PR #2565, opened 2016-11-13
Fixed uinstaller issues

SinghRajenM:
- Few files and folders are not deleted during uninstallation.
From InstallDir

1. Folder "localization" is not removed even all the files are removed from this folder.
2. Folder "pluging\APIs" is not removed even all the files are removed from this folder.
3. Three files are not removed (contextMenu.xml, functionList.xml and shortcuts.xml). There might be more files.
From "%appdata%\Notepad"
1. Bakcup folder (form %appdata%\Notepad++) is not removed if backup folder is not empty.
2. Folder "themes" is not removed even all the files are removed.
3. Because of above two items, parent folder ("%appdata%\Notepad") is also not removed.

- Rephrase user message when user tries to install 64 bit while 32 is already installed (vice versa).

### DIFF
--- a/PowerEditor/installer/nsisInclude/uninstall.nsh
+++ b/PowerEditor/installer/nsisInclude/uninstall.nsh
@@ -27,8 +27,10 @@
 
 Var themesParentPath
 Var doLocalConf
+Var keepUserData
 Function un.onInit
-  	; determinate theme path for uninstall themes
+	StrCpy $keepUserData "false"	; default value(It is must, otherwise few files such as shortcuts.xml, contextMenu.xml etc, will not be removed when $INSTDIR\doLocalConf.xml is not avaliable.)
+	; determinate theme path for uninstall themes
 	StrCpy $themesParentPath "$APPDATA\${APPNAME}"
 	StrCpy $doLocalConf "false"
 	IfFileExists $INSTDIR\doLocalConf.xml doesExist noneExist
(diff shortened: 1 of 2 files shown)


PowerEditor/installer/nsisInclude/uninstall.nsh
```diff
@@ -27,8 +27,10 @@
 
 Var themesParentPath
 Var doLocalConf
+Var keepUserData
 Function un.onInit
-  	; determinate theme path for uninstall themes
+  	StrCpy $keepUserData "false"	; default value(It is must, otherwise few files such as shortcuts.xml, contextMenu.xml etc, will not be removed when $INSTDIR\doLocalConf.xml is not avaliable.)
```
donho: If we initialize $keepUserData here...

PowerEditor/installer/nsisInclude/uninstall.nsh
```diff
@@ -116,7 +133,6 @@ Section un.UserManual
 SectionEnd
 
 
-Var keepUserData
 Function un.doYouReallyWantToKeepData
 	StrCpy $keepUserData "false"
```
donho: ... Why do we need to do it here again?

SinghRajenM: Right, it it not required in our case.
But it is always to good to reset the variable (or handle both Yes and No case) in a function which is meant to set/reset based on user interaction. Such a kind of functions can be called multiple times, so it is MUST to reset the value otherwise the variable may carry previous value if user clicks "No". 

In our case also the function is called twice (but in the flow it will be called only once). Let me know if you want me to remove it.

SinghRajenM:
@donho 
Is anything required from my end to get this PR merged?

PowerEditor/installer/nsisInclude/uninstall.nsh
donho: OK, let's keep it.

vlakoff:
With the added section "un.onUninstSuccess", isn't the "Remove remaining directories" part, at the very end of the file, now redundant?

Issue #1259, opened 2015-12-17
Double click word selection doesn't work after opening a link


fade0ff:
After opening a link from within NPP by double clicking on the URL in a text,  the next double click on a word will move the cursor to the end of the line instead of selecting and highlighting the word.
Only the next double click on a word will work as expected again.

PR #2569, opened 2016-11-14
Fix double click after double clicking hotspot

### LINKED ISSUE #2547: Can't select text properly after using links
1. Follow any link like
**https://notepad-plus-plus.org/fr/contribute/ text text**
2. Return focus to Noteped's window (previously) or select any text at once

dail8859:
Closes #2547. 

From [this section of code](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/src/Editor.cxx#L4210-L4214) it shows that the double click notification comes before the hotspot double click notification, so there is no need to store a flag for the next time `SCN_DOUBLECLICK` occurs.

### DIFF
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -383,7 +383,6 @@ private:
 
 	// For hotspot
 	bool _linkTriggered = true;
-	bool _isHotspotDblClicked = false;
 	bool _isFolding = false;
 
 	//For Dynamic selection highlight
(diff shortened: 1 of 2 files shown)


dail8859:
Also closes #1259

donho:
@dail8859 The fix could bring the side effect or regression. Let's see in the next release.

dail8859:
I spent some time digging through the git log and could not find the exact reason it was there. Is there a particular regression you know it might cause?

PR #2575, opened 2016-11-17
Added elif to the list of Inno Setup (.iss) preprocessor directive ke…

joeynovak:
…ywords.

Inno Setup has an "else if" preprocessor directive (elif).  See: http://www.jrsoftware.org/ispphelp/index.php?topic=funcs


chcg:
Diect link to description:
http://www.jrsoftware.org/ispphelp/topic_if.htm

PR #2585, opened 2016-11-19
Update for Python Lexer : add detect "end of line where string is not…

Cyrillev91:
Update for Python Lexer : add detect "end of line where string is not… closed" and "user defined keyword" already included in scintilla Python lexer

Close issue #2584

### 1. End of line where string is not… closed 
**before**
pythonlexer_before
**activate partially** (stylers.model.xml, without eolfilled):
pythonlexer_partial
**fully activated** (stylers.model.xml + enable eolfilled in ScintillaEditView.h):
pythonlexer_after

### 2. User defined keyword
sample
pythonlexer_userdefinedkeyword

Sources : 
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/lexers/LexPython.cxx
https://sourceforge.net/p/scintilla/scite/ci/default/tree/src/python.properties

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.h
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.h
@@ -742,6 +742,7 @@ protected:
 
 	void setPythonLexer() {
 		setLexer(SCLEX_PYTHON, L_PYTHON, LIST_0 | LIST_1);
+		execute(SCI_STYLESETEOLFILLED, SCE_P_STRINGEOL, true);
 	};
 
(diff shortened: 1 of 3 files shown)


MAPJe71:
I guess separate PR's would have been better.
Furthermore, why add a separate style for "USER DEFINED" keywords when it is already possible to enter user-defined keywords as part of the "KEYWORDS" style?

donho:
This behaviour is not present on the other language.
For the sake of consistence, it's better to keep the same behaviour everywhere.

PR #2606, opened 2016-11-25
Remove space before or after ":"

scootergrisen:
So all strings use the same style.


scootergrisen:
Why reject?
Would it not be better to use the same style in all string instead og extra spaces in some strings?

PR #2607, opened 2016-11-26
MAC to Mac and UNIX to Unix

scootergrisen:
So that is matches with:
```
                    <Item id="6403" name="Unix"/>
                    <Item id="6404" name="Mac"/>
```
I guess Unix could also be UNIX which ever way you prefer.


popcorner:
Actually, Mac might be an outdated name.
Apple called it Mac OS many years ago
and then renamed it to OSX

now it is called macOS
（change mac to lowercase and remove the space）

donho:

PR #2612, opened 2016-11-26
"Styler Dialog" to "Styler"

scootergrisen:
I suppose its not necessary to call a dialog a dialog in the title.


donho:

PR #2311, opened 2016-09-21
No notification if ShellExecute fails (Fixed#2179)

SinghRajenM:
If ShellExecute fails to execute command as if required software is not installed, then there is no notification to user.
This happens specially from "Run" menu (e.g. user clicks on Launch in Chrome while Chrome is not installed).

Accommodated reviews and prepared new PR. Original PR is #2181 and #2190.

shellexecute_failled

PR #2626, opened 2016-11-29
No notification if ShellExecute fails

SinghRajenM:
If ShellExecute fails to execute command as if required software is not installed, then there is no notification to user.
This happens specially from "Run" menu (e.g. user clicks on Launch in Chrome while Chrome is not installed).

Few review comments also have been accommodated as per #2311 .

image

### DIFF
--- a/PowerEditor/src/WinControls/StaticDialog/RunDlg/RunDlg.cpp
+++ b/PowerEditor/src/WinControls/StaticDialog/RunDlg/RunDlg.cpp
@@ -195,6 +195,28 @@ HINSTANCE Command::run(HWND hWnd)
 	expandNppEnvironmentStrs(argsIntermediate, args2Exec, args2ExecLen, hWnd);
 
 	HINSTANCE res = ::ShellExecute(hWnd, TEXT("open"), cmd2Exec, args2Exec, TEXT("."), SW_SHOW);
+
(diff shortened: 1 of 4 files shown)


PowerEditor/src/WinControls/StaticDialog/RunDlg/RunDlg.cpp
```diff
@@ -195,6 +195,28 @@ HINSTANCE Command::run(HWND hWnd)
...
+		errorMsg += TEXT("\n----------------------------------------------------------");
+		errorMsg += TEXT("\nCommand: ");
+		errorMsg += cmd2Exec;
+		errorMsg += TEXT("\nArguments: ");
+		errorMsg += args2Exec;
+		errorMsg += TEXT("\nError Code: ");
+		errorMsg += intToString(retResult);
+		errorMsg += TEXT("\n----------------------------------------------------------");
+
+		::MessageBox(hWnd, errorMsg.c_str(), TEXT("ShellExecute - ERROR"), MB_ICONERROR | MB_APPLMODAL);
```
donho: The big error icon makes users feel  the catasrophy is happening on Notepad++, whereas it's a small problem.
It's better to remove MB_ICONERROR IMO.

SinghRajenM: How about MB_ICONINFORMATION?
image

while only MB_OK looks as
image

PR #2631, opened 2016-11-30
Improve Folder Browse And Save Dialog

Motaz-Alnuweiri:
**Folder Browse & Save Dialog**
Enable the new user interface for `Folder Browse Dialog` including the edit box and enable Vista style for `Save Dialog`.

Folder Browse And Save Dialog

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -174,6 +174,11 @@ Notepad_plus::Notepad_plus()
 		is_admin = false;
 
 	_isAdministrator = is_admin ? true : false;
+
+	// Enable new style save dialog.
+	if (ver >= WV_VISTA) {
+		(NppParameters::getInstance())->setUseNewStyleSaveDlg(true);
+	}
 }
 
 Notepad_plus::~Notepad_plus()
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -174,6 +174,11 @@ Notepad_plus::Notepad_plus()
 		is_admin = false;
 
 	_isAdministrator = is_admin ? true : false;
+
+	// Enable new style save dialog.
+	if (ver >= WV_VISTA) {
```
SinghRajenM: Why new style dialog is applicable only for vista and higher version?

As per [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/bb773205(v=vs.85).aspx) , minimum version of Shell dll should be 5.0 and 6.0 for BIF_USENEWUI  and BIF_NONEWFOLDERBUTTON respectively. So I think XP also should be covered.

image

Motaz-Alnuweiri: @SinghRajenM  No this check is for `Save Dialog` only, because the save dialog style in the image above is introduced in Windows Vista.

For the `Folder Browse Dialog` I already tested it on Windows XP and it's work perfectly, but I forgot to say that.

donho: @SinghRajenM I don't think last version of Notepad++ can be run under XP.

Issue #1874, opened 2016-05-20
Opening a blank file should use same encoding as for a new document


vlakoff:
Use case: I often create a new file by right-click in Windows Explorer, then open this file with Notepad++.
- When creating a new file in Notepad++, it uses the `New Document > Encoding` preference (default UTF-8 w/o BOM).
- Furthermore, if I open a file containing only ANSI characters, it uses UTF-8 as well.
- BUT, if I open a blank file, it doesn't use UTF-8, but ANSI...

This is inconsistent and definitely has had caused me some troubleshooting headaches!

jtrent238:
So much problem. :pensive:

vlakoff:
@jtrent238, I'm unsure, is it irony? Seriously, this charset thing already has caused me many troubles…

jtrent238:
@vlakoff Does it do it on another computer? :question: 

> Don't have another computer Try [Virtual Box](http://bit.ly/OVMVB)!

vlakoff:
I encounter the same behaviour using a fresh settings profile.

jtrent238:
You got some screenshots?

molhanec:
I run to this also. Notepad++ is my favourite text editor and this is for me quite annoying because I rarely create new files directly in it. @vlakoff  thanks for reporting. (I don't know if jtrent238 is trying to be funny, but he is clearly not).

mzso:
[**@jtrent238**](https://github.com/jtrent238) commented on [2016. máj. 21. 22:07 CEST](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1874#issuecomment-220797782 "2016-05-21T20:07:08Z - Replied by Github Reply Comments"):
> You got some screenshots?

Why are you spamming? Try it yourself, and you'll see he's right.

I also got unpleasantly surprised by this a couple times...

mzso:
So many new features in v7. Yet no capacity to fix this small, but nasty bug.

jtrent238:
### **It don't do it for me.**

### **I haven't had an issue and I have not updated it.**


![](http://puu.sh/rkta7/e919516630.png)

mzso:
[**@jtrent238**](https://github.com/jtrent238) commented on [2016. szept. 23. 00:10 CEST](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1874#issuecomment-249043006 "2016-09-22T22:10:34Z - Replied by Github Reply Comments"):
[...]

PR #2633, opened 2016-12-01
Copy encoding fixes from loadFile() to reloadBuffer()

vlakoff:
(noticed this while investigating #1874)

Copy some code from `FileManager::loadFile` to `reloadBuffer`, to fix two bugs also when reloading file from disk. I think the bug 1 is quite important, as it affects "reload from disk" in a very common scenario (ANSI-only content, but force-detected as UTF-8 without BOM).

Bug 1:
1. The settings "New Document > UTF-8 + Apply to opened ANSI files" should be checked.
2. Create a file with only ANSI content (e.g. "aaa"). Save this file in ANSI. Close the file.
3. Open the file. It should be detected as UTF-8 (without BOM).
4. Execute menu command "Reload from Disk". The file is now detected as ANSI. Expected: should be detected as UTF-8.

Bug 2:
1. Create a file with this exact content:
    `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"> ééé`
2. Save it in ANSI and with a `.html` extension. Close the file.
3. Open the file. It should be detected as UTF-8 (and display garbage but that's "expected").
4. Execute command "Reload from Disk". The file is now detected as Windows-1255. Expected: should be detected as UTF-8.

<br>By the way, maybe we should drop the `getHtmlXmlEncoding` method. For instance look at the HTML regexes, they are outdated (no html5), not flexible and in most cases wouldn't match. Also, I think this method does more harm than good. These &lt;meta&gt; tags don't determine the file encoding, but the *expected* file encoding.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -689,13 +689,21 @@ bool FileManager::reloadBuffer(BufferID id)
 
 	if (res)
 	{
+		NppParameters *pNppParamInst = NppParameters::getInstance();
+		const NewDocDefaultSettings & ndds = (pNppParamInst->getNppGUI()).getNewDocDefaultSettings();
+
 		if (encoding == -1)
 		{
-			buf->setUnicodeMode(UnicodeConvertor.getEncoding());
+			UniMode um = UnicodeConvertor.getEncoding();
+			if (um == uni7Bit)
+				um = (ndds._openAnsiAsUtf8) ? uniCookie : uni8Bit;
+
+			buf->setUnicodeMode(um);
 		}
-		else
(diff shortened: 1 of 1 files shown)


vlakoff:
Of course, you might want to factorize this code from `loadFile` and `reloadBuffer` into a separate method.

donho:
@vlakoff  Have you tested your PR?
It doesn't work for me.

vlakoff:
I cannot compile, I just copied the existing code.

* Can you reproduce the bugs, at least the first one?
* What's your opinion about dropping `getHtmlXmlEncoding` as I described at the end of my first post?

donho:
I can reproduce the bug. But your PR doesn't work.
Please submit ONLY pr that you have tested. Thank you.

Issue #2641, opened 2016-12-04
Cannot open files when you drag them into Notepad++.


AraHaan:
### Description of the Issue
Since the update from 7.1 to 7.2.2 opening a file via drag and drop is broken.

### Steps to Reproduce the Issue
1. Update to 7.2.2
2. Open Notepad++ v7.2.2 after installing.
3. Drag and drop any file into editor and it wont open.

### Expected Behavior
To open the file from dragg & drop just like v7.1 allowed me to.

### Actual Behavior
Files dragged & dropped into Notepad++ are not opened.

### Debug Information

```
Notepad++ v7.2.2   (32-bit)
Build time : Nov 27 2016 - 16:31:07
Path : E:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7
Plugins : MenuIcons.dll mimeTools.dll NppConverter.dll nppcrypt.dll NppExec.dll NppExport.dll NppFTP.dll NPPJSONViewer.dll NppSaveAsAdmin.dll PluginManager.dll PyNPP.dll Python Indent.dll PythonScript.dll ReloadButton.dll Translate.dll ZenCoding-Python.dll zoomdisabler.dll 
```

dail8859:
Works fine for me...

```
Notepad++ v7.2.2   (32-bit)
Build time : Nov 27 2016 - 16:31:07
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7
Plugins : ComparePlugin.dll DoxyIt.dll LuaScript.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 
```

SinghRajenM:
There might be a case that Notepad++ is running as administrator (can see from debug info), while your explorer is running with lower privileges. I assume that you may have checked "open Noptepad++" on finish dialog which will elevate N++ (run with admin privilege).

If this is the case, then it is windows expected behavior. However, there are ways to handle it but again that will be lead to security issues otherwise why MS would have blocked it.
Some links for your reference - 
[stackoverflow](http://stackoverflow.com/questions/8239271/uac-elevation-does-not-allow-drag-and-drop)
[...]

Issue #556, opened 2015-07-24
Restart As Regular User After Installation


AtmaX:
Can Notepad++ restart itself as the regular logged on user after installing/updating itself, instead of as admin?

ariccio:
Duplicate of: [#159: Installer should drop privs when executing the installed binary](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/159)

Related: [#334: Notepad++ elevated save privileges](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/334)

Issue #159, opened 2015-06-03
Installer should drop privs when executing the installed binary


shuffle2:
The installer needs admin privileges in order to work, but then it runs notepad++.exe as admin. This is annoying (mainly because it breaks drag and drop).

ariccio:
It's even worse than drap & drop.

Notepad++ running with administrator privileges is a security nightmare. If someone exploits Notepad++, then they're free to literally do anything on the victim's computer.

shuffle2:
Well, that's not really an issue. Sometimes you need to run notepad++ as admin. It just shouldn't do this when autostarting after install.

donho:
The installer need administrator privileges to install Notepad++.
Notepad++ is launched by installer after the installation hence it inherits administrator privileges.
So actually I don't see any way to drop this privileges.
If anyone get the idea, let me know. I'll implement it.

The2rB:
My opinion:
It would be better to leave it closed after installation than to open it as administrator.
I were unable to open (any) files in Notepad++ after installation due to it running as administrator. Took a while before understanding what was going on. This made me wonder if the update had broken Notepad++.

milipili:
Using this plugin would not be better ? http://nsis.sourceforge.net/UAC_plug-in

dg9ngf:
I don't know what installer you use, but InnoSetup long has an option to start the process non-elevated. It does so because it keeps the non-elevated initial setup process around, which starts both the elevated setup, and the final application (in the original session).

GitMensch:
@milipili Yes, the UAC plugin seems to do what's necessary:

> The only thing the UAC plugin was designed to do is allowing the finish page's Run checkbox to launch the app at userlevel, after elevation has occured for the installer.

In any case: MS's "design" intend for installers is: start non-elevated (either as admin or as normal user), as soon as the rights will be necessary (for example you've checked a possible "install for all users" checkbox and/or install to ProgramFiles) present a button to the user for example "finish" with the UAC logo. If clicked you'll get an UAC prompt and this installation part will go on elevated (while the normal part stays alive).

The important part here is to only do the stuff in the elevated part which has to be elevated, see [UAC-Guidelines for Packages](https://msdn.microsoft.com/en-us/library/windows/desktop/aa368772%28v=vs.85%29.aspx) (applies to Windows Installer Packages but the principle is the same).

camthesaxman:
[...]

PR #2409, opened 2016-10-09
Include version info in Installer

SinghRajenM:
As of now, there is no version info available on the detail section of installer properties.
It is good to have version info.

file_info

PR #2643, opened 2016-12-04
Launch N++ with same privileges as explorer

SinghRajenM:
Fixes: #2641, #159, #556

Known Pending Action Item:
A shortcut is created in %temp% which should be deleted. But I didn't find way to handle it.
However, we can check for "nppInstance" mutex in a loop, but again keeping installer open (that too on finish page->finish page) for some more time (until N++ launched) is not good idea. Let me know if checking mutex if fine, I will update PR.

### DIFF
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -26,7 +26,22 @@
 ; Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
 Function LaunchNpp
-  Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '
+  ;Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '
+
+  ; create shortcut in temp with to launch as argument
+  CreateShortCut "$TEMP\notepad++.lnk" "$INSTDIR\notepad++.exe" "$INSTDIR\change.log"
+
+  ; Launch Notepad++ with same elevation as explorer
+  ; It will solve the drag an drop issue if explorer is running with lower privilege
+  Exec '"$WINDIR\explorer.exe" "$TEMP\notepad++.lnk"'
+
+  ; Caution : Please DONOT delete $TEMP\notepad++.lnk right after Exec command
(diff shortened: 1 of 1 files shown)


MAPJe71:
Why not use the [UAC plugin as mentioned by @milipilli](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/159#issuecomment-127913796)?

SinghRajenM:
UAC is very famous plugin and it is used widely. It should have been used in N++ as well. However, @donho  mentioned in his [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/159#issuecomment-125151670) if anyone gets idea then let him know, he will implement it.  So I wonder why it was not implemented till then?

It might be the case that 
1. @donho may not want an additional nsis plugin in repository as UAC plugin is NOT included in nsis by default.
2. I tried to automate the versioning ( #2409 ) by using an additional binary and some code which was not merged due to size for code [refer comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2409#issuecomment-260113923). So I wanted (infect want in every PR) to minimize the code, but using UAC plugin requires some code along with plugin dll in repository.

SinghRajenM:
UAC is very famous plugin and it is used widely. It should have been used in N++ as well. However, @donho  mentioned in his [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/159#issuecomment-125151670) if anyone gets idea then let him know, he will implement it.  So I wonder why it was not implemented till then?

It might be the case that 
1. @donho may not want an additional nsis plugin in repository as UAC plugin is NOT included in nsis by default.
2. I tried to automate the versioning ( #2409 ) by using an additional binary and some code which was not merged due to size for code [refer comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2409#issuecomment-260113923). So I wanted (infact want in every PR) to minimize the code, but using UAC plugin requires some code along with plugin dll in repository.

donho:
@SinghRajenM 
> However, we can check for "nppInstance" mutex in a loop, but again keeping installer open (that too on finish page->finish page) for some more time (until N++ launched) is not good idea. Let me know if checking mutex if fine, I will update PR.

I don't think it's a good idea to make a complex solution to remove a small shortcut from tmp.

Issue #2125, opened 2016-07-26
VHDL - Block Uncomment doesn't function


eebydesign:
### Description of the Issue

I presume this problem has been around for a while.  I found many instances of other folks seeing the same that the block uncomment for vhdl does nothing but the block comment does work.  Unfortunately if one selects block comment then realizes block uncomment does nothing then try and use the toggle single line comment, the results aren't 100% correct and one then needs to go back through each line to make sure what is commented really is what should be commented.  An example of this is if there were comments in the code, after doing block comment then toggle single line comment, these commented notes in the code are no longer commented.
### Steps to Reproduce the Issue

1.  Select some code (test.vhd)
2.  Right click Block comment
3.  Select same block of code.
4. Right click Block Un-comment
### Expected Behavior

Code should revert back to original code.
### Actual Behavior

Code stays commented.
### Debug Information

Notepad++ v6.9.2
Build time : May 18 2016 - 00:34:05
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll NppVHDL.dll PluginManager.dll

PR #2647, opened 2016-12-05
Potential fix to issue #2125, block uncomment feature not working.

blounsbury:
Notepad_plus::doStreamComment() had code to check for block comments and call Notepad_plus::doBlockComment(comment_mode). So Block Comment is working. However, undoStreamComment() was missing this code so Block Uncomment was not working.


blounsbury:
Any chance of this getting accepted or reviewed? It's a rather major bug that Block Uncomment does not function.

donho:
It's fixed or not, there's no "Potential".

blounsbury:
It's fixed based on my testing, but I thought someone might want to review the edits. That's why I said "Potential". Thanks for the feedback.

Issue #2573, opened 2016-11-15
Doc Switcher: context menu option doesn't make sense in vertical...


travisleithead:
### Description of the Issue
When triggering a context menu from the doc-switcher, the menu items labelled "Close All to the Left" and "Close All to the Right" don't make sense--since doc-switcher files are vertically-oriented.

![image](https://cloud.githubusercontent.com/assets/2152696/20318267/6704fca4-ab1e-11e6-86af-c07a9a36f2cb.png)

### Steps to Reproduce the Issue
1. Open a few documents
2. Enable the doc-switcher
3. Right click on a single document in the switcher and observe the menu items

### Expected Behavior
I would like to see the menu item names adjusted in this case to better match the semantics ("Close All to the Left" should be changed to "Close All above" and "Close All below" or some such.)

### Debug Information
Windows build 7.1 (32-bit)

Issue #2276, opened 2016-09-13
Alt+O does not trigger "Replace All in All Opened Documents"


MartinNFuller:
### Description of the Issue

In the Replace dialog (Ctrl+H), the keyboard shortcut Alt+O does not trigger "Replace All in All Opened Documents".
### Steps to Reproduce the Issue
1. Open 1+ documents
2. Ctrl+H
3. Enter some search text that appears in the documents
4. Alt+O
### Expected Behavior

Alt+O should have the same effect as clicking the button "Replace All in All Opened Documents"
### Actual Behavior

Alt+O has no effect.
### Debug Information

Notepad++ v6.9.2
Build time : May 18 2016 - 00:34:05
Path : D:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll

krilbe:
Please note that #3699 would probably solve this issue. Clashes between Alt shortcut letters from invisible tabs should not be an issue, i.e. Notepad++ should never, in any dialog with multiple tabs, have shortcuts on invisible tabs swallow Alt letter shortcuts that appear on the active tab. It's not realistic to require that all Alt shortcuts be unique across all tabs. Being unique on the active tab should suffice and work as expected.

Issue #2146, opened 2016-08-05
Quit after closing last document


ygoe:
I'd like to have an option to quit Notepad++ entirely when the last document has been closed. I'm using the "session restore" feature that always opens the documents at start that were open when Notepad++ was last closed. Much like my web browsers do. But when I don't need any documents permanently open, when I open one document and then close it again, I don't need Notepad++ staying around anymore. It should be closed then.

Right now I need to press Ctrl+W and then Alt+F4 to achieve this. Ctrl+W should be enough.

Another option would be not to close it but to minimise the window. I've seen this in TextPad but wouldn't use it today.

MAPJe71:
Possibly a duplicate of #1006 which was solved with pull request for #1923.

ygoe:
The issue itself is a duplicate but the solution is not. What should I do with a command line parameter? Nothing will ever specify that. It must be an option in the application itself to be useful.

SanyaWaffles:
@ygoe ygoe, are you using Windows 10 by chance.

ygoe:
I have indeed upgraded my computers to Windows 10 by now but the issue already existed on Windows 7 before. I don't think it's related to the Windows version, it should be purely application behaviour.

dail8859:
Closed by e31e37ebc3bdd8d34b0cb61f1e3b81477fcdfbaf

Issue #1978, opened 2016-06-17
rename will not move a file to a new drive


lcarlp:
I often use rename to move files between folders and that usually works fine, but I am noticing rename does not work when I try to use it to move a file from one drive to another, i.e., to a different drive letter.  The original file remains in place (good!).  The file is not copied to the target drive (not good),  And there is no error message or any indication that it did not work (bad).

I have tried this both using a drive letter for the target and also using a fully qualified UNC name.  It seems to behave the same in both cases.

I'm running N++ v6.9.2 on WIndows 7 Professional.  

I only recently started trying to do this, so it is possible this has never worked.  I can use "Save as" as a workaround, but I need to do another step to remove the original file, so it would be cleaner for me if I could use Rename.

I'm not shocked this does not work, since the Windows command-line "rename" command cannot be used to do this, but it can be done using a "move" command, and I normally find N++ does what I want even if Windows commands don't.  It's am amazingly reliable and powerful utility.

Thanks.

asanetargoss:
Can confirm this is still a bug in Notepad++ 7.2 (32 bit version) on Windows 7 SP1 64-bit

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
search if related topics exists. you are welcome to post either at similar
topics, or to create a new topic at Help Wanted or General Discussion.
(no extra account is needed, just use your github account to sign in)
```
[...]

PR #2649, opened 2016-12-07
Master

MawRojas:
Fixed / Added
rename will not move a file to a new drive 
       #1978
Alt+O does not trigger "Replace All in All Opened Documents"
      #2276
Doc Switcher: context menu option doesn't make sense in vertical
      #2573
Quit after closing last document - Added checbox in settings
      #2146
Change "cloud" terminology
      #928


dail8859:
I think I can safely assume this was a mistaken pull request. :)

Issue #2277, opened 2016-09-13
Do not add file to Recent Files List if it is just moved to the other view


pnedev:
### Description of the Issue

File is added to the _Recent Files List_ when it is only moved to the other view.
The behavior is wrong because the file is not actually closed.
### Steps to Reproduce the Issue

1. Open N++ and clear the _Recent Files List_.
2. Open two files.
3. Move one of them to the other view.
4. Look at the _Recent Files List_ - the moved file is added there.
### Expected Behavior

Just move the file to the other view without adding it to the _Recent Files List_.
### Actual Behavior

File is added to the _Recent Files List_ when it is only moved to the other view.
### Debug Information

When moving the file to the other view it is first closed in the current view which causes it to be added to the _Recent Files List_.

SinghRajenM:
Additional info:
If NPP is restarted, the file which was added to _Recent List_ is removed from the list.

Yaron10:
Hello,

Some simple changes in three NPP source files seem to solve this issue.

**In Notepad_plus.h replace**
`void doClose(BufferID, int whichOne, bool doDeleteBackup = false);`
**with**
`void doClose(BufferID, int whichOne, bool doDeleteBackup = false, bool addToClosedList = true);`

**In Notepad_plus.cpp replace**
`//just close the activate document, since thats the one we moved (no search)
doClose(_pEditView->getCurrentBufferID(), currentView());`
**with**
`//just close the activate document, since thats the one we moved (no search)
doClose(_pEditView->getCurrentBufferID(), currentView(), false, false);`

**In NppIO.cpp replace**
`void Notepad_plus::doClose(BufferID id, int whichOne, bool doDeleteBackup)`
**with**
`void Notepad_plus::doClose(BufferID id, int whichOne, bool doDeleteBackup, bool addToClosedList)`

**and also replace**
`if (PathFileExists(buf->getFullPathName()))`
**with**
`if (PathFileExists(buf->getFullPathName()) && addToClosedList)`

Thank you.
BR

SinghRajenM:
Fixed issue in #2665

pnedev:
Hello @SinghRajenM ,

Great, thank you for the fix.

Hello @donho ,

Thanks for accepting the fix so promptly.

BR

PR #2665, opened 2016-12-11
Add to recent file history only if file is removed from all the views

SinghRajenM:
Fixed issue: #2277 
There might be cases when file is cloned/moved to view. Don't add to recent list unless file is removed from all the views.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -611,7 +611,10 @@ void Notepad_plus::doClose(BufferID id, int whichOne, bool doDeleteBackup)
 	scnN.nmhdr.idFrom = (uptr_t)id;
 	_pluginsManager.notify(&scnN);
 
-	//add to recent files if its an existing file
+	// Add to recent file history only if file is removed from all the views
+	// There might be cases when file is cloned/moved to view.
+	// Don't add to recent list unless file is removed from all the views
+	generic_string fileFullPath;
 	if (!buf->isUntitled())
 	{
 		// if the file doesn't exist, it could be redirected
@@ -626,7 +629,7 @@ void Notepad_plus::doClose(BufferID id, int whichOne, bool doDeleteBackup)
 		}
 
 		if (PathFileExists(buf->getFullPathName()))
-			_lastRecentFileList.add(buf->getFullPathName());
+			fileFullPath = buf->getFullPathName();
 
 		// We enable Wow64 system, if it was disabled
 		if (isWow64Off)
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -611,7 +611,10 @@ void Notepad_plus::doClose(BufferID id, int whichOne, bool doDeleteBackup)
 	scnN.nmhdr.idFrom = (uptr_t)id;
 	_pluginsManager.notify(&scnN);
 
-	//add to recent files if its an existing file
+	// Add to recent file history only if file is removed from all the views
+	// There might be cases when file is cloned/moved to view.
+	// Don't add to recent list unless file is removed from all the views
+	generic_string fileFullPath(TEXT(""));
```
donho: The declaration of variable should be more near as possible to the first usage of variable.

PowerEditor/src/NppIO.cpp
```diff
@@ -626,7 +629,7 @@ void Notepad_plus::doClose(BufferID id, int whichOne, bool doDeleteBackup)
 		}
 
 		if (PathFileExists(buf->getFullPathName()))
-			_lastRecentFileList.add(buf->getFullPathName());
+			fileFullPath += buf->getFullPathName();
```
donho: Rather `fileFullPath = buf->getFullPathName();` IMO

SinghRajenM: It should be not be an issue as variable is declared with empty string. Let me know if it really needs to be changed.

Yaron10:
Hello @SinghRajenM,

Great work. 👍 
Shorter and better than my suggestion. :)

Thank you very much. I do appreciate it.
***
Hello @donho,

Thank you for reviewing and accepting it.

Best regards.

PowerEditor/src/NppIO.cpp
donho: OK, I didn't see it's used in another block. In this case, you're right.
OTOH, you don't need to do `generic_string fileFullPath(TEXT(""));`
`generic_string fileFullPath;` will create an empty string.

PowerEditor/src/NppIO.cpp
donho: true, but it's rather a sementic issue: 
`+=` is a concatanate operation, that's make us think there's already a string in it and we want to keep it.
So yes, please change it to make the intention more clear.

SinghRajenM: Done!

PowerEditor/src/NppIO.cpp
SinghRajenM: Done! used generic_string fileFullPath;

PR #2669, opened 2016-12-12
Extension of Finders

eglibicky:
Extension of finders, the result of "Find in Files" and "Find All in Current Document”.
This is an extension of finders, the result of "Find in Files" and "Find All in Current Document”.
The finders will have a name, that is the name of the search.
Four options will be added:
"Same name merge" will replace existing finder will a new result of search with the same name
"Only one finder" will create and maintain only one finder and replace it every time when there is a new "Find all" or "Find in Files" searches
"Show line once" will show only one line of code in finder, if multiple finds in the same line
"Close empty" will close all finders with 0 finds on the next search

In addition to that, a new close all finders menu option will be available in finder menu.
 master (#1)


cmeriaux:
can you clean your commit history and your commit messages ?
You must have 'merge pull request ...' in your commit history. Use branch and rebase instead of merge

PR #2684, opened 2016-12-13
Finder options

eglibicky:
Addition of Finders, the results of multiple searches of Find All in This Document or Find In Files


cmeriaux:
There is still "merge commit" that shouldn't be there. Use rebase -i (interactive) to keep useful commit only

Issue #364, opened 2015-06-29
Style configuration doesn't remember font & size


kobalicek:
I'm having an issue with Style Configurator.

I switched to a dark theme (obsidian), but when I close notepad++ and run it again it shows a wrong font (not monospace at all). If I go to "Style Configurator - Global Styles - Default Style" and set the font to "Courier New" the problem gets fixed for the current session, but even if I "Save & Close" next time I run notepad++ I have to do the same, the configuration is lost.

milipili:
Maybe related to #220

kobalicek:
Not sure, I selected "Courier New", which doesn't seem to have any unicode issues.

Tmp341:
Do you have admin rights?

kobalicek:
I think it has nothing to do with admin rights. It remembers the dark theme I selected, but it screws up the font by some reason.

kobalicek:
But to answer to your question - yes, I have.

Tmp341:
I had this issue that when i try to install a theme from desktop, it doesn't show up in the style config.I had to go to %Appdata%\Notepad++\themes and paste them.Didn't check it with the latest. Try this, go to %Appdata%\Notepad++\themes and open obsidian with npp. See the default style's fontName. And change it through style config and see if it changes there too.

alfrrr:
@kobalicek I posted an answer inside the issue #1299 that maybe fixes your problem

mjohnsonengr:
I have the same issue with the same theme -- Obsidian -- when I open Notepad++ as administrator, the font renders correctly as a monospaced font (Courier New I presume), but definitely not monospace in normal non-admin mode:

![image](https://cloud.githubusercontent.com/assets/7274059/13366667/04322118-dc9a-11e5-81af-27fa88d46d9b.png)

donho:
Cannot reproduce it.

xates:
@donho This happens only with Obsidian theme and you need to close and reopen Notepad++ to notice that. My patch #2695 that you rejected actually fixes this bug.

ttreadwellscs:
I have the same problem trying to use downloaded font 'M+ 1m medium' in Obsidian theme. Every time I restart Notepad++ I have to go to Settings > Style Configurator and reselect Obsidian theme to get the M+ font to show up. I tried the patch #2695 but it did not solve it.

QAnders:
+1
I have this problem too...

MichaelEugeneYuen:
[...]

PR #2695, opened 2016-12-15
Fixed Obsidian theme bug (issue #364)

xates:
Fixed this bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/364


xates:
Can I know why?

gamax92:
Also curious why this was rejected, almost 2 years later this is still an issue. Opening Notepad++ with the Obsidian theme has the wrong font and size and this patch still fixes the issue.

xates:
¯\_(ツ)_/¯

reyandme:
Bug is still there, sad

ner00:
In my case I found that I had permission issues with theme folders under AppData, there were more than one theme folder so I'm not sure which one was bad, but when I tried to delete the folders I didn't have permissions, which is weird, but overriding the permissions for AppData's `Notepad++/*` did the trick.

PR #2702, opened 2016-12-18
Update turkish.xml

gamzeakmann:
I added this line after  I opened issue # 2701.
Item id="44085" name="Çalışma Alanı Olarak Klasör" added this line.


mertcalik:
@gamzeakmann It seems correct. Can you verify these changes?

donho:
@gamzeakmann  it seems #2714 is the correct one. If it is, please close this one. Otherwise close this one.

PR #2725, opened 2016-12-28
Bulgarian language update [fixed]

rddim:


rddim:
added translation for menu `Tools` and submenu `MD5`

Issue #2600, opened 2016-11-23
NPP does not show toolbar if opened from system tray


SinghRajenM:
### Description of the Issue

Use minimize to system tray functionality and then try to open the npp. Toolbar is not visible.
### Steps to Reproduce the Issue

1. Open NPP (make sure toolbar is visible) and enable minimize to system tray
              Settings->preferences->MISC->Minimize to system tray
2. Now minimize the NPP
3. Try to launch npp from system tray
     right click on npp system tray icon, then click on either "Activate" or "Open...". If click on open, then don't open any file, simply click on cancel on file selection dialog.

### Expected Behavior

Toolbar must be visible.

### Actual Behavior

Toolbar is not visible.

### Debug Information

Notepad++ v7.2.1   (32-bit)
Build time : Nov 20 2016 - 22:42:51
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : DSpellCheck2.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

donho:
@SinghRajenM I cannot reproduce it from my Windows 10 machine with 32-bit as well.

vlakoff:
I can reproduce it. Windows 7 (64-bit) with Notepad++ 7.3.1 (32 bit).

![notepad-plus-plus_2600](https://cloud.githubusercontent.com/assets/544424/22173351/4958722c-dfc0-11e6-96c8-1507a110b3d0.png)

Actually there are several bugs:
* Toolbar missing
* Plugin panes missing (Light Explorer and SourceCookifier), but look at the small bits of them at top left and top right of window
* Status bar missing background

<br>If this can help to debug, I noticed the bug happens only the first time:
* Start Notepad++, minimize to tray, restore : bug
* Minimize again to tray, restore : still bug
* Switch window size from "maximized" to "normal" (then back to maximized if you wish):
  * fixes the display
  * the bug doesn't occur anymore when minimizing to tray / restoring

PR #2729, opened 2016-12-29
Fixed issue: show toolbar while launching from system tray

SinghRajenM:
Fixed issue: #2600 

Show toolbar (if applicable, here applicable mean show if toolbar enabled otherwise not) while launching from system tray.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2828,6 +2828,9 @@ void Notepad_plus::command(int id)
 		{
 			NppGUI & nppGUI = const_cast<NppGUI &>((NppParameters::getInstance())->getNppGUI());
 			::ShowWindow(_pPublicInterface->getHSelf(), nppGUI._isMaximized?SW_MAXIMIZE:SW_SHOW);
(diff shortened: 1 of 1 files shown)


donho:
Bug #2600 cannot be reproduced.

SinghRajenM:
@donho The issue is reproduced every time with 32 bit N++ version on Win10 64bit OS. Make sure N++window is maximized before clicking on minimize. Refer attached video capture and PSR (Problem Steps Recorder).

toolbarnotvisible
[ToolBarNotVisible.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/727581/ToolBarNotVisible.zip)

Let me know if you need any info analyze the issue.

donho:
With minimalize package under windows 10, I cannot reproduce it:
https://notepad-plus-plus.org/repository/7.x/7.3.1/npp.7.3.1.bin.minimalist.7z

Are you sure it's not plugins issue?

SinghRajenM:
Yes, it is be reproduced 100% even without single plugin. Refer the screenshot below from the package you shared.

image

PR #2737, opened 2017-01-01
Update german.xml

d-ahrens:
This wrong translation bothered me for months now.
"Assembly (language)" should be translated with "Assembler (Sprache)" and not with "Zusammengesetzt".
Compare https://en.wikipedia.org/wiki/Assembly_language and https://de.wikipedia.org/wiki/Assemblersprache


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2741

PR #2745, opened 2017-01-03
Update italian to v7.3

Ices-Eyes:
Update italian translation according to version 7.3 (MD5 tool)

### DIFF
--- a/PowerEditor/installer/nativeLang/italian.xml
+++ b/PowerEditor/installer/nativeLang/italian.xml
@@ -1,7 +1,7 @@
 <?xml version = "1.0" encoding = "utf-8" ?>
 <!--
-    Italian translation for Notepad++ 7.2.0
-    Last modified Sat, Nov 5th, 2016 by Luca Leonardi.
+    Italian translation for Notepad++ 7.3.0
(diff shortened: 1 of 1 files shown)


donho:
@Ices-Eyes I just add a new feature and several new translations:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6a05bedd05cbfce40106aa464e43448639d1913b#diff-2a8496e2537c34b95cbeca007cd2346b

Could you update the PR please? (so I can include it in new release)

Ices-Eyes:
Ok, will have them translated as soon as I can, I think today or tomorrow. Can I ask you some more insight on the new Word character list function? Just to be sure on how the translation need to sound like :)

donho:
@Ices-Eyes "Word character list" is a set of charcter for word.
For example: `A-Z` `a-z` `0-9` are parts of this set; `+-*/$?!` are not.
With the default set, when you double click on `this-is-a-word`, the whole word "this-is-a-word" won't be selected due to the `-`. 
The feature allows user to add '-' into the word character set, so the whole word "this-is-a-word" will be selected while double clicking on it.

PR #2751, opened 2017-01-05
Update bulgarian.xml to 7.3.1

rddim:


rddim:
is there any reason that @donho didn't accept this commit?

Issue #260, opened 2015-06-19
How to avoid disabling unstable plugins during update?


effemmeffe:
I use the Hex editor plugin.
Every time NPP get an update it disable that plugin due to instability, moving the file to the disabled folder.
And every time I have to move it back.
I don't know what are those instability, nor I do care.
I want to use that plugin and not being forced to fiddle with files at every upgrade.
Is there some way I can tell NPP not to disable unstable plugins?
TIA

milipili:
Proposing a list of *invalid* plugins with checkboxes before performing the action might be a solution.

pnedev:
@milipili, Are all plugins checked for validity and how is that accomplished if new plugins appear?

Wouldn't it be better to remove this check altogether? After all, the Plugin Manager shows the plugin stability when the user wants to install it. Also as it is now, the MultiClipboard plugin is considered to be unstable but it is actually working just fine. And there is no info why is it flagged as invalid.

effemmeffe:
The consistency of the plugin is beyond the scope of this question, but since pnedved introduced the stability of another plugin I can add that in my experience the Hex editor plugin is completely stable, I used it a lot and never found any issue. YMMV.

maddes-b:
The [Hex Editor plugin][1] works for me without any hassles.
I'm using [Notepad++][2] 6.8.8 with [Hex Editor][3] 0.9.5.

I only found one issue with the Hex Editor plugin, which is solvable, where Notepad++ fails to start correctly.
Notepad++ can be seen in Task Manager and that it is consuming more and more memory.
The reason for this behaviour is that the directory `%APPDATA%\Notepad++\plugins\config` is missing (e.g. for a newly created user).

To solve this:

 - End all the user's Notepad++ processes via Task Manager
 - Create the missing directory either via Windows Explorer or via the following command line<br>
`mkdir %APPDATA%\Notepad++\plugins\config`
 - After this Notepad++ will start normally again

Still it is a [bug in Hex Editor][4].
But in my point of view nothing to call the Hex Editor plugin unstable.
An option to exclude it from the "disable plugins check" of Notepad++'s installer would be great.

  [1]: http://docs.notepad-plus-plus.org/index.php?title=Plugin_Central#H
  [2]: https://notepad-plus-plus.org/
  [3]: https://sourceforge.net/projects/npp-plugins/files/Hex%20Editor/
  [4]: https://sourceforge.net/p/npp-plugins/bugs/215/

[...]

PR #2752, opened 2017-01-05
Do not remove plugins considered unstable on new install

pnedev:
As plugins are updated/re-built outside N++ it cannot know for sure which plugin is unstable so
don't touch installed plugins on new N++ install.


pnedev:
This fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/260.

AraHaan:
Hmm I got better idea what if the ``updater`` did not care about plugins so that way we dont have to worry about this false ``unstability`` issue.

donho:
The list of unstable plugins could be discutable, but there are some plugins are definitely unstable and they are must be on the list for the sake of stability of Notepad++.

PR #2800, opened 2017-01-17
shortcut mapper clear command

### LINKED ISSUE #2799: Shortcut mapper - add clear command
### Description of the Issue

In the Shortcut mapper,  removing a shortcut is very annoying.
Add au button to clear a shortcut.

cmeriaux:
In the shortcut mapper, shortcuts can be cleared easily

fixes #2799


PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -352,6 +357,120 @@ INT_PTR CALLBACK ShortcutMapper::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
+							_lastCursorRow[_currentState] = _babygrid.getSelectedRow();
+							fillOutBabyGrid();
+
+							isModified = true;
+
+							//Notify current Accelerator class to update everything
+							nppParam->getAccelerator()->updateShortcuts();
+							nppParam->setShortcutDirty();
+							break;
+						}
```
dail8859: There seems to be a good bit of code duplication among these `case` statements (and in other parts of the file as well). I'm not sure how easy it would be to pull out this common functionality into a separate function(s).

PowerEditor/src/WinControls/shortcut/shortcut.h
```diff
@@ -174,6 +174,14 @@ class Shortcut  : public StaticDialog {
 
 	void setName(const TCHAR * name);
 
+	void clear(){
+		_keyCombo._isCtrl = false;
+		_keyCombo._isAlt = false;
+		_keyCombo._isShift = false;
+		_keyCombo._key = 0;
+		return;
+	}
```
dail8859: It is probably worth changing the default constructor to call this method instead of duplicating the code.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -352,6 +357,120 @@ INT_PTR CALLBACK ShortcutMapper::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
+							isModified = true;
+
+							//Notify current Accelerator class to update everything
+							nppParam->getAccelerator()->updateShortcuts();
+							unsigned long cmdID = pcsc.getID();
+							ShortcutKey shortcut;
+							shortcut._isAlt = FALSE;
+							shortcut._isCtrl = FALSE;
+							shortcut._isShift = FALSE;
+							shortcut._key = '\0';
```
dail8859: Though technically not needed, this would be more readable as `shortcut.clear();`

cmeriaux:
@dail8859 thank you very much for your kind words.
I accept your comments, I'm trying to refactor.....

cmeriaux:
@dail8859 I did some refactors, but i'm not sur it worth submitting it because there's a lots of modifications. If you want to have a look, it's in my repository https://github.com/cmeriaux/notepad-plus-plus/commit/5efc20360137f8ee1fc75b3724d8919c1b34cef6?diff=split

dail8859:
Yeah that definitely is alot of modifications. I guess it will be up to Don which of these he prefers (or something between these two).

PR #2810, opened 2017-01-21
Update danish.xml to 7.3.1

scootergrisen:


AraHaan:
@donho, mind looking at my python keyword patch while you are at it as it would help for people who use Python 3.5 and newer that uses the pep mentioned in the issue as well.

PR #2813, opened 2017-01-21
Added PluginsAdmin Folder

thedx4:
Admin folder was missing from the include section producing errors in
the compile process


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f62bf83bcaea83331aa468ca7ec76dfff2f65d88

PR #2841, opened 2017-01-29
Update Ukrainian & Belarusian

a-polivanchuk:
Hi,

I'm updating Ukrainian translation with first update of Belarusian for the last 2 years...
+ new Crowdin setup file for GitHub integration.

Thanks!


donho:
@a-polivanchuk 
I don't understand why you make both Ukrainian and Belarusian content into English, then turn them back to both Ukrainian and Belarusian.

But one thing is sure that PR should always be done from the current file in repo. There's no exception.

PR #2849, opened 2017-01-30
Fixed localization installation issue [regression]

SinghRajenM:
**While breaking single nsi file into multiple nis (or nsh) files, a regression happened.**
User is not able to install desired language(s) on system while selecting under localization (in install options). However, he/she can install only one preferred locale (along with English) from first page, but cann't install multiple languages from install option.

Root Cause:
langs.nsh is included (in nppSetup.nsi) just before .onInit. Therefore, SectionGroup "Localization" localization is executed right after .onInit, while all the locale xmls are packed in function copyCommonFiles (in mainSectionFuncs.nsh) which comes later in the sequence.
```nsi
; Copy all the language files to the temp directory
	; than make them installed via option
	SetOutPath "$PLUGINSDIR\nppLocalization\"
	File ".\nativeLang\"
```
Resolution:
Break "langs.nsh" into two parts (langs.nsh and langs2.nsh). File "langs.nsh" include only the part which is being used to show the language to installation process and file "langs2.nsh" is being used to copy to selected languages.


MAPJe71:
AFAIK the language selection in the installer is _intended_ to install _one_ language only.
So this is not a fix for an issue but an implementation of a feature.

SinghRajenM:
@MAPJe71 , Nope!
If you have a look on V6.9.2 or any earlier version, it works fine.
There are two things.
1. Language used for installation
image
This will guide to install in preferred language and after installation N++ will use the selected language for all its UI.
2. User can install multiple languages and can change it anytime from preference setting
image

Moreover, the language count for point 1 and point are different. For example, Language Hindi is not available for 1 while it is available in 2. 
So in latest versions of N++, second part is missing. Let me know if you still any concern. Thanks!

PR #2857, opened 2017-01-31
Set _isNoSession to true always if not the first instance

### LINKED ISSUE #2856: _isNoSession should always be set to true for instances after the first, irrespective of _multiInstSetting
Currently (v7.3.1) in WinMain cmdLineParams._isNoSession is always set to true if this is not the first instance (`TheFirstOne`) **only** `if (nppGUI._multiInstSetting == multiInst)`.

This doesn't seem to make sense, until session.xml correctly supports multiple instances (by saving all of them separately) the "rememberLastSession" thing should always be disabled for instances after the first.

I remind that using multiple instances with auto-backup enabled can easily result in the loss of files (e.g. #819).

#819 could probably be closed once this is fixed, although it would be much better to solve it by really supporting multiple instances in session.xml.

---

By the way, it's not very clean to override from the code CmdLineParams members, it results in some confusion: they are not really the "Command line params" (the parameters passed on the command line) anymore.

Gabr-F:
Closes #2856

I made a pull request just to make it clear that it's trivial to fix.

I'm not really familiar with the code base though, or even with C++, and haven't tested it, so you'd better give it a check before accepting it.

I would still much rather have multiple instances in session.xml supported, of course, but in the meanwhile this will save some people.

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -330,10 +330,11 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int)
 	if (nppGUI._multiInstSetting == multiInst)
 	{
 		isMultiInst = true;
-		// Only the first launch remembers the session
-		if (!TheFirstOne)
-			cmdLineParams._isNoSession = true;
 	}
+
+	// Only the first launch remembers the session
+	if (!TheFirstOne)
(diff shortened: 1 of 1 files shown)


donho:
Please describe the bug (app's bad behaviour), and what you expect.
I cannot merge it if there's no bug (or bug is unclear).

donho:
No reply from pr author, pr closed.

PR #2881, opened 2017-02-08
Fixed file attribute managing in FileManager::saveBuffer

DankRank:
Original code was flawed:
 * On +S+H (hidden and system) files, the attributes are set 4 times, when there's only need for 2 calls.
 * When clearing +S bit, it is masked out from the _original_ attribute value, which means that if the +H was cleared previously, it is set back again. It doesn't seem like this was intended, so I assume it's a bug.
 * When restoring the attributes, there's a pointless OR operation.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -1043,8 +1043,7 @@ bool FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool isCopy, g
 
 	EventReset reset(writeEvent); // Will reset event in destructor.
 	Buffer* buffer = getBufferByID(id);
-	bool isHidden = false;
-	bool isSys = false;
+	bool isHiddenOrSys = false;
 	DWORD attrib = 0;
 
 	TCHAR fullpath[MAX_PATH];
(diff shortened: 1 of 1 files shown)


donho:
Could you point out the bug issue or provide a step-by-step instruction to reproduce the bug?

DankRank:
Reproduction steps:
1) create a file
2) run in cmd `attrib +S +H [filename]`
3) edit and save file in notepad++

This bug doesn't have any side effects per se, so there's no difference between expected behaviour and what actually happens. Below I have described in detail what is changed.

----
This PR changes the behaviour that happens when a file with both Hidden and System attributes is saved.

Normally the buffer is saved like this (when +S and +H are set):
* Read attributes
* Set attributes to +S -H
* Set attributes to -S +H (this is what looks like a bug: the H flag is set again. This also renders previous call useless. This doesn't seem to have any side effects though)
* Write to the file
* Set attributes to +S +H
* Set attributes to +S +H (useless call)

This PR changes this behaviour to:
* Read attributes
* Set attributes to -S -H
* Write to the file
* Set attributes to +S +H

When file only has Hidder _or_ System flag (or neither), this PR doesnt change any behaviour.

justanotheranonymoususer:
What if UnicodeConvertor.fopen fails? The encoding is not restored in this case.

PR #2888, opened 2017-02-09
Prevents flicker in the Folder as Workspace tree (issue #2203)

dangibson:
Fixes issue #2203 - Folder As Workspace blinks with high rate


donho:
Not sure if it's the only solution to use `while` and `sleep`. I will try to find a better solution.

dangibson:
The lines:
	int iTVS_EX_DOUBLEBUFFER = 0x0004;
	SendMessage(_hSelf, TVM_SETEXTENDEDSTYLE, iTVS_EX_DOUBLEBUFFER, iTVS_EX_DOUBLEBUFFER);
independently resolve the issue . The while is to process all pending notifications in one go, and the sleep is for when there are lots happening at once (like when a build is happening).

PR #2895, opened 2017-02-10
Show OS bit version as well along with OS name

SinghRajenM:
Sometimes it is useful to know the OS bit version as well to investigate the reported issue.

image


PowerEditor/src/Parameters.cpp
```diff
@@ -742,7 +742,7 @@ generic_string ThemeSwitcher::getThemeFromXmlFileName(const TCHAR *xmlFullPath)
 }
 
 
-winVer getWindowsVersion()
+winVer NppParameters::getWindowsVersion()
```
donho: could be const

PowerEditor/src/Parameters.h
```diff
@@ -1734,4 +1736,5 @@ class NppParameters final
 	void initScintillaKeys();	//these functions have to be called first before any modifications are loaded
 	int getCmdIdFromMenuEntryItemName(HMENU mainMenuHadle, generic_string menuEntryName, generic_string menuItemName); // return -1 if not found
 	int getPluginCmdIdFromMenuEntryItemName(HMENU pluginsMenu, generic_string pluginName, generic_string pluginCmdName); // return -1 if not found
+	winVer getWindowsVersion();
```
donho: could be const

PowerEditor/src/Parameters.cpp
SinghRajenM: No, it can't be as it sets value for member _platForm in switch statement below.

```C++
switch (si.wProcessorArchitecture)
	{
	case PROCESSOR_ARCHITECTURE_IA64:
		_platForm = PF_IA64;
		break;

PowerEditor/src/Parameters.h
SinghRajenM: No, it can't be as it sets value for member _platForm in switch statement below.

PR #2909, opened 2017-02-16
Open File didn't work when there were tabs in the line.

dangibson:
\t\tc:\path\to\file.txt didn't open due to tabs - now it does

c:\\\\path\\\\to\\\\file.txt didn't open due to \\\\ - now it does


donho:
fixed

PR #2920, opened 2017-02-17
iface docu correction of bufferID

chcg:
- corrections for interface of bufferID, as x64 requires usage of UINT_PTR aka WPARAM instead of INT
- used new EolType instead of formatType
- removed commented NPPM_ADDREBAR, NPPM_UPDATEREBAR, NPPM_REMOVEREBAR not available in n++ code
- corrected typos


chcg:
@donho What do you think about correction of:

		#define DOCSTAUS_READONLY 1
		#define DOCSTAUS_BUFFERDIRTY 2

which should be probably DOCSTATUS

donho:
@chcg 
> which should be probably DOCSTATUS
yes, indeed. Please do another commit in this PR for correcting them in the comment and in the code.

chcg:
Amended the commit with changes for DOCSTATUS

Issue #2713, opened 2016-12-22
link to UDL 2.1 documentation is outdated


pryrt:
![](http://i.imgur.com/WBdNGxY.png)
UDL 2.1.0.12 dialog box still links to the old "temporary" [http://udl20.weebly.com/](http://udl20.weebly.com/), which is a UDL 2.0 description.

The newer [http://ivan-radic.github.io/udl-documentation/delimiters/](http://ivan-radic.github.io/udl-documentation/delimiters/) more fully describes UDL 2.1 (including some of the changes between UDL 2.0 and UDL 2.1).  

Could the link in the dialog box be updated to the newer documentation site; or, if there's a documentation site even newer than that, could the dialog box link be set to whatever most fully covers 2.1.0.12?

Thanks

chcg:
@ivan-radic
Is this the official docu side for UDL2.1?

At https://sourceforge.net/p/notepad-plus/discussion/331753/thread/01f8c28f/?page=26
there is some dicussion on udl3, any news on this task?

The link needs also to be updated here:
http://docs.notepad-plus-plus.org/index.php/User_Defined_Languages

ivan-radic:
Hi,

new site (http://ivan-radic.github.io/udl-documentation/delimiters/) was my attempt to learn how github pages work. I never published it because I wanted to integrate it into official Notepad++ documentation site, not keep it as separate page/repository. I never did it due to lack of time and some health issues that affected my productivity.

Since I recently resumed my work on UDL, I am OK with putting new link into into UDL GUI. Check with Don Ho if that's OK with him too.

Here are relevant source files (search for 'weebly' string):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScitillaComponent/UserDefineDialog.cpp
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScitillaComponent/UserDefineDialog.rc

Best regards
Ivan Radic

ivan-radic:
Regarding UDL3. This is an ongoing task (recently resumed after longer break caused by burnout and health issues), that is kept in unpublished private repository.
There won't be any news or updates about it until I finish it.

Best regards
Ivan Radic

chcg:
Thanks Ivan for the update on this topic. Do you wanna provide a PR and checkback with Don Ho?

Best wishes,
 Christian

donho:
[...]

PR #2943, opened 2017-02-22
link to udl v2.1 documentation

chcg:
- update docu from udl v2.0 to v2.1, see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2713
- additional info:
The link needs also to be updated here:
http://docs.notepad-plus-plus.org/index.php/User_Defined_Languages

Open:
Is bounding box 40,31,160,13 ok or are there some graphical design rules it should follow apart from that the longer text has to fit


donho:
The current version of UDL is v2.0 in Notepad++.
I'll check with Loreia (the author of UDL v2) for an eventual update (or not).

chcg:
See 

[BUG_FIXED] (Author: Ivan Radić - aka Loreia) UDL 2.1.
donho committed on 27 Jan 2013
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d60466eed348e1d61577225fd260abeb5d061095#diff-ddbb1ae1c7d085bd1c35d143255d0f31

Current version from https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/include/SciLexer.h:

#define SCE_UDL_VERSION_MAJOR       2
#define SCE_UDL_VERSION_MINOR       1
#define SCE_UDL_VERSION_BUILD       0
#define SCE_UDL_VERSION_REVISION 12

Issue #2948, opened 2017-02-23
Folding style is not visible due to used colors


SinghRajenM:
### Description of the Issue

When folding style is not visible if user changes the folder margin style. However, on next restart it is visible.

### Steps to Reproduce the Issue

1. Open Notepad++                         
2. Change "Folder Margin Style" either as "Box Tree" or "Circle Tree" (Settings->Preferences->Editing-> check Box Tree under "Folder Margin Style")  
3. Select any folder element.

### Expected Behavior

Selected element folding style should be visible (ideally in red color. it may vary based on current theme if any. But should be clearly visible.)
### Actual Behavior

Selected element folding style is shown in while color which is hardly visible.

### Debug Information

Notepad++ v7.3.2   (32-bit)
Build time : Feb 12 2017 - 23:15:39
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck2.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

![image](https://cloud.githubusercontent.com/assets/14791461/23276140/3ba48f60-fa2f-11e6-9e37-23b65504af76.png)

PR #2949, opened 2017-02-23
Make folding style visible

SinghRajenM:
Fixed issue #2948 
Instead of gray and while colors, ask styler to provide color. If styler does not provide then only use gray and while. 
To make review easier => same piece of code is used in method 
```C++
void ScintillaEditView::performGlobalStyles()
```
Also marked two methods as deleted which were not used (rather I'd say not defined).

**Before**
foldingstylebefore

**After**
foldingstyleafter


chcg:
So as it is code duplication accordint to your commen a new method makes sense

SinghRajenM:
@chcg Actually these colors are being extracted using methods only. But yes, as this piece of code (include two methods) is repeated, it can be moved into single method as well. and I have done it now.

Thank you!

Issue #2010, opened 2016-06-24
Wrong mouse cursor over title bar, menu and document tabs


ygoe:
Every once in a while, I haven't found the trigger yet, the mouse cursor over the window title bar, the menu bar and the document tabs, so everything above the document content, is changed from the default arrow to the text input cursor. When moving the mouse cursor, the arrow flashes up shortly but is immediately replaced with the text mouse cursor as the mouse stops moving. Restarting Notepad++ resolves the issue. This happens when Notepad++ has been running for a day or longer and used all day (switching in and out, and between documents, opening and closing documents…)

Observed with version 6.9.1 but probably already before that.

ygoe:
Happened again now.

ygoe:
And again, just a few minutes after. Something's wrong here.

Ede123:
Sounds a lot like what I discussed with @donho at f2cd7790669ec3867977293f5c83bdec235b81b5 and #1018.
Could you check if this is the case?

ygoe:
Doesn't sound at all like the references you mentioned. I've had the issue a couple of times since my last report, most recently just now. It's still bugging me and I have to restart NPP to resolve it. Simply switching windows, either with Alt+Tab or by mouse interaction, is enough to run into this, it just takes a few days normal usage.

MAPJe71:
Debug Info?

ygoe:
@MAPJe71 What does that comment mean?

MAPJe71:
In Notepad++ open `?` menu, select `Debug Info`, select `Copy debug info into clipboard`.
Post clipboard contents as a comment below :)

ygoe:
> Notepad++ v6.9.2
> Build time : May 18 2016 - 00:34:05
> Path : C:\Program Files (x86)\Notepad++\notepad++.exe
> Admin mode : OFF
> Local Conf mode : OFF
> OS : Windows 10
> Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppSaveAsAdmin.dll PluginManager.dll ScrollPastEOFUni.dll

SinghRajenM:
I found the steps to reproduce the issue.

1. Open a file in (say a.txt) N++
2. Open the same file in windows Notepad. Now update and save it in Notepad.
3. Come to N++, it will ask you to reload the file.
4. As soon you click "Yes", we can see the reported behavior.

[...]

Issue #1761, opened 2016-04-15
Cursor flicker


msorscher:
I've installed the most recent upgrade to Notepad++: v6.9

Now whenever my cursor moves into the Notepad++ frame, it loses its arrow image, and takes on the edit shape (kind of like a capital I with serifs), or the double-headed arrow resize icon.  It flickers and doesn't keep it's shape.

rillig-nm:
Does it happen always? I experience the same behavior, but only after this sequence:

1. Open a file in Notepad++
2. Edit the file, but don't save it
3. Change the file outside of Notepad++
4. Switch back to Notepad++
5. In the Reload question dialog, answer Yes

When moving the mouse cursor now, the cursor flickers. This is especially visible in the title bar, where the cursor flickers between the insert text beam and the arrow.

Notepad++ v7   (64-bit)
Build time : Sep 21 2016 - 22:07:28
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll

SinghRajenM:
Possible duplicate of #2010 , #1018,
Fix has been submitted #2952

Issue #1018, opened 2015-10-17
Prevent automatic reload of files before shutdown


Ede123:
**Steps to reproduce:** ("File Status Auto-Detection" has to be activated, "Update silently" has to be deactivated, OS was Window 7 in the tests but I don't think it matters):
1 - Open a file in Notepad++
2 - Defocus Notepad++ and edit the file in another application
3 - Now try to close Notepad++ by clicking the close icon

**Expected behavior:**
Notepad++ closes immediately.

**Actual behavior:**
The "Reload" dialog opens ("The File has been modified by another program. Do you want to reload it").
However it does not matter what choice the user makes: Even when choosing to reload the file, Notepad++ is closed immediately afterwards. 

 &nbsp;

I had a quick look at the code but it seems the issue is harder to fix than I hoped:
- The reload (`Notepad_plus::notifyBufferChanged`) is triggered as soon as focus is restored to the Notepad++ window (which is when the mouse button is pressed over the close button)
- The `WM_CLOSE` message is however only triggered after the mouse button has been released.

Does somebody with more experience have an idea on how to fix this?

dail8859:
:+1: I agree this should be the default behavior.

Ede123:
In a first shot to fix this issue I tried to delay reloading by posting a `NPPM_INTERNAL_CHECKDOCSTATUS` message instead of directly reloading after receiving the `WM_ACTIVATEAPP` message (which immediately triggers the reload in the current code). See following diff:

```diff
diff --git a/PowerEditor/src/NppBigSwitch.cpp b/PowerEditor/src/NppBigSwitch.cpp
index bd6d522..23f850f 100644
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1335,13 +1335,20 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPa
 		{
 			if (wParam == TRUE) // if npp is about to be activated
 			{
-				const NppGUI & nppgui = pNppParam->getNppGUI();
-				if (LOWORD(wParam) && (nppgui._fileAutoDetection != cdDisabled))
+				if (Message == WM_ACTIVATEAPP)
 				{
-					_activeAppInf._isActivated = true;
-					checkModifiedDocument();
[...]

PR #2952, opened 2017-02-24
Fix: flickering issue if Doc is modified outside N++

SinghRajenM:
Fixes issue: #2010 , #1018, #1761 
@Ede123 Thank for giving idea about [fix](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2010#issuecomment-282321687).

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5001,6 +5001,18 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
                     // Then we ask user to update
 					didDialog = true;
 					
+					// This section deal with the flicking issue which was caused by f2cd779 (discussed in detail @ #1018)
+					// Idea is :
+					//		1. First get the current caret position.
+					//		2. Send WM_LBUTTONDOWN and WM_LBUTTONUP message (which acts as double click at 0,0)
+					//		3. Now set caret back to original position.
+					//
+					auto curPos = _pEditView->execute(SCI_GETCURRENTPOS);
+					::PostMessage(_pEditView->getHSelf(), WM_LBUTTONDOWN, 0, 0);
+					::PostMessage(_pEditView->getHSelf(), WM_LBUTTONUP, 0, 0);
+					::PostMessage(_pEditView->getHSelf(), SCI_SETSEL, curPos, curPos);
+					// Now the flickering issue is not seen.
+
 					if (doReloadOrNot(buffer->getFullPathName(), buffer->isDirty()) != IDYES)
 						break;	//abort
 				}


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -5001,6 +5001,18 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
...
+					// Idea is :
+					//		1. First get the current caret position.
+					//		2. Send WM_LBUTTONDOWN and WM_LBUTTONUP message (which acts as double click at 0,0)
+					//		3. Now set caret back to original position.
+					//
+					auto curPos = _pEditView->execute(SCI_GETCURRENTPOS);
+					::PostMessage(_pEditView->getHSelf(), WM_LBUTTONDOWN, 0, 0);
+					::PostMessage(_pEditView->getHSelf(), WM_LBUTTONUP, 0, 0);
+					::PostMessage(_pEditView->getHSelf(), SCI_SETSEL, curPos, curPos);
+					// Now the flickering issue is not seen.
```
dail8859: Would a single call to `_pEditView->execute(SCI_GRABFOCUS);` work?

SinghRajenM: Sorry, it does not work. :(

SinghRajenM: I mean _pEditView->execute(SCI_GRABFOCUS); does not work

dail8859: Good to know. It was just a thought.

Ede123:
You should probably check what messages are sent as a result of clicking the text area.
I guess sending one single message could already be enough to sort this out (or maybe only changing order of messages somewhere, as I assume this is everything the conflicting commit did)

SinghRajenM:
Sending WM_LBUTTONDOWN itself is enough to handle the flickering, but then caret behavior is unpredictable.
Sometimes, caret is set at top while sometime it is set somewhere else. 

Therefore, send WM_LBUTTONDOWN (and obviously WM_LBUTTONDOWN should be released so send WM_LBUTTONUP, other scintilla will consider that user has pressed down, so start selecting text based in mouse move).
Now we don't know where is the caret, so we must set it to original position. Similar thing is done few lines down.

``` C++
if (didDialog)
		{
			auto curPos = _pEditView->execute(SCI_GETCURRENTPOS);
			::PostMessage(_pEditView->getHSelf(), WM_LBUTTONUP, 0, 0);
			::PostMessage(_pEditView->getHSelf(), SCI_SETSEL, curPos, curPos);
			if (::IsIconic(_pPublicInterface->getHSelf()))
				::ShowWindow(_pPublicInterface->getHSelf(), SW_RESTORE);

			if (doCloseDoc) // buffer has been deleted, cannot (and no need to) go on
				return;
		}
```

Ede123:
I just created #2956 which contains and improves upon this fix.
I assume you won't mind if Don commited that one instead @SinghRajenM ?

Ede123:
@donho Have you seen my comment? Should I rebase #2956?

PR #2956, opened 2017-02-24
Improve behavior when notifying the user about a buffer change

Ede123:
This PR is an improved version of #2952:
- It also fixes the flickering issue described in #1018, #1761 and #2010 which was exposed by f2cd7790669ec3867977293f5c83bdec235b81b5 while avoiding code duplication
- On top of that it improves functionality: Now the Notepad++ window is restored and the document in question is activated *before* showing the dialog asking whether to reload/close the file. This was only done afterwards before forcing the user to choose blind.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -4963,6 +4963,27 @@ void Notepad_plus::drawTabbarColoursFromStylerArray()
 		TabBarPlus::setColour(stInact->_bgColor, TabBarPlus::inactiveBg);
 }
 
+void Notepad_plus::prepareBufferChangedDialog(Buffer * buffer)
+{
+	// immediately show window if it was minimized before
+	if (::IsIconic(_pPublicInterface->getHSelf()))
+		::ShowWindow(_pPublicInterface->getHSelf(), SW_RESTORE);
+
+	// switch to the file that changed
+	int index = _pDocTab->getIndexByBuffer(buffer->getID());
+	int iView = currentView();
+	if (index == -1)
+		iView = otherView();
+	activateBuffer(buffer->getID(), iView);	//activate the buffer in the first view possible
+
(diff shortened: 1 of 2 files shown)


SinghRajenM:
@donho 
This PR is really improved version of #2952. It handles additional case of flickering in case doc is deleted which was not taken care in previous PR. Also good code reusability by avoiding code duplication.

Overall a good PR except few code reformatting which visual studio(2015) does automatically many times within the scope. For example, as soon ending parenthesis (for if, for, switch or even function) is put, VS will reformat the code within the parenthesis. 

PS: Personally I like this PR and want to see it merged.

donho:
@SinghRajenM  
> It handles additional case of flickering in case doc is deleted which was not taken care in previous PR. 

I cannot reproduce it. Could you provide the instruction?

@Ede123 Could you rebase this PR on master so I can review it please?

SinghRajenM:
Steps to reproduce the flickering issue when doc is deleted.

1. Open a file in (say a.txt) N++
2. Now delete the file from disk. (If you're checking on Win10, then F5 is needed. Because sometimes Win10 does not refresh itself).
3. Come to N++, it will ask you to keep/remove the file in/from editor.
4. Move mouse pointer on Yes/No button or on any other file menu item. You can observe flickering and changing mouse pointer icon.

Ede123:
Rebase done.

@donho Those two issues are really the same and are caused by a confirmation dialog being shown.

Before this was handled by setting `didDialog` to true and then doing some workarounds afterwards, see [here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2956/commits/69b1b2502039e428a16a7c14a338aaaf83ae3b77#diff-53dd9a88b2bff8faa9e32059b865f2bfL5078). It's the same code SinghRajenM's and my PR are using with the difference of an additional `WM_LBUTTONDOWN` (before only the `WM_LBUTTONUP` message was sent). With my PR those things are always done immediately before showing the dialog.

PR #2958, opened 2017-02-25
Handle both File and Folder Drop

### LINKED ISSUE #2037: Drag and drop both files and folders doesn't open anything
A few days ago I noticed Notepad++ fails to open a combination of folders + files. During a train ride I devoted some time to fixing this issue, and I have a small pull request ready to go. The behaviour in the pull request now opens the folders in the directory viewer and the immediate directory files in editor windows.

SinghRajenM:
Fixes #2037

If both files and folder are drop together then also handle it.

Scenario 1: Open all files of folder instead of launching Folder as Workspace on folder dropping is checked
			Case 1 : Only File(s) drop = > Open file(s)
			Case 2: Only Folder(s) drop = > Open all the files from folder(s)
			Case 3: Folder(s) + File(s) drop = > Open all the files from the folder along with other selected file(s).

Scenario 2: Open all files of folder instead of launching Folder as Workspace on folder dropping is unchecked
			Case 1 : Only File(s) drop = > Open file(s)
			Case 2: Only Folder(s) drop = > Launch Folder as workspace and show all the files in it.
			Case 3: Folder(s) + File(s) drop = > List all the files from the Folder(s) in the Folder as Workspace and open all other selected file(s). #

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3127,77 +3127,68 @@ void Notepad_plus::dropFiles(HDROP hdrop)
 		else
 			switchEditViewTo(MAIN_VIEW);
 
+		/*
+		Scenarios 1: Open all files of folder instead of launching Folder as Workspace on folder dropping is checked
+		Case 1 : Only File(s) drop = > Open file(s)
+		Case 2: Only Folder(s) drop = > Open all the files from folder(s)
+		Case 3: Folder(s) + File(s) drop = > Open all the files from the folder along with other selected file(s).
+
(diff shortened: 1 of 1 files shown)


SinghRajenM:
Adding to above:
In current setup, all the cases **except case3 from scenario2** are handled. The purpose of this PR is used handle case3 from scenario2.

donho:
Current behaviour should be kept.
I will add an error message to notice users to disable Folder as Workspace dropping mode in order to enable the old feature.

SinghRajenM:
Deleting the branch as it will not be merged.

Issue #2828, opened 2017-01-26
Crashes if you hold right mousebutton and scroll a couple of times


Dunderbullen:
If you have a couple of tabs open and hold right mouse button and scroll a couple of times, NPP crashes.
This happens on Windows 10 at least.
It also kind of messes up the mouse cursor, and you need to move the mouse around a bunch before you can kind of control it again. 

### Steps to Reproduce the Issue

1. Hold right mouse button
2. Scroll
3. Profit

### Debug Information
Notepad++ v7.2.2   (32-bit)
Build time : Nov 27 2016 - 16:31:07
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll

SinghRajenM:
Duplicate of #2603

SinghRajenM:
Almost similar to #2603

Issue #2603, opened 2016-11-24
Disable right click + mouse wheel document switching


Grinderofl:
### Description of the Issue
Notepad++ features a facility where, when multiple tabs are opened, one can press and hold right mouse button, and use the mouse wheel to scroll through them. This opens a small popup window with the file names. I'm unable to speak for others, but I have never been able to get rid of that little popup without selecting a different document than my original one. This is a highly irritating feature and I would like to request a way to disable it completely, including the right click-scroll change tab behaviour that still persists even if you disable document switcher. It's far too easy for me to press this combination resulting in an unneeded document change. 

In addition, when it's the window popping up, it will stick even if you let go of the right mouse button. The whole reason I'm raising this bug request is that today my cup with this bug finally overfilled when I couldn't get rid of the popup window even after spamclicking it for five minutes straight trying different open documents to switch to. As this box blocks every other NPP functionality, including save, I lost a chunk of work as I had to kill the process. I cannot believe that I'd be the only user to whom this has been consistently happening ever since I started using it 6 years ago to the point of annoyance. 

### Steps to Reproduce the Issue
1. Open notepad++
2. Open more than one tab
3. Press and hold right mouse button in Notepad++ window and scroll the wheel
4. Let go of the right mouse button

### Expected Behavior
Document switcher closes, or does not open in the first place. My active tab does not change.

### Actual Behavior
Document switcher stays open. To get rid of it, I need to click on an entry. Same entry won't do it, has to be different, thus causing my tabs to switch.

### Debug Information
Windows, all the way from XP to 10, on all types of machines, vendor built, home built, macbook air+parallels/boot camp, imac 27"+parallels/boot camp, basically it doesn't matter where or what hardware or software, it just happens everywhere, because it's a built in feature. If you are unable to reproduce, I'm willing to record a video of this happening on a fairly consistent basis.

SinghRajenM:
Additionally, notepad crashed in below case always.
Steps to reproduce

1. Open notepad++
2. Open more than one tab (two tabs are enough)
3. Press and hold right mouse button in Notepad++ window and keep scrolling the wheel

At step 3, shadow of doc switcher start increasing. After certain amount of shadow, npp crashes.

![npp_crash](https://cloud.githubusercontent.com/assets/14791461/20604876/a9ea5830-b28e-11e6-8433-4a07a5537912.png)

SinghRajenM:
Additionally, notepad crashes in below case always.
[...]

Issue #1619, opened 2016-03-11
Application crash when using right-click + scroll too quickly


ArmlessMatt:
When there is more than on tab open in the app, Holding right-click and scolling up or down to quickly crash the entire app.

SinghRajenM:
Another duplicates #2603  and #2828

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2982

Issue #607, opened 2015-07-30
CTRL + mouse wheel menu not working on Windows 10


tzachyrm:
It looks like on every mouse wheel step a new menu is opened, and after too many of them, the process crashes.

SinghRajenM:
Another duplicates #1619 and #2828, 2603

PR #2971, opened 2017-02-28
Fix for issue #2828 Crashes if you hold right mousebutton and scroll a couple of times

richardebrock:
The application crashes when the user clicks the right mouse button and then scrolls back and forth between the list of files. The problem is caused by additional mouse wheel messages landing in the message queue before the dialog window has been completely initialized and is able to handle messages. As a result the main application window creates multiple TaskListDlg's which eventually leads to a stack overflow. The fix uses a static member variable in TaskListDlg to track the number of instances. This allows the main application window to check the static variable and only create the dialog window if there are no instances. This fix should then also address issues #1619, #2603 and #607 however this has to be confirmed.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2812,10 +2812,13 @@ void Notepad_plus::command(int id)
 				}
 				else
 				{
-					TaskListDlg tld;
-					HIMAGELIST hImgLst = _docTabIconList.getHandle();
-					tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
-					tld.doDialog();
+					if (TaskListDlg::_instanceCount == 0)
+					{
+						TaskListDlg tld;
+						HIMAGELIST hImgLst = _docTabIconList.getHandle();
+						tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
+						tld.doDialog();
+					}
 				}
 			}
 			_linkTriggered = true;
--- a/PowerEditor/src/WinControls/TaskList/TaskListDlg.cpp
+++ b/PowerEditor/src/WinControls/TaskList/TaskListDlg.cpp
@@ -31,6 +31,8 @@
 #include "Parameters.h"
 #include "resource.h"
 
+int TaskListDlg::_instanceCount = 0;
+
 LRESULT CALLBACK hookProc(UINT nCode, WPARAM wParam, LPARAM lParam)
 {
 	if ((nCode >= 0) && (wParam == WM_RBUTTONUP))
@@ -93,6 +95,7 @@ INT_PTR CALLBACK TaskListDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lP
 		{
 			_taskList.destroy();
 			::UnhookWindowsHookEx(_hHooker);
+			_instanceCount--;
 			return TRUE;
 		}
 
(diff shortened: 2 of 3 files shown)


SinghRajenM:
>The problem is caused by additional mouse wheel messages landing in the message queue
 
Correct.
>before the dialog window has been completely initialized and is able to handle messages. 

No, this is not true. To check it, just scroll once and wait for sometime. Now move mouse outside of the TasklistDlg area and start scrolling.
>As a result the main application window creates multiple TaskListDlg's which eventually leads to a stack overflow. 

True, multiple windows are created. But not because dialog was not completely initialized. Rather it is because of below code in "nppCommands.cpp" which is proceed for each mouse scroll.
```C++
	TaskListDlg tld;
	HIMAGELIST hImgLst = _docTabIconList.getHandle();
	tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
	tld.doDialog();
```

>The fix uses a static member variable in TaskListDlg to track the number of instances. This allows the main application window to check the static variable and only create the dialog window if there are no instances.

### _It creates regression. Now, you're restricted to scroll only over Tasklist dialog._

**IMO, WM_MOUSEWHEEL message in "nppBigSwitch.cpp" should be blocked until we find solution to handle it.**
```C++
case WM_MOUSEWHEEL:
{
	if (0 != (LOWORD(wParam) & MK_RBUTTON))
	{
		// redirect to the IDC_PREV_DOC or IDC_NEXT_DOC so that we have the unified process
		pNppParam->_isTaskListRBUTTONUP_Active = true;
		short zDelta = (short) HIWORD(wParam);
		return ::SendMessage(hwnd, WM_COMMAND, zDelta>0?IDC_PREV_DOC:IDC_NEXT_DOC, 0);
	}
	return TRUE;
}
'''
P.S. I've tried many options, but I could not find good solution to deal with it.

**Comment by SinghRajenM** [comment]
>The problem is caused by additional mouse wheel messages landing in the message queue
 
Correct.
>before the dialog window has been completely initialized and is able to handle messages. 

No, this is not true. To check it, just scroll once and wait for sometime. Now move mouse outside of the TasklistDlg area and start scrolling.
>As a result the main application window creates multiple TaskListDlg's which eventually leads to a stack overflow. 

True, multiple windows are created. But not because dialog was not completely initialized. Rather it is because of below code in "nppCommands.cpp" which is proceed for each mouse scroll.
```C++
	TaskListDlg tld;
	HIMAGELIST hImgLst = _docTabIconList.getHandle();
	tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
	tld.doDialog();
```

>The fix uses a static member variable in TaskListDlg to track the number of instances. This allows the main application window to check the static variable and only create the dialog window if there are no instances.

### _It creates regression. Now, you're restricted to scroll only over Tasklist dialog._

**IMO, WM_MOUSEWHEEL message in "nppBigSwitch.cpp" should be blocked until we find solution to handle it.**
```C++
case WM_MOUSEWHEEL:
{
	if (0 != (LOWORD(wParam) & MK_RBUTTON))
	{
		// redirect to the IDC_PREV_DOC or IDC_NEXT_DOC so that we have the unified process
		pNppParam->_isTaskListRBUTTONUP_Active = true;
		short zDelta = (short) HIWORD(wParam);
		return ::SendMessage(hwnd, WM_COMMAND, zDelta>0?IDC_PREV_DOC:IDC_NEXT_DOC, 0);
	}
	return TRUE;
}
```
P.S. I've tried many options, but I could not find good solution to deal with it.

richardebrock:
You're right about the initialization delay not being the cause, Ill close this pr and continue investigating.

PR #2975, opened 2017-02-28
Updated code for patching issue #2828

richardebrock:
As SinghRajenM correctly pointed out, my initial prognosis was incorrect and the issue is not caused by a timing issue between dialog initialization and the parent window. I stand by the _instanceCount member variable as a means to prevent subsequent dialog create calls, but this patch contains additional code to ensure that WM_MOUSEHWEEL messages are handled in the mouse hook and routed to the tasklist - this prevents the regressed user experience.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2812,10 +2812,13 @@ void Notepad_plus::command(int id)
 				}
 				else
 				{
-					TaskListDlg tld;
-					HIMAGELIST hImgLst = _docTabIconList.getHandle();
-					tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
-					tld.doDialog();
+					if (TaskListDlg::_instanceCount == 0)
+					{
+						TaskListDlg tld;
+						HIMAGELIST hImgLst = _docTabIconList.getHandle();
+						tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
+						tld.doDialog();
+					}
 				}
 			}
 			_linkTriggered = true;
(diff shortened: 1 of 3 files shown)


donho:
@SinghRajenM Could you review this PR please?

PowerEditor/src/NppCommands.cpp
```diff
@@ -2812,10 +2812,13 @@ void Notepad_plus::command(int id)
...
 				{
-					TaskListDlg tld;
-					HIMAGELIST hImgLst = _docTabIconList.getHandle();
-					tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
-					tld.doDialog();
+					if (TaskListDlg::_instanceCount == 0)
+					{
+						TaskListDlg tld;
+						HIMAGELIST hImgLst = _docTabIconList.getHandle();
+						tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
```
SinghRajenM: once dialog is created, then wheel scroll direction is eaten. It mean you can't change the tab switching in reverse way.

SinghRajenM:
Also some amount of testing is needed.
1. If Mouse is over Tasklist dialog, then both scrolling down works in reverse way. Scrolling up is not working at all.
2. When mouse is being scrolled outside Tasklist dialog, then scrolling up/down works in same direction.

richardebrock:
There are two problems in the code:

1) I'm neglecting to send the mouse data with the mouse wheel message and the delta data is lost, resulting in the mouse wheel handler always getting the same direction.

2) With point 1 fixed, there is still the problem that if the mouse is inside the tasklist, the list control is getting the WM_MOUSEWHEEL message from the UI subsystem and then again from the parent, this causes the line selector to move twice. I've implemented a workaround and now it's solved but I'm not sure the workaround is the most elegant way to go.  I'll post another pr once I've got that sorted.

PowerEditor/src/NppCommands.cpp
richardebrock: just to clarify: you are referring to the direction of the list scroll in this comment yes?

SinghRajenM: Correct. The flag direction is set once only. Afterwards, it is not handled whether user scrolls down or up.

PR #2980, opened 2017-03-01
fixing the mouse wheel to task list scroll bug (#2828)

richardebrock:
Sorts out the following:

1. Crash caused by scroll wheel + RMB selecting document in task list dialog
2. Direction of the scrolling in response to the mouse wheel that was present in previous patch attempts
3. Ensures that the item selected in the task list using either the mouse wheel or Ctrl + TAB is visible. This appears to be an unreported bug where a large number of files causes the task list to overflow the display height and hidden items are never shown even when selected.  

@SinghRajenM the dialog is always destroyed when the user selects a file in the task list and re-created, so I'm not sure if I'm misunderstanding what you were saying but I can't see any issue with direction being only set once in this latest patch.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2805,17 +2805,19 @@ void Notepad_plus::command(int id)
 			if (nbDoc > 1)
 			{
 				bool direction = (id == IDC_NEXT_DOC)?dirDown:dirUp;
-
 				if (!doTaskList)
 				{
 					activateNextDoc(direction);
 				}
 				else
 				{
-					TaskListDlg tld;
-					HIMAGELIST hImgLst = _docTabIconList.getHandle();
-					tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
-					tld.doDialog();
+					if (TaskListDlg::_instanceCount == 0)
+					{
+						TaskListDlg tld;
+						HIMAGELIST hImgLst = _docTabIconList.getHandle();
+						tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
+						tld.doDialog();
+					}
 				}
 			}
 			_linkTriggered = true;
--- a/PowerEditor/src/WinControls/TaskList/TaskList.cpp
+++ b/PowerEditor/src/WinControls/TaskList/TaskList.cpp
@@ -88,6 +88,7 @@ void TaskList::init(HINSTANCE hInst, HWND parent, HIMAGELIST hImaLst, int nbItem
 
 	ListView_SetItemState(_hSelf, _currentIndex, LVIS_SELECTED|LVIS_FOCUSED, LVIS_SELECTED|LVIS_FOCUSED);
 	ListView_SetBkColor(_hSelf, lightYellow);
+
 }
 
 void TaskList::destroy()
@@ -127,6 +128,11 @@ RECT TaskList::adjustSize()
 	ListView_SetColumnWidth(_hSelf, 0, _rc.right);
 	::SendMessage(_hSelf, WM_SETFONT, reinterpret_cast<WPARAM>(_hFont), 0);
 
+	//if the tasklist exceeds the height of the display, leave some space at the bottom
+	if (_rc.bottom > ::GetSystemMetrics(SM_CYSCREEN) - 120)
+	{
+		_rc.bottom = ::GetSystemMetrics(SM_CYSCREEN) - 120;
+	}
 	reSizeTo(_rc);
 
 	// Task List's border is 1px smaller than ::GetSystemMetrics(SM_CYFRAME) returns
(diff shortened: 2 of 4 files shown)


PowerEditor/src/NppCommands.cpp
```diff
@@ -2805,17 +2805,19 @@ void Notepad_plus::command(int id)
...
 				{
-					TaskListDlg tld;
-					HIMAGELIST hImgLst = _docTabIconList.getHandle();
-					tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
-					tld.doDialog();
+					if (TaskListDlg::_instanceCount == 0)
+					{
+						TaskListDlg tld;
+						HIMAGELIST hImgLst = _docTabIconList.getHandle();
+						tld.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), hImgLst, direction);
```
SinghRajenM: dialog is created once for per right click. What if user is scrolling down and suddenly changes mind and start scrolling up. Does it handle that case?

PowerEditor/src/WinControls/TaskList/TaskList.cpp
```diff
@@ -205,18 +211,19 @@ LRESULT TaskList::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lParam)
 			else
 			{
 				int32_t selected = (_currentIndex + 1) > (_nbItem - 1) ? 0 : (_currentIndex + 1);
-				ListView_SetItemState(_hSelf, _currentIndex, 0, LVIS_SELECTED|LVIS_FOCUSED);
+				ListView_SetItemState(_hSelf, _currentIndex, 0, LVIS_SELECTED|LVIS_FOCUSED);
```
SinghRajenM: Avoid reformatting to make code review easier and increase chances of PR acceptance.

PowerEditor/src/WinControls/TaskList/TaskList.cpp
```diff
@@ -205,18 +211,19 @@ LRESULT TaskList::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lParam)
...
-				ListView_SetItemState(_hSelf, _currentIndex, 0, LVIS_SELECTED|LVIS_FOCUSED);
+				ListView_SetItemState(_hSelf, _currentIndex, 0, LVIS_SELECTED|LVIS_FOCUSED);				
 				// tells what item(s) to be repainted
 				ListView_RedrawItems(_hSelf, _currentIndex, _currentIndex);
 				// repaint item(s)
 				UpdateWindow(_hSelf); 
 				ListView_SetItemState(_hSelf, selected, LVIS_SELECTED|LVIS_FOCUSED, LVIS_SELECTED|LVIS_FOCUSED);
 				// tells what item(s) to be repainted
-				ListView_RedrawItems(_hSelf, selected, selected);
+				ListView_RedrawItems(_hSelf, selected, selected);
```
SinghRajenM: Avoid reformatting to make code review easier and increase chances of PR acceptance.

PowerEditor/src/WinControls/TaskList/TaskList.cpp
richardebrock: Sorry about that, didn't realize it happened until after I had uploaded the pr

PowerEditor/src/NppCommands.cpp
richardebrock: The direction variable actually has no impact on the scrolling direction other than to set the initial selected item in the list, thereafter it is ignored. In this patch the scroll direction change is handled correctly and I didn't find any issue.

richardebrock: so short answer is yes

SinghRajenM: I tested thoroughly. So far it is working fine on win10. Did you check it on Win7?

I liked two more things (apart from the crash fix) ListView_EnsureVisible and List view height adjustment. 👍

richardebrock: I haven't tested on win 7 yet, Ill check win 7 and win 8.1 now and then post feedback here.

PowerEditor/src/WinControls/TaskList/TaskList.cpp
SinghRajenM: This is nice PR. So I would say, better you refresh the PR by removed reformatting and make sure it remains in single commit. You can use 
amend -a    to amend with previous  and
push -f    to force rewrite

donho: > This is nice PR. So I would say,

Agree.
:)

richardebrock: thanks for the tips. Testing on Win7 shows that if the mouse is outside the task list it jumps 2 items up or down instead of just 1.  So let me fix that up and Ill submit another pr, ill be sure to eliminate the formatting issues as well.

donho:
Thank you for reviewing @SinghRajenM  !

richardebrock:
I've found and fixed the win7 problem and confirmed that the updated code works on win8.1 and win10. Ill delete this PR then and submit a new one.

PR #2982, opened 2017-03-01
Fixing the mouse wheel to task list scroll bug (#2828), task list sizing and visible items

richardebrock:
This should do it now. Fixes the crash that occurs when a user holds down RMB and uses mouse wheel to scroll through document list. Also fixes the document list window size to prevent display overflow and ensures that the selected document item is visible. The patch includes an os version check and only forwards the mouse wheel messages on windows 10 to prevent duplicates getting through. I've tested on Win7, Win 7 embedded, Win 8.1 and Win 10.

### DIFF
--- a/PowerEditor/src/WinControls/TaskList/TaskListDlg.cpp
+++ b/PowerEditor/src/WinControls/TaskList/TaskListDlg.cpp
@@ -31,13 +31,25 @@
 #include "Parameters.h"
 #include "resource.h"
 
+int TaskListDlg::_instanceCount = 0;
+
 LRESULT CALLBACK hookProc(UINT nCode, WPARAM wParam, LPARAM lParam)
 {
 	if ((nCode >= 0) && (wParam == WM_RBUTTONUP))
     {
 		::PostMessage(hWndServer, WM_RBUTTONUP, 0, 0);
     }        
-	
+	if ((nCode >= 0) && (wParam == WM_MOUSEWHEEL))
+	{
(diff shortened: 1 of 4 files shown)


PowerEditor/src/WinControls/TaskList/TaskListDlg.cpp
```diff
@@ -38,16 +38,16 @@ LRESULT CALLBACK hookProc(UINT nCode, WPARAM wParam, LPARAM lParam)
 	if ((nCode >= 0) && (wParam == WM_RBUTTONUP))
     {
 		::PostMessage(hWndServer, WM_RBUTTONUP, 0, 0);
-    }        
-	if ((nCode >= 0) && (wParam == WM_MOUSEWHEEL))
+    } 
+	if ((nCode >= 0) && (wParam == WM_MOUSEWHEEL) && windowsVersion== WV_WIN10)
```
donho: Why Win10 only?

richardebrock: In my Win7 and Win8.1 tests the mouse wheel messages are sent to the control even if the mouse is outside the control whereas on Win10 they stop if you move outside the task list. So it appears to that Win10 handles things slightly differently, and that's what I put the check to only pass the messages on if its on Win10.

donho: Then it would be `windowsVersion >= WV_WIN10)`

Issue #2976, opened 2017-03-01
NPP Debug info: read version information from registry if available and use current version only as fallback


GitMensch:
Current debug info for the OS shows:
* `Windows 8.1` on on Windows Server 2012 R2 Standard 
* `Windows 10` on Windows 10 Pro

The OS info for the Server is obviously misleading, the OS info for Windows 10 is missing the important build number (and the concrete version).

**Suggestion**: Read this information from registry if possible.
The following key is available at least since Windows 7 (not checked older versions) and can be read with guest privileges, too: 

`HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion` in `ProductName`, `CurrentBuildNumber` and `UBR` (Update Build Revision). The last two values combined are the full "OS build" number mentioned in [MS Release list][1].

This provides:
* Windows Server 2012 R2 Standard / 9600 / 18543 on the server
* Windows 10 Pro / 14393 / 693 on the client, additional there is the `ReleaseId` (only Windows 10+) showing 1607

Using it for the debug info and using the current implementation only as a fall-back would lead to a string like `OS: Windows Server 2012 R2 Standard, Build 9600.18543` / `OS: Windows 10 (1607), Build 14393.693`.

[1]:https://technet.microsoft.com/en-us/windows/release-info

SinghRajenM:
Fix is submitted to show OS name properly now. However, it will show only OS name and bit info (32/64 bit).
As this info is meant for Notepad++ Dev Team to investigate the issue further. I think there is no point to show build number or flavor such Home/Professional/Enterprise Edition etc.

GitMensch:
Nice to have the OS name properly, the pull request makes the current implementation better.
Checking the current code there are some side effects of the current way of getting the OS which would be better with checking the registry:

* the current code is much more than the registry lookup (of a key that is officially documented and therefore unlikely to change and accessible for reading with nearly no rights)
* the current code doesn't apply to newer versions of windows - you'd always have to add code for this
* you don't get the flavor (this can be important as some features simply don't exist in minimal versions)
* for Windows 10+: you don't get the build number - after using Windows 10 including but not limited to the Insider builds I know that some strange behavior only exist in older builds (or in new Insider builds where some software just stops to work because MS changed something)

You may argue that NPP doesn't use these features but you don't know if this is true in the future and I wonder why it shouldn't show this information if it can get it reliable and more easy than the current implementation.

donho:
[...]

PR #2986, opened 2017-03-02
Show OS name properly on Debug Info

SinghRajenM:
Fix for #2976  There is no check to handle server name.

Fix is submitted to show OS name only. However, user reported issue saying show build number and OS flavor etc. As this info is meant for Notepad++ Dev Team to investigate the issue further. I think there is no point to show build number or flavor such Home/Professional/Enterprise Edition etc.

Let me know if those info is needed. I will be happy to update the PR.

PR is updated using [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724833(v=vs.85).aspx)

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -45,7 +45,7 @@ enum LangType {L_TEXT, L_PHP , L_C, L_CPP, L_CS, L_OBJC, L_JAVA, L_RC,\
 			   // The end of enumated language type, so it should be always at the end
 			   L_EXTERNAL};
 
-enum winVer{ WV_UNKNOWN, WV_WIN32S, WV_95, WV_98, WV_ME, WV_NT, WV_W2K, WV_XP, WV_S2003, WV_XPX64, WV_VISTA, WV_WIN7, WV_WIN8, WV_WIN81, WV_WIN10};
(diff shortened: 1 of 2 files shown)


donho:
winVer is used on several places for different usage.
It doesn't worth to take regression risk for Debug Info feature which is an accessory feature.

GitMensch:
I've checked the occurrences of winVer (using GitHub) and the fix is complete. The only parts where the fix was not correct is in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2986/commits/7c25cde1ee848911ca61863259b322259ea28b8f#diff-f898b2bd88be20761bd1b393779b810f as "several places" compare the enum with `>=` and therefore the order in the definition must use the "old" constants first. It must be 

```enum winVer{ WV_UNKNOWN, WV_WIN32S, WV_95, WV_98, WV_ME, WV_NT, WV_W2K, WV_XP, WV_S2003, WV_XPX64, WV_SHOME, WV_S2003R2, WV_VISTA, WV_S2008, WV_S2008R2, WV_WIN7, WV_WIN8, WV_S2012, WV_WIN81, WV_S2012R2, WV_WIN10, WV_S2016};```

SinghRajenM:
Deleting branch as PR will be merged.

GitMensch:
> Deleting branch as PR will be merged.

Did you mean "will not be merged"?

SinghRajenM:
oh, updated the comment. Yes, I mean "will not be merged."

PR #2990, opened 2017-03-04
Display MRU list properly (ctrl+TAB or Right Click+Scroll)

SinghRajenM:
If number of opened files are more, then MRU list exceed the length of the current display monitor and not visible which item is currently being selected. This PR fixed this issue and ensure that 

1. MRU list does not exceed current display height
2. Current selected item is visible

image


PowerEditor/src/WinControls/TaskList/TaskList.h
```diff
@@ -69,10 +70,13 @@ class TaskList : public Window
 		return (((TaskList *)(::GetWindowLongPtr(hwnd, GWLP_USERDATA)))->runProc(hwnd, Message, wParam, lParam));
 	};
 
+	void GetCurrentDisplayResolution();
```
MAPJe71: I'd expect a `get...Resolution` function to *return* the resolution i.e. both horizontal and vertical.

SinghRajenM: Okay, then let me change `get...Resolution` to `cal...Resolution`.

PR #2996, opened 2017-03-06
Display MRU list properly on seconday display

SinghRajenM:
If number of opened files are more and NPP is running on secondary display, then MRU list exceed the length of the current display monitor and not visible which item is currently being selected. This PR fixed this issue and ensure that

1. MRU list does not exceed current display height
2. Current selected item is visible

image


donho:
@SinghRajenM Please provide accurate way to reproduce it.
I cannot reproduce it at all.

SinghRajenM:
@donho , refer the steps to reproduce the issue as below - 

Case 1:
1. Connect two displays and make higher resolution display as primary display.
2. Open N++ and move it to secondary display which have comparatively lower resolution.
3. Open enough number of tabs (say 40~45).
4. Generate MRU list and see output. It will be same as displayed in previous comment.

Case 2: Refer the below sceeenshot.
image

SinghRajenM:
No response from long time. So closing the pull request.

Issue #2929, opened 2017-02-20
Pressing "Save all" by mistake on many unsaved documents causes multiple Save As dialogs


klanderfri:
### Description of the Issue
My workflow causes me to have 20+ unsaved documents open in Notepad++ that are never to be saved.
Sometimes I press "Save All" by mistake which causes multiple "Save As" dialogs. This forces me to select 
"Cancel" on a lot of dialogs.

### Steps to Reproduce the Issue
1. Create several documents and make sure they are changed and unsaved (Ctrl+N, then add some text into the document).
2. Ctrl+Shift+S
3. Press Cancel.
4. Press Cancel again.
5. And again...

### Expected Behavior
If the user presses Cancel then this choice should be applied to all subsequent dialogs.

### Actual Behavior
Every Save As are treated like an independent operation with no regard to previous user choices.

### Important
This issue is about unsaved documents that never been saved, ie only exists within the Notepad++ context.
Having several existing documents open, but unsaved, will NOT trigger the problem described in this issue.

### Debug Information
Notepad++ v7.3.2   (32-bit)
Build time : Feb 12 2017 - 23:15:39
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : HTMLTag_unicode.dll mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll

Carbone13:
same problem here

redmanmale:
See also #2724.
That dialog - Yes/No for all - seems like a more convenient solution for this, because your option - 'Cancel then this choice should be applied to all subsequent dialogs' is counter-intuitive imo.

salvarani:
"Cancel" button + "[X] Apply to all subsequent documents" checkbox
looks like the best solution to me.

donho:
This issue has been fixed by new added buttons "Yes to All" and "No to All".

klanderfri:
> This issue has been fixed by new added buttons "Yes to All" and "No to All".

In which version is this issue fixed? :-)

sasumner:
>In which version is this issue fixed?

the "...to all" additional buttons first appeared in v. 7.8[.0]

klanderfri:
[...]

PR #2999, opened 2017-03-06
Pressing "Save all" by mistake on many unsaved documents causes multiple Save As dialogs #2929

richardebrock:
Added a return inside the file save all loop to prevent multiple dialog windows


donho:
This pr change the current behaviour into an eventual unexpected behaviour.
The best behaviour to me is asking user if he/she want to cancel all after the first cancel.

PR #3019, opened 2017-03-11
Fix issue with clicking links in html/js/asp/php

dail8859:
The style at the position was reported has having all the higher bits set for some reason, thus when checking if it was a hotspot it was always false. For example the style at the position is reported as:

`4294967168 = 0b11111111111111111111111110000000`

However the styles are only 1 byte worth of data. I'm not sure why the higher bits are being set (possible due to using deprecated API calls) but this solves the immediate issue.


donho:
@dail8859  Have you the reported issue number or the instructions to reproduce the issue?

dail8859:
https://notepad-plus-plus.org/community/topic/13415/v7-3-3-fix-cia-hacking-notepad-issue/14

Issue #2998, opened 2017-03-06
Wrong very wrong translation


andre-paulo98:
So after like 30 minutes looking for Word Wrap in Portuguese Translation for Notepad++ i gave up and switched the language to English to finally be able to activate word wrap.
Could please the **Procura Contínua** be changed to **Moldar texto**
(not sure if i'm using the European or Brazilian Portuguese translation)

PR #3021, opened 2017-03-12
Update Portuguese translation as per issue #2998

govindragh:


donho:
@govindragh  Do you mind to provide more explanation for this change?

govindragh:
Sure, sorry! I noticed this issue (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2998), and verified that the translation in the issue is more accurate than the current one to the best of my abilities. I replaced all instances of "Procura Contínua" to "Moldar Texto" to match this.

Full disclaimer: I don't speak any Portuguese, and I was very bored when I pushed this change, so I won't feel hurt if you deny it. :)

PR #3024, opened 2017-03-12
Update estonian.xml

AndresTraks:


donho:
We need the ref of people (email) who know the language to maintain the translation. We don't need the ref of people who use the translator. Therefore this PR won't be accepted.

PR #3030, opened 2017-03-13
Misc and stuffs

milipili:


PowerEditor/src/winmain.cpp
```diff
@@ -37,6 +37,42 @@ namespace
 {
 
 
+void allowVmCopydataMessages(Notepad_plus_Window& notepad_plus_plus, const NppParameters* pNppParameters, winVer ver)
```
donho: Rather allow**W**mCopydataMessages ?

PowerEditor/src/MISC/Process/ProcessAvecThread/Process.h
```diff
@@ -29,6 +29,7 @@
 #pragma once
 
 #include <windows.h>
+#include "Common.h"
```
donho: Why not? But why?

donho: Why not? But why?

PowerEditor/src/winmain.cpp
```diff
@@ -161,7 +161,7 @@ void parseCommandLine(TCHAR * commandLine, ParamVector & paramVector) {
 	//the commandline generic_string is now a list of zero terminated strings concatenated, and the vector contains all the substrings
 }
 
-bool isInList(const TCHAR *token2Find, ParamVector & params)
+bool removeIfExists(const TCHAR *token2Find, ParamVector & params)
```
donho: The removal feature is rather for making next parsing easier.
I would say "isInList" is better sementically.

milipili: `isInList` does not do what it says. You expect something immutable, not something that will change the content of the input. In that regard, this function is badly named.

PowerEditor/src/winmain.cpp
milipili: Oh indeed.

PowerEditor/src/MISC/Process/ProcessAvecThread/Process.h
milipili: Please see the commit message:
```
process: add missing include Common.h (generic_string)
```
This include is required for the typedef `generic_string`.

milipili:
It seems that I can't mark the requested changes as done. So doing it here (and rebased from master)

donho:
@milipili Thank you. 
Have you run it with your modification?

Issue #3029, opened 2017-03-13
[Bug] Notepad++ problems with saving extensions for Session and Workspace files


Avaray:
Im using Win 8.1 64bit and Notepad++ 32bit.
I tested it with v 7.3.3 and v 6.9.2.

In 7.3.3 Im trying to save Session (File > Save Session).
Im naming file and choosing extension from list. And it is not saving that extension. I need to write extension manually.
And when Im trying to save Workspace there is no extension to choose at all.

In 6.9.2 it gives extension for Session file, but for Workspace file there is the same problem like in 7.3.3.

PR #3032, opened 2017-03-13
[Fixed regression]: Use extension for session file.

SinghRajenM:
Fixed regression for save session file #3029. This regression is caused in [6388d48](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6388d48e0ce5707e822f287203e46a2923f2881f). Reason was, by default selected index was -1, but if user toggles extension filters than issue does not occur.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1875,6 +1875,7 @@ const TCHAR * Notepad_plus::fileSaveSession(size_t nbFile, TCHAR ** fileNames)
 			sessionExt += TEXT(".");
 		sessionExt += ext;
 		fDlg.setExtFilter(TEXT("Session file"), sessionExt.c_str(), NULL);
+		fDlg.setExtIndex(0);		// 0 index for "custom extention types"
 	}
 	fDlg.setExtFilter(TEXT("All types"), TEXT(".*"), NULL);
 	sessionFileName = fDlg.doSaveDlg();


PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.h
```diff
@@ -26,16 +26,12 @@
...
+#pragma once
 
-//#include <windows.h>
-#ifndef DOCKINGDLGINTERFACE_H
 #include "DockingDlgInterface.h"
-#endif //DOCKINGDLGINTERFACE_H
-
 #include "TreeView.h"
 #include "ProjectPanel_rc.h"
+#include "FileDialog.h"
```
donho: Can use **class FileDialog;** instead.

SinghRajenM: Yes. It is done now.

donho:
@SinghRajenM This PR is for 2 issues, so please separate your commit in 2.

SinghRajenM:
@donho PR has been updated. This handles the regression. I will submit another one to handle workspace related issue #3029

donho:
@SinghRajenM Compiled under VS2013 release 32/64 bit. 
An extension "sss" has been set as session ext.
Save session with "sss" still doesn't work. No extension is added.

SinghRajenM:
But it work fine for me on my personal machine (win10 x64) with VS2015 (both release and debug). I have checked in my office dev PC win7 x64 using VS2013 for both release/debug (32/64 bit). I didn't see any issue.

donho:
Yes, it works your PR.
Sorry - I've got option "Use new style save dialog" enabled in Preferences dialog, that disabled the extension feature.

PR #3033, opened 2017-03-13
use scintilla namespace

chcg:
- activated usage of scintilla namespace in nmakefile
- corrected missing namespaces

-preparation for static build of scintilla lib, see discussion https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b869163609473f05c4f5d1d72a579b9f6af66ccd

### DIFF
--- a/scintilla/win32/scintilla.mak
+++ b/scintilla/win32/scintilla.mak
@@ -51,6 +51,9 @@ LDFLAGS=$(LDDEBUG) $(LDFLAGS)
 CXXFLAGS=$(CXXFLAGS) $(CXXNDEBUG)
 !ENDIF
 
+#unconditionally use scintilla namespace
+CXXFLAGS=$(CXXFLAGS) -DSCI_NAMESPACE
+
 INCLUDEDIRS=-I../include -I../src -I../lexlib
 CXXFLAGS=$(CXXFLAGS) $(INCLUDEDIRS)
 
--- a/scintilla/boostregex/AnsiDocumentIterator.h
+++ b/scintilla/boostregex/AnsiDocumentIterator.h
@@ -1,6 +1,10 @@
 ﻿#ifndef ANSIDOCUMENTITERATOR_H_12481491281240
 #define ANSIDOCUMENTITERATOR_H_12481491281240
 
+#ifdef SCI_NAMESPACE
+namespace Scintilla {
+#endif
+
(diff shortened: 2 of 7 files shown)


milipili:
@chcg What's this PR supposed to fix? (aka how this PR prepares for a static build) I've already a static build of scintilla via CMake (PR very soon) but I fail to see how those namespaces in those particular files are helpful (but I may have missed something).

milipili:
(even if they are useful for consistency with scintilla source files of course)

milipili:
@chcg Oh forget what I said. We just did not fix the problem the same way :)

scintilla/win32/scintilla.mak
```diff
@@ -19,7 +19,7 @@ LEXLIB=Lexers.lib
 
 LD=link
 
-CRTFLAGS=-D_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES=1 -D_CRT_SECURE_NO_DEPRECATE=1
+CRTFLAGS=-D_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES=1 -D_CRT_SECURE_NO_DEPRECATE=1 -DSCI_NAMESPACE=1
```
milipili: This should be another commit. First commit for adding the missing namespaces, then another for enabling those namespaces.

donho:
@chcg x64 build failed in your PR. Could you fix that?

chcg:
@donho x64 build errors a caused by   ..\src\WinControls\DocumentMap\documentMap.cpp(254): error C2220: warning treated as error - no 'object' file generated [C:\projects\notepad-plus-plus\PowerEditor\visual.net\notepadPlus.vcxproj]

What about the splitting in two commits like requested from @milipili ? Should it be two commit within this PR or two PRs?

donho:
@chcg 
OK, my bad.

>  Should it be two commit within this PR or two PRs?
Just leave it as it is, please. It's OK to me.

chcg:
@donho Added again with a slight modification that is added separately and also adapted vcxproj file

milipili:
@chcg That's why I asked to rebase :)

milipili:
@chcg Well since infinitely useful for code review and for other tools such as git bisect, I would strongly recommend to do it... However it seems that @donho does not really mind for now, you can probably ignore that one.

PR #3075, opened 2017-03-19
Fixes for multi-line feature

AngryGamer:
Changes:

1. Text on tabs with multi-line are no longer clipped (g, y, etc).
2. Fix tab switching hotkeys  (**shift+MW**, **ctrl+MW**, **shift+ctrl+MW**)
3. Hijack **shift+MW** hotkey to move tabs rather than switch to them (while drag and drop enabled)
4. Fix drag and dropping tabs with multi-line enabled.


PowerEditor/src/WinControls/TabBar/TabBar.h
```diff
@@ -223,6 +223,8 @@ public :
     // it's the boss to decide if we do the drag N drop
     static bool _doDragNDrop;
 	// drag N drop members
+	bool _mightBeDragging;
+	int _dragCount;
```
donho: Both **_mightBeDragging** & **_dragCount** should be initialized.

AngryGamer: Done!

PR #3076, opened 2017-03-19
Remove duplicate code

SinghRajenM:
Code clean up: remove duplicate code


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -978,6 +978,7 @@ HWND CreateToolTip(int toolID, HWND hDlg, HINSTANCE hInst, const PTSTR pszText)
 
 bool isCertificateValidated(const generic_string & fullFilePath, const generic_string & subjectName2check)
 {
+	bool bRetVal = false;
```
donho: I would replace bRetVal to isOK.

SinghRajenM: It has been changed now!

SinghRajenM:
Done.

PR #3092, opened 2017-03-25
Add separator for context menu

intika:
Add separator before and after npp in context menu


MAPJe71:
The Windows Explorer context menu that is.

donho:
I just don't see the advantage.

PR #3125, opened 2017-04-01
converted readme.txt to readme.md

RaymondDavidson:


MAPJe71:
This `readme.txt` file will end up on the target system during installation.
The `readme.md` does make it easier to read on GitHub but doesn't make much sense on a target system without an application that can render Markdown text.
Keeping two separate files makes it a maintenance nightmare.
An alternative would be to automatically generate the `.txt` from the `.md` (or vise versa) during setup creation with a to-be-provided script/tool.

PR #3126, opened 2017-04-01
converted readme.txt to readme.md

RaymondDavidson:


donho:
I see no advantage for that.

PR #3157, opened 2017-04-12
New resource for auto-completion in batch (batch.xml)

marcoM32:
A new resource for auto-completion in batch scripting environment
Please see the issue: #3585


SinghRajenM:
@marcoM32 nsis should have been updated. So that it can be deployed with installer.

Issue #2896, opened 2017-02-10
change.log Always re-opened on application entry [7.3.1]


mancoast:
### Description of the Issue

Recently auto-updated to Notepad++ 7.3.1 32bit on host: winsvr2016 x64.

### Steps to Reproduce the Issue

1.  Execute software.
2. Close "change.log"
3.  close/re-open software

### Expected Behavior

"change.log" remains closed

### Actual Behavior

"change.log" always present in open document tabs

### Debug Information

Notepad++ v7.3.1   (32-bit)
Build time : Jan 17 2017 - 00:21:10
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

![image](https://cloud.githubusercontent.com/assets/7040493/22827095/0c24391e-ef64-11e6-9514-f95c7392a62b.png)

dail8859:
See [this](https://notepad-plus-plus.org/community/topic/13102/changelog-opens-every-time-i-start-npp) discussion on the forum. It seems to be due to pinning the instance of Notepad++ that was ran immediately after installing. 

So if you unpin the shortcut, open a fresh instance of N++, and then repin, it then everything should be fine.

mancoast:
work-around confirmed.

pryrt:
Please see [this NPP Community forum discussion](https://notepad-plus-plus.org/community/topic/13623/change-log-opens-every-time-software-is-launched-7-3-3/): the evidence is starting to point to it either being the installer's shortcut, or Win10 auto-pinning the installer's initial call to "notepad++.exe change.log".  If it were the user manually pinning the shortcut, it would make sense to ask the user to fix the shortcut.  But if it’s the installer that’s creating the bad shortcut, the long-term fix needs to be on the installer side.  And even if it’s Win10 that’s creating an additional “pin” (probably in the “recently added” or “recently used” or “frequently used” section) that includes the change.log – it might be helpful to see if there’s a way during install to convince Win10 to auto-pin the correct version of the shortcut rather than the change.log shortcut.  It seems rather inconvenient (dare I say “rude”) to ask every Win10 user NPP x64 user to manually fix auto-generated shortcuts when the evidence is pointing to it being either the installer directly, or Win10’s auto-pin behavior, that’s causing the “bad” shortcut, and not anything the user has control over.

[...]

Issue #3014, opened 2017-03-10
7.3.3 will open the changelog every time you launch NP++


strollingHeifer:
### Steps to Reproduce the Issue
Open up Notepad++ after installing the 7.3.3 version.
Read the changelog, then close it.
Open another document.
Close NP++.
Open it again.
You will find the changelog to be open again in a separate tab.

### Expected Behavior
The changelog should not open again after you closed it for the first time after updating.

I'm not sure if this is related to the 64-bit version, as I made the switch from 32-bit to 64-bit with the 7.3.3 update. So this might have happened on earlier releases of the 64-bit variant as well. It just never happened to me with the 32-bit variant.
I did remove my custom settings with the Uninstaller before installing the new version.

dail8859:
Duplicate of #2979 and #2896.

See #2896 for a fix.

strollingHeifer:
I'm sorry. I did search for it before creating this issue. But I only looked for issues containing "changelog" instead "change.log". Thanks for the quick respond though.

Issue #2979, opened 2017-03-01
change.log always shown on startup


Lukeeone:
### Description of the Issue
change.log does always open on startup of n++, previously it was only shown for first usage after an update

### Steps to Reproduce the Issue
1. use Notepad++ v 7.3.2 64 bit, Win 10 Pro 64bit
2. open Notepad++

### Expected Behavior
change.log is only shown the first time the app is used after an update

### Actual Behavior
change.log is always shown when the app is opened (also if closed during a session)

### Debug Information
Notepad++ v7.3.2   (64-bit)
Build time : Feb 12 2017 - 23:17:55
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll

dail8859:
Duplicate of #2896. See that issue for a fix.

PR #3162, opened 2017-04-14
Don't open change.log everytime

SinghRajenM:
Fixes issue: #2896, #2979, #3014 

I found two ways to fix these issues.
1. Updating NSIS script
2. Updating Npp source code
        a) From installer launch npp with some parameter such as -install or /install (notice that it will elevated)
        b) If this parameter is seen, then launch npp as non elevated process. This can be achieved by grabbing access token from desktop shell, then duplicate this token and create new npp process using this token.

This PR uses method 1. If Method 2 sounds good, let me know I will update the PR.

### DIFF
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -26,22 +26,28 @@
 ; Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
 Function LaunchNpp
-  ;Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '
+  ; Open notepad instance with same integrity level as explorer,
+  ; so that drag n drop continue to function even
+  ; Once npp is launched, show change.log file (this is to handle issues #2896, #2979, #3014)
+  ; Caveats:
+  ;		1. If launching npp takes more time (which is rare), changelog will not be shown
+  ;		2. If previous npp is configured as "Always in multi-instance mode", then
+  ;			a. Two npp instances will be opened which is not expected
+  ;			b. Second instance may not support drag n drop if current user's integrity level is not as admin
+  Exec '"$WINDIR\explorer.exe" "$INSTDIR\notepad++.exe"'
 
-  ; create shortcut in temp with to launch as argument
-  CreateShortCut "$TEMP\notepad++.lnk" "$INSTDIR\notepad++.exe" "$INSTDIR\change.log"
+  ; Max 5 seconds wait here to open change.log
+  ; If npp is not available even after 5 seconds, exit without showing change.log
 
-  ; Launch Notepad++ with same elevation as explorer
-  ; It will solve the drag an drop issue if explorer is running with lower privilege
-  Exec '"$WINDIR\explorer.exe" "$TEMP\notepad++.lnk"'
-
-  ; Caution : Please DONOT delete $TEMP\notepad++.lnk right after Exec command
-  ; In slower system it may take some time to open N++ while Exec command exist quickly
-  ; Also putting sleep does not make sense here as
-  ; 	- DONOT know how much to wait
-  ;		- Keeps installer UI open wait time period
-  ;		- If wait time is more, installer UI may become irresponsive
-
+  ${ForEach} $R1 1 5 + 1				; Loop to find opened Npp instance
+	System::Call 'kernel32::OpenMutex(i 0x100000, b 0, t "nppInstance") i .R0'
+	IntCmp $R0 0 NotYetExecuted
+		System::Call 'kernel32::CloseHandle(i $R0)'
+		Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '
+		${Break}
+	NotYetExecuted:
+		Sleep 1000
+  ${Next}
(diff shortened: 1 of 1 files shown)


donho:
@SinghRajenM 
So this bug is triggered by the following 2 lines ?
```
CreateShortCut "$TEMP\notepad++.lnk" "$INSTDIR\notepad++.exe" "$INSTDIR\change.log"
Exec '"$WINDIR\explorer.exe" "$TEMP\notepad++.lnk"'
```

Could you explain how please?

SinghRajenM:
With current setup, npp is launched (from finish dialog) by executing a shortcut created in temp folder. This was done to make npp integrity level same explorer's integrity level, so that drag & drop functionality continues to work.

This is wired situation that first time pinning works in different way than the second time for same process.

1. Launch npp from installer page and pin it to taskbar
2. See its property.
3. Unpin it from taskbar and pin it again **without restarting npp**
4. See its property.

image

There are few articles ([superuser](https://superuser.com/questions/420085/how-can-i-pin-a-program-to-the-windows-7-taskbar-with-parameters), [msdn blog](https://blogs.technet.microsoft.com/deploymentguys/2009/04/08/pin-items-to-the-start-menu-or-windows-7-taskbar-via-script/)) which show how to pin a program with its command line argument. These articles also talk about the creating shortcut and then pin it to task bar.

But the point of interest is: window behaves in different ways for 1st pinning and 2nd pinning or so on for same process. 
1st pinning => includes command line arguments in it.
2nd pinning => does not includes command line arguments
I could not figure out why so?

PowerEditor/installer/nsisInclude/tools.nsh
```diff
@@ -26,22 +26,28 @@
...
+  ${ForEach} $R1 1 5 + 1				; Loop to find opened Npp instance
+	System::Call 'kernel32::OpenMutex(i 0x100000, b 0, t "nppInstance") i .R0'
+	IntCmp $R0 0 NotYetExecuted
+		System::Call 'kernel32::CloseHandle(i $R0)'
+		Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '
+		${Break}
+	NotYetExecuted:
+		Sleep 1000
+  ${Next}
 FunctionEnd
```
donho: It's too complex for a simple task.
I suggest the following code:
```
ExecWait '"$WINDIR\explorer.exe" "$INSTDIR\notepad++.exe"'
Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '
```
What do you think?

SinghRajenM: There are two reasons, I didn't use ```ExecWait``` - 

1. As name suggest ```ExecWait``` waits until exe terminates. So ideally it should wait for exe (here exe is explorer). Though above solution works proper in my win10 pc, but I'm not sure if this works in all the versions of windows.

image

1. ```ExecWait``` does not know whether npp is launched or not. If npp takes sometime to initialize itself (may be in first installation or in slower system). The original problem may continue to persist if npp is get launched because of second line ```Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log" '```

donho: @SinghRajenM 
> As name suggest ExecWait waits until exe terminates.
> ExecWait does not know whether npp is launched or not.
You're right about 2 points above.

As you said in the comment:
> If previous npp is configured as "Always in multi-instance mode", then 
> a. Two npp instances will be opened which is not expected 
> b. Second instance may not support drag n drop if current user's integrity level is not as admin 

this pr will make some regression if multi-instance mode is ON. Any alternative solution?

SinghRajenM: At this moment I can see two option.

1. Use UAC plugin (preferred). Many commercial s/w uses it.
2. As stated before, just execute ```Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log"'``` with additional flag (e.g. /Installed or -Installed or any other). If this parameter is seen, then launch npp as non elevated process. This can be achieved by grabbing access token from desktop shell, then duplicate this token and create new npp process using this token. UAC plugin internally does the same.

SinghRajenM: At this moment I can see two options.

1. Use UAC plugin (preferred). Many commercial s/w uses it.
2. As stated before, just execute ```Exec '"$INSTDIR\notepad++.exe" "$INSTDIR\change.log"'``` with additional flag (e.g. /Installed or -Installed or any other). If this parameter is seen, then launch npp as non elevated process. This can be achieved by grabbing access token from desktop shell, then duplicate this token and create new npp process using this token. UAC plugin internally does the same.

MAPJe71: Option 3: Remove opening `change.log`.

AngryGamer: Option 4: Just add the file to sessions.xml or add a setting to config.xml which toggles.

AngryGamer: Couldn't you also store build/version in config.xml and open change.log if it differs? This would bypass needing to launch an elevated notepad++

PR #3169, opened 2017-04-16
modify autocompletion to make it support .tex-specific / recognition

zyu47:
Any TeX-based keyword has a backslash / (such as /starttex and /endtex ) in front of them. The current version of NPP is not identifying a backslash for either extracting words or showing the list. This PR added the feature of the following:

1. These changes are specific to TeX files ending with '.tex';
2. Autocompletion keywords only show up when there is a / in front the typed word;
3. getWordArray ignores keywords that do not start with a /

### DIFF
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -65,7 +65,20 @@ bool AutoCompletion::showApiComplete()
 
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
-	_pEditView->showAutoComletion(curPos - startPos, _keyWords.c_str());
+
+	//determine if the file is a tex file
+	const TCHAR *fn = ((_pEditView)->getCurrentBuffer())->getFileName();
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -65,7 +65,20 @@ bool AutoCompletion::showApiComplete()
...
+	bool isTEX = fnsize >= 4 && fn[fnsize - 1] == 'x' && fn[fnsize - 2] == 'e' \
+		&& fn[fnsize - 3] == 't' && fn[fnsize - 4] == '.';
+
+	if (isTEX) { //if this is a TEX file
+		TCHAR initChars[2];
+		_pEditView->getGenericText(initChars, 2, startPos - 1, startPos);
+		if(initChars[0] == '\\')
+			_pEditView->showAutoComletion(curPos - startPos, _keyWords.c_str());
+	}else
+		_pEditView->showAutoComletion(curPos - startPos, _keyWords.c_str());
```
MAPJe71: You added the same code (more or less) four times to determine whether the file is a 'tex' file. Create a separate function instead e.g. `bool isTexFile( ... )`.
On the other hand it might not be necessary to create such a function as there are examples throughout the code how to determine whether special language conditions apply. Search for `getLangType` and find out how it's handled.

Issue #2532, opened 2016-11-05
Function completion ignores additionalWordChar


dail8859:
### Description of the Issue
Triggering the auto-completion (e.g. Ctrl+Space) does not use additionalWordChar to detect the keyword before the cursor.

### Steps to Reproduce the Issue

1. Use the xml auto complete file below (it is for Lua as an example)
2. Open new file
3. Set language to Lua
4. Type `tab` then Ctrl+Space
5. The auto complete will come up and suggest "table" and "table.insert" which is correct.
6. Continue typing so that the file contains `table.ins` then press Ctrl+Space
7. Nothing happens

### Expected Behavior
It should suggest the list of values from the auto-completion file. 

*Note:* If you type `table.insert` and then `(` it correctly shows the calltip because it uses the additionalWordChar to detect the function name.

### Actual Behavior
Nothing shows up

### Debug Information
```
Notepad++ v7.2   (32-bit)
Build time : Nov  2 2016 - 11:26:20
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7
Plugins : ComparePlugin.dll DoxyIt.dll LuaScript.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

```

The zipped up xml file:

[lua.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/573006/lua.zip)

donho:
@dail8859
> Triggering the auto-completion (e.g. Ctrl+Space) does not use additionalWordChar to detect the keyword before the cursor.

It seems trivial to me or am I missing somethings ?
If you cannot provide a PR for it, I will investigate this problem.

dail8859:
> If you cannot provide a PR for it

Due to my current schedule I do not believe I'll be able to get to it soon.

PR #3139, opened 2017-04-06
Fix buffer overrun for wordchar list

dail8859:
There have been reports of the wordchar list getting unexpected characters (see #3137, #3105, #3100, [here](https://notepad-plus-plus.org/community/topic/13568/double-click-selects-into-previous-line), [here](https://notepad-plus-plus.org/community/topic/13509/sci_wordleft-and-sci_wordright-treats-sentences-as-words-in-new-file/2), [here](https://notepad-plus-plus.org/community/topic/13593/inconsistent-double-click-selection-of-quoted-text)). Though I can't say for certain this PR will fix those bugs since I can't consistently reproduce the issue. Either way this still is a potential buffer overrun that needs fixed.

Also see 71edfb2dba3fd94b9b8b642e3d7dd7c00eeed307 which fixed a very similar issue.

PR #3198, opened 2017-04-22
Auto complete dialog not showing on AdditionaWordCharacter

tr12345:
Fixed issue displaying auto complete dialog when typing a defined additionalWordChar.

For example if auto complete list contains:
object
object.something
object.somethingelse

Then you configure your APIs\language.xml with additionalWordChar="." the dialog will not display when pressing "period" if you already selected "object" from the auto complete.

This update will add the characters defined in the language.xml to the current word characters list before displaying the autocomplete dialog then default back to the configured word characters list.

I did it this way so I could avoid putting the character in Notepad++ configuration for word character list.  This way auto complete will still display but if I double click a word in notepad++ it will not include the periods in the selection.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -46,6 +46,48 @@ static bool isAllDigits(const generic_string &str)
 	return std::all_of(str.begin(), str.end(), ::isdigit);
 }
 
+bool AutoCompletion::addLanguageCustomChars()
+{
+	if (_funcCalltip._additionalWordChar == TEXT(""))
+		return false;
+
+	string charList;
+	auto defaultCharListLen = _pEditView->execute(SCI_GETWORDCHARS);
+	char *defaultCharList = new char[defaultCharListLen + 1];
+	_pEditView->execute(SCI_GETWORDCHARS, 0, reinterpret_cast<LPARAM>(defaultCharList));
+	charList = defaultCharList;
+	delete[] defaultCharList;
+
+	string additionalWordChar = std::string(_funcCalltip._additionalWordChar.begin(), _funcCalltip._additionalWordChar.end());
+	string chars2addStr;
+	for (size_t i = 0; i < additionalWordChar.length(); ++i)
+	{
+		bool found = false;
+		char char2check = additionalWordChar[i];
+		for (size_t j = 0; j < charList.length(); ++j)
+		{
+			char wordChar = charList[j];
+			if (char2check == wordChar)
+			{
+				found = true;
+				break;
+			}
+		}
+		if (not found)
+		{
+			chars2addStr.push_back(char2check);
+		}
+	}
+
+	if (not chars2addStr.empty())
+	{
+		charList += chars2addStr;
+		_pEditView->execute(SCI_SETWORDCHARS, 0, reinterpret_cast<LPARAM>(charList.c_str()));
+		return true;
+	}
+
+	return false;
+}
 
 bool AutoCompletion::showApiComplete()
 {
@@ -734,6 +776,9 @@ void AutoCompletion::update(int character)
 	if (_pEditView->execute(SCI_AUTOCACTIVE) != 0)
 		return;
 
+	//If additionalWordChar defined for current Language include them in word characters list.
+	bool isCharsAdded = addLanguageCustomChars();
+	
 	const int wordSize = 64;
 	TCHAR s[wordSize];
 	_pEditView->getWordToCurrentPos(s, wordSize);
@@ -747,6 +792,10 @@ void AutoCompletion::update(int character)
 		else if (nppGUI._autocStatus == nppGUI.autoc_both)
 			showApiAndWordComplete();
 	}
+	
+	//After auto complete has been displayed default back to configured word characters.
+	if (isCharsAdded)
+		_pEditView->setWordChars();
 }
 
 void AutoCompletion::callTipClick(int direction) {
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.h
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.h
@@ -70,6 +70,8 @@ public:
 
 	bool setLanguage(LangType language);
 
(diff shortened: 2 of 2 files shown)


PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -46,6 +46,48 @@ static bool isAllDigits(const generic_string &str)
...
+bool AutoCompletion::addLanguageCustomChars()
+{
+	if (_funcCalltip._additionalWordChar == TEXT(""))
+		return false;
+
+	string charList;
+	auto defaultCharListLen = _pEditView->execute(SCI_GETWORDCHARS);
+	char *defaultCharList = new char[defaultCharListLen + 1];
+	_pEditView->execute(SCI_GETWORDCHARS, 0, reinterpret_cast<LPARAM>(defaultCharList));
+	charList = defaultCharList;
```
dail8859: This has a potential buffer overrun. See #3139 for a similar fix when using `SCI_GETWORDCHARS`.

PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -46,6 +46,48 @@ static bool isAllDigits(const generic_string &str)
 	return std::all_of(str.begin(), str.end(), ::isdigit);
 }
 
+bool AutoCompletion::addLanguageCustomChars()
+{
+	if (_funcCalltip._additionalWordChar == TEXT(""))
```
dail8859: It is recommend to use [`empty()`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#always-use-empty-for-testing-if-a-string-is-empty-or-not)

dail8859:
Does this also fix #2532 when using Ctrl+Space to show the autocompletion?

tr12345:
Applied your suggestions.  Thanks for the help.  It does't currently fix #2532 but with some minor tweaks it can.  Should have an update soon.

PowerEditor/src/ScitillaComponent/AutoCompletion.h
```diff
@@ -70,6 +70,8 @@ class AutoCompletion {
 
 	bool setLanguage(LangType language);
 
+	//Add language defined additional word characters to current word characters
+	bool addLanguageCustomChars();
```
dail8859: This can be made private.

PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -46,15 +46,65 @@ static bool isAllDigits(const generic_string &str)
...
+	charList = defaultCharList;
+	delete[] defaultCharList;
+
+	string additionalWordChar = std::string(_funcCalltip._additionalWordChar.begin(), _funcCalltip._additionalWordChar.end());
+	string chars2addStr;
+	for (size_t i = 0; i < additionalWordChar.length(); ++i)
+	{
+		bool found = false;
+		char char2check = additionalWordChar[i];
+		for (size_t j = 0; j < charList.length(); ++j)
```
dail8859: Could this `for` and next `if` statement be condensed to:

```c
if (charList.find(char2check) == string::npos)
{
	chars2addStr.push_back(char2check);
}
```

dail8859:
Functionality-wise this seems to work fine and can confirm this does fix #2532.

PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
dail8859: Actually I think this entire `for` loop starting at line 64 could be replaced with something like:

```c
size_t pos = 0;
while((pos = additionalWordChar.find_first_not_of(charList, pos)) != string::npos)
{
	chars2addStr.push_back(additionalWordChar[pos]);
	pos++;
}
```

I'd suggest double checking this though to make sure it works as expected.

tr12345: I'll take a look at this but originally this looping structure was copied from ScintillaEditView.cpp for consistency.

tr12345: This looks to work as expected I'll submit an update.

PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -46,15 +46,53 @@ static bool isAllDigits(const generic_string &str)
 	return std::all_of(str.begin(), str.end(), ::isdigit);
```
SinghRajenM: @dail8859 as per documentation ```isdigit``` behavior is undefined if the value of ch is not representable as ```unsigned char``` and is not equal to ```EOF```.

So in case of Asian languages this statement throws assert failure (Though it is irrelevant for release version).
Is there any issue if this is replaced with
``` C++
return str.find_first_not_of("0123456789") == std::string::npos;
```

dail8859: Ah I see what you mean. Is there some other form of `isdigit()` that would work? `find_first_not_of()` should work but a native function to check the character would probably be a bit more efficient and clearer...in the end shouldn't really matter though.

SinghRajenM: ```_tcstol``` for each character (because string may be large enough to convert into max possible integer) might an alternative. (using lamda)
```C++
TCHAR *dummy = nullptr;
TCHAR input[] = _T("5x");
long l = _tcstol(input, &dummy , 10);
if (dummy && _tcsclen(dummy ))
{
          cout<<"Not a valid string to convert to integer";
}
```

dail8859: Would something as straight forward as [`iswdigit()`](http://www.cplusplus.com/reference/cwctype/iswdigit/) work?

SinghRajenM: I never used it and moreover it's sample example code at [cpp reference](http://en.cppreference.com/w/cpp/string/wide/iswdigit) is bit confusing.

PR #3200, opened 2017-04-22
Improve whitespace visibility

SinghRajenM:
White spaces are not visible properly. Visibility of white-space which may be useful for languages in which white space is significant, such as Python.

image

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.h
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.h
@@ -362,6 +362,7 @@ public:
 
 	void showWSAndTab(bool willBeShowed = true) {
 		execute(SCI_SETVIEWWS, willBeShowed?SCWS_VISIBLEALWAYS:SCWS_INVISIBLE);
+		execute(SCI_SETWHITESPACESIZE, 2, 0);
 	};
 
 	void showEOL(bool willBeShowed = true) {


MAPJe71:
LOL, using LuaScript for that right now by adding the following to `startup.lua`:
```lua
-- make the space 'dots' more visible
editor.WhitespaceSize = 2
```

naml3i:
Is the status of the request "make size of whitespace symbol configurable" marked as "CLOSED"?
The only reasons I stay with version 7.3.3 is that since 7.4 the dot representing whitespace is made too much visible that it hurts my eye. I was living just fine with the tiny 1px dot so the 2x2 is too intrusive. Anyway I'm OK with people who want it bigger (although I guess they have eye sight problem?), so customizable dot size is nice.

MAPJe71:
@naml3i 
You can use the technique I mentioned above to make it smaller i.e. using LuaScript plugin and add the following to `startup.lua`:
```lua
-- make the space 'dots' less intrusive/hurtful to the eyes
editor.WhitespaceSize = 1
```

Makes it kind of customizable :grin:

naml3i:
@MAPJe71 Thanks a lot! Easier than I thought :smile: 
I gave a thought that it's too much of a hassle to even bother installing a plugin and edit some startup script. 

https://notepad-plus-plus.org/community/topic/13775/changing-colors-of-the-space-and-other-symbols-when-view-all-is-turned-on
Among the methods suggested here, I prefer LuaScript to PythonScript (it's more lightweight and straightforward).

PR #3218, opened 2017-04-27
FIxed Preference(MISC) UI Issue

SinghRajenM:
Issue:

1. Controls on preference (MISC) were overlapped due to newly added controls.
2. Tab order was not proper.
3. Changed phrase from "Snapshot on map" to "Snapshot on document map" as it is bit confusing what/which map.
image

After Fix:
Moved "Enable Notepad++ auto-updater" in the last
image


SinghRajenM:
@donho Done!

donho:
@SinghRajenM What you have done (the first try) was good, you need only put "Enable Notepad++ auto-updater" checkbox as the last item. 
But let's keep current layout.

Issue #3225, opened 2017-04-30
Feature request (with PR): "Remove lines containing..." in the FindReplace dialog


btomer:
### Feature description
The requested feature would allow to enter a search term (using the existing search modes) and having all lines that contain the search term deleted.
I find myself needing this feature all the time and I think it will be useful to others as well.

For example, for the following file:
```
Check is the first word of the document
Line without the search term 1
The last word in line is check
Line without the search term 2
A line in the middle that contains "check" in it
Line without the search term 3
Line without the search term 4
The last word in the document is check
```
Using the search term "check" (with match case disabled) and pressing "Remove containing lines" would result in the file:
```
Line without the search term 1
Line without the search term 2
Line without the search term 3
Line without the search term 4
```

### Status
I wrote the code for this feature, it compiles on my machine for x86 and for x64, and it passes the manual tests I put it through.

donho:
https://www.youtube.com/watch?v=SiPLZzRqEwM

PR #3226, opened 2017-04-30
"Remove lines containing..." feature (issue: #3225)

btomer:
This PR adds a new tab in the FindReplace dialog, that contains a "Remove containing lines" command.
The command description can be found in issue #3225 .

The PR contains:
* Some changes in FindReplaceDlg.cpp that make adding a new tab easier.
* The new "Remove Lines" tab and "Remove containing lines" button in the FindReplace dialog.
* The actual new functionality, which is inside processRange, and activated when it receives a new ProcessOperation type which is called "ProcessRemoveContainingLines".
* A new menu item for the tab, which is under the Search menu.
* A new quote easter egg that I've added. Please forgive me but I could not stop myself from trying to enter eternal glory :) If this is not appropriate, my apologies, and please still consider the real changes (I split the easter egg into a different commit so it will be easy to discard).

If you have any changes that you would like me to make to the code, please feel free to ask.


donho:
This feature should not be in Find & Replace dialog anyway.
And this feature exists already, and even more flexible to remove or keep only the lines :
https://www.youtube.com/watch?v=SiPLZzRqEwM

PR #3289, opened 2017-05-16
Auto-completion: ConTeXt lexer-specific backslash support

zyu47:
ConTeXt-based keyword has a backslash / (such as /starttext and /endtext ) in front of them. The current version of NPP is not identifying a backslash for either extracting words or showing the list. This PR added the feature of the following:

1. These changes are specific to ConTeXt lexer;
2. Autocompletion keywords only show up when there is a / in front of the typed word;
3. wordCompletion takes / into consideration


dail8859:
What is ConTeXt? A User Defined Language?

zyu47:
@dail8859 It's a type of a TeX-based typesetting system, like LaTeX. I guess we can make it general for both ConTeXt and TeX. See the wikipedia page: https://en.wikipedia.org/wiki/ConTeXt

dail8859:
I'm not the one to accept/reject this PR but these are some initial thoughts:
1. This seems overly specific to a lexer that Notepad++ doesn't support natively.
2. It is not easily extensible if/when more files are added as corner cases.
3. Constantly calling `getApiFileName()` + a string compare seems unnecessary. It would probably be better to save a flag and/or a string with extra chars within `setLanguage()` and then reuse it.
4. Is there a better place to grab language specific characters? I'm not sure what all the UDL files contain but I know at least for the auto complete files it has an `additionalWordChar` attribute that may be applicable (not saying this is the solution).

zyu47:
There is already a native TeX-language lexer that comes bundled with Npp. This fix to auto-completion can be made to work for the native TeX lexer in addition to any other lexer that uses backslash commands. Only the native tex file needs updating (add backslashes). Other files won't be affected.

Note that other scintilla editors such as Scite support autocompletion for TeX-backslash commands out of the box.

dail8859:
> already a native TeX-language lexer that comes bundled with Npp

Ah, gotcha. Well then it would be much more efficient to check the lang type to see if it is set to `L_TEX` instead of item 3 I listed above.

MAPJe71:
Please squash your commits into one commit to make it easier to review.

PR #3302, opened 2017-05-20
Remove duplicate javascript language definition

Sophist-UK:


MAPJe71:
AFAIK the duplicate is intentional that's why there is a comment 'Embedded javascript'.

Sophist-UK:
Yes- I saw that comment. But what purpose does this have except having less keywords. Let's see if someone who knows for certain can enlighten us.

donho:
@Sophist-UK If you don't know what you're doing, just don't do it.

PR #3318, opened 2017-05-25
Add strongly typed interface for Scintilla

dail8859:
Currently to interact with Scintilla, Notepad++ uses the `execute()` call which requires lots of casting and is not able to provide any type checking for what gets passed in. There are also numerous inconsistencies about variable types. For example to represent a "position" within a document, Notepad++ uses a variety of types, such as `size_t`, `long`, `int32_t`, `int`, and `LRESULT` with no real reason why. Scintilla only uses a small set of data types so most of the data that gets passed to Scintilla gets `static_cast<>()` immediately.

This PR adds a thin layer (most likely optimized away) which provides a readable and strongly typed interface for interacting with Scintilla. This PR is broken up into multiple commits for easier review. The first commit simply adds the interface. The 2nd and 3rd updates and cleans up the ScintillaEditView to utilize this interface, and the last replaces nearly all other calls to `execute()` with its alternative.

### About the Interface

Each call to Scintilla (e.g. `SCI_XXX`) has an associated method that makes the code readable and enforces correct variable types. For example:

```cpp
void AddText(int length, const char* text) const
{
    execute(SCI_ADDTEXT, length, text);
}
```

Some calls are also overloaded to accept `std::string`s such as:

```cpp
void AddText(const std::string& text) const
{
    execute(SCI_ADDTEXT, text.length(), text.c_str());
}
```

In other cases it is also able to allocate strings as return types (which avoids potential pitfalls such as buffer overruns):

```cpp
int GetText(int length, char* text) const
{
    sptr_t res = execute(SCI_GETTEXT, length, text);
    return static_cast<int>(res);
}

std::string GetText() const
{
    auto size = execute(SCI_GETTEXT, SCI_UNUSED, NULL);
    std::string text(size + 1, '\0');
    execute(SCI_GETTEXT, text.length(), &text[0]);
    trim(text);
    return text;
}
```

This interface can be generated with a Python script (which I can provide later) by reading the scintilla.iface file. This does away with quite a bit of casting when using the ScintillaEditView (other than casting `size_t` to `int` in places). 

I've attempted to make the minimum number of changes to use this interface throughout the code base. I'm hoping this will lead to more cleanup later by providing a cleaner way of getting data into and out of Scintilla.

---

Since this is a non-trivial code patch I am open to idea/feedback/comments/issues and hope that this gets thoroughly reviewed and tested by anyone willing.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -298,8 +298,8 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_subEditView.execute(SCI_SETEDGECOLUMN, svp1._edgeNbColumn);
 	_subEditView.execute(SCI_SETEDGEMODE, svp1._edgeMode);
 
-	_mainEditView.showEOL(svp1._eolShow);
-	_subEditView.showEOL(svp1._eolShow);
+	_mainEditView.SetViewEOL(svp1._eolShow);
+	_subEditView.SetViewEOL(svp1._eolShow);
 
 	_mainEditView.showWSAndTab(svp1._whiteSpaceShow);
 	_subEditView.showWSAndTab(svp1._whiteSpaceShow);
@@ -2533,7 +2533,7 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 			// Get previous line's Indent
 			if (prevLine >= 0)
 			{
-				indentAmountPrevLine = _pEditView->getLineIndent(prevLine);
+				indentAmountPrevLine = _pEditView->GetLineIndentation(prevLine);
 			}
 
 			// get previous char from current line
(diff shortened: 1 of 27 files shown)


donho:
I'm not the fan of ProperCase as method name (the name conversion is indicated in coding style BTW).
I don't like neither the change name of some ScintillaEditView methods (for me, it's more explicite than the message itself). 
Furthermore all the added methods makes ScintillaEditView less readable and maintainable.
With large amount of modification, we have to take a high risk of regression, plus all the minus above.

I'm aware of the conversion problem. but I don't see the solution brought by this PR.  I'll be opened for the other solutions with much lighter modification.

dail8859:
@donho

> I'm not the fan of ProperCase as method name

Ok, this can easily be changed.

> I don't like neither the change name of some ScintillaEditView methods

Yes some methods were changed since they simply wrap a single `execute()` call but these can be added back in if its that big of a deal.

> all the added methods makes ScintillaEditView less readable

These could be put in a separate class that the ScintillaEditView inherits from or these new methods put in a header file and simply add a `#include "additionalMethods.h"` inside the ScintillaEditView class.

> and maintainable

It can be generated by a Python script. So once the script is in place there is very little that needs done.

> we have to take a high risk of regression

Is the code that fragile it can not be made more robust?

> with much lighter modification.

I don't know how it could be made any lighter.

CookiePLMonster:
Naming convention issues aside, I don't see why this is rejected. While changes are quite substantial, they are easy to review and thus risk of regressions is low.

If anything, this PR is very likely to **uncover** issues caused by weak typing.

PR #3322, opened 2017-05-27
Update Italian Translation for document peeker, version 7.4.1

Ices-Eyes:
Add translations for document peeker


xdev31:
:+1: Translation is correct

PR #3334, opened 2017-05-31
Fix C6328: Potential argument type mismatch

shoelzer:
In several places string functions specified 'int' values but 'long' or 'size_t' was actually being used. I fixed them by specifying 'long' and casting inputs to 'long'. I found these using Visual Studio's Code Analysis feature.

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -455,7 +455,7 @@ void PluginsManager::runPluginCommand(size_t i)
 			catch (...)
 			{
 				TCHAR funcInfo[128];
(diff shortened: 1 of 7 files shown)


shoelzer:
@MAPJe71 I can update this PR, but I'm not clear what "C++ style" means. Do you mean like `long(i)`?

MAPJe71:
`static_cast<long>(i)`

shoelzer:
Updated PR to use `static_cast` everywhere.

donho:
What's the point of this PR ?

shoelzer:
@donho The point is to make argument types match the format string. All of these format strings specified `int` (using `%i` or `%d`), but the arguments change size/type when compiled as 32 or 64-bit. I believe my PR makes the types match in all cases.

Issue #449, opened 2015-07-09
Legibility in the find/replace dialog


Rosseiro:
Hi, 
I love Notepad++ but it appears that it's a little more difficult to use the find/replace dialog as it was supposed to be. Specially for us getting old, reading the input field in that default (unchangeable, as far as I know) font is hard. Especially when we're typing a regex, sometimes I can't see if I typed that pipe in the right place. So my suggestion is: allow us to change that font in the "find" field to a more readable, monospace font. 

Also, even with the transparency, it's sometimes tedious to switch between the dialog and the main window. Other editors have a built-in find/replace dialog in the main interface, some even with a highlight-as-you-type feature.

Thanks!

milipili:
It should be a fixed-width font anyway by default.

DavidBailey100:
I wish this dialog would have explicit search up and search down buttons. Sure you can set the direction to 'up', but then that remains set and thus easily messes up subsequent searches.

dail8859:
I'd even be ok with shift+enter or shift+"click search" to do a reverse search...although it's not a very intuitive solution.

DaveyD:
I triple David's suggestion..... Would be a very useful feature

alexhenzell:
@DavidBailey100
Just raised this issue (#506): Optional auto-reset of 'Find/Replace/Find in Files/Mark' settings
May be relevant to comment regarding 'up' direction

h-h-h-h:
@Rosseiro:
> allow us to change that font in the "find" field to a more readable, monospace font.

Issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/563 adds some suggestions as light gray text background.
>Other editors have a built-in find/replace dialog in the main interface, some even with a highlight-as-you-type feature.

Actually, there is a feature like this. You make it appear by hitting <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>I</kbd>. There are additional suggestions for this feature in issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/811.

@DavidBailey100:
> I wish this dialog would have explicit search up and search down buttons. Sure you can set the direction to 'up', but then that remains set and thus easily messes up subsequent searches.

Not a bad idea! What about this:
```
+-----------+ +---+
| Find Next | | ^ |
+-----------+ +---+
+-----------------+
|      Count      |
+-----------------+
```

---
[...]

PR #3337, opened 2017-06-01
#449 Monospace font in search and replace combobox

tonyk24:


tonyk24:
Before and after:

after
before

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -681,6 +681,15 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 	{
 		case WM_INITDIALOG :
 		{
+			HWND hFindCombo = ::GetDlgItem(_hSelf, IDFINDWHAT);
+			HWND hReplaceCombo = ::GetDlgItem(_hSelf, IDREPLACEWITH);
+			HWND hFiltersCombo = ::GetDlgItem(_hSelf, IDD_FINDINFILES_FILTERS_COMBO);
+
+			HFONT hf = CreateFont(-12, 0, 0, 0, 0, FALSE, 0, 0, 0, 0, 0, 0, 0, _T("Courier New"));
```
SinghRajenM: Somewhere ```DeleteObject``` should be called, shouldn't?

tonyk24: True that.

Accidently made change to a new pull request:

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3361

Ede123:
I'm afraid I don't like it:
* it unnecessarily blows up the size of the text boxes
* it's bad for the overall visual appearance of the dialog
* most importantly: it allows for a lot less characters to be visible when I'm already often using more than the full size of the input which reduces efficiency a lot

The only reason where a mono-spaced font would be useful is when opening the dropdown, but this is a very minor improvement an in my opinion it does not outweigh the downsides by far.

A useful change would the better visibility of upper case "i" vs lower case "l" but for that there are much better options (e.g. usage of [Source Sans Pro](https://fonts.google.com/specimen/Source+Sans+Pro) or similar)

sasumner:
With the relatively new (I think 7.4.2 and beyond) ability to resize the Find dialog window, I think I like this!  Without that ability I would have to agree with @Ede123 's "most importantly" comment.

donho:
Let's keep the current behaviour.

sasumner:
Wow, this is so old I'd forgotten about it.  VERY coincidentally, just yesterday I implemented a Pythonscript solution so that I can have this behavior with any Notepad++ version.  Yay for me! 

image

PR #3339, opened 2017-06-01
#449 Find previous and next instead of combo direction Down/Up

tonyk24:


dail8859:
Can you post a screenshot?

tonyk24:
Partly swedish.

https://ibb.co/fBRLXa

donho:
Interesting! I will consider it.

justanotheranonymoususer:
Am I the only one who doesn't like this change?

dinkumoil:
@justanotheranonymoususer : No, you are not. ;)

This change brought a bunch of issues:

1. Fixed button caption for "Find backwards" button, no translation ID available. This is annoying especially for users from countries using non-latin characters.
2. Even if there would be a translation ID for this button: its size is too small for some (if not most) languages in the world. Applies also to the "Find forwards" button.
3. When recording a macro using replace backwards no replacements are done at macro playback. Furthermore play back the recorded macro breaks the function of the "Find backwards" button - it searches forwards then. A restart of Npp is required to fix this.
4. When the "Replace" dialog is opened it is not possible to determine in which direction the "Find" respectively the "Replace"/"Replace all" buttons will work. Concerning "Replace all" this comes into effect if the "Wrap around" checkbox is unchecked.

Search and replace are important core functionalities of every text editor. Even if the bugs 1. to 3. would be fixed - the deficiency under 4. can not be fixed easily. But it's annoying in daily work.

I wish the changes would be reverted in the next release.

Issue #446, opened 2015-07-09
Use file type icons on the tabs


untroll:
I really like Notepad++ but when there are a lot of files opened, we can be lost easily because the icon on each tab is the same. It would be nice if Notepad++ could use the windows file type icons on each tab instead of having the same icon everywhere. It would give us a visual indication of which file we're looking for.

EditPad seems to have this feature already : 
![editpadlite 1](https://cloud.githubusercontent.com/assets/6537132/8605044/663e5b8e-2650-11e5-80f8-ad35e0980718.png)

If this feature is to hard to implement, maybe you should add way to let us associate an icon (from a library or from a file) to a tab and this icon has to be remembered if the file is closed and reopened.

Thank you!

Sébastien

untroll:
Will it be available in Notepad++ 7.0?

milipili:
I don't know. 7.0 I seriously doubt (unless the feature is trivial) but in the next release it is possible.

untroll:
Okay!

It would be a really usefull change!

ChesterLin:
Trying my hand at this:

![untitled](https://cloud.githubusercontent.com/assets/6014753/11025090/ede7d848-8664-11e5-98a0-4b181920c2d0.png)

Any concerns about styling? The overlay for unselected tabs needed to be removed to make icons look better. Not sure how to handle save vs unsaved files.

I'll submit a PR when this is closer to done.

Ede123:
One general note: I already have some improvements for the tab bar in line with #1075. It would be great if we could achieve those PRs to be compatible. Since @donho seems to be a little busy these days and does not merge PRs: Do you think it makes sense if you applied your changes on top of mine?
Since my PR mainly deals with positioning of elements you should be fine with just additionally changing the parts were the icons are drawn.

Regarding style:
* I see that you use the large (aka not "reduced") style of the tab bar. I'd increase the size of icons for this style (similar to the red diskette icon in your screenshot). On the other hand: Have you verified that your changes also work well with the reduced tab bar size (which I assume is what most people are using)?
* The feature should probably be optional. While it's definitely a cool feature, I personally like the current icons because they show me the state of files (saved, unsaved, write-protected, which is quite important to me) rather than the type of files (which is not that important to me).

ChesterLin:
@Ede123 Thanks for the comments. I believe that the changes can be done separately from your PR. I'm mainly working on refactoring IconList to support external icons at the moment.

[...]

PR #3354, opened 2017-06-04
#446 Shell icons used for file types

tonyk24:


tonyk24:
All four modes before and after (unedited, edited, monitoring and write protected.

Before:
file_type_icons_before

After (shell icons for unedited and edited, edited files are marked by *, monitoring icon and write protected icon kept as before): 
file_type_icons_after

donho:
The current file status icon is more consist than shell icon for Notepad++ IMO.

PR #3355, opened 2017-06-04
Improper autospacing of closing brace } #362

### LINKED ISSUE #362: Improper autospacing of closing brace "}"
Autospacing of the indent level gets confused if braces are used within string literals.
Known as issue within Javascript editor

Example-

``` javascript
function foo(bar)
{
    var bat = 0;
    while (bar.indexOf('{L', bat) >= 0) {
        var baz = bar.substring(bat, bar.indexOf('{', bat));
        }
    }
```

when type } on next line, it doesn't decrease the indentation level, but the next } does decrease indentation level, however it should close the function.  The editor highlights the correct matching { however.

tonyk24:
This is for C/C++/Java/C#/Objective-C/JS.

Not sure about the syntax rules regarding { in strings for languages Php/Jsp/Css.

Fix #362

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2481,35 +2481,6 @@ bool Notepad_plus::isConditionExprLine(int lineNumber)
 	return false;
 }
 
-int Notepad_plus::findMachedBracePos(size_t startPos, size_t endPos, char targetSymbol, char matchedSymbol)
-{
-	if (startPos == endPos)
-		return -1;
-
-	if (startPos > endPos) // backward
-	{
-		int balance = 0;
-		for (int i = int(startPos); i >= int(endPos); --i)
-		{
-			char aChar = static_cast<char>(_pEditView->execute(SCI_GETCHARAT, i));
-			if (aChar == targetSymbol)
-			{
-				if (balance == 0)
-					return i;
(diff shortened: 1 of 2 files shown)


tonyk24:
Updated commit now, using Scintillas function SCI_BRACEMATCH which takes into consideration the styling of the brackets as well.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2625,7 +2596,9 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 			int startPos = static_cast<int32_t>(_pEditView->execute(SCI_GETCURRENTPOS));
 			if (startPos != 0)
 				startPos -= 1;
-			int posFound = findMachedBracePos(startPos - 1, 0, '{', '}');
+
+			_pEditView->execute(SCI_COLOURISE, 0, -1);
```
dail8859: I don't think calling `SCI_COLOURISE` is a good idea because it forces the entire document to get re-lexed.

chcg: Maybe added because of https://scintilla-interest.narkive.com/TWBDw58B/problem-with-sci-bracematch

donho:
I just don't see how your PR could solve this general problem.

Issue #448, opened 2015-07-09
Add left margin to typing area


aaaxx:
Having a left margin of at least half a character in the typing (code) area would be really helpful. Right now the text butts up to bookmark bar, line number bar, or window frame (depending on what's set up to be displayed in Preferences) and it's hard to select a piece of it at the beginning of a line without placing a bookmark or selecting the whole line by clicking on line number.

donho:
I prefer the current behaviour.

PR #3356, opened 2017-06-04
#448 Add left margin to typing area.

tonyk24:


tonyk24:
Added a margin of 7 points. Too big or too small?
screenshot2

Ede123:
Why should a margin be needed at all? Just wastes space as far as I can see...

donho:
I prefer the current visual behaviour.

PR #3367, opened 2017-06-06
Fix hung UI thread while holding left click on a tab and not moving.

AngryGamer:
::DragDetect seems to being blocking UI updates for up to 2 seconds if you hold left click on a tab and don't move. This manifests itself as two issues depending on OS environment. Either [all the tabs completely disappear for 2 seconds](https://media.giphy.com/media/l0Iyldxa0P4StXTyM/giphy.gif) or the tab is selected but the document contents don't reflect the tab change.

To reproduce, enable drag and drop and either disable Multi-line or have only 1 row of tabs, then don't move while holding left click on a tab.

[Fixes issue reported here](https://notepad-plus-plus.org/community/topic/13881/tab-bar-flashes-when-dragging-tabs)

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -598,26 +598,7 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 
             if (_doDragNDrop)
             {
-				// ::DragDetect does not work with TCS_BUTTONS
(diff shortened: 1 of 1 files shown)


SinghRajenM:
I observed this issue too. But the interesting point is: this (UI blocking and complete tab disappears for 2 seconds) happens **only when Doc peeker on tab is activated**. _Otherwise it does not happen at all._

AngryGamer:
I can confirm the first issue is with Doc peeker enabled and the second issue is with Doc peeker disabled.

shravan2x:
I can confirm that this patch works.

donho:
@AngryGamer @SinghRajenM 
Any idea why Document Peeker causes such issue?

Issue #685, opened 2015-08-08
Non-blocking find/replace actions


h-h-h-h:
Hello,

the work Notepad++ does triggered by the find-and-replace dialog should be done in a worker thread and should be cancelable.

This is because such actions can sometimes last very long and can even cause you to think the GUI will never respond again and to force-terminate the program. This is especially true for large files.

If there's a way to get this information, there should even be a progress bar. Alternatively, the information of the last processing time would be valuable, if it's a long one.

An optional sound after long operations would also be appreciated.

---

Related issue: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/365

---

Because the find-and-replace dialog already has many buttons, probably there shouldn't be added an additional <q>Cancel</q> button. Because it makes sense to cancel a running action before closing the dialog, the <q>Close</q> button could turn into a <q>Cancel</q> button while an action is running.

The following image is an illustration of how the progress could be indicated. It shows the bottom of the find-and-replace dialog.

![gui](https://cloud.githubusercontent.com/assets/13482553/9569523/7f443238-4f6e-11e5-9543-c2d313780c8f.png)

The remaining time should appear as soon as enough data was gathered for a non-jumping presentation (short fixed amount of time?) and only if it's greater than or equal to 2 seconds at this point.

---

The developer of the PCRE2 library told me via e-mail that he is implementing an end offset feature for the functions `pcre2_match()` and `pcre2_substitute()` that would allow progress features (see issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/816). Because it would be a huge performance hit to do additional progress-related processing on every character position of the subject string, one should let the library process the string repeatedly from a start offset to an end offset and do progress-related processing between those calls. (Example: Try to match starting from offsets 0 to 100, update progress controls, try to match starting from offsets 101 to 200, and so on.)

Because the processing time of different regular expressions can differ very much, the application should first process a short range of the subject string and see whether the PCRE2 function returns quickly. If it returned quickly, the length of the ranges processed at once can be increased. This way, the length can be leveled out at a frequency of, say, 20 Hz (50 ms between GUI updates).

<sup>_Old text about conversation with the developer:_
[...]

PR #3368, opened 2017-06-06
Feature #685 progress bar in search dialog

tonyk24:


donho:
Could you provid the info about this PR?

tonyk24:
There is a progress bar at the bottom of the find dialog, and showing the remaining time of the search/count/replace operation.

The message queue is processed inbetween every  100th found occurance, making the find dialog non-blocking. All buttons but the close button is disabled during find. Pressing close cancels the operation and closes the dialog.

tonyk24:
screenshot_progress2

Ede123:
Is this only used for "Search" and "Replace" ("Highlight"?) tabs?
Or is "Search in Files" also updated?

I'm asking because there is already a progress bar for find in files and some consistency might be desirable.

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -703,6 +703,8 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			 _findClosePos.left = p.x;
 			 _findClosePos.top = p.y + 10;
 
+			 ::SendMessage(GetDlgItem(_hSelf, IDC_PROGRESSBAR), PBM_SETPOS, (WPARAM)0, 0);
```
MAPJe71: Please use C++ style typecast (see also [Guidelines for pull requests](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests)).

tonyk24:
Currently show progress for each individual file while doing Find in Files. Perhaps not really consistant.

I see some options:
- Remove progress for individual files while doing Find in Files.
- Remove Find in files progress and just use progressbar in Find dialog
- Keep both but add the filename to status bar showing progress of individual file

tonyk24:
Find/Replace/Mark: Will show remaining time, and pressing close in find dialog will cancel the operation. As a note, the replace operation can be cancelled mid-replace, without reversing the allready replaced items.
progress_find

Find in files: Will update the status bar find-dialog, but not possible to canel mid-file. Find in files is otherwise working as before.
progress_findinfiles

donho:
I don't like the idea for putting more elements into find dialog which is already overloaded.

PR #3372, opened 2017-06-06
Fix folding in user-defined languages for non-windows line endings

### LINKED ISSUE #350: Broken folding when using LF instead of CRLF
For a user defined language with parentheses and brackets as folders.

No problem with CRLF:
image

Switch to LF and see:
image

switching EOL on the text below will trigger this behavior.

```
    "token" (
        "token" (
            "token" -> [
                (
                    "token" (
                        "version" -> "a4afdc5"
                    )
                )
            ]
        )
    )
    view "token" widget "token" : "token" (
        "token" -> state "token" ( )
    )
```

version 6.7.9.2 on Windows 7 x64.

### LINKED ISSUE #2873: UDL folding issues with CR or LF only
When defining your own User-Defined Language, Operators and Delimiters will shift the folding by one line for every Operator and Delimiter that is followed by only a CR or LF, but works just fine if it is a CRLF
As a result, not only will the folding go beyond the original codeblock, it will also shift following codeblocks by the same amount of lines.

A couple of screenshots to highlight the issue:
Bugged:
cr
lf

Working as intended:
crlf

Also "Folding in code 2 style (seperators needed)" will not highlight the keyword if its at the end of the file. Folding works still fine thought.

ssk97:
fixes #2873 and fixes #350 (duplicate issues)

I also replaced two instances of "true" with EOL_SKIP_CHECK since that's what it evaluates to (and to reduce confusion since "true" would imply EOL_FORCE_CHECK)

### DIFF
--- a/scintilla/lexers/LexUser.cxx
+++ b/scintilla/lexers/LexUser.cxx
@@ -1715,8 +1715,8 @@ static void ColouriseUserDoc(unsigned int startPos, int length, int initStyle, W
                         // for delimiters that end with ((EOL))
                         if (newState != SCE_USER_STYLE_COMMENTLINE || (sc.ch != '\r' && sc.ch != '\n'))
                             sc.Forward(iter->length());
-                        if (sc.atLineStart)
-                            checkEOL = EOL_FORCE_CHECK;
+                        if (sc.atLineEnd)
+                            checkEOL = EOL_SKIP_CHECK;
 
                         // paint end of delimiter sequence
                         sc.SetState(newState);
--- a/scintilla/lexers/LexUser.cxx
+++ b/scintilla/lexers/LexUser.cxx
@@ -2085,6 +2085,8 @@ static void ColouriseUserDoc(unsigned int startPos, int length, int initStyle, W
                         // no closing sequence, start over from default
                         sc.SetState(SCE_USER_STYLE_IDENTIFIER);
                         dontMove = true;
+                        if (sc.atLineEnd)
+                            checkEOL = EOL_SKIP_CHECK;
                         break;
                     }
                 }
@@ -2129,7 +2131,7 @@ static void ColouriseUserDoc(unsigned int startPos, int length, int initStyle, W
                         sc.SetState(SCE_USER_STYLE_IDENTIFIER);
                         dontMove = true;
                         if (sc.atLineEnd)
-                            checkEOL = true;
+                            checkEOL = EOL_SKIP_CHECK;
 
                         levelNext--;
                         if (levelMinCurrent > levelNext)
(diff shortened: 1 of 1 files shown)


ssk97:
Update: I realized that the previous fix was only for operators (like semicolon in C), I've now fixed it so that delimiters (like quotes) don't break the folding either.

rmcurtis:
I'm so glad to see a fix for this!  It's been bugging me for years.  We even keep a bunch of files in Windows EOL format just so code folding works with our UDL.  Otherwise they should be Unix EOL format.

I managed to compile Notepad++ with your changes and it works great, just like it should.  Hope to see this fix in a released version soon.

Thanks again!

donho:
@ivan-radic Could you validate this PR please?

ivan-radic:
Thanks for notification @donho  and thanks to @ssk97  for the fix.
I agree that "true" should be replaced with EOL_SKIP_CHECK. Variable **checkEOL** used to be **bool**, but got converted to **int** during development. I wonder why compiler never warned about implicit conversion??

Anyway, looking at the code variable **checkEOL** is used to force EOL processing when Delimiter close statement "consumes" EOL. Typical example are line comments, EOL is consumed by line comment (or delimiter that is closed by ((EOL)) special character).

Of top of my head, I don't remember why was it necessary to skip EOL processing if Delimiter open and Folding keywords that are immediately followed by EOL char, but it obviously is missing for Opetaror1 stye. Thanks @ssk97 for finding that bug. First commit should be merged into main branch.

As for second commit, it is not good:

`
                         if (newState != SCE_USER_STYLE_COMMENTLINE || (sc.ch != '\r' && sc.ch != '\n'))
                             sc.Forward(iter->length());

\-                        if (sc.atLineStart)
\-                            checkEOL = EOL_FORCE_CHECK;
\+                        if (sc.atLineEnd)
\+                            checkEOL = EOL_SKIP_CHECK;
`

EOL at the end of line comments is not "skipped forward" because of potential nested line comment sequence. Therefore, we can't check **atLineEnd** but must instead rely on **atLineStart** which is safe even in case of multiple nested line comments.

ivan-radic:
Long story short: use first commit, discard second one. Some basic testing in pre-release phase by beta testers would be wise.

Best regards,
Ivan

Medal-x:
Still not work when using "SCE_USER_STYLE_COMMENTLINE" & "SCE_USER_STYLE_DELIMITER" (1-8) 
I added : 
`if (sc.atLineEnd) checkEOL = EOL_SKIP_CHECK;`
About everywhere after: 
`dontMove = true;`
in **LexUser.cxx**

Seem to work now!

Immueggpain:
Can we have this at 7.6.6 please?

donho:
Cherry pick `Fix folding in user-defined languages for non-windows line endings` but ignore `Fix delimiters as well as operators` as @ivan-radic  suggested.

Issue #774, opened 2015-08-21
Find Dialog have resize corner but it's actually not working (Win 8)


KvanTTT:
I am trying to resize this dialog but can not do it

![find dialog resize](https://cloud.githubusercontent.com/assets/1150330/9409506/c9eaa29c-4823-11e5-9240-0dc98fc65526.png)

Rikk:
It was always a false resize mark...

h-h-h-h:
It would be possible to remove the corner. But this feature would require it again to be resizable: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/563. So it's not a problem to make it resizable while the other feature not implemented.

dg9ngf:
I don't see how #563 is related to this issue.

IMO the resize grip simply needs to be removed. Should be a simple flag or option to set correctly to reflect the window's actual properties. #563 could add this grip again, but until then it's irritating.

dg9ngf:
Oh, and the browse for folder dialog is actually a pretty ancient one. Windows Vista introduced new file dialogs that are much easier to use. They can also be used to select folders now, replacing that ugly and unfriendly folder browse dialog from the older days. NPP already makes use of the latest file dialog version in open/save, so it should also do so in the find dialog.

h-h-h-h:
Because issue #563 contains a greater number of sub-issues, I meant it wouldn't hurt to make the dialog resizable separately. Let us see how the developer decides.

PR #3376, opened 2017-06-07
Ability to resize horizontally "Find in Files" dialog.

### LINKED ISSUE #2516: "Find in Files" dialog should be resizable
The Ctrl+F dialog "Find in Files" dialog isn't resizable, despite it has a resize handle at bottom right.

When I try to use this handle, the dialog doesn't change size (by the way, it produces a bit of "flashing", like some resizing was happening but immediately reset).

It would be great to be able to resize this dialog, possibly the current size becoming a "minimum size" instead.

Use case: I frequently use long directory name and it would be very helpful to see them entirely.

bahusoid:
Added ability to resize horizontally "Find in Files" dialog:
resizefindinfiles

It's very helpful (at least for me) when you dealing with long search strings or search directories.

I have zero experience with WinApi so comments/suggestions are welcome.

Closes #2516, #774


justanotheranonymoususer:
It would be very very nice if resizing vertically will show multiline find/replace boxes

vlakoff:
When resizing the dialog, texts in all input boxes become selected.

(it doesn't happen on the "Find" tab, the issue seems a bit different on the "Find what:" input)

bahusoid:
@vlakoff It's seems that strange behavior is "by design" in WinApi for combo boxes on resize. Text is selected only if it's present in combo box list items.  I'll see what can I do about it.

PR #3406, opened 2017-06-17
Fixed potential memory leak, found by Cppcheck.

orbitcowboy:
This PR fixes a potential memory leak found by the static code analysis tool [Cppcheck](http://cppcheck.sourceforge.net/devinfo/).

The potential memory leak happens, because new (by default) never returns a nullptr and therefore, the if-branch gets never executed. In order to catch memory allocation issues in C++, use try-catch or use (std::nothrow). 

Reference:
http://www.cplusplus.com/reference/new/operator%20new[]/


justanotheranonymoususer:
The checked pSearchFindAccel variable has nothing to do with `new`. As a side note, the allocation seems unnecessary, one can just use `_hFindAccTab = ::CreateAcceleratorTable(pSearchFindAccel, 1);`.

orbitcowboy:
I'll provide a new PR for this.

PR #3407, opened 2017-06-17
Remember empty "Replace with" on Notepad++ restart

bahusoid:
This PR allows to save to history empty values in Combo Boxes. It's useful for input fields where empty value is valid one and "Replace with" is one of them.

So I'm trying to fix the following condition:
emptyvaluenotsaved

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.h
@@ -401,7 +401,7 @@ private :
 	void updateCombo(int comboID);
 	void fillFindHistory();
     void fillComboHistory(int id, const std::vector<generic_string> & strings);
-	int saveComboHistory(int id, int maxcount, std::vector<generic_string> & strings);
+	int saveComboHistory(int id, int maxcount, std::vector<generic_string> & strings, bool saveEmpty);
 	static const int FR_OP_FIND = 1;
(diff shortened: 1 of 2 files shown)


vlakoff:
+1 on the principle. Not remembering the empty string surprised me many times.

PowerEditor/src/ScitillaComponent/FindReplaceDlg.h
```diff
@@ -401,7 +401,7 @@ private :
 	void updateCombo(int comboID);
 	void fillFindHistory();
     void fillComboHistory(int id, const std::vector<generic_string> & strings);
-	int saveComboHistory(int id, int maxcount, std::vector<generic_string> & strings);
+	int saveComboHistory(int id, int maxcount, std::vector<generic_string> & strings, bool saveEmpty = false);
```
donho: Why saveEmpty needs to be optional? And we should avoid paramter's default value - it can be a pitfall and make some bugs.

donho: OK, I see why it's optional. Please make filter (IDD_FINDINFILES_FILTERS_COMBO) the same behaviour (to true).
And remove the default value - let all calls explicit.

bahusoid:
@donho Done. I've made changes you requested. Let me know if you think anything else needs to be done.

PR #3413, opened 2017-06-19
update japanese.xml to v7.4.2

yasmise:
* Add new lines to catch up with v7.4.2 (Document peeker).


yasmise:
I've just noticed that "<< Find" button is introduced in Find/Replace dialog. So I've modified my commit to add that line.

PR #3416, opened 2017-06-19
Fixed text selection in combo boxes on FindReplace dialog resize

bahusoid:
Fixed problem reported by @vlakoff  here:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3376#issuecomment-309320220

If current text is present in combo box list item - it's selected on resize:
findinfilesallselected

Also while investigating this issue I noticed that some UI glitches are possible on rapid resizing. Adding SWP_NOCOPYBITS to SetWindowPos calls fixed it too.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -697,16 +697,28 @@ void FindReplaceDlg::resizeDialogElements(LONG newWidth)
 		IDREPLACE, IDREPLACEALL,IDC_REPLACE_OPENEDFILES, IDD_FINDINFILES_FIND_BUTTON, IDD_FINDINFILES_REPLACEINFILES, IDC_FINDPREV, IDOK, IDCANCEL,
 	};
 
-	_deltaWidth = newWidth - _initialClientWidth;
-	auto addWidth = _deltaWidth - _lastDeltaWidth;
-	_lastDeltaWidth = _deltaWidth;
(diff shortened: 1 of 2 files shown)


vlakoff:
Thanks for having a look at it :)

But with this patch, if some text had been selected before resizing, will this selection be discarded?

bahusoid:
@vlakoff  No. Active selection remains. Fix is the following:
Check before resizing if there is any active selection in combo box and clear selection after resizing only if there is no selection before resizing (endSelection == 0 - means no selection)

vlakoff:
So if I understood correctly, the issue would be fixed if there is no selection, but would still be present if there is a selection.

Still an improvement, as I suppose most of the time there is no selection.

I guess fixing all cases would require saving then restoring the selection? It would be probably worth it, though I can't say about coding difficulty, and performances.

But maybe someone is more familiar with WinApi and knows a way to fix this at the root?

bahusoid:
@vlakoff  I don't think you understand correctly. Logic I described is executed for each combo-box. Only partial selection is lost - as active edit box is fully selected by resize (it could also be fixed but I don't see importance of it - we lost any partial selection on any focus change):
resizeandselections

vlakoff:
That's what I meant: partial selection gets lost. It is clearly visible on your great gif.

Fixing this might help in some cases (e.g. cut/copy using right-clic menu, which doesn't take focus), but code complexity and performance should be considered, and as you said, many "focus change" scenarios already happen and alter the selection.

So, the situation should be good enough with this patch. I'm already happy this dialog is becoming resizable :)

PR #3440, opened 2017-06-24
Remove WS_EX_TOOLWINDOW style from dialog windows.

arnowelzel:
Using WS_EX_TOOLWINDOW does not make sense if the windows are dialog boxes. It is not needed at all.

In Windows 10 it is even worse: dialog windows using this style don't have a border (even if they are resizable) - so the "find" or "setting" dialogs for example will just be a plain white rectangle in front of a white text editor background - very confusing.

It is also confusing when some regular windows have smaller title bars due to the WS_EX_TOOLWINDOW style and others don't.


donho:
image

If you can remove also the upper left icon, without changing any current behaviour, I'll accept your PR.

arnowelzel:
Ok, it seems there was a confusion about the WS_SYSMENU style. If you just remove that as well, it will not appear any longer. Did you use WS_EX_TOOLWINDOW just to remove the system menu from the dialog windows?

arnowelzel:
PS: I can do this change as well and check the results. It will just take another day or so, as I'm a bit busy at the moment.

donho:
> Did you use WS_EX_TOOLWINDOW just to remove the system menu from the dialog windows?
Yes, this still is for removing the system icon but keeping the close button.
> PS: I can do this change as well and check the results. It will just take another day or so, as I'm a bit busy at the moment.
OK, then I close this PR. You can reopen it when you've done what I requested.

arnowelzel:
Ok, checked it. No - what you want is not possible.

If you want to have a close box, a windows must have WS_SYSMENU, because this IS the close box.

So I now understand, why you used the "toolbar window" style for all the dialog boxes, since you don't want to have the standard Windows GUI. But please reconsider this: even the Windows explorer uses the regular style for its dialog boxes - WS_SYSMENU but not WS_EX_TOOLWINDOW. So I would consider that the expected and normal behaviour: a normal dialog box has a system menu and a close box and is of course not a "toolbar window".

bahusoid:
@donho, @arnowelzel It's doable for dialogs by adding  DS_MODALFRAME instead of WS_EX_TOOLWINDOW.
For instance for preference.rc:

```C#
IDD_PREFERENCE_BOX DIALOGEX 0, 0, 560, 220
STYLE DS_SETFONT | DS_MODALFRAME | DS_FIXEDSYS | WS_POPUP | WS_CAPTION | WS_SYSMENU
EXSTYLE WS_EX_WINDOWEDGE
```
Will give us the following dialog (Win 10):
win10preferencesafter

and with go to dialog:

```C#
IDD_GOLINE DIALOGEX 26, 41, 261, 88
STYLE DS_SETFONT | DS_MODALFRAME | DS_FIXEDSYS | WS_POPUP | WS_CAPTION | WS_SYSMENU
EXSTYLE WS_EX_WINDOWEDGE
```

gotodialogframe

arnowelzel:
Well - there is a reason why the style is named "DS_MODALFRAME" - because it is for modal dialog windows and not for non-modal ones.

Nearly all dialog windows in Notepad++ are non-modal - you can just open them and continue to use the editor. A visible system menu icon is the visual indicator for this. Furthermore DS_MODALFRAME can not be combined with WS_THICKFRAME (which is an alias for WS_SIZEBOX) - but some dialog windows in Notepad++ use this to be resizable.

As an example: see the Explorer in Windows 10 and open the properties dialog for a selected file or folder. It will also have a system menu icon because it is also non-modal.

What's the problem with the system menu anyway? It is just part of Windows. So why not just use it as intended? If you don't like the icon you should fix the icon to be the right one and not try to use ugly "workarounds" to get rid of the standard UI.

bahusoid:
@arnowelzel Sorry but did you check your words before posting? I **do** make some basic checks before suggesting anything - resize and non-modal behavior are not affected in any way by DS_MODALFRAME (both cases were checked on FindReplace dialog). Where did you get it? The only thing I see in MSDN:

> Creates a dialog box with a modal dialog-box frame that can be combined with a title bar and window menu by specifying the WS_CAPTION and WS_SYSMENU styles.

Nothing about model dialog behavior - it's only about model dialog frame.

arnowelzel:
@bahusoid Yes, know exactly what I'm talking about. Maybe you should first read the official [UI style guide by Microsoft](https://msdn.microsoft.com/en-us/library/windows/desktop/dn688964(v=vs.85).aspx), which also refers to the already mentioned [document about dialog boxes](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742499(v=vs.85).aspx).

The resource editor of Visual Studio does not allow to combine DS_MODALFRAME and WS_THINKFRAME for good reasons: it makes no sense to give a Window a "modal" look if it is NOT modal and make it resizable as well. DS_MODALFRAME does not mean "hide system menu icon" but "show a window in a way, a modal dialog box looks like".

And please check existing dialog windows in applications by Microsoft - as I already mentioned the "properties" dialog for a file in the explorer in Windows 10 for example - and which styles they use. Maybe you will understand then.

So I would use DS_MODALFRAME **only** for dialog windows which **are** modal (which means you can not do anything else as long as the dialog is open) and not for every dialog just because it contains **standard elements** which one does not want to see.

But however the change will be in the end - using WS_EX_TOOLWINDOW is the worst of all and should be avoided! At the moment many dialog windows of Notepad++ really look awful in Windows 10.

bahusoid:
@arnowelzel Why are you referencing/suggesting UI style guidelines to me? The only thing I wrote is about actual implementation of behavior requested by donho. If it doesn't work  I'm interested to know why. The only helpful information from your post which explains at least what you meant by "DS_MODALFRAME can not be combined with WS_THICKFRAME":
> The resource editor of Visual Studio does not allow to combine DS_MODALFRAME and WS_THINKFRAME

arnowelzel:
@bahusoid I suggest to check the UI style guidelines so you understand why DS_MODALFRAME should not be combined with WS_THICKFRAME.

And about the resource editor: download a copy of Visual Studio (there is a free version as well) and try it yourself - Visual Studio will not let you create a dialog box resource with both styles set together in the visual resource editor.

PR #3447, opened 2017-06-26
Addition of multiple finders

eglibicky:
multiplefinders
## **It is a very useful feature when you need to search , as I do, many times every day in a existing code or text files and **switch** between the **results of your searches**. The results are presented at the bottom as search named tabs.**
This is an extension of finders, the **presentation of the  multiple search results of "Find" and "Find in Files" in the separate tabs**.
The finder tabs have names, which are the names of the searches.
In order to use this feature, the  **only one finder** option should not be selected in the   **Find and Find in Files dialog**.

Four options are added:

1. "**Only one finder**"Implements the old behavior if this option is selected, it creates and maintains only one finder and replace it every time when there are new "Find all" or "Find in Files" searches.
 The "Same name merge" and "Close empty" options are grayed out as not accessible, implementing an old behavior, except that you can still choose if you want to see the result line several times if the search string is found several times in the line.
**The default is false.**
onlyonefindermultiplefinders
2. **"Same name merge**" If selected when "Only one finder" is false, it replaces existing finder tab with a new result of search with the same name. If not selected, the results of the same search creates additional result tab, which can be useful if you search the same string in the different directories.
**The default is true.**
samenamemergemultiplefinders
Previous search result tab "shut" is close upon a new search and a new search result tab is open

3. "**Close empty**" closes all finders with 0 finds on the next search. 
**The default is true**
closeemptymultiplefinders

4. "**Show line once**" shows only one line of code in finder, if there are multiple finds in the same line. To show the result line once can be useful when you are writing a documentation and need to copy the results
**The default is true.**
manylinesmultiplefinders

In addition to that, a new “**Close all finders**” menu option is available in finder menu.

closeallmultiplefinders


Ede123:
I really like the idea as I'm often struggling with navigating in recent search results!

However there are some things I don't like:
* The options in the search dialog are hard to understand (it took me quite some time to figure it out even with the explanations) and probably not useful enough for most people. I suggest to have the "Only one finder" option in the preferences dialog and drop the other options in favor of reasonable defaults that should work for almost everybody:
  * Same name merge: Never merge searches unless they're identical (in that case update the last results)
  * Close empty: Do not even create a "finder" if there are no results - problem solved (maybe show a message box or print a message to the status bar of the find dialog in this case to inform the user)
  * Show line once (potentially worth a second PR): Improve the finder to support multiple search results per line. Ideally allow the user to jump to the correct occurence by clicking it (i.e. track where the user clicks). Working around this by dropping results feels wrong.

General note regarding wording (“Close all finders”, "Close this finder", "Find in this finder):
Finder is redundant here and introduces a concept unknown to the user (what's a"finder"?). It would be much clearer to simply use "Close all", "Close" and "Find in results". From the context the meaning should be obvious.

Additional feature request that might fit here:
I'm often missing context menu entries allowing to "Re-use search parameters" or "Update this search" since I find myself often doing the same search twice (e.g. after switching git branches or similar)

PR #2336, opened 2016-09-25
Scintilla/Boost build Update - Boostbuild.bat

MAPJe71:
- Improved argument detection and extended usage section accordingly.
- Corrected Boost build tool property for 64-bit builds i.e. using `address-model=64` instead of `architecture=ia64`. The `architecture=ia64`-property is intended for targets with an Itanium processor architecture which no one has (most probably). See also the '64-bit support'-section at http://www.boost.org/build/doc/html/bbv2/reference/tools.html
- Simplified Boost version detection i.e. get it directly from source, `getboostver.*`-files no longer needed.
- Replaced obsolete SourceForge forum link with Community forum link.
- Using for-loop to guess the used build tool set.
- Simplified library path assignment.
- Simplified `boostpath.mak`-file creation.

This PR replaces PR #2319.

PR #3456, opened 2017-06-29
Fix toolset flag parsing for BuildBoost.bat

### LINKED ISSUE #3455: BuildBoost.bat fails to set BOOSTPATH if toolset flag is used
### Description of the Issue

When you utilize the --toolset flag on the BuildBoost.bat, the batch script fails to set the BOOSTPATH. This causes the script to fail.

1. Follow the README steps up to "To Build SciLexer.dll with Boost"
2. Use the --toolset flag while trying to build boost e.g. BuildBoost.bat --toolset mvsc-12.0 C:\sources\boost_1_55_0 -x64

### Expected Behavior

The BuildBoost.bat script to utilize the toolset I selected and build boost. 

### Actual Behavior

The script echoed the USAGE block explaining how to format the command.

Gusgus01:
Fixes #3455

### DIFF
--- a/scintilla/boostregex/BuildBoost.bat
+++ b/scintilla/boostregex/BuildBoost.bat
@@ -15,18 +15,19 @@ IF [%1]==[] (
   GOTO PARAMCONTINUE
 )
 
-IF NOT [%1]==[--toolset] (
-     SET BOOSTPATH=%1
-)
-
 IF [%1]==[--toolset] (
   SET MSVCTOOLSET=%2
+  SHIFT
   SHIFT
 )
 
 IF [%2]==[-x64] (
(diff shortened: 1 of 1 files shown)


Gusgus01:
I noticed this while building from source, so I thought I would see if the changes I made were wanted. 

I saw some discussion on the community forums about more planned changes to the build process since the addition of the [certificate signing](https://notepad-plus-plus.org/community/topic/13457/how-do-i-run-npp-after-building-from-source), so I was hesitant to create this issue/pull request knowing that it might not be needed.

chcg:
See #2336 for a more intrusive, but also complete solution

Gusgus01:
Oh yep, I see that now. My search was not broad enough.

This can be closed!

PR #3463, opened 2017-07-01
Fixed PVS-Studio static analyser reported issues

SinghRajenM:
FindReplaceDlg.cpp: Release dynamically allocated memory.
BabyGrid.cpp: lpcs variable assigned twice successively.
ProjectPanel.cpp: A memory leak is possible.
TreeView.cpp: Wrong value is returned.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -1601,29 +1601,30 @@ bool FindReplaceDlg::processReplace(const TCHAR *txt2find, const TCHAR *txt2repl
 		// If the next find is the same as the last, then perform the replacement
 		if (nextFind.cpMin == currentSelection.cpMin && nextFind.cpMax == currentSelection.cpMax)
 		{
-			int stringSizeFind = lstrlen(txt2find);
-			int stringSizeReplace = lstrlen(txt2replace);
-
-			TCHAR *pTextFind = new TCHAR[stringSizeFind + 1];
-			TCHAR *pTextReplace = new TCHAR[stringSizeReplace + 1];
-			lstrcpy(pTextFind, txt2find);
-			lstrcpy(pTextReplace, txt2replace);
-		
 			bool isRegExp = replaceOptions._searchType == FindRegex;
 
 			int start = currentSelection.cpMin;
 			int replacedLen = 0;
 			if (isRegExp)
 			{
-				replacedLen = (*_ppEditView)->replaceTargetRegExMode(pTextReplace);
+				replacedLen = (*_ppEditView)->replaceTargetRegExMode(txt2replace);
 			}
 			else
 			{
 				if (replaceOptions._searchType == FindExtended)
 				{
-					Searching::convertExtendedToString(pTextReplace, pTextReplace, stringSizeReplace);
+					int stringSizeReplace = lstrlen(txt2replace);
+					TCHAR *pText2ReplaceExtended = new TCHAR[stringSizeReplace + 1];
+					Searching::convertExtendedToString(txt2replace, pText2ReplaceExtended, stringSizeReplace);
+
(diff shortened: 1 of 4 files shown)


PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -1644,6 +1644,17 @@ bool FindReplaceDlg::processReplace(const TCHAR *txt2find, const TCHAR *txt2repl
...
+			if (pTextFind)
+			{
+				delete[] pTextFind;
+				pTextFind = nullptr;
+			}
+			if (pTextReplace)
+			{
+				delete[] pTextReplace;
+				pTextReplace = nullptr;
+			}
```
bahusoid: What the point of pTextFind and pTextReplace local variables in the first place? It seems to me that they can be deleted completely - txt2find and txt2replace can be used directly instead. Am I wrong?

SinghRajenM: You're right about pTextFind. But pTextReplace is required because of statment
```C++
Searching::convertExtendedToString(pTextReplace, pTextReplace, stringSizeReplace);
```
However, I have updated PR for pTextFind.

bahusoid: Hmm. Wouldn't it be better to make this variable specific for "extended replace" branch of code.  So scope and purpose of it is much more clear. Something like:
```
if (isRegExp)
{
	replacedLen = (*_ppEditView)->replaceTargetRegExMode(txt2replace);
}
else
{
	if (replaceOptions._searchType == FindExtended)
	{
		int stringSizeReplace = lstrlen(txt2replace);
		TCHAR *txt2ReplaceExtended = new TCHAR[stringSizeReplace + 1];
		Searching::convertExtendedToString(txt2replace, txt2ReplaceExtended, stringSizeReplace);
 
		replacedLen = (*_ppEditView)->replaceTarget(txt2ReplaceExtended);
		
		delete[] txt2ReplaceExtended;
	}
	else
	{
		replacedLen = (*_ppEditView)->replaceTarget(txt2replace);
	}
}

```

donho:
@SinghRajenM  For "ProjectPanel.cpp", Could you point out where is th epossibility of a memory leak?

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -320,40 +320,63 @@ void ProjectPanel::destroyMenus()
...
+	bool bRetVal = false;
+	TiXmlDocument *pXmlDocProject = nullptr;
 
-	TiXmlNode *root = pXmlDocProject->FirstChild(TEXT("NotepadPlus"));
-	if (!root) 
-		return false;
+	try
+	{
+		pXmlDocProject = new TiXmlDocument(projectFileName);
+		if (!pXmlDocProject)
```
bahusoid: It looks like unnecessary check. `new` can't return null nowadays.

SinghRajenM:
@donho in original code are multiple return statements which may cause memory leak - 

```C++
	TiXmlDocument *pXmlDocProject = new TiXmlDocument(projectFileName);
	bool loadOkay = pXmlDocProject->LoadFile();
	if (!loadOkay)
		return false;

	TiXmlNode *root = pXmlDocProject->FirstChild(TEXT("NotepadPlus"));
	if (!root) 
		return false;

	TiXmlNode *childNode = root->FirstChildElement(TEXT("Project"));
	if (!childNode) 
		return false;

	if (!::PathFileExists(projectFileName))
		return false;
```

If something fails and return statement executes, then pXmlDocProject will not be deleted.

donho:
@SinghRajenM OK, I see.
I don't like to do a lot of modification which could bring regression for fixing an eventual memory leak.
Please just add `delete pXmlDocProject` before each return false.

SinghRajenM:
@donho changes have been made.

PR #3470, opened 2017-07-02
Do not treat strings with spaces as file extensions in the save dialog.

### LINKED ISSUE #3390: Save not adding filename extension
### Description of the Issue

Filename extension is not being added when a file is saved.

1. Open a new empty file.
2. Enter some text
3. Press the 'Save' icon.
4. Name the file, "save me!", Save as type: (default shown is Normal text file (*.txt)
5. Press Save button
6. Look for saved file in Windows file manager or use Save as to see file saved without extension.

### Expected Behavior

The extension .txt should be added to the filename.
### Actual Behavior

The extension .txt was not added when certain character(s) are the last character of a filename (e.g., '!').
Exclamation mark is the only character tested by accident.
Not always reproducible, went back to test again and it now adds the extension.

### Debug Information

Notepad++ v7.4.1   (32-bit)
Build time : May 18 2017 - 02:00:37
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (32-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll NppTextFX.dll PluginManager.dll SpellChecker.dll

irgendsontyp:
Fixes #3390.

A file name extension is now only considered valid when there is no space in the string following the last "." in the file name.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
@@ -481,7 +481,7 @@ generic_string changeExt(generic_string fn, generic_string ext, bool forceReplac
 	auto index = fnExt.find_last_of(TEXT("."));
 	generic_string extension = TEXT(".");
 	extension += ext;
-	if (index == generic_string::npos)
+	if (index == generic_string::npos || fnExt.find_first_of(' ', index + 1) != generic_string::npos)
 	{
 		fnExt += extension;
 	}


bahusoid:
I don't think it's a proper fix. What if it's some other common separator like "_" or "-". Or it's simply the end of file name like "ver2.1.txt". To me it looks like we don't need to guess extension at all if it's explicitly specified by user. So this auto guess extension should be used only when "Save as type"  equals "All types (*.*)".

irgendsontyp:
> What if it's some other common separator like "_" or "-". Or it's simply the end of file name like "ver2.1.txt".

Windows seems to treat everything except spaces after the last "." as file extension. You can check by creating a file in Windows Explorer and see the "Type" column. For example: I created a file named "foo._-" and the "Type" column says "_--File".

ver2.1 would indeed be a problem as Notepad++ takes "1" as the file extension. (That's what it already does, this was not introduced by my fix. Unfortunatelly there is no way to perfectly determine the file extension part of a file name because you never know whether there is one at all.

> To me it looks like we don't need to guess extension at all if it's explicitly specified by user. So this auto guess extension should be used only when "Save as type" equals "All types (*.*)".

The extension is guessed for example when the user provides a file name without extension or when the user changes the file type in the drop down menu of Windows' save file dialog. I guess this is done because you want to notify the user about the case when a file already exists. To check for that condition, you have to know the whole file name before saving. So you have to know the extension before the dialog closes.

donho:
Cannot reproduce #3390.
It seems it has been fixed.

sasumner:
Closure by @donho seems correct, as the conditions this PR addresses -- the "with spaces" part -- doesn't seem to cause any problem when tried in v.7.8.5.

Issue #3435, opened 2017-06-23
Changed find and replace dialog in v7.4.2: lost option to replace backwards


dinkumoil:
### Description of the Issue
The redesign of the find and replace dialog in v7.4.2 resulted in losing the option to replace words in direction backwards from the current cursor position.

### Expected Behavior
It should be possible to replace words backwards from the current cursor position as it was before the redesign of the find and replace dialog.

### Actual Behavior
The redesign of the find and replace dialog in v7.4.2 removed the radio buttons for search direction preference and replaced the "Find Next" button with two buttons for searching backwards and forwards. The "Replace" button at the "Replace" tab of the dialog leaved untouched. So we lost the option to **replace** backwards.

### Suggestion
Because there is already issue #3415 (the new "Find previous" and "Find next" buttons are too small to hold the whole text for some languages) I suggest to revert all changes instead of splitting the "Replace" button in two buttons too. It would be simply too much buttons placed close to another.

dinkumoil:
I noticed that the two new buttons at the "Find" tab also control the direction of the "Replace" function. OK, so we didn't lose the option to replace backwards. But it is still very inconvenient to switch to the "Find" tab only to set the direction of replacement.

So, #3471 does **not** really fix **this** issue. To get a Find/Replace dialog as comfortable as it was before v7.4.2 can only be achieved by introducing the posibility to set the replacement direction at the "Replace" tab.

dinkumoil:
Seems to be reverted. See 4d921b7

Issue #3415, opened 2017-06-19
Changed window find and replace in v7.4.2: Line break in german language


JackyRyan:
### Description of the Issue
In german language there is a poor line break in "find and replace" dialog (Weitersuche-n):  

![1](https://user-images.githubusercontent.com/29531994/27271478-453476a8-54c5-11e7-8aa5-7bf112754cab.PNG)

### Steps to Reproduce the Issue

1. Set language to german
2. Open find and replace dialog ctrl-f
3. See poor linebreak on right upper corner

### Expected Behavior
No line break

### Actual Behavior
Poor line break

### Debug Information
n/a
### In addition one further question:
Are the 2 arrows "<<" on find button your purpose? (I'd rather leave them...)

AndNovAtor:
I have similar behavior with Russian language (and "<<Find" button is also not translated). With English (text "Find next" ) this button is displayed normally.
It may be necessary to increase the width of the right column (column of "find" buttons).

JackyRyan:
OK, now I understand
"<< Find" is searching backward
"Find next" searching forward

![2](https://user-images.githubusercontent.com/29531994/27376160-9cf72f7c-5671-11e7-88f6-a5552109d830.PNG)

Why don't use a radio button to change between forward / backward? Why use arrows on search backward and "no arrows" on forward (don't like "Find >>"?)? Furthermore I'm used to click on first button after textbox to search and wondered why it only searches backward - but now I understand that there are 2 buttons forward / backward (other programs only have one button and use a radio box to change it).

Please update german text and button width:
"<< Suchen"
"Weitersuchen" + extend width (or just write "Suchen" and leave the width)

AndNovAtor:
Radio buttons was removed in THIS update.... https://notepad-plus-plus.org/download/v7.4.2.html

JackyRyan:
I only see:
3. Enhance Find Replace dialog (resizable & remove search direction radio buttons)

But what about changing the text and the arrows etc.? Was there any discussion?

AndNovAtor:
Hmm. For example: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3414
So, it will be fixed.

AndNovAtor:
Hm, i'm wrong. It is "custom" pull request. So, it is still bug.

Ilis:
The same in Russian

[...]

PR #3471, opened 2017-07-02
Update german.xml to v7.4.2 with newest fixes

schnurlos:
Following https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fb998007d7eaff074d990ff528026f94f21aeafc this closes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3415 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3435


dinkumoil:
This PR does not really fix issue #3435. To get a Find/Replace dialog as comfortable as it was before v7.4.2 can only be achieved by introducing the posibility to set the replacement direction at the "Replace" tab.

PR #3472, opened 2017-07-02
Fixed some coding mistake

SinghRajenM:
NppIO: It looks like typo
Parameter.h: Not all members were initialized in ctor and moved all initialization in single method
ScintillaEditView.cpp: redundant check of pointer
FileBrowser.cpp: unnecessary variable and compression
nsSBCharSetProber.h: Marked assignment operator overloading as deleted otherwise it should return something

### DIFF
--- a/PowerEditor/src/uchardet/nsSBCharSetProber.h
+++ b/PowerEditor/src/uchardet/nsSBCharSetProber.h
@@ -84,7 +84,7 @@ public:
   // prober has a hard-coded call to FilterWithoutEnglishLetters which gets rid
   // of the English letters.
   PRBool KeepEnglishLetters() {return mModel->keepEnglishLetter;} // (not implemented)
-  nsSingleByteCharSetProber operator=(const nsSingleByteCharSetProber&){};
+  nsSingleByteCharSetProber operator=(const nsSingleByteCharSetProber&) = delete;
 
 #ifdef DEBUG_chardet
   virtual void  DumpStatus();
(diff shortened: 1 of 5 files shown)


PowerEditor/src/uchardet/nsSBCharSetProber.h
```diff
@@ -84,7 +84,7 @@ class nsSingleByteCharSetProber : public nsCharSetProber{
   // prober has a hard-coded call to FilterWithoutEnglishLetters which gets rid
   // of the English letters.
   PRBool KeepEnglishLetters() {return mModel->keepEnglishLetter;} // (not implemented)
-  nsSingleByteCharSetProber operator=(const nsSingleByteCharSetProber&){};
+  nsSingleByteCharSetProber operator=(const nsSingleByteCharSetProber&) = delete;
```
donho: Could you provide more explanation for this change?

SinghRajenM: This is not proper assignment operator overloading. It should perform deep copy and should return ```*this```. 
Previous code (assignment operator overloading) does nothing. So instead of doing nothing, it should not be permitted at all.
There two ways to do it. 1. put = overloading in private section (similar to singleton class). Modern C++ offer another way mark it deleted (which is far better as compiler produce error when it see the usage of = in the code. Refer [msdn](https://msdn.microsoft.com/en-us/library/dn457344.aspx).).

PR #3473, opened 2017-07-03
Update Swedish translation to 7.4.2

magol:


PowerEditor/installer/nativeLang/swedish.xml
```diff
@@ -305,7 +306,8 @@
 					<Item id="50006" name="Sökvägskomplettering"/>
 				</Commands>
 			</Main>
-			<Splitter />
+            <Splitter>
+            </Splitter>
```
donho: From <Splitter /> to
<Splitter>
</Splitter>

???

magol: I saw that it was like so in the English base text. But that may not matter?

donho: I don't like unnecessary changes, also there are a lot of reformat (again).

magol: @donho I know that you don't like to make formatering changers, and I absolutely understand the motivation. I agree that it is not good to mix regular changes with formatering changes, as it is hard to see the changes that matter. 

The problem is that the quality of the code slowly deteriorates if it is not allowed to improve the formatting.

I agree that it is not wise to make big changes at once (which I stupidly tried to do), but it must be allowed to make small improvements in order to at least improve the quality.

I did not think it was so controversial to at least fix a few simple line feeds, but I restore them now.

The reason I want to do this is because it is very difficult to compare the Swedish translation with the English source. It had make it much simpler if I was allowed to do the translation equivalent to the English original.

donho: @magol OK, please do a major change in this PR, then afterward only the necessary modification afterward.

magol: @donho
I appreciate that you understand me and am willing to make exceptions to help me. I understand I've annoyed you, and it's absolutely anything I want. It's good that you keep your principles and have high ambitions.

Unfortunately, I've been on vacation, and I saw you already merged my PR.
Is it ok if I create a new PR that only sorts the strings and makes the file similar to the English source file? Then I promise not to annoy you anymore, and stick to the rules.

Issue #3474, opened 2017-07-03
German search dialog next button translation fails to be consistent


ophian:
NPP 7.4.2
See attached file for behaviour

![npp-7 4 2-search-dialog-de](https://user-images.githubusercontent.com/1105372/27784532-b62db82c-5fda-11e7-82c7-0b03748e8f15.png)

I think <<Finde | Finde>>, or <<Suche | Suche>>, or even untranslated would be just fine.
Thanks

arctic-blue:
>> fwd    &   << rwd    should be international. no need to translate it

schnurlos:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3471

dinkumoil:
Seems to be reverted. See 4d921b7

PR #3486, opened 2017-07-04
lang: Cleanup strange "&amp;" for both [en] [de] and finally fix "sea…

ophian:
…rch" for the latter

including Pull Request
 https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3471
 https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3471/commits/4e949c3bc04e19944ef7f19ad78d44130b1a6490

fixes #3471 and #3474 and #3415 and could be used as a "template" for other langs.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -421,7 +421,7 @@
 
             <StyleConfig title="Style Configurator">
                 <Item id="2"    name="Cancel"/>
-                <Item id="2301" name="Save && Close"/>
+                <Item id="2301" name="Save & Close"/>
                 <Item id="2303" name="Transparency"/>
                 <Item id="2306" name="Select theme: "/>
                 <SubDialog>
@@ -497,7 +497,7 @@
                     <Item id="1" name="OK"/>
                     <Item id="2" name="Cancel"/>
                 </StylerDialog>
-                <Folder title="Folder && Default">
+                <Folder title="Folder & Default">
                     <Item id="21101" name="Default style"/>
                     <Item id="21102" name="Styler"/>
                     <Item id="21105" name="Documentation:"/>
@@ -545,7 +545,7 @@
                     <Item id="22572" name="Styler"/>
                     <Item id="22622" name="Styler"/>
                 </Keywords>
-                <Comment title="Comment && Number">
+                <Comment title="Comment & Number">
                     <Item id="23003" name="Line comment position"/>
                     <Item id="23004" name="Allow anywhere"/>
                     <Item id="23005" name="Force at beginning of line"/>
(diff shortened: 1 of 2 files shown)


ophian:
added some more cleanup

ophian:
@schnurlos Hi
> Just a question: do you know what & means?

Yes, `&amp;` is the url or html entity for the ampersand `&`.
This has nothing to do in those language files and probably is some sort of wrong inserted editor stuff.

> And for the german ideas: "Suche vorwärts" and "Suche rückwärts" is much to long to fit in the field.

I don't have that. See all my Pull Request commits. It looks like this.

npp-7 4 2-search-dialog-de2

And I have checked all my other changes working good.
I'd like to see that included on next update.

ophian:
Just for the info,
the german.xml holds a bunch of translated vars, which are not in the english.xml (any more).
I think we should remove them, if not necessary.
But there are some, which make me wonder. These are subjected to IDs 44011 ... 46017, and some of them related to the character set encoding, like chinese Big5.

See https://github.com/ophian/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/german.xml#L284 ff

@donho 
Was that really removed in core?

bahusoid:
> Yes, `&amp;` is the url or html entity for the ampersand &.
> This has nothing to do in those language files and probably is some sort of wrong inserted editor stuff.

It's "[ampersand mnemonics](https://msdn.microsoft.com/en-us/library/190kw3at.aspx)". I believe you've just killed "Alt+[Key]" menu navigation in your languages (maybe only removed hints - access key underline when Alt key is pressed). In any case - if it's strange for you it doesn't mean it's useless.

ophian:
@bahusoid 
Uh yeah, thanks. Didn't know that. 

You mean things like: `ALT+0` and the reverse `SHIFT+ALT+0` ? Or: `ALT+h` hiding lines?
They still work and do what they should!

Now, what does that mean to us?
```diff
-                    <Item menuId="file" name="&amp;File"/>
-                    <Item menuId="edit" name="&amp;Edit"/>
-                    <Item menuId="search" name="&amp;Search"/>
+                    <Item menuId="file" name="File"/>
+                    <Item menuId="edit" name="Edit"/>
+                    <Item menuId="search" name="Search"/>
```
In the linked page it says:
> In the Properties Window, in the Caption property, type a new name for the control, typing an ampersand (&) in front of the letter you want as the access key for that control. For example, &Radio1.

Just checked it. Its the underline caption thing in the bar(s), activated by `ALT`. Both `&` and `&amp;` approaches work.
Checking translated files with the latter is not that nice, so I would pled to use the `&` generally.
What do you think?

bahusoid:
You should stick to how it was. It's XML specifics - you must encode & as  `&amp;` in xml. Otherwise it won't be well formatted XML.

donho:
please redo your PR... if it's really needed : 

PR #3488, opened 2017-07-04
Fixed problems found by VS 2017

bahusoid:
Common.cpp and pluginsAdmin.cpp: unicode string is treated as non-unicode
shortcut.cpp: signed and unsigned comparison

Aside from fixing potential problems it's also a precursor for making build under VS 2017 with built-in v141_xp  Platform Toolset. See attached log with compilation errors: 
[VS2017CompilationErros.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1123953/VS2017CompilationErros.txt)

### DIFF
--- a/PowerEditor/src/WinControls/shortcut/shortcut.cpp
+++ b/PowerEditor/src/WinControls/shortcut/shortcut.cpp
@@ -26,7 +26,6 @@
 // Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
 
-#include <algorithm>
 #include "shortcut.h"
 #include "Parameters.h"
 #include "ScintillaEditView.h"
@@ -483,7 +482,7 @@ INT_PTR CALLBACK Shortcut::run_dlgProc(UINT Message, WPARAM wParam, LPARAM)
 // return true if one of CommandShortcuts is deleted. Otherwise false.
 void Accelerator::updateShortcuts() 
 {
-	vector<int> incrFindAccIds;
+	vector<UINT> incrFindAccIds;
 	incrFindAccIds.push_back(IDM_SEARCH_FINDNEXT);
 	incrFindAccIds.push_back(IDM_SEARCH_FINDPREV);
 	incrFindAccIds.push_back(IDM_SEARCH_FINDINCREMENT);
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -765,7 +765,7 @@ COLORREF getCtrlBgColor(HWND hWnd)
 
 generic_string stringToUpper(generic_string strToConvert)
 {
-    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), ::toupper);
+    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), ::_totupper);
     return strToConvert;
 }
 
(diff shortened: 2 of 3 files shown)


PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -483,7 +482,7 @@ INT_PTR CALLBACK Shortcut::run_dlgProc(UINT Message, WPARAM wParam, LPARAM)
 // return true if one of CommandShortcuts is deleted. Otherwise false.
 void Accelerator::updateShortcuts() 
 {
-	vector<int> incrFindAccIds;
+	vector<UINT> incrFindAccIds;
```
bahusoid: For signed and unsigned comparison go to line 520 below:
```C#
// Special extra handling for shortcuts shared by Incremental Find dialog
if (std::find(incrFindAccIds.begin(), incrFindAccIds.end(), shortcuts[i].getID()) != incrFindAccIds.end())
```
shortcuts[i].getID() is unsigned int and  incrFindAccIds was int

SinghRajenM: shortcuts[i].getID() is ```unsigned long``` instead of ```unsigned int```. So ``` ULONG``` will be better ```UINT```.

bahusoid: Oh indeed. Thanks. I corrected my previous comment.  But still I think it's safe and better to use minimal required unsigned type here. Why to increase memory usage impact? )

SinghRajenM: > Why to increase memory usage impact?

Then it will end up with unnecessary type casting.

bahusoid: Could you please clarify implications of it?
With your suggested fix type casting is also needed (UINT -> ULONG for all int values that added to incrFindAccIds) plus more memory will be used in the end.

I write small code snippet to see how type conversion look and how it affects assembly code :
```C++
	unsigned int s1 = 1, s2 = 2;
00174BAE  mov         dword ptr [s1],1  
00174BB5  mov         dword ptr [s2],2  
	unsigned long l1 = 1000000000l;
00174BBC  mov         dword ptr [l1],3B9ACA00h  

	if (s1 == s2) //UINT AND UINT
00174BC3  mov         eax,dword ptr [s1]  
00174BC6  cmp         eax,dword ptr [s2]  
00174BC9  jne         main+52h (0174BE2h)  
//...
	if (s1 == l1) //UINT AND ULONG
00174BE2  mov         eax,dword ptr [s1]  
00174BE5  cmp         eax,dword ptr [l1]  
00174BE8  jne         main+71h (0174C01h)  
//...

	if (l1 == s1) //ULONG AND UINT
00174C01  mov         eax,dword ptr [l1]  
00174C04  cmp         eax,dword ptr [s1]  
00174C07  jne         main+90h (0174C20h)  
//...	


	if (l1 == s1) // first bigger type
00174C01  mov         eax,dword ptr [l1]  
00174C04  cmp         eax,dword ptr [s1]  
00174C07  jne         main+90h (0174C20h)  
//...	

```
I see no difference whatsoever - so it looks to me we shouldn't bother about safe implicit built-in type conversions.

donho:
I have no problem to compile under VS2017

bahusoid:
@donho  Did you try to compile with latest VS 2017 platform toolset  - v141_xp?

bahusoid:
platformtoolset

PR #3492, opened 2017-07-06
Try to make universal Visual Studio 2013-2017 project

bahusoid:
I think it's much easier to support single Visual Studio project instead of multiple VS  version dedicated projects.

That PR I believe makes notepadPlus.vs2015.vcxproj not needed (required changes were combined). To work with VS 2017 #3488 needs to be merged.

It works for me with VS 2015 and VS 2017. I don't have VS 2013 but it works with AppVeyor build based on VS 2013 image.

Summary of changes:
1) Added calculated NppVsPlatformToolset property. Needs to be updated for each new  VS version with proper PlatformToolset:
```xml
<!--VS 2017 -->
<NppVsPlatformToolset Condition="$(VisualStudioVersion) == '15.0'">v141_xp</NppVsPlatformToolset>
```
2) Added  VS 2015+ specifics from notepadPlus.vs2015.vcxproj  (suppress some warnings and added _CRT_NON_CONFORMING_WCSTOK preprocessor definitions);
3) Added some missing files;

Let me know what you think.

### DIFF
--- a/PowerEditor/visual.net/notepadPlus.vcxproj
+++ b/PowerEditor/visual.net/notepadPlus.vcxproj
@@ -1,5 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="NppSpecifics">
+    <!-- VS 2012 build tools don't support  C++x used in code, so no joy here-->
+    <!-- <NppVsPlatformToolset Condition="$(VisualStudioVersion) == '11.0'">v110_xp</NppVsPlatformToolset> -->
+    <!--VS 2013 -->
+    <NppVsPlatformToolset Condition="$(VisualStudioVersion) == '12.0'">v120_xp</NppVsPlatformToolset>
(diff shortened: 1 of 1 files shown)


linquize:
Nice idea!

mancoast:
Can we modify appveyor.yml to build using VS2017?

chcg:
See https://www.appveyor.com/docs/build-environment/#build-worker-images, currently used image is Visual Studio 2013. Visual Studio 2015 is the standard for current projects and Visual Studio 2017 is slightly different as just VS2015 and VS2017 are supported.

See e.g. https://github.com/chcg/notepad-plus-plus/blob/mingw_gcc/appveyor.yml for a config which added VS2015. VS2017 is a little bit different as vcvarsall.bat is not placed under C:\Program Files (x86)\Microsoft Visual Studio 15.0\VC\, see https://github.com/chcg/NppToolBucket/blob/master/appveyor.yml

donho:
VS2013 will be outdated. We are using VS2015 & VS2017 with the old vcxproj currently and it works fine.

bahusoid:
@donho  Did you try you solution with only VS 2017 installed? It won't work as  v140_xp platform toolset is not available. You solution require to install VS 2015 toolset.

PR #3494, opened 2017-07-07
Multiple Search Result Tabs

eglibicky:
##  It is a very useful feature when you need to search , as I do, many times every day in a existing code or text files and switch between the results of your searches. The results are presented at the bottom as search named tabs.
presentmultiplefinders

This is a the presentation of the multiple search results of "Find" and "Find in Files" in the separate tabs at the bottom.
The tabs have names, which are the names of the searches.
In order to use this feature, the only one finder option should not be selected in the Find and Find in Files dialog.
In order to activate a new feature, select "**Multiple Result Tabs**" in the Preference "Search Option".
If this option is **not selected**, the PR is not active and there is an **old behavior** with only one search result.
preferencesmultiplefinders


eglibicky:
Thank you for your input. I simplified this PR and it has much clear interface.

donho:
Quick play around, I've found 2 regression regarding "Find All in all opened documents" & "Find All in current document"
The option doesn't belong to "Search Engine Setting" clearly.

eglibicky:
The "Search Engine" was renamed to "Search Options". The user can easily find any search setting.
The long tab collapses into small one when it is not selected.

donho:
I prefer to keep the current way for this behaviour. Thank you for your PR anyway.

PR #3495, opened 2017-07-07
Make a copy of the command line so as to stop modifying the original

jay:
.. because other processes may want to read the command line that
started Notepad++. Prior to this change the original command line could
be seen by other processes as truncated due to parsing behavior.

Closes #xxxx

---

For example this command line:
"c:\program files (x86)\notepad++\notepad++.exe" foo bar\0

After Notepad++ parses it will change it to:
"c:\program files (x86)\notepad++\notepad++.exe" foo\0bar\0

Process Hacker reads the 64-bit PEB and will show:
"c:\program files (x86)\notepad++\notepad++.exe" foo bar

... but Process Explorer reads the 32-bit PEB and will show:
"c:\program files (x86)\notepad++\notepad++.exe" foo

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -315,7 +315,10 @@ void doException(Notepad_plus_Window & notepad_plus_plus)
 
 int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int)
 {
-	LPTSTR cmdLine = ::GetCommandLine();
+	// Use a copy of the command line in order to preserve the original to be read by other processes.
+	LPTSTR cmdLine = _tcsdup(::GetCommandLine());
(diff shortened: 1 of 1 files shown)


donho:
Which process would read the command line?
Any concrete example?

jay:
I'm writing a launcher for notepad++ (just wraps ExecInExplorer from Microsoft) because of the dll signature verification errors I get with a hard link. I was trying to diagnose why the command line was different from what I passed but then I realized it's just notepad++ modifying the command line passed to it. So one concrete example: sysinternals process explorer.

donho:
I see.
However, it misses a free of strdup has allocated.
And I prefer to do the allocation near where it needs.
I've done a fix. Thank you for your PR anyway.

PR #3499, opened 2017-07-07
Fixes #3246

### LINKED ISSUE #3246: Enable save button after underlying file is changed externally
### Description of the Issue
When a file is in the editor and it is changed by an external program, npp recognizes this and asks if you want to reload the file.  If you say no, it should enable the save button since the file is different than the one which is on the disk and is "dirty".

1.  Open a file
1a  Change the file
1b  Save the file 
1c Note that save button is disabled since file is not dirty
2.  Change the file with another editor
3.  Answer no when asked if you want to reload it

### Expected Behavior
The save button should be enabled since file has changes which can be saved

### Actual Behavior
Save button is not enabled and I must make and undo a change in order to save to disk

irgendsontyp:
Fixes #3246.

Set buffer state to dirty when the file on the file system has changed but the user decided to not reload it in Notepad++.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5012,7 +5012,7 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
 					// Then we ask user to update
 					if (doReloadOrNot(buffer->getFullPathName(), buffer->isDirty()) != IDYES)
 					{
-						// Since the file has but the user doesn't want to reload it, set state to dirty
+						// Since the file content has changed but the user doesn't want to reload it, set state to dirty
 						buffer->setDirty(true);
 
 						break;	//abort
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -5011,7 +5011,12 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
 
 					// Then we ask user to update
 					if (doReloadOrNot(buffer->getFullPathName(), buffer->isDirty()) != IDYES)
+					{
+						// Since the file has but the user doesn't want to reload it, set state to dirty
```
justanotheranonymoususer: "has changed"?

irgendsontyp: Thanks, I updated the comment! Guess I was not sure whether to say "the file has changed", "the file content has changed" or "the file has changed in the file system" or something else, so I might have decided to just omit the relevant parts of the sentence :)

irgendsontyp:
I'm curious when this PR (and my other two ones) will get a review. I'm kinda new to GitHub and those were my first three pull requests at all, so I do not know how much time it takes normally until a PR gets reviewed/merged.

Does someone know?

AngryGamer:
@irgendsontyp it can take a while, @donho generally reviews the PRs all at once.

vlakoff:
The kind of small change that greatly improves the workflow. Thanks :)

Issue #3384, opened 2017-06-10
Double FileTab problem (Bug)


alex77g2:
### Description of the Issue
Same file can be open in different (multiple) tabs. Which is usually prevented by NPP-editor.

### Steps to Reproduce the Issue
1. One tab has a file (e.g. D:\test.txt) open.
2. File+New gives new Tab (new 1).
3. Add test text "foo".
4. Save this new tab as D:\test.txt (same as above).
5. You see 2 different tabs having same file open and they could have different content.

### Expected Behavior
If NPP detects save-as-file identical to already open one, it should replace the existing tab. // false
Opening same file via open does only jump to existing one (not creating new tab). // correct

### Actual Behavior
Now it reopens another version with same tab-identifier-text.

### Debug Information
Notepad++ v7.4.1   (32-bit)
Build time : May 18 2017 - 02:00:37
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll SpellChecker.dll

DADMahindika:
I would like to work on this bug. Can someone please help me, and can I know where do I start this.

AngryGamer:
There's already some logic to prevent this problem. One would simply need to figure out why it isn't working.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppIO.cpp#L1331

DADMahindika:
But, even if we save the file with same name (also considering case sensitivity), still have this problem.

PR #3500, opened 2017-07-07
Fixed issue #3384 [same file in mutliple tab]

SinghRajenM:
Fixed issue #3384

Current code checks for opened file existence only in the other view (either in sub view or in main view), but does not check in same view.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1328,8 +1328,11 @@ bool Notepad_plus::fileSaveAs(BufferID id, bool isSaveCopy)
 
 	if (pfn)
 	{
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -1328,8 +1328,9 @@ bool Notepad_plus::fileSaveAs(BufferID id, bool isSaveCopy)
 
 	if (pfn)
 	{
+		BufferID same = _pDocTab->findBufferByName(pfn);
```
bahusoid: It seems second search is not really needed if first search is successful. So I think it's enough to have just one variable (maybe renamed to sameFile) and just do second search only if first search is failed.

SinghRajenM: Done!!!

PR #3506, opened 2017-07-09
Fix wrong File encoding auto-detection problem : binary file is detected as UTF-16 little-endian

### LINKED ISSUE #2692: Incorrectly file show binary

1.  Open file bad.txt
2.  File size must be 4096 but show 5875
3.  Incorrectly information show least characters must be 0xFF, but show 0xEF, or 0xBF ....
bad

### Actual Behavior

[bad.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/651024/bad.txt)

### Debug Information

Notepad++ v7.2.2   (32-bit)
Build time : Nov 27 2016 - 16:31:07
Admin mode : ON
Local Conf mode : OFF
OS : Windows 10
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

Cyrillev91:
Fix #2692

### DIFF
--- a/PowerEditor/src/Utf8_16.cpp
+++ b/PowerEditor/src/Utf8_16.cpp
@@ -181,7 +181,7 @@ size_t Utf8_16_Read::convert(char* buf, size_t len)
 
 void Utf8_16_Read::determineEncoding()
 {
-	INT uniTest = IS_TEXT_UNICODE_STATISTICS;
+	// INT uniTest = IS_TEXT_UNICODE_STATISTICS;
 	m_eEncoding = uni8Bit;
 	m_nSkip = 0;
 
@@ -204,12 +204,14 @@ void Utf8_16_Read::determineEncoding()
 		m_eEncoding = uniUTF8;
 		m_nSkip = 3;
 	}
+    /* Fix wrong File encoding auto-detection problem : binary file is detected as UTF-16 little-endian
 	// try to detect UTF-16 little-endian without BOM
 	else if (m_nLen > 1 && m_nLen % 2 == 0 && m_pBuf[0] != NULL && m_pBuf[1] == NULL && IsTextUnicode(m_pBuf, static_cast<int32_t>(m_nLen), &uniTest))
 	{
 		m_eEncoding = uni16LE_NoBOM;
 		m_nSkip = 0;
 	}
+    */
 	/* UTF-16 big-endian without BOM detection is taken away scince this detection is very week
     // try to detect UTF-16 big-endian without BOM
     else if (m_nLen > 1 && m_pBuf[0] == NULL && m_pBuf[1] != NULL)


donho:
@Cyrillev91 Could you provide the reason to comment these lines? (The reason that it fix the issue but won't have any regression)

Cyrillev91:
Bonjour @donho en fait cela est assez complexe 

Normalement un fichier encodé en UTF-16 little-endian commence par 0xFF 0xFE (appelé BOM) mais dans le code mis en commentaire on essai d'identifier un fichier encodé en UTF-16 little-endian même s'il ne contient pas de BOM.
2 cas sont possibles :
      - si un fichier hexa est identifié à tord comme UTF-16 LE without BOM alors il sera mal affiché (comme on peut le voir dans #2692)
        Remarque : si le fichier de #2692 est ouvert dans scintilla il est affiché correctement.
      - si le fichier est un vrai fichier "UTF-16 without BOM" il sera plus affiché tel que...

Il y a effectivement un choix à faire.

En parcourant l'historique du code on peut aussi voir que ce pb est apparue pour "try to detect UTF-16 big-endian without BOM" ==> et le code a été mis en commentaire.

Peut-être qu'il faudrait modifier le code et proposer 2 options désactivables :
     - try to detect UTF-16 big-endian without BOM
     - try to detect UTF-16 big-little without BOM

justanotheranonymoususer:
Please don't remove this check, it's useful. It can be improved, though. Notepad has a similar auto-detection mechanism, and it doesn't detect the binary file as Unicode. Perhaps Notepad++ can learn from it.

Cyrillev91:
Finalement il vaut mieux trouver une autre solution pour résoudre ce problème #2692 et conserver le mécanisme actuelle de détection de UTF-16 big-little without BOM

MAPJe71:
Google translate:
> Hello @donho in fact this is quite complex
>
>Normally a little-endian UTF-16 encoded file starts with 0xFF 0xFE (called BOM) but in the commented code we try to identify a little-endian UTF-16 encoded file even if it does not contain a BOM .
2 cases are possible:
>- if a hexa file is identified to twist as UTF-16 LE without BOM then it will be displayed incorrectly (as can be seen in # 2692)
>Note: If the # 2692 file is opened in scintilla it is displayed correctly.
>- if the file is a real file "UTF-16 without BOM" it will no longer be displayed such as ...
>
>There is indeed a choice to make.
>
>It is also possible to see that this pb appeared for "try to detect UTF-16 big-endian without BOM" ==> and the code has been commented out.
>
>Perhaps it would be necessary to modify the code and to propose 2 options désactivables:
>- try to detect UTF-16 big-endian without BOM
>- try to detect UTF-16 big-little without BOM

and

> Finally it is better to find another solution to solve this problem # 2692 and keep the current mechanism of detection of UTF-16 big-little without BOM

Issue #3545, opened 2017-07-20
Multi-line tab appears pressed even though it isn't the active tab


jay:
### Description of the Issue
Occasionally I see a tab appear to be pressed even though it is not the active tab. I don't have steps to reproduce. I can build Notepad++ on this computer so let me know if there's anything I can do to help gather more details the next time it happens.

### Expected Behavior
For only the selected tab to appear pressed.

### Actual Behavior
Screenshot:
![tab not popping up 742](https://user-images.githubusercontent.com/965580/28402283-04b7c1da-6ced-11e7-8ec9-518ff26a77db.PNG)

### Version
Notepad 7.4.2 x86 in Windows 8.1 x64.

AngryGamer:
I'm aware of this bug. It's fairly easy to replicate.

Step 1: Move a tab using drag and drop.
Step 2: Use a tab switching hotkey/feature which doesn't send TCN_SELCHANGE.

Some of the tab switching features include:

- MRU list (**Ctrl+Tab** or R**ight click+Scroll wheel**)
- Window->tab
- Ctrl+Page Up/Down

jay:
Thanks. I tried just now and can reproduce <50% of the time using CTRL+TAB (no drag-and-drop), which is probably how it happens since I use that occasionally to switch tabs.

AngryGamer:
I'll open a PR fixing this problem.

AngryGamer:
I got distracted, but [PR #3552](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3552) is now up! The AppVeyor builds ([x86](https://ci.appveyor.com/api/buildjobs/eluihal9qojtpvvw/artifacts/Notepad%2B%2B.Win32.Unicode%20Release.exe) and [x64](https://ci.appveyor.com/api/buildjobs/sodq3i5cqjq61d4a/artifacts/Notepad%2B%2B.X64.Unicode%20Release.exe)) are here.

PR #3552, opened 2017-07-22
Make sure previous tab does not keep focus when switching tabs.

AngryGamer:
Fixes [#3545](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3545)


jay:
Thanks! I tested this in Windows 8.1 today and it looks like you fixed it.

PR #3555, opened 2017-07-23
Update czech.xml translation to 7.4.2

xomx:


xomx:
Hmm, isn't it failing because of I used the &lt; and &gt; notation for the << and >> in the Find dlg?

schnurlos:
Seems to be. I would change it from "Najít &gt;&gt;" to "Najít >>" and from "&lt;&lt; Najít" to "<< Najít" as this is the only real change you made. And this works in english and other languages.

xomx:
Sure, I could switch to real <> chars but then I think it will not be a valid xml-file (invalid starting tag ...).

schnurlos:
Better invalid and working as valid and not working as the checks fail ...

PR #3575, opened 2017-07-30
winmain: fix command line argument parsing

jay:
Work with the arguments in a temporary array of pointers to the command
line before assigning them to paramVector as generic_string.

Follow up to afb3889. Since then the arguments were copied to
paramVector as generic_string too early, before the command line parsing
finished.

Closes https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3575


donho:
Thank you @jay !

Issue #3576, opened 2017-07-30
new feature : Open in Defaut Viewer


Cyrillev91:
Add Option to Open file in its default viewer. (Has the same effect as double–clicking this file in Windows Explorer)

Example : you modify a file with notepad++ and you want to see this file in the default viewer 

- Without this feature : Right clic on tab then select "Open Containing Folder in Explorer" then double-clic on File
- With this feature : Right clic on tab then select "Open in Defaut Viewer"

With tab menu
![openindefaultviewer](https://user-images.githubusercontent.com/12849364/28754321-3748f00c-7543-11e7-96e6-673da74d203e.jpg)

With File menu
![openindefaultviewerfilemenu](https://user-images.githubusercontent.com/12849364/28754327-45bb0ab2-7543-11e7-8d0c-f09466ef73ff.jpg)

PR #3577, opened 2017-07-30
Opens file in its default viewer. Has the same effect as double–click ing this file in Windows Explorer

Cyrillev91:
Add Option to Open file in its default viewer. (Has the same effect as double–clicking this file in Windows Explorer)

Example : you modify a file with notepad++ and you want to see this file in the default viewer 

- Without this feature : Right clic on tab then select "Open Containing Folder in Explorer" then double-clic on File
- With this feature : Right clic on tab then select "Open in Defaut Viewer"

With tab menu
openindefaultviewer

With File menu
openindefaultviewerfilemenu

Issue : #3576

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -110,7 +110,26 @@ void Notepad_plus::command(int id)
 			// Opens file in its default viewer. 
             // Has the same effect as double–clicking this file in Windows Explorer.
             BufferID buf = _pEditView->getCurrentBufferID();
-			ShellExecute(NULL, NULL, buf->getFullPathName(), NULL, NULL, SW_SHOW);
+			HINSTANCE res = ::ShellExecute(NULL, TEXT("open"), buf->getFullPathName(), NULL, NULL, SW_SHOW);
+
+			// As per MSDN (https://msdn.microsoft.com/en-us/library/windows/desktop/bb762153(v=vs.85).aspx)
+			// If the function succeeds, it returns a value greater than 32.
+			// If the function fails, it returns an error value that indicates the cause of the failure.
+			int retResult = reinterpret_cast<int>(res);
+			if (retResult <= 32)
+			{
+				generic_string errorMsg;
+				errorMsg += GetLastErrorAsString(retResult);
+				errorMsg += TEXT("An attempt was made to execute the below command.");
+				errorMsg += TEXT("\n----------------------------------------------------------");
+				errorMsg += TEXT("\nCommand: ");
+				errorMsg += buf->getFullPathName();
+				errorMsg += TEXT("\nError Code: ");
+				errorMsg += intToString(retResult);
+				errorMsg += TEXT("\n----------------------------------------------------------");
+				
+				::MessageBox(_pPublicInterface->getHSelf(), errorMsg.c_str(), TEXT("ShellExecute - ERROR"), MB_ICONINFORMATION | MB_APPLMODAL);
+			}
 		}
 		break;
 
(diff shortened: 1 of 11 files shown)


PowerEditor/src/NppCommands.cpp
```diff
@@ -105,6 +105,15 @@ void Notepad_plus::command(int id)
 		}
 		break;
 		
+		case IDM_FILE_OPEN_DEFAULT_VIEWER:
+		{
+			// Opens file in its default viewer. 
+            // Has the same effect as double–clicking this file in Windows Explorer.
+            BufferID buf = _pEditView->getCurrentBufferID();
+			ShellExecute(NULL, NULL, buf->getFullPathName(), NULL, NULL, SW_SHOW);
```
SinghRajenM: What if there is no program associated with a particular file type. For example: one might have python script file (.py extension), then?

IMO, the "Open in Default Viewer" (under tab and File Menu) should be enabled only when a program/executable is associated with a particular file type.

Cyrillev91: Hi,
If there is no association for a file type, widows will ask with it program to open it.
How do you open this file? (Comment voulez-vous ouvrir ce fichier ?)
windowsask

SinghRajenM:
Correct. But the point is: there is no use of showing this dialog as you
will end up with opening file again either with npp or widows notepad (in
most of the cases).
So better enable this option only when associated program is available.
Consider using ```AssocQueryString``` API to check association.

Cyrillev91:
I try this but is not perfect : 

        
	if (isFileExisting)
	{
		PTSTR ext = PathFindExtension(buf->getFullPathName());

		HRESULT hres;
		wchar_t buffer[MAX_PATH] = TEXT("");
		DWORD bufferLen = MAX_PATH;

		hres = AssocQueryString(NULL, ASSOCSTR_COMMAND, ext, NULL, buffer, &bufferLen);
				

		bool enable= (hres == S_OK)                                // check no no association in Windows XP or other error
			&& (wcsstr(buffer, TEXT("OpenWith.exe")) == NULL)  // check no no association in Windows 10
			&& (wcsstr(buffer, TEXT("OpenAs_RunDLL")) == NULL) // check no association in Windows 7
			&& (wcsstr(buffer, TEXT("notepad++")) == NULL);    // check association with notepad++

		_tabPopupMenu.enableItem(IDM_FILE_OPEN_DEFAULT_VIEWER, enable);
	}




1. **With file without association** : 

- With Windows XP : 
       - **hres** != S_OK 

- With Windows 7  : 
       - **hres** == S_OK 
       - **buffer** = C:\Windows\system32\rundll32.exe C:\Windows\system32\shell32.dll,OpenAs_RunDLL %1

- With Windows 10 : 
       - **hres** == S_OK 
       - **buffer** = C:\WINDOWS\system32\OpenWith.exe "%1" 

- Other Windows : no tested

- Next  Windows : ?

2. **With file associated with notepad++** : 
       - **hres** == S_OK 
       - **buffer** = C:\WINDOWS\system32\NOTEPAD.EXE %1




         
This is not perfect because one does not know the syntax in the next Windows, and the Exe of Notepad ++ could change name

What do you think ?

SinghRajenM:
We are using 
```C++
hres = AssocQueryString(ASSOCF_VERIFY|ASSOCF_INIT_IGNOREUNKNOWN, ASSOCSTR_EXECUTABLE, ext, NULL, NULL, &bufferLen);
```
instead of 
```C++
hres = AssocQueryString(NULL, ASSOCSTR_COMMAND, ext, NULL, buffer, &bufferLen);
```

in our commercial products and have not faced any issue so far. In our case, we are not interested in app name, so we are passing ```NULL```.

> This is not perfect because one does not know the syntax in the next Windows, and the Exe of Notepad ++ could change name

If worried more, then consider using the return code of ```ShellExecute```. If return code<=32, then show error message to user similar to commit 7f98067c128a0fb00f3714c661c

Cyrillev91:
Thank you for your help

PR #3581, opened 2017-08-01
FindReplace dialog: Introduce replace backwards option & language id for

tonyk24:
Introduce a replace backwards option. Language ids for localization and fixed a crash when macro recording "Find previous"-action.

Before: 
3339_before

After:
3339_after


donho:
@tonyk24  why don't we do "<< replace" & "replace >>" - with the same logic of find ?

tonyk24:
The problem with "Replace >>" and "<< Replace" is that you cannot control the direction of "Replace all", it has to replace either up or down, when option "Loop" is not selected.

dinkumoil:
@tonyk24 Nice to see that you try to fix the issues I collected in #3339 

But OK, where are we now?

Before v7.4.2 we had

1. a "Find" button
2. a "Replace" button
3. a "Replace all" button
4. a "Direction backwards" radio button
5. a "Direction forwards" radio button

This was a clear and easy understandable/usable design.

Now we have

1. a "Find backwards" button
2. a "Find forwards" button
3. a "Replace" button
4. a "Replace all" button
5. a "Replace backwards" checkbox

The number of GUI elements is the same but the design is more complicated and the usability is worse.

To get a "Find backward" and a "Find forwards" functionality one could install the very good "Customize Toolbar" plugin and place the related toolbar buttons in the toolbar. Furthermore there were already keyboard shortcuts for "Find backward" and "Find forwards". That means, the changes (in v7.4.2) which led to all this issues were simply not neccessary.

IMHO it is worth to think about a roll back.

donho:
@tonyk24  Does the direction really matter for users while replacing all ?
I don't think so. We need only choose a direction for this operation.

donho:
@tonyk24 Indeed, as @dinkumoil mentioned about, if there's no wrap, replace all will be an issue.
I'll think about it.

dinkumoil:
@donho 
> Does the direction really matter for users while replacing all ?

Definitely there are circumstances where to have the possibility to replace only in one direction (forwards or backwards) is very useful. I personally needed this feature occasionally in the past.

To have such a feature or not makes the difference between a good and an excellent editor.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4d921b7ffd3b5d077c618b5e87fc483db65f63e8

PR #3595, opened 2017-08-03
Improve file extension movement between ListBox

SinghRajenM:
Now mouse double click can be used to move File extension between ListBoxes.

### DIFF
--- a/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
+++ b/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
@@ -121,6 +121,32 @@ INT_PTR CALLBACK RegExtDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lPar
 
 		case WM_COMMAND :
 		{
+			// Handle File association list extension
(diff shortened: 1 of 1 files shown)


PowerEditor/src/MISC/RegExt/regExtDlg.cpp
```diff
@@ -121,6 +121,32 @@ INT_PTR CALLBACK RegExtDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lPar
...
+					if (-1 != ::SendDlgItemMessage(_hSelf, LOWORD(wParam), LB_GETCURSEL, 0, 0))
+					{
+						HWND(lParam) == ::GetDlgItem(_hSelf, IDC_REGEXT_LANGEXT_LIST) ?
+							::SendMessage(_hSelf, WM_COMMAND, IDC_ADDFROMLANGEXT_BUTTON, 0) :
+							::SendMessage(_hSelf, WM_COMMAND, IDC_REMOVEEXT_BUTTON, 0);
+					}
+					return TRUE;
+				}
+
+				// On Selecting an item in the list, either button
```
donho: I don't see the need of these comments.

SinghRajenM: 😋 
Actually I wanted to move related code here for maintainability/uniformity (e.g. keeping listbox related code at single place). But later drop the idea just to make code review easier.

PR #3601, opened 2017-08-05
Do not reload function list when switching to the same file in cloned view

### LINKED ISSUE #3592: Function List causing disruptive lag when switching between cloned views
### Description of the Issue
When viewing a large piece of code in clone mode with the function list panel open and visible, notepad++ will recompile the function list when switching the cursor between the two views causing a disruptive lag.

1. Open a large piece of code with function list support
2. Right click on the file tab and select "Clone to Other View"
3. Open the function list panel (View -> Function List)
4. Check that the function list panel shows function names as expected, then click in the code body either side. There will be a noticeable lag.

### Expected Behaviour
Zero lag switching between views. No need to recompile the function list if it is known that no changes have taken place since the last compilation.

### Actual Behaviour
Considerable lag when switching between views due to recompiling of the function list

protector1990:
fixes #3592 

Notepad_plus::switchEditViewTo() method now does not call notifyBufferActivated() always - only if it detects that the buffer we are switching to in other view is actually different than current one. Assumption is that we don't really need to activate the buffer if it is already active, so in that case we can skip some potentially lengthy operations - like Function list recompiling. In that case, we can just invalidate rects for active and inactive view to update the colors of the tab controls.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3510,6 +3510,8 @@ int Notepad_plus::switchEditViewTo(int gid)
 		// Make sure the colors of the tab controls match
 		::InvalidateRect(_mainDocTab.getHSelf(), NULL, FALSE);
 		::InvalidateRect(_subDocTab.getHSelf(), NULL, FALSE);
+		// Make sure that plugins still get notified
+		doNotifyBufferActivated(_pEditView->getCurrentBufferID());
 	}
 	return oldView;
 }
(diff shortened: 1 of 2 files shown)


dail8859:
I haven't had a chance to look deeper into this but I feel this may introduce problems into plugins specifically catching `NPPN_BUFFERACTIVATED` in order to know when the user switches to the other view.

protector1990:
Ok, after doing a little deeper analysis of the code, it looks to me that notifyBufferActivated() purpose should be to just notify plugins that a buffer has been activated. Also, I think that listening for NPPN_BUFFERACTIVATED notification is not the best way to figure out if a user has switched views and there should probably be a different notification for that. But again, I am not that knowledgeable about the code base, so please excuse my ignorance.
So, in order to keep NPPN_BUFFERACTIVATED and still solve this problem, I propose following:
1. Extract the code notifying the plugins into a new function, say doNotifyBufferActivated
2. Call doNotifyBufferActivated instead of notifyBufferActivated when we detect that we are switching to the same file in cloned view

If this makes sense to you, I can implement that change and ammend pull request.

dail8859:
> there should probably be a different notification for that.

Yes probably but changing the behavior of the current notifications would potentially break a lot of plugins.

> I can implement that change and ammend pull request.

I'm not the one that decides if the  updated pr gets accepted so we'll see.

donho:
Good point @dail8859 !

PR #3616, opened 2017-08-13
Update Bulgarian translation

rddim:


donho:
Always create a new branch for the new PR.
Next time I'll reject PR if you use always the same.
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/be610b4022a2e44c3e88de35141eea60187518f9

PR #3630, opened 2017-08-18
Update Corsican translation for Notepad++ 7.5

Patriccollu:
Corsican translation has been updated with new strings and some minor changes.


Patriccollu:
@donho Thanks Don. Best regards, Patriccollu.

PR #3649, opened 2017-08-24
Update Bulgarian translation

rddim:


PowerEditor/installer/nativeLang/bulgarian.xml
```diff
@@ -1045,9 +1045,9 @@
 		<MiscStrings>
 			<word-chars-list-tip value="Позволява добавяне на допълнителни символи към дума, чрез двойно щракане върху нея или при включена опция за търсене на &quot;Само цели думи&quot;."/>
 			<word-chars-list-warning-begin value="Бъдете внимателни: "/>
-				<!-- $INT_REPLACE$ is a place holder, don't translate it -->
+				<!-- $INT_REPLACE$ е контейнер, не се превежда -->
```
donho: It's really no need to translate it, just leave it.

PowerEditor/installer/nativeLang/bulgarian.xml
```diff
@@ -1045,9 +1045,9 @@
...
 			<word-chars-list-space-warning value="$INT_REPLACE$ интервал(а)"/>
-				<!-- $INT_REPLACE$ is a place holder, don't translate it -->
+				<!-- $INT_REPLACE$ е контейнер, не се превежда -->
```
donho: No need to be translated.

PowerEditor/installer/nativeLang/bulgarian.xml
rddim: Corrected

PowerEditor/installer/nativeLang/bulgarian.xml
rddim: Corrected

Issue #3643, opened 2017-08-22
Function list doesnt not work for C# files


maroonmoon13:
### Description of the Issue
Enabling functionlist for .cs files does not show anything in the function list sidebar.

### Steps to Reproduce the Issue

1. save a .cs file with example code:
```
//Hello World
using System;
class Solution
{
    static void Main(String[] args)
    {
        Console.WriteLine("Hello World!");
    }
}
```
2. Enable functionlist from View->functionlist
3. 

### Expected Behavior
Function list should show classname and Main function under it.

### Actual Behavior
Functionlist remains epty and shows nothing

### Debug Information

![desktop_new 1 cs - notepad](https://user-images.githubusercontent.com/31255968/29582805-b2764f4c-8743-11e7-8e3d-9b44ff602f31.png)

uskovboris:
Good morining!

The problem root is that programs with missed namespace closure:
namespace Test
{...
are treateed incorrect by N++ functionParser (error in functionList.xml), but successfully compile
by Visual Studio. Moreover it is common practice, to miss the "namespace" keyword in LinqPAD environment, for example

Actully we correctly parse programs like
using System;

**namespace TestIme**
{
	class Solution
	{
		public static void Main(String[] args)
		{
			Console.WriteLine("Hello World!");
		}
		
		static int get()
		{
			
		}
	}
**}**

uskovboris:
It parse file if we add i.e. ';' after class close '}':
using System;

class Solution
{
    public static void Main(String[] args)
    {
        Console.WriteLine("Hello World!");
    }
**};**
But unlike C++ it is not necessary in C# to finish class defenition with the ';' character

aks2161989:
I am getting the same problem with Notepad++ version 7.8.2 (32 bit)

When I open a saved C# file, Function List shows only the class with the Main() function and ignores other classes.

![Screenshot (27)](https://user-images.githubusercontent.com/20129646/71778852-0a927300-2fd9-11ea-9e65-2d71e3fba7ee.png)

Function List does NOT work at all for "new" tabs with C# code in them. It just shows a single entry (for example) "new 6" in the Function List.

MAPJe71:
>Function List does NOT work at all for "new" tabs with C# code in them. It just shows a single entry, for example, "new 6" in the Function List.

[...]

PR #3665, opened 2017-08-27
Fix: Function list doesnt not work for C# files #3643

uskovboris:
It needs to test this fix compatibility with different languages


MAPJe71:
Single commit per PR is preferred (especially when they are not directly related) and in its own branch.

uskovboris:
I'm sorry, I'd commit the "fix annoing reload from disk requests" in mistake. I'll recommit "resolve #3643" in the evening

aks2161989:
Hello,

Has this change been incorporated in Notepad++? I am using version 7.8.2 (32 bit) and Function List does NOT work for C#.

Please see [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3643#issuecomment-570900692)

donho:
>  It needs to test this fix compatibility with different languages

Other language depends on (and works well with) this core feature, you should fix in `functionList.xml`.

Issue #3645, opened 2017-08-23
In Python the variable "def" not working correctly in Function List.


Longhorn99:
When the word "def" is used as part of a variable, it shows up as a "function" in the Function List.

example:

USER_DEF = 1000

def test():
    print("hello")

will show up as:

= 1000def test()

MAPJe71:
Duplicate of #3136.

PR #3672, opened 2017-08-29
Fix Python the variable "def" not working correctly in Function List

uskovboris:
Fixed issue #3645
Fix the "In Python the variable "def" not working correctly in Function List"

i've tested it with following files:

USER_DEF = 1000

def test2():
	pass

def test():
	print("hello")

def a():
	pass

#there is the same problem
USER_CLASS = 67

class Test:

	USER_DEF = 1000

	def test2():
		pass

	def test():
		print("hello")

	def a():
		pass
	
def b():
      pass

USER_DEF = 1000

class Test:

	def test2():
		pass

	def test():
		print("hello")

	def a():
		pass
	
def b():

USER_CLASS = 1000

class Test:

	def test2():
		pass

	def test():
		print("hello")

	def a():
		pass
	
def b():

### DIFF
--- a/PowerEditor/src/functionList.xml
+++ b/PowerEditor/src/functionList.xml
@@ -746,7 +746,7 @@
 						<nameExpr expr="\w+(?=[\(|:])" />
 					</className>
 					<function
-						mainExpr="(?&lt;=def\x20).+?(?=:)"
+						mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
 					>
 						<functionName>
 							<funcNameExpr expr=".*" />
@@ -754,13 +754,14 @@
 					</function>
 				</classRange>
 				<function
-					mainExpr="(?&lt;=def\x20).+?(?=:)"
+					mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
 				>
 					<functionName>
 						<nameExpr expr=".*" />
 					</functionName>
 				</function>
 			</parser>
+
 
 			<!-- ======================================================== [ Bash ] -->
 			<!-- BASH - Bourne-Again Shell                                         -->


PowerEditor/src/functionList.xml
```diff
@@ -746,22 +746,23 @@
 						<nameExpr expr="\w+(?=[\(|:])" />
 					</className>
 					<function
-						mainExpr="(?&lt;=def\x20).+?(?=:)"
+						mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
```
MAPJe71: Replace `\sdef\x20` with `\bdef\h`.

PowerEditor/src/functionList.xml
```diff
@@ -746,22 +746,23 @@
...
 					>
 						<functionName>
 							<funcNameExpr expr=".*" />
 						</functionName>
 					</function>
 				</classRange>
 				<function
-					mainExpr="(?&lt;=def\x20).+?(?=:)"
+					mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
```
MAPJe71: Likewise.

PowerEditor/src/functionList.xml
```diff
@@ -746,22 +746,23 @@
...
 				>
 					<functionName>
 						<nameExpr expr=".*" />
 					</functionName>
 				</function>
 			</parser>
 
+
```
MAPJe71: Remove superfluous empty line.

MAPJe71:
Could try/test the following parser?
```xml
			<!-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
			|   https://sourceforge.net/p/notepad-plus/discussion/482781/thread/515001cd/#35ce
			|
			|   1) Strings is added to comments
			|   2) Detection of empty and nested classes
			|   3) Ability to go to the class header
			\-->
			<parser
				displayName="[#4 TODO] Python by itraveller/guy038"
				id         ="python_function"
				commentExpr="(?x)                                               # free-spacing (see `RegEx - Pattern Modifiers`)
								(?m-s:\x23.*$)                                  # Single Line Comment
							|	(?:                                             # Document String - Double or Single Triple-Quoted
									(?'TQT'\x22{3}|\x27{3})                     # ...start-of-string indicator
									(?s:.*?)                                    # ...whatever (dot matches line breaks), until
									(?:	\k'TQT'                                 # ...equal end-of-string indicator
									|	\Z                                      # ...or end-of-text
									)
								)
							|	(?:                                             # String Literal - Double or Single Quoted
									(?'QT'\x22|\x27)                            # ...start-of-string indicator
									(?s:.*?)                                    # ...whatever (dot matches line breaks), until
									\k'QT'                                      # ...equal end-of-string indicator
								)
							"
			>
				<!-- class ranges are limited via first level headers -->
				<classRange
					mainExpr    ="(?x)                                          # free-spacing (see `RegEx - Pattern Modifiers`)
							^
							class
							\K                                                  # discard text matched so far
							\h
							.+?
							(?=
								(?:\r?\n|\n?\r)
								(?-i:def|class)
								\h
							|
								\Z
							)
						"
				>
					<className>
						<nameExpr expr="\h+\K.+?(?=:)" />
					</className>
					<function
						mainExpr="(?x)                                          # free-spacing (see `RegEx - Pattern Modifiers`)
								(?m-i)                                          # ^ and $ match at line breaks - case-sensitive
									\A\h*\K\h\w+                                # class link
								|
									^\h+def\h+\K.+?(?=:)                        # nested functions
								|
									^\h+\Kclass\h.+?(?=:)                       # nested classes
							"
					>
						<!-- leading space is left to place the class link at the top of the list -->
						<functionName>
							<funcNameExpr expr=".+" />
						</functionName>
					</function>
				</classRange>
				<function
					mainExpr="(?x)                                              # free-spacing (see `RegEx - Pattern Modifiers`)
							(?m-i)                                              # ^ and $ match at line breaks - case-sensitive
								^\h*def\h+\K.+?(?=:)                            # upper and nested functions
							|
								^\h+\Kclass\h.+?(?=:)                           # nested classes
						"
				>
					<functionName>
						<nameExpr expr=".+" />
					</functionName>
				</function>
			</parser>
```

PowerEditor/src/functionList.xml
uskovboris: I think it is better to get the solution from
http://sourceforge.net/p/notepad-plus/discussion/482781/thread/515001cd/#b5d2
but it needs to improve the solutions suggested in that discussion. I'll comment in a one

uskovboris:
Actually both parsers are failed. The source

class Test:
	def test1():
		pass
	
	def test2(e):
		def intest2():
			pass
		pass

will be treated as


but there was not function "Test"

Moreover they treats "def test: " as a correct function definition, but it is not. A correct one is "def test(<optional-arguments-list>)"

uskovboris:
It is possible to correct both solutions. Now I doing my best to make a correct one

uskovboris:
Also, there are not nested classes in python. Or I'm confuse something?

uskovboris:
Also it does not work for 
<code>

class Test:
	def test1():
		pass
	
	def test2(e):
		def intest2():
			pass
		pass

		
class Test1:
    class Subclass1:
		def Subclass1test1():
			pass
		
		def Subclass1test2(e):
			def Subclass1intest2():
				pass
			pass
			
	def test1():
		pass

</code>

Actual:
image

uskovboris:
But original parser can't parse it in too:
image

uskovboris:
I have a suggestion: let introduce a simple and obvious way to treat with nested function and classes. It need to implement simplier method to define subcalsses and internal functions. It need to do in a notepad++ code side. Actually there are nested functions in Pascal/Delphy, C# 7.0 and many other languages.  It needs to have an entity like nestedClassRange, nestedFunction. And the product should search this objects "in-functions" and "in-classes" founded in the first level of file preprocessing

uskovboris:
TBD: example of special lang

<item type="class">
   <operator type="and">
       <clausure condition="match" value="class" />
        <clausure tupe=ends-with" value=":" />
         <info id="name" type="match" value="class(.*)" /> TDB....

MAPJe71:
Any change to the Function List source code and/or the `functionList.xml` schema will affect **all** language parsers i.e. possibly a huge impact on installed base.
Make sure you can prove/verify correct behaviour by also providing automated tests.

uskovboris:
Yes, I understand the seriousness of change. Yes I can provide functionality with auto-tests and also, I don't plan to make an alternative method of language elements parsing which will be live alongside with a current imolementation. I was inspired by the Mandiant IOC/Open IOC 1.1 format (uses in cybersecurity). I'm not shure, but it looks pretty convenient

donho:
@uskovboris I've got unknown repository so I can not merge your code - could you fix that?

PR #3670, opened 2017-08-29
BaanC FunctionList

oirfeodent:
Enable FunctionList for BaanC language

[BaanC_FunctionList.bc.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1262603/BaanC_FunctionList.bc.txt)

@MAPJe71 attached a source file with a list of possible combinations.
At this point, not all the combinations are displayed... But, not sure where the issue is.

Thx for your review.

PR #3675, opened 2017-08-30
BaanC Function List xml changes

oirfeodent:
Includes BaanC to the function list
[BaanC_FunctionList.bc.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1262824/BaanC_FunctionList.bc.txt)

A sample source file with the possibilities is attached for preview.

@MAPJe71 , this PR is a continuation of the #3670 .
I had to close it because, I messed up with the commits.
But, all the review comments are included, except the one about comments.
The '|' is the comment symbol and not the forward slash.

As of now, not all the possible ways of writing a function is captured by this regex.
Multi-line regex looks complicated to me, I have tested only the basic ones.
If you have any pointers on how to achieve all combinations let me know.

Regards,

### DIFF
--- a/PowerEditor/src/functionList.xml
+++ b/PowerEditor/src/functionList.xml
@@ -89,6 +89,7 @@
 			<association id=   "innosetup_syntax"     langID="46"                          />
 			<association id=  "powershell_function"   langID="53"                          />
 			<association id=  "javascript_function"   langID="58"                          />
+			<association id=       "baanc_function"   langID="60"                          />
 			<!-- ======================================================================== -->
 			<association id=         "krl_function"   userDefinedLangName="KRL"            />
 			<association id=         "krl_function"   ext=".src"                           />
@@ -1293,6 +1294,22 @@
 							)
 						"
 				/>
+			</parser>
+			
+			<!-- ======================================================= [ BaanC ] -->
+			<parser
+				displayName="BaanC"
+				id         ="baanc_function"
+				commentExpr="(?m-s:/|.*?$)"
+			>
+				<function
+					mainExpr="^\h*(?i:function)\s+[\w\.\s]+\("
+				>
+					<functionName>
+						<nameExpr expr="[\w\.\s]+\(" />
+						<nameExpr expr="[\w\.\s]+" />
+					</functionName>
+				</function>
 			</parser>
 
 			<!-- ================================================================= -->


MAPJe71:
>The '|' is the comment symbol and not the forward slash.

The '|' is the `or` operator in Regex, it has to be escaped with a *backward* slash.

MAPJe71:
Try this one:
```xml
			<parser
				displayName="BaanC"
				id         ="baanc_function"
				commentExpr="(?x)                                               # free-spacing (see `RegEx - Pattern Modifiers`)
								(?m-s:\x7C.*$)                                  # Single Line Comment
							"
			>
				<function
					mainExpr="(?x)                                              # free-spacing (see `RegEx - Pattern Modifiers`)
							(?im)                                               # case-insensitive, ^ and $ match at line breaks
							^\h*                                                # optional leading whitespace at start-of-line
							(?i:function)\s+
							(?:extern\s+)?
							(?:(?:boolean|double|long|string|void|domain\s+[A-Za-z_]\w*)\s+)?
							\K
							\w+(?:\s*\.\s*\w+)*
							\s*\(
						"
				>
					<functionName>
						<nameExpr expr="\w+(?:\s*\.\s*\w+)*" />
					</functionName>
				</function>
			</parser>
```

Present Function List implementation does not support "embedded" comments.

```baanc
| This is a BaanC comment.
| A comment starts with a pipe and ends at any EOL (Unix/Win/Mac)
| Comments can be placed anywhere.
|
| BaanC is case insensitive and space insensitive.
|
| Any number of SPACE and TAB can be used everywhere.
|
| Function definition:
| * Should start with the keyword "function" (preceding spaces/TABs allowed)
| * followed by simple return type (if any) and function name.
| * Simple returns are [string|boolean|long|double|void]. void is optional.
| * for complex return types keyword "domain <domainName>" is used.
| * '()' are used for function parameters
| * '{}' are used to indicate function open/close.
| * External functions are defined with two keywords "function extern" + all above conditions.
|
| Examples:

function func1(long var1)| Most Common practice for writing function
{
	| Do Something 1.
}

| Preceed an inline comment with at least 2 characters i.e. "func2" will not show in the FunctionList tree.
function string func2( | Next preferred method to write functions, when multiple incoming params
			double var2,
			string var2_1
			)
{
	| Do Something 2.
}

| Added an extra space so "func2.1" will show in the FunctionList tree.
function string func2.1(  | Next preferred method to write functions, when multiple incoming params
			double var2.1,
			string var2.1_1
			)
{
	| Do Something 2.1.
}

| "Embedded" comment is not supported i.e. "func3" will not show in the FunctionList tree.
function domain tcyesno func3 | Less used method, but both function and comments are valid.
				(long var3,
				 long var3_1)
{
	| Do Something 3.
}

| Moved "embedded" comment so "func3.1" will show in the FunctionList tree.
| Less used method, but both function and comments are valid.
function domain tcyesno func3.1 
				(long var3.1,
				 long var3.1_1)
{
	| Do Something 3.1.
}

| "Embedded" comment is not supported i.e. "func4" will not show in the FunctionList tree.
function 
	extern 	|This kind of comment and function are valid, but not a common practice.
		func4
			(boolean var4){ 
	| Do Something 4.
}

| Moved "embedded" comment so "func4.1" will show in the FunctionList tree.
|This kind of comment and function are valid, but not a common practice.
function 
	extern 	
		func4.1
			(boolean var4.1){ 
	| Do Something 4.1.
}

function extern void func5(long var5){ |This comment and function are valid. Common practice.
	| Do Something 5.
}

| "Embedded" comment is not supported i.e. "func6" will not show in the FunctionList tree.
function extern domain tcbool func6 |This comment and function are valid, but not a common practice.
(boolean var6){
	| Do Something 6.
}

| Moved "embedded" comment so "func6.1" will show in the FunctionList tree.
|This comment and function are valid, but not a common practice.
function extern domain tcbool func6.1 
(boolean var6.1){
	| Do Something 6.1.
}
```

oirfeodent:
@MAPJe71 thx for the review.
I have committed the same. 
However, not sure what I did for the third commit to happen and also not sure about the implication.

If you see any issue, I can close this PR and re-open a fresh one. 

Regards,

PR #3677, opened 2017-08-30
Auto compelte for Batch

SinghRajenM:
V7.5.1 change.log says [Auto complete support for batch file](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/d7f64d2027703b43e05206f15c5dc90bd1e93e39/PowerEditor/bin/change.log).
But auto complete file has not been included in installer.

## Before:
image

## After:
image


SinghRajenM:
Related Bug #3680

PR #3689, opened 2017-09-01
Aligned the installation script

marcoM32:


marcoM32:
Needless, the fix is already entered in the master branch

PR #3702, opened 2017-09-05
FunctionList for BaanC Language

oirfeodent:
Final changes after review with PR #3675 .


oirfeodent:
Will reopen new PR for Sections and include the same.

PR #3703, opened 2017-09-05
Update Indonesian Translation

ksatria:
I found that Backward Direction checkbox in Find window still using English. I translated it into "Cari ke belakang" as Indonesian version.


donho:
@ksatria Merge part shouldn't be in the PR, please redo your PR

PR #3732, opened 2017-09-12
fix feedScintKeys when more than two shortcuts are configured

### LINKED ISSUE #3720: Shortcuts for SCI_CUT, COPY, UNDO, REDO, etc are not saving
Steps to Reproduce:
1. Click Settings –> Shortcut mapper.
2. Choose "Scintilla commands" tab.
3. Double click on "SCI_CUT".
4. Check Alt and X. Leave Ctrl and Shift unchecked.
5. Press "Add" to add this shortcut to the list. This will be third shortcut.
6. Press "OK".
7. Close shortcuts window. Added shortcuts work yet.
8. Restart the Notepad++.

Expected Behavior:
Shortcuts are saved in shortcuts.xml and work after restarting Notepad++.

Actual Behavior:
Shortcuts are not saved in shortcuts.xml and do not work after restarting Notepad++.

cmeriaux:
fixes #3720 

The iterator of xml node was wrong. Iterator read only the first child, others children was ignored.
Now every children are read.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -2458,7 +2458,7 @@ void NppParameters::feedScintKeys(TiXmlNode *node)
 				KeyCombo kc;
 				for (TiXmlNode *nextNode = childNode->FirstChildElement(TEXT("NextKey"));
 					nextNode ;
-					nextNode = childNode->NextSibling(TEXT("NextKey")) )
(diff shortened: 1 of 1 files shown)


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2d1eee8b47ac1e0498252b578e6791abbd540e16

PR #3734, opened 2017-09-14
Updated existing solution to Visual Studio 2017 and added Scintilla project in it.

MojaveWastelander:
The "Notepad++" project now have all files filtered and replicate filesystem structure.
[Solution tree](https://imgur.com/0KinBZH)


donho:
Boost should be build with Scintilla - at this point there's no advantage in this PR.

PR #3743, opened 2017-09-15
Nitpicking - quotations

RenniePet:
Britain should be capitalized, one of the quotations is duplicated.


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6145,7 +6145,6 @@ Quote quotes[nbQuote] =
 	{"Anonymous #120", "Life is what happens to you while you're looking at your smartphone."},
 	//{"Anonymous #121", ""},
 	{"Anonymous #122", "Nerd?\nI prefer the term \"Intellectual badass\"."},
-	{"Anonymous #123", "How can you face your problem if your problem is your face?"},
```
MAPJe71: Instead of simply removing the line change it to an out-commented line as e.g. "Anonymous 121".

RenniePet:
I am not at all familiar with Git and GitHub, have tried to add a second pull request, but am unsure if it worked.

PR #3750, opened 2017-09-18
Regression: Fixed Nsis warning

SinghRajenM:
Moving mainSection from top to bottom created nsis warning.

image

I know there was fixed done related to uninstaller (installing x64 over x86 and vice versa). Instead of moving whole function, declaring variable ```Var diffArchDir2Remove``` just before use (before both mainSection and .onInit) could have fixed the issue.

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -86,13 +86,7 @@ page Custom ExtraOptions
 
 
 !include "nsisInclude\langs4Installer.nsh"
-
-
-
-
 !include "nsisInclude\mainSectionFuncs.nsh"
-
-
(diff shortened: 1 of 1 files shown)


donho:
I believe that I have move this block higher, then it made a regression due to a (some) variable(s) is (are) not initialized yet. So let's keep this warning. It doesn't matter while it's in compiling time but not in run time.

SinghRajenM:
> It doesn't matter while it's in compiling time but not in run time.

@donho 
No, this warning matters in run time as well. Side effects are explained in PR #3371.

SinghRajenM:
Closing as could not update it again here. Will create new as regression is not limited to warning it caused other issues as well.

PR #3751, opened 2017-09-18
Write version info to installer

SinghRajenM:
1. Write version info to the installer
2. Avoid manual work of updating version info every time. Use notepad++.exe to get version.

image

**Note: As per [nsis doc](http://nsis.sourceforge.net/Reference/!getdllversion) ```!getdllversion``` was introduced in nsis version v3.0a0. I'm assuming npp installer is generated using above this version.**

### DIFF
--- a/PowerEditor/installer/nsisInclude/globalDef.nsh
+++ b/PowerEditor/installer/nsisInclude/globalDef.nsh
@@ -29,10 +29,43 @@
 ; Define the application name
 !define APPNAME "Notepad++"
 
-!define APPVERSION "7.5.1"
-!define APPNAMEANDVERSION "${APPNAME} v${APPVERSION}"
-!define VERSION_MAJOR 7
-!define VERSION_MINOR 51
+; ------------------------------------------------------------------------
+; Get Notepad++ version from the notepad++ binary
+
+!ifdef ARCH64
+	!getdllversion "..\bin64\notepad++.exe" nppVer_
+!else
+	!getdllversion "..\bin\notepad++.exe" nppVer_
+!endif
+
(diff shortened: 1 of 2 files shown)


PowerEditor/installer/nsisInclude/globalDef.nsh
```diff
@@ -29,10 +29,43 @@
...
+; ------------------------------------------------------------------------
+; Get Notepad++ version from the notepad++ binary
+
+!ifdef ARCH64
+	!getdllversion "..\bin64\notepad++.exe" nppVer_
+!else
+	!getdllversion "..\bin\notepad++.exe" nppVer_
+!endif
+
+!define APPVERSION		${nppVer_1}.${nppVer_2}		; 7.5
```
donho: I just don't see how nppVer_1, nppVer_2, nppVer_3 & nppVer_4 are asigned.

MAPJe71: They are assigned by `!getdllversion`.

SinghRajenM: Right, it is assigned by ```!getdllversion```

From [nsis doc](http://nsis.sourceforge.net/Reference/!getdllversion)
>!getdllversion
>
>This is similar to GetDLLVersionLocal, only it stores the version number in defines and can therefore be used anywhere, not just inside functions and sections.
>
>!getdllversion "$%windir%\explorer.exe" expv_
>!echo "Explorer.exe version is ${expv_1}.${expv_2}.${expv_3}.${expv_4}"

PR #3789, opened 2017-10-01
Translations: Chinese(Taiwan): Replace tab indentation to align with rest of the style

Lin-Buo-Ren:
A few of the markup uses tab indentation instead of spaces and will display misaligned on viewers doesn't consider tab characters as certain space characters width, replace them with space indentation of the rest of the markup to fix this issue.

This patch is NOT tested as it is done on a Linux laptop ;), please review before merging.

Signed-off-by: 林博仁 &lt;<Buo.Ren.Lin@gmail.com>&gt;


donho:

Issue #3780, opened 2017-09-28
Crash with Folder as Workplace opened on folder refresh


YuriPetrov:
### Description of the Issue
Notepad++ throws an error "Queue Overflowed" and after a while crashes with Folder as Workplace opened when files and folders in a workplace tree are refreshed.

### Steps to Reproduce the Issue
1. Create ``C:\test`` folder.
2. Unzip ``create_files.bat`` from [create_files.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1340491/create_files.zip) into it.
3. Open Folder as Workplace panel in Notepad++ and Add ``C:\test``.
4. Run ``C:\test\create_files.bat`` several times.

### Actual Behavior
Folder as Workplace tree is blinking and throws an error "Queue Overflowed".
![image 139](https://user-images.githubusercontent.com/1536964/30962099-1fa3e2d0-a451-11e7-9b30-c4d5ebb0d671.png)
After a bunch of the same steps Notepad++ crashes.

### Debug Information
Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll PluginManager.dll

grishavanika:
Just for someone who want/will investigate the issue.

`CReadDirectoryChanges` is limited to detect up to 1000 changes because of `CThreadSafeQueue` (semaphore is used internally as a way to do notifications). Just curious: maybe auto-reset `Event` will work for the same purpose without limitations ?

donho:
Cannot reproduce the crash, with 5-6 launch of script. Any other way to reproduce it?

grishavanika:
@donho, I'm able to reproduce it with few runs (run, wait while process completes, ran again).
My guess is that you have faster system, maybe changing `create_files.bat` from 1000 files creation to, for example, 2000 will help you ?:
```
del /s /q /f C:\test\*.txt
for /l %%i in (1,1,2000) do copy C:\test\create_files.bat C:\test\create_files%%i.txt
```
(also, note, `C:\test` is hardcoded there so `create_files.bat` should be placed exactly in that folder)

MetaChuh:
[...]

PR #3793, opened 2017-10-01
Fix crash when too many directory changes happens

grishavanika:
Hi,

I'm new to making contribution to open source projects, so please correct me If I'm doing something wrong.

I started looking into [this issue (#3780)](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3780).

Main change is replacing Semaphore with Event as a way to do notifications for `CReadDirectoryChanges`.

Please take a look into changes. I did not learn the codebase, only the code around the crash, so maybe there is some deeper problem I did not noticed.

Thank you

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -39,6 +39,30 @@
 
 using namespace std;
 
+static void processChange(HWND notifyWindow, const TCHAR *fullFileName, Buffer *buf,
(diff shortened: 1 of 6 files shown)


donho:
@grishavanika The refactoring is good but not when you fix bugs.
Please provide a new PR (this one has conflict with current master) without factoring - I will check it in priority. Thank you.

PR #3832, opened 2017-10-19
adding sass

sadik-fattah:
from http://www.kuenzelit.de/blogContent/sass-user-defined.xml


dail8859:
Notepad++ does not ship with any user defined languages. Plus, just putting that file there does not include it in the installer.

PR #3870, opened 2017-11-03
Fix for "Toggle Single Line Comment" malfunctioning with HTML/XML

### LINKED ISSUE #3869: "Toggle Single Line Comment" malfunctioning with HTML/XML (fix now included)
### Description of the Issue
Normally upon pressing **Ctrl+Q**, user should be able to toggle selected lines between commented/uncommented states. However, when editing HTML/XML files (or possibly any types of files which have single line comments with both opening and closing tags - however, I haven't verified this claim yet) this feature seems to malfunction and continuous invocation of the functionality only results in **more comments** being added.

1. Open Notepad++ and create a new document
2. Set editing mode to XML **(Language -> XML)**
3. Input any text
4. Press **Ctrl+Q** to invoke "Toggle Single Line Comment".
5. Press **Ctrl+Q** to invoke it again.

### Expected Behavior
After step 4, line should be commented (for example, `<!-- foo -->`). It is the case at the moment.
After step 5, line should be uncommented again (for example, `foo`). Is is **not** the case at the moment.

### Actual Behavior
After step 5, line got a second set of comments around it (`<!-- <!-- foo --> -->`).

### Debug Information
Notepad++ v7.5.1   (64-bit)
Build time : Aug 29 2017 - 02:38:44
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)

CookiePLMonster:
This fixes #3869

As seen, a one off issue caused the last line character to be replaced by null terminator in the call to `getGenericText` (said function uses `_tcsncpy_s` with `_TRUNCATE` parameter which forces null terminator to be added if buffer is not large enough). Since provided buffer can store as many characters as the line has, null terminator replaces the last character in the line - and that prevents further parts of the code from detecting closing comment tags properly.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3951,7 +3951,7 @@ bool Notepad_plus::doBlockComment(comment_mode currCommentMode)
 	//The NORMAL MODE is used for all Lexers which have a commentLineSymbol defined.
 	//The ADVANCED MODE is only available for Lexers which do not have a commentLineSymbol but commentStreamSymbols (start/end) defined.
 	//The ADVANCED MODE behaves the same way as the NORMAL MODE (comment/uncomment every single line in the selection range separately)
-	//but uses two smbols to accomplish this.
+	//but uses two symbols to accomplish this.
 	bool isSingleLineAdvancedMode = false;
 
 	if (buf->getLangType() == L_USER)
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
Updated to correct buffer size - for normal mode replacements `linebuf` included space for comments (usually 2 or 3 characters) while it did not need to since `linebuf` only ever stores a raw text line, without adding any extra tags to it.

CookiePLMonster:
I have squashed both commits to conform with contributing guidelines more - and since both affect the same line the end result is objectively cleaner.

CookiePLMonster:
Making advantage of the burst of activity to attempt to bump this. @donho

PR #3871, opened 2017-11-05
Fix #3868 - Menu key does not work in Project Panel

### LINKED ISSUE #3868: Menu key does not work in Project Panel
### Description of the Issue
The [menu key](https://en.wikipedia.org/wiki/Menu_key) does not work when the Project Panel has focus.

1. Select any item in the Project Panel
2. Press the Menu Key

### Expected Behavior
The context menu to appear (as would happen if you right-clicked)

### Actual Behavior
Nothing

### Debug Information
Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : ComparePlugin.dll IndentByFold.dll JSMinNPP.dll NppExport.dll PluginManager.dll XMLTools.dll mimeTools.dll NppConverter.dll

luisffranca:
If the context menu is generated from the keyboard (for example, by the Menu key), then the x- and y-coordinates are -1 and the application should display the context menu at the location of the current selection.

### DIFF
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -25,6 +25,7 @@
 // along with this program; if not, write to the Free Software
 // Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
+#include <WindowsX.h>
 
 #include "ProjectPanel.h"
 #include "resource.h"
@@ -43,9 +44,6 @@
 #define INDEX_CLOSED_NODE    4
 #define INDEX_LEAF           5
 #define INDEX_LEAF_INVALID   6
-
-#define GET_X_LPARAM(lp)     LOWORD(lp)
-#define GET_Y_LPARAM(lp)     HIWORD(lp)
 
 INT_PTR CALLBACK ProjectPanel::run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam)
 {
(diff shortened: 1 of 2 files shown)


PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -44,8 +44,8 @@
 #define INDEX_LEAF           5
 #define INDEX_LEAF_INVALID   6
 
-#define GET_X_LPARAM(lp)     LOWORD(lp)
-#define GET_Y_LPARAM(lp)     HIWORD(lp)
+#define GET_X_LPARAM(lp) static_cast<short>(LOWORD(lp))
```
CookiePLMonster: Aren't those already defined in Windows headers?

DavidHansen-0x454d43:
@luisffranca This change doesn't work for me... GetCursorPos(&p) gets the current position of the mouse cursor, and not the position of the selected treeview item. In my testing, your code will only pop up the apps menu if the mouse is positioned over a workspace item. 

I think you need to use something like [this](https://msdn.microsoft.com/en-us/library/windows/desktop/bb773850(v=vs.85).aspx), to grab the bounding rectangle of the selected item.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
luisffranca: Yes, you're right. I've included the header now. Thanks!

luisffranca:
@DavidHansen-0x454d43 Thanks for the tip! I've changed the code and now the apps menu will pop up when a treeview item is selected.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -133,7 +131,30 @@ INT_PTR CALLBACK ProjectPanel::run_dlgProc(UINT message, WPARAM wParam, LPARAM l
...
+				if (xPos == -1 && yPos == -1)
+				{
+					HTREEITEM selectedItem = _treeView.getSelection();
+
+					if (selectedItem)
+					{
+						RECT selectedItemRect;
+						if (TreeView_GetItemRect(_treeView.getHSelf(), selectedItem, &selectedItemRect, TRUE))
+						{
+							showContextMenuFromMenuKey(selectedItem, selectedItemRect.left, selectedItemRect.top);
```
DavidHansen-0x454d43: Its working correctly now, good job @luisffranca !

My only suggestion would be to change the offset a bit so the apps menu doesn't entirely obscure the selected item:
showContextMenuFromMenuKey(selectedItem, (selectedItemRect.left+selectedItemRect.right)/2, (selectedItemRect.top+ selectedItemRect.bottom)/2);

luisffranca: @DavidHansen-0x454d43 I've done the change you suggested, and it looks better now :) Thanks!

Issue #3142, opened 2017-04-06
"Monitoring" doesn't detect changes


ndrewl:
### Description of the Issue

"Monitoring (tail-f)" feature doesn't work when writing to a file without closing it.
It doesn't update the view until file is closed.

### Steps to Reproduce the Issue

Instructions are given for python, for simplicity, but it's not specific for python
1. Open python in interactive mode:
```bash
python -i
```
Then type:
```python
f = open('test.txt', 'w')
```
2. Open file 'test.txt' in Notepad++ and click "Monitoring (tail -f)", the "eye" button.
3. Continue in python session, keeping eye on the notepad++ window:
```python
f.write('hey')
f.flush()
```
4. Nothing happens in the Notepad++ (but the message _is_ in file, you can open it with notepad.exe, for example, and check)
5. Again in python:
```python
f.close()
```
6. Now the message appears in the Notepad++

### Expected Behavior

Notepad++ should monitor for these changes, even if the writer doesn't close file.

### Actual Behavior

Output didn't change, Notepad++ doesn't "see" these updates.

### PS:
`tail -f` actually behaves correctly in this scenario.

knh1:
Would be cool if this got fixed at some point.  Switching tabs also updates the output.

jrocha:
Got this same behavior monitoring OpenVPN logs.

djspacedevil:
Hy,

still not fixed in Version 7.5.1 32-bit

jrocha:
@djspacedevil this version does not have the fix, v7.5.2 does not contain the fix as well. my PR is pending.

vdeconinck:
@donho Any idea when this will be integrated ? Looks like low hanging fruit one-liner, and without it, the "tail -f" feature is useless for me (and others :-)). FWIW, my use case is RealTerm - https://sourceforge.net/projects/realterm/ - which also keeps its capture file open for the whole session. Thx.

darakna:
7.5.4 still not fixed, used powershell instead: Get-Content -Path "C:\scripts\test.txt" -Wait
https://gist.github.com/marioflores/6509931

sawlogs:
Visual Studio Code with the "Log Viewer" extension supports this feature, so I had to switch to it. If anyone is looking for a text editor with tail -f support, this is my recommendation. It even supports large log files and highlights the log format.

epalm:
[...]

PR #3882, opened 2017-11-10
Fix for "Monitoring" doesn't detect changes #3142

jrocha:
Just added a flag to detect changes in file size.


drewler:
I have tested your code and it works perfectly 😄 

I hope it gets merged soon!

vdeconinck:
+1
This small fix should work 99.99% of the time (the exception being overwriting bytes in random access mode), and 100% of the logs, with no adverse effect. 
Nice catch.

billybooth:
Any word on getting this PR accepted?  Seems like an exceedingly safe, no-brainer that will make many of us very happy.

jrocha:
I really would like to know why this is still hanging over here, I really would like to use official binaries again.
@donho Any word on getting this PR accepted?

PR #3892, opened 2017-11-16
Fix #2049 - Request: Mute sound on find

### LINKED ISSUE #2049: Request: Mute sound on find
Can you have an option or is there an option already to mute the sound when find does not find anything? The flashing of the window is enough. I often search through multiple files and the sound is a distraction.

talhafarooq:
1. Added a checkbox in Preferences under MISC labelled Mute Text Not Found Sound
2. Added corresponding variable for checkbox in parameters.h
3. Added corresponding changes to save value when application exits
4. Changed FileReplaceDlg.cpp to check for this variable and to mute the sound if true


donho:

sasumner:
Maybe this is better addressed by having application level control on the sound.  What I mean by this (and I'm on Win7 so if later versions treat it differently this may not be relevant) is to change Notepad++ so that when opening the "sound" control (e.g. from the speaker icon in the tray) Notepad++ appears there (currently it doesn't) and one can set the sound settings from it.  Maybe this is overkill since N++ makes so few sounds, but I always lean towards solutions that don't involve adding another Preferences setting.  :)

PR #3896, opened 2017-11-17
Fix #1017 - GDI objects leak

### LINKED ISSUE #1017: GDI objects leak
After execute the explorer context menu with Notepad++ menu item ("Edit with Notepad++") leaked one GDI object.
Plugin named "Context menu entry" in default installer.

ATGH15102AFMLD:
Fixed a leak of one GDI object, which occurs after each opening of the Shell context menu in which there is "Edit with Notepad ++".
Changes have been made in the NppShell.dll project. A variable is added to the class, which is freed in the destructor.


CookiePLMonster:
Attempting to bump it - since it's fairly relevant, leaks in Shell are very nasty.

PR #3898, opened 2017-11-18
Fix #3863 - Ctrl+A Keyboard shortcut doesn't work in the MD5 Generator

### LINKED ISSUE #3863: Ctrl+A keyboard shortcut doesn't work in the MD5 generator
In the "Generate MD5 Digest" window ("Tools > MD5 > Generate..."), the Ctrl+A keyboard shortcut doesn't work. That's quite surprising, and this shortcut would be very useful (for selecting the current input, letting easily delete/replace it, which is currently tedious).

Note that most other shortcuts (e. g. Ctrl+C) do work already.

* Work : Ctrl + X (cut), C (copy), V (paste), Z (undo)
* Don't work : Ctrl + A (select all), Y (redo) (maybe this one would be harder to implement?)

Notepad++ 7.4.2

luisffranca:
When ES_MULTILINE is set for an Edit Control, Ctrl+A stop working. To fix that, I added a new window procedure for the MD5 Text edit control, where Ctrl+A is processed.

Although WM_KEYDOWN could be used, it would still present a beep sound. Therefore, WM_CHAR was used.

### DIFF
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -192,6 +192,13 @@ INT_PTR CALLBACK MD5FromTextDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 				DEFAULT_PITCH | FF_DONTCARE, "Courier New");
 			::SendMessage(::GetDlgItem(_hSelf, IDC_MD5_TEXT_EDIT), WM_SETFONT, reinterpret_cast<WPARAM>(hFont), TRUE);
 			::SendMessage(::GetDlgItem(_hSelf, IDC_MD5_RESULT_FOMTEXT_EDIT), WM_SETFONT, reinterpret_cast<WPARAM>(hFont), TRUE);
+
+			_hMD5TextEdit = ::GetDlgItem(_hSelf, IDC_MD5_TEXT_EDIT);
+			if (NULL != _hMD5TextEdit)
+			{
+				::SetWindowLongPtr(_hMD5TextEdit, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(this));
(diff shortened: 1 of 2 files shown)


PowerEditor/src/MISC/md5/md5Dlgs.cpp
```diff
@@ -255,6 +262,27 @@ INT_PTR CALLBACK MD5FromTextDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
 
+LRESULT MD5FromTextDlg::MD5TextEditRunProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+	switch (message)
+	{
+		case WM_CHAR:
+		{
+			// Select All if Ctrl + A
+			if (wParam == 1)
+			{
```
donho: It seems wired to me that wParam value being 1 means `Ctrl + A`. Any online document of MSDN about that?

donho:
@luisffranca  Could you provide the document of MSDN to explain (wParam == 1) correspond to "Ctrl + A" ?

PowerEditor/src/MISC/md5/md5Dlgs.h
```diff
@@ -44,5 +44,15 @@ public :
...
 	virtual INT_PTR CALLBACK run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam);
+
+	LRESULT MD5TextEditRunProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam);
+	static LRESULT CALLBACK MD5TextEditStaticProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam) 
+	{
+		return (((MD5FromTextDlg *)(::GetWindowLongPtr(hwnd, GWLP_USERDATA)))->MD5TextEditRunProc(hwnd, message, wParam, lParam));
+	};
+
+private :
+	HWND _hMD5TextEdit = nullptr;
```
donho: We don't need to have a member here.

donho:
See https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6034

Issue #3849, opened 2017-10-26
Character Count wrong


buggalicious:
### Description of the Issue

The character count in the summary reads it is counted without space, while in fact it also does count the spaces.

### Steps to Reproduce the Issue
1. Type 123 123
2. Double click the word "length" on the status bar or go to View/Summary
3. Read the first line of the summary "Character: (without blanks) and see the number "7"

### Expected Behavior
The number should be 6 without the white space

### Actual Behavior
The number is 7, because the blank space is counted despite the preceding statement describing the result

### Debug Information
none
The plugin TextFX/Tools/Word Count displays the correct values.
Besides, it would be great if one could have the actual value of "Character: (without blanks)" displayed on the status bar.

Sinhov:
Hello, new contributor here. I want to work on this issue. Could someone please guide me on how to get started? I don't know where to look for in the codebase to fix this issue.
Thanks.

sasumner:
It appears that the "Characters (without blanks)" value being reported is actually "Characters (without line-endings)".

**Proof** (for a document with multiple lines):
**Step 1:** Do a regular expression COUNT search on   (?s).   and note the value is same as "Current document length" value
**Step 2:**  Do a regular expression COUNT search on   (?-s).   and note the value is same as "Characters (without blanks)" value
The difference in the searches is the Step1 search will count line-ending characters and the Step2 will exclude line-ending characters

sasumner:
@Sinhov 
I would start by doing a "find-in-files" on the codebase for some text that appears in the View (menu) -> Summary popup window.  For example, search for `(without blanks):`.  This would get you quickly to the area of code that does this processing and you can study it and decide how to change it.  Good luck.

buggalicious:
Thank you, guys, for your interest in this matter. Is there any update on this?
I think this is something that should be correct as to the way the software handles it, it is part of it's reliability.
Best regards.

sasumner:
Sometimes changes are slow to be adopted.  It is possible to obtain your desired count by doing a Find-Count in regular expression search mode for `\S`:
![image](https://user-images.githubusercontent.com/30118311/32747723-a645aae4-c887-11e7-8fe3-bdea34c4bf84.png)

It is also possible to script a solution and have a running count of non-blank characters in the status bar.

But these are workarounds, not cures.

encore0117:
Working on this issue right now

encore0117:
I just split the char count into 2 cases.

[...]

PR #3907, opened 2017-11-21
Split the character counting into two cases based on encoding allowin…

encore0117:
For the issue #3849 
Split char counting into 2 cases to have counting without whitespaces.
For UTF-8, add one more measure in summary to have count without white spaces.
For other encodings, it remains the same.

Using UTF-8 encoding
first

Using ANSI encoding
second


donho:
Nothing to be fixed here.

PR #3908, opened 2017-11-22
Abort highlight search if the selection crosses a line boundry

klusark:
When selecting more, it becomes uselessly slow.

I don't think this is the best way to do it as it has a magic number, so treat it as more of a proof on concept.

Any ideas how to improve it would be appreciated.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/SmartHighlighter.cpp
+++ b/PowerEditor/src/ScitillaComponent/SmartHighlighter.cpp
@@ -127,6 +127,7 @@ void SmartHighlighter::highlightView(ScintillaEditView * pHighlightView, Scintil
 
 	auto curPos = pHighlightView->execute(SCI_GETCURRENTPOS);
 	auto range = pHighlightView->getSelection();
+	int textlen = range.cpMax - range.cpMin + 1;
 
 	// Determine mode for SmartHighlighting
 	bool isWordOnly = true;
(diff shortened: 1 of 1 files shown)


dail8859:
> When selecting more, it becomes uselessly slow.

Define "uselessly slow" because it works just fine for me even when selecting a nearly 2000 character word. Having a limit probably isn't a horrible idea but apparently I'm not running into the same issue you are.

klusark:
Sorry, I should have clarified that. 1024 is probably way too low a limit. My use case is opening a 10mb log file and selecting half to delete. If match word only is off it takes a few seconds.

dail8859:
Oh gotcha. Personally I think a limit like that is fine since the smart hilighting is meant to provide feedback to the user to quickly see other instances of selected text. If the chunk of text is large then it really doesn't provide much meaning as you probably won't be able to see other instances of the match. Though this is just my opinion.

As for the actual patch, see https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#do-not-use-java-like-braces

klusark:
I reworked it to just abort if the user selects multiple lines, since that already doesn't work during the search.

donho:
@klusark 
> I reworked it to just abort if the user selects multiple lines, since that already doesn't work during the search.

I didn't see it in your PR.

klusark:
@donho I'm not sure what you mean. lines 161 and 162 do exactly that. Should I add a comment to make it more clear?

donho:
@klusark Sorry I misunderstood your "selects multiple lines" by "multi-line selection & edition":
https://notepad-plus-plus.org/features/multi-editing.html

OTOH, in your PR, if you do select the last word of the 1st line and the first word of the 2nd line, the selected length will be smaller than the 2nd line length, and the smart highlight will still be performed.

PR #3911, opened 2017-11-22
I've done TODO and fixed issue #3910

### LINKED ISSUE #3910: FunctionList (classRange parser?) - single line comment is broken
Notepad v. 7.5.2. (32 bit)
Windows 7 SP1 Professional

Having Java as the selected language, the function list correctly parses this java class :

```
public class Car
{
	public int doSomething (int num)
	{
		return 1;
  	}
}
// something must be here ;)
```
add one commented out line now:

```
public class Car
{
//}  -- this line breaks the function list parser
	public int doSomething (int num)
	{
		return 1;
  	}
}
// something must be here ;)
```
### Expected Behavior
line `//}  -- this line breaks the function list parser` should be ignored

### Actual Behavior
the function list is broken.

This behaviour can be observed with other classRange based languages, with or as well without  open/closeSymbole in the classRange definition.

operasfantom:
fixed #3910

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2499,7 +2499,7 @@ int Notepad_plus::findMachedBracePos(size_t startPos, size_t endPos, char target
 	if (startPos > endPos) // backward
 	{
 		int balance = 0;
-		for (int i = int(startPos); i >= int(endPos); --i)
+		for (int i = static_cast<int>(startPos); i >= static_cast<int>(endPos); --i)
 		{
 			char aChar = static_cast<char>(_pEditView->execute(SCI_GETCHARAT, i));
 			if (aChar == targetSymbol)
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2516,6 +2516,20 @@ int Notepad_plus::findMachedBracePos(size_t startPos, size_t endPos, char target
 	}
 	else // forward - TODO
 	{
+		int balance = 0;
+		for (int i = (int)startPos; i <= (int)endPos; ++i) {
```
MAPJe71: 1. for type conversion use C++-style as stated in the [Contributing Guideline](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md)
2. or change to `for (int i = int(startPos); i <= int(endPos); ++i) {` like in `backward` branch.

scintilla/lexers/LexCPP.cxx
```diff
@@ -1327,10 +1327,13 @@ void SCI_METHOD LexerCPP::Fold(unsigned int startPos, int length, int initStyle,
...
 					char chNext2 = styler.SafeGetCharAt(i + 2);
-					if (chNext2 == '{') {
-						levelNext++;
-					} else if (chNext2 == '}') {
-						levelNext--;
+					if (!inLineComment) { //bracket not in singleline comment
+						if (chNext2 == '{') {
+							levelNext++;
+						}
+						else if (chNext2 == '}') {
```
MAPJe71: Don't change coding style (as stated in [Guideline](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests)) i.e. write lines 1333 and 1334 as: `} else if (chNext2 == '}') {`.

donho:
I have no idea about the fix in you PR, and IMO, there's nothing to be fixed in this file. 
The bug of #3910 is in Notepad++ but not in Scintilla.

PR #3917, opened 2017-11-25
Fix #3861 - Automplete detecting unwanted symbols

### LINKED ISSUE #3861: Automplete detecting unwanted symbols
### Description of the Issue
The autocomplete detects symbols it should not detect as variables in .js files. This has multiple times caused me to get random symbols in my code, that I did not intend to be there.

1. Make a .js file.
2. Write the following:
// somevar}
somev
3. "somevar" should autocomplete to "somevar}"

### Expected Behavior
I expected the autocomplete to omit the "}" symbol.

### Actual Behavior
The autocomplete did not omit the "}" symbol.

### Debug Information
Notepad++ v7.5.1   (64-bit)
Build time : Aug 29 2017 - 02:38:44
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)

Problem:
slett

annbumagina:
fixed #3861

### DIFF
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -135,7 +135,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	generic_string expr(TEXT("\\<"));
 	expr += beginChars;
-	expr += TEXT("[^ \\t\\n\\r.,;:\"()=<>'+!\\[\\]]+");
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -135,7 +135,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	generic_string expr(TEXT("\\<"));
 	expr += beginChars;
-	expr += TEXT("[^ \\t\\n\\r.,;:\"()=<>'+!\\[\\]]+");
+	expr += TEXT("[^ \\t\\n\\r.,;:\"(){}#$%&*-@^�|`~\\\\=<>'+!\\[\\]]+");
```
donho: I would exclude only {}.

PR #3918, opened 2017-11-25
cmake config update

chcg:
- update of cmake config to reflect the vs proj changes since the last update

### DIFF
--- a/PowerEditor/src/CMakeLists.txt
+++ b/PowerEditor/src/CMakeLists.txt
@@ -53,16 +53,23 @@ SET(projIncludDir	../../scintilla/include/
 					./WinControls/FunctionList/
 					./uchardet/
 					./WinControls/FileBrowser/
-					./WinControls/ReadDirectoryChanges/)
+					./WinControls/ReadDirectoryChanges/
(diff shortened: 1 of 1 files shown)


chcg:
See https://ci.appveyor.com/project/chcg/notepad-plus-plus/build/job/6w7a37kn44ojo0y1 for a testbuild with cmake. The mingw variant needs the changes from #4145 to have a chance to complete successfully.

donho:
@chcg Since vcproj files and makefile are maintained now, what's the point of CMakeList.txt ?

chcg:
By usage of cmake you could created make files for a lot of more build systems than just current MS vsproject or mingw make. You could generate vsproj files for different VS versions, jom, nmake, make, ninja, etc, see https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html. 

Additionally it would be possible to add a packaging step via the cmake part cpack (https://cmake.org/cmake/help/latest/manual/cpack-generators.7.html#manual:cpack-generators(7))

But in the end that all just makes sense if it is maintained at least a little bit.

PR #3930, opened 2017-11-27
Updated multiline comments in Cmake. Now they are correctly highliting.

Motyaspr:


dail8859:
These changes need submitted to the [Scintilla project](https://sourceforge.net/p/scintilla/bugs/) as they maintain all the lexer code. It is also possible this bug is fixed in their codebase.

donho:
CMake lexer is part of Scintilla project.
Please do you PR in http://www.scintilla.org/

PR #3942, opened 2017-11-30
Regression: Fixed multiple issues

SinghRajenM:
This PR fixes below issue
  **1. Nsis wanring**
&nbsp;&nbsp;&nbsp;&nbsp; Fix: It will provide estimated install size on component select page.

**2. Themes files installation is broken in V7.5.2**
&nbsp;&nbsp;&nbsp;&nbsp; Root cause: Variable $UPDATE_PATH was not before calling themes.nsh.

**3. Localization files installation is broken in V7.5.2**
&nbsp;&nbsp;&nbsp;&nbsp; Root cause: In "%temp%/nst143F.tmp" (nst143F.tmp is not fixed, it is created randomly by nsis), nppLocalization folder is not created. I could not figure out why. But yes, once installation is done (on finish dialog, before clicking finish), this folder appears again in "%temp%/nst143F.tmp".

P.S.  Could not reopen PR #3750, so opening new PR.

PR #3943, opened 2017-11-30
Fixed Dspell issues

SinghRajenM:
Issue:
1. DspellChecker is included V7.5.2, but it is not working. Because default the dictionary path should be "%appdata%/Notepad++/plugins/config/Hunspell". Referred V6.7.8 script. 

2. To install additional language (dictionary), npp must run as admin to add these files to "%programfiles/Notepad/plugins/config/Hunspell". Therefore, better to use %adddata% instead of INSTALLDIR.

3. Installed files are not removed on uninstallation.

Updated mainSectionFuncs.nsh, just to remove unreferenced label warning.

**Please note: This PR will work after PR #3942, Because $UPDATE_PATH will be available after that only. So PR #3942 is must for this PR.**

### DIFF
--- a/PowerEditor/bin/change.log
+++ b/PowerEditor/bin/change.log
@@ -1,3 +1,10 @@
+Notepad++ 7.5.3 bug-fixes:
+
+1.  Fix shell extension registration failure in installer.
+2.  Fix theme files installation failure in installer. 
+3.  Fix DSpellCheck incomplete installation in installer.
+
+
 Notepad++ 7.5.2 new features/enhancements & bug-fixes:
 
 1.  Fixed hanging issue while modifying JavaScript TAB settings.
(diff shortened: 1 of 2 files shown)


SinghRajenM:
God knows why it got closed automatically while resolving conflict.

dail8859:
@SinghRajenM Looks like it was force pushed and contains 0 commits.

SinghRajenM:
May be. I executed below commands.
```
git reset --hard origin/master
git commit -a
git push --force nppSingh nsis_DspellCheck_20171130
```

After that It has gone. Anyway, this is already fixed by @donho in except point 2.

SinghRajenM:
May be. I executed below commands.
```
git reset --hard origin/master
git commit -a
git push --force nppSingh nsis_DspellCheck_20171130
```

After that It has gone. Anyway, this is already fixed by @donho in except point 2.

PR #3958, opened 2017-12-06
Keep existing context menu text (Edit with Notepad++)

mlocati:
I use the keyboard a lot, and the `n` accelerator key is already taken by other entries in the Explorer context menu.
So, every time I install/update Notepad++, I change the `Edit with &Notepad++` to `Edit with Notepad&++`.

What about keeping the existing context menu entry if it already exists?


mlocati:
Closing because of lack of interest from the repo owner

sasumner:
>So, every time I install/update Notepad++, I change the Edit with &Notepad++ to Edit with Notepad&++

Is it possible in the installer to deselect updating the context menu?  If so, then perhaps deselecting that option will leave whatever is there alone, meaning that it won't remove it or touch it at all, leaving your previous modification alone.  Just a guess...

PR #3992, opened 2017-12-11
Fix broken indices in EncodingMapper

### LINKED ISSUE #3983: OEM 737 encoding broken from version 7.5.2
### Encoding to Greek OEM 737 character set is broken
Up until version 7.5.1, the display conversion to OEM 737 (Greek DOS codepage) worked fine. From version 7.5.2 this conversion function is broken.

1. Open a file that contains Greek written from a DOS window (copy con ....) in Notepad++ 7.5.2+.
2. Click on Encoding -> Character Sets -> Greek -> OEM 737.
3. You should see the Greek characters displayed, but instead you see another character set.

### Debug Information
Notepad++ v7.5.3   (64-bit)
Build time : Dec  5 2017 - 10:47:54
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

### LINKED ISSUE #3991: Character encodings are broken
### Description of the Issue
The heart of the issue lies in the way the menu item IDS interact with the EncodingMapper. IE a constant is subtracted from the ID and used as an index into the encoding map
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e339c0f600295267db7a8df01598d1cdb8f1c5cc/PowerEditor/src/NppCommands.cpp#L2174-L2177

However with the recent removal of a number of unused character encodings (60f1137), causes this to breakdown. the command IDs remain the same, however the indexes in the Encoding map have changed. This means that any encoding with a command id greater than the first deleted encoding, no longer behaves as expected.

dnakamura:
Fixes #3983
Fixes #3991

### DIFF
--- a/PowerEditor/src/EncodingMapper.cpp
+++ b/PowerEditor/src/EncodingMapper.cpp
@@ -50,12 +50,12 @@ static EncodingUnit encodings[] = {
     {28597, "greek ISO_8859-7 ISO-8859-7 csISOLatinGreek greek8 iso-ir-126 ELOT_928 ECMA-118"}, //IDM_FORMAT_ISO_8859_7 
     {28598, "hebrew ISO_8859-8 ISO-8859-8 csISOLatinHebrew iso-ir-138"},                        //IDM_FORMAT_ISO_8859_8 
     {28599, "latin5 ISO_8859-9 ISO-8859-9 csISOLatin5 iso-ir-148 l5"},                          //IDM_FORMAT_ISO_8859_9 
-    //{28600, ""/*"latin6 ISO_8859-10 ISO-8859-10 csISOLatin6 iso-ir-157 l6"*/},                //IDM_FORMAT_ISO_8859_10
-    //{28601, ""/*"ISO_8859-11 ISO-8859-11"*/},                                                 //IDM_FORMAT_ISO_8859_11
+    {-1,    ""/*"latin6 ISO_8859-10 ISO-8859-10 csISOLatin6 iso-ir-157 l6"*/},                  //IDM_FORMAT_ISO_8859_10
+    {-1,    ""/*"ISO_8859-11 ISO-8859-11"*/},                                                   //IDM_FORMAT_ISO_8859_11
     {28603, "ISO_8859-13 ISO-8859-13"},                                                         //IDM_FORMAT_ISO_8859_13
     {28604, "iso-celtic latin8 ISO_8859-14 ISO-8859-14 18 iso-ir-199"},                         //IDM_FORMAT_ISO_8859_14
     {28605, "Latin-9 ISO_8859-15 ISO-8859-15"},                                                 //IDM_FORMAT_ISO_8859_15
-    //{28606, ""/*"latin10 ISO_8859-16 ISO-8859-16 110 iso-ir-226"*/},                          //IDM_FORMAT_ISO_8859_16
+    {-1,    ""/*"latin10 ISO_8859-16 ISO-8859-16 110 iso-ir-226"*/},                            //IDM_FORMAT_ISO_8859_16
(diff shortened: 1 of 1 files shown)


SinghRajenM:
Oops!! That me who caused this issue. Even after thoroughly testing, this couldn't be caught.

However, I feel to avoid further confusion, updating encoding definition value in "menuCmdID.h" would be good deal (as commented items are no longer used.). 
Updating something like below - 
from
```
    #define    IDM_FORMAT_ISO_8859_9        (IDM_FORMAT_ENCODE + 17)
    //#define    IDM_FORMAT_ISO_8859_10     (IDM_FORMAT_ENCODE + 18)
    //#define    IDM_FORMAT_ISO_8859_11     (IDM_FORMAT_ENCODE + 19)
    #define    IDM_FORMAT_ISO_8859_13       (IDM_FORMAT_ENCODE + 20)
    #define    IDM_FORMAT_ISO_8859_14       (IDM_FORMAT_ENCODE + 21)
    #define    IDM_FORMAT_ISO_8859_15       (IDM_FORMAT_ENCODE + 22)
    //#define    IDM_FORMAT_ISO_8859_16     (IDM_FORMAT_ENCODE + 23)
    #define    IDM_FORMAT_DOS_437           (IDM_FORMAT_ENCODE + 24)
    #define    IDM_FORMAT_DOS_720           (IDM_FORMAT_ENCODE + 25)
```
to
```
    #define    IDM_FORMAT_ISO_8859_9        (IDM_FORMAT_ENCODE + 17)
    //#define    IDM_FORMAT_ISO_8859_10     (IDM_FORMAT_ENCODE + 18)
    //#define    IDM_FORMAT_ISO_8859_11     (IDM_FORMAT_ENCODE + 19)
    #define    IDM_FORMAT_ISO_8859_13       (IDM_FORMAT_ENCODE + 18)
    #define    IDM_FORMAT_ISO_8859_14       (IDM_FORMAT_ENCODE + 19)
    #define    IDM_FORMAT_ISO_8859_15       (IDM_FORMAT_ENCODE + 20)
    //#define    IDM_FORMAT_ISO_8859_16     (IDM_FORMAT_ENCODE + 23)
    #define    IDM_FORMAT_DOS_437           (IDM_FORMAT_ENCODE + 21)
    #define    IDM_FORMAT_DOS_720           (IDM_FORMAT_ENCODE + 22)
```

dail8859:
I don't think it is a good idea to change `IDM_xxx` constants. If a plugin uses these then it will break their usage of calling these menu commands.

dnakamura:
I think the best solution is to probable modify the encoding mapper to map based on the `IDM_xxx` constants rather than an index.  Yes it would make lookups more expensive, but its not like the user will be changing encodings frequently

dinkumoil:
@SinghRajenM 

Currently my [AutoCodepage plugin](https://sourceforge.net/projects/autocodepage/) is broken because of your changes in v7.5.2. Updating encoding definition value in "menuCmdID.h" would break it permanently.

I think the fix of @dnakamura is the best solution.

dail8859:
I found this in another part of the code, would this need updated as well (I honestly have no clue but if the EncodingMapper is changed this looks like it would have to change too).

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/440577f6479adb0b8edecffea9ece54916dab8ab/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L48-L97

dinkumoil:
@dail8859 

Seems to be used for this list (screenshot taken from Npp v7.3.3):

grafik

The list gets filled by the for-loop starting in line 1159 of preferenceDlg.cpp. This loop uses the array `encodings` as source for the encodings to be shown in the combobox. That means it would be possible to remove some items of the `encodings` array to reduce the set of encodings shown in the combobox.

I would say preferenceDlg.cpp has **not** to be changed.

PR #4002, opened 2017-12-13
Fix highlighting of <script> tags in XML files

dail8859:
Issue reported on [the forum](https://notepad-plus-plus.org/community/topic/14913/colorization-bug-in-xml).

Using `<script>` tags in XML has incorrect highlighting:

before

Since the HTML and XML lexer are the same, it treats `<script>` tags specially since HTML can embed code. The HTML lexer has a `"lexer.xml.allow.scripts"` property that should be used for XML to keep it from trying to parse embedded code. See this section within the HTML Lexer:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b2c3e82ce7a7eaa11b777c4c8dea6453adde8b7a/scintilla/lexers/LexHTML.cxx#L670-L672

After setting this property it now highlights it correctly:

after


dail8859:
Busy for the next week or so. I want to do a bit of extra testing on this before merging.

dail8859:
Closed by c97b35c8150532f7050114deb6f267127f8c9d92

Issue #3998, opened 2017-12-12
auto-complete only works with API file, should work for builtin-languages in any case


GitMensch:
### Description of the Issue
If a language is supported (defined in langs[.model].xml) but no API\\\${lang}.xml is available the keywords given for the language should be used for auto-complete "function completion".

### Steps to Reproduce the Issue
1. Activate auto-completion with "function completion only"
2. Create a new file and choose COBOL highlighter (as there is no API shipped)
3. try to use auto-completion

### Expected Behavior
If no API\\\${lang}.xml is available is available but langs[.model].xml defines the current language the entries under `Keywords` should be used for auto-completion.

### Actual Behavior
No function completion is done if API\\\${lang}.xml is missing.

### Debug Information
Notepad++ v7.5.3 (32-bit)
Build time : Dec 5 2017 - 10:45:46
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll DSpellCheck.dll LocationNavigate.dll mimeTools.dll NppConverter.dll NppExport.dll NppSaveAsAdmin.dll PluginManager.dll SourceCookifier.dll

donho:
Auto-completion list is not necessary the same with the keywords of syntax highlighting. In some languages like COBOL  it could be the case but in the others they may need some different standard libraries' API, which are not in the keyword list.

GitMensch:
> Auto-completion list is not necessary the same with the keywords of syntax highlighting. In some languages like COBOL it could be the case but in the others they may need some different standard libraries' API, which are not in the keyword list.

@donho I see. What do you think about using the keywords if API\${lang}.xml is missing?

GitMensch:
Ping @donho I completely understand you're very busy with this project alone, not even taking life into account. Just want to recheck "What do you think about using the keywords for the built-in languages if, and only if their API${lang}.xml is missing?" (this is what the issue is about, not about the case where an api file exists).

GitMensch:
Second thought: I vote to let the debugbuild open an msgbox if a builtin-language doesn't have an API file - this way the API file can simply be extracted from the keywords of the builtin-language and everything should be fine in the release builds, without the need (or a use) for a fallback.
Opinions?

PR #3393, opened 2017-06-13
FunctionList Update 3

MAPJe71:
- add parsers for ADA, AutoHotKey, Fortran Fixed and Free Form style (see Community Topics [#11059](https://notepad-plus-plus.org/community/topic/11059) & [#13553](https://notepad-plus-plus.org/community/topic/13553)), Haskell (see Community Topic [#12972](https://notepad-plus-plus.org/community/topic/12972)), Haskell Literate style, Haskell Literate style for LaTeX, MOX (see Community Topic [#13675](https://notepad-plus-plus.org/community/topic/13675)), PL/SQL (see Community Topic [#13393](https://notepad-plus-plus.org/community/topic/13393)), RAPID (see Community Topics [#10725](https://notepad-plus-plus.org/community/topic/10725) & [#12264](https://notepad-plus-plus.org/community/topic/12264))
 - update comment for parsers Assembly, AutoIt3, Batch, C, Java, KRL, NSIS, PowerShell and Sinumerik
 - update language ID table
 - improve the Bash parser for single-line, "Here Document" (type 1 & 2) and "Here String" comments
 - improve the INI parser by using \xnn notation for special characters
 - improve the Inno Setup parser by searching for procedures in addition to functions, discards forward declarations and searches in INI, ISPP and ISSI sections
 - improve the UniVerse BASIC parser by searching for functions and procedures in addition to numeric and alpha-numeric labels
 - improve the "Function list" XML parser for matching left and right quotes of function name filter

Issue #2731, opened 2016-12-30
Please update syntax highlighter for COBOL


GitMensch:
Some words are missing that can be found in the COBOL2014 standard and in different compiler's extensions. A not complete but quite nice list can be extracted from GnuCOBOL, see https://github.com/OpenCobolIDE/OpenCobolIDE/issues/389 for a list of words.

GitMensch:
necessary changes done, see https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4004 (note: the [original SciTE-file cobol.properties was updated](https://sourceforge.net/p/scintilla/scite/ci/ca90a391a4851d70e44ec49df9460bc304a08fba/), beforehand)

Additional note: I don't know why but the two reserved words `value` and `values` are in but not highlighted any more - as much more words are highlighted that were missing before I still consider it "ok"

GitMensch:
closed with finish of #4004 with a5ec6be

PR #4004, opened 2017-12-14
heavy COBOL update, first support within function list and auto-completion

GitMensch:
COBOL support in function list - new entries in functionList.xml (note: defaults to fixed-form reference format, manual switch to free-form reference-format possible in association for langID 50)

COBOL support in auto-completion - new file APIs/cobol.xml (note: defaults to upper-case, hopefully changed to support both, see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3998)
COBOL syntax highlighter - recreated list with many more reserved words

note: list of statements, functions and reserved words token from GnuCOBOL 3.0 (COBOL85,2002,2014 + extensions from IBM/MicroFocus/RM/ACUCOBOL), please report back if any words are missing


GitMensch:
Note: While the auto-complete works for the *keyword* [`FORMATTED-DATE`](https://github.com/GitMensch/notepad-plus-plus/blob/d53f6c596dfc90e110ae9ef9957807eaf2800e09/PowerEditor/installer/APIs/cobol.xml#L415) it doesn't work for the *function* parameters. Is there anything I can do to fix this or should I file a bug?

chcg:
@MAPJe71 Maybe interesting for  #3393

dail8859:
@GitMensch 

I have experienced the opposite when using `additionalWordChar`, for example using `.` for Lua. Using `Ctrl+Space` does not suggest words with the character already in them (see #2532), however opening the function (e.g. with `(`) works correctly. See the gif below. After it auto-completes the full function I delete a few chars and try to trigger the auto complete again but it doesn't work, however if I open the function it works correctly. I did not try any changes in this PR.


acdemo

PowerEditor/src/functionList.xml
```diff
@@ -90,6 +90,11 @@
 			<association id=  "powershell_function"   langID="53"                          />
 			<association id=  "javascript_function"   langID="58"                          />
 			<association id=       "baanc_section"    langID="60"                          />
+			<!--
+			   As there is currently only one langID for COBOL: change to cobol_section_free
+			   if this is your favourite format
+			-->
+			<association langID="50" id="cobol_section_fixed"/>
```
MAPJe71: Use same layout as is used for already available associations and position based on `langID` value i.e. between "innosetup_syntax" and "powershell_function" associations.

PowerEditor/src/functionList.xml
```diff
@@ -1468,6 +1473,28 @@
...
+			<!-- Variant for COBOL fixed-form reference format -->
+			<parser id="cobol_section_fixed" displayName="COBOL fixed-form reference format">
+				<function
+					mainExpr="(?m-s)^.{6}[ D][\t ]{0,3}(?!exit\s)[\w_-]+(\.|((?'seps'([\t ]|\*&gt;.*|([\n\r]+(.{6}([ D]|\*.*)|.{0,6}$)))+)section(\.|((?&amp;seps)(\.|[\w_-]+\.)))))"
+					displayMode="$functionName">
+					<functionName>
+						<nameExpr expr="[\w_-]+((?=\.)|((?'seps'([\t ]|\*&gt;.*|([\n\r]+(.{6}([ D]|\*.*)|.{0,6}$)))+)section((?=\.)|(?&amp;seps)((?=\.)|[\w_-]+(?=\.)))))"/>
+					</functionName>
+				</function>
+			</parser>
```
MAPJe71: 1. Use same layout as is used for existing parsers.
2. Remove `displayMode` attribute, it's not used.
3. Add comment/description.

PowerEditor/src/functionList.xml
```diff
@@ -1468,6 +1473,28 @@
...
+			<!-- Variant for COBOL free-form reference format -->
+			<parser id="cobol_section_free" displayName="COBOL free-form reference format">
+				<function
+					mainExpr="[\s\.](?!exit\s)[\w_-]+\s+section(\s*|(\s+[\w_-]+)?)(?=\.)"
+					displayMode="$functionName">
+					<functionName>
+						<nameExpr expr="[\w_-]+\s*section"/>
+				   </functionName>
+				</function>
+			</parser>
```
MAPJe71: See review comment for fixed-form parser.

GitMensch:
@MAPJe71 Applied most of your comments. Note: We tweaked these entries back in 2015 together, I would actually need time to fully understand and explain these regexes. But as the functionList still doesn't support COBOL at all I consider it better to use these as they are than waiting for me to come back to them in 2018/2019. Note: I don't know if the old comment boundaries problem is fixed by now (there was a request for you to redo your patch back then), if they are the commented one for free-form reference-format may be retested.

... just rechecked:
https://svn.code.sf.net/p/open-cobol/contrib/trunk/samples/prime_machine/prime_machine.cob works fine with the "simple" cobol_section_free variant which is active in the suggested PR but not with the extended one including the paragraphs.

GitMensch:
@donho The "common style issues" mentioned by @MAPJe71 were fixed, all CI and manual checks passed and #2731 is fixed by this - is there anything "open" on your side before including this one?

PR #4041, opened 2018-01-03
Updated some strings in romanian.xml

TigerxWood:


donho:
Is there any Romanian speaker can validate this PR?

Dambuster88:
No, it is not ok. Not all the changes.
Some of the verbs, like "Open" and "Close" are at the long infinitive form and sounds very bad.
https://en.wikipedia.org/wiki/Infinitive#Latin_and_Romance_languages

Dambuster88:
Check lines 911, 912, 914 and 919.
Why space before question mark?

simioff:
I'm not sure about the spaces before the question marks, but the rest looks good. At least other tools like Microsoft Office are using the same wording.

TigerxWood:
Hi guys, maybe you didn't notice but I redid the entire translation of this program. This last changes was some minor mistakes that I missed when I proofread it. This is the best translation you could have, instead of the old one that was imposible to use. Even this is not 100% according to romanian language rules. For example I didn't translate words like Clipboard or Cloud because I considered that because this program is used mainly by programmers they will understand better this english version instead of some crappy translation that noone could guess what it is, like Clipplanșetă for Clipboard.
About that infinitive stuff this is the right way to translate, for the consisystency with the rest of the translation: Salvare, Tipărire, etc.
About the spaces before question marks it was not intentional. I corrected them.

donho:
@TigerxWood  Could you check the newest updated english.xml here:

pfparesseux000:
In the file english.xml, because of issue 4040, items id="45005, id="45008, id="45010" and id="45011" changed. In the file romanian.xml, the sentences seem confused. These sentences should be updated too.

PR #4043, opened 2018-01-04
update japanese.xml to v7.5.4

yasmise:
* Replace '\r' by real carriage return
* Fix some translations for clear japanese


Tmp341:
@donho While you're on this, please change Turkish.xml version from v7.3.1 to v7.5.4, because it looks way too old.

Issue #4008, opened 2017-12-17
Error using vs2013 compile


softmgr:
### Description of the Issue
After compiling SciLexer.dll, I open the PowerEditor\visual.net\notepadPlus.vs2013.vcxproj, and then compile the project. ---- Json.hpp file error.

SinghRajenM:
Json.hpp support many new features (C++11 and C++14) and few of them are not supported in VS2013. Now you need VS2015 or VS2017 to compile Notepad++.

softmgr:
@SinghRajenM 
So please remove the file "PowerEditor\visual.net\notepadPlus.vs2013.vcxproj, so as not to cause 
misunderstandings, thank you!

chcg:
notepadPlus.vs2013.vcxproj removed with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d14cb43e9d3c58d02af97149cd573c81bd862c47#diff-91f7886abd2e7f21c7fa79557be379c9555d51a8df7353811ec7d1476e04b261

and VS2017 is required, see https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/BUILD.md

PR #4045, opened 2018-01-04
Update ReadMe and Remove V2013 project

SinghRajenM:
1. Update readme.md
Now Notepad++ is built with VS2015 and **can't be built with VS2013** because JSON.hpp needs VS2015 or higher. So updated ReadMe, to reflect the same.

2. VS2013 is no longer needed
As Notepad can't be built with VS2013 anymore and to avoid confusion, VS2013 should be removed.

Reported Bug: #4008


donho:
@SinghRajenM Have you tried to build Scintilla with PCRE support under VS2015 ?

SinghRajenM:
@donho No, I just tried without PCRE support and it worked fine. Now, I downloaded boost1.55 and it looks like VS2105 is not supported for Boost1.55. So let's drop idea to update readme. We can update it later when we move to latest scintilla.

donho:
Please reopen it when boost support VS2015

CookiePLMonster:
VS2015 is supported since a while - but boost 1.55.0 is too old to support it. 1.66.0 supports even VS2017.

PR #4053, opened 2018-01-06
avoid crash of #2727

chcg:
- fix of buffer overrun of issue #2727 with long UDL ext field
- PR just fixes the crash, but restricts the possible amount of extensions, also it is not immediately visible in the GUI, just after saving the udl on N++ shutdown and restart
- Unclear why MAX_PATH is needed as limitation for the complete ext field.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -652,7 +652,7 @@ typedef void (WINAPI *PGNSI)(LPSYSTEM_INFO);
 void cutString(const TCHAR* str2cut, vector<generic_string>& patternVect)
 {
 	TCHAR str2scan[MAX_PATH];
-	lstrcpy(str2scan, str2cut);
+	lstrcpyn(str2scan, str2cut, MAX_PATH);
 	size_t len = lstrlen(str2scan);
 	bool isProcessing = false;
 	TCHAR *pBegin = nullptr;


CookiePLMonster:
MAX_PATH in this place looks wrong and completely arbitrary. You'd probably be way better off either:

1. using a string class (whatever n++ prefers, looks like `generic_string` is the way to go)
2. doing strlen and allocating memory dynamically

I'd go with 1. - this way you'd also avoid using `lstr***` functions which are incredibly nasty.

chcg:
@CookiePLMonster Yes this is a reimpl. Just didn't want to introduce big code changes without the ok from somebody who understands why it was done this way. See https://msdn.microsoft.com/en-US/library/2c8d19sb.aspx, _tcstok is probably relevant and I'm not sure if compatibilty with Windows Runtime is a problem.

CookiePLMonster:
That's right, though this function is simple enough and it's probably easy to assure no new behaviour is introduced by rewriting it.

CookiePLMonster:
I have investigated the issue myself now and it seems pretty obvious that this MAX_PATH is arbitrary - so I'd say it's best to rewrite this function to range-construct strings for `patternVect` instead of tokenizing the array. This way you don't copy contents on the stack and thus remove a limit easily!

It also feels like this function should be returning a vector by value but that'd be a micro-optimization not worth the trouble.

chcg:
Close this one in favor of #4073

PR #4096, opened 2018-01-17
Shortcut Mapper filter

cmeriaux:
Add in the shortcut mapper a filter entry


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -769,6 +769,12 @@ generic_string stringToUpper(generic_string strToConvert)
     return strToConvert;
 }
 
+generic_string stringToLower(generic_string strToConvert)
+{
+    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), ::tolower);
```
MAPJe71: AFAIK the project is build with both `UNICODE` and `_UNICODE` defined.
As `generic_string` is defined as `std::basic_string<TCHAR>`, will make it a wide character string.
So shouldn't it be `::towlower` ?

rillig: Just to be on the safe side, always cast the argument to `::tolower` to `unsigned char`, and likewise the argument to `::towlower` to `wchar_t`. Otherwise the behavior is undefined. (ISO C99, 7.25.1p5)

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -124,123 +122,176 @@ void ShortcutMapper::initBabyGrid() {
 	_babygrid.setHighlightColorProtectNoFocus(RGB(230,194,190));
 }
 
+generic_string ShortcutMapper::getTextFromCombo(HWND hCombo)
+{
+	TCHAR str[128];
+	::SendMessage(hCombo, WM_GETTEXT, 127, reinterpret_cast<LPARAM>(str));
```
rillig: There's no point in asking for only 127 characters here. If you don't trust the Windows API, you must initialize `str[127]` to the null character before calling `SendMessage`. The `WM_GETTEXT` [documentation](https://msdn.microsoft.com/de-de/library/windows/desktop/ms632627) doesn't explicitly mention whether it copies the null character, so a super-safe bet would be to fill the complete buffer with zeroes before calling `SendMessage`.

Regarding the abstraction level: do you really need to work with `WM_GETTEXT` here? There should be a generic function that just takes a window handle and maybe a maximum string size and returns a string. This function would be useful in lots of other places and would keep this code really nice and short.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -124,123 +122,176 @@ void ShortcutMapper::initBabyGrid() {
...
+	case STATE_USER: {
+		nbItems = nppParam->getUserCommandList().size();
+		_babygrid.setLineColNumber(nbItems, 2);
+		break; }
+	case STATE_PLUGIN: {
+		nbItems = nppParam->getPluginCommandList().size();
+		_babygrid.setLineColNumber(nbItems, 3);
+		_babygrid.setText(0, 3, TEXT("Plugin"));
+		break; }
+	case STATE_SCINTILLA: {
```
rillig: This change of indentation doesn't belong in this commit. Plus, it makes the code layout inconsistent. See the `switch` statement immediately below, which also doesn't have a space.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -124,123 +122,176 @@ void ShortcutMapper::initBabyGrid() {
...
+	generic_string res(str);
+	return stringToLower(res);
+};
+
+bool ShortcutMapper::isFilterValid(Shortcut sc)
+{
+	bool match;
+	generic_string shortcut_name = stringToLower(generic_string(sc.getName()));
+	match = false;
+	if (_shortcutFilter[0] == TEXT('\0')){
```
rillig: The [TEXT macro](https://msdn.microsoft.com/de-de/library/windows/desktop/dd374074%28v=vs.85%29.aspx) is only defined for strings, not for individual characters. You should get a compiler warning here and the program should crash immediately.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -124,123 +122,176 @@ void ShortcutMapper::initBabyGrid() {
...
+				if (isFilterValid(cshortcuts[i])) {
+					_babygrid.setText(cs_index, 1, cshortcuts[i].getName());
+					if (cshortcuts[i].isEnabled()) //avoid empty strings for better performance
+						_babygrid.setText(cs_index, 2, cshortcuts[i].toString().c_str());
+	
+					if (isMarker)
+						isMarker = _babygrid.setMarker(false);
+					_shortcutIndex.push_back(i);
+					cs_index++;
+				}
```
rillig: This is quite a lot of code duplication. Is there a way to avoid this?

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
MAPJe71: > generic function

i.e. control APIs in `windowsx.h`.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -121,123 +122,176 @@ void ShortcutMapper::initBabyGrid() {
...
+	if (_shortcutFilter[0] == TEXT('\0')){
+		match = true;
+	}
+	// test the filter on the shortcut name
+	size_t match_pos = shortcut_name.find(_shortcutFilter);
+	if (match_pos != std::string::npos){
+		match = true;
+	}
+	return match;
+}
```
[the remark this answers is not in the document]
MAPJe71: or even better:
```cpp
if (_shortcutFilter.empty()) {
    return true;
}
```

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
rillig: @MAPJe71 Yes, I'm sure that the `TEXT` macro is only defined for strings. That's what the documentation says.

By the way, the same macro can be used as follows:

    TEXT(oadLibrary)("user32.dll")

It works, but is obviously not correct.

MAPJe71: @rillig The documentation does not say _only_ for strings.
I've been using it for years based on (old) MS documentation for both strings and characters (see also comment in `tchar.h` code snippet above) and never had a compiler warning (only used MSVS compilers).
What compiler gives the warning and what is the code?

rillig: @MAPJe71 There's no compiler warning I know of. I just thought that based on the definition of the macro in the MSDN there should be a type conflict, since the definition there says `void TEXT(LPTSTR string)`. But since it is a macro and not a function, that though of mine was wrong.

rillig: @MAPJe71 Just to be pedantic, the MSDN documentation clearly says _only for strings_, repeatedly:

* Identifies a **string** as Unicode
* Otherwise, the macro identifies a **string**
* void TEXT(**LPTSTR** string)
* Pointer to the **string**

Just because it _apparently works_ for single characters too doesn't mean it's guaranteed; see my `oadLibrary` example, which also works but isn't guaranteed.

The documentation does not explicitly need to mention the word _only_ because otherwise this word would have to appear too often and reading the text would get boring.

MAPJe71: Depends on what (part of the) documentation one is reading ([String and Character Literals](https://docs.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp)).

rillig: @MAPJe71 The documentation you cite only talks about the `L` modifier. It doesn't talk about the `TEXT` macro. This distinction was my point of argument the whole time. I know that you can write `L'a'` to get a wide character but that in itself and the knowledge that the `TEXT` macro is currently implemented such that it produces this sequence doesn't make it valid code. The `TEXT` macro might also be implemented in a way that produces a compiler-specific string literal that is only valid for strings but not for characters. And then the program blows up. Hopefully during compilation and not during runtime.

MAPJe71: > The TEXT macro might also be implemented in a way that produces a compiler-specific string literal that is only valid for strings but not for characters.

True but any macro or function _might_ be implemented differently in the future.
The documentation _might_ not be up to date or incomplete. Although the same can be said about the source code of the library.

Do a search for `TEXT('` in the Microsoft header files and be supprised how often they use the `TEXT` macro for character literals, apparently contradicting their own documentation (e.g. `sddl.h`, `strsafe.h`).

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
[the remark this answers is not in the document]
MAPJe71: otherwise it should have been (the project is build with both UNICODE and _UNICODE defined):
```cpp
if (_shortcutFilter[0] == L'\0') {
    return true;
}
```

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
rillig: @MAPJe71 Ok, you won. That last paragraph is very convincing. :)

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
cmeriaux: done

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
cmeriaux: last year, I tried to factorize and avoid code duplication but that was too much code modifications. 
Sorry

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
cmeriaux: big thread ! 
So my code is good, isn't it ?

cmeriaux:
arrrrh I can't compile Npp anymore due to the VS migration. My VS 2013 can't compile 'json' files. I can't set VS  2015

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
rillig: @cmeriaux No, your code is too complicated. Since you are not really interested in the first character of the string, you shouldn't ask for it. The only thing you want to know is whether the string is empty. So just ask `_shortcutFilter.empty()`.

PowerEditor/src/MISC/Common/Common.cpp
cmeriaux: done

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
cmeriaux: a lot's a code use this syntax in NPP. I keep it.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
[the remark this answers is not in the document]
cmeriaux: done

cmeriaux:
last commit fixes errors with "conflict marker" and filtered items

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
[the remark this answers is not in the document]
MAPJe71: Compacted (nit-picking?):
```cpp
bool ShortcutMapper::isFilterValid(Shortcut sc)
{
	if (_shortcutFilter.empty()) 
	{
		return true;
	}
	// does the name match the filter (which is  stored in lower case)
	return (stringToLower(generic_string(sc.getName())).find(_shortcutFilter) != std::string::npos);
}
```
Allthough that last line may be a bit to compact for readability.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -121,11 +122,37 @@ void ShortcutMapper::initBabyGrid() {
 	_babygrid.setHighlightColorProtectNoFocus(RGB(230,194,190));
 }
 
+generic_string ShortcutMapper::getTextFromCombo(HWND hCombo)
+{
+	const int NB_MAX = 128;
+	TCHAR str[NB_MAX];
```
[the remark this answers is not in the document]
cmeriaux: done

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -121,11 +122,37 @@ void ShortcutMapper::initBabyGrid() {
...
+	::SendMessage(hCombo, WM_GETTEXT, NB_MAX, reinterpret_cast<LPARAM>(str));
+	generic_string res(str);
+	return stringToLower(res);
+};
+
+bool ShortcutMapper::isFilterValid(Shortcut sc)
+{
+	bool match;
+	generic_string shortcut_name = stringToLower(generic_string(sc.getName()));
+	match = false;
```
chcg: Why is this not done directly in L136?

cmeriaux: done

donho:
It's always good to add more description for your PR - to make better chance to be accepted.

cmeriaux:
when searching in the plugin view, the filter is also applied on the plugin name

cmeriaux:
@donho here is two screenshot that helps you to understand what does the PR do.
search_shortcut_name
search_plugin_name_sc_name

PR #4104, opened 2018-01-19
Added bunch of improvements to C# intellisense

MCCshreyas:
1. Added parameter suggestion for methods in System.Console class. 
2. Added overloaded methods for Console.WriteLine and Console.ReadLine method.

PR #4105, opened 2018-01-19
Corrected CoffeeScript file extension

MCCshreyas:
1. Changed extension to .coffee instead of .litcoffee.

### DIFF
--- a/PowerEditor/src/langs.model.xml
+++ b/PowerEditor/src/langs.model.xml
@@ -86,7 +86,7 @@
             <Keywords name="instre2">sr kr ar ksmps nchnls</Keywords>
             <Keywords name="type1"></Keywords>
         </Language>
-        <Language name="coffeescript" ext="litcoffee" commentLine="#" commentStart="" commentEnd="">
+        <Language name="coffeescript" ext="coffee litcoffee" commentLine="#" commentStart="" commentEnd="">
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
Quick Googling shows:

`Literate CoffeeScript
Besides being used as an ordinary programming language, CoffeeScript may also be written in “literate” mode. If you name your file with a .litcoffee extension, you can write it as a Markdown document — a document that also happens to be executable CoffeeScript code. `

Aren't both extensions valid then? In which case you should append .coffee instead of replacing it.

MCCshreyas:
Added that one too.

dail8859:
@MCCshreyas This really needs to be one commit. If you are unable to squash these commits, I can do it to this PR for you if you'd like instead of opening a new PR.

MCCshreyas:
Yes. Please squash those two commit into one. Can you please do that for this one to? https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4104

dail8859:
Commits have been squashed for this PR. #4104 is already a single commit.

MCCshreyas:
OK. Thanks for squashing commits!

dail8859:
Closed by e0c4011982a6c3572258b13f066fd0a710258937

Thanks for the PR

Issue #4112, opened 2018-01-22
Issue after saving Macro


bryant112:
### Description of the Issue
After using "Save Current Recorded Macro", menu shows both Start Recording and Start Recording options as enabled in Macro Menu.

### Steps to Reproduce the Issue

1. Start Recording macro
2. Create Macro (my specific macro related to find and replace)
3. Stop Recording macro once all steps are complete
4. Select "Save Current Recorded Macro" from Macro menu

### Expected Behavior

Menu should be back at the state as before you clicked "Start Recording" from clicked
### Actual Behavior

Menu shows both "Start Recording" and "Stop Recording" enabled
### Debug Information

Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

![macroerror](https://user-images.githubusercontent.com/7781469/35226368-a9e403e4-ff50-11e7-8af1-e4b6fc6ee481.jpg)

Issue #4114, opened 2018-01-23
Menu item's state is not maintained


SinghRajenM:
### Description of the Issue

Menu Item's state under many menus  such as "File", "Edit", "View" "macro" etc are not maintained if a recorded macro is saved.

### Steps to Reproduce the Issue

**Case 1:**
1.  Start Recording macro
2.  Create Macro (my specific macro related to find and replace)
3.  Stop Recording macro once all steps are complete
4.  Select "Save Current Recorded Macro" from Macro menu

**Case 2:**
1. Try to delete saved macro from "Macro ->Modify Shortcut/Delete Macro"-> select macro -> delete"
2. Close Shortcut mapper window after deleting a macro

**Case 3:**
1. Make any changes in "Shortcut Mapper" dialog such (e.g. Delete/Modify any shortcut under any tab)

### Expected Behavior

State of all the menu items should be maintained.

### Actual Behavior

Menu items states are not maintained.
File menu -> "Open in Default Viewer", "Rename", "Move to recycle bin", "Reload from Disk" etc. (Even for unsaved file).
Edit Menu -> "Clear Read only flag" (Even for unsaved file).
Search Menu -> "Go to Matching brace", "Select all between matching braces".
....many more under all the menu items.

### Debug Information

Notepad++ v7.5.4   (32-bit)
Build time : Jan  1 2018 - 01:47:45
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll 

![image](https://user-images.githubusercontent.com/14791461/35281758-c3740ba4-0079-11e8-8fc4-a2bd3fbc8f17.png)

![image](https://user-images.githubusercontent.com/14791461/35282050-8901848c-007a-11e8-9bde-200eab7399f5.png)

SinghRajenM:
This is similar to #4112. Just to cover all other scenarios (Specially case 2 and 3), created this issue. Also, it does not limit to "Macro" menu, but cover all the menu items.

PR #4115, opened 2018-01-23
Menu item's state is not maintained

SinghRajenM:
Fixed issue #4112 and #4114


donho:
@SinghRajenM  Nice job!
But how come the following command makes a mess to the other command?
`enableCommand(IDM_MACRO_SAVECURRENTMACRO, !_macro.empty() && !_recordingSaved, MENU | TOOLBAR);`

PR #4077, opened 2018-01-12
Update code in FileManager::reloadBuffer to correctly detect line endings and encoding

CookiePLMonster:
Fixes #2637, #2843, #3557 and #4033 (maybe more?)

PR seems nearly identical to rejected #2633, however I have actually compiled and tested this and all test cases work fine.

This PR mirrors encoding/EOL detection present in `FileManager::loadFile` to `FileManager::reloadBuffer`. As a result, I observed following issues when reloading files modified outside of Notepad++ are fixed:

- EOF changes get detected and reflected in the GUI
- If default encoding is set as UTF-8, yet a modified file is a valid ANSI file then encoding gets set as UTF-8 (just like when opening) as opposed to ANSI - that is, default settings are now respected

PR #4034, opened 2018-01-01
[Fixed issue] - Detect Line ending changes on Reload

SinghRajenM:
Fixed issue #4033

Issue #2843, opened 2017-01-29
Inconsistent character encoding detection on open vs reload


Darkwolfdude:
Hi, using version 7.3.1 (it was also bugged on a previous version) I've been working on some UTF-8 XML (this XML class that i'm working with only accepts UTF-8 input and will crash out otherwise - and also putting the file through phps libxml error detection to confirm if it's UTF-8 or not - which it is when this error occurs).

When I open a newly created file the document displays in UTF-8 as expected, however when it automatically refreshes the file or reload the file after it has been updated, the file encoding changes from UTF-8 to TIS-620. It appears that the character that is mainly bugging it out is the left accented a; "á" (in my testing at least).

I took a look at a couple of the other issues raised and I tried going Preferences -> Misc -> Autodetect character encoding and turned that off. It worked after closing the original file, opening it, changing it externally then reloading it (but it didn't fix it after just changing the option then reloading the file after changing it externally).

I also have Preferences -> New Document -> Encoding -> UTF-8 > Apply to opened ANSI files [ON] (altering this didn't seem to do anything for me - but may not have tested that properly).

*When I reference UTF-8, it's without BOM.

### Debug Information
Notepad++ v7.3.1   (64-bit)
Build time : Jan 17 2017 - 00:28:43
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll 

Screenshot1 (Expected):
![image](https://cloud.githubusercontent.com/assets/2396866/22408952/e688e684-e679-11e6-8ed7-4d77ee36090b.png)

Screenshot2 (Actual):
![image](https://cloud.githubusercontent.com/assets/2396866/22408959/f8091f1e-e679-11e6-82c3-b571eca25e3a.png)

Gabr-F:
I don't know how Notepad++ attempts to detect UTF-8 files, but it's usually just possible to rule out that a file is _not_ UTF-8, if it contains codes illegal in UTF-8.
If you can you should really use BOMs for your unicode files, if you want to edit them conveniently with text editors; something that might be a BOM will probably always be regarded as a sure indicator of a unicode file, by unicode-aware editors.

(by the way your XML class will only crash if you happen to use illegal UTF-8 characters or sequences)

Issue #4079, opened 2018-01-13
0x00 symbols are omitted


SerVB:
### Description of the Issue
0x00 symbols are omitted

### Steps to Reproduce the Issue
1. Open attachment file 
[output.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1628086/output.txt) in ANSI Encoding.
2. See only 16 symbols (but not 32).

### Expected Behavior
The file contains 32 bytes: one '0x03' and 31 '0x00'. So I want to see "ETX NUL NUL..." (NUL 31 times).

### Actual Behavior
I see "ETX NUL NUL..." (NUL only 15 times). Where are other 16?

### Debug Information
Notepad++ v7.5.4   (32-bit)
Build time : Jan  1 2018 - 01:47:45
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

CookiePLMonster:
Looks like the "real" issue is that notepad++ identifies this file as UCS-2 Little Endian and using "Encode in ANSI" doesn't seem to take effect.

CookiePLMonster:
Potentially fixed with #3836

SerVB:
I agree, it's certainly an issue with encoding change.

PR #4124, opened 2018-01-28
Extracted parts of FileManager::reloadBuffer and FileManager::loadFile to a separate func

### LINKED ISSUE #2637: NP++ has issue changing status bar from ANSI to UTF-8 when the file is changed externally
### Description of the Issue
In notepad++ when a comma separated text files encoding changes from ANSI to UTF-8 with no byte order marker the status bar does not reflect the change.  The file needs to be closed and reopened for the status bar to reflect the change.

1. Load the ANSI file into NP++
2. Use a text editor, other than NP++, that can save an ANSI file as UTF-8 with no BOM
3. Go back to NP++ and allow it to reload the externally change file and note that the encoding doesn't change from ANSI to UTF-8

### Expected Behavior
I expected the status bar to change from ANSI to UTF-8

### Actual Behavior
The status bar stayed displaying ANSI

### Debug Information
Notepad++ v7.1   (32-bit)
Build time : Oct 15 2016 - 23:30:30
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll

CookiePLMonster:
Fixes #2637 and #2843

This is a de facto re-issue of #4077. I remade the PR after #4034 was merged. This time, however, I went ahead and extracted the common part of reloadBuffer and loadFile into a separate function.

My main reasoning behind that is code duplication - notice that if that logic was separate, this issue would not have existed in the first place! loadFile must have had its functionality expanded at some point, but reloadBuffer was left as is. I think extracting it to a separate function will ease maintenance.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -601,31 +601,13 @@ BufferID FileManager::loadFile(const TCHAR * filename, Document doc, int encodin
 		Buffer* buf = _buffers.at(_nbBufs - 1);
 
 		// restore the encoding (ANSI based) while opening the existing file
-		NppParameters *pNppParamInst = NppParameters::getInstance();
-		const NewDocDefaultSettings & ndds = (pNppParamInst->getNppGUI()).getNewDocDefaultSettings();
-		buf->setUnicodeMode(ndds._unicodeMode);
 		buf->setEncoding(-1);
 
 		// if no file extension, and the language has been detected,  we use the detected value
 		if ((buf->getLangType() == L_TEXT) && (detectedLang != L_TEXT))
 			buf->setLangType(detectedLang);
 
-		if (encoding == -1)
-		{
-			UniMode um = UnicodeConvertor.getEncoding();
-			if (um == uni7Bit)
-				um = (ndds._openAnsiAsUtf8) ? uniCookie : uni8Bit;
-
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -601,31 +601,13 @@ BufferID FileManager::loadFile(const TCHAR * filename, Document doc, int encodin
 		Buffer* buf = _buffers.at(_nbBufs - 1);
 
 		// restore the encoding (ANSI based) while opening the existing file
-		NppParameters *pNppParamInst = NppParameters::getInstance();
-		const NewDocDefaultSettings & ndds = (pNppParamInst->getNppGUI()).getNewDocDefaultSettings();
-		buf->setUnicodeMode(ndds._unicodeMode);
```
CookiePLMonster: Pre-emptive note on removal of this call - notice `setUnicodeMode` is (and was even before my changes) called down below in both branches of `if (encoding == -1)` - thus this call is useless.

dail8859:
@CookiePLMonster Does this PR also fix #1632?

CookiePLMonster:
Hard to tell, but I would say it does. However, the bug report mentions -binary- files so it may not be it... This one sounds more like #4079

PR #4135, opened 2018-02-02
Update danish.xml to 7.5.5

scootergrisen:


donho:
@scootergrisen  Could you check the newest updated english.xml here:

PR #4138, opened 2018-02-03
Simplify and fix memory leak in getSpecialFolderLocation

### LINKED ISSUE #399: CoTaskMemFree should be called when using `SHGetSpecialFolderLocation`
According https://msdn.microsoft.com/en-us/library/windows/desktop/bb762203%28v=vs.85%29.aspx, `CoTaskMemFree` should be called when calling `SHGetSpecialFolderLocation`

CookiePLMonster:
Fixes #399, which was assigned since July 2015 and no fix seems to exist, even though it's literally a few lines.

This PR rewrites `NppParameters::getSpecialFolderLocation`, improving upon a few things in this function:

1. Logic simplified overall - `SHGetSpecialFolderLocation` is deprecated, and the only use for the obtained `PIDL ` was to get a path to this folder - so by replacing it with `SHGetFolderPath` we achieve the same result with less code.
2. Fixed a memory leak - as noted in the documentation (and the issue linked), `SHGetSpecialFolderLocation` requires its output param to be freed with `CoTaskMemFree`. Now we don't need to worry about this.
3. Made the function have a single return point - this makes it eligible for NRVO optimization. I realize it's not going to gain anything since it's a simple and rarely used function, but if it was being rewritten anyway... then why not?

Sorry @milipili for kind of hijacking the issue, but I figured it's abandoned after 2.5 years.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -949,15 +949,15 @@ bool NppParameters::reloadLang()
 
 generic_string NppParameters::getSpecialFolderLocation(int folderKind)
 {
-	ITEMIDLIST *pidl;
-	const HRESULT specialLocationResult = SHGetSpecialFolderLocation(NULL, folderKind, &pidl);
-	if (!SUCCEEDED( specialLocationResult))
-		return generic_string();
-	
 	TCHAR path[MAX_PATH];
-	SHGetPathFromIDList(pidl, path);
+	const HRESULT specialLocationResult = SHGetFolderPath(nullptr, folderKind, nullptr, SHGFP_TYPE_CURRENT, path);
 
-	return path;
+	generic_string result;
+	if (SUCCEEDED(specialLocationResult))
+	{
+		result = path;
+	}
+	return result;
 }
 
 


PowerEditor/src/Parameters.cpp
```diff
@@ -949,15 +949,15 @@ bool NppParameters::reloadLang()
...
 generic_string NppParameters::getSpecialFolderLocation(int folderKind)
 {
-	ITEMIDLIST *pidl;
-	const HRESULT specialLocationResult = SHGetSpecialFolderLocation(NULL, folderKind, &pidl);
-	if (!SUCCEEDED( specialLocationResult))
-		return generic_string();
-	
 	TCHAR path[MAX_PATH];
-	SHGetPathFromIDList(pidl, path);
+	const HRESULT specialLocationResult = SHGetFolderPath(nullptr, folderKind, nullptr, SHGFP_TYPE_CURRENT, path);
```
SinghRajenM: Even ```SHGetFolderPath``` is deprecated ([MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762181(v=vs.85).aspx)). 

If I were there, I would use ```SHGetKnownFolderIDList``` or ```SHGetKnownFolderPath``` and fall back (for earlier than vista) ```SHGetPathFromIDList``` or ```SHGetFolderPath``.

CookiePLMonster: I realize `SHGetFolderPath` is also deprecated, however the other functions are even "more" deprecated.

I can use `SHGetKnownFolderPath` for Vista and newer, however it creates a few minor problems:

1. `SHGetKnownFolderPath` uses `KNOWNFOLDERID` instead of `CSLID` so the code isn't a drop-in replacement anymore. A function to convert those can be provided with ease, however.
2. Are there any examples in n++ code on how to pick a different function for newer OS? This is generally an easy task, but I don't want my solution to be too different to what n++ employs currently. I am thinking of replacing this entire function with a member variable being a function pointer, default to XP-friendly variation of the function and replace it with a Vista+ overload if available. As for importing `SHGetKnownFolderPath`, I'm pretty sure I'll find an example of how functions obtained via `GetProcAddress` should be stored to conform with current code style.
3. `SHGetKnownFolderPath` doesn't have A and W variations anymore - n++ is probably not going to support MBCS compilation anymore (no point really), but just for the sake of clarity it would probably be best not to include this code at all for non-UNICODE builds.

SinghRajenM: > Are there any examples in n++ code on how to pick a different function for newer OS?

In same file around line 1066

```C++
	//---------------------------------------------//
	// Dlg theme texture function for xp and vista //
	//---------------------------------------------//
	_hUXTheme = ::LoadLibrary(TEXT("uxtheme.dll"));
	if (_hUXTheme)
		_enableThemeDialogTextureFuncAddr = (WNDPROC)::GetProcAddress(_hUXTheme, "EnableThemeDialogTexture");
```

Also in winmain.cpp check method ```allowWmCopydataMessages```.

> n++ is probably not going to support MBCS compilation anymore

I think so. N++ has dropped ANSI support.

P.S. Considering others all the minor hiccups and simplicity, ```CoTaskMemFree``` can be used. Because there is no point in introducing a deprecated API to shadow another deprecated API.

CookiePLMonster: Very well then, I guess it makes sense to branch the code for XP/Vista+ then. While it may make the approach a bit less simple, I do think it makes sense to go for a slightly more complex solution and end up using the most up to date method available. Feels more future proof this way, and IMO with a project like this it should be a major decisive point.

My biggest issue with `SHGetSpecialFolderLocation` is that I believe it dates back to Win9x days - whereas `SHGetFolderPath` was deprecated "only" with the introduction on Vista.

CookiePLMonster: Seeing how this proposal is completely different to the original idea, maybe it makes more sense to create a secondary pull request (not relying on this one) and leave both for review, so it can be decided which one should be chosen?

CookiePLMonster: UPDATE: Literally everything breaks when trying to compile via MBCS, so no need to worry about this one 😆

donho: @CookiePLMonster 
My viewpoint is that Microsoft has a very different definition for `Deprecated` (comparing with Apple, for example).
So my approach is rather keeping the deprecated functions to have maximum compatibility until they doesn't work on the newest OS (or the rendered UI become really obsolete).

CookiePLMonster: Glad to hear that - in which case keeping `SHGetFolderPath` makes sense because:

1. Currently `PIDL` is obtained only to form a path out of it (so `SHGetFolderPath` achieves the same result with less code).
2. No need to worry about freeing up memory.

donho: BTW
https://twitter.com/Notepad_plus/status/959902540195299331
:)

PR #4139, opened 2018-02-03
update japanese.xml to v7.5.5

yasmise:
* Fix encoding names to match status bar
* Add a new translation ("2 find buttons mode")


donho:
@yasmise I try to update every language once per release, so please make sure you have done everything once, but not twice.

yasmise:
@donho I just confirmed that this japanese.xml is based on the latest english.xml, and I believe the translations are correct. I'm sure no more commit is required.

donho:
@yasmise Could you check the newest updated english.xml here:

yasmise:
Hi @donho 
Now I finished translating. Thanks for adding new translations!
But, I have two questions.
1. There are two `find-status-mark-nb-matches` items. [Here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/803917dd0989c69d97145a35f06ac93753668e26/PowerEditor/installer/nativeLang/english.xml#L1126) and [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/803917dd0989c69d97145a35f06ac93753668e26/PowerEditor/installer/nativeLang/english.xml#L1141). I think second one is not needed so this commit doesn't include it. Is it okay?
2. I've found `$INT_REPLACE$` is used for possibly string replacement. `CreateNewFileOrNot`, `CreateNewFileError`, `OpenFileError`, `FileBackupFailed` and `LoadStylersFailed`. They are from [Line 989](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/803917dd0989c69d97145a35f06ac93753668e26/PowerEditor/installer/nativeLang/english.xml#L989) to Line 995. Are they correct?

donho:
VERY NICE CATCH @yasmise 
Fixed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a1b4628b8b6e43bc8af59f58b9d5cde34b9afff6

donho:
@yasmise 
Please let me know when your translation is updated.

yasmise:
Translation updated to follow up ["Unix style path on open dialog"](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f9c8a0e22189404f78ba1f03581f19938f0b777b).

Issue #4131, opened 2018-02-01
Find In Files, fails on big files


RudiHansen:
When doing a Find in Files, you can get the error “File is too big to be opened by Notepad++” 
And every time you get that error the search stops, and does continue before you say OK to the error.

Perhaps you can have an option to skip large files in the search dialog?

dail8859:
Duplicate of #1889

ramgarden:
We tried to work on this issue yesterday but couldn't get the code to compile.  Need better instructions in the repo on how to compile for those of us that are VERY rusty on our C++.

MetaChuh:
please update your notepad++ to the latest 64 bit version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

Issue #1889, opened 2016-05-24
Find in Files...shouldn't prompt if file is too big to be opened


ramgarden:
when doing a text search in a folder with lots of subfolders and files (thousands) it shouldn't pop up any dialogs requiring the user to click OK like the "file is too big to be opened in notepad++".  A really long search should be able to run unattended and finish on its own.

danielddubois:
Agreed

kalendello:
This issue forces me to rely on other software for an unattended "find in files" search.  It'd be very easy to suppress the messagebox.

LazyBui:
I personally think there are several usability issues here:

1. It doesn't display the actual file name, so you have no idea what file it's stuck on.
2. It doesn't display the file size so you have no sense of scale.
3. It pops up and there's no way to silence them for a particular search operation or permanently (with a reset toggle in the configuration).
4. The search option granularity is such that it forces you to see this when you must search large or deep directory hierarchies looking for some information you're interested in. For example, configuration keys. You don't know if they're in .txt files, .ini files, .xml files, .cfg files, or what other sort of file. But you have a string that you know is in there. If these options were improved, the issue would be mitigated somewhat (being able to exclude .exe and .dll files from your *.* search filter, for example, would be a great start).

This information could also be collected throughout the operation and reported singularly at the end like the results are instead of utilizing the message dialog. It would be nice to be able to cancel the operation if one of these pops up, but that would be nice generally speaking and doesn't need to be related.

If keeping the dialog is desired, I've taken a look at the code and API revision will be required since a catch block divorced from the context of what's going on pops up the error and it's a simple `MessageBox`, not something more complex that could be customized. This load function is used by several places, so it may not be so straightforward to change the API.

I'm interested in solving this issue because I'm quite tired of having to hit OK hundreds of times for a search. The only question is basically how @donho wants to handle the work.

What I would personally like to see is something like this mockup:
![mock_big_options](https://user-images.githubusercontent.com/43490/33229601-10b7d4e8-d197-11e7-9099-f0b18c9cfa96.png)

The reason why I would like to see this is that I can see it being useful for some searches even if you don't want it most of the time. The ability to banish it temporarily or permanently makes me feel good about it being the default so the user knows it exists.

I realize that this would require quite a bit of API change, but I don't mind doing that as long as I know I'll be able to work with the author and get the code into mainline. Any comments or recommendations if I'm going to do this work?

ramgarden:
[...]

PR #4140, opened 2018-02-03
Don't halt find/replace operation because of large file

SinghRajenM:
Fixes defect: #4131 and #1889 

Many times, we get multiple logs (specially from server/production server) to analyze reported issues. Sometimes, these logs are more in number whereas few logs are huge in size. Notepad++ is very useful to find special keywords such as ERROR. **But sometimes the complete process gets stuck due to big file. Unless we kill the messagebox it does not proceed further**.

I have two proposals to deal with this.
1. By default ignore large files for search/replace and once operation is complete show the ignore file details. This message box can be shown/hidden using setting.
2. Provide a setting to exclude large files. Once operation is complete show the ignore file details. If exclude file setting is disabled, then keep the existing behavior.

This PR implements the second approach. If the proposal sounds good, I'll will improve it by moving strings to localization files.

image

## Further ToDo##
If user misses to enable setting for large file exclusion, provide him one more chance to disable for current search/replace session (may be by providing checkbox in messagebox or similar).

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1824,41 +1824,55 @@ void Notepad_plus::filePrint(bool showDialog)
 
 int Notepad_plus::doSaveOrNot(const TCHAR *fn)
 {
-	TCHAR pattern[64] = TEXT("Save file \"%s\" ?");
-	TCHAR phrase[512];
-	wsprintf(phrase, pattern, fn);
-	return doActionOrNot(TEXT("Save"), phrase, MB_YESNOCANCEL | MB_ICONQUESTION | MB_APPLMODAL);
+	return _nativeLangSpeaker.messageBox("DoSaveOrNot",
+		_pPublicInterface->getHSelf(),
+		TEXT("Save file \"$STR_REPLACE$\" ?"),
+		TEXT("Save"),
+		MB_YESNOCANCEL | MB_ICONQUESTION | MB_APPLMODAL,
+		0, // not used
+		fn);
 }
 
 int Notepad_plus::doReloadOrNot(const TCHAR *fn, bool dirty)
 {
-	TCHAR* pattern = TEXT("%s\r\rThis file has been modified by another program.\rDo you want to reload it%s?");
-	TCHAR* lose_info_str = dirty ? TEXT(" and lose the changes made in Notepad++") : TEXT("");
-	TCHAR phrase[512];
-	wsprintf(phrase, pattern, fn, lose_info_str);
-	int icon = dirty ? MB_ICONEXCLAMATION : MB_ICONQUESTION;
-	return doActionOrNot(TEXT("Reload"), phrase, MB_YESNO | MB_APPLMODAL | icon);
+	if (dirty)
+		return _nativeLangSpeaker.messageBox("DoReloadOrNotAndLooseChange",
+			_pPublicInterface->getHSelf(),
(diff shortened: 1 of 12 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1545,6 +1551,11 @@ bool Notepad_plus::replaceInFiles()
 
 	progress.close();
 
+	// Show large ignored files which could not be opened for find/repace in folder
+	// It is just to intimate user about the ingored file details
+	if (isIgnoreLargeFileMessage && ignoredFiles.size() > 0)
```
CookiePLMonster: !ignoredFiles.empty()

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1710,6 +1727,11 @@ bool Notepad_plus::findInFiles()
...
```
CookiePLMonster: !ignoredFiles.empty()

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1811,6 +1833,37 @@ bool Notepad_plus::findInCurrentFile()
 	return true;
 }
 
+void Notepad_plus::showIgnoredFilesMessage(ProcessOperation op, const std::vector<generic_string>& ignoredFiles)
+{
+	generic_string message;
+	switch (op)
+	{
+	case ProcessFindAll:
+		message += ignoredFiles.size() > 1 ?
```
CookiePLMonster: Does n++ account for grammar of non-English languages? In pretty much all Slavic languages this would require more than two strings (_1 plik, 2-4 pliki, 5+ plików_) but unless n++ actually accounts for that elsewhere, no point in worrying about this.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1811,6 +1833,37 @@ bool Notepad_plus::findInCurrentFile()
...
+			_T("Below file was ignored while replacing in folder.\r\n\r\n");
+		break;
+	default:
+		message += ignoredFiles.size() > 1 ?
+			_T("Below files were ignored.\r\n\r\n") :
+			_T("Below file was ignored.\r\n\r\n");
+		break;
+	}
+
+	for (auto file : ignoredFiles)
```
CookiePLMonster: either `auto&` or `message += std::move(file)` - preferably the former. Props for going with a foreach though, they are much underused in the codebase (iterating through `std::vector` with an index and comparing it against `size()` makes me shiver a bit).

PowerEditor/src/Notepad_plus.cpp
SinghRajenM: I feel it should not matter. Because 
1. If we use size then number of operations = 2 (one for calculating size + one for comparison again 0).
2. If we use empty then number of operations = 2 (one for comparing last==first + one for ! )

image

PowerEditor/src/Notepad_plus.cpp
SinghRajenM: I feel it should not matter.

PowerEditor/src/Notepad_plus.cpp
SinghRajenM: Will update as ```const auto&```.

PowerEditor/src/Notepad_plus.cpp
CookiePLMonster: I agree they both are equally performant, however you're talking about implementation details where I noticed the style. `empty()` just expresses your intent better.

Besides, even though we both know `size()` and `empty()` are always going to stay the same, it's not guaranteed by the standard. Standard only guarantees that both methods must be of constant complexity for `std::vector`.

PR #4141, opened 2018-02-03
Fix Unicode file name comparision to match Windows behaviour

### LINKED ISSUE #3820: Can't open two different files whose canonic names are the same
### Can't open two different files whose canonic names are the same
https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization

1. Create a file called "é.txt" (C form)
2. Create a file called "é.txt" (D form)
Open the first file. Open the second file.

### Expected Behavior
It should open a new tab for the second file.

### Actual Behavior
The tab of the first file is selected.

### Debug Information
Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Admin mode : ON
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppTextFX.dll PluginManager.dll XMLTools.dll 

image

CookiePLMonster:
Fixes #3820

This PR changes the way notepad++ compares file names - it used to compare file names linguistically, so two Unicode files with names which would be identical after canonization (but not identical as per the way Windows treats file names) were treated as same.

Since comparing file names seems more complex than one would initially expect, I have used a code snippet from Gregg Miskelly's MSDN blog:
https://blogs.msdn.microsoft.com/greggm/2005/09/21/comparing-file-names-in-native-code/

I might have overlooked some comparision functions on filenames, current changes reflect only the things I stepped by when debugging and then mass searching for `lstrcmp` and noticing 'filename' in one of parameters' names.

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -29,6 +29,7 @@
 #include <shlwapi.h>
 #include <shlobj.h>
 #include <uxtheme.h>
+#include <cassert>
 #include "StaticDialog.h"
 
 #include "Common.h"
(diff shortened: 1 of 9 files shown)


CookiePLMonster:
I have found more calls which compare file names - also included them for the sake of completeness (and potentially killing some edge cases if any of those hardcoded file names can be represented in more than one way).

PR #2478, opened 2016-10-24
Mingw GCC update

chcg:
- changes necessary to get back to a buildable version with mingw, after last adaptation with https://github.com/notepad-plus-plus/notepad-plus-plus/pull/1229
- fixed some gcc compiler errors and warnings, see https://ci.appveyor.com/project/chcg/notepad-plus-plus/build/1.0.3 to https://ci.appveyor.com/project/chcg/notepad-plus-plus/build/1.0.13
- remaining task is the wincontrol "ReadDirectoryChanges", which uses ATL, propably not available with mingw 
  @donho 
- what is your preference here, maybe it is easier to switch off this feature by ifdef mingw, than to adapt the control. I created a buildable version, which remove the dependency to ATL and replaced most parts by c++11 equivalents, but that is quite intrusive, see https://github.com/chcg/notepad-plus-plus/commits/mingw_gcc/PowerEditor/src/WinControls/ReadDirectoryChanges and not really tested for maybe newly introduced issues
- should a build target for mingw(, see https://www.appveyor.com/docs/installed-software/#mingw-msys-cygwin) be added to appveyor?

PR #1663, opened 2016-03-20
Fix many GCC compile warnings and some errors

marcows:
Fixes #1662

The MinGW build is not successful anymore since addition of the file browser. This pull request doesn't fix this problem.

PR #4145, opened 2018-02-04
update makefile for mingw gcc builds

chcg:
, see also #3918

Last update was  #2478.


chcg:
@GitMensch See #2478 with just this changes the build still fails, due to ATL, which is not available for mingw. So adding an additional build for it just makes sense, if it is planned to get this build running, either by disabling the problematic code via #ifdef or by some stub similar to https://github.com/notepad-plus-plus/notepad-plus-plus/pull/1663. Maybe in the meantime also some new issues were introduced.

donho:
@chcg Does this PR still be updated (compiled by MinGW) ?

chcg:
As this is 2 years old a rebase and adaption to current changes is necessary, reflecting:
- Revision: 581aff630af9b12b85b52c3673d15d8f6fde8031
Use standard C++11 mutex
And remove unecessary files
Close #6093
- Revision: 519fecbb827eb011e6da6d546e5da1d22fb982a8
Add SHA-256 implementation and use sha-256 instead of md5 in Plugin Admin

as far as I see. Some commits in the meantime are already helping to get closer to goal to have a mingw build like removal of the ATL by usage of C++11 mutex and Force cpp standard const char pointer on string literals

I could do a rebase and adapt.

chcg:
So this is WIP, appveyor will be incompatible modified until the makefile is updated.

chcg:
@donho See https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/28608301/job/85t95ficdgio8i5h/artifacts, seems to be a first working version, if combined with a signed SciLexer.dll. Quite huge as it is a debug build with all dependency libs linked in.

GitMensch:
This is very good news. Can you target for a non-static non-debug build [a different make target] next?

donho:
@chcg 
Looks good to me.
Sorry for the wait.

chcg:
@donho 
Some points remain I think:
- UserDefineDialog.cpp 
I think it should be possible to do the std::find() also for mingw
- functionListPanel.cpp 
check ofstream usage, maybe via basic_ofstream<TCHAR>
- appveyor.yml mingw config
scintilla is build just as release version
notepad++ seems to be just debug

Issue #3895, opened 2017-11-17
Double click a word is not working


ahmedhamdy90:
### Description of the Issue

After the recent update, when I double click a word, it is no longer highlighted. I have to slow my double-click speed in order to select a word, or I use mouse dragging to select it.
Double click working fine in chrome, edge, eclipse, sublime, and even the default windows notepad!

### Steps to Reproduce the Issue

1. open notepad++
2. try to double-click any word with normal (as usual) speed

### Expected Behavior

Selecting the current word and highlight all its occurrences 

### Actual Behavior

Nothing is happening, no word highlighting, no occurrences highlighting

### Debug Information

Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

### Things I did to resolve the issue
1. I tried to uninstall and install the 64-bit version --> still not fixed
2. I tried to make a clean uninstall (plugins, setting, and config) and install the 64-bit version --> still not fixed
3. I tried to make a clean uninstall again and install the 32-bit version again --> still not fixed 

Kindly find a gif showing the issue, this gif contains 3 double clicks: normal, slow, and fast double click on a word.
Normal and fast double click is not working properly on Notepad++, but they work fine in any other program.
![issue](https://user-images.githubusercontent.com/252999/32956707-347321d8-cbc2-11e7-9898-717ab6e33eb8.gif)

encore0117:
Does this issue persists if running on other machines?

I just tried running the same version in the same system, but I could not replicate the issue.

xv:
Are you double clicking with a mouse or a touchpad? If it's a touchpad, you're not alone; however, this seems to be a Scintilla issue not a NP++ one. I tried multiple Scintilla-based editors and they all have the same problem. I don't recall having this issue few months ago. I assume it's caused either by a Windows 10 update or a Synaptics driver update.

ahmedhamdy90:
@xv yes, I am using the touchpad. My laptop is dell inspiron 7559.

xv:
[...]

PR #4147, opened 2018-02-05
Fix for #3895 Workaround Windows bug

mitkoradev:
Scintilla code changed, making fix in Notepad++ seems harder, still probably possible too

http://prntscr.com/ia5egq

Looks like the issue is related to some crazy timing information returned by GetMessageTime
so time between last message and current message gets negative and wraps around as the comparison to doubleclick time is an unsigned one. Happens with touchpads, probably some stupid gesture recognition doesn't write right timestamps..

These casts for signed comparison seems to fix it

src/Editor.cxx line 4458

if ((((int)(curTime - lastClickTime)) < (int)Platform::DoubleClickTime()) && Close(pt, lastClick, doubleClickCloseThreshold)) {

### DIFF
--- a/scintilla/src/Editor.cxx
+++ b/scintilla/src/Editor.cxx
@@ -4135,7 +4135,7 @@ void Editor::ButtonDownWithModifiers(Point pt, unsigned int curTime, int modifie
 	if (shift && !inSelMargin) {
 		SetSelection(newPos);
 	}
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
Excuse me for not paying enough attention to the issue, but wouldn't using `WM_LBUTTONDBLCLK` message instead of counting times manually help mitigate the issue better?

donho:
@mitkoradev If it's Windows' bug, Microsoft should fix it, not on Notepad++ side.

PR #4150, opened 2018-02-05
Correctly use const qualifier for variables holding string literals

### LINKED ISSUE #4146: Codebase incompatible with /Zc:strictStrings
For info on `/Zc:strictStrings` see:
[https://msdn.microsoft.com/en-us/library/dn449508.aspx](url)

By default, Visual Studio allows for nonstandard C++ behaviour by allowing to assign string literals to non-const qualified types. For example, this is **not** standard:

`TCHAR *nbCharLabel = TEXT("Characters (without blanks): ");`

Standard requires such literal is const, so by definition it should not be possible to assign them like this. It's more than likely an attempt to write to the value pointed to by `nbCharLabel` will cause access violation.

`/Zc:strictStrings` enforces VS compiler to follow standard behaviour. Thus, it may be feasible to enable this switch for all compilation options (it's not exposed to project settings in 2015/2017, so it needs to be manually added to `Command Line` tab. When done, code will refuse to compile, throwing lots of errors (sadly, some of them are in `ScintillaComponent` - is it fine to fix things in there from within Notepad++). However, it's safe to just correct them - as if the code was relying on non-constness of such literals already, we'd end up with a crash any time code attempts to write to it. Thus, the only places which are not going to be trivially fixable (by adding a `const` qualifier to parameters and variables) are potential bugs in the code.

EDIT:
#3918 seems to reference this issue - tagging @chcg

CookiePLMonster:
Fixes #4146

This PR enables `/Zc:strictStrings` for a VS2015 project. Since this compilation flag disallows assigning string literals to non-const variables (like C++ standards states), code changes were made to ensure all variables storing pointers to string literals are const-qualified.

While is this purely a code clarity fix, it has potential for resolving one potential issue - previously, string literals were assigned to a non-const `pszText` field in `LVCOLUMN` structure. Documentation says nothing whether contents of this field are guaranteed not to be modified when `LVCOLUMN` is being read from (not populated to), so it was potentially unsafe.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1917,10 +1917,10 @@ void Notepad_plus::command(int id)
 
 			if (fileLen != -1)
 			{
-				TCHAR *filePathLabel = TEXT("Full file path: ");
-				TCHAR *fileCreateTimeLabel = TEXT("Created: ");
-				TCHAR *fileModifyTimeLabel = TEXT("Modified: ");
-				TCHAR *fileLenLabel = TEXT("File length (in byte): ");
+				const TCHAR *filePathLabel = TEXT("Full file path: ");
+				const TCHAR *fileCreateTimeLabel = TEXT("Created: ");
+				const TCHAR *fileModifyTimeLabel = TEXT("Modified: ");
+				const TCHAR *fileLenLabel = TEXT("File length (in byte): ");
 
 				characterNumber += filePathLabel;
 				characterNumber += curBuf->getFullPathName();
@@ -1939,13 +1939,13 @@ void Notepad_plus::command(int id)
 				characterNumber += TEXT("\r");
 				characterNumber += TEXT("\r");
 			}
-			TCHAR *nbCharLabel = TEXT("Characters (without blanks): ");
-			TCHAR *nbWordLabel = TEXT("Words: ");
-			TCHAR *nbLineLabel = TEXT("Lines: ");
-			TCHAR *nbByteLabel = TEXT("Current document length: ");
-			TCHAR *nbSelLabel1 = TEXT(" selected characters (");
-			TCHAR *nbSelLabel2 = TEXT(" bytes) in ");
-			TCHAR *nbRangeLabel = TEXT(" ranges");
+			const TCHAR *nbCharLabel = TEXT("Characters (without blanks): ");
+			const TCHAR *nbWordLabel = TEXT("Words: ");
+			const TCHAR *nbLineLabel = TEXT("Lines: ");
+			const TCHAR *nbByteLabel = TEXT("Current document length: ");
+			const TCHAR *nbSelLabel1 = TEXT(" selected characters (");
+			const TCHAR *nbSelLabel2 = TEXT(" bytes) in ");
+			const TCHAR *nbRangeLabel = TEXT(" ranges");
 
 			UniMode um = _pEditView->getCurrentBuffer()->getUnicodeMode();
 			auto nbChar = getCurrentDocCharCount(um);
@@ -2421,9 +2421,9 @@ void Notepad_plus::command(int id)
 		case IDM_SETTING_IMPORTPLUGIN :
         {
             // get plugin source path
-            TCHAR *extFilterName = TEXT("Notepad++ plugin");
-            TCHAR *extFilter = TEXT(".dll");
-            TCHAR *destDir = TEXT("plugins");
+            const TCHAR *extFilterName = TEXT("Notepad++ plugin");
+            const TCHAR *extFilter = TEXT(".dll");
+            const TCHAR *destDir = TEXT("plugins");
 
             vector<generic_string> copiedFiles = addNppComponents(destDir, extFilterName, extFilter);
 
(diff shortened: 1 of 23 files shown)


PowerEditor/src/WinControls/TabBar/ControlsTab.h
```diff
@@ -38,7 +38,7 @@ struct DlgInfo
     generic_string _name;
 	generic_string _internalName;
 
-	DlgInfo(Window *dlg, TCHAR *name, TCHAR *internalName = NULL): _dlg(dlg), _name(name), _internalName(internalName?internalName:TEXT("")) {};
```
CookiePLMonster: NON TRIVIAL CHANGE  - changed default parameter from `NULL` to `""` so there is no need for this conditional check. Checking through all uses of this constructor, no one ever passes `NULL` to it so current behaviour stays unchanged.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -534,18 +535,29 @@ BOOL WindowsDlg::onInitDialog()
 	lvColumn.mask = LVCF_WIDTH | LVCF_TEXT | LVCF_SUBITEM | LVCF_FMT;
 	lvColumn.fmt = LVCFMT_LEFT;
 
-	lvColumn.pszText = TEXT("\u21F5 Name");
```
CookiePLMonster: NON TRIVIAL CHANGE - `pszText` requires non-const strings so I used `CString` for that. It's a perfect candidate (better than `std::string`), since its `GetBuffer()` returns a non-const string pointer. Not possible with `std::string` until C++17 or so.

PowerEditor/visual.net/notepadPlus.vcxproj
```diff
@@ -106,6 +106,7 @@
       <SmallerTypeCheck>false</SmallerTypeCheck>
       <DisableSpecificWarnings>4091;4456;4457;4459</DisableSpecificWarnings>
       <WarningVersion>18</WarningVersion>
+      <AdditionalOptions>/Zc:strictStrings %(AdditionalOptions)</AdditionalOptions>
```
CookiePLMonster: NON TRIVIAL CHANGE - project settings.

PowerEditor/src/NppCommands.cpp
```diff
@@ -2404,9 +2404,9 @@ void Notepad_plus::command(int id)
 		case IDM_SETTING_IMPORTPLUGIN :
         {
             // get plugin source path
-            TCHAR *extFilterName = TEXT("Notepad++ plugin");
-            TCHAR *extFilter = TEXT(".dll");
-            TCHAR *destDir = TEXT("plugins");
+			const TCHAR *extFilterName = TEXT("Notepad++ plugin");
```
chcg: formatting issue, indentation has changed

PowerEditor/src/NppCommands.cpp
```diff
@@ -2427,9 +2427,9 @@ void Notepad_plus::command(int id)
         case IDM_SETTING_IMPORTSTYLETHEMS :
         {
             // get plugin source path
-            TCHAR *extFilterName = TEXT("Notepad++ style theme");
-            TCHAR *extFilter = TEXT(".xml");
-            TCHAR *destDir = TEXT("themes");
+            const TCHAR *extFilterName = TEXT("Notepad++ style theme");
+			const TCHAR *extFilter = TEXT(".xml");
```
chcg: formatting issue, indentation has changed

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -534,18 +535,29 @@ BOOL WindowsDlg::onInitDialog()
 	lvColumn.mask = LVCF_WIDTH | LVCF_TEXT | LVCF_SUBITEM | LVCF_FMT;
 	lvColumn.fmt = LVCFMT_LEFT;
 
-	lvColumn.pszText = TEXT("\u21F5 Name");
+	CString columnText;
```
chcg: generic_sting would be the normal choice for n++
ATL is not available for mingw.
See http://en.cppreference.com/w/cpp/string/basic_string/data regarding returning non const string data pointer. 
What about a const_cast?

CookiePLMonster: I am not a fan of `const_cast` personally. Do we have non-const `data()` available in VS2015?

CookiePLMonster:
Re-commited to fix indentation issues.

CookiePLMonster:
I tested a `generic_string` based implementation instead of ATL and it works fine - but will hold on changing it until #4154 is merged and rebase to avoid conflicts.

donho:
@CookiePLMonster With the consequent modification, there must be a better argument than "standard C++ const qualifications". So convince me.

CookiePLMonster:
For once, you **can't** assume `LVCOLUMN.pszText` can point to a non-modifable string. It works fine now, but maybe in the future it'll start modifying the string, like using `strtok` to tokenize it? Current approach breaks the contract.

The other const qualifiers are purely cosmetic changes, however. IMO it's good to have that conform to the standard, and it's just safer overall - such refactor ensures you won't ever try to modify those strings by mistake (and you could attempt that now when they are assigned to non const-qualified pointers).

AngryGamer:
This doesn't really help. You're supposed to initialize [LVCOLUMN.cchTextMax](https://msdn.microsoft.com/en-us/library/windows/desktop/bb774743(v=vs.85).aspx) when passing LVCOLUMN into something which writes to LVCOLUMN.pszText (for example TVM_GETITEM).

I don't see why someone knowingly initialize LVCOLUMN.cchTextMax on a const LVCOLUMN.pszText and pass it into TVM_GETITEM. If you're reusing a LVCOLUMN, then I'd hope you'd know the const-ness and size of the buffer you're going to write into...

CookiePLMonster:
Looks good! I'll rebase my code on the latest master and see what is still left.

CookiePLMonster:
Rebased on top of latest master.

Issue #2072, opened 2016-07-12
Prevent operation interruption by dialogs


Grismar:
Currently, when using functionality like "Find in Files", Notepad++ will interrupt operation with a dialog to report that a file is too big to open for Notepad++.

Although reporting this is valuable, the only option provided is 'Ok', but it still interrupts and pauses the operation. Notepad++ could instead opt to collect these messages, possibly even display them in the search dialog as they are being collected without interrupting the operation.

If the user decides this problem is reason to stop the operation, they can do so using the 'Cancel' button.

More in general, if at all possible, shifting the resolution of issues to the end of a batch operation (provided there's no dependency on order) is a useful feature to implement.

bladeoflight16:
Duplicate of #1889.

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/cf8ddc18c9833fb8f19df1c44dc769926a274d47

PR #4151, opened 2018-02-05
Don't halt find/replace operation because of large file

SinghRajenM:
Fixes defect: #4131 ,  #1889 and #2072

Many times, we get multiple logs (specially from server/production server) to analyze reported issues. Sometimes, these logs are more in number whereas few logs are huge in size. Notepad++ is very useful to find special keywords such as ERROR. **But sometimes the complete process gets stuck due to big file. Unless we kill the messagebox it does not proceed further**.

-----------------------------
### Update
~~I have two proposals to deal with this.~~
~~1. By default ignore large files for search/replace and once operation is complete show the ignore file details. This message box can be shown/hidden using setting.~~
~~2. Provide a setting to exclude large files. Once operation is complete show the ignore file details. If exclude file setting is disabled, then keep the existing behavior.~~

~~This PR implements the second approach. If the proposal sounds good, I'll will improve it by moving strings to localization files.~~

In this PR, I tried to solve the above issue. When "Find in Files" or "Replace in Files" operation is performed and npp is unable to open a file, then skip that file and continue with rest of the files. In the end of operation, show the details of skipped file(s) if any.

image

-----------------------------

## Further ToDo##
If the solution sounds good, I will update the PR to support localization.

## ##
Accidentally closed previous PR #4140 while resolving conflict. So opening again.


kevinf28:
Does this scale with 100s of too large files? I see that Ok button being buried quick. Scrollbars or ... after X entries ?

bladeoflight16:
Might I suggest taking a look at [my comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1889#issuecomment-374430377) in #1889? It should deal with the scaling problem more smoothly and negate the need for a setting.

SinghRajenM:
Improved it as suggested by @bladeoflight16.

xantari:
This hasn't been merged yet? Why not?

SinghRajenM:
>This hasn't been merged yet? 

No, it is not merged yet.

>Why not?

Only @donho knows. 😉

informatorius:
Would like to see this merged too :-)

SinghRajenM:
@donho Did you get time to review this PR?

SinghRajenM:
> 
> 
> @donho Did you get time to review this PR?

@donho still waiting for your response.

grantmatheny:
Please merge this as soon as possible! Still an issue in 7.8.5!

wonkawilly:
I'm really sorry if I'm bothering you all, but maybe the participants to this thread would want to give their opinion about this new and rich proposal about the improvement of the Find and Replace dialog at this link: #9627. Which will be a great improvement and perhaps even a real game changer for all of us that use N++ for more advanced uses than just editing the shopping list.
Please partecipate and ask others to. More opinion are welcome. Thank you all.

RobertTanenbaum:
@donho - This may not be a "bug", but it is definitely a feature that many people have requested. 
It is quite painful when I am searching a repository of thousands of .txt files and I have to monitor the search and click on the error popup whenever it occurs.
Being able to specify that the popup regarding too big files be suppressed, is sorely needed.
Is there any way to make an equivalent change in the current code base, so that this feature can be added?

SinghRajenM:
Good to see this PR reopened.
@donho , should I resolve conflicts if you want to take it further?

alankilborn:
@SinghRajenM 

I like your GUI approach to this at the very top.
I was curious if read-only files as well as files too large to open would be included in this?

donho:
@SinghRajenM 
Thank you for your feedback.
As you might notice, v8.2.2 release candidate is out, and in this version (for x64 build) the 2GB restriction will be removed. So normally it should work with v8.2.2 x64 binary (or with a minimum tuning). 
I opened it as a reminder, so it's not necessary do anything for this PR, at least not before the official release of v8.2.2.

donho:
@SinghRajenM @RobertTanenbaum 

After some testing, I confirm you that large file searching restriction has been removed. There is a large file opening confirmation which may be not expected, but it's working.

PR #4152, opened 2018-02-05
[Hindi] Updated localization

SinghRajenM:
1. Updated "Hindi.xml"
2. Corrected typos in English.xml (changed "NppIO.cpp" and "shortcut.rc" accordingly)
3. Synced "english_customizable.xml" with "English.xml"


donho:
@SinghRajenM Could you check the newest updated english.xml here:

SinghRajenM:
@donho It is already updated as per latest translation.

PR #4154, opened 2018-02-05
Warning/error fixes as per VS2017 code analysis

CookiePLMonster:
Since Visual Studio 2017 improved its warnings and static code analysis, I decided to give it a go and sanity check the code against those. Several fairly sensible warnings have been issues, so I decided to batch correct them. Those include two variable truncations, a few buffer overruns, one unsigned/signed mismatch and one instance of WinAPI misuse.

Please refer to reviewed comments for specific warnings/error which have been fixed - I think it's cleanest to pinpoint them this way.

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -766,7 +766,9 @@ COLORREF getCtrlBgColor(HWND hWnd)
 
 generic_string stringToUpper(generic_string strToConvert)
 {
-    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), ::toupper);
+    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), 
+        [](TCHAR ch){ return static_cast<TCHAR>(_totupper(ch)); }
+    );
     return strToConvert;
 }
 
--- a/PowerEditor/src/TinyXml/tinyxmlparser.cpp
+++ b/PowerEditor/src/TinyXml/tinyxmlparser.cpp
@@ -24,6 +24,7 @@ distribution.
 
 
 #include <sstream>
+#include <tchar.h>
 #include "tinyxml.h"
 
 //#define DEBUG_PARSER
@@ -156,7 +157,7 @@ const TCHAR* TiXmlBase::SkipWhiteSpace( const TCHAR* p )
 	}
 	while ( p && *p )
 	{
-		if ( isspace( *p ) || *p == '\n' || *p =='\r' )		// Still using old rules for white space.
+		if ( _istspace( *p ) || *p == '\n' || *p =='\r' )		// Still using old rules for white space.
 			++p;
 		else
 			break;
@@ -204,10 +205,10 @@ const TCHAR* TiXmlBase::ReadName( const TCHAR* p, TIXML_STRING * name )
 	// hyphens, or colons. (Colons are valid ony for namespaces,
 	// but tinyxml can't tell namespaces from names.)
 	if (    p && *p 
-		 && ( isalpha( (UCHAR) *p ) || *p == '_' ) )
+		 && ( _istalpha( *p ) || *p == '_' ) )
 	{
 		while(		p && *p
-				&&	(		isalnum( (UCHAR ) *p ) 
+				&&	(		_istalnum( *p ) 
 						 || *p == '_'
 						 || *p == '-'
 						 || *p == '.'
@@ -270,7 +271,7 @@ bool TiXmlBase::StringEqual( const TCHAR* p,
 		return false;
 	}
 
-    if ( tolower( *p ) == tolower( *tag ) )
+    if ( _totlower( *p ) == _totlower( *tag ) )
 	{
 		const TCHAR* q = p;
 
@@ -289,7 +290,7 @@ bool TiXmlBase::StringEqual( const TCHAR* p,
 		}
 		else
 		{
-			while ( *q && *tag && tolower( *q ) == tolower( *tag ) )
+			while ( *q && *tag && _totlower( *q ) == _totlower( *tag ) )
 			{
 				++q;
 				++tag;
@@ -338,7 +339,7 @@ const TCHAR* TiXmlBase::ReadText(	const TCHAR* p,
 				whitespace = true;
 				++p;
 			}
-			else if ( isspace( *p ) )
+			else if ( _istspace( *p ) )
 			{
 				whitespace = true;
 				++p;
@@ -533,7 +534,7 @@ TiXmlNode* TiXmlNode::Identify( const TCHAR* p )
 		#endif
 		returnNode = new TiXmlDeclaration();
 	}
-	else if (    isalpha( *(p+1) )
+	else if (    _istalpha( *(p+1) )
 			  || *(p+1) == '_' )
 	{
 		#ifdef DEBUG_PARSER
@@ -1013,7 +1014,7 @@ const TCHAR* TiXmlAttribute::Parse( const TCHAR* p, TiXmlParsingData* data )
 		// its best, even without them.
 		value = TEXT("");
 		while (    p && *p										// existence
-				&& !isspace( *p ) && *p != '\n' && *p != '\r'	// whitespace
+				&& !_istspace( *p ) && *p != '\n' && *p != '\r'	// whitespace
 				&& *p != '/' && *p != '>' )						// tag end
 		{
 			value += *p;
@@ -1126,7 +1127,7 @@ const TCHAR* TiXmlDeclaration::Parse( const TCHAR* p, TiXmlParsingData* data )
 		else
 		{
 			// Read over whatever it is.
-			while( p && *p && *p != '>' && !isspace( *p ) )
+			while( p && *p && *p != '>' && !_istspace( *p ) )
 				++p;
 		}
 	}
@@ -1136,7 +1137,7 @@ const TCHAR* TiXmlDeclaration::Parse( const TCHAR* p, TiXmlParsingData* data )
 bool TiXmlText::Blank() const
 {
 	for (size_t i = 0, len = value.length(); i < len; i++)
-		if ( !isspace( value[i] ) )
+		if ( !_istspace( value[i] ) )
 			return false;
 	return true;
 }
--- a/PowerEditor/src/WinControls/AnsiCharPanel/ansiCharPanel.cpp
+++ b/PowerEditor/src/WinControls/AnsiCharPanel/ansiCharPanel.cpp
@@ -129,7 +129,7 @@ void AnsiCharPanel::insertChar(unsigned char char2insert) const
 		bool isUnicode = ((*_ppEditView)->execute(SCI_GETCODEPAGE) == SC_CP_UTF8);
 		if (isUnicode)
 		{
-			MultiByteToWideChar(0, 0, charStr, -1, wCharStr, sizeof(wCharStr));
+			MultiByteToWideChar(0, 0, charStr, -1, wCharStr, _countof(wCharStr));
 			WideCharToMultiByte(CP_UTF8, 0, wCharStr, -1, multiByteStr, sizeof(multiByteStr), NULL, NULL);
 		}
 		else // ANSI
@@ -140,7 +140,7 @@ void AnsiCharPanel::insertChar(unsigned char char2insert) const
 	}
 	else
 	{
-		MultiByteToWideChar(codepage, 0, charStr, -1, wCharStr, sizeof(wCharStr));
+		MultiByteToWideChar(codepage, 0, charStr, -1, wCharStr, _countof(wCharStr));
 		WideCharToMultiByte(CP_UTF8, 0, wCharStr, -1, multiByteStr, sizeof(multiByteStr), NULL, NULL);
 	}
 	(*_ppEditView)->execute(SCI_REPLACESEL, 0, reinterpret_cast<LPARAM>(""));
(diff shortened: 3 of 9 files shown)


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -766,7 +766,9 @@ COLORREF getCtrlBgColor(HWND hWnd)
 
 generic_string stringToUpper(generic_string strToConvert)
 {
-    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), ::toupper);
+    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), 
+        [](TCHAR ch){ return static_cast<TCHAR>(::toupper(ch)); }
+    );
```
CookiePLMonster:
```
Warning	C4244	'=': conversion from 'int' to 'wchar_t', possible loss of data
```

PowerEditor/src/WinControls/AnsiCharPanel/ansiCharPanel.cpp
```diff
@@ -129,7 +129,7 @@ void AnsiCharPanel::insertChar(unsigned char char2insert) const
 		bool isUnicode = ((*_ppEditView)->execute(SCI_GETCODEPAGE) == SC_CP_UTF8);
 		if (isUnicode)
 		{
-			MultiByteToWideChar(0, 0, charStr, -1, wCharStr, sizeof(wCharStr));
+			MultiByteToWideChar(0, 0, charStr, -1, wCharStr, _countof(wCharStr));
```
CookiePLMonster:
```
Warning	C6386	Buffer overrun while writing to 'wCharStr':  the writable size is '20' bytes, but '40' bytes might be written.
```

(this argument specifies number of characters, not number of bytes).

PowerEditor/src/WinControls/AnsiCharPanel/ansiCharPanel.cpp
```diff
@@ -140,7 +140,7 @@ void AnsiCharPanel::insertChar(unsigned char char2insert) const
 	}
 	else
 	{
-		MultiByteToWideChar(codepage, 0, charStr, -1, wCharStr, sizeof(wCharStr));
+		MultiByteToWideChar(codepage, 0, charStr, -1, wCharStr, _countof(wCharStr));
```
CookiePLMonster: This one has been omitted by the static analysis, but still corrected it for the sake of completeness.

PowerEditor/src/WinControls/AnsiCharPanel/asciiListView.cpp
```diff
@@ -119,7 +119,7 @@ generic_string AsciiListView::getAscii(unsigned char value)
 			char ascii[2];
 			ascii[0] = value;
 			ascii[1] = '\0';
-			MultiByteToWideChar(_codepage, 0, ascii, -1, charStr, sizeof(charStr));
+			MultiByteToWideChar(_codepage, 0, ascii, -1, charStr, _countof(charStr));
```
CookiePLMonster:
```
Warning	C6386	Buffer overrun while writing to 'charStr':  the writable size is '20' bytes, but '40' bytes might be written.
```

(this argument specifies number of characters, not number of bytes).

PowerEditor/src/WinControls/DockingWnd/DockingSplitter.cpp
```diff
@@ -177,7 +177,7 @@ LRESULT DockingSplitter::runProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM
 			if (hookMouse)
 			{
 				::UnhookWindowsHookEx(hookMouse);
-				::SetCapture(NULL);
+				::ReleaseCapture();
```
CookiePLMonster:
```
Warning	C6387	'_Param_(1)' could be '0':  this does not adhere to the specification for the function 'SetCapture'.
```

PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
```diff
@@ -142,7 +142,7 @@ bool findStrNoCase(const generic_string & strHaystack, const generic_string & st
 	auto it = std::search(
 		strHaystack.begin(), strHaystack.end(),
 		strNeedle.begin(), strNeedle.end(),
-		[](char ch1, char ch2){return std::toupper(ch1) == std::toupper(ch2); }
+		[](TCHAR ch1, TCHAR ch2){return std::toupper(ch1) == std::toupper(ch2); }
```
CookiePLMonster:
```
Warning	C4244	'argument': conversion from 'const wchar_t' to 'char', possible loss of data
```

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -1359,7 +1359,7 @@ INT_PTR CALLBACK DefaultDirectoryDlg::run_dlgProc(UINT message, WPARAM wParam, L
 						TCHAR inputDir[MAX_PATH];
 						::SendDlgItemMessage(_hSelf, IDC_OPENSAVEDIR_ALWAYSON_EDIT, WM_GETTEXT, MAX_PATH, reinterpret_cast<LPARAM>(inputDir));
 						lstrcpy(nppGUI._defaultDir, inputDir);
-						::ExpandEnvironmentStrings(nppGUI._defaultDir, nppGUI._defaultDirExp, 500);
+						::ExpandEnvironmentStrings(nppGUI._defaultDir, nppGUI._defaultDirExp, _countof(nppGUI._defaultDirExp));
```
CookiePLMonster:
```
Warning	C6386	Buffer overrun while writing to 'nppGUI._defaultDirExp':  the writable size is '520' bytes, but '1000' bytes might be written.
```

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -467,10 +467,7 @@ INT_PTR CALLBACK Shortcut::run_dlgProc(UINT Message, WPARAM wParam, LPARAM)
 // return true if one of CommandShortcuts is deleted. Otherwise false.
 void Accelerator::updateShortcuts() 
 {
-	vector<int> incrFindAccIds;
-	incrFindAccIds.push_back(IDM_SEARCH_FINDNEXT);
-	incrFindAccIds.push_back(IDM_SEARCH_FINDPREV);
-	incrFindAccIds.push_back(IDM_SEARCH_FINDINCREMENT);
+	const unsigned long incrFindAccIds[] = { IDM_SEARCH_FINDNEXT, IDM_SEARCH_FINDPREV, IDM_SEARCH_FINDINCREMENT };
```
CookiePLMonster:
```Warning	C4389	'==': signed/unsigned mismatch
```

Since this vector always contains only those three entries, I went ahead and turned it into a plain array.

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -502,7 +499,7 @@ void Accelerator::updateShortcuts()
 			_pAccelArray[offset].key = shortcuts[i].getKeyCombo()._key;
 
 			// Special extra handling for shortcuts shared by Incremental Find dialog
-			if (std::find(incrFindAccIds.begin(), incrFindAccIds.end(), shortcuts[i].getID()) != incrFindAccIds.end())
+			if (std::find(std::begin(incrFindAccIds), std::end(incrFindAccIds), shortcuts[i].getID()) != std::end(incrFindAccIds))
```
CookiePLMonster: Changed so it can work on plain C arrays. Equivalent functionality wise otherwise.

PowerEditor/src/WinControls/shortcut/shortcut.cpp
chcg: should be possible to use std:array here instead of falling back to plain old c array

CookiePLMonster: You are absolutely right, I actually realized this as I reviewed it. Might as well... The only issue I have with it is that I have to hardcode the amount of entries, so must be done like

```std::array<unsigned long, 3> incrFindAccIds = { a, b, c };```

whereas with plain C array I don't need to do it.

File: PowerEditor/src/WinControls/AnsiCharPanel/ansiCharPanel.cpp (hunk 2 above)
Comment (chcg) [reply PowerEditor/src/WinControls/AnsiCharPanel/ansiCharPanel.cpp]: See caution note at https://msdn.microsoft.com/en-us/library/windows/desktop/dd319072(v=vs.85).aspx

File: PowerEditor/src/WinControls/DockingWnd/DockingSplitter.cpp (hunk 5 above)
Comment (chcg) [reply PowerEditor/src/WinControls/DockingWnd/DockingSplitter.cpp]: See https://msdn.microsoft.com/de-de/library/windows/desktop/ms646262(v=vs.85).aspx

File: PowerEditor/src/WinControls/Preference/preferenceDlg.cpp (hunk 7 above)
Comment (chcg) [reply PowerEditor/src/WinControls/Preference/preferenceDlg.cpp]: See https://msdn.microsoft.com/en-US/library/windows/desktop/ms724265(v=vs.85).aspx

nSize [in]

    The maximum number of characters that can be stored in the buffer pointed to by the lpDst parameter. When using ANSI strings, the buffer size should be the string length, plus terminating null character, plus one. When using Unicode strings, the buffer size should be the string length plus the terminating null character.

File: PowerEditor/src/MISC/Common/Common.cpp (hunk 1 above)
Comment (chcg) [reply PowerEditor/src/MISC/Common/Common.cpp]: See http://en.cppreference.com/w/cpp/string/byte/toupper why lamba expression is used.

ch  
character to be converted. If the value of ch is not representable as unsigned char and does not equal EOF, the behavior is undefined.

, but TCHAR is not just a char and therefore maybe 

http://en.cppreference.com/w/cpp/string/wide/towupper

matches better here.

Comment (CookiePLMonster) [reply PowerEditor/src/MISC/Common/Common.cpp]: Good point! MSDN points out `_totupper` exists

https://msdn.microsoft.com/pl-pl/library/45119yx3.aspx

in which case it would fit this code best. Correcting.

File: PowerEditor/src/MISC/Common/Common.cpp [hunk 10]
```diff
@@ -766,7 +766,9 @@ COLORREF getCtrlBgColor(HWND hWnd)
 
 generic_string stringToUpper(generic_string strToConvert)
 {
-    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), ::toupper);
+    std::transform(strToConvert.begin(), strToConvert.end(), strToConvert.begin(), 
+        [](TCHAR ch){ return static_cast<TCHAR>(_totupper(ch)); }
+    );
```
CookiePLMonster: In addition to previous changes, now `_totupper` is used, so it expands to `towupper` on `_UNICODE` compilations.

PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
```diff
@@ -142,7 +142,7 @@ bool findStrNoCase(const generic_string & strHaystack, const generic_string & st
 	auto it = std::search(
 		strHaystack.begin(), strHaystack.end(),
 		strNeedle.begin(), strNeedle.end(),
-		[](char ch1, char ch2){return std::toupper(ch1) == std::toupper(ch2); }
+		[](TCHAR ch1, TCHAR ch2){return _totupper(ch1) == _totupper(ch2); }
```
CookiePLMonster: In addition to previous changes, now `_totupper` is used, so it expands to `towupper` on `_UNICODE` compilations.

PowerEditor/src/WinControls/shortcut/shortcut.cpp
chcg: The solution would be http://en.cppreference.com/w/cpp/experimental/make_array, but to my feeling adding the size is still ok. Also because the compiler will detect if the initlist doesn't match with specified array size.

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -467,10 +468,7 @@ INT_PTR CALLBACK Shortcut::run_dlgProc(UINT Message, WPARAM wParam, LPARAM)
 // return true if one of CommandShortcuts is deleted. Otherwise false.
 void Accelerator::updateShortcuts() 
 {
-	vector<int> incrFindAccIds;
-	incrFindAccIds.push_back(IDM_SEARCH_FINDNEXT);
-	incrFindAccIds.push_back(IDM_SEARCH_FINDPREV);
-	incrFindAccIds.push_back(IDM_SEARCH_FINDINCREMENT);
+	const array<unsigned long, 3> incrFindAccIds = { IDM_SEARCH_FINDNEXT, IDM_SEARCH_FINDPREV, IDM_SEARCH_FINDINCREMENT };
```
CookiePLMonster: Now uses `std::array` as suggested.

CookiePLMonster:
Updated the PR with more instances of `isXXX` functions being used with `TCHAR` - replaced them with `_istXXX` functions.

NOTE: There are more cases like this, but I haven't addressed them all yet. Namely, Scintilla still likes to use `std::isspace`.

PowerEditor/src/TinyXml/tinyxmlparser.cpp
```diff
@@ -204,10 +205,10 @@ const TCHAR* TiXmlBase::ReadName( const TCHAR* p, TIXML_STRING * name )
 	// hyphens, or colons. (Colons are valid ony for namespaces,
 	// but tinyxml can't tell namespaces from names.)
 	if (    p && *p 
-		 && ( isalpha( (UCHAR) *p ) || *p == '_' ) )
+		 && ( _istalpha( *p ) || *p == '_' ) )
 	{
 		while(		p && *p
-				&&	(		isalnum( (UCHAR ) *p ) 
+				&&	(		_istalnum( *p )
```
CookiePLMonster: Notice a nasty truncation typecast is now gone.

CookiePLMonster:
UPDATE: `std::isspace` is templated, so no need to change anything. Thus, I think all offending cases have been addressed now!

Issue #4015, opened 2017-12-20
[Feature Request] Document Switcher Reordering/Pinning


Andronomos:
I would like to request the ability to reorder/pin documents in the document switcher. As a web developer/programmer on a normal day I can have 30+ files open. It's very frustrating when I've memorized the locations of documents in the list only to accidentally on the click doc switch title bar and have them all be reordered.

Having the ability to pin a document in place or be able to re order them by dragging would greatly enhance my productivity and lower my frustration while working in Notepad++.

nebosite:
+1

I use the switcher to manage the 20+ documents I have open and position is important to me.

nebosite:
Duplicate of: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1684

Issue #1684, opened 2016-03-26
[Feature Request] Doc Switcher Rearrangement


ghost:
The Doc Switcher gets a little annoying when you have to close and re-open documents for them to appear the way you want them to be arranged. The ability to rearrange the documents in it would be nice.

cajuncool:
I support this feature request.

You can sort the files in the list alphabetically in ascending or descending order, but you can never get back to the original order (as loaded or manually rearranged)!
So there should be another feature: Go back to the original/manual order.
Thank you!

nebosite:
Please please do this.   The switcher is already a great feature and this would put it over the top.

Andronomos:
Yes, this feature would greatly improve upon Notepad++. I already can't see myself using anything else and the lack of this feature is really my only complaint.

pidgeon777:
Also interested.

PR #4173, opened 2018-02-12
Keep Doc Switcher's Tab ordering in sync

### LINKED ISSUE #946: Doc Switcher: drag&drop rearranging
Just like user can drag&drop tabs to rearrange them, so could Doc Switcher allow drag&drop to rearrange.
Doc switcher could remember document order, just like tabs remember the order.
Doc Switcher could be a decent alternative to tabs in this regard.

### LINKED ISSUE #2342: Feature Request: tab re-ordering also re-orders Doc Switcher files
### Description of the Issue

I use Doc Switcher (aka Document List Panel) nearly all the time as a way to quickly bring a doc tab to focus. I also often move doc tabs around; I tend to actively work with tabs that are right-most whereas left-most tabs are things pending, temporarily sidelined, or reference docs. This re-ordering of tabs does not also re-order the Doc Switcher doc names. This makes Doc Switcher much less useful as, without re-ordering anything, the Doc Switcher order matches the tab order - top Doc Switcher file is left-most tab, bottom file is right-most tab. If these could kept in sync the Doc Switcher would be even more useful. Perhaps it could be an option somewhere.

1. Open 2 or more files.
2. Make the Doc Switcher panel appear: Settings > Preferences > General > Document List Panel > Show [X]
3. Move one of the files' tabs to a new position. Notice that the Doc Switcher order is unchanged.
### Desired Behavior

Ideally the Doc Switcher order would mirror the visible tab order, regardless of where one moves a doc's tab.
### Actual Behavior

Doc Switcher order does not change in response to moving doc tabs around.

Here's a capture before moving tabs around:
npp_doc_switcher_pre_tab_move
And here's one after switching around the two tabs:
npp_doc_switcher_post_tab_move
Notice the Doc Switcher order remains the same. If it could change in response to tab moves life would improve markedly here.

Thanks for reading and for Notepad++ - it rocks.

-csl

AngryGamer:
Closes #946 #1684 #2342 #4015

The patch introduces an "unsorted" direction (up, down, unsorted, up, down, unsorted, etc) which keeps in sync with the Tab Bar's ordering, but only while it's set to unsorted. This allows people to rearrange the Doc Switcher's files using the Tab Bar.

Changes Overview:

- Added "unsorted" sorting option on columns
- Tab order sync'd with Tab Bar while set to "unsorted"
- Sorting by "name" is preserved while toggling "exts" column
- Column widths no longer reset to default while toggling "exts" column (will resize if "exts" wont fit)

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -458,6 +458,11 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_INTERNAL_DOCORDERCHANGED :
 		{
+			if (_pFileSwitcherPanel)
+			{
+				_pFileSwitcherPanel->updateTabOrder();
+			}
+			
 			BufferID id = _pEditView->getCurrentBufferID();
 
 			// Notify plugins that current file is about to be closed
--- a/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcher.cpp
+++ b/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcher.cpp
@@ -34,25 +34,40 @@
 
 int CALLBACK ListViewCompareProc(LPARAM lParam1, LPARAM lParam2, LPARAM lParamSort)
 {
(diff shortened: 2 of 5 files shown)


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/883045d96fca53197e0d164a802db4b0171cedb7

PR #4175, opened 2018-02-13
Update chineseSimplified.xml

popcorner:
**What I do in this commit:**
- translate all the new strings till now (and will add translation for new strings before the next version)
- convert some indentation, just like 1f7f00b does.
- use more appropriate word for the encoding menu.
- use the same way to describe UTF-8 as english.xml


donho:
@popcorner Could you check the newest updated english.xml here:

popcorner:
@donho new strings added

PR #4178, opened 2018-02-13
enhance ShortcutMapper behavior while resizing:

cmeriaux:
- keep centered buttons while resizing
- set minimum width and height

capture_shortcut_mapper

### DIFF
--- a/PowerEditor/src/WinControls/Grid/ShortcutMapper.h
+++ b/PowerEditor/src/WinControls/Grid/ShortcutMapper.h
@@ -39,6 +39,7 @@ class ShortcutMapper : public StaticDialog {
 public:
 	ShortcutMapper() : _currentState(STATE_MENU), StaticDialog() {
 		_shortcutFilter = TEXT("");
+		_dialogInitDone = false;
 	};
 	~ShortcutMapper() {};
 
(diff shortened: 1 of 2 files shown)


cmeriaux:
fixes #4179

PowerEditor/src/WinControls/Grid/ShortcutMapper.h
```diff
@@ -100,6 +101,9 @@ protected :
 	};
 	LONG _clientWidth;
 	LONG _clientHeight;
+	LONG _initClientWidth;
+	LONG _initClientHeight;
+	boolean _dialogInitDone;
```
donho: Why boolean instead of bool ?

cmeriaux: done

cmeriaux:
@donho It's done.

dinkumoil:
The story still isn't finished. There is a bug in the calculation of the grid's height during vertical resizing, mostly the grid overlaps the description field underneath.

The problem is that the height of the grid always is a multiple of one grid line whereas the height of the dialog and thus the position of the description field can be adjusted pixel-wise.

grafik

Issue #3851, opened 2017-10-26
Tab issue


Zonekeeper26:
### Description of the Issue
When I click on the last tab in my tab bar when it's partially visible in the bar, the tab is 'picked up' as if dragging and dropping, and it does not release until I click somehwere

### Steps to Reproduce the Issue

1. Open enough tabs that you must scroll to see all of them. (or resize your window smaller than the length of the tab bar)
2. Scroll to where the last tab is partially visible. ( the filename may need to be long enough for this to happen)
3. Click the last tab while partially visible, and it acts like you're dragging the tab.

### Expected Behavior

When the tab is clicked, it shows the document and the tab does not get dragged.

### Actual Behavior

The tab acts as if it is being dragged despite not holding down the left mouse button.

### Debug Information

Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Path : C:\Users\<redacted>\Documents\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : none

Ankeraout:
Confirmed here on Windows 7 x64

thanhminhmr:
Confirm here.
~~~~
Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 
~~~~

tuyethoa08041997:
Just test this bug, it still there...

Ankeraout:
Check for updates, I have not experienced this bug for a long time now.

tuyethoa08041997:
Just checking for update, and it still there...
```
Notepad++ v7.5.4   (32-bit)
Build time : Jan  1 2018 - 01:47:45
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll DSpellCheck.dll JSMinNPP.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll Tidy2.dll 

```

[...]

PR #4182, opened 2018-02-14
Fix drag and drop erroneously being enabled

### LINKED ISSUE #4122: Tab sticks to mouse pointer after external update of a file opened in NPP
### Description of the Issue
After a file open in NPP is updated externally, its tab sticks to the mouse pointer if NPP is activated by clicking among the many tabs I have open.

1. Open multiple files in NPP, resulting in multiple tabs.
2. Externally change and save a file that is currently open in NPP.
3. Click on some random NPP tab when NPP doesn't have focus (it doesn't seem to matter which tab) .
4. Acknowledge the "Reload changed file" dialog.
5. Move the mouse pointer over NPP tabs.

### Expected Behavior
I expected nothing to happen.

### Actual Behavior
The tab for the changed file sticks to the mouse pointer until I click, usually resulting in the tab being moved somewhere else in the tab list.

### Debug Information
Notepad++ v7.5.4   (64-bit)
Build time : Jan  1 2018 - 01:50:29
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll 

The problem is repeatable and consistent.
By the way, Notepad++ is WONDERFUL! Thanks for a delightful text editor.

AngryGamer:
Fixes #4122 and probably #3851

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -634,7 +634,13 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 		{
 			if (_mightBeDragging && !_isDragging)
 			{
-				if (++_dragCount > 2)
+				// Grrr! Who has stolen focus and eaten the WM_LBUTTONUP?!
+				if (GetKeyState(VK_LBUTTON) >= 0)
+				{
+					_mightBeDragging = false;
+					_dragCount = 0;
+				}
+				else if (++_dragCount > 2)
 				{
 					int tabFocused = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURFOCUS, 0, 0));
 					int tabSelected = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0));


PowerEditor/src/WinControls/TabBar/TabBar.cpp
```diff
@@ -634,7 +634,13 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 		{
 			if (_mightBeDragging && !_isDragging)
 			{
-				if (++_dragCount > 2)
+				// Grrr! Who has stolen focus and eaten the WM_LBUTTONUP?!
+				if (GetKeyState(VK_LBUTTON) >= 0)
```
SinghRajenM: ##
[MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646301(v=vs.85).aspx) says:
Return value

Type: SHORT

The return value specifies the status of the specified virtual key, as follows:
    

- If the high-order bit is 1, the key is down; otherwise, it is up.
- If the low-order bit is 1, the key is toggled. A key, such as the CAPS LOCK key, is toggled if it is turned on. The key is off and untoggled if the low-order bit is 0. A toggle key's indicator light (if any) on the keyboard will be on when the key is toggled, and off when the key is untoggled.

##

So I feel, high-order bit should be checked to determine. Key is pressed or not.

AngryGamer: The return is a signed short. ON and Toggle ON will both be negative values, while OFF & Toggle OFF is 0 and 1 respectively

PR #4196, opened 2018-02-16
Update occitan.xml localisation to v.7.5.5

Louis-Berlic:
I improved existing translations and added new ones based on english.xml file.


donho:
@Louis-Berlic Could you check the newest updated english.xml here:

Louis-Berlic:
Done!

pfparesseux000:
Based on english.xml, items id="45005, id="45008, id="45010" and id="45011" changed because of the issue 4040. In this occitan.xml, the translated sentences are obsolete.

PR #4136, opened 2018-02-02
Update turkish.xml

Tmp341:
Updating to latest.

PR #4197, opened 2018-02-17
Update Turkish localization

Tmp341:
I've checked every english localization changes. I hope i haven't done any mistake. Closing #4136 as it is pretty much outdated.


donho:
@Tmp341 Could you check the newest updated english.xml here:

PR #4200, opened 2018-02-18
Update catalan.xml to v7.5.5

Hiro5:
Update to v.7.5.5.

Note that I saw small mistakes in english.xml (I think)
<ind-status-replaceall-readonly ...>
probably a typo, should be:
<find-status-replaceall-readonly ...>

and also, the line
<find-status-mark-re-malformed...>
is repeated twice.

Probably that should be fixed in english.xml.

Regards.

### DIFF
--- a/PowerEditor/installer/nativeLang/catalan.xml
+++ b/PowerEditor/installer/nativeLang/catalan.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!--
 Catalan translation for Notepad++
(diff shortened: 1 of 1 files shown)


donho:
@Hiro5 Could you check the newest updated english.xml here:

which has been updated by this commit:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/bd0867ee62b115b5f485abd8791da775c001ff8d

Hiro5:
Hi @donho 
I think it's OK now.
Not too much changes, since some of them I already did it the first time.

Regards.

PR #4203, opened 2018-02-19
A few corrections (. . ., "?", " + ", "CMD")

scootergrisen:
Not sure if there should be space before "?" in english but judging from the other software i translate im guessing there should.
"Cmd" with small "md" is used in others strings.


CookiePLMonster:
There should be no space before question marks.

scootergrisen:
Perhaps the title "Are you sure?" should be something else as the message is also a question:
`<SCMapperDoDeleteOrNot title="Are you sure?" message="Are you sure you want to delete this shortcut?"/>`
Maybe "Confirm delete/Confirmation/Delete shortcut" something like that.

PR #4219, opened 2018-02-24
Make column names translatable in Window->Window...

AngryGamer:


PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.h
```diff
@@ -30,6 +30,7 @@
 
 #include "SizeableDlg.h"
 #include "Common.h"
+#include "localization.h"
```
donho: All the header files `*.h` should try to add in source file `.cpp` as possible.

AngryGamer: done

PR #4226, opened 2018-02-26
File status detection under Windows XP (32-bit only)

d0vgan:
File status detection under Windows XP: using winapi_wstat for 32-bit only.
P.S. I don't have any idea why git shows changes in the functions isCertificateValidated and isAssoCommandExisting - I did not touch them at all!

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -34,6 +34,7 @@
 
 #include "Common.h"
 #include "../Utf8.h"
+#include <Parameters.h>
 
 WcharMbcsConvertor* WcharMbcsConvertor::_pSelf = new WcharMbcsConvertor;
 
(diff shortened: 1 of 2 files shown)


dail8859:
> P.S. I don't have any idea why git shows changes in the functions isCertificateValidated and isAssoCommandExisting - I did not touch them at all!

It appears this is due to the file originally having *mixed line endings*. As you can see from the screenshot below, opening the original file (on the left) and the new file from this PR (on the right) Notepad++ shows different line endings.

mle

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1158,38 +1158,84 @@ bool isCertificateValidated(const generic_string & fullFilePath, const generic_s
...
+    return (ull.QuadPart / 10000000ULL - 11644473600ULL);
+}
+
+int winapi_wstat(wchar_t const* _FileName, struct _stat* _Stat)
+{
+    // Starting from Visual Studio 2015, _wstat always returns -1 in Windows XP.
+    // So here is a WinAPI-based implementation of _wstat.
+    int nResult = -1;
+    DWORD dwAttr = ::GetFileAttributes(_FileName);
+    if (dwAttr != INVALID_FILE_ATTRIBUTES)
```
CookiePLMonster: Since you're checking for CreateFile return value, this check is redundant.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1158,38 +1158,84 @@ bool isCertificateValidated(const generic_string & fullFilePath, const generic_s
...
+            FILETIME creationTime, accessTime, writeTime;
+            if (::GetFileSizeEx(hFile, &fileSize) &&
+                ::GetFileTime(hFile, &creationTime, &accessTime, &writeTime))
+            {
+                ::ZeroMemory(_Stat, sizeof(struct _stat));
+                _Stat->st_atime = static_cast<decltype(_Stat->st_atime)>(filetime_to_time_ull(&accessTime));
+                _Stat->st_ctime = static_cast<decltype(_Stat->st_ctime)>(filetime_to_time_ull(&creationTime));
+                _Stat->st_mtime = static_cast<decltype(_Stat->st_mtime)>(filetime_to_time_ull(&writeTime));
+                _Stat->st_size = static_cast<decltype(_Stat->st_size)>(fileSize.QuadPart);
+                _Stat->st_mode = _S_IREAD | _S_IEXEC;
```
CookiePLMonster: Are directories executable?

PowerEditor/src/MISC/Common/Common.cpp
d0vgan: The intention was: if GetFileAttributes returns INVALID_FILE_ATTRIBUTES, we don't even call CreateFile. These are two different functions, so checking results of both sounds logical, no?

PowerEditor/src/MISC/Common/Common.cpp
d0vgan: As https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html says:
S_IEXEC : Execute (for ordinary files) or search (for directories) permission bit for the owner of the file

PowerEditor/src/MISC/Common/Common.cpp
CookiePLMonster: Sure, but you're not winning anything with this check. If the file/directory does not exist, `CreateFile` will fail anyway. Right now you are checking for file existence twice, and if `GetFileAttributes` returns valid data you still can't be sure that the file will still exist when `CreateFile` gets called (may very well be deleted between those two calls).

donho:
@d0vgan Please rebase from master for solving the conflict.

bahusoid:
Is it about this [issue](https://developercommunity.visualstudio.com/content/problem/19317/windows-c-function-wstat-does-not-work-correctly-w.html)? It seems it's fixed in latest VS 2017 platform toolset. So maybe it's better just to upgrade to VS 2017? Also there is some workaround suggested for existing project:

> First, unload the project by right clicking it and selecting "Unload Project". Then right click again and select "Edit myproject.vcxproj". The vcxproj is divided into property groups for each configuration. Everywhere that you have <PlatformToolset>v141_xp</PlatformToolset> (or any other XP toolset), you can add this tag in the same section: <TargetUniversalCRTVersion>10.0.10586.0</TargetUniversalCRTVersion>. You can use any SDK version in this section. This issue was fixed in 10.0.10586.0, but I recommend checking the SDK archive (https://developer.microsoft.com/en-us/windows/downloads/sdk-archive), downloading the latest SDK,

d0vgan:
The solution with <TargetUniversalCRTVersion>10.0.10586.0</TargetUniversalCRTVersion> does not work with Visual Studio 2015 at all because <TargetUniversalCRTVersion>10.0.10586.0</TargetUniversalCRTVersion> is not available in it!
I believe it will work when Visual Studio 2017 or additional SDK is installed, but, unfortunately, not with the original Visual Studio 2015.

d0vgan:
Actually, why not make this behavior _optional_? I mean, basing on this change, introduce a new option e.g. "WinXP-compatible file status" in the Settings dialog. And this option will alter the behavior of the function generic_stat. What do you think?
(Though it will probably mean the generic_stat function will need an additional argument e.g. isWinXPCompatible.)

AngryGamer:
I don't see a reason to waste space in the Settings dialog. If the objective is to always use _wstat on non-XP 32bit, then might as well do something like:
```
int winapi_wstat(wchar_t const* _FileName, struct _stat* _Stat)
{
	if (IsWindowsVistaOrGreater())
		return _wstat(_FileName, _Stat);
	
	// rest of code
}
```

donho:
@d0vgan 
Hi Vitaliy, Hope you are fine :)
Just wanna make the situation more clear:
 _wstat (of VS2015) has bugs only under windows XP 32 bits, but not under others platforms, is it correct?

However in your PR, you're using your implementation `winapi_wstat` instead of Microsoft's implementation `_wstat32` or `_wstat64i32` for all  32 bits platforms. That is overkilled IMO and might insert the unstable issue into Notepad++ for all the 32 bits platforms.

d0vgan:
OK, I've rebased the changes, plus now they are only for Windows XP (and 2000/2003 if someone succeds to compile Notepad++ under it.).

AngryGamer:
See https://msdn.microsoft.com/en-us/library/windows/desktop/ms724451(v=vs.85).aspx

> [GetVersionEx may be altered or unavailable for releases after Windows 8.1. Instead, use the Version Helper functions]

CookiePLMonster:
GetVersionEx being deprecated isn't really an issue - changes of it dissapearing are close to none, and it's a perfectly legimate way of detecting systems as old as Windows XP (could only cause issues if you try to detect 8.1 or 10, but I believe notepad++ is already manifested for Windows 10 so even that is fine).

d0vgan:
Updated according to the review comments: now using NppParameters::getInstance()->getWinVersion().

d0vgan:
Rebased against the master after handling the comments. Ready to merge, I suppose.

PR #4242, opened 2018-02-28
Plugins cannot know file status (RO or snapshot dirty) while Notepad++ start up (Fix #4201)

### LINKED ISSUE #4201: plugins cannot show the correct status of the opened files
plugins cannot show the correct status of the opened files after startup. 
The plugins are not notified until the status of a file changes after opening npp.
The plugins must be able to check the correct status of the files when loading.
For this purpose, there are no query possibilities for readonly modified and dirty flag of the files
show the Plugin Windowsmanager and Navigate To

open a session with readonly Files and you see in this Plugins all Files are writable.
Please add following NPP-Messages:
NPPM_GETREADONLYFROMPOS
NPPM_GETMODIFYFROMPOS
NPPM_GETDIRTYROMPOS

KlausiD:
plugins cannot show the correct status of the opened files after startup.
The plugins are not notified until the status of a file changes after opening npp.
The plugins must be able to check the correct status of the files when loading.
For this purpose, there are no query possibilities for readonly modified and dirty flag of the files
show the Plugin Windowsmanager and Navigate To

see issuse #4201


dail8859:
I haven't looked into the code but what is the difference between a buffer being "dirty" and being "modified"?

"dirty" meaning unsaved
"modified" meaning ....?

KlausiD:
"modified" is a not saved modified Document.
an example of dirty:
1. enable session snapshot ...   (at settings/Backup Page)
2. modified a open document
   or open a new document (button new) and edit it.
3. close npp but do not save the modified dokument!

now open npp again.
npp loading now the backupfile!
This file is now not modify but dirty!
Only if you change this Document at this session, now it is modified.

d0vgan:
I'd prefer to use BufferID rather than Pos. Please look at the existing messages such as NPPM_RELOADBUFFERID, NPPM_GETBUFFERENCODING and so on - they all use BufferID. And the NPPN_* notifications use BufferID as well.

d0vgan:
And I'd propose to use the message names of NPPM_ISBUFFERDIRTY and NPPM_ISBUFFERREADONLY.
Also, I believe that NPPM_ISBUFFERDIRTY can be used to get the both "dirty" and "modified" state by returning e.g. 1 ("dirty") or 2 ("modified").

KlausiD:
or is better one message NPPM_GETBUFFERSTATUS for all 3 Infos?
returning:
0 -> is read and writable
1 -> readonly
2 -> dirty
3 -> modified
as soon as this is clarified, i can revise the code.

d0vgan:
Agree, NPPM_GETBUFFERSTATUS makes sense. I assume there will be bit flags e.g.
BUFSTAT_READONLY = 0x01,
BUFSTAT_MODIFIED = 0x02,
BUFSTAT_LOADEDDIRTY = 0x03.
I've proposed a name BUFSTAT_LOADEDDIRTY for easier distinguishing between a file being modified while editing it and file being loaded "dirty" from the session snapshot.

d0vgan:
Oops, surely I mean:
BUFSTAT_LOADEDDIRTY = 0x04 
as the NPPM_GETBUFFERSTATUS is expected to return the combination of these bit flags according to the current buffer's (file) status.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -500,6 +500,60 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
+			DocTabView* pView = nullptr;
+
+			if (lParam == MAIN_VIEW)
+				pView = &_mainDocTab;
+			else if (lParam == SUB_VIEW)
+				pView = &_subDocTab;
+			else
+				return 0;
+
+			if ((size_t)wParam < pView->nbItem())
```
MAPJe71: Use C++ style type-casting.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -500,6 +500,60 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
```
MAPJe71: Likewise.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -500,6 +500,60 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
```
MAPJe71: Likewise.

KlausiD:
now i have replace the 3 message to one message (see the commits 078510d)
the funktion getBufferByID() should still be revised, since it only makes a cast and does not check the parameter bufferId. npp crashed when the bufferId is invalid.

KlausiD:
now i have update it to newest version!

KlausiD:
merge with master

donho:
[The notification `NPPN_SNAPSHOTDIRTYFILELOADED`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h#L591) is already there for plugins to notify plugins that a snapshot dirty file is loaded on startup. 
 The added message is not necessary.

d0vgan:
Don, the potential usages of `NPPN_SNAPSHOTDIRTYFILELOADED` and the proposed `NPPM_GETBUFFERSTATUS` are different.

Let's consider the following scenario:
You want to initiate file saving operation from within a plugin, and you want to know whether you need to physically save a file _at the current moment_.
With `NPPN_SNAPSHOTDIRTYFILELOADED`, you know that the file was loaded dirty _at Notepad++ startup_. To know the _current_ state of this file, you need to:
1. save the _initial_ "loaded dirty" state within a plugin;
2. constantly track Notepad++'es notifications for this file to understand whether a file is _currently_ saved or not;
3. if such file is closed, there is no need to track Notepad++'es notifications any more;
4. if such file is renamed or updated externally... - well, a plugin needs to decide what to do.

I.e. a plugin needs to do additional work (save the initial state + watch for all notifications) for all the files that were loaded dirty initially.
Instead, `NPPM_GETBUFFERSTATUS` would allow to get the _current_ status of the file _in that exact moment when a plugin is interested in it_.

Do you see the difference?

PR #4247, opened 2018-03-01
add translation for entering Administrator mode

popcorner:
it seems that the translation of administrator mode message box is not included in the language file.


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ecba7be9e488357b6de0116e5ce051d1ce332c6e

PR #4257, opened 2018-03-03
Fixed a piece of code

SinghRajenM:
**NppIO.cpp:**
Reduce the scope of local variable.

**ReadDirectoryChangesPrivate.cpp:** 
This seems to be a typo. Just think what if ```wstrFilename``` contains trailing backslash then what this statement ```wstrFilename = m_wstrDirectory + wstrFilename``` will do. 
But actual issue comes when ```m_wstrDirectory``` does not contains trailing backslash while ```wstrFilename``` contains.  
P.S. The probability of containing trailing backslash in ```wstrFilename``` is almost 0, that is the reason the piece of code is working fine so far.

### DIFF
--- a/PowerEditor/src/WinControls/ReadDirectoryChanges/ReadDirectoryChangesPrivate.cpp
+++ b/PowerEditor/src/WinControls/ReadDirectoryChanges/ReadDirectoryChangesPrivate.cpp
@@ -149,7 +149,7 @@ void CReadChangesRequest::ProcessNotification()
 
 		CStringW wstrFilename(fni.FileName, fni.FileNameLength/sizeof(wchar_t));
 		// Handle a trailing backslash, such as for a root directory.
-		if (wstrFilename.Right(1) != L"\\")
+		if (m_wstrDirectory.Right(1) != L"\\")
 			wstrFilename = m_wstrDirectory + L"\\" + wstrFilename;
 		else
 			wstrFilename = m_wstrDirectory + wstrFilename;
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -77,12 +77,12 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 			case WAIT_OBJECT_0 + 1:
 				// We've received a notification in the queue.
 			{
-				DWORD dwAction;
-				CStringW wstrFilename;
 				if (changes.CheckOverflow())
 					printStr(L"Queue overflowed.");
 				else
 				{
+					DWORD dwAction;
+					CStringW wstrFilename;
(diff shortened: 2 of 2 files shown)


PowerEditor/src/WinControls/ReadDirectoryChanges/ReadDirectoryChangesPrivate.cpp
```diff
@@ -149,7 +149,7 @@ void CReadChangesRequest::ProcessNotification()
 
 		CStringW wstrFilename(fni.FileName, fni.FileNameLength/sizeof(wchar_t));
 		// Handle a trailing backslash, such as for a root directory.
-		if (wstrFilename.Right(1) != L"\\")
+		if (m_wstrDirectory.Right(1) != L"\\")
```
donho: Is this modification intended to fix any bug?

SinghRajenM: Actually no bug has been seen so far, during debugging this piece of code, I came to know that this part should be modified. Refer debugger info as below for currently being monitored file (D:\Test\Test.txt)-

image

image

SinghRajenM: Also, as per existing code, refer the below screenshot 1 and with updated code screenshot 2.

**With existing code:**
image

**With Updated code:**
image

MAPJe71: So lines 89 to 93 of `NppIO.cpp` can be removed as their intended use (i.e. fix problem of monitoring files which are in root) are now (correctly) solved with this patch?

SinghRajenM: @MAPJe71, Yes, these lines can be removed. But there is no harm even if we keep those.

SinghRajenM:
Volla!

I was going through the [link](http://qualapps.blogspot.com/2010/05/understanding-readdirectorychangesw.html) given in source code and respective source code is located on [Git repo](https://github.com/jimbeveridge/readdirectorychanges/commit/2d9bc8c6003d6823b80beeaffbfb5fe2d760d352). Found there, same fix has been accommodated. 

image

PowerEditor/src/WinControls/ReadDirectoryChanges/ReadDirectoryChangesPrivate.cpp
MAPJe71: Might be wise to add some (extra) comment to these lines to prevent questions/confusion in the future.

PR #4272, opened 2018-03-07
Update thai.xml

orbitalz:
more translation on Thai.


donho:
@orbitalz Thank you for updating.
Could you update also the other entries please?
There was a major update recently here:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4286

donho:
Feel free to reopen it when the translation is done completely.

PR #4310, opened 2018-03-16
Update german

schnurlos:
Following https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fab3db50dfe68f2e94b5870e9d4af8d4bff751cd


donho:
FYI, there are 2 entries `DroppingFolderAsProjectModeWarning` in German localization.

PR #4320, opened 2018-03-19
Get rid of ATL dependency

d0vgan:
The ATL dependency is quite annoying when working in Visual Studio Express (where there's no ATL included). Moreover, currently it's easy to get rid of the ATL dependency at all - actually it does not add anything to the code that really worth this dependency.

### DIFF
--- a/PowerEditor/src/WinControls/ReadDirectoryChanges/ThreadSafeQueue.h
+++ b/PowerEditor/src/WinControls/ReadDirectoryChanges/ThreadSafeQueue.h
@@ -26,6 +26,7 @@
 //	http://qualapps.blogspot.com/2010/05/understanding-readdirectorychangesw.html
 //	See ReadMe.txt for overview information.
 
+#include "mutex.h"
 #include <list>
 
 template <typename C>
@@ -51,14 +52,15 @@ public:
 
 	void push(C& c)
 	{
-		CComCritSecLock<CComAutoCriticalSection> lock( m_Crit, true );
-		push_back( c );
-		lock.Unlock();
+		{
+			Yuni::MutexLocker lock( m_Mutex );
+			push_back( c );
+		}
 
 		if (!::ReleaseSemaphore(m_hSemaphore, 1, NULL))
 		{
 			// If the semaphore is full, then take back the entry.
-			lock.Lock();
+			Yuni::MutexLocker lock2( m_Mutex );
 			pop_back();
 			if (GetLastError() == ERROR_TOO_MANY_POSTS)
 			{
@@ -69,7 +71,7 @@ public:
 
 	bool pop(C& c)
 	{
-		CComCritSecLock<CComAutoCriticalSection> lock( m_Crit, true );
+		Yuni::MutexLocker lock( m_Mutex );
 
 		// If the user calls pop() more than once after the
 		// semaphore is signaled, then the semaphore count will
@@ -90,7 +92,7 @@ public:
 	// If overflow, use this to clear the queue.
 	void clear()
 	{
-		CComCritSecLock<CComAutoCriticalSection> lock( m_Crit, true );
+		Yuni::MutexLocker lock( m_Mutex );
 
 		for (DWORD i=0; i<size(); i++)
 			WaitForSingleObject(m_hSemaphore, 0);
@@ -110,7 +112,7 @@ public:
 protected:
 	HANDLE m_hSemaphore;
 
-	CComAutoCriticalSection m_Crit;
+	Yuni::Mutex m_Mutex;
 
 	bool m_bOverflow;
 };
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -82,9 +82,9 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 				else
 				{
 					DWORD dwAction;
-					CStringW wstrFilename;
+					std::wstring wstrFilename;
 					changes.Pop(dwAction, wstrFilename);
-					generic_string fn = wstrFilename.GetString();
+					generic_string fn = wstrFilename;
 
 					// Fix monitoring files which are under root problem
 					size_t pos = fn.find(TEXT("\\\\"));
--- a/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
+++ b/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
@@ -1336,7 +1336,7 @@ DWORD WINAPI FolderUpdater::watching(void *params)
 			// We've received a notification in the queue.
 			{
 				DWORD dwAction;
-				CStringW wstrFilename;
+				std::wstring wstrFilename;
 				if (changes.CheckOverflow())
 					printStr(L"Queue overflowed.");
 				else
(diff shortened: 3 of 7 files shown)


PowerEditor/src/WinControls/ReadDirectoryChanges/ThreadSafeQueue.h
```diff
@@ -28,6 +28,70 @@
 
 #include <list>
 
+class CCriticalSection
```
d0vgan: A very intuitive class, with its implementation quite similar to ATL's CComAutoCriticalSection. (Actually, there is nothing "Com" in it.)

PowerEditor/src/WinControls/ReadDirectoryChanges/ThreadSafeQueue.h
```diff
@@ -28,6 +28,70 @@
...
+        ::LeaveCriticalSection(&m_cs);
+    }
+
+    CCriticalSection(const CCriticalSection&) = delete;
+    CCriticalSection* operator=(const CCriticalSection&) = delete;
+protected:
+    CRITICAL_SECTION m_cs;
+};
+
+class CCriticalSectionLock
```
d0vgan: The implementation is quite similar to ATL's CComCritSecLock.

MAPJe71:
Why implement a new locking mechanism when there is already one available in `.\src\MISC\common\mutex.*`?

d0vgan:
Agree, Yuni::Mutex could be used instead. In such case I'll add a new class Yuni::MutexLockerEx that will include the m_bLocked member and additional methods Lock and Unlock - similar to CComCritSecLock.

MAPJe71:
There's already a Yuni::MutexLocker class. Why not use that one?

d0vgan:
@MAPJe71, please look at the implementation of Yuni::MutexLocker and then re-read my previous comment :) This is because of the following line in CThreadSafeQueue::push :
lock.Unlock();
So it's kind of lock_guard with additional manual methods Lock and Unlock.

MAPJe71:
Dispite using an autolocker/lock-guard on a mutex you can still manually (un-)lock it e.g.
```C++
template <typename C>
class CThreadSafeQueue : protected std::list<C>
{
public:

	void push(C& c)
	{
		Yuni::MutexLocker lock(m_Mutex); //<-- auto lock (RAII)
		push_back( c );
		m_Mutex.unlock(); //<-- manual unlock

		if (!::ReleaseSemaphore(m_hSemaphore, 1, NULL))
		{
			// If the semaphore is full, then take back the entry.
			m_Mutex.lock(); //<-- manual lock
			pop_back();
			if (GetLastError() == ERROR_TOO_MANY_POSTS)
			{
				m_bOverflow = true;
			}
		}
	} //<-- auto unlock

	bool pop(C& c)
	{
		Yuni::MutexLocker lock(m_Mutex);

        // 
	}

protected:
	Yuni::Mutex m_Mutex;
};
```

d0vgan:
@MAPJe71,
I bet you've already noticed what's wrong with that code.
Surely, it's this situation:
~~~
{
    Yuni::MutexLocker lock(m_Mutex); //<-- auto lock (RAII)
    m_Mutex.unlock(); //<-- manual unlock
} //<-- auto unlock - shit, unlocked twice!
~~~

MAPJe71:
Oooh, d#mn it you're right that won't work (as intended) 😁

@donho and @milipili could you extend the Yuni::MutexLocker class with manual un-/lock?
Would you consider a PR for [libyuni](https://github.com/libyuni/libyuni)?

d0vgan:
I'd rather propose to have 2 classes: MutexLocker and MutexLockerEx, without inheritance.
Here is why:
1) MutexLockerEx will have an additional bool variable (see the proposed CCriticalSectionLock implementation), which is simply not needed for a lightweight MutexLocker.
2) MutexLockerEx will have an additional argument in the constructor (see the proposed CCriticalSectionLock implementation), which is simply not needed for a lightweight MutexLocker.

d0vgan:
OK, guys - no additional classes now. Just a little thinking and a little bit more braces :)

d0vgan:
Now I believe it's good enough.

chcg:
@d0vgan additional benefit: building with mingw would be possible again, because ATL is not available there, see  #4145 and there #2478 with patches to avoid ATL, see https://github.com/chcg/notepad-plus-plus/commit/463bdc5e1416a5f2ee256d7316ed49e4865a0171#diff-5b536d07367d39b369aea9582ff0fe9d

with std::unique_lock<std::mutex> as replacement for CComCritSecLock<CComAutoCriticalSection>

d0vgan:
rebased + now everything in 1 commit

d0vgan:
Come on, why these ATL things are still here?
"Entities are not to be multiplied without necessity"!

d0vgan:
I used to think that it's me who is conservative and not fast - now I see I'm not alone :)

milipili:
@MAPJe71 From my point of view and from my experience, unlocking a mutex in the middle of the code where it is supposed to be managed by the scope is far from a good idea. Like deleting some resource managed by `unique_ptr`. For this reason, it is unlikely that an `unlock` mecanism comes in libyuni, better fix the code with proper raii (but feel free to bring some particular use-case).

Regarding libyuni/mutex, this is historical reason as `std::mutex` didn't exist before as you can guess. As of now, as the official minimum version for n++ is 2015 (if not mistaken, would be better to switch to 2017 from my point of view, there is a better support for multithread and brings multiple fixes), I would recommend to replace yuni/mutex by its stl counterpart as there is no added value.

That's said, @d0vgan is right using yuni/mutex here, for code consistency. That replacement has nothing to do with this MR and should be another one.

milipili:
@d0vgan I would encourage you next time to use multiple commits as it changes different things.

milipili:
@MAPJe71 Where did you find the snippet you posted regarding `CThreadSafeQueue`? I am surely missing something here as I can't find the `unlock()`

milipili:
@donho For information, libyuni is going full C++17, so `Yuni::Mutex`should disappear in the future (depending if users are still using it or not)

milipili:
@donho This class `CThreadSafeQueue` should die, or at least rewritten. No one must inherit from any stl container, as specified by the standard. Better use encapsulation instead of inheritance anyway.

milipili:
ATL dependency is really annoying. Nice job :)

d0vgan:
After some thinking, I think I vote for chcg's patch:
chcg/notepad-plus-plus@463bdc5#diff-5b536d07367d39b369aea9582ff0fe9d
It uses the standard STL's std::mutex - and it is better than relying on thirdparty's Yuni::MutexLocker, which additionally needs to be consireder to be updated.

MIvanchev:
@d0vgan: just to provide some aux. info, since I had the same fundamental problem as well. ATL/MFC is available with the [Visual C++ Build Tools 2015](https://blogs.msdn.microsoft.com/vcblog/2016/03/31/announcing-the-official-release-of-the-visual-c-build-tools-2015/). You can build Notepad++ easily that way.

chcg:
@MIvanchev The main point to get ride of ATL I think is to be able to build this under mingw.

MIvanchev:
> @MIvanchev The main point to get ride of ATL I think is to be able to build this under mingw.

Yes, I also think the mingw build should be restored. I wanted to raise awareness that the Build Tools include ATL/MFC, because I wasted 2 days trying to figure out things... =)

donho:
@d0vgan Sorry for the late review. Could you solve the conflict please?
Also please use lock_guard instead please?

donho:
@d0vgan Never mind, I will take in charge of it. Thank you for the PR.

Issue #4336, opened 2018-03-24
[Language] Turkish localization stays English after latest update


Tmp341:
### Description of the Issue

After the latest 7.5.6 update, Turkish language file isn't effective. Even though it is (Turkish.xml) locally presents. There was no problem before this.

### Steps to Reproduce the Issue

1. Installing latest update.
2. Changing to Turkish language.

### Expected Behavior

Language changed to Turkish.

### Actual Behavior

Language stays as English.

### Debug Information

Notepad++ v7.5.6   (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll nppcrypt.dll NppTextFX.dll PluginManager.dll Tidy2.dll 

![2](https://user-images.githubusercontent.com/1683388/37862187-d53932ba-2f59-11e8-8a74-95c1466641b8.jpg)

rddim:
The problem is that you have missing `>` at line 949 `</FindInFinder` must be `</FindInFinder>`
Also you have incorrect strings at line 1033 `<ÇevirOpenInAdminMode` must be `<OpenInAdminMode` and at line 1165 `<Çevirfinder-find-in-finder` must be `<finder-find-in-finder`

Tmp341:
How could i've missed those? 👍

Tmp341:
There is still an error or Notepad++ doesn't work with copy paste localizations. No change, even after locally error fixing.

SinghRajenM:
> No change, even after locally error fixing.

Follow below steps - 
1. Switch to English language (Settings->Preferences->General->Localization->English)
2. Update the Turkish localization in "%programfiles%\Notepad++\localization" or "%programfiles(x86)%\Notepad++\localization" based on your system architecture.
3. Now switch back to your localized language (similar to step 1).

It should work.

SinghRajenM:
Also there is one more issue around line 473.

```
<ShortcutMapperSubDialg title="Kısayol">
                <Item id="1" name=Tamam"/>
```

```name=Tamam"``` should contain quote after sign =.

Tmp341:
Because of those lines, my localization isn't working. I have to figure out my mistakes.

nesbilisim:
Solved,
Please add attached file in to localization.
[...]

PR #4338, opened 2018-03-24
Update Turkish translation

Tmp341:
- Trying to fix my errors made before #4336.

But tested directly, no luck. At least this fixes errors i've made.

### DIFF
--- a/PowerEditor/installer/nativeLang/turkish.xml
+++ b/PowerEditor/installer/nativeLang/turkish.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <NotepadPlus>
-	<Native-Langue name="Turkish" filename="turkish.xml" version="7.5.5">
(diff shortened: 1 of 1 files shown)


SinghRajenM:
Line [473](https://github.com/Tmp341/notepad-plus-plus/blob/016bb0adff49f82f6b7b2a93d1a597cef67e5b6d/PowerEditor/installer/nativeLang/turkish.xml#L473) also requires fix

```                <Item id="1" name=Tamam"/>```

**Comment by Tmp341** [comment]
Still not working. What is wrong with this?

**Comment by rddim** [comment]
Remove the space between `&amp;` and `Unix` at line 739 if you want to use <kbd>Alt</kbd>+<kbd>U</kbd> for enable/disable

**Comment by donho** [comment]
By using the following command, I've an error:
```
$ git checkout -b Tmp341-patch-1 master
fatal: A branch named 'Tmp341-patch-1' already exists.
```
You should create a new branch for a new PR instead of using the old branch.

PR #4364, opened 2018-04-03
Make addExt and deleteExts more robust

### LINKED ISSUE #4165: Extension registration doesn't work correctly
### Description of the Issue
When registering an extension, using npp file assoc builtin dialog, to be open with npp
it might happen that more extensions as expected get listed in builtin dialog.
The reason for this is, that the permission on the extension followed by the one selected
are breaking the code logic.
Possible solution:
https://notepad-plus-plus.org/community/topic/14931/does-not-save-the-files-i-need-to-use/17

see https://notepad-plus-plus.org/community/topic/14931/does-not-save-the-files-i-need-to-use/10

### Expected Behavior
only the chosen extension should be listed and in case of permission issue on this extension
a dialog explaining this would be nice

### Actual Behavior
all extensions are listed, even so they aren't registered

### Debug Information
all recent versions affected

ClaudiaFrank:
fixes #4165

### DIFF
--- a/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
+++ b/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
@@ -146,11 +146,12 @@ INT_PTR CALLBACK RegExtDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lPar
 					writeNppPath();
 
 					TCHAR ext2Add[extNameMax] = TEXT("");
+					bool addExt_successful = false;
 					if (!_isCustomize)
 					{
 						auto index2Add = ::SendDlgItemMessage(_hSelf, IDC_REGEXT_LANGEXT_LIST, LB_GETCURSEL, 0, 0);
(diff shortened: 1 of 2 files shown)


PowerEditor/src/MISC/RegExt/regExtDlg.cpp
```diff
@@ -330,47 +334,66 @@ void RegExtDlg::addExt(TCHAR *ext)
...
+	}
+	return true;
 }
 
 
 bool RegExtDlg::deleteExts(const TCHAR *ext2Delete)
 {
 	HKEY hKey;
-	::RegOpenKeyEx(HKEY_CLASSES_ROOT, ext2Delete, 0, KEY_ALL_ACCESS, &hKey);
+	long  nRet;
```
chcg: Could be just one line with definition and assignment.

ClaudiaFrank: Agreed, I just kept the style as used in addExt function. Should both be changed?

ClaudiaFrank:
not relevant

PR #4371, opened 2018-04-04
Added in the Character Panel two new columns showing HTML Codes

mmosoll:
- Removing some elements to avoid warnings and a compilation failure from previous pull request
- Changed the default title bar of the Character Panel
- Added in the Character Panel two new columns showing HTML Code and HTML Name for ANSI and Windows-1252 character sets.
- Modified NM_DBLCLK event to insert in document the current cell value when double clicked in the Character Panel.
- Modified translation files for French, English, Spanish and Spanish Argentina to include the new added columns in the Character Panel.
- Modified maximum length of the Character Panel titlebar in Notepad_plus::launchAnsiCharPanel().


PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1049,10 +1049,12 @@ Do you want to launch Notepad++ in Administrator mode?"/>
             <ColumnType name="Type"/>
         </WindowsDlg>
         <AsciiInsertion>
-            <PanelTitle name="ASCII Insertion Panel"/>
+            <PanelTitle name="ASCII/HTML Codes (ANSI or Windows-1252 character sets) Insertion Panel"/>
```
donho: Too long title

PR #4387, opened 2018-04-09
Fix issue 4192: Added close all as option in context

badearobert:


donho:
I don't think it's the good place to add this command.

Issue #4396, opened 2018-04-11
Unable to move untitled / dirty files from one n++ instance to another


badearobert:
### Description of the Issue
Unable to move untitled / dirty files from one n++ to another, as it requires closing the tab and reopening via a new n++ instance

### Steps to Reproduce the Issue

1.  write something on "new 1"
2.  create "new 2", write something else
3.  drag one of the files into a different window

### Expected Behavior
File to be moved accordingly

### Actual Behavior
Nothing happens, as it has early return on dirty files

donho:
Unfortunately, such limit cannot be removed.

PR #4398, opened 2018-04-11
Fixed issue #4396: Unable to move untitled / dirty files from one n++ instance to another

badearobert:
Fixed issue #4396

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3592,7 +3592,7 @@ void Notepad_plus::docOpenInNewInstance(FileTransferMode mode, int x, int y)
 {
 	BufferID bufferID = _pEditView->getCurrentBufferID();
 	Buffer * buf = MainFileManager->getBufferByID(bufferID);
-	bool isUntitledOrDirty = buf->isUntitled() || buf->isDirty();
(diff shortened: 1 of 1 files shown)


bahusoid:
But is it really moved as dirty/untitled? From code it looks like it simply opens "backup" file instead. Which is quite unexpected for "dirty file" case.

MAPJe71:
1. Please read the [guidelines for contributing](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests) and update your PR accordingly;
2. Did you test your changes?
3. If so, how did you test them (i.e. how can someone else verify them)?

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3592,7 +3592,7 @@ void Notepad_plus::docOpenInNewInstance(FileTransferMode mode, int x, int y)
 {
 	BufferID bufferID = _pEditView->getCurrentBufferID();
 	Buffer * buf = MainFileManager->getBufferByID(bufferID);
-	bool isUntitledOrDirty = buf->isUntitled() || buf->isDirty();
+	bool isUntitledOrDirty = (buf->isUntitled() || buf->isDirty());
```
donho: This line just does nothing.

donho:
The current PR might be a good initiative. However, it provides only a partial solution.

Issue #4393, opened 2018-04-10
menu name problem


nablas:
### Description of the Issue
A spell mistake in french version of last version of npp

### Steps to Reproduce the Issue

Install application in french, go to menu "paramétres" (french for parameters)

### Expected Behavior
should be paramètres, not paramétres

### Debug Information
When you think that the autor is french ...

hibyguy:
Grammar fixed!

donho:
9210bd37b6254988dce9fb1966e6d25a72332f85

hibyguy:
??

PR #4400, opened 2018-04-12
Fix properties menu french translation 

hibyguy:
Fixes issue #4393

"should be paramètres, not paramétres"


donho:
9210bd37b6254988dce9fb1966e6d25a72332f85

PR #4410, opened 2018-04-14
Remove "from" which is not a reserved JavaScript keyword

vlakoff:
References:
* http://www.javascripter.net/faq/reserved.htm
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords

All I found is the function [Array.from()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from), so it could be put in the "instre2" list, but this function is [pretty recent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility), and on the contrary the "instre2" list is quite outdated ("old school" javascript, e.g. lacks ES2015 functions).


vlakoff:
Bump, this one is an easy pick :)

PR #4411, opened 2018-04-14
Update marathi.xml

ganeshkoli:
Updated Marathi translation


donho:
@ganeshkoli  There are a lot of entries missing in it.
I will consider to integrate your commit if all the missing entries are added.

donho:
feel free to reopen it if my request is full filled.

PR #4419, opened 2018-04-17
More improvement towords C# intellisense

MCCshreyas:
1. Added `System.Array` class. 
2. Added `System.IO.File` class.
3. Added some missing methods for `System.Console` class. 
4. Added Popular `System.Windows.Forms.MessageBox` class. 
5. Added more methods of Thread class.


donho:
Make sure that you have commit all.
If a file needs to be updated 3 or 5 times in a short time, there's a problem!

MCCshreyas:
Done!

Issue #2433, opened 2016-10-17
Notepad++ exclude folder find in files filter


evandrocoan:
### Notepad++ exclude folder on the `Find in Files` filter

Originally posted on: http://stackoverflow.com/questions/4198829/notepad-find-in-files-filter-exclude

> I'm looking for a way to get n++'s find in files dialog to exclude certain files...
> 
> the only reference I've found so far is this http://sourceforge.net/project/shownotes.php?release_id=536795&group_id=189927 , but, unfortunately, it doesn't work. 
> 
> I'm using *[^*.dll] to, I think, search for everything BUT dll files.  
> 
> How can I resolve this ?

Is there how to exclude folders from the find in files dialog? I need to exclude the `.git` folders from the search, as the are very big and there are several of them.

![image](https://cloud.githubusercontent.com/assets/5332158/19440114/659f0d58-945f-11e6-9f2d-070851500037.png)

dail8859:
You can't. It doesn't work of real regular expressions. It uses the windows API and only supports wildcards (`*`) and I think question marks.

To search in multiple file types use something like `*.cpp *.h *.xml`

nikoudel:
Regex in Filters could become one of the best features of Notepad++ :(

robsiera:
At least every two years I find myself investigating if NotePad++ already has the long wanted "Exclude files filter".  I'm not complaining though. Keep up the good work!

Spin45:
Please add the ability to exclude folders/files in searches.

Nightblade:
Just checking, this isn't fixed yet is it?

PhilipOakley:
The solution from https://notepad-plus-plus.org/community/topic/10653/exclude-directories-from-the-search/5 is to temporarily set those folders you want to exclude to be 'hidden' folders (hiding just the top level sub-dir is sufficient), and this then stops Notepad++ descending into that sub-directory.

Worked for me just now!

Nightblade:
@PhilipOakley Thanks, but technically that's a workaround isn't it?  ;)

sasumner:
@Nightblade Sometimes...and sometimes sadly...workarounds are all we get. :)

PhilipOakley:
@Nightblade Correct, it's just a workaround, until such time as it is included in the (searchable) documentation :wink:

sudhir600:
Here is a easy solution -
[...]

PR #4429, opened 2018-04-19
#2433 Initial implementation of folder exclusion find in files.

jetand:
#2433 Initial implementation of folder exclusion find in files.


Lonzak:
Why hasn't that pull request been accepted? Lots of people waiting on that...

donho:
Is that some kind of joke?

image

Lonzak:
@jetand Can you fix that ui issue? I currently doesn't have my c++ env set up to have a look... But why that took half a year is beyond me...

dazey3:
@donho Rather rude, don't you think? Not at all constructive.

cfoxy17:
Notepad++ needs to be changed in one of two ways:

1. The UI issue with this "fix" needs to be resolved

OR

2. filters should allow you to enter multiple filters separated by a semicolon.. we should be able to exclude searching certain directories. This isn't a helpful feature, it's necessary. Perhaps it could be implemented by using a "!" before directories we want to omit.

cfoxy17:
You can exclude certain folders from the search by making the folders hidden that you want to omit. 

Right click the folder -> properties -> check the "Hidden" box -> click apply

jgreywolf:
@cfoxy17 This only works if the files you want to exclude are in specific folders.  If you want to exclude certain filetypes no matter where they are located within the folder structure you need to be able to exclude those files in a different way.

Being able to prepend ! before an extension seems the most logical way to handle this to me...

xylographe:
Implemented in commit dc48cee5f7f4efe1f70e7269fe5e46ac8b861b1b.

Lonzak:
@xylographe Really? Thank you!

sasumner:
@Lonzak : Well...I haven't looked at it super-closely, but I believe the commit mentioned by @xylographe is for file pattern matching based exclusion, and this issue _seems_ to be (from its title) for excluding folders, what you are interested in may truly _not_ be implemented by the commit.

Lonzak:
I had a look at the code and my c++ times have been awhile however it looks to me that its for files*:

```
bool matchInList(const TCHAR ***fileName**, const std::vector<generic_string> & patterns)
...
		if (patterns[i].length() > 1 && patterns[i][0] == '!')
		{
			if (PathMatchSpec(**fileName**, patterns[i].c_str() + 1))
				return false;

			continue;
		} 
```

except fileName means folders but than it is bad naming of variables imho

xylographe:
Yes, my comment was in response to [the previous comment by @jgreywolf](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4429#issuecomment-557190010).
Commit dc48cee5f7f4efe1f70e7269fe5e46ac8b861b1b implements glob pattern based _file_ exclusion.

Excluding _folders_ is still only possible with [the workaround posted by @cfoxy17](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4429#issuecomment-495633652).

Lonzak:
Perfect! So my last question in which version will this commit go? 7.8.2? 7.9?

xylographe:
@Lonzak If you do not want to wait for the next release (probably 7.8.2), you can download the updated Notepad++ executable from [the AppVeyor Artifacts](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/29050635). Do _not_ replace `SciLexer.dll`, only `notepad++.exe`.

Lonzak:
7.8.2 is out and its working! Finally :-)

dazey3:
Been waiting a long time to exclude file types! Great work @xylographe! 👍

donatelloOo:
What about the initial feature request which was "**folder** exclusion find in files" ?
Is it planed to have a way to exclude directories based on file path ?

For instance, never look into `target/` sub-folders using `!target/**`, or `!**/target/**`, or something like that ?

PR #1618, opened 2016-03-11
Add autocompletion file for AutoIt

vlakoff:
Add autocompletion file for the amazing [AutoIt](https://www.autoitscript.com/site/) software, which is already supported in syntax highlighting.

File directly taken from directory `Extras\Editors\Notepad++` of their release package. I only replaced the single quotes by double quotes for consistency with the other autocompletion files, and made sure this change didn't break anything.

I searched the author of this substantial file for proper credit, but I couldn't find it.

PR #4438, opened 2018-04-21
Add autocompletion file for AutoIt

vlakoff:
This is a resubmit of #1618.

* Updated `autoit.xml` file, taken from [version 3.3.14.5](https://www.autoitscript.com/site/autoit/downloads/) (for easier comparison and update, I keep the file as it is. No single to double quotes conversion as in #1618)
* Adapted to x64 installer refactoring https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fc014595a3e6dc31bc8e2fed4e4ef5cc3b183cf5


donho:
The xml file is 1 MB ! 
For me it's too much for a file for auto-completion.

vlakoff:
As I wrote in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/1618#issuecomment-243354057, 1 MB is nothing nowadays, and compressed into the installer it only adds 84 KB.

But I understand your position, and I can share it (I was a bit uncertain too about adding this file).

Issue #3805, opened 2017-10-05
Authenticode check failed unable to load Scintilla Library


dayton967:
### Description of the Issue
Upon Starting Notepad++

### Steps to Reproduce the Issue

1. Downloaded the ZIP for notepad++ 7.5.1
2. Ran Notepad++

### Expected Behavior
Should start notepad++

### Actual Behavior
Pop up for "Library Verification Failure" with the message "Authenticode Check Failure: signature or signing certificate are not recognized"
Followed by a Pop up  with "Exception On WM_CREATE" with the message "ScintillaEditView::init : SCINTILLA ERROR - Can not load the dynamic library"

### Debug Information
Notepad++ doesn't run.

![popup1](https://user-images.githubusercontent.com/4029174/31235186-cb0e2ffa-a9be-11e7-8fc3-493c4c7f9200.jpg)
![popup2](https://user-images.githubusercontent.com/4029174/31235187-ccaf41f0-a9be-11e7-9fe6-4033a2e6388f.jpg)

LuciferSam86:
seems this reason: https://notepad-plus-plus.org/news/notepad-7.3.3-fix-cia-hacking-issue.html

are time and date correct on the system?
Did you check the hashes for the program here: https://github.com/notepad-plus-plus/notepad-plus-plus/releases/tag/v7.5.1
Did you try do download again NPP?

dayton967:
The time is accurate. and I have tried to install a few times.
ff4b6e4fef1c1f8dd1f43337d26b9c711175ae58 *npp.7.5.1.bin.minimalist.7z

> signtool verify /v /pa "notepad++.exe"
Verifying: notepad++.exe

Signature Index: 0 (Primary Signature)
Hash of file (sha1): 3882FA198F4E83CF98156B94B6E8922BD377B0F2

Signing Certificate Chain:
    Issued to: DigiCert High Assurance EV Root CA
    Issued by: DigiCert High Assurance EV Root CA
    Expires:   Sun Nov 09 20:00:00 2031
    SHA1 hash: 5FB7EE0633E259DBAD0C4C9AE6D38F1A61C7DC25

        Issued to: DigiCert SHA2 High Assurance Code Signing CA
        Issued by: DigiCert High Assurance EV Root CA
        Expires:   Sun Oct 22 08:00:00 2028
        SHA1 hash: F7E0F449F1A2594F88856C0758F8E6F627E5F5A2


            Issued to: Notepad++
[...]

Issue #3504, opened 2017-07-09
NPP 7.4.2 Error after installation


aurum-aeolus:
### Description of the Issue

After install the program start with error in Scintilla component.

Library verification failed.
Authenticode  check failed: signature or signing certificate are not recognized.
Exception On WM_CREATE ScintillaEditView::init : SCINTILLA ERROR - Can not load the dynamic library

Win32Exception
An exception occurred. Notepad++ cannot recover and must be shut down.
The exception details are as follows:
Code: 0xC0000005
Type: Access violation
Exception address: 0x0038E4A8
[NppDump.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1133578/NppDump.zip)

### Steps to Reproduce the Issue

1. Install NPP 7.4.2 with default options
2. 
3. 

### Expected Behavior

Run the program normally.
### Actual Behavior

The program crash with error in Scintilla component.
### Debug Information

Debug from NPP 7.4.1

Notepad++ v7.4.1   (32-bit)
Build time : May 18 2017 - 02:00:37
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (32-bit)
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll 

The NPP 7.4.2 create dmp file.

IvanKostyuk:
How to resolve this issue? I did a workaround to comment `VerifySignedLibrary(sciLexerPath, SCINTILLA_SIGNER_KEY_ID, SCINTILLA_SIGNER_SUBJECT, SCINTILLA_SIGNER_DISPLAY_NAME, false, false);
` out in **ScintillaEditView.cpp**.
If this is a solution, may I provide a pull request?

Lujaina95:
1. if you tried to debug the project and it shows to you the following message: 
   " ScintillaEditView::init : SCINTILLA ERROR can not load the dynamic library "
   this message mean that you do not have scilexer.dll after you build scilexer.dll with boost 
   so,what you have to do? 
   -you need to back to VS2017 and click on project properties to tell the linker where to look for libraries to link to.
   To access this setting:
    - right-click on the project name in the Solution Explorer window,
    - then Properties ? Linker ? General ? Additional Library Directories.
    - Enter <boost_path>\stage\lib
[...]

PR #4456, opened 2018-04-25
fix 2 bugs #3805 , #3504

Lujaina95:
fixing the following bugs : 
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3805
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3504

which they are talking about having issue when try to run notepad++ in visual studio
and the message that appear to them is the following: 
"ScintillaEditView::init : SCINTILLA ERROR can not load the dynamic library "

i faced the same problem before, so these the steps that helped me to fix the problem

you need to back to VS2017 and click on project properties to tell the linker where to look for libraries to link to.
To access this setting:
right-click on the project name in the Solution Explorer window,
then Properties ? Linker ? General ? Additional Library Directories.
Enter <boost_path>\stage\lib
you need to change x64 to X86 which is in Visual studio because Notepad++ v7.5.4 (32-bit) it just work with x86
Also, you need to download SDK which you can find it at the following link:
https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk

these steps will be helpful for who use:
Notepad++ v7.5.4 (32-bit)
operating system x64
Windows 10
VS2017


MAPJe71:
Please close this pull request and stop creating PRs like this, it's not helpfull.

PR #4513, opened 2018-05-21
Cleanup in context menu and Run entries

vlakoff:
* Context menu: remove entry "Google Search"

  * redundant with the "Search On Internet" entry
  * … and just noneffective as the corresponding Run entry has been removed
  * introduced in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e451efbd290f36300fe08a3b5be5460c5dd94f84, see also my past comment on it: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e451efbd290f36300fe08a3b5be5460c5dd94f84#r22615404

* Run commands: remove entry "Launch in Safari"

  * Safari for Window is just [discontinued](https://apple.stackexchange.com/questions/68836/where-can-i-download-safari-for-windows/94010#94010)
  * … and nobody was using it

* Fix case: "Get php help" → "Get PHP help"


vlakoff:
Another nitpick: fix case "Search On Internet" → "Search on Internet"

(no change in practice, because the corresponding menu item was already with lowercase "on")

donho:
@vlakoff 
Even the dev of Safari under windows is discontinued, it's not the reason to remove this shortcut.
> and nobody was using it

what's the tangible proof to make you say that?

vlakoff:
I do maintain this menu entry is not worth it.
For me this is pretty clear, and I don't plan to argue about it.

"Less is more" principle, keeping such useless entry make the useful entries harder to spot.

But it's up to you :)

vlakoff:
If today, in a parallel universe, this entry was not present and there were a PR to add such an entry, would you accept it?

The Safari entry has been added along the Chrome one, [in 2009](https://github.com/vlakoff/notepad-plus-plus/commit/f5cca1add9da6d913411626facf4a82b71865a8a) (at that time, Chrome was [pretty new](https://en.wikipedia.org/wiki/Google_Chrome_version_history), far from the popularity it has now).

chcg:
See https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility
and (also in german) http://www.chip.de/downloads/Safari_27264736.html.

Latest safari version for windows is 5.1.7 from 2012, also for Mac it is 11.1. So the advice is not to install it any more due to security risks.

PR #4518, opened 2018-05-23
make carret line always visible in the find panel

### LINKED ISSUE #4510: In the search results area, I double-click on the search result and it will jump to the text area. However, I don’t know which row was used in the search results area.
In the search results area, I double-click on the search result and it will jump to the text area. However, I don’t know which row was used in the search results area.

### Description of the Issue

The first step below
ntp1

The second step below
ntp2

1. In the search results area
2.  I double-click on the search result and it will jump to the text area
3.  However, I don’t know which row was used in the search results area.

### Expected Behavior

After the second step, the search result area still shows which row is selected
### Actual Behavior

After the second step, I do not know which line is selected in the search result area
### Debug Information

nothing

nothing

cmeriaux:
fixes #4510


d0vgan:
+1

d0vgan:
Motivation:
https://notepad-plus-plus.org/community/topic/15835/find-result-window-s-current-active-line-colour

Issue #4529, opened 2018-05-26
On Debug Build, Korean Input has Assert. 


nhy3795:
### Description of the Issue

Assert is invoked because of the isdigit function called by the is isdgits function when entering Korean in the debug build.

### Steps to Reproduce the Issue

1.  Runs on a debug build.
2.  Input Korean. 
3.  Assert is executed.

### Expected Behavior

Debug environment without Assert.
### Actual Behavior

Debug environment with Assert.
### Debug Information

Input 'ㅁ', Not Copy and paste on clipboard. Input on keyborad. 

Call Stack
 	notepad++.exe!_VCrtDbgReportW(int nRptType, void * returnAddress, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist) line 502	C++
 	notepad++.exe!_CrtDbgReportW(int report_type, const wchar_t * file_name, int line_number, const wchar_t * module_name, const wchar_t * format, ...) line 278	C++
 	notepad++.exe!_chvalidator_l(__crt_locale_pointers * locale, int c, int mask) line 42	C++
>	notepad++.exe!_chvalidchk_l(const int _C, const int _Mask, __crt_locale_pointers * const _Locale) line 161	C++
 	notepad++.exe!_ischartype_l(const int _C, const int _Mask, __crt_locale_pointers * const _Locale) line 186	C++
 	notepad++.exe!_isdigit_l(int c, __crt_locale_pointers * locale) line 74	C++
 	notepad++.exe!isdigit(int c) line 79	C++
 	[External Code]	
 	notepad++.exe!isAllDigits(const std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> > & str) line 46	C++
 	notepad++.exe!AutoCompletion::getWordArray(std::vector<std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >,std::allocator<std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> > > > & wordArray, wchar_t * beginChars) line 133	C++
 	notepad++.exe!AutoCompletion::showApiAndWordComplete() line 94	C++
 	notepad++.exe!AutoCompletion::update(int character) line 750	C++
 	notepad++.exe!Notepad_plus::notify(SCNotification * notification) line 642	C++
[...]

PR #4530, opened 2018-05-26
Fix Unicode Problem.  Issue Name 'On Debug Build, Korean Input has As…

ffdd270:
When compiled in a Unicode environment, the isAllDigits function uses the iswdigit function.
Other environments use the isdigit function.

Fix Issue #4529.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -43,7 +43,7 @@ static bool isInList(generic_string word, const vector<generic_string> & wordArr
 
 static bool isAllDigits(const generic_string &str)
 {
-	return std::all_of(str.begin(), str.end(), ::iswdigit);
+	return std::all_of(str.begin(), str.end(), ::_istdigit);
 }
 
 
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -43,11 +43,7 @@ static bool isInList(generic_string word, const vector<generic_string> & wordArr
 
 static bool isAllDigits(const generic_string &str)
 {
-#ifdef _UNICODE
 	return std::all_of(str.begin(), str.end(), ::iswdigit);
-#else
-	return std::all_of(str.begin(), str.end(), ::isdigit);
-#endif
 }
 
 
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -43,7 +43,11 @@ static bool isInList(generic_string word, const vector<generic_string> & wordArr
 
 static bool isAllDigits(const generic_string &str)
 {
+#ifdef _UNICODE
+	return std::all_of(str.begin(), str.end(), ::iswdigit);
+#else
 	return std::all_of(str.begin(), str.end(), ::isdigit);
+#endif
 }
 
 


PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -43,7 +43,11 @@ static bool isInList(generic_string word, const vector<generic_string> & wordArr
 
 static bool isAllDigits(const generic_string &str)
 {
+#ifdef _UNICODE
```
SinghRajenM: Notepad++ is Unicode. ANSI version was dropped long back. Moreover, ```generic_string``` is ```typedef std::basic_string<TCHAR> generic_string;``` where ```TCHAR``` is ```WCHAR``` because of Unicode project setting. Also refer discussion - https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3198#discussion_r113758228

nhy3795: Thank you for reviewing. I will remove that compiler encounter.

nhy3795: Then, Can I erase the compiler encounters without considering ASCII situations? And was there an issue in the context of using iswdigit?

PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
```diff
@@ -43,11 +43,7 @@ static bool isInList(generic_string word, const vector<generic_string> & wordArr
 
 static bool isAllDigits(const generic_string &str)
 {
-#ifdef _UNICODE
 	return std::all_of(str.begin(), str.end(), ::iswdigit);
```
CookiePLMonster: `_istdigit` if you want to care about both. Since `generic_string` is typedef'd, I'd advice using a TCHAR version.

nhy3795: Thank you for Code Review. I will fix that on next commit.

but I have one question, generic_string is templated string? For Example. std::wstring same as generic_string<WCHAR>?

CookiePLMonster: Yep. From `Common.h`:
```cpp
typedef std::basic_string<TCHAR> generic_string;
typedef std::basic_stringstream<TCHAR> generic_stringstream;
```

nhy3795: Thank You For Answer. I fix that.

chcg:
There a some further code areas which use:

`::isdigit`

instead of 

`::_istdigit`

on generic_string, e.g.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.h#L54
or TCHAR:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ef13902206f80e00dcfd7d85b342066ed6f86d66/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp#L89

CookiePLMonster:
> and maybe also the scintilla component need to be checked.

Perhaps - but what's the convention on making changes inside scintilla? It's not exactly clear right now.

zufuliu:
Scintilla (except lexers, which mostly maintained by external developers) doesn't directly call ctype functions.

https://sourceforge.net/p/scintilla/code/ci/348e55f8107caddf4f48b8c514caa90f1c886c48/

PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
AjayVV: > Notepad++ is Unicode. ANSI version was dropped long back. Moreover, `generic_string` is `typedef std::basic_string<TCHAR> generic_string;` where `TCHAR` is `WCHAR` because of Unicode project setting. Also refer discussion - [#3198 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3198#discussion_r113758228)

Then why not replace all references of **T** in the entire project with **W**. Add some ``static_assert`` or `#define` all T-names to something that would cause a compiler error, and preventing T-uses.

LordNyriox:
@ffdd270:  The Scintilla version in Notepad++ has gone through a major upgrade.  Notepad++ now uses Scintilla `4.1.4` (instead of `3.5.6`).  

Please rebase your PR to be compatible with the current code.  

While you are doing so, you might also wish to look into @chcg's suggestion [<https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4530#issuecomment-434820177>].  ;}

Issue #4531, opened 2018-05-26
Find dialog: Add short key for button 'Find All in Current Document'


jannick0:
English version of npp 7.5.6: In the `Find` dialog there is no short key (displayed by underscoring the character in the string of the button `Find All in Current Document`) to be invoked together with ALT instead of clicking on the button.

Please add a short key such that, e.g. `ALT + d` invokes the search in the current document. I believe there is a short key in some locale of npp (other than English).

I hope I am not missing something trivial here to do it myself in some config.

Many thanks.

sasumner:
You ARE missing something.  :)

Here's what I do:  edit .\\localization\\english_customizable.xml and put `&amp;` just before the desired character in the "Find All In Current Document" string (or change the string to something else entirely if you want--but the important part is to put the `&amp;` in.  Then, in Preferences -> General -> Localization, set it to English (customizable).

jannick0:
@sasumner Many thanks for this.

Probably beneficial to add this to the standard English configuration like for other languages (see PR #4533).

sasumner:
So I really don't know the history on this, but I presume at one time some but not all keys were given an "alt" shortcut in the Find family of tabs.  Then later when the ability to change the text and/or add "alt" key combinations at the "user" level came into existence (okay, you have to edit a file...), having the developers pick and maintain all of these (across many different languages) basically dropped to zero importance level.

jannick0:
Understand. Let's see if the PR comes thru.

donho:
Will be in the next release

PR #4533, opened 2018-05-27
english.xml: add short key in find dialog

jannick0:
added short key for button `Find All in Current Document`  in `find` dialog (which should not infer with another configuration of short cut).

Cf. Issue #4531.


donho:
Wrong file to modify.
Will be in next release.

PR #4534, opened 2018-05-27
Add IAR Embedded Workbench file extensions to XML

pawelszramowski:
The following IAR Embedded Workbench files happen to be manual edited every so often:
- ewd: Project settings for C-SPY
- ewp: IAR Embedded Workbench project
- ewt: Project settings for C-STAT and C-RUN IDE IDE
- eww: Workspace file


donho:
Use "User ext." in Style Configurator to define your custom extensions.

pawelszramowski:
This is quite popular IDE for embedded software development and has been on the market for a long time, why should each user have to define these extensions on their own? There's already a bunch of application-specific file extensions on the list. Would you suggest Mozilla developers to add xul to user extensions, too?

Issue #3969, opened 2017-12-07
Changing Timezones is generating a popup that the file has been modified.


D3niz:
### Description of the Issue
With Notepad++ open and any file open, changing the computers timezone then gives a warning in Notepad++ with a popup "This file has been modified by another program, Do you want to reload it and lose the changes made in Notepad++?" or "This file has been modified by another program, Do you want to reload it"

This becomes annoying since it hasn't been changed and the time hasn't changed (still the same time UTC) only the time zone has changed.

### Steps to Reproduce the Issue
1.  Change your computer Timezone, e.g. set to +10
2.  Open notepad++ and any text file (the file needs to be saved, not just new file within notepad++)
3.  Change the Timezone to UTC
4.  go back to notepad++
4.  this pop up comes up "This file has been modified by another program, Do you want to reload it and lose the changes made in Notepad++?" or "This file has been modified by another program, Do you want to reload it"

### Expected Behavior
Don't display pop up if it is just a timezone change and the file hasn't been modified.
Haven't looked at code but I guess it should be always looking at the UTC time so it is consistent and timezone changes become irrelevant.

### Debug Information
Notepad++ v7.5.1   (64-bit)
Build time : Aug 29 2017 - 02:38:44
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll

dail8859:
Duplicate of #2535, #215

Issue #2535, opened 2016-11-06
Change timezone - Notepad acting crazy


kloentje2:
When I change my timezone, Notepad++ asks me on every file to reload it because another program has modified it.

Ik except notepad to continue with his current work, because I am working on a big project I have a lot of code open.

### Debug Information
Notepad++ v6.9.1
Build time : Mar 28 2016 - 19:48:40
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll

dail8859:
Duplicate of #215

fplatel:
I have the same issue : I travel a lot for my job, and I don't often shutdown my laptop : I only use hibernation. Each time I change my timezone, notepad++ ask me to reload each file, I suppose it sees that the local time for last modified is different, it should perform checking against universal time, not local time.

Issue #215, opened 2015-06-14
changing timezone causes all files to appear changed


ennorehling:
changing the windows timezone brings up the "this file has been changed by an external program" dialog for every open file. it looks like the timestamp comparison is not taking timezones into account?

donho:
Have you the instructions to reproduce this bug?

milipili:
Indeed the timezone is not taken into consideration. To avoid that n++ should work with UTC timestamps.
#792 show the way to fix the problem, even if I don't think that calling `_tzset` each time is the best way to fix that.

fplatel:
I have the same issue : I travel a lot for my job, and I don't often shutdown my laptop : I only use hibernation. Each time I change my timezone, notepad++ ask me to reload each file, I suppose it sees that the local time for last modified is different, it should perform checking against universal time, not local time.

mzso:
I think it shouldn't be based on time at all, but actual changes. Time is
unreliable, and accurate time is not always available.

2017. márc. 5. 22:49 ezt írta ("fplatel" <notifications@github.com>):

I have the same issue : I travel a lot for my job, and I don't often
shutdown my laptop : I only use hibernation. Each time I change my
timezone, notepad++ ask me to reload each file, I suppose it sees that the
local time for last modified is different, it should perform checking
against universal time, not local time.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/notepad-plus-plus/notepad-plus-plus/issues/215#issuecomment-284265796>,
or mute the thread

milipili:
I don't think you want that n++ spends its time performing checksums, especially when the opened files are big. However, n++ should definitively take the timezone into consideration when comparing timestamps. It does not matter if the timestamp is correct or not, the only thing that matters is that it has changed or not.

ColColonCleaner:
Just hit this today as I changed timezones. If this could be looked at I would be grateful.

HorizonNL:
I produced the issue and i want to fix it, what is the file name that i should work on it ?

HorizonNL:
This is my fix i hope it works with you.

`#pragma once
#include <vector>
#include <string>
#include <sstream>
#include <windows.h>
#include <iso646.h>
#include <cstdint>
#include <time.h>

const bool dirUp = true;
const bool dirDown = false;

#define NPP_CP_WIN_1252           1252
#define NPP_CP_DOS_437            437
[...]

PR #4541, opened 2018-05-31
Operate on UTC timestamps internally + deprecate stat/wstat

### LINKED ISSUE #4491: When time zone is changed in Windows 10, Notepad++ says that all open files have been modified
### Description of the Issue

When time zone is changed in Windows 10, Notepad++ says that all open files have been modified

1.  Open files
2.  Change timezone

### Expected Behavior

NotePad++ should NOT ask if the file shuold be reloaded.

### Actual Behavior

Notepad++ offers reloading all open files.

CookiePLMonster:
This PR is split in two parts to ease review. It should also be noted that the first commit is strictly a bug fix (albeit it forced quite a bit of refactoring), while the other commit unifies the rest of the code to be in line with these changes and is purely a code quality improvement.

1. First commit migrates stat-based timestamps to FILETIME based timestamps obtained from WinAPI instead of a CRT stat function. As per MSDN, timestamps obtained from GetFileAttributesEx (and other functions from that family) retrieve time in UTC and not system time, so that solves all issues where switching timezones triggers a "file modified" prompt.
2. Second commit unifies the code further by removing all instances of stat in favour of WinAPI based functions. This should improve code readability and also removes the need for `custom_wstat` hack added recently to circumvent an API bug on Windows XP. Since time/date formatting is also changed, "Summary" tab now shows date/time in locale-based format instead of a hardcoded format we used to have.

Fixes #4491, #3969, #2535 and #215.


PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -313,44 +313,72 @@ void Buffer::reload()
...
+			case ft_modified:
+				rawtime = attributes.ftLastWriteTime;
+				break;
+			default:
+				rawtime = attributes.ftLastAccessTime;
+				break;
+			}
+
+			SYSTEMTIME utcSystemTime, localSystemTime;
+			FileTimeToSystemTime(&rawtime, &utcSystemTime);
```
donho: Why this conversion? Does FILETIME is already UTC?
`Contains a 64-bit value representing the number of 100-nanosecond intervals since January 1, 1601 (UTC).`
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx

CookiePLMonster: Yes, but `getFileTime` returns a string for file statistics, so it should be presented to the user in his local time, not UTC. Therefore, `SystemTimeToTzSpecificLocalTime` needs to be used, and since it operates on `SYSTEMTIME`, first this file time needs to be converted to system time.

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -313,44 +313,72 @@ void Buffer::reload()
...
+			}
 
-			generic_strftime(tmpbuf, temBufLen, TEXT("%Y-%m-%d %H:%M:%S"), timeinfo);
-			return tmpbuf;
+			result += ' ';
+
+			cbSize = GetTimeFormat(LOCALE_USER_DEFAULT, 0, &localSystemTime, nullptr, nullptr, 0);
+			if (cbSize != 0)
+			{
+				TCHAR* buf = new TCHAR[cbSize];
```
donho: Call me a psychopath if you want to, I just hate dynamical allocation, and avoid it as I can.
So please make buf a static char buffer in the stack, with 256 byte as size which will be enough.

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -313,44 +313,72 @@ void Buffer::reload()
...
+			}
+
+			SYSTEMTIME utcSystemTime, localSystemTime;
+			FileTimeToSystemTime(&rawtime, &utcSystemTime);
+			SystemTimeToTzSpecificLocalTime(nullptr, &utcSystemTime, &localSystemTime);
+
+			int cbSize = GetDateFormat(LOCALE_USER_DEFAULT, 0, &localSystemTime, nullptr, nullptr, 0);
+			if (cbSize != 0)
+			{
+				TCHAR* buf = new TCHAR[cbSize];
```
donho: No memory allocation in the heap.

donho:
I really don't like dynamical allocation. The allocation costs, even in our miniscule case, not mention to  the delete operation - why bother if you can use static automatic variable?

CookiePLMonster:
I'll flip your question - why bother if a static allocation is not 100% safe? A constant 256 size will indeed be good for most cases, but can you guarantee it will **always** be enough? And if not, why should this code fail just because of an arbitrary limit, which can be circumvented by allocating dynamically?

Either way, I still propose `malloca`. In cases you want to cover (as in, up to 256 characters) it will **always** use the stack, so maybe even it will never ever use the heap - but if it ever has to, it will be handled correctly.

Issue #4579, opened 2018-06-13
Document highlight is incorrect when scrolling past end of file.


Kered13:
### Steps to Reproduce the Issue
1. Disable word wrapping.
2. Enable scrolling beyond the last line.
3. Open any document.
4. Scroll down beyond the end of the line.

### Expected Behavior
Highlighted area of document map should correctly reflect what is visible on screen.

### Actual Behavior
Highlighted area is incorrect, see screenshot.

![Screenshot](https://user-images.githubusercontent.com/21075688/41331321-817ff96a-6ea6-11e8-87cb-78d6f8301ba5.jpg)

Kered13:
I have prepared a fix for this in [pull request #4580](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4580).

PR #4580, opened 2018-06-13
Fix incorrect document map highlight when scrolling past end of file.

Kered13:
Fix for [Issue #4579](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4579).

Screenshot

### DIFF
--- a/PowerEditor/src/WinControls/DocumentMap/documentMap.cpp
+++ b/PowerEditor/src/WinControls/DocumentMap/documentMap.cpp
@@ -248,7 +248,7 @@ void DocumentMap::scrollMap()
 			if (lowerY == 0)
 			{
 				auto lineHeight = _pMapView->execute(SCI_TEXTHEIGHT, firstVisibleDocLine);
-				lowerY = nbLine * lineHeight + firstVisibleDocLine;
+				lowerY = nbLine * lineHeight + higherY;
 			}
 		}
 		else


Kered13:
This failure was caused by [this commit](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c94319e863e0f27e75a6e7999dd1fff9acac3d51), not my change.

donho:
It seems on the normal edit zone the wrap option is off whereas this option is ON on doc map.
I cannot reproduce it at all though.

Could you give me more detail for reproducing this bug please?

Kered13:
@donho I'm not sure what you mean about the wrapping. In the screenshot wrap is off in both the main edit region and the document map.

Messing around with a clean install it seems that the bug does not always manifest. If you toggle word wrap on then back on then the document map hightlight is correct, although it stops at the end of text. It may or may not be desirable to highlight past the end of text. It also doesn't always repro on short documents, though the highlight does go past the end of text in this case. However when I start Notepad++ with word wrap off and view a long document, like the Notepad++ LICENSE, then I consistently produce the bug.

Repro steps:
1. Disable word wrap. (View > Word Wrap)
2. Enable scrolling beyond last line (Tools > Preferences > Editing)
2. Restart Notepad++
3. Open Notepad++/LICENSE

PR #4583, opened 2018-06-13
Added 2 pdf files detailing steps for building Notepad++ with…

bfridkis:
… screenshots. These are 2 alternate means for doing so, and are to address difficulties expressed in issue reports 1228, 4474, & 4348.


donho:
These 2 pdf could be nice. But in my opinion they are rather a website as content than being included in the repository.

PR #4605, opened 2018-06-25
Add w/ BOM and remove w/o BOM in Russian locale

SerVB:
There is a confusion of `UTF-8-BOM` and simple `UTF-8` in the Russian localisation.

This is because:
* `UTF-8` was named as `UTF-8 without BOM` (in Russian: `UTF-8 без BOM`).
* `UTF-8-BOM` was named as `UTF-8`.

So many users (that don't know what BOM is) think that they should select `UTF-8` to work with simple `UTF-8` but they actually work with `UTF-8-BOM`. It has been happening for many years! Actually, I also stepped on these rakes when I started to learn programming many years ago.

I propose to change the localisation file to the [English file](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f4460075e6fe6400a7d595fa82bab54693d01929/PowerEditor/installer/nativeLang/english.xml#L263) style. Now:
* `UTF-8` is named as `UTF-8`.
* `UTF-8-BOM` is named as `UTF-8 with BOM` (in Russian: `UTF-8 с BOM`).

Also I added `with BOM` (in Russian: `с BOM`) to `UCS-2` encodings (again as in the English file).

I edited only the `<all menu item>` block because the `<NewDoc>` block had been already translated properly: it has the same style (check [this block](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f4460075e6fe6400a7d595fa82bab54693d01929/PowerEditor/installer/nativeLang/russian.xml#L687)).

### DIFF
--- a/PowerEditor/installer/nativeLang/russian.xml
+++ b/PowerEditor/installer/nativeLang/russian.xml
@@ -693,8 +693,8 @@
 					<Item id="6406" name="ANSI"/>
 					<Item id="6407" name="UTF-8"/>
 					<Item id="6408" name="UTF-8 c BOM"/>
-					<Item id="6409" name="UCS2 Big Endian c BOM"/>
-					<Item id="6410" name="UCS2 Small Endian c BOM"/>
+					<Item id="6409" name="UCS-2 Big Endian c BOM"/>
+					<Item id="6410" name="UCS-2 Small Endian c BOM"/>
 					<Item id="6411" name="Синтаксис по умолч.:"/>
 					<Item id="6419" name="Новый Документ"/>
 					<Item id="6420" name="Применить к откр. ANSI файлам"/>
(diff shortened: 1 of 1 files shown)


SerVB:
Hmm... It's not passing. Wasn't me!

donho:
@GOLEM777 Do you approve this PR?

GOLEM777:
@GOLEM777 Do you approve this PR?No problem! 23.07.2018, 02:39, "Don HO" <notifications@github.com>:@GOLEM777 Do you approve this PR?

GOLEM777:
I made a separate fix. SerVB thx.

SerVB:
@GOLEM777 , hmm, I just did the same ;)

SerVB:
@donho so?

PR #4609, opened 2018-06-28
Find in Project Panel files

Uhf7:
Changes in Project Panel
- Caption displays "Project 1", "Project 2", "Project 3" instead of "Project", "Project", "Project"
- Root node displays workspace file name instead of "Workspace".
- Root node has new context menu command: Find in Files.
- If more than one Project Panel is open, then the tabs display "1", "2", "3" instead of "Project", "Project", "Project".
- Tab sizing improved.

Changes in Find in Files Dialog
- Added Checkbox for each Project Panel.
- If Project Panel checkbox(es) are set, then "Find in Files" will search Project Panel files instead of directory.
- Checking a Project Panel checkbox will clear search directory and search directory releated checkboxes.
- Specifying a directory or checking any directory related checkboxes will clear all Project Panel checkboxes.

Behaviour change on start
Old: If a Workspace file is specified as command line parameter, then Notepad++ opens this workspace file in Project Panel 1. But if more than one Project Panel is already open, then the view shows the last Project Panel, and not the Project Panel with the Workspace file specified in the command line.
New: If a Workspace file is specified as parameter, then always Project Panel 1 will be displayed.

Bugfix
- If the NativeLanguageSpeaker class fails to initialize because the language files are missing, then a problem occured while it attempted to translate its own "... not found" message.


PowerEditor/visual.net/notepadPlus.vcxproj
```diff
@@ -23,12 +23,12 @@
     <ProjectGuid>{FCF60E65-1B78-4D1D-AB59-4FC00AC8C248}</ProjectGuid>
     <RootNamespace>Notepad++</RootNamespace>
     <Keyword>Win32Proj</Keyword>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
donho, line 26: notepadPlus.vcxproj should not be modified.

PowerEditor/src/localization.cpp
```diff
@@ -1299,6 +1299,10 @@ generic_string NativeLangSpeaker::getAttrNameStr(const TCHAR *defaultStr, const
 
 int NativeLangSpeaker::messageBox(const char *msgBoxTagName, HWND hWnd, const TCHAR *defaultMessage, const TCHAR *defaultTitle, int msgBoxType, int intInfo, const TCHAR *strInfo)
 {
+	if (!this)
```
donho, line 1302: In which circumstance `this` will be NULL ?

donho:
For each PR, there should be an issue created for the feature request or bug.
A big feature should be broken down to the small features so the PRs are easier to be reviewed.
 
Please refer to contribution rules:

Uhf7:
It was before the changes in Parameters.cpp committed at 2018-08-19 (3325da03a0bad59527a778a4aa0601bd1db3ef06). Since then, there is an "if (_pNativeLanguageSpeaker)"-statement is wrapped around every call of messageBox, and if _pNativeLanguageSpeaker is NULL, then the Windows MessageBox function is called instead. Placing this if statement directly within the messageBox function has the same effect and seemed easier to me. The original problem was a wrong configuration, which lead to an (unwrapped) early messageBox call in NppParameters::load (), I believe.

sasumner:
I'm confused.  This PR has been hanging around since June of 2018?  If @donho was going to review it now, I'm surprised that he didn't ask for the changes to be reapplied to the newest codebase.

Also, does this PR link back to an "issue"?  The proper protocol is that an issue is opened to discuss a proposed code change, and then a PR to make the change (and the PR is supposed to have a reference back to the issue).

Uhf7:
I reapplied the PR to the newest codebase at April, 9, 2020 unasked, because I wanted a new feature from the current version. And, yes, I did not follow the proper protocol, this was lack of experience, because this was my first activity on this platform ever. That might be the reason, why it was hanging around so long.
I try it again, with "issue" and all. And thank you for reviewing.

sasumner:
@Uhf7 Any chance to "try again" with these modifications?  They seem valuable.

daxliniere:
Rather than getting caught up on semantics, can we put that energy into implementing the feature? The Projects system seems like a great idea (Find in Projects especially), but there's not even any info on how to use it. Like, how do you 'move' an open file  into a project?

alankilborn:
@daxliniere "Projects" section of https://npp-user-manual.org/docs/session/ ?

daxliniere:
Dragging was the first thing I tried, but it's not a valid move:
image

alankilborn:
@daxliniere 

To quote from the documentation linked to earlier: "You can move a file from one place in the Project panel into a different Folder or Project by dragging the file within the Project Panel and dropping it on the name of the Project or Folder you want the file to move into. Other than that, **there aren’t other drag-and-drop features in the Project Panel.**"

daxliniere:
Sorry, so how do you get files into a project? 
Obviously this is for a separate feature request ticket, but from a UI point of view, the drag'n'drop method I alluded to in my screenshot seems like the easiest and most logical.

alankilborn:
@daxliniere 

---

>how do you get files into a project?

Right-click the project:

image

---

>from a UI point of view, the drag'n'drop method I alluded to in my screenshot seems like the easiest and most logical.

Yes, and as you also said, this is a separate feature-request-issue (if there isn't one already open for it).

---

@pryrt Is the user manual sufficient as-is for how to add files, or should more be added?

pryrt:
> Is the user manual sufficient as-is for how to add files, or should more be added?

I mean, _I_ think so.  The [Project Panel Specifics](https://npp-user-manual.org/docs/session/#project-panel-specifics) start off by saying that all Project Panel actions are accessed through right-clicking various entries in the panel, then it breaks it down into a list of what each command in the right click menu does for each of the entries in the panel; searching for "add" or "add file" on that page would take you quickly to the **Add Files** action that you highlighted from the screenshot.  

And in the [Project Panels](https://npp-user-manual.org/docs/session/#project-panels) prose description just above that, it gives conceptual level rather than menu level descriptions.  And there's an entire paragraph on what "drag and drop" features there are, and it's rather explicit that, "Other than that, there aren’t other drag-and-drop features in the Project Panel."

So I would think that if someone was able to find the Project Panels section in the User Manual, they would be able to figure out how to use any aspect of the Project Panels that they are interested in.

alankilborn:
@pryt said:

>I mean, I think so.

I think so, too, just wanted an independent opinion.

Others confused about the "project panel" feature should RTFM.

PR #4611, opened 2018-06-28
Restore window drawing after aborted shutdown

AngryGamer:
This bug is obnoxious. I've run into it a couple times and finally decided to fix it.

**How to replicate bug:**
Open scratch tab and input random text, then delete it from _AppData\Roaming\Notepad++\backup_ and try closing Notepad++ which will trigger a "Your backup file cannot be found (deleted from outside)" prompt, but click cancel on it. Enjoy using Npp after triggering this bug. 😋


donho:
Don't use `AngryGamer:master` as branch name for the next time.

Issue #852, opened 2015-09-03
Wish: "ignore first item" command line argument (for replacing Notepad)


maphew:
Please consider adding a new command line argument to NPP which says "discard the first item in the command line arguments list"

Example invocation: `notepad++.exe --isDebugger arg1 arg2 arg3`

Turns into: `notepad++.exe arg2 arg3`
# Why?

To allow replacing Notepad with Notepad++ without laboriously taking over all relevant file associations or touching system files. 
## Reasoning

The least intrusive and most light weight method to use NPP as the default text file application across the board is to use _Image File Execution Options_. This tell Windows to use notepad++.exe instead of notepad.exe any place notepad is the default action -- double clicking on _something.txt_, _foobaz.log_ for example.

Using image execution means a person can ignore [taking over file associations](http://superuser.com/a/474081/16966) (which is straightforward but involves _n_-dozen distinct changes to system and/or user defaults), and not dive into the dubious act of bypassing system files and their protective shield (rename notepad++.exe to notepad.exe and [replace all copies of notepad.exe](https://gist.github.com/amarraja/3938304) on C:).

It would also allow streamlining the thousand word explanation [Replacing Notepad](http://docs.notepad-plus-plus.org/index.php?title=Replacing_Notepad) wiki page down to 2 or three sentences and a single command

```
reg add "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /t REG_SZ /d "\"%ProgramFiles(x86)%\Notepad++\notepad++.exe\" --isDebugger" /f
```

(Even better, this single command could be added under _Settings > Preferences_ for an even smoother user experience and less documentation.)

---

Now this already works! You can use image execution right now. Try it without the made up `--isDebugger` parameter:

```
reg add "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /t REG_SZ /d "\"%ProgramFiles(x86)%\Notepad++\notepad++.exe\"" /f
```

...but it has the unfortunate side effect of opening the binary notepad.exe as a document in NPP everytime it's used.

![npp-opens-exe](https://cloud.githubusercontent.com/assets/486200/9667760/91d28956-5232-11e5-8f82-5a3f9e21a394.png)

To undo the change simply:

```
[...]

Issue #3987, opened 2017-12-11
Feature Request: Adding command line switches for "print" and "printto" to mimic notepad.exe


dodmi:
### Description of the Issue
I've replaced Notepad with Notepad++ [according to the official guide](http://docs.notepad-plus-plus.org/index.php?title=Replacing_Notepad).
This is working fine for opening documents. But the windows context menu also provides methods for printing files, associated with Notepad.exe and also some programs rely on these command line switches, which will fail if you replaced Notepad.exe.
`notepad.exe /P <filename>`
`notepad.exe /PT <filename> <printername> <driverdll> <port>`
It would be great if you could implement these command line switches to Notepad++ 

### Steps to Reproduce the Issue
1.  Replace Notepad by Notepad++ (by any method)
2.  Right click a text file and select "Print"

### Expected Behavior
The file gets printed without any further notification and closed afterwards (i.e. if Notepad++ wasn't already running, it will be closed afterwards, maybe a new instance should be used?)

### Actual Behavior
A message appears, that the file "C:\p" doesn't exist and if you like to create ist.

### Debug Information
Notepad++ v7.5.3   (32-bit)
Build time : Dec  5 2017 - 10:45:46
Path : C:\Program Files (x86)\Office\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)

donho:
I will see what I can do about it.

Big--Mike:
Thank you :)

donho:
@Big--Mike So could we consider this feature has been done by https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4620 ?

Big--Mike:
@donho It's partly done. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4620 emulates the /P parameter, not /PT. 

I'm using a Software which uses the command in the windows "PrintTo" context menu to convert various file types to pdf by sending them to a virtual PDF printer. The default for textfiles is **%SystemRoot%\system32\notepad.exe /pt "%1" "%2" "%3" "%4"**.
So, I have a use case, and Notepad++ will still not work in this situation. But this feature requires more changes as renaming the existing **-quickprint** to **/P**.

It would be nice to have support for this feature, but I understand, that the benefit is low compared to the feature in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4620, which helps many people to simply replace Notepad.

[...]

PR #4620, opened 2018-07-02
Implemented Wish: "ignore first item" command line argument (for replacing Notepad)

CookiePLMonster:
This PR is split into two commits to ease review:

1. First commit **hugely** simplifies command line parsing (105 LoC -> 12 LoC) by replacing code manually tokenizing arguments with a call to `CommandLineToArgvW` - which does exactly what the original function does, no more, no less. And improving readability (and reliability) by removing code is good, right?
2. Second commit implemented #852 - a new `-z` command line argument has been added, which will cause the argument following it to be ignored (in our case, removed from `ParamVector` entirely).

With changes from this PR, it is now possible to override notepad.exe with notepad++.exe as described in #852, by invoking a command like so:

```
reg add "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /t REG_SZ /d "\"%ProgramFiles(x86)%\Notepad++\notepad++.exe\" -z" /f
```

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -111,6 +111,41 @@ void stripIgnoredParams(ParamVector & params)
 	}
 }
 
+// 1. Converts /p to -quickPrint if it exists as the first parameter
+// 2. Concatenates all remaining parameters to form a file path, adding appending .txt extension if necessary
+// This seems to mirror Notepad's behaviour
+void convertParamsToNotepadStyle(ParamVector & params)
+{
+	ParamVector newParams;
+	auto it = params.begin();
+	if ( it != params.end() && lstrcmpi(TEXT("/p"), it->c_str()) == 0 ) // Notepad accepts both /p and /P, so compare case insensitively
+	{
+		++it;
+		newParams.emplace_back(TEXT("-quickPrint"));
+	}
+
+	bool ssHasContents = false;
+	generic_stringstream ss;
+	for ( ; it != params.end(); ++it )
+	{
+		ssHasContents = true;
+		ss << *it;
+		if ( std::next(it) != params.end() ) ss << TEXT(" ");
+	}
+	
+	if ( ssHasContents )
+	{
+		generic_string str = ss.str();
+		if ( *PathFindExtension(str.c_str()) == '\0' )
+		{
+			str.append(TEXT(".txt")); // If joined path has no extension, Notepad adds a .txt extension
+		}
+		newParams.push_back(std::move(str));
+	}
+
+	params = std::move(newParams);
+}
+
 bool isInList(const TCHAR *token2Find, ParamVector& params, bool eraseArg = true)
 {
 	for (auto it = params.begin(); it != params.end(); ++it)
@@ -247,6 +282,7 @@ const TCHAR FLAG_OPENSESSIONFILE[] = TEXT("-openSession");
 const TCHAR FLAG_RECURSIVE[] = TEXT("-r");
 const TCHAR FLAG_FUNCLSTEXPORT[] = TEXT("-export=functionList");
 const TCHAR FLAG_PRINTANDQUIT[] = TEXT("-quickPrint");
+const TCHAR FLAG_NOTEPAD_COMPATIBILITY[] = TEXT("-notepadStyleCmdline");
 
 
 void doException(Notepad_plus_Window & notepad_plus_plus)
@@ -288,6 +324,12 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int)
 	::CreateMutex(NULL, false, TEXT("nppInstance"));
 	if (::GetLastError() == ERROR_ALREADY_EXISTS)
 		TheFirstOne = false;
+
+	// Convert commandline to notepad-compatible format, if applicable
+	if ( isInList(FLAG_NOTEPAD_COMPATIBILITY, params) )
+	{
+		convertParamsToNotepadStyle(params);
+	}
 
 	bool isParamePresent;
 	bool showHelp = isInList(FLAG_HELP, params);
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -76,122 +76,27 @@ void allowWmCopydataMessages(Notepad_plus_Window& notepad_plus_plus, const NppPa
 	}
 }
 
-
-bool checkSingleFile(const TCHAR *commandLine)
-{
-	if (!commandLine || commandLine[0] == TEXT('\0'))
-		return false;
-
-	TCHAR fullpath[MAX_PATH] = {0};
-	const DWORD fullpathResult = ::GetFullPathName(commandLine, MAX_PATH, fullpath, NULL);
-
-	if (fullpathResult == 0)
-		return false;
-
-	if (fullpathResult > MAX_PATH)
-		return false;
-
-	if (::PathFileExists(fullpath))
-		return true;
-
-	return false;
-}
-
 //commandLine should contain path to n++ executable running
-void parseCommandLine(const TCHAR* commandLine, ParamVector& paramVector)
+ParamVector parseCommandLine(const TCHAR* commandLine)
 {
-	if (!commandLine)
-		return;
-
-	TCHAR* cmdLine = new TCHAR[lstrlen(commandLine) + 1];
-	lstrcpy(cmdLine, commandLine);
-
-	TCHAR* cmdLinePtr = cmdLine;
-
-	//remove the first element, since thats the path the the executable (GetCommandLine does that)
-	TCHAR stopChar = TEXT(' ');
-	if (cmdLinePtr[0] == TEXT('\"'))
+	ParamVector result;
+	int numArgs = 0;
+	LPWSTR* tokenizedCmdLine = CommandLineToArgvW( commandLine, &numArgs );
+	if ( tokenizedCmdLine != nullptr )
 	{
-		stopChar = TEXT('\"');
-		++cmdLinePtr;
+		result.assign( tokenizedCmdLine+1, tokenizedCmdLine+numArgs ); // If numArgs == 1, it will do nothing
+		LocalFree( tokenizedCmdLine );
 	}
-	//while this is not really DBCS compliant, space and quote are in the lower 127 ASCII range
-	while(cmdLinePtr[0] && cmdLinePtr[0] != stopChar)
-    {
-		++cmdLinePtr;
-    }
-
-    // For unknown reason, the following command :
-    // c:\NppDir>notepad++
-    // (without quote) will give string "notepad++\0notepad++\0"
-    // To avoid the unexpected behaviour we check the end of string before increasing the pointer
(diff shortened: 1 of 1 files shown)


donho:
@CookiePLMonster 
-z option works fine with `reg add "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /t REG_SZ /d "\"%ProgramFiles(x86)%\Notepad++\notepad++.exe\" -z" /f` (ie. no notepad.exe opened in Notepad++)
However, if double click on a file name contains space, `new file.txt` for example, Notepad++ will treat as 2 file `new` & `file.txt`.
Any solution for that?

Big--Mike:
@donho That's because notepad registers file types to run with _notepad.exe %1_ instead of _notepad.exe "%1"_.
It's working in Notepad, because Notepad analyzes its command line parameters. If the parameters build a path to an existing file, the parameters will be merged to one file path and the file will be opened.

Another issue with Notepad compatibility is printing file from explorer by right click menu. See [my issue](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3987)

donho:
@Big--Mike 
> That's because notepad registers file types to run with notepad.exe %1 instead of notepad.exe "%1".
It's working in Notepad, because Notepad analyzes its command line parameters. If the parameters build a path to an existing file, the parameters will be merged to one file path and the file will be opened.

Thank you for your answer. So is there any way (to add a `reg` command for example) to work around this problem?

CookiePLMonster:
We could probably replicate this behaviour

`If the parameters build a path to an existing file, the parameters will be merged to one file path and the file will be opened.`
easily by combining all commandline parameters (after stripping -z) and checking if their sum is a path to a file.

Let me run some tests on normal Notepad to verify if this is the case.

Big--Mike:
I think this must be done in the Notepad++ program logic itself.
The registry key above generally leads to the following behavior:
If anywhere in Windows **notepad.exe** is started with command line arguments **arg1**, **arg2**, it will lead to the following command line to be executed: **"%ProgramFiles(x86)%\Notepad++\notepad++.exe" -z "C:\Windows\System32\notepad.exe" arg1 arg2**
You can't influence the part after **-z** (therefore **-z** is used to simply omit the first argument, which is always the path to the Notepad executable). The arguments passed to notepad.exe, will be passed exactly to notepad++.exe instead.

CookiePLMonster:
Do you think joining all arguments on space and checking if that result points to an existing file is sufficient?

CookiePLMonster:
That is true, but it's very unlikely that a series of legimate arguments will end up forming a valid absolute path to an existing file.

Big--Mike:
It will probably work in 90% of cases...
Actually I just did some tests:
**notepad 1 2 /p** gives me an error, that the path was not found, while **notepad /p 1 2** would print the file "1 2" (If it exists, else Notepad asks to create "1 2.txt")
So, if there is a "real parameter", it's in the first argument. (I only know **/p** and **/pt**, but any other will also start with a **/**) 
**notepad 1 2** asks to create a file "1 2.txt", so it doesn't even check if the file exists, but simply puts quotes around the arguments (other then the first one, if it's a real parameter).

As Notepad++ supports plenty parameters on any position, it's not so easy there. 
You could optionally make this behavior dependent from the **-z** switch. As it's probably only used for the special case of replacing Notepad.exe, it could also switch the command line handling to a "Notepad compatibility mode".

Actually I'm using a vbs script to make the command line handling, which works great and does the following:
- throw away the first argument (this is what **-z** will do)
- iterate through the arguments
-- if the argument is **/p**, add **-quickPrint** to the future command line and throw the argument away
-- if the argument contains a space, put it into quotes and add it to the new arguments string (it was already quoted in this case)
-- if the argument doesn't contain a space, simply add it to the new arguments string

Afterwards, I test if the complete arguments string is a valid file name. If so, I quote it to have Notepad++ open the file, else Notepad++ a file for each argument.
I just can't handle Notepad's **/PT**, as there's no way to tell Notepad++, which printer to use.

CookiePLMonster:
@donho What do you think about the proposal above?

donho:
> As Notepad++ supports plenty parameters on any position, it's not so easy there.
You could optionally make this behavior dependent from the -z switch.

Yes, indeed. Let's treat such problem only in -z, but not globally.

> -- if the argument is /p, add -quickPrint to the future command line and throw the argument away

Nice to have feature. See:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3987

> -- if the argument contains a space, put it into quotes and add it to the new arguments string (it was already quoted in this case)
-- if the argument doesn't contain a space, simply add it to the new arguments string

Sounds good to me.

@CookiePLMonster Will you enhance in this PR?

CookiePLMonster:
Sure! Do you think it still makes sense to call it `-z` or maybe rename to something else indicating "notepad compatibility"?

donho:
> Do you think it still makes sense to call it -z or maybe rename to something else indicating "notepad compatibility"?

`-notepadCompatibility` is quite long, what's your suggestion?

CookiePLMonster:
Tough one... `-nc` historically stands for something else if I recall correctly, so might not be a good idea to use that.

Honestly, no idea.

cmeriaux:
What about -skipFirstArg

CookiePLMonster:
Since this would skip the arg and also change the way commandline is interpreted, not sure if it's a good idea.

Big--Mike:
Personally, I'm a fan of having the choice - speaking parameters in cases where understanding the purpose is more important than the length. And short ones for everyday typing.
On Linux for most command line tools, you have  two parameter for one purpose.
One to use in a script (better understanding) and one for the command line (for lazy people like me)
It's not as you have to type this parameter everyday and it you can guess the purpose with one glance in the registry key... So, I'd vote for a long version.

The parameters of Notepad++ aren't very consistent, by the way:
There are some short parameters, some long parameters and a double dashed help, which doesn't fit at all to the rest.
Maybe you could also use two parameters
notepad++.exe -notepadCompatibility -skipNextArg 
would work perfectly (please don't call it -skipFirstArg...)

donho:
> (please don't call it -skipFirstArg...

@Big--Mike  Why not?

Big--Mike:
Because the parameter causes Notepad++ to skip the next/following argument, not the first one...

CookiePLMonster:
No, -z ignores the argument directly following it, so:

`notepad++.exe -z arg1 arg2 arg3` = `notepad++.exe arg2 arg3`

or

`notepad++..exe arg1 -z arg2 -z arg3 arg4 -z` = `notepad++ arg1 arg4`, with the last -z doing nothing (because it's not followed by any argument).

ilendemli:
What if you right-click a text file and select print? It will add a /p after notepad.exe but before the file path. Maybe check for that aswell and open the print dialog?

Big--Mike:
/p will be handled in a way:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4620#issuecomment-407235706

ilendemli:
oh absolutely overflown that part of the conversation, sorry my bad

donho:
@CookiePLMonster Just wanna know if you will enhance this PR for managing the space in the argument given?

CookiePLMonster:
Sure, I kinda forgot about this - do we have anything detailing how it is
expected to work, then? A commandline argument to switch nop processor unto
a "notepad compatibility mode"?

Big--Mike:
My tests of Notepad's behavior are described here:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4620#issuecomment-406787495

CookiePLMonster:
That looks good! I'll look into it soon.

CookiePLMonster:
@Big--Mike @donho I have now added a new commit to this PR, adding `-notepadStyleCmdline`. It does the following:

1. Converts /p to -quickPrint if it exists as the first parameter
2. Concatenates all remaining parameters to form a file path, adding appending .txt extension if necessary

I have **not** tested this under a desired use case, however it appears that remaining arguments (that is, those left after discarding arguments via `-z`) form a file for Notepad++ to open. This should satisfy the requirements.

Big--Mike:
@CookiePLMonster Sounds great, thank you

donho:
so we should pass both -z and -notepadStyleCmdline ?

CookiePLMonster:
They do different things, so yes. `-z` only skips the parameter.

I felt like coupling both into one command line switch introduces too much implied behaviour, so they are separate.

donho:
I run the following command again, still doesn't work:
`reg add "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /t REG_SZ /d "\"%ProgramFiles(x86)%\Notepad++\notepad++.exe\" -z -notepadStyleCmdline" /f`

And even -z dosn't work anymore.

What's the command to use exactly?

CookiePLMonster:
`-z` needs to go last - after all, you're skipping **next** argument. In your case, you skipped `-notepadStyleCmdline`

donho:
OK, got it.
It's rather `reg add "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /t REG_SZ /d "\"%ProgramFiles(x86)%\Notepad++\notepad++.exe\" -notepadStyleCmdline -z" /f`

donho:
Why do we need to separate -z & -notepadStyleCmdline anyway?

CookiePLMonster:
> I felt like coupling both into one command line switch introduces too much implied behaviour, so they are separate.

I don't know about you, but I could imagine the case where people want a Notepad style command line but they don't need to skip any parameters (eg. old batch scripts which they'd like to migrate by doing Find & Replace from `notepad.exe` to `notepad++.exe -notepadStyleCmdline`.

Big--Mike:
@donho Yes, 
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4620#issuecomment-406787495
describes the behavior of Notepad in Windows 10 1803.

donho:
@CookiePLMonster & @Big--Mike 
I just did a note for replacing Notepad:
https://notepad-plus-plus.org/features/replace-notepad.html

Please let me know if there are some errors or some information is missing.
(BTW, what's the opposite command line to undo Notepad replacement?)

Big--Mike:
@donho Yes, you need to run the command from an elevated command prompt (Changing the registry in HKLM needs admin privileges). Also it needs to be cmd.exe if you use environment variables in this way, the command will fail in PowerShell.

The default install location depends on the OS:
x86 OS/x86 NP++: "%ProgramFiles%\Notepad++\"
x86 OS/x64 NP++: _won't work_
x64 OS/x86 NP++: "%ProgramFiles(x86)%\Notepad++\"
x64 OS/x64 NP++: _?? If the installer is also 64bit, it probably defaults to "%ProgramFiles%\Notepad++\", if the installer is x86 it will default to "%ProgramFiles(x86)%\Notepad++\"_

The command line to remove the Debugger value and to restore the default behavior is
**reg delete "HKLM\Software\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\notepad.exe" /v "Debugger" /f**

Issue #4621, opened 2018-07-02
Notepad++ can crash by launching it with specific commandline


CookiePLMonster:
### Description of the Issue
If running Notepad++ in mono-instance mode, it is possible to crash the instance by trying to open a new one with specific commandline arguments. A quick scan through the code reveals that commandline arguments are passed between instances via `WM_COPYDATA` - but it also tries to pass two fields of type `std::string`! Since it's illegal to move pointers between applications like this, code attempting to assign said strings will crash a currently opened Notepad++ instance. Therefore, it is technically possible to write a "npp killer" batch script.

### Steps to Reproduce the Issue
1. Make sure your Notepad++ is running in mono-instance mode.
2. Open any file via Notepad++ - without using any commandline arguments, like so:
```notepad++.exe file1.txt```
3. Open another file - this time, passing a file name and a valid string-based commandline argument (like `-L`):
```notepad++.exe -Lpl file2.txt```
4. Observe a crash.

### Expected Behavior
Notepad++ opens a new file in currently opened instance.

### Actual Behavior
Notepad++ crashes.

### Debug Information
I have verified the bug on released v7.5.6 and also on self-compiled up to date code, so v7.5.7 does not fix it.

Notepad++ v7.5.6   (64-bit)
Build time : Mar 19 2018 - 00:23:17
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

Since this is also a problem on current version, I will follow with a PR fixing this shortly.

CookiePLMonster:
Fixed with #4622

donho:
@CookiePLMonster I can reproduce the crash. Thank you.
How come `-Lpl` make Notepad++ crash?

CookiePLMonster:
The structure which was passed contained a non-POD type, std::string. Passing data is obviously a shallow copy, so with `-Lpl` this string held a "pl" string - thus, shallow copy of std::string passed pointers of instance A to instance B, but not their contents. Thus, trying to access the string crashed.

I created this DTO object as a PDO type, which can be safely copied around. I could have also passed those strings safely, but they are **not** used after being passed to a new instance anyway.

PR #4622, opened 2018-07-02
Fix for "Notepad++ can crash by launching it with specific commandline"

CookiePLMonster:
A fix for #4621. Please refer to this issue for detailed description and a repro path.

I created a new POD type which gets transferred between instances, instead of existing `CmdLineParams` which is **not** a POD type (because of `std::string`, whose contents don't need to be transferred between instances anyway).

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -235,6 +235,38 @@ struct CmdLineParams
 	}
 };
 
+// A POD class to send CmdLineParams through WM_COPYDATA and to Notepad_plus::loadCommandlineParams
+struct CmdLineParamsDTO
+{
+	bool _isReadOnly;
+	bool _isNoSession;
+	bool _isSessionFile;
+	bool _isRecursive;
+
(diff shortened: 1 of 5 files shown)


PowerEditor/src/Parameters.h
```diff
@@ -235,6 +235,38 @@ struct CmdLineParams
 	}
 };
 
+// A POD class to send CmdLineParams through WM_COPYDATA and to Notepad_plus::loadCommandlineParams
```
donho: POD? Ping Of Death? Penis Of Drag-queen ?

PowerEditor/src/Parameters.h
```diff
@@ -235,6 +235,38 @@ struct CmdLineParams
 	}
 };
 
+// A POD class to send CmdLineParams through WM_COPYDATA and to Notepad_plus::loadCommandlineParams
+struct CmdLineParamsDTO
```
donho: DTO ?

PR #4624, opened 2018-07-02

cmeriaux:
When we are restoring a wrapped buffer,
we send SCI_VISIBLEFROMDOCLINE message to convert a file line number to a visual line number,
scintilla seems to ignore the "wrap" option of this buffer.
Workaroud: execute SCI_ENSUREVISIBLE after set position and before executing SCI_VISIBLEFROMDOCLINE.

It seems that this issue has been encountered in ScitillaComponent/FindReplaceDlg.cpp in Searching::displaySectionCentered


cmeriaux:
Here is a video of the bug
https://youtu.be/Vpm-rwnGMzs

cmeriaux:
Here is a video of the first commit that is a partial fix
https://youtu.be/6KDzZzUbFng

cmeriaux:
Here is the video of the full fixed issue
https://youtu.be/0MHgfSEjemQ

Issue #4682, opened 2018-07-20
Document List Panel settings bug


LordRibblesdale:
### Description of the Issue
On Notepad++ 32bit version (tested on 7.5.6 and 7.5.7), the program crashes while setting up Document List Panel

### Steps to Reproduce the Issue
1. Open Setting -> Preferences
2. Check Show in Document List Panel
3. Check and uncheck Disable extension column

### Expected Behavior
Extension bar on the left should disappear, as do on 64bit version

### Actual Behavior
The program crashes

### Debug Information
Notepad++ v7.5.7   (32-bit)
Build time : Jun 29 2018 - 00:46:55
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

Yaron10:
@LordRibblesdale,

This is very interesting.
Did you build the binary yourself or you're using the official/original notepad++.exe?

See: https://notepad-plus-plus.org/community/topic/15438/multiple-errors-building-npp-7-5-6/11

LordRibblesdale:
Using official notepad++ from official installer from website!

Yaron10:
Thanks.

The crash does not occur on every machine or in any constellation.
But the issue should certainly be investigated.

AngryGamer:
I can't replicate this issue. Try out a [fresh portable copy](https://notepad-plus-plus.org/repository/7.x/7.5.7/npp.7.5.7.bin.x64.zip) extracted to a random directory so it's using stock settings, does the crash still happen?

LordRibblesdale:
On my system, also portable build still crashes

AngryGamer:
Can have I see debug info on the portable copy to ensure "Local Conf mode : ON"?

LordRibblesdale:
Notepad++ v7.5.7   (32-bit)
Build time : Jun 29 2018 - 00:46:55
Path : C:\Users\Green\Downloads\npp.7.5.7.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll 

Windows 10 Version 1607 14393.2363

AngryGamer:
[...]

PR #4684, opened 2018-07-21
Fixed crash Issue by checking & unchecking "Disable extension column" in preferences dialog

SinghRajenM:
Fixed Crash issue #4682 

**Also fixes additional issue** which can be seen only in x64 version (x86 version crashes 😉) with following steps.
1. Make sure both the checkbox ("Show" and "Disable extension column") are unchecked for "Document list panel". If not, then disable both of them and restart npp.
2. Now open preferences and check  "Disable extension column", then check "Show"
3. Observe "Doc switcher" panel. Only one column ("Name") is available.
4. Now uncheck "Disable extension column", and observe "Doc switcher" panel again. This time "Ext" is excepted but it is not visible.

image

### DIFF
--- a/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
+++ b/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
@@ -122,6 +122,7 @@ void VerticalFileSwitcherListView::initList()
 		{
 			// resize "Name" column when "exts" won't fit
 			LVCOLUMN lvc;
+			lvc.mask = LVCF_WIDTH;
 			SendMessage(_hSelf, LVM_GETCOLUMN, 0, reinterpret_cast<LPARAM>(&lvc));
 			
 			if (lvc.cx + 50 > totalWidth)


AngryGamer:
The patch seems to work. Npp x86 wasn't crashing on my systems though. The issue manifested itself in the form of the "Name" column's label vanishing.

SinghRajenM:
```C++
	LVCOLUMN lvc;
	SendMessage(_hSelf, LVM_GETCOLUMN, 0, reinterpret_cast<LPARAM>(&lvc));
		
	if (lvc.cx + 50 > totalWidth)
	{
		lvc.cx = totalWidth - 50;
		SendMessage(_hSelf, LVM_SETCOLUMN, 0, reinterpret_cast<LPARAM>(&lvc));
	}
```
From above code, we are interested to get value of `cx`. As per [MSDN](https://docs.microsoft.com/en-us/windows/desktop/api/Commctrl/ns-commctrl-taglvcolumna), if you want to get value for `cx`, then mask should be set to `LVCF_WIDTH`.

Also as per [MSDN](https://docs.microsoft.com/en-us/windows/desktop/controls/lvm-setcolumn), `LVM_SETCOLUMN` mask member should be set.

> LVM_SETCOLUMN message
Sets the attributes of a list-view column. You can send this message explicitly or by using the ListView_SetColumn macro.
Parameters
wParam
    Index of the column.
lParam
    Pointer to an LVCOLUMN structure that contains the new column attributes. The mask member specifies which column attributes to set. If the mask member specifies the LVCF_TEXT value, the pszText member is the address of a null-terminated string and the cchTextMax member is ignored.


image


Hope this answers your question.

Issue #4644, opened 2018-07-06
Folder as Workspace doesn't update as new files or folders are created.


warg0x4e:
### Description of the Issue
**Notepad++ version 7.5.7**
_Folder as Workspace_ doesn't update as new files or folders are created.

### Steps to Reproduce the Issue
The following steps are demonstrated at [https://youtu.be/nCy_MT46enY](https://youtu.be/nCy_MT46enY).
1.  Add a folder to _Folder as Workspace_.
2.  Create new files or folders in the folder.

### Expected Behavior
I expect to see the new files and folders in the _Folder as Workspace_.

### Actual Behavior
New files and folders aren't displayed in the _Folder as Workspace_.
![36704631_10216348229560021_4575886293489156096_n](https://user-images.githubusercontent.com/2155010/42358345-2a8b0f3a-80aa-11e8-9490-132f654c272a.png)

### Debug Information
Notepad++ v7.5.7   (64-bit)
Build time : Jun 29 2018 - 00:53:58
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

pryrt:
Confirmation: see https://notepad-plus-plus.org/community/topic/16040/folder-in-workspace/9 for more users who are seeing the same bug

For me (PeterJones in forum), it worked just fine in installed 7.5.6 32bit, but behaved as @warg0x4e reported in portable 7.5.7 64bit.

\--------
![](https://i.imgur.com/CmOoTNI.gif)

    Notepad++ v7.5.6   (32-bit)
    Build time : Mar 19 2018 - 00:26:59
    Path : C:\Program Files (x86)\Notepad++\notepad++.exe
    Admin mode : OFF
    Local Conf mode : OFF
    OS : Windows 10 (64-bit)
    Plugins : ComparePlugin.dll dbgpPlugin.dll DSpellCheck.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll NppFTP.dll PluginManager.dll PreviewHTML.dll PythonScript.dll XMLTools.dll

\--------
![](https://i.imgur.com/GiDb7Tv.gif)

	Notepad++ v7.5.7   (64-bit)
	Build time : Jun 29 2018 - 00:53:58
[...]

PR #4685, opened 2018-07-21
Fixed Regression issue: Folder as workspace not updating dir changes

SinghRajenM:
Fixed issue #4644 

The simplest fix is changing below code in all the three messages (FB_ADDFILE, FB_RMFILE and FB_RNFILE)

from 
```C++
generic_string separator = TEXT("\\\\");
size_t sepPos = file2Change[0].find(separator);
``` 
to 
```C++
generic_string separator = TEXT("\\");
size_t sepPos = file2Change[0].find_last_of(separator);
``` 

But just to remove duplicate code from all three sections, new method ```std::vector<DirChangeInfo> getDirChangeInfo(const std::vector<generic_string>& changeInfo);``` is introduced.

This regression was introduced by one of my PR  #4257

### DIFF
--- a/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
+++ b/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
@@ -158,26 +158,15 @@ INT_PTR CALLBACK FileBrowser::run_dlgProc(UINT message, WPARAM wParam, LPARAM lP
 		case FB_ADDFILE:
 		{
 			const std::vector<generic_string> file2Change = *(std::vector<generic_string> *)lParam;
-			generic_string separator = TEXT("\\\\");
+			std::vector<DirChangeInfo> changes = getDirChangeInfo(file2Change);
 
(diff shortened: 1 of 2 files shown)


donho:
> This regression was introduced by one of my PR #4257. Sorry for that. 

In this case, why don't we just revert PR #4257 ?

SinghRajenM:
@donho IMHO, we should not revert PR #4257 because of below reasons - 

1. As PR #4257 intended to fix the coding error (actually typo in the code).
2. We should not rely on ```generic_string separator = TEXT("\\\\");``` to find out folder and file name in the given path. What if tomorrow messages (FB_ADDFILE, FB_RMFILE and FB_RNFILE) are sent from somewhere else which may not contain double slash in it (e.g. Added new functionality to add/remove/rename file or folder from project panel's context menu. I may submit a PR for same soon.)?
3. Messages (FB_ADDFILE, FB_RMFILE and FB_RNFILE) contains duplicate code which should be moved to common code.

PR #4704, opened 2018-07-27
Improve Installer (Fixes #4083)

### LINKED ISSUE #4083: Uninstall DSpellCheck
### Description of the Issue
DSpellCheck is a new plugin that was installed with a recent NPP version. I don't need it and always disable the checking. Now I've divcovered that it hijacks the very important hotkey Alt+D which opens the File menu in the German language.

There used to be a plugin manager but that seems to have gone away, too. So there's no way to uninstall the spell checking and restore the correct function of the Alt+D hotkey.

1. Install Notepad++ with German translation.
2. Press Alt+D to open the File menu ("Datei")

### Expected Behavior
File menu opens

### Actual Behavior
Some other unknown context menu opens at the mouse cursor

### Debug Information
Notepad++ v7.5.4   (32-bit)
Build time : Jan  1 2018 - 01:47:45
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)

SinghRajenM:
This improved version of installer handles below scenarios - 
1. User can't remove auto updater.
2. User can't remove a specific/all the plugin(s) (of course talking about default plugin shipped in npp installer)

Even if user re-runs and unchecks either auto updater or any/all plugin(s) or both, s/he will not be able to remove them. S/he has to go the installation directory and delete them manually. _This is little annoying or out of reach for layman._

This also fixes defect #4083.

**Note: this PR handles only auto updater and plugin. Rest (auto-complete, Localization and themes) will be handled in separate PR as these items are less priority. Also it will make review easier.**

### DIFF
--- a/PowerEditor/installer/nsisInclude/binariesComponents.nsh
+++ b/PowerEditor/installer/nsisInclude/binariesComponents.nsh
@@ -102,54 +102,143 @@ ${MementoSection} "Auto-Updater" AutoUpdater
 !endif
 ${MementoSectionEnd}
 
-;Uninstall section
-SectionGroup un.Plugins
-	Section un.NppExport
+!macro RemoveAutoUpdater un
+	Function ${un}removeAutoUpdater
+		DetailPrint "Removing AutoUpdater files..."
+		Delete "$INSTDIR\updater\GUP.exe"
+		Delete "$INSTDIR\updater\libcurl.dll"
+		Delete "$INSTDIR\updater\gup.xml"
+		Delete "$INSTDIR\updater\License.txt"
+		Delete "$INSTDIR\updater\LICENSE"
+		Delete "$INSTDIR\updater\gpl.txt"
+		Delete "$INSTDIR\updater\readme.txt"
+		Delete "$INSTDIR\updater\README.md"
+		Delete "$INSTDIR\updater\getDownLoadUrl.php"
+		RMDir "$INSTDIR\updater"
+	FunctionEnd
+!macroend
+!insertmacro RemoveAutoUpdater ""
+!insertmacro RemoveAutoUpdater "un."
+
+!macro RemoveNppExport un
+	Function ${un}removeNppExport
+		DetailPrint "Removing NppExport plugin files..."
 		Delete "$INSTDIR\plugins\NppExport.dll"
 		Delete "$INSTDIR\plugins\NppExport\NppExport.dll"
 		RMDir "$INSTDIR\plugins\NppExport"
-	SectionEnd
-	
-	Section un.Converter
+	FunctionEnd
+!macroend
+!ifndef ARCH64
+!insertmacro RemoveNppExport ""
+!endif
+!insertmacro RemoveNppExport "un."
+
+!macro RemoveConverter un
+	Function ${un}removeConverter
+		DetailPrint "Removing Converter plugin files..."
 		Delete "$INSTDIR\plugins\NppConverter.dll"
 		Delete "$INSTDIR\plugins\NppConverter\NppConverter.dll"
 		RMDir "$INSTDIR\plugins\NppConverter"
-	SectionEnd
-	
-	Section un.MimeTools
+	FunctionEnd
+!macroend
+!insertmacro RemoveConverter ""
+!insertmacro RemoveConverter "un."
+
+!macro RemoveMimeTools un
+	Function ${un}removeMimeTools
+		DetailPrint "Removing MimeTools plugin files..."
 		Delete "$INSTDIR\plugins\mimeTools.dll"
 		Delete "$INSTDIR\plugins\mimeTools\mimeTools.dll"
 		RMDir "$INSTDIR\plugins\mimeTools"
-	SectionEnd
+	FunctionEnd
+!macroend
+!insertmacro RemoveMimeTools ""
+!insertmacro RemoveMimeTools "un."
 
-	Section un.PluginManager
+!macro RemovePluginManager un
+	Function ${un}removePluginManager
+		DetailPrint "Removing PluginManager plugin files..."
 		Delete "$INSTDIR\plugins\PluginManager.dll"
 		Delete "$UPDATE_PATH\plugins\Config\PluginManager.ini"
 		Delete "$INSTDIR\updater\gpup.exe"
-		RMDir "$INSTDIR\updater\"
-	SectionEnd
-	
- 	Section un.DSpellCheck
+		RMDir "$INSTDIR\updater"
+	FunctionEnd
+!macroend
+;!insertmacro RemovePluginManager "" # To remove nsis "not referenced" warning, should be uncommented when PluginManager is ready
+!insertmacro RemovePluginManager "un."
+
+!macro RemoveDSpellCheck un
+	Function ${un}removeDSpellCheck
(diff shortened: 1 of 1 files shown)


alexhass:
Why should users not able to remove auto-updater? 

I need to remove it to disallow users to update if MSI installer is used. Than the administrator is in charge of updating the software. The user also has no permission and no idea what version is compatible with company applications. So I seriously need to be able to remove updater.

chcg:
@alexhass From my understanding of the PR the issue is about scenario that the option for auto updater is unchecked during installation and the already available auto updater was not properly removed.

donho:
> User can't remove auto updater.

So with this PR, user can remove auto-updater now via installer? How?

alexhass:
Can you tell me how auto update can be force disables without killing plugin installer/updates?

SinghRajenM:
@donho 

In this PR we have created generic methods with the help of macro (like removeAutoUpdater, RemoveNppExport etc. ) which can be called from both installer and uninstaller.

**Let me explain usual uninstaller first**:
During uninstall `SectionGroup un.Plugins` will be executed which will intern call respect function e.g. Call `un.removeNppExport`. Also we have separate section `Section un.AutoUpdater` which will call `Call un.removeAutoUpdater`. **So there is no change in this process... It is an existing process. From code perspective only new methods are called instead of having own removal code.**

**Now let's take example of reinstall/update (_the purpose of this PR_)**: 
Now we have one more section ([here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4704/files#diff-968cf6cd6e0e4f4c60f6311647fcc324R185)). This is hidden section which means it will not be visible on installer UI. This section will be executed during install process (reinstall/update is also logically  install 😊). **Here we checking whether a particular section is checked or not. If NOT checked, then call respective method to remove respective files.**

For example: if auto update is unchecked,  then below steps will be executed
```
	${IfNot} ${SectionIsSelected} ${AutoUpdater}
		call removeAutoUpdater
	${EndIf}
```
It means if user is running installer first time, then this call has no effect because updater files do not exist in install dir. But if user running second time and installer was unchecked, then this call remove the updater files if already exists.

SinghRajenM:
@alexhass you can disable it from notepad++ UI

image

alexhass:
@Singh: nope, this does not work. The checkbox need to be unchecked and disables so a user cannot change it! That means force-disable. Like a GPO policy works.

Users are not allowed to install/update npp+. They are only allowed to install PER USER plugins to their appdata folder.

donho:
@SinghRajenM 
If we do remove auto-updater, then we will do remove default plugins, then auto-completion/localization files, etc...
I prefer to keep our logic as simpler as possible.

donho:
> @alexhass you can disable it from notepad++ UI

It make auto-update not check update for you periodically, but if user can still access to auto-updater via menu command to check availability of new update.

@alexhass the best way to disable it is not to install it.

SinghRajenM:
@donho 
> If we do remove auto-updater, then we will do remove default plugins, then auto-completion/localization files, etc...

**No, not at all. I feel I could not explain in better way.**

Whatever is unchecked only that will be removed (currently with this PR only default plugins and auto updater).

As you know during reinstall/upgrade all the previously installed section are by default checked (due to momentoSection). So Whatever user explicitly unchecking will be removed. Hope, this makes sense.

alexhass:
> It make auto-update not check update for you periodically, but if user can still access to auto-updater via menu command to check availability of new update.
> 
> @alexhass the best way to disable it is not to install it.

@donho  Bad that you do not know your own software. I explain the problem. If the gpup.exe is not in the system the user cannot use your plugin manager. My Users need plugin manager to install addons to APPDATA. You remember - you broken this feature in 7.6+. Can you this issue, please?

Additionally the user must not be able to click check for updates of NPP+ and the auto check also need to be disabled. The user do not has permission to update! He can only update his plugins.

Plugin manager requires the updater to function, but at the same time implements automatic update of NPP+. This need to be splitted like it was in past with plugin manager til 7.5. Can you fix this, please?

Please allow admins to add a registry key to HKLM to disable automatic updates of NPP (but not the per user plugins).

I hope you understand the problem now. Again users do not have permissions locally, but on their APPDATA folder.

sasumner:
@alexhass said:
>@donho Bad that you do not know your own software.

Ouch! :)  You'd better tread lightly.

donho:
@alexhass there's no `gpup.exe` (which is delivered by old PluginManager plugin). I guess you're talking about `gup.exe`.
gup.exe is used by Notepad++ for auto-updater and also by Plugin Admin part. So if you need Plugin Admin, you have to keep `gup.exe`, otherwise you can just remove it to prevent users from clicking it.
I do understand your problem, but that's the current design - I will see what I can do about it.

donho:
@sasumner 


>>    @donho Bad that you do not know your own software.

> Ouch! :) You'd better tread lightly.

No worry, I've got worse :)

alexhass:
@donho: What are you talking about? What schould I not install? Npp?

Are we really talking about the same?

1 users need per user plugins
2 to install plugins they need plugin manager
3 npp will be deployed via msi over ad deployment. Not your nsis setup that is not deployable. You remember?
4. user has no permission to install software locally, only administrators can.
5. because users have no local install permissions they do not need to check for updates of npp. 
6. Auto updates need to be disabled - not by the user - by admin via ad deployment!
7. users only need to update their per user plugins via plugin manager.

Why is it so difficult to understand all this professional requirements?

alexhass:
No worry, I've got worse :)

@donho: do you know that the plugin manager is no longer shown is updater exe file is removed or have you forgot about this? This makes no sense and was not the case in past. Plugin install/update and npp auto update need to be independent from each other. This is not the case since 7.6

donho:
@alexhass 
By "the best way to disable it is not to install it." I meant `GUP.EXE`.

>  to install plugins they need plugin manager

From 7.6, a build-in `Plugin Admin` is used to install/update/remove plugin.
AFAICT, the plugin `Plugin Manager` is outdated (it cannot work correctly with new Plugin Achetecture).

> npp will be deployed via msi over ad deployment. Not your nsis setup that is not deployable. You remember?

NSIS package can be deployed by SCCM. Well it costs, but it can be.

> Plugin install/update and npp auto update need to be independent from each other. This is not the case since 7.6

I'm aware of it, and I told you that I'll try to find a solution.

> user has no permission to install software locally, only administrators can.
> because users have no local install permissions they do not need to check for updates of npp.

OTOH, currently plugins are installed and work only in the same Notepad++ installed location. So if Notepad++ is installed in "Program Files", it'll be good to not install GUP.EXE during the deployment.
`/noUpdater` parameter of Notepad++ installer is there for such purpose.

alexhass:
Might have been confusing, but I meant Plugin Admin when I said Plugin Manager.

Most people do not have SCCM. This is why I build the MSI for you. NSIS cannot customized, MSI can be. NSIS has no repair, MSI has and so on. NSIS crap is past and should RIP.

Looking forward to see the feature to force disable automatic updates of NPP, but allow automatic updates of plugins. Please do not forget to add plugins per user in APPDATA back as users cannot install plugins to local machine. An UAC does not help as they have no admin user credentials. ProgramData folder is not really a good idea, too

PR #4707, opened 2018-07-28
Csharp autocompletion

MCCshreyas:
Added `System.Convert`, `System.IO.File`, `System.Threading.Task` class.


donho:
@MCCshreyas How comes an API file needs to be updated heavily every 2 months by the SAME person?
https://github.com/notepad-plus-plus/notepad-plus-plus/blame/master/PowerEditor/installer/APIs/cs.xml

Issue #4716, opened 2018-07-31
Selection of "working" tab is lost after restart so the first tab is always selected (by default?)


cristian-sulea:
### Description of the Issue

When reopening Notepad++ (with session snapshot enabled) the selected file (tab) at closing time is not reselected. It seems that in the session snapshot is not included anymore the "working" tab, so at reopening time the first tab is selected by default.
This was working in 7.4.x versions (i have 7.4.2 and is working) but in 7.5.x is not working any more.

### Steps to Reproduce the Issue
1. Open Notepad++
2. Open some files (so we have some tabs)
3. Select last opened file (tab) (but the bug can be reproduced with any tab)
4. Close Notepad++ (Preferences/Backup/Session snapshot -> is enabled)
5. Reopen Notepad++

### Expected Behavior
The tab selected at step 3 should be selected after reopening. 

### Actual Behavior
The first tab is always selected after reopening.

### Debug Information
Notepad++ v7.5.8   (64-bit)
Build time : Jul 23 2018 - 02:06:36
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

cristian-sulea:
today i updated to latest version and bug is still there ... uninstall

sasumner:
>today i updated to latest version and bug is still there 

@cristian-sulea Dude, seriously? A fix was committed 8 days ago but a release containing that commit hasn't happened yet.  WTF.

cristian-sulea:
SinghRajenM committed on Jul 31 -> the bug was fixed on JUL 31

SinghRajenM:
@cristian-sulea wait till next release happens. If you can't wait, then get the nightly build binary and replace it to the notepad installation directory.

https://ci.appveyor.com/api/buildjobs/5u4vljudt309nx33/artifacts/Notepad%2B%2B.X64.Unicode%20Release.exe

PR #4722, opened 2018-07-31
Fixed issue: Last active tab is ignored on N++ restart

SinghRajenM:
Fixed issue #4716 and #4755.
As of now this issue is seen only on x64 bit and RCA is below - 
```C++
size_t index2;
TiXmlElement *actIndex = viewRoots[k]->ToElement();
str = actIndex->Attribute(TEXT("activeIndex"), reinterpret_cast<int *>(&index2));
```
casting `size_t` which is ```unsigned __int64``` to ```int``` causes this issue. There is no impact on 32 bit npp as ```size_t``` is ```unsigned int``` on 32 bit.

**Special note:**
1. This is little difficult to reproduce.
2. If variable `index2` is initialized to 0, then I could not reproduce.
3. If variable `index2` is initialized to higher value (e.g. `size_t index2 = 380316380316`), then it is reproduced.
4. As of now, this variable is not initialized, so it can be reproduced only when `index2` is assigned with higher garbage value. If you want to reproduce, just assign higher value during debug to `index2` variable.
5. There might few more places where similar code might exists which requires attention.

image

See the value of `index2`. It should be `3`. Because of this below code gets affected.

image

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -1971,8 +1971,8 @@ bool NppParameters::getSessionFromXmlTree(TiXmlDocument *pSessionDoc, Session *p
 		return false;
 
 	TiXmlElement *actView = sessionRoot->ToElement();
-	size_t index;
-	const TCHAR *str = actView->Attribute(TEXT("activeView"), reinterpret_cast<int *>(&index));
+	int index = 0;
+	const TCHAR *str = actView->Attribute(TEXT("activeView"), &index);
 	if (str)
 	{
 		(*ptrSession)._activeView = index;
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Parameters.cpp
```diff
@@ -1986,15 +1986,15 @@ bool NppParameters::getSessionFromXmlTree(TiXmlDocument *pSessionDoc, Session *p
 	{
 		if (viewRoots[k])
 		{
-			size_t index2;
+			index = 0;
```
chcg: See https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d923ece38df175d8f3cccb187786a62bd07de0e3. Might reintroduce an issue.

SinghRajenM: @chcg do you mean reusing `index` variable again or whole PR? 

If you are talking about `index` variable, then I don't see any issue because function ```actIndex->Attribute(TEXT("activeIndex"), &index)``` is going to overwrite something on variable `index` (at least zero in that function). Am I missing something?

If you are talking about whole PR, then I need figure out more about [d923ece](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d923ece38df175d8f3cccb187786a62bd07de0e3) and what kind of compiling error that PR fixed. But one thing is sure, `size_t` should not be cast to `int` if building for x64. This kind of bugs are obvious while porting 32bit app to 64 bit app which require little extra care.

chcg: Unfortunately the commit doesn't specify more clearly what the observed bug was, but it introduced the additional variable index2. Maybe it is safer to keep it splitted into two independent variables and be sure to avoid crosstalking between activeIndex and activeView.

SinghRajenM: Okay, done!

Issue #2404, opened 2016-10-07
NPP Shell Extention dll is locked in explorer.exe


SinghRajenM:
### Description of the Issue

In a scenario where NPP shell extension is already triggered, "NppShell_06.dll" is not removed during uninstallation.
Npp shell extension is tightly attached to the explorer.exe, which is not released during the uninstallation.
### Steps to Reproduce the Issue

1. Install Npp with Shell Extension
2. Right click on any file (Edit with Notepad++ must be visible on Context menu)
3. Try to uninstall now.
4. Observe the install directory
### Expected Behavior

Install directory (C:\Program Files (x86)\Notepad++) should be removed completely.
### Actual Behavior

Install directory (C:\Program Files (x86)\Notepad++) is not removed as "NppShell_06.dll" is not removed.
### Debug Information

Notepad++ v7   (32-bit)
Build time : Sep 21 2016 - 22:00:56
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

This is happening for all most all the OS (checked on win7, Win Server 2016, Win10) for the versions (x86 and x64).

![npp_shell](https://cloud.githubusercontent.com/assets/14791461/19205278/06be8ea6-8cff-11e6-8f0b-83e3c5752d4a.png)

donho:
Current solution to deal with explorer lock is not perfect, but it works. So I won't do any codding effort for this issue.

SinghRajenM:
Okay, no problem!

But let me put the issue in two different ways. 

### 1. Update Notepad++
Consider N++ is released with new shell dll say **NppShell_07.dll** (contains fix for some major/critical issue) and new version is available via N++ auto updater as well.

1. A user has older version of N++ say **7.1** and he/she opens a text file using context menu
2. Now user will get a popup to update N++ from auto updater
3. User proceeds to update the N++

Now the issue comes, however the new update will be installed, but new shell dll will not be replaced.

### 2. User uninstalls Notepad++
1. User uninstalls N++ and somehow he/she notices that in N++ directory is not deleted from program files 
2. User tries to delete the directory manually, but it will not be deleted 
3. To delete this directory user has to either restart the explorer (if user is aware about it) or restart the system.

**So for both the scenarios, how you are going to deal? Is there any plan handle it?**

[...]

PR #2424, opened 2016-10-14
Delete Shell Extension dll more effectively

SinghRajenM:
Delete Npp Shell Extension dll more effectively
Fix issue #2404 
As soon Npp shell dll is loaded (right click on any file). It is locked inside the explorer.exe. Therefore, uninstaller/installer could not delete/replace the npp shell dll.
**Concept to resolve the issue is: move the locked dll to temp dir and delete that file on next reboot.**

PR #4772, opened 2018-08-18
Delete Shell Extension dll more effectively

SinghRajenM:
This is improved version of PR #2424. (instead of adding 1,2,3 so on, append current timestamp while moving `ShellDll` to %temp% so that nsis API `rename` does not fail.)

Fix issue #2404
As soon Npp shell dll is loaded (right click on any file). It is locked inside the explorer.exe. Therefore, uninstaller/installer could not delete the npp shell dll.

**Concept to resolve the issue is: move the locked dll to temp dir and delete that file on next reboot.**


Faq:
is this PR even valid as linked issue is closed?

SinghRajenM:
@Faq yes, still it is valid.

SinghRajenM:
No response from long time. So closing the pull request.

PR #4776, opened 2018-08-19
Make the MinGW/GCC build work again

marcows:
- updated the MinGW specific build files
- fixed build errors
- created stubs or excluded code parts for non-supported functionality

NP++ can now be built without errors and executed again.
Tested under Windows 7 using GCC 8.1.0 with POSIX threads and dwarf.


marcows:
Yes, I had seen them. It was hard getting patches integrated upstream in the past and I just wanted to try one last time with my existing work. I don't plan continuing, the time is better spent elsewhere. As long as the GCC build is only second-class citizen, you will always be fixing after and need luck getting them applied.

donho:
Duplicated:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4145

Issue #4655, opened 2018-07-10
Notepad++ is corrupting files and opening them with null values.


nightshaddow13:
### Description of the Issue
Occasionally when my computer goes to sleep and I have a '.pl' file open in notepad++ the file will become corrupt.  This has happened too me two times in the last 2 days.  The text seems to be completely deleted.

### Steps to Reproduce the Issue
1. Have open perl script
2. Put computer to sleep
3. Wake up device and open the document

### Expected Behavior
I expected my data to remain unscathed.

### Actual Behavior
The data was replaced with null data

### Debug Information
Notepad++ v7.5.7   (64-bit)
Build time : Jun 29 2018 - 00:53:58
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll 

![image](https://user-images.githubusercontent.com/41086688/42536703-c7bbb80e-8460-11e8-96f6-717bc9db7838.png)

sasumner:
Duplicate of #2381 and perhaps others.  Also see https://notepad-plus-plus.org/community/topic/15748/faq-desk-crash-caused-my-file-to-be-all-nulls

Long-standing, known issue...sad.  The #1 thing a text editor absolutely has to do is to not corrupt and not lose data.

sasumner:
@nightshaddow13 , has this problem recurred for you?  We're looking for a reliable reproduction sandbox.

pryrt:
@nightshaddow13, you seem to be the first I've seen here or in the [Community Forums](https://notepad-plus-plus.org/community/) that claims a reliable / repeatable method for causing the NULs (most people have it happen once, and never see it again).  How willing are you to help with the debug?

I don't have a build environment that would work for NPP, but there are plenty here who do.  I wonder if someone could create a debug-build executable for you, which would log extra debug info to a log file, flushing often: dump of all the settings (in case it's snapshot on/off, remember session on/off, etc), as well as recording various internal function calls -- if developers had a trace of the last few functions (or the last function that had enough time to flush the logfile to disk) that ran before the crash event, it might go a long way toward finding (and solving) the problem.

nightshaddow13:
[...]

Issue #4346, opened 2018-03-26
Complete loss of opened file content after a crash


pdelorenzo974:
### Description of the Issue
When Notepad++ crashes, every file with a RED TAB (modified & not saved state) is losing ALL it's content! When you restart Notepad++ it automatically reopen it completely empty!!!
I'm using latest version but this behavior exist for a long time and it's getting REALLY troublesome!
Do you really know what is it to lose a lot of work and not beeing able to recover it????
For now I just kept using Notepad++ because it doesn't crash OFTEN but now when it does it's a NIGTHMARE!!!

### Steps to Reproduce the Issue
1.  Open a file inside Notepad++
2.  Modify the file
3.  DO NOT save the file
4. Crash Notepad++
5. Re-start Notepad++
6. See for yourself!

### Expected Behavior
THE BEST EXPECTED BEHAVIOR: do not lose anything EVEN MODIFIED FILES AND FILE STATUS
THE MINIMUM EXPECTED BEHAVIOR YET TO BE: ALL SOFTWARE ARE ABLE TO AT LEAST RECOVER THE LATEST CONTENT SAVED ON DISK BUT THIS IS NOT EVEN TRUE WITH YOUR NOTEPAD++ ALL CONTENT IS WIPED OUT!!!!

### Actual Behavior
Each and every opened files with RED TAB (modified not yet saved) COMPLETELY WIPED OUT!!!!
EMPTY!!!! NOT MORE CONTENT!!!!!! COMPLETELY BLANK!!!!!!!

### Debug Information
Notepad++ v7.5.6   (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (32-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll NppTextFX.dll PluginManager.dll

stantonium:
This happened to me with the v7.5.6 64 bit version, could not get the debug info.  Unacceptable risk. Deleting npp++ until this bug is fixed.

stantonium:
duplicate of issue 2883?

Issue #2883, opened 2017-02-08
Notepad++ crash nulled the file


Anapher:
Hello,
after saving my php file, Notepad++ crashed and cleared the file to a size of 0 bytes. I wasn't able to find any cache or backup for that file. I'm a developer, I save my code frequently, I never thought that an issue could exist in Notepad++ which just erases the whole file on my HDD without any backup... 2 hours of work were destroyed there.

Anyways, that happend on Ctrl+S after I inserted XAML code into an HTML code tag like that:
```html
				<code class="xaml">
				<UserControl x:Class="ExtensionSpoofer.UserControl2"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
             xmlns:local="clr-namespace:ExtensionSpoofer"
             mc:Ignorable="d" 
             d:DesignHeight="300" d:DesignWidth="300">
    <Grid>
            
    </Grid>
</UserControl></code>
```

Two entries are found in the event log:
```
Name der fehlerhaften Anwendung: notepad++.exe, Version: 7.3.1.0, Zeitstempel: 0x587d55bc
Name des fehlerhaften Moduls: notepad++.exe, Version: 7.3.1.0, Zeitstempel: 0x587d55bc
Ausnahmecode: 0xc0000409
Fehleroffset: 0x000f4fdc
ID des fehlerhaften Prozesses: 0x4368
Startzeit der fehlerhaften Anwendung: 0x01d281f49d0c0806
Pfad der fehlerhaften Anwendung: C:\Program Files (x86)\Notepad++\notepad++.exe
Pfad des fehlerhaften Moduls: C:\Program Files (x86)\Notepad++\notepad++.exe
Berichtskennung: 6306db71-e7b5-44fb-9ff9-3a05982bb2fd
Vollständiger Name des fehlerhaften Pakets: 
Anwendungs-ID, die relativ zum fehlerhaften Paket ist: 
```

```
Fehlerbucket , Typ 0
Ereignisname: BEX
Antwort: Nicht verfügbar
CAB-Datei-ID: 0

Problemsignatur:
P1: notepad++.exe
P2: 7.3.1.0
P3: 587d55bc
P4: notepad++.exe
[...]

Issue #2381, opened 2016-10-02
Original (already saved) file and configuration settings lost after loss of power


jld123:
### Description of the Issue

File was open in editor. Put notebook in standby and left. Battery ran out of power and notebook went off. After my return I started up the notebook again. Notepad++ lost its configuration settings. But even more importanly, the file that was open in the editor only contained NUL NUL NUL .... Not only the differences to the last save point were lost but the entire file! This has been an issue for years (google notepad++ nul nul nul). 
### Steps to Reproduce the Issue

Tried to reproduce the issue by forcefully restarting shutting off the notebook (removed battery and pulled the plug), but that did not work. It does not seem to happen every time or only under certain circumstances

Sorry that I can't provide more information on how reproduce this bug. However, it seems to have been an issue for years. And the bug is of such a high severity (the original saved file should NEVER be destroyed) that I still decided to report this.

marcusfr123:
Even if this bug is not fixed it would be helpful to have infos under which circumstances this happens, what exactly happens (e.g. are all files open in the editor affected or only the one being edited or only changed files or...) and how to mitigate the problem (e.g. are notepad++ simple/verbose/session backup files also affected or can the original file be restored through those?)

SD-DAken:
The problem still exists in Notepad++ 7.5.1

The whole computer was hung (probably not Notepad++ 's fault) and had to be forcefully turned off, after restart the whole file that was currently open in Notepad++ was overwritten with NUL (0x00 bytes).

The file size seemed to be correct but every byte was replaced by 0x00 (NUL).

Even though no exact steps to reproduce exist the issue is severe enough that the code that saves files should be checked if there are hints what could result in such catastrophic failure.

I could understand if the unsaved changes are lost or the file ends up in an inconstent state if the programm / pc crashes, but the whole file being replaced by null bytes is unexpected and the worst that could possibly happen.

pryrt:
It's come up yet again in the [forums](https://notepad-plus-plus.org/community/topic/13302/fix-corrupted-txt-file-null/17).  

At one point last month, I tried using the windows API from an external process to try to kill the Notepad++ instance, and see if I could reliably replicate the issue, but I couldn't get it to trigger; apparently, even killing the process externally is too "clean" of an exit.

Today, I began to think that maybe there was a way to force a Windows Blue Screen Of Death while running NPP, and see if it could replicate the problem -- my thought was in a Virtual Machine, to protect any actual hardware.  

[...]

PR #4803, opened 2018-08-29
Fix possible file corruption during backup or power loss or other abn…

### LINKED ISSUE #240: Find/Replace On Large File Causes Crash
When I try to run a regular-expression find/replace operation on a large file, Notepad++ crashes.
- I don't think it's the fault of the regex. It's relatively simple - no backreferences, no nested repeats, and no repeatable sections which can match a newline (so it can't go mad and try to match the entire file)
- The thing I am replacing with is the blank string, so I'm pretty sure that's not the problem.
- It is definitely correlated with the size of the input to the find/replace; if I select a small enough part of the file and use 'In Selection', no crash:
  - Selection: 422932 characters, 8816 lines. Result: 1759 matches found and replaced, no crash
  - Full file: 1555727characters, 32212 lines. Result: 6433 matches found and replaced, but Notepad++ crashes as soon as the find/replace operation finishes.
- The crash appears to take place after the find/replace operation has finished - the 'X occurrences replaced' status on the Replace dialogue updates, and in the background I can see the file being re-rendered - but before control is actually returned to me after I have clicked the 'Replace All' button
- Error messages are as follows:
  - Win32Exception: "An exception occurred. Notepad++ cannot recover and must be shut down.
    The exception details are as follows:
    Code: 0xC0000005
    Type: Access violation
    Exception address: 0x5DCB0F3A
  - Do you want to save a dump file?
    Doing so can aid in developing
    [Yes] [No](I select yes.)
  - Failed to create dump file 'C:\Program Files (x86)\Notepad++\NppDump.dmp' (error 5)
  - Notepad++ will attempt to save any unsaved data. However, dataloss is very likely.
  - Notepad++ was able to successfully recover some unsaved documents, or nothing to be saved could be found.
    You can find the results at:
    C:\Users[redacted]\AppData\Local\Temp\N++RECOV
  - Notepad++: a free (GNU source code editor has stopped working
    A problem has caused the program to stop working correctly.
    Windows will close the program and notify you if a solution is available.
  - Sometimes an additional dialogue complaining about a 'pure virtual function call' appears, but I haven't been able to reproduce that reliably so I don't have the details.
- It does not appear to be correlated with memory usage; I watched the task manager during the find/replace and crash, and notepad++.exe's memory usage remained level throughout
- It appears to be correlated with the number (or possibly total size) of matches; I changed my regex by one character so it would require the same search pattern but fail to find any matches, and I got no crash.

pnedev:
…ormal N++ termination

Fixes #240, #2381, #2883, #4346, #4655 and probably more issues related to loss of data.


pnedev:
Due to the synchronization issues with backup operation I've made the backup run in the main N++ thread (the UI one) by posting in its message queue. This should cause any current long running user operation to first finish and then execute the backup save.

The problems so far were caused by the fact that some complex user operations were not considered _'long-running'_ (but actually are on very big files) and were not protected by the `LongRunningOperation` lock.

Strictly speaking, the `LongRunningOperation` lock during backup is not needed anymore because now all Scintilla buffer operations are serialized via channeling them through a single thread but I have left them in case I have overlooked something.

Barbapoux:
Hi, just to confirm case #240 with last hypothesis of @ClaudiaFrank 
when instant backup period is greater than duration for replacement, the crash doesn't occur anymore.
Thanks

donho:
@Barbapoux Can you reproduce steadily this issue?

Barbapoux:
Oui tout à fait. Je pense que tout le monde peut le faire.
Si ta question est "Est-ce que tu peux tester la version corrigée?", j'ai de nouveaux fichiers (125 Mo), donc oui c'est possible. Il faut juste m'indiquer la procédure souhaitée.
Cordialement.

donho:
@Barbapoux 
Cool ! Pourrais-tu me donner les instructions pour que je puisse le reproduire stp?

donho:
@pnedev Could you solve the conflicts please?

pnedev:
@donho Done.

Barbapoux:
Pardon, tout est décrit dans le thread #240 (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/240#issuecomment-200867146) , il y a un fichier pour le test, il suffit de l'éditer et faire supprimer les blancs de fin de ligne (ou toute opération un peu longue de remplacement, qui va mettre le bazar dans les pointeurs).
Avec la sauvegarde instantané à 7 secondes par défaut, en général, ça va au tas.

donho:
@Barbapoux  This PR is about fixing possible file corruption during backup. 
Whereas your comment is about Crash for Find/Replace on Large File.
I don't see the relation between these 2 issues.

Barbapoux:
I think @pnedev is on the right way. I had some issues with long replacement operations (replace all xxx xxx xxx by nothing on big file) after version 6.5.5, without any plugin, and I have found that issue with trailing blank that does the same error and easily reproductible with dummy data. On my point of view, the reason was the same, it seems you have solved replacement crash ( it was 2 years ago), as I don't have anymore garbage in my file with current version (replacement generated crashs or memory dump in middle of text file).
As @ClaudiaFrank has seen, disabling the periodic backup is solving the issue on trailing blank, then it is certainly because it is trying to copy a memory block that has changed.

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -957,47 +922,8 @@ bool FileManager::backupCurrentBuffer()
...
-			printStr(TEXT("WaitForSingleObject problem in deleteBufferBackup()!"));
-			return false;
-		}
-
-		// unlocled here, set to non-signaled state, to block all threads
-		::ResetEvent(writeEvent);
-	}
-
-	EventReset reset(writeEvent); // Will reset event in destructor.
-
```
donho: No `LongRunningOperation op;` instruction here?

pnedev: You mean in `FileManager::deleteBufferBackup()` ?
It won't hurt to add `LongRunningOperation op` but strictly speaking it is not needed at all. All buffer operations are now channeled through a single thread so no need for locking at all.
I can either remove all `LongRunningOperation` locks from `FileManager` in this file or add a lock in `deleteBufferBackup()` for consistency. What do you prefer?

donho: > All buffer operations are now channeled through a single thread

Could you point me this part?

pnedev: Yes, at the very beginning of this Git diff I have changed the direct call to `MainFileManager->backupCurrentBuffer()` to `::PostMessage(Notepad_plus_Window::gNppHWND, NPPM_INTERNAL_SAVEBACKUP, 0, 0);`.
This way the actual buffer backup will go through the message queue which is serialized - no concurrency.
The other operations here were not synchronized (locked) anyway but they also are always triggered from the GUI thread (again through the message queue). Am I missing anything?

donho: Well played!

Issue #4793, opened 2018-08-24
Feature request: Add color and icons for start screen tiles in windows 10


popcorner:
When I added some software to my start screen...
![image](https://user-images.githubusercontent.com/6312791/44568567-f8c87b00-a7a9-11e8-965c-7c850f0b2648.png)
It seems that other software have their own background color of the tiles, but notepad++ doesn't.
I checked the folder of chrome.exe and found chrome.VisualElementsManifest.xml:
```
<Application xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>
  <VisualElements
      ShowNameOnSquare150x150Logo='on'
      Square150x150Logo='68.0.3440.106\VisualElements\LogoLight.png'
      Square70x70Logo='68.0.3440.106\VisualElements\SmallLogoLight.png'
      Square44x44Logo='68.0.3440.106\VisualElements\SmallLogoLight.png'
      ForegroundText='dark'
      BackgroundColor='#FFFFFF'/>
</Application>
```
Just put a [softwarename].VisualElementsManifest.xml file with the program and the tile will get its color, it may be a good idea if notepad++ can add one too.

PR #4822, opened 2018-09-04
Add color for start screen tiles in windows 10

popcorner:
Just an example for this issue #4793 .
I don't know C++ or nsis.
I made this by imitating the file change.log and I'm not sure if it works properly.

### DIFF
--- a/PowerEditor/Test/FunctionList/nsis/unitTest
+++ b/PowerEditor/Test/FunctionList/nsis/unitTest
@@ -522,6 +522,7 @@ Section -"Notepad++" mainSection
 	File "..\..\LICENSE"
 	File "..\bin\SciLexer.dll"
 	File "..\bin\change.log"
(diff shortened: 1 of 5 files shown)


SinghRajenM:
It looks good, but I feel it would be great if both 70x70 and 150x150 images are also used. Have a look at Firefox icon.

image

popcorner:
@SinghRajenM The current icon for Notepad++ is not suitable if making it to PNG and scale by the operating system. If you make the PNG in 70px and 150px, the users who have high resolution screen(150% and more) will see a blurred icon. If you try to make a bigger PNG, the icon on the normal screen(100%) will be unclear. This icon doesn't design to be suitable for zooming.

And the current icon can fit to any resolution automatically, and seems not bad. You can see this picture, most of the icon doesn't as big as the one in Firefox.
image

PR #4934, opened 2018-10-18
Update swedish.xml

IluvBread:
Change "konfigurera programspråk" at line 278 to better represent what the menu actually is for.

### DIFF
--- a/PowerEditor/installer/nativeLang/swedish.xml
+++ b/PowerEditor/installer/nativeLang/swedish.xml
@@ -275,7 +275,7 @@
 					<Item id="10003" name="Flytta till ny instans"/>
 					<Item id="10004" name="Klona till ny instans"/>
 
-					<Item id="46001" name="Konfigurera programspråk"/>
+					<Item id="46001" name="Stilkonfiguration"/>
 					<Item id="46250" name="Definiera ditt språk..."/>
 					<Item id="46180" name="Användardefinierad"/>
 					<Item id="47000" name="Om Notepad++"/>


donho:
@magol  Could you validate it as well?

md2perpe:
Considering that the English wording is "Style Configurator..." the translation "Stilkonfiguration" would be more straightforward than "konfigurera programspråk" (meaning "configure programming language"). The Danish, Norwegian (both *bokmål* and *nynorsk*) and German translations also talk about style or look (*udseende* in Danish).

magol:
I agree that the current translation is not the best 
"Stilkonfiguration" is a literal translation, but I think the word is to long.
Maybe translate it to "konfiguration av stil" or "konfiguration av utseende"?

md2perpe:
I think that "konfigurera stil" would be best. Words like "av" has little content, are distracting, and should be avoided.

IluvBread:
"Konfigurera stil" would make the user have to read it all to understand what it configures while stilkonfiguration accomplishes the same thing with just a quick glance.

It's a little strange to prioritise the verb instead of the subject since we're literally inside the "settings" submenu.. I don't think anyone going into settings menu expects to do anything but change settings, so why would "Konfigurera stil" be more appropriate than "Stilkonfiguration" ?

magol:
@IluvBread You have a point with your objection. But I do not agree that "Stilkonfiguration" is easier to read than "Konfigurera stil". At a quick lock, I would not understand what "Stilkonfiguration" mean, I have to read the whole word to understand what it's all about.

But I have no strong opinions about it really, if most people thinks that "Stilkonfiguration" is the best, I accept that :-)

donho:
@magol I will close this issue and let you decide the best translation of this item for the next commit.
Thank you.

m4gnify:
Any new thoughts about this? "Konfigurera stil" is a lot better than the **still** current "Konfigurera programspråk". "Konfigurera programspråk" says nothing about what it actually does.

Issue #4679, opened 2018-07-19
CTRL+F4 does not close tab


nivs1978:
### Description of the Issue
When I press CTRL+F4 the currently open tab does not close.

### Steps to Reproduce the Issue
Open a few files and press CTRL+F4 to close the active tab/file.

### Expected Behavior
The currently active tab should close.

### Actual Behavior
Nothing happens

### Debug Information
Almost all Windows applications support the CTRL+F4 to close the current document/tab. Web browsers, Microsoft Word/Excel, OpenOffice, Adobe Reader etc. all support this shortcut that is just as generic as CTRL+C and CTRL+V in WIndows applications. Notepad++ should support this also.
It is on the official Windows Shortcut list: https://support.microsoft.com/en-us/help/12445/windows-keyboard-shortcuts

sasumner:
Dude, change it in the shortcut mapper...to Ctrl+F4...or whatever you like.

jrheisey:
As this is a Windows application I request that the default behavior default to the Windows standard of CTRL-F4 to close a tabbed view. Note that CTRL-F4 has been the Windows standard since Tabbed Views were implemented in Windows about 20 years ago.

sasumner:
@jrheisey Sure, but if it isn't acted upon, don't let it ruin your life if you want it -- remap it via the Shortcut Mapper...I did.  :)

MetaChuh:
@nivs1978 

please use ctrl+w to close the current tab.

thank you for your help and best regards.

IanKemp:
You can get this working using the NppExec plugin:

1. Go to Plugins -> NppExec -> Execute...
2. Paste the following snippet into the textbox:

```
NPP_SENDMSG WM_COMMAND IDM_FILE_CLOSE
```

3. Click `Save...` and name the script something like "Close current file".
4. Go to Plugins -> NppExec -> Advanced Options...
5. Under `Menu item` in the bottom-left, type something like "Close current file" into the `Item name` textbox.
6. From the `Associated script` dropdown, select the "Close current file" script, then click `Add/Modify`.
7. Click OK and the dialog will prompt you to restart Notepad++, do so.
8. Once NPP has restarted, go to Settings -> Shortcut Mapper... and select the `Plugin commands` tab.
9. Scroll down to the bottom and the new `Close current file` command should be there. Select its row and double-click, or click `Modify` to bring up the key binding dialog. Select CTRL and F4 as your shortcut, click OK.

[...]

PR #4941, opened 2018-10-20
CTRL+F4 does not close tab #4679

harish-515:
Made changes for the issue #4679. Where in user was unable to close the tab windows with Ctrl + F4 shortcut.


CookiePLMonster:
I disagree with this being a replacement. Ctrl+F4 may be a standard, but Ctrl+W is used just as widely. We shouldn't sacrifice one for the other.

sasumner:
At the risk of repeating myself, forget all this nonsense.  It has been defaulted to ctrl+w in Notepad++ for a long time, perhaps forever.  For those that don't like that fact, Shortcut Mapper gives you the ability to change it.  Utilize that feature.  Make it whatever you want...hint for the person that started this:  Make it ctrl+f4 (for you).

donho:
A better solution is both. But since we cannot do both so far, replace the default one is just a lousy solution.

Issue #4942, opened 2018-10-21
Read-Only flag (user read only) is not synced


SinghRajenM:
### Description of the Issue

User read only flag under "tab context menu" and "edit menu" is not synced in different cases.

### Steps to Reproduce the Issue

**Case 1:**
If File system attribute is read only. Then "User read only" is disabled under "tab context" menu while _**same is enabled under "edit menu"**_.

It should be disabled at both the places.

**Case 2:**
If File system attribute is read only and user enables "tail monitoring". Then "User read only" is checked and disabled under "tab context" menu while same is disabled **_but checked_** under "edit menu".

It should be disabled and checked at both the places.

**Case 3:**
File system attribute is normal and user enables "User read only" (via either tab context menu or edit menu). "User read only" flag is checked and enabled at the places which is absolutely fine.
Now set file system attribute as set read only,  then "User read only" is checked and disabled under "tab context" menu while same is **_enabled and unchecked_** under "edit menu".

It should be disabled and checked at both the places.

### Debug Information

Notepad++ v7.5.9   (64-bit)
Build time : Oct 14 2018 - 15:19:55
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

PR #4943, opened 2018-10-21
Fixed Read-Only flag (user read only) syncing issue

SinghRajenM:
Fixed issue #4942

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1965,19 +1965,13 @@ void Notepad_plus::checkDocState()
 	enableCommand(IDM_VIEW_LOAD_IN_NEW_INSTANCE, !(isCurrentDirty || isCurrentUntitled), MENU);
 
 	bool isSysReadOnly = curBuf->getFileReadOnly();
-	if (isSysReadOnly)
-	{
-		::CheckMenuItem(_mainMenuHandle, IDM_EDIT_SETREADONLY, MF_BYCOMMAND | MF_UNCHECKED);
-		enableCommand(IDM_EDIT_SETREADONLY, false, MENU);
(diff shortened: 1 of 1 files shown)


SinghRajenM:
@donho Could you please review this PR?

donho:
@SinghRajenM Thank you for the fix.
I just found another glitch: when the file is set RO via Notepad++, then set this file system RO outside of Notepad++, "Set RO" command is not disabled. Could it be enhanced with the minimum effort?

SinghRajenM:
>I just found another glitch: when the file is set RO via Notepad++, then set this file system RO outside of Notepad++, "Set RO" command is not disabled. Could it be enhanced with the minimum effort?

@donho , I will check it soon. Thank you!

SinghRajenM:
@donho I could not reproduce this issue with latest code.

However, figured out an MS bug (mostly due to cache). 😜 
1. Set a file RO in notepad++
2. Now set system read only file outside of npp (right click on file-> properties->check Read-Only). **Just click on apply, DON'T click OK that makes sure that properties dialog remains open**
3. Now clear system read only flag using from Npp using command (clear read-Only flag).
4. Now come back to file properties dialog. It will indicate that file is still read only (but in actual not, as we removed that flag using npp). Anyway, now uncheck and check again. Now click on OK/Apply and close the properties dialog.
5. Now open properties dialog again, **You will notice that still file is not read only, but we tried to make it read only in step 4.**

I feel, you might have encounter this scenario. If not let me know.

MSBug

donho:
> I just found another glitch: when the file is set RO via Notepad++, then set this file system RO outside of Notepad++, "Set RO" command is not disabled. Could it be enhanced with the minimum effort?

Never mind, I guess it's due to the File Status autto-detection was disabled.
Thank you for your investigation.

PR #4944, opened 2018-10-21
Update chinese.xml

david082321:
EN:
Change "$STR_REPLACE" into "$STR_REPLACE$". To fix the problem that file name can't display.

ZH:
把 $STR_REPLACE改成 $STR_REPLACE$，修正檔名無法正常顯示的問題。


david082321:
@donho Please fix tihs bug in Chinese language.
default

PR #4947, opened 2018-10-23
[WIP] Installer is DPI-aware

stanciuadrian:
And no longer blurs text.

Before: 

before

and after: 

after

I noticed the image is smaller only after inserting these 2 screenshots. [Elsewhere](http://forums.winamp.com/showthread.php?t=443754) somebody suggests 

> showing different bitmaps according to the DPI setting


donho:
@stanciuadrian Could you solve the conflict ?

stanciuadrian:
Done.

PR #4962, opened 2018-10-25
Avoiding loading / verifying SciLexer.dll in a static initializer.

### LINKED ISSUE #4952: Problematic call of loadSciLexerDll in a static initializer
The following line

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ec340000ccde72debc8d0ac721ec376d07dec3e4/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp#L50

is potentially dangerous, but the opinion of a C++ expert is required. The logic behind the method is quite complex and uses the standard library, including streams. As far as I understand, the order of static initialization in C++ is specified only within the same module so the standard library might not be ready yet. This could have serious implications and I recommend the call not to be made in the static initializer, but for example in `ScintillaEditView::init`.

MIvanchev:
Tested with `Unicode Debug|x64` and `Unicode Release|x64`. With the proposed change, SciLexer.dll will be loaded and verified on WM_CREATE, during the initialization the Scintilla views. The loading only takes place once; if it fails the program exists. Nothing really changes in the error handling timing. Fixes #4952.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -47,7 +47,7 @@ TCHAR * getSciLexerFullPathName(TCHAR * moduleFileName, size_t len)
 	return moduleFileName;
 };
 
-HINSTANCE ScintillaEditView::_hLib = loadSciLexerDll();
+HINSTANCE ScintillaEditView::_hLib = NULL;
 int ScintillaEditView::_refCount = 0;
 UserDefineDialog ScintillaEditView::_userDefineDlg;
 
@@ -225,7 +225,11 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 {
 	if (!_hLib)
 	{
-		throw std::runtime_error("ScintillaEditView::init : SCINTILLA ERROR - Can not load the dynamic library");
+		_hLib = loadSciLexerDll();
+		if (!_hLib)
+		{
+			throw std::runtime_error("ScintillaEditView::init : SCINTILLA ERROR - Can not load the dynamic library");
+		}
 	}
 
 	Window::init(hInst, hPere);


PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -225,7 +225,11 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 {
 	if (!_hLib)
 	{
-		throw std::runtime_error("ScintillaEditView::init : SCINTILLA ERROR - Can not load the dynamic library");
+		_hLib = loadSciLexerDll();
```
CookiePLMonster: Is the function called only once? If so, the first `if (!_hLib)` can probably go.

MIvanchev: Sadly it's called at least twice:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b01a3966681cacdfb1fce8313822c234566d98f3/PowerEditor/src/Notepad_plus.cpp#L222

and also for each plugin requested Scintilla:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b01a3966681cacdfb1fce8313822c234566d98f3/PowerEditor/src/ScitillaComponent/ScintillaCtrls.cpp#L36

It's what makes refactoring the code difficult :/

MIvanchev:
> Would threading be an issue here? ...

Very good question, didn't think about threading. My analysis shows it won't be an issue, but please correct me if I'm wrong. The primary and secondary views are initialized upon WM_CREATE, this is safe. The remaining question is whether createSintilla is called from multiple threads. I answer this with no, since the call hierarchy is

```
<GUI thread>
Notepad_plus_Window::Notepad_plus_Proc
   Notepad_plus_Window::runProc
      Notepad_plus::process (on NPPM_CREATESCINTILLAHANDLE)
          ScintillaCtrls::createSintilla
```

So this should be safe...

CookiePLMonster:
You are right, it should be safe to assume `WM_CREATE` won't be called from more than one thread at a time.

donho:
Such fix / modification should address to bugs.
If yes, please provide your instructions of bug reproduction and reopen it.

PR #4967, opened 2018-10-27
Improved deleteMarkedLines

RipleyTom:
PR aims to minimize the number of calls to replaceTarget which is a very expensive operation by removing consecutive bookmarked/unbookmarked lines at the same time.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2175,7 +2175,26 @@ void Notepad_plus::deleteMarkedLines(bool isMarked)
 	for (int i = lastLine ; i >= 0 ; i--)
 	{
 		if (bookmarkPresent(i) == isMarked)
-			deleteMarkedline(i);
+		{
+			int last_bookmark = i;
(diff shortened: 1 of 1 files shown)


sasumner:
Does this come from an "issue"?  If so, shouldn't the issue be linked here?

RipleyTom:
It comes from me using notepad++ to clean a log and noticing it took 20+ mins on a  500MB log to remove bookmarked lines(and it took less than a min to bookmark them).

After a quick search I couldn't find a relevant issue.

I've noticed drastic improvement when cleaning my logs with this pr(they often have consecutive bookmarks) but even in worst case scenarios(no consecutive bookmarks) the performance loss in the extra check is insignificant.

Ideally for the best performance you'd recreate the text from scratch by concatening all non removed lines but I've no idea if it is possible through scintilla.

sasumner:
So in summary this is designed to speed up the case where there are a lot of consecutive lines that are bookmarked...

Any reason why you didn't change the similar code in `cutMarkedLines()` in a similar way, or better, have a common section to do this?

PR #4977, opened 2018-10-31
AutoIt autoCompletion

chcg:
added autoit.xml autoCompletion from http://docs.notepad-plus-plus.org/index.php?title=User_Defined_Language_Files#A -> http://notepad-plus.sourceforge.net/commun/userDefinedLang/Autoit.xml


PowerEditor/installer/APIs/autoit.xml
```diff
@@ -0,0 +1,16099 @@
+<?xml version="1.0" encoding="Windows-1252" ?>
```
chcg: Is there are reason not to use utf-8 as encoding?

PowerEditor/installer/APIs/autoit.xml
```diff
@@ -0,0 +1,16099 @@
...
+<NotepadPlus>
+	<AutoComplete>
```
chcg: Should language be added?

PowerEditor/installer/APIs/autoit.xml
donho: IMO, most of API are in English, so ANSI is OK

PowerEditor/installer/APIs/autoit.xml
donho: which language?

donho:
@chcg Could you solve the conflict please?

PowerEditor/installer/APIs/autoit.xml
chcg: Other files have e.g.:
`
	<AutoComplete language="C">

`

Is there a specific reason why enoding is windows-1252 and not utf-8?

chcg:
Rebasing done.

PR #4978, opened 2018-10-31
Lua auto completion

chcg:
added lua autoCompletion from http://ageekandhisblog.com/notepad-plugin-lua-corona-api-autocompletion/

as the one mentioned under http://docs.notepad-plus-plus.org/index.php?title=User_Defined_Language_Files#L -> http://www.davidcarrigg.com/uploads/LuaLangFile.zip is no longer available


dail8859:
It seems like alot of the standard functions are there just glancing through. Honestly I'm not sure if its a good idea to include the Corona API since it is a specific framework.

Issue #4946, opened 2018-10-22
Opening D:/*.log leads to a crash


barisdemiray:
### Description of the Issue
I wanted to copy and paste a path given by a Visual Studio command, to open all the `.log` files and it led to a crash. Path I entered was this, `D:/*.log`.

It happens with version 7.5.8 (32-bit). With French interface if that matters.

### Steps to Reproduce the Issue
1. Ctrl+O or File/Open.
2. Paste `D:/*.log` in the file name input box.
3. Click Open.

### Expected Behavior
Opening all the `*.log` files in `D:\`.

### Actual Behavior
Crash.

### Debug Information
I saved a crash dump, [NppDump.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/2501000/NppDump.zip).

PS: I got a `Cannot save crash dump (error: 5)` message while trying to save a crash dump. A more clear message like `Cannot save a crash dump under Program Files without Administrator rights` would make more people to send a bug report.

SinghRajenM:
Don't you think `D:/*.log` should be `D:\*.log` (notice slash). Moreover, npp does not supports wildcard for file opening.

SinghRajenM:
@jsnells1 really?
If I am not wrong, it will just open the folder path, then manually you need to select the file(s) and click ok or again you have type *.* or some *.log or similar.

for example <kbd>Ctrl</kbd>+ <kbd>O</kbd> then type `C:\Users\Xyz\Desktop\Garbage\*.txt` will open path till `C:\Users\Xyz\Desktop\Garbage`, then again either you have to select all files or type `*.txt` again, right?

jsnells1:
Oh, sorry you're right. It navigates to the directory and then you can do *.txt once there. I thought you meant all wildcard file opening wasn't supported so I chimed in.

sasumner:
This is fraught with difficulty, and most of it comes from the setting:  "Use new style dialog (without file extension feature and Unix style path capacity"...

barisdemiray:
> Don't you think `D:/*.log` should be `D:\*.log` (notice slash). Moreover, npp does not supports wildcard for file opening.

I agree but we have `Cmake` as the build system and we do slash replacement to avoid escaping. That path probably must be one of them. In any case, end result should not be a crash.

sasumner:
And shouldn't `D:/*.log` actually do something useful when "Unix style path capacity" is enabled?

[...]

PR #4986, opened 2018-11-04
Crash fix for invalid input path

SinghRajenM:
Fixed issue #4946 
root cause is below line
```const TCHAR * substring = wcsrchr(fileName.c_str(), TCHAR('\\'));```

`substring` coming `NULL` as per reported case.

**Minor code enhancement:**
File `FileDialog.cpp` uses invalid flag in it. However, inclusion or exclusion of flag `DS_CENTER` does not have any impact as long flag `OFN_PATHMUSTEXIST` exists. Because the value is same for both the flags.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -352,15 +352,18 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive, 
             if (globbing)
             {
                 const TCHAR * substring = wcsrchr(fileName.c_str(), TCHAR('\\'));
-                size_t pos = substring - fileName.c_str();
+				if (substring)
+				{
+					size_t pos = substring - fileName.c_str();
 
-                patterns.push_back(substring + 1);
-                generic_string dir(fileName.c_str(), pos + 1); // use char * to evoke:
-                                                               // string (const char* s, size_t n);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: PowerEditor/src/NppIO.cpp [hunk 1]
```diff
@@ -352,15 +352,18 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive,
             if (globbing)
             {
                 const TCHAR * substring = wcsrchr(fileName.c_str(), TCHAR('\\'));
-                size_t pos = substring - fileName.c_str();
+				if (substring)
```
SinghRajenM: If needed some message can be shown in `else` case of [line](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4986/files#diff-76c62ef652708b4b7b02f049a16b0629R391). Like "File is not available for opening" or something similar, but that is optional.

**But at any cost app should not crash, so this `null` checking is MUST.**

donho:
Cannot reproduce the crash at all. Is there anything I am missing?

SinghRajenM:
This issue occurs when "Use New Style Dialog" setting is not used .
<kbd>Ctrl</kbd>+<kbd>O</kbd>, then type `D:/*.log` and hit <kbd>Enter</kbd>.

Issue #4409, opened 2018-04-13
Read-only Flag Cleared Upon Close & Reopen


nathanm:
### Description of the Issue
I usually keep a text file open in Notepad++ with special characters that I use to paste into documents. Because I don't want to inadvertently change it, I have the read-only flag set, just in case. However, with the latest update to Notepad++, the read-only flag is cleared when I open Notepad++.

I tried setting the file's read-only attribute in Windows Explorer, while Notepad++ was closed. Then, after reopening Notepad++, the flag was still set. However, after clearing it and setting it again in Notepad++, the wrong behavior continued.

### Steps to Reproduce the Issue
1. Set the read-only flag on a file
2. Close Notepad++
3. Open Notepad++

### Expected Behavior
The read-only flag should remain set or cleared for each file, whatever its state was when last open.

### Actual Behavior
The read-only flag is cleared for a file that was previously set.

### Debug Information
Notepad++ v7.5.6   (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll GedcomLexer.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

sasumner:
It has worked this way as long as I've been using Notepad++.  It's meant to be a temporary thing.  But who knows...there is no documentation about it that I know of.

dinkumoil:
I guess you are mixing up two different mechanisms for write protecting a file. But the UI of Notepad++ is admittedly confusing in this point. 

There are two menu entries which tell us that they manipulate the file's write protection:

1. `Edit - Set Read-only`
2. `Edit - Clear Read-only flag`

The first one only toggles an internal flag of Notepad++ (or the underlying Scintilla component) whereas the second one clears the file system's "Read only" file attribute - and this is an one-way ticket because it's not possible to **set** this attribute from within Notepad++.

dinkumoil:
I think this can't be achieved in the way you would prefer. Imagine Notepad++ would store the state of its internal read-only flag for **all** files which it has been opened ever - the config file would grow to GBs over time.

nathanm:
I don't mean every individual file, just the ones that were open when Notepad++ was closed.

Issue #2970, opened 2017-02-28
[Feature Request / Bug Fix] Read-Only attribute not saved in session Xml file


salvarani:
### Description of the Issue ###
When saving a session where there are Read-Only documents, the Read-Only flag is not saved in the .XML session file; when restarting Notepad++ and loading the session, all documents are editable... now imagine that you have to re-apply the Read-Only flag to dozens of documents...

andry81:
Is a better idea to just create a command line key to request new session where directly edit read only files w/o need to clearing the flag and further save it at a session level.

To save a file, Notepad++ can, for example:
1. remove the flag
2. save file
3. restore flag

For an ability to differentiate between readonly not changed and read only changed file, there is a need to redesign the tabs view, because there is about 3 versions of an icon:
* read only file
* read/write not changed
* read/write changed

When needs 4:
* read only not changed
* read only changed
* read/write not changed
* read/write changed

Issue #4872, opened 2018-09-23
Feature Request: Remember read-only status of opened files


Lokiare:
### Description of the Issue

NPP fails to remember the manually set read-only status (set from the context menu for each tab/file), when re-opened.

### Steps to Reproduce the Issue

1. Open a file in NPP
2. Right-click on tab/file set read-only status
3. Close NPP
4. Re-open NPP

### Expected Behavior

It's expected that the tab/file opens and remembers whether it was set to read-only or not.

### Actual Behavior

The tab/file does not remember that it was set to read-only.

### Debug Information

N/A

dinkumoil:
Duplicate of #4409.

Issue #2548, opened 2016-11-08
Status of Read Only files is ignored on Session Save/Load operations


terablade2001:
### Description of the Issue
Status of Read-Only marked files seems not to be saved when a session is saved into a file. Thus the next time the client is load a session from a file, all Read-Only files will be editable again!

### Steps to Reproduce the Issue
1. Create a new file, add some text to it and save it somewhere.
2. Set Read-Only status: Edit->Set Read-Only. Now the file can not be edited.
3. Save the current session from File->Save Session... somewhere in your disk.
4. Close notepad++ and reopen it.
5. Load the previous saved session from File->Load Session... and select the session file which was used before.

### Expected Behavior
The file which was marked as Read-Only (step 2) should be still be marked as Read-Only and should not be able to be edited.

### Actual Behavior
The file which was marked as Read-Only (step 2), is now NOT marked as Read-Only and can be edited.

### Debug Information
Notepad++ v7.1   (32-bit)
Build time : Oct 15 2016 - 23:30:30
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7
Plugins : none

thenlich:
I agree that this behavior is unexpected. Same for "Monitoring (tail -f)" feature.
Can reproduce on:

Notepad++ v7.5.3   (64-bit)
Build time : Dec  5 2017 - 10:47:54
Path : C:\Daten\Programme\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

PR #4987, opened 2018-11-04
[Feature Request]: Retain read only flag state in session as well

SinghRajenM:
Implemented feature request.
Fixes: #4409 and #4872 

[Edit]:
Fixes: #2970, #2548

**Please note:**
This PR is subsequent of PR #4943 and contains all the changes from that PR as well. This PR does not work well if those changes are not included.

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -162,8 +162,8 @@ private:
 
 struct sessionFileInfo : public Position
 {
-	sessionFileInfo(const TCHAR *fn, const TCHAR *ln, int encoding, const Position& pos, const TCHAR *backupFilePath, FILETIME originalFileLastModifTimestamp, const MapPosition & mapPos) :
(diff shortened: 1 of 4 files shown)


SinghRajenM:
@donho Could you please review this PR?

PowerEditor/src/Parameters.h
```diff
@@ -177,6 +177,8 @@ struct sessionFileInfo : public Position
 	std::vector<size_t> _marks;
 	std::vector<size_t> _foldStates;
 	int	_encoding = -1;
+	bool _isUserReadOnly = false;
+	bool _isMonitoring = false;
```
donho: Why do we need this member variable?

SinghRajenM: @donho 

we need `_isMonitoring` because we can't rely only on `_isUserReadOnly` as this flag is also set when monitoring is ON.

So to save readonly for a file in session we need to check whether the file is really set as readonly or it is set readonly because of monitoring. This check is being done [here](
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4987/files#diff-35d82bdf4e1fdc62d667eb514cbc8cd2R3145)

PR #4989, opened 2018-11-05
Add comment on shortcut(installer)

Marr11317:
Add a little comment in the installer, since all major programs do it.
This is a minor pull request. Maybe it should me merged manually.


Marr11317:
Ok, why not?

Issue #4065, opened 2018-01-10
Change from 32 to 64bit, registry not addapted


Orca75:
### Description of the Issue
When I reinstalled npp 7.5.4 to the 64bit version, the registry entry for HKEY_CLASSES_ROOT\Applications\notepad++.exe\shell\open\command was not changed.

So it was not possible to assing an extension with "Open with..."

### Steps to Reproduce the Issue

1. 
2. 
3. 

### Expected Behavior
C:\Program Files\Notepad++\notepad++.exe

### Actual Behavior
C:\Program Files(x86)\Notepad++\notepad++.exe

Issue #3101, opened 2017-03-27
x64 version can't be set as default app for anything


Igor-Yavych:
### Steps to Reproduce the Issue
1. Click "open with" in context menu
2. Locate Notepad++ exe
3. Select it
4. Nothing happens. It doesn't appear in the list

When you try to set it as default app via control panel, nothing happens either. 

### Expected Behavior
It should appear in the list but it doesn't. This problem doesn't exist with Notepad++ x32 version. 

I'm using Win 7 x 64

sekile12:
hi

Igor-Yavych:
Hello? Any news? Any additional information you need?

Gusgus01:
FYI, the expected behavior occurs on Windows 10 using v7.4.2 x64. Maybe a Windows 7 issue or not an issue anymore.

Gusgus01:
I think this is a duplicate of #2454

A workaround is available there are well.

mlaffargue:
I had the same problem (with the 32bits version), and resolved it with this  [Microsoft support note](https://support.microsoft.com/en-us/help/2159999/sometimes-you-can-not-associate-a-program-with-an-extension-in-windows)

I installed 64bits version of Notepad++ and because of some problem (plugin manager if I remember) I went back to 32bits version.

It seems that modifying the installed version didn't modify the Registry key : HKEY_CLASSES_ROOT\Applications\notepad++.exe\shell\open\command

It still pointed to `Program Files` instead of `Program Files(x86)`

Putting the right folder made everything work again.

ke352802081770314:
I just freshly installed a Win7 Pro x64 machine, and installed Notepad++ x64 on it, and the first time I doubleclick on a .php file, I am asked to choose the program. While I would have preferred being able to choose Notepad++ from the suggestion list without going through browse and selecting the exe, having done that it works fine for me. No register file extension tricks, no registry manipulation needed.

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

Issue #2817, opened 2017-01-22
Bug with transition from x86 to x64 version. Registration of the app turns out to be invalid.


zalima:
### Description of the Issue
If I have already installed and removed an x86 version, the installation of x64 version turns out to be buggy. The app looks as it is not registered in Windows OS: file associations do not work, and the "Open With" dialogue does not suggest Notepad++ as an option.

### Steps to Reproduce the Issue
1. Install x86 version.
2. Remove x86 version.
3. Install x64 version.
4. Try to associate Notepad++ with .txt by any standard way.

### Expected Behavior
It must allow to be associated with different file types like any other application.

### Actual Behavior
Standard ways do not work.

### Debug Information
This happens because the x86 version sets
[HKEY_CURRENT_USER\Software\Classes\Applications\notepad++.exe\shell\open\command]
@="\"C:\\Program Files (x86)\\Notepad++\\notepad++.exe\" \"%1\""
and the x64 installer does not change this.
If you manually change it to
[HKEY_CURRENT_USER\Software\Classes\Applications\notepad++.exe\shell\open\command]
@="\"C:\\Program Files\\Notepad++\\notepad++.exe\" \"%1\""
then associations by standard means of Windows work fine. (Though associations made from the options menu of Notepad++ seam do not work properly still. But the reason of this possibly is different. Possibly because they do not work properly from the standard user account. Some further investigation is neaded.)
So, when an x64 version is installed, the installer must re-register the application properly.

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
search if related topics exists. you are welcome to post either at similar
[...]

Issue #2440, opened 2016-10-18
Not updating registery when upgrading 32bit to 64bit


tknight0000:
Upgrading from the v6.x.x 32bit version to the v7.1 64bit version, via "Notepad++ Installer 64-bit x64", does not update the registry key in "Computer\HKEY_CLASSES_ROOT\Applications\notepad++.exe\shell\open\command". Which can be tricky to diagnose. As file associations just break and the bad registry reference prevents the application from being added to "Open With" window again.

**Previous Value**:
"C:\Program Files (x86)\Notepadd++\notepad++.exe" "%1"

![registry](https://cloud.githubusercontent.com/assets/22911684/19477157/c6718808-950a-11e6-8986-6546fd75edb1.jpg)

**New Needed Value**:
"C:\Program Files\Notepadd++\notepad++.exe" "%1"

My machine is Windows7 64bit (with windows updates on)

alexfeigin:
I also had to fix similar an issue on my pc, in my case I found the bad entry for each users-id at:
HKEY_USERS\<user-id>\Software\Classes\Applications\notepad++.exe\shell\open\command
same behavior on win 7 - bad entry keeps me from re-adding notepad++ as an application to handle file extension association 
I understand from [this](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724475(v=vs.85).aspx) that HKEY_CLASSES_ROOT is just a view over user/machine keys - so a fix might need to be on each user and not classes_root

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

PR #5000, opened 2018-11-09
Fixed file association bugs and enhancement

SinghRajenM:
This PR is can be broken into two pieces:
1. Fixes below defects
	#1052
	#2817
	#3101
	#2440
	#1786
may be few others
**[Edit]**
#4065

	Nsis script is changed to handled these cases.
	1. Application is registered in global scope (HKLM registry is written) which helps to associate an extension properly.
	2. Whenever, there is an upgrade from x86 to x64 or vice versa, registry should details should be updated. So that existing file extensions will not be broken.
	3. On uninstall, 
		3.1 Un register the application from the global and user scope.
		3.2 Refresh shell icons so that npp icon from the associated file extensions will be changed. 
	
2. Enhance File association
	1. File extension can't be associated from "Settings->Preferences->File Association" if user does not have administrative rights. Moreover, writing HKCR is not recommended way to associate file extensions. As per [MSDN](https://docs.microsoft.com/en-us/windows/desktop/sysinfo/hkey-classes-root-key), HKCR is merged view of HKCU and HKLM and it is primarily intended for compatibility with the registry in 16-bit Windows. So instead of writing to HKCR, write to HKCU which does not require admin access.
	2. Refresh the shell to update icon of files whenever file extension is registered or de-registered to/from npp.

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -193,6 +193,7 @@ Section -"Notepad++" mainSection
 	${If} $diffArchDir2Remove != ""
 		!insertmacro uninstallRegKey
 		!insertmacro uninstallDir $diffArchDir2Remove 
+		Call unRregisterApplication
 	${endIf}
 
 	Call copyCommonFiles
@@ -203,6 +204,8 @@ Section -"Notepad++" mainSection
 	
 	Call shortcutLinkManagement
 	
+	Call registerApplication
+
 SectionEnd
 
 ; Please **DONOT** move this function (SetRoughEstimation) anywhere else
--- a/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh
+++ b/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh
@@ -282,4 +282,17 @@ Function shortcutLinkManagement
 	SetShellVarContext current
 FunctionEnd
 
+Function registerApplication
+	; Register progID
+	; using class as "Notepad++_File" as it was used in npp source code as well
+	WriteRegStr HKLM "SOFTWARE\Classes\Notepad++_File" "" "Notepad++ Document"
+	WriteRegStr HKLM "SOFTWARE\Classes\Notepad++_File\DefaultIcon" "" '"$INSTDIR\notepad++.exe", 0'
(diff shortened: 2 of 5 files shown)


alexhass:
> File extension can't be associated from "Settings->Preferences->File Association" if user does not have administrative rights. Moreover, writing HKCR is not recommended way to associate file extensions. As per MSDN, HKCR is merged view of HKCU and HKLM and it is primarily intended for compatibility with the registry in 16-bit Windows. So instead of writing to HKCR, write to HKCU which does not require admin access.

Interresting. VSCode setup does it wrong, too and changes HKCR. Have you fully tested this change?

donho:
Using installer to deal with the extension is a way but not the best one IMO.
I will see what I can do about modifying file extension in Notepad++ with no administrator privilege mode.

SinghRajenM:
> Using installer to deal with the extension is a way but not the best one IMO.

No @donho, no!!! How you are going to handle below case then - 

>ii. Whenever, there is an upgrade from x86 to x64 or vice versa, registry should details should be updated. So that existing file extensions will not be broken.

Unless you write an another module which will be executed on migration (from x86 to x64 or vice versa) to update the registry or you implement some command line argument to notepad++ itself and call it silently. I don't see any worth in both the approach just to beat [one line](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5000/files#diff-9ef16f6f0c6b2faa63ea9fac44ccbb0dR290)
```WriteRegStr HKLM "SOFTWARE\Classes\Notepad++_File\shell\open\command" "" '"$INSTDIR\notepad++.exe" "%1"'```

Similarly how you are going to handle below case - 
>iii. On uninstall,
3.2 Refresh shell icons so that npp icon from the associated file extensions will be changed.

What will be the alternative just to beat another one line `system::Call 'shell32.dll::SHChangeNotify(i, i, i, i) v (${SHCNE_ASSOCCHANGED}, ${SHCNF_IDLIST}, 0, 0)'`,?


Moreover, installer code is just to register the app and cleanup the things (like de-registering the app, refresh shell icons etc.) on upgrade or uninstall. And IMHO, these two things MUST be only and only in installer code....

Anyway, it's your call. You can decide what suit best to npp. Thanks!

SinghRajenM:
@donho have you found way to deal with this issue.

Community link (apart from the big list in bug tracker) to asking about the solution-
https://community.notepad-plus-plus.org/topic/18320/there-is-no-other-way-to-associate-files-from-np-preferences-only-if-you-start-the-app-with-admin-rights

PR #5025, opened 2018-11-17
Fix menu bar for Hebrew

ophers:
Also, I'll commence work on closing the gap on the Hebrew translation (bring it up to par with the English source).
I'm bilingual, Hebrew is my native tongue and I have excellent command of it.
Back in the day (i.e. 2006) I localized [CUPS](https://github.com/apple/cups) to Hebrew.


chcg:
Seems to be ok according to https://translate.google.com/#iw/en/%D7%97%D7%9C%D7%95%D7%9F%0A%D7%AA%D7%9E%D7%99%D7%9B%D7%94

PR #5066, opened 2018-11-28
PHP: Add styler entry for complex variables

### LINKED ISSUE #5065: PHP: Support styling of complex variables in strings
The PHP lexer in Scintilla v3.6.5 already detects complex variables in strings (`SCE_HPHP_COMPLEX_VARIABLE` value `104`). However, **stylers.model.xml** doesn't target this styleID anywhere, thus a default installation of Notepad++ will cause such variables to fall back to the global default text style (with no way to change the style solely via the GUI).

As a matter of convenience, **stylers.model.xml** should be modified so that Notepad++ ships with a way to easily customize the style of these variables. Something like:

`            <WordsStyle name="COMPLEX VARIABLE" styleID="104" fgColor="808080" bgColor="FEFCF5" fontName="" fontStyle="1" fontSize="" />`

bakedpotat0:
Resolves #5065

### DIFF
--- a/PowerEditor/src/stylers.model.xml
+++ b/PowerEditor/src/stylers.model.xml
@@ -850,6 +850,7 @@
             <WordsStyle name="QUESTION MARK" styleID="18" fgColor="FF0000" bgColor="FDF8E3" fontName="" fontStyle="0" fontSize="" />
             <WordsStyle name="DEFAULT" styleID="118" fgColor="000000" bgColor="FEFCF5" fontName="" fontStyle="0" fontSize="" />
             <WordsStyle name="STRING" styleID="119" fgColor="808080" bgColor="FEFCF5" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="COMPLEX VARIABLE" styleID="104" fgColor="808080" bgColor="FEFCF5" fontName="" fontStyle="1" fontSize="" />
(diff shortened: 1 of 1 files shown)


ArkadiuszMichalski:
@donho I'm wondering where this PR dissapeared. After this PR (https://github.com/notepad-plus-plus/notepad-plus-plus/commits/master/PowerEditor/src/stylers.model.xml) next don't have this rule. This rule is correct https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/lexers/LexHTML.cxx#L739.

dail8859:
@ArkadiuszMichalski Good catch.

@chcg It looks like your PR #5112 removed the changes that this commit made. Was this intentional?

donho:
@ArkadiuszMichalski 
Sorry I don't follow you at all. Could you elaborate it?

ArkadiuszMichalski:
@donho Hmm, I didn't notice that @chcg remove this https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0686435d1ac6caac55520b57b934f736593a7d4c#diff-0e8d56e12e5d6853a3d3e6f56a732ee8L853 line, which solved this bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5065. So this PR dissapeared, the only question is why he removed it.

chcg:
@ArkadiuszMichalski @dail8859 The removal was no intended change. Just xml should have been changed to a valid syntax.

PR #5067, opened 2018-11-28
Update pluginsAdmin.h, using _TCHAR and _istdigit() 

0017031:
use _TCHAR and _istdigit() to avoid wchar/char mismatch, and possible undefined behavior.

https://en.cppreference.com/w/cpp/string/byte/isdigit


SinghRajenM:
Can be complied with VS2017 now.

donho:
@SinghRajenM  Cannot be complied with VS2017 without this PR?

PR #5071, opened 2018-11-28
Corrected minor typos

SinghRajenM:
In parameters.h, only extra semicolons are removed.


PowerEditor/resources/nppPluginList.json
```diff
@@ -15,7 +15,7 @@
 			"folder-name": "mime-tools",
 			"display-name": "Mime tools",
 			"repository": "https://github/npp_plugins/mimetools/",
-			"description": "MIME plugin for Notepad++ implements several main functionnalities defined in MIME (Multipurpose Internet Mail Extensions).",
```
chcg: Already correct at https://github.com/notepad-plus-plus/nppPluginList/blob/master/src/pl.x86.json

SinghRajenM: @chcg Do we need this file in npp repository? I don't think so.

chcg: @SinghRajenM You are right. This file is not needed any longer in the npp repo itself and is outdated version 1.0 anyway.

donho: @SinghRajenM Yes, this file should be removed.

PowerEditor/src/Parameters.h
```diff
@@ -1280,7 +1280,7 @@ const int RECENTFILES_SHOWONLYFILENAME = 0;
 class NppParameters final
```
donho: Please keep `Parameters.h` as it is.

SinghRajenM:
@donho review comments have been accommodated.

PR #5093, opened 2018-12-01
Xml syntax check failing

chcg:
checked all files of the APIs dir for xml syntax issues and corrected the invalid files

Tested with plugin tools and checked also against https://www.xmlvalidation.com/

The current invalid xml syntax seems to have no bad influence on the functionality of the files.

### DIFF
--- a/PowerEditor/installer/APIs/autoit.xml
+++ b/PowerEditor/installer/APIs/autoit.xml
@@ -171,7 +171,7 @@
 			</Overload>
 		</KeyWord>
 		<KeyWord name="AutoItSetOption" func="yes">
-			<Overload retVal="" descr="Changes how literal strings and % symbols are interpreted.  By default strings are treated literally, this option allows you to use %environment% variables inside strings, e.g., &quot;The temp directory is: %temp%&quot;. 1 = expand environment variables (similar to AutoIt v2) 0 = do not expand environment variables (default) Without this option the usual way would be: 	&quot;The temp directory is: &quot; & EnvGet(&quot;temp&quot;) ">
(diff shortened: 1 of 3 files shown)


donho:
@chcg  I let you commit your pr directly. Thank you.

donho:
@chcg Any problem for you to merge your PR into master?

donho:
BTW, it'll be good to check XML syntax for every build - is it possible?

chcg:
@donho 
I think I did:
b. copy 2 commands "git checkout -b USER-BRANCH_NAME master" and "git pull https://github.com/USER/nppPluginList.git USER-BRANCH_NAME" to execute on your local git repository with git command.
c. copy 2 commands "git checkout master" and "git merge --no-ff USER-BRANCH_NAME" from the github PR page then execute on your local git repository with git command.
skipped as I thought that is not needed as there were no conflicts:
d. execute "git rebase -i" on your local git repository with git command to fix up if needs.
e. execute "git commit --amend" on my local git repository with git command to make a more pertinent commit comment if needs.
and then:
f. "git push"

, but it seems the history look different from your merges. Is point d. needed therefore?

chcg:
I think it should be possible to do a similar check as for the json format via python at https://github.com/notepad-plus-plus/nppPluginList/blob/master/validator.py also for the xmls under ...\notepad-plus-plus\PowerEditor\installer\APIs and ...\notepad-plus-plus\PowerEditor\installer\nativeLang. Maybe even add an xml schema to even better check the content of the xmls.

I will take a look at that.

MAPJe71:
> Maybe even add an xml schema to even better check the content of the xmls.

IMO that's the right way (e.g. see [functionList.xsd](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/Test/FunctionList/functionList.xsd)).

Issue #5057, opened 2018-11-25
Number of matches cropped in Find/Replace dialog


vlakoff:
I have updated from 7.4.2 to 7.6, and noticed this crop while using the Mark tab, please see screenshot below where I have circled the issue:

![screenshot](https://user-images.githubusercontent.com/544424/48983179-f39e0380-f0eb-11e8-8dc5-d441b44010c6.png)

Possible regression from 0e60bd89579cd939f2d72800ff4ee4e716c78a1a (in [FindReplaceDlg.rc](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0e60bd89579cd939f2d72800ff4ee4e716c78a1a#diff-8d40907d761514e336b88b08481fc222)), but I don't understand why this hasn't been reported previously.

It's a huge usability regression for me, and the fix is presumably straightforward :)

vlakoff:
This is actually a regression from 213af533899b720fa2abcdd4abcc2f3c577d8d7b, where the text has been changed from "Mark: N&nbsp;matches." to "N&nbsp;matches."

(search for "find-status-mark-" in FindReplaceDlg.cpp and english.xml)

As it's inconsistent with the other status bar strings, I guess this missing "Mark: " is actually an overlook.

<br> The slight crop at the beginning is probably a different, longer-standing issue.

sasumner:
> I guess this missing "Mark: " is actually an overlook.

Put "Mark: " back in!  :-)

vlakoff:
I don't think it's related to [optimization](https://github.com/notepad-plus-plus/notepad-plus-plus/labels/optimization). Maybe you could add [regression](https://github.com/notepad-plus-plus/notepad-plus-plus/labels/regression) instead.

sasumner:
@vlakoff I think what @chcg is saying with that labelling is that the current way it looks is workable but not ideal, so this issue is an "optimization" of that situation. :-)

chcg:
@sasumner @vlakoff From the available labels I thought bug is not the right one as the dialog is still usable, more like a visual glitch. So it might be not a immediate must have fix. I now added also the regression label.

vlakoff:
Fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/157846c60a84558f5ac6e642a1c306542325fce4

PR #5094, opened 2018-12-02
Fix status bar messages in dialog of the Mark All command

vlakoff:
Fix for #5057. Also removed some remaining duplicate entries.

Some languages gave me difficulties:
* Arabic: displays strangely because of right-to-left; I guess my changes are fine, but a double-check would be welcome.
* Chinese: apparently there was a confusion between "mark" and "count"; I did my best to fix it, I let you check it ;)


vlakoff:
I don't think it's related to [optimization](https://github.com/notepad-plus-plus/notepad-plus-plus/labels/optimization) ;)

vlakoff:
Rebased :)

PR #5095, opened 2018-12-02
Update with plugins admin

Ices-Eyes:


SinghRajenM:
You need to update ```<Item id = "48503" name = "Genera negli appunti" />``` as well as per db8b20dd156109c8ab0db6455b6c2b66e02701b4

Issue #4492, opened 2018-05-09
ENH REQ : Change tab's icon to lock when Read-Only Flag Set


perlsmith:
### Description of the Issue

Along the lines of View > Monitoring which changes the icon to an Eye, it would be great if we could change the icon to be a lock when the Read-Only Flag is set.

It has been pointed out that the Disk-Icon gets grayed out, so there is a visual indicator. However, there is a bug in that implementation in that the icon is NOT grayed out when the tab is not selected. Either way, a lock icon would be a nice feature. We might give users a choice between the two.

### Steps to Reproduce the Issue

1. Edit > Set Read-Only
2. See that the icon for the tab is NOT changed to a lock picture.
3. 

### Expected Behavior

When set to read-only, the icon should change to a lock to indicate the file is read-only. It would be nice if the Title-bar text also indicated "Read Only" like Excel and Word do..

sasumner:
You only addressed the case where this status is set via the Edit menu option.  What's your feeling about when a file with the OS's read-only attribute is set on a file that gets opened in Notepad++ ?  I would presume you'd want to see the same "lock" icon, but it is your feature to define.  :)

perlsmith:
Absolutely. Thanks for catching that. 

How easy is it to save the user's "behavior" - as in, "last time you opened this file, you quickly set it to read-only. It hasn't been edited in a while, so it's probably safe to assume you don't want to edit. Would you like to set it to read-only to view"? That would be cool if we could do it. Give the user a pop-up to respond to and achieve his intent quickly.. 

I'll take anything I can get..

SinghRajenM:
Possible duplicate of #3170

alankilborn:
This feature exists when "Alternate icons" are selected in "Settings" > "Preferences" > "General" > "Tab Bar".

It looks like this:

![image](https://user-images.githubusercontent.com/77065706/182616348-8b83a43b-05d0-4e2d-aca3-f896a5a6f2bf.png)

Tested in N++ 8.4.4.

Suggest that this issue be CLOSED.

Issue #3170, opened 2017-04-17
Read-only status not visible on background tabs


GotLag:
### Description of the Issue

![npp snip](https://cloud.githubusercontent.com/assets/13514321/25075679/7ecc722e-2358-11e7-8ce7-b34890b6e807.PNG)
Tabs not in foreground show the standard blue file icon instead of the greyed-out read-only icon. Both of the documents in the above picture are flagged read-only

### Steps to Reproduce the Issue

1. Open multiple documents
2. Set one to read-only
3. Switch to another document

### Expected Behavior

Tabs should indicate read-only status regardless of whether they're active or not

### Actual Behavior

They don't.

### Debug Information

Notepad++ v7.3.3   (32-bit)
Build time : Mar  8 2017 - 03:37:37
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll XMLTools.dll

fhaag:
I came here to report the same issue - but as it has already been reported, I would only like to add: Read-only status and whether or not a file has been saved (red vs. blue disk) are really two orthogonal things. Maybe read-only status should be indicated by some overlay symbol that would then be visible on both foreground and background tabs.

**Use Case**

I sometimes have plenty of files open, some of which serve as examples, while others are what I am working on. Sometimes, the examples and the files being edited even have the same names. Being able to distinguish the write-protected examples at one glance from the writeable files helps improve the overview in this situation.

lars18th:
Hi,

I have a similar problem. In fact, the feedback of the "Read-Only" mode is insufficient.

I suggest to change the "FileName" to something like `[RO] filename.ext`. With the append of this `[RO]` will be more easy to identify the used mode.

I hope to see this implemented soon.
Regards.

sasumner:
+1.  Wow.  I have been using Notepad++ for quite a while and I NEVER noticed this before.

Issue #4595, opened 2018-06-19
File association and registering as the default application


HerculesIL:
File association and registering as the default application

### Description of the Issue
Can't  registering Files like *.TXT , *.LOG etc. as the default application when installed x86 version on Win x64

### Steps to Reproduce the Issue
I have Win10 Ent.
I had mistaken and I Installed x86 version of notepad++ instead x64 version.
So I Uninstall this version and ReInstall the correct version x64 (64bit) .

After Installation I notice that files couldn’t registering as the default application with notepad++

I tried to Replacing all C:\Program Files (x86)\Notepad++ references by C:\Program Files\Notepad++ in registry but this didn’t fixed the problem.

I also tried to Uninstall and run CCleaner to delete all the history data , reboot and Reinstall (x64) but this didn’t fixed the problem either .

### Request a fix / patch
Is there any cleaner tool to delete all the history data from previous installation (Registry) ? 
I met a lot of people with a similar case at Notepad++ community, Is it possible to solve this issue without reinstall Windows O.S ?

warg0x4e:
For now, check out [https://www.nirsoft.net/utils/file_types_manager.html](https://www.nirsoft.net/utils/file_types_manager.html).

chcg:
@SinghRajenM  Could take a look at this and the a similar once:
 https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1786
 https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2440
 https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2454
 https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2817

and some corresponding community threads
https://notepad-plus-plus.org/community/topic/16558/can-t-set-notepad-as-a-default-program-for-any-extension-windows-10
https://notepad-plus-plus.org/community/topic/16558/can-t-set-notepad-as-a-default-program-for-any-extension-windows-10

SinghRajenM:
@chcg, sure I will check in sometime.

chcg:
See PR #5000

PR #5097, opened 2018-12-02
Enhancement: Make readonly tab identifiable even if tab is not active

SinghRajenM:
Fixes: #4492, #3170

image


donho:
It's very clear about the indication. However the style of icon has changed completely - which doesn't go with the others at all. Hence this PR is not accepted.

sasumner:
NIHS?

SinghRajenM:
@sasumner 
> NIHS?

What? I'm hearing this abbreviation first time. 😉

SinghRajenM:
@donho ,
You're right, but you're looking at one side of coin. We already have icon for monitoring which also does not go with others at all. And I felt, lock icon would be best to suit this, so it was used.

Anyway, the idea was to differentiate the readonly tab which should be taken somehow. Thanks!

sasumner:
What does "not go with the others at all" mean?  This is a ridiculous reason for rejection. @SinghRajenM , why would you ever bother to do another PR on Notepad++ code?  Expend your efforts elsewhere, where they might be appreciated.

SinghRajenM:
> What does "not go with the others at all" mean? This is a ridiculous reason for rejection. @SinghRajenM , why would you ever bother to do another PR on Notepad++ code? Expend your efforts elsewhere, where they might be appreciated.

@sasumner  yes sometime, it demotivates. I also felt the same as you. Not only this, but there are around 60 PRs (mine are 8~10) PRs are ready to merge, but still pending. Many of PRs got reject because of the amount of code change stating more code changes are prone to error/regression. This kind of attitude in software industry is not good at all. People's effort should be valued. 

@chcg asked me to look into file association issue #4595 . I was knowing these issues and fix for those already, but just to avoid wastage of time, I didn't paid much attention. But when @chcg asked, I look on those issues again and submitted the PR. But what after that? No response, no feedback, I would say not even reviewed.

I remember the case of @Rikk and why he left this community.

If we look at the bug tracker, as on today (06, dec, 2018) there are 2400 bugs reported. If we exclude duplicate or invalid cases , still around 1500 bugs can be seen which is not a good sign for a popular software. Instead of providing less valued features (e.g. doc peeker while we already have document map in place, Tools->MD5 etc.), some time should be spent to clear bug list as well. I'm not against new features, but meanwhile bug-fixes and enhancement also should be considered.

The only motivation have is: I would like to improve the software which I use daily basis and I love to code.

chcg:
@SinghRajenM Hm, but which PR is releated to 4595? It is not referenced there.

Anyhow just donho is doing the work here and my feeling is that the inflow is a little bit to much for just one guy not working fulltime on it. Or maybe even too much for just one in fulltime.
And I guess a bug only mode with feature requests/enhencements/etc. are not rejected immediately to reduce the current amount of issues might be ok for a profit project but as this is just spare time the focus is probably more driven by current interests as user needs/problems. 
As my part is just https://github.com/notepad-plus-plus/nppPluginList I try to sort and classify things here.
I think some maintainer for various areas of N++ would be helpful as e.g. normal translation PR merges might be also done by someone else apart from donho with some multi-language background.

SinghRajenM:
@chcg just for your reference, #4595 is from windows side (I think starting from 1803). But other defects referenced in that defect (and many reported on community) are handled in #5000. This PR deals around 6~7 issues (may be more) from tracker and hot topic on community as well. So don't you think it is also users needs/problems.

So concern is not this PR. The actual concern is something else. For example, as @sasumner pointed, what is wrong with this PR? If you don't like it does not mean that other will not like. Others will like if it really makes a difference. Similarly plugin admin design is another thing which currently people are discussing a lot.

**Moral of story**: BOSS is not always ➡️. Sometimes he takes ⬅️ and tries to defend that it is ➡️. Unfortunately, everyone is trying to become BOSS.

sasumner:
One thing for sure, though:  BOSS is always BOSS !  :-)

PR #5105, opened 2018-12-06
fixed for change Chinese display name GB2312 to GBK

qxo:
cause after Window 95 936 means GBK not GB2312

GB2312 is subset of GBK

如实际是GBK字符，但在notepad++状态栏显不的是GB2312是一种误导
-- ie: '镕'  is a GBK (936) , not a GB2312


donho:
@popcorner Do you validate it?

popcorner:
@donho I'm not sure how to validate this, and I tried some characters in GBK and not GB2312(passed), and some characters included in GB18030 but not GBK(failed). My OS is in simplified Chinese. The actual encoding seems to be GBK and the description above should be right.

donho:
It's not clear to me to accept this PR.
Until there's other complain, I will keep current status.

Issue #406, opened 2015-07-04
Double-click splitter to resize both panes to equal size


Eldaw:
It would be nice if it was possible to double-click the splitter dividing 2 panes and have it automatically move the splitter to the middle, making both panes exactly equal in size.

MetaChuh:
implemented in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4f24f81eb8e0d1adc87da8d047531504d1b8232e

PR #5118, opened 2018-12-09
Double click splitter resets panes to equal size

jmbockhorst:
Added feature #406.


donho:
Build error on both x86 & x64
@jmbockhorst could you check it and fix it?

jmbockhorst:
I fixed my error and the build is now passing.

Issue #2105, opened 2016-07-20
Add Encoding - Character sets - Chinese - GB18030 (Extended Chinese)


jonilaw:
This character set is formally called "Chinese National Standard GB 18030-2005: Information technology — Chinese coded character set". GB abbreviates Guójiā Biāozhǔn (国家标准), which means national standard in Chinese. The standard was published by the China Standard Press, Beijing, November 8, 2005. Only a portion of the standard is mandatory. Since May 1, 2006, support for the mandatory subset is officially required for all software products sold in the PRC. Due to its Unicode equivalence, GB18030 supports both simplified and traditional Chinese characters.

Quoted from Wikipedia

GB18030 is a must have encoding for text editor now. Thank you.

yzwduck:
GB18030 is supported on Windows XP and later. 
Since n++ is no longer targeting Windows 2000, I suggest replacing GB2312 with GB18030. Changing the code page identifier from 936(GB2312) to 54936(GB18030) in EncodingMapper.cpp, and it should work.

jonilaw:
How do I compile the n++ after editing the EncodingMapper.cpp? Many thanks.

dail8859:
See the [README section](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/README.md#to-build-notepad-from-source).

jonilaw:
Will they include this update in the future release? Thanks.

mirsking:
Looking forward to the GB18030 within notepad++ .

lindasczx:
Or any one knows how to edit n++ src to add gb18030 support?

lindasczx:
@yzwduck 
No, it doesn't work. I've tried it and found Scintllla doesn't support it.

yzwduck:
@lindasczx 
N++ does NOT support GB18030 now. You have to patch N++ to get GB18030 support.

Diff: (I'm not sure whether it is complete or not)
```
diff --git a/PowerEditor/src/EncodingMapper.cpp b/PowerEditor/src/EncodingMapper.cpp
index 15fd1a0b..d7ca8ab6 100644
--- a/PowerEditor/src/EncodingMapper.cpp
+++ b/PowerEditor/src/EncodingMapper.cpp
@@ -73,7 +73,7 @@ static EncodingUnit encodings[] = {
     {866,   "IBM866 cp866 oem866 866"},                                                         //IDM_FORMAT_DOS_866
[...]

PR #5124, opened 2018-12-10
add support for GB18030

qxo:
ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2105

54936 is GB18030

and current GB2312 should change to  GBK ( 936 means GBK after Window 95)


popcorner:
GB18030...

donho:
The Scintilla part should do the PR on Scintilla project.

qxo:
OK 
https://sourceforge.net/p/scintilla/feature-requests/1257/

chcg:
@donho See:

>  The patch is against an old 3.x version of Scintilla so will not apply. For example, Platform::DBCSCharLength no longer exists. The patch does not add GB18030 cases to 4.x methods like IsDBCSLeadByteInvalid and IsDBCSTrailByteInvalid.
> 
> The patch does not appear to handle all GB18030 cases. GB18030 includes 4-byte characters and Scintilla's DBCS support includes code where only 1 or 2-byte characters will work, such as in SafeSegment:
> 
> 

So it seems to be unlikely that this change will ever appear on a 3.x branch. If this would be added to scintilla how to you plan to backport this for N++?

Issue #1898, opened 2016-05-27
No Python code folding past level 1


HactarCE:
Alt + 1 works fine to collapse the first level of code, but Alt + 2-9 don't work (not do the menus). Everything is fine when using syntax highlighting for other languages. Manually collapsing the code (by hitting the [-] button by the line numbers) works fine, however.

HactarCE:
So forgive me if I'm wrong, but a quick-and-dirty solution would be to divide the "FoldLevel" by 4?

dail8859:
I looked deeper into the code and it seems to be based on indentation so it's not as easy as just dividing. Plus I'd hate to have to add a special case just for python. I submitted a bug report to Scintilla to see if the odd fold level is intended or not.

cmeriaux:
Hello guys,
I also noticed the problem sometimes ago.
I did some test and my conclusions are : 
scintilla python lexers 
- doesn't support indentation with tabs.
- every line indented with 1 space will be attached on fold level 1
- every line indented with 2 spaces will be attached on fold level 2
- every line indented with n spaces will be attached on fold level n

So if your python code uses a 4 spaces tabulation, the first level level of indentation will be associated with the fold level 5.

@HactarCE your quick-and-dirty solution wouldn't be divide the "FoldLevel" by "tab size"
 ....

dail8859:
> doesn't support indentation with tabs.

Tabs have always worked fine for me.

> every line indented with n spaces will be attached on fold level n

Yes this is the conclusion I came to as well. Also, indenting with a tab sets it to a fold level of 8 no matter what the width of the tab character is set to.

Here is the Scintilla bug [report](https://sourceforge.net/p/scintilla/bugs/1834/).

dail8859:
The Scintilla author reports the Python lexer is working as intended. To properly fold the `n`th level of indentation is not straight forward. It is possible this kind of thing happens to other lexers.

pembo13:
Any updates on this issue?

victorclaessen:
I would really like this to work with tabs as well.

MaxCorman:
I would seriously like this feature to be resolved since i use Notepad++ for normal text categorization as the Python, Nimrod, and Matlab languages are very convenient since they don't require putting characters everywhere for folding.

Is there some kind of workaround to this?

MaxCorman:
[...]

PR #5131, opened 2018-12-12
Fixed issue #1898 by using a fold level stack.

zufuliu:


victorclaessen:
This is a really useful feature for Python programmers, and I would like to express my hope that this gets integrated soon.

zufuliu:
Since NPP updated Scintilla to v4.1.4 (f548db82ed3cc19c1716bec3e80ab197868ca9e6) This PR is out-dated.
`ScintillaEditView::isFoldIndentationBased()` need update to what been used, i.e. find `IndentAmount` in used lexers. e.g. `SCLEX_MATLAB` is indented by keywords in the new Scintilla.

By the way, here is also a buit-in Markdown lexer (LexMarkdown), `userDefinedLang-markdown.default.modern.xml` maybe changed to use it. Anothor lexer LexIndent like lexNull but comes with indentation based folding.

zufuliu:
Scintilla 4.1.5 is already released,  https://www.scintilla.org/ScintillaHistory.html

zufuliu:
Following change in `CodePageFromCharSet()` (in ScintillaWin.cxx) may cause regression bug for ANSI code pages other than DBCS and 1252, e.g. on Russian (1251), Greek (1253), Turkish (1254), Thai (874) systems.

```diff
-- case SC_CHARSET_DEFAULT: return documentCodePage;
++ case SC_CHARSET_DEFAULT: return documentCodePage ? documentCodePage : 1252;
```

Here are some discussion on https://sourceforge.net/p/scintilla/bugs/2093/

donho:
@zufuliu  So if the update of Scintilla 4.1.4 fixes #1898 ?

zufuliu:
@donho No, no changes on these indentation based folding even in 4.1.5 or latest mercurial source.

donho:
> Since NPP updated Scintilla to v4.1.4 (f548db8) This PR is out-dated.

So will you update this PR for fixing the bug mentioning above?

zufuliu:
It's updated.

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -1949,23 +1949,83 @@ void ScintillaEditView::bufferUpdated(Buffer * buffer, int mask)
...
 void ScintillaEditView::collapse(int level2Collapse, bool mode)
 {
 	execute(SCI_COLOURISE, 0, -1);
 
-	int maxLine = static_cast<int32_t>(execute(SCI_GETLINECOUNT));
+	const int maxLine = static_cast<int32_t>(execute(SCI_GETLINECOUNT));
+	int line = 0;
 
-	for (int line = 0; line < maxLine; ++line)
+	if (isFoldIndentationBased())
```
donho: In order to avoid the regression and for the sake of maintenance, the old code should be kept. 
Use `isFoldIndentationBased()` to call the new code.
So you can also simplify the new code.

LordNyriox:
@zufuliu:  Perhaps you can squash the commits in this PR, for readability?

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
zufuliu: It's moved to a new method collapseFoldIndentationBased().
Don't know how that can be simplified.

donho: @zufuliu
It looks OK for me. Thank you.

PR #5175, opened 2018-12-29
Update functionList.xml C++ parser

Baptistou:
Hello,

I use Notepad++ every days, most often for C++ language and I noticed that the Function List feature by default doesn't recognize most function and class definitions. I have updated the C++ parser in functionList.xml and solved this issue for files with code syntax from C++98 to C++20.
I also tested it with your C++ unit test and I found an error in your unitTest.expected.result, I will open an issue for that.

Improvements :
- add templated classes (and template-nested templated classes)
- add namespaced classes
- add classes which inherits another class without the accessibility keyword (private is optional for example)
- add classes with virtual inheritance
- add functions defined with the following keywords
  - (left) consteval|constexpr|friend|inline|static|virtual
  - (right) const|final|noexcept|override
- add functions defined with templated types (and template-nested templated types)
- add operator definitions


Baptistou:
Concerning your first point, do you suggest to update the current C++ [unitTest](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/Test/FunctionList/cpp/unitTest) file, to create another unitTest file or to attach code samples and screenshots here ?

MAPJe71:
Update the set of the current C++ unitTest file and its related result file or create additional sets.

PowerEditor/Test/FunctionList/unitTestLauncher.ps1
```diff
@@ -1,46 +1,25 @@
...
-..\..\bin\notepad++.exe -export=functionList -ljava .\java\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -ljavascript .\javascript\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lnsis .\nsis\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lperl .\perl\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lphp .\php\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lpowershell .\powershell\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lpython .\python\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lruby .\ruby\unitTest | Out-Null
-..\..\bin\notepad++.exe -export=functionList -lxml .\xml\unitTest | Out-Null
+$languages = @("asm","autoit","bash","batch","c","cpp","cs","ini","inno","java","javascript","nsis","perl","php","powershell","python","ruby","xml")
```
MAPJe71: Nice :thumbsup:

Baptistou:
I couldn't use postfix as .\language\unitTest* would also match unitTest.expected.result. Unless to rename all files, prefix is the best solution.

MAPJe71:
Isn't it possible to filter out the files with `.expected.result` extension?

Baptistou:
Maybe I could add a pipe or a if to filter *.expected.result and *.result.json files.

donho:
@Baptistou  Sorry for the long waiting process, could you please solve the conflict so I can review your code?

donho:
Nice job @Baptistou !

Issue #3650, opened 2017-08-25
Auto-detect double default extensions (same extension for both R and REBOL)


DoubleYouGTT:
### Description of the Issue
Since the addition of REBOL to the auto-detection, all my R files are detected as REBOL. This is due to the setting of the default extension 'r' for both R and REBOL.

![defaultext](https://user-images.githubusercontent.com/15901992/29712881-53a5b182-899c-11e7-91f2-be1e8d2e15bb.png)

### Possible Fix
Removing the default extension 'r' from the REBOL item in langs.xml.

djhurio:
I affirm this issue. Thank you for the suggestion how to fix this:

    <Language name="rebol" ext="reb" commentLine=";" commentStart="" commentEnd="">

vikramvarun:
How to resolve this issue on GitHub website. My repository https://github.com/vikramvarun/10-Days-of-Statistics has codes written R but GitHub is showing them as Rebol.

dcomtois:
For a temporary fix:
Remove R key from REBOL section in the file ‘langs.xml’ (path: User\AppData\Roaming\Notepad++\langs.xml )

source: https://stackoverflow.com/questions/38317644/notepad-not-recognizing-r-language

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/627c00b4338876b585870cf556578d491dd6affa

FVFaleiro:
This error persist in the Notepad ++ version 7.7.1.

Even with these lines below are apparently correct, the software recognize .R files as Rebol.
<Language name="rebol" ext="r2 r3 wreb" commentLine=";" commentStart="" commentEnd="">
<Language name="r" ext="r s splus" commentLine="#">

Debug information: 
Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
@FVFaleiro 

Try this:

Grab a portable version of 7.7.1 and extract and run it.  Drag a `.r` file into it.  Note that status bar says `R programming language`.  At least it does for me.

xylographe:
@FVFaleiro 
[...]

PR #5203, opened 2019-01-06
Remove "r" from ext for Rebol

dcomtois:
Related to issue #3650


dcomtois:
> 
> 
> See https://fileinfo.com/extension/r, seems r is most commonly used for R in the first place.

Yes indeed.

Issue #5194, opened 2019-01-03
Hang and show "bad allocation" dialog when view the attached text file.


supermingwu:
### Description of the Issue

Hang and show "bad allocation" dialog when view the attached text file.

### Steps to Reproduce the Issue

1. Open the attached text file: 2019-01-02-20-26-58_android.txt
2. Scroll the view (or press PageDown) to the end of text file.
3. Show "bad allocation" dialog during scrolling and we can't scroll sown any more.

### Expected Behavior

We can view all content of the text file. 
We can do this by PSPad and WordPad, but fail with NotePad++.

### Debug Information

Notepad++ v7.5.9   (32-bit)
Build time : Oct 14 2018 - 15:02:52
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll ComparePlugin.dll Explorer.dll HexEditor.dll NppColumnSort.dll NppTextFX.dll PluginManager.dll 

[2019-01-02-20-26-58_android.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/2722649/2019-01-02-20-26-58_android.txt)

SinghRajenM:
@Predelnik
This crash is due to DSpellCheck.dll and seen with DSpellCheck V1.4.8 which can't be seen with V1.4.9.

@supermingwu , could you update DSpellCheck version and check whether issue exists or not.

Predelnik:
Yes, in 1.4.9 I fix an issue with spell-checking files containing invalid utf-8 symbols which this file has in abundance e.g.
![image](https://user-images.githubusercontent.com/2133957/50851479-04e65b00-138e-11e9-90c7-e12e812a0651.png)

supermingwu:
Yes, this bug is fixed after update DSpellCheck to V1.4.9.

PR #5212, opened 2019-01-08
Make exception error more clear

SinghRajenM:
Easy to identify issues like #5194 

**Now**
image

**Then**
image


sasumner:
Wow.  What a great idea!  Nice one.

PR #5222, opened 2019-01-11
Update langs.model.xml

EbATpy:
open *.arxml as *.xml


donho:
Use user customized ext.

Issue #5253, opened 2019-01-22
Can't undo replaced text


ElSchiffer:
Notepad++ v7.5.9
Build time : Oct 14 2018 - 15:02:52
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
OS : Windows 10

### Description of the Issue
When user search and replace some text to another and accidentally click "Replace all in all opened document" user can't undo all document that replaced

### Steps to Reproduce the Issue
1.  Open find and replace
2. Click "Replace all in all opened document"
3.

sasumner:
It's possible to undo, you just have to move to each document and execute Undo.  Not great, but not catastrophic either.

Myrcellion:
But now "Replace All in All Opened Documents" sounds like a dangerous feature, maybe it should ask for confirmation before executing the replacement.

sasumner:
>did dev noticed this issue?

Did you happen to notice that there are 2476 open issues?  It's really hard to "notice" a specific one within the context of that many.

donho:
As @sasumner said:
> It's possible to undo, you just have to move to each document and execute Undo. Not great, but not catastrophic either.

sasumner:
@donho I guess the each-doc-undo is okay for sane people like you and me :) that run with a small number of tabs. There are others that have hundreds (I'd guess).  Maybe requesting confirmation is not a bad idea (although obviously not to apply if RAOD occurs in a macro). Reconsider? But it's nothing to me! :)

keanerickson:
Much love for notepad++ but this feature is really a terrible liability and I wish I could remove this button entirely.

MetaChuh:
@keanerickson 

if you don't know what you are doing, when pressing a certain notepad++ button, please don't press it.

as all files can be undone, as long as notepad++ is not closed, the only liability sits between keyboard and chair in this case.

note: if you only notice human errors after a while and not immediately, please enable verbose backup, so that you can always revert to any older save state of any specific file.

many thanks and best regards.

keanerickson:
How about putting a button for an incredibly niche feature more than 10 pixels away from a button that is viable for every day use?

sasumner:
Released in N++ version 7.8.8

PR #5260, opened 2019-01-23
Add confirmation dialog for "Replace All in All Opened Documents"

Myrcellion:
The issue #5253 states that a user can accidentally replace all occurrences of a string in all opened documents rather than in a single document. When it happens, the user needs to undo in every opened document, one after another. In order to prevent this, I thought that adding a confirmation dialog before executing the replace could be a solution.

replace

The code behind the dialog is inspired from the doCloseOrNot function from Notepad_plus.cpp
code


vlakoff:
This would be consistent with "Find in Files > Replace in Files" which already has a confirmation dialog.

On the other hand, a mistake with "Replace All in All Opened Documents" is not a disaster (you can undo the changes and the files are not saved) whereas "Find in Files > Replace in Files" is quite dangerous (not undoable, files saved in the process).

Also, maybe the two dialogs could be uniformized a bit (for example one has Yes/No the other has OK/Cancel).

Myrcellion:
I updated the options to OK/Cancel and modified the text of the dialog to match the "Find in Files > Replace in Files" dialog.
npp03
A mistake with "Replace All in All Opened Documents" is indeed not a disaster, it is simply annoying and potentially time-consuming as it has to be undone for every opened document, which may be a boring operation if there are many. This additional dialog is certainly not vital, but as you said, it would be consistent with "Find in Files > Replace in Files" and prevent some mistakes.

vlakoff:
I'd tend to agree with the addition of this dialog.

I use a lot the "Replace All" feature, but much less the "Replace All in All Opened Documents" feature, and when I do it's for "large works" (refactoring, etc) where an additional prompt doesn't add much time to the total time spent on the task.

sasumner:
I would have thought that "Yes/No" is better than "OK/Cancel" to an "Are you...?" question.  Meh.

vlakoff:
Personally I'd vote for OK/Cancel, because Yes/No is a choice between two options, whereas OK is for confirming ("go forward") and Cancel is a "do nothing, panic escape" action.

But this could be debated for ages, let's be happy as long as the two dialogs are consistent :)

Myrcellion:
Well, the "Find in Files > Replace in Files" already exists and uses Ok/Cancel, the changes to the text of "Replace All in All Opened Documents" are also based on this already existing dialog, for consistency purpose.

donho:
As @sasumner said:
> It's possible to undo, you just have to move to each document and execute Undo. Not great, but not catastrophic either.

PR #5277, opened 2019-01-28
Remove autocomplete directory on uninstall

SinghRajenM:


vlakoff:
On a related note, on program update the directory `plugins\APIs` is emptied but not deleted. Maybe this could be fixed in [autoCompletion.nsh](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nsisInclude/autoCompletion.nsh) :)

PR #5286, opened 2019-01-30
Only checking for filesystem changes when the tab is selected

ivanmari:
A simple fix to avoid having to close every externally modified file notification at once.
Issue fixed: notepad-plus-plus#1087


jonathonmartin:
As someone with chronic too many tabs syndrome, I could not be happier that this is finally getting some attention. It's so cognitively disruptive to be pulled away from something you are trying to look at, wade through dialogs, then hunt for the tab you were originally interested in.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ba60fbdd715d647aa5ce7be5b6b75182b4a52951

PR #5306, opened 2019-02-10
Fixed build error VS2017

SinghRajenM:
VS 2017 complains about it and says type mismatch. Therefore, no object warning as error.
Moreover, ```_nbUserLang``` is ```unsigned char```, then there is no need to declare ```begin``` and ```end``` as ```int```.

image

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -1681,11 +1681,11 @@ bool NppParameters::getUserParametersFromXmlTree()
 std::pair<unsigned char, unsigned char> NppParameters::addUserDefineLangsFromXmlTree(TiXmlDocument *tixmldoc)
 {
 	if (!tixmldoc)
-		return std::pair<unsigned char, unsigned char>(0, 0);
+		return std::pair<unsigned char, unsigned char>('\0', '\0');
 
 	TiXmlNode *root = tixmldoc->FirstChild(TEXT("NotepadPlus"));
 	if (!root)
-		return std::pair<unsigned char, unsigned char>(0, 0);
+		return std::pair<unsigned char, unsigned char>('\0', '\0');
 
 	return feedUserLang(root);
 }
@@ -2607,7 +2607,7 @@ bool NppParameters::getShortcuts(TiXmlNode *node, Shortcut & sc)
 
 std::pair<unsigned char, unsigned char> NppParameters::feedUserLang(TiXmlNode *node)
 {
-	int iBegin = _nbUserLang;
+	unsigned char begin = _nbUserLang;
 
 	for (TiXmlNode *childNode = node->FirstChildElement(TEXT("UserLang"));
 		childNode && (_nbUserLang < NB_MAX_USER_LANG);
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Parameters.cpp
```diff
@@ -1681,11 +1681,11 @@ bool NppParameters::getUserParametersFromXmlTree()
 std::pair<unsigned char, unsigned char> NppParameters::addUserDefineLangsFromXmlTree(TiXmlDocument *tixmldoc)
 {
 	if (!tixmldoc)
-		return std::pair<unsigned char, unsigned char>(0, 0);
+		return std::pair<unsigned char, unsigned char>('0', '0');
```
jstine35: in C++ `0 != '0'`   This needs to be rewritten as either:

```
return std::pair<unsigned char, unsigned char>((unsigned char)0, (unsigned char)0);
```
or optionally, using `escape character`  syntax for ASCII NUL:
```
return std::pair<unsigned char, unsigned char>('\0', '\0');
```

PowerEditor/src/Parameters.cpp
```diff
@@ -1681,11 +1681,11 @@ bool NppParameters::getUserParametersFromXmlTree()
 std::pair<unsigned char, unsigned char> NppParameters::addUserDefineLangsFromXmlTree(TiXmlDocument *tixmldoc)
 {
 	if (!tixmldoc)
-		return std::pair<unsigned char, unsigned char>(0, 0);
+		return std::pair<unsigned char, unsigned char>('0', '0');
 
 	TiXmlNode *root = tixmldoc->FirstChild(TEXT("NotepadPlus"));
 	if (!root)
-		return std::pair<unsigned char, unsigned char>(0, 0);
+		return std::pair<unsigned char, unsigned char>('0', '0');
```
jstine35: in C/C++,  `0 != '0'`  -- This needs to be changed.  See inline comment above for details.

PowerEditor/src/Parameters.cpp
SinghRajenM: Nice catch. 👍

donho:
@SinghRajenM I just built 32/64 both releases under my VS2017, no error has been found.

SinghRajenM:
@donho may be you tried with VS2015 toolset. As I can see this error in both VS2017 and VS2019 Preview. Try to change the toolset and see the magic. 😜 

image

image

jstine35:
To clarify:  This is a warning which is being promoted to error (I'm not that familiar with NP++ build policy but I assume it might have warnings-as-errors set as a policy).  Another option to fix this compile error is to disable the warning.

Many Visual C++ warnings are frivolous and unproductive.  Disabling them is not only the easy solution, but quite often also the smart one.

SinghRajenM:
@donho  please look into this again as I need to modify `Parameters.cpp` all the time whenever I take latest from main trunk. This is little tedious process.

I don't have VS2015 installed so I build npp using vs2017 with its toolset which always reports this error.

@jstine35 
>Another option to fix this compile error is to disable the warning

I feel this is not a good idea. For example unused variable warning `(warning C4189: 'a': local variable is initialized but not referenced)` is also treated as error which is good and we should removed used items.

PR #5312, opened 2019-02-11
Bug Fix for Tab context menu

SinghRajenM:
1. Disable "Open containing Folder in Explorer" and "Open Containing Folder in CMD" for unnamed tab.
2. Now it is aligned with "File->Open Containing Folder->Explorer" and "File->Open Containing Folder->Cmd".

For Code review, Refer [Notepad_plus.cpp line 2000 and 2001](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Notepad_plus.cpp#L2000)

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -565,6 +565,8 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			bool isFileExisting = PathFileExists(buf->getFullPathName()) != FALSE;
 			_tabPopupMenu.enableItem(IDM_FILE_DELETE, isFileExisting);
 			_tabPopupMenu.enableItem(IDM_FILE_RENAME, isFileExisting);
+			_tabPopupMenu.enableItem(IDM_FILE_OPEN_FOLDER, isFileExisting);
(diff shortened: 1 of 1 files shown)


donho:
@SinghRajenM Do we have the bug report on issue tracker?

SinghRajenM:
@donho , no there is no bug reported for it. I found it myself. Do you want a defect to be logged?

donho:
I just tried to reproduce the bug, but I have to understand what it is the bug firstly.
What do you mean by "Disable "Open containing Folder in Explorer" and "Open Containing Folder in CMD" for unnamed tab" ?

sasumner:
I can jump in and explain.  If you have an unnamed tab, e.g. "new 2", then it has no folder (it is unsaved). Thus "open containing folder" options make no sense and should be disabled/greyed.  Unless perhaps it is desired to be taken to the folder containing the "scratch" files Notepad++ uses for files of this type?

SinghRajenM:
Continue to @sasumner ,
"open containing folder" take you to "My PC" or "My Computer" while
"Open Containing Folder in CMD" takes you to Notepad++ installation path.


Additionally, Options at "File->Open Containing Folder->Explorer" and "File->Open Containing Folder->Cmd" are grayed out which are expected. Which is not sync with context menu of tab.

PR #5314, opened 2019-02-12
Crash Fixed: Regression

SinghRajenM:
When try to save an unsaved file, crash happens. This regression caused by 7fe3cda1d09d133b3d5cdb982c8ffd4b7c7fdd12

image

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
@@ -122,7 +122,7 @@ int FileDialog::setExtsFilter(const TCHAR *extText, const TCHAR *exts)
 	extFilter += TEXT(")");	
 	
 	// Resize filter buffer
-	int nbCharAdditional = static_cast<int32_t>(extFilter.length() + lstrlen(exts) + 3); // 3 additional for nulls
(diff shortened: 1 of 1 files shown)


donho:
@SinghRajenM Thank you for your heads up.
I prefer my one line fix - clear and efficient:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/1ab162480079c7c8870af90bb45b9239ac5ad427

SinghRajenM:
One line fix is clear and efficient **only in look** @donho. It does not solve the problem. Did you really test your change?

wrongfix

SinghRajenM:
> Did you really test your change?

I'm asking because I did the same thing. But in testing found, that it is not proper. That's why my one line fix got converted to 3~4 line automatically. 😊

donho:
Interesting! So did the bug exist before the replacement of lstrcpy to wcscpy_s ?
I'll check your solution tomorrow.

SinghRajenM:
> Interesting! So did the bug exist before the replacement of lstrcpy to wcscpy_s ?

No, it was working fine earlier.

donho:
@SinghRajenM  Nice fix. Could you resolve the conflict please.

SinghRajenM:
@donho done!

PR #5323, opened 2019-02-17
Support More localization

SinghRajenM:
Support localization for few more items:
1. All the controls on "Tab rename" dialog
2. Label ("Max Size: ") on "recent file history" under preference dialog
3. Label ("Tab Size: ") on "Language" under preference dialog

## Edit
4. Label ("Nb of column: ") on "editing" under preference dialog
5. Label ("Nb char: ") on "auto-completion" under preference dialog
6. Title and static control on the dialog which appears when "Create New...", "Rename" and "Save as" is clicked.

PR #5311, opened 2019-02-11
Feature: Support rename for unnamed tabs

SinghRajenM:
This PR implements request described in #3994.

ToDo: (Only if this feature sounds good, ToDo will be implemented)
 - [ ] Support localization for ```StringDlg```
 - [ ] Extend class ```StringDlg```, so that characters which are not allowed in a file name such as `|`, `:`, `>`, `<` etc. will be restricted for tab's new name so that there will be no problem while saving backup file.

renametab

PR #5324, opened 2019-02-17
Don't allow restricted characters for tab file rename

SinghRajenM:
As promised in PR #5311, below items have been implemented now - 

- [x] Support localization for ```StringDlg``` 
=> This is done in PR #5323
- [x] Extend class ```StringDlg```, so that characters which are not allowed in a file name such as `|`, `:`, `>`, `<` etc. will be restricted for tab's new name so that there will be no problem while saving backup file. 
=> This is done in the current PR.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/UserDefineDialog.h
+++ b/PowerEditor/src/ScitillaComponent/UserDefineDialog.h
@@ -399,13 +399,17 @@ class StringDlg : public StaticDialog
 {
 public :
     StringDlg() : StaticDialog() {};
-	void init(HINSTANCE hInst, HWND parent, const TCHAR *title, const TCHAR *staticName, const TCHAR *text2Set, int txtLen = 0, bool bGotoCenter = false) {
+	void init(HINSTANCE hInst, HWND parent, const TCHAR *title, const TCHAR *staticName, const TCHAR *text2Set, int txtLen = 0, const TCHAR* restrictedChars = nullptr, bool bGotoCenter = false) {
 		Window::init(hInst, parent);
 		_title = title;
 		_static = staticName;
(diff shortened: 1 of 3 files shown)


SinghRajenM:
@donho 

As "rename unsaved tab" feature will be available from next release, we need to make backup of renamed unsaved tab is also maintained.

Current implementation doesn't restrict user from entering reserved keys (such as < > : \ /  | ? * etc). If user enters one of these characters for file name, then backup will not be created.

So before we start getting complains, please look into this PR.

PowerEditor/src/ScitillaComponent/UserDefineDialog.h
```diff
@@ -417,12 +421,20 @@ public :
...
+	LRESULT static CALLBACK customEditProc(HWND hEdit, UINT msg, WPARAM wParam, LPARAM lParam);
+
+	bool isAllowed(const generic_string& txt);
+	void HandlePaste(HWND hEdit);
+
 private :
     generic_string _title;
     generic_string _textValue;
     generic_string _static;
+	generic_string _restrictedChars;
```
Faq: extra tab?

SinghRajenM: No.  Moreover, it's not me. Visual Studio does it automatically.

`    generic_string _static;` => There are 4 spaces
`	generic_string _restrictedChars;` => Here it is a tab.
image

LordNyriox: @SinghRajenM:  I usually solve this issue by highlighting the "tab" character, typing `Ctrl`+`H` (find and replace), then entering 4 spaces into the "replace" field.

Issue #5339, opened 2019-02-24
Junk character on macro playback


SinghRajenM:
### Description of the Issue

Playing recorded macro produces junk character on find window (under mark tab)

### Steps to Reproduce the Issue

1. Open notepad++ and start recording macro
2. <kbd>Ctrl</kbd>+<kbd>F</kbd>and click on "Mark"
3. Type some text and click on "Mark All"
4. Now stop macro recording.
5. Now start playback (menu->macro->playback) **without closing find window**
6. Observe text on Find window status bar

### Expected Behavior

Find window status bar should have respective marked word counts.

### Actual Behavior

Find window status bar should has junk characters.

### Debug Information

Notepad++ v7.6.3   (64-bit)
Build time : Jan 27 2019 - 17:16:47
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll 

![image](https://user-images.githubusercontent.com/14791461/53303050-cb10db80-388b-11e9-83d1-01bf9bcc77d9.png)

![image](https://user-images.githubusercontent.com/14791461/53303064-f0054e80-388b-11e9-9361-976f88555137.png)

MetaChuh:
@SinghRajenM 

nice find and very intriguing.

i can confirm and reproduce your discovery on **all** my current test setups, notepad++ 7.5.5, 7.5.9, 7.6, 7.6.2 and 7.6.3, installed as well as portable, both x86 and x64.

it will probably be the same behaviour on **all older versions**, that have a **status bar at the find window**.
(i don't see this character junk on e.g. version 5.9.3, because it is so old, that it didn't have the find status bar implemented yet)

PR #5340, opened 2019-02-24
[Potential crash cause]: Fixed macro playback junk char issue

SinghRajenM:
Fixed defect #5339

image

~~#### P.S: I know "0 matches" is not proper string, it should be "0 match". But is very minor and similar minor grammar mistakes can be found many places (as can been seen in below screenshot as well). The purpose of this is PR is to remove the junk character which eventually can be cause to crash.
image~~

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2710,10 +2710,7 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, generic_string s
 						}
 						else
 						{
-							TCHAR moreInfo[128];
-							NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance())->getNativeLangSpeaker();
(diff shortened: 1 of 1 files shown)


sasumner:
>I know "0 matches" is not proper string, it should be "0 match"

Why do you think this?  I think "0 matches" is totally correct.

SinghRajenM:
I feel it should be 
 "0 match.", "1 match.", "2 matches", "3 matches" and so on.......

> "1 match.", "2 matches", "3 matches" and so on.......
image
image


are already there.

>  "0 match."

Why not this then. 

OTOH why not "`1 matches`"?

LordNyriox:
@SinghRajenM: 

>I feel it should be
>"0 match.", "1 match.", "2 matches", "3 matches" and so on....... 

English language conventions can be so complicated…and I am a native speaker of the tongue. 

In US English, it is typical to use the plural tense for all value counts except the absolute singular.  In other words, we say "zero *matches*", "one-point-five *matches*", but "one *match*".  

I cannot speak for British English because I have never been outside North America, but in both common use and popular media I have repeatedly seen the pattern I described above—and almost never the pattern you described.

sasumner:
@LordNyriox is correct.  "1 match", for all other N -> "N matches"

SinghRajenM:
Very funny!!! I have updated the description now. 😎

PR #5369, opened 2019-03-03
On program update, remove old autoCompletion directory if empty

vlakoff:
As noted in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5277#issuecomment-464397468

ping @SinghRajenM :)

### DIFF
--- a/PowerEditor/installer/nsisInclude/autoCompletion.nsh
+++ b/PowerEditor/installer/nsisInclude/autoCompletion.nsh
@@ -178,6 +178,9 @@ SectionGroup "Auto-completion Files" autoCompletionComponent
 		SetOutPath "$INSTDIR\autoCompletion"
 		File ".\APIs\autoit.xml"
 	${MementoSectionEnd}
+
+	; remove old autoCompletion directory if empty
+	RMDir "$INSTDIR\plugins\APIs\"
 SectionGroupEnd
 
 
@@ -302,7 +305,6 @@ SectionGroup un.autoCompletionComponent
 		Delete "$INSTDIR\plugins\APIs\lua.xml"
 		Delete "$INSTDIR\autoCompletion\lua.xml"
 	SectionEnd
-
 
 	Section un.autoit
 		Delete "$INSTDIR\plugins\APIs\autoit.xml"


SinghRajenM:
@vlakoff In which update scenario autoCompletion directory will  be empty? As of now, I could not think about it.

Say if user already has all/few autocomplete APIs installed and user unchecks all of those autoCompletion APIs during update, there is no provision to remove already installed APIs as of now. So autoCompletion directory will not empty. So I can't think in which scenarios this PR is applicable.


@donho 
> Say if user already has all/few autocomplete APIs installed and user unchecks all of those autoCompletion APIs during update, there is no provision to remove already installed APIs as of now.

This is a bug. During update/reinstall, installer should provide option to remove already installed plugin/themes/autocomplete etc. I have submitted PR  #4704 to handle this scenario (only for plugins as of now). If it sounds good, will be enhanced for other options too.

vlakoff:
I updated Notepad++, with all the autoCompletion boxes checked.

The installer put all new files into `$INSTDIR\autoCompletion`, and removed the corresponding files from `$INSTDIR\plugins\APIs`. As a result the latter got entirely emptied.

SinghRajenM:
> The installer put all new files into $INSTDIR\autoCompletion, and removed the corresponding files from $INSTDIR\plugins\APIs. As a result the latter got entirely emptied.

In this case, user knows what s/he is going to do. I was expecting this may come as answer. 😜 
Moreover, this is not correct/usual approach. User may remove directory, but as you know autocomplete writes memento registry as well (`Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Notepad++`), who is going to remove those then?

So instead of removing them manually. Reinstall/update should allow user to remove already installed components.

vlakoff:
My head already hurts enough

vlakoff:
Something like this?

```nsis
Var autocompleteNSIS
StrCpy $autocompleteNSIS "false"
${MementoSection} "NSIS" NSIS
    StrCpy $autocompleteNSIS "true"
${MementoSectionEnd}

Delete "$INSTDIR\plugins\APIs\nsis.xml"

${If} $autocompleteNSIS == "true"
    SetOutPath "$INSTDIR\autoCompletion"
    File ".\APIs\nsis.xml"
${Else}
    Delete "$INSTDIR\autoCompletion\nsis.xml"
${EndIf}
```

I'm removing the old files by purpose. Better having the newest files than obsolete files. If ever the user has customized a file on its own, he should be aware the installer replaces the content of "%ProgramFiles%\Notepad++".

ping again @SinghRajenM :D

SinghRajenM:
If the intention is only to remove empty autoCompletion directory on update, then [this line](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/157846c60a84558f5ac6e642a1c306542325fce4/PowerEditor/installer/nppSetup.nsi#L266) is the best candidate for statement `RMDir "$installPath\autoCompletion\"`.


> If ever the user has customized a file on its own, he should be aware the installer replaces the content of "%ProgramFiles%\Notepad++".

No, this is not true in all the cases. In case of autoComplete, files are not replaced because of [this](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/157846c60a84558f5ac6e642a1c306542325fce4/PowerEditor/installer/nsisInclude/autoCompletion.nsh#L30).

Above sample code will install api.xml if user checks respective checkbox and if user unchecks, then existing api.xml will be removed. But still autoCompletion directory will not be removed. Moreover, there should be generic/common function which will do the same thing for all the api.xml files instead of having separate code for each api.xml file.

vlakoff:
* Removing directory if empty (can't harm) and deleting files if unchecked are two different issues.
* My PR was for deleting the *old* directory if empty.
* I'd suggest removing that `SetOverwrite off`.
* Indeed, code should be factorised if possible.

vlakoff:
The code factorisation may be something like:
```nsis
Var /GLOBAL autocompleteState

!macro autocompleteFile name file
    StrCpy $autocompleteState "false"
    ${MementoSection} "${name}" name
        StrCpy $autocompleteState "true"
    ${MementoSectionEnd}

    Delete "$INSTDIR\plugins\APIs\${file}"

    ${If} $autocompleteState == "true"
        SetOutPath "$INSTDIR\autoCompletion"
        File ".\APIs\${file}"
    ${Else}
        Delete "$INSTDIR\autoCompletion\${file}"
    ${EndIf}
!macroend

!insertmacro autocompleteFile C "c.xml"
!insertmacro autocompleteFile C++ "cpp.xml"
!insertmacro autocompleteFile Java "java.xml"
!insertmacro autocompleteFile C# "cs.xml"
!insertmacro autocompleteFile HTML "html.xml"
```
My NSIS is rusty (and TBH, NSIS language is shitty), the above code probably has syntax errors (I mainly think of the "name" parameter).

Issue #5368, opened 2019-03-02
Folder paths with double slashes causing issue


melban:
### Description of the Issue
When clicking to open a file and pasting in a full path with double slashes results in locked application
c:\\\\

### Steps to Reproduce the Issue
1. File -> Open
2. type C:\\\\
3. Click enter

### Expected Behavior
Navigates to specified folder to pick files

### Actual Behavior
A loop or unforseen action is happening and locks program.

MetaChuh:
@melban 

i can confirm this issue, but only if you are using the old open/save dialog style.

please go to `settings > preferences > default directory` and make sure that `use new style dialog` is activated, as seen at the screenshot below.

![Imgur](https://i.imgur.com/fibS0kv.png)

please retry, after you have activated the new style dialog.
attempting to open `c:\\` as a file, will now give you an expected, invalid file name alert, instead of looping/freezing.

sasumner:
Still, it should be fixed, for the "old style" case. :)

MetaChuh:
@sasumner 

yes, you are absolutely correct, and i had more or less exactly the same sentence written down already.
(my deleted end quote: "but you're right, it should work on the old style dialog too.")

but i hesitated and removed that, as i don't know yet, if any fixing change at the old style dialogue code might get in conflict with it's functionality on wine platforms.

melban:
@MetaChuh When I enabled "use new style" I do not have the issue and am prompted with the dialog box.  Thank you.

PR #5370, opened 2019-03-03
Bug fix: Open/Save file hangs

SinghRajenM:
Fixed issue: #5368 

If `use new style dialog` (`settings > preferences > default directory`) is not activated and user enters double backslashes (specially due to copy/paste), then npp hangs as message `CDN_FILEOK` is sent infinite times.


PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
```diff
@@ -474,15 +474,19 @@ BOOL APIENTRY FileDialog::run(HWND hWnd, UINT uMsg, WPARAM, LPARAM lParam)
 					{
 						// change to backslash, and insert trailing '\' to indicate directory
 						hFileDlg = ::GetParent(hWnd);
-						std::wstring _fnStr(fileName);
```
SinghRajenM: As `_fnStr` is not member, preceding underscore is removed and it is changed to `filePath` which looks more appropriate as it is path.

PR #5378, opened 2019-03-05
Johab

grizlupo:
Johab is now rarely used, but many older documents are in this charset. So I added a menu and prober to support Johab.


donho:
It's rather a FR for uchardet project.
Please make your change to be merged in uchardet project. 
I'll only include their new release.

chcg:
@grizlupo See https://www.freedesktop.org/wiki/Software/uchardet/ with code at https://cgit.freedesktop.org/uchardet/uchardet/.  Seems https://gitlab.freedesktop.org/uchardet/uchardet/issues could be used also to create a new issue with a request to support johab + https://gitlab.freedesktop.org/uchardet/uchardet/merge_requests to create a PR. Seems there was no request like this one before by searching project bugzilla https://bugs.freedesktop.org/enter_bug.cgi?product=uchardet

grizlupo:
thanks for information. i did it. https://gitlab.freedesktop.org/uchardet/uchardet/merge_requests/1

PR #5379, opened 2019-03-05
Update Turkish localization

Tmp341:
Updated according to 5423b50, 157846c and da3a17e commits.


donho:
Please add also the change:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/28f87ca6dedc44a3862668afc282fb05d1fd60f8

PR #5380, opened 2019-03-05
Added Missing Styles to Deep Black

majorthorn:
Subsection was Deprecated with the introduction of SectionGroup in NSIS 2.05
Added PAGEEX introduced in NSIS 2.0
Added Function Definitions which should map to functiondef
Added Comment box which will add styling to comments between /* and */ comments.


majorthorn:
@donho Here is the fixed PR

donho:
You should create a new branch for each PR

Issue #1116, opened 2015-11-09
Dozens of "File doesn't exist" warnings prevent closing window (even with Update Silently enabled) and shown incorrectly with Network Share


PowerAccess:
## Several Issues with "File doesn't exist" Showing When Shouldn't
1. Dozens of "File doesn't exist" warnings often prevent me from closing Notepad++ window after deleting or moving a folder. Those warnings shouldn't be shown when closing the application.
2. "File doesn't exist" most often get shown when inappropriate or inaccurate, such as for files on a network share, mapped drive, removable drive, or temp files opened from a zip archive, even when the original file does still exist or becomes available again, the prompts still get shown.
3. "File doesn't exist" shouldn't be shown when "Update Silently" is enabled (though another option could be added if felt needed for these). 

If absolutely necessary, such a warning could just be deferred until if/when the user attempts to save the file to disk, or a new option to allow users to show this warning could be added if needed, but most of the complaints and questions about File Status Detection are related to avoiding this annoying warning.

Please note that disabling File-status Detection is not a viable workaround as that is needed for the primary use case of preventing unintended overwriting of files modified on disk.
## Details

As seen in the below screenshot, the "Keep non existing file" warning prompting for "The file {FilePath} doesn't exist anymore. Keep this file in editor?" appears when the file is no longer found on disk (whether lost connection to network share or removed portable drive, or manually moved deleted before going to close Notepad++), sometimes even repeated 2x for such an event, even when "Update silently" preference is enabled. 
## Screenshot

![unnecessary non existing file warnings despite update silently](https://cloud.githubusercontent.com/assets/4053734/11049804/d7a38c12-86f5-11e5-857e-6267904a96c9.png)
## Test Cases & Common Causes

For files on a network share (whether NAS, VPN, FTP, WebDAV, etc.) or even on a thumb drive or portable drive, any time the connection is lost or drive is removed, then this annoying prompt shows up and then another prompt is shown again once the file becomes available again.  T

Also, these warnings occur when opening a temp file such as from a zip archive, once it's no longer available.

Even for local files, this often occurs when I am done with a temp folder or ready to move it after previewing its contents, and then go to close Notepad, and get prompted for nearly every single open tab before I can safely close the app.

[...]

Issue #5386, opened 2019-03-06
Opening file when already opened files are changed selects wrong tab


vwolf010:
### Description of the Issue
Opening file when already opened files are changed selects wrong tab

### Steps to Reproduce the Issue
1. Open np++ and open a file (say a.txt)
2. Change a.txt outside np++ (with another editor)
3. Now open another file (say b.txt)

### Expected Behavior
1. Notification that a.txt has been changed with question if I want to reload
2. Open b.txt in a new tab and jump to the tab of b.txt

### Actual Behavior
1. Notification that a.txt has been changed with question if I want to reload
2. Tab a.txt stays selected and the jump to b.txt does not happen

### Debug Information
n/a

Issue #4747, opened 2018-08-07
Feature request: "Close" and "Yes/No/Close All" options for externally-modified files


Pr0methean:
### Description of the Issue
In my usage, when Notepad++ detects that a file has been modified externally, it usually means that leaving it open up to that point was a mistake. Thus, it'd be handy to have the option to close files for which that'd happened. It'd be even more handy if I only had to answer the prompt once, even when multiple files were affected.

### Steps to Reproduce the Issue
1. Clone https://github.com/Pr0methean/BetterRandom
2. Open several files from that branch in Notepad++
3. Check out the branch java7
4. Open another file

### Expected Behavior
I'm interrupted only once about whether to reload files that have changed externally, and can choose to close them.

### Actual Behavior
I get one prompt for *each* file that's changed externally, and closing the file isn't one of the choice 

### Debug Information
Notepad++ v7.2.2   (64-bit)
Build time : Nov 27 2016 - 16:35:52
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7
Plugins : mimeTools.dll NppConverter.dll

Issue #4592, opened 2018-06-18
A wrong behavior when an open file was updated outside notepad++


dyura:
notepad++ checks file timestamps only when the application is getting focus. The program doesn't verify changes in timestamps right before saving. However, this feature is essential. All advanced editors check timestamps on Save. (Even some editors are monitoring file on disk during editing sessions to inform users as soon as possible when the file was changed outside.)
Lack of this feature leads to two problems.
1. when we are editing or viewing files in notepad++, and any of open files will be modified outside the editor, users are not informed, and notepad++ makes possible to easily overwrite the disk content. It is unacceptable. 
2. When the file was changed outside notepad++, and the program gets focus, the editor notifies users, giving an option to reload the file and to discard changes. However, discarding changes is not what users normally want, and they press Cancel. But, later, when users intend to save, notepad++ doesn't remind that the buffer timestamp is old. All popular editors remind all the time when the file was updated on disk outside the editing session.

PR #5389, opened 2019-03-06
Fixed a long waiting issue about file autto change detection

SinghRajenM:
@donho This mega PR fixes below defects:
#1087
#3349
#3333
#3528
#3138
#3934
#3160
#1822
#745
#3599
#4329
#3971
#4592

Following defects are no longer valid, as these cases will not occur
#5386
#4189
#1995
#2867
#3306
#2776
#2724
#2405
#4072
#4747

**_The list is huge and there are many more similar defects. I will keep on adding them here._**

**Edit:** _Below screenshot is no longer valid as changes were made to accommodate review comments._
image

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -45,12 +45,6 @@ using namespace std;
 #define WM_DPICHANGED 0x02E0
 
 
-DWORD WINAPI CheckModifiedDocumentThread(LPVOID)
-{
-	MainFileManager->checkFilesystemChanges();
-	return 0;
-}
-
 struct SortTaskListPred final
 {
 	DocTabView *_views[2];
@@ -1542,7 +1536,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case NPPM_INTERNAL_CHECKDOCSTATUS:
 		{
 			const NppGUI & nppgui = pNppParam->getNppGUI();
-			if (nppgui._fileAutoDetection != cdDisabled)
+			if (nppgui._fileAutoDetection & cdEnabled)
 			{
 				checkModifiedDocument();
 				return TRUE;
@@ -1621,15 +1615,15 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_INTERNAL_ENABLECHECKDOCOPT:
 		{
-			NppGUI & nppgui = const_cast<NppGUI &>((pNppParam->getNppGUI()));
+			NppGUI& nppgui = const_cast<NppGUI&>((pNppParam->getNppGUI()));
 			if (wParam == CHECKDOCOPT_NONE)
 				nppgui._fileAutoDetection = cdDisabled;
 			else if (wParam == CHECKDOCOPT_UPDATESILENTLY)
-				nppgui._fileAutoDetection = cdAutoUpdate;
+				nppgui._fileAutoDetection = (cdEnabled | cdAutoUpdate);
 			else if (wParam == CHECKDOCOPT_UPDATEGO2END)
-				nppgui._fileAutoDetection = cdGo2end;
+				nppgui._fileAutoDetection = (cdEnabled | cdGo2end);
 			else if (wParam == (CHECKDOCOPT_UPDATESILENTLY | CHECKDOCOPT_UPDATEGO2END))
-				nppgui._fileAutoDetection = cdAutoUpdateGo2end;
+				nppgui._fileAutoDetection = (cdEnabled | cdGo2end | cdAutoUpdate);
 
 			return TRUE;
 		}
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3783,25 +3783,38 @@ bool Notepad_plus::activateBuffer(BufferID id, int whichOne)
 			return false;
 	}
 
+	bool isFileChangeDetection = (NppParameters::getInstance()->getNppGUI()._fileAutoDetection & cdEnabled) ? true : false;
+
 	if (reload)
 	{
 		performPostReload(whichOne);
+	}
+	else if(isFileChangeDetection)
+	{
+		// Buffer has been activated, now check for file modification
+		// If enabled to change detection
+		pBuf->checkFileState();
 	}
 
 	notifyBufferActivated(id, whichOne);
 	return true;
 }
 
-void Notepad_plus::performPostReload(int whichOne) {
-	NppParameters *pNppParam = NppParameters::getInstance();
-	const NppGUI & nppGUI = pNppParam->getNppGUI();
-	bool toEnd = (nppGUI._fileAutoDetection == cdAutoUpdateGo2end) || (nppGUI._fileAutoDetection == cdGo2end);
+void Notepad_plus::performPostReload(int whichOne)
+{
+	NppParameters* pNppParam = NppParameters::getInstance();
+	const NppGUI& nppGUI = pNppParam->getNppGUI();
+	bool toEnd = (nppGUI._fileAutoDetection & cdGo2end) ? true : false;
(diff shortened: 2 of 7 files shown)


SinghRajenM:
Fixes:
#1116
#2082

No longer valid:
#4180

sasumner:
Big big thumbs up...if if gets accepted/integrated.  Otherwise, meh. :)

LordNyriox:
@SinghRajenM:  **Grammar Warning:** 

Though not necessarily incorrect in English, the article in "all the tabs" suggests that there is a *specific set* of tabs being referenced—rather than a generic global definition. 

Perhaps you can change the option text to read "for all tabs" instead?

LordNyriox:
@SinghRajenM:  **Grammar Warning:** 

Though not necessarily incorrect in English, the article in "`all the tabs`" suggests that there is a *specific set* of tabs being referenced—rather than a generic global definition. 

Perhaps you can change the option text to read "`for all tabs`" instead? 

**P.S.**  Also note that there is no article in the label for the newly added option "for current tab".  Please either include "the" in both labels, or neither of them.

LordNyriox:
@SinghRajenM:  **Grammar Warning:** 

Though not necessarily incorrect in English, the article in "`all the tabs`" suggests that there is a *specific set* of tabs being referenced—rather than a generic global definition. 

Perhaps you can change the option text to read "`for all tabs`" instead? 

**P.S.**  Also note that there is no article in the label for the newly added option "`for current tab`".  Please either include "`the`" in both labels, or neither of them.

sasumner:
@LordNyriox I see "for all the tabs" in the image above; maybe the image was changed AFTER your comment?

SinghRajenM:
@sasumner No. No change till now.

As per @LordNyriox 
either remove "the" from "Enable for all ~~the~~ tabs" 
or add "the" in "Enable only for **the** current tab".

sasumner:
IMO the best thing would be to drop "the" in both.

wqweto:
Can you please merge #5273 in this PR (a simple 5 lines change) as it's totally ignored so far by the maintainer.

The proposed fix allows to **undo** the automatic/manual file reloading. We are humans making mistakes and can confirm file reload even though we don't want to lose our unsaved modifications. Let's give a chance to undo such mistake!

SinghRajenM:
> IMO the best thing would be to drop "the" in both.

Done now.

>Can you please merge #5273 in this PR (a simple 5 lines change) as it's totally ignored so far by the maintainer.

@wqweto It is better to keep #5273 separate to adhere [Contributing Policy](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests). Also as I can @donho  has assigned #5273 to himself. Once he will get time, he will review and take action. I would say, don't worry, it is not ignored, just taking time.

Tatsuran:
@SinghRajenM thank you so much. Waiting for update with this fix a lot, switching tabs issue driving me nuts

donho:
@SinghRajenM  What is the new behaviour? A detail description will help to integrate your PR.

SinghRajenM:
@donho 
Please find the description as below - 
1. **If new checkbox is not checked, then old behavior is retained. No change at all.** It will help those users who want to detect changes for all the opened files. So it is retained as it is.
2. If new checkbox is checked, then
     a. If file change will be detected only for active tab. So user will see only one messagebox for reload. If file other than active tab is modified, then user will be see reload message only when s/he activates that particular tab.
    b. Other two options are same as they were working before for all the tab (but now will work for current tab too.). Like if user check "update silently", then the active will be updated silently without showing messagebox. Same for "Go to end" and combination of both.

From UI perspective:
1. "Update silently" and "Scroll to the end after update" will be enabled only if either one of options (Enable for all tabs or Enable for current tab) is selected.
2. At a given time, user can choose "Enable for all tabs" or "Enable for current tab" or None of them.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1621,15 +1617,15 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_INTERNAL_ENABLECHECKDOCOPT:
 		{
-			NppGUI & nppgui = const_cast<NppGUI &>((pNppParam->getNppGUI()));
+			NppGUI& nppgui = const_cast<NppGUI&>((pNppParam->getNppGUI()));
```
SinghRajenM: @donho 
**Additional note:**
I feel "Doc Monitor" plugin developer needs to intimated. With this current implementation, new functionalities can not be enabled. So this plugin developer should be informed and we may need to provide additional support to him/her if s/he is actively updating plugin.

donho:
Good job.
It seems to me the new behaviour is better than the old one.
Let's remove the old behaviour and make new one as standard behaviour.

SinghRajenM:
@donho Updated the PR as per review comments.

bahusoid:
Please confirm that Save All operation does nothing for unchanged tabs, so it's safe to click it even if some tabs were changed externally.

And what about "Replace All in All Opened Documents" operation? Does it mean that with this new behavior replace will be silently applied to old document and external changes are lost on Save All?

SinghRajenM:
@bahusoid 
> Please confirm that Save All operation does nothing for unchanged tabs, so it's safe to click it even if some tabs were changed externally.

Yes, save all will not impact unchanged tab as well as externally modified file.


>And what about "Replace All in All Opened Documents" operation?

This is behavioral decision here. 
**If only file changes detection is enabled**, "Replace All in All Opened Documents" or "File all in All opened Doc" will be performed on current state of all the tab. That means if something is externally modified will not be considered. **But, if "Update silently"** is also enabled then externally modified changes will be considered during "Replace All in All Opened Documents" and "File all in All opened Doc" operations. So it is all user's choice.

bahusoid:
> This is behavioral decision here.
>If only file changes detection is enabled, "Replace All in All Opened Documents" or "File all in All opened Doc" will be performed on current state of all the tab. 

But it's not desired behavior. I want to be informed on changes. And when I do "Replace All" I want to be sure that this operation is done on up to date files.
With this pr it's either silent update or  silent replace on outdated file  - not good.

donho:
@SinghRajenM What @bahusoid pointed out is an expected behaviour but hard to implement.
To prevent from users' complain about unexpected side effect of the new behaviour, I will keep both old and new behaviours. So users can switch back the old one if they want. You don't need to do anything on this PR, I kept your previous one on my local branch.

Thank you for this helpful improvement.

SinghRajenM:
@donho 

> @SinghRajenM What @bahusoid pointed out is an expected behaviour but hard to implement.

I think I found the solution for this issue as well. Likely one line solution 😜 

``` 
bool Notepad_plus::replaceInOpenedFiles() {
......
.....
if (pBuf->isReadOnly())
				continue;

//These two line are newly added
			if(!pBuf->isDirty())
				pBuf->checkFileState();
```

Say there are 5 tabs opened.
Tab 1: Saved
Tab 2: Externally modified
Tab 3: Externally modified
Tab 4: A dirty tab (may be saved, but dirty)
Tab 5: A saved dirty tab, but it is modified outside as well.

Currently, I'm on Tab 1. I do replace all in opened tab.

Result:
**Tab 1**: replacement will happen without any issue.
**Tab 2**: It will ask for confirmation to reload
     a) if user says yes to reload, then file will be reloaded and then replacement will happen.
     b) if user says no to reload, then replacement will happen for old data
**Tab 3**: Unless user come out of tab 2, it will be waiting for its turn. Once tab 2 operation is complete, replacement will happen for this tab as per 2.a and 2.b
**Tab 4**: After tab 3, replacement will happen on this tab directly.
**Tab 5**: After tab 4, replacement will happen for this tab as well _but with old data_.

Please note: User can do <kbd>Ctrl</kbd> + <kbd>z</kbd> for all the 5 tabs.

I feel this is expected. I feel this can be applied to find operation as well. I will check in sometime today.

bahusoid:
@SinghRajenM Sounds good! But that makes  "Do for all" request relevant again :)

PR #5397, opened 2019-03-08
zulu and kurdish translation fix

chcg:
- corrected xml encoding from UTF8 with BOM -> required UTF8 for zulu and kurdish
- added zulu to localization list, see also #5209
- space -> tab in localizationString.h according to the rest of the list

### DIFF
--- a/PowerEditor/installer/nativeLang/kurdish.xml
+++ b/PowerEditor/installer/nativeLang/kurdish.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8" ?>
 <NotepadPlus>
     <Native-Langue name="كوردی" RTL="yes" filename="kurdish.xml" version="7.5">
         <Menu>
--- a/PowerEditor/installer/nativeLang/zulu.xml
+++ b/PowerEditor/installer/nativeLang/zulu.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8" ?>
 <!--
 ***********************************************************************************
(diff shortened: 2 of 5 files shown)


PowerEditor/installer/nativeLang/kurdish.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8" ?>
```
donho: What's the change of this line?

PowerEditor/installer/nativeLang/zulu.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8" ?>
```
donho: Same question as above one.

chcg: Changed from UTF8 with BOM -> plain UTF8 otherwise the switch to this language is not working.

PowerEditor/installer/nativeLang/kurdish.xml
chcg: Changed from UTF8 with BOM -> plain UTF8 otherwise the switch to this language is not working.

PowerEditor/installer/nativeLang/zulu.xml
donho: Good job then

donho: @chcg could you merge a commit properly?
I can always do it, but if you can, it'll be better.

chcg:
Sorry I was to late for merging it on my own.

donho:
@chcg  No problem :)

PR #5406, opened 2019-03-11
Ignore deprecation warning for GetVersionEx

stevoisiak:
GetVersionEx() was depreciated in Windows 8.1. (See https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getversionexw)


donho:
The concerning warning has been disabled.

PR #5411, opened 2019-03-12
Remove redundant call to GetVersionEx

stevoisiak:
Simplies logic when checking Windows version.


donho:
The code was there is for the compatibility.

stevoisiak:
> The code was there is for the compatibility.

I’m not sure I understand. Is there a version of Windows that requires two calls to GetVersionEx?

PR #5422, opened 2019-03-14
Delete autoCompletion folder on uninstall

SinghRajenM:


PowerEditor/installer/nsisInclude/uninstall.nsh
```diff
@@ -54,7 +54,7 @@ Function un.onUninstSuccess
 	RMDir "$INSTDIR\"
 
 	RMDir "$APPDATA\${APPNAME}\plugins\"
-	RMDir "$installPath\autoCompletion\"
+	RMDir "$INSTDIR\autoCompletion\"
```
donho: Should this line be before `RMDir "$INSTDIR\"` ?

SinghRajenM: @donho Yes, you are right. I could not catch it because in my testing NppShell_06.dll was locked. So Notepad++ directory was not deleted. I should have not assumed only NppShell_06.dll as reason.
I have fixed it now.

However, It is good to remove NppShell_06.dll as well on installed. So that Notepad++ dir also will be removed.

PowerEditor/installer/nsisInclude/uninstall.nsh
```diff
@@ -271,8 +271,8 @@ Section Uninstall
 		Delete "$APPDATA\${APPNAME}\userDefineLangs\userDefinedLang-markdown.default.modern.xml"
 		RMDir /r "$APPDATA\${APPNAME}\plugins\"
 		RMDir /r "$APPDATA\${APPNAME}\backup\"	; Remove backup folder recursively if not empty
-		RMDir "$APPDATA\${APPNAME}\themes\"	; has no effect as not empty at this momenet, but it is taken care at un.onUninstSuccess
-		RMDir "$APPDATA\${APPNAME}"		; has no effect as not empty at this momenet, but it is taken care at un.onUninstSuccess
+		RMDir "$APPDATA\${APPNAME}\themes\"	; has no effect as not empty at this moment, but it is taken care at un.onUninstSuccess
+		RMDir "$APPDATA\${APPNAME}"		; has no effect as not empty at this moment, but it is taken care at un.onUninstSuccess
```
LordNyriox: @SinghRajenM:  If you are going to correct the spelling of a comment, you should also correct the *wording* of a comment. 

>it is taken care at 

…should be… 

>it is **dealt with** at 

"Taken care" is not a valid English phrase.  "Taken care of" is the correct form of the phrase, but it would be awkward in this particular sentence.  ("it is **taken care of** at un.onUninstSuccess")

chcg:
Appveyor build failure:
> Fallback clouds stack is too big.

Issue #4566, opened 2018-06-06
Percentage of zooming in and out


wogns7534:
### Description of the Issue
When expanding or shrinking in Notepad ++, it would be better to display the percentage in the lower right corner as in Word. There is a key to restore the default, but it would be more convenient if it represented a percentage.

### Steps to Reproduce the Issue
1. Write the code for zooming in and out.
2. Apply it to Notepad ++.
3. It is possible to enlarge and reduce more conveniently than before.

### Expected Behavior
It is more convenient to zoom in and out without having to use the default restore key.

### Actual Behavior
The percentage of zoom in and out is displayed in the lower right corner (next to UTF-8).

### Debug Information
![image](https://user-images.githubusercontent.com/37335948/41055027-17d38180-69fb-11e8-8eab-63812c8391fe.png)

sasumner:
The "zoom" is controlled by the Scintilla editing component so there isn't much Notepad++ development can do to (reasonably) change it.  

As to displaying the zoom level in the status bar, you might see this: https://notepad-plus-plus.org/community/topic/15679/seeing-the-zoom-level-with-pythonscript
That could be probably be adapted to display some sort of percentage but you'd have to figure out a reasonable algorithm for that.

donho:
We have obvious way to zoom in/out on UI. Not on status bar anyway.

Issue #1268, opened 2015-12-19
Character position not displayed in status bar


badon:
If I need to debug string manipulation on a text file, I have to use workarounds because Notepad++ does not display character position in the status bar. String manipulation is universally done with character position numbers (AKA "index position"). Notepad++ only displays line and column numbers. 

There are 2 workarounds for this missing feature. The first one is to begin selecting text at the first character position of the file, and then use the count of selected characters to determine the cursor's character position. The other workaround is to use the "Go to..." function in the search menu or CTRL+G, and then select "Offset" and enter the character position you want to go to. 

Both workarounds are inadequate given the importance of knowing the character position, so I request that character position be displayed in the status bar.

taxi16:
I just have the same problem and totally agree with the judgement of the relevance of that enhancement request. I was really astonished that it is not available (as Notepad++ for me is by far the best tool in it's domain, not to write letters, but for code development :) ). So adding that would be great!

Huud:
You can put the cursor over where you want to find the offset, and then double click on the (Ln, Col, Sel) section on the bottom bar to get a popup window that contains your current offset.

But yeah, an option to display the offset directly on the bar instead of the (Sel) when nothing is selected would be ideal.

MetaChuh:
we are closing this issue due to multiple newer notepad++ versions since it was filed.

please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins.
if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as detailed and reproducible as possible.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
search if related topics exists. you are welcome to post either at similar
topics, or to create a new topic at Help Wanted or General Discussion.
```
[...]

PR #5434, opened 2019-03-17
PR to discuss about zoom value on status bar

SinghRajenM:
Hi @donho and All,

I am trying to implement a feature to show zoom value on status bar which was discussed on community and issue tracker.
https://notepad-plus-plus.org/community/topic/15679/seeing-the-zoom-level-with-pythonscript
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4566
https://notepad-plus-plus.org/community/topic/17321/feature-request-current-zoom-to-status-bar

I was able to reach half way. Current solution is partially working as it works on click on zoom in/out from toolbar button and menu item (View->Zoom->options). It is not working for shortcuts e.g. <kbd>Ctrl</kbd>+<kbd>+</kbd>, or <kbd>Ctrl</kbd>+ <kbd>mouse scroll</kbd> etc.
But now I'm stuck. I'm not able to figure out how scintilla shortcut works. Like how scintilla accelerator works, which function is called or which event/message is sent? A help to understand this would be appreciated. 

image


sasumner:
IMO Zoom is not important enough to waste valuable status bar space on. :)

Ekopalypse:
Scintilla shortcut is CTRL and +(plus sign from numpad) and notification is [SCN_ZOOM](https://www.scintilla.org/ScintillaDoc.html#SCN_ZOOM)

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2346,6 +2351,28 @@ void Notepad_plus::setLangStatus(LangType langType)
 	_statusBar.setText(getLangDesc(langType).c_str(), STATUSBAR_DOC_TYPE);
 }
 
+void Notepad_plus::setPercentStatus(int whichOne, int percent)
```
MAPJe71: Renaming parameter `whichOne` to `whichView` would make its purpose more clear IMO.

Ekopalypse:
@SinghRajenM 
I don't want to hijack this PR with another request but as it belongs to statusbar and there is already
a request open #1268 I wanted to take the chance to ask if this could be implemented as well.
From my point of view, when working with scintilla this one would be quite nice to have.
I know, you can use goto window to see where am I but I think it would be even nicer to have it,
maybe after len:... line:..., visible in the statusbar.
If you disagree with it - don't mind and ignore this request.
Thank you

PowerEditor/src/Notepad_plus.cpp
donho: What's your suggestion @MAPJe71 ?

donho:
@SinghRajenM 
I second what @sasumner said.

SinghRajenM:
Cool!!!

@donho but there should be way to know current level, isn't it?

donho:
@SinghRajenM  It could be "nice to have", but it's not a mandatory info IMO. 
This info could be on both tooltip of zoom - if the implementation is trivial. Otherwise let's focus our coding effort on more important feature and critical bug.

n3tcom:
May I ask, on the record, is it the official position of Notepad++ that this critical accessibility feature is not of importance?

sasumner:
>May I ask, on the record, is it the official position of Notepad++ that this critical accessibility feature is not of importance?

Official position, on the record:  Not important.

kipusoep:
What is wrong with giving people the power to decide for themselves if this is valuable or not?
Implement the feature and make it optional (opt-in) would make both sides happy I'd say...

PR #5441, opened 2019-03-20
Fixed multi cursor bugs

jmbockhorst:
Thanks a lot to @MetaChuh for testing the multi cursor feature and finding some bugs. I have fixed a few issues:

- Using enter and backspace repeatedly caused the cursors to be incorrect and get jumbled.
- Multi cursors functions did not work with LF (Unix) or CR (Macintosh) line endings.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1636,21 +1636,49 @@ void Notepad_plus::command(int id)
 			// Get the number of selection
 			int numSelections = int(_pEditView->execute(SCI_GETSELECTIONS));
 
+			if (numSelections == 0)
+			{
+				_pEditView->execute(SCI_NEWLINE);
+				break;
+			}
+
+			// Assume line endings is CR LF
+			int selectionMultiplier = 2;
+
+			// Check if line endings is LF or CR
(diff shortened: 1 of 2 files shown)


sasumner:
I for one am nervous about this now. :)
Why can't we just get the functionality via a Scintilla upgrade?

jmbockhorst:
One thing to note. When there is only one cursor (so for most users), it still uses the default Scintilla functions. So there is no worries about anything breaking for the normal user.
Also, I feel like now (after these fixes), this is working pretty well, and is a useful upgrade for those who could use it. But please test it for yourself if you are still worried about it.

MetaChuh:
@jmbockhorst 
cc @donho as testing report

* [x] all new multi cursor editing bugs i have found, including the crashes, were re-tested and are successfully fixed with this pull. 👍

* [x] i was also not able to find any further issues for now, even when switching live between eol modes when multiple cursors are already set, and even if multi-cursors are shifted left or right across line endings.
(see combined screencast of the fixed version below, tested on both x86 and x64)

Imgur

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/1360ff82ba995e0464abb79dc2525b566b8e5921

@sasumner  Yes, Scintilla SHOULD be (and WILL be) updated in short term.

PR #5453, opened 2019-03-24
Fixed detection of shorcut key conflicts when filtering

### LINKED ISSUE #5374: shortcut mapper conflict doesn't display right when filtered
### Description of the Issue

When the shortcut mapper view is filtered, it doesn't display a keystroke conflict correctly.  (See [this Community post](https://notepad-plus-plus.org/community/topic/17222/nppexec-f6-not-working/5) for context)

1. Have a conflict with one of the plugin shortcuts.  In my example, both **Main Menu [3]: `Explorer`** and **Plugin Commands: NppExec.dll `Execute...`** were mapped to `F6`
2. In Shortcut Mapper, on the Plugin commands tab, scroll normally to `Execute...` and see it's red, and the message box shows `Main menu  |  3   Explorer  ( F6 )`.  This is expected behavior.  
3. Filter by `Execute`, then click on one of the others (like "save all files on execute") then back on the red-highligted `Execute...` row: the message box now shows `No shortcut conflicts for this item.`, even though there is still a conflict, and it's properly highlighted red. 

### Expected Behavior

Showing `Main menu  |  3   Explorer  ( F6 )` whether the filter is applied or not

### Actual Behavior

Correctly shows `Main menu  |  3   Explorer  ( F6 )` when no filter applied, but claims `No shortcut conflicts for this item.` when there is a filter applied

### Debug Information

```
Notepad++ v7.6.3   (64-bit)
Build time : Jan 27 2019 - 17:16:47
Path : C:\usr\local\apps\notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll DSpellCheck.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppFTP.dll PythonScript.dll XMLTools.dll 
```

jmbockhorst:
Currently, when the shortcut mapper is filtering shortcuts, it doesn't display key conflicts properly. This is due to the indexes of the filtered shortcuts being different (since there are less) than the original list of shortcuts. Fixes #5374.

### DIFF
--- a/PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
+++ b/PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
@@ -1062,37 +1062,37 @@ INT_PTR CALLBACK ShortcutMapper::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 								case STATE_MENU:
 								{
 									vector<CommandShortcut> & vShortcuts = nppParam->getUserShortcuts();
-									findKeyConflicts(&conflictInfo, vShortcuts[currentIndex].getKeyCombo(), currentIndex);
+									findKeyConflicts(&conflictInfo, vShortcuts[_shortcutIndex[currentIndex]].getKeyCombo(), _shortcutIndex[currentIndex]);
 								}
 								break;
 
 								case STATE_MACRO:
 								{
 									vector<MacroShortcut> & vShortcuts = nppParam->getMacroList();
-									findKeyConflicts(&conflictInfo, vShortcuts[currentIndex].getKeyCombo(), currentIndex);
+									findKeyConflicts(&conflictInfo, vShortcuts[_shortcutIndex[currentIndex]].getKeyCombo(), _shortcutIndex[currentIndex]);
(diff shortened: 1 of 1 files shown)


sasumner:
The changes to the cpp file look like roughly the same dozen line repeated four different times.  Couldn't that have been made into a function to consolidate, and avoid possible future changes missing some of these sections?  Just askin.

jmbockhorst:
Yes I know, it’s not ideal. There is a good amount of code in this area that is repetitive as well, but consolidating it is a bit tricky. I am working on it.

sasumner:
A sure way to get a PR rejected is to change existing code for such a reason.  So I'm suggesting it only for code you are putting in. :)

jmbockhorst:
It turns out, the problem was a lot simpler than I had originally thought. I updated it with the easy fix. Thanks for the nudging me to look into it closer. :)

donho:
@jmbockhorst 
Could you solve the conflict please?

PR #5455, opened 2019-03-24
Fix functionList.xml for php

WinterSilence:
Add `# comment` + foreach contruction.

### DIFF
--- a/PowerEditor/src/functionList.xml
+++ b/PowerEditor/src/functionList.xml
@@ -110,7 +110,7 @@
 			<parser
 				id         ="php_syntax"
 				displayName="PHP"
-				commentExpr="(?s:/\*.*?\*/)|(?m-s://.*?$)"
+				commentExpr="(?s:/\*.*?\*/)|(?m-s://.*?$)|(?m-s:#.*?$)"
 			>
 				<classRange
 					mainExpr    ="^\s*(class|abstract\s+class|final\s+class|interface|trait)[\t\x20]+[A-Za-z_\x7F-\xFF][\w\x7F-\xFF]*(\s*|\s*(extends|implements|(extends\s+(\\|[A-Za-z_\x7F-\xFF][\w\x7F-\xFF]*)+\s+implements))\s+(\,\s*|(\\|[A-Za-z_\x7F-\xFF][\w\x7F-\xFF]*))+\s*)?\{"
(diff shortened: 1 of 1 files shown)


chcg:
Seems n++ function list changes are still blocked by https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3393 , as no tests are in place yet.

PowerEditor/src/functionList.xml
```diff
@@ -123,21 +123,21 @@
...
 					</function>
 				</classRange>
 				<function
 					mainExpr="^\s*function\s+\w+\s*\("
 				>
 					<functionName>
-						<nameExpr expr="(?!(if|while|for))\w+\s*\(" />
-						<nameExpr expr="(?!(if|while|for))\w+" />
+						<nameExpr expr="(?!(if|while|for|foreach))\w+\s*\(" />
+						<nameExpr expr="(?!(if|while|for|foreach))\w+" />
```
vlakoff: Do you know why `switch` is missing here? Probably an overlook, isn't it?

PowerEditor/src/functionList.xml
```diff
@@ -123,21 +123,21 @@
 						<nameExpr expr="\w+\Z" />
 					</className>
 					<function
-						mainExpr="^\s*((static|public|protected|private|final)*(\s+(static|public|protected|private|final))+\s+)?(function\s+)+(\w+(\s+\w+)?(\s+|\*\s+|\s+\*|\s+\*\s+))?(\w+\s*::)?(?!(if|while|for|switch))[\w~]+\s*\([^\{]*\{"
+						mainExpr="^\s*((static|public|protected|private|final)*(\s+(static|public|protected|private|final))+\s+)?(function\s+)+(\w+(\s+\w+)?(\s+|\*\s+|\s+\*|\s+\*\s+))?(\w+\s*::)?(?!(if|while|for|switch|foreach))[\w~]+\s*\([^\{]*\{"
 					>
 						<functionName>
-							<funcNameExpr expr="(?!(if|while|for|switch))\w+\s*\([^\{]*" />
+							<funcNameExpr expr="(?!(if|while|for|switch|foreach))\w+\s*\([^\{]*" />
```
vlakoff: Should we also add [`do`](https://www.php.net/manual/en/control-structures.do.while.php)?

vlakoff: However, I can't reproduce a false positive happening with these structures.

Could you please provide an example showing the need for these regex negative assertions?

PowerEditor/src/functionList.xml
```diff
@@ -123,21 +123,21 @@
 						<nameExpr expr="\w+\Z" />
 					</className>
 					<function
-						mainExpr="^\s*((static|public|protected|private|final)*(\s+(static|public|protected|private|final))+\s+)?(function\s+)+(\w+(\s+\w+)?(\s+|\*\s+|\s+\*|\s+\*\s+))?(\w+\s*::)?(?!(if|while|for|switch))[\w~]+\s*\([^\{]*\{"
+						mainExpr="^\s*((static|public|protected|private|final)*(\s+(static|public|protected|private|final))+\s+)?(function\s+)+(\w+(\s+\w+)?(\s+|\*\s+|\s+\*|\s+\*\s+))?(\w+\s*::)?(?!(if|while|for|switch|foreach))[\w~]+\s*\([^\{]*\{"
```
vlakoff: (yet) outside the scope of this PR, but I guess the `(\s+|\*\s+|\s+\*|\s+\*\s+)` could be shortened to `(\s+)`, as there can't be such `*` in PHP. Probably the rule has been copied from another language.

vlakoff: Same with the double colon `::`, that can occur in static call but not in function definition.

WinterSilence:
@vlakoff pls, if u want change something send PR

PowerEditor/src/functionList.xml
WinterSilence: @vlakoff  function list in `futures` - not all work in curret version. See https://notepad-plus-plus.org/features/function-list.html

donho:
@WinterSilence  Could you provide a sample php code to show your enhancement?

sasumner:
@donho said

>@WinterSilence Could you provide a sample php code to show your enhancement?

An "issue" would have been the perfect place to discuss this and show what a possible enhancement could have done.  But the protocol of "issue first, then PR" seems to have been skipped for this case.

Perhaps priority of PR integration consideration should be given to clearly described issues and well-done PRs.  That way, the people that put the effort in, see the fruits of their work integrated sooner.  :-)

donho:
Just gave some tries, cannot see any point for this PR.
Please provide the samples in an issue and reopen it.

Issue #5400, opened 2019-03-08
UDL: If a user defined language's name contains () brackets, like Markdown (Default), the language bullet will not be set correctly to that specific udl.


MetaChuh:
### Description of the Issue

if a user defined language contains () brackets in its name, for example `Markdown (Default)`, the language bullet will not be set correctly.

### Steps to Reproduce the Issue

1. open up any file.
2. go to the `language` menu and select `Markdown (Default)` or any other udl that has `()` brackets in it's name.
3. go to the `language` menu again and watch where the bullet, which defines the current language, is placed.

### Expected Behaviour

the language bullet is expected to be at the currently selected udl, e.g. `Markdown (Default)`.

### Actual Behaviour

the language bullet is shown at the `User-Defined` template udl, as shown at the screenshot below:

![Imgur](https://i.imgur.com/KcljFhS.png)

### Debug Information

Notepad++ v7.6.4   (32-bit)
Build time : Mar  6 2019 - 02:52:22
Path : C:\Users\User\Desktop\npp.7.6.4.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

**note**: this behaviour apparently dates back and applies to all older versions which have an udl editor.
i've only discovered this recently by coincidence, as the latest `Markdown (Default)` udl contains `()` brackets in its name, my apologies for that.

**additional note** this issue might also be related to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4714

predicted issue priority for the majority of users: low.
predicted issue priority for power users: low to medium.

MetaChuh:
**update and more apologies**:

i have just retested issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4714 , and the `Markdown (Default)` issue is **solely based on an udl name of 15 characters or more.**
the `()` brackets are not the cause, sorry about my premature assumption.

bpersonick:
Hey Meta, I'm glad I was able to help you determine the root cause of your issue by having described mine and my testing of it.  Hopefully someone gives these a look. :)

MetaChuh:
and i'm glad you write the way you do, so it was very easy to find, very easy to read and relate to the issue i've encountered while testing the deployment issues of the now distributed `Markdown (Default)` udl.

yes, hopefully someone gives these a look.
[...]

Issue #4714, opened 2018-07-30
Language Menu does not highlight active UDL, when name greater than 14 characters, no syntax highlighting for UDL when name greater than 21 characters.


bpersonick:
### Description of the Issue
The User-defined language interface allows UDLs to be saved with names up to 21 characters.

Importing an XML through the UDL menu allows seems to have no small limit on the number of characters.

 in all instances, the Language menu displays the UDL Names exactly as defined in the XML or through saving them.

 However, if a UDL name is 15 characters or greater, the language menu will not show that language as selected and instead will show the generic "User-Defined" as selected.

Furthermore, if a User-defined Language has 22 or greater characters in its name NP++ will not display the syntax highlighting when selecting it.

### Steps to Reproduce the Issue
Issue A:
1. Save a known working UDL as a name greater than 15 characters
2. Open Language menu, Language displays as expected.
3. Select that language name
4. Syntax Highlighting is applied as expected
5. Look in Language menu, "User-Defined" is highlighted instead of UDL name.

Issue B:
1. Export a known working UDL
2. Edit the XML to have a name longer than 22 characters
3. Import the XML for the UDL through the UDL interface (or add it by editing the xml directly)
4. Open Language menu, Language displays as expected.
5. Select that language name
6. Syntax Highlighting is NOT applied as expected, instead default "User-Defined" style is applied.
7. Look in Language menu, "User-Defined" is highlighted instead of UDL name.

### Expected Behavior
For any length name I can save through the interface, I would expect it to apply and to show as selected when applied.

When importing longer names, I woudl expect we shodul have a higher max, but if this is not t possibel I would expect NP++ to indiate that it needs to have the name shortened to 21 characters and asking you to change the name if needed, and so long as there not more than say 5 instances which woudl need to be changed (if mor eit shodl fail noting to edit the XMl and shorten all names longer than 21 characters).

However I woudl much prefer to have names that are a full int (255 characters) or at least 50+ characters as a more resonable amount.  I have run into this issue int he past but lways though i had a malformed UDL frustrating to know I have wasted effort it he past.

### Debug Information

Notepad++ v7.5.6   (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)
[...]

PR #5458, opened 2019-03-26
Fixed issue and allow some more characters in UDL name

SinghRajenM:
Fixed issue #5400 and #4714

UDL name length has been extended to 64 characters as it is required. examples are as below - 

Generic printer Description language (GPD)
Generic Description Language (GDL)
page description language (PDL)
Printer Command Language (PCL)
PostScript Printer Description (PPD)

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2086,18 +2086,19 @@ void Notepad_plus::checkLangsMenu(int id) const
 			if (curBuf->isUserDefineLangExt())
 			{
 				const TCHAR *userLangName = curBuf->getUserDefineLangName();
-				const int nbChar = 16;
-				TCHAR menuLangName[nbChar];
+				TCHAR menuLangName[langNameLenMax];
 
 				for (int i = IDM_LANG_USER + 1 ; i <= IDM_LANG_USER_LIMIT ; ++i)
 				{
-					if (::GetMenuString(_mainMenuHandle, i, menuLangName, nbChar-1, MF_BYCOMMAND))
+					if (::GetMenuString(_mainMenuHandle, i, menuLangName, langNameLenMax, MF_BYCOMMAND))
+					{
 						if (!lstrcmp(userLangName, menuLangName))
 						{
 							HMENU _langMenuHandle = ::GetSubMenu(_mainMenuHandle, MENUINDEX_LANGUAGE);
 							doCheck(_langMenuHandle, i);
 							return;
 						}
+					}
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2086,18 +2086,19 @@ void Notepad_plus::checkLangsMenu(int id) const
...
 			{
 				const TCHAR *userLangName = curBuf->getUserDefineLangName();
-				const int nbChar = 16;
-				TCHAR menuLangName[nbChar];
+				TCHAR menuLangName[langNameLenMax];
 
 				for (int i = IDM_LANG_USER + 1 ; i <= IDM_LANG_USER_LIMIT ; ++i)
 				{
-					if (::GetMenuString(_mainMenuHandle, i, menuLangName, nbChar-1, MF_BYCOMMAND))
+					if (::GetMenuString(_mainMenuHandle, i, menuLangName, langNameLenMax, MF_BYCOMMAND))
```
chcg: See https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-getmenustringa

> Note  The GetMenuString function has been superseded. Use the GetMenuItemInfo function to retrieve the menu item text.

But the really relevant issue is:

> Remarks
> 
> The nMaxCount parameter must be one larger than the number of characters in the text string to accommodate the terminating null character.
> 
> If nMaxCount is 0, the function returns the length of the menu string.
> Security Warning
> The lpString parameter is a TCHAR buffer, and nMaxCount is the length of the menu string in characters. Sizing these parameters incorrectly can cause truncation of the string, leading to possible loss of data.

, so the necessary string buffer length could be retrieved dynamically, if wanted.
I think also langNameLenMax still needs the -1 as before.

SinghRajenM: Hmmm... Same link says: (also during debug below points stand true). Sometime MS documentation is confusing...

>Type: LPTSTR
The buffer that receives the null-terminated string. **If the string is as long or longer than lpString, the string is truncated and the terminating null character is added**. If lpString is NULL, the function returns the length of the menu string.
cchMax
Type: int
The maximum length, in characters, of the string to be copied. **If the string is longer than the maximum specified in the nMaxCount parameter, the extra characters are truncated**. If nMaxCount is 0, the function returns the length of the menu string.

Look from "GPD GPD GPD GPD GPD GPD GPD GPD GPD GPD GPD GPD GPD GPD GPD 1122" last character is truncated and null termination is added. If we use -1 in length then we are going to lose one more character which means.... New length will be 62 characters instead of 63. 😊 

image

SinghRajenM: @chcg just for reference. look at the screenshot when -1 is used..

image

chcg: @SinghRajenM Thanks for debugging this. So if the terminating null is there also in case of truncation everything is fine.

bpersonick:
Thanks Very much @SinghRajenM and @donho!  Super excited to be able to use more than 14 characters in the names of UDLS!  :)

PR #5467, opened 2019-03-29
Support old Mac line endings in "Remove Consecutive Duplicate Lines" command

vlakoff:
See https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0467b7d6cd19aa9bea373a9eabb44e73de0cea62#commitcomment-32962912.

As usual with such code, it's important the "\r\n" occurs before the "\r".

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1408,13 +1408,13 @@ void Notepad_plus::removeDuplicateLines()
 	// whichPart : line head or line tail
 	FindOption env;
 
-	env._str2Search = TEXT("^(.*\\r?\\n)(\\1)+");
+	env._str2Search = TEXT("^(.*(?:\\r\\n|\\n|\\r))(\\1)+");
 	env._str4Replace = TEXT("\\1");
     env._searchType = FindRegex;
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, true);
 
 	// remove the last line if it's a duplicate line.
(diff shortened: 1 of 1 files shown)


MetaChuh:
possible fix to make it work on files with <kbd>cr</kbd> only line endings:

`env._str2Search = TEXT("^(.+)\\r?\\n?(\\1)$");`

(instead of `env._str2Search = TEXT("^(.+)\\r?\\n(\\1)$");`)

vlakoff:
Yes, this works too, as we are only looking for one occurrence of line return.
Such code would be wrong in scenarios where there can be multiple line breaks, but that's not the case here so it's fine.

@donho if you prefer this version, you can replace both occurrences of `(?:\\r\\n|\\n|\\r)` by `\\r?\\n?`

MetaChuh:
both variants work, so it doesn't really matter which one is used.

currently i prefer don's `^(.+)(\r?\n|\r)(\1)$` variant, because it puts "\r\n" before "\r", as you mentioned above, making it more future proof.
... and ... if anything goes wrong, it wasn't me 😉

sasumner:
I certainly didn't mean to start a which-regex-is-better war. :)

MetaChuh:
@sasumner 

no, not at all. it's just me, who didn't realise that i was posting at a pr until now.
i thought it was just an opened issue 😉 sorry about that @vlakoff 🙏
(at least i wrote that it doesn't really matter which one is used)

second embarrassment of the day ... the first one was not testing the old <kbd>cr</kbd> classic 68k mac line endings, even though i'm the mac only dude ... unless you switched lately 😄

PR #5482, opened 2019-04-03
Updating Turkish localization

Tmp341:
Added new entries according to 54d2bc5 commit.


donho:
You should ALWAYS create the new PR.
I won't accept your PR no more if it happen again because it wastes my time.

Tmp341:
I always clear my branch after you accept it. I forgot this time it seems.

PR #5523, opened 2019-04-13
Add contextual menu for file encoding in status bar

### LINKED ISSUE #400: Add encoding context menu on status bar
Recently, [this commit](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/251e987a715cdb7ff829e23319677e849705a6b8#commitcomment-11999121) added 2 context menus on status bar:

> Add language context menu and EOL conversion context menu on status bar.

A natural addition to this would be the encoding menu.

vlakoff:
Fixes #400 and #756.
That would be a simple but very convenient feature.

The implementation seems to be [very easy](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/9b5a39b58f7844e62e46d4952155c29964c7c6f3).
The only thing I can't say is if it works correctly on the "Character sets" submenu. If yes, that's wonderful!

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -418,23 +418,23 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			if (notification->nmhdr.hwndFrom == _statusBar.getHSelf())
 			{
 				LPNMMOUSE lpnm = (LPNMMOUSE)notification;
-				if (lpnm->dwItemSpec == DWORD(STATUSBAR_CUR_POS))
-				{
-					bool isFirstTime = !_goToLineDlg.isCreated();
-					_goToLineDlg.doDialog(_nativeLangSpeaker.isRTL());
-					if (isFirstTime)
-						_nativeLangSpeaker.changeDlgLang(_goToLineDlg.getHSelf(), "GoToLine");
-				}
-				else if (lpnm->dwItemSpec == DWORD(STATUSBAR_DOC_SIZE))
-				{
-					command(IDM_VIEW_SUMMARY);
-				}
-				else if (lpnm->dwItemSpec == DWORD(STATUSBAR_DOC_TYPE))
+				if (lpnm->dwItemSpec == DWORD(STATUSBAR_DOC_TYPE))
 				{
 					POINT p;
 					::GetCursorPos(&p);
 					HMENU hLangMenu = ::GetSubMenu(_mainMenuHandle, MENUINDEX_LANGUAGE);
 					TrackPopupMenu(hLangMenu, 0, p.x, p.y, 0, _pPublicInterface->getHSelf(), NULL);
+				}
+				else if (lpnm->dwItemSpec == DWORD(STATUSBAR_DOC_SIZE))
+				{
+					command(IDM_VIEW_SUMMARY);
+				}
+				else if (lpnm->dwItemSpec == DWORD(STATUSBAR_CUR_POS))
+				{
+					bool isFirstTime = !_goToLineDlg.isCreated();
+					_goToLineDlg.doDialog(_nativeLangSpeaker.isRTL());
+					if (isFirstTime)
+						_nativeLangSpeaker.changeDlgLang(_goToLineDlg.getHSelf(), "GoToLine");
 				}
 				else if (lpnm->dwItemSpec == DWORD(STATUSBAR_EOF_FORMAT))
 				{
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppNotification.cpp
```diff
@@ -418,23 +418,23 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			if (notification->nmhdr.hwndFrom == _statusBar.getHSelf())
 			{
 				LPNMMOUSE lpnm = (LPNMMOUSE)notification;
-				if (lpnm->dwItemSpec == DWORD(STATUSBAR_CUR_POS))
+				if (lpnm->dwItemSpec == DWORD(STATUSBAR_DOC_TYPE))
```
chcg: Is there a reason why STATUSBAR_DOC_TYPE and STATUSBAR_CUR_POS are swapped?

vlakoff: There are 2 commits in this PR. The [first commit](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5523/commits/5bb118d96e1c516c59435bd13b778b269f52347a) is code reordering for readability, the [second commit](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5523/commits/9b5a39b58f7844e62e46d4952155c29964c7c6f3) is the actual change.

chcg: Thanks for the hint: "Sort conditionals like the display order" explains it all :-)

vlakoff:
Would someone be very kind, to compile with this change and try it?

chcg:
@vlakoff A compiled version is already available with the appveyor build, see e.g. for x64: https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/23821550/job/8x46cl6qnj8jm64f/artifacts

vlakoff:
Woooo, that's really great to know!

Thus, I tried the binary, and my change seems to be perfectly working :)

donho:
The Encoding menu is too large and complex to show as context menu.
So this feature won't be in Notepad++.

sasumner:
>Add contextual menu for file encoding in status bar

The title is confusing, but seems to want it in the status bar?

>The Encoding menu is too large and complex to show as context menu.

Even Don was confused, and doesn't mention status bar, although there really isn't room in the status bar for it anyway.

I didn't try it, but can't one edit `contextMenu.xml` and put what is desired in the right-click-of-editor-window-context-menu, perhaps even in an `Encodings >` grouping??  That would seem to satisfy as much as having something for it in the status bar.

vlakoff:
Indeed, it's the context menu _on the status bar_, not on the main window.
See the linked issues, there are demands for this.

Did you at least try it?
The menu works perfectly. And yes, it fits well.

As more, people are usually interested by the "root" encodings (ANSI, UTF-8), they more rarely dig into the submenus (which, again, are perfectly working too).

xylographe:
> I didn't try it,

By all means, give it a try. This patch makes the encoding in the status bar behave consistent with the _end of line mode_, and the _language_ : a right click or a double click opens the corresponding menu. The encoding menu is smaller than the language menu, hence lack of room is not an issue.

sasumner:
@vlakoff / @xylographe As you can tell, I didn't fully familiarize myself with all the background (which is a bit disjointed, due to age, perhaps) before my recent comment.  After doing so, it all seems fine to me.  I officially retract all of my earlier comments. :)

vlakoff:
Could this please be considered for reopening?

vlakoff:
Could this PR be considered again? There is demand for it, and it works perfectly well.

Currently, to set EOL and charset of a file, we have to do:
* bottom right of screen, right click on current EOL in status bar, choose the wanted EOL
  (much more convenient than "menu bar > Edit > EOL Conversion > ...")
* then move all the way up to top left of screen, "menu bar > Encoding > ..."

The usability issue is that the two actions are often used at the same time, but they are far apart.

Whereas with this PR, all we have to do is two right clicks on the status bar.

alankilborn:
>the two actions are often used at the same time,

I don't know that that is true for everyone that might want this, but I agree that this issue makes sense and should be adopted as a new feature!

The argument for declining was:  `The Encoding menu is too large and complex to show as context menu.`
I understand this argument if this would appear in a more general-purpose context, i.e., editor right click.
But this is a dedicated spot where user knows what he's getting into when invoking, so I see no harm in it.

vlakoff:
This has been addressed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a630b63c9a557a640a522ed22b2f7f1b18a0379a.

PR #5536, opened 2019-04-16
* fix non-idomatic use of std::remove_if

0017031:
1.  **erase-remove idiom**
please refer to the example code at [https://en.cppreference.com/w/cpp/algorithm/remove](https://en.cppreference.com/w/cpp/algorithm/remove)

2.  **the lambda predicate and** **`_istspace()`**,  **`_TUCHAR`** 
     Use such [Generic-Text Mappings](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/isspace-iswspace-isspace-l-iswspace-l?view=vs-2019#generic-text-routine-mappings) to align with the use of **`generic_string`**.
    When non-Unicode, [_TUCHAR ](https://docs.microsoft.com/en-us/cpp/text/generic-text-mappings-in-tchar-h?view=vs-2015#generic-text-data-type-mappings)is `unsigned char`; otherwise it is `wchar_t`. 
    Under both  situation we can avoid the undefined behavior of [isspace()](https://en.cppreference.com/w/cpp/string/byte/isspace) and [iswspace()](https://en.cppreference.com/w/cpp/string/wide/iswspace).

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -464,11 +464,10 @@ void Notepad_plus::command(int id)
 			if (nppGui._searchEngineChoice == nppGui.se_custom)
 			{
 				url = nppGui._searchEngineCustom;
-				remove_if(url.begin(), url.end(), isspace);
-
+				url.erase(std::remove_if(url.begin(), url.end(), [](_TUCHAR x){return _istspace(x);}),
+						  url.end());
 				auto httpPos = url.find(TEXT("http://"));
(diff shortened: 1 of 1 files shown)


donho:
@chcg Is it a bug?
I don't see in any way it's a bug.
Or if we can reproduce the bug? In this case, reopen this issue with the instructions to reproduce the bug.

0017031:
it is a bug. 
the codes are used to remove extra spaces in user's custom-search-engine-string
without std::erase(), the spaces is not actually removed, resulting in an incorrect search-engine definition

Issue #5516, opened 2019-04-11
Resizing panes allows automatic resizing of Other View


andrewlvaldez:
### Steps to Reproduce the Issue
1. Open the Other View along with the search results pane.
2. Resize the height of the results pane then mouse over the splitter between Views left to right.

### Expected Behavior
It is expected that mousing over the splitter does nothing.

### Actual Behavior
The Views resize without actually clicking on the splitter.

This happens when you resize any of the other panes, ie document map, functions list, search results.

Using v7.6.6 on Windows 7.

Originally reported as issue #680 using v6.7.9 on Windows 7.

MetaChuh:
@andrewlvaldez 

verified as reproducible.

**view splitter bar** displaces and responds to **mouseover without mousedown**,
after resizing a **docked** search result panel, as seen at the screencast below.

steps to reproduce:
* resize a docked search result (find result) panel, when multi-view is enabled.
* release the left mouse button and slowly hover over the splitter without holding any mouse button.
* **the splitter will shift to the right, and only to the right**.
(note: the screen recorder also does not detect a mouse up release)
* after clicking somewhere else, all reverts to normal.

![#5516 andrewlvaldez, Resizing panes allows automatic resizing of Other View](https://user-images.githubusercontent.com/45773305/55989749-53b1d280-5ca6-11e9-93ef-be99c45529b1.gif)

<kbd>Notepad++ v7.6.6   (64-bit)
Build time : Apr  3 2019 - 23:52:32
Path : C:\Users\User\Desktop\npp.7.6.6.bin.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll</kbd>

<kbd>Notepad++ v7.6.6   (32-bit)
Build time : Apr  3 2019 - 23:49:50
Path : C:\Users\User\Desktop\npp.7.6.6.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll</kbd>

Issue #2097, opened 2016-07-18
Resizing two-view mode gets stuck if you resize search view near intersection


kutluguzhin-g-soft:
Title says it all: if you have two-view mode and search view and try to resize search view and grab it at exact intersection of two views resizing for views gets stuck and follows your mouse when you are nearby.

andrewlvaldez:
I believe this might be a duplicate of issue #680.

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

PR #5554, opened 2019-04-20
Fixed defect #5516

### LINKED ISSUE #680: Resizing panes allows automatic resizing of Other View
Open the Other View along with the search results pane. 

Resize the height of the results pane then mouse over the splitter between Views left to right. 
The Views resize without actually clicking on the splitter.

This happens when you resize any of the other panes, ie document map, functions list, search results.

Using v6.7.9 on Windows 7.

SinghRajenM:
Fixes #680, #2097 and #5516

1. Resizing any dockable window (e.g. search result, function list, folder as workspace) also resizes the doc splitter (other view).
2. Corrected typo (`gotoRightBouuom` => `gotoRightBottom`)
3. Make use of in class initialization. (`RECT _rect` => `RECT _rect = {};`).

### DIFF
--- a/PowerEditor/src/WinControls/SplitterContainer/Splitter.cpp
+++ b/PowerEditor/src/WinControls/SplitterContainer/Splitter.cpp
@@ -37,17 +37,6 @@ bool Splitter::_isVerticalFixedRegistered = false;
 
 #define SPLITTER_SIZE 8
 
-
-
-Splitter::Splitter()
-{
-	_rect.left   = 0; // x axis
-	_rect.top    = 0; // y axis
-	_rect.right  = 0; // Width of the spliter.
-	_rect.bottom = 0; // Height of the spliter
-}
-
-
 void Splitter::init( HINSTANCE hInst, HWND hPere, int splitterSize, double iSplitRatio, DWORD dwFlags)
 {
 	if (hPere == NULL)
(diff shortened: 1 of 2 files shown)


PowerEditor/src/WinControls/SplitterContainer/Splitter.cpp
```diff
@@ -389,9 +379,10 @@ LRESULT CALLBACK Splitter::spliterWndProc(UINT uMsg, WPARAM wParam, LPARAM lPara
 
 		case WM_LBUTTONUP:
 		{
-			if (!_isFixed)
+			if (!_isFixed && _isLeftButtonDown)
 			{
 				ReleaseCapture();
+				_isLeftButtonDown = false;
```
AjayVV: I may be wrong in observation here. `_isLeftButtonDown` may remain true on mouse button up if `_isFixed` is true

SinghRajenM: Hmm, no. `_isLeftButtonDown` is set to `true` only if `_isFixed` is not true. [Look here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5554/files#diff-a39360c3bc45e15d195381c6bfdcc4a2R274)

SinghRajenM:
@donho , it's worth to look into this PR as related bug is pending for long time.

PR #5565, opened 2019-04-22
Resolves #5493, Allows you to do CTRL + [ or ] or reduce or increase indentation

### LINKED ISSUE #5493: [Feature Request] Use "Ctrl" + "[" "]" To Indent Line
### Description of the Issue

The above shortcut to quickly indent a line wherever the caret may be. The one feature of Brackets editor (JS-based) which is missing in Notepad++ —would be nice. Currently you have to hit Home or select the line & hit Tab key. Please add this in a future update, would save me tons of effort while coding [for AutoHotkey.] Ctrl + ] moves the tab indentation (current line) forward while Ctrl + [ moves it backwards. Thanks in advance.

1. 
2. 
3.

tojaroslaw:
Resolves #5493, Allows you to do CTRL + [ or ] or reduce or increase the current line indentation, this feature was requested so that notepad++ could have the same behavior as brackets

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -425,6 +425,8 @@ static const ScintillaKeyDefinition scintKeyDefs[] =
 	{TEXT(""),                            SCI_NEWLINE,                 false, false, true,  VK_RETURN,   0},
 	{TEXT("SCI_TAB"),                     SCI_TAB,                     false, false, false, VK_TAB,      IDM_EDIT_INS_TAB},
 	{TEXT("SCI_BACKTAB"),                 SCI_BACKTAB,                 false, false, true,  VK_TAB,      IDM_EDIT_RMV_TAB},
+	{TEXT("SCI_TAB"),                     SCI_TAB,                     false, false, false, VK_OEM_6,    IDM_EDIT_INS_TAB}, // CTRL + "]" indents line by 1
+	{TEXT("SCI_BACKTAB"),                 SCI_BACKTAB,                 false, false, false, VK_OEM_4,    IDM_EDIT_RMV_TAB}, // CTRL + "[" unindents line by 1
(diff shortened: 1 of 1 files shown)


Ekopalypse:
Correct me if I'm wrong but it looks like your commit does NOT assign the CTRL modifier because
all modifier flags are false. 
In addition, the proposed flags are already in use by SCI_PARAUP and SCI_PARADOWN messages and last but not least, I assume your commit does not do what #5493 was asking for. I understand, that,
regardless where the caret is, those shortcuts should trigger an action to in-/dedent the current line.

sasumner:
Why is this even necessary?  If a user wants to do ctrl+[ for SCI_TAB and ctrl+] for SCI_BACKTAB...then why wouldn't that user just go into Shortcut Mapper and configure these keystrokes?

PowerEditor/src/Parameters.cpp
```diff
@@ -425,6 +425,8 @@ static const ScintillaKeyDefinition scintKeyDefs[] =
 	{TEXT(""),                            SCI_NEWLINE,                 false, false, true,  VK_RETURN,   0},
 	{TEXT("SCI_TAB"),                     SCI_TAB,                     false, false, false, VK_TAB,      IDM_EDIT_INS_TAB},
 	{TEXT("SCI_BACKTAB"),                 SCI_BACKTAB,                 false, false, true,  VK_TAB,      IDM_EDIT_RMV_TAB},
+	{TEXT("SCI_TAB"),                     SCI_TAB,                     false, false, false, VK_OEM_6,    IDM_EDIT_INS_TAB}, // CTRL + "]" indents line by 1
+	{TEXT("SCI_BACKTAB"),                 SCI_BACKTAB,                 false, false, false, VK_OEM_4,    IDM_EDIT_RMV_TAB}, // CTRL + "[" unindents line by 1
```
donho: Have you ever tested your PR ?

PR #5580, opened 2019-04-24
Update chineseSimplified.xml

popcorner:


PowerEditor/installer/nativeLang/chineseSimplified.xml
```diff
@@ -295,10 +298,11 @@
                     <Item id="48018" name="编辑弹出菜单"/>
                     <Item id="48009" name="管理快捷键..."/>
                     <Item id="48011" name="首选项..."/>
+                    <Item id="48014" name="打开插件文件夹..."/>
                     <Item id="48015" name="插件管理..."/>
                     <Item id="48501" name="生成..."/>
                     <Item id="48502" name="从文件生成..."/>
-                    <Item id="48503" name="生成到剪贴板"/>
+                    <Item id="48503" name="从选区生成并复制到剪贴板"/>
                     <Item id="48504" name="生成..."/>
```
donho: The original translation of 48503 should be 生成到剪贴板 but not 从选区生成并复制到剪贴板
Your modification makes 48503 and 48506 have the same translation, whereas they are 2 different commands.

popcorner: @donho but 48503 and 48506 are the same in english.xml, I'm not sure why I can't make them the same in Chinese.

popcorner:
@donho I think the modification of item 48503 is right. I did the original translation when the MD5 was added. And I did the SHA256 translation when that was added, but I didn't find that the translation of 48503 has changed at that time. I noticed this change and modify the 48503 the same as 48506 when making this pull request, just like the english.xml did.

PowerEditor/installer/nativeLang/chineseSimplified.xml
donho: @popcorner  Yes my bad. You're right about it.

PR #5582, opened 2019-04-24
update to marathi.xml

gagarekumar:
Changed some spellings in Marathi language and added item id " 46250 ".

Why so important:
In Marathi language , we use चौथी for the word "fourth" .Like this,all the necessary changes have been made in "marathi.xml" file

### DIFF
--- a/PowerEditor/installer/nativeLang/marathi.xml
+++ b/PowerEditor/installer/nativeLang/marathi.xml
@@ -5,6 +5,9 @@
 	** created By:- Nikhil Tamhankar **
 		Email- iamnik.mailme@gmail.com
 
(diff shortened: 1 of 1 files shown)


donho:
Can anyone validate this PR?

SinghRajenM:
@donho , I don't know Marathi much. But it seems ok to me. There is no much difference except new item addition 46250.

P.S. India is big democratic country where so many languages exist. On top of that each language has its own flavor based on region. Like every 20~30 km, you can notice the change in accent for same language.

gagarekumar:
I know marathi well.I am from this region.Though at first site this seems minute difference.. but in marathi it means a lot..thanks

Issue #5586, opened 2019-04-25
ISSUE: RELOAD dialog appears when MONITORing a file


Daklu:
### Description of the Issue
When monitoring a file, the reload dialog appears when the file is updated.  This behavior is not observed in v7.6.3, and the behavior is slightly different in v7.6.4 than it is in 7.6.6.

### Steps to Reproduce the Issue
1. Monitor a file
2. Change the contents of the file externally

### Expected Behavior
The file is updated with no reload dialog

### Actual Behavior
v7.6.3 - No reload dialog appears.  (Expected behavior.)
v7.6.4 - Sometimes the reload dialog appears, sometimes it does not.  Can't cleanly repro.  May be related to whether or not Notepad++ has focus.
v7.6.6 - The reload dialog appears every time the window is updated.

### Debug Information
Notepad++ v7.6.6   (64-bit)
Build time : Apr  3 2019 - 23:52:32
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

MetaChuh:
@Daklu 

thank you for your debug information and details.

we can currently not reproduce this on 7.6.6. x64 or x86.
(see screencast, with working monitoring on 7.6.6 x64)
![Imgur](https://i.imgur.com/dfLdYUp.gif)

but we had a [second report about this at the notepad++ community](https://notepad-plus-plus.org/community/topic/17513/monitor-tail-f-prompts-to-reload-file), so we suppose it might be a setting, or leftover files from an old installation that could cause this.

**please test the following, to verify, if a clean notepad++ (without any plugins or settings that could cause this) would monitor as expected on your system:**

* download the notepad++ 7.6.6. 64 bit portable version from [>>> here <<<](https://notepad-plus-plus.org/repository/7.x/7.6.6/npp.7.6.6.bin.x64.zip).
(the portable version does not require any installation and runs completely independent and isolated from your installed version. all custom settings, as well as any stuff you might want to try at the portable version, will not modify your installed version)

* extract `npp.7.6.6.bin.x64.zip` to e.g. your desktop.

* **important note:** make sure to close all instances of notepad++ that might be running, before starting the portable version at the next step, to be sure you are using the portable version for this test.

[...]

Issue #4847, opened 2018-09-13
Suppress reload dialog when monitoring is enabled


dougbreaux:
### Description of the Issue

Sometimes - it appears to be maybe when I have more than one file monitored at the same time - I get regular reload dialogs when the monitored files change. Sometimes - maybe when I have only one monitored - this doesn't occur. And clearly the desired behavior is to not see such dialogs when monitoring.

### Steps to Reproduce the Issue

1.  Monitor one or more files
2.  Have those files change externally

### Expected Behavior

Keep updating file with no reload dialogs appearing

### Actual Behavior

Reload dialogs appear frequently

### Debug Information

Notepad++ v7.5.8   (32-bit)
Build time : Jul 23 2018 - 02:03:53
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : NPPJSONViewer.dll XMLTools.dll DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll 

Pull request #4527 seems to be about the same thing, code there might be relevant.

dinkumoil:
Can you confirm that the following setting is **NOT** checked while monitoring?

![grafik](https://user-images.githubusercontent.com/24699812/45509015-a968df80-b796-11e8-8740-b715a425b6f3.png)

dougbreaux:
No, it's checked. I do want detection of non-monitored files. Or am I misunderstanding something?

dinkumoil:
If the setting above is turned off you won't get reload confirmation dialogs anymore, this is my experience. But I agree that this behaviour isn't comfortable and should be changed. Maybe you should extend your report and include this fact.

PR #5591, opened 2019-04-27
Fixed File Monitoring Reload issue

SinghRajenM:
Fixed issue #5586 and #4847....

### Root cause:
In member function`void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)` below code gets executed ...
```C++
case DOC_NEEDRELOAD: // by log monitoring
			{
				doReload(buffer->getID(), false);
```
```doReload(buffer->getID(), false);``` calls ```activateBuffer(id, currentView());``` and ```activateBuffer(id, currentView());``` calls ```checkFileState()```.

Now in ```checkFileState()``` below code causes this issue. As I said [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5586#issuecomment-487273042), it's all about the thread timing. So timestamp in npp does not match with timestamp of actual file. Therefore, `_currentStatus` becomes `DOC_MODIFIED`.

```C++
		if (CompareFileTime(&_timeStamp, &attributes.ftLastWriteTime) != 0)
		{
			_timeStamp = attributes.ftLastWriteTime;
			mask |= BufferChangeTimestamp;
			_currentStatus = DOC_MODIFIED;
			mask |= BufferChangeStatus;	//status always 'changes', even if from modified to modified
		}
```

**P.S. Updating `Buffer.cpp` is enough to deal with this issue. But to be safer side, it is good to block reloading in `case DOC_MODIFIED:` in file `Notepad_plus.cpp` too.**

### Edit
Commit 2 deals with  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487313886 which is because of Microsoft and described here (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487341747) why it is MS.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3328,6 +3328,19 @@ void Notepad_plus::checkModifiedDocument(bool bCheckOnlyCurrentBuffer)
 	MainFileManager->checkFilesystemChanges(bCheckOnlyCurrentBuffer);
 }
 
+void Notepad_plus::checkModifiedDocumentIfMonitoring()
+{
+	// This is an workaround to deal with Microsoft issue in ReadDirectoryChanges notification
+	// If command prompt is used to write file continuously (e.g. ping -t 8.8.8.8 > ping.log)
+	// Then ReadDirectoryChanges does not detect the change
+	// Fortunately, notification is sent if right click or double click happens on that file
+	// Let's leverage this as workaround to enhance npp file monitoring functionality.
+
+	Buffer* currBuf = getCurrentBuffer();
+	if (currBuf && currBuf->isMonitoringOn())
+		::PathFileExists(currBuf->getFullPathName());
+}
+
 void Notepad_plus::getMainClientRect(RECT &rc) const
 {
     _pPublicInterface->getClientRect(rc);
@@ -5129,6 +5142,10 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
 			}
 			case DOC_MODIFIED:	//ask for reloading
 			{
+				// Since it is being monitored DOC_NEEDRELOAD is going to handle the change.
+				if (buffer->isMonitoringOn())
+					break;
+
 				bool autoUpdate = (nppGUI._fileAutoDetection & cdAutoUpdate) ? true : false;
 				if (!autoUpdate || buffer->isDirty())
 				{
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -224,7 +224,10 @@ void Buffer::setFileName(const TCHAR *fn, LangType defaultLang)
 
 bool Buffer::checkFileState() // returns true if the status has been changed (it can change into DOC_REGULAR too). false otherwise
 {
- 	if (_currentStatus == DOC_UNNAMED)	//unsaved document cannot change by environment
+	// 1. Unsaved document cannot change by environment
+	// 2. Monitoring is sent by NPPM_INTERNAL_RELOADSCROLLTOEND
+	// So no need to check file status for both the above cases
+	if (_currentStatus == DOC_UNNAMED || isMonitoringOn())
 		return false;
 
 	WIN32_FILE_ATTRIBUTE_DATA attributes;
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -458,6 +458,7 @@ private:
 	void checkSyncState();
 	void dropFiles(HDROP hdrop);
 	void checkModifiedDocument(bool bCheckOnlyCurrentBuffer);
+	void checkModifiedDocumentIfMonitoring();
 
     void getMainClientRect(RECT & rc) const;
 	void staticCheckMenuAndTB() const;
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1532,6 +1532,13 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_INTERNAL_CHECKDOCSTATUS:
 		{
+			// This is an workaround to deal with Microsoft issue in ReadDirectoryChanges notification
+			// If command prompt is used to write file continuously (e.g. ping -t 8.8.8.8 > ping.log)
+			// Then ReadDirectoryChanges does not detect the change
+			// Fortunately, notification is sent if right click or double click happens on that file
+			// Let's leverage this as workaround to enhance npp file monitoring functionality.
+			checkModifiedDocumentIfMonitoring();
+
 			const NppGUI & nppgui = pNppParam->getNppGUI();
 			if (nppgui._fileAutoDetection != cdDisabled)
 			{


MetaChuh:
@SinghRajenM 

**problem:**

using `ping -t 8.8.8.8 > ping.log` and monitoring it:

your fix removes the reload dialogue, but the monitored file
**does not get updated automatically and does not get updated on focus change**.

only a tab change or a close file (<kbd>ctrl</kbd>+<kbd>c</kbd>) updates the monitored view, as seen at the screencast:

7 6 6  monitoring bug with #5591

<kbd>Notepad++ v7.6.6   (32-bit)
Build time : Apr 27 2019 - 12:39:26
Path : C:\Users\User\Desktop\npp.7.6.6.bin\Notepad++.7.6.6 (#5591 rajendra - Fixed File Monitoring Reload issue).exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll</kbd>

many thanks and sorry for being slower than usual this week.

SinghRajenM:
>problem:
using ping -t 8.8.8.8 > ping.log and monitoring it:
your fix removes the reload dialogue, but the monitored file
**does not get updated automatically and does not get updated on focus change.**

@MetaChuh this is not an issue with npp. But this is design (or may be bug) from MS. `ping` command result is not flush immediately. MS keeps it in buffer and writes to file periodically to avoid extensive IO operation. Buffer is flush immediately while write click or double clicking on the output file name. See the image below.   

Running below commands give proper output
```batch
@echo off
setlocal

:loop
echo %time% >> time.txt
timeout /t 5
goto :loop
```


I started `ping -t 8.8.8.8 > time.txt` at 9:25 and at 9:37 also explorer show 1KB file with last update time as 9:25 only...
image

image



**After double clicking on file:**

image

MetaChuh:
@SinghRajenM 

the section
```
if (buffer->isMonitoringOn())
	break;
```
will prevent `doReloadOrNot`, but it will also prevent
```
doReload(buffer->getID(), false);
if (mainActive || subActive)
{
	performPostReload(mainActive?MAIN_VIEW:SUB_VIEW);
}
```
from happening.

ps: i can understand you perfectly, and i know it's rather a windows 10 file handling problem as long as a file is not closed, but a typical notepad++ user will currently expect, that a monitored file will at least get updated, if he invokes a manual focus change, so imho we should not swap one issue for another issue if possible.

if needed, please have a look again at the stock 7.6.6 video from https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5586#issuecomment-487299912
the file gets updated under the same circumstances on focus change, albeit producing a reload dialogue.

many thanks for your help and best regards

SinghRajenM:
> but it will also prevent
> 
> ```
> doReload(buffer->getID(), false);
> if (mainActive || subActive)
> {
> 	performPostReload(mainActive?MAIN_VIEW:SUB_VIEW);
> }
> ```

@MetaChuh above is expected too. BTW, `performPostReload(mainActive?MAIN_VIEW:SUB_VIEW);` is equivalent (almost) to below lines which are part of ``` c++ case DOC_NEEDRELOAD: // by log monitoring```
```C++
// not only test main view
				if (buffer == _mainEditView.getCurrentBuffer())
				{
					_mainEditView.execute(SCI_GOTOLINE, _mainEditView.execute(SCI_GETLINECOUNT) - 1);
				}
				// but also test sub-view, because the buffer could be clonned
				if (buffer == _subEditView.getCurrentBuffer())
				{
					_subEditView.execute(SCI_GOTOLINE, _subEditView.execute(SCI_GETLINECOUNT) - 1);
				}
```

> if needed, please have a look again at the stock 7.6.6 video from #5586 (comment)
the file gets updated under the same circumstances on focus change, albeit producing a reload dialogue.

File is getting updated because you have enabled "File auto change detection". When I disable "File auto change detection", file is not getting updated when I focus between cmd prompt and notepad++. However, switching between tabs within npp, updates the file in both new and old implementation. So we can try leverage this part and improve monitoring functionality.


> but a typical notepad++ user will currently expect, that a monitored file will at least get updated, if he invokes a manual focus change, so imho we should not swap one issue for another issue if possible.

True, I second it. But here we are not swapping issue as here file is getting updated because of other feature (auto file change detection).

SinghRajenM:
@MetaChuh a workaround has been provided in commit 2 for https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487313886.

MetaChuh:
@SinghRajenM 

monitoring works better than ever before on win 10, with your modified fix. 👍
verified fix for both issues, monitoring multiple files.
please note, that processing will take longer than usual.

many thanks and best regards.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3328,6 +3328,19 @@ void Notepad_plus::checkModifiedDocument(bool bCheckOnlyCurrentBuffer)
...
+{
+	// This is an workaround to deal with Microsoft issue in ReadDirectoryChanges notification
+	// If command prompt is used to write file continuously (e.g. ping -t 8.8.8.8 > ping.log)
+	// Then ReadDirectoryChanges does not detect the change
+	// Fortunately, notification is sent if right click or double click happens on that file
+	// Let's leverage this as workaround to enhance npp file monitoring functionality.
+
+	Buffer* currBuf = MainFileManager->getCurrentBuffer();
+	if (currBuf && currBuf->isMonitoringOn())
+		::PathFileExists(currBuf->getFullPathName());
```
AjayVV: Does `PathFileExists` imitate the behavior of double/right click to the file?

SinghRajenM: > Does PathFileExists imitate the behavior of double/right click to the file?

No, it does not. It just checks for file existence. But internally, I might be gathering some info such as file size, last modified, last access or something similar. So when this API is called, that time on Windows10, file info is updated automatically. So it is just being utilized here.

AjayVV: I understand that this function is part of Shell API, and might be doing something internally to flush the existing open file buffers (Explorer is very close to OS/file-system). However, I couldn't find, anywhere, that this function does flush the file buffers of open files. Hence, this solution is not provably correct.
Also, it might be working only on Windows 10, and may not exhibit the same behavior on other OS. 
It should also be noted that file monitoring and antivirus software, which do install file-system drivers, might also result in this behavior.

LordNyriox:
@MetaChuh:  **Off-Topic Question** 

How do you make Github keystroke text-boxes like the ones you used in [<https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487313886>]? 

I have been unsuccessfully trying to use those for ages.  :\

LordNyriox:
@MetaChuh:  **Off-Topic Question** 

How do you make Github keystroke text-boxes like the ones you used in [<https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487313886>]? 

I have been (unsuccessfully) trying to figure out how to use those, and Google is not being much of a help.  :\

PowerEditor/src/Notepad_plus.cpp
SinghRajenM: > Hence, this solution is not provably correct.

This is a work around. 

> Also, it might be working only on Windows 10, and may not exhibit the same behavior on other OS.
This occurs only in win10 (as of now). There is no issue observed in other OS. Something has been broken on windows 10.

AjayVV: I understand now. Then this function should be called when OS >= Windows 10 only. However, I don't see `winVer` enum having Win10 as one of the value.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3328,6 +3328,19 @@ void Notepad_plus::checkModifiedDocument(bool bCheckOnlyCurrentBuffer)
...
 
+void Notepad_plus::checkModifiedDocumentIfMonitoring()
+{
+	// This is an workaround to deal with Microsoft issue in ReadDirectoryChanges notification
+	// If command prompt is used to write file continuously (e.g. ping -t 8.8.8.8 > ping.log)
+	// Then ReadDirectoryChanges does not detect the change
+	// Fortunately, notification is sent if right click or double click happens on that file
+	// Let's leverage this as workaround to enhance npp file monitoring functionality.
+
+	Buffer* currBuf = MainFileManager->getCurrentBuffer();
```
donho: Use getCurrentBuffer() of class Notepad_plus instead.

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -554,6 +554,10 @@ Buffer* FileManager::getBufferByIndex(size_t index)
 	return _buffers.at(index);
 }
 
+Buffer* FileManager::getCurrentBuffer()
+{
+	return _pNotepadPlus ? _pNotepadPlus->getCurrentBuffer() : nullptr;
+}
```
donho: Not necessary.

PowerEditor/src/ScitillaComponent/Buffer.h
```diff
@@ -80,6 +80,7 @@ class FileManager final
 	int getBufferIndexByID(BufferID id);
 	Buffer * getBufferByIndex(size_t index);
 	Buffer * getBufferByID(BufferID id) {return static_cast<Buffer*>(id);}
+	Buffer* getCurrentBuffer();
```
donho: Not necessary.

PowerEditor/src/Notepad_plus.cpp
SinghRajenM: You're right. I didn't notice it.

PowerEditor/src/ScitillaComponent/Buffer.cpp
SinghRajenM: Removed.

PowerEditor/src/Notepad_plus.cpp
SinghRajenM: Fixed it now.

PowerEditor/src/ScitillaComponent/Buffer.h
SinghRajenM: Removed it.

donho:
please keep separate issue per pr.
Open another PR for the rest of commit please.

SinghRajenM:
I intentionally put together because both issues were interlinked. Anyway, I will create new PR for below issue - 

Commit 2 deals with #5591 (comment) which is because of Microsoft and described here (#5591 (comment)) why it is MS.

Issue #5576, opened 2019-04-23
Plugins Admin removes itself and all other plugins


rddim:
### Description of the Issue

Plugins Admin removes itself and all other plugins when removed manually installed plugins
### Steps to Reproduce the Issue

1. Manually install some plugin (for example [AutoSave_dll_1v60.zip](https://sites.google.com/site/fstellari/nppplugins/AutoSave_dll_1v60.zip?attredirects=0))
2. Go to `Plugins` > `Plugins Admin...` > `Installed`
3. Check the `AutoSave.dll` and click `Remove`

### Expected Behavior

To not remove Plugins Admin and other plugins
### Actual Behavior

After step `3` and N++ restart everything is removed
### Debug Information

Notepad++ v7.6.6   (64-bit)
Build time : Apr  3 2019 - 23:52:32
Path : D:\Install\Office Programs\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : AutoSave.dll ComparePlugin.dll ShtirlitzNppPlugin.dll VisualStudioLineCopy.dll 

### Other Info
All the `7.6.x` versions are affected

MetaChuh:
duplicate reference #5279

SinghRajenM:
**Root Cause:**
For manually installed plugins, there is no value for `folder-name` in npp code. So `i->_folderName` is empty which could not make path as `C:\Program Files\Notepad++\plugins\AutoSave`. So the remove directory remains `C:\Program Files\Notepad++\plugins\` which does magic to remove complete plugin folder rather than `plugin\AutoSave`
```
else // op == pa_remove
{
	// add folder to operate
	updaterParams += TEXT(" \"");
	updaterParams += i->_folderName;
	updaterParams += TEXT("\"");
}
```


```
{
			"folder-name": "mimeTools",
			"display-name": "Mime tools",
			"version": "2.2",
			"id": "c89b2bb4ca3e8ae5e3f6d3169fcbe830165b330b1c4639ecb4149ed91fabba06",
			"repository": "https://github.com/npp-plugins/mimetools/releases/download/v2.2/mimetools.v2.2.zip",
			"description": "MIME plugin for Notepad++ implements several main functionalities defined in MIME (Multipurpose Internet Mail Extensions).",
			"author": "Don HO",
			"homepage": "https://github.com/npp-plugins/mimetools"
		},

```

Issue #5279, opened 2019-01-29
Removing an non-plugin-admin-managed plugin causes whole plugins folder to disappear


TaoK:
### Description of the Issue

When using plugin admin to remove a plugin that is not supported by plugin admin (yet), it seems that the plugin admin system accidentally removes the whole "plugins" folder.

### Steps to Reproduce the Issue
1. Install v7.6.3 (32-bit) with default options, so you have the Plugins menu with plugin admin and the default plugins
2. Add the Poor Man's T-Sql Formatter plugin (eg download from  http://architectshack.com/GetFile.aspx?Page=PoorMansTSqlFormatter&File=SqlFormatterNppPlugin.1.6.10.b.zip, put it the right place)
3. Restart NPP, confirm Poor Man's T-SQL Formatter is available
4. Open plugin admin, find Poor Poor Man's T-SQL Formatter in the "installed" tab
5. "Remove" it, accepting the restart warning, and the admin-mode popup thing

### Expected Behavior

When Notepad++ restarts, the one plugin you requested to remove should be gone.

### Actual Behavior

When Notepad++ restarts, the entire Plugins menu is gone, including plugin admin.

Nu11u5:
Reproduced this issue exactly with the older Scout-Soft CSV SQL plugin (http://www.scout-soft.com/sql/).

NPP v7.6.4 (32-bit).

chcg:
Maybe this is not an issue directly within N++ PA, but from https://github.com/notepad-plus-plus/wingup. This needs to be further investigated.

MetaChuh:
duplicate reference #5576

SinghRajenM:
>Maybe this is not an issue directly within N++ PA, but from https://github.com/notepad-plus-plus/wingup. This needs to be further investigated.

@chcg Nope, this is not an issue with N++ PA. Actual issue lies within npp. Npp must handle this special case......

**Root Cause:**
For manually installed plugins, there is no value for `folder-name` in npp code. So `i->_folderName` is empty which could not make path as `C:\Program Files\Notepad++\plugins\AutoSave`. So the remove directory remains `C:\Program Files\Notepad++\plugins\` which does magic to remove complete plugin folder rather than `plugin\AutoSave`
```
else // op == pa_remove
{
	// add folder to operate
	updaterParams += TEXT(" \"");
	updaterParams += i->_folderName;
	updaterParams += TEXT("\"");
}
```


```
{
[...]

PR #5593, opened 2019-04-27
Bug Fix: Don't remove whole plugin System

SinghRajenM:
Fixed issue #5576 and #5279

### DIFF
--- a/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
+++ b/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
@@ -487,7 +487,17 @@ bool PluginsAdminDlg::exitToInstallRemovePlugins(Operation op, const vector<Plug
 		{
 			// add folder to operate
 			updaterParams += TEXT(" \"");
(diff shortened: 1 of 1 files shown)


PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
```diff
@@ -487,7 +487,17 @@ bool PluginsAdminDlg::exitToInstallRemovePlugins(Operation op, const vector<Plug
...
 			updaterParams += TEXT(" \"");
-			updaterParams += i->_folderName;
+			generic_string folderName = i->_folderName;
+			if (folderName.empty())
+			{
+				auto lastindex = i->_displayName.find_last_of(TEXT("."));
+				if (lastindex != generic_string::npos)
+					folderName = i->_displayName.substr(0, lastindex);
+				else
+					folderName = i->_displayName;	// This case will never occur, but in case if it occurs too
```
LordNyriox: @SinghRajenM:  Perhaps change the wording (of the comment) for clarity? 

>// This case will never occur

...could be...

>// This case *should* never occur

Issue #1487, opened 2016-02-17
There are string not settable from XML


akittino:
Take a look into these screenshots:

Every text that is in English on screens are not settable in XML.
Maybe I did not defined proper entry, if you know that, please let me know :)
1. Untranslatable check box.:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13099985/a7f93a98-d538-11e5-8f8b-b35c164c045d.png)
2. Dialog # 1:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13099992/b79c186c-d538-11e5-8157-071c9c1d913a.png)
3. Dialog # 2:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13099994/bbeae4b6-d538-11e5-8541-dabbd6a520ab.png)
4. Dialog # 3:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13100000/c1ce2d16-d538-11e5-8cf4-85b0769c2e75.png)
5. Restore recent closed file:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13100004/c98203ac-d538-11e5-8dc3-39ec7a83d193.png)
6. Column mode tip message box:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13100007/ce365db2-d538-11e5-9506-28ec33519091.png)
7. SummaryBox:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13100010/d1e32d00-d538-11e5-8c35-25aea6099778.png)
8. Shortcut mapper:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13100012/d5c059ca-d538-11e5-8bb5-afff42100468.png)
9. Trim Trailling and Save:
   ![image](https://cloud.githubusercontent.com/assets/6968485/13100015/d9c9c8da-d538-11e5-84a8-b0e48f750873.png)
10. Run entries:
    ![image](https://cloud.githubusercontent.com/assets/6968485/13100017/dfa58a96-d538-11e5-830a-ab052f07f705.png)

MetaChuh:
we are closing this issue due to multiple newer notepad++ versions since it was filed.

please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html).
if your issue persists, and it is still of relevance to you, please feel free to file a new issue as detailed and reproducible as possible.

thank you for your help and best regards.

[...]

Issue #1476, opened 2016-02-14
Restore Recent Closed File translation not picked up


batagy:
Hi!

Translation related issue.
Below item's translation is not pickup by the program properly:

Item id=“41021” name=“Restore Recent Closed File”

Around 2015-05-10 , this was fixed, so that translation was picked up properly in Notepad++, but later in version 6.8.1 it went bad again. So the translation is not pickup up again since 6.8.1 version.

The issue I've mentioned already in PR #542  towards the end of discussion, but finally it was ignored.

Please note! If in Settings, if you change the language to English, then back to the Language, then the translation is refreshed and picked up! But as soon as you open Notepad++ next time, this item is forgotten again.

Thanks a lot
György

Tmp341:
I've stated that in #1016, and i thought it was my mistake. But it looks like the problem is occuring other than English. Just like you said : 

> if you change the language to English, then back to the Language, then the translation is refreshed and picked up! But as soon as you open Notepad++ next time, this item is forgotten again.

batagy:
Hi!
Why is that so hard to correct such a minor issue?! 
Should we wait years somebody to correct this damn issue?
Thanks a lot

batagy:
@Rikk 
@milipili 
Hi guys! Could you please check this annoying bug?
Why Don ignores this for months?
Who can correct this?

(PS: this issue was already mentioned in #542 .)

Rikk:
Why me? I don't help Npp anymore, the entire project is handled and slowed by only one sir: donho.
There are hundreds of proposed fixes that are simply ignored. I, at least, don't like to help this kind of projects.

batagy:
Hi!
This issue is still present in version 7.3.

schnurlos:
And still in 7.6.1 ...

chcg:
Still repoducible with:

Notepad++ v7.6.6   (64-bit)
Build time : Apr  3 2019 - 23:52:32
Path : C:\Users\christian\Source\Repos\npp.7.6.6.bin.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll PluginManager.dll XMLTools.dll

MetaChuh:
way to reproduce summary:
* change to any language (except english)
(`file > restore recent closed file` is correctly shown translated)
* close notepad++
* reopen notepad++

[...]

PR #5594, opened 2019-04-27
add capabilty to translate file summary dialog info

chcg:
See #1487, there is a left over from the non translatable texts.

Still open:
7. SummaryBox:

Meintioned also in:
Restore Recent Closed File translation not picked up #1476


schnurlos:
Hopefully this will be included soon ... thx for adding that

PR #5624, opened 2019-05-06
changed for loop on line 56 to a range based for loop.

servebotfrank:
Hey I noticed this line was written using the pre c++11 style with indices so I tried to fix that by turning it into a range based for loop. I hope this is acceptable.


chcg:
@servebotfrank See https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/24349132/job/v9lsbjt5du2eisvg/messages compilation issue.

PR #5636, opened 2019-05-09
Suggestion: Store scintilla sources as git submodule

bahusoid:
I think it would be better if you create separate scintilla fork  https://github.com/notepad-plus-plus/scintilla  based on https://github.com/mirror/scintilla and link it here as git submodule instead of copy\paste.

Also maybe create separate branch for Notepad++ specific modifications(something like [this](https://github.com/bahusoid/scintilla/tree/npp))
That will give clear idea what  was changed (see [here](https://github.com/bahusoid/scintilla/commit/8d986378f8774e79f7ffb430096a1a5d31477ccb)) and simplify further upgrades

What do you think? I can do all manipulations in this PR but I need you to create scintilla fork.


LordNyriox:
@donho:  What do you think of this idea?

donho:
Submodule creates a lot confusion by its nature, so no submodule in Notepad++ project.

PR #5659, opened 2019-05-16
Fix possible data loss on save failure

pnedev:


donho:
@pnedev 
Could you provide the bug number or bug description?
What kind of "save failure" are you talking about?

pnedev:
@donho ,

I couldn't find filed bug describing the issue but I recall several user complaints about the same problem.
**[Here](https://notepad-plus-plus.org/community/topic/17632/crital-help-file-content-lost-when-save/1)** is a link from the forum describing the issue.

As a summary:
1) Backup is disabled
2) File from disk is opened in Notepad++
3) Disk space runs out
4) User tries to save the file after adding several lines
5) File save fails and user is asked to try in Admin mode
6) User answers "No" to admin mode
7) The file is corrupted, no Undo is possible - data lost

I don't know if my change fixes the issue but since on save failure we return without executing
`_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, _scratchDocDefault);`
I thought that might be causing the problems.

I just wanted to point that out as a potential cause for you to review as you certainly know the code better than me.

Thanks.

MetaChuh:
@pnedev 

i haven't had time to fill up a hard drive to zero space left, and verify if it is really reproducible.
were you able to simulate what the user has experienced with and without your fix ?

pnedev:
@MetaChuh ,

I haven't tried to reproduce the issue. As you said, leaving the disk with 0 space is not something you can quickly reproduce.

I just examined the file save logic in the code and see the pointed thing as a potential problem.
Following the issue description and the code we'll end-up with `_pscratchTilla`  (the Notepad scratch Scintilla instance) keeping the original view doc as it's doc pointer.

donho:
The code in PR provide a more correct code, in which the reference of document to save is decreased correctly (no memory leak), but that's all. 
I don't think it's fixed anything, but the code is what it should be, and since out of disk space situation happens rarely, this fix will go into 7.7 exceptionally.

pnedev:
@donho ,

Thanks for merging the change.
It really doesn't solve the issue but as you said is the correct code in this case.

I found the root cause for the problem and filed a new issue.
Please have a look at it **[here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664)** for the description of the cause.

@MetaChuh ,

> were you able to simulate what the user has experienced

Yes, the issue is reproducible.
I used a USB stick to check it.

Issue #3467, opened 2017-07-01
Project for visual Studio 2017


tarung65:
### Description of the Issue
Unable to Build project in Visual Studio 2017 

### Steps to Reproduce the Issue

1. Open notepadPlus.vs2015.vcxproj in visual Studio 2017
2. Build the project
3. Compilation failed

### Expected Behavior

A new project to be created for Visual studio 2017

irgendsontyp:
As a workaround, which works for me, you can modify the notepadPlus.vs2015.vcxproj by changing "v140_xp" to "v141" and then try again.

bahusoid:
If the only thing that needs to be changed is  "Platform Toolset" then I think separate project is an overkill for VS 2017. I think it's better to adjust existing vs2015 project to calculate proper "Platform Toolset". I don't have VS 2017 but I prepared prototype for such project:

[notepadPlus.vs2015.vcxproj.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1122331/notepadPlus.vs2015.vcxproj.zip)

Could you please check if it works with VS 2017?

tarung65:
But as the newer version of visual studio is rolling out it is always better to switch to that and community version of Visual Studio 2017 is free so why we are using its older version.It is always better to upgrade to the latest tools.

bahusoid:
I'm not saying that you shouldn't use it. But why to force it? Adding additional solution is additional burden to support. And what the benefits?

tarung65:
if that is the reason then the older version of project also should be removed

bahusoid:
If for VS 2013  it's also only about "Platform Toolset" - it's easy doable too. Could you please confirm that my changes work with VS 2017?

tarung65:
Yes, your changes works but we have to ignore warnings.

bahusoid:
Ok. Thanks. If I understand correctly build is failing due to warnings (so you need to disable Treat Warnings as Error flag) That's actually a no go. I will check how can I make it build without warnings.

dail8859:
Closing as it is now buildable with VS2017

chcg:
See https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5306 for compilation issues with platform toolset of VS2017 v141 or v141_xp.

PR #5693, opened 2019-05-24
Followup on issue #3467

chcg:
- update toolset from v140_xp -> v141_xp
- removed outdated VS2013 project file


SinghRajenM:
@chcg  the build is failing and reason is explained in PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5306. But unfortunately, it was not accepted.

@donho  you may need to look into PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5306 again or @chcg include the changes here.

donho:
@SinghRajenM  There's never problem for me to build under VS2017 with the current code, so I don't see the reason to change it.

SinghRajenM:
> restore all v140_xp instead of v141_xp

@donho ,
Didn't get the context why you want to keep the toolset from VS2015 while both projects (npp and scintilla) have already been migrated to VS2017?

Just for reference from [MS Doc](https://docs.microsoft.com/en-us/cpp/build/how-to-modify-the-target-framework-and-platform-toolset?view=vs-2019)
>**Note**
**To change the target platform toolset, you must have the associated version of Visual Studio or the Windows Platform SDK installed.** For example, to target the Itanium platform with the Windows7.1SDK platform toolset, you must have Microsoft Windows SDK for Windows 7 and .NET Framework 4 SP1 installed; however, you could use another compatible version of Visual Studio to do your development work, provided that you target the correct Framework version and platform toolset.

Currently to build npp, it is required to install toolset of VS2015. Fortunately, MS made it easy to install different versions of toolsets using "Visual studio Installer", but it comes with additional cost. In VS2019 it cost around 3GB extra space.
**Additionally, v140 can be installed, but v140_xp can't be installed using "Visual studio Installer 2019". Not sure about "Visual studio Installer 2017". Hopefully, "Visual studio Installer 2017" should allow to install v140_xp. If not, then VS2015 must be installed separately to build npp.**

image

**P.S. Just curious to know why do you want to keep VS2015 toolset at all instead of VS2017 toolset?**

chcg:
@SinghRajenM You are right, I have local modification for the v141_xp build issues. 
@donho I will update the PR to avoid the compiler issues. Ok?

donho:
@chcg As I said, please restore all v140_xp instead of v141_xp to make PR accept.
@SinghRajenM Yes, I saw that v141_xp come with VS2017. Strangely, it's not installed on my PC.
Since all the releases are built from my PC, I try to keep the build environment as stable as possible.

chcg:
@donho As you want to stay at older VS2015 toolchain the only thing what was left is the removal of the outdate vs2013 project file you did already.
But still unclear as @SinghRajenM pointed out why you want to stick to vs140_xp instead of vs141_xp. See also https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/win32/SciLexer.vcxproj#L30 , so scintilla is also using vs141.

PR #5732, opened 2019-06-03
Fix typo on change.log

DavidePastore:
**removes** instead of **romoves**.


MetaChuh:
@DavidePastore 

you can impossibly be serious.
a typo fix for the old change.log, which will not be part of the next release, does not make much sense, does it ?

peace.

sasumner:
@MetaChuh Maybe it was a practice PR.  :)

DavidePastore:
I've never seen such a confident and friendly repository.

sasumner:
@DavidePastore Nono, we're friendly.  Please, give as good as you get. Don't be discouraged. We aren't above some humor here. :)

PR #5753, opened 2019-06-07
change.log typo fixed

sunzone93:
change.log typo 'romoves' fixed to 'removes'


MetaChuh:
@sunzone93 

this line will not be present at the next change.log, so there is no need to correct the typo.
note: please do not create a pull request without a [valid, open and accepted issue](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#pull-requests) for it.

many thanks and best regards.

PR #5773, opened 2019-06-11
Continuation of PR #5591 (workarround to MS ReadDirectoryChanges bug)

SinghRajenM:
This PR is to deal with  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487313886 which is because of Microsoft and described here (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5591#issuecomment-487341747) why it is MS.

Please note: This issue occurs only in Windows 10.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1532,6 +1532,17 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_INTERNAL_CHECKDOCSTATUS:
 		{
+			// This is an workaround to deal with Microsoft issue in ReadDirectoryChanges notification
+			// If command prompt is used to write file continuously (e.g. ping -t 8.8.8.8 > ping.log)
+			// Then ReadDirectoryChanges does not detect the change.
+			// Fortunately, notification is sent if right click or double click happens on that file
+			// Let's leverage this as workaround to enhance npp file monitoring functionality.
+			// So calling "PathFileExists" is a workaround here.
+
+			Buffer* currBuf = getCurrentBuffer();
+			if (currBuf && currBuf->isMonitoringOn())
+				::PathFileExists(currBuf->getFullPathName());
+
 			const NppGUI & nppgui = pNppParam->getNppGUI();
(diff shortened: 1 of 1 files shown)


MetaChuh:
@donho 

my apologies, the original 2 in 1 commit was my fault, not @SinghRajenM 's.

without this second commit, running `ping -t 8.8.8.8 > ping.log` and monitoring this file, will not even update it on focus change.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3328,6 +3328,19 @@ void Notepad_plus::checkModifiedDocument(bool bCheckOnlyCurrentBuffer)
 	MainFileManager->checkFilesystemChanges(bCheckOnlyCurrentBuffer);
 }
 
+void Notepad_plus::checkModifiedDocumentIfMonitoring()
```
donho: Remove this function

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1532,6 +1532,13 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_INTERNAL_CHECKDOCSTATUS:
 		{
+			// This is an workaround to deal with Microsoft issue in ReadDirectoryChanges notification
+			// If command prompt is used to write file continuously (e.g. ping -t 8.8.8.8 > ping.log)
+			// Then ReadDirectoryChanges does not detect the change
+			// Fortunately, notification is sent if right click or double click happens on that file
+			// Let's leverage this as workaround to enhance npp file monitoring functionality.
+			checkModifiedDocumentIfMonitoring();
```
donho: put the following cod instead of this function:
```
	Buffer* currBuf = getCurrentBuffer();
	if (currBuf && currBuf->isMonitoringOn())
		::PathFileExists(currBuf->getFullPathName());
```

SinghRajenM:
@donho Review comments have been accommodated.

> my apologies, the original 2 in 1 commit was my fault, not @SinghRajenM 's.

Actually, I wanted to keep interlinked issue in same PR but with different commits. So @MetaChuh, it's not your fault. Rather I would say it was my intention. Thank you for testing extensively and reporting issues.

Issue #5768, opened 2019-06-10
IDM_EDIT_CLEARREADONLY makes file readonly


briddums:
### Description of the Issue

Calling IDM_EDIT_CLEARREADONLY from a plugin on a file that is not readonly will cause the file to become readonly

### Steps to Reproduce the Issue

1. Create a plugin that can call IDM_EDIT_CLEARREADONLY
2. Open notepad++ with a file that is not read only
3. Run the plugin command

### Expected Behavior

The file should continue to be writable

### Actual Behavior

The file became read only

### Debug Information

Notepad++ v7.5.8   (32-bit)
Build time : Jul 23 2018 - 02:03:53
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll ScrollPastEOFUni.dll XMLTools.dll OppyNppPlugin.dll

Ekopalypse:
Seems not to be the case using PythonScript plugin and the following snippet
notepad.menuCommand(MENUCOMMAND.EDIT_CLEARREADONLY)
Files are still editable

Notepad++ v7.6.6   (64-bit)
Build time : Apr  3 2019 - 23:52:32
Path : D:\PortableApps\Npp\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll PythonScript.dll

briddums:
You did not follow my steps to reproduce.  You ran a snippet from another plugin, I don't know how the internals of PythonScript work.  It could be doing something different than directly calling IDM_EDIT_CLEARREADONLY.

The code within NppCommands.cpp > IDM_EDIT_CLEARREADONLY uses XOR:

```
dwFileAttribs ^= FILE_ATTRIBUTE_READONLY;
```

Therefore if the readonly flag is turned on, it will be turned off

```
0001
0001
----
0000
```

But if it is turned off, it will then be turned on
```
0000
0001
----
0001
```

Ekopalypse:
actually PythonScript calls IDM_EDIT_CLEARREADONLY via SendMessage but my error was to believe that
this is the READONLY feature which can be set via the tab context menu, instead it is the file attribute.
You are right, same happens with Pythonscript.
Sorry for confusion.

PR #5774, opened 2019-06-11
Fix bug in IDM_EDIT_CLEARREADONLY

briddums:
Using ^= for the readonly attribute will cause it to always change the bit.  So if CLEARREADONLY is called on a file that isn't read only, the file will become read only.

Changing it to &= ~FILE_ATTRIBUTE_READONLY; will result in it always clearing the readonly flag.


sasumner:
You are supposed to tie a PR to its issue.  Helper: #5768

Issue #5760, opened 2019-06-07
Plugins Admin: sort plugins by name or date


cmeriaux:
### Description of the Issue
In plugin admin UI, we can't sort elements

### Expected Behavior
Add a sort capability by name or release date

PR #5780, opened 2019-06-12
plugin admin date

### LINKED ISSUE #5742: Plugins Admin: add plugin release date
### Description of the Issue
Add in the plugin admin UI the plugins release date

cmeriaux:
add plugin release date. Fixes #5742
plugins can be sorted by name or date #5760

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -442,6 +442,7 @@
             <PluginsAdminDlg title="Plugins Admin" titleAvailable = "Available" titleUpdates = "Updates" titleInstalled = "Installed">
                 <ColumnPlugin   name="Plugin"/>
                 <ColumnVersion  name="Version"/>
+                <ColumnDate name="Date"/>
                 <Item id="5501" name="Search:"/>
                 <Item id="5503" name="Install"/>
                 <Item id="5504" name="Update"/>
--- a/PowerEditor/installer/nativeLang/french.xml
+++ b/PowerEditor/installer/nativeLang/french.xml
@@ -437,6 +437,7 @@
             <PluginsAdminDlg title="Gestionnaire des modules d'extension" titleAvailable = "Disponibles" titleUpdates = "Mises à jour" titleInstalled = "Installés">
                 <ColumnPlugin   name="Modules d'extension"/>
                 <ColumnVersion  name="Version"/>
+                <ColumnDate name="Date"/>
                 <Item id="5501" name="Rechercher:"/>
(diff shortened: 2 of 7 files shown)


rddim:
@cmeriaux 

Please add `<ColumnDate  name="Date"/>` to the `english.xml` for helping the translations. Must be under the `<PluginsAdminDlg title=....` section. Thanks

cmeriaux:
@rddim done

xylographe:
Thank you, Christophe.

Would it be possible to perform a case-insensitive sort when sorting by plug-in name?

ATM the few names that start with a lower case letter come at the bottom of the list,

PluginAdmin

xylographe:
It would also be nice — if technically possible — to have an indicator in the heading of the sort column.
For example,  **`Date     ▲`** for descending sort, and **`Date     ▼`** for ascending sort.

donho:
I don't think this new added information is pertinent.
By adding date info, we (and plugin authors),have more thing to control, and (a lot) more code for maintaining. So this feature won't be added.

cmeriaux:
@donho I disagree with your point. This is a nice feature to find quickly the new plugin or recent update.
On the nppPluginList  side, it's already implemented and it's already mandatory for json validation. No plugin author seems to be annoyed by this new field since June. 
For the first time, plugin admin would have a nice feature that old "plugin admin" didn't have.
Please reconsider your point.

donho:
Not because nppPluginList has such information that PluginAdmin should integrate such feature.
I just saw your merged commit:
https://github.com/notepad-plus-plus/nppPluginList/commit/a701bb505183dfe56e9391f2633a69c0f3ef8c2c

I would like to know how did you get all these release dates for the plugins? Are they accurate?

cmeriaux:
They are accurate. I got those dates in each plugin owner repository (usually github release or sourceforge). For some plugins, I found the release date in the old plugin admin xml files.

donho:
IMO, users don't care to learn a new released plugins, they need to find the ones they need. Or if users want to know the new update, they can go to Update tab to install them.
So "release date" is not even nice to have, it's just useless.

The commit notepad-plus-plus/nppPluginList@a701bb5 shouldn't be merged without this feature being integrated into Notepad++, therefore it will be reverted.

chcg:
@donho I'm not so happy about the revert. Even if the last release date might be not of interest to the user ( also requested, see https://github.com/bruderstein/nppPluginManager/issues/29) it is of interest to me to distinguish between plugins not updated for a long time and actively developed ones. Especially if an scintilla component update like with 7.7 happens the compatibility might break and the release date in this case is of interest to identify the plugins which should be tested again.

PR #5791, opened 2019-06-15
Update german.xml

schnurlos:
Small update, following https://github.com/notepad-plus-plus/notepad-plus-plus/commit/074a4bca9311374daef0750083b911fe48d0af5d


donho:
duplicated

Issue #4885, opened 2018-09-27
Tab dragdrop makes other tabs unusable


Wurstbrotc4:
### Description of the Issue
Dragging a tab to another position has the effect that other tabs can't be activated by clicking anymore. Additionally, dragging such a tab drags the first one instead (the one that was dragged before).

Without code change, it can either be resolved by closing and restarting the whole program, or by dragging the dragged tab back to its original position.

### Steps to Reproduce the Issue
I can always reproduce it in the latest version (7.5.8), without plugins. It works for me with older 6.x.x versions in the very same environment.

### Expected Behavior
After tab dragging, the tab bar still works 100%

### Pull request will be opened in some minutes...

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/895c315df0c5649bfdf4737472e6937242347358

AngryGamer:
I can't replicate this issue and no debug information is provided. Please provide more information about your setup so that someone can replicate the problem (such as OS, use of a compatibility layer like WINE, etc).

Also does [PR#5792](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5792) impact your problem in anyway?

PR #5792, opened 2019-06-16
Fix tab dragging on WINE and ReactOS

AngryGamer:


donho:
@AngryGamer I can reproduce #4885 under WINE.
But just applied your PR and test it under my WINE, it dosn't fix this issue :(

AngryGamer:
Alright, I was able to replicate #4885. This should fix it all!

donho:
@AngryGamer Bad news: the new modification in this PR doesn't change anything. After drag & drop a tab, it remains active (which is good), but I cannot activate the other tabs via mouse click. until I close the active one.

AngryGamer:
There's just no way. Are you sure your local build has all patches applied? Try the AppVeyor builds just in case. [win32](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/25346596/job/xa155uoiy4qmn2d5/artifacts) [x64](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/25346596/job/nd525m7nlv8sfqrs)

I was getting weird behavior under this setup, but it's fully functional with all patches applied.
`[user@localhost ~]$ cat /etc/fedora-release`
`Fedora release 29 (Twenty Nine)`
`[user@localhost ~]$ wine --version`
`wine-4.0.1`

AngryGamer:
Actually, I know what's wrong. I'll update the patch again.

AngryGamer:
This should do it.

donho:
@AngryGamer  Good news, it works with the latest modification.
But my concerning is that it might bring some regression ?

AngryGamer:
I think it's probably fine, but I can only vouch for Windows 10. I'd put it into a Release Candidate and let the masses tell us. 😊

LordNyriox:
@AngryGamer: 

>I'd put it into a Release Candidate and let the masses tell us. 

You can test the result using the PR's [AppVeyor build](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/25371531/job/qlgfctopj8b3g5o1/artifacts).  

Just replace the exe from your own Notepad++ 7.7 installation with the equivalent AppVeyor artifact.  But leave the dll alone, because the AppVeyor dll fails NPP's on-launch signature check.

PR #5808, opened 2019-06-20
Include additional Keyword Sets.

oirfeodent:


donho:
@oirfeodent What is your PR for ?
If it's for fixing a bug or adding a feature, please provide the issue number(s).
Or at least provide some description.

sasumner:
"Or at least provide some description."  <--- I would not have added this part.  The protocol is that all PRs start from "issues".  In theory, issues should be "accepted" before anyone does a PR.

PR #4861, opened 2018-09-20
Update Malayalam translation

rathodroshan0137:
notepad-plus-plus/PowerEditor/installer/nativeLang/Malayalam.xml

PR #5832, opened 2019-06-26
Zulu is not availbel on contributer page 

rathodroshan0137:
@chcg ,

Zulu language available in the editor of notepad plus but i can't find on https://notepad-plus-plus.org/contribute/binary-translations.html Binary Translations page.
Would you please check this screenshot ( https://www.screencast.com/t/EsraYsXBwzS9 ) editor of notepad plus

Would you please provide positive reply on this 

Thanks,


chcg:
@rathodroshan0137 Update of the webpage has currently started at https://github.com/notepad-plus-plus/npp-usermanual. Until now the https://github.com/notepad-plus-plus/npp-usermanual/blob/master/content/binary-translation/_index.en.md is empty.

What is the relation of this PR for https://en.wikipedia.org/wiki/Malayalam to Zulu? Seems the complain should be done in a separate issue independent from this PR.

Regarding this PR:
You are providing just the translation file. To be used by N++ also e.g. similar to zulu:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/localizationString.h#L119
and
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Parameters.cpp#L3944
need to be adapted.

donho:
> What is the relation of this PR for https://en.wikipedia.org/wiki/Malayalam to Zulu? Seems the complain should be done in a separate issue independent from this PR.

Indeed. I feel weird and confused that Malayalam translator claim Zulu should be in the translation page. 

And I do remember this rejected PR:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4861

We need people who know about Malayalam to valid this PR, instead of accepting it blindly.

xylographe:
@donho 
This is exactly the same file as in #4861, including `& quot;` errors.

donho:
I've checked myself, it's the same file as #4861.
@rathodroshan0137  Improve your translation or don't waste our time.

Issue #5822, opened 2019-06-24
Input characters might missing when encoding is ANSI


ahdung:
### Steps to Reproduce the Issue
1. make sure current encoding is ANSI
2. type 立马

### Expected Behavior
get 立马

### Actual Behavior
only 立

### Debug Information
Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\Users\Administrator\Desktop\npp.7.7.1.bin\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

donho:
Could you test with Notepad++ v7.7 to see if the same bug is reproduced?

zufuliu:
@donho this was fixed in Scintilla 4.1.5, you need update Scintilla to 4.1.5 or later.

https://sourceforge.net/p/scintilla/bugs/2093/
https://www.scintilla.org/ScintillaHistory.html

zufuliu:
Simply fix without upgrade Scintilla is copy Editor::AddCharUTF() (the block above NotifyChar) from latest Scintilla release.

Related commits for this bug (other commits on [#2093](https://sourceforge.net/p/scintilla/bugs/2093/) is not critical for this bug):
https://sourceforge.net/p/scintilla/code/ci/a7d3bc738c23d87e109c88a049c3a2b0ddad8da0/
https://sourceforge.net/p/scintilla/code/ci/1dcd0ca936187fa6564f2d24df99765c7e9092fc/

ahdung:
@donho v7.7 occur too.
@zufuliu Thank you, but I don't know what is Scintilla, a npp plugin? what should I do?

andrecool-68:
@ahdung This is the library that the notepad ++ works with. (This file is SciLexer.dll)

ahdung:
@andrecool-68 So I just need wait for v7.7.2 ?

zufuliu:
Since no one seems working on the copy and paste, I made PR #5834 for this,

ahdung:
@zufuliu Thank you.

donho:
@ahdung  @zufuliu  I just upgraded Scintilla from 414 to 420:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5934

Could you test both 32 and 64 build here:
https://notepad-plus-plus.org/temp/

[...]

PR #5834, opened 2019-06-26
Fix exception when inserting DBCS text in ANSI encoding, issue #5822.

zufuliu:
This was fixed in Scintilla 4.1.5, patch copied from Scintilla 4.1.7.


chcg:
Fix, see https://sourceforge.net/p/scintilla/bugs/2093/

zufuliu:
Update to Scintilla 4.1.7 should very smooth.
If npp use unmodified Scintilla (except CodePageFromCharSet in ScintillaWin.cxx), just copy all file from scintilla417.zip would finish the upgrade.

zufuliu:
Since the code no longer compatible with Scintilla 4.2.0 (released on today), I closed the PR. Please instead upgrade to Scintilla 4.2.0 or later.

PR #5845, opened 2019-06-27
'-nosession' must not overwrite config.xml (Fix #3938)

### LINKED ISSUE #3938: -nosession overwrites settings
### Description of the Issue

When `-nosession` is passed as argument, this setting is stored (and overwritten) when exiting the instance. However, `-nosession` should only be applied for the duration of the session and not stored permanently (if I wanted to have it permanently, I could just change the setting).

**I am fairly confident this used to work already** but suddently my settings are always reset when I start (a different instance) with `-nosession`. I think this bug must have introduced some time recently.

A similar issue existed for `-notabbar` (Bug #5274 on SF) and was fixed.

Background: I would like to use NPP as an advanced editor and as notepad replacement at the same time.

As an advanced editor (when I just execute notepad++.exe) I like the tabs, that my session is saved and that quick new documents are saved in the session althouth the file itself was not saved yet.

As a notepad replacement (which I use for Total Commander) I want a single session (new instance for each file), hence to tabs and and no session management.

For the first case I just use notepad++ as normal, enabling session restoration, tabs etc.

For the second case, when I use

`notepad++.exe -multiInst -nosession -notabbar`

It seems that the session setting is saved in the global configuration, i.e. it interferes with my "main advanced editor": When I close all instances and just open "notepad++.exe" again, the session setting is suddenly reset.

However, I want that "-nosession" only applies to the new and current instance I start with "notepad++.exe -multiInst -nosession -notabbar".

1. Start notepad++ without command line switches and turn session management on (in "Backup").
2. Create another instance with `notepad++.exe -multiInst -nosession -notabbar`
3. Close both instances
4. Open notepad++ again without command line switches and the session management setting is reset

### Expected Behavior
Command line switches should override configuration but never interfere with them

### Actual Behavior
Command line switches interfere with setting (i.e. settings get reset)

### Debug Information
Latest version (7.2.2).

MichaWiedenmann:
When `-nosession` is given as a command line parameter the attribute
`isSnapshotMode` in `config.xml` is overwritten with `no` regardless of
its original value.

    <GUIConfig name="Backup" action="0" useCustumDir="no"
               dir="" isSnapshotMode="yes" snapshotBackupTiming="7000" />

Change this to keep the original value.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -5679,7 +5679,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("useCustumDir"), _nppGUI._useDir ? TEXT("yes") : TEXT("no"));
 		GUIConfigElement->SetAttribute(TEXT("dir"), _nppGUI._backupDir.c_str());
 
-		GUIConfigElement->SetAttribute(TEXT("isSnapshotMode"), _nppGUI.isSnapshotMode() ? TEXT("yes") : TEXT("no"));
+		GUIConfigElement->SetAttribute(TEXT("isSnapshotMode"), _nppGUI._isSnapshotMode ? TEXT("yes") : TEXT("no"));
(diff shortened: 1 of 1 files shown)


xylographe:
Glad to see this one make the ‘cut’. :)

PR #5853, opened 2019-06-28
Load plug-ins with LoadLibraryEx

### LINKED ISSUE #5802: Allow extra plugin DLL files to be loaded automatically
### Description of the Issue
Many plugins come with their own DLL files needed for the plugin to operate correctly. 

Normally [Run-Time Dynamic Linking](https://docs.microsoft.com/en-us/windows/desktop/dlls/using-run-time-dynamic-linking) would be required by the plugin to use local DLL files within the plugin's folder. However, in cases where it is not possible to do Run-Time Dynamic (due to a large number of exported functions, such as PythonScript), the extra DLL files must be located in the same directory as `notepad++.exe` (**note** that the Plugin Admin does not support copying DLL files to the root installation directory).

### Proposed Feature
Enable [Load-Time Dynamic Linking](https://docs.microsoft.com/en-us/windows/desktop/dlls/using-load-time-dynamic-linking) (required DLLs are found and loaded by the OS) when loading plugins. Notepad++ would have to slightly change the way it loads plugins so that any required DLL files are allowed to be loaded from the plugins folder. 

---

CC @ClaudiaFrank @xylographe

xylographe:
Allow plug-ins to load private DLL files from the plugins folder.

Close #5802

Loading plug-ins with `LoadLibraryEx` and the `LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR` flag allows for implicitly linked (Load-Time Dynamic Linking) plug-ins, which is currently not possible without user intervention.

To stay on the side of caution `LoadLibrary` is still used when running under Windows XP, while `LoadLibraryEx` is used when running under a more recent version of Windows.

Regarding compatibility, Windows XP users can still use implicitly linked plug-ins if they add the plug-in's directory (e.g. `%ProgramFiles%\Notepad++\plugins\LuaScript`) to the PATH environment variable.

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -114,6 +114,9 @@ cleanup: // release all of our handles
 	return machine_type;
 }
 
+#define	LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR	0x00000100
+#define	LOAD_LIBRARY_SEARCH_DEFAULT_DIRS	0x00001000
+
 int PluginsManager::loadPlugin(const TCHAR *pluginFilePath)
 {
 	const TCHAR *pluginFileName = ::PathFindFileName(pluginFilePath);
@@ -130,7 +133,8 @@ int PluginsManager::loadPlugin(const TCHAR *pluginFilePath)
 		if (GetBinaryArchitectureType(pluginFilePath) != ARCH_TYPE)
 			throw generic_string(ARCH_ERR_MSG);
 
-	    pi->_hLib = ::LoadLibrary(pluginFilePath);
+        const DWORD dwFlags = GetProcAddress(GetModuleHandle(TEXT("kernel32.dll")), "AddDllDirectory") != NULL ? LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS : 0;
+        pi->_hLib = ::LoadLibraryEx(pluginFilePath, NULL, dwFlags);
         if (!pi->_hLib)
         {
 			generic_string lastErrorMsg = GetLastErrorAsString();


chcg:
See https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-loadlibraryexw for further infos on the used parameters and the restriction regarding supported windows versions.
@xylographe  Couldn't be libloaderapi.h (include Windows.h) included instead of the defines LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR and LOAD_LIBRARY_SEARCH_DEFAULT_DIRS ? Are the defines not available with the platformset v140_xp\v141_xp

xylographe:
> Are the defines not available with the platformset v140_xp\v141_xp

No, that is the first thing I tried, and the flags that are not supported by Windows XP are not available.

I had to choose between converting from v140_xp to v140, or adding the defines. I believe the latter might be more acceptable than the former.

CookiePLMonster:
`LoadLibraryEx` is available for XP, so I guess logic can be simplified by just zero-ing `LoadLibraryEx` flags for XP.

Also how @chcg said it's better to query for this feature by trying to `GetProcAddress` specific functions instead of checking for XP/Vista presence.

xylographe:
> it's better to query for this feature by trying to GetProcAddress specific functions 

I don't understand. **`LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR`** and **`LOAD_LIBRARY_SEARCH_DEFAULT_DIRS`** are not functions. How would I use `GetProcAddress` to determine whether these flags are supported?

CookiePLMonster:
From MSDN's LoadLibraryEx page:

> Windows 7, Windows Server 2008 R2, Windows Vista and Windows Server 2008:  The LOAD_LIBRARY_SEARCH_ flags are available on systems that have KB2533623 installed. To determine whether the flags are available, use GetProcAddress to get the address of the AddDllDirectory, RemoveDllDirectory, or SetDefaultDllDirectories function. If GetProcAddress succeeds, the LOAD_LIBRARY_SEARCH_ flags can be used with LoadLibraryEx.

xylographe:
@CookiePLMonster 
Sorry, a slight misunderstanding. I thought you were referring to Windows XP.

CookiePLMonster:
That also extends to WinXP - it supports `LoadLibraryEx` so just like on later Windows you just need to check for those functions - that is, you don't need to check for OS version ever.

xylographe:
Thank you, @CookiePLMonster 

New commit: Use `GetProcAddress` i/o `getWinVersion` to determine availability of the LOAD_LIBRARY_SEARCH_ flags.

Ekopalypse:
Test revealed that all of the plugins currently available via plugins admin behave the same concerning LoadLibray(LL) vs LoadLibraryEx(LLEx) usage.

This doesn't mean, all plugins work properly but rather that if the plugin doesn't work properly with LL it doesn't work with LLEx  either and vice versa but still behave the same, meaning, if CustomerLinenumbers didn't show all line numbers using LL it showed the same line numbers using LLEx, still not correct but the same as with LL.

But test revealed another area of issues, namely, plugin incompatibilities. For example it happened that npp sometimes crashed while using Bookmarks@Dook with all other plugins loaded, but didn't crash if I tested Bookmarks@Dook alone or like PoorMansTSqlFormatter caused Npp to crash on exit with all plugins loaded but not if used it alone ... (?!?)
This happened with npp using  LL as well as with npp using LLEx.

Note, I cannot say if these two plugins are the cause of the problem or whether they just suffer from a behavior of another plugin.

So, from my point of view changing library loading seems not to introduce further issues.

PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
```diff
@@ -114,6 +114,9 @@ static WORD GetBinaryArchitectureType(const TCHAR *filePath)
 	return machine_type;
 }
 
+#define	LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR	0x00000100
+#define	LOAD_LIBRARY_SEARCH_DEFAULT_DIRS	0x00001000
```
donho: It's strange that you have to define 2 symbols yourself - is there no header files to be included ?

donho: nm, I read the previous comment.

PR #5883, opened 2019-07-04
Fix Notepad commandline style parsing

### LINKED ISSUE #5861: Double Click in Windows opens the wrong file
In Windows Explorer I have a folder with two files in a new folder:

new 2.txt     (this has one space in the filename)
new  2.txt    (this has two spaces in the filename)

When I double-click on the filename with two spaces, Notepad++ (v7.7.1) loads the one with the single space in the filename.

My system:  Win8.1, Notepad++ 7.7.1, Disk is an SSD with NTFS filesystem.
My test folder is in the Documents folder.

I was comparing these two text files and was about to delete one of them because they appear to contain the same contents, however I noticed that one was 1kB and the other 5kB so I looked into it more and found this bug.

CookiePLMonster:
This PR refines `-notepadStyleCmdline` parsing to e.g. properly account for filenames which have multiple spaces in the file name.

To verify, consider this case:
1. Have Notepad++ replace Notepad, as described here: https://notepad-plus-plus.org/features/replace-notepad.html
2. Create a file called `file name.txt` (with one space).
3. Create a file called `file  name.txt` (with two spaces).
4. In the directory containing those files, run `notepad file  name` (with one space).

### Observed result

`file name.txt` (with one space) is opened in Notepad++.

### Expected result

`file  name.txt` (with two spaces) is opened in Notepad++.

Fixes #5861

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -81,70 +81,48 @@ void allowWmCopydataMessages(Notepad_plus_Window& notepad_plus_plus, const NppPa
 ParamVector parseCommandLine(const TCHAR* commandLine)
 {
 	ParamVector result;
-	int numArgs = 0;
-	LPWSTR* tokenizedCmdLine = CommandLineToArgvW( commandLine, &numArgs );
-	if ( tokenizedCmdLine != nullptr )
+	if ( commandLine[0] != '\0' )
 	{
-		result.assign( tokenizedCmdLine+1, tokenizedCmdLine+numArgs ); // If numArgs == 1, it will do nothing
-		LocalFree( tokenizedCmdLine );
+		int numArgs;
+		LPWSTR* tokenizedCmdLine = CommandLineToArgvW( commandLine, &numArgs );
+		if ( tokenizedCmdLine != nullptr )
+		{
+			result.assign( tokenizedCmdLine, tokenizedCmdLine+numArgs );
+			LocalFree( tokenizedCmdLine );
+		}
 	}
 	return result;
-}
-
-// Looks for -z arguments and strips command line arguments following those, if any
-void stripIgnoredParams(ParamVector & params)
-{
-	for ( auto it = params.begin(); it != params.end(); )
-	{
-		if (lstrcmp(it->c_str(), TEXT("-z")) == 0)
-		{
-			auto nextIt = std::next(it);
-			if ( nextIt != params.end() )
-			{
-				params.erase(nextIt);
-			}
-			it = params.erase(it);
-		}
-		else
-		{
-			++it;
-		}
-	}
 }
 
 // 1. Converts /p to -quickPrint if it exists as the first parameter
 // 2. Concatenates all remaining parameters to form a file path, adding appending .txt extension if necessary
 // This seems to mirror Notepad's behaviour
-void convertParamsToNotepadStyle(ParamVector & params)
+ParamVector convertParamsToNotepadStyle(PWSTR pCmdLine)
 {
-	ParamVector newParams;
-	auto it = params.begin();
-	if ( it != params.end() && lstrcmpi(TEXT("/p"), it->c_str()) == 0 ) // Notepad accepts both /p and /P, so compare case insensitively
+	ParamVector params;
+	if ( _tcsnicmp(TEXT("/p"), pCmdLine, 2) == 0 ) // Notepad accepts both /p and /P, so compare case insensitively
 	{
-		++it;
-		newParams.emplace_back(TEXT("-quickPrint"));
+		params.emplace_back(TEXT("-quickPrint"));
+		pCmdLine += 2; // Length of "/p"
 	}
 
-	bool ssHasContents = false;
-	generic_stringstream ss;
-	for ( ; it != params.end(); ++it )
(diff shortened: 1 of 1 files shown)


madivad:
I'm slightly confused why you would expect the filename with two spaces to be opened when you'e specifically asked for one. 

If you had of designed the test for failure in the opposite direction (asking for the filename with two spaces and only getting the one with ONE space, THEN I'd understand). Can you elaborate?

CookiePLMonster:
> I'm slightly confused why you would expect the filename with two spaces to be opened when you'e specifically asked for one.

GitHub folds those multiple spaces into one, hence me commenting those with (one space) and (two spaces).


> perusing this code, why is the command line tokenised to begin with? Why not accept the filenames as presented by the operating system?

It's tokenized as if it was passed as `argc` and `argv`. Notepad++ has numerous commandline arguments, so this is pretty much a necessity.

And no, this very pull request addresses issues occuring when a filename is passed **not** escaped. You may consider this a bogus use case, but Windows Notepad handles it correctly (as per my notes), and the entire purpose of `-notespadStyleCmdline` switch is to support Notepad commandline arguments perfectly, no matter how weird they may be.

xylographe:
> GitHub folds those multiple spaces into one

Not if you use U+00A0 (NO-BREAK SPACE),
 `file  name.txt` or
 `file name.txt`. :)

PowerEditor/src/winmain.cpp
```diff
@@ -101,6 +105,21 @@ void stripIgnoredParams(ParamVector & params)
...
 			{
+				PWSTR ignoredString = wcsstr(pCmdLine, nextIt->c_str());
+				if ( ignoredString != nullptr )
+				{
+					ignoredString += nextIt->length();
+
+					// Advance to the first non-whitespace and not quotation mark character
+					while ( iswspace( *ignoredString ) || *ignoredString == L'"' )
+					{
+						if ( *ignoredString == '\0' )
```
AngryGamer: Does this do anything? I don't see either while condition letting a NULL character through.

PowerEditor/src/winmain.cpp
```diff
@@ -110,41 +129,37 @@ void stripIgnoredParams(ParamVector & params)
...
-	bool ssHasContents = false;
-	generic_stringstream ss;
-	for ( ; it != params.end(); ++it )
+	// Advance to the first non-whitespace character
+	while ( iswspace( *pCmdLine ) )
 	{
-		ssHasContents = true;
-		ss << *it;
-		if ( std::next(it) != params.end() ) ss << TEXT(" ");
+		if ( *pCmdLine == '\0' )
```
AngryGamer: ditto

PowerEditor/src/winmain.cpp
CookiePLMonster: Good point - according to https://en.cppreference.com/w/cpp/string/wide/iswspace null terminator is not considered whitespace, and thus it should be out of this loop by then. Will correct tomorrow.

PowerEditor/src/winmain.cpp
```diff
@@ -313,10 +328,10 @@ void doException(Notepad_plus_Window & notepad_plus_plus)
 
 
 
-int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE, LPSTR, int)
+int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
```
donho: Why do we need `wWinMain` ?

CookiePLMonster: I need access to a wide char command line parameter. `GetCommandLineW` does fit my use case, because it also includes the program name. See:

```
lpCmdLine

Type: LPSTR

The command line for the application, excluding the program name. To retrieve the entire command line, use the GetCommandLine function.
```

CookiePLMonster:
Oh my, oh my! I approached this code intending to resolve concerns raised in a review and I was in shock. Have I even tested this!?

First of all, this commit appears to be bigger than it actually is, because I had to move `stripIgnoredParams` elsewhere. Here's what I fixed, though:
1. `-notepadStyleCmdline` was not stripped from a joint filename path, only `-z` was That means `notepad++.exe -notepadStyleCmdline file name.txt` attempted to open `-notepadStyleCmdline file name.txt`. Now fixed.
2. `.txt` was appended even if there was no filename. That means `notepad++.exe -notepadStyleCmdline` attempted to open `.txt`. Now fixed.
3. Skipping over commandline parameters was not checking if entire words were matched. This means `notepad++.exe filenamewith-z-inparam -z paramtoskip` would be totally broken - when attempting to skip `-z`, it matched the one in `filenamewith-z-inparam`. Now fixed.

This is pretty embarrassing. I hope it's all fine now.

donho:
@CookiePLMonster  Indeed, I was thinking "This function is quite complicated for the divers features, let's uni-test it. to avoid the regression". I installed gTest via Visual Studio and begin to do some tests, then I realized the test part should be a lib or a dll, so I quit. If there's the unit-test framework which doesn't need to modify the structure of Notepad++ part. I'm all for unit-test but I hate to modify the main project's structure for adapting the tests.

donho:
Tried also catch2 without success. I quit for unit-test for the moment.

PR #5887, opened 2019-07-05
[Cosmetic changes]: Yes/No to All dlg enhancement

SinghRajenM:
1. Resize the `LTEXT` width and height
    * Width is increased to fit the full file name
    * Height is reduced, but it is sufficient for max allowed file name
2. Adjust the buttons positions
3. Remove minimize and maximize button from sys menu

## Comparison
image

### Max file name:
image


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4ebcf6191701963566ba575a65a11a56944570f9

Issue #5666, opened 2019-05-17
The Read Only icon is not distinguishable when a tab becomes inactive.


dyura:
NPP uses one of 3 icons inside the tabs. All 3 have the same drawing but different colors: red for unsaved buffer, blue for a saved buffer, and grey for read-only.   When a buffer is  inactive the icon is dim. It is almost impossible to distinguish between dim blue and grey icons. I just would suggest using another drawing for Read-Only. For example, I re-compiled  using readonly.ico.old from Notepad++ repository, and the visibility in my version is perfect now.  readonly.ico.old  is already in the repository

SinghRajenM:
Duplicate of #4492 and #3170. A try was given at https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5097.

An another try here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5890

dyura:
Thanks for the links to these duplicates. I see that at least four more people pointed to this bug. However, this bug still exists, and we still cannot distinguish inactive tabs as read-only. I understand that this is not a critical bug, but it is VERY EASY to correct.
What is the problem? All 3 icons, read-only, saved, and unsaved, have the same style, but different colors. Practically the difference between gray and blue, when the icon is darkened, is not noticeable . My suggestion is to change the read-only icon, while preserving the accepted style as much as possible. Let's make the upper right corner black. You can see my attached example of this icon.
![readonly ](https://user-images.githubusercontent.com/39743460/61194890-ab3f5000-a692-11e9-8875-14d0ba6ea7bb.png)
 My graphic design skills are bad. But even with my bad skills, I believe that the current icon style symbolizes a floppy disk. Let's think that the black corner symbolizes a write-protected floppy disk. Look at the screenshot when I used this icon. As for me, the style is almost the same, but the black spot in the upper corner allows us to perfectly distinguish read-only tabs.
![unnamed](https://user-images.githubusercontent.com/39743460/61194894-b1cdc780-a692-11e9-9908-d14cb0d05684.png)

dyura:
My previous post was not answered. After my post, I noticed  that  there was another pull request, [#5890](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5890 ) , intending to change the read only icon. This request was rejected with a comment: “The colour looks weird.”
[...]

PR #5890, opened 2019-07-05
Enhancement: Make readonly tab identifiable even if tab is not active

SinghRajenM:
Fixes: #4492, #3170 and #5666

Previous PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5097. This time, tried to kept the icon same just by changing the color.

image


donho:
The colour looks wired.

PR #5892, opened 2019-07-06
Fix typos in change.log

Atalanttore:


donho:
change.log has been distributed with new release 7.7.1. So this fix won't take any effect.
That says, I'll install spelling checker plugin on my build machine for the release note.

PR #5927, opened 2019-07-18
Improve GUI in Find dialog

donho:
Thanx to:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/449#issuecomment-135799762


donho:
@MAPJe71 
> Shouldn't the text on the button (i.e. "▼ Find Next") be translated?
it's in english.xml already.

rddim:
The border around the `Count` button should be fixed

before
npp_before

after
npp_after

donho:
@rddim You are in Windows 7, right?
Could you check it under Windows 10 please?

MAPJe71:
See following picture using Dutch translation:

- The border (i.e. group indication) around `In selection` and the `Count`-button is also in the wrong position on Windows 10.
- Allthough there is a translation for "Find Next" in `dutch.xml` it's used for the 1-button search-mode. The "▼ Find Next" for 2-button search mode is NOT translated i.e. no entry in `dutch.xml` and `english.xml`. Furthermore, when I remember correctly the text on the 'previous' and 'next' buttons was replaced with arrows w/o text to prevent discussion on the translated text not fitting on the buttons for some/several languages.

Capture-2019-07-19-22 27 20

xylographe:
@MAPJe71
Dutch translation for items 1721/1723 (Dialog.Find) added in commit e36bd31bf604bcef99fabb39673979bdf23efc11.

I fully agree that the equally sized **`<<`** and **`>>`** buttons are vastly superior to this new stuff.

rddim:
@donho 
Yes, Windows 7
No, I can't check under Windows 10

PR #5934, opened 2019-07-21
Upgrade Scintilla from 4.1.4 to 4.2.0

donho:


GOLEM777:
I think this is a good idea.
Better to update than to edit.

donho:
Please test both 32 and 64 build here:
https://notepad-plus-plus.org/temp/

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/92bad0a60ad606b30df9ed97aecf4ff27bb6e967

GOLEM777:
Hello! ANSI feels good (typing and saving). At the moment everything is working and stable. FOR ME!!
Win10 (64).

Issue #5958, opened 2019-07-27
Error when building SciLexer.dll with boost


Omcsesz:
Hello, Team!

When I try to build SciLexer.dll with Boost 1.70.0 using the following command:

`nmake BOOSTPATH=C:\github\boost\boost_1_70_0\ BOOSTREGEXLIBPATH=C:\tmp`

the build fails with the following error:

```
Microsoft (R) Program Maintenance Utility Version 14.16.27032.1
Copyright (C) Microsoft Corporation.  All rights reserved.

        cl -Zi -TP -MP -W4 -EHsc -std:c++17 -D_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES=1 -D_CRT_SECURE_NO_DEPRECATE=1 -D_SC_SECURE_NO_WARNINGS=1  -O1 -MT -DNDEBUG -GL -I../include -I../src -I../lexlib -IC:\github\boost\boost_1_70_0\ -DSCI_OWNREGEX -D_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS -c ../boostregex/BoostRegexSerch.cxx
Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27032.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

BoostRegexSearch.cxx
../boostregex/BoostRegexSearch.cxx(12): fatal error C1083: Cannot open include file: 'stdlib.h': No such file or directory
NMAKE : fatal error U1077: '"c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64\cl.EXE"' : return code '0x2
'
Stop.
```
Can you help me in this?

Ekopalypse:
Do you know this?
https://github.com/bfridkis/Building-Notepad-Plus-Plus-from-Source-A-Beginners-Guide

Omcsesz:
No, not yet.
This seems promising, thanks!

donho:
Non related with Notepad++ project - PBKAC.
Suggestion; Search before asking. Google is your good friend.

Omcsesz:
I thought the readme.md is up-to-date in this regard.

sasumner:
> I thought the readme.md is up-to-date in this regard.

I guess when you find the solution, if `readme.md` is not accurate in that regard, submit a pull-request to make it right!

Omcsesz:
That's true!

PR #5959, opened 2019-07-27
Update Readme.md

Omcsesz:
The older, cmd method has been replaced with Visual Studio-based method, hence the update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,68 +28,8 @@ https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 To build Notepad++ from source:
 -------------------------------
 
-There are two components that need to be built separately:
-
- - `notepad++.exe`: (depends on `SciLexer.dll`)
- - `SciLexer.dll` : (with nmake)
-
-You can build Notepad++ with *or* without Boost - The release build of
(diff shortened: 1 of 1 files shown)


donho:
The readme has been updated recently and as fas as I know, nobody has the problem with it.

Omcsesz:
Actually, it has been suggested to me here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5958

donho:
The suggestion told you if you find any inaccurate description, you're free to do a PR to correct it.
Not erase the whole section to put a link for people who have never done dev.

sasumner:
> it has been suggested to me here

It would have been more appropriate to take any clarifying points you found at that external link and integrate them into readme.md to make it a more accurate and cohesive document.

Omcsesz:
> 
> 
> > it has been suggested to me here
> 
> It would have been more appropriate to take any clarifying points you found at that external link and integrate them into readme.md to make it a more accurate and cohesive document.

True, I will do that.

sasumner:
@Omcsesz You are very agreeable; hopefully if you put the effort in, it will be accepted and all will benefit if the way it is now is deficient.  But, I for one have no control over whether or not that will happen.

PR #5982, opened 2019-07-31
Update french.xml to 7.7.2

xylographe:
- Add missing entries.
- Remove surplus entries.
- Modify encoding menu entries according to #5951.


donho:
@xylographe 
Here is the error message:
```
$ git checkout -b xylographe-pr/french master
fatal: A branch named 'xylographe-pr/french' already exists.
```
When I said "branch per PR", it means you should, for each PR you wanna merge, create a new branch with the different name. 
I would give `update_french_190731` as name to this branch.

Sorry for the inconvenience, thank you for your comprehension.

xylographe:
Sorry, slight misunderstanding. I wasn't aware that you keep the old PR branches, I thought something had gone wrong with the sqash/rebase, hence I removed the old PR branch, and created a new one, but using the same name again. :)

PR #5989, opened 2019-08-01
[Installer] Improve configurability of shortcuts creation

vlakoff:
* If user is admin: allow choosing (default: create for all users).
* If user is not admin: hide controls, create for current user.


vlakoff:
UI has to be improved a bit…

installer-shortcuts

vlakoff:
Ahhhh, much better:

installer-shortcuts-v2

vlakoff:
It _seems_ to work, though some reviews would be greatly appreciated.

A further improvement could be to remember controls' states when navigating page back and forth again.

donho:
This feature might be interesting for few people for taking risk of regression of installer.

vlakoff:
I have managed to craft a fully functional page (including the "restore state" feature), and with code actually robuster than the current one.

But agree with keeping the code of this page as simple as possible. Even the "restore state" feature alone (just for the 2 checkboxes of the current page) significantly increases the code complexity.

PR #5995, opened 2019-08-02
Make 'Clear all marks' and 'Inverse Bookmark' macroable (fixes # 1948)'

### LINKED ISSUE #1948: Marking issues in recorded macro
**Have the following text:**
This is a text is want to keep
This is a text is want to remove
This is a text is want to keep
This is a text is want to keep
This is a text is want to remove
This is a text is want to keep

I.
Start recording a macro.
1. Go at the beginning of the text
1. Go to Find and Replace, Mark tab, check Bookmark line option and enter text _keep_ in Find what area
2. Press Mark all (leave window open) -> lines 1,3,4,6 are marked 
3. Go to Search - Bookmark - Remove unmarked lines -> lines 2 and 5 are removed
4. Press Clear all marks.  -> all marks are cleared
5. Close window and stop recording the macro.
Restore the original text and run the macro.

Expected: 
_This is a text is want to keep_ appears on 4 lines, all of which are **unmarked**,

Actual:
_This is a text is want to keep_ appears on 4 lines, all of which are **marked** 

II. 
**Have the same text.**
Start recording a macro.
1. Go at the beginning of the text
1. Go to Find and Replace, Mark tab, check Bookmark line option and enter text _keep_ in Find what area
2. Press Mark all (leave window open) -> lines 1,3,4,6 are marked 
3. Go to Search - Bookmark - Inverse bookmark -> lines 2 and 5 are marked, lines 1,3,4,6 are unmarked 
4. Go to Search - Bookmark - Remove bookmarked lines. -> Lines 2 and 5 are removed
5. Press Clear all marks.  -> all marks are cleared
6. Close window and stop recording the macro.
Restore the original text and run the macro.

Expected: 
**_This is a text is want to keep_ appears on 4 lines**, all of which are unmarked,

Actual:
**_This is a text is want to remove_ appears on 2 lines** 

System details: Intel i5 at 2.6 GHz, 4GB RAM, Windows 7 Ultimate 64 bit
Application version: Notepad++ v6.9.2

sasumner:
Fixes #1948 .

1948 describes two problems with bookmarking and macros, split into Case I (clear marks) and Case II (invert bookmarks).  This PR addresses both problems.  The Case I fix is entirely in `FindReplaceDlg.cpp` and the Case II fix is fully in `NppCommands.cpp`.

This is my second try at a PR for this, so fingers crossed.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -1331,6 +1331,7 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 				{
 					if (_currentStatus == MARK_DLG)
 					{
+						if (isMacroRecording) saveInMacro(wParam, FR_OP_FIND);
 						(*_ppEditView)->clearIndicator(SCE_UNIVERSAL_FOUND_STYLE);
 						(*_ppEditView)->execute(SCI_MARKERDELETEALL, MARK_BOOKMARK);
 						setStatusbarMessage(TEXT(""), FSNoMessage);
@@ -2509,7 +2510,7 @@ void FindReplaceDlg::saveInMacro(size_t cmd, int cmdType)
 {
 	int booleans = 0;
 	::SendMessage(_hParent, WM_FRSAVE_INT, IDC_FRCOMMAND_INIT, 0);
-	::SendMessage(_hParent, WM_FRSAVE_STR, IDFINDWHAT,  reinterpret_cast<LPARAM>(_options._str2Search.c_str()));
+	::SendMessage(_hParent, WM_FRSAVE_STR, IDFINDWHAT,  reinterpret_cast<LPARAM>(cmd == IDC_CLEAR_ALL ? TEXT("") : _options._str2Search.c_str()));
 	booleans |= _options._isWholeWord?IDF_WHOLEWORD:0;
 	booleans |= _options._isMatchCase?IDF_MATCHCASE:0;
 	booleans |= _options._dotMatchesNewline?IDF_REDOTMATCHNL:0;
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -2764,6 +2769,13 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 						break;
 					}
 
+					case IDC_CLEAR_ALL:
+					{
+						(*_ppEditView)->clearIndicator(SCE_UNIVERSAL_FOUND_STYLE);
```
LordNyriox: @sasumner:  I am probably wrong on this, but are you sure this line is not supposed to be `SCI_UNIVERSAL_FOUND_STYLE`?

xylographe:
Ty, Scott. Works as ‘advertised’.

sasumner:
I don't see it anywhere in here, but I got emailed by Github with this:

@LordNyriox commented on this pull request.

In PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp:

> @@ -2764,6 +2769,13 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 						break;
 					}
 
					case IDC_CLEAR_ALL: 
					{
						(*_ppEditView)->clearIndicator(SCE_UNIVERSAL_FOUND_STYLE);
@sasumner: I am probably wrong on this, but are you sure this line is not supposed to be SCI_UNIVERSAL_FOUND_STYLE?

My response:

If you'll notice in the commit where I added `if (isMacroRecording) saveInMacro...` you can see the line right below (which I didn't add) uses the same code you question above.  This is the interactive (non-macro) code that runs when the user presses the `Clear all marks` button.  My added line for this, in the code that executes when a macro is being run, is identical.  So yes, I think the usage of `SCE_...` is correct.

xylographe:
@sasumner 
> I don't see it anywhere in here,

Presumably, deleted later on, perhaps, because SCI_UNIVERSAL_FOUND_STYLE doesn't exist? :-)

LordNyriox:
@sasumner: 

>I don't see it anywhere in here, but I got emailed by Github with this 

I deleted the comment once I took a second look at the existing code surrounding the PR, specifically line `1335` of `FindReplaceDlg.cpp`. 

Unfortunately, it is not possible for me to delete e-mail notifications of my stupid comments.  ;}

sasumner:
Late thoughts:  

Note that support for "Clear all marks" was added, not a bug fix to existing functionality.  Of course, one could consider it a bug fix because the original support was missing when it should have been there?!  :)

Thus, the macro code number for "Clear all marks" (which is 1633) will have to be added to the documentation.  Since at the moment, documentation doesn't exist any longer, I will be creating some (macro-encoding of Find-family commands) based on the old wiki pages.

PR #5998, opened 2019-08-03
[Installer] Fix initial checkbox state that was not handled

vlakoff:
Follow-up to #5989.

As I guessed, the programmatic initial check of the checkbox was ineffective (unless the user clicks the checkbox).

I tried putting `NSD_OnClick` before `NSD_Check`, no change, and replacing `OnClick` with `OnChange`, event doesn't work at all.

`${NSD_Check}` just doesn't trigger that "click" event. Therefore the solution is to manually call the event handler.

### DIFF
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -112,6 +112,7 @@ Function ExtraOptions
 	${If} $1 == "$PROGRAMFILES"
 	${ORIF} $2 == "$PROGRAMFILES64"
 		${NSD_Uncheck} $NoUserDataCheckboxHandle
+		Call OnChange_NoUserDataCheckBox
 		EnableWindow $NoUserDataCheckboxHandle 0
 	${Else}
 		EnableWindow $NoUserDataCheckboxHandle 1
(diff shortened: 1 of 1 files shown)


donho:
Is there an issue to report that?

vlakoff:
I just tested the installer, found a bug and fixed it.

sasumner:
Proper procedure is this:

* Create an issue and discuss the "bug" or whatever (feature request, etc).
* The community will then use the issue comments to discuss it
* Ideally it gets marked as "accepted", if there is an agreeable resolution, but in practice this isn't mandatory
* A PR is created that fixes or otherwise implements the issue (sort of risky -- could be a waste of your time -- if the issue wasn't accepted)
* (Ideally) PR is accepted and merged, leading to an improved Notepad++ ... but this is not guaranteed to happen, even when official team members do the code changes.

vlakoff:
But, I'm simply fixing a bug… 😭

I'll just link this: [Process by Management](https://thedailywtf.com/articles/process-by-management), I found it funny.
(but sadly it's a true and common story 😭)

sasumner:
@vlakoff 

TL;DR

If you want to contribute, you have to follow the process for THIS project.

PR #6001, opened 2019-08-04
Make Combobox font monospace in Find dialog

donho:
Close ##5920, close #3361


xylographe:
Works fine for me!
Just wondering if this won't be a problem for Asian users. Courier New seems to include glyphs for Latin, Greek, Arabic, and Cyrillic scripts, but I don't see support for Thai, Chinese, Japanese, etc.?

donho:
@xylographe Good point. I have tested with Chinese Traditional under my Windows 10 in English. It seems no problem to me.
Let's wait & see.

Ede123:
What changed since #3337?

I guess there are at least as many reasons to keep previous behavior. There are just no issues requesting a non-monospaced font because - well - it was the default.

Issue #5907, opened 2019-07-08
Queue overflow bug in 7.7


AnReZa:
Bug #3453 can be reproduced again in 7.7 and 7.7.1. Please reopen the original ticket, or fix it in this one,

Issue #3740, opened 2017-09-15
Suggestion: Use lazy loading for Folder as Workspace component


samek-h:
### Description of the  Issue
Folder as Workspace component appears to be using recursive traversal on the tree of open folders even when the folders are collapsed. This might cause a behavior described in other reported issues, e.g. #1566, #2203, #2519,  #3453.

### Steps to Reproduce the Issue
1. Use "Open Folder as Workspace" from "File" menu for several folders containing a large number of files and nested directories.
2. Try to delete or change a large number of files inside the open folders.

### Expected Behavior
No or reduced blinking caused by tree refreshing. No error dialog on filesystem updates.

### Actual Behavior
A frequent blinking appears, in some cases followed by "Queue overflowed" error dialog.

PR #6005, opened 2019-08-05
Fix crash when too many directory changes happen [without refact]

xylographe:
@donho
This is the rebased, un-refactored version of PR #3793.

I don't know if this fixes the problem it claims to fix. I never experienced that problem. All I know is that this branch builds successfully.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -39,7 +39,6 @@
 
 using namespace std;
 
-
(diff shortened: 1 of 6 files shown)


SinghRajenM:
@xylographe , even I haven't seen the crash, but "Queue overflow" error occurs. And the reason was "Semaphore" with max limit of 1000. There are many defect reported about it. Few of them are #5907, #3740 and many more has been closed recently as part of old outdated defect closure.


And, converting Semaphore to Event fixes "queue overflow" issue.

PR #6019, opened 2019-08-08
Fix EncodingMapper singleton memory leak

wjx0912:
Fix EncodingMapper singleton memory leak

### DIFF
--- a/PowerEditor/src/EncodingMapper.cpp
+++ b/PowerEditor/src/EncodingMapper.cpp
@@ -83,8 +83,6 @@ static EncodingUnit encodings[] = {
     {20866, "koi8_r csKOI8R"}                                                                   //IDM_FORMAT_KOI8R_CYRILLIC
 };
 
-EncodingMapper * EncodingMapper::_pSelf = new EncodingMapper;
-
 bool isInListA(const char *token, const char *list) {
 	if ((!token) || (!list))
 		return false;
--- a/PowerEditor/src/EncodingMapper.h
+++ b/PowerEditor/src/EncodingMapper.h
@@ -35,7 +35,10 @@ struct EncodingUnit {
 
 class EncodingMapper {
 public:
-    static EncodingMapper * getInstance() {return _pSelf;};
+    static EncodingMapper * getInstance() {
+        static  EncodingMapper  _pSelf;
+        return &_pSelf;
+    };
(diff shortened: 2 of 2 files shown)


sasumner:
What issue number does this PR attempt to fix??

wjx0912:
hi, @sasumner 
no issue, just fix a memory leak,
best regards~

sasumner:
The site protocol is to open an issue, THEN create a PR that references that issue.

wjx0912:
issue: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6031

wjx0912:
#6031

donho:
The singleton lasts the whole life of application so when Notepad++ exits, the OS (Windows) manages to deallocate the allocated memory by application.
OTOH, it's a good practice to instance the singleton only when we need, without allocate in heap zone. Furthermore, the code is more modern (C++11). Hence the acceptation.

sasumner:
>Hence the acceptance

Disappointed to see the acceptance without the analysis and discussion by the author in the "issue"...

SinghRajenM:
When you say singleton, copy and assignment should not be allowed, should it?

donho:
@sasumner 
> Disappointed to see the acceptance without the analysis and discussion by the author in the "issue"

It's indeed a "leak of memory" since in EncodingMapper, a `new` is done but no `delete` (not as the 3 other singletons, see https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3dbb2c4b8ef47fddfd4e78aed086b28108fd036d and https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3643a1875b588e730d132a3e401f099a83b866ef).

donho:
@SinghRajenM 
> When you say singleton, copy and assignment should not be allowed, should it?

You're right about it.
Your enhancement on 4 singletons is welcome.

sasumner:
@donho 

I just like to see PR makers do a nice job with issue creation, before fixing code.  :)

I guess if the absolutely aren't going to do this, and team members quickly see that they are pointing out a real code problem with the PR, then, meh, ..., it's OK.

But I don't think we like it when we see an issueless PR and then have to go guessing at what the heck they are trying to accomplish with it (this is a generalization, maybe doesn't apply this time).

PR #6027, opened 2019-08-08
Fix "Command Argument Help" MessageBox

### LINKED ISSUE #4067: MsgBox "Command Line Arguments" not on top
### Description of the Issue
click Menu / ? / Command Line Arguments --> MsgBox disapears if Mouse is moved away from this (small) MsgBox-window,

1. click Menu / ? / Command Line Arguments
2. move mouse over underlying (but usually larger) Full-NPP-window (it seems that perhaps "only" the Editor-field (where the file content is displayed) is catching this MouseOver-event, the Menu or Window-title seems to have no effect here).

### Expected Behavior
for "Debug Info" MsgBox it works as expected (not going behind NPP-MainWindow).

### Actual Behavior
Cmd Line Arg - MsgBos goes away (behind MainWindow).
Additionally it can be opened multiple times, which makes no sense at all. Again Debug-Info cannot be open multiple times, 

### Debug Information
Notepad++ v7.5.4   (32-bit)
Build time : Jan  1 2018 - 01:47:45
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

xylographe:
Fix #4067

Prevent problems reported in [issuecomment-519245178](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4067#issuecomment-519245178).
Makes "Command Argument Help" mbox behave the same as "Summary" mbox.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2855,7 +2855,7 @@ void Notepad_plus::command(int id)
 		case IDM_CMDLINEARGUMENTS:
 		{
 			// No translattable
-			::MessageBox(NULL, COMMAND_ARG_HELP, TEXT("Notepad++ Command Argument Help"), MB_OK);
+			::MessageBox(_pPublicInterface->getHSelf(), COMMAND_ARG_HELP, TEXT("Notepad++ Command Argument Help"), MB_OK | MB_APPLMODAL);
 			break;
 		}
 


PowerEditor/src/NppCommands.cpp
```diff
@@ -2855,7 +2855,7 @@ void Notepad_plus::command(int id)
 		case IDM_CMDLINEARGUMENTS:
 		{
 			// No translattable
-			::MessageBox(NULL, COMMAND_ARG_HELP, TEXT("Notepad++ Command Argument Help"), MB_OK);
+			::MessageBox(_pPublicInterface->getHSelf(), COMMAND_ARG_HELP, TEXT("Notepad++ Command Argument Help"), MB_OK | MB_APPLMODAL);
```
donho: Personally, I don't like modal dialog.
It's not harmful if users want n windows of help.

sasumner: @donho The problem is it is easy to get the Help window open and BEHIND the main window (or other windows).  Then shortcuts in the main window DON'T work. This happened to me the other day and it took quite some time to understand what was happening.  Read https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4067#issuecomment-519245178 for details.  This PR is a good fix; I tested it. :)

xylographe: Did you read Scott's comment?

> In addition to the problems already noted, if the Notepad++ Command Argument Help window remains open and the user returns to editing within a Notepad++ editor tab view, odd things happen.
> 
> For an example of one, if the user presses ctrl+s to save the active document, it is not saved and a DC3 character is placed at the current position in the editor buffer. Another example is ctrl+w to close the active document; in this case an ETB character is inserted into the text instead. Closing out the open Notepad++ Command Argument Help window restores normal functionalities for these two examples.

sasumner: Finished testing the artifact build.  No problems found.  @donho Thank you for reconsidering. @xylographe Thanks for doing this code change.  :)

PR #6040, opened 2019-08-12
Bookmark all lines of multiline match (iss 6018)

### LINKED ISSUE #6018: 'Bookmark line' Marking doesn't bookmark all lines of a multi-line match
### Description of the Issue
When a `Mark`ing search with the `Bookmark line` option enabled finds a match that span lines, EVERY line that is part of the match should receive a bookmark; in reality only the FIRST line of the match gets bookmarked.

1. Paste this text into a new Notepad++ editing tab:

    Once upon a midnight dreary, while I pondered, weak and weary,
    Over many a quaint and curious volume of forgotten lore-
    While I nodded, nearly napping, suddenly there came a tapping,
    As of some one gently rapping, rapping at my chamber door.
    "'Tis some visitor," I muttered, "tapping at my chamber door-
    Only this and nothing more."

2. Invoke the Mark dialog via `Search` menu -> `Mark...`
3. Search for the `Find what` expression `(?s)lore.*?tor` as follows:

image

4. Press the `Mark All` button.

### Expected Behavior
Bookmarks should be placed as the following shows:

image

### Actual Behavior
A single bookmark is placed for this multiline match:

image

### Debug Information
Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\......\npp.7.7.1.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)

### Additional Information
This is likely the same issue as #3541.  However, since that issue is now closed and the example provided didn't reproduce, presumably because the poster didn't express it correctly rather than the issue not being present, this current issue was opened (with hopefully a better, more correct example).

The usefulness of this desired bookmarking behavior is mostly in the ability to `Copy Bookmarked Lines` and get ALL lines containing the match.  Copying the first line only of a multiline match is of dubious value.

If one is purely using `Next Bookmark` of `Previous Bookmark` to move around between matches, perhaps this proposed behavior change "gets in the way" by producing too many bookmarks.  For that case, one could replicate the first-line-only bookmarking behavior by a slight regex change in the `Find what`, e.g. `(?s)lore(?=.*?tor)` (note positive-lookahead), for the above example.  Or, even better, simply use `Search` -> `Jump Up` (or `Jump Down`) -> `Find Style` instead of going to the next/previous bookmark!

sasumner:
Fixes #6018 .

Note usage of `targetEnd - 1`.  The `- 1` part is necessary because if the expression to be marked ends with a line-ending, using `targetEnd` will result in the following line (not part of the match) also being bookmarked.  As a marking operation won't ever hit zero-length matches, `targetEnd - 1` is always >= `targetStart`.

Suggested `change.log` entry, if merged:  `Mark's Bookmark Lines feature now marks all lines of a multi-line match, instead of bookmarking only the first matching line.`

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2134,10 +2134,15 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 				if (_env->_doMarkLine)
 				{
 					auto lineNumber = pEditView->execute(SCI_LINEFROMPOSITION, targetStart);
-					auto state = pEditView->execute(SCI_MARKERGET, lineNumber);
+					auto lineNumberEnd = pEditView->execute(SCI_LINEFROMPOSITION, targetEnd - 1);
 
-					if (!(state & (1 << MARK_BOOKMARK)))
-						pEditView->execute(SCI_MARKERADD, lineNumber, MARK_BOOKMARK);
+					for (auto i = lineNumber; i <= lineNumberEnd; ++i)
+					{
+						auto state = pEditView->execute(SCI_MARKERGET, i);
+
+						if (!(state & (1 << MARK_BOOKMARK)))
+							pEditView->execute(SCI_MARKERADD, i, MARK_BOOKMARK);
+					}
 				}
 				break; 
 			}


xylographe:
@sasumner 
Perhaps, in the interest of legibility and maintainability (Contributing guide: “Always prefer a variable name that describes what the variable is used for.”), use slightly more “explanatory” variable names, like `firstLineNumber`, `lastLineNumber`, `(lineNumber = firstLineNumber; lineNumber <= lastLineNumber`?

sasumner:
@xylographe In general I agree with what the contributing guide says in that regard (who wouldn't?, it's common sense).  I think that statement in the guide is meant more for entirely new variables and/or new sections of code, though, but I could fully be wrong.

My attempt in this case was to not change the name of something existing that I was continuing to use.  In theory it makes less lines of difference (but in practice here that is only 1 line) for reviewers to examine.  Kind of following the idea that "more lines of diff == more likely to be rejected".

I'm inclined to let it stand for the moment at least.  I don't even know if @donho likes the functionality change.  If Don likes the change but agrees some of the variable names could be better,  I will certainly make those changes.

Issue #6047, opened 2019-08-14
Build failure: error C2672 in ThreadSafeQueue.h


nkolotov:
Methods of [dependent base class template `std::list<C>`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ab4cf883367eb3818fc0e6dea57d39d21f9ec618/PowerEditor/src/WinControls/ReadDirectoryChanges/ThreadSafeQueue.h#L32)  (such as `push_back` `empty` etc) are used without qualifier  causing standard conformant compilers to reject the code.

donho:
What "Build failure" ? 
Could you clarify the context ?

nkolotov:
So I have hard time trying to build notepad++. All the compilers that I've tried (vc++14.16, vc++14.22, clang5, clang7) will reject this code unless I switch to permissive mode. See [example](https://godbolt.org/z/It6GfJ).

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d38559b339a833c90f0df484297ddc249d6c43df

PR #6048, opened 2019-08-14
Fixed build failure caused by use of members of dependent base class template without qualifier.

nkolotov:
fix for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6047


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d38559b339a833c90f0df484297ddc249d6c43df

Issue #6049, opened 2019-08-14
Use of deprecated function std::bind2nd in WindowsDlg.cpp


nkolotov:
Function `std::bind2nd` has been deprecated in C++11 and then removed in C++17. [Use of this function  in WindowsDlg.cpp](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/06db9ee338ac9659b01b960325486a1791b249b0/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp#L783) is causing standard conformant compilers either to emit a warning or to reject the code.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e74cb34e22323b255ee9fc7603eb82f90cd60d28

PR #6050, opened 2019-08-14
Fixed use of deprecated function std::bind2nd.

nkolotov:
Fix for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6049
I took a straightforward approach and replaced `std::bind2nd` with `std::bind` though it would probably be better idea to use lambda or dedicated functor here.


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e74cb34e22323b255ee9fc7603eb82f90cd60d28

Issue #6053, opened 2019-08-14
Build failure: error C2362 in PluginsManager.cpp


nkolotov:
[Jumps to `cleanup` label](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/06db9ee338ac9659b01b960325486a1791b249b0/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp#L87) bypass `peHdr` variable initialization. Even though `peHdr` is not used anyhow after that label [most of the modern compilers flag this as fatal error](https://godbolt.org/z/-RpMVW).

Issue #6051, opened 2019-08-14
Build failure: error C3861 in Sorters.h


nkolotov:
[Function `std::make_pair` is used without qualificator](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/8b151ac991478a90d5064966f710ed566982c47e/PowerEditor/src/MISC/Common/Sorters.h#L292) and can not be found with argument-dependent lookup causing standard conformant compilers to reject the code.

donho:
Please clarify your "Build failure" context.

nkolotov:
So I have hard time trying to build notepad++. All the compilers that I've tried (vc++14.16, vc++14.22, clang5, clang7) will reject this code unless I switch to permissive mode. See [example](https://godbolt.org/z/g7ZxhS).

donho:
The project file of Notepad++ is VC 2017 (Mine is 15.8.9) and the warning level is set the most strict & treat warning as error. I have no problem of compiling.
What do you mean by "permissive mode" ?

nkolotov:
I've already answered the same question at [#6054
](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6054#issuecomment-521965028), but here is recap:

[...]

Issue #6043, opened 2019-08-13
Incomplete use of "Invalid regular expression" error message


sasumner:
### Description of the Issue

The error message `Find: Invalid regular expression` is used to alert the user to a syntax problem with their regular expression ONLY when a `Find Next` action is invoked.  An invalid-regular-expression message needs to be shown on the Find window's status bar whenever ANY type of find/replace/count/mark operation is initiated and the user's regular expression is malformed.  (Obviously this applies only when the user sets the `Search mode` to `Regular expression`.)

### Steps to Reproduce the Issue

1. Invoke the Find window (e.g. ctrl+f)
2. Put `abc[def` in the `Find what` box.  (This is a bad regex because the `[` pattern metacharacter has no corresponding `]`).  
3. Select `Regular expression` for the `Search mode`.
4. Press the `Find Next` button.  Observe the Find window's status bar indicates in red: `Find: Invalid regular expression`.  (This is correct and desired behavior and is only shown here to contrast with the following incorrect behavior.)
5. Press the `Count` button.  Observe the Find window's status bar indicates in blue: `Count: 0 matches`.  The status bar at this point should indicate that the regular expression is invalid.
6. Try other search actions (besides `Find Next`) that support regular expressions without changing the `Find what` box data.  Observe that none report on the invalidity of the regular expression, but rather indicate zero matches.  [Indeed, a find-in-files operation in a big folder tree will work for a long time to report zero matches, when it should abort very quickly at the start to indicate that the regex is bad.]

### Expected Behavior
I expect to be told that my regular expression is not correct, and other search activities to be aborted.  Zero matches is misleading when the true problem is that a search parameter is incorrect.

### Actual Behavior
See 5 and 6 above in Steps to Repro...

### Debug Information
Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\............\npp.7.7.1.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : ColumnTools.dll Explorer.dll LocationNavigate.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll PythonScript.dll

sasumner:
[...]

PR #6054, opened 2019-08-14
Fixed build failure caused by jump bypassing variable initialization.

nkolotov:
Fix for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6053
For now I've just reduced variable scope so it will never be left in uninitialized state, however it would probably be a better idea to get rid of goto horrors completely.

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -94,12 +94,13 @@ static WORD GetBinaryArchitectureType(const TCHAR *filePath)
 	if (addrHeader == NULL)
 		goto cleanup; // couldn't memory map the file
 
-	PIMAGE_NT_HEADERS peHdr = ImageNtHeader(addrHeader);
-	if (peHdr == NULL)
-		goto cleanup; // couldn't read the header
-
-	// Found the binary and architecture type
-	machine_type = peHdr->FileHeader.Machine;
+	{
+		PIMAGE_NT_HEADERS peHdr = ImageNtHeader(addrHeader);
+		if (peHdr == NULL)
+			goto cleanup; // couldn't read the header
+		// Found the binary and architecture type
+		machine_type = peHdr->FileHeader.Machine;
+	}
 
 cleanup: // release all of our handles
 	if (addrHeader != NULL)


donho:
Please clarify your "Build failure" context.

nkolotov:
So I have hard time trying to build notepad++. All the compilers that I've tried (vc++14.16, vc++14.22, clang5, clang7) will reject this code unless I switch to permissive mode. See [example](https://godbolt.org/z/It6GfJ).

donho:
The project file of Notepad++ is VC 2017 (Mine is 15.8.9) and the warning level is set the most strict & treat warning as error. I have no problem of compiling.
What do you mean by "permissive mode" ?

nkolotov:
The project from the repo uses VS2015 (XP) toolset (vc++14.0 compiler). Perhaps project indeed builds fine with that compiler, but I haven't tried it. https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6047, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6049, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6051, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6043 issues that I've reported describe problems when building using "standard conformant compilers". And by "standard conformant compilers" I mean compilers that claim to follow ISO C++ standard such as gcc and clang. Visual C++ has been notorious for doing things its own way, but recently MS [made a considerable efforts](https://devblogs.microsoft.com/cppblog/announcing-msvc-conforms-to-the-c-standard/) to [improve standard conformance](https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements?view=vs-2019). As a result newer vc++ compilers have conformance mode that can be enabled with [`/permissive-` switch](https://docs.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=vs-2019) (btw it is ON by default for new projects). While permissive mode still allows them to build most of invalid c++ code that was accepted by older vc++ compilers.

So in order to reproduce described build failures you need to (1) switch to newer toolset and corresponding Windows SDK, at least to 14.16 (2) disable permissive mode by toggling Project properties → C/C++ → Language → Conformance mode (3) optionally setting target c++ standard on the same property page to c++14 or even to c++17. If you go further and enable [built-in static analysis](https://docs.microsoft.com/en-us/visualstudio/code-quality/code-analysis-for-c-cpp-overview?view=vs-2019) there additionally will be ~400 warnings reported. It may also be beneficial to set up clang build as well because clang offers better conformance and issues clearer diagnostics.

Btw why does the project use VS2015 (XP) toolset instead of VS2017 (XP) toolset even though you use VS2017? Even if there is a particular reason to still use legacy VS2015 tools making code standard conformant will keep it future proof and will certainly make life of the developers using up-to-date tools easier.

PR #6055, opened 2019-08-15
Performance improvement part 1

SinghRajenM:
1. When an object of a class is created, the constructors of all member variables are called consecutively in the order the variables are declared, even if you don't explicitly write them to the initialization list. 
In order to follow above statement, avoid assigning a in ctor. Use the beauty of constructor initialization list to assign the value.

2. Many member variable were not assigned in the ctor (by normal human tendency), so use in class initialization to avoid such issues.

3. Pass variables are `const ref` instead of pass by value.

4. Treeview.h and treeview.cpp implemented under path ` PowerEditor/src/WinControls/TreeView/` were not used in the project. Treeview is being used from `PowerEditor/src/WinControls/ProjectPanel/`. So deleted it.


donho:
> When an object of a class is created, the constructors of all member variables are called consecutively in the order the variables are declared, even if you don't explicitly write them to the initialization list.
In order to follow above statement, avoid assigning a in ctor. Use the beauty of constructor initialization list to assign the value.

What's the concerning modification ?

> Many member variable were not assigned in the ctor (by normal human tendency), so use in class initialization to avoid such issues...

Why not?
But the whole changes makes code review difficult and it'll be "all or nothing".
Please create new PRs for each topic.

PR #6056, opened 2019-08-15
Performance improvement part 2

SinghRajenM:
1. `Common.cpp` don't return reference of local variable. Though, there implicit conversion exist (from vector to string). But let's follow KISS principle.
2. `Minidumer.cpp` reduce the scope of local variables and keep them initialized during declaration.
3. `pluginmanager.cpp` catch exception by reference.
4. In rest all other four files, member variables are initialized (either using initialization list or in class initialization).


donho:
Please create new PRs for each topic.

PR #6034, opened 2019-08-10
Handle Ctrl+A keyboard shortcut in MD5/SHA256 Generator

xylographe:
Added a new window procedure for the HASH_TEXT_EDIT control,
where Ctrl+A is now processed.

Fix #3863, close #3898, close #6034

@donho 
This is the rebased (and slightly adjusted) version of PR #3898.

According to [Microsoft ‘WM_CHAR message’ documentation](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-char) `wParam` contains “the character code of the key.” For Ctr+A that would be 0x0001 (UTF-16).

PR #6059, opened 2019-08-15
Handle Ctrl+A keyboard shortcut in MD5/SHA256 Generator

xylographe:
Added a new window procedure for the HASH_TEXT_EDIT control,
where Ctrl+A is now processed.

@donho
This PR is the successor of PR #6034. Apparently, the default Win7 EDITBOX proc informs windows it wants notification (WM_CHAR message) for all keys, and the default Win10 EDITBOX proc does *not*. Hence, by processing the WM_GETDLGCODE message, we can inform Win10 to generate the missing WM_CHAR messages.

@sasumner 
Thank you for testing under Windows 10.

### DIFF
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -34,8 +34,18 @@ INT_PTR CALLBACK HashFromFilesDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 			HFONT hFont = ::CreateFontA(fontDpiDynamicalHeight, 0, 0, 0, 0, FALSE, FALSE, FALSE, ANSI_CHARSET,
 				OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS, DEFAULT_QUALITY,
 				DEFAULT_PITCH | FF_DONTCARE, "Courier New");
-			::SendMessage(::GetDlgItem(_hSelf, IDC_HASH_PATH_EDIT), WM_SETFONT, reinterpret_cast<WPARAM>(hFont), TRUE);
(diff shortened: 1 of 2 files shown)


donho:
Works only on top edit control, not on the bottom part.

xylographe:
@donho 
Yes, it didn't, because the bottom part (hash result) is read-only.
With the new commit Ctrl+A now works in both edit controls.

donho:
@xylographe  It's better now, but it's still not perfect.
It does work for both text fields in dialog which launched by `Generate...`.
However, it doesn't work for both text fields in dialog which launched by `Generate from files...`

sasumner:
It is somewhat pointless to have ctrl+a work in the bottom boxes of both `Generate...` and `Generate from files...`.  Why?:  Because there is a `Copy to Clipboard` button there below the box which copies the entire contents of the lower box (even if there is a partial selection of its text); thus no need for ctrl+a.

Note:  Partial selection copy (if desired) can be achieved by making the selection, then using right-click-Copy.

But..for consistency in the UI, user expects ctrl+a to select-all, wherever they want to use it.  So ctrl+a should work in every text selection situation.

donho:
Yes, for the sake of consistency of UI, it’s good to have the same behavior. And it’s shame to not add this feature in them for 2 lines of code more.

xylographe:
@donho 
With this new commit Ctrl+A works in **HashFromFilesDlg** (HASH_PATH_EDIT & HASH_RESULT_EDIT controls), and in **HashFromTextDlg** (HASH_TEXT_EDIT & HASH_RESULT_FOMTEXT_EDIT controls).

donho:
`LRESULT HashFromFilesDlg::run_textEditProc(WNDPROC oldEditProc, HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)`
and `LRESULT HashFromTextDlg::run_textEditProc(WNDPROC oldEditProc, HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)` have the same code, so it's not necessary to duplicate the code. 
I'll do this part of enhancement.

PR #6062, opened 2019-08-16
Code enhancement: remove goto label

### LINKED ISSUE #6053: Build failure: error C2362 in PluginsManager.cpp

donho:
Close #6053, close #6054

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -76,40 +76,38 @@ bool PluginsManager::unloadPlugin(int index, HWND nppHandle)
     return true;
 }
 
-static WORD GetBinaryArchitectureType(const TCHAR *filePath)
+static WORD getBinaryArchitectureType(const TCHAR *filePath)
 {
-	WORD machine_type = IMAGE_FILE_MACHINE_UNKNOWN;
-	HANDLE hMapping = NULL;
-	LPVOID addrHeader = NULL;
-
 	HANDLE hFile = CreateFile(filePath, GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_READONLY, NULL);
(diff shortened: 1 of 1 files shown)


PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
```diff
@@ -76,40 +76,38 @@ bool PluginsManager::unloadPlugin(int index, HWND nppHandle)
...
+static WORD getBinaryArchitectureType(const TCHAR *filePath)
 {
-	WORD machine_type = IMAGE_FILE_MACHINE_UNKNOWN;
-	HANDLE hMapping = NULL;
-	LPVOID addrHeader = NULL;
-
 	HANDLE hFile = CreateFile(filePath, GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_READONLY, NULL);
 	if (hFile == INVALID_HANDLE_VALUE)
-		goto cleanup;
+	{
```
SinghRajenM: @donho instead of multiple return statement why not single return. e.g. put everything in try and through the result, catch the result and return it. This way multiple `CloseHandle` can be avoided. Just a thought.

donho: It's another way to deal with error.
Since it's a small function, I'll keep this way. Thank you for the suggestion.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/77da7062077f03d75f1e4c27648596e186f894b6

PR #6064, opened 2019-08-16
Don't return reference of local varibales

SinghRajenM:
Don't return reference of local variable. Though, there implicit conversion exist (from vector to string). But let's follow KISS principle and keep it neat and clean.


donho:
The current code works and I don't see the advantage of your modification.

SinghRajenM:
> 
> 
> The current code works and I don't see the advantage of your modification.

@donho Cool, but error is an error.....
image

donho:
@SinghRajenM How can I produce it ?

SinghRajenM:
@donho you need to use some code analyzer. I used cppCheck.

PR #6065, opened 2019-08-16
Make ctor/dtor explicit default if possible

SinghRajenM:
Make ctor/dtor explicit default if possible otherwise make use of initialization list.


donho:

PR #6066, opened 2019-08-16
Make progress class more elegant

SinghRajenM:
Make progress class more elegant


donho:

PR #6067, opened 2019-08-16
Provide default value for class members

SinghRajenM:
Provide default value for class members (either using in class initialization or initializer list in ctor)


donho:

Issue #6069, opened 2019-08-17
Incorrect font used to draw captions of document tabs when non-reduced.


nkolotov:
When "Reduced" option for tab bar is disabled in settings "system" font will be used for tab captions. This font lacks cleartype antialiasing, moreover, it has reduced set of symbols so most of Unicode file names will be displayed using symbols from some other font (that may or may not have cleartype antialiasing) causing captions to be rather messed up.
![tab bar problem](https://user-images.githubusercontent.com/50573584/63210939-64150680-c0f9-11e9-8c5c-798719c49c94.png)

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6070#issuecomment-524455571

PR #6070, opened 2019-08-17
Switched to use larger font based on DEFAULT_GUI_FONT rather than SYSTEM (that lacks antialiasing and most of unicode symbols) for drawing non-reduced tab bar captions.

nkolotov:
Fix for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6069

tab bar fixed

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -343,18 +343,24 @@ void TabBarPlus::init(HINSTANCE hInst, HWND parent, bool isVertical, bool isMult
 
 	LOGFONT LogFont;
 
-	_hFont = (HFONT)::SendMessage(_hSelf, WM_GETFONT, 0, 0);
+	_hFont = reinterpret_cast<HFONT>(::SendMessage(_hSelf, WM_GETFONT, 0, 0));
 
 	if (_hFont == NULL)
-		_hFont = (HFONT)::GetStockObject(DEFAULT_GUI_FONT);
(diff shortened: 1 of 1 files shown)


donho:
`before` display the unicode symbole correctly or am I missing somethings ?
I've tried to use the Chinese  transitional file name under my English Windows 10 then open it, it displays no problem on both reduced and enlarged mode.

nkolotov:
Some symbols are using cleartype antialiasing while some dont't use any anitialiasing. Those without antialiasing are from system font, while the rest with cleartype antialiasing seem to be upscaled symbols from default gui font. So literally all of them are rendered incorrectly.

donho:
Here is the screenshot after applying this PR:
image

On my PC even reduce to `LONG const large_height = (normal_height * 120) / 100;`, a very long file name still can't fit.
`LONG const large_height = normal_height;` can make any size of file name fit, but in this case, non-reduced mode loss its meaning. 

The antialiasing issue is a cosmetic and small issue, so I prefer keep it as it is now.

nkolotov:
Yeah, the text placement is definitely off, however it is mostly the problem with tab size / positioning calculation. (Actually I was going to add some follow up issues on this matter). It seems to be somewhat incorrect even on regular DPI but gets even worth with high DPI. And it is not just text problem, for example "Saved" icon on tabs is placed incorrectly as well, it clearly goes beyond and blow left tab corner. 

> The antialiasing issue is a cosmetic and small issue
For me and some other people non-reduced tab bar is unusable in its current state.

PR #6072, opened 2019-08-18
Update JSON for Modern C++

CarysT:
Does not present any errors or warnings in notepad++, but I assume it might break compatibility with some plugins. The newer versions of the library have some nice features.

Updated from 2.1.1 to 3.7.0


donho:
I don't see any need so far.
If there's in the future, I'll take care of it myself.

PR #6075, opened 2019-08-19
fix so that Find-All-in-All-Opened-Docs doesn't search cloned doc 2 times

### LINKED ISSUE #6074: Find-All-in-All-Opened-Documents searches cloned document two times
### Description of the Issue
If I have a cloned document open in both views, a `Find All in All Opened Documents` search that has hits in the document will report each and every hit two times in the `Find result` window.

1. Open the `change.log` file.
2. Right-click its tab and choose `Clone to Other View`.
3. Invoke `Find` (ctrl+f).
4. Put `enhancements` into `Find what`, all other options unticked.
5. Press `Find All in All Opened Documents` button.

### Expected Behavior
I expect, since it is the same document in both views (because it is cloned), that I would only get unique hits in `Find result`, i.e., a second, unnecessary search of the file would not be conducted.  For the example given I would expect to see `1 hit in 1 file`.

### Actual Behavior
The same hits are reported twice in `Find result`, see screenshot below; for the example provided I see `2 hits in 2 files`.

### Debug Information
Notepad++ v7.7.1   (64-bit)
Build time : Jun 16 2019 - 21:24:47
Path : C:\........\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)

image

sasumner:
Fixes #6074 .

If adopted the suggested `change.log` entry is:
"Fix for Find-All-in-All-Opened-Documents searching a cloned document two times, and reporting same results two times."

----------------------

Note: I don't know why the github diff is showing my `continue` line as indented too many times. Perhaps it has to do with the intermixing of tabs and spaces in the code before I changed it?  

Here's what my change looks like in VS (boxed in red), note tab-indenting and space-indenting intermingled at top and bottom of screenshot (but **I** didn't do that):

image

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1802,6 +1802,10 @@ bool Notepad_plus::findInOpenedFiles()
 		for (size_t i = 0, len2 = _subDocTab.nbItem(); i < len2 ; ++i)
 	    {
 			pBuf = MainFileManager.getBufferByID(_subDocTab.getBufferByIndex(i));
+			if (_mainDocTab.getIndexByBuffer(pBuf) != -1)
+			{
+				continue;  // clone was already searched in main; skip re-searching in sub
+			}
 			_invisibleEditView.execute(SCI_SETDOCPOINTER, 0, pBuf->getDocument());
 			auto cp = _invisibleEditView.execute(SCI_GETCODEPAGE);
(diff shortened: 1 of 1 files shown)


xylographe:
> I don't know why the github diff is showing my continue line as indented too many times.

GitHub tabulator width is 8 columns, not 4.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e102ecb919fd3f5ed524d136c104f9208cfedec6

Issue #6020, opened 2019-08-08
Feature Request: Increase the number of keywords List/set


oirfeodent:
### Description of the Issue

The list of keyword sets currently used by NPP seems to be 7.
There are some cases where the keyword list used by some languages goes beyond 7.
It could be useful to increase the keyword list to possibly 9/10.

### Steps to Reproduce the Issue

Not sure if the code location can be mentioned here... but will give a try.
If there is need to amend this, will do later.
1. Currently [**notepad-plus-plus\PowerEditor\src\ScitillaComponent\ScintillaEditView::setLexer()**](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/92bad0a60ad606b30df9ed97aecf4ff27bb6e967/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp#L1179) is used to set the keyword list for any language.
2. Though the array [**(*pKwArray[10])**](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/92bad0a60ad606b30df9ed97aecf4ff27bb6e967/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp#L1183) is declared as 10, only 7 is defined to be used with [**setLexer()**](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/92bad0a60ad606b30df9ed97aecf4ff27bb6e967/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp#L1179)
3. Due to this the languages which have more keywords beyond 7 cannot use those keywords list.

### Expected Behavior

If possible to increase the keyword list/set increase the same to the 10 available.

### Actual Behavior

Cannot use more than 7 keyword list/set.

### Debug Information

Notepad++ v7.7.1   (64-bit)
Build time : Jun 16 2019 - 21:24:47
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 8.1 (64-bit)
[...]

PR #6081, opened 2019-08-21
Increase Keyword Set. The set is declared as 10 but not coded.

oirfeodent:
This will enable languages having more than 7 keyword sets to use it.
Will close issue #6020 .

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -550,6 +550,8 @@ int getKwClassFromName(const TCHAR *str)
 	if (!lstrcmp(TEXT("type3"), str)) return LANG_INDEX_TYPE3;
 	if (!lstrcmp(TEXT("type4"), str)) return LANG_INDEX_TYPE4;
 	if (!lstrcmp(TEXT("type5"), str)) return LANG_INDEX_TYPE5;
+	if (!lstrcmp(TEXT("type6"), str)) return LANG_INDEX_TYPE6;
+	if (!lstrcmp(TEXT("type7"), str)) return LANG_INDEX_TYPE7;
 
 	if ((str[1] == '\0') && (str[0] >= '0') && (str[0] <= '8')) // up to KEYWORDSET_MAX
 		return str[0] - '0';
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -1226,6 +1226,19 @@ void ScintillaEditView::setLexer(int lexerID, LangType langType, int whichList)
 		const char * keyWords_char = wmc.wchar2char(pKwArray[LANG_INDEX_TYPE5], CP_ACP);
 		setKeywords(langType, keyWords_char, LANG_INDEX_TYPE5);
 	}
+
+	if (whichList & LIST_7)
+	{
+		const char * keyWords_char = wmc.wchar2char(pKwArray[LANG_INDEX_TYPE6], CP_ACP);
+		setKeywords(langType, keyWords_char, LANG_INDEX_TYPE6);
+	}
+
+	if (whichList & LIST_8)
+	{
+		const char * keyWords_char = wmc.wchar2char(pKwArray[LANG_INDEX_TYPE7], CP_ACP);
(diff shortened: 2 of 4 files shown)


PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -1226,6 +1226,24 @@ void ScintillaEditView::setLexer(int lexerID, LangType langType, int whichList)
...
+		setKeywords(langType, keyWords_char, LANG_INDEX_TYPE6);
+	}
+
+	if (whichList & LIST_8)
+	{
+		const char * keyWords_char = wmc.wchar2char(pKwArray[LANG_INDEX_TYPE7], CP_ACP);
+		setKeywords(langType, keyWords_char, LANG_INDEX_TYPE7);
+	}
+
+	if (whichList & LIST_9)
```
donho: `You can set up to 9 lists of keywords for use by the current lexer. keyWordSet can be 0 to 8:` 
https://www.scintilla.org/ScintillaDoc.html#SCI_SETKEYWORDS

PowerEditor/src/ScitillaComponent/ScintillaEditView.h
```diff
@@ -95,6 +95,9 @@ const int CP_GREEK = 1253;
 #define LIST_4 16
 #define LIST_5 32
 #define LIST_6 64
+#define LIST_7 128
+#define LIST_8 256
+#define LIST_9 512
```
donho: `You can set up to 9 lists of keywords for use by the current lexer. keyWordSet can be 0 to 8:` 
https://www.scintilla.org/ScintillaDoc.html#SCI_SETKEYWORDS

donho: So LIST_9 is useless.

PowerEditor/src/Parameters.cpp
```diff
@@ -550,6 +550,9 @@ int getKwClassFromName(const TCHAR *str)
 	if (!lstrcmp(TEXT("type3"), str)) return LANG_INDEX_TYPE3;
 	if (!lstrcmp(TEXT("type4"), str)) return LANG_INDEX_TYPE4;
 	if (!lstrcmp(TEXT("type5"), str)) return LANG_INDEX_TYPE5;
+	if (!lstrcmp(TEXT("type6"), str)) return LANG_INDEX_TYPE6;
+	if (!lstrcmp(TEXT("type7"), str)) return LANG_INDEX_TYPE7;
+	if (!lstrcmp(TEXT("type8"), str)) return LANG_INDEX_TYPE8;
 
 	if ((str[1] == '\0') && (str[0] >= '0') && (str[0] <= '8')) // up to KEYWORDSET_MAX
```
xylographe: Perhaps, `(str[0] <= '9')`?

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
xylographe: "_keyWordSet can be 0 to 8 (actually 0 to **KEYWORDSET_MAX**)_"

In Scintilla.h: `#define KEYWORDSET_MAX 30`.

Should be OK, depending on lexer (BaanC?) ?

oirfeodent: @donho I have removed LIST_9.
However, I would like to highlight that [KEYWORDSET_MAX](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/92bad0a60ad606b30df9ed97aecf4ff27bb6e967/scintilla/include/Scintilla.h#L1003) is currently 30... but, BAANC is using the Maximum Keyword sets, which will be fulfilled with LIST_8.

PowerEditor/src/ScitillaComponent/ScintillaEditView.h
oirfeodent: @donho LIST_9 is removed.

PowerEditor/src/Parameters.cpp
oirfeodent: @xylographe LIST_9 is removed in the recent amend, so should be ok.

Issue #6101, opened 2019-08-28
Unoptimal performance of string ↔ wstring conversion.


nkolotov:
[`string2wstring`](https://github.com/SinghRajenM/notepad-plus-plus/blob/fb78d366d6b60827ef7c3af7374744084ad91420/PowerEditor/src/MISC/Common/Common.cpp#L545) and [`wstring2string`](https://github.com/SinghRajenM/notepad-plus-plus/blob/fb78d366d6b60827ef7c3af7374744084ad91420/PowerEditor/src/MISC/Common/Common.cpp#L559) functions allocate unnecessary temporary buffer vector and then copy its content into output string. Length of the input string will be recalculated twice by `MultiByteToWideChar` even though it is known at advance, length of the output string will be calculated once (during construction of the string returned) even though it is also known. Moreover, most of `wstring2string` invocations involve construction of unnecessary temporary string objects ( [example](https://github.com/SinghRajenM/notepad-plus-plus/blob/fb78d366d6b60827ef7c3af7374744084ad91420/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp#L683) ) wasting even more time on memory reallocation.

chcg:
@nkolotov Did you do any measurements on this changes? Are the performance gains significant?

nkolotov:
@chcg No, but the functions changed don't seem to be used on hot paths anyway.

donho:
We need the measurement to prove there's need to take the regression risk to change. Otherwise, the current code works fine without any performance issue. I don't consider it as bug.

PR #6103, opened 2019-08-28
Improved performance of string ↔ wstring conversion.

nkolotov:
Fix for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6101


donho:
In case that it can be measured that your conversion is much performance, let's remain the readable code for the maintenance and avoid the regression.

Issue #6105, opened 2019-08-29
Left margin width incorrect


koernertim:
### Description of the Issue
When you open an blank Notepat++ file, the width of the line number margin is configured to display up to 3-digit numbers (so up to line 999). In conclusion the margin shoud get wider when you reach line 1000. However, this is not the case. The margin already gets wider once you reach line 999. The same thing applies to all other potencies of 10. The margin gets wider on the line before you reach 1000/10000/etc...

### Steps to Reproduce the Issue

1.  Open an blank Notepad++ file
2.  Paste 999 new lines
3.  Look at the number line margin width

### Expected Behavior
Margin gets wider once you reach line 1000

### Actual Behavior
Margin gets wider once you reach line 999

![image](https://user-images.githubusercontent.com/48256264/63914167-a6601100-ca32-11e9-9d12-fc82e934badf.png)

![image](https://user-images.githubusercontent.com/48256264/63914188-b37d0000-ca32-11e9-8b96-177949d9059e.png)

sasumner:
Closing due to trivial nature of the complaint.  (Sorry)

PR #6107, opened 2019-08-29
Fix issue #6105

koernertim:
Fix issue #6105 which has been closed due to trivial nature of the complaint. Pretty simple fix.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -1444,7 +1444,7 @@ void ScintillaEditView::defineDocType(LangType typeDoc)
     }
 	setSpecialIndicator(*pStyle);
 
-    // Il faut surtout faire un test ici avant d'exécuter SCI_SETCODEPAGE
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -1444,7 +1444,7 @@ void ScintillaEditView::defineDocType(LangType typeDoc)
     }
 	setSpecialIndicator(*pStyle);
 
-    // Il faut surtout faire un test ici avant d'ex�cuter SCI_SETCODEPAGE
```
koernertim: This happend automatically due to encoding issues

xylographe:
This changes the minimum from 3-digit numbers to 2-digit numbers.
Also, it will set space for 3-digit numbers if last line is for example 1002.

Perhaps, you meant to replace
```
auto lastVisibleLineDoc = execute(SCI_DOCLINEFROMVISIBLE, lastVisibleLineVis);
```
with
```
auto lastVisibleLineDoc = execute(SCI_DOCLINEFROMVISIBLE, lastVisibleLineVis) - 1;
```
This should be extensively tested, though, with and without line wrap.

donho:
It's not bug for me so nothing to fix.

PR #6115, opened 2019-09-01
Use C++11 range based for loop

SinghRajenM:
1. Use C++11 range based for loop
2. Avoid using `using namespace std;` as it is considered bad practices.


donho:
I'll take in charge the code enhancement myself.
Thank you anyway for the heads up.

PR #6116, opened 2019-09-01
Avoid using "using namespace std;", rather use "std::"

SinghRajenM:
Avoid using `using namespace std;` rather use `std::`. Using `using namespace std;` is considered bad practice.

Some links which advise not to use `using namespace std;`
https://isocpp.org/wiki/faq/coding-standards#using-namespace-std
https://akrzemi1.wordpress.com/2019/03/14/not-using-namespace-std/
https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -50,7 +50,6 @@
 #include "fileBrowser.h"
 #include "Common.h"
 
-using namespace std;
 
 enum tb_stat {tb_saved, tb_unsaved, tb_ro};
 #define DIR_LEFT true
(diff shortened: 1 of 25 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6541,9 +6540,9 @@ DWORD WINAPI Notepad_plus::threadTextPlayer(void *params)
 
 	//writeLog(TEXT("c:\\tmp\\log.txt"), "\n\n\n\n");
 	const wchar_t* quoter = qParams->_quoter;
-	wstring quoter_str = quoter;
+	std::wstring quoter_str = quoter;
 	size_t pos = quoter_str.find(TEXT("Anonymous"));
-	if (pos == string::npos)
+	if (pos == std::wstring::npos)
```
SinghRajenM: It should have been `wstring::npos` instead of `string::npos`

PowerEditor/src/Parameters.cpp
```diff
@@ -3348,25 +3346,25 @@ void NppParameters::feedUserKeywordList(TiXmlNode *node)
...
 				//int len = _tcslen(kwl);
-				basic_string<TCHAR> temp{TEXT(" ")};
+				generic_string temp{TEXT(" ")};
 
 				temp += kwl;
 				size_t pos = 0;
 
 				pos = temp.find(TEXT(" 0"));
-				while (pos != string::npos)
+				while (pos != generic_string::npos)
```
SinghRajenM: It should have been `wstring::npos` instead of `string::npos`.

donho:
Code enhancement is nice, if it improves the performance. Otherwise let's focus on new feature and fixing bugs.

PR #6117, opened 2019-09-02
Reduce useless member variable from class

SinghRajenM:
Code enhancement: Let's talk in term of performance improvement. Avoid unnecessary memory footprint by removing useless member of class.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/UserDefineDialog.h
+++ b/PowerEditor/src/ScitillaComponent/UserDefineDialog.h
@@ -56,7 +56,6 @@ class GlobalMappers
 		std::unordered_map<int, generic_string> styleNameMapper;
 
 		std::unordered_map<generic_string, int> temp;
-		std::unordered_map<generic_string, int>::iterator iter;
 
 		std::unordered_map<int, int> nestingMapper;
 		std::unordered_map<int, int> dialogMapper;
@@ -75,8 +74,8 @@ class GlobalMappers
             temp[TEXT("Words4")]                        = SCE_USER_KWLIST_KEYWORDS4;
 
             // in case of duplicate entries, newer string should overwrite old one !
-            for (iter = temp.begin(); iter != temp.end(); ++iter)
-                keywordNameMapper[iter->second] = iter->first;
+			for (auto const& iter : temp)
+				keywordNameMapper[iter.second] = iter.first;
             keywordIdMapper.insert(temp.begin(), temp.end());
             temp.clear();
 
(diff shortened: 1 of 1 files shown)


donho:
The procedure for accepting PR is: create an issue, where you describe the bug for reproducing it or an performance issue which can be measured. Once it is accepted, you can do your PR. 

SinghRajenM:
>an performance issue which can be measured.

@donho Even a novice can tell the what will be size of below two classes. IMHO, such a very small basic things do not need measurement or discussion. BTW, rule is rule.
``` C++
class OneVariable
{
        int m_nVal = 0;
}
```

``` C++
class TwoVariable
{
        int m_nVal = 0;
       std::unordered_map<generic_string, int>::iterator iter;
}
```
Anyway for your reference----

image

sasumner:
@SinghRajenM :

My 2c:

>a novice can tell the what will be size of below two classes

I don't consider myself a novice, but without doing the analysis you provided just above, I would not have known that `TwoVariable` is 32.  Perhaps the point was to just show that using `iter` at all takes up _some_ space.  But, yes, see:  In the discussion there is stuff to be learned!

>such a very small basic things do not need measurement or discussion

I think every proposed code change deserves these things.  And especially code that works as it exists.  Changing because there are newer ways to do things now that there were when the code was first written (that's just a guess for this situation) isn't in-and-of-itself a great reason to change it.  But finding a significant performance improvement (for example) in a new way of doing something _would_ be a great reason.

All code changes involve some amount of risk of unintended consequences.  An author/owner of a codebase assumes that risk, and thus must be convinced that changes are needed and worth taking even a small amount of risk on.

Perhaps the best argument for discussion:  Discussion is great to change the minds of those who would look at only a code change and say "meh" and dismiss it.  :)

That all being said, I have seen Don accept PRs of this type before, without discussion in an "issue".  I do not understand the distinction between why "those" would be accepted and "these" wouldn't.  I would _like_ to understand, because sometimes here I chide people that do a PR without first creating an issue, and then (I suppose) I look foolish for doing that when Don goes ahead and accepts such a PR.

If the contribution guidelines say that an issue must be created before a PR (and they do), then that's what all PRs should have.

>rule is rule

I'm not sure what this is saying.  If there's a C++ thing that says "this is how one shall do something", then perhaps citing that rule would be a good idea (in the "discussion").  But again, if the rule (or language construct) was not in existence at the time of original code-writing, it is not a definite that such a change should be made.

Again, just my 2c, I'm trying to _understand_, not to enflame.  I think you've made some great contributions to N++ and I hope to see many more in the future.

SinghRajenM:
> I don't consider myself a novice, but without doing the analysis you provided just above, I would not have known that TwoVariable is 32.

@sasumner greetings! it is not about `32`, you will get different result when you will run it on `x86` architecture. There are many more things involved when we talk about the size of class like "structure padding", hidden pointers e.g. __vptr etc.

**The point is**: when a member is added to class, definitely, class size (more precisely object size) will increase which is clearly known. How much size will be increased is other part of story. So why to add useless members to class.

PR #6118, opened 2019-09-02
Fixed code error (may be typo)

SinghRajenM:
Fixing some code error that seems to be typo.

- Notepad_plus.cpp: 
Performing find operation on `wstring` should compare with `wstring::npos` instead of `string::npos`. Mind `w' here.

- Parameters.cpp
`basic_string<TCHAR>` is typedef as `generic_string`. So replace `basic_string<TCHAR>` by `generic_string`. 
Also use `::npos` from `generic_string` instead from `string::npos`.


donho:
The procedure for accepting PR is: create an issue, where you describe the bug for reproducing it or an performance issue which can be measured. Once it is accepted, you can do your PR. 

SinghRajenM:
@donho @sasumner 
Few things are self-explanatory ---
image

When dealing with `wstring` why to think about `string` at all.

PR #6119, opened 2019-09-02
LexA68k.cxx minor comment fix

mag92:
"toggle" -> "toggled"


chcg:
@mag92 This part is taken over from https://www.scintilla.org/ and it doesn't make sense to fix it here as it will be vanished again with the next update to a newer scintilla version. You should consider to contribute directly at scintilla.

mag92:
Makes sense. Thanks for explaining.

PR #6164, opened 2019-09-20
Implements native Win32 API file access and uses it to write user files

### LINKED ISSUE #5664: File is erased on save when there is no free space left on disk
### Description of the Issue

2. Add several lines.
3. Save the file.

### Expected Behavior

### Actual Behavior

### Root cause
The problem is caused by
in `FileManager::saveBuffer`

### Possible solutions

### Notes on solutions

pnedev:
Fixes #5664 and hopefully #6133 (config and session files are excluded because TinyXML lib still uses CRT file APIs)

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -1,0 +1,153 @@
+// This file is part of Notepad++ project
+// Copyright (C)2019 Don HO <don.h@free.fr>
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either
+// version 2 of the License, or (at your option) any later version.
+//
+// Note that the GPL places important restrictions on "derived works", yet
+// it does not provide a detailed definition of that term.  To avoid
+// misunderstandings, we consider an application to constitute a
+// "derivative work" for the purpose of this license if it does any of the
+// following:
+// 1. Integrates source code from Notepad++.
+// 2. Integrates/includes/aggregates Notepad++ into a proprietary executable
+//    installer, such as those produced by InstallShield.
+// 3. Links to a library or executes a program that does any of the above.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, write to the Free Software
+// Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+
+#include "FileInterface.h"
+
+
+CFile::CFile(const char *fname, Mode fmode) : _hMode(fmode)
+{
+	if (fname)
+	{
+		DWORD access, share, disp, attrib;
+		fillCreateParams(access, share, disp, attrib);
+
+		_hFile = ::CreateFileA(fname, access, share, NULL, disp, attrib, NULL);
+	}
+
+	if ((_hFile != INVALID_HANDLE_VALUE) && (_hMode == Mode::APPEND))
+	{
+		LARGE_INTEGER offset;
+		offset.QuadPart = 0;
+
+		::SetFilePointerEx(_hFile, offset, NULL, FILE_END);
+	}
+}
+
+
+CFile::CFile(const wchar_t *fname, Mode fmode) : _hMode(fmode)
+{
+	if (fname)
+	{
+		DWORD access, share, disp, attrib;
+		fillCreateParams(access, share, disp, attrib);
+
+		_hFile = ::CreateFileW(fname, access, share, NULL, disp, attrib, NULL);
+	}
+
+	if ((_hFile != INVALID_HANDLE_VALUE) && (_hMode == Mode::APPEND))
+	{
+		LARGE_INTEGER offset;
+		offset.QuadPart = 0;
+
+		::SetFilePointerEx(_hFile, offset, NULL, FILE_END);
+	}
+}
+
+
+void CFile::Close()
+{
+	if (IsOpened())
+	{
+		if (_written)
+		{
+			::SetEndOfFile(_hFile);
+			::FlushFileBuffers(_hFile);
+		}
+
+		::CloseHandle(_hFile);
+
+		_hFile = INVALID_HANDLE_VALUE;
+	}
+}
+
+
+int_fast64_t CFile::GetSize()
+{
+	LARGE_INTEGER r;
+	r.QuadPart = -1;
+
+	if (IsOpened())
+		::GetFileSizeEx(_hFile, &r);
+
+	return static_cast<int_fast64_t>(r.QuadPart);
+}
+
+
+long CFile::Read(void *rbuf, long buf_size)
+{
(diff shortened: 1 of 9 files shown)


PowerEditor/src/MISC/Common/FileInterface.cpp
```diff
@@ -0,0 +1,167 @@
...
+	{
+		_hFile = INVALID_HANDLE_VALUE;
+	}
+
+	if ((_hFile != INVALID_HANDLE_VALUE) && (_hMode == APPEND))
+	{
+		LARGE_INTEGER offset;
+		offset.QuadPart = 0;
+
+		::SetFilePointerEx(_hFile, offset, NULL, FILE_END);
```
chcg: See https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfilepointerex

> You can use SetFilePointerEx to determine the length of a file. To do this, use FILE_END for dwMoveMethod and seek to location zero. The file offset returned is the length of the file. However, this practice can have unintended side effects, such as failure to save the current file pointer so that the program can return to that location. It is simpler and safer to use the GetFileSizeEx function instead.

chcg:
@pnedev How is this affecting portability of the code? I guess the mingw build should have no problem with this change.

See also https://stackoverflow.com/questions/38602652/how-to-detect-insufficient-disk-space-when-using-stdio-file-operations-in-c-c. There seems to be indeed no solution based on the stdio file operations :-(

PowerEditor/src/Utf8_16.cpp
```diff
@@ -285,52 +284,64 @@ Utf8_16_Write::~Utf8_16_Write()
 	fclose();
 }
 
-FILE * Utf8_16_Write::fopen(const TCHAR *_name, const TCHAR *_type)
+bool Utf8_16_Write::fopen(const TCHAR *name)
 {
-	m_pFile = ::generic_fopen(_name, _type);
+	m_pFile.reset(new CFile(name, CFile::WRITE));
```
SinghRajenM: `std::make_unique` is preferred over `new`

PowerEditor/src/MISC/Common/FileInterface.h
```diff
@@ -0,0 +1,77 @@
...
+
+
+private:
+	CFile() = delete;
+	CFile(const CFile&) = delete;
+	CFile& operator=(const CFile&) = delete;
+
+	void fillCreateParams(DWORD &access, DWORD &share, DWORD &disp, DWORD &attrib);
+
+	HANDLE	_hFile;
```
SinghRajenM: In class initialization should be used for `_hFile` and `_hMode` too.

PowerEditor/src/MISC/Common/FileInterface.h
```diff
@@ -0,0 +1,77 @@
...
+	void Close();
+
+	int_fast64_t GetSize();
+
+	long Read(void *rbuf, long buf_size);
+	bool Write(const void *wbuf, long buf_size);
+
+
+private:
+	CFile() = delete;
```
SinghRajenM: If parameterized ctor is defined, then default ctor is automatically overruled. So no need to mark it `deleted` explicitly,

PowerEditor/src/MISC/Common/FileInterface.cpp
```diff
@@ -0,0 +1,167 @@
...
+CFile::CFile(const char *fname, Mode fmode) : _hMode(fmode)
+{
+	if (fname)
+	{
+		DWORD access, share, disp, attrib;
+		fillCreateParams(access, share, disp, attrib);
+
+		_hFile = ::CreateFileA(fname, access, share, NULL, disp, attrib, NULL);
+	}
+	else
```
SinghRajenM: In class initialization will take care of `else` statement if defined. So `else` is not needed

pnedev:
> @pnedev How is this affecting portability of the code? I guess the mingw build should have no problem with this change.
> 

Well, It is the native Win32 API but Notepad++ extensively uses it. MinGW doesn't have problems with that. The `FileInterface` idea is to act like operating system abstraction layer. It can be rewritten in the future for other OS APIs if needed and the code that uses it will not change.

> See also https://stackoverflow.com/questions/38602652/how-to-detect-insufficient-disk-space-when-using-stdio-file-operations-in-c-c. There seems to be indeed no solution based on the stdio file operations :-(
> 
> https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/write?view=vs-2019
> has ENOSPC and might be another option.

The issue is not because of _write() failing but because of fopen() deleting the file contents on open.
But you are right - _sopen() and _write() could be used as well. I don't see why one should prefer _sopen() over CreateFile() though.

PowerEditor/src/MISC/Common/FileInterface.h
pnedev: I prefer to be explicit when I'm not sure about something :)
OK, thanks.

PowerEditor/src/Utf8_16.cpp
pnedev: Perhaps but there's nothing wrong with this code, correct me if I'm wrong.

LordNyriox:
@donho:  What is your opinion on this PR?

donho:
@pnedev Sorry I have to decline it - The change is so heavily that could bring some regression.

> But you are right - _sopen() and _write() could be used as well. I don't see why one should prefer _sopen() over CreateFile() though.

Yes, why not - from the viewpoint of a project maintainer, the best fix is with less lines - if it does fix the bugs.

pnedev:
@donho ,

I know that what I'm about to say won't make any difference but I'll say it anyway.

Some fixes require more than a few lines of change, you know that. There have been enough bigger changes recently that could have also brought regressions but they had been done nevertheless.
It seems that the issues this PR fixes are not quite important to you.

About the change:
It is not heavy at all. It simply replaces `fopen()` with `CreateFile` and `fread()` / `fwrite()` with `ReadFile()` / `WriteFile()`. It does it in a way that is convenient to be used anywhere in the code and be easily modified in the future if such need arises without introducing new big and heavy to review new changes. It makes the code more manageable in the future.

As a remark `_sopen()` and `_write()` would not make a smaller change.

P.S. Since I made the fix I have been using my patched notepad++.exe version daily without facing any problems at all.

xylographe:
> P.S. Since I made the fix I have been using my patched notepad++.exe version daily without facing any problems at all.

Same experience here (using this patch since 2019-09-23).

chcg:
@pnedev @xylographe @donho 
Maybe it is acceptable to split this changes up into several smaller steps:
1. introduce file interface with just encapsulating the current functionality from the generic_fopen
2. replace by win32 api at a minimum level
3. extend teh win32 api usage

pnedev:
@chcg ,

I appreciate you being constructive on this but I don't think it will make things easier to review.
I also consider it double-work for both sides.

I can describe the change in more details if that will help.

pnedev:
@donho ,

Please read this:
https://community.notepad-plus-plus.org/post/48725

pryrt:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133#issuecomment-644329771

The problem keeps occurring for users

donho:
@pnedev 
The branch is deleted so I cannot reopen this PR.
Could you redo it? That'll be our last try to fix #6133

pnedev:
@donho,

Here is the new PR: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10598

PR #6191, opened 2019-10-02
Fix Portuguese localization

### LINKED ISSUE #6167: pt-PT translation error in search box "Wrap around"
### Description of the Issue
The portguese (pt-PT) translation translates the search box's "Wrap Around" to "Moldar Texto", which is not an accurate translation. Ideally, it would be something like "Pesquisa circular", which means "circular search". "Moldar texto" means "Wrap text" and does not represent the actual behaviour.

xylographe:
Replace incorrect "Moldar Texto" with "Voltar ao início"

Fixes #6167


donho:
@xylographe  Could you add new translation entries for some hotkeys (3b77016) please?

xylographe:
@donho Done

Issue #5158, opened 2018-12-20
Native support for Windows 10 on ARM (ARM64 Win32)


jkunkee:
While I'm filing this feature request as a long-time fan of Notepad++, please note that I'm also a Microsoft software engineer whose day job is expanding the Windows 10 on ARM (ARM64, both Win32 and UWP) software ecosystem.

I use the x86 ("32-bit") version of Notepad++ on a regular basis on a Windows 10 on ARM device, and it works very well. There is also a wide variety of compiled plugins available for it. There's a solid argument for not doing the extra work to support another architecture, though I see x64 hasn't been out of the question. :)

Notepad++ is a wonderfully light application, so the benefits of native ARM64 support are simple:

 * Lower power consumption
    * The idle loop takes ~0.2% CPU, but I tend to leave Notepad++ running. A lot.
 * Slightly better responsiveness
    * Seriously, when I started it a few minutes ago typing was just slightly slow to render. I tried again and couldn't notice the difference again. It's slight.

With Visual Studio 2017 15.9+ and Windows SDK 10.0.17763.0, ARM64 Win32 is fully supported (no `WindowsSDKDesktopARMSupport` even). Projects often just need to add an ARM64 configuration to build, the biggest questions being about binary dependencies and assembly. 

I may eventually get to prepping a PR, but in the meantime I'm happy to answer questions about what this requires.

zufuliu:
Hi @jkunkee 

Seems arm and arm64 SDK is different from x86 and x64, files count from 
```
Windows Kits\10\Lib\10.0.17763.0\um
    arm              348
    arm64            358
    x64              474
    x86              475
```
and shlwapi.lib is missing from arm and arm64 folder.

zufuliu:
After repair Windows SDK, all lib is here:
```
Windows Kits\10\Lib\10.0.17763.0\um
    arm              444
    arm64            454
    x64              474
    x86              475
```
But "Inherited values" for linker input is different:
for x64:
```
gdi32.lib
winspool.lib
comdlg32.lib
advapi32.lib
shell32.lib
ole32.lib
oleaut32.lib
uuid.lib
odbc32.lib
odbccp32.lib
```

and for arm64:
```
kernel32.lib
user32.lib
```

jkunkee:
[...]

PR #6196, opened 2019-10-04
update for arm64 build

chcg:
see build https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/27890847 for issue  #5158

Seems at least adding the config makes sense even, if not build on a regular base via appveyor.

### DIFF
--- a/PowerEditor/visual.net/notepadPlus.vcxproj
+++ b/PowerEditor/visual.net/notepadPlus.vcxproj
@@ -1,6 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Unicode Debug|ARM64">
+      <Configuration>Unicode Debug</Configuration>
+      <Platform>ARM64</Platform>
+    </ProjectConfiguration>
     <ProjectConfiguration Include="Unicode Debug|Win32">
       <Configuration>Unicode Debug</Configuration>
       <Platform>Win32</Platform>
@@ -8,6 +12,10 @@
     <ProjectConfiguration Include="Unicode Debug|x64">
       <Configuration>Unicode Debug</Configuration>
       <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Unicode Release|ARM64">
+      <Configuration>Unicode Release</Configuration>
+      <Platform>ARM64</Platform>
     </ProjectConfiguration>
     <ProjectConfiguration Include="Unicode Release|Win32">
       <Configuration>Unicode Release</Configuration>
@@ -38,12 +46,23 @@
     <CharacterSet>Unicode</CharacterSet>
     <WholeProgramOptimization>true</WholeProgramOptimization>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Unicode Release|ARM64'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <PlatformToolset>v141</PlatformToolset>
+    <CharacterSet>Unicode</CharacterSet>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
+  </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Unicode Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <PlatformToolset>v141</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Unicode Debug|x64'" Label="Configuration">
+    <ConfigurationType>Application</ConfigurationType>
+    <PlatformToolset>v141</PlatformToolset>
+    <CharacterSet>Unicode</CharacterSet>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Unicode Debug|ARM64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <PlatformToolset>v141</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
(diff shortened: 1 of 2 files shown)


donho:
@chcg I do believe it does compile Notepad++ ARM64 binary. However, Scintilla should be in ARM64 as well, that implies to compile PCRE of boost - without this part we cannot make a minimalist Notepad++ run. And I'm not even talking about WinGUp, PluginList and all plugins. Any suggestion?

chcg:
@donho See https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/27890847/job/rxifl9trr50sln29, also scintilla is build with it, see at:

  Creating library ..\bin\SciLexer.lib and object ..\bin\SciLexer.exp
LINK : warning LNK4010: invalid subsystem version number 5.01; default subsystem version assumed
Generating code
Finished generating code
	LIB /OUT:..\bin\libscintilla.lib .\AutoComplete.obj .\CallTip.obj .\CaseConvert.obj .\CaseFolder.obj .\CellBuffer.obj .

I should be possible to use at least c++11 regex by scintilla. I could also test the compatiblity together with boost. Guess this might work also without to big troubels.

Also did already some time ago an experiment with npppluginlist:
https://github.com/chcg/nppPluginList/tree/appveyor_for_arm
, so that is also no big issue.

WinGup might be more interesting. I would start to provide PRs for the plugins part of the installer as a starting point, if you think arm64 support is of interest.

donho:
@chcg Could you provide me the steps to follow to compile Scintilla (with boost's PCRE) please ?

chcg:
@donho The magic is done via:

```
 if "%platform%"=="arm64" set archi=amd64_arm64
 - call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" %archi%
```
to setup the compiler flags accordingly for arm64. Apart from that the scintilla build is just done like the normal ones for x64,x86:


> - if "%configuration%"=="Unicode Debug" set scintilla_debug=DEBUG=1
> - if "%configuration%"=="Unicode Release" set scintilla_debug=
> - nmake SUPPORT_XP=1 %scintilla_debug% -f scintilla.mak

also SUPPORT_XP=1 doesn't make sense for arm64. It is probably better to remove that flag in case of the arm build.

chcg:
@donho I overlook the "with boost" part. I haven't check that out yet.

Further hint:
See https://www.scintilla.org/ScintillaHistory.html for 3.6.6:

> 
>  Released 24 May 2016.
> C++ 11 <regex> support built by default. Can be disabled by defining NO_CXX11_REGEX. 

That one uses https://en.cppreference.com/w/cpp/regex/ecmascript as default mode, see
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/src/Document.cxx#L3065

Maybe the readme should be update to reflect that change compared to the previous builtin limited POSIX version with the scintilla update to 4.1.4/4.2.0.

donho:
@chcg 

C++11 RegEx is attractive because it allows to get rid of the dependency on boost.
However, some Notepad++ built-in features uses heavily (like Function List & html/xml tag matching) so there could be some regressions if we switch the RE engin.

> Maybe the readme should be update to reflect that change compared to the previous builtin limited POSIX version with the scintilla update to 4.1.4/4.2.0.

Feel free to modify this part.

sasumner:
@donho said:

>C++11 RegEx is attractive because ... 

>.. some Notepad++ built-in features uses heavily ... so there could be some regressions (due to C++11 RE performance and its different syntax flavor) if we switch the RE engin.

Not to mention ticking off the user base that is used to the current regex engine.  I think there would have to be significant, demonstrable benefits to changing the engine in order to actually do so.

chcg:
@donho My hint was not meant for the main version which should still be build with boost to avoid loosing PCRE. Just for the build without boost which no longer uses the oldstyle limited regular engine, but due to the changed default by scintilla the C++11 regex flavour.

donho:
@sasumner 
>  I think there would have to be significant, demonstrable benefits to changing the engine in order to actually do so.

What does make you say so? Could you elaborate it?

donho:
@chcg I have tried to build Scintilla come with Notepad++ without boost:
[SciLexer.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/3775342/SciLexer.zip)
(release, signed) 
After some quick tests, html/xml tag matching works but not  Function List.

sasumner:
@donho 

>What does make you say so? Could you elaborate it?

I already said it:  Making the existing user base angry at having to learn something new.  But, give them something better (yet to be determined) and it will be okay. :-)

donho:
@sasumner 
English is not my native language so sometimes I could misunderstand or ignore the hidden meaning :)
So you're saying that you are for changing RExpr engin from boost's one to C++11's one, because that make the community learn another RE syntax flavor, even it could make users irate.

Do I understand you correctly ?

LordNyriox:
@donho:  English is my native language, so I cannot help but comment here… 

> Do I understand you correctly? 
Not exactly.  Near as I can tell, @sasumner is arguing in favor of changing the regex engine *despite* the inevitable frustration—but advises that you also add some new features to NPP's regex subsystem to give the users something to look forward to.

donho:
@LordNyriox 
> but advises that you also add some new features to NPP's regex subsystem to give the users something to look forward to.

Thank you for the interpretation :)
Concretely, what kind of regex subsystem could it can be? It's still vague for me.

LordNyriox:
@donho: 

> Concretely, what kind of regex subsystem could it be? It's still vague for me. 

I get the impression that @sasumner was leaving that one up to you to decide.  Either that or he figured the final decision could be left until after Boost was already removed as a dependency.  [*shrug*] 

I presume some sort of quality-of-life feature would be best—but I really am not the best person to ask in this case. 

I do not understand regex well enough to use it, much less make informed recommendations about enhancing others' usage of it.

sasumner:
@donho @LordNyriox Given the apparent language barrier, please disregard any of my prior comments.

chcg:
Seems boost could be build with commandline, e.g.:

`.\b2 -j2 architecture=arm link=static runtime-link=static threading=multi address-model=64 stage --toolset=msvc-14.1 --without-python`

the relevant parameter which differ from the x86/x64 build is:
architecture=arm
and
address-model=64
as we havearm64.

Further hint for x86 and x64:
Instead of building the regex lib on your own it is possible to use the prebuild windows binaries:
https://bintray.com/boostorg/release/boost-binaries/1.70.0#read
or
https://sourceforge.net/projects/boost/files/boost-binaries/1.70.0/

chcg:
Close this one as the changes are now part of https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9594

Issue #4339, opened 2018-03-24
Inconsistent crash when dragging tabs


vtad4f:
### Description of the Issue
With several tabs open, when I drag them around notepad++ will occasionally crash. It happened 2-3 times one day and not again since. 

### Steps to Reproduce the Issue
1. Open a few tabs
2. Re-order the tabs by dragging-dropping

### Expected Behavior
Not crashing

### Actual Behavior
Crashing

### Debug Information
Notepad++ v7.5.6 (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

MAPJe71:
Can not reproduce on Windows 10 64-bit with Notepad++ 7.5.6 32 bit or Notepad++ 7.5.6 64 bit.

vtad4f:
Hi. Sorry for not providing this earlier. The info below is my current debug info. I saw this issue within the last two weeks, so it could also be 7.5.5 (32 bit).

Notepad++ v7.5.6   (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

vtad4f:
I also frequently run with these cmd line options: **-multiInst -nosession**

AngryGamer:
What kind of crash? Does notepad++ close itself? Does the Tab Bar freeze? Do you get an error message?

vtad4f:
It wasn't just the tab bar that froze. I did not get an error message. Windows popped up the standard "Check online for a solution and re-open", "Close the program" options or something similar.

I haven't seen it since that day. It may have been an issue that was only present in 7.5.5.

Pokechu22:
I can also reproduce this; it's happened 2 or 3 times (on separate days).  I did _not_ get a message from windows about the application crashing; it just hung for a few seconds and then closed.

I will note that when dragging the tab, the crash happens after the drag has stopped, and the file name in the dragged tab is blank.

[...]

PR #6202, opened 2019-10-08
tab fix

balagansky:
This resolves the issues I posted about in this issue thread:

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4339

### DIFF
--- a/PowerEditor/bin/change.log
+++ b/PowerEditor/bin/change.log
@@ -1,20 +1,39 @@
-Notepad++ v7.7.1 enhancements and bug-fixes:
+Notepad++ v7.8 enhancements and bug-fixes:
 
-1.  Fix drag and drop tab regression: dragging tab is switched to another tab issue.
-2.  Fix an input regression: Cyrillic, Turkish and other languages input issue in ANSI mode.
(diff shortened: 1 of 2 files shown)


balagansky:
Few more observations:

activateAt is identical to setActiveTab, except the former is used much more rarely and does not actually send a selection change message, despite preparing it. This could use a closer look.

It's possible that the focus message is also inappropriate in other situations besides session loading.

There likely may be a more appropriate/direct fix than not sending the focus message. I didn't dig too deeply to determine exactly why sending the focus message in that circumstances causes problems.

As far as I can tell, not sending the focus message on session load has no ill effects - at least not in my testing.

balagansky:
Looks like the problem can still occur via other triggers besides session loading, even if the focus message is completely commented out. So this fix won't help much, it seems. Hmm...

balagansky:
Here's a better fix.

Still not completely sure whether this is the best fix possible, but it seems to work.

balagansky:
Re-created this PR from a branch (instead of master - oops): https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6210

PR #6210, opened 2019-10-11
Fix to tab clicks under multi-line mode

balagansky:
MOUSEMOVE message could be sent continuously while a mouse button is held down, even if the mouse isn't moving, so using _dragCount is to determine whether we are dragging is unsafe.

This addresses at least some of the issues mentioned here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4339

Previously, merely clicking (without dragging) a tab could cause unexpected tab switching, sometimes corruption (tab becoming blank/inaccessible), and eventually crashing.

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -586,6 +586,9 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 
 		case WM_LBUTTONDOWN :
 		{
+			_dragRefPoint.x = LOWORD(lParam);
+			_dragRefPoint.y = HIWORD(lParam);
+
 			if (_drawTabCloseButton)
 			{
 				int xPos = LOWORD(lParam);
@@ -633,15 +636,18 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 
 		case WM_MOUSEMOVE :
 		{
+			POINT p;
+			p.x = LOWORD(lParam);
+			p.y = HIWORD(lParam);
+
 			if (_mightBeDragging && !_isDragging)
 			{
 				// Grrr! Who has stolen focus and eaten the WM_LBUTTONUP?!
 				if (GetKeyState(VK_LBUTTON) >= 0)
 				{
 					_mightBeDragging = false;
-					_dragCount = 0;
 				}
-				else if (++_dragCount > 2)
+				else if (_dragRefPoint.x != p.x || _dragRefPoint.y != p.y)
 				{
 					int tabFocused = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURFOCUS, 0, 0));
 					int tabSelected = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0));
@@ -663,10 +669,6 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 					}
 				}
 			}
-
-			POINT p;
-			p.x = LOWORD(lParam);
-			p.y = HIWORD(lParam);
 
 			if (_isDragging)
 			{
@@ -771,7 +773,6 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 		case WM_LBUTTONUP :
 		{
 			_mightBeDragging = false;
-			_dragCount = 0;
 
 			int xPos = LOWORD(lParam);
 			int yPos = HIWORD(lParam);
--- a/PowerEditor/src/WinControls/TabBar/TabBar.h
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.h
@@ -224,7 +224,7 @@ protected:
     static bool _doDragNDrop;
 	// drag N drop members
 	bool _mightBeDragging = false;
-	int _dragCount = 0;
+	POINT _dragRefPoint = {LONG_MAX, LONG_MAX};
 	bool _isDragging = false;
 	bool _isDraggingInside = false;
     int _nSrcTab = -1;


balagansky:
I also noticed that activateAt is identical to setActiveTab except the former is used much more rarely and does not actually send a selection change message, despite preparing it. This could use a closer look (perhaps at least delete the message preparation code that has no effect in activateAt).

donho:
There are several iteration recently to enhance tab dragging issue. It should be fixed.
I have tried your PR, and I don't see any benefit. If you cannot point out the difference before and after your PR, I have to decline it.

balagansky:
I don't know why the problems this PR addresses doesn't manifest for everybody / inconsistently. Perhaps it has to do with certain Windows updates. It also seems sensitive to exactly what tabs are open / the width of the window.

All I can really do is show a recording of the problems this addresses. I'm quite sure that the problems had not been fixed as of submission of this PR, because I tried running the current code and still saw the same behavior.

I'm having trouble reproducing some of the issues (the crashing/corruption in particular) with a fresh set of files. For now, here's a recording of a problem that's reproducible on I think any set of files:

npptabs

I'll see if I can repro the others later.

balagansky:
Also I have been using this PR locally since I wrote it. I have had zero issues since then, whereas I had issues constantly before that.

balagansky:
So, visually, what I see is most similar to this post: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4339#issuecomment-429061948

Except I don't actually need to drag anything for that to occur. Merely clicking on an inactive tab would cause that sometimes. But it appears related to npp thinking I'm dragging when I'm not.

I think basically the issue is that npp goes into a drag state at an unexpected/invalid time, and somehow this then causes other problems.

balagansky:
In that gif, clicking tabs and moving the mouse slightly. The circles represent clicks.

To reliably reproduce that behavior, you may need to start by loading a session file, and I think the active tab needs to be something other than the first one. It's certainly possible to repro under other circumstances, but I don't know how to do so reliably.

Which debug info are you looking for?

donho:
> Which debug info are you looking for?

Menu `? -> Debug Info...`

donho:
> In that gif, clicking tabs and moving the mouse slightly. The circles represent clicks.

I cannot reproduce it at all. I believe this bug has been fixed recently. We don't see any similar issue reported since.

My Debug Info:
Notepad++ v7.8   (32-bit)
Build time : Oct 21 2019 - 12:53:19
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1903
OS Build : 18362.418

Please check with the latest version.

balagansky:
I spent some more time investigating this.

**The issue in the gif seems to be distinct from the issues I was initially experiencing**. I am able to repro it on two different machines (Win 7 and Win 10), with or without my fix, it turns out, and on HEAD (7.8.1). 

I still cannot repro the issues I was seeing before again, but I _did_ repro them (clicking on inactive tab creates a corrupt blank tab, followed by a crash on clicking the blank tab) _one time_ yesterday using 7.7.1 - unfortunately I wasn't recording or debugging.

Here's the debug info of the machine where I've experienced the click on tab -> corruption -> crash issues:

Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)

Here's the debug info of another machine where I can also repro the gif issue:

Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)

This particular issue isn't debilitating and I'm not going to worry about it now - I'll assume it's not related to the initial issues I was concerned with (but maybe it is). I'll open a separate issue item for it.

Back to this PR. According to my memory, when I was initially compelled to look into this and write the PR, I was able to reproduce the crash issue fairly reliably - using the latest release and the HEAD. Now I cannot, but I have reproduced it one time, at least, as mentioned above, so I know I'm not imagining it or making a mistake re using an old version.

With my fix in place, the crash has occurred zero times, and when I was initially testing it, there seemed to be a clear difference between a build with the fix and without the fix - one crashed reliably, the other never crashed.

What I meant by "quick fix, not carefully tested", is that I think there are other problems that in combination with the handling of WM_MOUSEMOVE cause these symptoms. My addresses situations where WM_MOUSEMOVE may be sent without the mouse actually moving. There might be other mistakes in logic elsewhere that are combining with this somehow to cause the buggy behavior. I didn't have time to look much deeper into that, but since my fix appeared to work, that also seemed good enough for now, at least.

A bit more on why WM_MOUSEMOVE might be sent repeatedly... It seems this can happen for various reasons. When I was initially testing it, it appeared, that I was seeing WM_MOUSEMOVE events at rapid successive intervals even if I was not moving the mouse. Currently it seems I get them roughly once per second. It turns out that in my current environment, the culprit is Task Manager's process list tab. When the process list tab is active, Task Manager apparently sends a WM_MOUSEMOVE event to every window (whee!). There are apparently other potential causes of this as well, and the moral of the story is that you cannot rely on successive WM_MOUSEMOVE events having differing mouse positions.. which iirc the tab code appeared to assume.

Am I 100% sure that this fix really works? Not quite, but I now have a week of accumulated usage that seems to suggest it does, in addition to my initial testing. In any case there appears to be no downside to this change and it clearly corrects the determination of dragging state. Without this change, the program can enter dragging state even when the mouse is not moved - whether or not that leads to a crash, this is not good.

Sorry, I also wish that this was a more straightforward issue!

balagansky:
Heh.

While trying to record a different issue, I managed to repro the crash, and then managed to do it again, 3/3 times. Both with 7.7.1 and with my fix applied. So it turns out that my fix doesn't work after all, and up until this point not seeing the crash with the change in place was just a coincidence.

Here's the gif. I'll repost it on the linked issue as well.

npptabscorruption3

The first time I ran into that, instead of blank tabs, I got chinese characters - like someone else had in the issue thread. I have a recording of that too, but I can't post it because it revealed proprietary information behind the window after the window disappeared. Here's a screenshot instead:

image

balagansky:
So this PR may not actually prevent any current buggy behavior. I think you should still consider merging it so that drag state is calculated more correctly - that could at least prevent issues or confusing later on. Feel free to close the PR if you don't think the change adds value.

AngryGamer:
I wonder whether this patch would exacerbate erroneous drag toggling during fast motion. For example, if someone were to inadvertently click before their mouse is at a complete stop or if they click a tab and immediately dart the cursor down to click something else. If this is a problem and this patch exacerbates it, then it would require the addition of time checks to ignore like 50ms to 100ms of mouse movement.

balagansky:
> 
> 
> I wonder whether this patch would exacerbate erroneous drag toggling during fast motion. For example, if someone were to inadvertently click their mouse right before it's at a complete stop or if they click a tab and immediately dart the cursor down to click something else. If this is a problem and this patch exacerbates it, then it would require the addition of time checks to prevent dragging for >100ms after a mouse click. Other than that, I don't see a problem with this enhancement.

Dragging shouldn't really do anything until the mouse is moved far enough to be over another tab anyway, so that doesn't seem like a problem, although I'm not intimately familiar with this code. Actually, the bug I was trying to fix with this is basically that - something was happening even if the mouse never moved, but turns out the fix for that is not this change (luckily my commit message doesn't specify that, so no need to update it).

One way to mitigate that potential issue is to require a certain distance of pixels to be traveled before entering a drag state. Currently the distance is implicitly 1 pixel, but it could be set to, say, 5 pixels, or some value based on resolution. I don't think this is necessary, though, as I didn't notice any ill effects from erroneous microdragging.

donho:
@balagansky I don't understand why you tested on 7.7.1 whereas the latest version is 7.8.
I have tested v7.8 and cannot reproduce the issue you showed while dragging tab. I didn't reproduce the crash either.
```
Notepad++ v7.8   (32-bit)
Build time : Oct 21 2019 - 12:53:19
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1903
OS Build : 18362.418
```

balagansky:
7.7.1 was the latest auto-update version. I have tested/repro'd this initially about 20 days ago, before the version increment to 7.8 on HEAD, and again this week, in addition to on 7.7.1.

I'm not particularly surprised that you can't repro as the preconditions seem very specific and possibly not entirely deterministic. It my most recent repro, I used a specific session file at the start of the repro steps, and it may depend on what tabs are in the session file, but unfortunately I can't share it since they are proprietary work files. If I have time later, I'll try to repro starting with a public set of files. I don't think that part of the repro process is as finicky as the rest of it.

I am certain the issues are still there and have not been fixed, and that I've repro'd them on two different machines, two different OS's now.

balagansky:
The update mechanism from within the tool still seems to point to 7.7.1 (says "no update available"). I grabbed the 7.8 installer on my win 7 machine from the main site and was immediately able to repro, as before.

Notepad++ v7.8   (32-bit)
Admin mode : ON
Local Conf mode : OFF
OS Build : 7601.0

balagansky:
Here is a complete repro gif on my win 7 machine using the 7.8 installer and only the files in the npp program directory. Gifs below for both the tab swap issue and the crash (for some reason now, on this machine, attempting to close the corrupt tabs doesn't crash - nothing much happens, but the tabs are unusable).

Notepad++ v7.8   (32-bit)
Admin mode : ON
Local Conf mode : OFF
OS Build : 7601.0

tab swap:
npptabclick

tab corruption:
npptabcrash

If you still cannot repro based on this, I'm not sure what else I can do.

LordNyriox:
@balagansky:  I noticed all your reports are using 32-bit Notepad++ (even though your OS is apparently 64-bit). 

Out of curiosity, have you tried to reproduce these bugs in the 64-bit build of Notepad++?

balagansky:
> @balagansky: I noticed all your reports are using 32-bit Notepad++ (even though your OS is apparently 64-bit).
> 
> Out of curiosity, have you tried to reproduce these bugs in the 64-bit build of Notepad++?

Yes. Makes no difference.

> **P.S.** Also, why the heck are you opening the Notepad++ main executable as a tab in Notepad++?

Because I'm trying to provide repro steps with readily available public files.

> An executable trying to read its own binary data (as text, no less) sounds like a quick way to trigger program crashes.

Perhaps it sounds like that, but that's not true. The only situation in which that might be a problem is the general situations of file lock or read/write contention, but since everything is reading, nothing is writing, and npp reads without locking, there is no issue here.

balagansky:
I took a quick look at the code to see what this PR actually would fix, and here is one example (there might be others):

Click + Ctrl+Scroll in the tab area rotate the active tab. Without this PR, if MOUSEMOVE events are being sent externally when the mouse is not moved the user is, this rotation will stop working after the next MOUSEMOVE event is processed.

Basically logic based on _isDragging is affected.

donho:
Fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fce74dd894a33bc169d0be75b789a0606aae8474

balagansky:
> 
> 
> Fixed by [fce74dd](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fce74dd894a33bc169d0be75b789a0606aae8474)

That commit does _not_ address the issue in this PR. This is a separate issue.

balagansky:
@AngryGamer,

About unintentional drag swaps - I feel that's a fringe concern and I'm not sure that that's a bigger issue than the reverse case where you _intentionally_ want to do a drag quickly. With "protection" against "fast dragging", some users may on the contrary expect a drag and not get it. Which expectation is more correct is not clear cut.

There's also another technique sometimes used to decrease accidental drags - and that's to require some number of pixels traveled greater than 1. But that's also subject to tuning and could run into problems with resolution, DPI, etc.

My current opinion is that we're better off with no such "protection". But I acknowledge that others may feel differently. Ultimately I don't particularly care which is chosen, but I'd still like to see the assumption that MOUSEMOVE implies actual mouse movement resolved.

If we really do want a delay before entering a drag state, using a counter is not a great solution because it's dependent on how often the input check is called. I think as you suggested, using a time check might work better for that. Maybe 100ms is reasonable.

Perhaps even adding both techniques (time and distance) would be best. Time works for the situations you describe, but distance would work for cases where a user intends to click, not drag, but accidentally shifts the mouse slightly, and takes longer than 100ms or whatever the time threshold is to execute the click.

Feel free to fork this PR and resubmit it with "protection" added. But I'd lean towards just putting this one in as is and then adding PRs on top of that for refinement.

AngryGamer:
@balagansky 

>  I'd still like to see the assumption that MOUSEMOVE implies actual mouse movement resolved.

I don't necessarily disagree, but as I pointed out in my first paragraph: 
`at present it doesn't offer any tangible benefit [to the end user] over existing code ... [incorrectly toggled] _isDragging [is] remedied on LBUTTONUP ... So basically, [there's not much difference during normal use except with regard to rapid actions]`

> Feel free to fork this PR and resubmit it with "protection" added. But I'd lean towards just putting this one in as is and then adding PRs on top of that for refinement.

I'm not submitting PRs or planning to attempt much in terms of contributing here _for reasons_. I'm replying to _you_ because I've got no animosity towards _you_ and I'm familiar with the code. Also, getting this PR accepted as-is would likely require consensus on the protection issue not being necessary. I don't care one way or the other about it, but I was merely pointing it out.

---
[_repost from other discussion_](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fce74dd894a33bc169d0be75b789a0606aae8474#commitcomment-35996841)
> _In theory_ #6210 is more accurate, but at present it doesn't offer any tangible benefit over existing code. The reason is that _isDragging being incorrectly toggled by phantom MOUSEMOVE messages during LBUTTONDOWN _should be_ remedied on LBUTTONUP. So basically, no actual harm is done unless someone managed to move their cursor over another tab during this transition. This conclusion leads to me to postulate that:
> 
> There's only a potential problem during **rapid actions**. For example, you click on the edge of a tab and immediately dart past another tab to click on something else. This would likely trigger a tab swap. Was it intentional? My opinion is that _when it's that rapid_ **probably not**. However, #6210 has no protection against this [as suggested here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6210#issuecomment-546583791), but the existing code does in fact offer some protection (_so #6210 is a regression in some respects_), but the existing protection would be ineffective during instances of phantom MOUSEMOVE messages (_so existing protection isn't always effective_).
> 
> **My conclusion is that:**
> 
> If accidental tab swapping during rapid actions _is a problem_ and _should be protected against_, then #6210 has _the potential_ to be a better solution. Not only is it inherently immune to phantom MOUSEMOVE messages, but protection against accidental tab swaps could easily be added to it.

AngryGamer:
@balagansky 

>  I'd still like to see the assumption that MOUSEMOVE implies actual mouse movement resolved.

I don't necessarily disagree, but as I pointed out in my first paragraph: 
`at present it doesn't offer any tangible benefit [to the end user] over existing code ... [incorrectly toggled] _isDragging [is] remedied on LBUTTONUP ... So basically, [there's not much difference during normal use except with regard to rapid actions]`

> Feel free to fork this PR and resubmit it with "protection" added. But I'd lean towards just putting this one in as is and then adding PRs on top of that for refinement.

I'm not submitting PRs or planning to attempt much in terms of contributing to this project _for reasons_. I'm replying to _you_ because I've got no animosity towards _you_ and I'm familiar with the code. Also, getting this PR accepted as-is would likely require consensus on the protection issue not being necessary. I don't care one way or the other about it, but I was merely pointing it out.

---
[_repost from other discussion_](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fce74dd894a33bc169d0be75b789a0606aae8474#commitcomment-35996841)
> _In theory_ #6210 is more accurate, but at present it doesn't offer any tangible benefit over existing code. The reason is that _isDragging being incorrectly toggled by phantom MOUSEMOVE messages during LBUTTONDOWN _should be_ remedied on LBUTTONUP. So basically, no actual harm is done unless someone managed to move their cursor over another tab during this transition. This conclusion leads to me to postulate that:
> 
> There's only a potential problem during **rapid actions**. For example, you click on the edge of a tab and immediately dart past another tab to click on something else. This would likely trigger a tab swap. Was it intentional? My opinion is that _when it's that rapid_ **probably not**. However, #6210 has no protection against this [as suggested here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6210#issuecomment-546583791), but the existing code does in fact offer some protection (_so #6210 is a regression in some respects_), but the existing protection would be ineffective during instances of phantom MOUSEMOVE messages (_so existing protection isn't always effective_).
> 
> **My conclusion is that:**
> 
> If accidental tab swapping during rapid actions _is a problem_ and _should be protected against_, then #6210 has _the potential_ to be a better solution. Not only is it inherently immune to phantom MOUSEMOVE messages, but protection against accidental tab swaps could easily be added to it.

balagansky:
@AngryGamer, I gotcha. As I said in the commit thread (https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fce74dd894a33bc169d0be75b789a0606aae8474#commitcomment-35997274), I believe there _is_ still a tangible benefit to the user with this PR. Actually, I just found it now - I mentioned it earlier in this thread.

"Click + Ctrl+Scroll in the tab area rotate the active tab. Without this PR, if MOUSEMOVE events are being sent externally when the mouse is not moved the user is, this rotation will stop working after the next MOUSEMOVE event is processed."

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6210#issuecomment-547205198

AngryGamer:
@balagansky
> "Click + Ctrl+Scroll in the tab area rotate the active tab. 

Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling (which would toggle dragging and keep it active). A high DPI mouse is even more likely to cause this problem.

AngryGamer:
@balagansky
> "Click + Ctrl+Scroll in the tab area rotate the active tab. 

Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling the mousewheel (which would toggle dragging and keep it active). A high DPI mouse is even more likely to cause this problem.

AngryGamer:
@balagansky
> "Click + Ctrl+Scroll in the tab area rotate the active tab. 

Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling mousewheel (which would toggle dragging and keep it active). A high DPI mouse is even more likely to cause this problem.

AngryGamer:
@balagansky
> "Click + Ctrl+Scroll in the tab area rotate the active tab. 

Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling mousewheel (which would toggle dragging and keep it active). A high DPI mouse is even more likely to cause this problem.

The way you'd normally fix this problem is by preventing or terminating any dragging while key modifiers are held.

AngryGamer:
@balagansky
> "Click + Ctrl+Scroll in the tab area rotate the active tab. 

Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling mousewheel (which would toggle dragging and keep it active). A high DPI mouse is even more likely to cause this problem.

The way you'd normally fix this problem is by preventing or terminating any dragging while [modifier keys](https://en.wikipedia.org/wiki/Modifier_key) are held.

AngryGamer:
@balagansky
> "Click + Ctrl+Scroll in the tab area rotate the active tab. 

Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling mousewheel (which would toggle dragging and keep it active). A mouse with a very sensitive sensor is even more likely to cause this problem.

The way you'd normally fix this problem is by preventing or terminating any dragging while [modifier keys](https://en.wikipedia.org/wiki/Modifier_key) are held.

balagansky:
> 
> 
> @balagansky
> 
> > "Click + Ctrl+Scroll in the tab area rotate the active tab.
> 
> Are you talking about the CTRL+MouseWheel hotkey for cycling open tabs?

Yeah.

> I never considered the hotkey would be used while holding left click. I'm not sure this PR would solve that problem either. For example, a common habit for people is to move their mouse a little while scrolling mousewheel (which would toggle dragging and keep it active). A mouse with a very sensitive sensor is even more likely to cause this problem.

It would solve the truly unexpected cause at least when the mouse isn't moving.

> The way you'd normally fix this problem is by preventing or terminating any dragging while [modifier keys](https://en.wikipedia.org/wiki/Modifier_key) are held.

In that case, yeah, probably.

This was a contrived example I inferred from glancing at the code. The issues in practice are probably minor, although perhaps some less obscure situation exists that's affected.

This PR could be redone to preserve the previous "expected" behavior in terms of requiring two updates instead of one before going into a drag state while still addressing the issue of assuming WM_MOUSEMOVE implies actual movement.

Regardless, I don't really care at this point either. Probably not worth anybody's time if we're not aiming for high robustness.

AngryGamer:
@balagansky 
> This PR could be redone to preserve the previous "expected" behavior in terms of requiring two updates instead of one before going into a drag state while still addressing the issue of assuming WM_MOUSEMOVE implies actual movement.

The existing behavior toggles dragging on the third update.
0->1
1->2
3->Dragging

> Regardless, I don't really care at this point either. Probably not worth anybody's time if we're not aiming for high robustness.

That would be purely a bug fix/enhancement with no drawbacks over existing code.

You could try another PR and see what happens, but [you should be somewhat pessimistic.](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6274#issuecomment-550088540) After all, this is basically a fix for inconsistencies in when dragging starts under different circumstances.

AngryGamer:
@balagansky 
> This PR could be redone to preserve the previous "expected" behavior in terms of requiring two updates instead of one before going into a drag state while still addressing the issue of assuming WM_MOUSEMOVE implies actual movement.

The existing behavior toggles dragging on the third update.
0->1
1->2
3->Dragging

> Regardless, I don't really care at this point either. Probably not worth anybody's time if we're not aiming for high robustness.

That would be purely a bug fix/enhancement with no drawbacks over existing code.

You could try another PR and see what happens, but [you should be somewhat pessimistic.](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6274#issuecomment-550088540) After all, this is basically a fix for inconsistencies in when dragging starts under different circumstances and the benefits are very minor.

AngryGamer:
@balagansky 
> This PR could be redone to preserve the previous "expected" behavior in terms of requiring two updates instead of one before going into a drag state while still addressing the issue of assuming WM_MOUSEMOVE implies actual movement.

That would be purely a bug fix/enhancement with no drawbacks over existing code.

And also, the existing behavior toggles dragging on the third update.
0->1
1->2
3->Dragging

> Regardless, I don't really care at this point either. Probably not worth anybody's time if we're not aiming for high robustness.

You could try another PR and see what happens, but [you should be somewhat pessimistic.](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6274#issuecomment-550088540) After all, this is basically a fix for inconsistencies in when dragging starts under different circumstances and the benefits are very minor.

PR #6231, opened 2019-10-18
Update brazilian_portuguese.xml 

francescosacco:


chcg:
@francescosacco See #6231, could you create a common version together with @hdstec.

PR #6233, opened 2019-10-19
Fix initialization of _monospacedFontFinfDlg parameter

xylographe:
Fixes the initialization of the _monospacedFontFinfDlg parameter, which was always set to false.

Also fixes typo **`FinfDlg`** ==> **`FindDlg`** (in a separate commit for convenience of review).


xylographe:
@chcg 
Actually, this is a bug fix, not an optimization: The initial value of `_monospacedFontFinfDlg` is **false**, and while processing `config.xml` the value should be set to **true** if the stored configuration is `yes`, but without this patch that doesn't happen.

The problem was introduced in commit dde38183d3407f0b39ee67260da255bde51b324c (see `Parameters.h` and `Parameters.cpp`).

donho:
@xylographe Agree with you from the code.
But how can I reproduce the bug ?

donho:
Oups, sorry. I can reproduce it now.

PR #6237, opened 2019-10-20
Update czech.xml translation to v7.8.1

xomx:


xomx:
Maybe it is just me, but the new "Use monospaced font in Find dialog" checkbox setting in Preferences->MISC. does not survive the N++ restart (e.g. checked checkbox will became unchecked again...).

xylographe:
@xomx 

> checked checkbox will become unchecked again

Correct. Fixed in PR #6233.

PR #6239, opened 2019-10-20
Fix Italian localization

xylographe:


Ices-Eyes:
@xylographe hope i've made it right... :)

xylographe:
Thank you, @Ices-Eyes.

Issue #6244, opened 2019-10-21
Alt+Tab doesn't show NPP


PalashBansal:
### Description of the Issue
When a subwindow is opened inside NPP, and we press windows provided shortcut Alt+Tab to switch to other opened application, NPP doesn't come up in options.
This happens only for some subwindows like "Rename Current Tab" or "Close All > Save dialog" opened. If "Preference" dialog is opened, issue doesn't occur.

### Steps to Reproduce the Issue
1. Open either "Rename current tab" by right clicking on new tab(doesn't exist in file system) followed by "Rename" option. OR Keep 2 or more files unsaved and click "Close All" option to get "Save" dialog.
2. Keep "Alt+Tab" shortcut pressed and see NPP missing.
3. Switch to some other application.
4. Again try above step 2. And switching to NPP is not possible.
**Workaround-**
To switch to NPP from taskbar only.

### Expected Behavior
"Alt+Tab" should show NPP application always(if NPP is in launched state)

### Actual Behavior
NPP doesn't appear in application options from "Alt+Tab"

### Debug Information
Seems like issue is easily reproducible for every user, still enjoy-
Notepad++ v7.8   (64-bit)
Build time : Oct  7 2019 - 00:55:55
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.23833
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
>Seems like issue is easily reproducible for every user

Not for me.  I can successfully alt+tab back to Notepad++ in any of the above scenarios.  Maybe the difference is one of Win7 versus Win10?

Me:

Notepad++ v7.8   (64-bit)
Build time : Oct  7 2019 - 00:55:55
Path : C:\.........\Releases\npp.7.8\npp.7.8.bin.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1709
OS Build : 16299.1387
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

Ekopalypse:
It works for me on Windows 7 x64 BUT only one time.
Follow the steps as described, 
- switch to another running application by using alt+tab
- switch back to npp using alt+tab
- press alt+tab again -> Npp icon is gone.

[...]

PR #6252, opened 2019-10-22
Fix for Alt+Tab doesn't show NPP #6244

PalashBansal:
**Issue is because of Tool window being used instead of normal window for these 2 dialogs- "Close All" and "Rename"**
Known issue already discussed on- https://stackoverflow.com/questions/3760571/showdialog-makes-app-window-disappear-from-windows-alt-tab-list
Excerpt from https://docs.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles -

_WS_EX_TOOLWINDOW 0x00000080L_ | The window is intended to be used as a floating toolbar. A tool window has a title bar that is shorter than a normal title bar, and the window title is drawn using a smaller font. **A tool window does not appear in the taskbar or in the dialog that appears when the user presses ALT+TAB.** If a tool window has a system menu, its icon is not displayed on the title bar. However, you can display the system menu by right-clicking or by typing ALT+SPACE.
-- | --

_DS_MODALFRAME_ is added for "Close All" dialog, because after making it normal window, it started showing default empty windows file icon.


donho:
I can not accept your PR if I cannot reproduce the bug.
Could you provide the detail instructions to reproduce it?

xylographe:
Just as a reference to all, reproduction instructions in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6244#issuecomment-545389473.

donho:
@xylographe 
Please confirm me if this PR fix the issue.

xylographe:
@donho 
Yes, this PR fixes the issue in both cases: "Rename Current Tab" and "Save" dialogue.

OS Build : 7601.0

PR #6262, opened 2019-10-23
Update unitTest

Mithil467:
Fixed typos in comments

### DIFF
--- a/PowerEditor/Test/FunctionList/python/unitTest
+++ b/PowerEditor/Test/FunctionList/python/unitTest
@@ -104,7 +104,7 @@ def bfs(G, s):
         Args :  G - Dictionary of edges
                 s - Starting Node
         Vars :  dist - Dictionary storing shortest distance from s to every other node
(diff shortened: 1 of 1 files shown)


PowerEditor/Test/FunctionList/python/unitTest
```diff
@@ -277,4 +277,4 @@ def krusk(E_and_n):
                     break
                 s[j].update(s[i])
                 s.pop(i)
-                break
```
MAPJe71: Though not documented as such the "no new line at end of file" might be intended for testing proper operation of the Python Function List parser.

Mithil467: Oh, did I mess anything at the end of the file? I didn't change it though, I just edited the words in the comments, and that too in github web editor. What should I do to fix it?

donho:
PR for typo won't be accepted.
An issue is enough.

PR #6264, opened 2019-10-23
Capitalized first character of 'wrap' in 'Word wrap'

Mithil467:
All of the items have uppercase characters excluding some cases like 'Folder as Workspace'. So I thought it would be better to change 'Word wrap' to 'Word Wrap' to unify things up a bit.
Before:
Screenshot (67)
After:
Screenshot (66)


donho:
Only the bug fixing or feature request will accepted :

Issue #6274, opened 2019-10-26
Inconsistent multi-line tab click activation behavior


balagansky:
### Description of the Issue
When multi-line tabs are enabled, behavior when clicking on a tab is inconsistent depending on whether there is one row or more than one row of tabs.

When there is one row of tabs, mouse down activates an inactive tab. But when there multiple rows, inactive tabs only become active on mouse up.

### Steps to Reproduce the Issue
1. Enable multi-line tabs
2. Load up tabs such that there is more than one tab row
3. Click on an inactive tab

### Expected Behavior
Inactive tabs always become active on mouse down

### Actual Behavior
Inactive tabs only become active on mouse up when there are multiple tab rows

### Debug Information
Notepad++ v7.7.1   (32-bit)
Build time : Oct  7 2019 - 19:07:35
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1803
OS Build : 17134.1069
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

Also reproducible on master/HEAD

![npptabclick](https://user-images.githubusercontent.com/56180447/67611182-bbb09d80-f74c-11e9-8866-9535c62c729c.gif)

balagansky:
The changes suggested by @AngryGamer here seem to fix this: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6273#issuecomment-546753080

donho:
v7.7.1 is not head version and such issue could be fixed in v7.8
I cannot reproduce the issue you reported at all by using v7.8.

balagansky:
I can reproduce it just as easily on 7.8, on multiple machines.

I am truly baffled that you cannot.

donho:
Sorry. I read it again and it's not a bug to me.

Issue #6273, opened 2019-10-25
Erroneous Multi-line tab single row click/drag tab swap


balagansky:
### Description of the Issue
When using multi-line tabs, sometimes clicking and then dragging only slightly will cause tabs to swap even if the mouse is barely moved.

### Steps to Reproduce the Issue
1. Enable multi-line tabs
2. Open multiple tabs
3. Make sure the active tab is not the first tab
4. Re-open Notepad++, restoring the session
5. Click (and maybe hold a bit) an inactive tab
6. Move mouse slightly

### Expected Behavior
Nothing

### Actual Behavior
Previously active tab is swapped with clicked tab.

### Debug Information
Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

![npptabs](https://user-images.githubusercontent.com/56180447/67610259-62de0680-f746-11e9-8162-c3f5ee38f1ad.gif)

balagansky:
Removing these blocks of code seems to avoid the issue, but that alone may not be the right fix:

	// TCM_SETCURFOCUS is busted on WINE/ReactOS for single line (non-TCS_BUTTONS) tabs...
	// We need it on Windows for multi-line tabs or multiple tabs can appear pressed.
	if (::GetWindowLongPtr(_hSelf, GWL_STYLE) & TCS_BUTTONS)
	{
		::SendMessage(_hSelf, TCM_SETCURFOCUS, index, 0);
	}

AngryGamer:
@balagansky
> Removing these blocks of code seems to avoid the issue, but that alone may not be the right fix:
> 
> ```
> // TCM_SETCURFOCUS is busted on WINE/ReactOS for single line (non-TCS_BUTTONS) tabs...
> // We need it on Windows for multi-line tabs or multiple tabs can appear pressed.
> if (::GetWindowLongPtr(_hSelf, GWL_STYLE) & TCS_BUTTONS)
> {
> 	::SendMessage(_hSelf, TCM_SETCURFOCUS, index, 0);
> }
> ```

This section of code shouldn't impact single line/row of tabs at all.

[...]

PR #6277, opened 2019-10-27
Fix for multi-line tabBar click activation #6274

PalashBansal:
If Multi-line TabBar is enabled, mouse right click for tab selection still works fine without any change-
```
case WM_RBUTTONDOWN :	//rightclick selects tab aswell
        setActiveTab(nTab);
```
**Regarding Issue:**
I can repro. this issue very easily, I hope you too. If not, freely ask for help.
**Regarding PR:**
This is a quick fix kinda. I debugged a lot but couldn't find the root cause.
```
-> TabBar.cpp Line-602
-> Under case WM_LBUTTONDOWN :
-> ::CallWindowProc(_tabBarDefaultProc, hwnd, Message, wParam, lParam);
```
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/a156bf10529dad33714a73d825529d3ce5ff5a28/PowerEditor/src/WinControls/TabBar/TabBar.cpp#L602
This statement immediately switches tab, but only in single line tab mode, not multiline. Weird!

I added additional code snippet, which will be executed ONLY if end-user performs LDOWN while Multi-line is enabled. Not changing any previous functionality(LDOWN on Multi-line disabled still works same)
I have done rigorous testing from my side. It will be good if someone does the final testing.
Or, if someone has a better solution(especially after finding the root cause of problem), will be happy to know the correct approach and solution. :)

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -599,8 +599,20 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 				}
 			}
 
-            ::CallWindowProc(_tabBarDefaultProc, hwnd, Message, wParam, lParam);
-			int currentTabOn = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0));
+			if (::GetWindowLongPtr(_hSelf, GWL_STYLE) & TCS_BUTTONS)
+			{
+				int nTab = getTabIndexAt(LOWORD(lParam), HIWORD(lParam));
+				if (nTab != -1 && nTab != static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0)))
+				{
+					setActiveTab(nTab); // executes when Multi-line TabBar is enabled
+				}
+			}
+			else
+			{
+				int currentTabOn = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0));
+				notify(NM_CLICK, currentTabOn);
+			}
+			::CallWindowProc(_tabBarDefaultProc, hwnd, Message, wParam, lParam);
 
 			if (wParam == 2)
 				return TRUE;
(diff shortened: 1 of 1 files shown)


donho:
Which bug do you address to? Could you provide description, steps to reproduce and your debug info?

PalashBansal:
@donho 
**For issue-** **https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6274**

Here is my **debug info.**-
Notepad++ v7.8   (64-bit)
Build time : Oct  7 2019 - 00:55:55
Admin mode : ON
Local Conf mode : OFF
OS Version : 1803
OS Build : 17134.1069

I can repro. this issue on both win7 and win10 with latest NPP

If you see my file changes, I am also doing same fix which is suggested in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6273#issuecomment-546753080

**How to reproduce the issue:**
1. Keep your Settings>Preferences>General>TabBar>**Multi-line option unchecked.**
2. Keep many tabs opened, click to any tab(not current) and keep your **mouse left button click holded for a while, and then release**
3. In step2 above, you will observe that tab got **already** switched to new one, when you are clicking(LBUTTONDOWN) not releasing(LBUTTONUP) **which is expected behavior.**
**BUT-**
4. Keep your Settings>Preferences>General>TabBar>Multi-line option **checked.**
5. Keep many tabs opened, click to any tab(not current) and keep your mouse left button click holded for a while, and then release. (Note that while doing this, even if multi-line option is checked, there should be >=2 number of tab lines), keep your NPP window smaller or number of tabs opened higher.
**NOT this-**
image
**But should be this-**
image

6. In step5 above, you will observe that tab haven't switched to new one, when you are clicking(LBUTTONDOWN) but switched when releasing(LBUTTONUP)

Please reply back if usecase is not clear, or if still issue is not reproducible.

balagansky:
I think this PR fixes addresses the root cause: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/6278

PalashBansal:
> I think this PR addresses the root cause: #6278

I just looked into the changes in your PR
I don't see any additional **relevant** change in this PR- #6278
However, you are touching functionality of single line tabbar LBUTTON click which is already working perfectly.
Also, for multiline tabbar LBUTTON click, you are calling `notify` 2 times instead of 1. Note that setActiveTab already calls it, which is also the reason it is not mentioned in WM_RBUTTONDOWN

On a side note: You can suggest changes(if any) here instead of creating a duplicate PR.

balagansky:
If you have time, please try out that PR and see for yourself. I was also surprised that it seemed to fix the issue you were working on with this PR as that was _not_ the intent, but it was a pleasant side effect. For further info on that PR, check out the referenced issues.

As I wrote in that PR, it implements @AngryGamer's suggested changes, so I can only dare to claim partial credit at best. At a glance, I don't see a problem with the dual calls to notify() since the calls are notifying on a different code.

donho:
It's not a bug to me

PR #6278, opened 2019-10-28
Fix tab dragging issues related to switching off TCS_BUTTONS under multi-line tabs mode (per AngryGamer's suggestions)

balagansky:
Addresses the following:

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6274
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6273
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4339

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -586,6 +586,15 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 
 		case WM_LBUTTONDOWN :
 		{
+			if (::GetWindowLongPtr(_hSelf, GWL_STYLE) & TCS_BUTTONS)
+			{
+				int nTab = getTabIndexAt(LOWORD(lParam), HIWORD(lParam));
+				if (nTab != -1 && nTab != static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0)))
+				{
+					setActiveTab(nTab);
+				}
+			}
+
 			if (_drawTabCloseButton)
 			{
 				int xPos = LOWORD(lParam);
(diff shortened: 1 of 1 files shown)


balagansky:
Implements @AngryGamer's suggestion here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6273#issuecomment-546753080

PowerEditor/src/WinControls/TabBar/TabBar.cpp
```diff
@@ -643,16 +655,9 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 				}
 				else if (++_dragCount > 2)
 				{
-					int tabFocused = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURFOCUS, 0, 0));
 					int tabSelected = static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0));
```
AngryGamer: Just in case, put everything below it into a conditional block:

`if (tabSelected != -1)`

PowerEditor/src/WinControls/TabBar/TabBar.cpp
```diff
@@ -586,6 +586,18 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
...
+			if (::GetWindowLongPtr(_hSelf, GWL_STYLE) & TCS_BUTTONS)
+			{
+				int nTab = getTabIndexAt(LOWORD(lParam), HIWORD(lParam));
+				if (nTab != -1 && nTab != static_cast<int32_t>(::SendMessage(_hSelf, TCM_GETCURSEL, 0, 0)))
+				{
+					setActiveTab(nTab);
+				}
+			}
+
+			_dragRefPoint.x = LOWORD(lParam);
```
AngryGamer: 😋

PowerEditor/src/WinControls/TabBar/TabBar.cpp
balagansky: Sure.

balagansky:
Grr. I'm working in a different dev env than usual and getting confused. I have no idea where those extra lines even came from.

Will fix in a bit.

donho:
I can produce it now.
It's a regression which is inserted by 8342da3df8009f8d76d1d4990f3b5695eae88dd8 (#3552) to fix #3545. 
I don't like to add layers and layers of code to fix issue, and create more regression eventually.
I will just revert  8342da3df8009f8d76d1d4990f3b5695eae88dd8  to fix this regression.
Thank you for your heads up and your coding effort.

balagansky:
@donho, thanks for looking into it.

Although the commit you suggest to revert does trigger this problem, I don't think reverting it is the best option because that commit does fix a real issue that would otherwise regress.

This commit addresses both the original issue and the newly created one. Essentially this commit fixes the fix, while preserving the original fix.

balagansky:
Actually, reverting that commit _WILL NOT_ fix the problem (I verified by applying the change locally).

Reverting would remove one TCM_SETCURFOCUS, but leaves the other one, and does nothing to address the real problem of TCM_GETCURFOCUS being called even in single-line mode (not TCS_BUTTONS).

This PR addresses the problems directly. Please reconsider.

AngryGamer:
@donho I'm rendered speechless. 

-----------------------------------------------

**First off,** incorrect tab dragging wasn't the only symptom exhibited by this bug. 

**Secondly,** all you've done is "fixed" one way to trigger the incorrect tab dragging, but it also happens in other ways.

**Thirdly,** you've **reintroduced old bugs** and **caused new ones.**

All in all, your "fix" was a huge regression.

-----------------------------------------------

**tl;dr You didn't fix anything.** This PR _**is**_ the correct fix.

balagansky:
Thanks, @donho, for reconsidering and accepting.

Issue #7499, opened 2019-11-03
'Move to Recycle Bin' > select 'No' -- "Delete Fail" dialog seems unnecessary


PalashBansal:
### Description of the Issue
**On selecting 'No' option in 'Move to Recycle Bin' confirmation dialog, user can see that file is still there. So, "Delete file failed" dialog seems redundant.**
Moreover, its a move operation. Also, 'fail' signifies failing to do the intended operation. If a message is meant to be there- Better msg would be "Move file aborted".
Another issue is NPP going to background because of "Delete file failed" dialog. See gif-
![gif](https://user-images.githubusercontent.com/6822117/68085918-2dcf6380-fe6c-11e9-9c77-2c2594d248e3.gif)

### Steps to Reproduce the Issue
1. Keep a file opened in NPP. Open 'File' menu
2. Select 'Move to recycle bin'
3. Select 'No' in the confirmation dialog

### Debug Information
Debug info.-
Notepad++ v7.8.1   (64-bit)
Build time : Oct 27 2019 - 22:57:19
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : **Windows 7** Professional (64-bit) 
OS Build : 7601.23833
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

**Note** that NPP going to background problem is reproducible only in Win7(not Win10)

sasumner:
Your example shows this action being done on file "new 5".  Since it isn't possible to move an "unsaved" file to the recycle bin, probably that action should be disabled in the menus.

PalashBansal:
> Your example shows this action being done on file "new 5".

Actually i saved the file as "new 5"(default NPP given name itself). If you observe, file is saved and tabbar is not highlighted.
Sorry for misleading info, from now on i will raise issues with example having non default filenames.

> Since it isn't possible to move an "unsaved" file to the recycle bin, probably that action should be disabled in the menus.

With unsaved file, 'move to recycle bin' option is already disabled.

sasumner:
@PalashBansal Ah, okay...I definitely think that your example was "tricky". :-)

donho:
I cannot reproduce it on my windows 10.
I suppose it's the OS behaviour, therefore it's not bug in Notepad++

PalashBansal:
> I cannot reproduce it on my windows 10.
[...]

PR #7500, opened 2019-11-03
remove unnecessary delete fail msg

PalashBansal:
Resolves issue- 'Move to Recycle Bin' > select 'No' -- "Delete Fail" dialog seems unnecessary #7499


donho:
Cannot reproduce under Windows 10 - it's rather an OS issue.

sasumner:
I also cannot reproduce on Windows 10.

PR #7521, opened 2019-11-06
Add theme file extension for ini

xnumad:
"A .theme file is a .ini text file [...]." - https://docs.microsoft.com/en-us/windows/win32/controls/themesfileformat-overview


sasumner:
@xnumad What issue number does this go with?

donho:
@xnumad The concerning issue should be opened before to provide the detail of you expected behaviour before your PR.
Anyway, the extensions in langs.model.xml are usually universal. `theme` as extension is specific for MS, so it won't be accepted. 
You can always add it as customizable ext via Style Configurator dialog.

PR #7524, opened 2019-11-07
Fixed #7508 (kind of regression)

### LINKED ISSUE #7508: New Save Prompt Locks Minimized Window
### New Save Prompt Locks Minimized Window
If you have a minimized Notepad++ window open with unsaved changes, the new prompt behaves incorrectly. Tested on Windows 7 and Windows 10 - it's worse on 10.

1. Open a new window
2. Type text into the new window
3. Minimize the new window
4. Close the new window through the taskbar

### Expected Behavior
The window should restore itself and the prompt should show as a standard Modal prompt for the window.

### Actual Behavior
Nothing happens.
On Windows 7: You can click on the taskbar icon or preview and the window pops up, but is in Modal mode, and nothing can be clicked until you hit Escape.
On Windows 10: The window becomes locked in a minimized state. The only way to get it to pop up again is through the "Bring to front" feature in the Task Manager. Then, like Windows 7, it's in Modal mode and nothing can be clicked until you hit Escape.

PS: I'd also like to suggest hiding the "All" buttons for windows that have only one tab open, but that's sort of a separate issue.

SinghRajenM:
Fixed issue #7508.
This issue was introduced in commit 3f7956dc1bd16445f031a38bca8aa9e72806a408

**Root cause:** 
When main windows is minimized and user tries to close it from taskbar, then save confirmation dialog does not come front. Why? Reason is still unknown, but it seems either there is some problem in constructing modal dialog using `DialogBoxParam` or may be problem with Style/styleEX used in rc of `IDD_DOSAVEORNOTBOX` dialog.

But bringing main windows front (if minimized) and then showing save confirmation, resolves the reported issue. I have verified on Win10 x64.

### DIFF
--- a/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
+++ b/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
@@ -306,7 +306,11 @@ void DebugInfoDlg::doDialog()
 
 void DoSaveOrNotBox::doDialog(bool isRTL)
 {
-	
+	if (::IsIconic(_hParent))
+	{
+		::ShowWindow(_hParent, SW_RESTORE);
+	}
+
 	if (isRTL)
 	{
 		DLGTEMPLATE *pMyDlgTemplate = NULL;


PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -306,7 +306,11 @@ void DebugInfoDlg::doDialog()
 
 void DoSaveOrNotBox::doDialog(bool isRTL)
 {
-	
+	if (::IsIconic(_hParent))
```
donho: For unknown reason, ::IsIconic(_hParent) return always false under my Windows 10, even Notepad++ is iconized.
Have you tested your code under Windows 10 ?

SinghRajenM: >Have you tested your code under Windows 10 ?

Yes @donho , I have tested. I verified on Win10 x64.

donho:
@SinghRajenM 
Please consider the following code for closing Notepad++ via iconized zone:
```
		case IDM_SYSTRAYPOPUP_CLOSE:
		{
			_pPublicInterface->setIsPrelaunch(false);
			_pTrayIco->doTrayIcon(REMOVE);
			if (!::IsWindowVisible(_pPublicInterface->getHSelf()))
				::SendMessage(_pPublicInterface->getHSelf(), WM_CLOSE, 0,0);
		}
		break;
```
Notepad++ icon state has been removed by `_pTrayIco->doTrayIcon(REMOVE);`.
I don't know how the test `if (::IsIconic(_hParent))` can be true.

PR #7528, opened 2019-11-08
Fix GCC compile errors under Linux due to case sensitivity in #include

marcows:


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/98f4b1503ee191eae482069fd7e0bc11c5da178a

PR #7536, opened 2019-11-09
remove unnecessary delete fail msg

PalashBansal:
Resolves issue- "Delete Fail" dialog seems unnecessary if user selects 'no' #7499

Solution is to only show dialog in case of exception.
To reproduce the issue- See my detailed comment- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/7499#issuecomment-551215145


PalashBansal:
@sasumner Can you please provide your feedback for this fix?

PR #7537, opened 2019-11-09
Add disable In-Selection when multiple selections active for Count, Replace-All and Mark-all commands

### LINKED ISSUE #7532: Count / ReplaceAll / MarkAll option "In Selection" should not be available with multiple selections
### Description of the Issue

The Count / ReplaceAll / MarkAll functions of the Find window allow a user to specify "In Selection" with multiple selections active.  Logically, this makes sense, but those functions only deal with a single "main" selection.  Until such a time when those functions are expanded to allow true multiple selection counting/replacing/marking, "In Selection" should not be allowed as a choice when multiple selections are active, because Count / ReplaceAll / MarkAll will affect a different amount of text than the user expects.

1. Make sure that `Multi-Editing Settings` are enabled in the Preferences' `Editing` area:
image

2. In a new Notepad++ tab, paste the following text:
```
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa

aaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
3. Use the mouse to select the first three lines. This is now the "main" selection, as it is the only selection:
image
4. Hold Ctrl and select (with the mouse) the last three lines into a second selection.  This is now the "main" selection as it is the most recently made selection:
image
5. Invoke the Mark dialog from the Search menu and set it up as follows:
image
6. Press the `Mark All` button and observe the results.

### Expected Behavior

I expected every `a` in *both* active selections to be marked.

### Actual Behavior
image
Only the `a`s on the last three lines (the current "main" selection) have been marked.  This is confusing to the user as there are clearly `a`'s in a selection that didn''t get marked.  This confusion could be avoided by disallowing `In selection` as a choice when multiple selections are active, much like it is disallowed when a rectangular selection is active.

### Debug Information

Notepad++ v7.8   (32-bit)
Path : C:\........\npp.7.8.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1709
OS Build : 16299.1451

### Notes

MarkAll was used as an example, but Count and ReplaceAll work similarly with multiple selections.

sasumner:
Fixes #7532.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -957,8 +957,8 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 						_options._isInSelection = true;
 					}
 				}
-				// Searching/replacing in column selection is not allowed 
-				if ((*_ppEditView)->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE)
+				// Searching/replacing in multiple selections or column selection is not allowed 
+				if (((*_ppEditView)->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE) || ((*_ppEditView)->execute(SCI_GETSELECTIONS) > 1))
 				{
 					checkVal = BST_UNCHECKED;
 					_options._isInSelection = false;


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/485b5aa7c79425b6c3df4b89189989185800f477

Issue #7543, opened 2019-11-10
Losing file changes on 'move to recycle bin'


PalashBansal:
### Description of the Issue
Condition1(Good): For a saved file on disk -> 'Move to Recycle Bin' button is enabled.
Condition2(Good): For a new document(not on disk) ->'Move to Recycle Bin' button is disabled(dimmed)

But, **Condition3(Erroneous):** For a document saved in disk but content is modified without saving changes -> 'Move to Recycle Bin' button is enabled. Clicking on 'Move to Recycle Bin' button **ignores all changes** and moves file from disk to recycle bin.

### Steps to Reproduce the Issue
1. Open a file from disk to NPP
2. Make some changes(but don't save)
3. 'Move to Recycle Bin' button is unfortunately enabled. Click it!
4. File moves to Recycle bin and gets closed from NPP.
5. Go to recycle bin and Restore file.
6. Open the file and observe that changes are lost.

### Expected Behavior
'Move to Recycle Bin' button should be disabled like Condition2

### Actual Behavior
'Move to Recycle Bin' button is enabled(allowed to click)

PalashBansal:
Fix is in PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7544

Ekopalypse:
I don't think this is expected behavior, at least not for me. If I want to delete a file, aka move to recycle bin, then I expect it to be deleted regardless if there are changes or not open and it is obvious that
if I delete a file which has "open changes" that I don't care about it, otherwise I wouldn't have deleted the file in first place and if it is by accident, there is still the confirmation dialog which need to be answered.

PalashBansal:
> I don't think this is expected behavior, at least not for me. If I want to delete a file, aka move to recycle bin, then I expect it to be deleted regardless if there are changes or not open and it is obvious that
> if I delete a file which has "open changes" that I don't care about it, otherwise I wouldn't have deleted the file in first place and if it is by accident, there is still the confirmation dialog which need to be answered.

Thats a good point. But what if user forgets to save file moving it to recycle bin to be restored later. Btw, confirmation dialog doesn't tell about unsaved changes, but only message about moving/closing file.

Ekopalypse:
Maybe I do have a totally different workflow then you do but for me this sounds strange.
Of course I could have, accidentally, hit the delete button but then I get the confirmation dialog and at
[...]

PR #7544, opened 2019-11-10
fix to also check for dirty file Issue #7543

PalashBansal:
**Fix for Issue- Losing file changes on 'move to recycle bin' #7543**

fix is to- not only check file existence in disk but **also check whether file has unsaved changes or not**.


donho:
Nothing is need to be fixed.

PR #7565, opened 2019-11-13
Fix incorrect use of std::remove_if().

bobsayshilol:
std::remove_if() will not erase elements from the underlying container, and since what remains outside of the new valid range is implementation defined we should be erasing those elements otherwise we are left with unspecified characters at the end of the URL.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -462,7 +462,8 @@ void Notepad_plus::command(int id)
 			if (nppGui._searchEngineChoice == nppGui.se_custom)
 			{
 				url = nppGui._searchEngineCustom;
-				remove_if(url.begin(), url.end(), _istspace);
+				auto newEnd = remove_if(url.begin(), url.end(), [](TCHAR c){ return _istspace(c); });
+				url.erase(newEnd, url.end());
 
 				auto httpPos = url.find(TEXT("http://"));
 				auto httpsPos = url.find(TEXT("https://"));


sasumner:

bobsayshilol:
So I read the contributions markup (hence why all of the changes are broken up into individual PRs rather than only a few "misc fixes" PRs) but I assumed that that document was only a guideline since PRs such as #7522 appear to have no related issue and PRs such as #6233 have more than one commit (and also doesn't have an issue?), so I assumed that creating an issue for each of these would only add noise as all of the issues would be copies of the descriptions provided in the PRs.

I also don't know what would need to be discussed since, as is the case for this PR, it's easy to verify from the code that it's a bug so having a change ready for a maintainer to be able to test rather than having them ask how best to fix it seemed like a more appropriate and simpler action than discussing it.

sasumner:
#7522 is a translation update which is sort of an exception to the "issue" rule.  #6233 should have had an "issue".  There are exceptions, but for suggested changes such as yours, to have the most likelihood of being considered, I suggest following the protocol.  But you can do what you want. :-)

bobsayshilol:
Fair enough. I thought I was making it easier for the maintainers by providing a change ready for landing but I guess not. I had 12 additional commits on top of the ones I've already submitted as PRs but if it has to be a 3 step process for each one before I can even submit the PR for review then that's too much hassle for me. I'll close all these off and someone else can file them as issues if they'd like. Sorry for the spam.

chcg:
#7956 Depends on this one.

PR #7566, opened 2019-11-13
Balance unbalanced GlobalLock/Unlock() pairs

bobsayshilol:
The commit messages explain the reason for both, but as mentioned in the title the `GlobalLock()`/`GlobalUnlock()` pairs needed balancing to avoid leaks.

Note that the double call to `GlobalUnlock()` was chosen over casting the result of `GlobalLock()` to two separate types to avoid a strict-aliasing violation.


sasumner:

bobsayshilol:

chcg:
See https://github.com/Microsoft/wil/wiki/RAII-resource-wrappers. unique_ptr usage on GlobalLock via:

`unique_hglobal_locked`

PR #7569, opened 2019-11-13
Remove unnecessary macro checks

bobsayshilol:
The macro `SCINTILLA_EDIT_VIEW_H` is never defined and the header it's guarding already has a `#pragma once` inside it so there's no need to check for it before including the header.


sasumner:

bobsayshilol:

PR #7570, opened 2019-11-13
Remove some unnecessary string constructions

bobsayshilol:
The callees already take the parameters by reference so there's no need to construct temporaries in the caller.


sasumner:

bobsayshilol:

PR #7613, opened 2019-11-18
Python builtins highlighting

PavelBlend:
Added highlighting to the built-in python library.
Changed the backlight so that it became correct for python 3.
Namely:

- Added purple highlighting for all built-in functions, constants and exceptions.

- now print is not a statement, but a function. Accordingly, this function will be highlighted in purple (instead of blue)

- exec is now also highlighted in purple. Since in python 3 this operator has become a function.

- The same goes for the inline constants True, False, and None.

But there are problems.

standard functions will always be highlighted, regardless of whether these names are methods or not:

python-syntax-test


xylographe:
Perhaps, themes should be adjusted too?

Issue #7618, opened 2019-11-19
Simple exclude files in Find in Files filter


alef162:
### Exclude files in Find in Files filter

When searching for a text in multiple files, would be useful to exclude some file types, like binary files, or not relevant files.

For example using this filter: 
\*.\* -\*.exe -\*.obj -\*.log

This should be a simple implementation, as noted in issue #2453

donho:
Is such notation standard?

sasumner:
I've seen it used with command-line tools.

No way to use it with filespecs where you need to use a space, but this is also a limitation of the existing syntax. :-)

alef162:
The current syntax uses space as a separator.  So this is a current limitation I didn't want to fix in this pull request.
I've seen the "-" syntax in other editors.

sasumner:
>The current syntax uses space as a separator.

I implied that rather than saying it.  Thanks for filling in the gap.

>I've seen the "-" syntax in other editors

Good to cite specific examples if you've got them.  I didn't cite specific command-line tools because I couldn't think of any that I've used recently that support it.

alef162:
Sublime uses similar syntax: 
https://forum.sublimetext.com/t/exclude-files-directories-from-find-in-files/825

xylographe:
Git uses `!` before a glob pattern in `.gitignore`.

donho:
Why not 2 fields for filter : include & exclude ?
What do you guys think?

xylographe:
One field might sometimes be easier to comprehend, for example,
```
*.abc !x*.abc
```
Reading from left to right: include files with extension _abc_, unless the file name starts with an _x_.

Or, more realistic:
```
*.cs !*.Designer.cs
```
Include C# source files, unless auto-generated by VS.

donho:
Humm...
Between `*.* -*.exe -*.obj -*.log` and `*.* !*.exe !*.obj !*.log`, I prefer the 2nd.

alef162:
I created a new pull request with your recommendation.
 #7625 
You can decide what to merge.

bahusoid:
Both `-` and `!` are valid in filenames. Ideally it should be something not allowed like `|`.

sasumner:
@bahusoid Ideally, yes, but `!` (or `-`) is rarely seen (at least in my experience) as the first character of a filename.  It seems the goal here is to have something mnemonic, and `!` (or `-`) achieves that.

alef162:
[...]

Issue #7262, opened 2019-10-30
中国自强


penghubingzhou:
### Steps to Reproduce the Issue

1. 
2. 
3.

PR #7619, opened 2019-11-19
#7618 Simple exclude files in "Find in Files" filter

alef162:
#7618 Simple exclude files in "Find in Files" filter
Filter usage example:
\*.\* -\*.exe -\*.obj -\*.log

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -973,12 +973,21 @@ bool str2Clipboard(const generic_string &str2cpy, HWND hwnd)
 
 bool matchInList(const TCHAR *fileName, const std::vector<generic_string> & patterns)
 {
+	bool is_matched = false;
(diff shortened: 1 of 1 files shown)


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -973,12 +973,21 @@ bool str2Clipboard(const generic_string &str2cpy, HWND hwnd)
...
 	{
+		if (patterns[i].length() > 1 && patterns[i][0] == '-') 
+		{
+			if (PathMatchSpec(fileName, patterns[i].c_str() + 1))
+				return false;
+
+			continue;
+		} 
+
 		if (PathMatchSpec(fileName, patterns[i].c_str()))
```
bahusoid: No need to match pattern if it's already matched.
```suggestion
		if (is_matched == false && PathMatchSpec(fileName, patterns[i].c_str()))
```

alef162:
Continue work on #7262

PR #7625, opened 2019-11-20
#7618 Simple exclude files in "Find in Files" filter using "!"

alef162:
Filter usage example:
\*.\* !\*.exe !\*.obj !\*.log


alef162:
Issue with github.com connection:

git clone -q https://github.com/notepad-plus-plus/notepad-plus-plus.git C:\projects\notepad-plus-plus
fatal: unable to access 'https://github.com/notepad-plus-plus/notepad-plus-plus.git/': Failed to connect to github.com port 443: Timed out

Will try new pull request.

PR #7648, opened 2019-11-23
Improve Japanese localizations

Phroneris:
- Add access keys to `DoSaveOrNot` buttons
- Fix the second sentence of `find-status-replaced-next-not-found` lacking a "next" nuance


Phroneris:
Sorry, this PR is bad.
Maybe because this brunch is secondhand, meaning made not from a plain original `notepad-plus-plus:master` but from mine used for my old PR, needless commits and review comments of those days are included here.

Closed.

Issue #5937, opened 2019-07-22
[Feature Request] Enable second SQL keyword list


CurbedLarry:
### Description
Scintilla supports multiple keyword lists in its SQL lexer, however Notepad++ only passes one list. This means all SQL keywords have the same style. Ideally data types would be highlighted in a different style.

### Changes
I believe the changes required **might** be as simple as follow:

**stylers.model.xml**
Add:
`<WordsStyle name="KEYWORD2" styleID="16" fgColor="8000FF" bgColor="FFFFFF" fontName="" fontStyle="1" fontSize="" keywordClass="instre2" />`

**langs.model.xml**
Move data type keywords to a second keyword row:
`<Keywords name="instre2">bigint binary bit ... varchar xml</Keywords>`

**ScintillaEditView.h**
Edit: `setLexer(SCLEX_SQL, L_SQL, LIST_0);`
To: `setLexer(SCLEX_SQL, L_SQL, LIST_0 | LIST_1);`

### Debug Information
Notepad++ v7.7.1 (32-bit)
OS : Windows 10 (64-bit)

xylographe:
Perhaps, the themes (`PowerEditor\installer\themes\*.xml`) should be adjusted too?

Ekopalypse:
The lexer, currently, supports up to [8 wordlist](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/lexers/LexSQL.cxx#L263), why restricting to two?

CurbedLarry:
> Perhaps, the themes (`PowerEditor\installer\themes\*.xml`) should be adjusted too?

Agree, all those theme files could be updated too.

> The lexer, currently, supports up to [8 wordlist](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/lexers/LexSQL.cxx#L263), why restricting to two?

The 3rd and 4th keyword lists are aimed at Oracle syntax. I don't know if they could be skipped? I would just be happy give data types a different colour.  Another colour for functions would be nice but trying to limit scope of the change.

Ekopalypse:
I don't think that it must be skipped. If the 3rd + 4th list doesn't contain any words it doesn't trigger anything. Would you mind giving a short example and explanation which parts should be colored how then I would give it a try. I don't have any SQL skills. But if you say you want to keep your changes as you've posted, I'm fine with it as well :-)

CurbedLarry:
Hi,

[...]

PR #7693, opened 2019-12-04
Enable two more SQL keyword list

xylographe:
KEYWORD2 – data types, highlighted in a different style.
USER1 – user-defined keywords, empty list by default.

Implements #5937. See https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5937#issuecomment-514519313 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5937#issuecomment-514572374 in particular.
Adds previously missing SQL keywords as well.

**Before:**

**After:** (with "user1" as a user-defined keyword)


donho:
Red? Could they be other color than red (for example purple)?
My reason  is red leads ppl think of error.

xylographe:
Purple (#800080) is fine by me. What do you think, @CurbedLarry?

SQL-after

donho:
@xylographe it's perfect to me.
Ping me when you merge your modif in this branch.

CurbedLarry:
> Purple (#800080 [[Purple color codes](https://www.rapidtables.com/web/color/purple-color.html)]) is fine by me. What do you think, @CurbedLarry?

Would only say that some darker themes already use pink/purple for the number style (eg Plastic Code Wrap).  Perhaps in those cases another colour can be default?

xylographe:
@donho
Modifications applied.

Red was used as the KEYWORD2 foregroundcolour in 5 themes:
```
  #FF0000 (Red) ==> #800080 (Purple    )   [Default    ]
  #FF0000 (Red) ==> #600080 (Purple-ish)   [Hello Kitty]
  #FF0000 (Red) ==> #A00090 (Purple-ish)   [Navajo     ]
  #FF0000 (Red) ==> #007FFF (Azure     )   [Obsidian   ]
  #FF0000 (Red) ==> #008060 (Teal-ish  )   [Khaki      ]
```

Issue #6194, opened 2019-10-03
regex replace is reapplied to its output on each line until matching stops


bughit:
### Description of the Issue
matching is supposed to happen on the original text, not recursively on the result of the replacement

```
aaa
aaa
```

A replacement in the above of "^a" => "" should match and remove the first "a" in every line.

Instead it removes all "a" chars, because it keeps rerunning the match/replace on the output of the previous match/replace until it stops matching (all "a" chars are gone)

### Debug Information
Notepad++ v7.7.1 (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)

bughit:
@donho

sasumner:
is the `@ donho` supposted to do something special?

SusilRamarao:
Hi can you give more info like replicating the issue? Thanks

kyleballing:
Escalate Please

Example Text

> foo, item1, item2, item3
> 643!@%^ bar, item1, item2, item3
> gar  6 asdf bly, item1, item2, item3
> , item1, item2, item3
>        , item1, item2, item3

Suppose I want to remove all text prior to the comma separated list (where Item1 could be anything). 
Using a regular expression "^.*?," I get the following output.

> item3
> item3
> item3
> item3
> item3

Replace All removes all the leading garbage, but then starts again at the beginning of the document and removes item1, then it starts again and removes item2, leaving only item 3.

I would expect Replace All, to replace all matches in the document as it exists before replacement only.

I can see how recursively replacing could be handy, but in most text editors and regex searches, this is not the default and the accepted behavior is to only iterate once.

Notepad++ v7.8.1   (32-bit)
Build time : Oct 27 2019 - 22:46:07
Path : D:\Users\kyleba\Desktop\VanillaNotepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1803
OS Build : 17134.1130
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
@kyleballing 's comment doesn't really add any new information, aside from the fact that he would like it addressed. :-)

>but then starts again at the beginning of the document 

[...]

PR #7702, opened 2019-12-05
backward direction checkbox always enabled, even in regular exp. search mode

### LINKED ISSUE #3640: Backward-direction checkbox enabled with Reg Exp search mode (at N++ startup)
### Description of the Issue

### Expected Behavior

### Actual Behavior
See point 4 above.

### Debug Information
Notepad++ v7.5   (32-bit)
Admin mode : OFF
Local Conf mode : ON
OS : Windows 7 (64-bit)

image

sasumner:
Fixes #3640

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -332,9 +332,6 @@ void FindReplaceDlg::fillFindHistory()
 		::SendDlgItemMessage(_hSelf, IDWHOLEWORD, BM_SETCHECK, BST_UNCHECKED, 0);
 		::EnableWindow(::GetDlgItem(_hSelf, IDWHOLEWORD), (BOOL)false);
 
-		//regex upward search is disable in v6.3 due to a regression
-		//::EnableWindow(::GetDlgItem(_hSelf, IDC_FINDPREV), (BOOL)false);
-		
 		// If the search mode from history is regExp then enable the checkbox (. matches newline)
 		::EnableWindow(GetDlgItem(_hSelf, IDREDOTMATCHNL), true);
 	}
@@ -1400,16 +1397,10 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 						//regex doesn't allow whole word
 						_options._isWholeWord = false;
 						::SendDlgItemMessage(_hSelf, IDWHOLEWORD, BM_SETCHECK, _options._isWholeWord?BST_CHECKED:BST_UNCHECKED, 0);
-
-						//regex upward search is disable in v6.3 due to a regression
-						::SendDlgItemMessage(_hSelf, IDC_BACKWARDDIRECTION, BM_SETCHECK, BST_UNCHECKED, 0);
-						_options._whichDirection = DIR_DOWN;
 					}
 
 					::EnableWindow(::GetDlgItem(_hSelf, IDWHOLEWORD), (BOOL)!isRegex);
 
-					//regex upward search is disable in v6.3 due to a regression
-					::EnableWindow(::GetDlgItem(_hSelf, IDC_BACKWARDDIRECTION), (BOOL)!isRegex);
 					return TRUE; }
 
 				case IDWRAP :


chcg:
@sasumner  Do you known what the regression in 6.3 was? Might this be fixed or modified by the updates of scintilla or boost?

sasumner:
> @sasumner Do you known what the regression in 6.3 was? Might this be fixed or modified by the updates of scintilla or boost?

I don't.  6.3 was before my time.  I started with 6.7.5 four and a half years ago.  I have tried to elicit conversation on what this was a few times over the years, both here and on the Community, and anyone who might have known what the regression was either did not see it or chose not to respond.  Do you have any objection to this change?  It is fairly clear that the thing that is disabled in this one manner is allowed in at least two alternate ways of invoking, so this merely unifies the user interface...

dinkumoil:
This logic is strange.

There is an inconsistency in GUI behaviour that allows users in some cases to perform actions that produce invalid results.

Instead of fixing the GUI inconsistencies in order to allow only valid user actions, you suggest a change that makes the GUI consistent in a way that users are **always** allowed to perform actions that produce invalid results.

That's not making things better, it's declaring a bug as a world-standard.

sasumner:
@dinkumoil 

You posted the same comment here and in issue #3640 .  No sense in that or in posting the same reply here; please see my response in the comments for that issue.

dinkumoil:
@sasumner 

Originally I wanted to leave my comment only here at the PR thread, but at first I accidentally left it at the issue thread. So I repeated my comment here because IMO the PR is more important than the issue.

To future readers: Please consider [this conversation](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3640) as well.

xylographe:
There seems to be some confusion about backward regex search. It is correct that for some regex patterns the number of matches differs between forward and backward search. However, the conclusion that therefore backward regex search produces _invalid results_  is incorrect. The yielded results are consistent and logically valid. I don't see a compelling reason to prohibit backward regex search.

Hence, I go along with the uncomplicated solution to a minor issue, as provided by this PR.

donho:
@xylographe 
> the conclusion that therefore backward regex search produces _invalid results_ is incorrect.

Yes, it doesn't work for old version of scintilla. However, it could be fine now with the new version.
Have you any example that regExpr backward search doesn't work?

sasumner:
@donho 

I provide a good example regex search scenario in the issue for this PR.  Direct link: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3640#issuecomment-562273006

I don't think it has to do with Scintilla.  "Working" is sometimes a user interpretation.  Sometimes the results of a backward regex search may be unexpected, at first.  But if you take the time to understand it, it is consistent.  Thus, there isn't a "problem" with it, it simply sometimes gives different results than a forward regex search.

But in the end it is a decision for @donho backed by group voices here.  If no decision is made "soon", I will withdraw the PR and look at alternate "solutions".

xylographe:
I guess, the alternative would be to always disable (grey) "backward direction" when "regular expression" is selected. Personally, I won't miss it, but I think, it's a tad patronizing.

dinkumoil:
For sure, disabling backward direction regex search is indeed some kind of patronizing. BUT, all commenters involved in this PR have experience in giving support in the user forum. Wouldn't you agree that it is sometimes better to slightly patronize users than getting again and again the same user complaints and support requests why this or that feature doesn't work like expected?

A great example of wrong user expectations is [this thread](https://community.notepad-plus-plus.org/topic/18536/how-to-center-text) in the community forum. In short: The user complained about the lack of a "center text" feature in Notepad++ though allegedly there is a "align left" and a "align right" feature. Though he has been told that this kind of features is very uncommon for plaintext editors but are typical features of word processing software, he insisted that "center text" **should** be included in Notepad++ because the program is for text processing and additionally refered to his 35 years of experience working with computers (in marketing and advertisement business). Finally he tried to make the commenters feel ashamed by posting the advice he was given in another forum where someone told him to use the LTR/RTL writing feature to achive some kind of "align left"/"align right" behaviour. That clearly shows that he even didn't understand what he was doing when using those features.

I don't claim that backward regex search gives results which are wrong in a pure technical meaning. If one thinks about the results, he might have to admit that they are correct. Nevertheless, in most cases the results won't be that what a user has expected. Moreover, I bet they are even not helpful in most cases. And I bet there will be users complainig about the results.

So, why allow a feature which is not helpful and likely will create annoying support requests? Of course, it is a quick fix for the current inconsistency but it is "taking the lazy way" instead of doing it right (what I mean with "doing it right" I have elaborated about in [my posting in the issue thread](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3640#issuecomment-563196863)). And to say it again: I'm not a C++ dev, thus I can not do these changes by myself.

sasumner:
As an additional data point, guy038 had some private email conversation with me on this issue (not sure why he didn't just comment here, but maybe he's not a github member and has his reasons for that).  Guy was more in favor of disabling backward regex searches entirely, rather than full-bore enabling.

>...getting again and again the same user complaints and support requests why this or that feature doesn't work like expected?

I think it's a catch-22 regarding user complaints with this.  It's either (a) _Why do backward regex searches work strangely in certain cases?_ or (b) _Why is backward search disabled when using regex?_.  Empirically, since I've been a Notepad++ user (version 6.7.5) backward regex search has been possible (methods not involving the direction checkbox), and I've seen exactly ZERO user complaints for (a).  We don't know the user reaction for (b), to a full disallow.  But, if we just consider the _Backward direction_ checkbox being disabled for regex search (current N++ behavior) only, I can recall exactly ONE user complaint...or maybe it wasn't a complaint but just a question.  The difference between zero and one complaint/query is likely statistically negligible over 4+ years; maybe the most conclusive thing we can pull from this is that user complaints probably would not be a problem.

With my _user_ hat on, rather than _developer_ hat, I know that I often walk back and forth among the matches in a search, and I don't remember any (non-contrived) time when the regex results were different enough to make me question the matches.  Not being able to move forward and backward through my hits, even in regex mode, would likely irk me every time I encounter it.  But that's just me as a _user_.

Maybe the best answer is to substantially change the search routines.  Instead of starting at the caret, all searches should cover the entire doc (BOF -> EOF) or maybe just BOF to caret and build a list of match positions within this space.  Then, code can look through the match list and see if the caret location satisfies.  Because the search to the "left" of the caret position will be covered by a forward search through that section of the doc, it is easy to determine the first, second, etc. backward moving result through the data.  This solves other problems as well, like the one posed in #6194.  So forward or backward searching would produce the same result, and there would be no need for disabling/disallowing anything.

dinkumoil:
> (a) Why do backward regex searches work strangely in certain cases?
> I've seen exactly ZERO user complaints for (a)

The reason could simply be that users didn't try to use menu entries, toolbar buttons or whatever to do a regex search in backward direction while the search dialog has been open. And even if a user has tried to search in backward direction, when confronted with weird results I guess he might have thought "Oh well, weird results, but the 'search backward' checkbox is disabled so using the menu entry for that is possible but it is likely not to be a desired user action".

>... and build a list of match positions within this space. ... Because the search to the "left" of the caret position will be covered by a forward search through that section of the doc, it is easy to determine the first, second, etc. backward moving result through the data.

This is a really great suggestion though it causes a lot more effort.

donho:
> I provide a good example regex search scenario in the issue for this PR. Direct link: #3640 (comment)

Indeed, still buggy in the current version of Scintilla.

Ekopalypse:
I agree with @dinkumoil about the assumption users might have. 
For example, if I start with an forward search to find something and then overshoot my target and want to go back, I want to find the same result as I did before.
Technically it might be right to return a different match, but, as said, I think most users won't expect it.
From my point of view there are only two options:
1. disable regex backward search or
2. In case of a backward search do an forward search always and keep a list of the matches to get backward from current caret position

I did a test with the given example from @sasumner  with these editors
Notepad++
CudaText, Geany  ( those don't allow backward searches )
and Eclispe, VSCode, Sublime, Atom, Notepad2, PSPad, WingWare, PyCharm (these do forward searches and jump back to those matches)

donho:
@sasumner 
> I don't think it has to do with Scintilla. "Working" is sometimes a user interpretation. Sometimes the results of a backward regex search may be unexpected, at first. But if you take the time to understand it, it is consistent. Thus, there isn't a "problem" with it, it simply sometimes gives different results than a forward regex search.

I've got your point. But Notepad++ users don't have the time to understand such logic. They just want a tool do things as they expected. 

Let's look at alternate "solutions" -> please grey out also arrow down ▼ button in reg expr mode.

Issue #4881, opened 2018-09-26
Scroll and cursor position lost


bernard01:
I edit a dozen of raw text files, with word wrap turned on. I switch between them back and forth using [Ctrl+Tab].

The files are small, each only 2 pages.

I expect that each file / tab maintains its scroll and cursor position. However that is not the case. Both are lost.

For some time, that is perhaps ok, and most users perhaps can't be bothered reporting this.

It gets really annoying and impossible to work with when a problem becomes complex and requires strong focus with a stable view.

The problem most often occurs at the bottom third of the page where the content becomes hidden because the page just scrolls back to the top.  I can reproduce the problem each time.

sasumner:
Possible duplicate of #4825 and probably others; please search open issues before creating new (better to add new or slightly different info to an open issue)

Issue #4825, opened 2018-09-05
Scrollbar changes positions swapping tabs with word wrap


error17:
### Description of the Issue

With word wrap on, and a document with lines that are split due to word wrap, with many lines so there is a scroll bar, if you swap to another tab and back, the scroll bar will be in a different position than where it was left.

### Steps to Reproduce the Issue

1. create a document with long lines and many lines.
2. scroll somewhere in the middle. Make sure the top is not on the starting line, but one of the split word-wrapped lines (best to have line number enabled so you can see starting lines).
3. swap tabs.

### Expected Behavior

The scroll bar should not move, should stay in the same position.

### Actual Behavior

The scroll bar will change where it is. If the top line is a new line, there is no issue; but if the top line is on a word-wrapped line, it will snap to starting line on tab switch, can be disorientating on very long lines that have been split into many word-wrapped lines.

vertigo220:
Came here to report this as well. It should ideally stay exactly where it was when switching, e.g. if it's showing lines 30-60, switching to another tab and back it should still be showing lines 30-60, not 24-54 or whatever. I'm constantly switching back and forth and this is constantly causing me to have scroll back to where I was. Otherwise, this is an amazing program, so thank you for it!

Antti29:
I think this was at one point fixed but a later update regressed it.

bersbersbers:
I agree: that earlier bug was #4571 (which itself was #2078, #2576, #264, #3261, #3570, #3015), and a current open duplicate is #4881.

xylographe:
The cause of this behaviour has been explained in [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4571#issuecomment-396406247).

It was fixed in commit 40192bda0ccbd18077abb02fac30292dfe8e62f4, and reverted in commit fdf954f714325b41a86b745f3e3525277ba17ca4, because the ‘cure’ caused [a serious permance issue](https://notepad-plus-plus.org/community/topic/16468/notepad-7-5-9-release/16).

bersbersbers:
Thanks for investigating! Would someone care to organize the heap of duplicates (see my previous post)? Because that list includes several open issues and even a closed, apparently fixed one (#4571).

Issue #3570, opened 2017-07-28
Scroll position is changed for very long wrapped lines


ygoe:
### Description of the Issue
When the user switches between file tabs, the scroll position of the file is not kept. So the user sees another part of the file than where they previously left the tab.

### Steps to Reproduce the Issue
1. Activate automatic line wrapping and scrolling beyond the last line
2. Open or create a file with a very long line that wraps over more than a screen page of display lines
3. Add a few more shorter lines
4. Scroll the view so that it begins with some of the (not all) wrapped lines and following complete lines are also visible
5. Open or create another file
6. Switch back to the other file

### Expected Behavior
The vertical scrolling should be where I left it, i. e. wrapped lines are at the top and the following lines are also visible.

### Actual Behavior
The scrolling is reset so that the beginning of the wrapped line is at the top (can be seen at the line numbers) and the following lines are not visible because they're too far down. The cursor position is not affected by this. So if the cursor was previously visible in one of the following lines, it now no longer is.

### Debug Information
Notepad++ v7.4.2   (32-bit)
Build time : Jun 18 2017 - 23:34:19
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll

fml2:
I can confirm this with NP++ v. 7.5.1 (official download) on Win XP.

Issue #2576, opened 2016-11-17
Window returns to incorrect location on tab change with long lines and wordwrap mode enabled


PH0lder:
### Description of the Issue
WIndow returns to incorrect location on tab change with long lines and wordwrap mode enabled.

I have View | Word Wrap enabled, and a few lines that are longer than the width of the window (i.e. are wrapped.)  When I switch to another editor tab and then come back, the cursor stays on the correct line, but sometimes it's not actually visible in the window because it is scrolled to the wrong place.  I assume it's not counting the line number correctly when there are long wrapped lines.

### Steps to Reproduce the Issue
1.  Have long lines that need word wrapping in a long document.
2.  Enable View | Word Wrap.
3.  Scroll to the bottom of the document.
4, Switch to another editing tab.
5. Switch back to the first tab with the long word wrapped document.

### Expected Behavior
When I come back to the window after switching tabs, the window should have stayed in the same state (location and scrolled to position) it was before I switched tabs.

### Actual Behavior
The cursor remains in the correct location, but the location moves, including not even being visible on screen.

### Debug Information
Notepad++ v7.2   (64-bit)
Build time : Nov  2 2016 - 11:27:56
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll 

This has occurred in older versions for some time, including in the 32bit version before I upgraded to the 64bit version.

ysrn:
Please fix this, it's making it impossible to work with word-wrapped files.

NicolBolas:
Is there any movement on this issue? Any at all? Is this a Scintella problem or a Nodepad++ problem?

fml2:
This issue is probably related to #3570.

zhuhang:
Any updates on this issue?

fml2:
Yes, this is better now but still not quite well. If the cursor is at the end of the file in an empty line (i.e. the last text line has a CR at the end), then, after switching the tab to another file and back, the cursor is still o the right line, but it is barely visible. The file got scrolled one line up.

The file has 2792 lines, the last one is empty.

sasumner:
@fml2 
Can you post a before and after screenshot of this?

fml2:
Sorry, I can't post screenshots with the old version of NP++ since I already deleted it. But I can post some imags with NP++ 7.5.9 (the latest version as of now).

[...]

PR #7713, opened 2019-12-08

### LINKED ISSUE #2078: Wrong screen position after tab switch (with word wrap)
I would like to report an annoying issue when using the word wrap feature. If you switch between documents while editing and you got the word wrap feature enabled, then the screen position is off.

Here is how to reproduce it:
Open a document with very long lines in it. Like this: <unwrapped.png>
Turn on word wrap and move to the end of the document. Like this <wrapped.png>
Then switch to another open document and switch back. And it will look like something like this: <wrapped_after_tabswitch.png>

Expected: The screen should go back to exactly where it was.
Result: The screen is off. The cursor is not visible on the screen. This doesn't happen when word wrap is off.
unwrapped
wrapped
wrapped_after_tabswitch

cmeriaux:
Fix scintilla restoring wrapped buffer issue: If you switch between tabs, text jump to another lines.
It's an old bug, fixed in v7.5.9, but it had performance regression so it was reverted in v6.0.0.

- First step: set selection, set anchor, set xoffset...
- Second step: once Scintilla has send the notification SCN_PAINTED, we can scroll several lines to set the first visible line to the correct wrapped line.

Fixes #2078, #2576, #3570, #4825, #4881

Here is a video of the test https://youtu.be/LPcjKSV6GT4 

`

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -1784,7 +1784,7 @@ void ScintillaEditView::saveCurrentPos()
 	//Save data so, that the current topline becomes visible again after restoring.
 	int32_t displayedLine = static_cast<int32_t>(execute(SCI_GETFIRSTVISIBLELINE));
 	int32_t docLine = static_cast<int32_t>(execute(SCI_DOCLINEFROMVISIBLE, displayedLine));		//linenumber of the line displayed in the top
-	//int offset = displayedLine - execute(SCI_VISIBLEFROMDOCLINE, docLine);		//use this to calc offset of wrap. If no wrap this should be zero
+	int32_t offset = displayedLine - static_cast<int32_t>(execute(SCI_VISIBLEFROMDOCLINE, docLine));		//use this to calc offset of wrap. If no wrap this should be zero
 
 	Buffer * buf = MainFileManager.getBufferByID(_currentBufferID);
 
@@ -1796,6 +1796,7 @@ void ScintillaEditView::saveCurrentPos()
 	pos._xOffset = static_cast<int>(execute(SCI_GETXOFFSET));
 	pos._selMode = static_cast<int32_t>(execute(SCI_GETSELECTIONMODE));
 	pos._scrollWidth = static_cast<int32_t>(execute(SCI_GETSCROLLWIDTH));
+	pos._offset = offset;
 
 	buf->setPosition(pos, this);
 }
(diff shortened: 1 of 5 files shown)


cmeriaux:
Fixes #2078, #2576, #3570, #4825, #4881

PowerEditor/src/ScitillaComponent/ScintillaEditView.h
```diff
@@ -632,6 +633,9 @@ friend class Finder;
     void setTabSettings(Lang *lang);
 	bool isWrapRestoreNeeded() const {return _wrapRestoreNeeded;};
 	void setWrapRestoreNeeded(bool isWrapRestoredNeeded) {_wrapRestoreNeeded = isWrapRestoredNeeded;};
+	bool isPositionRestoreNeeded() const {return _positionRestoreNeeded;};
+	void setPositionRestoreNeeded(bool isPositionRestoredNeeded) { _positionRestoreNeeded = isPositionRestoredNeeded;};
+
```
donho: Both getter and setter are not necessary here since `_positionRestoreNeeded` is only used internally (in the object).

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -1817,9 +1816,24 @@ void ScintillaEditView::restoreCurrentPos()
 		execute(SCI_SETXOFFSET, pos._xOffset);
 	}
 	execute(SCI_CHOOSECARETX); // choose current x position
-
 	int lineToShow = static_cast<int32_t>(execute(SCI_VISIBLEFROMDOCLINE, pos._firstVisibleLine));
-	scroll(0, lineToShow);
+	execute(SCI_SETFIRSTVISIBLELINE, lineToShow);
+	if (isWrap()){
```
donho: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#general

PowerEditor/src/ScitillaComponent/ScintillaEditView.h
cmeriaux: Ok ! I've removed the setter.

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
cmeriaux: oups !!! sorry. It's fixed

cmeriaux:
@donho  Thanks for the review.
I've fixed everythings and rebased my branch
Thanks

donho:
@cmeriaux 
It doesn't work to me:
https://www.youtube.com/watch?v=dG8D9LmWq-g

donho:
On the other view, it does work better, but still not working 100%:
https://www.youtube.com/watch?v=DBiyD3BxU5E

cmeriaux:
Arff ! what a nightmare to handle scintilla.
It seems that there is a corner case when the first line is a wrapped line, and the displayed portion is small.
Let's try again to fix it

donho:
On this video the right view scrolled at the bottom, switch off and switch back - it's scrolled up:
https://www.youtube.com/watch?v=DBiyD3BxU5E

It's also the current behaviour (bug) but I think your PR should address it as well.

cmeriaux:
@donho  new PR #7751 that fixes bugs you found.
main differences: now I wait several SCN_PAINTED notification before scrolling to the correct position.

Issue #5522, opened 2019-04-12
Function list in C++ does not detect functions with parameter comments containing parantheses (reopen of #4899)


kderh:
Description of the Issue
Steps to Reproduce the Issue

Add this code in a .cpp file. "Working" is detected by the function list but "NotWorking" is not, until you remove the ( in the comment. This style of documentation is common I think...
``
void Working(
int a // SomeText
)
{
}

void NotWorking(
int a // SomeText(
)
{
}
Expected Behavior

Both should display.
Actual Behavior

Some functions are not displayed.
Debug Information

V7.6.6 Win32
![grafik](https://user-images.githubusercontent.com/43849620/56069805-08222600-5d85-11e9-86cd-09b30487fe30.png)

Firecharmlily:
Have narrowed down possible reasons being connected to functionparser and/or xml file tree

PR #7729, opened 2019-12-12
Fixed Function list in C++ does not detect functions with parameter comments containing parantheses (reopen of #4899) #5522

julian-robledo:
I added a class range to the C function definition due to the fact it wasn't capable of identifying a function with a comment within its parameters.  The class range just takes the preset function parameters and is pasted in a class range so it can more easily identify them.  The code also takes in only the function name and not the entire line before the parenthesis.

### DIFF
--- a/PowerEditor/src/functionList.xml
+++ b/PowerEditor/src/functionList.xml
@@ -176,6 +176,94 @@
 							|	(?s:\x27(?:[^\x27\x5C]|\x5C.)*\x27)             # String Literal - Single Quoted
 							"
 			>
+				<classRange
+					mainExpr    ="^[\t\x20]*(class|struct)[\t\x20]+\w+\s*(final)?\s*(:\s*(AAAAAAAAAAAAAAAAAAAAAAAAAAAA99)\s+\w+\s*)?\{"
+					openSymbole ="\{"
+					closeSymbole="\}"
+				>
+					<function
+						mainExpr="(?x)                                              # Utilize inline comments (see `RegEx - Pattern Modifiers`)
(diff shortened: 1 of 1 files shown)


sasumner:
Is there an _issue_ discussing this deficiency in the function list?  PRs aren't accepted without a detailed discussion in an issue first.  This is so developers considering the merge can see/test/understand the deficiency and the proposed solution easily.  Refer to the "contributing" guidelines.

julian-robledo:
Yes, there is: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5522

PowerEditor/src/functionList.xml
```diff
@@ -176,6 +176,94 @@
 							|	(?s:\x27(?:[^\x27\x5C]|\x5C.)*\x27)             # String Literal - Single Quoted
 							"
 			>
+				<classRange
+					mainExpr    ="^[\t\x20]*(class|struct)[\t\x20]+\w+\s*(final)?\s*(:\s*(AAAAAAAAAAAAAAAAAAAAAAAAAAAA99)\s+\w+\s*)?\{"
```
MAPJe71: `(AAAAAAAAAAAAAAAAAAAAAAAAAAAA99)` does not seem right to me.
Furthermore, you applied the changes to the wrong parser i.e. the `C` instead of `C++` parser.
Did you read the guidelines for contributing?

donho:
@julian-robledo 
Could you explain what "AAAAAAAAAAAAAAAAAAAAAAAAAAAA99" is for?

Issue #5056, opened 2018-11-25
Feature request:  Add "Save As" button between "Save" and "Save All" buttons.  Thanks.


somebadhat:
Feature request:  Add "Save As" button between "Save" and "Save All" buttons.  Thanks.

sasumner:
"Customize Toolbar" plugin will allow you to do this.  Unlikely to be changed in Notepad++ itself due to the existence of this plugin.  Probably best to CLOSE this issue.

dinkumoil:
I have to second @somebadhat  - there is no "Save **As**" button in Notepad++. Also the latest release of the _CustomizeToolbar_ plugin (v4.1) doesn't include such a button.

dinkumoil:
> is there any reason a "custom" ... button for SaveAs can't be made (by the end-user) in the CustomizeToolbar plugin?

Well, it is possible to create custom buttons with this plugin. But I found it is not working reliably.

Often the buttons are greyed out after starting Notepad++, but if you start it via Explorer context menu (Edit with Notepad++) the buttons work.

Furthermore, adding buttons depends on the UI language since buttons have to be connected to a menu item using a chain of the texts of all menu entries involved.

Finally the buttons have to be in a special BMP format (256 colors i.e. 8 bits per pixel). This format is not very common, the only program I know which can convert normal BMPs to 8 BPP is IrvanView.

But besides all that mentioned before - regardless that there is a plugin which let users define own buttons the feature request of @somebadhat is a valid one. To decline that would be the same like saying "You miss a feature in Notepad++? Oh, there is the Python plugin, you can write a script to get the functionality you want, no need to integrate it in Notepad++ itself".

Let the developers decide if someone wants to invest the effort for the requested button.

sasumner:
If I'd have remembered about SaveAs and the Customize Toolbar plugin...I would have approached it differently with my comments.  At the time of writing I thought it was pretty much plug-and-play, with maybe a tad bit of customization (setting up a button).  I guess it isn't, which is a shame because here we have a dedicated plugin to do something (that is likely never to get attention in Notepad++ itself) and it is apparently working poorly.  :(  And if it *were* working, this discussion would likely be over.  :-)

Sure, this issue can stay open.  I don't see it getting mainline attention, but that's a different thing.

[...]

PR #7731, opened 2019-12-13
[ENHANCEMENT] added save as button as per Issue#5056

Nandith-Narayan:
Implemented a Save-As button as per [Issue#5056](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5056).
image


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5056#issuecomment-565652795

Issue #6078, opened 2019-08-20
A bug for Notepad++


yunfan188:
I found a bug about auto-input. 
On the setting->Preference->auto-complete->auto-input, I select the (), [], {}, "", ''.
When I edit code, () -->(())，it can auto input, that`s OK, however, (1) --> (1())，it can`t work, rather then it becomes (1(), even if I press the “shift+( + )” keys, it still becomes (1(), and then I press "ctrl+z" keys, it can become the normal format (1()). The issuse has same behavior on other symbols, like [].

bib12ag:
I fixed that bug.
In the 'insertMatchedChars()' of 'Autocompletion.cpp', several lines have been commented out so that they can be completed in parentheses and brackets.
Just comment out only `if (isCharNextBlank || isInSandwich)` for each switch case ( '(', '[', '{' ).

ganego:
See also: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1038

ganego:
Holy crap, after four years you finally fixed it and I no longer have to compile Npp myself.
Thank you @donho

PR #7733, opened 2019-12-14
Fix a autocompletion bug in parentheses and brackets #6078

bib12ag:
In issue #6078 , the problem is that when insert '(', '[', '{' with letters in '()', '[]', '{}', they are not autocomplete despite setting auto-input.
([]), (()) {()}... this is not problem.
I want to write `void function(int arr[])`, but that is written as 'void function(int arr[ )'.
I have to write ']' manually.

(closetag switch caluse in function 'insertMatchedChars' of 'AutoCompletion.cpp')
```
case int('('):
if (matchedPairConf._doParentheses)
{
    if (isCharNextBlank || isInSandwich)
    {
	    matchedChars = ")";
	    _insertedMatchedChars.add(MatchedCharInserted(static_cast<char>(character), caretPos - 1));
    }
}
```
I wondered if I needed an `if (isCharNextBlank || isInSandwich)` clause.
So I comment out only `if (isCharNextBlank || isInSandwich)` for each switch case ( '(', '[', '{' ).

```
case int('('):
if (matchedPairConf._doParentheses)
{
    matchedChars = ")";
    _insertedMatchedChars.add(MatchedCharInserted(static_cast<char>(character), caretPos - 1));
}
```


sasumner:
Closing.  Too many irrelevant files included in PR.

Issue #7737, opened 2019-12-14
User-configurable command line interpreter


xylographe:
It would be nice if it would be possible to replace the invoked command line interpreter with something more user-friendly than the archaic `cmd.exe` without having to modify the source code. For example, by adding a _CommandLineInterpreter_ setting in `config.xml`.

donho:
What interpreter can user set other than cmd?

sasumner:
I might guess "Powershell"...?

donho:
Why not? But such request has never made (except this one). I wonder if it's needed such feature.

xylographe:
> What interpreter can user set other than cmd?

I have been happily using powershell for several years. At first, I replaced `TEXT("cmd")` with `TEXT("powershell")`, but later on I decided to make it more flexible with a configurable setting.

Other interpreters, better than cmd, would be zsh, bash, tcsh, or fish.

> But such request has never made

Perhaps, I'm the last surviving shell aficionado? :wink:

sasumner:
@donho My feeling this is a reasonable change.  Unless you want an explicit "Open...Powershell" to sit alongside "Open...CMD".

@xylographe said "I'm the last surviving shell aficionado?"  For me, my user base is still most-familiar with CMD, but it seems that Micro$oft is finally promoting something else (Powershell) with Windows 10, so it may change.

PR #65, opened 2015-05-14
x64 build

NN---:
#64 Work in progress on x64 build.

PR #7738, opened 2019-12-14
Add "commandLineInterpreter" setting in config.xml

xylographe:
Sets cli used to "Open Containing Folder in cmd".

Implements #7737

Allows user to set the command line interpreter that is invoked when “Open Containing Folder in cmd” (or “CMD here”) is selected. This setting is intentionally not accessible in the **Preferences** dialogue,
because changing this from _cmd_ (the default value) to something else will most likely require some preparation, like adjusting PATH in the Registry, or creating a hard link in a suitable directory, or even writing a small utility acting as a bridge between NPP and the targeted shell. Hence, I believe, it is wiser to treat this setting like the _dlgAlwaysVisible_ setting: documented in the user manual, but not available in **Preferences**.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -2076,7 +2076,6 @@ bool NppParameters::getSessionFromXmlTree(TiXmlDocument *pSessionDoc, Session *p
 					(childNode->ToElement())->Attribute(TEXT("endPos"), &position._endPos);
 					(childNode->ToElement())->Attribute(TEXT("selMode"), &position._selMode);
 					(childNode->ToElement())->Attribute(TEXT("scrollWidth"), &position._scrollWidth);
-
 					MapPosition mapPosition;
 					int32_t mapPosVal;
 					const TCHAR *mapPosStr = (childNode->ToElement())->Attribute(TEXT("mapFirstVisibleDisplayLine"), &mapPosVal);
@@ -5096,7 +5095,16 @@ void NppParameters::feedGUIParameters(TiXmlNode *node)
 			const TCHAR * optDocPeekOnMap = element->Attribute(TEXT("docPeekOnMap"));
 			if (optDocPeekOnMap)
 				_nppGUI._isDocPeekOnMap = (lstrcmp(optDocPeekOnMap, TEXT("yes")) == 0);
-
+		}
+		else if (!lstrcmp(nm, TEXT("commandLineInterpreter")))
+		{
+			TiXmlNode *node = childNode->FirstChild();
+			if (node)
+			{
+				const TCHAR *cli = node->Value();
+				if (cli && cli[0])
+					_nppGUI._commandLineInterpreter.assign(cli);
+			}
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


sasumner:
Build appears to be failing due to my #7730 

How can it have succeeded as part of the auto-check on 7730 and failed here??

Why does the auto-check build use ming?

xylographe:
@sasumner Only [mingw-w64_810_X64 build](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/29527207/job/xppmwbyoocb7ml20) fails. Look at commit 53452d96e0bf17f9f5eaaeec9c6beb2424f6d8cd, and you'll see a clickable red cross at the left of the commit message. I don't use mingw, so I won't be able to resolve the issue (probably, a missing `#include`).

donho:
@xylographe 
> Allows user to set the command line interpreter that is invoked when “Open Containing Folder in cmd” (or “CMD here”) is selected.

What interpreter can user set other than cmd?

xylographe:
Out of the box, powershell should work. But, if you want to invoke 64-bit posh from 32-bit NPP a small adjustment (hard link in %windir%) will be necessary. Shells like bash and zsh must be installed first, and to invoke them from NPP at least the PATH environment variable should be adjusted.

PowerEditor/src/Parameters.cpp
```diff
@@ -5096,7 +5096,16 @@ void NppParameters::feedGUIParameters(TiXmlNode *node)
 			const TCHAR * optDocPeekOnMap = element->Attribute(TEXT("docPeekOnMap"));
 			if (optDocPeekOnMap)
 				_nppGUI._isDocPeekOnMap = (lstrcmp(optDocPeekOnMap, TEXT("yes")) == 0);
-
+		}
+		else if (!lstrcmp(nm, TEXT("CommandLineInterpreter")))
```
donho: The naming should follow the convention.

xylographe: Name of setting changed from **`CommandLineInterpreter`** to **`commandLineInterpreter`** (also in commit message).

PowerEditor/src/Parameters.cpp
```diff
@@ -5861,6 +5870,13 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("onAnotherView"), _nppGUI._smartHiliteOnAnotherView ? TEXT("yes") : TEXT("no"));
 	}
 
+	// <GUIConfig name = "commandLineInterpreter">< / GUIConfig>
```
donho: > This setting is intentionally not accessible in the Preferences dialogue,
because changing this from cmd (the default value) to something else will most likely require some preparation, like adjusting PATH in the Registry, or creating a hard link in a suitable directory, or even writing a small utility acting as a bridge between NPP and the targeted shell. 

Good point. In this case, I think it'll be even more cautious to remove this block. So if user needs this feature, he/she adds it manually.

xylographe: Good idea. I had not thought about the possibility to make it a ‘read-only’ option.

Code block removed.

xylographe: @donho 
Oops, that doesn't work, because the new `config.xml` will no longer contain the _commandLineInterpreter_ setting. Perhaps, something like this:
```
	// <GUIConfig name = "commandLineInterpreter">< / GUIConfig>
	if (_nppGUI._commandLineInterpreter.compare(TEXT("cmd")))
	{
		TiXmlElement *GUIConfigElement = (newGUIRoot->InsertEndChild(TiXmlElement(TEXT("GUIConfig"))))->ToElement();
		GUIConfigElement->SetAttribute(TEXT("name"), TEXT("commandLineInterpreter"));
		GUIConfigElement->InsertEndChild(TiXmlText(_nppGUI._commandLineInterpreter.c_str()));
	}
```
Only add the _commandLineInterpreter_ setting if its value does not equal "cmd"?

sasumner: I like it better when the block is present even if it says "cmd".  Then, a power user that wants to change it doesn't have to look up the syntax, they just change "cmd" to what they want, and they are up and running with it.  Easy.  My 2c.

donho:
@xylographe 
I let you update Notepad++ User Manual:
https://github.com/notepad-plus-plus/npp-usermanual

xylographe:
> I let you update Notepad++ User Manual

Bien, je m'en occuperai.

donho:
Attention, tu ne devrais pas utiliser cette langue civilisée à la place de la langue de Shakespeare - puisque toi-même tu es pour la langue "standard" ;)

xylographe:
[Notepad++ User Manual PR \#65](https://github.com/notepad-plus-plus/npp-usermanual/pull/65)

PR #7739, opened 2019-12-14
Fix a autoCompletion close tag bug #6078

bib12ag:
Improvement #7733 

In issue #6078 , the problem is that when insert '(', '[', '{' with letters in '()', '[]', '{}', they are not autocomplete despite setting auto-input.
([]), (()) {()}... this is not problem.
I want to write `void function(int arr[])`, but that is written as 'void function(int arr[ )'.
I have to write ']' manually.

(closetag switch caluse in function 'insertMatchedChars' of 'AutoCompletion.cpp')
```
case int('('):
if (matchedPairConf._doParentheses)
{
    if (isCharNextBlank || isInSandwich)
    {
	    matchedChars = ")";
	    _insertedMatchedChars.add(MatchedCharInserted(static_cast<char>(character), caretPos - 1));
    }
}
```
I wondered if I needed an `if (isCharNextBlank || isInSandwich)` clause.
So I comment out only `if (isCharNextBlank || isInSandwich)` for each switch case ( '(', '[', '{' ).

```
case int('('):
if (matchedPairConf._doParentheses)
{
    matchedChars = ")";
    _insertedMatchedChars.add(MatchedCharInserted(static_cast<char>(character), caretPos - 1));
}
```

### DIFF
--- a/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScitillaComponent/AutoCompletion.cpp
@@ -582,34 +582,24 @@ void AutoCompletion::insertMatchedChars(int character, const MatchedPairConf & m
 		case int('('):
 			if (matchedPairConf._doParentheses)
 			{
-				if (isCharNextBlank || isInSandwich)
-
-				{
-					matchedChars = ")";
-					_insertedMatchedChars.add(MatchedCharInserted(static_cast<char>(character), caretPos - 1));
-				}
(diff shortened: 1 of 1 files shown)


donho:
The old algorithm is not perfect, but it prevent from insertion close symbol in any case.
For example, when a parenthesis is opened from the beginning of sentence, the close parenthesis symbol won't be inserted immediately, so users can decide to close the parenthesis where they want.

PR #2266, opened 2016-09-10
Added more functionalities to "Folder as workspace"

SinghRajenM:
Added more functionalities to "Folder as workspace"
1. Fold All
2. UnFold All

PR #7749, opened 2019-12-17
Add "Fold All" toolbar command in Folder As Workspace

donho:


xylographe:
Nice complement to the AIM button!

SinghRajenM:
A try was given to implement something similar (**not the fancy button, but functionality wise same**) exactly 3 years back. But it was rejected saying `I don't think this feature is useful for a lot of users.` https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2266

Now, suddenly it became useful. Interesting!!!

donho:
@SinghRajenM  

The "Select/show current editing file" feature makes this feature more value. And GUI part does it as well. A good feature is not just a feature, it's the whole solution.
BTW, it's not at all the same implementation - this one is more compact and more consistent.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c8b70dacd7a25642c336b560194ab99cb7279b18
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e0d995c1a6a045c8effb9f412325ba14f0d329d1

PR #7751, opened 2019-12-17

### LINKED ISSUE #2078: Wrong screen position after tab switch (with word wrap)

cmeriaux:
If you switch between tabs while wrap mode is enable, text jump to another lines.
It's an old bug, fixed in v7.5.9, but it had performance regression so it was reverted in v7.6.0.

- First step: set selection, set anchor, set xoffset, set first visible line...
- Second step: once Scintilla has send the notification SCN_PAINTED, 
  - check the first visible line is correct. If it's not, set again the first visible line and wait for another notification
   - if there is an offset to apply, check the first visible line 'wrap count' value (see scintilla doc https://www.scintilla.org/ScintillaDoc.html#SCI_WRAPCOUNT )
    if value differ from the saved one, it means scintilla is still updating the wrapped buffer. So wait for the next notification
  - finally the wrapped buffer seems updated and ready to be scrolled to the necessary line.

   there is a protection to avoid infinite loops. After trying to restore position 8 times, we abort the process of restoring. It can occures when the wrapped buffer changes its topology (example when you moved a wrapped buffer to the other view. Line lengths changes and exact wrapped postion can't be exactly restored.

Fixes #2078, #2576, #3570, #4825, #4881

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -946,6 +946,17 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			if (not notifyView)
 				return FALSE;
 
+			// Check if a restore position is needed. 
+			// Restoring a position must done after SCN_PAINTED notification so that it works in every circumstances (including wrapped large file)
+			if (_mainEditView.isPositionRestoreNeeded())
+			{
+				_mainEditView.restoreCurrentPosAfterPainted();
+			}
+			if (_subEditView.isPositionRestoreNeeded())
+			{
+				_subEditView.restoreCurrentPosAfterPainted();
+			}
+
 			// ViewMoveAtWrappingDisableFix: Disable wrapping messes up visible lines.
 			// Therefore save view position before in IDM_VIEW_WRAP and restore after SCN_PAINTED, as doc. says
 			if (_mainEditView.isWrapRestoreNeeded())
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -2076,6 +2076,8 @@ bool NppParameters::getSessionFromXmlTree(TiXmlDocument *pSessionDoc, Session *p
 					(childNode->ToElement())->Attribute(TEXT("endPos"), &position._endPos);
 					(childNode->ToElement())->Attribute(TEXT("selMode"), &position._selMode);
 					(childNode->ToElement())->Attribute(TEXT("scrollWidth"), &position._scrollWidth);
+					(childNode->ToElement())->Attribute(TEXT("offset"), &position._offset);
+					(childNode->ToElement())->Attribute(TEXT("wrapCount"), &position._wrapCount);
 					MapPosition mapPosition;
 					int32_t mapPosVal;
 					const TCHAR *mapPosStr = (childNode->ToElement())->Attribute(TEXT("mapFirstVisibleDisplayLine"), &mapPosVal);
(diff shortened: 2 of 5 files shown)


chcg:
@cmeriaux Is this change dependent to on the scintilla update done in the meantime or a new way to handle this old problem?

cmeriaux:
@chcg it's a new way to handle this old problem. 
Scite (scintilla for windows) has the same issue

donho:
Every new PR should have new branch name:
Could you recreate a new PR with a NEW branch please?

donho:
I copied  first part Step `1: From your project repository, check out a new branch and test the changes.` of "command line instructions", then paste in my git console:
```
donh@DESKTOP-5BODV9D MINGW64 /c/sources/notepad-plus-plus (master)
$ git checkout -b cmeriaux-restore_wrapped_buf master
fatal: A branch named 'cmeriaux-restore_wrapped_buf' already exists.

donh@DESKTOP-5BODV9D MINGW64 /c/sources/notepad-plus-plus (master)
$ git pull https://github.com/cmeriaux/notepad-plus-plus.git restore_wrapped_buf
remote: Enumerating objects: 9, done.
remote: Counting objects: 100% (9/9), done.
remote: Total 10 (delta 9), reused 9 (delta 9), pack-reused 1
Unpacking objects: 100% (10/10), done.
From https://github.com/cmeriaux/notepad-plus-plus
 * branch              restore_wrapped_buf -> FETCH_HEAD
Auto-merging PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
Auto-merging PowerEditor/src/Parameters.h
Auto-merging PowerEditor/src/Parameters.cpp
Auto-merging PowerEditor/src/NppNotification.cpp
gpg: AllowSetForegroundWindow(14932) failed: Access is denied.

Merge made by the 'recursive' strategy.
 PowerEditor/src/NppNotification.cpp                | 11 ++++
 PowerEditor/src/Parameters.cpp                     |  4 ++
 PowerEditor/src/Parameters.h                       |  2 +
 .../src/ScitillaComponent/ScintillaEditView.cpp    | 65 ++++++++++++++++++++--
 .../src/ScitillaComponent/ScintillaEditView.h      |  5 ++
 5 files changed, 82 insertions(+), 5 deletions(-)
```
Then `git log` gave the following result:
```
$ git log
commit b7531459c948c3843c6b41d0a7632d3b862a4e79 (HEAD -> master)
Merge: a1d048fd 4b8262dd
Author: Don HO <don.h@free.fr>
Date:   Thu Dec 26 13:12:00 2019 +0500

    Merge branch 'restore_wrapped_buf' of https://github.com/cmeriaux/notepad-pl                                                                                           us-plus

commit a1d048fdd9367d37dbe17134cb88e4e24f83c311 (origin/master, origin/HEAD, for                                                                                           k/master)
Author: Don HO <don.h@free.fr>
Date:   Wed Dec 25 20:03:47 2019 +0500

    Update localization files for a new option

    Move an old entry to an another section and add a new one for added new opti                                                                                           on in:
    7a1096de5b3618bc3e6611e6c25caac952d41ae2
```

@cmeriaux If you do remember that once we have lost your credit on your commit, I guess that's the cause.

cmeriaux:
@donho  Ok I've created a new PR with a new branch name.

For the record, delete my branch before checking out the new one may have worked. But I've never been a maintainer

git branch -D cmeriaux-restore_wrapped_buf
git checkout -b cmeriaux-restore_wrapped_buf master

donho:
I clean the branch around periodically. But I won't bother to clean the branch on every merge.
The rule is for the sake of the workflow. Hope you understand that.

PR #5152, opened 2018-12-18
Add zulu.xml

RoshanRathod0137:

PR #7767, opened 2019-12-22
Venetan localization

conky77:
new localization for Venetan language


chcg:
See https://en.wikipedia.org/wiki/Venetian_language, an Italian "dialect".
@conky77 It is not sufficient for a new language just to add the xml. See e.g. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/5152/files
There is also an entry necessary in:
PowerEditor/installer/nsisInclude/langs4Npp.nsh
and
PowerEditor/src/localizationString.h

conky77:
Done: I created two new Pull-requestes

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7770

PR #7770, opened 2019-12-23
Add Venetian localization [complete]

xylographe:
PR #7767 extended.

PR #7771, opened 2019-12-23
Venetan localization

conky77:


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7770/files

Issue #4571, opened 2018-06-08
Jump to another lines when you switch tabs


EgorAnatolievich:
### Description of the Issue

Sorry, my English is not well, but I'll try.
If you switch between tabs, where a lot of lines, text jump to another lines.

### Steps to Reproduce the Issue

1. Open two tabs with a lot of lines.
2. In tab_1 pagedown anywhere
3. In tab_2 pagedown anywhere
4. Switch between the tabs and focus lose in tab_1 or tab_2

Look the screenshots.
![1](https://user-images.githubusercontent.com/3139930/41167382-662b0762-6b4b-11e8-9d0c-5cb26ef3b69f.png)
![2](https://user-images.githubusercontent.com/3139930/41167383-664a034c-6b4b-11e8-8aa0-565eb98146a5.png)
![3](https://user-images.githubusercontent.com/3139930/41167384-666922cc-6b4b-11e8-96d8-10901d3b5710.png)

### Expected Behavior

When you switch between  the tabs, focus didn't lose. You always see lines what lines you saw before switch.

### Actual Behavior

When you switch between the tabs, lines are jump.

### Debug Information

Notepad++ v7.5.6 (64bit)
Build time: Mar 19 2018 - 00:23:17
Windows 10 Pro 64-bit Versions 1803 (OS Build 17134.48)

sasumner:
What plugins do you have installed?  Do `?` menu, choose `Debug Info`.  If I'm remembering correctly, this kind of issue always ends with a plugin causing it.

EgorAnatolievich:
Debug info:

Notepad++ v7.5.6   (64-bit)
Build time : Mar 19 2018 - 00:23:17
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll DSpellCheck.dll mimeTools.dll NppConverter.dll PluginManager.dll PythonScript.dll XMLTools.dll

sasumner:
I would try removing the plugins one at a time and see if behavior changes.  Remove the plugins that do not come standard with Notepad++ first, as that is where you are likely to find the problem (if it is to be found that way).

EgorAnatolievich:
Sorry for duplicate report. Very sad that this bug din't fix :(
Should I close my issue?

sasumner:
[...]

PR #964, opened 2015-09-24
Update korean.xml

CNBLACK:
Updated XML Contents, and Optimized

PR #7781, opened 2019-12-26

cmeriaux:
If you switch between tabs while wrap mode is enable, text jump to another lines.
It's an old bug, fixed in v7.5.9, but it had performance regression so it was reverted in v6.0.0.

- First step: set selection, set anchor, set xoffset...
- Second step: once Scintilla has send the notification SCN_PAINTED, we can scroll several lines to set the first visible line to the correct wrapped line.

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -946,6 +946,17 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			if (not notifyView)
 				return FALSE;
 
+			// Check if a restore position is needed. 
+			// Restoring a position must done after SCN_PAINTED notification so that it works in every circumstances (including wrapped large file)
+			if (_mainEditView.isPositionRestoreNeeded())
+			{
+				_mainEditView.restoreCurrentPosAfterPainted();
+			}
+			if (_subEditView.isPositionRestoreNeeded())
+			{
+				_subEditView.restoreCurrentPosAfterPainted();
+			}
+
 			// ViewMoveAtWrappingDisableFix: Disable wrapping messes up visible lines.
 			// Therefore save view position before in IDM_VIEW_WRAP and restore after SCN_PAINTED, as doc. says
 			if (_mainEditView.isWrapRestoreNeeded())
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -950,24 +950,24 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			// Restoring a position must done after SCN_PAINTED notification so that it works in every circumstances (including wrapped large file)
 			if (_mainEditView.isPositionRestoreNeeded())
 			{
-				_mainEditView.restoreCurrentPosAfterPainted();
+				_mainEditView.restoreCurrentPosPostStep();
 			}
 			if (_subEditView.isPositionRestoreNeeded())
 			{
-				_subEditView.restoreCurrentPosAfterPainted();
+				_subEditView.restoreCurrentPosPostStep();
 			}
 
 			// ViewMoveAtWrappingDisableFix: Disable wrapping messes up visible lines.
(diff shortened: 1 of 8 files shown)


PowerEditor/src/NppNotification.cpp
```diff
@@ -946,6 +946,17 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			if (not notifyView)
 				return FALSE;
 
+			// Check if a restore position is needed. 
+			// Restoring a position must done after SCN_PAINTED notification so that it works in every circumstances (including wrapped large file)
+			if (_mainEditView.isPositionRestoreNeeded())
+			{
+				_mainEditView.restoreCurrentPosAfterPainted();
```
donho: Here we do restoreCurrentPosAfterPainted() then we do restoreCurrentPos() (line #964) - it's quite confused for me.
Could you enlight me about this point please?

donho: > First step: set selection, set anchor, set xoffset...
> Second step: once Scintilla has send the notification SCN_PAINTED, we can scroll several lines to set the first visible line to the correct wrapped line.

So if I do understand, first step is `restoreCurrentPos()` (line number 964) to enable `positionRestoreNeeded`, then the SCN_PAINTED message will be sent again.
`restoreCurrentPosAfterPainted()` is the 2nd step ?

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -1825,9 +1826,63 @@ void ScintillaEditView::restoreCurrentPos()
...
+
+void ScintillaEditView::restoreCurrentPosAfterPainted()
+{
+	static int32_t restoreDone = 0;
+	Buffer * buf = MainFileManager.getBufferByID(_currentBufferID);
+	Position & pos = buf->getPosition(this);
+
+	++_restorePositionRetryCount;
+
+	if (_restorePositionRetryCount > 8)
```
donho: Why do we need to retry 7 times?

PowerEditor/src/NppNotification.cpp
cmeriaux: Oui, vous avez totalement raison. Juste une précision, scintilla peut envoyer plusieurs noticiation SCN_PAINTED avant que le buffer soit prés à êter afficher. cette seconde étape est donc itérée plusieurs fois dans un maximum de 7 itérations. 7 est un chiffre arbitraire que j'ai choisi. 2 est un minimum (trouvé avec des tests). Je ne connais pas de maximum.

Pour information, dans mon PR précédent, les tests que tu as fait étaient KO car cette seconde étape n'était pas itérable.

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
cmeriaux: See my previous comment

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -1825,9 +1826,63 @@ void ScintillaEditView::restoreCurrentPos()
...
+			_positionRestoreNeeded = false;
+		}
+	}
+	else
+	{
+		// Buffer position is correct, and there is no scroll to apply
+		_positionRestoreNeeded = false;
+	}
+
+	return;
```
donho: This return is not necessary.

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
donho: OK, I see. So please add comment on this magic number with your above explanation.

PowerEditor/src/NppNotification.cpp
donho: The name of function `restoreCurrentPosAfterPainted` is not explicit to me and bring some confusion.
Do we have another better naming for this function?

cmeriaux:
@donho  I've done the requested changes.
Thanks

cmeriaux:
Fixes #4571

PR #7783, opened 2019-12-26
Delete nppGpgPub.asc

duythangqn1102:


donho:
@duythangqn1102 
Could you provide the reason to remove nppGpgPub.asc ?

sasumner:
>Could you provide the reason to remove nppGpgPub.asc ?

If @duythangqn1102 was following the contributing protocol, this question would not have to be asked:  An issue would have been opened _FIRST_, discussing the reason.

PR #7815, opened 2020-01-05
Update polish.xml

knowyourway:
Fixed some typos


donho:
Could you add the following missing translation?

f44c357

cbd1e83

ab207db

PR #7885, opened 2020-01-22
Fix HashFrom(Files|Text)Dlg::doDialog()

### LINKED ISSUE #7884: Wrong language in SHA-256 hash dialogues
When the SHA-256 hash dialogues are opened more than once, the caption (window title) and button text (in hash from files), will be reset to English.

1. Select any language (not English) in Preferences. I'll use French for demonstration.
2. **Select Tools > SHA-256 > Generate...** (**Outils > SHA-256 > Générer...**).

SHA256A

3. Close (Fermer) the dialogue.
4. **Select Tools > SHA-256 > Generate...** (**Outils > SHA-256 > Générer...**).

SHA256B

Q.e.d.

xylographe:
Fixes #7884.

Four different dialogues are used,
```
_md5FromTextDlg
_md5FromFilesDlg
_sha2FromTextDlg
_sha2FromFilesDlg
```
Hence, there is no reason to change the text of captions and/or buttons after initialisation.
Removing the unnecessary, offending lines solves the problem.

### DIFF
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -181,18 +181,20 @@ void HashFromFilesDlg::setHashType(hashType hashType2set)
 void HashFromFilesDlg::doDialog(bool isRTL)
 {
 	if (!isCreated())
+	{
 		create(IDD_HASHFROMFILES_DLG, isRTL);
 
-	if (_ht == hash_sha256)
-	{
-		generic_string title = TEXT("Generate SHA-256 digest from files");
-		::SetWindowText(_hSelf, title.c_str());
+		if (_ht == hash_sha256)
+		{
+			generic_string title = TEXT("Generate SHA-256 digest from files");
+			::SetWindowText(_hSelf, title.c_str());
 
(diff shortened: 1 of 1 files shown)


donho:
Your PR create the following regression:
Stay in English, launch "Generate MD5 digest from file", close it. then launch "Generate SHA2 digest from file" - the dialog title is stuck on the title bar of the first launch.

xylographe:
Very strange,
1. Language = English
2. Launch "Generate MD5 digest from file"
3. Launch "Generate SHA-512 digest from file"

No problem,
SHA256

But, if I close the MD5 dialogue between (2) and (3), then the text does indeed
show MD5 instead of SHA-512. I'll look into it.

donho:
@xylographe  
For me, set title actions should be kept. And enhance them with action to retrieve the localization string.

BTW, SHA-512 is not implemented yet :)

xylographe:
@donho 
I learned something new today. :)  The **`_nativeLangSpeaker.changeDlgLang()`** method does absolutely nothing at all if the UI language is English. Fortunately, that makes the solution of this issue very simple: the English title/button text is changed — if necessary — after creation of the dialogue. If the UI language is not English, then the translated text will be set after **`doDialog()`** returns.

PR #7893, opened 2020-01-25
Update Swedish localization

xylographe:
It's not 7.8.3, but it's a step in the right direction: 75 missing items were added, only 3 left,
```
<Item id="4008" name="Please exit Notepad++ and relaunch Notepad++ in Administrator mode to use this feature."/>

<Item id="6348" name="Don't fill find field in Find dialog with selected word"/>

<GUpProxyConfNeedAdminMode title="Proxy Settings" message="Please relaunch Notepad++ in Admin mode to configure proxy."/>
```


donho:
So why don't you update it to 7.8.3, directly?

xylographe:
Because I can't. The  "[**timint** authored](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7893/commits)" says it all.

Issue #905, opened 2015-09-13
Find/replace dialog should remember its last position


ygoe:
The dialog normally appears in the middle of the editor window, which is most likely in the way. So I move it a bit to the side to see the document while interacting with the search function. This position is retained when I close and re-open that window. But the position is reset again after Notepad++ was restarted. So this window placement should be included in the session state and be restored after quitting and restarting Notepad++.

This is pretty annoying when restarting Notepad++ often.

xylographe:
Although I restart NPP only once a month on average, I would welcome this enhancement.

dg9ngf:
I only noticed because I was starting it quite often today during bug hunting. Normally I'd keep it open until the next patchday as well.

h-h-h-h:
> @xylographe:
I restart NPP only once a month on average

&nbsp;

> @dg9ngf:
Normally I'd keep it open until the next patchday as well.

And in between you constantly put Windows in hibernation, or not even that?

xylographe:
No, Win7 remains active 24/7. Usually 3‒5 weeks before reboot is necessary because it becomes too slow.

dg9ngf:
I can easily run my Windows 7 for several months, but Windows updates become too annyoing then so I give in and let them install. This definitely leads to a restart after more than a month.

h-h-h-h:
What a luck that Notepad++ is green:
> Notepad++ is trying to reduce the world carbon dioxide emissions. When using less CPU power, the PC can throttle down and reduce power consumption, resulting in a greener environment.
(https://notepad-plus-plus.org/)

MetaChuh:
we are closing this issue due to its age and multiple newer notepad++ versions since it was filed.

please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html).
if your issue persists, and it is still of relevance to you, please feel free to file a new issue.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
[...]

PR #7898, opened 2020-01-28
Make Find window remember its position across multiple runs

sasumner:
Implements #905

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -4511,6 +4511,39 @@ void NppParameters::feedGUIParameters(TiXmlNode *node)
 			if (val)
 				_nppGUI._isMaximized = (lstrcmp(val, TEXT("yes")) == 0);
 		}
+
+		else if (!lstrcmp(nm, TEXT("FindWindowPosition")))
+		{
+			RECT oldRect = _nppGUI._findWindowPos;
+			bool incomplete = true;
+			int i;
+
+			if (element->Attribute(TEXT("left"), &i))
+			{
+				_nppGUI._findWindowPos.left = i;
+
+				if (element->Attribute(TEXT("top"), &i))
+				{
+					_nppGUI._findWindowPos.top = i;
+
+					if (element->Attribute(TEXT("right"), &i))
+					{
+						_nppGUI._findWindowPos.right = i;
+
+						if (element->Attribute(TEXT("bottom"), &i))
+						{
+							_nppGUI._findWindowPos.bottom = i;
+							incomplete = false;
+						}
+					}
+				}
+			}
+			if (incomplete)
+			{
+				_nppGUI._findWindowPos = oldRect;
+			}
+		}
+
 		else if (!lstrcmp(nm, TEXT("NewDocDefaultSettings")))
 		{
 			int i;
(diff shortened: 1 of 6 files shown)


Ekopalypse:
Just tested the x64 appveyor build.
Yes, works like expected.
Remembered position after npp relaunch.
Centered dialog after monitor disconnect as well as after screen reconfiguration.

PowerEditor/src/Parameters.cpp
```diff
@@ -4511,6 +4511,39 @@ void NppParameters::feedGUIParameters(TiXmlNode *node)
...
+							incomplete = false;
+						}
+					}
+				}
+			}
+			if (incomplete)
+			{
+				_nppGUI._findWindowPos = oldRect;
+			}
+		}
```
xylographe: Perhaps, something like,
```C++
RECT newRect;
if (element->Attribute(TEXT("top"), &newRect.top) &&
    element->Attribute(TEXT("left"), &newRect.left) && 
    element->Attribute(TEXT("right"), &newRect.right) &&
    element->Attribute(TEXT("bottom"), &newRect.bottom))
{
    _nppGUI._findWindowPos = newRect;
}
```

sasumner: I suppose.  My intent was to be very much like the code that reads in the configuration settings for `_appPos` (the main window positioning setting).   If you'll notice, this new code is almost identical to that, except I chose *not* to use the variable name `fuckUp` like Don did (I chose to use the name `incomplete`).  Maybe Don's choice IS better in this case, for descriptiveness, but I don't have the moxie to put something like that into someone else's codebase, even when duplicating code. ;)

xylographe:
Excellent! Thank you, @sasumner.

PowerEditor/src/Parameters.cpp
donho: > except I chose not to use the variable name fuckUp like Don did

Did I? I thought I watch always my language (only C++) :)

However, I prefer @xylographe 's way - more compact and readable.

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
```diff
@@ -77,37 +77,111 @@ void StaticDialog::goToCenter()
...
+			candidateRc = getViewablePositionRect(testPositionRc);
+
+			if ((testPositionRc.left != candidateRc.left) || (testPositionRc.top != candidateRc.top))
+			{
+				::MoveWindow(_hSelf, candidateRc.left, candidateRc.top, 
+					candidateRc.right - candidateRc.left, candidateRc.bottom - candidateRc.top, TRUE);
+			}
+		}
+		else
+		{
```
donho: So this part could be replace by `getViewablePositionRect()` or I am missing somethings ?

PowerEditor/src/Parameters.cpp
sasumner: Ok, will change it to be the more compact version.  I think if I would have left the "fuck" variable in there, you would have preferred that version. :)

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
sasumner: The `display` function is currently called for many of the different child windows.  The intent with this code change was to allow this enhanced positioning code to be introduced ONLY for the Find window (to stay in context of #905 and not have to validate ALL windows that use it -- too large of a change).  Other windows that can get "lost" by having positions outside viewable space would need to be handled one at a time (separate future "issues"), but in theory could use this new call option to display().  Once all child windows have been evaluated/changed to use the new code, the code I've moved to the "else" clause could be removed.

donho: Makes sens. Let's keep it in this PR.

donho:
BTW, just add a new commit into this PR.

PowerEditor/src/Parameters.cpp
sasumner: A problem with attempting to change it to the more-compact version is that the RECT struct members are LONG and there is no facility in the xml element Attribute function for reading into that data type.  But, because int and long are the same size (on Windows, and in 32 or 64 bit), one can solve this problem via casting.  But it is ugly now, and I tend to favor the original code in this PR.  What do you guys think about it?:

```
RECT newRect;
if (element->Attribute(TEXT("top"), reinterpret_cast<int *>(&newRect.top)) &&
    element->Attribute(TEXT("left"), reinterpret_cast<int*>(&newRect.left)) &&
    element->Attribute(TEXT("right"), reinterpret_cast<int*>(&newRect.right)) &&
    element->Attribute(TEXT("bottom"), reinterpret_cast<int*>(&newRect.bottom)))
{
    _nppGUI._findWindowPos = newRect;
}

```

sasumner:
Oops, maybe I added my comment above when I should have added it here at the end.  See above where I say "A problem with attempting to change it to the more-compact version..."

PowerEditor/src/Parameters.cpp
donho: OK, let's not stick on the small detail - both solutions work for me, so we can keep the way you prefer.

xylographe: @sasumner 
Agreed. First implementation is better than casting-version.

Issue #7905, opened 2020-01-30
Feature request: Include doc-size column in Windows-Selection dialog


FootStark:
### Description of feature request

Add a column with the document size to Windows overview (Menu >> Window >> Windows...).
This help comparing file sizes and finding very large open documents.

### Implementation Hints
File for changes: WindowsDlg.cpp
Source of column content: `buf->docLength()`, format with `StrFormatByteSize` (WinAPI: shlwapi.h)

PR #7911, opened 2020-02-01
Add document length to Windows-dialog / 

FootStark:
### This pull request addresses:
- Adds a size column to the windows dialog as requested in #7905 
- Fixes display errors in list when sorting/closing tabs (wrong windows size in _rc)
- Inverted arrows for sort-indication to respect common conventions
- Does some code cleanup to avoid some warnings

Sorry for addressing multiple issues at once, I found the display issue while working on the code. 

### Implementation details:
- The current document sizes are read via `FileManager.docLength()`. As this seems to be very slow (> 10 ms per call), and asynchronous operations don't seem to be supported, a timer loop is used to keep the application responsive. 
- As I am not to firm with C++ naming conventions, the enum names might need review.
- The `getClientRect(_rc);` should likely be promoted to `SizeableDlg::onSize` (not sure why _rc is defined in SizeableDlg but neither updated or used anywhere)

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1113,7 +1113,7 @@ Continue?"/>
             <ColumnName name="Name"/>
             <ColumnPath name="Path"/>
             <ColumnType name="Type"/>
-			<ColumnSize name="Size"/>
+            <ColumnSize name="Size"/>
         </WindowsDlg>
         <AsciiInsertion>
             <PanelTitle name="ASCII Codes Insertion Panel"/>
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1113,6 +1113,7 @@ Continue?"/>
             <ColumnName name="Name"/>
             <ColumnPath name="Path"/>
             <ColumnType name="Type"/>
+			<ColumnSize name="Size"/>
         </WindowsDlg>
         <AsciiInsertion>
             <PanelTitle name="ASCII Codes Insertion Panel"/>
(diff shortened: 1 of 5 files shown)


sasumner:
It is A LOT of code...

FootStark:
Yes, I was surprised myself. Should have been a no-brainer at first (add a column displaying a field value), but there were some challenges that had to be conquered. Went through it again and fixed some lines, but not much to be left out. If you need further information or a walk-through to accept, let me know.

donho:

FootStark:
Ok, just to make sure: Is this a general rejection or shall I break up the changes into single consecutive chunks?
If so, is code refactoring welcome (better future maintenance, e.g. use enum values instead of numeric col idx), or rather keep redundant code with magic numbers for ease of review?

xylographe:
@FootStark
No code refactoring at all!
Read [CONTRIBUTING](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md).

donho:
@FootStark 

1. Your should create the issue for each feature and bug-fix.
2. Create a PR per bug-fix/feature, and ref the issue you created.
3. Refactoring could be accepted (if it really makes sens) but after your PR are merged or rejected.

In your case, I suggest you create a bug report on the issues firstly.

sasumner:
>suggest you create a bug report on the issues firstly

To be fair to @FootStark , the main point of his change _has_ an issue, which he references in the first line of his PR:

>Adds a size column to the windows dialog as requested in #7905

But in general, yes, all the other stuff should be broken out into separate issues, as granular as is reasonable, and then code changes and PRs made, one at a time.

You'll know if your issues are as small as reasonable if your PR text is something like "Fixes #xxxx" or "Implements #xxxx" and nothing more needs to be said.

PR #7912, opened 2020-02-02
Fix Venetian translation

conky77:


donho:
I'm wondering how come so many errors have been done in the first versions since you are the original translators ?

conky77:
Venetan is a spoken language (at home, with friends and very often at work), but at school, it is learnt only Italian. (and in Brasil, Portuguese). So Venetan has many contaminations from Italian(and in Brasil Portuguese):  I translated N++ to Venetan from Italian, so there were many words in Italian, In this version, I tried to avoid Italian contamination, make it more natural, respect the accents rules. I checked and also compared my translation with other software translation Venetian (like Libre Office).Venetan is recognised by Veneto-region and by Brasil, but not yet from Italy: maybe you will create a N++ version for Venetan recognition :-)

Issue #7950, opened 2020-02-15
Sound after an unsuccessful search cannot be disabled


jpvanleeuwen:
### Description of the Issue
After doing an unsuccessful search, the system 'wrong!' error bell is played which I find very annoying but I cannot find a way to disable it. Disabling all sounds or the system error sound in the OS is not a solution because in other cases I do want to hear the error bell, just not after an unsuccessful search  (why would it be an error anyway?)

### Steps to Reproduce the Issue
Search for word that is not in a text file

### Expected Behavior
I would expect to be able to disable the system error bell after an unsuccessful search somewhere in the settings

### Actual Behavior
The system error bell after an unsuccessful search cannot be disabled.

The help file says:
https://npp-user-manual.org/docs/searching/
In addition, the system error bell will sound if a Find Next or Replace action results in the Find what text not being encountered; the bell cannot be disabled.

A similar issue is #2049
where MetaChuh commented on Apr 24, 2019
"please update your notepad++ to the latest version Notepad++ 7.6.6 and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as easy to understand and reproducible as possible.
thank you for your help and best regards."

But I have a even newer version and I cannot disable the annoying error bell.

### Debug Information
Notepad++ v7.8.4   (64-bit)
Build time : Jan 29 2020 - 01:31:03
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Home (64-bit) 
OS Version : 1903
OS Build : 18362.657
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
This seems a reasonable request to me; I will look into it...

sasumner:
>I would expect to be able to disable the system error bell after an unsuccessful search somewhere in the settings

After a code change, it can be done in the Windows 10 "Volume Mixer" by muting the app:

![image](https://user-images.githubusercontent.com/30118311/74597885-a382c800-5034-11ea-857a-ba3c8bb17138.png)

To get to the "Volume Mixer", right-click on the "speaker" icon in the system try, and choose "Open Volume Mixer":

![image](https://user-images.githubusercontent.com/30118311/74597907-07a58c00-5035-11ea-9dea-dfb1e541e807.png)


[...]

Issue #7743, opened 2019-12-15
Publish MinGW Build Instructions


sasumner:
@donho to publish MinGW build instructions for Notepad++, Scintilla, Boost

Also, perhaps the contributing guidelines should be updated so that contributors know that they should check their PR for build problems with MinGW before submitting.

Reference link: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/53452d96e0bf17f9f5eaaeec9c6beb2424f6d8cd#r36432951

sasumner:
I've noticed in the Visual Studio project settings, the setting to treat warnings as errors is turned on.  

* Is this same option turned on in the VS appveyor builds?

* Is an equivalent option turned on in the MinGW appveyor builds?  (I can answer this one and say "no" because I was browsing some of the MinGW appveyor build output and saw warning output although the overall build was successful; here's what I was browsing: https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/29761761/job/3urhas35fcxeg3oh )

xylographe:
Yes in the VS appveyor builds,
- `/W4`  show severe, significant, production quality and informational warnings.
- `/WX`  treat all compiler warnings as errors.

The g++ equivalent of `/WX` is `-Werror`.

sasumner:
Begs the question, why isn't it turned on in the MinGW builds?

xylographe:
Because the MinGW build would fail with `-Werror`. First all the warnings would need to be resolved.

sasumner:
I'll revise the question:  Why isn't it turned on in the MinGW builds and the warnings resolved? :-)

donho:
@sasumner 
The aim of mingw build is for providing more options to build Notepad++ but since the official release  is not built by it, the project won't be aligned with it.

sasumner:
@donho said:

>The aim of mingw build is for providing more options to build Notepad++ but since the official release is not built by it, the project won't be aligned with it

Does that mean we won't be getting build instructions for mingw?
If so we can Close this issue.
But I have further questions/comments:

- I would like to build my PRs with mingw before submitting so that I know appveyor builds won't fail on mingw build.  This has happened to me two times previously.

- Rare discussion of mingw builds seem involve Linux.  I would like to do mingw builds on Windows machine as I currently have no Linux.  Is this reasonable?

- If mingw build is to remain "secret", can we stop appveyor building it?

[...]

PR #7951, opened 2020-02-16
Add external sound control capability for unsuccessful search bell

sasumner:
Implements #7950 .

Adds _winmm.lib_ to project due to the use of the _PlaySound_ function.
I'm not sure what is going to happen with the Appveyor MINGW build of the project because of this, but since there still are no published MINGW build instructions (see #7743), I can't check it.

### DIFF
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -250,7 +250,7 @@ CXX				= $(CROSS_COMPILE)g++
 CXXFLAGS		= $(INCLUDESPECIAL) -DTIXML_USE_STL -DTIXMLA_USE_STL $(UNICODE) -std=c++17 -fpermissive
 INCLUDES		= $(patsubst %,-I%,$(DIRS)) -I./include
 LDFLAGS			= -Wl,--subsystem,windows -municode -mwindows
-LIBS			= -lcomdlg32 -lcomctl32 -lgdi32 -lole32 -loleacc -lshell32 -lshlwapi -ldbghelp -lversion -lcrypt32 -lsensapi -lwintrust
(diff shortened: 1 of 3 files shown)


xylographe:
@sasumner 
I believe, MinGW might build again if you add `-lwinmm` to the LIBS in `PowerEditor/gcc/makefile`,
```
LIBS = -lcomdlg32 -lcomctl32 -lgdi32 -lole32 -loleacc -lshell32 -lshlwapi -ldbghelp -lversion -lcrypt32 -lsensapi -lwintrust -lwinmm
```

donho:
Nice one. But I wonder if it works under Windows 7 ?

sasumner:
>I wonder if it works under Windows 7

Hmmm, I had thought so, but maybe not.  :-(
**_Is it worth keeping as a conditional Windows 10 thing?_**
Win 7 has already passed at least one end-of-support date...

donho:
> Win 7 has already passed at least one end-of-support date...

The support date is like theory vs practice to me.
Microsoft creates the theory to make their wish comes true but in practice there are still a lot of people use Windows 7. 
It's worth to break the retro-compatibility (for 7) for an important issue, but not for such enhancement (even for xp).

I will try to test under Windows 7 and xp - if it doesn't pass we can always do  an option for it.

donho:
Work on Windows 7 as under Windows 8.1 & 10 - Congrats!
I let you modify also the User Manual to show this feature.

donho:
It will be awesome that there are some illustration in the User Manual like you show in:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/7950#issuecomment-586659949

PR #7953, opened 2020-02-17
Fix

Henry2o1o:


chcg:
@schnurlos You may want to have a look.

xylographe:
I prefer the current "Nicht druckbare Zeichen" to the proposed "Symbole anzeigen".

schnurlos:
But Henry2o1o is right. The original text in english is "Show Symbol" ...

Issue #7944, opened 2020-02-14
Build fails with Visual Studio 2019


Margen67:
### Description of the Issue
Build fails due to a warning being treated as an error.
```
1>C:\Users\Margen67\Documents\GitHub\notepad-plus-plus\PowerEditor\src\NppCommands.cpp(465,5): error C2220: the following warning is treated as an error
1>C:\Users\Margen67\Documents\GitHub\notepad-plus-plus\PowerEditor\src\NppCommands.cpp(465,5): warning C4834: discarding return value of function with 'nodiscard' attribute
```
Line it's referencing: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppCommands.cpp#L465

Disabling `Treat Warnings As Errors` produces a seemingly functional build, but obviously is a workaround.

### Steps to Reproduce the Issue
1. Open [PowerEditor\visual.net\notepadPlus.vcxproj](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/BUILD.md#to-build-notepadexe)
2. Build
3. Error

### Expected Behavior
Build completes successfuly.

### Actual Behavior
Build fails due to a warning being treated as an error.

### Debug Information
N/A

sasumner:
`Debug Information` IS important; in this case it should be what version of Notepad++ you are attempting to build.  Moving ahead with assumption it is 7.8.4 which is current as of this writing.

Current Build instruction as of this writing references VS2017 not 2019.  The reference is rather obscure but it is there; it could be made more prominent.  If/when VS2019 becomes the standard tool for building, the build instruction will be changed and the warnings fixed so that a clean build results.

Margen67:
@sasumner I was using latest master.
Why not fix it so VS2017 doesn't have to be used?

CookiePLMonster:
This is a perfectly valid warning to display, so I don't understand why (as always) it has to wait until things break for real instead of fixing it pre-emptively. But whatever floats your boat, I guess.

sasumner:
>latest master

which doesn't mean a whole lot if someone reads this at some future point, without some legwork to figure out what it means; just sayin.

>why (as always) it has to wait

Perhaps the answer is primary-dev @donho, who would be the one to fix it, has limited time?
Don't know; I'm not privy to all of the thought-processes...

If primary-dev wants to reopen this, he will.

CookiePLMonster:
[...]

PR #7956, opened 2020-02-19
Add GitHub Actions CI

Margen67:
`windows-latest` (VS2019) currently fails due to #7944: https://github.com/Margen67/notepad-plus-plus/actions/runs/41560612

`vs2017-win2016` works, but isn't ideal since it's removal is planned: actions/virtual-environments#68

Why to add:
AppVeyor is slow.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,99 @@
+name: CI
+
+on:
+  push:
+    paths-ignore:
+      - 'README.md'
+      - 'appveyor.yml'
+  pull_request:
+    paths-ignore:
+      - 'README.md'
+      - 'appveyor.yml'
+
+jobs:
+  build-vs2017:
+    runs-on: vs2017-win2016
+    strategy:
+      fail-fast: false
+      matrix:
+        configuration: [Unicode Release, Unicode Debug]
+        platform: [Win32, x64]
+        include:
+          - configuration: Unicode Debug
+            scintilla_debug: DEBUG=TRUE
+          - platform: Win32
+            archi: x86
+            platform_input: Win32
+          - platform: x64
+            archi: amd64
+            platform_input: x64
+    steps:
+    - uses: actions/checkout@v2
+    - uses: seanmiddleditch/gha-setup-vsdevenv@v1
+      with:
+        arch: ${{ matrix.archi }}
+    - name: NMAKE Scintilla/SciLexer
+      working-directory: scintilla\win32
+      run: nmake SUPPORT_XP=1 ${{ matrix.scintilla_debug }} -f scintilla.mak
+    - uses: warrenbuckley/Setup-MSBuild@v1
+    - name: Build Notepad++
+      working-directory: PowerEditor\visual.net
+      run: msbuild notepadPlus.vcxproj /p:configuration="${{ matrix.CONFIGURATION }}" /p:platform=${{ matrix.PLATFORM_INPUT }}
+    - name: Set artifact path
+      run: |
+        if ('${{ matrix.PLATFORM_INPUT }}' -eq 'x64' -And '${{ matrix.CONFIGURATION }}' -eq 'Unicode Release') {
+            echo "::set-env name=artifact::PowerEditor\bin64\Notepad++.exe"
+        }
+        if ('${{ matrix.PLATFORM_INPUT }}' -eq 'x64' -And '${{ matrix.CONFIGURATION }}' -eq 'Unicode Debug') {
+            echo "::set-env name=artifact::PowerEditor\visual.net\x64\Unicode Debug\Notepad++.exe"
+        }
+        if ('${{ matrix.PLATFORM_INPUT }}' -eq 'Win32' -And '${{ matrix.CONFIGURATION }}' -eq 'Unicode Release') {
+            echo "::set-env name=artifact::PowerEditor\bin\Notepad++.exe"
+        }
+        if ('${{ matrix.PLATFORM_INPUT }}' -eq 'Win32' -And '${{ matrix.CONFIGURATION }}' -eq 'Unicode Debug') {
+            echo "::set-env name=artifact::PowerEditor\visual.net\Unicode Debug\Notepad++.exe"
+        }
+    - name: Upload Notepad++ artifact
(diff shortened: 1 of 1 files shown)


chcg:
@Margen67 What about using https://github.com/actions/virtual-environments/blob/master/images/win/Windows2016-Readme.md in the meantime until N++ is ready for 2019.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,105 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - uses: seanmiddleditch/gha-setup-vsdevenv@v1
+      with:
+        arch: ${{ matrix.archi }}
+    - name: NMAKE Scintilla/SciLexer
+      working-directory: scintilla\win32
+      run: nmake ${{ matrix.scintilla_debug }} -f scintilla.mak
+    - name: Make Scintilla/SciLexer
+      working-directory: scintilla\win32
```
chcg: Is this needed also for mingw?

Margen67:
@chcg `vs2017-win2016` works, but isn't ideal since it's removal is planned: https://github.com/actions/virtual-environments/issues/68
Therefore #7565 is preferred, since it would also remove the requirement of building with VS2017.

chcg:
@Margen67 Thanks. I also think vs2019 should be used with github actions. Do it have v141_xp or did you mean v142_xp?

See https://packages.ubuntu.com/en/bionic/all/mingw-w64-x86-64-dev/filelist with
/usr/x86_64-w64-mingw32/lib/libsens.a
/usr/x86_64-w64-mingw32/lib/libsensapi.a
/usr/x86_64-w64-mingw32/lib/libsenscfg.a
/usr/x86_64-w64-mingw32/lib/libsensorsapi.a
vs. 
https://packages.ubuntu.com/en/bionic/all/mingw-w64-i686-dev/filelist with just
/usr/i686-w64-mingw32/lib/libsensorsapi.a

Margen67:
v142_xp doesn't exist. It does have 141_xp though: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md#visual-studio-2019-enterprise

donho:
@chcg Could you check if it work please? If it work, is it better than old system?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,105 @@
...
+            platform_input: Win32
+          - platform: x64
+            archi: amd64
+            platform_input: x64
+    steps:
+    - uses: actions/checkout@v2
+    - uses: seanmiddleditch/gha-setup-vsdevenv@v1
+      with:
+        arch: ${{ matrix.archi }}
+    - name: NMAKE Scintilla/SciLexer
```
chcg: @Margen67 Is the step NMAKE Scintilla/SciLexer needed also for mingw?

chcg:
@donho It is working, you could see that at https://github.com/Margen67/notepad-plus-plus/runs/463058432?check_suite_focus=true. The advantage as far as I see with github actions is that the build run in parallel and therefore the overall build time is just 5min 37sec vs. 19 min 19 sec (https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/30996803), also most of the individual builds are slower on  github( >4min) than on appveyor (<3:40), but that may vary depending on the available build engines in the cloud.

Apart from the CI functionality github actions might have some further benefit, if steps currently done manually like checking xml files for a specific format could be checked automatically by some script on each commit or some other action.

.github/workflows/CI.yml
Margen67: No.

donho:
@chcg Thank you for your detail information.
In this case, could you remove appveyor ? There's no point to keep both systems.
OK for enabling Notepad++ to support VS2019.

donho:
I just saw the file is added in `.github` folder, that I dislike strongly. 
Is there anyway to NOT create this folder on the root of project?

Margen67:
> 
> 
> I just saw the file is added in `.github` folder, that I dislike strongly.
> Is there anyway to NOT create this folder on the root of project?

No.

donho:
Then I prefer to not integrate this PR - Sorry for that.
`.github` is not hidden under Windows and I don't think it's reasonable to have a folder for github.
IMO any project could be in any git platform so github structure shouldn't be here.

Margen67:
So you would rather have a slower CI just because it has to be in a folder? That makes no sense. I don't see how it's any worse than a file like AppVeyor has.

CookiePLMonster:
> Then I prefer to not integrate this PR - Sorry for that.
> `.github` is not hidden under Windows and I don't think it's reasonable to have a folder for github.
> IMO any project could be in any git platform so github structure shouldn't be here.

What on Earth is this reasoning? At this point reasons to reject stuff cannot get any more arbitrary than this, it's ridiculous.

If you have a problem with how it's set up, feel free to complain to GitHub Support. I'm sure they'll hear you out.

sasumner:
Prestatement:  I'm not on one side or the other on this issue. :-)

>feel free to complain to GitHub Support. I'm sure they'll hear you out.

Why would they (care)?

>That makes no sense.

> one great way to shoot yourself in the foot

One thing we have to keep in mind is that this is @donho 's project.  He can arbitrarily decide to like something or not like something.  If he doesn't like it, for whatever reason, he should not have it in his project.  And he shouldn't be pressured to.  (Of course, history has shown that he's pretty good at not feeling pressured)

If others feel strongly otherwise, I suppose they should simply spare themselves the stress and move on to another text editing product; there are many available.

donho:
>> feel free to complain to GitHub Support. I'm sure they'll hear you out.

> Why would they (care)?

Good point

> Not to mention that looks like you liked the change and then you rejected it solely over a folder name - that's one great way to shoot yourself in the foot.

Whereas my decision could be unfair, the goal that Notepad++ is released with GPL is that you can fork the project and do whatever you want (by keeping it as GPL of course).

donho:
> So you would rather have a slower CI just because it has to be in a folder? That makes no sense. I don't see how it's any worse than a file like AppVeyor has.

The integrate build is for making sure every commit doesn't break anything. So the building time does not matter to me. What does really matter to me it's Notepad++ running performance.

Issue #4955, opened 2018-10-24
Monitoring (tail -f) is not working 


dia-susca:
### Description of the Issue
I use notepad++ to monitor Websphere application logs, on Windows x64. If I choose the monitor option, the files aren't actually monitored. I still have to right-click on the tab and choose Reload for the file to get updated. 
The automatic updates have a delay that is too big compared to what monitoring should do.

I have version  7.5.9 (64-bit) of notepad++

### Steps to Reproduce the Issue
1. Open a log-file that is written by another (java) application
2. Click the monitor button
3. The log is continuously updated externally

### Expected Behavior
I expect notepad++ to act as a tail -f command

### Actual Behavior
But it actually reloads files just as it did before, whenever it detects a change after a longer time interval

I have tried to manually refresh the file faster, but it is cumbersome to keep right-clicking the file tab. And there is no option in the file editing context-menu to reload the file.
As a work-around, I assigned a shortcut to "Reload From Disk" in Settings/Shortcut Mapper, but it does not work -  I keep pressing the keys without any effect. Only right-clicking the tab and clicking Reload does the expected refresh.

SinghRajenM:
It works fine for me.

Could you save below content in bat file and run the batch file. Then monitor the `time.txt` and see the result. If this works fine, then there might be issue with java logging mechanism.

```Batch
@echo off
setlocal

:loop
echo %time% >> time.txt
timeout /t 2
goto :loop
```


![capture](https://user-images.githubusercontent.com/14791461/47433856-e2956600-d7be-11e8-9c46-41faf947242e.gif)

dinkumoil:
I personally gave it up to use Notepad++ for file monitoring. I found a utility called TailSharp, it can be downloaded [here](http://oette.info/) (look in the right side bar entitled "Downloads"). 

This utility is able to monitor the file created with the script posted by @SinghRajenM above even if you omit the `timeout /t 2` command.

dia-susca:
You are correct that your example works fine. But, for java applications, it doesn't. So it is still a bug.

Here's a sample application that uses log4J (runnable jar, cfg file and sources) that proves that notepad++ monitoring does not work. The file is updated only if you right-click on the file tab.

[TestJavaLogs.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/2518367/TestJavaLogs.zip)

Other monitoring apps like BareTail work on windows, so it must be fixable.

Thank you

[...]

PR #4527, opened 2018-05-25
Supressed reload dialog when Monitoring is enabled  #3142

jrocha:
When Monitoring is enabled N++ was asking for a file reload.

PR #7969, opened 2020-02-22
fix for #3142 and #4955 - "Monitoring" doesn't detect changes

jrocha:
This patch adds an active monitor to detect changes on files since Windows isn't honoring `FILE_NOTIFY_CHANGE_SIZE` or `FILE_NOTIFY_CHANGE_LAST_WRITE`  on `ReadDirectoryChangesW` correctly if the file writer keep it opened.

This solution is based on GNU tail for Windows does. It does it at
`static void tail_forever (struct File_spec *f, int nfiles, double sleep_interval)` on `tail.c`.

This patch is related to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3142 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4955 and https://github.com/notepad-plus-plus/notepad-plus-plus/pull/4527

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -35,6 +35,7 @@
 #include "VerticalFileSwitcher.h"
 #include "functionListPanel.h"
 #include "ReadDirectoryChanges.h"
+#include "ReadFileChanges.h"
 #include <tchar.h>
 #include <unordered_set>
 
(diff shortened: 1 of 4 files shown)


donho:
@jrocha 
I need to be reproduce the bug to validate your fix, could you provide me the way to reproduce this bug?

jrocha:
@donho I've been using this C# [code](https://gist.github.com/jrocha/1fd533ae51bbe9e91278747e1b304dc6) to reproduce this bug.

donho:
@jrocha In this case, could you provide your binary ?

donho:
> @jrocha In this case, could you provide your binary ?

OK, I can reproduce it with your C# code in debug mode.
And check that your code fix it. Good job.

PR #7971, opened 2020-02-23
Added the doc-size column to the Windows Selection dialog

eljefe7000:
Per feature request #7905, I added the doc-size column to the Windows Selection dialog and wired it up with the sort functionality that was already there.  I also did not resize the entire dialog, but made all of the columns fit the control.  Not sure if that matters.

### DIFF
--- a/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
+++ b/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
@@ -156,7 +156,7 @@ struct BufferEquivalent
 
 	bool compare(int i1, int i2) const
 	{
-		if (_iColumn >= 0 && _iColumn <= 2)
+		if (_iColumn >= 0 && _iColumn <= 3)
 		{
 			BufferID bid1 = _pTab->getBufferByIndex(i1);
 			BufferID bid2 = _pTab->getBufferByIndex(i2);
@@ -369,6 +369,16 @@ INT_PTR CALLBACK WindowsDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM lPa
 								generic_strncpy(pLvdi->item.pszText, lang->getLangName(), len-1);
 							}
 						}
+						else if (pLvdi->item.iSubItem == 3) // size
+						{
+							int docSize = buf->docLength();
+							string docSizeText = to_string(docSize);
+							wstring wstr = wstring(docSizeText.begin(), docSizeText.end());
+							const wchar_t * wstrp = wstr.c_str();
+							int docSizeTextLen = lstrlen(wstrp);
+							generic_strncpy(pLvdi->item.pszText, wstrp, docSizeTextLen);
+							pLvdi->item.pszText[docSizeTextLen] = 0;
+						}
 					}
 					return TRUE;
 				}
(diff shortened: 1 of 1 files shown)


PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -521,6 +531,11 @@ BOOL WindowsDlg::onInitDialog()
 	lvColumn.cx = 50;
 	SendMessage(_hList, LVM_INSERTCOLUMN, 2, LPARAM(&lvColumn));
 
+	columnText = TEXT("\u21F5 ") + pNativeSpeaker->getAttrNameStr(TEXT("Size"), WD_ROOTNODE, WD_CLMNNAME);
```
AngryGamer: @eljefe7000

FWIW, this change won't properly translate "size". At present, the column will be translated to "name". See 13801e3540 for guidance.

Also, I bet the "size" translation can be copied from Styler Dialog's `id="25032"` in every language's translation file. Examples: [English](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/13801e35405841cfb8ff0af65023e341ba3dca34/PowerEditor/installer/nativeLang/english.xml#L496) & [French](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/13801e35405841cfb8ff0af65023e341ba3dca34/PowerEditor/installer/nativeLang/french.xml#L538)

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -598,6 +613,23 @@ void WindowsDlg::updateColumnNames()
 	lvColumn.pszText = const_cast<TCHAR *>(columnText.c_str());
 	lvColumn.cx = static_cast<int>(SendMessage(_hList, LVM_GETCOLUMNWIDTH, 2, 0));
 	SendMessage(_hList, LVM_SETCOLUMN, 2, LPARAM(&lvColumn));
+
+	columnText = pNativeSpeaker->getAttrNameStr(TEXT("Size"), WD_ROOTNODE, WD_CLMNTYPE);
```
AngryGamer: @eljefe7000

FWIW, this change won't properly translate "size". At present, the column will be translated to "type". See 13801e3540 for guidance.

Also, I bet the "size" translation can be copied from Styler Dialog's `id="25032"` in every language's translation file. Examples: [English](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/13801e35405841cfb8ff0af65023e341ba3dca34/PowerEditor/installer/nativeLang/english.xml#L496) & [French](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/13801e35405841cfb8ff0af65023e341ba3dca34/PowerEditor/installer/nativeLang/french.xml#L538)

donho: @AngryGamer 
Fixed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/eb6da8195b67a3077310a4d787fd4ac37cd21c76 now.
Thank you for your heads up.

eljefe7000: Thank you both for your help.

PR #7980, opened 2020-02-27
Add support for auto-indenting Perl and Rust.

### LINKED ISSUE #7958: Auto-Indent not working with Perl script

1. Open or create a file named by *.pl ("Perl source file" on left-bottom)
2. Input {}
3. Press Enter between {}

### Expected Behavior

This image records  other editors, but npp has same behavior in C/C++ and JAVA mode
3VbVyT

### Debug Information

Nope, but I tried 7.61x64, 7.6.6x32, 7.81x64, 7.81x32

Here after compare with SciTE, I presume it's Abug?
https://community.notepad-plus-plus.org/topic/18940/is-there-any-plugin-can-make-npp-realtime-auto-indent/10

### LINKED ISSUE #7961: Auto-indent does not support Rust.
### Description of the Issue
When editing a Rust file (extension .rs) auto-indent does not automatically increase the indentation level after an open brace, and does not reduce the indentation when a closing brace is inserted. Other braces languages have this behavior. Although existing indentation is still copied to new lines.

Create a new file called "main.rs" and enter the following code:

    fn main() {

With your cursor at the end of this line, press the enter key.

### Expected Behavior
A new line is created with increased indentation:

    fn main() {
        |  <-- The cursor is now here.

### Actual Behavior
    fn main() {
    |  <-- The cursor is now here.

### Debug Information
Notepad++ v7.8.2   (64-bit)
Build time : Dec  4 2019 - 01:39:49
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.657
Plugins : BetterMultiSelection.dll Comment Wrap.dll ElasticTabstops.dll Explorer.dll mimeTools.dll NppConverter.dll NppExport.dll Python Indent.dll SurroundSelection.dll VisualStudioLineCopy.dll

Kered13:
These languages do no support single line control structures without braces, so they are excluded from isConditionExprLine.

Fixes #7958
Fixes #7961

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2571,8 +2571,12 @@ void Notepad_plus::addHotSpot()
 	_pEditView->execute(SCI_SETSTYLING, 0, 0);
 }
 
-bool Notepad_plus::isConditionExprLine(int lineNumber)
+bool Notepad_plus::isConditionExprLine(int lineNumber, LangType type)
 {
+	// These languages support single line blocks without braces.
+	if (!(type == L_C || type == L_CPP || type == L_JAVA || type == L_CS || type == L_OBJC ||
+	      type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS))
+		return false;
 	if (lineNumber < 0 || lineNumber > _pEditView->execute(SCI_GETLINECOUNT))
 		return false;
 
@@ -2638,8 +2642,10 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 		(eolMode == SC_EOL_CR && ch == '\r')) && prevLine >= 0 && _pEditView->getLineLength(prevLine) == 0)
 		return;
 
+	// These languages use braces for control structures.
 	if (type == L_C || type == L_CPP || type == L_JAVA || type == L_CS || type == L_OBJC ||
-		type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS)
+		type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS ||
+		type == L_PERL || type == L_RUST)
 	{
 		if (((eolMode == SC_EOL_CRLF || eolMode == SC_EOL_LF) && ch == '\n') ||
 			(eolMode == SC_EOL_CR && ch == '\r'))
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2571,8 +2571,12 @@ void Notepad_plus::addHotSpot()
 	_pEditView->execute(SCI_SETSTYLING, 0, 0);
 }
 
-bool Notepad_plus::isConditionExprLine(int lineNumber)
+bool Notepad_plus::isConditionExprLine(int lineNumber, LangType type)
 {
+	// These languages support single line blocks without braces.
```
donho: Do we really need that ?

Kered13: It explains why the function supports those languages, and why it does not support the other languages (Perl and Rust don't support single line blocks without braces). If someone adds new languages in the future, it should help them understand whether they should include the new language in that function or not.

sasumner: Was @donho talking about only the comment line, or the comment line plus the 3 lines that follow?  I find it hard to believe Don would object to the comment alone.

Kered13: Well the lines below are needed because Perl and Rust (and probably other languages I don't know about) _don't_ have single line control structures without braces, so it would be incorrect to indent there if the user did not insert a brace. (I assume this difference is the reason that Perl has not been supported for autoindent before, unlike every other braces language. Rust is pretty new so I sort of understand why it isn't supported yet.)

donho: @Kered13 I need a concrete example :
which code in Rust & Perl working and not working if we keep and remove these lines of code in PR ?

Kered13: With these lines, in Rust:
```
if condition
| <-- indentation goes here
```

In Perl:
```
if (condition)
| <-- indentation goes here
```


Without these lines, in Rust:
```
if condition
    | <-- indentation goes here
```

In Perl:
```
if (condition)
    | <-- indentation goes here
```

The problem with the second examples is that Rust and Perl _require_ braces in an if statement (and all similar statements). It is a syntax error to not have braces. Therefore this line should be treated as a simple multi-line statement, and Notepad++ does not automatically indent multi-line statements.

donho: In this case, I prefer add the follow lines:
```
			else if (type == L_PERL || type == L_RUST)
			{
				_pEditView->setLineIndent(curLine, indentAmountPrevLine);
			}
```
After
```
			else if (nextChar == '{')
			{
				_pEditView->setLineIndent(curLine, indentAmountPrevLine);
			}
```
instead of modifying the behaviour of `isConditionExprLine` function.
If makes the code simpler and more readable.

Issue #7979, opened 2020-02-27
Discuss to improving korean translation


sokcuri:
### Description of the Issue
This issue was created to discuss the improvement of the Korean translation of notepad++
이 이슈는 notepad++의 한국어 번역 개선을 토론하기 위해 만들어졌습니다

A better translation has been discussed at https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7935. so create a new translation related issue. thanks.
PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7935 에서 좀 더 나은 번역에 대한 논의가 이루어져 이슈를 새로 파게 되었습니다. 양해 부탁드립니다

nulLeeKH:
I think this is a very good phenomenon. I hope this will improve the quality of Korean translation.

We also appreciate the PR#7935 has helped make this issue.

sokcuri:
Here are some good looking suggestions

**`블록` -> `선택 영역`**
For me, it sounds more clear.

_Originally posted by @Party4Bread in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7935_

**`모두 책갈피` -> `책갈피 모두 적용`**
I think it looks better than previous one.

_Originally posted by @itholic in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/7935_

**`숫자 앞 0 채우기` -> `숫자 앞에 0 채우기`
Or, maybe "숫자 앞에 0 채우기" is better

_Originally posted by @itholic in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDM4NDk1NzA4Mg==/comments/review_comment_

**`Windows` -> `윈도우`, `Unix` -> `유닉스`, `Mac` -> `맥`
I'm actually fine with this change, as they seem to be commonly transliterated.

For what it's worth, there are two possible transliterations "윈도" and "윈도우" (the former complies to the national standard, the latter is more colloquial). Historically Microsoft never transliterated Windows by themselves, presumably in order to not deal with this situation at all. As we are talking about what Korean speakers would feel comfortable, I would go with "윈도우".

_Originally posted by @lifthrasiir in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MzczNTU5NTc3/timeline/more_items_

sinoru:
> Windows -> 윈도우, Unix -> 유닉스, Mac -> 맥
[...]

PR #7982, opened 2020-02-27
Update korean.xml

nulLeeKH:
Modify korean.xml based on issue #7979


sokcuri:
> About 4008, how about the "관리자 모드 다시 실행해 주세요" change to "관리자 모드로 실행해주세요"
> 
> Because already mentioned about exit program("Notepad를 종료한 뒤") before that sentence.

It was fixed in the latest commit.

PR #8014, opened 2020-03-06
Update Turkish localization

Tmp341:
* Added according to eb6da81 commit.
* Updated according to 5483ad1 commit.


donho:
Thank you @Tmp341

PR #8028, opened 2020-03-10
Updated CSS Keywords

Gitoffthelawn:
I recently went through a bunch of my own source code as well as about half of Mozilla's source code for Firefox.  As I found CSS keywords that were not included in Notepad++, I added them locally.
I then merged them with existing keywords, and sorted the combined list alphabetically.
The result is reflected in this pull request.
I hope it helps you and others.


chcg:
@Gitoffthelawn Is there a relevant spec for the keywords you added? Maybe https://developer.mozilla.org/en-US/docs/Web/CSS/Reference#Keyword_index.

Gitoffthelawn:
@chcg Yes, that's one of the best specs available.  Mozilla is excellent at providing exact syntax, examples, and browser compatibility tables.

Technically, however, the official specs from W3 are located here:
https://www.w3.org/Style/CSS/

With descriptions located here:
https://www.w3.org/Style/CSS/specs.en.html

Gitoffthelawn:
Would you like to merge this into the codebase?

PR #8039, opened 2020-03-13
Update chineseSimplified.xml

popcorner:
make the translation file up to date


popcorner:
new strings added

donho:
Thank you @popcorner

PR #8104, opened 2020-04-11
Changes localization name back from "taiwaneseMandarin.xml"

duangsuse:
So, Hong Kong is using Chinese Traditional now, is HK a part of Taiwan?
Or have they need to use "台灣繁體" instead of "香港繁體"?

~~I'm a radical Honkongness, so I want to change this back.~~ :confused:

It's an attack for Chinese Mainland's dictatorship and legal discrimination against homosexuality, but it's also a __bad example for language/culture/region naming__.

I don't think it's "technical correct", just see below:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bf2cd8e05abf5ca6c93cdc8011e0023292aeedb8/PowerEditor/installer/nsisInclude/langs4Npp.nsh#L343-L349

If you think Taiwan is not a part of China, please do not break our culture into parts, thanks.

It's ridiculous to say that Traditional Chinese and Simplified Chinese are fundamentally different, just like `([''] != 0)` in JavaScript.

### DIFF
--- a/PowerEditor/installer/nativeLang/chineseTraditional.xml
+++ b/PowerEditor/installer/nativeLang/chineseTraditional.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <NotepadPlus>
-    <Native-Langue name="台灣繁體" filename="taiwaneseMandarin.xml" version="7.8.6">
+    <Native-Langue name="正體中文" filename="chineseTraditional.xml" version="7.8.6">
         <Menu>
             <Main>
                 <!-- Main Menu Entries -->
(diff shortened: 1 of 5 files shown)


donho:
I see your view-point, though I don't agree with you.
It's an open source project so you can fork this project, and do things in your way in your repository.

duangsuse:
> I see your view-point, though I don't agree with you.
> It's an open source project so you can fork this project, and do things in your way in your repository.

But, please note this line should be removed (no-op file deletion), since Chinese is no longer a culture in NPP now :laughing: .

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bf2cd8e05abf5ca6c93cdc8011e0023292aeedb8/PowerEditor/installer/nsisInclude/langs4Npp.nsh#L344

PR #8128, opened 2020-04-17
disable upward searches in regular expression mode

### LINKED ISSUE #3640: Backward-direction checkbox enabled with Reg Exp search mode (at N++ startup)

sasumner:
Fixes #3640, although that is only part of the story.  The full story, debated before, is that regular expression searching in a backward direction from the caret position causes matches that the user does not expect.  The best thing that was decided to do (group decision) is to fully disable upward searching.

@donho , I don't know if you disallow this type of coding, but it isn't in the contributing guide, so it is OK until you say it is not.  :-)
If you don't like it, I will change it.  
Here's what I mean:

image

I think this IF statement, used above in the image:

`if ((id == IDM_SEARCH_FINDPREV) && (op._searchType == FindRegex))`

has a bit clearer meaning than its opposite:

`if (!((id == IDM_SEARCH_FINDPREV) && (op._searchType == FindRegex)))`


donho:
>  I don't know if you disallow this type of coding, but it isn't in the contributing guide, so it is OK until you say it is not. :-)

@sasumner It is OK for me such explicit way.
However, I think it's overlooked in this way to disable exgexpr upward search. It's enough to disable it from UI (https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5a951531ad440c0080880b5ddcfda76f28071455).

sasumner:
@donho 

>It's enough to disable it from UI (5a95153).

I do not think that it is enough to do just that.  
I have tested that commit and found out that:

1. Shift+F3 will incorrectly perform a backward regex search
2. With input focus in Find window, Shift+Enter will incorrectly perform a backward regex search
3. Macros that were recorded with an earlier version that allows backward regex search will still play back that search (should be disallowed)
4. Macros that are hand-edited to for a backward regex search will still play back that search (should be disallowed)

Point 3 is arguable, but you can't have point 4 without point 3.

donho:
OK, good point.
Please solve the conflict and I will merge it.

donho:
Thank you @sasumner for the merge.
However the whole contet of `FindReplaceDlg.cpp` has beed removed and added.
Could you check it please?

sasumner:
@donho 

Hmm, I simply used github's web-based conflict resolver for this very simple case.
I guess I don't know what it does behind the scenes?

sasumner:
I suppose I will just give up.

Issue #5997, opened 2019-08-02
Wrong characters shown when reloading a file that changed its encoding.


feqwix:
### Description of the Issue
When reloading a file whose content and encoding was changed by a third-party application, the encoding is correctly detected, but the characters shown are incorrect. After closing and reopening the file the correct characters are shown.

### Steps to Reproduce the Issue
Create a file, add some ANSI encoded text and open it with notepad++:

```
> chcp 1252
> echo ñaña > test.txt
```

<img width="485" alt="pic1" src="https://user-images.githubusercontent.com/13251313/62403166-e3cf9c80-b561-11e9-8d0d-9761642cd66b.png">

Keep file opened in notepad++ window. Return to console and rewrite file content with some UTF-8 endoced text:

```
> chcp 65001
> echo ñañaña > test.txt
```

Return to the notepad++ windows and allow to reload the content.
The encoding detection works fine, but some weird characters are shown:

<img width="469" alt="pic2" src="https://user-images.githubusercontent.com/13251313/62403203-15486800-b562-11e9-89a2-5d6902a9679f.png">

Close and repoen the file and now encoding and text are correctly shown:

<img width="463" alt="pic3" src="https://user-images.githubusercontent.com/13251313/62403210-209b9380-b562-11e9-9e38-996a90acc9de.png">

### Expected Behavior
Notepad++ should show the correct characters using the detected encoding after reloading the file.

### Debug Information

Tested using minimalist package:  `npp.7.7.1.bin.minimalist.x64.7z`

```
Notepad++ v7.7.1   (64-bit)
Build time : Jun 16 2019 - 21:24:47
Path : C:\prog\npp\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : none
```

PR #8132, opened 2020-04-18
Wrong characters shown when reloading a file that changed its encoding. #5997

eljefe7000:
Wrong characters shown when reloading a file that changed its encoding. #5997


xylographe:
After opening a UTF-8-BOM encoded file, (displays fine)

Greek-A
After triggering **Reload**, (the file is unchanged, but display is a mess)

Greek-B

eljefe7000:
I think I have a fix for this, so I'll close this pull request and try again.

Issue #8074, opened 2020-03-26
[Feature request] sort and move macros


GhbSmwc:
this:
![image](https://user-images.githubusercontent.com/13095760/77676684-08e7a400-6f65-11ea-9945-071e565a24e6.png)
On windows, the system environment variables at least lets you move each “command”:
![image](https://user-images.githubusercontent.com/13095760/77676825-3fbdba00-6f65-11ea-81ab-0dd7f17402eb.png)

And also the ability to sort by description alphabecially in case the user have lots of macros.

sasumner:
While it certainly would be nice to have a great user interface for arranging the ordering of macros, it's worth noting that you can currently change the ordering of macros by editing the text of shortcuts.xml and physically moving the text for them around.

GhbSmwc:
Sorry for the late reply, I did check, and did not see the list of macros I created:
![image](https://user-images.githubusercontent.com/13095760/79386394-0f4fb700-7f38-11ea-9014-abcb1cbe916d.png)

sasumner:
@GhbSmwc You are probably looking at the wrong `shortcuts.xml`.  Check the one under `%APPDATA%`.

GhbSmwc:
Thanks. Kinda unintuitive, hopefully the upcomming update lets the user move them around.

sasumner:
>Kinda unintuitive, 

Agreed that it is a deficiency.  What about arranging them with submenus as well?

>hopefully the upcomming update lets the user move them around.

No code changes have been made concerning it, so I'd say the likelihood hovers near 0%.

szaboch:
I'd like to be assigned this issue if possible. I've already implemented/tested the move up/down buttons locally and am currently implementing the sort button

sasumner:
@szaboch It doesn't have to be assigned to you; you are free to submit a PR if you'd like.

PR #8134, opened 2020-04-18
Feature #8074 Macro Move/Sort Buttons Implemented

szaboch:
Implemented the Macro move up/down/sort button as requested at Issue #8074.

Currently only supported for the Macro tab of the Shortcut Mapper menu, however the code could be expanded upon to support the other Shortcut Mapper tabs


szaboch:
Capture

szaboch:
> I don't think a comment like "this is bad" is going to get accepted.

That comment is in there multiple times for some of the other buttons in the Shortcut Mapper. I only included it because I saw the other buttons had it

For example if you're in the Scintilla Commands section of the shortcut mapper, the clear and delete buttons should not be clickable. You can see inside of the the associated switches:

`case STATE_SCINTILLA:
{
	return FALSE;			//this is bad
}`

sasumner:
> I saw the other buttons had it

Coding styles change over the years.
A "new" comment like that begs for an explanation of why it is "bad".
Yours isn't "new", so it's probably OK.  Forget what I said.

donho:
This PR is rejected for the following reasons:
1. These 3 buttons are unnecessary - move up/down could be mouse drag and drop, and sort could be controlled by the column header.
2. These 3 buttons work only for Macros, it makes them look clumsy.

sasumner:
>move up/down could be mouse drag and drop

Two comments:
* Is there this kind of thing anywhere else in the UI?
* Mouse drag-and-drop only doesn't respect keyboardists.

PR #8135, opened 2020-04-19
Update ukrainian.xml

a-polivanchuk:


donho:
@a-polivanchuk 
Please also add the following change:
93f7853

Thank you

Issue #6272, opened 2019-10-25
Simple UX Improvement - Recent Files Vertical Alignment


dsnvwlmnt:
Users who use Recent files extensively will know this pain very well, especially if they have a deep nested structure of folders.

In a long list of recent files, finding the one you want is a giant pain, because the most important part - the filename itself - is not in vertical alignment. Your eyes have to zig-zag instead of quickly scan straight down a list.

Three possible solutions:

1) Put the filename on the left side of the menu, with the folder in brackets (this method is preferred as the folder names align more logically than in 2):
```
1: filename.xyz (C:\foo\bar)
2: filename.123 (D:\bar\foo\bar)
```

2) Right-align the files/folder list instead of the current left-alignment. This way the filenames are all aligned:
```
1: --- C:\foo\bar\filename.xyz
2: D:\bar\foo\bar\filename.123
```

(late edit option # 3) Don't list the containing folder, and put the full path+file in a tooltip (LibreOffice-style):
```
1: filename.xyz
2: filename.123
```

Below is what I'm looking at right now. It's extremely cumbersome and easily improved:

![Untitled](https://user-images.githubusercontent.com/14173022/67607831-b8141b00-f73a-11e9-9b6d-d0fae0c23a9e.png)

sasumner:
I have always simply been glad that I can find a file on the Recent Files list.  

But with the goal of making things better: I rather like the idea advanced by option 1.

MAPJe71:
However, there will always be combinations that have your eyes go zig-zag i.e. it will never be ideal for everyone.

Option 1
```
1: otherfilename.xyz (C:\foo\bar)
2: filename.otherextension (D:\bar\foo\bar)
```
Option 2:
```
1: --------- C:\foo\bar\otherfilename.xyz
2: D:\bar\foo\bar\filename.otherextension
```
Option 3:
```
1: otherfilename.xyz
2: filename.otherextension
```

dsnvwlmnt:
> However, there will always be combinations that have your eyes go zig-zag i.e. it will never be ideal for everyone.

It's a matter of aligning to what 99% of users care about. I suppose there are 3 ways to align things:

1. drive name (current behavior)
2. file name (suggested behavior)
3. file extension (right-aligned)

1. I can't imagine many people primarily care about aligning to drive name, but correct me if I'm wrong. It's the current behavior in np++, but I imagine it was developed this way as an afterthought and never looked at again.

[...]

PR #8136, opened 2020-04-19
Enhancement #6272 Recent Files UI Change

szaboch:
This change re-formats the Recent File List as described at #6272 
Capture1

### DIFF
--- a/PowerEditor/src/lastRecentFileList.cpp
+++ b/PowerEditor/src/lastRecentFileList.cpp
@@ -157,10 +157,22 @@ void LastRecentFileList::updateMenu()
 	{
 		::RemoveMenu(_hMenu, _lrfl.at(i)._id, MF_BYCOMMAND);
 	}
+
+	size_t separator = 0;
 	//Then readd them, so everything stays in sync
 	for (int j = 0; j < _size; ++j)
 	{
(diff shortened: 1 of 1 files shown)


sasumner:
Why not leave the filename off the complete path part and just have the directory?
Example:
`c:\foo\bar\file1.txt`
could be listed as:
`file1.txt (c:\foo\bar\)`
It saves space and eliminates the redundant information, which I thinks gives it a cleaner look.

Some more random thoughts that should have gone into the issue, not the PR:

Would this be best as an _In Submenu_ thing?
In other words, if user doesn't enable _In Submenu_ then display full paths only in main menu.
But if user ticks _In submenu_, have there be another choice under that for what this PR has created?

I'm talking about this preference setting:

image
 
Thus maybe we'd have:

`_ Only File Name`
`_ Full File Name Path`
`_ File Name (File Directory\)`

And now that I'm looking at it, doesn't _Customize Maximum Length_ seem like it should be a _separate_ setting?  I'm guessing that what it does is basically the same as _Full File Name Path_ but truncates to the set number of characters.

donho:
This PR makes the item length much longer.
Even under the `filename.ext (fileDirectoryFullPath)` format, it makes longer.
The current behaviour should be kept, which is strait, simple and clean.

sasumner:
More probably the reason this was rejected is was that it didn't consider preserving existing behavior for users that are used to that behavior and like it.

PR #8113, opened 2020-04-14
Update german.xml to 7.8.7

schnurlos:
Following https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5483ad13d0d580b7858176032819582d9daccc10

Issue #7927, opened 2020-02-09
Grouping box wrapping 'In-Selection' and 'Count' is misplaced


sasumner:
### Steps to Reproduce the Issue

0. (Added post-issue creation).  Start Windows with your primary display one that is greater than 96 d.p.i.  I started mine with my laptop's screen (120 d.p.i) as my primary (and only) display.

1.  Invoke `Replace` window; note positioning of grouping box around `In selection` and `Replace All` controls:  

![image](https://user-images.githubusercontent.com/30118311/74109966-26e47b00-4b56-11ea-8de4-b534d344a6c9.png)

2.  Invoke `Find` window; note positioning of grouping box around `In selection` and `Count` controls:

![image](https://user-images.githubusercontent.com/30118311/74109967-2c41c580-4b56-11ea-8f97-37c3369ddf6c.png)

3.  Compare the results of 1 and 2 and notice that while 1 looks nice and is well-placed; 2 looks bad and is mis-aligned.  For 2, the grouping box needs to be repositioned "upward" a few pixels.

### Expected Behavior
I expect the UI to look nice in this regard. :-)

### Actual Behavior
The UI does not look nice in this regard. :-( 

### Debug Information
Notepad++ v7.8.4   (32-bit)
Build time : Jan 29 2020 - 01:33:47
Path : C:\.........\npp.7.8.4.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1709
OS Build : 16299.1565
Plugins : mimeTools.dll NppConverter.dll NppEditorConfig.dll NppExport.dll

donho:
Is there someone who's got the correct display on "Count" ?
If yes, what's your screen resolution?

sasumner:
@donho Do _YOU_ have the correct display?

donho:
I have had, but not now.
That's why I ask.

sasumner:
I restarted my PC with my primary display at boot time set to one that is 96 d.p.i.  That results in it looking as it should:

![image](https://user-images.githubusercontent.com/30118311/74112054-839c6180-4b67-11ea-9991-9e2a7ea68d80.png)

At the time of issue creation, I had last rebooted on my laptop without any external displays attached, and my laptop screen is 120 d.p.i.

So it appears to be a d.p.i. issue?  It seems like if that group box is badly positioned due to d.p.i., then there would be more things poorly located.

[...]

PR #8140, opened 2020-04-20
Add ability to find-all in selected text of the current document

sasumner:
Implements #8113 

To conduct a search for all occurrences within a (single) selected range of text, tick the _In Selection_ checkbox before pressing the _Find All in Current Document_ button.  Results appear in the _Find result_ window.

Before this change, the only type of search in the _Find_ window/tab that respected _In Selection_ was a _Count_ search.  Thus _In Selection_ checkbox and the _Count_ button appeared boxed in with a "group box".  

Because the  _Find All in Current Document_ button is NOT adjacent to the _Count_ button so that the grouping box could simply be made larger to contain both buttons, it was decided to remove the grouping box on the _Find_ tab (conveniently also solving #7927 !).  

To make up for the now-missing UI cues about the feature, it was decided to add a tooltip to the _In Selection_ checkbox:

image

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1209,6 +1209,11 @@ Continue?"/>
 
 Find in all files except exe, obj &amp;&amp; log:
 *.* !*.exe !*.obj !*.log"/>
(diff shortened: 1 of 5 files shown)


sasumner:
Withdrawing for now; will come back later. :-)

PR #8151, opened 2020-04-21
fix bug of clearing too much when clearing marking in selected text

### LINKED ISSUE #8150: Bug with clearing marking (red background) in selected text
### Description of the Issue

When a selection is active containing text that has been previously "Mark'd" with a red background, and it is desired to clear the red background ONLY WITHIN THAT SELECTION, the red background is cleared at higher document positions than is covered by the selection (i.e., OUTSIDE the active selection).

1) Create a new tab with the following text:

```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```

2a) Do a _Mark All_ operation with the following settings:

image

2b) Observe:

image

3a) Make the following selection of text (selection covers lines 3 and 4) and then do a _Clear all marks_ operation with the following settings:

image

3b) Observe **"Actual Behavior"**; too much marking is removed (see line 5 where the bookmark remains but the red background has been eliminated):

image

3c) Realize that **"Desired Behavior"** is that the red background be removed from only lines 3 and 4, and NOT LINE 5 since line 5 was not in the active selection when _Clear all marks_ was invoked.

### Debug Information

Notepad++ v7.8.4   (32-bit)
Build time : Jan 29 2020 - 01:33:47
Path : C:\...........\npp.7.8.4.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039
Plugins : BetterMultiSelection.dll DSpellCheck.dll linefilter2.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll NppMenuSearch.dll NppToolBucket.dll PythonScript.dll qkNppReverseLines.dll XMLTools.dll

sasumner:
Fix #8150

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2844,7 +2844,7 @@ void FindReplaceDlg::clearMarks(const FindOption& opt)
 		int endPosition = cr.cpMax;
 
 		(*_ppEditView)->execute(SCI_SETINDICATORCURRENT, SCE_UNIVERSAL_FOUND_STYLE);
-		(*_ppEditView)->execute(SCI_INDICATORCLEARRANGE, startPosition, endPosition);
+		(*_ppEditView)->execute(SCI_INDICATORCLEARRANGE, startPosition, endPosition - startPosition);
 
 		if (opt._doMarkLine)
 		{


donho:
@sasumner 
I think there is more serious bug here:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8150#issuecomment-620963478

If it's Scintilla issue, it may be better to remove "In Selection" from "Mark" section.

sasumner:
@donho said:

>I think there is a more serious bug here

I don't see a problem with the fix proposed by this PR.
I just tried the appveyor 32-bit exe from this PR on your data.
I got the following (expected) result after pressing _Clear_ :

image

donho:
@sasumner 
I should try your PR before reporting the bug I discovered. Sorry about that.

Issue #8108, opened 2020-04-13
Compile Notepad++ with /guard:cf


savchenko:
### Description of the Issue
Control Flow Guard is not enabled for `notepad++.exe`

### Steps to Reproduce the Issue
1.  Install latest release (v7.8.5.0 at the time of writing)
2. Check running process with the Process Explorer or use [NetSPI/PESecurity](https://github.com/NetSPI/PESecurity) as following:
```powershell
Get-PESecurity -File 'C:\Program Files\Notepad++\notepad++.exe'

FileName         : C:\Program Files\Notepad++\notepad++.exe
ARCH             : AMD64
DotNET           : False
ASLR             : True
DEP              : True
Authenticode     : True
StrongNaming     : N/A
SafeSEH          : N/A
ControlFlowGuard : False
HighentropyVA    : True
```

### Expected Behavior
`ControlFlowGuard : True`

### Actual Behavior
`ControlFlowGuard : False`

### Debug Information
```
Notepad++ v7.8.5   (64-bit)
Build time : Mar  4 2020 - 11:08:34
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.720
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 
```

Ekopalypse:
@asvc - am I right to assume that activating this feature will kick out earlier windows versions than Windows 8.1? By that I mean, once compiled with this flag npp doesn't run on, let's say, on Windows7?

asvc:
Not sure, but I am happy to spin-up Windows 7 VM and try a binary with CFG enabled if that is of a concern.

asvc:
Not sure, but I am happy to spin-up Windows 7 VM and try a binary with CFG enabled if that is a concern.

Ekopalypse:
Maybe to confirm what MS states from [here](https://docs.microsoft.com/en-us/windows/win32/secbp/control-flow-guard)
_We strongly encourage developers to enable CFG for their applications. You don't have to enable CFG for every part of your code, as a mixture of CFG enabled and non-CFG enabled code will execute fine. But failing to enable CFG for all code can open gaps in the protection. **Furthermore, CFG enabled code works fine on "CFG-Unaware" versions of Windows and is therefore fully compatible with them.**_

chcg:
[...]

PR #8159, opened 2020-04-22
added ControlFlowGuard setting for N++ from #8108

chcg:
- added also to unused scintilla VS project

corresponding  issue #8108

Open point:
- Should it be also added to the make file for scintilla

During my testing on win10 64bit I didn't see problems, no obvious performance penalty


donho:
With some google search, I don't really see the advantage, but it could bring performance issue. So I would say let's keep this PR opened, we integrate it when there's a good reason.

donho:
After one year, we don't feel the need for implementing this feature.
We can close it safely I think.

PR #8161, opened 2020-04-23
Fix wrong Project panel at start issue

### LINKED ISSUE #8126: [Bug] Wrong Project Panel at Start
### Description of the Issue
The intended switch to Project panel 1 after starting Npp with a Workspace file name as parameter 1 doesn't work.

1. Configure *Settings->Preferences->MISC->WorkspaceFileExtension* with a valid workspace extension.
2. Start Npp with the name of a valid workspace file as first parameter.
3. Open at least one more Project panel with *View->Project->Project panel 2*
There have to be at least two Project panels open to see the effect. If already are at least two Project  panels open, you can omit this step.
4. End Npp.
5. Start Npp with the name of a valid workspace file as first parameter again.

### Expected Behavior
Npp should show the first Project panel, wich should show the project described in the workspace file.

### Actual Behavior
Npp shows the Project panel with the highest number. If I switch to the first Project panel, then this shows the correct content.

### Debug Information
Notepad++ v7.8.5   (32-bit)
Build time : Apr 16 2020 - 11:48:34
Path : C:\isys-vision\Tools\Npp\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Plugins : none

### My workaround
The intended switch is programmed in **NppIO.cpp**, near line 211 as follows:

    if (isFileWorkspace(longFileName) && PathFileExists(longFileName))
    {
        nppParam.setWorkSpaceFilePath(0, longFileName);
        command(IDM_VIEW_PROJECT_PANEL_1);
        return BUFFER_INVALID;
    }
The `command(IDM_VIEW_PROJECT_PANEL_1)` call shows no effect, since the Project panels are loaded later in the startup process. This load ends always with the last Project panel being the current Project panel.

In **NppIO.cpp**, instead of calling `command(IDM_VIEW_PROJECT_PANEL_1)`, I set a variable, `__WorkspaceFileLoadedFromCommandLine`, to `true`.

Then, later in the startup process, in **Notepad_plus_Window.cpp**, around line 242:

    // Launch folder as workspace after all this dockable panel being restored from the last session
    // To avoid dockable panel toggle problem.
    if (cmdLineParams->_openFoldersAsWorkspace)
    {
        _notepad_plus_plus_core.launchFileBrowser(fns, true);
    }
    else if (_notepad_plus_plus_core._WorkspaceFileLoadedFromCommandLine)
    {
        ::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1, 0);
    }
At this stage, `IDM_VIEW_PROJECT_PANEL_1` works.

It is my work-around only. There might be a better way to solve this.

This issue seems to be related to #3519, but can be solved without remembering the last open Project panel.

Uhf7:
Fixes #8126

I do not say, that this is the right way to fix it.

I only say, the `IDM_VIEW_PROJECT_PANEL_1` command works in **Notepad_plus_Window.cpp**, but not in **NppIo.cpp**. 

What I did is the most none-invasive solution, I kept the checks where they were and shifted the `IDM_VIEW_PROJECT_PANEL_1` command to a later point in the startup process, where it works.

### DIFF
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -388,6 +388,8 @@ private:
                                               // then WM_ENDSESSION is send with wParam == FALSE
                                               // in this case this boolean is set true, so Notepad++ will quit and its current session will be saved 
 
+    bool _isWorkspaceFileLoadedFromCommandLine = false;
+
 	ScintillaCtrls _scintillaCtrls4Plugins;
 
 	std::vector<std::pair<int, int> > _hideLinesMarks;
--- a/PowerEditor/src/Notepad_plus_Window.cpp
+++ b/PowerEditor/src/Notepad_plus_Window.cpp
@@ -245,6 +245,10 @@ void Notepad_plus_Window::init(HINSTANCE hInst, HWND parent, const TCHAR *cmdLin
 	{
 		_notepad_plus_plus_core.launchFileBrowser(fns, true);
 	}
+	else if (_notepad_plus_plus_core._isWorkspaceFileLoadedFromCommandLine)
+	{
+		::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1, 0);
+	}
 
 	// Notify plugins that Notepad++ is ready
 	SCNotification scnN;
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -211,7 +211,8 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive, 
 	if (isFileWorkspace(longFileName) && PathFileExists(longFileName))
 	{
 		nppParam.setWorkSpaceFilePath(0, longFileName);
-		command(IDM_VIEW_PROJECT_PANEL_1);
+		_isWorkspaceFileLoadedFromCommandLine = true;
+//		command(IDM_VIEW_PROJECT_PANEL_1);
 		return BUFFER_INVALID;
 	}
 


sasumner:
>I do not say, that this is the right way to fix it.

Honesty is appreciated.
However, it may not be the path to getting a PR accepted.
If YOU feel it isn't the best way, what are others to think? :-)

Uhf7:
Thank You for that hint.

The pull request was created only to make it easier to review my workaround I described in issue #8161. I'm actually not eager to get my pull request accepted, I want to use Project panels with Workspace files. And I think, more people could use them, if they would work a little better. It's nearly all there, if one would see the workspace specified in the command line after start, and if one could search them through, it would be close to perfect.

Now, after your comment, I debugged the startup sequence of events again, and I'm shure, that it is hard to make this fix cheaper (3 lines of code) or more side-effect-free. But there are many who know  much more about Notepad++ internals than I do. And they might find something better.

Final thought: Given the existing code base, the workaround I suggested is safe and easy to apply. The efford to find, if it exists, the "right" solution (like changing initialization sequences or similar restructuring), is too unpredictable to solve this little problem.

donho:
Fix #8126

donho:
@Uhf7 
This PR won't work on the following scenario:
1. Open 1 Project Panel, add "Project 1" on the the WorkSpace and save it as WS1.ws.
2. Open 2nd Project Panel, add "Project 2" on the the WorkSpace and save it as WS2.ws.
3. notepad++.exe c:\yourWsDir/WS2.ws

Uhf7:
@donho,

Actually I think it does work, but we seem to have different objectives here. The current behavior without this PR is, when performing step 3 from your list:
- Npp opens two Project Panels. The first Project Panel contains my WS2 workspace, the second Project Panel is empty. But, the second Project Panel becomes the active Project Panel, and if I want to see my workspace, I have to switch back to the first project panel.

The behavior with this PR is:
- Npp opens two Project Panels. The first Project Panel contains my WS2 workspace, the second Project Panel is empty. But now, the first Project Panel becomes the active Project Panel, no switching necessary anymore.

What I think what you are thinking is: You want to implement a kind of recognition mechanism, which correlates the workspace file name from the command line with the earlier opened workspaces, like this:

- Npp opens two Project Panels. The first Project Panel contains the WS1 workspace, because it is memorized in the `config.xml`. The second Project Panel opens WS2, because it is memorized in the `config.xml` too. And then, Npp switches to the workspace specified in the command line.

This mechanism was never implemented nor requested by me. It also has many holes in it. What happens, if all Project Panels are busy, and I specifiy a new WS file in the command line? Discard the oldest Project Panel? Discard the 1st? Or the 3rd? No good idea.

So I would go even a step further and let the command line override the memorized Project Panels. So if I don't specify a WS file in the command line, I get my memorized Project Panels. If I specify a WS in the command line, only one Project Panel opens, wich contains my WS.

Another point: The workspace files itself are independent from each other, there is not stored "I'm workspace 1", or "I' m workspace 3" in the workspace files. And this is good. And I'm not confident about linking a workspace file name from the command line with something stored earlier to restore an erlier state. This is good, if I start Npp without parameters, but not in this case.

donho:
@Uhf7
I just tried my scenario with the following steps:

1. Open 1 Project Panel, add "Project 1" on the the WorkSpace and save it as WS1.ws.
2. Open 2nd Project Panel, add "Project 2" on the the WorkSpace and save it as WS2.ws.
3. close Notepad++
4. `notepad++.exe c:\yourWsDir/WS2.ws`

Now WS2 is opened in both Panels. It's a more serious regression than old behaviour.

Uhf7:
@donho,
your are right in one point, the second workspace was empty in my test, because there was no workspace file stored for it in the `config.xml`.

But I still can't confirm the regression. I just tried with the current master (the official version, not mine), and I did as follows:
- Start notepad++ without command line parameter.
- Close project panel.
- Close notepad++
- Edit `config.xml` with Windows-Notepad, so that the previously stored workspace file names are gone and it looks like this:
`    <ProjectPanels>`
`        <ProjectPanel id="0" workSpaceFile="" />`
`        <ProjectPanel id="1" workSpaceFile="" />`
`        <ProjectPanel id="2" workSpaceFile="" />`
`    </ProjectPanels>`
- Open Notepad++ without command line parameters again.
- View Project Panel 1. It opens without a project in it.
- Add new project to Project Panel 1, name it "Project1".
- Right click on "Workspace", "Save As", Filename `Ws1.npj` (my ending for workspace files).
- View Project Panel 2. It opens without a project in it.
- Add new project to Project Panel 2, name it "Project2".
- Right click on "Workspace", "Save As", Filename `Ws2.npj`.
- Close Notepad++

Before watching what happens, here is how the `config.xml` now looks like:
`    <ProjectPanels>`
`        <ProjectPanel id="0" workSpaceFile="C:\Test\ws1.npj" />`
`        <ProjectPanel id="1" workSpaceFile="C:\Test\ws2.npj" />`
`        <ProjectPanel id="2" workSpaceFile="" />`
`    </ProjectPanels>`

Now I start notepad++ with the command line parameter `ws2.npj`. It opens two Project Panels. Project Panel 1 contains Project2 from `ws2.npj`. Project Panel 2 contains Project2 from `ws2.npj` too. The active Project Panel is Project Panel 2.

Now I close notepad++ without doing anything else (I only switched to Project Panel 1 to see what's in it). Project Panel 1 is still active, when I close it. Now, the `config.xml` looks like this:
`    <ProjectPanels>`
`        <ProjectPanel id="0" workSpaceFile="c:\Test\ws2.npj" />`
`        <ProjectPanel id="1" workSpaceFile="C:\Test\ws2.npj" />`
`        <ProjectPanel id="2" workSpaceFile="" />`
`    </ProjectPanels>`

There is no regression, it has always been this way. That's why I suggested, in my previous comment on this,  to omit loading the workspace files from the `config.xml` at all, if a command line parameter with a workspace file name is provided. 

But the behaviour of the original program code is: If there is a command line parameter with a workspace file is specified, replace the workspace in Project Panel 1 with the one from the parameter file, load all other workspaces from the `Session.xml`, and switch to Project Panel 1. It is coded this way, intended or not. And the switch to Project Panel 1 doesn't work in the original code, because it is called too early in the timeline. And that's all what this PR does, moving the switch to Project Panel 1 to a later point, where it works.

donho:
Sorry the term "regression" may not accurate in this context.
What I mean is the current behaviour is better than the one of this PR under the scenario I provided.

2 workspace ws1 and ws2 were opened in 2 panels in the previous session, if you run `notepad++.exe c:\yourWsDir/WS1.ws` or `notepad++.exe c:\yourWsDir/WS2.ws` won't change anything, and you may need click the tab to switch to the workspace you want - that's the current behaviour.

But the behaviour of this PR could make the same workspace opened in 2 panels - it's a bug for me, that's why I use "regression" term.

Uhf7:
>  But the behaviour of this PR could make the same workspace opened in 2 panels 

If I open notepad++.exe c:\myWsDir\WS2.ws, then I have Ws2.ws loaded in both Project Panels, like I described above. And this is so in the **current version**. And if this is a bug, then it exists in the current version, and it is not a result of this PR. That's why I refuse the term regression so hard.

sasumner:
>And this is so in the current version

Instead of saying it this way, actually state the current version number.  It is "history-safe", and it is shorter to just type "x.x.x" rather than "current version".

Uhf7:
Thank you, I ment Version 7.8.6, fdcd11ccc6160d51de583326441a620ad9905611. But it is same since at least 7.5.7, when I wondered the first about it.

donho:
@Uhf7 
> If I open notepad++.exe c:\myWsDir\WS2.ws, then I have Ws2.ws loaded in both Project Panels, like I described above. And this is so in the current version. And if this is a bug, then it exists in the current version, and it is not a result of this PR. That's why I refuse the term regression so hard.

My bad. Indeed, I can reproduce it in the current version.

PR #8163, opened 2020-04-23
Update belarusian.xml

a-polivanchuk:
This update contains the latest translations from @BORNAME via Crowdin


donho:
@a-polivanchuk 
Please also add the following change:
93f7853

Thank you

PR #4266, opened 2018-03-06
function list: Perl function regex update

yuzhy8701:
Fixing the issue #4265

PR #8198, opened 2020-05-01
update perl keywords and auto-complete

pryrt:
update perl syntax highlighting based on Perl v5.30
* update perl's keywords in langs.model.xml, adding missing keywords
* update perl's autocompletion list, adding missing keywords and adding parameter completion


chcg:
@pryrt Maybe you also want to take a look at #4266 and if the new perl version introduces updates to the function list, but since some time function list improvement is blocked by missing unit tests for it.

pryrt:
The #4266 was in functionList.xml; this is in langs.model.xml and autocompletion's perl.xml ... but maybe there's a block to improvements in those two config files, too? (I don't know if there are tests for those).  If it's suspended or rejected like that other PR, I'll just publish them as a gist and advertise them in the Community Forum, but I was hoping that we could get the config files updated in the official distribution, which is why I submitted them.

chcg:
@pryrt Sorry for the confusion. There is no problem as far as I know with updating keywords and autocompletion. Just the corresponding functionlist for perl is problematic. 
From my understanding these 3 config files (or sections of them) of N++ together with the support from scintilla are making up the programming language support for a language.

chcg:
@pryrt See https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/37078284/messages fails due to 

	<KeyWord name="goto" func="yes"><Overload retVal="" descr="create spaghetti code"><Param name="&NAME" /></Overload><Overload retVal="" descr="create spaghetti code"><Param name="EXPR" /></Overload><Overload retVal="" descr="create spaghetti code"><Param name="LABEL" /></Overload></KeyWord>

is &NAME intended here or should it be just NAME?

pryrt:
> @pryrt See https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/37078284/messages fails due to....
> from #8198. Is &NAME intended here or should it be just NAME?

It needs the ampersand (it's telling the perl author that it needs a sub reference in that position).

The XML should this have `&amp;NAME`

PR #8201, opened 2020-05-02
mingw build with all warnings enabled

chcg:
See https://github.com/notepad-plus-plus/notepad-plus-plus/issues/7743. Enabled all warnings and started to fix them.

- same warning setting as scintilla
- fixed the pedantic ones
- removed multiple unused vars
- corrected indentation warning with if clauses due to mixed spaces/tabs within one function
- reorder warning on constructor init list not matching with the order given by the occurrence in the header file


donho:
The aim of mingw build is for providing more options to build Notepad++. I will fix some code error like unused variables myself but since the official release  is not built by it, the project won't be aligned with it.

chcg:
@donho Is any other class of warnings acceptable to be added as PR ? At least enabling more warnings to have them visible.

Issue #8133, opened 2020-04-18
Add "Find all in selection" capability


sasumner:
In current Notepad++ (7.8.5) there is no ability to conduct a search operation in a selected region of text.  

The ability should exist to do so for a single (i.e., "main") selection.  

The results of the search should go to the `Find result` window, just like, for example, current *Find All in Current Document* output does.

sasumner:
So in implementing this the hardest part is going to be dealing with similar effects to the problem reported here:  #7927 
:-(
:-(

sasumner:
User desire:  https://community.notepad-plus-plus.org/topic/15209/find-in-selected-text

Uhf7:
After reading #8207, I have another idea to implement this consistently:
- Remove the **Find All In Current Document** button from the **Find** page.
- Replace the **Count** button of the **Find** page with a **Find All** button, equal to the **Replace All** button in the **Replace** page.
  _Assumption:_ We don't need the **Count** button, because the list of search results contains the number of hits anyway.
- Now, the new **Find All** button finds all occurrences of the search target within the current document or in the current selection, depending on the checkmark beside it. And, **Find All** now works very similar to **Replace All** in the **Replace** page.

The function seems to be already there, only that it will be triggered by **Find All** instead of **Find All In Current Document**.

sasumner:
@Uhf7 

Part of the goal when changing long-established software is to not change it too much; we wouldn't want existing users to be looking for a "missing" **_Find All in Current Document_** button, putting issues in saying "where did it go?"  And they would, I have no doubt of that.

Another note on "changing too much": In my PR notes, I talked about how I considered and discarded the idea of swapping the placement of two buttons -- as too disruptive to an established UI.

**_Count_** on its own is still useful, at least to me as a user as well as a developer.

**_Find All_** would be confusing, because the Find in Files tab already has a **_Find All_** button.

That all being said, I'm certainly not vehemently opposed to your suggestions, but @donho, as software owner, has final decision.  Let's see if he has any input on it.  We all want the best design.

Uhf7:
The problem of the **_Find All_** button in two different pages could be solved by renaming the button in the **Find in Files** page to **_Find in Files_**. Then it would perfectly fit to the button below: **_Replace in Files_**.

sasumner:
>The problem of the Find All button in two different pages could be solved by renaming...

Yes, but again you are talking about changing an established UI.

[...]

PR #8207, opened 2020-05-04
Add ability to do a find-all in a selection

sasumner:
Implements #8133 

To conduct a search for all occurrences within a (single) selected range of text, tick the **_In Selection_** checkbox before pressing the **_Find All in Current Document_** button. Results appear in the **_Find result_** window.

Before this change, the only type of search in the **_Find_** window/tab that respected **_In Selection_** was a **_Count_** search. Thus **_In Selection_** checkbox and the **_Count_** button appeared boxed in with a "group box".

Because the **_Find All in Current Document_** button is NOT adjacent to the **_Count_** button so that the grouping box could simply be made larger to contain both buttons, it was decided to remove the grouping box on the **_Find_** tab (conveniently also solving #7927 !).

The alternative of swapping positions of **_Find All in Current Document_** button and the **_Find All in All Open Documents_** button was discarded for another reason as well:  possibly too disruptive to users used to the current placement.

To make up for the now-missing UI cues about the feature, it was decided to add a tooltip for the In Selection checkbox; text very similar to the following:

image

When a In Selection find-all is conducted the output in the Find result window will reflect it by using the verbage "selection" instead of "file"; example:

image

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -219,7 +219,8 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case WM_FINDALL_INCURRENTDOC:
 		{
-			return findInCurrentFile();
+			bool searchScopeIsSelectedText = (wParam == 0) ? false : true;
+			return findInCurrentFile(searchScopeIsSelectedText);
 		}
 
 		case WM_FINDINFILES:
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1209,6 +1209,12 @@ Continue?"/>
 
 Find in all files except exe, obj &amp;&amp; log:
 *.* !*.exe !*.obj !*.log"/>
+            <inselection-checkbox-tip value = "Applies ONLY to:
+- Count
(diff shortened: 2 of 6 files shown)


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -219,7 +219,8 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case WM_FINDALL_INCURRENTDOC:
 		{
-			return findInCurrentFile();
+			bool searchScopeIsSelectedText = (wParam == 0) ? false : true;
```
xylographe, line 222: Perhaps, **`bool searchScopeIsSelectedText = wParam != 0`**?

sasumner: Yea, my code is strange there.
I've got to believe it came from an evolution of a slightly different bit of logic I was working on there, but I don't have any way to remember that. :-)

sasumner:
@donho 

This recent commit to make the _Find result_ window language translatable: 
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2b345c8f39310a921fc90ae38196ed323cb1a66a

contains changes that would make acceptance of this current PR difficult.

I'll withdraw this PR, make it compatible with the other changes, and resubmit.

PR #8212, opened 2020-05-04
Uploading new DansLeRuSH dark theme

DansLeRuSH:
As seen with Don, I propose a new dark theme for Notepad++

Best regards,
Franck

### DIFF
--- a/PowerEditor/installer/themes/DansLeRuSH-Dark.xml
+++ b/PowerEditor/installer/themes/DansLeRuSH-Dark.xml
@@ -1,0 +1,910 @@
+<?xml version="1.0" encoding="Windows-1252" ?>
+<!--// DansLeRuSH's dark theme for Notepad++ / Notepad plus plus
+----------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


donho:
@DansLeRuSH  Your xml file is not a valid xml file (ie. it cannot be opened in any browser)

donho:
BTW your branch name `DansLeRuSH-master` is a bad name. For the sake of workflow, you should create a new branch with a unique name for each PR (for example: `DansLeRuSH-theme-20200510`).
Please check [the rules of contribution](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests) then recreate a new PR.

DansLeRuSH:
> 
> 
> @DansLeRuSH Your xml file is not a valid xml file (ie. it cannot be opened in any browser)

Hi,

My bad I put too much "-" in comments.
I correct and do another PR with a better name.

Best regards,
Franck

PR #8231, opened 2020-05-07
Fix Issue 8067 Doc switcher icon state update for cloned documents

### LINKED ISSUE #8067: Doc switcher "dirty" icon state does not stay in sync when documents are cloned
### Description of the Issue
Doc switcher "dirty" icon state does not stay in sync when documents are cloned (split view). 

1. Enable doc switcher (Preferences -> Document List Panel -> Show)
2. Create new document, then right-click and Clone to Other View
3. Start typing. 

### Expected Behavior
Both blue icons in the Doc Switcher panel  should go red

### Actual Behavior
Only one blue icon goes red. 

### Debug Information
Notepad++ v7.8.5   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1903
OS Build : 18362.592
Plugins : HexEditor.dll mimeTools.dll NppConverter.dll NppExport.dll 

doc switcher bug

Uhf7:
Fixes #8067

### DIFF
--- a/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
+++ b/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
@@ -242,7 +242,6 @@ void VerticalFileSwitcherListView::setItemIconStatus(BufferID bufferID)
 			{
 				ListView_SetItemText(_hSelf, i, 1, (LPTSTR)::PathFindExtension(buf->getFileName()));
 			}
-			break;
 		}
 	}
 }


PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
```diff
@@ -242,7 +242,6 @@ void VerticalFileSwitcherListView::setItemIconStatus(BufferID bufferID)
 			{
 				ListView_SetItemText(_hSelf, i, 1, (LPTSTR)::PathFindExtension(buf->getFileName()));
 			}
-			break;
```
sasumner, line 245: I might have changed the `break;` line to a comment, like:

`// do not break here because we could be dealing with a cloned document`

I would do this because the next reader of this source code might think: "This is inefficient, why don't we just break out once we've found the matching buffer" and then they are inclined to put the `break` back in.  :-)

But it is OK the way it is; don't change it.
Wait for acceptance or rejection.

sasumner:
@donho 
@Uhf7 

I have tested this PR and I believe that it solves the originating issue.

PR #8238, opened 2020-05-08
Find-in-Files "Filters" box of ONLY "exclusion" items finds no matches; should auto-add *.*

### LINKED ISSUE #8169: Find-in-Files "Filters" box of ONLY "exclusion" items finds no matches; should auto-add *.*
### Description of the Issue
Notepad++ 7.8.2 added the ability to use "exclusion" specs in the _Filters_ box for _Find in Files_.

Example: Put `!*.log` in _Filters_ to exclude files ending with `.log` from the search.  Multiple exclusion specs can be used, such as `!*.log !*.obj` to suppress searching of log and object files.

However, as pointed out here:

https://community.notepad-plus-plus.org/topic/18609/question-regarding-new-find-in-files-filter-excluding-ability-in-version-7-8-2

using **only** "exclusion" specs results in no possible matches.

The technique that works is to put `*.* !*.log` in the _Filters_ box.  Note the addition of the `*.*`

But, one can leave the _Filters_ box totally empty and the effective filter spec will be `*.*`, so why shouldn't the case of "only exclusion filters" also infer a `*.*`?

1. Perform any _Find in Files_ search leaving the _Filters_ box empty.  Note results; the assumption is that you search for something that will generate some number of hits.
1a. Note also that after the search, Notepad++ has put `*.*` in the _Filters_ box for you.  How nice.
2. Remove the `*.*` and put _only_ `!*.zzz` in the _Filters_ box, where `zzz` is any file extension that you don't have present in the directory tree of the search.
3. Run the _Find in Files_ search again.

### Expected Behavior

I expected the same hits from the search in step 3 as were obtained in step 1.

### Actual Behavior

No hits occur, i.e. `(0 hits in 0 files)` .

**BOTTOM LINE FOR THIS ISSUE**:  Using only "exclusion" specs in the _Filters_ box should result in a search of all files except the exclusion matches.

### Debug Information
Notepad++ v7.8.6   (32-bit)
Path : C:\.............\npp.7.8.6.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039

eljefe7000:
Fixes #8169

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -653,6 +653,7 @@ void cutString(const TCHAR* str2cut, vector<generic_string>& patternVect)
 {
 	if (str2cut == nullptr) return;
 
+	bool aPositivePatternFound = false;
 	const TCHAR *pBegin = str2cut;
 	const TCHAR *pEnd = pBegin;
 
@@ -661,15 +662,24 @@ void cutString(const TCHAR* str2cut, vector<generic_string>& patternVect)
 		if (_istspace(*pEnd))
 		{
 			if (pBegin != pEnd)
+			{
 				patternVect.emplace_back(pBegin, pEnd);
+				aPositivePatternFound = (generic_string(pBegin)[0] != '!');
+			}
 			pBegin = pEnd + 1;
-		
+
 		}
 		++pEnd;
 	}
 
 	if (pBegin != pEnd)
+	{
 		patternVect.emplace_back(pBegin, pEnd);
+		aPositivePatternFound = (generic_string(pBegin)[0] != '!');
+	}
+
+	if (!aPositivePatternFound)
+		patternVect.emplace_back(TEXT("*.*"));
 }
 
 


sasumner:
For this to be a proper PR, it should have the issue it fixes as its first "comment", instead of being left like this:

image

As it appears to address issue 8169, I will attempt to supply that information in the correct place.

eljefe7000:
My apologies.  I added the description in the commit, but forgot to add it for the PR.

sasumner:
>My apologies. 

No worries.

>I added the description in the commit, but forgot to add it for the PR.

This part confuses me, because I was talking about the issue number in the PR comment, and wasn't saying anything about the "commit description".

The "commit description" should be a relevant description of the change.  This is because, going forward, that's what becomes visible in the master history.

If you just copy the issue description (or close to it) as the commit message, it may not make sense.  

For THIS change I might have used a commit message of `Assume *.* when only exclusion specs are specified in Filters box`, or something like that.

Describing the defect in the issue title makes perfect sense, including it as part of the description of the commit that addresses the defect may or may not make sense.

But, don't make any further changes to it now, as that would just make it confusing.

eljefe7000:
Okay, I will leave it alone until further notice.  Thanks for the advice.

sasumner:
@donho 
@eljefe7000 

Because I was the one that created the original issue, I tested this PR.
I believe it works to resolve the issue.

PowerEditor/src/Parameters.cpp
```diff
@@ -653,6 +653,7 @@ void cutString(const TCHAR* str2cut, vector<generic_string>& patternVect)
 {
```
donho, line 653: `cutString()` is a generic function which uses white space to build token array.
Not mention that this function is used in other places,  adding other semantic here for the specific need here is just wrong.

sasumner: Note that I didn't look at code for this PR other than what is shown in github's web view.
I did run the code from Appveyor build and it worked, but obviously that is only part of what is required for successful PR.
I'm sorry for not looking deeper at it.  I fully agree with rejection reason.

eljefe7000: My apologies for choosing cutString as the place for the change.  I initially put the code in FindReplaceDlg but thought it might cut down on code repetition if it were in cutString.  I'll try to do a better job of analyzing the code in the future.

sasumner: @eljefe7000 Live and learn, like always...everyone...every day...

Issue #8096, opened 2020-04-07
Feature Request: Add Regex cheat sheet


NealBScott:
MS Visual studio search and replace dialog has a 'cheat sheet' to help remember what the common regex expressions are.   Would be really nice to see similar in Notepad-plus-plus especially for the folks that don't use them every day!

![image005](https://user-images.githubusercontent.com/63323208/78728140-94b5e480-7904-11ea-96bf-b284c77ad631.png)

sasumner:
Probably too many nuances in regex to make this truly useful.

Maybe better would be to create some hyperlinks on the Run menu, that reference the Boost regex documentation, or maybe https://www.regular-expressions.info/

alankilborn:
The "Run" menu suggestion seems a good one.

* Click "Run" menu
* In box that appears, put `https://npp-user-manual.org/docs/searching/#regular-expressions`
![image](https://user-images.githubusercontent.com/77065706/183104046-08b28cf1-cb3a-456d-acee-6ac305f8844e.png)

* Click "Save"
* In the name box that appears, put `Regular expression help`
![image](https://user-images.githubusercontent.com/77065706/183104111-614c60c5-70dd-4e5d-bea6-913f317b9df6.png)
* Click "OK"
* Click "Cancel"
* Now an item will appear when clicking on the "Run" menu:
![image](https://user-images.githubusercontent.com/77065706/183104401-46f04cab-f270-4e90-ae23-e5ac233d1149.png)
Selecting the item will open a browser window with the Notepad++ user manual on regular expressions in view.

PR #8241, opened 2020-05-08
Add Regex Help Menu Item

JahJoey:
Pull Request For #8096

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2755,6 +2755,12 @@ void Notepad_plus::command(int id)
 			break;
 		}
 
+		case IDM_REGEX:
+		{
+			::ShellExecute(NULL, TEXT("open"), TEXT("https://www.regular-expressions.info/"), NULL, NULL, SW_SHOWNORMAL);
+			break;
+		}
+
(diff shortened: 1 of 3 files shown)


sasumner:
This is more appropriate for a _Run_ menu entry, set up by end-users.

Also, information on that site is not specific to the regular expression engine used in Notepad++, so it could potentially confuse Notepad++ users on certain aspects of regex.

PowerEditor/src/NppCommands.cpp
```diff
@@ -2755,6 +2755,12 @@ void Notepad_plus::command(int id)
 			break;
 		}
 
+		case IDM_REGEX:
+		{
+			::ShellExecute(NULL, TEXT("open"), TEXT("https://www.regular-expressions.info/"), NULL, NULL, SW_SHOWNORMAL);
```
xylographe, line 2760: Apparently, NPP is build with Boost 1.70:
```
https://www.boost.org/doc/libs/1_70_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html
https://www.boost.org/doc/libs/1_70_0/libs/regex/doc/html/boost_regex/format/perl_format.html
```

Issue #2946, opened 2017-02-23
Add an option to toggle the "Search Results Window"


MaDill:
### Description of the Issue
After a search or when press F7 the search results window is opened. It would be nice to close it again with F7. Maybe an option to allow this toggle can be added. At the moment I use a Autohotkey script for this, but it would be nice if this is integrated already

### Expected Behavior
Press F7 toggle the search window. If shown, then close it. If closed, then open it.

### Actual Behavior
F7 only show the window. When it is already there, nothing happened.

### Debug Information
Notepad++ v7.3.1   (64-bit)
Build time : Jan 17 2017 - 00:28:43
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7
Plugins : mimeTools.dll NppConverter.dll 

The link to the autohotkey script: http://superuser.com/questions/700357/create-a-hotkey-keyboard-shortcut-to-close-the-notepad-find-results-window

HerbM:
Second and bump this please.

The AutoHotKey solution is not a great choice (even though I habitually run AHK) because it is very difficult to get it to run (with same key) through a wide variety of remote desktop and screen sharing software.

Please, this would be small but constantly significantly useful feature.

Close search without breaking flow by moving hands from keyboard.

JJT319:
Same as #3766. 
I would think we do not want to change the existing now expected behaviour of the menu item 'Search Results Window' Default map 'F7'.
I would suggest a workable solution would be to add a menu item 'Hide Results Window' Default map 'Ctrl+F7'.
I would happily code this if the admin approves the issue.

sasumner:
@JJT319 Does your proposal consider adding keyboard-based closure of the "Search Results Window" (aka the window with **_Find result_** in its title bar -- not sure why this window had/needs two different names) when input focus is in THAT window?

JJT319:
@sasumner apologies, I am not quite sure exactly what you are asking. (New to codeTriage, hope I am doing this right)

First of all yes I am not sure why it has two different names, probably would be nice to have one consistent name and also in the code. 
Adding to that, it is strange that the menu item 'Search Results Window' is clickable before the window is created. This is probably a topic for another issue, but essentially the object is not created until a search in all files is performed and so the menu item should be greyed out until this has happened or it should open an empty window. Whereas currently prior to a search nothing happens on menu click, or 'F7', 

Back on this topic.
[...]

PR #8245, opened 2020-05-10
Add menu item to hide search results window

JJT319:
Resolves issue #2946 
Adds a new menu item to "Hide Search Results Window". Has a corresponding keyboard shortcut mapping. 
Renamed menu item "Search Results Window" to "Show Search Results Window" to clarify purpose and fit in with new menu item.


donho:
I don't like the idea to add a new command just for that. Let's make ESC key to hide Search Results Window when it is focused.

PR #8248, opened 2020-05-10
find incremental dialod: add support of extended search and regexp se…

### LINKED ISSUE #8247: incremental find dialog: add support of regexp and extended search mode
### Description of the Issue

The incremental find dialog only support a normal search. Add the support of extended mode (\r \n \t ....), regexp mode, regexp mode with dot that match new line. 
In another world, add support of every mode that classical find supports.

Also increase the width of the serach field cause it's frequently asked by the community

### Expected Behavior
Add a dropdown menu to select the desired mode.
image

### Actual Behavior
No mode is configurable today
image

### Debug Information

Notepad++ v7.8.5   (32-bit)
Path : C:\notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1909
OS Build : 18363.778
Plugins : AutoEolFormat.dll CodeAlignmentNpp.dll Explorer.dll ExtSettings.dll FileSwitcher.dll HexEditor.dll LightExplorer.dll LightExplorer2.dll mimeTools.dll NavigateTo.dll NppConverter.dll NppExport.dll NppFTP.dll NppJumpList.dll NppMenuSearch.dll NppPluginTemplate.dll nppplugin_svn.dll NppSaveAsAdmin.dll NppTextFX.dll PythonScript.dll SherloXplorer.dll WindowManager.dll

cmeriaux:
Fix #8247

The incremental find dialog only support a normal search. Add the support of extended mode (\r \n \t ....), regexp mode, regexp mode with dot that match new line.
In another world, add support of every mode that classical find supports.

Also increase the width of the serach field cause it's frequently asked by the community

screencast: https://youtu.be/lnfJBeb2RkU


cmeriaux:
@sasumner  yes this is a good idea.
another idea is to add the incremental search parameters to the saved xml configuration so that parameters are persistent.

donho:
> Since this would be first changes to Incremental Search in a long time, maybe it should also add language translation capability for existing controls and new?

@sasumner  OK, go ahead please.
Thank you

donho:
@cmeriaux 

If user wants regexpr, they can use Find dialog.
The incremental search is aimed for a minimal interface to provide a quick search. So this feature is pointless to me.

cmeriaux:
@donho  using a regexp could be tricky and a nightmare. It's very usefull to have a instantly.
Have you look the sceencast ? This feature is pretty handy

You can check the thread https://community.notepad-plus-plus.org/topic/12722/survey-incremental-search-usefulness/41 This is what community likes

Ekopalypse:
One, for me, important reason why I prefer incremental search to the find dialog is,
that I have to work with only one monitor more often than I like.
In these situations, the Find dialog is not optimally usable.
Either I have to close the dialog after each search or I always have to move it. 
The incremental search panel has a clear advantage here.

sasumner:
I started putting my comments about this into the "issue" ( #8247 ) because I think that is more where they belong, but if everyone else wants to use the "PR" comments to discuss it, then OK.

@Ekopalypse 

>I have to close the dialog after each search or I always have to move it.

Maybe this is why Visual Studio has _both_ kinds, the larger more full-featured one and this smaller one that is reasonable to always have open:

image

But this sounds more like a different feature than the one currently under discussion?
Because of the nature of incremental search, I mean.

Ekopalypse:
>Maybe this is why Visual Studio has both kinds

@sasumner - maybe it is and VSCode has only the little one with the incremental feature.

image

sasumner:
Incremental search just seems so, well, sudden.  :)
Interesting how these other programs choose `.*` as the button text for regex as in general that regex is a "bad idea".

sasumner:
Side note:  I use incremental search, but I never noticed until just now that Shift+Enter moves backward through matches (like the regular Find dialog).  Maybe I am just late to the party on discovering this. :)

Ekopalypse:
Btw. Sublime and Atom do both use incremental regex search and also having the caption set as .*. 
image

image

Notepad2 and Geany seem not to have such a feature, I mean the little find panel.

sasumner:
Just thinking here, maybe what others are thinking:  What if incremental search were reworked a bit more to be a minimal height (already is!) find interface with the incremental part--the immediate match jump--being optional.  It would have the other features of course, similar to the Visual Studio / VSCode, meaning little (minimal size) buttons to set case, word, etc.  If the "jump" is optional, an initial Enter press after editing the text field would be required.  Thinking about it, this PR does indeed go down this road, so maybe I'm adding nothing new. :)

Ekopalypse:
For me personally, it doesn't matter what the button says. It's a matter of getting used to it. 
The bigger the input box, the better.

sasumner:
@Ekopalypse And buttons can certainly have tooltips!

asvc:
Just wanted to note the existence of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8278 which is directly relevant to this pull request.

As rightfully noted by @sasumner , there are two caveats:
1. Using “Highlight All” makes N++ scroll the viewer despite that valid match is already present on the screen. Potentially related to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3640 et al?
82169299-39621380-9900-11ea-8700-339c28824943

2. Not all matches are highlighted with "highlight all" enabled. Test snippet:
    ```javascript
    {
        "guid": "{xx-xx-xx-xx}",
        "name": "Command Prompt",
        "commandline": "cmd.exe",
        "icon": "ms-appx:///ProfileIcons/{xx-xx-xx-xx}.png",
        // "colorScheme": "Campbell",
        "antialiasingMode": "cleartype",
        "closeOnExit": "graceful",
        "cursorShape": "bar",
        "fontFace": "Consolas",
        "fontSize": 10,
        "hidden": false,
        "historySize": 9001,
        "padding": "8, 8, 8, 8",
        "snapOnInput": true,
        "startingDirectory": "%USERPROFILE%",
        "useAcrylic": false
    }
    ```
    Only the first one is highlighted immediatly and only two are highlighted after pressing “Enter”:

    1589770153522-npp_regexp


Minus the caveats above, it does what it says. @cmeriaux , do you think these are because of the introduced changes or due to the existing Scintilla / Boost bugs?

sasumner:
@asvc said:

>Using “Highlight All” makes N++ scroll the viewer despite that valid match is already present on the screen. Potentially related to #3640 et al?

Why do you think there's a connection between scrolling the viewer and backward-regex?
I mean, maybe there is, and I just don't see it?

cmeriaux:
@donho as @asvc request a similar feature, is it OK for you to reopen this ticket ?

asvc:
/ping @sasumner .

It's been a month, is there a chance we can get back to this discussion? Especially in light of https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8248#issuecomment-626689047

KnIfER:
The bar is long enough to put in another functionality ——

What about "incremental/persistent" goto line? The default Goto-Line is just an one-shot dialog.


image (3)

znsoooo:
@cmeriaux I came from another issue (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11902) and read throughout everyone's opinions.
I'm disappointed that it's not considered as a useful feature. But I agree with you that this feature would be useful for quickly validating regular expressions.
Do you have any released portable versions? I am not good at build program. Can you release it in your fork or send it to me? I am very interested and thank you very much! :)

cmeriaux:
@znsoooo  you can have a try by downloading a 64b portable version , then replacing the notepad++.exe by this one (built by CI/CD tools from my fork) https://ci.appveyor.com/project/cmeriaux/notepad-plus-plus/builds/44218065/job/hf64p34gtipya8nl/artifacts

znsoooo:
@cmeriaux Thank you very much! It even work on 32 bit portable version, I dont know why.

PR #8252, opened 2020-05-11
DansLeRuSH dark theme (20200511)

DansLeRuSH:
Hi team,

I resubmit my dark theme after a correction on too much "-" in comments and with a better name for the branch as seen with Don.

Best regards,
Franck

### DIFF
--- a/PowerEditor/installer/themes/DansLeRuSH-Dark.xml
+++ b/PowerEditor/installer/themes/DansLeRuSH-Dark.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="Windows-1252" ?>
 <!--// DansLeRuSH's dark theme for Notepad++ / Notepad plus plus
-----------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


dail8859:
If your expectation is to have this theme available via the installer, just dropping the xml file in that directory is not enough.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nsisInclude/themes.nsh Needs updated as well.

donho:
@dail8859 Nice catch. I'll update it if this PR is accepted.

DansLeRuSH:
> 
> 
> If your expectation is to have this theme available via the installer, just dropping the xml file in that directory is not enough.
> 
> https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nsisInclude/themes.nsh Needs updated as well.

OK so sorry to missed this (not yet very familiar with all the process) ...

> 
> 
> @dail8859 Nice catch. I'll update it if this PR is accepted.

... and thank you for that ! 🖖😃

PR #8259, opened 2020-05-12
Update russian.xml to 7.8.7

GOLEM777:
add new lines


PowerEditor/installer/nativeLang/russian.xml
```diff
@@ -1220,6 +1220,14 @@ Updated to v7.8.6:
...
 			<summary-nbsel2              value=" байт) в "/> 
 			<summary-nbrange             value=" диапазоне(ах)"/>
-		</MiscStrings>
-	</Native-Langue>
+            <replace-in-files-confirm-title      value="Вы уверены?"/>
+            <replace-in-files-confirm-directory  value="Вы точно хотите заменить все совпадения в каталоге :"/>
+            <replace-in-files-confirm-filetype   value="Для типа файла(ов) :"/>
+            <find-result-caption                 value="Результаты поиска"/>
+            <find-result-title                   value="Поиск"/>
+            <find-result-title-info              value="($INT_REPLACE1$ совпадения в $INT_REPLACE2$ файлах из $INT_REPLACE3$ найденных)"/>
```
sasumner, line 1228: @donho 

Can we request that users NOT supply translations for new strings until AFTER they appear in a released version?
Or not doing translations for a period of time after new strings are introduced into the code?

Reasons:

I think there is a bug with `find-result-title-info` in `english.xml` and if translators copy that bug to do "early" translations then it becomes a mess.

The bug is output can appear like this, in English:
`(5 hits in 2 files of 3 searched) - Line Filter Mode: only display the filtered results`
instead of:
`(5 hits in 2 files of 3 searched - Line Filter Mode: only display the filtered results)`
It is a minor bug but it creates an inconsistency between untranslated and translated formatting.

Also, I was going to do a new feature that would require a slight adjustment to this string's name:
Instead of calling it `find-result-title-info` I would change it to `find-result-title-info-files`.  
Again, for this type of thing if we have very early translations it creates trouble.

GOLEM777: Okay! I track changes in the code. I will be make changes.

sasumner: Maybe my request to wait was not so good.
Software changes and contains mistakes.
We'll just have to deal with it. :-)

artem78:
You overtake me. I wanted to translate these strings too.

donho:
@GOLEM777 
Please add the following change:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/93f78533334bb11af006e1171cc45d42848346ef

Thank you

PowerEditor/installer/nativeLang/russian.xml
donho: @sasumner  The only way to prevent translators from translating until the new release is to not release the new modification in `english.xml`. The disadvantage is that the translation unit could be forgotten. And also it's not really efficient - AFAICT some translators follow the commits on master to make their translation.

artem78:
Please fix word duplicate: "**без без** возможности пути"

PR #8267, opened 2020-05-14
Fix loadFileData() does not detect error properly issue

### LINKED ISSUE #3381: FileManager::loadFileData() does not detect error properly
### Description of the Issue
FileManager::loadFileData() will return success if the fread() returns 0 due to failure. fread() does not distinguish between end-of-file and error, and application should be checking feof() or ferror() to detect which condition occurred. NP++ doesn’t seem to do this and assumes eof.

This resulted in a "succesful" open of a blank document when a network file was able to fopen() but the fread() failed. User should have been given an error that NP++ failed to read the file properly.

1. Open a file where the fopen() will succeed, but fread() will fail
2. File opens "successfully" but is blank
3. No way for user to know if they truly successfully opened an empty file or if there was an error.
4. I suppose this could manifest itself at any point fread() failed, i.e. could get a "partial" file opened.

### Expected Behavior
User is given an error.

### Actual Behavior
NP++ did not report an error and just presented a blank document.

### Debug Information
Notepad++ v7.3.3   (32-bit)
Build time : Mar  8 2017 - 03:37:37
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll PluginManager.dll SpellChecker.dll

Dimon4eg:
show error if read file fails:
image

Fix #3381

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -1351,6 +1351,11 @@ bool FileManager::loadFileData(Document doc, const TCHAR * filename, char* data,
 		do
 		{
 			lenFile = fread(data+incompleteMultibyteChar, 1, blockSize-incompleteMultibyteChar, fp) + incompleteMultibyteChar;
+			if (ferror(fp) != 0)
+			{
+				success = false;
+				break;
+			}
 			if (lenFile == 0) break;
 
             if (isFirstTime)


PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -1426,6 +1426,10 @@ bool FileManager::loadFileData(Document doc, const TCHAR * filename, char* data,
 		success = false;
 	}
 
+	const int error = ferror(fp);
+	if (error != 0)
+		success = false;
```
jurko-gospodnetic, line 1431: do this straight after the `fread()` call? as I expect that's what this is intended to check?

Dimon4eg: agree. done.

xylographe: Obviously, it is up to @donho, but I believe the original solution was more efficient as well as more elegant..

jurko-gospodnetic: not sure what you mean by more efficient and more elegant 😕 

- there is a new `error` variable introduced in long code without limiting its scope
- the error checking is far removed from the actual call, which together with 'check global state' error handling seems like a real bad thing
- there is no guarantee that the `lenFile == 0` check would catch the error as the function might return less than the requested number of bytes before failing
- efficiency seems a non-issue as we're checking a variable as opposed to doing I/O which is an incomparable cost

PR #8269, opened 2020-05-15
Disallow backward regex searches due to sometimes surprising results

### LINKED ISSUE #3640: Backward-direction checkbox enabled with Reg Exp search mode (at N++ startup)

sasumner:
Fixes #3640, although that is only part of the story. The full story, debated before, is that regular expression searching in a backward direction from the caret position causes matches that the user does not expect. The best thing that was decided to do (group decision) is to fully disable upward searching.

Note:  This is the corrected version of PR #8128 , which was hopelessly scrambled, or at least too messed up for me to figure out how to fix.  There was nothing wrong with the code in 8128, just the way the commit ended up due to a botched merge.

There aren't nearly as many lines of code changed as are shown below.  One more downside of github:  It's difference reports are poor.  BeyondCompare4 does a much better job at showing the real changes.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1022,23 +1022,31 @@ void Notepad_plus::command(int id)
 		case IDM_SEARCH_FINDNEXT :
 		case IDM_SEARCH_FINDPREV :
 		{
-			if (!_findReplaceDlg.isCreated())
-				return;
-
-			FindOption op = _findReplaceDlg.getCurrentOptions();
-			op._whichDirection = (id == IDM_SEARCH_FINDNEXT?DIR_DOWN:DIR_UP);
-			generic_string s = _findReplaceDlg.getText2search();
-			FindStatus status = FSNoMessage;
-			_findReplaceDlg.processFindNext(s.c_str(), &op, &status);
-			if (status == FSEndReached)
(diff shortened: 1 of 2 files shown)


sasumner:
Hello @guy038 

>My reaction to your decision to disable, once and for all, the search for regular expressions in reverse order !

Well, all I was really trying to do was to "unify" a user-interface where backward-regex searching has been _somewhat_ disabled ever since I started as a Notepad++ user.

My first feeling about it (as a developer, not as a user) was to clean up the UI so that backward regex searching was fully ENABLED in the UI.  Then someone spoke out against that idea, because of the odd matches that sometimes can occur for a user.  I was swayed by that; someone cared enough to comment about it.

So then I let it set for a long time.  Then I finally decided that I started something, so I should finish it.  So I did, with apparently mixed results.

Anyway, some advice for you so you are not "blindsided" in the future, by changes:  Monitor the activity here, and comment when you don't like something, or when you like something.  Feedback to issue ideas is good for any and all (potential) developers thinking about fixing/implementing a feature.  Had I known backward regex was such a big deal to you, development likely would have proceeded differently.

PR #8272, opened 2020-05-15
Update Turkish Localization

Tmp341:
Updated according to 924e4b2, 2b345c8, f637901, 2b713fb and 5a95153 commits. Hope nothing missed.

Btw, ``<replace-in-files-confirm-filetype value="For file type :"/>`` does that **file type** mean like ``txt`` or ``type this here``?


donho:
Please add the following change:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/93f78533334bb11af006e1171cc45d42848346ef

Thank you

donho:
It's timing! Thank you @Tmp341 for the quick fix.

> Btw, <replace-in-files-confirm-filetype value="For file type :"/> does that file type mean like txt or type this here?

Here file type is the type of file like *.txt or *.cpp.
Anything will you modify again?

Tmp341:
I want to check it again.

PR #8280, opened 2020-05-18
Fix Document map overlay stuck to first monitor

### LINKED ISSUE #1317: Document map overlay stuck to first monitor
Hello,

Moving Notepad++ to another monitor using the <kbd>Windows</kbd> + <kbd>Shift</kbd> + <kbd>►</kbd> shortcut causes the document map overlay to stay behind on the first monitor.

Here's a screenshot showing the issue:

npp_document_map_stuck

The overlay will stay there until I minimize, and then maximize the window.

Notepad++ 6.8.8

Uhf7:
Fixes #1317

This PR fixes the following effects:
- Document map overlay remains visible after minimize to tray.
- Document map overlay is left behind on current monitor after moving Npp with Windows+Shift+Arrow to another monitor.
- Document map overlay escapes to another monitor without Npp, when Windows+Shift+Arrow is pressed while the input focus is at the document map.


donho:
It fixes the bug but creates a regression:
Mouse-click doesn't move the orange rectangle under the mouse clicked position.

Uhf7:
I cannot verify this regression in my environment. A left-click into the Document Map window moves the orange window to a position centered around the mouse pointer in Y direction. It is the same (untouched) mechanism like before:

`ViewZoneDlg::run_dlgProc --> WM_LBUTTONDOWN` sends a `DOCUMENTMAP_MOUSECLICKED` to its `_hParent` and `DocumentMap::run_dlgProc` processes it.

What exactly went wrong? No reaction to mouse clicks at all?  Or do I have to try it from a certain initial situation?

donho:
@Uhf7  If your Notepad++ is not whole screen (reduce it to half), could you reproduce the regression?

donho:
@Uhf7 Just checked again:
In the master there's the same problem, so it's not a regression.
However, I feel it's more accurate in master (ie. mouse clicked is not at the center, but in the orange rectangle).

PR #8286, opened 2020-05-19
Fix Mouse remains in clicked state after resizing docked windows

### LINKED ISSUE #8285: Mouse remains in clicked state after resizing docked windows
If I resize a docked window and move the mouse pointer to the Document Map after this, then the orange window marking the current range of the document sticks to the mouse pointer as if the left mouse button was pressed.

1. Start Npp. 
2. Open a sufficiently large document. That means, it must have at least more lines than visible lines in the edit window.
2. View the Document Map.
3. Configure Document Map to be in a docked state (if it not already is).
4. Use the mouse to resize the Document Map.
5. Move the mouse pointer into the Document map, without any further mouse click.

### Expected Behavior
I can move the mouse pointer in the Document Map without any impact.

### Actual Behavior
The orange window, wich marks the current range in the document, sticks to the mouse pointer as if the left mouse button was pressed. The effect goes away as soon as I press and release the left mouse button anywhere within Npp.

### Further information
The effect is independent from which docked window is resized, to resize the Document Map itself is only the most simple way to reproduce it.

It is similar to #3380, but I cannot be sure that this one fully covers #3380.

### Debug Information
Notepad++ v7.8.6   (32-bit)
Build time : May 19 2020 - 13:16:33
Path : c:\Tools\Npp\Notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Plugins : none

Uhf7:
Fixes #8285.

The reason for the effect is the mouse hook procedure, which seems to be quite obsolete in contemporary systems. Look, what it does. It catches the mouse messages and posts them a second time to the same window. I assume, that was ment to speed up the shifting of the splitters on "prehistoric systems" (That's what you call every system not supporting transparent windows). I see no regression without this hook.

What happens is, the hook catches the `WM_LBUTTONUP` message and posts it to `DockingSplitter::runProc`. Here, the hook is released and `ReleaseCapture()` is called. All fine.

But Windows fails to update its internal key state, when the `WM_LBUTTONUP` comes as an posted-by-a-hook message. So, if you call `GetKeyState(VK_LBUTTON)` after `ReleaseCapture()` in the original code, then it is still in a pressed state. Which confuses the Document Map later.

If I should have overseen something about the usefulness of this hook procedure here, please let me know.

### DIFF
--- a/PowerEditor/src/WinControls/DockingWnd/DockingSplitter.cpp
+++ b/PowerEditor/src/WinControls/DockingWnd/DockingSplitter.cpp
@@ -35,37 +35,6 @@
 BOOL DockingSplitter::_isVertReg = FALSE;
 BOOL DockingSplitter::_isHoriReg = FALSE;
 
-static HWND		hWndMouse		= NULL;
-static HHOOK	hookMouse		= NULL;
-
-#ifndef WH_MOUSE_LL
-#define WH_MOUSE_LL 14
-#endif
-
-static LRESULT CALLBACK hookProcMouse(int nCode, WPARAM wParam, LPARAM lParam)
-{
-    if (nCode >= 0)
-    {
-		switch (wParam)
-		{
-			case WM_MOUSEMOVE:
-			case WM_NCMOUSEMOVE:
-				::PostMessage(hWndMouse, static_cast<UINT>(wParam), 0, 0);
-				break;
-
-			case WM_LBUTTONUP:
-			case WM_NCLBUTTONUP:
-				::PostMessage(hWndMouse, static_cast<UINT>(wParam), 0, 0);
-				return TRUE;
-
-			default:
-				break;
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


chcg:
@Uhf7 There are more hook like the one removed, see https://github.com/notepad-plus-plus/notepad-plus-plus/search?q=hookProcMouse&unscoped_q=hookProcMouse

donho:
@chcg 
Let's remove one by one per PR if needed.

Uhf7:
Yes, I have seen them too. But I don't want to remove all of them at once. And I don't want to remove this one actually, if it serves any hidden purpose. But otherwise, if not, then the software works better without it.

The reason I'm going for this one is the negative effect for the mouse button status, as described in the issue. And currently, this is the only mouse hook for which I can observe a certain negative effect, and I can remove this effect most effectively by removing the hook.

Issue #8294, opened 2020-05-20
Improve Document Map precision


Uhf7:
The orange rectangle in the Document Map is covering the same range in the Document Map as the range which is visible in edit view. This should be done more precisely.

Some points are:
- Upper edge of orange window is shifted 2 pixels toward the top.
- Lower edge of orange window does not cover the last line visible in edit view. This is especially visible with short files.
- Height of orange window changes when reaching the end of file, although edit window didn't change height.
- Wrapped view scrolls to wrong position if a line is very long. To reproduce this, turn word wrap of, copy/paste-create a line with 64000 characters or more, turn word wrap on and scroll downward in the document. The orange window will disappear.

PR #8295, opened 2020-05-20
Improve Document Map precision

Uhf7:
Improves #8294

To solve the 2-pixel offset problem, we have to use the actual Scintilla window for position calculation, because it has a `WS_EX_CLIENTEDGE` style attached to it, which shifts it contents. The new way to calculate the position is implemented in `DocumentMap::doMove()`.

The calculation of the current visible range is now based on the document positions in the corners of the edit window, and map scrolling is done by Scintilla itself now.

In non-wrapped views, the hight of the orange window is calculcated now on the basis of the height of the edit window: If there is only a half line visible at the bottom, than the orange window covers only a half line in the Document Map.

In wrapped views, the number of lines visible in the edit window can differ from the number of lines in the Document Map view for the same range of text, because wrapping can be different. Hence the size of the orange window can change in wrapped views, although the height of the edit window remains constant. This is tricky, and the wrapped view does not work as good as the non-wrapped view, but a little better than before.

### DIFF
--- a/PowerEditor/src/WinControls/DocumentMap/documentMap.cpp
+++ b/PowerEditor/src/WinControls/DocumentMap/documentMap.cpp
@@ -191,6 +191,7 @@ void DocumentMap::wrapMap(const ScintillaEditView *editView)
 		_pMapView->execute(SCI_SETWRAPINDENTMODE, pEditView->execute(SCI_GETWRAPINDENTMODE));
 
 	}
+	doMove();
 }
 
 int DocumentMap::getEditorTextZoneWidth(const ScintillaEditView *editView)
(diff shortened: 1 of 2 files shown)


Uhf7:
Sorry, it did not link to the issue automatically, "Improve" seems not to be a magic word.

sasumner:
>Sorry, it did not link to the issue automatically, 

I don't understand this.
Your first/opening comment appears fine and links to the issue.

>"Improves" seems not to be a magic word.

The "magic"(?) is in the `#8294` part -- github creates a link for you.
Is that the "magic" you mean?

The "unwritten rule" of the first line of a PR being short and making it easy to find the originating issue seems to have been followed here.  You can use "Fixes", "Enhances", "Implements", "Improves" or whatever, but they do no "magic".

Uhf7:
@sasumner,

it worked, when I wrote "Fixes #xxxx". And now it didn't work, when I wrote "Improves #xxxx". So I thought, "Fixes"is a magic word, and "Improves" isn't. And it seems not to react on "#xxxx" alone either.

sasumner:
@Uhf7 

Sometimes it doesn't "do its magic" while you are composing in the "Write" tab, but if you switch to the "Preview" tab it should do the auto-linking thing.

"Fixes" and "Improves" are not relevant, as far as I know.

donho:
@Uhf7 
I just realized (by checking the behaviour of your PR), the current code works fine for the precision, if the `Word Wrap` is not enabled - it's the same behaviour of your PR.

donho:
@Uhf7 
That's the bin of your PR:
https://www.youtube.com/watch?v=cFUU_hLXkLc

Uhf7:
@donho
As I wrote in my initial comment, wrapped views are tricky.

The upper edge of the orange window was already better in wrapped views. The lower edge is more inaccurate, because we have to decide for an integral line height here. But you are right, `SCI_POINTYFROMPOSITION` delivers the top of the line, so after adding one line height it looks better. 

Thank you for the video too, but actually I prefer screen shots, in which I can count the pixels more easily. This document map characters are so small.

Uhf7:
Current Document Map unwrapped:

DMap_Cur_4

New Document Map unwrapped:

DMap_New_4

The orange window in the current Document map (unwrapped) is
- is 4 Pixels too wide, 2 pixels left and 2 pixels at the right.
- starts 2 pixels higher then it should do, the line hight is 6 pixels, its over 30 %.

It does not look much better at the lower edge. The screen captures above are zoomed, here are the same ones in original size.

Current Document Map:

DMap_Cur

New Document Map:

DMap_New

The lower edge of the current Document Map in unwrapped view is very dependent from the size of the edit view and the size of the document. Here another example of the current Document Map, when it is at the end of a short document:

DMap_Cur2

The upper edge of the orange window doesn't look perfect here, the lower edge is definitely wrong. This is what I want to fix with this PR.

donho:
Just retried this PR. There is some problem in the PR comparing to the current behaviour:

This PR:
image

Current behaviour:
image

donho:
Bug mentioned aside, in this PR, the "windows" (orange rectangle) is resized to match editor zone, and a little more accurate, which is nice.
However, the bug in the wrap mode is still not fixed, which is more annoying than the small tuning IMO. 
Can this PR be improved to fix the inaccurate clicking bug in the wrap mode?

Uhf7:
@donho,
sorry for contradicting again, but there is an observation error here. I described this effect in the last paragraph of my initial comment to this PR. And if I say it's tricky, tricky it is.

But first things first, thank you for testing and thank you for the screen shots.

Now let's have a look at them. The second one shows the current (7.8.6) behavior. We see the first line in the edit view, line 75, is  the first line marked by the orange window of the Document Map. Off target 2 pixels, but the intention is good. I cannot say this for the last line marked by the orange window. In the edit window, it is line 94, the line immediately after a comment line. The last line in the orange window is not immediately after a comment line. Read the last paragraph of my first comment again and then look at the different wrappings. In the edit view, all comment lines are wrapped, in the Document Map, no comment line is wrapped. This makes the difference. Now count the comment lines in the edit view. I count 6. And then count the comment lines within the orange window of the Document Map. I count 10. See the problem? That's why the orange window in my PR is much shorter than in the original.

And, sorry again, but you don't care about marking text 2 pixels off target? In a text editor software? That's a statement!

donho:
After checking your PR again with my glasses, you're right about it.
The wrapped sentences on the map are not the same height than the height in editor zone.
I will see what can I do about it.

Issue #8311, opened 2020-05-24
Automatic file extension for session files and workspace files


Uhf7:
In _Preferences --> MISC_, two special file extensions can be specified. One for session files, and another one for workspace files. There is another setting in _Preferences --> Default directory_, named Use _new style dialog ..._.

When I save a new session file or a new workspace file, then the behavior of the **Save As** dialog is unnecessarily dependent of whether I use the _new style_ dialog or not. Using the standard **Save As** dialog, the extension of the session/workspace file is automatically appended, if specified in _Preferences --> MISC_. Using the _new style_ dialog, it is not.

I understand, that not automatically appending the extension is the designed behavior of the _new style_ dialog. This may be good for the user's document files, but it seems not to be so good for session and workspace files.

As soon as I specify an extension for session files or workspace files in _Preferences_, I modify the way in which these files are seen by Npp anyway, and from that point on, it is quite useless for me to store those files without their assigned extensions.

So, long story short, I want to append the extensions for session files and workspace files automatically in the _new style_ **Save As** dialog too, and only, if the extensions are specified in _Preferences_. That would it make easier to save new session or workspace files with the correct extension. (I prefer the _new style_ dialog for the list with my hard drives on the left side instead of the big icons.)

To facilitate the usability for the many users who are already used to type in those extensions manually, or select an existing file from the file list, it should be implemented in the same way as the Windows Notepad appends the **.txt** extension automatically: This extension is only appended, if it is not already there.

donho:
@Uhf7 
The inconvenient old style save dialog should be removed. It's kept because file extension adding feature and Unix style path capacity could not implemented in the new style dialog.
Now file extension adding feature is added by your PR. If Unix style path capacity can be done in the new style dialog, then the old style will be removed (no COM Object implementation) - that will improve user XP and make the code simpler.

PR #8312, opened 2020-05-24
Automatic file extensions for session files and workspace files

Uhf7:
Implements #8311.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
@@ -267,6 +267,7 @@ TCHAR * FileDialog::doSaveDlg()
 
 	NppParameters& params = NppParameters::getInstance();
 	_ofn.lpstrInitialDir = params.getWorkingDir();
+	_ofn.lpstrDefExt = _defExt.c_str();
 
 	_ofn.Flags |= OFN_OVERWRITEPROMPT | OFN_HIDEREADONLY | OFN_ENABLESIZING;
 
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -2217,6 +2217,7 @@ const TCHAR * Notepad_plus::fileSaveSession(size_t nbFile, TCHAR ** fileNames)
 			sessionExt += TEXT(".");
 		sessionExt += ext;
 		fDlg.setExtFilter(TEXT("Session file"), sessionExt.c_str(), NULL);
+		fDlg.setDefExt(ext);
 		fDlg.setExtIndex(0);		// 0 index for "custom extension types"
 	}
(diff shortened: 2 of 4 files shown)


PowerEditor/src/WinControls/OpenSaveFileDialog/FileDialog.cpp
```diff
@@ -267,6 +267,7 @@ TCHAR * FileDialog::doSaveDlg()
 
 	NppParameters& params = NppParameters::getInstance();
 	_ofn.lpstrInitialDir = params.getWorkingDir();
+	_ofn.lpstrDefExt = _defExt.c_str();
```
donho, line 270: **Very interesting this line**: It's fixed the problem that auto-adding file ext (by changing file type) does not work in new style dialog.
I was searching to fix this problem but the only solution I found was using ugly COM Object.
I still don't get why and how this line is the solution so @Uhf7 if you have any explanation please let me know.

Uhf7: It's actually a rather complex mechanism of **GetOpenFileName** and **GetSaveFileName**.

If the field `_ofn.lpstrDefExt = NULL`, then there is no default extension specified, and if there is no default extension, then no extension is appended automatically ever.

If the field `_ofn.lpstrDefExt` points to an empty string (`_ofn.lpstrDefExt = L""`) , the behavior changes: Now, the selected extension from the _file types drop down_ is automatically appended. And, if there is nothing selected in the _file types drop down_ (`"*.*"`), then `_ofn.lpstrDefExt` is appended, which is in this example nothing.

If the field `_ofn.lpstrDefExt` points to a not-empty string (`_ofn.lpstrDefExt = L"npj"`), then the selected extension from the _file types drop down_ is automatically appended, regardless of the `"npj"` in `_ofn.lpstrDefExt`. Only if nothing is selected in the _file types drop down_ (`"*.*"`), then the default ending is appended, which would then be `".npj"`.

A point to be watched carefully is the handling of the starting dot of the default extension. If there is an empty string specified as default extension, then no dot is added. If there is a not-empty string specified as default extension, then the dot is inserted automatically. If the specified default extension starts with a dot already, then we have a double dots. Means for this project: We could enforce a no-dot policy for the default session and workspace filenames in _Preferences-->MISC_.

Another point I realized while writing this is, that I forgot to improve the load procedures for workspace and session files too. The line `_ofn.lpstrDefExt = _defExt.c_str();` should also be included into `FileDialog::doOpenSingleFileDlg();`. Then the **Open Workspace** function works automatically, the **Open Session** function needs a `fDlg.setDefExt(ext)` call too at loading. Could force-push a new Version of this PR easily with this modifications.

donho: @Uhf7 Thank you for the very detailed explanation.

> that I forgot to improve the load procedures for workspace and session files too. 

Ah, too late. Creating a new PR will be fine. Thank you again for fixing this bug.

Issue #8314, opened 2020-05-25
Display file name as root of workspace instead of "Workspace"


Uhf7:
If I open a new workspace, no file is associated to it, so it looks like this:

![Ws_New](https://user-images.githubusercontent.com/40629737/82841360-0efef000-9ed6-11ea-9912-e38cb6e94990.png)

After I open a workspace file, it looks like this:

![Ws_File_Ws](https://user-images.githubusercontent.com/40629737/82841396-2e961880-9ed6-11ea-8f48-aefee2890b78.png)

Wouldn't it be nice to display the workspace file name instead of "Workspace"? Like this:

![Ws_File_Fn](https://user-images.githubusercontent.com/40629737/82841446-5e452080-9ed6-11ea-9803-b7088e862036.png)

It's only a very small code modification, if you would consider this to be good too. As long as there is no file associated, it remains "Workspace". As soon as a file is associated, the filename is displayed.

sasumner:
Is it best to keep the extension on it, or only show `ws1` ?
Would it be good to show its complete path as a tooltip for mouse hover?

Uhf7:
The complete path is already shown as tool tip.

I find, if the extension is omitted, then it is not so easy to identify it being a file name. It looks then like an editable title, same as "Project1" is.

sasumner:
>I find, if the extension is omitted, then it is not so easy to identify it being a file name. It looks then like an editable title, same as "Project1" is.

Seems reasonable; keep the extension.

PR #8315, opened 2020-05-25
Display file name as root of workspace instead of "Workspace".

Uhf7:
Implements #8314

### DIFF
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -398,7 +398,8 @@ bool ProjectPanel::openWorkSpace(const TCHAR *projectFileName)
 
 	NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
 	generic_string workspace = pNativeSpeaker->getAttrNameStr(PM_WORKSPACEROOTNAME, "ProjectManager", "WorkspaceRootName");
-	HTREEITEM rootItem = _treeView.addItem(workspace.c_str(), TVI_ROOT, INDEX_CLEAN_ROOT);
+	TCHAR * fileName = PathFindFileName(projectFileName);
+	HTREEITEM rootItem = _treeView.addItem(fileName, TVI_ROOT, INDEX_CLEAN_ROOT);
 
 	for ( ; childNode ; childNode = childNode->NextSibling(TEXT("Project")))
 	{
@@ -453,6 +454,9 @@ bool ProjectPanel::writeWorkSpace(TCHAR *projectFileName)
     HTREEITEM tvRoot = _treeView.getRoot();
     if (!tvRoot)
       return false;
+
+	TCHAR * fileName = PathFindFileName(projectFileName);
+	_treeView.renameItem(tvRoot, fileName);
 
     for (HTREEITEM tvProj = _treeView.getChildFrom(tvRoot);
         tvProj != NULL;


Uhf7:
Jumped too short. Need two additional code lines to display the correct filename after **Save Workspace As** too.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -398,7 +398,8 @@ bool ProjectPanel::openWorkSpace(const TCHAR *projectFileName)
 
 	NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
 	generic_string workspace = pNativeSpeaker->getAttrNameStr(PM_WORKSPACEROOTNAME, "ProjectManager", "WorkspaceRootName");
```
MAPJe71, line 400: Doesn't your change make this line (400) obsolete?

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -454,6 +455,9 @@ bool ProjectPanel::writeWorkSpace(TCHAR *projectFileName)
     if (!tvRoot)
       return false;
 
+	TCHAR * fileName = PathFindFileName(projectFileName);
+	_treeView.renameItem(tvRoot, fileName);
+
```
MAPJe71, line 460: 1. Wrong indentation characters used.
2. Change to `TCHAR * fileName = PathFindFileName(fn2write);` (see first two lines of method).

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
Uhf7: That's observed correctly. Thank you. I should remove it.
The translation itself is used now only in `ProjectPanel::newWorkSpace()`.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
Uhf7: For filename: Thank you again, this one would crash after the _...workspace was modified. Do you want to save it?_ request. I should be more careful.

For indentation: In most of the code lines in this project, tabs are used for indentation.  In this function, however, in 18 lines spaces are used, and in 9 lines tabs. I added two lines with tabs, because I thought to make it equal to the majority of code lines. This has the unfortunate effect of the wrong indentation in this preview. In my IDE, it looks fine. But I can change to spaces here too.

Question remaining: spaces or tabs?

sasumner: >Question remaining: spaces or tabs?
> I thought to make it equal to the majority of code lines.

I think you did the right thing.

>wrong indentation in this preview. In my IDE, it looks fine.

Yes, this happens sometimes.

Uhf7: @MAPJe71 

Since the PR made it into the master without the modifications you recommended, I checked if the crash I was afraid of would really happen. It does not. `PathFindFileName` returns `NULL`, if `NULL` is given as parameter, and `treeView.renameItem(tvRoot, fileName)` returns immediately without action, if `NULL` is specified as `fileName`.

The function `ProjectPanel::writeWorkSpace` is only called with `NULL` as parameter before Npp ends or before a new workspace is loaded, so that luckily not even a wrong display name is left behind.

Thank you again for watching, such things can go wrong.

PR #8322, opened 2020-05-27
Update Slovak translation

kubalav:


donho:
`kubalav:master` cannot be used. You should create a new branch for per PR. See:

Issue #3747, opened 2017-09-18
Unicode characters inconsistently cannot be displayed in Notepad++


charliehoward4dp:
### Description of the Issue

In a Notepad++ document that is encoded as UTF-8 (no BOM), many Unicode characters are not displayed, but the hollow square appears in their place.  If a displayable Unicode character is added to a line containing undisplayable Unicode characters, those undisplayable ones suddenly appear. Removing the "good" one makes the others revert to the hollow square.  A simple example:

☆◬⊗⊠⋆⧆⨂

Paste that line into NP++ and you will see all the characters.  Remove the leading star ☆ and the others become squares.  Restore the star and the others re-appear.
### Steps to Reproduce the Issue

1. Create a UTF-8 (no BOM) text file.  (This is the only hard part of the procedure.)
2.  copy & paste the following string into a UTF-8 (no BOM) Notepad++ document: ☆◬⊗⊠⋆⧆⨂
3.  all of those characters will display properly
4. delete the leading star ☆
5. the other characters become hollow squares
6. restore the ☆ and the other characters reappear
### Expected Behavior

All of the characters always should appear.
### Actual Behavior

They only appear if an always-acceptable Unicode character is on the same line. If an always-acceptable Unicode character is in the document but not on the same line, certain Unicode characters, such as, but not limited to, the ones shown above, will not be displayed properly.

### Debug Information

Notepad++ v7.5.1   (32-bit)
Build time : Aug 29 2017 - 02:35:41
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll NppTextFX.dll PluginManager.dll SpellChecker.dll 

This occurs with characters from many of the Unicode blocks.

dennis-thomas-o:
I was able to reproduce this as well.

I tested with the *Default Style* in *Style Configurator* set to `Courier New`, `Consolas`, `Arial` and `Times New Roman`.

I believe this issue would happen any time you enter a character that is **NOT** contained in the selected font and then add/remove on the same line a character which **IS** contained in the selected font.

IMHO seems like something not quite right with the font-substitution routines.

### Debug Information

Notepad++ v7.5.9   (64-bit)
Build time : Oct 14 2018 - 15:19:55
[...]

Issue #2287, opened 2016-09-15
Ligature support


Graucsh:
### Description of the Issue

Fonts that support ligatures are not drawn with them
### Steps to Reproduce the Issue
1. Select a font that supports ligatures, such as Fira Code
2. Select a C-based language
3. key in "!="
### Expected Behavior

a long equals sign with a slash through it
### Actual Behavior

discrete characters
### Suggestion

Set the Technology property of the Scintilla objects to DirectWrite, which draws with ligatures, or provide an option to turn it on.

![default](https://cloud.githubusercontent.com/assets/22226368/18569157/f666bb42-7b6e-11e6-96e8-cd6b940e3e6d.png)
![directwrite](https://cloud.githubusercontent.com/assets/22226368/18569193/28329178-7b6f-11e6-8013-28223d189218.png)

dail8859:
Duplicate of #1651 

I looked into this at one point and just using `SC_TECHNOLOGY_DIRECTWRITE` is not enough. Notepad++ uses Scintilla v3.5.6 and it seems that a newer version of Scintilla is required to support ligatures. So once Scintilla is updated internally this will most likely be possible.

Graucsh:
I added the following line to void Notepad_plus::loadBufferIntoView(BufferID id, int whichOne, bool dontClose):
::SendMessage(viewToOpen->getHSelf(), SCI_SETTECHNOLOGY, SC_TECHNOLOGY_DIRECTWRITE, 0);

I realize that's probably not the right place, but I got the following result using the Fira Code font:

![npp_ligatures](https://cloud.githubusercontent.com/assets/22226368/19666062/31076d1c-9a16-11e6-961d-01dac6ca62df.png)

dail8859:
That's interesting. I tried the exact same thing before without any success. Though I tried it on Win7 which might be a difference.

Graucsh:
Maybe it makes a difference where the message is sent?
![npp_ligatures_win7](https://cloud.githubusercontent.com/assets/22226368/19691255/a9898a32-9aa1-11e6-9862-0f669d243f42.png)

I put the code after:
MainFileManager->addBufferReference(id, viewToOpen);

dail8859:
This is what I tried:

- Install the Fira Code font (I think I used the medium)
- Set the global Notepad++ font to use Fira Code
[...]

PR #8326, opened 2020-05-27
Allow optional use of Scintilla's DirectWrite technology.

Uhf7:
Implements #3747. Solves a problem involving incorrectly displayed characters, if those characters are not part of the current font.

Affects many other issues, like #2287, for example.

The problem seems to come up every some months, and is solved until now for insiders only with the help of a plugin. May be there are even multiple plugins out there which solve this problem.

But I, as normal user here, didn't have even the idea of searching for such a plugin to solve this problem. And, see #3747 and others, I'm not alone with this.

The solution I suggest with this PR includes an additional setting in the _MISC_ page of _preferences_, named _Use DirectWrite technology_. 

Changing this option requires restart of Npp. Why? Because it is designed to preserve the previous behavior, if it is not checked. That means, no `SCI_SETTECHNOLGY` call is made, as long as the new option is unchecked. So any existing plug-ins using `SCI_SETTECHNOLOGY` behave like before.

For the same reason I refrained from making the change visible in the moment the check mark is clicked.

Thanks to @Ekopalypse who came up with the `SCI_SETTECHNOLOGY` function in issue #3747, I would have blamed it to Scintilla alone without much further searching.


dail8859:
What are the downsides of DirectWrite? Do some platforms not support it?

Ekopalypse:
Afaik, [XP is not supported] (https://docs.microsoft.com/en-us/windows/win32/directwrite/direct-write-portal) and it has been reported that on some systems the fonts do not look as sharp as with gdi.
BUT it is an optional setting, so both of these points are not actually affected.

dail8859:
Which is the exact reason why I asked. If it was widely supported then you could do away with yet another option for a user to be aware of. But if there is a reason to be able to control that then this PR make sense

Uhf7:
`ScintillaDoc.html` says, it's supported from Windows Vista on.

The implementation of `SCI_SETTECHNOLOGY` in `ScintillaWin.cxx` reveals, that the command does nothing, if the initialization of Direct2D fails. So technically, the worst thing I expect to happen is, that after checking the new option everything remains as it is (If Direct2D in the version required by Scintilla is not available on the target system).

@dail8859, if you ask why I made it optional, this is for 3 reasons:
- I have tested it on two systems  only: Windows 10 and Windows 7. I'm convinced that the option helps on this systems, but I cannot be sure about any other system.
- Strange behaviors can occur using Windows programs on Windows remote desktops or in other remote controlled sessions, TeamViewer for example.
- There seem to be plugins out in the world, which use the `SCI_SETTECHNOLOGY` function too, mainly to set the technology to `SC_TECHNOLOGY_DIRECTWRITE`. But I wanted to have chance to avoid conflicts with those plugins. Imagine a plugin setting technology to `SC_TECHNOLOGY_DIRECTWRITEDC` instead, then this could raise a conflict.

So, making it an option means having a save fallback, after turning off the option you can be sure that nothing is worse than before.

xylographe:
When **`Technology`** is set to **`DirectWrite`** ([`SCI_SETTECHNOLOGY(SC_TECHNOLOGY_DIRECTWRITE)`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETTECHNOLOGY)), should **`BufferedDraw`** be turned off ([`SCI_SETBUFFEREDDRAW(0)`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETBUFFEREDDRAW)) ?

Uhf7:
No, let's leave it as it is. It's default _on_.

Take _on_ as a synonym for "flicker-free". Yes, it may be "flicker-free" with DirectDraw anyway, but on the other hand, the performance loss by rendering a text line into a bitmap first and then transferring it to the screen instead of transferring it to the screen directly, cannot be so much that it is worth the effort. And the performance loss should be smaller than with the classic GDI technology anyway.

So, turning `SCI_SETBUFFEREDDRAW` off seems to me like calling for additional negative effects, and not like increasing the performance.

ValZapod:
Wait a second. Why it was not originally DirectWrite if it still has fallback? Also, can we @ somenody from Scintilla? Also I support fixing #2287.

Uhf7:
@ValZapod
I cannot follow you here. What do you mean exactly?

Uhf7:
Added some Windows Vista screenshots to #3747, which seem to support the necessity to make the DirectWrite technology feature configurable.

Uhf7:
I see this "technology" by now as a choice between two completely different rendering engines. The default engine comes with some negative effects, and the DirectDraw engine too. The user should decide, which effects he can live with best. 

I stumbled over an effect with composed characters.

The following picture shows the cursor moving through line 23. The 3rd character in the line is  "-ْ", composed of a Minus sign combined with U+0652, just for illustration (I used that character set, because it was 1st in the list). 

Now, there is a different behavior with cursor positions between DirectDraw technology and default technology, looking like this:

CursorInCharWithDirectDraw

- With default technology, the cursor is at the same position in columns 3 and 4, directly before the composed character. 
- With DirectDraw technology (red frame), the cursor is positioned wrong in column 3.

donho:
@Uhf7 Could you solve the conflict please?

ValZapod:
Just wanna propose this UI, maybe it is more cool? Okay, if no))

KOLANICH:
I don't think it is cooler, just garbages GUI. One usually has to almost never change this pref.

sasumner:
Rest assured, the _Preferences_ aren't going to be given a different look.  I don't know where ideas like that come from.

PowerEditor/src/Parameters.h
```diff
@@ -835,6 +835,7 @@ struct NppGUI final
 	bool _backSlashIsEscapeCharacterForSql = true;
 	bool _stopFillingFindField = false;
 	bool _monospacedFontFindDlg = false;
+	bool _useDirectWrite = false;
```
donho, line 838: There are 4 different possible values (0-3) and users could ask for adding 2 & 3 in the future.
So using an enum variable `_writeTechnologyEngine` with 2 values (0 & 1) might preserve the settings interface in `config.xml`. So if there's any change in  the future, we just need to change the GUI without touching the `config.xml`

PowerEditor/src/Parameters.cpp
```diff
@@ -5919,6 +5923,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("fileSwitcherWithoutExtColumn"), _nppGUI._fileSwitcherWithoutExtColumn ? TEXT("yes") : TEXT("no"));
 		GUIConfigElement->SetAttribute(TEXT("backSlashIsEscapeCharacterForSql"), _nppGUI._backSlashIsEscapeCharacterForSql ? TEXT("yes") : TEXT("no"));
 		GUIConfigElement->SetAttribute(TEXT("monospacedFontFindDlg"), _nppGUI._monospacedFontFindDlg ? TEXT("yes") : TEXT("no"));
+		GUIConfigElement->SetAttribute(TEXT("useDirectWrite"), _nppGUI._useDirectWrite ? TEXT("yes") : TEXT("no"));
```
donho, line 5926: I would use `writeTechnologyEngine` instead of `useDirectWrite` for 0 or 1 value here.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -126,6 +126,7 @@ BEGIN
     CONTROL         "Enable Notepad++ auto-updater",IDC_CHECK_AUTOUPDATE, "Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,37,143,217,10
     CONTROL         "Don't fill find field in Find dialog with selected word", IDC_CHECK_STOPFILLINGFINDFIELD, "Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,37,158,217,10
     CONTROL         "Use Monospaced font in Find dialog (Need to restart Notepad++)",IDC_CHECK_MONOSPACEDFONT_FINDDLG, "Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,173,325,10
+    CONTROL         "Use DirectWrite technology (Need to restart Notepad++)",IDC_CHECK_DIRECTWRITE_ENABLE, "Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,188,325,10
```
donho, line 129: Can we find another term more user friendly than "Use DirectWrite technology"?

KOLANICH: IMHO we should mention DirectWrite explicitly anyway. Maybe `Render text using DirectWrite (may be beneficial for performance and quality)`

Uhf7: Actually, I don't know a better name. But I'm open for suggestions.

I agree with @KOLANICH, that we should display, what we are doing, and not something obscure. But I also worry about the length of the name a little.

So my second thought was `Use DirectWrite API (Need to restart Notepad++)`.

donho: "Render text using DirectWrite (may be beneficial for performance and quality)" sounds better.
Agree.

sasumner: @Uhf7 said:  "I ... worry about the length of the name a little."

@donho suggests: _Render text using DirectWrite (may be beneficial for performance and quality)_

Conclusion:  Don's not worried about the length!  :-)

Is it reasonable to do?: 
_Render text using DirectWrite_ 
and then as a hover tooltip?:
_May be beneficial for performance and quality; need to restart Notepad++_

donho: Good idea @sasumner !
What do you think @Uhf7 ?

donho: Maybe it's better, for aligning with `Use Monospace font in Find dialog (need to restart Notepad++)`:
```
Render text using DirectWrite (need to restart Notepad++)
and then as a hover tooltip?:
May be beneficial for performance and quality
```

sasumner: A downside is that it creates two separate things to language translate.

donho: @sasumner  It's a very little sacrifice if it's more clear for the users :)

Uhf7: I don't find many tooltips in the preferences menu. Could this be the 1st one? And if, do we really want to start this? I could create a 2nd line, for example, instead of the tooltip.

Uhf7: Better text: _Use DirectWrite (May improve rendering special characters, need to restart Notepad++)_. Improving performance is a too vague promise perhaps.

donho: @Uhf7  OK, sounds good enough.
Please make one line label - We have enough room. and please make it as long as possible even in English it doesn't take all room: the translation takes more room than the original text generally.

PR #8330, opened 2020-05-28
Update slovak.xml

kubalav:


donho:
@kubalav  Sorry about rejecting your PR again.
`kubalav:patch-1` has been used before and it makes a error when I try to integrate your commit:
```
$ git checkout -b kubalav-patch-1 master
fatal: A branch named 'kubalav-patch-1' already exists.
```
I could delete your branch of course but my goal is keep the workflow from impact.
So please use the date of PR as branch name like `20200528`.

PR #8336, opened 2020-05-28
Fix Document Map regression under Windows 7

### LINKED ISSUE #8335: Document Map regression under Windows 7
I recently created some PR's which fixed some issues with the Document Map, in which the transparent window marking the current text range within the document has been detached from the rest of Npp and was displayed then on another virtual desktop or another physical desktop than Npp itself. This modifications, however, only work under Windows 10. They destroyed the Document Map under Windows 7 and Windows Vista. I realized this only now while testing another issue, see #3747. I feel guilty. I want to send my apologies to all Windows 7 users with the upcoming PR, which keeps all advantages of the new behavior under Windows 10, but reverts them for Windows 7. The technical reason for it is, that Windows 7 does not support layered style for child windows.

Uhf7:
Fixes #8335.

The following effects where always there in the versions before my PR's and are still there under Windows 7 now, but not under Windows 10:
- Transparent window of Document Map is displayed above the Find/Replace dialog.
- Transparent window of Document Map is left behind if Npp is minimized to tray.
- Transparent window of Document Map can detach from Npp in multi monitor environments. Fortunately, there are no virtual desktops under Windows 7.

The increased precision of the orange window remains increased under Windows 7 with this PR.

Did I mention already, that I find the Document Map tricky?


donho:
@Uhf7 
> Did I mention already, that I find the Document Map tricky?

Yes, you did. Have you a better solution?

Uhf7:
>  Have you a better solution?

No. Since it works fine now under Windows 10, and as good as before under Windows 7, I will not go for a better solution. Much too tricky :-)

PR #8368, opened 2020-06-03
make cpp the first extension for C++

pryrt:
since save-as appends the first extension from the list to all files, the first extension for C++ should be cpp, not h


donho:
Only for C++ ?

pryrt:
I went briefly through all the other languages.  The languages I have exposure to (languages that I "know", to some extent or another) all used extensions that I would consider "default" for the language in the first slot.  The other languages all started with an extension that seemed reasonable for the name.

Issue #8361, opened 2020-06-02
[Project] Double clicking a file doesn't move focus to edit area


ArkadiuszMichalski:
The same problem as for [Folder as Workspace](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4656). There are two ways to open a file from Project:
1. Double click a file
3. Select a file and press Enter

But double click method doesn't move focus to the edit area. The focus should move to the edit area or  maybe this behavior for double left click is intentional?

PR #8369, opened 2020-06-03
Fix input focus not in edit window after double click

### LINKED ISSUE #4656: Folder as Workspace: Double clicking a file doesn't move focus to edit area
### Description of the Issue
As far as I know, there are three ways to open a file from Folder as Workspace:

1. Double click a file
2. Right click a file -> Open
3. Select a file and press Enter

Of these, the double click method is the only one that doesn't move focus to the edit area.

### Expected Behavior
The focus should move to the edit area.

### Actual Behavior
Focus remains in the Folder as Workspace area.

Uhf7:
Fixes #4656 and #8361 

This one bothers me too.

Replace `SendMessage` with `PostMessage`, and it works.

As always with `PostMessage`, the life time of the parameters is important.

In the `ProjectPanel` class, the filename parameter already lives long enough in `fullPathStrs`.

In the `FileBrowser` class, I had to add a new member `_fullPath` instead of leaving the filename at the stack.

### DIFF
--- a/PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
+++ b/PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
@@ -176,6 +176,8 @@ protected:
 	HMENU _hFileMenu = NULL;
 	std::vector<FolderUpdater *> _folderUpdaters;
 
+	generic_string _selectedNodeFullPath; // this member is used only for PostMessage call
+
 	std::vector<SortingData4lParam*> sortingDataArray;
 
 	void initPopupMenus();
--- a/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
+++ b/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
@@ -513,15 +513,15 @@ void FileBrowser::openSelectFile()
 	HTREEITEM selectedNode = _treeView.getSelection();
 	if (!selectedNode) return;
 
-	generic_string fullPath = getNodePath(selectedNode);
+	_selectedNodeFullPath = getNodePath(selectedNode);
 
 	// test the path - if it's a file, open it, otherwise just fold or unfold it
-	if (!::PathFileExists(fullPath.c_str()))
+	if (!::PathFileExists(_selectedNodeFullPath.c_str()))
 		return;
-	if (::PathIsDirectory(fullPath.c_str()))
+	if (::PathIsDirectory(_selectedNodeFullPath.c_str()))
 		return;
 
-	::SendMessage(_hParent, NPPM_DOOPEN, 0, reinterpret_cast<LPARAM>(fullPath.c_str()));
+	::PostMessage(_hParent, NPPM_DOOPEN, 0, reinterpret_cast<LPARAM>(_selectedNodeFullPath.c_str()));
 }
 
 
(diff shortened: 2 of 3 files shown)


sasumner:
You forgot double-clicking on a node in the `Function List` window! :-)
(Okay, maybe that one is a bit different -- but still, that one is my pet peeve!

donho:
@Uhf7  Good job.
My question is why PostMessage makes the difference?

PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
```diff
@@ -176,6 +176,8 @@ class FileBrowser : public DockingDlgInterface {
 	HMENU _hFileMenu = NULL;
 	std::vector<FolderUpdater *> _folderUpdaters;
 
+	generic_string _fullPath;
```
donho, line 179: Could you change this line to the following:
```
generic_string _selectedNodeFullPath; // this member is used only for PostMessage call
```

Uhf7: > My question is why PostMessage makes the difference?

Short answer: Because the **TreeView** double click handler claims back the input focus after calling the double click notification procedure.

Long answer: You cannot debug it with Spy++, but by including `OutputDebugString` calls into the source code of Npp. If you write a debug message starting with '++' before certain calls and a 2nd debug message starting with '--' after the same calls, then it gives the following picture:

	SendMessage
	===========
	001 ++TreeView WM_LBUTTONDBLCLK          <===== POSTED
	002   ++ProjectPanel::notified(NM_DBLCLK)
	003     ++SendMessage(_hParent, NPPM_DOOPEN)
	004       ++Notepad_plus::process(NPPM_DOOPEN)
	005         TreeView WM_KILLFOCUS
	006         ScintillaEditView WM_SETFOCUS
	007       --Notepad_plus::process(NPPM_DOOPEN)
	008     --SendMessage(_hParent, NPPM_DOOPEN)
	009   --ProjectPanel::notified(NM_DBLCLK)
	010   ScintillaEditView WM_KILLFOCUS
	011   TreeView WM_SETFOCUS
	012 --TreeView WM_LBUTTONDBLCLK

	PostMessage
	===========
	101 ++TreeView WM_LBUTTONDBLCLK          <===== POSTED
	102   ++ProjectPanel::notified(NM_DBLCLK)
	103     ++PostMessage(_hParent, NPPM_DOOPEN)
	104     --PostMessage(_hParent, NPPM_DOOPEN)
	105   --ProjectPanel::notified(NM_DBLCLK)
	106 --TreeView WM_LBUTTONDBLCLK
	107 ++Notepad_plus::process(NPPM_DOOPEN) <===== POSTED
	108   TreeView WM_KILLFOCUS
	109   ScintillaEditView WM_SETFOCUS
	110 --Notepad_plus::process(NPPM_DOOPEN)

First shown is the sequence of events using `SendMessage`. The `NPPM_DOOPEN` command is processed during the `SendMessage` call, the input focus is transferred to Scintilla in lines `005` and `006`, and is claimed back by  **TreeView** in lines `010` and `011`, before the processing of the double click message ends.

The second sequence shows what happens using `PostMessage`. `PostMessage` returns immediately, leaving the `NPPM_DOOPEN` command in the message queue of Npp. The processing of the double click message by **TreeView** ends without focus fighting, and after that, the `NPPM_DOOPEN` command is processed as intended. The input focus switching at lines `108` and `109` cannot be contested anymore by the **TreeView** double click handler.

Important to note is, that only the events marked as POSTED in the sequences shown above come from the message queue. All other actions are plain calls, `SendMessage` performs nothing else than a plain call of the Window procedure of the window specified.

donho: @Uhf7 Thank you for the very detail explanation!

PR #8370, opened 2020-06-04
Change focus to edit window after activating function in function list

### LINKED ISSUE #4356: no focus on double click at funktionlist
### Description of the Issue
when i make a double click ant a funktion in the functionlist, the editview shows the function but the editview becomes no focus. I have to go first click into the editview to edit the function or scroll at the view.

1.  open a *.c File
2. double click a function at the function list
3. press a arrow key to navigate at the editview

### Expected Behavior
navigate at the editview

### Actual Behavior
navigate at the functionlist

-> the editview becomes no focus by a double click a function at the functionlist

Uhf7:
Fixes #4356.

Makes function list behave like "Project Panel" and "Folder As Workspace" with almost zero effort: If a function is choosen with return key or double click, then the focus switches to the edit window.

### DIFF
--- a/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
+++ b/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
@@ -557,6 +557,7 @@ void FunctionListPanel::notified(LPNMHDR notification)
 			case NM_DBLCLK:
 			{
 				openSelection(treeView);
+				PostMessage(_hParent, WM_COMMAND, SCEN_SETFOCUS << 16, reinterpret_cast<LPARAM>((*_ppEditView)->getHSelf()));
 			}
 			break;
 
@@ -570,7 +571,9 @@ void FunctionListPanel::notified(LPNMHDR notification)
 					{
 						HTREEITEM hItem = treeView.getSelection();
 						treeView.toggleExpandCollapse(hItem);
+						break;
 					}
+					PostMessage(_hParent, WM_COMMAND, SCEN_SETFOCUS << 16, reinterpret_cast<LPARAM>((*_ppEditView)->getHSelf()));
 				}
 			}
 			break;


sasumner:
What about caret placement in edit window?

Uhf7:
As far as I see, it is positioned at the 1st character of the function name.

This is same like before, done in

 `FunctionListPanel::openSelection` ==> `(*_pEditView)->scrollToCenterPos`,

the first action of `scrollToCenterPos` is `execute(SCI_GOTOPOS, pos)`, which moves the caret to the start of function name.

PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
```diff
@@ -557,6 +557,7 @@ void FunctionListPanel::notified(LPNMHDR notification)
 			case NM_DBLCLK:
 			{
 				openSelection(treeView);
+				PostMessage(_hParent, WM_COMMAND, SCEN_SETFOCUS << 16, reinterpret_cast<LPARAM>((*_ppEditView)->getHSelf()));
```
MAPJe71, line 560: Curious...why left shift 16?

Uhf7: Because `SCEN_SETFOCUS` is expected in the high word of `wParam` in the Big Switch. It's a Scintilla-defined non-standard `WM_COMMAND` message. Now, since you asked, I could have used 

	PostMessage(_hParent, NPPM_INTERNAL_SWITCHVIEWFROMHWND, 0, reinterpret_cast<LPARAM>((*_ppEditView)->getHSelf()));

too, it does absolutely the same and looks even better.

MAPJe71: Or use `MAKEWPARAM`-macro:
```cpp
PostMessage(_hParent, WM_COMMAND, MAKEWPARAM(SCEN_SETFOCUS, 0), reinterpret_cast<LPARAM>((*_ppEditView)->getHSelf()));
```

Uhf7: That's, what you mean. It wouldn't have appeared to me in a dream to use `MAKEWPARAM`, but you are right. So, if I would change this one again, I would go for the `NPP_INTERNAL_SWITCHVIEWFROMHWND` message. But in the future, I will use the `MAKE*PARAM` macros for such things, if possible. Thank you for the hint.

Yaron10:
Just to complete this PR.

If the Function List has focus and you close it via `View -> Function List` or pressing its toolbar button - the editor should get the focus.

If the Function List is not focused, do nothing (let the focused element keep it).

***

In `NppCommands.cpp`

After
```
if (_pFuncList && (not _pFuncList->isClosed()))
{
```
Add
```
if(::IsChild(_pFuncList->getHSelf(), ::GetFocus()))
    ::SetFocus(_pEditView->getHSelf());
```
***
**Folder as Workspace should also be fixed.**

PR #8374, opened 2020-06-04
Fix Load project file by dragging doesn't work

### LINKED ISSUE #8324: [Projcet] Load project file by dragging doesn't always work
After setting project extension (in Preferences) dragging project file doesn't always work. Step to reproduce:
1. Open Notepad++ and open empty project panel.
2. Drag project file to Notepad++ main window.

Another example:
1. Open Notepad++ (project panel is hide).
2. Drag project file to Notepad++ main window. << this work
3. Drag another project file to Notepad++ main window. << this not work

Looks like dragging project file works only when we start Notepad++ without activate any project panel. After activate this panel (even when we immediately closing it) dragging project file will only activate project panel (without process this file). It doesn't make sense because project file is not processed and is not even open for edit.

Uhf7:
Fixes #8324 

To say it clearly, this never worked even before my PR #8161.

The only exception was: If no Project Panel was stored in `config.xml` before starting Npp, then it worked exactly one time. My PR #8161 wrecked even this. So I felt obligated to fix the drag and drop issue too.

The drag and drop operation as well as the command line use Project Panel 1 as target. I extended the safety question which occurs when a modified workspace file is being replaced by another workspace file to the drag and drop operation. This made it necessary to transfer this safety request into a new function.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5999,10 +5999,9 @@ void Notepad_plus::launchFileBrowser(const vector<generic_string> & folders, boo
 
 void Notepad_plus::launchProjectPanel(int cmdID, ProjectPanel ** pProjPanel, int panelID)
 {
+	NppParameters& nppParam = NppParameters::getInstance();
 	if (!(*pProjPanel))
 	{
-		NppParameters& nppParam = NppParameters::getInstance();
-
 		(*pProjPanel) = new ProjectPanel;
 		(*pProjPanel)->init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf());
 		(*pProjPanel)->setWorkSpaceFilePath(nppParam.getWorkSpaceFilePath(panelID));
@@ -6039,6 +6038,11 @@ void Notepad_plus::launchProjectPanel(int cmdID, ProjectPanel ** pProjPanel, int
 
 		(*pProjPanel)->setBackgroundColor(bgColor);
 		(*pProjPanel)->setForegroundColor(fgColor);
+	}
+	else
+	{
+		if ((*pProjPanel)->saveWorkspaceRequest())
+			(*pProjPanel)->openWorkSpace(nppParam.getWorkSpaceFilePath(panelID));
 	}
 	(*pProjPanel)->display();
 }
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -388,8 +388,9 @@ private:
                                               // then WM_ENDSESSION is send with wParam == FALSE
                                               // in this case this boolean is set true, so Notepad++ will quit and its current session will be saved 
 
-    bool _isWorkspaceFileLoadedFromCommandLine = false;
-
+    bool _isWorkspaceFileLoadedFromCommandLine = false; // Set during Notepad_plus::doOpen if workspace file is opened.
+                                                        // But it is only evaluated during startup, when doOpen
+                                                        // has been called while command line interpretation.
 	ScintillaCtrls _scintillaCtrls4Plugins;
 
 	std::vector<std::pair<int, int> > _hideLinesMarks;
--- a/PowerEditor/src/Notepad_plus_Window.cpp
+++ b/PowerEditor/src/Notepad_plus_Window.cpp
@@ -247,6 +247,10 @@ void Notepad_plus_Window::init(HINSTANCE hInst, HWND parent, const TCHAR *cmdLin
 	}
 	else if (_notepad_plus_plus_core._isWorkspaceFileLoadedFromCommandLine)
 	{
+		// Switch back to Project Panel 1, when a workspace file has been specified in the command line.
+		// This code is executed only once in lifetime of the process, at the first initialization. It is necessary, because
+		// the Project Panels are not loaded by Notepad_plus::doOpen only, but also by the Plugin Manager restoring the state
+		// of the last session from config.xml.
 		::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1, 0);
 	}
 
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -212,7 +212,9 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive, 
 	{
 		nppParam.setWorkSpaceFilePath(0, longFileName);
 		_isWorkspaceFileLoadedFromCommandLine = true;
-//		command(IDM_VIEW_PROJECT_PANEL_1);
+		// This line switches to Project Panel 1 while starting up Npp
+		// and after dragging a workspace file to Npp:
+		command(IDM_VIEW_PROJECT_PANEL_1);
 		return BUFFER_INVALID;
 	}
 
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -863,6 +863,35 @@ HTREEITEM ProjectPanel::addFolder(HTREEITEM hTreeItem, const TCHAR *folderName)
 	return addedItem;
 }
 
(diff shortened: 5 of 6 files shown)


Uhf7:
Checked this PR again, because I don't want to mess up things too deeply. The PR leads to a persistent behavior of Npp now, what I worry about is, do we want it behave this way in the first place?

The question is about workspace files and session files. As soon as there is an extension specified for both them, they are treated special by Npp:

1. If I specify a session file in the command line (, or if I double-click a session file in the Explorer), not the session file itself is opened, but the session stored in it.
2. If I specify a workspace file in the command line (, or if I double-click a workspace file in the Explorer), not the workspace file itself is opened, but the Project Panel 1 switches to the specified workspace file.
3. If I specify a session file in the `File` menu `Open Ctrl-O`, not the session file itself is opened, but the session stored in it.
4. If I specify a workspace file in the `File` menu `Open Ctrl-O`, not the workspace file itself is opened, but the Project Panel 1 switches to the specified workspace file.
5. If I drag a session file to Npp, not the session file itself is opened, but the session stored in it.
6. If I drag a workspace file to Npp, not the workspace file itself is opened, but the Project Panel 1 switches to the specified workspace file.

This behavior is perfectly fine for me. But some questions may arise:
- How can I open a session file in the edit window now?
- How can I open a workspace file in the edit window now?

I think, this was the wish behind issue #518. The problem of using command line options to open project or session files is, that double clicking in Explorer wouldn't work anymore.

So one could argue, now referring to my numbered list above, that only in case 1 the session should be loaded, and in case 3 and 5 the session file itself should be opened in the edit window. There is still the way over the `File` menu command `Open session ...` to open a session file from within Npp.

And one could argue, that only in case 2 the workspace should be loaded into Project Panel 1, and in case 4 and 6 the workspace file itself should be opened. There is still the context menu in the Project Panel to open a workspace file from with Npp.

Again, the current behavior with specified session file and workspace file extensions is fine with me, but if you want to go the other way, this PR could be modified quite easy to reach this.

The current behavior (7.8.6) should not be kept, it is too inconsistent for workspace files.

xylographe:
I believe, your new proposal (open session/workspace file for editing in cases 3, 4, 5 and 6, instead of loading the session/workspace) would offer a more understandable/predictable behaviour. Of course, I can only speak for myself, but when, for example, opening a session file via **Ctrl+O** (case 3) leads to session being loaded rather than opening the file for editing, this behaviour would surprise me, and I surely would wonder how to edit this session file in NPP without renaming (or linking) it.

ArkadiuszMichalski:
I think 1/2 can't completely replace 5/6 because what if I don't make file associations and double clicks on session/project file don't work? Dragging (as load/process this two type of files) is neccessary in this case. Theoretically, we could drag the project directly to the project panel (to load) and to the edit area (for editing), but for the session file there is no separate panel.
What, e.g. if dragging with some key pressed will open the file for editing (e.g. with Ctrl, Shift or other). This will not change current behavior, it will only improve it a bit. I ask because the same problem will apply to `Folder as Workspace` when we drag the entire folder, it would be nice to be able to add this folder to `Folder as Workspace` panel or open all its files (without using the Preferences window to change that behaviour).
In general, before accepting this PR, it is better to analyze all the pros and cons more carefully so that you do not change it once again in the future.

Uhf7:
> I think 1/2 can't completely replace 5/6 because ...

I cannot understand this concern, because the existence of Windows file associations for workspace or session files affects only, which program is started if someone tries to open the file. If no program is associated with the workspace or session files, then this files cannot be opened immediately with the Explorer by pressing enter or by double-click, a selection box with possible applications will pop up.

If we want to open a file with double click from the Explorer, we need a file association to make sure, that Npp is started. And within the Preferences of Npp, we decide, how to handle this file. If an extension for session files is specified, and the file has this extension, then it is opened as session. If an extension for workspace files is specified, and the file has this extension, then it is opened as workspace. In all other cases, it is opened as "normal" text file.

Dragging a file to a link to `Notepad++.exe` on the desktop, or even to `Notepad++.exe` itself in a folder, works like the command line `Notepad++.exe` with the file name dragged to it as the first parameter. So this is covered by case 1 and 2. This applies also for the case, that there is already an instance of Npp open and someone drags a file to the link on the desktop.

Dragging as such does occur only if Npp is running and someone drags a file into its area. This dragging operation works independently of any Windows file associations and is controlled by Npp alone.

> In general, before changing this behavior it is better to analyze all the pros and cons more carefully

In general, I would agree with you, but this PR does not go for the final and best and perhaps never coming solution, including various different drop zones inside the Npp window. It goes for a for a bad effect: Project file by dragging doesn't work. And if one looks closer, one can see how bad it doesn't work (7.8.6, it changes the filename of the workspace in Project Panel 1 without switching the workspace itself to this filename, leaving Project Panel 1 in an inconsistent state). So let's go for a consistent behavior here first, in an easy-to-implement way, before extending the functionality further and introduce more effects. But I'm open to any way. Except leaving it inconsistent.

ArkadiuszMichalski:
@Uhf7 
Of course that your PR is better than the current situation, which corrects the bug I reported. You just asked an additional question about a possible change in the current behavior, so I referred to it, and also to the opinion of @xylographe. I use dragging to process these files and eventually change this behavior to opening them for editing would be painful for me, unless there would be some extra drag-and-drop option for editing - like with dragging with pressed additional key or other solution.

I use many different portable version of NPP at once (through `doLocalConf.xml`), without any file associations in system, and setting extensions for session/project + dragging is better than selecting this files through NPP  window (we don't have to go through many directories). That's why I would not potentially be in favor of changing dragging behavior, and that's all I wanted to add:)

Uhf7:
>  I use dragging to process these files and eventually change this behavior to opening them for editing would be painful for me

I see you need the dragging like it is. Fine for me too. What about changing behavior in case 3 and 4 only? One could say, that it is actually expected to open a file for editing with `File`, `Open`. And there is an extra menu command `Open Session File ...`, and the workspace file can be changed using the context menu of the Project Panel easily.

Changing intended behavior is hypothetical anyway, before we have @donho's opinion about it. But I brought up this issue here, because it fits into the range of program code touched by this PR. And sometimes, I edit workspace and session files myself, using Windows Notepad.

I didn't consider nor test so far, what happens, if you drag a folder to Npp. The `Folder as Workspace` issue seemed too far away from this issue for me. There cannot be special extensions defined in the Preferences menu for folders, and I don't believe that you can associate Folders with Npp. So this one works technically in another way and can be (perhaps) improved in an extra effort.

xylographe:
> What about changing behavior in case 3 and 4 only?

I think, that would be a good compromise as it would maintain consistent **File > Open** behaviour (i.e. open selected file(s) for editing independent of file extension).

ArkadiuszMichalski:
>What about changing behavior in case 3 and 4 only?

From my point of view, it doesn't matter because I hate choosing files through the window and I try to avoid it. But even if (hypothetically) the dragging behavior changes I can live with it (if it is better for 99% of users).
 
I focus on dragging because you nice describe all possible way to operate on session/project file, and dragging is what interests me the most. Like @xylographe said dragging session/project file (when we set extension for them) and load/process them (not open for edit) for me it is also sometimes problematic, but on the other hand it is also convenient for load/process this file (as I mentioned above). I use dragging often for load session/project file - not for edit them. 

Returning to `Folder as Workspace`, it also has a similar conflict when it comes to dragging. Depending on the settings by dragging we can only add folder to `Folder as Workspace` or open folder's content (all nested files to edit). It is not possible to add folder to `Folder as Workspace` or open folder's content for edit by dragging (without switching option). It is very similar to project/session file, we can't load or open this file by dragging (without changing extension in settings). That's why I mentioned it, if anyone would consider resolving this problem with dragging in the future, keep in mind that this applies to these 3 things and could probably solve it with one consistent behavior.

But first let this PR pass, and any further (wilder) changes (like for dragging) maybe I discuss in a new bug. I don't want to make too much noise here.

donho:
I agree with @Uhf7 - once an user extension is defined, Notepad++ open the session or the project - for the sake of consistence. The session and project configure files are not supposed to be edited manually. 
If a defined conflict happens (for example, a non session or project configure file with the defined extension), user can always remove the defined extension from Preferences dialog temporarily to open it.

PR #7710, opened 2019-12-07
Update catalan.xml to v7.8.2

Hiro5:
Time to update the catalan #translation ... 7 months already? :)
Added some missing strings.

Regards to everybody,
Hiro5.

PR #8384, opened 2020-06-06
Update catalan translation to 7.8.7

xeanhort:
- Sort catalan.xml file like english.xml file to track changes easier.
- Removed ids: 24001, 6209, 6212, 6115, 6307
- Updated ids: 23242, 6857, 2902
- Added ids: 6237, 4008, 6115, 6348, GUpProxyConfNeedAdminMode, ColumnSize, AsciiInsertion/PanelTitle, replace-in-files-confirm-title, replace-in-files-confirm-directory, replace-in-files-confirm-filetype, find-result-caption, find-result-title, find-result-title-info, find-result-title-info-extra, find-result-hits.
- Remove empty spaces from english.xml.

PR #8396, opened 2020-06-10
Update catalan.xml to v7.8.7

Hiro5:
Updated Catalan location for npp 7.8.7.

Sorry I didn't read #7710

Regards

### DIFF
--- a/PowerEditor/installer/nativeLang/catalan.xml
+++ b/PowerEditor/installer/nativeLang/catalan.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!--
 Catalan translation for Notepad++
-Updated 07.12.2019, v7.8.2
+Updated 10.06.2020, v7.8.7
 By Hiro5 <groccat at gmail>
 -->
 <NotepadPlus>
-	<Native-Langue name="Català" filename="catalan.xml" version="7.8.2">
+	<Native-Langue name="Català" filename="catalan.xml" version="7.8.7">
 		<Menu>
 			<Main>
 				<Entries>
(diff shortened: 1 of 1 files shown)


chcg:
@Hiro5 Could you have a look at #8384 from @xeanhort. Just one updated translation could be used. Maybe you could review your changes and agree how to proceed.

xeanhort:
First contribution here, but at #8384 there's the full description of the changes. I think that the main difference is the sorting.

Hiro5:
> First contribution here, but at #8384 there's the full description of the changes. I think that the main difference is the sorting.

Hi everyone,
The main reason of the catalan.xml order, it's because it's a Visual Order.
If you compare the order of the < Commands> section in english.xml, with the File menu in np++... you will see that some parts are in different order.
While in the catalan.xml it's in the same order as you can see it in the notepad++ app, visually.
Thanks to that, in the past, I was able to see some errors in english.xml file. Like repeated entries, items in the wrong place, and entries that were missing in the english.xml file, but they did exist in the source code, and I just had to add them.

All that said,
I don't mind to change it to the english.xml order, even if it's a big change :)
To me, both ways are OK. And both have their own reason to be that way.

I also can see that you added the comments translated, I removed them because they're not needed, and they were like redundancy: If I see < Menu>< Main>< Entries> ... I don't need a comment to tell me that those are the "< !-- Main Menu Entries -->".

I can see that there are some more differences,
maybe some old entry must be removed, or something else needs to be changed,
I'll check it out.
Thank you, @xeanhort :)

Regards.

xeanhort:
> If you compare the order of the < Commands> section in english.xml, with the File menu in np++... you will see that some parts are in different order.
> While in the catalan.xml it's in the same order as you can see it in the notepad++ app, visually.
> Thanks to that, in the past, I was able to see some errors in english.xml file. Like repeated entries, items in the wrong place, and entries that were missing in the english.xml file, but they did exist in the source code, and I just had to add them.

I've noticed it at the beginning, but I thought that the english version should be the one more updated, so I changed it to be able to detect changes easily comparing files.  I think it's better to keep all sorted the same way to track changes, but you should say it :)

> I also can see that you added the comments translated, I removed them because they're not needed, and they were like redundancy: If I see < Menu>< Main>< Entries> ... I don't need a comment to tell me that those are the "< !-- Main Menu Entries -->".

As I said, I just tried to mimic the format of english.xml to make it easier to compare and maintain 

> I can see that there are some more differences,
> maybe some old entry must be removed, or something else needs to be changed,
> I'll check it out.
👍

Issue #8398, opened 2020-06-10
7.8.7 new "Save As" feature **always** adds selected extension


vinsworldcom:
The new "Save As" feature of always adding the language extension (or default to "Normal text file (.txt)" to the "File name" in the "Save As" is great, but it may be too aggressive?  If I change the "File name" in the "Save As" to a different "unknown" extension (one not in the "Save As Type" list), the name in the "File name" should be saved, the "Save As Type" extension should **not** be added *back* onto the file when saved.

**Workaround**
When saving a file type that does not have an extension in the "Save As Type" box of the "Save As" dialogue, one has to manually either:

- select "All types (*.*)" in the "Save As Type" drop down

or 

- put the "File name" text in double quotes

in order for the "File name" with non-available extension to be saved as filename.non-available-ext.

**Description of the Issue**
Open a new file in Notepad++ and type some text.  File=>Save As..., the "File name" in the "Save As" dialogue shows as "new 1.txt" and the "Save As Type" is "Normal text file (.txt)".  Rename in the "File name" text box of the "Save As" dialogue to "mygedcomfile.ged" and press "Save" button.  File is saved as "mygedcomfile.ged.txt".  It should be "mygedcomfile.ged" without the additional ".txt" extension.

**Steps to Reproduce the Issue**
1. Open a new file in Notepad++ and type some text. 
2. File=>Save As...
3. Change the "File name" in the "Save As" dialogue to "mygedcomfile.ged"
4. Press "Save" button
5. File is actually saved as "mygedcomfile.ged.txt"

**Expected Behavior**
File is save only as "mygedcomfile.ged"

**Actual Behavior**
File is actually save as "mygedcomfile.ged.txt"

Notepad++ v7.8.7   (64-bit)
Build time : Jun  5 2020 - 02:38:56
Path : C:\usr\bin\npp64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1217
Current ANSI codepage : 1252
[...]

PR #8404, opened 2020-06-11
Allow unregistered file types in Save As dialog.

Uhf7:
Here the possible solution for #8398 I described there for testing.


donho:
Not sure I will do somethings for #8398, but I'll check this PR.

Uhf7:
As I said in #8398, it's a Windows dialog box, featuring Windows behavior. That's fine with me. The code will be better without integrating this PR.

donho:
I prefer the current behaviour (while switching the filter user can see which ext is applied) than the one of this PR (nothing happen visually).  So let's keep the current behaviour.

PR #8409, opened 2020-06-12
Fix URL underline always black, invisible in dark themes

### LINKED ISSUE #8405: Bug: Styles: new 7.8.7 underline always black, invisible on dark themes
### Description of the Issue

Under 7.8.7's new link behavior, the underline for links is always black, and invisible on dark themes making no cosmetic link distinction unless you hover over the link.

1. In 7.8.7, select a dark theme (zenburn, twilight, choco, black board and so on)
2. Enter one or more links into a document
3. Observe that the link underline is no longer visible

### Expected Behavior

Clear visible link indicator.

### Actual Behavior

No visible link indicator.

### Debug Information

Notepad++ v7.8.7   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.900
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll ElasticTabstops.dll Explorer.dll mimeTools.dll NppConverter.dll nppcrypt.dll NppExport.dll NppFavorites.dll NppFTP.dll NppHasher.dll NPPJSONViewer.dll NppQrCode32.dll NppTextFX.dll PoorMansTSqlFormatterNppPlugin.dll pork2sausage.dll Tidy2.dll _CustomizeToolbar.dll 

2020-06-11 10_26_43-Window

### Request
Please add link styling controls. While allowing me to use the same color for the underline as the current font-color would likely be sufficient (this was the behavior in 7.8.6 and prior), I'd love the ability to assign the underline color or set a background color instead or completely style links along the same formatting as other elements in N++.

Uhf7:
Fixes #8405.

Measures to fix it:
- Underline color is adjusted to default text color now.
- Use `INDIC_BOX` style for hovering instead of `INDIC_FULLBOX`. This style draws a solid box around the link while hovering, the color of which can be configurated in the _URL hovered_ style, and this is better visible than the transparent `INDIC_FULLBOX` style, especially in dark themes.
- Call `addHotSpot` after resizing too. It is necessary to correct the underline color after showing the window, and it fixes another effect also: If the edit window is enlarged by dragging the window edge with the mouse, and the newly visible portions of the edit window contain clickable links, they where not immediately indicated.
- Call `addHotSpot` while processing `WM_UPDATESCINTILLAS`.

One related problem remains, which I don't know how to solve it correctly: The _URL hovered_ style is not included in all themes, and if a theme is active, in which the style is not included, I cannot change it. Is it allowed to "extend" themes?


donho:
Personally, I prefer `INDIC_FULLBOX` - more elegant and clear.

Uhf7:
Added one line without request: `execute(SCI_INDICSETALPHA, URL_INDIC, 70);`.

This boosts the visibility of the `INDIC_FULLBOX` style dramatically. Standard is 50. If the value is too high, text visibility will be impaired.

If you don't like it, I remove it again. But please try it (if you have not already).

Uhf7:
Almost forgot: The cursor flicker while moving slowly over the URL is described in https://sourceforge.net/p/scintilla/bugs/2170/. And it is almost gone in the current Scintilla version.

So a theoretical question arises: Would modifications to the modified Scintilla version used in this project possible or out of question?

donho:
> And it is almost gone in the current Scintilla version.

Hmm... it's fixed in the current version, or not. "almost gone" are strange words to describe.
If it's fixed in the current version, then I will find a timing (and version) to upgrade Scintilla.

> So a theoretical question arises: Would modifications to the modified Scintilla version used in this project be possible or out of question?

It's difficult to judge like that. I would say if it's a bug-fix, then rather doing a PR on Scintilla project, then Notepad++ will upgrade Scintilla. For the specific feature it's another story. If the feature is worth, then it's possible.

Uhf7:
> "almost gone" are strange words.

Yes, indeed. The cursor change on every mouse move is definitely gone. I believe to have seen occasional changes of the mouse cursor while moving fast over the URL, but I'm not sure whether they were triggered because I left the URL while fast-moving or not. It's way better in any case. Perhaps you can try it yourself with the original SciLexer.dll (4.4.3), the Npp extensions are not needed for this test.

Uhf7:
@Donho, sorry, but the flicker reappeared in 4.4.3 too. Cannot reproduce the flicker free state I saw. At least, 2170 of Scintilla is still open.

donho:
@Uhf7 OK. Thank you for the info.

Issue #8213, opened 2020-05-04
Order of Find All / Replace All buttons


nicolasnoakes:
### Position/order of Find All and Replace All buttons is likely to cause unintended mistakes

The logical sequence of the operational buttons on the "Find" and "Replace" tabs is not consistent, which will lead to users making unintended replacements.

### Steps to Reproduce the Issue

In the Find dialogue, the buttons on the "Find" tab are sequenced as follows, top to bottom:

1. Find Next
2. Count
3. **Find All in All Opened Documents**
4. **Find All in Current Document**
5. Close

But on the "Replace" tab, the logical order of the last operation buttons (buttons 3 & 4) is reversed.

Find 

1. Next
2. Replace
3. **Replace All** [in Current Document]
4. **Replace All in All Opened Documents**
5. Close

### Expected Behavior
I propose that it would be better and more logical (increasing scope from top to bottom) to swop buttons 3 & 4 on the "Find" tab. Thus the "in Current Document" and the "in All Opened Documents" buttons would be consistently positioned.

![notepad-find](https://user-images.githubusercontent.com/14802452/81004162-fc832080-8e19-11ea-8047-992729b8f355.png)

![notepad-replace](https://user-images.githubusercontent.com/14802452/81004186-06a51f00-8e1a-11ea-9398-bfb21b11ec48.png)

Uhf7:
Yes, I'am with you. And I would even go a step further and make it look like this:
![FindDlg001](https://user-images.githubusercontent.com/40629737/81048010-d3b36780-8ebb-11ea-9faf-571011c5995c.png)
Then the **_In Selection_** checkmark can work for both **_Count_** and **_Find All_**.
And **_Find All_** is a better match to **_Replace All_** than **_Find All in Current Document_**. 
And optically, it is easier to make a difference between the **_Find All_** button and the **_Find All in All Opened Documents_** button.

sasumner:
>Then the In Selection checkmark can work for ..... and Find All.

Note that currently *In Selection* has no effect on *Find All in Current Document* (a.k.a. *Find All* as proposed here).  It is simply a feature under consideration; see #8133

Uhf7:
I took the function itself as implemented in PR #8207. That is why I trust this function is possible. But I didn't look at the details of the PR right now.

nicolasnoakes:
[...]

PR #8413, opened 2020-06-14
Add capability to find-all in selected text of current document

### LINKED ISSUE #7927: Grouping box wrapping 'In-Selection' and 'Count' is misplaced

sasumner:
Implements #8133 , #8213 
Fixes #7927
Perhaps implements some of the spirit of #5089 
Perhaps implements some of the spirit of  this Community [THREAD](https://community.notepad-plus-plus.org/topic/17802/block-commands)

--------------------------------------------------

The implementation is to make _Find All in Current Document_ respect the _In Selection_ checkbox, and to change the text in the _Find result_ window to reflect that a search was done at the "selection" scope:

image

In the _Find_ window, to emphasize that _In Selection_  can apply to _Find All in Current Document_ , the groupbox UI element that contained _In Selection_ and _Count_ buttons is changed to contain all three of _In Selection_ and _Count_ and  _Find All in Current Document_ :

image

To make the containing groupbox approach feasible, it was needed to swap the positions of _Find All in Current Document_ and _Find All in All Opened Documents_ buttons.  While I don't like to change the user interface, it makes sense to do so in this case, with additional argument for doing it provided in #8123 .

--------------------------------------------------

The subject of this PR previously had two submitter-aborted tries (sorry for the noise on this):  #8207 and #8140


GOLEM777:
Does this really work with separate selection? 
Оr $INT_REPLACE3$ rudimentary and show number of files?

sasumner:
@GOLEM777 I have no idea what you are trying to ask with that.

Uhf7:
Only a suggestion, but I think, that the _Find_ dialog looks too much different compared to the _Replace_ dialog. My preferred look would still be as said in #8213 (_Find_ on the left side, _Replace_ on the right side):

FindDlg001

This would not change functionality, but the dialogues would fit much nicer together, because:

- The **Find next** buttons are the same and look the same, they are good as they are.
- The **Replace** button exists only in the _Replace_ dialog, naturally.
- The **Find all** button is the equivalent to the **Replace all** button and could look similar.
- The **... in all Documents** buttons and the **Close** buttons already look similar as they should.
- The **Count** button in the _Find_ dialog is a special function. It needs the **In selection** frame around it, so it could be placed like shown. PRO: The **In selection** frame remains much more compact.

And, since we swap the **Find all in Current Document** and **Find all in All Opened Document** buttons anyway, the slim design could help the users to adopt to the new button order, instead of accidentally hitting the wrong of the equal sized **Find all in ...** buttons, because they changed places.

donho:
I guess that's is what @GOLEM777 is talking about:
image
If my guess is right, it should not work with separate selection (like `Count` behaviour) and 
`N hits in 1 file of 1 searched` format should be kept as it is.

GOLEM777:
capture_06152020_195338

donho:
@Uhf7 Though I don't like the UI of this PR, at least it keeps the same logic (all the commands are in the same column on the right) IMO.

sasumner:
@Uhf7 I _did_ note your comments on UI.  However, I don't like to change existing UI except when absolutely necessary, because it is an unpopular thing to do (with users and author), even when it makes things better.  This PR changes UI only because I felt it had to.  Because it "had to", I took the suggestion in #8213 and that only in its minimal sense.

sasumner:
@donho said:

>I guess that's is what @GOLEM777 is talking about ,,, If my guess is right, it should not work with separate selection (like Count behaviour) 

Existing released Notepad++ behavior:  _In Selection_ is greyed when more than one selection is active.  Thus it is not possible to _Count_ in more than one selection, nor do the subject of this PR in more than one selection, no do _any_ Find/Replace/Mark operation when more than one selection is active.  

Doing what @GOLEM777 suggests is beyond the scope of the current PR, because it opens the door to having _Count_ handle multiple selections as well.  That functionality would be fine, except it should have its own issue, suggest: "Find/Replace/Mark functions should work on multiple selections".

>and `N hits in 1 file of 1 searched` format should be kept as it is.

Wouldn't that be misleading?  You could do two searches back-to-back, one with entire file scope, the other with selection scope, and they would look the same except for a different number of hits:

`Search "foo" (5 hits in 1 file of 1 searched)`
`Search "foo" (2 hits in 1 file of 1 searched)`

>Though I don't like the UI of this PR

What specifically bothers you here?
We can fix it, whatever it is.  :-)

Uhf7:
> at least it keeps the same logic (all the commands are in the same column on the right)

Yes, but one could say that **Count** is a special command here, the only one which does not do real action, except of displaying the number of results. It is a kind of preview of what is to expect when the **Find all** function beside it is executed. And, one could say, we have a huge empty white space wasted below the **Find what** input box anyway, why not use it? 

And, as I cannot say often enough (sorry for repeating), the _Find_ dialog and the _Replace_ dialog look like made by different persons on different days. It's like a car with a green door on the left and a blue door on the right, for me.

Uhf7:
>  I don't like to change existing UI except when absolutely necessary.

Yes, this strategy works if no UI change at all is necessary. Then one can say, I leave it (as bad) as it is, no harm done. 

But here, 
- the UI is changed anyway, 
- two nearly equal looking buttons are swapped,
- the _Find_ dialog is designed different to the _Replace_ dialog, optically even more after this PR than before,

so, why not try to pull it straight. Of course, its all subjective ...

sasumner:
So is there value in the scope of this PR, aside from the UI?
If the answer is No, we close it.
If the answer is Yes, then @donho tells me how to resolve this concern:

>Though I don't like the UI of this PR

It really is _that_ simple. :-)

donho:
@sasumner 
> What specifically bothers you here?

I just feel it's not pretty, but I have no better solution, so let's keep it as it is in PR.

>> and N hits in 1 file of 1 searched format should be kept as it is.

> Wouldn't that be misleading? You could do two searches back-to-back, one with entire file scope, the other with selection scope, and they would look the same except for a different number of hits:...

Not sure what you mean, but I what I meant for the expected result is what you have implemented.

sasumner:
@donho said:

>I just feel it's not pretty, but...

Maybe you mean the big groupbox with a lot of open space around _In selection_, and the checkbox being uncentered vertically?
Well, if you'll notice, what I did was made it so _In selection_ never moves on the three of four tabs that implement it.  For some reason, for me that is satisfying. :-)

Thank you for valuing the content of the PR, even if not the UI.  At work, we have to do a lot of searches of this type.  At least weekly some colleague will ask me how to search only within selected text with Notepad++.  Soon I will be able to tell them how to do this.

Issue #8393, opened 2020-06-09
Regular expression zero-length-match calltip text should be language translatable


sasumner:
### Description of the Issue

The "call tip" text that appears when a zero-length match occurs (when a regular expression search using _Find Next_ is executed) is currently hard-coded in Notepad++ source:

```
(*_ppEditView)->execute(SCI_CALLTIPSHOW, start, reinterpret_cast<LPARAM>("^ zero length match"));
```

This should be made language-translatable.

Perhaps the circumflex ( `^` ) and its following space should remain as fixed, and only the `zero length match` part should be affected.

### Debug Information
Tested in 7.8.6 but occurs in earlier versions as well.

sasumner:
Released in N++ version 7.8.8

PR #8414, opened 2020-06-14
Make regex zero-length-match calltip transatable

sasumner:
Implements #8393


donho:
@sasumner Could you remind me how to get regex zero length match result please?

sasumner:
@donho said:

>remind me how to get regex zero length

Put `(?=.)` into _Find what_ box.
Press _Find Next_.
This will give you a match (and the calltip) at any position in your document.
More _Find Next_ presses will advance the calltip one character at a time.

PR #8420, opened 2020-06-16
Improve three small focus issues.

### LINKED ISSUE #8419: Improve three small focus issues
This is a continuation of #8369 and #8370, brought up by @Yaron10 in #8370.

Two improvements were requested:
- Focus goes not back to edit window if Function List has focus and and the Function list is closed by the menu command **View -> Function List**.
- Focus goes not back to edit window if **Folder as Workspace** has focus and and the **Folder as Workspace** is closed by the menu command **View->Folder as Workspace**.

The third would be:
- Switch focus to edit window after loading a Workspace file from command line.

Uhf7:
Fixes #8419.

### DIFF
--- a/PowerEditor/src/Notepad_plus_Window.cpp
+++ b/PowerEditor/src/Notepad_plus_Window.cpp
@@ -253,6 +253,7 @@ void Notepad_plus_Window::init(HINSTANCE hInst, HWND parent, const TCHAR *cmdLin
 		// of the last session from config.xml.
 		::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1, 0);
 	}
+	::SendMessage(_hSelf, WM_ACTIVATE, WA_ACTIVE, 0);
 
 	// Notify plugins that Notepad++ is ready
 	SCNotification scnN;
--- a/PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
+++ b/PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
@@ -1185,6 +1185,7 @@ void DockingCont::showToolbar(tTbData* pTbData, BOOL state)
 int DockingCont::hideToolbar(tTbData *pTbData, BOOL hideClient)
 {
 	int iItem = searchPosInTab(pTbData);
+	BOOL hadFocus = ::IsChild (pTbData->hClient, ::GetFocus());
 
 	// delete item
 	if (TRUE == ::SendMessage(_hContTab, TCM_DELETEITEM, iItem, 0))
@@ -1218,6 +1219,9 @@ int DockingCont::hideToolbar(tTbData *pTbData, BOOL hideClient)
 			{
 				::SendMessage(_hParent, WM_SIZE, 0, 0);
 			}
+			// set focus to current edit window if the docking window had focus
+			if (hadFocus)
+				::PostMessage(::GetParent(_hParent), WM_ACTIVATE, WA_ACTIVE, 0);
 		}
 
 		// keep sure, that client is hide!!!
--- a/PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
+++ b/PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
@@ -361,8 +361,6 @@ LRESULT DockingManager::runProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM l
 		{
 			tTbData	TbData	= *(reinterpret_cast<DockingCont*>(lParam))->getDataOfActiveTb();
 			LRESULT res = SendNotify(TbData.hClient, DMN_CLOSE);	// Be sure the active item is OK with closing
-			if (res == 0)	// Item will be closing?
-				::PostMessage(_hParent, WM_ACTIVATE, WA_ACTIVE, 0);	// Tell editor to take back focus
 			return res;
 		}
 		case DMM_FLOATALL:


Yaron10:
@Uhf7,

Great. Thank you.

Uhf7:
I debugged it some time again. There is an unnecessary difference between closing a docking window with the mouse (by hitting the close icon) and closing a docking window by closing the last docking content window in it. Hitting the close icon brings the input focus to the edit window. Closing the last content window of a docking window per menu command does not.

To solve this problem, I suggest to move the `Postmessage(WM_ACTIVATE)` line, which brings the focus back to the edit window,  from `DockingManager::DMM_CLOSE` to `DockingCont::hideToolbar`. Since this is hard to explain, here the recorded sequences of events:

	Closing docking window containing single File browser dock with mouse
	=====================================================================

	001 ++doClose by WM_LBUTTONUP (DockingCont)
	002   ++DockingCont::doClose
	003     ++DMM_CLOSE (DockingManager)
	004       ++DMN_CLOSE (FileBrowser)
	005         ++IDM_VIEW_FILEBROWSER (not isClosed() branch)
	006           ++NPPM_DMMHIDE
	007             ++DockingManager::showDockableDlg (false)
	008               ++DockingCont::hideToolbar
	009               --DockingCont::hideToolbar
	010             --DockingManager::showDockableDlg (false)
	011           --NPPM_DMMHIDE
	012         --IDM_VIEW_FILEBROWSER (not isClosed() branch)
	013       --DMN_CLOSE (FileBrowser)
	014     --DMM_CLOSE (DockingManager)
	015     ++DockingCont::hideToolbar
	016     --DockingCont::hideToolbar
	017   --DockingCont::doClose
	018 --doClose by WM_LBUTTONUP (DockingCont)

In this sequence, the `WM_ACTIVATE` message is posted at line `014`, before `DMM_CLOSE` returns. It would be moved to line `009`.

	Closing Folder as Workspace with Menu command
	=============================================

	101 ++IDM_VIEW_FILEBROWSER (not isClosed() branch)
	102   ++NPPM_DMMHIDE
	103     ++DockingManager::showDockableDlg (false)
	104       ++DockingCont::hideToolbar
	105       --DockingCont::hideToolbar
	106     --DockingManager::showDockableDlg (false)
	107   --NPPM_DMMHIDE
	108 --IDM_VIEW_FILEBROWSER (not isClosed() branch)

In this sequence, the `WM_ACTIVATE` message is not posted in the current version, and would be posted at line `105` after this PR.

We have to notice here, that the `DockingCont::hideToolbar` function is already responsible for handling what happens when the last docking content window is closed. Then it closes the docking window itself and resizes the other windows, if necessary. So it's easy to burden it with the focus throwing too.

Also to mention: The focus should only be thrown to the edit window, if the focus before was in the closed docking window. If we have two docking windows, and the 1st has the input focus, it can keep  the focus while the 2nd docking window is closed with a menu command.

Last thing to notice: The `DockingCont::hideToolbar` call in line `015` is actually redundant, but it compensates the lack of proper mechanisms to close Project panels, which could be subject of another PR later.

Yaron10:
@Uhf7,
:+1:

While you're at it. :)
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8426

PowerEditor/src/Notepad_plus_Window.cpp
```diff
@@ -252,6 +252,7 @@ void Notepad_plus_Window::init(HINSTANCE hInst, HWND parent, const TCHAR *cmdLin
 		// the Project Panels are not loaded by Notepad_plus::doOpen only, but also by the Plugin Manager restoring the state
 		// of the last session from config.xml.
 		::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1, 0);
+		::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_SWITCHTO_OTHER_VIEW, 0);
```
donho, line 255: What are you trying to do here?

Uhf7: I switch the focus to the edit view after switching to the correct Project panel, and this only while starting Npp with a workspace file name in the command line.

This is to achieve a similar behavior to other IDE's, easiest example here is MS VC++. The solution file .sln is the equivalent to the workspace file, and if it is opened with MS VC++, the input focus is not in the "Solution explorer", but in the edit window.

Of course, the message I used to achieve this was chosen more to make the code line look similar to the previous one. I looked for what was available in the `Notepad_plus::command` function and found nothing better.

Since I get deeper into it day by day, I can now offer:

		::SendMessage(_hSelf, WM_ACTIVATE, WA_ACTIVE, 0);

And, we could open the big packet and compare the behavior of 
- _Folder as Workspace_
- _Project Panel_
- _Function List_
- _Document Switcher_

What we see is, that the _Function List_ switches back to the edit window in the `Notepad_plus::launchFunctionList()` function, the other three do not. So what happens is, that the input focus tends to be in one of those three side windows instead of in the edit window after startup. Now we could insert a `_pEditView->getFocus()` call at the end of
- `Notepad_plus::launchFileBrowser()`
- `Notepad_plus::launchProjectPanel()`
- `Notepad_plus::launchFileSwitcherPanel()`
(It is already there in `Notepad_plus::launchFunctionList()`)

Then we wouldn't need a separate focus change message while startup. But, it would change the behavior of the three panels when they are started from the menu. Today, after starting them from the menu, they have the input focus. If we make them behave like the _Function List_, then the input focus will be in the edit window after starting them from the menu.

So I asked the MS VC++ IDE here, closed the "Solution Explorer" and opened it again with "View->Solution Explorer", result: The "Solution Explorer" has the input focus.

Which means, the current behavior may not be unexpected by many users. Based on this, I would give the input focus to the side window when it is launched from the menu. But I would also expect, that the edit window has the input focus after startup. The easiest way to solve all of it is to move the `WM_ACTIVATE` message one bracket down. Like this:

	if (cmdLineParams->_openFoldersAsWorkspace)
	{
		_notepad_plus_plus_core.launchFileBrowser(fns, true);
	}
	else if (_notepad_plus_plus_core._isWorkspaceFileLoadedFromCommandLine)
	{
		::SendMessage(_hSelf, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1, 0);
	}
	::SendMessage(_hSelf, WM_ACTIVATE, WA_ACTIVE, 0);

It fixes the startup input focus for all three panels without changing there current behavior while starting them from the menu. Additionally, you could think about removing the `_pEditView->getFocus()` call from `Notepad_plus::launchFunctionList()`, to make the _Function List_ behave like the other panels.

Alternatively, we could put the `_pEditView->getFocus()` in the `launch...` functions of the three other panels and accept the change of behavior while starting them from the menu. Would be fine with me too.

Many choices. What would you like?

Yaron10:
When the user is opening a window, the expected behavior IMO is that this window should get the focus.

The user might want to look through the various functions, and although this can be done even if the editor has the focus - it should be better to let the Function List keep the focus.

And therefore I'd choose removing the `_pEditView->getFocus()` call from `Notepad_plus::launchFunctionList()`.

PowerEditor/src/Notepad_plus_Window.cpp
donho: `IDM_VIEW_SWITCHTO_OTHER_VIEW` makes the focus change from a Scintilla Edit View to another one, if you have 2 edit views opened side by side. So I don't think it's what you want.

Uhf7: You are right, that's not what I want to do. But `IDM_VIEW_SWITCHTO_OTHER_VIEW` works, because it switches the input focus to the current edit view, as long as the input focus is not at the current edit view. And in this situation, the input focus is at the Project panel.

But as I wrote, I would use `WM_ACTIVATE` now instead, and I would make it work for _Folder As Workspace_ and _Doc Switcher_ too, by shifting it one line down.

donho: > But IDM_VIEW_SWITCHTO_OTHER_VIEW works,

Just as having sex with my sister - it works doesn't mean it's right :)

> But as I wrote, I would use WM_ACTIVATE now instead,

Yes, please do it.

Uhf7: Do you approve the described move of the `WM_ACTIVATE` message one line down too? This would make the following change to the original PR: During startup, the input focus would be set to the edit window not only in the case when a Workspace file is specified in the command line, but also if any of the following side windows is loaded during startup:
- _Folder as Workspace_
- _Project Panel_
- _Doc Switcher_

The current state is, that if any of this side windows is loaded automatically during startup, the input focus is in the side window instead of the edit window. This would change for good if we move the `WM_ACTIVATE` line as I described in my long comment.

donho: @Uhf7 
> This would make the following change to the original PR: During startup, the input focus would be set to the edit window not only in the case when a Workspace file is specified in the command line, but also if any of the following side windows is loaded during startup:

Yes, the focus should be always on the edit view after startup.

PR #8425, opened 2020-06-17
Include functionList.xml in minimalist packages.

### LINKED ISSUE #8371: functionList.xml is missing in minimalist packages
I'm using the minimalist packages of Npp and noticed today, that the Function List is always empty. After some digging I found, that the `functionList.xml` is not part of `npp.7.8.6.bin.minimalist.7z` and `npp.7.8.6.bin.minimalist.x64.7z`. Could it be included in the next installation?

Uhf7:
Fixes #8371.


donho:
As its name suggests, `minimalist` is the minimal package which contains nothing but needed components to run Notepad++.
If user needs functionList.xml, he/she can download zip/7z package, or add this file manually. So let's keep minimalist packages as it is.

Issue #8428, opened 2020-06-17
Limit scope of Remove Empty Lines to a selection, if one is active


sasumner:
### Description of the Issue

Rarely do I want to remove empty lines in my entire document.  Often I'd like it to be done in limited scope, say, the currently selected lines.

As of Notepad++ 7.8.7, this is not possible (without a regular expression replacement operation or, of course, manually deleting each one, but these are hardly user-friendly methods).

### Steps to Reproduce the Issue

1.  Create some text with some empty lines:
```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.

Up Jack got
And home did trot,
As fast as he could caper;
And went to bed
And plastered his head
With vinegar and brown paper.

When Jill came in
How she did grin
To see Jack's paper plaster;
Mother vexed
Did whip her next
For causing Jack's disaster.
```

2.  Select some of the lines, being sure that at least one empty line is part of the selection:
![image](https://user-images.githubusercontent.com/30118311/84936666-9d672b80-b0a8-11ea-8e50-0e22329e7aa4.png)

3.  Attempt to remove empty lines, limited to the scope of selected text, using _Edit_ menu > _Line Operations_ > _Remove Empty Lines_

### Expected Behavior
I expected the empty lines in my active selection to be removed, and any other empty lines to be left untouched:
![image](https://user-images.githubusercontent.com/30118311/84937016-22eadb80-b0a9-11ea-935f-a6190c9b3197.png)

### Actual Behavior
All empty lines in the document were removed:
![image](https://user-images.githubusercontent.com/30118311/84936803-cd163380-b0a8-11ea-8a86-145e412b6c6e.png)

### Debug Information
Notepad++ v7.8.7   (32-bit)
Build time : Jun  5 2020 - 02:34:28
Path : C:\............\npp.7.8.7.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

### Other:

This should also apply to _Remove Empty Lines (Containing Blank characters)_

sasumner:
Released in N++ version 7.8.8

PR #8429, opened 2020-06-17
Allow scope of Remove Empty Lines to be limited by an active selection

sasumner:
Implements #8428

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1400,10 +1400,13 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		env._str2Search = TEXT("^$(\\r\\n|\\r|\\n)");
 	}
 	env._str4Replace = TEXT("");
-    env._searchType = FindRegex;
-
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, true);
-
+	env._searchType = FindRegex;
+	auto mainSelStart = _pEditView->execute(SCI_GETSELECTIONSTART);
+	auto mainSelEnd = _pEditView->execute(SCI_GETSELECTIONEND);
+	auto mainSelLength = mainSelEnd - mainSelStart;
+	bool isEntireDoc = mainSelLength == 0;
+	env._isInSelection = !isEntireDoc;
+	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
 	if (isBlankContained)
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1400,10 +1400,13 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		env._str2Search = TEXT("^$(\\r\\n|\\r|\\n)");
 	}
 	env._str4Replace = TEXT("");
-    env._searchType = FindRegex;
-
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, true);
-
+	env._searchType = FindRegex;
+	auto mainSelStart = _pEditView->execute(SCI_GETSELECTIONSTART);
```
sasumner, line 1404: Is it nicer to do this here instead?
`Sci_CharacterRange mainSelection = _pEditView->getSelection();`
And then work with that in the next lines?
Is it dangerous to comment on one's own code?  Is that asking for trouble? :-)

donho: > Is it nicer to do this here instead?:
Sci_CharacterRange mainSelection = _pEditView->getSelection();

Yes, why not?

> Is it dangerous to comment on one's own code? Is that asking for trouble? :-)

Is it a philosophical question? Am I supposed to answer this question?  :)

donho:
PR tested and it works as usual.
However, there's a behaviour you might not consider:
If a word or some words of the same line is/are selected, do we treat the selected zone in one line or we ignore this case (so treat the whole document) for line*s* operation?

donho:
@sasumner Any thought about the behaviour mentioned above?

donho:
@sasumner Any thought about the behaviour mentioned above?

sasumner:
@donho 
I'm typing my response.
Give me a time to complete it please. :-)

sasumner:
@donho

>If a word or some words of the same line is/are selected, 

Overall, it might be strange for users until they learn about selection-based removal of empty lines...

>do we treat the selected zone in one line 

Nothing would change for the user in this case; it would look like their command didn't do anything.
That's OK because they would think "What did I do wrong?" and then "Oh, this must now be a selection-based command".

>or we ignore this case (so treat the whole document) for lines operation?

I would think NO, we do not treat the whole document here.
User must give up the selection completely in order to treat the whole document.
Or, of course, do ctrl+a first.

PR is OK as it is for this functionality.

But if you want changes, just tell me what they are.

donho:
User could double click a word accidentally (that makes word selection), then scroll down, and run this command. In this case, no empty line is removed, and user is confused. That's the scenario that I can think about. 

> But if you want changes, just tell me what they are.

Just want to get your thought about it.

sasumner:
>User could ... makes ... selection, then scroll down, and run this command. In this case, no empty line is removed, and user is confused.

I agree that is a possible scenario.
If adopted, the new behavior could be strange to users for a bit of time, until they are exposed to it and become familiar.

donho:
OK, let's do it in this way currently. We will see if there are some users complain it in the future.

PR #8437, opened 2020-06-19
Fix Find-in-Finder produces false hit

### LINKED ISSUE #8435: Find-in-Finder produces hit when it shouldn't when searching in previously found lines
1. Create some text in a new Notepad++ tab:
```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```
2. Do a Ctrl+S and give that file a name in the file system (IMPORTANT! Do not continue these steps with an "unnamed" file, e.g. "new 1"):
image

3. Initiate a _Find_ by pressing Ctrl+f and fill out the search this way:
image
4. Conduct the search by pressing the _Find All in Current Document_ button
5. Observe the following result:
image
6. Right-click in the _Find result_ window to obtain:
image
7. Select _Find in these found results..._
8. Fill out the resulting window this way:
image
9. Press the _Find All_ button in the _Find in found results_ window
10. Observe:
image

### Expected Behavior
I expect Notepad++ to indicate that no hits occurred because I told it to _Search only in found lines_.  Because _fetch_ doesn't occur in any of the previously found lines, it should result in a decision of "0 hits".

### Actual Behavior
Notepad++ displays "1 hit" but does not show that single hit, so I don't really know what it is thinking. :-)

### Debug Information
Notepad++ v7.8.7   (32-bit)
Path : C:\\...........\npp.7.8.7.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252

Uhf7:
Fixes #8435.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2087,7 +2087,7 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 
 		int foundTextLen = targetEnd - targetStart;
 		int replaceDelta = 0;
-
+		bool processed = true;
 				
 		switch (op)
 		{
@@ -2136,12 +2136,6 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 
 				const TCHAR *pFileName = pFindersInfo->_pFileName ? pFindersInfo->_pFileName : TEXT("");
 
-				if (!findAllFileNameAdded)	//add new filetitle in hits if we haven't already
-				{
-					pFindersInfo->_pDestFinder->addFileNameTitle(pFileName);
-					findAllFileNameAdded = true;
-				}
-
 				auto lineNumber = pEditView->execute(SCI_LINEFROMPOSITION, targetStart);
 				int lend = static_cast<int32_t>(pEditView->execute(SCI_GETLINEENDPOSITION, lineNumber));
 				int lstart = static_cast<int32_t>(pEditView->execute(SCI_POSITIONFROMLINE, lineNumber));
@@ -2163,14 +2157,14 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 				SearchResultMarking srm;
 				srm._start = start_mark;
 				srm._end = end_mark;
-				
-				if (pOptions->_isMatchLineNumber)
+				processed = (!pOptions->_isMatchLineNumber) || (pFindersInfo->_pSourceFinder->canFind(pFileName, lineNumber + 1));
+				if (processed)
 				{
-					if (pFindersInfo->_pSourceFinder->canFind(pFileName, lineNumber + 1))
-						pFindersInfo->_pDestFinder->add(FoundInfo(targetStart, targetEnd, lineNumber + 1, pFileName), srm, line.c_str());
-				}
-				else
-				{
+					if (!findAllFileNameAdded)	//add new filetitle in hits if we haven't already
+					{
+						pFindersInfo->_pDestFinder->addFileNameTitle(pFileName);
+						findAllFileNameAdded = true;
+					}
 					pFindersInfo->_pDestFinder->add(FoundInfo(targetStart, targetEnd, lineNumber + 1, pFileName), srm, line.c_str());
 				}
 				break;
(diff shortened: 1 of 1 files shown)


sasumner:
@donho Have you noticed this one?

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -2263,7 +2257,7 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 			}
 			
 		}	
-		++nbProcessed;
+		if (processed) ++nbProcessed;
```
donho, line 2260: Next time, please separate the instruction from the condition (put the instruction in a new line).

sasumner: @donho, @Uhf7 

Is the rule _always_ to put the instruction in a new line?
I think sometimes (like this one) it is very concise and very readable to do it the way @Uhf7 did it.
But if the rule is to always do it, then I suggest it should also use curly braces as well, like this:
```
if (processed)
{
    ++nbProcessed;
}
```
What is the thinking about that?

donho: > Is the rule always to put the instruction in a new line?

It depends.
In this case, it's just easier for reading (only for me maybe?).
On the following case, I would say keep it in the same line:
```
	switch (_winVersion)
	{
		case WV_WIN32S: return TEXT("Windows 3.1");
		case WV_95: return TEXT("Windows 95");
		case WV_98: return TEXT("Windows 98");
		case WV_ME: return TEXT("Windows Millennium Edition");
		case WV_NT: return TEXT("Windows NT");
		case WV_W2K: return TEXT("Windows 2000");
		case WV_XP: return TEXT("Windows XP");
		case WV_S2003: return TEXT("Windows Server 2003");
		case WV_XPX64: return TEXT("Windows XP 64 bits");
		case WV_VISTA: return TEXT("Windows Vista");
		case WV_WIN7: return TEXT("Windows 7");
		case WV_WIN8: return TEXT("Windows 8");
		case WV_WIN81: return TEXT("Windows 8.1");
		case WV_WIN10: return TEXT("Windows 10");
		default: /*case WV_UNKNOWN:*/ return TEXT("Windows unknown version");
	}
```
//But if the rule is to always do it, then I suggest it should also use curly braces as well, like this:
```
  if (processed)
  {
    ++nbProcessed;
  }
```

I'm not against such style, but I would use the following one:
```
  if (processed)
    ++nbProcessed;
```
so it looks more compact.

Uhf7: I'm under the impression that increasing line count is negative for PR acceptance. All recommendations, I read so far, say so. It's an optimization criteria for PRs. So my single line can be seen as cheating here, I admit this. I'll avoid this in the future. Thank you for the hint.

sasumner: On a separate line also makes it more obvious what is going on when single stepping with the debugger.

Uhf7: It's much easier to debug, if it is in two separate lines, I sign this immediately.

donho: > I'm under the impression that increasing line count is negative for PR acceptance. All recommendations, I read so far, say so.

True. I avoid the tourists and bad coders.
But the readability of codes you added is important as well, so "bad" coding style also brings some negative points :)

sasumner: "tourists" LOL

Uhf7: And, @sasumner, to answer your actual question, I've been involved in projects already, where opening a new code block (opening curly brace after if's) was compulsory. So I kind of understand this wish.

But it's readable enough for me without the curly brackets, as long as the indentation is correct, and as long it is only a simple short statement. A multi-parameter call spreading over several lines deserves curly brackets.

 And, since the Notepad++ code contains already many of those single statements after if's, I would simply continue this tradition.

donho: > I'm under the impression that increasing line count is negative for PR acceptance. All recommendations, I read so far, say so.

Your statement is totally true in *.h BTW

sasumner: @Uhf7 The real reason to like the curly braces even with a single statement, is that when that statement becomes 2 or more, it is easy to fall into the trap of believing the indentation. :-)

Uhf7: Believing in indentation? That could happen if a Python programmer is at work ...

Issue #8432, opened 2020-06-18
Remove the "Replace All in All Opened Documents" 'feature'


Gewath:
### Description of the Issue
Problem: accidentally clicking "Replace All in All Opened Documents" ruins potentially all your text files, causing you to have to go through potentially hundreds of thousands of open text files and fix them one by one, and ruins them all permanently if you've closed Notepad++ in-between (#2855 #5253 #5260).

Previous topics on this feature were closed because "you can just fix them by going through them one by one, and clicking undo". This is akin to saying "you can just replace all in all opened documents, by going through them one by one and clicking replace all".

As the function is now admitted to be completely pointless, remove it entirely, it clutters the "replace" tool in Notepad++.

### Steps to Reproduce the Issue
1. Open every file you've ever edited in Notepad++.
2. Delete original save files, keep files in editor. (This is how most people use Notepad++)
3. Ctrl+F.
4. Replace.
5. Check tickbox 'regular expression'.
6. Find what: [A-Za-z0-9]
7. Replace with: I shall not add one-click global file corrupting buttons to my program.
8. Hover mouse over 'replace all'.
9. Move mouse 1cm down.
10. Left click.
11. Close Notepad++.

![notepad++ replace all in all](https://user-images.githubusercontent.com/27696856/85042820-f360cf80-b18b-11ea-9772-60754ed8605d.png)

### Debug Information
Notepad++ v7.8.5   (32-bit)
Build time : Mar  4 2020 - 11:04:20
Path : E:\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Home (64-bit) 

OS Version : 1703
OS Build : 15063.674
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

Screenshot has been modified.

sasumner:
Here's a nice little Pythonscript to remove the feature.  
Move the key-binding of Ctrl+h from the default over to running this script.  
Note that the _Replace All in All Opened Documents_ button is now gone!:

![image](https://user-images.githubusercontent.com/30118311/85062664-de2d7600-b176-11ea-859b-b9675072353f.png)

The script:
```
# -*- coding: utf-8 -*-

from Npp import editor, notepad, MENUCOMMAND
import ctypes
from ctypes.wintypes import BOOL, HWND, LPARAM

try:

    REPLACE_DIALOG().show()

except NameError:

[...]

Issue #2855, opened 2017-01-31
Some suggestions that would be very convenient.


jginfo:
Hello

Some suggestions that would be very convenient.

1) Horizontal scrolling of top Files tabs with thumbwheel (when the mouse pointer hovers over the Top Files tabs)
2) pop-up confirmation 'yes/no' for the option 'Close All BUT This' (when you right click on Top Files tab)

3) pop-up confirmation 'yes/no' for button 'Replace All in All Opened Documents' (in the pop-up 'Replace')

4)  in the pop-up 'Find' tab 'Mark'
make 2 buttons instead of the 'Clear' button
- button 'Clear Text' purging the marks in the text alone (NOT in the column of line numbers)
- button 'Clear Line' purging the marks in the column of line numbers

personally, I prefer the marking options in the 'Find' tab ... (As it was a time!)

I am French and below the French version
----------------------------------------------------
Quelques suggestions qui seraient bien pratique.

1) Pouvoir scroller horizontal les onglets des fichiers avec la molette (lorsque le pointeur de la souris survol les onglets des fichiers)
2) pop-up de confirmation 'oui/non' pour l'option 'fermer tout sauf ce fichier' (lorsque l'on click droit sur un onglet de fichier)

3) pop-up de confirmation 'oui/non' pour le bouton 'Remplacer dans tous les documents ouverts'
4) Dans la popup 'Cherche' onglet 'Marquer'
faire 2 boutons au lieu du seul bouton 'Purger'
- bouton 'Purge Texte' purgeant les marques dans le texte seul (PAS dans la colonne des n° de lignes)
- bouton 'Purge Lignes' purgeant les marques dans la colonne des n° de lignes

perso, je préférais les options de marquage dans l'onglet rechercher... (comme il fut un temps)

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
[...]

PR #8438, opened 2020-06-19
Add confirmation prompt to Replace-all-in-all-opened-docs command

sasumner:
Implements #5253
Probably implements the implied request in #8432 ??
Implements point 3 in #2855

Also implements desires expressed here:
* https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8213#issuecomment-624167936
* https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8213#issuecomment-632230473

So this was easy enough to implement.
Although, personally, I did not mind the existing behavior as I've never (in 5 years of using N++) pressed _Replace All in All Opened Documents_ by mistake.

My feeling is that I don't care one bit if this gets accepted or rejected.
If it gets rejected, that will stand as the final decision on this feature, and any future issues opened of this nature will be quickly closed without debate.


keanerickson:
Thanks for addressing this despite the protestations of some who would just blame "user error".

PR #8452, opened 2020-06-23
BUILD.md file updated

ayushsinha-13:
2nd step of Build Notepad++.exe changed


donho:
Let's keep this section as generic as possible.

ayushsinha-13:
ok

PR #8475, opened 2020-06-25
Fix horizontal scroll bar position

### LINKED ISSUE #8466: Horizontal scrollbar position is reset on minimise - restore of window
### Description of the Issue

If you have an open tab with a horizontal scrollbar that's been scrolled to the right.
Minimising then restoring the Notepad++ window resets the scrollbar position back to the left even though the content is still scrolled.

Clip below shows content is scrolled but scrollbar has been reset after minimise/restore.

image

1. Open any file that would cause the horizontal scrollbar to show
2. Scroll to the right
3. Minimise the Notepad++ window
4. Restore the Notepad++ window

### Expected Behavior

Horizontal scrollbar to keep scroll position

### Actual Behavior

Horizontal scrollbar position is reset

### Debug Information

Notepad++ v7.8.6   (64-bit)
Build time : Apr 21 2020 - 15:23:03
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.900
Plugins : ComparePlugin.dll DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

Uhf7:
Fixes #8466.

This effect does not only occur after restoring a minimized Npp, but also after starting Npp with a saved session. The lines included force Scintilla to update the horizontal scroll bar(s) without changing the current x-offset.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1646,6 +1646,10 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case WM_ACTIVATE:
 		{
 			_pEditView->getFocus();
+			auto x = _pEditView->execute(SCI_GETXOFFSET);
+			_pEditView->execute(SCI_SETXOFFSET, x);
+			x = _pNonEditView->execute(SCI_GETXOFFSET);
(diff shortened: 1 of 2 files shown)


Uhf7:
Had to add one more thing to it: After-live calls to `_pScintillaFunc` should be avoided. They are caught, invisible to the user, but they are ugly while debugging.

Issue #8400, opened 2020-06-10
Rectangular selection does not move when inserting new chars.


Ekopalypse:
### Description of the Issue

When creating a rectangular selection and adding new chars the selection does
not move to its new positions

### Steps to Reproduce the Issue
1. Create a few lines
2. Put the cursor to column 0 and create a rectangular selection
3. press tab key

### Expected Behavior
The rectangular selection has been placed after the inserted tab characters

### Actual Behavior
The rectangular selection is before the inserted tab characters

### Debug Information
Notepad++ v7.8.7   (64-bit)
Build time : Jun  5 2020 - 02:38:56
Path : D:\...\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.24544
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

Example 
npp 7.8.7:
![image](https://user-images.githubusercontent.com/47723516/84299892-c9872780-ab51-11ea-850c-0adc1891876a.png)

npp 7.8.6:
![image](https://user-images.githubusercontent.com/47723516/84299832-ad838600-ab51-11ea-98d4-9a4365c395fc.png)

Uhf7:
What happens here: Scintilla doesn't like `SCI_INDICSETSTYLE` or `SCI_INDICSETHOVERSTYLE` commands while editing. Whenever such a command occurs, it calls a kind of internal reset/redraw function, which wrecks the cursor position of block selection.

There is nothing in their documentation saying don't change indicator styles while rectangular selection is active, so nobody saw it coming. 

It occurs in other situations too, for example, when changing zoom while the rectangular selection is active, but that is far less disturbing.

Fortunately, there is an easy way to fix it, I will create a PR showing this way.

sasumner:
@Uhf7 Is it worth putting a bug report in on the Scintilla site?  If so I will ask you to please do it.

Uhf7:
Actually, I don't know if they would see it as a bug. They could see it as "normal" behavior, which is not described in the `ScintillaDoc.html`.

And it's hard to file an example of the effect, which they can reproduce, but I think about it.

donho:
@Uhf7  There was an Windows update waiting - after updating and restarting my laptop, the PR is working. Thank you for the fix.

PR #8496, opened 2020-06-29
Fix URL hovered style with no underline

### LINKED ISSUE #8493: URL Hovered colour disappears after modifying any Clickable Link Settings
### Description of the Issue
Trying to change `Clickable Link Settings` in Preferences dialog makes loose the URL hovered colour.

1. Start Notepad++, make sure  URL hovered colour is present.
2. Open Preferences dialog, click "No Underline" in `Clickable Link Settings`.
3. Move mouse on a URL.

### Expected Behavior

the URL hovered colour should be always present.

### Actual Behavior

the URL hovered colour is absent.

### Debug Information
Notepad++ v7.8.8   (32-bit)
Build time : Jun 24 2020 - 00:15:36
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1903
OS Build : 18362.900
Current ANSI codepage : 1252

Uhf7:
Fixes #8493.

The reason is, that the `SCI_INDICSETSTYLE` command unfortunately resets the indicator hover style too, so that it has to be set again after calling `SCI_INDICSETSTYLE`. 

The desired functionality, for what this trouble began, was to make the change of the _No Underline_ option visible immediately. The first solution was, to set it every time in `addHotSpot`, until it was realized, that the cursor position was reset while block selection. The solution for this was to call the `SCI_INDICSETSTYLE` command only, if the _No Underline_ option changed. _Edit:_ This was the occasion, when the new bug crawled in (660284011776830a4f483ceeadf244dd5522d7f9). I moved `SCI_INDICSETHOVERSTYLE` to the init function of Scintilla Edit view, because I thought, this never changes, it needs to be set only once.

Now, the `SCI_INDICSETHOVERSTYLE` is called after every `SCI_INDICSETSTYLE` too, the PRO for this is, that all indicator style setting commands are at the same place again.  As before, the indicator style setting commands are only applied if the style has been changed, so that the problem with the block selection remains absent.

Successfully tested with this PR:
- Cursor position while inserting chars in block selection (Issue #8400)
- Underline color in dark themes (Issue #8405)

Sorry for the trouble.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2498,8 +2498,15 @@ void Notepad_plus::addHotSpot(ScintillaEditView* view)
 	LPARAM indicStyle = (urlAction == 2) ? INDIC_PLAIN : INDIC_HIDDEN;
 
 	LPARAM indicStyleCur = pView->execute(SCI_INDICGETSTYLE, URL_INDIC);
-	if (indicStyleCur != indicStyle)
+	LPARAM indicHoverStyleCur = pView->execute(SCI_INDICGETHOVERSTYLE, URL_INDIC);
+
+	if ((indicStyleCur != indicStyle) || (indicHoverStyleCur != INDIC_FULLBOX))
+	{
 		pView->execute(SCI_INDICSETSTYLE, URL_INDIC, indicStyle);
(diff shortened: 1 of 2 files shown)


Uhf7:
Just investigated about `SCI_INDICSETHOVERFORE`. This would be reset with `SCI_INDICSETFORE`, which is not used anymore, since the underline color is determined by the indicator value because of the dark themes. The URL hover color remains correct.

PR #8497, opened 2020-06-29
Retain local config option during update installation

kenijo:
Check for the existence of the doLocalConfig.xml file in the installer so that during installation of Notepad++ we don't have to manually recheck the "Don't use %APPDATA%" checkbox.
This is even more valuable when doing silent updates so it retains the existing configuration.


donho:
It would be a good feature if it worked.
Just tried the PR: the checkbox of "Don't use %APPDATA%" is checked when `doLocalConf.xml` is present, but the file `doLocalConf.xml` is still removed after the installation.

Issue #8480, opened 2020-06-26
Add VHDL suppport to functionList


el-coder-sb:
### Add VHDL suppport to functionList.xml
Would be fine to have a functionList entry for VHDL.

Since VHDL is not a software language it does not follow the constructs (e.g. classes and class methods) known from software. This leads to several different possible (and reasonable) "translations" of the VHDL structures to the given class-function pattern of functionList (e.g. https://community.notepad-plus-plus.org/topic/11554/function-list-for-vhdl/2 and further topics).

But as there is currently no VHDL support at all at functionList, I suggest to implement at least one reasonable entry for VHDL so (new) users will have at least a good overview about the content/modules inside a VHDL file and a good starting point if further functionList requirements exist.

### Expected Behavior
I propose the following behavior:
Detect all entities, architectures, components (declarations not instantiations), blocks and processes in a file and interpret them as simple "function" without class.

Example:
[simple_functionList_test.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/4835974/simple_functionList_test.txt)

could give (the last entry is new in VHDL-2008):

     --- entity ent1 is
     --- architecture rtl of ent1 is
     --- component compo1 is
     --- component reset_controller is
     --- proc1: process (reset_reset_n, clk_clk)
     --- led_block: block
     --- doing_so_inst_4c : entity work.doing_so

I had further ideas like interpreting the port maps as class-functions. [NppForum-FunctionList](https://community.notepad-plus-plus.org/topic/19517/add-vhdl-to-functionlist-xml-and-interpret-port-map-as-class/10)
But as we (thanks a lot @MAPJe71 ) could not find a reliable solution for all the possibilities VHDL allows, I would propose to use only the above simple "function" detection.

I will provide a pull request if this issue is accepted.

PR #8509, opened 2020-07-02
Add VHDL to functionList.xml

el-coder-sb:
Implements #8480.

Add VHDL to functionList.xml supporting ENTITY, BLOCK, PROCESS, COMPONENT, ARCHITECTURE.


PowerEditor/src/functionList.xml
```diff
@@ -936,6 +937,34 @@
...
+				<function
+					mainExpr="(?x)                                          # free-spacing (see `RegEx - Pattern Modifiers`)
+						  ^                                             # match at beginning of line
+						  \h*                                           # optional leading whitespace
+						  (\w+\h*:)?                                    # ID of process
+						  (\h(?!END)\w+\h)*                             # make sure this is not the end of a module
+						  \h*
+						  (?(1)(ENTITY|BLOCK|PROCESS)|                  # keywords if ':' was detected before
+						  (ENTITY|COMPONENT|ARCHITECTURE))              # keywords else
+						  \h*
```
MAPJe71, line 959: `\h+` to have at least 1 blank between a keyword and the identifier.

PowerEditor/src/functionList.xml
```diff
@@ -936,6 +937,34 @@
...
+						  [A-Za-z0-9_\(\)., ]*                          # ID or sensitivity list
```
MAPJe71, line 960: `[...]+` to have at least one character in the identifier or sensitivity list.

PowerEditor/src/functionList.xml
```diff
@@ -936,6 +937,34 @@
...
+				id         ="vhdl_syntax"
+			>
+				<function
+					mainExpr="(?x)                                          # free-spacing (see `RegEx - Pattern Modifiers`)
+						  ^                                             # match at beginning of line
+						  \h*                                           # optional leading whitespace
+						  (\w+\h*:)?                                    # ID of process
+						  (\h(?!END)\w+\h)*                             # make sure this is not the end of a module
+						  \h*
+						  (?(1)(ENTITY|BLOCK|PROCESS)|                  # keywords if ':' was detected before
```
MAPJe71, line 957: Likewise, `(?:ENTITY|BLOCK|PROCESS)` to make it non-capturing group.

PowerEditor/src/functionList.xml
```diff
@@ -936,6 +937,34 @@
...
+						  (ENTITY|COMPONENT|ARCHITECTURE))              # keywords else
```
MAPJe71, line 958: Likewise, `(?:ENTITY|COMPONENT|ARCHITECTURE)` to make it non-capturing group.

PowerEditor/src/functionList.xml
```diff
@@ -936,6 +937,34 @@
...
+			<parser
+				displayName="VHDL"
+				id         ="vhdl_syntax"
+			>
+				<function
+					mainExpr="(?x)                                          # free-spacing (see `RegEx - Pattern Modifiers`)
+						  ^                                             # match at beginning of line
+						  \h*                                           # optional leading whitespace
+						  (\w+\h*:)?                                    # ID of process
+						  (\h(?!END)\w+\h)*                             # make sure this is not the end of a module
```
MAPJe71, line 955: `(?:\h(?!END)\w+\h)*` to make it non-capturing group.

el-coder-sb:
Thanks for your help.

I will do the changes!

And I will provide a unittest.
But I can´t run the unitTestLauncher.ps1. How do I do without building notepad++ from source?

What I did:
I downloaded/cloned the current npp repo. I run unitTestLauncher.ps1 but this fails as there is only 'Notepad++.exe.lastcodeanalysissucceeded' instead of a real exe. OK.
Then I changed the path in the ps-script to my current installed npp version but then it tells me: 
Unexpected token "-export"

PowerEditor/src/functionList.xml
el-coder-sb: Done. Changed to` [\h\(]+` to allow parentheses after keyword process without blank.

PowerEditor/src/functionList.xml
el-coder-sb: Done.

PowerEditor/src/functionList.xml
el-coder-sb: Done, but I can´t see the benefit of non-capturing mode?!?

PowerEditor/src/functionList.xml
el-coder-sb: Done, but I can´t see the benefit of non-capturing mode?!?

PowerEditor/src/functionList.xml
el-coder-sb: Done, but I can´t see the benefit of non-capturing mode?!?

el-coder-sb:
@MAPJe71 any further comments/ideas?

MAPJe71:
Comment: 
1. Get rid of the leading and trailing whitespace in the Function List entries.

Ideas: 
1. Make 'entity' and 'architecture' branches. Note though that the 'entity' of your example will not show because it is empty.

```
\-- unitTest
    +-- entity ent1
    |   \-- ...
    \-- architecture rtl of ent1
        +-- component compo1
        +-- component reset_controlleR
        +-- proc1: process (reset_reset_n, clk_clk)
        +-- block1: block
        \-- comp_per_entity : entity work.doing_so port map ( Reset_s 
```

donho:
@el-coder-sb 

What's the input file of uni-test?

el-coder-sb:
> What's the input file of uni-test?

@donho : VHDL/unitTest is a syntactical correct vhdl file with some possible notation formats. It is reduced to the relevant part: e.g. the process does not have any body here as it isn´t necessary for the functionList unittest.

Hope this answers your question?!

el-coder-sb:
>     1. Make 'entity' and 'architecture' branches. Note though that the 'entity' of your example will not show because it is empty.

@MAPJe71 : I decided to not to do so as there was the problem with the port maps of the components. As I did not implement any port map detection I did not implement any port detection at all (i.e. not for entities).

Still one can argue that making the architecture a branch (with it´s components etc.) would be reasonable. It is. I didn´t make it a branch as it would lead to one extra click (to expand the branch) every time I open a new file and want to see all the used components and processes (which is more likely than wanting to know the architecture as the usual practice is to insert exactly one architecture in one file). 
As [npp-forum: FunctionList for VHDL ](https://community.notepad-plus-plus.org/topic/11554/function-list-for-vhdl/7) proposes it the same way I decided to do so as well.

If you very prefer the cleaner but less practical solution with the branch I could do so. However I prefer the practical solution.

MAPJe71:
> If you very prefer...

It was just a suggestion. I'm not a VHDL user myself, was wonderering if it would be more convenient.

donho:
@el-coder-sb Sorry, my bad. It looks fine. It'll be merged in few minutes.

donho:
> But I can´t run the unitTestLauncher.ps1. How do I do without building notepad++ from source?

You should launch `Powershell`, then go to the directory `c:\...\notepad-plus-plus\PowerEditor\Test\FunctionList` (where there is the powershell script `unitTestLauncher.ps1`) in the launched powershell, then type `.\unitTestLauncher.ps1` then run it - the script will run  `c:\...\notepad-plus-plus\PowerEditor\bin\notepad++.exe` to generate the json result to test, then compare it with the expected result. Don't forget to copy your modified `functionList.xml` into `c:\...\notepad-plus-plus\PowerEditor\bin\`, because `c:\...\notepad-plus-plus\PowerEditor\bin\notepad++.exe` loads only `functionList.xml` from the same directory.

el-coder-sb:
> It was just a suggestion. I'm not a VHDL user myself, was wonderering if it would be more convenient.
@MAPJe71 Thanks a lot for your ideas! I appreciate your opinion! In this case I have (at the moment) a different opinion. Thanks for accepting this. 
I will be open to any further ideas in future. And may be with some more time working with the function list I will get a changing opinion by myself... We will see. 

Thanks a lot for your effort!! (@donho as well!!)

PR #8512, opened 2020-07-03
Update Italian localization to 7.8.8

Ices-Eyes:
Update to 7.8.8 and typos and corrections


donho:
Quick check with https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8385, the changes are the same.
Since https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8385 comes firstly, I'll take that one.

PR #8516, opened 2020-07-03
Fix Found line may not be centered if word wrap is on

### LINKED ISSUE #8489: Found line may not be centered if Word-Wrap is ON
**STR:**
Word-Wrap: ON.
Open `NppCommands.cpp` (_caret at the beginning_).
Open the Find dialog and search `IDM_FILE_RELOAD`.
Switch to another tab.
In the `Find Results` window double-click the *second result*.

**Result:**
The found line is not centered in the editor (it might depend on NPP window size).
תמונה

If you double-click the second result and the current tab is `NppCommands.cpp`, the found line is centered properly.
תמונה

(Reproducible also in `Find in Files` if a file is opened after double-clicking a found line). 
***
Notepad++ v7.8.7   (32-bit)
Path : C:\---\npp.7.8.7.bin.minimalist\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.23964
Current ANSI codepage : 1255
Plugins : none

Uhf7:
Fixes #8489.


Yaron10:
:+1:

Issue #8461, opened 2020-06-24
[Find Result] go to wrong tab when file is temporary


Tequiller:
### Description of the Issue
When we have several files open which many temporary (never saved), and we wish to reach at a research result line from the "Find result", if this line is from a temporary file the NPP go to the equivalent line to the tab already opened.

### Steps to Reproduce the Issue
1. Open saved files
2. Copy to the clipboard one of those text
3. Open new file
4. Write several lines in the temporary file
5. Then, paste text from the clipboard to this temporary file
6. Search a word that appear in both files thanks to [Find All in All Opened Documents]
7. From the [Find result], double click on the result from the saved file (should go into the good tab)
8. From the [Find result], double click on the result from the temporary file (if before the double click you were to the saved file, NPP keep this tab opened, and show you the equivalent characters position of the temporary file.

### Expected Behavior
NPP should recognised the good one tab and go to it even if it is temporary.

### Actual Behavior
NPP keep the actual tab open in the case of line result is from a temporary file.

### Debug Information
Notepad++ v7.8.6   (32-bit)
Build time : Apr 21 2020 - 15:17:06
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS Name : Windows 8 (64-bit) 
OS Build : 9200.0
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
Can't reproduce with N++ 7.8.7 -- closing.
@Tequiller If you can provide a more specific example where it _does_ reproduce, of course we will reopen.

Tequiller:
Thanks for your reponse...
Indeed there is a 7.8.7 version... strangely my NPP don't tell me there's a new version... (bug ???)
I go to update it manually

![image](https://user-images.githubusercontent.com/46250710/85552064-e8da8600-b622-11ea-8812-3d38d68b8c72.png)

sasumner:
>strangely my NPP don't tell me there's a new version

At this point in time, that's expected.
New versions always have some "settling time" to see if early adopters have any critical problems.

>Is it possible I'had click one time in "update it later" that desables the [Update Notepad++] option?

No.

More importantly, does your issue not appear in 7.8.7?  Or can you provide more-specific example that will cause it to appear for maybe more than just you?

Tequiller:
I just make same test with the NPP 7.8.7
Bug is actual yet.

[...]

PR #8522, opened 2020-07-04
Fix Flickering cursor after double-clicking a found line

### LINKED ISSUE #8487: Find Results: Flickering cursor after double-clicking a found line
**STR:**
Open file `A`.
Select a word.
Open the Find dialog.
Find.
Close file `A`.
Delete file `A` in Windows Explorer.
In the `Find Results` window double-click a found line under file `A`.

**Result:**
The `File No Longer Exists` message is displayed (correct).
The cursor is intensely flickering. You'd have to exit NPP.
***
Notepad++ v7.8.3   (32-bit)
Build time : Jan 17 2020 - 18:20:02
Path : C:\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.23964

Uhf7:
Fixes #8487.

Same thing as before. The Scintilla double click notifications need their `WM_LBUTTONUP` message badly. This time it has been obfuscated by a `MessageBox` call, which works surprisingly different compared to the browser start when double clicking an URL. This is why we have to use `SendMessage`  here. The `MessageBox` belongs to the Npp process, the browser does not. So the `MessageBox` changes the input focus within Npp, and Scintilla's double click mechanism doesn't accept `WM_LBUTTONUP` messages after input focus change.

_Edit:_ Removed second part of this PR, it didn't work good enough with new ("temporary") buffers.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -489,6 +489,11 @@ bool Finder::notify(SCNotification *notification)
 		{
 			// remove selection from the finder
 			isDoubleClicked = true;
+
+			// WM_LBUTTONUP can go to a "File not found" messagebox instead of Scintilla here, because the mouse is not captured.
+			// The missing message causes mouse cursor flicker as soon as the mouse cursor is moved to a position outside the text editing area.
+			::SendMessage(_scintView.getHSelf(), WM_LBUTTONUP, 0, 0);
+
 			size_t pos = notification->position;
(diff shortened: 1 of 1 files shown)


Uhf7:
The flicker is removed with this PR, and it doesn't mark old find results in the wrong buffer anymore, if the message box

"_File_ doesn't exist. Create it? [Yes] [No]" 

is answered with [No].

What remains is, that Npp invites the user to create a new file in order to mark the old find results, but this cannot be solved as easy as I thought. It seems to be closely related to #8461.

Yaron10:
:+1:

Issue #8477, opened 2020-06-25
did not auto scroll to last line after update


johnw2018:
i found a bug .
Settings > Preferences > MISC -> File Status Auto-Detection
‘Update Silently’ and ‘Scroll to last line after update’
but did not auto scroll to the last line after update

i used 
notepad++ v7.8.7 
windows 7 64bit

Yaron10:
It only occurs if Word-Wrap is ON, and it's a regression in 7.8.3.

Yaron10:
The issue is not fixed with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4b8d83a8f882e6cd7c0acb570f871a519104e691.

donho:
Thank you @Yaron10 for checking it.
I'll take a look on this issue.

PR #8529, opened 2020-07-05
Fix open file command line with scroll option (-n/-l)

### LINKED ISSUE #8476: Open file from command line fails to jump to the correct line if file not already loaded
### Description of the Issue

When working with Cppcheck, notepad++ is used as external editor to edit source files. Since some time, notepad++ no longer jumps to the correct line number passed by option "-n" when the affected source file is not already loaded into notepad++.

1. Create a batch job (Windows .bat) calling notepad++ with a filename and a line number (option -n).
2. Execute the batch job, notepad++ will open the file but not jump to the line number passed.
3. Execute the batch job again - don't close the file before - and notepad++ will now jump to the line number passed.

### Expected Behavior

Notepad++ should always jump to the line number passed by option -n.
### Actual Behavior

The file is opened but positioned at line #1.

### Debug Information

Notepad++ v7.8.6   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.900
Plugins : GhNppExec.dll mimeTools.dll NppConverter.dll NppExport.dll

cmeriaux:
Fixes #8476
The scroll action was canceled by the "restore position"  mechanism. I add an accessor to disable the "restore position" mechanism in case of scroll action passed by command line argument.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5462,7 +5462,7 @@ std::vector<generic_string> Notepad_plus::loadCommandlineParams(const TCHAR * co
                 auto pos = _pEditView->execute(SCI_FINDCOLUMN, ln-1, cn-1);
                 _pEditView->execute(SCI_GOTOPOS, pos);
             }
-
+			_pEditView->setPositionRestoreNeeded(false);
 			_pEditView->scrollPosToCenter(_pEditView->execute(SCI_GETCURRENTPOS));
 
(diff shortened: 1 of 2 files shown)


Yaron10:
@cmeriaux,

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8477 might be related as well.

donho:
Adding the setter for this boolean smells bad.
I'll fix it (by using the hint provided by @Yaron10 ) in my way.

Issue #8471, opened 2020-06-25
Toggle Panel on/off


davidbrigden:
### Some panels can be toggled on/off from menu entries.  Others can not.  Have all toggle

In Notepad ++ there are a number of panels supplementing the editor itself.  All can be opened by selecting in the menu.  Some can be closed again from the menu  
* Project as Workspace
* Document Map
* Function List.  

Others can only be closed by closing the panel directly
* Character Panel
* Clipboard History
* Project Panel (1, 2, 3)
* Search Results

This seems inconsistent.  I think that all panels should toggle on/off.   I say this because we can have multiple panels in a stack docked to the side or bottom of N++.  If you close a docked panel, you close the entire stack. 

I would rather not close an entire stack just to close a single panel.  I also would rather not drag a panel off the stack to close it.  It seems the best way is having the menu entry (or toolbar entry or shortcut) toggle the panel

ArkadiuszMichalski:
@Uhf7  Is there any way to close all panels in stock with one action? When I have a lot of them in one stock now I must click [x] or middle mouse button on each panel. Overall this solution with [x] is better than previouse, but now there is no option to close all :) Maybe with Ctrl + [X] or other key can do it?

PR #8533, opened 2020-07-07
Allow toggle Project Panels from menu.

Uhf7:
Project Panels can now be toggled in the menu like "Folder As Workspace".

That's part of Issue #8471.

That's a big one, but its hard to divide into single small steps. I will write a few comments to the code, but this seems to work only after the PR is created.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -806,17 +806,17 @@ bool Notepad_plus::saveProjectPanelsParams()
 {
 	if (_pProjectPanel_1)
 	{
-		_pProjectPanel_1->checkIfNeedSave(TEXT("Project Panel 1"));
+		if (!_pProjectPanel_1->checkIfNeedSave()) return false;
 		(NppParameters::getInstance()).setWorkSpaceFilePath(0, _pProjectPanel_1->getWorkSpaceFilePath());
 	}
 	if (_pProjectPanel_2)
 	{
-		_pProjectPanel_2->checkIfNeedSave(TEXT("Project Panel 2"));
+		if (!_pProjectPanel_2->checkIfNeedSave()) return false;
 		(NppParameters::getInstance()).setWorkSpaceFilePath(1, _pProjectPanel_2->getWorkSpaceFilePath());
 	}
 	if (_pProjectPanel_3)
 	{
-		_pProjectPanel_3->checkIfNeedSave(TEXT("Project Panel 3"));
+		if (!_pProjectPanel_3->checkIfNeedSave()) return false;
 		(NppParameters::getInstance()).setWorkSpaceFilePath(2, _pProjectPanel_3->getWorkSpaceFilePath());
 	}
 	return (NppParameters::getInstance()).writeProjectPanelsSettings();
@@ -6046,6 +6046,44 @@ void Notepad_plus::launchFileBrowser(const vector<generic_string> & folders, boo
 	_pFileBrowser->setClosed(false);
 }
 
+void Notepad_plus::checkProjectMenuItem()
+{
+	HMENU viewMenu = ::GetSubMenu(_mainMenuHandle, MENUINDEX_VIEW);
+	int viewMenuCount = ::GetMenuItemCount(viewMenu);
+	for (int i = 0; i < viewMenuCount; i++)
+	{
+		HMENU subMenu = ::GetSubMenu(viewMenu, i);
+		if (subMenu)
+		{
+			int subMenuCount = ::GetMenuItemCount(subMenu);
+			bool found = false;
+			bool checked = false;
+			for (int j = 0; j < subMenuCount; j++)
+			{
+				UINT const ids [] = {IDM_VIEW_PROJECT_PANEL_1, IDM_VIEW_PROJECT_PANEL_2, IDM_VIEW_PROJECT_PANEL_3};
+				UINT id = GetMenuItemID (subMenu, j);
+				for (int k = 0; k < _countof(ids); k++)
+				{
+					if (id == ids [k])
+					{
+						found = true;
+						UINT s = GetMenuState(subMenu, j, MF_BYPOSITION);
+						if (s & MF_CHECKED)
+						{
+							checked = true;
+							break;
+						}
+					}
+				}
+			}
+			if (found)
+			{
+				CheckMenuItem(viewMenu, i, (checked ? MF_CHECKED : MF_UNCHECKED) | MF_BYPOSITION);
+				break;
+			}
+		}
+	}
+}
 
 void Notepad_plus::launchProjectPanel(int cmdID, ProjectPanel ** pProjPanel, int panelID)
 {
@@ -6053,7 +6091,7 @@ void Notepad_plus::launchProjectPanel(int cmdID, ProjectPanel ** pProjPanel, int
 	if (!(*pProjPanel))
 	{
 		(*pProjPanel) = new ProjectPanel;
-		(*pProjPanel)->init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf());
+		(*pProjPanel)->init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), panelID);
 		(*pProjPanel)->setWorkSpaceFilePath(nppParam.getWorkSpaceFilePath(panelID));
 
 		tTbData	data;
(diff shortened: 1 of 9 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -806,17 +806,17 @@ bool Notepad_plus::saveProjectPanelsParams()
...
+		if (!_pProjectPanel_2->checkIfNeedSave()) return false;
 		(NppParameters::getInstance()).setWorkSpaceFilePath(1, _pProjectPanel_2->getWorkSpaceFilePath());
 	}
 	if (_pProjectPanel_3)
 	{
-		_pProjectPanel_3->checkIfNeedSave(TEXT("Project Panel 3"));
+		if (!_pProjectPanel_3->checkIfNeedSave()) return false;
 		(NppParameters::getInstance()).setWorkSpaceFilePath(2, _pProjectPanel_3->getWorkSpaceFilePath());
 	}
 	return (NppParameters::getInstance()).writeProjectPanelsSettings();
```
Uhf7, line 822: The changes here are:
- Save operation can be cancelled. If **checkIfNeedSave** returns false, the save operation should be aborted.
- Project Panels have there own (translated) title now, don't need the title parameter anymore.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6046,14 +6046,52 @@ void Notepad_plus::launchFileBrowser(const vector<generic_string> & folders, boo
...
+				}
+			}
+			if (found)
+			{
+				CheckMenuItem(viewMenu, i, (checked ? MF_CHECKED : MF_UNCHECKED) | MF_BYPOSITION);
+				break;
+			}
+		}
+	}
+}
```
Uhf7, line 6087: This new function does serve only one purpose: to add a check mark to the _Project_ menu entry, if one or more of it's sub menu entries are checked. Since the _Project_ menu entry has a submenu instead of an ID, it's hard to access.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6073,14 +6111,10 @@ void Notepad_plus::launchProjectPanel(int cmdID, ProjectPanel ** pProjPanel, int
...
-		if (title_temp.length() < 32)
-		{
-			wcscpy_s(title, title_temp.c_str());
-			data.pszName = title;
-		}
+		generic_string title_no = to_wstring (panelID + 1);
+		generic_string title_temp = pNativeSpeaker->getAttrNameStr(PM_PROJECTPANELTITLE, "ProjectManager", "PanelTitle") + TEXT(" ") + title_no;
+		(*pProjPanel)->setPanelTitle(title_temp);
+		data.pszName = (*pProjPanel)->getPanelTitle();
 		::SendMessage(_pPublicInterface->getHSelf(), NPPM_DMMREGASDCKDLG, 0, reinterpret_cast<LPARAM>(&data));
```
Uhf7, line 6118: Here is a change. The title of all Project Panels were "Project", I changed them to "Project Panel 1", "Project Panel 2" and "Project Panel 3". Easier to tell which is which. Since **data.pszName** points now to different strings for each Project Panel, the title cannot be stored anymore in a single static variable. I included the **_panelName** member to store the title, this comes in handy for the **checkIfNeedSave** function too.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6095,6 +6129,10 @@ void Notepad_plus::launchProjectPanel(int cmdID, ProjectPanel ** pProjPanel, int
 			(*pProjPanel)->openWorkSpace(nppParam.getWorkSpaceFilePath(panelID));
 	}
 	(*pProjPanel)->display();
+
+	checkMenuItem(cmdID, true);
+	checkProjectMenuItem();
+	(*pProjPanel)->setClosed(false);
 }
```
Uhf7, line 6136: This last three new lines in **launchProjectPanel** are actually the same as in **launchFileBrowser**: Check the menu item and set **_isClosed** to **false**.

PowerEditor/src/Notepad_plus.h
```diff
@@ -387,10 +387,6 @@ friend class FileManager;
                                               // and "Enable session snapshot and periodic backup" is not enabled
                                               // then WM_ENDSESSION is send with wParam == FALSE
                                               // in this case this boolean is set true, so Notepad++ will quit and its current session will be saved 
-
-    bool _isWorkspaceFileLoadedFromCommandLine = false; // Set during Notepad_plus::doOpen if workspace file is opened.
-                                                        // But it is only evaluated during startup, when doOpen
-                                                        // has been called while command line interpretation.
 	ScintillaCtrls _scintillaCtrls4Plugins;
```
Uhf7, line 390: This removal is interesting too. I finally dared to move the call of the **loadCommandLineParams** function to a later point, where the plug-ins already have been restored. Due to this move the additional effort to switch back to Project Panel 1 after restoring the plug-insis no longer necessary.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1813,8 +1816,6 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				saveGUIParams(); //writeGUIParams writeScintillaParams
 				saveFindHistory(); //writeFindHistory
 				_lastRecentFileList.saveLRFL(); //writeRecentFileHistorySettings, writeHistory
-				saveProjectPanelsParams(); //writeProjectPanelsSettings
-				saveFileBrowserParam();
 				//
```
Uhf7, line 1819: Here I moved the parameter saving functions for the _Project Panels_ and the _Folder as Workspace_ to an earlier position. Reason: **saveProjectPanelParams** function can be cancelled now, same way as the save of the edit buffer files.

PowerEditor/src/NppCommands.cpp
```diff
@@ -682,18 +683,34 @@ void Notepad_plus::command(int id)
...
+						checkProjectMenuItem();
+					}
+				}
+				else
+				{
+					launchProjectPanel(id, pp [idx], idx);
+				}
+			}
 		}
 		break;
```
Uhf7, line 716: Unified the **IDM_VIEW_PROJECT_PANEL_n** functions here, major difference: They toggle now, and not always launch or switch to foreground. Side effect: If we want to launch a Project Panel now, we cannot call 

`command (IDM_VIEW_PROJECT_PANEL_n)`

 anymore, we have to call **launchProjectPanel** for this. We will see this later.

PowerEditor/src/NppIO.cpp
```diff
@@ -211,10 +211,9 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive,
 	if (isFileWorkspace(longFileName) && PathFileExists(longFileName))
 	{
 		nppParam.setWorkSpaceFilePath(0, longFileName);
-		_isWorkspaceFileLoadedFromCommandLine = true;
 		// This line switches to Project Panel 1 while starting up Npp
 		// and after dragging a workspace file to Npp:
-		command(IDM_VIEW_PROJECT_PANEL_1);
+		launchProjectPanel(IDM_VIEW_PROJECT_PANEL_1, &_pProjectPanel_1, 0);
 		return BUFFER_INVALID;
```
Uhf7, line 217: (Re)launch instead of toggle.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -595,7 +600,12 @@ void ProjectPanel::openSelectFile()
...
 void ProjectPanel::notified(LPNMHDR notification)
 {
-	if ((notification->hwndFrom == _treeView.getHSelf()))
+	if (notification->code == DMN_CLOSE)
+	{
+		::SendMessage(_hParent, WM_COMMAND, IDM_VIEW_PROJECT_PANEL_1 + _panelID, 0);
+		SetWindowLongPtr (getHSelf(), DWLP_MSGRESULT, _isClosed ? 0 : 1);
+	}
+	else if ((notification->hwndFrom == _treeView.getHSelf()))
 	{
```
Uhf7, line 609: The **DWLP_MSGRESULT** mechanism is available in the _DockingManager_ and allows to prevent a tab from being closed. This mechanism gets triggered here if we press the close icon of the side window containing one or more _Project Panels_ or other docking windows. The _Project Panel_ now sends the **IDM_VIEW_PROJECT_PANEL_n** command to close itself, like the FileBrowser already does. If the close failed because the Save-Request has been cancelled, then **DWLP_MSGRESULT** is set to **1**, which prevents the tab from closing.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.h
```diff
@@ -113,9 +121,11 @@ class ProjectPanel : public DockingDlgInterface {
 	HMENU _hProjectMenu = nullptr;
 	HMENU _hFolderMenu = nullptr;
 	HMENU _hFileMenu = nullptr;
+	generic_string _panelTitle;
 	generic_string _workSpaceFilePath;
 	generic_string _selDirOfFilesFromDirDlg;
 	bool _isDirty = false;
+	int _panelID = 0;
```
Uhf7, line 129: Added support for 2 new members needed here: **_panelID** and **_panelTitle**.

PowerEditor/src/NppCommands.cpp
donho: There might be an impact if plugin call it. We'll see.

PowerEditor/src/NppCommands.cpp
```diff
@@ -682,18 +683,34 @@ void Notepad_plus::command(int id)
...
-			launchProjectPanel(id, &_pProjectPanel_3, 2);
+			ProjectPanel** pp [] = {&_pProjectPanel_1, &_pProjectPanel_2, &_pProjectPanel_3};
+			int idx = id - IDM_VIEW_PROJECT_PANEL_1;
+			if (*pp [idx] == nullptr)
+			{
+				launchProjectPanel(id, pp [idx], idx);
+			}
+			else
+			{
+				if (not (*pp[idx])->isClosed())
```
donho, line 697: Next time, please use modern operator (`!` instead of `not`).

sasumner: I tend to like _not_ myself, but if you want ! then now I know it.

Issue #8131, opened 2020-04-18
"Find" settings need their own sub-page in Settings -> Preferences


sasumner:
There are getting to be a few setting related entirely to _Find_ on the _MISC._ page in the _Preferences_ :

![image](https://user-images.githubusercontent.com/30118311/79641410-73a38e00-8165-11ea-9be6-8d59c550ec36.png)

These should be broken out into their own sub-page in the box on the left:

![image](https://user-images.githubusercontent.com/30118311/79641400-553d9280-8165-11ea-9498-69971f35ac43.png)

This also makes room for new "Find" related settings, and frees up space on the `MISC.` subpage for new items there, too.

One new setting could be added as well:

`Keep find window open after search that outputs to Find-result window`

which would control the existing `config.xml` entry:

`<FindHistory ... dlgAlwaysVisible="no/yes" ... />`

sasumner:
Related to #8130

sasumner:
Of course, I have some ideas about new features for "Find" that will necessitate some new Preference settings, so that has something to do with opening this issue.  :-)

donho:
> One new setting could be added as well:
> Keep find window open after search that outputs to Find-result window

This hidden option can be made visible in this panel as well:
https://npp-user-manual.org/docs/preferences/#preferences-for-advanced-users

sasumner:
>Keep find window open after search that outputs to Find-result window

Yes, this was actually in my plan.  :-)

>The hidden option Allow regex backward search can be made visible in this panel as well

Yes, OK.

sasumner:
I will make separate PR for these, after this current PR is accepted:

* Keep find window open after search that outputs to Find-result window
* Allow regex backward search

sasumner:
Changes from this issue released in Notepad++ version 7.9

PR #8541, opened 2020-07-08
Add 'Searching' subpage to Preferences for Find-Replace-Mark settings

sasumner:
Implements #8131

The reason for this is several more configurable settings relating to Find/Replace/Mark will be coming in the future, and the _MISC._ area is out of space.  Plus this is a better logical grouping than with _MISC._

This moves two existing items from _MISC._ to _Searching_.

Looks like this:

image


donho:
Please move 4 checkboxes down to make dialog prettier:
image

PR #8543, opened 2020-07-09
Update italian.xml v7.8.8

Ices-Eyes:
Correct some typos and position of entry 6237


donho:
@Ices-Eyes 
Thank you for your update.
Could you check the diff from v7.8.9 to the newest english.xml to make sure there's nothing's missing?

PR #8551, opened 2020-07-10
Fix italian.xml, position item id=6237

conky77:


conky77:
there is already another PR for the same correction

PR #8552, opened 2020-07-11
Update polish.xml to 7.8.9

ArkadiuszMichalski:
Update polish.xml to 7.8.9 + one small correct.


donho:
@ArkadiuszMichalski 
Thank you for your update. 

ArkadiuszMichalski:
@donho 
Any place where can I download the latest compiled version of NPP? This would make easier to update translation, because without "live preview of the current changes" I can only add the missing entries in their original translation. It would be nice if there was a version for translators before the RC version (2-3 days earlier) so that the update translations could go to the RC version.
 
Btw, what number would be mark next NPP version? In english.xml we still see `version="7.8.7"` which looks out of date.

donho:
@ArkadiuszMichalski It'll be v7.9

> if there was a version for translators before the RC version (2-3 days earlier) so that the update translations could go to the RC version.

The workflow is include everything in the RC.

ArkadiuszMichalski:
@donho 
I'm not sure if everything fit to the GUI (missing binary version), if no I will correct it next time.

PR #8556, opened 2020-07-12
Update italian.xml Set order like English.xml

conky77:


donho:
Since Italian localization is maintained by @Ices-Eyes mainly, I prefer it's him who does such change if he judges necessary.

Issue #8555, opened 2020-07-12
Don't collapse multiline editing on cursor movement


Febbe:
### Description of the Issue
Using Crtl + mouse click to add cursors works just fine, but when using Alt + Mouse there are several issues:
1. It is not possible to mark a block with Mouse + Move + Alt, or Mouse + Alt + Move, only with Alt + Mouse + Move. The Alt should be a toggle.
2. The cursors created with the Alt- Multiline feature collapse, when the cursors are moved with the direction keys, but should all retain like using the Crtl + Mouse Multiline feature.

### Steps to Reproduce the Issue
1. Open a new Document add some Multiline Text 
2. Try to use the multiline feature with the Alt key but click [and move ]the mouse first

2. Create a functional Alt multiline selection (e.g. even 0 x 1+ has this problem)
3. Move it with the arrow keys

### Expected Behavior
1st problem: The order of pressing the buttons doesn't make any difference.
2nd problem: I would exspect my selection to move according to the cursor button directions, instead of collapsing especially for [0 x n] n-line selections,

### Actual Behavior
For the first problem: just nothing happens
For the second: The cursors collaps to one single cursor. Even pressing Ctrl + Alt + Mouse + Move wont create multiple movable cursors.

### Debug Information
not required

sasumner:
These behaviors are largely Scintilla-dependent (Notepad++ 's underlying editing component -- a separate project not owned by Notepad++).  Notepad++ tries to leave this stuff alone, allowing that the Scintilla maintainers know best about how such things should be handled.

To your point 1, Scintilla offers a configuration that will change an already-begun stream selection into a column selection, if the Alt key is pressed while the stream selection is being made.  Details on this are here: https://www.scintilla.org/ScintillaDoc.html#SCI_SETMOUSESELECTIONRECTANGULARSWITCH

To your point 2, again Notepad++ leaves this pretty much to Scintilla, but you might want to try the _Better Multiselection_ plugin for a bit different behavior you might like in this situation.

sasumner:
>Debug Information: not required

Why do you think that?
Debug Info is _always_ required, that's why it is in the issue template.

sasumner:
PR #8557 does about all that can be done to satisfy this issue.

sasumner:
Changes from this issue released in Notepad++ version 7.9

Febbe:
Awesome, thank you for your work :).

KnIfER:
[...]

PR #8557, opened 2020-07-12
Allow user to start making a stream selection and then change to column block

sasumner:
Improves #8555 point 1, in editing views and also in _Find result_ and _Find in these found results_ windows.  My feeling is that if this PR is accepted, 8555 should be closed.

Test by starting to create a normal stream selection with the mouse, then in the middle of dragging to cover text add a press of the Alt key.  Note that the stream selection has now turned into a column block.

Behavior before this PR:  Once a stream selection is started, it couldn't be dynamically changed into a column block selection.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -323,6 +323,10 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_mainEditView.execute(SCI_SETMULTIPASTE, SC_MULTIPASTE_EACH);
 	_subEditView.execute(SCI_SETMULTIPASTE, SC_MULTIPASTE_EACH);
 
+	// allow user to start selecting as a stream block, then switch to a column block by adding Alt keypress
+	_mainEditView.execute(SCI_SETMOUSESELECTIONRECTANGULARSWITCH, true);
+	_subEditView.execute(SCI_SETMOUSESELECTIONRECTANGULARSWITCH, true);
+
 	// Let Scintilla deal with some of the folding functionality
 	_mainEditView.execute(SCI_SETAUTOMATICFOLD, SC_AUTOMATICFOLD_SHOW);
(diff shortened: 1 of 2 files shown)


donho:
@sasumner 
> Test by starting to create a normal stream selection with the mouse, then in the middle of dragging to cover text add a press of the Alt key. Note that the stream selection has now turned into a column block.

I don't follow what this PR does exactly, neither the FR #8555. Could you tell me how I can turn a stream selection into a column block with this PR ?

sasumner:
@donho 

I don't think I can explain it any better than what you quoted from me above, starting with "Test by starting to create..."

@chcg 

Can you add anything to my explanation to help understanding here?

sasumner:
@Ekopalypse , same to you: Can you add anything to my explanation to aid understanding here?

Ekopalypse:
Let's assume we have the following text in a document

line 1
line 2
line 3
line 4 with additional text
line 5
line 6

and do a selection like this.

image

If I now select line 5, I see that this is the wrong selection, because I want to add something after the number, so I need a rectangular selection. The current solution is to start selecting again, but make sure I press the ALT key before I start selecting the text.

This PR would allow you to keep the selection you made before (by holding down the left mouse button) and change it to a rectangular selection at any time after the selection by pressing the ALT key.

image

donho:
@Ekopalypse Thank you for explanation.
@sasumner  It's a neat feature, very cool :)

sasumner:
@donho @Ekopalypse 

>It's a neat feature, very cool

My prior editor had a cool feature regarding this type of thing.
Instead of adding the Alt key after beginning to drag, if you tapped and released the right mouse button (while still holding the left), the selection type would change -- it would cycle through "rectangular block" -> "full lines selection" -> "stream selection" -> (repeat)

I tried a long time ago to Pythonscript that same behavior, without success.
Maybe it could be worked on as a "native" Notepad++ behavior.
Or maybe a challenge to @Ekopalypse to see what really could be done from the scripting side? :-)

Issue #8563, opened 2020-07-14
Mouse right-click cancels selection unless click is inside the selection


sasumner:
### Description of the Issue

A mouse right-click cancels any currently active selection unless the click is inside the contrasting area defining the selection.  Notepad++ shouldn't cancel the selection if the click is outside.

### Steps to Reproduce the Issue
1. Create some text in a new tab, example:

```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```

2. Make some sort of selection of that text, example:

![image](https://user-images.githubusercontent.com/30118311/87473617-bc95a200-c5ef-11ea-8793-4e3d8d306980.png)

3. Right-click a spot in the editing window away from the selection, example is where the red arrow points:

![image](https://user-images.githubusercontent.com/30118311/87473594-b6072a80-c5ef-11ea-92b4-c493bcbafb57.png)

### Expected Behavior

I expected my selection to remain intact, regardless of where I right-clicked inside the editor window.  
I expected the context menu to appear, allowing me to copy my selection.
The context menu _does_ appear, but by then my selection is no longer there so that I can copy it.

### Actual Behavior

Right-clicking outside the selection cancels it, moving the caret to the location clicked (if a valid document position).  This can be seen in the screenshot above; the one with the big red arrow.
As left-click moves the caret for me, I don't need right-click to do that.

### Debug Information

Notepad++ v7.8.9   (32-bit)
Build time : Jul  9 2020 - 21:32:00
Path : C:\\.........\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

### Other

This issue was first raised on the Community site:

https://community.notepad-plus-plus.org/topic/19253/editor-right-click-outside-current-text-selection-lose-this-selection/

donho:
The behaviour you expected was in Scintilla (very) old version.
For me the current behaviour (right-click outside of selection cancels the selection) is more consistence, since it's really "contextual".

sasumner:
Changes from this issue released in Notepad++ version 7.9

PR #8564, opened 2020-07-15
Off-selection right-click should not cancel selection in editor

sasumner:
Implements #8563


PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -513,6 +514,23 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
...
+			{
+				// if right-click in the editing area (not the margins!),
+				// don't let this go to Scintilla because it will 
+				// move the caret to the right-clicked location,
+				// cancelling any selection made by the user
+				return TRUE;
+			}
+			break;
+		}
 	}
```
Uhf7, line 534: The function itself is good, it behaves like WIndows-Notepad now.

I did some experiments to find out why it works, because it looked strange to me. _Why is the text position needed to find out the margin width?_

The `execute(SCI_POINTXFROMPOSITION, clickPos)` is actually wrong. The Scintilla doc says:
`SCI_POINTXFROMPOSITION(<unused>, position pos) → int`.

So the correct call would be `execute(SCI_POINTXFROMPOSITION, 0, clickPos)`. Which wrecks the function.

This code works, because the `execute(SCI_POINTXFROMPOSITION, clickPos)` always returns the x pixel position of the text position 0, which is equal to the margin width.

So the same effect can be reached with less code lines:

		case WM_RBUTTONDOWN:
		{
			int clickX = GET_X_LPARAM(lParam);
			int marginX = static_cast<int>(execute(SCI_POINTXFROMPOSITION, 0, 0));
			if (clickX >= marginX)
			{
				. . .

Yaron10:
@sasumner,

I think that many users would prefer the current behavior.
It has a **major advantage**: you can right-click-copy at line 1 and then right-click-paste at line 2 without the extra left-click.

sasumner:
@Yaron10 

Maybe it is not a good feature after all then.
I did it because of the request for it on the Community.

For myself, I always paste with ctrl+v because with the right-click paste, the paste position gets obscured by the popped up context menu.  I never know truly if I'm pasting in the right spot.  This PR helps with that, so to me it seemed a good thing.

I'm not a big believer in an infinite number of _Preference_ settings, so I think if it comes to that being desired as a way to have both new and old behavior for something like this, we just kill it instead.

Yaron10:
> I did it because of the request for it on the Community.

The request there is legit, but most of the participant did not understand the need for a change.
Your work is appreciated anyway. :)

> I never know truly if I'm pasting in the right spot.

Increasing the caret width might help. 

> I'm not a big believer in an infinite number of Preference settings, so I think if it comes to that being desired as a way to have both new and old behavior for something like this, we just kill it instead.

:+1: 

***

Dealing with it might help you solve another related issue.
Right-clicking in the Lines-Margin on the *first line* of a selection removes the selection.
I can open a separate issue.

sasumner:
@Yaron10 

>Increasing the caret width might help.

I already have the maximum caret width and I have changed the caret width to bright red, since a long time ago.

>Right-clicking in the Lines-Margin on the first line of a selection removes the selection.  I can open a separate issue.

Probably best not to open an issue on this.  It is Scintilla behavior so it would just get labeled "Scintilla dependent" and nothing would ever happen on it.

Yaron10:
>  It is Scintilla behavior so it would just get labeled "Scintilla dependent" and nothing would ever happen on it.

:+1:

donho:
I'm not against this feature. But if it should be optional between `Enable scrolling beyond last line` & `Disable advanced scrolling feature`.

sasumner:
@donho Okay, I will add the _Preference_ stuff for it.

sasumner:
@donho 

Preference control of this feature has been added.
I called it `Right-click does not move caret` because I could not think of anything better.

donho:
@sasumner 
> I called it Right-click does not move caret because I could not think of anything better.

Even it's the most accurate description for this feature, I would suggest:
`Keep selection while right-clicked outside of selection` 
for the following reasons:
1. It's the most understandable description for the users
2. People see more the context menu and observe more the selection than the change of caret position.
3. People who want to change this option for keeping the selection but not for keeping the caret position.

What do you think?

sasumner:
Yes, yours is good.
I'll rework the user interface text.
More proper English version of yours is `Keep selection when right-clicking outside of selection`

donho:
I think the period should be removed:
image

donho:
image

Sorry for the request again: please make this control label 2 lines.

sasumner:
>Sorry for the request again: please make this control label 2 lines

No problem.
Sorry for my poor testing; I should have seen that this was a problem, for such long text.
In the fix it made sense to adjust the Y coordinates on several controls.

PR #8584, opened 2020-07-18
Fix URL is opened inadvertently when clicking white space outside document

### LINKED ISSUE #8581: Documents with a URL as the first line will open that URL when white space is clicked elsewhere in the document
### Description of the Issue

When the first line of a new file is a URL, double-clicking white space anywhere after the URL in the document will open that url.

1.  Open a new document
2.  Type: http://www.google.com, then return a few times to create some new empty lines.  Heck, add more content if you want.
3.  Double click on new lines near the EOL.

### Expected Behavior

Nothing.

### Actual Behavior

URL in the first line opens in a new browser tab.

Note: moving the url to the second line of code or adding a space before the URL prevents this issue from happening.

### Debug Information

Notepad++ v7.8.7   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1809
OS Build : 17763.1282
Current ANSI codepage : 1252

Uhf7:
Fixes #8581.

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -817,18 +817,19 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				// Check wether cursor is within URL
 				auto indicMsk = notifyView->execute(SCI_INDICATORALLONFOR, notification->position);
 				if (!(indicMsk & (1 << URL_INDIC))) break;
-
-				// Revert selection of current word. Best to this early, otherwise the
-				// selected word is visible all the time while the browser is starting
-				notifyView->execute(SCI_SETSEL, notification->position, notification->position); 
+				auto startPos = notifyView->execute(SCI_INDICATORSTART, URL_INDIC, notification->position);
+				auto endPos = notifyView->execute(SCI_INDICATOREND, URL_INDIC, notification->position);
+				if ((notification->position < startPos) || (notification->position > endPos)) break;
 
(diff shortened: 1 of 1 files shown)


donho:
@Uhf7 After applying this PR it seems the issue is fixed. However, if 4 lines have given after the url, then you double-click on the 1st, the 2nd or the 3rd line, the cursor will jump to the 4th line. Replace URL by any non-URL string then such behaviour stops.
Could this issue be fixed as well?

Uhf7:
Thank you, I would never have tried this.

PowerEditor/src/NppNotification.cpp
```diff
@@ -818,19 +818,22 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				auto indicMsk = notifyView->execute(SCI_INDICATORALLONFOR, notification->position);
 				if (!(indicMsk & (1 << URL_INDIC))) break;
 
-				// Revert selection of current word. Best to this early, otherwise the
-				// selected word is visible all the time while the browser is starting
-				notifyView->execute(SCI_SETSEL, notification->position, notification->position); 
-
-				// WM_LBUTTONUP goes to opening browser instead of Scintilla here, because the mouse is not captured.
+				// WM_LBUTTONUP goes to opening browser instead of Scintilla here, because the mouse 	is not captured.
```
donho, line 821: This line should not be changed.

Issue #5721, opened 2019-05-31
Indenting a single selected line with the tab key doesn't work


EliasHasle:
When selecting multiple lines, or a single line including the trailing line break, the selection is indented when pressing tab. However, when selecting a single line without the trailing line break, the selection is instead replaced with four spaces. I find this counter-intuitive.

Notepad++ version is 7.6.6 (64-bit). This is on Windows 10. I think I have no reason to believe this is platform-dependent.

(This post and its title was edited following the below discussion.)

sasumner:
If you have the entire line including line-ending selected, a press of the Tab key will indent the selected line.   If you have the line only partially selected (and this could include selecting everything EXCEPT the line-ending--so it appears as if the whole line is selected) and press Tab, then the line disappears and is replaced with 4 spaces as @EliasHasle points out.  Personally, I'm OK with this behavior.

Side note: I often exploit the somewhat unusual behavior of where you have a PARTIAL first line selected, then several complete lines, then a partial LAST line selected...and press Tab, and voila, the lines touched by selection all get indented one "tab stop".

EliasHasle:
I can confirm that selecting the line with the break works.

Now I tried selecting the whole line, without the line break, in a non-Python file, for which the four spaces replacement rule is not enabled, and the behavior is the same. The whole line is replaced by a tab.

If this is indeed the desired behavior, I guess this issue may be closed. (Personally, I would prefer having the line indented when it is selected to the end, even without the line break.)

sasumner:
If you are satisfied that this works reasonably (although maybe not to your preference), then YOU should close this issue.

I'm curious how your workflow goes such that you get the line selected without the line-ending.  Sure you can do it, but it is not nearly as easy as selecting the line WITH the line-ending selected as well.

EliasHasle:
No, I am not satisfied. But the issue report may be resting on false premises, in the sense that it is not specific to the tabs-to-spaces setting. If that is the case, I am surprised I did not encounter the problem before.

Shift+End selects the rest of the line.

sasumner:
>But the issue report may be resting on false premises,

So just change the title of the issue report then, to remove the tab/spaces reference. :)

>Shift+End selects the rest of the line.

So in your opinion, if you have a line like the following:

    bbbbif (x==0) { y=3; }

(Note that the `bbbb` at the beginning is 4 spaces representing a "tab", or maybe it is one true tab character...)

[...]

PR #8623, opened 2020-07-27
Improve single-line indentation and dedentation functionality

sasumner:
Implements #5721


Yaron10:
@sasumner,

I haven't tested it yet.
If a line is selected without the EOL and you press `Tab`, the line should be *kept* and indented. Is that correct?

`Edit -> Increase Line Indent`: shouldn't the line be indented (as it is) even if it's *not selected*?
I've been using a Python-Script implementing it and I think that this is the expected behavior.

sasumner:
@Yaron10 The best way to understand it is to experiment with it. :-)

But:

No selection or a single selection only on one line -> indent/dedent that single line, maintaining caret position and/or selection the way it was. (New behavior as desired by this issue and the community posting I referenced in the issue)

A selection of more than one line -> indent/dedent the lines as a group (retains existing behavior of Scintilla SCI_TAB / SCI_BACKTAB).

Also, keycombos are decoupled:  After this Tab key defaults to SCI_TAB, Shift+Tab defaults to SCI_BACKTAB, "Indent" has no default keycombo, "Dedent" has no default keycombo.  I'm not really sure there was a good reason for having the SCI_* functions tied to the menu items for indent/dedent.

Yaron10:
@sasumner,

After testing and experimenting: excellent! :+1: 

Thank you.

donho:
@sasumner 
I don't feel right for the behaviour of this PR.
The indent action should be only one (by TAB). The dissociation of TAB key with "Insert Line indent" command will create confusion.

sasumner:
@donho said:

>I don't feel right for the behaviour of this PR.

So this PR does **_not_** change _Tab_ key (or _Shift+Tab_ key) behavior.
Those keys function as SCI_TAB and SCI_BACKTAB per _Scintilla_ page of _Shortcut Mapper_, as they always have.

This PR allows an enhanced indention/dedention behavior, via the _Edit_ menu entries, that users have called for (see discussion and links in the linked issue).  The repurposing of the _Edit_ menu entries allows these new functions to be shortcut-mapped without adding any new entries anywhere.

I actually thought it was confusing before this PR, where we had _Tab_ and _Shift+Tab_ in TWO places (directly as Scintilla commands -- and also on the _Edit_ menu).

So old behavior is retained, and an optional new behavior is added.  What's not to like? :-)

donho:
The behaviour of this PR is confusing, because both `TAB` (plus Shift-TAB) and `Increase Line Indent` (plus `Decrease Line Indent`) command do the same thing on selected multi-lines mode, BUT do the different things on the other mode.
This PR could be accepted if  `SCI_TAB` had been changed also its behaviour to aligne with `Increase Line Indent`.
(but then we need to hack in Scintilla code, that I try to avoid).

Yaron10:
@donho,

That's a good point. :+1: 

However, this is the behavior in MS Visual Studio and other editors.
I think the logic is as follows:

If the explicit `Indent` is used, `Indent` it must be in all cases.
If a line is selected *without the EOL* and you press `Tab`, it's NOT `Indent` because for indenting a line you must indent the EOL as well.
If a line is selected *with the EOL* and you press `Tab`, we assume it *is* `Indent` as is the case in multiple selected lines.

donho:
@Yaron10 
> However, this is the behavior in MS Visual Studio and other editors

Indeed MS VS is has the same behaviour. Could you provide me the name of "other editors" please ?

Yaron10:
@donho,

EditPad.

donho:
> EditPad.

Now I'm asking myself why should I listen an unfaithful Notepad++ user :P 
Joke aside, argument accepted.

Yaron10:
Notepad++ is a superb editor.
We appreciate your work.

EditPad has a good RTL support.

sasumner:
@Yaron10 

>EditPad has a good RTL support.

Whoa!  You kinda just slipped that one in there.  
As a major hint, I'm sure.  LOL

sasumner:
@donho 

>Now I'm asking myself why I should listen an unfaithful Notepad++ user.

Acceptance of this PR just gave @Yaron10 one more reason to be a more faithful N++ user.  :-)

BTW, knowledge of other editors is not unfaithfulness.
We appreciate when those with knowledge of good features of other editors bring them to us.
In other industries this is called "competitive analysis" -- being familiar with the rivals, and stealing their best features to make one's own product better.   :-)
BUT...I do hate it when the only argument for something is that "Visual Studio does it that way". :-(

sasumner:
@donho You should also close #5721 with acceptance of this PR.

Yaron10:
**I swear allegiance to Notepad++, to its author and all developers. I swear that I will faithfully follow the RULES. So help me God almighty!**
***
@sasumner,

> Whoa! You kinda just slipped that one in there.
As a major hint, I'm sure. LOL

Absolutely not! That's not me.
I actually wanted to write "EditPad has a good RTL support (I know it's the Scintilla component in Notepad++).", but was too lazy.
***
> Acceptance of this PR just gave @Yaron10 one more reason to be a more faithful N++ user.

I do think it's one of the most important improvements to NPP since I've started using it.

donho:
@sasumner 
> BUT...I do hate it when the only argument for something is that "Visual Studio does it that way". :-(

Well, the convention from a famous one helps the decision, but not only.
As you did, I was not comfortable about this feature on the first sight. All we need is just the time to get used to it.

PR #8636, opened 2020-07-29
Update slovak.xml to 7.8.9

kubalav:


donho:
@kubalav 
Thank you for your update. 

Issue #8632, opened 2020-07-28
Add scope info to ReplaceAll / Count / Mark search output in Find window status bar


sasumner:
### Description of the Issue
There are various search scopes possible for the _Replace All_, _Count_ and _Mark_ searching functions; they are:

* Selected text (_In selection_ ticked)
* Entire document (_Wrap around_ ticked and _In selection_ not ticked)
* BOF-to-caret (_Backward direction_ ticked and _Wrap around_ and _In selection_ not ticked)
* caret-to-EOF (none of _Backward direction_ , _Wrap around_ , _In selection_ ticked)

It would help the user to output the scope that was searched along with the search results in the Find window status bar.

This allows a user to confirm that their settings were correct (i.e., what they intended), and also provides a more complete picture of the action just completed.

So an example might be, instead of outputting:

`Count: 2 matches`

Output this:

`Count: 2 matches in selected text`

Note: This issue originated from the discussion here: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4eac7d8e94f7e1e32f67ddbe611a95e6764d0ade#commitcomment-40968491

### Debug Information
Issue based on functionality in:

Notepad++ v7.8.9   (32-bit)
Build time : Jul 15 2020 - 20:26:50
Path : C:\\.........\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll

sasumner:
Changes from this issue released in Notepad++ version 7.9

PR #8639, opened 2020-07-30
Add search scope to Find window status bar output

sasumner:
Implements #8632 

Some sample output:

image

image

image

image


donho:
The periods remain in the labels:
image

sasumner:
>Remove the periods.

Sorry.  It is interesting that the default text in the code itself had the periods, but the localization files (e.g. english.xml) do not have them.

PR #8641, opened 2020-07-30
Fix mouse cursor flicker while hovering

### LINKED ISSUE #8588: on clickable links - mouse cursor is flickering
I have "clickable links" enabled,

and have noticed significant mouse cursor flickering when mouse is over the link
in Notepad++

It hasn't been before, I suspect some new version of Notepad++ started this...

Anybody experience the same? :-( 
Please fix this...

Uhf7:
Fixes #8588 and #8647. 

The reason, why the flicker occurs, is the following: There are two Windows messages sent by Windows to Scintilla, when the mouse is moved: `WM_MOUSEMOVE` and `WM_SETCURSOR`.

- `WM_MOUSEMOVE` informs Scintilla, that the mouse has been moved inside its window.
- `WM_SETCURSOR` informs Scintilla, that the mouse cursor has been moved. Anywhere, not necessarily in its window.

Scintilla calls the Windows function `SetCursor` (, which sets the mouse cursor shape, not the position), while processing both messages. Unfortunately, Scintilla uses different ways to calculate the desired cursor shape. So, whenever the mouse cursor is moved, two `SetCursor` calls are applied, sometimes with two different cursor shapes.

On `WM_MOUSEMOVE`,  Scintilla calls `ButtonMoveWithModifiers`, which sets the correct cursor shape.
On `WM_SETCURSOR`, Scintilla calls `SetCursor` directly, sometimes with the wrong cursor shape.

This PR shows how to eliminate this effect in the modified Scintilla version used by Notepad++. This may be the fastest way to get results without introducing new effects.

The current original Scintilla version (Version 4.4.4, downloaded 2020-07-30, 5d134721c303ceecbdcb28ec82b28f0cbbdb4a55) has the same effect and can be fixed in the same way, although the `WM_SETCURSOR` code has been changed a little. Before updating to a new Scintilla version, we should try to get it fixed in the original Scintilla.

### DIFF
--- a/scintilla/src/Editor.cxx
+++ b/scintilla/src/Editor.cxx
@@ -4486,6 +4486,12 @@ void Editor::DwellEnd(bool mouseMoved) {
 
 void Editor::MouseLeave() {
 	SetHotSpotRange(nullptr);
+	// Fix the following issue:
+	// #8647
+	// It could be fixed in Scintilla release superior to v4.4.4 - to verify.
+	// ++ added
+	SetHoverIndicatorPosition(INVALID_POSITION);
+	// -- added
 	if (!HaveMouseCapture()) {
 		ptMouseLast = Point(-1, -1);
 		DwellEnd(true);
@@ -4675,6 +4681,12 @@ void Editor::ButtonDownWithModifiers(Point pt, unsigned int curTime, int modifie
 	lastClick = pt;
 	lastXChosen = static_cast<int>(pt.x) + xOffset;
 	ShowCaretAtCurrentPosition();
+	// Fix the following issue:
+	// #8647
+	// It could be fixed in Scintilla release superior to v4.4.4 - to verify.
+	// ++ added
+	if (inSelMargin) SetHoverIndicatorPosition(INVALID_POSITION);
+	// -- added
 }
 
 void Editor::RightButtonDownWithModifiers(Point pt, unsigned int, int modifiers) {
@@ -4848,18 +4860,37 @@ void Editor::ButtonMoveWithModifiers(Point pt, unsigned int, int modifiers) {
 			}
 			hotSpotClickPos = INVALID_POSITION;
 		}
-
+		// Fix the following issue:
+		// #8647
+		// It could be fixed in Scintilla release superior to v4.4.4 - to verify.
+		// ++ added
+		if ((vs.fixedColumnWidth > 0) && PointInSelMargin(pt)) {
+			SetHoverIndicatorPosition(INVALID_POSITION);
+		}
+		// -- added
 	} else {
 		if (vs.fixedColumnWidth > 0) {	// There is a margin
 			if (PointInSelMargin(pt)) {
 				DisplayCursor(GetMarginCursor(pt));
 				SetHotSpotRange(nullptr);
+				// Fix the following issue:
+				// #8647
+				// It could be fixed in Scintilla release superior to v4.4.4 - to verify.
+				// ++ added
+				SetHoverIndicatorPosition(INVALID_POSITION);
+				// -- added
 				return; 	// No need to test for selection
 			}
 		}
 		// Display regular (drag) cursor over selection
 		if (PointInSelection(pt) && !SelectionEmpty()) {
 			DisplayCursor(Window::cursorArrow);
+			// Fix the following issue:
+			// #8647
+			// It could be fixed in Scintilla release superior to v4.4.4 - to verify.
+			// ++ added
+			SetHoverIndicatorPosition(INVALID_POSITION);
+			// -- added
 		} else {
 			SetHoverIndicatorPoint(pt);
 			if (PointIsHotspot(pt)) {
--- a/scintilla/win32/ScintillaWin.cxx
+++ b/scintilla/win32/ScintillaWin.cxx
@@ -1463,7 +1463,16 @@ sptr_t ScintillaWin::WndProc(unsigned int iMessage, uptr_t wParam, sptr_t lParam
 						} else if (PointIsHotspot(PointFromPOINT(pt))) {
 							DisplayCursor(Window::cursorHand);
 						} else {
-							DisplayCursor(Window::cursorText);
+							// Fix the following issue:
+							// #8588
+							// It should be fixed in Scintilla release superior to v4.4.4 - to verify.
+							// ++ added
+							Sci::Position pos = PositionFromLocation(PointFromPOINT(pt), true, true);
+							if ((pos != INVALID_POSITION) && (hoverIndicatorPos != Sci::invalidPosition))
+								DisplayCursor(Window::cursorHand);
+							else
+							// -- added
+								DisplayCursor(Window::cursorText);
 						}
 					}
 				}


Uhf7:
The effect is fixed now in the original Scintilla too, so you have the choice between fixing it in your modified 4.2.0 or upgrading to the current version of Scintilla. I would prefer the fastest way, which is in my opinion to fix the already modified version, just to see whether it works on the many different systems out there.

 The fixed original version of Scintilla I refer to here is still 4.4.4, downloaded 2020-07-31, 7b1106eeebbb7b425cf336e87a4b8fa54da2fe66. I tested it on my system, and it worked as expected. The Scintilla issue providing more information about this is https://sourceforge.net/p/scintilla/bugs/2170/.

Uhf7:
There is one new effect introduced with this modification, which is much less annoying than the flicker, but we should try remove it. This effect is: If the clickable link is placed at the end of file (no non-link character after it), and a mouse click is done to the area after the end of file, the cursor switches to the hand icon. It switches back, as soon as the mouse is moved. I didn't see it in the first place, but I it can be removed easily.

There is another old effect too. This is described in #8647, seems to be in Scintilla too, so while we are at it, perhaps we should try to solve the hovering issues of the standard indicators all at once, at least the ones we know right now.

donho:
> The effect is fixed now in the original Scintilla too, so you have the choice between fixing it in your modified 4.2.0, or upgrading to the current version of Scintilla. I would prefer the fastest way, which is in my opinion to fix the already modified version, just to see whether it works on the many different systems out there.

I just tried Scintilla 4.4.4, source code download from here:
https://www.scintilla.org/ScintillaDownload.html

And I don't see the fix from the binary I compiled. Are you sure about it?
[SciLexer.444.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5006870/SciLexer.444.zip)

Uhf7:
I think you have the release 4.4.4. I downloaded the current version from the https://sourceforge.net/p/scintilla/code/ci/default/tree/ page, using the function "Download snapshot".

donho:
OK, I will wait for 4.4.5 for updating Scintilla then.

Uhf7:
I'm going to push the fixes for the remaining changes here, so that the artifacts can be used for testing, and try to get the fixes included in the original Scintilla then.

donho:
Thank you. Does the latest commit fix #8647 ?

Uhf7:
@Yaron10 says yes, My tests saying yes too. I just created an ticket for this in Scintilla:
https://sourceforge.net/p/scintilla/bugs/2193/

Uhf7:
I just pushed the latest version of my Scintilla modifications for testing, they should solve the two effects mentioned last by @Yaron10 in #8647.

scintilla/win32/ScintillaWin.cxx
```diff
@@ -1463,7 +1463,11 @@ sptr_t ScintillaWin::WndProc(unsigned int iMessage, uptr_t wParam, sptr_t lParam
 						} else if (PointIsHotspot(PointFromPOINT(pt))) {
 							DisplayCursor(Window::cursorHand);
 						} else {
-							DisplayCursor(Window::cursorText);
+							Sci::Position pos = PositionFromLocation(PointFromPOINT(pt), true, true);
```
donho, line 1466: One more stuff, in order to make future Scintilla updating easier, could you add the following comment:
// Fix the following issue:
// https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8588
// It should be fixed in Scintilla release superior to v4.4.4 - to verify.

scintilla/src/Editor.cxx
```diff
@@ -4486,6 +4486,7 @@ void Editor::DwellEnd(bool mouseMoved) {
 
 void Editor::MouseLeave() {
 	SetHotSpotRange(nullptr);
+	SetHoverIndicatorPosition(INVALID_POSITION);
```
donho, line 4489: One more stuff, in order to make future Scintilla updating easier, could you add the following comment:
// Fix the following issue:
// https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8647
// It could be fixed in Scintilla release superior to v4.4.4 - to verify.

scintilla/src/Editor.cxx
```diff
@@ -4675,6 +4676,7 @@ void Editor::ButtonDownWithModifiers(Point pt, unsigned int curTime, int modifie
 	lastClick = pt;
 	lastXChosen = static_cast<int>(pt.x) + xOffset;
 	ShowCaretAtCurrentPosition();
+	if (inSelMargin) SetHoverIndicatorPosition(INVALID_POSITION);
```
donho, line 4679: One more stuff, in order to make future Scintilla updating easier, could you add the following comment:
// Fix the following issue:
// https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8647
// It could be fixed in Scintilla release superior to v4.4.4 - to verify.

scintilla/src/Editor.cxx
```diff
@@ -4849,17 +4851,22 @@ void Editor::ButtonMoveWithModifiers(Point pt, unsigned int, int modifiers) {
 			hotSpotClickPos = INVALID_POSITION;
 		}
 
+		if ((vs.fixedColumnWidth > 0) && PointInSelMargin(pt)) {
```
donho, line 4854: One more stuff, in order to make future Scintilla updating easier, could you add the following comment:
// Fix the following issue:
// https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8647
// It could be fixed in Scintilla release superior to v4.4.4 - to verify.

scintilla/src/Editor.cxx
```diff
@@ -4849,17 +4851,22 @@ void Editor::ButtonMoveWithModifiers(Point pt, unsigned int, int modifiers) {
...
 
+		if ((vs.fixedColumnWidth > 0) && PointInSelMargin(pt)) {
+			SetHoverIndicatorPosition(INVALID_POSITION);
+		}
 	} else {
 		if (vs.fixedColumnWidth > 0) {	// There is a margin
 			if (PointInSelMargin(pt)) {
 				DisplayCursor(GetMarginCursor(pt));
 				SetHotSpotRange(nullptr);
+				SetHoverIndicatorPosition(INVALID_POSITION);
```
donho, line 4862: One more stuff, in order to make future Scintilla updating easier, could you add the following comment:
// Fix the following issue:
// https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8647
// It could be fixed in Scintilla release superior to v4.4.4 - to verify.

scintilla/src/Editor.cxx
```diff
@@ -4849,17 +4851,22 @@ void Editor::ButtonMoveWithModifiers(Point pt, unsigned int, int modifiers) {
...
 				return; 	// No need to test for selection
 			}
 		}
 		// Display regular (drag) cursor over selection
 		if (PointInSelection(pt) && !SelectionEmpty()) {
 			DisplayCursor(Window::cursorArrow);
+			SetHoverIndicatorPosition(INVALID_POSITION);
```
donho, line 4869: One more stuff, in order to make future Scintilla updating easier, could you add the following comment:
// Fix the following issue:
// https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8647
// It could be fixed in Scintilla release superior to v4.4.4 - to verify.

donho: Please add whole link instead of issue number.

donho:
Thank you @Uhf7  !

Issue #7908, opened 2020-01-31
Control over "Find" dialogue box behaviour.


vikrant123pundir:
### Expected Behavior
I want "Find" dialogue box to not to get closed untill I hit "escape" key or "close" button.

### Actual Behavior
Find dialogue box closes by itself after "Find All in All Documents" & "Find All in Current Document" commands.

### My Solution
I know that for programmers, the present scenario is ideal. But I use notepad++ for research purposes & it is one of the core application of my workflow.
I would be really happy if there was a checkbox under "Settings"--->"Preferences..."--->"MISC." for controlling "Find" dialogue box behaviour.
That will not upset the programmers either.

sasumner:
Quit Notepad++, open config.xml in notepad.exe, change the value of the existing parameter:

`dlgAlwaysVisible="no"`

to 

`dlgAlwaysVisible="yes"`

I believe this is in the new help documentation.

sasumner:
Reopening -- I will make this behavior controllable via a _Preference_ setting on the newly-created _Searching_ subpage (from issue 8131).

sasumner:
Changes from this issue released in Notepad++ version 7.9

PR #8651, opened 2020-07-31
Make (Find) dlgAlwaysVisible a true Preference setting

sasumner:
Implements #7908 

Removes the old hidden way of providing the same functionality by hand-editing `config.xml` and setting `FindHistory` section's `dlgAlwaysVisible` to `yes`.

Here's what the new _Preference_ looks like:

image


sasumner:
@donho 
This one has been inactive for some time, wondering if there is a problem?
I thought it would be an "easy" one to accept, as you seemed to be in favor of it here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8131#issuecomment-664009375

donho:
I let you modify the description about `dlgAlwaysVisible` in:
https://npp-user-manual.org/docs/searching/#dialog-based-searching

sasumner:
>I let you modify the description about dlgAlwaysVisible in:  https://npp-user-manual.org/docs/searching/#dialog-based-searching

Okay, I will do so.
BTW, @pryrt , I have added a number of new features lately, but I have forgotten about updating the manual.
Is this something you are slowly taking care of (I haven't noticed) or (probably) I should be doing it?

pryrt:
If it's mentioned in release notes, I try to get to it (I try to modify with the RC release and merge the PR after the final release)... but I don't think I made any specific updates for v7.8.8 - v7.8.9 (maybe because I thought the change list in the forum announcement didn't sound like new features... but maybe I was just too busy to check the manual).

All that to say, if you've added features and don't see a PR that updated the docs, you should go ahead and add them in. :-)

sasumner:
@pryrt said

>I thought the change list in the forum announcement didn't sound like new features...

Don puts that together, and sometimes is rather "brief". :-)
Actually @guy038 does a nice job, post-release, of linking the change log items to issues on github, so it is fairly easy to get a more "in-depth" view of what something is, and if it affects the manual.

Issue #8617, opened 2020-07-27
Unsaved file closed on "Close All BUT this" if also a non-existing file is open


GunnarBernsteinHH:
### Description of the Issue

When you have 2 or more files open you can use "Close All BUT this" to close everything else. 
The currently selected tab should stay open. Especially when it is a new unsaved tab.

### Steps to Reproduce the Issue

1. Open any text file as tab 1. 
2. Open a new tab and write something in it as tab 2, it will be named "new 1".
3. Delete the text file of tab 1 in the file system.
4. Right-click on the unsaved tab 2 and do "Close All BUT this"

### Expected Behavior
Tab 1 will be closed.
Tab 2 will remain WITH content.

### Actual Behavior
1. A question appears "The File1 does not exists anymore. Do you want to keep it in the editor?"
   This question itself is useless as I already ordered it to be closed.
2. Confirm with "No". The first tab will be closed.
3. The "new 1" Tab will also be closed. **All content is lost.**
4. A new empty "new 2" tab will be created

### Debug Information
Notepad++ v7.8.7   (64-bit)
Build time : Jun  5 2020 - 02:38:56
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 7 Ultimate (64-bit) -> actually Windows 10
OS Version : 1903
OS Build : 7600.900
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll NppXmlTreeviewPlugin.dll XMLTools.dll

Ekopalypse:
I can see that behavior if the position of file, which has been deleted already, is before the new file and in the same view.
If the position is after the new file or if the file is on the other view then the new file is not closed.

OK:
new 1 | file_which_has_bee_already_deleted_on_disk.txt
new 1 (in view 0) and file_which_has_bee_already_deleted_on_disk.txt (in view 1)

BUG:
file_which_has_bee_already_deleted_on_disk.txt | new 1

tested on:
Notepad++ v7.8.9   (64-bit)
Build time : Jul  9 2020 - 21:26:26
Path : D:\Tests\npp\_latest\x64_clean\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.24544
Current ANSI codepage : 1252
[...]

PR #8657, opened 2020-08-02
Fix for issue 8617

Roshni-P:
The active index should be updated correctly for the scenario specified in Issue #8617.
Updating the active index ensured that the contents of the New Unsaved file were not lost.


donho:
Your PR insert a regression : if no file is deleted, the current tab is closed after running "Close All but this" command.

sasumner:
>Your PR insert a regression

Plus, it needs to follow the contributing guidelines better.

Yaron10:
@donho,

> if no file is deleted, the current tab is closed after running "Close All but this" command.

I get the same result with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a82819972a6388668a9ce823738aa24c5572d8f4. Anything I'm missing?

Yaron10:
Fixed with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/1eb7d2bef5e7c42c071379154fe11fa30166162a.

PR #8663, opened 2020-08-05
update korean 7.8.9

yunbj:


donho:
@yunbj 
Thank you for your update. 

yunbj:
deprecated.

#8832

PR #8669, opened 2020-08-05
Fix URL detection in first line of document.

### LINKED ISSUE #8668: URL link detection may not work in first line of document
### Description of the Issue
In unwrapped view, URL links are not detected correctly, if the beginning of the URL is not visible in the edit window because of horizontal scrolling.

1. Create new edit buffer.
2. Turn word wrap off and URL underlining on.
3. Type any long URL like http://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
4. Make sure, that the link is in the first line of the document and that there is only this one line in the document.
5. Resize Notepad++, until its width is significantly smaller than the URL.
6. Go to the end of the line and type a space. 
7. Move text cursor to the left with arrow key.

### Expected Behavior
URL remains underlined or highlighted.

### Actual Behavior
As soon as the start of the URL becomes invisible because of horizontal scrolling, the end of the URL becomes unmarked.

This behavior is for URL(s) in the first line only. The effect has been introduced inadvertently with PR _Edit:_ #8467, which fixed _Edit:_ #5503. 

### Debug Information
Notepad++ v7.8.9   (32-bit)
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Current ANSI codepage : 1252
Plugins : none

Uhf7:
Fixes #8668. 

The function `ScintillaEditView::getVisibleStartAndEndPosition` is used to detect the text range processed by `Notepad_plus::addHotSpot`. It is not used for any other purposes.

The start of the range detected with `SCI_POSITIONFROMPOINT` does not include the beginning of the line, if it is invisible due to horizontal scrolling.

The end of the range, detected with `SCI_POSITIONFROMPOINT` too, contains the end of the last visible line only, if word wrap is turned off. With word wrap, the end of the current line may not be included if it's not visible.

Now, the function returns the range from the beginning of the first line partially visible to the end of the last line partially visible, and this whether wrapped or not. Sorry for disregarding this in my PR #8467.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -2260,9 +2260,12 @@ void ScintillaEditView::getVisibleStartAndEndPosition(int * startPos, int * endP
 	// Get the position of the 1st and last showing chars from the edit view
 	RECT rcEditView;
 	getClientRect(rcEditView);
-	*startPos = static_cast<int32_t>(execute(SCI_POSITIONFROMPOINT, 0, 0));
-	*endPos = static_cast<int32_t>(execute(SCI_POSITIONFROMPOINT, rcEditView.right - rcEditView.left, rcEditView.bottom - rcEditView.top));
-
+	LRESULT pos = execute(SCI_POSITIONFROMPOINT, 0, 0);
+	LRESULT line = execute(SCI_LINEFROMPOSITION, pos);
+	*startPos = static_cast<int32_t>(execute(SCI_POSITIONFROMLINE, line));
+	pos = execute(SCI_POSITIONFROMPOINT, rcEditView.right - rcEditView.left, rcEditView.bottom - rcEditView.top);
+	line = execute(SCI_LINEFROMPOSITION, pos);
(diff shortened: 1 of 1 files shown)


sasumner:
@Uhf7 the `getVisibleStartAndEndPosition ` got me thinking about what happens if a URL is so long that its left side is not in the user's view and its right side is also not in the user's view (in this case we definitely have a horizontal scrollbar visible).  So, like this, where the "box" is what the user sees in their editing window:

```
                                 +--------------------------------+
                                 |                                |
https://www.google.com/search?q=blah&rlz=1C1GCEB_enUS851US851&oq=blah&aqs=
                                 |                                |
                                 +--------------------------------+
```

Does this case work appropriately for all of your URL detection code changes?
Are you glad you got involved in changing that aspect of N++? :-)

Uhf7:
@sasumner, That's a cool special case, but it should be covered by being `line` the same for the start and the end of the range. So, the start is `SCI_POSITIONFROMLINE (line)`, and the end is `SCI_GETLINEENDPOSITION (line)`, and `line` is the same for both calls.

PR #8680, opened 2020-08-07
Make CURRENT_LINE and CURRENT_COLUMN expansions one-based

### LINKED ISSUE #8679: CURRENT_LINE and CURRENT_COLUMN should be one-based not zero-based
### Description of the Issue

To be user-useful, the `$(CURRENT_LINE)` and `$(CURRENT_COLUMN)` expansions offered in the _Run_ menu's _Run..._ > _The Program to Run_ box should be "one-based" and not "zero-based".

1. In a Notepad++ tab, move your caret to line 1, column 1 (start of document)
2. Invoke the _Run_ menu's _Run..._ command.
3. Put this text in the input combobox underneath _The Program to Run_ :  
`cmd /c echo CURRENT_LINE=$(CURRENT_LINE) CURRENT_COLUMN=$(CURRENT_COLUMN) && pause`:

image

4. Press the `Run` button.

### Expected Behavior

A CMD window opens with `CURRENT_LINE=1 CURRENT_COLUMN=1` in it.

### Actual Behavior

A CMD window opens with `CURRENT_LINE=0 CURRENT_COLUMN=0` in it:

image

The line and column data is always "one less" than it should be, for any line and any column the caret is on.

### Additional

1. If it is thought that users are using these commands as they are, the "fix" could be to implement new expansion variables, suggest: `$(CURRENT_LINE1)` and `$(CURRENT_COLUMN1)`

2. See also https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8629#issuecomment-667582029

### Debug Information
Notepad++ v7.8.9   (32-bit)
Path : C:\\.............\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252

sasumner:
Fixes #8679

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -824,12 +824,12 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_GETCURRENTLINE:
 		{
-			return _pEditView->getCurrentLineNumber();
+			return _pEditView->getCurrentLineNumber() + 1;	// one-based, not zero-based
 		}
 
 		case NPPM_GETCURRENTCOLUMN:
 		{
-			return _pEditView->getCurrentColumnNumber();
+			return _pEditView->getCurrentColumnNumber() + 1;	// one-based, not zero-based
 		}
 
 		case NPPM_GETCURRENTSCINTILLA:


dail8859:
Couldn't this potentially break plugins that are expecting it to be zero based? Not saying it is correct, just that the behavior will change.

sasumner:
>Couldn't this potentially break plugins that are expecting it to be zero based?

It could.  Perhaps plugins was its primary use?
I can't see this at a user-level at all, as all other tools I've seen that deal with line and column numbers deal with "user" concepts of one-based numbers.

Should I back off and create `$(CURRENT_LINE1)` and `$(CURRENT_COLUMN1)` ?

dail8859:
I guess that would be Don's call. That suggestion would ensure compatibility and would be my preference.

sasumner:
>I guess that would be Don's call. That suggestion would ensure compatibility and would be my preference.

Let's just go with that, then.
No need to wait for Don on it.
I'll rework it...

Yaron10:
Do plugin use `$(CURRENT_LINE)` or `NPPM_GETCURRENTLINE`?

If the latter is correct, how about changing `wsprintf(expandedStr, TEXT("%d"), lineNumber);` to `wsprintf(expandedStr, TEXT("%d"), lineNumber + 1);` instead of the current change?

(In `RunDlg/RunDlg.cpp`).

Issue #8624, opened 2020-07-27
Find Results: Add a word-wrap option


Yaron10:
Add an option to word-wrap (toggle state in the context menu).

Thank you.

donho:
Not sure it's nice visually. I'll check it anyway.

sasumner:
@donho said

>Not sure it's nice visually. 

I think it is nice visually.
See a sample screenshot [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8525#issuecomment-654244664)
The screenshot also has visible whitespace turned on.

Yaron10:
Seems nice to me too.

sasumner:
Perhaps it could be a _Preference_ setting--for how you usually want it--and then a _Find result_ window right-click context menu option as a temporary (until you restart N++) override?

Yaron10:
You did a great job with the new Searching page and you want to fill it up now. :)

On a serious note: your proposed structure can be helpful in some cases to some users.
But, all in all, allow me to disagree: I think that one permanent context-menu setting would be more convenient and less confusing.

Thank you for working on that too.

sasumner:
>You did a great job with the new Searching page and you want to fill it up now.

Thanks.
Yes, I have "big plans" for it.
It seems like search/replace/etc. is a big area for me, where a lot of the work I've done has been concentrated.

>one permanent context-menu setting

Perhaps you are right.
And it would leave more room in the _Preferences_ 's _Searching_ page for other things!!

Yaron10:
> It seems like search/replace/etc. is a big area for me, where a lot of the work I've done has been concentrated.

I'm glad it is. Thanks again for your work.
I've also made a lot of changes in this area for my personal use.

The Incremental Search Bar is my main search tool.
![תמונה](https://user-images.githubusercontent.com/10229320/89132236-701cf280-d51b-11ea-81da-8eb7c8010115.png)

I've replaced the EditBox with a ComboBox.
Find First/Prev/Next.
Display All in Current Document (Ctrl -> in All Open Documents).
Whole-Word. 
Case.
Mark.

I've also made the default size of the Find dialog much smaller. 
![תמונה](https://user-images.githubusercontent.com/10229320/89132356-3d272e80-d51c-11ea-8aa6-a16e74e91671.png)

Double-clicking the title toggles "More/Less".
[...]

Issue #8525, opened 2020-07-04
'Find in Files': Not possible to jump to every single hit in certain cases


tstone-1:
### Description of the Issue

When using 'Find in Files', Notepad++ will find all occurrences of the string of interest. However, it is not possible to open the respective files from the list of hits, jumping to the particular occurrence.

### Steps to Reproduce the Issue

1. Within a folder `mwe`, create a file `mwe.tex` containing
```
Aryl-Aryl

Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Aryl-Aryl.

[...]

Issue #8270, opened 2020-05-15
Searching using extended mode doesn't handle \0 properly


aussiechip:
### Description of the Issue

In a binary file that contains a NUL, searching for \0 on its own works properly, and searches with anything before a \0 seems to work, but searching for \0 followed by anything else fails. It generally finds a \0 followed by some other character. 

### Steps to Reproduce the Issue

1. Open a file with lots of NUL characters in it. Perhaps you could create one that has \0 followed by every other ASCII character, as a test?
2. Look for a pattern that involves a \0 followed by another character.
3. If it finds the right one, try searching for a \0 followed by some different character

### Expected Behavior

I expected it to find the string I was searching for! (I happened to be looking for \0A\0P)

### Actual Behavior

It found something else

### Debug Information

Notepad++ v7.8.6   (64-bit)
Build time : Apr 21 2020 - 15:23:03
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1903
OS Build : 18362.418
Plugins : JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
> looking for \0A\0P

A potential workaround while this is being looked into is to do the search in _Regular expression_ mode.  That mode does not really like dealing with NUL characters either (and of course Notepad++ as a whole doesn't like them--and probably never will), but if your files use the most basic of encodings, you might have luck searching for `[^\x01-\xff]A[^\x01-\xff]P`.  A bit painful, but it seems to work.

sasumner:
Actually a simpler regex search of `\x00P` should work as well.
Apparently I was thinking that regular expression searches had trouble with NUL characters as well.

aussiechip:
OK. I had kindof assumed that that was what was happening behind the scenes anyway - the extended mode string would be internally expanded to a regexp, and then passed to the search engine. It was a pretty rash assumption, I'l admit.

As I was testing this, it became pretty clear that the search was locating random characters that happened to be in the search string memory before it was copied out of the dialog box's field. The copy of the search string from the dialog presumably terminates with the first NUL it finds, and then the search string length is passed separately, or something like that.

Anyway, that might give a bug hunter a starting point.

aussiechip:
oops. Sorry.

sasumner:
[...]

Issue #7723, opened 2019-12-11
DOUBLE (edit: was REMOVE) the line-length limit in Find-result window


sasumner:
### Description of the Issue

There has been a complaint about long lines not being "well-represented" in the `Find result` window.  See https://notepad-plus-plus.org/community/topic/18028/how-do-i-see-the-full-line-in-find-results-window.

Long lines are truncated and an ellipsis appended if the line is too long.  If the "hit" for the line occurs in the truncated part, the hit marking is either not present, or is sometimes put in a questionable location.

The line-length limit is represented by `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` which resolves to 1024.  This #define comes from a Scintilla #include but there doesn't seem to be any rationale for it; perhaps it is a vestige from the early days?

Removing the line-length limit would solve the problem, but would it create any new problems?

### Steps to Reproduce the Issue

See the Community issue linked earlier.

### Expected Behavior

Hits beyond column (roughly) 1024 should have their matched hit-text fully visible in the `Find-result` window.

### Actual Behavior

Lines with length greater than (roughly) 1024 are truncated and a `...` is appended to the end.  Thus, the true hit text is not knowable from simply looking at the line in `Find-result`.  Note that if one double-clicks the line in `Find-result`, the true hit-text will be revealed because the user's caret selection will indicate it in an editor tab.

### Debug Information
Notepad++ v7.8   (32-bit)
Build time : Oct  7 2019 - 00:58:16
Path : C:\........\npp.7.8.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1709
OS Build : 16299.1508
Plugins : ColumnTools.dll Explorer.dll LocationNavigate.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll PythonScript.dll

sasumner:
Just making some notes here:

There's some code in a `case ProcessFindAll` block that looks like this:

```
TCHAR lineBuf[1024];
if (nbChar > 1024 - 3)
	lend = lstart + 1020;
```

This looks suspiciously like it is trying to enforce a length limit, but instead of using the #define for `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` it is using the magic number of 1024.

Also the `- 3` part is also reminding me of the length of "`...`"

[...]

PR #8681, opened 2020-08-07
Add word-wrap option to find-result window context menu

sasumner:
Implements #8624


sasumner:
Closing for now as this PR was to build artifacts for @Yaron10 to experiment with.

Yaron10:
@sasumner,

:+1: 
Thank you for a great job!
With really long lines this option is a MUST. 

I couldn't find any faults. :)
The only possible enhancement I can think of is colorizing `Line x`. It's slightly less prominent with `Wrap` ON.

תמונה

But it's probably not worth it.

***
I've tested it with [this](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5044169/extensions.zip) file.
I searched for `id`. The results baffled me and I had to test it with a fresh portable NPP 7.8.9.

תמונה

I don't want to open a new issue if it's already known.

sasumner:
>I've tested it with this file....I don't want to open a new issue if it's already known.

I believe that is a manifestation of the same thing as in #8525
I'm still looking at that one, but it appears the search-result lexer gets confused because there are multi-byte characters in the data.

sasumner:
>only possible enhancement I can think of is colorizing Line x

Well, in theory you can colorize/bold/change-font-size yourself using the _Style Configurator_ settings for `Search result`.
However, because this data screws up the lexer (again, # 8525 ), the results of doing so doesn't look so good:

image

The line-wrapping looks super, though!!
:-)

sasumner:
@Yaron10 
Did you try wrapping with a _Find in these found results..._ search?

Yaron10:
> The line-wrapping looks super, though!!

That's correct. :)

> Did you try wrapping with a Find in these found results... search?

I did and It looked as great. Any issue there?

> Well, in theory you can colorize/bold/change-font-size yourself using the Style Configurator...

:+1: 

***
> I believe that is a manifestation of the same thing as in 8525

I didn't follow all the steps there. In our case wrong characters are highlighted in the Find Results window. 
No need to open a new issue?  

Here's another case:
Paste the following line in a new document.
```
user_pref("browser.uiCustomization.state", "{\"placements\":{\"widget-overflow-fixed-list\":[],\"nav-bar\":[\"back-button\",\"forward-button\",\"stop-reload-button\",\"_6e1ccc3a-6961-4b99-9455-e04cb868e7c2_-browser-action\",\"_d07ccf11-c0cd-4938-a265-2a4d6ad01236_-browser-action\",\"customizableui-special-separator1\",\"_8883111b-4bc8-4a99-b37f-5672f6cf9834_-browser-action\",\"_4853d046-c5a3-436b-bc36-220fd935ee1d_-browser-action\",\"customizableui-special-separator2\",\"personal-bookmarks\",\"urlbar-container\",\"downloads-button\",\"ublock0_raymondhill_net-browser-action\",\"find-button\",\"_8883111b-4bc8-4a99-b37f-5672f6cf9825_-browser-action\",\"_7773111b-4bc8-4a99-b37f-5672f6cf9825_-browser-action\",\"customizableui-special-separator3\",\"privatebrowsing-button\",\"new-window-button\",\"customizableui-special-separator4\"],\"toolbar-menubar\":[\"menubar-items\"],\"TabsToolbar\":[\"tabbrowser-tabs\",\"new-tab-button\",\"alltabs-button\"],\"PersonalToolbar\":[]},\"seen\":[\"developer-button\",\"_6e1ccc3a-6961-4b99-9455-e04cb868e7c2_-browser-action\",\"_8883111b-4bc8-4a99-b37f-5672f6cf9825_-browser-action\",\"_8883111b-4bc8-4a99-b37f-5672f6cf9834_-browser-action\",\"ublock0_raymondhill_net-browser-action\",\"_4853d046-c5a3-436b-bc36-220fd935ee1d_-browser-action\",\"_7773111b-4bc8-4a99-b37f-5672f6cf9825_-browser-action\",\"_d07ccf11-c0cd-4938-a265-2a4d6ad01236_-browser-action\"],\"dirtyAreaCache\":[\"PersonalToolbar\",\"nav-bar\",\"TabsToolbar\",\"toolbar-menubar\"],\"currentVersion\":16,\"newElementCount\":134}");
```

Search for `browser`.

-- 16 hits. 
-- The search-term `browser` does not appear in the last 7 results.

I think that in such cases NPP should truncate the part *before* the search-term, add `...` and show the search-term.

Known?

sasumner:
>Did you try wrapping with a Find in these found results... search?
>Any issue there?
No issue, I was just wondering if you had a problem with the behavior.
Opening a secondary finder it will use the "wrap" setting from the original finder.
If you open a third finder from the second, it will still use the setting from the original finder, even if you've changed the setting in the second.
It's okay with me. :-)

sasumner:
>...a manifestation of the same thing as in 8525
>No need to open a new issue?

No, not unless something new comes to light.  8525 covers it.

sasumner:
>I think that in such cases NPP should truncate the part before the search-term, add ... and show the search-term.

Wow, with the addition of the "wrap" stuff, this becomes more of a "problem". :-)
Before, if you didn't see the search-term highlighted, you just assumed it was "off to the right somewheres".
With "wrap" on, you fully expect to see it, somewhere...
To implement your suggestion would required a change to the Scintilla part of the code, as it does the lexing for `Search result`.  As I don't fully understand how this lexing works (or I would have already fixed 8525), I don't see changes to this in the future, at least made by me. :-)

>Known?

Well, it is known that long lines (> ~1024) are purposefully truncated in `Find result` window, with perhaps no real reason for it.  See #7723

Yaron10:
> Opening a secondary finder it will use the "wrap" setting from the original finder.
If you open a third finder from the second, it will still use the setting from the original finder, even if you've changed the setting in the second.
...
No, not unless something new comes to light. 8525 covers it.
...
Well, it is known that long lines (> ~1024) are purposefully truncated in Find result window, with perhaps no real reason for it. See 7723

:+1: 

> Wow, with the addition of the "wrap" stuff, this becomes more of a "problem"

Alas, that's true. :)

sasumner:
Reopening this pull request after user acceptance testing and the following nice feedback:

image

Sample 1:  Before wrapping of long lines is turned on:

image

Sample 2:  AFTER wrapping of long lines is turned on:

image

donho:
@sasumner @Yaron10 
Have you ever tried (by applying this PR) searching in a .exe ? There might be some unwanted results.

sasumner:
@donho said:

>Have you ever tried (by applying this PR) searching in a .exe ? 

We advise users on the support forums not to use a text editor for searching non-text (e.g. binary) files.  One reason for this is that NUL ('\0') support in N++ is not complete (yet!).  Searching an .exe is going to cause NULs to appear in the _Find result_ window and this causes several known issues (one is #8270 , there are others, including one I can't find right now regarding copying from the _Find result_ window stops at any '\0' character in a line).

That being said, if you did search an .exe, and the longness of the result bothered you, would you just not right-click and untick "Word wrap long lines"?

I tried some searches on the N++ exe with this feature enabled and did not notice any problems.

>There might be some unwanted results.

What are some other "unwanted results" you are thinking of?

donho:
> What are some other "unwanted results" you are thinking of?

Finally, nothing. :) 
I worried about somethings will never happen.

It looks good this PR, just use "!" instead of "not".

sasumner:
@donho 

>Use the modern operators notation (Use ! instead of not).

OK, will change.
Note that this PR was submitted BEFORE our discussion yesterday in another PR, about preferring usage of `!` instead of `not`.
BTW, isn't `not` more "modern" than `!` ?
BTW, I think you don't like `not` because it is Englishy, and `!` has no language ties. :-)
It's fine, you can like whatever you like.

donho:
@sasumner 
> BTW, isn't not more "modern" than ! ?

Interesting... I thought `not` is like `and` and `or`, which are part of c++98.
If you have any link about this info, please provide it to me.

sasumner:
@donho I did some more research and it appears that `!` actually is more modern than `not`.  So I was wrong.  Perhaps I am thinking about Python where I think it is the other way...or maybe I'm wrong there too.
Anyway, I will use your preference of `!` in Notepad++ code from now on.  (Although a quick search shows a lot of `not`)

Yaron10:
@sasumner,

**STR:**
Search for a relatively long term.
Collapse the Search line.

**Result:**
תמונה

Originally, the underline serves as a separator between segments (search term, file name etc.). 
With "Wrap" enabled, the multiple underlines are slightly confusing.

It's relevant to a wrapped file name as well.
תמונה

I'm not sure it's worth the extra work/code. What do you think?

sasumner:
The underline is an effect from the fold being collapsed, with the first line of the folded "code" being really long.

>I'm not sure it's worth the extra work/code. 

There's no work/code that could be done here --  this is Scintilla at its finest. :-)

>the multiple underlines are slightly confusing

I don't think it is a bad effect.  It's not _ideal_, but...

I say we forget about it and move on. :-)

Yaron10:
Moving on. :)

Thank you.

Issue #8634, opened 2020-07-29
A web link is not processed if it contains Cyrillic


andrecool-68:
A web link is not processed if it contains Cyrillic.

Notepad++ v7.8.6   (32-bit)
Build time : Apr 21 2020 - 15:17:06
Path : I:\Tools_Servis\TextCode\NPP++\npp.7.8.6.bin\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Home Premium (64-bit) 
OS Build : 7601.0
Plugins : ComparePlugin.dll DSpellCheck.dll Explorer.dll HexEditor.dll HTMLTag_unicode.dll ImgTag.dll JSMinNPP.dll LocationNavigate.dll mimeTools.dll MultiClipboard.dll NativeLang.dll NppConverter.dll NppExport.dll NppMarkdownPanel.dll NppSnippets.dll PreviewHTML.dll ShtirlitzNppPlugin.dll Tidy2.dll VisualStudioLineCopy.dll WebEdit.dll WindowManager.dll XMLTools.dll _CustomizeToolbar.dll 

![1](https://user-images.githubusercontent.com/3471892/88802224-70cc2680-d1b3-11ea-9cf5-7c571c7c4d33.png)

sasumner:
URL processing has been undergoing change in recent versions of N++.
Can you retest in 7.8.9 and see if it is the same as you described above in that version?

andrecool-68:
@sasumner 
The latest version has exactly the same behavior

![Безымянный](https://user-images.githubusercontent.com/3471892/88806417-0f0ebb00-d1b9-11ea-8bc4-6f53e7a43aee.png)

Yaron10:
NPP uses the following RegEx which needs to be changed:
`URL_REG_EXPR "[A-Za-z]+://[A-Za-z0-9_\\-\\+~.:?&@=/%#,;\\{\\}\\(\\)\\[\\]\\|\\*\\!\\\\]+"`

BTW, a `.` at the end is interpreted as part of the URL which, obviously, is wrong.

This is a task for @guy038 the RegEx master. :)

sasumner:
So, not any kind of new problem relating to the change to URL displaying, but rather a long-standing thing...

andrecool-68:
@sasumner  I noticed this error only today

Yaron10:
> So, not any kind of new problem relating to the change to URL displaying, but rather a long-standing thing

[...]

PR #8685, opened 2020-08-09
Fix URLs not underlined immediately after copy/paste

Uhf7:
This PR fixes an effect brought up here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8634#issuecomment-669503226. It can be seen as another step of the effort to put standard indicators into operation.

The effect can be reproduced as follows:
- New file
- URL underlining on
- Type four lines (the 4th line is empty):
				
		http://aaa
		http://bbb
		http://ccc

- Select them, copy them, paste them several times and watch.

The effect is, that the URLs are not underlined immediately after pasting. Only a cursor movement after the paste gets them underlined.

The reason for it is, that Notepad++ has no reaction to the `SC_MOD_CHANGEINDICATOR` notification implemented. So what happens is:
- The paste occurs
- After the paste, `addHotSpot` is called correctly
- `addHotSpot` marks the links using the indicator
- Scintilla notifies Notepad++ that indicators have been modified
Notepad++ should call `InvalidateRect(EditWindow)` now, what I suggest in this PR.

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -88,6 +88,10 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				prevWasEdit = false;
 			}
 
+			if (notification->modificationType & SC_MOD_CHANGEINDICATOR)
+			{
+				::InvalidateRect(notifyView->getHSelf(), NULL, FALSE);
+			}
 			break;
 		}
 


PowerEditor/src/NppNotification.cpp
```diff
@@ -88,6 +88,10 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				prevWasEdit = false;
 			}
 
+			if (notification->modificationType & SC_MOD_CHANGEINDICATOR)
+			{
+				::InvalidateRect(notifyView->getHSelf(), NULL, FALSE);
+			}
```
donho, line 94: Hmm... I'm wondering if it will bring a performance issue.

donho: It seems ok for me.

Uhf7: It does not bring a performance issue, because calls to `InvalidateRect` do not immediately trigger `WM_PAINT` messages. So calling `InvalidateRect` repeatedly in a loop does no harm.

I checked it out by inserting `OutputDebugString` calls into the code before the `InvalidateRect` calls and the `WM_PAINT` messages sent to Scintilla (in `ScintillaEditView.cpp`). There is one additional `WM_PAINT` message sent per `addHotSpot` call.

But `addHotSpot` calls are seldom, because they are only executed after view range changes or text modifications. I can imagine that `addHotSpot` will be optimized soon, so that it causes no indicator change notifications anymore as long as no URLs have been modified in the view. But this will not affect performance, it will only look better in the debug logs.

Another source of indicator change notifications is the `xmlMatchedTagsHighlighter`. This one does its work after each text cursor move, and also causes one more `WM_PAINT` message per text cursor move. It could be optimized too. And this would also have a close-to-none effect on performance.

What struck me while this tests was, that most `WM_PAINT` messages sent to Scintilla have as update area only _the current line_ or _the current line together with the line above_ or _the current line together with the line below_. This `WM_PAINT` messages occur, when the text cursor is moved and when the text cursor is blinking. And exactly this `WM_PAINT` messages hide most of the potential update problems we may have. I think, this is why the problem has not been seen before. So, whenever it comes to test update problems, I would set the blink rate of the text cursor to max first.

donho: @Uhf7 Thank you very much for the testing and very detail explanation!

PR #8690, opened 2020-08-11
Separate command shortcut names and menu names.

### LINKED ISSUE #8689: Different shortcut names in Shortcut Mapper table and menu
This is about the Shortcut Mapper, tab 1, _Main Menu_.

The shortcut names displayed in the first column are mostly derived from the names of the according menu entries. This is good for most of them, but some commands call for different names in the menu and in the shortcut table. Example:
Menu -> View -> Collapse Level -> 1
The menu entry is "1", and so is the name in the Shortcut Mapper. Not exactly intuitive.

All localized names in this page of the Shortcut Mapper come from the Menu/Main/Commands element of the Language.xml. This contains sub-elements, which have the attributes **id** and **name**. **id** contains the command number, **name** contains how to show it in the menu.

My proposal here is, to add another, optional attribute (I named it **shortcut**). It has not to be there at all, but if it is there, then it is used as display name in the Shortcut Mapper. If it is not there, then the attribute **name** is used for the Shortcut Mapper like before. Useful things can be done with it:

                    <Item id="44051" name="1" shortcut="Collapse Level 1"/>
                    <Item id="44052" name="2" shortcut="Collapse Level 2"/>
                    <Item id="44053" name="3" shortcut="Collapse Level 3"/>

or:

                    <Item id="44081" name="Project Panel 1" shortcut="Toggle Project Panel 1"/>
                    <Item id="44082" name="Project Panel 2" shortcut="Toggle Project Panel 2"/>
                    <Item id="44083" name="Project Panel 3" shortcut="Toggle Project Panel 3"/>
                    <Item id="44085" name="Folder as Workspace" shortcut="Toggle Folder as Workspace/>

In all examples, we want to keep the short name in menu, but need something more in the Shortcut Mapper. The names in the Shortcut Mapper could be updated step by step, there is no rush to change all language files at once. And, in the native English version, which works without language file, the Shortcut Mapper names come from the `winKeyDefs []` table in `Parameters.cpp`, and the menu names come from the menu resource in `notepad_plus.rc`.

This is also a preparation for some new commands, which I want to put into the command list of the Shortcut Mapper, but which I do not want to have in the menu:

                    <Item id="44xxx" name="Project Panel 1" shortcut="Launch Project Panel 1"/>
                    <Item id="44xxx" name="Project Panel 2" shortcut="Launch Project Panel 2"/>
                    <Item id="44xxx" name="Project Panel 3" shortcut="Launch Project Panel 3"/>
                    <Item id="44xxx" name="Folder as Workspace" shortcut="Launch Folder as Workspace"/>

For all of those commands it seems important, that they are labeled correctly in the Shortcut Mapper.

Uhf7:
Fixes #8689.

It's actually an improvement, I wrote fix to link it to the issue.


sasumner:
>It's actually an improvement, I wrote fix to link it to the issue.

When I do an "enhancement", I write "Implements #xxxx" as the description.
It's just what I do, certainly not some rule.

Uhf7:
@sasumner, yes, this sounds right. But in the moment, it does not react to "Improves" or "Enhances" or "Implements", as far as I know. That's why I wrote "Fixes", only to link it to the issue. OK, it's kind of cheating too. I don't do it again. I would not have done this, if the issue would have been created by someone else. But I created it mainly for this PR, and I felt, if the PR is gone, accepted or rejected, the issue can be closed too.

sasumner:
@Uhf7 Not sure I understood that.  Not cheating, IMO.  You can certainly implement issues you create.

Uhf7:
I cheated the Github portal software by tricking it into linking that PR to the issue.

donho:
A question (from a non-native English speaker) does nothing about cheating:
Why you guys use `Fixes` and `Closes` instead of `Fix` and `Close` ?

sasumner:
@donho said:

>A question...

It is a shorter way of saying "This PR fixes issue XX".

If you tried it as "This PR fix XX", it just isn't right. :-)
Maybe that one still seems OK to you? :-)

It doesn't really matter, the intent is clear.
Just keep doing it the way you do it; we'll keep doing it the way we do it.

BTW, Notepad++ is amazing in an English sense.
Thank you for making it as a default English project.
Maybe doing anything else did not make sense?
It has a few rough spots.
The worst one may be "Find in these found results...". :-)

donho:
@sasumner 

> If you tried it as "This PR fix issue XX", it just isn't right. :-)

I didn't think in this way - thank you for the answer.

> The worst one may be "Find in these found results..."

As you know, English is not my native language, so feel free to do a PR if it hurts your eyes :)

sasumner:
@donho 

>As you know, English is not my native language, so feel free to do a PR

When the English is chosen for a new command, I don't like to see it changed in the future, because the translators have already translated based on the original.  Does this make sense?  Or should we feel free to make these sort of changes later?

And of course I have already changed it in my personal english_customizable.xml file. :-)

Uhf7:
> When the English is chosen for a new command, I don't like to see it changed in the future

I would contradict here. Example: There is the _View_ menu, containing the entries
- Fold All
- Unfold All
- Collapse Current Level
- Uncollapse Current Level

and so on. This is an example of inconsistency. And we have to keep it for ever now? For what?

If we would change "Uncollapse" into "Unfold", the English version would get better and all the translations would not become wrong or unusable because of it.

sasumner:
It's not a "should never do" thing, but only when really necessary.
Other points: 
* existing documentation can become out of date
* references to things by textual name on support forums become meaningless or are hard to track

donho:
@Uhf7 
About this PR, I don't think it's a good idea to create the dependency of `nativeLang.xml` because : 
1. The modification of englis.xml and of other localization xml files is necessary. 
2. Without `nativeLang.xml`, the menu items remain the same.
So I won't take this solution.

However, I agree that some item names in shortcut mapper are meaningless. The best solution to me is concatenate items' parent menus name, for example, `Auto-Completion->Function Completion` or `Collaps level->3`. Such method will solve the problem for all languages. But my question is: Can we get the menu parents' name with little coding effort?

Uhf7:
Actually, the nativeLang.xml's can stay as they are or be upgraded step by step. There is no harm done, if they are not updated, then everything stays as it is. If they are updated, then the Shortcut Mapper table improves, everything else stays as it is too.

Without nativeLang.xml (Settings _English_) , the name displayed in the menu comes from the resource, and the name displayed in the Shortcut Mapper comes from the e `winKeyDefs []` table in `parameters.cpp`.  This table can be updated step by step too. I included the three names for _Toggle Project Panel_ as example.

Why we need different names instead of concatenated menu items? Example sub window menu entries: They are named now _Folder As Workspace_, for example, and do actually _Toggle Folder As Workspace_. But this is bad in the menu, because to show this, the menu entry has a check mark displayed before it. And you cannot create a good name for it by concatenating menu entries. And the concatenations will get quite long sometimes. And I think, some translators may want to write different names in the menu and in the Shortcut Mapper table (A short name in menu, a more descriptive name in the Shortcut Mapper).

It was meant as compatible change, to open more potential in the future.

donho:
My main point is the PR doesn't change anything for English users, and I believe they are the major users of Notepad++.

Uhf7:
@donho , maybe there is a misunderstanding here. Please help me with this.

You said, that there is no change without `nativeLang.xml`. Do you mean settings _English_?

If yes, then we have to consider, that setting it to _English_ from any other Language, then Notepad++ works without `nativeLang.xml` only after restart. Because `_nativeLangSpeaker.init` is called with `loadIfEnglish = false` only at the start of Notepad++ in the constructor of `Notepad_plus`. When changing languages, it is called with `loadIfEnglish = true`, which makes it use the `English.xml` for the _English_ setting until restart.

What changed for English users without `nativeLang.xml` with this PR were three entries in the Shortcut Mapper:
- _Project Panel 1_ became _Toggle Project Panel 1_
- _Project Panel 2_ became _Toggle Project Panel 2_
- _Project Panel 3_ became _Toggle Project Panel 3_

This were exactly the changes I expected right now for this case. This changes were reached by inserting names for `IDM_VIEW_PROJECT_PANEL_<n>` into the he `winKeyDefs []` table, not by changing any language.xml. I did nothing insert into this table for the `IDM_VIEW_FOLD_<n>` commands and the `IDM_VIEW_UNFOLD_<n>` commands, but this PR would have opened the possibility to do that or could be extended to this. I just don't wanted to change everything at once.

Can you please check, if you see the _Toggle Project Panel 3_ entries after starting Notepad++ in _English_ language? Perhaps I have coded something wrong, so that it does not work on your system. Thank you in advance for this.

donho:
@Uhf7

Uhf7:
> Why don't you rename for all in this PR ?

I just didn't want to change everything at once. But if you say yes to it, I revise this PR and rename the other entries with unclear names too. And I read the localized names from the `<ShortcutMapper>` section instead of from the `<CommandSection>`.

Then we would have a great new Shortcut Mapper main table (new names at least), ready for commands like "Launch Folder As Workspace" or "Launch Project Panel n". Why will I go for this additional commands? They are not needed in the menu, but they can be used to switch to the specified panel using a keyboard shortcut.

What is my plan after that? Close a single tab of the docked window instead of the whole stack. The prerequisite for this was to make the Project Panels switchable. But this is still a few PRs ahead.

donho:
> I just didn't want to change everything at once. But if you say yes to it, I revise this PR and rename the other entries with unclear names too. 

Surely it was my bad to close this PR too soon, since it does exactly what I want (good job).
But I close it for not seeing what I expected, so I think it's worth to mention the procedure of validation/merging PR (generally):
1. Checking the bug description, and try to reproduce it.
2. If bug is reducible, get the PR branch, compiling it and see if it fix the bug.
3. If the bug has been fixed in PR, then review the codes.

> What is my plan after that? Close a single tab of the docked window instead of the whole stack. The prerequisite for this was to make the Project Panels switchable. But this is still a few PRs ahead.

Yes, sounds great.

sasumner:
@donho said

>2. If bug is reducible, get the PR branch, compiling it and see if it fix the bug.

That sounds like more effort than:  "If bug reproduces, get the PR artifact exe and see if it fixes the bug."

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -168,10 +168,10 @@ generic_string Shortcut::toString() const
 	return sc;
 }
 
-void Shortcut::setName(const TCHAR * name)
+void Shortcut::setName(const TCHAR * menuName, const TCHAR * shortcutName)
 {
-	lstrcpyn(_menuName, name, nameLenMax);
-	lstrcpyn(_name, name, nameLenMax);
+	lstrcpyn(_menuName, menuName, nameLenMax);
+	TCHAR const * name = shortcutName ? shortcutName : menuName;
```
donho, line 174: Shouldn't it be `const TCHAR *` instead of `TCHAR const *` ?

sasumner: This may be an English speaker thing, but...
If you have, `TCHAR const * foo`, you read it (RTL) out loud as "foo is a pointer to constant TCHAR".
Putting the `const` in a different place, it doesn't read so nice.
THAT is why I like it this way.
But, does it really matter? :-)
That leads into things like, is this correct: `TCHAR *foo` or is this correct: `TCHAR* foo`... or should the `*` be jammed up against the variable name, the type, or floating in space (like line 174 has it above)?
But...these are "small" things, and I'm sure throughout the N++ code we have a variety of all of them. :-)

Uhf7: Both should have the same effect, see

https://stackoverflow.com/questions/890535/what-is-the-difference-between-char-const-and-const-char

From this:
> The first thing to the left of the "const" is what's constant. If "const" is the thing the farthest to the left, then the first thing to the right of it is what's constant

> To avoid confusion, always append the const qualifier.
int       *      mutable_pointer_to_mutable_int;
int const *      mutable_pointer_to_constant_int;
int       *const constant_pointer_to_mutable_int;
int const *const constant_pointer_to_constant_int;

The **const** itself is necessary, because the parameter is **const**. This **const** prevents me from writing `menuName[3]='7'` wihin this function. The **const** has to go over to **name** to satisfy the very accurate compiler, so I appended **const** out of habit. In most of the code lines of Notepad++ **const** seems to be first in the line, and I may have obeyed this in the past, so if you want, I swap it.

Uhf7: >  is this correct: `TCHAR *foo` or is this correct: `TCHAR* foo`... or

I admit that I act totally random about spaces around the `*`. I use the way it is written at the nearest point in code I see, and this not in Notepad++ only. Although at 1st glance, the 1st version seems more correct because of `THCAR *foo, foo1, *foo2;`

sasumner: >I admit that I act totally random about spaces around the *

Yes, use what the code around the area you are affecting does.
However, I've noticed that Visual Studio has its own opinion on this, and sometimes changes it on you.

Uhf7: > Visual Studio has its own opinion on this, and sometimes changes it on you.

Does it? I never noticed. When?

donho: @sasumner @Uhf7 OK, thank you for the explanation. I didn't know both are the same semantically.
Let's keep it as it is.

PR #8706, opened 2020-08-13
Fixed naming error, where "new 1" is followed up by "new 3"

### LINKED ISSUE #8677: NPP jumps from "New 1" to "New 3" if "Find Results" is displayed
**STR:**
Open NPP (for simplifying the case: no session, `New 1` is open).
Open the Find dialog and press `Find All in Current Document`.
-- The `Find Results` window is displayed.
Open a new file.

**Result:**
`New 3` is opened instead of `New 2`.

(The same happens if you open Python Script console).

**Apparently a minor issue, but https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8461 might be related.** 

***
Notepad++ v7.8.9   (32-bit)
Build time : Aug  1 2020 - 16:33:01
Path : C:\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.23964
Current ANSI codepage : 1255

Esco441-91:
Fixes #8677 

**The problem**
As described in the issue 8677, second unnamed document was named "new 3" instead of "new 2" when "Find Results" window was open.

**Reason**
I found out that Find Results window is an unnamed document (enumerated with state DOC_UNNAMED) to help writing lines to it, using same code that is used for writing any document. The document is then treated in the code as any other unnamed document whenever it is listed while trying to find the next number for a new document.

**Solution**
While this maybe could be fixed by adding new enumeration to the document types, the document types are used in so many places that I don't think I could make such a massive change. My solution was to add a new member, _isUtilityBuffer, to document buffers to indicate that it's not a normal document, but rather used for something else. So when Find Results -window is opened, it is marked as an utility buffer and will be ignored whenever deciding a name for new unnamed window.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -1103,7 +1103,7 @@ size_t FileManager::nextUntitledNewNumber() const
 	for (size_t i = 0; i < _buffers.size(); i++)
 	{
 		Buffer *buf = _buffers.at(i);
-		if (buf->isUntitled())
+		if (buf->isUntitled() && buf->isUtilityBuffer() == false) // Skip utility buffers (Find Results)
 		{
 			// if untitled document is invisible, then don't put its number into array (so its number is available to be used)
 			if ((buf->_referees[0])->isVisible())
--- a/PowerEditor/src/ScitillaComponent/Buffer.h
+++ b/PowerEditor/src/ScitillaComponent/Buffer.h
@@ -374,6 +374,9 @@ public:
 
 	void langHasBeenSetFromMenu() { _hasLangBeenSetFromMenu = true; };
 
+	void setUtilityBuffer(bool state) { _isUtilityBuffer = state; };
+	bool isUtilityBuffer() { return _isUtilityBuffer; };
+
 private:
 	int indexOfReference(const ScintillaEditView * identifier) const;
 
@@ -432,6 +435,8 @@ private:
 	bool _isMonitoringOn = false;
 
 	bool _hasLangBeenSetFromMenu = false;
+
+	bool _isUtilityBuffer = false; // Marks that DOC_UNNAMED is used for something else than a "real document". i.e. the Find Results -window.
 
 	MapPosition _mapPosition;
 
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2330,6 +2330,7 @@ void FindReplaceDlg::findAllIn(InWhat op)
 		originalFinderProc = SetWindowLongPtr(_pFinder->_scintView.getHSelf(), GWLP_WNDPROC, reinterpret_cast<LONG_PTR>(finderProc));
 
 		_pFinder->setFinderReadOnly(true);
+		_pFinder->_scintView.setUtilityView(true); // Mark the view as utility so it won't be considered as "new x"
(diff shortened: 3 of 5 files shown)


Yaron10:
@Esco441-91,

Congratulations! Thank you for this PR.

It does fix the problem in the following case:

Open NPP (for simplifying the case: no session, `New 1` is open).
**Open a saved file.**
**Close the saved file.**
Open the Find dialog and press `Find All in Current Document`.
-- The `Find Results` window is displayed.
Open a new file.

**Result:**
`New 2`. 

:+1: 

In the following case, however, another **`New 1`** is opened.

Open NPP (for simplifying the case: no session, `New 1` is open).
Open the Find dialog and press `Find All in Current Document`.
-- The `Find Results` window is displayed.
Open a new file.

Another push. :)

Esco441-91:
@Yaron10,

Thank you for the info. You are right on the new issue generated, but it didn't even require Find Results window was opened. Just opening the second document was enough.  I had made the mistake in estimating where specific functions are called. The same function is used whenever npp is opened, and the first "new 1" is opened in the same way as any invisible view/buffer. 

This also resulted that more than just the Find Results window were marked as utility. While I don't see it as a functional problem, it was not intended so, and it's a good thing you were able to bring this up :)

I made some adjustments so that my new setUtilityBuffer isn't called from so deep in the stack when Find Results window is opened. I made a new push, so feel free to test it if you want :)

Yaron10:
@Esco441-91,

Great job. :+1: 

As I've mentioned in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8677, the problem also occurs if you open the Python Script console (and possibly other panels).

This problem still exists.
Another push? :)

Thanks again.

sasumner:
@Esco441-91 The sign of a good code change is that you've analyzed the code thoroughly and found all of the "trouble spots" (yourself, through the analysis) and fixed them.  Relying on someone (@Yaron10 in this case) to point them out one-by-one, and then you make commit-after-commit to fix each one, is not a sign that you've done the proper research and have come up with an encompassing solution.  Thus, confidence in the solution is eroded when the time comes for someone to check your work and make a yes/no decision about accepting your PR.

Esco441-91:
@sasumner,

Thanks for your feedback. Maybe I have taken in more than I thought. The issue felt easy and the research and solution I made felt correct, but I must've failed at some point. It seems I have more to learn in analyzing code I don't fully know yet. For the latest part, I had forgotten about the Python Console while looking at the Find Results -window. 

I understand your concern about the confidence part and I will close the PR as I don't want it to cause you guys any trouble on acception decision. I think I'll still fix the issue in my forked repo just so I can give myself some peace :)

Thank you and @Yaron10 for the feedback, help and understanding for my state of experience.

sasumner:
@Esco441-91 

It wasn't my intention to be "discouraging", just sort of explaining the reality of it. :-)

>The issue felt easy and the research and solution I made felt correct, but...
YES!  I've been in that situation myself SOOOO MANY times!

Esco441-91:
@sasumner,

Don't worry about it :). Your response couldn't have been better for this situation. I would be lying if I said I didn't feel a bit sad about not being able to fix the code in one go, but I feel that this PR taught me just as much about contribution as a succesful PR would have. I know that I can do better next time, whenever and whichever repo that'll happen to be.

Yaron10:
@Esco441-91,

I'd like to thank you again for your work. It is appreciated.

Submitting a PR or even a commit and then realizing it requires some changes happens frequently to the most experienced coders.

Please update if/when you've found a solution.

Best regards.

Issue #8527, opened 2020-07-04
Add option to disable "are you sure" 'feature' for Replace All in All Opened Documents


Jay91150:
Replace All in All Opened Documents is, or rather was, a very versatile tool for some of us, but has been hobbled with this new release. Despite what some (ahem [#8432](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8432)) insinuate, no, people don't delete their active open files and leave themselves with no backups, and then misclick Replace All in All Opened Documents instead of Replace All. (BTW, even if there are people who misclick, the latter would be fixed by giving the Replace All in All Opened Documents function a hotkey like it should have had from the beginning, i.e. Alt+O or Alt+U or something similar. And if you are steadfastly going to refuse to add a hotkey, then don't underline a letter (currently the 'u' in Documents) to imply that there is a hotkey enabled.)

I use this feature in repetition on sets of large files on a routine basis, and I'm far from the only one who does. Not everyone is like [#8432](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8432)'s OP and we should not be forced to take this new functionality which will slow down workflows drastically. Please add a checkbox or setting that will allow us to disable the 'are you sure' prompt, and use this feature quickly and efficiently as it was meant to be used.

### Steps to Reproduce the Issue
1. Open multiple files
2. Ctrl+H
3. Enter values to find and to replace
4. Mouse over to Replace All in All Opened Documents (because you can't seem to add a hotkey for this function) and click it

### Expected Behavior
All values of 'find' are replaced with 'replace' in all opened documents

### Actual Behavior
Extra popup appears asking me if I'm really sure I meant to click on the command I deliberately clicked on.

### Debug Information
Notepad++ v7.8.8   (32-bit)
Build time : Jun 24 2020 - 00:15:36
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2004
OS Build : 19041.329
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
It's not a big deal.  Just confirm the operation.

sasumner:
I still don't think it is a "big deal", but I will reopen this.

There have been a few mentions of it; for example here is another one: 
[...]

PR #8708, opened 2020-08-13
Add confirmation preference for Replace All in All Opened Documents operation

sasumner:
Implements #8527

Here is sample of new UI.
I anticipate maybe some desired changes to it? :-)

**_Edit: Removed image of initial UI to avoid future confusion_**


sasumner:
@Yaron10 Sorry, I didn't implement your RTL or quoting suggestions in this. :-(

donho:
I think the following layout has more advantage:
image
for some reasons:
1. It's prettier :) 
2. More space for the translation.

Here's code:
```
IDD_PREFERENCE_SEARCHINGSETTINGS_BOX DIALOGEX 0, 0, 455, 185
BEGIN
	CONTROL         "Don't fill find field in Find dialog with selected word", IDC_CHECK_STOPFILLINGFINDFIELD,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 10, 350, 10
	CONTROL         "Use Monospaced font in Find dialog (Need to restart Notepad++)", IDC_CHECK_MONOSPACEDFONT_FINDDLG,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 25, 350, 10
	CONTROL         "Find dialog remains open after search that outputs to results window", IDC_CHECK_FINDDLG_ALWAYS_VISIBLE,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 40, 350, 10
	GROUPBOX        "Replace Confirmations", IDC_SEARCH_CONFIRMATIONS_STATIC, 30, 66, 279, 84, BS_CENTER
	CONTROL         "Confirm interactive Replace-All-in-All-Opened-Documents", IDC_CHECK_CONFIRMREPLOPENDOCS,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 80, 265, 10
	CONTROL         "Confirm Replace-All-in-All-Opened-Documents in macro", IDC_CHECK_CONFIRMMACROREPLOPENDOCS,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 95, 263, 10
	CONTROL         "Confirm interactive Replace-in-Files", IDC_CHECK_CONFIRMREPLFILES,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 111, 224, 10
	CONTROL         "Confirm Replace-in-Files in macro", IDC_CHECK_CONFIRMMACROREPLFILES,"Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 126, 196, 10
END
```

BTW, I don't think it's a good idea to make confirm dialog disable optional. I can imagine that the user disable it by ignorance, then he/she hits the Replace in Files button accidentally. It's not like Replace in all Opened files (all the modified files are kept in the editor, so user can undo them file by file.

sasumner:
@donho

It's prettier, but it takes up a lot more of the available space that way. :-(
We have many more "Searching" preferences to come, so where will they fit? :-)

Should I remove both of the _Replace in Files_ items?
I do recall someone asking for it to be removed during macro execution, but I can't find the reference on that right now.
Is it reasonable to remove the interactive one, leaving 3 new checkboxes with this?

donho:
> Should I remove both of the Replace in Files items?

User should not be able to disable the confirmation of Replace in File both button/macro action for me.


OTOH, the confirmation of Replace all in All opened document macro action should be always absence  - if user recorded such macro and run it by error, it's not our problem but PEBKAC.


> I do recall someone asking for it (the confirmation) to be removed during macro execution, but I can't find the reference on that right now.

There are always people who ask somethings. It doesn't mean what they ask is pertinent and we  should implement the feature they ask. But if you see a feature is request more than 3 times, it may need to be implemented.

> Is it reasonable to remove the interactive one, leaving 3 new checkboxes with this?

So we keep only one checkbox "Confirm Replace All in All Opened Documents action". No word "interactive", neither the "-". And we can remove the groupbox as well.

sasumner:
>There are always people who ask somethings. It doesn't mean what they ask is pertinent and we should implement the feature they ask

Right, but when you're implementing something related, it doesn't hurt to also do a related request that you remember.

>So we keep only one checkbox "Confirm Replace All in All Opened Documents action". No word "interactive", neither the "-". And we can remove the groupbox as well.

I agree that this makes sense.  My main reason for agreeing is that it saves space on the page. :-)

I will do it.

donho:
> but when you're implementing something related, it doesn't hurt to also do a related request that you remember.

Hmm... It could hurt. I do believe KISS - too many options kill application. Users prefer a simple, easy to use program.

sasumner:
>too many options kill application

I really do hate preference settings for things, so, yes, total agreement

sasumner:
Revised UI:

image

Issue #8654, opened 2020-08-01
URLs have unexpected colours


ygoe:
### Description of the Issue
In any document, URLs have an unexpected formatting when hovered with the mouse cursor or when the text cursor is inside the URL text span. This formatting does not correspond to what I've set in my styles (#0000ff foreground, white background, underlining).

### Steps to Reproduce the Issue
1. Open or create any document
2. Find or type a URL like this: https://github.com
3. Place the mouse cursor over the URL / place the text cursor inside the URL

### Expected Behavior
In previous versions (up to 7.8.8) the mouse cursor would change the URL to blue text colour and underline it. The background remains unchanged (white). The text cursor did not alter the URL's appearance in any way.

Previous versions didn't seem to care for my style setting either. At least my styles had a grey background and no underlining, which has never been the way how it was displayed anyway. But the default was okay for me.

### Actual Behavior
With both, mouse and text cursor, the URL is now displayed with a blue-ish background, the text colour remains unchanged (black for text, grey for Bash comments, whatever you have) and it's not underlined. The text contrast is a little low depending on the current foreground colour.

Additionally, when moving the text cursor on the URL (the background turns blue) and then moving the mouse cursor that's not over the URL, the blue background goes away. Text and mouse cursor seem to compete on the URL formatting and the later input event wins.

![Screenshot](https://user-images.githubusercontent.com/868315/89096408-d64c2d00-d3d6-11ea-9693-42ef62c86863.png)

### Debug Information
Notepad++ v7.8.9   (32-bit)
Build time : Jul 15 2020 - 20:26:50
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1709
OS Build : 16299.1087
Current ANSI codepage : 1252
Plugins : CustomizeToolbar.dll mimeTools.dll NppConverter.dll NppExport.dll NppSaveAsAdmin.dll

PR #8713, opened 2020-08-15
Bring blue underlined URLs back.

Uhf7:
This PR shows, how to bring the blue underlined URLs back, which seem to be so important to some Notepad++ users.

Last complain about it I read in #8654.

This can only be activated with a new release of Notepad++. This release would have to include a new _SciLexer.dll_, and together with this, the hover style for the URLs can be changed to `INDIC_EXPLORERLINK`.

I filed this modification also as https://sourceforge.net/p/scintilla/feature-requests/1371/, but they say: "Not in this version anymore". So if you want it, you will probably have to modify the Notepad++ version of _SciLexer.dll_. Luckily, there are only four lines to change.


Yaron10:
@Uhf7,

I'm on of those "some Notepad++ users" who prefer the old behavior on hover.
Thank you for this PR. The new implementation is now perfect.
:+1: 

`pView->execute(SCI_INDICSETALPHA, URL_INDIC, 70);` is not needed anymore, correct?

Uhf7:
@Yaron10,

I would wait, until a new Notepad++ specific SciLexer.dll is released, before modifying Notepad++.exe. This modifications should be done synchronous  with the next release, if  @donho goes this way.

And yes, using the new indicator style, the `SCI_INDICALPHA` is no longer necessary. This was to improve visibility of the `FULLBOX` style in dark themes. Which is another thing to think about, may be, some modifications of the themes must be taken back. Sorry for all that trouble, but in the end, it will be better and, hopefully, more future-safe, than before.

Yaron10:
If you check `No Underline`, links *are* underlined on hover.
This is actually not so bad IMO.

Or, did you mean some more changes were needed in NPP code?

Yaron10:
@Uhf7,

Mid-air collusion. :)

Uhf7:
> If you check No Underline, links are underlined on hover.

Yes, you would need two different hover styles too.

`INCDIC_TEXTFORE` (_no underline_) or `INDIC_EXPOLORERLINK` (_underline_) while hover, and `INDIC_HIDDEN` (_no underline_) or `INDIC_PLAIN` (_underline_) while normal.

Yaron10:
:+1: 
Thank you.

donho:
Isn't it easier to reverse to the old implementation?

sasumner:
> Isn't it easier to reverse to the old implementation?

That seems drastic.
It feels like a lot of good work has been put into changing this.

Uhf7:
> Isn't it easier to reverse to the old implementation?

This would have downsides too:
- The syntax-highlighting didn't work around URLs, that's why the new method has been introduced originally.
- The Scintilla doc says, that style byte partitioning is deprecated, so they will not fix syntax highlighting errors caused by it.
- The new implementation is shorter and easier to maintain.

Of course, Scintilla has not much indicator styles right now, but perhaps this can be improved over the time.

Notepad++ seems to have a pioneering role in using this indicators together with hovering, I didn't realize this at the start, but I wouldn't turn back being so close to finishing it.

donho:
After applying this PR and recompiling Scintilla, the behaviour remains the same:
image

Am I missing somethings?

Uhf7:
To put it into action, the hovered URL style must be changed in Notepad++ from `INDIC_FULLBOX` to the new `INDIC_EXPLORERLINK`. 

I didn't do this within this PR, because `INDIC_EXPLORERLINK` is unknown to all _SciLexer.dlls_  without this PR. So it can actually be done together with a new release of Notepad++ only. This release would have to include the new _SciLexer.dll_, and together with this, the indicator styles used for URLs should be changed as follows:
- normal
  - `INDIC_HIDDEN` (no underline) or `INDIC_PLAIN` (underline)
- hovered
  - `INCDIC_TEXTFORE` (no underline) or `INDIC_EXPOLORERLINK` (underline)

The `SCI_INDICSETALPHA , URL_INDIC, 70` call should be removed too.

I can include this modifications into this PR too, but then, it should only be integrated, when a new release of Notepad++ comes out.  Because, after it is integrated, the current developer version of the _Notepad++.exe_ cannot be used together with the signed version of the _SciLexer.dll_ from the last release 7.8.9.

donho:
Indeed, it works after replace `INDIC_FULLBOX` by `INDIC_EXPOLORERLINK` in Notepad++ codes.
However replace `INDIC_TEXTFORE` by `INDIC_EXPOLORERLINK` don't have expected effect (underline). In your modified Scintilla code, I can see `INDIC_EXPOLORERLINK` is new added and doing actually nothing. Could you give me a hint?

Yaron10:
```
{
	ScintillaEditView* pView = view ? view : _pEditView;

	int urlAction = (NppParameters::getInstance()).getNppGUI()._styleURL;
	LPARAM indicStyle = (urlAction == 2) ? INDIC_PLAIN : INDIC_HIDDEN;
	LPARAM indicHoverStyle = (urlAction == 2) ? INDIC_EXPOLORERLINK : INDIC_TEXTFORE;

	LPARAM indicStyleCur = pView->execute(SCI_INDICGETSTYLE, URL_INDIC);
	LPARAM indicHoverStyleCur = pView->execute(SCI_INDICGETHOVERSTYLE, URL_INDIC);

	if ((indicStyleCur != indicStyle) || (indicHoverStyleCur != indicHoverStyle))
	{
		pView->execute(SCI_INDICSETSTYLE, URL_INDIC, indicStyle);
		pView->execute(SCI_INDICSETHOVERSTYLE, URL_INDIC, indicHoverStyle);
		pView->execute(SCI_INDICSETFLAGS, URL_INDIC, SC_INDICFLAG_VALUEFORE);
	}
```

I suppose this is what @Uhf7 meant.

donho:
@Yaron10 Thank you. But I'm saying that: From my understanding, his PR doesn't seem to be able to have any effect.
But I just realized what it does (why it works) by writing to you.  
Thank you anyway.

@Uhf7 OK, I see why and how it work - `INDIC_EXPOLORERLINK` is an indication only for Notepad++ so Scintilla project will never accept integrating it.
OTOH, I prefer the current behaviour (`INDIC_FULLBOX `) so I would say let's see the reaction of users in the future.

Yaron10:
@donho,

:+1: 

>  I prefer the current behaviour `(INDIC_FULLBOX )`...

To each their own. :)

Uhf7:
> Scintilla project will never accept integrating it.

Actually, I wouldn't say this, but there may be a long way until we would have it in the Notepad++ _SciLexer.dll_. 

And if the original _SciLexer.dll_ makers say, we do it  now, then it may come with the next _SciLexer.dll_-Version (may be 5.xxx), and then, ...

I suggested that change, because it is a reasonably small change which will work on several new versions of _SciLexer.dll_ ahead, which means, no matter what they invent or change, the new `INDIC_EXPLORERLINK` style can be added easily to that new _SciLexer.dlls_.

As I see it, you have a lot of heavy modifications within _SciLexer.dll_ already, to transform it to a Notepad++ _SciLexer.dll_, so it would make no major difference, to add some more enhancements, like the `INDIC_EXPLORERLINK` style.

Issue #8674, opened 2020-08-06
Sorting with column block selection uses incorrect key


sasumner:
### Steps to Reproduce the Issue
1. Consider this text in an editor tab:
```
abcdef
abcddef
```
2. Make this column block selection:
![image](https://user-images.githubusercontent.com/30118311/89571291-67ffd400-d7f5-11ea-8828-c80da896f919.png)

3. Execute _Edit_ > _Line Operations_ > _Sort Lines Lexicographically Ascending_.

### Expected Behavior
Since the selected column key contains the same characters for both lines, the lines should remain in the original order after the sort.

### Actual Behavior
The lines incorrectly change order.

### Additionally
A zero-width column-block selection should use as the sort key each line's content from the column of the selection out to the end of line.  Currently, this type of selection will cause the **_single character_** at the caret column to be used as the key.

So, using the same data as above, and this selection:

![image](https://user-images.githubusercontent.com/30118311/89571437-aac1ac00-d7f5-11ea-98cb-42d1e2077d88.png)

will result in no change in line order, when it should swap the lines.

### Debug Information
Notepad++ v7.8.9   (32-bit)
Build time : Jul 15 2020 - 20:26:50
Path : C:\\.............\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll

donho:
Column selection should not be considered while sorting, otherwise there'll be a lot of cases (operations) should be taken into consideration.

sasumner:
@donho said:

>Column selection should not be considered while sorting

I don't understand this.  Column selection currently (N++ 7.8.9) **_IS_** considered while sorting, the current code is just doing it **_WRONGLY_**.

>a lot of cases (operations) should take column and multi-selection into consideration.

I agree that currently these types of selection operation support is lacking in N++, but it doesn't relate to current issue where column selection is already supported.

sasumner:
Changes from this issue released in Notepad++ version 7.9

PR #8717, opened 2020-08-15
Fix column key sort with col block starting in col1 bug

### LINKED ISSUE #8716: Sorting with column block key starting in first column sorts incorrectly

1. Consider this text in an editor tab:
```
abcdef
abcddef
```

2. Make this column block selection:
image

3. Execute *Edit > Line Operations > Sort Lines Lexicographically Ascending*.

### Expected Behavior
Since the selected column-sort-key contains the same characters for both lines, the lines should remain in the original order after the sort.

### Actual Behavior
The lines incorrectly change order:
image

### Additional

This bug sounds similar to the bug described in #8674 but it is distinctly different.

### Debug Information
Notepad++ v7.8.9 (32-bit)
Path : C:\.............\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252

sasumner:
Fixes #8716

Note that when a column block begins in (user) column 1, `_fromColumn` is zero; in the original code this would cause `isSortingSpecificColumns()` to return `false`.

### DIFF
--- a/PowerEditor/src/MISC/Common/Sorters.h
+++ b/PowerEditor/src/MISC/Common/Sorters.h
@@ -50,9 +50,9 @@ protected:
 	{
 		if (isSortingSpecificColumns())
 		{
-			// prevent an std::out_of_range exception
 			if (input.length() < _fromColumn)
 			{
+				// prevent an std::out_of_range exception
 				return TEXT("");
 			}
 			else if (_fromColumn == _toColumn)
@@ -74,7 +74,7 @@ protected:
 
 	bool isSortingSpecificColumns()
 	{
-		return _fromColumn != 0 || _toColumn != 0;
+		return _toColumn != 0;
 	}
 
 public:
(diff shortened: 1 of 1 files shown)


donho:
@sasumner 
Isn't `return !( _fromColumn == 0 && _toColumn == 0);` more explicit?
Just want to have your thought.

sasumner:
@donho

>Isn't return !( _fromColumn == 0 && _toColumn == 0); more explicit?

I thought about doing it that way.
It seems more complicated but it probably reads better.

I think this would also work:
`return _toColumn != 0;`

Let me know if I should change it to your version; it's fine.

sasumner:
Implemented the `return _toColumn !=0;`.

Also, moved an unrelated comment slightly to have it make more sense.
Should have been moved with #8674 changes but wasn't.
I hope it is OK to move that comment here.

donho:
@sasumner 
I'm not a psychopath against to all "unrelated" modification. I'm OK with all modification in the right direction, if it doesn't prevent me from code review. :)

Issue #8719, opened 2020-08-16
Add keyboard shortcuts for switching to Project Panels, Folder as Workspace and Function List


Uhf7:
I would like to add user-definable keyboard shortcuts for the following functions:

- Switch to Project Panel 1
- Switch to Project Panel 2
- Switch to Project Panel 3
- Switch to Folder as Workspace
- Switch to Function List

In the View menu, this panels can be turned on and off, there is a checkmark beside them, while they are turned on.
The keyboard shortcuts for this (toggling) menu commands are already there too.

Now I want to add keyboard shortcuts, which can be used to switch the focus to this panels without mouse. This keyboard shortcuts are not needed/useless in the menu.

donho:
A quick and rough spec needs to be talked here:

For all docking panels, there are 3 main actions for managing the window:
1. Launch it 
2. Switch it
3. Close it

I see the intention is use the same commands `A` for togging (launch & close) and the invisible commands`B` for switching.
My question is, what it happens if a panel isn't launched yet (for example "Function List"), but the user trigger the shortcut of `B`?

donho:
So AFAICT, `A` is done for all the docking panels, you want to implement `B`, and they are shown only in Shortcut Mapper.
Do you confirm?

How about for Document Map ?

Uhf7:
Yes, I confirm, the `A` commands are already there, and the `B` commands are only needed as keyboard shortcuts. From the menu, the `A` commands can be used.

And I left out the Document Map, because this can only be used with the mouse, so there should be no need to switch the keyboard input focus to it with a keyboard shortcut.

PR #8720, opened 2020-08-16
Add keyboard shortcuts for switching to panels.

Uhf7:
Implements #8719.

For all panels except the Function List, everything went smooth. The Function List, however, needed a few optimizations:
- Removed an useless `reload` call from `Notepad_plus::launchFunctionList`, because reloading the Function List implies parsing, which is causing delay. So I rely on the `reload` call in `DMN_SWITCHIN` notification now, when the Function List is launched.
- In `reload`:
  - Moved the folding state restauration to after the sort, to avoid loosing the current position in the tree.
  - Minimized the number of calls of `WM_SETTEXT` for the filter edit control.
  - Removed one sort call, because sorting is already done as a result of the `WM_SETTEXT` call. It triggers `EN_CHANGE`, even if the text remains the same, and ends up sorting the main view or the filtered view if needed.


PowerEditor/src/NppCommands.cpp
```diff
@@ -777,8 +789,9 @@ void Notepad_plus::command(int id)
 		break;
 
 		case IDM_VIEW_FUNC_LIST:
+		case IDM_VIEW_SWITCHTO_FUNC_LIST:
```
donho, line 792: Should IDM_VIEW_SWITCHTO_FUNC_LIST share IDM_VIEW_FUNC_LIST's routine?
I've got a performance here by switching to function list: function list has been launched and displayed, and the focus is on Folder as Workspace panel. After switching to function list by trigger assigned shortcut, function list reloaded and focused. But what I expected is no reload, just focus on it (Like mous clicked).

Uhf7: Yes, omitting the reload on re-launches would make the delays go away and the restore tree position would not occur. This would be nicer, for `IDM_VIEW_FUNC_LIST` too. The Function List is rebuilt anyway when the edit buffer switches (at least as long as it is visible). I will look into the mechanisms again and omit the reparse/rebuild process if possible. If the function list changed because of modifications in the edit buffer, then the refresh button can be used to reparse/rebuild.

Uhf7: I did three modifications:
- I removed the `reload` call for the Function List from the `DMN_SWITCHIN` notification. 
- Instead, I call `reload` in `launchFunctionList` again, but only, when the Function List is created the first time.
The reload after buffer switching is already done in `Notepad_plus::notifyBufferActivated`.
- I Modified the `cmdID` passed to `launchProjectPanel`: If a Project Panel is started with the new `IDM_VIEW_SWITCHTO_PROJECT_PANEL_n` command, then it gets the corresponding `IDM_VIEW_PROJECT_PANEL_n` as command id. This was necessary for two reasons:
  - Keep the same numbers in config.xml, independent of the command the Project Panel was launched with.
  - The `tTbData.dlgID` contains the toggling commands for all other panels, the Project Panels should do the same.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6214,12 +6214,10 @@ void Notepad_plus::launchFunctionList()
 
 		_pFuncList->setBackgroundColor(bgColor);
 		_pFuncList->setForegroundColor(fgColor);
+		_pFuncList->reload();
```
donho, line 6217: This modification made a regression, step to reproduce:
1. Open 2 different xml documents - xml_1 & xml_2, xml_2 is current document
2. Open Function List Panel - in which xml_2 is loaded.
3. Open Project Panel 1, if it is not in the same docking window of Function List, drag it in it.
4. Switch to xml_1 from xml_2
5. Switch to  Function List from Project Panel 1 (by using mouse or shortcut). 

Result before applying this PR:
Function List loads xml_1 content - Correct

Result after applying this PR:
Function List remains xml_2 content - Incorrect

Conclusion: reload() shouldn't be here. It should be where it was.

PowerEditor/src/NppCommands.cpp
donho: What I meant by "Should IDM_VIEW_SWITCHTO_FUNC_LIST share the same routine of IDM_VIEW_FUNC_LIST?" is that you should do the following without changing the current logic to avoid the regression:
```
		case IDM_VIEW_SWITCHTO_FUNC_LIST:
		{
			if (_pFuncList->isVisible())
			{
				_pFuncList->getFocus();
			}
			else
			{
				checkMenuItem(IDM_VIEW_FUNC_LIST, true);
				_toolBar.setCheck(IDM_VIEW_FUNC_LIST, true);
				launchFunctionList();
				_pFuncList->setClosed(false);
			}
		}
		break;
		
		case IDM_VIEW_FUNC_LIST:
		{
                ...
```

PowerEditor/src/Parameters.cpp
```diff
@@ -401,6 +401,12 @@ static const WinMenuKeyDefinition winKeyDefs[] =
 	// In localized environments, the text comes preferably from xml Menu/Main/Commands.
 	{ VK_NULL,    IDM_UPDATE_NPP,                               false, false, false, TEXT("Update Notepad++") },
 	{ VK_NULL,    IDM_CONFUPDATERPROXY,                         false, false, false, TEXT("Set Updater Proxy...") },
+
+	{ VK_NULL,    IDM_VIEW_SWITCHTO_PROJECT_PANEL_1,            false, false, false, TEXT("Switch to Project Panel 1") },
+	{ VK_NULL,    IDM_VIEW_SWITCHTO_PROJECT_PANEL_2,            false, false, false, TEXT("Switch to Project Panel 2") },
+	{ VK_NULL,    IDM_VIEW_SWITCHTO_PROJECT_PANEL_3,            false, false, false, TEXT("Switch to Project Panel 3") },
+	{ VK_NULL,    IDM_VIEW_SWITCHTO_FILEBROWSER,                false, false, false, TEXT("Switch to Folder as Workspace") },
+	{ VK_NULL,    IDM_VIEW_SWITCHTO_FUNC_LIST,                  false, false, false, TEXT("Switch to Function List") },
```
donho, line 409: They should be added right after the following line:
`{ VK_NULL,    IDM_VIEW_FUNC_LIST,                           false, false, false, TEXT("Toggle Function List") },`
so users can spot them in shortcut mapper much easier.

Uhf7: This one is easy, it is coming with the next revision of this PR.

PowerEditor/src/NppCommands.cpp
Uhf7: Didn't see this while writing my last comment, I will check it out tomorrow.

donho (approved):

Issue #8697, opened 2020-08-12
Bug when using Window menu to close last file open in second view


sasumner:
### Steps to Reproduce (STR) the Issue
1. Start with "new 1" as the only N++ tab (from a fresh install or portable version unzip)
2. Create new tab with Ctrl+n; now have "new 2" in addition to "new 1"
3. Right-click "new 2" and choose _Move to Other View_
4. Invoke the _Window_ menu and choose _Windows..._; observe:
![image](https://user-images.githubusercontent.com/30118311/90026985-0670b700-dc86-11ea-81d2-8e0cbb6d36f3.png)
5. Click "Close window(s)" button

### Expected Behavior
"new 2" and the second view should close, and the "Windows" window should also close.

### Actual Behavior
Observe that "new 2" and the second view closes, but the "Windows" window remains open and still shows a now-unknown "new 2":
![image](https://user-images.githubusercontent.com/30118311/90027079-1e483b00-dc86-11ea-9c05-fe69eba83f89.png)

### Second set of STR (slightly different conditions)

If "new 2" is a modified file, it is a bit different.
Repeat the steps above but modify "new 2" in some way before invoking the _Window_ menu:
![image](https://user-images.githubusercontent.com/30118311/90027421-87c84980-dc86-11ea-9065-a041ac150098.png)
After clicking "Close window(s)" button (on the _Windows_ window) and declining to save the modified "new 2" tab...
Observe that "new 2" and the second view closes, but the "Windows" window now shows some unknown "new 3":
![image](https://user-images.githubusercontent.com/30118311/90027472-944ca200-dc86-11ea-904f-1ceb99b10059.png)

### Debug Information
Notepad++ v7.8.9   (32-bit)
Build time : Jul 15 2020 - 20:26:50
Path : C:\\........\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1339
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll

donho:
@sasumner  Could you confirm if https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8721 has fixed the bug and do the code review please?

PR #8721, opened 2020-08-16
Issue 8697 bug when using window menu to close last file open in second view

eljefe7000:
This is a fix for Issue [8697](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8697) (Bug when using Window menu to close last file open in second view).  

The code was not checking to see if _idxMap was empty, and _idxMap gets decremented when the user clicks on the "Close Window(s)" button, so I added code to send the dialog a WM_CLOSE message if _idxMap.size() is less than 1.


sasumner (dismissed):
In testing the PR, I noticed that `new 2` remains in the `Document Switcher` window after closure of that file in the second view -- it should also be removed from there:

image

donho:
@sasumner  Just checked it in master, indeed I can reproduce it as well: 
With 2 new docs in 2 views respectively, this bug can be reproduced not only in Window dialog, but also close it directly via the close button on document tab.

sasumner:
@donho So it seems unrelated to this PR then.
Should open a separate issue for that, I suppose?
Continue evaluating this PR but ignoring that?

donho:
@sasumner 
This PR works around the problem - it's a solution though I don't really like it. In the mean while, I have no better solution so it'll be accepted.

> Should open a separate issue for that, I suppose?

Yes, please. I will close #8697 with this PR.

Issue #8447, opened 2020-06-21
Sorting of stream selection can do one too many lines


sasumner:
### Description of the Issue
When using the _Edit_ menu > _Line Operations_ > _Sort_ > ... operations on a stream selection, the line at the bottom end (at the higher position) can be unintentionally included in the sort.

### Steps to Reproduce the Issue
1. Create some text in a new N++ tab:
```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```
2. Select the lines I wish to sort via a single stream selection, by clicking in the line number margin starting in line 1 and dragging down to cover the lines I want to operate on:
![image](https://user-images.githubusercontent.com/30118311/85214274-b08f2b00-b336-11ea-841a-29dd982fb8e3.png)

3. Invoke the _Edit_ menu > _Line Operations_ > _Sort_ > _Sort Lines Lexographically Ascending_ command.

### Expected Behavior
I expected line 6 `And Jill came tumbling after.` to **_NOT_** be included in the lines to be sorted.

### Actual Behavior
Line 6 `And Jill came tumbling after.` **_IS_** included in the lines sorted:

![image](https://user-images.githubusercontent.com/30118311/85214304-f77d2080-b336-11ea-9703-8681ecd79e6d.png)

### Other

I do realize that if I don't include the line ending of line 5 in the stream selection, line 6 will not be included in the group of lines to be sorted.  However, this feels unnatural and is hard to remember to do before conducting the sort.

### Debug Information
Notepad++ v7.8.7   (32-bit)
Build time : Jun  5 2020 - 02:34:28
Path : C:\\..............\npp.7.8.7.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

Yaron10:
NPP would show **6** selected lines in the status bar.

IMO it's wrong and I've changed `bool getSelectedCount(int & selByte, int & selLine) const` in `ScintillaEditView.h` to:
```
	bool getSelectedCount(int & selByte, int & selLine) const
	{
		// return false if it's multi-selection or rectangle selection
[...]

PR #8740, opened 2020-08-18
Fix no-selection sort including empty last line of document

### LINKED ISSUE #8739: No-selection sort awkwardly includes empty line at end of document
### Description of the Issue

With no active selection and uniform lines in a document tab, sorting produces an odd result; see STR and results below:

1. Create the following text in a new Notepad++ tab:
```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```

2. Turn on visible line-endings; observe:
image

3. With no active selection (as shown), execute _Edit > Line Operations > Sort Lines Lexicographically Ascending_.

### Expected Behavior

image

### Actual Behavior

I obtain an unwanted empty line at the top of my file, and no line-ending on the last line of my file:

image

### Additional

Sorting the document data after doing a _Select All_ (Ctrl+a) produces the expected result (perhaps owing to the fixing of #8447 ).

### Debug Information

Notepad++ v7.8.9   (32-bit)
Path : C:\\...........\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1339
Current ANSI codepage : 1252

sasumner:
Fixes #8739

Also fixes a situation where a harmless one-line document is let through to actually be sorted (consider the original code when `SCI_GETLINECOUNT` returns a 1).

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -599,6 +599,19 @@ void Notepad_plus::command(int id)
 					// No selection.
 					fromLine = 0;
 					toLine = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+					if (_pEditView->execute(SCI_LINELENGTH, toLine) == 0)
+					{
+						if (toLine > 0)
+						{
+							// Last line is zero-length, do not include it in the lines to be sorted
+							--toLine;
+						}
+					}
+					if (toLine == 0)
+					{
+						// Sorting a single line is not allowed
+						return;
+					}
 				}
 			}
 


donho:
I use this sample to test sorting:
image

Before applying this PR:
image


After applying this PR:
image

So for me this "new feature" will bring more confusion to users than the current behaviour which is coherent.

sasumner:
@donho 

I understand your example but I don't think it is a typical use-case.
Your sample contains intentional empty lines at end of document.  

Truly, in that case, if user sorts entire document, they probably don't want any of the empty lines included in the sort.
But we don't try to fix that as their empty lines were intentional.

The case I was trying to catch is where user wants every line to have a line-ending, but doesn't want the "fake" line at the end to be pulled into the sort.

Inevitably, following such a sort, user will have to:

* take time to figure out why this strange empty line is now at top of document
* navigate to top of document to remove
* notice that last line no longer has an end-of-line character(s) on it
* re-add the end-of-line character to the last line

There is support on Community by at least 4 users for this issue/PR:  
https://community.notepad-plus-plus.org/post/57151
Start reading with "Alan Kilborn" posting that starts with _Here’s a situation that is still bad IMO even with 7.8.9_ and read several postings.

donho:
My test is not an extravagant sample.
For me this behaviour is not consistent so I won't include it.

sasumner:
With rejection of this, I'm not sure if a (minor) inconsistency is left remaining.

In the "issue" for this PR was the following note:
`Sorting the document data after doing a Select All (Ctrl+a) produces the expected result (perhaps owing to the fixing of #8447 ).`

If there's no difference between "act on entire document" and "act on selection created by ctrl+a" then I think we have an inconsistency remaining.

Issue #8692, opened 2020-08-11
[Feature Request] Add number of hit lines to file-level search output


sasumner:
Much like #8154 augmented the file-level search output with additional helpful information, it would be nice to enhance it again in such a way as to see the number of lines having hits in each file.  A picture may describe the proposal best:

Current output from a file-level search (N++ 7.8.9):

![image](https://user-images.githubusercontent.com/30118311/89946781-71bc7980-dbf1-11ea-9513-7731803849e0.png)

Proposed improved output:

![image](https://user-images.githubusercontent.com/30118311/89946375-cf03fb00-dbf0-11ea-9bbb-70c8937ee904.png)

Yaron10:
@sasumner,

This is a nice idea and would be a useful addition. :+1: 

I suppose the option to display only one line with multiple *clickable* occurrences has been suggested and discussed before.
Here is an illustration from EditPad.

![תמונה](https://user-images.githubusercontent.com/10229320/90056345-664b7a00-dce7-11ea-98ce-814fd1a90ed6.png)

sasumner:
>I suppose the option to display only one line with multiple clickable occurrences has been suggested and discussed before.

So it is Off-Topic for the current issue, but:

Well... the "multiple-clickable-occurrences" part gets into touching the code for "Search result" lexing, which already has some bugs as discussed before (8525 and others), which is "part of" Scintilla and thus not updated frequently, and anyway quite frankly scares me.  :-)

This general idea is put forth, without the "multiple-clickable-occurrences" part, in #1499, and indeed I am considering implementing that (in a limited way, sorry, for now no multiple-clickables!).

I noticed [THIS](https://community.notepad-plus-plus.org/topic/19837/find-all-in-current-document-not-displaying-complete-lines) where someone wanted (or actually, didn't want) **_"...the cursor jumps to the main window but at the location of the characters that were found, not iat the position I was in the Find results window."_** and so that gave me the idea that double-clicking (probably with a modifier, e.g., Ctrl) could jump the user not to the hit, but to the position double-clicked at.  I mention it here because it is semi-related to what you were saying; perhaps it could become part of 1499 's implementation, or it could even stand alone as its own issue.

sasumner:
ON-Topic:

>This is a nice idea and would be a useful addition. 

Thank you (for offering an opinion, the fact that it is a positive one is just a bonus :-) )
[...]

PR #8741, opened 2020-08-18
Add line count to hits output in search results

sasumner:
Implements #8692

@donho :
It appears to have some value, although it was my own feature request. :-)
Maybe it is too intrusive in the output?
I don't know, let me know what you think...


donho:
For me it's too many lines of code for this minor feature. Sorry, but I don't think it's worth.

sasumner:
@donho 

>For me it's too many lines of code for this minor feature. 
Sorry, but I don't think it's worth.

Okay.
Perhaps if I'd have said it was the "first part" of a larger feature I was working on, it would have won you over to my side.  :-)

donho:
@sasumner  You can always say it :)

PR #8752, opened 2020-08-21
Fix regular expression has an unescaped '.' (dot) characters

dwisiswant0:
### Description

Escaped  `.` character before-and-after `scintilla`

Escape all meta-characters appropriately when constructing regular expressions for security checks, pay special attention to the `.` meta-character.
If a regular expression implements such a check, it is easy to accidentally make the check too permissive by not escaping the . meta-characters appropriately. Even if the check is not used in a security-critical context, the incomplete check may still cause undesirable behaviors when it accidentally succeeds.


donho:
You should submit the PR to [Scintilla project](https://www.scintilla.org/).
But still, I don't think it's a bug.

PR #8753, opened 2020-08-23
Close a single panel instead of the entire stack.

Uhf7:
Clicking the "Close" icon in docked windows closes the current panel now instead of the entire stack.

This implements the last part of #8471.

### DIFF
--- a/PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
+++ b/PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
@@ -1136,33 +1136,25 @@ void DockingCont::onSize()
 
 void DockingCont::doClose()
 {
-	int	iItemOff = 0;
 	int	iItemCnt = static_cast<int32_t>(::SendMessage(_hContTab, TCM_GETITEMCOUNT, 0, 0));
+	int iItemCur = getActiveTb();
 
-	for (int iItem = 0; iItem < iItemCnt; ++iItem)
+	TCITEM		tcItem		= {0};
+
+	tcItem.mask	= TCIF_PARAM;
+	::SendMessage(_hContTab, TCM_GETITEM, iItemCur, reinterpret_cast<LPARAM>(&tcItem));
+	if (tcItem.lParam)
 	{
-		TCITEM		tcItem		= {0};
-
-		// get item data
-		selectTab(iItemOff);
-		tcItem.mask	= TCIF_PARAM;
-		::SendMessage(_hContTab, TCM_GETITEM, iItemOff, reinterpret_cast<LPARAM>(&tcItem));
-		if (!tcItem.lParam)
-			continue;
-
 		// notify child windows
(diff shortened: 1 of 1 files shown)


Yaron10:
Just found out that `Shift + Clicking Close` closes the entire stack in MS VS.
For example: `Find Results 1` + `Find Results 2`.

Uhf7:
OK, I see. There is no way to get around it :-). I would prefer Shift + clicking Close then.

In MS VS, it is even configurable, whether the close button affects only the active panel or the entire stack : "Close button affects active tool window only" is the option. Must have been checked all the time I used it. And this setting is (in MS VS) negated by the Shift key:
- Option "Close button affects active tool window only" on: Click on close closes active tool window, Shift + click on Close closes all tool windows
- Option "Close button affects active tool window only" off: Click on close closes all tool windows, Shift + click on Close closes active tool window.

So, I see some white space left in the MISC. page of preferences ...

Yaron10:
In MS VS you can't see the forest for the trees.

> So, I see some white space left in the MISC. page of preferences ...

If you think it's important. :)

Thank you. 
:+1:

KnIfER:
I think it's more convenient to press ctrl+w and close a single panel .  

(I have a mouse button bound  to ctrl+w, and I am using it every where )

Uhf7:
> I think it's more convenient to press ctrl+w and close a single panel 

Started Notepad++, pressed Ctrl+W. While in text windows, it inserts a control character ETB, while in docking windows, nothing happens.

Is this a kind of hidden function that already exists? Or is it just a kind of feature request or another way in which it could be done?

sasumner:
@Uhf7 

>Started Notepad++, pressed Ctrl+W. While in text windows, it inserts a control character ETB...

Default behavior of Ctrl+w should be to close the tab.

If it is intended to be hidden, it isn't hidden very well:

image

But, historically, in Windows, Ctrl+w has been a "close tab" command.  I think it has fallen out of favor and Ctrl+F4 is now the favorite for this.  Why F4 I have no idea.

Uhf7:
> historically, in Windows, Ctrl+w has been a "close tab" command

OK, but this cannot be applied here, because the "Close" keyboard shortcut is for documents only. So, if we would want to use it for side windows too, we would have to make the "Close" keyboard shortcut dependent on in which window it is pressed. If it is pressed in the edit window, then the current document would be closed, if it is pressed in a side window, the current tab of the side window would be closed. That's actually not what I intent here.

The question still lingering from my side is, in which degree the click to the "Close" icon of a side window should be configurable. Is it enough to provide a Shift-Click for closing all tabs in the side window, or would a configuration be acceptable, so that all users, who got used to it, can continue to close the entire stack instead of the current tab only.

sasumner:
>OK, but this cannot be applied here, because the "Close" keyboard shortcut is for documents only.

Correct, I was NOT suggesting it, just relaying some history.

PR #8760, opened 2020-08-24
Disable controls on unshown tabs of Find dialog

### LINKED ISSUE #8696: Find dialog keyboard accelerators can be broken by localization
### Description of the Issue
If I use the `english_customizable.xml` localization, then two buttons in the Find/Replace dialog have **Alt-R** as accelerator:
- _Replace_ in the _Replace_ page.
- _Replace in Files_ in the _Find in Files page_.

It happens, that I have the _Replace_ page open, and with pressing **Alt-R** in the _Replace_ page I start the _Replace in Files_ function from the _Find in Files page_. Which has a safety question, which made me happy here.

1. Start Notepad++ and set localization to _English (customizable)_.
2. Open the Find/Replace dialog.
3. Verify, that the _Replace_ button in the _Replace_ page has an underlined "R" (sometimes it becomes visible only after pressing the Alt key).
4. Verify, that the _Replace in Files_ button in the _Find in Files_ page has an underlined "R".
5. Go back to the _Replace_ page. 
6. Fill in something into the _Find what_ and _Replace with_ fields, like _A_ and _A_.
7. Press **Alt-R**. The function works like expected.
8. Use the **Tab** key to move the input focus to the _Replace_ button (in the _Replace_ page).
9. Press **Alt-R**. again.

### Expected Behavior
A normal _Replace_ function as happened in step 7.

### Actual Behavior
The _Replace in Files_ function from the _Find in Files_ page starts.

### Debug Information
Notepad++ v7.8.9   (32-bit)
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Current ANSI codepage : 1252
Plugins : none

sasumner:
Fixes #8696 

...and does some other good things as well, in the course of fixing 8696.

My PR:

- Fixes the original issue
- Fixes the problem of being able to affect states of controls on unseen tabs via keyboard accelerators (e.g. in N++ 7.8.9, on _Find_ tab, note the state of the _Wrap around_ control; move to the _Find in Files_ tab (which has no _Wrap around_ control) and press Alt+p, then go to the Find tab and notice that the state of _Wrap around_ has changed)
- Fixes the problem of each tab in the Find family needing to have unique keyboard accelerators (before it was not realistically possible to have the same accelerator on more than one tab, after this PR, it is fine); example: Alt+r is defined by default as the accelerator for _Replace_ on the _Replace_ tab and also _Replace in Files_ on the _Find in Files_ tab.  Now these can peacefully coexist -- the _Replace_ button action will be fired when Alt+r is pressed while the _Replace_ tab is visible, and the _Replace in Files_ action will occur when Alt+r is pressed when the user is looking at the _Find in Files_ tab.
- Removes the unknown/unused control `IDD_FINDINFILES_GOBACK_BUTTON`

So before this PR, controls on the Find dialog that should not be seen by the user were hidden using ::ShowWindow(..., SW_HIDE).  This was fine as it allowed the enable/disable state to change independently regardless of whether a control was hidden or shown.  The bad part with this method is that keyboard accelerators can be invoked by the user on hidden controls (that are enabled), and will invoke the control's functionality.

What is needed instead, and what is provided by this PR, is to disable (::EnableWindow(..., FALSE) ) a control right after it is hidden with SW_HIDE.  However, this means that the true enabled/disabled state of the control must be remembered in another way.

As a control is being hidden, its enable/disable state is remembered into a map/dictionary (surprisingly this would be a first for N++ -- the use of std::map) keyed by control id.  Then the control is hidden AND disabled.  When the time comes to show the control again, its enable/disable state is looked up in the map and the control is shown and either enabled or disabled, as appropriate from the map data.

When code executes that changes the enable/disable state of a hidden control, with the intention of enabling it, the control is NOT enabled, but rather the map remembering the desired state is updated.  At such a time when the control is shown again, the map is examined and the correct enable/disable state is applied based upon that.

Thus the code provides replacement functions for `::ShowWindow()` and `::EnableWindow()` that a programmer should call when hiding/showing and enabling/disabling controls in the Find dialog:  After this PR the programmer should instead call `showFindDlgCtrl()` and `enableFindDlgCtrl()`, respectively.


sasumner:
One thing that will make me sad if this PR adopted:  I have Alt+t set to run _Find All in Current Document_.  Currently this works from ANY of the four Find window tabs.  If this PR is adopted, then one of my personal favorite things stops working. :-(

But...the PR is the right thing to do, so I could adapt to switching my tab over to the _Find_ tab and then pressing my Alt+t. :-)

donho (changes requested):
1. Rename the functions' name and their parameter's name.
2. Remove `&amp;` from `<Item id="1660" name="&amp;Replace in Files"/>` in both `english.xml` & `english_customizable.xml` to  make them match with the current situation (in .rc).

donho (approved):

PR #8761, opened 2020-08-25
Enhancement ergonomic of Auto-Completion

GOLEM777:
1 - Increased block Auto-indent (little space for some languages)
2 - Enhancemed ergonomics for future options
ps
Need  change "Auto-indent" on "Auto-indent with new line"

capture_08252020_034652


donho:
It's quite subjective, but I don't feel the layout you suggested is better than current one.

Issue #8757, opened 2020-08-24
Uni-tests for function-list


donho:
As mentioned in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/3393, We need the uni-tests for function-list so users can enhance `functionList.xml` without regression.
 
Currently Notepad++ is capable to get input source file and output a result in JSON format:
Command:
```
notepad++.exe -export=functionList -lcpp c:\funcListTests\cppTest.cpp
```
the behaviour:

> Notepad++ will open "c:\funcListTests\cppTest.cpp" as C++ source file, and write result to `c:\funcListTests\cppTest.result.json` - the old `cppTest.result.json` will be erased by the new one.

All we need is have all input files in `A` folder and generate expected results in `B` folder, then in build system (AppVeyor) to integrate the ability to evoke notepad++ (in each build) to take files from A folder, and generate test files into A folder - this system should compare result files between `A` & `B` folder.

Any help for providing **input files** and for **AppVeyor** is welcome.

GitMensch:
Nice one, so everything s setup to add new unit tests now.
The only thing is that there's so far only the possibility for one unitTest per language; is this the way to go (so always add to the existing unit test if it exists for the language) or should there be a wildcard (so generating `$testRoot$__\unitTest*` and later comparing it with the appropriate expected result)?

Just asking to be able to come up with a README for those unit tests later...

donho:
@GitMensch Currently there's only way to run unit-tests - run them all. We can do however a generic script aside for running the test by passing directory name as argument. You can create a feature request in issue tracker if you think it's useful.

GitMensch:
I see and that is good to remember. My question was: should it be possible to have more than one unittest for one language identifier?

donho:
@GitMensch  2 or more unittests for one language? Currently no, and I don't see why we will need it in the future.

PR #8762, opened 2020-08-25
Integrate unit-tests for function-list

donho:
Implement #8757, close #3393

`unitTestLauncher.ps1` should be launched by Appveyor, if this script returns -1 then build should fail, otherwise it will return 0 so build continues.


PowerEditor/Test/FunctionList/bash/unitTest.expected.result
```diff
@@ -1 +1 @@
-{"leaves":["setenv","action_needed","canonicalize_path","make_module","db2_name","versioned_copy"],"root":"unitTest"}
\ No newline at end of file
+{"leaves":["action_needed","canonicalize_path","make_module","db2_name","versioned_copy"],"root":"unitTest"}
```
donho, line 1: "setenv" should be present IMO, so there has been already a regression.

donho:
@GitMensch @chcg Here the working unit-tests for function-list.
Any idea to integrate it into appveyor.yml ?

chcg:
@donho It did a first version, see https://github.com/chcg/notepad-plus-plus/tree/issue_8762 with result https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/34984122. The problem seems to come from scintilla SciLexer.dll which is not build with boost regex support by appveyor. Probably the CI build could be modified to use the prebuild boost libraries from the appveyor environment, see https://www.appveyor.com/docs/windows-images-software/#boost

I will give that a try.

PS:
Additionally there seems to be a problem with the powershell script, if it runs on the #net core version (https://github.com/PowerShell/PowerShell/releases/tag/v7.0.3)

chcg:
Boost is just available as source code, but not with binary libs for regex. So probably the easiest check would be against some prebuild version like the last release e.g or at least the SciLexer.dll from such a build.

PR #8769, opened 2020-08-26
Scintilla fix for hovered indicators

Uhf7:
To make the hovered blue underline links really shine, Scintilla needs another fix.

Without it, the URL becomes only partially hover-highlighted, if
- The hovered style is `INDIC_EXLORERLINK` or `INDIC_TEXTFORE`
AND
- There is a line break or a syntax highlighting change within the URL.
  (The line break is only possible with word wrapping on)

The problem was discovered by @Yaron10's thorough testing and is mentioned 1st in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8634#issuecomment-678768659. 

I found out, that the detection, whether a character belonging to an indicator range, is currently hovered or not, didn't work in the text drawing part of the code (The part of the code, which is responsible for the decorations, like underlines or boxes, works).

I filed this as https://sourceforge.net/p/scintilla/bugs/2199/, and a fix for this has been integrated in the original Scintilla. 

What is now left to do, is to integrate the fix in the Notepad++ version of the `SciLexer.dll` too. Sorry for the trouble, but we are real indicator pioneers here.

### DIFF
--- a/scintilla/src/EditView.cxx
+++ b/scintilla/src/EditView.cxx
@@ -1810,9 +1810,12 @@ void EditView::DrawForeground(Surface *surface, const EditModel &model, const Vi
 					const int indicatorValue = deco->ValueAt(ts.start + posLineStart);
 					if (indicatorValue) {
 						const Indicator &indicator = vsDraw.indicators[deco->Indicator()];
-						const bool hover = indicator.IsDynamic() &&
-							((model.hoverIndicatorPos >= ts.start + posLineStart) &&
-							(model.hoverIndicatorPos <= ts.end() + posLineStart));
+						bool hover = false;
+						if (indicator.IsDynamic()) {
+							const Sci::Position startPos = ts.start + posLineStart;
+							const Range rangeRun(deco->StartRun(startPos), deco->EndRun(startPos));
+							hover =	rangeRun.ContainsCharacter(model.hoverIndicatorPos);
+						}
 						if (hover) {
 							if ((indicator.sacHover.style == INDIC_TEXTFORE) || (indicator.sacHover.style == INDIC_EXPLORERLINK)) {
 								textFore = indicator.sacHover.fore;


donho:
@Uhf7 
> To make the hovered blue underline links really shine, Scintilla needs another fix.

A (2) picture(s) will explain what you mean by "really shine".

Uhf7:
> A (2) picture(s) will explain what you mean by "really shine".

The picture has been provided by @Yaron10, in the https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8634#issuecomment-678768659 comment already mentioned. The blue part of the last link in this picture is the only partly highlighted link. With this Scintilla fix, the whole link gets blue while hovering.

> BTW, It's already too many modification in scintilla for such feature.

This is a bug in Scintilla, and it has been fixed in the original version of Scintilla already. I can't image that you want to keep it. Or that you want to wait until the new Scintilla version comes out. I'm only the messenger here.

And I can't promise, that this is the last Scintilla bug regarding indicators that needs to be fixed. And I can't promise either, that I will not to have suggest further fixes or improvements of the `addHotSpot` function either.

The only thing I know is, that we are going forward here, and this is better than backward.

donho:
> The picture has been provided by @Yaron10, in the #8634 (comment) comment already mentioned. The blue part of the last link in this picture is the only partly highlighted link. With this Scintilla fix, the whole link gets blue while hovering.

I mean a pic BEFORE applying your PR and a pic AFTER applying your PR

> Or that you want to wait until the new Scintilla version comes out. I'm only the messenger here.

I will judge it after seeing the screenshots.

Yaron10:
**Case 1:**

Word-Wrap: ON.
Paste a link in a normal text file.
Make sure the link is wrapped.
Hover.

**Before:**
תמונה

**After:**
תמונה

***

**Case 2:**

Paste `https://www.youtube.com/watch?v=VmcftreqQ6E&list=PnQIRE5O5JpiLL&index=1` in an XML file.
Hover.

**Before:**
תמונה

**After:**
תמונה

PR #8776, opened 2020-08-28
Enable and disable controls correctly for reg exp search mode in find-in-finder

### LINKED ISSUE #8768: Find-in-found-results allows whole-word option when searching for regular expression
### Description of the Issue
Some controls in the _Find in these found results_ window aren't correctly enabled/disabled based upon the _Search mode_ (mainly related to the _Regular expression_ mode)

1. Do a _Find All in Current Document_ search to get the _Find result_ window populating and showing.
2. Right-click in the _Find result_ window and select _Find in these found results..._ from the popup menu that appears.
3. In the _Find in found results_ window that has appeared, ensure that the _Search mode_ is set to _Regular expression_.

### Actual Behavior
Observe that the _Match whole word only_ control is ENABLED and possibly CHECKED; see screenshot.

image

### Expected Behavior
When in a regular expression searching mode, a whole-word search is not reasonable, so the _Match whole word only_ control should be DISABLED and UNCHECKED.

Additionally, the _. matches newline_ control should be DISABLED (but can be left either checked or unchecked) when **not** in _Regular expression_ _Search mode; currently it is ALWAYS enabled.

### Debug Information
Notepad++ v7.8.9   (32-bit)
Path : C:\\.........\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1339
Current ANSI codepage : 1252

sasumner:
Fixes #8768

-------------

@donho 

This change adds a new function `setChecked` into `StaticDialog.h` for use by the other code of this PR.

It simplifies an API call like this:

`::SendDlgItemMessage(_hSelf, IDC_MATCHLINENUM_CHECK_FIFOLDER, BM_SETCHECK, _options._isMatchLineNumber ? BST_CHECKED : BST_UNCHECKED, 0);
`

into something cleaner and easier to look at:

`setChecked(IDC_MATCHLINENUM_CHECK_FIFOLDER, _options._isMatchLineNumber);
`

I would probably not have added this function, but I noticed that `StaticDialog.h` already had the function `isCheckedOrNot` and I felt like if you have that function, why not create the companion function and call it `setChecked`?

This sets up a "refactoring" opportunity as there are about 200 other calls in the codebase that could be converted over to this cleaner/simpler style.  What do you think about doing that refactor?

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -686,15 +686,19 @@ void FindInFinderDlg::initFromOptions()
 	HWND hFindCombo = ::GetDlgItem(_hSelf, IDFINDWHAT_FIFOLDER);
 	addText2Combo(_options._str2Search.c_str(), hFindCombo);
 
-	::SendDlgItemMessage(_hSelf, IDC_MATCHLINENUM_CHECK_FIFOLDER, BM_SETCHECK, _options._isMatchLineNumber ? BST_CHECKED : BST_UNCHECKED, 0);
-	::SendDlgItemMessage(_hSelf, IDWHOLEWORD_FIFOLDER, BM_SETCHECK, _options._isWholeWord ? BST_CHECKED : BST_UNCHECKED, 0);
-	::SendDlgItemMessage(_hSelf, IDMATCHCASE_FIFOLDER, BM_SETCHECK, _options._isMatchCase ? BST_CHECKED : BST_UNCHECKED, 0);
-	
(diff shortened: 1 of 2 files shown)


donho:
@sasumner 
> This sets up a "refactoring" opportunity as there are about 200 other calls in the codebase that could be converted over to this cleaner/simpler style. What do you think about doing that refactor?

I wouldn't call them "refactoring" since there's no optimization of performance/structure, but only improvement of readability.
We can use this method when we create new code, but it's not necessary to replace the existing calls.

sasumner:
>I wouldn't call them "refactoring" since there's no optimization of performance/structure, but only the improvement of readability.

I just call it "refactoring" because it is shorter than "improvement of readability". :-)

>We can use this method when we create new code, but it's not necessary to replace the existing calls.

Sure.
The "problem" with that is:
* people may not realize that the new function exists
* people may be afraid to change the "style" of existing code, thinking it hurts the chance to get PR accepted

Issue #8524, opened 2020-07-04
[Feature Request] Enhance usefulness of status bar's "Sel :" area


sasumner:
### Description of the Feature Request

The `Sel :` area of the status bar is currently only useful in one situation:  

When I have a single stream selection it will show me the number of characters in my selection and the number of lines that my selection touches.  

**_Example 0_**: `Sel : 25 | 2` for 25 characters selected on two lines.
This is a good feature.

But what about when I have other types of selection situations?

**_Example 1_**: 
Multiple selections (after enabling _Multi-Edit Settings_ in the _Preferences_).

Result 1: 
In the status bar I see `Sel : N/A`
Not terribly useful.

**_Example 2_**:
Rectangular selections after selecting a column-block via Shift+Alt+arrows or Alt+LClick+drag.

Result 2:
In the status bar I see `Sel : N/A`
Not terribly useful.

**_Example 3_**:
One selection of zero characters; the normal single-carat situation.

Result 3:
In the status bar I see `Sel : 0 | 0`
Not terribly useful.

Wouldn't it be super if Notepad++ could display useful information for these situations?

Additionally, there is an off-by-1 error in the current output of the number of lines touched by a single selection; this should be corrected during any revamping of `Sel :` status bar functionality; for reference see https://community.notepad-plus-plus.org/topic/13087/sel-in-status-bar-is-off-by-1-on-line-count

### Expected Behavior

I expect to be shown some useful information in all types of selection situations.

### Actual Behavior

Described above.

### Debug Information

Reference version:

Notepad++ v7.8.8   (32-bit)
Build time : Jun 24 2020 - 00:15:36
Path : C:\\...............\npp.7.8.8.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit)
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

### Other

I decided to create a new and very specific issue for this based upon prior discussion in other issues:

* https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8447#issuecomment-647110458
* #7716

Yaron10:
@sasumner,

:+1: :+1: :+1: 
Please share anyway. :)

sasumner:
@donho 
Is there a reason for "suspended" label?
[...]

PR #8780, opened 2020-08-29
Enhance info provided in Sel portion of main status bar

sasumner:
Implements #8524

No selection:
Example: `Pos : 1,234`
Meaning: Single caret is at position 1233 in document (user position 1234)

Single stream selection (**no change to existing behavior**):
Example:  `Sel : 27 | 2`
Meaning:  27 characters selected on 2 lines

Multiple stream selections (this PR behavior):
Example:  `Sel 3 : 72 | 6`
Meaning:  3 selections of 72 characters on 6 lines

Rectangular selection with no virtual space in column block (this PR behavior):
Example:  `Sel : 2x4 = 8`
Meaning:  selection of 2 rows(lines) by 4 columns; 8 characters in that selection

Rectangular selection with some virtual space in column block (this PR behavior):
Example:  `Sel : 2x4 🡢 6`
Meaning:  selection of 2 rows(lines) by 4 columns; 6 real characters in that selection

------------------------

Note: Seemingly unrelated `IDM_EDIT_SPLIT_LINES` logic needed to be changed due to slight alteration in shared `getSelectionLinesRange()` function.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3228,33 +3228,65 @@ int Notepad_plus::wordCount()
 
 void Notepad_plus::updateStatusBar()
 {
-    TCHAR strLnCol[128];
+	TCHAR strLnCol[128];
 	TCHAR strSel[64];
-	int selByte = 0;
-	int selLine = 0;
 
-	_pEditView->getSelectedCount(selByte, selLine);
+	int selCount = static_cast<int>(_pEditView->execute(SCI_GETSELECTIONS));
 
-	long selected_length = _pEditView->getUnicodeSelectedLength();
-	if (selected_length != -1)
-		wsprintf(strSel, TEXT("Sel : %s | %s"), commafyInt(selected_length).c_str(), commafyInt(selLine).c_str());
+	const int maxSelsToProcessLineCount = 99;  // limit the number of selections to process, for performance reasons
+	const std::pair<int, int> selCharsAndLines = _pEditView->getSelectedCharsAndLinesCount(maxSelsToProcessLineCount);
+
+	if (_pEditView->execute(SCI_SELECTIONISRECTANGLE))
+	{
+		int rectAnchor = static_cast<int>(_pEditView->execute(SCI_GETCOLUMN, _pEditView->execute(SCI_GETRECTANGULARSELECTIONANCHOR)));
+		rectAnchor += static_cast<int>(_pEditView->execute(SCI_GETRECTANGULARSELECTIONANCHORVIRTUALSPACE));
+		int rectCaret = static_cast<int>(_pEditView->execute(SCI_GETCOLUMN, _pEditView->execute(SCI_GETRECTANGULARSELECTIONCARET)));
+		rectCaret += static_cast<int>(_pEditView->execute(SCI_GETRECTANGULARSELECTIONCARETVIRTUALSPACE));
+		int rectWidth = std::abs(rectCaret - rectAnchor);
+		bool hasVirtualSpace = selCharsAndLines.first != selCount * rectWidth;
+
+		wsprintf(strSel, TEXT("Sel : %sx%s %s %s"),
+			commafyInt(selCount).c_str(),
+			commafyInt(rectWidth).c_str(),
+			hasVirtualSpace ? TEXT("🡢") : TEXT("="),
+			commafyInt(selCharsAndLines.first).c_str());
+	}
+	else if (selCount > 1)
+	{
+		wsprintf(strSel, TEXT("Sel %s : %s | %s"),
+			commafyInt(selCount).c_str(),
+			commafyInt(selCharsAndLines.first).c_str(),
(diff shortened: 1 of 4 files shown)


sasumner:
Closing for now as this PR was just to build artifacts for some test volunteers.

Perhaps @Yaron @Ekopalypse wish to have a look?

Yaron10:
@sasumner,

:+1: 

The `TEXT("🡢")` idea if `hasVirtualSpace` is very nice.
It seems you've found a solution to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8524#issuecomment-654295098, correct?

I still think that the display would be much better and clearer if some of my format-comments were implemented.
I understand it's not entirely up to you.

Apropos comments: It would help if the various relevant threads were referenced in one place.

Thank you.

sasumner:
>The TEXT("🡢") idea if hasVirtualSpace is very nice.

I think that means the "right arrow"?
That one was all @donho !

>It seems you've found a solution to #8524 (comment), correct?

Yes and no.
If there are more than 99 selections (somewhat arbitrary number), then it will display `Sel XXX : cccc | ...` where the `...` is the thing I'm trying to point out.  In other words, it doesn't even try to figure out how many lines in this case, it just shows `...`.  This is because calculating that accurately takes a lot of processing, and we don't want to potentially impact performance in the worst of circumstances.

>I still think that the display would be much better and clearer if some of my format-comments were implemented.

I feel the same way about my ideas. :-)
The way of the world is compromise.

>Apropos comments: It would help if the various relevant threads were referenced in one place.

I don't know what this means.

Thank you for your comments.
It perhaps may give me confidence to reopen this PR. :-)
Let's see if we have other testers and commenters...

Yaron10:
> That one was all @donho !

So, well done @donho! :)

> Yes and no.
If there are more than 99 selections...

The "Yes" is very good. :+1: 
The "No" (more than 99 selections) is absolutely insignificant IMO. 

> The way of the world is compromise.

True.
I just wanted to emphasize that the format can make the difference between clarity and confusion.

>  Apropos comments: It would help if the various relevant threads were referenced in one place.
...
I don't know what this means.

We've discussed this issue in 3 or 4 different threads. It would be convenient to have them all referenced here.

> It perhaps may give me confidence to reopen this PR.

You certainly should.
I'm sure it would be very useful to those who frequently use multiple-selections.

sasumner:
>The "No" (more than 99 selections) is absolutely insignificant IMO.

Well, you'll get the users that run a script that selects every character in a document into its own one-character-long selection...  :-)

sasumner:
>We've discussed this issue in 3 or 4 different threads. It would be convenient to have them all referenced here.

I think they could all be tracked back via the issue comments.
And this PR links back to the issue.

Ekopalypse:
Looks good to me - if I'm right, then this **Sel 5: 25 | 4**  translates to 
5 selections with 25 characters selected and 4 lines involved. It is not intuitive but not confusing either.
I guess most users will understand when they have made a few selections.
As @Yaron10 already stated, the extra "virtual space" symbol is nice too.

sasumner:
>It is not intuitive but not confusing either.

Well, given that no one ever explained to me how it worked BEFORE this PR:

(Before) example:  `Sel : 25 | 4` meaning 25 characters selected on 4 lines

I don't think the multi-selection (after) example (which adds the `5` to this, for 5 selections) really boggles anyone's mind.  :-)

But yea, @Ekopalypse , I think you're right, with:  

>most users will understand when they have made a few selections.

Thanks for giving it a try.

sasumner:
Reopening after positive tester feedback.
Thank you to testers.

sasumner:
So just a bit of weirdness from this but nothing to get anxious about.:

STR:

Enable multi-editing.
Make a column selection covering 4 lines and 3 columns.
No weirdness:  statusbar shows `Sel : 4x3 = 12` as intended.
Hold Ctrl and drag the mouse over 5 characters of a word on another line.
**_Observe weirdness_**: statusbar shows `Sel 5 : 17 | 5`
Apparently the 4x3 column block has been converted to 4 selections of 3 characters each, on 4 lines.  Adding one selection of 5 chars on another line and the data on the statusbar makes sense.
I've always found it odd that Scintilla doesn't cancel the rectangular selection when user attempts to add a second (stream) selection via the "multi-editing" capability.  It certainly does cancel it if you try to add a second rectangular block by holding Ctrl+Alt while dragging.

Ekopalypse:
I checked this previously and I think it is correct as, from my
understanding, you cannot have different selections active.
Either rectangle or stream but not both.

Am Sa., 5. Sept. 2020 um 16:29 Uhr schrieb sasumner <
notifications@github.com>:

sasumner:
@Ekopalypse Yes, but it would feel more right if Scintilla would cancel the rectangular selection in this scenario, rather than "converting" it to multiple stream selections (if that is what it really is doing; I think so, but I didn't look hard at it).

Ekopalypse:
From logical point of I agree, but to be honest, I prefer the way it is doing right now as from workflow point of view it makes sense because otherwise you have to re-select what was already selected.

PR #8785, opened 2020-09-01
Optimized images files

tiagofilipesilva:
In the spirit of Notepad Plus Plus, these image files were optimized without quality loss, so they are processed faster and consume less memory.

Kudos to @donho for this amazing project.


tiagofilipesilva:
So… why was the PR closed?

Issue #8790, opened 2020-09-01
Add word-count in selected text to View > Summary output


sasumner:
### Description of the Issue

It was requested to add the count of words in selected text to the _View_ menu's _Summary..._ output data; meaning this:

![image](https://user-images.githubusercontent.com/30118311/91887491-d7cc7780-ec58-11ea-82ea-e4c3d7741e49.png)

As one can see, there's a total word count from the entire document in the `Words:` output section.  However, there isn't a words-in-selection output.

This behavior was requested as part of another issue, here:  https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1136#issuecomment-683906397

### Debug Information
Notepad++ v7.8.9   (32-bit)
Build time : Jul 15 2020 - 20:26:50
Path : C:\\....................\npp.7.8.9.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1339
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll

Issue #1136, opened 2015-11-12
Show word count on status bar


abdekker:
Using v6.3 here. Currently the status bar has a panel that says "length : 31003   lines : 507" (in my example). The number of bytes is interesting but in some circumstances not of interest, whereas the number of words might be. My suggestion is to be able to toggle between "file size" and "word count" in the status bar. The status bar would then appear as:
- "length : 31003   lines : 507" [if File size (default) is selected]
- "words: 4943  lines: 507" [if Word count is selected]

I realise there is the menu option View > Summary... gives some of this information, but that's a bit buried away and requires several clicks.

Great job on notepad++, love it!

denis-bogdanas:
I subscribe to this request.

joergneulist:
You can also access the Summary by double clicking on the status bar. Still, I would also love to see the word count without interacting!

abdekker:
@joergneulist: Thanks! That is new to me. The original feature request is still valid, but this is useful (hidden) feature.

mrspaceman:
This would be a great feature for me.

jamesford42:
+1

vbasarkar:
+1

stoptheinsanity:
I use the double click method, but for me, if I could just glance at the status bars and clearly see the word count it wouldn't break the flow of writing.

laocoi:
Vote for this feature

Aenadon:
Absolutely! Please do this!

bitseeker:
+1 from me. Would be very progressive in workflow.

igavronski:
+1 -- or even better, a customizable status bar would be great.

sasumner:
Possibly relevant:  https://notepad-plus-plus.org/community/topic/14889/footer-sel-to-include-word-count

a4jp-com:
Can I request the number of words selected get counted and displayed automatically in the bottom info bar? Everything else I wanted has been fixed and this program is wonderful. This would be very useful for students submitting homework that need to be certain lengths. As well as university students and SEO (Search Engine Optimization) workers like me that need to also find text throughout multiple files.

Like this: selected count/total count

sasumner:
Maybe I'm off base here but perhaps an idea on why this hasn't been implemented is that it could be "expensive" to calculate the number of words in a really large document.  The standard things currently shown on the status bar are all direct Scintilla calls to data items that are already maintained.

a4jp-com:
[...]

PR #8791, opened 2020-09-01
Add word-count in selected text to document Summary output

sasumner:
Implements #8790

Example:

image

No attempt is made to handle multiple or column-block selections.

If invoked with no selection, will show output of `0` ; this serves as a prompt to the user that the feature is available to count words in a single selection.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1339,6 +1339,7 @@ Find in all files except exe, obj &amp;&amp; log:
             <summary-filemodifytime value="Modified: "/>
             <summary-nbchar value="Characters (without line endings): "/>
             <summary-nbword value="Words: "/>
+            <summary-nbwordinsel value="Words (in 1 selection): "/>
             <summary-nbline value="Lines: "/>
             <summary-nbbyte value="Document length: "/>
             <summary-nbsel1 value=" selected characters ("/>
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3224,11 +3224,12 @@ size_t Notepad_plus::getSelectedBytes()
 	return result;
 }
 
-int Notepad_plus::wordCount()
+int Notepad_plus::wordCount(bool isEntireDoc)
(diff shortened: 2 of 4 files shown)


donho:
Why do we need it?

sasumner:
>Why do we need it?

It was a user request (which another user upvoted), after a full dynamically-updating word count on the status bar was denied (a denial which I agree with).  

It's a minor add, but if you want to reject it, user could count words in selection with a regular expression count operation.

donho:
IMO, if we want to add selected words information, we should add here:
image

In this case, the management of multi-selections should be considered.

donho:
@sasumner 
OTOH, 35 persons voted word count on status feature:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1136

Do you think word count on status will bring some performance issue?

donho:
Answer myself the question:
> Do you think word count on status will bring some performance issue?

There is.

sasumner:
@donho said:

>Do you think word count on status will bring some performance issue?

Yes, that's why I said earlier "a denial I agree with".  :-)

sasumner:
@donho said:

>if we want to add selected words information, we should add here: `18 selected characters (18 bytes) in 2 ranges

I was trying to keep the in-selection word-counting simple and confined to ONE selection.

It is more complicated to do multi-selection searching; that's why it still isn't supported for _In selection_ during _Find_ / _Count_ operations

How about:
`Words in document (selection): 1,234 (125)`

If more than one selection currently active, it shows existing behavior:
`Words: 1,234`

donho:
IMO selection info is not really the interesting info for users, but since we have already provided such info, we could add a complete info in this field. In the case that we cannot align with it, just don't provide it and we keep a cleaner (a little messed already) interface - too much info, no info. 
My role is find a balance between the application's rich features and its usability. Hope you understand.

sasumner:
@donho said:

>... Hope you understand.

No problem.  
I showed requesting user how to do a macro-based solution here --> https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1136#issuecomment-687779764
Is it as nice as this PR? No.
It seems OK though.

Issue #5142, opened 2018-12-14
No confirmation for Closing an emptied new file when an undo buffer is present


Davidebyzero:
### Description of the Issue
Notepad++ almost always asks for user confirmation when the user Closes a file that has been edited. However, if it's a new file (that does not exist as an actual file on disk), and some edits were done, followed by clearing everything, and then a Close is done, there is no confirmation, and the file is instantly closed, even though there actually was data in the Undo buffer that might have been important.

I consider this to be dangerous, as I have become accustomed to reliable Undo buffers and always being prompted before closing anything that has been edited. Having an exception to the rule like this is unexpected and could result in unintended loss of work.

Microsoft Visual Studio handles this in the way I prefer. On the other hand, Windows Notepad handles this in the same flawed way as Notepad++.

### Steps to Reproduce the Issue
1. Create a New file in Notepad++
2. Type something into the new file
3. Select All, then press Backspace or Delete
4. Close the file
5. Notepad++ instantly closes the file, without prompting for confirmation.

### Expected Behavior
When there is unsaved data in the Undo buffer, and the file has not been saved in its current edited state, Notepad++ should always ask for Close File confirmation, regardless of whether it is a named file on disk or a not-yet saved new file.

### Actual Behavior
In the scenario described in "Steps to Reproduce" above, Notepad++ discarded potentially valuable data in the Undo buffer without asking for user confirmation.

Contrast this with another situation: open or create a file in Notepad++, clear all its contents, and then save it to disk. Now, if the file is Closed, there should be no confirmation (and there isn't, which is correct), because it has already been saved (indicating there was actual user intent to erase the file's contents) and no edits have been subsequently done. But if edits were done after the clearing of its contents, a File Close should trigger a confirmation dialog (and it does, which is correct).

Also, if a named file on disk is being edited in Notepad++, and is saved as a zero-byte, empty file, and then subsequently edited to be non-empty (without being saved), then cleared of its contents (without being saved), a subsequent File Close prompts the user for confirmation. This is the correct behavior, and Notepad++ should continue doing the same thing in this scenario.

And finally, in the situation of creating a new file in Notepad++ (e.g. with File→New) and then closing it without having done any edits, it should be instantly closed with no confirmation. This is currently being done and is the correct behavior.

sasumner:
>Notepad++ discarded potentially valuable data in the Undo buffer without asking for user confirmation.

This is asking a bit much...

[...]

PR #8813, opened 2020-09-04
Prompt to save unnamed but modified empty tab

sasumner:
Implements #5142

Give user a final chance to reclaim data in undo buffer of a currently zero-length document, instead of automatically simply closing such a tab.

Mimics Visual Studio behavior in the same scenario.


donho:
If user wants to save somethings he/she should restore information then close it to save.
Close an empty (even dirty) document and ask user to save it - it doesn't seems a smart behaviour to me.
Such behaviour exists since (very) long time ago but I've never got any complain about it.
I won't integrate such behaviour, sorry.

PR #8822, opened 2020-09-08
Prevent line number margin width user-noticeable glitch during boundary cross

### LINKED ISSUE #5670: [Feature Request] Add a setting for choosing a custom width for the Line Number Margin, to avoid the annoyance of seeing the resizing process after line 999.
### Description of the Issue
From line 1 to 999 of a file, the space for the Line Number Margin is the same width. When it reaches 1000, the width increases, which causes my text to be shifted slightly. This is very annoying. Imagine scrolling up and down repeatedly near that point and seeing the shift multiple times. I assume that reaching the "10,000" point will cause another resize.

1. Scroll past line number 999.
2. The width of the line number will increase, causing text to shift rightward.

### Desired Behavior
1. An option called "Constant Line-Number Width".
2. If my custom length is "5", the width of the margin will be the same size from line number 1 to 99,999. It won't cause a resize until it reaches line number 100,000.
3. The lowest number that people are allowed to choose should not be lower than "3", because Notepad++ already shows lines "1 to 999" without resizing.

If the creators of this program know the absolute limit of the size of a text file that this program can process, then set that as the maximum width that can be chosen.
Example:
If the creators know that current computers cannot open a text file that has 10 million lines, they should set the limit of the "Custom Length" setting to 7, so it can show "9,999,999" lines without the width of the line-number margin resizing.

Putting millions of lines in one file may be ridiculous, even if computers can handle it, but putting 10,000 lines in one program is not too strange, so the "Constant Line-Number Width" should definitely be allowed to be set to "5".

sasumner:
Fixes #5670

As issue states, visual glitch occurs when scrolling through an area of the document where the line number changes from N digits to N+1 digits (first occurrence at line 999/1000, second occurrence at 9999/10000, etc.).

The existing code in `updateLineNumberWidth()` seems to do a bit much "thinking".
Is this really necessary?

Taking a cue from @xylographe 's https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5670#issuecomment-499312475 , this PR simplifies the logic by having a single-width line number margin for all lines in the document, corresponding to the size necessary for the largest line number in the document.  This has some advantages:

* solves the "glitching" issue
* faster execution for better performance (this code is called constantly as it is part of the PAINT process)
* simpler code

Tested with a document with several hundred million lines.

The disadvantage is that at small line numbers, the margin is wider than it truly needs to be.  In practice, even for the case of a doc with several hundred million lines, this (subjectively) is not really a big downside, except perhaps when one is evaluating a PR such as the current one. :-)

Note: PR also gives 4 more pixels to the width, by using an additive constant of 12 in the `pixelWidth` calculation (existing code uses a constant 8 pixels for this).

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -2717,38 +2717,17 @@ void ScintillaEditView::updateLineNumberWidth()
 {
 	if (_lineNumbersShown)
 	{
-		auto linesVisible = execute(SCI_LINESONSCREEN);
-		if (linesVisible)
+		auto lineCount = execute(SCI_GETLINECOUNT);
+		int numDigits = 0;
+		while (lineCount)
 		{
-			auto firstVisibleLineVis = execute(SCI_GETFIRSTVISIBLELINE);
-			auto lastVisibleLineVis = linesVisible + firstVisibleLineVis + 1;
-
-			if (execute(SCI_GETWRAPMODE) != SC_WRAP_NONE)
-			{
-				auto numLinesDoc = execute(SCI_GETLINECOUNT);
-				auto prevLineDoc = execute(SCI_DOCLINEFROMVISIBLE, firstVisibleLineVis);
-				for (auto i = firstVisibleLineVis + 1; i <= lastVisibleLineVis; ++i)
-				{
-					auto lineDoc = execute(SCI_DOCLINEFROMVISIBLE, i);
-					if (lineDoc == numLinesDoc)
-						break;
-					if (lineDoc == prevLineDoc)
-						lastVisibleLineVis++;
-					prevLineDoc = lineDoc;
-				}
-			}
-
-			auto lastVisibleLineDoc = execute(SCI_DOCLINEFROMVISIBLE, lastVisibleLineVis);
-			int i = 0;
-
-			while (lastVisibleLineDoc)
-			{
-				lastVisibleLineDoc /= 10;
-				++i;
-			}
-
-			i = max(i, 3);
-			auto pixelWidth = 8 + i * execute(SCI_TEXTWIDTH, STYLE_LINENUMBER, reinterpret_cast<LPARAM>("8"));
+			lineCount /= 10;
+			++numDigits;
+		}
+		numDigits = max(numDigits, 3);
+		auto pixelWidth = 12 + numDigits * execute(SCI_TEXTWIDTH, STYLE_LINENUMBER, reinterpret_cast<LPARAM>("8"));
+		if (execute(SCI_GETMARGINWIDTHN, _SC_MARGE_LINENUMBER) != pixelWidth)
+		{
 			execute(SCI_SETMARGINWIDTHN, _SC_MARGE_LINENUMBER, pixelWidth);
 		}
 	}
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -2726,10 +2726,7 @@ void ScintillaEditView::updateLineNumberWidth()
 		}
 		numDigits = max(numDigits, 3);
 		auto pixelWidth = 12 + numDigits * execute(SCI_TEXTWIDTH, STYLE_LINENUMBER, reinterpret_cast<LPARAM>("8"));
-		if (execute(SCI_GETMARGINWIDTHN, _SC_MARGE_LINENUMBER) != pixelWidth)
-		{
-			execute(SCI_SETMARGINWIDTHN, _SC_MARGE_LINENUMBER, pixelWidth);
-		}
+		execute(SCI_SETMARGINWIDTHN, _SC_MARGE_LINENUMBER, pixelWidth);
 	}
 }
 


Yaron10:
@sasumner,

:+1: 
This PR also solves some issues in Compare plugin. Thank you.

Allow me some questions:

The extra 4px padding are added *only to the left* (`8` to `12`; tested with "Bookmarks Margin" hidden).
Why is that needed? - Apparently it only increases the difference between the left and right padding. 

Would it be possible to center the numbers within the margin? 

Assuming that the execution of `SCI_GETMARGINWIDTHN` and `SCI_SETMARGINWIDTHN` take approximately the same time, wouldn't it better to remove the `if (execute(SCI_GETMARGINWIDTHN` line?
-- With the `if`, one command is always executed and in some cases two.
-- Without the `if`, one command is executed in all cases.

sasumner:
>The extra 4px padding...Why is that needed?

During testing I noticed that for some ranges of line numbers, the left side of the number was being chopped very slightly, so I wanted to give it a bit more room.  I don't recall the range; there is potentially a lot to examine when one "looks" at several hundred million lines.  :-)

>Would it be possible to center the numbers within the margin?

I don't believe this is currently reasonably achievable.

>Assuming that the execution of SCI_GETMARGINWIDTHN and SCI_SETMARGINWIDTHN take approximately the same time

I did not assume that a "get" takes the same amount of effort as a "set".  I would think that a "get" just grabs a internally stored number and returns it, whereas a "set" would do much more.
I did get into Scintilla's code for the "set" and it seemed a bit involved before it did its own "get".
But I will look at it (both) again.

Yaron10:
> there is potentially a lot to examine when one "looks" at several hundred million lines. 

Correct. :)
With an Image Editor you can see the exact number and location of pixels added.
`8` also adds padding to the left only, but I thought it might be justified in some cases due to the window border/edge proximity.
And, obviously, `12` also further increases the margin width.

> I did not assume that a "get" takes the same amount of effort as a "set". I would think that a "get" just grabs a internally stored number and returns it, whereas a "set" would do much more.

I also assumed there was a difference between "get" and "set".
But I wanted to "sharpen" the question.

> But I will look at it (both) again.

Thank you.

sasumner:
>With the if, one command is always executed and in some cases two.
>Without the if, one command is executed in all cases.

Before looking at code, just thinking about the above a bit more...

The "two" case only occurs right at the time when the lines-in-doc crosses a border (e.g. from <= 999 to >= 1000).
How often does this happen in practice? Not very often.
So I'd say that pretty much only "one" command is executed here.
But still, I will examine it.

sasumner:
There is not a real advantage to calling SCI_GETMARGINWIDTHN as SCI_SETMARGINWIDTHN will check for "no change" in requested width before doing anything of substance anyway, so I'll push new code that eliminates the "get".

Yaron10:
@sasumner,

Thank you. :+1: 

> But I will look at it (both) again.

I'm now realizing you meant both "get" and "set". I thought it was `if` and `12`.
Should I assume the change to `12` is necessary or you're still investigating it?

sasumner:
>Should I assume the change to 12 is necessary

"Necessary" is a bit subjective in this case.
I couldn't find the line range that I thought it was actually doing some chopping for; I could have been mistaken.

So I'll go forward with the opinion that the `+ 12` sample looks visually more appealing/balanced:
image

than the `+ 8` sample does:
image

Here's another `+ 12` sample, maybe at a better line number to look at:
image

But, really, the main effort here is to fix the bump/glitch, so to me the `+ 8` or `+ 12` is a "take your pick"...
...expecially since we might end up with this being a "patch" anyway, given my very recent history of suggested changes. ;-)

Yaron10:
> So I'll go forward with the opinion that the + 12 sample looks visually more appealing/balanced

Yes, in your screenshots the `+ 12` looks better.

> given my very recent history of suggested changes.

Come on! NPP recent versions' fixes and enhancements include so many of your PRs.

**"Add line count to hits output in search results"**: I've integrated it in my build. IMO It's certainly worth the extra code.

**"Fix no-selection sort including empty last line of document"**: I've rarely used the sorting functionalities in "real life" but I integrated that part too. I can see Don's point but I tend to agree with you.

**"Prompt to save unnamed but modified empty tab"**: 
**"Add word-count in selected text to document Summary output"**:
I can see your point but I tend to agree with Don.

So, in my list you have only 2 "rejects". That's quite good. :)

donho:
@sasumner 

The so called "visual glitch" is, in fact, a feature. It makes the margin take as less room as possible.
  
It seems less code so more performance, but when I was scrolling the document, I feel your implementation is less efficient than the original code.
In deed, here you use a loop to replace a if condition so I really don't think it's "faster execution for better performance", and it's even worse - as you said "this code is called constantly as it is part of the PAINT process", for each scroll line, it could loop from 1 to 5 or 6 times to get the number of digit.

I don't really feel interesting about this implementation

donho:
@Yaron10 
> This PR also solves some issues in Compare plugin. Thank you.

Could you elaborate the problem of Compare plugin?

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -2717,40 +2717,16 @@ void ScintillaEditView::updateLineNumberWidth()
 {
 	if (_lineNumbersShown)
 	{
-		auto linesVisible = execute(SCI_LINESONSCREEN);
-		if (linesVisible)
+		auto lineCount = execute(SCI_GETLINECOUNT);
+		int numDigits = 0;
+		while (lineCount)
```
donho, line 2722: I would do the following code:
```
		if (lineCount >= 1 && lineCount <= 9) numDigits = 1;
		else if (lineCount >= 10 && lineCount <= 99) numDigits = 2;
		else if (lineCount >= 100 && lineCount <= 999) numDigits = 3;
		else if (lineCount >= 1000 && lineCount <= 9999) numDigits = 4;
		else if (lineCount >= 10000 && lineCount <= 99999) numDigits = 5;
		else if (lineCount >= 100000 && lineCount <= 999999) numDigits = 6;
		else
		{
			while (lineCount)
			{
				lineCount /= 10;
				++numDigits;
			}
		}
```
From 1 to 6 digits will catch 90% of cases - so no loop on scrolling in these cases.

donho:
@sasumner Hmm...
> It seems less code so more performance, but when I was scrolling the document, I feel your implementation is less efficient than the original code.

My bad. I just saw that I have the same code:
```

			while (lastVisibleLineDoc)
			{
				lastVisibleLineDoc /= 10;
				++i;
			}
```
I will optimize it as I suggested. Then let's see later if we remove this feature later.

Yaron10:
@donho,

During the past few years @pnedev has used some creative workarounds for issues caused by the changing margin width.

Here is an open issue.

Please use the latest **ComparePlus** development version.
https://ci.appveyor.com/project/pnedev/compare-plugin/builds/34184726
(NOTE: the x64 version is not yet fully compatible with the current SciLexer.dll).

https://github.com/pnedev/compare-plugin/issues/244

Thank you for looking into it. :+1:

sasumner:
@donho 

>The so called "visual glitch" is, in fact, a feature. It makes the margin take as less room as possible.

This has always been understood.
My opinion is that a wider margin is an acceptable trade to avoid the visual glitch.

donho:
@sasumner 
I prefer to preserve this feature - but there are some demands for this so let's make it optional (dynamic by default).
And since some plugins need a constant margin width (as Compare plugin does that @Yaron10 mentionned), a plugin API `NPPM_DISABLEDYNAMICLINEMARGIN (BOOL makeItDisableed)` will be necessary as well.

Regarding GUI, in Editing section of Preferences is quite crowded - OTOH, I think it's better to group `Display line number` option. What's your suggestion ?

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
Yaron10: That's clever. :+1: 

Just curious: why not `if (lineCount < 10) numDigits = 1;` and so on?

Yaron10:
@donho,

> I prefer to preserve this feature - but there are some demands for this so let's make it optional (dynamic by default).
And since some plugins need a constant margin width (as Compare plugin does that @Yaron10 mentionned), a plugin API NPPM_DISABLEDYNAMICLINEMARGIN (BOOL makeItDisableed) will be necessary as well.

The dynamic width is certainly better in many cases.
Thank you for this solution. :+1: 

@sasumner,

Thanks again.
I'm starting to think I've been jinxing you. :)

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
Yaron10: Actually we can start with `if (lineCount < 1000) numDigits = 3;` and remove the `max()` later.

donho:
@Yaron10 @sasumner 
Sorry I was not clear. I didn't reject the idea - this PR could be elaborated to make the current behaviour optional: dynamic & less dynamic.
For that we have to add an option in Preferences and 1 API call. So users can set what they want via UI, and plugins can use call to disable dynamic display.

donho:
And sorry for the conflict - I have fixed a bug in the current behaviour by removing the obsolete code.

Yaron10:
@donho,

You were quite clear and I understood you correctly.
I just wrote that the dynamic width could be useful too, and therefore the pref was a good solution.

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
donho: @Yaron10
> why not if (lineCount < 10) numDigits = 1; and so on?

You're right about it. Just tune up and commited (for the dynamic display part):
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/8e73f5d116b159485bb1b4f95e8f9904efd27632

Yaron10: :+1:

donho: > Actually we can start with if (lineCount < 1000) numDigits = 3; and remove the max() later.

Good thinking. I guess it's optimized maximum:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d82aba7b0ca40470b9cdeef624aa35bdf6376b42

Yaron10: Great. :+1: 

I wouldn't have thought of removing `include "tchar.h"`. :)

Yaron10:
@sasumner,

> With the OnSwitchFile and OnModification events it might be possible to implement a custom scheme.
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5670#issuecomment-506703577

This indeed seems better than calling `updateLineNumberWidth()` on `SCN_PAINTED` (if the width is not dynamic).
What do you think?

Thank you.

sasumner:
@Yaron10 Performance enhancement wasn't really my goal with this; removing the glitch was.  Of course, I wasn't opposed to removing some code that didn't seem useful, as long as it was within the `updateLineNumberWidth()` function.  But rearchitecture outside the function is probably best left to Don; if someone else does it, Don probably gets nervous that something subtle hasn't been broken.  Perhaps where things have gone, the whole thing might be best left in Don's care as well, not sure at this point.

Yaron10:
@sasumner,

I thought you closed this PR because you wanted to open a new clean one.
Whatever you decide - thanks again for work.

Yaron10:
I thought you closed this PR because you wanted to open a new clean one.
Whatever you decide - thanks again for your work.

sasumner:
>I thought you closed this PR because you wanted to open a new clean one.

Well, mainly I closed it because commits were made external to it that would make continued work in this one fairly conflicting.  :-)

sasumner:
@donho said:

>Regarding GUI, in Editing section of Preferences is quite crowded - OTOH, I think it's better to group Display line number option. What's your suggestion ?

I'd suggest:

image

Ticking the box would enable the NEW functionality; leaving it unticked would retain the EXISTING functionality.

sasumner:
@donho

How does the "plugin interface" work with this?
If a plugin calls to "disable" this, what is it disabling?  The existing behavior?  And thus the new behavior comes into effect?
What if two plugins want it set the opposite way?

donho:
@sasumner 
> I'd suggest:
>
> Ticking the box would enable the NEW functionality (from this issue).
> Leaving it unticked would retain the EXISTING functionality.

Before doing it, I would like to clean the **Editing** page: Move out **Folder Margin Style**, **Border** and **Vertical Edge Settings**, put them into a new section **Margin/Border/Edge**. What do you think?

sasumner:
@donho said

>Move out Folder Margin Style, Border and Vertical Edge Settings, put them into a new section Margin/Border/Edge. What do you think?

I already had a start on adjusting the existing subpage so that we can get this one setting in.
How about I continue with that for now (for this issue/PR)?

And for the future I'll do the "new section Margin/Border/Edge"
I'll make an issue and do a PR for that separately?

donho:
@sasumner 
> If a plugin calls to "disable" this, what is it disabling? The existing behavior? And thus the new behavior comes into effect?

I would say: while `VOID NPPM_SETLINECOUNTBASEDMARGINWIDTH(0, BOOL val2set)` is called by a plugin, if val2set is TRUE, then the new behaviour of line number magin will be enabled, otherwise staying in old behaviour.

> What if two plugins want it set the opposite way?

Then the last one will win. There's no way to have a consensuses in Plugins system unfortunately.

donho:
> And for the future I'll do the "new section Margin/Border/Edge"
I'll make an issue and do a PR for that separately?

Yes, make a PR separately please.

Issue #8828, opened 2020-09-08
Off-selection right-click should not cancel selection in Find-result window


sasumner:
An off-selection right-click should not cancel selection in a _Find result_ (or derivative) window, when Notepad++ is so configured in the Preferences.

### Description of the Issue
Reference #8563 and associated PR #8564 which affected the main editing view(s).
The commit from the PR created the following Preference setting:
![image](https://user-images.githubusercontent.com/30118311/92496467-27b1be00-f1c6-11ea-8eac-62ce08906719.png)
The behaviour controlled by this setting should also occur in the _Find result_ window.

donho:
@sasumner 
> The behaviour controlled by this setting should also occur in the Find result window.

Just tested, the behaviour controlled by this setting occurs also in the Find result window (without applying https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8829).

sasumner:
@donho 

>Isn't it the case for you ?

Hmm.  Well, testing now, it seems to work without applying the PR.
Not sure where I noticed it happen in order to create this issue + PR.
Since it is working, I'll close this issue and the PR.

PR #8829, opened 2020-09-08
Off-selection rclick in find-result windows should follow preference

sasumner:
Implements #8828

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -27,6 +27,7 @@
 
 #include <shlobj.h>
 #include <uxtheme.h>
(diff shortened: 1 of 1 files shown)


donho:
@sasumner 
If I do understand, this PR makes Find result windows has the same behaviour as Main edit areas' ones, according `Keep selection when right-click outside of selection` is checked or not ?

sasumner:
@donho 

>this PR makes Find result windows has the same behaviour as Main edit areas' ones

Yes, it was something I didn't consider originally.
It makes the behavior consistent.

donho:
Find result windows uses also class `ScintillaEditView` so what you have done in ScintillaEditView's procedure is kept. I don't think we need to do somethings more to keep Find windows aligns with main edit zones.

sasumner:
Closing due to not being needed.

Issue #6095, opened 2019-08-26
Marked text needs to be copyable


sasumner:
### Description of the Issue
When I use the `Mark...` feature to mark some large amount of text, often I want to copy only that marked text to a different place.  Currently there is no easy way to do that.

### Steps to Reproduce the Issue

1. Create the following text in a new Notepad++ editing view tab:

    A baseball team from the fictional town of "Mudville" (implied to be the home
    team) is losing by two runs in its last inning.  Both the team and its fans
    (a crowd of 5,000, according to the poem) believe they can win if Casey,
    Mudville's star player, gets to bat.  However, Casey is scheduled to be the
    fifth batter of the inning, and the first two batters (Cooney and Barrows)
    fail to get on base.  The next two batters (Flynn and Jimmy Blake) are
    perceived to be weak hitters with little chance of reaching base to allow
    Casey a chance to bat.  Surprisingly, Flynn hits a single, and Blake follows
    with a double that allows Flynn to reach third base.  Both runners are now in
    scoring position and Casey represents the potential winning run.  Casey is so
    sure of his abilities that he does not swing at the first two pitches, both
    called strikes.  On the last pitch, the overconfident Casey strikes out
    swinging, ending the game and sending the crowd home unhappy.

2. Invoke the `Search` > `Mark...` function.
3. In the `Find what` box enter `(?-s)\bb.*?\b`
4. Tick the `Wrap around` box.
5. Select `Regular expression` for `Search mode`.
6. All other options _unticked_.
7. Press the `Mark All` button.
8. Observe text marked like so:

![image](https://user-images.githubusercontent.com/30118311/63689730-c5636680-c7d8-11e9-9a4b-d0d06c6e83c1.png)

### Expected Behavior
I expected to find a way to copy my marked text to the clipboard.

### Actual Behavior
No easy way to copy marked text to the clipboard for use elsewhere.

### Debug Information

Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\.........\npp.7.7.1.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : ColumnTools.dll Explorer.dll LocationNavigate.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll PythonScript.dll

sasumner:
[...]

PR #8867, opened 2020-09-17
Add ability to copy marked text to the clipboard

sasumner:
Implements #6095

This one may be a bit controversial (perhaps why I haven't finished it before now), but the need keeps coming up on the Community site.
Basically, it is a user need to copy (red)marked text to the clipboard.
It really is only useful after doing regular-expression based marking via the _Mark_ dialog.

So here is where it would appear in the menus (is there a better place?):

image

Here's an example of the paste result after running the _Marked Text to Clipboard_, after marking as indicated by the search shown:

image

When the search matches run across line-endings, some sort of delimiter is needed to show where one match ends and another begins.  I chose `----` for this in the output, but it could be something else if that is not liked:

image


guy038:
Hi, @sasumner, and **All**,

First, **many thanks** for your implication in the improvement of N++ and, especially, for everything related to the "Find" window tabs. You are going to become "The" specialist of these 4 tabs ;-))

Your idea of adding the possibility of copying **all marked** occurrences of a search in the **clipboard** is really valuable. And the future location of this feature ( **`Edit > Copy to Clipboard > Marked text to Clipboard`** is quite **logical**, too.

I've thought of an **other** location : **Search > Bookmark > Copy Bookmarked Text`**. Note that this **new** syntax is different from the existent **Search > Bookmark > Copy Bookmarked Lines`**

In this way, this new option could be **easily** reached by **right**-clicking on the **Bookmark** margin, that you lately implemented ;-))

What do you think of ? May be, code or structure **problems** would arise, choosing my solution ?

As usual, **Scott**, just a suggestion !

Best Regards,

guy038

guy038:
Hello, @sasumner, and **All**,

As, apparently, I do **not** have, anymore, the possiblity of editing a **previous** post, here is an **updated** version that I hope will be more readable !

---

First, **many thanks** for your implication in the improvement of N++ and, especially, for everything related to the **"Find"** window tabs. You are going to become "The" specialist of these **4** tabs ;-))

Your idea of adding the possibility of copying **all marked** occurrences of a search, in the **clipboard**, is really valuable. And the **future** location of this feature ( **`Edit > Copy to Clipboard > Marked text to Clipboard`** ) is quite **logical**, too.

I've thought of an **other** location : **`Search > Bookmark > Copy Bookmarked Text`**. Note that this **new** syntax is different from the **existent** **`Search > Bookmark > Copy Bookmarked Lines`** option !

In this way, this **new** option could be **easily** reached by **right**-clicking on the **Bookmark** margin, that you lately implemented ;-))

What do you think of ? May be, code or structure **problems** would arise, choosing my solution ?

As usual, **Scott**, just a suggestion !

Best Regards,

guy038

sasumner:
>many thanks ... especially, for everything related to the "Find" window tabs. You are going to become "The" specialist of these 4 tabs

It DOES seem like many of my efforts are concentrated here!

BTW, did you notice a recent change I did allows a user to reuse the same Alt+letter keyboard accelerators in the four tabs?  So, you could have Alt+r do replace in the _Replace_ tab, and Alt+r do replace-in-files in the _Find in Files_ tab (although that is not a great example!).  Until this change, I was using your long ago remapping so that every function in the 4 tabs had a unique accelerator -- remember that one? :-)

>this new option could be easily reached by right-clicking on the Bookmark margin

I like it.  I will add it.  I suppose there is no reason that it shouldn't appear in two places in the UI.

sasumner:
>I like it. I will add it. I suppose there is no reason that it shouldn't appear in two places in the UI.

I thought some more about it.
I will add it as an enhancement after this PR is merged (if it is).

donho:
`Edit->Copy to Clipboard` is a quite logical choice. However, why not a button "Marked Text to Clipboard" just under "Mark All" in Find dialog's Mark section? It will more easy to use and understand.

sasumner:
>why not a button "Marked Text to Clipboard" just under "Mark All" in Find dialog's Mark section?

Sure, we can do that.  That text is kind of long for a button; could be multiline I suppose...

donho (changes requested):
1. Add a button "Marked text to Clipboard".
2. Add a generic méthod `markedToClipboard(int indiStyle)` for the future commands "Marked Style N text to Clipboard".

sasumner:
I shortened the text for the button a little bit so that it doesn't need to be multiline.
I tried it as a multiline button, but the reality of it was that it was so big on the screen that way that it looked like a very important feature (really this is a minor thing).
I left the menu entry in the _Edit_ menu because that way the command is macroable without having to be supported in the complicated way that other find-related macros get recorded.

PowerEditor/src/Notepad_plus.rc
```diff
@@ -283,6 +283,8 @@ BEGIN
                 MENUITEM "Current Full File path to Clipboard",    IDM_EDIT_FULLPATHTOCLIP
                 MENUITEM "Current Filename to Clipboard",          IDM_EDIT_FILENAMETOCLIP
                 MENUITEM "Current Dir. Path to Clipboard",         IDM_EDIT_CURRENTDIRTOCLIP
+                MENUITEM SEPARATOR
+                MENUITEM "Marked Text to Clipboard",               IDM_EDIT_MARKEDTOCLIP
```
donho, line 287: IMO, it's not necessary - that's why the existence of the button.

sasumner: Did you see my comment https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8867#issuecomment-700976991 about macro-ability?

Do we leave it unable to be recorded in a macro?

Or do we give it this type of treatment: https://npp-user-manual.org/docs/searching/#searching-actions-when-recorded-as-macros
This type seems like a lot of effort for such a simple command.

donho: OK, it's not so bad to have menu command.
We'll have "Marked Style N text to Clipboard" (N:1..5) &  "All Marked text to Clipboard" 6 commands:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8943
We may make sure the name "Marked Text to Clipboard" is not ambiguous among these commands.

donho: BTW, feel free to assign to yourself if you're interested in doing https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8943

PowerEditor/src/ScitillaComponent/FindReplaceDlg.rc
```diff
@@ -80,6 +80,7 @@ BEGIN
     PUSHBUTTON      "Replace All in All Opened Doc&uments",IDC_REPLACE_OPENEDFILES,268,74,91,21,BS_MULTILINE
     PUSHBUTTON      "Find All",IDD_FINDINFILES_FIND_BUTTON,268,20,91,14,WS_GROUP
     PUSHBUTTON      "Replace in Files",IDD_FINDINFILES_REPLACEINFILES,268,38,91,14
+    PUSHBUTTON      "Copy Marked Text",IDC_COPY_MARKED_TEXT,268,56,91,14
```
donho, line 83: Nice one :)

PowerEditor/src/Notepad_plus.rc
sasumner: OK, I will do it, once this one is committed; that way there shouldn't be any conflicts to resolve around MENUITEM in Notepad_plus.rc.

sasumner:
@guy038 said:

>I've thought of an other location : Search > Bookmark > Copy Bookmarked Text.
>In this way, this new option could be easily reached by right-clicking on the Bookmark margin, that you lately implemented

I think this "need" goes away with @donho 's suggestion of adding a button to the _Mark_ window, so I won't do it. :-)

PowerEditor/src/Notepad_plus.rc
sasumner: >make sure the name "Marked Text to Clipboard" is not ambiguous 

Like the existing _Search_ > _Jump up_ > ... commands, we can use something like _Copy Style #**_N_** Text to Clipboard_

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -3390,6 +3404,48 @@ bool FindReplaceDlg::replaceInOpenDocsConfirmCheck(void)
 	return confirmed;
 }
 
+void FindReplaceDlg::markedTextToClipboard(int indiStyle)
```
donho, line 3407: According https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8943 ,  `markedTextToClipboard` method shouldn't belong to `FindReplaceDlg`

donho (changes requested):
Find a suitable class for `markedTextToClipboard` method.

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
sasumner: Haha. Future issues affect current issues.

donho: > Haha. Future issues affect current issues.

Not really, markedTextToClipboard has been made more generic, so it should be in a place more generic :)

sasumner: Probably to do it well, we should cancel this PR, combine the two issues into one, and have a cohesive approach.
I'm usually in favor of separating things, but in this case maybe together is better.
If you agree, close this PR?

donho: Do you prefer to move markedTextToClipboard in the PR for #8943 ?

sasumner: Short answer: Yes. :-)

Issue #4479, opened 2018-05-07
When case insensitive sorting will be added in Notepad++ x64 (without TextFX)?


PiotrMP006:
When case insensitive sorting will be added in Notepad++ x64 (without TextFX)?

insensitive sorting vs sensitive sorting

A A
a B
B C
b a
C b
c c

sasumner:
Related Community posting:
https://community.notepad-plus-plus.org/topic/19659/sort-case-insensitive-on-64-bit

Corey-Keller:
NOTE: This is just some armchair code ideas from someone who hasn't done ANYTHING C(++) in 10+ years, so it may be completely invalid. If so, please let me know.

[Here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0bf3eea08ae83c9abcfbef722c51a994b3800bac/PowerEditor/src/MISC/Common/Sorters.h#L80) is the code that is used for lexicographic sorting. Couldn't you either:

1. Just make a new duplicate class ("LexicographicSorterCaseInsensitive") 
2. add a parameter (bool isCaseSensitive, like there already is for ascending/descending)

where the comparison function contents changes `return a.compare(b) < 0;` to `return toupper(a).compare(toupper(b)) < 0;`?

sasumner:
@Corey-Keller 
It's a bit more involved than that in the Unicode world, but it's certainly doable of course.
I'll be working on it.

Corey-Keller:
> 
> 
> @Corey-Keller
> It's a bit more involved than that in the Unicode world, but it's certainly doable of course.
> I'll be working on it.

Right. After posting I saw [this](https://qr.ae/TW8wnX) that points out just how difficult it can be with non-English languages, much less all of the extra issues dealing with the full Unicode 13 set.

deus-e-x:
I think it would be sufficient to support case insensitive sorting of the English character set only.

I just checked with NppTextFX, it sorts lines containing special foreign characters like ‘ß’ or ‘Ф’ at the top of a sort result. I.e. NppTextFX does not try to sort such lines but places them at a prominent place for the user to easily get aware.

sasumner:
@donho Does implementing this leave us open to everyone wanting/expecting something different out of it?  And a lot of future "issues" created when it doesn't do what individual users expect?

I'm not opposed to putting effort into it, but I'd like to have an "approved" direction on it before doing so.

[...]

PR #8869, opened 2020-09-18
Add case insensitive sorting

sasumner:
Implements #4479


PowerEditor/src/MISC/Common/Sorters.h
```diff
@@ -130,6 +130,49 @@ class LexicographicSorter : public ISorter
...
 
+// Implementation of lexicographic sorting of lines, ignoring character casing
+class LexicographicCaseInsensitiveSorter : public ISorter
+{
+public:
+	LexicographicCaseInsensitiveSorter(bool isDescending, size_t fromColumn, size_t toColumn) : ISorter(isDescending, fromColumn, toColumn) { };
+
+	std::vector<generic_string> sort(std::vector<generic_string> lines) override
+	{
+		// Note that both branches here are equivalent in the sense that they give always give the same answer.
```
chcg, line 141: "they give always give" seems to be one give to much

sasumner: If you search the code for those four words BEFORE this PR, you will find them.
Thus, this code is blameless for that, as it is a copy+paste. :-)
But, I will fix both.

sasumner:
@donho I was expecting more conversation about this rather than acceptance with no changes. :-)  We will see what the users think of it.

donho:
@sasumner 
What kind of point are we supposed to talk about ?

sasumner:
@donho 

Nothing specific.
Maybe I am just too worried about this one. :-)

Issue #8068, opened 2020-03-25
Accessibility: tabbar saved vs edited vs readonly


pryrt:
### Description of the Issue

Please: the icons and/or the names on the tabbar need to better indicate saved vs edited vs readonly state.  

Please do **not** reject this out-of-hand.  It's not just aesthetics (though that could be improved) or personal preference; it **needs** to be done.  As mentioned in https://community.notepad-plus-plus.org/topic/19110/different-indicator-for-unsaved-tabs, certain varieties of color blindness make the saved/blue vs unsaved/red icon indistinguishable.  

At the very minimum, the tabbar needs to use the *, just like the title bar use.  But that's the least preferred.  The best solution would be changing the icon on each, to three distinct and easily-recognizable icons.  

I know there's #5666 still open, but this takes it farther.
I know PR #5890 was rejected because the orange color looked weird.
I know PR #5097 was rejected because "it looked different"

Instead of rejecting outright, please work with the Community to find an acceptable solution, because this is an issue of accessibility.

The point is to make them all look as different as possible, to make it easy for everyone (colorblind or not) to quickly recognize the difference between saved, edited, and readonly.  Combing different icons with different colors is the best combination, which is better for everyone.

Doing something like Blue Disk (or Green Disk, or Green Checkmark if you are bold enough for not making tradition dictate an out-of-date icon -- who uses 3.5" floppies?!) for saved, Red Lock (or yellow or orange or whatever color doesn't look weird) for readonly, and Red Pencil for edited.

This would make it easier to distinguish, at a glance, for everyone.

If you don't like the exact suggestions, work with us to find alternatives that _do_ work.  I bet @SinghRajenM  would be willing to implement whichever suggestion gets consensus and do a new PR, or someone could at least start from those PRs as a basis, and tweak it to match what's agreed upon.

If changing the icon is absolutely unacceptable (which seems strange, since the icon changes to a creepy eyeball when the file is in "monitoring (tail -f)" mode), then at the very minimum, implement three different states in the text of the  tabbar: maybe `*` or `(*)` for edited and `(x)` or `(ro)` for readonly; or, if Unicode characters are allowed, 🔒 U+1F512 LOCK for readonly and ✎ U+270E LOWER RIGHT PENCIL (or one of the other pencil characters) for edited.  But really, icons with different shapes is the best solution for everyone.

sasumner:
Upon first read, I have to give this a +1 ... or a +10, +100, ...  

[...]

PR #8871, opened 2020-09-19
Add alternative icon set for tab bar

sasumner:
Implements #8068

A decision was made to implement this as requiring a restart of N++ for it to take effect.
This was because there are several places (tab bar, doc switcher, ctrl+tab) where these icons are used, and making the change on-the-fly dynamic would involve a lot of rewriting.
It certainly could be done without requiring restart, but this is likely a one-time change for the user anyway, so IMO having to restart is not a big problem.


sasumner:
image

image

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -230,9 +234,12 @@ LRESULT Notepad_plus::init(HWND hwnd)
     const ScintillaViewParams & svp1 = nppParam.getSVP();
 
 	int tabBarStatus = nppGUI._tabStatus;
+	
 	_toReduceTabBar = ((tabBarStatus & TAB_REDUCE) != 0);
-	int iconDpiDynamicalSize = nppParam._dpiManager.scaleY(_toReduceTabBar?13:20);
-	_docTabIconList.create(iconDpiDynamicalSize, _pPublicInterface->getHinst(), docTabIconIDs, sizeof(docTabIconIDs)/sizeof(int));
+	int iconDpiDynamicalSize = nppParam._dpiManager.scaleY(_toReduceTabBar ? 12 : 18);
```
donho, line 239: Is any reason for changing the scale size?

PowerEditor/src/Notepad_plus.rc
```diff
@@ -116,8 +116,11 @@ IDI_UNDO_DISABLE_ICON           ICON      "icons/undo_dis.ico"
 IDI_REDO_DISABLE_ICON           ICON      "icons/redo_dis.ico"
 //
 IDI_SAVED_ICON                  ICON      "icons/saved.ico"
+IDI_SAVED1_ICON                 ICON      "icons/saved1.ico"
```
donho, line 119: `saved2.ico` for me is more suitable:
"Shreck 2" came after "Shreck", but there's no "Shreck 1".

donho (changes requested):
Change icon name suffix from 1 to 2.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -56,7 +56,11 @@ enum tb_stat {tb_saved, tb_unsaved, tb_ro};
 #define DIR_LEFT true
 #define DIR_RIGHT false
 
-int docTabIconIDs[] = {IDI_SAVED_ICON, IDI_UNSAVED_ICON, IDI_READONLY_ICON, IDI_MONITORING_ICON};
+int docTabIconIDs[][4] =
+{
+	{ IDI_SAVED_ICON,  IDI_UNSAVED_ICON,  IDI_READONLY_ICON,  IDI_MONITORING_ICON },
+	{ IDI_SAVED1_ICON, IDI_UNSAVED1_ICON, IDI_READONLY1_ICON, IDI_MONITORING_ICON },
```
donho, line 62: Same here, change the suffix from 1 to 2

PowerEditor/src/Notepad_plus.rc
sasumner: It originated from the possibility of having in the future, possibly:

* Select Alternate icon set 1 <-- this current one
* Select Alternate icon set 2 <-- possible future set

But yes, no problem, I can change this.

No `c` in `Shrek` :-)

PowerEditor/src/Notepad_plus.cpp
sasumner: Yes

PowerEditor/src/Notepad_plus.cpp
sasumner: @donho I changed it because of discoveries we discussed here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8068#issuecomment-608635130

sasumner:
@donho 

>It's OK, I will modify them myself

Sorry, I would have done it but yesterday was a busy day. :-(
I'm glad you liked the rest of the PR.

---------------------

How would you feel about an experiment:
Set this new icon set option as _enabled by default_.

Rationale:
* It gives N++ a fresh new look. :-)
* Users that might like it probably won't know that it exists.  Most don't troll through the _Preferences_, they just use the product.
* I want to see how many people simply hate any kind of change, even a change for the better! (We'd know this by count of tech support questions which are easily answered -- untick the preference)

donho:
> Sorry, I would have done it but yesterday was a busy day. :-(

No problem at all. I was playing and modifying your PR, then I told myself that it's stupid to make you do another commit, whereas it's minor and I have had the result I want. 

> Set this new icon set option as enabled by default.

I prefer to keep the current ones as default. 
> It gives N++ a fresh new look. :-)

Well, personally, I prefer the old ones. The new ones are not pretty but good for telling difference while dark background is using.

> Users that might like it probably won't know that it exists. Most don't troll through the Preferences, they just use the product.

We will make the document for that.

> I want to see how many people simply hate any kind of change, even a change for the better! (We'd know this by count of tech support questions which are easily answered -- untick the preference)

Users are timid animals, the least change will make them run away.

sasumner:
>I prefer to keep the current ones as default.

Sure.

>new ones are not pretty 

I think they are fairly good.
They come from the icon set that N++ has historically used.

>good for telling difference while dark background 

You mean active versus inactive tab dark background.
Yes, I agree.
They will definitely help color-impaired users tell the difference in any form (active/inactive, large/small, etc)
Plus readonly icon won't cause all users problems on inactive tab.

>Users are timid animals, the least change will make them run away.

LOL.
Well, they won't run away (from the product) because the product is good.
But they do like to complain. :-)
I thought it would be interesting to see the depths of that complaint.

Well, (as a user) I'm switching to this icon set permanently for myself.
It is too bad that I can't convince the "old man" in you to let go of the "floppy disk" and make the change yourself. :-)

mere-human:
How about adding an option for choosing the icon set during the installation?
I got here from the discussion [there ](https://community.notepad-plus-plus.org/topic/20639/newbie-here-a-few-quick-questions)

Otherwise, the new users won't even know this opportunity exists.
Only if they are curious enough to get to the Community forum.

It could be done in a similar way there was an option for "Use the old, obsolete and monstrous icon".
Maybe, even with image previews to see a difference.

Issue #8833, opened 2020-09-09
Add a separate style for Bookmarks Margin


Yaron10:
**STR:**
Enable `Lines Number Margin` and `Bookmarks Margin`.

**Result:**
There's no clear border between the two margins.
The user might want to add a bookmark but actually click in the `Lines Margin`. And vice-versa.

**Suggested:**
Add a separate style for `Bookmarks Margin`.

***
Notepad++ v7.8.9   (32-bit)
Build time : Sep  5 2020 - 02:32:39
Path : C:\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

sasumner:
@Yaron10 

What I do, with Pythonscript, is colorize my bookmarks margin:

![image](https://user-images.githubusercontent.com/30118311/92593233-c42f9b00-f26e-11ea-8832-2b20c1db5d2e.png)

But, yes, you are right, and this has long bothered me, even before I colored it via script.

Yaron10:
@sasumner,
:+1:

Yaron10:
@sasumner,

When you find some time for this. :)

In `ScintillaEditView.cpp` before `COLORREF foldMarginColor = grey;`.

Add
```
	COLORREF bookmarksMarginBG = grey;
	i = stylers.getStylerIndexByName(TEXT("Bookmarks Margin"));
	if (i != -1)
	{
		Style & style = stylers.getStyler(i);
		bookmarksMarginBG = style._bgColor;
	}
	execute(SCI_SETMARGINTYPEN, _SC_MARGE_SYBOLE, SC_MARGIN_COLOUR);
	execute(SCI_SETMARGINBACKN, _SC_MARGE_SYBOLE, bookmarksMarginBG);
```

In `stylers` after `<WidgetStyle name="Line number margin"`.

Add
```
<WidgetStyle name="Bookmarks Margin" styleID="0" bgColor="EBEBEB" fontStyle="0" />
```

or one of the following.
```
        <!-- <WidgetStyle name="Bookmarks Margin" styleID="0" bgColor="E2E2EF" /> -->
[...]

PR #8874, opened 2020-09-20
Allow bookmark margin color setting

sasumner:
Implements #8833


Yaron10:
@sasumner,

:+1: 
Thank you.

PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
```diff
@@ -2633,6 +2633,16 @@ void ScintillaEditView::performGlobalStyles()
 	execute(SCI_SETFOLDMARGINCOLOUR, true, foldMarginColor);
 	execute(SCI_SETFOLDMARGINHICOLOUR, true, foldMarginHiColor);
 
+	COLORREF bookmarkMarginColor = veryLiteGrey;
```
donho, line 2636: discrete & elegant. Nice choice!

sasumner: Looking at it again 2 weeks later, to me it looks a bit _too_ discreet.
But, at least now user can change it to make it more contrasting if they'd like. :-)

PR #8876, opened 2020-09-21
Enhance Sel info provided in main status bar

sasumner:
Implements #8524

Second try on this one; performance was an issue on the first try so hopefully it is better this time.

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -890,7 +890,14 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				}
 			}
 
-			updateStatusBar();
+			bool selectionIsChanged = (notification->updated & SC_UPDATE_SELECTION) != 0;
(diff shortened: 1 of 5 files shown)


donho:
I didn't try the binary yet, but just check your code - and I cannot tell where is enhanced for the performance regarding the PR of last time. Do you mind to elaborate it in 2 words?

sasumner:
Wow.  Two words only?  :-(

How about?:  **_TRIED IT_**

The performance complaint with the earlier PR #8780 was during scrolling of a "large" cpp or xml file.
With this new PR I tried this, and compared result to the same files using 7.8.9.
I did not notice a difference that I could quantify.

PowerEditor/src/NppNotification.cpp
```diff
@@ -890,7 +890,14 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				}
 			}
 
-			updateStatusBar();
+			bool selectionIsChanged = (notification->updated & SC_UPDATE_SELECTION) != 0;
```
donho, line 893: Well played!

PR #8883, opened 2020-09-22
Avoid unnecessary notifications while deactivating window

Uhf7:
This is an enhancement of PR #8475 which fixed (and still fixes) #8466.

It takes into account, that `WM_ACTIVATE` is not called only, when the window is being activated, but also, when the window is being **de**activated. In the latter case it is not necessary to do the activation actions.

This was pointed out to me by @Predelnik, who had a problem with the additional `SCN_UPDATEUI/SC_UPDATE_H_SCROLL` notification occurring when the window is being deactivated, see https://github.com/Predelnik/DSpellCheck/issues/221#issuecomment-696652074 for details.


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/1a37b64fe697ceb8c1aa4ce8683d1ea38dee5c8c

PR #8885, opened 2020-09-22
Add preference for save type of untitled normaltext files

### LINKED ISSUE #8866: New file default save as type
i would like the save as dialog to have Save as type: All types (`*.*`) as default for new files when normal text is the default language. that way i can type in whatever extension i want to save the file as, for example, test.html, without needing to select the dropdown menu. it's easier and quicker

sasumner:
Fixes #8866 and potentially others (see links in issue).

Adds this preference setting:

image

Note:  Default setting is **_UNCHECKED_**

It is a bit strange in the _New Document_ page, as the other settings on that page influence document creation as a tab (thus, early on) and this one influences when the new document is saved for the first time (thus, later).  But it seems OK to me.


donho:
It isn't more clear the following text:
"Set Save dialog file extension filter as "*.*" for Normal Text"
?

donho:
Also, I would put this option into "MISC." because it's not about the "New Document".

sasumner:
> Also, I would put this option into "MISC." because it's not about the "New Document".

It is kind-of about the _New Document_ in the sense that it concerns "first-save" of it, but in the true spirit of existing _New Document_ options, you're right.

I will move it.

donho:
> It is kind-of about the New Document in the sense that it concerns "first-save" of it, but in the true spirit of existing New Document options, you're right.

True. But it's concerning also Save As command for existing files.

> I will move it.

Thank you.

BTW, maybe it's even more clear:
"Set Save dialog file extension filter to `.*`  instead of `.txt` for Normal Text"

donho (changes requested):
1. Rename checkbox label as "Set Save dialog file extension filter to `.*` instead of `.txt` for Normal Text"
2. Move this option into MISC. (Move up "Enable Notepad++ auto-updater" and put it just below)

sasumner:
The requested changes make it burdensome to continue working with this PR.
I will close this PR, and submit a new one, because that is the easiest path.

PR #8898, opened 2020-09-24
Close Search Results Window with ESC

JJT319:
Resolves issue #2946 ESC now closes the focused Search Results Window


sasumner:
I was curious to see what it would do for the default case of "docked at bottom", along with the NppExec console open.  I'm happy to report that it closes only the search results tab and not also the NppExec console.  :-)

PR #8905, opened 2020-09-26
Fixed smart highlighting of cloned view

### LINKED ISSUE #7910: Smart Highlighting issue in a cloned document, when scrolling up/down the original document
### Description of the Issue :

When a document is **cloned** in the **secondary** view, and some occurrences of a word are **highlighted** in green, in the **primary** view, scrolling up/down the **original** document **only** highlights, in the **secondary view**, the **visible** occurrences of that word, of the **primary view**

### Expected Behaviour :

When a document is **cloned** in the **secondary** view, and some occurrences of a word are **highlighted** in green, in the **primary** view, scrolling up/down the **original** document, in **primary view** should not matter and, **all** the occurrences of that word, in the **secondary view**, should be **highlighted** in green, too !

### Debug Information :

Notepad++ v7.8.3   (32-bit)
Build time : Jan 12 2020 - 19:05:42
Path : D:\@@\783\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Microsoft Windows XP (32-bit) 
OS Build : 2600.0

### Hypotheses :

- Notepad++ version used : **`v7.8.3`**

- **No extra** plugin installed, just the **`3`** **default** ones

- Options ticked in **`Preferences... > Highlighting > Smart Highlighting`** : **`Enable`**, **`Match whole word only`** and **`Highlight another view`**. **All** the other options **un-ticked**

- Disable the **multi-line** tabs feature by **un-ticking** the option **`Preferences... > General > Tab Bar > Multi-line`** => **`1`** **row** of tabs, only !

- The **Horizontal** splitter between **primary** and **secondary** is selected. If not, simply **right**-click on the **splitter** and choose the  **`Rotate to left`** option

- **Default** zoom is used, both, in **primary** and **secondary** view ( **`Ctrl + Numpad /`** )

- No **horizontal** nor **vertical** synchronization of the **two** views. So, **un**-tick the options **`View > Synchronize Vertical Scrolling `** and **`View > Synchronize Horizontal Scrolling`**

- For readability, **enable** the **line numbers** column, in **`Preferences... > Editing > Display line number`**

---

### Steps to reproduce the Issue :

- Open a **new** tab

- Hit the **`Enter`** key to create **`20`** **empty** lines

- Then , append **`10`** lines containing the string **`ABCDE`** ( so, from lines **`21`** to **`30`** )

- End with **`2`** **empty** lines ( lines **`31`** and **`32`** )

- Save it as **`Test.txt`**

- Select the option **`View > Move/Clone Current Document > Clone to Other View`** => The **cloned** document **`Test.txt`** is located at **bottom** of screen

- Move line **`16`** as the **first** line in the **secondary** view

- Now, select the **original** **`Test.txt`** tab, on **top** of screen

- In the same way, move line **`16`** as the **first** line of the **primary** view

- **Double**-click on line **`23`** of the **primary** view => The **ten** lines **`ABCDE`**, in **each** view, are **highlighted**, as **expected**

- Then, while keeping the mouse pointer **over** the **primary** view, scroll **down** , with the mouse **wheel** ( or with the **vertical** bar ) , until **no** text is present => Only the **visible** line(s) of the **primary** view is/are **highlighted** in the **secondary** view !

- Now, scroll **up**, with the mouse **wheel**, until **no** text is present => In the same way, only the **visible** line(s) of the **primary** view is/are **highlighted** in the **secondary** view !

---

### Remarks :

- If, without changing the **active top** tab ( **primary** view ), we place the mouse pointer over the **secondary** view and scroll **up/down**, with the mouse **wheel**, nothing is changed in the **primary** view and the **ten** lines are always **highlighted**

- If we **stop** and **re-start** Notepad++ with the **`Test.txt`** displayed in **each** view, this minor bug **still** occurs

- I also noticed this behaviour, with the **old** N++ **`7.2`** version, which was the **first** release which included the **`Highlight another view`** option, regarding the **Smart Highlighting** feature !

- Although **not** tested, I suppose that the **recent** **`v7.8.4`** version does produce the same **bug**
 
---

### Important :

- Instead of **cloning** the file **`Test.txt`**, in the **secondary** view, let's create a **copy** of the file **`Test.txt`**, named **`Copy.txt`**

- Open this **new** file, in the **secondary** view, with the option **`View > Move/Clone Current Document > Move to Other View`**

- This time, if we repeat the **above** steps, whatever the **visible** lines **`ABCDE`**, in the **primary** view, obtained by **scrolling** text up or down, the **ten** lines **`ABCDE`** of the identical file **`Copy.txt`** are always **highlighted**, in the **secondary** view, as expected !

Best Regards,

guy038

jofon:
Smart highlighting with a cloned view and "Highlight another view" enabled was only highlighting the visible occurrences of the word in the Primary view, ignoring the cloned view. 

With this PR, the cloned view is highlighed correctly as if it weren't cloned.

Fixes #7910

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -861,8 +861,7 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			if (notification->nmhdr.hwndFrom != _pEditView->getHSelf()) // notification come from unfocus view - both views ae visible
 			{
 				//ScintillaEditView * unfocusView = isFromPrimary ? &_subEditView : &_mainEditView;
-				if (nppGui._smartHiliteOnAnotherView &&
-					_pEditView->getCurrentBufferID() != notifyView->getCurrentBufferID())
+				if (nppGui._smartHiliteOnAnotherView)
 				{
 					TCHAR selectedText[1024];
 					_pEditView->getGenericSelectedText(selectedText, sizeof(selectedText)/sizeof(TCHAR), false);
--- a/PowerEditor/src/ScitillaComponent/SmartHighlighter.cpp
+++ b/PowerEditor/src/ScitillaComponent/SmartHighlighter.cpp
@@ -171,11 +171,15 @@ void SmartHighlighter::highlightView(ScintillaEditView * pHighlightView, Scintil
 
 	highlightViewWithWord(pHighlightView, text2FindW);
 
-	if (nppGUI._smartHiliteOnAnotherView && unfocusView && unfocusView->isVisible()
-		&& unfocusView->getCurrentBufferID() != pHighlightView->getCurrentBufferID())
-	{
-		// Clear marks
-		unfocusView->clearIndicator(SCE_UNIVERSAL_FOUND_STYLE_SMART);
+	if (nppGUI._smartHiliteOnAnotherView && unfocusView && unfocusView->isVisible())
+	{		
+		// Clear the indicator only when the view is not a clone, or it will clear what we have already hightlighted in the pHighlightView
+		if (unfocusView->getCurrentBufferID() != pHighlightView->getCurrentBufferID())
+		{	
+			unfocusView->clearIndicator(SCE_UNIVERSAL_FOUND_STYLE_SMART);
+		}
+		
+		// Hightlight the unfocused view even if it's a clone, as it might be in a different area of the document
 		highlightViewWithWord(unfocusView, text2FindW);
 	}
 


RFImages:
This issue is not resolved until  Settings->Preferences->Highlighting->Smart Hightlighting->Highlight another view is checked by default or this option removed. Why would anyone not want this checked and tolerate unacceptable behavior that requires multiple posts in here for people to realize there is a bug on/off switch that is on by default?

Issue #7791, opened 2019-12-28
Some URLs are not parsed correctly


djhenderson:
### Description of the Issue
When a URL is immediately followed by some special characters, the URL plus the special character is underlined, highlighted on hover, and sent to your browser on double-click. When the URL is immediately followed by a single- or double-quote character, the quote is not included in the URL.

Due to the official URL format, this is correct behaviour, however in some cases, it is not helpful behaviour for an editor.

The markdown syntax for a link uses a link description between square brackets, followed by a URL in parenthesis. These are typically written with the parenthesis immediately adjacent to the URL, e.g.
    \[here]\(https://github.com/mohzy83/NppMarkdownPanel/releases).

### Steps to Reproduce the Issue

1. Open a new file and insert markdown link of the form \[name]\(URL)  e.g.
    \[here]\(https://github.com/mohzy83/NppMarkdownPanel/releases).
2. Inspect the source to observe the underlined text
3. hover over the URL and observe the underlined and highlighted text
4. double click the highlighted text and observe the URL visited by your browser.

### Expected Behaviour
2. only the URL is underlined
3. only the URL is highlighted and underlined
4. the URL sent to the browser opens the target page

### Actual Behaviour
When the right-parenthesis is immediately adjacent to the URL, you will see
2. the underline text includes the right-parenthesis
3. the highlighted and underlined text includes the right-parenthesis
4. the URL sent to the browser results in a 404 page or 404 error.

### Debug Information
Notepad++ v7.8.2   (64-bit)
Build time : Dec  4 2019 - 01:39:49
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Home (64-bit) 
OS Version : 1809
OS Build : 17763.914
Plugins : Comment Wrap.dll ComparePlugin.dll mimeTools.dll NotepadStarterPlugin.dll NppConverter.dll NppExport.dll NppMarkdownPanel.dll NppSaveAsAdmin.dll Python Indent.dll zoomdisabler_x64.dll 

### Comments
The syntax parsing of a URL in markdown source should only include a trailing right-parenthesis when there is an unmatched left-parenthesis in the trailing component of the URL. This is still not strictly correct, however it is, I believe, more correct than the current behaviour for markdown.

Please consider excluding a trailing right-parenthesis from URLs in markdown source.

[...]

Issue #6155, opened 2019-09-17
hyperlinks don't recognise international characters in URLs


HairyPhil:
### Description of the Issue

Conversion of URLs to hyperlinks stops when it encounters an international character.

### Steps to Reproduce the Issue

Copy and paste the following example URL to a blank txt document:
https://www.amazon.fr/GRÜNTEK-Ebrancheur-Cisailles-crémaillère-SATISFAIT/dp/B01COZUA4E/

(Alternative example:)
https://www.amazon.fr/Gardena-coupe-branches-TeleCut-520-670-télescopiques/dp/B07JLJ4N44/

### Expected Behavior

The whole URL should be rendered as a hyperlink.

### Actual Behavior

The hyperlink incorrectly ends after the R in GRÜNTEK because of the umlaut.

### Debug Information
Notepad++ v7.7.1   (64-bit)
Build time : Jun 16 2019 - 21:24:47
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll

HairyPhil:
Note that GitHub renders the hyperlinks correctly.

ArkadiuszMichalski:
@sasumner 
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8634 is not the same? The title suggest to Cyrillic, but the new regex probably cover a wider range of characters (including those here). Maybe close this bug?

Issue #5029, opened 2018-11-18
highlighting and executing http link not ignoring trailing (closing) bracket


jannick0:
### Description of the Issue
* When parsing a http links a trailing closing bracket `)` is not ignored and highlighted. Moreover, the trailing bracket is passed on to the browser when double-clicking it.
* Use case: E.g., markdown syntax for web links  `[ABC](http://www.abc.com)` or web references enclosed in brackets.
* I am not sure if brackets in an URI are required to be matching acc. to the standard. 

### Steps to Reproduce the Issue

1. Type `[Google](http://www.google.com)`
2. Doube-click on link

### Expected Behavior

* Link highlighed in NPP should **not** include trailing bracket.
* Trailing bracket in link passed to browser should be discarded.

### Actual Behavior

* Link highlighed in NPP **does** include trailing bracket.
* Trailing bracket in link passed to browser *not* discarded.

sasumner:
Possibly related to #3353

jannick0:
- [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): Uniform Resource Identifier (URI), Generic Syntax, Section 3.
- If matching brackets are the issue, recursive regular expressions could help. It might be even possible to put the formal URI definition into a lengthy regex definition (cf., e.g., [URI regular expression](jmrware.com/articles/2009/uri_regexp/URI_regex.html) ... untested).
- It would be great if someone could point (me) to the NPP code snip where the link parsing is happening (regexp, manual parsing, etc). Probably worth while looking at that in the light of the increasing usage of markdown et al. (#3353 is definitely related). Many thanks!

PopeSpaceous:
Hi! I have been looking into this issue. 
It seems like the regular expression for finding the url in the text editor could use some work. 
Currently it is this "[A-Za-z]+://[A-Za-z0-9_\\-\\+~.:?&@=/%#,;\\{\\}\\(\\)\\[\\]\\|\\*\\!\\\\]+"
Essentially it says.
[A-Za-z]+                       ://               [A-Za-z0-9_\\-\\+~.:?&@=/%#,;\\{\\}\\(\\)\\[\\]\\|\\*\\!\\\\]+
I'll take any number of alphabetical characters until :// then any number of alphanumeric, _ -,+~.:?&@=/%#,; and brackets. 
So to fix your issue we need a better regular expression. 
[...]

PR #8921, opened 2020-09-29
URL detection with parser.

Uhf7:
Intended to solve #3912,  #3353, #4643,  #5029, #6155, #7791, #8634 and probably others.

Wininet.lib is required, so I had to extend the `.vcxproj` file. The function I used from this, `InternetCrackURL`, exists since Windows 2000, so there should be no problem with older systems.

Special thanks go to @Yaron10 and @Ekopalypse, they did preliminary tests which helped a lot to improve it.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -27,6 +27,7 @@
 
 #include <time.h>
 #include <shlwapi.h>
+#include <wininet.h>
 #include "Notepad_plus.h"
 #include "Notepad_plus_Window.h"
 #include "FileDialog.h"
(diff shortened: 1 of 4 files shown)


sasumner:
Fails on mingw build due to missing wininet.lib.
Add by looking at how winmm.lib was added (this one bit me).
Still no mingw build instructions. :-(

Uhf7:
I simply forgot mingw. So I extended the gcc makefile now too.

Yaron10:
I've been using the parser for the past month.
Apart from fixing the above mentioned issues, the performance (CPU etc. - on opening, fast scrolling) is noticeably better.
Please test it with [these files](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5021684/Test.zip).

Thank you @Uhf7 again for your work. :+1:

chcg:
@Uhf7 See also https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/CMakeLists.txt#L360 which should be extended by winmm and wininet.

Uhf7:
@chcg, 

thank you for that hint. What confuses me here are the different library lists in the gcc `makefile`, in the `CMakeLists.txt`, and in the MS VC project file `.vcxproj`.

Since I don't use CMake for this project, I'm not fully knowing what I'm doing, if I modify this. For the other two I can see the consequences at least while the AppVeyor build.

The libs are, sorted alphabetically, without wininet (before this PR):
1. `makefile`:  comctl32 comdlg32 crypt32 dbghelp gdi32 ole32 oleacc sensapi shell32 shlwapi uuid version winmm wintrust  
2. `.vcxproj`: comctl32 crypt32 dbghelp sensapi shell32 shlwapi oleacc version winmm wintrust
3. `CMakeLists.txt`: comctl32 crypt32 dbghelp sensapi shlwapi version wintrust

I think there are some possibilities:
- All three ways use different standard libraries.
- CMake wasn't used for a long time and is incomplete anyway.
- Both, `makefile` and `.vcxproj`, could contain some unnecessary libs.

If it would be sure that CMake builds the `.vcxproj` file and it includes the listed libraries as additional dependencies into this `.vcxproj` file, then we would need to include shell32, oleacc and (the already mentioned) winmm too, so that the library lists in 2. and 3. are equal.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2530,23 +2825,36 @@ void Notepad_plus::addHotSpot(ScintillaEditView* view)
...
-	}
-	if (endPos > startPos)
-		pView->execute(SCI_INDICATORCLEARRANGE, startPos, endPos - startPos);
+	UINT cp = static_cast<UINT>(pView->execute(SCI_GETCODEPAGE));
+	char *encodedText = new char[endPos - startPos + 1];
+	pView->getText(encodedText, startPos, endPos);
+	TCHAR *wideText = new TCHAR[endPos - startPos + 1];
+	int wideTextLen = MultiByteToWideChar(cp, 0, encodedText, endPos - startPos + 1, (LPWSTR) wideText, endPos - startPos + 1) - 1;
+	if (wideTextLen <= 0) return;
+	delete[] encodedText;
```
donho, line 2834: There are somethings incoherent here:
if (wideTextLen <= 0)  you return then encodedText is allocated -> memory leak.
Or am I missing somethings ?

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2530,23 +2825,36 @@ void Notepad_plus::addHotSpot(ScintillaEditView* view)
 	LRESULT indicFore = pView->execute(SCI_STYLEGETFORE, STYLE_DEFAULT);
 	pView->execute(SCI_SETINDICATORVALUE, indicFore);
 
-	pView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
-	pView->execute(SCI_SETTARGETRANGE, startPos, endPos);
-	int posFound = static_cast<int32_t>(pView->execute(SCI_SEARCHINTARGET, strlen(URL_REG_EXPR), reinterpret_cast<LPARAM>(URL_REG_EXPR)));
```
donho, line 2535: If `URL_REG_EXPR` is unused, the following line should be removed from `Notepad_plus.h`:
`#define URL_REG_EXPR "[A-Za-z]+://[A-Za-z0-9_\\-\\+~.:?&@=/%#,;\\{\\}\\(\\)\\[\\]\\|\\*\\!\\\\]+"`

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2530,23 +2825,36 @@ void Notepad_plus::addHotSpot(ScintillaEditView* view)
...
+	delete[] encodedText;
+
+	int startWide = 0;
+	int lenWide = 0;
+	int startEncoded = 0;
+	int lenEncoded = 0;
+	while (true)
+	{
+		bool r = isUrl(wideText, wideTextLen, startWide, & lenWide);
+		if (lenWide <= 0) break;
```
donho, line 2843: It's not in the rules of coding style, but it's preferred (you can follow the coding style of existent code):
``` 
if (lenWide <= 0)
    break;
```
then
`if (lenWide <= 0) break;`

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2497,6 +2498,300 @@ void Notepad_plus::setUniModeText()
...
+{
+	int dist, schemeLen;
+	if (scanToUrlStart(text, textLen, start, & dist, & schemeLen))
+	{
+		if (dist)
+		{
+			*segmentLen = dist;
+			return false;
+		}
+		int len;
```
donho, line 2770: Good practice: initialize the variable.
```
int len = -1; // or 0
```

donho:
Impressive! Great job!

image

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2497,6 +2498,300 @@ void Notepad_plus::setUniModeText()
...
+
+bool isUrlSchemeSupported (INTERNET_SCHEME s)
+{
+	switch (s)
+	{
+		case INTERNET_SCHEME_FTP:
+		case INTERNET_SCHEME_HTTP:
+		case INTERNET_SCHEME_HTTPS:
+		case INTERNET_SCHEME_MAILTO:
+			return true;
```
donho, line 2553: The current scheme allow all the forms of URI: `[A-Za-z]+://...`
We can see in the future if users really need it
However, maybe we can consider at least the schemes of **file**, **data**, **ldap**, **telnet** and **irc**?
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

donho: OK, just checked the header, it's quite limited... in this case, why not to add INTERNET_SCHEME_FILE ?
It'll be nice to support such url:
file:///C:/Users/xxx/Downloads/sxxxcilier-fr-202010-invoice00239637.pdf

donho (changes requested):
1. Remove potential memory leak.
2. Remove unused RegExpr `URL_REG_EXPR`.
3. Initialize all variables on their declaration.
4. (Not mandatory) `Add file:///` support

PowerEditor/src/Notepad_plus.cpp
Uhf7: Unfortunately, including the `file:` scheme doesn't mean to extend `isUrlSchemeSupported` only. File names require a different parsing:
- They can contain spaces. And spaces in file names should certainly be allowed when the file name is enclosed in double quotes and otherwise not.
- They can contain only one colon (located after the drive letter).
- They may not contain some other characters, which are allowed in `http:`-URLs.

I included some code to support the `file:` scheme, which works for me. I made it removable, if don't like it.

ArkadiuszMichalski:
@Uhf7 What standard/spec you follow when write parser? Or maybe try cover most problems reported by Notepad++ users?

Uhf7:
Spec!? Would be glad to have one which covers our problem here.

I used input from different sources:

- Notepad++ issues
- Best of the behavior of MS VC, MS Outlook and the Github.com web interface in which I write here
- Wikipedia
- https://skorks.com/2010/05/what-every-developer-should-know-about-urls/ (my favorite)
- http://www.w3.org/Addressing/URL/url-spec.txt

In this order, I would guess. The main problem we have, is not the URL format as such, but where to draw the dividing line between the normal text and the URL (An URL with three closing parenthesis at the end and no opening one may match the spec perfectly, but we don't want it to highlight it this way).

The first two points listed above are about delimitation between normal text and highlighted URL, the rest of it describe more or less the URL format.

ArkadiuszMichalski:
Ok, you can also check https://url.spec.whatwg.org/, and maybe it test (https://github.com/web-platform-tests/wpt/tree/master/url, like https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json). Of course it's only for URL format, not delimiters. Maybe it will help you (for other schames), maybe not, you decide.  It's great that finally this URL will be better recognized in Notepad++, big thx!!!

Uhf7:
@ArkadiuszMichalski , thank you for those links, the last one is the most interesting for me. Although I wouldn't accept every example from various the "href"-fields there.

PowerEditor/src/Notepad_plus.cpp
Yaron10: > They can contain spaces. And spaces in file names should certainly be allowed when the file name is enclosed in double quotes and otherwise not.

File-names containing spaces are quite common.
And currently `file:///C:/Test 1.txt` is not parsed correctly.

If the fix is not simple, I'm wondering if removing `INTERNET_SCHEME_FILE` shouldn't be better.

Uhf7: @Yaron10,

file names containing spaces are supposed to be enclosed in "double quotes", like in command lines. There is no other way to detect the end, except guessing.

Yaron10: @Uhf7,

A file containing the `INTERNET_SCHEME_FILE` would usually be `file:///C:/Test 1.txt` (NOT `"file:///C:/Test 1.txt"`).
So the parsing would not work properly in many cases.

IMO this can cause confusion and removing `INTERNET_SCHEME_FILE` would be better.

Thank you.

Uhf7: @Yaron10,
the correct way would be `file:///C:/Test%201.txt`. I added the special handling for the file scheme with quotes to allow really spaces in file names, but it seems not to be too intuitively.

Some ways to go here:
- Remove file scheme completely.
- Remove special handling for file scheme, then
  - Spaces have always to be coded as `%20`
  - It is possible to construct invalid file names, at least, if this method is used to open local files
- Allow the percent-encoding in the special handling for file scheme, then the spaces would work in quotes and the percent-encoding would work too.

I'm open to everything here.

Yaron10: Continued in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8975.

PR #8923, opened 2020-09-29
Update Danish translation to 7.9.1

scootergrisen:


donho:
@scootergrisen 

Thank you!

scootergrisen:
I would prefer to use the changes as they are now and then i can update again at a later time.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5d6e5c6e9cacd4da54d43c501314d0fde02c128a

PR #8935, opened 2020-09-30
Update italian.xml

conky77:


donho:
@conky77 

Thank you!

conky77:
@donho  Italian translation complete

PR #8940, opened 2020-10-01
Remove TreeView Beep when pressing return key.

### LINKED ISSUE #8938: Annoying bell sound on keystroke ENTER in Function list & Folder as Workspace
### Description of the Issue
In both Function list & Folder as Workspace, if select an item then type ENTER, it produce an annoying bell sound.

1. In Function list or Folder as Workspace, select an item
2. Type ENTER

### Expected Behavior
No sound

### Actual Behavior
Bell sound

### Debug Information
Notepad++ v7.9   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2004
OS Build : 19041.508
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppXmlTreeviewPlugin.dll

Uhf7:
Fixes #8938.

The Windows TreeView control initiates a Default Beep sound, if it's `NM_RETURN` notification remains unhandled.

### DIFF
--- a/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
+++ b/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
@@ -633,6 +633,9 @@ void FileBrowser::notified(LPNMHDR notification)
 			}
 			break;
 
+			case NM_RETURN:
+				SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, 1);
+			break;
 
 			case TVN_KEYDOWN:
 			{
--- a/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
+++ b/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
@@ -561,6 +561,10 @@ void FunctionListPanel::notified(LPNMHDR notification)
 			}
 			break;
 
+			case NM_RETURN:
+				SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, 1);
+			break;
+
 			case TVN_KEYDOWN:
 			{
 				LPNMTVKEYDOWN ptvkd = (LPNMTVKEYDOWN)notification;
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -628,6 +628,10 @@ void ProjectPanel::notified(LPNMHDR notification)
 				openSelectFile();
 			}
 			break;
+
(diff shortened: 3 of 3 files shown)


donho:
Terrific! Could you tell me how to find this tip if it's not a secret? (the keywords you used on google to find this trick)

donho (changes requested):
1. Use `DWLP_MSGRESULT` instead of `0` or `Use `DWL_MSGRESULT` to have both x86 & x64 build.
2. Remove the beep of `TAB` & `ESCAPE` for Function list as well

Uhf7:
The most important hint was in https://docs.microsoft.com/en-us/windows/win32/controls/wm-notify. There is the statement "If the message handler is in a dialog box procedure, you must use the `SetWindowLong` function with `DWL_MSGRESULT` to set a return value".

The way to it is:
- Remove any reaction to the return key from Notepad++ and be astonished about the fact, that the "beep" survives this.
- Read several internet pages about how this problem has been solved for a Visual Basic wrapper class around the TreeView control.
- Read the TreeView control documentation, especially here: https://docs.microsoft.com/en-us/windows/win32/controls/tvn-keydown and here: https://docs.microsoft.com/en-us/windows/win32/controls/nm-return-tree-view-. This two articles reveal, that the return code of `TVN_KEYDOWN` is unimportant. And, that a non-zero return code of `NM_RETURN` prevents the default processing, which seems to be "beep".
- Wonder, why the Windows procedures in Notepad++ return TRUE for all `WM_NOTIFY` messages, but still the "beep" is there. Then, you come to the `DWL_MSGRESULT` for dialog box procedures.

Uhf7:
> Use DWLP_MSGRESULT instead of 0 or DWL_MSGRESULT to have both x86 & x64 build.

That's reasonable.

Remove the beep of TAB & ESCAPE for Function list as well

Do we have Christmas already? This PR was only intended to remove the "beep" from the enter key presses.
If there are more problems, I would be happy to solve them later, but I'm a bit busy right now.
If you have a better solution until then, please implement it without me, I don't need my name in the log for this.

donho:
> Do we have Christmas already? 

Soon, but not yet. But even though, I would never allow myself to ask gifts from people. I just thought it's fair to give you the credit for this feature, since it's you who found the solution.

Thank you anyway for your help - the life changes without the beep!

donho:
BTW, beep of ESC & ESCAPE key are removed in this commit:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c94f84000b7806866740d92a13529788f1487bbb

Issue #8943, opened 2020-10-02
Add "Marked Style N text to Clipboard" menu commands


donho:
Add "Marked Style N text to Clipboard" (N:1..5) &  "All Marked text to Clipboard" 6 commands.
So user can mark text with any style and copy them to the clipboard.

sasumner:
Would this be to copy only _unique_ text?

Example:
If a user styles one text `Jack` in a document, ALL occurrences get the styling:
![image](https://user-images.githubusercontent.com/30118311/94950600-e0f77100-04b0-11eb-9a85-4b1aabad626c.png)
Is there much use in copying this data THREE times?

donho:
@sasumner 
Here is an example:
![image](https://user-images.githubusercontent.com/90293/94951456-9394f700-04e4-11eb-85fc-7e08d33d8705.png)

sasumner:
@donho 

Yes, I understand the core concept, but for your example, should output (clipboard contents) for the "all" case be (A):

```
Notepad++
source code editor
GNU General Public License
Notepad++
Notepad++
a greener environment
```

Or (B):

```
Notepad++
source code editor
GNU General Public License
a greener environment
```

I think it should be (B).
User likely won't want redundant data, but has no way to NOT get all the duplicates at time of doing the styling.

sasumner:
For naming suggestion for new commands, see https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8867#discussion_r498952885

ArkadiuszMichalski:
Copy bookmarks allows duplicates ...

donho:
(A) "all" case

It's for user to eliminate the redundant text if they don't want them.

donho:
OTOH, I'm not sure where is suitable to put these commands.
It could be in `Edit >> Copy to Clipboard` or add `Copy to Clipboard` just behind `Search >> Jump down`.

sasumner:
>It's for user to eliminate the redundant text if they don't want them.

OK...
But they will ask: 
"I only selected `Notepad++` once, how come I got 25 copies of it when I pasted?"
"How can I remove the redundant text that happened when I pasted?"
And they will not like the answer when someone starts talking about regular expression replacement. :-)

It could be a problem, because the core of the "styling" feature really is all about multiple occurrences being visually highlighted.

Just pointing out the potential problems.

Possible ideas (for sometime in the future, if problems result):

[...]

PR #8864, opened 2020-09-17
Update Corsican translation for Notepad++ 7.9.1

Patriccollu:
Hello,

All the strings containing `styles/tabs/groups/occurrences` with a numbering have been changed to use superscript instead of normal text.

Following up commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/55d671719c8f00f02ed32b15575bc2e6587a5e30 and additional minor changes.

Cheers,
Patriccollu.

PR #8957, opened 2020-10-05
Add ability to copy to clipboard text styled with styles 1 thru 5

sasumner:
Implements #8943

This is the second part of the change begun with #6095 and #8864 
Those implemented copying (red)marked text to the clipboard.


sasumner:
Closing; will create new PR to resolve conflicts.

PR #8967, opened 2020-10-07
Update czech.xml to v7.9.1

xomx:


donho:
@xomx 

Thank you!

xomx:
The translation has been revised according to the diff.

PR #8992, opened 2020-10-11
Fix possible file corruption on file open for writing

### LINKED ISSUE #5664: File is erased on save when there is no free space left on disk

pnedev:
Do not destroy file contents when opening user file for writing.
This makes sure original file is not corrupted if file writing fails for some reason.

Microsoft documentation for reference:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fseek-fseeki64?view=vs-2019

Fix #5664

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -879,7 +879,21 @@ bool FileManager::backupCurrentBuffer()
 				::SetFileAttributes(fullpath, dwFileAttribs);
 			}
 
-			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc"));
+			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("rb+c")); // Use "r+" to NOT destroy existing file contents on open
+			if (fp)
+			{
+				// On "r+" open success move file position to the very beginning of file
+				if (_fseeki64(fp, 0, SEEK_SET))
+				{
+					UnicodeConvertor.fclose(true);
+					fp = NULL;
+				}
+			}
+			else
+			{
+				fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc")); // If "r+" open fails file might not exist so try opening it anew with "w"
+			}
+
 			if (fp)
 			{
 				int lengthDoc = _pNotepadPlus->_pEditView->getCurrentDocLen();
@@ -1004,7 +1018,21 @@ bool FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool isCopy, g
 
 	int encoding = buffer->getEncoding();
 
-	FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc"));
+	FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("rb+c")); // Use "r+" to NOT destroy existing file contents on open
+	if (fp)
+	{
+		// On "r+" open success move file position to the very beginning of file
+		if (_fseeki64(fp, 0, SEEK_SET))
+		{
+			UnicodeConvertor.fclose(true);
+			fp = NULL;
+		}
+	}
+	else
+	{
+		fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc")); // If "r+" open fails file might not exist so try opening it anew with "w"
+	}
+
 	if (fp)
 	{
 		_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, buffer->_doc);	//generate new document
--- a/PowerEditor/src/Utf8_16.cpp
+++ b/PowerEditor/src/Utf8_16.cpp
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "Utf8_16.h"
+#include <io.h>
 
 const Utf8_16::utf8 Utf8_16::k_Boms[][3] = {
 	{0x00, 0x00, 0x00},  // Unknown
@@ -432,7 +433,7 @@ void Utf8_16_Write::setEncoding(UniMode eType)
 }
 
 
-void Utf8_16_Write::fclose()
+void Utf8_16_Write::fclose(bool keepOriginalLength)
 {
 	if (m_pNewBuf)
 	{
@@ -442,6 +443,12 @@ void Utf8_16_Write::fclose()
 
 	if (m_pFile)
 	{
+		if (!keepOriginalLength)
+		{
+			const int fd = _fileno(m_pFile);
+			_chsize_s(fd, _ftelli64(m_pFile));
+		}
+
 		::fflush(m_pFile);
 		::fclose(m_pFile);
 		m_pFile = NULL;
--- a/PowerEditor/src/Utf8_16.h
+++ b/PowerEditor/src/Utf8_16.h
@@ -143,7 +143,7 @@ public:
 
 	FILE * fopen(const TCHAR *_name, const TCHAR *_type);
 	size_t fwrite(const void* p, size_t _size);
-	void   fclose();
+	void   fclose(bool keepOriginalLength = false);
 
 	size_t convert(char* p, size_t _size);
 	char* getNewBuf() { return reinterpret_cast<char*>(m_pNewBuf); }
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -125,7 +125,7 @@ generic_string relativeFilePathToFullFilePath(const TCHAR *relativeFilePath)
 
 void writeFileContent(const TCHAR *file2write, const char *content2write)
 {
-	FILE *f = generic_fopen(file2write, TEXT("w+c"));
+	FILE *f = generic_fopen(file2write, TEXT("wc"));
 	fwrite(content2write, sizeof(content2write[0]), strlen(content2write), f);
 	fflush(f);
 	fclose(f);


PowerEditor/src/Utf8_16.cpp
```diff
@@ -442,6 +443,12 @@ void Utf8_16_Write::fclose()
 
 	if (m_pFile)
 	{
+		if (updateEOF)
+		{
+			const int fd = _fileno(m_pFile);
+			_chsize_s(fd, _ftelli64(m_pFile));
+		}
```
donho, line 450: Why do we need reset the file length here?
Does `fflush `not put what we need exactly?

pnedev: When file is opened with "rb+" its content is not erased and the EOF (end-of-file) mark is not updated as well.
We begin to write the new data over the previous and if the file grows bigger the EOF mark is updated properly (file is enlarged).
On the other hand, if the written new data is less and the file is supposed to shrink the EOF mark is not automatically updated (because the file is opened for read/write access and the remaining old data we had not updated is kept intact).
So if we want to shrink the file we have to explicitly truncate it to the end of the new written content. `_ftelli64()` should tell us the current position in the file since the last successful write and we move the EOF there truncating the unneeded content.
Unfortunately I couldn't find standard CRT function that directly sets the EOF mark position and thus I had to use `_chsize_s()`

donho: So are these 2 lines for shrinking the closing file ?

pnedev: Yes

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -879,7 +879,21 @@ bool FileManager::backupCurrentBuffer()
 				::SetFileAttributes(fullpath, dwFileAttribs);
 			}
 
-			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc"));
+			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("rb+c")); // Use "r+" to NOT destroy existing file contents on open
+			if (fp)
+			{
+				// On "r+" open success move file position to the very beginning of file
+				if (_fseeki64(fp, 0, SEEK_SET))
```
donho, line 886: Here if `_fseeki64` failed to put file pointer on the beginning of file, you consider there's no more room on the hard disk, right?

PowerEditor/src/Utf8_16.h
```diff
@@ -143,7 +143,7 @@ class Utf8_16_Write : public Utf8_16 {
 
 	FILE * fopen(const TCHAR *_name, const TCHAR *_type);
 	size_t fwrite(const void* p, size_t _size);
-	void   fclose();
+	void   fclose(bool updateEOF = true);
```
donho, line 146: Please change to:
`void   fclose(bool keepOriginalLength = false);`

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -1004,7 +1018,21 @@ bool FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool isCopy, g
...
 	int encoding = buffer->getEncoding();
 
-	FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc"));
+	FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("rb+c")); // Use "r+" to NOT destroy existing file contents on open
+	if (fp)
+	{
+		// On "r+" open success move file position to the very beginning of file
+		if (_fseeki64(fp, 0, SEEK_SET))
+		{
+			UnicodeConvertor.fclose(false);
```
donho, line 1027: Please change to:
`UnicodeConvertor.fclose(true);`

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -879,7 +879,21 @@ bool FileManager::backupCurrentBuffer()
...
 			}
 
-			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc"));
+			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("rb+c")); // Use "r+" to NOT destroy existing file contents on open
+			if (fp)
+			{
+				// On "r+" open success move file position to the very beginning of file
+				if (_fseeki64(fp, 0, SEEK_SET))
+				{
+					UnicodeConvertor.fclose(false);
```
donho, line 888: Please change to:
`UnicodeConvertor.fclose(true);`

PowerEditor/src/Utf8_16.cpp
```diff
@@ -432,7 +433,7 @@ void Utf8_16_Write::setEncoding(UniMode eType)
 }
 
 
-void Utf8_16_Write::fclose()
+void Utf8_16_Write::fclose(bool updateEOF)
```
donho, line 436: Please change `updateEOF` to `keepOriginalLength` so it's more explicit

donho (changes requested):
This logic of this PR works but obviously not tested :)
After fixing the error, this PR fix the file trucked issue after failure of saving (tested with my 3 MB vitual drive)

PowerEditor/src/ScitillaComponent/Buffer.cpp
pnedev: > Here if _fseeki64 failed to put file pointer on the beginning of file, you consider there's no more room on the hard disk, right?

No. I simply make sure `_fseeki64` does its job properly. If for some reason it doesn't, then we'll simply report write failure.

> According the description of MS fseek's document:
`On devices incapable of seeking, the return value is undefined.`

`On devices incapable of seeking, the return value is undefined.` are other types of streams (for example `stdin`). By `devices` its meant streams that are accessed strictly sequentially - no seeking possible for those.

pnedev:
@donho ,
The requested changes are done.

PowerEditor/src/Utf8_16.cpp
```diff
@@ -442,6 +443,12 @@ void Utf8_16_Write::fclose()
 
 	if (m_pFile)
 	{
+		if (!keepOriginalLength)
```
donho, line 446: @pnedev 
Here `if (!keepOriginalLength)` should be `if (keepOriginalLength)` IMO.
That's why I ask you to inverse the boolean... or am I missing somethings?

pnedev: By default the file length should be changed.
When you open the file with `fopen` and rewind it to position 0 with `_fseek64` you start doing writes with `fwrite`.
Each `fwrite` call updates the position (increments it with the bytes written).
When you are done writing you close the file with `fclose`. Now if you have written less data than the original file you need to update the file length to truncate the remaining data that was in the file before your edit.
So by default the change in file length should happen after all writing is done and before you close the file. This way the file will always be with size equal to the position after the last write.

The only case when we don't want to change file length is if after `fopen`, `fseek` fails to set the position in the beginning of the file (for some unknown reason).

donho: @pnedev 
> By default the file length should be changed.

"By default" do you mean in the normal situation? If yes, I don't see the point, because saving function works fine (except in such situation) in Notepad++ without these 2 lines.

However, I tested again with both  `if (!keepOriginalLength)` & `if (keepOriginalLength)` under a very small room situation - both makes my file truncated. So it doesn't work this PR for me. 
Have you tested your PR?

pnedev: @donho ,

> "By default" do you mean in the normal situation?

Yes.

> If yes, I don't see the point, because saving function works fine (except in such situation) in Notepad++ without these 2 lines.

When we open file with mode "rb+c" (as is in the PR) the saving doesn't work fine without these 2 lines.

If the file was originally this:
`aaaaaaaa`
and we edit it in Notepad++ with mode "rb+c" to be
`bbbb`
after the save WITHOUT these 2 lines we'll get:
`bbbbaaaa`.

So these 2 lines are needed to truncate the file when we save it with shorter length.

> Have you tested your PR?

Yes, I have.
But I test it with RAM disk and it worked.
After your comment I tested the original Notepad++ with RAM disk and the file was again not erased when there was no space left. It appears RAM disk doesn't behave quite like normal disk.

So I'll re-test both the original Notepad++ and the PR with normal USB drive and will get back to you soon.

donho: @pnedev 
> If the file was originally this:
> aaaaaaaa
> and we edit it in Notepad++ with mode "rb+c" to be
> bbbb
> after the save WITHOUT these 2 lines we'll get:
> bbbbaaaa.

I don't like such hack. - it "fixes" a bug in a extreme context, but could affect app's working in 98% normal situation.

> But I test it with RAM disk and it worked.

I use VeraCrypt to create a disk, then mount it as a drive - quite easy to do.

pnedev: I have to test this on native Windows.
My tests on Linux under Wine always show the file contents are preserved.

I'll write when I manage to test it.

pnedev: @donho 

> I don't like such hack. - it "fixes" a bug in a extreme context, but could affect app's working in 98% normal situation.

It is not a hack. It is simply to open the file in "update" mode - read, write, change some content.
The problem is that changing existing content doesn't automatically shrink the file.

And it also doesn't affect app's working at all - It opens the file more consciously to make sure file is not immediately destroyed on open (when we use `fopen`). Because if currently `fopen` succeeds opening the file and `fwrite` fails for any reason (not only 'no space left' problem) the file is gone.

But it is of course up to you to decide if you prefer this approach or will think of another.
If you don't like this PR simply reject it so I don't waste more time in vain.

Thanks.

sasumner: >If you don't like this PR simply reject it so I don't waste more time in vain.

I think if the idea wasn't liked, that would have already happened.
Since it hasn't, please, both of you, keep "negotiating" until something here is merged.

donho: @pnedev 
normally, to write data to the disk the procedure is : 
1. fopen
2. fwrite
3. fflush
4. fclose

I don't see the necessity to resize the file. 
I can accept that we modify the normal procedure to save in an extreme condition, but calling `_chsize_s` to resize the file length is not part of normal procedure. Therefore I cannot take this way as the solution.

> If you don't like this PR simply reject it so I don't waste more time in vain.

OK, I will close this PR. My role is keeping the product working so if your future PR is about the special treatment only on the special case, and keep the normal routing as it is, it will be welcome.  

@sasumner 
> Since it hasn't, please, both of you, keep "negotiating" until something here is merged.

No offense. It's normal that devs defense their viewpoint. OTOH, I have my principle to keep.

pnedev: @donho ,

Current "normal" save procedure:

1. fopen (mode "wbc") - File is implicitly truncated to zero size (`_chsize_s(file, 0)` implicitly). ANY fail in down functions lead to data loss.
2. fwrite
3. fflush
4. fclose

Proposed procedure (no "less normal" at all):

1. fopen (mode "rb+c") - File is NOT implicitly truncated to zero size - fail in down functions DO NOT lead to data loss.
2. fseek - to start of file 
3. fwrite
4. _chsize_s - explicitly (meaning in our control) to the new data size
5. fflush
6. fclose

All the above functions are not "weird" and "not normal". They are valid functions from the file API and are there for a reason. They wouldn't exist otherwise.

----------------------------------------------------

Will you accept the change if I use `_chsize_s` ONLY when the file needs to be shrunk?
Or you'll use another approach and fix it on your own?

Thanks.

sasumner:
@donho said:

> It's normal that devs defense their viewpoint.

Yes, I understand.
But, if dev has a nice core idea for improvement, it is good to somewhat lead the dev to an acceptable implementation.
At least, that's how I feel about it.

@pnedev Thanks for pressing forward. :-)

PowerEditor/src/Utf8_16.cpp
donho: @pnedev 
> Will you accept the change if I use _chsize_s ONLY when the file needs to be shrunk?

I will accept to keep the current "normal" save procedure as it is for the normal situation (since it's working fine), and use any way you want (including _chsize_s ) in the "no more space on disk" situation.

pnedev: @donho ,

> I will accept to keep the current "normal" save procedure as it is for the normal situation (since it's working fine), and use any way you want (including _chsize_s ) in the "no more space on disk" situation.

Then we'll need completely new approach because `fopen` is the source of the possible corruption but you don't want to use different `fopen` mode that is MUCH safer.
Since you forbid the use of other `fopen` mode we'll need to do something BEFORE calling `fopen` to make sure the size is enough to write the entire file.

Here is one solution:

1. Check remaining free disk size (MSDN `GetDiskFreeSpaceEx()`).
2. If size is enough continue with your "normal" save procedure:
3. `fopen`
4. `fwrite`
5. `fflush`
6. `fclose`

So do you prefer to check the remaining free disk size before EVERY file save more that the approach in the current PR?

donho: > Then we'll need completely new approach because fopen is the source of the possible corruption but you don't want to use different fopen mode that is MUCH safer.

The reason is just because the code is there and it is working fine (except for no free space situation). I don't see why we should replace it by the code which could bring some regression.


> 1. Check remaining free disk size (MSDN GetDiskFreeSpaceEx()).
> 2. If size is enough continue with your "normal" save procedure:
> ...
> So do you prefer to check the remaining free disk size before EVERY file save more that the approach in the current PR?

It depends on the performance of binary and the complexity of the code.
Check disk space on each save is not reasonable to me. It could be done once on the first saving for each buffer though.

pnedev: @donho ,

> The reason is just because the code is there and it is working fine ...

But it is apparently not working fine since it erases the original file before writing the memory contents to the disk which introduces a serious risk to loose data. Just because this doesn't happen normally it doesn't mean the risk is not there. And it is intrinsic to using `fopen` with `w` mode.

> I don't see why we should replace it by the code which could bring some regression.

The change is so minimal and it can be well tested. You can even keep use the fixed exe for some time and see if you'll encounter any problem.

I have made another PR that is working fine and fixes the problem - https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9024

> Check disk space on each save is not reasonable to me.

It is not reasonable to me either. And also not as secure as the current approach.
Besides, it is better to call `_chsize_s` ONLY when needed than `GetDiskFreeSpaceEx` everytime.

> It could be done once on the first saving for each buffer though.

That would be only half-measure. We'll need to keep status that the buffer is already saved once but more importantly:
there might be user downloading task in the background and just when the user tries to save his/hers file the free space to run-out.

PR #8998, opened 2020-10-12
Widen length/lines part of status bar

### LINKED ISSUE #4157: Number of lines in the file is missing the last digit
### Description of the Issue
The number of lines in the file (displayed in the bottom status bar) is missing the last digit when there are 10+ million lines in the file.
image

1. Create a file with more than 12 million lines of data (for example, 14,225,877).
2. The "lines" count that is displayed in the bottom center of the status bar will be missing the very last digit (in this example, you will see "lines: 14,225,87")

### Expected Behavior
You expect that all digits of the line count would be displayed.

### Actual Behavior
The last digit of the line count is missing.

### Version Info
This bug was discovered in version 7.5.4 of Notepad++ (32-bit)

sasumner:
Fixes #4157

Surely(?) a bit more room can be spared in the status bar for the `length: x,xxx  lines: x,xxx` part?

This could be reduced down to `210` and still fit the magnitude of numbers the original complainant cited, if the feeling is that `220` is "too much".

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -376,7 +376,7 @@ LRESULT Notepad_plus::init(HWND hwnd)
     //--Status Bar Section--//
 	bool willBeShown = nppGUI._statusBarShow;
     _statusBar.init(_pPublicInterface->getHinst(), hwnd, 6);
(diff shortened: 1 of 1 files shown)


donho:
220 is good - complain no more.

Issue #9004, opened 2020-10-13
Replace wont replace with regex lookbehind assertion


marko-savic:
### Description of the Issue
Replace will not actually replace the character.

### Steps to Reproduce the Issue
1. Open new txt file
2. Add two lines fo example: 
2020-08-31-09:24:26,5.23,4.22,3.64,26
2020-08-31-09:24:27,0.00,0.00,0.00,26
3. Open replace with regex search, try this assertion: Find What: "(?<=:..)," and Replace With: "-"

### Expected Behavior
Matched character ',' is replaced with the character '-'

### Actual Behavior
A match occurs but no actual replacing is done

### Debug Information

Notepad++ v7.9   (64-bit)
Build time : Sep 22 2020 - 03:19:04
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1903
OS Build : 18362.1082
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
Some notes:

* The _Expected Behavior_ occurs if a _Replace All_ is done.
* The _Actual Behavior_ occurs if a step-by-step _Replace_ is attempted.

sasumner:
Duplicates/similars:  #713 #1870 #2216 #2360 #5688

Issue #4855, opened 2018-09-16
An alleged BUG in regular expression search


meirivry:
### Description of the Issue
a regular expression search for '\b' should find a word boundary, however it finds boundary between each individual characters.

### Steps to Reproduce the Issue
1. Type few words like "Hello world!!!".
2. Click CTL-F for search.
3. On the Search Mode select regular expression
4. On the "find what :" field type \b and click "find next" button several times.

### Expected Behavior
It should find the inter-character place before "H" of the "Hello world!!!", the place after the "o", the place before "w" and the place after "d".

### Actual Behavior
It finds all inter-character places.

sasumner:
It hits as you describe because the buffer it searches STARTS (when you Find-Next) at the caret position.  There is NOTHING to the left of the caret position (as far as the search is concerned).  Thus if there is a word character to the right of the caret position when you initiate a search you will get a hit, no matter what you can see to the left.  

So think about your insert caret between the "w" and the "o" in your example.  If you hit Find-Next here, it hits the "o" because there is NOTHING to the left of it (for searching purposes).

I'm not trying to be judge-and-jury on this being correct/incorrect behavior, just providing the WHY for existing behavior.

There is more discussion about this type of behavior here:  https://notepad-plus-plus.org/community/topic/16104/how-to-find-and-highlight-a-specific-occurance-of-a-symbol

Issue #2360, opened 2016-09-28
Find & Replace Broken w/ Regex Lookaround


sampalmer:
### Description of the Issue

When doing a Find & Replace using a regular expression with a lookahead or lookbehind, the "replace" button may fail to work.
### Steps to Reproduce the Issue

1. Open a blank Notepad++ tab
2. Insert the following text `,NULL,NULL`
3. Open the Find & Replace dialog
4. Put this in the find box: `(?<=,)NULL`
5. Set "Search Mode" to "Regular expression"
6. Press "Replace" multiple times
### Expected Behavior

The "Replace" button should result in the current match being replaced with the specified replacement text.
### Actual Behavior

The "Replace" button didn't perform any replacement, but instead moved to the next match.
### Debug Information

Notepad++ v7   (32-bit)
Build time : Sep 21 2016 - 22:00:56
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 10
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

Note that this doesn't occur when you press "Replace All"; it's specific to the "Replace" button.

dail8859:
Duplicate of #2216

PR #8991, opened 2020-10-11
Split functionList.xml into multifile in functionList Folder

donho:
New "functionList" will be added beside of notepad++.exe or "%APPDATA%\Notepad++\", according the installation mode.
All splited files are localized in this folder with the explicit language name.

"overrideMap.xml" is optional for overriding the default functionList parse rule files, and for adding UDL parse rule files.

Close #4896

PR #9008, opened 2020-10-14
Fix RegEx look behind operations

### LINKED ISSUE #713: Replacing from my regex find, does nothing
I'm using the regex `(?<=\d)"(?=\s)` to find the quote marks at the end of a number in a tab/newline delimited file. In the find and replace dialog, when I click on <kbd>Replace</kbd> my selection is advanced to the next find result, but the previously selected " isn't replaced. My replace value seems irrelevant, I started with no replace value (removing " ) and tried using some random keysmash with no different results. 

Notepad++ v6.7.9.2

### LINKED ISSUE #8434: Regular expression containing \K does not work with step-by-step Replace
### Description of the Issue
When walking thru replacements using the _Replace_ button, regular expressions that use the `\K` syntax in the _Find what_ field are _NOT_ replaced when matches clearly occur.

1.  Create some text in a new tab; note the blank line at the top:
```

Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```
2. Put your caret at the first position in the file, on line 1
3. Press Ctrl+h to invoke the _Replace_ dialog window
4. Set up the following:
image
5. Press _Replace_ and observe the editor window displays the next occurring match by selecting it:
image
6. Note that due to the `\K` syntax, the part of the match before the `\K` is "forgotten about" and not included in the selection.  This seems OK as that is what `\K` is supposed to provide.
7. Press _Replace_

### Expected Behavior
I expected the `ck` that was selected in line 2 (as shown in the screenshot of step 5 above) to be replaced with `CK`.

### Actual Behavior
No replacement was made on line 2 and the editor window now looks like this (following match is selected):
image

### Debug Information
Notepad++ v7.8.7   (32-bit)
Path : C:\\..................\npp.7.8.7.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039
Current ANSI codepage : 1252

### Other

Note that the replacements will be made if _Replace All_ is used, but that's not what this issue is about.

### Some references that discuss this behavior of `\K` with step-by-step Replace

* https://community.notepad-plus-plus.org/topic/19570/how-to-delete-text-after-the-2nd-specific-character-in-each-row
* https://community.notepad-plus-plus.org/topic/19417/replace-specific-word-between-words-in-string/
* https://community.notepad-plus-plus.org/topic/19417/replace-specific-word-between-words-in-string/
* https://community.notepad-plus-plus.org/topic/76/marking-a-large-text

Uhf7:
Intended to fix #713, #1870, #2360 and #9004.

Does not fix #8434.

_Edit:_ should solve #4855 too, collides with PR #8926.

The modification cannot be verified in the AppVeyor build, because the AppVeyor build does not contain the Boost library. The modification cannot be applied to the build-in RegEx, because this doesn't support look-behind operations.

The modification itself is self-explanatory: If we want RegEx to look behind, then we need to specify a backstop position.

_Edit:_ I used Boost version 1_73_0 for this, may be, there are older Boost versions which don't have this parameter.

_Edit2:_ Just re-tested it with Boost version 1_70_0, it works there too. There are no significant code changes between 1_70_0 and 1_73_0 in the RegEx part of Boost, so that an upgrade to 1_73_0 would have no advantages.

### DIFF
--- a/scintilla/boostregex/BoostRegExSearch.cxx
+++ b/scintilla/boostregex/BoostRegExSearch.cxx
@@ -288,13 +288,8 @@ Sci::Position BoostRegexSearch::FindText(Document* doc, Sci::Position startPosit
 			regex_constants::ECMAScript
 			| (caseSensitive ? 0 : regex_constants::icase);
 		search._regexString = regexString;
-		
-		const bool starts_at_line_start = search.isLineStart(search._startPosition);
-		const bool ends_at_line_end     = search.isLineEnd(search._endPosition);
 		search._boostRegexFlags = 
-			  (starts_at_line_start ? regex_constants::match_default : regex_constants::match_not_bol)
-			| (ends_at_line_end     ? regex_constants::match_default : regex_constants::match_not_eol)
-			| ((sciSearchFlags & SCFIND_REGEXP_DOTMATCHESNL) ? regex_constants::match_default : regex_constants::match_not_dot_newline);
+			((sciSearchFlags & SCFIND_REGEXP_DOTMATCHESNL) ? regex_constants::match_default : regex_constants::match_not_dot_newline);
 		
 		const int empty_match_style = sciSearchFlags & SCFIND_REGEXP_EMPTYMATCH_MASK;
 		const int allow_empty_at_start = sciSearchFlags & SCFIND_REGEXP_EMPTYMATCH_ALLOWATSTART;
@@ -344,15 +339,13 @@ template <class CharT, class CharacterIterator>
 BoostRegexSearch::Match BoostRegexSearch::EncodingDependent<CharT, CharacterIterator>::FindTextForward(SearchParameters& search)
 {
 	CharacterIterator endIterator(search._document, search._endPosition, search._endPosition);
+	CharacterIterator baseIterator(search._document, 0, search._endPosition);
 	Sci::Position next_search_from_position = search._startPosition;
 	bool found = false;
 	bool match_is_valid = false;
 	do {
-		search._boostRegexFlags = search.isLineStart(next_search_from_position)
-			? search._boostRegexFlags & ~regex_constants::match_not_bol
-			: search._boostRegexFlags |  regex_constants::match_not_bol;
 		const bool end_reached = next_search_from_position > search._endPosition;
-		found = !end_reached && boost::regex_search(CharacterIterator(search._document, next_search_from_position, search._endPosition), endIterator, _match, _regex, search._boostRegexFlags);
+		found = !end_reached && boost::regex_search(CharacterIterator(search._document, next_search_from_position, search._endPosition), endIterator, _match, _regex, search._boostRegexFlags, baseIterator);
 		if (found) {
 			const Sci::Position  position = _match[0].first.pos();
 			const Sci::Position  length   = _match[0].second.pos() - position;
--- a/PowerEditor/installer/functionList/python.xml
+++ b/PowerEditor/installer/functionList/python.xml
@@ -16,13 +16,13 @@
 			commentExpr="(?s:'''.*?''')|(?m-s:#.*?$)"
 		>
 			<classRange
-				mainExpr    ="(?&lt;=^class\x20).*?(?=\n\S|\Z)"
+				mainExpr    ="^class\x20\K.*?(?=\n\S|\Z)"
 			>
 				<className>
 					<nameExpr expr="\w+(?=\s*[\(|:])" />
 				</className>
 				<function
-					mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
+					mainExpr="\sdef\x20\K.+?(?=:)"
 				>
 					<functionName>
 						<funcNameExpr expr=".*" />
@@ -30,7 +30,7 @@
 				</function>
 			</classRange>
 			<function
-				mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
+				mainExpr="\sdef\x20\K.+?(?=:)"
 			>
 				<functionName>
 					<nameExpr expr=".*" />
--- a/PowerEditor/installer/functionList/ruby.xml
+++ b/PowerEditor/installer/functionList/ruby.xml
@@ -16,7 +16,7 @@
 		>
 			<!-- within a class-->
 			<classRange
-				mainExpr    ="(?&lt;=^class\x20).*?(?=\n\S|\Z)"
+				mainExpr    ="^class\x20\K.*?(?=\n\S|\Z)"
 			>
 				<className>
 					<nameExpr expr="\w+" />
@@ -25,7 +25,7 @@
 					mainExpr="^\s*def\s+\w+"
 				>
 					<functionName>
-						<funcNameExpr expr="(?&lt;=def\s)\w+" />
+						<funcNameExpr expr="def\s\K\w+" />
 					</functionName>
 				</function>
 			</classRange>
@@ -34,7 +34,7 @@
 				mainExpr="^\s*def\s+\w+"
 			>
 				<functionName>
-					<nameExpr expr="(?&lt;=def\s)\w+" />
+					<nameExpr expr="def\s\K\w+" />
 				</functionName>
 			</function>
 		</parser>


scintilla/boostregex/BoostRegExSearch.cxx
```diff
@@ -352,7 +353,7 @@ BoostRegexSearch::Match BoostRegexSearch::EncodingDependent<CharT, CharacterIter
 			? search._boostRegexFlags & ~regex_constants::match_not_bol
 			: search._boostRegexFlags |  regex_constants::match_not_bol;
 		const bool end_reached = next_search_from_position > search._endPosition;
-		found = !end_reached && boost::regex_search(CharacterIterator(search._document, next_search_from_position, search._endPosition), endIterator, _match, _regex, search._boostRegexFlags);
+		found = !end_reached && boost::regex_search(CharacterIterator(search._document, next_search_from_position, search._endPosition), endIterator, _match, _regex, search._boostRegexFlags, baseIterator);
```
sasumner, line 356: This is interesting.  I don't see this form of the call to `regex_search` in this documentation?: 
https://www.boost.org/doc/libs/1_74_0/libs/regex/doc/html/boost_regex/ref/regex_search.html
What am I missing here?

Uhf7: In the documentation, you missed nothing, as far as I see. It seems to be outdated or incomplete.

I came to the solution by examining the source code.

sasumner:
How can we know what other regex behavior this might break?

Here's a (admittedly bad) example:

* Searching for `\A` with the code from this PR finds no hits.  Ideally, it would find one zero-length match at start of file (I think).

* Searching for `\A` with the code without this PR finds A LOT hits.  Too many, in fact (one at every position of the file).

But the fact that the behavior is different makes me wonder about other constructs...

As we have no regression-testing code for regex operations, do we adopt a "it fixes something, so we just go with it" attitude?  (Not saying that is right, or wrong).  Just asking.

Uhf7:
Can you explain your example a little more detailed, please? You did search what? In what? And expected what? I'm missing the data here to reproduce it.

Uhf7:
OK, made up my mind: It was a good example. It is fixed now too.

sasumner:
>It was a good example. It is fixed now too.

Sorry I didn't get back to you sooner.  I had to sleep, well, at least a little bit, before getting back to what's important. :-)

I still don't think it was a "good" example, because `\A` was "broken" before.
I am interested to see if it is fixed (or less, or equivalently broken) now.
I will test later...

Uhf7:
> do we adopt a "it fixes something, so we just go with it" attitude?

No. The decision has been made to use boost library for the regex operations. And the only thing I'm trying here is to allow the boost library to work as expected by giving it the parameters it needs. And the improvements are obvious.

sasumner:
My point was, without any real way of regression testing regex changes, how do we know that an improvement in one area doesn't change behavior unexpectedly in some other area (of regex search/replace).  I already found one example of this with `\A` working differently with the change, but that was just some playaround poking I was doing...

Uhf7:
Fun fact: I just realized, that this PR solves #4855 too.

sasumner:
> Fun fact: I just realized, that this PR solves 4855 too.

It's more than just a fun-fact:
Since 4855 is currently "open", you should add it to your `Intended to fix 713, 1870, 2360 and 9004.` line at the top of the PR, so that when it gets accepted, Don will close it.
There are probably a few (or several) more issues that this could close.

It is too bad we didn't know about this somewhat-secret (not in docs) parameter a long time ago!

Uhf7:
Preparing my next iteration. I'm quite certain now, that the `match_not_bol` flag is not needed anymore, it was needed only, because regex wasn't allowed to look behind and handled the start position of the search like the start of a line. Which was unhealthy for searches starting mid-line. With the baseIterator, regex can look behind to see whether it is at the beginning of a line or not, hence the toggling of `match_not_bol` should be obsolete.
--

donho:
@sasumner 
> My point was, without any real way of regression testing regex changes, how do we know that an improvement in one area doesn't change behavior unexpectedly in some other area (of regex search/replace).

We have talked about Unit-test on RegExpr, without conclusion/solution. But I just thought of Unit-test of Function List - the several parsers are used RegExpr for building Function List, they may not cover all the case but we can add lexers for UDL, so we can extend to cover more case. So it will be the solution.

The only downside is that it won't be automatic, since the SciLexer.dll in Appveyor doesn't contain boost's RegExpr. But we can launch them manually with our local build Scilexer.dll to make sure there's no regression.

donho:
@Uhf7 There's a good news and a bad news for your PR. The good news is almost all unit-tests pass. The bad news is not for the python parser.

Please paste the following code to a new doc, set it to Python, then launch Function list panel. 
You will see under `SampleClassBad`, there are 2 items of `__init__ (self, fileName)`. Without applying this PR, there was only 1 item.
```
INPUT_FILE_NAME = "example.txt"

class SampleClassBad (object):
    # WrongClassName: this will appear in the function list

    def __init__ (self, fileName):
        self.fileName = fileName


class SampleClassGood(object):
    # this works as expected: notice the missing space between
    # the class name and the left parenthesis.

    def __init__ (self, fileName):
        self.fileName = fileName




```

Of course, it could be the parser's problem. Here's the Python parser:
```
<parser
	displayName="Python"
	id         ="python_syntax"
	commentExpr="(?s:'''.*?''')|(?m-s:#.*?$)"
>
	<classRange
		mainExpr    ="(?&lt;=^class\x20).*?(?=\n\S|\Z)"
	>
		<className>
			<nameExpr expr="\w+(?=\s*[\(|:])" />
		</className>
		<function
			mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
		>
			<functionName>
				<funcNameExpr expr=".*" />
			</functionName>
		</function>
	</classRange>
	<function
		mainExpr="(?&lt;=\sdef\x20).+?(?=:)"
	>
		<functionName>
			<nameExpr expr=".*" />
		</functionName>
	</function>
</parser>

```

Can you see any problem in the parser ?

sasumner:
@donho said:

> We have talked about Unit-test on RegExpr, without conclusion/solution

Yes, that was here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8926#issuecomment-701398373

> Unit-test of Function List - RegExpr are used by the different language parsers for building Function List, they may not cover all the case...So it will be the solution.

It is better than nothing, so why not indeed.

>since the SciLexer.dll in Appveyor doesn't contain boost's RegExpr.

Can we change that?  See here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8900#issuecomment-711081909

Uhf7:
I can verify the problem, using the following example:

	INPUT_FILE_NAME = "example.txt"

	class SampleClassBad (object):
	    # Class functions detected twice

	    def __init1__ (self, fileName):
	        self.fileName = fileName

	    def __init2__ (self, fileName):
	        self.fileName = fileName


	class SampleClassSad(object):
	    # Class functions detected twice

	    def __init3__ (self, fileName):
	        self.fileName = fileName

	    def __init4__ (self, fileName):
	        self.fileName = fileName


	class SampleClassMad(object):
	    # Class functions detected twice

	    def __init5__ (self, fileName):
	        self.fileName = fileName

	    def __init6__ (self, fileName):
	        self.fileName = fileName


	class SampleClassGood (object):
	    # seems OK, but adding one non-whitespace character after it, the class functions here are detected twice too.

	    def __init7__ (self, fileName):
	        self.fileName = fileName

	    def __init8__ (self, fileName):
	        self.fileName = fileName

I'm going to investigate further next weekend (3 days ahead).

Uhf7:
Now I found out what happens with the Python parser. It is indeed the only parser using regex look-behind operations, and the `functionParser.cpp` is made in a way, that it works only, when the look-behind operations don't really look behind. 

But the regex expressions of the Python parser can be replaced easily by there `\K` equivalents, then the parser works as before with and without the rest of this PR.

----

It's a little bit complicated to explain what happens, I try it anyway. The function `functionMixParser::parse` is, where it begins: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/77d460696703731d2fc0036884f87de760a777ab/PowerEditor/src/WinControls/FunctionList/functionParser.cpp#L715-L739

The first `classParse` call in it searches for the class regex `(?<=^class\x20).*?(?=\n\S|\Z)` and finds:

	sample source                                 find results
	-----------------------------------           ------------------------------------
	class SampleClass1 (object):            Hit1: SampleClass1 (object):
	    def __init1__ (self, fileName):               def __init1__ (Self, fileName):
	        self.fileName = fileName                      self.fileName = fileName
	
	class SampleClass2(object):		Hit2: SampleClass2 (object):
	    def __init2__ (self, fileName):               def __init2__ (self, fileName):
	        self.fileName = fileName                      self.fileName = fileName

Notice, that both find results do not contain `class` anymore. In these two find results, the `classParse` function searches out successfully the class name and all function names. So far, so good.

Then comes the loop commented as `// the second level`. The `classParse` call in line 726 is applied to all the find results of the first call. So imagine, the expression `(?<=^class\x20).*?(?=\n\S|\Z)` is applied again to the find result "Hit 1" from above:
- As long as regex doesn't really look behind, nothing is found.
- If regex really looks behind, then it sees before the word `SampleClass1` still the word `class`, because the find result is still within it's original buffer. So it finds the same class again and inserts the same class functions again.

This is, why we cannot use behind operations in the regex expression of this parser.

----

The last question is: Why are the functions not duplicated in the last class? This is the result of a combination of effects.
- This break in  `FunctionZoneParser::classParse` https://github.com/notepad-plus-plus/notepad-plus-plus/blob/77d460696703731d2fc0036884f87de760a777ab/PowerEditor/src/WinControls/FunctionList/functionParser.cpp#L598 has the following consequence: All scanned zones are passed to `funcParse`,  except if the end of the scanned zone matches the `end` parameter of the `classParse` function. In this case, this scanned zone is not passed to `funcParse`.
- The `classParse` function is called with start=0 and end=236, in the first level, and with start=142 and end=234 in the second level, when parsing `SampleClass2` again.

If end=236, the `break` is not triggered, if end=234, it breaks.

This last effect seems only to affect parsers which have no open/close symbols, so I don't thin there is immediate danger from here.

Uhf7:
OK, it doesn't merge automatically, but the green lines in the `functionList.xml` work.

ArkadiuszMichalski:
@Uhf7 `functionList.xml` was split https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8991.

Uhf7:
While checking the new function list xml's, I saw, that ruby might be affected too. But it passes the unit test.

scintilla/boostregex/BoostRegExSearch.cxx
```diff
@@ -292,9 +292,7 @@ Sci::Position BoostRegexSearch::FindText(Document* doc, Sci::Position startPosit
 		const bool starts_at_line_start = search.isLineStart(search._startPosition);
 		const bool ends_at_line_end     = search.isLineEnd(search._endPosition);
```
MAPJe71, line 293: AFAICT these bools are no longer used, so lines can be removed.

MAPJe71 (changes requested):

scintilla/boostregex/BoostRegExSearch.cxx
MAPJe71, line 293: AFAICT these bools are no longer used, so lines can be removed.

MAPJe71 (changes requested):

donho:
@Uhf7 
> This is, why we cannot use behind operations in the regex expression of this parser.

Thank you for the very detail information. A notice will be added into npp-usermaual:
https://github.com/notepad-plus-plus/npp-usermanual/pull/120

I would like to know if you have finish tuning this PR? Is it ready to merge or you'll add something else?

Uhf7:
OK, I would change the look behind expression in ruby too, to leave no "negative" example behind. And I would change or leave out the notice in the user manual, because it is not true. They work now. The only thing is, that they have been used in a wrong way, because they didn't work until now. When a new function list language xml is written in the new environment, then this will not happen again. And there are negative look behind expressions `(?<!xxx)` too, which will not run into this problem.

Uhf7:
Now I have no further tuning for this PR.

donho:
Thank you @Uhf7 !

sasumner:
@Uhf7 Is the guidance to use `\K` in future function-list work, instead of the positive-lookbehind construct?  I'm unclear as to if it (positive-lookbehind) was used in the Python parser wrongly or as was simply a side-effect of how this construct worked (before this PR) in N++.

Uhf7:
@sasumner,

it seems to be complicated. Before being able to describe the various regex constructs which can be used or not, we would need a more detailed description about how the function list parser works. And this is a hard thing to do. The problem for the Python parser was the second level search for classes in `FunctionMixParser::parse`. There is nothing mentioned about this in FunctionList.md. It seems not to be needed for current Python parser, but perhaps for other languages.

What we can say generally, is, that the parser divides the parsed code into sections, and within this sections, other regex expressions are applied to divide them further. Since the sections are not copied out of the buffer to divide them into sub-sections, but being parsed in place instead, look-behind operations may work in an unexpected way. But such a text could raise more questions than it answers too.

Issue #1499, opened 2016-02-18
Single line of output to Find-result window for lines with 1+ hits


DanielTheGreat:
I don't use it often, but I like Notepad++ for the occasional coding (hobby) project!  Just one little thing slightly annoys me, so I wondered whether others would agree to a small change to the way the  'Find All in...' feature operates...

I sometimes use 'Find All in All Opened Documents' (or the other one) to find all occurrences of a particular string that I want to check and/or edit.  But if I search for a string (perhaps a variable or register name) that occurs twice (or more) in one line, the search results pane shows that line twice (or more).  This simply clutters the search results, and seems to serve no useful purpose.

Here's an example:
# define    vca25    vca2_addr.bit.b5    // Voltage detection 0 enable bit

If I'm searching for 'vca2' to find everywhere the register is accessed, I don't need to see that line appear twice in the results;  I just want to know every file and every line where that register appears.

Would it be practical (and acceptable to most users) to change the 'Find All' code such that when a string is identified in a particular file/line, **that** line is then skipped from further searching so it appears only once in the results pane?

I realise this will complicate the code a little, since I presume there is much common code between 'Find All in...' and 'Replace All in...' functions, whereas for 'Replace All in...' you **do** want every instance found, not just the first per line!

Thanks, Daniel
(Gerroa, Australia)

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your support and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
search if related topics exists. you are welcome to post either at similar
topics, or to create a new topic at Help Wanted or General Discussion.
(no extra account is needed, just use your github account to sign in)
```
[...]

PR #9014, opened 2020-10-15
Fix find result line length

### LINKED ISSUE #8525: 'Find in Files': Not possible to jump to every single hit in certain cases

### LINKED ISSUE #9009: Often nothing happens when double clicking on find in files results
### Description of the Issue

When I do a normal "find in files" search often in the results window clicking on a result line does nothing.

1. Put the attached file in an empty directory
2. Load it
3. Open “Find in files”
4. Search for `stupiscono` in *.txt (“Normal” search in all subfolders, no other checked option)
5. In the search results window you have one result line from the file (see attached screenshot)
6. Double click on it. Nothing happens.

### Expected Behavior
The editor cursor goes to line containing `stupiscono`

### Actual Behavior
Nothing happens.
If you delete lines after the matching line suddenly the double click works again. Deleting lines before the matching line seems have no effect.
In the results window, if you collapse the Search "stupiscono" section, the matching line is not collapsed and remains visible, as if it is not part of this section.

### Debug Information
Notepad++ v7.9 (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2004
OS Build : 19041.508
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll DSpellCheck.dll HexEditor.dll HTMLTag_unicode.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll NppFavorites.dll NppFTP.dll NPPJSONViewer.dll NppSaveAsAdmin.dll NppTextFX.dll XMLTools.dll _CustomizeToolbar.dll

The file to reproduce the problem: [bug.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5378450/bug.txt)

image

Uhf7:
Fixes #9009
Edit by @sasumner : Fixes #8525 as well!

The reason for the effect was, that the search result parser cannot handle lines longer than `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` characters. The length limitation of the displayed found line has been made in wide chars instead of in UTF-8 chars. When it was converted into UTF-8 after this, it was getting actually longer than the limit.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -32,6 +32,7 @@
 #include "Notepad_plus_msgs.h"
 #include "UniConversion.h"
 #include "localization.h"
+#include "Utf8.h"
 
 using namespace std;
 
@@ -2158,15 +2159,15 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 				int nbChar = lend - lstart;
 
 				// use the static buffer
-				TCHAR lineBuf[1024];
+				TCHAR lineBuf[SC_SEARCHRESULT_LINEBUFFERMAXLENGTH];
 
-				if (nbChar > 1024 - 3)
-					lend = lstart + 1020;
+				if (nbChar > SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - 3)
+					lend = lstart + SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - 4;
 
 				int start_mark = targetStart - lstart;
 				int end_mark = targetEnd - lstart;
 
-				pEditView->getGenericText(lineBuf, 1024, lstart, lend, &start_mark, &end_mark);
+				pEditView->getGenericText(lineBuf, SC_SEARCHRESULT_LINEBUFFERMAXLENGTH, lstart, lend, &start_mark, &end_mark);
 
 				generic_string line = lineBuf;
 				line += TEXT("\r\n");
@@ -2191,15 +2192,15 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 				int nbChar = lend - lstart;
 
 				// use the static buffer
-				TCHAR lineBuf[1024];
-
-				if (nbChar > 1024 - 3)
-					lend = lstart + 1020;
+				TCHAR lineBuf[SC_SEARCHRESULT_LINEBUFFERMAXLENGTH];
+				
+				if (nbChar > SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - 3)
+					lend = lstart + SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - 4;
 
 				int start_mark = targetStart - lstart;
 				int end_mark = targetEnd - lstart;
 
-				pEditView->getGenericText(lineBuf, 1024, lstart, lend, &start_mark, &end_mark);
+				pEditView->getGenericText(lineBuf, SC_SEARCHRESULT_LINEBUFFERMAXLENGTH, lstart, lend, &start_mark, &end_mark);
 
 				generic_string line = lineBuf;
 				line += TEXT("\r\n");
@@ -3537,14 +3538,25 @@ void Finder::add(FoundInfo fi, SearchResultMarking mi, const TCHAR* foundline)
 	mi._end += static_cast<int32_t>(str.length());
 	str += foundline;
 
-	if (str.length() >= SC_SEARCHRESULT_LINEBUFFERMAXLENGTH)
+	WcharMbcsConvertor& wmc = WcharMbcsConvertor::getInstance();
+	const char *text2AddUtf8 = wmc.wchar2char(str.c_str(), SC_CP_UTF8, &mi._start, &mi._end); // certainly utf8 here
+	size_t len = strlen(text2AddUtf8);
+
+	if (len >= SC_SEARCHRESULT_LINEBUFFERMAXLENGTH)
 	{
-		const TCHAR * endOfLongLine = TEXT("...\r\n");
-		str = str.substr(0, SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - lstrlen(endOfLongLine) - 1);
-		str += endOfLongLine;
+		const char * endOfLongLine = " ...\r\n"; // perfectly Utf8-encoded already
+		size_t lenEndOfLongLine = strlen(endOfLongLine);
+		size_t cut = SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - lenEndOfLongLine - 1;
+
+		while ((cut > 0) && (!Utf8::isValid(& text2AddUtf8 [cut], (int)(len - cut))))
+			cut--;
+
+		memcpy ((void*) & text2AddUtf8 [cut], endOfLongLine, lenEndOfLongLine + 1);
+		len = cut + lenEndOfLongLine;
 	}
+
 	setFinderReadOnly(false);
-	_scintView.addGenericText(str.c_str(), &mi._start, &mi._end);
+	_scintView.execute(SCI_ADDTEXT, len, reinterpret_cast<LPARAM>(text2AddUtf8));
 	setFinderReadOnly(true);
 	_pMainMarkings->push_back(mi);
 }


PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -2158,15 +2159,15 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 				int nbChar = lend - lstart;
 
 				// use the static buffer
-				TCHAR lineBuf[1024];
+				TCHAR lineBuf[SC_SEARCHRESULT_LINEBUFFERMAXLENGTH];
 
-				if (nbChar > 1024 - 3)
-					lend = lstart + 1020;
+				if (nbChar > SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - 3)
+					lend = lstart + SC_SEARCHRESULT_LINEBUFFERMAXLENGTH - 4;
```
sasumner, line 2165: @Uhf7 
You eliminated the `1024` "magic number" (good!).  
How about doing something better for the `3` and the `4` magic numbers as well?
I'm fairly certain that `3` refers to the length of `...`, but that's because I'm familiar with this code (future code readers may not know this).
Perhaps the `4` should not be used anyway -> it should be `- 3 - 1` (before `3` is eliminated)

sasumner:
Just as a semi-related aspect, there have been user-requests for increasing the max. _Find result_ line length, e.g. #7723

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
Uhf7: Actually, I'm not even sure if there should be a new constant for this. Reason: The limit in wide chars determines, how many characters we want to see in the Find Result window. The `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` constant says, how many UTF-8 characters the parser can handle. And between the new constant and the `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH`  constant should be the factor 4.

Uhf7:
Actually, the line in the Find result window shouldn't be cut at the end. It should be cut around the actual find result within the line. But this is too much change at once, I wanted only make the double click working for long results too.

sasumner:
> ...shouldn't be cut at the end. It should be cut around the actual find result...

Isn't this an arguable point?

Maybe user wants to find some text that is about at the same column in each matched line and then quickly see what is ~150 columns to the right of that -- currently user can position the horizontal scroll bar in the _Find result_ window, then scroll vertically to examine such data.
That is just an example -- maybe it is too contrived? :-)

sasumner:
@Uhf7 
Good points, especially:

>If a find result line is cut, then the marked result in it gets invisible, if it is beyond the cutting point.

Being invisible isn't desired, probably all can agree on that.

For the future, I wonder if it is better to:

* allow full "source" lines in the _Find result_ (back to #7723 again, with "unlimited" increase)
OR
* "cut" but always keep the match text visible (as much as possible -- match text could be longer than the size limit)

On a personal note, as a N++ user, I'm glad that I don't often deal with really long lines.

Uhf7:
I always imagine a line like "abc abc abc abc", with so many "abc"s in it, that the system is at the limit already. Then, I search for "abc" (Find all in current document). And then, this ultra-long line gets multiplied in the Find Result window. That's why I would vote against an unlimited size. But I'm for increasing the limit to, lets say 2047 wide characters and 8191 UTF-8 characters. I would go much further. In the moment, we cannot increase the size because the `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` constant crosses the boundary to the SciLexer.dll. It can only be done in a new release.

sasumner:
> In the moment, we cannot increase the size because the SC_SEARCHRESULT_LINEBUFFERMAXLENGTH constant crosses the boundary to the SciLexer.dll

Yes, I gave some discussion of this in #7723

>I always imagine a line like "abc abc abc abc", with so many "abc"s in it, .... Then, I search for "abc" (Find all in current document). And then, this ultra-long line gets multiplied in the Find Result window.

Yes, that is unliked by several users as well, see #1499

sasumner:
>How to mark the multiple results in the line, especially, when they overlap

They can't overlap, but perhaps you mean "are contiguous" as in `abcabcabcabc` and you search for `abc`

>But this all is not the purpose of this PR.

Right, I'll stop now. :-)

PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
```diff
@@ -32,6 +32,7 @@
 #include "Notepad_plus_msgs.h"
 #include "UniConversion.h"
 #include "localization.h"
+#include "../Utf8.h"
```
sasumner, line 35: The `../` part not needed here, I think.

Uhf7: I copied it from Common.cpp, of course, without thinking about it. Removing the `../` from the FindReplaceDlg.cpp ...

donho:
@Uhf7 
> The reason for the effect was, that the search result parser cannot handle lines longer than SC_SEARCHRESULT_LINEBUFFERMAXLENGTH characters. The length limitation of the displayed found line has been made in wide chars instead of in UTF-8 chars. When it was converted into UTF-8 after this, it was getting actually longer than the limit.

In this case, increase value of `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` to 4096 will  fix #9009, won't it?
Since we're talking about increasing  `SC_SEARCHRESULT_LINEBUFFERMAXLENGTH` in #7723.

sasumner:
>increase value of SC_SEARCHRESULT_LINEBUFFERMAXLENGTH to 4096 will fix 9009 ?

No, changing to 4096 only won't fix the general problem.
Something like this change is needed (although I have not finished my own review of this yet)

donho:
@sasumner 
> No, changing to 4096 only won't fix the general problem.
There's no way to solve general problem - the line length limit is non-exist generally.

@Uhf7 
> The reason for the effect was, that the search result parser cannot handle lines longer than SC_SEARCHRESULT_LINEBUFFERMAXLENGTH characters. The length limitation of the displayed found line has been made in wide chars instead of in UTF-8 chars.

For me, `str.length()` should return the same length as `strlen(text2AddUtf8)`
Your code does fix the bug but I still don't know how.

donho:
@Uhf7 
OK I see.
Just display the length of `str.length()` & `strlen(text2AddUtf8)`, the values are 1021 & 1031 respectively.
wstring::length() indeed less accurate then strlen(char*).
Since SCI_ADDTEXT take UTF8 `char*`, it'll be more accurate to count the length with `char*` directly.

Uhf7:
The problem still remaining is, that a find result of 1024 wide characters can expand to up to 4096 utf8 characters. And we display maximal 1024 utf8 characters. This has an obvious effect:

- If the search result line consists of 1024 characters `a`, then we see about 1000 characters `a` in the find result window.
- If the search result line consists of 1024 characters `ä`  then we see about 500 characters `ä` in the find result window. 
- If the search result line consists of 1024 characters `ạ`  then we see about 300 characters `ạ` in the find result window.

So actually, we would need two different constants, one for the maximal length in wide characters, describing how many characters we wish to see in the find result window, and a second one for the maximal length in utf8 characters. But this can only be done in the moment of a new release of the `SciLexer.dll`.

And, as already said , I wouldn't make it unlimited (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9014#issuecomment-709405971).

sasumner:
>we would need two different constants

This could certainly be a future development effort.
I think doubling the buffer size for now ( as described at end of #7723 comments) is a reasonable short-term compromise for the users that have reported having a problem with this.

sasumner:
@Uhf7 I'll add one more bullet point to what you started:

* If the search result line consists of 1024 characters 𒊹 then we see about 250 characters 𒊹 in the find result window.

PR #9024, opened 2020-10-17
Fix possible file corruption on file open for writing

### LINKED ISSUE #5664: File is erased on save when there is no free space left on disk

pnedev:
Microsoft documentation for reference:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=vs-2019
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fseek-fseeki64?view=vs-2019

Fix #5664

### DIFF
--- a/PowerEditor/src/Utf8_16.cpp
+++ b/PowerEditor/src/Utf8_16.cpp
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "Utf8_16.h"
+#include <io.h>
 
 const Utf8_16::utf8 Utf8_16::k_Boms[][3] = {
 	{0x00, 0x00, 0x00},  // Unknown
@@ -276,8 +277,9 @@ Utf8_16_Write::Utf8_16_Write()
 	m_eEncoding = uni8Bit;
 	m_pFile = NULL;
 	m_pNewBuf = NULL;
-	m_bFirstWrite = true;
 	m_nBufSize = 0;
+	m_bFirstWrite = true;
+	m_bWriteFailure = false;
 }
 
 Utf8_16_Write::~Utf8_16_Write()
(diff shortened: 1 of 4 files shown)


pnedev:
@donho ,
This PR is based on https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8992 because I believe it is the right approach to deal with this problem but I tested it with Notepad++ 7.9 32bit on Win10 (64bit) and on my setup behaves exactly as supposed to.

`_chsize_s` is called ONLY when it needs to be called (when the file is shortened/shrunk).

PowerEditor/src/Utf8_16.cpp
```diff
@@ -442,6 +457,15 @@ void Utf8_16_Write::fclose()
 
 	if (m_pFile)
 	{
+		if (!keepOriginalLength && !m_bWriteFailure)
```
donho, line 460: > _chsize_s is called ONLY when it needs to be called (when the file is shortened/shrunk).

No, it's not what I want. As what I said, _chsize_s should be called only after writing on  disk failure.
This condition let all normal cases have this special treatment.

donho:
Also I tested this PR, it didn't fix the bug - the same result as the current one (before applying this PR).

PowerEditor/src/Utf8_16.cpp
pnedev: @donho ,

Calling `_chsize_s ` after the write failure is pointless (or very bad hack to say the least if your idea is to extend the file to its initial length after the fail).

pnedev:
@donho ,
 
> Also I tested this PR, it didn't fix the bug - the same result as the current one (before applying this PR).

Just curious how did you test it?
The PR perfectly works on my setup.

donho:
@pnedev 
1. I use VeraCrypt to create a small drive:

image


2. I added 3 larges JPG to make disk almost full, then I copied change.log, opened it in Notepad++, modified it then save it:

image


3. Then I copied the content, pasted it repeatedly and crazily: 

image

4. I saved it, then the following dialog appeared:

image


5. Answered No, another dialog came:

image


6. Answer No again. Reload the file from disk, here the result:

image

pnedev:
Thanks for the detailed description @donho .

I suppose it is not very "clean" test or one that can be fully trusted because we don't know for sure how VeraCrypt handles it's virtual drives.
Do you know what is the file system of the VeraCrypt drive?

Anyway, no matter what we discuss here you've made-up your mind that this PR's approach is dangerous, not normal or whatever. So I'll test one last thing and if it works I'll make a last PR trying to minimize the impact of this serious bug.

donho:
> I suppose it is not very "clean" test or one that can be fully trusted because we don't know for sure how VeraCrypt handles it's virtual drives.

It's the only way I found, sorry for not having a 3 MB USB flash drive in 2020.
OTOH, I don't see the difference once the drive is mounted by the system.

> Anyway, no matter what we discuss here you've made-up your mind that this PR's approach is dangerous, not normal or whatever.

As I said, and I say again: There's no reason to change current working code for fixing this bug. So let's treat only the special case.

PR #9030, opened 2020-10-18
Increase untruncated Find-result window line length to 2048

### LINKED ISSUE #7723: DOUBLE (edit: was REMOVE) the line-length limit in Find-result window

sasumner:
Fixes #7723


donho:
@sasumner I suppose you have check if it works and there's no regression?

sasumner:
@donho said:

>I suppose you have check if it works and there's no regression?

Well, I test all of my PRs. :-)
I ran a few simple tests on this one, seemed to work.
I did not exhaustively test it as it is "only" a length change.
A recent PR changed hardcoded occurrences `1024` (the old value for this) to use the `#define` name, so that doesn't seem to be a concern.
Do you have some other concerns?
If so, can you be specific?
I am willing to run more tests, I'm just not sure what they should be.

donho:
> Do you have some other concerns?
If so, can you be specific?

No no, I'm just lazy :)

> I am willing to run more tests, I'm just not sure what they should be.

I guess it's enough to test 1 or 2 case(s) to make sure it's working, and 1 or 2 usual case(s) to ensure there's no regression.

sasumner:
BTW, for the testing I did do, I concentrated on lines with A-z0-9 data and line lengths slightly less than the new threshold (to NOT see the `...`) and lines lengths slightly more than the threshold (to definitely see the `...`).

PR #9034, opened 2020-10-18
Improve URL update on resize

### LINKED ISSUE #9032: URL update on resize
URL update doesn't work in every case when the edit window height is increased. The first description of this effect I saw came from @Yaron10 in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8975#issuecomment-711171169.

### Description of the Issue
If there is any docked window above or below a document window, and the size of the document window is increased by moving the splitter between the document window and the other docked window, then the URLs in the new portion of the document window may not by displayed as URLs.

1. Create a test file with 50 lines in it, each line contains one URL, like https://github.com/notepad-plus-plus/notepad-plus-plus.
2. Open this test file with Notepad++.
3. Use the function "Find all" to make the Find results window appear. By default, this will appear below the document.
4. Resize the Find results window so that both, the document window and the Find results window, have about the same height.
5. Close the test file.
6. Open the test file again.
7. Resize the Find results window by dragging it's upper edge down, so that the Find result window height decreases and the document window height increases.

### Expected Behavior
The URLs, which become newly visible by the enlargement of the document window will be highlighted as URL.

### Actual Behavior
The URLs, which become newly visible by the enlargement of the document window are not highlighted.

### Debug Information
Notepad++ v7.9   (32-bit)
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll LuaScript.dll mimeTools.dll NppConverter.dll NppExport.dll 

This can be reproduced with older versions too (tried 7.8.5 for example).

Uhf7:
Fixes #9032. 

The reason for the effect is, that the `addHotSpot` calls for resizing are in the `WM_SIZE` of the application. This works, as long as the size of the whole application changes, but not, when the application size remains constant and only the size of the document window inside the application changes.

The solution for this is, to remove the `addHotSpot` calls from the `WM_SIZE` of the application and add them to the `DocTabView::reSizeTo` function instead. Since `addHotSpot` cannot be called directly from this function, I took the detour over `NPPM_INTERNAL_UPDATECLICKABLELINKS`.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -525,9 +525,6 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				_pDocMap->reloadMap();
 			}
 
-			addHotSpot(_pEditView);
-			addHotSpot(_pNonEditView);
-
 			result = TRUE;
 			break;
 		}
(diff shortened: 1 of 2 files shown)


Yaron10:
@Uhf7,

:+1: 
Very nice. Thank you for the quick fix.

PR #9046, opened 2020-10-22
fix #8783: incorrect icon for monitored file in tab bar dialog and vertical dir switch

### LINKED ISSUE #8783: Monitored file has incorrect icon in Ctrl+Tab switcher
### Steps to Reproduce the Issue (STR)
1. Have several tabs open in N++
2. For a disk file, turn monitoring on (press the "eye" icon in the toolbar).
3. Cycle through the files using Ctrl+Tab repeatedly.

### Expected Behavior

I expected to see the "eye" icon like the one used on the tab bar, for the monitored file.

### Actual Behavior

Icon used for the monitored file seems to be the "readonly" icon (grey diskette).

image

### Debug Information

Notepad++ v7.8.8   (64-bit)
Build time : Jun 23 2020 - 23:59:56
Path : C:\.........\npp.7.8.8.bin.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1339
Current ANSI codepage : 1252

### Additional Information

For color comparison, here's a cap of a non-monitored file (note the blue icon -- contrast with the grey icon in the earlier cap):

image

ErwanDelhoveUCL:
Fixes #8783

Small fixed of 8783
Awaiting comments! ^^

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -7256,7 +7256,7 @@ void Notepad_plus::minimizeDialogs()
 void Notepad_plus::restoreMinimizeDialogs()
 {
 	size_t nbDialogs = _sysTrayHiddenHwnd.size();
-	for (int i = (nbDialogs - 1); i >= 0; i--)
+	for (int i = (static_cast<int>(nbDialogs) - 1); i >= 0; i--)
 	{
 		::ShowWindow(_sysTrayHiddenHwnd[i], SW_SHOW);
 		_sysTrayHiddenHwnd.erase(_sysTrayHiddenHwnd.begin() + i);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -53,7 +53,7 @@
 
 using namespace std;
 
-enum tb_stat {tb_saved, tb_unsaved, tb_ro,tb_monitored};
+enum tb_stat {tb_saved, tb_unsaved, tb_ro, tb_monitored};
 #define DIR_LEFT true
 #define DIR_RIGHT false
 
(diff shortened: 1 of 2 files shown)


PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
```diff
@@ -222,7 +222,7 @@ void VerticalFileSwitcherListView::setItemIconStatus(BufferID bufferID)
 	LVITEM item;
 	item.pszText = fn;
 	item.iSubItem = 0;
-	item.iImage = buf->getUserReadOnly()||buf->getFileReadOnly()?2:(buf->isDirty()?1:0);
+	item.iImage = buf->getUserReadOnly()?3:buf->getFileReadOnly()?2:(buf->isDirty()?1:0);
```
donho, line 225: `(buf->getUserReadOnly() || buf->getFileReadOnly())` -> Both are read only, so it's the same icon.
You need use another way to detect monitoring status.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -53,7 +53,7 @@
 
 using namespace std;
 
-enum tb_stat {tb_saved, tb_unsaved, tb_ro};
+enum tb_stat {tb_saved, tb_unsaved, tb_ro,tb_monitored};
```
donho, line 56: Add a while space in front of tb_monitored

PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
ErwanDelhoveUCL: Do you think I should replace the int number with an enumeration or it isn't useful here?
I will make a method isReadOnly() like in the Notepad_plus.cpp.

ErwanDelhoveUCL: My bad the method isReadOnly() is already there ^^

donho: @ErwanDelhoveUCL

After applying your PR, when I set document to ReadOnly, the icon become monitoring:

image

donho (changes requested):
Keep  `buf->getUserReadOnly()||buf->getFileReadOnly()` for 2 and use `isMonitoringOn()` for detecting 3

donho (approved):

Issue #9031, opened 2020-10-18
Apostrophe in a url breaks clickable link creation


Gorrrg:
### Description of the Issue
When there's a url that contains an apostrophe, the parsing stops before the apostrophe which creates an incomplete clickable link.

### Steps to Reproduce the Issue
1. Paste a url into a text document, for example [https://en.wikipedia.org/wiki/Parkinson's_law](https://en.wikipedia.org/wiki/Parkinson's_law)
2. Click the partial link
3. You will end up at [https://en.wikipedia.org/wiki/Parkinson](https://en.wikipedia.org/wiki/Parkinson)

### Expected Behavior
The entire legitimate url should be incorporated into the clickable link

### Actual Behavior
Clickable link stops before the apostrophe.

### Debug Information
Notepad++ v7.9   (64-bit)
Build time : Sep 22 2020 - 03:19:04
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2004
OS Build : 19041.572
Current ANSI codepage : 65001
Plugins : JsMapParser.NppPlugin.dll JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll XMLTools.dll

Uhf7:
https://en.wikipedia.org/wiki/Murphy's_law ...

Yaron10:
@Uhf7,

Thank you for improving the parser. 
Great work. :+1:

dottoreCapitano:
In my python scripts, I sometimes have constructs like:
`  page = requests.get('https://www.tum.de/', headers=headers)`
Double-clicking the URL opens https://www.tum.de/' which gives an error in this case.
Do you have any idea how to realize a nice behaviour also in this situation?

Uhf7:
@dottoreCapitano 
Thank you for the post. The current URL parser works without looking back, which means, it disregards what character is before the link. So, seemingly clear decisions like "the URL is enclosed in quotes or double quotes" are in many cases not respected as expected. I tried to look back before, but there where massive critics in #8921 and #8975. So I abandoned this. But I still think it's a reasonable way to improve the URL detection.

[...]

Issue #9056, opened 2020-10-24
URLs not marked as links in unfocused view


Yaron10:
**STR:**
Open files `A` and `B` containing some 50 links.
(Do not scroll at this stage).
Move `B` to the other view.
Hover over the unfocused `A` and wheel-scroll down.

**Result:**
The lines revealed after scrolling down in `A` are not marked as links. 

![תמונה](https://user-images.githubusercontent.com/10229320/97081928-68e31d80-160e-11eb-8dc0-8037eae106e5.png)
***
Notepad++ v7.9   (32-bit)
Build time : Oct 23 2020 - 20:53:55
Path : C:\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

Yaron10:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b93f49a67791be6ed19c521b5e20bcb8efd375c2/PowerEditor/src/NppNotification.cpp#L855

Replacing that line with `addHotSpot(notifyView);` solves the issue.

***

Shouldn't we check if both views are visible on calling `addHotSpot(_pNonEditView);`?

donho:
Fix in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9059

Yaron10:
Thank you @Uhf7 and @donho.

PR #9059, opened 2020-10-25
Test tool to verify URL parser

Uhf7:
In preparation of solving #9031 (which will hopefully not be linked automatically to this PR), I would like to introduce a test tool, which can detect, which effects further modifications of the URL parser could have.


donho:
Very interesting... however it adds a dependency (Lua script plugin) to Appveyoer.
@chcg is there a way to have lua interpreter directly  in Appveyor?

donho:
@Uhf7 
The displayed result of verifyUrlDetection_1b.txt is kind of confusing. But I guess it's rather for you to make progression of URL detection.

My concern is, if this PR can be integrate to Appveyor for the building test?

Uhf7:
> My concern is, if this PR can be integrate to Appveyor for the building test? I don't see the possibility due to lack of output files.

I integrated it in the Appveyor building tests, to check, whether it is possible. My concern was, that the test itself could not work on the Appveyor VM, because it basically checks the real output in the edit window. But this seems OK.

Now, the Lua script can be run in 2 ways:
- From within the editor (Plugins --> LuaScript --> Execute Current File)
- By running the .\verifyUrlDetection.ps1 power shell script. Before starting this, Notepad++ should be closed.

PowerEditor/src/NppNotification.cpp
```diff
@@ -852,7 +852,7 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			// replacement for obsolete custom SCN_SCROLLED
 			if (notification->updated & SC_UPDATE_V_SCROLL)
 			{
-				addHotSpot();
+				addHotSpot(notifyView);
```
donho, line 855: I guess it's for fixing #9056.
Next time please do it separately, since this fix has nothing to do with the URL test issue.

Uhf7: Sorry for that, it wasn't meant to be here. I just tested it quickly, but I forgot to remove it after the test.

donho: No problem. Just a friendly reminder.

PR #9073, opened 2020-10-29
Add some missing items to Visual Studio project

### LINKED ISSUE #9063: Update Visual Studio project file, mainly for missing items
I noticed that the Visual Studio project [FILE](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0689a9445343a6c99f11959d8b4a91d598bc8495/PowerEditor/visual.net/notepadPlus.vcxproj#L1) is a bit out-of-date; it doesn't include certain files as part of the project, thus potentially causing identifiers to be missed when a "Find in Files" search (a Visual Studio one, NOT a Notepad++ one) is conducted with scope "Entire Solution".

What made me notice this is that I searched on an identifier defined in `FindReplaceDlg_rc.h` and Visual Studio did not tell me where it was defined.

I see that some other `*_rc.h` files are not part of the "project".  Is there any reason for this?
Otherwise, I plan to go through the project file and update it for missing items and other problems...

sasumner:
Fix #9063

Notes:

* Additionally, removes unneeded `../` prefix on include of `Utf8.h` in `Common.cpp`.  (Submitting with hope it doesn't fail the mingw build!)

* `localizationString.h` had a case difference in the project file, so it was removed and readded to fix this.

@donho:

These files are noticed to not be used and could be removed from the project and the github repository:

```
...\PowerEditor/src/MISC/PluginsManager/deprecatedSymbols.h
...\PowerEditor/src/MISC/crc16/Crc16.h
...\PowerEditor/src/StaticControl.h
...\PowerEditor/src/WinControls/TreeView/TreeView.cpp
...\PowerEditor/src/WinControls/TreeView/TreeView.h
...\PowerEditor/src/WinControls/WindowInterface.h
```

After potential removal of files, these "folders" will be empty and could be removed as well:

```
...\PowerEditor/src/MISC/crc16/
...\PowerEditor/src/WinControls/TreeView/
```

@donho: Do you want a PR for such removal?

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -36,7 +36,7 @@
 #include "StaticDialog.h"
 
 #include "Common.h"
-#include "../Utf8.h"
+#include "Utf8.h"
 #include <Parameters.h>
 
 void printInt(int int2print)
--- a/PowerEditor/visual.net/notepadPlus.vcxproj
+++ b/PowerEditor/visual.net/notepadPlus.vcxproj
@@ -539,17 +539,31 @@ copy ..\src\contextMenu.xml ..\bin64\contextMenu.xml
     <None Include="..\src\cursors\drag_plus.cur" />
   </ItemGroup>
   <ItemGroup>
+    <ClInclude Include="..\src\clipboardFormats.h" />
+    <ClInclude Include="..\src\dpiManager.h" />
+    <ClInclude Include="..\src\keys.h" />
+    <ClInclude Include="..\src\localizationString.h" />
+    <ClInclude Include="..\src\MISC\Common\Sorters.h" />
     <ClInclude Include="..\src\MISC\Common\verifySignedfile.h" />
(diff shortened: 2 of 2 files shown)


donho:
To remove:
```
...\PowerEditor/src/MISC/PluginsManager/deprecatedSymbols.h
...\PowerEditor/src/StaticControl.h
...\PowerEditor/src/WinControls/TreeView/TreeView.cpp
...\PowerEditor/src/WinControls/TreeView/TreeView.h
...\PowerEditor/src/WinControls/WindowInterface.h
```

Please keep:
```
...\PowerEditor/src/MISC/crc16/Crc16.h
```
I will do a CRC feature with it.

Please move `TreeView.h` & `TreeView.cpp` from `...\PowerEditor\src\WinControls\ProjectPanel\`
to `...\PowerEditor/src/WinControls/TreeView/` since this class is used in common by several classes.

> Do you want a PR for such removal?

Yes please do in another PR, no need to create an issue for that.

sasumner:
>I will do a CRC feature with it.

Hmm, something for `Tools` menu, perhaps...

PR #9075, opened 2020-10-29
Remove some unused files from project

sasumner:
Reference https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9073#issuecomment-718792506 for the reason for this PR.

Hopefully mingw build is OK with this PR. I hate to guess, but what is one to do?

Also from the above comment was:

>Please move TreeView.h & TreeView.cpp from `...\PowerEditor\src\WinControls\ProjectPanel\`
to `...\PowerEditor/src/WinControls/TreeView/` since this class is used in common by several classes.

Since this PR removes an unused version of the TreeView files, let's move the same-named files in a PR after this one is done.  The reasoning for this is that trying to do it all at once may confuse github?


donho:
Good news: `All checks have passed` :)

PR #9076, opened 2020-10-29
Move TreeView cpp and h files to different folder in project

sasumner:
See originating request https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9073#issuecomment-718792506

Not sure why github shows every line of the cpp/h files as a "diff", it clearly knows the files were simply moved/renamed:

```
On branch remove_some_unused_files_part2_XXXX
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        renamed:    PowerEditor/src/WinControls/ProjectPanel/TreeView.cpp -> PowerEditor/src/WinControls/TreeView/TreeView.cpp
        renamed:    PowerEditor/src/WinControls/ProjectPanel/TreeView.h -> PowerEditor/src/WinControls/TreeView/TreeView.h

```

PR #9083, opened 2020-10-31
Move treeview files in the project structure

sasumner:
Changes needed to move the TreeView.cpp and TreeView.h files from the ProjectPanel folder to their own folder, as these files are shared among several Notepad++ features.

Adjustments to the Visual Studio project file to keep the project building.

gcc makefile was adjusted with earlier commit to keep mingw build from failing.

See discussion in the aborted #9076 for more info.


donho:
> gcc makefile was adjusted with earlier commit to keep mingw build from failing.

I modified makefile in your old branch so in this new branch you have to import the same modification to make it work.

sasumner:
>I modified makefile in your old branch

Ah, okay, I didn't notice that detail. :-)

Issue #9055, opened 2020-10-24
Provide a mechanism for plugins to get a Scintilla handle or Document pointer from BufferID.


Kered13:
The use case here is that you want to retrieve or manipulate text in a file that is not currently in view. For example, when handling a NPPN_FILEOPENED notification the new file is not the "current Scintilla" (unless it is the first file opened when starting NPP).

I can imagine a couple ways this could be accomplished:

* A new message that will create and return a new Scintilla handle for a given BufferID. This handle can be destroyed with NPPM_DESTROYSCINTILLAHANDLE when no longer needed.
* A new message that will return the Document pointer associated with a given BufferID. The plugin can use SCI_SETDOCPOINTER to attach this Document to a Scintilla handle that it has created with NPPM_CREATESCINTILLAHANDLE. This option requires more work from the plugin, but gives the plugin more flexibility. For example it could create a single long-lived Scintilla handle and associated it with multiple Documents in turn. A reverse function (Document pointer to BufferID) might have some utility as well.

I don't mind doing the work to implement this, but I'd like to get your feedback on how you think this should be implemented first.

Kered13:
I would like to start working on this, can it be accepted and assigned to me?

My plan is to go with the second approach I described above, creating a new message:

Document NPPM_GETDOCUMENTFROMBUFFER(BufferID)

donho:
`Document` object should not be used outside of Notepad++.
What do you try to do concretely?

Kered13:
As I said before, I would like to be able to read and modify documents that are in the background. In particular, in response to NPPN_FILEOPENED and SCI_MODIFIED, and possibly other, notifications. As far as I am aware, this is not currently possible for a plugin. I have previously made [a thread](https://community.notepad-plus-plus.org/topic/20207/how-to-get-text-and-or-scintilla-handle-for-background-file) on the plugin development forum where you can see a bit more discussion, but no one came forward with a solution.

What is the security risk of exposing Document? Internally it is a pointer, but as far as the APIs exposed by Scintilla are concerned it is essentially an opaque identifier, so I don't see it how it is any worse than exposing the Scintilla handles.

[...]

PR #9085, opened 2020-10-31
Add NPPM_GETDOCUMENTFROM buffer message.

Kered13:
This message returns a Scintilla document pointer for the given buffer ID.

This implements #9055.


donho:
For the sake of security and stability, `Document` object should keep internal usage only.

Issue #5547, opened 2019-04-18
Find All in... feature request (formerly #1499)


DanielTheGreat:
MetaChuh closed this issue without any good explanation why, so I'm re-opening it, as suggested for unresolved issues.  S/he suggested I try Notepad++ version 7.6.6, which I did, but with no change to the unwanted behaviour.  If it was closed because the requested improvement was not thought worthy of doing, then please say so.

I don't use it often, but I like Notepad++ for the occasional coding (hobby) project! Just one little thing slightly annoys me, so I wondered whether others would agree to a small change to the way the 'Find All in...' feature operates...

I sometimes use 'Find All in All Opened Documents' (or the other one) to find all occurrences of a particular string that I want to check and/or edit. But if I search for a string (perhaps a variable or register name) that occurs twice (or more) in one line, the search results pane shows that line twice (or more). This simply clutters the search results, and seems to serve no useful purpose.

Here's an example:
define vca25 vca2_addr.bit.b5 // Voltage detection 0 enable bit

If I'm searching for 'vca2' to find everywhere the register is accessed, I don't need to see that line appear twice in the results; I just want to know every file and every line where that register appears.

Would it be practical (and acceptable to most users) to change the 'Find All' code such that when a string is identified in a particular file/line, that line is then skipped from further searching so it appears only once in the results pane?

I realise this will complicate the code a little, since I presume there is much common code between 'Find All in...' and 'Replace All in...' functions, whereas for 'Replace All in...' you do want every instance found, not just the first per line!

Thanks, Daniel
(Gerroa, Australia)

krilbe:
I think this change should be optional, if implemented. I often use this type of search in Visual Studio, which also shows every occurrence on every line. This is great when the purpose is to find all references to something in order to assess if the reference needs to be modified due to some change to it or its semantics.

The V File Viewer also shows every occurrence on every line, I think.

DanielTheGreat:
krilbe, I think you might have misunderstood the problem I'm commenting on...

Of course we want the Search to find every occurrence of a search term, but we don't need to have the same line DISPLAYED TWICE just because the search term appears twice in that one line.  I can't see that that achieves anything useful.

Daniel

sasumner:
Check out my workaround presented in #1499

sasumner:
[...]

Issue #4525, opened 2018-05-25
Suggestion: don't duplicate lines with multiple hits in "find results"


Layarion:
This is how I would like multiple hits to be handled on Notepad++.
https://i.imgur.com/43NPong.png
The above image in EditPad Pro 7.6.4

It's not though. Instead, lines with multiple hits get duplicated. I feel that this makes it harder to scroll through and read. like so: https://i.imgur.com/AfjlI35.png

If you feel that this would be a problem with the general public, make it optional.

sasumner:
Duplicate of #2987.  Please spend some time searching for existing issues before creating new ones.

Layarion:
I don't remember if I did. I probably couldn't think of the keywords that person was using if I did.

That post was made in 2017, and I feel like my images deliver a clearer image (no pun intended) of what's wanted.

though it's saddening, to see proof that it likely won't happen.

weirdalsuperfan:
Man I wish this was an option

sasumner:
This feature-request is REJECTED by the author of Notepad++ ( https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9088#issuecomment-720016310 )
It will not be reconsidered.
Please do not open any new issues on this topic as they will be immediately closed.

The suggestion is to use this workaround:
* Set your _Search mode_ to _Regular expression_.
* Add the following to the end of your _Find what_ expression:  `(?-s).*`

Example 1: To show only the first hit on a line for `foo`, make your _Find what_ expression `foo(?-s).*`

If the literal text you are intending to search for contains regular expression metacharacters, add `\Q` to its start and `\E(?-s).*` at its end.

Example 2: To show only the first hit on a line for `*foo*`, make your _Find what_ expression `\Q*foo*\E(?-s).*`

Layarion:
"To show only the first hit on a line" I do not think you understand what we asked for, because that's not anything close to what we wanted. Please read it again, because it is obvious you did not understand what is being asked for.

sasumner:
@Layarion 

It is 100% obvious you are...

My response and closure of this some hours ago was because this issue is one of a number of similar ones, and that closure comment was boilerplate used in the closure of all of them.  Apologies if it doesn't exactly "fit" with your issue.

Issue #3224, opened 2017-04-28
Find in active/files returns same line multiple times


ecki:
### Description of the Issue
When I search in multiple files or in all active files I get a list of search results in the lower pane which is incredibly useful, thanks for this. However when using a search word which occurs multiple times in a line, it will show the line also multiple times (each time with another occurrence of the word higlited. In my cases for logfiles it is up to 10 times, which makes the result sometimes harder to read than the actual file.

I can see that this is sometimes useful, but most of the time it stops you from quickly scanning the list. So a feature in the search mask would be "report fond lines instead of found occurences" or something line that. From my point of view the new behavior would be the better default, but I am not sure what the motives for the current behavior is.

### Steps to Reproduce the Issue
Create 2 files in an empty dir, both containing
a
b b
a

Search for "b" in all files, you will receive 4 lines instead of 2:
file 1:
line 2: **b** b
line 2: b **b**
file 2
line 2: **b** b
line 2: b **b**

### Expected Behavior
Get a 
file 1
line 2: **b** **b**
file 2
line 2: **b** **b**

I see this in 7.3.2 32bit but it is like this for years.

tergabri:
I don't think it is a bug because it make it easier to look for it.
For example if you look for b in a line with 1000 non 'b' characters, with one b in position 200 and 600 you will prefer to have it repeated. This will help you locate your 'b' characters.

I can try to implement something to fix it if we all agree it is a bug.

Hope that helps

ecki:
This is a feature request (but I somewhat fail to see how this makes it any easier to see the multiple matches)

bahusoid:
It helps you to find exact position in the string of each match by clicking on each result.
What is expected behavior on click with you suggested fix - select only first match?

ecki:
Well, I only care about jumping to that line so I can see the context. selecting the first match (which would highlite the others) is a good choice. Selecting the whole line or nothing is also acceptable as long as the cursor is positioned in the line.

nikunjbhatt:
I was also going to request for this feature.

If possible, what can/should be done is: show single line in search window and when clicked on a found string, the cursor should jump to clicked string in editor.

sasumner:
@nikunjbhatt said:

>show single line in search window and when clicked on a found string, the cursor should jump to clicked string in editor.

First, that is exactly what already happens.

[...]

Issue #2987, opened 2017-03-03
Feature Request: Find All to return only unique lines even if some lines have multiple matches of the text or expression


celticvalley:
### Current Behavior
When using Normal or RegEx expressions to "Find All" in current or all documents, one will end up with duplicated lines in the find results if the search text or expression appears 2 or more times in a line.   I fully understand that this could be considered normal and expected as a matter of how the search mechanism works.  However, for us regular human beings, we want to see only one line returned for each expression matched, even for lines that are matched multiple times.

### Current Work-around
There are a few RegEx ways to return only one line.  I provide one example here.  The problem with these work-arounds is they nullify a nice feature of Notepad++ where Notepad++ highlights the text matched in the find results.  The highlighting feature is nice because it draws my attention to the text I am searching for.  But the RegEx work-arounds causes Notepad++ to either highlight much more text than what I am interested in (one method), or to highlight only the first matching text I search for in the line, not all of the matching text (another method which I show here).

Example work-around:
^.*?\Ktargetword  (this returns only unique lines, but also only highlights the first occurrence of "targetword".

### Desired Behavior for new feature
Ideally Notepad++ would have a check box for whether to return only unique lines.  If unchecked, behavior is as it is now.  But if checked, when Find All is used only unique lines are returned and every instance of the found text is highlighted, even if 2 or more instances of the text is found in a line.  This desired feature would be available for both Normal and RegEx searches.  Stated another way, Notepad++ would do the work of consolidating what would have been multiple lines into a unique line in the background on my behalf.  This would be so much more intuitive than what we have today.

Thanks!

nikunjbhatt:
Duplicate of #3224

sasumner:
This feature-request is REJECTED by the author of Notepad++ ( https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9088#issuecomment-720016310 )
It will not be reconsidered.
Please do not open any new issues on this topic as they will be immediately closed.

The suggestion is to use this workaround:
* Set your _Search mode_ to _Regular expression_.
* Add the following to the end of your _Find what_ expression:  `(?-s).*`

Example 1: To show only the first hit on a line for `foo`, make your _Find what_ expression `foo(?-s).*`

If the literal text you are intending to search for contains regular expression metacharacters, add `\Q` to its start and `\E(?-s).*` at its end.

[...]

PR #9088, opened 2020-10-31
Add Find All option to show max of 1 hit per source line

### LINKED ISSUE #1499: Single line of output to Find-result window for lines with 1+ hits

sasumner:
Fix #1499, #5547, #2987, #2481, #4525, #3224 and possibly others

Also, satisfies requests made in these Community threads:

* https://community.notepad-plus-plus.org/topic/20227/let-result-lines-for-text-search-pattern-appear-only-once-even-if-there-are-more-occurrences

* https://community.notepad-plus-plus.org/topic/15631/find-results-multiple-matches-per-line-would-like-non-duplicate-line-result-highlighting-every-match

* https://community.notepad-plus-plus.org/topic/20146/find-all-current-duplicate-rows

* https://community.notepad-plus-plus.org/topic/15727/find-all-in-current-doc-reports-dup-lines-when-dup-occurence

User interface changes:

image

image

Notice in the output that there are TWO hits per line, but only ONE (the first) is shown in the results window:

image

Note the verbage `Condensed Hits Mode: Display only 1st hit per line` is added to the result for this type of search.

Note that TOTAL hits (6 in this case) is still displayed; it is only the lines output that is altered.

This PR makes this feature macroable.

This PR makes this feature work in _Find in these found results..._ output windows as well:

image

The user may move between the condensed hits using F4 and Shift+F4.  This will jump between the hits displayed.  If the user wishes to move between _all_ hits, not just the first on the lines, this procedure may be followed:

* locate the line of interest in the condensed results using F4 or Shift+F4
* press Enter to move input focus to editor window (the text of the "hit" is selected)
* press Ctrl+f (invokes Find window and puts the text of the "hit" string in the _Find what_ box)
* press Enter (finds the _second_ hit on the same line, if there is one; repeat for _third_ hit, etc.)


donho:
It's indeed a point that we can improve but I really don't like to add more and more options, because that makes application more and more difficult to use. 

Only the following result will be accept : 
image

In this way, no more unnecessary option in the middle of find dialog which makes my eyes bleeding.
But I don't want this feature badly - not so badly to make me accept a big modification which may bring the regression and the complexity to maintain.

sasumner:
>But I don't want this feature badly

Okay, I will close all existing issues with a rejection notice and a warning not to open new issues on same topic.

Any new issues that are opened will quickly be closed with a reference to here saying author declines the feature.

In the past you have said "let's see what users say" -- with all of the issues and links I've cited above, users are clearly wanting such a feature; I'm fine with denying them. :-)

I will add the workaround to everyplace I post the rejection:
"Search in regular expression mode and add `(?-s).*` to the end of your search text"

sasumner:
@donho

I'm OK with rejection. :-)
But it does open the door to more discussion, below. :-)

>It's ... a point ... we can improve but I ... don't like ... more options

It would be _impossible_ to add it without an option.
Users would complain that they liked the old behavior, thus it needs a way to choose.

>but I really don't like to add more and more options

We do this all the time, in the _Preferences_.

Personally I don't like everything having to retain backward compatibility:
I think if the designers make a change that is "for the better", it doesn't have to keep the old way -- users should adapt.

>Only the following result will be accept

With no way to turn it on/off?
Visually highlighting ALL of the matches on a single line would involve a lexer change (more complication, more possibility of regression).
It is of limited value without some sort of way for the user to jump the editing caret back to the doc at the point of the 2nd, 3rd, ... match on a line.

With this current PR's approach, the user knows that they will jump to the only highlighted match on the line (the first match).

>unnecessary option in the middle of find dialog which makes my eyes bleeding

"eyes bleeding" -Nice!  :-)

Still not sure why it is "unnecessary" but comments above cover that.

The checkbox location was just a proposal; had the feature been liked it could have been negotiated where to put it, what its exact text should be, etc.

The _Find_ dialog has no more vertical space (e.g. under _Wrap around_) without making the window itself larger vertically.  I was reluctant to do that because of existing user complaints about the large size of the Find window.

>(won't) accept a big modification which may bring the regression and the complexity to maintain

I did put a good deal of time in to make sure the feature was solid.
I don't feel it was complex to maintain.
I think you just don't like it for the earlier reason and these are just additional statements.
It's OK to not like it.

But it brings some related questions:

* **What is a "big modification"?**
If I want to do a change, and it is more than 25 (50? 100?) touched lines, should I just not do it?
Not risk a "too big" rejection?
Not risk wasted time.
There were some lines changed in this PR that didn't have to be that simply made the code better.
Look at the addition of this line: `FindOption const * pFinderFindOpts = ...`.
This made sense to do for the new code accessing a "find option", but it also allowed some existing code to be "simplfied".
It did not make sense to add `pFinderFindOpts` for the new code without simplifying the existing code.
BUT...this made the overall change "bigger".
Was simplification a bad thing because it touched more lines?
Should we be so afraid of regression that code that could be made "cleaner" is left "sloppy"?
Doesn't that just make it harder to read and maintain?

* **How are devs to know when you "don't want this feature badly"?**
You admit you don't follow "issues" well, just PRs.
If you followed and contributed to issues reliably, devs would know it is OK to work on something without it being rejected as "not wanted" after work effort is expended and PR prepared.
Of course it could be rejected later if not implemented well/correctly.
But that opens the door to an easy "too big" rejection if the mind is changed.
What regular devs need is a reliable way to contact you for such discussion.
You could say that you will try to follow issues better.
But sometimes a dev might want discussion that they don't want to take place in "issue" comments (it invites ANYONE to add their opinions which maybe aren't wanted at that point).
You have a (small) set of regular devs that are expending their effort to make the product better -- why risk lowering their enthusiasm for it?
**What are your suggestions for improving this situation?**

donho:
> In the past you have said "let's see what users say" -- with all of the issues and links I've cited above, users are clearly wanting such a feature; I'm fine with denying them. :-)

It's true that if there's a lot of requests for a feature, then I would implement it. But here for me the requests have reached the limit of implementation. In this PR, there's 2 disadvantages: 1. we add one more option 2. this option works partially, and also add more info for explanation. For me the real solution is make it work as expected, without redundant option - I'm sure you agree that screenshot I have provided is the best solution.

donho:
> It would be impossible to add it without an option.

True. Just as "it's impossible to not eat", it doesn't mean that we should put anything edible in our mouth :) 

> Users would complain that they liked the old behavior, thus it needs a way to choose.

Users also are confused by huge options in front of them, the most of users are not power users, and they just ignore all options without complaining. It could be worse  - they just don't use Notepad++ for another more easier editor.

> We do this all the time, in the Preferences.

Then we must reduce this part in the future.

> Personally I don't like everything having to retain backward compatibility:
I think if the designers make a change that is "for the better", it doesn't have to keep the old way.

Agree, but I do believe a lot of backward compatibilities in Notepad++ are necessary. 

>  users should adapt

You've got the point - but which part of users who should adapt? Those who are using this way for long time, or these who want other ways in the future?

>> Only the following result will be accept

> With no way to turn it on/off?

It's the correct way my suggestion, so why turn it off?
 
> Visually highlighting ALL of the matches on a single line would involve a lexer change (more complication, more possibility of regression).

True. That was my point "But I don't want this feature badly - not so badly to make me accept a big modification which may bring the regression and the complexity to maintain"

> It is of limited value without some sort of way for the user to jump the editing caret back to the doc at the point of the 2nd, 3rd, ... match on a line.

It doesn't matter, user jumps to the 1st occurrence, then he/she will find the 2nd and 3rd... in the same line.

> With this current PR's approach, the user knows that they will jump to the only highlighted match on the line (the first match).
Still not sure why it is "unnecessary" but comments above cover that.

"Unnecessary" is related to our screenshot suggestion, what I meant is, this option is no need to be there with my suggested solution. Sorry for not being clear.

> The checkbox location was just a proposal; had the feature been liked it could have been negotiated where to put it, what its exact text should be, etc.
The Find dialog has no more vertical space (e.g. under Wrap around) without making the window itself larger vertically. I was reluctant to do that because of existing user complaints about the large size of the Find window.


Different from Preferences dialog, Find/Replace dialog is launched all the times for searching, so I prefer to not add any control on it as possible.

>I did put a good deal of time in to make sure the feature was solid.
I don't feel it was complex to maintain.
I think you just don't like it for the earlier reason and these are just additional statements.
It's OK to not like it.

I'm still not convinced it's a good idea to add it as part of searching functionality.
OTOH, I wouldn't say no if we add a command `"Show only the 1st found line"` or `"Remove the redundant lines"` to the context menu of Find result - you can use "Hide lines" feature to archive it.  

> What is a "big modification"?

"big modification" it's vague, indeed. by "big modification" is not necessary the large quantity of lines of code. It could be the modification of core function, like save or searching. 

> How are devs to know when you "don't want this feature badly"?

Unfortunately, there's no reply to this question. It's kinda my feeling, but I know it's not a good answer.
You can write to me don--dot--h--at--free--dot--fr before you implementing a big feature, that will avoid wasting of your time.

Thank you for your feedback and sorry for your wasting time.

nikunjbhatt:
@donho @sasumner 
This seems better:
> `"Show only the 1st found line"` or `"Remove the redundant lines"` to the context menu of Find result

sasumner:
@donho 

Thank you for spending the time to reply.

>For me the real solution is make it work as expected, without redundant option and information - I'm sure you agree that screenshot I have provided is the best solution.

Visually it is a good solution.
But users, who do not think their requests through all the time before making them, will expect a quick way to jump to any of the highlighted matches on those lines with multiple hits.
This PR made it very clear, by marking only the first match on a line, that this is the only one that is "jumpable".
There could be other ways to make multiple matches per output line "jumpable", but more code, more changes, more risk...

>don--dot--h--at--free--dot--fr before you implementing a big feature, that will avoid wasting of your time.

Am I certain to receive a reply?
I believe I have done this in the past once and received no reply. :-(
To be fair, I think I did this twice and one time received a reply.

I will consider the rest of your comments and consider the possibility of reworking the subject of this PR somewhat so that it acceptable.  OR... I may just leave things as they are with the workaround I proposed.

sasumner:
@donho So something like this would be OK? :

image

The "Show only..." would control changing the view of the right-clicked search result.

The "Always show..." would control how future searches get output, for users that always prefer one behavior.  This setting would be remembered in config.xml.

donho:
@sasumner 
> To be fair, I think I did this twice and one time received a reply.

Sorry, I could misdelete some email. But if I read the mail, I would answer for sure.

>  The "Show only..." would control changing the view of the right-clicked search result.
The "Always show..." would control how future searches get output, for users that always prefer one behavior. This setting would be remembered in config.xml.

How about a option "Show only..." in the "Searching" section of Preference dialog ?

nikunjbhatt:
@sasumner @donho 
"Show only ..." should be at both places, in find result window and in preferences window and should be synced.

However, if there is any performance issue of adding/showing/hiding multiple lines of same search term in the search results window then it is better to keep the option only in the preferences window and always show only 1 line if the option is checked.

sasumner:
@nikunjbhatt 

>"Show only ..." should be at both places, in find result window and in preferences window and should be synced.

Well, that (both places) we WON'T do. Sorry.

donho:
@sasumner Agree.

@nikunjbhatt If "Show only ..." **command** is on Search result context menu, then user should run it (or not) after each search. Or "Show only ..." **option** is on Preference, then if it's checked, after each search this operation will be applied. We won't have both. I think user will prefer to have it as option.

jue71:
Hi all, first: thanks a lot for all the work you did. I love it and I am very pleased always I use it.

Although the thread/PR seems to be closed the discussion was ongoing and so my 5 cents may help... don't want to waste time, but maybe some experience and ideas help's about the topic:

- sometimes I'm fooled a bit by showing a line multiple times and recognize later that it was the same line with multiple findings.
- on the other hand I fear that I'm fooled if only the first finding is shown

=> ideas:
- collapsing the same line to one line with all findings marked (as shown in donho's screenshot on 1 Nov 2020) I would appreciate very much
- maybe the count of findings (at the begin of the result lines) can give a hint to multiple findings (see example below)
- double click jumps to first finding in line. To hop to the next findings it is easy to press "F3" as the search pattern is active in search up do now (maybe a hint for people who handle very long lines).
- option for old behaviour is not important for me. But it can also be in preferences as I don't think that this will be changed very often.

Example:
Line 1 (2): **ab**cd**ab**
Line 2 (2): **ab**cd**ab**
Line 3 (2): **ab**cd**ab**
Line 4 (1): **ab**c

sasumner:
@jue71 

Thank you for your comments.
This feature isn't "dead"... it is just "hibernating".  :-)

>collapsing the same line to one line with all findings marked (as shown in donho's screenshot on 1 Nov 2020) I would appreciate very much

Not sure to what you're referring as far as the screenshot, but @donho did have some good suggestion regarding this and getting the feature finally accepted.  I think another of the ideas was hiding/unhiding lines.

>Line 1 (2): abcdab
Line 2 (2): abcdab
Line 3 (2): abcdab
Line 4 (1): abc

While this is a good idea, I don't think it is reasonable because it would cause a disruption to the lexer code (the thing that colorizes the hits in the output).

>double click jumps to first finding in line

I agree with that.

>To hop to the next findings it is easy to press "F3" as the search pattern is active in search

I think what you mean here is to use the keycombo for _Select-and-find-next_ for this case.
Your (F3) suggestion works if you are working with the most recent search, but _Select-and-find-next_ seems like it would always work (after you've forced a move from the _Search results_ window to the editor -- with double-click for example).

>option for old behaviour is not important for me. But it can also be in preferences as I don't think that this will be changed very often.

I'd have to refresh myself on the discussion about this, but it may be best to have it in the _Search results_ right-click menu (like the setting for wrapping long lines).

Again, thanks for your input.

jue71:
> Thank you for your comments.
> This feature isn't "dead"...

I'm quite happy that you appreciate my comments and that the feature isn't dead

>  it is just "hibernating". :-)

 :-) What a luck that spring is coming soon ;-)

>> collapsing the same line to one line with all findings marked [...]
>
> Not sure to what you're referring as far as the screenshot,

It was just pointing out that it is good that all findings of a line are marked not just the first one as shown in screenshots before (**bold** shows the marking of findings):
Line 1 (2): **ab**cdab
It then would be hard to find the other findings in a line on a first glance.

>> Line 1 (2): **ab**cd**ab**
>> Line 2 (2): **ab**cd**ab**
>> Line 3 (2): **ab**cd**ab**
>> Line 4 (1): **ab**c
> 
> While this is a good idea, I don't think it is reasonable because it would cause a disruption to the lexer code (the thing that colorizes the hits in the output).

Sorry, I'm not familiar with the code... so it was just an idea which may help the users (including me ;-) ) not to overlook findings especially if they are in a long line in the part which is currently not shown. I have no glue how much work it is to count the findings of a line and write this number right after the line number.

>> To hop to the next findings it is easy to press "F3" as the search pattern is active in search
> 
> I think what you mean here is to use the keycombo for _Select-and-find-next_ for this case.

You mean Ctrl-F3? Again learned something :-) but no, this was not my intention.

> Your (F3) suggestion works if you are working with the most recent search, but _Select-and-find-next_ seems like it would always work (after you've forced a move from the _Search results_ window to the editor -- with double-click for example).

ah, you are right... I forgot the older searches... so, yes, you are right... Ctrl-F3 would to the trick after double clicking on the according result line

>> option for old behaviour is not important for me. But it can also be in preferences [...]
> 
> [...] it may be best to have it in the _Search results_ right-click menu (like the setting for wrapping long lines).

Also a good idea... I think it would be helpful then, if the state is memorized for the next search: this means that if someone selected collapsing multiple findings of a line to just one line then it should be always the same as long the user doesn't select the opposite.

PR #9089, opened 2020-11-01
Add color samples on menu items for styling features

### LINKED ISSUE #8983: [Feature Request] style tokens: "in use" indicator; color preview; more style tokens
1) It would be nice, if the entries in the _style token_ menu had an indicator, if they are currently used, so you cannot use the same token for different markings by accident (assuming f.i., that you want to mark certain words, but in different colors).

2) While you are at it, you could add a box showing the color of the style to the menu entries

3) Is where a specific reason, why there are only 5 different styles available? Is this configurable? For certain tasks/workflows it would be useful to have more styles.

sasumner:
Fix #8983 (2nd request only; other requests are rejected)

New UI samples:

image

image

image

image

image

image

image

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -411,6 +411,8 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	// Menu Section //
 	// ------------ //
 
+	setupColorSampleBitmapsOnMainMenuItems();
+
 	// Macro Menu
 	std::vector<MacroShortcut> & macros  = nppParam.getMacroList();
 	HMENU hMacroMenu = ::GetSubMenu(_mainMenuHandle, MENUINDEX_MACRO);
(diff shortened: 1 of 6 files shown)


donho:
Awesome!
I will check it as soon as v7.9.1 is released.

Yaron10:
@sasumner,

This is beautiful. :+1: 
Thank you.

Using "Windows Classic Theme" on Windows 7, the images are cut off (to 13x13).
תמונה

Interestingly, *icons* are displayed correctly.
תמונה

sasumner:
@Yaron10 

>Using "Windows Classic Theme" on Windows 7, the images are cut off (to 13x13).

I kind of like the look you show; rather 3D :-)

> icons are displayed correctly.

Where is that from? (I can't tell without it being in English)

Yaron10:
@sasumner,

> I kind of like the look you show; rather 3D

That's why I like the "Classic Theme".

> Where is that from?

`Mark All` + `Unmark All`.
With the `MenuIcons` plugin (some parts of NPP code need to be modified for that plugin to work properly).

> I don't know how long we hold on to Windows 7, especially for purely visual things.
In other words, I'm not inclined to put more effort into it as I have no way to test Win7 things anymore.

Fair enough. Ill try it later with `const int bitmapXYsize = 13;`.

You've edited *my* post.

sasumner:
>You've edited my post.

Oops.  Sorry.  Not sure how that happened.  Will fix.

sasumner:
>With the MenuIcons plugin (some parts of NPP code need to be modified for that plugin to work properly).

Hmmm, not sure I like the idea of that plugin.
Because if N++ devs want to do nice things with the menus on their own, would there be conflicts?
I certainly would not want to be restricted from doing this, just to stay compatible with a plugin's behavior, in some way.
This treads on a developer's turf, IMO.

Yaron10:
> Because if N++ devs want to do nice things with the menus on their own, would there be conflicts?

As of now there are no conflicts.
And, obviously, NPP devs should not be restricted whatsoever by this plugin. :)

Yaron10:
@sasumner,

It's displayed properly with `const int bitmapXYsize = 13;`.

תמונה

That's a nice and helpful addition. :+1: 
Thanks again.

***

> I started out using smaller color sample bitmaps.
But I noticed that it made the menu items vertically closer together (so that it was noticeably different than the main menu items).

This is also the result on my machine with "Windows Aero": the vertical gap between menu-items is reduced from 14px to 11px.

sasumner:
>This is also the result on my machine with "Windows Aero": the vertical gap between menu-items is reduced from 14px to 11px.

Yea, I'm not sure why it would do that; it doesn't seem to make sense but maybe there is some hidden reason for it.

Yaron10:
I've just noticed another issue with the "Classic Theme": on hover - the original/correct color is changed to a completely different one. It seems the colors are inverted.

תמונה

Well, someone might want to look into it. :)

Yaron10:
> The selected and clear bitmaps should be monochrome. The system uses the Boolean AND operator to combine bitmaps with the menu so that the white part becomes transparent and the black part becomes the menu-item color. If you use color bitmaps, the results may be undesirable.

https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setmenuitembitmaps#remarks

sasumner:
>The selected and clear bitmaps should be monochrome

I saw that, but I didn't know what it means.
Still don't. :-)
If they are monochrome, how can one express color.
Maybe one can't reliably.
Maybe it kills the whole idea.

Yaron10:
Interestingly, it seems to be working properly with "Windows Aero" and your Win 10.
Could you test it with other themes on Win 10?

> Maybe it kills the whole idea.

That would be unfortunate. I really liked it.
@donho wrote he would test it later. Fingers crossed. :)

Anyway, thanks again.

sasumner:
@Yaron10 

Maybe I just keep some of my better ideas for my own implementation.
Then I don't have to be concerned with such troubles, because I'm not going to be changing my theme.
:-)

Yaron10:
> Maybe I just keep some of my better ideas for my own implementation.

Please don't. 
It was a clever idea and a nice execution. And, who knows, some minor change can still fix it.

> because I'm not going to be changing my theme.

I meant just for testing. As you wish. :)

sasumner:
>I meant just for testing.

Yes, I know what you meant.
What I meant was that if I "keep some of my better ideas for my own implementation" then probably they are less work -- if they work for me, they work, and I don't have to consider themes, Window 7, blah, blah, blah, etc.  :-)

sasumner:
@donho 

Did you read any of the comments above between @Yaron10 and myself ?
In light of those comments, I have some serious doubts about this PR !  :-)

donho:
No, I didn't. will do now.

donho:
@sasumner
So do we have the solution for 2 problems (square size and inverse color) of Win7 found by @Yaron10 ?

donho:
`const int bitmapXYsize = 13;` is too small - 16 is the way better in my win10.

donho:
Sorry @Yaron10, but I would say let's keep the PR as it is. Since Windows 7 is not even supported by Microsoft:
https://support.microsoft.com/en-us/windows/what-happens-when-windows-7-support-ends-5c740944-b787-a89e-ad23-69ec848ae763#:~:text=After%2010%20years%2C%20support%20for,including%20security%20updates%2C%20from%20Microsoft.

Yaron10:
@donho and @sasumner,

I'd like to emphasize that the problems seem to occur only with "Windows Classic Theme".
I suppose that theme is not widely used these days. :)

Typing these lines, your last comment popped-up. :+1: 

Thank you.

donho:
No. Thank you. Thank you for your contributions to this project / community.

Yaron10:
@donho,

Well, thank you for this project. I, with millions of fellow users, highly appreciate your work. :+1:

donho:
***Our*** work

sasumner:
@donho 

Well, if we are all done praising each other ...  :-)

So acceptance of this PR, and the comments above, seems to indicate that we're okay with this feature, and users that use obscure Windows theme setups where it may not appear looking great are to just be told "too bad, live with it"?  

I'm fine with that stance, I just want to be sure that that is the feeling.  :-)

donho:
> So acceptance of this PR, and the comments above, seems to indicate that we're okay with this feature, and users that use obscure Windows theme setups where it may not appear looking great are to just be told "too bad, live with it"?

Yes @sasumner , unfortunately we cannot make everyone happy. But we can at least make the most of people happy.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2157,6 +2159,68 @@ void Notepad_plus::checkSyncState()
...
+			const int bitmapXYsize = 16;
+			HBITMAP hNewBitmap = CreateCompatibleBitmap(hDC, bitmapXYsize, bitmapXYsize);
+			HDC hDCn = CreateCompatibleDC(hDC);
+			HBITMAP hOldBitmap = static_cast<HBITMAP>(SelectObject(hDCn, hNewBitmap));
+			RECT rc = { 0, 0, bitmapXYsize, bitmapXYsize };
+
+			// paint full-size black square
+			HBRUSH hBlackBrush = CreateSolidBrush(RGB(0,0,0));
+			FillRect(hDCn, &rc, hBlackBrush);
+			DeleteObject(hBlackBrush);
```
Yaron10, line 2202: @sasumner,

The default background color is black.
We'd get the nice black frame without that block. Can you test it please? 

Thank you.

PR #9099, opened 2020-11-04
Update spanish.xml

swazmo:


donho:
@swazmo 
Could you provide some description about your fix?

PR #9105, opened 2020-11-05
Update ukrainian.xml

a-polivanchuk:


donho:
@a-polivanchuk 

donho:
@a-polivanchuk 
With no reply from you, I'll merge your PR as its current state. 
If there's further modification for updating Ukrainian localization will wait for the next release (about 1 month later).

Issue #9098, opened 2020-11-03
When the "Find result" panel refers to a search on the unsaved file "NEW 1", only, any further search, with the "Find in these found results..." option, does NOT match anything !


guy038:
### Description of the Issue

A search, using the **`Find in these found results...`** option, does **not** match anything if a **`Find All in Current Document`** was performed on the **non-saved** file **`new 1`**, **exclusively** !

### Steps to Reproduce the Issue

1. Start Notepad++

2. Click **`3`** times on the **`Ctrl + N`** shortcut  => The **`3`** tabs **`new 1`** , **`new 2`** and **`new 3`** appear

3. Open the well-known file **`License.txt`**

4. Copy all its contents and **paste** it into each **non-saved** file

5. Select the **`new 3`** file, first

7. Open the Find dialog ( **`Ctrl + F`** )

3. SEARCH **`your`**

4. **Tick** the **two** options **`Match whole word only`** and **`Wrap around`**, **only**

5. Select the **`Normal`** search mode

6. Click on the **`Find All in Current Document`** button

=> The **`Find result`** appears with the message *Search "your" (14 hits in 1 file of 1 searched)*

7. **Right**-click in the **`Find result`** panel

8. Select the **`Find in these found results...`** option

9. SEARCH **`that`**

10. **Tick** the **two** options **`Search only in found lines`** and **`Match whole word only`**

11. Select the **`Normal`** search mode

12. Click on the **`Find All`** button

=> A second **`Find result`** panel appears with  the message *Search "that" (6 hits in 1 file of 1 searched) - Line Filter Mode: only display the filtered results*  ( **OK** )

13. Click on the **`X`** button to close the second **Find result** panel

14. **Right**-click in the main **`Find result`** panel

15. Select the **`Clear All`** option to get an **empty** panel

---

16. Select the **`new 2`** file and **repeat** the steps **`7.`** to **`15.`**

 => Again, you **correctly** get the message *Search "that" (6 hits in 1 file of 1 searched) - Line Filter Mode: only display the filtered results*

17. Select the **`new 1`** file and **repeat** the steps **`7.`** to **`15.`**

### Actual Behavior

=> This time, the search in **`Find in these found results...`** panel does **not** match anything and you get the message *Search "that" (0 hits in 0 files of 1 searched) - Line Filter Mode: only display the filtered results*

[...]

PR #9115, opened 2020-11-06
Fix bug where search-results won't jump focus into 'new 1' file

### LINKED ISSUE #8461: [Find Result] go to wrong tab when file is temporary

sasumner:
Fixes #8461 , #9098 

When a "buffer" is created, it is given a name "new (digit)" where "digit" is allowed to repeat if the buffer being created is not visible to the user.  This is likely so that if the user has only one tab open and it is labeled "new 1" and they press Ctrl+n that they aren't faced with the "awkwardness" of having "new 5" being opened, if the "system" has opened some invisible buffers for its own use in the meanwhile.

A real problem when the code wants to get a buffer by its name, because potentially several "new 1" (for example) buffers can simultaneously exist -- which one should be returned?  The answer should be "the visible one", which this PR satisfies.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScitillaComponent/Buffer.cpp
@@ -1463,7 +1463,12 @@ BufferID FileManager::getBufferFromName(const TCHAR* name)
 	for (size_t i = 0; i < _buffers.size(); i++)
 	{
 		if (OrdinalIgnoreCaseCompareStrings(name, _buffers.at(i)->getFullPathName()) == 0)
-			return _buffers.at(i)->getID();
+		{
+			if (_buffers.at(i)->_referees[0]->isVisible())
+			{
+				return _buffers.at(i)->getID();
+			}
+		}
 	}
 	return BUFFER_INVALID;
 }


Yaron10:
@sasumner,

One of your most important fixes. :+1: :+1: 
Thank you.

sasumner:
@Yaron10 

>most important fixes.

Really...?
:-)

Yaron10:
Absolutely. The fixed issues are major IMO.

Yaron10:
@sasumner,

The issue you brought up in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8461#issuecomment-650849558 is still valid.
But that's not a common case.

sasumner:
>But that's not a common case.

And it probably should not have been allowed -- being able to name two temporary tabs with the same name -- in the first place .
If someone opens an issue on that, and that issue is dealt with, it will take care of the current concern.

Yaron10:
> And it probably should not have been allowed -- being able to name two temporary tabs with the same name -- in the first place .

Yes, that makes sense.

PowerEditor/src/ScitillaComponent/Buffer.cpp
```diff
@@ -1463,7 +1463,12 @@ BufferID FileManager::getBufferFromName(const TCHAR* name)
 	for (size_t i = 0; i < _buffers.size(); i++)
 	{
 		if (OrdinalIgnoreCaseCompareStrings(name, _buffers.at(i)->getFullPathName()) == 0)
-			return _buffers.at(i)->getID();
+		{
+			if (_buffers.at(i)->_referees[0]->isVisible())
+			{
+				return _buffers.at(i)->getID();
+			}
+		}
```
donho, line 1471: So this bug is due to the other hidden "new #" buffer is found?
Nice catched!

sasumner:
In the opening comments to this PR, I said:

>...if the user has only one tab open and it is labeled "new 1" and they press Ctrl+n that they aren't faced with the "awkwardness" of having "new 5" being opened, if the "system" has opened some invisible buffers for its own use in the meanwhile.

Which reminds me much of #8677
:-)

PR #9124, opened 2020-11-07
Add ellipsis to Rename and Print on tab bar context menu

sasumner:
The _File_ menus' rename and print commands have ellipsis (`...`) after their text, indicating that activating the menu item will cause an input window of some type to open:

image

But the "tab bar" right-click context menu equivalents have no ellipsis:

image

This issueless PR corrects this situation.


sasumner:
@donho @vinsworldcom

Semi-related:  Do you think `Rename...` is sufficient so that user knows this command can also move a file to a different folder/drive on disk?  Or should it be `Rename or move...` ?

donho:
@sasumner Good question (from the aspect of devs/IT guys). However, I think it's no need to complicate the meaning of this command. "Rename" is pretty clear for both technical and non-technical people.

sasumner:
@donho 

Best answer:  RTFM (if it is in there!) 
:-)

PR #9126, opened 2020-11-08
Update Slovak translation

kubalav:


donho:
@kubalav 

kubalav:
Translation completed!

PR #9127, opened 2020-11-08
Prevent names of untitled tabs from duplication

### LINKED ISSUE #9119: Unsaved files should not be allowed to have the same name
### Description of the Issue

Using the _File_ > _Rename_ feature on an unnamed/unsaved document allows something more memory-jogging than `new 4` or `new 2` on a tab.

However, currently two or more untitled documents may be given the same name, without error.

This is problematic for at least one circumstance (described below), but there may be others, and really, is it helpful to have two tabs with the exact same temporary name?

1. Press Ctrl+n twice to create two new documents entitled `new 1` and `new 2`.
2. Put the text `abc` in `new1` and the text `abcdef` in `new 2`.
3. Right-click the `new1` tab and choose _Rename_ and specify `foo` as the new name.
4. Similarly, Right-click the `new2` tab and choose _Rename_ and specify `foo` as the new name.
5. Make the _first_ `foo` tab the active one by clicking the tab.
6. Run a simple _Find All in All Opened Documents_ search for `abc` and observe:
image
7. Verify that the leftmost `foo` tab is still the active one.
8. Double-click `abcdef` in the _Find result_ window.

### Expected Behavior

The second (rightmost) `foo` tab will be made active to show the `abcdef` match.

### Actual Behavior

The second (rightmost) `foo` tab was **NOT** made active to show the `abcdef` match.
The user is still viewing the first (leftmost) `foo` tab and only sees `abc`.

### See Also 

The comment here:  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9115#issuecomment-723301529

### Debug Information
Notepad++ v7.9.1   (64-bit)
Path : C:\\.......\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1518
Current ANSI codepage : 1252

sasumner:
Fixes #9119

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1119,6 +1119,7 @@ Continue?"/> <!-- HowToReproduce: when you openned file is modified and saved, t
             <DocReloadWarning title="Reload" message="Are you sure you want to reload the current file and lose the changes made in Notepad++?"/>
             <FileLockedWarning title="Save failed" message="Please check whether if this file is opened in another program"/>
             <FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/> <!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
+            <RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use on another tab."/> <!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
             <DeleteFileFailed title="Delete File" message="Delete File failed"/> <!-- HowToReproduce: this message prevents from system failure. It's hard to reproduce. -->
 
             <NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1119,7 +1119,7 @@ Continue?"/> <!-- HowToReproduce: when you openned file is modified and saved, t
             <DocReloadWarning title="Reload" message="Are you sure you want to reload the current file and lose the changes made in Notepad++?"/>
             <FileLockedWarning title="Save failed" message="Please check whether if this file is opened in another program"/>
             <FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/> <!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
-            <RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use on another tab."/> <!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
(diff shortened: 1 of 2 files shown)


Yaron10:
@sasumner,

Very nice! Thank you. :+1: 

I've just quickly tested it.

1. How about changing "The specified name is already in use on another tab." to "The specified name is already in use."?
The user might try to "rename" the current "Name" to "Name".
2. Interestingly, "new 1" is not allowed even if there currently isn't a "new 1". That's not the case with "new 2". Not a big deal. :)

sasumner:
>The user might try to "rename" the current "Name" to "Name"

That would be fairly dumb, but...I'm sure there are dumb users out there.
I will change it to your suggestion.
It also covers the use in the following:

>"new 1" is not allowed even if there currently isn't a "new 1"

Don't you remember the invisible "new 1"s ?
Yea, I don't think it is worth bothering about.

donho:
> That would be fairly dumb, but...I'm sure there are dumb users out there.

Agree. I'm part of dumb users, that's why Notepad++ is popular - because I made it for me and other dumb people.
;)

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1119,6 +1119,7 @@ Continue?"/> <!-- HowToReproduce: when you openned file is modified and saved, t
             <DocReloadWarning title="Reload" message="Are you sure you want to reload the current file and lose the changes made in Notepad++?"/>
             <FileLockedWarning title="Save failed" message="Please check whether if this file is opened in another program"/>
             <FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/> <!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
+            <RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use."/> <!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
```
donho, line 1122: Let's use explicit (dumb) one:
"The specified name is already in use on another tab."

PowerEditor/src/NppIO.cpp
```diff
@@ -1741,23 +1741,40 @@ bool Notepad_plus::fileRename(BufferID id)
...
+			{
+				sameNamedBufferId = _pNonDocTab->findBufferByName(tabNewName);
+			}
+			
+			if (sameNamedBufferId != BUFFER_INVALID)
 			{
-				generic_string oldBackUpFile = buf->getBackupFileName();
+				_nativeLangSpeaker.messageBox("RenameTabTemporaryNameAlreadyInUse",
+					_pPublicInterface->getHSelf(),
+					TEXT("The specified name is already in use."),
```
donho, line 1754: Let's use explicit (dumb) one:
"The specified name is already in use on another tab."

donho (commented):
Let's use explicit (dumb) one:
"The specified name is already in use on another tab."

My reason is not only English native speaker use English, but also the other non-English speakers in IT/Software domain prefer to use Software in English. So it's better to be as explicit as possible.

PowerEditor/src/NppIO.cpp
sasumner: @donho 
What you want is how the FIRST commit is.
Can you just accept the first commit and ignore the second?

donho: @sasumner OK, cool.
I'll do a cherry pick.

Yaron10:
@donho,

> I'm part of dumb users, that's why Notepad++ is popular - because I made it for me and other dumb people.

Notepad++ wouldn't have been popular at all if its users were as "dumb" as you. :)
(To avoid any misunderstanding: that was a compliment).

Regarding the error message: reading your comments, I think we had some misunderstanding here. But it's not that important.

donho:
@Yaron10 
Yes, I misread your comment indeed. :)

Issue #9125, opened 2020-11-08
Caret not visible if "Go to -> Offset" is EOL


Yaron10:
**STR:**
Open a file formatted in "Windows (CR LF)". 
Open the "Go to" dialog.
Select "Offset", enter an EOL position and run it.
(For example: `149` in `stylers.model.xml`).

**Result:**
The caret is not visible.

(It also happens when opening NPP from the Command Line with the argument `-p`).
***
Notepad++ v7.9.1   (32-bit)
Build time : Nov  2 2020 - 01:03:56
Path : C---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

sasumner:
>(For example: 149 in stylers.model.xml).

So for Windows (CRLF) files, the position between the CR and the LF is treated specially.  In other words, you can't normally "get there".
You can see an example of this by turning visible line endings on and trying to move the caret through the line-ending part.  
It won't place the caret between the CR and the LF.

Maybe this is another example of "what is position, really?"
Scintilla has one notion; perhaps I have a different notion.
Notepad++ might have something different as well. :-)

I think "position" should be purely user-level.  No notion of a unicode character (potentially) taking up 4 "units" of position, and perhaps no notion of CRLF in Windows files taking up 2 "units" of position??  Hmm.

More about position here: #9095 
More about related things (encoding/position/bytes) here: https://community.notepad-plus-plus.org/topic/20218/tests-and-impressions-on-the-view-summary-functionality
and here: https://community.notepad-plus-plus.org/post/59379

Yaron10:
> So for Windows (CRLF) files, the position between the CR and the LF is treated specially. In other words, you can't normally "get there".
You can see an example of this by turning visible line endings on and trying to move the caret through the line-ending part.
It won't place the caret between the CR and the LF.

I realized that.
But the caret must be visible as is the case in "trying to move the caret through the line-ending part".
Perhaps the simplest solution should be programmatically executing "Arrow-Right" and then "Arrow-Left". 

> Maybe this is another example of "what is position, really?"
Scintilla has one notion; perhaps I have a different notion.
Notepad++ might have something different as well. 

[...]

PR #9129, opened 2020-11-09
Disallow Goto dlg offset option from moving to position inside multibyte char or between CR and LF

### LINKED ISSUE #9101: The "Search > Go to..." feature should not allow moving inside a multi-byte encoding of a character !
**Remark** : This issue has been **first** noticed by **`Alan Kilborn`**, revisited by **`Peter Jones`** and **discussed** in that topic :

https://community.notepad-plus-plus.org/post/59397

### Description of the Issue

When using the **`Search > Go to...`** feature, with the **`Offset`** option **ticked**, the different **offsets** corresponding to each **byte** of a **multi-bytes** encoding, after the **first** one, should be **inaccessible** !

- Open a new tab in N++

- if, necessary, use the **`Encoding > Convert to UTF-8`** to get an **empty** **`UTF-8`** encoded file

- Just type in the text **`A👨Z`** on the **first** line

Note that, as the **`emoji`** **MAN** 👨 is the **Unicode** character of code-point **`U+1F468`**, we can describe this line, in an **`UTF-8`** **encoded** file, as :

~~~z
Characters :   A         👨        Z
Bytes      :   41  F0  9F  91  A8  5A
Offset     :   0   1   2   3   4   5
~~~

- If you move the caret right **before** the **`A`** char, the **`Search > Go to...`** feature says you're at offset **`0`**

- If you move the caret right **before** the **`👨`** char, the **`Search > Go to...`** feature says you're at offset **`1`**

- If you move the caret right **before** the **`Z`** char, the **`Search > Go to...`** feature says you're at offset **`5`**

All these offsets are **correct**. But these values **should** be the **only** possible **offsets** to type in in the **`You want to go to`** zone !

### Actual Behavior

Now, let's force a move to offset **`3`**, exactly in the **middle** of the **multi-bytes** sequence of the **emoji** char ( byte **`91`** ) and then click on the **`Go`**  button

- Seemingly, the caret seems right **before** the **`Z`** letter. In fact :
 
    - If you hit the **`Backspace`** key, you get the text A**x91xA8**Z, so the **first two** bytes of the encoding **`xF0x9F`**, **before** the offset, are **deleted**

    - If you hit the **`Delete`** key, you get the text A**xF0x9FxA8**Z, so the next **`x91`** byte, **after** the offset, is **deleted**

- In addition, as you can see, the action of the **two** keys **`Backspace`** and **`Delete`** are **not** *symmetrical* as the former deletes **two** bytes ( the beginning of the **multi-bytes** sequence ) whereas the latter just deletes **one** byte ( **`x91`** )

### Expected Behavior

The offsets values, relative to the **individual** bytes of a multi-bytes sequence, **after** the **`1`** **byte**, in a **Unicode** encoded file, should **not** be allowed ! For instance, in the example above, the **allowed** values should be, **exclusively**, **`0`**,  **`1`** and **`5`**

Then, the **`Backspace`** and **`Delete`** would just act on **one** character, only, as **expected** !

Best Regards,

guy038

sasumner:
Fixes #9101 and #9125 

Quick reproduction methods:

for 9101:
UTF-8 document containing this single character: 𝅙
use goto dialog (Ctrl+g) to move to offset 1
type any character, e.g. `a`
observe data corruption:
image

for 9125:
UTF-8 document with Windows format with 2 lines, no characters on line 1 (just a CRLF line-ending)
turn visible line endings on
use goto dialog (Ctrl+g) to move to offset 1
type any character, e.g. `a`
observe data corruption in line-endings:
image

The fix moves the caret to the nearest position after the desired position that puts the caret on a real character border.  Other types of fixes are certainly possible, if desired.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/GoToLineDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/GoToLineDlg.cpp
@@ -62,11 +62,19 @@ INT_PTR CALLBACK GoToLineDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM)
 						}
 						else
 						{
-							auto sci_line = (*_ppEditView)->execute(SCI_LINEFROMPOSITION, line);
+							int posToGoto = 0;
+							if (line > 0)
+							{
+								// make sure not jumping into the middle of a multibyte character
+								// or into the middle of a CR/LF pair for Windows files
+								auto before = (*_ppEditView)->execute(SCI_POSITIONBEFORE, line);
+								posToGoto = static_cast<int>((*_ppEditView)->execute(SCI_POSITIONAFTER, before));
+							}
+							auto sci_line = (*_ppEditView)->execute(SCI_LINEFROMPOSITION, posToGoto);
 							(*_ppEditView)->execute(SCI_ENSUREVISIBLE, sci_line);
-							(*_ppEditView)->execute(SCI_GOTOPOS, line);
+							(*_ppEditView)->execute(SCI_GOTOPOS, posToGoto);
 						}
-                    }
+					}
 
 					SCNotification notification = {};
 					notification.nmhdr.code = SCN_PAINTED;


Yaron10:
@sasumner,

I was not by my PC in the last couple of days.

Thank you for this fix. :+1: 

> Other types of fixes are certainly possible, if desired.

Do you mean `-p` from the command-line?

sasumner:
>>Other types of fixes are certainly possible, if desired.

>Do you mean -p from the command-line?

No, I suppose I meant putting up a message box "Invalid offset" or something to that effect.
I didn't do this for 2 reasons:

* currently one can enter an offset that is beyond the end-of-file offset and no msg box error results

* the continuing debate over what "offset" really means may drive near-future changes, so why bother to make this really polished now

BUT... I should go off and examine `-p` now for the same problem.

Yaron10:
> currently one can enter an offset that is beyond the end-of-file offset and no msg box error results
...
the continuing debate over what "offset" really means may drive near-future changes, so why bother to make this really polished now

Good points.

> BUT... I should go off and examine -p now for the same problem.

Thank you. :+1:

PR #9133, opened 2020-11-11
update korean.xml (7.9 -> 7.9.1)

yunbj:
update korean.xml. (based on 0405e230)


yunbj:
I will fix it and update file for 7.9.2

PR #9143, opened 2020-11-13
Prevent corruption possibility when using -p command line parameter

### LINKED ISSUE #9142: Specifying -p on command line can lead to data corruption
### Description of the Issue

Similar to the data corruption that can happen with the _Goto_ dialog's _Offset_ option (see #9101 and  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9129#issuecomment-726063497 ) , the `-p` command line parameter can set the current position between the bytes of a multibyte-encoded UTF-8 character, or between the bytes of a Windows' line-ending of a CRLF pair.  This should not be allowed to occur.

Multibyte UTF-8 characters should be be considered "atomic" (strong feeling about) and so should Windows' line-endings (less strong of a feeling, but still fairly strong).

1. Turn visible line-endings on via _View_ menu > _Show symbol_ > _Show End of Line_
2. Open `test_4byte_utf8.txt` file (attached, below), observe the UTF-8 character (after zooming):
image
3. Optional, using HexEditor plugin, look at hex view, observe:
image
4. Open `test_crlf.txt` file (attached, below), observe:
image
5. Close all files; quit Notepad++
6. Run the command line: `notepad++.exe -p1 test_4byte_utf8.txt` using the attached file of the same name
7. After the file loads but before doing anything else, type `a`
8. Observe data is corrupted as the 4-byte UTF-8 character has been split:
image
9. Repeat steps 5 through 7 using the `test_crlf.txt` file instead of the `test_4byte_utf8.txt` file in step 6.
10. Observe line-endings, which should be CRLF, are "corrupted"; one line-ending is CR, the other is LF:
image

### Expected Behavior

No data corruption.

### Actual Behavior

The data corruption shown in steps 8 and 10.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Path : C:\\........\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1518
Current ANSI codepage : 1252

### Test files
[test_4byte_utf8.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5533419/test_4byte_utf8.txt)
[test_crlf.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5533424/test_crlf.txt)

sasumner:
Fix #9142

...and some localized space-indentation / poor-formatting. :-)

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5914,24 +5914,28 @@ std::vector<generic_string> Notepad_plus::loadCommandlineParams(const TCHAR * co
 		}
 
 		if (ln != -1 || cpos != -1)
-		{	//we have to move the cursor manually
+		{
+			//we have to move the cursor manually
 			int iView = currentView();	//store view since fileswitch can cause it to change
 			switchToFile(bufID);	//switch to the file. No deferred loading, but this way we can easily move the cursor to the right position
 
 			if (cpos != -1)
 			{
+				// make sure not jumping into the middle of a multibyte character
+				// or into the middle of a CR/LF pair for Windows files
+				auto before = _pEditView->execute(SCI_POSITIONBEFORE, cpos);
+				cpos = static_cast<int>(_pEditView->execute(SCI_POSITIONAFTER, before));
 				_pEditView->execute(SCI_GOTOPOS, cpos);
 			}
-            else
-			if (cn == -1)
+			else if (cn == -1)
 			{
 				_pEditView->execute(SCI_GOTOLINE, ln-1);
 			}
-            else
-            {
-                auto pos = _pEditView->execute(SCI_FINDCOLUMN, ln-1, cn-1);
-                _pEditView->execute(SCI_GOTOPOS, pos);
-            }
+			else
+			{
+				auto pos = _pEditView->execute(SCI_FINDCOLUMN, ln-1, cn-1);
+				_pEditView->execute(SCI_GOTOPOS, pos);
+			}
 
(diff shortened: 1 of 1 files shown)


sasumner:
@donho You were very quick on this one.
I realized that `-p0` will result in the position after the first character being set (I won't call that position 1 !).

Yaron10:
> I realized that -p0 will result in the position after the first character being set (I won't call that position 1 !).

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/26428bd2ba8df757091e2a9732dc12ba1a042fc4/PowerEditor/src/Notepad_plus.cpp#L5931

Can you try `cpos = static_cast<int>(_pEditView->execute(SCI_POSITIONAFTER, before -1));`?

**Pros:**
- `-p0` will result in "Pos: 1".
- It will match "Pos" in the status bar.
- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9125#issuecomment-723668973: the caret *is* at the end of that line. IOW: simulate manually placing the caret there.

**Cons:**
- "Go to" dialog would have to be adjusted.
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9129#issuecomment-726063497: the continuing debate over what "offset" really means may drive near-future changes, so why bother to make this really polished now.

donho:
@sasumner 
> @donho You were very quick on this one.
I realized that -p0 will result in the position after the first character being set (I won't call that position 1 !).

No problem. Just do another PR (without issue it's OK).

sasumner:
@donho said:

>No problem. Just do another PR (without issue it's OK).

Completed with #9150

Issue #2435, opened 2016-10-17
HTTP links get high lighted, but context menu has no "copy link" option


ronww:
In a file with HTTP links, the text of the link is high lighted, but the context menu does not have a "copy link" action.
### Steps to Reproduce the Issue
1. Create a file and enter some sample text.
2. Include the text: http://example.com/file.html
3. Click else where in file
4. Hover mouse cursor over link text
5. Observe link text being high lighted
6. Right click on link text to get context menu
   7/ Observer there is no "copy link" option
### Expected Behavior

that the context menu would have a "copy link" option
### Actual Behavior

there was no "copy link" option
### Debug Information

Notepad++ v6.9.2
Build time : May 18 2016 - 00:34:05
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7
Plugins : mimeTools.dll NppConverter.dll NppExport.dll NppFTP.dll NppTextFX.dll PluginManager.dll RunMe.dll XMLTools.dll

MAPJe71:
The highlight of the links might just indicate that you have `Clickable Link Settings` enabled.
Double click on a link and the applicable page is opened in your default browser.

To copy the link: select the link text (left-mouse-down just before left most character of link text, drag mouse to right side of right most character in link text, let go of mouse button) and use Ctrl+C or context-menu's copy.

ronww:
Thanks for your reply.

Yes, I know that, but "copy link" in the context menu is an increasingly
common feature. NP++ Seems to have the link recognition part working, so
makes sense to make the feature fully functional.

Mxywes:
Any updates on this feature? Or is it possible to implement it as a plugin?

KarenArzumanyan:
Good suggestion. I also need to copy the link. From the context menu it would be very good.

RyouMisaki:
I also try everytime to find "copy url/link" option on links, since I mostly use private tabs on the browser, please add "copy link" option to context menu, it would be great value added

sasumner:
@Uhf7 What do you think? You are the hyperlinks guru at the moment. :-)

Uhf7:
@sasumner, it's a wish which I would share.

There are two ways:
- An extra entry in the context menu, which is only visible if there is a link under the cursor.
[...]

PR #9154, opened 2020-11-14
Context menu with "Copy link" option.

Uhf7:
Implements #2435.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -160,6 +160,7 @@ The comments are here for explanation, it's not necessary to translate them.
                     <Item id="42048" name="Copy Binary Content"/>
                     <Item id="42049" name="Cut Binary Content"/>
                     <Item id="42050" name="Paste Binary Content"/>
+                    <Item id="42082" name="Copy link"/>
                     <Item id="42037" name="Column Mode..."/>
                     <Item id="42034" name="Column Editor..."/>
                     <Item id="42051" name="Character Panel"/>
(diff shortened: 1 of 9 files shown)


Yaron10:
@Uhf7,

Thank you for another great PR. :+1: 

NPP enables the user to customize the context-menu (Menu -> Settings -> Edit Context Menu).
Would it be possible to control *both* the appearance and position of "Copy Link" via `contextMenu.xml`?

If NOT, I think the "Copy Link" addition should be optional.
And regarding the position: IMO it would be better to add a separator + "Copy Link" to the end instead of replacing "Copy".
(For my personal use it's even more important: I use the "MenuIcons" plugin and it works in the context-menu only if "Copy" is there). 

Should it require a lot of extra work?

Uhf7:
@Yaron10, thank you for the feedback.

I intentionally tried it this way to avoid too much impact on the infrastructure. The basic idea is, that "Copy" is useless without selection, disabled anyway in this case, so simply make it to "Copy link", if appropriate.

_Configuration:_ Since it replaces "Copy", "Copy link" appears instead of "Copy" where ever you configure it in the `contextMenu.xml`. If there is no "Copy" command in the `contextMenu.xml`, then no "Copy link" appears. If there are no clickable links enable in Settings->MISC, no "Copy link" appears either. That is the plan and seems enough configuration for me.

_MenuIcons plugin:_ I just tried the `MenuIcons_dll_1v22_x32.zip` from the GitHub repository (https://github.com/francostellari/NppPlugins), and it worked perfect. The icons seem to be identified by the file names, and if I copy a `Copy Link.ico` file in the `MenuIcons\!` folder of this plugin, then this icon gets displayed beside the "Copy link" menu entry.

Yaron10:
@Uhf7,

> I intentionally tried it this way to avoid too much impact on the infrastructure.

That's a good point.

> Since it replaces "Copy", "Copy link" appears instead of "Copy" where ever you configure it in the contextMenu.xml.

*So, the "MenuIcons" plugin shouldn't be a consideration at all.* Thank you for trying it.
(It does work with the original `contextMenu.xml`, but doesn't with some [customization](https://github.com/notepad-plus-plus/notepad-plus-plus/files/5542897/contextMenu.zip)).

donho:
I love how it changes dynamically - high value on the usability. Bravo @Uhf7 !

PR #9158, opened 2020-11-14
Finder dialog needs new parent

### LINKED ISSUE #8199: "Find in these found results" does nothing when first "Find result" window opened via macro execution
### Description of the Issue

If a macro is run that causes the first output to the "Find result" window in a Notepad++ run, then right-clicking in the "Find result" window and choosing "Find in these found results" results in nothing further happening with that.

1. Record and save a macro that causes data to be written to the "Find result" window (the simplest being a "Find All in Current Document" search).
2. Restart Notepad++
3. Play the previously saved macro.
4. Right-click in the "Find result" window and choose "Find in these found results..." from the popup menu.

### Expected Behavior
After step 4 is executed, I expect to see this dialog box appear:

image

### Actual Behavior
Nothing happens (no dialog opens) after step 4.

### Debug Information
Notepad++ v7.8.6   (32-bit)
Path : C:\.........\npp.7.8.6.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039

Uhf7:
Fixes #8199. 

Problem: All dialogs which inherit from `DockingDlgInterface` have `_pPublicInterface->getHSelf()` as parent, except the `Finder` dialog.

This has drawbacks:
- The display method of the `Finder` dialog doesn't work, that's why it has been replaced by `SendMessage NPPM_DMMSHOW/HIDE` calls to the parent of Find/Replace dialog in many places
- All messages for the public interface are sent to parent of parent instead of parent
- If the `Finder` starts before the Find/Replace dialog has been displayed (see #8199 how to do this), then the `Finder` dialog has no parent.

This PR gives the `Finder` docking dialog the parent window it is entitled to: the public interface window.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -544,7 +544,7 @@ void Finder::gotoFoundLine()
 	const FoundInfo fInfo = *(_pMainFoundInfos->begin() + lno);
 
 	// Switch to another document
-	if (!::SendMessage(::GetParent(_hParent), WM_DOOPEN, 0, reinterpret_cast<LPARAM>(fInfo._fullPath.c_str()))) return;
+	if (!::SendMessage(_hParent, WM_DOOPEN, 0, reinterpret_cast<LPARAM>(fInfo._fullPath.c_str()))) return;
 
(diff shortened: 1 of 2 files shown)


donho:
The Finder has a better parent now 👍 !

PR #9159, opened 2020-11-15
Publish MinGW Build Instructions

### LINKED ISSUE #7743: Publish MinGW Build Instructions

donho:
Close #7743

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -63,24 +63,25 @@ To build SciLexer.dll without PCRE support (for both 32-bit & 64-bit):
 Note: If building the 32-bit and 64-bit versions in the same folder structure, after building one, it may be necessary to do a `nmake -f scintilla.mak clean` before building the other.
 
 
-## Build `Notepad++` & `SciLexer.dll` with GCC:
+## Build 64 bits binaries with GCC:
 
-If you have installed [MinGW](http://www.mingw.org/), then you can compile Notepad++ binary & SciLexer.dll with GCC.
+If you have installed [MinGW-w64](https://mingw-w64.org/doku.php/start), then you can compile Notepad++ & SciLexer.dll 64 bits binaries with GCC.
 
 * Compile Notepad++ binary
 
 1. Launch cmd.
 2. Change dir into `notepad-plus-plus\PowerEditor\gcc`.
-3. Type `mingw32-make.exe`
+3. Type `mingw32-make.exe -j%NUMBER_OF_PROCESSORS%`
 4. `NotepadPP.exe` is generated in `notepad-plus-plus\PowerEditor\bin\`.
 
 * Compile SciLexer.dll
 
 1. Launch cmd.
 2. Change dir into `notepad-plus-plus\scintilla\win32`.
-3. Type `mingw32-make.exe`
+3. Type `mingw32-make.exe -j%NUMBER_OF_PROCESSORS%`
 4. `SciLexer.dll` is generated in `notepad-plus-plus\scintilla\bin\`.
 
-You can download MinGW from https://sourceforge.net/projects/mingw-w64/files/. On Notepad++ Github page (this project), the build system use [MinGW 8.1](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z).
+You can download MinGW-w64 from https://sourceforge.net/projects/mingw-w64/files/. On Notepad++ Github page (this project), the build system use [MinGW 8.1](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z).
 
-Website: https://mingw-w64.org/doku.php/start
+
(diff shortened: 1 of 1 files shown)


donho:
@chcg 
Could you review the added part to make sure it's accurate please?

ArkadiuszMichalski:
Ok, I finally compile it, 2 things:
- when we use MinGW from the package (7z) whe need manually add the `MinGW/bin` folder to system Path variable to make `mingw32-make.exe` invoke works (in the case what we see in above instruction).
- is there any quick tip to compile 32-bit NPP?

donho:
@ArkadiuszMichalski 
> when we use MinGW from the package (7z) whe need manually add the MinGW/bin folder to system Path variable to make mingw32-make.exe invoke works (in the case what we see in above instruction).

True. Added the notice in 
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9159/commits/a3791c7d0b42d1966ae4c8495327d6254adb307c

> missing -j%NUMBER_OF_PROCESSORS% but I don't see any differences with / without it

Added also in above commit.
It's for optimizing the performance of binary, I guess the default value is 2 - so it could be slower (that you can't even feel about it).

> is there any quick tip to compile 32-bit NPP?

We don't provide 32 bit instructions since MinGW build is not for the production and since there's 64 bits (better).
But I believe the instructions are similar, if you download the right tool chain from http://www.mingw.org/

ArkadiuszMichalski:
@donho `%PATH%` is always finished by `;` (semicolon)? I can't check because I add some and don't know what is default state.
So `set PATH=%PATH%C:\xxxx\mingw64\bin` maybe be wrong, but `set PATH=C:\xxxx\mingw64\bin;%PATH%` is safer (take from https://en.wikipedia.org/wiki/PATH_(variable)#DOS.2C_OS.2F2.2C_and_Windows).

BUILD.md
```diff
@@ -61,3 +61,27 @@ To build SciLexer.dll without PCRE support (for both 32-bit & 64-bit):
...
+4. `NotepadPP.exe` is generated in `notepad-plus-plus\PowerEditor\bin\`.
+
+* Compile SciLexer.dll
+
+1. Launch cmd.
+2. Change dir into `notepad-plus-plus\scintilla\win32`.
+3. Type `mingw32-make.exe -j%NUMBER_OF_PROCESSORS%`
+4. `SciLexer.dll` is generated in `notepad-plus-plus\scintilla\bin\`.
+
+You can download MinGW-w64 from https://sourceforge.net/projects/mingw-w64/files/. On Notepad++ Github page (this project), the build system use [MinGW 8.1](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z).
```
chcg, line 84: For 32bit build https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/sjlj/i686-8.1.0-release-posix-sjlj-rt_v6-rev0.7z could be used.

chcg (approved):

chcg:
- With the -j option the number of parallel make jobs could be influenced. So the output shouldn't be different apart from a faster build time.
- Path variable should be ending in a semicolon but adding one and placing the compiler at the front of the path var might be good to avoid that other settings of the user like e.g. cygwin are causing conflicts.

donho:
@chcg Thank you for your review and information.
I think it's better to provide only the info about what we have in our build system.

donho:
> It's for optimizing the performance of binary, I guess the default value is 2 - so it could be slower without indicate it (that you can't even feel about it).

So it's rather for optimizing the performance of compiling, not binary - Thank you @chcg for the correction.

ArkadiuszMichalski:
Yes, it's for performance of compiling, but I don't see any differences with / without it. Maybe the Notepad ++ project is too small to notice this. I saw some discussion that `-j` not always works, but it's old (https://stackoverflow.com/questions/9420825/how-to-compile-on-multiple-cores-using-mingw-inside-qtcreator/9421254). In practice, the second compilation is fast (probably a lot of data has been cached in RAM) so it's not a problem.

sasumner:
> it's for performance of compiling, but I don't see any differences with / without it.

If your computer is older, it can't take full advantage of this, maybe?

ArkadiuszMichalski:
Maybe, from 2011, i7 2600K (4/8) and Win7, no difference in core load. In later I will check it on Win10 with something newer from 2019 (but still 4/8).

@sasumner Do you see the difference after using `-j` on your machine?

sasumner:
>@sasumner Do you see the difference after using -j on your machine?

I haven't had time to try it yet; hopefully soon.

donho:
@sasumner 
> I haven't had time to try it yet; hopefully soon.

You have claimed this tutorial, you have to test it.
(Muhahahaha...)

sasumner:
>You have claimed this tutorial, you have to test it.

Oh, I will.
And by the time I get to it, others will have worked out any bugs. :-)

donho:
Bugs ?
There's no bug in Notepad++, only the random features :)

ArkadiuszMichalski:
@donho Compiled SciLexer.dll in this method is without boost?

donho:
@ArkadiuszMichalski  
> Compiled SciLexer.dll in this method is without boost?

No, without boost - as the build in MSVC.

PR #9191, opened 2020-11-22
Makes find next/prev result work on find in finder results

### LINKED ISSUE #9080: Next/Prev Search Results don't act on current Find-result window
### Description of the Issue
The menu commands _Next Search Result_ and _Previous Search Result_ act only on the original _Find result_ window, not the active _Find result_window.

1. Create and save a file with the following text:
```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.

Up Jack got
And home did trot,
As fast as he could caper;
And went to bed
And plastered his head
With vinegar and brown paper.

When Jill came in
How she did grin
To see Jack's paper plaster;
Mother vexed
Did whip her next
For causing Jack's disaster.
```
2. Do a simple _Find All in Current Document_ search for the text `Jack`
3. Observe:
image
4. Right-click in the client area of the _Find result_ window and choose _Find in these found results..._
5. Untick all checkboxes and search for `Jill`
6. Observe:
image
7. Click in the main editor window
8. Use the menu commands _Next Search Result_ and _Previous Search Result_ (or F4 / Shift+F4, respectively), several times, randomly.

### Actual Behavior
Commands in step 8 moved between results of the first search, finding `Jack`

### Expected Behavior
Commands in step 8 should move between results of the second search for `Jill`, since it is the _Find result_ window that is active and currently showing to the user.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Build time : Oct 27 2020 - 05:04:37
Path : C:\\..........\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1457
Current ANSI codepage : 1252

clachen:
Fixes #9080

This is a one line fix that fixes the issue whereby the "find next result" button does not work when searching inside the result found by a previous search, and instead, showed the original search result's search terms. 
Now it works as expected, allowing us to cycle through the new search term's result instead. 

The fix works by assigning the data from  pFindersInfo->_pDestFinder which has conveniently already stored the information from finding the new find in finder result to _pFinder where it can be accessed by the Find Next Result function. 

The Find Next/Prev button has not been changed, and as such it will still cycle through the initial search result. For example if you search for "Jack" in a file first, and then search for "Jill" in the search results for "Jack", the "Find Next" button will still cycle through "Jack".

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2333,7 +2333,7 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 		else if (op == ProcessFindInFinder)
 		{
 			if (pFindersInfo && pFindersInfo->_pDestFinder)
-				pFinder = pFindersInfo->_pDestFinder;
+				_pFinder = pFindersInfo->_pDestFinder;
 			else
 				pFinder = _pFinder;
 		}
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -2332,8 +2332,11 @@ int FindReplaceDlg::processRange(ProcessOperation op, FindReplaceInfo & findRepl
 		}
 		else if (op == ProcessFindInFinder)
 		{
-			if (pFindersInfo && pFindersInfo->_pDestFinder)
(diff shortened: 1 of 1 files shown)


donho:
By assigning `_pFinder` to this volatile variable, you also put the other options we don't want into this object.
So it may "fix" the "bug" (which is not considered as bug to me:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9080#issuecomment-735426267 )
but it brings more regressions.

Issue #9155, opened 2020-11-14
svn urls not links since 7.9.1


nikita-leontiev:
### Description of the Issue
Since 7.9.1 svn urls are not links any more, so can't be open, for example, via TortoiseSVN.
Suppose related to https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9090

### Steps to Reproduce the Issue
Add any svn URL to editor, for example: svn://svn.notepad-plus-plus.org

### Expected Behavior
Link.

### Actual Behavior
Simple text.

### Debug Information
Notepad++ v7.9.1   (32-bit)
Build time : Nov  2 2020 - 01:03:56
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS Name : Windows 7 Home Premium (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1251
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll

donho:
In order to fix several bugs of URL link, the parsers of URL has been rewritten (by @Uhf7).
The current implementation handles much better URL. OTOH, it has some limit to support other protocol, like `svn://`.
@Uhf7 please correct/complete if what I mentioned above is not accurate.

Uhf7:
That's correct so far. The previous regex based parser accepted nearly everything as scheme name, the new version limits the allowed scheme names to the schemes which can be handled by the installed web browser and the installed email software.

`svn:` is a custom URL scheme, which may be installed on computers, but it is not available "out of the box". That may be one reason, why custom URL schemes are not detected by MS VC++, MS Outlook or MS Word. The other reason is, that every URL scheme actually has it's own format, although the `svn:` scheme format looks very similar to the `http:` scheme format.

How to get custom schemes supported? I can imagine a configurable list of custom schemes, which would then be handled like the `http:` scheme, if this is desperately needed. I cannot imagine to parse each and every custom scheme installed on a given computer or even check for the custom schemes installed. Many schemes I see on my computer seem to be for internal use by the software which installed them.

FreeAndNil:
@Uhf7 How about a checkbox for enabling the old behavior?

donho:
@Uhf7 
[...]

PR #9195, opened 2020-11-22
Add GUI in preferences dialog for adding URI customized schemes

donho:
Related: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5168bdb7c5395821e458aaf20297cb7e8c76ef02 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9155


PowerEditor/src/Parameters.h
```diff
@@ -849,7 +849,7 @@ struct NppGUI final
 	bool _isWordCharDefault = true;
 	std::string _customWordChars;
 	urlMode _styleURL = urlUnderLineFg;
-	generic_string _customizedShemes = TEXT("svn:// cvs:// git:// imap:// irc:// irc6:// ircs:// ldap:// ldaps:// news: telnet:// gopher:// ssh:// sftp:// smb:// skype: snmp:// spotify: steam:// sms: slack:// chrome:// bitcoin:");
+	generic_string _uriShemes = TEXT("svn:// cvs:// git:// imap:// irc:// irc6:// ircs:// ldap:// ldaps:// news: telnet:// gopher:// ssh:// sftp:// smb:// skype: snmp:// spotify: steam:// sms: slack:// chrome:// bitcoin:");
```
sasumner, line 852: @donho Spelling `_uriShemes` -> `_uriSchemes` :-)

sasumner: But you didn't change the spelling problem?  :-)

sasumner:
@donho

Okay, so I didn't actually review the code so much as try out the feature.

Taking a cue from [HERE](https://community.notepad-plus-plus.org/post/60031) (a user complaint about his custom linking behavior that used to work but is broken in 7.9.1), I recreated this user's setup on my system and tried to get it to work in the current (as of this writing) codebase.

I will report that I had success.

Summary:

I added this:
image

And after that I was able to achieve links in text such as these:
image
And then when clicking the links, the indicated files opened in Notepad++ for me.

I did notice that the links with pathnames must all contain forward slashes in order to be recognized as links, for example, this is NOT recognized as a link:
image

This is in contrast to earlier versions of N++ (I used 7.8.6) where backslashes WERE recognized, and mixed slashes as well, examples:
image

donho:
@sasumner 
It's about @Uhf7 's implementation of URI. As he has mentioned, it's impossible to manage all the custom URI format, so it's based on URL format - I agree with his choice.

sasumner:
>I agree with his choice.

I agree also.

PR #9197, opened 2020-11-23
update korean.xml (7.9.2)

yunbj:
fix #9182 and update


donho:
@yunbj 

PR #9210, opened 2020-11-30
Enhance -n -c -p command line args behavior

### LINKED ISSUE #9146: Command Line Parameters: -1 vs. -2
**STR:**
Open a file.
Place the caret at line 3.

Run that file from the command-line with `-n-1`.
-- The caret remains at line 3. 

Run that file from the command-line with `-n-2`.
-- The caret is at line 0.

**Expected:**
The same result in both cases.
***
Notepad++ v7.9.1 (32-bit)
Build time : Nov 2 2020 - 01:03:56
Path : C---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.23964
Current ANSI codepage : 1255

sasumner:
Fix #9146

@Yaron10 Please test

Notes:

* -p remains zero-based (we weren't out to change that with this issue/PR).
* -p has priority over the others if it is specified along with -n and/or -c

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5884,10 +5884,13 @@ std::vector<generic_string> Notepad_plus::loadCommandlineParams(const TCHAR * co
 		return std::vector<generic_string>();
 	}
 
- 	LangType lt = pCmdParams->_langType;
-	int ln =  pCmdParams->_line2go;
-    int cn = pCmdParams->_column2go;
-    int cpos = pCmdParams->_pos2go;
+	LangType lt = pCmdParams->_langType;
+	int ln = pCmdParams->_line2go;
+	bool lnPresent = pCmdParams->_line2goPresent;
+	int cn = pCmdParams->_column2go;
+	bool cnPresent = pCmdParams->_column2goPresent;
+	int cpos = pCmdParams->_pos2go;
+	bool pnPresent = pCmdParams->_pos2goPresent;
 	bool recursive = pCmdParams->_isRecursive;
 	bool readOnly = pCmdParams->_isReadOnly;
 	bool openFoldersAsWorkspace = pCmdParams->_openFoldersAsWorkspace;
@@ -5917,15 +5920,33 @@ std::vector<generic_string> Notepad_plus::loadCommandlineParams(const TCHAR * co
 			pBuf->setLangType(lt);
 		}
 
-		if (ln != -1 || cpos != -1)
+		if (pnPresent || lnPresent || cnPresent)
 		{
 			//we have to move the cursor manually
 			int iView = currentView();	//store view since fileswitch can cause it to change
 			switchToFile(bufID);	//switch to the file. No deferred loading, but this way we can easily move the cursor to the right position
 
-			if (cpos != -1)
+			auto currPos = _pEditView->execute(SCI_GETCURRENTPOS);
+
+			if (pnPresent)
 			{
-				if (cpos > 0)
+				int docLength = static_cast<int>(_pEditView->execute(SCI_GETLENGTH));
+				if (docLength == 0)
+				{
+					cpos = 0;
+				}
+				else
+				{
+					if (cpos < 0)
+					{
+						// -1 means last position of file, 
+						// -2 means next-to-last position of file, etc.
+						cpos += docLength + 1;
+						cpos = max(cpos, 0);
+					}
+					cpos = min(cpos, docLength);
+				}
+				if (cpos != 0)
 				{
 					// make sure not jumping into the middle of a multibyte character
 					// or into the middle of a CR/LF pair for Windows files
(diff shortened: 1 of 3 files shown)


chcg:
@sasumner Fails to compile because isParamePresent is now an unused parameter and setting warning as error makes the build break.

sasumner:
@chcg Thanks, not sure how that was missed...pushing the fix for that...

ArkadiuszMichalski:
Works as specified in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9146#issuecomment-727277500. Only some things:
- now alone `-c` works fine (`-n` is set by default, same as manually passing `-n` or `-n0`)
- preceding 0 in value are ignored, so `009` is same as `9`.
- `-0` and `+0` in value is same as `0`.
- if `-p` remains zero-based then `-p` or `-p0` does not mean a remembered last position (as is the case for `-n` or `-c`), but set it to 0.

sasumner:
>-p or -p0 does not mean a remembered last position

Just to add something to this:  With `-p` there is nothing "remembered" anyway, so...

ArkadiuszMichalski:
> > -p or -p0 does not mean a remembered last position
> 
> Just to add something to this: With `-p` there is nothing "remembered" anyway, so...

Right, and if changing to the numbering from 1, then `-p` and `-p0` must be determined somehow, but it can be done in next bug (if this behavior will be changed).

sasumner:
@ArkadiuszMichalski 
Well, if positioning is changed to be 1-based, then `-p0` is an error and should probably do nothing.  Same for `-p`.

ArkadiuszMichalski:
Yes, and in practice, it would mean the last remembered position (line and column).

sasumner:
>Yes, and in practice, it would mean the last remembered position (line and column).

Well, I'm confused by what that statement is saying, I think.
If a `-p` value is fed in, and that value is an "error", then logic that changes position would be aborted at that point and no change-of-position would be done.
So...if we're saying the same thing two different ways, that's fine, but it feels like we are each saying something different.

ArkadiuszMichalski:
If `-p` will faild then where carrect will be positioned, on the begining in line 1? Actually, that's how it is now, so it could stay that way.

sasumner:
I suppose `-p1` or `-p0` could do the same (put caret at very top of file), but let's save final judgment on that for a future time/issue.

Yaron10:
Nice work. Assuming it's needed and useful. :)

**1.** `-c` -> ignored (correct). `-n` NOT ignored but moves the caret to Col 1.
**2.** I think that ignoring `-n0` is a bit weird (any other number is not ignored). I'd treat it as 1.
**3.** Why should Pos override any other args? Wouldn't it better if the first (or probably better *last*) arg overrides?

sasumner:
>Why should Pos override any other args? Wouldn't it better if the first (or probably better last) arg overrides?

It was the old behavior, just being continued.
There isn't any detection-order logic in the code; adding something like that for this purpose seems like a lot to ask.

donho:
@sasumner  Thank you for the PR.

The algorithm of current code is:
If `pos` argument is present, then got to pos (and ignore the 2 other arguments), otherwise no `pos` and no `col` then go to `line`. otherwise both `col` &  `line` must be present, then go to line & go to col. Simple and straight forward.
 
The code of this PR is much darker (more difficult to understand) comparing with the current code. From my viewpoint, there's only one added factor (positive or negative) for each argument - could the code be improved to make it more readable?

sasumner:
@donho

>code...is much darker

LOL. I've never heard of "dark code" before, but I like that phrase.

>could the code be improved to make it more readable?

Probably not.
Certainly more comments could be added, but I don't think that is what you mean.

While I feel this code "works", I don't really like this code.

So here's the sequence of events that got us here:

* I notice #9146 and think that it is unimportant because a argument to `-n` is always >= 1
* However, I remember a user that wanted a command-line way to open a file and position caret on the last line
* My Python background makes me think that `-n-1` is a nice syntax for that, and could "fix" 9146 at the same time
* Back-and-forth discussion with other users in issue comments causes feature to "bloat"
* In my desire to please everyone, I finished the "bloated" version of the feature and submitted PR

Here's my suggestions:

* Close this PR (I will do this)
* Advise a user with need to jump to last line of file to use `-n2147483647` (or similar)
* Close 9146 with the advice that "negative arguments to `-n` are not defined, and could cause undefined behavior" which is an implied suggestion of "don't do that" and that we "won't fix"

I think, moving forward, that I will work on small things that interest me, in the way that I think they should be handled.
I will submit PR, and if it can be accepted, that's good.
If changes to PR are requested, and they are very-small and reasonable to me, I will make them.
Otherwise I will just self-close the PR.
I need to concentrate on keeping it "fun". :-)

donho:
> I need to concentrate on keeping it "fun". :-)

I couldn't agree more with you! 

> In my desire to please everyone, I finished the "bloated" version of the feature and submitted PR

What I've learnt from this project is, you can't please everyone. 

> Close this PR (I will do this)

Sorry for making you code but not integrating it. And it was my fault - I should make myself more clear for the PR acceptation criteria before you coding. I'll pay more attention about it for the next time.

sasumner:
@donho 

> It was my fault

I don't think there is any blame or fault here.
It just sort of happened.
:-)

PR #9233, opened 2020-12-06
Make Line text in Search-results translatable

### LINKED ISSUE #9224: Ability to localize "Line" preceding each line number on Search Results
Hello,

When using `Find All in Current Document` or `Find All in All Opened Documents`, that kind of result is displayed:

```
Circà "margine" (3 risultati in 1 ducumenti frà 1 ducumenti circati)
  C:\Users\patriccollu\Downloads\corsican 201203.xml (3 risultati)
	Line 854:                 <MarginsBorderEdge title="Margine è bordu">
	Line 855:                     <Item id="6201" name="Margine di piegatura"/>
	Line 932:                     <Item id="6607" name="Preferenze di margine (Unita : mm)"/>
```

Everything is translated, excepted the word "**Line**" preceding each line number. Would it be possible to localize this word?

Thanks,
Patriccollu.

sasumner:
Fix #9224

Notes: 

* Changes lexer code so that it no longer looks for a four-character wide fixed token on "Line" lines of search result; it looks for the first space character following the token instead.  This allows the translation of "Line" to be less or more characters long.  Translation restriction for "Line": Must not contain a space or a colon.  [**EDIT:** Second commit lifts the restriction on a colon being in the "Line" translation string, if user wants this for some reason.]

* Prepares the translated text for "Search" and "Line" output lines only once, when a search is begun (otherwise, for "Line" it would pointlessly and repeatedly prepare the same text over and over for each "Line" of the result).

* Eliminates the styling of the space character between "Line" and the digits of the line number (see image [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9224#issuecomment-738864831) for why this was wrong and was changed) 

* Performs some additional error checking on the user's translation for "Search" so that it is now impossible for a "Search" line to appear to the lexer as a "file" line or a "Line" line.  Translation restrictions for "Search": Must not start with a space or tab.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -3435,8 +3435,7 @@ generic_string Finder::getHitsString(int count) const
 
 void Finder::addSearchLine(const TCHAR *searchName)
 {
-	NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
-	generic_string str = pNativeSpeaker->getLocalizedStrFromID("find-result-title", TEXT("Search"));
+	generic_string str = _searchResultSearchPrefix;
 	str += TEXT(" \"");
 	str += searchName;
 	str += TEXT("\" \r\n");
@@ -3527,8 +3526,11 @@ void Finder::addSearchHitCount(int count, int countSearched, bool isMatchLines, 
 
 void Finder::add(FoundInfo fi, SearchResultMarking mi, const TCHAR* foundline)
 {
+	// build up and add a line like this: "\tLine 123: xxxxxxHITxxxxxx"
+	
 	_pMainFoundInfos->push_back(fi);
-	generic_string str = TEXT("\tLine ");
+
+	generic_string str = TEXT("\t") + _searchResultLineNbPrefix + TEXT(" ");
 
 	TCHAR lnb[16];
 	wsprintf(lnb, TEXT("%d"), fi._lineNumber);
@@ -3596,16 +3598,19 @@ void Finder::wrapLongLinesToggle()
 
 bool Finder::isLineActualSearchResult(const generic_string & s) const
 {
-	const auto firstColon = s.find(TEXT("\tLine "));
-	return (firstColon == 0);
+	// actual-search-result lines are the only type that start with a tab character
+	// sample: "\tLine 123: xxxxxxHITxxxxxx"
+	return (s.find(TEXT("\t")) == 0);
 }
 
 generic_string & Finder::prepareStringForClipboard(generic_string & s) const
 {
-	// Input: a string like "\tLine 3: search result".
-	// Output: "search result"
+	// Input: a string like "\tLine 123: xxxxxxHITxxxxxx"
+	// Output:              "xxxxxxHITxxxxxx"
+
 	s = stringReplace(s, TEXT("\r"), TEXT(""));
 	s = stringReplace(s, TEXT("\n"), TEXT(""));
+
 	const auto firstColon = s.find(TEXT(':'));
 	if (firstColon == std::string::npos)
 	{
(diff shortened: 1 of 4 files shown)


ArkadiuszMichalski:
> Translation restriction for "Line": Must not contain a space or a colon.

It's worth adding it as a comment in `english.xml`. 

```<find-result-line-nb-prefix value="Line"/> <!-- here -->```

If other items also have some restrictions it is worth mentioning them in the comment. This is important for people doing translations.

**Comment by sasumner** [comment]
@ArkadiuszMichalski Good suggestion, I will add that.
I realized that a colon is actually OK in the "Line" text, so I'm going to do a slight revision to this anyway; I'm closing until I get that finished, then I'll reopen.

**Comment by donho** [comment]
@sasumner 
I do believe the code of PR is needed for the translation. But the code is the way too much to translate "Search" & "Line".
Sincerely, I don't think we do need to translate these 2 simple English words, especially with such massive code.

**Comment by sasumner** [comment]
@donho 

>I don't think we do need to translate these 2 simple English words

Let's consider `"Search"` ...

"Search" is already translated: `<find-result-title value="Search"/>`

But without some additional simple "protections" in code, some obscure bugs could result (users do crazy things) -- why not prevent them if it is easy to do?

The following screenshots were all generated from 7.9.1 code, not the PR code:

image
(Note above the "Search" line is green like a path line and the folding is wrong)

image
(Note above a lot of things messed up here)

image
(Note above it is even worse if you run more than one search)

**Comment by sasumner** [comment]
@donho 

>I don't think we do need to translate these 2 simple English words

Let's consider `"Line"` ...

Well, "Line" is just fine with me, but I am a native English speaker.

But what if I weren't?  How would I feel about it as a user?
I'd probably get annoyed every time I saw this "ugly symbol" in my "Search results":

image

I mean, just what IS that ugly symbol??  :-)

But, on a serious note, why not give the user a complete translation in this area of the UI?
The changes the PR presents are super-easy to validate.

**Comment by donho** [comment]
@sasumner 
> (Note above a lot of things messed up here)

So it's rather a bug? How can I reproduce it?

> I'd probably get annoyed every time I saw this "ugly symbol" in my "Search results":

I don't follow you... Do you mean the word "Line", these 4 characters, is an ugly symbol for non English users?

**Comment by sasumner** [comment]
>So it's rather a bug? 

So it is obscure, but I discovered the possibility of it when I was investigating the main thrust of this PR, meaning allowing "Line" to be translated.

>How can I reproduce it?

If you start this string with a space or a tab:

`<find-result-title value="Search"/>`

Example:

`<find-result-title value="         Search"/>`

Again, it is obscure, but I was "in the code" so I thought I would close the hole.

>Do you mean the word "Line", these 4 characters, is an ugly symbol for non English users?

Isn't any foreign word ugly to someone when all of the rest is in a comfortable language?

**Comment by sasumner** [comment]
@donho 

What's the feeling, for a new issue and PR, that fixes **_ONLY_** THIS problem with the lexer:

image

**Comment by sasumner** [comment]
>But the code is the way too much to translate "Search" & "Line".
>I don't think we do need to translate these 2 simple English words, especially with such massive code.

This is "code is ... way too much" and is "massive code" ??

Here is my response to that:

* Some quantity of code in the PR could be reduced if we wanted to translate "Line" on every single line of output in the "Search result" window.  But it seems "wasteful" to do that constantly for perhaps thousands+ of potential output lines.  Why not do the translation one time per search?

* Much of the "code" is in the form of comments to help define the uncoupled interface between "PowerEditor" and "Scintilla" code.  This interface was VERY POORLY known (until I took the trouble to figure it out), so I thought I would add comments to help clarify how this works, for the betterment of the code for the future.

* The code in `Finder::isLineActualSearchResult()` could have remained unchanged, but it was misleading: It used a variable called `firstColon` but there were no colons involved at all, so I thought by changing this code it makes it make more sense.

* Comments in `Finder::prepareStringForClipboard()` did not have to be changed, but the change makes the code better

But again, there is little question of the code of this PR working correctly, so what does it matter the quantity of code changed?

**Comment by sasumner** [comment]
If it is not important to translate "Line" for the convenience of the user, let's just say that and close PR and issue.

**Comment by donho** [comment]
@sasumner 

My opinion about the translation of "Line" is here:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9224#issuecomment-740267055

> Example:
`<find-result-title value=" Search"/>`

OK, I can reproduce it.
However, I don't consider it as "bug" but rather a "PEBKC". Firstly all the official localization files contain the correct translation format (without ws/TAB before) - so such situation won't happen if the localization files are untouched. Secondly, I don't see why the translator add any ws or TAB before the translation. Even if it could happen, it gives the indicator to the translator to fix it.
 
Basically, I agree with your arguments, especially if it's crash issue or critical issue are caused, but not such minor visual issue which could happen very rarely. As a result, this PR won't be included, hope you understand my position.

**Comment by ArkadiuszMichalski** [comment]
@donho If this is too much code so why not only make this two labesl translatable (without checking correct value) and just add a comment inside `englis.xml` file to not use some chars  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9233#issuecomment-739501257. It's very odd when you have a translation system and you omit some elements.

**Comment by donho** [comment]
@ArkadiuszMichalski 

Actually it's only 1 label to translate in this PR (my mistake to say 2 labels to translate).
OK, I will give it a try (in my way).

**Comment by donho** [comment]
@ArkadiuszMichalski 

Done in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e3455a0f7e2e8aa2c97d1f7a6d2ddfde2d83f95e

**Comment by sasumner** [comment]
@donho 

> this PR won't be included, hope you understand my position

No worries, although my nice comments that could have lived forever in the code just have to live on in my personal notes.  :-)

Can you either approve or disapprove of this idea ? :  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9233#issuecomment-740088620

**Comment by donho** [comment]
@sasumner 
> What's the feeling, for a new issue and PR, that fixes ONLY THIS problem with the lexer? :

It's the problem of lexer. I will take care of this issue.

**Comment by donho** [comment]
@sasumner 
Fixed: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2f6454a085b772f17d3052e28d7f9d674218a725

File: PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp [hunk 1]
```diff
@@ -3596,16 +3598,19 @@ void Finder::wrapLongLinesToggle()
 
 bool Finder::isLineActualSearchResult(const generic_string & s) const
 {
-	const auto firstColon = s.find(TEXT("\tLine "));
```
sasumner, line 3599: Just a comment so that I can get a link directly to "here"

Issue #8338, opened 2020-05-28
 [Feature Request] Add tabs counter in visible place


ArkadiuszMichalski:
Please consider adding a tabs counter in somewhere in a visible place, for example:
1. Before `X` button in Menu bar.
2. Or somewhere in Status bar.
3. Or any other better visible place.

This is a very convenient improvement for people who work on a large number of tabs, and sometimes by mistake we close some tab/tabs. With tabs counter we can easy control if our sets of tabs is complete.

Format for tabs counter can be just `Number` or `Number/Number` (for two views), or more obvious like `Tabs: Number` or `Tabs: Number/Number` in Status bar. This can also be optional for people who don't need it (some checkbox in Preferences).  In detail, this can be done in different ways, it is more important to accept this future.

ArkadiuszMichalski:
We have information about: numbers of chars, numbers of lines, numbers of words, numbers of chars in selections, etc., but not how many files we have opened. At least in this window, which is dedicated for tabs.
![image](https://user-images.githubusercontent.com/2730894/83683092-1fede680-a5e5-11ea-9044-8a3665768409.png)
Let it be so, I will look for a solution through a plugin.

sasumner:
>I will look for a solution through a plugin.

A scripting plugin could easily do it; example with Pythonscript, with 4 open tabs in Notepad++:

![image](https://user-images.githubusercontent.com/30118311/83684372-eb623680-a5b4-11ea-98cd-6b4586bbc7c0.png)

sasumner:
Another method is to do a _Find All in All Opened Documents_ with a _Find what_ of `\z` in _Regular expression_ search mode.  This will give you something like `...(4 hits in 4 files)` which would indicate that 4 total tabs are open.

ArkadiuszMichalski:
When I doing the translation I just noticed this information in the search result window, but with usability it has little to do, it will come across in a different direction. But I still think it could be added at least here:
![image](https://user-images.githubusercontent.com/2730894/83743544-261eaa00-a65b-11ea-8c26-5cb18ea28fa1.png)
In fact, thanks to the OK button we have free space for additional info. But from my perspective, I still need a real preview, not hidden somewhere, so look around for something else.

sasumner:
Adding it in the box that appears when one chooses _Window_ menu and selects _Windows..._ as shown immediately above seems a reasonable way to do it.  Reopening.

donho:
[...]

PR #9245, opened 2020-12-10
Add all tabs counter for windows dialog

X1ngyuL1u:
As requested by #8338,
Added a status bar to the windows dialog which shows the total number of tabs and will update along with other elements in the dialog. It will also resize well with the whole window. I had to move the "OK" button a little bit higher in order to add this status bar.


PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -865,6 +870,19 @@ void WindowsDlg::doClose()
...
+}
+
+//this function will be called everytime when close is performed
+//as well as each time refresh is performed to keep updated
+void WindowsDlg::doCount()
+{
+	TCHAR count[32];
+	wsprintf(count, TEXT("%d"), _idxMap.size());
+	generic_string msg = TEXT("");
+	msg += TEXT("Total: ");
```
sasumner, line 883: `Total` should be made translatable.

X1ngyuL1u: Updated the total. Now it should be translatable.

donho:
I don't like adding a status bar just for this info.
I think a static text on the bottom of Window dialog to display this info (`Total document number: 25`) will be enough.

X1ngyuL1u:
The original code uses CWinMgr which is hard to manipulate around, plus I think the added status bar can extend spaces for more information in the future if needed, e.g. total number of selected.

X1ngyuL1u:
> I don't like adding a status bar just for this info.
> I think a static text on the bottom of Window dialog to display this info (`Total document number: 25`) will be enough.

I tried to work around with the WinMgr the original code is using, however, I still cannot put he static text to the right place. So this time I tried to put the count to the Title, and now the title of the dialog would be "Windows -- Total tabs opened: XX".

donho (changes requested):
Use Window Caption (title bar) to display info instead of status bar:
"Window (total documents open: 46)"

donho (approved):

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -228,7 +229,7 @@ BEGIN_WINDOW_MAP(WindowsDlgMap)
 		BEGINCOLS(WRCT_REST,0,0)                       // Begin list control column
 			BEGINROWS(WRCT_REST,0,0)
 				RCREST(IDC_WINDOWS_LIST)
-				RCSPACE(20)
+				RCSPACE(15)
```
donho, line 232: Restore `RCSPACE(20)` which is not necessary.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -242,6 +243,7 @@ BEGIN_WINDOW_MAP(WindowsDlgMap)
 			RCTOFIT(IDC_WINDOWS_SORT)
 			RCREST(-1)
 			RCTOFIT(IDCANCEL)
+			RCSPACE(14)
```
donho, line 246: Remove `RCSPACE(14)` which is not necessary.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -664,7 +666,7 @@ void WindowsDlg::updateColumnNames()
 
 void WindowsDlg::onSize(UINT nType, int cx, int cy)
 {
-	MyBaseClass::onSize(nType, cx, cy);
+	MyBaseClass::onSize(nType, cx, cy);
```
donho, line 669: Remove all the space characters after `MyBaseClass::onSize(nType, cx, cy);`.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -865,6 +868,22 @@ void WindowsDlg::doClose()
 		}
 		ListView_SetItemCount(_hList, _idxMap.size());
 	}
+	doCount();
```
donho, line 871: Why do we need to count the document while closing the dialog?

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -865,6 +868,22 @@ void WindowsDlg::doClose()
...
+//as well as each time refresh is performed to keep updated
+void WindowsDlg::doCount()
+{
+	NativeLangSpeaker* pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
+	
+	TCHAR count[32];
+	wsprintf(count, TEXT("%d"), _idxMap.size());
+	
+	generic_string msg = TEXT("Windows -- ");
+	msg += pNativeSpeaker->getAttrNameStr(TEXT("Total tabs opened: "), WD_ROOTNODE, WD_TABTOTAL);
```
donho, line 884: Please replace `TEXT("Total tabs opened: ")` by `TEXT("Total documents opened: ")`.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlgRc.h
```diff
@@ -49,4 +49,3 @@
 	#define  IDM_WINDOW_WINDOWS   (IDR_WINDOWS_MENU + 1)
 	#define  IDM_WINDOW_MRU_FIRST (IDR_WINDOWS_MENU + 20)
 	#define  IDM_WINDOW_MRU_LIMIT (IDR_WINDOWS_MENU + 29)
-
```
donho, line 52: Restore this empty line.

donho (changes requested):
Please take consideration on the remarks.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
X1ngyuL1u: Restored

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
X1ngyuL1u: Removed

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
X1ngyuL1u: Removed

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
X1ngyuL1u: This is not closing the dialog, this is the button event which closes the chosen tab in the dialog, thus we need to update the count after we closing tabs

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
X1ngyuL1u: Replaced

PowerEditor/src/WinControls/WindowsDlg/WindowsDlgRc.h
X1ngyuL1u: Restored

donho (approved):

PR #9253, opened 2020-12-12
Update venetian.xml to 7.9.2

conky77:


donho:
@conky77 

donho:
@conky77 
With no reply from you, I'll merge your PR as its current state. 
If there's further modification for updating Venetian localization will wait for the next release (about 1 month later).

Issue #9240, opened 2020-12-09
Possible layout issue with the workspace toolbar


eljefe7000:
### Description of the Issue
Not sure if this is a bug or by design, but the workspace toolbar appears to be hard-configured with an RTL layout, which causes any icons you want to use to also be rendered in RTL format.  In the first screenshot, you can see an icon I created using the NPP "New File" icon as a base.  Note that the upper right corner is folded:
![image](https://user-images.githubusercontent.com/10099377/101658535-1cfc0480-3a13-11eb-943c-a237f7fc0eaa.png)

If I add the icon to the toolbar, it is rendered with the upper left corner folded:
![image](https://user-images.githubusercontent.com/10099377/101658992-91cf3e80-3a13-11eb-981f-4a0ae7a1fc39.png)

Again, this could be by design, but I wanted to point it out just in case.

### Steps to Reproduce the Issue
1. Create an icon whose left and right sides are non-symmetrical.
2. Add the icon to the workspace toolbar in code.
3. Run the code and open a workspace.

### Expected Behavior
The icon should be rendered exactly as it was designed.

### Actual Behavior
The icon is rendered in RTL format.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Build time : Dec  8 2020 - 22:21:25
Path : C:\Users\jdsmith\source\repos\notepad-plus-plus\PowerEditor\visual.net\x64\Unicode Debug\Notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.1198
Current ANSI codepage : 1252
Plugins : RestApiToText.dll

donho:
I see your point. However, all the icons on Folder as Workspace are symmetric, and I don't see the possibility to add new icons in the feature.

eljefe7000:
@donho , The reason I brought it up is that I was looking at another request - [9166: Change order of items in 'Folder as workspace'](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9166).  The request asks for menu items or buttons to move files up and down in the tree view.  I can make symmetric icons (up an down arrow) to satisfy the request.

donho:
The goal of using RTL layout is make "Locate current file" near the current file in edit zone. If this rule is respected, I have no problem to remove this property.

PR #9257, opened 2020-12-12
9240 workspace toolbar layout issue

eljefe7000:
This fix is for issue [9240: Possible layout issue with the workspace toolbar](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9240).  The toolbar layout style was set to WS_EX_LAYOUTRTL (Right to left mirroring), which meant the toolbar would render its button icons in RTL mirroring order also.  The result was that any non-symmetric icon would look "backwards" from the way it was designed.  Set the style to WS_EX_LEFT, which is the default.


donho:
What we need is have toolbar buttons on the right, so your PR doesn't reach my goal.

Issue #9166, opened 2020-11-16
Change order of items in 'Folder as workspace'


nelisse:
### Change order of items in 'Folder as workspace'
It would be useful to change the ordering of the items in 'Folder as workspace'.
E.g. add an up and down button, to move the selected item up/down in the list

### Steps to Reproduce the Issue
1. Currently not available

donho:
After checking the difficulty for implementing of this feature, this feature won't be implemented.

PR #9261, opened 2020-12-13
Issue 9166 change order of items in workspace folder

eljefe7000:
This pull request is for [Issue 9166: Change order of items in 'Folder as workspace'](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9166).  Added two new menu items, two new buttons and bitmap files, and code to move a file up or down in the workspace treeview.


donho:
@eljefe7000 
Your implementation works perfectly, however it's not complete - such feature is interesting ONLY IF the altered position can be crossed sessions. That means A LOT OF information will be memorized for this feature.

That's why I rejected this FR - change tree item's position means recording every item's position in tree, not only it's not trivial, but also it could bring some performance issue.

PR #9265, opened 2020-12-14
Fix scroll to last line problem after main window resize

### LINKED ISSUE #9123: Scrolling view to the end by -n or-p argument doesn't work properly for resize main window
### Description of the Issue
When we resize NPP main window then scrolling view to the end by `-n` or `-p` argument doesn't work properly.

1. Open NPP and resize main window (for example to 1024x703).
2. Close NPP.
3. Make test for `stylers.model.xml` file (it has 1407 lines and 170624 length)
```
notepad++ -n1407 stylers.model.xml
notepad++ -n1500 stylers.model.xml

notepad++ -p170624 stylers.model.xml
notepad++ -p180000 stylers.model.xml
```

We get cursor in right place, but view stick on 1401 line (when mian window size is 1024x703).

image

I don't know if it's a bug or intentional but if we want jump to last line/char (or we pass greater value than them) then NPP at least should force scrolling view to the end (as much as possible).

### Debug Information
Notepad++ v7.9.1 (32-bit)
Build time : Nov 2 2020 - 01:03:56
Path : D:_test\NPPClean\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.0
Current ANSI codepage : 1250
Plugins : none

sasumner:
Fix #9123

### DIFF
--- a/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScitillaComponent/ScintillaEditView.cpp
@@ -3347,6 +3347,7 @@ void ScintillaEditView::scrollPosToCenter(size_t pos)
 		middleLine = lastVisibleDocLine -  nbLine/2;
 	int nbLines2scroll =  line - middleLine;
 	scroll(0, nbLines2scroll);
+	execute(SCI_ENSUREVISIBLEENFORCEPOLICY, line);
 }
 
 void ScintillaEditView::hideLines()


sasumner:
@ArkadiuszMichalski Can you test this, to be sure?

ArkadiuszMichalski:
@sasumner Works fine.

sasumner:
> Works fine.

Okay then, reopening on your authority. :-)

PR #9267, opened 2020-12-14
Plugin admin to use UTF-8 transformation for author/description

### LINKED ISSUE #9266: PluginAdmin doesn't show Unicode texts
### Description of the Issue
UTF8 json input from plugin list is not displayed properly in PluginAdmin

Input from https://github.com/notepad-plus-plus/nppPluginList/blob/master/src/pl.x86.json#L833
			"author": "João Rosa",

is displayed as shown below.

### Expected Behavior
João  should be displayed as shown also here. 

### Actual Behavior
grafik

### Debug Information
N++ 7.9.1

See
https://stackoverflow.com/questions/583562/json-character-encoding-is-utf-8-well-supported-by-browsers-or-should-i-use-nu
on utf8 support for json.

chcg:
Fix #9266

- use UTF-8 input transformation for plugin author and description
- rreverted (update nlohmann json to v3.9.1 with performance and security improvements)

grafik

### DIFF
--- a/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
+++ b/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
@@ -656,10 +656,10 @@ bool loadFromJson(PluginViewList & pl, const json& j)
 			pi->_displayName = wmc.char2wchar(valStr.c_str(), CP_ACP);
 
 			valStr = i.at("author").get<std::string>();
-			pi->_author = wmc.char2wchar(valStr.c_str(), CP_ACP);
+			pi->_author = wmc.char2wchar(valStr.c_str(), CP_UTF8);
 
 			valStr = i.at("description").get<std::string>();
-			pi->_description = wmc.char2wchar(valStr.c_str(), CP_ACP);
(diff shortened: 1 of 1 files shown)


sasumner (commented):
Why only author and description changed to CP_UTF8?

chcg:
Just to be conservative in the change and it is unlikely that unicode is used for http links, foldernames, sha256 key and the version.

sasumner:
@chcg Just FYI because it appears you didn't know:
If you put `Fix #xxxx` as the first line of your PR comments, it will automagically link it back to issue "xxxx"

chcg:
@sasumner Thanks for the hint. Indeed I was not aware of that.

donho:
@chcg 
Why do we need to update json.hpp ?

chcg:
The update of nlohmann json to v3.9.1 is not necessary for the GUI change,

donho:
@chcg 
Then please remove the update part - I will take in charge of the update of "json.hpp".

donho (changes requested):
Restore "json.hpp" as it was.

donho:
Thank you @chcg !

chcg:
> 
> 
> Restore "json.hpp" as it was.

Reverted to old json.hpp version.

donho (approved):

PR #9268, opened 2020-12-15
Make find in finder result's text direction follows the search results' one

donho:


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/21e9239203fe734c24111edc29012deb364fdc5a

Issue #9254, opened 2020-12-12
Clipboard History text insertion does not respect current N++ tab line-ending type


sasumner:
### Description of the Issue
Data inserted into a document from the _Clipboard History_ panel does not have its line-ending types converted to the current documents line-ending type at the time of insertion, corrupting a document.

### Steps to Reproduce the Issue

1. Set the line-ending type of a N++ tab to `Unix (LF)` ; observe in status bar:
![image](https://user-images.githubusercontent.com/30118311/101984613-92a0e400-3c50-11eb-887e-b1aad57b6f11.png)
2. Create some data of multiple lines, which will appear as follows (when visible line-endings turned on):
![image](https://user-images.githubusercontent.com/30118311/101984560-376ef180-3c50-11eb-99fe-6f4eeb29d5b4.png)
3. Press Ctrl+a followed by Ctrl+c ; observe data has been remembered in the _Clipboard History_:
![image](https://user-images.githubusercontent.com/30118311/101984581-6b4a1700-3c50-11eb-81de-8fad140e9bd6.png)
4. Change line-ending type of the document to `Windows (CR LF)` ; observe in status bar:
![image](https://user-images.githubusercontent.com/30118311/101984646-bc5a0b00-3c50-11eb-87f0-72bcd4f31c65.png)
5. Observe data in tab now appears like this (when visible line-endings turned on):
![image](https://user-images.githubusercontent.com/30118311/101984676-deec2400-3c50-11eb-99f4-15d996419a67.png)
6. Move caret to start of line 2.
7. Double-click on the `abcdefghi` entry in the _Clipboard History_ panel to insert its data into the current document ; observe:
![image](https://user-images.githubusercontent.com/30118311/101984703-1f4ba200-3c51-11eb-87e2-828f57a8ab2c.png)

### Actual Behavior
After the insertion, the document becomes one of mixed-line-ending type, meaning that the document is corrupted.  This happens "silently" if the user is not operating with visible line-endings turned on.

### Expected Behavior
I expected, that just like any other paste, line-endings on the data are converted to the type of line-endings used by the document (and shown in the status bar) at the time of the paste.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Build time : Nov  2 2020 - 01:07:46
Path : C:\\.............\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
[...]

PR #9271, opened 2020-12-15
Fixed line endings when pasting from clipboard history

Joseph-Flanagan-Umich:
In response to #9254, fixes EOL when pasting from clipboard history


sasumner:
It hasn't been decided yet if that is the correct way to handle it.
I'm not saying that won't end up being the way, but closing for now...

Issue #9255, opened 2020-12-12
[FR] Clipboard History text data should widen/shrink as panel widens/shrinks


sasumner:
### Description of the Issue

The default size of the _Clipboard History_ panel is well-suited to the size of its text data entries, example:
![image](https://user-images.githubusercontent.com/30118311/101985120-02fd3480-3c54-11eb-9b01-55499cbacc03.png)

However, when the panel is resized with the mouse in order to widen it and see more of the text data content, the panel widens but the data does not:
![image](https://user-images.githubusercontent.com/30118311/101985148-30e27900-3c54-11eb-96fd-4fcaad0324d6.png)

### Steps to Reproduce the Issue

1. Show the _Clipboard History_ panel via _Edit_ menu's _Clipboard History_ entry
2. Copy some data to the clipboard
3. Observe the data in the  _Clipboard History_ panel
4. Use the mouse to change the horizontal size of the  _Clipboard History_ panel

### Actual Behavior
Observe that the amount of data shown does not grow/shrink with the sizing of the panel.

### Expected Behavior
I expected to be shown more of the text data content, because otherwise the screen area of the panel is empty and wasted, and I as a user perhaps can't see some important part of the data I need.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Build time : Nov  2 2020 - 01:07:46
Path : C:\\.............\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1809
OS Build : 17763.1577
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

PR #9274, opened 2020-12-16
Allowed clipboard history text to scale with panel

Joseph-Flanagan-Umich:
As requested by #9255 

Not sure if this is the best way. Also haven't been able to test on all resolutions. But this works for 1920 x 1080.

Let me know what I should do to further test this or if there is a better way.


sasumner:
I see "glitches" when using this PR and widening/reducing the Clipboard History panel:

Examples:

image

image

The window probably needs a horizontal scrollbar when the data is too wide to fit?

donho:
Please fix the glitch then redo another PR

PR #9292, opened 2020-12-21
Add option to disable minimal margin

### LINKED ISSUE #9291: [FR] Add an option to disable all print margins
As it currently stands, Notepad++ doesn't allow you to print a document with no margins whatsoever, i.e. with the text starting exactly in the upper left corner. This is a major issue for users of thermal line printers, especially ESC/P receipt printers, where the printable area is abysmally small (possibly narrower than 57mm).

The `printMarge` variable in Printer.cpp introduces a minimum margin of two columns and two rows, which, combined with the offset of the guillotine from the printing head, results in a lot of printable space going to waste. Moreover, it prevents the printer from operating in continuous feed mode, where each printout is seamlessly appended to the previous one.

This could be easily solved by introducing a toggle to disable the effects of `printMarge`:
Screenshot 2020-12-21 212424

Notepad++ is a very robust text editor that goes hand-in-hand with whitespace-sensitive applications the likes of POS printers. It would be a shame if such a simple limitation rendered it unsuitable for this job.

Note: the work needed to implement this is done. Just requesting your approval for the PR.

Lachcim:
This is the solution to the issue I mentioned in #9291. Summary of the changes:

* Made the value of `printMarge` in Printer.cpp toggleable
* Added a new checkbox to the print tab in Preferences
* Updated the preference dialog to link the checkbox to program logic
* Updated the config loader to load and save the preference

Fixes #9291


donho:
@Lachcim 
Sorry for the late review.
Could you restore the branch for this PR? Or you don't mind that I take the patch of this PR?

Lachcim:
Thanks for getting back to me @donho. Restoring the branch would take a lot of work, if you think you can manage without it, feel free to do so.

However, I'd like to draw your attention to the fact that this is likely a failed feature request that would probably fare better as a fork. It modifies the UI to implement a rather specialized feature which failed to gather support from the userbase. I would advise you to close this request as well as #9291, which perhaps I should have done myself.

donho:
@Lachcim 
> However, I'd like to draw your attention to the fact that this is likely a failed feature request that would probably fare better as a fork. It modifies the UI to implement a rather specialized feature which failed to gather support from the userbase. I would advise you to close this request as well as #9291, which perhaps I should have done myself.

OK, I will keep this PR opened, and see if there are the similar requests coming in the future. If there are the same sames requests, I will apply your patch here.

Thank you for your contribution.

Issue #6113, opened 2019-08-30
Tab bar's text does not respect system colors


Fierelier:
### Description of the Issue
The tab bar's text does not adapt the user's Windows theme colors. This is especially noticeable when using a dark theme.

### Steps to Reproduce the Issue
1. In Windows 7, right click the desktop and select "Personalize"
2. Apply the Windows Classic theme
3. Hit "Window Color" at the bottom
4. Apply a striking color to all text-related items and hit OK
5. Observe how none of the colors apply to the tab bar's text
The same can be achieved in Windows 8 and 10 with third party tools.

### Expected Behavior
The tab bar text adapts the "3D Objects" text-color.

### Actual Behavior
The tab bar text is always black.

### Debug Information
Notepad++ v7.7.1   (32-bit)
Build time : Jun 16 2019 - 21:14:50
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (32-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

### Screenshot
![image](https://user-images.githubusercontent.com/22013816/64030732-6fc7ea80-cb47-11e9-9ebd-981457e5804d.png)

antoniovazquezblanco:
### Description of the Issue

The issue also affects to me. Just wanted to clarify some details below.

 ### Steps to Reproduce the Issue

1. In Windows 10, right click the desktop and select "Personalize"
2. Go to the "Color" section.
3. Select "Dark" in windows theme and apply.
4. Observe how other windows apply the theme (Firefox, File explorer...) and Notepad++ stays in the light theme.

### Expected Behavior

The window frame and components respect windows theming.

### Actual Behavior

The tab bar text is always white.

### Debug Information

Notepad++ v7.7.1   (64-bit)
Build time : Jun 16 2019 - 21:24:47
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll 

### Screenshot

![imagen](https://user-images.githubusercontent.com/304193/66388433-693e5580-e9c6-11e9-8388-f651d06c5f92.png)

EthanR:
I would also appreciate this change, seems related to this other issue:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4463


[...]

PR #9298, opened 2020-12-23
Edit themes to use system scheme color for active tab text

Fierelier:
See issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6113#issuecomment-749948928


donho:
The system scheme could be changed by user. So it's no enough to change the color in theme.

Fierelier:
@donho You are wrong. The issue is that you defined the color in the themes. That means my change is superior, because it actually DOES respect the system scheme by REMOVING the color, while the current default does not.

PR #9301, opened 2020-12-23
Update venetian.xml

conky77:


donho:
Please check [v7.9.3 translation diff comparing with v7.9.2](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9455). If nothing is missed, please confirm it with me, so I will include this PR in the next release. Thank you!

conky77:
@donho What is the deadline for the updating?

donho:
@conky77 
There's no deadline. I'll release a new version soon. So if you cannot do it now ,I can always take the current PR as it is and you do the modification for the next release (v7.9.5).

PR #9328, opened 2021-01-01
Improve option for setting save dialog filter to All Types

### LINKED ISSUE #9325: Improve 'Set Save dialog file extension filter to .* ...' option functionality
### Description of the Issue

Some users have complained that the _Save As_ dialog functionality change, first initiated by @Uhf7, and then later compensated for (by me, with a PR to issue #8866) with a preference setting choice, doesn't provide the functionality they need.

Here's one complaint (source: https://community.notepad-plus-plus.org/post/60417 ):

> *_I deployed 7.9.1 to our users yesterday, but they are still unhappy, as the new checkbox only returns the old behavior partly.
> Without the checkbox enabled, any new file you “Save as” and an edited file you “save as” suggests you to save as .txt.
> With the old notepad++ versions it was for new and edited files .
> Now with the checkbox it is `*.*` for new files, but not for opened files.
> Example:
> We explicitly have the problem, that our users have to edit .STA files (which are just text files with STA as a file ending).
> So they open them via right click context menu, edit them and click “save as” and regardless that it’s a STA file, notepad++ 7.9.1 with the new checkbox enabled, always wants to save that file as .txt.
> That is a huge problem._*
> 

Here's another complaint (source: https://community.notepad-plus-plus.org/post/61186 ):

> *_What put it over the edge for me was not only the inconvenience of having to scroll up to and pick All file types, but that any time I had a non-.TXT extension (i.e., every file I ever edit!), when I went to “Save as” in another location, or, say as the next file in a sequence (i.e., just change number in file name), it would automatically add the .TXT extension even though A FILE EXTENSION WAS ALREADY PRESENT–and Notepad++ should have been smart enough to understand that. Nor can I begin to count up how many times I accidentally saved PHP source files as something.php.txt and then wondered why something didn’t work._*

Here's another perhaps more minor complaint (source: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9081#issuecomment-747741068 ):

> *_Once saved, it never easily saves without the extension (except after manually changing the file type to 'All types (`*.*`)' )
> Name your file 'new1' (without extension) -- That works fine.
> Try again "Save As..." (the combo box after 'Save as type' selects now: 'Normal text file (`*.txt`)')
> You cannot save it as 'new2' unless you change the type to `*.*`_*
> 

### Possible resolution

Currently the influence of the new preference setting is qualified by:

* the buffer that is being saved-as is "untitled", e.g. "new 2", and
* the language of the buffer is "Normal text"

What if these two qualifiers are removed?  
That way, with the preference setting enabled, any time the SaveAs dialog appears it will have "Save as type:" set to `All types (*.*)`.

Probably this eases the pain of the users complaining above.
But does it cause any new pain for other users?

### Debug Information
Comments above based upon functionality in Notepad++ 7.9.1

sasumner:
Fix #9325

Remove requirement for the `Set Save dialog file extension filter to `.*` instead of .txt for Normal Text` to affect only untitled (e.g. "new 4") and "text" (e.g. "Language" = "None (Normal text)" files.  With this PR, with the option enabled (ticked), the SaveAs dialog will unconditionally set the filter to `*.* (All types)` before showing the dialog to the user.

The only real change here is the logic controlling this (in NppIO.cpp) and the UI text of the option.

New UI text of the option:  `Set Save dialog file extension filter to *.*`

All other changes are changes to variable names to more accurately reflect the functionality.


mere-human (approved):
Tried this locally.
Now the dialog correctly shows "All types" for new file "Save" and opened file "Save as..." with non txt extension.
image
Good change!

bReathl3sS:
@sasumner This would be an awesome fix, thank you. I got really annoyed by this problem - thinking that 7.9.1's new tick box option would fix it just to find out that it won't.

sasumner:
@bReathl3sS 

>thinking that 7.9.1's new tick box option would fix it just to find out that it won't.

Part of the problem with such things is that there are a lot of differing opinions on how things should work. :-)

bReathl3sS:
Obviously, yeah.
As I’m not familiar with the Notepad++ release cycles: When will this fix be coming to the main installer? 7.9.3?

Gesendet von Mail für Windows 10

Von: sasumner
Gesendet: Mittwoch, 6. Januar 2021 13:19
An: notepad-plus-plus/notepad-plus-plus
Cc: bReathl3sS; Mention
Betreff: Re: [notepad-plus-plus/notepad-plus-plus] Improve option for settingsave dialog filter to All Types (#9328)

@bReathl3sS
thinking that 7.9.1's new tick box option would fix it just to find out that it won't.
Part of the problem with such things is that there are a lot of differing opinions on how things should work. :-)

sasumner:
>When will this fix be coming to the main installer? 7.9.3?

Unknown.
It is even unknown if this PR will be accepted by the Notepad++ author.
As of this writing, it is an "open" PR.

sasumner:
@bReathl3sS 

Now that this has been committed into the master branch, it will be in the next release, presumably 7.9.3.

sasumner:
@mere-human @donho 

If we move the "main" file _SaveAs_ dialog to the "newest style", perhaps the preference setting from this PR would be better served by being a checkbox right on the _SaveAs_ dialog?!  It would need to be "remembered", just like the setting currently is.

( This comment made due to changes to "session" _SaveAs_ dialog with PR #9286 )

mere-human:
> If we move the "main" file _SaveAs_ dialog to the "newest style", perhaps the preference setting from this PR would be better served by being a checkbox right on the _SaveAs_ dialog?! It would need to be "remembered", just like the setting currently is.

Do you want the "Save As" dialog to look like this?

image

As far as I understood, it makes sense to show such a checkbox only for Normal text (Language = None) files being saved.

Or we can still keep it in preferences if it is not changed often. This could save ordinary users from making hard decisions about the checkbox 🙂.

sasumner:
@mere-human It was just a thought...   We can keep the idea floating for now...

>it makes sense to show such a checkbox only for Normal text (Language = None) files being saved.

Hmm, perhaps my latest change (this accepted PR) went "too far", as I removed the Language=None qualifier.
We will see what the users think.
(This PR was kind of a hard one for me, because, as a user, I never ever had trouble with extensions at save-as time.  Other people, however, seemed to have no end of troubles)

donho:
@sasumner 
> If we move the "main" file SaveAs dialog to the "newest style", perhaps the preference setting from this PR would be better served by being a checkbox right on the SaveAs dialog?! It would need to be "remembered", just like the setting currently is.

That makes perfect sense! 👍 

@mere-human 
> Do you want the "Save As" dialog to look like this?

Awesome!

> As far as I understood, it makes sense to show such a checkbox only for Normal text (Language = None) files being saved.

Nope, in this merged PR, @sasumner has made change for all language:
image

So let's do it by following @sasumner PR's change.

sasumner:
>So let's do it by following @sasumner PR's change.

Can I suggest going with THIS PR's already committed change, as it is, for the next N++ release?
It will give users a chance to comment on the functionality change, without a location shift as well.

They are already somewhat used to this _Preference_ setting where it is, and changing its location _and_ its functionality at the same time may be "too much" change all at once.

If that is agreeable, after the next release I'll create an issue to move it into the _SaveAs_ dialog.

Also, this will give the "new dialog stuff" some time to settle out (before piling new functionality on).

VERY IMPORTANT ----> Also, if XP support is to be maintained (as discussed elsewhere) with a complicated scheme to keep the "old dialogs", then this setting can't really be moved into the new _SaveAs_ dialog, because how would users of the "old" see it?

donho:
@sasumner 
OK, let's do it this way.

> VERY IMPORTANT ----> Also, if XP support is to be maintained (as discussed elsewhere) with a complicated scheme to keep the "old dialogs", then this setting can't really be moved into the new SaveAs dialog, because how would users of the "old" see it?

If Notepad++ can continue being run under XP, with the minimal effort, then as I said in [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9378#issuecomment-761292695), we can keep it running under XP but stop fixing the bugs specific for XP.

sasumner:
> If Notepad++ can continue being run under XP, with the minimal effort, then as I said in [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9378#issuecomment-761292695), we can keep it running under XP but stop fixing the bugs specific for XP.

It wasn't my intent to start any kind of separate commentary about keeping XP or not in this thread, except to say how it relates to current thread:

* if keeping XP support : this thread's preference setting should stay where it is
* if dropping XP support : this thread's setting _could_ move into the _SaveAs_ dialog

donho:
@sasumner 
I'm checking if there is other simpler way to have new style dialog and keep Notepad++ running under XP.
If there's no other way, then XP will be dropped.
I just want to make sure.

mere-human:
I wanted to check out a solution with loading a necessary system .dll in runtime as it was suggested in [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9378#issuecomment-761527377) in another discussion.
I'll post any info in that thread as soon as I know it.
So many things to do!

sasumner:
>So many things to do!

That is _ALWAYS_ the case here.  Welcome to our world!  :-)

sasumner:
Continuing from some of the old discussion, beginning (specifically) here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9328#issuecomment-757176413
I have a new suggestion for what this might look like:

image

Hmm, I don't know that I like things in the "negative" sense.
Maybe it should be "Suggest a type"  ?

mere-human:
> Hmm, I don't know that I like things in the "negative" sense.
> Maybe it should be "Suggest a type" ?

Could be also something like:
"Ignore language"
"Ignore file type"
"Override suggested extension"

sasumner:
"Always prompt with \*.\* "

donho:
"Never suggest a type", "Ignore language", "Ignore file type" and "Override suggested extension" are more substile ways to name it. However, the most people work in IT/dev are not substile (including me), not mentioning that part of Notepad++ users are not native English user (me again).
So I would say `Always prompt with *.*` is a better way to word this item.

sasumner:
>substile

I think what is meant here is _subtle_

donho:
> I think what is meant here is subtle

👍 You see?!

PR #9339, opened 2021-01-02
add xml syntax checks according to #9330

### LINKED ISSUE #9330: Check XML validation in Notepad++ build system
An invalid xml could make disfunction in Notepad++ as problem in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9329 .

To avoid such problem, it's necessary to check the whole set of XML files in Notepad++'s AppVeyor settings to ensure the validity of these files.

chcg:
Fix #9330 

- Added xml syntax check via python script just for the win32 debug build because there already the functionlist check happens 
- Fixed an issue found for perl

Open points:
- Maybe the script should be placed under dir Test and not at toplevel dir.
- Some xml files/folders have 
`<?xml version="1.0" encoding="Windows-1252" ?>`
, why not using encoding UTF-8 for all xml files?
- For shortcuts.xml the header with the encoding is missing.


donho:
Thank you @chcg. 

> Maybe the script should be placed under dir Test and not at toplevel dir.

Yes, good point.

> Some xml files/folders have
<?xml version="1.0" encoding="Windows-1252" ?>
, why not using encoding UTF-8 for all xml files?

I'll check them.

> For shortcuts.xml the header with the encoding is missing.

Thank you for your heads up. I will add the header then.

donho:
@chcg 
> Maybe the script should be placed under dir Test and not at toplevel dir

In the new PR, please create `xmlValidator` under "Tests" and put the script into this created directory like:
C:\xxxxx\notepad-plus-plus\PowerEditor\Test\xmlValidator\validator_xml.py

PR #9351, opened 2021-01-04
Update german.xml v7.9.3

schnurlos:
Following https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9333

### DIFF
--- a/PowerEditor/installer/nativeLang/german.xml
+++ b/PowerEditor/installer/nativeLang/german.xml
@@ -2,7 +2,7 @@
 
 <!--
 	German localization for Notepad++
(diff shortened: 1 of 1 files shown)


donho:
Please check (v7.9.3 translation diff comparing with v7.9.2)[https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9455]. If nothing is missed, please confirm it with me, so I will include this PR in the next release. Thank you!

schnurlos:
@donho I can confirm the diff comparing, as I do always in that way. Also the https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9333 is done in my first fix here.

donho:
Thank you @schnurlos

PR #9381, opened 2021-01-10
Improve high-dpi display

### LINKED ISSUE #8115: External display, blurry display
External display, blurry display.
I have two monitors,The second monitor does not display properly

1.The first display shows normal

2.The second monitor does not display properly.The menu bar is obviously blurred, and there is a key dislocation in the upper right corner.Text editing area, I also feel blurred

sasumner:
Fix #8115 (and probably others)

Seems to improve things on high-dpi monitors? (without any otherwise obvious problems)
Here's my result (a definite improvement) -- perhaps the screenshot removes some of the crispness of reality, though!
On the left is N++ 7.9.2; on the right is the result from this PR -- font size settings are the same for each.
This is running on Win10.
Would appreciate if others would try on various OS and monitor situations!

image

Credit (or blame) to @xianyun666

### DIFF
--- a/PowerEditor/src/dpiAware.manifest
+++ b/PowerEditor/src/dpiAware.manifest
@@ -1,9 +1,9 @@
 <?xml version='1.0' encoding='UTF-8' standalone='yes'?>
 <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
   <asmv3:application xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
-    <asmv3:windowsSettings
-         xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
-      <dpiAware>true</dpiAware>
+    <asmv3:windowsSettings>
+      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
+      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitor</dpiAwareness>
     </asmv3:windowsSettings>
   </asmv3:application>
 </assembly>
--- a/PowerEditor/src/dpiAware.manifest
+++ b/PowerEditor/src/dpiAware.manifest
@@ -3,7 +3,7 @@
   <asmv3:application xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
     <asmv3:windowsSettings>
       <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
-      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitor</dpiAwareness>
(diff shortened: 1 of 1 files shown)


chcg:
See https://docs.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows

> Per-Monitor V1 DPI awareness mode (PMv1) was introduced with Windows 8.1. This DPI awareness mode is very limited and only offers the functionality listed below. It is recommended that desktop applications use Per-Monitor v2 awareness mode, supported on Windows 10 1703 or above.

So maybe "PerMonitorV2" might be even better on modern Win10 versions.

sasumner:
@chcg said:

>So maybe "PerMonitorV2" might be even better on modern Win10 versions.

Let's try it!  (it is in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9381/commits/80d47809ae3f842fc4f6a847878075586f1a1241)

xianyun666:
Have you tried the option of PerMonitorV2, there is a display problem, have you forgotten it?@sasumner

image

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8115

donho:
@sasumner 
Could you confirm that you can still reproduce the problem that @xianyun666 have mentioned?

sasumner:
>Could you confirm that you can still reproduce the problem that @xianyun666 have mentioned?

Interestingly, with `PerMonitorV2`, this time the bad effects were lessened, so I missed them.
But, yes, with V2 there are some bad UI elements...

So, I'll back this PR off to only use `PerMonitor` where I haven't yet seen any bad effects.
But it would be nice if many others could try it as well!

Here's a new screenshot, again with `PerMonitor` on my high-dpi screen, and again with 7.9.2 on the left, and the `PerMonitor` version of this PR on the right:

image

sasumner:
Note:  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9381/commits/0cec1bda1dc2c79c348ca0360a517ce7861f4b9c removes `PerMonitorV2` and restores `PerMonitor`

Sorry for the flip-flopping and possible confusion it generates.

donho:
> But, yes, with V2 there are some bad UI elements...

Could you provide the screen shot please?

sasumner:
>Could you provide the screen shot please?  (using `PerMonitorV2`)

Ok, so to make screenshots I put `PerMonitorV2` back in place.
And I have no explanation for this, but it looks A LOT worse than when I looked at it yesterday or even earlier today.
But here are some of the bad UI examples:

_Find_ when opened onto a high-dpi monitor:
image

_Find_ when opened onto a non-high-dpi monitor:
image

_Shortcut Mapper_ when opened onto a high-dpi monitor:
image

_Shortcut Mapper_ when opened onto a non-high-dpi monitor:
image

donho:
Thank you very much @sasumner for your screenshot.
I don't know such display bug from Windows or Notepad++ - or it's due to part of your system (your OS) is not updated.
In any case, let's do it without V2.

donho:
In fact, with V2 render the better standard GUI. OTOH, all the non standard GUI part are f*cked up.

PR #9387, opened 2021-01-13
Restore split-lines split-on-vertical-edge functionality

### LINKED ISSUE #8262: [v.7.8.6 regression] Vertical Edge Settings no longer works with Split Lines (Ctrl-I)
### Description of the Issue

v.7.8.6 introduced the ability to have multiple Vertical Edges, but it breaks an existing
feature, i.e. Split Lines (Ctrl-I) no longer works with single Vertical Edge.

1. Settings/Preferences/Editing/Vertical Edge Settings: enter a single number, e.g. 80, into the text box.
2. Leave "Background mode" unchecked.
3. Enter text with more than 80 characters in a single line
4. press Ctrl-I (Split Lines)

### Expected Behavior

The single line is expected to split into multiple lines, where each line is less than 80 characters long.

### Actual Behavior

Nothing happened after pressing Ctrl-I.

### Debug Information

Notepad++ v7.8.6   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.778

image

sasumner:
Fix #8262

Implements a fix whereby the RIGHTMOST user specified value in the vertical edge box is the column to split by, e.g. 60 in the following example:

image

If the contents of the box were instead `60 40`, the split-by column would be 40, as that is the RIGHTMOST value.

-----------------------------------------------------------------

Also, fixes the problem pointed out [HERE](https://community.notepad-plus-plus.org/post/61615), specifically:

_"If the font size of the line numbers margin is set to a different value than the font size for documents, the “word-wrap at single-edge-line” code calculates a wrong wrapping position. This position has to be provided in pixels. To calculate this pixel value, a Scintilla function has to be called that calculates the width in pixels of a provided string from a provided text style. It seems like Notepad++'s internal code erroneously takes the text style of the line numbers margin font when calling this Scintilla function."_

by changing usage of `STYLE_LINENUMBER` to `STYLE_DEFAULT`.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1598,14 +1598,22 @@ void Notepad_plus::command(int id)
 				auto caretPos = _pEditView->execute(SCI_GETLINEENDPOSITION, lineRange.second);
 				_pEditView->execute(SCI_SETSELECTION, caretPos, anchorPos);
 				_pEditView->execute(SCI_TARGETFROMSELECTION);
-				if (_pEditView->execute(SCI_GETEDGEMODE) == EDGE_NONE)
+				int edgeMode = static_cast<int>(_pEditView->execute(SCI_GETEDGEMODE));
+				if (edgeMode == EDGE_NONE)
 				{
-					_pEditView->execute(SCI_LINESSPLIT);
+					_pEditView->execute(SCI_LINESSPLIT, 0);
 				}
 				else
 				{
-					auto textWidth = _pEditView->execute(SCI_TEXTWIDTH, STYLE_LINENUMBER, reinterpret_cast<LPARAM>("P"));
-					auto edgeCol = _pEditView->execute(SCI_GETEDGECOLUMN);
+					auto textWidth = _pEditView->execute(SCI_TEXTWIDTH, STYLE_DEFAULT, reinterpret_cast<LPARAM>("P"));
+					auto edgeCol = _pEditView->execute(SCI_GETEDGECOLUMN); // will work for edgeMode == EDGE_BACKGROUND
+					if (edgeMode == EDGE_MULTILINE)
+					{
+						NppParameters& nppParam = NppParameters::getInstance();
+						ScintillaViewParams& svp = const_cast<ScintillaViewParams&>(nppParam.getSVP());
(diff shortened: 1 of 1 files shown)


donho:
@sasumner 
Why not LEFTMOST?

sasumner:
>Why not LEFTMOST?

Why not, indeed?  :-)
It certainly _could_ be the left-most.

However, @pryrt pointed out, [HERE](https://community.notepad-plus-plus.org/post/61569) that:

"I am not sure it’s ever overly useful to wrap at anything but the rightmost (edge), because otherwise, why draw those other (edges)?"

So I proceeded based upon that "nudge".

When stating that, @pryrt is operating on the assumption that users will enter multiple edges into the input box from lowest to highest, e.g. `40 60 80`.  But this is simply perhaps a natural or "feel good" way of doing it for the user, as entering `80 40 60` would give you the very same edges.

So @donho , if you will accept the PR if I change it to the "leftmost", that is not a problem at all and I will do it.

But, if I am learning how to correctly read your comments, I think you may just be asking the question, and not really objecting to "rightmost" -- am I correct?  :-)

The important part is to give the users back their desired split-on-edge functionality.
Please advise me on how to proceed.

pryrt:
> When stating that, @pryrt is operating on the assumption that users will enter multiple edges into the input box from lowest to highest, e.g. 40 60 80. But this is simply perhaps a natural or "feel good" way of doing it for the user, as entering 80 40 60 would give you the very same edges.

When I said "RIGHTMOST" in the forum, I was actually intending to specify in the "largest column number", which in European-based languages (left-to-right direction) is also the "rightmost" column number.   I probably did have the tacit assumption that people would define the margin columns from smallest to largest in the text box when answering it, but order inside the entry wasn't actively on my mind -- I was thinking about column number.  And my comment of "I am not sure it’s ever overly useful to wrap at anything but the rightmost (edge)" only makes sense if the rightmost edge is the rightmost in terms of column number....  Because if you have `60 40` and wrap at column 40 but have another edge drawn at column 60, it makes no sense, because no text will ever hit the edge marker at column 60.  But that's just making no sense to me.  As long as the documentation that you write up for this feature makes it clear that writing `60 40` will have the wrap at column 40 but draw both edges, that's all that really matters.  (And if someone else _does_ have a reason for wrapping before the final edge, this implementation gives them a way to have it the way they want, despite my incredulity at the need for it.)

PR #9396, opened 2021-01-15
Add ability to delete from Find combobox history

### LINKED ISSUE #9366: [Feature Request] Clear Directories In Find In Files
I'd likr a way to clear the directory listing in Find In Files.  I have a few I no longer need to search in (one directory has been deleted), and I'd like to see a way to delete individual directories in the pull down menu. Don't need a clear all, as some I still use.

sasumner:
Fix #9366 

When any of the _Find_ window comboboxes are dropped and an entry is highlighted, pressing the _Delete_ key will remove the highlighted entry:

* _Find what_ combobox
* _Replace with_ combobox
* _Directory_ combobox
* _Filters_ combobox

No method is provided for clearing all entries at once, but as the "depth" of the comboboxes is limited to a small amount, clearing one entry at a time (when one wants to clear all) is not overly burdensome.

### DIFF
--- a/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp
@@ -865,12 +865,16 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			COMBOBOXINFO cbinfo = { sizeof(COMBOBOXINFO) };
 			GetComboBoxInfo(hFindCombo, &cbinfo);
 			originalComboEditProc = SetWindowLongPtr(cbinfo.hwndItem, GWLP_WNDPROC, reinterpret_cast<LONG_PTR>(comboEditProc));
+			SetWindowLongPtr(cbinfo.hwndItem, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(cbinfo.hwndCombo));
 			GetComboBoxInfo(hReplaceCombo, &cbinfo);
 			SetWindowLongPtr(cbinfo.hwndItem, GWLP_WNDPROC, reinterpret_cast<LONG_PTR>(comboEditProc));
+			SetWindowLongPtr(cbinfo.hwndItem, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(cbinfo.hwndCombo));
 			GetComboBoxInfo(hFiltersCombo, &cbinfo);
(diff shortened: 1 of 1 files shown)


sasumner:
Closing temporarily as a defect was noted:  If this is used to remove _all_ entries of a combobox, what is remembered after a N++ restart is the original contents, NOT "nothing".  Will fix this and reopen...

sasumner:
@chcg calls it a "feature"; I call it an "enhancement".  :-)

sasumner:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9396/commits/91648c40b586b00ff1e749800eacf37bb5142dbd clears up the problem of an empty combobox not being remembered.  Also cleaned up some space versus tab issues in the source close to this change.

mere-human (commented):
This works pretty well already!
If the text in the edit box is not the same as the current selection that we want to delete, then it is changed.
Can we avoid that?

Steps:
1. Expand combo box:
image
2. Press the "Delete" key button with the same selection.
3. Result: Edit box text changes
image

sasumner:
>If the text in the edit box is not the same as the current selection that we want to delete, then it is changed.
Can we avoid that?

I didn't (and still don't) think that is an important enough thing to bother about.

mere-human:
> > If the text in the edit box is not the same as the current selection that we want to delete, then it is changed.
> > Can we avoid that?
> 
> I didn't (and still don't) think that is an important enough thing to bother about.

Ok. It is fine.

mere-human (approved):

donho:
I doubt how many users will discover this neat feature, even it'll be documented in https://npp-user-manual.org/

mere-human:
> I doubt how many users will discover this neat feature, even it'll be documented in https://npp-user-manual.org/

It is present in some form in other applications. 
For example, a user can delete search suggestions in the browser address bar dropdown via _Shift + Del_:
https://www.gearpatrol.com/tech/a429171/how-to-forget-a-wrong-url-in-chrome/

sasumner:
>I doubt how many users will discover this neat feature, even it'll be documented...

Well, there's no harm in having undiscovered features -- when users request them, we can tell them they already exist!

BTW, I do remember one other request (aside from the one in the linked issue) in the past for this.  But I can't find a reference to it now.  So, at least two users can be made happy by this change.

>other applications...delete search suggestions in the browser address bar dropdown via Shift + Del...

I didn't consider that there might be a "standard" way of performing the action.
To me the _Delete_ key alone was intuitive.
Also, I considered mouse right-click, with popup menu, for it, but that means localization/translations are involved -- and that just feels too heavy and complicated for this.

PR #9403, opened 2021-01-17
Allow forward slash file name in CustomFileDialog

### LINKED ISSUE #9374: Save/Open (including Save/Load Session) file dialogs don't accept forward slashes in 'File name:' box
### Description of the Issue
Post integration of PR #9286, I noticed that the SaveAs dialog used when saving a session does not support forward slashing in the pathname.  It would be nice if such slashing were allowed.

1. _File_ > _Save Session...
2. Attempt to specify a (valid) full path with forward slashes in its name into the _SaveAs_ dialog's _File name_ box:
image
3. Press the _Save_ button and observe the error:
image
4. Change the slashes to backward ones:
image
5. Observe session file saves without error into the destination pathname

### Expected Behavior
To be able to use either forward ( `/` ) or backward ( `\` ) slashes in pathnames in the SaveAs (and Open) dialogs.

### Actual Behavior
Unable to use forward slashes.

### Debug Information
Notepad++ source after this commit: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b9f498f0412bc1806ee238627b6099d26e76e9cf

mere-human:
Allow forward-slash file name in CustomFileDialog.
Add necessary functionality to be able to use CustomFileDialog widespread.
See commit messages for details.

Fix #9374


PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -130,22 +359,29 @@ class CustomFileDialog::Impl
...
 	bool show()
 	{
-		HRESULT hr = _dialog->Show(_hwndOwner);
-		return SUCCEEDED(hr);
+		bool okPressed = false;
+		HRESULT hr = FileDialogEventHandler::createInstance(IID_PPV_ARGS(&_events));
+		if (SUCCEEDED(hr))
+		{
+			DWORD dwCookie;
+			hr = _dialog->Advise(_events, &dwCookie);
```
mere-human, line 374: Should we check the preference that enables the Unix-style file path?

mere-human:
Also, it works for some file picker dialogs (e.g. _File > Open Folder as Workspace_). 
Should I disable this feature for the file picker?

For the file picker, the input with the "Enter" key button doesn't work (yet). It works only when"Select Folder" is pressed. 
Should I fix it also?

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
sasumner: >Should we check the preference that enables the Unix-style file path?

I don't think so, because if this all works out, that pref setting will be going away.  :-)

donho:
Firstly, as I mentioned [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9374#issuecomment-763778321), the UNIX style path should work for all file dialogs (save/open & save/load session & open folder, etc...). Obviously, in this PR, it's done only for save session dialog (not even for load session).

Secondly, Unix style **folder** path doesn't work with save session dialog, whereas Windows style folder path does. To reproduce: you can give `c:\tmp` to browse in "tmp" directory by clicking button "Save", but if you provide `c:/tmp`, the error message displayed.

donho (changes requested):
1. remove unnecessary define.
2. follow Windows style folder path browsing behaviour
3. allow UNIX style path also in Save/Open file dialog

mere-human:
In the current state, the forward-slashes should be acceptable by file dialogs.
Also, FileDialog is completely replaced by CustomFileDialog.

You can try reviewing per commit instead of per file.
Let me know if I can split this PR into smaller parts to make the review process easier.

mere-human:
**Test scenarios:**
• Forward slashes in file or directory name (relative, absolute)
• Confirm the dialog with a button or with an Enter key
• Make sure default directory preference is respected
• Session load/save
• Workspace (Project Panel) save/load
• Project Panel > Add files... (multi-select)
• UDL import/export
• Run > Run.. > [...] (.exe, .com, .bat | all)
• Tools > MD5 > Genererate from files
• Settings > Import plugins (.dll, multi-select)
• Settings > Import theme styles (XML, multi-select)
• File > Open/Save/Save As (filters, read-only, initial file name, multi-select)
• Save As: empty file should be with .txt extension
• Save As: change extension (.java, .v, etc) - name should be updated
• Save As: type name without extension - new file with extension according to type
• Save As: type a name with extension - it should be preserved regardless of type

mere-human:
I had to rebase it against the latest master.

donho (approved):

donho (changes requested):
1. PowerEditor/bin/NotepadPP.exe_dynamic & scintilla/bin/libscintilla.a shouldn't be commited. Remove them.
2. Please rebase with last mast to solve the conflict.
3. Fix a crash (I'll leave the instructions to reproduce it).

donho:
I have had a crash while save file. Here are the instruction to reproduce it:
1. Open "new 1" & "new 2", write somethings in both files, quit Notepad++
2. Reopen Notepad++, "new 1" & "new 2" are opened with the state "dirty"
3. "Ctrl-S" to save "new 1", then click "Cancel".
4. Switch to "new 2", "Ctrl-S" to save "new 2"
5. Notepad++ will hang few seconds, the crash (disappears)

Note I have checked "Set Save dialog extension filter to *.*" in Preferences dialog. Not sure if it matters.

donho:
Problem found. "Set Save dialog extension filter to ." does matters.
In the following line:
```
const generic_string ext = get1stExt(_filterSpec[_fileTypeIndex].ext);
```
The length of `_filterSpec` is 1, but `_fileTypeIndex` is 1 instead of 0.

mere-human:
By the way, what purpose does the MinGW build serve? This is not stated anywhere, I am just curious.

sasumner:
>By the way, what purpose does the MinGW build serve? This is not stated anywhere, I am just curious.

I don't really know, but I've used such "dual built" systems in the past as a "check" on cpmpilers that issue questionable warnings, or not enough warnings.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -17,24 +17,39 @@
...
+const GUID ComTraits<T>::uid = __uuidof(T);
+
 // Smart pointer alias for COM objects that makes reference counting easier.
 template<class T>
-using com_ptr = _com_ptr_t<_com_IIID<T, &__uuidof(T)>>;
+using com_ptr = _com_ptr_t<_com_IIID<T, &ComTraits<T>::uid>>;
 
 // _com_ptr_t needs this function.
 // Define it here to avoid linking with "comsupp.lib".
+// MinGW defines its own version of this function.
```
sasumner, line 44: Is there a downside to linking with comsupp.lib ?

sasumner:
This entry for _File name_ won't open the _functionList_ folder when Enter/Open is pressed, but `.\functionList` will.  Both should be correct and open the folder:

image

As it is, when Enter/Open is pressed, all that happens is that the _File name_ box is cleared.

sasumner:
Not a problem relating to this PR, but, if an invalid path is in the _Directory_ field on the _Find in Files_ dialog when the `...` button is pressed, no dialog opens.

Would it be better to open the folder-browser dialog to the folder of the path most-valid from the _Directory_ field?

Example:
_Directory_ box contains `c:\aaaa\bbbb\cccc\dddd\eeee`
but `c:\aaaa\bbbb\cccc\dddd\` is not a valid path, but `c:\aaaa\bbbb\cccc\` is valid
In this case should the folder-brower dialog open to present the user a view of folder `cccc` ?

sasumner:
Removal of this setting (no longer necessary) should also be part of this PR ... ? :

image

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
mere-human: No, I am not aware of any.
It just seemed like overkill to include a .lib just for a single simple function.
Do you think we can add it and let the optimizer do the thing?

mere-human:
> Would it be better to open the folder-browser dialog to the folder of the path most-valid from the _Directory_ field?

To me, it is more like a nice to have feature. I'd better implement it separately. This PR seems pretty big already.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
sasumner: To me a dozen characters added to the link specification beats a bunch of source lines of conditional code.
Not sure what @donho prefers.

mere-human:
> Removal of this setting (no longer necessary)

Done.

> This entry for _File name_ won't open the _functionList_ folder when Enter/Open is pressed, but `.\functionList` will. Both should be correct and open the folder

Done.

donho (approved):

donho:
Sorry for the long review process. Since the PR is quite huge, and it may bring some regression, I will check the binary (tests) without really checking the source code for the first step. So the PR will be modified with several iteration.

donho (changes requested):
1. Fix the related path not working for Unix style: for browsing if the directory is on `c:\tmp` and I want to change to `c:\toto`, Windows style `..\toto` works fine, but `../toto` doesn't work. (the same problem of `.\functionList` that @sasumner mentioned) 
2. Remove all changes for localization except `english.xml` & `english_customizable.xml`. The reason is that whole bunch files make review more complicated. The translators (or I) will take in charge of it. Keep in mind that your PR has more chance to be accepted if it is smaller, cleaner and more compact (ie. easier to be reviewed).
3. Link with comsupp.lib and remove the source code come from comsupp.lib.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
donho: @mere-human 
How can it "overkill" to link a .lib? 
@sasumner is right. Let's link with comsupp.lib - I prefer to include comsupp.lib in the project file than including a bunch of source lines which is not belong to us (and I don't think we have the right - since it's not under GPL).

PowerEditor/visual.net/notepadPlus.vcxproj
```diff
@@ -301,7 +301,6 @@ copy ..\src\contextMenu.xml ..\bin64\contextMenu.xml
     <ClCompile Include="..\src\ScitillaComponent\DocTabView.cpp" />
     <ClCompile Include="..\src\WinControls\DocumentMap\documentMap.cpp" />
     <ClCompile Include="..\src\EncodingMapper.cpp" />
-    <ClCompile Include="..\src\WinControls\OpenSaveFileDialog\FileDialog.cpp" />
```
donho, line 304: Shouldn't we add `..\src\WinControls\OpenSaveFileDialog\CustomFileDialog.cpp`?

PowerEditor/visual.net/notepadPlus.vcxproj
```diff
@@ -594,7 +593,6 @@ copy ..\src\contextMenu.xml ..\bin64\contextMenu.xml
     <ClInclude Include="..\src\ScitillaComponent\DocTabView.h" />
     <ClInclude Include="..\src\WinControls\DocumentMap\documentMap.h" />
     <ClInclude Include="..\src\EncodingMapper.h" />
-    <ClInclude Include="..\src\WinControls\OpenSaveFileDialog\FileDialog.h" />
```
donho, line 597: Shouldn't we add `..\src\WinControls\OpenSaveFileDialog\CustomFileDialog.h`?

PowerEditor/visual.net/notepadPlus.vcxproj
mere-human: It is already there, see https://github.com/notepad-plus-plus/notepad-plus-plus/blob/611790310c89e1bf1357b1be7a9fdfd3abbebd68/PowerEditor/visual.net/notepadPlus.vcxproj#L295

PowerEditor/visual.net/notepadPlus.vcxproj
mere-human: It is already there, see https://github.com/notepad-plus-plus/notepad-plus-plus/blob/611790310c89e1bf1357b1be7a9fdfd3abbebd68/PowerEditor/visual.net/notepadPlus.vcxproj#L579

mere-human:
> 1. Fix the related path not working for Unix style: for browsing if the directory is on `c:\tmp` and I want to change to `c:\toto`, Windows style `..\toto` works fine, but `../toto` doesn't work. (the same problem of `.\functionList` that @sasumner mentioned)

Have you tried the latest version of the PR?
This should have fixed the relative path issue https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9403/commits/20a541c72f527fbac12f360d3942867a85051eea
I've just checked locally (release x86) and it works as expected:

https://user-images.githubusercontent.com/9664141/105688855-c4b98d00-5f02-11eb-8893-a4e9ed87d0e3.mp4


Also, what build did you try (release/debug/x86/x64)?

donho:
> Have you tried the latest version of the PR?
This should have fixed the relative path issue 20a541c
I've just checked locally (release x86) and it works as expected

Yes, indeed - it's fixed in the last state of PR. Thank you!

mere-human:
> * Remove all changes for localization except `english.xml` & `english_customizable.xml`. The reason is that whole bunch files make review more complicated. The translators (or I) will take in charge of it. Keep in mind that your PR has more chance to be accepted if it is smaller, cleaner and more compact (ie. easier to be reviewed).

Right, sorry. Fixed.
This PR is big. Maybe, I should have done the preference removal in a separate PR.

> * Link with comsupp.lib and remove the source code come from comsupp.lib.

Done. Probably not needed at all since my build succeeded without it. Maybe, it was necessary when the minimal Windows version was XP.

Suesskirsche:
Hi,
I'm not a programmer but just a user with eyes spotting typos ;-)... 
So I hope I don't confuse. 

Shouldn't the 2nd line contain 
   "...\src\ScintillaComponent\..." 
instead of 
   "...\src\ScitillaComponent\..." (missing "n")
?
Regards 
Michael 


Am 25. Januar 2021 04:24:05 MEZ schrieb Don HO <notifications@github.com>:

sasumner:
@Suesskirsche said

>Shouldn't the 2nd line contain
   "...\src\ScintillaComponent\..."
instead of
   "...\src\ScitillaComponent\..." (missing "n")
?

Nope

mere-human:
> Hi, I'm not a programmer but just a user with eyes spotting typos ;-)... So I hope I don't confuse. Shouldn't the 2nd line contain "...\src\ScintillaComponent\..." instead of "...\src\ScitillaComponent\..." (missing "n") ? 

Looks like a typo. Have to be fixed someday.

sasumner:
>Looks like a typo. Have to be fixed someday.

It's just a folder name.  
Unfortunately it was misspelled when it was first created, but does no harm.
It could be renamed, I suppose, if someone wants to go to the effort.
I believe github can be told not to lose the history of the contained files if it is renamed someday.

donho (approved):

donho:
@mere-human 

Well done! 

You can create a new PR to remove all Item where id="6430" if you want.
Otherwise I can do it.

Thank you for the PR which makes Notepad++ look younger!

PR #9405, opened 2021-01-18
Update polish.xml

klewi1983:
Fixed a typo


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9457

PR #9411, opened 2021-01-19
Add def of _WIN32_WINNT

donho:


mere-human:
Take a look at my comment:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/051b17ca44ca7402ad8a6f6a7cbc0be0c0fe1184#r46119100
You should change this file instead `PowerEditor\gcc\include\various.h`

chcg:
@donho 
Comment from @mere-human to change PowerEditor\gcc\include\various.h is correct. It would be also possible to change it in the makefile itself, but as it already exists in various.h it is easier to adapt it directly there.

Further infos also added to the checkin comment:

See e.g. https://mariusbancila.ro/blog/2018/12/07/troubles-with-windows-target-versions/.

So both defines _WIN32_WINDOWS and _WIN32_WINNT should be changed according to the VS project changes and be updated to 0x0600.

Most likely _WIN32_WINDOWS is obsolete in the meantime as it seems it was used for Win95/98/ME which is even more outdated than WinXP:
https://github.com/Alexpux/mingw-w64/blob/d0d7f784833bbb0b2d279310ddc6afb52fe47a46/mingw-w64-headers/include/winresrc.h
https://github.com/Alexpux/mingw-w64/blob/d0d7f784833bbb0b2d279310ddc6afb52fe47a46/mingw-w64-headers/include/w32api.h
, but it should do no harm to have it still available.

PR #9378, opened 2021-01-09
Use new style dialog to browse a folder

mere-human:
Add new methods to CustomFileDialog:
- setTitle() sets the dialog title
- setFolder() sets an initial directory
- pickFolder() shows a file open dialog to select a folder

Use CustomFileDialog in folderBrowser().

Affected areas:
- Search > Find in Files
- File > Open Folder as Workspace
- Preferences > Default Directory
- Preferences > Backup > Custom Backup
- Preferences > Cloud & Link

Fix #8513

PR #9412, opened 2021-01-19
Fix startup error on Windows XP caused by SHCreateItemFromParsingName

mere-human:
Load shell32.dll and resolve the necessary symbol at runtime.
This makes it possible to run the app on Windows XP.

Address discussion comments in issue #9378

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -38,6 +38,32 @@
 #include "CustomFileDialog.h"
 #include "Parameters.h"
 
+namespace // anonymous
+{
+	IShellItem* createShellItem(const TCHAR* name)
+	{
+		using SHCreateItemFromParsingName_t =
+			HRESULT(WINAPI *)(PCWSTR pszPath, IBindCtx *pbc, REFIID riid, void **ppv);
+		static SHCreateItemFromParsingName_t pCreateShellItem = nullptr;
+
+		// Load a symbol at runtime. Otherwise, the application won't start on Windows XP.
+		if (!pCreateShellItem)
+		{
+			HMODULE hShell32 = ::GetModuleHandle(_T("shell32.dll"));
+			pCreateShellItem = (SHCreateItemFromParsingName_t)GetProcAddress(hShell32, "SHCreateItemFromParsingName");
+		}
+
+		if (pCreateShellItem)
+		{
+			IShellItem* psi = nullptr;
+			pCreateShellItem(name, nullptr, IID_PPV_ARGS(&psi));
+			return psi;
+		}
+		return nullptr;
+	}
+
+} // anonymous namespace
+
 // Private impelemnation to avoid pollution with includes and defines in header.
 class CustomFileDialog::Impl
 {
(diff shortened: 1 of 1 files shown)


Uhf7:
I don't think it is necessary to go this way, if we drop XP support anyway. 

But another question arises here. If we drop XP support for one of the most interesting base functions of the software (saving files), than perhaps it would be good not to start Notepad++ under XP at all.

ArkadiuszMichalski:
If some critical things no longer work under XP, an alert should pop up as to which latest version is running under this system and the program should be closed (not to cause damage).

mere-human:
> If we drop XP support for one of the most interesting base functions of the software (saving files), then perhaps it would be good not to start Notepad++ under XP at all.

For now, only some places are affected like the dialog from _File > Save Session_.
But we can come up with a workaround. For example, we can show `FileDialog` that uses `GetSaveFileName` on Windows XP.

Uhf7:
> an alert should pop up as to inform which latest version working under this system

This would be ideal, but the problem here is, that in order to show such a message box under XP, we have to assure, that no post-XP function is used without a `LoadLibrary` wrap around. As soon as one post-XP function is used in the natural way, then Notepad++ would stop working before showing such a message box. 

To load all post-XP functions at run time with `LoadLibrary`, seems not practicable to me. In so far, the system error `Notepad++ cannot load SHCr...` message box was not too bad at all. It assures that the user cannot start Notepad++ accidentally.

> For now, only some places are affected like the dialog from File > Save Session.
But we can come up with a workaround. For example, we can show `FileDialog` that uses `GetSaveFileName` on Windows XP.

I think, all of the load/save file dialogs will follow sooner or later, and over the long run, it will not be easy to maintain and test 3 versions of each dialog every time. Since the new style file dialogs have some advantages, it would be better and more future-save to drop XP like suggested here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9378#issuecomment-762555009 and focus on a single implementation of each load/save dialog.

ArkadiuszMichalski:
I open separate bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9423 because support for older systems appears in other bugs and it is worth having a clear position on this matter.

sasumner:
>it will not be easy to maintain and test 3 versions of each dialog every time

+1

donho:
It took time to download & install XP SP3 and configure VirtualBox to make file transfer working from host to client.
Anyway, I finally did some tests with this PR under XP.

The test code contains :
**1)** this PR
**2)** revert of https://github.com/notepad-plus-plus/notepad-plus-plus/commit/45912a3321877bc159ebffebf788dcbcc48c34ea
**3)** https://github.com/notepad-plus-plus/notepad-plus-plus/commit/051b17ca44ca7402ad8a6f6a7cbc0be0c0fe1184  

But since **2)** is revert, due to symbol `COPY_FILE_NO_BUFFERING` in **3)**, I have had to add *A* or *B* into `NppIO.cpp`  :

*A*
```
#if defined(_WIN32_WINNT) && (!defined(_WIN32_WINNT_VISTA) || (_WIN32_WINNT < _WIN32_WINNT_VISTA))
#undef _WIN32_WINNT
#define _WIN32_WINNT _WIN32_WINNT_VISTA
#endif
```

*B*
```
#ifndef COPY_FILE_NO_BUFFERING
#define COPY_FILE_NO_BUFFERING                0x00001000
#endif
```

The result with *B* works fine, Notepad++ is launched without any problem, whereas the result with *A* crashed.

Conclusion: If both *A* & *B* worked, I would include this PR and keep making Notepad++ running under XP, because this compact PR (little effort) can make these prehistoric people happy. 
The failure (with *A*) reminds us that keeping supporting very old OS will decrease our code quality.
So XP will be dropped in the next release. Thank you @mere-human for your effort (still).

sasumner:
>So XP will be dropped in the next release. 

If only I could believe that this is a **_FINAL_** decision.  :-)

>Thank you @mere-human for your effort (still).

Remember, it was mere-human that started this whole thing, what with those fancy new dialogs.
...which is nothing but praise for mere-human because the new dialogs are great!

mere-human:
> Remember, it was mere-human that started this whole thing, what with those fancy new dialogs.
> ...which is nothing but praise for mere-human because the new dialogs are great!

I plan to go on 🙂

PR #9436, opened 2021-01-21
Fix Switch-To keyboard shortcuts cannot be used in context menu

### LINKED ISSUE #9435: Switch-To-Keyboard shortcuts cannot be used in context menu
### Description of the Issue

This is a partial problem arisen in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9347#issuecomment-756129872. The Switch-To commands, intended to be available as keyboard shortcuts only, cannot be used in the context menu.

Add the following line to `contextMenu.xml`:
`<Item ItemNameAs="Switch to Project Panel 1" id="44104"/>`

### Expected Behavior
A new context menu item appears, named "Switch to Project Panel 1". Clicking this context menu item will switch the input focus to Project Panel 1.

### Actual Behavior
A new context menu item appears, it has an empty name, it is disabled and checked. Since it is disabled, no click on it is accepted.

### Debug Information
Notepad++ v7.9.2   (32-bit)
Build time : Dec 31 2020 - 03:58:36
Path : C:\Tools\Npp\Npp_792\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Current ANSI codepage : 1252

This issue has been created as an anchor for an upcoming PR, which will improve this behavior.

Uhf7:
Fixes #9435.

### DIFF
--- a/PowerEditor/src/WinControls/ContextMenu/ContextMenu.cpp
+++ b/PowerEditor/src/WinControls/ContextMenu/ContextMenu.cpp
@@ -112,14 +112,13 @@ void ContextMenu::create(HWND hParent, const std::vector<MenuItemUnit> & menuIte
 		{
 			lastIsSep = true;
 		}
-
 		if (mainMenuHandle)
 		{
 			UINT s = ::GetMenuState(mainMenuHandle, item._cmdID, MF_BYCOMMAND);
 			if (s != -1)
 			{
-				bool isEnabled = (::GetMenuState(mainMenuHandle, item._cmdID, MF_BYCOMMAND)&(MF_DISABLED|MF_GRAYED)) == 0;
-				bool isChecked = (::GetMenuState(mainMenuHandle, item._cmdID, MF_BYCOMMAND)&(MF_CHECKED)) != 0;
+				bool isEnabled = (s & (MF_DISABLED | MF_GRAYED)) == 0;
+				bool isChecked = (s & (MF_CHECKED)) != 0;
 				if (!isEnabled)
 					enableItem(item._cmdID, isEnabled);
 				if (isChecked)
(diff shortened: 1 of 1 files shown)


PowerEditor/src/WinControls/ContextMenu/ContextMenu.cpp
```diff
@@ -115,12 +115,16 @@ void ContextMenu::create(HWND hParent, const std::vector<MenuItemUnit> & menuIte
...
 		if (mainMenuHandle)
 		{
-			bool isEnabled = (::GetMenuState(mainMenuHandle, item._cmdID, MF_BYCOMMAND)&(MF_DISABLED|MF_GRAYED)) == 0;
-			bool isChecked = (::GetMenuState(mainMenuHandle, item._cmdID, MF_BYCOMMAND)&(MF_CHECKED)) != 0;
-			if (!isEnabled)
-				enableItem(item._cmdID, isEnabled);
-			if (isChecked)
-				checkItem(item._cmdID, isChecked);
+			UINT s = ::GetMenuState(mainMenuHandle, item._cmdID, MF_BYCOMMAND);
+			if (s != -1)
```
donho, line 119: I know MS says:
```
Return value
Type: UINT

If the specified item does not exist, the return value is -1.
...
```
But I thought it would generate warning on compiling, since you're comparing between an unsigned value and a signed one. Or is the flat earth plot true ?

Uhf7: Microsoft states here: https://docs.microsoft.com/en-us/cpp/cpp/type-conversions-and-type-safety-modern-cpp?view=msvc-160

> The compiler does not warn about implicit conversions between signed and unsigned integral types. Therefore, we recommend that you avoid signed-to-unsigned conversions altogether. 

Nice recommendation, but hard to follow. It's easier to live on the flat earth.

I could
- add `(UINT)` before `-1`
- change the type of `s` to `int`
- leave it as it is

All variants produce the same code. I tend to leave it.

In this moment, I see that I actually forgot to use the new variable `s`. So I will revise this PR and remove the 2 unnecessary calls of `GetMenuState`.

Issue #9423, opened 2021-01-19
Guidelines for supporting legacy systems.


ArkadiuszMichalski:
@donho 
There should be some official information about supporting for older systems. My point is whether the support for eg WinXP or Win7 has been completely abandoned, or is it still supported, for example, if the corrections do not require a lot of additional code.
This is important when responding to bugs, because some people report them from the older systems, and sometimes this bugs do not appear in the current Win10. What then, bugs should be closed?

If support for a given system has been completely dropped, you can add a short note as to which last version of NPP supported the given system. This information would then be good as ref links for future bugs. This can be given at https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md or in a separate subpage.

mere-human:
> whether the support for eg WinXP or Win7 has been completely abandoned, or is it still supported, for example,

With these changes, the Windows XP support has officially ended:
45912a3321877bc159ebffebf788dcbcc48c34ea
ca1bcb1ec0bedbb4aad227c369667aef30624618
The app just won't start on that OS.

> add a short note as to which last version of NPP supported the given system

This would be nice to have.

donho:
I just opened a PR for this issue:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9443

Not sure my way is clear enough. If not, please suggest your way.
Feel free to reword this PR.

Issue #9347, opened 2021-01-03
Focus does not return to a Project panel


zanud:
### Description of the Issue
If a Project panel has lost focus _temporarily_, then it will not get it back.

### Steps to Reproduce the Issue
There is a lot of different scenarios. For example:

**A) Operations of Notepad++**

1. Launch Notepad++.
2. Open any file.
3. Open the Project panel. (Optionally, you may create a project in it and add some files/folders to it. But this is not necessary.)
4. Invoke any operation that opens it's dialog window. The operation may be project-related or not - this is irrelevant. (For example, "Add file to the project" / "Remove file(folder) from the project" / "Remove project" / "Open Workspace" / "Save workspace as..." / "View" -> "Summary" / etc.) The only important thing - this operation must open some dialog.
5. Close the dialog by any button ("Yes" / "No" / "OK" / "Cancel" - irrelevant).

**B) Actions outside of Notepad++**

1. Launch Notepad++.
2. Open any file.
3. Open the Project panel. (Optionally, you may create a project in it and add some files/folders to it. But this is not necessary.)
4. Switch from Notepad++ to any other application (either by mouse click or by Ctrl-Tab).
5. Switch back to Notepad++ by Ctrl-Tab.

### Actual Behavior
After Step 5 focus will be on the file tab: cursor (caret) will blink in the text, and you will be able to move it by arrow keys or to type new text here.

### Expected Behavior
Focus should return to the workspace tree in the Project panel.

### Debug Information
Notepad++ v7.9.2   (32-bit)
Build time : Dec 31 2020 - 03:58:36
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Microsoft Windows XP (32-bit) 
OS Build : 2600.0
Current ANSI codepage : 1251
Plugins : none

ArkadiuszMichalski:
Temporarily, the shortcut may be used:
![image](https://user-images.githubusercontent.com/2730894/103782181-6c941700-5037-11eb-983c-930144a1ca86.png)

@sasumner This `Switch...` commands can be use as entry in context menu (they have any ids)? I don't see them directly in `View` menu.

zanud:
> Temporarily, the shortcut may be used

Does this really works for you?
[...]

PR #9443, opened 2021-01-23
Add guidelines for supporting legacy systems

donho:
This PR is concerning to #9423

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,13 @@ _Since the release of version 7.6.5 Notepad++ is signed using GPG with the follo
 https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 
 
+Supported OS
+------------
+
+All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ uses can afford a modern system. Here are the [table of supported system](SUPPORTED_SYSTEM.md) in case you are one of them.
+
+
+
 Build Notepad++
 ---------------
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 Supported OS
 ------------
 
-All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can afford a modern system. Here is the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
+All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can or want to use the newest system. Here is the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
 
 
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 Supported OS
 ------------
 
-All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ uses can afford a modern system. Here are the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
+All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can afford a modern system. Here are the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
 
 
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 Supported OS
 ------------
 
-All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ uses can afford a modern system. Here are the [table of supported system](SUPPORTED_SYSTEM.md) in case you are one of them.
+All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ uses can afford a modern system. Here are the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
 
 
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,8 @@ https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 Supported OS
 ------------
 
-All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can afford a modern system. Here are the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
+All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can afford a modern system. Here is the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
+
 
 
 
@@ -47,5 +48,4 @@ Contribution
 Code contribution is welcome. Here are some [rules](CONTRIBUTING.md) that your should follow to make your contribution accepted easily. 
 
 [Notepad++ Contributors](https://github.com/notepad-plus-plus/notepad-plus-plus/graphs/contributors)
-
 
--- a/SUPPORTED_SYSTEM.md
+++ b/SUPPORTED_SYSTEM.md
@@ -3,7 +3,7 @@
 
 * The column **OS** is the list of different MS Windows generations.
 * The column **can be run (version can be run)** is for the status of the latest release to be run under the OS. If the latest version cannot be run under the specific OS (with the status "NO"), then the last version number which can be run will be indicated in the paratheses.
-* The column **supported** contains the status of Notepad++ team supporting Notepad++ under the specific OS. It means if a bug is present under a OS no more supported, but not under supported OS, then this issue won't be treated by team.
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -28,6 +28,13 @@ _Since the release of version 7.6.5 Notepad++ is signed using GPG with the follo
 https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/nppGpgPub.asc
 
 
+Supported OS
+------------
+
+All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can afford a modern system. Here are the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.
```
mere-human, line 34:
```suggestion
All the Windows systems still supported by Microsoft are supported by Notepad++. However, not all Notepad++ users can afford a modern system. Here is the [Supported systems information](SUPPORTED_SYSTEM.md) you may need in case you are one of them.

```

SUPPORTED_SYSTEM.md
```diff
@@ -0,0 +1,21 @@
+
+## Supported OS Table
+
+* The column **OS** is the list of different MS Windows generations.
+* The column **can be run (version can be run)** is for the status of the latest release to be run under the OS. If the latest version cannot be run under the specific OS (with the status "NO"), then the last version number which can be run will be indicated in the parentheses.
+* The column **supported** contains the status of Notepad++ team supporting Notepad++ under the specific OS. It means if a bug is present under a OS no more supported but not under supported OS, then this issue won't be treated by team.
```
mere-human, line 6:
```suggestion
* The column **supported** contains the status of Notepad++ team supporting Notepad++ under the specific OS. It means if a bug is present under an OS no more supported but not under supported OS, then this issue won't be treated by the team.
```

mere-human (commented):
Maybe, it could be also nice to state that even if the OS version isn't supported, the developers should avoid breaking it. But not if the maintenance cost will be high.

donho:
> Maybe, it could be also nice to state that even if the OS version isn't supported, the developers should avoid breaking it. But not if the maintenance cost will be high.

Nice one. Please add such info in this PR.

sasumner:
>Maybe, it could be also nice to state that even if the OS version isn't supported, the developers should avoid breaking it. But not if the maintenance cost will be high.

I'm totally confused as to what this means.  @mere-human can you elaborate/explain?

ArkadiuszMichalski:
When does an unsupported system become blocked? If the team will skips bugs for unsupported systems (but still not blocked), this means that team can completely ignore them and, for example, break the functions that worked in it until now (when this changes not affect supported systems)?

donho:
@sasumner 
My interpretation is: the devs should be careful of avoiding the regression on unsupported OS while adding the new features (under the supported OS), not as what he has done @mere-human  :)

ArkadiuszMichalski:
@donho 
So please add such info because:
```
The column **supported** contains the status of Notepad++ team supporting Notepad++ under the specific OS. It means if a bug is present under an OS no more supported but not under supported OS, then this issue won't be treated by the team.
```
because it is not clear whether the changes should still be checked in unsupported (but not blocked) systems or not.

mere-human:
> > Maybe, it could be also nice to state that even if the OS version isn't supported, the developers should avoid breaking it. But not if the maintenance cost will be high.
> 
> I'm totally confused as to what this means. @mere-human can you elaborate/explain?

I remember we had a conversation related to this topic in the scope of the newest file dialogs. 
So maybe it makes sense to rephrase these comments and add that info to the doc.
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9378#issuecomment-761292695:

> I would say, if Notepad++ still runs under XP, it'll be shame to stop running under this OS, due to (only) this new folder dialog.
OTOH, we can stop "supporting" XP (for example the bug Focus does not return to a Project panel #9347 under XP won't be fixed.

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9378#issuecomment-760963142:

> IMO: Could be done, but it sounds like complicating the code unnecessarily.

> IMO: XP users want to use an ancient OS that is no longer updated by its maker, they can use older versions of N++ (also which are not updated by ITS maker) that DO support their OS.

mere-human:
> voiding the regression on unsupported OS while adding the new features (under the supported OS), not like what he has done @mere-human :)

Yep, unfortunately, shame on me. :pensive:
But as far as I remember, it was decided that supporting 3 versions of the dialog could be hard.

donho:
@mere-human 
> Yep, unfortunately, shame on me. 

I was kidding :)

Joke aside, XP is a special case - it was a good OS (not like vista) and at the same time it's very old (there are also a lot of security issues). So the decision is not easy. But still, the most people use Windows 10 now so there's no reason to not use the modern save/open dialog feature because of it. OTOH, sooner or later XP will be drop (by Notepad++) in the the medium/short term, so this new feature in open/save dialog gives the opportunity for dropping it - it won't be the same case for Windows 7. But at this point (concerning Windows 7), I think everyone agrees with that.

Uhf7:
> However, not all Notepad++ users can afford a modern system.

Just reading this in the new `nppGpgPub.asc`, I would challenge this statement: In most cases I don't use a "modern" system, this is because the new system sucks and I want to keep my old (working) system as long as possible. So perhaps this could be changed in something like "not all Notepad++ users can or want to use the newest system" or something like that.

donho:
@Uhf7 
Poor people have always their excuse :D 
But the excuse accepted.

mere-human (approved):

donho:
@ArkadiuszMichalski 
> When does an unsupported system become blocked?

Until a new feature related to system in Notepad++ (needed API) which doesn't be supported in the unsupported system.
There's no restrict rule for it. An example is, dropping Vista will be different from dropping XP. Dropping Vista will be easier because of its bad popularity.  

> If the team will skips bugs for unsupported systems (but still not blocked), this means that team can completely ignore them and, for example, break the functions that worked in it until now (when this changes not affect supported systems)?

The bugs should be reproduced under the newest supported OS. If they can't, then we can say the bug is due to the different implementation of API (Such case is rare, but it can happen). In such situation, The bug won't be treated since it's only in unsupported system.

donho:
@ArkadiuszMichalski 
> So please add such info because from:
...
it is not clear whether the changes should still be checked in unsupported (but not blocked) systems or not.

Sorry but it's not clear to me.
Could you add the sentence where it should be please?

chcg:
Probably for WinXP the latest SP3 is required to run N++ with 7.9.2. Maybe a general note should be added that the list is meant for the last SP of each version.
What about Win8.1?

ArkadiuszMichalski:
@donho 
>Sorry but it's not clear to me.
Could you add the sentence where it should be please?

What if modifying something by one PR (which now works fine in Win7 and Win10) will break working on Win7 (but in Win10 it will works fine).  Should a person working on PR  pay attention to this when someone reports such a problem? I don't expect that  person working on PR will test unsupported systems now, but the changes may break them, even in a slight sense. In such a situation, what should be done, looking for a working solution for all unblocked/unsupported systems, or completely not look at those old unblocked/unsupported  systems? 

Since the unsupported systems are not blocked and there is no need to test new PR in them, it's very easy to break things that work in them now.

Uhf7:
Remember the preprocessor definition `WIN32_WINNT` here. As long as this is not upgraded to a "new version", no Windows API functions of this "new version" can be used. So, the only thing to do here is to assure that no one dares to re-define `WIN32_WINNT`.

donho:
@ArkadiuszMichalski 
> What if modifying something by one PR (which now works fine in Win7 and Win10) will break working on Win7 (but in Win10 it will works fine).

I couldn't say it won't happen, because it has happened recently and we have had dropped XP. But since Notepad++ is based Win32 API, the situation you mentioned above happens rarely.

> Since the unsupported systems are not blocked and there is no need to test new PR in them, it's very easy to break things that work in them now.

I wouldn't say "very easy" but "it could happen". Well, a regression, even on an unsupported OS, should be fixed of course. However, an issue (that I can't find now) tested by @Uhf7 has the different behaviour under Windows 10 & Windows 7. In this case, since Windows 7 is unsupported OS, the bug won't be fixed.

ArkadiuszMichalski:
Btw, looks like Win7 is still popular:
[gs.statcounter.com](https://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide)
[netmarketshare.com](https://netmarketshare.com/operating-system-market-share.aspx?options=%7B%22filter%22%3A%7B%22%24and%22%3A%5B%7B%22deviceType%22%3A%7B%22%24in%22%3A%5B%22Desktop%2Flaptop%22%5D%7D%7D%5D%7D%2C%22dateLabel%22%3A%22Trend%22%2C%22attributes%22%3A%22share%22%2C%22group%22%3A%22platformVersion%22%2C%22sort%22%3A%7B%22share%22%3A-1%7D%2C%22id%22%3A%22platformsDesktopVersions%22%2C%22dateInterval%22%3A%22Monthly%22%2C%22dateStart%22%3A%222019-11%22%2C%22dateEnd%22%3A%222020-10%22%2C%22segments%22%3A%22-1000%22%7D)

donho:
@ArkadiuszMichalski 
> Btw, looks like Win7 is still popular:

Yes, I'm aware of it. It's for sure that Windows 7 will last for a while, just like XP.
And the most of support for Windows 7 in Notepad++ will kept.

PR #9449, opened 2021-01-26
Remove localization of "Use new style dialog" with id="6430"

mere-human:
The option isn't available since FileDialog was removed.

For more context, see discussion in #9403 and #9374


sasumner:
@mere-human 

You do that with N++ and regex _Replace in Files_ ?  :-)

mere-human:
> You do that with N++ and regex _Replace in Files_ ? :-)

Exactly, this is a great opportunity to test features 🙂

PR #9458, opened 2021-01-28
Update russian.xml to 7.9.3

GOLEM777:
This is not my fault.
I think problem is due to a line that was removed after create request.
Request was trying to delete a line id6430  that was no longer there.

-----
The file is fully working and tested!


GOLEM777:
This is not my fault.
I think problem is due to a line that was removed after create request.
Request was trying to delete a line that was no longer there.
-----
The file is fully working and tested!

donho:
OK, I understand.
Sorry for the inconvenience.
And thank you for your effort.

PR #9460, opened 2021-01-28
Default Open/Save file Directory more informative

GOLEM777:
capture_01282021_073704


sasumner:
Totally unnecessary.

donho:
I don't think this adding information is needed either.

PR #9469, opened 2021-01-30
Update italian.xml v7.9.3

conky77:


conky77:
@donho I don't know if I'm in time for the next version for Italian Translation

PR #9496, opened 2021-02-09
Make tab splitter menu translatable

### LINKED ISSUE #8955: Last not translated text
Please can add this for completeness.

1 - Incremental Search

capture_10052020_202353

~~2 - Rotate border~~    - **DONE!**

capture_09142020_201103

mere-human:
Rename menu IDs in tab splitter to be in English.

Address #8955 partially.


mere-human:
I changed the location of some controls so that the translated labels can fit.
There is no auto-layout feature by default in Win32 dialog.
By the way, what is the preferred way to deal with such issues?

Before:
Screenshot 2021-02-09 235220
After:
Screenshot 2021-02-08 225901

donho:
This part is not translatable because there's no room for the translation.

mere-human:
> This part is not translatable because there's no room for the translation.

Ok. Then, I'll remove the part for Incremental Search until we figure out a solution.

Issue #4949, opened 2018-10-23
Save sometimes does not work (file changes are not detected)


c33s:
### Description of the Issue
in very seldom cases the save button/shortcut does nothing. i thought it was gone after i updated to 7.5.4 but it happend again 2 times in the last 2 weeks. 

### Steps to Reproduce the Issue
for me it isn't really reproduceable, it just happens sometimes. 
1. have open about 30 text files (some saved, some never saved at all)
2. open a new file from windows (totalcommander shift + f4)
3. write some text "hello"
4. ctrl + s
5. write more text " world"
6. ctrl + s
7. close notepad++
8. reopen notepad (previous session is reopened)

### Expected Behavior
the file should contain "hello world"

### Actual Behavior
the file contains only "hello"
interesting fact is that before `3.` the save icon in the menu is blue, after `4.` the color change to grey (nonclickable) and writing more text normally would change it back to blue but it stays grey. this is no problem if you use the mouse for saving but if you use ctrl + s there is a very high chance you wont recognize it.

### Debug Information
```
Notepad++ v7.5.4   (32-bit)
Build time : Jan  1 2018 - 01:47:45
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll JSLintNpp.dll JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll NppTextFX.dll NppToolBucket.dll PluginManager.dll SessionMgr.dll XMLTools.dll 
```

shimscharf:
I can confirm this bug is a regular occurrence. Every time.

1) Create new file.
2) write text "hello"
3) save "filename.ext"
4) continue writing text "world"
5) save
6) close notepad
7) look at file --> only contains "hello"

In the last 2 versions of Notepad++ this has been occurring.

I have been been getting around it by:
1) create new file
2) write "hello"
3) save file
4) close file
5) open file
6) write "world"
7) save file

dsonewald:
Also in Notepad++ v7.5.8   (32-bit)
Build time : Jul 23 2018 - 02:03:53
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
[...]

PR #9498, opened 2021-02-10
Fix save of new files problems with Session Manager plug in.

### LINKED ISSUE #9475: Changes not saved for newly created *.R file!
### Description of the Issue
After creating and saving new *.R file, further modifications are not being saved! Neither in the c:\Users\Tomas\AppData\Roaming\Notepad++\backup\ directory, and not even when saved explicitly by Ctrl-S! Status of the file (the color of the disquette icon on the tab) is not updated.

1. Open new tab by Ctrl-N
2. Write some R code, e.g. "a <- 1"
3. Save as "a.R".
4. Add another line of code, e.g. "b <- 2". 
5. Save by Ctrl-S.

### Expected Behavior
In step 4, status of the file should update (disquette icon on the tab should turn red). After few seconds, the modified file should be seen in c:\Users\Tomas\AppData\Roaming\Notepad++\backup\. 

In step 5, changes should be saved in the file "a.R".

### Actual Behavior
In step 3, first version of the file is saved successfully, which is allright.

But in step 4, status of the file doesn't update! (Disquette icon on the tab stays blue, doesn't turn red). No changes saved in the c:\Users\Tomas\AppData\Roaming\Notepad++\backup\ directory.

In step 5, no changes are saved in the file "a.R"!!

### Notes

- Doing the same process with "*.txt" file shows no such problems. 
- It is only problem for newly created files. If I close the file and open it again in NPP, the problem disappears.

### Debug Information

Notepad++ v7.9.1   (32-bit)
Build time : Nov  2 2020 - 01:03:56
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1803
OS Build : 17134.1845
Current ANSI codepage : 1250
Plugins : mimeTools.dll NppAutoIndent.dll NppConverter.dll NppExport.dll NppTextFX.dll SessionMgr.dll 

There is 1.6 GB of free space on the disk.

Uhf7:
Fixes #9475 

Mechanics: The `FileManager` class uses `_pscratchTilla` to manipulate documents. The Session Manager plugin calls `NPPM_SAVECURRENTSESSION` while it is notified about the language change. While saving the session, `FileManager::docLength` is called, which should not switch the document attached to `_pscratchTilla`.

Maybe other functions of the `FileManager` class should restore the document pointer too.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1501,8 +1501,9 @@ int FileManager::getFileNameFromBuffer(BufferID id, TCHAR * fn2copy)
 
 int FileManager::docLength(Buffer* buffer) const
 {
+	Document curDoc = _pscratchTilla->execute(SCI_GETDOCPOINTER);
 	_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, buffer->_doc);
 	int docLen = _pscratchTilla->getCurrentDocLen();
-	_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, _scratchDocDefault);
+	_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, curDoc);
 	return docLen;
 }


donho:
@Uhf7 
Thank you for the PR. I just try to make the context more clear: 
So you have reproduced the bug with Session Manager plugin installed right?
I suppose that you have tested your PR and the bug has gone?

donho:
> While saving the session, FileManager::docLength is called,

Could you point it out (link) please?

Uhf7:
Ok, I was quite short in my description.

The original problem is, that the diskette icon doesn't turn red when modifying the file by typing something.

Prerequisite to do this is:
- Session Manager is active
- Auto save in Session Manager is turned on

The steps are:
1. Open a new file with Ctrl-N
2. Type something
3. Press Ctrl-S, and save the new file with a type different than .txt. The diskette icon turns blue now.
4. Type something again, the diskette icon remains blue, and the file will never be saved again.

The switch of the diskette icon between blue and red is done in `NppNotification.cpp`, when processing 		`SCN_SAVEPOINTREACHED` and `SCN_SAVEPOINTLEFT`. To make this work, a save point is needed before, which is created wile saving by a `SCI_SETSAVEPOINT` call here: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ea33a0cb765f93a974319d873d2b8a09a42c1878/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1061

I copied some lines from there to explain it:

		1004 _pscratchTilla->execute(SCI_SETDOCPOINTER, 0, buffer->_doc);
		...
		1057 buffer->setFileName(fullpath, language);
		1058 buffer->setDirty(false);
		1059 buffer->setStatus(DOC_REGULAR);
		1060 buffer->checkFileState();
		1061 _pscratchTilla->execute(SCI_SETSAVEPOINT);
		1062 _pscratchTilla->execute(SCI_SETDOCPOINTER, 0, _scratchDocDefault);

- In 1004, the document intended to save is assigned to `_pscratchtilla`.
- In 1057, the language of the current buffer is changed. This is causing the Session Manager to save the session, if the language has changed. If the language remains text, the Session Manager performs no auto-save. If the Session Manager calls `NPPM_SAVECURRENTSESSION`, then this calls the `FileManager::docLength` function, which resets the document assigned to `_pscratchtilla` to `_scratchDocDefault`.
- In 1058, the diskette icon turns blue.
- In 1061, the save point is set.
  - If the Session Manager saved the file, the save point is set for the document `_scratchDocDefault` (wrong). 
  - If the Session Manager didn't save, the save point is set for the document assigned in 1004 (the right one).

donho:
@Uhf7 
Thank you for the instructions. 

After installing Session Manager:

image

then I create a new file, type somethings, save to a file. The type something in the create file, it turns always to red, then I can save it.

Repeat the same procedure several time, still can't reproduce this bug. 
Am I missing something?

Uhf7:
Am I missing something?

- The Session Manager settings are the same as mine.
- Session Manager's About menu says it's version is 1.4.2.0.

Step 1: Open a new file. The diskette icon is blue:



Step 2: Type a char. The diskette icon is red now:



Step 3: Press Ctrl-S. Save as .R file:



The diskette is blue now:



Step 4: Type another char. The diskette icon remains blue:

mere-human:
> Am I missing something?
> 
> * The Session Manager settings are the same as mine.
> * Session Manager's About menu says it's version is 1.4.2.0.

I've just reproduced the issue with Notepad++ 7.9.3 and Session Manager 1.4.3.

mere-human (approved):
This PR fixed the issue for me.

Uhf7:
@mere-human,
I wonder, where version 1.4.3 comes from. I only found 1.4.2 here:

http://mfoster.com/npp/download.html

mere-human:
> @mere-human,
> I wonder, where version 1.4.3 comes from. I only found 1.4.2 here:
> 
> http://mfoster.com/npp/download.html

Plugin Admin:
image

ArkadiuszMichalski:
@Uhf7 
> I wonder, where version 1.4.3 comes from. I only found 1.4.2 here:

This is fork made by chcg https://github.com/chcg/npp-session-manager/.

Uhf7:
Thank you, @mere-human, thank you @ArkadiuszMichalski,

with your hints I found it under https://github.com/chcg/npp-session-manager/releases.

donho:
OK, I can reproduce with `.R` - I wonder why `.txt` doesn't trigger the bug.

Uhf7:
> I wonder why .txt doesn't trigger the bug.

Because the Session Manager plugin calls  `NPPM_SAVECURRENTSESSION` only, if it gets a language change notification. And this comes from here:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/d7819cf601051510b7fa1640ca17fedad297e1f8/PowerEditor/src/ScintillaComponent/Buffer.cpp#L203

And this is only called, if `_lang`  is changed from the default `L_TEXT` to - in this case - `L_R`.

shimscharf:
fixes this too:

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4949

PR #9504, opened 2021-02-12
Fix #9499 - JSON auto-indent for brace

### LINKED ISSUE #9499: Auto-indent in JSON file
### Description of the Issue
I believe JSON does not allow single line control structures without braces.  In other words, if I specify and object:

```
"myobject": { "key": "value" }
```

I need to use the braces `{` and `}`.  I can put them on separate lines:

```
"myobject": { 
    "key": "value"
}
```

but they need to be there.  When I type the `{` brace and press enter, Notepad++ does not auto-indent a "tab" length for me like it does in other language where braces delimit control structures (e.g., Perl, Powershell).

**!!! PLEASE !!!** Someone else weigh in on this and let me know if that assumption is correct??

I think this can be easily fixed by editing [this](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3419a749d9bb9fcd5dce1d98723dccb3170c3d0e/PowerEditor/src/Notepad_plus.cpp#L3070) to this:

```
			else if (type == L_PERL || type == L_RUST || type == L_POWERSHELL || type == L_JSON)
```

1. Set file type to JSON
2. Enter:
```
"myobject": { 
```

3.  Press "Enter" 

You get:

```
"myobject": { 
|
```

Where the pipe `|` is the cursor.

### Expected Behavior

```
"myobject": { 
    |
```

Where the pipe `|` is the cursor - indented by the language specific tab setting.

### Actual Behavior
No auto-indent.

### Debug Information
Notepad++ v7.9.3   (64-bit)
Build time : Feb  8 2021 - 18:16:35
Admin mode : OFF
Local Conf mode : ON
OS Version : 1909
OS Build : 18363.1316
Current ANSI codepage : 1252
Plugins :

vinsworldcom:
Fixes #9499 .  Adds JSON to list of languages that get an auto-indent after brace + Enter key.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3024,7 +3024,7 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 		return;
 
 	if (type == L_C || type == L_CPP || type == L_JAVA || type == L_CS || type == L_OBJC ||
-		type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS || type == L_PERL || type == L_RUST || type == L_POWERSHELL)
+		type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS || type == L_PERL || type == L_RUST || type == L_POWERSHELL || type == L_JSON)
 	{
 		if (((eolMode == SC_EOL_CRLF || eolMode == SC_EOL_LF) && ch == '\n') ||
 			(eolMode == SC_EOL_CR && ch == '\r'))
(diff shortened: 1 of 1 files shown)


mere-human (approved):
Works well.
This didn't fix the indent for brackets [] though (as it was stated in the issue discussion).
ezgif-7-21e7efcc4119

vinsworldcom:
@mere-human 

Thank you for the review.

> This didn't fix the indent for brackets [] though (as it was stated in the issue discussion).

Yes, sorry, thought that was clear - the brackets `[` are a separate issue as auto-indent for brackets does not work in any of the languages tested - C/C++, Powershell, Perl, etc.  I think if we want auto-indent for brackets `[`, I recommend opening a new issue and addressing it across all the file types, not just JSON.

Would you agree with that?

Cheers.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3024,7 +3024,7 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 		return;
 
 	if (type == L_C || type == L_CPP || type == L_JAVA || type == L_CS || type == L_OBJC ||
-		type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS || type == L_PERL || type == L_RUST || type == L_POWERSHELL)
+		type == L_PHP || type == L_JS || type == L_JAVASCRIPT || type == L_JSP || type == L_CSS || type == L_PERL || type == L_RUST || type == L_POWERSHELL || type == L_JSON)
```
sasumner, line 3027: It's quite a long "if" statement.  What's the "tipping point" for a possible change to switch/case ?

PR #9507, opened 2021-02-14
Add a preference to mute all sounds

### LINKED ISSUE #7950: Sound after an unsuccessful search cannot be disabled

mere-human:
Fix #7950


mere-human:
The new preference is located under the auto-updater check-box in the MISC section.
image

It allows to disable various sounds e.g. when a search has no match in a file.

sasumner:
IIRC, @donho was happy with the `PlaySound` implementation as it is, and didn't want an extra preference setting for this.
The sounds not relating to "unsuccessful" search are rare and have never been mentioned by a user AFAIK.

mere-human:
> The sounds not relating to "unsuccessful" search are rare and have never been mentioned by a user AFAIK.

See #7950
They still complain about these sounds. I think it is better to make an easy option to disable it once and for all.

I can make the option elsewhere or rename it or add a scrollbar in the MISC section.

mere-human:
The Notepad++ appears in the volume mixer only after it tries to make a sound.
Disabling the volume this way is not easy and not intuitive.

sasumner:
@mere-human 

Hmm, surprised you didn't "ask to be assigned" to this one before just "fixing" it.  Of course, not an absolute requirement--to be assigned--but it can shortcut potentially wasted time if an issue has a "history".  Hmm, lots of air-quotes on this :-).

Well, you put in some work, let's see if Don has had a change of mind on this.

Codingale:
I believe the Search category isn't a terrible place instead of Misc. I just hate that Windows never mutes it for longer than one restart or even the same session sometimes. On top of that, it's not like having one preference check is going to cause the entire codebase to run slower by a noticeable amount unless the implementation was done poorly. `PlaySound` also  doesn't respect Windows Sound Settings, as I have everything off and it still plays some beep from some archaic method that does what it wants.

mere-human:
> @mere-human
> 
> Hmm, surprised you didn't "ask to be assigned" to this one before just "fixing" it. Of course, not an absolute requirement--to be assigned--but it can shortcut potentially wasted time if an issue has a "history". Hmm, lots of air-quotes on this :-).
> 

Yes, sorry for rushing.
I just had some free time

donho:
@sasumner 
> IIRC, @donho was happy with the PlaySound implementation as it is, and didn't want an extra preference setting for this.
The sounds not relating to "unsuccessful" search are rare and have never been mentioned by a user AFAIK.

I was indeed. However, I've received several users' feedback for this requests because they don't find this option in Notepad++ : disable it in the system sound notification might be a solution but it's not certainly an obvious place for users.

donho (changes requested):
Revert 
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d3cb2ceb360c8f0d4219929466b2a7330de08597

PowerEditor/src/Parameters.h
```diff
@@ -838,6 +838,7 @@ struct NppGUI final
 	bool _monospacedFontFindDlg = false;
 	bool _findDlgAlwaysVisible = false;
 	bool _confirmReplaceInAllOpenDocs = true;
+	bool _muteSounds = true;
```
mere-human, line 841: Should I make it "false" by default?

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -416,7 +416,8 @@ BEGIN
     CONTROL         "Peek on document map",IDC_CHECK_ENABLEDOCPEEKONMAP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,269,71,140,10
 
 	// "Enable Notepad++ auto-updater" should be always the 1st item, because it'll be hidden if GUP.exe is absent
-    CONTROL         "Enable Notepad++ auto-updater",IDC_CHECK_AUTOUPDATE,"Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,37,94,217,10
+    CONTROL         "Mute all sounds", IDC_CHECK_MUTE_SOUNDS, "Button", BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP, 37, 94, 217, 10
```
mere-human, line 419: Should I add a localization for this string?

sasumner:
>So please Revert d3cb2ce

Don't forget to revert the docs (see https://github.com/notepad-plus-plus/npp-usermanual/commit/40dbb47f889de157fe2c0812b079e5bd8f820048 ) as well, when you are adding documentation about the new Preference setting

PowerEditor/src/Parameters.h
donho: Yes, please.

PowerEditor/src/WinControls/Preference/preference.rc
donho: Yes, please.
As well, add the modification in `english.xml`.

PowerEditor/src/Parameters.h
sasumner: I think the general idea is that existing behavior should be the default, so that is why this should be `false`

mere-human:
> > So please Revert [d3cb2ce](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d3cb2ceb360c8f0d4219929466b2a7330de08597)
> 
> Don't forget to revert the docs (see [notepad-plus-plus/npp-usermanual@40dbb47](https://github.com/notepad-plus-plus/npp-usermanual/commit/40dbb47f889de157fe2c0812b079e5bd8f820048) ) as well, when you are adding documentation about the new Preference setting

I will update the manual as soon as this PR is submitted.

donho (approved):

PR #9508, opened 2021-02-14
UI text improvments to ShortcutMapper

### LINKED ISSUE #9505: Shortcut Mapper UI text clarifications
Making this an issue rather than straight-to-PR in case there is need of debate over the non- `...` part.

### Description of the Issue

image

should be changed to:

image

Also, when pressing the _Delete_ button, the prompt can be misleading, as it implies that the shortcut keycombo and not the command will be deleted:

image

Suggest changing it to this for maximum safety:

image

### Debug Information

Comments based upon English text in N++ 7.9.2

sasumner:
Fix #9505


donho:
> I guess by going down this path, I have somewhat thwarted my own desires. :-(
Stupid me!

If you have some remorse about it, then there might be some problems in the guideline you provided.
We don't really need to apply this PR - I don't really care about this ellipsis and I agree with @ArkadiuszMichalski  "These dots just introduce more problems than benefits". So I will put this PR on suspend status, and let's see if it's really needed by user in 3 months. 

Does it sound good to you?

sasumner:
>If you have some remorse about it,

The guideline I provided I found on the internet, and I provided it because I thought it was interesting/valuable.
I didn't mean it as something I personally 100% agree with.
I thought it only fair to present it, because, knowing about it, I could not withhold it and feel I was being honest.
In the future if I find a relevant document that I don't completely agree with, I'll hide it and deny all knowledge of it. 
:-)

>I don't really care about this ellipsis

There has been an effort before (not by me) to add in `...` where it should be, and remove it where it should not be, so I thought it was something you cared enough about to try to "get it correct".

>"These dots just introduce more problems than benefits"

I'm not sure this is true?
Otherwise, why would other software use them?
What "problems"?

>see if it's really needed by user in 3 months

Well, there is no "user need" for it, at all.
So if you think users are going to yell for it...nope.  :-)

And, I was misled by the prompt that I've changed as part of this PR: the "...command entirely" one, so much so that I deleted a Run menu command when I just wanted to delete the shortcut.
So I think that part alone makes the PR worth it.  
Don will now point to PEBKAC. :-)

>Does it sound good to you?

I still think it is a good PR.
But why wait?  Accept/reject now, as you like...

donho:
> What "problems"?

The translation - we have to enlarge the buttons - and not for only ones which need ellipsis.

> But why wait? Accept/reject now, as you like...

I would reject it if I'm the only user of this project.
But it's not the case so I prefer to see if there are a lot of people think it's worth.

sasumner:
>>What "problems"?

>The translation - we have to enlarge the buttons - and not for only ones which need ellipsis.

I did it in the PR.  Took 5 minutes and a piece of paper to scratch some math on.  The buttons were a touch on the small side anyway, surprised that no translators ever complained.

>...a lot of people think it's worth...

It's not a detail that anyone would bother typing an opinion on.

Just...kill...it, get it out of sight.  :-)
I'd close it myself, but I refuse to close what I think has value.

ArkadiuszMichalski:
> What "problems"?

For example, entries in the context menu or buttons made with plugins will stop working.  Each such change is a potential corruption of them (when they operate on the name ). It would be a good idea to go through all the places where these dots are missing or unnecessary and change it by once. And for new items would be better to think about whether they need a ellipsis or not before they are added.

I do not consider UI elements, where adding them may require changing the dimensions of the elements, which in some translations may still not be enough.

sasumner:
>Each such change is a potential corruption of them (when they operate on the name ).

It's why they _shouldn't_ operate on the name, when at all possible.
If they _do_ operate on the name, right there they have trouble, because "localization" makes the name a moving target.

>And for new items would be better to think about whether they need a ellipsis or not before they are added.

Well, sure, in a perfect world this is done this way.  We're living in _this_ world, though.  Things get missed, or not thought of.

With the advantage of hindsight, this issue is "not that important", and I wish I'd never started it.

>I do not consider UI elements, where adding them may require changing the dimensions of the elements, which in some translations may still not be enough.

I'm not sure at all what that sentence is saying.

Suesskirsche:
As an ordinary user I am used to and expect software to stick to standards so I don't get confused. 

It took ages from the first DOS and windows applications to standardize the file menu to be named "File" and be on the leftmost position, followed be "Edit" and then "Help" to the rightmost position. Until then every company / developer created their own "standard", horrible to work with various programs having different names and positions for the same functionality! 

The ellipsis for buttons opening sub-windows requiring user input / selection also developed over several years, and I wouldn't want to miss it either! 

What made me hesitate to write my opinion is that I don't know if this is the right "channel" to plea for the ellipsis. Having just read the (I find excellent) suggestions in #9529 by Willy Wonca about redesigning the find & replace dialog and the reason for it's closure I fear my user's opinion is disturbing the developers talk. Unfortunately I don't know where else user's opinions should be expressed or wishes entered. Maybe on github (don't understand its concept as a non-programmer; was glad to enlist to this interesting exchange of thoughts)? 

Regards and thanks for developing and improving N++!


Am 17. Februar 2021 19:29:19 MEZ schrieb Don HO <notifications@github.com>:

sasumner:
> I fear my user's opinion is disturbing the developers talk. 

NO!  

>Unfortunately I don't know where else user's opinions should be expressed or wishes entered.

You're absolutely 100% in the right place!

donho:
Finally, I will reject this PR: I'm not convinced there's no strong need for this fixes. It brings more troublesome than the advantage.

PR #9531, opened 2021-02-17
Update Turkish localization

Tmp341:
* According to d6c9410, d5ad025, e26199a, 41c4180, adce1a8, 6e43ba6, 35584b3, 5c884a8 and a044cefe commits.


donho:
Thank you for the update. Could you adapt the diff of english.xml (v7.9.4 vs v7.9.3):
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9613

and confirm me when it's done? Thank you.

Tmp341:
I think it is done now. @donho

PR #9532, opened 2021-02-17
Fix extension appending depending on file type in file dialog

### LINKED ISSUE #9510: File doesn't save as extension type
### Description of the Issue
In Notepad++ 7.9.2 I can do "Save as", file name: "lol" and "Save as type": "Normal text file (*.txt)" and the output file will be "lol.txt". Choosing the same options in 7.9.3 and the output file is named "lol" with no extension.

I don't have the option "Set Save dialog file extension filter to \*.\*" checked.

Steps are above

### Expected Behavior
Above

### Actual Behavior
Above

### Debug Information
Notepad++ v7.9.3   (64-bit)
Admin mode : ON
Local Conf mode : OFF
OS Build : 7601.0
Current ANSI codepage : 1252

mere-human:
When changing file path, ensure both dialog data and edit control are set.
Use mouse button state change to track dialog confirmation.

Fix #9510

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -398,6 +398,9 @@ private:
 	// Sets the file name and waits until it is processed by the edit control.
 	void sendDialogFileName(const TCHAR* name)
 	{
+		// Update both the edit control and the dialog internal data.
+		if (_dialog)
+			_dialog->SetFileName(name);
 		::SendMessage(_hwndNameEdit, WM_SETTEXT, 0, reinterpret_cast<LPARAM>(name));
 	}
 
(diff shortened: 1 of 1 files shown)


mere-human:
This PR should fix the problems with the file dialog on Windows 7 as well.

donho (changes requested):
Fix: 
Press tab until the Save button has the input focus, then press return: .txt is not appended.
ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9510#issuecomment-779079803

sasumner:
@donho said:

>To be fixed:
Press tab until the Save button has the input focus, then press return: .txt is not appended.

Testing this PR, I agree with @donho findings.

Curiously, though, if you do this:

Press tab until the Save button has the input focus, then press **spacebar**: .txt ***IS*** appended!

mere-human:
> To be fixed:
> `Press tab until the Save button has the input focus, then press return: .txt is not appended.`
> ref: [#9510 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9510#issuecomment-779079803)

Originally, I tested this scenario and everything was fine. So I was puzzled as to what should I fix.
But then I tried to change the _Save as type_, then remove the extension from file name and then press the button (by any means).
It didn't appended the extension. Everything should be fine now.

Issue #8125, opened 2020-04-16
[Feature request] Make project files searchable by file list (not by directory).


Uhf7:
I use the "View->Project->Project Panel 1" command to open a project panel. Then I pick my project
files from various different directories. Then I save the workspace file with the self-choosen extension ".npj".

I set Preferences->MISC->"Workspace file ext." to ".npj". Now I can open Npp with my Workspace file as first parameter, and I'm back in my project. So far, so good.

My problem is, that the "Find in files" function of Npp cannot search through Project file sets. It can only search directories. This is bad for me, because my project files are kind of scattered around different places.

I miss the option "search all files in project", which most Compiler IDEs have. Now I could use a Compiler IDE, MS VC++ for instance, but the start up time of those IDE's is simply too long.

What I ended up with, was building an own version of Npp, which has additional check marks in the "Find in Files" dialog, Looks like this:

![FindInFiles](https://user-images.githubusercontent.com/40629737/79457986-6e253700-7ff1-11ea-9014-ab1dbf2cf02c.png)

Selecting one of the Project check marks tells Npp, that it should search all files listed in this Project panel.

This extension has the following impact:

- Modificatition of the Find in files dialog: 3 new check marks in FindReplaceDlg.rc, some shifting of existing controls, and 6 lines FindReplaceDlg_rc.h.

- Modificatition of the Find in files dialog: 3 new check marks in FindReplaceDlg.rc, some shifting of existing controls, and 6 lines FindReplaceDlg_rc.h.

- Changes to handle the new check marks (on/off/mutual exclusion against other check marks), ca. 110 Lines in FindReplaceDlg.cpp, 7 lines in FindReplaceDlg.h, 13 new lines in Parameters.cpp to
  save/load the state of the new checkboxes, 3 new lines in Parameters.h

- A new function ProjectPanel::enumWorkSpaceFiles(...),  ca. 30 new lines in ProjectPanl.cpp, 1 line in ProjectPanel.h

- Modification of the find function itself,  ca. 50 lines new or changed in Notepad_plus::findInFiles(...).

**The benefit is, that Notepad++ can now search through all files of a project, like the most Compiler IDEs can do.**

If you like this new feature, I would be happy to create a pull request, including only the modifications listed above, for reviewing.

sasumner:
Just kind of throwing this out there, without much thought, but why not just invoke this special functionality if the _Directory_ box is set to `Project1` for example?  Less impact to the existing UI.

[...]

PR #9534, opened 2021-02-18
Add function for searching files in project panels.

Uhf7:
Implements #8125


sasumner:
Did I push you into doing it?
If so, I was happy to do it.  :-)

Uhf7:
Screenshot is already in the issue, I didn't wanted to post it twice.

Uhf7:
Screenshot is already in the issue, I didn't want to post it twice.

Uhf7:
Screenshot is already in the issue, I didn't want to post it twice.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1268,6 +1271,7 @@ Continue?"/>
                     <Item id="3126" name="Save As..."/>
                     <Item id="3127" name="Save a Copy As..."/>
                     <Item id="3121" name="Add New Project"/>
+                    <Item id="3128" name="Find in Files.../>
```
chcg, line 1274: From https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/37827195/job/evdo8fivme4cv4gu/messages

> PowerEditor/installer/nativeLang\english.xml: PowerEditor/installer/nativeLang/english.xml:1275:17:FATAL:PARSER:ERR_LT_IN_ATTRIBUTE: Unescaped '<' not allowed in a

closing " is missing in line 1274

Uhf7: @chcg, Thank you. Some errors cannot be detected by simply compiling it. I should perform the xml checks too ...

sasumner:
>Screenshot is already in the issue, I didn't want to post it twice.

Ah, ok.  Often the implementation in the PR ends up to be different from the proposal (issue).

Uhf7:
> Often the implementation in the PR ends up to be different from the proposal (issue).

In this case not really. The function was implemented already at the time of the screen shot. The only difference is that I added localization and therefore it is "Project Panel _n_" now instead of "Project _n_", to use the same names everywhere.

sasumner:
Now looking at the screenshot in the issue...

I think the vertical increase in size of the _Find_ window is going to be objectionable.
Would it be reasonable to have _Project Panel 1_ (right where you have it) and then, to the right (at the same vertical level), have the two other checkboxes simply labeled "2" and "3" ?

>it is "Project Panel n" now instead of "Project n"

Perhaps go back to the simpler _Project n_ to save horizontal space, if the simple "2" and "3" idea is adopted.

ArkadiuszMichalski:
Some time ago I also wondering how to narrow it and I even did various combinations in paint :) Finally I ended up with:
```
  In project:
[ ]1  [ ]2  [ ]3
```
We save one line and have more space for translation in the first line. But I don't know if this is better than 3 separate lines.

Uhf7:
Like this?

FindInFiles_1

Would be fine with me, the only problem perhaps is the small room for the translations of "Project".

ArkadiuszMichalski:
It is worth checking which "Project" is the longest in the current translations.

mere-human:
How about using a dropdown combo box instead of the check boxes for "Project 1"?
It could be made wider if needed. And also, that could allow a potential extension to include "Folder as Workspace" there.
image

sasumner:
Good ideas, all.  

Of the current ideas,  @ArkadiuszMichalski 's idea of a 2-line solution might be the best so far.  

@mere-human A dropdown has merit, but wouldn't it eliminate the ability to, for example, search project 1 and project 3 at the same time?  I presume that is @Uhf7 's current design (haven't checked the code, but the UI design implies that).

OTOH, perhaps @donho won't object to increasing the vertical size of the window (but I seriously doubt that).

Uhf7:
@ArkadiuszMichalski

it would look like this:

FindInFiles_2

Could come with or without frame.

@mere-human,

the combo box has the problem, that there can only one Project Panel selected at the time. Not too bad for me. But it is harder to operate than a checkbox, I need mostly 2 clicks and some concentration to change a setting here. Like @sasumner wrote while I was creating this:

FindInFiles_3

But it's not exactly simpler than the checkmarks, in operating and in coding.

mere-human:
> the combo box has the problem, that there can only one Project Panel selected at the time.

The combo box seems more "generic" to me. It could be combined with other items. For example, "All Projects" could be added:
image
This is an example from Visual Studio.
But that's just my personal opinion.

> A dropdown has merit, but wouldn't it eliminate the ability to, for example, search project 1 and project 3 at the same time?

Do you think, this is going to be used often? 
For me, the more valid use case seems to be either "search a particular project" or "search all".

Sorry, if I am overcomplicating things.

sasumner:
As @Uhf7 just showed, this might be a better UI -- well, the radio button part -- because it makes clear that it is either a _Directory_ search OR a _Project_ search, not both:

image

And there might currently be vertical room to add something like this below the existing _Match case_ (refer to existing UI, NOT the screenshots above):

`O Panel : [x]Project 1 [x]2 [x]3 [x]Folder As Workspace`

sasumner:
@mere-human said:

>Do you think, this is going to be used often?

Don't know.

>For me, the more valid use case seems to be either "search a particular project" or "search all".

>Sorry, if I am overcomplicating things.

You're not.  This is hard-core UI design.  It's awesomely fun!  :-)

Uhf7:
> Dropdown can't be multiselected, as we have in browsers?

You cannot see the current selection in a drop down box with multiple selections.

sasumner:
>You cannot see the current selection in a drop down box with multiple selections.

I suppose the dropdown could show `(multiple)` for such a case (when it is not open), but would then require opening it to see what the true selections are.

donho:
My eyes was bleeding.
It's a good feature. Let's not ruin it with the terrible UI.
I would add "Find in Project" page, hide "Directory" and related checkbox, and put Panel controls instead.
So we keep the same UI, and it's not overloaded with all control, plus users know better where to do their specific operation. In term of code it's also isolated for this feature.
What do you think?

Uhf7:
@donho 

I think this is a very good idea. Now there is enough space again. Would you agree with this dialog?

FindInFiles_6

ArkadiuszMichalski:
I assume in the future [requests to include a Folder as Workspace search](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9534#issuecomment-781369144) here will be rejected? Now this can be called from the context menu on the folder in FAW panel and we're moved to Find in Files with automatic root selection.

Uhf7:
@ArkadiuszMichalski 

what is the difference between a search through "Folder as Workspace files" and a search through a directory? I mean, if the path specified in the directory field of the "Find In Files" subpage is the same which is used by the "Folder as Workspace" side window, then the results are the same, or am I wrong here?

sasumner:
Does moving the tab position of the _Mark_ tab to insert a new tab bother anyone?
I know that I have some automation that depends upon the Mark tab being index 3 here -- would this impact a good number of "real" users as well?  I can certainly change my automation, but I'm guessing this possibility would also bother @ArkadiuszMichalski ! :-)

sasumner:
>what is the difference between a search through "Folder as Workspace files" and a search through a directory?

One difference (for me) is that when I'm using FIF on a FAW, I find myself having to untick _In all sub-folders_, but I normally want to have that ticked when I'm doing a non-FAW FIF.  It is a bit annoying to have to look at this choice constantly.  I don't think it matters for the current issue, but it gave me a chance to talk about it. :-)

Uhf7:
> I have some automation that depends upon the Mark tab being index 3 

How can automation use tab positions in a dialog box of the UI? It's not the official plugin interface, isn't it?

sasumner:
>How can automation use tab positions in a dialog box of the UI?

So many things are possible with the scripting plugins. :-)
It's probably not a factor for any users but me, so, nevermind.
I would put any new tab at the end of existing tabs, but if @donho likes it where you've put it, that's what matters.

ArkadiuszMichalski:
@Uhf7
>what is the difference between a search through "Folder as Workspace files" and a search through a directory? I mean, if the path specified in the directory field of the "Find In Files" subpage is the same which is used by the "Folder as Workspace" side window, then the results are the same, or am I wrong here?

In practice none, it's search just one root folder, as we can do it manually. So asking for cheeckbox for FAW (a bit earlier ) looks strange for me. It shouldn't be a problem for anyone to keep it as it is. 
 
@sasumner 
> Does moving the tab position of the Mark tab to insert a new tab bother anyone?
I know that I have some automation that depends upon the Mark tab being index 3 here -- would this impact a good number of "real" users as well? I can certainly change my automation, but I'm guessing this possibility would also bother @ArkadiuszMichalski ! :-)

If only I knew how to automate tabs (by index) or other UI elements, then probably be a problem for me (having scripts already written). But if this is not strictly based on the tab's position in the window (pixels) then changing the index in the script shouldn't be a problem. Maybe elsewhere you can share how we can do something like this (I assume with PythonScript).

Uhf7:
> I would put any new tab at the end of existing tabs

I would order the tabs by their function, not by their age. But I actually don't care were the new tab will be. So it is an administrative decision.

sasumner:
@ArkadiuszMichalski I only said I thought it would bother you because you pointed out recently that adding/removing ellipsis from command names could break things for existing users; this is kind of similar, but probably only breaks things for me. :-)

sasumner:
>I would order the tabs by their function, not by their age.

IMO a good reason for putting it on the far right is that it would probably be the least-used tab by users.
Of course, for myself I have never used the "project" feature in N++ because there was no easy way to search the files making up the project -- this is changing now, so I will start using projects and thus maybe the tab actually will get a lot of use.  
But yes, "administrative decision" = all about Don. :-)

donho:
> I would put any new tab at the end of existing tabs

It makes more sense to me to put it beside of "Find in Files" tab since they have the similar functionality. As Notepad++ is still an oriented UI editor, making UI consistence is more important than making some (few) users changing the script, especially the scripting is not supported in Notepad++, IMO. 

> So asking for cheeckbox for FAW (a bit earlier ) looks strange for me.

Agree. User can always use the context menu command on FAW for searching more finely in some sub-directory, so I see no point to add "include a Folder as Workspace search".

donho:
@Uhf7 
> I think this is a very good idea. Now there is enough space again. Would you agree with this dialog?

I think it's the best compromise for the 3 checkboxes. Note that:
1. if no project checked, Find All and Replace All should be disabled.
2. "Find in Projects" context menu command should be added while right-clicking on "Workspace" of project panel.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1732,8 +1722,44 @@ bool Notepad_plus::findInFiles()
 		patterns2Match.insert(patterns2Match.begin(), TEXT("*.*"));
 	}
 
-	vector<generic_string> fileNames;
-	getMatchedFileNames(dir2Search, patterns2Match, fileNames, isRecursive, isInHiddenDir);
+	if (not dir2Search [0])
```
donho, line 1725: Please use modern C++ symbol.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1732,8 +1722,44 @@ bool Notepad_plus::findInFiles()
...
+		{
+			_pProjectPanel_2->enumWorkSpaceFiles (NULL, patterns2Match, fileNames);
+		}
+		if (_findReplaceDlg.isProjectPanel_3() && _pProjectPanel_3)
+		{
+			_pProjectPanel_3->enumWorkSpaceFiles (NULL, patterns2Match, fileNames);
+		}
+		if (fileNames.size() == 0) return false;
+	}
+	else if (not ::PathFileExists(dir2Search))
```
donho, line 1741: Please use modern C++ symbol.

Uhf7: Can you elaborate, please? What symbol is not modern? I'm unable to see it right now.

mere-human: @Uhf7 I think it is the `not` operator alias. I remember I had similar comment.

Uhf7: So '!' is preferred?

sasumner: Yes, `!` is preferred.  
Don and I went through this ourselves a while ago.  I like `not` because it is more readable (and Python uses it), but apparently `not` in C/(C++) is an antiquated notion, but supported.

ArkadiuszMichalski: Maybe it is worth mentioning in  in [Coding style](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#coding-style)?

Uhf7: OK, I was confused, because some sites in the internet say

> In Modern C++, you can also use and, or and not as boolean operators which means &&, || and ! respectively which is identical to languages like Python. 

Source of this example is https://dev.to/delta456/modern-c-and-or-and-not-as-boolean-operators-2jgf, but I'm sure I have read it on other places too. 

So, the actually request here is not to use `not` modern C++?. Means `!` instead of `not`. But `not` is modern, `!` is old. Or I'm confused.

sasumner: When Don and I debated it, IIRC he proved to me that `not` was old-style, or at least well-enough to satisfy me then.  Since `!` is most widely used in the current code-base for Notepad++, it seems fine to stick with it; plus it is  what Don prefers.  If a wide group of people prefer `not` , @ArkadiuszMichalski 's suggestion to put something about using `!` / `&&` / `||` in the contributing guidelines is smart.

Uhf7: Sorry, IIRC??? Don't know this shortcut. Please help.

sasumner: IIRC = If I Recall Correctly
I thought that one was very common.
(Here's where @ArkadiuszMichalski jumps on me because I got on him one time for using some lesser known abbreviation in a posting here)

Uhf7: @sasumner , "IIRC", Thank you, I decrypt more and more of such shortcuts, but this one was not available on the Internet instantly.

So, the essence here is, that there is a different opinion of what is modern and what is not.

Since I don't care about writing `!` or `not`, I will change it to `!`. I hope, this was really the problem @donho meant. Thank to @mere-human, @sasumner and @ArkadiuszMichalski for the hints.

donho: @Uhf7 Yes, I confirm `!` is more modern operator than "not", I cannot find the link right now, but for the sake of coding style in Notepad++, please use "!" instead of "not".

donho (commented):
Add "Find in projects" tab beside of "Find in files" tab by adding the 3 checkboxes. Note that:

1. if no project checked, Find All and Replace All should be disabled.
2. "Find in Projects" context menu command should be added while right-clicking on "Workspace" of project panel.

donho (changes requested):
Add "Find in projects" tab beside of "Find in files" tab by adding the 3 checkboxes. The following points should be done:

1. if no project checked, Find All and Replace All should be disabled.
2. "Find in Projects" context menu command should be added while right-clicking on "Workspace" of project panel.

PowerEditor/src/Notepad_plus.cpp
sasumner: @donho Add to contributing guidelines?:

```
Please conform to existing coding style when contributing to the Notepad++ project.  Observe the code near your intended change, and attempt to maintain the style of that code.  As an example, the majority of the Notepad++ code uses `!` , `&&` and `||` operators instead of `not`, `and`, and `or`.  So if you submit code that uses `not`, etc., it won't be accepted.
```

Or similar...

donho: @sasumner 
Is it really necessary? It's just common sense to me.
OTOH, I will clean up all the `not` operators in the project - I didn't aware of it from the beginning (I did some "contribution" probably surly).

sasumner: @donho said "is it really necessary?"

Well, if it had just come up ONCE, I'd say no (reference our discussion some time ago concerning it).
But it came up earlier this week with @mere-human code.
And now it has resurfaced again with @Uhf7 code.
Needing an explanation each time.
As you like, draw your own conclusion as to if it is necessary.
But...there were other nice phrases in my suggestion...the parts about taking note of the code around you, and mimicking its style.
Again, always your choice.

donho: @sasumner 
OK, you're right.
Please do a PR for that.

donho: Done in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/dbac42edd0f9c25bfdd8a90cf59ec4f701352f53
There were a bunch of "not" !!!

donho: @Uhf7 
Sorry for the conflict :(

ArkadiuszMichalski:
I test it and it seems to work fine. I'm also wondering if adding checkbox to `[] Include hidden files` wouldn't make it easy to skip something (if necessary), simillar what we have for `Find in Files`, as `[ ]In hidden folders`. We have here one empty line (because drop here `Directory` filed).

Uhf7:
@ArkadiuszMichalski,

Thank you for testing.

I don't like the concept of hidden files in a project very much. But if a hidden file is part of a project, it should be searched too.

The "Match whole word" options is available on all pages of the Find dialog, so it should remain in place and not move around while switching pages.

ArkadiuszMichalski:
But that will not change position of "Match whole word" , it will be where he is now. Unless it cannot be added without changing the position.
image

> I don't like the concept of hidden files in a project very much. But if a hidden file is part of a project, it should be searched too.

But with a switch, you can decide whether to search or not (default can be as search). It can be convenient to exclude certain areas quickly without modifying the whole project.  This is nothing urgent and can be discuss as a separate bug. I also have one more question regarding the search in the project, but not here. Let it finally pass :)

sasumner:
MY "CONFUSION":
The nomenclature of "project" and "workspace" is somewhat confusing in its usage in Notepad++.  I suppose it is so because the way to using the feature is through something called "project".  Meaning, get started via _View_ > _Project_ > ...
But then we have what is called a project _below_ the "workspace".
So we have a double-usage of the term "project".
If I think of "workspace" as Visual Studio's "solution" concept (and forget about the "upper level project naming" in N++, then it gets easier for me.

So in light of MY CONFUSION, and perhaps because I wasn't reading the comments in this thread tightly enough, I was under the impression that this new _Find in Project_ feature would allow me to search only in the active "project" (of the current workspace), not the entire workspace at once.  Search only the active "project of workspace" would meet my need as a user much better that the entire thing, but that could be just the way I use it, and not representative of most users.

But it might be nice to have both capabilities, much like Visual Studio allows "Entire Solution" and "Current Project".
If user right-clicks the top-level workspace node in a "Project X" panel and there is a "search" choice, it could take the user to the new _Find in Projects_ tab which would be set to search the entire workspace.  If user right-clicks a project-under-workspace in a "Project X" panel and there is a "search" choice, it could again take the user to _Find in Projects_, this time set (however that would look) for only searching that project.

Just some random thoughts.
I'm certainly not trying to up-scope this PR.

Uhf7:
The terminology confused me too. My main problem was, that there is a "Workspace" in Project Panel,  and there is a function "Folder as Workspace", and both have nothing to do with each other.

I prefer to hide the word "Workspace" in the Project Panel by saving it in a file. Then the file name of the workspace file is displayed, and I can regard it as what it is: A group of projects. And the right click on this group has now a menu entry "Find in Projects...", which does exactly what it says, search through all projects in this group.

The possibility to click on any node below the workspace (project group) to search only through the files in this subtree would be a nice thing too, but I wouldn't do it all at once.

ArkadiuszMichalski:
I vote for it, in the Polish translation I try don't use "workspace" word in relation to the project. Note that in `Preferences> MISC.` we have `Workspace file .ext:`, wouldn't it be better to use `Project file .ext:`?

@Uhf7 
>The possibility to click on any node below the workspace (project group) to search only through the files in this subtree would be a nice thing too, but I wouldn't do it all at once.

Heh, this is exactly the question I was planning to ask about the search https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9534#issuecomment-784311195.

sasumner:
>I vote for it, in the Polish translation I try don't use "workspace" word in relation to the project (because of the Folder as Workspace function). Note that in Preferences> MISC. we have Workspace file .ext:, wouldn't it be better to use Project file .ext:?

This gets to the bigger question of, should the nomenclature be changed in a global way in Notepad++.
It's bigger than just what is in this PR.

Uhf7:
@ArkadiuszMichalski,

it's actually a Project Group file. If it would be called Project file, then there could be confusion about the multiple projects in it. 

The only reason for me to get rid of the word "Workspace" here would be the similar sounding "Folder as Workspace" function. But in the moment, I only see "Workspace" while creating a new Project Group, so I don't care too much about the name.

sasumner:
@Uhf7 said:

>The possibility to click on any node below the workspace (project group) to search only through the files in this subtree would be a nice thing 

+1

ArkadiuszMichalski:
> it's actually a Project Group file.

I understand it like that we have something like a project, which is saved in a separate file. We have 3 panels for loading single project (from project file or live adding). Each such project can have different things added, we can call them labels, folders, sub-projects, works, jobs, etc., but as a whole for me it is just one project that is displayed in the project panel (as the name suggests).

image

Of course, it's just playing with words and as everyone explains it in their own way. I will just add that there are `<Project>` nodes in the project file, so talking about Project Group file also makes sense. Well, it is not a place for this type of discussion, because we want to finally be able to search it :)

donho (approved):

sasumner:
@Uhf7 Can you provide user manual updates for the feature you've added?  See https://github.com/notepad-plus-plus/npp-usermanual/issues/166

sasumner:
Is it reasonable to wish for an entirely separate _Filters_ when using _Find in Projects_ versus _Find in Files_ ?
For me, if I'm searching in a project I've defined, then I want it to search all of those files every time, equivalent of `*.*`.
But if I'm doing a non-project file search I typically don't want this.
It is annoying to be doing both types in a session and having to change _Filters_ often.
I can certainly see where there are times where a user _would_ want to specify a filter when using a project...but it is likely a different filter than a non-project search.
Thoughts?

mere-human:
> Is it reasonable to wish for an entirely separate _Filters_ when using _Find in Projects_ versus _Find in Files_ ?

Do you mean, to make these _Filters_ separate in different tabs?

sasumner:
>>Is it reasonable to wish for an entirely separate Filters when using Find in Projects versus Find in Files ?

>Do you mean, to make these Filters separate in different tabs?

Well, that would help ME and the way I (will) use it.  But I'm just one user.  :-)  That's why I'm asking for opinions.
So the general idea might be that there is a _second_ filters, only for _Find in Projects_.  The UI does _NOT_ change.

Uhf7:
> Is it reasonable to wish for an entirely separate Filters when using Find in Projects versus Find in Files

Seems reasonable. I actually don't use _Folder As Workspace_ and _Project Panel_ at the same time, but for someone who does it may be better. And it should be easy to implement.

sasumner:
>I actually don't use Folder As Workspace and Project Panel at the same time

To be clear, I wasn't referring to FaW, but just the general _Find in Files_ situation.
But, both of these are similar in that when searching, you typically want _some_ filter in place -- because maybe you don't control what's in the folders besides the files you typically edit.

Contrast this with _Find in Projects_ where you most-often (but, granted, not 100% of the time) want everything -- because you've directly specified what belongs to the project, and thus, by inference what doesn't.

If no one else comments in a negative fashion on the idea of a second set of filters, I'll make a separate issue of it.

Uhf7:
>  I wasn't referring to FaW, but just the general Find in Files situation

Yes, you are right, Find in Files works independent from FaW, I mixed this up.

Suesskirsche:
How about "remembering" the last x searches and offering them in a dropdown list (most recent search on top, oldest at the bottom) where they can be reactivated easily? 
Actually I'd like to have that in Word because there I could really need it... ;-)
But it would be neat in N++ too.

Am 7. März 2021 14:45:58 MEZ schrieb sasumner <notifications@github.com>:

sasumner:
@Suesskirsche said

>How about "remembering" the last x searches and offering them in a dropdown list (most recent search on top, oldest at the bottom) where they can be reactivated easily?

First, this is rather a bit too off-topic for this discussion thread.
This discussion thread is in relation to the feature of the PR.

Second, what you're asking for is currently the behavior in Notepad++.

PR #9543, opened 2021-02-19
TiXmlBase::StringEqual: ignoreCase parameter is mistreated

QoL15155:
**`tinyXml - StringEqual`** :
The parameter _ignoreCase_ is reversed - so we don't ignore case when we should.

That means that in TiXmlNodeA::Identify for instance:
``` c++
const char* xmlHeader = { "<?xml" };
const char* commentHeader = { "<!--" };

if ( StringEqual( p, xmlHeader, true ) )
```
string '<?xml' would be compared as case sensitive, but  _commentHeader_ on the other characters would be lower-cased. (performance)

There are others examples showing the parameter should be treated in reverse (and not just renamed)

### DIFF
--- a/PowerEditor/src/TinyXml/tinyXmlA/tinyxmlparserA.cpp
+++ b/PowerEditor/src/TinyXml/tinyXmlA/tinyxmlparserA.cpp
@@ -293,7 +293,7 @@ bool TiXmlBaseA::StringEqual( const char* p,
 	{
 		const char* q = p;
 
-		if (ignoreCase)
+		if (!ignoreCase)
 		{
 			while ( *q && *tag && *q == *tag )
 			{
--- a/PowerEditor/src/TinyXml/tinyXmlA/tinyxmlparserA.cpp
+++ b/PowerEditor/src/TinyXml/tinyXmlA/tinyxmlparserA.cpp
@@ -920,6 +920,7 @@ const char* TiXmlUnknownA::Parse( const char* p, TiXmlParsingDataA* data )
 	if ( !p )
 	{
 		if ( document )	document->SetError( TIXMLA_ERROR_PARSING_UNKNOWN, 0, 0 );
+		return 0;
 	}
 	if ( *p == '>' )
 		return p+1;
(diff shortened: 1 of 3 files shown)


donho:
Thank you for your PR.
However, you have to report the bug in the issue tracker, provide the description of bug and instructions to reproduce the bug. Then create a PR to refer the bug fixed in this PR.

sasumner:
Agree.  I was very confused as to why you think this code change is necessary, given that the code you changed has been in place for a long period of time with no trouble reported.

QoL15155:
Hi,
Sorry if I made a mess, this is actually my first time pushing into an open source github repository.
There are actually 3 commits here, each fixes a different issue.
I guess  the first one is kinda dangerous in term of backward compatability (though I still think it should be fixed), but the other two are pretty straight forward bugs.

# TiXmlBase::StringEqual
The argument's name is "ignoreCase", where in fact it does the opposite, when 
_ignoreCase_ is on, we don't ignore the case and vice versa. I think that either the name should be changed or the logic to avoid confusion. Tracking who calls _StringEqual_ function, I think suggest the logic is what should be changed. 

As for the 2 other commits, each describes a buffer overflow.

# TiXmlUnknown::Parse

Very hard to replicate, but its easy to see in the code that if `p==0` and we dont' return, then on condition `(*p=='>')` we try to dereference the address 0x00000000 and crash

``` c++
	if ( !p )
	{
		if ( document )	document->SetError( TIXMLA_ERROR_PARSING_UNKNOWN, 0, 0 );
	}
	if ( *p == '>' )
		return p+1;
```

# Shortcut::setName

Can be replicated by editing *shortcuts.xml* and adding a __Command_ with 64 characters (or more).
For instance:
``` xml
<Command name="noaksdlkasjdfsfsdfsdfsdfsdfsdfsdfsdgfsdffssdaqwekawkeawerjerwerjwerlaaskdjalsdjalskdjalsdjalskdjasld" Ctrl="no" Alt="yes" Shift="no" Key="111">http://www.php.net/$(CURRENT_WORD)</Command>
```
Then open-up notepad++ and under _Run_ menu, as consequences of the overflow there will be a shortcut with an empty name representing our new command name. 

As it happens to work, when __name_ variable is overflow with off-by-one, it zeroes out __menuName_ variable (since in memory __menuName_ comes after the __name_ variable)

sasumner:
>Sorry if I made a mess, this is actually my first time pushing into an open source github repository.

No problem, just read and follow the contributing guidelines.

>There are actually 3 commits here, each fixes a different issue.

So this is bad, you should open a separate "issue" for each.
I see you've broken them down above and explained them, but you really need to copy and paste that stuff into 3 new issues.
Then do a separate PR for each.

Issue #9493, opened 2021-02-08
Add some instruction how to test latest NPP build.


ArkadiuszMichalski:
I often see people reporting something that has been fixed in last build version. Maybe they don't know how to check something in this version of the program without compiling it (via the last appveyor).

Some detailed step-by-step instructions on how to do it in the portable version (maybe with pics) would be recommended. And when opening new bugs, you should link to it somewhere at the very beginning.

mere-human:
This seems like a good idea.
But where is this supposed to be?
Manual page? https://npp-user-manual.org/

sasumner:
>I often see people reporting something that has been fixed in last build version.

They do that, often?

Actually, I didn't know about getting the last-build, myself, so I found out:

This link gets you to where you can download the "current build" version:  https://ci.appveyor.com/project/donho/notepad-plus-plus

At that, if you click on one of the debug jobs (non-ming) you will be taken to a screen where you can download an .exe by clicking on the "artifacts" link.

The .exe is 99% of the time the only thing you need.  You can paste that .exe alongside your (relatively) current notepad++.exe (it has a different name so it won't overwrite), and then to test it all you need to do is exit Notepad++ and restart it using that different exe.

ArkadiuszMichalski:
> They do that, often?

Yes, at least I've seen a couple of them not so long ago. Additionally, some people don't realize that checking "some bug" in the last build is very easy as you described above, like here https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9488#issuecomment-774554736. It would be good to have it described somewhere to just link how to do it, and not to duplicate this description every time.

sasumner:
I'm not sure where it would go.
"Build instructions" would be a candidate location, but it isn't an exact fit there I guess.

ArkadiuszMichalski:
I think separate subpage will be better, and put its link below "Build Notepad++" (https://github.com/notepad-plus-plus/notepad-plus-plus#build-notepad), as another header like ""How test different Notepad++ version". The description will take into account the build from appveyor", but also general notes how to use the portable version for test (because thanks to it we can test different version without mess up the current installation, if there is one). 

Also it makes no sense to repeating that someone should checks if the portable version also has some error, since we can just give him a link to the description. Many of the reported problems are due to outdated settings through the installation version (located in user folder), and such a thing can be determined immediately thanks to the testing in fresh portable version.

sasumner:
[...]

PR #9552, opened 2021-02-21
Fix single-step Replace for regex backslash K case

### LINKED ISSUE #8434: Regular expression containing \K does not work with step-by-step Replace

sasumner:
Fix #8434

Let's consider it experimental.
Hopefully @guy038 will give it some testing time.

All the credit for the algorithm to @Uhf7.
I'm not a big fan of the logic here, but I considered some other ways without success, so...

Since it is experimental, I will close this PR after the appveyor build completes cleanly.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1743,9 +1743,7 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
 	(*_ppEditView)->execute(SCI_SETSEARCHFLAGS, flags);
 
 	bool regexSearchContainsBackslashK = false;
-	if ((pOptions->_searchType == FindRegex) && 
-		((findNextType == FINDNEXTTYPE_REPLACENEXT) ||
-		(findNextType == FINDNEXTTYPE_FINDNEXTFORREPLACE)))
+	if ((pOptions->_searchType == FindRegex) && (findNextType == FINDNEXTTYPE_FINDNEXTFORREPLACE))
 	{
 		bool escaped = false;
 		for (TCHAR const *p = pText; *p != 0; ++p)
@@ -1795,14 +1793,7 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
 			}
 
 			//new target, search again
-			if (regexSearchContainsBackslashK)
-			{
-				posFind = (*_ppEditView)->searchInTargetWithReachback(pText, stringSizeFind, startPosition, endPosition);
-			}
-			else
-			{
-				posFind = (*_ppEditView)->searchInTarget(pText, stringSizeFind, startPosition, endPosition);
-			}
+			posFind = (*_ppEditView)->searchInTarget(pText, stringSizeFind, startPosition, endPosition);
 		}
 
 		if (posFind == -1)
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1742,8 +1742,34 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
 
 	(*_ppEditView)->execute(SCI_SETSEARCHFLAGS, flags);
 
+	bool regexSearchContainsBackslashK = false;
+	if ((pOptions->_searchType == FindRegex) && 
+		((findNextType == FINDNEXTTYPE_REPLACENEXT) ||
+		(findNextType == FINDNEXTTYPE_FINDNEXTFORREPLACE)))
+	{
+		for (TCHAR const *p = pText; ; ++p)
+		{
+			p = wcsstr(p, TEXT("\\K"));
+			if (p == NULL)
+			{
+				break;
+			}
+			if ((p == pText) || (*(p - 1) != TEXT('\\')))
+			{
+				regexSearchContainsBackslashK = true;
+				break;
+			}
+		}
+	}
 
-	posFind = (*_ppEditView)->searchInTarget(pText, stringSizeFind, startPosition, endPosition);
+	if (regexSearchContainsBackslashK)
+	{
+		posFind = (*_ppEditView)->searchInTargetWithReachback(pText, stringSizeFind, startPosition, endPosition);
+	}
+	else
+	{
+		posFind = (*_ppEditView)->searchInTarget(pText, stringSizeFind, startPosition, endPosition);
+	}
 	if (posFind == -1) //no match found in target, check if a new target should be used
 	{
 		if (pOptions->_isWrapAround) 
@@ -1765,7 +1791,14 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
 			}
 
 			//new target, search again
-			posFind = (*_ppEditView)->searchInTarget(pText, stringSizeFind, startPosition, endPosition);
+			if (regexSearchContainsBackslashK)
+			{
+				posFind = (*_ppEditView)->searchInTargetWithReachback(pText, stringSizeFind, startPosition, endPosition);
+			}
+			else
+			{
+				posFind = (*_ppEditView)->searchInTarget(pText, stringSizeFind, startPosition, endPosition);
+			}
 		}
 
 		if (posFind == -1)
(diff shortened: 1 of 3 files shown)


sasumner:
@guy038

For getting the version to test, try out these beta instruction steps:  https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9493#issuecomment-781344297

Look for the heading `Using a pull-request (PR) version of Notepad++`

Provide feedback on both:
* those instructions
* the functioning of this PR

Closing now and awaiting feedback.

Uhf7:
@sasumner,

Thank you for the PR, I just tested it.

I think, `searchInTargetWithReachback` should be applied only for `FINDNEXTTYPE_FINDNEXTFORREPLACE`. Reason: The only thing we want to reach here is, that the Replace command, which follows immediately after a successful Find command, works correctly. We don't want to change the behavior for finding the next occurrence after the replace.

Example:

	Jack hacks
	Jack hacks
	Jack hacks

Search for the Regex `Ja.*?\Kck`. Replace with X.

Replace All results in

	JaX hacks
	JaX hacks
	JaX hacks

Single step replace results now in 

	JaX haXs
	JaX haXs
	JaX haXs

If `searchInTargetWithReachback` would be applied only for `FINDNEXTTYPE_FINDNEXTFORREPLACE`, then Single step replace and Replace All would have the same results again.

sasumner:
>I think, searchInTargetWithReachback should be applied only for FINDNEXTTYPE_FINDNEXTFORREPLACE. Reason: The only thing we want to reach here is, that the Replace command, which follows immediately after a successful Find command, works correctly. We don't want to change the behavior for finding the next occurrence after the replace.

You beat me to it on this one.  I was already working on the next commit for this which does exactly this. :-)

Uhf7:
Works fine for me now, but I miss the limitation of how far it does reach back in line 2371 of `ScintillaEditView.cpp`. It reaches back now to the beginning of the file

	size_t fp = 0;

but the reach back should be limited to some position before the current position

	size_t fp = 0; // (fromPos > 2048) ?  fromPos - 2048 : 0;

There is a Pro and a Con for that limitation:
- Pro: no performance issues in large files
- Con: if the expression skipped by `\K` is larger than the limit, then it still doesn't work.

To test the performance issues, I made a file with16384 lines of 183 repetitions of "Jack hacks ", each line was about 2012 columns long. The file length displayed was 32,997,376 bytes. No, I went to line 16384 and searched `Ja.*?\Kck`. Then I pressed Replace. Without the limit, it was kind of delayed.

Although the limit can be much larger without big lags after pressing Replace, we have to take into account here too, that users can create incredibly long regular expressions, which may take longer to evaluate than this easy one.

sasumner:
@Uhf7 

Correctness was favored over performance.
The potential performance impact was probably why I let the issue idle for so long.
If this PR is judged to be worthy of inclusion, I leave it to a higher authority to decide if reachback needs to be limited or can go all the way back to start-of-file.

sasumner:
@mere-human So I reopened this to "mark as draft".  In the past, I've just closed a PR if it wasn't ready for merging and then reopened to add more commits (for build check) or if it was finally ready.  So I expected to see "mark as ready" or something like that where the link text for "mark as draft" was, but nope.  So I'm not sure what to do when it is finally ready now. :-)  But, I suppose I will figure that out when/if the time comes. :-)

sasumner:
So, in response to @guy038 's comments [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8434#issuecomment-784583153) 

_**Side note**:  Guy, you should comment on a proposed solution in the PR comments for that "solution", NOT in the "issue" comments.  The reason for this is that a PR could be poor (and to be discarded) while an "issue" is totally valid and is waiting for a better solution (a different PR).  If you comment on a specific PR in the "issue" comments, and the PR you are talking about is closed and discarded at some future point, the "issue" comments become irrelevant and possibly confusing to future developers wanting to try their own solution._

So I said [above](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9552#issuecomment-783430862) that "Correctness was favored over performance".
Guy has shown one example of where a replacement using _Replace All_ button versus using this PR's changed _Replace_ button behavior produces different results. Although Guy seemed to downplay that difference, to me it is important.
Are we to trade one bad behavior (`\K` use causing _Replace_ to mostly not work) for a behavior that is perhaps better but still shown to be incorrect for certain data?

mere-human:
> I expected to see "mark as ready" or something like that where the link text for "mark as draft" was, but nope. So I'm not sure what to do when it is finally ready now. :-) But, I suppose I will figure that out when/if the time comes. :-)

Ok, there should be a button "Ready for review" in the bottom of your PR.
My information wasn't exact because I didn't remember it.
FYI: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request

Uhf7:
If Line 2378 of `ScintillaEditView.cpp` would be

	fp = fa + ((fa > fp) ? 0 : 1);

instead of 

	fp = fa + 1;

then all samples from @guy038 would work.

sasumner:
>then all samples from @guy038 would work.

Indeed.  Let's give it a try.

Uhf7:
OK, I repeated the tests with the samples introduced by @guy038, using a hand-crafted macro which I integrated in my `shortcuts.xml` file: 
[9552_Guy038_Tests.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6217034/9552_Guy038_Tests.txt)

It is intended to be run in an empty text buffer. It produces two lines of output for each sample, the first line is the result of the single-step replace, the second line is the result of the Replace All operation.

The official version 7.9.5 shows this results:
```
A:
123@456@0@789@0
123@456@0@789@0
B:
123@56@@89@
123@56@@89@
C:
12@45@@78@0
12@45@@78@0
D:
123abc456abc0abc789abc0
123@456@0@789@0
E:
123abc456abc0abc789abc0
12@45@0abc78@0
F:
123abc456abc0abc789abc0
12@45@0@78@0
G:
123@456@0abc789abc0
12@45@0@78@0
H:
12@45@@78@0
12@45@@78@0
I:
123abc456abc0abc789abc0
123@456abc0abc789@0
J:
abcdefxyzdefabcdefxyz
abc@xyzdefabc@xyz
```

The artifact of this PR shows this results:
```
A:
123@456@0@789@0
123@456@0@789@0
B:
123@56@@89@
123@56@@89@
C:
12@45@@78@0
12@45@@78@0
D:
123@456@0@789@0
123@456@0@789@0
E:
12@45@0abc78@0
12@45@0abc78@0
F:
12@45@0@78@0
12@45@0@78@0
G:
12@45@0@78@0
12@45@0@78@0
H:
12@45@@78@0
12@45@@78@0
I:
123@456abc0abc789@0
123@456abc0abc789@0
J:
abc@xyzdefabc@xyz
abc@xyzdefabc@xyz
```
Much better.

I'm still not happy with searching from the start of the file every time. It's not only an issue of performance, it can also happen, that a replace operation, which would work with a limited reach-back window, doesn't work, because the file contains "indigestible" content at the start. So, if the file contains something at the start, which overcharges the RegEx engine, then the single-step-replace does not work for the entire file. 

A kind of compromise solution, I could imagine, is to do one or more iterative steps. The first iteration starts 2048 bytes before current pos. In most real world cases, this will be enough. There  is no danger of stack overflows in the RegEx engine with this short window, and the result will not be compromised by file content 10 MB before the current position. If the first iteration fails, then, for correctness, a second iteration could start from the beginning of the file.

PR #9566, opened 2021-02-23
Catch regex_search exceptions

### LINKED ISSUE #9565: Failed regular expression searches can mark whole file as search result
### Description of the Issue
If a regular expression search fails because of an internal stack overflow or a similar exception, the whole file is marked as search result.

1. Open the text file contained in this zip: [BackslashK-Testfile.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6030665/BackslashK-Testfile.zip)
2. Search for the regular expression in line 2 of the test file: `abc(.|\R)*?\Kdef`, easiest with wrap around option on.
3. Repeat the search, easiest with F3.

### Expected Behavior
- All occurrences of `def`, which are preceded by an `abc`, are found step-by-step.
- _Edit 1:_ If the search fails, then it should not select the whole file.

### Actual Behavior
- _Edit 2:_ Not all occurrences of `def`, which are preceded by an `abc`, are found step-by-step.
- The `def` occurrences are found only as long as the distance between `abc` and `def` is shorter than 22000 to 64000 characters. An exact number cannot be given, because the internal error which occurs here is a stack overflow.
- If the distance between `abc` and `def` is too long, then the whole file is selected as search result. _Edit_ 3: This is dangerous, especially for replace operations.

### Debug Information
Notepad++ v7.9.3   (32-bit)
Build time : Feb 14 2021 - 03:12:31
Path : C:\Tools\Npp\Npp_793\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1607
OS Build : 14393.576
Current ANSI codepage : 1252

Uhf7:
Fixes #9565.


Uhf7:
There have been critics on this PR from major developers, so that I'm going to revise it. The new goal is to display the original error text of the exception in the status line of the Find dialog. In the case of #.9565 this would be "Ran out of stack space trying to match the regular expression". Naturally, this will take some days. Please leave the PR open during this time.

Uhf7:
OK, it was easier than I thought. This PR does two things now:
- Avoid selecting the whole file if the regex expression is correct but the regex search itself fails. This is due to the catch statements added.
- Propagate the error text of the regex-search to Notepad++

I tested the 32-bit version only until now. To test it, the SciLexer.dll has to be compiled with the boost regex extensions.

sasumner:
@Uhf7 

>I tested the 32-bit version only until now.

Did you every try the 64-bit?

I'm not really set up for 32-bit building (don't have Boost at hand).
I was going to test this, so I attempted to build SciLexer.dll (64-bit) with VS2019.  I first built the code without this PR's change to make sure all was fine, and it was.  The I applied the BoostRegexSearch.cxx/.h changes and it failed to build:

```
BoostRegexSearch.cxx
../boostregex/BoostRegexSearch.cxx(257): error C2664: 'int wsprintfA(LPSTR,LPCSTR,...)': cannot convert argument 2 from 'const wchar_t [32]' to 'LPCSTR'
../boostregex/BoostRegexSearch.cxx(257): note: Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast
C:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um\winuser.h(326): note: see declaration of 'wsprintfA'
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.24.28314\bin\HostX64\x64\cl.EXE"' : return code '0x2'
Stop.
```

Is it me?  Any ideas?

mere-human:
> I applied the BoostRegexSearch.cxx/.h changes and it failed to build:

It looks like it tries to build non-Unicode version.

sasumner:
>It looks like it tries to build non-Unicode version.

I suppose it is possible.
Is the code that this PR adds the only code in the file where this would matter? (off to go look...)

Uhf7:
Until 45 minutes ago, I never compiled a 64 bit version with boost regex. But now I tried it and it worked for me.

I use MS VC 2017. The `SciLexer.vcxproj` is based on that from the repository, the setting "General\Character Set" is "Use Unicode Character Set". So my `wsprintf` expands to `wsprintfW`.

mere-human:
> Is the code that this PR adds the only code in the file where this would matter?

What command line do you use for build?
I haven't tried this PR yet. I can do this later.
Take a look at this: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/BUILD.md#build-scilexerdll-with-boost
Ensure you are using the correct Developer Command Prompt.

Uhf7:
Now I got it. The mistake is on my side. I used `L"xxx"` instead of `TEXT("xxx")` in `BoostRegexSearch.h`. With `TEXT` instead of `L` it compiles with both character set settings.

sasumner:
@mere-human said

>Take a look at this: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/BUILD.md#build-scilexerdll-with-boost
Ensure you are using the correct Developer Command Prompt.

Seriously?  You're really telling me this?  :-) :-)

@Uhf7 said

>It has succeeded in Appveyor, because this (still) builds without Boost.

Ah, yes.  A constant source of confusion.  Hopefully to be changed soon.

scintilla/boostregex/BoostRegExSearch.cxx
```diff
@@ -251,6 +251,17 @@ RegexSearchBase *CreateRegexSearch(CharClassify* /* charClassTable */)
 #endif
 }
 
+void propagateErrorMessage(char const *msg)
```
mere-human, line 254: Maybe, just use global string buffer instead of this cumbersome method for message propagation?

mere-human (commented):
Just a minor suggestion.
Otherwise, i tested it and it works.
image

scintilla/boostregex/BoostRegExSearch.cxx
Uhf7: There is a global string buffer? I was not aware of this. But if it's easy, I'm open to it. Where is the API?

mere-human: @Uhf7 May I push to your PR? You can revert my commit later if you don't like it.

sasumner:
@Uhf7 As you continue to work this PR, is it possible to bring in the change from https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d6c6a49d1b2f763c13c77b865a9693e9d912cc37 so that others, for example, @guy038 , can test this?

I tested it.  

In @mere-human 's screenshot above we see the error message on the status bar.  Typically such status bar messages begin with `Find:` or `Count:` and then proceed to say what they're going to.

It works great for a `Find Next` button press.  But for a `Count` press or a `Find All in Current Document`, and probably others I didn't try, there is no indication that a problem was encountered, only partial results are presented.
I presume you are going to continue working on such aspects?

mere-human:
Also, now we get **all** the exceptions from Boost.Regex for better or for worse.
For example, see:
image

The same on master branch:
image

Note that this message is without localization. Will it make sense to hide them under a preference?

scintilla/boostregex/BoostRegExSearch.cxx
mere-human: Here is the idea: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3d08ed23302ec544aac5c015c900e980c2ffb90d
You can get this commit to your PR if you like it (git cherry-pick, git apply diff, etc).

Uhf7:
@sasumner,

> It possible to bring in the change from d6c6a49

OK, I'm going to rebase it.

> But for a Count press or a Find All in Current Document, and probably others I didn't try, there is no indication that a problem was encountered, only partial results are presented.

The first intention here is to prevent to select the whole file at once, if the search fails for technical reasons, but the regex itself is ok.

The second intention is to transport the error message.

The third intention is to display the error message. I find it more informative to see what is wrong with my Regex expression in English than to read "Find: Invalid regular expression" in localized language. But of course, this is my "take the easiest way" approach. Another approach would be:
- Let SciLexer.dll provide three return codes: (-1 for not found, -2 for regex wrong and -3 for search failed). Then -1 and -2 could be displayed as localized umbrella messages (-2: "Find: Invalid regular expression") and (-3: "Find: Regular expression matching failed").
- Put an extra button in the status line to show the not localized detailed error message for experts.

The forth intention is to improve the other functions, but this should be done in another PR then.

scintilla/boostregex/BoostRegExSearch.cxx
Uhf7: Ah, now I get it. You want to export a new function from SciLexer.dll. Seems to be the better way here, technically. Thank you. I look into it.

mere-human:
> * Let SciLexer.dll provide three return codes: (-1 for not found, -2 for regex wrong and -3 for search failed). Then -1 and -2 could be displayed as localized umbrella messages

This is a good idea! 
If we care only about the type of the error, perhaps, the return code is better than an error message.

If you need to distinguish between different error codes, you might also return exception's error code:
https://www.boost.org/doc/libs/1_34_0/libs/regex/doc/error_type.html
To avoid confusion with other values, it might be returned as a negative value with offset if needed:
`return -(100 + ex.code());`

Uhf7:
>  perhaps, the return code is better than an error message

I want the original messages. They can contain more information than the error code itself. In case of wrong regular expressions, this is, where the parsing error happened.

The original messages are quite technical. Which implies, that it is not easy to localize them without loss of information. That's why I suggested to localize the general error messages only (-1, -2, -3, see previous post), and provide the original error text as an addition for experts.

And I guess, three different negative return codes are easier to handle than an arbitrary negative error number.

Uhf7:
The detailed message could be displayed as tooltip. This would look like this:

FindTooltip

sasumner:
Handling invalid regular expression messages seems to go outside the scope of what this PR intends?
I leave it to you to set your own scope, as it may be more convenient to wrap this all into it.

My idea, in the context of #6043, was to analyze the regex in the _Find what_ box, as the user types, and turn the box red if the content is an invalid regex.  This avoids the user pressing a button to start a search and only THEN realizing the regex is bad.  I can see the tooltip on the status bar above, originating from a red _Find what_ box in this situation.
Just ideas...

Uhf7:
@sasumner 

> My idea ... was to analyze the regex in the Find what box, as the user types, and turn the box red if the content is an invalid regex.

This is really out of the scope of this PR, but this PR will neither hinder nor replace such a function in the future. This PR deals only with the result of the `searchInTarget` function. Before this PR, -1 was returned for "not found", -2 was returned for "invalid regular expression",  and 0 was returned for "search failed". After this PR, -3 is returned for search failed. Since -3 is was not recognized before this PR, I was forced to introduce this at some places.

sasumner:
>This is really out of the scope of this PR

Sure.  I wasn't trying to adjust any scope of _this_ PR; as I said you are free to set your own.
I'm just commenting where some future development is semi-related, as it may be helpful to know.

donho:
@Uhf7 It seems still in iteration - please let me know when PR is ready.

mere-human:
> It seems still in iteration - please let me know when PR is ready.

By the way, "set PR as draft" can be used for that.

Uhf7:
@donho,

this was it from my side. Instead of only catching all exceptions, this PR includes now 3 localized general error messages in the status bar of the Find dialog:
- "Find: Can't find the text" (was there before)
- "Find: Invalid regular expression" (was there before)
- "Find: Search failed"  (new).
 
The latter two general errors are refined by a tooltip for the Find dialog status bar with the original regex error text.

What I needed to do was to react to the `searchInTarget` result of `-3` everywhere, so I replaced all `if (res == -1 || res == -2)` conditions by `if (res < 0)` and all `if (res != -1 && res != -2)` by `if (res >= 0)`. I guess, this is the easiest way to handle it, since buffer sizes above `INT_MAX` bytes aren't allowed anyway.

I'm aware that "Count", "Find All" and "Replace All" functions treat the result of `-2` and `-3` as "not found", but this is no regression to the current behavior and should be fixed with another PR.

Uhf7:
@mere-human,

> By the way, "set PR as draft" can be used for that.

Where exactly do I click to do this?

mere-human:
> Where exactly do I click to do this?

@Uhf7 
image

Uhf7:
@mere-human,
now I see it. Wasn't there before, wasn't it?
Is the way back similar easy?

mere-human:
> Is the way back similar easy?

Yes, something like "Mark as ready" in the same place.

Uhf7:
@mere-human,
cool, I will try it as soon as appropriate.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -340,6 +340,8 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 	}
 	//Get the startup document and make a buffer for it so it can be accessed like a file
 	attachDefaultDoc();
+
+	_pGetExceptionMsgFunc = reinterpret_cast<GetExceptionMessageFunc>(::GetProcAddress(_hLib, "GetBoostExceptionMessage"));
```
donho, line 344: With the [fusion of notepad++.exe & SciLexer.dll](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9594) (coming soon), there won't be `_hLib` to get process' address. What'll be the solution?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1795,11 +1795,39 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
...
-		generic_string msg = pNativeSpeaker->getLocalizedStrFromID("find-status-invalid-re", TEXT("Find: Invalid regular expression"));
-		setStatusbarMessage(msg, FSNotFound);
+		generic_string  msgGeneral;
+		if (posFind == -2)
+		{
+			 msgGeneral = pNativeSpeaker->getLocalizedStrFromID("find-status-invalid-re", TEXT("Find: Invalid regular expression"));
+		}
+		else
+		{
+			msgGeneral = pNativeSpeaker->getLocalizedStrFromID("find-status-search-failed", TEXT("Find: Search failed"));
```
donho, line 1808: tooltip is a very good way to provide the exception message. It would be even better to let user know he/she can get more info for the failure of search.
I would put "Find: Search failed - move the mouse here to get more info" instead of "Find: Search failed".

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1324,6 +1324,7 @@ Find in all files except exe, obj &amp;&amp; log:
             <find-status-replaceinopenedfiles-nb-replaced value="Replace in Opened Files: $INT_REPLACE$ occurrences were replaced"/>
             <find-status-mark-re-malformed value="Mark: The regular expression to search is malformed"/>
             <find-status-invalid-re value="Find: Invalid regular expression"/>
+            <find-status-search-failed value="Find: Search failed"/>
```
donho, line 1327: Reword the message.

donho (changes requested):
1. Consider the situation in the future to link Scintilla statically.
2. Consider to reword the search failure message to let user know how to get the detail info.

sasumner:
@mere-human showed a screenshot where the status bar reports:  `Ran out of stack space trying to match the regular expression`.  Is this text coming from Boost, or is it something made up?  If not coming from Boost, is there a better way to express it?  I'm trying to think like a user that doesn't know or care about "stack space" and is simply wondering what went wrong and what to do next.  Hopefully what he thinks to do/try next is to NOT put in a support ticket.

Uhf7:
> Ran out of stack space trying to match the regular expression

This text is the original message coming from Boost. The current plan is, not to translate this. That's why I took the effort to display translated general messages in the status line and provide the original messages as tooltip.

sasumner:
>Find: Search failed

My first reaction to this was that user might just key in on the "search failed" and then start looking to their data to see why a search that they thought should have succeeded did not, not suspecting a badly-formed regex or a regex-engine problem.

>Find: Search failed - move the mouse here to get more info

So I see @donho requested this, which I suppose is better.
Translators will translate the "move the mouse here to get more info" part, but when the user does that what pops up will be English text.  Ironic, but nothing to do about it.

guy038:
Hello @uhf7, @sasumner, @mere-human, @don-ho,

From the list of **Boost** regex errors, provided by @mere-human, below, I noticed an **small** error ! You're going to say I'm a bit **pretentious**.!

https://www.boost.org/doc/libs/1_34_0/libs/regex/doc/error_type.html

But, indeed, the **message** of the **`error_badrepeat`** seems inaccurate. In fact, not the message itself but the provided **example** !

Please, consider, however, my **post** as a classical **non-urgent** message !

---

Back to the problem, it states that the **`a*+`** regex is an **illegal** syntax . That's **not** true : it's the letter **`a`** followed with the **possessive** quantifier **`*+`**

Just consider the **two** regexes, below, against this text **`xyzbaaaaaaaa12---------baaaaaaaa123456789`**, in a **new N++** tab :

- Regex **A** : **`ba*\w{4}`**

- Regex **B** : **`ba*+\w{4}`**

As you can see, the regex **`A`** matches **two** strings and regex **`B`** matches just **one** range of **word** chars. I'm not going to explain, **here**, why the regex **`B`**, with **possessive** quantifier, is more **restrictive**. But this **clearly** shows that, anyway, the **`a*+`** syntax is a **valid** one ;-))

---

Conversely, the regex **`a+*`** and the similar **`ba+*\w{4}`** are **illegal** syntaxes, which, **presently**, display the message **`Find: Invalid regular expression`**  !

Have **nice coding** time, all !

Best Regards,

guy038

mere-human:
> @mere-human showed a screenshot where the status bar reports: `Ran out of stack space trying to match the regular expression`. Is this text coming from Boost, or is it something made up?

This is a text from the Boost exception.

> it states that the **`a*+`** regex is an **illegal** syntax . That's **not** true : it's the letter **`a`** followed with the **possessive** quantifier **`*+`**

Yes, probably not the best example. But this is a question to the Boost authors.

sasumner:
@guy038 said:

> it states that the a*+ regex is an illegal syntax . That's not true : it's the letter a followed with the possessive quantifier *+

FWIW, Regex buddy doesn't like it, not sure if it is making its own judgement or interpreting something the Boost engine is returning:

image

guy038:
Hello, @sasumner,

But I **do** agree with **`RegexBuddy`**, as well !!

As I said at the **end** of my **previous** post

> Conversely, the regex **`a+*`** and the similar **`ba+*\w{4}`** are **illegal** syntaxes, which do **not** work in N++, too

But the regexes **`a*+`** and the similar **`ba*+\w{4}`** are **correct** regexes : it's a **greedy** quantifier **`*`** changed as **atomic**, due to the trailing **`+`** symbol !

**Scott**, just try the regex **`a*+`** in **`RegexBuddy`** It should work nicely !

BR

guy038

sasumner:
>just try the regex a*+ in RegexBuddy It should work nicely !

I guess I was having a dyslexic day today, messing up my ordering of `*+` and `+*`.  :-(
Sorry for the "noise".

donho:
@sasumner 
> Translators will translate the "move the mouse here to get more info" part, but when the user does that what pops up will be English text. Ironic, but nothing to do about it.

The non-translated text from boost provides more detailed info. If user can not read it, he/she can always translate it on Internet.
So I don't feel ironic way. It's like sex to me - even it's bad, it's better than nothing.

donho:
@Uhf7 The 2 points of my request are optional - your reply is appreciated.

Uhf7:
I'm actually waiting for the statically linked version. Then I can replace the `GetBoostExceptionMessage` call by some sort of symbol exported by the lib pointing to the "global string buffer", and then I can request the next review.

sasumner:
>I'm actually waiting for the statically linked version. Then I can replace the GetBoostExceptionMessage call by some sort of symbol exported by the lib pointing to the "global string buffer", and then I can request the next review.

Maybe that's a good reason to "convert to draft" this PR , until such a time ?  :-)

Uhf7:
> Maybe that's a good reason to "convert to draft" this PR.

No, I don't think so. First, it's suspended now anyway, and second, it was not supposed to be an experimental PR made for others to test. I really want it.

I'm considering a way now to make it work for both - static linking and dynamic loading.

sasumner:
@Uhf7 

In my mind, "draft" status is like your own "suspension".  
It wouldn't have anything to do with "really wanting it" or not.
It would keep it out of @donho 's sight until it was really ready.  You said yourself it wasn't ready because you were waiting for the statically linked changes to become final -- at which point I presume you rebase this to get them.
Only at that point would you then take it out of "draft" status.
But perhaps I misunderstand because I am new to making draft PRs.

>I'm considering a way now to make it work for both - static linking and dynamic loading.

Is there value in dynamic if static is the future?

Uhf7:
> It would keep it out of @donho 's sight until it was really ready.

The opposite is what I want. The PR I originally made was perfectly ready. It was no draft, it was what I intended to do. 

Then, critics came in: Oh, error handling is not good enough. So I improved it. I passed the error message.

More critics came in: Oh, no localization. So I improved it. I work with general localized error messages now and display the original message as tooltip.

More critics came in: Oh, the way of passing the value is cumbersome (although my original way worked in both cases, statically linked and dynamically loaded). So I improved it. Actually, @mere-human did this for me.

Now, the rules of game changed, static linking is the new ambition ...

I'm going to make this PR fit for this too, but it is no draft. It's the normal process, in which this PR is reviewed by other developers and in this case, there where some reasonable doubts. I will continue working to clear this doubts. 

But I'm not convinced that this is a reason to convert this PR into a draft. A draft means: "All you developers out there, don't look at it until I'm ready". And that's definitely not what I intend here.

> Is there value in dynamic if static is the future?

Much value. If it works both ways, the PR could be applied immediately.

PowerEditor/installer/nativeLang/english.xml
MrScottAtt: The message should help the user understand and solve the issue.  _Why_ did the search fail?  What should they do about it?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1769,7 +1769,7 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
 		}
 
 		if (posFind == -1)
```
MrScottAtt, line 1771: In the preceding files, posFind was compared to -1 or -2, which was modified to compare to 0.  This compares only to -1.  Is this an intentional comparison to a specific error return code?

MrScottAtt, line 1771: Could be "IsFailNotFound(posFind)", and "IsFailInvalidRegex(posFind)" below.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
MrScottAtt: Any exception message should be copyable in some way.

PowerEditor/installer/nativeLang/english.xml
sasumner: The message _does_ help the user understand what went wrong:

image

As far as what to do about it, well, for the pictured case that may require some good understanding of regex (on how to recraft the regex so that it might be simpler as to not cause the engine any troubles, if possible) but I don't think that is something Notepad++ is going to be able to tell you how to solve.

In other cases, it does pinpoint an invalid regex problem, example:

image

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3394,6 +3412,41 @@ void FindReplaceDlg::drawItem(LPDRAWITEMSTRUCT lpDrawItemStruct)
 	RECT rect;
 	_statusBar.getClientRect(rect);
 	::DrawText(lpDrawItemStruct->hDC, ptStr, lstrlen(ptStr), &rect, DT_SINGLELINE | DT_VCENTER | DT_LEFT);
+
+	if (_statusbarTooltipMsg.length() == 0) return;
+
+	SIZE size;
+	::GetTextExtentPoint32(lpDrawItemStruct->hDC, ptStr, lstrlen(ptStr), &size);
+	int s = (rect.bottom - rect.top) & 0x70; // limit s to available icon sizes and avoid uneven scalings
```
MrScottAtt, line 3420: Please turn magic number 0x70 into a symbolic constant, maybe something like ICON_SIZE_MASK.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1795,11 +1795,22 @@ bool FindReplaceDlg::processFindNext(const TCHAR *txt2find, const FindOption *op
...
 	}
-	else if (posFind == -2) // Invalid Regular expression
-	{
+	else if (posFind < -1) 
+	{ // error
 		NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
-		generic_string msg = pNativeSpeaker->getLocalizedStrFromID("find-status-invalid-re", TEXT("Find: Invalid regular expression"));
-		setStatusbarMessage(msg, FSNotFound);
+		generic_string  msgGeneral;
+		if (posFind == -2)
```
MrScottAtt, line 1802: May be out of scope here, however I would really like to see these -1, -2, etc. magic numbers turned into (meaningfully named) symbolic constants.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
Uhf7: There is no advantage in a one time used constant. Compared to the comment, ICON_SIZE_MASK says nothing, as less as any other name I could invent. It's simply an arithmetic operation, like + 3 or - 5. I wouldn't have the idea to invent constants like FIVE or THREE.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
Uhf7: Replacing this by constants would make indeed more sense, but the goal was not to improve infrastructure. Until now, it's hardcoded -1 and -2 everywhere, and since the good/bad check is now `r >= 0` instead of `(r != -1) && (r >= -2)`, it's improved already.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -991,7 +991,7 @@ int Notepad_plus::getHtmlXmlEncoding(const TCHAR *fileName) const
 		_invisibleEditView.execute(SCI_SETTARGETRANGE, startPos, endPos);
 
 		auto posFound = _invisibleEditView.execute(SCI_SEARCHINTARGET, strlen(xmlHeaderRegExpr), reinterpret_cast<LPARAM>(xmlHeaderRegExpr));
-		if (posFound != -1 && posFound != -2)
+		if (posFound >= 0)
```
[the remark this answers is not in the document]
Uhf7: The evaluation of a single term `r >= 0` is clear enough for me. There is a simple decision: Negative values are errors. To wrap this into a function, is obfuscation. Abstract, abstract, abstract, until nobody knows anymore what actually happens...

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
MrScottAtt: That works for a single-use constant.

This is not the place for an extended discussion, however I will say that I would not ask for constants such as "FIVE" or "THREE"; they wouldn't add _meaning_ vs. 5 or 3.

Code is a shared facility, read 10 to 100 times as often as it is written.  The idea is to communicate your _intent_ to the next developer who is reading and trying to understand the code.  

It also helps keep them from doing a global search-and-replace of 3.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
Uhf7: As indicated here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9566#discussion_r594644310, I could imagine to add three constants to  the `scintilla\include\BoostRegexSearch.h`. Something like

	#define ERR_NOT_FOUND (-1)
	#define ERR_WRONG_REGEX (-2)
	#define ERR_SEARCH_FAILED (-3)

But no extra function for this mega-complex issue to detect which of the three error codes has been delivered.

And, as I mentioned before, I believe in generalized error codes and original regex error messages. Honestly, there is a correlation here: Users, who are able to compose the most complicated regex expressions (I ever saw), are certainly able to interpret the original regex error messages and can get more out of the original messages than out of any translation I could provide.

Uhf7:
@MrScottAtt 
Thank you for the review, although I not agree with your opinion in all points. But constants for -1, -2 and -3 would be good, I agree at least to that.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
Uhf7: Yes, this argumentation works for the negative error numbers, but not for the

`size = size & 0x70`

Here, the comment in the line is hint enough.

PowerEditor/installer/nativeLang/english.xml
donho: @sasumner Brilliant!

donho (approved):

PowerEditor/installer/nativeLang/english.xml
sasumner: @donho LOL, OK...but I didn't do anything but make some nice screenshots! :-)

PR #9574, opened 2021-02-24
Boost appveyor build

Uhf7:
Make boost regex available in Appveyor builds.

There are unnecessary differences between Appveyor builds and Notepad++ release builds:
- The release builds of Notepad++ contain a SciLexer.dll with improved RegEx behavior
- The Appveyor builds do not.

This PR intents to overcome this problem.

_Edit 2:_ Since Appveyor has no 1.70 version of boost installed for the current compiler MS VC 2017, I used boost 1.69 here. I believe, that this is better than no boost at all.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ build_script:
     - cd "%APPVEYOR_BUILD_FOLDER%"\scintilla\win32
     - if "%configuration%"=="Unicode Debug" set scintilla_debug=DEBUG=1
     - if "%configuration%"=="Unicode Release" set scintilla_debug=
-    - if "%archi%" NEQ "" nmake SUPPORT_XP=1 %scintilla_debug% BOOSTPATH=C:\Libraries\boost_1_69_0\ BOOSTREGEXLIBPATH=C:\Libraries\boost_1_70_0\lib32-msvc-14.1\ -f scintilla.mak
+    - if "%archi%" NEQ "" nmake SUPPORT_XP=1 %scintilla_debug% BOOSTPATH=C:\Libraries\boost_1_70_0\ BOOSTREGEXLIBPATH=C:\Libraries\boost_1_70_0\lib32-msvc-14.1\ -f scintilla.mak
     - if "%Platform%"=="mingw-w64_810_X64" mingw32-make -j%NUMBER_OF_PROCESSORS%
     - cd "%APPVEYOR_BUILD_FOLDER%"\PowerEditor\visual.net\
     - if "%archi%" NEQ "" msbuild notepadPlus.vcxproj /p:configuration="%configuration%" /p:platform="%platform_input%" /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ build_script:
     - cd "%APPVEYOR_BUILD_FOLDER%"\scintilla\win32
     - if "%configuration%"=="Unicode Debug" set scintilla_debug=DEBUG=1
     - if "%configuration%"=="Unicode Release" set scintilla_debug=
-    - if "%archi%" NEQ "" nmake SUPPORT_XP=1 %scintilla_debug% BOOSTPATH=C:\Libraries\boost_1_69_0\ BOOSTREGEXLIBPATH=C:\Libraries\boost_1_69_0\lib32-msvc-14.1\ -f scintilla.mak
+    - if "%archi%" NEQ "" nmake SUPPORT_XP=1 %scintilla_debug% BOOSTPATH=C:\Libraries\boost_1_69_0\ BOOSTREGEXLIBPATH=C:\Libraries\boost_1_70_0\lib32-msvc-14.1\ -f scintilla.mak
     - if "%Platform%"=="mingw-w64_810_X64" mingw32-make -j%NUMBER_OF_PROCESSORS%
     - cd "%APPVEYOR_BUILD_FOLDER%"\PowerEditor\visual.net\
     - if "%archi%" NEQ "" msbuild notepadPlus.vcxproj /p:configuration="%configuration%" /p:platform="%platform_input%" /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -33,9 +33,8 @@ build_script:
     - cd "%APPVEYOR_BUILD_FOLDER%"\scintilla\win32
     - if "%configuration%"=="Unicode Debug" set scintilla_debug=DEBUG=1
     - if "%configuration%"=="Unicode Release" set scintilla_debug=
-    - if "%archi%" NEQ "" nmake SUPPORT_XP=1 %scintilla_debug% -f scintilla.mak
+    - if "%archi%" NEQ "" nmake SUPPORT_XP=1 %scintilla_debug% BOOSTPATH=C:\Libraries\boost_1_69_0\ BOOSTREGEXLIBPATH=C:\Libraries\boost_1_69_0\lib32-msvc-14.1\ -f scintilla.mak
```
mere-human, line 36: I guess you need to download the boost first. Maybe, use a package from NuGet or some other package manager?
Also, it could be checked out from git repo and built. However, that might slowdown AppVeyor build significantly.

Uhf7: No, I don't. Everything is pre-installed. I'm trying to check out where exactly right now.

Uhf7: It will not slow down the Appveyor build.

 What I found out yet:
- The Appveyor doc is correct. For MS2017, only boost 1.69 is available
- Boost 1.69 works for the 32-bit version.

Next:
- Make it work for the 64-bit version.

Uhf7: Test 3 revealed, that Boost 1.70 is not available for MS VC 2017.

Uhf7:
OK, after Test 4 I know, that Boost 1.70 is really not available.

chcg:
You could see the installed boost images for appveyor here:
https://www.appveyor.com/docs/windows-images-software/#boost

To my knowledge that are just the sourcecode versions of boost. So just header only components could be used directly. Regex is not of that kind and need to be build, see e.g. https://www.nuget.org/packages/boost_regex-vc141/ for a nuget package.
With some older version such a packages was also part of the scintilla build via the vcxproj file, see https://github.com/notepad-plus-plus/notepad-plus-plus/blob/v7.4/scintilla/win32/packages.config

Uhf7:
@chcg,
I looked exactly at this page (https://www.appveyor.com/docs/windows-images-software/#boost), my problem here is, that boost 1.69 is pre-installed for MS VC2017, but boost 1.70 (which is used by the Notepad++ SciLexer version) is not. So, what I'm currently trying, is to use 1.69 to create the SciLexer.dll with boost. If this works, we have to evaluate what is bad in 169 compared to 170. If there are no magnificent improvements of 170 compared to 169, then I would say, let's use 169 with Appveyor builds, it's better than no boost regex at all ...

mere-human:
Maybe, it is about time to upgrade Notepad++ to be built with MSVC 2019.
I've wanted that for some time🙂

ArkadiuszMichalski:
There were some changes: 
https://www.boost.org/users/history/version_1_70_0.html

Uhf7:
@ArkadiuszMichalski,

too many lines in that, I cannot read it today, and, honestly, I'm not going to read it tomorrow. If you have a certain point in this list, which is important to you, please name it.

The question here is, do we believe, that a 1.69 regex boosted version of SciLexer.dll will allow users to test certain issues closer to reality than with a not-boosted SciLexer.dll version. I believe, that version 1.69 is good enough...

And, of course, @mere-human, if there would be an administrative decision to use MS VC 2019 from now on, then we could use the 1.73 version of boost regex with no Appveyor constraints.

sasumner:
@ArkadiuszMichalski 

>There were some changes:
https://www.boost.org/users/history/version_1_70_0.html

Curious, that history says nothing about the regex part of the library -- does that mean there were no changes from 1.69 to 1.70?  Hmm, looking back a few links 1_69_0, 1_68_0, 1_67_0, no mention of regex there either.  I guess I don't know how to read Boost history docs correctly?

@mere-human 

>Maybe, it is about time to upgrade Notepad++ to be built with MSVC 2019.
I've wanted that for some time

My secret:  I've NEVER had MSVC 2017; always used MSVC 2019 on N++...with the slight inconvenience doing that causes.

@Uhf7 

> I believe, that version 1.69 is good enough...

I tend to agree.

appveyor.yml
donho: `SUPPORT_XP=1` is not necessary anymore.

Uhf7:
@ArkadiuszMichalski,

sorry for complaining about the long list, this is of course the thing which should be checked. Today, it looks much shorter and it seems indeed not to contain anything about regex.

I compared the regex source code of versions 1.69 and 1.70 too, there is no more difference than a new `#define BOOST_REGEX_NO_EXTERNAL_TEMPLATES` for Cygwin systems in `config.hpp`. All other regex files are the same. I didn't check all the dependencies within boost, but there seems no intentional change of regex behavior between this two versions.

ArkadiuszMichalski:
Unfortunately, Notepad ++ still lacks regex tests, it would be a lot easier with them. Function List test passed on this version? Better this than nothing. 
One more question for formalities. Will build (appveyor) .dll and release .dll use the same or different version of Boost in future? If Boost doesn't actually change that much from version to version, then maybe different versions won't be a reason for unnecessary bug reports.

sasumner:
Now that @donho has joined this thread...  The last time this kind of thing came up, Don mentioned he would "look into static linking" of Scintilla and thus by inference Boost.

donho:
@ArkadiuszMichalski 
>  Function List test passed on this version? 

Yes, all tests passed with new ScinLexer.dll v4.4.6.

> Unfortunately, Notepad ++ still lacks regex tests, it would be a lot easier with them.

We can add regex test via Function List by using different user defined language.

@sasumner 
> Don mentioned he would "look into static linking" of Scintilla and thus by inference Boost.

Yes sure. I will check it after releasing of next version.

sasumner:
So the summary is, now that this is accepted, that future appveyor builds of SciLexer will contain Boost 1.69.

`SUPPORT_XP` was retained, is that relevant, or was a good opportunity to remove it missed?

donho:
@sasumner 
> So the summary is, now that this is accepted, that future appveyor builds of SciLexer will contain Boost 1.69.

Correct.

> `SUPPORT_XP` was retained, is that relevant, or was a good opportunity to remove it missed?

I create a new issue here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9581 and assign to @chcg.
It could be done in this PR.

PR #9596, opened 2021-02-28
Update german.xml

schnurlos:
Following last changes of english.xml - 1


donho:
Thank you for the update. Could you adapt the diff of english.xml (v7.9.4 vs v7.9.3):
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9613

and confirm me when it's done? Thank you.

Issue #9597, opened 2021-03-02
Notepad++ has problems with display of chars in UTF8 and UTF16-LE


HarKe298:
### Description of the Issue
I have export files which displayed as "UTF-16 LE" in Windows-Notepad.
This program show all chars correctly but I need the functionality of line sorting from Notepad++.
NPP show wrong sign for e.g.: "⁇�"
NPP show the file encoding "UCS-2 LE". This seams to be ok.

![npp-encoding-error](https://user-images.githubusercontent.com/79909084/109637855-9f492000-7b4d-11eb-9196-6e02d55dc872.jpg)

I canot use NPP for this file with this encoding.
What I have to do to work with my export files with this encoding?

### Steps to Reproduce the Issue

1. Open the file: 
[⁇🅟-Language.csv.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6068065/-Language.csv.txt)
with Notepad and with NPP and compare the content visually.

2. 
3. 

### Expected Behavior

The chars should show correctly and save with minimal changings should create a valid usable file with neccessary chars.

### Actual Behavior

Work with this file will destroy the content of this file.

### Debug Information

Notepad++ v7.9.1   (32-bit)
Build time : Nov  2 2020 - 01:03:56
Path : C:\@\PortableApps\Notepad++Portable\App\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.1379
Current ANSI codepage : 1252
Plugins : AutoSave.dll BigFiles.dll ComparePlugin.dll DSpellCheck.dll HexEditor.dll HTMLTag_unicode.dll JsMapParser.NppPlugin.dll linefilter2.dll Linefilter3.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll nppplugin_svn.dll NppXmlTreeviewPlugin.dll urlPlugin.dll VisualStudioLineCopy.dll XMLTools.dll XPatherizerNPP.dll

ArkadiuszMichalski:
@donho 
I get a crash when try open above `.txt` file and last NPP with new Scintilla:

![image](https://user-images.githubusercontent.com/2730894/109641422-f9e47b00-7b51-11eb-9adb-be75d92f77bd.png)

@HarKe298 

What if you try with this option?

[...]

PR #9599, opened 2021-03-03
Fix UTF-16 decoding/encoding for code points above U+0FFFF.

Uhf7:
The problem addressed with this PR is, that the Windows functions `MultiByteToWideChar` and `WideCharToMultiByte` support code points between U+10000 and U+10FFFF, but the internal conversion routines of Notepad++ do not. This leads to problems like described in #9597.

How to test it? Download the following test files:
[Unicode-Testfiles.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6075709/Unicode-Testfiles.zip)

The Utf-8 test file works with and without this PR.
The Utf-16 test file shows control characters for the code points above U+FFFF without this PR, and the corresponding Unicode characters with this PR.

Of course, most Unicode characters are not present in the fonts used and are still displayed as square. But now there should be no difference anymore to the Windows-driven Utf-8/Utf-16 conversion.


ArkadiuszMichalski:
How does this relate to the UCS-2 shown below? UCS-2 use only 2 byte (cover only BMP plane). Will a new entry be added for UTF-16 or will UCS-2 be changed to UTF-16? 

image

Uhf7:
The menu entries could be changed to UTF-16 too. Although I worry less about the naming here.

It's better to say UCS-2 and then to support UTF-16 instead, than to say UTF-16 and then to support UCS-2 only. And that's, how it is right now: An UTF-16 byte order mark is read from the file, an the file is handled as UCS-2. An UTF-16 byte order mark is written to the file, but only UCS-2 code points work. So, I think, forget UCS-2.

If this PR would be accepted, and there would be a change request like "replace the text UCS-2 in the menu entries with UTF-16", I would be happy to do this.

sasumner:
@Uhf7

>If this PR would be accepted, and there would be a change request like "replace the text UCS-2 in the menu entries with UTF-16", I would be happy to do this.

If this happens, please also change the naming that is used internally as well, for example, the `UCS_2` in the following:

    #define    IDM_FORMAT_UCS_2BE           (IDM_FORMAT + 6)
    #define    IDM_FORMAT_UCS_2LE           (IDM_FORMAT + 7)
    #define    IDM_FORMAT_CONV2_UCS_2BE     (IDM_FORMAT + 12)
    #define    IDM_FORMAT_CONV2_UCS_2LE     (IDM_FORMAT + 13)

Perhaps this is a rather obvious thing, but I've seen before where this is NOT done, and it just makes things more confusing in the future.

donho (changes requested):
Change all UCS-2 in project to UTF-16.

donho:
@Uhf7 
My request `Change all UCS-2 in project to UTF-16.` is due to what you said:

> If this PR would be accepted, and there would be a change request like "replace the text UCS-2 in the menu entries with UTF-16", I would be happy to do this.

and also the suggest of @sasumner : 

> If this happens, please also change the naming that is used internally as well, for example, the `UCS_2` in the following:

    #define    IDM_FORMAT_UCS_2BE           (IDM_FORMAT + 6)
    #define    IDM_FORMAT_UCS_2LE           (IDM_FORMAT + 7)
    #define    IDM_FORMAT_CONV2_UCS_2BE     (IDM_FORMAT + 12)
    #define    IDM_FORMAT_CONV2_UCS_2LE     (IDM_FORMAT + 13)

Do you mind to modify the 2 points above?

donho (approved):

sasumner:
I created a test file for UTF16 with intended content of `abc⁇` where the last character is https://www.compart.com/en/unicode/U+2047

When I pull this file into N++ 7.9.5, the file content appears to be as intended:

image

However, when using the release exe from this PR, pulled into a 7.9.5 portable testbed, I get:

image

When using the debug exe, I get this:

image

Clicking `Ignore` numerous times results finally results in a view of the file as shown above with the `xE2` in it.

Can anyone duplicate this behavior?

Test file is attached:
[UTF16BEBOM.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6365563/UTF16BEBOM.txt)

ArkadiuszMichalski:
For portable (last PR) I get same results as above (for release and debug).

Uhf7:
You got me. Thank you for testing. The conversion while reading the file works only, if the file has a 7-bit ASCII character at the end. And all my tests worked only, because my test files ended with CR/LF.
I just fixed it. PR comes tomorrow, I better test twice.

sasumner:
>Thank you for testing

I wasn't specifically testing this PR.
The data I picked was just something random I quickly made in order to do some other dev work.
I'm glad that it exposed a problem!

PR #9607, opened 2021-03-06
Add Cloud Config and Command Line Arguments to Debug Info

### LINKED ISSUE #9598: add to "Debug Info"
Would you be willing to add two lines to the Debug Info?
1. `Cloud Config: OFF` or `Cloud Config: "c:\path\to\cloud\folder"`, depending on the cloud settings
2. if `-settingsDir` is set, show `SettingsDir Option:` as either `OFF` or `"c:\path\to\settings.dir\"`

Similar to `Local Conf` and `Admin Mode`, those would help us recognize right away whether someone who has a problem with config settings is just not looking at the files in the right folder.

(While answering https://community.notepad-plus-plus.org/topic/20821, trying to figure out why settings weren't getting saved, I was about to ask whether the user had Cloud enabled... though I stopped, because it doesn't look like he's using cloud, and I didn't want to confuse him further.  But if Cloud and now `-settingsDir` information were always available in Debug Info, it would eliminate us having to ask that question. )

mere-human:
Fix #9598

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -1504,6 +1504,9 @@ public:
 	}
 	const CmdLineParamsDTO & getCmdLineParams() const {return _cmdLineParams;};
 
+	const generic_string& getCmdLineString() const { return _cmdLineString; }
+	void setCmdLineString(const generic_string& str) { _cmdLineString = str; }
+
 	void setFileSaveDlgFilterIndex(int ln) {_fileSaveDlgFilterIndex = ln;};
 	int getFileSaveDlgFilterIndex() const {return _fileSaveDlgFilterIndex;};
 
@@ -1731,6 +1734,7 @@ private:
 	int _nbExternalLang = 0;
 
 	CmdLineParamsDTO _cmdLineParams;
+	generic_string _cmdLineString;
 
 	int _fileSaveDlgFilterIndex = -1;
 
--- a/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
+++ b/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
@@ -143,6 +143,11 @@ INT_PTR CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM /
 			_debugInfoStr += nppFullPath;
 			_debugInfoStr += TEXT("\r\n");
 
+			// Command line as specified for program launch
+			_debugInfoStr += TEXT("Command Line : ");
+			_debugInfoStr += nppParam.getCmdLineString();
+			_debugInfoStr += TEXT("\r\n");
+
 			// Administrator mode
 			_debugInfoStr += TEXT("Admin mode : ");
 			_debugInfoStr += (_isAdmin ? TEXT("ON") : TEXT("OFF"));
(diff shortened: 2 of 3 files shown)


mere-human:
Now the Debug Info dialog looks like this:
image

mere-human:
**Question:** If a user changes the _Cloud & Link_ preference and then opens the _Debug Info_ dialog, the new changes won't be shown there.
It is not hard to fix that. Should I do this in current PR?

sasumner:
>Question: If a user changes the Cloud & Link preference and then opens the Debug Info dialog, the new changes won't be shown there.
It is not hard to fix that. Should I do this in current PR?

So if you "fix" it:  User changes cloud pref setting then opens Debug Info and copies data and then reports a bug...wouldn't then we be looking at a setting that is not fully in effect for the current run, and it potentially being misleading? 

Other thoughts:

The original issue asked for settingsDir.  I suppose showing the command line represents this.  Is it possible for settingsDir to expand command-line vars, e.g. %USER% ?  If so, it might be nice to see the exact command line in the Command Line section, and the expanded result under a separate line item.

In the screenshot you show Cloud Config : OFF.  I presume if it is not off, it shows that path here, as requested in the issue.

The command line can be quite long e.g. multiple files are specified for opening.  How does this look in your PR code?  My thinking here is that it might be nicer to show the command line arguments at the bottom of the debug info, so that a long command line isn't "disruptive" to look at, right in the middle of things.

If I have a currently-running Notepad++ and I get a command-prompt and I type `notepad++ anotherfile.txt` then my anotherfile.txt file is opened in the running instance.  My most-recent command line arguments would then be `anotherfile.txt`.  Would this affect what is shown in the Debug Info?

I'm not trying to overcomplicate things, and maybe some things I've said aren't important, I'm just having random thoughts as I ponder this.  :-)

mere-human:
> User changes cloud pref setting then opens Debug Info and copies data and then reports a bug...wouldn't then we be looking at a setting that is not fully in effect for the current run, and it potentially being misleading?

This is also true. 
My original suggestion was because the _Debug Info_ dialog initializes its text only once, when it is shown for the first time. 
If you open it second time it will show the same info as the first time regardless of any changes.
Maybe, this is something that should be addressed separately.

> it might be nice to see the exact command line in the Command Line section, and the expanded result under a separate line item.

If notepad++ is launched from the command line, or from the shell, the OS expands environment variables. So we don't need to do this. Here, I launched it with `-settingsDir="%TMP%"`:
image


> I presume if it is not off, it shows that path here, as requested in the issue.

Sure. Here it is:
image


> The command line can be quite long e.g. multiple files are specified for opening. How does this look in your PR code?

The _Debug Info_ dialog has a scroll bar. So I don't think it is a problem. The plugins list might be long too. Are you sure it should be changed?
Example:
image


> My most-recent command line arguments would then be `anotherfile.txt`. Would this affect what is shown in the Debug Info?

I assume the multi-instance is off as by default. No, it will show the command line arguments of the first launched instance only.

sasumner:
>If you open it second time it will show the same info as the first time regardless of any changes.
Maybe, this is something that should be addressed separately.

Sure, can you create an issue?  I'll assign to you if you want it.

>If notepad++ is launched from the command line, or from the shell, the OS expands environment variables

I had forgotten about this but yes, that makes it a non-issue.

>>The command line can be quite long e.g. multiple files are specified for opening. How does this look in your PR code?

>The Debug Info dialog has a scroll bar. So I don't think it is a problem. The plugins list might be long too. Are you sure it should be changed?

It isn't a problem, per se, I was just sort of concerned with how it would look visually.
As long as all the data is there, I think there is no problem.

>>My most-recent command line arguments would then be anotherfile.txt. Would this affect what is shown in the Debug Info?

>I assume the multi-instance is off as by default. No, it will show the command line arguments of the first launched instance only.

I was wondering how this would work with just a single instance.
It is not a problem that command line would only be shown from the original running.

mere-human:
> Sure, can you create an issue? I'll assign to you if you want it.

The thing is, it can't be observed in the current build from the master branch.
Because all the available information can't be changed without restarting the app.
For example, _Plugins_ - any install/uninstall will trigger restart. _Local Conf_ - detects a `doLocalConf.xml` file during startup.
So I can report the issue as soon as it's observable. That is after this PR is merged this will be sort of observable with _Cloud Config_.

PR #9609, opened 2021-03-06
Fix Project workspace changes lost on save cancel

### LINKED ISSUE #9605: Project workspace changes lost on save cancel
### Description of the Issue

When attempting to quit Notepad++ with a project workspace that has unsaved changes, a prompt appears to ask you if you want to save.  If you say Yes, you are presented with a FileSave dialog.  If you cancel this dialog you will lose the opportunity to save your changes, and Notepad++ exits (but it is nice in that it informs you).

1. Start with no Project panels open and only a single `new 1` tab open that is blue (no mods).
2. Open a Project panel, e.g. `Project 2`, via `View` menu > `Project` > `Project Panel 2` to see:
image
3. Rightclick the word `Workspace` and choose `Add New Project` :
image
4. Accept the default value of "Project Name" to see:
image
Note that the workspace icon has turned red, indicating unsaved changes.
5. Attempt to exit N++ by clicking the main window's `X` in the upper right corner.  N++ will prompt you:
image
6. Click `Yes`
7. A FileSave dialog appears and has two buttons `Save` and `Cancel`.
8. Press `Cancel`.
9. N++ displays:
image
10. Press `OK`

### Actual Behavior
Notepad++ exits

### Expected Behavior
I expected Notepad++ to continue running, so that I could save my workspace changes at some future point in time, and not lose them.

For illustration of correct behavior, note what happens in a similar situation when you have a modified tab:
If your settings are such that if you have a red `new 1` file and you attempt to quit N++, you will see:
image
If you click `Yes` you are presented with a FileSave dialog appears and has two buttons `Save` and `Cancel`.
If you then press `Cancel` N++ does not exit, giving you the opportunity to still save the changes to this file.

### Debug Information
Notepad++ v7.9.3   (64-bit)
Path : C:\...........\npp.7.9.3.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1909
OS Build : 18363.1316
Current ANSI codepage : 1252

Uhf7:
Fixes #9605

This PR is also intended to improve the handling of write errors to Project Panel workspace files. Notepad++ distinguishes now between write errors and cancelled Save As dialogs. Write errors are always reported with a message box, Save As cancellations are silent.

I simulated write errors to the workspace file with the command line `Notepad >> abc.npj`. There is no write possible to `abc.npj`, until the Windows Notepad is terminated. I write this here because of the comment in the `English.xml`, which says the write error is hard to reproduce.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1178,7 +1178,8 @@ Do you want to recover your langs.xml?"/> <!-- HowToReproduce: Close Notepad++. 
 Please remove its root from the panel before you add folder &quot;$STR_REPLACE$&quot;."/> <!-- HowToReproduce: Add a folder like "c:\temp\aFolder\" into "Folder as Workspace" panel, then try to add "c:\temp\". -->
 
             <ProjectPanelChanged title="$STR_REPLACE$" message="The workspace was modified. Do you want to save it?"/>
-            <ProjectPanelChangedSaveError title="$STR_REPLACE$" message="Your workspace has not been saved."/> <!-- HowToReproduce: this message prevents from system failure. It's hard to reproduce. -->
+            <ProjectPanelSaveError title="$STR_REPLACE$" message="An error occurred while writing your workspace file.
+Your workspace has not been saved."/> <!-- HowToReproduce: this message prevents from system failure. It's hard to reproduce. -->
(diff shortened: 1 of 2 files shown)


sasumner:
>I write this here because of the comment in the English.xml, which says the write error is hard to reproduce.

If it isn't too long of a description (and it doesn't seem like it is), why not put it in the english.xml comment in place of the "hard to reproduce" text?

PR #9618, opened 2021-03-07
italian.xml maintenance

conky77:


donho:
@conky77 
You should modify this file only the parts needed. No re-order the entries or reformat the file.
I remember that we have talked about that.

conky77:
@donho  I confirm that you said in #8556 
"Such modification is not accepted normally because the difference can not be seen.
Since Italian localization is maintained by @Ices-Eyes mainly, I prefer it's him who does such change if he judges necessary." 
but he has never confirmed that he wants to keep the current order and he hasn't made a contribution for 6 months (since #8543). I propose that we can ask him if he would mind changing the order and we can wait for an answer from him. If he doesn't answer, I propose to make a decision with who answers. Let me know if you agree with my proposal. Thanks

donho:
@conky77 
> I propose that we can ask him if he would mind changing the order and we can wait for an answer from him. If he doesn't answer, I propose to make a decision with who answers. Let me know if you agree with my proposal. Thanks

Sounds good. Let's do it in this way.
@Ices-Eyes What do you think for the point we mentioned about?

donho:
@conky77 
Let's wait for about 2 weeks - if there's no reply from @Ices-Eyes , then please close this PR and create a new one with the new format/order you prefer.

Ices-Eyes:
Hi @donho and @conky77
main reason I would not change the order is that having string ID in order make it simple to check if a string code is already present in the file. Since for years the main way to detect new or modified strings was via the translation section of the N++ forum, where the tips was mainly as added string xxx yyy zzz, it was for me easier to detect those changes and check if already translated or not. But that days where I was the only translator for italian are ended, so if someone prefer to have another method or or sort of the keys, no problem on my side. ;)

donho:
@Ices-Eyes 
Thank you for your reply.

@conky77 
OK for the reorder and reformat in the next PR.

conky77:
Thanks, @donho  and @Ices-Eyes 
I have created PR #9635 (Italian.xml) with the English file order. So it is easier to use WinMerge to compare it with the English (and the Venetian) file. Thanks

PR #9651, opened 2021-03-13
Added batch processing of added and removed files in Folder as Workspace

### LINKED ISSUE #9203: Folder as Workspace: High CPU usage when files added in bulk
### Description of the Issue

When using "Folder as Workspace", if files are added to a subfolder (that need not be open in the view) in bulk on the background (e.g. pasted in via Explorer or written by an application), NP++ makes the CPU run at a full core's worth (or several, if several instances are open), to a point where it becomes unusable and needs to be terminated.

This has become a regular issue with a new routine where I'm clearing and rewriting the (1K+) cache files for an application that has its main folder added into the Folder as Workspace panel. The /cache/whatever/ subfolder is not (ever) open, but is included as one of the many subfolders of the application.

When NP++ is terminated and restarted, I can see all the newly added/generated files included in the subfolder, and this apparently requires next to no CPU or other resources, regardless of the volume of files. There must be something triggering this in the NP++ routine that watches folders on the background.

1. Open a folder with subfolders in "Folder as Workspace"
2. Copy a large amount of files (say, 1000+) into a subfolder on the background (e.g. via Explorer, or with a shell script that writes a batch of dummy files)
3. Watch NP++ churning the CPU forever

### Expected Behavior

NP++ would be momentarily done with the indexing (?), allowing me to continue writing code.

### Actual Behavior

NP++ maxes out the CPU and there's a constant flicker in the "Folder as Workspace" view. I haven't had the patience to see how long this might continue, suffice it to say this goes on and on.

### Debug Information

Notepad++ v7.9   (64-bit)
Build time : Sep 22 2020 - 03:19:04
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2009
OS Build : 19042.572
Current ANSI codepage : 1252
Plugins : AnalysePlugin.dll BigFiles.dll ComparePlugin.dll mimeTools.dll NavigateTo.dll NppConverter.dll nppcrypt.dll NppExport.dll NPPJSONViewer.dll NppTextViz.dll NppXmlTreeviewPlugin.dll pork2sausage.dll PythonScript.dll XMLTools.dll 

image

*This here routine breaks down a large JSON file into individual objects per file, saved in the /data/tgt/ subfolders (seen in the view). When I wipe and rebuild (PHP + Apache) a couple of thousand items, NP++ chokes on the process.*

jofon:
When using Folder as Workspace, every file added or removed is processed one at a time, so when many files are added or removed, Npp takes a long time to get back to a usable state (see #9203).

This PR changes processing of added or removed files to be done in batches. If multiple changes of the same type are received, they're grouped together for processing (with a maximum amount so Npp doesn't stay stuck forever). Processing of the files then groups them according to their paths, reducing the amount of calls to functions that traverse the TreeView.

Fixes #9203

### DIFF
--- a/PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
+++ b/PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
@@ -103,6 +103,8 @@ private:
 	HANDLE _watchThreadHandle = nullptr;
 	HANDLE _EventHandle = nullptr;
 	static DWORD WINAPI watching(void *param);
+
+	static void processChange(DWORD dwAction, std::vector<generic_string> filesToChange, FolderUpdater* thisFolderUpdater);
 };
 
 struct SortingData4lParam {
@@ -143,11 +145,10 @@ public:
 	void addRootFolder(generic_string rootFolderPath);
 
 	HTREEITEM getRootFromFullPath(const generic_string & rootPath) const;
-	HTREEITEM findChildNodeFromName(HTREEITEM parent, const generic_string&) const;
+	HTREEITEM findChildNodeFromName(HTREEITEM parent, const generic_string& label) const;
 
-	bool addInTree(const generic_string& rootPath, const generic_string& addItemFullPath, HTREEITEM node, std::vector<generic_string> linarPathArray);
 	HTREEITEM findInTree(const generic_string& rootPath, HTREEITEM node, std::vector<generic_string> linarPathArray) const;
-	bool deleteFromTree(const generic_string& rootPath, HTREEITEM node, const std::vector<generic_string>& linarPathArray);
+
 	void deleteAllFromTree() {
 		popupMenuCmd(IDM_FILEBROWSER_REMOVEALLROOTS);
 	};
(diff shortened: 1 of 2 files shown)


donho:
@jofon 
Not bad. But I find a critical regression :
Drop a folder in FaW and delete 1 file or folder in Explorer - that remove whole tree including the root.

donho (changes requested):
Fix the regression:
Drop a folder in FaW and delete 1 file or folder in Explorer - that remove whole tree including the root.

jofon:
Fixed the regression

donho (approved):

PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
```diff
@@ -188,6 +189,25 @@ class FileBrowser : public DockingDlgInterface {
 
 	bool selectCurrentEditingFile() const;
 
+	struct FilesToChange {
+		generic_string commonPath; // Common path between all the files. rootPath + linarWithoutLastPathElement
+		generic_string rootPath;
+		std::vector<generic_string> linarWithoutLastPathElement;
+		std::vector<generic_string> files; // file/folder names
+	};
```
donho, line 197: The member variable of class or struct should have the prefix `_`.
(Voir https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#member-variables)

donho (changes requested):
Add the prefix `_` on the member variable of struct.

PowerEditor/src/WinControls/FileBrowser/fileBrowser.h
jofon: Done

donho (approved):

PR #9654, opened 2021-03-14
Fix regression: Find button not enabled when activating window.

Uhf7:
Fixes the regression described here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8125#issuecomment-798801190


mere-human (approved):

Ekopalypse:
Yep, I downloaded the artifacts and tested them on Win7 x64. The reported problem did not occur.

PR #9666, opened 2021-03-16
Fix UNC path with quotes handling in File Open/Save Dialog

### LINKED ISSUE #9582: Open dialog: when pasting a complete UNC (\\server\share\...) path into the file name box, all but one leading backslash are removed when pressing enter.

1. Ctrl-o (Open dialog)
2. Enter UNC path to some file on some file share (\\myServer\myShare\folderx\myFile.xml)
3. Press ENTER: 
### Expected Behavior

File is opened from server.

### Actual Behavior

you see one leading backslash removed, followed by an error message

### Debug Information

Notepad++ v7.9.3   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2004
OS Build : 19041.804
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll HexEditor.dll HTMLTag.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll nppcrypt.dll NppExport.dll NPPJSONViewer.dll NppQrCode64.dll NppXmlTreeviewPlugin.dll RegexTrainer.dll SQLinFormNpp64.dll XMLTools.dll

mere-human:
Remove a workaround for double backslash that leads to problems.
Originally, the double backslash workaround was needed to address #5368.
But now, the app isn't stuck when such path is used, so the problem is gone.

Fix #9582

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -391,18 +391,6 @@ private:
 				return eq;
 			},
 			'\\');
-		// If there are two or more double backslash, then change it to single.
-		// Start from 2nd element to keep a path that starts from "\\".
-		auto last = std::unique(fileName.begin() + 1, fileName.end(),
-			[](generic_string::value_type a, generic_string::value_type b)
-			{
-				return a == b && a == '\\';
-			});
-		if (last != fileName.end())
-		{
-			fileName.erase(last, fileName.end());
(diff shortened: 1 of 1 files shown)


mere-human:
As reported in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9582#issuecomment-799393721

With this PR, paths like these should be opened without a problem:
* `\\127.0.0.1\c$`
* `"\\127.0.0.1\c$\Program Files (x86)\Notepad++\change.log”`
* `"\\127.0.0.1\c$\Python36-32\LICENSE.txt"`

PR #9667, opened 2021-03-18
Add "when" reserved word of C# 6.0 in auto-completion

digocesar:


donho:
It doesn't seems to me `when` is the keyword in C#. Could you provide any reference?

digocesar:
`when` was include in C# 6.0 for `catch` and in C# 7.0 for `switch/case`.
https://docs.microsoft.com/pt-br/dotnet/csharp/language-reference/keywords/when

ArkadiuszMichalski:
If accepted add also to `langs.model.xml` for  `<Language name="cs" ext="cs" commentLine="//" commentStart="/*" commentEnd="*/">`

image

digocesar:
> If accepted add also to `langs.model.xml` for `<Language name="cs" ext="cs" commentLine="//" commentStart="/*" commentEnd="*/">`

Done. ;)

PR #9669, opened 2021-03-18
Fix Italian.xml

conky77:


donho:
@conky77 

Thank you!

PR #9674, opened 2021-03-19
Fix file doesn't save as extension type on Windows 7

### LINKED ISSUE #9657: NP++  7.9.4 File doesn't save as extension type (win 7)
### Description of the Issue

Same issue as https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9510

Should have been fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f96d81af11de99fd48e4ef37e0265aa5e45033fd unless I misunderstand?

Unfortunately the commit did not fix the issue on win 7 at least.

In Notepad++ 7.9.2 I can do "Save as", file name: "lol" and "Save as type": "Normal text file (*.txt)" and the output file will be "lol.txt". Choosing the same options in 7.9.4 and the output file is named "lol" with no extension.

I don't have the option "Set Save dialog file extension filter to \*.\*" checked.

Edit for exact steps:
ALT -> F -> A -> this for for the "Save As..." menu, then I type my filename and hit Return.

The CTRL+ALT+S shortcut will also replicate the issue.

### Expected Behavior
Above

### Actual Behavior
Above

### Debug Information
Notepad++ v7.9.4   (64-bit)
Build time : Mar 15 2021 - 01:03:10
Admin mode : ON
Local Conf mode : OFF
OS Build : 7601.0
Current ANSI codepage : 1252

mere-human:
Initialize file dialog window procedures later so that
all the controls have been created by this moment.

Fix #9657

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -250,11 +250,9 @@ public:
 	{
 		return S_OK;
 	}
-	IFACEMETHODIMP OnFolderChange(IFileDialog* dlg) override
+	IFACEMETHODIMP OnFolderChange(IFileDialog*) override
 	{
 		// First launch order: 3. Custom controls are added but inactive.
-		if (!_dialog)
-			initDialog(dlg);
 		return S_OK;
 	}
 	IFACEMETHODIMP OnFolderChanging(IFileDialog*, IShellItem*) override
@@ -262,9 +260,11 @@ public:
 		// First launch order: 2. Buttons are added, correct window title.
 		return S_OK;
 	}
-	IFACEMETHODIMP OnSelectionChange(IFileDialog*) override
+	IFACEMETHODIMP OnSelectionChange(IFileDialog* dlg) override
 	{
 		// First launch order: 4. Main window is shown.
+		if (!_dialog)
+			initDialog(dlg);
 		return S_OK;
 	}
 	IFACEMETHODIMP OnShareViolation(IFileDialog*, IShellItem*, FDE_SHAREVIOLATION_RESPONSE*) override
@@ -300,6 +300,8 @@ private:
 	FileDialogEventHandler(FileDialogEventHandler&&) = delete;
 	FileDialogEventHandler& operator=(FileDialogEventHandler&&) = delete;
 
+	// Inits dialog pointer and overrides window procedures for file name edit and ok button.
+	// Call this as late as possible to ensure all the controls of the dialog are created.
 	void initDialog(IFileDialog * d)
 	{
 		assert(!_dialog);
@@ -347,9 +349,8 @@ private:
 		return fileName;
 	}
 
+	// Modifies the file name if necesary after user confirmed input.
 	// Called after the user input but before OnFileOk() and before any name validation.
-	// Prefer SendMessage communication with the edit box here rather than IFileDialog methods.
-	// The setter methods post the message to the queue, and it may not be processed in time.
 	void onPreFileOk()
 	{
 		if (!_dialog)


mere-human:
It would be nice to check this on Windows 7. I've done it on VM.

mere-human:
So the problem was that the Edit box control for the file name wasn't created yet when `EnumChildWindows` had been called.
It seems that Windows 7 has a different order of controls creation for this dialog.
This is the structure of the file dialog controls (just for the reference):
Screenshot 2021-03-19 112622

Now `EnumChildWindows` is called a bit later and everything is okay. 
Works as before on Windows 10.

sasumner:
>It seems that Windows 7 has a different order of controls creation for this dialog.

I didn't look but hopefully this didn't take too much effort to "fix".
If it had, it probably should have fallen into the "don't fix on non-current OS" bin.

mere-human:
> I didn't look but hopefully this didn't take too much effort to "fix".

No, it didn't. The fix is simple and the risk is minimal (as for me).
It took some time to figure out the root cause in my limited free time.

@donho Consider including this in 7.9.5.

donho:
@mere-human 
Has this PR been tested under Windows 10?

sasumner:
@mere-human said:

>Works as before on Windows 10.

donho:
@sasumner 
I saw his comment. I just want to make sure he did the test before testing it. To avoid the regression like before.

mere-human:
> Has this PR been tested under Windows 10?

Yes, I tested it on Windows 10. 
I tested it on Windows 7 VM too, but I did less number of test cases.
But the more people test it the better.

Below is my test list if anybody is interested.

Generic test cases:
- File dialogs:
  - File > Open, Save, Save as
  - Search > Find in files, enter something in “Directory”, press [...] (should open directory)
  - File > Open Folder as Workspace (should open directory)
  - Run > Run.. (exe file type)
  - etc
- Dialog confirmation:
  - Enter (Return) key
  - Press the button with a mouse
  - Select the button with tab and hit Enter/Space key
  - Press Alt + S shortcut
- File type
  - Language > Pick any (should be reflected in save dialog)
  - File > Save as > Select file type, enter name, save (should append extension)
  - File > Save as > Select file type (C, C++), enter a name with a different extension “tmp.txt” (extension should be preserved)
  - File > Save as > Change type then change name
  - Remove extension and ensure it’s added after confirmation
- Preferences
  - Settings > Default directory (should affect initial directory in file dialog)
  - Settings > MISC > Set Save dialog file extension filter to *.*
    - Shouldn’t append any extension
  - Settings > MISC > Session file ext (pick any)
    - View > Folder as Workspace (check)
    - File > Save Session, check “Save Folder as Workspace”
    - File type should be as set in preferences or *.*
    - Modify panel, open saved file
  - Settings > MISC >Workspace file ext (pick any)
    - View > Project Panel > Panel 1, add some entries to Project
    - Project Panel > Workspace > Save as
    - File type should be as set in preferences or *.*
    - Modify panel, open saved file

Path cases:
- File name
  - 1.txt
- Folder name - should navigate to it
  - e:\tmp
- Relative path
  - ..\tmp\4.txt
- Absolute path
  - e:\tmp\4.txt
- Forward slash path
  - e:/tmp/4.txt
  - ../Share
- Environment variable
  - %appdata%\tmp.txt
- UNC name:
  - \\127.0.0.1\c$
  - "\\127.0.0.1\c$\Program Files (x86)\Notepad++\change.log"
  - "\\127.0.0.1\c$\Python36-32\LICENSE.txt"

sasumner:
@mere-human 

I edited your posting above to change your "test list" into a code block.
Otherwise, the markdown consumes your `*` usage and some of your `\` usage.

mere-human:
@sasumner Thank you! It was displaying fine for me. 
Look:
image
I guess it could be rendered wrong in email.

donho:
@mere-human 
Thank you for your confirmation.

> Below is my test list if anybody is interested.

With all the bug-fixes and added features in the modern style file dialog, It would be great to have unit-tests for testing the list you provide. Any idea to process unit-tests for the implementation of file dialog?

mere-human:
> It would be great to have unit-tests for testing the list you provide. Any idea to process unit-tests for the implementation of file dialog?

Yes, good idea. I thought about the same.
I can provide more info on this later (as a separate issue).

PR #9677, opened 2021-03-20
Added Edge to the same case as the other browsers

### LINKED ISSUE #9313: Can't view file in Edge if name contains spaces
### Description of the Issue
Using the Chromium-based Edge browser, NPP's View Current File in Edge command doesn't work if the filename contains spaces. All of the other browsers in that same submenu display the file correctly.

1. Open or create a file with a name that contains one or more spaces.
2. Do View > View Current File in > Edge.

### Expected Behavior
Edge displays the file.

### Actual Behavior
Edge opens multiple tabs - one for each word in the file name - all of which display an error message.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.1256
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll HexEditor.dll mimeTools.dll NppConverter.dll NppExport.dll

jofon:
This correctly opens the files with spaces in the path, since the command ends up substituting the spaces for %20, which also works for Edge.

Fixes #9313

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2032,7 +2032,18 @@ void Notepad_plus::command(int id)
 				{
 					::ShellExecute(NULL, TEXT("open"), valData, fullCurrentPath.c_str(), NULL, SW_SHOWNORMAL);
 				}
-				else
+				else if (id == IDM_VIEW_IN_EDGE)
+				{
+					// Try the Legacy version
+
+					// Don't put the quots for Edge, otherwise it doesn't work
+					//fullCurrentPath = TEXT("\"");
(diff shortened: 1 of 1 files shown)


chcg:
Legacy edge will be uninstalled with the April update:
https://techcommunity.microsoft.com/t5/microsoft-365-blog/new-microsoft-edge-to-replace-microsoft-edge-legacy-with-april-s/ba-p/2114224

jofon:
I was unaware of the difference between the legacy and new version of Edge, and this PR breaks the functionality for Legacy. Thanks for the heads up @chcg. I'll take a look and see if there's a better way to do this without breaking it for the Legacy version.

PR #9678, opened 2021-03-20
SaveAs adds original path to recent files history

### LINKED ISSUE #9673: [Feature Request] STOP deleting file from recent files history when it was 'Saved As'
### Description of the Issue
File - Save As drops the original file from the recent files history.
I know there is File - Save A Copy As, but that is not my desire.

1. File - Save As
2. Look for original File in recent files history.
3. It is gone.

### Expected Behavior
I want recent files history to respect all files opened regardless of renaming events.

### Actual Behavior
Recent files history drops renamed files from its list.

### Debug Information
Notepad++ v7.9.3   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2009
OS Build : 19042.804
Current ANSI codepage : 1252

sasumner:
Fix #9673 

(also clean up some local leading tab/space inconsistency)


mere-human (approved):
I've just tried it and it works good!

PR #9683, opened 2021-03-22
Fix sort with column key selection that appears after tab characters

### LINKED ISSUE #9682: Sorting with column-key doesn't work following a tab character
### Description of the Issue
If one uses the sorting feature with a column-block defining the sort key, and this column-block occurs after a tab character on the lines, sorting doesn't work correctly.  
For some additional info, see https://community.notepad-plus-plus.org/post/63877

1. Create a new tab with the following data:
```
	b
	a
	f
	c
	e
	d

```
Same data shown with _View_ > _Show Symbol_ > _Show White Space and TAB_ enabled:
image
2. Define a column-block, as shown here:
image
3. Execute _Edit_ > _Line Operations_ > _Sort Lines Lexicographically Ascending_

### Expected Behavior
image

### Actual Behavior
No change; data is still in original order as shown in step 1 above.

### Debug Information

Notepad++ v7.9.4   (64-bit)
Build time : Mar 15 2021 - 01:03:10
Path : C:\\.......\npp.7.9.4.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1909
OS Build : 18363.1379
Current ANSI codepage : 1252

sasumner:
Fix #9682

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -596,8 +596,10 @@ void Notepad_plus::command(int id)
 					int endPos = max(leftPos, rightPos);
 					fromLine = _pEditView->execute(SCI_LINEFROMPOSITION, startPos);
 					toLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
-					fromColumn = _pEditView->execute(SCI_GETCOLUMN, leftPos);
-					toColumn = _pEditView->execute(SCI_GETCOLUMN, rightPos);
+					size_t fromLineStartPos = _pEditView->execute(SCI_POSITIONFROMLINE, fromLine);
+					size_t toLineStartPos = _pEditView->execute(SCI_POSITIONFROMLINE, toLine);
+					fromColumn = leftPos - fromLineStartPos;
+					toColumn = rightPos - toLineStartPos;
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


donho:
Regression. Sorting the 2nd column of the following line doesn't work anymore after applying this PR:
wb
xa
qf
ac
fe
hd

donho (changes requested):
Fix the regression due to this PR

sasumner:
@donho

>Sorting the 2nd column of the following line doesn't work anymore after applying this PR

I cannot reproduce this result.
If I make a selection one character wide, consisting of the second column, and sort-lex-ascending, it sorts to:

```
xa
wb
ac
hd
fe
qf
```

Same result if I begin with a zero-width caret between the two columns.

Can you describe your "steps to reproduce" a bit more?
Can you post exact text used in a code block?

donho:
OK. I can reproduce the bug steadily with whichever characters before the column: 
1. using the example I provided: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9683#issuecomment-808838753
2. put the caret at the end of hd.
3. column selection from d letter to b letter, from the lower-right to upper-left direction.
4. Menu "Edit->Line Operations->Sort Lines Lexicographically Ascendant"
5. Nothing sorted.

donho (approved):

PR #9684, opened 2021-03-22
Add ability to style only current instance of text

### LINKED ISSUE #9263: Style only the current selection/word, not all copies of identical text
### Description of the Issue
When styling text using _Search_ > _Mark All_ > _Using_ **Nth** _style_ , I'd like to be able to color ONLY the text I currently have selected, or only the current word at the caret if no selection is active.  This goes against the menu item's "Mark **ALL**" concept but in certain circumstances it can be valuable to do this.  One circumstance is when copying styled text to the clipboard and wanting to avoid redundant data.

Some other user requests for this:

* https://community.notepad-plus-plus.org/topic/20466
* https://community.notepad-plus-plus.org/post/59207

I alluded to the user needing this feature when I added the "Copy Styled Text" capability to Notepad++ via this issue and its PR:  #8943 , specifically in these comments:

- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8943#issuecomment-702878124
- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8943#issuecomment-702862998

1. Create some text in a Notepad++ tab:
```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.

Up Jack got
And home did trot,
As fast as he could caper;
And went to bed
And plastered his head
With vinegar and brown paper.

When Jill came in
How she did grin
To see Jack's paper plaster;
Mother vexed
Did whip her next
For causing Jack's disaster.
```

2. Place the caret inside the word `Jack` in the first line
3. Invoke _Search_ menu > _Mark All_ > _Using 1st style_
4. Observe all occurrences of `Jack` are styled in the cyan color:
image

### Expected Behavior
I expected a way to have only the currently selected text, or the current word at the caret, to be styled with one of the five styling colors, not all prior and future instances of the text.

### Actual Behavior
All occurrences of identical text were styled with the color I chose.  This was not what I needed.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Path : C:\\...........\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1577
Current ANSI codepage : 1252

sasumner:
Fix #9263


donho:
It's a good feature, but:
After checking "Style single occurrence when mark with #1 - #5", the menu entry group still named "Mark All".
I would say let's remove "Style single occurrence when mark with #1 - #5" checkbox and keep "Mark All" commands as they are. Add another group "Mark selected only" for this feature. What do you think about it?

sasumner:
>Add another group "Mark selected only" for this feature. What do you think about it?

We could do that. My original attempt was to keep it simple, although simple isn't always best, I suppose.

Perhaps it raises more questions, though. For example, we only have a clear-all -- do we also need to add a "clear one" capability?

donho:
> We could do that. My original attempt was to keep it simple, although simple isn't always best, I suppose.

It's also my intention - that's why this suggestion: user chose one (mark all) or another (mark selected only), no more option - so simpler.

> Perhaps it raises more questions, though. For example, we only have a clear-all -- do we also need to add a "clear one" capability?

We don't need to add a "clear one" capability, and user don't need it IMO. Currently "Clear all" group commands do exactly as the commands name, so there's no ambiguity.

donho (changes requested):
This feature is too good to not be added in the context menu:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/contextMenu.xml

sasumner:
>This feature is too good to not be added in the context menu

I previously considered it, but didn't want to add to the lack of localization in the context menu, reference https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9713

But, ok, I will add it in.

sasumner:
I didn't feel I could leave the existing context menu folder name as it was.
Changing it made it longer.
Not sure what will be thought of that.
But I suppose I will find out. :-)

donho (approved):

donho:
@sasumner 
You can always reopen #9713 then work on it.
User can use `subMenuId` instead of `FolderName` then the text of menu entry will be from the (translated) application instead of provided string.
I would accept the PR if the code is simple and straightforward. If there are a lot coding effort, then forget about it.

PR #9694, opened 2021-03-24
Update Slovak translation to 7.9.5

kubalav:


kubalav:
bad diff via browser github

PR #9704, opened 2021-03-26
Add command line parameter to add string to title bar

### LINKED ISSUE #9539: Add some argument that it's value will be added as substring in the title
### Description of the Issue
Currently, it is not possible to distinguish a particular instance of Notepad ++ (from the same `.exe` file or different `.exe` files). This cause that from another program I cannot refer to a specific instance of NPP and open the file in it. Let say we have two separate Notepad++ folder (portable):
```
A\notepad++.exe (shortly A)
B\notepad++.exe (shortly B)
```

Both are run at the same time. Step to reproduce:
1. run `A -multiInst`
2. run `B -multiInst` << let say this NPP was the last active
3. go to other window folder (or program)
4. do sth like this `A\notepad++.exe file.txt` (in console or through another program)

Our file will be open in `B` (not `A`) - it looks like due to recent activity on `B`. Even if we use only `A` and we open two instance we cannot distinguish them other than the order in the system bar or the PID of the process. Both solutions are not permanent and they are changes when we restart the program or when we change the order on the system bar.

With an additional argument that adding something to the title, it is very easy to distinguish different processes from each other. So when I starting Notepad ++ I can set some text in title, and then I can easily select such a window based on the title and then open the specific file / files in it:
1. run `A -multiInst -subtitle="main"`
2. run`B -multiInst`
3. go to other window folder (or program)
4. in console use some commands/tools that select/active window base on (sub)title and do sth like this `A\notepad++.exe file.txt`

Our file will be open in `A` (not `B` as previouse example).

In my case lack of such possibility is a problem. Very often I work with many versions of NPP at the same time (each for different purposes), and I am not able to delegate files to the correct instances of the program. Very often I forget to select the correct window before I running the openning action and have to repeat the whole process. 

### Debug Information
```
Notepad++ v7.9.2   (32-bit)
Build time : Jan 27 2021 - 17:53:35
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.0
Current ANSI codepage : 1250
Plugins : HexEditor.dll 
```

sasumner:
Fix #9539 

Example:

`notepad++ -titleAdd="for ArkadiuszMichalski"`

Result:

image

### DIFF
--- a/PowerEditor/src/Notepad_plus_Window.h
+++ b/PowerEditor/src/Notepad_plus_Window.h
@@ -20,7 +20,7 @@
 
 const TCHAR COMMAND_ARG_HELP[] = TEXT("Usage :\r\
 \r\
-notepad++ [--help] [-multiInst] [-noPlugin] [-lLanguage] [-LlangCode] [-nLineNumber] [-cColumnNumber] [-pPosition] [-xLeftPos] [-yTopPos] [-nosession] [-notabbar] [-ro] [-systemtray] [-loadingTime] [-alwaysOnTop] [-openSession] [-r] [-qn=\"Easter egg name\" | -qt=\"a text to display.\" | -qf=\"D:\\my quote.txt\"] [-qSpeed1|2|3] [-quickPrint] [-settingsDir=\"d:\\your settings dir\\\"] [-openFoldersAsWorkspace] [filePath]\r\
+notepad++ [--help] [-multiInst] [-noPlugin] [-lLanguage] [-LlangCode] [-nLineNumber] [-cColumnNumber] [-pPosition] [-xLeftPos] [-yTopPos] [-nosession] [-notabbar] [-ro] [-systemtray] [-loadingTime] [-alwaysOnTop] [-openSession] [-r] [-qn=\"Easter egg name\" | -qt=\"a text to display.\" | -qf=\"D:\\my quote.txt\"] [-qSpeed1|2|3] [-quickPrint] [-settingsDir=\"d:\\your settings dir\\\"] [-openFoldersAsWorkspace]  [-titleAdd=\"additional title bar text\"][filePath]\r\
 \r\
 --help : This help message\r\
 -multiInst : Launch another Notepad++ instance\r\
--- a/PowerEditor/src/Notepad_plus_Window.h
+++ b/PowerEditor/src/Notepad_plus_Window.h
@@ -48,6 +48,7 @@ notepad++ [--help] [-multiInst] [-noPlugin] [-lLanguage] [-LlangCode] [-nLineNum
 -quickPrint : Print the file given as argument then quit Notepad++\r\
 -settingsDir=\"d:\\your settings dir\\\": Override the default settings dir\r\
 -openFoldersAsWorkspace: open filePath of folder(s) as workspace\r\
+-titleAdd=\"string\": add string to Notepad++ title bar\r\
 filePath : file or folder name to open (absolute or relative path name)\r\
 ");
 
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -1672,6 +1672,15 @@ public:
 		_cmdSettingsDir = settingsDir;
 	};
 
+	void setTitleBarAdd(const generic_string& titleAdd)
+	{
+		_titleBarAdditional = titleAdd;
+	}
+	generic_string getTitleBarAdd()
+	{
+		return _titleBarAdditional;
+	}
+
 	DPIManager _dpiManager;
 
 	generic_string static getSpecialFolderLocation(int folderKind);
@@ -1750,6 +1759,8 @@ private:
 	bool _isx64 = false; // by default 32-bit
 
 	generic_string _cmdSettingsDir;
+
+	generic_string _titleBarAdditional;
 
 public:
 	void setShortcutDirty() { _isAnyShortcutModified = true; };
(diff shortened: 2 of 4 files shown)


PowerEditor/src/Parameters.h
```diff
@@ -1672,6 +1672,16 @@ class NppParameters final
 		_cmdSettingsDir = settingsDir;
 	};
 
+	void setTitleBarAdd(const generic_string& titleAdd)
+	{
+		_titleBarAdditional = titleAdd;
+	}
+
+	generic_string getTitleBarAdd()
```
mere-human, line 1680: It's better to declare as `const generic_string& getTitleBarAdd() const {...}`

PowerEditor/src/winmain.cpp
```diff
@@ -464,11 +465,20 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
 		nppParameters.setCmdSettingsDir(path);
 	}
 
+	generic_string titleBarAdditional;
+	if (getParamValFromString(FLAG_TITLEBAR_ADD, params, titleBarAdditional))
+	{
+		// could contain double quotes if contains white space
+		if (titleBarAdditional.c_str()[0] == '"' && titleBarAdditional.c_str()[titleBarAdditional.length() - 1] == '"')
```
mere-human, line 472: 1. Check that `titleBarAdditional` is not empty.
2. You can actually do `titleBarAdditional[0]` or even `titleBarAdditional.front()` and `titleBarAdditional.back()`.
3. What if `titleBarAdditional == "\""`?

sasumner: This is copy/paste/change/don't-think from recent DonHo change involving "settings dir" that suffers from same problems.  :-(

PowerEditor/src/Parameters.h
sasumner: True.  Something felt strange while I was doing it.  :)

sasumner: Changed.  Thanks @mere-human.

PowerEditor/src/winmain.cpp
sasumner: Changed.  Thanks @mere-human.

PowerEditor/src/Notepad_plus_Window.h
```diff
@@ -48,6 +48,7 @@ notepad++ [--help] [-multiInst] [-noPlugin] [-lLanguage] [-LlangCode] [-nLineNum
 -quickPrint : Print the file given as argument then quit Notepad++\r\
 -settingsDir=\"d:\\your settings dir\\\": Override the default settings dir\r\
 -openFoldersAsWorkspace: open filePath of folder(s) as workspace\r\
+-titleAdd=\"string\": add string to Notepad++ title bar\r\
```
mere-human, line 51: Probably, this also has to be added to the list above, right before `[filePath]`.

mere-human (commented):
Looks good:
image
image

By the way, if you specify empty quotes, it will be added as is. Not sure if it is bad or not:
image

sasumner:
>By the way, if you specify empty quotes, it will be added as is. Not sure if it is bad or not

I'm not sure, either.
Maybe it would serve as a prompt to the user that they forgot to put something BETWEEN the `"`?

I think I will remove it by comparing `>=2` rather than `>2`.
So that if user specifies `=""` then nothing will be shown.

If user specifies `="` however then it will look like this (can't protect always from crazy users):
image

ArkadiuszMichalski:
@sasumner 
`notepad++ qt="` also operates on text (but in edit area) and does not enter this first ". Is there any escape character if I really want put such char? I've tried `& \ <> ^ | or even %34` and neither works. 
```
notepad++ -qt="test"s" > display "test"s
notepad++ -qt="test"s` > display "test"s
notepad++ -qt="test` > display "test
notepad++ -qt="` > display nothing

notepad++ -qt="I want " and more texts."
``` 

But I see that the syntax for `-qt`, `-qf` and `-qn` has changed. We haven't used `=""` before, and space was replaced by `%20%`.

sasumner:
First, I don't care one whit about the `-qX` options so I can't address any concerns regarding those.

Second, I don't see it as a concern for the simple feature that is the subject of this PR/issue.
Use `'` instead if you really have to have some type of quoting in the string that is added to the title bar.
I was willing to add the feature; however, I'm not willing to make it infinitely flexible unless a reason can be substantiated.
Remember what happened with the `-n` , `-c` and `-p` command line arguments?  The infinite flexibility got it killed.

ArkadiuszMichalski:
>Remember what happened with the -n , -c and -p command line arguments? The infinite flexibility got it killed.

I remember, I want only add that if `-qt="` does nothing, maybe `-titleAdd="` should do the same, just more consistency. I thought maybe there is some hidden escape character that I don't know about, I'm not asking you to enter it.

sasumner:
>I remember, I want only add that if -qt=" does nothing, maybe -titleAdd=" 

We cannot examine the entirety of Notepad++ 's history when doing such an addition.
It is just such a small feature it is not important enough to do something like that.
Like the `-n` etc options, the answer for this one is: "don't specify `-titleAdd="` (and why would you?)
Thanks to @mere-human for calling attention to the single `"` case!  :-)

donho:
> By the way, if you specify empty quotes, it will be added as is. Not sure if it is bad or not

If an user add an appalled quote, then he/she is appalled, I wouldn't blame to this feature but him/her :D 
Users have to take the responsibility for what they do. Notepad++ should execute the instructions faithfully, period.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/cae2f77daf045408b683d194bbf1b09061edb872

PR #9718, opened 2021-03-29
Update russian.xml to v7.9.6

GOLEM777:
Updated to v7.9.6:
- added new lines
- minor fix


donho:
@GOLEM777 

Thank you!

PR #9728, opened 2021-04-03
Reword command name of Run menu's Open file in new instance

### LINKED ISSUE #9724: Cannot spawn another instance of Notepad++ for an empty non-saved document tab
### Description of the Issue
Cannot spawn another instance for a new document tab that have never been saved.

1. Open Notepad++ with other documents open in tabs.
2. Click "File -> New".
3. Drag the new tab off Notepad++ to spawn another shell.  It doesn't work.

Workaround is to type something, save the document as new.txt or whatever, and drag that tab off Notepad++.   Then, a new notepad++ will be spawned but the document had to be saved first.

### Expected Behavior
A new spawned Notepad++ with new empty non-saved tab.

Alternatively, the File menu could have a menu option to spawn a new Notepad++ with an empty non-saved tab.

### Actual Behavior
Notepad++ does not allow spawning another Notepad++ for a non-saved new document tab.

sasumner:
Fix #9724

...at least the really broken part.

@donho I saw no reason for `$(CURRENT_WORD)` to be a parameter to the command, so this PR removes it -- do you see any reason for it to be there?  I suspect a copy-paste-forgetToRemove bug from the previous command when this one was created.

### DIFF
--- a/PowerEditor/src/shortcuts.xml
+++ b/PowerEditor/src/shortcuts.xml
@@ -10,7 +10,7 @@
     <UserDefinedCommands>
         <Command name="Get PHP help" Ctrl="no" Alt="yes" Shift="no" Key="112">https://www.php.net/$(CURRENT_WORD)</Command>
         <Command name="Wikipedia Search" Ctrl="no" Alt="yes" Shift="no" Key="114">https://en.wikipedia.org/wiki/Special:Search?search=$(CURRENT_WORD)</Command>
-        <Command name="Open file in another instance" Ctrl="no" Alt="yes" Shift="no" Key="117">$(NPP_FULL_FILE_PATH) $(CURRENT_WORD) -nosession -multiInst</Command>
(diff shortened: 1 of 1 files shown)


donho:
`$(CURRENT_WORD)` should be kept and "Open file in another instance" should be "Open file in another instance via selected file path". However, it's too long as menu entry name. If you have any suggestion for a better menu entry name, please let me know.

mere-human:
Maybe, show a tooltip that says the selected word will be used as a file path?

> `$(CURRENT_WORD)` should be kept and "Open file in another instance" should be "Open file in another instance via selected file path". However, it's too long as menu entry name. If you have any suggestion for a better menu entry name, please let me know.

sasumner:
How about ""Open selected file path in new instance?"

donho:
@sasumner 
OK for me.

sasumner:
@mere-human said

>Maybe, show a tooltip that says the selected word will be used as a file path?

I don't think tool tips are supported on menu items, but that might be a useful thing for the future.

PR #9733, opened 2021-04-06
Correct locale while compiling the regex.

Uhf7:
This is a correction of PR #9707. @MarkusBodensee noticed [here](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/86c66bba90cc3f49323cdcae77173099122c75dc#r49146650), that this line doesn't stick to the plan.

### DIFF
--- a/scintilla/boostregex/BoostRegExSearch.cxx
+++ b/scintilla/boostregex/BoostRegExSearch.cxx
@@ -416,7 +416,7 @@ void BoostRegexSearch::EncodingDependent<CharT, CharacterIterator>::compileRegex
 {
 	if (_lastCompileFlags != compileFlags || _lastRegexString != regex)
 	{
-		std::locale l = std::locale("");
+		std::locale l = std::locale::global(std::locale(""));
 		_regex = Regex(CharTPtr(regex), static_cast<regex_constants::syntax_option_type>(compileFlags));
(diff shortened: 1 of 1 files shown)


donho:
@Uhf7 
> This is a correction of PR #9707.

Thank you for your PR.
If it's a "correction", could you provide the instructions to reproduce the bug or the regression please?

Uhf7:
There is no visible bug or functional regression due to the wrong line. The problem is another. The idea, or plan, of PR 9707 is to wrap each RegEx operation into two additional lines:
```
std::locale l = std::locale::global(std::locale(""));
<regex operation>
std::locale::global(l);
```
The first line is intended to set the global locale to "", the last line is to restore the previous locale, so that the locale is not changed for the rest of the program because of searching.

Due to the wrong line before the compilation of the RegEx, kind of opposite is reached. After the compilation of the RegEx, the global locale is changed to "" now, which is not intended. It has, however, in the moment, no effect for the function, so that I didn't see it while testing. And, for the issue 9636, the locale during the RegEx compilation seems to have no influence onto the result. I cannot know if this is the case under all circumstances, the idea is, make the locale equal for all calls into the Boost RegEx engine.

To avoid possible unknown effects in the future, the global locale should be set and restored correctly in all three places. Sorry for your additional time and effort.

PR #9749, opened 2021-04-09
Update czech translation to v7.9.6

xomx:


donho:
@xomx 

Thank you!

xomx:
@donho 
It is done.
Thanks for the Dark Mode, I like it! Also the new icons are superb!

PR #9765, opened 2021-04-13
Add context menu to ShowAllCharacters toolbar button

### LINKED ISSUE #9740: "Show all characters" toolbar button forgets "Show symbol" View setting.
### Description of the Issue
It would be convenient if the "Show all characters" toolbar button respected/remembered the "Show symbol" View setting. Right now, toggling (i.e. clicking twice) the "Show all characters" toolbar button will disable the "Show white space and tab" View option.

1. Enable `View` > `Show symbol` > `Show white space and tab`
2. Text shows orange dots and arrows, but no LF.
3. Click `Show all characters` toolbar button
4. Text shows orange dots and arrows, **and LF**.
5. Click `Show all characters` toolbar button

### Expected Behavior
6. Text shows orange dots and arrows, **but no LF**.

### Actual Behavior
6. Text shows no orange dots, no arrows, and LF.
7. `View` > `Show symbol` has `Show white space and tab` unchecked

### Debug Information
```Notepad++ v7.9.1   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2004
OS Build : 19041.746
Current ANSI codepage : 1252
Plugins : BetterMultiSelection.dll CodeAlignmentNpp.dll ComparePlugin.dll DSpellCheck.dll JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll XMLTools.dll 
```

### Additional information
I haven't actually tested this, but I think this is how this feature used to work in the past. Regardless, it's sort of inconvenient because I like to always see the spaces and tabs (they are not very intrusive, with the light-orange colours), but I only enable the end-of-line display occasionally (because it's much more noisy with the heavy black characters).

sasumner:
Fix #9740 

Or at least make the button more useful by adding a right-click context menu that quickly allows setting any of the FOUR setting choices (rather than only TWO states that the button currently allows).

In the following the ¶ was right-clicked upon to generate the popup menu:

image

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -499,13 +499,33 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				}
 			}
 			else // From tool bar
+			{
+				LPNMMOUSE lpnm = (LPNMMOUSE)notification;
+				if (lpnm->dwItemSpec == IDM_VIEW_ALL_CHARACTERS)
+				{
+					ContextMenu viewAllCharsToolbarButtonContextMenu;
+					std::vector<MenuItemUnit> itemUnitArray;
+					NativeLangSpeaker* nativeLangSpeaker = NppParameters::getInstance().getNativeLangSpeaker();
+					itemUnitArray.push_back(MenuItemUnit(IDM_VIEW_TAB_SPACE, 
+						nativeLangSpeaker->getNativeLangMenuString(IDM_VIEW_TAB_SPACE, TEXT("Show White Space and TAB"))));
+					itemUnitArray.push_back(MenuItemUnit(IDM_VIEW_EOL, 
+						nativeLangSpeaker->getNativeLangMenuString(IDM_VIEW_EOL, TEXT("Show End of Line"))));
+					itemUnitArray.push_back(MenuItemUnit(IDM_VIEW_ALL_CHARACTERS, 
+						nativeLangSpeaker->getNativeLangMenuString(IDM_VIEW_ALL_CHARACTERS, TEXT("Show All Characters"))));
+					viewAllCharsToolbarButtonContextMenu.create(_pPublicInterface->getHSelf(), itemUnitArray);
+					ScintillaViewParams& svp = (ScintillaViewParams&)(NppParameters::getInstance()).getSVP();
+					viewAllCharsToolbarButtonContextMenu.checkItem(IDM_VIEW_TAB_SPACE, svp._whiteSpaceShow && !svp._eolShow);
+					viewAllCharsToolbarButtonContextMenu.checkItem(IDM_VIEW_EOL, !svp._whiteSpaceShow && svp._eolShow);
+					viewAllCharsToolbarButtonContextMenu.checkItem(IDM_VIEW_ALL_CHARACTERS, svp._whiteSpaceShow && svp._eolShow);
+					viewAllCharsToolbarButtonContextMenu.display(p);
+				}
 				return TRUE;
-			//break;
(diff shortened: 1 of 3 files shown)


sasumner:
It might make more sense to have the context menu NOT have "Show All Characters".

The button, without the menu, toggles between showing none and showing all, so that capability is provided for.  
So really, all that is really needed is to add a quick way of choosing either the "white space..." and "EOL" choices.

What's the opinion on this?
@vinsworldcom ?

vinsworldcom:
That sounds logical - current button functionality is maintained and the right click context menu provides the additional options.  I didn't request this though, just think it's a super idea so hopefully the OP weighs in.  Otherwise I like your latest recommendation, button toggles all on/off and context menu provides the other 2 choices.

Cheers.

sasumner:
@vinsworldcom Right, soliciting your opinion directly because if I don't then I probably don't get ANY feedback (a problem here).  Often we don't fix the problem the way the OP directly wants, so I don't care if the OP weighs in or not. :)  Your comment of "super idea" is enough for me. :)

Since I already completed the PR as it is, I'll leave it and see what @donho says at evaluation time; he may like that the context menu mirrors the _Show Symbol_ submenu contents.  Or I'll be asked to change it as discussed.  Or he might not even like any part of this at all...

GOLEM777:
Maybe just add button of `Show White Space and TAB`  to the panel?

KnIfER:
Right-click context menu, if is use frequenlty, could be further enhanced. (Not necessary, but possible.)

Imagine when the user what to click an item, he needs to right click on the button -> move the mouse -> left click on the menu item, thus three actions in total are required. We can merge those three actions into "one" :  Right click on the button, without releasing the mouse button, move the mouse cursor to the position of the menu item, and finally release the right mouse button. Actually this is the default behaviour of window's "left-click" menus, very fluent.

sasumner:
>Right-click context menu on the buttons, if is used frequenlty, could be further enhanced. (Not necessary, but possible.)

>Imagine when the user what to click an item, he needs to right click on the button -> move the mouse -> left click on the menu item, thus three actions in total are required. We can merge these three actions into "one" : Right button down on the button, (show the context menu), without releasing the mouse button, move the mouse cursor to the position of the menu item, and finally release the right mouse button. Actually this is the default behaviour of window's "left-click" menus, very fluent.

I think it might actually be 5 actions (currently):

* right click 
* release right click (context menu appears at this point)
* move mouse to desired item
* left click
* release left click (to select current item)

Versus a possible 3 actions:

* right click (context menu appears)
* move mouse to desired item
* release right click (to select current item)

Nevertheless, something like that, if wanted, should be implemented throughout Notepad++.
Feel free to create a new issue concerning it.

>this is the default behaviour of window's "left-click" menus, very fluent.

I never thought about it before, but yes, it is a valid point.

GOLEM777:
But I would replace `All Characters` on `Show White Space and TAB`! )))
But the idea is good. Easter egg for the initiated.

sasumner:
IMO:
The problem is that most users will want "white space and tab" or "nothing".
Toolbar button only provides "whitespace and tab and line endings" or "nothing".
The line ending are so damn intrusive they can't be left turned on for longer than it takes to spot check a suspected problem.

vinsworldcom:
@sasumner 1000% agree - "white space and tab" or "nothing".  "The line ending are so damn intrusive they can't be left turned on for longer than it takes to spot check a suspected problem."

Cheers.

GOLEM777:
And if we redraw the icons with designation  submenu (context)?
And implement a change of icon for the selected operation?

Issue #5148, opened 2018-12-17
Applying Style Tokens is case sensitive until search is done


ebouden:
### Steps to Reproduce the Issue

Preparation to default case insensitive search:
1. Open Notepad++
2. Press Crtl-N  (create a new file)
3. type next text:
Case
case
CASE
Case
caSE
4. press Crtl-F, uncheck [Match case]
5. Press [Find Next]
6. press Escape
7. save the file (with some valid name on some valid path)
8. Close Notepad++
Now reproduce the issue with this case-insensitive default setting:
9. Open Notepad++ (it opens with the earlier saved file)
10. Right-click on the first word ("Case") in this same file > Style Token >  Using 1st Style
11. Now check that only the two exact occurrences of the word "Case" are highlighted in blue.
12. Press Crtl-F, Press Find Next, Press excape
13. Right click the 2nd word "case" in this same file > Style Token > Using 1st Style
14. Now check that ALL occurrences of the word "case" are highlighted in blue, no matter their case.

### Expected Behavior

I expected that all occurrences were already colored blue after executing step 10.

### Actual Behavior

They were not colored blue after executing step 10, instead, some occurrences that did not exactly match the case were skipped.
The correct coloring only happened after step 13, so when a case insensitive search was done before it.

### Debug Information

Notepad++ v7.5.9   (32-bit)
Build time : Oct 14 2018 - 15:02:52
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll AnalysePlugin.dll NppTextFX.dll PluginManager.dll PythonScript.dll 

Motivation:
I created a search engine (launched with hotkey in Notepad++) operating on a structured text file which produces a "search result"-file - a simple text file - as result, which is then opened by launching a fresh Notepad++ instance with it. Plugin PythonScript is installed. Its Startup.py is modified, such that it searches every opened file for some tags (which are only present in the specific search result file).
The script reads the tags and some keywords (which were used by the search engine), then it applies Style Tokens to the first 5 keywords. But as the colors are currently only applied to the case-exact matches of the keywords, the search overview is not complete (and the hot-keys Crtl-1..Crtl-5 will only jump to the colored ones). I'd like to see all cases to be colored - and so usable with Crtl-1..Ctrl-5.

[...]

PR #9769, opened 2021-04-14
Add MarkAll settings for case/word

### LINKED ISSUE #9751: Style Token not applied consistently within words
I noticed an issue with the "style token" feature. The problem is that styling is not applied consistently within words.

### Case 1

1. Close and open Notepad++.
1. Type "training" on line 1 and "train" on line 2.
1. With nothing selected, right-click on "train" on line 2 > Style token > Using 1st Style.

Under these conditions, only "train" on line 2 is styled.

screen capture where line 2 is styled

----

### Case 2 

1. Close and open Notepad++.
1. Type my test: "training" on line 1 and "train" on line 2.
1. *Open the Find dialog*.
1. *Clear 'Match whole word only'*.
1. With nothing selected, right-click on "train" on line 2 > Style token > Using 1st Style.

Under these conditions, "train" on lines 1 and 2 are styled: The token will be styled within words.

screen capture where both lines are styled

Style Token will continue to match within words after opening the Find dialog once. 

### Expected Behavior

I am accustomed to case 2 where style applies to matches within words. This is how Notepad++ operates most of the time. 

It doesn't make sense that a Find dialog option would affect Style Token. I suggest decoupling Style Token from the Find dialog. If people want to control this behavior then I suggest adding a formal preference: Style Token Matching. 

### Debug Information
Notepad++ v7.9.5   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2009
OS Build : 19042.867
Current ANSI codepage : 1252

sasumner:
Fix #9751 #5148

I ended up going with this UI:

image

as it is "less busy" than the earlier one I proposed [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9751#issuecomment-817137211) and it also somewhat mirrors the preferences for _Smart Highlighting_ :

image

Like _Smart Highlighting_, here the case and word choices may both be "checked" simultaneously.

The "classic" setting maintains the behavior existing as the only choice in 7.9.5 and earlier.

I solicited @donho 's opinion [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9751#issuecomment-817137211) but didn't get a response, so I proceeded on my own; hopefully things turn out well for the direction I chose. :-)

Did a bit of renaming of variables, as the code section that referred to these two made things hard to understand:

```z
TCHAR text2Find[strSize];
TCHAR text2Find2[strSize];
```


donho:
1. The settings should be on menu command.
2. I think we can find the fix without reconfiguring "Mark all" - at least not 6 options.

sasumner:
>at least not 6 options.

Hmmm, I count 4.

sasumner:
Pre-closure, I had some revisions completed due to @ArkadiuszMichalski suggestion.
I will reopen so that I can get the changes archived and a build artifact created.
After that I will close again.

sasumner:
This was the change I wanted to get in:

image

Re-closing.

@ArkadiuszMichalski What do you think?  Haha, not that it matters much. :-)

ArkadiuszMichalski:
Generally, I don't use this function because it's not remembered, so I can't say what's better and what's not. However, I think that if I used it often, and needed different variants, I would rather have them quickly available as above, than hidden somewhere in the Preferences window.

donho:
> Hmmm, I count 4 in the screenshot above.

Oups, not all Asians are good at math :)
I was talking this suggestion:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9751#issuecomment-817137211

> This was the change I wanted to get in:

It makes more sense than this one: 
image

but still, in the main menu let's keep the action logic. If these 4 options are necessary, we put them into Preferences diag.

sasumner:
>If these 4 options are necessary, we put them into Preferences

It can be done, but for me as a user (how many others?) it makes it difficult for me to use the feature, because I change these settings often (currently I do this poorly due to limited control).  As well, I can't macro-ize a setting in the Preferences.

As I consider all of this, I wonder if it the whole feature should be another tab in the _Find_ window family.
However, I've always considered this a "lightweight" feature, not needing a dedicated UI that that would bring.

sasumner:
>Generally, I don't use this function because it's not remembered

To be explicit, you mean across runs of Notepad++.
Like, for example, bookmarks are remembered.

Is that a call for a feature-request, in order to do this remembering?  :-)

ArkadiuszMichalski:
>Is that a call for a feature-request, in order to do this remembering? :-)

I think I've seen a question about this somewhere before, but I don't know if it's still open or rejected. If it was remembered, maybe I would start using it, but at the moment there are more important things that I miss. In general, mentioned bookmarks are what I need, but I wish we could label them and display this label (e.g. in the function list, as a subgroup).  They are non-invasive for the content and actually remembered, so they're perfect for making shortcuts to certain things in our code for all file types. I will open a separate issue for this.

donho:
@sasumner 
> It can be done, but for me as a user (how many others?) it makes it difficult for me to use the feature, because I change these settings often (currently I do this poorly due to limited control). As well, I can't macro-ize a setting in the Preferences.

I do understand the need. But the coherence in the program should be kept. While the exact settings of smart highlighting existing in Preferences dialog, I don't see the reason to not reuse them or at least regroup them at the same place.

Issue #9706, opened 2021-03-27
Recent file not existing is removed from recent list


jringbox:
### Description of the Issue

Drive not mounted causes recent file to be removed from recent list.   I have to navigate all the way to the file again to open the file.  This is time consuming and annoying.

To be clear, this is the user's fault but removing from the recent list without giving options causes more work for the user to put the file back in the recent list, esp if this accidentally happens frequently for different files.  I guess I'm absent minded sometimes.

### Steps to Reproduce the Issue

1.  Click on recent file in list where the mount drive has not been mounted.
2.  Message is displayed with "Cannot open file" with "FILE cannot be opened: Folder FOLDER doesn't exist." with "OK" button below.
3. Click "OK".  Go to the File menu and see the file is gone from the recent list.

### Expected Behavior
Show this message instead: "FILE cannot be opened: Folder FOLDER doesn't exist.  Do you wish to remove from recent list?" with "Yes" or "No" button below.  "Yes" button can be defaulted focus.

### Actual Behavior
This message is shown:  "Cannot open file" with "FILE cannot be opened: Folder FOLDER doesn't exist." with "OK" button.  After clicking OK, there is no way to avoid the invalid file from not being removed from the recent list.

Joreminga:
I have implemented the feature request as specified above. I will issue a pull request shortly.

PR #9772, opened 2021-04-16
Implemented upgrade request based on issue #9706

Joreminga:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9706

Added a yes or no option to the file does not exist dialog to give the user the option to keep a file in the recent files list if it does not exist. 

Use case: the file may exist on a usb drive that is not currently connected to the computer or a network drive and the user does not want to remove the file from the recents list.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -198,11 +198,11 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive, 
 	{
 		wcscpy_s(longFileName, targetFileName.c_str());
 	}
-    _lastRecentFileList.remove(longFileName);
 
 	generic_string fileName2Find;
 	generic_string gs_fileName{ targetFileName };
 
+	bool removeFileFromList = true;
 
 	// "fileName" could be:
 	// 1. full file path to open or create
@@ -232,17 +232,20 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive, 
                 ::SendMessage(_pPublicInterface->getHSelf(), WM_SIZE, 0, 0);
             }
         }
+		_lastRecentFileList.remove(longFileName);
         return foundBufID;
     }
 
     if (isFileSession(longFileName) && PathFileExists(longFileName))
     {
+		_lastRecentFileList.remove(longFileName);
         fileLoadSession(longFileName);
         return BUFFER_INVALID;
     }
 
 	if (isFileWorkspace(longFileName) && PathFileExists(longFileName))
 	{
+		_lastRecentFileList.remove(longFileName);
 		nppParam.setWorkSpaceFilePath(0, longFileName);
 		// This line switches to Project Panel 1 while starting up Npp
 		// and after dragging a workspace file to Npp:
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -302,8 +305,14 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive,
 				str2display += longFileName;
 				str2display += TEXT("\" cannot be opened:\nFolder \"");
 				str2display += longFileDir;
-				str2display += TEXT("\" doesn't exist.");
-				::MessageBox(_pPublicInterface->getHSelf(), str2display.c_str(), TEXT("Cannot open file"), MB_OK);
+				str2display += TEXT("\" doesn't exist.");
```
sasumner, line 308: New UI text should be made language translatable.

PowerEditor/src/NppIO.cpp
```diff
@@ -302,8 +305,14 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive,
 				str2display += longFileName;
 				str2display += TEXT("\" cannot be opened:\nFolder \"");
 				str2display += longFileDir;
-				str2display += TEXT("\" doesn't exist.");
-				::MessageBox(_pPublicInterface->getHSelf(), str2display.c_str(), TEXT("Cannot open file"), MB_OK);
+				str2display += TEXT("\" doesn't exist."); 
+				str2display += TEXT("Do you wish to remove the file from the recent list?");
+				int msgBoxResponse = ::MessageBox(_pPublicInterface->getHSelf(), str2display.c_str(), TEXT("Cannot open file"), MB_YESNO);
```
sasumner, line 310: Good opportunity to make existing UI text language translatable.

sasumner:
Does this deserve an option in the _Preferences_ to enable/disable it?

Should this have a "Yes to all" / "No to all" capability for the case of multiple files in this condition?

PowerEditor/src/NppIO.cpp
Joreminga: How is translation handled in Notepad++? 
What would I need to add in order to enable localization?
Please let me know and I will fix it.
Thanks!

Joreminga:
I think it would be a good idea to add in a "Yes to all" or "No to all" option for when the user tries to close multiple files.
Let me know if you would like me to push up a new commit with changes to enable this.
Thanks!

sasumner:
>Let me know if you would like me to push up a new commit with changes to enable this.

It's your PR, so it is your decision.
I was merely posing it as a question for thought.

PowerEditor/src/NppIO.cpp
mere-human: What you need is `NativeLangSpeaker`, see https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/localization.h.
Take a look at other places that are already localized for examples.

Joreminga: Perfect, I'll get the localization work done tonight and get a fix in. Thanks for the help!

Joreminga:
I think I will actually leave out the "Yes to all" or "No to all" option. When you attempt to close multiple files with unsaved work at the same time there isn't a "Yes to all" or "No to all" option so I will try to keep the software consistent with that standard and just ask individually for each file.

sasumner:
>When you attempt to close multiple files with unsaved work at the same time there isn't a "Yes to all" or "No to all" option

???
Yes there is:

image

Joreminga:
Ah I see now, I was being dumb.
In that case I will add in the "Yes to all" and "No to all" options.
Thanks for the help!

Joreminga:
I looked into what it will take to implement a "Yes to all" and "No to all" option for this enhancement request.
If we want everything to be translatable using NativeLangSpeaker then I will need to add some code there to replicate the functionality displayed in the close multiple files dialog (DoSaveOrNotBox). 

Additionally I noticed that when multiple files do not exist on the system and you attempt to load them Notepad++ does not give a "Yes to all" or "No to all" option there.

My proposal would be to close this pull request out and I will create a new issue (enhancement) to add in "Yes to All" and "No to All" options to both the dialog present in this issue and the create file if not exists dialog I mentioned above.

Please let me know if that is an acceptable course of action.
Thanks, Josh.

PR #9773, opened 2021-04-16
Update localization: Hong Kong Cantonese...

Edditoria:
As at commit 9c66ff9 of this repo.


donho:
@Edditoria 

Thank you!

Edditoria:
@donho Updated. Thanks for notice.

I assume that the next version will be `v7.9.6`. Please adjust it if it is wrong.

Issue #9766, opened 2021-04-13
Search option 'whole word' does no longer work


m-wim:
### Description of the Issue

Search option 'whole word' does no longer work

### Steps to Reproduce the Issue

1.  open search dialog, enter text e.g. file
2.  chose option to search for whole word 
3.  perform the search

### Expected Behavior

Search result shows only locations with the full word e.g. file and not files

### Actual Behavior

Search result shows all occurrences 

### Debug Information

Notepad++ v7.9.5   (64-bit)
Build time : Mar 21 2021 - 02:13:17
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.1377
Current ANSI codepage : 1252
Plugins : AnalysePlugin.dll BetterMultiSelection.dll ComparePlugin.dll Explorer.dll mimeTools.dll NppConverter.dll NppEventExec.dll NppExec.dll NppExport.dll NppGTags.dll NppTextViz.dll NppXmlTreeviewPlugin.dll

sasumner:
>Search option 'whole word' does no longer work

It does work.  
Closing until a specific counter-example is provided (if that's possible).

m-wim:
More details: the regexp search checkbox was enabled but the search option 'whole word' was not disabled. I switched the two options several times, and now the 'whole word' option is unchecked and grayed with regexp option enabled - but it wasn't before. This could have been the first start after version update.

sasumner:
Easy reproduction steps:

Open N++ and press Ctrl+f and set up a regular expression search of any text and press _Find Next_.  The search does not need to be successful.
Close N++.
Open N++ and press Ctrl+f and notice that _Match whole word only_ checkbox is enabled (_Search mode_ is still set to _Regular expression_ from the previous search).  At this point it should be disabled.

PR #9792, opened 2021-04-24
#9766 - Fix for 'Match Whole Word' option being enabled in Regular Expression Search

Roshni-P:
Ensured 'Match Whole Word' option is disabled in Regular Expression Search - #9766


sasumner (commented):
Where did _Match whole word only_ go in _Normal_ search mode??

image

I think this change is too simplistic.

PR #9796, opened 2021-04-24
Add MarkAll Preference settings for case and word

### LINKED ISSUE #9751: Style Token not applied consistently within words

sasumner:
Fix #9751 #5148


donho:
image

It's quite wired as behaviour. What do you think @sasumner ?

sasumner:
I did not change any logical behavior with this PR, only the sources for case and word matching.

Before PR: case and word matching were pulled from case and word choices on _Find_ dialog.
After PR: case and word matching are pulled from choices on _Highlighting_ tab of _Preferences_.

User of 9571 wanted decoupling of _Find_ controls -- achieved.
User of 5148 had complaint about case effect before a search was run -- that was addressed by removing _Find_ settings from influence.

Other changes were deemed outside of scope of this PR -- but we can certainly consider them, if desired.

donho:
Good, I create issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9803

Issue #9525, opened 2021-02-16
Feature Request: Move tab to 'end' of the tabs


ggnabasik:
Would like to have an option on the right/click on a tab to move the selected file to the 'end' of the tabs for easy access.  i often find myself having 100 files open and when I click on the file from the project folder view it jumps to the middle of the tabs where I last opened it.  i have to close it and then re-open it, losing my spot in the file, in order to group the 3 or 4 files I'm working with at that time so i can easily jump between them.

sasumner:
So drag-and-drop of tabs is a bit "shaky", but it seems like it would work for such a need?

To each his own, but is it practical to have 100 files open? (no answer needed)

wonkawilly:
I like the idea to have two right click commands one to move the tab to the end and one command to move it to the start in just a click

sasumner:
Here's a PythonScript that, when run, will move the currently-active tab to the right-most position in the collection of tabs:

```
from Npp import notepad

curr_view = notepad.getCurrentView()
curr_index = notepad.getCurrentDocIndex(curr_view)
count = 0
for (filename, _, _, view) in notepad.getFiles():
    if view == curr_view: count += 1
for _ in range((count - 1) - curr_index):
    notepad.menuCommand(MENUCOMMAND.VIEW_TAB_MOVEFORWARD)
```

Obviously to use the script you'd have use the PythonScript plugin...

KnIfER:
What about a quick access entry to the multiline mode? In multiline mode you can drag your tab anywhere and in my [implementation](https://github.com/KnIfER/Textrument/commit/f178cf8b0bf5784e7c5a555f690aece3b4ab288a), when you switch back to the single line mode, the current tab would be centered in the tabbar  automatically. 

Something like this : 

Use hot key (ctrl+shift+s) to open the settings menu, where resides some quick access entry.

![quick_settings_menu](https://user-images.githubusercontent.com/11593903/110232905-20e0e980-7f64-11eb-8144-deb1465532c4.png)

One good point of this ***quick settings menu*** is that it is invoked as a popup around the mouse cursor and you can still acess it in the fullscreen mode.

Here is it in action with 100 files opened : [---link---](https://drive.google.com/file/d/1M01Xg7_8Tt4RRCxlhalI5g2isCVAeKUR/view?usp=sharing)

pchris7489:
[...]

Issue #4437, opened 2018-04-20
Feature request: Possibility to "Replace" without automatically doing "Find next"


P-Si:
### Description of the Issue
I'm doing a fairly complex RegEx "search & replace", and would like to see what my RegEx actually did to the firstly found string before jumping ahead to next matching string in my file. 

### Steps to Reproduce the Issue
Do any "Replace" and after replacing the cursor jumps automatilcally to the next item, so I cannot see what the replace actually did to the found string.

### Desired Behavior
Have a possibility to press (for example) "alt" key while pressing "Replace" button in order to just replace without jumping ahead to next matching string in file

pekka

dinkumoil:
I would like to have such a feature too.

Yyyyes:
Hi.  I would like to have such a feature too.  - I have that same problem over and over again.  Thanks

sasumner:
So a known workaround to this, is after doing your replace (single), switch back to the editor tab and do an Undo than a Redo (use keystrokes).  Your caret will be left where the substitution occurred.  Ideal? Nope.  Workable? Well, I find it to be.

Yyyyes:
I used the workaround for quite some time, but I am fed up with it..  
Not having the text moved on the screen would make it much easier to check the replacement (presently, when coming back, the doc doesn't display at the same line as before, so I have to search for the cursor position too).  Very often, I ask myself "did it do what I expected?,"  And I go back to check...   "  a real pain, with the risk of adding errors in the process.  

What makes it so difficult to split the <replace> button into two button ----- a first<replace & stay// go next if just replaced here> and  --- a second <replace & go to next>?  
For the shortcut keys, the SAME shortcut key should allow, first hit to <replace and stay> and, second hit, if no cursor move has been done, to <go to next>.  
Alternatively, a solution coud be to optionnaly put a delay (1-2 seconds)  in order to view the replacement result before proceeding to next occurence. 
Why is it such an extraordinary request to be able to check the result of the operation before going to the next one  ?   For instance, Textpad does it with a <Replace>  <Replace Next> pair of buttons.

Y.
PS I would also ask for shortcut keys for the <   >>   find next> and < <<   find previous> operations.

sasumner:
>I used the workaround for quite some time, but I am fed up with it.

Here's another workaround that can be handy, but you won't like:  Press Replace All.  The screen doesn't change so you can see your first replacement.  You can undo it like the other workaround and then continue on...

[...]

PR #9798, opened 2021-04-24
Adds ability to move tab to 'end' or 'front' of tabs along with option to 'Replace' without automatically jumping to next occurence

gosseh:
Adds features requested by #9525 and #4437

Added the options when right clicking a tab to move the tab to the front or end of list. Also added a checkbox so that when the user presses 'Replace', the next occurrence is not automatically jumped to.

I use
Windows 10


sasumner:
Please do not combine multiple unrelated issue fixes into one PR as this will not be accepted.

PR #9801, opened 2021-04-26
Add pref setting to allow Replace to stop after replacement

### LINKED ISSUE #4437: Feature request: Possibility to "Replace" without automatically doing "Find next"

sasumner:
Fix #4437

image


donho:
I kinda like this behaviour :)

sasumner:
>I kinda like this behaviour :)

Yea, it's one of those "You have to try it to appreciate it" kind of things. :-)

PR #9800, opened 2021-04-25
Dark Mode independent of OS, find/replace dialog box, more utility code for future dark mode stuff, including buffered paint animation

adzm:
dark mode can be used independent of OS setting. experimental changes on the fly are still iffy with regards to the dark mode menus and scrollbar hack but everything else seems okay.
more utility code subclassing common controls
buffered paint for post-xp themed drawing (animation for checkbox states etc)

Issue #9742, opened 2021-04-07
NppFluentUI Icons for NppDarkMode


WildByDesign:
I created this separate issue so that icon related details don't clutter up the main discussion PR #9587 that @adzm has created. Essentially, I don't want to create any kind of noise in that discussion because it is more programming related whereas this is more image/icon related.

**Microsoft Fluent UI System Icons:** https://github.com/microsoft/fluentui-system-icons
Easy to navigate site for Fluent UI icons: https://bennymeg.github.io/ngx-fluent-ui/

### My first WIP:
![NppFluentIcons](https://user-images.githubusercontent.com/26308319/113876962-3d03c080-9786-11eb-82d8-943ffc882f84.png)

The first WIP NppFluentUI Icons for NppDarkMode was based on the regular Fluent UI icons.

### My second WIP:
![NppFluentIcons-filled](https://user-images.githubusercontent.com/26308319/113877139-6ae90500-9786-11eb-8a7c-402ba2f80259.png)

The second WIP focused on the filled Fluent UI icon type in the hopes that it would be softer on the eyes. I also focused more on the icons in the tab bar as well. I got rid of the constant save button icon that is always present on each tab. Instead, I made it so that the icon shows up on the tab only when changes are made, making it clear to the user that the latest changes have not yet been saved.

### Challenges and Issues:

Part of the reason why I made the change on the tab bar with the saved icon is because of a problem where the saved icon was being colored constantly regardless of the color that I was using. I assume this is related to the following issue mentioned over at (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9587#issuecomment-814289896):

> Unfortunately, since the tab color styling has not really worked in a long
> while due to windows' visual styles, pretty much every theme (including the
> default notepad theme) is using incredibly jarring and garish colors for
> tab styles, like bright red and green backgrounds. Ick.

So the saved icon was being color-shaded red or green depending on whether the tab was in focus or not and I believe that this is something that requires coding changes and cannot be solved with icons in tab bar.

There are also a lot of challenges with scaling SVG to tiny sizes (16x16) and converting to old BMP format. So while I could technically batch convert all icons to make the process easier for myself, the batch converted icons suffered from some imperfections and pixel related issues that show up more at 16x16. So I have to do each icon manually and give each pixel some love.

### Release Challenges and Timing:

[...]

PR #9802, opened 2021-04-26
Dark Mode independent of OS setting; statusbar improvements and dark buttons

adzm:
Still in progress but functional. A bit tricky with some things being able to toggle while Notepad++ is running, but that might be acceptable since we already state changing the experimental options requires restarting the application.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -876,22 +876,12 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			return TRUE;
 		}
 
-		case WM_SETTINGCHANGE:
-		{
-			if (NppDarkMode::handleSettingChange(_hSelf, lParam))
-			{
-				// dark mode may have been toggled by the OS
-				NppDarkMode::refreshDarkMode(_hSelf, true);
-			}
-			break;
-		}
-
 		case WM_INITDIALOG :
 		{
 			if (NppDarkMode::isExperimentalEnabled())
 			{
-				NppDarkMode::allowDarkModeForWindow(_hSelf, true);
-				NppDarkMode::refreshTitleBarThemeColor(_hSelf);
+				NppDarkMode::allowDarkModeForWindow(_hSelf, NppDarkMode::isEnabled());
+				NppDarkMode::setTitleBarThemeColor(_hSelf, NppDarkMode::isEnabled());
 			}
 
 			EnumChildWindows(_hSelf, [](HWND hwnd, LPARAM) {
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -876,6 +876,12 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			return TRUE;
 		}
 
+		case NPPM_INTERNAL_REFRESHDARKMODE:
+		{
+			NppDarkMode::autoThemeChildControls(_hSelf);
+			return TRUE;
+		}
+
 		case WM_INITDIALOG :
 		{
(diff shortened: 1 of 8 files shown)


adzm:
@donho this is what we were hoping for initially - not requiring someone to enable dark mode for the whole system just to get dark mode for Notepad++! This works well as it is now though does not dynamically switch quite right while Notepad++ is running, but never leaves anything in an unusable state. I'm experimenting with other changes but wanted to get this last-functional version up here for now.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -848,8 +849,62 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			return TRUE;
 		}
 
+		case WM_CTLCOLOREDIT:
```
mere-human, line 852: Isn't it the same code as in your other PR? 
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9800/files#diff-62585f441298d642a3dbd871af2800eac3606f9fe148372ab119feed21e050b0

adzm: Yeah, there are a bunch of conflicts if I rebase on the current master, once #9800 is merged I'll rebase this one and mark it as non-draft.

adzm:
FYI this is pretty much ready to go now; unfortunately getting the menus and scrollbar hack to dynamically change between dark and light at runtime is not feasible at the moment, but with these changes we can enable dark mode for the app without needing it set for the OS. Waiting for #9800 and then I can rebase this and get it completed.

adzm:
Merging this into #9800 for simplicity

adzm:
Sorry about the PR open/close song and dance. Rushing to finish this on my lunch break and deal with merge conflicts. Looks like everything is working well now though! @donho ready for review and merge.

adzm:
image

Note the dark buttons in find dialog and the status bar is now also dark there, as well as drawing the size gripper thingy

donho:
# THIS-IS-SO-GOOD!

mere-human:
Looks good!

By the way, @adzm do you have any plans to make it "less dark"?
I mean, use a background color other than black? Or tune it down a little.
For example, Windows Explorer uses grayish colors (except menu bar which is black):
image
The same colors could be used in Notepad++.

The similar case for Visual Studio.
Even GitHub has this setting:
image

adzm:
Yeah there is no reason we can't have a dimmed versus completely dark version. I believe I kept everything in NppDarkMode namespace so it can easily be swapped out mostly for other colors. There was the idea of having them all be themeable like the rest of the colors but that has its own obstacles. At some point this week and/or weekend I'll try to get the new dark mode icons in and see if we can also add a less intense dark mode option too.

adzm:
I am all about picking a dark theme to use for default. I am wary of removing all the other options though since we are using some undocumented things and hooks, figured it might be useful to have those available just in case. But maybe it is worth just keeping it simple as well.

Should we do that stuff in a separate PR? Back when I did this initially I was modifying the style configurator thing to have a separate dark theme section, which is probably what we'd want (a different theme for dark mode vs non).

donho:
> since we are using some undocumented things and hooks, figured it might be useful to have those available just in case. 

We can always keep the different options in `config.xml`. We just remove 3 options from the UI. 

> Should we do that stuff in a separate PR?

I'll merge this PR this night, So please do it in the next PR.

>  I was modifying the style configurator thing to have a separate dark theme section, which is probably what we'd want (a different theme for dark mode vs non).

Then please define this dark theme in the next PR and copy it as `notepad-plus-plus\PowerEditor\installer\themes\darkThemeDefault.xml`, then commit it. So When user switchs to Dark mode via Notepad++ preferences, or via Windows 10 system, `darkThemeDefault.xml` will be used automatically.

It's the same logic for the toolbar and tabbar icons - see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9742#issuecomment-826182763

donho:
I second what @mere-human said with the different (easier) UI way:

image

The slide bar can be 1-3 (3 options) or 1-4 (4 options), so users have choice & it won't waste too much room.

PR #9807, opened 2021-04-27
PHP : Add "curlopt_ssl_verifystatus"

l3afblow3r:


chcg:
See https://www.php.net/manual/en/ref.curl.php

Issue #1314, opened 2016-01-03
no foreground color options in selected text color style


jbrosecity:
I am new to Notepad++. I have been using Textpad for many years but I recently got fed up with them because they have never taken the rather obvious suggestion of allowing one to scroll font size up and down like almost every other Windows application. For aging eyes this is a must. I found Notepad++ does this and a whole lot more. I have spent some time in Style Configurator and am pretty happy with all the options, except that for some reason when I highlight or select any given text the font color or foreground remains the same. In most applications, as in Textpad, when you select text this applies the background color and the foreground goes to white. Even in the Style Configurator window itself, when any item under Style is selected the item is highlighted in blue and the text changes to white. It would be almost impossible to read it if it remained black. Even as I type this request, if I select a word the black font changes to white under the blue selection color. I’m working around this in NPP by using a fairly light color for highlighting, so I can still read the black through it, but I would prefer a darker highlighting with white font or foreground. Hope someone adds this rather obvious option soon.

MetaChuh:
we are closing this issue due to multiple newer notepad++ versions since it was filed.

please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins.
if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as detailed and reproducible as possible.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
search if related topics exists. you are welcome to post either at similar
topics, or to create a new topic at Help Wanted or General Discussion.
```
[...]

PR #9817, opened 2021-04-29
Allow selected text foreground color to be changed

### LINKED ISSUE #3397: Foreground Color of selected text cannot be changed
### Description of the Issue

Notepad++ does not use the standard selected text colors of white text on blue background. Because of this, the user should be able to change both the foreground and background colors of selected text. However, using the Style Configurator  and changing the Foreground Color of selected text does nothing and the color is always black. 

1. Click "Settings" 
2. Click "Style Configurator"
3. Select "Global Style" 
4. Select "Selected text colour"
5. Change the foreground color and save the changes.

### Expected Behavior

The foreground text color of selected text should  change. Alternatively, the default secret text color should be the Windows standard OD white text on blue background. 
 
### Actual Behavior

The foreground text color of selected text does not change.

### Debug Information

Notepad++ v7.4.1   (64-bit)
Build time : May 18 2017 - 02:02:28
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll PluginManager.dll 

The color of the selected text in the npp dialog boxes as shown below is the windows standard and should be available in the npp editor.

sasumner:
Fix #3397 #1314

I really don't know the history (if there is any) on why this wasn't fixed long ago...
Maybe there is some good reason for it?

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2566,24 +2566,26 @@ void ScintillaEditView::performGlobalStyles()
 		execute(SCI_SETCARETLINEBACK, style._bgColor);
 	}
 
-    COLORREF selectColorBack = grey;
-
+	COLORREF selectColorBack = grey;
+	COLORREF selectColorFore = black;
 	i = stylers.getStylerIndexByName(TEXT("Selected text colour"));
 	if (i != -1)
-    {
-        Style & style = stylers.getStyler(i);
+	{
+		Style & style = stylers.getStyler(i);
 		selectColorBack = style._bgColor;
-    }
+		selectColorFore = style._fgColor;
+	}
 	execute(SCI_SETSELBACK, 1, selectColorBack);
+	execute(SCI_SETSELFORE, 1, selectColorFore);
 
-    COLORREF caretColor = black;
+	COLORREF caretColor = black;
 	i = stylers.getStylerIndexByID(SCI_SETCARETFORE);
 	if (i != -1)
-    {
-        Style & style = stylers.getStyler(i);
-        caretColor = style._fgColor;
-    }
-    execute(SCI_SETCARETFORE, caretColor);
+	{
+		Style & style = stylers.getStyler(i);
+		caretColor = style._fgColor;
+	}
+	execute(SCI_SETCARETFORE, caretColor);
 
 	COLORREF edgeColor = liteGrey;
 	i = stylers.getStylerIndexByName(TEXT("Edge colour"));
--- a/PowerEditor/src/stylers.model.xml
+++ b/PowerEditor/src/stylers.model.xml
@@ -1377,7 +1377,7 @@
         <WidgetStyle name="Brace highlight style" styleID="34" fgColor="FF0000" bgColor="FFFFFF" fontName="" fontStyle="1" fontSize="10" />
         <WidgetStyle name="Bad brace colour" styleID="35" fgColor="800000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
         <WidgetStyle name="Current line background colour" styleID="0" bgColor="E8E8FF" />
-        <WidgetStyle name="Selected text colour" styleID="0" bgColor="C0C0C0" />
+        <WidgetStyle name="Selected text colour" styleID="0" bgColor="C0C0C0" fgColor="000000" />
         <WidgetStyle name="Caret colour" styleID="2069" fgColor="8000FF" />
         <WidgetStyle name="Edge colour" styleID="0" fgColor="80FFFF" />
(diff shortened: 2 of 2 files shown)


donho:
It doesn't be done because it's cool to select text w/o losing the syntax highlighting.
I would say "why not" if it's optional that user can change back to keep syntax highlighting on selection...

But no. I prefer to keep the current behaviour.

sasumner:
@donho

OK.  Sure.
At least we have documented the "why" of it.

I (with the gracious help of another person) spent some time yesterday trying to research the "why" and we couldn't find anything.  (Later edit: we should have read issue 3397 comments closer!)
Certainly more time was spent on the research than coding a change!
The "best" (most entertaining) thing that was found was something from 2004 that didn't even seem to answer the question asked:  https://sourceforge.net/p/notepad-plus/discussion/331754/thread/40dc962c/

For myself, I've had this code in place in my `startup.py` for probably 5 years:

```
        for e in (editor1, editor2):
            # set foreground text for selections to be white instead of the default black:
            e.setSelFore(1, (255, 255, 255))
```

My point in mentioning my 5-year-workaround is that I have never -- ever, not one single time -- noticed that while I was selecting text, the lexing colors/styles are not retained on the text that is selected.
Sure enough, if I evaluate it now, they are not retained (surprise!)
If I had noticed this, I likely wouldn't have cared, because IMO it simply isn't that big of a deal.
(But, aha, there are some hints of this behavior if one reads the comments closely in #3397 )

So we have multiple user complaints over the years about not being able to set selected text foreground color; here are only some that I could find:

- https://community.notepad-plus-plus.org/topic/11039/foreground-options-for-selected-text
- https://community.notepad-plus-plus.org/topic/12578/can-t-change-foreground-color-of-selected-text
- https://community.notepad-plus-plus.org/topic/20156/how-to-change-color-of-letters-of-highlighted-text

I have no reason to believe that people are going to stop requesting this, nor stop denegrating Notepad++ because it doesn't support it.

So what are we left with?
FAQ?
Does our FAQ maintainer @ArkadiuszMichalski want to take care of that?

donho:
@sasumner 
The fact that everyone is doing it doesn't mean it's right.

Selecting text without losing the syntax highlighting is advanced feature for Notepad++ certainly, though for some it's not. 
We see some people complain for such behaviour which's come from its 1st release, but if we apply the PR you provided - there will be much more complain (even the insult) coming.

The only way to satisfy both sides is make it as a hidden option - if users want to modify the current behaviour with the foreground colour, they have to put an empty xml file named `enableSelectForegroundColour.xml` beside of `config.xml`. We active this option in the code only if `enableSelectForegroundColour.xml` exists. What do you think?

sasumner:
@donho

>The only way to satisfy both sides is make it as a hidden option - if users want to modify the current behaviour with the foreground colour, they have to put an empty xml file named enableSelectForegroundColour.xml beside of config.xml. We active this option in the code only if enableSelectForegroundColour.xml exists.

How about a different "hidden" option that isn't so development intensive?
What about:

* if user sets selected-text foreground color to the same as selected-text background color, we preserve existing behavior (respect background choice only)
* if user sets foreground different from background, we respect both settings (no lexing in selected text)

mere-human:
> they have to put an empty xml file named `enableSelectForegroundColour.xml` beside of `config.xml`. We active this option in the code only if `enableSelectForegroundColour.xml` exists.

Just a suggestion: what if the option would still be in `config.xml` but won't be exposed in the Preferences UI?
Probably, we have some similar pro-users options.

donho:
@sasumner 
> How about a different "hidden" option that isn't so development intensive?

It's not intensive at all, and it's quite easy to do - otherwise I wouldn't suggest.


> What about:
> * if user sets selected-text foreground color to the same as selected-text background color, we preserve existing behavior (respect background choice only, lexing in selected text)
> * if user sets foreground different from background, we respect both settings (therefore no lexing in selected text)

It could be a way for a power user who has RTFM, but the average users (including me)  will be lost.
If an user change one of these 2 colour by accident, how can he/she knows to set both the same colour to get back the current settings? Not mention that the foreground colour should be always disabled.

Users know what they are doing by adding this empty file, so the other the users who want the current behaviour have no risk to be bothered.

@mere-human 
> Just a suggestion: what if the option would still be in config.xml but won't be exposed in the Preferences UI?

I still think adding an empty file `enableSelectForegroundColour.xml` explicitly to enable this feature is a better solution. The advantage is easier to add (no need to hack config.xml).

donho:
@sasumner 
I will reopen this PR and push some commits into your branch, so you has nothing to do but validate my commits.

e-t-l:
I created enableSelectForegroundColour.xml and the Foreground Color option for Selected Text is still disabled. I am on version 8.6.4. Is there something else I need to do to enable this feature?

PR #9820, opened 2021-05-01
Improve performance of Search results Open all

### LINKED ISSUE #9819: "Open all" from "Search results" window is SLOW
### Description of the Issue
I had need to use the _Open all_ feature in the right-click context menu of the _Search results_ window.
I had previously done a search which resulted in `(3055 hits in 12 files of 31 searched)`.
The largest file containing a hit was 197157 bytes.
Unfortunately I can't provide my fileset as the files contain proprietary data belonging to my employer.

1.  Do a _Find in Files_ search that is representative of the above.
2.  Right-click in the _Search results_ window.
3.  Choose _Open all_ from the context menu.

### Actual Behavior
It took Notepad++ roughly one minute to open the twelve files.
During that minute, the tab bar and the status bar was observed to be flickering.

### Expected Behavior
I expect Notepad++ to be able to open twelve moderately-sized files relatively quickly.
I don't expect to see flickering in the tab bar and status bar areas.

### Debug Information
Notepad++ v7.9.5   (64-bit)
Path : C:\..............\npp.7.9.5.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1909
OS Build : 18363.1440
Current ANSI codepage : 1252

sasumner:
Fix #9819

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -3988,11 +3988,9 @@ void Finder::removeAll()
 
 void Finder::openAll()
 {
-	size_t sz = _pMainFoundInfos->size();
-
-	for (size_t i = 0; i < sz; ++i)
+	for (generic_string path : getResultFilePaths())
 	{
-		::SendMessage(_hParent, WM_DOOPEN, 0, reinterpret_cast<LPARAM>(_pMainFoundInfos->at(i)._fullPath.c_str()));
+		::SendMessage(_hParent, WM_DOOPEN, 0, reinterpret_cast<LPARAM>(path.c_str()));
 	}
 }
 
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3988,11 +3988,9 @@ void Finder::removeAll()
 
 void Finder::openAll()
 {
-	size_t sz = _pMainFoundInfos->size();
-
-	for (size_t i = 0; i < sz; ++i)
+	for (generic_string path : getResultFilePaths())
```
mere-human, line 3991: It's better to declare `path` as `const generic_string&` or `generic_string&&` or even `auto&&` to avoid unnecessary copies.

sasumner: Sure, why not? :-)

PR #9823, opened 2021-05-02
Update basque.xml

izendegi:
I've reviewed all the translated strings and translated some of the untranslated ones.
I'll do another PR with the pending strings once this one gets integrated.


izendegi:
I close this because I've just done another PR with a full translation.:

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9826

PR #9826, opened 2021-05-03
Update basque.xml (full translation)

izendegi:


donho:
@izendegi 

Thank you!

izendegi:
I've checked the latest report and changed the last string changes.

PR #9831, opened 2021-05-06
Fluent UI icons for Dark mode

donho:
When user switches from light mode to dark mode, toolbar icons set switch to Fluent UI icons set in dark mode by default if used icon set is standard one.


PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
```diff
@@ -50,6 +50,28 @@ void IconList::addIcon(int iconID) const
...
+
+	if (hColorInvertedIcon)
+		ImageList_AddIcon(_hImglst, hColorInvertedIcon);
+
+	::DestroyIcon(hIcon);
+
+	return hColorInvertedIcon != 0;
+}
+
+HICON IconList::invertColour(HICON /*hIcon*/) const
```
donho, line 69: @adzm 
I know it's quite naive, but is it possible to implement this function to invert the colour of an icon ?
If the answer is yes, then we don't need to add dark mode icons for dark mode.
Ref our conversation:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9742#issuecomment-826142399

donho: @adzm 
Never mind. After checking [the link](https://github.com/derceg/explorerplusplus/commit/839a3fb36a9084fa4def1e0c6113e86c0a170738) provided by @WildByDesign , I think we will putting a lot of codding effort for not so good quality of icons.

WildByDesign:
@donho  Excellent work, sir.

From my past understanding with inverting icons with code, there always seemed to be a certain degree of quality lost in the sharpness of icons. So in this case, we may have more icons added to the code base, but at least they will retain sharpness and look fantastic. In the future, we can look into ways to use SVG as icons directly.

I have a nice surprise for you later today with all new icons for tab bar (monitoring, read only, saved, unsaved, close buttons, etc.) and I am really please with the outcome now. Dark mode tab bar icons are all done. I just have to invert them and create the light mode tab bar icons and at that point I will be able to provide them to you.

donho:
@WildByDesign 
Thank you. I look forward for your new icons!

donho:
@adzm 
The dark mode is almost ready in the release, it needs a bit more work IMO:
1. In preferences, remove all options but keep only "Enable dark mode".
2. In preferences, add a slide bar to allow users to tune the dark depth.

I can do the first one. Is is possible for you to do the 2nd point? 
I just want to synchronize with what you wanna do, so no pressure.

WildByDesign:
I have completed the tab bar icons for dark mode but I did run into some issues with the way that the light mode versions turned out. I don't want to hold up this PR any longer, therefore I have attached the dark mode tab bar icons to this comment. I will have to finish the light mode versions in the next few days but this way you can continue forward with dark mode implementation now.

I recreated the various closeTabButton icons as well. Now, they will have a better distinction of showing the current tab in comparison to the inactive tabs. I feel that they look much better this way.

Download: 
[FluentUI-tabbar.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6446837/FluentUI-tabbar.zip)

Cheers!

donho:
@WildByDesign Integrated. Thank you.

sasumner:
So the "alternate" tab icons still have some problems (no worries, though, I know this is an "evolution", and in general everything looks FABULOUS) so just wanted to point out:

* the "lock" icon for a readonly file looks a bit strange (see screenshot immediately below)
* a saved file doesn't have an icon? (see screenshot immediately below)

image

The above comments and screenshot is if I start up N++ with _Alternate icons_ enabled.

---------------------------------------------

If I then (N++ still running) turn off alternate icons, I get:

image

where the "non alternate" icons don't respect "dark mode" coloring.
But...since these icons (the diskette) are not distinguished by shape, only color, I don't see how they are going to survive into a dark-mode N++?  Maybe the intent is that they don't survive, and the "alternate" icons become the only icons for saved/modified/readonly?

---------------------------------------------

If I from that point turn alternate icons back on again, I get icons that don't obey the dark colors either:

image

but I'd just like to say that I LIKE THOSE!!!  A LOT!!!  :-)

So anyway, some odd things with these tab-bar icons specifically, but in general I am liking dark-mode a lot!

donho:
> a saved file doesn't have an icon? (see screenshot immediately below)

It was wired at the beginning to me too, then I thought "why not", since the document is in the "clean" mode.

> where the "non alternate" icons don't respect "dark mode" coloring.
But...since these icons (the diskette) are not distinguished by shape, only color, I don't see how they are going to survive into a dark-mode N++? Maybe the intent is that they don't survive, and the "alternate" icons become the only icons for saved/modified/readonly?

We may need a struct to memorize the settings of light mode and dark mode in the future. But I don't want to be bothered
 for the moment. For the users who wanna try dark mode then switch back light mode, they can always reset it back. For those who want to be in the dark mode, they will always stay on the dark side.

> If I from that point turn alternate icons back on again, I get icons that don't obey the dark colors either

Then user can switch to light mode then switch back to the dark mode. The logic of PR is, when user try the dark mode, we suggest the best UX that we (I) believe to users, but they can always override these settings with their own taste after switching in the dark mode. Users usually stay on only one side but if some (or a lot) of them want to switch between 2 mode, we will consider if we do somethings about it (Gray mode or Michael Jaksons mode?).

sasumner:
>>a saved file doesn't have an icon? (see screenshot immediately below)

>It was wired at the beginning to me too, then I thought "why not", since the document is in the "clean" mode.

It's not a terrible idea...

I didn't really understand most of the rest.
But it isn't important right now, I guess.
I was just trying to point out what I saw as inconsistencies.

adzm:
I have almost completed the changes needed here. The slide for the dark
mode intensity is a bit trickier than I thought; I ended up making the
'default' use similar colors to Explorer which is not completely black as
was suggested. I'll get the implementation of the slide tested today or
tomorrow and get it in for review though.

Do you have an easier way to edit resources like dialogs? Visual Studio
doesn't seem to handle the way they are laid out well and I end up having
to do a lot of manual changes to get that working well.

mere-human:
> Do you have an easier way to edit resources like dialogs? Visual Studio doesn't seem to handle the way they are laid out well and I end up having to do a lot of manual changes to get that working well.

I do it this way:
* edit the resource (e.g. a dialog) in Visual Studio to make it look like I want
* copy the part of the involved resource (for dialog it is everything from `IDD_* DIALOGEX` till `END`)
* revert the changes in the corresponding *.rc file
* add the copied RC code to the *.rc file or replace the existing one

This way allows to minimize unwanted changes.

WildByDesign:
> * a saved file doesn't have an icon? (see screenshot immediately below)

This is interesting and something that I have wondered as well. I had already created a small checkmark icon for dark mode that we can easily add if that is desired.

I am torn on this one. On one hand, I really like the clean look without having a whole bunch of redundant icons indicating that documents are saved. It just felt unnecessary. It looks clean without it.

But on the other hand, it also leaves that empty space there which is also unfortunate. It wasn't until after-the-fact that I realized the importance of that space for the other states such as read only, unsaved, monitoring, etc. In my mind, I am still debating this one back and forth.

donho:
@adzm 
> I have almost completed the changes needed here. 

It's quite vague. Could you provide the accurate detail? Or a PR (in progress) will give me a hint. The idea is to prevent me from doing the same things.

> Do you have an easier way to edit resources like dialogs? 

The way that @mere-human suggests is the way to go.

donho:
@WildByDesign 
Let's keep the current empty icon for saved (clean) document for dark mode. It's indeed more simple & efficient to me. We will see the feedback of users.

PR #9836, opened 2021-05-06
Add Ctrl+C and Ctrl+A standard keyboard accelerators to Search results context menu

### LINKED ISSUE #9824: Add keyboard accelerators to Search-results context menu
Add keyboard accelerators (for standard keycombos for common editing commands) to Search-results context menu:

* Add `Ctrl+C` to the menu entry for verbatim `Copy`
* Add `Ctrl+A` to the menu entry for `Select All`

A fairly self-explanatory request, especially with the following images...

In Notepad++ 7.9.5:

image

The requested additions:

image

sasumner:
Fix #9824

Not sure this will be "wanted" but minimal effort expended.  :-)

This would be the addition:

image

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -4320,7 +4320,7 @@ INT_PTR CALLBACK Finder::run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam)
 				generic_string copyLines = pNativeSpeaker->getLocalizedStrFromID("finder-copy", TEXT("Copy Selected Line(s)"));
 				generic_string copyVerbatim = pNativeSpeaker->getLocalizedStrFromID("finder-copy-verbatim", TEXT("Copy"));
 				copyVerbatim += TEXT("\tCtrl+C");
-					generic_string copyPaths = pNativeSpeaker->getLocalizedStrFromID("finder-copy-paths", TEXT("Copy Pathname(s)"));
+				generic_string copyPaths = pNativeSpeaker->getLocalizedStrFromID("finder-copy-paths", TEXT("Copy Pathname(s)"));
 				generic_string selectAll = pNativeSpeaker->getLocalizedStrFromID("finder-select-all", TEXT("Select all"));
 				selectAll += TEXT("\tCtrl+A");
 				generic_string clearAll = pNativeSpeaker->getLocalizedStrFromID("finder-clear-all", TEXT("Clear all"));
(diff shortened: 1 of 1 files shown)


sasumner:
>Okay, what if I have a different hotkey?

What hotkey would this be?
Remember, this is in the _Search results_ window, where hotkeys are not assignable.
Plus, these are standard Windows function assignments:
* Ctrl+c = Copy
* Ctrl+a = Select All

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4319,8 +4319,10 @@ INT_PTR CALLBACK Finder::run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam)
 				generic_string uncollapseAll = pNativeSpeaker->getLocalizedStrFromID("finder-uncollapse-all", TEXT("Uncollapse all"));
 				generic_string copyLines = pNativeSpeaker->getLocalizedStrFromID("finder-copy", TEXT("Copy Selected Line(s)"));
 				generic_string copyVerbatim = pNativeSpeaker->getLocalizedStrFromID("finder-copy-verbatim", TEXT("Copy"));
+				copyVerbatim += TEXT("\tCtrl+C");
```
donho, line 4322: Not sure in the other language, but in German it's rather "Strg-C"

donho:
> So in general, it is an acceptable idea? Just need to figure out the localization part (and not have that be too much burden).

Yes, it makes Copy command more clear. Localization "Ctrl-C" is necessary indeed.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
sasumner: @donho Do you have screenshot of N++ showing "Strg-C" on, for example _Edit_ of main menu?
I can make `notepad.exe` show it that way, but not Notepad++ (still shows "Ctrl-C") while on same screen notepad.exe shows "Strg-C".

donho: I don't have screenshot since I don't use any German system.
But I'm sure @chcg can provide one.

mere-human:
Let me share some screenshots I made on Windows after switching display language to Deutsch (Germany).
Notepad:
image
Notepad++ 7.9.5 portable with its language set to Deutsch:
image

As you can see, the shortcuts aren't localized in Notepad++.
It would be nice to confirm the same on someone's other machine. Maybe, I'm missing something.

But in general, I think, it should be fine to leave this PR as is without localizing the shortcuts.

donho:
@mere-human 
> But in general, I think, it should be fine to leave this PR as is without localizing the shortcuts.

You are right. 

@sasumner I take this PR as it is.

PR #9843, opened 2021-05-08
Update Turkish Localization

Tmp341:
* According to 2e7c5e3, 926e6e9, 6834d79, 874f0d0, 4369718, 1ae39c2, cbf3d2c, 1a9307b, e3dbeda, a0177e8, 0dbdef3, b657f58, 70515c8 and ddd4448 commits.


donho:
@Tmp341 

Thank you!

Tmp341:
What is difference with these?

`<Item id="6352" name="Match case"/>`
`<Item id="6332" name="Match case"/>`

I've translated id 6352 without checking below, if they are not same, i might need to fix one of them.

donho:
@Tmp341 
> What is difference with these?

image

Tmp341:
Oh, they were same (Lower/Upper Case) meaning? Let me fix that.

PR #9844, opened 2021-05-09
Fix Python Function List not showing functions

littlej541:
Functions defined at the top of the file or below a class with nothing in between would not show in the Function List.

This changes the function regex to fix this. A small unit test that the old regex won't pass but the new will included.

Old regex:
oldregex

New regex:
newregex

### DIFF
--- a/PowerEditor/Test/FunctionList/python/baddeftest/unitTest
+++ b/PowerEditor/Test/FunctionList/python/baddeftest/unitTest
@@ -1,0 +1,25 @@
+def invisible_function1(): # invisible because at file start
+    pass
+
+class VisibleClass1:
+    def __init__(self):
+        pass
+
+
+
+
+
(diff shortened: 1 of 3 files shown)


donho:
@littlej541 
I remove all linebreaks between `VisibleClass1` and `invisible_function2` and tested with current version (v7.9.5), but `invisible_function2` is still invisible. Any idea about it?

ArkadiuszMichalski:
@donho  Works for me. Maybe you replaced the regex on the wrong line? It's best to replace the entire `python.xml` file and check.

   
image

donho:
@littlej541 
I put invisible_function2 on the same line of class by adding some white space.  Now I can see it's recognized.

@ArkadiuszMichalski 
I didn't apply the fix, just with the UnitTest to see if remove line breaks can make it works. (it work only without line breaks according the description of provided script)

PR #9868, opened 2021-05-15
Rename sort randomly menu item, move it and reverse lines out of sorting section

sasumner:
Note: no associated "issue".

* Rename "Sort Lines Randomly" on _Edit_ > _Line Operations_ menu to "Randomize Line Order"
* Move "Reverse Line Order" and "Randomize Line Order" so that they appear in the menu above the bottom section (which contains only the true sorting operations).

Screenshot relating to the change are here: https://github.com/notepad-plus-plus/npp-usermanual/pull/220#issuecomment-841648842


sasumner:
Note to self: When this PR is accepted, work on https://github.com/notepad-plus-plus/npp-usermanual/issues/216

donho (approved):

PR #9876, opened 2021-05-16
Add access keys to non-keyboard-accessible menu items

### LINKED ISSUE #9875: Most Edit menu items are inaccessible via keyboard
In some Notepad++ menus, no items have hardcoded access keys (e.g. `View`, `Encoding`, `Language`, `Settings`, `Tools`,...), so you activate them by pressing the key for the first letter of the item (multiple times, if more than such items exist). In `File`, `Edit` and `Search` menus, some items have hardcoded access keys, and others don't; so the underlined letters activate them, no matter the first letters of other items.

I find the interwoven way this is done in the `Edit` menu particularly problematic. E.g. the `Line Operations` item has the most subitems in any Notepad++ menu, most of which don't have keyboard shortcuts; and even if you assign keyboard shortcuts to all those 23 items, it is difficult to remember all 23 of them so you would still need to frequently open the menu. But it's not possible to access the menu by pressing `Alt+E` and then `L`, because doing so activates the `Select All` item, which has the `L` key hardcoded into it. And what's worse, the `Line Operations` item is right on the middle of the `Edit` menu, so the only way to reach it with keyboard is, pressing `Alt+E` and then pressing `Down arrow` 11 times, or `Up arrow` 13 times!

To me, the exact opposite of the current behavior would make more sense, where the mundane items in the first two sections of the `Edit` menu do not have hardcoded access keys, and the items in the later sections should have access keys instead. Alternatively, hardcoded access can be removed from all, so that only the first letters are used to access all items (though 10 items start with `C` in English localization, so those ones would be a bit overloaded). Alternatively, hardcoded access keys can be added to all items.

Here is an example of what those alternatives might look like:

Edit menu access key alternatives
*1 (current menu), 2 (opposite), 3 (none), 4 (all)*

I called the items in the first two sections 'mundane' as I can't imagine there being any significant number of keyboard users who "Select All", by pressing `Alt+E` and then `L`, instead of just `Ctrl+A`. I would assume that everybody who is proficient enough with keyboard shortcuts to know to activate menus with `Alt` keys, is already also familiar with and uses the `Ctrl+A` shortcut, which are standard across all programs. I am sure that there are mouse users who do very much use the `Edit -> Select All` menu item, but they would be using the mouse to select it. (In fact, they wouldn't even know that `L` is the access key for the item, since underlined keys don't appear if you click `Edit` with mouse instead of `Alt+E`, and mouse users wouldn't have bothered with activating the Windows ease of access setting "Underline access keys when available".) Similarly for other hardcoded shortcuts; the `Copy` item blocks access to items like `Convert Case to` which includes many items with difficult to remember/nonstandard shortcuts. And it is hard to imagine anybody using `Alt+E` and then `C` to copy things; they would either use `Ctrl+C` if they prefer the keyboard, or click `Edit` and then `Copy` (or the toolbar button) if they prefer the mouse.

Even though I would personally prefer **2** for this reason, I thought it might be strange not to have access keys only for the initial items (and I was also wary of breaking existing workflows, even though I assume them to be extremely rare), so I made a pull request for **4**. (While at it, I also added access keys to the items in the `File` and `Search` menus that lacked them.) I think the alternative **3** would also be an improvement over the existing situation, if that will be a more acceptable solution. And whichever approach is preferred, I can then modify the `english.xml` file accordingly as well.

debiedowner:
Fix #9875

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -23,26 +23,26 @@ The comments are here for explanation, it's not necessary to translate them.
                 </Entries>
                 <!-- Sub Menu Entries -->
                 <SubEntries>
-                    <Item subMenuId="file-openFolder" name="Open Containing Folder"/>
-                    <Item subMenuId="file-closeMore" name="Close More"/>
+                    <Item subMenuId="file-openFolder" name="Open Containing &amp;Folder"/>
+                    <Item subMenuId="file-closeMore" name="Close &amp;More"/>
                     <Item subMenuId="file-recentFiles" name="&amp;Recent Files"/>
-                    <Item subMenuId="edit-copyToClipboard"  name="Copy to Clipboard"/>
-                    <Item subMenuId="edit-indent" name="Indent"/>
-                    <Item subMenuId="edit-convertCaseTo" name="Convert Case to"/>
-                    <Item subMenuId="edit-lineOperations" name="Line Operations"/>
-                    <Item subMenuId="edit-comment" name="Comment/Uncomment"/>
-                    <Item subMenuId="edit-autoCompletion" name="Auto-Completion"/>
-                    <Item subMenuId="edit-eolConversion" name="EOL Conversion"/>
-                    <Item subMenuId="edit-blankOperations" name="Blank Operations"/>
-                    <Item subMenuId="edit-pasteSpecial" name="Paste Special"/>
-                    <Item subMenuId="edit-onSelection" name="On Selection"/>
-                    <Item subMenuId="search-markAll" name="Mark All"/>
-                    <Item subMenuId="search-markOne" name="Mark One"/>
-                    <Item subMenuId="search-unmarkAll" name="Unmark All"/>
-                    <Item subMenuId="search-jumpUp" name="Jump Up"/>
-                    <Item subMenuId="search-jumpDown" name="Jump Down"/>
-                    <Item subMenuId="search-copyStyledText" name="Copy Styled Text"/>
-                    <Item subMenuId="search-bookmark" name="Bookmark"/>
+                    <Item subMenuId="edit-copyToClipboard"  name="Cop&amp;y to Clipboard"/>
+                    <Item subMenuId="edit-indent" name="&amp;Indent"/>
+                    <Item subMenuId="edit-convertCaseTo" name="Con&amp;vert Case to"/>
+                    <Item subMenuId="edit-lineOperations" name="&amp;Line Operations"/>
+                    <Item subMenuId="edit-comment" name="Co&amp;mment/Uncomment"/>
+                    <Item subMenuId="edit-autoCompletion" name="&amp;Auto-Completion"/>
+                    <Item subMenuId="edit-eolConversion" name="&amp;EOL Conversion"/>
+                    <Item subMenuId="edit-blankOperations" name="&amp;Blank Operations"/>
+                    <Item subMenuId="edit-pasteSpecial" name="&amp;Paste Special"/>
+                    <Item subMenuId="edit-onSelection" name="&amp;On Selection"/>
+                    <Item subMenuId="search-markAll" name="Mark &amp;All"/>
+                    <Item subMenuId="search-markOne" name="Mark &amp;One"/>
+                    <Item subMenuId="search-unmarkAll" name="&amp;Unmark All"/>
+                    <Item subMenuId="search-jumpUp" name="&amp;Jump Up"/>
+                    <Item subMenuId="search-jumpDown" name="Jump &amp;Down"/>
+                    <Item subMenuId="search-copyStyledText" name="&amp;Copy Styled Text"/>
+                    <Item subMenuId="search-bookmark" name="&amp;Bookmark"/>
                     <Item subMenuId="view-currentFileIn" name="View Current File in"/>
                     <Item subMenuId="view-showSymbol"  name="Show Symbol"/>
                     <Item subMenuId="view-zoom"  name="Zoom"/>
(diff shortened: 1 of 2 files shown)


sasumner:
If user selects _English_ in _Preferences_ > _General_ > _Localization_, the changes introduced by this PR are no longer seen.

So the localization box may already be set to _English_ when you go there.  Drop down the box, choose _English_ (even though this makes no sense to do because it was already set to _English_).  Observe the underlined letters from this PR are no longer present.

debiedowner:
@sasumner I know, I mentioned it at the end of the linked issue — though this point relates to this specific pull request, so perhaps I should have mentioned it here, instead of that (already too long) issue. I don't know which of the alternatives, if any, would be preferred; so I didn't want to go through the effort to modify the `english.xml` file with these changes right away.

After making the request, I noticed that GitHub has a draft pull request feature that I probably should have used, to make this more clear. Or maybe I shouldn't have made a pull request at all at this stage until there was some discussion; I thought it would be better if people could test how it behaves rather than just looking at screenshots, and I needed to make the code change to take the screenshots anyway. In any case, once there is any indication that the changes as they exist in this pull request (or any other alternative) are acceptable, I can modify the `english.xml` file accordingly straight away, and add it to this pull request.

sasumner:
> I mentioned it at the end of the linked issue

True enough.  I didn't see it there -- sorry.

>people could test how it behaves

I think that is self-explanatory -- no testing needed. :-)

Are you volunteering to be the _english.xml_ maintainer? :-)

debiedowner:
> Are you volunteering to be the english.xml maintainer? :-)

I am so woefully inexperienced that I was confused by the question, so probably not :-)

My only previous pull request so far, which was modeled after (or even imitated) your earlier similar pull request, added the item to both `Notepad_plus.rc` and `english.xml`, so I assumed that anyone modifying `Notepad_plus.rc` ought to also edit `english.xml` to keep them consistent. I would happily cede that responsibility to the maintainer, if someone is already tasked with doing that. If I can get away with less work, great :) But it seems to me like, whenever someone proposes a change to `Notepad_plus.rc`, asking them to change `english.xml` to be the same would be a simpler solution.

sasumner:
>whenever someone proposes a change to Notepad_plus.rc, asking them to change english.xml to be the same would be a simpler solution.

This presupposes that someone doing such a change cares about what you seem to care about, these little underline thingys.
To do them "right" requires thinking and care.
I know that I don't care about such things, and (likely) would not put that kind of care into adding new menu items.
Unless of course it was "mandated" to do so, but I don't think the author of Notepad++ cares about such things highly, either, or there'd be some sort of rule about it, and maintenance to it...

So currently, if a developer makes a change to a menu, they are supposed to also modify english.xml.
english.xml is what keys the "translators" on what they need to put into french.xml, spanish.xml, etc.
Before a release of Notepad++, the translators are kicked off on this "mission".
That way, all new things that developers have put into english.xml get properly translated.

As a developer, I would prefer to NOT modify english.xml, and have someone watching the other changes for, for example, modifications to `Notepad_plus.rc` and any other source code changes that have a language component, and then that person would modify english.xml appropriately, before it gets into the hands of the other language translators.  This person could serve as a "sanity" check on the appropriateness of the English used by the developer, as well as the details of "does this thing have the proper underlined character in the menus?"

Of course these opinions and thoughts are my own...

debiedowner:
> > whenever someone proposes a change to `Notepad_plus.rc`, asking them to change english.xml to be the same would be a simpler solution.
> 
> This presupposes that someone doing such a change cares about what you seem to care about, these little underline thingys.

Oh, I wasn't talking about access keys, but any change to `Notepad_plus.rc`, adding items, renaming them etc. As you confirmed that my understanding is correct ("*So currently, if a developer makes a change to a menu, they are supposed to also modify english.xml.*"), all I am saying is that I will complete this fix by changing `english.xml` once I know which alternative is acceptable.

But if you meant that "neither @donho nor I care about access keys, you won't get much commentary on the linked issue about the alternatives", that's also fine; I guess than I should go ahead and modify `english.xml` right away with this alternative, and hope that it gets merged. Like I said, I would have preferred alternative 2 if I was just making a build for my personal use, it's more practical; but I think alternative 3 or 4 makes more sense overall, that's why I made a pull request for 4 rather than 2. I guess "we don't care about these" could also mean "let's go with alternative 3, no more underlined keys to maintain", that's also fine by me, though slightly less convenient in everyday use, but much better than the current situation.

By "*Are you volunteering to be the english.xml maintainer? :-)*" if you meant whether I would be willing to maintain the access keys for other menus because others don't care about these, then I guess so. I do care about those very much, I even deliberately use a very customizable web browser (Vivaldi) where I can modify these access keys and thoughtfully assign access keys to items that I use in a way that makes most sense to me. Notepad++ menus are not so easily customizable (except for the right-click context menu), so it matters that it has sensible ones by default. I just wanted to fix the most glaring issue and not alter the existing conventions, that's why I didn't add any access keys to e.g. `View` menu where none existed. But I would be willing to do that, as well as other changes for `Alt` access keys users like me. But the `Edit` menu situation is the most egregious one, that's why I would start with that in any case.

donho (changes requested):
Please integrate the added access keys to `english.xml`

sasumner:
@debiedowner 

>But if you meant that "neither @donho nor I care about access keys"

I thought I made it clear that I speak for no one but myself:  "Of course these opinions and thoughts are my own..."
Please don't infer that anything I say also applies for @donho ; the only thing that can happen with that is that I get into trouble. :-)

I think, if you're inclined to, that you can monitor future changes (turn on all notification emails, or manually watch all master commits) and be a "check" that the underlined characters in menus / dialog boxes / etc. are being done sensically and consistently.

debiedowner:
@sasumner Sorry about misunderstanding your comment referring to *the author of Notepad++*. In any case, Notepad++ has many idiosyncrasies about keyboard navigation that indicates that it hasn't been a priority, or at least that it has been a bit neglected recently; but if there is enough care to at least tolerate my possible future fixes, I would love to make them.

For example, in the `Find` and `Replace` windows, all buttons have access keys, so that e.g. to count the word, I press `Alt+T`. But in the `Mark` tab of the same window, the buttons in the same positions (e.g. `Clear all marks`) have no access keys, so I need to press `tab` key repeatedly to access them. Same for `Find in Files` and `Find in Projects`. Presumably these tabs are more recent additions since the time the last person caring about keyboard access keys stopped caring.

Even more strangely, when tabbing to the aforementioned buttons that lacks access keys, another Notepad++ idiosyncrasy emerges and focus shifts randomly between options on the left and buttons on the right. But I am so averse to using the cursor that I have learned my way around these, e.g. to get to the `Clear all marks` button, 5 `tab` presses. Same problem in the Style Configurator: After choosing a style, I would expect to `Shift+Tab` 2 times to reach the `Save & Close` button at the end, but the button's tab order is weirdly sandwiched between different options on the window, so I can only reach it after 7 `tab` presses. I change the style at least twice a day (daytime and nighttime), so I actually wrote an AutoHotkey script to work around this quirk of Notepad++.

It never occurred to me that I could propose changes to fix them in Notepad++ rather than developing workarounds for them in AutoHotkey. (Although, in my defense, [you are partially the one who set me on this path 5 years ago](https://community.notepad-plus-plus.org/topic/12864/notepad-should-not-fill-the-search-field-with-default-text-when-search-window-was-already-open-and-filled), I noticed just now :D ) I assumed that these quirks were there because they were hard to fix or there was a compelling reason for them to be the way they are; but now, looking at Notepad++ source code, they are trivially easy to fix: adding access keys to those is as easy as adding ampersands to `FindReplaceDlg.rc` and `english.xml` files, and correcting the order is as easy as literally reordering lines in `FindReplaceDlg.rc`! It's just that, evidently recently nobody cared much about keyboard access, just as long as they looked fine on the screen and worked fine with mouse. (Evidently there wasn't anybody caring much about screen reader behavior either; as far as I know, those follow the same order as tabbing on keyboard, and the focus shifting randomly must be jarring for the visually impaired). Apparently I am in a very small minority who absolutely prefers keyboard use for everything. (Or maybe people like me prefer using Vim or Emacs or something? I don't know; I am happy with Notepad++ outside of these few quirks.) In any case, if those other changes will be welcome (which I guess they might be given that this change seems to be accepted), I will be happy to fix them in Notepad++ code rather than working around them in my custom AutoHotkey scripts.

For now, I just added these exact same changes to the `english.xml` file, as @donho asked.

sasumner:
@debiedowner 

I will have a bit more to add later (crushingly busy right now), but I wanted to quickly warn you about the _Find_ dialog in case you are getting to work on it now.

It used to be (not sure of the current state) that any underscore-accelerators had to be unique across all four (now five) tabs of the _Find_ window.  Meaning that you couldn't have (for example, totally made up scenario) Alt+z be one thing on _Replace_ tab and a different thing on _Mark_ tab.

More later...

debiedowner:
@sasumner

When briefly looking at the file, I did notice that most elements (and of course, their access keys) were shared between tabs; but I don't know if it would have occurred to me that non-shared elements should avoid access key clashes, so thanks for the heads-up.

I am not working on it now though, or maybe not even soon; but probably the next time when it bothers me too much while doing repetitive text manipulation...

sasumner:
@debiedowner 

I guess I didn't have that much to say after all..except:

>Presumably these tabs are more recent additions since the time the last person caring about keyboard access keys stopped caring.

It's probably not that, it's probably a throw-hands-in-air-because-all-keycombos-had-to-be-unique-across-4-tabs-and-they're-all-used-so-what's-a-person-to-do situation.  :-)

>in my defense, you are partially the one who set me on this path 5 years ago, I noticed just now

Haha.  Well, it appears that 5 years ago that was how I did such things as well.  I've long since given up those external utilities in favor of a "less external" utility: Pythonscript.

>I will be happy to fix them in Notepad++ code rather than working around them in my custom AutoHotkey scripts.

Go for it.  Create a separate issue and then PR please.

debiedowner:
When modifying `english.xml` to be consistent with `Notepad_plus.rc`, I noticed that there was a capitalization inconsistency (`Jump up/down` in `Notepad_plus.rc`, `Jump Up/Down` in `english.xml`). While not strictly related to this pull request, since I was working on making those lines consistent, I fixed that inconsistency as well, I hope that's OK. It seemed too insignificant to make a separate pull request, and I was modifying both lines in both files in this pull request anyway.

PR #9877, opened 2021-05-16
Update dutch.xml

RockyTDR:
Update Dutch translations according to the following commits to English.xml:
2021-05-16 https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e623e76d0b2ffc1612417993b182b4537f9aca3f (Rename sort randomly menu item, move it and reverse lines out of sorting section)
2021-05-20 https://github.com/notepad-plus-plus/notepad-plus-plus/commit/32dce9b54e40c2cb070f2739fd43a0d905841df3 (Add access keys to non-keyboard-accessible menu items)
~~2021-05-22 **Remove all "amp" shortcuts to enable keyboard access to all menu items using first character of each entry**~~

### DIFF
--- a/PowerEditor/installer/nativeLang/dutch.xml
+++ b/PowerEditor/installer/nativeLang/dutch.xml
@@ -5,7 +5,7 @@ Modifications until 2018-03-26 by Klaas Nekeman (knekeman(at)gmail.com).
 Modifications until 2020-05-26 by xylographe <wr86420@gmail.com>.
 Modifications from 2020-01-28 and onwards by Thomas De Rocker (RockyTDR)
 
-Last modified on 2021-05-16 by Thomas De Rocker (RockyTDR).
+Last modified on 2021-05-22 by Thomas De Rocker (RockyTDR).
 -->
 <NotepadPlus>
     <Native-Langue name="Nederlands" filename="dutch.xml" version="7.9.5">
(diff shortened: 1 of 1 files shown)


sasumner:
>2021-05-22 Remove all "amp" shortcuts to enable keyboard access to all menu items using first character of each entry

That isn't quite as good as providing an underlined accelerator.

As an example, your intent is to "open a folder as workspace".  I'll walk through it in English because I'm not fluent with Dutch.

With what you've done:

* Press and release Alt to highlight `File` menu; press Enter to open that menu
* You press `o` because that is the first letter of `Open Folder as Workspace`
* `Open...` is highlighted by the program
* You press `o` again; `Open Containing Folder` is highlighted by the program
* You press `o` again; `Open in Default Viewer` is highlighted by the program
* You press `o` again; `Open Folder as Workspace...` is highlighted by the program
* You press Enter to execute `Open Folder as Workspace...`

How it would be done with underlined accelerators:

* Press and release Alt to highlight `File` menu; press Enter to open that menu
* You press `w` because that is the underlined letter of `Open Folder as Workspace`
* `Open Folder as Workspace...` is highlighted by the program
* You press Enter to execute `Open Folder as Workspace...`

RockyTDR:
> That isn't quite as good as providing an underlined accelerator.

@sasumner Your arguments are completely valid. However, this was done intentionally. See my post (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9875#issuecomment-846381440) in issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9875 which was handled in PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9876

**Downsides** of removing all underlined accelerators:
- some menu items need multiple key strokes to be activated

**Upsides:**
- **all** menu items are accessible by keystrokes according to the first character of every menu item (Warning: adding only one accelerator in a menu breaks this system!)

**Alternatives:**

- As @debiedowner described, his option `4` would enable all menu items to be accessible by keystrokes, but developers need to keep in mind to add accelerators to all menu items in the source code, while it requires translators to carefully choose the right accelerator character.

Because of the up- and downsides described above, and because users can still create custom shortcuts if they wish, I opted to remove all accelerators.
If the community deems this a very bad decision, I could always roll it back.

sasumner:
This is just my guess, but I would think @donho would want a unified approach, not an "English does it one way, Dutch does it another way" situation.

>but developers need to keep in mind to add accelerators to all menu items in the source code

I suggested that @debiedowner become the "watchman" on that and become the "English translator".
I don't know how well that suggestion was received. :-)

>...because users can still create custom shortcuts if they wish

They _can_ but probably in practice very few users do this.

>If the community deems this a very bad decision, I could always roll it back.

I suppose you'll get your answer on the when @donho accepts or rejects this PR.  :-)

RockyTDR:
> I suppose you'll get your answer on the when @donho accepts or rejects this PR. :-)

I reverted the removal of the accelerators in this PR to prevent it to interfere with the "normal" translations. This can always be done later, in a separate PR 😊
 
> This is just my guess, but I would think @donho would want a unified approach, not an "English does it one way, Dutch does it another way" situation.

I agree! This should be discussed further to reach a well thought out decision.

donho:
> This is just my guess, but I would think @donho would want a unified approach, not an "English does it one way, Dutch does it another way" situation.

For me each access key should be unique for access the menu item more efficient. Otherwise, just don't put it.

sasumner (approved):

sasumner:
>>This is just my guess, but I would think @donho would want a unified approach, not an "English does it one way, Dutch does it another way" situation.

>For me each access key should be unique for access the menu item more efficient. Otherwise, just don't put it.

The aspect in question was : English is full of underlined characters in menu items; Dutch wanted to not have any at all.  My thought was: Either all have them or all not.  That's all.

donho:
@sasumner 
> The aspect in question was : English is full of underlined characters in menu items; Dutch wanted to not have any at all. My thought was: Either all have them or all not. That's all.

I don't think so. For the less frequent commands (though it's subjective) it's possible to not have access key. However, for the menu entries of the 1st level (File, view, ...etc), we should have an access key for each to be able to access the sub-menu & menu command.

RockyTDR:
> For the menu entries of the 1st level (File, view, ...etc), we should have an access key for each to be able to access the sub-menu & menu command.

This is not strictly needed, I think. When removing all access keys, typing the first character of the menu item changes focus to that menu item (this works even in the menu bar).

PR #9906, opened 2021-05-25
Add access keys to more Find/Replace dialog items

### LINKED ISSUE #9886: Add access keys to more Find/Replace dialog items
I was working on adding access keys to more Find/Replace dialog items, paying attention not to use the same access keys on different tabs, in light of @sasumner's warning in #9876. I was going to make a very limited pull request due to this constraint (The only available and usable keys are `&j` and `&k`; remaining `&q`, `&v` and `&z` are also available but not usable). But upon experimenting, I observed no clash when same access keys are used in different tabs. Looking at old issues, this problem indeed seems to be a thing as recently as 2019. I don't know what changed, but it is no longer the case in my testing.

Actually, this is easy to test without making any changes: @donho's fix 8fd691a for this problem only changed `FindReplaceDlg.rc`, not `english.xml`; so if you temporarily set your locale to English, both `Find All in All Opened Documents` in the `Find` tab and `Replace All in All Opened Documents` in the `Replace` tab have the underlined access key `&O`, but they work fine. Pressing `Alt+O` while in the `Find` tab correctly finds all in all opened documents, and pressing `Alt+O` while in the `Replace` tab correctly replaces all in all opened documents. I thought maybe the problem only happens for `FindReplaceDlg.rc`, but after modifying that and building, it still works fine.

Does either of you or anyone else have any insight or warning to share about this? Something like *oh, we fixed it* or *it sometimes works and it's sometimes buggy, we don't know what causes it so don't mess with it*? If not, I will assume that this was just a bug due to some component in the old build process and no longer something we need to worry about, me and other language translators. Because I agree with Swedish translator's comment in the issue that donho's commit fixed (#2276):

> Clashes between Alt shortcut letters from invisible tabs should not be an issue, i.e. Notepad++ should never, in any dialog with multiple tabs, have shortcuts on invisible tabs swallow Alt letter shortcuts that appear on the active tab. It's not realistic to require that all Alt shortcuts be unique across all tabs. Being unique on the active tab should suffice and work as expected.

So if may be permitted to assume that now it will work as expected, I will add new access keys to items accordingly in a pull request; and also revert the existing ones to use the more intuitive keys instead of the workarounds due to this earlier bug, e.g. in the case of the aforementioned inconsistency between `english.xml` and `FindReplaceDlg.rc`, I will use the access keys of `english.xml`, thereby reverting donho's commit fixing this bug.

debiedowner:
Fix #9886

Personally I mostly wanted an access key for "Clear all marks", and added access keys to other items while at it. I didn't add access keys to the first buttons and Close buttons, as they can accessed by using the `Enter` and `Escape` keys (at least that's presumably the reason they lacked access keys, and I didn't change this).

Thanks to #8760, I was able to add the same access key to similar buttons on different tabs, e.g. `Replace`, `Replace in Files`, `Replace in Projects` can all be activated using `Alt+R`. I did the same thing for a few existing access keys; e.g. `Find All in All Opened Documents` and `Replace All in All Opened Documents` now share the access key `Alt+O` (i.e. reversing the earlier fix 8fd691a), `Find All` and `Replace All` share the access key `Alt+A` etc. In addition to consistency, I decided to do this because it uses more intuitive access keys instead of something caused by a workaround due to an earlier bug, and because it is unlikely to disrupt people's workflows as they are already used to using that access key when they performed a parallel function on a different tab, and because it is easier to keep in mind that buttons performing similar functions on different tabs share the same access keys. (This was already the case with most controls (e.g. `Match case` is `Alt+C` in all tabs); there is no reason various `Replace` buttons in different tabs not have the same access key `Alt+R` other than the earlier technical limitation.) But, if preferred, I can reverse this change and limit this pull request to just adding new access keys.

Another thing: While the fix at #8760 mostly works fine, one quirk is that if controls with checkboxes (rather than push buttons) share access keys in different tabs, it doesn't work fully as expected: it activates the control just fine, but it doesn't flip the checkbox; another `Space` press is required for that. To avoid this annoyance, I made this pull request paying attention to keeping the checkbox access keys unique across tabs. Push buttons not having any problems with non-unique access keys allowed free up unique access keys for checkboxes. E.g. since `Replace All in All Opened Documents` now uses `Alt+O` and not `Alt+U`, I was able to assign `Alt+U` to `In all sub-folders` uniquely. This is another justification for making the few minor changes to existing access keys that I did.


debiedowner:
And another thing: I found the inconsistent language in the `Find` and `Replace` tabs strange. As the last button, both tabs have the `Find All in All &Opened Documents` and `Replace All in All &Opened Documents`; which is fine, the phrasing is the same and I made the access keys same as well. But right above that, the buttons are named `Find &All in Current Document` and `Replace &All`. Again, I made the access keys same, but why the different phrasings? Either the clarification is unnecessary, so they can be named `Find &All` and `Replace &All`, or the clarification is necessary, so they should be named `Find &All in Current Document` and `Replace &All in Current Document`.

I didn't make a change in the text as it goes beyond access key changes I wanted to make in this pull request, but if it is acceptable, I could modify `Replace &All` to `Replace &All in Current Document` and adjust the button size accordingly. (Similarly, I could instead change `Find &All in Current Document` to `Find &All`, but that would cause the button text to be the same as `Find in Files` / `Find in Projects` buttons; maybe that's what this phrasing was seeking to avoid?) Or maybe making text consistent across tabs doesn't matter much, in which case this pull request can be merged as is, if everything else is OK.

sasumner:
@debiedowner 

Have you developed and tested this PR in a branch that contains this change?:  https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6c19e882424e3f1471b315c7970b1fb21fd79dda
I'm just curious if that change has any impact; ideally it shouldn't.

>I didn't add access keys to the first buttons and Close buttons, as they can accessed by using the Enter and Escape keys

By "first" buttons I think you mean the "default" command buttons.

Presumably this button (in yellow) also got no underlined accelerator:

image

>Find All in All Opened Documents and Replace All in All Opened Documents now share the access key Alt+O

I can see users complaining that they meant to do a find but actually did a replace in many files, because they weren't on the tab they thought they were on when they pressed Alt+o.  A PBKAC for sure. though.  And they'd have to have disabled the confirm on RAOD for this to do any real harm.

>one quirk is that if controls with checkboxes (rather than push buttons) share access keys in different tabs, it doesn't work fully as expected: it activates the control just fine, but it doesn't flip the checkbox; another Space press is required for that. 

The "quirk" was first noted here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8696#issuecomment-700261217

Regarding captions on buttons:

* using _Replace All in Current Document_ is kind of a good suggestion, although, due to the past user confusion about this button not truly replacing "all" (unless _Wrap around_ is also ticked), I personally would rather see something like "Replace All in Current Direction" if there is to be a change.  Of course that maybe makes things unclear if _Wrap around_ is actually ticked.  A digression: Even _Wrap around_ is bad in this case, because what it really is implying is "Affect entire document".  Conclusion: UI design -- a battle with winners and losers.

* _Find All in Current Document_ should remain as it is.  Changing it to _Find All_ is problematic; as you point out that would create two buttons on different tabs with the same caption.  This is not a technical problem but rather a support issue.  Users sometimes aren't very specific in their bug reports, so I could see reports of "Find All not working", leading to further back and forth to establish more about what they mean.

* Unfortunately, I now notice that the _Find in Projects_ tab uses _Find All_ on its button.  Sigh, I did not implement that feature, nor review it on a deep level.  This one is pretty glaring.  :-(  

* Conclusion: If I were you, I wouldn't attempt changing button caption text. :-)  But, do as you will.

debiedowner:
> Have you developed and tested this PR in a branch that contains this change?

I did, and I just retested, but no issues related to that change.

> Presumably this button (in yellow) also got no underlined accelerator:

Right. I could have added an accelerator to it, but it would be of no help unless you have a ▲ character available on your keyboard layout. And I didn't want to add/change text just to add an access key; that's also the reason why the directory selector in `Find in Files` doesn't have an access key. You need to use the access key for the `Directory` field and then tab to it. (I could add `.` as an access key, but that's already being used by `. matches newline`; I could in turn try to change that's access key, but again, I didn't want to make too much changes.

In any case, like the "access key" for the default buttons is `Enter`, the `▲` button also has a sort of access key, `Shift+Enter`, and I imagine most keyboard users use that. I don't even have the `▲` button displayed (the checkbox to its right is unchecked) since I never use mouse to click them anyway.

> because they weren't on the tab they thought they were on when they pressed Alt+o.

This is the nature of access keys / accelerator keys, being context dependent; e.g. pressing `Alt+E` (or just `E`) closes all your tabs if you are in the File menu and opens `EOL Conversion` if you are in the Edit menu. Or pressing `Enter` obviously has vastly different actions depending on the dialog that is open. If anything, that is their advantage, compared to the keyboard shortcuts which mostly need to be unique across the application. They are just a shortcut to press the button visible on the screen; if the user uses the mouse to just blindly click the button at that location without paying attention to where they are, they will have the same problems.

My main concern in having to change a few existing access keys was breaking people's habits; that's why I considered this commonality an advantage rather than a problem. E.g. in that example, users are already used to pressing `Alt+O` to `Find All in All Opened Documents`, so there should be no difficulty in adjusting to press `Alt+O` to `Replace All in All Opened Documents`; this is how it was intended before it had to be changed due to the bug, and it is probably more in line with what new users would expect anyway. I certainly appreciate that `Enter` key is common in all dialogs and I don't need to look for a separate key on each dialog. If the action is too dangerous or destructive, maybe it shouldn't have any access key at all, but like you said, an additional confirmation dialog takes care of this better.

> Regarding captions on buttons:

Thanks for the feedback; I was just too allured by the possibility of symmetry between the two tabs, but it doesn't really matter and the current ones are probably perfectly fine.

donho:
For me, there's no reason to change the current access keys in case of no clash. A lot of people use them since long time.

PR #9921, opened 2021-05-29
Invert plugin toolbar icon color

donho:
Plugin can use NPPM_ADDTOOLBARICON for adding the command & icon on toolbar:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h#L148

This PR try to invert ICO format (provided by plugin) color for matching with dark theme.

### DIFF
--- a/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
+++ b/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
@@ -122,11 +122,30 @@ void ToolBarIcons::reInit(int size)
 		_iconListVector[HLIST_DEFAULT2].addIcon(i._hIcon);
 		_iconListVector[HLIST_DISABLE2].addIcon(i._hIcon);
 
+		ICONINFO iconinfoSrc;
+		GetIconInfo(i._hIcon, &iconinfoSrc);
+		HBITMAP hBmpOld = iconinfoSrc.hbmColor;
 
-		_iconListVector[HLIST_DEFAULT_DM].addIcon(i._hIcon);
-		_iconListVector[HLIST_DISABLE_DM].addIcon(i._hIcon);
-		_iconListVector[HLIST_DEFAULT_DM2].addIcon(i._hIcon);
-		_iconListVector[HLIST_DISABLE_DM2].addIcon(i._hIcon);
+		HDC dcMem = ::CreateCompatibleDC(NULL);
+		HDC hDestDc = ::CreateCompatibleDC(NULL);
+		BITMAP bmp;
+		::GetObject(hBmpOld, sizeof(BITMAP), &bmp);
+		::SelectObject(dcMem, hBmpOld);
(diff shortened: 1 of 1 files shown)


donho:
@adzm 

The following picture is the current state of this PR:
image

The first ICO is provided by DSpellCheck plugin, the 2nd visible ICO provided by NppPluginDemo:
http://download.notepad-plus-plus.org/repository/MISC/NppPluginDemo.7z

I did expect the icon keep its transparency and invert the color so plugin authors don't need to provide the dark mode ICO. 
If you have any idea to improve the result (of this PR), please let me know.

adzm:
Oh this is an interesting one. I remember dealing with a lot of
ico/dib/bitmap internals back when Vista came out with the png icons. My
hunch is that there is an issue with the mask vs transparency. I'll check
out those plugins and see if I can figure something out.  IIRC I ended up
drawing the icon with DI_MASK and having to check for alpha channel info in
the actual bitmaps in order to get things to work consistently on those
with alpha and without.

donho:
@adzm 
Feel free to create your PR so I'll use your solution instead of this PR.

adzm:
I'm messing with this now and still am trying to figure out hicon and mask
workings but I'm optimistic.

However instead of inverting the colors I am thinking I should invert the
brightness. That way a dark green check becomes a light green check and
stays green instead of turning pink.

adzm:
Do you know a project that uses the bitmaps instead of ico? I wonder if we can figure out a way to get it to work with those, too, at runtime somehow.

donho:
@adzm 
Thank you for the help!

> However instead of inverting the colors I am thinking I should invert the
brightness. That way a dark green check becomes a light green check and
stays green instead of turning pink.

Since all the icons (except standard BMP) are changed to the mono color (black in light mode, white in dark mode), I don't think it's interesting to transform from darker green to lighter green or somethings similar. The current state of most plugins just don't support ICO format, so these plugins have to be updated by adding ICO format to display the icons while one of Fluent UI icons is chosen. In this case, plugin authors will (hopefully) align with the color of Fluent UI - so dark or white.

> Do you know a project that uses the bitmaps instead of ico? I wonder if we can figure out a way to get it to work with those, too, at runtime somehow.

Notepad++ uses BMP to display the standard icons (and ICONS on Fluent UI by using imagelist):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/ToolBar/ToolBar.cpp#L327

Otherwise, @WildByDisgn has provided a link which may be useful to you:
https://github.com/derceg/explorerplusplus/commit/839a3fb36a9084fa4def1e0c6113e86c0a170738

adzm:
#9935 is a fixed implementation of this; it inverts the lightness rather than the colors but also works properly with the masks and everything. @donho I think this is a pretty good approach for those who do not want to create their own dark mode icons. If other plugins are still registering toolbar icons via a bitmap we could use this same approach to generate an icon from that as well, even inverting lightness like this one, so that way backwards compatibility remains intact.

PR #9925, opened 2021-05-30
Revert BrandingText to follow code of conduct

yakahack:
The BrandingText is the text shown on the license page on the installer


donho:
I prefer the current one.

PR #9927, opened 2021-05-31
update japanese.xml to v8.0

yasmise:
Update Japanese translation. Corresponding commit hash of english.xml is shown in each commits.

### DIFF
--- a/PowerEditor/installer/nativeLang/japanese.xml
+++ b/PowerEditor/installer/nativeLang/japanese.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <NotepadPlus>
-    <Native-Langue name="Japanese" filename="japanese.xml" version="7.9.6">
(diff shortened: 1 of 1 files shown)


donho:
@yasmise 
Thank you for the update.
Are these access key you added conventional comparing with all Japanese applications?

yasmise:
@donho
I'm not sure I understand correctly, but here's my answer.

According to my experiences, it seems like most applications which are translated from English to Japanese use same access keys to its original English version. So, access keys added in this PR is same to English version of Notepad++.

If you are referring `Copy (&amp;C)` style, yes, it is conventional. I'm sure most applications use this style.

donho:
@yasmise 
> According to my experiences, it seems like most applications which are translated from English to Japanese use same access keys to its original English version. So, access keys added in this PR is same to English version of Notepad++.

Yes, that's my concern. Thank you for your confirmation.

Issue #8922, opened 2020-09-29
Add possibility to close all panels in stack with single action


ArkadiuszMichalski:
### Description of the Issue
After this bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8471 (PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8753) there is no way to close all stack of panels with single command/action. The previous behavior has been around for years and some may get used to it, like @Uhf7 https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8753#issuecomment-683467400 wrote. 

I know it's enough to click [x] a few times ... but I would prefer once (because such an option existed and it is no longer there).

Overall this solution with [x] is better than previouse, but now there is no option to close all :) Maybe with Ctrl (or Shift) + [X] or other key can (even middle buton) do it?

PS, Even for undocked panels when we go to context menu >> Close Alt+F4 command only one panel is close. but when I press Alt+F4 it close all Notepad++ (weird).

### Steps to Reproduce the Issue

1. Open few panels in one docked area.
2. How close all with one action?

Generally, we can live without this option, but if it was available that would be fine too. If this feature will be rejected, then this description will be as reference to any future same questions.

### Debug Information

Notepad++ v7.9   (32-bit)
Build time : Sep 22 2020 - 03:24:22
Path : D:\_test\NPPClean\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Ultimate (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1250
Plugins : none

Uhf7:
> 2. How close all with one action?

I have nothing against an additional function for closing all panels at once, the question still is: Shift-Click or configurable?

> PS, Even for undocked panels when we go to context menu >> Close Alt+F4 command only one panel is close. but when I press Alt+F4 it close all Notepad++ (weird).

That's indeed a thing to go for, if you press Alt-F4 in MS VC, while a floating side window has the focus, then the entire MS VC closes. This should be implemented this way in Notepad++ too, I will go for it.

A click on the close icon of a floating window should behave the same way as the click on the close icon in the docked window, this part is already right and will not change with implementing the additional "close all" function.

ArkadiuszMichalski:
Firstly, if @donho accept such possibility we can deliberate:), otherwise it's a pity to waste time (except for unofficial releases). If yes then...

@Uhf7
> Shift-Click or configurable?

[...]

PR #9930, opened 2021-05-31
Close all tabs in stack with single action

Uhf7:
Implements #8922. Shift-click closes the whole stack now.

I included the "Close" tooltip in localization, but I didn't  extend this tooltip yet for the following reason: If the side window is not docked, it has an original Windows title bar. The tooltip for the Close button in this title bar is always "Close", and cannot be modified in any easy way. So, if we extend the tooltip in the docked state only, we have two different tooltips, docked and undocked.

This could be overcome only by drawing an own title bar for the undocked side windows. In this case, the title bars of docked and undocked side windows could made look the same way. But I did not any experiment in this direction right now.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1436,6 +1436,7 @@ Find in all files except exe, obj &amp;&amp; log:
             <session-save-folder-as-workspace value="Save Folder as Workspace" />
             <tab-untitled-string value="new " />
             <file-save-assign-type value="&amp;Append extension" />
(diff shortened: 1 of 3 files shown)


donho:
OK, let's keep the current way.
Does "Close" tooltip in localization work in both docked & undocked ways?

Uhf7:
> Does "Close" tooltip in localization work in both docked & undocked ways?

Not in the same way. If docked, the localization of Notepad++ works. If undocked, the localization mechanism of Windows works. So it may end up in different tooltips, if both localizations are not synchronous.

PR #9935, opened 2021-06-02
Invert plugin toolbar icon lightness for dark mode

adzm:
Alternative implementation of #9921 that inverts lightness. This can be updated to invert the actual colors if desired by simply setting the RGB components to 255 - R etc instead of using the HSL lightness inversion.

### DIFF
--- a/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
+++ b/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
@@ -16,8 +16,9 @@
 
 
 #include <stdexcept>
+#include <memory>
 #include "ImageListSet.h"
-
+#include "NppDarkMode.h"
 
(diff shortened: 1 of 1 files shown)


adzm:
I think this is a pretty good approach for those who do not want to create their own dark mode icons. If other plugins are still registering toolbar icons via a bitmap we could use this same approach to generate an icon from that as well, even inverting lightness like this one, so that way backwards compatibility remains intact.

adzm:
While RGB inverting will work out fine for grayscale, it does for example turn green into pink. Using HSL and inverting the lightness preserves the hue (green etc) while simply inverting the lightness, which while not perfect will at least keep colors of any icons intact, and works just as well for gray values.

donho:
@adzm 
Awesome! It does work like a champ!

Issue #9922, opened 2021-05-30
Make Notepad++ compilable with VS2019


donho:
Currently Notepad++ is built with VS2017.
It's time to switch to VS2019 for Notepad++ future versions.

donho:
@chcg 
I guess you have tried it and made it. If yes, could you create a PR for it please?
I did try, but I didn't make it.

mere-human:
I use VS2019 to build Notepad++. It works pretty well.  I've heard @sasumner uses VS2019 as well.
To do so, I simply installed VS2017 build tools in VS2019:
![image](https://user-images.githubusercontent.com/9664141/120101749-3d9a5100-c150-11eb-8caa-c16c68c72c5e.png)

To use VS2019 by default, I think, one has to switch the Platform Toolset.
![image](https://user-images.githubusercontent.com/9664141/120101681-ebf1c680-c14f-11eb-92d1-fc96d7d87101.png)

sasumner:
What I do (and what I've always done since I never had VS2017):

* change _Platform Toolset_ (as @mere-human mentions) -- but I don't install v141, I just change the setting to v142
* change _Treat Warnings As Errors_ to _No_

Perhaps if I'd ever installed v141, I would not have had to change the warnings-as-errors setting?

donho:
@mere-human @sasumner
Thank you for the reply. 
I suppose that you install VS2019 upon your VS2017? 
I download VS2019 community in my new laptop, and my settings screen (for SciLexer) has less options:  
![image](https://user-images.githubusercontent.com/90293/120200421-9b986880-c224-11eb-95f3-80df548830be.png)

And it gives me the following error while I compile Scintilla:
![image](https://user-images.githubusercontent.com/90293/120201089-54f73e00-c225-11eb-9c67-3d948a32a3c8.png)

v142 seems work for @sasumner so I guess I'm missing somethings in my installation...

donho:
Here's the other screen shot of settings:
![image](https://user-images.githubusercontent.com/90293/120201801-1f068980-c226-11eb-9ea0-412681187b07.png)

donho:
@mere-human 
Indeed, Toolset **v141** DOES make my Notepad++ built under VS2019 !

> change Platform Toolset (as @mere-human mentions) -- but I don't install v141, I just change the setting to v142

[...]

Issue #9479, opened 2021-02-03
notepad++ always opens on first desktop


laurijssen:
### Description of the Issue
When you have multiple desktop on windows notepad++ always opens on the first.
It should open a new instance on the current active desktop

### Steps to Reproduce the Issue
1.  Start up notepad++
2. Create a new desktop (winkey + tab => click '+'
3. Go to new desktop (winkey + right arrow)
4. Start up notepad++

### Expected Behavior
Expect a new instance of notepad++ to start on the second desktop

### Actual Behavior
Windows switches to the first desktop and puts notepad++ on the foreground

mere-human:
It seems to me, that the Notepad++ rather should remember its last position be it n the first desktop or on the second one.

laurijssen:
> 
> 
> It seems to me, that the Notepad++ rather should remember its last position be it n the first desktop or on the second one.

Ok, so not consider an improvement? 
My habit is that when working on multiple projects to divide them over desktops. EG a python project on the first and a C# on the second. With Visual studio (+ Code) it works fine but notepad only opens all files in one instance always

mere-human:
> Ok, so not consider an improvement?

It was just an opinion.

By the way, have you tried changing this option?
![image](https://user-images.githubusercontent.com/9664141/106725015-8febbb00-6611-11eb-9d7d-a43b45e1a64b.png)
If you work with multiple projects you might want to keep them in different instances.
I tried this, and a new instance is opened on the currently active desktop.

laurijssen:
That's it indeed

laurijssen:
Reopened because with multi-instance enabled notepad++ open a full instance for every file even when on one desktop.
In my opinion it should work that for every desktop there is one instance and on the desktop where a file is opened it's added to the list of open files on that desktop instance.

carlosvsilva:
I would also much prefer if Notepad++ respected the current Windows virtual Desktop open and by default opened a new file in the current desktop instead of switching to the Notepad++ window open on a different virtual desktop and adding a file to that other window.

Please make Notepad++ aware of Windows virtual Desktops, they're really useful to keep work and personal stuff separate, in the age of remote working and screen sharing. Thanks!

q--:
Linking the various duplicate issues to hopefully make it easier to keep track of things: #12097 #11345 #11489

Issue #6225, opened 2019-10-15
The app should appear on the a single virtual desktop by default


andkirby:
### Description of the Issue

By default all apps appears on a one Windows 10 Virtual Desktop ([Virtual Desktop manual](https://www.howtogeek.com/197625/how-to-use-virtual-desktops-in-windows-10/)).
But NPP always appearing on all ones by default.
It should not work in this way be default.

### Steps to Reproduce the Issue

1.  Create an Virtual Desktop ([manual](https://www.howtogeek.com/197625/how-to-use-virtual-desktops-in-windows-10/))
2. Run NPP
3. Move to another Virtual Desktop

### Expected Behavior

The app stays on the initial virtual destop.
Another Virtual Desktop don't have launched NPP.

### Actual Behavior

Appears on all virtual desktops.

### Debug Information

Desktop # 1
![Desktop # 1](https://content.screencast.com/users/andKirby/folders/Jing/media/95e8aa9c-dbae-4d03-bbe8-c9d986b5b9ca/2019-10-15_1751.png)

Desktop # 2
![Desktop # 2](https://content.screencast.com/users/andKirby/folders/Jing/media/c0d0a801-4ef5-403e-b714-464537c494a5/2019-10-15_1751.png)

Result after `WIN`+`TAB`:
![](https://content.screencast.com/users/andKirby/folders/Jing/media/21926939-33c1-40dd-a599-f9a1f0ae0a0e/2019-10-15_1759.png)

Nu11u5:
This seems to be a bug as this is not normal behavior for any app.  If this can be made an (optional) intentional feature that would be interesting.

andkirby:
@Nu11u5, yeah, true.
I was able to disable that by some application. But it's really overhead to use it always when you run NPP.
So, there is definitely enabled redundant option for this app.

PR #9936, opened 2021-06-02
Windows 10 Virtual Desktop Support

ameisen:
This is an initial, basic implementation of support for Windows 10 Virtual Desktops.

Presently, you can either:

1. Specify that Notepad++ should be in multi-instance mode, where every time you open a file it is a new instance.
2. Specify it should be in 'session' mode, where sessions are in their own instances.
3. Specify that it should be in single-instance mode, where there is only one instance of Notepad++.

None of these work well with virtual desktops.

This adds a new checkbox to mark awareness of virtual desktops. When it is checked (and it is by default) then if Notepad++ were to prevent itself from starting due to Notepad++ already running, it will now *only* do so if any of the already-running Notepad++ instances *are running in the same virtual desktop as the user presently*. 

This way, you won't be jumped around virtual desktops just because you opened a file, which is incredibly jarring.


PowerEditor/src/NppCommands.cpp
```diff
@@ -487,7 +487,7 @@ void Notepad_plus::command(int id)
 			if (nppGui._searchEngineChoice == nppGui.se_custom)
 			{
 				url = nppGui._searchEngineCustom;
-				remove_if(url.begin(), url.end(), _istspace);
+				static_cast<void>(std::remove_if(url.begin(), url.end(), _istspace));
```
ameisen, line 490: Visual Studio 2019 refused to compile due to `remove_if` being annotated as `nodiscard`.

PowerEditor/src/NppNotification.cpp
```diff
@@ -280,10 +280,9 @@ BOOL Notepad_plus::notify(SCNotification *notification)
...
 
 					HWND hWinParent = ::GetParent(hWin);
-					const rsize_t classNameBufferSize = MAX_PATH;
-					TCHAR className[classNameBufferSize];
-					::GetClassName(hWinParent,className, classNameBufferSize);
-					if (lstrcmp(className, _pPublicInterface->getClassName()) == 0 && hWinParent != _pPublicInterface->getHSelf()) // another Notepad++
+					TCHAR className[_countof(Notepad_plus_Window::ClassName) + 1];
+					::GetClassName(hWinParent, className, _countof(className));
+					if (generic_strncmp(className, Notepad_plus_Window::ClassName, _countof(className)) == 0 && hWinParent != _pPublicInterface->getHSelf()) // another Notepad++
 					{
```
ameisen, line 286: I believed that it was best to use the same mechanism of checking if a window was a Notepad++ window in both locations; only checking against a string that is just one character longer than the expected value suffices; it will either be a null-terminator, or it will be different.

PowerEditor/src/NppNotification.cpp
```diff
@@ -280,10 +280,9 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 					fileNamesData.cbData = long(quotFileName.length() + 1)*(sizeof(TCHAR));
 
 					HWND hWinParent = ::GetParent(hWin);
-					const rsize_t classNameBufferSize = MAX_PATH;
-					TCHAR className[classNameBufferSize];
-					::GetClassName(hWinParent,className, classNameBufferSize);
-					if (lstrcmp(className, _pPublicInterface->getClassName()) == 0 && hWinParent != _pPublicInterface->getHSelf()) // another Notepad++
+					TCHAR className[_countof(Notepad_plus_Window::ClassName) + 1];
```
ameisen, line 283: I believed that it was best to use the same mechanism of checking if a window was a Notepad++ window in both locations; only checking against a string that is just one character longer than the expected value suffices; it will either be a null-terminator, or it will be different.

ameisen:
I cannot link this to any issues, but it is related to:

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9479 (which it resolves)

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6225 (related but does not resolve, switching virtual desktops is possible but not implemented here)

PowerEditor/src/NppCommands.cpp
mere-human: If you're using VS2019, you should build using VS2017 vc141 toolset.
See more info here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9922

PowerEditor/src/winmain.cpp
```diff
@@ -21,6 +21,78 @@
 #include "verifySignedfile.h"
 #include "NppDarkMode.h"
 
+// Include that contain COM classes and IDLs
+// The COM interfaces that we are about here are only available if we specify Windows 10 or higher,
+// So we override the default _WIN32_WINNT that is set in the project file for this include.
+#if defined(_WIN32_WINNT) && (_WIN32_WINNT < _WIN32_WINNT_WIN10)
+#	undef _WIN32_WINNT
+#endif
+#define _WIN32_WINNT _WIN32_WINNT_WIN10
```
mere-human, line 30: Will it work on Windows 7?

PowerEditor/src/winmain.cpp
```diff
@@ -21,6 +21,78 @@
...
+#endif
+#define _WIN32_WINNT _WIN32_WINNT_WIN10
+#include <ShObjIdl.h>
+
+// MinGW and such have IDL headers that do not have IVirtualDesktopManager.
+// Thus, we need to provide the declarations in that case, and ideally
+// declarations that are syntactically valid for both MSVC _and_ MinGW-g++.
+// 
+// These are lifted from ShObjldl.h and ShObjldl_core.h
+#ifndef __IVirtualDesktopManager_INTERFACE_DEFINED__
```
mere-human, line 38: I'd suggest moving it to a separate header file and including here.

PowerEditor/src/winmain.cpp
ameisen: No idea.

Microsoft only defines the class and interfaces in the header if you specify that you want Windows 10-level support. It's possible that the class and interface are registered on 7, but I don't have a system to check with or to figure out what the class and interface IDs are.

So, on Windows 7, it will either work, or it will do nothing because said class and interface IDs don't exist.

PowerEditor/src/winmain.cpp
ameisen: Probably a good idea.

It's frustrating as it only exists because the MinGW headers are lacking these IDL definitions.

PowerEditor/src/NppCommands.cpp
ameisen: In lieu of installing the 141 toolset, why not just use 142 by default?

ameisen:
I removed the VC2019 changes since those are handled in another PR, and moved the big chunk of bringing in the IDLs to its own header.

PowerEditor/src/winmain.cpp
ameisen: Looking further into it, while Windows 7/8.1 do technically support virtual desktops as far as I know, there is no public API for them.

A further implementation would:

1. Require me to have a 7/8.1 system set up.
2. To reverse how to interface with them.

So, I'm not sure how to get the functionality working on them, Microsoft doesn't appear to intend for it.

Unless you meant if this would break the program for Windows 7, in which case, I don't see why it would. While doing that can potentially change some GUIDs, it's being done in a very localized fashion, and only for getting this specific GUID.

sasumner: @ameisen Your PR does not have to bring Virtual Desktop functionality to Windows 7.  Nor should it if extensive coding would be required to do so.  However, your PR must not prevent Notepad++ from running on Windows 7 nor degrade its performance there.

sasumner: Hopefully someone can test it under Windows 7.

ameisen: Well, I added a bit more code to try and make sure that that define gets restored to its original value, and that nothing else is impacted.

I cannot really do any testing, though. It _should_ work, I don't particularly see why it wouldn't, but I cannot assert that it would.

mere-human:
I checked this locally on Windows 10. It seems to open a new instance (when multi-instance is on) on the active virtual desktop which is the right thing.

1. For some reason, the state of the preference "Windows Virtual Desktop aware" isn't remembered. 
If I uncheck it, then close the application, then start it again, it remains checked. 
I had only one instance of Notepad++ so several instances writing to the same config isn't the case.
image

2. I've checked this on Windows 7 VM. It runs without any noticeable problem. 
It would be nice to disable the checkbox control (grey-out) if it has no effect on Windows 7.

3. Also, I'd suggest a better wording of the preference name. For example "Open on the active desktop". Or invert it and name it like "Always open on the first desktop".

PowerEditor/src/winmain.cpp
```diff
@@ -548,51 +554,126 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
...
-			if (::IsZoomed(hNotepad_plus))
-				sw = SW_MAXIMIZE;
-			else if (::IsIconic(hNotepad_plus))
-				sw = SW_RESTORE;
+					shouldTerminate = enumWindowsDataPair.second;
+				}
 
-			if (sw != 0)
-				::ShowWindow(hNotepad_plus, sw);
+				pVDM->Release();
```
mere-human, line 632: Minor suggestion: You can use [com_ptr](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp#L40) or similar smart pointer to avoid memory leaks.
If you chose the former, then its declaration should probably be moved to some header.

PowerEditor/src/winmain.cpp
mere-human: Ok. My concern was that the GUID isn't defined on Windows 7. But since we compile it on Windows 10, it doesn't matter.
When running on Windows 7 it seems like `CoCreateInstance` just doesn't succeed and since we're checking the result, everything is fine.

ameisen: Even compiling on Windows 7, even if the header doesn't have the GUID it would just take the fallback route that MinGW uses and define it with the hardcoded value. So, that should _still work_™.

PowerEditor/src/winmain.cpp
ameisen: I had actually implemented a rather deep `ComPtr` type that included reference counting and awareness of the thread that it was initialized in (due to `CoInitialize` semantics) before realizing that `com_ptr` existed, but had removed it to keep the PR smaller :|.

ameisen:
> 
> 
> I checked this locally on Windows 10. It seems to open a new instance (when multi-instance is on) on the active virtual desktop which is the right thing.
> 
>     1. For some reason, the state of the preference "Windows Virtual Desktop aware" isn't remembered.
>        If I uncheck it, then close the application, then start it again, it remains checked.
>        I had only one instance of Notepad++ so several instances writing to the same config isn't the case.
>        image
> 
>     2. I've checked this on Windows 7 VM. It runs without any noticeable problem.
>        It would be nice to disable the checkbox control (grey-out) if it has no effect on Windows 7.
> 
>     3. Also, I'd suggest a better wording of the preference name. For example "Open on the active desktop". Or invert it and name it like "Always open on the first desktop".

Odd, it does remember it for me. I'll look at that.

I don't like the nomenclature "Always open on the first desktop", as that isn't what it does. It always opens it _on the first desktop it was opened on_, not necessarily the first desktop overall.

ameisen:
@mere-human

My last commit should resolve all of those issues.

ameisen:
The test failure looks like a sporadic failure in the tests?

PowerEditor/src/MISC/Com.h
```diff
@@ -0,0 +1,36 @@
+// This file is part of Notepad++ project
```
mere-human, line 1: 👍

mere-human (approved):
I verified this locally. The behavior is as described below.
Launch Notepad++ on the first desktop. Create a second virtual desktop using `Win+Tab` > `+`.
Version 8:
1. Mono-instance - when opening Notepad++ on second desktop, it brings you back to the first desktop.
2. Multi-instance - when opening Notepad++ on second desktop, it opens you another instance on the current desktop.

With this PR _and_ Preference > Multi-instance > Desktop aware ON:
1. Mono-instance -  after you open Notepad++ on second desktop, it creates another instance on second desktop. Other than that it behaves the same as before but only _within the current desktop_.
2. Multi-instance - no changes, seem to work as before (or I didn't notice).

I didn't find a way to test "Open session in a new instance" option because I don't have association with the file type and I checked using portable version. But it seems to work as before within the same desktop.

ameisen:
So, I have actually noticed sporadic failures of functionality.

99% of the time, this works fine. However, what I've found is that in certain cases, it will open on the wrong desktop:
* If you try to open a file that is already open on another desktop's instance
* Sometimes I've noticed it just go to the wrong desktop overall. Windows' virtual desktop API can sporadically fail, and the documentation specifies that. It just doesn't specify what to do in the case of failure.

In the former, I am thinking there are more places where I need to disable instance checking and force a specific instance per-desktop. In the latter, I'm unsure. There are private APIs that people use, but those are volatile and the GUID changes from version to version (even if the APIs don't) so I'm very wary of using those - the main advantage is that those APIs let you get the virtual desktop's GUID (which is actually _necessary_ for some of the normal APIs - I have no idea how they're meant to be used without that).

PR #9942, opened 2021-06-03
Appveyor vs2019 with toolchain v142

### LINKED ISSUE #9922: Make Notepad++ compilable with VS2019

chcg:
fix #9922

- adapted to build with VS2019 + v142
- as stated in #9922 the build with VS2019 +v141 is possible without code modifications (therefore the split into 2 commits, see https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/39403038)
- see https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/39478493 for a build of this changes with the older v141

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -487,7 +487,8 @@ void Notepad_plus::command(int id)
 			if (nppGui._searchEngineChoice == nppGui.se_custom)
 			{
 				url = nppGui._searchEngineCustom;
-				remove_if(url.begin(), url.end(), _istspace);
+				url.erase(std::remove_if(url.begin(), url.end(), [](_TUCHAR x) {return _istspace(x); }),
+					url.end());
 
 				auto httpPos = url.find(TEXT("http://"));
 				auto httpsPos = url.find(TEXT("https://"));
--- a/boostregex/BoostRegExSearch.cxx
+++ b/boostregex/BoostRegExSearch.cxx
@@ -13,6 +13,9 @@
 #include <iterator> 
 #include <vector>
 #include <memory>
+#include <string>
+#include <string_view>
+#include <stdexcept>
 #include "Scintilla.h"
 #include "Platform.h"
 #include "ILoader.h"
--- a/boostregex/UTF8DocumentIterator.cxx
+++ b/boostregex/UTF8DocumentIterator.cxx
@@ -1,10 +1,14 @@
 ﻿#include "UTF8DocumentIterator.h"
 
 
+#include <string>
+#include <string_view>
+#include <stdexcept>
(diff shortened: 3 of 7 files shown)


boostregex/UTF8DocumentIterator.cxx
```diff
@@ -1,10 +1,14 @@
 ﻿#include "UTF8DocumentIterator.h"
 
 
+#include <string>
```
ameisen, line 4: On my local build, including `<string>` isn't required, only `<string_view>` and `<stdexcept>`.

PowerEditor/src/NppCommands.cpp
```diff
@@ -487,7 +487,8 @@ void Notepad_plus::command(int id)
 			if (nppGui._searchEngineChoice == nppGui.se_custom)
 			{
 				url = nppGui._searchEngineCustom;
-				remove_if(url.begin(), url.end(), _istspace);
```
mere-human, line 490: Nice! You probably fixed some bug 🙂

boostregex/UTF8DocumentIterator.cxx
```diff
@@ -58,6 +62,27 @@ UTF8DocumentIterator::UTF8DocumentIterator(const UTF8DocumentIterator& copy) :
 		}
 }
 
+UTF8DocumentIterator& UTF8DocumentIterator::operator ++ ()
```
mere-human, line 65: But why is it needed?

PowerEditor/src/NppCommands.cpp
chcg: yes, see #5536

boostregex/UTF8DocumentIterator.cxx
chcg: Because I didn't want to add 
#include <string_view>
#include <stdexcept>
for
#include "Platform.h"
in the header file.

boostregex/UTF8DocumentIterator.cxx
chcg: Yes, I just added it because of some copy&paste from BoostRegExSearch.cxx and <string> is not needed here.

chund:
what about the *.vcxproj files?

boostregex/BoostRegExSearch.cxx
```diff
@@ -13,6 +13,9 @@
 #include <iterator> 
 #include <vector>
 #include <memory>
+#include <string>
+#include <string_view>
```
ozone10, line 17: `<string_view>` is part of `<string>` so it's not needed here.

PR #9945, opened 2021-06-03
Fix append extension to work when ok button clicked

### LINKED ISSUE #9939: Append extension doesn't work in Save As dialog
### Description of the Issue

This was originally reported on the Community Forum https://community.notepad-plus-plus.org/post/66533
Thanks to @pryrt  and Alan Kilborn for providing details and @sasumner for testing.
Some discussion was in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9732.

1. Press _Ctrl+N_ to create new file; new1 tab appears
2. Press _Ctrl+Alt+S_ to invoke Save As dialog
3. Check the Append extension checkbox:
image
4. Set the file name to anything without extension (e.g. `z`):
image
5. Click on _Save_ button using left mouse button

### Expected Behavior

File is saved with extension as `z.txt`

### Actual Behavior

File is saved without extension as `z`

Note:
In step 5 if mouse click is replaced by
* _Enter_ key - works as expected (.txt extension added automagically)
* _Alt+S_ - doesn't work (no extension added)

### Debug Information

Notepad++ v8   (64-bit)
Build time : May 27 2021 - 22:10:00
Path : E:\Installs\npp\npp.8.0.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.985
Current ANSI codepage : 1252

mere-human:
Fix #9939

In some cases, checkbox was found instead of OK button and window
procedure was overriden for it. Now, the OK button is identified
using style and label checks.

Also, in some cases IFileDialog::GetFileTypeIndex() returns
the old value. To fix that, always remember the file type selected.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -22,7 +22,6 @@
 #endif
 #include <comdef.h>		// _com_error
 #include <comip.h>		// _com_ptr_t
-
 #include "CustomFileDialog.h"
 #include "Parameters.h"
 
@@ -51,6 +50,7 @@ namespace // anonymous
 
 	static const int IDC_FILE_CUSTOM_CHECKBOX = 4;
 	static const int IDC_FILE_TYPE_CHECKBOX = IDC_FILE_CUSTOM_CHECKBOX + 1;
+	static const TCHAR TEMP_OK_BUTTON_LABEL[] = _T("Save");
 
 	// Returns a first extension from the extension specification string.
 	// Multiple extensions are separated with ';'.
@@ -284,6 +284,10 @@ public:
 	{
 		if (dialogIndex == 0)
 			return false;
+		// Remember the current file type index.
+		// Since GetFileTypeIndex() might return the old value in some cases.
+		// Specifically, when called after SetFileTypeIndex().
+		_currentType = dialogIndex;
 		generic_string name = getDialogFileName(_dialog);
 		if (changeExt(name, dialogIndex - 1))
 			return SUCCEEDED(_dialog->SetFileName(name.c_str()));
@@ -311,20 +315,19 @@ public:
 	{
 		if (id == IDC_FILE_TYPE_CHECKBOX)
 		{
-			bool ok = false;
+			UINT newFileType = 0;
 			if (bChecked)
 			{
-				ok = SUCCEEDED(_dialog->SetFileTypeIndex(_lastSelectedType));
-				ok &= OnTypeChange(_lastSelectedType);
+				newFileType = _lastSelectedType;
 			}
 			else
 			{
-				UINT currentIndex = 0;
-				ok = SUCCEEDED(_dialog->GetFileTypeIndex(&currentIndex));
-				if (ok && currentIndex != 0 && currentIndex != _wildcardType)
-					_lastSelectedType = currentIndex;
-				ok &= SUCCEEDED(_dialog->SetFileTypeIndex(_wildcardType));
+				if (_currentType != 0 && _currentType != _wildcardType)
+					_lastSelectedType = _currentType;
+				newFileType = _wildcardType;
 			}
+			_dialog->SetFileTypeIndex(newFileType);
+			OnTypeChange(newFileType);
 			return S_OK;
 		}
 		return E_NOTIMPL;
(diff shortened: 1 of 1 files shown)


PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -311,20 +315,19 @@ class FileDialogEventHandler : public IFileDialogEvents, public IFileDialogContr
...
-				UINT currentIndex = 0;
-				ok = SUCCEEDED(_dialog->GetFileTypeIndex(&currentIndex));
-				if (ok && currentIndex != 0 && currentIndex != _wildcardType)
-					_lastSelectedType = currentIndex;
-				ok &= SUCCEEDED(_dialog->SetFileTypeIndex(_wildcardType));
+				if (_currentType != 0 && _currentType != _wildcardType)
+					_lastSelectedType = _currentType;
+				newFileType = _wildcardType;
 			}
+			_dialog->SetFileTypeIndex(newFileType);
```
ameisen, line 329: Should there be a sanity check just to make sure that `newFileType` is not the same as `_currentType`? That is, that it actually changed?

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -51,6 +50,7 @@ namespace // anonymous
 
 	static const int IDC_FILE_CUSTOM_CHECKBOX = 4;
 	static const int IDC_FILE_TYPE_CHECKBOX = IDC_FILE_CUSTOM_CHECKBOX + 1;
+	static const TCHAR TEMP_OK_BUTTON_LABEL[] = _T("Save");
```
ameisen, line 53: Following C++14+, this should probably be `static constexpr const...`.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -826,7 +830,7 @@ class CustomFileDialog::Impl
 
 ///////////////////////////////////////////////////////////////////////////////
 
-CustomFileDialog::CustomFileDialog(HWND hwnd) : _impl{std::make_unique<Impl>()}
+CustomFileDialog::CustomFileDialog(HWND hwnd) : _impl{ std::make_unique<Impl>() }
```
ameisen, line 833: Strangely, MSVC (GCC and Clang do not) generates different code depending on if you brace or parantheses-initializer. Also, generates different code if you use `new` instead of `std::make_unique`.

https://godbolt.org/z/3WfKaqEYP

mere-human: I don't see any significant differences between braces vs parentheses init. Below is a diff of your example when I left only int overload and only float overload:
image

As for new, yes, it would be different as expected. 
But still, using `make_unique` over bare `new` is a good practice to avoid leaks if constructor throws:
https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
mere-human: Good point.
But I believe, it could never happen.

This function is called each time the checkbox is clicked.
`_lastSelectedType` could be anything but `_wildcardType` (condition at the line 325).
And when the new file type is selected (user clicks an item in the dropdown list) the checkbox is checked if it was unchecked.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
mere-human: In this context, I think it doesn't matter.
The compiler generates the same assembly.

If I were to declare this as a static member of a class, then it makes more sense. 
Since I could initialize `constexpr` in place while for a `static const` I need an initializer in the `.cpp`.

mere-human:
I also checked this PR on Windows 7 VM. Seems to be working fine.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
ameisen: Sure, I just make it a habit nowadays to use `constexpr` wherever it is relevant instead of just `const` alone.

Many things will generate the same machine code output, but will have slightly different semantics, but in general I keep it `constexpr` because it encourages me to use it more often.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
ameisen: I don't disagree, I was just surprised at the difference as I'd expected the optimizer to reduce them to the same thing in the end.

Don't look at the `wrapper_type`s as they are `__forceinline`d - look at the outputs for `func0`/`1`/`2`.

PR #9948, opened 2021-06-04
Hungarian translation update for 7.9.5

batagy:


batagy:
This is closed as #9949  replaces it.

PR #9954, opened 2021-06-06
fix #2124 and fix #9931 Added a Save all confirm dialog and an option under Preferences->MISC to toggle this

### LINKED ISSUE #2124: "Save All" button - Should it have a confirm action dialog?
### Description of the Issue

The save all button has no confirm, so a mis-click or mis-key will init the save-frenzy. This has undesirable or even dangerous effects if you are in a session with old files, possibly production, possibly in NppFTP....which is exactly what happened recently.
1. Open a session at work, use NppFTP if you like. This session may have lots of old work in tabs. Work on a critical file(s).
2. Go home and work on those files some more via NppFTP using a different PC and session.
3. Come back to work, continue on 1 file, but accidently hit save all, watch as NppFTP overwrites all the old files open in the session, effectively wiping out the work you did at home.
4. Erm....go back home to get the files? I dunno it's just annoying from here on out :)
### Expected Behavior

Exactly what Np++ did. Although some sort of "are you sure you want to save everything in this session" dialog would be very much appreciated.
### Actual Behavior

Np++ did exactly what is expected and normal behavior, it saved everything. This can have undesirable effects if you aren't expecting it.
### Workaround

By using the "customize toolbar" plugin one can remove the icon for save all, reducing mis-clicks. Changing the shortcut to something more complex helps too. But for people who are using stock Np++ and fire off a save all by mistake at the wrong time, a confirm action dialog could save their butts.

### LINKED ISSUE #9931: [FR] Please add "Cancel All" button next to "Cancel" button on file "Save As" dialog window
### Description of the Issue

I am one of those people who has hundreds of open file tabs in the NotePad++ application.  A large percentage of those open file tabs are untitled "works in progress".  I am also a klutz; sometimes I accidentally click the "Save All" button instead of the "Save" button.  When that happens and I have several hundred open file tabs which are still untitled "works in progress" and not yet ready to be committed to hard disk, I am inundated with a long queue of back-to-back "Save As" file dialogs.  I would be eternally grateful if functionality could be added to the "Save As" file dialog window which does the following:

1.  Recognizes that it is not a single file but a whole queue of pending files to be saved.
2.  When so recognized, a "Cancel All" button is added to the dialog window next to the existing "Cancel" button.  
3.  When the Cancel All" button is clicked, the current file "save as" operation and all pending file "save as" operations are immediately bypassed and purged from the pending event queue.

This new feature would spare me from cursing at myself and suffering the burden of holding down the "Esc" key for several minutes in order to dismiss the barrage of pop-up "Save As" dialog windows.

Thank you in advance for your time and consideration.

1.  Create a lot of new files (about a hundred or so, to appreciate the struggle).
2.  Click the "Save As" button in the toolbar.
3.  Sigh loudly while holding down the "Esc" key on the keyboard for as long as it takes for all of the file "Save As" dialogs to stop popping up.

### Debug Information

Notepad++ v7.9.5   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.1556
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NavigateTo.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll XMLTools.dll 

NotePad++FeatureRequest

Ashfaaq18:
fix #2124 and fix #9931
The MISC toggle option will save its state

image


sasumner:
@donho I like the approach described in #9931 where an errant _SaveAll_ is cancellable.  
While this PR's approach theoretically works, I just like the other better.

PR #9967, opened 2021-06-08
Search Results Window locked greyed until the first search

### LINKED ISSUE #9655: Make the Search Results Window locked (grayed) until the first search
### Description of the Issue
I can't open Search Results Window without making one search. It's intentional? Even when I delete all the results, I can open them and it is empty. So why it's blocked at first time? This causes that toggling options in context menu will require one dumy search. 

1. Open NPP.
2. Press F7 or select from menu bar `Search > Search Results window` 
3. Nothing happen.

If this is intentional, and it must be so for some reason, then at least after start NPP let the `Search > Search Results window`  entry in menu bar will be blocked (greyed), until the first search.  So that I don't have to wonder why something is not working when it is visible in this menu. 

image

### Debug Information
```
Notepad++ v7.9.3   (32-bit)
Build time : Mar 14 2021 - 00:16:00
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.0
Current ANSI codepage : 1250
Plugins : none
```

Ashfaaq18:
Animation

fix #9655, The search result window option will only enable after the first Find All search

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -534,6 +534,10 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_pluginsAdminDlg.setPluginsManager(&_pluginsManager);
 	_pluginsManager.setMenu(_mainMenuHandle, NULL, enablePluginAdmin);
 
+	//Search menu
+
+		//disable "Search Results Window" under Search Menu 
+	::EnableMenuItem(_mainMenuHandle, IDM_FOCUS_ON_FOUND_RESULTS, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
 
 	//Main menu is loaded, now load context menu items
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -534,6 +534,10 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_pluginsAdminDlg.setPluginsManager(&_pluginsManager);
 	_pluginsManager.setMenu(_mainMenuHandle, NULL, enablePluginAdmin);
 
+	//Search menu
+
+		//disable "Search Results Window" under Search Menu
```
donho, line 539: Remove the empty line and align the comment.

donho (changes requested):
Remove the empty line and align the comment.

donho (approved):

PR #9968, opened 2021-06-08
Added a Save all confirm dialog and an option under MISC to toggle this

### LINKED ISSUE #2124: "Save All" button - Should it have a confirm action dialog?

### LINKED ISSUE #9931: [FR] Please add "Cancel All" button next to "Cancel" button on file "Save As" dialog window

Ashfaaq18:
fix #2124 and fix #9931
The settings->preferences->MISC toggle option will save its state

Sol2


sasumner:
@Ashfaaq18 
The text in your dialog box needs to be localizable.  Code change and english.xml change required.
Your new text in the _Preferences_ also.  No code change but change to english.xml required.

@donho I like the approach described in #9931 where an errant _SaveAll_ is cancellable.
While this PR's approach theoretically works, I just like the other better.

PowerEditor/src/NppIO.cpp
```diff
@@ -1598,26 +1598,54 @@ bool Notepad_plus::fileSaveSpecific(const generic_string& fileNameToSave)
...
 {
-	if (viewVisible(MAIN_VIEW))
+	bool confirmed = false;
+
+	if (NppParameters::getInstance().getNppGUI()._saveAllConfirm)
 	{
-		for (size_t i = 0; i < _mainDocTab.nbItem(); ++i)
+		int answer = _nativeLangSpeaker.messageBox("SaveAllConfirm",
+			_pPublicInterface->getHSelf(),
+			TEXT("Are you sure you want to save all documents??"),
```
donho, line 1609: One question mark is enough.

PowerEditor/src/NppIO.cpp
```diff
@@ -1598,26 +1598,54 @@ bool Notepad_plus::fileSaveSpecific(const generic_string& fileNameToSave)
...
+			TEXT("Are you sure?"),
+			MB_YESNO);
```
donho, line 1611: Use MB_YESNOCANCEL instead of MB_YESNO.
And add one sentence after "Are you sure you want to save all documents?" : 
`Choose "Cancel" if you want to dis-activate this dialog. You can always activate this option later in Preferences dialog (MISC section).`

PowerEditor/src/NppIO.cpp
```diff
@@ -1598,26 +1598,54 @@ bool Notepad_plus::fileSaveSpecific(const generic_string& fileNameToSave)
...
+
+		if (answer == IDYES)
```
donho, line 1613: After block of `if (answer == IDYES)`
Add `if (answer == IDCANCEL)` to dis-activate `_saveAllConfirm`

donho (changes requested):
1. Adapt the code for people who want to disable this confirm dialog directly.
2. Add localization entries in `english.xml`

donho:
@sasumner 
If you mean ["Cancel All" button on the save dialog](https://user-images.githubusercontent.com/3978924/120402059-4422f700-c30f-11eb-89a5-604943303780.png), I prefer the approach of this PR, for the following reasons:
1. Saving all existent files may not what users want if they click "Save All" by accident.
2. Adding this button in save dialog makes a complex dialog even more complex.

mere-human:
> 2\. Adding this button in save dialog makes a complex dialog even more complex.

I agree. Save dialog shouldn't contain `[Cancel all]` button because the dialog requires user input for a _specific_ file not for _all_ files.

I think, it would be more convenient to change the current `Are you sure?` dialog so that it provides 3 options:
`[OK] [Cancel] [Cancel all]`
And change its title to something like: `Save all?`
And keep a way to disable the dialog in preferences (consistent old behavior).

donho:
@mere-human 
> I think, it would be more convenient to change the current Are you sure? dialog so that it provides 3 options:
[OK] [Cancel] [Cancel all]

What I wanted is even better: `[Yes] [No] [Never ask me]`. But if we use the standard dialog, The label of button is not changeable unfortunately.

donho:
@Ashfaaq18 
I see you have pushed new commits, but there's no change the code of the PR since my last request.
Just want to make sure that you're aware of it.

Ashfaaq18:
> @Ashfaaq18
> I see you have pushed new commits, but there's no change the code of the PR since my last request.
> Just want to make sure that you're aware of it.

oops, yeah. was tryin to update my forked repo without the merge commit. must've messed it up along the way.  Ill re-request a review once I fix this.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1232,6 +1233,10 @@ Do you want to launch Notepad++ in Administrator mode?"/>
 Notepad++ will be restarted after all the operations are terminated.
 Continue?"/>
             <NeedToRestartToLoadPlugins title="Notepad++ need to be relaunched" message="You have to restart Notepad++ to load plugins you installed."/> <!-- HowToReproduce: Import a plugin via menu "Settings->Import->Import Plugin(s)...". -->
+            <SaveAllConfirm title="Are you sure?" message="Are you sure you want to save all documents?
```
donho, line 1236: The caption should not be "Are you sure?" but rather "Save All Confirmation" since the option name is "Enable Save All confirm dialog"

PowerEditor/src/NppIO.cpp
```diff
@@ -1598,26 +1598,62 @@ bool Notepad_plus::fileSaveSpecific(const generic_string& fileNameToSave)
...
-	if (viewVisible(MAIN_VIEW))
+	bool confirmed = false;
+
+	if (NppParameters::getInstance().getNppGUI()._saveAllConfirm)
 	{
-		for (size_t i = 0; i < _mainDocTab.nbItem(); ++i)
+		int answer = _nativeLangSpeaker.messageBox("SaveAllConfirm",
+			_pPublicInterface->getHSelf(),
+			TEXT("Are you sure you want to save all documents?\r\rChoose \"Cancel\" if your answer will always be \"Yes\" and you won't be asked this question again.\rYou can re-activate this dialog in Preferences dialog later."),
+			TEXT("Are you sure?"),
```
donho, line 1610: "Save All Confirmation"

donho (changes requested):
Change the caption of dialog.

donho (approved):

PR #9972, opened 2021-06-08
Remove Gitter badge from Readme

ocdtrekkie:
According to @donho on the Gitter, the channel has been shut down, and inquiries should be sent to the community site. I assumed that Gitter was actively in use on this project because the badge was in the readme, so it would be good to remove it, so newcomers no longer attempt to go to Gitter with their discussions or questions.


mere-human:
This should also be removed then.
image

ocdtrekkie:
@mere-human That does not appear on my Notepad++ 8.0 release.

mere-human:
> Here's the commit where it was deleted: [71fabb1](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/71fabb15676f9f654ef1203cfc28abf95d475749)

Ok, I must've missed that. And my config is outdated.

PR #9973, opened 2021-06-08
Fix Open dialog OK button label

### LINKED ISSUE #9966: "Open File" dialog shows "Save" button
### Description of the Issue
When Opening a file, the dialog box shows "Save" and "Cancel"

1. Open Notepad ++
2. Select File / Open

### Expected Behavior
The window should show "Open" / "Cancel"

### Actual Behavior
(See description of the issue)

### Debug Information
Notepad++ v8   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.985
Current ANSI codepage : 1252

image

mere-human:
Now Open dialog shows Open label on the button instead of Save.
Don't change any button labels.
Instead, get the leftmost button when overriding the window procedure.

Fix #9966

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -50,7 +50,6 @@ namespace // anonymous
 
 	static const int IDC_FILE_CUSTOM_CHECKBOX = 4;
 	static const int IDC_FILE_TYPE_CHECKBOX = IDC_FILE_CUSTOM_CHECKBOX + 1;
-	static const TCHAR TEMP_OK_BUTTON_LABEL[] = _T("Save");
 
 	// Returns a first extension from the extension specification string.
 	// Multiple extensions are separated with ';'.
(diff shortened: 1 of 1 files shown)


mere-human:
Checked:
- Windows 10, Windows 7
- Multi-monitor
- English, Arabic (RTL) OS language
- Advanced name scenarios (Unix-style path, relative path, etc)

donho:
Could you please point me out in which PR this regression was introduced?

donho:
> Could you please point me out in which PR this regression was introduced?

OK, I see: It came from your latest PR.

So why and how it did happen?
It's not for blaming you, just want to prevent such regression in the future.

mere-human:
> So why and how it did happen?

This was introduced in my previous PR.

I search for some controls of the Save/Open dialog in order to override window procedures for them.
The problem it addressed was that `EnumChildWindows()` function found the wrong button (checkbox not OK button).

So, I decided to change the OK button label, then use the label to differentiate from other buttons during search (Cancel, etc).
After the search is done, I called `SetOkButtonLabel(nullptr)` which restored the label to what it was before (even localized one).
But turned out, it works only for Save dialog but not for Open dialog.
For Open dialog, the call did nothing.
MSDN docs say nothing about it.
Here is the change:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9945/files#diff-eeb5624a35a43795da4eb970149a9ce7d22858b678a242affd2357520ea3e9f2R592

Maybe, because I was in a hurry. But I'm surprised nobody else had noticed it before the release.

donho:
@mere-human 

> Maybe, because I was in a hurry. Sorry about that.

It's not only your fault, it's under my responsibility - since it was me who approve and merge the PR.

> But I'm surprised nobody else had noticed it before the release.

Such error is not easy to find - who would think of checking Open File dialog?
Once again, It's not for blaming anyone, just want to prevent such regression in the future.

> I think I should make that tests stuff that I was talking about a priority now.

Unit tests work fine for some component, but not for UI component.

I saw that you use the same procedure/template for both save & open file dialog. 
It might be compact, but makes code more fragile. My suggestion is do a refactoring to separate this 2 features (even separate 1 file to 2 or several files), so you know what you are doing on which part exactly. But since it's your code, I leave the decision to you.

Issue #9984, opened 2021-06-10
Dark mode - remove white line under menubar


adzm:
### Description of the Issue

remove white line under menubar.

I'll fill this out with examples shortly, just want an ID to track this under

### Steps to Reproduce the Issue

1.  Enable dark mode
2.  See white line under menu!

PR #9985, opened 2021-06-10
#9984 Dark mode menubar white line can be overdrawn

adzm:
Fixes the white line issue #9984

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -353,6 +353,34 @@ namespace NppDarkMode
 		}
 	}
 
+	void drawUAHMenuNCBottomLine(HWND hWnd)
+	{
+		MENUBARINFO mbi = { sizeof(mbi) };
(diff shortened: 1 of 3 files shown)


adzm:
Need to test in some other windows situations, DPIs etc to make sure this works right in all those situations. This is admittedly a hack but appears to have no visible flicker. The white line must be overdrawn in response to DrawMenuBar via NCPAINT or NCACTIVATE

WildByDesign:
@adzm This is fantastic. Great work!

I never knew that I hated that white line until now.

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -353,6 +353,34 @@ namespace NppDarkMode
 		}
 	}
 
+	void drawUAHMenuNCBottomLine(HWND hWnd)
+	{
+		MENUBARINFO mbi = { sizeof(mbi) };
+		if (!GetMenuBarInfo(hWnd, OBJID_MENU, 0, &mbi))
```
martijnlaan, line 359: `mbi` is not actually used in the remainder of the code?

PR #9995, opened 2021-06-11
give FileLockedWarning when file could not be saved

### LINKED ISSUE #9994: No error message when locked file could not be saved
### Description of the Issue

No error message when file could not be saved.

1. have file which is locked by another program
2. start notepad++ in admin mode
3.  open the file and edit
4. save with Ctrl + S or the Symbol bar
5. => no error that the file was not saved (because it is locked by another process)

### Expected Behavior

An Error message, that there was a sharing violation or the file was locked, or whatever Windows gives as error 

### Actual Behavior
File is not saved, no error message.
File still shows as edited (what is OK).

### Debug Information

Notepad++ v7.9.5   (64-bit)
Admin mode : ON
Local Conf mode : OFF
OS Name : Windows Server 2016 Standard (64-bit) 
OS Version : 1607
OS Build : 14393.4467
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

chund:
fix #9994
when file was locked and could not be saved, no error message was given.
Now the error message is given. The else with an obviously already existing error message was inserted.

Let me know if it would be desirable to also output the Windows error message, why the file could not be opened/saved.

### DIFF
--- a/PowerEditor/installer/nativeLang/slovenian.xml
+++ b/PowerEditor/installer/nativeLang/slovenian.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!--
-        Slovenian localization for Notepad++ v7.9.4
-        Last modified 15. Mar 2021 by dr. Vinko Kastelic
+        Slovenian localization for Notepad++ v8.0.0
+        Last modified 08. Jun 2021 by dr. Vinko Kastelic
         
 -->
 <NotepadPlus>
-    <Native-Langue name="Slovenski jezik" filename="slovenian.xml" version="7.9.4">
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -649,6 +649,14 @@ bool Notepad_plus::doSave(BufferID id, const TCHAR * filename, bool isCopy)
 			}
 
 		}
+		else
+		{
+			_nativeLangSpeaker.messageBox("FileLockedWarning",
```
mere-human, line 654: At first I thought this isn't localized. But turns out, there is a localization for `FileLockedWarning`.
But for some reason, it's not used anywhere in the code.
Anyway, I would recommend adding `MB_ICONWARNING`.

PR #10001, opened 2021-06-13
Use Scintilla's Regex instead of boost's one

### LINKED ISSUE #9975: Auto-Completion (Word-Completion) is very slow in v8
### Description of the Issue 
Auto-completion (Word completion only) is too slow to use - If the word list is long, the box takes 2-3 minutes to appear.

1. Make the text file of 5-6 MB,
2. open in v8. 
3. Set the following settings:
image
4. Type a character

### Expected Behavior
The Auto-completion box appears immediately.

### Actual Behavior
If the word list is long, the box takes 2-3 minutes to appear 

### Debug Information
Notepad++ v8   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1252

donho:
Another option to fix #9975
If all the tests can pass, then this solution can be considered.

### DIFF
--- a/scintilla/win32/SciLexer.vcxproj
+++ b/scintilla/win32/SciLexer.vcxproj
@@ -73,7 +73,7 @@
   <ItemDefinitionGroup>
     <ClCompile>
       <WarningLevel>Level4</WarningLevel>
-      <PreprocessorDefinitions>WIN32;SCI_LEXER;_CRT_SECURE_NO_DEPRECATE;_SCL_SECURE_NO_WARNINGS;_WINDOWS;_USRDLL;_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES;SCI_OWNREGEX;SCI_NAMESPACE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>WIN32;SCI_LEXER;_CRT_SECURE_NO_DEPRECATE;_SCL_SECURE_NO_WARNINGS;_WINDOWS;_USRDLL;_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES;SCI_NAMESPACE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <AdditionalIncludeDirectories>..\include;..\src;..\lexlib;..\..\boostregex</AdditionalIncludeDirectories>
       <BrowseInformation>true</BrowseInformation>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>


sasumner:
Dubious, as it would likely change much behavior users depend upon.

chcg:
I agree with @sasumner without boost regex PCRE syntax (https://www.boost.org/doc/libs/1_76_0/libs/regex/doc/html/boost_regex/syntax.html) gets lost and just ECMAScript or Posix style remains.

donho:
@sasumner 
I am not sure about it...
My reason is that the function list use RegEx heavily, if all the build tests pass, then we can tell quite safely it's compatible.
OTOH, could you provide me an RegEx which works with boost's RegEx engin but not on Scintilla's one ?

donho:
@MAPJe71 
What's your opinion about it?

sasumner:
> OTOH, could you provide me an RegEx which works with boost's RegEx engin but not on Scintilla's one ?

It might work fine for internal things such as function-list, but users that have gotten used to the advanced features and even the quirks of the Boost engine will be greatly upset. For those that depend upon regex heavily for their work, they will be unhappy.  Probably.

I have no concrete examples to give.

donho:
@chcg @sasumner 

OK, after trying what @ozone10 has provide:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9975#issuecomment-860659516

I'm convinced that it's better to keep boost's regex.

MAPJe71:
Sorry I'm late to the party...

> the function list use RegEx heavily

and if I remember correctly some of these uses some syntax specific for the Boost RegEx engine.
IMO switching to a different RegEx engine shouldn't be taken lightly. That doesn't mean it's impossible though, there's just a high risk of regressions.

The [regular-expressions website](http://www.regular-expressions.info) has a [comprehensive list](http://www.regular-expressions.info/tools.html) of available RegEx libraries/engines.
Furthermore, that website also has the possibility to [compare the syntax of the libraries/engines] (http://www.regular-expressions.info/refcharacters.html).

donho:
@MAPJe71 
> and if I remember correctly some of these uses some syntax specific for the Boost RegEx engine.
IMO switching to a different RegEx engine shouldn't be taken lightly. That doesn't mean it's impossible though, there's just a high risk of regressions.

This PR has not passed unit-tests about regex, but since the result of script was not checked by Appveyor script, and I didn't verify the unit-tests result manually, I thought all the unit-tests of function list are passed. Hence that question. 

Reopen it for checking the build result - the unit-tests are not passed:

image

Issue #9636, opened 2021-03-10
Boost replacement case modifiers do not work as expected


guy038:
### Description of the Issue

The Boost case modifiers ( **`\U`**, **`\L`**; **`\u`** and **`\l`** ), in **replacement**, do **not** change the **case** of any **accentuated** character, so for any **letter** char with **Unicode** code-point **over** **`U+007F`** !

### Steps to Reproduce the Issue

- Place the simple **French** sentence, below, in a **new** tab

~~~diff
C'est là, près de la forêt, dans un gîte où régnait un grand capharnaüm, que l'aïeul ôta sa flûte et son bâton de son canoë.
~~~

- Open the **Replace** dialog ( **`Ctrl + H`** )

    - SEARCH **`\w`**
	
	- REPLACE **`\U$0`**    or    **`\u$0`**

    - Tick the **`Wrap around`** option

    - Click on the **`Replace All`** button

### Expected Behavior

**After** the replacement, the text should be changed like below :

~~~diff
C'EST LÀ, PRÈS DE LA FORÊT, DANS UN GÎTE OÙ RÉGNAIT UN GRAND CAPHARNAÜM, QUE L'AÏEUL ÔTA SA FLÛTE ET SON BÂTON DE SON CANOË.
~~~

These **expected** modifications can be done, of course, by **selecting** the text and using the **default** **`Shift + Ctrl + U`** shortcut

### Actual Behavior

**After** the replacement, we get this **output** :

~~~diff
C'EST Là, PRèS DE LA FORêT, DANS UN GîTE Où RéGNAIT UN GRAND CAPHARNAüM, QUE L'AïEUL ôTA SA FLûTE ET SON BâTON DE SON CANOë.
~~~

It's obvious that **all** the **accentuated** characters have **not** been modified, in their **uppercase** form, by the regex S/R !

The reasoning would be **identical** with the **`\L`** or **`\l`** **case** modifiers, applied on an initial **uppercase** text

## Notes 

- This issue occurs, **both**, in **`ANSI`** or **`Unicode`** encoded files, as **`UTF-8`**

- This issue exists since the implementation of the **Boost** regex library, on N++ **`v6.0.0.`** !

- The **last** N++ version, used for the **tests**, was the **`v7.9.2`** release

---

Best Regards,

guy038

guy038:
Hi all, 

Sorry ! In the sentence :

[...]

PR #10010, opened 2021-06-15
performance issue introduced with switch to static linkage

### LINKED ISSUE #9975: Auto-Completion (Word-Completion) is very slow in v8

chcg:
fix #9975
performance issue introduced with switch to static linkage of libSciLexer

change was originally added with:  Improve character case handling in RegEx (#9707)

### DIFF
--- a/boostregex/BoostRegExSearch.cxx
+++ b/boostregex/BoostRegExSearch.cxx
@@ -306,11 +306,9 @@ Sci::Position BoostRegexSearch::FindText(Document* doc, Sci::Position startPosit
 			);
 		search._skip_windows_line_end_as_one_character = (sciSearchFlags & SCFIND_REGEXP_SKIPCRLFASONE) != 0;
 		
-		std::locale l = std::locale::global(std::locale(""));
 		Match match =
 			isUtf8 ? _utf8.FindText(search)
 			       : _ansi.FindText(search);
-		std::locale::global(l);
 
 		if (match.found())
 		{


mere-human:
So, was it locale slowing down the regex? Interesting.
Should the other 2 usages of locale introduced in the original PR be also reverted?

donho:
@mere-human 
>So, was it locale slowing down the regex? Interesting.

Rather `locale` call slowing down the regex in **static link** case. 

>Should the other 2 usages of locale introduced in the original PR be also reverted?

For me yes.

donho:
@chcg 
Good work! 
Could you please let us know how you did find this commit (which is even before the merge of static link commit) is the cause of performance issue?

It seems `locale` call is culprit of regex performance issue on static link, don't you think it's better to revert both #9707 & #9733 ?

donho:
Just checked 2 versions of binaries:
A. the current fix
B. remove all `locale` calls (ie. Revert both  #9707 & #9733)

Result:
A. performance: fixed - no regression
B. performance: fixed (very slightly faster in some circumstances) - with regression of #9636

So this PR will be taken as it is.

chcg:
Something I missed to add:
- The fix for #9636 seems to be not working for me in N++ 7.9.2 and 7.9.5, but with 8.0. @donho Could you confirm that?

How I found it:
As my computer is quite slow with the regression the search in the lore ipsum text is taking about 1 min and I attached the debugger to the release (as I expected debug might show me wrong findings) version and just stopped randomly multiple times and often ended up to be in the change of the locale. First I thought it might be in relation to the exception handling there, but after commenting out the try catch it still happened. Also tried different settings from https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/boostregex/boost/regex/config.hpp like https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/boostregex/boost/regex/config.hpp#L320, but that made no difference.
Maybe profiling would be more elegant nowadays, but I have not tested it so far.

donho:
@chcg 
> The fix for #9636 seems to be not working for me in N++ 7.9.2 and 7.9.5, but with 8.0. @donho Could you confirm that?

That's correct.

> How I found it:
As my computer is quite slow with the regression the search in the lore ipsum text is taking about 1 min and I attached the debugger to the release ...

The old PCs never die. They just slow down.
Thank you for sharing your debug method!

Issue #10022, opened 2021-06-17
[Feature Request] Improve DarkMode (TitleBar and ScrollBar)


ozone10:
### Description of the Issue

Some parts are not in darkmode
### Steps to Reproduce the Issue

1.  Open Notepad++ v8.1
2.  Enable DarkMode
3. a. Display some docked window (Project panel, Funstion list, ...) with scrollbar
b. Some dialog windows don't have dark titlebar.

### Expected Behavior

Scrollbar and titlebar are dark
### Actual Behavior

Scrollbar is sometimes light, titlebar is light

donho:
Fixed by @ozone10

Issue #9958, opened 2021-06-07
Dark mode missed a few spots


xcasxcursex:
Title. A major issue is plugins. As far as the base app goes:
![image](https://user-images.githubusercontent.com/58130821/120977326-8daf8e00-c7b6-11eb-9ecc-2ea3c174de68.png)

Dark mode though... finally... Feels good man!

richardcrossley:
Thanks for raising this, I was going to report the same. 

Notepad++, rocks, thanks this will be on my work PC in the morning.

sasumner:
Clearing refers to version 8.
I'm not looking forward to slogging through all of the "this bit was forgotten for dark mode!" issues that are probably soon to be created. :-)

pixeltester:
Plugins icons have disappeared from the toolbar even with darkmode deactivated or rolling back to previous release, is this the same issue? I can't see how to reinstate them.

pixeltester:
Plugin icons are invisible in light mode also.

pixeltester:
![invisible](https://user-images.githubusercontent.com/48839175/121021585-71e9bf00-c799-11eb-9441-7093469d6953.png)

Invisible plugin icons in light mode also.

McCovican:
Tabs seem to revert to light mode when multi-line tabs are enabled and you open enough tabs to go onto a second line.

sasumner:
@McCovican Maybe like described in #9957 ?

tfkinc:
The editor stayed for me in light mode on one system. I did restart NPP several times - no changes.

![image](https://user-images.githubusercontent.com/41904824/122230888-5168e580-ceba-11eb-9618-befb464c53f9.png)

On a secondary system it works just fine:

![image](https://user-images.githubusercontent.com/41904824/122231346-ac024180-ceba-11eb-855e-0932eb28914c.png)

xcasxcursex:
> The editor stayed for me in light mode on one system. I did restart NPP several times - no changes.
Settings... Style Configurator... Select Theme: ...

xcasxcursex:
Just a little bump for the tab overflow buttons: 

![image](https://user-images.githubusercontent.com/58130821/129551034-6407fe47-f8f5-4f65-8df1-40244b70e1b7.png)

Thanks :)

ArkadiuszMichalski:
Duplicate of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10171.

PR #10023, opened 2021-06-17
Improve DarkMode (TitleBar)

ozone10:
Make DarkMode title bar to apply on all dialog windows. (Now should be missing only message box dialog windows)

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10022
partially fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9958


donho:
@ozone10 
Thank you for your PR.
However, I don't see the difference before & after applying your PR.
A screenshot of comparison will help.

ozone10:
Note the dialog windows title bar and scrollbar in ascii panel
Notepad++ v8.1
image


This pull request
image

Now dark title bar is applied almost everywhere, I think it is missing only for windows created by MessageBox function

donho (commented):
Make 2 PR 1 for Titlebar, another for ScrollBar.

donho (changes requested):
Make 2 PR 1 for Titlebar, another for ScrollBar.

ozone10:
They are using same function. Should I split the function `void setDarkThemeViaWinAPI(HWND hwnd, bool setTitle);`
to one with "bool true" and one with "bool false".

ozone10:
Now only affect title bar.

donho:
@ozone10 
Thank you, the code is more readable now.
However, there is a regression comes with this PR: while we are in full dark mode (ie. restart Notepad++ to have titlebar and scrollbar in black), then we switch in light mode, the titlebar doesn't turn light.
The current implementation turns the titlebar light without restarting Notepad++.

donho (approved):

donho (changes requested):
Fix the regression that titlebar doesn't turn into light mode directly.

donho:
@ozone10 
Almost...
image

PowerEditor/src/WinControls/ClipboardHistory/clipboardHistoryPanel.rc
```diff
@@ -24,5 +24,5 @@ EXSTYLE WS_EX_TOOLWINDOW | WS_EX_WINDOWEDGE
 CAPTION "Clipboard History"
 FONT 8, "MS Sans Serif", 0, 0, 0x0
 BEGIN
-	LISTBOX         IDC_LIST_CLIPBOARD,50,44,78,120,LBS_OWNERDRAWFIXED | LBS_NOINTEGRALHEIGHT | WS_VSCROLL | WS_TABSTOP
+	LISTBOX         IDC_LIST_CLIPBOARD,50,44,78,120,LBS_OWNERDRAWFIXED | LBS_NOINTEGRALHEIGHT | WS_VSCROLL | WS_TABSTOP | NOT WS_BORDER
```
donho, line 27: Why do we need to remove `WS_BORDER` ?

donho: OK, I see the visual effect.

donho:
@ozone10 
Now you can add Scrollbar PR please

PR #10034, opened 2021-06-18
Fix unsupported compiler option when building Notepad++ x64 & ARM64

### LINKED ISSUE #10033: Wrong Enhanced Instruction Set (compiler)
I have noticed when compiling x64 release I am getting
```
Command line warning D9002: ignoring unknown option '/arch:IA32' 
```
This option is not valid for x64 and arm64.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fd5bfe42527029355966167f75d54f346fd0ef92/PowerEditor/visual.net/notepadPlus.vcxproj#L318

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fd5bfe42527029355966167f75d54f346fd0ef92/PowerEditor/visual.net/notepadPlus.vcxproj#L371

mere-human:
Remove unsupported option /arch:IA32 from x64 & ARM64 Release for Notepad++ project.

Fix #10033

### DIFF
--- a/PowerEditor/visual.net/notepadPlus.vcxproj
+++ b/PowerEditor/visual.net/notepadPlus.vcxproj
@@ -315,7 +315,6 @@ copy ..\src\contextMenu.xml ..\bin\contextMenu.xml
       <WarningLevel>Level4</WarningLevel>
       <TreatWarningAsError>true</TreatWarningAsError>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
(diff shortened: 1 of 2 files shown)


chcg:
@mere-human The iterator warnings are not in Scintilla 3rd party code, but in the extensions from N++ to replace the builtin regex with boost. This is N++ specific. So it would be also possible to modify the iterators to not be derived.

mere-human:
> The iterator warnings are not in Scintilla 3rd party code, but in the extensions from N++ to replace the builtin regex with boost

Indeed. Does it make sense to revert that part and make a separate PR?

PR #10036, opened 2021-06-19
Remove std::iterator from Boost Regex search code

### LINKED ISSUE #10035: Deprecated iterator warning when building Notepad++
### Description of the Issue

The warning is reported on custom iterator code that allow Scintilla to work with Boost.Regex.

1. Open notepadPlus.sln in Visual Studio
2. Build > Rebuild Solution

### Expected Behavior

No warning should be reported.
The code in question should be fixed not to use deprecated `std::iterator`.

### Actual Behavior

C4996 warning is reported about std::iterator deprecation (10 times).

### Debug Information

```
Notepad++ v8.1   (64-bit)
Path : ...\Notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1052
Current ANSI codepage : 1252
Plugins : none
```

Detailed warning message:
```
1>c:\program files (x86)\microsoft visual studio\2019\community\vc\tools\msvc\14.16.27023\include\xutility(462): warning C4996: 'std::iterator<std::bidirectional_iterator_tag,char,ptrdiff_t,_Ty *,_Ty &>::iterator_category': warning STL4015: The std::iterator class template (used as a base class to provide typedefs) is deprecated in C++17. (The <iterator> header is NOT deprecated.) The C++ Standard has never required user-defined iterators to derive from std::iterator. To fix this warning, stop deriving from std::iterator and start providing publicly accessible typedefs named iterator_category, value_type, difference_type, pointer, and reference. Note that value_type is required to be non-const, even for constant iterators. You can define _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
1>        with
1>        [
1>            _Ty=char
1>        ] (compiling source file ..\..\boostregex\BoostRegExSearch.cxx)
1>c:\program files (x86)\microsoft visual studio\2019\community\vc\tools\msvc\14.16.27023\include\xutility(4294): note: see declaration of 'std::iterator<std::bidirectional_iterator_tag,char,ptrdiff_t,_Ty *,_Ty &>::iterator_category'
1>        with
1>        [
1>            _Ty=char
1>        ] (compiling source file ..\..\boostregex\BoostRegExSearch.cxx)
1>c:\program files (x86)\microsoft visual studio\2019\community\vc\tools\msvc\14.16.27023\include\xutility(491): note: see reference to class template instantiation 'std::_Iterator_traits_base<_Iter,void>' being compiled
1>        with
1>        [
1>            _Iter=Scintilla::AnsiDocumentIterator
1>        ] (compiling source file ..\..\boostregex\BoostRegExSearch.cxx)
1>...\notepad-plus-plus\boostregex\boost\regex\v5\match_results.hpp(52): note: see reference to class template instantiation 'std::iterator_traits<BidiIterator>' being compiled
1>        with
1>        [
1>            BidiIterator=Scintilla::AnsiDocumentIterator
1>        ] (compiling source file ..\..\boostregex\BoostRegExSearch.cxx)
1>...\notepad-plus-plus\boostregex\boostregexsearch.cxx(206): note: see reference to class template instantiation 'boost::match_results<CharacterIterator,std::allocator<_Ty>>' being compiled
1>        with
1>        [
1>            CharacterIterator=Scintilla::AnsiDocumentIterator,
1>            _Ty=boost::sub_match<Scintilla::AnsiDocumentIterator>
1>        ]
1>...\notepad-plus-plus\boostregex\boostregexsearch.cxx(236): note: see reference to class template instantiation 'BoostRegexSearch::EncodingDependent<char,Scintilla::AnsiDocumentIterator>' being compiled
```

mere-human:
std::iterator was deprecated in C++17.
Remove it to fix the warnings and avoid other issues.

Fix #10035

### DIFF
--- a/boostregex/AnsiDocumentIterator.h
+++ b/boostregex/AnsiDocumentIterator.h
@@ -1,13 +1,35 @@
-﻿#ifndef ANSIDOCUMENTITERATOR_H_12481491281240
+﻿// This file is part of Notepad++ project
+// Copyright (C) 2021 Notepad++ authors.
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// at your option any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef ANSIDOCUMENTITERATOR_H_12481491281240
 #define ANSIDOCUMENTITERATOR_H_12481491281240
 
 #include "Position.h"
 
 namespace Scintilla {
 
-class AnsiDocumentIterator : public std::iterator<std::bidirectional_iterator_tag, char>
+class AnsiDocumentIterator
 {
 public:
+	using iterator_category = std::bidirectional_iterator_tag;
+	using value_type = char;
+	using difference_type = ptrdiff_t;
+	using pointer = char*;
+	using reference = char&;
+
 	AnsiDocumentIterator() {};
 
 	AnsiDocumentIterator(Document* doc, Sci::Position pos, Sci::Position end) :
--- a/boostregex/UTF8DocumentIterator.h
+++ b/boostregex/UTF8DocumentIterator.h
@@ -1,8 +1,23 @@
-﻿#ifndef UTF8DOCUMENTITERATOR_H_3452843291318441149
+﻿// This file is part of Notepad++ project
+// Copyright (C) 2021 Notepad++ authors.
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// at your option any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef UTF8DOCUMENTITERATOR_H_3452843291318441149
 #define UTF8DOCUMENTITERATOR_H_3452843291318441149
 
 #include <stdlib.h>
-#include <iterator>
 #include <vector>
 #include <memory>
 #include "Position.h"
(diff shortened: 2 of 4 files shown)


boostregex/UTF8DocumentIterator.h
```diff
@@ -12,80 +27,86 @@ namespace Scintilla {
...
-		}
-
-		bool operator != (const UTF8DocumentIterator& other) const
+	using iterator_category = std::bidirectional_iterator_tag;
+	using value_type = wchar_t;
+	using difference_type = ptrdiff_t;
+	using pointer = wchar_t*;
+	using reference = wchar_t&;
+
+	UTF8DocumentIterator() {};
```
mere-human, line 39: This is just a whitespace change.
I changed 2 tabs to 1 tab to make it consistent with other code..

boostregex/AnsiDocumentIterator.h
```diff
@@ -1,13 +1,35 @@
-﻿#ifndef ANSIDOCUMENTITERATOR_H_12481491281240
+﻿// This file is part of Notepad++ project
+// Copyright (C) 2021 Notepad++ authors.
```
donho, line 2: Next time you can use "Anonymous" or "Nobody" instead of "Notepad++ authors" if you don't want to leave your name.
BTW, hiding your name could be a solution. My way is:
"Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live." 
;)

mere-human: > Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live.

Yep. This is exactly what I have in mind.
I don't want to be sued for accidentally breaking some contract conditions😄

But honestly, I wasn't the author of this file. So I just added a "generic" comment header.

boostregex/AnsiDocumentIterator.h
```diff
@@ -1,13 +1,35 @@
...
-class AnsiDocumentIterator : public std::iterator<std::bidirectional_iterator_tag, char>
+class AnsiDocumentIterator
 {
 public:
+	using iterator_category = std::bidirectional_iterator_tag;
+	using value_type = char;
+	using difference_type = ptrdiff_t;
+	using pointer = char*;
+	using reference = char&;
+
```
donho, line 32: Any explanation for using `using ...` to replace  ` public std::iterator<std::bidirectional_iterator_tag, char>` please?

mere-human: `std::iterator` was deprecated in C++17.
It is [defined](https://en.cppreference.com/w/cpp/iterator/iterator) as following:

```
template< 
    typename Category,
    typename T,
    typename Distance = std::ptrdiff_t,
    typename Pointer = T*,
    typename Reference = T& 
> struct iterator;
```

It was deprecated mainly because it wasn't clear which type is what when specifying template parameters.
Using declaration allow to be more specific.

> What is deprecated is the technique of inheriting from std::iterator to define those types. That’s it.
> 
```
class MyOutputIterator : public std::iterator<std::output_iterator_tag, void, void, void, void>
{
```
> This code doesn’t say to which type of the interface (value_type, reference_type…) each of the types passed corresponds.
> 
> A more explicit way to go about it is to write the using declarations (or typedefs if you’re before C++11) directly inside of the iterator

More info:
https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

Originally, I just wanted to suppress the warning. But @chcg suggested to modify the code so that the warning isn't generated and the code is a valid C++17 code.

donho: Thank you for the explanation. However, we cannot just count on this article.
Has this modification been tested? If yes, how?

mere-human: Yes, I did some testing. I opened the file 
[lorem.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6682034/lorem.txt)
I checked auto complete and regex search/count with simple expressions like `q\w+s\b`. 

What else proofs are needed?
https://en.cppreference.com/w/cpp/iterator/iterator
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0174r2.html

The generated assembler is exactly the same.
The only difference is that _with_ `std::iterator` the type alias (using/typedef) is defined in the base class.
And _without_ `std::iterator` the type aliases are defined in the same class (AnsiDocumentIterator).

Below is a short example.
With iterator base:
https://godbolt.org/z/xvffq8EnW
Without iterator base:
https://godbolt.org/z/bMGs9aPqz
If you do a diff of the generated assembly, there is no difference.

donho:
@mere-human 
Could you solve the conflict please?

mere-human:
> @mere-human
> Could you solve the conflict please?

Done.
I had to rebase on top of master branch and then do a force push in order to resolve conflicts.

PR #10038, opened 2021-06-19
Fix GCC compiler warnings in Notepad++

### LINKED ISSUE #10037: GCC reports numerous warnings when Notepad++ is built
### Description of the Issue

The warnings make it harder to spot the compiler error if it's present.
Also, they could signal about a potential problem

1. Build Notepad++ with MinGW using instructions in [BUILD.md](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/BUILD.md#build-64-bits-binaries-with-gcc)

### Expected Behavior

No warnings are generated. Or the amount of warnings is minimal.

### Actual Behavior

98 warnings are reported.

### Debug Information

```
Notepad++ v8.1   (64-bit)
Path : ...\Notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1052
Current ANSI codepage : 1252
Plugins : none
```

mere-human:
Fix #10037

### DIFF
--- a/PowerEditor/src/DarkMode/IatHook.h
+++ b/PowerEditor/src/DarkMode/IatHook.h
@@ -30,7 +30,8 @@ PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName,
 			continue;
 
 		auto import = RVA2VA<PIMAGE_IMPORT_BY_NAME>(moduleBase, impName->u1.AddressOfData);
-		if (strcmp(import->Name, funcName) != 0)
+		static_assert(sizeof(*(import->Name)) == sizeof(char), "Name should be 1 byte");
+		if (strcmp(reinterpret_cast<const char*>(import->Name), funcName) != 0)
 			continue;
 		return impAddr;
 	}
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2656,7 +2656,7 @@ bool isUrlTextChar(TCHAR const c)
 		case '{':
 		case '}':
 		case '?':
-		case '\0x7f':
+		case '\u007F':
 			return false;
 	}
 	return true;
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1335,7 +1335,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_CREATESCINTILLAHANDLE:
 		{
-			return (LRESULT)_scintillaCtrls4Plugins.createSintilla((lParam == NULL?hwnd:reinterpret_cast<HWND>(lParam)));
+			return (LRESULT)_scintillaCtrls4Plugins.createSintilla((lParam == 0?hwnd:reinterpret_cast<HWND>(lParam)));
 		}
 
 		case NPPM_INTERNAL_GETSCINTEDTVIEW:
@@ -1413,7 +1413,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			else if (wParam == NPPMAINMENU)
 				return (LRESULT)_mainMenuHandle;
 			else
-				return NULL;
+				return FALSE;
 		}
 
 		case NPPM_LOADSESSION:
@@ -1573,7 +1573,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				for (size_t i = 0, len = _hModelessDlgs.size() ; i < len ; ++i)
 				{
 					if (_hModelessDlgs[i] == reinterpret_cast<HWND>(lParam))
-						return NULL;
+						return FALSE;
 				}
 
 				_hModelessDlgs.push_back(reinterpret_cast<HWND>(lParam));
@@ -1589,7 +1589,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 						{
 							vector<HWND>::iterator hDlg = _hModelessDlgs.begin() + i;
 							_hModelessDlgs.erase(hDlg);
-							return NULL;
+							return FALSE;
 						}
 					}
 					return lParam;
@@ -2146,7 +2146,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case NPPM_DMMGETPLUGINHWNDBYNAME : //(const TCHAR *windowName, const TCHAR *moduleName)
 		{
 			if (!lParam)
-				return NULL;
+				return FALSE;
 
 			TCHAR *moduleName = reinterpret_cast<TCHAR *>(lParam);
 			TCHAR *windowName = reinterpret_cast<TCHAR *>(wParam);
@@ -2167,7 +2167,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 					}
 				}
 			}
-			return NULL;
+			return FALSE;
 		}
 
 		case NPPM_ADDTOOLBARICON_DEPRECATED:
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -1042,7 +1042,7 @@ namespace NppDarkMode
 	{
 		if (NppDarkMode::isEnabled())
 		{
-			int msg = NULL;
+			UINT msg = 0;
 			switch (type)
 			{
 			case NppDarkMode::ToolTipsType::toolbar:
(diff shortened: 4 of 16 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2656,7 +2656,7 @@ bool isUrlTextChar(TCHAR const c)
 		case '{':
 		case '}':
 		case '?':
-		case '\0x7f':
+		case '\u007F':
```
mere-human, line 2659: @Uhf7 Was this an intention?
This line was addded in the commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/401bfcb0cd133d43f550dbe69275faf28663606b
It has decimal value `127`

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2822,25 +2822,19 @@ void ScintillaEditView::currentLinesDown() const
...
 
 	switch (caseToConvert)
 	{
 		case UPPERCASE:
 		{
-			for (int i = 0; i < nbChars; ++i)
-			{
-				strWToConvert[i] = (WCHAR)(UINT_PTR)::CharUpperW((LPWSTR)strWToConvert[i]);
-			}
+			::CharUpperBuffW(strWToConvert, nbChars);
```
mere-human, line 2832: This works exactly the same as CharUpperW:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-charupperbuffa

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2851,12 +2845,12 @@ void ScintillaEditView::changeCase(__inout wchar_t * const strWToConvert, const
 				if (::IsCharAlphaW(strWToConvert[i]))
 				{
 					if ((i < 1) ? true : !::IsCharAlphaNumericW(strWToConvert[i - 1]))
-						strWToConvert[i] = (WCHAR)(UINT_PTR)::CharUpperW((LPWSTR)strWToConvert[i]);
+						strWToConvert[i] = (WCHAR)(UINT_PTR)::CharUpperW(reinterpret_cast<LPWSTR>(strWToConvert[i]));
```
mere-human, line 2848: Note that CharUpperW does the replacement in place, so the assignment is not necessary:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-charupperw
Also, the behavior for a single character isn't very clear (for me) but it seems most people use it like this.

PowerEditor/src/Notepad_plus.cpp
sasumner: I presume the warning here is "duplicate case value"

mere-human: @sasumner No, the warning was `character constant too long for its type`.
As I understood, `'\0xNN'` format is intended only for ASCII char literals.

U007F is _DELETE_ symbol: https://www.fileformat.info/info/unicode/char/7f/index.htm
But I wanted to ensure if it was intended.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1413,7 +1413,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			else if (wParam == NPPMAINMENU)
 				return (LRESULT)_mainMenuHandle;
 			else
-				return NULL;
+				return FALSE;
```
donho, line 1416: Here returning FALSE modifies the semantic of this message.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
donho: Though it could work exactly the same, we take the risk of regression, not even for bug-fixing or a new feature, but for removing the warning of a compiler that we don't use for releasing.

donho (commented):
Please provide the arguments of changing.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
mere-human: I checked using commands from _Edit > Convert case to_ menu (to upper, to lower, inverting mixed case, etc). It works the same.
I thought, that in some cases calling a function on entire string rather than on each character individually could give a better result.
If you want, I can revert this part back to using `CharUpperW`.

PowerEditor/src/Notepad_plus.cpp
mere-human: GCC Warning:
```
../src/Notepad_plus.cpp:2659:8: warning: multi-character character constant [-Wmultichar]
   case '\0x7f':
        ^~~~~~~
```
This is not a valid character literal. If you slightly modify the code, and compile it using MSVC, it also gives you a warning.
The code:
```
#include <iostream>
int main()
{
    wchar_t c = '\0x7f';
    std::cout << c << '\n' << (int)c;
}
```
MSVC Warning:
```
1>e:\Development\cpp\etc\ConsoleTest\ConsoleApplication\Source.cpp(4,21): warning C4305: 'initializing': truncation from 'int' to 'char'
1>e:\Development\cpp\etc\ConsoleTest\ConsoleApplication\Source.cpp(4,14): warning C4309: 'initializing': truncation of constant value
```
Also, even without compiling:
image

---

In this case, GCC is more demanding. And I think this is good since it helps us to spot a potential bug.

PowerEditor/src/DarkMode/IatHook.h
```diff
@@ -30,7 +30,8 @@ PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName,
 			continue;
 
 		auto import = RVA2VA<PIMAGE_IMPORT_BY_NAME>(moduleBase, impName->u1.AddressOfData);
-		if (strcmp(import->Name, funcName) != 0)
+		static_assert(sizeof(*(import->Name)) == sizeof(char), "Name should be 1 byte");
+		if (strcmp(reinterpret_cast<const char*>(import->Name), funcName) != 0)
```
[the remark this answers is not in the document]
mere-human: This is just signed/unsigned conversion. It is better to be explicit. Also, added static_assert to make it safer.
GCC Warning:
```
../src/darkmode/IatHook.h: In function 'IMAGE_THUNK_DATA64* FindAddressByName(void*, PIMAGE_THUNK_DATA, PIMAGE_THUNK_DATA, const char*)':
../src/darkmode/IatHook.h:33:22: warning: invalid conversion from 'BYTE*' {aka 'unsigned char*'} to 'const char*' [-fpermissive]
   if (strcmp(import->Name, funcName) != 0)
              ~~~~~~~~^~~~
```

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -1065,7 +1065,7 @@ namespace NppDarkMode
 			}
 			else
 			{
-				auto hTips = reinterpret_cast<HWND>(SendMessage(hwnd, msg, NULL, NULL));
+				auto hTips = reinterpret_cast<HWND>(SendMessage(hwnd, msg, WPARAM{}, LPARAM{}));
```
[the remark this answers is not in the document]
mere-human: Apparently, both WPARAM and LPARAM aren't pointer types. So passing NULL isn't right.
I could change it to `0`.
GCC Warning:
```
../src/NppDarkMode.cpp:1068:74: warning: passing NULL to non-pointer argument 3 of 'LRESULT SendMessageW(HWND, UINT, WPARAM, LPARAM)' [-Wconversion-null]
     auto hTips = reinterpret_cast<HWND>(SendMessage(hwnd, msg, NULL, NULL));
```

PowerEditor/src/NppBigSwitch.cpp
mere-human: I can change it to `LRESULT{}` or just `0`.

mere-human:
Reasons to fix these warnings:
1. Smaller compiler output. Easier to spot compiler errors. Less clutter.
2. Fix potential problems which may be a source of hidden bugs.

PowerEditor/src/NppDarkMode.cpp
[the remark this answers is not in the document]
ozone10: I used `NULL`, because of some copy&paste, but according to documentation, it should be both zero.

donho:
> Smaller compiler output. Easier to spot compiler errors. Less clutter.
> Fix potential problems which may be a source of hidden bugs.

Agree.
However, the semantic of code change a lot by adapting to GCC (making code less readable). 
I don't think it's wise to take the risk of regression by removing the warning of a compiler that we don't use for releasing.
Since the release tool is VS Studio, let's maintain no warning only in VS Studio.

sasumner:
So, in summary, the GCC build of Notepad++ will only hurt and not help.
That doesn't seem like a great strategy.
How about removing it entirely?
If something isn't going to help, and is only going to hurt --> get rid of it.

mere-human:
> However, the semantic of code change a lot by adapting to GCC (making code less readable).

I can change this PR to make it more readable and less error-prone.

donho:
@sasumner 
It is needed by some people, so let's keep it.
But I don't want to make more effort for it.

donho:
@mere-human 
You can keep trying it. But if an API returns a handle or NULL pointer, and gcc need a BOOL as return, I don't see what you can do more about it.

PR #10039, opened 2021-06-19
Make Project Panel toolbar conform to dark mode

donho:
Project Panel toolbar is still in gray when Notepad++ in dark mode.
The goal of this PR is to fix this bug.
 
image

### DIFF
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -804,20 +804,28 @@ void ProjectPanel::notified(LPNMHDR notification)
 	}
 	else if (notification->code == NM_CUSTOMDRAW && (notification->hwndFrom == _hToolbarMenu))
 	{
-		NMTBCUSTOMDRAW* nmtbcd = reinterpret_cast<NMTBCUSTOMDRAW*>(notification);
-		//if (nmtbcd->nmcd.dwDrawStage == CDDS_PREERASE)
+		if (NppDarkMode::isEnabled())
 		{
-			if (NppDarkMode::isEnabled())
+			static bool isVSDisabled = false;
+			if (!isVSDisabled)
 			{
-				FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, NppDarkMode::getBackgroundBrush());
-				nmtbcd->clrText = NppDarkMode::getTextColor();
-				SetTextColor(nmtbcd->nmcd.hdc, NppDarkMode::getTextColor());
-				SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_SKIPDEFAULT);
+				NppDarkMode::disableVisualStyle(_hToolbarMenu);
+				isVSDisabled = true;
 			}
-			else
-			{
-				SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_DODEFAULT);
-			}
+
+			auto nmtbcd = reinterpret_cast<LPNMTBCUSTOMDRAW>(notification);
+			FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, NppDarkMode::getBackgroundBrush());
+			nmtbcd->clrText = NppDarkMode::getTextColor();
+			// highlight color when hover
+			// same color when hovering above menu 
+			// RGB(65, 65, 65) should be added to NppDarkMode.cpp
+			// needed because, visual style is disabled
+			nmtbcd->clrHighlightHotTrack = RGB(65, 65, 65);
+			SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_NOTIFYSUBITEMDRAW | TBCDRF_HILITEHOTTRACK);
+		}
+		else
+		{
+			SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_DODEFAULT);
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


donho:
@adzm 
I copied your code for the FolderAsWorkspace, it should work but here's the result:
image

Have you any idea what I am missing?

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -802,6 +802,24 @@ void ProjectPanel::notified(LPNMHDR notification)
...
+	{
+		NMTBCUSTOMDRAW* nmtbcd = reinterpret_cast<NMTBCUSTOMDRAW*>(notification);
+		//if (nmtbcd->nmcd.dwDrawStage == CDDS_PREERASE)
+		{
+			if (NppDarkMode::isEnabled())
+			{
+				FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, NppDarkMode::getBackgroundBrush());
+				nmtbcd->clrText = NppDarkMode::getTextColor();
+				SetTextColor(nmtbcd->nmcd.hdc, NppDarkMode::getTextColor());
+				SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_SKIPDEFAULT);
```
ozone10, line 815: This will not work. Because visual style is enabled.
For FolderAsWorkspace [linea 717-718](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp#L717) have no effect because there is no text.
Even with text `SetTextColor` will have no effect.

donho: @ozone10 
So do you have any suggestion?

ozone10: I would remove FolderAsWorkspace lines 717-718, as they do nothing.

For this:
You can disable visual style with `SetWindowTheme(_hToolbarMenu, L"", L"");`.

I would probably add it to NppDarkMode.cpp so we don't have to use `#include <Uxtheme.h>`

quickly prototyping, this seems to work, you can adapt it
```
case WM_NOTIFY:
		{
			auto notification = reinterpret_cast<LPNMHDR>(lParam);

			switch (notification->code)
			{
				case NM_CUSTOMDRAW:
				{
					if (NppDarkMode::isEnabled() && notification->hwndFrom == _hToolbarMenu)
					{
						auto nmtbcd = reinterpret_cast<LPNMTBCUSTOMDRAW>(notification);
						FillRect(nmtbcd->nmcd.hdc, &nmtbcd->nmcd.rc, NppDarkMode::getBackgroundBrush());
						nmtbcd->clrText = NppDarkMode::getTextColor();
                                                // highlight color when hover
                                                // same color when hovering above menu 
                                                // RGB(65, 65, 65) should be added to NppDarkMode.cpp
                                                // needed because, visual style is disabled
						nmtbcd->clrHighlightHotTrack = RGB(65, 65, 65); 
						SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_NOTIFYSUBITEMDRAW | TBCDRF_HILITEHOTTRACK);
					}
					else
					{
						SetWindowLongPtr(_hSelf, DWLP_MSGRESULT, CDRF_DODEFAULT);
					}
					break;
				}
				default:
				{
					notified(notification);
				}
			}
		}
return TRUE;
```

For function list it should also work

donho: @ozone10 
Thank you! Your solution works fine.
Please let me know if you have suggestion about the current state of PR.

ozone10: Maybe when switching back to light mode, enable visual style with this `SetWindowTheme(hwnd, nullptr, nullptr);`

donho: > Maybe when switching back to light mode, enable visual style with this SetWindowTheme(hwnd, nullptr, nullptr);

Without doing so, it switches back without problem on my machine... I will take as it is now. We will see if it's really needed on some different configuration.

donho: @ozone10 
> I would remove FolderAsWorkspace lines 717-718, as they do nothing.

Line 717 & 718 contains a curly opened brace and a empty line respectively.
Could you point out explicitly the no effect lines please?

donho: @ozone10 
Removed : https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5df7c97f98915ed293366d9d521fe5115b2c7c45

Thank you. 

> also if visual style is not enabled, highlight seems different
you can check it when starting in light mode, enable dark mode and then disable without restart

I cannot see the difference :(

ozone10: @donho
start in light mode
image

switch from dark mode
image

donho: @ozone10 
Thank you for screenshots.
Fixed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/296e1c695e0e38a4233ab2a70278fcb675419e63

ozone10:
@donho 
some issue if you have more project panels open only one will have white text.
image

donho:
@ozone10 
> some issue if you have more project panels open only one will have white text.

OK I see. It's due the static variable which control the calling of `NppDarkMode::setExplorerTheme(_hToolbarMenu, false);` (to make sure it's called only once when switching in dark mode from the light mode), though I don't have any the solution so far.

donho:
@ozone10 
Fixed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4ae106377414a308b06787059f4643916f526284

PR #10042, opened 2021-06-20
Make line above panel toolbar dark.

ozone10:
For Project panel, Function list panel and Folder as workplace panel.

before
image
after
image

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10043


donho:
@ozone10 
Could it be changed dynamically according the choice of UI (dark mode or light mode) ?

PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
```diff
@@ -166,8 +167,14 @@ INT_PTR CALLBACK FileBrowser::run_dlgProc(UINT message, WPARAM wParam, LPARAM lP
 			_treeView.makeLabelEditable(false);
 			_treeView.display();
 
-            return TRUE;
-        }
+			return TRUE;
+		}
```
ozone10, line 171: this is probably because of whitespace

PR #10048, opened 2021-06-21
Fix GCC compiler warnings in Notepad++

### LINKED ISSUE #10037: GCC reports numerous warnings when Notepad++ is built

mere-human:
The semantic of the code remains the same.
Now GCC doesn't generate warnings during the build.
Main reasons for this change:
1. Smaller compiler output (245KB => 200KB). Easier to spot compiler errors. Less clutter.
2. Fix potential problems which may be a source of hidden bugs.

Fix #10037

### DIFF
--- a/PowerEditor/src/DarkMode/IatHook.h
+++ b/PowerEditor/src/DarkMode/IatHook.h
@@ -30,7 +30,6 @@ PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName,
 			continue;
 
 		auto import = RVA2VA<PIMAGE_IMPORT_BY_NAME>(moduleBase, impName->u1.AddressOfData);
-		static_assert(sizeof(*(import->Name)) == sizeof(char), "Name should be 1 byte");
 		if (strcmp(reinterpret_cast<const char*>(import->Name), funcName) != 0)
 			continue;
 		return impAddr;
--- a/PowerEditor/src/DarkMode/IatHook.h
+++ b/PowerEditor/src/DarkMode/IatHook.h
@@ -30,7 +30,8 @@ PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName,
 			continue;
 
 		auto import = RVA2VA<PIMAGE_IMPORT_BY_NAME>(moduleBase, impName->u1.AddressOfData);
-		if (strcmp(import->Name, funcName) != 0)
+		static_assert(sizeof(*(import->Name)) == sizeof(char), "Name should be 1 byte");
+		if (strcmp(reinterpret_cast<const char*>(import->Name), funcName) != 0)
 			continue;
 		return impAddr;
 	}
--- a/PowerEditor/src/DarkMode/IatHook.h
+++ b/PowerEditor/src/DarkMode/IatHook.h
@@ -30,7 +30,7 @@ PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName,
 			continue;
 
 		auto import = RVA2VA<PIMAGE_IMPORT_BY_NAME>(moduleBase, impName->u1.AddressOfData);
-		if (strcmp(reinterpret_cast<const char*>(import->Name), funcName) != 0)
+		if (strcmp(import->Name, funcName) != 0)
 			continue;
 		return impAddr;
 	}
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1413,7 +1413,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			else if (wParam == NPPMAINMENU)
 				return (LRESULT)_mainMenuHandle;
 			else
-				return FALSE;
+				return 0;
 		}
 
 		case NPPM_LOADSESSION:
@@ -1573,7 +1573,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				for (size_t i = 0, len = _hModelessDlgs.size() ; i < len ; ++i)
 				{
 					if (_hModelessDlgs[i] == reinterpret_cast<HWND>(lParam))
-						return FALSE;
+						return 0;
 				}
 
 				_hModelessDlgs.push_back(reinterpret_cast<HWND>(lParam));
@@ -1589,7 +1589,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 						{
 							vector<HWND>::iterator hDlg = _hModelessDlgs.begin() + i;
 							_hModelessDlgs.erase(hDlg);
-							return FALSE;
+							return 0;
 						}
 					}
 					return lParam;
@@ -2146,7 +2146,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case NPPM_DMMGETPLUGINHWNDBYNAME : //(const TCHAR *windowName, const TCHAR *moduleName)
 		{
 			if (!lParam)
-				return FALSE;
+				return 0;
 
 			TCHAR *moduleName = reinterpret_cast<TCHAR *>(lParam);
 			TCHAR *windowName = reinterpret_cast<TCHAR *>(wParam);
@@ -2167,7 +2167,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 					}
 				}
 			}
-			return FALSE;
+			return 0;
 		}
 
 		case NPPM_ADDTOOLBARICON_DEPRECATED:
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1335,7 +1335,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_CREATESCINTILLAHANDLE:
 		{
-			return (LRESULT)_scintillaCtrls4Plugins.createSintilla((lParam == 0?hwnd:reinterpret_cast<HWND>(lParam)));
+			return (LRESULT)_scintillaCtrls4Plugins.createSintilla((lParam ? reinterpret_cast<HWND>(lParam) : hwnd));
 		}
 
 		case NPPM_INTERNAL_GETSCINTEDTVIEW:
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1335,7 +1335,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_CREATESCINTILLAHANDLE:
 		{
-			return (LRESULT)_scintillaCtrls4Plugins.createSintilla((lParam == NULL?hwnd:reinterpret_cast<HWND>(lParam)));
+			return (LRESULT)_scintillaCtrls4Plugins.createSintilla((lParam == 0?hwnd:reinterpret_cast<HWND>(lParam)));
 		}
 
 		case NPPM_INTERNAL_GETSCINTEDTVIEW:
@@ -1413,7 +1413,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			else if (wParam == NPPMAINMENU)
 				return (LRESULT)_mainMenuHandle;
 			else
-				return NULL;
+				return FALSE;
 		}
 
 		case NPPM_LOADSESSION:
@@ -1573,7 +1573,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				for (size_t i = 0, len = _hModelessDlgs.size() ; i < len ; ++i)
 				{
 					if (_hModelessDlgs[i] == reinterpret_cast<HWND>(lParam))
-						return NULL;
+						return FALSE;
 				}
 
 				_hModelessDlgs.push_back(reinterpret_cast<HWND>(lParam));
@@ -1589,7 +1589,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 						{
 							vector<HWND>::iterator hDlg = _hModelessDlgs.begin() + i;
 							_hModelessDlgs.erase(hDlg);
-							return NULL;
+							return FALSE;
 						}
 					}
 					return lParam;
@@ -2146,7 +2146,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case NPPM_DMMGETPLUGINHWNDBYNAME : //(const TCHAR *windowName, const TCHAR *moduleName)
 		{
 			if (!lParam)
-				return NULL;
+				return FALSE;
 
 			TCHAR *moduleName = reinterpret_cast<TCHAR *>(lParam);
 			TCHAR *windowName = reinterpret_cast<TCHAR *>(wParam);
@@ -2167,7 +2167,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 					}
 				}
 			}
-			return NULL;
+			return FALSE;
 		}
 
 		case NPPM_ADDTOOLBARICON_DEPRECATED:
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1413,7 +1413,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			else if (wParam == NPPMAINMENU)
 				return (LRESULT)_mainMenuHandle;
 			else
-				return 0;
+				return NULL;
 		}
 
 		case NPPM_LOADSESSION:
@@ -1573,7 +1573,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				for (size_t i = 0, len = _hModelessDlgs.size() ; i < len ; ++i)
 				{
 					if (_hModelessDlgs[i] == reinterpret_cast<HWND>(lParam))
-						return 0;
+						return NULL;
 				}
 
 				_hModelessDlgs.push_back(reinterpret_cast<HWND>(lParam));
@@ -1589,7 +1589,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 						{
 							vector<HWND>::iterator hDlg = _hModelessDlgs.begin() + i;
 							_hModelessDlgs.erase(hDlg);
-							return 0;
+							return NULL;
 						}
 					}
 					return lParam;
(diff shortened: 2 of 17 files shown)


mere-human:
Don [said](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10038#issuecomment-864589218) I can keep trying to fix it, so I did.
Now, the change is more clear and less error prone.
I'm fine if this won't get accepted.

PowerEditor/src/NppIO.cpp
```diff
@@ -334,7 +334,7 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive,
 	BufferID buffer;
 	if (isSnapshotMode)
 	{
-		buffer = MainFileManager.loadFile(longFileName, NULL, encoding, backupFileName, fileNameTimestamp);
+		buffer = MainFileManager.loadFile(longFileName, Document{}, encoding, backupFileName, fileNameTimestamp);
```
donho, line 337: Could you provide the reference about the empty curly braces notation in c++?

mere-human: Yes. It is called "uniform initialization":
https://isocpp.org/wiki/faq/cpp11-language#uniform-init
When there is nothing inside the curly braces, a proper default value is used. 
It doesn't matter whether it's a built-in integral type (`0`) or some class (`std::string("")`).

In this case, Document is defined as following:
`typedef sptr_t Document;`
But using uniform initialization, we don't care about the actual type.

Do you think it's better to use `0`?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -346,7 +346,7 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
 	{
 		case WM_MOUSEHWHEEL :
 		{
-			::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) > 0)?SB_LINERIGHT:SB_LINELEFT, NULL);
+			::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) > 0)?SB_LINERIGHT:SB_LINELEFT, 0);
```
donho, line 349: OK

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -361,9 +361,9 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
 			if (LOWORD(wParam) & MK_SHIFT)
 			{
 				// move 3 columns at a time
-				::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) < 0) ? SB_LINERIGHT : SB_LINELEFT, NULL);
-				::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) < 0) ? SB_LINERIGHT : SB_LINELEFT, NULL);
-				::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) < 0) ? SB_LINERIGHT : SB_LINELEFT, NULL);
+				::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) < 0) ? SB_LINERIGHT : SB_LINELEFT, 0);
+				::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) < 0) ? SB_LINERIGHT : SB_LINELEFT, 0);
+				::CallWindowProc(_scintillaDefaultProc, hwnd, WM_HSCROLL, ((short)HIWORD(wParam) < 0) ? SB_LINERIGHT : SB_LINELEFT, 0);
```
donho, line 366: OK

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2831,15 +2831,15 @@ void ScintillaEditView::changeCase(__inout wchar_t * const strWToConvert, const
 		{
 			for (int i = 0; i < nbChars; ++i)
 			{
-				strWToConvert[i] = (WCHAR)(UINT_PTR)::CharUpperW((LPWSTR)strWToConvert[i]);
+				strWToConvert[i] = (WCHAR)(UINT_PTR)::CharUpperW(reinterpret_cast<LPWSTR>(strWToConvert[i]));
```
donho, line 2834: 👍

PowerEditor/src/Utf8_16.cpp
```diff
@@ -222,7 +222,7 @@ void Utf8_16_Read::determineEncoding()
 		m_nSkip = 3;
 	}
 	// try to detect UTF-16 little-endian without BOM
-	else if (m_nLen > 1 && m_nLen % 2 == 0 && m_pBuf[0] != NULL && m_pBuf[1] == NULL && IsTextUnicode(m_pBuf, static_cast<int32_t>(m_nLen), &uniTest))
+	else if (m_nLen > 1 && m_nLen % 2 == 0 && m_pBuf[0] != 0 && m_pBuf[1] == 0 && IsTextUnicode(m_pBuf, static_cast<int32_t>(m_nLen), &uniTest))
```
donho, line 225: 👍

PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
```diff
@@ -739,7 +739,7 @@ BrowserNodeType FileBrowser::getNodeType(HTREEITEM hItem)
 		return browserNodeType_file;
 	}
 	// Root
-	else if (tvItem.lParam != NULL && !reinterpret_cast<SortingData4lParam*>(tvItem.lParam)->_rootPath.empty())
+	else if (tvItem.lParam != 0 && !reinterpret_cast<SortingData4lParam*>(tvItem.lParam)->_rootPath.empty())
```
donho, line 742: Semantic problem: it could be `nullptr` but not `0`.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -501,7 +501,7 @@ void ProjectPanel::buildProjectXml(TiXmlNode *node, HTREEITEM hItem, const TCHAR
 	{
 		tvItem.hItem = hItemNode;
 		SendMessage(_treeView.getHSelf(), TVM_GETITEM, 0, reinterpret_cast<LPARAM>(&tvItem));
-		if (tvItem.lParam != NULL)
+		if (tvItem.lParam != 0)
```
donho, line 504: Semantic problem: it could be `nullptr` but not `0`.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -829,7 +829,7 @@ NodeType ProjectPanel::getNodeType(HTREEITEM hItem)
 		return nodeType_project;
 	}
 	// Folder
-	else if (tvItem.lParam == NULL)
+	else if (tvItem.lParam == 0)
```
donho, line 832: Semantic problem: it could be `nullptr` but not `0`.

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
```diff
@@ -24,7 +24,7 @@ StaticDialog::~StaticDialog()
 	if (isCreated())
 	{
 		// Prevent run_dlgProc from doing anything, since its virtual
-		::SetWindowLongPtr(_hSelf, GWLP_USERDATA, NULL);
+		::SetWindowLongPtr(_hSelf, GWLP_USERDATA, 0);
```
donho, line 27: Semantic problem: it could be `nullptr` but not `0`.

PowerEditor/src/WinControls/TreeView/TreeView.h
```diff
@@ -37,7 +37,7 @@ class TreeView : public Window {
 
 	virtual void init(HINSTANCE hInst, HWND parent, int treeViewID);
 	virtual void destroy();
-	HTREEITEM addItem(const TCHAR *itemName, HTREEITEM hParentItem, int iImage, LPARAM lParam = NULL);
+	HTREEITEM addItem(const TCHAR* itemName, HTREEITEM hParentItem, int iImage, LPARAM lParam = 0);
```
donho, line 40: Semantic problem: it could be `nullptr` but not `0`.

PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
```diff
@@ -1119,7 +1119,7 @@ INT_PTR CALLBACK ShortcutMapper::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 bool ShortcutMapper::findKeyConflicts(__inout_opt generic_string * const keyConflictLocation,
 										const KeyCombo & itemKeyComboToTest, const size_t & itemIndexToTest) const
 {
-	if (itemKeyComboToTest._key == NULL) //no key assignment
+	if (itemKeyComboToTest._key == 0) //no key assignment
```
donho, line 1122: 👍

PowerEditor/src/DarkMode/IatHook.h
```diff
@@ -30,7 +30,7 @@ PIMAGE_THUNK_DATA FindAddressByName(void *moduleBase, PIMAGE_THUNK_DATA impName,
 			continue;
 
 		auto import = RVA2VA<PIMAGE_IMPORT_BY_NAME>(moduleBase, impName->u1.AddressOfData);
-		if (strcmp(import->Name, funcName) != 0)
+		if (strcmp(reinterpret_cast<const char*>(import->Name), funcName) != 0)
```
donho, line 33: Please keep this one unchanged. The type of `Name` is a `char` but not `char *`.

donho (changes requested):
Please follow my comments for keeping or revert the change.

PowerEditor/src/NppIO.cpp
donho: OK, thank you for the info.
Please remove all the empty curly braces notation - it's not explicit enough. 
I prefer that pointers are initialized by a `NULL` or `nullptr`, but not `0` or `{}`.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1413,7 +1413,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			else if (wParam == NPPMAINMENU)
 				return (LRESULT)_mainMenuHandle;
 			else
-				return NULL;
+				return 0;
```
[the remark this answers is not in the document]
mere-human: This function has return type of `LRESULT`. 
This type is not a pointer but an integer. So, I don't think it is right to return `NULL`.
`LRESULT` is defined as following in [minwindef.h](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/shared/minwindef.h#L188):
`typedef LONG_PTR            LRESULT;`
LONG_PTR is in turn defined as following in [basetsd.h](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types#LONG_PTR)
`typedef __int64 LONG_PTR, *PLONG_PTR;`

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1573,7 +1573,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				for (size_t i = 0, len = _hModelessDlgs.size() ; i < len ; ++i)
 				{
 					if (_hModelessDlgs[i] == reinterpret_cast<HWND>(lParam))
-						return NULL;
+						return 0;
```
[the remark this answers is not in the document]
mere-human: `LRESULT` is not a pointer but an integer. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655490155).

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1589,7 +1589,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 						{
 							vector<HWND>::iterator hDlg = _hModelessDlgs.begin() + i;
 							_hModelessDlgs.erase(hDlg);
-							return NULL;
+							return 0;
```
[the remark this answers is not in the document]
mere-human: `LRESULT` is not a pointer but an integer. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655490155).

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2146,7 +2146,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case NPPM_DMMGETPLUGINHWNDBYNAME : //(const TCHAR *windowName, const TCHAR *moduleName)
 		{
 			if (!lParam)
-				return NULL;
+				return 0;
```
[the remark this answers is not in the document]
mere-human: `LRESULT` is not a pointer but an integer. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655490155).

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2167,7 +2167,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 					}
 				}
 			}
-			return NULL;
+			return 0;
```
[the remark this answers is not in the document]
mere-human: `LRESULT` is not a pointer but an integer. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655490155).

PowerEditor/src/WinControls/FileBrowser/fileBrowser.cpp
mere-human: `LPARAM` is not a pointer but an integer. So it's not right to compare it with or assign a `NULL` to it.
It is a defined as following in [minwindef.h](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/shared/minwindef.h#L187):
`typedef LONG_PTR            LPARAM;`

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
mere-human: `LPARAM` is not a pointer but an integer. So it's not right to compare it against `NULL`. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655494587).

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
mere-human: `LPARAM` is not a pointer but an integer. So it's not right to compare it against `NULL`. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655494587).

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
mere-human: `SetWindowLongPtr` function's 3rd argument is `LONG_PTR` which is not a pointer but an integer:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptra

PowerEditor/src/WinControls/TreeView/TreeView.h
mere-human: `LPARAM` is not a pointer but an integer. So it's not right to compare it against `NULL`. See my previous [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10048#discussion_r655494587).

PowerEditor/src/DarkMode/IatHook.h
mere-human: Done.
The type of `import->Name` is `CHAR`. For some reason, it is defined as `BYTE` in GCC which is `unsigned char` hence the warning.

PowerEditor/src/NppIO.cpp
mere-human: `Document` is integer but not a pointer. Changed to 0.

In [Buffer.h](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.h#L28):
`typedef sptr_t Document;`

In [Scintilla.h](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/include/Scintilla.h#L35):
`typedef intptr_t sptr_t;`

In `<stdint.h>`:
`typedef __int64          intptr_t;`

PowerEditor/src/DarkMode/IatHook.h
mere-human: Oh, turns out, the latest MinGW header has defined it right:
https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/include/winnt.h#L7226
So this is not needed anymore.

mere-human:
Fixed all the comments that should be fixed.
Added explanation why the rest of the code should be left as is (`LRESULT` and `LPARAM`).
Now, the code should be easier to read and no warnings are generated.

mere-human:
For some warnings might seem not that important.
But sometimes, the warning show us potential problems.
I want to make the project clean and tidy.
It is similar something similar to [broken window theory](https://en.wikipedia.org/wiki/Broken_windows_theory).

PowerEditor/src/NppIO.cpp
donho: > Document is integer but not a pointer. Changed to 0.

Yes, I know. I checked it.
But with the same logic, every pointers are an integer, period.

donho (changes requested):
All about semantic, so where the variables are used as pointers, let's return NULL or nullptr. No need to check the original type - because a pointer of object is an integer after all.

Please restore NULL instead of 0 in the case that I indicated.

mere-human:
> All about semantic, so where the variables are used as pointers, let's return NULL or nullptr. No need to check the original type - because a pointer of object is an integer after all.
> 
> Please restore NULL instead of 0 in the case that I indicated.

Maybe, disable this specific warning for GCC and revert the related chages?

donho:
> Maybe, disable this specific warning for GCC and revert the related chages?

In makefile? that could be a solution.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -227,8 +227,8 @@ void Searching::displaySectionCentered(int posStart, int posEnd, ScintillaEditVi
 	pEditView->execute(SCI_CHOOSECARETX);
 }
 
-LONG_PTR FindReplaceDlg::originalFinderProc = NULL;
-LONG_PTR FindReplaceDlg::originalComboEditProc = NULL;
+LONG_PTR FindReplaceDlg::originalFinderProc = {};
+LONG_PTR FindReplaceDlg::originalComboEditProc = {};
```
donho, line 231: As I said, no `{}` please.

donho (changes requested):
Rermove all `{}`

donho (approved):

PR #10053, opened 2021-06-21
Corrections & additions greek.xml 

cpolychronidis:
Corrected some mistranslated parts, added some translations and fixed format to be identical to the one in english.xml in order to make translation easier.


donho:
Don't reformatting : 

If you continue maintaining Greek localization file and the formatting is really necessary, then  we will see about it.
No formatting for this PR so the modification can be checked easily.

Issue #10061, opened 2021-06-24
Regression:  Explorer theme for list- and tree-views unreadable in NON-dark mode


vinsworldcom:
### Description of the Issue
As per [discussion](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c8be679cde9936c94782d6b02e6dae32cee9b9f8#commitcomment-52586681) on commit c8be679cde9936c94782d6b02e6dae32cee9b9f8, using:

- Windows 10 64-bit with system dark mode
- Notepad++ 64-bit commit c8be679cde9936c94782d6b02e6dae32cee9b9f8
- Notepad++ in NON-dark mode using Obsidian-like theme

The function list highlighting and foreground text combination becomes unreadable.

### Steps to Reproduce the Issue
As described above, the view is:

![image](https://user-images.githubusercontent.com/9657169/123111224-be283680-d40a-11eb-9e99-c53f83006636.png)

Notice the head of the function list is "highlighted" and the foreground text is black leading to unreadable.

### Expected Behavior
Expected behavior using Notepad++ version 8.1 or any commit **_before_** that referenced above:

![image](https://user-images.githubusercontent.com/9657169/123192555-2c074900-d471-11eb-9b2a-216554ea87dc.png)

Notice the head of the function list is "highlighted" white and the foreground text is black making it readable - but perhaps the white highlight is undesirable for Notepad++ dark mode (which I'm NOT using in this example).

### Debug Information (pre-commit c8be679cde9936c94782d6b02e6dae32cee9b9f8 showing expected result)
Notepad++ v8.1   (64-bit)
Build time : Jun 22 2021 - 10:27:44
Path : C:\usr\bin\npp64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.1621
Current ANSI codepage : 1252
Plugins : 

### Debug Information (post-commit c8be679cde9936c94782d6b02e6dae32cee9b9f8 showing actual, undesirable result)
Notepad++ v8.1   (64-bit)
Build time : Jun 24 2021 - 01:54:24
Path : C:\usr\bin\npp64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
[...]

PR #10062, opened 2021-06-24
Disable Explorer Theme for TreeView

ozone10:
I am quite busy now, so until I have time and figure out how to do custom draw for treeview, I will disable eplorer theme.


ozone10:
There is issue that in dark mode scroll bar can revert back to light mode. But this was before commit with Explorer Theme.

Issue #10083, opened 2021-06-29
[DarkMode] Improve scrollbar dynamic switching


ozone10:
Not all scrollbars support dynamic switching from light to dark mode and vice versa.

PR #10084, opened 2021-06-29
Improve dynamic switching for scrollbars.

ozone10:
For main and sub edits, finder (search result docking window) and clipboard history panel.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10083

now only missing tree- and list-view scrollbars, and scrollbars in dialog windows


donho:
@ozone10 
Why not all panel? (ie. FaW, ASCII Insert panel & Function list... etc)

ozone10:
@donho 
> Why not all panel? (ie. FaW, ASCII Insert panel & Function list... etc)

Because both use different approach and require restart-less darkmode.
And treeview has issue with highlight.

For dialog (About, MD5, ...) it is possible, but I think it would be better do one by one gradually, so there is no dialog with too much mixed elements..

PR #10088, opened 2021-06-30
Improve link to user manual on questionmark (help) menu

### LINKED ISSUE #10087: ?-menu link to online User Manual should be named consistently
### Description of the Issue

The **?**-menu's link to the [online user manual](npp-user-manual.org) doesn't have consistent naming.  By default (per the  [Notepad_plus.rc](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9a2dcaa5f8eb678e2b0c6494c16a8b9df37f06bb/PowerEditor/src/Notepad_plus.rc#L1135)), it is labeled "Notepad++ Online Document".  But if you use the **Settings > Preferences > General > Localization** pulldown to choose either [english.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f9d6fb9e31719a0ef9bbc8bd42c262560575adf8/PowerEditor/installer/nativeLang/english.xml#L322) or [english_customizable.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/585608b456d5c7373bbae9e02a86d3e91059b775/PowerEditor/installer/nativeLang/english_customizable.xml#L319), that menu entry will shrink down to "Online Documentation" -- so even though the default is English, it doesn't match the two English localization files.

The naming should be consistent between Notepad_plus.rc, english.xml, and english_customizable.xml.  Also, because the branding of the npp-user-manual is "Notepad++ User Manual", I think it would be better to make the English name for all three locations as "Notepad++ User Manual".

1. Fresh installation / unzip
2. Look at **?** menu and see "Notepad++ Online Document"
3. **Settings > Preferences > General > Localization**: select `English (customizable)`
4. Look at **?** menu and see "Online Documentation"
5. **Settings > Preferences > General > Localization**: select `English`
6. Look at **?** menu and see "Online Documentation"

### Expected Behavior

See the same text whether it's the default menu text, english.xml menu text, or english_customizable.xml menu text.  Suggested text would be "Notepad++ User Manual"

### Actual Behavior

See different text in the default menu text vs the english.xml or english_customizable.xml menu text.

### Debug Information

```
Notepad++ v8.1   (64-bit)
Build time : Jun 17 2021 - 01:56:07
Path : C:\usr\local\apps\npp\npp.8.1.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 1903
OS Build : 18362.1256
Current ANSI codepage : 1252
```

sasumner:
Fix #10087

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -319,7 +319,7 @@ The comments are here for explanation, it's not necessary to translate them.
                     <Item id="47010" name="Command Line Arguments..."/>
                     <Item id="47001" name="Notepad++ Home"/>
                     <Item id="47002" name="Notepad++ Project Page"/>
-                    <Item id="47003" name="Online Documentation"/>
+                    <Item id="47003" name="Notepad++ Online User Manual"/>
                     <Item id="47004" name="Notepad++ Community (Forum)"/>
                     <Item id="47012" name="Debug Info..."/>
                     <Item id="47005" name="Get More Plugins"/>
--- a/PowerEditor/src/Notepad_plus.rc
+++ b/PowerEditor/src/Notepad_plus.rc
@@ -1132,7 +1132,7 @@ BEGIN
         MENUITEM SEPARATOR
         MENUITEM "Notepad++ Home",                 IDM_HOMESWEETHOME
         MENUITEM "Notepad++ Project Page",         IDM_PROJECTPAGE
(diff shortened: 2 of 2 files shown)


pryrt:
what about english_customizable?

sasumner:
>what about english_customizable.xml?

It really should just be a copy of english.xml.
I'm not sure of its update schedule, or how it is done.

Issue #10085, opened 2021-06-29
multi line editing control text in Italian translation


Trip20:
### Description of the Issue
Hello everyone, I would kindly like to draw attention to the incorrect and non-exhaustive Italian translation used for the checkbox that refers to the enabling of the multiline mode discussed in 3d https://community.notepad-plus-plus.org/topic/21409/multi-line-editing/5

### Steps to Reproduce the Issue

I would suggest adding a function specification item regarding the "enable (Ctrl + CLICK / SELECTION)" checkbox in Configuration >> Preferences >> Display

### Expected Behavior
![cornice](https://user-images.githubusercontent.com/59014838/123870235-56a24780-d932-11eb-9e80-28dc0e7fd99d.png)

### Actual Behavior
![multilinea](https://user-images.githubusercontent.com/59014838/123869462-52295f00-d931-11eb-8c73-df432d480c23.jpg)

### Debug Information
Notepad++ v7.9.5   (64-bit)
Build time : Mar 21 2021 - 02:13:17
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 7 Ultimate (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll

sasumner:
@Trip20 

In the Community postings, you were asked to supply a suggestion for the text -- you neglected to do so. :-(

Looking at the Italian for 7.9.1 (the most-recent version to have _Multi-Editing Settings_ in a groupbox), we see:

![image](https://user-images.githubusercontent.com/30118311/123958368-f8e61d80-d97a-11eb-81b8-8c528b9dc426.png)

So a suggestion here would be to change the non-groupboxed text for this control to:

`Abilita Selezione Multipla (Ctrl+clic/selezione)`

A concern is, is this text too long to fit?

Or maybe there is a better translation for "Enable  Multi-Editing".
But an Italian speaker would have to supply this; not sure what the reluctance on the part of @Trip20 is to actually do this.

Trip20:
hi sasumner, no reluctance on my part to do this!
In the "Expected Behavior" I have reported the picture of how it could be displayed.
The problem is that I followed the translation of my thought: in the image I translated it too and I have not reported it in Italian. I am fused!
[...]

PR #10093, opened 2021-07-01
Fix incorrect translation "Enable Multi Editing" in italian.xml

conky77:
Fix incorrect translation "Enable Multi Editing" (Issue #10085 )


donho:

donho:
@conky77 Never mind.
I'm doing the v8.1.2 Release Candidate, so I will integrate your update as it is.
The upcoming update will be in the next release. Thank you.

PR #10100, opened 2021-07-03
Use current file directory in File Rename dialog

### LINKED ISSUE #10095: File Rename should suggest current directory, not Default Save Directory
As mentioned [HERE](https://community.notepad-plus-plus.org/topic/21397) a simple _File_ > _Rename_ of file should probably not suggest the directory specified with _Default Open/Save file Directory_ when that setting is not _Follow current document_.   Rather, it should always suggest the directory that the file is currently in.  Optionally, a new preference setting could be made for controlling this functionality.

### Debug Information
Notepad++ v8.1   (64-bit)
Build time : Jun 17 2021 - 01:56:07
Path : C:\\............\\npp.8.1.portable.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 1909
OS Build : 18363.1556
Current ANSI codepage : 1252

mere-human:
Fix #10095

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1717,7 +1717,7 @@ bool Notepad_plus::fileRename(BufferID id)
 
 		fDlg.setExtFilter(TEXT("All types"), TEXT(".*"));
 		setFileOpenSaveDlgFilters(fDlg, false);
-
+		fDlg.setFolder(buf->getFullPathName());
 		fDlg.setDefFileName(buf->getFileName());
 		generic_string fn = fDlg.doSaveDlg();
 


donho:
It doesn't work.
Have you tested the PR you provided? :)

mere-human:
> It doesn't work, and it makes even a regression.
> Have you tested the PR you provided? :)

Yes, I tested it on Windows 10 and Windows 7 VM.
Does rename still uses the Default Directory for you?

mere-human:
How to test:
* Open Notepad++
* Menu > Settings > Preferences > Default Directory
* Change to `%USERPROFILE%\Desktop`
* Open some text file (not in Desktop folder)
* File > Rename
* Expected: Save As dialog is opened in the file's directory

mere-human:
Here is a video of the new behavior:

https://user-images.githubusercontent.com/9664141/124386482-582d9000-dce3-11eb-9345-9cbe182a01e1.mp4



Note the directory after Rename command is different from the directory after Open command.

sasumner:
I tested it as well.
With _Follow current document_ in effect, the directory that is opened is very strange.  (Note that I did not try the other two choices for _Default Open/Save Directory_.)
It is one that I frequently use in my daily-use version of N++, but that directory is not known to my running debug instance.
It is certainly not the folder of the current document.
(On Windows 10.)

donho:
@mere-human 
Here's the result of my test of Rename op:

image

donho:
BTW, can "Save as..." to be changed into "Rename..." on title bar?

mere-human:
> It is certainly not the folder of the current document.

> Here's the result of my test of Rename op:

Ok, thank you all for testing it. I will look into the problems you've mentioned.


> BTW, can "Save as..." to be changed into "Rename..." on title bar?

I think it is possible. I can try.

mere-human:
I fixed all the problems and updated **How to test** [above](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10100#issuecomment-873587168).

donho:
It doesn't work when `Remember used directory` is set.
The way to reproduce it:
Save a doc as c:\temp\a.txt, choose a file with different path, say c:\foo\b.txt
then do"Rename" & "Save as" several times without click OK (but cancel 
then go to c:\faa\c.txt and "Save as" - the last used directory becomes c:\foo\

BTW, the caption should not be "Rename current tab" but just "Rename" is OK.

mere-human:
> then go to c:\faa\c.txt and "Save as" - the last used directory becomes c:\foo\

Isn't it the expected behavior? "Save as" follows the last used directory. "Rename" does not. In the previous step, you used "Save as" for `c:\foo\b.txt`:

> then do"Rename" & "Save as" several times without click OK (but cancel

 so it was the last used directory displayed for `c:\faa\c.txt`.

donho:
> Isn't it the expected behavior? 

No. It should always "c:\temp\" since "c:\temp\a.txt," was the last saved file. On `b.txt` & `c.txt` the "save as" dialog is only evoked but not confirmed (ie. OK button pushed down). Therefore the paths of these 2 files should not be considered as last used directories.

> "Save as" follows the last used directory

Not if the Cancel button in Save As dialog is clicked.

mere-human:
> It should always be "c:\temp" since "c:\temp\a.txt," was the last saved file. On `b.txt` & `c.txt` the "save as" dialog is only evoked but not confirmed (ie. OK button pushed down). Therefore the paths of these 2 files should not be considered as last used directories.

This is a different problem.
I can reproduce it starting from Notepad++ version 8.0 (portable). It is not present in version 7.9.5.
I would suggest to create a new issue and fix it separately. What do you think?

mere-human:
Ok, I actually don't want to wait for response or create a new issue myself.
Moreover, the fix is 1 line of change.
So I'll fix it here rightaway.

mere-human:
I updated How to test comment [above](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10100#issuecomment-873587168).
Also, I found out that the behavior of _Remember last used directory_ is not properly documented: https://npp-user-manual.org/docs/preferences/#default-directory

donho:
@mere-human 
>Also, I found out that the behavior of Remember last used directory is not properly documented: https://npp-user-manual.org/docs/preferences/#default-directory

Which part?
But for me, the current behaviour is not logical. I created an issue for that:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10115

mere-human:
> Which part?

I described it in this issue https://github.com/notepad-plus-plus/npp-usermanual/issues/258
Basically, the part about changing the current directory only if the dialog was confirmed is missing.

> But for me, the current behaviour is not logical. I created an issue for that:

Yes, I agree. This is a regression in v8.0.

PR #10102, opened 2021-07-04
Open a file from command line by applying an existent UDL via UDL name

donho:
Add command line flag `-udl="UDL name"` to open a file by applying an existent UDL via its name.


PowerEditor/src/winmain.cpp
```diff
@@ -482,6 +483,19 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
 		nppParameters.setTitleBarAdd(titleBarAdditional);
 	}
 
+	generic_string udlName;
+	if (getParamValFromString(FLAG_APPLY_UDL, params, udlName))
+	{
+		if (udlName.length() >= 2)
```
mere-human, line 489: This condition will allow an empty name like `""`. Is it expected?

PR #10104, opened 2021-07-05
Update Brazilian Portuguese translation for Notepad++ 7.9.5

luxyapp:


donho:

luxyapp:
Updated!

PR #10106, opened 2021-07-05
Dark mode listview header

ozone10:
Use different approach to make listview header dark. 

@donho 
This will allow highlight, hot-tracking and sort arrow.

image

image

btw
I think I have figured how to make dark mode restartless, still need some code cleaning but I think I could do PR in few days.
Still some issue with scrollbar in treeview. But for other it is working.

Also dark mode seems to work in Windows 11.


donho:
@ozone10 
Bravo! It's even better in both light & dark mode!

> I think I have figured how to make dark mode restartless, still need some code cleaning but I think I could do PR in few days.

Even for the titlebar? It'll be awesome !

> Also dark mode seems to work in Windows 11.

That's a good news - because the dark mode use some undocumented API - I hope that they will become official ones.

PR #10108, opened 2021-07-05
Add global regex unitTest via UDL way

donho:
Use User Defined Language to have function list on Unit-tests.
It could reduce regression happening while the code of boost regex or the related code is changed.


donho:
@chcg 
I just realized that the script of AppVeyor doesn't take into account of the Unit-test result.
The unit-tests (`unitTestLauncher.ps1`) exit with `0` if everything is OK, other `1`, `-1` or `-2`.

However, everything is fine except the new added unit-test `udl-regexGlobalTest`, which works fine when I test locally.
What am I missing?

chcg:
@donho See https://help.appveyor.com/discussions/problems/16519-failure-in-ps1-script-results-in-successful-build and https://help.appveyor.com/discussions/problems/4498-powershell-exception-in-test_script-does-not-fail-build there seems to be a strange behaviour with powershell error codes.
You already added a solution for that I see.

-1 mean that unitTest.expected.result and unitTest.result.json differ.

donho:
@chcg 
> See https://help.appveyor.com/discussions/problems/16519-failure-in-ps1-script-results-in-successful-build and https://help.appveyor.com/discussions/problems/4498-powershell-exception-in-test_script-does-not-fail-build there seems to be a strange behaviour with powershell error codes.

It seems the 2nd is the solution. At least the exit code of our unit-tests are taken into account now. 

> -1 mean that unitTest.expected.result and unitTest.result.json differ. Where is the udl regexGlobalTest of
loaded from? Shouldn't it be at:

The new add flag `-udl="regexGlobalTest"` no more need UDL named "regexGlobalTest" but only the parser with the same ID to make function list work.
The files differ could means the generated result in the test is empty - that means the function list is empty after applying the parser. However, I did test it locally and it does work on my PC (A good excuse for all programmers :P ). I'm trying to figure out what it is the difference between my PC & build system (Appveyor). Please let me know if you see somethings with your fresh eyes.

donho:
Just realized I've forgotten a small modif while I was replying you.
Committed and building... let's wait & see.

donho:
Hallelujah! It passed! Thank you @chcg !

PR #10111, opened 2021-07-05
Update russian.xml v8.1.2

GOLEM777:


donho:

PR #10112, opened 2021-07-06
Update Slovak translation

kubalav:


donho:

kubalav:
@donho Slovak translation done!

PR #10128, opened 2021-07-09
Make Notepad++ color customizable

donho:
Notepad++ is dark-modable now, but why not in green/purple/yellow/red/blue/etc... mode?
This PR is try to make Notepad++ color customizable.
image


donho:
@ozone10 
I made all members in `darkColors` structure more blue to see the result:

image

As you can see in the screenshot, there are at least 3 places remain unchanged, and it seems the 3 API are the undocumented ones, controlled by WDM. Do you know how to change the colors?

ozone10:
Unfortunately, unless we owner draw it, it is not possible. And for title I don't think it's possible.

donho:
@ozone10 
Then these colors won't be changed and I will make only dark mode but with the different colors for choice.
Thank you for your reply.

mere-human:
> This PR is try to make Notepad++ color customizable.

This looks nice even though other areas aren't customizable!

donho:
@mere-human 
More color tones supported. Check it out ;)

mere-human:
@donho 
> @mere-human
> More color tones supported. Check it out ;)

I checked it out. Looks good!
image

However, for some reason, when I switched from default Black (Dark mode) to e.g. Red tone, the editor style became default white as you can see above.

mere-human:
Also, how about to show small color swatches near each tone here?
image
For example, like here in the Style Configurator:
image

GOLEM777:
Also, how about to show add color picker for custom customizing self color?

PR #10134, opened 2021-07-09
Use `!` instead of `not` as per guideline # 13

durmusk:
Contradicts guideline # 13 "Use `!` instead of `not`"


donho:
According the guideline, use issue instead of PR for such request.

sasumner:
I made a PR for this change: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10158

PR #10141, opened 2021-07-11
Improve DarkMode toggling

ozone10:
Make treeview scrollbar style more reliable in light mode.
Disable hottracking for treeview with classic style.
Make choosing dark tone to not require turning back to light mode.

image

image

image


PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -100,15 +100,20 @@ IDD_PREFERENCE_SUB_DARKMODE DIALOGEX 0, 0, 455, 185
...
+    CONTROL         "Green tone",IDC_RADIO_DARKMODE_GREEN,  "Button",BS_AUTORADIOBUTTON,35,70,100,10
+    CONTROL         "Blue tone",IDC_RADIO_DARKMODE_BLUE,    "Button",BS_AUTORADIOBUTTON,35,85,100,10
+    CONTROL         "Purple tone",IDC_RADIO_DARKMODE_PURPLE,"Button",BS_AUTORADIOBUTTON,35,100,100,10
+    CONTROL         "Cyan tone",IDC_RADIO_DARKMODE_CYAN,    "Button",BS_AUTORADIOBUTTON,35,115,100,10
+    CONTROL         "Olive tone",IDC_RADIO_DARKMODE_OLIVE,  "Button",BS_AUTORADIOBUTTON,35,130,100,10
+
+    GROUPBOX        "Treeview Style",IDC_DARKMODE_TV_GB_STATIC,165,20,105,65,BS_CENTER
+    CONTROL         "Classic",IDC_RADIO_DARKMODE_TVCLASSIC, "Button",BS_AUTORADIOBUTTON | WS_GROUP | WS_TABSTOP,175,35,80,10
+    CONTROL         "Light",IDC_RADIO_DARKMODE_TVLIGHT,     "Button",BS_AUTORADIOBUTTON,175,50,80,10
+    CONTROL         "Dark",IDC_RADIO_DARKMODE_TVDARK,       "Button",BS_AUTORADIOBUTTON,175,65,80,10
```
donho, line 116: Treeview style cannot be disassociated from dark mode, not only it could confuse users, but also for preventing from the following case:
 
image

The correct behaviour would be: when dark mode is enabled, treeview will be switched to "Dark" option automatically; When dark mode is disabled, treeview will be switched to "Light" or "Classic" (I guess "Light" looks better, but it's up to you).

donho (changes requested):
Remove Treeview Style from Preferences dialog.
Use dark tree style/light (or classic) when Notepad++ dark mode is enabled/disabled.

PowerEditor/src/WinControls/Preference/preference.rc
ozone10: I can make it automatic,

but the issue here is that user can use dark theme with light mode and vice versa
image
This was mentioned in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10061

Would it be better if it is moved in style configurator

ozone10:
@donho 
I made it so toggling between dark and light mode will reset treeview style to classic, but still there is issue I mention above (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10141#discussion_r667479341)

PowerEditor/src/WinControls/Preference/preference.rc
donho: Let's keep KISS - Dark option belongs to the dark mode.

donho:
@ozone10 
> I made it so toggling between dark and light mode will reset treeview style to classic, but still there is issue I mention above (#10141 (comment))

Then let's use classic style under light mode.

PowerEditor/src/WinControls/Preference/preference.rc
ozone10: @donho 
> Let's keep KISS principle - Dark option belongs to the dark mode.

There might be someone who want dark mode but use light theme.
image

So should I remove this treeview style setting from preferences to make it hidden option (so it can be changed only via editing config.xml).

donho: Treeview style GUI option has to be removed for sure.
Hidden option is not necessary in my opinion as well.
KISS should not be applied to GUI, but to the source code as well.

ozone10: > Treeview style GUI option has to be removed for sure.
> Hidden option is not necessary in my opinion as well.
> KISS should not be applied to GUI, but to the source code as well.

Ok so should I keep only classic style, scrollbar fixes and enhancement to dark tone switching.

ozone10: Ok I will do commit later, some IRL stuff.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -449,7 +449,7 @@ BEGIN
     CONTROL         "Peek on tab",IDC_CHECK_ENABLEDOCPEEKER,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,269,58,140,10
     CONTROL         "Peek on document map",IDC_CHECK_ENABLEDOCPEEKONMAP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,269,71,140,10
 
-	// "Enable Notepad++ auto-updater" should be always the 1st item, because it'll be hidden if GUP.exe is absent
+    // "Enable Notepad++ auto-updater" should be always the 1st item, because it'll be hidden if GUP.exe is absent
     CONTROL         "Enable Notepad++ auto-updater",IDC_CHECK_AUTOUPDATE,"Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP, 37, 94, 210, 10
     CONTROL         "Mute all sounds", IDC_CHECK_MUTE_SOUNDS, "Button", BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP, 37, 109, 217, 10
     CONTROL         "Autodetect character encoding",IDC_CHECK_DETECTENCODING,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,124,217,10
```
donho, line 455: restore `preference.rc`
I'm modifying the same part so there would be some conflicts to me if it remains this way.

PowerEditor/src/WinControls/TreeView/TreeView.cpp
```diff
@@ -26,9 +26,9 @@ void TreeView::init(HINSTANCE hInst, HWND parent, int treeViewID)
 	Window::init(hInst, parent);
 	_hSelf = ::GetDlgItem(parent, treeViewID);
 
-	auto treeViewStyles = WS_HSCROLL | WS_TABSTOP | TVS_LINESATROOT\
+	const auto treeViewStyles = WS_HSCROLL | WS_TABSTOP | TVS_LINESATROOT\
 						| TVS_HASLINES | TVS_HASBUTTONS | TVS_SHOWSELALWAYS\
-						| TVS_EDITLABELS | TVS_INFOTIP | TVS_TRACKSELECT;
+						| TVS_EDITLABELS | TVS_INFOTIP/* | TVS_TRACKSELECT*/;
```
donho, line 31: `/* | TVS_TRACKSELECT*/` can be removed.

PowerEditor/src/WinControls/TreeView/TreeView.cpp
```diff
@@ -44,6 +44,7 @@ void TreeView::init(HINSTANCE hInst, HWND parent, int treeViewID)
 							nullptr);
 
 	NppDarkMode::setDarkTooltips(_hSelf, NppDarkMode::ToolTipsType::treeview);
+	//NppDarkMode::setTreeViewStyle(_hSelf);
```
donho, line 47: This line can be removed.

donho (changes requested):
Please follow the comments for the modification.

donho:
@ozone10 

BTW, there's no dark mode but only classic mode:

image

Is there something missing?

ozone10:
> BTW, there's no dark mode but only classic mode:

> Is there something missing?

I removed it because with light theme expand symbol will became invisible.
image

donho:
@ozone10 
> I removed it because with light theme expand symbol will became invisible and issue with highlight.

Could the treeview dark style background be light gray so it can suit both dark theme and light theme?

donho (approved):

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -939,20 +941,17 @@ INT_PTR CALLBACK DarkModeSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
 					NppDarkMode::setDarkTone(nppGUI._darkmode._colorTone);
 					changed = true;
+					forceRefresh = true;
 					break;
 			}
 
 			if (changed)
 			{
-				NppDarkMode::refreshDarkMode(_hSelf);
+				NppDarkMode::refreshDarkMode(_hSelf, forceRefresh);
```
donho, line 954: Nice job! 👍

ozone10:
@donho 
> Could the treeview dark style background be light gray so it can suit both dark theme and light theme?

Background is using color from style configurator. 

I am thinking of using different approach.
Add option to use non-classic style in preference or style configurator, and dynamically switch dark/light style based on background.

To determine brightness using formulas from https://stackoverflow.com/questions/596216/formula-to-determine-perceived-brightness-of-rgb-color .

donho:
@ozone10 
> To determine brightness using formulas from https://stackoverflow.com/questions/596216/formula-to-determine-perceived-brightness-of-rgb-color .

Looks quit overkill... but if the code is not large and no performance issue, why not?

PR #10158, opened 2021-07-14
Change not usage to !

sasumner:
Because another guideline says to use "!" instead of "not", examples that are called "Good" should not use "not".


CONTRIBUTING.md
```diff
@@ -168,7 +168,7 @@ In short: The easier the code review is, the better the chance your pull request
 
   * ###### Good:
     ```cpp
-    if (not string.empty())
+    if (!string.empty())
```
donho, line 171: 👍

PR #10168, opened 2021-07-15
Deleting from search results removes hierarchically correctly

### LINKED ISSUE #8160: Deletion of a last "child" line in "Find result" should also delete ancestors
1. Start with this text in a Notepad++ tab:

```
Jack and Jill
Went up the hill
To fetch a pail of water
Jack fell down
And broke his crown,
And Jill came tumbling after.
```

2. Run the _Find All in Current Document_ search shown in the following and observe the results (also shown):

image
 
3.  Click on the line containing `Line 5: And broke his crown,` in the _Find result_ window to make it active.  Press the Delete key on the keyboard and observe the following:

image

The `Line 5: And broke his crown,` line was removed (correct).

Because the `Line 5: And broke his crown,` line was the only "child" of the `new 1 (1 hit)` "parent", the `new 1 (1 hit)` line should also have been removed.  

In turn, because the `new 1 (1 hit)` line should have been removed and it is the only "child" of the `Search "crown" (1 hit in 1 file)` line, the `Search "crown" (1 hit in 1 file)` ("grandparent") line should also have been removed.

4. Clear all _Find result_ window results by right-clicking the window and choosing _Clear All_.

5. Repeat step 2 and again observe the results screenshot from step 2.

6. Click on the line containing `new 1 (1 hit)` in the _Find result_ window to make it active.  Press the Delete key on the keyboard and observe the following:

image

The `new 1 (1 hit)` line was removed (correct).

Because the `new 1 (1 hit)` line was the only "child" of the `Search "crown" (1 hit in 1 file)` "parent", the `Search "crown" (1 hit in 1 file)` line should also have been removed. 

### Debug Information

Notepad++ v7.8.6   (32-bit)
Path : C:\.........\npp.7.8.6.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039

sasumner:
Fix #8160

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -546,40 +546,78 @@ void Finder::gotoFoundLine()
 
 void Finder::deleteResult()
 {
-	auto currentPos = _scintView.execute(SCI_GETCURRENTPOS); // yniq - add handling deletion of multiple lines?
+	auto currentPos = _scintView.execute(SCI_GETCURRENTPOS);
 
-	auto lno = _scintView.execute(SCI_LINEFROMPOSITION, currentPos);
-	auto start = _scintView.execute(SCI_POSITIONFROMLINE, lno);
-	auto end = _scintView.execute(SCI_GETLINEENDPOSITION, lno);
-	if (start + 2 >= end) return; // avoid empty lines
+	auto lineOfDelPress = _scintView.execute(SCI_LINEFROMPOSITION, currentPos);
+
+	auto start = _scintView.execute(SCI_POSITIONFROMLINE, lineOfDelPress);
+	auto end = _scintView.execute(SCI_GETLINEENDPOSITION, lineOfDelPress);
+	if (start + 2 >= end) return; // don't process a delete request on the trailing blank line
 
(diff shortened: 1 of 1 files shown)


sasumner:
If this is used to remove all content from a _Find in these search results..._ generated window, perhaps the panel tab itself should be closed?  (this PR does not do this):

image

mere-human:
> perhaps the panel tab itself should be closed?

It could look a bit unexpected, as for me. Despite the fact that empty search results window isn't very useful.

sasumner:
>>perhaps the panel tab itself should be closed?

>It could look a bit unexpected, as for me. Despite the fact that empty search results window isn't very useful.

I can't tell with 100% certainty if that is a vote for keeping the (empty) panel, or auto-closing it.
I _think_ it is a vote to keep it. :-)

mere-human:
> I can't tell with 100% certainty if that is a vote for keeping the (empty) panel, or auto-closing it.

Yes, I'm for keeping it open.

sasumner:
Note that the "primary" search results window won't be auto-closed even if empty, because it will receive future standard search results.  A "secondary" results window only ever received a single result, so I thought perhaps it was a more obvious choice for autodeletion if that one result is ever removed.

But I do agree that keeping it is probably for the best.

donho:
Such improvement, why not? But what I don't like is quite massive modification which could bring the regression for small gain.

mere-human:
> Such improvement, why not? But what I don't like is quite massive modification which could bring the regression for small gain.

But why rejecting it without a strong reason? We could test different scenarios. There was a problem here anyway.

PR #10170, opened 2021-07-15
Allow CtrlTab to switch tabs in Find dialog

### LINKED ISSUE #10019: CTRL+F should upgrade to CTRL+H when pressed
## Currently
When on the CTRL+F window, pressing CTRL+H does nothing (or something unexpected).

## Expected
What _should_ intuitively happen like in most (if not all) other editors:

When on the CTRL+F window, **pressing CTRL+H should "upgrade" to what should normally show when pressing CTRL+H**.

## Debug
Notepad++ v7.9.5   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2009
OS Build : 19043.1055
Current ANSI codepage : 1252
Plugins : JSLintNpp.dll JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
Fix #10019

Doesn't implement exactly what the requester wanted, but implements a way to switch tabs once _Find_ dialog is focused that was impossible before.

Ctrl+Tab will switch to the right one tab.
Ctrl+Shift+Tab (awkward so maybe rarely used) will switch to the left one tab.

A lot of code for a simple thing (runs high risk of rejection), but it is easily validated to not break existing functionality:
* test Delete key functionality to delete an entry when a combobox is dropped down (still works)
* test Ctrl+Backspace functionality to delete a word in an edit box (still works)

Additionally, the groundwork is now put down for easily implementing new behaviors for more control types.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -3575,15 +3575,12 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwndEdit, UINT message, WP
 		delLeftWordInEdit(hwndEdit);
 		return 0;
 	}
-	else
+	else if (message == WM_KEYDOWN && wParam == VK_DELETE)
 	{
-		// check for Delete key press in dropped combobox in order to remove an item
+		// find the associated combobox for the edit control that triggered this
 
+		static HWND hwndCombo = 0;
 
-
-
-		/*
-		HWND hCombo = 0;
 		EnumChildWindows(pFindReplaceDlg->_hSelf, [](HWND hwnd, LPARAM lParam) {
 			HWND hwndEdit = reinterpret_cast<HWND>(lParam);
 			wchar_t className[32] = { 0 };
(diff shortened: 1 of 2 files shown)


donho:
I don't expect the modification is so important for such small gain...

donho:
@sasumner 
Ctrl-TAB is nice-to-have small feature, but not nice enough to take the regression with such important modification.
Sorry, but I have to reject this PR.

mere-human:
I tried it locally. It works very good.
Old _Tab_ behavior to select a particular control is preserved.
_Delete_ and _Ctrl + Backspace_ work as before.
Here is a small demo:

https://user-images.githubusercontent.com/9664141/125902441-ce81640b-f124-48f6-9b10-fbca90eca2c2.mp4

Note that tabs are selected using _Ctrl + Tab_ (the cursor isn't moving).

donho:
@mere-human 
The `Ctrl-H` is more intuitive than `Ctrl-TAB`, Here is the implemetation of `Ctrl-H` detected in Find dialog, I will add the other tabs' detection:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10177

@sasumner 
Sorry I was not able to indicate this way to you. I didn't recall that the shortcut of Find dialog command has been already capturable by Find dialog (implemented 5 years ago in ac871cacd48434375f81d28773dc316677716b22).

ValZapod:
> The `Ctrl-H` is more intuitive than `Ctrl-TAB`

No, it is not. The chrome ctrl-tab, ctrl-shift-tab is more intuitive.

PR #10177, opened 2021-07-16
Use Ctrl-H to switch to Find-Replace dialog from the other tabs

### LINKED ISSUE #10019: CTRL+F should upgrade to CTRL+H when pressed

donho:
Add find/replace/... shortcuts detection capacity in Find Replace dialog.

Fix #10019

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1208,7 +1208,6 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 
 				case IDM_SEARCH_FIND:
 					enableReplaceFunc(false); // enable relace false so only find
-					goToCenter();
 					return TRUE;
 
 				case IDM_SEARCH_REPLACE:
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1210,6 +1210,10 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 					goToCenter();
 					return TRUE;
 
+				case IDM_SEARCH_REPLACE:
+					enableReplaceFunc(true);
+					return TRUE;
+
 				case IDREPLACE :
 				{
 					std::lock_guard<std::mutex> lock(findOps_mutex);
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1207,11 +1207,20 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 				return TRUE;
 
 				case IDM_SEARCH_FIND:
+					enableReplaceFunc(false); // enable relace false so only find
 					goToCenter();
 					return TRUE;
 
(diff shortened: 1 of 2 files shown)


PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -582,11 +581,13 @@ void Accelerator::updateShortcuts()
...
+	size_t nbFindReplaceAcc = findReplaceAcc.size();
+	if (nbFindReplaceAcc)
 	{
-		ACCEL *tmpFindAccelArray = new ACCEL[1];
-		tmpFindAccelArray[0] = *pSearchFindAccel;
-		_hFindAccTab = ::CreateAcceleratorTable(tmpFindAccelArray, 1);
+		ACCEL* tmpFindAccelArray = new ACCEL[nbFindReplaceAcc];
+		for (size_t i = 0; i < nbFindReplaceAcc; ++i)
+			tmpFindAccelArray[i] = findReplaceAcc[i];
+		_hFindAccTab = ::CreateAcceleratorTable(tmpFindAccelArray, nbFindReplaceAcc);
```
mere-human, line 590: warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1207,7 +1207,19 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 				return TRUE;
 
 				case IDM_SEARCH_FIND:
-					goToCenter();
```
mere-human, line 1210: By the way, how is it managed by the system?
Is there a way to "find" the Find dialog now?

mere-human:
I verified it locally.
It actually works and allows you to switch from Find (_Ctrl+F_) to Replace (_Ctrl+H_) which is good.
It kind of solves the problem.

However, as for me, _Ctrl+Tab_ would still be more logical.
Since, I personally don't remember all the shortcuts.

Also, there is no shortcut for _Find in Projects_ so there is no way to select it using a keyboard.
image

image

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: I tested roughly. If Find dialog is off screen but Notepad++ is on the screen, Ctrl-F will bring back Find dialog in the main frame zone of Notepad++ with or without dialog being focused.

donho:
> However, as for me, Ctrl+Tab would still be more logical. Since, I personally don't remember all the shortcuts.

Ctrl-F & Ctrl-H are much more intuitive, that's why such request exists. For the Ctrl-Shift-F (Find in files) & Ctrl-M (Mark), users will check the shorcuts if needed. 

These shortcuts are already used, so it makes more sense to me to add them. Not mentioning without the lines and lines added code which could bring bugs/regression.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
mere-human: Indeed, I can confirm.

PR #10180, opened 2021-07-16
Add C++ unit tests to the project

### LINKED ISSUE #9705: Integrate unit tests to the project
### Description of the Issue
I suggest integrating unit tests into the project. This has been brought up in some discussions, for example, here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9674#issuecomment-803472169

Unit tests have a lot of advantages:
* ensure individual parts are correct
* helps spotting problems early on (after modifications are made)
* define the desired behavior
* reduce risk of regressions
* make code modifications and refactoring safer

**What needs to be done:**
1. Choose a unit test framework and add it to the project.
2. Create an initial set of unit tests (as a proof of concept).
3. Integrate unit tests into the build process. This can be a part of the local build, AppVeyor build, or both.

I would start with an area familiar to me (file dialogs) and then probably cover other parts of the project.

### Choosing unit test framework
There are several commonly used C++ test frameworks.

My preferred option is [GoogleTest](https://github.com/google/googletest) framework:
* Rich set of API and predefined checks
* Death tests, fatal and non-fatal tests
* Good documentation and easy to use
* Automatic test case registration
* Built-in mock objects support
* Actively supported
* Visual Studio integration, a lot of command-line options
* I have a lot of experience using it

Other options: [Boost.Test](https://www.boost.org/doc/libs/1_72_0/libs/test/doc/html/boost_test/intro.html), [Catch2](https://github.com/catchorg/Catch2), [CppUnit](https://freedesktop.org/wiki/Software/cppunit/), [cpputest](https://github.com/cpputest/cpputest), etc.
See references below for more information.

Main disadvantages:
* No built-in mock objects support
* Dependency on other libraries (such as Boost)
* Outdated, poor documentation
* Not xUnit
* I have less (Boost.Test) or no experience with these

### Notes and References
The goal would be to increase [test coverage](https://en.wikipedia.org/wiki/Code_coverage) of the project.
We even can add a badge showing coverage.

[Mock object](https://en.wikipedia.org/wiki/Mock_object) simulates the behavior of a real object. It helps to test one part of the code in isolation from other parts that should not be tested in a particular case. This helps a lot to add tests incrementally.

Unit test frameworks:
* [Comparison of C++ unit test frameworks | Stack Overflow](https://stackoverflow.com/questions/242926/comparison-of-c-unit-test-frameworks)
* [Most Popular C++ Unit Testing Frameworks | Reddit](https://www.reddit.com/r/cpp/comments/4e9afx/most_popular_c_unit_testing_frameworks/)
* [List of unit testing frameworks | Wikipedia](https://en.wikipedia.org/wiki/List_of_unit_testing_frameworks#C++)
* [Write unit tests for C/C++ in Visual Studio | Microsoft](https://docs.microsoft.com/en-us/visualstudio/test/writing-unit-tests-for-c-cpp?view=vs-2019)

### Debug Information

```
Notepad++ v7.9.4   (32-bit)
Build time : Mar 15 2021 - 01:00:14
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2009
OS Build : 19042.867
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll HexEditor.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll NppMarkdownPanel.dll NppTextFX.dll PythonScript.dll 
```

mere-human:
Use Google Test framework package (via NuGet).
Add test cases for some functions that are mainly used in the file dialog.
This is mainly a proof of concept and can be extended further.

UnitTests project includes the code to test (.cpp) directly.
Other approaches such as linking with .obj files or creating a static .lib were not implemented because of some reasons.
Main reasons are:
- keep the code simple
- avoid an error during linking because Notepad++ uses static CRT (more info [here](https://docs.microsoft.com/en-us/answers/questions/240332/assert-when-using-mtd-dynamic-library.html))

Fix #9705

Relevant links:
- [How to use Google Test for C++ in Visual Studio](https://docs.microsoft.com/en-us/visualstudio/test/how-to-use-google-test-for-cpp?view=vs-2019)
- [Run unit tests with Test Explorer](https://docs.microsoft.com/en-us/visualstudio/test/run-unit-tests-with-test-explorer?view=vs-2019)


donho:
It's miles & miles of news files + a lot of modification...
Unit-tests are supposed for testing each unit for prevent from the regression, but I smells the regression from your modification...

PowerEditor/visual.net/notepadPlus.sln
```diff
@@ -44,6 +76,28 @@ Global
...
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Debug|Win32.ActiveCfg = Debug|Win32
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Debug|Win32.Build.0 = Debug|Win32
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Debug|x64.ActiveCfg = Debug|x64
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Debug|x64.Build.0 = Debug|x64
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Release|ARM64.ActiveCfg = Debug|Win32
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Release|ARM64.Build.0 = Debug|Win32
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Release|Win32.ActiveCfg = Release|Win32
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Release|Win32.Build.0 = Release|Win32
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Release|x64.ActiveCfg = Release|x64
+		{2581D488-956B-4D3B-BEE4-A9F1164759E2}.Unicode Release|x64.Build.0 = Release|x64
```
donho, line 100: Why these above added lines?

donho:
I'm not sure to integrate the Unit-Tests system - it's overhead and bring the small benefice for the project.

mere-human:
> It's miles & miles of news files + a lot of modification...

I know. But the new files are added for these reasons:
* test cases code (in UnitTests directory)
* moving the code from one place to another (e.g. moving from `CustomFileDialog.cpp` to `COM.h`)
* auxiliary files for building (e.g. `UnitTests.vcxproj`)

Now there are 7 test cases. 
I can reduce the new code to just 1 test case as a proof-of-concept. 
Then add more test cases in a separate PR. 
Maybe, it will help to make it easier to understand.
Will it be better?

There were PRs with a lot of changes before. For example, do you remember when Notepad++ switched from SciLexer DLL to static library? Or when Boost.Regex was added to the repository? There were a lot of changes too.

mere-human:
> but I smells the regression from your modification...

The only change to Notepad++ here is moving code from one file to another without any changes to the code.

mere-human:
> I'm not sure to integrate the Unit-Tests system - it's overhead, build time consuming and bring the small benefice for the project.

But isn't any change has a chance of regression? Even introducing the dark mode to the Preferences dialog has a chance of regression.
The PR looks complex because it adds a whole new system.
I created it specifically because I want to _reduce_ the amount of regressions.

Here is the result from AppVeyor:
image
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/40025118/job/vsh2ljwkc4sk9uv3
It means that functions are executed with sample inputs and the outputs are checked.
So any unexpected change to the code that make it work wrong or not as we expected will make unit tests fail.
So then we will know that something went wrong at the stage of running unit tests not at the stage of release candidate.
I think it is very good.

PowerEditor/visual.net/notepadPlus.sln
mere-human: These are configurations for UnitTests projects generated by Visual Studio. I can try to reduce it.

ArkadiuszMichalski:
Without such tests, any change is a regression risk, and the repeated of the sentence like "... quite massive modification which could bring the regression, so reject" can discourage any, even the most stubborn programmer (from fixing anything). Currently  we have around 1400 open issues, I wonder how many of them should be closed, because due to extensive changes they will not pass anyway.

donho:
> repeated of the sentence like  "... quite massive modification which could bring the regression, so reject"

Nope, The reason of reject is not only the regression probability, but especially the complexity of maintenance, for the feature I judge not worth.

mere-human:
> but especially the complexity of maintenance, for the feature I judge not worth.

I'm here to answer any questions or to help educating about how this should be maintained.
I'm planning to work on it. 
But even if there's nobody left who knows how to maintain it, the unit tests could be safely removed from the project without any harm to the rest of the code.

ArkadiuszMichalski:
>@donho 
Nope, The reason of reject is not only the regression probability, but especially the complexity of maintenance, for the feature I judge not worth.

Maybe add more `accepted` label to the issues you consider most expected? Then there would be more choice for people interested in fixing something, with less risk that their work will be rejected (compared to when they take a completely random issue). Currently, there is little of it:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues?q=is%3Aissue+is%3Aopen+label%3Aaccepted

donho:
@mere-human 
>> but especially the complexity of maintenance, for the feature I judge not worth.

>I'm here to answer any questions or to help educating about how this should be maintained.

Well, it's not what I meant. I was answering @ArkadiuszMichalski 's question for the criteria of PR acceptance.

OTOH, we have had a talk about my thought for Unit-test. My opinion remain the same:
I saw a lot of project use unit-test and make it cover 100% as goal of the project. A lot of refactoring for making unit-testable and only for that. 
If there's a huge non-graphic framework used by several products, I would say it's a good idea to unit-test the it if the framework keep evaluating - in Notepad++ project the majority of are graphic components, and I really don't want the refactoring entire project for the 20-40% unit-testable part. Unit-test is a way for preventing regression, not the goal of the project.

So sorry for your effort in vain - To me the disavantage (change the coding logic, refactoring the current code) of unit-test for this project will be greater than its advantage.

donho:
@ArkadiuszMichalski 
> Maybe add more `accepted` label to the issues you consider most expected? Then there would be more choice for people interested in fixing something, with less risk that their work will be rejected (compared to when they take a completely random issue).

I could do more effort about that. 
However, it won't be binary like that:
A "Nice to have" feature could be implemented in several ways.
My criteria is KISS & lazy (easy to review).
If it's not possible, isolate the code to prevent from spaghetti code could help.
And it's not all... A "Nice to have" feature could be present intelligently in term of usability, or not.
For accepting the "Nice to have" features, I need not only "I have a good feature", but also a brainstorming of contributors to integrate them intelligently. If it's not possible, it may be better do without them.

Sincerely, I would like make every contributor happy if I could. But I can't.

image

Notepad++ is not yet like the 3rd one in the above picture, but not far from it. There are so many features to implemented, the filtering is necessary.

mere-human:
> change the coding logic, refactoring the current code

There are no changes in logic. Just moving the code from one file to another.

> in Notepad++ project the majority of are graphic components, and I really don't want the refactoring entire project for the 20-40% unit-testable part

I disagree. There are a lot of bugs in code that is not related to GUI. The amount of such bugs could be minimized if there were unit tests.
I don't even say that it's possible to test GUI with unit tests (rendering to bitmap, mocking UI interfaces, etc).

For example, lets take a look at commit history. 
For the past ~3 weeks, these commits are not related to GUI and could be covered with tests:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3482673fd8cdf5d15f365470c4a12a3fd7feae5f
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5e28b47ff88e9b8b779923de136452a926e5d6d9
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0472fd7f2aa16f9b295a007739860acdbdf2d23

My opinion is that such program as Notepad++ could drown in bugs/regressions without tests. 
Is it drowning already? What is the way to keep it stable when introducing new features? Or when fixing some bugs?
How many of 1.4k open bugs are not related to GUI?

PR #10183, opened 2021-07-18
Retain Functions list scrollbar position on reload

### LINKED ISSUE #9659: Functions list keeps scrolling to the top every time the current file is saved
Simple bug: the functions list on the right side keeps scrolling back to the top every single time I save the current file. Because I work in technology I don't trust technology and I'm constantly pressing CTRL+S to negate someone else's irresponsibility from negating my progress. At 2560x1440 my functions list for many common files I edit is vertically longer than the available space and this is exceptionally obnoxious especially right now since I'm going back and forth between common functions and I have to stop my mind-flow and waste time thinking about having to scroll to get the functions back in to view to quickly switch over.

Here is the debug information:
---------------------------------------------------------------------

Notepad++ v7.9.1   (64-bit)
Path : C:\MEDIA\OFFICE\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.418
Current ANSI codepage : 1252

Ashfaaq18:
fix #9659

Saving the file before fix does not retain scroll bar position:

funcListSaveFix_before

Saving the file after fix retains scroll bar position:

funcListSaveFix_after

### DIFF
--- a/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
+++ b/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp
@@ -315,6 +315,13 @@ bool FunctionListPanel::serialize(const generic_string & outputFilename)
 
 void FunctionListPanel::reload()
 {
+	bool isScrollBarOn = GetWindowLongPtr(_treeView.getHSelf(), GWL_STYLE) & WS_VSCROLL;
+	//get scroll position
+	if (isScrollBarOn)
+	{
+		GetScrollInfo(_treeView.getHSelf(), SB_VERT, &si);
+	}
+
 	// clean up
 	_findLine = -1;
 	_findEndLine = -1;
@@ -396,6 +403,12 @@ void FunctionListPanel::reload()
 
 	// invalidate the editor rect
 	::InvalidateRect(_hSearchEdit, NULL, TRUE);
+
+	//set scroll position
+	if (isScrollBarOn)
+	{
+		SetScrollInfo(_treeView.getHSelf(), SB_VERT, &si, TRUE);
+	}
 }
 
 void FunctionListPanel::markEntry()
@@ -498,6 +511,11 @@ void FunctionListPanel::init(HINSTANCE hInst, HWND hPere, ScintillaEditView **pp
 			_funcParserMgr.init(funcListDefaultXmlPath, funcListDefaultXmlPath, ppEditView);
 		}
 	}
+
+	//init scrollinfo structure
+	ZeroMemory(&si, sizeof(si));
+	si.cbSize = sizeof(si);
+	si.fMask = SIF_POS;
 }
 
 bool FunctionListPanel::openSelection(const TreeView & treeView)
--- a/PowerEditor/src/WinControls/FunctionList/functionListPanel.h
+++ b/PowerEditor/src/WinControls/FunctionList/functionListPanel.h
@@ -109,6 +109,7 @@ private:
 	TreeView _treeView;
 	TreeView _treeViewSearchResult;
 
+	SCROLLINFO si;
 	long _findLine = -1;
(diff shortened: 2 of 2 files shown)


Yaron10:
@Ashfaaq18,

The issue still exists if `Order by Name` is ON.

Thank you.

Ashfaaq18:
hey @Yaron10, thanks for testing it out. Seems like it is not working when the 'Sort' option is on. Ill try to patch it up.

Ashfaaq18:
hey @Yaron10, I'm assuming the 'Order by Name' you mean is the 'Sort' option in function list. I tried this with the 'Sort' option ON and its still working as intended. 
Btw, this commit was accepted and closed only recently, which was only an hour ago since you commented. I hope you pulled this commit after it was merged to the master and tested it then.

Yaron10:
Hello @Ashfaaq18,

Thanks for replying and for your work.

Yes, I meant `Sort by Name`.

***
When you first open the `Function List`, you get the following state.

תמונה

Double-click `isUrl`, edit that function body, Save and return focus to the FL by clicking on its caption.

**Result:** the root `Notepad_plus.cpp` node is selected. - Wrong.

***
If you expand the `Notepad_plus` node, you get the following state.

תמונה

Double-click `replaceInFiles`, edit that function body, Save and return focus to the FL by clicking on its caption.

**Result:** the `replaceInFiles` node is selected. - Correct.

If you repeat the last steps when `Sort by Name` is ON, the root `Notepad_plus.cpp` node is selected. - Wrong.

I've tested those cases _with and without your PR_ and got the same results.
I'm sure your PR does fix some cases. Could you enlighten me?

Ashfaaq18:
Hey, @Yaron10. Thanks for the information on how to recreate the issue. I did manage to recreate the problems you've mentioned. 

My PR simply,
Stores the scroll bar position before a reload takes place to the Function list, 
and sets back the stored scroll bar position after the reload has taken place.
I did not alter any node highlighting here.

In this issue you've mentioned. It seems like the problem is the default highlighting of the root node, which I did not alter. This default highlighting must be happening after the reload takes place, thus scrolling back to the top and also weirdly collapsing the expanded function trees when 'Sort by Name' is ON.

Since this is currently a problem to be fixed, would you be willing to open a new issue on this (regarding node highlighting) ? I can work on fixing this.

Yaron10:
If you don't mind, I'd rather understand the case your PR fixes before opening a new issue.
Can you please list the exact case/STR?

Thanks again.

Ashfaaq18:
My PR fixes the following case;
Assuming 'Sort by Name' is OFF. If a function list has a visible scroll bar and if that bar is scrolled down to a certain point. The next save will keep the scroll bar in the same scrolled down position (not reset it).

in the above 3 problem scenarios you showed, only the second one worked because it has a visible scroll bar with 'Sort by Name' OFF. 
The 3rd scenario is when you told to repeat the second with 'Sort by Name' ON, in this case when the user saves, 
the function list collapses thus removing the scroll bar. This prevents my fix from working.

(btw if you have a file with a functions list longer than the window and WITHOUT collapsible tree dropdowns in it, my fix will work even with 'Sort by Name' ON because the scroll bar is still visible (there is no tree to collapse)  )

Yaron10:
👍 
Thank you.

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10280

A point to consider:
I think the functions list code requires some major optimization/refactoring regardless of those issues.
(If you open a file containing many functions, the FL may take a few seconds to load).
Fixing our issues by some patches, might slow the FL even more.

Ashfaaq18:
Thanks for creating that issue. And yeah, I totally agree with you about the function list's responsiveness. It sure needs a full rework.

PR #10194, opened 2021-07-19
Copy file names from Windows dialog

### LINKED ISSUE #10123: Feature Request: Get file paths from all open tabs
### Description of the Issue

I want have an ability to copy paths of all opened documents.
This was discussed on a community forum: 
https://community.notepad-plus-plus.org/topic/21447/how-to-get-file-paths-from-all-open-tabs
There are other suggestion there. Please, see my suggestion below.

### Actual Behavior
This copies file path only for current document:
Edit > Copy to Clipboard > Current Full File path to Clipboard

### Expected Behavior

Add a way to copy the file path for all opened document.

**Suggestion:**
Make the text from Window > Windows copyable.
It already supports multi-select. The only thing missing is Ctrl+C and put a list of of file paths to clipboard.
image

### Debug Information

Notepad++ v8   (64-bit)
Build time : May 27 2021 - 22:10:00
Path : C:\Program Files\npp-portable\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1052
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll HexEditor.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExport.dll NppMarkdownPanel.dll XMLTools.dll

mere-human:
Ctrl+C copies selected path names.
Context menu allows to copy either name or path of selected files.

Fix #10123


mere-human:
Windows dialog now supports Ctrl+C shortcut and context menu.
The context menu has 2 items that are disabled if nothing is selected.
image

Copy names gives a result like this:
```
new 1
notepadPlus.vcxproj.user
readonly.txt
new 4
```
Copy path names gives a result like this:
```
new 1
E:\Development\notepad-plus-plus\PowerEditor\visual.net\notepadPlus.vcxproj.user
E:\tmp\readonly.txt
new 4
```

### How to test:
* Open/create several of files
* Open some read-only files
* Modify some files
* Open from menu Window > Windows
* Ensure the sort functionality and the buttons works as before
* Select some files (or all via Ctrl+A)
* Ctrl+C and paste somewhere
* RMB select menu item and paste somewhere

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -335,34 +353,20 @@ INT_PTR CALLBACK WindowsDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM lPa
...
-
-						//const Buffer& buffer = _pView->getBufferAt(index);
-						BufferID bufID = _pTab->getBufferByIndex(index);
-						Buffer * buf = MainFileManager.getBufferByID(bufID);
+						generic_string text;
 						if (pLvdi->item.iSubItem == 0) // file name
 						{
-							int len = pLvdi->item.cchTextMax;
-							const TCHAR *fileName = buf->getFileName();
-							generic_strncpy(pLvdi->item.pszText, fileName, len-1);
```
mere-human, line 350: There was a buffer overrun that was revealed by my changes.
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strncpy-strncpy-l-wcsncpy-wcsncpy-l-mbsncpy-mbsncpy-l?view=msvc-160
It seems, in some cases more characters were copying than the length of the `pszText` (here or below). 
This led to some other memory being overwritten which resulted in a crash in Scintilla code:
image

Now the code uses generic_string that allocates memory dynamically and is safer than string copy functions.
The only time string copy is called is at line 401 with the proper size check.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -907,6 +937,57 @@ void WindowsDlg::doSortToTabs()
...
+		}
+
+		if (!selection.empty() && selection.back() != '\n')
+			selection += '\n';
+	}
+	if (!selection.empty())
+		str2Clipboard(selection, _hList);
+}
+
+Buffer* WindowsDlg::getBuffer(int index) const
```
mere-human, line 978: Extracted this function from the code in `case WM_NOTIFY`.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
mere-human: It was done in a separate commit to make it easier for review. So that it can be reviewed separately.

Issue #10069, opened 2021-06-24
[DarkMode] Splitter background – opinion


ozone10:
Question, this is more about aesthetic. What do you like more?

Currently splitter have lighter color and middle one doesn't have "dots".
![image](https://user-images.githubusercontent.com/55940305/123295156-f7dd6800-d504-11eb-87d0-16e5e1f2bdbb.png)

These splitter have same color as main GUI and "dots" are restored.
![image](https://user-images.githubusercontent.com/55940305/123294353-4a6a5480-d504-11eb-9074-7cd897634e29.png)

btw, ignore that document map has border and different colors.

mere-human:
For me, 2nd looks better. 1st has the same color as the tab so it's harder to distinguish it.

donho:
> For me, 2nd looks better. 1st has the same color as the tab so it's harder to distinguish it.

+1

PR #10199, opened 2021-07-20
Restore Splitter appearance in dark mode

ozone10:
and allow dynamic color change for arrows.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10069

Only changes for non-docking splitter.

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -1,4 +1,4 @@
-﻿#include "nppDarkMode.h"
+﻿#include "NppDarkMode.h"
 
 #include "DarkMode/DarkMode.h"
 #include "DarkMode/UAHMenuBar.h"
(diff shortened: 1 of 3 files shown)


PowerEditor/src/NppDarkMode.cpp
```diff
@@ -1,4 +1,4 @@
-﻿#include "nppDarkMode.h"
+﻿#include "NppDarkMode.h"
```
ozone10, line 1: Finally found out why vs2019 has trouble with finding function definition.

donho: So file name is case sensitive now ?

ozone10: @donho
> So file name is case sensitive now ?

It depends on compiler implementation, the most used will usually compile fine, some with warning.

This is more about vs2019 as IDE, as sometimes it can have trouble with connecting the right header to source file,
so I can not use `ctrl` + clicking on function name to go to function declaration or definition.

PR #10207, opened 2021-07-20
Update french.xml

fylip22:
Update french.xml

### DIFF
--- a/PowerEditor/installer/nativeLang/french.xml
+++ b/PowerEditor/installer/nativeLang/french.xml
@@ -23,7 +23,7 @@
                     <Item subMenuId="file-openFolder" name="Ouvrir le répertoire du fichier en cours"/>
                     <Item subMenuId="file-closeMore" name="Fermer +"/>
                     <Item subMenuId="file-recentFiles" name="Fichiers récents"/>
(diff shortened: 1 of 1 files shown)


SubActif:
I am not used to working on a github and as I do not want to do anything here is my French translation version.
I have been translating on my side for a while and I adapt according to the available space and at best the texts.
I use a order by IDs number (not the original order) to find my way around what changes the place of certain items in relation to the online file.

I added elements not present in the current file (surely forgotten) and removed duplicate IDs that had no logic unless I had different IDs.
On the other hand, it is also missing in English elements present in the View menu which are not suddenly translated because they do not know their IDs:
- Hide Line Number Margin
- Hide Bookmark Margin
- Hide Folder Margin

This can disturb I think the fact that I align the names of the XML elements in column, it is for readability but on Github it may not be practical, I do not know how GitHub will handle these empty spaces and if they will be the same size as the original.
For some variables I have also removed the manual line breaks to replace them with a &#x0A; also for ease of translation but hey I suspect that this will disturb some.

[french.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/6852329/french.txt)

Issue #10205, opened 2021-07-20
[DarkMode] Edge/border around main window/edit zone


ozone10:
Currently in Npp v8.1.2 `No edge` has no effect in dark mode.
![image](https://user-images.githubusercontent.com/55940305/126368419-ccd30482-747c-47f2-afdf-a18e86216654.png)

What about using `WS_BORDER` style instead of `WS_EX_CLIENTEDGE` in dark mode.
![image](https://user-images.githubusercontent.com/55940305/126368739-730189ce-ce9a-40f9-81a3-42ddd87494d0.png)

mere-human:
Shall we create a special tag for dark mode issues?

ozone10:
I think there is already Dark mode label.

PR #10211, opened 2021-07-20
Restore edit zone border in dark mode

ozone10:
Restore functionality for option `No edge` in preference when using dark mode.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10205

I did not run tabify on "Notepad_plus.cpp" and "ScintillaEditView.cpp", because both of them contain nearly same amount of tab and space lines.


donho:
@ozone10 
Please revert all the formatting and keep only the modification- I prefer to keep the complete history when I use blame.

donho (changes requested):
Remove all the reformatting for using "git blame" afterward.

donho (approved):

Issue #10027, opened 2021-06-18
Dark Mode: Find Dialog, Visual Glitch


MarkusBodensee:
### Description of the Issue
There is a visual glitch at the `Find what` combo box when opening the find dialog in dark mode. Does not happen in normal mode.

### Steps to Reproduce the Issue
1. Enable Dark Mode and restart Nodepad++ as recommended.
2. Open find dialog with `Ctrl + F`.
3. See visual glitch at the `Find what` combo box.

- Hovering the combo box or click on any other control --> visual glitch disappears.
- Click to edit view, which makes find dialog inactive --> visual glitch disappears.

### Expected Behavior
No visual glitch.

### Actual Behavior
![SearchReplaceV8 1](https://user-images.githubusercontent.com/82021174/122533510-f18d4e80-d021-11eb-9c6e-f768abdcd5e7.png)

### Debug Information
Notepad++ v8.1   (64-bit)
Build time : Jun 17 2021 - 01:56:07
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.1556
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll

rimrul:
> Hovering the combo box or click on any other control --> visual glitch disappears.
    Click to edit view, which makes find dialog inactive --> visual glitch disappears.

Resizing the dialog window makes it re-appear in v8.1.1

PR #10220, opened 2021-07-21
Dark find and replace dialog enhancement and fixes

ozone10:
Make color more consistent.
Fix bug with trackbar (Transparency option) not changing background,
when changing background color.
Fix visual glitch.
Enhancement to combobox, including DPI fixes.
Make listbox of combobox use dark scrollbar.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10027
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10221


PowerEditor/src/NppDarkMode.cpp
```diff
@@ -983,7 +983,7 @@ namespace NppDarkMode
 		GetThemeBackgroundContentRect(buttonData.hTheme, hdc, BP_GROUPBOX, iStateID, &rcBackground, &rcContent);
 		ExcludeClipRect(hdc, rcContent.left, rcContent.top, rcContent.right, rcContent.bottom);
 
-		DrawThemeParentBackground(hwnd, hdc, &rcClient);
+		//DrawThemeParentBackground(hwnd, hdc, &rcClient);
```
ozone10, line 986: This is causing visual glitch, currently only commented, in case there is some side effect.
But I have not noticed anything for now.

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -93,7 +93,7 @@ namespace NppDarkMode
 		HEXRGB(0xE0E0E0),	// textColor
 		HEXRGB(0xC0C0C0),	// darkerTextColor
 		HEXRGB(0x808080),	// disabledTextColor
-		HEXRGB(0x808080)	// edgeColor
+		HEXRGB(0x646464)	// edgeColor
```
ozone10, line 96: use same color as system border

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -184,7 +184,7 @@ namespace NppDarkMode
...
```
ozone10, line 187: use same color as system border

donho (changes requested):
Remove all the reformatting for using "git blame" afterward.

ozone10:
@donho 
> Remove all the reformatting for using "git blame" afterward.

you mean trimmed lines?

donho:
@ozone10 
Yes. Not only for the `git blame`, but also it's hard to review the code - I know I can hide the space from GitHub interface, but there are the other contributors with whom I apply "No formatting" rule.

ozone10:
Ok will fix it now.

ozone10:
btw I don't know if github support blame with ignoring whitespace.
but `git blame -w` command should allow to ignore whitespace changes

donho:
Noted. 
I use usually git blame in VS or on Github. So let's keep the current white spaces.

donho (approved):

PR #10225, opened 2021-07-22
Update portuguese.xml

hugok79:
For commit:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/8e38b9dabab11da0b7f3972a99e2d15a0a6cdeb6


donho:
Here is the diff between v8.1.2 & v8.1.3:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10297

Could you check it and confirm me it's all done (or you prefer do it later in the next release)?
Thank you.

hugok79:
Replaced with "PR 10298"
@donho

Issue #10226, opened 2021-07-22
[DarkMode] Dark Style Configurator


ozone10:
Notepad++ v8.1.2 doesn't apply dark mode on Style Configurator.
![image](https://user-images.githubusercontent.com/55940305/126678101-9415849e-bb0e-410c-81c8-8e7c6977398d.png)

Part of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10171

PR #10227, opened 2021-07-22
Apply dark mode on Style Configurator

ozone10:
Also create link text color to be used too in about, debug and preference dialogs.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10226

image


ozone10:
@donho 
I did not create color option in preference because currently it only affect style configurator (yellow title text).

btw both `NppDarkMode.cpp` and `NppDarkMode.h` don't have license comment header.

PR #10228, opened 2021-07-23
update korean.xml 8.1.2

yunbj:


donho:
Here is the diff between v8.1.2 & v8.1.3:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10297

Could you check it and confirm me it's all done (or you prefer do it later in the next release)?
Thank you.

yunbj:
@donho
It's ready. Thanks for your comment.

Issue #10230, opened 2021-07-23
[DarkMode] Dark Windows dialog


ozone10:
Notepad++ v8.1.2 does not apply dark mode on Windows dialog.

![image](https://user-images.githubusercontent.com/55940305/126815008-2e61dfcb-3ae6-4dbb-9354-23b502c763ef.png)

related https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10171

PR #10231, opened 2021-07-23
Apply dark mode on Windows dialog

ozone10:
Use background and text colors from style theme in listview.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10230

Currently does not support dynamic redraw, dialog needs to be closed and reopened to apply new settings.

image
image
image


donho:
> Currently does not support dynamic redraw, dialog needs to be closed and reopened to apply new settings.

I think it's normal, since it's a modal dialog - and there's very little chance that users switch dark mode on/off while Window dialog is opened.

Issue #10236, opened 2021-07-24
[DarkMode] Dark Column Editor


ozone10:
Notepad++ v8.1.2 does not apply dark mode on Column Editor.
![image](https://user-images.githubusercontent.com/55940305/126873275-a6e259d7-1a8c-4de9-b254-0898198fb8a8.png)

PR #10237, opened 2021-07-24
Apply dark mode on column editor

ozone10:
image

<s>Only small issue is that "Format" text is not in center. </s>
Ok I may have solution, still need testing, will do another PR later.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10236

### DIFF
--- a/PowerEditor/src/ScintillaComponent/columnEditor.cpp
+++ b/PowerEditor/src/ScintillaComponent/columnEditor.cpp
@@ -44,6 +44,8 @@ INT_PTR CALLBACK ColumnEditorDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 	{
 		case WM_INITDIALOG :
 		{
+			NppDarkMode::autoSubclassAndThemeChildControls(_hSelf);
+
 			switchTo(activeText);
 			::SendDlgItemMessage(_hSelf, IDC_COL_DEC_RADIO, BM_SETCHECK, TRUE, 0);
 			goToCenter();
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScintillaComponent/columnEditor.cpp
```diff
@@ -299,7 +348,6 @@ void ColumnEditorDlg::switchTo(bool toText)
 	::EnableWindow(::GetDlgItem(_hSelf, IDC_COL_INCREASENUM_EDIT), !toText);
 	::EnableWindow(::GetDlgItem(_hSelf, IDC_COL_REPEATNUM_STATIC), !toText);
 	::EnableWindow(::GetDlgItem(_hSelf, IDC_COL_REPEATNUM_EDIT), !toText);
-	::EnableWindow(::GetDlgItem(_hSelf, IDC_COL_FORMAT_GRP_STATIC), !toText);
```
ozone10, line 302: Groupbox has no state so EnableWindow should not be used. When used, it can caused some issue with text color.

ozone10:
Sorry for this commit mess, forgot to save file.

PowerEditor/src/ScintillaComponent/columnEditor.cpp
donho: Even in light mode?

ozone10: In light mode it does nothing, but in dark mode it will reset back to black color.

donho: It's OK for this PR, because the enabled zone and the disabled zone are quite distinguishable. On the others PR we may make it be able to be disabled under only the light mode.

ozone10: @donho
I think only "Custom Backup Directory" in preference in backup also has this issue.

PR #10241, opened 2021-07-25
Fix file name edit box not functional on Windows Vista

### LINKED ISSUE #10238: Can't change file name in Open/Save dialog on Windows Vista
### Description of the Issue

image

1. File > New > Save as
2. Type letters to _File name_ edit box
3. Press Save

### Expected Behavior

Text in edit box changes. File is saved with a new name.

### Actual Behavior

Text doesn't change. File is saved with old (default) name.
The same happens with open dialog.

### Debug Information

Notepad++ v8.1.2   (32-bit)
Path : C:\Users\Public\npp.8.1.2.portable\notepad++-orig.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows (TM) Vista Ultimate (64-bit) 
OS Build : 6002.0
Current ANSI codepage : 1252

mere-human:
Avoid the same window procedure being used for several different edit boxes.
The problem was found on Windows Vista but _it can happen in other environment too_.

Fix #10238

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -465,14 +465,14 @@ private:
 		const int bufferLen = MAX_PATH;
 		static TCHAR buffer[bufferLen];
 
-		if (GetClassName(hwnd, buffer, bufferLen) != 0)
+		if (IsWindowEnabled(hwnd) && GetClassName(hwnd, buffer, bufferLen) != 0)
 		{
 			if (lstrcmpi(buffer, _T("ComboBox")) == 0)
 			{
 				// The edit box of interest is a child of the combo box and has empty window text.
-				// Note that file type dropdown is a combo box also (but without an edit box).
+				// Note that there might be other combo boxes (file type dropdown, address bar, etc).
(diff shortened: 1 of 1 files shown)


mere-human:
Checked it on Windows 10, 7, Vista. File save and file open seem to work fine. Other features like Unix path are good too.

donho:
@mere-human 
> The problem was found on Windows Vista but it can happen in other environment too.

So what was the problem? Is it not only related to Vista? What kind of other environment?

mere-human:
> @mere-human
> 
> > The problem was found on Windows Vista but it can happen in other environment too.
> 
> So what was the problem? Is it not only related to Vista? What kind of other environment?

@donho The problem is that there are more than one ComboBox + Edit controls in save dialog:
image

We need to find the one responsible for _File name_ to override its window procedure.

For some reason, on Windows 10 and Windows 7 only one Edit control is present while _EnumChildProc_ is called.
But on Windows Vista, 2 Edit controls appear in _EnumChildProc_. Maybe, 2nd one is disabled later.
This was the cause of the problem because we were trying to use the same window procedure for 2 Edit controls which made _File name_ unusable.

While I'm not sure what leads to such problem, I'm sure that with this PR the code is safer and more resistant to similar problems.

alphacentauris44:
I used the build for PR # 10174 and I confirm that I no longer reproduce the problem encountered with, under WIndows 10 x64.

Issue #10242, opened 2021-07-25
[DarkMode] Color inconsistency


ozone10:
Notepad++ v8.1.2 has some color inconsistency in dark mode.
Some items should have different colors, e.g. panel toolbars.

PR #10243, opened 2021-07-25
Make dark colors consistent

ozone10:
Make dark colors consistent for panel toolbars.
Make dark colors consistent style configurator edit controls.
Make dark colors consistent for docking elements.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10242

-  `BackgroundColor()`, `BackgroundBrush()` – should be used for items which have some interactions, or need to be different from Darker variant.
-  `DarkerBackgroundColor()`,  `DarkerBackgroundBrush()` – should be used for static/top backgrounds.
-  `SofterBackgroundColor()`,  `SofterBackgroundBrush()` – should be used for items , which have interactions, or have focus.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -181,7 +181,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			{
 				RECT rc = { 0 };
 				GetClientRect(hwnd, &rc);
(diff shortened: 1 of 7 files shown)


donho:
@ozone10 
I don't see the point - the current dark colors are quite consist to me.
Could you provide the screenshots "before & after" please?

ozone10:
This is mainly for custom tone, as default tones have same colors for top and main colors.

before
image

after
image

Also to avoid unnecessary color flicker, when moving splitter, because the erase color is different.

Issue #10253, opened 2021-07-25
[DarkMode] Dark Value dialogs


ozone10:
Notepad++ v8.1.2 does not apply dark mode on "mini" dialogs created in preference

![image](https://user-images.githubusercontent.com/55940305/126905240-dbbaffce-9254-43ea-84f2-84c86e9d2f66.png)

Issue #10210, opened 2021-07-20
Distraction-free mode's exit buttton not affected by dark mode


mrandreastoth:
See screenshot below. Note how the distraction-free mode's exit button is not taking on the dark mode theme when it should.

![image](https://user-images.githubusercontent.com/26897855/126392239-cab5f4b1-4f3b-406f-aea8-7b10f3048132.png)

ArkadiuszMichalski:
The same is for `Full Screen Mode (F11)` and `Post-It (F12)`.

Issue #10171, opened 2021-07-15
[DarkMode} Make other dialog windows dark (WIP)


ozone10:
Apply dark mode on other dialog windows. Currently WIP.
I need to fix changing between light/dark mode. 

Missing plugin admin, user defined language, windows...  (from window menu) and shorcut mapper.
These are more complex.

![image](https://user-images.githubusercontent.com/55940305/125859973-80c9581a-647e-494a-adeb-35681588b3c5.png)

donho:
toolbar truncated chevron not visible in dark mode:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10103

kczx3:
> > Tabbar overflow (updown control)
> 
> ![image](https://user-images.githubusercontent.com/58130821/130950057-8dea3513-2c66-4012-ade9-b5bf3cf2daa1.png)
> 
> This one is super bugging me.

It seems you can call GetTopWindow passing the HWND of the Tab control to get the HWND of the UpDown since it is a child of the Tab control.  Or maybe you could monitor [WM_PARENTNOTIFY](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/inputmsg/wm-parentnotify) to see when the child UpDown is created/destroyed.  Even so, I'm not sure how you'd subclass the UpDown to paint it yourself.

kczx3:
WM_PARENTNOTIFY doesn't seem to send for the creation of the UpDown control even though I don't see it having the WS_EX_NOPARENTNOTIFY style.

alicepachecoo:
Hi everyone. 
I'm using the latest version (8.9.5) and I want to report that the save popup (not the same as the SaveAll) doesn't respect the Dark Mode style (from the application or windows 11), it's the only element of the UI that has this mismatch. Evidence below:

// Video
[https://drive.google.com/file/d/1vYqni7jh-1Ekkjgipiff0E9fJGuK1c8e/view?usp=sharing](url)
// Picture
[https://drive.google.com/file/d/1kj0LiZvgw2EUbmEMZYeWGQGdMHfEnxB9/view?usp=sharing](url)
// Windows 11 - dark mode style
[https://drive.google.com/file/d/1L9d3QC7jKVfos900-1Iw1Cbhum2WvjcL/view?usp=sharing](url)

ozone10:
@alicepachecoo 
[...]

PR #10254, opened 2021-07-25
Apply dark mode on value dialogs

ozone10:
Apply dark mode on value dialogs (dialogs created in preference).
Change value dialogs border style.
Apply dark mode on close button for other views (full-screen, post-it
and distraction free).

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10253, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10210

image
image

"Mini" dialogs don't support dynamic redraw, does not matter, as it is not possible to change dark mode when it's opened.


GOLEM777:
How about?
capture_07252021_220553

ozone10:
@GOLEM777 
> How about?

This is part of tabbar, unfortunately I still have no idea how to make it dark.
you can check https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10171 for current state of dark mode elements

Issue #10255, opened 2021-07-25
[DarkMode] Dark hash dialogs


ozone10:
Notepad++ v8.1.2 does not apply dark mode on hash dialogs.
![image](https://user-images.githubusercontent.com/55940305/126910990-d4c5bc4c-c73d-458c-928a-ac8a9a00811f.png)

PR #10256, opened 2021-07-25
Apply dark mode on hash dialogs

ozone10:
Also allow typing newline (`enter`) to use multiple lines for generating hash.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10255

image

I will be now taking break, I will do other dialogs later.


PowerEditor/src/MISC/md5/md5Dlgs.cpp
```diff
@@ -22,13 +22,16 @@
 #include "CustomFileDialog.h"
 #include "Parameters.h"
 #include <shlwapi.h>
+#include "resource.h"
 
 INT_PTR CALLBACK HashFromFilesDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM /*lParam*/)
```
donho, line 27: `LPARAM /*lParam*/)` should be uncommented.

donho (changes requested):
Solve the build error

PowerEditor/src/MISC/md5/md5Dlgs.cpp
ozone10: fixed

Issue #10264, opened 2021-07-26
[DarkMode] Dark Preference, About, Debug and Save dialogs


ozone10:
Notepad++ v8.1.2 does not apply dark mode on Preference, About, Debug and Save dialogs.

![image](https://user-images.githubusercontent.com/55940305/127050709-83521c71-47e8-43fd-9ed9-77810b358640.png)

PR #10265, opened 2021-07-26
Apply dark mode on preference and other dialogs

ozone10:
Apply dark mode on about, debug and save dialogs.
Make link text aware of dark mode.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10264

image

Preference, about and debug dialogs depend on `URLCtrl.cpp` so they need to be done together.

### DIFF
--- a/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
+++ b/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
@@ -211,6 +211,42 @@ INT_PTR CALLBACK PreferenceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM 
 			if (enableDlgTheme)
 				enableDlgTheme(_hSelf, ETDT_ENABLETAB);
 
+			NppDarkMode::autoSubclassAndThemeChildControls(_hSelf);
+
+			return TRUE;
+		}
+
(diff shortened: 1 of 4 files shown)


PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -3189,7 +3560,7 @@ void BackupSubDlg::updateBackupGUI()
 		isEnableGlobableCheck = true;
 		isEnableLocalCheck = BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_BACKUPDIR_CHECK, BM_GETCHECK, 0, 0);
 	}
-	::EnableWindow(::GetDlgItem(_hSelf, IDC_BACKUPDIR_USERCUSTOMDIR_GRPSTATIC), isEnableGlobableCheck);
+	//::EnableWindow(::GetDlgItem(_hSelf, IDC_BACKUPDIR_USERCUSTOMDIR_GRPSTATIC), isEnableGlobableCheck);
```
ozone10, line 3563: @donho
again the same issue with groupbox, in light mode it does nothing, in dark mode reset color to black

if you need some changes, could it be done in separate PR, I won't available for a while, and I don't want block this PR

Issue #10269, opened 2021-07-27
[DarkMode] Dark Color Popup


ozone10:
Notepad++ v8.1.2 does not apply dark mode on Color Popup.

![image](https://user-images.githubusercontent.com/55940305/127198533-6770bf92-1093-4784-aa16-34315a781d1e.png)

PR #10270, opened 2021-07-27
Apply dark mode on color popup

ozone10:
Fix use of dark mode edge color in light mode for color picker button.
Properly redraw disabled color picker button.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10269

light mode new style (border)
image
dark mode
image


mere-human:
Why is the dialog look different in new light mode? It is gray in current build:
image

ozone10:
@mere-human 
> 
> Why is the dialog look different in new light mode? It is gray in current build:

I can look into it, but it is issue? Should I restore the light mode look?

I have changed border style because in dark mode it create white border.

donho:
@mere-human 
My BEFORE & AFTER screenshot under windows 10:

image

Do you think it's an issue?

mere-human:
> Do you think it's an issue?

No, I just wanted to point out it's different.

Issue #10196, opened 2021-07-19
Changing settings in Dark Mode in Preferences should not reset Style Configurator


Edditoria:
### Description of the Issue

In v8.1.2, changing settings in Dark Mode in Preferences will reset Style Configurator to "DarkModeDefault". (Should it be "Dark Mode Default?)
They should not interact in this way.

### Steps to Reproduce the Issue

1. Open Preferences via Settings.
2. Click "Dark Mode".
3. Click something including:
    - "Enable dark mode"
    - Select any tone
    - Disable "Enable dark mode"
4. Change "Style Configurator" again.
5. Loop step 3 and 4.

### Expected Behavior

User can select their favourite tone (app-theme) and styler (syntax-theme) separately.
These two type of configs are in different dialog, anyway.

### Actual Behavior

When changing settings related to dark mode, the styler switches to "DarkModeDefault".
When disable "Enable dark mode", the styler switches to "Default (styler.xml)".

ozone10:
related (maybe duplicate) to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10049

donho:
@Edditoria 
I'm looking for a way to switch to `markdown.zenburn.udl.xml` :
https://github.com/Edditoria/markdown-plus-plus/blob/master/udl/markdown.zenburn.udl.xml

if dark mode is enabled.

Edditoria:
Let me evalue how dark mode should work:

- There are 3 windows related to color appearance:
  1. Preferences - Dark Mode.
  2. Style Configurator.
  3. UDL.
- In contrast of syntax themes, users may often switch between dark mode and light mode (when dark mode is disable).
- You may want to create tones for light mode in future.
- Users have their favourite themes in light and dark, and the themes are **most possibly from different authors**.
- So, it is not good to integrate different UDLs in one file.
- User may want to use a dark theme in light mode.
- Pre-defined dark/light in UDL does not work either.
- Manually change configs in UDL may be too complicated for users, especially when they update their UDLs in bulk.

To make it simple and easy to understand for user:

- There will be 2 main categories: dark mode and light mode.
- There will be user-selectable syntax themes: 1 for dark (apply to all tones), and 1 for light.
- Syntax theme:
  - ~~In Preferences - Dark Mode, users can select 1 syntax theme for dark and 1 for light.~~ (Sorry it is Microsoft-way)
[...]

Issue #10049, opened 2021-06-21
[Feature Request] Don't force style and tab icons when toogling dark mode


ozone10:
### Description of the Issue

Don't force style and tab icons when toggling dark mode.
Make it optional, or allow to choose default styles and icons.

I am using my own dark and light theme, and it can be annoying when toggling between modes.
For icon currently many plugins don't support new darkmode icons, so they appear blank, when using fluent icons.
So I currently prefer standard icon set.
### Steps to Reproduce the Issue

1. Open preferences
2. Enable or disable dark mode

### Expected Behavior

Style and icons remain same.

### Actual Behavior

Icon is changed to fluent and style to Default or DarkModeDefault.

donho:
@ozone10 
In order to provide a better UX for ppl who switch to dark mode for the first time, it's inevitable to show users the "suitable" configuration. For me, if users have tried dark mode, they should decide which side they wanna stay. Maybe it could not the case for the developers of this project (ie. you and others who are working in dark mode), but for the most of users, you are in dark mode, or you are not.

ozone10:
@donho 
What about adding checkbox `Use default style theme`,  which will be enabled by default.
Same with toolbar icons.

donho:
@ozone10 
If we add more options for 1 option, then there's a problem.
Too many options, no option.

PR #10271, opened 2021-07-27
Don't reset style & icons, when changing dark tone

ozone10:
Improve title bar dynamic redraw when changing mode.

This will only reset style and toolbar icons when toggling between dark and light mode.

Related https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10049 , https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10196


ozone10:
Forgot parameter

Issue #10287, opened 2021-07-28
[DarkMode] Dark Plugins Admin


ozone10:
Notepad++ v8.1.2 does not apply dark mode on Plugins Admin.

![image](https://user-images.githubusercontent.com/55940305/127382584-f11417de-709e-4af2-ae3d-80f048e4d413.png)

PR #10288, opened 2021-07-28
Apply dark mode on plugins admin

ozone10:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10287

image

Currently does not support dynamic change for  listview colors from style configurator, Notepad++ needs restart. 

@donho 
This may be last PR from me for while, I will be away.


donho:
@ozone10 
> This may be last PR from me for while, I will be away.

OK, no problem.
Thank you a lot for pushing Notepad++ darker. I'll prepare a release as soon as possible.

Issue #10299, opened 2021-07-31
[DarkMode] Windows 11 dialog background issue


ozone10:
On Windows 11 `WM_CTLCOLORDLG` and theme seem to work differently than on older Windows, 

![image](https://user-images.githubusercontent.com/55940305/127749055-e441b637-4d69-4593-be31-1880f6c6b1bb.png)

Edit:
because of this
https://docs.microsoft.com/en-us/windows/apps/get-started/make-apps-great-for-windows#desktop-apps-win32

lilalien:
Yeah I still get this on the latest version in the dialogs, and on the about box. Looking forward to the fix, its coming along nicely in dark mode!

PR #10300, opened 2021-07-31
Fix issues with dialog background on Windows 11

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10299

### DIFF
--- a/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
+++ b/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
@@ -137,6 +137,18 @@ INT_PTR CALLBACK RegExtDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lPar
 			break;
 		}
 
+		case WM_ERASEBKGND:
+		{
+			if (NppDarkMode::isEnabled())
+			{
+				RECT rc = { 0 };
+				getClientRect(rc);
+				::FillRect(reinterpret_cast<HDC>(wParam), &rc, NppDarkMode::getDarkerBackgroundBrush());
+				return TRUE;
+			}
+			break;
+		}
+
 		case WM_DRAWITEM :
 		{
 			HICON hIcon = ::LoadIcon(_hInst, MAKEINTRESOURCE(IDI_DELETE_ICON));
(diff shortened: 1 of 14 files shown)


donho:
Windows 11 is not out yet officially, so it could be a bug in Windows 11 (MS always does care about the retro-compatibility).
I would say let's wait for its coming to see if it needs to be fix.

ozone10:
Hopefully `WM_CTLCOLORDLG` won't end like `WM_CTLCOLORMSGBOX`.

`WM_ERASEBKGND` should be also more reliable on older Windows.
Windows 11 is already on insider beta channel, but as you said it could be bug, so maybe wait for at least preview channel.

ozone10:
@donho 
Notepad++ is using uxtheme textures for "white" background, for example here
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2576bf884b37a916cf5f597d87e436fb59864cec/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L212

and it seems that Windows 11 will handle uxtheme api differently
```
Win32 controls that use UXTheme will automatically get a "repaint" of the Light theme for select rejuvenated control visuals.
```
https://docs.microsoft.com/en-us/windows/apps/get-started/make-apps-great-for-windows#desktop-apps-win32

donho:
@ozone10 
> and it seems that Windows 11 will handle uxtheme api differently

So no need to retrieve and call `enableDlgTheme` under Windows 11 anymore?

ozone10:
@donho 
It's needed to make background white in light mode, without it will be "classic gray", which does not look good.

ozone10:
@donho 
I've tried workaround with `EnableThemeDialogTexture` and it works great, I will do separate PR

donho:
@ozone10 So could we close this PR since the better way has be done in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/8ac13d90654e5ebaff2167582c7b24751d38b6ce ?

ozone10:
@donho 
Sure, this PR is not needed for now.
In case it is needed I can always reopen or do another PR.

btw Microsoft has just released Windows 11 ISO for insider

PR #10303, opened 2021-08-01
Fix crash when several instances of the file dialog are shown

### LINKED ISSUE #10290: Bug in  CustomFileDialog, when trying to close multiple of them, result in nullptr and crash
### Description of the Issue
https://community.notepad-plus-plus.org/topic/21575/bug-when-double-mouse-click-run-button-in-run-a-macro-multiple-times-dialog

1. Open Npp, create a new empty tab, write something in it to make it dirty.
2. In "Run a Macro Multiple Times" dialog, choose `Trailing space and save`, click the "run" btn multiple times.
3. Dismiss the Save-as dialog**s**.

### Expected Behavior
Save-as dialogs are closed one by one

### Actual Behavior
Npp crashed.

### Debug Information
Notepad++ v8.1.2   (64-bit)
Path : D:\Code\FigureOut\Textrument\秘籍\npp.8.0.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1110
Current ANSI codepage : 65001
Plugins : MarkdownText.dll mimeTools.dll MultiClipboard.dll NppConverter.dll NppExport.dll 

### Additional Information

[video by LNT](https://drive.google.com/file/d/1R4pC3m35EgDuymZRcYIp7_s67g8YBL33/view?usp=sharing)

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fcea83ba988e9f319549d4bdfa64facfe8a7ab0b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp#L558

```
Exception thrown: read access violation.
FileDialogEventHandler::_staticThis was nullptr.

```

mere-human:
Pass a pointer to FileDialogEventHandler instance via GWLP_USERDATA
instead of using static variable. This way each window can have its
own pointer.

Fix #10290

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -180,6 +180,15 @@ namespace // anonymous
 		return {};
 	}
 
+	LRESULT callWindowClassProc(const TCHAR* className, HWND hwnd, UINT msg, WPARAM wparam, LPARAM lparam)
+	{
+		WNDCLASSEX wndclass = {};
+		wndclass.cbSize = sizeof(wndclass);
+		if (GetClassInfoEx(nullptr, className, &wndclass) && wndclass.lpfnWndProc)
+			return CallWindowProc(wndclass.lpfnWndProc, hwnd, msg, wparam, lparam);
+		return FALSE;
+	}
+
 	// Backs up the current directory in constructor and restores it in destructor.
 	// This is needed in case dialog changes the current directory.
 	class CurrentDirBackup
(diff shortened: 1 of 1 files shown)


donho:
@mere-human 
The behaviour has been changed with this PR: multiple Save as dialogs are launched while I clicked "Run" button of Macro dialog.
However, when I click on "Run" +10 times very rapidly, the crash still happens.

We don't have to align the behaviour with the standard save dialog, The save dialog could have unique instance then ignore the other  launch command. It might be a better behaviour.

mere-human:
@donho 
Indeed, I tried clicking it ~10 times fast.
It seems now it crashes in a different place, not in the file dialog.

Release build:
image

Debug build:
image

So, do you want only one instance of file dialog to be shown simultaneously?
It could fix the problem, but need to check.

donho:
> So, do you want only one instance of file dialog to be shown simultaneously?

Yes, it seems to me a good behaviour.

mere-human:
> > So, do you want only one instance of file dialog to be shown simultaneously?
> 
> Yes, it seems to me a good behaviour.

Done.
Now, when clicking 2nd time on the Run button in _Run a Macro Multiple Times_ dialog and the file dialog is active, then nothing happens (until the dialog is closed/confirmed).
Note also, when set the value to 10 times then click run, the file dialog will be shown 10 times one after another without any problems.

PR #10304, opened 2021-08-01
Apply explorer style on treeview

ozone10:
Enable hottracking for explorer style.
Allow dynamic change based on treeview background.


chcg:
@ozone10 See appveyor build for mingw:

```
../src/NppDarkMode.cpp: In lambda function:
../src/NppDarkMode.cpp:400:11: error: 'pow' was not declared in this scope
    return pow(((colorChannel + 0.055) / 1.055), 2.4);
           ^~~
../src/NppDarkMode.cpp:400:11: note: suggested alternative: 'putw'
    return pow(((colorChannel + 0.055) / 1.055), 2.4);
```

likely include for math.h is missing, see https://www.cplusplus.com/reference/cmath/pow/

PR #10314, opened 2021-08-02
Use CRLF line ending in Copy command from Windows dialog

### LINKED ISSUE #10311: Copy from Windows dialog uses wrong line ending
### Description of the Issue

This may lead to file names being pasted without a visual delimiter in Notepad.
It was reported here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10123#issuecomment-890594047

1. Create or open several files
2. Menu > Window > Windows
3.  Select all (Ctrl + A)
4. RMB > Copy names
5. View > Show Symbol > Show End of Line
6. Create a new document
7. Edit > Paste Special > Paste Binary Content

### Expected Behavior

File names are separated by CR LF

### Actual Behavior

File names are separated by LF

### Debug Information

Notepad++ v8.1.2   (64-bit)
Build time : Aug  2 2021 - 21:35:49
Path : C:\Notepad++.exe
Command Line : -multiInst -nosession
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1110
Current ANSI codepage : 1252
Plugins : none

mere-human:
Also, move endsWith() function to Common.h and reuse it in WindowsDlg.

Fix #10311

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1297,6 +1297,15 @@ void trim(generic_string& str)
 	else str.erase(str.begin(), str.end());
 }
 
+bool endsWith(const generic_string& s, const generic_string& suffix)
+{
+#if defined(_MSVC_LANG) && (_MSVC_LANG > 201402L)
+#error Replace this function with basic_string::ends_with
+#endif
+	size_t pos = s.find(suffix);
+	return pos != s.npos && ((s.length() - pos) == suffix.length());
+}
+
 int nbDigitsFromNbLines(size_t nbLines)
 {
 	int nbDigits = 0; // minimum number of digit should be 4
--- a/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
+++ b/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
@@ -984,6 +984,7 @@ void WindowsDlg::putItemsToClipboard(bool isFullPath)
 	constexpr int pathColumn = 1;
 
 	TCHAR str[MAX_PATH] = {};
+	const generic_string crlf = _T("\r\n");
 
 	generic_string selection;
 	for (int i = -1, j = 0; ; ++j)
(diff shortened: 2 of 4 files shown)


mere-human:
@donho This should make the feature more usable for Windows users.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1297,6 +1297,15 @@ void trim(generic_string& str)
 	else str.erase(str.begin(), str.end());
 }
 
+bool endsWith(const generic_string& s, const generic_string& suffix)
+{
+#if defined(_MSVC_LANG) && (_MSVC_LANG > 201402L)
+#error Replace this function with basic_string::ends_with
+#endif
```
donho, line 1304: Why this define?

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
```diff
@@ -984,6 +984,7 @@ void WindowsDlg::putItemsToClipboard(bool isFullPath)
 	constexpr int pathColumn = 1;
 
 	TCHAR str[MAX_PATH] = {};
+	const generic_string crlf = _T("\r\n");
```
donho, line 987: I would say it could be CR, LF or CRLF - it depends on the current document's EOL (Windows, Unix or Macintosh)

donho (changes requested):
Consider the given suggestion.

PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp
donho: After testing, the pasted EOL will conform to the destination's EOL.
So you're right, let's make it as Windows format.

donho (approved):

PowerEditor/src/MISC/Common/Common.cpp
mere-human: Because starting from C++20, this feature is available in standard library https://en.cppreference.com/w/cpp/string/basic_string/ends_with
When the project starts using C++20, this function can be removed.

Issue #10195, opened 2021-07-19
Mark all - add option to follow "find dialog settings"


cmeriaux:
### Description of the Issue

Version 8.0 has introduce a regression on mark all feature. It can't be configured through "find dialod settings"
It's so handy to configure that feature through find menu, instead opening Preferences/Highlighting/Mark all menu. That's why there was an option for smart highlighting, so that smar highlighting is configured through find menu
I suggest to a add a menu item to configure it: see bellow the screenshot

![image](https://user-images.githubusercontent.com/12024941/126213391-50a8f2a2-e620-4b37-a751-2d8524060a76.png)

### Steps to Reproduce the Issue

1. 
2. 
3. 

### Expected Behavior

I excpect the behavior of v7.9.5: mark all is configurable through find menu

### Actual Behavior

Mark all is only configurable through preferences.

### Debug Information

Notepad++ v8   (32-bit)
Build time : Jun  6 2021 - 18:41:52
Path : G:\dev\npp.8.0.portable\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2009
OS Build : 19043.1110
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

Same issue with V8.1.1 abd V8.1.2

mere-human:
@cmeriaux But isn't it the same?
![image](https://user-images.githubusercontent.com/9664141/126216294-3c030f09-80fb-4776-bff2-b48ab0034e19.png)

cmeriaux:
@mere-human this is only for smart highlighting

mere-human:
This problem is related to https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9796
Just for the reference.

ArkadiuszMichalski:
This name "Mark All" is used in two places and is confusing because represents two different mechanisms. Can't use some word other than "Mark" for one of them? In context menu we have "Style", which is more obvious.

![image](https://user-images.githubusercontent.com/2730894/128362295-aba3f0e7-07d6-407e-9939-d75cc8848224.png)

Again, I took some time to figure it out. Only for me is it confusing?

mere-human:
> This name "Mark All" is used in two places and is confusing because represents two different mechanisms.

[...]

PR #10351, opened 2021-08-07
Substitute "Mark" for "Style" in the menu entries

donho:
In order to avoid users' confusion, all "Mark" occurrences of menu entries are substituted for "Style", so users won't confuse with "Mark" fonction in Find dialog.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10195#issuecomment-893481499


PowerEditor/installer/nativeLang/english.xml
```diff
@@ -547,41 +547,41 @@ The comments are here for explanation, it's not necessary to translate them.
...
+                    <Item id="43036" name="Previous style of 4th style"/>
+                    <Item id="43037" name="Previous style of 5th style"/>
+                    <Item id="43038" name="Previous style of Mark (Find dialog)"/>
+                    <Item id="43039" name="Next style of 1st style"/>
+                    <Item id="43040" name="Next style of 2nd style"/>
+                    <Item id="43041" name="Next style of 3rd style"/>
+                    <Item id="43042" name="Next style of 4th style"/>
+                    <Item id="43043" name="Next style of 5th style"/>
+                    <Item id="43044" name="Next style of Mark (Find dialog)"/>
+                    <Item id="43055" name="Copy Styled Text of 1st Style"/>
```
ArkadiuszMichalski, line 578: In all copy this "Style" can be given with lowercase, in accordance with all other cases.

PowerEditor/src/Parameters.cpp
```diff
@@ -189,41 +189,41 @@ static const WinMenuKeyDefinition winKeyDefs[] =
...
+	{ VK_5,       IDM_SEARCH_GOPREVMARKER5,                     true,  false, true,  TEXT("Previous style of 5th style") },
+	{ VK_0,       IDM_SEARCH_GOPREVMARKER_DEF,                  true,  false, true,  TEXT("Previous style of Mark (Find dialog)") },
+	{ VK_1,       IDM_SEARCH_GONEXTMARKER1,                     true,  false, false, TEXT("Next style of 1st style") },
+	{ VK_2,       IDM_SEARCH_GONEXTMARKER2,                     true,  false, false, TEXT("Next style of 2nd style") },
+	{ VK_3,       IDM_SEARCH_GONEXTMARKER3,                     true,  false, false, TEXT("Next style of 3rd style") },
+	{ VK_4,       IDM_SEARCH_GONEXTMARKER4,                     true,  false, false, TEXT("Next style of 4th style") },
+	{ VK_5,       IDM_SEARCH_GONEXTMARKER5,                     true,  false, false, TEXT("Next style of 5th style") },
+	{ VK_0,       IDM_SEARCH_GONEXTMARKER_DEF,                  true,  false, false, TEXT("Next style of Mark (Find dialog)") },
+	{ VK_NULL,    IDM_SEARCH_STYLE1TOCLIP,                      false, false, false, TEXT("Copy Styled Text of 1st Style") },
```
ArkadiuszMichalski, line 220: In all copy this "Style" can be given with lowercase, in accordance with all other cases.

ArkadiuszMichalski (approved):
We have also settings in the Style Configurator window for this two mechanism. They seem obvious here (due 1,2,3,4,5 variants), but we also have a mix of "mark" and "style" word.

image

PowerEditor/src/Parameters.cpp
donho: Nice spotted.
All the menu items are in Proper Case, so I will modify them to conform to the convention.

donho:
@ArkadiuszMichalski 
> We have also settings in the Style Configurator window for this two mechanism. They seem obvious here (due 1,2,3,4,5 variants), but we also have a mix of "mark" and "style" word.

What's your suggestion then?

ArkadiuszMichalski:
I don't have anything special, just the following came to mind:
```
Style 1 for token
Style 2 for token
Style 3 for token
Style 4 for token
Style 5 for token

or

1st style for token
2nd style for token
3rd style for token
4th style for token
5th style for token
```

donho:
hmm...

```
Mark (Find dialog)
1st token style
2nd token style
3rd token style
4th token style
5th token style
```

It's shorter to display and user won't lose the reference.
What do you think?

ArkadiuszMichalski:
It looks good and it's clear.

donho:
> one more thing, in Search > Jump Up/Down we have `Find Style` but in Copy Styled Text we have `Find Style (Marked)`. Shouldn't they have the same names? If you use "Mark (Find dialog)" everywhere would not be better (as above)?

In this case `Find Mark Style` everywhere to reference to Mark op in Find dialog? Do you see any incoherence in this way?

ArkadiuszMichalski:
"Find Mark Style" in "Search > Jump Up/Down" and "Search > Copy Styled Text" is fine.

PowerEditor/src/Parameters.cpp
donho: @ArkadiuszMichalski 
After checking, these commands are shown in shortcut mapper for explaining what it is instead of displaying `1st Style`, so they are not the same as menu items, so I would keep the "free style".

donho:
@ArkadiuszMichalski 
Just push a new modification. Hope it's OK this time.

ArkadiuszMichalski:
Yes, it's more consistent now.

donho:
@ArkadiuszMichalski 
Thank you for the checking!

PR #10359, opened 2021-08-09
No toolbar tooltip - fixed

### LINKED ISSUE #10355: Issue: No Toolbar Tooltips
After commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/733c7e73de6d5dece3bcd9eb8784fd2bf61da359 main toolbar tooltips are not showing

Ashfaaq18:
Fix #10355

@Yaron10 and @ozone10 , can you guys give this a thorough check. RTL and LTR tooltips should be working accordingly. 
Check the drag and drop as well.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3795,10 +3795,8 @@ void Notepad_plus::dropFiles(HDROP hdrop)
 	{
 		// Determinate in which view the file(s) is (are) dropped
 		POINT p;
-		::GetCursorPos(&p);
-		HWND hWnd = WindowFromPoint(p);
-		::MapWindowPoints(NULL, hWnd, &p, 1);
-		HWND hWin = ::RealChildWindowFromPoint(hWnd, p);
+		::DragQueryPoint(hdrop, &p);
+		HWND hWin = ::ChildWindowFromPointEx(_pPublicInterface->getHSelf(), p, CWP_SKIPINVISIBLE);
 		if (!hWin) return;
 
 		if ((_subEditView.getHSelf() == hWin) || (_subDocTab.getHSelf() == hWin))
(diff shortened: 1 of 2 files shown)


Yaron10:
@Ashfaaq18,

It's working perfectly on my machine. 👍 

donho:
@Ashfaaq18 
Do you mind to explain why the change made the regression and why this PR solve the regression?

Ashfaaq18:
Hi @donho, Initially I thought the problem was with only using ScreenToClient, since in this [winapi](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-screentoclient) documentation they suggest using MapWindowPoints for RTL application. But just using that did not work.

So I guessed maybe something should be wrong with the handle "_pPublicInterface->getHSelf()". Therefore, I tried to replicate the problem by returning the handle at the cursor point through WindowFromPoint, this fixed the problem (my previous PR). But I did not realize that the toolbar tips will stop working.

After, @ozone10 brought up the toolbar tip issue, looks like returning the handle through WindowFromPoint is bringing up another problem. This is where I realized the exact problem. 

In both of these scenarios (the initial RTL issue and the recent toolbar tip issue) the program was incorrectly executed at the if else conditions (line 919 in NppNotifications.cpp), this condition works with the handles returned by the ::RealChildWindowFromPoint() function, after further testing there seemed to be something wrong with the handles returned by this function. 

So, after taking a look at the documentations, I used the function ChildWindowFromPointEx to get further control of the handle returned at the cursor point. By skipping the invisible child windows (RealChild skips transparent), I got the problem fixed.

The problem seemed to lie only at the "RealChildWindowFromPoint" statement, nothing to do with the handles I talked about initially, the "_pPublicInterface->getHSelf()" and "WindowFromPoint". So i reverted it back to what it was before since i didn't want to break anything which was already working.

donho:
Thank you @Ashfaaq18  for the very detailed explanation!

Ashfaaq18:
You're welcome @donho, Happy to help and be a part of this project.

Issue #10353, opened 2021-08-08
Future Request: Add TypeScript as a  new  programing language


Brandonbr1:
It would be nice if Typescript could be added as a programing language in the Languages > T >Typescript

![notepad++_N6R6akgpsn](https://user-images.githubusercontent.com/74426342/128619436-5d27513b-521e-43c3-9df3-7ebb18bb3c33.png)

PR #10361, opened 2021-08-09
Add support TypeScript

0xlay:
Support TypeScript was added: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10353#issue-963344194


donho:
the extension "ts" is recognized as JavaScript, which applies to Cpp lexer also. What makes your PR better?
A screenshot of Before and After could help.

0xlay:
@doho TypeScript has extensions ts, tsx. JavaScript has extensions js, jsx. These are two different languages. TypeScript has static typing, but JavaScript does not.

My PR adds TypeScript lang as a new language.

donho:
> My PR adds TypeScript lang as a new language.

Yes, I do understand what you have added in your PR. However, besides adding  TypeScript lang, I don't see the advantage of syntax highlighting in TypeScript.

0xlay:
@donho I added new keywords to the "langs.model.xml" file for the TypeScript lang that are not in JavaScript. And some of the keywords I removed. I think the logical solution would be to add TypeScript as a new language, rather than mixing everything with one JavaScript. But it's decide you.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3448,6 +3448,8 @@ LangType Notepad_plus::menuID2LangType(int cmdID)
             return L_TXT2TAGS;
         case IDM_LANG_VISUALPROLOG:
             return L_VISUALPROLOG;
+		case IDM_LANG_TYPESCRIPT:
+			return L_TYPESCRIPT;
```
donho, line 3452: Indentation should be aligned with other lines.

PowerEditor/visual.net/notepadPlus.vcxproj
```diff
@@ -530,6 +530,11 @@ copy ..\src\contextMenu.xml ..\binarm64\contextMenu.xml
   <ItemGroup>
     <Image Include="..\src\icons\allChars_off.ico" />
     <Image Include="..\src\icons\allChars_on.ico" />
+    <Image Include="..\src\icons\asciiPanel.ico" />
+    <Image Include="..\src\icons\asciiPanel2.ico" />
+    <Image Include="..\src\icons\chameleon.ico" />
+    <Image Include="..\src\icons\clipboardPanel.ico" />
+    <Image Include="..\src\icons\clipboardPanel2.ico" />
```
donho, line 537: All the diff in `notepadPlus.vcxproj` should be restored.
There's no reason to modify project file.

donho (changes requested):
Please correct the 2 requested points.

0xlay:
@donho This project has a code style problem (mixing tabs and spaces). GitHub doesn't like the mixing of tabs and spaces.

donho (approved):

donho:
@0xlay 
I just realized that in dark mode there's no syntax highlighting in TypeScript document, do you mind to add it in `DarkModeDefault.xml` please ?

donho:
@0xlay 
Another request for auto-completion & function list of type script - but it could be done later.

0xlay:
@donho Ok. I will add in next PR support dark theme and auto-completion & function list of type script

PR #10369, opened 2021-08-10
Fix security flaw issue

### LINKED ISSUE #6165: Nessus reports vulnerability "Microsoft Windows Unquoted Service Path Enumeration"
Opened on behalf of [this forum post](https://community.notepad-plus-plus.org/topic/18279/nessus-reports-vulnerability-microsoft-windows-unquoted-service-path-enumeration)

The installer doesn’t quote the uninstall string appropriately.
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Notepad++\UninstallString is

C:\Program Files\Notepad++\uninstall.exe

when it should be

“C:\Program Files\Notepad++\uninstall.exe”

And similarly in wow64node for 32-bit version.
Can you please tweak the installer to quote this so that people don’t have to waste time “remediating” this?

### LINKED ISSUE #10191: Security flaw
not an issue per say, but a flaw non the less :)

the uninstall path in regedit is not quoted

image

donho:
There's a security flaw for the un-installation of Notepad++ in Windows Register, the string without quotes:
`C:\Program Files\Notepad++\uninstall.exe`, whereas it should be `"C:\Program Files\Notepad++\uninstall.exe"`.
The reason is, hacker can create a file called `c:\program.exe`, then Windows could interpret  `Files\Notepad++\uninstall.exe` as the argument, so the system could run `c:\program.exe`. Ref:
https://isc.sans.edu/diary/Help+eliminate+unquoted+path+vulnerabilities/14464

Fixed by @ozone10:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10191#issuecomment-894811688

Fix #10191, fix #6165

### DIFF
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -171,7 +171,7 @@ Function writeInstallInfoInRegistry
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "Publisher" "Notepad++ Team"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MajorVersion" "${VERSION_MAJOR}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MinorVersion" "${VERSION_MINOR}"
-	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "UninstallString" "$INSTDIR\uninstall.exe"
+	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "UninstallString" '"$INSTDIR\uninstall.exe"'
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "DisplayIcon" "$INSTDIR\notepad++.exe"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "DisplayVersion" "${APPVERSION}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "URLInfoAbout" "${APPWEBSITE}"
--- a/PowerEditor/Test/FunctionList/nsis/unitTest
+++ b/PowerEditor/Test/FunctionList/nsis/unitTest
@@ -30,7 +30,7 @@
 !define VERSION_MAJOR 6
 !define VERSION_MINOR 9
 
-!define APPWEBSITE "http://notepad-plus-plus.org/"
+!define APPWEBSITE "https://notepad-plus-plus.org/"
 
 !define UNINSTALL_REG_KEY "Software\Microsoft\Windows\CurrentVersion\Uninstall\${APPNAME}"
 !define MEMENTO_REGISTRY_ROOT HKLM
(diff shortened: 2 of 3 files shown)


PowerEditor/installer/nsisInclude/tools.nsh
```diff
@@ -171,7 +171,7 @@ Function writeInstallInfoInRegistry
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "Publisher" "Notepad++ Team"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MajorVersion" "${VERSION_MAJOR}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MinorVersion" "${VERSION_MINOR}"
-	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "UninstallString" "$INSTDIR\uninstall.exe"
+	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "UninstallString" '"$INSTDIR\uninstall.exe"'
```
mere-human, line 174: Should this be double quoted?
Like this
image

mere-human: For some reason, GitHub failed to upload my image. So here it is:
https://imgur.com/2d7Ef59

donho: @mere-human 
Shouldn't it be? What's your argument?

donho: @mere-human 
Just updated my explanation of this security flaw.

MarkusBodensee: @mere-human The single quotes are correct to make nested double quotes appear in the string which will actually go to the registry.

See this link for confirmation, originally provided by @ozone10 : https://nsis.sourceforge.io/mediawiki/index.php?title=How_can_I_use_quotes_in_a_string%3F

MarkusBodensee:
Should the same be done for other Path strings, too? Or is it a really special case for the uninstaller?

Especially, the right click context menu entry without quotes looks dangerous.

See:
RightClickRegistry

PowerEditor/installer/nsisInclude/tools.nsh
mere-human: Thanks @MarkusBodensee that answers my question.

dail8859 (approved):

donho:
@MarkusBodensee 

It should be quoted. Could you provide the path of entry in registry please?

MarkusBodensee:
@donho 
On my machine it is on the following two paths, but I am not sure if the GUID is the same on every machine:
`HKEY_CLASSES_ROOT\CLSID\{B298D29A-A6ED-11DE-BA8C-A68E55D89593}\Settings`
`HKEY_LOCAL_MACHINE\SOFTWARE\Classes\CLSID\{B298D29A-A6ED-11DE-BA8C-A68E55D89593}\Settings`

I just searched in the registry for "notepad++" to find locations which could be problematic.

I tried to find the code in the installer files, but haven't found it until now.

donho:
@MarkusBodensee 
Thank you for your info.

I managed to make NppShel.dll registry settings (Notepad++ context menu) to have the quote:

image

However, when I selected a file, and right click for the context menu then run "Edit with Notepad++", I've got the following error:

image

I think it's a special case for the uninstaller.

MarkusBodensee:
I rechecked the other registry entries which I thought could be problematic at first glance, but could not make them break anything or find any misbehaviour. And at least they look the same like for example vlc media player, too. So they should be ok. Uninstaller is really a special one.

PR #10378, opened 2021-08-12
Use only ASCII characters in LICENSE text

### LINKED ISSUE #10370: Seems smart quotes in License text transform into strange symbols
### Description of the Issue

Screenshot

(It probably occurs in few more places in License text)

This issue was found in Notepad++ v8.1.2 (though it probably spans since v7.9.3) Windows 32-bit
Some details were offered in forum thread [https://community.notepad-plus-plus.org/topic/21622/notepad-v8-1-2-license-strange-symbols](https://community.notepad-plus-plus.org/topic/21622/notepad-v8-1-2-license-strange-symbols)
It was suggested that this problem started with commit [ab5c1d3](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ab5c1d3)

mere-human:
Replace “ and ” by ", replace © by (C)

Fix #10370


ValZapod:
The actual fix is of course to fix the charset in the installer.

mere-human:
> The actual fix is of course to fix the charset in the installer.

Sure. But the current solution in this PR was easier and faster to do.
I don't think it is worth spending more time on this minor issue.

PR #10383, opened 2021-08-15
Make dark mode compatible with Windows 11

ozone10:
Use documented way to enable UxTheme dialog texture.
Remove undocumented way.
Disable UxTheme dialog texture in dark mode.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10299
alternative to https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10300

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -1433,10 +1433,7 @@ namespace NppDarkMode
 			, theme
 		};
 
-		if (subclass)
-		{
-			::EnableThemeDialogTexture(hwndParent, theme ? ETDT_ENABLETAB : ETDT_DISABLE);
-		}
+		::EnableThemeDialogTexture(hwndParent, theme && !NppDarkMode::isEnabled() ? ETDT_ENABLETAB : ETDT_DISABLE);
 
 		EnumChildWindows(hwndParent, [](HWND hwnd, LPARAM lParam) {
 			auto& p = *reinterpret_cast<Params*>(lParam);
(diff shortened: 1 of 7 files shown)


donho:
@ozone10 
I am not at home for the moment so I don't have Windows 8 machine with me to test this PR. But I do believe it'll make some visual progression in Windows 8.1, some old Windows 10 (no upgraded) and even in Windows 7.

Have you tested under some of mentioned OS above for preventing from the regression?

ozone10:
@donho 
`EnableThemeDialogTexture` is supported from Windows Vista (https://docs.microsoft.com/en-us/windows/win32/api/uxtheme/nf-uxtheme-enablethemedialogtexture)

I have tested it in Windows 7, 10 (19041 and latest – supported versions) and 11. 
There should be no change how it works in Windows 8, 8.1 and older 10 (unsupported versions).
The only version which this PR will break is Windows XP.

I've also not removed `NPPM_GETENABLETHEMETEXTUREFUNC`, related `getEnableThemeDlgTexture()` function and variables in `Parameters.h`, because I was not sure if it is used by plugins.

mere-human:
> The only version which this PR will break is Windows XP.

It's been broken for a while. The latest Notepad++ won't install/launch on XP.

donho:
@ozone10 
Thank you for the info & testing.
As @mere-human mentioned above, XP has been dropped in previous release.
I'll do a test next weekend under Windows 8 to make sure there's no regression.

donho:
@ozone10 
Sorry about that, could you solve the conflict please?

PR #10387, opened 2021-08-15
Use current buffer's path in Find in Files dialog

### LINKED ISSUE #8045: Find-in-Files "Follow current doc" not working when absolute Default Directory set
### Description of the Issue
When a *Default Directory (Open/Save)* setting is set to be an absolute folder path (the 3rd of 3 radio buttons) in *Preferences* > *Default Directory* > *Default Directory (Open/Save)*, the *Follow current doc.* checkbox on the *Find in Files* tab of the *Find* window DOES NOT work to change the *Directory:* box value for *Find in Files*.  The *Directory:* box value is always set to the chosen absolute path in the *Preferences*.

1. Open *Preferences* > *Default Directory* and tick the 3rd (lowest) of 3 radio buttons, and provide an absolute path, for example: `W:\`
2. Open a file on C: drive, for example, `readme.txt` that comes with Notepad++
3. Invoke *Find in Files* by pressing Ctrl+Shift+f
4. Tick the *Follow current doc.* checkbox on the *Find in Files* tab.  Or, tick and untick it  a few times, and observe that the C: drive folder path of `readme.txt` _NEVER_ appears in the *Directory:* box, as it should.

### Expected Behavior
The folder path of the active file in the editor should appear in the *Directory:* box after an untick/tick cycle of *Follow current doc.*

### Actual Behavior
The correct folder path never appears in *Directory:*.

### Debug Information
Notepad++ v7.8.5   (32-bit)
Path : C:\.......\npp.7.8.5.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.973

### Relevant Screenshots

image

image

### Additional information
Even if the *Preference* settings is moved away from the 3rd (bottommost) absolute path setting for *Default Directory (Open/Save)*, if there is **ANY TEXT** at all in the box to the right of this bullet, then the *Follow current doc.* checkbox on the *Find in Files* tab of the *Find* window DOES NOT work to change the *Directory:* box value for *Find in Files*.

mere-human:
"Follow current doc." just took the working directory path when checked.
But the working directory could be set to an absolute directory in "Default Directory" preferences.

Now when "Follow current doc." is checked, use the current path first.
If that fails, fallback to the working directory as it was before.

Fix #8045

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -16,7 +16,6 @@
 
 #include <shlwapi.h>
 #include <uxtheme.h>
-#include "Common.h"
 #include "FindReplaceDlg.h"
 #include "ScintillaEditView.h"
 #include "Notepad_plus_msgs.h"
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -14,8 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-#include <shlobj.h>
+#include <shlwapi.h>
 #include <uxtheme.h>
+#include "Common.h"
 #include "FindReplaceDlg.h"
 #include "ScintillaEditView.h"
 #include "Notepad_plus_msgs.h"
@@ -1822,9 +1823,19 @@ INT_PTR CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 
 					if (findHistory._isFolderFollowDoc)
 					{
-						NppParameters& nppParam = NppParameters::getInstance();
-						const TCHAR * dir = nppParam.getWorkingDir();
-						::SetDlgItemText(_hSelf, IDD_FINDINFILES_DIR_COMBO, dir);
+						// Working directory depends on "Default Directory" preferences.
+						// It might be set to an absolute path value.
+						// So try to get the current buffer's path first.
+						generic_string currPath;
+						const Buffer* buf = (*_ppEditView)->getCurrentBuffer();
+						if (!(buf->getStatus() & (DOC_UNNAMED | DOC_DELETED)))
+						{
+							currPath = buf->getFullPathName();
+							PathRemoveFileSpec(currPath);
+						}
+						if (currPath.empty() || !PathIsDirectory(currPath.c_str()))
+							currPath = NppParameters::getInstance().getWorkingDir();
+						::SetDlgItemText(_hSelf, IDD_FINDINFILES_DIR_COMBO, currPath.c_str());
 					}
 
 				}


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -14,8 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-#include <shlobj.h>
+#include <shlwapi.h>
 #include <uxtheme.h>
+#include "Common.h"
```
donho, line 19: `Common.h` is not needed.
Please remove it.

donho (changes requested):
Remove line `#include Common.h`

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
mere-human: It is needed because I used `PathRemoveFileSpec()` overload for `generic_string` declared in `"Common.h"`.

donho: The signature of common.h's version is `PathRemoveFileSpec(generic_string&)`
shlwapi.h's version one is `PathRemoveFileSpec(TCHAR*)`
The `PathRemoveFileSpec` you use here passed a generic_string object as argument so there's no ambiguity.
"Common.h" is already included in "localization.h" which is included in "FindReplaceDlg.cpp". 
So even you remove the line `#include "Common.h"`, compiler can link to the right function.
For the sake of maintenance facility, in both *.h & *.cpp, we try to include as minimun header as passible.

mere-human: I think this is micro-management.
Each header should have either [pragma once](https://en.wikipedia.org/wiki/Pragma_once) or [include guard](https://en.wikipedia.org/wiki/Include_guard) (`Common.h` has `#pragma once`).
So it is the task of the compiler (not the programmer) to handle the fact that a file is included multiple times.

Each file should include all headers it requires to make it self-sufficient. It is bad to rely on includes in other headers. What if `Common.h` will be removed from `localization.h`? Then you'll have to fix every place that was relying on the fact that it included `Common.h`. This makes the code really fragile.

On the other hand, if you're _really_ concerned with the build time, I'd suggest you to create a dedicated GitHub issue.
Be sure, I have some ideas for that too.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-#include <shlobj.h>
+#include <shlwapi.h>
```
mere-human, line 17: How about this change? Well, with `shlobj.h` I had a weird error that could be a result of includes order, redefinition or whatever. 
`findreplacedlg.cpp(1833): error C2664: 'BOOL PathRemoveFileSpecW(LPWSTR)': cannot convert argument 1 from 'generic_string' to 'LPWSTR'`
Including `shlwapi.h` instead fixes the problem.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: It's not even a "management". I try to make all contributed code into master conform my standard/style so I could maintain in the future.

> Each file should include all headers it requires to make it self-sufficient. It is bad to rely on includes in other headers. What if Common.h will be removed from localization.h? Then you'll have to fix every place that was relying on the fact that it included Common.h. This makes the code really fragile.

I got your point but I don't consider it's "fragile". The situation you described comes only on compiling time but not on the runtime - the dev adds the header needed and the compiling problem solved.

> On the other hand, if you're really concerned with the build time,  I'd suggest you to create a dedicated GitHub issue.
Be sure, I have some ideas for that too.

The goal of including as less headers as possible is for the maintenance, not for reducing the build time. I care about runtime whereas the build time it doesn't matter to me. But feel free to create issues if you have some suggestions by tagging me.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: No idea about this error.
But I think it's better use shlwapi instead of shlobj.

donho (approved):

PR #10400, opened 2021-08-17
Fix file dialog append extension with RTL language

### LINKED ISSUE #10397: NPP RTL: file wrongly saved without an extension.
**STR:**
Windows RTL + NPP RTL.
Open a new file and press Save.
Append Extension is checked and the "Save As Type" field contains "Normal Text File (*.txt)".
In the "File Name" field replace "New X.txt" with "Test" (without .txt).
Save.

**Result:**
"Test" is saved without an extension.

**Expected:**
"Test" should be saved as "Test.txt".
This is the result in MS-Notepad and some other programs I've tried.
It also used to be the behavior in older Notepad++ versions.

***

Notepad++ v8.1.3   (32-bit)
Build time : Aug 10 2021 - 00:32:53
Path : C:\\notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255

mere-human:
When both Windows and Notepad++ use RTL language, the Save/Open button
is swapped with Cancel button.
Now take that into account when searching the button.

Fix #10397

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -465,6 +465,7 @@ private:
 	{
 		const int bufferLen = MAX_PATH;
 		static TCHAR buffer[bufferLen];
+		static bool isRTL = false;
 
 		auto* inst = reinterpret_cast<FileDialogEventHandler*>(param);
 		if (!inst)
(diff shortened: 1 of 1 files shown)


mere-human:
RTL language - any language that reads from right to left (Arabic, Hebrew, etc).
Windows 10 language can be changed in Settings > Time & Language > Language > Add Language.
Notepad++ UI language affects dialogs. It can be changed in Preferences > General > Localization.
**Note:** View > Text direction RTL _does not affect_ this case.

**Possible language combinations:**
1. Windows EN, N++ RTL - file dialog EN, main window RTL
2. Windows RTL, N++ RTL - file dialog EN, main window EN
3. Windows RTL, N++ RTL - file dialog RTL, main window RTL

The problem is reproducible only in combination 3.
I tested it on Windows 10 because I don't have a localization for my Windows 7 VM.
For English (and any other LTR) language, the behavior should be the same as before.

**Note:** I don't know Arabic or any other RTL language.

Save dialog with Arabic language:
لقطة الشاشة 2021-08-16 231345

Window style for a button in this case:
لقطة الشاشة 2021-08-16 232612

Window style for a button with English language:
لقطة الشاشة 2021-08-16 232650

mere-human:
Actually, I had another idea how to fix it: use window hooks instead of sub-classing controls.
But I decided to went the easiest way possible.
Maybe, if more file dialog problems show up, I'll try that window hook idea then.

Yaron10:
@mere-human,

Thank you for the quick fix. It's working perfectly now. 👍 
Special thanks for changing your system language.

donho:
@mere-human 
> When both Windows and Notepad++ use RTL language, the Save/Open button
is swapped with Cancel button.

According to the description of #10397 (provided by @Yaron10) , it's a bug of Append Extension checkbox (not working), and Save button works (Otherwise he cannot save the file). It doesn't much what your description. Or am I missing somethings?

mere-human:
@donho 
> According to the description of #10397 (provided by @Yaron10) , it's a bug of Append Extension checkbox (not working), and Save button works (Otherwise he cannot save the file). It doesn't much what your description. Or am I missing somethings?

Save button works. 
But the reason why the extension isn't appended is that `onPreFileOk()` isn't being called.
It is not called because we failed to find the Save button and override it's window procedure.

PR #10406, opened 2021-08-19
Update czech.xml

xomx:


donho:
@xomx 
Could you please update your translation with it:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10408

xomx:
@donho 
Done.
I have also reviewed again the Mark&Style stuff, I hope that this time I got it right (I am not using it).

donho:
Please redo your PR, the file is changed entirely:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10406/files

xomx:
@donho 
Sorry, I cannot do that until Monday. I have the source czech.xml file at my PC at work and unfortunately I switched it off before weekend, so I cannot connect to it via RD. And I did not succeed to restore the newest file from the closed PR, Github are showing to me only my 2 patches (but incomplete, as they are very long - it incorrectly thinks that these are completely different). And when I look into my patch-36, the czech.xml file is empty.

donho:
No problem @xomx !

Issue #8574, opened 2020-07-16
Checkbox overlap in incremental search bar


Myvh:
### Description of the Issue
In the incremental search bar at the bottom of the window, the checkbox labeled "Match case" is slightly hidden by the "Highlight all" checkbox as soon as the "Highlight all" checkbox is moused over. If "Match case" is moused over, the checkbox is refreshed and is displayed normally again.

![Annotation 2020-07-16 153542](https://user-images.githubusercontent.com/68384832/87678094-d3b4cb00-c77a-11ea-80d1-8b1a0920a977.png)

### Steps to Reproduce the Issue
1. Open incremental search.
2. Mouse over the "Highlight all" checkbox.

### Debug Information
Notepad++ v7.8.7   (64-bit)
Build time : Jun  5 2020 - 02:38:56
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.959
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

sasumner:
I suspect this is a DPI issue.  I couldn't replicate on my normal setup (main display is 96dpi) but while I was on vacation using only my laptop's display (120dpi), I noticed this same visual artifcact.

mshingote:
I'm working on it.

mere-human:
I can't replicate it also
![image](https://user-images.githubusercontent.com/9664141/130316741-d7f5bec6-ad1b-4e4f-85c0-ca384e70a7ad.png)
```
Notepad++ v8.1.3   (64-bit)
Build time : Aug 16 2021 - 10:26:44
Path : ...\Notepad++.exe
Command Line : -multiInst -nosession
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2009
OS Build : 19043.1165
Current ANSI codepage : 1252
Plugins : none

```

mshingote:
@mere-human It's reproducible at my end.
You need to hover mouse pointer on `Highlight all` checkbox

mshingote:
Before fix widths are overlapping
![image](https://user-images.githubusercontent.com/8279344/130317148-2a07934a-ffaa-4a83-9826-58b126d65f30.png)

After fix overlap is not observed
[...]

PR #10419, opened 2021-08-21
Fixing overlapped checkbox in incremental search bar

mshingote:
'Highlight all' and `Match case` checkbox width are overlapping that is the reason for UI overlapping on mouse hover.
This PR corrected width of `Highlight all` same as `Match case` after this no overlap observed.

Open #8574


mshingote:
@donho any comment is appreciated

mere-human:
Just a thought: with this change, there can be not enough room for a label in a different language. Have you checked other localizations?

mshingote:
good point, another way I can increase x coordinate of next label so that it wont compromise with size.. next change is on the way

PR #10420, opened 2021-08-21
Make Incremental Search panel translatable

donho:
Fix also checkbox overlap in incremental search bar.

Fix #8955, fix #8574

PR #10421, opened 2021-08-21
Fixing overlapped checkbox in incremental search bar

mshingote:
Width of 'Highlight all' and 'Match case' are overlapping that is the reason for UI overlapping on mouse hover.
This PR moved 'Match case' to right side by adjusting its x-coordinate and width so that there will be room for other localizations.
After this change no overlap observed. #8574


donho:
Thank you for the PR.
However, the PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10420 will fix the issue. So this PR will not be taken.

Issue #10052, opened 2021-06-21
Dark Mode: Document Map Improvements


WildByDesign:
This is a feature request for improvements to the Document Map in Dark Mode.

I have noticed that one of the last remaining things that needs some improvements in Dark Mode is the Document Map. In particular, that awful Orange color combined with a frosted white over the background of the doc map is dreadful in dark mode. Possibly Light Mode could do without the dreaded Orange.

**Here is a working example for Dark Mode:** 

![NppDarkMode-DocumentMap](https://user-images.githubusercontent.com/26308319/122781098-7c1dba00-d27d-11eb-8104-33adff58e336.png)
_(Before on left and After on right.)_

Please keep in mind that I am not a programmer. However, I have achieved this with a small code change that benefits Dark Mode. But likely would need a real programmer to make it change automatically between Dark Mode and Light Mode.

I changed the following code:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/DocumentMap/documentMap.cpp#L449

Changed to:

```
	const COLORREF orange = RGB(0x80, 0x80, 0x80);
	const COLORREF white = RGB(0x30, 0x30, 0x30);
```

After that small change and compiling the binary, Document Map in Dark Mode is smooth and silky. I wish that I had time to create and animated gif just to show how easy it is on the eyes while scrolling the document map around. But trust me, it is beautiful.

**EDIT:** I just decided to try it with Light Mode as well. It's actually pretty good there as well, but not perfect. It's still better in comparison to the Orange color. So it might need to changes for Light Mode and particularly switching back and forth between Dark and Light modes.

ozone10:
One issue is that user can use dark mode but with light style theme, or vice versa.

Solution would be allow user to define colors themself via style configurator.

WildByDesign:
@ozone10  Great point and perfect suggestion as well. I agree that exposing both foreground color and background color of the Document Map in Style Configurator would be the most ideal solution.

WildByDesign:
@ozone10 You are a hero! Cheers!

PR #10422, opened 2021-08-21
Allow custom color for Document Map

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10052
image
image

Document map style is added only to default stylers, still need to update other stylers.


donho:
BTW, DarkModeDefault.xml is not modified.

PowerEditor/src/WinControls/DocumentMap/documentMap.cpp
```diff
@@ -442,14 +444,36 @@ INT_PTR CALLBACK DocumentMap::run_dlgProc(UINT message, WPARAM wParam, LPARAM lP
...
+			return;
+	}
+}
+
 void ViewZoneDlg::drawPreviewZone(DRAWITEMSTRUCT *pdis)
 {
 	RECT rc = pdis->rcItem;
-	
-	const COLORREF orange = RGB(0xFF, 0x80, 0x00);
-	const COLORREF white = RGB(0xFF, 0xFF, 0xFF);
```
ozone10, line 450: @donho 

> IMO, the background colour should remain as the main edit zone - think of Dark Mode. Don't you think so?

Part outside focus is using by default white color (`RGB(0xFF, 0xFF, 0xFF)`) so it may seems like there is no change in background color.

Those images are examples, and background color is changed to make notice of difference,

PR #10426, opened 2021-08-21
Refresh the working directory when changing Follow current doc. preference

### LINKED ISSUE #10396: Wrong default folder in Open/Save dialog
**STR:**
Settings -> Preferences -> Default Directory.
Check the third RadioButton and in the EditBox enter "C:\Users\UserName\Downloads".
Close NPP.

Restart NPP (New 1, no session).
Open a file located in any folder other than Downloads.
Settings -> Preferences -> Default Directory.
Check "Follow current document".
File -> Open (or Save As).

**Result:**
Downloads is the default folder in the dialog.
***

Notepad++ v8.1.3.
The issue existed in older NPP versions as well.

mere-human:
Fix #10396

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5943,9 +5943,7 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
 			command(IDM_VIEW_REFRESHTABAR);
 		checkDocState();
 		setTitle();
-		generic_string dir(buffer->getFullPathName());
-		PathRemoveFileSpec(dir);
-		setWorkingDir(dir.c_str());
+		setWorkingDir(buffer);
 	}
 
 	if (mask & (BufferChangeLanguage))
@@ -6000,9 +5998,7 @@ void Notepad_plus::notifyBufferActivated(BufferID bufid, int view)
 	setUniModeText();
 	setDisplayFormat(buf->getEolFormat());
 	enableConvertMenuItems(buf->getEolFormat());
-	generic_string dir(buf->getFullPathName());
-	PathRemoveFileSpec(dir);
-	setWorkingDir(dir.c_str());
+	setWorkingDir(buf);
 	setTitle();
 	//Make sure the colors of the tab controls match
 	::InvalidateRect(_mainDocTab.getHSelf(), NULL, FALSE);
@@ -6281,6 +6277,14 @@ void Notepad_plus::setWorkingDir(const TCHAR *dir)
 	{
 		params.setWorkingDir(dir);
 	}
+}
+
+void Notepad_plus::setWorkingDir(const Buffer* buf)
+{
+	assert(buf);
+	generic_string dir(buf->getFullPathName());
+	PathRemoveFileSpec(dir);
+	setWorkingDir(dir.c_str());
 }
 
 int Notepad_plus::getLangFromMenuName(const TCHAR * langName)
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5943,7 +5943,7 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
 			command(IDM_VIEW_REFRESHTABAR);
 		checkDocState();
 		setTitle();
-		setWorkingDir(buffer);
+		setWorkingDir(buffer->getFullPathName());
 	}
 
 	if (mask & (BufferChangeLanguage))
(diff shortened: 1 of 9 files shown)


Yaron10:
@mere-human,

Thank you for your work. 👍 

Your PR fixes the case in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10396, but it creates another issue.

**STR:**
Settings -> Preferences -> Default Directory.
Check the third RadioBox and in the EditBox enter "C:\Users\UserName\Downloads".
Close NPP.

Restart NPP (New 1, no session).
Open a file located in any folder other than Downloads.
Settings -> Preferences -> Default Directory.
Check "Follow current document".
_Re-check the third RadioBox (Downloads)._
File -> Open (or Save As).

**Result:**
The current doc folder appears in the dialog. 

***



When "Follow current doc" or "Last used directory" is checked on starting NPP (New 1) and "File -> Open" - NPP installation folder appears in the dialog. 

**STR:**
Settings -> Preferences -> Default Directory.
Check the third RadioBox and in the EditBox enter "C:\Users\UserName\Downloads".
Close NPP.

Restart NPP (New 1, no session).
Settings -> Preferences -> Default Directory.
Check "Follow current document" or "Last used directory".
File -> Open (or Save As).

**Result:**
Downloads is the default folder in the dialog.

This case can be regarded as a minor issue or a non-issue.

mere-human:
@Yaron10 the 1st case should be fixed now. 

The 2nd case remains as it is since I consider it as a non-issue. 
Because when the current file doesn't exist on disk, there's no last used directory.
Or is it really necessary to have the installation directory as working directory in this case?

Yaron10:
> the 1st case should be fixed now.

👍 
It's working beautifully now.

> The 2nd case remains as it is since I consider it as a non-issue.

OK. 
I thought it might be a minor issue because of the inconsistency, but it's probably not worth the extra code.

PowerEditor/src/NppCommands.cpp
```diff
@@ -3594,6 +3594,12 @@ void Notepad_plus::command(int id)
 		}
 		break;
 
+		case IDC_REFRESH_WORK_DIR:
+		{
+			setWorkingDir(_pEditView->getCurrentBuffer());
+		}
+		break;
+
```
donho, line 3602: Except some legacy code, `NppCommands.cpp` contains only the user commands (ie. menu commands).
The above code should be rather in `NppBigSwitch.cpp`

PowerEditor/src/resource.h
```diff
@@ -431,6 +431,7 @@
     //#define IDC_KEY_END                        (IDCMD+7)
     //#define IDC_KEY_SELECT_2_HOME            (IDCMD+8)
     //#define IDC_KEY_SELECT_2_END            (IDCMD+9)
+    #define IDC_REFRESH_WORK_DIR           (IDCMD+10)
```
donho, line 434: It should be rather `NPPM_INTERNAL_REFRESHWORKDIR`.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6283,6 +6279,20 @@ void Notepad_plus::setWorkingDir(const TCHAR *dir)
...
+{
+	if (buf)
+	{
+		generic_string dir(buf->getFullPathName());
+		PathRemoveFileSpec(dir);
+		setWorkingDir(dir.c_str());
+	}
+	else
+	{
+		setWorkingDir(static_cast<const TCHAR*>(nullptr));
```
donho, line 6292: Why do we need to add this method, whereas `setWorkingDir(TCHAR*)` is called inside ?

donho (changes requested):
1. Remove `setWorkingDir(const Buffer*)` and enhance (if needed) and call `:setWorkingDir(const TCHAR*)` directly.
2. Define `NPPM_INTERNAL_REFRESHWORKDIR` instead of `IDC_REFRESH_WORK_DIR`

PowerEditor/src/resource.h
mere-human: Thanks for explanation. I'm not very familiar with this part of the app.

PowerEditor/src/NppCommands.cpp
mere-human: I see, thanks.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2669,6 +2669,13 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return TRUE;
 		}
 
+		case NPPM_INTERNAL_REFRESHWORKDIR:
+		{
+			const Buffer* buf = _pEditView->getCurrentBuffer();
+			setWorkingDir(buf ? buf->getFullPathName() : nullptr);
+		}
+		return TRUE;
```
MarkusBodensee, line 2677: Return statement should be inside parenthesis like in the other cases.

donho (approved):

donho:
@mere-human
Thank you for the change - it's much easier for reviewing now and especially that we are sure there'll be no regression.

PR #10429, opened 2021-08-22
Make "Save All" dialog more clear

### LINKED ISSUE #10339: Consider renaming cancel button in save all dialogue
The ``cancel`` button on the save all dialogue is kind of non intuitive and should be renamed to ``Yes, don't ask again``.
Yes, while the box says exactly what the ``cancel`` button does, I think it's function can be made more descriptive by renaming it to say ``Yes, don't ask again``.
Background of this is, the save and save-all button are right next to each other, so a misclick is guaranteed to happen at some point. When that happens I assume most people don't read what the dialogue box says but just hit the ``cancel`` button because essentially everywhere cancel means don't do whatever I just clicked. Renaming that button to ``Yes, don't ask again`` most likely causes people to hit ``no`` instead.
image

### Debug Information
Notepad++ v8.1.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1110
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll NppSaveAsAdmin.dll

### LINKED ISSUE #10340: Change function of save all dialogue "x" button
### Description of the Issue
In the save all dialogue hitting the ``x`` to close the popup behaves the same as when hitting the ``cancel`` button.

1. Make sure the ``enable save all confirm dialogue`` box under preferences is ticked
2. Modify a file and hit the save-all button
3. Press the ``x`` on the popup window
image

### Expected Behavior
File does not get saved when hitting the ``x``.

### Actual Behavior
File gets saved when hitting the ``x`` and instead should behave like pressing ``no``.

### Debug Information
Notepad++ v8.1.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1110
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll NppSaveAsAdmin.dll

donho:
Fix #10339, fix #10340


ArkadiuszMichalski:
Pressing ESC and [X] mean the same as Always YES? Because now it is like that, and it should probably just close this window without any action?

mere-human:
> Pressing ESC and [X] mean the same as Always YES?

I agree, closing the dialog should _cancel the saving process_. It shouldn't modify any preferences.
image
Also, the dialog is lacking the Cancel button.
Please, take the File > Close All dialog as an example:
image

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -453,3 +453,93 @@ INT_PTR CALLBACK DoSaveOrNotBox::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
+		::GlobalFree(hMyDlgTemplate);
+	}
+	else
+		::DialogBoxParam(_hInst, MAKEINTRESOURCE(IDD_DOSAVEALLBOX), _hParent, dlgProc, reinterpret_cast<LPARAM>(this));
+}
+
+void DoSaveAllBox::changeLang()
+{
+	generic_string msg;
+	generic_string defaultMessage = TEXT("Are you sure you want to save all documents?\r\rChoose \"Always Yes\" if your answer will always be \"Yes\" and you won't be asked this question again.\rYou can re-activate this dialog in Preferences dialog later.");
```
mere-human, line 475: > Choose \"Always Yes\" if your answer will always be \"Yes\"

Don't you think it's a bit of a tautology?🙂
I would remove this text at all and show it as a tooltip.

ArkadiuszMichalski:
>@mere-human 
Also, the dialog is lacking the Cancel button. 

Probably because NO means the same thing, it doesn't make sense to use two buttons. At the moment this dialog applies to all files (only) so NO can be treated as Cancel.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
donho: It's OK if it's simple & clear to users. it doesn't matter tautology to me. And tooltip is not necessary finally.

donho: But if you have a better text, feel free to provide.

donho:
> Pressing ESC and [X] mean the same as Always YES? Because now it is like that, and it should probably just close this window without any action (same as NO)?

It's fixed now.

mere-human:
> Pressing ESC and [X] mean the same as Always YES?

By the way, here is the issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10340

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
mere-human: I would change it to something like this:
```
Choose "Always Yes" if you don't want to see this dialog again.
You can re-activate this dialog in Preferences later.
```
It is pretty clear what "Always Yes" does, so don't need to repeat it.

mere-human:
I would also recommend to reorder the buttons this way:
`[Yes] [Always Yes] [No]`
Negative option is usually the rightmost one.

PR #10453, opened 2021-08-28
Add custom date time insert

### LINKED ISSUE #10445: "Insert date time" should insert date and then time
### Description of the Issue
With the commit `0733e6f2413636f55f2b7a4b6631c4531730ef61`, the feature "Insert date and time" was implemented. It inserts first the time and then (after a space) the date.

I personally would expect it the other way around, i.e. first the date and then the time. Am I alone in my expectation?

I assume the current implementation was copied from the standard windows editor. I find the implementation there somewhat weird.

### Expected Behavior
"Insert date time" inserts `<date> <time>` and not `<time> <date>`.

donho:
Make Insert date time customizable.
Also make date time order reversable for short & long format.

Fix #10445


donho:
Any suggestion/idea is welcome.

PowerEditor/src/NppCommands.cpp
```diff
@@ -69,19 +70,44 @@ void Notepad_plus::command(int id)
...
-			GetDateFormatEx(LOCALE_NAME_USER_DEFAULT, dateFlag, &st, NULL, dateStr, sizeof(dateStr) / sizeof(dateStr[0]), NULL);
-			GetTimeFormatEx(LOCALE_NAME_USER_DEFAULT, TIME_NOSECONDS, &st, NULL, timeStr, sizeof(timeStr) / sizeof(timeStr[0]));
+			GetDateFormatEx(LOCALE_NAME_USER_DEFAULT, dateFlag, &currentTime, NULL, dateStr, sizeof(dateStr) / sizeof(dateStr[0]), NULL);
+			GetTimeFormatEx(LOCALE_NAME_USER_DEFAULT, TIME_NOSECONDS, &currentTime, NULL, timeStr, sizeof(timeStr) / sizeof(timeStr[0]));
 
-			generic_string dateTimeStr = timeStr;
-			dateTimeStr += TEXT(" ");
-			dateTimeStr += dateStr;
+			generic_string dateTimeStr;
+			if (!NppParameters::getInstance().getNppGUI()._dateTimeReverseDefaultOrder)
```
mere-human, line 84: Prefer non-inverted conditions in if/else. It is usually easier to understand.
```
if (..._dateTimeReverseDefaultOrder)
{
...
}
else
{
...
}
```

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1328,3 +1328,26 @@ int nbDigitsFromNbLines(size_t nbLines)
 	}
 	return nbDigits;
 }
+
+generic_string getDateTimeStrFrom(const generic_string& dateTimeFormat, const SYSTEMTIME& st)
```
mere-human, line 1332: Why is it needed?
We can just use `GetDateFormatEx` here as well. As for me, it is more powerful and stable than manual string parsing.
You can call it with `LOCALE_NAME_INVARIANT`.
If users like it, we even could expose locale settings, etc in preferences.

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -169,7 +169,7 @@ INT_PTR CALLBACK PreferenceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			_wVector.push_back(DlgInfo(&_searchingSubDlg, TEXT("Searching"), TEXT("Searching")));
 			_wVector.push_back(DlgInfo(&_backupSubDlg, TEXT("Backup"), TEXT("Backup")));
 			_wVector.push_back(DlgInfo(&_autoCompletionSubDlg, TEXT("Auto-Completion"), TEXT("AutoCompletion")));
-			_wVector.push_back(DlgInfo(&_multiInstanceSubDlg, TEXT("Multi-Instance"), TEXT("MultiInstance")));
+			_wVector.push_back(DlgInfo(&_multiInstanceSubDlg, TEXT("Multi-Instance & Date"), TEXT("MultiInstance")));
```
mere-human, line 172: How about to put it to "Edit" or "Misc." category?
It looks like it doesn't belong to "Multi-instance".

MarkusBodensee:
It depends on how complex you want your implementation to be. ;-)

What about controlling the addition of leading zeroes? Or the number of digits used for the year?
Example: 11. May 2009 at time 01:02:15
```
YYYY-MM-DD hh:mm:ss --> 2009-05-11 01:02:15
YY-MM-DD h:mm:ss --> 09-05-11 1:02:15
Y-M-D h:m:s --> 0-5-11 1:2:15
YYYY-MM-D --> 2009-05-11
```

Or controlling 24h or 12h format with AM/PM? Some ideas [here](https://www.c-sharpcorner.com/blogs/date-and-time-format-in-c-sharp-programming1).

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
donho: There's no more room in both "Edit" & "Misc." categories.
"Multi-instance" is one of rare pages which have still some room.

mere-human: How about adding a new page?

mere-human:
> What about controlling the addition of leading zeroes? Or the number of digits used for the year?

This should be possible by exposing the `GetDateFormatEx` to users. It's format spec should show that.

PowerEditor/src/MISC/Common/Common.cpp
donho: GetDateFormatEx is only for Date, with the time it's necessary to use GetTimeFormatEx.
For format "hh:mm:ss YYYY-MM-DD", we still need to parse the format string to split it into 2 format.
Here my goal is to provide an easy format - so it's easy to understand for user and easy to implement for dev.

donho:
@MarkusBodensee 
It's intuitive and conventional. But difficult to implemented.

@mere-human 
Win32 API uses 2 functions for formatting date and time. Whereas my goal is 1 format for both. So I take the easy solution for providing an easy format to users.

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
donho: I'm trying not overload the left menu, so it's not considerable for this category.

mere-human:
> Win32 API uses 2 functions for formatting date and time. Whereas my goal is 1 format for both. So I take the easy solution for providing an easy format to users.

@donho This is not a problem. You still can use 2 functions subsequently on one format string.
Take a look at this example:
https://github.com/mere-human/win32-samples/commit/2b53701a85a894bd6e76c903e5ae3179e76efefa#diff-68197c399ce0334fef526c3b9a4f6e09e3627caecf70f13f0e180d0818d50998R65-R78
With a single format string like `yyyy-MM-dd hh:mm:ss` I was able to obtain this:
image

So I would suggest to reconsider the implementation. I can create an issue if needed and provide a PR.

donho:
> So I would suggest to reconsider the implementation. I can create an issue if needed and provide a PR.

OK, go ahead - please provide both issue (a small specification of format) & PR.

mere-human:
> > So I would suggest to reconsider the implementation. I can create an issue if needed and provide a PR.
> 
> OK, go ahead - please provide both issue (a small specification of format) & PR.

@donho You can try it https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480
It is still work in progress.
The functionality works. But if you like it, we may need to change some UI text.

PR #10458, opened 2021-08-29
Document List Improvements: column size fix and new file path column

### LINKED ISSUE #4715: Column in document switcher reset size/position when resized
### Description of the Issue

Column in document switcher reset size/position when resized:

npp-issue

### Debug Information
```
Notepad++ v7.5.8   (64-bit)
Build time : Jul 23 2018 - 02:06:36
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll NPPJSONViewer-x64.dll PluginManager.dll DSpellCheck.dll mimeTools.dll NppConverter.dll 
```

### LINKED ISSUE #9679: [Feature Extension Request] Doc Switcher -- Additional column with full file path for sorting
### Description of the Issue
Currently in Doc Switcher are two columns available "Filename" and "Extension" -- so the list can be sorted by filename and by extension (great feature btw).
It would be really helpful, when there would be a third (optional) column with the full file path.
Then sorting would be possible by filename, filepath and extension.

### Expected Behavior
Sorting all files in Doc Switcher by full file path when clicking in header row of the new "path" column.

grafik

### LINKED ISSUE #10363: Problem with Document List Panel Extension column
### Description of the Issue
Document List Panel's Extension column is right justified and clipped, it makes it impossible to know an extension which is more than 2 characters long.

image

When the columns are resized to address the issue
image

All is good until you switch away from Npp (example minimize / maximize)
And it's back to its original clipped state.

1. Enable Document List panel and show the extensions 
2. Extension column is clipped
3. Resize column to fix it
4. Minimize Npp
5. Maximize Npp
6. Problem is back

### Expected Behavior
I expect the columns not to be clipped as they are by default (at least allow more characters than 2)
And if one resizes the columns, I expect them to be persistent until user resizes again, ie it survives restarts let alone minimize / maximize.

### Actual Behavior
See Description and Screenshots above

### Debug Information
Notepad++ v8.1.2   (32-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1151
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll DSpellCheck.dll Explorer.dll JSMinNPP.dll LocationNavigate.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll NPPJSONViewer.dll NppSaveAsAdmin.dll NppSnippets.dll NppTextFX.dll Tidy2.dll _CustomizeToolbar.dll

Ashfaaq18:
Fix #10363:
On npp startup, columns are auto sized to their longest row. All column items will be fully visible during startup, no clipping.

DocListResizeCol_Fix

Fix #4715:
If users change column size during runtime, the columns will not change back when minimizing ,tabbing or resizing.
(column size will revert when adding new files)

DocListResize1_Fix

Fix #9679
Added file path column to document list, can enable and disable from settings.

DocListPathCol_Fix


PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -48,7 +48,9 @@ BEGIN
     CONTROL         "Standard icons: small",IDC_RADIO_STANDARD,"Button",BS_AUTORADIOBUTTON,64,101,119,10
     GROUPBOX        "Document List Panel",IDC_DOCLIST_GB_STATIC,57,125,147,37,BS_CENTER
     CONTROL         "Disable extension column",IDC_CHECK_DOCLIST_NOEXTCOLUMN,
-                    "Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,64,140,124,10
+                    "Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,64,135,124,10
+    CONTROL         "Disable file path column",IDC_CHECK_DOCLIST_NOPATHCOLUMN,
```
mere-human, line 52: I would recommend avoiding negative commands. A better way could be "Enable..." instead of "Disable...". Usually it is easier to understand.

mere-human: Ideally, it would be nice to change `disable → enable` in previous option as well.
Usually, unnecessary inversions make it harder to understand.

Yaron10:
@Ashfaaq18,

Thank you for your great work. 👍 

Allow me some comments.

**1.** If the "Path" column is disabled, you get the following columns width (even after NPP restart):
תמונה

I think it would be more convenient and visually nicer if it looked like that (*current behavior*): 
תמונה

And if "Ext" is also disabled:
תמונה

IOW: take both the window and values width into account.

**2.** Some user might request to display "Path" only. I think it would make sense.

**3.** You did a great job moving "Toggle Document List" from Preferences to View.
The "Columns to display" settings should also be moved from Preferences to the "Document List" itself.
(Preferably - display a context-menu on right-clicking the columns headers).

**4.** is the dot after "Path" (`Path.`) intentional?

PowerEditor/src/WinControls/Preference/preference.rc
donho: > Ideally, it would be nice to change disable → enable in previous option as well.

It could be in the rule of the other application. However, it's not necessary to be the rule in Notepad++.
If the default value is ON, and people ask to make it optional, then it's usually "Disable XXX", and set it unchecked by default, of course.

donho:
* Single feature or bug-fix per PR.
* In short: The easier the code review is, the better the chance your pull request will get accepted.
ref:

Ashfaaq18:
@donho, looks like you've closed the other 2 bugs related to HiDPI (seems like they've already been fixed).

shall I redo a PR for only the path feature (take @Yaron10's points into consideration) ?

donho:
> shall I redo a PR for only the path feature (maybe add date modified in there as well

OK for the path column. However, is modified date column really necessary? I don't see any request for that.

>  and take @Yaron10's points into consideration) 

Yes. But please keep the PR as small as possible - you can do them separately from adding path column so the review will be easier to me.

Ashfaaq18:
> is modified date column really necessary?

Not really, Just thought of spicing it up a little.

> Yes. But please keep the PR as small as possible

yeah sure. Ill try keeping it to a minimum

Issue #10185, opened 2021-07-18
Nodepad++ compiled with mingw-w64 GCC 8.1 x64 is almost unusable


ivan-u7n:
### Description of the Issue
While there seems to be the working instruction for building Notepad++ with gcc, the result is almost unusable: no toolbar, no Plugin Admin, unsaved preferences.

### Steps to Reproduce the Issue
1. Compile Notepad++ with mingw-w64 GCC 8.1 using BUILD.md as a guide.
2. Run it

### Expected Behavior
A usable version of Notepad++.

### Actual Behavior
![notepadpp-gcc](https://user-images.githubusercontent.com/81974034/126069308-b06ebe3c-1406-4e82-ac6f-2f3394246fac.png)
Neither toolbar nor Plugin Admin to be seen. The opened file is the saved upon closing the application config.xml and is the reason why preferences aren't saved.
I event tried downloading the “Plugin Manager” plugin manually also with no success — it just doesn't show up anywhere.

### Debug Information
Notepad++ v8.1.1   (64-bit)
Build time : Jul 18 2021 - 14:09:15
Path : F:\develop\notepad-plus-plus-master\PowerEditor\bin\NotepadPP.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise LTSC 2019 (64-bit) 
OS Version : 1809
OS Build : 17763.2061
Current ANSI codepage : 1251
Plugins : none

mere-human:
What for do you need MinGW version?

ivan-u7n:
For a temporary replacement of the current version build with my patches applied to the bundled Scintilla (see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5264) as I have no desire to install full-blown Microsoft Visual Studio just for this.

I saw the build instructions and thought it will be ideal temporary solution, but alas Notepad++, while compiles with GCC and even starts, isn't usable as a replacement of the official build, thus the bug report. Otherwise there is no sense in making a piece of software buildable by a toolchain, but crippled in its functionality.

mere-human:
As far as I know, the MinGW build exists rather for verification purpose, something like "sanity check". Also, MinGW build could provide some useful compiler checks.
For example: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9403#issuecomment-766347210
I don't think it was designed to be a fully-functional product. But I might be wrong.
[...]

Issue #10328, opened 2021-08-04
[MinGW] Questions about appveyor builds


ozone10:
Both MinGW appveyor builds (release and debug) are same, any reason building it twice?

Any reason to build dynamic binary, static is somehow working, but dynamic will crash for me.

Should warning flags `-Wall -Wpedantic` be enabled, to check for issues with code?

donho:
@ozone10 
> Both MinGW appveyor builds (release and debug) are same, any reason building it twice?

What do you mean by `release and debug are same`? It follows the same schema of x86, x64 and ARM64, doesn't it?

ArkadiuszMichalski:
Flag was commented:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/appveyor.yml#L38

ivan-u7n:
> Any reason to build dynamic binary, static is somehow working, but dynamic will crash for me.

That's strange, for while I disable the build of the dynamic binary, I do it for speed of getting a binary after changing something in bundled Scintilla.

So I checked and it works for me (after copying required MinGW libraries into PowerEditor\bin):
```
Notepad++ v8.1.2   (64-bit)
Build time : Jul 26 2021 - 00:04:19
Path : F:\develop\notepad-plus-plus-master\PowerEditor\bin\NotepadPP_dynamic.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise LTSC 2019 (64-bit) 
OS Version : 1809
OS Build : 17763.2061
Current ANSI codepage : 1251
Plugins : DSpellCheck.dll mimeTools.dll nppConverter.dll NppExport.dll NppFTP.dll 
```

ozone10:
@donho 
as @ArkadiuszMichalski  said flag was commented so appveyor now builds "release" version twice

@ivan-u7n 
I think I have figured it out, you need to have right version mingw installed to get the correct version of dll files.
If you have no mingw installed, it will give you error with not able to find function in dll and if you have wrong version it can crash.

So I think dynamic build can be disabled.

ivan-u7n:
> I think I have figured it out, you need to have right version mingw installed to get the correct version of dll files.
> If you have no mingw installed, it will give you error with not able to find function in dll and if you have wrong version it can crash.

[...]

PR #10469, opened 2021-08-31
Issue 10328

chcg:
Fix 10328
Correct mingw debug build, added patch provided within #10328


ivan-u7n:
This will also partially fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10185 (the missing toolbar part).

PowerEditor/gcc/include/various.h
```diff
@@ -1,52 +1,52 @@
...
-// from MinGW include/driverspecs.h
-#define __inout
-#define __inout_opt
+//this file contains definitions not available in gcc 3.4.5,
+//but are needed for Notepad++
+//the makefile will automatically include this header file
+
+//GetLongPathName = 410
+//Multimonitor: 410
+//#define _WIN32_WINDOWS 0x0410
```
mere-human, line 7: It would be better just to remove the lines that aren't needed anymore (here and below) rather that just commenting.
This will reduce clutter.
The original code will still be available in Git history.

PR #10474, opened 2021-08-31
Update dutch.xml

RockyTDR:
Updated Dutch translations since 2021-08-31


donho:
@RockyTDR 
Thank you for the update.
Could you please check the latest difference between v8.1.4 & v8.1.5 on the following issue:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10548

And confirm me if your translation is all updated?
Thank you.

PR #10480, opened 2021-08-31
Enhance Insert Custom Date command

### LINKED ISSUE #10467: Insert Date Time Customized doesn't support some formats
For more context, see discussion here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10453

1. Settings > Preferences > Multi-Instance & Date
2. Custom Format

### Expected Behavior

A way to specify short year, etc.

### Actual Behavior

No way to specify short year (e.g. 98), leading zeroes in month and day (01), etc.

### Debug Information

Notepad++ v8.1.4   (64-bit)
Build time : Aug 30 2021 - 20:46:31
Path : ...\Notepad++.exe
Command Line : -multiInst -nosession
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1165
Current ANSI codepage : 1252
Plugins : none

mere-human:
Fix #10467

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1331,23 +1331,68 @@ int nbDigitsFromNbLines(size_t nbLines)
 
 generic_string getDateTimeStrFrom(const generic_string& dateTimeFormat, const SYSTEMTIME& st)
 {
-	generic_string dateTimeStr = dateTimeFormat;
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("Y"), std::to_wstring(st.wYear));
-	wchar_t buf[3];
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wMonth);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("M"), buf);
+	const TCHAR* localeName = LOCALE_NAME_USER_DEFAULT;
+	const DWORD flags = 0;
 
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wDay);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("D"), buf);
+	constexpr int bufferSize = MAX_PATH;
+	TCHAR buffer[bufferSize] = {};
+	int ret = 0;
 
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wHour);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("h"), buf);
+	// 1. Escape 'tt' that means AM/PM or 't' that means A/M.
+	// This is needed to avoid conflict with 'M' date format that stands for month.
+	generic_string newFormat = dateTimeFormat;
+	const TCHAR escapedChar = 0x1;
+	bool hasMidday = false;
+	for (auto& ch : newFormat)
+	{
+		if (ch == 't')
+		{
+			ch = escapedChar;
+			hasMidday = true;
+		}
+	}
 
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wMinute);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("m"), buf);
+	// 2. Format the AM/PM string.
+	generic_string midday = _T("tt");
+	if (hasMidday)
+	{
+		ret = GetTimeFormatEx(localeName, flags, &st, midday.c_str(), &midday[0], static_cast<int>(midday.size()));
+		if (ret == 0)
+			midday = _T("tt");
+	}
 
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wSecond);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("s"), buf);
+	// 3. Format the time (h/m/s/t/H).
+	ret = GetTimeFormatEx(localeName, flags, &st, newFormat.c_str(), buffer, bufferSize);
+	if (ret != 0)
+	{
+		// 4. Format the date (d/y/g/M). 
+		// Now use the buffer as a format string to process the format specifiers not recognized by GetTimeFormatEx().
+		ret = GetDateFormatEx(localeName, flags, &st, buffer, buffer, bufferSize, nullptr);
+	}
 
-	return dateTimeStr;
+	if (ret != 0)
+	{
+		// 5. Now format only the AM/PM string.
+		if (hasMidday && !midday.empty())
+		{
+			// 'ret' holds the actual length of the string
+			for (int i = 0; i < ret; ++i)
+			{
+				if (buffer[i] == escapedChar)
+				{
+					// 't' => 'A' or 'P'
+					buffer[i] = midday[0];
+					if (((i + 1) < ret) && (buffer[i + 1] == escapedChar))
+					{
+						// 'tt' => 'AM' or 'PM'
+						++i;
+						buffer[i] = midday[1];
+					}
+				}
+			}
+		}
+		return buffer;
+	}
+
+	return {};
 }
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1338,60 +1338,29 @@ generic_string getDateTimeStrFrom(const generic_string& dateTimeFormat, const SY
 	TCHAR buffer[bufferSize] = {};
 	int ret = 0;
 
-	// 1. Escape 'tt' that means AM/PM or 't' that means A/M.
-	// This is needed to avoid conflict with 'M' date format that stands for month.
+	// 1. Remove format specifiers that we don't support:
+	//    - 'tt' that means AM/PM or 't' that means A/M
+	//    - 'g', 'gg' that means B.C. or A.D.
+	// Note: If 'tt' is supported it will need to be properly escaped
+	// to avoid conflict with 'M' date format that stands for month.
 	generic_string newFormat = dateTimeFormat;
-	const TCHAR escapedChar = 0x1;
-	bool hasMidday = false;
-	for (auto& ch : newFormat)
-	{
-		if (ch == 't')
-		{
-			ch = escapedChar;
-			hasMidday = true;
-		}
-	}
+	auto it = std::remove_if(newFormat.begin(), newFormat.end(),
+		[](generic_string::value_type x) { return x == 't' || x == 'g'; });
+	newFormat.erase(it, newFormat.end());
 
-	// 2. Format the AM/PM string.
-	generic_string midday = _T("tt");
-	if (hasMidday)
-	{
-		ret = GetTimeFormatEx(localeName, flags, &st, midday.c_str(), &midday[0], static_cast<int>(midday.size()));
-		if (ret == 0)
-			midday = _T("tt");
-	}
-
-	// 3. Format the time (h/m/s/t/H).
+	// 2. Format the time (h/m/s/t/H).
 	ret = GetTimeFormatEx(localeName, flags, &st, newFormat.c_str(), buffer, bufferSize);
 	if (ret != 0)
 	{
-		// 4. Format the date (d/y/g/M). 
+		// 3. Format the date (d/y/g/M).
 		// Now use the buffer as a format string to process the format specifiers not recognized by GetTimeFormatEx().
 		ret = GetDateFormatEx(localeName, flags, &st, buffer, buffer, bufferSize, nullptr);
 	}
 
 	if (ret != 0)
 	{
-		// 5. Now format only the AM/PM string.
-		if (hasMidday && !midday.empty())
-		{
-			// 'ret' holds the actual length of the string
-			for (int i = 0; i < ret; ++i)
-			{
-				if (buffer[i] == escapedChar)
-				{
-					// 't' => 'A' or 'P'
-					buffer[i] = midday[0];
-					if (((i + 1) < ret) && (buffer[i + 1] == escapedChar))
-					{
-						// 'tt' => 'AM' or 'PM'
-						++i;
-						buffer[i] = midday[1];
-					}
-				}
-			}
-		}
-		return buffer;
+		// 'ret' holds the actual length of the string
+		return generic_string(buffer, buffer + ret);
 	}
 
 	return {};
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1329,6 +1329,75 @@ int nbDigitsFromNbLines(size_t nbLines)
 	return nbDigits;
 }
 
+namespace
+{
+	constexpr TCHAR timeFmtEscapeChar = 0x1;
+	constexpr TCHAR middayFormat[] = _T("tt");
+
+	// Returns AM/PM string defined by the system locale for the specified time.
+	// This string may be empty or customized.
+	generic_string getMiddayString(const TCHAR* localeName, const SYSTEMTIME& st)
+	{
+		generic_string midday;
+		// Obtain the length of the resulting string first.
+		int ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, nullptr, 0);
+		if (ret > 0)
+		{
+			midday.resize(ret);
+			ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, &midday[0], static_cast<int>(midday.size()));
+			if (ret != 0)
+				midday.resize(ret - 1); // Remove the null-terminator.
+		}
+		if (ret == 0 || midday[0] == '\0')
+			midday.clear();
+		return midday;
+	}
+
+	// Replaces conflicting time format specifiers by a special character.
+	bool escapeTimeFormat(generic_string& format)
+	{
+		bool modified = false;
+		for (auto& ch : format)
+		{
+			if (ch == middayFormat[0])
+			{
+				ch = timeFmtEscapeChar;
+				modified = true;
+			}
+		}
+		return modified;
+	}
+
+	// Replaces special time format characters by actual AM/PM string.
+	void unescapeTimeFormat(generic_string& format, const generic_string& midday)
+	{
+		if (midday.empty())
+		{
+			auto it = std::remove(format.begin(), format.end(), timeFmtEscapeChar);
+			if (it != format.end())
+				format.erase(it, format.end());
+		}
+		else
+		{
+			size_t i = 0;
+			while ((i = format.find(timeFmtEscapeChar, i)) != generic_string::npos)
+			{
+				if (i + 1 < format.size() && format[i + 1] == timeFmtEscapeChar)
+				{
+					// 'tt' => AM/PM
+					format.erase(i, std::size(middayFormat) - 1);
+					format.insert(i, midday);
+				}
+				else
+				{
+					// 't' => A/P
+					format[i] = midday[0];
+				}
+			}
+		}
+	}
+}
+
 generic_string getDateTimeStrFrom(const generic_string& dateTimeFormat, const SYSTEMTIME& st)
 {
 	const TCHAR* localeName = LOCALE_NAME_USER_DEFAULT;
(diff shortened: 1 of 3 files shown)


mere-human:
Example
* format: `hh:mm:ss tt; ddd, MMM dd yy g`
* output: `11:29:40 PM; Wed, Aug 31 94 A.D.`

mere-human:
For more info about format:
https://docs.microsoft.com/en-us/windows/win32/intl/day--month--year--and-era-format-pictures
https://docs.microsoft.com/en-us/windows/win32/api/datetimeapi/nf-datetimeapi-gettimeformatex#remarks

donho:
`tt` doesn't work for me:
image

I would say that we keep only the string format of API and don't support `tt` or `gg`.
I will accept only the simplest implementation.

mere-human:
> `tt` doesn't work for me:

What is your current locale?

donho:
French keyboard but with USA date time currency format.
Insert long date time in my Notepad++ gives me : `8:35 PM Wednesday, September 1, 2021`

rddim:
Bulgarian date time currency format, by default there is no `AM` `PM`

* If the `AM symbol` field is empty the whole text after `tt` is missing

npp_dt_no_am_pm

* If the `AM symbol` field contain `AM` the text after `tt` is available, but nothing for `tt`

npp_dt_am_no_pm

It is interesting that if `AM symbol` is empty, but the `PM symbol` contain `PM` the result is as the 1st screenshot. I don't know if I test it tommorow before 12:00h

mere-human:
> Bulgarian date time currency format, by default there is no `AM` `PM`

> French keyboard but with USA date time currency format.

Thank you for testing! I'll look into it.

donho:
>     * If the `AM symbol` field contain `AM` the text after `tt` is available, but nothing for `tt`

That's why I don't want a complex system.

PowerEditor/src/Parameters.h
```diff
@@ -875,7 +875,7 @@ struct NppGUI final
 	generic_string _uriSchemes = TEXT("svn:// cvs:// git:// imap:// irc:// irc6:// ircs:// ldap:// ldaps:// news: telnet:// gopher:// ssh:// sftp:// smb:// skype: snmp:// spotify: steam:// sms: slack:// chrome:// bitcoin:");
 	NewDocDefaultSettings _newDocDefaultSettings;
 
-	generic_string _dateTimeFormat = TEXT("Y-M-D h:m:s");
+	generic_string _dateTimeFormat = TEXT("yyyy-MM-dd HH:mm:ss");
```
mere-human, line 878: It would be nice if it was possible to update the value already saved to config XML.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1331,23 +1331,37 @@ int nbDigitsFromNbLines(size_t nbLines)
...
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("s"), buf);
+	if (ret != 0)
+	{
+		// 'ret' holds the actual length of the string
+		return generic_string(buffer, buffer + ret);
+	}
 
-	return dateTimeStr;
+	return {};
 }
```
donho, line 1367: Dosn't the following simpler code work? If it does for any case, no need to remove `tt` & `gg`. I just don't like the extra processing.
```
{
	const DWORD flags = 0;

	constexpr int bufferSize = MAX_PATH;

	GetDateFormatEx(localeName, flags, &st, dateTimeFormat.c_str(), buffer, bufferSize, nullptr);
	GetTimeFormatEx(localeName, flags, &st, buffer, buffer, bufferSize);

	return buffer;
}
```

donho: Hmm... I see why you process time before date - due to `ddd`

mere-human: Right.
Processing time before date almost works.
Except for `tt`. If it inserts AM then `M` is considered as month specifier for date format.
So the string like `hh tt yy` will be formatted as `11 A9 21` instead of `11 AM 21` for `2021.09.02 11:00`.

mere-human: I can return `tt` processing like before but leave it empty for case when it's not defined in the current locale (as on your system).

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1331,23 +1331,37 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+	int ret = 0;
+
+	// 1. Remove format specifiers that we don't support:
+	//    - 'tt' that means AM/PM or 't' that means A/M
+	//    - 'g', 'gg' that means B.C. or A.D.
+	// Note: If 'tt' is supported it will need to be properly escaped
+	// to avoid conflict with 'M' date format that stands for month.
+	generic_string newFormat = dateTimeFormat;
+	auto it = std::remove_if(newFormat.begin(), newFormat.end(),
+		[](generic_string::value_type x) { return x == 't' || x == 'g'; });
```
donho, line 1348: Removing AM & PM solve substitution of `M` problem. 
`g` can be restored without any consequence.

donho (changes requested):
Restore `g` & `gg` abiliity.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -392,8 +392,8 @@ BEGIN
     LTEXT           "* The modification of this setting needs to restart Notepad++",IDD_STATIC_RESTARTNOTE,110,70,239,20
     GROUPBOX        "Customize insert Date Time",IDC_DATETIMEFORMAT_GB_STATIC,90,97,268,105,BS_CENTER
     CONTROL         "Reverse default date time order (short && long formats)",IDD_DATETIMEFORMAT_REVERSEORDER_CHECK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,103,113,241,10
-    RTEXT           "Y-M-D h:m:s\nh:m:s D/M/Y\nM D, Y h:m",IDC_STATIC,134,142,77,25
-    LTEXT           "1985-10-26 01:24:00\n01:24:00 26/10/1985\n10 26, 1985 01:24",IDC_STATIC,234,142,94,25
+    RTEXT           "yyyy-MM-dd HH:mm:ss\nHH:mm:ss dd/MMM/yy\nMM dd, yyyy h:m",IDC_STATIC,134,142,77,25
```
donho, line 395: Please make string as "yyyy-MM-dd HH:mm:ss\nHH:mm:ss dd/MM/yy\nMMM d, yyyy h:m"

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -392,8 +392,8 @@ BEGIN
     LTEXT           "* The modification of this setting needs to restart Notepad++",IDD_STATIC_RESTARTNOTE,110,70,239,20
     GROUPBOX        "Customize insert Date Time",IDC_DATETIMEFORMAT_GB_STATIC,90,97,268,105,BS_CENTER
     CONTROL         "Reverse default date time order (short && long formats)",IDD_DATETIMEFORMAT_REVERSEORDER_CHECK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,103,113,241,10
-    RTEXT           "Y-M-D h:m:s\nh:m:s D/M/Y\nM D, Y h:m",IDC_STATIC,134,142,77,25
-    LTEXT           "1985-10-26 01:24:00\n01:24:00 26/10/1985\n10 26, 1985 01:24",IDC_STATIC,234,142,94,25
+    RTEXT           "yyyy-MM-dd HH:mm:ss\nHH:mm:ss dd/MMM/yy\nMM dd, yyyy h:m",IDC_STATIC,134,142,77,25
+    LTEXT           "1985-10-26 01:24:00\n01:24:00 26/Oct/85\n10 26, 1985 1:24",IDC_STATIC,234,142,94,25
```
donho, line 396: Please make string as "1985-10-26 01:24:00\n01:24:00 26/10/85\nOct 26, 1985 1:24"

donho (changes requested):
Modify the sample.

PowerEditor/src/MISC/Common/Common.cpp
donho: `tt` worked in my system if I use only GetTimeFormatEx without GetDateFormatEx.
Even with GetDateFormatEx called after, it still works - just M is replaced by a number (I didn't modify my settings).
So I guess there were some bug in your old commit.

mere-human: You can check this one:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/40594850
It should work for `tt`.

donho: No, it doesn't.

image

donho: Even it worked, algorithm would be too complex to go into master.

mere-human:
Will finish it in several days.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1331,23 +1331,37 @@ int nbDigitsFromNbLines(size_t nbLines)
...
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("m"), buf);
+	const TCHAR* localeName = LOCALE_NAME_USER_DEFAULT;
+	const DWORD flags = 0;
+
+	constexpr int bufferSize = MAX_PATH;
+	TCHAR buffer[bufferSize] = {};
+	int ret = 0;
+
+	// 1. Remove format specifiers that we don't support:
+	//    - 'tt' that means AM/PM or 't' that means A/M
```
MarkusBodensee, line 1342: Typo in comment:
't' means A/P and not A/M.

MarkusBodensee:
Since 't' is just replaced with 'A' or 'P', this case is not problematic for GetDateFormatEx(). So is it acceptable to allow 't', but not 'tt' ?

OTOH, the user can specify customized AM/PM specifiers in the Windows time settings. So there seems no easy solution for this scenario at all.

MarkusBodensee:
Possible solution as pseudo algorighm. Just a random thought as input for discussion, but not completely easy since there is no easy:

1. check if format string contains 't' or 'tt'
2. remember 't' or 'tt' and replace it with '$' or '$$' in format string
3. do GetTimeFormatEx(...) and GetDateFormatEx(...) with modified format string like done before
4. do GetTimeFormatEx(...) with the remembered 't' or 'tt' from 2.
5. finally replace '$' or '$$' in the result of 3. with result from 4.

mere-human:
@MarkusBodensee 

> Possible solution as pseudo algorighm. Just a random thought as input for discussion, but not completely easy since there is no easy

This was almost what I did as an initial version. You can take a look at it here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480/commits/8cef015f9916946aeccc0cc6df743995b63f7607

But it has some [problems](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480#issuecomment-910550535).
I'll try to improve the solution a bit later, when I have time. 
If @donho likes the new solution, we'll keep it. 
If not, then `tt` won't be allowed in the custom format.

donho:
> If @donho likes the new solution, we'll keep it.

What is the new solution?

donho:
BTW, I have provide a solution (to include `tt`) here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10467#issuecomment-911653446

PowerEditor/src/MISC/Common/Common.cpp
mere-human: Done.

PowerEditor/src/WinControls/Preference/preference.rc
mere-human: Done.

PowerEditor/src/WinControls/Preference/preference.rc
mere-human: Done.

mere-human:
> BTW, I have provided a solution to include `tt` here: [#10467 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10467#issuecomment-911653446)

This suggested solution may have different problems. Let's assume the current time is assume time is September 14, 2021, 21:00.
1. It requires one more GetDateFormatEx() call just to get the value "M" separately. This value might collide with other values of hours, minutes, etc.
2. What to do if a user specifies a format like "hh tM" and expects an output like "9 P9"? With the suggested solution it will be incorrectly formatted as "9 PM"
3. How to deal correctly with the cases when "M" produces a number consisting of several digits "12"?
4. Other edge cases that I haven't found yet

mere-human:
So I provided an upgraded version of my very first solution. Now it works fine for various cases like this:
Screenshot 2021-09-15 000440
Screenshot 2021-09-15 000516
Screenshot 2021-09-15 000953

If there are some problems, I'm ready to provide a simple solution that doesn't handle `tt` format.

donho:
@mere-human 
I still prefer [my solution](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10549/files) - looks much simpler. :)

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1329,25 +1329,112 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+	generic_string getMiddayString(const TCHAR* localeName, const SYSTEMTIME& st)
+	{
+		generic_string midday;
+		// Obtain the length of the resulting string first.
+		int ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, nullptr, 0);
+		if (ret > 0)
+		{
+			midday.resize(ret);
+			ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, &midday[0], static_cast<int>(midday.size()));
+			if (ret != 0)
```
MarkusBodensee, line 1348: Better to check `if (ret > 0)` here, too?

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1329,25 +1329,112 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+	const DWORD flags = 0;
+
+	constexpr int bufferSize = MAX_PATH;
+	TCHAR buffer[bufferSize] = {};
+	int ret = 0;
 
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wDay);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("D"), buf);
+	// 1. Get the AM/PM string.
+	const generic_string midday = getMiddayString(localeName, st);
```
MarkusBodensee, line 1411: Step 1 can be moved inside of `if (hasMiddayFormat)` body to remove overhead. `midday` is only needed if there is something to replace.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1329,25 +1329,112 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+		// Now use the buffer as a format string to process the format specifiers not recognized by GetTimeFormatEx().
+		ret = GetDateFormatEx(localeName, flags, &st, buffer, buffer, bufferSize, nullptr);
+	}
 
-	_snwprintf(buf, sizeof(buf), TEXT("%02d"), st.wSecond);
-	dateTimeStr = stringReplace(dateTimeStr, TEXT("s"), buf);
+	if (ret != 0)
+	{
+		if (hasMiddayFormat)
+		{
```
MarkusBodensee, line 1430: Move current step 1 to here.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1329,25 +1329,112 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+			// 5. Now format only the AM/PM string.
+			generic_string result = buffer;
+			unescapeTimeFormat(result, midday);
```
MarkusBodensee, line 1433: Why copying `buffer` to `result`? Why not pass `buffer` to `unescapeTimeFormat()`? And `return` statement for `buffer` is already there, so `return result` is not needed.

MarkusBodensee (commented):
I left a few thoughts. I like this approach much, because it handles all the edge cases I had in mind. And especially the customized by user formats, too.

mere-human:
> But I still prefer [my solution](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10549/files) - way much simple. :)

1. It doesn't handle `t` format.
image

2. AM/PM string can be customized via Settings > Region > Additional date, time & regional settings > Change date, time or number formats. Your solution doesn't handle it.
image
It'll look like this:
image

3. Other collision problems. Consider date 1985-10-10 10:10.
image
Now let's hack it:
image

I can continue searching bugs🙂

Considering the facts above, I think the current solution works better.
I know, the algorithm isn't very simple. But this provides the maximum functionality to the users.
I commented it well and split the code into functions so that it is easier to understand.

We still can go back to a simple solution without `tt` support but that would be dull.

PowerEditor/src/MISC/Common/Common.cpp
mere-human: `GetTimeFormatEx` returns 0 (in case of error) or the length of the resulting string. 
The error is unlikely here because we check the length first at line 1343. But I check it nevertheless. 
Empty AM/PM string is handled here as well (it will result in `midday.resize(0)`). 
`std::string` works not very well in some cases if it contains `'\0'`.

PowerEditor/src/MISC/Common/Common.cpp
mere-human: `unescapeTimeFormat` works with a reference. I need to copy buffer to a string and then modify it.

mere-human: If I understood you correctly, it would look like this:
```
generic_string unescapeTimeFormat(const generic_string& buffer, const generic_string& midday)
{
  generic_string result = buffer;
  // modify result...
  return result;
}
```
But here we're creating an extra copy of buffer which I wanted to avoid.

MarkusBodensee: You are correct, sorry, I messed it up.

PowerEditor/src/MISC/Common/Common.cpp
MarkusBodensee: Yes, both `GetTimeFormatEx` and `GetDateFormatEx` return 0 in error case and should never ever return negative values. To be on the safest side one can be, a check for `if (ret > 0)` after every call of the methods would even cover this. But it's just a minor remark.

Another question. Can it be avoided to call `GetTimeFormatEx` two times, because it looks like overhead? Maybe like this, or is there a downside?

```
constexpr int size = MAX_PATH;
generic_string midday;
midday.resize(size);

int ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, &midday[0], size);
if (ret > 0)
{
    midday.resize(ret - 1); // Remove the null-terminator.
}
else
{
    midday.resize(0); // Clear.
}

return midday;
```

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1329,25 +1329,112 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+	{
+		generic_string midday;
+		// Obtain the length of the resulting string first.
+		int ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, nullptr, 0);
+		if (ret > 0)
+		{
+			midday.resize(ret);
+			ret = GetTimeFormatEx(localeName, 0, &st, middayFormat, &midday[0], static_cast<int>(midday.size()));
+			if (ret != 0)
+				midday.resize(ret - 1); // Remove the null-terminator.
```
donho, line 1349: Why bother to remove null terminator?

PowerEditor/src/MISC/Common/Common.cpp
donho, line 1411: whereas midday is generic_string ?

PowerEditor/src/MISC/Common/Common.cpp
mere-human: Because string algorithms such as `std::string::erase` and `std::string::insert` will lead to incorrect results otherwise.
It had actually happened to me until I figured out.

PowerEditor/src/MISC/Common/Common.cpp
mere-human: ?

mere-human: `midday` can be empty, or AM/PM or custom value specified by a user in Settings (see my previous comments in this PR).

ivan-u7n:
```c++
{
	const DWORD flags = 0;

	size_t tPos = 0;
	size_t tLength = std::string::npos;
	{
		if (tLength == std::string::npos)
		{
			GetTimeFormatEx(localeName, flags, &st, L"t", buffer, bufferSize);
			tLength = wcslen(buffer);
		}
		size_t tEnd = tPos + 1;
		while (tEnd < formatEx.length() && formatEx[tEnd] == L't') ++tEnd;
		if (tLength)
		{
			tPos += tEnd + 8;
		}
		else
		{
			formatEx.erase(tPos, tEnd - tPos);
		}
	}

	formatEx = buffer;

	return buffer;
}
```
image

Although it doesn't handle user specified escaping it does seem much simpler…

donho:
@mere-human 
Is it a right result `2021-09-15 16:21:46 PM` to you ?

mere-human:
> @mere-human
> Is it a right result `2021-09-15 16:21:46 PM` to you ?

@donho Seems legit.

PowerEditor/src/MISC/Common/Common.cpp
mere-human: Done.

PowerEditor/src/MISC/Common/Common.cpp
mere-human: Done.

PR #10501, opened 2021-09-04
Fix CSS pseudo-elements not recognized issue

### LINKED ISSUE #10425: CSS pseudo-elements not recognized
Notepad++ doesn’t recognize CSS pseudo-elements in the form `::pseudo-element`, unlike `:pseudo-element`. SciTE doesn’t have this problem.

donho:
Update CSS keywords and add new styles.

Fix #10425


PowerEditor/src/stylers.model.xml
```diff
@@ -342,6 +342,17 @@
             <WordsStyle name="ID" styleID="10" fgColor="0080FF" bgColor="FFFFFF" fontName="" fontStyle="1" fontSize="" />
             <WordsStyle name="IMPORTANT" styleID="11" fgColor="FF0000" bgColor="FFFFFF" fontName="" fontStyle="1" fontSize="" />
             <WordsStyle name="DIRECTIVE" styleID="12" fgColor="0080FF" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="DOUBLESTRING" styleID="13" fgColor="808080" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="SINGLESTRING" styleID="14" fgColor="808080" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="IDENTIFIER2" styleID="15" fgColor="0040E0" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
```
ArkadiuszMichalski, line 347: @donho 
IDENTIFIER2, IDENTIFIER3, EXTENDED_IDENTIFIER, EXTENDED_PSEUDOCLASS, EXTENDED_PSEUDOELEMENT will not work without coresponding list for them:  LIST_2,  LIST_3, LIST_5,  LIST_6,,  LIST_7 setting by `setLexer()` as above.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/lexers/LexCSS.cxx#L75

Could be listed all of them, then it will be easier to add something in the `langs.model.xml` itself, without changing `ScintillaEditView.h`.

donho: For CSS2 & CSS3 (IDENTIFIER2 & IDENTIFIER3 respectively) are in current existing list, so I don't think we will add (separate) them in the future. Anyway, I will comment all not used keyword lists for the eventual future added.
Thank you for your review.

ArkadiuszMichalski:
@donho 
What about with this old notation `:` for pseudoelement? Should be recognized as pseudoclass because `after before first-letter first-line` are still listed in pseudoclasses (`instre2`)?

donho:
@ArkadiuszMichalski 
I removed `after` from `instre2` (by keeping `after` in `type3`), here's the result:

image

ArkadiuszMichalski:
And it's correct, because `:after` is a pseudo-element (in old notation) and shouldn't use the pseudo-class colors, now it just has a color from `UNKNOWN_PSEUDOCLASS`. New notation `::after` use `PSEUDOELEMENT` style.

Like I wrote here https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10425#issuecomment-912715474, there is a trick to handle `:after` with the same colors as `::after`, without changing the lexer. An additional list `EXTENDED_PSEUDOCLASS` (`LIST_6`) can be used.

```
setLexer(SCLEX_CSS, L_CSS, LIST_0 | LIST_1 | LIST_4 | LIST_6);

<Keywords name="type5">after before first-letter first-line</Keywords>

<WordsStyle name="PSEUDOELEMENT" styleID="18" fgColor="FF8000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
<WordsStyle name="EXTENDED_PSEUDOCLASS" styleID="20" fgColor="FF8000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />

```

Now bot `:after` and `::after` will use same style. Unfortunately, the style configurator will show `EXTENDED_PSEUDOCLASS`, which can be confusing for others (because it will cover these 4 pseudoelement cases and not everyone has to know about it).

donho:
You have mentioned in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10425#issuecomment-912656203
> @donho
:after is also pseudoelement (notation with : comes from old specs). In new specs only :: should be used for pseudoelement (browsers still support old notation for compatibility reasons).

So I suppose :after should be recognized. Now my question is: Is `:after` (and the other `:*`) valid in the most of browsers?

ArkadiuszMichalski:
>So I suppose :after should be recognized. 

It would be most appropriate, but the lexer does not understand it now (it can only consider them as pseudo-classes or use abve trick).

>Now my question is: if :after (and the other :*) is (are) valid in the most of browsers, should we make it valid as well?

Yes, for sure the 4 old ones are valid in the browser (here nobody wants to break backwards compatibility), I did not check the rest (because it is possible that all pseudo-elements with ":" are supported in browsers). I'll check it out later.

> should we make it valid as well?

It would be most appropriate, then the question is how? Some options:

- recognize them as pseudoclass
- not recognize so treat as UNKNOWN_PSEUDOCLASS
- use EXTENDED_PSEUDOCLASS list for them and use same style as for pseudoelement
- open issue in Lexilla to recognize this old notation

donho:
OK, I'll do the following:
* `after before first-letter first-line` will added into LIST_6 (EXTENDED_PSEUDOCLASS 20)
* `marker` will be added to the <Keywords name="type3"> 

Which keywords should I remove from PsudoClass besides of `after before first-letter first-line` ?

ArkadiuszMichalski:
>Which keywords should I remove from PsudoClass besides of after before first-letter first-line ?

also `selection`

donho:
BTW, since `:after :before :first-letter :first-line` are psudo-element, why don't we add them rather into EXTENDED_PSEUDOELEMENT (21) ?

ArkadiuszMichalski:
`PSEUDOELEMENT` and `EXTENDED_PSEUDOELEMENT (21)` and thieir coresponding list recognize only "::" notation, we need ":" so `EXTENDED_PSEUDOCLASS` may be the solution here (we don't use this list anyway).

ArkadiuszMichalski:
@donho 

But we can do this:
```
<WordsStyle name="LEGACY_PSEUDOELEMENT" styleID="20" fgColor="FF8000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" keywordClass="type5" /> /* Oryginal name is EXTENDED_PSEUDOCLASS */
``` 

We can use any `name=""` so in Style Configurator won't be any confusing.

image

ArkadiuszMichalski (changes requested):
Check last comment https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10501#issuecomment-913016155.

PowerEditor/installer/themes/DarkModeDefault.xml
```diff
@@ -338,6 +338,17 @@ License:             GPL2
...
+            <WordsStyle name="DOUBLESTRING" styleID="13" fgColor="808080" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="SINGLESTRING" styleID="14" fgColor="808080" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" />
+            <!--WordsStyle name="IDENTIFIER2" styleID="15" fgColor="0040E0" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" /-->
+            <WordsStyle name="ATTRIBUTE" styleID="16" fgColor="DFDFDF" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" />
+            <!--WordsStyle name="IDENTIFIER3" styleID="17" fgColor="00A0E0" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" /-->
+            <WordsStyle name="PSEUDOELEMENT" styleID="18" fgColor="CEDF99" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize=""  keywordClass="type3"/>
+            <!--WordsStyle name="EXTENDED_IDENTIFIER" styleID="19" fgColor="7F7F00" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" /-->
+            <WordsStyle name="EXTENDED_PSEUDOCLASS" styleID="20" fgColor="CEDF99" bgColor="3F3F3F" fontName="" fontStyle="0" fontSize="" keywordClass="type5" />
```
ArkadiuszMichalski, line 348: Make analogous correct as in `stylers.model.xml`.

PowerEditor/src/langs.model.xml
```diff
@@ -119,8 +119,10 @@
...
-            <Keywords name="instre2">active after before check checked disabled empty enabled first first-child first-letter first-line first-of-type focus hover indeterminate invalid lang last-child last-of-type left link not nth-child nth-last-child nth-of-type nth-last-of-type only-child only-of-type optional read-only read-write required right root selection target valid visited</Keywords>
-        </Language>
+            <Keywords name="instre2">active check checked disabled empty enabled first first-child first-of-type focus hover indeterminate invalid lang last-child last-of-type left link not nth-child nth-last-child nth-of-type nth-last-of-type only-child only-of-type optional read-only read-write required right root target valid visited</Keywords>
+            <Keywords name="type3">after before first-letter first-line selection cue file-selector-button placeholder slotted part</Keywords>
```
ArkadiuszMichalski, line 123: Add `marker` here.

ArkadiuszMichalski (changes requested):

ArkadiuszMichalski (approved):

PR #10503, opened 2021-09-04
Fix gcc build: xml writing, .gitignore

### LINKED ISSUE #9163: [MinGW] After closing Notepad++ some xml config files will be broken
### Description of the Issue
After closing Notepad++ some xml config files will be broken. Just check `config.xml` or `session.xml`. Without a working configuration, some things cannot be tested.

image

### Debug Information
Notepad++ v7.9.1   (64-bit)
Build time : Nov 15 2020 - 18:37:13
Path : C:\Users\Arek\Desktop\npp.7.9.1.portable.x64\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Build : 7601.24136
Current ANSI codepage : 1250

ivan-u7n:
As a non-unicode build is no longer possible, wide-char specifiers can be used.

Fixes #9163 and the second part of #10185.

### DIFF
--- a/PowerEditor/src/TinyXml/tinyxml.cpp
+++ b/PowerEditor/src/TinyXml/tinyxml.cpp
@@ -559,7 +559,7 @@ void TiXmlElement::Print( FILE* cfile, int depth ) const
 		generic_fprintf( cfile, TEXT("    ") );
 	}
 
-	generic_fprintf( cfile, TEXT("<%s"), value.c_str() );
+	generic_fprintf( cfile, TEXT("<%ls"), value.c_str() );
 
 	TiXmlAttribute* attrib;
 	for ( attrib = attributeSet.First(); attrib; attrib = attrib->Next() )
@@ -581,7 +581,7 @@ void TiXmlElement::Print( FILE* cfile, int depth ) const
 	{
 		generic_fprintf( cfile, TEXT(">") );
 		firstChild->Print( cfile, depth + 1 );
-		generic_fprintf( cfile, TEXT("</%s>"), value.c_str() );
+		generic_fprintf( cfile, TEXT("</%ls>"), value.c_str() );
 	}
 	else
 	{
@@ -598,7 +598,7 @@ void TiXmlElement::Print( FILE* cfile, int depth ) const
 		generic_fprintf( cfile, TEXT("\n") );
 		for( i=0; i<depth; ++i )
 		generic_fprintf( cfile, TEXT("    ") );
-		generic_fprintf( cfile, TEXT("</%s>"), value.c_str() );
+		generic_fprintf( cfile, TEXT("</%ls>"), value.c_str() );
 	}
 }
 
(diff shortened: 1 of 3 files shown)


PowerEditor/src/TinyXml/tinyxml.cpp
```diff
@@ -1025,7 +1025,7 @@ void TiXmlUnknown::Print( FILE* cfile, int depth ) const
 {
 	for ( int i=0; i<depth; i++ )
 		generic_fprintf( cfile, TEXT("    ") );
-	generic_fprintf( cfile, TEXT("%s"), value.c_str() );
+	generic_fprintf( cfile, TEXT("%ls"), value.c_str() );
```
donho, line 1028: I am wondering if `%ls` work for `fwprintf`.

My doubt comes from the following link:
https://stackoverflow.com/questions/26816547/whats-the-difference-between-printfs-printfls-wprintfs-and-wp

donho:
> Should fix #9163 and the second part of #10185.

What's the 2nd part of #10185?

PowerEditor/src/TinyXml/tinyxml.cpp
ivan-u7n: Yes, it works in both `fprintf` and `fwprintf` to mean a `wchar_t*` argument for both GCC and MSC.

ArkadiuszMichalski:
I downloaded [Notepad++.mingw.Release.exe](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/40645210/job/k079mkvivafxaqtw/artifacts) and the xml files (like `config.xml` or `session.xml`) are no longer corrupted.

ivan-u7n:
> What's the 2nd part of #10185? The config.xml shown as following?
> https://user-images.githubusercontent.com/81974034/126069308-b06ebe3c-1406-4e82-ac6f-2f3394246fac.png

Yes, the first part was the missing toolbar due to an absent manifest.

> **"Should"** is an ambiguous word - have you tested without and with this PR to make sure it's fixed?

Sorry for the incorrect wording.

I tested the PR with GCC and all xml-configs are now written correctly, moreover I use a variant of this PR to build my locally used version of Notepad++.

After AppVeyor build was finished, I downloaded the MSC x64 version from https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/40645210 to check for any regressions with MSC and found none.

I'll change the text of this PR.

PowerEditor/src/TinyXml/tinyxml.cpp
ozone10: @donho 
`The various printf and wprintf functions take a format string and optional arguments and produce a formatted sequence of characters for output. `
from https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-160

and list of function using these format specification: https://docs.microsoft.com/en-us/cpp/c-runtime-library/stream-i-o?view=msvc-160

And this is why GCC has issue with `%s`, but MSC not
`An ls, lS, ws, or wS type specifier is synonymous with S in printf functions and with s in wprintf functions.`
from https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-160#size-prefixes-for-printf-and-wprintf-format-type-specifiers

PR #10507, opened 2021-09-04
gcc build: 32-bit binary, remove -fpermissive, auto -j in make

### LINKED ISSUE #9989: Fix or update instruction for MinGW 32-bit
[This sentence](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/BUILD.md#build-64-bits-binaries-with-gcc) looks insufficient:
> Note 2: For 32-bit build, https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/sjlj/i686-8.1.0-release-posix-sjlj-rt_v6-rev0.7z could be used. The rest of the instructions are still valid.

I'm trying to compile the current code with 32-bit tools but without success. Scintilla passed but Notepad++ not :
```
../src/darkmode/DarkMode.cpp: In function 'void FixDarkScrollBar()':
../src/darkmode/DarkMode.cpp:227:99: error: invalid static_cast from type 'FixDarkScrollBar()::<lambda(HWND, LPCWSTR)>' to type 'fnOpenNcThemeData' {aka 'void* (__attribute__((stdcall)) *)(HWND__*, const wchar_t*)'}
     addr->u1.Function = reinterpret_cast<ULONG_PTR>(static_cast<fnOpenNcThemeData>(MyOpenThemeData));
                                                                                                   ^
mingw32-make.exe: *** [makefile:284: ../src/darkmode/DarkMode.o] Error 1
mingw32-make.exe: *** Waiting for unfinished jobs....
```

Compiling with MinGW 64-bit works fine.

ivan-u7n:
* Building 32-bit binary with GCC is now possible.
* The warning requiring the `-fpermissive` flag was fixed by explicit cast.
* Simple make calls now apply the `-j` flag to submakes when applicable.
* The GCC part of /BUILD.md was updated and additional info about building 32-bit binaries was added.
* The 32-bit binary build was added to /appveyor.yml, as a result the file got better structured.

Fixes #9989.

EDIT: Debug Info:
```
Notepad++ v8.1.4   (32-bit)
Build time : Sep  4 2021 - 22:18:15
Built with : GCC 8.1.0
Path : F:\develop\git\notepad-plus-plus\PowerEditor\bin\NotepadPP-release.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise LTSC 2019 (64-bit) 
OS Version : 1809
OS Build : 17763.2114
Current ANSI codepage : 1251
Plugins : none
```


donho:
Is it possible to build x86 & x64 release binaries into `bin` & `bin64` respectively ?

ivan-u7n:
> Is it possible to build x86 & x64 release binaries into `bin` & `bin64` folder respectively ?

I'm afraid without either some wild guesswork or crippling simple `mingw32-make` invocation it's impossible.

The `make` program doesn't care neither about the compiler it calls nor about the architecture of binaries it produces, so it's impossible to distinguish into which directory to put the resulting binary. Moreover the output filename of Scintilla's makefile is the same no matter the options. So I need to `mingw32-make clean` every time I change building options (e.g. the release/debug configuration, the architecture, and so on).

The `target` option was added into `PowerEditor/gcc/makefile` only because I put the same environment variable into `appveyor.yml`. But as I mentioned in `BUILD.md`, a 32-bit MinGW installation can be fixed permanently by 3 simple commands and won't require `target` option anymore.

Summing up all the above I'm more inclined to generate a resulting binary with the constant name in the build directory (or a sibling one to it). And the name of that directory will contain the target triplet and the configuration. The target triplet is then can default to 64-bit version and needs to be specified for 32-bit build. The minimum required XML-files can be copied by makefile into this directory for Notepad++ to start properly.

ArkadiuszMichalski:
> The make program doesn't care neither about the compiler it calls nor about the architecture of binaries it produces, so it's impossible to distinguish into which directory to put the resulting binary. 

But from the makefile we can check if 32-bit or 64-bit GCC will be used and set the appropriate directories depending on it?

donho:
@ivan-u7n 
> But from the makefile we can check if 32-bit or 64-bit GCC will be used and set the appropriate directories depending on it?

As @ArkadiuszMichalski has mentioned above, there must be a way to detect x86/x64 mode of GCC?

BUILD.md
```diff
@@ -41,33 +40,46 @@ More about the previous build process: https://community.notepad-plus-plus.org/t
...
 
-Note 1: if you use MinGW from the package (7z), you need manually add the MinGW/bin folder path to system Path variable to make mingw32-make.exe invoke works (or you can use command :`set PATH=%PATH%;C:\xxxx\mingw64\bin` for adding it on each time you launch cmd).
+```
+gendef %windir%\System32\SensApi.dll
+dlltool -d SensApi.def -k -l libsensapi.a
+del SensApi.def
+```
 
-Note 2: For 32-bit build, https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/sjlj/i686-8.1.0-release-posix-sjlj-rt_v6-rev0.7z could be used. The rest of the instructions are still valid.
+The resulting `libsensapi.a` then can be moved from the current directory to `$MinGW-root$\i686-w64-mingw32\lib`.
```
donho, line 85: > The missing linking library can also be generated manually to fix a MinGW installation permanently by executing the following commands via cmd in a writable directory with $MinGW-root$\bin added to PATH:
> 
> gendef %windir%\System32\SensApi.dll
dlltool -d SensApi.def -k -l libsensapi.a
del SensApi.def

However, the result for me is:

```
C:/appz/mingw/mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lsensapi
collect2.exe: error: ld returned 1 exit status
mingw32-make.exe[1]: *** [makefile:141: ../bin/NotepadPP-release.exe] Error 1
mingw32-make.exe: *** [makefile:106: all] Error 2

C:\dev\notepad-plus-plus\PowerEditor\gcc>gendef %windir%\System32\SensApi.dll
 * [C:\Windows\System32\SensApi.dll] Found PE image

C:\dev\notepad-plus-plus\PowerEditor\gcc>dlltool -d SensApi.def -k -l libsensapi.a

C:\dev\notepad-plus-plus\PowerEditor\gcc>del SensApi.def

C:\dev\notepad-plus-plus\PowerEditor\gcc>mingw32-make.exe
mingw32-make.exe[1]: '../bin/libscilexer.a' is up to date.
linking ../bin/NotepadPP-release.exe
C:/appz/mingw/mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lsensapi
collect2.exe: error: ld returned 1 exit status
mingw32-make.exe[1]: *** [makefile:141: ../bin/NotepadPP-release.exe] Error 1
mingw32-make.exe: *** [makefile:106: all] Error 2

C:\dev\notepad-plus-plus\PowerEditor\gcc>
```
Are there some instructions missing? BTW, what's the `SensApi.dll` ? 
I would say that we keep only MinGW 64 bit since it's more modern and it works better without any extra support.

donho (changes requested):
1. Remove 32-bit binary instructions from `build.MD`
2. Make 64-bit binary in the folder of `bin64`

ivan-u7n:
@donho 
> As @ArkadiuszMichalski has mentioned above, there must be a way to detect x86/x64 mode of GCC?

That's what I called "wild guesswork". I didn't want to go that way, because there exist multilib builds of GCC which can produce both 64- and 32-bit binaries. And they most certainly will be detected as x64. However it seems I ought to, I'll look into it.

> Are there some instructions missing? BTW, what's the SensApi.dll ?

In the first case you missed adding `target=i686` to `mingw-make` call. In the second case you missed the moving part of "The resulting `libsensapi.a` then can be moved from the current directory to `$MinGW-root$\i686-w64-mingw32\lib`".

I think I'll make some sort of detection not to require adding `target=i686` to `mingw32-make` invocation and bailing out if the DLL in question cannot be found in case of running not on Windows.

This DLL provides two functions for detecting internet-connection availability.

> I would say that we keep only MinGW 64 bit since it's more modern and it works better without any extra support.

Still, 32-bit build is good to have for the sanity check at least, in the current state it "just works" at least at AppVeyor.

> 2. Make 64-bit binary in the folder of bin64

May I propose the opposite, as the 64-binary seems to be the primary one: `bin` for 64-bit binary and `bin32` for the 32-bit one? Or maybe even something like `bin-i686` and `bin-x86_64`?

ArkadiuszMichalski:
@ivan-u7n  
>That's what I called "wild guesswork". I didn't want to go that way, because there exist multilib builds of GCC which can produce both 64- and 32-bit binaries. And they most certainly will be detected as x64. However it seems I ought to, I'll look into it.

But Notepad++ is build/tested on a specific version of MinGW-w64 as the build instruction says. Does it make sense to worry about every available tools? It's good if it will work universally, but nobody will check it all anyway, the priority should be a specific MinGW-w64. If someone uses a different tool, then he can correct the makefiles for it.

@donho 
>Remove 32-bit binary instructions from build.MD

Why, due to old 32-bit plugins this build is sometimes necessary. And that nicely explains how to deal with it. As a last resort, this can be described directly one the issue, FAQ or forum.  

Can't AppVeyor only check 64-bit by default, but keep 32-bit option (disabled by default by some flag)? Then from time to time it will also be possible to check 32-bit and fix it (if necessary) also through AppVeyor. Two additional builds permanently on (debug and normal) for 32-bit make the process a bit longer. I see that MinGW isn't a priority anyway, so one build type should be enough.

ivan-u7n:
@ArkadiuszMichalski 
> But Notepad++ is build/tested on a specific version of MinGW-w64 as the build instruction says. Does it make sense to worry about every available tools? It's good if it will work universally, but nobody will check it all anyway, the priority should be a specific MinGW-w64. If someone uses a different tool, then he can correct the makefiles for it.

If that's the case, then it'll simplify the task.

I also want to try extracting the debug information into separate files to shorten the compile time.

donho:
@ivan-u7n 
> In the first case you missed adding target=i686 to mingw-make call. 

Yes, my bad. Adding `target=i686` makes the build go through the end.
I still don't understand why mingw build 32 release needs to link to `SensApi.dll` - is this dll part of MS Windows system ?
Can we remove `SensApi.dll` API part and provide the command line explicitly:
```
mingw32-make target=i686
```

My goal is make this build instruction as simple as possible and prevent users from confusing.

BUILD.md
```diff
@@ -41,33 +40,46 @@ More about the previous build process: https://community.notepad-plus-plus.org/t
...
+# Building 64-bit binaries with GCC
 
-If you have installed [MinGW-w64](https://mingw-w64.org/doku.php/start), then you can compile Notepad++ & libscilexer.a 64 bits binaries with GCC.
+If you have [MinGW-w64](https://mingw-w64.org/doku.php/start) installed, then you can compile Notepad++ and optionally libscilexer.a 64-bit binaries with GCC.
 
-* Compile libscilexer.a
+You can download MinGW-w64 from https://sourceforge.net/projects/mingw-w64/files/. Notepad++ uses [MinGW 8.1 x86_64-8.1.0-release-posix-seh-rt_v6-rev0](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z).
 
-This step is not necessary anymore as it will be build in the process of compiling the Notepad++ binary. It is here just for completeness sake as this option is still available.
+## Compile libscilexer.a (optional)
```
donho, line 49: Remove the part `## Compile libscilexer.a (optional)` and keep only the part `## Compile Notepad++ binary`
Users (most of them) don't care about libscilexer part. All they want is to build Notepad++

BUILD.md
```diff
@@ -41,33 +40,46 @@ More about the previous build process: https://community.notepad-plus-plus.org/t
...
+This step is not necessary anymore as the library will be build in the process of compiling the Notepad++ binary. It is here just for completeness sake as this option is still available.
+
+1. Launch `cmd`.
 2. Change dir into `notepad-plus-plus\scintilla\win32`.
-3. Type `mingw32-make.exe -j%NUMBER_OF_PROCESSORS%`
+3. Type `mingw32-make -j%NUMBER_OF_PROCESSORS%`
 4. `libscilexer.a` is generated in `notepad-plus-plus\scintilla\bin\`.
 
-* Compile Notepad++ binary
+## Compile Notepad++ binary
```
donho, line 58: Rather `## Compile Notepad++ x64 binary`

BUILD.md
```diff
@@ -41,33 +40,46 @@ More about the previous build process: https://community.notepad-plus-plus.org/t
...
 
-To have a debug build just add `DEBUG=1` to the `mingw32-make.exe` invocation. The binary will be called `NotepadPP-debug.exe` in this case.
+To have a debug build just add `DEBUG=1` to the `mingw32-make` invocation. The binary will be called `NotepadPP-debug.exe` in this case.
 
 To see commands being executed add `VERBOSE=1` to the same command.
 
-You can download MinGW-w64 from https://sourceforge.net/projects/mingw-w64/files/. On Notepad++ Github page (this project), the build system use [MinGW 8.1](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z).
+**Note:** if you use MinGW from a package (7z), you need to manually add the `$MinGW-root$\bin` directory to the system `PATH` environment variable for `mingw32-make` invocation to work (one can use a command like `set PATH=%PATH%;$MinGW-root$\bin` each time `cmd` is launched).
+
+# Building 32-bit binaries with GCC
```
donho, line 71: Rather `## Compile Notepad++ x86 binary`

BUILD.md
```diff
@@ -41,33 +40,46 @@ More about the previous build process: https://community.notepad-plus-plus.org/t
...
+
+Building a 32-bit binary of Notepad++ is tested with [MinGW 8.1 i686-8.1.0-release-posix-dwarf-rt_v6-rev0](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/dwarf/i686-8.1.0-release-posix-dwarf-rt_v6-rev0.7z).
+
```
donho, line 74: Keep the download part above, so users know where to download the compiler.

BUILD.md
donho, line 85: Replace above info by:
```
1. Launch `cmd`.
2. Change dir into `notepad-plus-plus\PowerEditor\gcc`.
3. Type `mingw32-make target=i686`
4. `NotepadPP-release.exe` is generated in `notepad-plus-plus\PowerEditor\bin\`.
```

donho (changes requested):
Rework BUILD.md to make the procedure more explicit to users.

Issue #10511, opened 2021-09-06
A weird Find dialog glitch 


Yaron10:
**STR:**
Download a portable NPP 8.14.
Download [Notepad++.Win32.Release.exe](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/40651601/job/708pabs0b2886x10/artifacts) (https://github.com/notepad-plus-plus/notepad-plus-plus/commit/156784722a46fd250128ef415d6813a8171de5c7) and place it in the portable NPP 8.14 folder.
Run Notepad++.Win32.Release.exe.
Open the Find dialog.
Move/Resize the Find dialog.
Close NPP.
*Delete config.xml in the portable NPP 8.14 folder.*
Restart NPP.
Open the Find dialog.

**Result:**
![תמונה](https://user-images.githubusercontent.com/10229320/132248674-2bfc45f1-6833-4fef-9e98-53536abf8b91.png)

![תמונה](https://user-images.githubusercontent.com/10229320/132249670-60851578-7bb0-4ed1-b83a-7680d4c784a1.png)

**NOTE:** If the problem doesn't happen after the first time, repeat the STR.
I've encountered the issue in other scenarios too, but the above STR are relatively consistent.
I *suppose* other similar issues might pop up.

***

Notepad++ v8.1.4   (32-bit)
Build time : Sep  5 2021 - 14:35:33
Path : C:\Users\---\Downloads\npp.8.1.4.portable\Notepad++.Win32.Release.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

ivan-u7n:
The commit and the bug are not related, I was able to reproduce the bug with [a build from the previous commit](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/40650714/job/v9sgys2c8vg0l4t9/artifacts) too:
![image](https://user-images.githubusercontent.com/81974034/132254316-0a92c445-1981-4509-8704-a167070a6f2d.png)
```
Notepad++ v8.1.4   (32-bit)
Build time : Sep  5 2021 - 11:30:28
Path : F:\develop\git\notepad-plus-plus\PowerEditor\bin\Notepad++.Win32.Release.exe
[...]

PR #10524, opened 2021-09-08
properly fix #10454 (a crash after open) without causing #10511 (Find dialog glitch)

### LINKED ISSUE #10454: BUG with commit 68d293e - crash when open file
### Description of the Issue
Using the commit 68d293e , when using File => Open and opening a file, Notepad++ crashes.  The previous commit (e12b161) does not exhibit this behavior leading me to believe the bug was introduced in the 68d293e commit.

Based on the [discussion](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/68d293e7118a761aa4b6bc81fa934851480a950a#commitcomment-55542338) it may be related to #10206 since the 68d293e commit changes this value [from 30 to 38](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/68d293e7118a761aa4b6bc81fa934851480a950a#diff-11356ce8679f14c3c38f087779b0625511031cdef2dab261d775c8a40af1866c).

1. Download the AppVeyor Notepad++ artifact (I used 64-bit Release) for commit 68d293e 
2. Open Notepad++
3. Use File => Open and select a file to open, press the "Open" button
4. File does not open, Notepad++ crashes.

### Debug Information
Notepad++ v8.1.4   (64-bit)
Build time : Aug 25 2021 - 23:17:29
Command Line : -noPlugin
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 1909
OS Build : 18363.1734
Current ANSI codepage : 1252
Plugins : none

Windows 10, 64-bit, system-wide dark mode
Using a [custom theme](https://gist.github.com/vinsworldcom/71bba70153fc978818bf00863989b2ea) in Notepad++ (based on Obsidian), **NOT** using Notepad++ dark mode.

ivan-u7n:
The main bad guy was found: `Notepad_plus_Window` created on the stack in `wWinMain()`. `gdb` claims its (or inlined within `Notepad_plus`) size is a little under 2MB.

Will finally fix #10454 (a crash after open) and restore commits reversed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e9cf428388fc4e7f927b269be081d5876abb9963 without causing #10511 (Find dialog glitch) to appear again.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -681,13 +681,11 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	// Initialize the default foreground & background color
 	//
 	{
-		StyleArray & globalStyles = nppParam.getGlobalStylers();
-		int i = globalStyles.getStylerIndexByID(STYLE_DEFAULT);
-		if (i != -1)
+		const Style * pStyle = nppParam.getGlobalStylers().findByID(STYLE_DEFAULT);
+		if (pStyle)
 		{
-			Style & style = globalStyles.getStyler(i);
-			nppParam.setCurrentDefaultFgColor(style._fgColor);
-			nppParam.setCurrentDefaultBgColor(style._bgColor);
+			nppParam.setCurrentDefaultFgColor(pStyle->_fgColor);
+			nppParam.setCurrentDefaultBgColor(pStyle->_bgColor);
 		}
 	}
 
@@ -2259,12 +2257,9 @@ void Notepad_plus::setupColorSampleBitmapsOnMainMenuItems()
 
 	for (int j = 0; j < sizeof(bitmapOnStyleMenuItemsInfo) / sizeof(bitmapOnStyleMenuItemsInfo[0]); ++j)
 	{
-		StyleArray& stylers = NppParameters::getInstance().getMiscStylerArray();
-		int iFind = stylers.getStylerIndexByID(bitmapOnStyleMenuItemsInfo[j].styleIndic);
-		
-		if (iFind != -1)
+		const Style * pStyle = NppParameters::getInstance().getMiscStylerArray().findByID(bitmapOnStyleMenuItemsInfo[j].styleIndic);
+		if (pStyle)
 		{
-			Style const* pStyle = &(stylers.getStyler(iFind));
 
 			HDC hDC = GetDC(NULL);
 			const int bitmapXYsize = 16;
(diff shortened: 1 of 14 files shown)


ivan-u7n:
Everything with MSC-produced binaries seems good so far: no crash after opening a file in the 64-bit build and no Find dialog glitch in neither 32- nor 64-bit build.

donho:
But it doesn't demonstrate why Find dialog worked before the conversion of array to vector.

ivan-u7n:
It's hard to tell the exact reason in this case, but it seems that the uninitialized variable in `FindReplaceDlg`, as a part of `Notepad_plus_Window` allocated on the stack, just so happened to **always** be located in a zeroed memory block. But when something in the memory layout were changing, that became untrue.

I think the part "allocated on the stack" may be a part of the reason. The memory is managed by the OS, and it may have zeroed some regions of the stack address space for security or some other reasons. And when the size of the object had changed by using vectors instead of arrays, the offset of the variable has changed too to be mapped to an uninitialized memory block and to result in the glitch.

donho:
So the commit 14bd25a5e8b59ad9218083a94e99cfaa077fca82 (**an attempt to properly fix #10454**) is not necessary, is it?

ivan-u7n:
It seems so yes:
```
f:\develop\git\notepad-plus-plus\PowerEditor\gcc>gdb ..\bin\NotepadPP-debug.exe
GNU gdb (GDB) 8.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-w64-mingw32".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ..\bin\NotepadPP-debug.exe...done.
(gdb) p sizeof(Notepad_plus_Window)
$1 = 11216
(gdb)
```
11216 bytes seems rather small.

But I'd still apply it in case `Notepad_plus_Window` grows large again.

donho:
@ivan-u7n 
I just removed my revert (the last commit on master) by forcing push. So could you reset your local repository with the following instructions:
https://stackoverflow.com/questions/1125968/git-how-do-i-force-git-pull-to-overwrite-local-files

then redo a PR only with 31c3bdb3ea0cfe8735cd5869052a63e878d47232. for fixing the bug #10511.
(Do another PR for 14bd25a5e8b59ad9218083a94e99cfaa077fca82 if you judge necessary)

Sorry for the inconvenience, I'm trying to keep a clean repository.

PR #10531, opened 2021-09-09
fix #10454 (open crash) properly

### LINKED ISSUE #10454: BUG with commit 68d293e - crash when open file

ivan-u7n:
`Notepad_plus_Window` was rather huge (megabytes). Now it's small (tens of kilobytes), but to safeguard against future size increase move its allocation from the stack into the heap.

fix #10454 (open crash)

### DIFF
--- a/PowerEditor/src/Notepad_plus_Window.h
+++ b/PowerEditor/src/Notepad_plus_Window.h
@@ -57,11 +57,6 @@ filePath : file or folder name to open (absolute or relative path name)\r\
 class Notepad_plus_Window : public Window
 {
 public:
-	static Notepad_plus_Window & getInstance() {
-		static Notepad_plus_Window * instance = new Notepad_plus_Window;
-		return *instance;
-	}
-
 	void init(HINSTANCE, HWND, const TCHAR *cmdLine, CmdLineParams *cmdLineParams);
 
 	bool isDlgsMsg(MSG *msg) const;
--- a/PowerEditor/src/Notepad_plus_Window.h
+++ b/PowerEditor/src/Notepad_plus_Window.h
@@ -57,6 +57,11 @@ filePath : file or folder name to open (absolute or relative path name)\r\
 class Notepad_plus_Window : public Window
 {
 public:
+	static Notepad_plus_Window & getInstance() {
+		static Notepad_plus_Window * instance = new Notepad_plus_Window;
+		return *instance;
+	}
+
 	void init(HINSTANCE, HWND, const TCHAR *cmdLine, CmdLineParams *cmdLineParams);
 
 	bool isDlgsMsg(MSG *msg) const;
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus_Window.h
```diff
@@ -57,6 +57,11 @@ filePath : file or folder name to open (absolute or relative path name)\r\
 class Notepad_plus_Window : public Window
 {
 public:
+	static Notepad_plus_Window & getInstance() {
+		static Notepad_plus_Window * instance = new Notepad_plus_Window;
```
donho, line 61: `free` is missing.

donho (changes requested):
Add `delete` for freeing the allocated memory.

PowerEditor/src/Notepad_plus_Window.h
ivan-u7n: That was the same Singleton technique as the one used in `NppParameters`. Still to be identical in behaviour, I switched to `std::unique_ptr` in `wWinMain()`.

PowerEditor/src/winmain.cpp
```diff
@@ -609,7 +610,8 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
         }
 	}
 
-	Notepad_plus_Window notepad_plus_plus;
+	auto upNotepadWindow = std::make_unique<Notepad_plus_Window>();
```
donho, line 613: `up` in `upNotepadWindow` means `unique pointer` I suppose?

ivan-u7n: Exactly, `up` in `upNotepadWindow` stands for "**u**nique **p**ointer".

donho: It's OK for "up" - I just want to make sure.

PR #10539, opened 2021-09-11
Document List: Additional path column feature

### LINKED ISSUE #9679: [Feature Extension Request] Doc Switcher -- Additional column with full file path for sorting

Ashfaaq18:
fix #9679

DocListPath_Fix

The column widths are stored in the config file to retain the widths when closing and reopening Npp.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2451,11 +2451,16 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return langDesc.length();
 		}
 
-		case NPPM_DOCLISTDISABLECOLUMN:
+		case NPPM_DOCLISTDISABLEPATHCOLUMN:
+		case NPPM_DOCLISTDISABLEEXTCOLUMN:
 		{
 			BOOL isOff = static_cast<BOOL>(lParam);
 			NppGUI & nppGUI = nppParam.getNppGUI();
-			nppGUI._fileSwitcherWithoutExtColumn = isOff == TRUE;
+
+			if(message == NPPM_DOCLISTDISABLEEXTCOLUMN)
+				nppGUI._fileSwitcherWithoutExtColumn = isOff == TRUE;
+			else
+				nppGUI._fileSwitcherWithoutPathColumn = isOff == TRUE;
 
 			if (_pDocumentListPanel)
 			{
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2457,7 +2457,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			BOOL isOff = static_cast<BOOL>(lParam);
 			NppGUI & nppGUI = nppParam.getNppGUI();
 
-			if(message == NPPM_DOCLISTDISABLEEXTCOLUMN)
+			if (message == NPPM_DOCLISTDISABLEEXTCOLUMN)
 				nppGUI._fileSwitcherWithoutExtColumn = isOff == TRUE;
 			else
 				nppGUI._fileSwitcherWithoutPathColumn = isOff == TRUE;
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -833,6 +833,9 @@ struct NppGUI final
 	generic_string _themeName;
 	MultiInstSetting _multiInstSetting = monoInst;
 	bool _fileSwitcherWithoutExtColumn = false;
+	int _fileSwitcherExtWidth = 50;
+	bool _fileSwitcherWithoutPathColumn = false;
+	int _fileSwitcherPathWidth = 50;
 	bool isSnapshotMode() const {return _isSnapshotMode && _rememberLastSession && !_isCmdlineNosessionActivated;};
 	bool _isSnapshotMode = true;
 	size_t _snapshotBackupTiming = 7000;
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -832,9 +832,9 @@ struct NppGUI final
 	TCHAR _defaultDirExp[MAX_PATH];	//expanded environment variables
 	generic_string _themeName;
 	MultiInstSetting _multiInstSetting = monoInst;
-	bool _fileSwitcherWithoutExtColumn = false;
+	bool _fileSwitcherWithoutExtColumn = true;
 	int _fileSwitcherExtWidth = 50;
-	bool _fileSwitcherWithoutPathColumn = false;
+	bool _fileSwitcherWithoutPathColumn = true;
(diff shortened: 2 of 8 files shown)


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2451,11 +2451,16 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
 
-		case NPPM_DOCLISTDISABLECOLUMN:
+		case NPPM_DOCLISTDISABLEPATHCOLUMN:
+		case NPPM_DOCLISTDISABLEEXTCOLUMN:
 		{
 			BOOL isOff = static_cast<BOOL>(lParam);
 			NppGUI & nppGUI = nppParam.getNppGUI();
-			nppGUI._fileSwitcherWithoutExtColumn = isOff == TRUE;
+
+			if(message == NPPM_DOCLISTDISABLEEXTCOLUMN)
```
donho, line 2460: Use `if (` instead of `if(`.

donho (changes requested):
Please take consideration of following remarks:

Yaron10:
@Ashfaaq18,

👍 
Thank you for your great work.

I think the window needs to be redrawn in some cases.

**STR:**
Run this build for the first time with the old `config.xml`.
Increase (significantly) the width of the Path column.
Uncheck Ext.
Scroll horizontally.

**Result:**
תמונה

Recheck Ext.

**Result:**
תמונה
(You see the Ext. column).

Minimizing and restoring NPP solves the issue.

*If you run this build again with the new`config.xml`, you get the glitch immediately.* 

***

What's the purpose of displaying the Path column?
Is it only for sorting or also for actually seeing the path?

If the latter is correct, I'd suggest to have an option to display the Path column only.
It would make the context-menu implementation more complex.

What do you and @donho think?

Yaron10:
Should `New X` be displayed in the Path column?

Ashfaaq18:
> 
> What's the purpose of displaying the Path column?
> Is it only for sorting or also for actually seeing the path?
> 

The issue linked to this pr mentions sorting. 

I would prefer to see the path, (To help me navigate faster when having more than 2 same name files open from different directories)

>
>If the latter is correct, I'd suggest to have an option to display the Path column only.
>It would make the context-menu implementation more complex.
>

I can try adding an "only path" option in the context menu, 
When the user checks this, it will remove all the other columns in the list and, uncheck and grey out the other options in the context menu. To revert back, they'll have to uncheck the "only path" option.

Or

have the column list like this
image
the above is an example for: D:\Temp\textFile.txt 
it looks nicer since it flows naturally


> Should `New X` be displayed in the Path column?

I can remove this if its not much code.

@donho and anyone else, drop in your thoughts

donho:
> If the latter is correct, I'd suggest to have an option to display the Path column only.
> It would make the context-menu implementation more complex.
> 
> What do you and @donho think?

I would say it's better to remove file spec (the name part) and keep only the folder path in the path column.

Yaron10:
I think we can assume that many users would display the Path column for seeing the path and not only for sorting.
We should allow those users to see max-path-characters in min-panel-width (the editing area is an "expensive real estate"; especially in small monitors).

If changing the context-menu is too much, I'd suggest to keep the current implementation as it is.
The user could uncheck the Ext. column, push the Name column aside and see the path in one column.  

> Should `New X` be displayed in the Path column?
>> I can remove this if its not much code.

👍 
I think it's worth it.
This is how it's implemented in the Window dialog.

Yaron10:
@Ashfaaq18,

Thank you for fixing that glitch. 👍 

I'm afraid it requires some more fine-tuning.

**STR:**
Have the Ext column enabled and the Path column disabled.
Increase the Ext column width so that the horizontal scroll bar should appear.
Uncheck Ext.
Recheck Ext.

**Result:**
תמונה
And the horizontal scroll bar is not displayed.

It turns out to be quite complex.

PowerEditor/src/Parameters.h
```diff
@@ -832,9 +832,9 @@ struct NppGUI final
 	TCHAR _defaultDirExp[MAX_PATH];	//expanded environment variables
 	generic_string _themeName;
 	MultiInstSetting _multiInstSetting = monoInst;
-	bool _fileSwitcherWithoutExtColumn = false;
+	bool _fileSwitcherWithoutExtColumn = true;
```
Yaron10, line 835: I think this should be `false`.

donho: If my recall is good, the default value of withoutExt is true - to be verified in the old release.

Ashfaaq18:
> @Ashfaaq18,
> 
> Thank you for fixing that glitch. 👍
> 
> I'm afraid it requires some more fine-tuning.
> 
> **STR:**
> Have the Ext column enabled and the Path column disabled.
> Increase the Ext column width so that the horizontal scroll bar should appear.
> Uncheck Ext.
> Recheck Ext.
> 
> **Result:**
> תמונה
> And the horizontal scroll bar is not displayed.
> 
> It turns out to be quite complex.

@Yaron10 , If I'm right this happens when the increased "Ext." column width is still lower than the width of the dockable panel window. If you go past the width of the dockable window. The horizontal bar will appear even after an uncheck and recheck.

The reason for this is the column snapping code, where the column hugs the right side of the window. 
Uncheck-recheck of the column refreshes the column widths(window sizing as well), This refresh changes the width of the "Name" column to adjust according to the docked window while the "Ext." columns width remains fixed.

The current Npp ext cloumn has a fixed size, no resizing possible. 
I can set the ext column back to this constant if necessary.

donho:
@Ashfaaq18 
I just noticed that on path column header there's no sort mark after clicking it (like on name or ext one). Can it be added?

donho (approved):

Ashfaaq18:
> @Ashfaaq18
> I just noticed that on path column header there's no sort mark after clicking it (like on name or ext one). Can it be added?

@donho, The sort arrow shows up when I click the path column header. 

DocListPathSort_Prob

Is it a bug?

donho:
> > @Ashfaaq18
> > I just noticed that on path column header there's no sort mark after clicking it (like on name or ext one). Can it be added?
> 
> @donho, The sort arrow shows up when I click the path column header.
> 
> 
> Is it a bug?

Strange... On my Windows 10, mark on path column appears when ext column is disabled. However, there's no mark on path column header if all 3 columns are present.

We will see the other users' reaction.

Yaron10:
@Ashfaaq18,

I couldn't reply earlier.
Unfortunately, this PR was closed prematurely.

Thanks again for your work. I do appreciate it. 👍

donho:
@Yaron10 
> Unfortunately, this PR was closed prematurely.

Is there somethings wrong in this PR for you?

PR #10540, opened 2021-09-12
gcc build: 32-bit build, remove -fpermissive, auto -j in make and more

### LINKED ISSUE #9989: Fix or update instruction for MinGW 32-bit

ivan-u7n:
A new version of #10507 to fix #9989. I again messed-up with git…

Updates:
* The makefile now generates the binary in either `bin.i686` or in `bin.x86_64` in the current directory and also copies all required files into it. The naming of the directory depends on the target CPU of the compiler in use. Also the makefile now automatically creates missing `libsensapi.a` for the `i686` build to success.
* `BUILD.md` was also simplified to reflect the changes in the makefile. Now there is one common section about building Notepad++ with GCC with a note that `i686` build works only on a Windows system.
* Debug builds with GCC were disabled in `appveyor.yml` because they were of little practical use but huge in size and time consuming to produce. To be of use to someone, they need to have GDB installed. But then it's most certainly means they also have GCC to create a debug build by themselves.

`SensApi.dll` is a Windows system DLL and Notepad++ uses its functions in `SecurityGard::verifySignedLibrary()`, but for some reason `i686` version of MinGW-w64 GCC doesn't include a linking library for this DLL, thus it is generated on the fly.


ivan-u7n:
The `BUILD.md` in a diff is a mess, it's better to look at its [rendered version](https://github.com/ivan-u7n/notepad-plus-plus/blob/gcc-32bit/BUILD.md).

BUILD.md
```diff
@@ -41,33 +40,25 @@ More about the previous build process: https://community.notepad-plus-plus.org/t
...
-2. Change dir into `notepad-plus-plus\scintilla\win32`.
-3. Type `mingw32-make.exe -j%NUMBER_OF_PROCESSORS%`
-4. `libscilexer.a` is generated in `notepad-plus-plus\scintilla\bin\`.
+## Comping Notepad++ binary
 
-* Compile Notepad++ binary
+1. Launch `cmd` and add `$MinGW-root$\bin` to `PATH` if necessary.
+2. `cd` into `notepad-plus-plus\PowerEditor\gcc`.
+3. Run `mingw32-make`.
+4. `notepad++.exe` will be generated either in `bin.x86_64` or in `bin.i686` depending on the compiler used. The path to the directory is displayed at the end of the build process and it will contain everything needed for Notepad++ to run.
```
donho, line 56: why `bin.x86_64` **or** in `bin.i686` ? 
Does it depend on the compiler for building x64 or x86 bin ?
It's confusing to me (and for users surely).

donho (changes requested):
I've noticed a regression in this PR:
I cannot generate x64 bin in this PR. Calling mingw32-make.exe of mingw64:
It generated `notepad++.exe` in `PowerEditor\gcc\bin.i686\` as 32bit bin, and  copies all files from `PowerEditor\bin\` to `PowerEditor\gcc\bin.i686\` (or 64-bits notepad++.exe is replaced by x32-bits one while copying all files from `PowerEditor\bin\`).

Please fix this regression and correct evoked points.

ivan-u7n:
@donho 
> I've noticed a regression in this PR:
> I cannot generate x64 bin in this PR. Calling mingw32-make.exe of mingw64:
> It generated `notepad++.exe` in `PowerEditor\gcc\bin.i686\` as 32bit bin, and copies all files from `PowerEditor\bin\` to `PowerEditor\gcc\bin.i686\` (or 64-bits notepad++.exe is replaced by x32-bits one while copying all files from `PowerEditor\bin\`).
> 
> Please fix this regression and correct evoked points.

The first `g++` found in `PATH` wins. The makefile checks the target CPU of this `g++` and uses this value to make adjustments for the `i686` build. However the part about "the first `g++` found in `PATH` wins" hasn't changed but became more explicit. So please check which version of GCC is the first in your `PATH`. Although now only required files from `PowerEditor/bin` are copied, not the whole directory.

> `https://mingw-w64.org/doku.php/start` is no more available.
> Use `https://mingw-w64.org/` for replacing it.

The link to MinGW-w64 was changed.

> why `bin.x86_64` **or** in `bin.i686` ?
> Does it depend on the compiler for building x64 or x86 bin ?
> It's confusing to me (and for users surely).

Yes, it depends on the compiler, I tried to reword this sentence to be less confusing.

---

As a side note: the makefile now supports multilib GCC and this was tested with multilib GCC 10.2 which defaults to `x86_64`, but also supports building for `i686`. But it's not advertized in `BUILD.md` because the compiler lacks full support for C++11 threading and thus requires some additional workarounds.

donho:
@ivan-u7n 
Yes, it was my bad manipulation:
I have 2 32 & 64 compiler installed in my system, and I set only mingw for 32-bit in %PATH%. I used the full path of mingw32-make.exe for x64 build, but I guess gcc serch in 32-bit directory - that makes the build always in 32 bits.

Now I open 2 cmd : one set path including migw 32 & another set path including mingw 64.
The 32-bits bin is built with no issue, in the "bin.i686" folder. However, the 64-bits bin has not been generated, with the following error message:
```
* compiling ../src/WinControls/WindowsDlg/WindowsDlg.rc
* compiling ../src/WinControls/shortcut/RunMacroDlg.rc
* compiling ../src/WinControls/shortcut/shortcut.rc
* creating TARGET_DIRECTORY bin.x86_64
* compiling manifest.rc
* copying bin.x86_64/contextMenu.xml
        1 file(s) copied.
* copying bin.x86_64/langs.model.xml
        1 file(s) copied.
* copying bin.x86_64/shortcuts.xml
        1 file(s) copied.
* copying bin.x86_64/stylers.model.xml
        1 file(s) copied.
* copying bin.x86_64/change.log
        1 file(s) copied.
* copying bin.x86_64/doLocalConf.xml
        1 file(s) copied.
* copying bin.x86_64/userDefineLangs
2 File(s) copied
* copying bin.x86_64/readme.txt
        1 file(s) copied.
* copying bin.x86_64/autoCompletion
27 File(s) copied
* copying bin.x86_64/functionList
33 File(s) copied
* copying bin.x86_64/themes
22 File(s) copied
* copying bin.x86_64/localization
94 File(s) copied
* compiling ../src/WinControls/shortcut/shortcut.cpp
* compiling ../src/localization.cpp
* compiling ../src/winmain.cpp
* linking bin.x86_64/notepad++.exe
C:/appz/mingw/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: skipping incompatible bin.x86_64.build\libscilexer.a when searching for -lscilexer
C:/appz/mingw/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lscilexer
collect2.exe: error: ld returned 1 exit status
mingw32-make.exe[1]: *** [makefile:216: bin.x86_64/notepad++.exe] Error 1
mingw32-make.exe: *** [makefile:178: all] Error 2

C:\xxx\notepad-plus-plus\PowerEditor\gcc>
```

Any idea about the error ?

ivan-u7n:
@donho 
> Now I open 2 cmd : one set path including migw 32 & another set path including mingw 64.
> The 32-bits bin is built with no issue, in the "bin.i686" folder. However, the 64-bits bin has not been generated, with the following error message:
> 
> ```
> ...
> * linking bin.x86_64/notepad++.exe
> C:/zzzz/mingw/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: skipping incompatible bin.x86_64.build\libscilexer.a when searching for -lscilexer
> C:/zzzz/mingw/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lscilexer
> collect2.exe: error: ld returned 1 exit status
> mingw32-make.exe[1]: *** [makefile:216: bin.x86_64/notepad++.exe] Error 1
> mingw32-make.exe: *** [makefile:178: all] Error 2
> 
> C:\xxx\notepad-plus-plus\PowerEditor\gcc>
> ```
> 
> Any idea about the error ?

That's the unfortunate result of not very adequate makefile of Scintilla combined with using different compilers in parallel. In this case `mingw32-make clean all` will work as it will clean all intermediate object-files in `scintilla/win32`. I didn't add this into `BUILD.md` assuming always having the clean source tree.

donho:
> That's the unfortunate result of not very adequate makefile of Scintilla combined with using different compilers in parallel. In this case `mingw32-make clean all` will work as it will clean all intermediate object-files in `scintilla/win32`. I didn't add this into `BUILD.md` assuming the source tree being always clean.

After clean both Scintilla win32 dire and Notepad++ dire, x64 bin has been built smoothly.
Now both binaries are generated in different folder - nice work!

I would reword the following sentence even more explicitly:
```
The 64-bits or 32-bits binary notepad++.exe will be generated either in bin.x86_64 or in bin.i686 directory respectively, depending on the target CPU of the compiler — look for the full path to the resulting binary at the end of the build process. The directory with the binary will contain everything needed for Notepad++ to start.
```

What do you think?

ivan-u7n:
> I would reword the following sentence even more explicitly:
> 
> ```
> The 64-bits or 32-bits notepad++.exe will be generated either in bin.x86_64 or in bin.i686 directory respectively, depending on the target CPU of the compiler — look for the full path to the resulting binary at the end of the build process. The directory with the binary will contain everything needed for Notepad++ to start.
> ```
> 
> What do you think?

[Done](https://github.com/ivan-u7n/notepad-plus-plus/blob/gcc-32bit/BUILD.md#comping-notepad-binary). I just changed the order. I also moved all other points into a list and added an item into it about switching compilers and modes.

donho:
> Building a 32-bit version of Notepad++ currently works only on Windows because %windir%\system32\SensApi.dll is required for the build process to succeed — the tested i686 version of GCC doesn't have a linking library for this system DLL, so it is generated during the build process.

Do you mean only notepad++.exe 32-bits built by GCC cannot run with WINE under Linux (or OSX)?

ivan-u7n:
> > Building a 32-bit version of Notepad++ currently works only on Windows because %windir%\system32\SensApi.dll is required for the build process to succeed — the tested i686 version of GCC doesn't have a linking library for this system DLL, so it is generated during the build process.
> 
> Do you mean only notepad++.exe 32-bits built by GCC cannot run with WINE under Linux (or OSX)?

No, although I haven't tested it, it should run, provided WINE has this DLL.

I mean exactly what is written: 32-bit version of Notepad++ can't be *compiled* without `%windir%\system32\SensApi.dll` being present, because the build process uses this DLL straight from the system to create missing linking library from it. When the binary is compiled, it uses this DLL the same way as all other Windows system ones.

donho:
> I mean exactly what is written: 32-bit version of Notepad++ can't be _compiled_ without `%windir%\system32\SensApi.dll` being present, because the build process uses this DLL straight from the system to create missing linking library from it.

Why only for compiling 32-bits notepad++.exe ?
And is `%windir%\system32\SensApi.dll` always present, isn't it ?

ivan-u7n:
> Why only for compiling 32-bits notepad++.exe ?
> And is `%windir%\system32\SensApi.dll` always present, isn't it ?
> 
> Edit:
> OK, answer my 1st question with your previous answer:
> 
> ```
> but for some reason i686 version of MinGW-w64 GCC doesn't include a linking library for this DLL, thus it is generated on the fly.
> ```

Yes, even more recent version I found (the 10.2 multilib from https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10540#issuecomment-917718427) have this library for its 64-bit variant, but still lacks it for its 32-bit variant.

According to [Microsoft](https://docs.microsoft.com/en-us/windows/win32/api/sensapi/nf-sensapi-isnetworkalive) it appears in Windows XP, so yes one can say it is always present.

donho:
So the following statement:
```
Building a 32-bit version of Notepad++ currently works only on Windows because %windir%\system32\SensApi.dll is required for the build process to succeed — the tested i686 version of GCC doesn't have a linking library for this system DLL, so it is generated during the build process.
```
is for explaining the following code:

```
ifeq "$(TARGET_CPU)" "i686"
$(TARGET_BINARY): $(BUILD_DIRECTORY)/libsensapi.a
$(BUILD_DIRECTORY)/libsensapi.a: | $(BUILD_DIRECTORY)
	@echo * generating $@
	$(AT)gendef $(call normalize-path,$(firstword $(wildcard $(windir)/syswow64/SensApi.dll $(windir)/system32/SensApi.dll)))
	$(AT)dlltool -mi386 -f--32 -d SensApi.def -k -l $(call normalize-path,$@)
	$(AT)$(RM) SensApi.def
endif
```

Such info is very interesting for dev who maintain makefile, but not for "end users" who grab code source and try to have a working build - Not only they (the end users) don't care about such info, but also is confusing for them. Since you have already managed it for the end users in makefile, I would suggest to move this info from `BUILD.md` to `gcc\makefile` (as comment of course).

donho (changes requested):
Move the info **Notes on using GCC targeting i686** from BUILD.md to `gcc\makefile` as comment

donho (approved):

MarkusBodensee:
@ivan-u7n
Sorry for the noobish question, but did I get it correct, that `mingw32-make.exe` is responsible for building both, 32 bit and 64 bit version, depending on the available MinGW in PATH? I am just confused of the terminology of `mingw32-make.exe`.

I read somewhere in the comments that the correct architecture is choosen automatically. If this is true, is it worth adding a comment about this to the `.md` file? Or am I just the only one which is confused ;)

Another question which just came to my mind: How is it possible to distinguish between the 32 bit and 64 bit build? Appart from About dialogue and so on where this info is added "by hand/code". Can it be seen from the binary?

ivan-u7n:
@MarkusBodensee
Yes, the name of the executable for GNU Make in both `i686` and `x86_64` builds of GCC by MinGW-w64 is `mingw32-make` which might be a bit confusing.

The target CPU of the first found GNU C++ compiler in `PATH` determines the name of the output directory. But if you have only one compiler, it's no brainer to guess in which architecture the result will be. Moreover something similar is already in `BUILD.md`:
> 4. The 32-bit or 64-bit `notepad++.exe` will be generated either in `bin.i686` or in `bin.x86_64` directory respectively, depending on the target CPU of the compiler — look for the full path to the resulting binary at the end of the build process.

Without external tools one can't distinguish bitness of a binary.

MarkusBodensee:
@ivan-u7n
Thank you very much for explanation and for the details.

PR #10541, opened 2021-09-12
Make code more sanitary

donho:
Fix some vulnerable codes detected by VS's analyzing.


mere-human:
I wonder, what sort of analysis did you use? 
Are there any extra warnings reported by Visual Studio? 
If so, should we mark them as errors?

donho:
@mere-human 
> I wonder, what sort of analysis did you use?

In VS:

image

> Are there any extra warnings reported by Visual Studio?
If so, should we mark them as errors?

I don't think so. At least, I don't know how.

PR #10549, opened 2021-09-14
Use Microsoft DateTime format for "Insert Custom DateTime" command

donho:


PowerEditor/src/Parameters.h
```diff
@@ -784,7 +784,7 @@ struct NppGUI final
 	generic_string _uriSchemes = TEXT("svn:// cvs:// git:// imap:// irc:// irc6:// ircs:// ldap:// ldaps:// news: telnet:// gopher:// ssh:// sftp:// smb:// skype: snmp:// spotify: steam:// sms: slack:// chrome:// bitcoin:");
 	NewDocDefaultSettings _newDocDefaultSettings;
 
-	generic_string _dateTimeFormat = TEXT("Y-M-D h:m:s");
+	generic_string _dateTimeFormat = TEXT("yyyy-MM-dd hh:mm:ss");
```
ivan-u7n, line 787: Maybe it's better to use `HH` as the default for hours?

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1333,23 +1333,33 @@ int nbDigitsFromNbLines(size_t nbLines)
...
+	generic_string dateTimeStr = buffer;
+
+	generic_string AM_number = TEXT("A");
+	AM_number += M_number;
+
+	generic_string PM_number = TEXT("P");
+	PM_number += M_number;
+
+	dateTimeStr = stringReplace(dateTimeStr, AM_number, TEXT("AM"));
+	dateTimeStr = stringReplace(dateTimeStr, PM_number, TEXT("PM"));
```
ivan-u7n, line 1363: Maybe instead of postprocessing, a bit of preprocessing will be a better option:
```c++
	size_t tPos = 0;
	{
		size_t tEnd = tPos + 1;
		for ( ; tEnd < formatEx.length() && formatEx[tEnd] == L't'; ++tEnd);
		tPos += tEnd + 8;
	}

	formatEx = buffer;
```
This way no matter into what `t` or `tt` is converted, it will always be escaped for `GetDateFormatEx()`.

donho: Amazing @ivan-u7n !
The test input:
`hh:mm:ss, tt; M dd yy g t`
transformation:
```
hh:mm:ss, ''''tt''''; M dd yy g ''''t''''
01:24:00, 'AM'; M dd yy g 'A'
```
Expected output: 
`01:24:00, AM; 10 26 85 A.D. A`

My question is: how did you know the trick of `''''xxx''''` for `GetTimeFormatEx`? Any ref about that?
Why `GetDateFormatEx` remove only 1 pair of quotes instead of 3?

donho: BTW, may I integrate your code or you prefer to create a PR under your name?

ivan-u7n: The "trick" is from https://docs.microsoft.com/en-us/windows/win32/intl/hour--minute--and-second-format-pictures:
> Single quotation marks can be used to mark characters that should be displayed exactly as specified. If the application must display a single quotation mark, it should place two single quotation marks in a row. For example, 'abc''bar', is displayed as "abc'bar".

So to get a single quotation mark, there are must be 4 of them: two at the ends are for the start and the end of the escaped string, and two in the middle are turned into one in the output.

However the code above will break if a time marker is absent or there's a single quotation mark in it. That code also doesn't do anything to user escaped strings. This one will only break if there a single quotation mark in a time marker:
```c++
{
	const DWORD flags = 0;

	size_t findPos = 0;
	size_t tResultLength = std::string::npos;
	while ((findPos = formatEx.find_first_of(L"'t", findPos)) != std::string::npos)
	{
		if (formatEx[findPos] == L'\'')
		{
			// single quotation mark escape
			size_t findEnd = findPos + 1;
			while ((findEnd = formatEx.find_first_of(L"'", findEnd)) != std::string::npos)
			{
				if (findEnd + 1 < formatEx.length() && formatEx[findEnd + 1] == L'\'')
				{
					formatEx.insert(findEnd, 2, L'\'');
					findEnd += 4;
					continue;
				}
				break;
			}
			if (findEnd == std::string::npos)
			{
				// no closing single quotation mark
				formatEx.push_back(L'\'');
				findEnd = formatEx.length();
			}
			formatEx.insert(findEnd, 2, L'\'');
			formatEx.insert(findPos, 2, L'\'');
			findPos = findEnd + 4 + 1;
		}
		else
		{
			// time marker
			if (tResultLength == std::string::npos)
			{
				// if time marker is defined in locale...
				GetTimeFormatEx(localeName, flags, &st, L"t", buffer, bufferSize);
				tResultLength = wcslen(buffer);
			}
			size_t findEnd = formatEx.find_first_not_of(L"t", findPos + 1);
			if (findEnd == std::string::npos)
				findEnd = formatEx.length();
			if (tResultLength)
			{
				// ...then escape its result...
				formatEx.insert(findEnd, 4, L'\'');
				formatEx.insert(findPos, 4, L'\'');
				findPos = findEnd + 8;
			}
			else
			{
				// ...or remove it: its somewhat magical removal by GetTimeFormatEx() breaks escaping
				formatEx.erase(findPos, findEnd - findPos);
			}
		}
	}

	formatEx = buffer;

	return buffer;
}
```
`hh:mm:'s''s' tt; 'yy''yy'-MM-dd` => `07:48:s's ; yy'yy-09-15`

Should I try to fix the last small bug (the breakage if there a single quotation mark in a time marker) and then open a PR for my version?

mere-human: Is this code easier than my [PR](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480)?

donho: > However the code above will break if a time marker is absent or there's a single quotation mark in it. That code also doesn't do anything to user escaped strings

Example please ?

>. This one will only break if there a single quotation mark in a time marker:

Example please ?

donho: @mere-human 
> Is this code easier than my [PR](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480)?

This one yes:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10549#discussion_r709139701

donho: @ivan-u7n 
> Should I try to fix the last small bug (the breakage if there a single quotation mark in a time marker) and then open a PR for my version?

The code becomes more complex than the first version you provided.
If the most cases work with the first version, it'll be fine.

MarkusBodensee: Small summary of the latest approaches:

This one [#10549 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10549#discussion_r709139701) will not work if the AM/PM symbols are absent/empty in the windows settings.

This one [#10480 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480#issuecomment-920032850) will cover the cases of empty AM/PM symbols and is not much more complicated. So it is the more reliable solution.

If single quotation mark should be handled in customised AM/PM symbols, the [#10480 approach](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10480) of @mere-human looks like the way to go.

If more escaping should be handled, everything gets complex ;-)

ivan-u7n: > If more escaping should be handled, everything gets complex ;-)

No overly complex, only ~100 lines of code: #10556 😎

donho:
@ivan-u7n 
Just change the Region settings from USA to France:
Here the result with the PR applied:

image

I would say these 2 MS API work in mysterious way. :/

ivan-u7n:
> @ivan-u7n
> Just change the Region settings from USA to France:

My original is Russia and it also doesn't have things like `AM` or `PM`.

> I would say these 2 MS API work in mysterious way. :/

I'll repeat here:
*`t` or `tt` processing in a format string requires special handling because GetTimeFormatEx(), in case of an empty time-marker value in locale settings, **removes everything preceding a time-marker token up to another Time token**; so if a time-marker token is the first Time token in a string with only Date tokens before it, the output will have only results of tokens after that time-marker token*

> _**Edit: I've tested with your first version, the result is the same :(**_

That's why the more complex processing than one would think is required. :(

mere-human:
@donho this PR works fine with French region.
image
I really don't get it why you don't like it. 

How do you define "complexity" of code? 
Is it the number of lines? Is it the amount of if-else branches? 
Should it be listed in the contribution guidelines somehow?
What matters the most is the _value that feature brings to the user_.

There are a lot of complex (code if you could call it so) in Notepad++ already: dark mode is complex, [Notepad_plus](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Notepad_plus.cpp#L123) is big (~8K lines) and complex, [CustomFileDialog](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp#L881) is complex. But all these things bring some value needed by users.

donho:
@ivan-u7n @mere-human 
I have tested both code of your PR for changing the regions, both PR work fine.
However, the PR of @mere-human is much less complex, so I prefer to use his PR.
OTOH, I will keep the PR of @ivan-u7n opened in case the edge cases that we haven't seen after release. 
Thank you for the effort of both of you!

ivan-u7n:
The only thing I don't see in the PR of @mere-human — the handling of verbatim lines, e.i. to have verbatim 'date', 'time' or the like in the output. Although a user will just need to use 3 single quotes instead of one for such desires.

mere-human:
> the handling of verbatim lines, e.i. to have verbatim 'date', 'time' or the like in the output.

This can be handled in a separate feature request if there's a need.

PR #10556, opened 2021-09-15
implement custom datetime format via Windows API

ivan-u7n:
The implementation got complex to support all corner cases.
Two showcases that work:
1. `hh:mm:'s''s'tt; 'yy''yy'-MM-dd` giving `10:27:s'sa'md; yy'yy-09-15` if `PM` string is set to `a'md`.
2. `hh:mm:'s''s' tt; 'yy''yy'-MM-dd` giving `10:30:s's; yy'yy-09-15` if `PM` string is not set.

EDIT: The current simplified version doesn't handle the cases above, but everything in examples works well:
image

EDIT2: A case for a more complex processing: someone wanting to set format to `'date:' yyyy-MM-dd; 'time:' HH:mm:ss;` expecting to get `date: 2021-09-15; time: 22:30:31;`, but getting `15ae: 2021-09-15; ime: 22:30:31;` and becoming frustrated…


ivan-u7n:
@donho 
Here it is. Although it did become somewhat complex. But if all corner cases aren't important (like `AM` or `PM` strings containing quotes, any user escaping, and so on) then it can be simplified a bit.

BTW, maybe it's better to disable AppVeyor autobuilds if only xml or text files were changed?

donho:
@ivan-u7n 
Could you provide some samples which work in the current PR but not work in your first version:
 
```
{
	const DWORD flags = 0;

	size_t tPos = 0;
	{
		size_t tEnd = tPos + 1;
		while (tEnd < formatEx.length() && formatEx[tEnd] == L't') ++tEnd;
		tPos += tEnd + 8;
	}

	formatEx = buffer;

	return buffer;
}
```

ivan-u7n:
@donho ofcourse:
1. `yyyy-MM-dd tt HH:mm:ss` in locales without a time-marker will result in ` 14:33:22` thanks to over-aggressive token removal by `GetTimeFormatEx()`.
The reason is documented in the comment in the code:
*GetTimeFormatEx(), in case of an empty time-marker value in locale settings, removes everything preceding and succeeding a time-marker token up to another time token; so if a time-marker token is the first time token in a string with only date tokens before it, the output will have only tokens after the time-marker one*
(although I may be wrong about the "succeeding" part)
2. The reversed `HH:mm:ss tt yyyy-MM-dd` in locales without a time-marker will result in `14:33:22 yyyy-MM-dd` due to `tt` removal eating one escape delimiter.
3. The same as 1. `yyyy-MM-dd tt HH:mm:ss` but in locales with a time-marker with a single quotation mark (I set `PM` my marker to `a'md`) will result in `2021-09-16 am16 14:38:07`.
4. Any format string with escaping, for example `'date:' yyyy-MM-dd; 'time:' tt HH:mm:ss;` will result in ` 14:43:26;` or in `'16ate:' 2021-09-16; time:' PM 14:43:26;` depending on the presence of the time-marker in locale settings (in this it is simple `PM`).

However this PR is based on the same principles as the prototype I showed. Proper handling of escape delimiters and empty time-marker values I think is a must, and all other corner cases are conditioned based on the presence of escape delimiters in either format string or in a time-marker value.

PR #10566, opened 2021-09-19
Fix undoing a modification make deleted file from outside undirty

### LINKED ISSUE #10401: Deleting a file externally, and then undoing+redoing in notepad++, marks the file as saved even though it isn't

1. In Windows Explorer, create a new file named `test.txt`.
2. Open that file in Notepad++
3. Enter `Test`, then hit <kbd>control</kbd>+<kbd>S</kbd> to save.
4. Delete the file in Windows Explorer.  Either a regular delete/move to recycle bin (<kbd>delete</kbd>) or a real delete (<kbd>shift</kbd>+<kbd>delete</kbd>) will work.
5. In Notepad++, when prompted to keep the file in editor, select yes.
6. Undo with <kbd>control</kbd>+<kbd>Z</kbd>
7. Redo with <kbd>control</kbd>+<kbd>Y</kbd>

### Expected Behavior
The file should be marked as unsaved, since it is unmodified.

### Actual Behavior
After redoing, the file is no longer marked as unsaved, and the save button is greyed out, even though the file no longer exists.

Interestingly, "File &rarr; Reload from Disk" and "File &rarr; Move to Recycle Bin" are greyed out, consistent with a nonexistent file.  Also, Notepad++ detects if the file is created again and prompts to reload it (and also marks the file as dirty in that case, and re-enables those operations).  Both of these are useful behaviors that should probably be preserved if this issue is fixed.

### Debug Information
```
Notepad++ v8.1.2   (32-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1165
Current ANSI codepage : 1252
```

ripmoneygone2020:
for issue #10401. _mainDocTab saved status isn't being updated correctly when a external file delete of the currently opened buffer happens. Initially, if there is text in the buffer it will display as unsaved. But after a undo, and redo operation, the _mainDocTab displays the incorrect saved status. If the file has been deleted externally, and the user opts to keep the file open and there is text in the buffer it should be displayed as unsaved until the file is saved.

problem may also be in NppNotification.cpp (starting line 122). because if a SCN_SAVEPOINTREACHED the code says you can no longer undo, which results isDirty being set to false. and when bufferUpdated is called _mainDocTab will see that the buffer isn't dirty and not display the correct saved status

Fix #10401

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -123,8 +123,15 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 			bool isSnapshotMode = NppParameters::getInstance().getNppGUI().isSnapshotMode();
 			if (isSnapshotMode && !isDirty)
 			{
+				DocFileStatus status = buf->getStatus();
+				bool isDocDeleted = status == DOC_DELETED;
+				bool isDocModified = status == DOC_MODIFIED;
+				bool isDocRegular = status == DOC_REGULAR;
+
 				bool canUndo = _pEditView->execute(SCI_CANUNDO) == TRUE;
 				if (!canUndo && buf->isLoadedDirty() && buf->isDirty())
+					isDirty = true;
+				else if(canUndo && (isDocDeleted || isDocModified) && !isDocRegular && buf->isDirty())
 					isDirty = true;
 			}
 			buf->setDirty(isDirty);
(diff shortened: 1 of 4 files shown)


ripmoneygone2020:
closing this, its not fixed.

PR #10568, opened 2021-09-19
gcc-build: don't depend on Scintilla's makefile

ivan-u7n:
* All parts of Scintilla are now built with Notepad++'s makefile ignoring Scintilla's one entirely. This prevents build errors when switching between compilers (like https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10540#issuecomment-917721735).
* Update `BUILD.md`.
* Fix warnings in `boostregex/BoostRegExSearch.cxx` (`-Wall -Wpedantic` are the default for Scintilla).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,8 +151,7 @@ PowerEditor/bin/SourceCodePro-Regular.ttf
 #-------------------------------------------------------------------------------
 # MinGW-w64 GCC
 
-bin.*
-scintilla/win32/*.o
+bin.*/
 
 
 PowerEditor/src/MISC/md5/RCa06792
--- a/BUILD.md
+++ b/BUILD.md
@@ -44,18 +44,23 @@ uses features from Boost's `Boost.Regex` library.
 
 If you have [MinGW-w64](https://www.mingw-w64.org/) installed, you can compile Notepad++ with GCC.
 
-MinGW-w64 can be downloaded [here](https://sourceforge.net/projects/mingw-w64/files/). Building Notepad++ is regularly tested on a Windows system with [x86_64-8.1.0-release-posix-seh-rt_v6-rev0](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z) for building 64-bits binary and with [i686-8.1.0-release-posix-dwarf-rt_v6-rev0](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/dwarf/i686-8.1.0-release-posix-dwarf-rt_v6-rev0.7z) versions for building 32-bits binary. Other versions may also work but are untested.
+MinGW-w64 can be downloaded [here](https://sourceforge.net/projects/mingw-w64/files/). Building Notepad++ is regularly tested on a Windows system using:
+* [x86_64-8.1.0-release-posix-seh-rt_v6-rev0](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z) for building a 64-bit binary;
+* [i686-8.1.0-release-posix-dwarf-rt_v6-rev0](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/dwarf/i686-8.1.0-release-posix-dwarf-rt_v6-rev0.7z) for building a 32-bit binary.
+Other versions may also work but are untested.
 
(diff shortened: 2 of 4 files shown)


donho:
@ivan-u7n 
The new versions of Scintilla will be upgraded into Notepad++, so the build of SciLexer should be independent of Notepad++ project. It will be an extra burden for us to maintain Scintilla's makefile in Notepad++.
Hence the reject of this PR.

ivan-u7n:
@donho
> The new versions of Scintilla will be upgraded into Notepad++, so the build of SciLexer should be independent on Notepad++ project. It will be an extra burden for us to maintain Scintilla's makefile in Notepad++.
> Hence the reject of this PR.

I disagree, because problems if any can arise only when there are major updates like extracting Lexilla from Scintilla. But in this case they will always happen.

Moreover, the makefile is written in such a way, that minor changes like adding or removing files won't be a problem at all. Also it was reworked with the future addition of Lexilla into the mix in mind. However if required, I can extract bits responsible for external libraries into individual files, but I prefer to have all configurable parameters in one spot.

ivan-u7n:
@donho
Will something like https://github.com/ivan-u7n/notepad-plus-plus/commit/cca49c4438902e9997f176456f3c5c055369e8fe with the ability to use Scintilla's makefile in case it is needed be accepted?

I'll copy the whole commit message containing details here:
> **gcc: build Scintilla with Notepad++'s makefile**
>
> Added the unified build mode. Notepad++'s makefile now by default builds Scintilla on its own ignoring the latter's makefile entirely. This prevents build errors when switching between compilers by moving all intermediate files into one directory.
>
> The previous independent build mode using Scintilla's own makefile can still be activated via passing `INDEPENDENT=1` to `mingw32-make`. Although a workaround for the errors above was added into this mode too, it prevents incremental rebuild of Scintilla when its source code is changed.
> 
> Updated `BUILD.md` to document `clean` and `fullclean` makefile goals with rewording of other parts.
>
> Fixed warnings in `boostregex/BoostRegExSearch.cxx` (`-Wall -Wpedantic` are the default for Scintilla).

donho:
> Moreover, the makefile is written in such a way, that minor changes like adding or removing files won't be a problem at all. Also it was reworked with the future addition of Lexilla into the mix in mind. 

We will see when it happens, then we talk about it


> The previous independent build mode using Scintilla's own makefile can still be activated via passing INDEPENDENT=1 to mingw32-make. Although a workaround for the errors above was added into this mode too, it prevents incremental rebuild of Scintilla when its source code is changed.

My argument remains unchanged: For the facility of maintaining, Scintilla's makefile can be call inside of Notepad++ makefile, but we don't include it. Hope you understand.

> Fixed warnings in boostregex/BoostRegExSearch.cxx (-Wall -Wpedantic are the default for Scintilla).

This part is acceptable of course.

ivan-u7n:
> My argument remains unchanged: For the facility of maintaining, Scintilla's makefile can be call inside of Notepad++ makefile, but we don't include it. Hope you understand.

I've got your point. And I'd wholeheartedly agree if one makefile was copy-pasted into another — that's a sure call for all kinds of troubles. But in this case the only things really copied were two defines: `SCI_LEXER` to enable lexers and `SCI_OWNREGEX` to use Boost.Regex. Handling of everything else is done by the features of Notepad++'s makefile already used for building Notepad++ itself — I just add several top-level Scintilla's directories to the list.

Moreover, as this unified build mode seems more intended for developers working on Scintilla/Lexilla with Notepad++ as its gateway and using GCC as the main compiler, I think I'll inverse the logic to make the unified mode opt-in and the current independent one — the default. This way more developer friendly build mode will be available, but wouldn't be forced. And in time it can become the default one if it proves itself trouble-free.

donho:
If the structure of Scintilla project changes completely in the future, then we have to change also the makefile of Notepad++ -  That is what I prevent from.
The makefile of Notepad++ should build only notepad++.exe and link Scintilla lib.

My decision remains the same.

ivan-u7n:
> If the structure of Scintilla project changes completely in the future, then we have to change also the makefile of Notepad++ - That is what I prevent from.

Here is a contradiction, for in case of Scintilla's layout change, the makefile of Notepad++ needs to change anyway to at least point to a new location of Scintilla's makefile.

> The makefile of Notepad++ should build only notepad++.exe and link Scintilla lib.

I can understand that desire. So now I'm going to see if I can make Scintilla's makefile more cooperative without rewriting it from scratch. It will be excellent if I can move object files it generates into some directory which external callers can specify. Scintilla's makefile is already modified, so I hope one more little change won't be a problem.

donho:
> I can understand that desire. So now I'm going to see if I can make Scintilla's makefile more cooperative without rewriting it from scratch. It will be excellent if I can move object files it generates into some directory which external callers can specify. Scintilla's makefile is already modified, so I hope one more little change won't be a problem.

Scintilla's makefile has been modified for a good reason, and the modification is minor : 
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ab58c8ee3ed1b8f8fa84a06c0f7d304fb0f0cd76#diff-81078873c7192ded6b5a93da2d62b0e6f86a5516905529f465fa5b7c33bb2df4

The PR of modification of Scintilla's makefile won't be accepted if the advantage is not significant, in the aim of making Scintilla updating as easier as possible in the future.

ivan-u7n:
Although I found a rather hackish way to do what I wanted without modifying Scintilla's makefile, I decided it's better to patch makefiles upstream: Lexilla: https://github.com/ScintillaOrg/lexilla/pull/31; Scintilla: https://sourceforge.net/p/scintilla/bugs/2283/.

🤞 for making it into 5.1.2 release.

ivan-u7n:
Unfortunately patches didn't make it into 5.1.2, so my PR #10580 uses the hackish way for now.

PR #10571, opened 2021-09-19
UDL: add checkbox for DarkMode-ness to the UDL dialog

### LINKED ISSUE #10584: The UDL dialog has no indicator whether current UDL is for Dark Mode or not
### Description of the Issue
Currently in the UDL dialog there is neither an indicator to see whether current UDL is for Dark Mode or not, nor a way to change it. This can be done only by checking and changing the corresponding XML file.

1. Open the UDL dialog.

### Expected Behavior
image

### Actual Behavior
image

### Debug Information
```
Notepad++ v8.1.4   (64-bit)
Build time : Sep 22 2021 - 20:47:49
Built with : GCC 8.1.0
Path : F:\develop\git\notepad-plus-plus\bin.x86_64\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise LTSC 2019 (64-bit) 
OS Version : 1809
OS Build : 17763.2183
Current ANSI codepage : 1251
Plugins : none
```

ivan-u7n:
Currently there is no way in GUI to mark UDL for Dark Mode. Adding a checkbox solves this problem.

This PR may be too late for 8.1.5. That's why I made two PRs — so this one can wait after 8.1.5 is released and the other one can be merged having no impact on translations.

Fix #10584


donho:
Please create an issue to describe the bug and how to reproduce it, and link the PR to the issue (Fix #XXXXX).

ivan-u7n:
@donho done

donho:
As my explanation [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10584#issuecomment-925427961) , this feature won't be implemented.

ivan-u7n:
OK, I see the point.

I though about it more thorough: the dark mode (DM) is all about colors, so maybe it's better to add another pair of colors for DM in all necessary places and to switch between them when the mode changes. This will allow to have a single UDL definition for both modes and won't require manual actions when switching between modes. And that's only what I came up with on the go.

What do you think? Is it worth trying to implement?

donho:
@ivan-u7n Sorry, didn't see the reply - please tag me next time so I will be notified by email.

> I though about it more thorough: the dark mode (DM) is all about colors

Yes, it was the intention - to ignore styles' white bg color and to apply the dark bg color for all styles, so the same UDL file could be fit in both dark mode & light mode. But the solution is far from perfect since the fg color matters as well.

> so maybe it's better to add another pair of colors for DM in all necessary places and to switch between them when the mode changes. This will allow to have a single UDL definition for both modes and won't require manual actions when switching between modes. And that's only what comes first to my mind.

As I mentioned above, the fg color matters as well. A group of keywords set in dark blue in light mode cannot be distinguished in dark mode for example. 

So I won't make more effort for this issue. Thank you anyway for thinking of it.

ivan-u7n:
@donho
> Sorry, didn't see the reply - please tag me next time so I will be notified by email.

I though so, will do next time.

However I don't think you got my intention: I wasn't saying that all old UDLs will automatically become dark mode compatible. Although something akin to this can be done if crudely by just inverting the light mode colors, I don't think the result will be pretty, but as a last resort it will do. But I digress.

What I was proposing is adding two more color choices in UI: for dark mode fg and bg colors — in addition to existing light mode fg and bg colors. Users then can change all four in one spot.

I don't think it will be easy, but I am willing at least to try, but only if it's not one of those features that won't be implemented, thus the question.

donho:
> What I was proposing is adding two more color choices in UI: for dark mode fg and bg colors — in addition to existing light mode fg and bg colors. Users then can change all four in one spot.

OK, I see your point now. It's not the solution because after switching into dark mode, user can switch to the theme they prefer, and the themes are not all dark mode, there are also in light mode. So this solution can only cover the default Dark mode theme.

ivan-u7n:
The details can be worked out. The sensible default is to use light colors in both modes, unless the dark ones are defined. Again this won't make all current themes magically dark mode aware. My proposal isn't about this. It is about giving more flexibility to theme and UDL developers by having the ability to specify distinct colors for light and dark modes and thus paving a way to create themes and UDLs which will be dark mode aware. On the UDL front already there is recently opened #10593 which isn't a proper bug report, but still…

To sum it up, as I don't see straight "no" from you, I'll look into it to see if my proposal can be implemented without changing too much code. Ofcourse I'll open a PR when I have something to look at.

donho:
>  It is about giving more flexibility to theme and UDL developers by having the ability to specify distinct colors for light and dark modes and thus paving a way to creating themes and UDLs which will be dark mode aware.

So if I do understand, UDL will double its size for both light theme and dark theme. But this solution won't cover every situation since the background color for both theme can vary. For example:

image

> To sum it up, as I don't see straight "no" from you, I'll look into it to see if my proposal can be implemented without changing too much code. Ofcourse I'll open a PR when I have something to look at.

If we double the size of UDL for the above result, it's a straight "No" of course.

ivan-u7n:
> If we double the size of UDL for the above result, it's a straight "No" of course.

Ofcourse not, [preliminary implementation](https://ci.appveyor.com/project/ivan-u7n/notepad-plus-plus/builds/40910518) shows the size increase of about 10% from 6.0 to 6.6 kilobytes for Markdown UDL.

It's still in the early stage, but it does show that what I proposed is possible. Moreover it is already switches between modes both for themes and for UDLs using a single style source.

Issue #4795, opened 2018-08-24
Close All needs verification pop up


jrheisey:
### Description of the Issue
When clicking the "Close All" toolbar button by accident it is very inconvenient to remember each file and where to find them in a large project. 
I had about 20 open files.

### Steps to Reproduce the Issue
1. Open a set of files
2. Click the "Close All" button on the toolbar
3. All files are immediately closed without any confirmation

### Expected Behavior
I expect a pop up dialog confirming to close all the tab views.

### Actual Behavior
Tabbed views are immediately closed

### Debug Information
N/A

donho:
Not like "Save All" which needs a warning popup because the operation modifies all the open files' content, "Close All" close all opened file without modification. A popup will be annoying for the most users.
Save your session for opening all your files so the problem solved.

PR #10572, opened 2021-09-19
for #4795 "enhancement"

ripmoneygone2020:
for Close All needs verification pop up #4795.

enhancement #4795 found that the Close All action didn't prompt the user to confirm the action. This will now prompt the user to confirm their intent to Close all files, but not if they are closing the application.

You can press the "ESC" key to cancel the action, in addtion to pressing the "no" and "cancel" buttons. Only clicking yes, will proceed and close all files.

added setting to tab group box under general preferences to allow the user to enable this behavior


donho:
This feature won't be implemented.

PR #10577, opened 2021-09-21
Clipboard text memory overrun fix

raspopov:
Fixed the memory overrun of the clipboard text copy operation caused by missed zero terminator at the end of data induced by some ill-behavior programs.

Fix for: #10590

Signed-off-by: raspopov <raspopov@cherubicsoft.com>


donho:
Please create an issue to describe the bug and how to reproduce it, and link the PR to the issue (Fix #XXXXX).

donho:
Cannot reproduced in v8.1.4

PR #10580, opened 2021-09-21
gcc: revamp makefile

ivan-u7n:
Notepad++'s makefile was revamped to be more modular and at the same time simpler. The main reason behind the changes — build errors caused by Scintilla when switching between compilers. Now Scintilla is also built in its own directory for each configuration. However the implementation is somewhat hackish. The patches to proper support separate build directories are sent upstream. When they are committed, the current not ideal implementation can be replaced with the proper one during Scintilla's update to the respective version.

Also build directories for Notepad++ are now one level deep instead of the hierarchy repeating the source one. But that's more a cosmetic change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,7 +151,7 @@ PowerEditor/bin/SourceCodePro-Regular.ttf
 #-------------------------------------------------------------------------------
 # MinGW-w64 GCC
 
-bin.*
+bin.*/
 scintilla/win32/*.o
 
 
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -18,45 +18,68 @@
 # definitions
 #
 
-GCC_DIRECTORY := ../gcc
-GCC_EXCLUDE := $(GCC_DIRECTORY)/gcc-%
-SRC_DIRECTORY := ../src
-SRC_EXCLUDE := $(SRC_DIRECTORY)/tools/%
-BIN_DIRECTORY := ../bin
-INSTALLER_DIRECTORY := ../installer
+ORIGIN_DIRECTORY = ..
+# will be properly defined later
+BUILD_DIRECTORY = .
+TARGET_DIRECTORY = .
 
-TARGET_BINARY := notepad++.exe
-SRC_DATA := contextMenu.xml langs.model.xml shortcuts.xml stylers.model.xml
-BIN_DATA := change.log doLocalConf.xml readme.txt userDefineLangs/
-INSTALLER_DATA := autoCompletion/ functionList/ localization/ themes/
+SOURCE_DATA =
+SOURCE_IGNORE =
+SOURCE_DATA += $(ORIGIN_DIRECTORY)/gcc=np
+SOURCE_IGNORE += $(ORIGIN_DIRECTORY)/gcc/gcc-%
+SOURCE_DATA += $(ORIGIN_DIRECTORY)/src=np
+SOURCE_IGNORE += $(ORIGIN_DIRECTORY)/src/tools/%
 
-SCINTILLA_DIRECTORY := ../../scintilla
-SCINTILLA_TARGET := libscilexer.a
+SCINTILLA_ROOT = $(ORIGIN_DIRECTORY)/../scintilla
+SCINTILLA_LINK = scilexer
+SCINTILLA_TARGET = $(BUILD_DIRECTORY)/lib$(SCINTILLA_LINK).a
+BUILD_DIRECTORY.SCINTILLA = $(SCINTILLA_ROOT)/$(BUILD_DIRECTORY)
 
-CXX := $(CROSS_COMPILE)g++
-CXXFLAGS := -include $(GCC_DIRECTORY)/gcc-fixes.h -std=c++17 -Wno-conversion-null
-RC := $(CROSS_COMPILE)windres
-RCFLAGS :=
-CPP_PATH := $(SCINTILLA_DIRECTORY)/include
-CPP_DEFINE := UNICODE _UNICODE _WIN32_WINNT=0x0600 TIXML_USE_STL TIXMLA_USE_STL
-LD := $(CXX)
-LDFLAGS := -municode -mwindows
-LD_PATH :=
-LD_LINK := comctl32 crypt32 dbghelp ole32 sensapi shlwapi uuid uxtheme version wininet wintrust
-LD_LINK += $(patsubst lib%.a,%,$(SCINTILLA_TARGET)) imm32 msimg32 ole32 oleaut32
-SUBMAKEFLAGS := -O --no-print-directory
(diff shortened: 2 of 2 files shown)


donho (changes requested):
I recommand to separate & breakdown the boostregex/BoostRegExSearch.cxx warning part & BUILD.md into different PR. As contribute guideline mentioned:
*In short: The easier the code review is, the better the chance your pull request will get accepted.*

ivan-u7n:
@donho
> I recommand to separate & breakdown the boostregex/BoostRegExSearch.cxx warning part & BUILD.md into different PR.

Am I right to understand that you wish to leave only makefile's revamp in this PR and for me to create two new ones:
1. for fixing warnings;
2. for updating BUILD.md?

> Could you please explain all the foreach parts?

Three `foreach`es in `source-items` are for:
1. going over all directories specified in `SOURCE_DATA`: paths where to look for sources and headers — into this list I previously tried to add several Scintilla's top-level directories;
2. going over all patterns for files to filter — this part exists because `patsubst` in `3.` operates only on a single pattern;
3. going over all files matching `2.` in `1.` — this is the main loop setting up relationship between object and source files.

However now thinking about it, I think I can eliminate `2.` as there is no need to replace file extensions, just appending the required one will also do.

PS. Have you seen my comment https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10571#issuecomment-925736584 about an approach to handling dark mode styles?

donho:
> Am I right to understand that you wish to leave only makefile's revamp in this PR and for me to create two new ones:
> 
>     1. for fixing warnings;
> 
>     2. for updating BUILD.md?

Yes, please.

ivan-u7n:
Sorry for all the noise, I tried to rename the branch, but alas — GitHub doesn't properly apply it to referencing PRs.

> Yes, with your modification we could have the same result. However, it makes sense to me semantically the original code to check the index until to the beginning, then reset -1 to 0. So please keep it as it is.

The reason I changed this line:
```c++
        while (startPos >= 0 && styler.StyleAt(--startPos) == initStyle);
```
is that `startPos >= 0` is always true for unsigned types (which `Sci_PositionU` is) and the loop terminates only thanks to the second condition becoming false — in case of out of bounds requests `Accessor::StyleAt()` returns `SCE_USER_STYLE_DEFAULT` (default-constructed `char`) which `initStyle` thankfully can't be equal to, otherwise the loop will turn into infinite.

Nonetheless, having even a potential infinite loop doesn't sound good to me. If it's still OK with you, I'll leave this line unchanged.

donho:
> The reason I changed this line:
> 
> ```c++
>         while (startPos >= 0 && styler.StyleAt(--startPos) == initStyle);
> ```
> 
> is that `startPos >= 0` is always true for unsigned types (which `Sci_PositionU` is) and the loop terminates only thanks to the second condition becoming false — in case of out of bounds requests `Accessor::StyleAt()` returns `SCE_USER_STYLE_DEFAULT` (default-constructed `char`) which `initStyle` thankfully can't be equal to, otherwise the loop will turn into infinite.
> 
> Nonetheless, having even a potential infinite loop doesn't sound good to me. If it's still OK with you, I'll leave this line unchanged.

Yes, you're right about it.
OK for the change, please do it in another PR.

ivan-u7n:
> OK for the change, please do it in another PR.

Everything is done:
* this PR is now only about the makefile;
* there are 3 more PRs:
1. updating `BUILD.md`;
2. fixing warnings during Scintilla build;
3. eliminating a potential infinite loop in `scintilla/lexers/LexUser.cxx`.

donho (approved):

donho:
@ivan-u7n 
Could we separate the copy necessary files script from the build (compile & link) makefile?

ivan-u7n:
@donho 
> Could we separate the copy necessary files script from the build (compile & link) makefile?

As some files are still required to be copied into the resulting directory, I propose to make a goal `optional` in the current makefile which when called like `mingw32-make optional` will copy everything optional.

However all the files copied aren't big, and it's good to gather everything scattered in the source tree in one directory where freshly compiled Notepad++ can utilize it all. Thus I'd leave the coping as is, especially as the destination is the single directory that can be easily deleted (or moved if something akin portable version is needed).

donho:
makefile for me is for build purpose (compile & link), ie. for generating of executable binary only.
For the sake of maintenance, all the purpose beyond the build should be in a script. This script can be called by makefile, but be out of makefile.

donho:
@ivan-u7n 
Is it considerable to you to separate the copy files part from makefile ?

ivan-u7n:
@donho Sorry, got overrun by tasks at work.

To me a makefile is basically just a tool for automation. And I consider the copying to be a part of the build process, like so many pieces of open source software build themselves with `make` and then are installed via `make install`.

So my proposal is the following: let `mingw32-make` build Notepad++ and copy only essential for it to start files into the target directory while `mingw32-make install` will copy all the extra useful files scattered over the source tree into the same target directory.

Also, in my opinion, separating the copying part from the makefile will make the maintenance harder, because, as I said earlier, some copying is still required to occur during the build process, and the extraction will lead to two files in need to be kept in sync over what is copied by which one.

donho:
> So my proposal is the following: let mingw32-make build Notepad++ and copy only files essential for it to start into the target directory while mingw32-make install will copy all the extra useful files scattered over the source tree into the same target directory.

Yes, that sounds great.

> Also, in my opinion, separating the copying part from the makefile will make the maintenance harder, because, as I said earlier, some copying is still required to occur during the build process, and the extraction will lead to two files in need to be kept in sync over what is copied by which one.

Sorry, I didn't make myself clear. What I meant is, when an user type `mingw32-make` (or `mingw32-make install`), all the needed files will be copied, but it's done by another script, which is triggered by `makefile`. It'll make the current `makefile` even simpler. What do you think?

ivan-u7n:
> Yes, that sounds great.

I'll make the respective change to this PR hopefully next week.

> Sorry, I didn't make myself clear. What I meant is, when an user type `mingw32-make` (or `mingw32-make install`), all the needed files will be copied, but it's done by another script, which is triggered by `makefile`. It'll make the current `makefile` even simpler. What do you think?

The problem with this approach is the duplication of effort as there will be needed at least two scripts: one for the Windows default `cmd` and another one for a UNIX-like shell akin MSYS2 on Windows and Bash on Linux/macOS. However GNU Make interprets its makefiles the same way no matter the system it runs on, an thus the current approach just works with minor tweaks in both environments. (Although I haven't yet tested the makefile in anything but `cmd`, but plan to at least with MSYS2 in the near future.)

To sum it up, I agree that moving out the copying part will make the current makefile simpler, but not the overall maintenance burden. For me the cons overweight the pros.

donho:
This PR is closed due to the lake of reply.

PR #10594, opened 2021-09-25
gcc: update `BUILD.md`

ivan-u7n:
Updated `BUILD.md` to document `clean` and `fullclean` makefile goals with rewording of other parts.

Should be applied together with the makefile changes as the explicit mention of the problem solved by those changes got converted into general description of the `clean` goal.


donho:
This PR is closed due to the lake of reply.

PR #10596, opened 2021-09-25
gcc: fix warnings during Scintilla build

ivan-u7n:
Fixed warnings generated during Scintilla build in `boostregex/BoostRegExSearch.cxx` and in `scintilla/lexers/LexUser.css` (`-Wall -Wpedantic` are the default for Scintilla). Fixing warnings in the latter file was done by removing the check that was always false.

### DIFF
--- a/boostregex/BoostRegExSearch.cxx
+++ b/boostregex/BoostRegExSearch.cxx
@@ -72,7 +72,7 @@ private:
 			set(m._document, m.position(), m.endPosition());
 			return *this;
 		}
-		Match& operator=(int /*nullptr*/) {
+		Match& operator=(void* /*nullptr*/) {
 			_position = -1;
 			return *this;
 		}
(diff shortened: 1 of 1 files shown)


scintilla/lexers/LexUser.cxx
```diff
@@ -808,8 +808,6 @@ static inline void ReColoringCheck(Sci_PositionU & startPos, int & nestedLevel,
         do
         {
             ch = styler.SafeGetCharAt(--startPos);
-            if (startPos == -1)
-                startPos = 0;
```
donho, line 812: The deletion of these 2 lines shouldn't be in this PR, should it?

ivan-u7n: It should as it is the fix for a warning:
> Fixing warnings in the latter file was done by removing the check that was always false.

donho: Yes, I know.
My point is : there's no impact to remove these 2 lines in this PR, whereas there are more modification above these deleted lines in  PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10597. Both modifications in the same file but not in the same PR could make regression due to the separated review.

donho (changes requested):
Move the change in `LexUser.cxx` to the PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10597

donho (approved):

PR #10597, opened 2021-09-25
UDL lexer: fix potential infinite loop

ivan-u7n:
When fixing warnings in `scintilla/lexers/LexUser.css`, in a few lines above the spot I noticed and fixed a condition in a loop which is always true (unsigned >= 0) and can potentially turn the loop infinite.

And then I noticed and removed a check after the loop which now is also always false.

### DIFF
--- a/scintilla/lexers/LexUser.cxx
+++ b/scintilla/lexers/LexUser.cxx
@@ -786,20 +786,20 @@ static inline void ReColoringCheck(Sci_PositionU & startPos, int & nestedLevel, 
                                    int & isCommentLine, bool & isInComment, Accessor & styler, vector<nestedInfo> & lastNestedGroup,
                                    vector<nestedInfo> & nestedVector, /* vector<int> & foldVector, */ int & continueCommentBlock)
 {
-    // re-coloring always starts at line beginning !!
-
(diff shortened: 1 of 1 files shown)


chcg:
Seems the line

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/lexers/LexUser.cxx#L792

need a check for startPos > 0, otherwise the style at unsigned (-1) is checked and might return something undefined.

ivan-u7n:
I haven't touched it because `StyleAt()` for out of bounds requests returns default-constructed `char`, e.g. `SCE_USER_STYLE_DEFAULT`. Although I can add the check — it isn't hard.

ivan-u7n:
Pulled the whole loop under the `startPos > 0` check for the sake of sanity.

chcg:
I think by:

           while (startPos > 0 && styler.StyleAt(--startPos) == initStyle);
 
startPos is modified and could be 0 afterwards. So the startPos > 0 is still needed afterwards to avoid a wraparound within the next do-while-loop.

ivan-u7n:
Nicely spotted, thanks! Inserted the check between the blocks.

PR #10598, opened 2021-09-25
Win32file api

pnedev:
Use native Win32 API to access files (currently only used for saving files to fix the reported file corruption problems).
Settings files are still written using TinyXML library and are thus still prone to those problems.

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.h
+++ b/PowerEditor/src/MISC/Common/FileInterface.h
@@ -1,0 +1,64 @@
+// This file is part of Notepad++ project
+// Copyright (C)2021 Don HO <don.h@free.fr>
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// at your option any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+#pragma once
+
+#include <windows.h>
+#include <tchar.h>
+#include <cstdint>
+
+
+class CFile
+{
+public:
+	enum class Mode
+	{
+		READ,
+		WRITE,
+		APPEND
+	};
+
+	CFile(const char *fname, Mode fmode = Mode::READ);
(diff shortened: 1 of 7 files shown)


PowerEditor/src/MISC/Common/FileInterface.h
```diff
@@ -0,0 +1,64 @@
...
+	{
+		return (_hFile != INVALID_HANDLE_VALUE);
+	}
+
+	void Close();
+
+	int_fast64_t GetSize();
+
+	unsigned long Read(void *rbuf, unsigned long buf_size);
+	bool Write(const void *wbuf, unsigned long buf_size);
```
ivan-u7n, line 53: Why not the standard `size_t` for the sizes? I'm really curious — maybe I don't know something…

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -868,15 +869,14 @@ bool FileManager::backupCurrentBuffer()
...
 			}
 
-			FILE *fp = UnicodeConvertor.fopen(fullpath, TEXT("wbc"));
-			if (fp)
+			if (UnicodeConvertor.fopen(fullpath))
 			{
 				int lengthDoc = _pNotepadPlus->_pEditView->getCurrentDocLen();
 				char* buf = (char*)_pNotepadPlus->_pEditView->execute(SCI_GETCHARACTERPOINTER);	//to get characters directly from Scintilla buffer
-				size_t items_written = 0;
+				long items_written = 0;
```
ivan-u7n, line 876: …as then it seems this switching from `size_t` to `long` and later casts won't be necessary.

donho (changes requested):
Please change the code according the following remarks:
 
1. Coding convention: Change the 1st letter of methods of CFile class to lowercase.
2. Make `Utf8_16_Write::fwrite()` return `bool`.
3. Change `long items_written = 0;` to `boolean isWrittenSuccessful = false;`.

PowerEditor/src/MISC/Common/FileInterface.h
ivan-u7n: @donho What I meant is that it seems more reasonable for these member functions to have `buf_size` or return type as `size_t` to concentrate the necessary casts or other adjustments in one spot than to spread them all over the code.

Also on a 64 bit system this code will fail to write a buffer over 4GB in size, although it's unlikely to happen, but still…

pnedev:
@donho,

Sorry, I don't have time to work or comment on that anymore.
Feel free to modify, accept, decline this PR as you find appropriate.

donho:
> Sorry, I don't have time to work or comment on that anymore.
> Feel free to modify, accept, decline this PR as you find appropriate.

OK, no problem. 
Thank you for your PR and all your effort concerning this issue.

donho (approved):

PR #10600, opened 2021-09-27
RFC: basic support for additional dark mode colors in themes and UDLs

ivan-u7n:
The support is basic because, while all four colors (background and foreground for light and dark modes) are stored independently, the UI (Styles Configurator and Define Your Language dialogs) doesn't yet have controls to change all four colors independently but instead operates on the pair specific for the current mode.

The UI can be improved at a later date using this RFC as the basis. Or may be exactly what is needed and it ought to be merged as it is. All said, my current goal is to get some feedback and to see if this idea even meets some demand.

In case a theme or a UDL doesn't have dark colors defined, the light ones will be copied to take the vacant spots while loading styles from a file. There is also a check to prevent color duplication while writing styles into a file. As it's just a copy, there is no impact on further manipulations via the UI.

The dark mode Markdown UDL was folded into the unified one. The default dark mode theme was folded into the unified default theme. Or to be precise, in the latter case the default dark mode theme was used as the basis and light mode colors were added into it. Also, the current default theme and bundled Markdown UDL files are the same as the ones produced by this version of Notepad++.


donho:
The solution of this PR doesn't cover the following case:

image

ivan-u7n:
@donho 
> The solution of this PR doesn't cover the following case:

It wasn't supposed to. It was supposed to cover only the basic usage of the default theme and the preinstalled Markdown UDL with possibilities for the future dark-mode-aware themes and UDLs to appear.

However, now that I think about it, for the case above to work properly, the basic style information for UDLs must be provided by the theme. And then UDLs having only the format definitions without styles or with minimal overrides will work with any theme. I can look into it if there is a demand, but this will be a rather breaking change as at least external themes and UDLs will for sure require the overhaul.

donho:
> It wasn't supposed to. It was supposed to cover only the basic usage of the default theme and the preinstalled Markdown UDL with possibilities for the future dark-mode-aware themes and UDLs to appear.

However, it's what I previewed:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10571#issuecomment-927204416

> However, now that I think about it, for the case above to work properly, the basic style information for UDLs must be provided by the theme.

It's already done. Open Document List or Folder as Workspace Panel, and change the theme in Styler Configurator - you'll see the bg color changes in the panel.

>  And then UDLs having only the format definitions without styles or with minimal overrides will work with any theme.

Yes, that's what I intended to try: override the bg color in UDL file and it's even no need to add other information in UDL file. (maybe just only one propriety optional `bgColorCanBeOverriden="no"` to make defined bg color in the tag persist if overriding is needed) 

OTOH, it's about UDL enhancement. The modification of styler.xml's structure is surely a straight no.

donho:
@ivan-u7n 
> Yes, that's what I intended to try: override the bg color in UDL file and it's even no need to add other information in UDL file. (maybe just only one propriety optional bgColorCanBeOverriden="no" to make defined bg color in the tag persist if overriding is needed)

In fact, this feature is already done since long time ago (not only for UDL, but for the global styler): https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10619

`colorStyle` with its possible value: 1 (`COLORSTYLE_FOREGROUND`), 2 (`COLORSTYLE_BACKGROUND`) or 3 (`COLORSTYLE_FOREGROUND | COLORSTYLE_BACKGROUND`) (see: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Parameters.h#L377) make bg or fg be ignored, but by default both are used. This feature is not well known by the most of users I guess, and even forgotten by me (who have implemented it).

Sorry for your wasted time & thank you for investing your time in it.

PR #10611, opened 2021-09-30
Fix incorrect text size calculation for drawing the dark mode status bar which can cause an invalid character to be displayed.

### LINKED ISSUE #10564: "Empty box" character in status bar ONLY in dark mode.
I use Windows 10 x64 + NPP v8.1.4
I'm attaching an image showing the problem I'm experiencing.
Apparently, from the screenshots I've seen from others, it's not a common issue. However, I can reproduce it easily:

- Install NPP
- Defaults to light mode > no problem
- **Change to dark mode > problem**
- Change to light mode > no problem

Might be related to the font NPP uses for its UI? Could I change it to test if that's the case?

Compare

adzm:
Fixes #10564


donho:
@adzm It makes sens your fix. However, I cannot reproduce the bug.
I will ask the author of this issue for testing.

PR #10612, opened 2021-09-30
Fix previous session lost issue after MS update restarting Windows

### LINKED ISSUE #10402: NPP opens with a Blank Session, after Win10 Automatic Restart following Update
### Description of the Issue
NPP opens with a Blank Session, after Win10 Automatic Restart following Update.

It is very difficult to stop Win 10 from restarting by itself following an Update, during Inactive hours. The Internet is full of people searching on tips to avoid that.

What happens with NPP is that (it doesn't happen every time, but about once in 3 times), it does not safely save and close the Session.xml file.

I believe that when NPP is re-launched, it finds the Session.xml file corrupted, and over-writes it with a Blank Session file, and opens in a Blank Session.

It becomes a nag, because we have so many text files we are working on, arranged in a particular order, and working between them, as a Writer.
It's very difficult to recall which files were open and in which order, and then to get back to a semblance of where we where, we have to spend nearly half a day till we are somewhere comfortable again.

I had reported on this earlier, along with a Null file issue I had with unsaved files.

I had suggested for NPP to look at having a set of Time-stamped Session files, say about the latest 10 at all times, at whatever time interval you'll feel optimal.
At every new save, the oldest one would be removed.

So, for whatever reason, NPP crashes, or is force-closed before it can save and exit properly, which causes the Session.xml file to get corrupt.

NPP at next re-launch could check the integrity of the Session.xml file, attempt repair, and if not possible, load with the previous Session.xml file, and alert the user that this has happened and it has taken this action...

I am sure users would bless you for this, and feel so so much more secure using NPP.

Thanks a load for your wonderful program, and for all your support. 

### Debug Information
Notepad++ v8.1.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1252
Plugins : AutoSave.dll ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll SessionMgr.dll

donho:
Use Windows IO API instaed POSIX IO (output only) on tinyXml component for writting xml file (config.xml, session.xml, etc...).

Fix #10402

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.h
+++ b/PowerEditor/src/MISC/Common/FileInterface.h
@@ -51,6 +51,10 @@ public:
 	unsigned long read(void *rbuf, unsigned long buf_size);
 	bool write(const void *wbuf, unsigned long buf_size);
 
+	bool writeStr(const char* str) {
+		return write(str, static_cast<unsigned long>(strlen(str)));
+	};
+
 private:
 	HANDLE	_hFile		{INVALID_HANDLE_VALUE};
 	Mode	_hMode		{Mode::READ};
--- a/PowerEditor/src/MISC/Common/FileInterface.h
+++ b/PowerEditor/src/MISC/Common/FileInterface.h
@@ -22,43 +22,39 @@
 #include <cstdint>
 
 
-class CFile
+class Win32_IO_File final
 {
 public:
-	enum class Mode
-	{
+	enum class Mode {
 		READ,
 		WRITE,
 		APPEND
 	};
 
-	CFile(const char *fname, Mode fmode = Mode::READ);
-	CFile(const wchar_t *fname, Mode fmode = Mode::READ);
+	Win32_IO_File(const char *fname, Mode fmode = Mode::READ);
+	Win32_IO_File(const wchar_t *fname, Mode fmode = Mode::READ);
 
-	~CFile()
-	{
-		Close();
-	}
+	Win32_IO_File() = delete;
+	Win32_IO_File(const Win32_IO_File&) = delete;
+	Win32_IO_File& operator=(const Win32_IO_File&) = delete;
 
-	bool IsOpened()
-	{
+	~Win32_IO_File() {
+		close();
+	};
+
+	bool isOpened() {
 		return (_hFile != INVALID_HANDLE_VALUE);
-	}
+	};
 
-	void Close();
-
-	int_fast64_t GetSize();
-
-	unsigned long Read(void *rbuf, unsigned long buf_size);
-	bool Write(const void *wbuf, unsigned long buf_size);
+	void close();
+	int_fast64_t getSize();
+	unsigned long read(void *rbuf, unsigned long buf_size);
+	bool write(const void *wbuf, unsigned long buf_size);
 
 private:
-	CFile(const CFile&) = delete;
-	CFile& operator=(const CFile&) = delete;
-
-	void fillCreateParams(DWORD &access, DWORD &share, DWORD &disp, DWORD &attrib);
-
 	HANDLE	_hFile		{INVALID_HANDLE_VALUE};
 	Mode	_hMode		{Mode::READ};
 	bool	_written	{false};
+
+	void fillCreateParams(DWORD& access, DWORD& share, DWORD& disp, DWORD& attrib);
 };
--- a/PowerEditor/src/MISC/Common/FileInterface.h
+++ b/PowerEditor/src/MISC/Common/FileInterface.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <windows.h>
+#include <string>
 #include <tchar.h>
 #include <cstdint>
 
@@ -51,8 +52,8 @@ public:
 	unsigned long read(void *rbuf, unsigned long buf_size);
 	bool write(const void *wbuf, unsigned long buf_size);
 
-	bool writeStr(const char* str) {
-		return write(str, static_cast<unsigned long>(strlen(str)));
+	bool writeStr(const std::string& str) {
+		return write(str.c_str(), static_cast<unsigned long>(str.length()));
 	};
 
 private:
--- a/PowerEditor/src/MISC/Common/FileInterface.h
+++ b/PowerEditor/src/MISC/Common/FileInterface.h
@@ -1,5 +1,5 @@
 // This file is part of Notepad++ project
-// Copyright (C)2021 Don HO <don.h@free.fr>
+// Copyright (C)2021 Pavel Nedev (pg.nedev@gmail.com)
 
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
--- a/PowerEditor/src/TinyXml/tinyxml.cpp
+++ b/PowerEditor/src/TinyXml/tinyxml.cpp
@@ -562,24 +562,24 @@ const generic_string tagCloseSymb = TEXT(">");
 const char tagCloseChar[] = ">";
 const char tagCloseSymbFinal[] = " />";
 
-void TiXmlElement::Print( CFile& cfile, int depth ) const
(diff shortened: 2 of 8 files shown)


PowerEditor/src/TinyXml/tinyxml.cpp
```diff
@@ -551,20 +551,37 @@ void TiXmlElement::SetAttribute( const TCHAR * name, const TCHAR * _value )
...
+const char tagCloseSymbFinal[] = " />";
+
+void TiXmlElement::Print( CFile& cfile, int depth ) const
 {
 	int i;
 	for ( i=0; i<depth; i++ )
 	{
-		generic_fprintf( cfile, TEXT("    ") );
+		cfile.Write(four_ws, 4);
+		//generic_fprintf( cfile, TEXT("    ") );
```
mere-human, line 571: It's better to remove code rather than comment it (here and below). You can always find it in Git history.

PowerEditor/src/TinyXml/tinyxml.cpp
```diff
@@ -551,20 +551,37 @@ void TiXmlElement::SetAttribute( const TCHAR * name, const TCHAR * _value )
...
+const char tagCloseChar[] = ">";
+const char tagCloseSymbFinal[] = " />";
+
+void TiXmlElement::Print( CFile& cfile, int depth ) const
 {
 	int i;
 	for ( i=0; i<depth; i++ )
 	{
-		generic_fprintf( cfile, TEXT("    ") );
+		cfile.Write(four_ws, 4);
```
mere-human, line 570: Instead of using the magic number "4" you can use `std::size(four_ws)`.
Also, you can add an overloaded method to CFile that accepts only a string and does `strlen` on it to determine the size.

By the way, I'm not sure using a named constant improves the readability of code here.
Here we have simple and short string literals that are easy to understand.
Also,  reducing such modifications could help merging with a newer version of TinyXML library.

PowerEditor/src/TinyXml/tinyxml.cpp
```diff
@@ -551,20 +551,37 @@ void TiXmlElement::SetAttribute( const TCHAR * name, const TCHAR * _value )
...
+		//generic_fprintf( cfile, TEXT("    ") );
 	}
 
-	generic_fprintf( cfile, TEXT("<%ls"), value.c_str() );
+	generic_string tagOpenValue = tagOpenSymb + value;
+	std::string tagOpenValueA = wstring2string(tagOpenValue, CP_UTF8);
+	cfile.Write(tagOpenValueA.c_str(), static_cast<unsigned long>(tagOpenValueA.length()));
```
mere-human, line 576: I'm not sure you have to convert the string to ASCII.
Since, without the change `fwprintf` was used.

PowerEditor/src/TinyXml/tinyxml.cpp
donho: It's for the future debug need, especially for such word to word modification.
So I prefer to keep it.

PowerEditor/src/TinyXml/tinyxml.cpp
donho: TinyXML libary in Notepad++ hasn't been updated since long time ago, and won't be updated in the future. There's no bug in it from the usage of Notepad++, so why bother?

cfile.Write needs a pointer of char and `cfile.Write("    ", 4);` won't work, so a named variable cannot be avoid. Since this 4 ws value is used in the different place, I prefer to declare it as constant outside of this function. Use `4` instead of std::size(four_ws) is to prevent from using cast. They are not magic number for these trivial `const char*`.

PowerEditor/src/TinyXml/tinyxml.cpp
donho: https://stackoverflow.com/questions/28618715/c-writefile-unicode-characters

PowerEditor/src/TinyXml/tinyxml.cpp
mere-human: You can add an overload `bool Write(const char* wbuf, size_t buf_size);`

pnedev: I see @mere-human has proposed the same thing as me here: https://github.com/donho/notepad-plus-plus/commit/a8ec068965b2a0eed010b53d88a18d14bb678e58#commitcomment-57345690 - to make an overloaded Write method of the CFile class that accepts directly a `std::string` - will make things much clearer.

chcg: Anyhow there is no possibility to update, see https://sourceforge.net/projects/tinyxml, apart from changing to https://github.com/leethomason/tinyxml2, which uses UTF-8 and probably needs to be adapted for windows utf-16.

PowerEditor/src/MISC/Common/FileInterface.h
```diff
@@ -51,6 +51,10 @@ class Win32_IO_File final
 	unsigned long read(void *rbuf, unsigned long buf_size);
 	bool write(const void *wbuf, unsigned long buf_size);
 
+	bool writeStr(const char* str) {
+		return write(str, static_cast<unsigned long>(strlen(str)));
+	};
+
```
pnedev, line 57: @donho ,

That method is a good addition but IMO it is good to also add another `write()` that accepts `const std::string& str` as parameter directly and use `str.length()` as size inside.
Using `strlen()` on `string.c_str()` is not optimal because `strlen()` will cycle through all string chars to find the terminating `\0` and find the length and that is just wasting time since `std::string` method `length()` directly has the length. The bigger the string is the better this optimization will be.

PowerEditor/src/TinyXml/tinyxml.cpp
```diff
@@ -567,19 +567,19 @@ void TiXmlElement::Print( Win32_IO_File& cfile, int depth ) const
...
+		cfile.writeStr(four_ws);
 		//generic_fprintf( cfile, TEXT("    ") );
 	}
 
-	generic_string tagOpenValue = tagOpenSymb + value;
-	std::string tagOpenValueA = wstring2string(tagOpenValue, CP_UTF8);
-	cfile.write(tagOpenValueA.c_str(), static_cast<unsigned long>(tagOpenValueA.length()));
+	std::string tagOpenValue = tagOpenSymb;
+	tagOpenValue += wstring2string(value, CP_UTF8);
+	cfile.writeStr(tagOpenValue.c_str());
```
pnedev, line 576: Look my suggestion above.
In case you implement `write(const std::string& str)` method you'll simply do `cfile.writeStr(tagOpenValue);` here and inside `tagOpenValue.length()` will be used instead of `strlen()`. Thus no time will be wasted to loop through all the string chars.
This is valid for all `std::string` `write()` invocations below as well.

PowerEditor/src/TinyXml/tinyxml.cpp
```diff
@@ -555,9 +555,9 @@ const char one_ws[] = " ";
 const char four_ws[] = "    ";
 const char eol[] = "\r\n";
 
-const generic_string tagOpenSymb = TEXT("<");
-const generic_string tagOpenSlash = TEXT("</");
-const generic_string tagCloseSymb = TEXT(">");
+const char tagOpenSymb[] = "<";
+const char tagOpenSlash[] = "</";
+const char tagCloseSymb[] = ">";
```
pnedev, line 560: If you add `write(const std::string& str)` method as suggested above maybe it will be a good idea to make all `const char` values here to be `const std::string`. This will remove the `strlen()` call when writing them to the file from every place where that happens (and it seems it will happen a lot).

donho: It'll be even better to remove them.
For the sake of readability like @mere-human has mentioned, I can reintegrate these strings into code thanks to `writeStr(const std::string& str)`.

PR #10604, opened 2021-09-28
Update swedish.xml

Regnander:
In this PR I...
* added the untranslated lines from the English file
* reworded the warning/error dialog messages to make the text flow better in Swedish
  * for example "Continue?" -> "Do you want to continue?"
* replaced the TAB indents with regular spaces to match the format of the English file
* trimmed some existing lines where the UI was truncating text
* made sure each menu item has a unique [mnemonic](https://en.wikipedia.org/wiki/Mnemonics_(keyboard)) in the `File`, `Edit` and `Search` menus
* used [Microsoft's Language Portal](https://www.microsoft.com/en-us/language) to make sure the translation is using recognizable and consistent terminology for the more technical terms

I should also mention that I took some liberties translating the color picker labels in `Preferences` > `Dark Mode` since they won't make any sense in Swedish when translated word-for-word. Some examples are "Top" = "Primary window color", "Menu hot track" = "Hover color" and "Active" = "Active elements".

PR #10620, opened 2021-10-04
Update swedish.xml

Regnander:
Due to numerous failed attempts to push a few commits to my existing PR (#10604) in order to fix the indents, I decided to make a new PR to not mess up thing even more.

Here is the original description, although slightly edited to reflect the new commits:
>In this PR I...
>* added the untranslated lines from the English file
>* reworded the warning/error dialog messages to make the text flow better in Swedish
>  * for example "Continue?" -> "Do you want to continue?"
>* ~replaced the TAB indents with regular spaces to match the format of the English file~
>* trimmed some existing lines where the UI was truncating text
>* made sure each menu item has a unique [mnemonic](https://en.wikipedia.org/wiki/Mnemonics_(keyboard)) >in the `File`, `Edit` and `Search` menus
>* used [Microsoft's Language Portal](https://www.microsoft.com/en-us/language) to make sure the >translation is using recognizable and consistent terminology for the more technical terms
>
>I should also mention that I took some liberties translating the color picker labels in `Preferences` > `Dark >Mode` since they won't make any sense in Swedish when translated word-for-word. Some examples are >"Top" = "Primary window color", "Menu hot track" = "Hover color" and "Active" = "Active elements".

### DIFF
--- a/PowerEditor/installer/nativeLang/swedish.xml
+++ b/PowerEditor/installer/nativeLang/swedish.xml
@@ -373,7 +373,7 @@ The comments are here for explanation, it's not necessary to translate them.
                 <Item CMID="10" name="Byt namn på fil..."/>
                 <Item CMID="11" name="Flytta till papperskorgen"/>
                 <Item CMID="12" name="Skrivskydda"/>
-                <Item CMID="13" name="Inaktivera skrivskyddet"/>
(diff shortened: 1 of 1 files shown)


Regnander:
The whole merge shenanigans was a desperate attempt on my end to make the commits appear in the previous PR. 😞

Issue #10593, opened 2021-09-24
Editing .md files in dark mode


timint:
Editing .md files in dark mode is a challenge. Maybe someone has a fix for that?
I thought I would bring this up as it's been this way for many versions now.

![image](https://user-images.githubusercontent.com/359192/134749469-eb6b3ac8-8b9a-4afa-ab2d-16abf7b6ce8c.png)

timint:
And so I discovered a separate Language setting for dark mode markdown which feels half baked. Is it not possible to have it honouor the selected style theme instead?

![image](https://user-images.githubusercontent.com/359192/134749774-700787bd-b4dd-471c-83c1-732166b644e4.png)

thomassmith311:
Those Markdown settings are actually "User Defined Languages" created with Edditoria: https://github.com/Edditoria/markdown-plus-plus.

You'll need to go to that link and follow the instructions.
> 
    Download the source code in latest release page. It should be a zip file.
    Open the zip file and go to <udl\> folder.
    Copy a XML file of your favorite theme, and paste in <userDefineLangs\> folder of Notepad++.
        (The directory is usually <%AppData%\Notepad++\userDefineLangs\>.)
    Restart Notepad++.
    Open and test with a Markdown file e.g. test/at-a-glance.md.

Copy the XML file of whatever theme you're using (looks look Obsidian?) into the userDefineLangs folder, close and reopen N++, and it should be there. That link also tells you how to create custom MD files for your own custom themes.

timint:
I want to thank you for the time and effort spent in replying to me. I see the markdown support is a complete side project.

Indeed it is Obsidian. Well spotted. 😀

I copied `markdown-plus-plus-master.zip\udl\markdown.obsidian.udl.xml` to `%appdata%\Notepad++\userDefineLangs\` and that solved the problem, thank you. Is this the intended solution by design? Or is this just a workaround due to framework limitations?

Spontaneously I would have thought the most optimal implementation would be to add it as a Language like any other languages built in. Yeah, markdown is not a language, but seeing YAML in the list makes me think why not? But then I am a web developer and cpp is just not my thing. Maybe @Edditoria have some good input on this?

rdipardo:
> I would have thought the most optimal implementation would be to add it as a Language like any other languages built in.

You're not alone; to quote from this [this related thread](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5600#issuecomment-609781469):

[...]

PR #10621, opened 2021-10-05
Replacing markdown udls with unified udl

thomassmith311:
I am replacing the two markdown udl files with a unified file that should, ideally, work with all themes. I've tested my udl file on all the various themes included in Notepad++, and there are no issues I discovered.


thomassmith311:
This is in connection to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10593

CennoxX:
If switching back to a unified udl, the changes of this commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4cc3d9fd1a5d4eeb7be093729a8885d3f049b622 should also be reverted.

But first, there is an issue with the changes above: the colored italic / bold texts are gone:
old:
grafik
new:
grafik

thomassmith311:
Right. I don't know how to fix the fact that the color is switched back to the plain color, rather than that sepia color. Your pull request makes the color blue in a light theme but sepia in a dark theme - which is great, but that, of course, depends on having two different udl files. Having the sepia color in a light theme makes it impossible to read.

How much does it matter if it does not have the color any longer?

donho:
@thomassmith311 
What's the advantage to merge these 2 UDL files in one file?

CennoxX:
As mentioned above, if the two udl's are unified the changes of the commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4cc3d9fd1a5d4eeb7be093729a8885d3f049b622 where the udl was seperated should also be reverted.

thomassmith311:
@donho Could it be a good idea to move the Markdown option up the dropdown under "M", along with Makefile and Matlab, etc., rather than having it be a user-defined language? I'm just not sure how to do that.

Edditoria:
@donho Please consider to upgrade UDL system that **overrides the background color** of the document space, rather than individual character only. I think it can quickly solve the existing UI problem. And it is urgent because dark mode is a popular topic.

Btw, the existing features of Markdown display is limited by UDL. For example:

- 8 delimiter styles are not enough.
- Tree-based nesting styles that is limited on 1 delimiter (example), rather than the child style will affect the whole document.
- Only 1 style for operators, comments, and others.

It is not easy and demand may be low for other languages, tho.

thomassmith311:
> @thomassmith311
> 
> > Could it be a good idea to move the Markdown option up the dropdown under "M", along with Makefile and Matlab, etc., rather than having it be a user-defined language? I'm just not sure how to do that.
> 
> If you mean in the menu `Languages`, it'll be quite difficult to implement it, since Markdown is supported as UDL but not native.

Yeah, I thought so. I don't know, it just feels a little weird to have this one language sitting down on the bottom.

PR #10730, opened 2021-10-29
Update german.xml with mainly the date time translations

MarkusBodensee:
Add all missing date time translations from v8.1.5.
Update 42083 and 42078 to not use the word sort (sortieren) because english.xml doesn't either.
Move items 41025 and 44109 to the same order like in english.xml for more easy diffing.
Delete obsolete/doubled item 42079 (it was there two times in german.xml).

See my comments in the last update of german.xml at #10428. Review welcome.

@schnurlos Feel free to take these changes over to your next update. Just wanted to create a PR to have the translation in a next release.

### DIFF
--- a/PowerEditor/installer/nativeLang/german.xml
+++ b/PowerEditor/installer/nativeLang/german.xml
@@ -2,7 +2,7 @@
 
 <!--
 	German localization for Notepad++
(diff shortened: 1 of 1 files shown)


schnurlos:
@MarkusBodensee: Thanks for your updates which I can completely confirm. Sorry for missing your comments at #10428.

PR #10748, opened 2021-11-04
Fix VHDL comments highlighting

oswald3141:
Comments styles in stylers.model.xml are inconsistent with Scintilla's states
described in /scintilla/include/SciLexer.h. Moreover, block comments are
highlighted with not the same style as line comments, but with style used for
"COMMENT LINE BANG" in other languages.
I eliminated the "COMMENT LINE" style, which Scintilla does not generate, fixed
the "COMMENT BLOCK" style to be the same as regular "COMMENT", and added a
separate style, "COMMENT LINE BANG", for special comments.

### DIFF
--- a/PowerEditor/src/stylers.model.xml
+++ b/PowerEditor/src/stylers.model.xml
@@ -1325,8 +1325,8 @@
         <LexerType name="vhdl" desc="VHDL" ext="">
             <WordsStyle name="DEFAULT" styleID="0" fgColor="000000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
             <WordsStyle name="COMMENT" styleID="1" fgColor="008000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
-            <WordsStyle name="COMMENT LIne" styleID="2" fgColor="008000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
-            <WordsStyle name="COMMENT BLOCK" styleID="15" fgColor="008080" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="COMMENT LINE BANG" styleID="2" fgColor="008080" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
+            <WordsStyle name="COMMENT BLOCK" styleID="15" fgColor="008000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
             <WordsStyle name="NUMBER" styleID="3" fgColor="FF8000" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
             <WordsStyle name="STRING" styleID="4" fgColor="808080" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />
             <WordsStyle name="STRING EOL" styleID="7" fgColor="808080" bgColor="FFFFFF" fontName="" fontStyle="0" fontSize="" />


donho:
@oswald3141 
Could you provide the code example so I can see the difference after applying your PR?

oswald3141:
@donho 
Of course. I used this piece of code as an example:
```
-- This is a line comment
/*This is a block comment*/
--! This is a line comment bang
```
This is how it gets highlighted now:
before
And this is how it gets highlighted when the stylers file is corrected:
after

el-coder-sb:
Well, the different highlighting of block comment and line comment was intentional (and is common in other languages/editors as well). May be the way was not the cleanest one. However would be great to get a different highlighting again.

oswald3141:
Sorry, I didn't think it was intentional. It wasn't mentioned in the description of 7bdb370 that the purpose of the change was introducing a different highlighting style for comment *blocks*, so I though there just was a typo.  
I also noticed that in Verilog the style of block comments is the same as of line comments. So that's why I made this PR.  
Anyway, now it's enough to change the fgColor in the line below to change the color of the block comments highlighting. I'm not sure, though, that it should be the same as "COMMENT LINE BANG".
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e87342fef680d497db43af62c08d57a16870c8a6/PowerEditor/src/stylers.model.xml#L1329

el-coder-sb:
You´re right, the descirption of was  7bdb370 was too short. The purpose was mainly introducing comment blocks at all as VHDL2008 was not supported up to that date. The different colors were of minor importance and therefore not mentioned.

>  I'm not sure, though, that it should be the same as "COMMENT LINE BANG".

I agree, it would be better to not have the same color. Looking at stylers.model.xml the third color value used for comments is: fgColor="8050A0". 

@oswald3141 feel free to submit this change if you are fast working with github. If not I´ll see to do so at the end of the week.

oswald3141:
I looked through stylers.model.xml, and it seems like for most languages, Notepad++ applies the same highlighting colour for both line and block comments. I also looked through my own VHDL code with your suggestion implemented. I sometimes use both line and block comments in the same file, and in such places, I got a bit distracted by the different highlighting colours. Of course, it may be just a matter of a habit. But still, could you please explain why you think block and line comments should have different highlighting colours? After all, they both represent the same language construct.

el-coder-sb:
I do some Python and there exist the line comment sign <#> and the doc string sign <'''>. With the latter one being mulitline. And npp does color them different (COMMENT vs. TRIPLE). I adopted that for my vhdl coding. 

However you are right, that in vhdl:
> After all, they both represent the same language construct.

Afaik that is valid for the <--!> as well, as it is not a vhdl construct but just a convention used by doc generators.

So yes, it is just a matter of habit. For me coloring equally is losing a possibility to distinguish different types/levels of comments (commented code vs. doc strings vs. .TODO vs. ...) and making the multiline comment - in my eyes - needless.
However I understand if others do so by different means. That is: If you like to keep it, it won´t bother me.

oswald3141:
Thank you for spending your time writing this explanation. I see your point now. However, I still personally like the idea of keeping highlighting colours at the minimum more. Still I wouldn't mind either if you nonetheless decide to change the colour in a new PR.  

By the way, thank you very much for adding VHDL-2008 support to npp! I write the code using features added by you almost every day.

el-coder-sb:
Great to hear that you benefit from VDHL-2008 support. 

At the moment I´d keep the coloring as is and look how I deal with it.

Coming to an end just a hint to another topic:
If you use VHDL in npp, the function list view could be helpful (#8480). May be you have some ideas to make this better   ...

PR #10755, opened 2021-11-07
Automatically get file type for renaming

### LINKED ISSUE #10729: [Feature request] Match or set default extension in rename window
### Actual Behavior
In the rename window,  by default the "Save as" menu is always set to "All Types (\*.*)".

rename

### Expected Behavior
In the rename window, by default the "Save as" menu match with current file extension.

### Desired Implementation
I request two items to be added in the settings:

1. Match the current file extension in rename window
2. Set default extension in rename window

I hope it's possible to do this and that it does not require too much effort.
Thanks

termdew:
This merge request introduces the following functionality:

When the user wants to rename a file, Notepad++ automatically recognizes the file extension and sets the filter in the file dialog accordingly.

Fixes #10729.


PowerEditor/src/NppIO.cpp
```diff
@@ -1756,12 +1756,27 @@ bool Notepad_plus::fileRename(BufferID id)
...
+		{
+			if (bufLang == item.LangID)
+			{
+				fDlg.setExtFilter(item.longName, item.lexerName);
+			}
+			else
+			{
+				fDlg.setExtFilter(TEXT("All types"), TEXT(".*"));
+			}
+		}
```
MarkusBodensee, line 1769: Should the above construct look more like below? I haven't tested it, it's just a feeling. Your current algorithm sets a filter with every loop iteration, which doesn't look like desired, because in fact, only the last filter item is set then, isn't it?

```
fDlg.setExtFilter(TEXT("All types"), TEXT(".*"));
for (auto item : ScintillaEditView::langNames)
{
	if (bufLang == item.LangID)
	{
		fDlg.setExtFilter(item.longName, item.lexerName);
		break;
	}
}
```

termdew:
Yeah, your code snippet looks indeed more plausible than mine. I will test it and integrate the changes into my branch.

termdew:
@MarkusBodensee unfortunately I could not make your code snippet work. Seems like the ExtFilter cannot be set twice.
I committed a new version, which works as intended now.

MarkusBodensee:
@termdew
Nevertheless, I think you should add a break after you found a matching extension to avoid further looping... Performance reasons...

If you cannot call setExtFilter() more than one time, I think you should consider another way to set the default *.* filter when not finding a matching extension. Maybe use some boolean flag or something.

termdew:
@MarkusBodensee I pushed a new version of the code, it would be nice if you could review it once again 🙂  Thank you! 🙂

MarkusBodensee:
@termdew It looks nearly ok from code review/logical point of view, considering that your statement is valid that `fDlg.setExtFilter(...)` can only be called once.

But now, after a test with the latest binary from the build artifacts, I saw that the behaviour is not like I would expect it.

What I expect: The correct extension entry is selected from the list. If extension is unknown, the first entry of the list is selected, which is `All types (*.*)`.
What actually happens with this PR: The first entry of the list (which normally is `All types (*.*)`) gets replaced with a found extension entry. So the list is modified instead of just the correct list entry is selected. Furthermore the unknown extension case doesn't work, too. Normal text file .txt is displayed.

PowerEditor/src/NppIO.cpp
```diff
@@ -1755,13 +1755,28 @@ bool Notepad_plus::fileRename(BufferID id)
...
+			if (bufLang == item.LangID)
+			{
+				fDlg.setExtIndex(langTypeIndex + 1);
+				const bool defaultAllTypes = NppParameters::getInstance().getNppGUI()._setSaveDlgExtFiltToAllTypes;
+				const generic_string checkboxLabel = _nativeLangSpeaker.getLocalizedStrFromID("file-save-assign-type",
+					TEXT("&Append extension"));
+				fDlg.enableFileTypeCheckbox(checkboxLabel, !defaultAllTypes);
+				break;
+			}
+		}
```
donho, line 1772: The loop is not necessary: `setFileOpenSaveDlgFilters` return the index of `bufLang`.

donho (changes requested):
Remove the loop (`setFileOpenSaveDlgFilters` returns the index of `bufLang`).
Please inspire from the implementation of `fileSaveAs`.

donho:
Due to the absence of reply on the change request, this PR is closed.

PR #10758, opened 2021-11-07
Failed to compile with VS 2022

SinghRajenM:
Compiled npp after a long time and compilation failed in VS 2022, it may fail in VS2019 as well. See if below makes sense. Thanks!

> typedef struct

It results into error (`error C7626: unnamed class used in typedef name cannot declare members other than non-static data members, member enumerations, or member classes`). 

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5208?view=msvc-160
As per above link, this warning (becomes error due to treat warning as error) is new to VS 2019 16.6 onwards.

> constexpr int nameColumn = 0;

`warning C4189: 'nameColumn': local variable is initialized but not referenced`
This warning got converted into error.

### DIFF
--- a/PowerEditor/src/WinControls/DockingWnd/Docking.h
+++ b/PowerEditor/src/WinControls/DockingWnd/Docking.h
@@ -46,7 +46,7 @@
 #define DWS_DF_FLOATING		0x80000000			// default state is floating
 
 
-typedef struct {
+struct tTbData {
 	HWND hClient = nullptr;                // client Window Handle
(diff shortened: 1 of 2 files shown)


donho:
Compiling with VS 2019 is fine.
But the ANSI C way for declaring struct is obsolete indeed.

donho:
BTW welcome back @SinghRajenM :)

SinghRajenM:
> 
> 
> BTW welcome back @SinghRajenM :)

Thank you. :)

PR #10759, opened 2021-11-07
Use stdcpp17

SinghRajenM:
SciLexer already uses `stdcpp17`.


SinghRajenM:
Closing as in release mode, it produces error due to deprecated stuffs warning in cpp17 such as `C4996: 'std::codecvt_utf8`

PR #10809, opened 2021-11-21
Handle environment variables update at runtime

### LINKED ISSUE #10749: Cmd shell created by Notepad++ does not retrieve updated environment variables unless Notepad++ is restarted
### Description of the Issue
When using built-in right click context menu option "Open Containing Folder in cmd", the environment variables available to the shell do not update (whether added, modified, or removed). The environment variables are updated only if Notepad++ itself is restarted.

1.  Open Notepad++
2.  Modify Environment Variable (for example, just add a random new one)
3.  Notepad++'s shell does not get access to any environment variable changes, regardless of how often the shell is opened or closed (normally that's how you update them)
4. Close and re-open Notepad++ for the shell's environment variables to update

### Expected Behavior
The shells should have their environment variables updated when it is closed and re-opened.

### Actual Behavior
Environment variables do not update.

### Debug Information
Notepad++ v8.1.5   (64-bit)
Build time : Sep 26 2021 - 15:23:23
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1288
Current ANSI codepage : 1252

xomx:
Fixes #10749 .

Uses undocumented shell32!RegenerateUserEnvironment.

Keeps current environment variables of the running N++ process, updates the global ones (system & user profile variables).

In case of a conflict (RegenerateUserEnvironment updates a variable, which has been already updated by running N++ process somehow in the local N++ environment set), the global RegenerateUserEnvironment value wins.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -8153,7 +8153,7 @@ void Notepad_plus::handleEnvironmentSettingChange(LPARAM lParam)
 								strValue += *i;
 							}
 
-							DWORD dwRet = ::GetEnvironmentVariable(strName.c_str(), szBuf, BUFSIZE);
+							DWORD dwRet = ::GetEnvironmentVariable(strName.c_str(), szBuf, BUFMAXLEN + 1);
 							if (0 == dwRet)
 							{
 								if (ERROR_ENVVAR_NOT_FOUND == ::GetLastError())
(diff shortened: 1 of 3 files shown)


donho:
@xomx 
I thought (without searching on Internet) there would be a much simple way to get the current global environment settings then erase the old ones with the current one. But obviously it's not that simple. Checking setting by setting to add or remove it, it's risky and complex solution to me - especially it's for a feature not frequently (rarely) requested. So I prefer to not integrate this PR. Hope you understand my position.

xomx:
@donho 
I understand, do not worry. Anybody who maintains a SW project has to always weight pros and cons. Especially when the SW is widespread or critical in some way. Just look, what problems have brought a simple POSIX IO change for WIN32. On a positive note - I have learned something new and now I am using this technique in my private SW with success.

PR #10834, opened 2021-11-25
Update Slovak translation to 8.1.9.2

kubalav:


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/1574deb57010da3fab993ca6aad1733ec1bef932

Issue #10757, opened 2021-11-07
Daylight saving time causing data loss if revisions overwrite and Notepad++ 'Reload' false positive?


AscendChina:
So I have a sync'd MEGA.NZ (like dropbox) folder to a local folder/directory on my harddrive. I also have set up Mega to use revisions. **I'm working on a text based document late into November 6th 2021, and I forgot that its daylight savings where the time goes BACKWARDS 1 hour at exactly 2AM back to 1AM....** so all throughout the hours of the night I'm typing and saving the text document in notepad++ and Mega is doing a good job of keeping sync'd revisions etc.... **then suddenly at around 1:59AM local time, the clock on my system goes back to 1:00AM local time.... I don't even notice it, except for the fact that I get a prompt in notepad++ saying that the text document has changed and that if I wanted to 'reload' the new version from disk.... except I only had one copy of the text file open and hadn't made any changes in the last couple minutes**, certainly not at the time I got the prompt.... so the only thing I can think of is the notepad++ got confused when it all of a sudden saw a different timestamp for last modified.... 

**Could it be possible that due to the time jump issue with daylight savings time** that Mega accidentally or inadvertently went back to an earlier revision and erased the entire hour of work I had done from 1:00am to 2:00am (when 2:00am went back to 1:00am) etc because of the time reversion due to daylight savings?

And/or is this a **bug/flaw in the way notepad++ auto detects changes? Did the going back one hour at the daylight savings time trip it up?!**

monkyyx:
The core issue is related to #10688 & #10753 and fixed by #10774. I've tested it by manually changing the timezone back an hour.

donho:
@AscendChina 
Please check the old issue you created instead of creating the new issue.

Issue #10753, opened 2021-11-07
Notepad++ has a Reload false alarm / bug? "This file has been modified by another program. Do you want to reload it"? even when it hadn't been recently modified at all?!


AscendChina:
Had a weird incident recently where I had multiple text (.txt) based documents that I had written myself (basically notes, reminders, scratchpad stuff) and I'm using notepad++ and with multiple tabs open. I don't have the MS notepad open nor do I have any other editor application (such as Word etc) open/running. I only have one instance of (multi-tabbed) notepad++ open. 

I was doing something in the web browser and switched to the notepad++ application, suddenly I get the prompt about "Reload -- This file has been modified by another program. Do you want to reload it?" and out of habit/instinct I just clicked on "Yes" to get the nag dialog box out of the way, hoping then I could tab over to the text file that I had been using as a pastebin for my copy/paste clipboard of things I was copying from the firefox/chrome browser....

Immediately I'm presented with another such prompt, asking me the exact same question about if I wished to Reload this new tab/ additional text file, that is when I got a bit suspicious since I new for sure I hadn't been updating two different text files... so I took a screenshot with greenshot of it.... then to my surprise after this time clicking that No I didn't wish to reload, I'm prompted a THIRD time about if I want to reload yet ANOTHER tab/instance/textfile! I also click No....

Afterwards since each time it shows the exact path of the text file in question that is asking about reload, I go to the path on the Windows Explore (I'm using Windows 10) and right click each of the text files to check the Created, Modified, Accessed etc and they were from back in late October!

These text files are in a directory that is sync'd by MEGA (sort of like dropbox) with revisions, so I log into my MEGA account and check the revision histories to see what is going on.... Then I download a couple of prior revisions and confirmed that indeed the data looks good, no sign of data corruption or somehow a revision accidentally reverting by mistake, and indeed the MEGA revision confirmed what Windows Explorer confirmed, and that is there hadn't been any changes since October to the files....

Yet, I reboot my machine each night, and my computer hadn't been up for more than 12 hours from the last reboot, I don't do sleep or hibernation, so there is no way I could have had this notepad++ instance open for more than a few hours.... The tabs that I had open weren't edited/modified since October, and yet tonight for some strange reason I got prompted by Notepad++ stating that these files have been changed and asking me if I wish to reload?

My question **is there a bug with Notepad++ that could be doing this false alarm/alert** or is it that my computer is compromised somehow and files are changing behind the scenes without me being aware?

[...]

Issue #10751, opened 2021-11-06
Files cannot be saved sometimes for no reason


SamweisWeb:
### Description of the Issue

As of the latest release 8.1.9 of NP++ 64bit, more than occasionally I'm being prompted with a strange message while trying to save a file. The message informs me about not being able to save the file and asks me whether to restart NP++ in the administrator mode.

### Steps to Reproduce the Issue

None, for it is occasionally.

### Expected Behavior

File saving without this strange message.

### Actual Behavior

See above.

### Debug Information

None.

None.

donho:
What"s the file's path? Files are protected in some location (folder), user needs Admin privilege to modify/delete these files.

blackraz:
> ### Description of the Issue
> As of the latest release 8.1.9 of NP++ 64bit, more than occasionally I'm being prompted with a strange message while trying to save a file. The message informs me about not being able to save the file and asks me whether to restart NP++ in the administrator mode.
> 
> ### Steps to Reproduce the Issue
> None, for it is occasionally.
> 
> ### Expected Behavior
> File saving without this strange message.
> 
> ### Actual Behavior
> See above.
> 
> ### Debug Information
> None.
> 
> None.

I also have this problem :(

Docteh:
@blackraz got any debug information? I've been running 8.1.9 x64 on Windows 10, haven't seen this pop up yet.

Notepad++ v8.1.9   (64-bit)
Build time : Oct 21 2021 - 23:37:53
Path : C:\Users\Kyle\Downloads\npp.8.1.9.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Home (64-bit) 
OS Version : 2009
OS Build : 19043.1288
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

blackraz:
I've been running np++ 8.1.9  x64 on Windows 11 and I have prompted, that file cannot be saved.
 I have downgraded np++ 8.1.8, 8.1.7, 8.1.6 and no luck, after downgrading np++ to 8.1.5 the problem has been solved

SamweisWeb:
@donho
The file is in no special folder. I use it for years, no Admin privileges..

donho:
@blackraz @SamweisWeb 
Could you provide your debug info please?
[...]

Issue #10688, opened 2021-10-18
Reload file message if saved on network drive since v8.1.6


monkyyx:
### Description of the Issue

Since v8.1.6 the reload file message appears after editing & saving a file on a network drive ~~and switching between open files~~.
### Steps to Reproduce the Issue

1. Open a file saved on a network drive
2. Edit and save the file
3. ~~Switch between open files or minimize/maximize Notepad++~~
(sometimes it takes a minute or two before the message appears)

### Expected Behavior

no reload file message
### Actual Behavior

reload file message appears
### Debug Information

Notepad++ v8.1.7   (64-bit)
Build time : Oct 15 2021 - 01:20:46
Path : C:\Users\xxx\Desktop\npp.8.1.7.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19042.1288
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

![npp](https://user-images.githubusercontent.com/88665772/137716394-8ca6e2b4-a882-449f-8ae4-4c71ab2e92fe.PNG)

michelgvandijk:
I had the same problem in v8.1.9 (64-bit), went back to v8.1.5 (64-bit), haven't had the problem since.

rdipardo:
Looks like a regression caused by 11b2dd0:

> The goal of Release 8.1.6 is fix 2 [long time critical issues] [. . .]. In order to remedy these 2 problems, the __file saving API is changed from POSIX IO functions to Win32 native IO API__

https://notepad-plus-plus.org/news/v816-released (edited)

[long time critical issues]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133#issuecomment-929074742

donho:
I have no network drive/samba so I can't reproduce this issue.
However, I find some incoherent code of timestamp detection and did a blend fix:
http://download.notepad-plus-plus.org/repository/MISC/

@monkyyx @michelgvandijk @Docteh @SirHesti @rdipardo 
Could you guys test the fixed binary (download above) and confirm me if it work (and the message you see) please?

rdipardo:
1. Reprocduced in 8.1.9 (32-bit, portable):

[...]

PR #10847, opened 2021-11-28
Add log ability for debugging network drive file status detection issue

donho:
To activate log, user should:
1. Add an empty "nppLogNetworkDriveIssue.xml" file beside of notepad++.exe, or if user has no admin privilege, he/she can add this file into %APPDATA%\Notepad++\.
2. Create "C:\temp\" directory, if it doesn't exist yet.
3. Start notepad++.exe, and wait for the file status (timestamp) detection error from the network drive. If the errors occur, there should be some trace in "C:\temp\nppLogNetworkDriveIssue.log".

People who have had the network drive file status detection issue in #10688, #10753, #10757, #10751 & #10787 are welcome to download the binary and provide the generated log in order to fix this issue.

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -115,7 +115,7 @@ generic_string relativeFilePathToFullFilePath(const TCHAR *relativeFilePath)
 
 void writeFileContent(const TCHAR *file2write, const char *content2write)
 {
-	Win32_IO_File file(file2write, Win32_IO_File::Mode::WRITE);
+	Win32_IO_File file(file2write);
 
 	if (file.isOpened())
 		file.writeStr(content2write);
@@ -124,10 +124,29 @@ void writeFileContent(const TCHAR *file2write, const char *content2write)
 
 void writeLog(const TCHAR *logFileName, const char *log2write)
 {
-	Win32_IO_File file(logFileName, Win32_IO_File::Mode::APPEND);
+	const DWORD accessParam{ GENERIC_READ | GENERIC_WRITE };
+	const DWORD shareParam{ FILE_SHARE_READ | FILE_SHARE_WRITE };
+	const DWORD dispParam{ OPEN_ALWAYS };
+	const DWORD attribParam{ FILE_ATTRIBUTE_NORMAL | FILE_FLAG_POSIX_SEMANTICS | FILE_FLAG_WRITE_THROUGH };
+	HANDLE hFile = ::CreateFileW(logFileName, accessParam, shareParam, NULL, dispParam, attribParam, NULL);
 
-	if (file.isOpened())
-		file.writeStr(log2write);
+	if (hFile != INVALID_HANDLE_VALUE)
+	{
+		LARGE_INTEGER offset;
+		offset.QuadPart = 0;
+		::SetFilePointerEx(hFile, offset, NULL, FILE_END);
+
+		SYSTEMTIME currentTime = { 0 };
+		::GetLocalTime(&currentTime);
+		generic_string dateTimeStrW = getDateTimeStrFrom(TEXT("yyyy-MM-dd HH:mm:ss"), currentTime);
+		std::string log2writeStr(dateTimeStrW.begin(), dateTimeStrW.end());
+		log2writeStr += "  ";
+		log2writeStr += log2write;
+		log2writeStr += "\n";
+
+		DWORD bytes_written = 0;
+		::WriteFile(hFile, log2writeStr.c_str(), log2writeStr.length(), &bytes_written, NULL);
+	}
 }
 
 
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -126,8 +126,8 @@ void writeLog(const TCHAR *logFileName, const char *log2write)
 {
 	const DWORD accessParam{ GENERIC_READ | GENERIC_WRITE };
 	const DWORD shareParam{ FILE_SHARE_READ | FILE_SHARE_WRITE };
-	const DWORD dispParam{ OPEN_ALWAYS };
-	const DWORD attribParam{ FILE_ATTRIBUTE_NORMAL | FILE_FLAG_POSIX_SEMANTICS | FILE_FLAG_WRITE_THROUGH };
+	const DWORD dispParam{ OPEN_ALWAYS }; // Open existing file for writing without destroying it or create new
+	const DWORD attribParam{ FILE_ATTRIBUTE_NORMAL | FILE_FLAG_WRITE_THROUGH };
 	HANDLE hFile = ::CreateFileW(logFileName, accessParam, shareParam, NULL, dispParam, attribParam, NULL);
 
 	if (hFile != INVALID_HANDLE_VALUE)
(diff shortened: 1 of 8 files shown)


donho:
@pnedev @xomx 
If there's more log worth to add, please let me know.

xomx:
@donho 
Nice, the logging works for me and I think it is enough (especially the timestamps are substantial, see below). Unfortunately I have not been able to reproduce the problem yet (I am trying DropBox and Synology NAS).

I try to summarize the problem:
- affected users save a file to network storage place, then they deactivate/activate relevant N++ tab (e.g. minimize and then maximize N++ or simply switch back to N++ window), so the file-change detection mechanism of the N++ is called and they are getting a false positive(?) messagebox from the N++ about a change in file
- because they had not this problem before with the POSIX IO fwrite, I suspect that the WIN32 WriteFile returns after called, but the file on the network storage is not updated yet (this is not an error in the WriteFile, but probably a network layer implementation problem), I do not think that there is a problem with the saved data correctness (see below)
- one of the affected users says that he does not have this issue every time, so I think that the data saving is correct (this is not surprising, as the fwrite CRT implementation at the end calls down the same WriteFile), but sometimes somewhat slower than in the case of fwrite
- another user says "When I save a file to a network drive (which is on a Linux Ubuntu machine) and then try to run it on the Linux machine it says 'text file busy'...", so this is another sign of that the network saving is in fact not finished yet
- I looked at the MS CRT source code (fwrite.c) and from what I saw, they e.g. split the data saved to smaller chunks, so maybe that stuff is what does the magic for a network layer later...
- as I said, I cannot reproduce issues of the affected users, so I have no proof, but if I am right, we can overcome this situation by:

A) delaying the GetFileAttributesEx in Buffer::checkFileState somehow (even the Sleep(1) to switch the N++ thread context once could help)

B) remembering also the last previous timestamp for a file and when we get CompareFileTime == 1, compare also the currently obtained value against it (if it matches, try Sleep(1) or similar and then fire another GetFileAttributesEx check)

C) distinguishing in between local and network file and create some additional mechanism for the network ones

pnedev:
Very good analysis @xomx :+1: 

This is also kind of what I suspected.
It is clear that the problem is connected with the network shared files meaning that in some cases the write has not yet finished (by 'write' I mean our full file access - write - setEndPos - flush).
I haven't had the chance yet to make a setup on my side and try reproducing the issue but as a quick fix I think we can open the file for writing with flag `CREATE_ALWAYS` instead of flag `OPEN_ALWAYS`.

monkyyx:
After the reload file message these entries showed up:

2021-11-29 10:13:12  J:\xxx.txt  in checkFileState(): attributes.ftLastWriteTime (440555049/30926081) < _timeStamp (440558191/30926081)
2021-11-29 10:29:37  J:\xxx.txt  in checkFileState(): attributes.ftLastWriteTime (1058810873/30926083) < _timeStamp (1058815339/30926083)

xomx:
@pnedev 
Yes, for writing we should generally use the CREATE_ALWAYS, good catch.

But I am afraid that this will not solve the network saving problem. One user already posted his log, there is "SetEndOfFile call failed: The process cannot access the file because it is being used by another process.".That means the ERROR_SHARING_VIOLATION code. We are in a situation that N++ successfully opened a file handle for writing and holds it, then it successfully use WriteFile, but then it cannot use SetEndOfFile while it still holds the opened file handle?! I suspect that this user's network storage place supports some kind of transaction operations (another write operation is delayed until the previous is not completed). And SetEndOfFile is not asynchronous, so it returns with this error code instead.

I have one more idea, but I do not know, if it is applicable to the current N++ code. We can switch the order of the calls (1st use SetFilePointerEx at the future data end, then use SetEndOfFile, set the fileptr back to start and only then write the buffer data to the file). But this implicates we know the exact data size beforehand.

pnedev:
@xomx ,

We need `SetEndOfFile` only if the open mode is `OPEN_ALWAYS`. If we change it to `CREATE_ALWAYS` then the file is completely wiped before opening so we won't need to use `SetEndOfFile` (as file size will be set to zero and `WriteFile` will automatically update it) and we will be kind of emulating the `fopen` and `fwrite` behavior.
Right now I suspect `WriteFile` is not completely finished on the network side and thus `SetEndOfFile` fails.

But let Don first try swapping the call order of `SetEndOfFile` and `FlushFileBuffer` to see if this will help `WriteFile` finish before we call `SetEndOfFile`.
Let changing `OPEN_ALWAYS` to `CREATE_ALWAYS` be our last measure because that will bring back this problem: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664.

Thanks!

donho:
@monkyyx 
Could you try the fix: [32 bits](https://ci.appveyor.com/api/buildjobs/36xeeoeu68fh6jen/artifacts/Notepad%2B%2B.Win32.Release.exe) or [64 bits](https://ci.appveyor.com/api/buildjobs/hui4t4w7t6ni265m/artifacts/Notepad%2B%2B.x64.Release.exe)
Please remove the old log file before testing.

monkyyx:
@donho Here you go. Looks the same to me.

2021-11-29 19:35:52  J:\xxx.txt  in checkFileState(): attributes.ftLastWriteTime (3747267540/30926159) < _timeStamp (3747271704/30926159)
2021-11-29 19:37:21  J:\xxx.txt  in checkFileState(): attributes.ftLastWriteTime (502281556/30926160) < _timeStamp (502286928/30926160)
2021-11-29 19:37:25  J:\yyy.txt  in checkFileState(): attributes.ftLastWriteTime (519011895/30926160) < _timeStamp (519026223/30926160)

donho:
@monkyyx 
Thank you. I suppose that `J` drive is the mounted network drive, right?

@xomx 
> C) distinguishing in between a local and a network file and adding some additional mechanism to handle the network ones

On the example provided by @monkyyx , how can a network drive be distinguishable?

xomx:
@donho 
I translated the @monkyyx log-data to more readable form (I converted his FILETIME values to UTC SYSTEMTIME and then translated it to localtime). The timestamp differences are mostly below 1ms. E.g. here are his three last log-lines in a more readable form (rrrr/mm/dd hh:mm:ss.ms):

attributes.ftLastWriteTime  2021/11/29  19:35:24.316  <  _timeStamp  2021/11/29  19:35:24.316
attributes.ftLastWriteTime  2021/11/29  19:37:09.314  <  _timeStamp  2021/11/29  19:37:09.315
attributes.ftLastWriteTime  2021/11/29  19:37:10.987  <  _timeStamp  2021/11/29  19:37:10.988

So can you modify the Win32_IO_File::close in this PR like:

FlushFileBuffers(hFile);
Sleep(1); // thread context switching only (let's see if that helps to solve the x64 problem for @GTR8000)
SetEndOfFile(_hFile);
Sleep(1); // thread context switching only (this should help @monkyyx)

About the network drive detection:
Yes, it is a good question. I do not know a reliable one. Network places mapped as a drive letter or a folder in Windows should have somewhere the real network path visible. But if e.g. a client SW like DropBox watches a folder or file marked as shared for this service and makes all filedata transfers transparent to the user, we are at the dead end here.

monkyyx:
@donho Correct. It's an SMB share mounted as a network drive.

donho:
@xomx 
I really don't like Sleep solution not only because it smell the hack but also because its uncertain status - you don't really know how much time you will need.
I prefer the solution that distinguish the network drive files from local drive ones then disable the timestamp update from the past.

PR #10874, opened 2021-12-04
Fix dirty status of reloaded files (isUnsync true)

### LINKED ISSUE #10796: reload file but the label is still red
### Description of the Issue

v8.1.9 (64-bit)

1. open git.c
2. Modify git.c with another software, and save file
3. back to notepad++, choose No
image
4. Reload from the disk (Ctrl+R) and choose Yes
5. table is still red
image

xomx:
Fixes #10796 .

After a successful reload, the N++ filebuffer should be marked as not dirty, because of now it is the same as its file source on disk. But that was not true, when the filebufer had been marked as unsynced before the reloading. Steps to reproduce the previous incorrect behavior:

1. Open a file in N++.
2. Switch to another editor, edit the same file there and save it.
3. Switch back to N++, there will be the changed-file messagebox with options to reload / not reload, choose 'No'.
4. The N++ filebufer is now marked as dirty (this is ok).
5. Then initiate the same reloading manually (eg via Ctrl+R), this time choose 'Yes'.
6. After a successful reload, the filebuffer in N++ is now incorrectly marked as dirty.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -721,7 +721,18 @@ bool FileManager::reloadBuffer(BufferID id)
 	buf->setLoadedDirty(false);	// Since the buffer will be reloaded from the disk, and it will be clean (not dirty), we can set _isLoadedDirty false safetly.
 								// Set _isLoadedDirty false before calling "_pscratchTilla->execute(SCI_CLEARALL);" in loadFileData() to avoid setDirty in SCN_SAVEPOINTREACHED / SCN_SAVEPOINTLEFT
 
+	bool bUnsyncStateBeforeReload = buf->isUnsync(); // store for a possible unsync state restoration later 
+	buf->setUnsync(false);	// we have to ensure that isUnsync is false before we try to call the loadFileData below for the reloading
+							// - after a successful reload, the source file and its filebuffer in N++ are the same, so the isDirty should be false (and the isUnsync too)
+							// - but when the isUnsync is true at this point, any successfully reloaded filebuffer later will end up marked as dirty!
+							//   (similar reason as in the above setLoadedDirty(false), but now it is the "_pscratchTilla->execute(SCI_SETSAVEPOINT)" in loadFileData(),
+							//    which ends up in the Notepad_plus::notify(), case SCN_SAVEPOINTLEFT, where is logic "if (buf->isUnsync()) isDirty = true")
+
 	bool res = loadFileData(doc, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
+
+	if (!res)
+		buf->setUnsync(bUnsyncStateBeforeReload);
+
 	buf->_canNotify = true;
 
 	if (res)
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -722,11 +722,7 @@ bool FileManager::reloadBuffer(BufferID id)
 								// Set _isLoadedDirty false before calling "_pscratchTilla->execute(SCI_CLEARALL);" in loadFileData() to avoid setDirty in SCN_SAVEPOINTREACHED / SCN_SAVEPOINTLEFT
 
 	bool bUnsyncStateBeforeReload = buf->isUnsync(); // store for a possible unsync state restoration later 
-	buf->setUnsync(false);	// we have to ensure that isUnsync is false before we try to call the loadFileData below for the reloading
-							// - after a successful reload, the source file and its filebuffer in N++ are the same, so the isDirty should be false (and the isUnsync too)
-							// - but when the isUnsync is true at this point, any successfully reloaded filebuffer later will end up marked as dirty!
-							//   (similar reason as in the above setLoadedDirty(false), but now it is the "_pscratchTilla->execute(SCI_SETSAVEPOINT)" in loadFileData(),
-							//    which ends up in the Notepad_plus::notify(), case SCN_SAVEPOINTLEFT, where is logic "if (buf->isUnsync()) isDirty = true")
+	buf->setUnsync(false);	// ensure that isUnsync is false before calling the loadFileData, otherwise isDirty and isUnsync will not be false after a successful reload (issue fixed: #10796)
 
 	bool res = loadFileData(doc, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
 
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -721,13 +721,13 @@ bool FileManager::reloadBuffer(BufferID id)
 	buf->setLoadedDirty(false);	// Since the buffer will be reloaded from the disk, and it will be clean (not dirty), we can set _isLoadedDirty false safetly.
 								// Set _isLoadedDirty false before calling "_pscratchTilla->execute(SCI_CLEARALL);" in loadFileData() to avoid setDirty in SCN_SAVEPOINTREACHED / SCN_SAVEPOINTLEFT
 
-	bool bUnsyncStateBeforeReload = buf->isUnsync(); // store for a possible unsync state restoration later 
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -721,7 +721,18 @@ bool FileManager::reloadBuffer(BufferID id)
 	buf->setLoadedDirty(false);	// Since the buffer will be reloaded from the disk, and it will be clean (not dirty), we can set _isLoadedDirty false safetly.
 								// Set _isLoadedDirty false before calling "_pscratchTilla->execute(SCI_CLEARALL);" in loadFileData() to avoid setDirty in SCN_SAVEPOINTREACHED / SCN_SAVEPOINTLEFT
 
+	bool bUnsyncStateBeforeReload = buf->isUnsync(); // store for a possible unsync state restoration later 
+	buf->setUnsync(false);	// we have to ensure that isUnsync is false before we try to call the loadFileData below for the reloading
+							// - after a successful reload, the source file and its filebuffer in N++ are the same, so the isDirty should be false (and the isUnsync too)
+							// - but when the isUnsync is true at this point, any successfully reloaded filebuffer later will end up marked as dirty!
+							//   (similar reason as in the above setLoadedDirty(false), but now it is the "_pscratchTilla->execute(SCI_SETSAVEPOINT)" in loadFileData(),
+							//    which ends up in the Notepad_plus::notify(), case SCN_SAVEPOINTLEFT, where is logic "if (buf->isUnsync()) isDirty = true")
```
donho, line 729: It's nice to have comment, but it's kind of too verbose.
I would give a simple explanation then provide the URL of issue.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -721,7 +721,18 @@ bool FileManager::reloadBuffer(BufferID id)
...
+	buf->setUnsync(false);	// we have to ensure that isUnsync is false before we try to call the loadFileData below for the reloading
+							// - after a successful reload, the source file and its filebuffer in N++ are the same, so the isDirty should be false (and the isUnsync too)
+							// - but when the isUnsync is true at this point, any successfully reloaded filebuffer later will end up marked as dirty!
+							//   (similar reason as in the above setLoadedDirty(false), but now it is the "_pscratchTilla->execute(SCI_SETSAVEPOINT)" in loadFileData(),
+							//    which ends up in the Notepad_plus::notify(), case SCN_SAVEPOINTLEFT, where is logic "if (buf->isUnsync()) isDirty = true")
+
 	bool res = loadFileData(doc, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
+
+	if (!res)
+		buf->setUnsync(bUnsyncStateBeforeReload);
```
donho, line 734: So it's enough to call setUnsync(true) once only when the `res` is true, isn't it?

donho (changes requested):
1. Make the fix simpler.
2. Make the comment (if any) simpler.

PowerEditor/src/ScintillaComponent/Buffer.cpp
xomx: Ok, I will shorten that.

PowerEditor/src/ScintillaComponent/Buffer.cpp
xomx: No.
When the `res` is true, reloading succeeded and the filebuffer `_isUnsync` and `_isDirty` should be both false.
So only when the `res` is false, I am taking care and restoring it to the previous state whatever it is.

donho: Sorry correction:
So it's enough to call setUnsync(f**alse**) once only when the res is true, isn't it?

donho: If res is true, then set unsync false unconditionally, otherwise, just do nothing.
It's not necessary to call it twice.

xomx: There were in fact two problems, when the `_isUnsync` was true at this point:
- this flag caused incorrect activation of the _isDirty flag later in the loadFileData()
- nobody cared to reset it back to false after a successful reload

I solved both problems by calling `buf->setUnsync(false)` unconditionally before the loadFileData().
Only if the reloading in loadFileData() fails, one should care about resetting it back to a possible previous unsynced state.
So maybe I should rather use:

```
if (!res && bUnsyncStateBeforeReload)
	buf->setUnsync(true);
```
Is that form better?

donho: bUnsyncStateBeforeReload is useless IMO.
The fix can be much simpler:
```cpp
if (res)
    buf->setUnsync(false);
```
We can ignore current unsync state until reload result is true, then unsync state must be set to false. Or am I missing something?

donho: BTW, you can set inside of the following block:
```cpp
if (res)
	{
		setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
	}
	```

Comment (xomx) [reply PowerEditor/src/ScintillaComponent/Buffer.cpp:730]: > We can ignore current unsync state until reload result is true, then unsync state must be set to false. Or am I missing something?

Yes, it will not fix the issue, _isDirty will be triggered in the loadFileData().

It can be:

```
bool res = loadFileData(doc, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
if (res)
{
	buf->setUnsync(false);
	buf->setDirty(false);
}
```
But I will not put it in the 
```
if (res)
{
	setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
}
```
because of there is the `buf->_canNotify = true;` before it and I do not want to verify possible consequences.

So do you agree with this:

```
bool res = loadFileData(doc, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
if (res)
{
	buf->setUnsync(false);
	buf->setDirty(false);
}

buf->_canNotify = true;

if (res)
{
	setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
}

return res;
```

Comment (donho) [reply PowerEditor/src/ScintillaComponent/Buffer.cpp:730]: @xomx 
> it will not fix the issue, _isDirty will be triggered in the loadFileData().

You are right about it 

Let's use the following version - it'll be more clear IMO. 
```cpp
bool res = loadFileData(doc, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
if (res)
{
	buf->setUnsync(false);
	buf->setDirty(false);
}

buf->_canNotify = true;

if (res)
{
	setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
}
```

xomx: Ok.

donho: > because of there is the buf->_canNotify = true; before it and I do not want to verify possible consequences.

Just check it, you can safe to put 2 line into the block :
```
if (res)
{
	buf->setUnsync(false);
	buf->setDirty(false);
	setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
}
buf->_canNotify = true;
```
`buf->_canNotify = false;` is set in the beginning of function `reloadBuffer` for locking the loading period (especial the call of `loadFileData`) without notification. 

But if you don't want to do it, it's OK, I'll modify it afterward.

xomx: I am not as familiar with the N++ code as you, so I was just careful not to screw up something. Modify the code as you think it fits. Good night.

donho (approved):

PowerEditor/src/ScintillaComponent/Buffer.cpp
donho: Thank you @xomx for your fix and consistence!

PR #10884, opened 2021-12-06
Add "nameof" reserved word to C#.

### LINKED ISSUE #10870: C# reserved words 'nameof' is not marked as reserved
I'm using Notepad++ v.8.1.9.2 (32-bit) and the reserved words `when` and `nameof` from C# is not marked as reserved (as blue, in my case).
Pull requests #9667 and #8633 implement these reserved words. But it looks like they haven't been included in the most recent releases yet.

digocesar:
The reserved word `nameof` was missing in one file to be marked in blue on editor.
Closes #10870


donho:
Don't add the keyword one by one per PR. Last time while "when" was added, it would be better to add the others.
It's fastidious to not merge the PR which contains all info. Therefore I will add it myself.

digocesar:
> Don't add the keyword one by one per PR. Last time while "when" was added, it would be better to add the others. It's fastidious to not merge the PR which contains all info. Therefore I will add it myself.

Excuse me.
The two keywords were included at different times, months ago. That's why they are in separate PR.
But when I added `nameof` I didn't know that I needed to add in 2 files to work.

However, in my opinion, it's better to have separate PR because it can happen that one word is accepted and the other not. But this is my opinion!

PR #156, opened 2015-06-02
Use GWLP_* instead of GWL_* and DWLP_* instead of DWL_* 

NN---:
Since we use GetWindowLongPtr (GWLP) and SetWindowLongPtr.

These functions work with LONG_PTR and not LONG.

This PR fixed merge problems of #129 and available for merge.

PR #152, opened 2015-06-01
DlgProc returns INT_PTR and not BOOL.

NN---:
Use correct prototype and remove casts.
See https://msdn.microsoft.com/en-us/library/windows/desktop/ms645469(v=vs.85).aspx

PR #10912, opened 2021-12-13
update boost regex version to 1.78.0 

chcg:
from https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.7z
changes from https://www.boost.org/users/history/version_1_78_0.html:

Regex:

    Breaking Change: Change \B to be the opposite of \b as per Perl behaviour.
    Change w32_regex_traits.hpp so that windows.h is no longer included.
    Fixed fuzzing related issues #151, #152, #153, #156.


donho:
@chcg 
So there's no change in boost 1.77 for regexp package?

chcg:
@donho Yes, there was no change for 1.77.0 regarding the regex part.

Issue #10907, opened 2021-12-12
[MinGW] Build not work under path with space


ArkadiuszMichalski:
If we put Notepad++ source code to the path with space then build will fail.  This is an old GNU make problem ([read this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10886#issuecomment-991798146)). For Git Bash tool I make [some tweak ](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10886) and it should work, but with the Notepad++ source code inside such path there is still a problem.

Belowe is a result when I put source code to `Program files` folder: 

```
mingw32-make[1]: *** No rule to make target 'files/NPP/PowerEditor/gcc/bin.i686.
build/libscilexer.a'.  Stop.
mingw32-make: *** [makefile:212: bin.i686.build/libscilexer.a] Error 2
```

Also it's possible that the problem exist in the Scintilla's makefile.

If it is not going to be repaired, it is worth mentioning it in the build.md. It is also worth detecting it in the makefile and interrupting its work with an appropriate message.

PR #10914, opened 2021-12-14
Fix [MinGW] Build not work under path with space

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10907. Before test put Notepad++ source code to path with including space. I checked CMD and Git Bash and it works, but some confirmation from others would be advisable.


ArkadiuszMichalski:
The problem occurs when we run makefile outside it's directory, like appveyor does:
 https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ee884f87bbb66d1431f0d530d0cd25bb7cde6b35/appveyor.yml#L104

Handling this case in line: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ee884f87bbb66d1431f0d530d0cd25bb7cde6b35/PowerEditor/gcc/makefile#L105
it is not very flexible. In the MAKEFILE_LIST variable we have the paths to each processed makefile, but the separator is a space, so if any of them has a space in their path it will be a problem. I have no idea how to get only the path of the current makefile. I'll try to get the value of the `-f` command (if possible).

ArkadiuszMichalski:
Yes, above comment is for this failure.

ArkadiuszMichalski:
@donho 

After this fix spaces in the source path are allowed until we run the `makefile` directly from the `PowerEditor/gcc` folder. Unfortunately outside of this folder will not work (for example when from another folder we specify a `makefile` with the `-f` option). Fixing this would require passing the path to the makefile in an extra variable, and then doing a mass of space conversion to something else (and back it to space again) which will only obfuscate all the code. I just added a short note in `build.md` to avoid spaces in the path for such cases.

donho:
Under both x86 & x64 Native Tool Command Prompt for VS 2022, I've got the same result:

```
C:\tmp\New folder\notepad-plus-plus\PowerEditor\gcc>mingw32-make.exe
* creating BUILD_DIRECTORY bin.i686.build
mingw32-make.exe[1]: *** No rule to make target 'folder/notepad-plus-plus/PowerEditor/gcc/bin.i686.build/libscilexer.a'.  Stop.
mingw32-make.exe: *** [makefile:199: bin.i686.build/libscilexer.a] Error 2
```

I suppose the PR is only for git bash - in this case, I prefer change the Build.md to force users to not put into the folder containing white spaces.

ArkadiuszMichalski:
@donho Can you try directly from CMD running on the system? It looks like you are testing an old makefile or this tool implicitly uses `-f`. On first screenshot I have relative path `../../PowerEditor/gcc/bin.i686.build/libscilexer.a`, and that's why it works because spaces aren't given explicitly. But in your log I can still see absolute path.

image

image

donho:
image

donho:
oups, my bad.
Instead of copying makefile, I copied makefile.rc

PR #10924, opened 2021-12-19
Add exclude folder(s) capacity in Find in Files (obsoled)

### LINKED ISSUE #2433: Notepad++ exclude folder find in files filter

### LINKED ISSUE #8350: How to exclude one or more subfolders from folder opening in command line?
How to exclude one or more subfolders from folder opening in command line?

It is possible to open a folder:
- notepad++.exe "base"

- base
-- subfolder1
--- sub-subfolder1-1
--- sub-subfolder1-2
---- sub-sub-subfolder1-2-1
---- sub-sub-subfolder1-2-2
-- subfolder2
--- sub-subfolder2-1
-- subfolder3
--- sub-subfolder3-1

How to open:
- base without subfolder1
- base without sub-subfolder1-2
- base without sub-sub-subfolder1-2-2

Thanks in advance.

donho:
A "Exclude" combobox is added, that allows users to provide the list of excluding folder names with the same format as "Filters" - allow `*` and use white space as separator. The provided name(s) will be used for excluding only the 1st level of matched directory. If users need to exclude folders with the same name (or names matched the specific pattern) in all levels, a prefix `+` should be added. 

Here is a sample:
Filters: `*.*`
Directory: `c:\myProject\`
Exclude: `bin +log* +test`

So the excluded directories could be:
1. *c:\myProject\bin\\*
2. *c:\myProject\log\\*
3. *c:\myProject\logs\\*
4. *c:\myProject\src\log4j\\*
5. *c:\myProject\test\\*
6. *c:\myProject\src\test\\*

The following directories will not be excluded:
1. *c:\myProject\foo\tests\\*
2. *c:\myProject\foo\bin\\*

Fix #2433, fix  #8350


Yaron10:
@donho,

That's a useful addition. Thank you for your work. 👍 
(I haven't tested it yet).

I'm wondering if the extra `Exclude` ComboBox is necessary. 
Why not use `Filters` for excluding both files and folders? - This would be more "natural" IMO.

1. Place `Directory` *before* `Filters`.
2. Add the syntax to the tooltip.

Directory: `c:\myProject\`
Filters: `!*.exe !\bin !\+log* !\+test`

(`!\` for excluding a folder is an example. the preferred syntax can be discussed).

donho:
@Yaron10 Interesting!
I would say:
Filters: !*.exe !bin\ !+log*\ !+test\

What do you think?

Yaron10:
@donho,

I suppose you have a reason for preferring `!bin\`. I think it would be fine.
`!\` might be slightly more intuitive as it translates to "Not Folder".

Thanks again for you work.

donho:
@Yaron10 
Your suggestion is almost perfect. I say "almost perfect" because users could provide the following patterns with the analogy of this syntax:
`*.text !*.exe !\bin !\+log* !\+test \src \+doc`

There's a logic problem in the above patterns and the result of any possible process could be considered as bug.

Yaron10:
@donho,

Let's try and exclude the "almost". :)

I hope I understand you correctly.
`\src \+doc` should not be allowed.
For files you can use implicit exclusion (`*.cpp`) and explicit exclusion (`!*.exe`). 
For folders you can only use explicit exclusion (`!\bin`) - *just like your implementation with the `Exclude` ComboBox*.

***

Implicit folder exclusion can be useful in some cases.
If you have 20 sub-folders under `c:\myProject\` and you only want to search in the `A` and `B` sub-folders, it should be easier to use `\A \B` in `Filters`.
The question is if you want to implement it at this stage.

donho:
@Yaron10 
> For folders you can only use explicit exclusion (!\bin) - just like your implementation with the Exclude ComboBox.

Let's go in this direction for the moment.

Yaron10:
Great.

Allow me one more question:
How about changing `!\+log*` to `!\log*+`? - All the "arguments" at the end.

I'll test the new version.
Thank you. I appreciate your work.

donho:
@Yaron10 
Done. Please check https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10927

donho:
Use #10927 instead.

PR #10927, opened 2021-12-20
Add exclude folder(s) capacity in Find in Files

### LINKED ISSUE #2433: Notepad++ exclude folder find in files filter

### LINKED ISSUE #8350: How to exclude one or more subfolders from folder opening in command line?

donho:
This PR allows users to provide the list of excluding folder names/patterns in "Filters" field.
Eclusion operator is alway "!" at the begining. In order to distinguish folder from file, "\\" should be used as prefix of the folder name/pattern, following "!". That allows the exclusion of the directories under the root directory you want to search (the 1st level of matched directories).

If users need to exclude folders with the same name (or names matched the specific pattern) in all levels, the `+` should be put between "!" and "\\" to exclude them recursively.

Here is a sample:
Filters: `*.* !\bin !+\test !+\log*`
Directory: `c:\myProject\`

So the excluded directories could be:
1. c:\myProject\bin\
2. c:\myProject\log\
3. c:\myProject\logs\
4. c:\myProject\src\log4j\
5. c:\myProject\test\
6. c:\myProject\src\test\

The following directories will not be excluded:
1. c:\myProject\foo\tests\
2. c:\myProject\foo\bin\

Note that "inclusion of folder" is not allowed, and such pattern will be ignored.

Fix #2433, fix  #8350

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -959,23 +959,18 @@ bool matchInExcludeDirList(const TCHAR* dirName, const std::vector<generic_strin
 {
 	for (size_t i = 0, len = patterns.size(); i < len; ++i)
 	{
-		if (patterns[i].length() > 2 && patterns[i][0] == '!' && patterns[i][1] == '\\') // if it's dir it should have at least "!\" as prefix for excluded dir
+		size_t patterLen = patterns[i].length();
+
+		if (patterLen > 3 && patterns[i][0] == '!' && patterns[i][1] == '+' && patterns[i][2] == '\\') // check for !+\folderPattern: for all levels - search this pattern recursively
 		{
-			size_t patterLen = patterns[i].length();
-			if (patterLen > 3 && patterns[i][patterLen - 1] == '+') // for all levels - search this pattern recursively
-			{
-				generic_string realPattern = patterns[i].substr(2, patterLen - 3);
-				if (PathMatchSpec(dirName, realPattern.c_str()))
-					return true; // excluded folder matched
-			}
-			else // exclusive pattern for only the 1st level
-			{
-				if (level == 1)
-				{
-					if (PathMatchSpec(dirName, patterns[i].c_str() + 2))
-						return true; // excluded folder matched
-				}
-			}
+			if (PathMatchSpec(dirName, patterns[i].c_str() + 3))
+				return true;
+		}
+		else if (patterLen > 2 && patterns[i][0] == '!' && patterns[i][1] == '\\') // check for !\folderPattern: exclusive pattern for only the 1st level
+		{
+			if (level == 1)
+				if (PathMatchSpec(dirName, patterns[i].c_str() + 2))
+					return true;
 		}
 	}
 	return false;
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -955,6 +955,32 @@ bool matchInList(const TCHAR *fileName, const std::vector<generic_string> & patt
 	return is_matched;
 }
 
+bool matchInExcludeDirList(const TCHAR* dirName, const std::vector<generic_string>& patterns, size_t level)
+{
+	for (size_t i = 0, len = patterns.size(); i < len; ++i)
+	{
+		if (patterns[i].length() > 2 && patterns[i][0] == '!' && patterns[i][1] == '\\') // if it's dir it should have at least "!\" as prefix for excluded dir
+		{
+			size_t patterLen = patterns[i].length();
+			if (patterLen > 3 && patterns[i][patterLen - 1] == '+') // for all levels - search this pattern recursively
+			{
+				generic_string realPattern = patterns[i].substr(2, patterLen - 3);
+				if (PathMatchSpec(dirName, realPattern.c_str()))
+					return true; // excluded folder matched
+			}
+			else // exclusive pattern for only the 1st level
+			{
+				if (level == 1)
+				{
+					if (PathMatchSpec(dirName, patterns[i].c_str() + 2))
+						return true; // excluded folder matched
+				}
+			}
+		}
+	}
+	return false;
+}
+
 bool allPatternsAreExclusion(const std::vector<generic_string> patterns)
 {
 	bool oneInclusionPatternFound = false;
--- a/PowerEditor/src/MISC/Common/Common.h
+++ b/PowerEditor/src/MISC/Common/Common.h
@@ -83,6 +83,7 @@ std::string getFileContent(const TCHAR *file2read);
 generic_string relativeFilePathToFullFilePath(const TCHAR *relativeFilePath);
 void writeFileContent(const TCHAR *file2write, const char *content2write);
 bool matchInList(const TCHAR *fileName, const std::vector<generic_string> & patterns);
+bool matchInExcludeDirList(const TCHAR* dirName, const std::vector<generic_string>& patterns, size_t level);
 bool allPatternsAreExclusion(const std::vector<generic_string> patterns);
 
 class WcharMbcsConvertor final
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1465,8 +1465,9 @@ void Notepad_plus::removeDuplicateLines()
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 }
 
-void Notepad_plus::getMatchedFileNames(const TCHAR *dir, const vector<generic_string> & patterns, vector<generic_string> & fileNames, bool isRecursive, bool isInHiddenDir)
+void Notepad_plus::getMatchedFileNames(const TCHAR *dir, size_t level, const vector<generic_string> & patterns, vector<generic_string> & fileNames, bool isRecursive, bool isInHiddenDir)
 {
+	level++;
 	generic_string dirFilter(dir);
 	dirFilter += TEXT("*.*");
 	WIN32_FIND_DATA foundData;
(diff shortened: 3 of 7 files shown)


Yaron10:
@donho,

I'd like to make sure we're on the same page.

> In order to distinguish folder from file, "" should be used as prefix of the folder

I suppose "" is a typo and you meant `\`, correct?

I have 2 sub-folders: `test` and `test2`.
Using `!\test*`, `test2` is NOT excluded. Am I missing anything?

***
Did you intentionally not change the order of `Filters` and `Directory`?
(Now that `Filters` applies to `Directory` as well, shouldn't it be the last combo?).

Are you planning to add the new syntax to the Filter's tooltip? Or the docu should suffice?

alankilborn:
>(Now that Filters applies to Directory as well, shouldn't it be the last combo?).

It would have been nice to ALWAYS have had THAT top-to-bottom ordering. :-)

donho:
> I suppose "" is a typo and you meant \, correct?

Yes. Fixed.

> Using !\test*, test2 is NOT excluded. Am I missing anything?

Strange, it does work for me:


Without exclusion:

image


vs with exclusion:

image

Yaron10:
@alankilborn,

Does `!\test*` work for you?

Yaron10:
@donho,

[This](https://github.com/notepad-plus-plus/notepad-plus-plus/files/7747851/tmp.zip) is my `tmp` folder. Could you please test it?

תמונה

donho:
@Yaron10 
Could you provide me the link of binary you use? x64 or x86 ?
Could you provide me also your debug info?

Yaron10:
@donho,

Interesting.

https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/41957233/job/4gdqu0osturf3h4k/artifacts

Notepad++ v8.1.9.3   (32-bit)
Build time : Dec 20 2021 - 17:53:18
Path : C:\---\notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255

donho:
@Yaron10 The 1st test with my local build:

image

Notepad++ v8.1.9.3   (64-bit)
Build time : Dec 20 2021 - 20:36:18
Path : C:\ddddd\notepad-plus-plus\PowerEditor\bin64\Notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 22000.376
Current ANSI codepage : 1252
Plugins : none

ArkadiuszMichalski:
For me also not work for above test with `!\Test*`:
image
```
Notepad++ v8.1.9.3   (32-bit)
Build time : Dec 20 2021 - 17:53:18
Command Line : -multiInst
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Build : 7601.0
Current ANSI codepage : 1250
```

Yaron10:
@donho,

I've downloaded a fresh portable x64, replaced the binary with https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/41957233/job/an9ikvlpo56mw8wr/artifacts and got the same result.

תמונה

donho:
@ArkadiuszMichalski @Yaron10 

With the binary link that @Yaron10 has provide, It works as expected : 

image


Notepad++ v8.1.9.3   (32-bit)
Build time : Dec 20 2021 - 17:53:18
Path : C:\sources\notepad-plus-plus\PowerEditor\bin\Notepad++.Win32.Release.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 22000.376
Current ANSI codepage : 1252
Plugins : none


I'm using Windows11 - Maybe it's reproducible under Windows 10 and version before ?
I'm gonna test under my windows 10

Yaron10:
@donho,

Thank you.
Could you please test my tmp files?

@ArkadiuszMichalski,

Thank you for testing and sharing.

Yaron10:
@donho,

> With the binary link that @Yaron10 has provided, It works as expected

Do you mean both x32 and x64?

donho:
@Yaron10 
> Could you please test my tmp files?

Sure. just upload it.

donho:
> Do you mean both x32 and x64?

Only 32

Yaron10:
I've uploaded it in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10927#issuecomment-998255797. :)

donho:
@Yaron10 
Good news! I can reproduce the bug with your files under both Win10 & win11 !
I will provide the fix ASAP - thank you & @ArkadiuszMichalski for the bug detection and tests!

Yaron10:
@donho,

Great. 👍 
Thank you very much.

With your permission, I'll continue tomorrow.

donho:
> With your permission, I'll continue tomorrow.

It's not me who gives you the permission, it's rather your wife or you gf :P
No problem for that - thank you for your help. Good night then.

donho:
@Yaron10 
After debugging the code, I saw it's the following code which failed your folder `Test2`:
```cpp
if (PathMatchSpec(dirName, pattern))
	return true; // excluded folder matched
```

With your searching sample, it's PathMatchSpec("Test2", "test*") => the returned value should be `true`, but it failed with your folder "Test2". 
[PathMatchSpec](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathmatchspeca) is one of MS Win32API so it should be bug free.
Therefore I renamed your "Test2" to "test3", then rename it again back to "Test2" via Explorer. With the same data to do find in files, now the result is correct (the folder "Test2" of the first level under the root is excluded):

image

That reminds me an old colleague has shown me the hacker uses Cyrillic character to replace English letter (both letter looks like almost the same) on the URL for phishing. So I copy your original Test2 folder from the zip to my test directory. Now both "Test2" folders are sitting together:

image

All the letters looks identical, my best guess is there is one or more invisible unicode character in "Test2" string, inserted by error.

@ArkadiuszMichalski Could you make your test samples without using the zip provided by @Yaron10 please? I'm sure that it's OK to you.

ArkadiuszMichalski:
Yes, this name "Test2" has something invisible before `T`, just try raname it's name and press right arrow, need 3 steps to jump after `T`. I didn't even know we could do that in the paths.

donho:
image


Under git bash is more clear:

image

Yaron10:
@donho,

Thank you for investigating and finding the problem.

(@ArkadiuszMichalski, thank you too for the additional info).

It's very interesting.
When I create a new folder in File Explorer and keep the default name - copying the name/path via a simple script, I get some gibberish characters at the beginning.
*The problem normally does not occur after renaming the folder*.
Something in the way I renamed/copied the folders in the "tmp" folder I uploaded probably caused those hidden characters to remain there.

*Copying the default "New Folder" name via Windows `Copy As Path`, I get the proper name without the gibberish characters.*

Well, another RTL issue.

***

Testing the new feature (without the hidden characters) - it's working like a charm. Great job. 👍 

One issue:
`+` is a legit folder-name character.
So if you have a `Test+` sub-folder and you use `!\Test+` - you might not get the expected results.

alankilborn:
>+ is a legit folder-name character

Well, if we're going there...  spaces are legit folder name characters, too. :-)

ArkadiuszMichalski:
>\+ is a legit folder-name character
So if you have a Test+ sub-folder and you use !\Test+ - you might not get the expected results.

So it is best to have all pattern characters before the `\` which is forbidden in the name (`+Test` is also a valid name).

donho:
@Yaron10 
Here is the list of forbidden characters under Windows file system: 
https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names
I don't see any suitable one for the semantic "recursive" or "global" (except `*`, but it is part of pattern symbol). What's another better choice in your opinion?

ArkadiuszMichalski:
It cannot simply be `+!\Test` or `!+\Test`? Everything related to the name will be after `\`, and the behavior control will be before `\`.

donho:
@ArkadiuszMichalski 
> So it is best to have all pattern characters before the \ which is forbidden in the name (+Test is also a valid name).

Yes. I guess it's a better solution!

Yaron10:
@ArkadiuszMichalski's is good.

How about `!\Test\*+`?

donho:
@Yaron10 
> How about `!\Test\*+`?

It's a little bit too complex (IMO).
`!\1stLevelFolder` & `!+\allFolders` is simple & quite intuitive. It's easier to implement also.

ArkadiuszMichalski:
Any plan to support space in some special way? I haven't checked it, but maybe `Program*Files` will work at this point, it may not be ideal for every case, but it's worth knowing about it.

Yaron10:
> Any plan to support space in some special way?

That's a very good point.
I'm now realizing that perhaps this is what @alankilborn meant in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10927#issuecomment-998791511.

donho:
@Yaron10  @ArkadiuszMichalski 
Just commit the change, could you do some tests and confirm me it's OK to merge?
Thanks.

donho:
> Any plan to support space in some special way? I haven't checked it, but maybe Program*Files will work at this point, it may not be ideal for every case, but it's worth knowing about it.

There's no plan to support space in the pattern. `Program*Files` pattern works to match to "Program Files" so users have the way to do what they want.

> What should be easier to implement - !\Folder Name\*+ or !*+\"Folder Name"?

That will need to modify the function which cuts the Filters field string to put into the pattern arrays. I didn't see yet the feature request for that so far. OTOH, it's not the scope of this PR.

ArkadiuszMichalski:
@donho Works fine. But don't forget to update the description in the tooltip as well as the localization files.

Yaron10:
@donho,

It works like a charm. 👍 
Thanks again for your great work.

Neustradamus:
It is possible to have x86/x64 build to test it?

Thanks in advance.

donho:
32:
https://ci.appveyor.com/api/buildjobs/gqhdnphwy3k56cl7/artifacts/Notepad%2B%2B.Win32.Release.exe
64:
https://ci.appveyor.com/api/buildjobs/929tkboa0c1cmsu4/artifacts/Notepad%2B%2B.x64.Release.exe

Neustradamus:
Thanks, I will look it!

Have you planned a date for the next release?

PR #10956, opened 2021-12-27
Document List icons now support darkmode icons and alternate icons (Fix 2)

### LINKED ISSUE #10740: Icons for Document List unchanged in dark mode
### Description of the Issue
image
Notice that in dark mode, the pencil icon on the tab bar to the left changes to a symbolic icon, being solid white. This is something that I like a lot. Unfortunately, the document list icons do not change accordingly. I'd expect the document list icons to also look like symbolic icons of pencils, rather than red/blue save files.

Ashfaaq18:
fix #10740

fix demo:
DocListIconSupport_Fix

Note: when you toggle the app to dark mode or toggle alternate icons, the document list docking panel will reset back to the left side of notepad++. So if the panel was floating or docked to the right, it'll go back to its default position (left). This is due to it restarting.


thomassmith311:
Great, thank you!

PR #10958, opened 2021-12-27
Fix UTF-8 detected wrongly as TIS-620 issue

### LINKED ISSUE #940: UTF-8 section symbol (0xC2A7) invokes TIS-620 decoding
I noticed some strange behavior today with how Notepad++ handles encoding detection: if a file contains 1 or more section symbols (`§`) and no other non-ASCII characters, Notepad++ will decode the file using the [TIS-620](https://en.wikipedia.org/wiki/Thai_Industrial_Standard_620-2533) charset, resulting in each section symbol being decoded to `ยง`. This behavior is entirely inconsistent and does not seem to occur for any other non-ASCII Latin character, and will not occur if other non-ASCII symbols are present in the file.

Edit: I should mention this occurs on multiple machines. all with an English locale active.

### LINKED ISSUE #3172: Encoding problems - changes symbols on re-opening file
Hi,
I figured out n++ has some issues saving files properly.
I just lost an important password using a simple txt-file as some special characters changed when reopening the file.

As I'm using npp for more than 5 years now and never noticed this behavior before, I wonder what changed.
Might this be a buggy release? (7.3.3 32bit)
Or might there be a maleous plugin?

Installed plugins are:
-compare
-converter
-customize toolbar
-falling bricks
-indent by fold
-mime tools
-npp export
-nppFtp

### LINKED ISSUE #3188: UTF-8 encoding option goes wrong with file with misidentified encoding
### Description

I've got a file (will attach one which replicated this issue), which is almost entirely ASCII, but with a UTF-8 encoded character in it. This file does not have a leading BOM.

Notepad++ misidentifies the encoding as Thai TIS-620. By itself this is a minor inconvenience, but following this, the encoding options in the Encoding menu seem to end up broken.

1. Open the file.
2. Notepad++ misidentifies the encoding as TIS-620.
3. Select Encode in UTF-8 from the Encoding menu.

### Expected Behavior

Change the encoding to UTF-8 and display the character correctly.

### Actual Behavior

Encoding at the bottom of the screen flickers, and reverts to TIS-620, and the character is not displayed correctly.

Selecting a different encoding from the Character Sets (e.g. ISO 8859-1) from the menu does change the encoding, but obviously ends up displaying the multiple characters. There doesn't seem to be any option under the Character Sets submenu for forcing 'Unicode' as the character set.

### Debug Information

Notepad++ v7.3.3   (32-bit)
Build time : Mar  8 2017 - 03:37:37
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

[test.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/943687/test.zip)

### LINKED ISSUE #3588: (bug) Encoding change from UTF-8 to TIS-620 when saving and closing
### Description of the Issue
When I create and save a text file containing only the text "Хз." (in Russian), which has Windows string ending and UTF-8 w\o BOM encoding, it transforms to TIS-620 encoding, and containing text becomes "ะฅะท." Seems to not happen with longer texts inside files. Also doesn't happen if you change "Хз" to some other two letters, for ex. "Чо", "Иа", "Ха" (in Russian).
Checked on versions 7.3.3 (32bit) and 7.4.2 (64bit).
Plugins disabling makes no effect.
OS is Windows 7 64-bit.

1. Create a file with Windows string ending and UTF-8 w\o BOM encoding
2. Write "Хз." in it (in Russian)
3. Save and close
4. Reopen this file

### Debug info
Notepad++ v7.4.2   (64-bit)
Build time : Jun 18 2017 - 23:38:43
Path : D:\Program Files\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)

screen

### LINKED ISSUE #4932: Opened with wrong encoding
### Description of the Issue

I saved with only 今 (means "now" in Japanese) on the first line with UTF-8. Then, I re-opened the file. However, the file was opened as TIS-620.

1. [File] ->[New]
2. type 今 on the first line
3. save
4. [File] -> [Reload from Disk]

### Expected Behavior

open the file as encoding UTF-8
### Actual Behavior

encoding was changed to TIS-620
### Debug Information

Notepad++ v7.5.8   (64-bit)
Build time : Jul 23 2018 - 02:06:36
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : ComparePlugin.dll DSpellCheck.dll MarkdownViewerPlusPlus.dll NPPJSONViewer.dll PluginManager.dll PythonScript.dll mimeTools.dll NppConverter.dll

### LINKED ISSUE #8755: (bug) Encoding change from UTF-8 to TIS-620 when saving and closing (2)
### Description of the Issue
This bug is similar to the following issue:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3588

I have a text file in UTF-8 containing lots of text and the portuguese word "dá". When the file is closed and opened again, the program change the encoding of the file to TIS-640, buggin the word to "dรก-".

Just want to give this feedback, since uncheking "Autodetect character encoding" solves the problem.

1. Create a simple text file with UTF-8 encoding and the word "dá";
2. Save and close;
3. Reopen the file.

### Expected Behavior
That the file would still have UTF-8 encoding and the word as was written.

### Actual Behavior
The program changes the file encoding and bugs the word to "dรก-".

### Debug Information
Notepad++ v7.8.5   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 7 Professional (32-bit) 
OS Build : 7601.0

### LINKED ISSUE #10492: UTF-8 character not displayed
Notepad+++ 8.1.4 (64-bit)

Opening a text file with certain UTF-8 character, for example `û` does not detect it. It's recognized as Thai TIS-620, and changing Encoding to UTF-8 doesn't change to it. Example file attached below.

[small-u-circumflex.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/7101889/small-u-circumflex.zip)

### LINKED ISSUE #10916: Incorrect character set detection
Consider the following text:

```
Superscript[¹](#notes)

Blah...

...

#notes

 - This is an anchor.
```

The `¹` character in the text above is [Unicode Superscript One](https://www.compart.com/en/unicode/U+00B9) (`U+00B9`).

If you save this as an UTF8 file using latest version of Notepad++ on the latest version of Windows 10, then close and reopen the file it will be detected as Thai TIS-620 character set and `¹` character will be garbled:

image

It seems that the presence of non-alpha character (i.e. symbols or digits) immediately before `¹` is triggering the problem.

donho:
uchardet detects usually wrongly UTF-8 as TIS-620, hence TIS-620 detection is disabled in this commit.

More info:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10916#issuecomment-1001671957

Fix #10916, fix #940, fix #8755, fix #3588, fix #3188, fix #4932, fix #3172, fix #10492

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1229,11 +1229,13 @@ BufferID FileManager::bufferFromDocument(Document doc, bool dontIncrease, bool d
 
 int FileManager::detectCodepage(char* buf, size_t len)
 {
+	int codepage = -1;
 	uchardet_t ud = uchardet_new();
 	uchardet_handle_data(ud, buf, len);
 	uchardet_data_end(ud);
 	const char* cs = uchardet_get_charset(ud);
-	int codepage = EncodingMapper::getInstance().getEncodingFromString(cs);
+	if (stricmp(cs, "TIS-620") != 0) // TIS-620 detection is disabled here because uchardet detects usually wrongly UTF-8 as TIS-620
+		codepage = EncodingMapper::getInstance().getEncodingFromString(cs);
 	uchardet_delete(ud);
 	return codepage;
 }


alankilborn (commented):
I wouldn't call this a "fix".  What about users that work with TIS-620 and want it autodetected?

donho:
@alankilborn 
> I wouldn't call this a "fix". 

Agree. It's a workaround.

> What about users that work with TIS-620 and want it autodetected?

See:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10916#issuecomment-1001671957

hyoretsu:
Thank God someone with common sense fixed this issue for numerous UTF-8 users that were having this problem. (Definitely waaay more adopted than Thailand's char set)

There was even an issue as far back as 2015, 6 years ago, with this exact same bug.

Issue #10941, opened 2021-12-23
[MinGW] Support both MingGW multilib versions


ArkadiuszMichalski:
MinGW can come as a multilib version, which means that a given version can generate both 32-bit and 64-bit binary. In this variant one pack contains two kinds of library. We don't need have two completely separate tools and switching between them through the `PATH` variable.

MinGW multilib version is available on the locations:
[MinGW-W64 GCC-8.1.0 i686-posix-sjlj](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/8.1.0/threads-posix/sjlj/)
[MinGW-W64 GCC-8.1.0 x86_64-posix-sjlj ](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/sjlj/)
This is the same version of GCC 8.1.0 that appveyor uses, just a different exception handling (here `sjlj` for both types). We need only download one of them and we can generate 32-bit and 64-bit Notepad++. I checked both and they work fine. There are even newer versions, but I haven't checked them:
[Multilib Toolchains(Targetting Win32 and Win64)](https://sourceforge.net/projects/mingw-w64/files/Multilib%20Toolchains(Targetting%20Win32%20and%20Win64)/)

We can check if our version of gcc supports multilibs with the `-v` option, by invoking `gcc -v` (find `--enable-multilib` or `--disable-multilib` state).

Currently in the `makefile` it is possible to force generation of 32-bit Notepad ++ if we have 64-bit MinGW. But the reverse is not possible, I mean to generate 64-bit Notepad++ if we have 32-bit MingGW. This behaviour is controlled by explicitly passing the `TARGET_CPU` variable in the command line.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/30c586d94c71b41b708716750f63c908306d8aa4/PowerEditor/gcc/makefile#L85
To support oposite case we need only add one more condition that checking `x86_64` value in passing `TARGET_CPU` and set proper flags.

BTW, we can also add detection for multilib support, but that will complicate the conditions a bit (it seems unnecessary). I assume that if someone wants to use the `TARGET_CPU` variable, they know that his version of MinGW supports multilib.

donho:
> Currently in the makefile it is possible to force generation of 32-bit Notepad ++ if we have 64-bit MinGW.

[...]

PR #10962, opened 2021-12-28
Support both MinGW multilib versions 

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10941.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -55,8 +55,11 @@ MinGW-w64 can be downloaded [here](https://sourceforge.net/projects/mingw-w64/fi
 3. Run `mingw32-make`.
 4. The 32-bit or 64-bit `notepad++.exe` will be generated either in `bin.i686` or in `bin.x86_64` directory respectively, depending on the target CPU of the compiler — look for the full path to the resulting binary at the end of the build process.
 
+### Some additional information:
+
 * The directory containing `notepad++.exe` will also contain everything needed for Notepad++ to start.
 * To have a debug build just add `DEBUG=1` to the `mingw32-make` invocation above. The output directory then will be suffixed with `-debug`.
 * To see commands being executed add `VERBOSE=1` to the same command.
-* When switching between compilers or between release/debug modes, `mingw32-make clean` must be executed first.
-* If you are building the project outside of the `PowerEditor/gcc` directory, for example when using `-f` option, then the entire project path must not contain any spaces. Additionally, the path to `makefile` of this project should be listed as first.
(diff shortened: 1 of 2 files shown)


donho:
@ArkadiuszMichalski 
So the indicated package on build.md:
https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/seh/x86_64-8.1.0-release-posix-seh-rt_v6-rev0.7z
is not MinGW multilib package?

If not, it will better to use this link:
https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win64/Personal%20Builds/mingw-builds/8.1.0/threads-posix/sjlj/x86_64-8.1.0-release-posix-sjlj-rt_v6-rev0.7z
to replace the original one, for putting the advantage of this PR. What do you think?

ArkadiuszMichalski:
@donho
I checked MinGW  8.1.0 packages and only `sjlj` variant is multilib (for both case, 32/64 bit). I would not change the recommended versions in `build.md` because these are the same ones that `appveyor` uses. It will be easier to manage any error in the future (if someone reports them). so better to encourage testing in the same version.

Until now, the makefile had MinGW 64-bit multilib support, but the 32-bit variant was missing. so I added it (since it already partially exists). The additional information in `build.md` may be useful for people who have and using such versions permanently in the `Path` env. 

If in the future problems arise from using different versions (I mean onelib vs multilib) it is best not to support multilib at all. Well, I have not seen such a report yet, so it does not bother me. Sometimes it's easier and faster to generate a second version of Notepad++ without changing the `Path` to other MinGW variant. Personally, I have permanently added MinGW 32-bit sjlj to system `Path` and generation of 64-bit Notepad++ works fine.

donho:
@chcg 
Is it hard to change to MinGW sjlj W64 in our build system so we don't have to deal with 2 packages of MinGW ?
What's your opinion on it?

chcg:
At least there seems to be a performance penalty with sjlj compared to seh, see
https://stackoverflow.com/questions/30739099/what-is-the-difference-between-mingw-seh-and-mingw-sjlj/30739394.
I'm not sure how complicated it is to install the sjlj package on appveyor. Timing might be also of interest then and maybe caching.

ArkadiuszMichalski:
@donho 

I'm not sure if single package multilib will always work the same as 2 separate ones. Generally I'm not working on Notepad++ generated from MinGW. Separate versions MinGW are probably more promoted (more popular and available), especially in the new versions of MinGW.

This got me interested when I started analyzing the makefile from Notepad++. Multilib support has arrived in this new makefile that was made quite recently. Maybe Ivan U7n had some experience with it and wanted to be able to handle them as well (well, at least 64-bit packages), but only through `TARGET_CPU` variable.

donho:
@ArkadiuszMichalski @chcg 
OK, I see your points. Thank you for your answers.

PR #10334, opened 2021-08-05
Fix RTL inverse horizontal arrow key movement

Ashfaaq18:
Fix #8553, Fix #7678 and Fix #9730

Before the fix:

rtlArrowKeyFixbefore

After the fix:

rtlArrowKeyFixafter

PR #10963, opened 2021-12-28
Fix caret left-right inverse moving in R2L mode issue

### LINKED ISSUE #7678: arrows text cursor in Arabic RTL MODE is reversed
Write in Arabic and move the left and right arrows reversed, after switch to RTL.

The text cursor is supposed to go in the right direction but is reversed when i using RTL MODE.

if i press right arrow he gone to left, vice versa.

as you see here i press left arrow but he gone right !?
https://imgbbb.com/images/2019/11/30/arrr.gif

### LINKED ISSUE #8553: Problem in moving the arrow left and right in the Arabic language after RTL Mode

1. After CTRL+ALT+R to RTL Mode
2. everything working good but when i press like arrow left he moving right, Or the opposite is true!!?
3. he must moving right, right and left, left

### Expected Behavior

I think Arabic is well supported but unfortunately not complete..

### Debug Information

Notepad++ v7.8.7   (32-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll DSpellCheck.dll HexEditor.dll mimeTools.dll NppConverter.dll NppExport.dll NppTextFX.dll 

and

Notepad++ v7   (64-bit)
Build time : Sep 21 2016 - 22:07:28
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 7

### LINKED ISSUE #9730: Arrow keys moving caret in wrong direction in RTL mode
### Description of the Issue
When in RTL (Right to Left) mode, pressing the left and right keyboard arrow keys in Notepad++ causes the caret to move in the opposite direction. Left moves right, right moves left.

1. Open a new file in Notepad++.
2. Write any text in any language, RTL or otherwise (try "Hello World", "שלום עולם")
3. Switch to RTL mode (View->Text Direction RTL)
4. Press the arrow keys to move the caret.
5. The caret will move in the opposite direction

### Expected Behavior
Pressing the left arrow key should move the caret left (as the right arrow key currently does).
Pressing the right arrow key should move the caret right (as the left arrow key currently does).
Other text editors flip the keys correctly, including the default Windows notepad.
The arrows should definitely be flipped in RTL languages, and maybe in LTR languages as well (subject to debate).

### Actual Behavior
Pressing the right arrow key moves the caret left.
Pressing the left arrow key moves the caret right.

donho:
ref:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10334#issuecomment-1002054488

Fix #8553, fix #7678, fix #9730


Yaron10:
@donho,

Nice and clever! I was impressed. 👍 

Assigning the keys in `changeTextDirection()` - the behavior in the "Search Results" window is fixed as well.
And, obviously,  the code is shorter and the structure is better. 

Also, from now on - changing the text direction applies to both views.

Thank you.

donho:
@Yaron10 
> Also, from now on - changing the text direction applies to both views.

We will see the users' feedback. We may set an option if needed.

> I'm wondering about the key-assigning on each and every search. What do you think?

Well spotted. Fixed now.

Yaron10:
@donho,

I haven't tested it.
But - how about `if ((isRTL && isFinderR2L != isRTL) || (!isRTL && isFinderR2L == isRTL))`?
Also: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ea56b024535d37f31854cfddf3a0b4c11ed4385c/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2758.

Another option: implement something like `if ((toRTL && isRTL) || (!toRTL && !isRTL))` at the beginning of `changeTextDirection()` itself.
(And modify https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ea56b024535d37f31854cfddf3a0b4c11ed4385c/PowerEditor/src/NppCommands.cpp#L3477?).
I haven't thought about it sufficiently.

I'll continue tomorrow. Good night.

donho:
@Yaron10 
You're right. I didn't consider the case we switch from R2L to L2R.
Fixed now. Please let me know if it works correctly.
Thank you!

Yaron10:
@donho,

I might have confused you here. :)

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/98e7d79960833b7e826d8ca60830e38253d076c8/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2710

Should be `if ((toRTL && !isRTL) || (!toRTL && isRTL))`.
(Call `changeTextDirection()` only if there's a mismatch).

Thanks again.

Yaron10:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/98e7d79960833b7e826d8ca60830e38253d076c8/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2711

Should be `_pFinder->_scintView.changeTextDirection(toRTL);`.

Yaron10:
I don't have VS on the computer I'm using now, but the following change seems correct.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/98e7d79960833b7e826d8ca60830e38253d076c8/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2759

Should be:
```
if (isRTL)
   pFinder->_scintView.changeTextDirection(true);
```

donho:
@Yaron10 
New commit added, should be good now.
Thank you.

Yaron10:
@donho,

Great. 👍 

Thanks again for your work.

PR #10970, opened 2021-12-30
Update SUPPORTED_SYSTEM.md with Windows 11

thomassmith311:
Windows 11 is becoming available for more and more Windows computers. I propose that we add a slot on the table for Windows 11.


donho:
I will update it myself.

PR #10990, opened 2022-01-03
Advertise environment variables update at runtime

xomx:
Implemented #10749 .

Detect a global (user & system profile) environment variables change.

Advertise the change above in the N++ GUI by new menu command mark (& its tooltip) feature.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1,4 +1,4 @@
-﻿// This file is part of Notepad++ project
+// This file is part of Notepad++ project
 // Copyright (C)2021 Don HO <don.h@free.fr>
 
 // This program is free software: you can redistribute it and/or modify
@@ -740,6 +740,8 @@ LRESULT Notepad_plus::init(HWND hwnd)
 
 void Notepad_plus::killAllChildren()
 {
+	destroyMarkOnMenuItemById(MCM_ALL_MENU_COMMANDS);
+
 	_toolBar.destroy();
 	_rebarTop.destroy();
 	_rebarBottom.destroy();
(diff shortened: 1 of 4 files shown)


PowerEditor/src/Notepad_plus.h
```diff
@@ -636,4 +636,47 @@ friend class FileManager;
...
+		const TCHAR* _szMarkStr;			// single char like "!" or a very short string (the menuitem check-area is limited...)
+		COLORREF _clrfMark;					// foreground color of the above mark, use the CLR_INVALID here to use the current foreground menucolor
+		BYTE _alphaOpacity;					// 0 ... transparent, 255 ... opaque
+		SIZE _sizeMarkArea;					// used at runtime
+		const char* _szTooltipStrId;		// optional, it can be NULL (this is string ID for the NativeLangSpeaker)
+		const TCHAR* _szTooltipDefaultStr;	// optional, it can be NULL
+		HWND _hTooltip;						// used at runtime
+		TOOLINFO _tiTooltip;				// used at runtime
+		UINT _uMaxTooltipWidth;				// multiline wrapping point for longer tooltip strings
+	} MENUCOMMANDMARK, *LPMENUCOMMANDMARK;
```
donho, line 667: About the structure `st_MenuCommandMark`:
1. Coding standard of Notepad++ is not respected: the "type prefix" of variable naming is not used in Notepad++ code, whereas it's everywhere in this PR.
2. In modern C++ structure (no `typedef`), all the member variables can be initialized directly in the structure - it will improve the code readability.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -8102,3 +8104,243 @@ void Notepad_plus::updateCommandShortcuts()
...
+		}
+
+		NativeLangSpeaker* pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
+		if (!pNativeSpeaker)
+			return false;
+		generic_string tip2show = pNativeSpeaker->getLocalizedStrFromID(pMenuCommandMark->_szTooltipStrId, pMenuCommandMark->_szTooltipDefaultStr);
+		TCHAR szBuf[2048]; // we cannot use the tip2show.c_str() directly later
+		_sntprintf_s(szBuf, _countof(szBuf), _TRUNCATE, TEXT("%s"), tip2show.c_str());
+
+		pMenuCommandMark->_hTooltip = CreateWindowEx(pNativeSpeaker->isRTL() ? WS_EX_LAYOUTRTL : 0, TOOLTIPS_CLASS, NULL,
```
donho, line 8277: Since the intent of this PR is the code be part of Notepad++, the reuse of Notepad++ is necessary.
Using the function `CreateToolTip` of Notepad++ could make this PR shorter and more readable: 
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/Common/Common.cpp#L1015

donho:
@xomx 
I can see that you put a lot of effort in this PR, and though the code works, it doesn't work totally:
1. under High DPI:

image


2. The context menu has no ! mark, but has the tooltip:

image


Anyway, the provided PR works globally. However I have to decline this PR, for the following reasons:
1. The coding standard of Notepad++ is not respected
2. The code of Notepad++ is not reused while it is needed 
3. The modern C++ style is not used, which is important for the readability
4. (the most important) The amount of code (too important) is not proportional for the small feature as this one.
 
For the point 4, it's true that adding the symbols on the menu with the associated tooltip displayed is quite cool, but for the moment I don't see the need. 

Thank you for your coding effort anyway.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1,4 +1,4 @@
-﻿// This file is part of Notepad++ project
+// This file is part of Notepad++ project
```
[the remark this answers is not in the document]
xomx: Strange, I have no intention to edit this.

xomx:
> 4\. (the most important) The amount of code (too important) is not proportional for the small feature as this one

All the other reported problems could be probably fixed except this one, so I am leaving this idea.

Issue #8972, opened 2020-10-07
Translation is missing in several places [summary]


ArkadiuszMichalski:
### Description of the Issue
Still missing translation in several places:
~~1. Context menu in Shortcut Mapper: `Modify`, `Delete`, `Clear`. [PR #13120](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13120)~~ (done)

Other reported places:

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15426
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13107~~ ~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3450~~ (Update and Proxy Settings dialogs belong to another repo [WinGup](https://github.com/notepad-plus-plus/wingup))
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4274~~ (reject)
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8004 - still missing "length/lines/Ln/Col/Pos" in the status bar

~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15105~~  (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15375~~  (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13286~~ (reject) - Tooltips are not updated after changing translation (reason https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13288#issuecomment-1454937340)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13166~~ (reject)  - `Command Argument Help` dialog will not be translated (reason https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13166#issuecomment-1454836182).
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13175~~  (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8020~~  (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10985~~ (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10984~~ (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10983~~ (done)
~~https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10356~~ (done)
[...]

PR #11031, opened 2022-01-11
Make menu folders on context menu translatable

### LINKED ISSUE #9713: Three text contextual options are not localized
Hello,

I just noticed that three text contextual options are not localized: "Style token", "Remove style" and "Plugin commands".

In the screenshot below, you can see that everything is translated in Corsican except these three options.

Notepad++ Two text contextual options are not localized

In the Search menu, two options - `Mark All` > `Using 1st [to 5th] Style` and `Unmark All` > `Clear 1st [to 5th] Style` - are similar to the first two above and properly localized.

Thanks,
Patriccollu.

donho:
Make the following context menu items translatable:
1. "Style all occurrences of token"
2. "Style one token"
3. "Clear style"
4. "Plugin commands"

**Note for translators: to test your translation, you have to replace the old "contextMenu.xml" with the new one (modified in this commit).**

Fix partially #8972
Fix #9713


ArkadiuszMichalski:
Works as as expected:
image

PR #11037, opened 2022-01-13
Add Find/Replace text swap button in Find replace dialog

### LINKED ISSUE #6010: [IMPROVEMENT] Find and Replace Dialog
### Description of the Issue
What about to add a new button in **Find and Replace** dialog and in **Find and Replace in Files** dialog to switch the text content of the **Find what** text field with the **Replace with** text field?

The aim is to facilitate the user which needs to invert the text content of those fields. In fact at this moment in case you need to do the operation you have to do a lot of steps that make the operation really tedious, especially if you need to do it a lot of times for particular editing of the text file(s) you are modifying:

1. select the text of the **Find** field
2. copy/cut the text of that field
3. paste it elsewhere
4. select the text of the **Replace** field
5. copy/cut the text of that field
6. paste it into the **Find** field
7. select the text of the **Find** field previously pasted elsewhere
8. copy/cut that text
9. paste it into the **Replace** field

There are nine passages!!! IMHO it's way too much for a so simple operation that can be done automatically with a click on a button. Unfortunately I'm one that needs to do those operations really a lot of times per day so a button placed right there could facilitate a lot my complex editing and find and replace operations.

Look at the attached image file to see the aspect of the button I propose to implement: as simple as that:

Notepad++ Find Replace Form New Button
Notepad++ Find Replace in Files Form New Button

Thanks for comprehension.

donho:
Fix  #6010


xomx:
@donho 
I had a funny moment when I tried appveyor build with this PR. 1st everything was ok, then I switched language to my Czech in Preferences and when I got back after a while, the little new button had been mangled somehow (see the pic below). Switching back to English did not help. So I restarted N++ and tried other languages as well, but only Czech had this problem. Then with a little help of WinSpy I read the mangled text in button and I could identify an old remnant in the czech.xml file (it had no equivalent in current english.xml).
wrong_res_id

donho:
@xomx 
Well catched! The conflict with the obsolete translation is avoided by the last commit.
Thank you for your heads up!!

PR #11044, opened 2022-01-14
Remove 2GB restriction for 64 bit binary

### LINKED ISSUE #10921: [FR] Support editing of files greater than 2GB.
The software currently cannot edit very large files (i.e. > 2GB). This makes the tool unusable for tasks that included large files (e.g. Data Science). Please enhance the software to handle large files.

donho:
Fix #10921

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1353,7 +1353,7 @@ bool FileManager::loadFileData(Document doc, const TCHAR * filename, char* data,
 		if (nppParam.archType() == IMAGE_FILE_MACHINE_I386)
 		{
 			pNativeSpeaker->messageBox("FileTooBigToOpen",
-				NULL,
+				_pNotepadPlus->_pEditView->getHSelf(),
 				TEXT("File is too big to be opened by Notepad++"),
 				TEXT("File size problem"),
 				MB_OK | MB_APPLMODAL);
@@ -1364,7 +1364,7 @@ bool FileManager::loadFileData(Document doc, const TCHAR * filename, char* data,
 		else // x64
 		{
 			int res = pNativeSpeaker->messageBox("WantToOpenHugeFile",
-				NULL,
+				_pNotepadPlus->_pEditView->getHSelf(),
 				TEXT("Opening a huge file of 2GB could take more than 3 minutes.\nDo you want to open it?\nNote that Word Wrap feature will be turned OFF before loading file."),
 				TEXT("Opening huge file warning"),
 				MB_YESNO | MB_APPLMODAL);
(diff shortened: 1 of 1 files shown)


donho:
Tested with the same  2.04 GB file, the result is quite random - loading time from almost 3 minutes to almost 6 minutes.
Not sure if this removal restriction is useful.

donho:
After opening a  2.04 GB file, we can close only the file - it's instantly and won't have any problem.
However, if we exit Notepad++ without closing this huge file, Notepad++ will hang.
There's no meaning to allow users to open any huge file in Notepad++ IMO.

rddim:
Please, update the `english.xml` and add dark mode support :) 

Изображение 1
Изображение 2

dfs-:
@donho is the new message text translatable?

> Please, update the `english.xml` and add dark mode support :)
> 
> Изображение 1 Изображение 2

You should be able to use tags OpenHugeFileButWrapIsOn and WantToOpenHugeFile in your language.xml; just search for "FileTooBigToOpen" and then add those two below them in the same fashion.

It's XPath is `/NotepadPlus/Native-Langue/MessageBox/FileTooBigToOpen`.

rddim:
@dfs- 

Thanks, I know that, but all languages are based on english and for me (in this case) is better to create one PR for all instead of new just for the locale.

donho:
@rddim 
Due to the very poor performanceon a 2GB+ file, this PR is not going to be merged.

ArkadiuszMichalski:
@donho What file did you test? I check this PR and on my machine for 2GB it take 7s.

donho:
@ArkadiuszMichalski 
> What file did you test? I check this PR and on my machine for 2GB it take 7s, 3GB 7s, 4GB 15s.

I tested with movie files (MKV or AVI).
They are opened extremely quickly in SciTE 5.1.6

ArkadiuszMichalski:
Content is correct, other editors display text. I use some bat to generate file, like this:
```
@echo off
echo "This is just a sample line appended  to create a big file. " > dummy2GB.txt
for /L %%i in (1,1,25) do type dummy2GB.txt >> dummy2GB.txt
```
```
@echo off
echo "This is just a sample line appended  to create a big file. " > dummy4GB.txt
for /L %%i in (1,1,26) do type dummy4GB.txt >> dummy4GB.txt
```

```
@echo off
echo "This is just a sample line appended  to create a big file. " > _dummy3GB.txt
for /L %%i in (1,1,24) do type _dummy3GB.txt >> _dummy3GB.txt
for /L %%i in (1,1,3) do type _dummy3GB.txt >> dummy3GB.txt
del _dummy3GB.txt
```

But even when Notepad++ opens such files there is a problem with editing, it scrolls to the end. Editing in SciTE works.

donho:
@ArkadiuszMichalski 
It seems there are a bug (for your test) and performance issue in the v8.2 (at least).
I just tried Notepad++ v8.1.5 portable version, the loading time for 1.5GB took only 3-4 seconds, whereas in v8.2 with the same file, if Notepad++ doesn't hang, it takes 3-7 minutes!

ArkadiuszMichalski:
For me there is no problem with version 8.2, file 1.5GB loads 5s.

donho:
Culprit found: it's due to `NppLspClient` plugin.
OK, then I will redo a PR for the removal of 2GB restriction.
Thank you @ArkadiuszMichalski  to put me in the right direction!!

donho:
@ArkadiuszMichalski 
I confirm NULL character at the end of file (happen to me now). It could be the same effect of Exchange 2022 mentioned by @chcg 
I'm gonna take a look at it.

Issue #8802, opened 2020-09-02
Opening 620 MB SQL file leaves it very slow


luapfr:
I’m trying to manipulate some SQL or XML files but they are large files and notepad ++ crashes.

My computer has a 
Windows 10 64 Bits
Core I7 8750h 32GB RAM
1TB SSD
I’m using Notepad ++ 7.8.8 64 Bits

I did the forum procedures where I ran the same without a plugin
https://community.notepad-plus-plus.org/topic/19939/opening-620-mb-sql-file-leaves-it-very-slow/6

Notepad++ v7.8.8   (64-bit)
Build time : Jun 23 2020 - 23:59:56
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2004
OS Build : 19041.450
Current ANSI codepage : 1252
Plugins : none

dail8859:
Since renaming the file to txt solved the issue, this points the SQL Lexer which is part of the [Scintilla project](https://scintilla.org/). You will have to open an issue with that project for them to look into this.

luapfr:
it is not just SQL if it is XML or any other file format that is large gives the same problem.

dail8859:
Again if a regular text file works then it still indicates its whatever lexer is being used (SQL, XML, etc) so it is still part of the Scintilla project.

luapfr:
Ok

https://sourceforge.net/p/scintilla/bugs/2200/

linquize:
Are the lines too long?

luapfr:
I tested
SQL
XML
JSON
CSV

They all give the same problem
in the case in XML the longest line is 200 characters

In SQL the longest line is 237 characters long

luapfr:
Attached an XML file and a Video showing the Problem.

https://mega.nz/file/Buol0YyB#nHXTOSQFwI1LfmZLRfHJ3eL8L691Pkr21D89arbk81c

luapfr:
[...]

PR #11047, opened 2022-01-16
Remove 2GB file open restriction for x64 binary

### LINKED ISSUE #10921: [FR] Support editing of files greater than 2GB.

donho:
This PR is still in progress, because for supporting 2GB or larger file, it needs to use int64 as integer type, for all the data structures which deal with Scintilla, especially for searching features. It's not enough to remove only the restriction.

The current state of this PR can open 1GB (that 32bits binary cannot even open it) to 5GB (tested). 
All the files under 2G opened flawlessly. However, once the file size is more than 2GB, file can be opened, but with plenty of NUL characters at the end of file, as mention in another PR:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11044#issuecomment-1013236115
 I have still no idea how to fix it. If anyone has any idea here, please let me know.    

Fix #10921

### DIFF
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -360,8 +360,8 @@ private:
 	//Synchronized Scolling
 	struct SyncInfo final
 	{
-		INT_PTR _line = 0;
-		INT_PTR _column = 0;
+		intptr_t _line = 0;
+		intptr_t _column = 0;
 		bool _isSynScollV = false;
 		bool _isSynScollH = false;
 
@@ -372,7 +372,7 @@ private:
 	bool _isUDDocked = false;
 
 	trayIconControler* _pTrayIco = nullptr;
-	INT_PTR _zoomOriginalValue = 0;
+	intptr_t _zoomOriginalValue = 0;
 
 	Accelerator _accelerator;
 	ScintillaAccelerator _scintaccelerator;
@@ -496,13 +496,13 @@ private:
 		::CheckMenuItem(_mainMenuHandle, itemID, MF_BYCOMMAND | (willBeChecked?MF_CHECKED:MF_UNCHECKED));
 	}
 
-	bool isConditionExprLine(INT_PTR lineNumber);
-	INT_PTR findMachedBracePos(size_t startPos, size_t endPos, char targetSymbol, char matchedSymbol);
+	bool isConditionExprLine(intptr_t lineNumber);
+	intptr_t findMachedBracePos(size_t startPos, size_t endPos, char targetSymbol, char matchedSymbol);
 	void maintainIndentation(TCHAR ch);
 
 	void addHotSpot(ScintillaEditView* view = NULL);
 
-    void bookmarkAdd(INT_PTR lineno) const {
+    void bookmarkAdd(intptr_t lineno) const {
 		if (lineno == -1)
 			lineno = _pEditView->getCurrentLineNumber();
 		if (!bookmarkPresent(lineno))
@@ -516,14 +516,14 @@ private:
 			_pEditView->execute(SCI_MARKERDELETE, lineno, MARK_BOOKMARK);
 	}
 
-    bool bookmarkPresent(INT_PTR lineno) const {
+    bool bookmarkPresent(intptr_t lineno) const {
 		if (lineno == -1)
 			lineno = _pEditView->getCurrentLineNumber();
 		LRESULT state = _pEditView->execute(SCI_MARKERGET, lineno);
 		return ((state & (1 << MARK_BOOKMARK)) != 0);
 	}
 
-    void bookmarkToggle(INT_PTR lineno) const {
+    void bookmarkToggle(intptr_t lineno) const {
 		if (lineno == -1)
 			lineno = _pEditView->getCurrentLineNumber();
 
@@ -546,7 +546,7 @@ private:
 	void inverseMarks();
 	void replaceMarkedline(size_t ln, const TCHAR *str);
 	generic_string getMarkedLine(size_t ln);
-    void findMatchingBracePos(INT_PTR& braceAtCaret, INT_PTR& braceOpposite);
+    void findMatchingBracePos(intptr_t& braceAtCaret, intptr_t& braceOpposite);
     bool braceMatch();
 
     void activateNextDoc(bool direction);
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -360,8 +360,8 @@ private:
 	//Synchronized Scolling
 	struct SyncInfo final
 	{
-		int _line = 0;
-		int _column = 0;
+		INT_PTR _line = 0;
+		INT_PTR _column = 0;
 		bool _isSynScollV = false;
 		bool _isSynScollH = false;
 
@@ -496,8 +496,8 @@ private:
 		::CheckMenuItem(_mainMenuHandle, itemID, MF_BYCOMMAND | (willBeChecked?MF_CHECKED:MF_UNCHECKED));
 	}
 
-	bool isConditionExprLine(int lineNumber);
-	int findMachedBracePos(size_t startPos, size_t endPos, char targetSymbol, char matchedSymbol);
+	bool isConditionExprLine(INT_PTR lineNumber);
+	INT_PTR findMachedBracePos(size_t startPos, size_t endPos, char targetSymbol, char matchedSymbol);
 	void maintainIndentation(TCHAR ch);
 
 	void addHotSpot(ScintillaEditView* view = NULL);
@@ -546,7 +546,7 @@ private:
 	void inverseMarks();
 	void replaceMarkedline(size_t ln, const TCHAR *str);
 	generic_string getMarkedLine(size_t ln);
-    void findMatchingBracePos(int & braceAtCaret, int & braceOpposite);
+    void findMatchingBracePos(INT_PTR& braceAtCaret, INT_PTR& braceOpposite);
     bool braceMatch();
 
     void activateNextDoc(bool direction);
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -502,7 +502,7 @@ private:
 
 	void addHotSpot(ScintillaEditView* view = NULL);
 
-    void bookmarkAdd(int lineno) const {
+    void bookmarkAdd(INT_PTR lineno) const {
 		if (lineno == -1)
 			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
 		if (!bookmarkPresent(lineno))
@@ -523,7 +523,7 @@ private:
 		return ((state & (1 << MARK_BOOKMARK)) != 0);
 	}
 
-    void bookmarkToggle(int lineno) const {
+    void bookmarkToggle(INT_PTR lineno) const {
 		if (lineno == -1)
 			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
 
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -372,7 +372,7 @@ private:
 	bool _isUDDocked = false;
 
 	trayIconControler* _pTrayIco = nullptr;
-	int _zoomOriginalValue = 0;
+	INT_PTR _zoomOriginalValue = 0;
 
 	Accelerator _accelerator;
 	ScintillaAccelerator _scintaccelerator;
@@ -509,16 +509,16 @@ private:
 			_pEditView->execute(SCI_MARKERADD, lineno, MARK_BOOKMARK);
 	}
 
-    void bookmarkDelete(int lineno) const {
+    void bookmarkDelete(size_t lineno) const {
 		if (lineno == -1)
 			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
 		while (bookmarkPresent(lineno))
 			_pEditView->execute(SCI_MARKERDELETE, lineno, MARK_BOOKMARK);
 	}
 
-    bool bookmarkPresent(int lineno) const {
+    bool bookmarkPresent(INT_PTR lineno) const {
 		if (lineno == -1)
-			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
+			lineno = _pEditView->getCurrentLineNumber();
 		LRESULT state = _pEditView->execute(SCI_MARKERGET, lineno);
 		return ((state & (1 << MARK_BOOKMARK)) != 0);
 	}
@@ -542,10 +542,10 @@ private:
 	void cutMarkedLines();
 	void deleteMarkedLines(bool isMarked);
 	void pasteToMarkedLines();
-	void deleteMarkedline(int ln);
+	void deleteMarkedline(size_t ln);
 	void inverseMarks();
-	void replaceMarkedline(int ln, const TCHAR *str);
-	generic_string getMarkedLine(int ln);
+	void replaceMarkedline(size_t ln, const TCHAR *str);
+	generic_string getMarkedLine(size_t ln);
     void findMatchingBracePos(int & braceAtCaret, int & braceOpposite);
     bool braceMatch();
 
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -504,14 +504,14 @@ private:
 
     void bookmarkAdd(INT_PTR lineno) const {
 		if (lineno == -1)
-			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
+			lineno = _pEditView->getCurrentLineNumber();
 		if (!bookmarkPresent(lineno))
 			_pEditView->execute(SCI_MARKERADD, lineno, MARK_BOOKMARK);
 	}
 
     void bookmarkDelete(size_t lineno) const {
 		if (lineno == -1)
-			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
+			lineno = _pEditView->getCurrentLineNumber();
 		while (bookmarkPresent(lineno))
 			_pEditView->execute(SCI_MARKERDELETE, lineno, MARK_BOOKMARK);
 	}
@@ -525,7 +525,7 @@ private:
 
     void bookmarkToggle(INT_PTR lineno) const {
 		if (lineno == -1)
-			lineno = static_cast<int32_t>(_pEditView->getCurrentLineNumber());
+			lineno = _pEditView->getCurrentLineNumber();
 
 		if (bookmarkPresent(lineno))
 			bookmarkDelete(lineno);
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -647,14 +647,14 @@ BufferID FileManager::loadFile(const TCHAR * filename, Document doc, int encodin
 	if (!fp)
 		return BUFFER_INVALID;
 	_fseeki64(fp, 0, SEEK_END);
-	size_t fileSize = _ftelli64(fp);
+	int64_t fileSize = _ftelli64(fp);
 	fclose(fp);
-
+	
 	bool ownDoc = false;
 	if (!doc)
 	{
 		// If file exceeds 200MB, activate large file mode
-		doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT, 0, fileSize < (200 * 1024 * 1024) ? 0 : SC_DOCUMENTOPTION_STYLES_NONE | SC_DOCUMENTOPTION_TEXT_LARGE);
+		doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT, 0, fileSize < NPP_STYLING_FILESIZE_LIMIT ? 0 : SC_DOCUMENTOPTION_STYLES_NONE | SC_DOCUMENTOPTION_TEXT_LARGE);
 		ownDoc = true;
 	}
 
@@ -751,7 +751,7 @@ bool FileManager::reloadBuffer(BufferID id)
 	if (!fp)
 		return false;
 	_fseeki64(fp, 0, SEEK_END);
-	size_t fileSize = _ftelli64(fp);
+	int64_t fileSize = _ftelli64(fp);
 	fclose(fp);
 
 	bool res = loadFileData(doc, fileSize, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
@@ -1352,14 +1352,14 @@ LangType FileManager::detectLanguageFromTextBegining(const unsigned char *data, 
 	return L_TEXT;
 }
 
-bool FileManager::loadFileData(Document doc, size_t fileSize, const TCHAR * filename, char* data, Utf8_16_Read * unicodeConvertor, LoadedFileFormat& fileFormat)
+bool FileManager::loadFileData(Document doc, int64_t fileSize, const TCHAR * filename, char* data, Utf8_16_Read * unicodeConvertor, LoadedFileFormat& fileFormat)
 {
 	FILE *fp = generic_fopen(filename, TEXT("rb"));
 	if (!fp)
 		return false;
 
 	// size/6 is the normal room Scintilla keeps for editing, but here we limit it to 1MiB when loading (maybe we want to load big files without editing them too much)
-	size_t bufferSizeRequested = fileSize +min(1 << 20, fileSize / 6);
+	int64_t bufferSizeRequested = fileSize +min(1 << 20, fileSize / 6);
 	
 	NppParameters& nppParam = NppParameters::getInstance();
 	NativeLangSpeaker* pNativeSpeaker = nppParam.getNativeLangSpeaker();
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -515,7 +515,7 @@ int Buffer::removeReference(ScintillaEditView * identifier)
 }
 
 
-void Buffer::setHideLineChanged(bool isHide, int location)
+void Buffer::setHideLineChanged(bool isHide, size_t location)
 {
 	//First run through all docs without removing markers
 	for (int i = 0; i < _references; ++i)
@@ -937,7 +937,7 @@ bool FileManager::backupCurrentBuffer()
 
 			if (UnicodeConvertor.openFile(fullpath))
 			{
-				int lengthDoc = _pNotepadPlus->_pEditView->getCurrentDocLen();
+				size_t lengthDoc = _pNotepadPlus->_pEditView->getCurrentDocLen();
 				char* buf = (char*)_pNotepadPlus->_pEditView->execute(SCI_GETCHARACTERPOINTER);	//to get characters directly from Scintilla buffer
 				boolean isWrittenSuccessful = false;
 
@@ -950,8 +950,8 @@ bool FileManager::backupCurrentBuffer()
 				else
 				{
 					WcharMbcsConvertor& wmc = WcharMbcsConvertor::getInstance();
-					int grabSize;
-					for (int i = 0; i < lengthDoc; i += grabSize)
+					size_t grabSize;
+					for (size_t i = 0; i < lengthDoc; i += grabSize)
 					{
 						grabSize = lengthDoc - i;
 						if (grabSize > blockSize)
@@ -959,7 +959,7 @@ bool FileManager::backupCurrentBuffer()
 
 						int newDataLen = 0;
 						int incompleteMultibyteChar = 0;
-						const char *newData = wmc.encode(SC_CP_UTF8, encoding, buf+i, grabSize, &newDataLen, &incompleteMultibyteChar);
+						const char *newData = wmc.encode(SC_CP_UTF8, encoding, buf+i, static_cast<int>(grabSize), &newDataLen, &incompleteMultibyteChar);
 						grabSize -= incompleteMultibyteChar;
 						isWrittenSuccessful = UnicodeConvertor.writeFile(newData, static_cast<unsigned long>(newDataLen));
 					}
@@ -1063,7 +1063,7 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 	{
 		_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, buffer->_doc);	//generate new document
 
-		int lengthDoc = _pscratchTilla->getCurrentDocLen();
+		size_t lengthDoc = _pscratchTilla->getCurrentDocLen();
 		char* buf = (char*)_pscratchTilla->execute(SCI_GETCHARACTERPOINTER);	//to get characters directly from Scintilla buffer
 		boolean isWrittenSuccessful = false;
 
@@ -1082,8 +1082,8 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 			}
 			else
 			{
-				int grabSize;
-				for (int i = 0; i < lengthDoc; i += grabSize)
+				size_t grabSize;
+				for (size_t i = 0; i < lengthDoc; i += grabSize)
 				{
 					grabSize = lengthDoc - i;
 					if (grabSize > blockSize)
@@ -1091,7 +1091,7 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 
 					int newDataLen = 0;
 					int incompleteMultibyteChar = 0;
-					const char* newData = wmc.encode(SC_CP_UTF8, encoding, buf + i, grabSize, &newDataLen, &incompleteMultibyteChar);
+					const char* newData = wmc.encode(SC_CP_UTF8, encoding, buf + i, static_cast<int>(grabSize), &newDataLen, &incompleteMultibyteChar);
 					grabSize -= incompleteMultibyteChar;
 					isWrittenSuccessful = UnicodeConvertor.writeFile(newData, static_cast<unsigned long>(newDataLen));
 				}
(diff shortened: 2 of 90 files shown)


PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1342,26 +1342,25 @@ bool FileManager::loadFileData(Document doc, const TCHAR * filename, char* data,
...
+	unsigned __int64 bufferSizeRequested = fileSize + min(1<<20, fileSize/6);
 	
 	NppParameters& nppParam = NppParameters::getInstance();
 	NativeLangSpeaker* pNativeSpeaker = nppParam.getNativeLangSpeaker();
 
-	// As a 32bit application, we cannot allocate 2 buffer of more than INT_MAX size (it takes the whole address space).
-	// As a 64bit binary, we have more address for the allocation. However loading a 2GB file takes from 3 minutes to 7 minutes, which makes Notepad++ unusable:
-	// https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11044
-	// So here we fix the size limit to 2GB
 	if (bufferSizeRequested > INT_MAX)
```
mere-human, line 1350: Should this be changed as well?

donho: @mere-human 
It should be kept for 32bits binary.

xomx:
The change is huge! And what about plugins...

Here are some, you have not addressed yet:

.\PowerEditor\src\ScintillaComponent\FindReplaceDlg.cpp
- remove static_cast<int32_t> in Finder::notify() at line 500 (v8.2):
  pos = static_cast<int32_t>(_scintView.execute(SCI_GETLINEENDPOSITION, notification->line));
  
.\PowerEditor\src\ScintillaComponent\FunctionCallTip.h (and FunctionCallTip.cpp)
- redeclare ints _curPos, _startPos and _currentParam to size_t
- and adjust appropriately in FunctionCallTip.cpp: FunctionCallTip::updateCalltip(), FunctionCallTip::showCalltip(), FunctionCallTip::getCursorFunction()

.\PowerEditor\src\NppBigSwitch.cpp
- remove static_cast<long> from lParam in Notepad_plus::process() -> case WM_FRSAVE_INT:
  _macro.push_back(recordedMacroStep(static_cast<int32_t>(wParam), 0, static_cast<long>(lParam), NULL, recordedMacroStep::mtSavedSnR));

chcg:
From https://www.scintilla.org/ScintillaHistory.html

> External interfaces use the Sci_Position and Sci_PositionU typedefs instead of int and unsigned int to allow for changes to a 64-bit interface on 64-bit platforms in the future. Applications and external lexers should start using the new type names so that they will be compatible when the 64-bit change occurs. There is also Sci_PositionCR (long) for use in the Sci_CharacterRange struct which will also eventually become 64-bit. 

Shouldn't we use Scintilla defines instead of Windows or C types more often?

xomx:
There seems to be a big difference in memory consumption/handling in between N++ and Scite. Scite apparently does not need 2x filesize memory buffer (tested with 5.26GB text file, moreover N++ needs another 512MB big buffer for the same file):

scite-bigfile-memory
npp-bigfile-memory

xomx:
SetDlgItemInt/GetDlgItemInt cannot be used for big integers, e.g. look at the now incorrect `(UINT)limit` here:

```
void GoToLineDlg::updateLinesNumbers() const 
{
	size_t current = 0;
	size_t limit = 0;
	
	if (_mode == go2line)
	{
		current = (*_ppEditView)->getCurrentLineNumber() + 1;
		limit = (*_ppEditView)->execute(SCI_GETLINECOUNT);
	}
	else
	{
		current = (*_ppEditView)->execute(SCI_GETCURRENTPOS);
		size_t currentDocLength = (*_ppEditView)->getCurrentDocLen();
		limit = (currentDocLength > 0 ? currentDocLength - 1 : 0);
	}
    ::SetDlgItemInt(_hSelf, ID_CURRLINE, (UINT)current, FALSE);
    ::SetDlgItemInt(_hSelf, ID_LASTLINE, (UINT)limit, FALSE);
}
```


> once the file size is more than 2GB, file can be opened, but with plenty of NUL characters

If 'Line' mode is active in the 'Go To...' dlg, the `getCurrentLineNumber()` returns nonsense in the `GoToLineDlg::updateLinesNumbers()`. To test:
- open a file bigger than 4GB
- after successful opening in N++, do not move the cursor from the start position
- go to the Search- > Go to...
- see there, that the 'You are here' is not 1.

The reason is, that the `execute(SCI_LINEFROMPOSITION, 0)` call is returning nonsense somehow. There should be some invalid Scintilla setting for this in N++ and I think that this could be responsible for the NUL-chars problem described above. While scrolling thru a bigfile in N++, sometimes I see invalid CR/LF chars before I get the NUL-chars and then the N++ often freezes and when I Debug->BreakAll in MSVS, I see in callstack, that there is the SCI_LINEFROMPOSITION used before.

If you switch to 'Offset'mode in 'Go To...' dlg, both execute(SCI_GETCURRENTPOS) and getCurrentDocLen() works ok, but then the `limit` is truncated in `::SetDlgItemInt(_hSelf, ID_LASTLINE, (UINT)limit, FALSE)`, so you will not see the correct 'You can't go further than'.

donho:
@xomx 
>  Here are some, you have not addressed yet:

Done. Thank you. Please let me know if you find the other ones.

> The change is huge!

Yes, I know. But if we want Notepad++ allows users to open 2GB+ files, these changes are necessary. I believe you have realized it...

> And what about plugins...

I did check the API of Notepad++ here:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
And I didn't see any imcompability/change. If you see anything, please let me know.
For 2 Sintilla handles in plugins, plugin authors access to original Scintilla document (on Internet or in the distribution of Notepad++ source code). They should know how to use the API of Scintilla correctly (not like me :) ). At the worst case, their plugins wont work on 2GB+ files, which would be the rare case because people use Notepad++ for source files.

donho:
@chcg 
> Shouldn't we use Scintilla typedefs instead of Windows or C types more often?

Are you talking for the plugins? If yes, it's not necessary IMO, here is my reply about plugin: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11047#issuecomment-1014758441

Otherwise, I prefer to use the system provided prototype - One day human being has 128 bits machine, I believe INT_PTR & size_t will support upto 128bits :)

donho:
@xomx 
> There seems to be a big difference in memory consumption/handling in between N++ and Scite. Scite apparently does not need 2x filesize memory buffer (tested with 5.26GB text file, moreover N++ needs another 512MB big buffer for the same file):
(retested with Sc446.exe to have the same version of Scintilla, the result is the same)

Very interesting. I will investigate this fact.

donho:
@xomx 
> The reason is, that the execute(SCI_LINEFROMPOSITION, 0) call is returning nonsense somehow. There should be some invalid Scintilla setting for this in N++ and I think that this could be responsible for the NUL-chars problem described above. 

Indeed. Not only SCI_LINEFROMPOSITION call is returning nonsense, but also SCI_GETCOLUMN returns always 2. I thought it was the lake of buffer room (https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1399), but even I doubled its room, I've got the same result. Except the allocated room, I don't think there are other "invalid Scintilla setting". 

OTOH, it could be some of stacks corrupted, because of the important data loaded, which corrupts also Scintilla's data, make its dysfunction.

xomx:
- in `FunctionListPanel::addEntry` the `_i64toa` is needed under 64-bit and also removing of the casting:
  generic_itoa(`static_cast<int32_t>`(pos), posStr, 10);
- in` ScintillaEditView::activateBuffer` the _i64toa is needed under 64-bit and also removing of the casting:
  itoa(`static_cast<int>`(numLines), numLineStr, 10);   (and isn't it a remnant code there?)
- in `AutoCompletion::getWordArray` remove casting:
  posFind = _pEditView->searchInTarget(expr.c_str(), `static_cast<int32_t>`(expr.length()), wordEnd, docLength);
- in `FileManager::loadFileData` there is the following filesize (`lenFile`) truncation):
  const char *newData = wmc.encode(fileFormat._encoding, SC_CP_UTF8, data, `static_cast<int32_t>`(lenFile), &newDataLen, &incompleteMultibyteChar);
- in `Finder::notify` invalid casting:
  pos = `static_cast<int32_t>`(_scintView.execute(SCI_GETLINEENDPOSITION, notification->line));
- in `FunctionListPanel::getBodyClosePos` invalid casting:
  targetEnd = `static_cast<int32_t>`(begin);
- in .\PowerEditor\src\Notepad_plus.h remove invalid casting in `bookmarkAdd`, `bookmarkDelete` and `bookmarkToggle`:
  lineno = `static_cast<int32_t>`(_pEditView->getCurrentLineNumber());
- in .\PowerEditor\src\NppBigSwitch.cpp -> `process()`:
  `case WM_FRSAVE_INT` and `WM_FRSAVE_STR` ... remove `static_cast<long>` from `lParam`
- in .\PowerEditor\src\NppCommands.cpp -> `command()` -> `case IDM_TOOL_SHA256_GENERATEINTOCLIPBOARD:` ... invalid `strLen` type and remove casting:
  `int32_t` strLen = `static_cast<int32_t>`(selectionEnd - selectionStart);

xomx:
@donho 
> OTOH, it could be some of stacks corrupted, because of the important data loaded, which corrupts also Scintilla's data, make its dysfunction.

Yes, it could be that. I followed the SCI_LINEFROMPOSITION deep into the Scintilla code and the error originates from .\scintilla\src\Partitioning.h (see the PROBLEM: places):

```
	T PartitionFromPosition(T pos) const noexcept {
		if (body->Length() <= 1)
			return 0;
		if (pos >= (PositionFromPartition(Partitions())))
			return Partitions() - 1;
		T lower = 0;
		T upper = Partitions();
		do {
			const T middle = (upper + lower + 1) / 2; 	// Round high
			T posMiddle = body->ValueAt(middle);  ->  PROBLEM: posMiddle becomes negative...
			if (middle > stepPartition)
				posMiddle += stepLength;
			if (pos < posMiddle) {
				upper = middle - 1;
			} else {
				lower = middle;
			}
		} while (lower < upper);
		return lower;  ->  PROBLEM: returns 84824893 instead of 0
	}
```

donho:
> There seems to be a big difference in memory consumption/handling in between N++ and Scite. Scite apparently does not need 2x filesize memory buffer (tested with 5.26GB text file, moreover N++ needs another 512MB big buffer for the same file):

Indeed. SciTE uses only the needed memory whereas Notepad++ load 2 times needed room. I compared the loading way (sync) between Notepad++ & SciTE, I don't see much difference. However, I see SciTE uses currently the async way which is not used by Notepad++. It could be the reason - to verify.

donho:
@xomx 
All done from:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11047#issuecomment-1014858726

Except the 2 following points:

> in FileManager::loadFileData there is the following filesize (lenFile) truncation):
const char *newData = wmc.encode(fileFormat._encoding, SC_CP_UTF8, data, static_cast<int32_t>(lenFile), &newDataLen, &incompleteMultibyteChar);

The `lenFile` argument goes down until to pass to function `MultiByteToWideChar` which takes `int` as type of argument.
OTOH though `lenFile` is type of size_t, the value is around blockSize (`static const int blockSize = 128 * 1024 + 4;`) so it should be OK.

> case WM_FRSAVE_INT and WM_FRSAVE_STR ... remove static_cast<long> from lParam

The constructor prototype is:
`recordedMacroStep(int iMessage, uptr_t wParam, uptr_t lParam, const TCHAR *sParam, int type)`
In which the iMessage will never need to be 64bits, so it's OK.

xomx:
Further problematic parts I met during debugging:

- ColumnEditorDlg::run_dlgProc -> case WM_COMMAND :
  lines 173 & 290:
  (*_ppEditView)->replaceTarget(s2r.c_str(), `int`(lineBegin), `int`(lineEnd));
- DocumentMap::scrollMapWith
  higherY = _pMapView->execute(SCI_POINTYFROMPOSITION, 0, `static_cast<int32_t>`(mapPos._higherPos));
- FunctionCallTip::showCalltip
  _currentOverload = `static_cast<int32_t>`(i);
- FunctionParser::getInvertZones
  if (`int`(begin) < sourceZones[0].first)
- FunctionListPanel::getBodyClosePos
  targetEnd = `static_cast<int32_t>`(begin);
- Notepad_plus::filePrint
  printer.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), _pEditView, showDialog, `int`(startPos), `int`(endPos), _nativeLangSpeaker.isRTL());
  and declaration of Printer::init:
  void Printer::init(HINSTANCE hInst, HWND hwnd, ScintillaEditView *pSEView, bool showDialog, `int` startPos, `int` endPos, bool isRTL)

And I have some progress in debugging of the main problem. N++ has no problem up to a 2GB file, but when I tried to open exactly (or a little bit bigger) 2GB file, it silently exited. Filesize (2147483648 - 1) was ok, but 2147483648 (exactly 2GB) was not ok. First problem appeared after exact 2GB filedata loading in Notepad_plus::addHotSpot at this place:
`pView->getText(encodedText, startPos, endPos);`
The underlying Scintilla throwed write access violation exception in .\scintilla\src\Editor.cxx  -> `Editor::WndProc` -> `case SCI_GETTEXTRANGE`.

The reason for this behavior lies within `Sci_TextRange` used in .\PowerEditor\src\ScintillaComponent\ScintillaEditView.cpp here:
```
void ScintillaEditView::getText(char *dest, size_t start, size_t end) const
{
	Sci_TextRange tr;
	tr.chrg.cpMin = static_cast<Sci_PositionCR>(start);
	tr.chrg.cpMax = static_cast<Sci_PositionCR>(end);
	tr.lpstrText = dest;
	execute(SCI_GETTEXTRANGE, 0, reinterpret_cast<LPARAM>(&tr));
}
```
`tr.chrg.cpMin` & `tr.chrg.cpMax` are still defined in the .\scintilla\include\Sci_Position.h as `long`!
But unlike Unix, MS Windows x64 platform is LLP64, so the `long` size is still 32-bit there, so any 2GB+ number causes `cpMin`/`cpMax` overflow (it becomes a negative number).

chcg:
From Sci_Position.h

> // For Sci_CharacterRange  which is defined as long to be compatible with Win32 CHARRANGE
> typedef long Sci_PositionCR;

See also:
https://docs.microsoft.com/en-us/windows/win32/api/richedit/ns-richedit-charrange

It seems to be intentional since scintilla 3.6, as it is stated

> External interfaces use the Sci_Position and Sci_PositionU typedefs instead of int and unsigned int to allow for changes to a 64-bit interface on 64-bit platforms in the future. Applications and external lexers should start using the new type names so that they will be compatible when the 64-bit change occurs. There is also Sci_PositionCR (long) for use in the Sci_CharacterRange struct which will also eventually become 64-bit.

xomx:
@chcg 
Yes, but the question is - how the SciTe handles that, because it works ok with 2GB+ files (and N++ not).
Obviously we cannot use the current N++ code like
```
	tr.chrg.cpMin = static_cast<Sci_PositionCR>(start);
	tr.chrg.cpMax = static_cast<Sci_PositionCR>(end);
```
as it overflows the signed `long` 32-bit integer with 2GB+ integers.

donho:
@xomx 
If you change `typedef long Sci_PositionCR;` to `typedef INT_PTR Sci_PositionCR;`, what will happen?

donho:
@xomx 
If I do understand about the ref provided by @chcg, it's on the application level (so here Notepad++ in our case) to adapt the definition of Sci_PositionCR for our need. Therefore we should modify `typedef long Sci_PositionCR;` to `typedef INT_PTR Sci_PositionCR;`. 

@chcg Are we on the same page?

xomx:
@donho 


> If you change `typedef long Sci_PositionCR;` to `typedef INT_PTR Sci_PositionCR;`, in "notepad-plus-plus\scintilla\include\Sci_Position.h", what will happen?

I tried that yesterday, but it caused a lot of build errors&warnings, so I did not follow further.

xomx:
So finally I solved this debugging challenge, I have found the main source of the 2GB+ files problem.

tldr - just adjust this N++ code:
```
BufferID FileManager::loadFile(const TCHAR * filename, Document doc, int encoding, const TCHAR *backupFileName, FILETIME fileNameTimestamp)
{
	bool ownDoc = false;
	if (!doc)
	{
		doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT);
		ownDoc = true;
	}
```

to this:
```
BufferID FileManager::loadFile(const TCHAR * filename, Document doc, int encoding, const TCHAR *backupFileName, FILETIME fileNameTimestamp)
{
	bool ownDoc = false;
	if (!doc)
	{
		doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT, 0, SC_DOCUMENTOPTION_TEXT_LARGE);
		ownDoc = true;
	}
```

Longer story:
- to be able to get at least over the problem with `Sci_PositionCR`, I redefined it again from the original `long` to the `INT_PTR` and adjusted manually all the N++ source code for the resulting errors
- this helped, but then I got only to the same problematic place as before (".\scintilla\src\Partitioning.h" -> `PartitionFromPosition(T pos)`)
- the error origin was in the `body->ValueAt(middle)`, which returned negative (2GB-4GB range) or nonsense (4GB-6GB) etc. integers for the 2GB+ files
- MSVS showed me the actual returning type: `type const int &` for the `body->ValueAt()`, so I knew what causes the problem (it should return `__int64` type instead), but for a long time I could not track this to a specific N++ source code or setting
- in the end it turned out, that there is a runtime switch in ".scintilla\src\CellBuffer.cxx":
```
  CellBuffer::CellBuffer(bool hasStyles_, bool largeDocument_) :
	hasStyles(hasStyles_), largeDocument(largeDocument_) {
	readOnly = false;
	utf8Substance = false;
	utf8LineEnds = 0;
	collectingUndo = true;
	if (largeDocument)
		plv = std::make_unique<LineVector<Sci::Position>>();
	else
		plv = std::make_unique<LineVector<int>>();
  }
```
- then it was easy to find info abnout `SC_DOCUMENTOPTION_TEXT_LARGE` in the [Scintilla Doc](https://www.scintilla.org/ScintillaDoc.html)


This PR is far from finished with this finding, but now the N++ is stable and works with 2GB+ files:
npp-bigfile-linepos-fixed

donho:
@xomx 
Terrific! It can not only open a 2GB+ files, but also not allocate double of file size, thanks to you!
 I just did a new commit, which opens 200MB+ files without syntax highlighting. For me a 200 MG code source is not reasonable, if it's not impossible. 
I will also add new commit regarding your remarks in your previous comment. Thank you very much for your help.

xomx:
@donho 

I like that use of SC_DOCUMENTOPTION_STYLES_NONE for 200MB+ files. I tested your last commit with 200MB+ files and the memory footprint was as you described.

We talked before about switching off the 'Word wrap' for large files by default. What do you think about the possible backups and session snapshots options? Let's say for files larger than 2GB.

I still do not like much the patch of `Sci_PositionCR` to `INT_PTR` in the Scintilla "Sci_Position.h", but it is working and the changes needed for the N++ source code are not as big as I originally expected. Moreover the Scintilla author is eventually going to change the `long` definition to something 64-bit on the Windows platform, thus this patch should be ok.

PowerEditor/src/Notepad_plus.h
```diff
@@ -360,8 +360,8 @@ friend class FileManager;
 	//Synchronized Scolling
 	struct SyncInfo final
 	{
-		int _line = 0;
-		int _column = 0;
+		INT_PTR _line = 0;
```
mere-human, line 363: Consider using fixed-size integers from <cstdint> header.
https://en.cppreference.com/w/cpp/types/integer
For example: `int64_t`, `uint64_t`. Or use some typedef like `using line_t = int64_t`
This will make it more obvious and it will be less dependent on Win32 headers.

donho:
@xomx 
> I still do not like much the patch of Sci_PositionCR to INT_PTR in the Scintilla "Sci_Position.h", but it is working and the changes needed for the N++ source code are not as big as I originally expected.

Yes. I've done this part  Sci_PositionCR to long long instead of INT_PTR  - small changes really.

> it is better to use some predefined constant for that like NPP_STYLING_FILESIZE_LIMIT instead of 'magic numbers'

Done. 

> or maybe make this a configurable setting

I was thought about it. But I prefer to have users' feedback firstly.

> We talked before about switching off the 'Word wrap' for large files by default. What do you think about the possible backups and session snapshots options? Let's say for files larger than 2GB.

I will test this part tomorrow - if the impact exists, I'll do it definitely.

PowerEditor/src/Notepad_plus.h
donho: @mere-human 
I don't see your point.

mere-human: I suggested using fixed size integer types from standard C++ headers instead of `INT_PTR` from Windows header.
This could add more clarity and reduce ambiguity.
Also there's `intptr_t` .
What was the idea behind using `INT_PTR`?

donho: There's no idea behind INT_PTR, just the type I found which is signed "size_t" but indeed, intptr_t looks better.

donho:
@xomx 
> [Further problematic parts I met during debugging:](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11047#issuecomment-1015939419)

Done. Thank you.

xomx:
Where previously 32-bit only, but now potentially 64-bit, integers are handled by `GetDlgItemInt`/`SetDlgItemInt`, such places should be re-coded by using `GetDlgItemText`/`SetDlgItemText`. Now I know only about the `GoToLineDlg`.

".\PowerEditor\src\ScintillaComponent\GoToLineDlg.h":

`int` getLine() const {
	BOOL isSuccessful;
	`int line = ::GetDlgItemInt(_hSelf, ID_GOLINE_EDIT, &isSuccessful, FALSE);`
	return (isSuccessful?line:-1);
};

".\PowerEditor\src\ScintillaComponent\GoToLineDlg.cpp":

line 91:  `int` line = getLine();
line 188: `::SetDlgItemInt(_hSelf, ID_CURRLINE, (UINT)current, FALSE);`
line 189: `::SetDlgItemInt(_hSelf, ID_LASTLINE, (UINT)limit, FALSE);`

===

A note for the use of `SC_DOCUMENTOPTION_TEXT_LARGE`:

There is another place where the `SCI_CREATEDOCUMENT` is used - at `FileManager::newEmptyDocument()`.
I think that for surety we should change the Scintilla `Document` creation there in 64-bit builds to:
`Document doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT, 0, SC_DOCUMENTOPTION_TEXT_LARGE);`
(when an user creates a new tab (Ctrl+N) and then paste a 2GB+ data in there, the previously fixed erratic behavior could be back...)

For the similar reason I suggest adjusting the `FileManager::loadFile` place to:
`doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT, 0, fileSize < NPP_STYLING_FILESIZE_LIMIT ? SC_DOCUMENTOPTION_TEXT_LARGE : SC_DOCUMENTOPTION_STYLES_NONE | SC_DOCUMENTOPTION_TEXT_LARGE);`

Otherwise we have to monitor the current N++ filebuffer size for the 2GB limit crossing (then e.g. we ensures reloading of the buffer with the large Scintilla `Document` variant).

===

And to save time with otherwise useless testing - do anyone know, what is the N++ behavior for situations, where another memory allocation is no longer possible? I mean e.g. when I am writing or appending another data to a N++ filebuffer and then Scintilla fails with `SC_STATUS_BADALLOC`...
(32-bit SciTe just showed me during another data pasting a messagebox with "Memory exhausted. SciTE will now close.")

PowerEditor/src/Notepad_plus.h
mere-human: By the way, `intptr_t` is 4 bytes in x32 app and 8 bytes in x64 app. Is it expected?
What is the expected behavior in x32 version of the app for the fixed bug?

mere-human:
I decided to test this PR.
So I generated a 2GB text file using these commands:
```
fsutil file createnew big.txt 2147483648
echo "This is just a sample line appended to create a big file.. " >> big.txt
```
While opening the file, I no longer see the "File is too big to be opened by Notepad++" message box.
However, in debug build I see an assert followed by a crash:
```
Assertion [useCount == 0] failed at e:\development\notepad-plus-plus\scintilla\src\positioncache.cxx 372
notepad++.exe has triggered a breakpoint.
```

Also, I compared the amount of occupied RAM on master vs this PR for a relatively big file (a valid Lua script, 60MB).
And it's about the same - ~180MB (debug build). The same for x32 build.

xomx:
@mere-human 
> However, in debug build I see an assert followed by a crash:

This seems to me as a Scintilla limitation, not a N++ bug (BTW - SciTe446.exe has the same problem with your big.txt).
The big.txt is full of NUL-bytes with no CR/LF, so in fact one huge line with 2GB+ NUL-chars. The problem is probably here:
"\scintilla\src\EditView.cxx" -> `EditView::LocationFromPosition`:
const `int` posInLine = `static_cast<int>`(pos.Position() - posLineStart);

So there will be a 32-bit integer overflow. Maybe somewhere in the Scintilla doc will be a note about a maximum line length(?)


> Also, I compared the amount of occupied RAM on master vs this PR for a relatively big file (a valid Lua script, 60MB).
> And it's about the same - ~180MB (debug build). The same for x32 build.

Yes, it should be that. Only for the files over 200MB you will see the half of the memory footprint.

donho:
> Yes, it should be that. Only for the files over 200MB you will see the half of the memory footprint.

I  would reword it to:
Yes, it should be that. Only for the files over 200MB you won't see the allocation of double of file size in memory.

mere-human:
> This seems to me as a Scintilla limitation, not a N++ bug (BTW - SciTe446.exe has the same problem with your big.txt).

Yes, I've tried opening the same file with SciTE 5 and it crashed as well. Maybe, I need to prepare a different big file...

donho:
@xomx 
> Where previously 32-bit only, but now potentially 64-bit, integers are handled by GetDlgItemInt/SetDlgItemInt, such places should be re-coded by using GetDlgItemText/SetDlgItemText. Now I know only about the GoToLineDlg.
...

Done in the last commit. Nice spotted!

> A note for the use of SC_DOCUMENTOPTION_TEXT_LARGE:
There is another place where the SCI_CREATEDOCUMENT is used - at FileManager::newEmptyDocument().
I think that for surety we should change the Scintilla Document creation there in 64-bit builds to:
Document doc = (Document)_pscratchTilla->execute(SCI_CREATEDOCUMENT, 0, SC_DOCUMENTOPTION_TEXT_LARGE);
(when an user creates a new tab (Ctrl+N) and then paste a 2GB+ data in there, the previously fixed erratic behavior could be back...)

It could happen very rarely. 
Though I see no harmful to use SC_DOCUMENTOPTION_TEXT_LARGE, I prefer keep the old behaviour so far.

donho:
Tested with an old laptop of 6-7 years (the 1st generation of Surface book), here are the results:
1. 2GB files take at least 1 minute, 4GB files take almost 3 minutes to load.
2. When Word Wrap is ON, even on the old laptop, I can tell the difference barely.

According the tested result, the behaviour of x64 binary to open huge file will be the following (IMO):
1. No "allowToOpenHugeFile.xml", either no GUI option to allow open huge file, but a dialog will always appear while users try to open any 2GB+ file. (SciTE displays always the dialog.)
2. No disabling Word Wrap - it could be done afterward if needed.

xomx:
@donho 

> It could happen very rarely.
> Though I see no harmful to use SC_DOCUMENTOPTION_TEXT_LARGE, I prefer keep the old behaviour so far.

Leaving intentionally hidden bugs in the code?

> [Restore Scintilla header](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11047/commits/ef9c4522844448c7f73b9bd801b24849a6e7171e) 

The x64 release binary, compiled with PlatformToolset v141, is not crashing for me on startup or later.

But now, with the reverted Sci_Position.h header, it is now consistently crashing (just open a text file with the exact 2GB size and scroll down to the end of that file...). The somewhat specific crash is only a visible signal, that any N++ functionality, which depends on the Sci_PositionCR somehow (e.g. ScintillaEditView::getText), will have problems with 2GB+ files.

donho:
@xomx
> Leaving intentionally hidden bugs in the code?

How can it be "hidden bugs"? It's good to think of the case of copying 2GB+ of code/text and paste it. But for me it's useless, and cannot happen. If it can, then let's users talk, we will see what we'll do about it. 
My job is keep the incremental change, so any regression can be tracked and fixed easily.

> But now, with the reverted Sci_Position.h header, it is now consistently crashing (just open a text file with the exact 2GB size and scroll down to the end of that file...). The somewhat specific crash is only a visible signal, that any N++ functionality, which depends on the Sci_PositionCR somehow (e.g. ScintillaEditView::getText), will have problems with 2GB+ files.

I cannot reproduce the crash. And I don't see why reverse Scintilla header could make it more unstable.
I'll do a RC soon so the community can test it.

PowerEditor/src/Notepad_plus.h
donho: Yes, it's expected.
In this PR, the file length can be beyond 4bytes for x64 binary so the position of cursor can be beyond 4bytes, therefore we need 8 bytes to hold this value. 32bits binary keeps the same behaviour and nothing to be fixed.

donho:
FYI:
https://community.notepad-plus-plus.org/topic/22438/notepad-v8-2-2-release-candidate

ArkadiuszMichalski:
>@xomx
>But now, with the reverted Sci_Position.h header, it is now consistently crashing (just open a text file with the exact 2GB size and scroll down to the end of that file...). 

I have file with `2,00 GB (bytes: 2 147 483 648)` and I have no crash. What exactly is the file causing this?

xomx:
@ArkadiuszMichalski 

> What exactly is the file causing this?

dl 7-Ziped ~430MB [here](https://drive.google.com/file/d/11r2sdOFUzSino7lrq6QveBhensqTFcCf/view?usp=sharing)

1. Run a 8.2.2RC x64
2. Open Gutenberg_#10802_combined_txt_files-cut-2GBexact.txt
3. Scroll down to the end of file

The culprit seems to be the reverted fix for the Sci_Position.h.
BUT - I could consistently reproduced the crash, now suddenly I cannot. Strange.
However from the debug crash pic below is evident, how the missing patch for the `Sci_PositionCR` influences the code (e.g. look at the tr->chrg.cpMax overflowed-int value there)

npp.8.2.2.portable.x64:
2GB-crash-8 2 2RC

debug version of the same notepad-plus-plus-master snapshot:
2GB-crash-debug

ArkadiuszMichalski:
Here's a crash:
https://community.notepad-plus-plus.org/topic/22438/notepad-v8-2-2-release-candidate?_=1643228223595

donho:
@xomx 
> dl 7-Ziped ~430MB here

I cannot reproduce the crash on [RC2](https://community.notepad-plus-plus.org/topic/22446/notepad-v8-2-2-release-candidate-2?_=1643305164610) with the file you provided.

> So we need back the fix in the Sci_Position.h. 

I rebuilt with the modif `typedef intptr_t Sci_PositionCR;`, here's the 64bis binary (built by the release machine):
http://download.notepad-plus-plus.org/repository/8.x/Sci_Position_header_modified/notepad++.zip

This release binary crashes on startup from a scratch environment (no session, basic conf file). However, the debug binary doesn't crash, and can load 2GB file.

Since the 64 release binary built by the release machine crashes, I cannot use the above code for our release. If you have any good solution, please let me know.

> Otherwise it is actually better, if the code crashes immediately, as this starts to overwrite unintentional memory (e.g. on the pic above it does: tr->lpstrText[-4294967292] = '\0', whereas the len should be a small positive number instead...). And that is only one specific case.

Sorry, I don't follow you, could you elaborate it?

donho:
@xomx 
Here is crashed release binary launch in debug mode:
image

xomx:
@donho 


> Sorry, I don't follow you, could you elaborate it?

32-bit Sci_PositionCR problem:

After opening a 2GB+ file, everytime the actual N++ fileview reaches the 2GB+ position, in the N++ is called:

`void Notepad_plus::addHotSpot(ScintillaEditView* view)`

and there is this code:

```
char *encodedText = new char[endPos - startPos + 1];
pView->getText(encodedText, startPos, endPos);
```

- endPos & startPos there are of the correct size_t and these enclose the actual visible fileview space in the N++ (`endPos – startPos + 1` is typically something small like 2000)
- `encodedText` is there a ptr to a small allocated heap memory buffer

Now look at the called `ScintillaEditView::getText`:
```
void ScintillaEditView::getText(char *dest, size_t start, size_t end) const
{
	Sci_TextRange tr;
	tr.chrg.cpMin = static_cast<Sci_PositionCR>(start);
	tr.chrg.cpMax = static_cast<Sci_PositionCR>(end);
	tr.lpstrText = dest;
	execute(SCI_GETTEXTRANGE, 0, reinterpret_cast<LPARAM>(&tr));
}
```
- as soon as the `start` or `end` is at least 2147483648, the `static_cast<Sci_PositionCR>` causes 32-bit int buffer overflow ... e.g. the 2147483648 (0x80000000) becomes -2147483648

and then we go deep in the Scintilla code (".\scintilla\src\Editor.cxx"):
```
case SCI_GETTEXTRANGE: {
			if (lParam == 0)
				return 0;
			Sci_TextRange *tr = static_cast<Sci_TextRange *>(PtrFromSPtr(lParam));
			Sci::Position cpMax = static_cast<Sci::Position>(tr->chrg.cpMax);
			if (cpMax == -1)
				cpMax = pdoc->Length();
			PLATFORM_ASSERT(cpMax <= pdoc->Length());
			Sci::Position len = cpMax - tr->chrg.cpMin; 	// No -1 as cpMin and cpMax are referring to inter character positions
			pdoc->GetCharRange(tr->lpstrText, tr->chrg.cpMin, len);
			// Spec says copied text is terminated with a NUL
			tr->lpstrText[len] = '\0';
			return len; 	// Not including NUL
		}
```
- but there `tr->chrg.cpMax/cpMin` is now an invalid negative number
- not even the code is doing nonsense now, it is now overwriting random N++ heap memory at:
  `tr->lpstrText[len] = '\0';`
  (`len` there should be some small positive integer, but it can becomes e.g. a big negative one)
- seeing the access violation crash here depends on the current memory layout only, if there is just now the write access allowed e.g. at tr->lpstrText[-421359878], it succeeds (and corrupts that memory!!!)

xomx:
@donho 


> I rebuilt with the modif `typedef intptr_t Sci_PositionCR;`, here's the 64bis binary (built by the release machine): http://download.notepad-plus-plus.org/repository/8.x/Sci_Position_header_modified/notepad++.zip
> 
> This release binary crashes on startup from a scratch environment (no session, basic conf file).

Not for me, I can open it in an otherwise empty folder and load a 2GB+ files without problem.
Strange.



> Since the 64 release binary built by the release machine crashes, I cannot use the above code for our release. If you have any good solution, please let me know.

Ok, I will think about it.

donho:
@xomx 
The culprit of crash is DSpellCheck plugin (FYI @Predelnik).
I will add it back `typedef intptr_t Sci_PositionCR;`
Thank you for insistence :)

ArkadiuszMichalski:
@donho 
It's intentional that when we do reload we use more RAM as closing and reopening the big file? When open 2GB file I see RAM pick to 4,2GB and drop to 2,3GB. Now when I make reload this file I get RAM pick to 6,4GB and drop to 2,3GB.

xomx:
@donho 

> The culprit of crash is DSpellCheck plugin

Yes, with that plugin included I can reproduce the problem both in Release and Debug builds.
In Debug it first throws this assert:
"Assertion [cpMax <= pdoc->Length()] failed at .\notepad-plus-plus-master\scintilla\src\editor.cxx 6170"
(culprit - this plugin calls the `SCI_GETTEXTRANGE` with now somewhat mangled `Sci_TextRange`, where `tr->chrg.cpMin = 0x00000000ffffffff` and `tr->chrg.cpMax = 0x00000187a3e3f9a0` instead of `0` and `0`)

? Maybe a recommendation note for plugins authors to update their Sci_Position.h header, if they are using the Sci_TextRange, Sci_CharacterRange or Sci_PositionCR somehow in their plugin ...
  
NPP_STYLING_FILESIZE_LIMIT
- maybe the 200MB is too low, as I saw also this usage: [Opening 620 MB SQL file leaves it very slow
](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8802)

xomx:
@ArkadiuszMichalski 

Are you talking about these?
memory_reload_spikes

If so, I think that it is ok. Important is, that the memory usage is not increasing (a leak....).

ArkadiuszMichalski:
@xomx Yes, because when I reload 4GB file it need 13GB in pick, when open it need less, only 8.2GB. This means that file handling requires even more free RAM than 2x the file.

donho:
@ArkadiuszMichalski 
> It's intentional that when we do reload we use more RAM as closing and reopening the big file? When open 2GB file I see RAM pick to 4,2GB and drop to 2,3GB. Now when I make reload this file I get RAM pick to 6,4GB and drop to 2,3GB. It will be more noticeable for larger files.

After checking the code of Notepad++ reload routine, I didn't find the reason. I believe it's the Scintilla internal allocation. Anyway, the allocated memory has been de-allocated, that's the most important thing :)

donho:
@xomx 
> maybe the 200MB is too low, as I saw also this usage: Opening 620 MB SQL file leaves it very slow

Do you mean to increase it to 600 MB?
If yes, I am not for it:
I have tested 260MB XML files today, it made Notepad++ hanging on my gaming machine while loading it. Such unstable behaviour leads to [the improvement](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/774321e0997e9561482124cd66007cee3c224758). 

I think it's important to make this large file limitation as 200MB.

PR #11050, opened 2022-01-17
Fix file dialog checkbox doesn't work in empty dir

### LINKED ISSUE #10436: Save Dialog "Append Extension" doesn't work if Default Directory is empty

1. Preferences > Default Directory
2. Select 3rd radio button, specify a path to an empty directory without any files
3. File > New
4. File > Save As
5. Check "Append Extension" if it is OFF
6. Change name to "test" (no extension)
5. Press "Save"

### Expected Behavior

File is saved as "test.txt".

### Actual Behavior

File is saved as "test" without an extension.

### Debug Information

Notepad++ v8.1.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1165
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll HexEditor.dll JSMinNPP.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppEditorConfig.dll NppExport.dll NppMarkdownPanel.dll NppMenuSearch.dll NppTags.dll PythonScript.dll XMLTools.dll

mere-human:
Use hooks for keyboard and window procedure instead of subclassing controls.
Use a handle map for transferring instance data to the hook procedure.
This approach should be more reliable than the previous one.

Fix #10436

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -22,6 +22,7 @@
 #endif
 #include <comdef.h>		// _com_error
 #include <comip.h>		// _com_ptr_t
+#include <unordered_map>
 #include "CustomFileDialog.h"
 #include "Parameters.h"
 
@@ -171,13 +172,16 @@ namespace // anonymous
 		return {};
 	}
 
-	LRESULT callWindowClassProc(const TCHAR* className, HWND hwnd, UINT msg, WPARAM wparam, LPARAM lparam)
+	HWND getDialogHandle(IFileDialog* dialog)
 	{
-		WNDCLASSEX wndclass = {};
-		wndclass.cbSize = sizeof(wndclass);
-		if (GetClassInfoEx(nullptr, className, &wndclass) && wndclass.lpfnWndProc)
-			return CallWindowProc(wndclass.lpfnWndProc, hwnd, msg, wparam, lparam);
-		return FALSE;
+		com_ptr<IOleWindow> pOleWnd = dialog;
+		if (pOleWnd)
+		{
+			HWND hwnd = nullptr;
+			if (SUCCEEDED(pOleWnd->GetWindow(&hwnd)))
+				return hwnd;
+		}
+		return nullptr;
 	}
 
 	// Backs up the current directory in constructor and restores it in destructor.
@@ -251,21 +255,22 @@ public:
 	}
 	IFACEMETHODIMP OnFolderChange(IFileDialog*) override
 	{
-		// First launch order: 3. Custom controls are added but inactive.
+		// Dialog startup calling order: 3. Custom controls are added but inactive.
+		if (!foundControls())
+			findControls();
 		return S_OK;
 	}
 	IFACEMETHODIMP OnFolderChanging(IFileDialog*, IShellItem* psi) override
 	{
 		// Called when the current dialog folder is about to change.
-		// First launch order: 2. Buttons are added, correct window title.
+		// Dialog startup calling order: 2. Buttons are added, correct window title.
 		_lastUsedFolder = getFilename(psi);
 		return S_OK;
 	}
 	IFACEMETHODIMP OnSelectionChange(IFileDialog*) override
 	{
-		// First launch order: 4. Main window is shown.
-		if (shouldInitControls())
-			initControls();
+		// This event isn't triggered in an empty folder.
+		// Dialog startup calling order: 4. Main window is shown.
 		return S_OK;
 	}
 	IFACEMETHODIMP OnShareViolation(IFileDialog*, IShellItem*, FDE_SHAREVIOLATION_RESPONSE*) override
@@ -274,7 +279,7 @@ public:
 	}
 	IFACEMETHODIMP OnTypeChange(IFileDialog*) override
 	{
-		// First launch order: 1. Inactive, window title might be wrong.
+		// Dialog startup calling order: 1. Inactive, window title might be wrong.
 		UINT dialogIndex = 0;
 		if (SUCCEEDED(_dialog->GetFileTypeIndex(&dialogIndex)))
 		{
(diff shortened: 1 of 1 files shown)


mere-human:
## How to test

Environment: Vista, 7, 10, 11, etc. (I tested on Windows 10 & Windows 7 VM)

I tested all of the following. But you can pick any subset of it. Because not all the scenarios are actually affected.

**Generic test cases:**
* File dialogs:
  * File > Open, Save, Save as
  * Search > Find in files, enter something in “Directory”, press [...] (should open directory)
  * File > Open Folder as Workspace (should open directory)
  * Run > Run.. (exe file type)
  * etc
* Dialog confirmation:
    * Enter (Return) key
    * Press the button with a mouse
    * Select the button with tab and hit Enter/Space key
    * Press Alt + S shortcut (Save) or Alt + O (Open)
    * Buttons can be swapped if both Windows and Notepad++ language is RTL
* File type
    * Language > Pick any (should be reflected in save dialog)
    * File > Save as > Select file type, enter name, save (should append extension)
    * File > Save as > Select file type (C, C++), enter a name with a different extension “tmp.txt” (extension should be preserved)
    * File > Save as > Change type then change name
    * Remove extension and ensure it’s added after confirmation
* Preferences
    * Settings > Default directory (should affect initial directory in file dialog)
        * Set to an empty directory and ensure the checkbox works
    * Settings > MISC > Session file ext (pick any)
        * View > Folder as Workspace (check)
        * File > Save Session, check “Save Folder as Workspace”
        * File type should be as set in preferences or .
        * Modify panel, open saved file
    * Settings > MISC >Workspace file ext (pick any)
        * View > Project Panel > Panel 1, add some entries to Project
        * Project Panel > Workspace > Save as
        * File type should be as set in preferences or .
        * Modify panel, open saved file
* RTL
    * Windows EN, N++ AR (RTL) - Save dialog EN, main window AR (RTL)
    * Windows AR, N++ EN - Save dialog EN, main window EN
    * Windows AR (RTL), N++ AR - Save dialog AR (RTL), main window AR (RTL)
* Other
    * Rename in a dialog
        * Select a file or a folder when a file dialog is open
        * Press F2 to rename it
        * Pick a different name and press Enter
        * The file dialog shouldn’t be closed
    * Multiple dialogs from a macro
        * Modify a document
        * Macro > Run a macro
        * Select “Trim Trailing Spaces..”
        * Click “Run” several times or specify some number > 1
        * Several dialogs should be open

**Path cases:**
* File name
    * 1.txt
* Folder name - should navigate to it
    * e:\tmp
* Relative path
    * ..\tmp\test-files\keys.c
* Absolute path
    * e:\tmp\test-files\keys.c
* Forward slash path
    * e:/tmp/test-files/keys.c
    * ../Share
* Environment variable
    * %appdata%\tmp.txt
* UNC name:
    * \\127.0.0.1\c$
    * "\\127.0.0.1\c$\Program Files (x86)\Notepad++\change.log"
    * "\\127.0.0.1\c$\Python36-32\LICENSE.txt"

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -362,31 +369,71 @@ class FileDialogEventHandler : public IFileDialogEvents, public IFileDialogContr
...
-	// Overrides window procedures for file name edit and ok button.
+	// Find window handles for file name edit and ok button.
 	// Call this as late as possible to ensure all the controls of the dialog are created.
-	void initControls()
+	bool findControls()
 	{
 		assert(_dialog);
-		com_ptr<IOleWindow> pOleWnd = _dialog;
-		if (pOleWnd)
+		if (HWND hwndDlg = getDialogHandle(_dialog))
```
chcg, line 377: assignment in if clause should be typically avoided to find easily the wrong usage of = instead of ==, see https://wiki.sei.cmu.edu/confluence/display/c/EXP45-C.+Do+not+perform+assignments+in+selection+statements

mere-human: Right. However, on the other hand, it reduces the scope of a variable and makes code more compact.
But I agree, let's change it.

Yaron10:
@mere-human,

Great. 👍 
Thank you for your work.

mere-human:
@Yaron10 Did you check the build? Does it fix the problem for you?

Yaron10:
@mere-human,

Yes, I did test it before posting my previous reply.
The issue is fixed. 👍 

Thanks again.

mere-human:
@donho What do you think about this fix?

mere-human:
> OTOH, I did not have confidence in this PR, due to the change of mechanism from the Inheritance to the hooking - such change cause the regression, almost inevitably.

If there would be a regression, then I will fix it.

> I'm wondering why the fix cannot be done by keeping the same Inheritance mechanism whereas it works almost perfectly so far?

You answered your own question - "almost" perfectly. 
There were numerous problems with the sub-classing approach.
For example:
#9510
#9657
#10290
It seems, we hit the limit of this approach, while there are still problems that are not fixed like this one.
I believe the suggested approach in this PR is superior:
- no need to rely on obscure window procedure messages from controls like WM_GETDLGCODE
- no need to rely on the controls init order and do the sub-classing in OnSelectionChange()
- the code now should be simpler and more robust

mere-human:
We can include it in the next release, so it's separate from 2GB file changes.

donho:
@mere-human 
> If there would be a regression, then I will fix it.

Thank you for your engagement. But it's better to not have the regression, isn't it? :)

Well, what I'm worrying about, is the mechanism of CustomFileDialog changes again for fixing the other bugs/regressions.
Are you sure the hooking is the final solution?

mere-human:
> @mere-human
> 
> > If there would be a regression, then I will fix it.
> 
> Thank you for your engagement. But it's better to not have the regression, isn't it? :)
> 

@donho Sure, it's better without a regression.
But this fix is addressing another regression as well.

> Well, what I'm worrying about, is the mechanism of CustomFileDialog changes again for fixing the other bugs/regressions. Are you sure the hooking is the final solution?
> 
I think yes. Well, I hope so.

PR #10996, opened 2022-01-05
Update danish.xml to 8.2

Regmos:

PR #11057, opened 2022-01-19
Update danish.xml to 8.2.1

Regmos:


donho:
@Regmos
By checking the hash of 4 commits, it's the same PR comparing with https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10996 
Please create a new PR with a new commit - not the old ones since I couldn't rebase them.

Issue #10640, opened 2021-10-09
Shell doesn't register on ARM based systems


FlashD:
### Description of the Issue
The provided shell file NppShell_06.dll in the ARM package fails to register with Windows on ARM. The end result is the same using both the installer or registering it by hand.

### Steps to Reproduce the Issue

1. Install Notepad++ using the installer.
2. Try to open any file using the context menu
2. Notice missing "Edit with Notepad++" in context menu
3. Open CMD with elevated privileges
4. Navigate to the folder Notepad++ is installed at
5. Try to register the NppShell_6.dll with "regsvr32 /i NppShell_6.dll"

### Actual Behavior

<img width="671" alt="Error2" src="https://user-images.githubusercontent.com/5943012/136655522-4d1ed230-f704-4139-9b61-309313df9e2f.png">

### Debug Information

Notepad++ v8.1.5   (ARM 64-bit)
Build time : Sep 26 2021 - 15:10:12
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (Unknown-bit) 
OS Version : 2009
OS Build : 19043.1237
Current ANSI codepage : 1252
Plugins : HexEditor.dll

donho:
could you provide me your `NppShell_06.dll` so I can make sure it's arm64 architecture ?

jof284:
I attempted to install v8.4.4 and the installer attempts to call multiple instances of Microsoft Register Server (regsvr32) eventually locking up the machine. I tested this with Windows 11 Pro (22000.832) on a Lenovo X13s (Snapdragon (TM) 8cx Gen 3, 16 GB RAM).

PR #11069, opened 2022-01-22
Arm64 bitness

### LINKED ISSUE #11068: arm64 version string
### Description of the Issue
From code inspection arm64 is missing in the platform information and therefore also the version string doesn't contain 64bit attribute.

See https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info

Check debug info text

### Expected Behavior

### Actual Behavior
OS Name : Windows 10 Pro (Unknown-bit)

### Debug Information
N/A

chcg:
Fix #11068


PowerEditor/src/Parameters.cpp
```diff
@@ -791,6 +791,10 @@ winVer NppParameters::getWindowsVersion()
 		_platForm = PF_X86;
 		break;
 
+	case PROCESSOR_ARCHITECTURE_ARM64:
```
donho, line 794: It seems the symbol `PROCESSOR_ARCHITECTURE_ARM64` is not recognized by MinGW32 :(

ArkadiuszMichalski: Only [new version](https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/winnt.h#l4877) has it. Maybe add it to [gcc-fixes.h](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/gcc/gcc-fixes.h)?

chcg: Corrected.

donho:
@chcg 
I have no ARM device for testing, but it seems `getWinVerBitStr()` returns the right info (ie. 64 bits) currently on DebugInfo box under ARM system: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10640

chcg:
@donho See the issue you mentioned:
OS Name : Windows 10 Pro (Unknown-bit)

Issue #11081, opened 2022-01-26
User Mode Shell Context Menu


elsonwei:
Allow the shell extension handler be installed per user.
This make the users who don't have the administrator permission can use the shell context menu.
Or the context menu can be used in portable version.

victorel-petrovich:
> Or the context menu can be used in portable version.

Just trying to understand, did you mean you want users of portable version to use context menu as well, or that they can already use it in portable version?

PR #11082, opened 2022-01-26
Support user mode shell context menu

elsonwei:
Run the command "regsvr32 /n /i:user NppShell.dll" to install the shell extension handler for current user.
And can be uninstall by run "regsvr32 /u /i:user NppShell.dll"

fixing #11081

### DIFF
--- a/PowerEditor/src/tools/NppShell/src/NppShell.cpp
+++ b/PowerEditor/src/tools/NppShell/src/NppShell.cpp
@@ -24,10 +24,10 @@ TCHAR szDefaultMenutext[] = TEXT("Edit with &Notepad++");
 
 #ifdef WIN64
 TCHAR szShellExtensionTitle[] = TEXT("ANotepad++64");
-TCHAR szShellExtensionKey[] = TEXT("*\\shellex\\ContextMenuHandlers\\ANotepad++64");
+TCHAR szShellExtensionKey[] = TEXT("%s*\\shellex\\ContextMenuHandlers\\ANotepad++64");
 #else
 TCHAR szShellExtensionTitle[] = TEXT("ANotepad++");
-TCHAR szShellExtensionKey[] = TEXT("*\\shellex\\ContextMenuHandlers\\ANotepad++");
+TCHAR szShellExtensionKey[] = TEXT("%s*\\shellex\\ContextMenuHandlers\\ANotepad++");
 #endif
 
 #define szHelpTextA "Edits the selected file(s) with Notepad++"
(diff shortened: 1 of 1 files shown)


donho:
@elsonwei 
Doesn't the installer do it already? What else does your PR add ?

dfs-:
@donho I guess this can be used from the settings inside N++. I mean in the "File Association" tab.

PowerEditor/src/tools/NppShell/src/NppShell.cpp
```diff
@@ -200,22 +212,26 @@ BOOL RegisterServer() {
 //---------------------------------------------------------------------------
 // UnregisterServer
 //---------------------------------------------------------------------------
-BOOL UnregisterServer() {
+BOOL UnregisterServer(BOOL bUserMode) {
 	TCHAR szKeyTemp[MAX_PATH + GUID_STRING_SIZE];
 
-	RegDeleteKey(HKEY_CLASSES_ROOT, szShellExtensionKey);
+	HKEY hRootKey = (bUserMode) ? HKEY_CURRENT_USER : HKEY_CLASSES_ROOT;
+	wchar_t* szPrefix = (bUserMode) ? L"SOFTWARE\\Classes\\" : L"";
```
dfs-, line 219: Should use HKEY_LOCAL_MACHINE instead of HKEY_CLASSES_ROOT. This will remove requirement of making `szPrefix` dependant of `bUsermode` but always use the _SOFTWARE\Classes_ subkey.

elsonwei: No, Users who don't have administrative privileges cannot modify the HKLM.

elsonwei:
> @elsonwei Doesn't the installer do it already? What else does your PR add ?
NPP cannot be installed if the user doesn't have the administration privileges.
That is, NPP doesn't support user mode installer like Vs Code or WinMerge.
This PR makes the shell extension handler be installed by the user.

elsonwei:
> @donho I guess this can be used from the settings inside N++. I mean in the "File Association" tab.

elsonwei:
> @donho I guess this can be used from the settings inside N++. I mean in the "File Association" tab.

Speaking of "File Association", this also requires the administration privileges.
User should have the ability to set file association individually.
What I want to do is make NPP can be run without requiring privileges.

PowerEditor/src/tools/NppShell/src/NppShell.cpp
dfs-: Sorry if I might sound rude, but users also don't have "administrative privileges" in HKCR. My point is, that writing to HKCR in it's entirety is a bad idea, since it's an aggrregated/composite node [(see here)](https://docs.microsoft.com/en-us/windows/win32/sysinfo/hkey-classes-root-key?redirectedfrom=MSDN).

The linked article states everything that is needed to endorse my comment, so I won't bother elaborating my request any further.

PR #11088, opened 2022-01-26
Autocomplete icons

### LINKED ISSUE #11087: Autocompletion - function autocompletion spruced up!
### Description of the Issue
Scintilla allows for an icon to be displayed in the autocomplete popup that could indicate the type of suggestion the autocomplete is from.  Notepad++ allows autocomplete from words in the current doc, keywords from the autoComplete/*.xml files, or both.  I have both turn on and it would be nice to distinguish between which autocomplete suggestions are from the current doc versus those coming from the autoComplete/*.xml files.  Fortunately, this is a pretty easy lift.

We just need to include an XPM icon (I used the ./PowerEditor/src/icons/functionList2.ico - converted with a free web-based file converter), [register the icon ID](https://www.scintilla.org/ScintillaDoc.html#SCI_REGISTERIMAGE) and [set the type separator](SCI_AUTOCSETTYPESEPARATOR), then just append the type separator and image ID to each keyword.

#### Duplicates with word + function enabled

The only issue I've seen with my testing is that once a keyword is inserted into a file, it is picked up by the word search and then added again when the keywords are merged.  Use `print` as an example:

`print` is found as a document word and merged into the word list
`print` is found as a keyword and merged into the keyword list as `print?1000` (with the type separator and image ID)
Now, when comparing the keywords list to the word list to merge them, `print` != `print?1001` to they are both added resulting in:

image

The second autocomplete in the picture above, typing a keyword `print` when it is already in the document shows twice in the autocomplete suggestions, once as a document word, once as the keyword (from autoComplete/*.xml).

#### Word only autocomplete

There is no "issue" with this, just that I originally tried to put an icon for word autocomplete as well and it "works", but we get the duplicate issue above in spades.

Every time they plain word `print` is found in the doc, it is compared to the building list which already has `print?5000` added (assuming 5000 for the WORD icon image ID).  `print` != `print?5000` so it's added again.  Then the next instance of `print` is found, compared and again, `print` != `print?5000` already in the list, so it's added again ... and again!

SO .. I gave up on trying to find a way to add an icon for word autocomplete.  Frankly, it seems better this way - simple word complete is just that, but using any form of function complete - the keywords for functions get icons.

### Review

I'll submit a pull request but PLEASE someone review!  I think I made use of the appropriate typing and put the XPM data in the header file rather than a global in the C++ source file.

Cheers.

vinsworldcom:
Fix #11087 

PLEASE someone review the code to make sure I used the appropriate types for strings and put the images XPM data in the right place - as private in the header file.

Tested with work, function and both autocomplete modes in 64-bit 8.2.1.

Cheers.


PowerEditor/src/ScintillaComponent/AutoCompletion.h
```diff
@@ -88,7 +88,65 @@ class AutoCompletion {
 	generic_string _keyWords;
 	size_t _keyWordMaxLen = 0;
 
-	FunctionCallTip _funcCalltip;
+    const char* xpmfn[53] = {
```
mere-human, line 91: FYI: You can declare it as `const char* xpmfn[] =`. This way it will compute the array length from the initializer list.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
```diff
@@ -58,6 +58,8 @@ bool AutoCompletion::showApiComplete()
 		return false;
 
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
+	_pEditView->execute(SCI_AUTOCSETTYPESEPARATOR, WPARAM('?'));
+	_pEditView->execute(SCI_REGISTERIMAGE, 1000, LPARAM(xpmfn));
```
mere-human, line 62: It's better to use a named constant for image ID instead of a magic number.

PowerEditor/src/ScintillaComponent/AutoCompletion.h
vinsworldcom: I originally had that, but got a compiler error:

```
E0070 incomplete type is not allowed
```

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
vinsworldcom: named constant?  Sorry my C++ skills are pretty low, do you mean like:

```
#define FUNC_IMG_ID 1000
```

mere-human: Yes, something like that will do.

mere-human: Try calling intToString() from "Common.h"

vinsworldcom: Thanks @mere-human !  That worked, so I cleaned up a bit and pushed a new commit on this PR.

Cheers.

PowerEditor/src/ScintillaComponent/AutoCompletion.h
mere-human: Okay, I see. 
Maybe it's better to move this declaration to `AutoCompletion.cpp` since it's used only there.
This way the header file won't be cluttered with a huge XPM string.

vinsworldcom: Done!  Thanks again.

mere-human (approved):
Looks good
Screenshot 2022-02-04 210410

PowerEditor/src/ScintillaComponent/AutoCompletion.h
```diff
@@ -21,6 +21,7 @@
 #include "tinyxml.h"
 
 const size_t tagMaxLen = 256;
+const auto FUNC_IMG_ID = 1000;
```
chcg, line 24: Seems this could be moved to the cpp file as just used there. This is likely also the case for tagMaxLen. This has the advantage to be not visible for all who include AutoCompletion.h and within the global namespace.

constexpr for strings is not available before c++20 (https://www.modernescpp.com/index.php/constexpr-vector-and-string-in-c-20)

vinsworldcom: moved the FUNC_IMG_ID to `.cpp` file.  The `tagMaxLen` was not added by this PR so I'm leaving it there for now.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
```diff
@@ -58,6 +118,8 @@ bool AutoCompletion::showApiComplete()
 		return false;
 
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
+	_pEditView->execute(SCI_AUTOCSETTYPESEPARATOR, WPARAM('?'));
+	_pEditView->execute(SCI_REGISTERIMAGE, FUNC_IMG_ID, LPARAM(xpmfn));
```
donho, line 122: Here `SCI_REGISTERIMAGE` are called each time  `showApiComplete` is called.
Whereas we need only call it once for all (tested).
I guess it's the same case for both `SCI_AUTOCSETSEPARATOR` & `SCI_AUTOCSETTYPESEPARATOR`, but I didn't tested yet.

donho:
As I mentioned above, SCI_REGISTERIMAGE is needed to be called only once.
But I saw there are other calls (out of scope of this PR) can be also optimized.
This PR will be accepted and merged as it is. I'll do the optimization personally.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
vinsworldcom: Yes, since `SCI_AUTOCSETSEPARATOR` was already there, I assumed adding the others in the same would work.  This could be considered defensive in case some plugin changes the `SCI_AUTOCSETSEPARATOR` or registers an image with the same ID.  Off chance on the latter, but I use autocomplete in 2 plugins (QuickText, TagLEET) as well as some PythonScripts.  I always keep the `SCI_AUTOCSETSEPARATOR` = ` " "` and `SCI_AUTOCSETTYPESEPARATOR` = `?`; others - who knows?

Cheers.

PR #11092, opened 2022-01-27
Update german.xml to v8.3

dfs-:
* fix two functions not available in Shortcut Mapper because they shared the same name as other entries
* make translation more consistent and more faithful to the original
* fix cut-off text in MD5 and SHA-256 tool dialog as well as incremental search and the Shortcut Mapper
* add previously untranslated drop-down items in Preferences > Print
* add translation for WantToOpenHugeFile as referenced by #11047 / commit 46011e3


dfs-:
Requires #11115 to be merged to not expose the issue described in #11114.

PR #11093, opened 2022-01-27
Update english.xml to v8.3

dfs-:
* add translation for WantToOpenHugeFile as referenced by commit 46011e3
* add previously untranslated drop-down items in Preferences > Print

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -3,7 +3,7 @@
 The comments are here for explanation, it's not necessary to translate them.
 -->
 <NotepadPlus>
-    <Native-Langue name="English" filename="english.xml" version="8.2.1">
+    <Native-Langue name="English" filename="english.xml" version="8.2.2">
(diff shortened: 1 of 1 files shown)


rddim:
@dfs- 
Could you check that the `Add` button stop working if you translate for example `Full file name path` to something different please?
I add bulgarian localization for the combo strings, they appear in the menu, but the `Add` button does not work. This make me think that N++ need some code changes to work with translated strings?!

GOLEM777:
@rddim 
You are right, the code stopped working!

dfs-:
Will investigate later. Thought I had this tested, but seems I did not. :/

xomx:
@rddim 
Confirmed for czech.xml, so I am reverting the changes. Thanks.

dfs-:
@rddim @GOLEM777 @xomx

Implemented a fix that is a pull request on its own: #11115

dfs-:
@donho 
I guess you already solved it - thanks for the merge.

PR #11108, opened 2022-01-30
Update belarusian.xml

a-polivanchuk:
This is a revised Belarusian translation with many changes since the last update.
Please review it so it can be landed in future releases.


dfs-:
You might want to check out #11093 for changes made for v8.2.2

a-polivanchuk:
@dfs- If you don't mind I'd like to keep this PR as it is. I only pull changes from master :)
Thanks!

dfs-:
@a-polivanchuk yep, not a problem. 

I only thought it would be a good idea, as you seem to have invested a good amount of time updating the belarusian translation. The changes made in the aforementioned PR account for changes that are already pushed to master.

PR #11132, opened 2022-02-05
fixed cmdline '-n' param not working

### LINKED ISSUE #11131: -n doesn't work
8.3
-n doesn't work. For example from grepWin. But the faggots will love the banner.

xomx:
Fixes #11131 .

This fixes the signed/unsigned mismatch there arisen from the current 2GB+ changes.


Yaron10:
@xomx,

👍 
Thank you for your work.

PR #11158, opened 2022-02-08
fix for the /D= NSIS param regression (x64)

### LINKED ISSUE #11072: Target directory for the installation is ignored in silent mode
### Description of the Issue
The installer of version **8.2.1** ignores the provided target directory in silent mode.

```
notepadpp-8.2.1-x64.exe /S /D="C:\Program Files\NPP"
```

### Expected Behavior
The program should be installed in the target directory `C:\Program Files\NPP`

### Actual Behavior
The program is always installed in the directory `C:\Program Files\Notepad++`, no matter which target directory is specified.
This behavior has only started with version 8.2.1.

xomx:
Detailed explanation is in the .onInit patch comments.

This regression has been caused by the previous PR: #11013 .

Fixes #11072 .
Other report & testing:  https://community.notepad-plus-plus.org/topic/22495/d-programdata-notepad-not-working

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -111,7 +111,57 @@ InstType "Minimalist"
 Var diffArchDir2Remove
 Var noUpdater
 
+!ifdef ARCH64
+; this is needed for the 64-bit InstallDirRegKey attribute patch
+!include "StrFunc.nsh"
+${StrStr} # Supportable for Install Sections and Functions
+!endif
+
 Function .onInit
+
+	; --- PATCH BEGIN (it can be deleted without side-effects, when the NSIS N++ x64 installer binary becomes x64 too)---
+	;
+	; 64-bit patch for the NSIS attribute InstallDirRegKey (used in globalDef.nsh)
+	; - this is needed because of the NSIS binary, generated for 64-bit Notepad++ installations, is still a 32-bit app (this can be changed in the future),
+	;   so the InstallDirRegKey attribute checks for irrelevant RegKey (HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Notepad++), see also this:
+	;   https://nsis.sourceforge.io/Reference/SetRegView
+	; - for the closest InstallDirRegKey simulation possible, this should be right at the top of .onInit
+	;
+	; Notes:
+	;
+	; - before .onInit is executed, $INSTDIR has been preset by the InstallDir attribute
+	; - then if the value defined by a possible InstallDirRegKey attribute is valid, it overrides original InstallDir
+	; - finally, if the user used "/D=..." at runtime (at installer cmdline), $INSTDIR has been changed to that parameter before .onInit
+	;
+	; - so we have to know here, if the "/D" was used, but there is a problem (citation from the NSIS documentation):
+	;   "For parsing out the PARAMETER portion, see GetParameters. If /D= is specified on the command line (to override the install directory) it won't show up in $CMDLINE."
+	;
+	; - so one solution could be not to use the 'Installdir' & 'InstallDirRegKey' at all and then we can use here:
+	;	${If} $INSTDIR != ""
+	;		;/D was used for sure, so obey and do nothing
+	;	${Else}
+	;	...
+	;
+	; - another solution is to parse the original cmdline for the "/D=..." and then act accordingly (this is the chosen solution below)
+	;
+!ifdef ARCH64 || ARCHARM64 ; x64 or ARM64
+	${If} ${RunningX64}
+		System::Call kernel32::GetCommandLine()t.r0 ; get the original cmdline (where a possible "/D=..." is not hidden from us by NSIS)
+		${StrStr} $1 $0 "/D="
+		${If} "$1" == ""
+			; "/D=..." was NOT used for sure, so we can continue in this InstallDirRegKey x64 patch 
+			SetRegView 64 ; disable registry redirection (enable access to 64-bit portion of registry)
+			ReadRegStr $0 HKLM "Software\${APPNAME}" ""
+			${If} "$0" != ""
+				; a valid previous installation path has been detected, so offer that as the $INSTDIR
+				StrCpy $INSTDIR "$0"
+			${EndIf}
+			SetRegView 32 ; restore the original state (this is only for simulating the non-patched state as closely as possible)
+		${EndIf}
+	${EndIf}
+!endif
+	;
+	; --- PATCH END ---
 
 	${GetParameters} $R0 
 	${GetOptions} $R0 "/noUpdater" $R1 ;case insensitive 
@@ -148,9 +198,6 @@ updaterDone:
 	${If} ${RunningX64}
 		; disable registry redirection (enable access to 64-bit portion of registry)
 		SetRegView 64
-		
-		; 64-bit patch for the NSIS attribute InstallDirRegKey (used in globalDef.nsh)
-		ReadRegStr $INSTDIR HKLM "Software\${APPNAME}" ""
 		
 		; change to x64 install dir if needed
 		${If} "$InstDir" != ""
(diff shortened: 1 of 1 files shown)


donho:
@xomx 
I don't really like this solution but it seems the hack is needed for NSIS 32 installer's limitation and I don't see other solution. Hope there's no other regression caused by this PR.
Can comment be less verbose? It's good to a reviewer of PR, but it's too much for the comments which explain the code.

donho (changes requested):
Make comment less verbose.

donho (changes requested):
Make comment less verbose.

xomx:
@donho 

> Make comment less verbose.

Done.

> I don't really like this solution but it seems the hack is needed for NSIS 32 installer's limitation and I don't see other solution.

Me neither.

The right solution is to generate x64 NSIS binary for the x64 N++. NSIS team still not officially supports that ([explanation](http://forums.winamp.com/showthread.php?t=454174)), but I am successfully using the "Target AMD64-Unicode" possibility with the help of this [fork](https://github.com/negrutiu/nsis).

donho (approved):

donho:
@xomx 
I've got an error while compiling it:

image

Did you have compiled it and tested it?

PowerEditor/installer/nppSetup.nsi
```diff
@@ -111,8 +111,40 @@ InstType "Minimalist"
 Var diffArchDir2Remove
 Var noUpdater
 
+!ifdef ARCH64
+; this is needed for the 64-bit InstallDirRegKey patch
+!include "StrFunc.nsh"
+${StrStr} # Supportable for Install Sections and Functions
+!endif
+
```
donho, line 119: Remove this block please.

PowerEditor/installer/nppSetup.nsi
```diff
@@ -111,8 +111,40 @@ InstType "Minimalist"
...
 Function .onInit
 
+	; --- PATCH BEGIN (it can be deleted without side-effects, when the NSIS N++ x64 installer binary becomes x64 too)---
+	;
+	; 64-bit patch for the NSIS attribute InstallDirRegKey (used in globalDef.nsh)
+	; - this is needed because of the NSIS binary, generated for 64-bit Notepad++ installations, is still a 32-bit app,
+	;   so the InstallDirRegKey checks for the irrelevant HKLM\SOFTWARE\WOW6432Node\Notepad++, explanation:
+	;   https://nsis.sourceforge.io/Reference/SetRegView
+	;
+!ifdef ARCH64 || ARCHARM64
```
donho, line 129: Add `!include "StrFunc.nsh"` here.

donho (changes requested):
Fix ARM64 build error.

PowerEditor/installer/nppSetup.nsi
xomx: I have fixed that ARM64 build error, but this cannot be removed from where it is:
https://nsis.sourceforge.io/StrFunc#3.1_How_To_Use_Commands_In_Install_Sections_and_Functions

PowerEditor/installer/nppSetup.nsi
xomx: ditto above

xomx:
@donho 

> I've got an error while compiling it:
> 
I have fixed that, sorry for the inconvenience.

> Did you have compiled it and tested it?

Yes, of course, but only x64. I did not have the arm64 support installed in my MSVS, where I deal with this issue. So I had to download & reinstall the whole MSVS only for that.  Now I can build even the arm64. Unfortunately I cannot do a real testing with that, as I have not such HW.

chcg:
@donho @xomx 

I created a branch for testing the installer script, see https://github.com/chcg/notepad-plus-plus/tree/installer_build_check with appveyor result:
https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/42544468

Used https://github.com/snapcore/snapcraft/blob/main/windows/generate-self-signed-cert.ps1 to create a dummy certificate added to the repo. Additionally copied the necessary build data from release 8.3 to avoid building all the stuff.

Findings:
- unused files (https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/bin/NppShell.dll and https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/bin/NppShell64.dll), shouldn't they be deleted
- it was necessary to copy ..\src\langs.model.xml, ..\src\stylers.model.xml , ..\src\shortcuts.xml , ..\src\contextMenu.xml to bin which is normally done by the N++ postbuild step, maybe a conditional copy action could be added
- @donho do you use a script to build N++ and NppShell and copy Updater, Plugins, PluginList ? 
- @donho With newer versions of the signtool environment (Windows 10 SDK builds 20236 and newer) you need to specifying at least \fd (https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool). Failure on appveyor build visible from https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/42544385. 

> /fd SHA256

 is probably the appropriate setting.

donho:
@xomx 
No problem. I realized the error from the ARM64 build. It's not trivial to (think of) test(ing) ARM64.
And indeed, we need installer script (NSIS) part to be tested also in our CI process to prevent from such problem.

@chcg 
I just checked the new commit in your branch:
https://github.com/chcg/notepad-plus-plus/commit/30719e055513423f4502b8174ea1b9e9f0ccb060

There are too many files are not needed (IMO) in the repository. If testing NSIS scripts need all these files, I prefer to not test it in CI.

Here's how I build release: 
I build 3 notepad++.exe and 3 nppPluginList.dll for 3 platforms, with VS2017. I then copy 3 nppPluginList.dll on the right directories.
Before launching **packageAll.bat** `NPP_CERT_PWD`, `NPP_CERT` and `SIGN` variables are set. Then all the zip, 7z and nsis installer.exe are build by **packageAll.bat**. Note that we can set SIGN=0 for skipping the certificate signature.
But launching **packageAll.bat** is not necessary. We need only the following commands for checking the consistency of installer script:
```
"%PROGRAMFILES(x86)%\NSIS\makensis.exe" nppSetup.nsi
"%PROGRAMFILES(x86)%\NSIS\makensis.exe" -DARCH64 nppSetup.nsi
"%PROGRAMFILES(x86)%\NSIS\makensis.exe" -DARCHARM64 nppSetup.nsi
```

> unused files (https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/bin/NppShell.dll and https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/bin/NppShell64.dll), shouldn't they be deleted?

These 2 dll (NppShell.dll & NppShell64.dll) are identical as NppShell_06.dll of 32 bits and 64 bits respectively. They are the latest build  (so current version *_06) of NppShell.

> it was necessary to copy ..\src\langs.model.xml, ..\src\stylers.model.xml , ..\src\shortcuts.xml , ..\src\contextMenu.xml to bin which is normally done by the N++ postbuild step, maybe a conditional copy action could be added

Both vsproj file and packageAll.bat copy these files on the right place. But vsproj file provides the compile-test environment & packageAll.bat copy them for the build package purpose. It's important to dissociate between the both files IMO.

> @donho do you use a script to build N++ and NppShell and copy Updater, Plugins, PluginList ?

No. I compile them manually with VS2017, and copy them on the right place, on my PROD PC, which has the only role to build release packages. So only the necessarily updated components to be built and copied on each release.

> @donho With newer versions of the signtool environment (Windows 10 SDK builds 20236 and newer) you need to specifying at least \fd (https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool). Failure on appveyor build visible from https://ci.appveyor.com/project/chcg/notepad-plus-plus/builds/42544385.

OK, I'll check it - could you open the issue and assign it to me please?

PR #11191, opened 2022-02-12
Fix autocomplete case insensitive list

### LINKED ISSUE #11187: Fix autocomplete case insensitive list
Issue mentioned at Notepad++ community topic [Issue with autocompletion - comparing 7.3.3 and 8.3].

Excerpts of autocomplete results of `Tra` mentioned from my [reply]:

## 7.3.3 autocomplete displays:

```
TraceLine
TransformAndDrawLineList
```

## 8.3 autocomplete displays:

```
TRACE.ALL
TRACE.FLOODFILL
TRACE.FLOODFILLENT
TRACE.GRATE
TRACE.OPAQUE
TRACE.PLAYER
TRACE.PLAYERCLIP
TRACE.SHOT
TRACE.SMOKEBOMB
TRACE.SOLID
TRACE.VISIBLE
TRACE.WATER
```

These autocomplete lists are based on creation of a UDL named *test*, set with defaults, and this [autocomplete xml file] named as test.xml.

Need to set `1` to `3` in:

`Preferences -> Auto-Completion -> From 3 th character`

else you need to cancel autocomplete on the 2nd character and then type the 3rd character to display the autocomplete list for the 3rd character.

Set language to *test* in *Language* menu.

The autocomplete list for `T` can be different to the autocomplete list for `Tra` which important to note.

8.3 should not display that list for either ignorecase=no or ignorecase=yes, in my opinion.

 * If ignorecase=no and `TRA` was typed, then OK, though `Tra` was typed.
 * If ignorecase=yes, I would expect to see also `TraceLine` and `TransformAndDrawLineList` in the list.

This lead me to debug with help from Michael Vincent who pointed to the source file with the issue. I have considered with testing to have found a fix. Now the autocomplete list has a suitable selection. The results from `getWordArray()` was inadequate because of the search `flags` being used. `canStop` condition did not help either as the list is unsorted (yet alone case in...sensitivity) so I cannot see how it can `break` out of the `for` loop at the correct iteration.

Patch to fix the issue:

```diff
diff --git a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
index e7cf9efc..81e363ca 100644
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -164,7 +164,6 @@ bool AutoCompletion::showApiAndWordComplete()
 
 	// Add keywords to word array
 
-	bool canStop = false;
 	for (size_t i = 0, kwlen = _keyWordArray.size(); i < kwlen; ++i)
 	{
 		int compareResult = 0;
@@ -183,12 +182,6 @@ bool AutoCompletion::showApiAndWordComplete()
 		{
 			if (!isInList(_keyWordArray[i], wordArray))
 				wordArray.push_back(_keyWordArray[i]);
-			canStop = true;
-		}
-		else if (canStop)
-		{
-			// Early out since no more strings will match
-			break;
 		}
 	}
 
@@ -233,7 +226,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | _ignoreCase ? 0 : SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
 
 	_pEditView->execute(SCI_SETSEARCHFLAGS, flags);
 	intptr_t posFind = _pEditView->searchInTarget(expr.c_str(), expr.length(), 0, docLength);

```

If you want the patch as a PR, let me know, as I have a branch ready. The guidelines mentions that an issue is needed before a PR can be accepted.

  [Issue with autocompletion - comparing 7.3.3 and 8.3]: https://community.notepad-plus-plus.org/topic/22539/
  [autocomplete xml file]: https://subversion.assembla.com/svn/omnibot/Enemy-Territory/0.8/tools/Notepad++/autoCompletion/omni-bot.xml
  [reply]: https://community.notepad-plus-plus.org/post/74048

mpheath:
This a fix for issue titled **Fix autocomplete case insensitive list**

 * Fix #11187

Issue was submitted by myself and this is the PR to finalize.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -164,7 +164,6 @@ bool AutoCompletion::showApiAndWordComplete()
 
 	// Add keywords to word array
 
-	bool canStop = false;
 	for (size_t i = 0, kwlen = _keyWordArray.size(); i < kwlen; ++i)
 	{
 		int compareResult = 0;
@@ -183,12 +182,6 @@ bool AutoCompletion::showApiAndWordComplete()
 		{
 			if (!isInList(_keyWordArray[i], wordArray))
 				wordArray.push_back(_keyWordArray[i]);
-			canStop = true;
-		}
-		else if (canStop)
-		{
-			// Early out since no more strings will match
-			break;
 		}
 	}
 
@@ -233,7 +226,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | _ignoreCase ? 0 : SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
 
 	_pEditView->execute(SCI_SETSEARCHFLAGS, flags);
 	intptr_t posFind = _pEditView->searchInTarget(expr.c_str(), expr.length(), 0, docLength);
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -226,7 +226,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | _ignoreCase ? 0 : SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | (_ignoreCase ? 0 : SCFIND_MATCHCASE) | SCFIND_REGEXP | SCFIND_POSIX;
 
 	_pEditView->execute(SCI_SETSEARCHFLAGS, flags);
 	intptr_t posFind = _pEditView->searchInTarget(expr.c_str(), expr.length(), 0, docLength);
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -226,7 +226,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | (_ignoreCase ? 0 : SCFIND_MATCHCASE) | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
 
 	_pEditView->execute(SCI_SETSEARCHFLAGS, flags);
 	intptr_t posFind = _pEditView->searchInTarget(expr.c_str(), expr.length(), 0, docLength);


mpheath:
I have added parentheses around the ternary so that it can be evaluated before the rest of the line. This should help to avoid related compile and/or runtime issues.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
```diff
@@ -183,12 +182,6 @@ bool AutoCompletion::showApiAndWordComplete()
 		{
 			if (!isInList(_keyWordArray[i], wordArray))
 				wordArray.push_back(_keyWordArray[i]);
-			canStop = true;
-		}
-		else if (canStop)
-		{
-			// Early out since no more strings will match
-			break;
```
donho, line 191: Why `canStop` is removed?

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
```diff
@@ -233,7 +226,7 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | (_ignoreCase ? 0 : SCFIND_MATCHCASE) | SCFIND_REGEXP | SCFIND_POSIX;
```
donho, line 229: `getWordArray` method is for getting word list for word completion, but has nothing to do with function list of auto-completion.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: I used `OutputDebugString()` in the `for` loop. This is the saved log from Sysinternals DebugView from testing:

```
00000001	7:17:35 PM	[4012] Tra	
00000002	7:17:35 PM	[4012] int: "1"	
00000003	7:17:35 PM	[4012] TRA	
00000004	7:17:35 PM	[4012] TRA	
00000005	7:17:35 PM	[4012] TRA	
00000006	7:17:35 PM	[4012] TRA	
00000007	7:17:35 PM	[4012] TRA	
00000008	7:17:35 PM	[4012] TRA	
00000009	7:17:35 PM	[4012] TRA	
00000010	7:17:35 PM	[4012] TRA	
00000011	7:17:35 PM	[4012] TRA	
00000012	7:17:35 PM	[4012] TRA	
00000013	7:17:35 PM	[4012] TRA	
00000014	7:17:35 PM	[4012] TRA	
00000015	7:17:35 PM	[4012] -- break	
00000016	7:17:35 PM	[4012] -- break	
00000017	7:17:35 PM	[4012] -- break	
00000018	7:17:35 PM	[4012] -- break	
00000019	7:17:35 PM	[4012] -- break	
00000020	7:17:35 PM	[4012] -- break	
00000021	7:17:35 PM	[4012] -- break	
00000022	7:17:35 PM	[4012] -- break	
00000023	7:17:35 PM	[4012] -- break	
00000024	7:17:35 PM	[4012] -- break	
00000025	7:17:35 PM	[4012] -- break	
00000026	7:17:35 PM	[4012] Tra	
00000027	7:17:35 PM	[4012] Tra	
00000028	7:17:35 PM	[4012] -- break	
00000029	7:17:35 PM	[4012] -- break	
00000030	7:17:35 PM	[4012] -- break	
```

up to `00000462`.

 * 1st line is value of `beginChars`, which is what was typed into the edit pane.
 * 2nd line is value of `_ignoreCase`, which is to ignore case as shown.
 * 3rd to 462nd is the values of `kwSufix.c_str()` and the `break` statement is commented so that ` -- break` is printed out instead.

With `canStop` set to `true` on the 1st break, then only the `TRA` entries are processed and not the `Tra` entries. This is probably the main issue causing the missing entries in the autocomplete list.

`canStop` removed as it does not allow all entries for the autocomplete list.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: `getWordArray` is used in `AutoCompletion::showApiAndWordComplete()` and I noticed it was fixed  with `SCFIND_MATCHCASE` so that is what I changed 1st. Then I found still issues with the `for` loop with the autocomplete list so I found that the `break` ing out of the loop too early appears to be the issue. If handling `SCFIND_MATCHCASE` setting is not desired, then that could be undone. I considered handling `SCFIND_MATCHCASE` as OK, but perhaps consider as not needed now.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
donho: Just tested your PR:
It seems work with [test.xml](https://subversion.assembla.com/svn/omnibot/Enemy-Territory/0.8/tools/Notepad++/autoCompletion/omni-bot.xml). However, if we change ignoreCase to "no" in test.xml:
`<Environment ignoreCase="no" startFunc="(" stopFunc=")" paramSeparator="," terminal=";" />`
It doesn't work at all - the list contains a lot entries which should be excluded.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: I just tested with restoring

```
int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
```

and the output in DebugView is the same, so this line is not the main issue for the autocomplete so it can remain as it was before.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: At set on `ignoreCase` set to `no` in *test.xml*.

I get the same lists as in the issue report. `Tra` shows 2 entries like as the 7.3.3 list and `TRA` shows 12 entries like as the 8.3 list.

I can only recommend retest to see if same.

mpheath: Also make sure the setting **Preferences -> Auto-completion -> From `3`th character** is set so the autocomplete list is based on the 3rd character, not the 1st character. The autocomplete list should be more filtered on the 3rd character.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
donho: @mpheath 
My bad. It was set on `Function Completion` instead of `Function and Word Completion`. Sorry for the wrong accusation.
It seems "showApiAndWordComplete" that you modified is more accurate than "showApiComplete" now.
However, I have to test it more - to prevent from regression

On the meanwhile, please revert `(_ignoreCase ? 0 : SCFIND_MATCHCASE)` change.
Word completion by default is case sensitive, and it has nothing to do with the function completion's setting.

donho (changes requested):
Revert `int flags = SCFIND_WORDSTART | (_ignoreCase ? 0 : SCFIND_MATCHCASE) | SCFIND_REGEXP | SCFIND_POSIX;`

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: Revert done. We both made some mistakes. Appreciate the concern.

Issue #11135, opened 2022-02-05
Feature Request: line number alignemnet in the find all window


Rudo79:
### Description of the Issue
results of **Find all in the current document** are not aligned in the output window

### Steps to Reproduce the Issue
search for text located in several lines of the document

### Expected Behavior
![expected](https://user-images.githubusercontent.com/99095503/152656454-003510ca-5d06-4817-98e2-0097c79c248d.JPG)

### Actual Behavior
![current](https://user-images.githubusercontent.com/99095503/152656376-9cb9f367-20c9-4e7b-af34-62cc96727dae.JPG)

### Possible Solution
I'm happy with this simple solution in the file FindReplaceDlg.cpp
wsprintf(lnb, TEXT("%6u"), static_cast<int>(fi._lineNumber));
![solution](https://user-images.githubusercontent.com/99095503/152656545-de0e1ed1-d3af-4099-aea9-fbb51609f73f.JPG)

Rudo79:
the feature is already requestet by #11119
sorry for double post

PR #11195, opened 2022-02-12

Rudo79:
this small change aligns the line numbers in the search results window
it is hardcoded at 6 digits so it is better that without any alignment
see more in the issue #11135

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -4062,7 +4062,7 @@ void Finder::add(FoundInfo fi, SearchResultMarking mi, const TCHAR* foundline)
 	str += TEXT(" ");
 
 	TCHAR lnb[16];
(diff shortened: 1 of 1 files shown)


alankilborn:
Dubious that a 6-character-wide hardcoding will be accepted.

Rudo79:
> Dubious that a 6-character-wide hardcoding will be accepted.

I know that it is not the best solution but very good workaround for lot of notepad++ users
line shift is visible and distracting in the first lines up to 5 digits

alankilborn:
Isn't a better solution to retrieve the number of lines and then pad appropriately the the _right_ number of leading spaces?

Rudo79:
yes, it is. I fully agree with you
It was also my first idea but the information about number of lines is not available in the Finder class.
I am not yet that familiar with the whole Notepad++ source code and I have currently no time to dig deeper onto it.
maybe, you can help with that?

Rudo79:
I found how to get number of lines. see below (just for the testing purpose)
I will do it in the right way and extend the finder class. ASAP
@alankilborn thanks for your support!
leading_spaces

gawkla:

PR #11208, opened 2022-02-14

Rudo79:
added appropriate number of leading spaces for line number in the search result window


Rudo79:
pushed changed VS project :(

PR #11209, opened 2022-02-14

Rudo79:
added appropriate number of leading spaces for line number in the search result window
the issue #11135

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -3996,15 +3996,15 @@ void Finder::addFileNameTitle(const TCHAR * fileName)
 
 void Finder::setLineNumFormatStr()
 {
-   size_t NumLines = (*_ppEditView)->execute(SCI_GETLINECOUNT);
+	size_t NumLines = (*_ppEditView)->execute(SCI_GETLINECOUNT);
 	size_t NumDigits = 0;
-   while (NumLines != 0) {
-      NumLines = NumLines / 10;
-      ++NumDigits;
-   }
+	while (NumLines != 0) {
+		NumLines = NumLines / 10;
+		++NumDigits;
+	}
 
-   TCHAR strFormat[10];
-   wsprintf(strFormat, TEXT("%%%uu"), NumDigits);
+	TCHAR strFormat[10];
+	wsprintf(strFormat, TEXT("%%%uu"), NumDigits);
 	_LineNumFormatStr = strFormat;
 }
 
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3993,6 +3994,20 @@ void Finder::addFileNameTitle(const TCHAR * fileName)
 	_pMainMarkings->push_back(EmptySearchResultMarking);
 }
 
+void Finder::setLineNumFormatStr()
+{
+   size_t NumLines = (*_ppEditView)->execute(SCI_GETLINECOUNT);
+	size_t NumDigits = 0;
```
dfs-, line 4000: I guess this shouldn't be indented this far?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3993,6 +3994,20 @@ void Finder::addFileNameTitle(const TCHAR * fileName)
...
+   size_t NumLines = (*_ppEditView)->execute(SCI_GETLINECOUNT);
+	size_t NumDigits = 0;
+   while (NumLines != 0) {
+      NumLines = NumLines / 10;
+      ++NumDigits;
+   }
+
+   TCHAR strFormat[10];
+   wsprintf(strFormat, TEXT("%%%uu"), NumDigits);
+	_LineNumFormatStr = strFormat;
```
dfs-, line 4008: I guess this shouldn't be indented this far?

dfs-:
Works well for a single file, but looks a bit wonky using Find in Files:

image

Maybe you could modify the code to use the maximum length format for Find in Files?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
Rudo79: yes, you are right; I'll correct that together with the improvement for "Find in Files"

Rudo79:
I fixed the coding style of my changes

Rudo79:
I took a look at that it will be little bit more work to implement that.
I will do it in a separate task (if this implementation will be accepted)

alankilborn:
>"Find in Files": I took a look at that it will be little bit more work to implement that.
I will do it in a separate task (if this implementation will be accepted)

I don't think that will be accepted -- solution for all cases should be one submission.

Rudo79:
I'm going to create a new PR with fixed leading spaces for the "Find in Files" as well

gawkla:

PR #11237, opened 2022-02-18

Rudo79:
Added leading spaces for line numbers in Find All results
issue #11135
grafik

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1816,6 +1816,35 @@ bool Notepad_plus::findInFilelist(std::vector<generic_string> & fileNames)
 
 	const bool isEntireDoc = true;
 
+	size_t MaxLineNum = 0;
+	for (size_t i = 0; i < filesCount; ++i)
+	{
+		if (progress.isCancelled()) break;
+
+		bool closeBuf = false;
+		BufferID id = MainFileManager.getBufferFromName(fileNames.at(i).c_str());
+		if (id == BUFFER_INVALID)
+		{
+			id = MainFileManager.loadFile(fileNames.at(i).c_str());
+			closeBuf = true;
+		}
+
+		if (id != BUFFER_INVALID)
+		{
+			Buffer* pBuf = MainFileManager.getBufferByID(id);
+			_invisibleEditView.execute(SCI_SETDOCPOINTER, 0, pBuf->getDocument());
+			size_t nbLine = _invisibleEditView.execute(SCI_GETLINECOUNT);
+			if (nbLine > MaxLineNum)
+			{
+				MaxLineNum = nbLine;
+			}
+		}
+
+		if (closeBuf)
+			MainFileManager.closeBuffer(id, _pEditView);
+	}
+	_findReplaceDlg.SetNumOfLines(MaxLineNum);
+
 	for (size_t i = 0, updateOnCount = filesPerPercent; i < filesCount; ++i)
 	{
 		if (progress.isCancelled()) break;
@@ -1892,6 +1921,19 @@ bool Notepad_plus::findInOpenedFiles()
 
 	if (_mainWindowStatus & WindowMainActive)
 	{
+		size_t MaxLineNum = 0;
+		for (size_t i = 0, len = _mainDocTab.nbItem(); i < len; ++i)
+		{
+			pBuf = MainFileManager.getBufferByID(_mainDocTab.getBufferByIndex(i));
+			_invisibleEditView.execute(SCI_SETDOCPOINTER, 0, pBuf->getDocument());
+			size_t nbLine = _invisibleEditView.execute(SCI_GETLINECOUNT);
+			if (nbLine > MaxLineNum)
+			{
+				MaxLineNum = nbLine;
+			}
+		}
+		_findReplaceDlg.SetNumOfLines(MaxLineNum);
+
 		for (size_t i = 0, len = _mainDocTab.nbItem(); i < len ; ++i)
 		{
 			pBuf = MainFileManager.getBufferByID(_mainDocTab.getBufferByIndex(i));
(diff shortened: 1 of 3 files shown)


alankilborn:
@Rudo79 

Why is there so _much_ whitespace in front of the line numbers?

I would expect:

    Search "..."
        D:\...
            Line 11: ...
            Line 49: ...
            Line 87: ...
        D:\...
            Line  11: ...
            Line  49: ...
            Line  87: ...
            Line 125: ...
            Line 186: ...
            Line 247: ...

Rudo79:
because one of the test files has 161513 lines, i.e. 6 digits 
grafik

alankilborn:
Well that might work great for you, but features in Notepad++ have to work for everyone.
I wouldn't like it to work like you show.

I would, however, like this feature if it worked like I showed in my mocked up output (files are independent).

Also, it shouldn't use the line-count of the file for determining the amount of leading spacing, it should use the width of the highest line number with a hit in it.
So for your example, even though there are 161513 lines in the file, if the highest line number with a hit is line 125, then there should be no padding on the left for lines greater than 99, one padding space for lines 10 to 99, and two spaces of padding for lines 1-9.

dfs-:
@alankilborn: Since it should work for everyone, I found what you are expecting to look really bad, since you don't have the results aligned, see #11209. The new PR by @Rudo79 fixes this behaviour. IMHO this is a huge improvement both over the original variant, where search results where randomly jumping left to right and the original PR, which still hat the jumping when using the Find in Files search.

alankilborn:
>I found what you are expecting to look really bad, since you don't have the results aligned

It's a matter of opinion.  
I _do_ have the results aligned; I just think the alignment should "reset" with every file -- like I showed.
You think something else--fine.

And there shouldn't be huge amounts of whitespace, like @Rudo79 shows, regardless.  Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11237#issuecomment-1046015487

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3993,6 +3993,19 @@ void Finder::addFileNameTitle(const TCHAR * fileName)
 	_pMainMarkings->push_back(EmptySearchResultMarking);
 }
 
+void Finder::setLineNumFormatStr(size_t NumLines)
+{
+	size_t NumDigits = 0;
+	while (NumLines != 0) {
```
donho, line 3999: Bad brace position. Please check contribution guide line for the coding convention of this project.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1816,6 +1816,35 @@ bool Notepad_plus::findInFilelist(std::vector<generic_string> & fileNames)
 
 	const bool isEntireDoc = true;
 
+	size_t MaxLineNum = 0;
```
donho, line 1819: The variable name starts always with lowercase. Please check contribution guide line for the coding convention of this project.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1816,6 +1816,35 @@ bool Notepad_plus::findInFilelist(std::vector<generic_string> & fileNames)
...
+	for (size_t i = 0; i < filesCount; ++i)
+	{
+		if (progress.isCancelled()) break;
+
+		bool closeBuf = false;
+		BufferID id = MainFileManager.getBufferFromName(fileNames.at(i).c_str());
+		if (id == BUFFER_INVALID)
+		{
+			id = MainFileManager.loadFile(fileNames.at(i).c_str());
```
donho, line 1828: Loading every file to search (only for the aligning the line numbers) twice will make the performance issue in Find in Files feature.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -165,6 +167,7 @@ protected :
 	bool _purgeBeforeEverySearch = false;
 
 	generic_string _prefixLineStr;
+	generic_string _LineNumFormatStr;
```
donho, line 170: The variable name starts always with lowercase. Please check contribution guide line for the coding convention of this project.

donho:
It's a good feature. However, the PR will create the performance issue in Find in Files feature due to Loading every file to search twice. I'll see what I can do for this feature but it cannot happen to load file twice just for it.

Rudo79:
hello all;
FYI: I am on vacation until March 6th
I agree with @dfs- that the aligned results over all searched file should be the better solution for every one.
@alankilborn: It will have a big impact on performance in finding the minimum number of required leading spaces; at least with current implementation of searching
@donho: every help is appreciated!
I'll try to find better way to do it after my vacation.

alankilborn:
>I agree with @dfs- that the aligned results over all searched file should be the better solution for every one.

So I could do a first search, and there will be some chosen number of blank spaces for padding.
Then I could do a second, third, etc search and for each one there could be some different number of spaces.
Then I copy all results to the clipboard and try to manipulate them and I still have some "poor alignment" issues with the data.

Isn't a better and simpler solution to have a _Preference_ setting on the _Search_ tab such as `Space-pad line numbers in Search results to ___ digits"?  A default of 0 would indicate no padding (current behavior).

Rudo79:
@alankilborn: I was writing about one search with "Find in Files->Find all" and "Find all in All opened Documents"; so the results should have the same line number alignment. the next search will have the same or different alignment depend on searched files.
I can also imagine a solution where all serch results will be adapted to particular max line number but not with current search and draw implementation. I'll take a look at that after my vacation.

alankilborn:
>I can also imagine a solution where all serch results will be adapted to particular max line number 

I think this hints at my suggestion of:

>a Preference setting on the Search tab such as `Space-pad line numbers in Search results to ___ digits"? 

I also think that would be a very small code change, which I've noticed has a higher likelihood of acceptance.

gawkla:

PR #11255, opened 2022-02-21
fix autosaving session on exit (regression)

### LINKED ISSUE #11249: Session is Auto-Saved on Exit
### Description of the Issue
Session is saved after exiting npp even though "Multi-Instance settings" is set to "Default (Mono-instance)"

(using latest version 8.3.1)
1. In Preferences:
1.1 set "Multi-Instance" mode to "Default (Mono-instance)"
1.2 In "Backup", uncheck "Remember current session for next launch"
2. open npp session file ("notepad++.exe"  -multiInst -openSession "c:\path\to\mySession.nppSession")
3. make some session-modifications (e.g. change active tab, close tab, open another tab)
4. close npp
5. re-open the session file (as in step 2)

### Expected Behavior
session looks the same as in the end of step (2)

### Actual Behavior
session looks the same as in the end of step (3)

xomx:
A regression caused by the #10935 arisen from the incomplete (cmdline session loading was omitted) implementation of the loaded session auto-saving on exit.

Fixes #11249 .

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -6104,6 +6104,7 @@ std::vector<generic_string> Notepad_plus::loadCommandlineParams(const TCHAR * co
 		return std::vector<generic_string>();
 
 	NppParameters& nppParams = NppParameters::getInstance();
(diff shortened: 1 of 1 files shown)


donho:
This one also is not a regression to me but a refined behaviour.

PR #11259, opened 2022-02-21
N++ unnecessarily stalled at exit (regression)

### LINKED ISSUE #11219: N++ unnecessarily stalled at exit with a large file opened within (regression)
### Description of the Issue
This is only a small regression from the fix #11017 .

N++ is unnecessarily stalled in the situation, when closing a N++ with a large file opened within, but the session.xml file cannot be updated at all due to the current N++ settings chosen.

1) Open a 64-bit N++ (v8.3+).
2) Ensure:
   UNCHECKED ... Settings -> Preferences... -> Backup -> 'Remember current session for next launch'
   SELECTED ... Settings -> Preferences... -> Multi-Instance & Date -> Multi-Instance settings -> Default (mono-instance)
3) Open a larger file (the larger will be the file, the more visible will be the exit-delay, also choosing a Debug, instead of the usual Release, build helps to observe this behavior).
4) Close the opened N++ - there will be a noticeable delay before the N++ main window disappears.

### Expected Behavior
N++ window/process almost immediately exits with the settings described above.

### Actual Behavior
Unnecessarily frozen N++ window at the exit.

### Debug Information
Related to all the N++ v8.2.1+, but really visible for larger files, so v8.3+.

The delay originates from the ".\PowerEditor\src\NppBigSwitch.cpp" -> `Notepad_plus::process()` -> `case WM_CLOSE:`

```
	...
	Session currentSession;
	getCurrentOpenedFiles(currentSession, true);
	...
```

Inside the `Notepad_plus::getCurrentOpenedFiles()` is the following loop across the entire opened file, which is the real source of the stalled N++ at the exit, when a large file has been opened:

```
	for (size_t j = 0 ; j < maxLine ; ++j)
	{
		if ((_invisibleEditView.execute(SCI_MARKERGET, j) & (1 << MARK_BOOKMARK)) != 0)
		{
			sfi._marks.push_back(j);
		}
	}
```
	

Anyway, calling the `getCurrentOpenedFiles` is pointless with the settings described above, so here is a quick fix proposal:

```
	...
	Session currentSession;
	if (!((nppgui._multiInstSetting == monoInst) && !nppgui._rememberLastSession))
		getCurrentOpenedFiles(currentSession, true);
	...
```

xomx:
This is only a small regression from the fix #11017 .

N++ is unnecessarily stalled in the situation, when closing a N++ with a large file opened within, but the session.xml file cannot be updated at all due to the current N++ settings chosen.

Fixes #11219 .

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1975,7 +1975,8 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				}
 
 				Session currentSession;
-				getCurrentOpenedFiles(currentSession, true);
+				if (!((nppgui._multiInstSetting == monoInst) && !nppgui._rememberLastSession))
+					getCurrentOpenedFiles(currentSession, true);
 
 				if (nppgui._rememberLastSession)
 				{


donho:
It's not a regression to me but an enhancement.

PR #11267, opened 2022-02-22
Fix uppercase character in SUPPORTED_SYSTEM.md

GMLudo:
Hi,

The case of **NO** is different for Windows 8, this pull request fixes that.

Thanks for the merge


alankilborn:
The contributing guidelines say typos should just be an issue, not made into a PR.

PR #11124, opened 2022-02-03
Update french.xml 

fylip22:
Update french.xml

PR #11273, opened 2022-02-23
Update french.xml

Mateos81:
With most of [@fylip22](https://github.com/fylip22) changes and I believe all [@donho](https://github.com/donho) comments taken into account from https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11124


Mateos81:
Thank you!

I'd like to use this space for a question
I'm going through the file to ease its future maintenance
I'd like to edit it in 3 steps:
- Moving existing lines to match the english.xml file order;
- Adding missing entries;
- Editing existing ones if necessary, but at least the version at the top of the file

I'm going to do a commit per step
The question is: would you like a single PR for the 3 commits, or a PR per commit?

Issue #11253, opened 2022-02-21
Feature request: more NPPM_MESSAGES.


Leonard-The-Wise:
Extracted from this post [here](https://community.notepad-plus-plus.org/post/74432)

As I am writting a custom Lexer plugin that have custom indentation (It's on Git, but project still private, sorry), I currently have no way to read the Auto-Indent settings from Notepad and it causes conflicts with my plugin. Something like `NPPM_GETCURRENTAUTOINDENTSTATUS` would be nice to have, so I won't execute my indentation code with that flag turned on.

If possible, if I was able to disable the built-in auto-indentation temporarily from a command would be nice either (although not mandatory). Then I would check the current status, auto-set it to disable, and when language changed (or program exitted) I would restore the user previous state.

Also, I can’t read whether the user is currently recording/playing a macro properly. This also causes conflict with the plugin functionality. I already studied the code and found a `WM_GETCURRENTMACROSTATUS` but it is not enough, since it doesn’t tell me if a macro is running.

Thanks.

Ekopalypse:
I second this, in general it would be beneficial for plugin developers if a notification is available informing about configuration changes. Something like NPPN_CONFIGURATION_CHANGED with wparam an ID that points to the setting and lparam 0 or 1 for Boolean on/off and -1 when it is necessary to call another NPPM_ message to see the current status.

PR #11275, opened 2022-02-23
Add support for plugin custom Auto-Indentation

Leonard-The-Wise:
Based on this request:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11253

I added small changes to the code to allow plugins to control auto-indentation flux. Needed a couple new messages.
Also, as it is related to this function, Plugins now can also query for macro status { Idle, Recording, Stopped, PLaying }, so they can change behavior appropriatedly.


Leonard-The-Wise:
Actually, I've decided to change NPPM_GETLANGUAGEAUTOINDENTATION to separate the search status from the ACTUAL type of indentation. Will reopen pull in a second...

PR #11277, opened 2022-02-23
Add custom auto-indentation support (for external languages)

Leonard-The-Wise:
I added small changes to the code to allow plugins to control auto-indentation flux. Needed a couple new messages.
Also, as it is related to this function, Plugins now can also query for macro status { Idle, Recording, Stopped, PLaying }, so they can change behavior appropriatedly, and can query for Auto-Indentation settings from user preferences.

Associated Issue: #11253


Leonard-The-Wise:
Typo in MessageID codes (jumped 105)... remaking. :(

PR #11282, opened 2022-02-24
Bring light-mode tabs visuals in line with dark-mode

### LINKED ISSUE #10919: Bring light-mode tabs visuals in line with dark-mode
### Description of the Issue
Currently light-mode tabs use system default themeless visuals. This is inconsistent with what DarkMode tabs look like and also with themed system tab visuals (Windows 10). Also they just look ugly and I think many will agree.
DarkMode tabs are painted manually in WM_PAINT and are just plain rectangles which appears to be an accepted way of doing it.

Being a proud user of light theme and feeling constantly displeased with these visuals I made a little patch for personal use, which I would like to improve on and make it into official release if this gets approved. Screenshots are attached (no vertical, multi-line or coloured bar yet since I don't use them) 

### Additional info to discuss a fix
I remember finding out that tabs theme gets stripped in init TabBarPlus during _hwndArray processing or couple of lines after that point, but i didn't look into it more closely since manually painting in WM_PAINT might provide better visuals than default windows style and since it is already implemented like that for DarkMode

Launch npp with DarkMode disabled on Windows 7 or newer

### Expected Behavior
Visuals similar to themed system apps

### Actual Behavior
Ugly themeless visuals from Win98 era
npptabs

flotsamarch:
Resolves #10919


donho:
What's improved is minor (IMO) for the modification of code.
So this PR won't be integrated.

PR #11292, opened 2022-02-25
fix for UDL comment config (regression)

### LINKED ISSUE #11239: UDL comment configuration input fields are broken
### Description of the Issue
When opening the UDL dialog, select an existing UDL with something setup in the comment block, the input fields are empty. When type something in there, close notepad++ and reopen it again it is again empty. But when there is an existing UDL with configured comments they are used for syntax highlighting.

1. Open an existing UDL xml file, for example markdown._preinstalled.udl.xml and check <Keywords name="Comments"> if there is something configured
2. Open the UDL dialog and 
3. select the before checked language
4. switch to the "Comment & Number" tab
5. check "Comment line style" and "Comment style" blocks => they are empty
6. type in all the Fields something
7. close the dialog
8. close notepad++
9. open everything again and the fields are again empty

The xml is changed when something is written into the input fields, but not with the values what was typed in.
Example from markdown._preinstalled.udl.xml
Before: <Keywords name="Comments">00# 01 02((EOL)) 03&lt;!-- 04--&gt;</Keywords>
Open NP++, open the dialog and type something in the comment input fields. Close NP++
After: <Keywords name="Comments">0 1 2 3 4</Keywords>

### Expected Behavior
Display configured comment configuration and allow to config a new configuration for comments.

### Actual Behavior
See "Steps to Reproduce the Issue"

### Debug Information
Notepad++ v8.3.1   (64-bit)
Build time : Feb 13 2022 - 15:41:12
Path : C:\_PortablePrograms\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19044.1466
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll HexEditor.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll PythonScript.dll XMLTools.dll 

udl

xomx:
Caused by mistake in [#11047](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/94154b0f0e281b4bbfad9c59f8d01c9dff305c12) .

Fixes #11239 .

### DIFF
--- a/PowerEditor/src/ScintillaComponent/UserDefineDialog.cpp
+++ b/PowerEditor/src/ScintillaComponent/UserDefineDialog.cpp
@@ -535,11 +535,12 @@ void CommentStyleDialog::setKeywords2List(int id)
             IDC_COMMENT_CLOSE_EDIT
         };
 
-        for (size_t i = 0; i < sizeof(list)/sizeof(int); ++i)
+        TCHAR intBuffer[10] = { '0', 0 };
+        for (int i = 0; static_cast<size_t>(i) < sizeof(list) / sizeof(int); ++i)
         {
-            wstring intStr = std::to_wstring(i);
+            generic_itoa(i, intBuffer + 1, 10);
             ::GetDlgItemText(_hSelf, list[i], buffer, max_char);
-			convertTo(newList, max_char, buffer, intStr.c_str());
+            convertTo(newList, max_char, buffer, intBuffer);
         }
 
 		wcscpy_s(_pUserLang->_keywordLists[index], newList);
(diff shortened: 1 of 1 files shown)


donho:
@xomx 
Good job.
However, I don't really get it:
`prefix` is not modified in `void convertTo(TCHAR *dest, int destLen, const TCHAR *toConvert, const TCHAR *prefix)`, 
Why using wstring instead did cause the regression?

xomx:
@donho 
> Why using wstring instead did cause the regression?

Using `wstring` is not the source of this regression. You deleted the preformatted `intBuffer`, e.g. here is the relevant code in `CommentStyleDialog::setKeywords2List()` before(ok) and after (regression):
UDL-comment-regression

xomx:
@donho 

The original way (without regression):
- the preformated `intBuffer` just initialized (it starts with the '0'-char, the rest 9 ones are NULs)
UDL-intBuffer-init
- then the current `i` integer is written at the `intBuffer+1`!
- so when the `i` is e.g. 1, the original way creates "01" for the `prefix` param of `convertTo()` (and not "1"...)

You could probably use the wstring-way, if you keep the right format.

(As I am not well-conversant with the UDL-stuff, I chose the exact old handling way and changed only your intended signed/unsigned correction in the for-cycle).

donho:
> then the current i integer is written at the intBuffer+1!

That's the detail that I didn't notice.

Thank you for the explanation @xomx !

PR #11295, opened 2022-02-25
Update langs.model.xml

Mr-Update:
- Extensions added
- Name corrected
- Indentation corrected


alankilborn:
I don't think this is going to be accepted...

PowerEditor/src/langs.model.xml
```diff
@@ -178,8 +178,8 @@
...
         </Language>
-        <Language name="ini" ext="ini inf url wer" commentLine=";"/>
-        <Language name="inno" ext="iss" commentLine=";" commentStart="" commentEnd="">
+        <Language name="ini" ext="cfg conf hhp ini inf url vmoptions wer" commentLine=";"/>
```
dfs-, line 181: cfg/conf added as ini file? This is plain wrong tbh, as these extensions are used by many applications with many different formats.

donho:
only `aspx` is approved.
I'll add it myself.

PR #11314, opened 2022-03-01
AutoComplete icons for func=yes and others

### LINKED ISSUE #11311: Fx icon shown for keywords that are not functions
### Description of the Issue

After the implementation of #11087, "fx" icon is shown for all keywords from auto-completion file. The icon should be shown only for keywords with attribute `func="yes"`.

1. Open empty file.
2. Set language to `C`.
3. Type `ab`.

screenshot

### Expected Behavior

"fx" icon is shown only for functions (`abort`, `abs`, `access`, `acos`), not for keywords (`absread`, `abswrite`, `acosl`, `allocmem`, `arc`).

### Actual Behavior

"fx" icon is shown for all keywords (including ones that are not functions) from auto-completion file.

### Debug Information

Notepad++ v8.3.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1526
Current ANSI codepage : 1250
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll XMLTools.dll

vinsworldcom:
Fix #11311

When using the new "fx" icon in autocomplete for "autoComplete/*.xml" files, there is a desire to distinguish between actual "functions" - those that have the XML parameter `func="yes"` set and those that don't.

This PR adds another XPM image taken from the Notepad++ function list (the icon for a function name - a light purple box) and uses that for non-function entries in the "autoComplete/*.xml" files.  We add the new XPM, register it with the other one and then in the routine where we add the "\x1E IMG_ID", we just test if the XML node has the `func` parameter set to `yes`.  If so, use the "fx" icon, if not, use the new box one.

Examples are shown in the referenced issue.

Cheers.


donho:
As I mentioned in #11311, `fx` for distinguishing from word entries. 
This PR will make more complain for the API files without `func` attribute at all, then we will run after the problem by adding the icons over and over again.
I prefer to keep the concept as simple as possible. Sorry that I have reject this PR.

vinsworldcom:
@donho understand about simplicity, my [argument for the counterpoint is in the issue](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11311#issuecomment-1057485498).  Understood if you ultimately choose to reject this - no worries.

Cheers.

PR #11316, opened 2022-03-01
added QuietUninstallString capability for the winget

### LINKED ISSUE #11310: Register QuietUninstallString in Registry to be fully compatible with Winget
Hello,

as you can see in [this](https://github.com/microsoft/winget-pkgs/issues/49601) issue I created in winget-pkgs, Notepad++ silent uninstall would work in winget if it would implement writing the QuitUninstallString registry value. With that, winget would know what arguments to use for a silent uninstall.

Is it possible to implement this? We're automating Notepad++ silent installation on corporate machines with winget, and the absence of this feature forces us to bypass winget and use a custom uninstall command.

Thank you!

xomx:
It makes the silent uninstall of N++ via 'winget' possible.

Fixes #11310 .

Refs:
https://docs.microsoft.com/en-us/windows/package-manager/winget/uninstall#options
https://nsis.sourceforge.io/Add_uninstall_information_to_Add/Remove_Programs#Recommended_values

### DIFF
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -172,6 +172,7 @@ Function writeInstallInfoInRegistry
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MajorVersion" "${VERSION_MAJOR}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MinorVersion" "${VERSION_MINOR}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "UninstallString" '"$INSTDIR\uninstall.exe"'
+	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "QuietUninstallString" "$\"$INSTDIR\uninstall.exe$\" /S"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "DisplayIcon" "$INSTDIR\notepad++.exe"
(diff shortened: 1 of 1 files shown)


PowerEditor/installer/nsisInclude/tools.nsh
```diff
@@ -172,6 +172,7 @@ Function writeInstallInfoInRegistry
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MajorVersion" "${VERSION_MAJOR}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "MinorVersion" "${VERSION_MINOR}"
 	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "UninstallString" '"$INSTDIR\uninstall.exe"'
+	WriteRegStr HKLM "${UNINSTALL_REG_KEY}" "QuietUninstallString" "$\"$INSTDIR\uninstall.exe$\" /S"
```
dfs-, line 175: While your definition is technically correct I would suggest using the same format as with `UninstallString`, which would result in `'"$INSTDIR\uninstall.exe"' /S`.

Only a suggestion to make it more readable and in line with the others.

xomx: Noticed that, but after all the tests I & the others did and I have the habit of not changing anything after. Guess this time it will be ok, so I am adjusting that.

donho:
@xomx Has anyone tested this PR?

xomx:
@donho 
> Has anyone tested this PR?

Yes. Me (I only checked the correct writing of the `QuietUninstallString` to the Registry by N++ installer) and two external users also the MS Winget part. See inside the issue.

PR #11323, opened 2022-03-02
Re-enable possibility to center the FindReplaceDlg on parent (#9201)

### LINKED ISSUE #9201: Find dialog positioning should be alongside Notepad++
### Description of the Issue

The Find dialog saves it's last position, thus being sometimes far away from the main Notepad++ application. Especially in multi-monitor environments. I often have to move Notepad++ to different monitors for various reasons (i. e. other workspace or a RDP connection related to my work in Notepad++ is on the screen I used Notepad++ beforehand). The Find dialog then may be opened on the "wrong" screen.

In a three-monitor setup, where the monitors rest aside
image

1. Move Notepad++ to monitor 3 (leftmost)
2. Open the Find dialog -> will be also placed on monitor 3
3. Close the Find dialog
4. Move Notepad++ to monitor 1 (rightmost)
5. Open the Find dialog -> will still be placed on monitor 3, leaving you wondering if it has even opened

### Expected Behavior

The find dialog should always be placed on the same screen as Notepad++.

### Actual Behavior

The find dialog stores it's absolute position making it sometimes detach from Notepad++.

### Debug Information

Notepad++ v7.9.1   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 2009
OS Build : 19042.572
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll XMLTools.dll

dfs-:
Re-enable possibility to center the FindReplaceDlg on parent (fixes #9201)

* add goToCenter to hotkey check in IDM_SEARCH_(FIND|REPLACE|FINDINFILES|MARK)

Before commit aa69711d it was possible to center the dialog on Notepad++ using a second CTRL+F but was lost during the implementation of #10019, #10177.

This will re-add the function:
* When the hotkey used is not associated with the current tab, it will switch to the new tab (as before).
* If the hotkey used is associated with the current tab, it will center the FindReplaceDlg on Notepad++.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1287,22 +1287,34 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 				return TRUE;
 
 				case IDM_SEARCH_FIND:
-					enableReplaceFunc(false); // enable relace false so only find
+					if (_currentStatus == FIND_DLG)
+						goToCenter();
+					else
+						enableReplaceFunc(false);
 					return TRUE;
 
 				case IDM_SEARCH_REPLACE:
-					enableReplaceFunc(true);
+					if (_currentStatus == REPLACE_DLG)
+						goToCenter();
+					else
+						enableReplaceFunc(true);
 					return TRUE;
 
 				case IDM_SEARCH_FINDINFILES:
-					enableFindInFilesFunc();
+					if (_currentStatus == FINDINFILES_DLG)
+						goToCenter();
+					else
+						enableFindInFilesFunc();
 					return TRUE;
 
 				case IDM_SEARCH_MARK:
-					enableMarkFunc();
+					if (_currentStatus == MARK_DLG)
+						goToCenter();
+					else
+						enableMarkFunc();
 					return TRUE;
 
-				case IDREPLACE :
+				case IDREPLACE:
 				{
 					std::lock_guard<std::mutex> lock(findOps_mutex);
 


alankilborn:
At first my comment was going to be "you only handled 4 of the 5 tabs of the Find window".  But after some research, I see that _Find in Projects_ isn't supported as fully as the others; I wonder why?

alankilborn:
@dfs- See https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11324

donho:
@dfs- 
Without this PR, if the focus is on Notepad++ then you type Ctrl+F, the Find dialog will be back alone side of Notepad++.
That's why this ability was removed.

dfs-:
@donho 
Please pardon me, but I don't understand your point. As referenced in the issue #9201 I often have a multi-monitor setup where Notepad++ gets moved from one screen to another regularly based on my current workflow. 

Doing this will detach the FindReplaceDlg, as it stores it's own position. So maybe I had Notepad++ open und the left-most screen and used the dialog there, then closing the finder and moving Notepad++ it to the right-most screen. Opening the find dialog again might have it on the left-most screen whereas Notepad++ still is on the right-most. 

It was very practical to just use CTRL+(F, F) to quickly center the find dialog again on the Notepad++ main window. As of now, I am not able to do this due to the change linked in the commit. Is there any way of accomplishing said workflow without this PR that I am not aware of?

OzBob:
@xomx 
the Ctrl+F followed by the Ctrl+H 
will not centre the Find Dialog

double Ctrl+F 
then Ctrl+H 
will

xomx:
@OzBob 
> the Ctrl+F followed by the Ctrl+H
> will not centre the Find Dialog

From my side it was only a confirmation that this PR does not interfere with the existing Ctrl+F + Ctrl+H functionality. I did not mean to use that as an alternative shortcut for the re-centering...

PR #11327, opened 2022-03-02
fix for the possible crash when loading an older session file

### LINKED ISSUE #11325: NPP crashes on opening a session file
**STR:**
Create two files: _C:\T1.txt_ and _C:\T2.txt_.
Create a session file containing the following lines:
```
<NotepadPlus>
    <Session activeView="0">
        <mainView activeIndex="0">
            <File firstVisibleLine="0" xOffset="0" scrollWidth="514" startPos="0" endPos="0" selMode="0" lang="None (Normal Text)" encoding="-1" filename="C:\T1.txt" backupFilePath="" originalFileLastModifTimestamp="1462017716" />
        </mainView>
        <subView activeIndex="0">
            <File firstVisibleLine="0" xOffset="0" scrollWidth="514" startPos="0" endPos="0" selMode="0" lang="None (Normal Text)" encoding="-1" filename="C:\T2.txt" backupFilePath="" originalFileLastModifTimestamp="1461982719" />
        </subView>
    </Session>
</NotepadPlus>
```
Open that session file (i.e. double-click on it or drag it into NPP).

**Result:**
NPP crashes.

**Caused by** https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ca8c302805f882ed61717bb26efbc141a6991ea7.

***

Notepad++ v8.3.2   (64-bit)
Build time : Feb 27 2022 - 16:11:01
Path : C:\---\Notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

xomx:
The 2GB+ adaptation has forced the change in session integers saving via TinyXML (now saving indirectly as strings). While the `MapPosition` part inside the `NppParameters::getSessionFromXmlTree()` has been already fortified against possible missing session-file internal items, the `Position` part not. So implemented likewise the `MapPosition` way there.

Fixes #11325 .

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -2175,14 +2175,31 @@ bool NppParameters::getSessionFromXmlTree(TiXmlDocument *pSessionDoc, Session& s
 				if (fileName)
 				{
 					Position position;
-					position._firstVisibleLine = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("firstVisibleLine"))));
-					position._xOffset = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("xOffset"))));
-					position._startPos = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("startPos"))));
-					position._endPos = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("endPos"))));
-					position._selMode = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("selMode"))));
-					position._scrollWidth = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("scrollWidth"))));
-					position._offset = static_cast<intptr_t>(_ttoi64((childNode->ToElement())->Attribute(TEXT("offset"))));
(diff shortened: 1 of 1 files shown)


Yaron10:
@xomx,

👍 
Thank you for the quick fix.

donho:
No, it's not enough.
You have review ca8c302805f882ed61717bb26efbc141a6991ea7 to fix the crash.

donho (changes requested):
Review ca8c302805f882ed61717bb26efbc141a6991ea7 to add the fix for preventing the crash.

xomx:
@donho 

I re-checked and I did not find any other problem like that in the rest of my 2GB+ patch.

The relevant problematic part of the `NppParameters::getSessionFromXmlTree()` worked ok before my 2GB+ patch, because of the underlying TinyXML `TiXmlElement::Attribute( const TCHAR * name, int* i )` returns 0 as a default in the `i` in case of invalid/missing `name` of the item from the session.xml file.


The rest of the code, I rewrote in my 2GB+ patch, has already been of the following type anyway, so it took into account even a possible NULL-string:
```
int32_t mapPosVal;
const TCHAR* mapPosStr = (childNode->ToElement())->Attribute(TEXT("mapFirstVisibleDisplayLine"), &mapPosVal);
if (mapPosStr)
	mapPosition._firstVisibleDisplayLine = mapPosVal;
```

donho (approved):

PR #11342, opened 2022-03-05
Fixed blurry disabled static texts in Dark mode

### LINKED ISSUE #10823: Unavailable options text "ghost" on some windows in Dark Mode
### Description of the Issue
In the preferences window in the Dark mode tab, the text for options that is unavailable appears to print twice:
image
That can be seen for the text beside the large checkboxes. It looks like there is a ghost of the text on top of the original text and slightly displaced. This also appears in some other places where there are unavailable options in the various windows that can be opened. This only happens in Dark Mode.

### Expected Behavior
Unavailable text should be gray (while normal text is white).

### Actual Behavior
Text appears to print twice.

### Debug Information
Notepad++ v8.1.9.2   (64-bit)
Build time : Nov 21 2021 - 04:30:20
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1348
Current ANSI codepage : 1252

### LINKED ISSUE #11331: Highlighted entry in Plugins Admin with dark mode not distinguishable
### Description of the Issue
When using dark mode a searched highlighted entry in plugins admin is very difficult to recognize, in light mode it is easy viewable.

1. use dark mode
2. open plugins admin
3. search a plugin
4. compare with light mode

### Expected Behavior
The highlighted entry should be distinguishable from the not highlighted entries

### Actual Behavior
The highlight is very difficult to distinguish from the rest

grafik

compare that to light mode:
grafik

Notepad++ v8.3.2   (32-bit)
Build time : Feb 26 2022 - 23:11:29
Path : C:\Users\Chromwell\Documents\Tools\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19043.1526
Current ANSI codepage : 65001
Plugins : BetterMultiSelection.dll ComparePlugin.dll CustomizeToolbar.dll LuaScript.dll mimeTools.dll NppCalc.dll NppConverter.dll NppExport.dll NppMenuSearch.dll

Ashfaaq18:
Fix #10823, fix #11331

Preferences -> Dark mode tab

BlurryTextFix_GIF

Preferences -> File Association tab

image

Preferences -> Backup

image

Preferences -> Auto-Completion

BlurryTextFix1_GIF

### DIFF
--- a/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
+++ b/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
@@ -125,7 +125,7 @@ intptr_t CALLBACK RegExtDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lPa
 				result = NppDarkMode::onCtlColorDarker(reinterpret_cast<HDC>(wParam));
 			}
 			//set the static text colors to show enable/disable instead of ::EnableWindow which causes blurry text
-			if (
+			if  (
 				(HWND)lParam == ::GetDlgItem(_hSelf, IDC_SUPPORTEDEXTS_STATIC) ||
 				(HWND)lParam == ::GetDlgItem(_hSelf, IDC_REGISTEREDEXTS_STATIC)
 				)
--- a/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
+++ b/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
@@ -83,6 +83,7 @@ void RegExtDlg::doDialog(bool isRTL)
 
 intptr_t CALLBACK RegExtDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM lParam)
 {
+	NppParameters& nppParam = NppParameters::getInstance();
 	switch (Message)
 	{
 		case WM_INITDIALOG :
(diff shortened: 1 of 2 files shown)


Yaron10:
@Ashfaaq18,

I don't use Dark Mode. Just tested it for your PR.

Very nice! 👍
Thank you for your work. 

I've modified the system fonts.
Still, you might want to have a look.

תמונה

תמונה

תמונה

תמונה

תמונה

***

BTW, the Dark Mode tab needs some adjustment for RTL.

תמונה

Ashfaaq18:
>I've modified the system fonts.

You changed this by modifying your PC's registry settings yeah? 

>Still, you might want to have a look.

There's quite a lot to be fixed in this case. If this fix is good to go, ill continue on with the rest.

> BTW, the Dark Mode tab needs some adjustment for RTL.
> 
> תמונה

would be great if you could open a new issue regarding this

Yaron10:
@Ashfaaq18,

> You changed this by modifying your PC's registry settings yeah?

Yes.

> If this fix is good to go, ill continue on with the rest.

👍 

***
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11343

donho:
Hmm... So if we add a (some) text static text(s) and if we need to disable it (them), we have to use your way instead of EnableWindow()...

Ashfaaq18:
> Hmm... So if we add a (some) text static text(s) and if we need to disable it (them), we have to use your way instead of EnableWindow()...

unfortunately yeah, The EnableWindow() function on static texts adds a second white font layer beneath the disabled text. This is of course not clearly visible in a normal white background. 
So instead of disabling the static texts as they are not interactive, I simply recolor them to grey when disabled.

I found an old [stackoverflow post](https://stackoverflow.com/questions/18945787/win32-static-font-wrong-when-grayed-out) regarding this. It does not have a solution to the problem, but there is a suggestion. Which is to handle the control in the WM_CTLCOLORSTATIC message.

ozone10:
@Ashfaaq18 
great work.

Can you also do the same for Column editor and Style configurator?
image

Ashfaaq18:
> @Ashfaaq18 great work.
> 
> Can you also do the same for Column editor and Style configurator? image

@ozone10 , yeah sure ill try fixing this. btw @donho, should we open a new issue for this? or ill simply link a new PR to this closed issue?

Issue #7901, opened 2020-01-29
Folding for CMake incorrectly folds functions and elseif


ssproessig:
### Description of the Issue`
Example CMake file `example.cmake`

```cmake
function(function1)
endfunction()

function(function2 PARAM)
    if(${CMAKE_CXX_COMPILER_ID} STREQUAL GNU)
        list(APPEND options ${COMPILE_OPTIONS_GCC})
    elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL Clang)
        list(APPEND options ${COMPILE_OPTIONS_CLANG})
    elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
        list(APPEND options ${COMPILE_OPTIONS_MSVC})
    endif()

    target_compile_options(${TARGET} PRIVATE ${options})
endfunction()

function(function3 PARAM P2)
endfunction()
```

### Steps to Reproduce the Issue
1.  Copy `example.cmake` into a file
2. Open in Notepad++

### Expected Behavior
- three functions to fold (`function1`, `function2`, `function3`)
- each `if`/`elseif` can be folded

### Actual Behavior
- `elseif` folding does not handle multiple `elseif` correctly (fold from first `elseif` to `endif`)
- `function` folding breaks in `function2` - folding everything until last `endfunction`

### Screenshots
![cmake_folding_1](https://user-images.githubusercontent.com/1576323/73391277-e8d69400-42d7-11ea-9560-84deaf913bdb.png)

![cmake_folding_3](https://user-images.githubusercontent.com/1576323/73391280-e96f2a80-42d7-11ea-8a3d-0141570b9d17.png)

![cmake_folding_2](https://user-images.githubusercontent.com/1576323/73391279-e96f2a80-42d7-11ea-82fa-e2112d28e2cc.png)

### Debug Information
```
Notepad++ v7.8.2   (64-bit)
Build time : Dec  4 2019 - 01:39:49
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.592
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 
````

KeyArts:
Hello, 

this error has been bothering me for a while, so I started to investigate it.

I think the solution is quite simple.

[...]

PR #11352, opened 2022-03-07
LexCmake.cxx: Avoid incrementation of foldLevel in case of "ElseIf" Token

KeyArts:
Fix notepad-plus-plus#7901

Tested with ExampleCode given in Bug-Report. 
Tested with "fold.at.else" enabled and disabled


chcg:
Scintilla changes should not be done here, but at https://github.com/ScintillaOrg/lexilla/blob/master/lexers/LexCmake.cxx. See also comment at #7901

donho:
v8.3.3 RC will be out in a couple of days and it should be stable enough.
I will work on Scintilla 5+ update as soon as v8.3.3 release is done.
And I believe this bug has been fixed in Scintilla 5+. If not, then you should do a PR toward to Scintilla project.

PR #11355, opened 2022-03-07
Fix append extension when saving with Return (win 7)

### LINKED ISSUE #11305: NP++ 8.3.1+Extension type not appended when saving with Return (win 7)
### Description of the Issue

As reported previously in this issue: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9657

The bug where the extension type is not appended when saving with the Return key is back on win 7 in NP++ 8.3.1 or higher.

NP++ 8.2.1 does not have this issue so I'm guessing it is as a result of the following commit: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b5a5baf13bb3f0b041f49698afb74af2b7059f6a

@mere-human should be familiar with this having fixed the previous issues.

ALT -> F -> A -> for the "Save As..." menu, type filename and hit Return.

The CTRL+ALT+S shortcut followed by hitting Return will also replicate the issue.

### Expected Behavior
Output: lol.txt

### Actual Behavior
Output: lol

### Debug Information

Notepad++ v8.3.2   (64-bit)
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Build : 7601.0
Current ANSI codepage : 1252

mere-human:
For some reason, the file name ComboBox isn't present when enumerating windows in OnFolderChange().
Because of that, we can't call onPreFileOk() in keyboard hook procedure.
Fix that by retrying window search in OnSelectionChange() if we didn't find them previously.

Fix #11305

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -269,8 +269,10 @@ public:
 	}
 	IFACEMETHODIMP OnSelectionChange(IFileDialog*) override
 	{
-		// This event isn't triggered in an empty folder.
+		// This event may not be triggered in an empty folder (Windows 10+).
 		// Dialog startup calling order: 4. Main window is shown.
+		if (!foundControls())
+			findControls();
 		return S_OK;
 	}
 	IFACEMETHODIMP OnShareViolation(IFileDialog*, IShellItem*, FDE_SHAREVIOLATION_RESPONSE*) override


mere-human:
So I found a VM with Windows 7 and somehow was able to fix the issue.
Also, it would be nice to re-test this bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10436

donho:
@mere-human 
Thank you for fixing it under the war.
Take care & good luck!
You're not alone, the world is with Ukraine, Notepad++ is with Ukraine.

mere-human:
> @mere-human Thank you for fixing it under the war. Take care & good luck! You're not alone, the world is with Ukraine, Notepad++ is with Ukraine.

Thank you for standing with Ukraine and spreading the information!
This matters a lot.
Let's hope this war ends soon.

PR #11356, opened 2022-03-08
Add new ability for filtering incompatible plugins

### LINKED ISSUE #11349: notepad++ 8.3.2 crashes on Windows 7
### Description of the Issue
notepad++ 8.3.2 crashes on Windows 7 (Version 6.1.7601)
notepad++ 8.2.1 works on Windows 7 (Version 6.1.7601)

### Expected Behavior
To not crash.

### Actual Behavior
Crashes on startup.

### LINKED ISSUE #11358: Improve Notepad++ stability by filtering better incompatible plugins on startup
In commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f9118dd13c637a4054c101e685c9030115f790c8 "npp-compatible-versions" has been added in plugin list for helping not only Pllugin Admin but also Plugin Manager (Plugin Loader) to check the compatibility with the latest version of plugin (if the version of installed plugin is the latest one).

However, for Plugin Manager if the version of installed plugin is the previous one, there's no info to know if this plugin with old version is compatible with running Notepad++. In order to filter better old version plugins, "old-versions-compatibility" is added in plugin list, so Plugin Manager can use it to check the compatibility to decide load it or not.

donho:
In commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f9118dd13c637a4054c101e685c9030115f790c8 **"npp-compatible-versions"** has been added in plugin list for helping not only Plugin Admin, but also Plugin Manager (Plugin Loader) to check the compatibility with the latest version of plugin (if the version of installed plugin is the latest one).
However, if the version of installed plugin is the previous one, there's no info to know if this plugin with old version is compatible with running Notepad++. In order to filter better old version plugins, **"old-versions-compatibility"** is added in plugin list, so Plugin Manager can use it to check the compatibility to decide load it or not.

The value of "old-versions-compatibility" is a string made by 2 interval versions as following:
`"old-versions-compatibility": "[,2.6][,8.2.1]"`

The first interval versions are for old plugin versions: [,2.6] means from all the versions to v2.6 included.
The second interval versions are for Notepad++ versions: [,8.2.1] means from all the versions to v8.2.1 included.
Both interval versions together means: the plugin in question from the 1st version to v2.6 is compatible with Notepad++ v8.2.1 and all Notepad++ previous versions.

Here is the sample:
```json
{
	"folder-name": "mimeTools",
	"display-name": "Mime tools",
	"version": "2.7",
	"npp-compatible-versions": "[8.3,]",
	"old-versions-compatibility": "[,2.6][,8.2.1]",
	"id": "b65fbfaa15b443131eb69188069cacbff04eca66e0cb84130631303a1d3895f8",
	"repository": "https://github.com/npp-plugins/mimetools/releases/download/v2.7/mimetools.v2.7.zip",
	"description": "Implements several main functionalities defined in MIME.",
	"author": "Don HO",
	"homepage": "https://github.com/npp-plugins/mimetools"
},
```
Considering these 2 scenarios:
1. The current distributed plugin version is v2.7 (the latest one normally), and it is compatible with all Notepad++ version from v8.3. If installed MineTools plugin is v2.7 and running Notepad++ is v8.3.3 then it's OK, and MineTools v2.7 will be loaded.

2. If installed MineTools plugin is v2.6 and running Notepad++ is v8.3.3 then Plugin Manager learns it's a previous version (v2.6 < v2.7), so it will check "old-versions-compatibility" - it will find v2.6 is in the 1st interval [,2.6] - it's matched, then Plugin Manager keep checking the running Notepad++'s compatibility: v8.3.3 is not in the 2nd interval [,8.2.1] - it's not compatible, then MineTools plugin v2.6 won't be loaded.

Here are the test instructions for this PR:
1.  download x64 debug binary here:
https://ci.appveyor.com/api/buildjobs/5ba501ecu21k28i7/artifacts/Notepad%2B%2B.x64.Debug.exe
2. Create an empty file "nppPluginList.json" in your `npp\plugins\Config\`
and copy the following content 
https://github.com/notepad-plus-plus/nppPluginList/blob/5e23fce86813ae434dd123bb33b1dcd7db346050/src/pl.x64.json
and paste into `npp\plugins\Config\nppPluginList.json`
You have to copy "updater" folder (which contains GUP.exe) to enable plugin list.
3. Download DSpellCheck plugin v1.4.22 which is not compatible with Notepad++ from v8.3 and later version here:
https://github.com/Predelnik/DSpellCheck/releases/download/v1.4.22/DSpellCheck_x64.zip
Install it into plugin folder manually.
4. Now launch Notepad++ x64 debug binary and watch it crash.
5. Edit "nppPluginList.json" by adding `"old-versions-compatibility": "[,1.4.22][,8.2.1]",` into DSpellCheck section.
6. Startup Notepad++. Notepad++ doesn't crash. and now you can check DSpellCheck plugin is not loaded on the startup.

Fix  #11358, fix #11349, close  #11356

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -355,10 +355,19 @@ bool PluginsManager::loadPluginsV2(const TCHAR* dir, const PluginViewList* plugi
 						// Find plugin version
 						Version v;
 						v.setVersionFrom(pluginsFullPathFilter);
-						if (pui->_version == v)
+						if (v == pui->_version)
 						{
 							// Find compatible Notepad++ versions
 							isCompatible = nppVer.isCompatibleTo(pui->_nppCompatibleVersions.first, pui->_nppCompatibleVersions.second);
+						}
+						else if (v < pui->_version && // If dll version is older, and _oldVersionCompatibility is valid (not empty), we search in "_oldVersionCompatibility"
+							!(pui->_oldVersionCompatibility.first.first.empty() && pui->_oldVersionCompatibility.first.second.empty()) && // first version interval is valid
+							!(pui->_oldVersionCompatibility.first.second.empty() && pui->_oldVersionCompatibility.second.second.empty())) // second version interval is valid
+						{
+							if (v.isCompatibleTo(pui->_oldVersionCompatibility.first.first, pui->_oldVersionCompatibility.first.second)) // dll older version found
+							{
+								isCompatible = nppVer.isCompatibleTo(pui->_oldVersionCompatibility.second.first, pui->_oldVersionCompatibility.second.second);
+							}
 						}
 					}
 				}
@@ -404,10 +413,19 @@ bool PluginsManager::loadPluginsV2(const TCHAR* dir, const PluginViewList* plugi
 							// Find plugin version
 							Version v2;
 							v2.setVersionFrom(pluginsFullPathFilter2);
-							if (pui2->_version == v2)
+							if (v2 == pui2->_version)
 							{
 								// Find compatible Notepad++ versions
 								isCompatible2 = nppVer.isCompatibleTo(pui2->_nppCompatibleVersions.first, pui2->_nppCompatibleVersions.second);
+							}
+							else if (v2 < pui2->_version && // If dll version is older, and _oldVersionCompatibility is valid (not empty), we search in "_oldVersionCompatibility"
+								!(pui2->_oldVersionCompatibility.first.first.empty() && pui2->_oldVersionCompatibility.first.second.empty()) && // first version interval is valid
+								!(pui2->_oldVersionCompatibility.first.second.empty() && pui2->_oldVersionCompatibility.second.second.empty())) // second version interval is valid
+							{
+								if (v2.isCompatibleTo(pui2->_oldVersionCompatibility.first.first, pui2->_oldVersionCompatibility.first.second)) // dll older version found
+								{
+									isCompatible2 = nppVer.isCompatibleTo(pui2->_oldVersionCompatibility.second.first, pui2->_oldVersionCompatibility.second.second);
+								}
 							}
 						}
 					}
--- a/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
+++ b/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
@@ -558,6 +558,27 @@ std::pair<Version, Version> getIntervalVersions(generic_string intervalVerStr)
 	return result;
 }
 
+// twoIntervalVerStr format:
+// "[4.2,6.6.6][6.4,8.9]"  : The 1st interval from version 4.2 to 6.6.6 inclusive, the 2nd interval from version 6.4 to 8.9
+// "[8.3,][6.9,6.9]"       : The 1st interval any version from 8.3 to the latest version, the 2nd interval present only version 6.9
+// "[,8.2.1][4.4,]"        : The 1st interval 8.2.1 and any previous version, , the 2nd interval any version from 4.4 to the latest version
+std::pair<std::pair<Version, Version>, std::pair<Version, Version>> getTwoIntervalVersions(generic_string twoIntervalVerStr)
+{
+	std::pair<std::pair<Version, Version>, std::pair<Version, Version>> r;
+	generic_string sep = TEXT("][");
+	generic_string::size_type pos = twoIntervalVerStr.find(sep, 0);
+	if (pos == string::npos)
+		return r;
+
+	generic_string intervalStr1 = twoIntervalVerStr.substr(0, pos + 1);
+	generic_string intervalStr2 = twoIntervalVerStr.substr(pos + 1, twoIntervalVerStr.length() - pos + 1);
+
+	r.first = getIntervalVersions(intervalStr1);
+	r.second = getIntervalVersions(intervalStr2);
+
+	return r;
+}
+
 bool loadFromJson(std::vector<PluginUpdateInfo*>& pl, const json& j)
 {
 	if (j.empty())
@@ -602,12 +623,20 @@ bool loadFromJson(std::vector<PluginUpdateInfo*>& pl, const json& j)
 				pi->_nppCompatibleVersions = getIntervalVersions(nppCompatibleVersionStr);
 			}
 
+			if (i.contains("old-versions-compatibility"))
+			{
+				json jOldVerCompatibility = i["old-versions-compatibility"];
+
+				string versionsStr = jOldVerCompatibility.get<std::string>();
+				generic_string oldVerCompatibilityStr(versionsStr.begin(), versionsStr.end());
+				pi->_oldVersionCompatibility = getTwoIntervalVersions(oldVerCompatibilityStr);
+			}
+
 			valStr = i.at("repository").get<std::string>();
 			pi->_repository = wmc.char2wchar(valStr.c_str(), CP_ACP);
 
 			valStr = i.at("homepage").get<std::string>();
 			pi->_homepage = wmc.char2wchar(valStr.c_str(), CP_ACP);
-
 
 			pl.push_back(pi);
 		}
--- a/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.h
+++ b/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.h
@@ -40,6 +40,13 @@ struct PluginUpdateInfo
 	                                                    // <4.2, 6.6.6>: from v4.2 (included) to v6.6.6 (included)
 	                                                    // <0.0.0.0, 8.2.1> all until v8.2.1 (included)
 	                                                    // <8.3, 0.0.0.0> from v8.3 (included) to all
+
+	// Optional
+	std::pair<std::pair<Version, Version>, std::pair<Version, Version>> _oldVersionCompatibility; // Used only by Plugin Manager to filter plugins while loading plugins
+	                                                                                              // The 1st interval versions are for old plugins' versions
+	                                                                                              // The 2nd interval versions are for Notepad++ versions
+	                                                                                              // which are compatible with the old plugins' versions given in the 1st interval
+	
 	generic_string _homepage;
 	generic_string _sourceUrl;
 	generic_string _description;


rdipardo:
> The first interval versions are for old plugin versions: [,2.6] means from all the versions to v2.6 included.
> The second interval versions are for Notepad++ versions: [,8.2.1] means from all the versions to v8.2.1 included.
> Both interval versions together means: the plugin in question from the 1st version to v2.6 is compatible with Notepad++ v8.2.1 and all Notepad++ previous versions.

But the second interval is just the *complementary set* of the compatible plugins (i.e., the hollow portion of this Venn diagram):



So the outcome of scenario n.<sup>o</sup> 1 will __*always*__ be the same:

1. If the max is older than the current version, __*always*__  say no

2. If the min is older or even with the current version, __*always*__ say yes

The value of "old-versions-compatibility" is redundant and could be automatically generated. I doubt any plugin authors will fill it in, or even know how to. It's only relevant if they maintain back-ported versions in parallel with feature releases.

As for scenario n.<sup>o</sup> 2, a "zero trust" approach would be easier and safer. The mere __*presence*__ of a compatibility restriction should block the library loader until "npp-compatible-versions" is satisfied, no questions asked.

donho:
@rdipardo 
Sorry I don't get your point about the 2 circles illustration.

> The value of "old-versions-compatibility" is redundant and could be automatically generated.

How?

> I doubt any plugin authors will fill it in, or even know how to.

True. But the community member can help to fill it out, at least I can. And it could solve our current situation like #11349.
The best example to the problem is Compare or DSpellCheck plugin:
```json
		{
			"folder-name": "ComparePlugin",
			"display-name": "Compare",
			"version": "2.0.2",
			"npp-compatible-versions": "[8.3.1,]",
			"id": "4151fbc9778047991cf4b900363d846bda5b0d1783e5fed9eb77e4c8253ba315",
			"repository": "https://github.com/pnedev/compare-plugin/releases/download/v2.0.2/ComparePlugin_v2.0.2_X64.zip",
			"description": "Shows the differences between 2 files (side by side).",
			"author": "Ty Landercasper, Jean-Sebastien Leroy, Pavel Nedev",
			"homepage": "https://github.com/pnedev/compare-plugin"
		}
```
Here if user uses Notepad++ v8.3.3 on Compare plugin 2.0.2, there's no problem, compatibility can be checked.
If user uses Notepad++ v8.3.3 on Compare plugin 2.0.1, you don't know if v2.0.1 is compatible with Notepad++ v8.3.3, since you can not use the value of "version" ("2.0.2") to deduct v2.0.1 is the version compatible with [,8.2.1].
  
"old-versions-compatibility" (maybe you can find a better name) is the tracker of the old version of plugins which have had critical issue with a specific version of Notepad++. Plugin Loader gets the plugins version from the dll binary, compare with the value of "version" in plugin list, if it's not matched, it checks "old-versions-compatibility" to determinate if the binary version is in the versions which have had issue with Notepad++, if yes then it will compare Notepad++ binary version with the interval versions of Notepad++, otherwise,  the old version of plugin is considered no issue, and it'll be loaded.

I believe Notepad++ with plugin stability issue can be improved in this way, and it's not  redundant.

Yaron10:
@donho,

I think it's a good idea. Thank you for your recent work on this issue. 👍 

If you think it should be tested, could you please provide a plugin that is NOT supposed to be loaded?

rdipardo:
@donho,
>> The value of "old-versions-compatibility" is redundant and could be automatically generated.

> How?

Because the currently running N++ version is the only __*useful*__ point of reference. `[2.6,][,8.2.1]` is not useful information, unless N++ becomes able to downgrade itself.

The only action that should be taken is upgrading the plugin to the listed compatible version (2.7, if it's mimeTools). You already have that information in the "version" and  "npp-compatible-versions" fields.  

You're scenario would ultimately upgrade the plugin anyway; so just upgrade it without complicating the plugin list schema!

donho:
@rdipardo 

> Your scenario would ultimately upgrade the plugin anyway; so just upgrade it without complicating the plugin list schema!

There may be one thing you're missing:
The scenarios I provided is not for Plugin Admin updating, but for plugin loader not loading the incompatible plugins on Notepad++ startup. Please check https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f9118dd13c637a4054c101e685c9030115f790c8.

donho:
@Yaron10 
Thank you. 
You should 
1.  download x64 debug binary here:
https://ci.appveyor.com/api/buildjobs/5ba501ecu21k28i7/artifacts/Notepad%2B%2B.x64.Debug.exe
2. Create an empty file "nppPluginList.json" in your `npp\plugins\Config\`
and copy the following content 
https://github.com/notepad-plus-plus/nppPluginList/blob/5e23fce86813ae434dd123bb33b1dcd7db346050/src/pl.x64.json
and paste into `npp\plugins\Config\nppPluginList.json`
You have to copy "updater" folder (which contains GUP.exe) to enable plugin list.
3. Download DSpellCheck plugin v1.4.22 which is not compatible with Notepad++ from v8.3 and later version here:
https://github.com/Predelnik/DSpellCheck/releases/download/v1.4.22/DSpellCheck_x64.zip
Install it into plugin folder manually.
4. Now launch Notepad++ x64 debug binary and look it crash.
5. Edit "nppPluginList.json" by adding `"old-versions-compatibility": "[,1.4.22][,8.2.1]",` into DSpellCheck section.
6. Startup Notepad++. It shouldn't crash. and now you can check DSpellCheck plugin is not loaded on the startup.

Let me know if you have any problem/question.

@rdipardo You can also try it to see the purpose of  "old-versions-compatibility".

rdipardo:
It does exactly what you expect:

assert

I edit `nppPluginList.json`, _et voilà!_

Screenshot 2022-03-09 033914

rdipardo:
Does this mean the proper functioning of N++ now depends on a JSON file having just the right content?

Yaron10:
@donho,

Thank you for the detailed instructions.
Great work. 👍 

How about a short message?
```
Some plugins are not compatible with the current Notepad++ version and have been disabled.
Newer and compatible versions of those plugins might be available in "Plugins -> Plugins Admin".
```

For those who use the portable version and haven't updated the required components, some warning can be displayed the first time v8.3.3 is run.

donho:
@rdipardo 
> It does exactly what you expect:

Indeed, the behaviour is different between the debug and release binary, DSpellCheck triggers the assertion box you saw in Notepad++ debug. DSpellCheck makes Notepad++ binary crash if the binary is in release mode. 

Here the feature of new added "old-versions-compatibility" is demonstrated: When Notepad++ startup, all the installed plugins will be loaded usually. Thanks to "old-versions-compatibility" of the section of DSpellCheck in "nppPluginList.json", Notepad++ v8.3.2 skips to load DSpellCheck v1.4.22 because of compatibility issue.

> Does this mean the proper functioning of N++ now depends on a JSON file having just the right content?

Yes and no.
If nppPluginList.dll is present and it's enabled (with GUP.exe), then yes - Notepad++ can skip loading incompatible plugins. And of course, the json content in nppPluginList.dll should be correct. In the case of absence of nppPluginList.dll, Notepad++ have no info for filtering, therefore all the plugins will be loaded and lead to eventual crash.

donho:
@Yaron10 
Thank you for your tests and help!

> How about a short message?

Well it'll be disturbing for users for each startup of Notepad++.
If users notice the plugins they use are not loaded, they will ask support from the community or on Github.

donho:
@Yaron10 
> How about a short message?

One point is considerable in such situation:
As @rdipardo suggested in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11358#issuecomment-1062692927:
Plugin Loader skips loading the incompatible plugins on startup, and Plugin Admin can update the incompatible plugin to the newest one (if any) on Notepad++ exit (closed by users). But it needs a lot of coding effort so such feature won't be in upcoming release.

Yaron10:
@donho,

> Well it'll be disturbing for users for each startup of Notepad++.

Good point. But it can be displayed once with a msgShown flag.

> If users notice the plugins they use are not loaded, they will ask support from the community or on Github.

That's a solution too. :)

> As @rdipardo suggested in...

Yes, that would be great.

Thanks again.

rdipardo:
For the record, here's easy this could have been (__*Note*__ I reverted a84cbe8f before applying this):

```diff
diff --git a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
index 9ddd15cd..49f6a254 100644
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -401,10 +401,25 @@ bool PluginsManager::loadPluginsV2(const TCHAR* dir, const PluginViewList* plugi
 						PluginUpdateInfo* pui2 = pluginUpdateInfoList->findPluginInfoFromFolderName(foundFileName2, index2);
 						if (pui2)
 						{
+							// Does plugin have a version constraint?
+							const bool hasConstraint = !(pui2->_nppCompatibleVersions.first.empty() && pui2->_nppCompatibleVersions.second.empty());
+
 							// Find plugin version
 							Version v2;
 							v2.setVersionFrom(pluginsFullPathFilter2);
-							if (pui2->_version == v2)
+							TCHAR msg_buf[1024] = { 0 };
+							const TCHAR* plName = pui2->_displayName.c_str();
+
+							if (hasConstraint && pui2->_version != v2)
+							{
+								wsprintf(msg_buf, L"%s %s may be incompatible with your Notepad++ version.\r\nPlease upgrade to %s and restart Notepad++.", 
+										plName, 
+										v2.toString().c_str(), 
+										pui2->_version.toString().c_str());
+								::MessageBox(NULL, msg_buf, TEXT("Unsafe plugin"), MB_OK);
+								isCompatible2 = false;
+							}
+							else
 							{
 								// Find compatible Notepad++ versions
 								isCompatible2 = nppVer.isCompatibleTo(pui2->_nppCompatibleVersions.first, pui2->_nppCompatibleVersions.second);

```
Screenshot 2022-03-09 224401

[notepad++.x64.Debug-53cbe1c2.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/8220074/notepad%2B%2B.x64.Debug-53cbe1c2.zip)

[notepad++.x64.Debug-53cbe1c2.sha256.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/8220168/notepad%2B%2B.x64.Debug-53cbe1c2.sha256.txt)

donho:
@rdipardo 
You can "guess" 1.4.22 could be incompatible. It works in this case.
But when DSpellCheck release new version v1.4.24, the plugin list will be updated to:
```
		{
			"folder-name": "DSpellCheck",
			"display-name": "DSpellCheck",
			"version": "1.4.24",
			"npp-compatible-versions": "[8.3.1,]",
		},
``` 
Then with your code, if installed DSpellCheck version is v1.4.23, it will suggest to update v1.4.23 to v1.4.24, and not load this compatible plugin at all.

Hence "old-versions-compatibility".

rdipardo:
Only future N++ releases we ever perform these checks. If 1.4.24 is available by then, what's wrong with a upgrading to it?

A plugin downgrade is not possible, because the only maximum on [the list] today is 8.2.1, and 8.2.1 will never execute this code.

After the upgrade, the `else` branch will validate the required version. If that fails, then the plugin list version is newer than the N++ release, which is never going to happen either.


[the list]: https://github.com/notepad-plus-plus/nppPluginList/blob/5ff20d20c0156c05cb870f1a609b8e0b571244fd/src/pl.x64.json#L265

rdipardo:
No one's talking about a package manger here. 
This is just a safety check against (future) breaking changes.

donho:
@rdipardo 
Let's focus on my question:

the context:
The installed DSpellCheck version is v1.4.23, 
the plugin list for DspellCheck :
```
		{
			"folder-name": "DSpellCheck",
			"display-name": "DSpellCheck",
			"version": "1.4.24",
			"npp-compatible-versions": "[8.3.1,]",
			...
		},
```
What does it do your code when Notepad++ startup?

rdipardo:
How does `"version": "1.4.24"` get onto the list unless N++ has been __*upgraded*__, or the user edited the manifest, which would be their own fault? 

Breakage happens because plugins fall behind N++'s development cycle. Upgrading to the next version will always be appropriate, unless you plan to back-port this feature to _N_ past releases. Never has a user been told "Downgrading that plugin will solve the crash you're experiencing."

rdipardo:
A more complicated schema makes nothing safer; it only shifts responsibly to the plugin maintainers.
That is truly the last word I will waste on this topic.

donho:
> How does "version": "1.4.24" get onto the list unless N++ has been upgraded, or the user edited the manifest, which would be their own fault?

It's Notepad++ newest version, let's say v8.3.4, come with this new plugin list above. It's a normal situation. 

> Breakage happens because plugins fall behind N++'s development cycle. Upgrading to the next version will always be appropriate, unless you plan to back-port this feature to N past releases. Never has a user been told "Downgrading that plugin will solve the crash you're experiencing."

I agree about updating plugins automatically would be a better solution. But it needs a lot of coding effort in Plugin Admin, so we are not there yet. What we have now is not loading the incompatible plugins.

> A more complicated schema makes nothing safer; it only shifts responsibly to the plugin maintainers.
That is truly the last word I will waste on this topic.

I have demonstrate that it's not a complicated situation by showing to you 1 + 1 = 2, so I've got nothing more to say neither.

PR #11403, opened 2022-03-17
Fix file saving critical bug under Symantec encryption desktop

### LINKED ISSUE #11339: Notepad ++ is not working with encryption software when saved with newer versions
### Description of the Issue
We have a file location where we store out ITAR files. This file share location is 100% encrypted due to the ITAR files that are in there. We are using Symantec Desktop Encryption for the encrypted files. When we save files with a newer version of Notepad ++ we have started to see issues where files reopening as NULL and we can not see the info in the files. When saved in older versions we are not seeing the issues. If we try and switch the association of the files to just plan old Notepad it opens just fine not matter what version of Notepad ++ we use.

Files save fine with Notepad++ (v7.8.2) but if we use other installs both 8.2.1 and 8.3 they didn’t work. Issue we are seeing is that we are constantly updating users to the newer version of Notepad ++ with Ivanti our update software deployment software. Curious if anyone has seen this issue and has a fix or a place to look for additional trouble shooting.

1.  user goes and tries to open a Notepad++ file in the file share and text disappears and it just says NULL in there
2.  if user changes the file association or rolls back to v7.8.2 Notepad ++ it works file but the moment we update them text goes to NULL again 
3. So it has something to do with a change that was made from v7.8.2 to 8.2.1 or 8.3

### Expected Behavior
When we open up the file we want to keep the text with in the file and not say NULL in it. It works just fine on the older version - we rolled it back to version 7 and it works and with the versions 8.2.1 we uninstalled plugins and still see the issues

### Debug Information
Notepad++ v8.3.2 (64-bit)
Path : C:\ProgramData\Notepad++\notepad++.exe
Command Line :
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19044.1526
Current ANSI codepage : 1252

took out plugins and did a debug again

Notepad++ v8.3.2 (64-bit)
Path : C:\ProgramData\Notepad++\notepad++.exe
Command Line :
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19044.1526
Current ANSI codepage : 1252
Plugins : none

Not sure that bug report is exactly what you are looking for - let me know of additional information you would like me to provide im sure more details will be needed for additional trouble shooting

donho:
Fix #11339

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -127,7 +127,7 @@ void writeLog(const TCHAR *logFileName, const char *log2write)
 	const DWORD accessParam{ GENERIC_READ | GENERIC_WRITE };
 	const DWORD shareParam{ FILE_SHARE_READ | FILE_SHARE_WRITE };
 	const DWORD dispParam{ OPEN_ALWAYS }; // Open existing file for writing without destroying it or create new
-	const DWORD attribParam{ FILE_ATTRIBUTE_NORMAL | FILE_FLAG_WRITE_THROUGH };
+	const DWORD attribParam{ FILE_ATTRIBUTE_NORMAL };
 	HANDLE hFile = ::CreateFileW(logFileName, accessParam, shareParam, NULL, dispParam, attribParam, NULL);
 
 	if (hFile != INVALID_HANDLE_VALUE)
@@ -149,6 +149,7 @@ void writeLog(const TCHAR *logFileName, const char *log2write)
 		::WriteFile(hFile, log2writeStr.c_str(), static_cast<DWORD>(log2writeStr.length()), &bytes_written, NULL);
 
 		::FlushFileBuffers(hFile);
+		::CloseHandle(hFile);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


xomx:
@donho 

Good file handle leak catch in the `writeLog` func!

And all your changes are ok, at least I think so.

Maybe consider adding of the FlushFileBuffers possible error states into the NppLogNulContentCorruptionIssue log, e.g.:

```
void Win32_IO_File::close()
{
	if (isOpened())
	{
		DWORD dwFlushError = NOERROR;
		if (_written)
		{
			if (!::FlushFileBuffers(_hFile))
				dwFlushError = ::GetLastError();
		}
		::CloseHandle(_hFile);
		
		...
```
and log the `dwFlushError` further in the code, if `dwFlushError != NOERROR`...

donho:
@xomx 
The code you suggest: `if (!::FlushFileBuffers(_hFile))`
However:
`If the function succeeds, the return value is nonzero.`
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers

Issue #11359, opened 2022-03-09
Problem with Toolbar


VFederau:
### Description of the Issue
After Switching in Option to "Dark Mode" he changes the icon at the toolbar. When you disable later the "Dark Mode" he switches everything to normal - except for the Toolbar Icons.

### Steps to Reproduce the Issue
1. Click to Options
2. Enable the "Dark Mode" 
3. Disable the "Dark Mode"

### Expected Behavior
After deactive the "Dark Mode" i would see the normal icons - not the "Dark Mode" Icons any more.

### Actual Behavior
After Switching in Option to "Dark Mode" he changes the icon at the toolbar. When you disable later the "Dark Mode" he switches everything to normal - except for the Toolbar Icons.

### Debug Information
Notepad++ v8.2.1   (32-bit)
Build time : Jan 19 2022 - 18:38:49
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2009
OS Build : 19042.1526
Current ANSI codepage : 1252
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll 

![BugNotepadPlusPlus](https://user-images.githubusercontent.com/10497094/157427709-d7ef1d0d-8ef4-4473-b28d-a812259598e1.PNG)

VFederau:
Found a workaround - he Switches in "Dark Mode" the Toolbar to "Fluent UI: little symbols". Here in the Menu (unter Options the first point in the list) you can change it back to "little default symbols" (last entry in list).
![NppWorkaround](https://user-images.githubusercontent.com/10497094/157432195-ae19a60c-3ac8-44fa-a7d2-32c339dc1eb6.png)

ArkadiuszMichalski:
Duplicate of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10544.

Issue #10544, opened 2021-09-13
Disabling dark mode in preferences does not set back icon type used in light mode. When starting in light mode with "Standard icons : small"


GryXion:
### Description of the Issue
Enabling dark mode reset toolbar ui, but Disabling dark mode does not make it back to previous one for standards icons specificaly.

Hi, in notepad ++ since version 8.1.0 up to 8.1.4, in WINDOWS version 32bits and 64bits alike.

Bug tested on 8.1 up to 8.1.4 on WINDOWS 7 and WINDOWS 10 lastest update, in 32bits and 64bits version not portable, installed with installer .msi for windows.

### Steps to Reproduce the Issue
1.  open notepad++
2.  click on toolbar => "Settings"
3.  in "General" check Standard icons : small
4. then on "Dark Mode" menu in settings panel still, click on checkbox "Enable dark mode"
5. The unexpected/ might be wanted and not bothering me happen : in General, Toolbar panel is set to "Fluent UI: small"
6. The really unexpected thing is, going back to "Dark Mode" in settings, Disabling darkmode with a click on checkbox "Enable dark mode"
7. In General panel in settings, Toolbar panel is still set in "Fluent UI: small" or whatever was set in dark mode enabled, but do not revert to the setting that was set in light mode.

### Expected Behavior
1. I expected that disabling dark mode would put back on the previous settings that was present in light mode, before dark mode was enable.
In my case, it would be to show up "Standard icons : small again" that was up in light mode when dark mode is disable.
2. _And if possible un option to keep the current Tool bar icon size and type when darkmode is enable/disable. Or a way to keep light mode settings in light mode, and dark mode settings in dark mode.

i think there is a need to think about it to not impact wrongly other people expectation, as such, maybe a setting, or something to keep the current settings of light mode, and a way to get back to light mode settings would be great. I did not figure it out the best way yet, thx for your help.

### Actual Behavior
"Standard icons : small again" is not reset on when dark mode is disable, while it was on before dark mode beeing enable.

### Debug Information
Bug is present from v8.1.0 to v8.1.4 checked in WIN10 and WIN7, in both 64 and 32 bits, and both not portable version. unknown on the other versions

exemple :
Notepad++ v8.1.4   (32-bit)
Build time : Aug 21 2021 - 12:58:42
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
[...]

Issue #10212, opened 2021-07-20
Enable dark mode switches the Toolbar Icons


DMI-1407:
If "Standard Icons: small" is selected and then Dark Mode gets enabled the Toolbar Icons are set to "Fluent UI: small" and "Fluent UI: small" is selected.
_(This is really annoying...)_

### Steps to Reproduce the Issue
1. Disable:  Preferences -> Dark Mode -> Enable Dark Mode
2. Select:  Preferences -> General -> Toolbar -> Standard icons: small
3. Enable:  Preferences -> Dark Mode -> Enable Dark Mode

### Expected Behavior
No change in:  Preferences -> General -> Toolbar

### Actual Behavior
Sets option to:  Preferences -> General -> Toolbar -> Fluent UI: small

### Debug Information
Notepad++ v8.1.2   (32-bit)
Build time : Jul 16 2021 - 01:56:21
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : none

DMI-1407:
label:bug

ArkadiuszMichalski:
Any reason why after turning off Dark Mode the previous icon pack is not selected? I usually use `Standard Icons: small` and I have to manually select them again after turning off Dark Mode, it's annoying.

donho:
For simplifying UX, once user switch on/off dark mode, a lot of settings will be change.
But in v8.1.3, once you stay in dark mode, your settings will be kept while you're changing the ton or customizing your dark mode.
Such behaviour will be remained.

PR #11405, opened 2022-03-18
#11359 Switch the toolbar back to standard view when disabling the da…

ViktorGraz:
…rk mode
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11359, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10544, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10212


PowerEditor/src/WinControls/Preference/preference_rc.h
```diff
@@ -20,6 +20,7 @@
 	#define	PREF_MSG_ISCHECKED_GENERALPAGE (WM_USER + 1) // wParam:checkbox/radiobutton ID in General page. lParam is type of "bool *" to get result
 	#define	PREF_MSG_SETTOOLICONSFROMSTDTOSMALL (WM_USER + 2) 
 	#define	PREF_MSG_DISABLETABBARALTERNATEICONS (WM_USER + 3) 
+	#define PREF_MSG_SETTOOLICONSFROMSMALLTOSTD (WM_USER + 4)
```
Mateos81, line 23: Small remark: the previous entries have a tab between the #define keyword and the constant name, while yours has a space in-between

ViktorGraz: @Mateos81, thanks for the review, I have changed it!

ArkadiuszMichalski:
@ViktorGraz You always back to `Standard icon:small`, even if someone have choosed other sets before switch `Dark mode`. This is probably not the expected behavior. Also after turn off `Dark mode` we have this state in `Preferences`:
image

ViktorGraz:
@ArkadiuszMichalski , thanks for the review!

PR #11411, opened 2022-03-19
Window Sort Document List Fix

### LINKED ISSUE #11272: Window -> Sort BY: Update Document List
**STR:**
Open some files.
Open the _Document List_ panel. 
Change the documents position via _Window -> Sort By_.

**Result:**
The order in _Document List_ is not updated.
Sorting via the Window dialog, the _Document List_ **is** updated..

***
Using the latest https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2d1ba019996d3d599abb46c446bd4ad299672ce9 in Master.

Ashfaaq18:
Fix #11272
Document List now updates itself with respect to the tabs sorted by the windows dialog
WinSortDocList_Fix_GIF

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2880,8 +2880,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 						}
 						activateBuffer(_pDocTab->getBufferByIndex(_pDocTab->getCurrentTabIndex()), currentView());
 
-						if (_pDocumentListPanel)
-							_pDocumentListPanel->reload();
+						::SendMessage(_pDocTab->getHParent(), NPPM_INTERNAL_DOCORDERCHANGED, 0, _pDocTab->getCurrentTabIndex());
 
 						break;
 					}
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2879,6 +2879,10 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 							_pDocTab->setBuffer(i, tempBufs[nmdlg->Items[i]]);
 						}
 						activateBuffer(_pDocTab->getBufferByIndex(_pDocTab->getCurrentTabIndex()), currentView());
+
+						if (_pDocumentListPanel)
+							_pDocumentListPanel->reload();
+
 						break;
 					}
 				}


Yaron10:
@Ashfaaq18,

Thank you for fixing yet another of "my issues". 👍 

I've used:
```
// Update _pDocumentListPanel and notify plugins.
// It would be much better to send it only after verifying the order has indeed been changed.
::SendMessage(hwnd, NPPM_INTERNAL_DOCORDERCHANGED, 0, nmdlg->curSel);
```
What do you think?

***

**NOTE:** regarding `nmdlg->curSel` - I've modified some code in WindowsDlg.cpp.
```
void WindowSort::doSort()
{
	size_t count = _pTab->nbItem();
	std::vector<UINT> items(count);
	NMWINDLG nmdlg = {};
	nmdlg.nItems = static_cast<UINT>(count);
	nmdlg.Items = items.data();

	for (size_t i = 0; i < count; ++i)
		nmdlg.Items[i] = _idxMap[i];

	std::vector<int>::iterator it = std::find(_idxMap.begin(), _idxMap.end(), _pTab->getCurrentTabIndex());
	nmdlg.curSel = static_cast<UINT>(it - _idxMap.begin());

	::SendMessage(_pTab->getHParent(), WDN_NOTIFY, 0, LPARAM(&nmdlg));
}
```

Ashfaaq18:
> 
> I've used:
> 
> ```
> // Update _pDocumentListPanel and notify plugins.
> // It would be much better to send it only after verifying the order has indeed been changed.
> ::SendMessage(hwnd, NPPM_INTERNAL_DOCORDERCHANGED, 0, nmdlg->curSel);
> ```
> 
> What do you think?
> 

May i know inside which function you are using this statement? And yeah, its better to run it only when a sort actually happens and not for every sort click. 
My one is running for every click at the moment. I will have to amend this PR.

> **NOTE:** regarding `nmdlg->curSel` - I've modified some code in WindowsDlg.cpp.
> 
> ```
> void WindowSort::doSort()
> {
> 	size_t count = _pTab->nbItem();
> 	std::vector<UINT> items(count);
> 	NMWINDLG nmdlg = {};
> 	nmdlg.nItems = static_cast<UINT>(count);
> 	nmdlg.Items = items.data();
> 
> 	for (size_t i = 0; i < count; ++i)
> 		nmdlg.Items[i] = _idxMap[i];
> 
> 	std::vector<int>::iterator it = std::find(_idxMap.begin(), _idxMap.end(), _pTab->getCurrentTabIndex());
> 	nmdlg.curSel = static_cast<UINT>(it - _idxMap.begin());
> 
> 	::SendMessage(_pTab->getHParent(), WDN_NOTIFY, 0, LPARAM(&nmdlg));
> }
> ```

Im assuming the above is actually the WindowDlg::doSort() function and not WindowSort::doSort(), if yes, that's a significant code modification. I haven't yet gone through it fully to differentiate the functionality difference.

Yaron10:
@Ashfaaq18,

Thank you for replying.

> May i know inside which function you are using this statement?

After https://github.com/notepad-plus-plus/notepad-plus-plus/blob/d9c5688635ae78ad04b739fb59bb7868f1903890/PowerEditor/src/NppBigSwitch.cpp#L2881

(Instead of your `if (_pDocumentListPanel)`).

> Im assuming the above is actually the [WindowDlg::doSort()](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/d9c5688635ae78ad04b739fb59bb7868f1903890/PowerEditor/src/WinControls/WindowsDlg/WindowsDlg.cpp#L949-L985) function...

I've refactored WindowsDlg.cpp. I just wanted to explain `nmdlg->curSel`.
If you think that `:SendMessage()` is better, you could use `::SendMessage(hwnd, NPPM_INTERNAL_DOCORDERCHANGED, 0, _pDocTab->getCurrentTabIndex());`

This is also called even if the order has not been changed. It was a note to myself.

Yaron10:
@Ashfaaq18,

I just want to clarify that my comment/question was not about checking if the order has been changed. I'm not sure it's worth the extra code.

`NPPM_INTERNAL_DOCORDERCHANGED` is used in `IDM_VIEW_TAB_MOVEFORWARD` and Drag&Drop. I thought it might be better to use it here too. 

Thanks again.

Ashfaaq18:
> @Ashfaaq18,
> 
> I just want to clarify that my comment/question was not about checking if the order has been changed. I'm not sure it's worth the extra code.

@Yaron10, Got it.

> `NPPM_INTERNAL_DOCORDERCHANGED` is used in `IDM_VIEW_TAB_MOVEFORWARD` and Drag&Drop. I thought it might be better to use it here too.
> 
> Thanks again.

Actually i prefer your method since it also notifies the plugins. Btw, do you know of any plugins which uses this info?

On another note i found a small issue with both these methods,
So i tested the NPPM_INTERNAL_DOCORDERCHANGED case and came across a scenario where this fix of mine and the NPPM_INTERNAL_DOCORDERCHANGED case wont work. 

Steps to reproduce:
1. Enable Document List
2. Sort the Document List Column to either ascending or descending
3. Now go to the windows dialog or windows->sort by and sort
4. This fix and the NPPM_INTERNAL_DOCORDERCHANGED case will not work

In this case I'm not sure if we should override the sorted Document List columns or just leave them as it is.

Yaron10:
@Ashfaaq18,

> Btw, do you know of any plugins which uses this info?

No, unfortunately I don't.

> In this case I'm not sure if we should override the sorted Document List columns or just leave them as it is.

IMO we should leave the order in the Document List as it is.
This is the result on manually Drag&Drop, and I think it's the expected behavior.

Thanks again for your work. 👍 

***

BTW, there's a major difference between the Dialog and the Menu.

Sorting via the Dialog, the selected tab may change: the previously **tab-index** gets selected.
For example: if you have **A (selected)** on the left and B on the right, and after sorting the order changes - **B is selected**.

Sorting via the Menu, the previously selected tab remains selected. I think that's better.

I realized that yesterday after retesting whether the Document List is updated on sorting via the Dialog (without your PR).

Ashfaaq18:
> 
> Thanks again for your work. 👍
> 

Your welcome.

> BTW, there's a major difference between the Dialog and the Menu.
> 
> Sorting via the Dialog, the selected tab may change: the previously **tab-index** gets selected. For example: if you have **A (selected)** on the left and B on the right, and after sorting the order changes - **B is selected**.
> 
> Sorting via the Menu, the previously selected tab remains selected. I think that's better.
> 
> I realized that yesterday after retesting whether the Document List is updated on sorting via the Dialog (without your PR).

Would be great if you could open a new issue to solve this case.

Yaron10:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11422.

Thank you. 👍

Issue #10504, opened 2021-09-04
Notes for updating N++'s Scintilla beyond 5.x


ArkadiuszMichalski:
Notepad++ 8.1.4 (current version as of this writing) uses Scintilla version 4.4.6. Starting from Scintilla 5.0.0 we have:
>First version that separates Lexilla from Scintilla. Each of the 3 projects now has a separate history page but history before 5.0.0 remains combined.

https://github.com/ScintillaOrg/lexilla
https://www.scintilla.org/Lexilla.html
https://www.scintilla.org/LexillaDoc.html
https://www.scintilla.org/ScintillaDoc.html#alpha

Issues fixed after the update:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10925
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10651
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10908
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10706
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10065
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10423
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3858
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9229
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8359
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8087
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/63
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5264
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5805, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5750

Issues that can be fixed after the update:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11413
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10171
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10991
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11014
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10301
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1260

Even some early version (Notepad++ with new Scintilla/Lexilla) for testing would be advisable to catch possible problems.

[...]

PR #11419, opened 2022-03-21
Scintilla5

chcg:
PR as requested by #10504

@donho Should I squash the changes into a single commit?

Open issues:
- support for external lexers seems to need a change to the existing lexers (tested with https://github.com/blu3mania/npp-papyrus)
- objc lexer change, unclear why it is needed
- hardcoded, spreaded lexer names


PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -65,7 +67,7 @@ LanguageName ScintillaEditView::langNames[L_EXTERNAL + 1] = {
 {TEXT("c"),				TEXT("C"),					TEXT("C source file"),									L_C,			SCLEX_CPP},
 {TEXT("cpp"),			TEXT("C++"),				TEXT("C++ source file"),								L_CPP,			SCLEX_CPP},
 {TEXT("cs"),			TEXT("C#"),					TEXT("C# source file"),									L_CS,			SCLEX_CPP},
-{TEXT("objc"),			TEXT("Objective-C"),		TEXT("Objective-C source file"),						L_OBJC,			SCLEX_CPP},
+{TEXT("objc"),			TEXT("Objective-C"),		TEXT("Objective-C source file"),						L_OBJC,			SCLEX_OBJC},
```
chcg, line 70: Unclear why this change is needed to get objc lexer working

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -630,7 +632,7 @@ void ScintillaEditView::setXmlLexer(LangType type)
 {
 	if (type == L_XML)
 	{
-        execute(SCI_SETLEXER, SCLEX_XML);
+		execute(SCI_SETILEXER, 0, reinterpret_cast<LPARAM>(CreateLexer("xml")));
```
chcg, line 635: "xml" and strings below might be replaced again by some definition/constexpr from N++

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -1215,7 +1217,12 @@ void ScintillaEditView::setKeywords(LangType langType, const char *keywords, int
 
 void ScintillaEditView::setLexer(int lexerID, LangType langType, int whichList)
 {
-	execute(SCI_SETLEXER, lexerID);
+#pragma warning( push )
```
chcg, line 1220: considered as workaround

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2242,7 +2249,7 @@ char * ScintillaEditView::getWordFromRange(char * txt, size_t size, size_t pos1,
         pos2 = tmp;
     }
 
-    if (size < pos2 - pos1)
+    if (size < pos2-pos1)
```
chcg, line 2252: unneeded change

PowerEditor/src/ScintillaComponent/ScintillaEditView.h
```diff
@@ -681,16 +681,12 @@ friend class Finder;
 	};
 
 	void setMakefileLexer() {
-		execute(SCI_SETLEXER, SCLEX_MAKEFILE);
-		makeStyle(L_MAKEFILE);
+		setLexer(SCLEX_MAKEFILE, L_MAKEFILE, LIST_NONE);
```
chcg, line 684: needs review, if this could be done, "needed" because of compilation issues if code is placed within header file

PowerEditor/src/ScintillaComponent/ScintillaEditView.h
```diff
@@ -681,16 +681,12 @@ friend class Finder;
...
 
 	void setMakefileLexer() {
-		execute(SCI_SETLEXER, SCLEX_MAKEFILE);
-		makeStyle(L_MAKEFILE);
+		setLexer(SCLEX_MAKEFILE, L_MAKEFILE, LIST_NONE);
 	};
 
 	void setIniLexer() {
-		execute(SCI_SETLEXER, SCLEX_PROPERTIES);
+		setLexer(SCLEX_PROPERTIES, L_INI, LIST_NONE);
```
chcg, line 688: dito

dail8859:
Do we know/understand what kind of plugin compatibility issues this might cause? As we've seen in the past even a small change to the Scintilla ABI can cause numerous plugins to break.

chcg:
I think the relevant cases are (based on the current changes):
- ExternalLexers: Incompatible ILexer5 needs to be implemented
- Plugins using also Scintilla interfaces: might need adaptations depending on the used scintilla functionality
- Plugins using N++ interfaces: considered to be still functional without adaptations

vinsworldcom:
> Do we know/understand what kind of plugin compatibility issues this might cause?

I have by no means tested exhaustively, but am running the artifact from this commit for a bit with all my usual plugins and no noticeable issues yet.

Notepad++ v8.3.3   (64-bit)
Build time : Mar 21 2022 - 07:28:18
Path : C:\Users\mvincent\tmp\npp64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1586
Current ANSI codepage : 1252
Plugins : AutoSave.dll BetterMultiSelection.dll ChangedLines.dll CodeAlignmentNpp.dll ComparePlus.dll CustomizeToolbar.dll DSpellCheck.dll Explorer.dll GitSCM.dll JSMinNPP.dll MarkdownPanel.dll NppExec.dll NppFTP.dll NppMenuSearch.dll PythonScript.dll QuickText.dll SurroundSelection.dll TagLEET.dll TagsView.dll XMLTools.dll ZoomDisabler.dll 

Was able to color the background of autocomplete as a new Scintilla 5+ feature:

image

donho:
@chcg 
> @donho Should I squash the changes into a single commit?

Yes please. 

I created a new branch which is based on your Scintilla5 branch but I modified manually for understanding the difference between old & new Scintilla and for also adding the Notepad++ specific Scintilla modifications:
https://github.com/donho/notepad-plus-plus/commit/b55d3d5ce18a4adac07f26b868a2207fee528777

This PR will be used anyway for being merged to upgrade to Scintilla5 since its binary works much better than the one of my branch.

I will provide you the diff soon between 2 branches so you can integrate the modifications into this PR.

Yaron10:
Thank you for your work.

I've just tested this build for a few minutes.
Should we report issues at this stage?

1. Current line is highlighted if "Highlight Current Line" is _unchecked_ and vice versa.
2. Selected text background color changes when the editor doesn't have focus.

***

Should I be able to build the new source files with Visual Studio 2017?

chcg:
@donho Squash done.

donho:
@Yaron10 
> Current line is highlighted if "Highlight Current Line" is unchecked and vice versa.

I cannot reproduce it.

@vinsworldcom 
> Was able to color the background of autocomplete as a new Scintilla 5+ feature:

One thing at a time :) It will be done for sure.

@dail8859 
> Do we know/understand what kind of plugin compatibility issues this might cause? As we've seen in the past even a small change to the Scintilla ABI can cause numerous plugins to break.

@chcg has summarized very well about the eventual plugin compatible issues. I just want to add that in the case of crash issue, we have the way to not loading the crash plugin via Plugin list compatibility feature, if the incompatible plugins are discovered before the next release. That guarantees Notepad++'s stability and also gives the time to plugin authors to adapt the plugins to Scintilla5.

Yaron10:
@chcg,

Thank you for the detailed reply. I appreciate it.

@donho,

I've downloaded a fresh portable v8.3.3 and replaced the binary with https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/42993546/job/i961vxwbvsc8e665/artifacts.

תמונה

תמונה

vinsworldcom:
> @Yaron10
> 
> > Current line is highlighted if "Highlight Current Line" is unchecked and vice versa.
> 
> I cannot reproduce it.

@donho I can reproduce it.

1) Download fresh 8.3.3 portable and unzip
2) Download the Scintilla 5 `notepad++.exe` artifact and replace the unzipped portable `notepad++.exe` with the Scintilla 5 artifact.
3) Open this new `notepad++.exe` and observe the line 1 is highlighted
4) Open menu Settings => Preferences... => Editing and observe "Enable current line highlighting" is unchecked

image

Assume it may have something to do with Scintilla now using `SC_ELEMENT_CARET_LINE_BACK` as opposed to [`SCI_SETCARETLINEBACK`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETCARETLINEBACK) the way we do it currently?

Cheers.

ArkadiuszMichalski:
For me this str is valid:

1) Download fresh 8.3.3 portable and unzip
2) Download the Scintilla 5 `notepad++.exe` artifact and replace the unzipped portable `notepad++.exe` with the Scintilla 5 artifact.
3) Open this new `notepad++.exe` and observe the line 1 is highlighted
4) Open menu Settings => Preferences... => Editing and uncheck "Enable current line highlighting"
5) Restart Notepad++ and see that current line is still highlighting

Basically when Notepad ++ is running again then current line is always highlighting (regardless of the settings).

PowerEditor/scintilla.original.forUpdating/scintillaUpdatingWorkFlow.txt
```diff
@@ -1,2 +1,2 @@
-This file (scintilla.original.forUpdating.7z) is the original Scintilla release of current Scintilla.
-While updating a new version of scintilla, this scintilla will be compared with Notepad++'s scitilla, in order to get the patch for adding it into new scintilla release.
+This file (scintilla<version>.zip and lexilla<version>.zip) are the original Scintilla and Lexilla releases of current Scintilla and Lexilla.
+While updating a new version of scintilla or lexilla, the content of the zips will be compared with Notepad++'s scintilla or lexilla, in order to get the patch for adding it into new scintilla or lexilla release.
```
donho, line 2: 👍

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4291,7 +4293,12 @@ void Finder::finishFilesSearch(int count, int searchedCount, bool isMatchLines,
 	addSearchHitCount(count, searchedCount, isMatchLines, searchedEntireNotSelection);
 	_scintView.execute(SCI_SETSEL, 0, 0);
 
-	_scintView.execute(SCI_SETLEXER, SCLEX_SEARCHRESULT);
```
donho, line 4294: So the above code was not necessary at all for Scintilla446?
Or Scintilla5 doesn't need to set it again?

donho (changes requested):
Great job!
Please consider the the revealed points in comments.

ArkadiuszMichalski:
@donho Please follow this str https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1077748440.

donho:
@ArkadiuszMichalski 
> Download fresh 8.3.3 portable and unzip

OK so it seems the first step is important.

donho:
@Yaron10
> Current line is highlighted if "Highlight Current Line" is unchecked and vice versa.

Could you open an issue for it please?
Some behaviours in Scintilla5 have been changed. But the old logic of this feature has been implemented badly.
It's not bug in this PR, so I will fix it after merge this PR.

Yaron10:
@donho,

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11433.
Thank you. :+1:

donho:
Thank you @Yaron10 !

rdipardo:
> * Plugins using also Scintilla interfaces: might need adaptations depending on the used scintilla functionality

Just watch out for SCI_GETTEXT, SCI_GETSELTEXT and SCI_GETCURLINE [^1]; early adopters [got really burned][1] by those.

Sorry, @dail8859, but it looks like [ElasticTabstops][2] may need extra attention:

    $ git grep -niE "SCI_GET(TEXT|SELTEXT|CURLINE),"

    src/ScintillaEditor.h:216:              sptr_t res = Call(SCI_GETCURLINE, length, text);
    src/ScintillaEditor.h:221:              auto size = Call(SCI_GETCURLINE, SCI_UNUSED, NULL);
    src/ScintillaEditor.h:223:              Call(SCI_GETCURLINE, text.length(), &text[0]);
    src/ScintillaEditor.h:1130:             sptr_t res = Call(SCI_GETSELTEXT, SCI_UNUSED, text);
    src/ScintillaEditor.h:1135:             auto size = Call(SCI_GETSELTEXT, SCI_UNUSED, NULL);
    src/ScintillaEditor.h:1137:             Call(SCI_GETSELTEXT, SCI_UNUSED, &text[0]);
    src/ScintillaEditor.h:1242:             sptr_t res = Call(SCI_GETTEXT, length, text);
    src/ScintillaEditor.h:1247:             auto size = Call(SCI_GETTEXT, SCI_UNUSED, NULL);
    src/ScintillaEditor.h:1249:             Call(SCI_GETTEXT, text.length(), &text[0]);


LuaScript seems to be in the clear.

[^1]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10504#issuecomment-1010295809

[1]: https://github.com/geany/geany/commit/142b8ffdd4b7372b77966abbc99435644210515a
[2]: https://github.com/dail8859/ElasticTabstops/issues/28

dail8859:
> LuaScript seems to be in the clear.

Unfortunately not :) It doesn't directly use the `SCI_xxx` constants but does wrap these Scintilla messages.

E.g.
https://github.com/dail8859/LuaScript/blob/066aeb615c08581898e124f216396f3d8aa8c1c8/src/SciIFaceTable.cpp#L2871

chcg:
@donho Added the requested changes. Please check that I didn't miss something.

donho (approved):

donho:
@chcg 
Thank you for the change. The issue that you mentioned: 
> support for external lexers seems to need a change to the existing lexers 

Could you elaborate it please?

donho (approved):
After some googling, it seems "SCI_LOADLEXERLIBRARY" is abandoned by Scintilla5. we have to remove the external lexer support in the next release.

For the '<' & '>' of #define, don't worry about it - I will do it in the same time of removing the code of external lexer.

Yaron10:
@donho,

> Selected text background color changes when the editor doesn't have focus.

Should I open an issue for that?

donho:
@Yaron10 
>> Selected text background color changes when the editor doesn't have focus.
> Should I open an issue for that?

I don't think it's a bug - it's rather the new behaviour of Scintilla5 IMO.
Could you please check SciTE to confirm this new behaviour?

Yaron10:
@donho,

I haven't used SciTE for years. It makes you appreciate Notepad++ even more. :)
I've just downloaded the latest version and the selected text background color changes on losing focus.

I suppose we'll have to get used to it.

Thank you for your work. 👍

Yaron10:
@dail8859,

That's absolutely great! 👍 
Thank you very much.

vinsworldcom:
@Yaron10 

I set it with PythonScript, which doesn't yet know about the new Scintilla API, so a bit clunky:

```
import ctypes

user32 = ctypes.WinDLL('user32', use_last_error=True)
# change background color of inactive selection text (Scintilla 5+)
user32.SendMessageW(editor1.hwnd, 2753, 17, rgb2int(96,96,96))
user32.SendMessageW(editor2.hwnd, 2753, 17, rgb2int(96,96,96))
```

Cheers.

Yaron10:
@vinsworldcom,

Thank you for sharing. 👍 

What do you import for `rgb2int`?

I must leave now. I'll be back on Thursday.

alankilborn:
>What do you import for rgb2int?

Probably something like this:

    def rgb(red, green, blue): return (((blue << 8) | green) << 8) | red

vinsworldcom:
@Yaron10 @alankilborn whoops, forgot about that.  I import it from a utility module I have.  The function (and its inverse):

```
def rgb2int(r, g, b, RGB=False):
    """
        Convert RGB to integer.  Returns value as per COLORREF 0x00BBGGRR:
        https://docs.microsoft.com/en-us/windows/win32/gdi/colorref
        Set RGB arg to True for 0x00RRGGBB value.
    """

    r = int(r)
    g = int(g)
    b = int(b)
    if RGB:
        color = (r << 16) + (g << 8) + b 
    else:
        color = (b << 16) + (g << 8) + r 
    return color

def int2rgb(color):
    """Convert integer to R,G,B tuple"""

    if type(color) == str:
        color = color.split('#')[-1]
        color = int(color, 16)
    r = (color >> 16) & 255
    g = (color >> 8) & 255
    b = color & 255
    return (r,g,b)
```

You could of course just put a number in the last argument of `user32.SendMessageW()` from my previous post, for example:

```
import ctypes

user32 = ctypes.WinDLL('user32', use_last_error=True)
# change background color of inactive selection text (Scintilla 5+)
user32.SendMessageW(editor1.hwnd, 2753, 17, 6316128)
user32.SendMessageW(editor2.hwnd, 2753, 17, 6316128)
```

Cheers.

Yaron10:
@vinsworldcom,

It won't work for me. I'm using the following code in "startup.py":
```
def rgb2int(r, g, b, RGB=False):
    r = int(r)
    g = int(g)
    b = int(b)
    if RGB:
        color = (r << 16) + (g << 8) + b
    else:
        color = (b << 16) + (g << 8) + r
    return color

import ctypes
from Npp import editor

user32 = ctypes.WinDLL('user32', use_last_error=True)

notepad.hwnd = user32.FindWindowW('Notepad++', None)
editor1.hwnd = user32.FindWindowExW(notepad.hwnd, None, "Scintilla", None)
editor2.hwnd = user32.FindWindowExW(notepad.hwnd, editor1.hwnd, "Scintilla", None)

# change background color of inactive selection text (Scintilla 5+)
user32.SendMessageW(editor1.hwnd, 2753, 17, rgb2int(30,30,30))
user32.SendMessageW(editor2.hwnd, 2753, 17, rgb2int(30,30,30))
```

Thanks again.

Thanks also to @Ekopalypse for https://community.notepad-plus-plus.org/post/75217.

***
@alankilborn,

Thank you for the `rgb2int` alternative.

vinsworldcom:
@Yaron10 not sure you need to find the hwnd of the editor objects.  I think PythonScript does that for you.  I don't FindWindow() in my startup.py.

Yaron10:
@vinsworldcom,

תמונה

It won't work from a regular script either.

Yaron10:
@vinsworldcom,

No error. It just makes no difference.
I'm using PT v2.

Thanks again to you and @alankilborn.

Unfortunately I can't use Scintilla5 until https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11476 is fixed.
But it's still interesting. :)

Yaron10:
@vinsworldcom & @alankilborn,

Adding `execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_SELECTION_INACTIVE_BACK , selectColorBack);` after
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/721f994df8c5ea8085a6c9b82a4d6a85110f6f6e/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L2586

does work.

I don't understand why the script doesn't.

donho:
FYI:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3b0d5242acd23b4a727560ea98139cc5f4a8b88a

PR #11432, opened 2022-03-24
Update Inno Setup keyword list

sssousa:


donho:
@sssousa 
I didn't find any info by googling on it: 
Is Inno Setup case-sensitive?

sssousa:
@donho 
It's not.

donho:
@sssousa 
> It's not.

Any ref from the official description? 
And even though, people use Camel case or Pascal case conventionally, don't they?

sssousa:
@donho 
> Any ref from the official description? And even though, people use Camel case or Pascal case conventionally, don't they?

I couldn't find it on the official documentation either, so I tested at my owns scripts to be sure. But you are correct, the common use is Pascal Case.

donho:
@sssousa 
Thank you for your info.
I will take the PR as it is.
If you wanna transform all inno setup autocompletion keywords to Pascal Case afterward, you're welcome.

PR #11442, opened 2022-03-27
NSIS syntax update

sredna:
Instruction list auto-generated directly from NSIS source. This means some of the v1.0 instructions you have in your list (but we (NSIS) removed 20 years ago) are no longer recognized.


PowerEditor/installer/APIs/nsis.xml
```diff
@@ -98,22 +107,27 @@
 		<KeyWord name="GetDLLVersion" />
 		<KeyWord name="GetDLLVersionLocal" />
 		<KeyWord name="GetErrorLevel" />
-		<KeyWord name="GetExeName" />
-		<KeyWord name="GetExePath" />
```
donho, line 102: Why are both APIs removed?

PowerEditor/installer/APIs/nsis.xml
```diff
@@ -203,7 +236,6 @@
 		<KeyWord name="SetRegView" />
 		<KeyWord name="SetShellVarContext" />
 		<KeyWord name="SetSilent" />
-		<KeyWord name="SetStaticBkColor" />
```
donho, line 206: It seems a valid API too.

PowerEditor/installer/APIs/nsis.xml
sredna: They do not even exist in the first CVS2SVN check-in 20 years ago: https://sourceforge.net/p/nsis/code/625/tree/NSIS/trunk/Source/tokens.cpp There are some macros with those names but the syntax would be ${GetExeName} and they are not a part of the base instructions.

$EXEPATH and $EXEFILE variables have been around for 15 years.

PowerEditor/installer/APIs/nsis.xml
sredna: Replaced in 2003: https://sourceforge.net/p/nsis/code/2347/

Issue #11446, opened 2022-03-28
Function List enhancements on sorting


dfs-:
### Description of the Issue
Opening a document while function list is enabled and the function list parser has mixed mode parser (classes and functions, like i.e. Python) the classes are always listed first, despite their position in the document. Sorting the list using the appropriate button in the function list panel and then unsorting it will sort the elements in the correct order.

Also it would be great to have a preference setting to always enable sorting for newly opened documents.

### Steps to Reproduce the Issue
1. Open a Python document with functions and classes
2. Open the function list and notice how classes are always on top regardless of functions being in the same file much earlier
3. Sort the function list (-> gets sorted lexicographically)
4. Unsort the function list (-> gets sorted correctly)

### Expected Behavior
Sorting should always be consistent between states.

### Actual Behavior
Inital sorting has classes above functions regardless of the their position.

### Debug Information
Notepad++ v8.3.3   (64-bit)
Build time : Mar 13 2022 - 17:20:02
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19044.1586
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll PythonScript.dll XMLTools.dll 

See also:
https://community.notepad-plus-plus.org/topic/22756/function-list-weirdness-and-sorting

PR #11447, opened 2022-03-28
Function list default sorting (#11446)

dfs-:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11446:
* add new preference setting to allow function list to be sorted lexicographically by default
* fix sorting for mixed mode function list parsers on new file open
* add new translations to english.xml and german.xml

image

Stored in `config.xml`
```<?xml version="1.0" encoding="UTF-8" ?>
<NotepadPlus>
    ...
    <GUIConfigs>
        ...
        <GUIConfig name="MISC" ... sortFunctionList="yes" ... />
        ...      
    </GUIConfigs>
</NotepadPlus>
```


ArkadiuszMichalski:
Don't you think that it is more convenient to put the options of individual panels in their context menu? This makes switching them much more convenient and faster. Similar what we have for `Document List` (`on/off Ext. and Path.` - it was also in `Preferences` in the past, but has been moved.).

dfs-:
> Don't you think that it is more convenient to put the options of individual panels in their context menu?

I definitely see your point. 
I considered it but thought it would be a bit weird to have it as a single line context menu that also has nothing to do with the elements displayed in the function list. Another approach that I just now came up with is having a third button in the panel depicting a gear or something which would allow for the option to be changed...

dfs-:
@ArkadiuszMichalski Something like this?
image

dfs-:
> Btw, what is correlation beetwen this new option and AZ button? When I check this option then button is not active, but when I restart Notepad++ button is active. Same is when I turn off this option and restart Notepad++.

It will set the default sorting when a new document is opened. Some people like me prefer the list to be sorted by name rather than by where in the document the specific class/function might be found. 

It willingly does not align the currently opened documents function sorting as one might have specifically changed the setting for documents currently active/opened.

donho:
@dfs-  @ArkadiuszMichalski 

Tested this PR, there are several problems:
image

1. Choose `Sort functions (A to Z) by default` doesn't check the menu item, you have choose it again to make check mark appear.
2. Choose `Sort functions (A to Z) by default` doesn't execute Sorting A-Z command.
3. Sort functions (A to Z) by default` is not memorized after Notepad++ being restarted

Or am I missing the point of this feature?

ArkadiuszMichalski:
> @donho Or am I missing the point of this feature?

You're right, after the recent changes it doesn't work at all. One another problem is when we start or restart Notepad++ and try switch this option then we have two times click on it (first time doesn't change it state), after this single clicking change it state (until next restart).

donho:
@ArkadiuszMichalski @dfs- 
It's a good idea to put this option into Function List Panel.
But there might be some GUI incoherence.
The current behaviour is: sorting option is applied to each document individually, and it's kept only in the current session.
I suppose `Sort functions (A to Z) by default` will apply sorting A-Z not only to the current document, but also to all opened documents of the current session. And of course, it will be remembered through the sessions.
Is my description accurate?

ArkadiuszMichalski:
From this https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11447#issuecomment-1081091604 intention was change behaviour only for new opened document (which also means that after launching Notepad++ it will apply to everyone). But I didn't call for this option so I don't have any particular preferences here, just test what has been done. Maybe @Yaron10 has some additional suggestions?

dfs-:
@donho @ArkadiuszMichalski 
I'll recheck the changes and see if I can reproduce your problems.

dfs-:
@donho 

> * Choose `Sort functions (A to Z) by default` doesn't check the menu item, you have choose it again to make check mark appear.

Should have been fixed.

> * Choose `Sort functions (A to Z) by default` doesn't execute Sorting A-Z command.

Currently this is this implemented behaviour. The option should enable the "A-Z" button only for files that are opened after the setting has been changed, as all other files already have their state stored in the "previous params list", see also:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/658b626257d5f000575e3361a325c707dc5cf697/PowerEditor/src/WinControls/FunctionList/functionListPanel.cpp#L393-L410

If it should behave differently, it should be decided (by whom?) if the checkbox automatically enables/disables the sorting for all opened files or only the current one.

> * Sort functions (A to Z) by default` is not memorized after Notepad++ being restarted

Should have been fixed.

Yaron10:
@ArkadiuszMichalski,

Sorry for the late response. I was not by my PC yesterday.

I haven't tested the second version of this PR (i.e the toolbar button, before @donho's comments).
It seems to work as expected now.
Thank you all.

***

Isn't the use of the the toolbar button somewhat unconventional?
Left-clicking a toolbar button normally executes a command.

> I considered it but thought it would be a bit weird to have it as a single line context menu that also has nothing to do with the elements displayed in the function list.

The context-menu can be displayed only on right-clicking the toolbar or an empty space as implemented in the Documents List.

PR #11464, opened 2022-04-02
Fix case-conversion code to handle changing string lengths

### LINKED ISSUE #11463: Case conversion malfunctions when the string length changes
To reproduce:

* Paste the text `ȺRGH` into a UTF-8 encoded document.
* Select it and press Ctrl+U to lower-case it.

Expected result: `ⱥrgh`

Actual result: `ⱥrg`

This happens because `Ⱥ` encodes to two bytes and `ⱥ` encodes to three bytes, and the case-conversion code assumes that the length in bytes doesn't change. If the length increases (as above) then you lose bytes from the end. If it decreases (e.g. upper-casing `ⱥrgh`) then you get NULs and other garbage at the end.

benrg:
Fixes an issue where, e.g., upper-casing U+2C65 adds garbage to the
document because its capital form U+023A has a shorter UTF-8 encoding.
Incidentally fixes another bug: rows would be truncated at the first
NUL when changing case in column mode.

Fix #11463

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2778,6 +2778,43 @@ void ScintillaEditView::currentLinesDown() const
 	execute(SCI_SCROLLRANGE, execute(SCI_GETSELECTIONEND), execute(SCI_GETSELECTIONSTART));
 }
 
+// Case converts the document byte range [start:end] in place and returns the
+// change in its length in bytes. On any error, does nothing and returns zero.
+intptr_t ScintillaEditView::caseConvertRange(intptr_t start, intptr_t end, TextCase caseToConvert)
+{
+	if (end <= start || uintptr_t(end) - uintptr_t(start) > INT_MAX/2)
+		return 0;
+
+	unsigned codepage = getCurrentBuffer()->getUnicodeMode() == uni8Bit ? _codepage : CP_UTF8;
+
+	int mbLen = int(end - start);
+	const int mbLenMax = 2 * mbLen + 1;  // allow final NUL + substantial expansion
+
+	char *mbStr = new char[mbLenMax];
+	getText(mbStr, start, end);
+
+	if (int wideLen = ::MultiByteToWideChar(codepage, 0, mbStr, mbLen, NULL, 0)) {
+		wchar_t *wideStr = new wchar_t[wideLen];  // not NUL terminated
+		::MultiByteToWideChar(codepage, 0, mbStr, mbLen, wideStr, wideLen);
+
+		changeCase(wideStr, wideLen, caseToConvert);
+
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2778,7 +2778,9 @@ void ScintillaEditView::currentLinesDown() const
 	execute(SCI_SCROLLRANGE, execute(SCI_GETSELECTIONEND), execute(SCI_GETSELECTIONSTART));
 }
 
-void ScintillaEditView::changeCase(__inout wchar_t * const strWToConvert, const int & nbChars, const TextCase & caseToConvert) const
+namespace {
+
+void changeCase(__inout wchar_t * const strWToConvert, int nbChars, TextCase caseToConvert)
```
donho, line 2783: Please restore this part.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2900,72 +2902,81 @@ void ScintillaEditView::changeCase(__inout wchar_t * const strWToConvert, const
...
-	if (um != uni8Bit)
-	codepage = CP_UTF8;
+} // namespace
 
-	if (execute(SCI_GETSELECTIONS) > 1) // Multi-Selection || Column mode
-	{
-        execute(SCI_BEGINUNDOACTION);
+// Case converts the document byte range [start:end] in place and returns the
+// change in its length in bytes. On any error, does nothing and returns zero.
+intptr_t ScintillaEditView::caseConvertRange(intptr_t start, intptr_t end, TextCase caseToConvert)
```
donho, line 2909: Please move this method at the end.

donho (changes requested):
The diff is a mess that make review impossible.
I suggest you to restore ScintillaEditView.cpp by using `git checkout ScintillaEditView.cpp` then to  modify only where it does need, and finally to add caseConvertRange after convertSelectedTextTo, - to make code review possible. Thank you.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
benrg: Moving it after convertSelectedTextTo didn't fix the diff, but moving it before changeCase did. Sorry for the trouble.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
benrg: I reverted that and a few other drive-by changes.

donho (changes requested):
Please follow the coding style.
Otherwise good PR for the fix & the refactoring.

donho:
nm...
I'll correct them myself.
Thank you for the fix.

PR #11478, opened 2022-04-05
Fix squarish artifacts around rounded button corners

shriprem:
Related to fixes for: #10171, #10255, #10264, #10269

These squarish artifacts are visible in dark mode only in Window 7 & Windows 11 OS versions. But NOT in Windows 10 since the buttons in this OS version have inherently squarish corners.

Sample screenshot with this issue, before the fix:
image

Fixed in the following dialogs:
1. About Notepad++
2. Debug Info
3. Save All Files prompt
4. Save or Not Save Files prompt
5. MD5/SHA Generate (in both from text and From Files)
6. Incremental Search
7. Colour Chooser popup

Also, fixed in the following but could not test yet (since I am not aware of a path to invoke these dialogs in NPP):
1. Progress dialog (code in FindReplaceDlg.cpp)
2. TaskListDlg.cpp

### DIFF
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -51,6 +51,7 @@ intptr_t CALLBACK HashFromFilesDlg::run_dlgProc(UINT message, WPARAM wParam, LPA
 		}
 		return TRUE;
 
+		case WM_CTLCOLORBTN:
 		case WM_CTLCOLORDLG:
 		{
 			if (NppDarkMode::isEnabled())
@@ -371,6 +372,7 @@ intptr_t CALLBACK HashFromTextDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 			break;
 		}
 
+		case WM_CTLCOLORBTN:
 		case WM_CTLCOLORDLG:
 		{
 			if (NppDarkMode::isEnabled())
@@ -393,15 +395,6 @@ intptr_t CALLBACK HashFromTextDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 				{
 					return NppDarkMode::onCtlColorDarker(reinterpret_cast<HDC>(wParam));
 				}
-			}
-			break;
-		}
-
-		case WM_PRINTCLIENT:
-		{
-			if (NppDarkMode::isEnabled())
-			{
-				return TRUE;
 			}
 			break;
 		}
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -51,7 +51,6 @@ intptr_t CALLBACK HashFromFilesDlg::run_dlgProc(UINT message, WPARAM wParam, LPA
 		}
 		return TRUE;
 
-		case WM_CTLCOLORBTN:
 		case WM_CTLCOLORDLG:
 		{
 			if (NppDarkMode::isEnabled())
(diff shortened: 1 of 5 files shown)


shriprem:
@donho , in the other NPP dialogs where you have already handled this issue, you have actually used the following code:
```C++
		case WM_PRINTCLIENT:
		{
			if (NppDarkMode::isEnabled())
			{
				return TRUE;
			}
			break;
		}
```

But while implementing the dark mode for my own plugins ([FWDataViz](https://github.com/shriprem/FWDataViz) & [GotoLineCol](https://github.com/shriprem/Goto-Line-Col-NPP-Plugin)), I found that I could slip in the `case WM_CTLCOLORBTN` together with the required code for `case WM_CTLCOLORDLG` to achieve the same results, as in the following sample:
```C++
		case WM_CTLCOLORBTN:
		case WM_CTLCOLORDLG:
		{
			if (NppDarkMode::isEnabled())
			{
				return NppDarkMode::onCtlColorDarker(reinterpret_cast<HDC>(wParam));
			}
			break;
		}
```

To fix the leftover dialogs today, I have used the latter approach, since it required just an line extra line of code per dialog/run_dlgProc.

Actually, I had noticed this issue only in a few NPP dialogs. Far fewer than the number of dialogs I eventually fixed. For an exhaustive search for dialogs with this issue, I searched for `case WM_CTLCOLORDLG` and then checked to see if that case was also accompanied with a `case WM_PRINTCLIENT` or not. I fixed the ones with the missing `case WM_PRINTCLIENT` occurrences.

ozone10:
Cool, this was bothering me, and could not figure how to fix it. 

Unfortunately without WM_PRINTCLIENT some dialogs will have visual glitch when switching from light mode to dark mode.
image

shriprem (commented):
Replacing with `case WM_PRINTCLIENT` code to fix the reported issue with `case WM_CTLCOLORBTN`

shriprem:
@ozone10 , thanks for reporting the issue. I have submitted a fresh commit into this PR. This commit is using the `case WM_PRINTCLIENT` instead of `case WM_CTLCOLORBTN` in all the 10 instances added in the initial commit.

I was able to replicate the issue with the non-usage of `case WM_PRINTCLIENT` by following this narrow replication path below:
1. Open NPP. Change to light mode.
2. Open and close any of these following dialogs: About NPP, Debug Info, MD5/SHA Generate from text,  MD5/SHA Generate from Files, Colour chooser popup, Incremental Search.
3. Now switch to dark mode.
4. Reopen any of the dialogs opened in step 2. You will notice the squarish corners around buttons and sometimes artifacts around other controls. These artifacts will go away only when you close and reopen NPP afresh.

With the new commit in this PR, those artifacts and squarish corners are being properly handled well.

donho:
@shriprem 
It's OK for this time, but please create always a new branch for a new PR:

shriprem:
Noted. And thanks for the consideration.

Saw the contributing guidelines after submitting the PR. I have been spoiled by the _ad hoc_ nature of PRs at the NPPPlugins repo. Will not happen again.

donho:
I cannot merge your PR finally.
Please create a new branch for a new PR please.

shriprem:
Sure. Will do so.

donho:
FYI: it seems to me that the last commit "[Merge branch 'notepad-plus-plus:master' into master](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11478/commits/f6a75e52947edf7bc2d8476800d9edd260f333e0)" caused the problem.

So creating the new branch for a new PR is the way to prevent from such situation.

Issue #11337, opened 2022-03-04
Please sort listing of languages in Style Configurator in alphabetic order


labeldevops:
While I am at it: another thing that you could change probably with one line: 

please sort listing of languages in Style Configurator in alphabetic order.

 I do not get why you do not changed that already - did you ever use it? Did you ever scroll the language list with full extra attention because it is not sorted in alphabetic order and you were looking for one language that you can not find on the list for the third time?

ULTRA annoying! Please sort it, THANKS!

Yaron10:
@labeldevops,

You can sort the languages in stylers.xml, and that should be the order in _Style Configurator_.

vinsworldcom:
You can just type the letters of the language and it should bring you there.

CennoxX:
This is already sorted in the default themes and should also be sorted in other themes 👍
To sort this, take every Theme in https://github.com/notepad-plus-plus/notepad-plus-plus/tree/master/PowerEditor/installer/themes and sort the LexerType-Nodes by desc-property.

PR #3726, opened 2017-09-09
Language menu and Style configurator sorted alphabetically

Cyrillev91:
**Language** menu (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/2566#issuecomment-321282203) and **Style configurator** (#1989) **sorted alphabetically** and use the same language name in the Style configurator as the language menu.

Rename **nimrod** Language to **nim**. https://nim-lang.org/blog/2014/12/29/version-0102-released.html

Change list extension for **REBOL** language (delete *.r because confict with r language). #3650

PS : I made a single PR because the last 2 changes are minimal and concerns the same files.

Issue #1989, opened 2016-06-21
Language list in Style Configurator is not sorted for all themes


sirzooro:
Language list in Style Configurator is not sorted, please sort it alphabetically.

N++ 6.9.2, windows

dail8859:
It is in alphabetical order other than "Global Styles" at the beginning and "Search Results" at the end.

sirzooro:
For some reason they are not sorted on my PC:

![styles](https://cloud.githubusercontent.com/assets/1969912/16299768/3b43a970-393c-11e6-9bf2-0dc7d1feca77.jpg)

sirzooro:
I just restarted N++ and list is sorted now. Strange.

dail8859:
Closing for now. If you find there is some way to reproduce the issue feel free to reopen with any additional information.

sirzooro:
I found one thing - when I scrolled theme list dropdown with mouse scroll, I found that for some of them list is sorted, while for other don't - it looks as on screenshot added earlier.

I also noticed that sorted list is not completely sorted - Javascript, JSON and Fortran(fixed) are below "Search result" at the end.

How can I open this ticker? I do not see any "Reopen" button or something like this.

dail8859:
Reopening since some themes (e.g. Monokai) is not sorted.

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
the notepad++ source code, please visit us at the notepad++ community forum, and
search if related topics exists. you are welcome to post either at similar
topics, or to create a new topic at Help Wanted or General Discussion.
(no extra account is needed, just use your github account to sign in)
```
[...]

PR #11484, opened 2022-04-06
sort languages in themes alphabetically

CennoxX:
sort all languages in themes alphabetically and place "Search result" at the end like already done in Default and DarkModeDefault, missplaced tabs in the xml are deleted/replaced with spaces.

This was already fixed for some themes:  #1989
A Pull Request for this was closed, because it was outdated: #3726
This would fix: #11337

### DIFF
--- a/PowerEditor/installer/themes/Bespin.xml
+++ b/PowerEditor/installer/themes/Bespin.xml
@@ -41,6 +41,113 @@ Credits:
 //-->
 <NotepadPlus>
     <LexerStyles>
+        <LexerType name="actionscript" desc="ActionScript" ext="">
+            <!--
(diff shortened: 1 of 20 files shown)


CennoxX:
Beware, due to moving the complete LexerType nodes, the diff here on Github is complicated to read...

donho:
@CennoxX 
It'll fastidious if updating theme each time needs sorting effort manually.
The better solution (which is implemented in #11508) would be sorting automatically so we don't mind how is the order in any theme file, including in `stylers.xml`
Thank you for your PR.

CennoxX:
Okay, that's understandable and a good solution to the problem. Thanks.

PR #11487, opened 2022-04-07
set default language for installer

### LINKED ISSUE #10582: [Feature Request]: Use language of installed version instead of OS language as default in updater
### Description of the Issue

The N++ installer asks for the installation language. The default seems to be set to the OS language (or current location). For people who want their installation in a different language, it would be convenient to have the language of the installed version as a default. Since many developers in non-English speaking countries prefer English for coding, this would be beneficial for many people.

1.  Open updater e.g. on a German Windows computer while having installed the English version of notepad++.

### Expected Behavior

Language of the installed version should be set as default.

### Actual Behavior

OS language is set as default.

DannyGersh:
resolve #10582
when a language is selected in the installer, it is stored in a registry key. the next time the installer installs, it loads that language from the registry.
the key is HKLM "SOFTWARE\${APP NAME}\InstallerLanguage"

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -145,6 +145,14 @@ Function .onInit
 	;
 	; --- PATCH END ---
 
+; look for previously selected language
+ClearErrors
+Var /GLOBAL temporaryLng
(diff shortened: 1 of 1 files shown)


dfs- (changes requested):
I'd suggest changing the indentation to allign with the statements before and after your sections.

Issue #10778, opened 2021-11-13
Installation message change request


poolshrk:
I hope this is the correct place for this request.

I am affiliated with a large international High School/Middle school competition.

I absolutely love Notepad++, and I like to put it on competition VMs that students use.

Unfortunately we can no longer use Notepad++ due to the installation message "Software is like sex: It's better when it's free".

It would be great if the message could be replaced with something more non-controversial for school systems.

If not, no worries, I will continue to love Notepad++. Thank you all for your time, hard work, and amazing software.

89z:
This is being discussed online:

https://news.ycombinator.com/item?id=30959244

I have to agree with issue poster, its kind of a dumb reason to lose users over.

89z:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11503

PR #11503, opened 2022-04-08
changed text that some people might find offensive

sandreas:
The installer text may be not the best to keep people attracted. So I changed it to a less provocative quote from Linus Torvalds.

Reference https://news.ycombinator.com/item?id=30959244

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -306,6 +306,6 @@ Section -FinishSection
   Call writeInstallInfoInRegistry
 SectionEnd
 
-BrandingText "Software is like sex: It's better when it's free"
+BrandingText "In open source, we feel strongly that to really do something well, you have to get a lot of people involved."
 
 ; eof


alankilborn:
I think this will be rejected, as it has come up before, and the author really likes the original quote.

89z:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10778

sdegutis:
> I think this will be rejected, as it has come up before, and the author really likes the original quote.

You never know, people can change. I read somewhere that in movies/books, the villains are static and the heroes dynamic.

russellballestrini:
veto

russellballestrini:
I like the original one.

donho:
> I like the original.

Me too.

alankilborn:
@sandreas 

>>I like the original.

>Me too.

Can't say I didn't warn you!

sandreas:
@donho Ok, thank you for your quick feedback.

@alankilborn Just to be clear: The text does not offend ME personally, I'm ok with it. I just thought, if it prevents someone using Notepad++, it may be *more professional* to get rid of it. When the author disagrees, it is his decision.

Issue #11433, opened 2022-03-24
"Enable current line highlighting" may function incorrectly 


Yaron10:
Using https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/42993546/job/i961vxwbvsc8e665/artifacts.
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1075460543. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1077716546. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1078171813.

Current line is **highlighted** if "Enable current line highlighting" is **unchecked** and vice versa.

PR #11511, opened 2022-04-10
Fix for "Enable current line highlighting" may function incorrectly

shriprem:
Fix for #11433

This is being proposed as a quick & temporary fix.

The documentation for [SCI_SETCARETLINEVISIBLE](https://www.scintilla.org/ScintillaDoc.html#SCI_SETCARETLINEVISIBLE) states:
>You can choose to make the background colour of the line containing the caret different with these messages. To do this, set the desired background colour with SCI_SETCARETLINEBACK, then use SCI_SETCARETLINEVISIBLE(true) to enable the effect.

Based on this, I first tried setting a value with SCI_SETCARETLINEBACK before the **SCI_SETCARETLINEVISIBLE** call. No luck.

Therefore, I surmised that maybe the Scintilla5 control is not fully initialized and ready to be able to process the **SCI_SETCARETLINEVISIBLE** call during the time when NPP is still launching. So, I tried using **PostMessage** instead of the **SendMessage** that was being used for the **SCI_SETCARETLINEVISIBLE** call. That seems to have fixed the issue.

Effectively, this PR change is:
```C++
	//_mainEditView.execute(SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow);
	//_subEditView.execute(SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow);
	PostMessage(_mainEditView.getHSelf(), SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow, NULL);
	PostMessage(_subEditView.getHSelf(), SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow, NULL);
```

The **_execute_** function is using a Scintilla Direct Function pointer to process a **SendMessage** API call. Since **PostMessage** API call is not processed immediately, a direct function pointer is redundant for the SCI_SETCARETLINEVISIBLE calls on both views.

In any case, this PR is intended as a quick & temporary fix until a better, long term solution can be found.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -273,8 +273,8 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_mainEditView.setWrapMode(svp._lineWrapMethod);
 	_subEditView.setWrapMode(svp._lineWrapMethod);
 
-	_mainEditView.execute(SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow);
-	_subEditView.execute(SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow);
+	PostMessage(_mainEditView.getHSelf(), SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow, NULL);
+	PostMessage(_subEditView.getHSelf(), SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow, NULL);
 
 	_mainEditView.execute(SCI_SETENDATLASTLINE, !svp._scrollBeyondLastLine);
 	_subEditView.execute(SCI_SETENDATLASTLINE, !svp._scrollBeyondLastLine);


shriprem:
#### Key Tip:
This issue can be reproduced by starting a fresh Notepad++ session that is using a config.xml with the attribute value of `currentLineHilitingShow="hide"` in the `ScintillaPrimaryView` GUIConfig element. Or, see the 5-step instruction here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1077748440.

Yaron10:
@shriprem,

Thank you for this nice temporary fix. 👍 
Thanks also for the **Key Tip**. This is how I encountered the issue.

shriprem:
@donho, please note that I have posted a 2nd way of fixing issue #11433 in PR #11518. If you find the other PR is the way to go, please reject this PR.

Personally, I feel #11518 is better than this PR.

I am also planning a 3rd PR for this issue fix. That 3rd PR will try to be compliant with Scintilla's new recommendation to use the [SCI_SETELEMENTCOLOUR](https://www.scintilla.org/ScintillaDoc.html#SCI_SETELEMENTCOLOUR) API instead of SCI_SETCARETLINEVISIBLE & SCI_SETCARETLINEBACK calls. I will need a bit more time to do that.

I am presenting these alternatives for a fix in separate PRs. I think this approach may facilitate an easier side-by-side evaluation than if they were all put together in multiple commits within a single PR. Please advise if you differ. Thanks!

Yaron10:
@shriprem,

> I am also planning a 3rd PR for this issue fix. That 3rd PR will try to be compliant with Scintilla's new recommendation to use the [SCI_SETELEMENTCOLOUR](https://www.scintilla.org/ScintillaDoc.html#SCI_SETELEMENTCOLOUR) API instead of SCI_SETCARETLINEVISIBLE & SCI_SETCARETLINEBACK calls. I will need a bit more time to do that.

👍 

Did you see https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1078171813? :)

donho:
The option 3 is taken, hence this PR is closed.

PR #11518, opened 2022-04-10
Fix option 2 for #11433 "Enable Current Line Highlighting"

### LINKED ISSUE #11433: "Enable current line highlighting" may function incorrectly

shriprem:
Fix option 2 of 3 for issue #11433 

#### Plus points for this approach
* This PR fix does **_NOT_** involve the use of an asynchronous message call that was used in PR #11511.
* The change involved with this fix is also much smaller than the first one. 

This method involves just adding the `isCurrentLineHiLiting()` check inside `ScintillaEditView::performGlobalStyles()`. So, the relevant lines of code after the change will be:

```C++
	const Style * pStyle = stylers.findByName(TEXT("Current line background colour"));
	if (pStyle && isCurrentLineHiLiting())
	{
		execute(SCI_SETCARETLINEBACK, pStyle->_bgColor);
	}
```

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2570,7 +2570,7 @@ void ScintillaEditView::performGlobalStyles()
 	StyleArray & stylers = nppParams.getMiscStylerArray();
 
 	const Style * pStyle = stylers.findByName(TEXT("Current line background colour"));
-	if (pStyle)
+	if (pStyle && isCurrentLineHiLiting())
 	{
 		execute(SCI_SETCARETLINEBACK, pStyle->_bgColor);
 	}


shriprem:
#### Please note:
I am also planning a 3rd PR for this issue fix. That 3rd PR will try to be compliant with Scintilla's new recommendation to use the [SCI_SETELEMENTCOLOUR](https://www.scintilla.org/ScintillaDoc.html#SCI_SETELEMENTCOLOUR) API instead of SCI_SETCARETLINEVISIBLE & SCI_SETCARETLINEBACK calls. I will need a bit more time to do that.

I am presenting these alternatives for a fix in separate PRs. I think this approach may facilitate an easier side-by-side evaluation than if they were all put together in multiple commits within a single PR.

donho:
The option 3 is taken, hence this PR is closed.

PR #11519, opened 2022-04-11
Fix option 3 for #11433 using SCI_SETELEMENTCOLOUR

### LINKED ISSUE #11433: "Enable current line highlighting" may function incorrectly

shriprem:
### Note: This is the 3rd alternative way to fix #11433. For the other two methods, see PR #11511 & PR #11518.

This PR fix uses the **SCI_SETELEMENTCOLOUR** API recommended by Scintilla instead of the _discouraged_ **SCI_SETCARETLINEBACK** & **SCI_SETCARETLINEVISIBLE** API calls.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -3745,8 +3745,8 @@ void Notepad_plus::command(int id)
 		case IDM_VIEW_CURLINE_HILITING:
 		{
 			COLORREF colour = (NppParameters::getInstance()).getCurLineHilitingColour();
-			_mainEditView.setCurrentLineHiLiting(!_pEditView->isCurrentLineHiLiting(), colour);
-			_subEditView.setCurrentLineHiLiting(!_pEditView->isCurrentLineHiLiting(), colour);
+			_mainEditView.setCurrentLineHiLiting(_pEditView->isCurrentLineHiLiting(), colour);
+			_subEditView.setCurrentLineHiLiting(_pEditView->isCurrentLineHiLiting(), colour);
 		}
 		break;
 
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -3744,8 +3744,11 @@ void Notepad_plus::command(int id)
 
 		case IDM_VIEW_CURLINE_HILITING:
 		{
-			COLORREF colour = (NppParameters::getInstance()).getCurLineHilitingColour();
-			bool hilite{ NppParameters::getInstance().getSVP()._currentLineHilitingShow };
+			NppParameters& nppParams = NppParameters::getInstance();
+
+			COLORREF colour{ nppParams.getCurLineHilitingColour() };
+			bool hilite{ nppParams.getSVP()._currentLineHilitingShow };
+
 			_mainEditView.setCurrentLineHiLiting(hilite, colour);
 			_subEditView.setCurrentLineHiLiting(hilite, colour);
 		}
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -3745,8 +3745,9 @@ void Notepad_plus::command(int id)
 		case IDM_VIEW_CURLINE_HILITING:
 		{
 			COLORREF colour = (NppParameters::getInstance()).getCurLineHilitingColour();
-			_mainEditView.setCurrentLineHiLiting(_pEditView->isCurrentLineHiLiting(), colour);
-			_subEditView.setCurrentLineHiLiting(_pEditView->isCurrentLineHiLiting(), colour);
+			bool hilite{ NppParameters::getInstance().getSVP()._currentLineHilitingShow };
+			_mainEditView.setCurrentLineHiLiting(hilite, colour);
+			_subEditView.setCurrentLineHiLiting(hilite, colour);
 		}
 		break;
 
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2570,9 +2570,9 @@ void ScintillaEditView::performGlobalStyles()
 	StyleArray & stylers = nppParams.getMiscStylerArray();
 
 	const Style * pStyle = stylers.findByName(TEXT("Current line background colour"));
-	if (pStyle)
+	if (pStyle && isCurrentLineHiLiting())
 	{
-		execute(SCI_SETCARETLINEBACK, pStyle->_bgColor);
+		execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, pStyle->_bgColor);
 	}
 
 	COLORREF selectColorBack = grey;
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2567,12 +2567,16 @@ void ScintillaEditView::expand(size_t& line, bool doExpand, bool force, intptr_t
 void ScintillaEditView::performGlobalStyles()
 {
 	NppParameters& nppParams = NppParameters::getInstance();
-	StyleArray & stylers = nppParams.getMiscStylerArray();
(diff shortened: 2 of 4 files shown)


shriprem (commented):
No need for the SCI_SETCARETLINEVISIBLE calls in Notepad_plus.cpp since `isCurrentLineHiLiting` is being called to check if SCI_SETELEMENTCOLOUR (on SC_ELEMENT_CARET_LINE_BACK) should be run or not in `ScintillaEditView::performGlobalStyles`.

shriprem (commented):
In NppCommands.cpp, the negations are not needed anymore since the `isCurrentLineHiLiting` call is now returning the NPP parameter value, and not the edit view's SCI_GETCARETLINEVISIBLE current value.

shriprem (commented):
For the SCI_SETELEMENTCOLOUR calls in both ScintillaEditView.cpp & ScintillaEditView.h files, the color value is being passed unmodified in its RGB format, and not in the RGBA format. Since using  either the `bgColor | 0xFF000000` or the (redundantly ineffective) `bgColor | 0x00000000` made no difference in the current line background color in NPP. Hence using the simplest `bgColor` value. But will need to watchout for any RGB/RGBA issues with future version changes of Scintilla.

Yaron10:
@shriprem,

👍 

> But the old logic of this feature has been implemented badly.

@donho in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11419#issuecomment-1078171813.

This PR seems very good. Let's wait and see what Don thinks. 

***

What do you think of the following minor change?
```
if (isCurrentLineHiLiting())
{
	pStyle = stylers.findByName(TEXT("Current line background colour"));
	if (pStyle)
	{
		execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, pStyle->_bgColor);
	}
}
```

Thank you.

shriprem:
@Yaron10 , if there is some runtime cost involved in the `stylers.findByName` call, then your restructured code would be better. I will defer to @donho on this. Thanks!

PowerEditor/src/ScintillaComponent/ScintillaEditView.h
```diff
@@ -469,14 +469,14 @@ friend class Finder;
...
-		execute(SCI_SETCARETLINEBACK, bgColor);
+		if (isHiliting)
+			execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, bgColor);
+		else
+			execute(SCI_RESETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, NULL);
 	};
 
 	bool isCurrentLineHiLiting() const {
-		return (execute(SCI_GETCARETLINEVISIBLE) != 0);
+		return NppParameters::getInstance().getSVP()._currentLineHilitingShow;
```
donho, line 479: The goal of `isCurrentLineHiLiting` is for detecting the state current line highlighting via Scintilla handle, and the same idea applies to the whole `ScintillaEditView` class.
There's no advantage to use this method since Scintilla API has been changed its behaviour. 
I suggest to remove `isCurrentLineHiLiting` method completely and use `NppParameters::getInstance().getSVP()._currentLineHilitingShow;` directly.

PowerEditor/src/NppCommands.cpp
```diff
@@ -3745,8 +3745,8 @@ void Notepad_plus::command(int id)
 		case IDM_VIEW_CURLINE_HILITING:
 		{
 			COLORREF colour = (NppParameters::getInstance()).getCurLineHilitingColour();
-			_mainEditView.setCurrentLineHiLiting(!_pEditView->isCurrentLineHiLiting(), colour);
-			_subEditView.setCurrentLineHiLiting(!_pEditView->isCurrentLineHiLiting(), colour);
+			_mainEditView.setCurrentLineHiLiting(_pEditView->isCurrentLineHiLiting(), colour);
+			_subEditView.setCurrentLineHiLiting(_pEditView->isCurrentLineHiLiting(), colour);
```
donho, line 3749: Use `NppParameters::getInstance().getSVP()._currentLineHilitingShow;`  instead of `isCurrentLineHiLiting`.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2570,9 +2570,9 @@ void ScintillaEditView::performGlobalStyles()
 	StyleArray & stylers = nppParams.getMiscStylerArray();
 
 	const Style * pStyle = stylers.findByName(TEXT("Current line background colour"));
-	if (pStyle)
+	if (pStyle && isCurrentLineHiLiting())
```
donho, line 2573: Use `NppParameters::getInstance().getSVP()._currentLineHilitingShow;`  instead of `isCurrentLineHiLiting`.

donho (changes requested):
Please consider the change request.

PowerEditor/src/ScintillaComponent/ScintillaEditView.h
shriprem: Sure. Have pushed a new commit with this change.

PowerEditor/src/NppCommands.cpp
shriprem: Done. Using a bool variable to do it once instead of two separate lookups.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
shriprem: Done. Also incorporated @Yaron10 's suggestion assuming that the findByName has some runtime cost which can be avoided when the current line highlighting preference is off.

shriprem:
GitHub is still displaying **Changes requested** in red even though I have provided a _Resolve conversation_ reply separately for each change requested. Hopefully this _summary_ comment will make it go away.

PowerEditor/src/NppCommands.cpp
```diff
@@ -3745,8 +3745,9 @@ void Notepad_plus::command(int id)
 		case IDM_VIEW_CURLINE_HILITING:
 		{
 			COLORREF colour = (NppParameters::getInstance()).getCurLineHilitingColour();
-			_mainEditView.setCurrentLineHiLiting(!_pEditView->isCurrentLineHiLiting(), colour);
-			_subEditView.setCurrentLineHiLiting(!_pEditView->isCurrentLineHiLiting(), colour);
+			bool hilite{ NppParameters::getInstance().getSVP()._currentLineHilitingShow };
```
donho, line 3748:
```cpp
COLORREF colour = (NppParameters::getInstance()).getCurLineHilitingColour();
...
bool hilite{ NppParameters::getInstance().getSVP()._currentLineHilitingShow };
```
could be 
```cpp
NppParameters& nppParams = NppParameters::getInstance();
COLORREF colour = nppParams.getCurLineHilitingColour();
...
bool hilite{ nppParams.getSVP()._currentLineHilitingShow };
```
for the sake of readability.

donho (changes requested):
I believe it's the last request for this PR :)
Thank you in advance!

Yaron10:
@shriprem & @donho,

Great work and comments. 👍 
Thank you.

donho (approved):

PowerEditor/src/NppCommands.cpp
donho: > So, I have converted the COLORREF color & bool hilite assignments to direct brace initialization. Since nppParams is taking reference assignment, my old habit still favors copy initialization for it.

I have no problem with such way for initialization. Though I would use the traditional way - You know that you can't teach an old dog new tricks :)

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e094e996979deb10ffbe8e8228e4b238ddbf202c

Issue #11501, opened 2022-04-07
Framed Current Line option


shriprem:
### Change Request
It would be nice to have the **Framed Current Line** option in Notepad++ similar to the ones in Visual Studio or VS Code.

Currently it is achievable with either the [FWDataViz](https://github.com/shriprem/FWDataViz) plugin OR by running the following command with the [PythonScript]( https://github.com/bruderstein/PythonScript) plugin: `editor.setCaretLineFrame`.

The result would be something like:
[  _The clip below has line wrapping enabled on purpose. It is capturing the related settings line in NPP's config.xml._  ]

![image](https://user-images.githubusercontent.com/45252729/162333290-22e53d4d-f423-4f91-950b-3712a71395db.png)

---

### Preference dialog

In the Notepad++ Preferences dialog, this setting can be handled either with a checkbox or a horizontal slider control ranging between values 0 to 10. If we implement it with a checkbox, an ideal value to use with the `SCI_SETCARETLINEFRAME` call would be: 2

Mock screenshot with the new checkbox control in Preferences:

![image](https://user-images.githubusercontent.com/45252729/162334062-62c6cd89-a9c2-4245-8e9c-7545cbd4a6d6.png)

---

### Implementation

The current-line-framing feature is achievable via the Scintilla API function [SCI_SETCARETLINEFRAME](https://www.scintilla.org/ScintillaDoc.html#SCI_SETCARETLINEFRAME).

The _executing_ code for this new setting would be the last two lines below added inside Notepad_plus.cpp:
[ _The first two lines below are pre-existing, and are only indicated here for the location context._ ] 
```C++
...
	_mainEditView.execute(SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow);
	_subEditView.execute(SCI_SETCARETLINEVISIBLE, svp._currentLineHilitingShow);

	_mainEditView.execute(SCI_SETCARETLINEFRAME, svp._currentLineFrameWidth);
	_subEditView.execute(SCI_SETCARETLINEFRAME, svp._currentLineFrameWidth);
...
```

Of course, there will be other minor additions to: Parameters.h, Parameters.cpp, preference.rc, preferenceDlg.cpp, config4zipPackage.xml. Plus, additions to unitTest scripts for functionList and xml.  Lastly, this will also impact various translation files.

alankilborn:
[...]

PR #11526, opened 2022-04-12
11501 - Framed Current Line option

shriprem:
New feature: #11501

* Adds a new checkbox _Enable current line frame_ and a slider control to specify the width.
* The width slider control ranges between 1 and 6 (pixels).
* The width slider control is enabled only when _Enable current line frame_ is checked.

image

#### Related changes:
* **currentLineUseFrame** and **currentLineFrameWidth** attributes added to the **ScintillaPrimaryView** GUIConfig element in **config.xml**.
* The default value for **currentLineUseFrame** is _false_ for continuity of user experience.
* The default value for **currentLineFrameWidth** is _1_.
* The two new default values have also been added in the **config.4zipPackage.xml** file.
* The two new values are also added in the unit test files.
* New menu command function defined in **menuCmdId.h**: `IDM_VIEW_CURLINE_FRAMING           (IDM_VIEW + 121)`

#### Pending:
* Translation handling within NPP code?
* Updates to individual language files for the new preference controls?
* Updates to documentation?


Yaron10:
@shriprem,

👍 
Thank you for your work.

> If the Enable current line highlighting is unchecked, the framing is automatically turned off by Scintilla. However for the users clarity, the Frame Width slider should be disabled when the Enable current line gets unchecked.

@shriprem in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11501#issuecomment-1093433973.

The Frame CheckBox should be disabled as well, shouldn't it?
I'd also suggest to indent the Frame CheckBox so that it's clear it depends on "Enable current line highlighting"

alankilborn:
>Translation handling within NPP code?

This should be done. (Answering this way because you put a `?` at the end)

>Updates to individual language files for the new preference controls?

This is not necessary, except for english.xml.  There are people that specialize in most languages that will do the other translations.

>Updates to documentation?

This would be refreshing...to see someone that submits a code change also submit the corresponding user manual change.

PowerEditor/src/menuCmdID.h
```diff
@@ -365,7 +365,9 @@
     #define    IDM_VIEW_SWITCHTO_FUNC_LIST        (IDM_VIEW + 108)
     #define    IDM_VIEW_SWITCHTO_DOCLIST          (IDM_VIEW + 109)
 
-    #define    IDM_VIEW_GOTO_ANOTHER_VIEW        10001
+    #define    IDM_VIEW_CURLINE_FRAMING           (IDM_VIEW + 121)
```
Yaron10, line 368: This file defines **menu commands** only.

shriprem: @Yaron10,

> This file defines menu commands only.

I was merely replicating this action in the same way how the _Enable current line highlighting_ action is also implemented. AFAIK, _Enable current line highlighting_ is also not a menu option. But we find this in **NppCommands.cpp**:

```C++
		case IDM_VIEW_CURLINE_HILITING:
		{
			NppParameters& nppParams = NppParameters::getInstance();

			COLORREF colour{ nppParams.getCurLineHilitingColour() };
			bool hilite{ nppParams.getSVP()._currentLineHilitingShow };

			_mainEditView.setCurrentLineHiLiting(hilite, colour);
			_subEditView.setCurrentLineHiLiting(hilite, colour);
		}
		break;

		case IDM_VIEW_CURLINE_FRAMING:
		{
			const ScintillaViewParams& svp = NppParameters::getInstance().getSVP();

			bool useFrame{ svp._currentLineUseFrame };
			unsigned char frameWidth{ svp._currentLineFrameWidth };

			_mainEditView.setCurrentLineFraming(useFrame, frameWidth);
			_subEditView.setCurrentLineHiLiting(useFrame, frameWidth);
		}
		break;
```

The benefit from this is that plugins/scripts will gain the ability to invoke these actions via the [NPPM_MENUCOMMAND](https://npp-user-manual.org/docs/plugin-communication/#nppm-menucommand) message.

alankilborn:
>The benefit from this is that plugins/scripts will gain the ability to invoke these actions via the [NPPM_MENUCOMMAND](https://npp-user-manual.org/docs/plugin-communication/#nppm-menucommand) message.

This is a "nice thing" but isn't the usual situation.
It creates an inconsistency if only _some_ of the _Preferences_ are controllable in this manner.
:-(

shriprem:
@alankilborn, such inconsistencies are already there. I am sure if we take the time to look for them, we will find more such non-menu actions in addition to the (currently existing) IDM_VIEW_CURLINE_HILITING and the (proposed) IDM_VIEW_CURLINE_FRAMING actions.

To rephrase, it was not actually my intention to put _Enable current line framing_ together with other menu commands. I was only trying to stick with the same implementation pattern of how _Enable current line highlighting_ has been implemented. The starting point for that pattern is here, in **preferenceDlg.cpp**:

```C++
   case IDC_CHECK_CURRENTLINEHILITE:
      svp._currentLineHilitingShow = (BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_CHECK_CURRENTLINEHILITE, BM_GETCHECK, 0, 0));
      ::SendMessage(_hParent, WM_COMMAND, IDM_VIEW_CURLINE_HILITING, 0);
      return TRUE;

   case IDC_CARETLINEFRAME_WIDTH_CHECK:
      svp._currentLineUseFrame = (BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_CARETLINEFRAME_WIDTH_CHECK, BM_GETCHECK, 0, 0));
      ::EnableWindow(::GetDlgItem(_hSelf, IDC_CARETLINEFRAME_WIDTH_SLIDER), svp._currentLineUseFrame);
      redraw();

      ::SendMessage(_hParent, WM_COMMAND, IDM_VIEW_CURLINE_FRAMING, 0);
      return TRUE;
```

So, I wonder why the exclusion rule be enforced only to IDM_VIEW_CURLINE_FRAMING?

alankilborn:
I was just pointing out the inconsistency.  It's messy.
Of course, not something you created, @shriprem

Yaron10:
@shriprem,

I do apologize. I wrote that comment without properly checking it.
Let's wait for Don's review.

Thank you.

shriprem:
@Yaron10, no need for apologies.

We all want the best from NPP. So a civil & constructive discussion is very vital. That's what I've seen in all your feedback.

While implementing this feature, for the most part, I was trying to be very faithful to the existing coding style and patterns. I was on auto-pilot while doing some of the code changes. For example, in the `PowerEditor/Test/FunctionList/cpp/unitTest` file on line 5670, I even modified this _commented_ line by adding `currentLineUseFrame="hide" currentLineFrameWidth="1" ` just because `currentLineHilitingShow="show"` was there on that line. Because I wasn't sure if the testing utility is sensitive even to commented lines like how some doc-gen programs are.

alankilborn:
>What do you think?

I have never seen a PR changing unit test files either.

shriprem:
> The Frame CheckBox should be disabled as well, shouldn't it?

There are 3 ways to handle this:
1. Leave both the _checked_ & _enabled_ states of the _Enable current line highlighting_ & _Enable current line framing_ checkboxes independent. This was the case as of the [07634b4](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526/commits/07634b4bd73fbff97df50eed696091917b4cc9bb) commit.
2. Manage the _enabled_ state of the _Enable current line framing_ checkbox and the slider based on _checked_ state of the _Enable current line highlighting_ checkbox. This would be the conventional approach. Easy for everyone of us here to visualize even without an actual implementation.

However, there is a third way...

### 3. The SmartSense or IntelliSense approach
* Automatically uncheck the _Enable current line framing_ checkbox & disable the slider only when the _Enable current line highlighting_ checkbox is unchecked by the user.
* Automatically check the _Enable current line highlighting_ checkbox when the _Enable current line framing_ checkbox is checked by the user.

This approach is more intuitive of the user's intent, and gets the job done in a single click instead of two clicks. This is very much akin to how IDEs like Visual Studio & Eclipse automatically add the #include statement for the relevant header file when a function call just typed out by a programmer requires it.

I have this implementation in the [315e2b2](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526/commits/315e2b224a763d81db07b5cf9a1dcefa9e459040) commit. If after assessing this approach, the consensus is to revert back to the conventional approach, we can easily do so. But let us try this out first.

The partial lockstep traits for the two checkboxes is being achieved within the two `if`  statement blocks below.
```C++
   case IDC_CHECK_CURRENTLINEHILITE:
      svp._currentLineHilitingShow = (BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_CHECK_CURRENTLINEHILITE, BM_GETCHECK, 0, 0));
      if (!svp._currentLineHilitingShow)
      {
         ::SendDlgItemMessage(_hSelf, IDC_CARETLINEFRAME_WIDTH_CHECK, BM_SETCHECK, false, 0);
         ::EnableWindow(::GetDlgItem(_hSelf, IDC_CARETLINEFRAME_WIDTH_SLIDER), false);
         redraw();
      }

      ::SendMessage(_hParent, WM_COMMAND, IDM_VIEW_CURLINE_HILITING, 0);
      return TRUE;

   case IDC_CARETLINEFRAME_WIDTH_CHECK:
      svp._currentLineUseFrame = (BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_CARETLINEFRAME_WIDTH_CHECK, BM_GETCHECK, 0, 0));
      if (svp._currentLineUseFrame)
      {
         ::SendDlgItemMessage(_hSelf, IDC_CHECK_CURRENTLINEHILITE, BM_SETCHECK, true, 0);
      }

      ::EnableWindow(::GetDlgItem(_hSelf, IDC_CARETLINEFRAME_WIDTH_SLIDER), svp._currentLineUseFrame);
      redraw();

      ::SendMessage(_hParent, WM_COMMAND, IDM_VIEW_CURLINE_FRAMING, 0);
      return TRUE;
```

alankilborn:
>Automatically uncheck the Enable current line framing checkbox & disable the slider only when the Enable current line highlighting checkbox is unchecked by the user.
Automatically check the Enable current line highlighting checkbox when the Enable current line framing checkbox is checked by the user

This certainly seems like a good approach to me.

shriprem:
>I'd also suggest to indent the Frame CheckBox so that it's clear it depends on "Enable current line highlighting"

@Yaron10, I have added this indent. To do this, the width allocated for the checkbox had to be reduced so that it could be shifted slightly to the right. The width allocated for the _Enable current line framing_ checkbox in the resource file is only 180 units. As compared to 270 units allocated for the _Enable current line highlighting_ checkbox above it. I have a slight concern whether this reduced width may prove insufficient for some of the language translations.

Right now, the frame width slider and the two static texts on either side of it are aesthetically aligned to the width of the _Line Wrap_ groupbox above. If we were to increase the width of the _Enable current line framing_ checkbox, the slider group will need to be shifted further to the right, and may end up looking odd and out of balance.

shriprem:
Just pushed a new commit: [Added missing NPP & Scintilla state changes for the lockstep approach](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526/commits/b3d61ac0d3c2de20fcb5a1a9658642931f92a974).

Owing to this, I will edit the comment above to paste the modified code.

Yaron10:
@shriprem & @alankilborn,

> The SmartSense or IntelliSense approach
> > This certainly seems like a good approach to me.

This is clever, and I can see the advantage in this approach.

But if the user wants to toggle between Frame and _No-Highlight_, they would need to use two different controls: Frame for ON and Highlight for OFF.
I find it confusing.

I also don't like the "father" box being automatically checked on checking the "child" box.

I'd prefer the second traditional approach (and keep the Frame state even if it's disabled).
But I'm in the minority here. :) 

***

> To do this, the width allocated for the checkbox had to be reduced...

:+1: 

Thanks again for your work.

donho:
Several bugs have been detected:
1. Checking option makes strange behaviour:

image

2. Unchecking option makes line highlighting disappear:

image

3. When checking option works (seeing the frame), changing the width value via slidebar makes nothing change.

shriprem:
New commit: [ccd1cbf](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526/commits/ccd1cbf29c44e62c3b89b51283d3c875bf9fff92)

#### Enclosing std::stoi call for currentLineFrameWidth param within a try/catch block

1. It was not an issue if the **currentLineFrameWidth** attribute was totally absent in config.xml. 
2. The issue was if that attribute was present but was either empty or had a non-numeric value. i.e., currentLineFrameWidth="" or currentLineFrameWidth="x". Could possibly happen if the user inadvertently edits config.xml in a different editor and then tries launching NPP. Very far fetched, but still possible. If so, NPP would crash silently before launching. This commit fixes that.

Please see the code modified in this commit. Suggest if there is a better pattern used elsewhere in NPP for such string to numeric conversions.

Yaron10:
@donho,

> 2. Unchecking option makes line highlighting disappear:
> 3. When checking option works (seeing the frame), changing the width value via slidebar makes nothing change.

Did you test it with the last build?

Yaron10:
@shriprem,

> The issue was if that attribute was present but was either empty or had a non-numeric value. i.e., currentLineFrameWidth="" or currentLineFrameWidth="x". Could possibly happen if the user inadvertently edits config.xml in a different editor and then tries launching NPP. Very far fetched, but still possible. If so, NPP would crash silently before launching. This commit fixes that.

:+1: 
You're so thorough! :)

NPP can crash by manually modifying other values in config.xml.

Yaron10:
@donho,

We would appreciate your opinion about two points we've discussed here.

1. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526#pullrequestreview-939464678, and the following good replies.
2.  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526#issuecomment-1097016786, and the following reply.

Thank you.

Good night all.

shriprem:
I've found an issue with split view.

When launching, NPP is applying the specified color and frame width to both views (restored from a previous NPP session).

Then start fiddling with the frame width slider control. The frame width is getting applied only in the primary view. And, the color changes to a different color in the secondary view while there is no change in frame width here.

In the clip below, the primary view has the specified current line color (red) and the frame width. While the sub-view on the right has a black(?) color and frame width hasn't changed.

But close & reopen NPP. Now, both views are consistent with user specified settings.

image

Strange but will be fascinating to resolve this.

PowerEditor/Test/FunctionList/cpp/unitTest
```diff
@@ -5667,7 +5687,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("onAnotherView"), _nppGUI._smartHiliteOnAnotherView ? TEXT("yes") : TEXT("no"));
 	}
 
-	// <GUIConfig name="ScintillaPrimaryView" lineNumberMargin="show" bookMarkMargin="show" indentGuideLine="show" folderMarkStyle="box" lineWrapMethod="aligned" currentLineHilitingShow="show" scrollBeyondLastLine="no" disableAdvancedScrolling="no" wrapSymbolShow="hide" Wrap="no" borderEdge="yes" edge="no" edgeNbColumn="80" zoom="0" zoom2="0" whiteSpaceShow="hide" eolShow="hide" borderWidth="2" smoothFont="no" />
+	// <GUIConfig name="ScintillaPrimaryView" lineNumberMargin="show" bookMarkMargin="show" indentGuideLine="show" folderMarkStyle="box" lineWrapMethod="aligned" currentLineHilitingShow="show" currentLineUseFrame="hide" currentLineFrameWidth="1" scrollBeyondLastLine="no" disableAdvancedScrolling="no" wrapSymbolShow="hide" Wrap="no" borderEdge="yes" edge="no" edgeNbColumn="80" zoom="0" zoom2="0" whiteSpaceShow="hide" eolShow="hide" borderWidth="2" smoothFont="no" />
```
donho, line 5690: Why do we need to change this file?

PowerEditor/Test/FunctionList/xml/unitTest
```diff
@@ -40,7 +40,7 @@
...
         <GUIConfig name="searchEngine" searchEngineChoice="2" searchEngineCustom="" />
         <GUIConfig name="SmartHighLight" matchCase="no" wholeWordOnly="yes" useFindSettings="no" onAnotherView="no">yes</GUIConfig>
-        <GUIConfig name="ScintillaPrimaryView" lineNumberMargin="show" bookMarkMargin="show" indentGuideLine="show" folderMarkStyle="box" lineWrapMethod="aligned" currentLineHilitingShow="show" scrollBeyondLastLine="no" disableAdvancedScrolling="no" wrapSymbolShow="hide" Wrap="no" borderEdge="yes" edge="no" edgeNbColumn="80" zoom="0" zoom2="0" whiteSpaceShow="hide" eolShow="hide" borderWidth="2" smoothFont="no" />
+        <GUIConfig name="ScintillaPrimaryView" lineNumberMargin="show" bookMarkMargin="show" indentGuideLine="show" folderMarkStyle="box" lineWrapMethod="aligned" currentLineHilitingShow="show" currentLineUseFrame="hide" currentLineFrameWidth="1" scrollBeyondLastLine="no" disableAdvancedScrolling="no" wrapSymbolShow="hide" Wrap="no" borderEdge="yes" edge="no" edgeNbColumn="80" zoom="0" zoom2="0" whiteSpaceShow="hide" eolShow="hide" borderWidth="2" smoothFont="no" />
```
donho, line 43: Why do we need to change this file?

PowerEditor/Test/FunctionList/xml/unitTest.expected.result
```diff
@@ -1 +1 @@
...
...rrentLineFrameWidth=\"1\" scrollBeyondLastLine=\"no\" disableAdvancedScrolling=\"no\" wrapSymbolShow=\"hide\" Wrap=\"no\" borderEdge=\"yes\" edge=\"no\" edgeNbColumn=\"80\" zoom=\"0\" zoom2=\"0\" whiteSpaceShow=\"hide\" eolShow=\"hide\" borderWidth=\"2\" smoothFont=\"no\"","GUIConfig name=\"DockingManager\" leftWidth=\"200\" rightWidth=\"200\" topHeight=\"200\" bottomHeight=\"200\"","ActiveTabs cont=\"0\" activeTab=\"-1\"","ActiveTabs cont=\"1\" activeTab=\"-1\"","ActiveTabs cont=\"2\" activeTab=\"-1\"","ActiveTabs cont=\"3\" activeTab=\"-1\"","FindHistory nbMaxFindHistoryPath=\"10\" nbMaxFindHistoryFilter=\"10\" nbMaxFindHistoryFind=\"10\" nbMaxFindHistoryReplace=\"10\" matchWord=\"no\" matchCase=\"no\" wrap=\"yes\" directionDown=\"yes\" fifRecuisive=\"yes\" fifInHiddenFolder=\"no\" dlgAlwaysVisible=\"no\" fifFilterFollowsDoc=\"no\" fifFolderFollowsDoc=\"no\" searchMode=\"0\" transparencyMode=\"1\" transparency=\"150\" dotMatchesNewline=\"no\" isSearch2ButtonsMode=\"no\"","History nbMaxFile=\"10\" inSubMenu=\"no\" customLength=\"-1\"","ProjectPanel id=\"0\" workSpaceFile=\"\"","ProjectPanel id=\"1\" workSpaceFile=\"\"","ProjectPanel id=\"2\" workSpaceFile=\"\""],"root":"unitTest"}
```
donho, line 1: Why do we need to change this file?

PowerEditor/src/NppCommands.cpp
```diff
@@ -3746,14 +3746,26 @@ void Notepad_plus::command(int id)
 		{
 			NppParameters& nppParams = NppParameters::getInstance();
 
-			COLORREF colour{ nppParams.getCurLineHilitingColour() };
-			bool hilite{ nppParams.getSVP()._currentLineHilitingShow };
+			const COLORREF colour{ nppParams.getCurLineHilitingColour() };
+			const bool hilite{ nppParams.getSVP()._currentLineHilitingShow };
```
donho, line 3750: Please add a white space before `{` for the sake of visibility.
```
const COLORREF colour { nppParams.getCurLineHilitingColour() };
const bool hilite { nppParams.getSVP()._currentLineHilitingShow };
```

PowerEditor/src/menuCmdID.h
donho: > AFAIK, Enable current line highlighting is also not a menu option. But we find this in NppCommands.cpp:

It was. Then it's been moved into Preferences dialog so it's the historic reason. 

> The benefit from this is that plugins/scripts will gain the ability to invoke these actions via the [NPPM_MENUCOMMAND](https://npp-user-manual.org/docs/plugin-communication/#nppm-menucommand) message.

I'm wondering if it's a good thing - since this option has its GUI part, there will be the incoherence between the state of this option and its GUI.

donho (changes requested):
Please consider the requests.
For the GUI part, I'm checking another way, I'll let you know soon.

alankilborn:
>>The benefit from this is that plugins/scripts will gain the ability to invoke these actions via the NPPM_MENUCOMMAND message.

>I'm wondering if it's a good thing - since this option has its GUI part, there will be the incoherence between the state of this option and its GUI.

Script writers could always keep this incoherence possibility in mind.  

Example: recent commit added allowing caret to move beyond end of line as a preference setting.  For a long time, I've had a script where I could toggle this via a shortcut key combo...because sometimes I want it on and sometimes I want it off...without wanting to be bothered to invoke preferences dialog, find where the setting is that I want, turn it off/on, and save it.

That being said, it probably wouldn't be good (consistent) that only some commands would be this way.  I suppose it should be all-or-none.

PowerEditor/Test/FunctionList/cpp/unitTest
shriprem: No need. Reverting back.

PowerEditor/Test/FunctionList/xml/unitTest
shriprem: No need. Reverting back.

PowerEditor/Test/FunctionList/xml/unitTest.expected.result
shriprem: No need. Reverting back.

PowerEditor/src/NppCommands.cpp
shriprem: Done. (It was the VS auto-formatting at work.)

shriprem:
@donho, I have reverted the definition & its usage of IDM_VIEW_CURLINE_HILITING in NppCommands.cpp, and substituted its role with a new definition of NPPM_INTERNAL_CARETLINEFRAME and its usage in NppBigSwitch.cpp.

For the GUI, I am planning a new PR as well. That will facilitate a side-by-side evaluation of the GUIs from the two PRs. Details for that will follow later.

donho:
@shriprem 
Thank you!

> For the GUI, I am planning a new PR as well. That will facilitate a side-by-side evaluation of the GUIs from the two PRs. Details for that will follow later.

I'm also doing a mock-up for finding the good way to do GUI part. But you can try on your side either.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -84,15 +84,27 @@ BEGIN
...
+    LTEXT           "1",IDC_CARETLINEFRAME_WIDTH_DISPLAY,353,116,12,8
+    CONTROL         "Enable smooth font",IDC_CHECK_SMOOTHFONT,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,54,129,250,10
+    CONTROL         "Enable virtual space",IDC_CHECK_VIRTUALSPACE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,54,142,270,10
+    CONTROL         "Enable scrolling beyond last line",IDC_CHECK_SCROLLBEYONDLASTLINE,
+                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,54,155,270,10
+    CONTROL         "Keep selection when right-click outside of selection",IDC_CHECK_RIGHTCLICKKEEPSSELECTION,
+                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,54,168,270,10
+    CONTROL         "Disable advanced scrolling feature due to touchpad issue",IDC_CHECK_DISABLEADVANCEDSCROLL,
+                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,54,181,270,10
 END
```
donho, line 104: Please copy/paste the following codes:
```
IDD_PREFERENCE_SUB_EDITING DIALOGEX 0, 0, 455, 203
BEGIN
    GROUPBOX        "Caret Settings",IDC_CARETSETTING_STATIC,25,6,156,67,BS_CENTER
    LTEXT           "Width :",IDC_WIDTH_STATIC,53,27,37,8,0,WS_EX_RIGHT
    COMBOBOX        IDC_WIDTH_COMBO,97,25,40,30,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
    RTEXT           "Blink rate :",IDC_BLINKRATE_STATIC,28,52,60,8
    CONTROL         "",IDC_CARETBLINKRATE_SLIDER,"msctls_trackbar32",TBS_AUTOTICKS | TBS_BOTH | TBS_NOTICKS | TBS_TRANSPARENTBKGND | WS_TABSTOP,105,52,57,13
    LTEXT           "S",IDC_CARETBLINKRATE_S_STATIC,166,52,12,8
    LTEXT           "F",IDC_CARETBLINKRATE_F_STATIC,93,52,12,8,0,WS_EX_RIGHT

    GROUPBOX        "Line Wrap",IDC_LW_GB_STATIC,193,6,78,67,BS_CENTER
    CONTROL         "Default",IDC_RADIO_LWDEF,"Button",BS_AUTORADIOBUTTON | WS_GROUP,203,21,59,10
    CONTROL         "Aligned",IDC_RADIO_LWALIGN,"Button",BS_AUTORADIOBUTTON,203,36,60,10
    CONTROL         "Indent",IDC_RADIO_LWINDENT,"Button",BS_AUTORADIOBUTTON,203,51,62,10

    GROUPBOX        "Current Line Mark",IDC_CURRENTLINEMARK_STATIC,284,6,129,83,BS_CENTER
    CONTROL         "None",IDC_RADIO_CLM_NONE,"Button",BS_AUTORADIOBUTTON | WS_GROUP,307,22,59,10
    CONTROL         "Highlight",IDC_RADIO_CLM_HILITE,"Button",BS_AUTORADIOBUTTON,307,37,60,10
    CONTROL         "Frame",IDC_RADIO_CLM_FRAME,"Button",BS_AUTORADIOBUTTON,307,52,62,10
    LTEXT           "Width :",IDC_CARETLINEFRAME_WIDTH_STATIC,299,67,37,8,0,WS_EX_RIGHT
    CONTROL         "",IDC_CARETLINEFRAME_WIDTH_SLIDER,"msctls_trackbar32",TBS_AUTOTICKS | TBS_BOTH | TBS_NOTICKS | TBS_TRANSPARENTBKGND | WS_TABSTOP,337,67,57,13
    LTEXT           "1",IDC_CARETLINEFRAME_WIDTH_DISPLAY,396,67,12,8
	
    CONTROL         "Enable Multi-Editing (Ctrl+Mouse click/selection)",IDC_CHECK_MULTISELECTION,
                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,26,114,270,10
    CONTROL         "Enable smooth font",IDC_CHECK_SMOOTHFONT,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,26,129,250,10
    CONTROL         "Enable virtual space",IDC_CHECK_VIRTUALSPACE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,26,142,270,10
    CONTROL         "Enable scrolling beyond last line",IDC_CHECK_SCROLLBEYONDLASTLINE,
                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,26,155,270,10
    CONTROL         "Keep selection when right-click outside of selection",IDC_CHECK_RIGHTCLICKKEEPSSELECTION,
                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,26,168,270,10
    CONTROL         "Disable advanced scrolling feature due to touchpad issue",IDC_CHECK_DISABLEADVANCEDSCROLL,
                    "Button",BS_AUTOCHECKBOX | WS_TABSTOP,26,181,270,10
END
```

to have the new look:

image

PowerEditor/src/WinControls/Preference/preference_rc.h
```diff
@@ -136,6 +136,11 @@
 
 	#define	IDC_CHECK_VIRTUALSPACE (IDD_PREFERENCE_SUB_EDITING + 45)
 
+	#define	IDC_CARETLINEFRAME_WIDTH_CHECK (IDD_PREFERENCE_SUB_EDITING + 46)
+	#define	IDC_CARETLINEFRAME_WIDTH_STATIC (IDD_PREFERENCE_SUB_EDITING + 47)
+	#define	IDC_CARETLINEFRAME_WIDTH_SLIDER (IDD_PREFERENCE_SUB_EDITING + 48)
+	#define	IDC_CARETLINEFRAME_WIDTH_DISPLAY (IDD_PREFERENCE_SUB_EDITING + 49)
+
```
donho, line 143: Please erase your local `preference_rc.h` by using this one.

[preference_rc.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/8483401/preference_rc.zip)

donho (changes requested):
* Restore the unnecessary modified files
* Re-implement the new GUI part by using the provide codes.

donho:
I realize there are 2 Frame-current-line PR which based on the same branch.
It's confusing for the reviewing.
I will close both of them.
Please create a new branch for a new PR and let's focus on 1 feature for PR!

PowerEditor/src/WinControls/Preference/preference.rc
alankilborn: _Mark_ in _Current Line Mark_ might be poor wording, as the UI already has a lot of uses for this word.

How about _Current Line Emphasis_ for the title of this groupbox?

Also, _Highlight_ on its own is now weak -- what about _Colour Background_ instead?

In _Style Configurator_ the wording is _Current line background colour_.  That's fine but user needs to be aware that this will also serve for the _Frame_ colour.  I'm not saying there needs to be any change to _Style Configurator_, I'm just pointing this out.  Perhaps it is best handled in user manual.

alankilborn:
@Yaron10 Do my comments here have any value? https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526#discussion_r849739408  What do you think?

PowerEditor/src/WinControls/Preference/preference.rc
Yaron10: @alankilborn,

> Mark in Current Line Mark might be poor wording, as the UI already has a lot of uses for this word.

I agree.

> How about Current Line Emphasis for the title of this groupbox?

Translating "Emphasis" to my native language, it's not the best term.
But, if you're a native English speaker, you know better. :)

> In Style Configurator the wording is Current line background colour.

That's a good point.

So the title can be "Current line background colour".
And then:
"None"
"Entire line"
"Frame only"

If "Current line background colour" is too long, "Current line highlighting" seems good too.

Issue #11530, opened 2022-04-13
[DarkMode] Dark shortcut mapper


ozone10:
Notepad++ 8.3.3 does not apply dark mode on shortcut mapper
![image](https://user-images.githubusercontent.com/55940305/163193013-7a08d894-3f0a-4d77-97b3-4e72a6bc6492.png)

Issue #10358, opened 2021-08-09
Shortcut mapper / Macros tab shows garbled/cropped text in status area


artie-finkelstein:
### Description of the Issue
Text displayed in the "Shortcut mapper" status area (e.g., "No shortcut conflicts for this item.") only shows a few of the pixels of the bottom of the string but with most of the string shifted up past the top of the proper display area.

### Steps to Reproduce the Issue
1.  Npp 8.13rc or Npp 8.13rc2
2. click on \Macro\Modify Shortcut/Delete Macro...
3. 

### Expected Behavior
I expected to be able to read the text (see attached file "macro-good status.png") 

### Actual Behavior
I got the results show in the attached file "macro-bad status.png

### Debug Information
Notepad++ v8.1.3   (32-bit)
Build time : Aug  8 2021 - 13:17:47
Path : C:\Programs\Notepad++\notepad++_8.13rc2.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : none

This cropping does not occur with 8.12 or earlier releases.  It's only visible under 8.13rc & 8.13rc2 and shows up when '\Macro\Modify Shortcut/Delete Macro...' is used to initially display the 'Shortcut mapper' menu.  This cropping can also show up on the 'Plugin commands' tab, and may then show up on other tabs if they are selected after 'Plugin commands'.  (For this portion of testing I obviously had plugins enabled.)
```
[...]

PR #11531, opened 2022-04-13
Add initial dark mode support for Shortcut Mapper

ozone10:
Minor tweaks for drawAutocompleteColoursFromTheme to avoid double calling
const Style* pStyle = ...

Before I leave for while, some code for dark shortcut mapper.

partially fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11530
image

Missing is  dark grid.
Also there is still issue with https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10358


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2dbc25f91818e05ec1f8d8a518d8b700a57ec793

PR #11532, opened 2022-04-13
11501 - Framed Current Line with a simpler Preferences UI

shriprem:
New Feature: #11501

The Preferences GUI with this PR is much simpler as compared to PR #11526.

Due to the elimination of the _Enable current line frame_  checkbox, the GUI will be simpler for the end users to comprehend and use. Instead of the checkbox, user guidance is provided in the feedback text next to the slider (now with a range from **_0_** to 6).

While evaluating the two PRs, please see @Yaron10's analysis [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11501#issuecomment-1094116463).

image

image


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1869,10 +1877,6 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 			drawDocumentMapColoursFromStylerArray();
 
-			drawAutocompleteColoursFromTheme();
-			AutoCompletion::drawAutocomplete(_pEditView);
-			AutoCompletion::drawAutocomplete(_pNonEditView);
-
```
donho, line 1875: How can they be erased?

shriprem:
@donho , I saw it too. I have restored it back in the 2nd commit. Not sure how that happened.

Also, in the 3rd commit, I restored 3 preexisting lines in **preference.rc** that VS had reformatted on its own. So, the file change should only display only the lines I have added or modified.

donho:
I realize there are 2 Frame-current-line PR which based on the same branch.
It's confusing for the reviewing.
I will close both of them.
Please create a new branch for a new PR and let's focus on 1 feature for PR!

PR #11533, opened 2022-04-13
11501 - Current Line Frame with a simpler Preferences UI

shriprem:
New Feature: #11501

The Preferences GUI with this PR is much simpler as compared to PR #11526.

Due to the elimination of the _Enable current line frame_  checkbox, the GUI will be simpler for the end users to comprehend and use. Instead of the checkbox, user guidance is provided in the feedback text next to the slider (now with a range from **_0_** to 6).

While evaluating the two PRs, please see @Yaron10's analysis [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11501#issuecomment-1094116463).

image

image

PR #11534, opened 2022-04-14
Fully implemented - Don Ho's version of Preference dialog

shriprem:
New Feature: #11501

This PR implements the version of Preferences dialog UI designed by @donho 

For a different version of Preferences GUI implementation, see #11533

This PR:
image


shriprem:
@donho

* Please note that I have taken the liberty to change the title of the group box to **Current Line Indicator** as that sounds slightly better than **Current Line Mark** -- at least in my view. If you feel it needs to be changed back, let me know.

*  I have used the **preference_rc.zip** file as you provided it in your [code review](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11526#discussion_r849723382). That is, it also includes three commented lines of retired control IDs. If that needs to be totally removed from the file, let me know.

shriprem:
>Note: users who preferred not to highlight the current line in previous versions, would have the line highlighted after upgrading to the new version.
They would only need to re-apply their preference.

Yes. The plus point with that is it will draw attention of those users to the possibility of using the new _Frame_ option, assuming that earlier they were finding the full highlight too obtrusive and so had turned it off.

Yaron10:
> Yes. The plus point with that is it will draw attention...

👍 

I wrote those lines a note only. Not an issue.

alankilborn:
> users who preferred not to highlight the current line in previous versions, would have the line highlighted after upgrading to the new version

The MINUS point is that it will probably become an annoying tech support question, asked many dozens of times. :-(

shriprem:
@alankilborn,

>The MINUS point is that it will probably become an annoying tech support question, asked many dozens of times. :-(

In this initial iteration, it eased my thought process by keeping the radio control IDs, enum values, state variables and config values similar in their names/text values. It helped me get things working right. Now that is done, we can easily substitute the attribute names and values for backward compatibility. And, it needs to be done at only two places in the Parameters.cpp file.

In the clip below, all we need to do is to revert to the earlier attribute name and values in red, and add in the third state for "frame". That will take care of the _MINUS POINT_.

[ _But before I go ahead with this code change, what do you and @Yaron10 think about the naming for the groupbox in the first commit? Do I need to switch it back to what Don had it in his mockup?_ ]


image

alankilborn:
>what do you and @Yaron10 think about the naming for the groupbox in the first commit? Do I need to switch it back to what Don had it in his mockup?

Given that @Yaron10 didn't like "Emphasis" due to translation issues, "Indicator" is probably the next best thing, and certainly better than "Mark".  But...final call is Don's, so he may insist on "Mark".

Yaron10:
@shriprem,

> In the clip below, all we need to do is to revert to the earlier attribute name and values...

Thank you for that too. 👍 

> what do you and @Yaron10 think about the naming...

Let the ruler be a ruler. :)

donho:
How about `Mark Current Line with` instead of `Current Line Indicator` ?

shriprem:
@donho, sure I can do that.

donho:
> Note: users who preferred not to highlight the current line in previous versions, would have the line highlighted after upgrading to the new version.
They would only need to re-apply their preference.

@Yaron10  has revealed a very good point.


> The MINUS point is that it will probably become an annoying tech support question, asked many dozens of times. :-(

I second @alankilborn 's point. It'll be considered as a bug by user.

@shriprem 
> Yes. The plus point with that is it will draw attention of those users to the possibility of using the new Frame option, assuming that earlier they were finding the full highlight too obtrusive and so had turned it off.

I don't think it's a good way to "draw attention" of users.
The user's settings should be kept if it's possible. And it's quite easy to be done in this PR:
1. We should keep reading "currentLineHilitingShow" on Notepad++ starup.
2. If it exists we use the settings, otherwise read "markCrurrentLineWith" which cotains 0 (None), 1 (HighLight) or 2 (Frame).
3. On exit, write "markCrurrentLineWith" value but not "currentLineHilitingShow".

In this way, no user's complain, and their settings data is migrated seamlessly.

shriprem:
>We should keep reading "currentLineHilitingShow" on Notepad++ starup.
If it exists we use the settings, otherwise read "markCrurrentLineWith" which cotains 0 (None), 1 (HighLight) or 2 (Frame).
On exit, write "markCrurrentLineWith" value but not "currentLineHilitingShow".

@donho, I had already pushed the commit with the groupbox name change a few minutes prior to this. I can implement this new change for the config attribute renaming. But that will be in the next commit. Thanks!

---

#### PostScript:

To be quite honest, I used that earlier name of "Current Line Indicator" based on my familiarity with the [SetRowFocusIndicator](https://docs.appeon.com/pb2017r2/datawindow_reference/ch09s181.html) function name that applies on the current row in a datawindow (PowerBuilder terminology for the data grid). And that function uses the enumerated values of "Off!", "FocusRect!" and "Hand!". A very similar context to what we are dealing with the current line in NPP.

donho:
> To be quite honest, I used that earlier name of "Current Line Indicator" based on my familiarity with the [SetRowFocusIndicator](https://docs.appeon.com/pb2017r2/datawindow_reference/ch09s181.html) function name that applies on the current row in a datawindow (PowerBuilder terminology for the data grid). And that function uses the enumerated values of "Off!", "FocusRect!" and "Hand!". A very similar context to what we are dealing with the current line in NPP.

I do understand your reason for the new term. My concern is find an association with the old behaviour.
Let's see the opinion of others - We can always change the name of this feature before its release.

shriprem:
> We should keep reading "currentLineHilitingShow" on Notepad++ starup.
If it exists we use the settings, otherwise read "markCrurrentLineWith" which cotains 0 (None), 1 (HighLight) or 2 (Frame).
On exit, write "markCrurrentLineWith" value but not "currentLineHilitingShow".

Just confirming that these changes are in the latest commit [53cfc49](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/53cfc495fff99848726d3cbfe7ad4225bf8c42c0). Builds have completed as well.

Yaron10:
Two points about the latest implementation:
1. Checking for both "currentLineHilitingShow" and "markCurrentLineWith" in the code.
2. After running this binary and testing an older version, the current line is highlighted even if the preference was OFF.

shriprem:
>Two points about the latest implementation:
>
>1. Checking for both "currentLineHilitingShow" and "markCurrentLineWith" in the code.
>2. After running this binary and testing an older version, the current line is highlighted even if the preference was OFF.

@Yaron10, good observations.

Yes, in the preceding [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) commit, the attribute name of "currentLineHilitingShow" was retained and only the third state value of "frame" was added. The retention of the attribute name yielded both the backward and the first-forward-and-then-backward compatibilities that you mention in point 2.

However, after incorporating the [changes suggested](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534#issuecomment-1100271521), the dropping of the the old attribute name and adoption of a new attribute name have contributed to having only the one-time backward compatibility in the latest [53cfc49](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/53cfc495fff99848726d3cbfe7ad4225bf8c42c0) commit.

> 3. In (most) other GroupBoxes titles you don't use a verb.

Yes, I do agree with you on this point. The new groupbox name is now a verbal phrase.

Yaron10:
@shriprem,

Thank you. :)

> Yes, in the preceding [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) commit...

That's what I meant.
Don's GUI solution was very good. I didn't understand the last "changes request".

donho:
@Yaron10 
> I didn't understand the last "changes request".

The latest request is for solving the problem that you pointed out:
> Note: users who preferred not to highlight the current line in previous versions, would have the line highlighted after upgrading to the new version.
They would only need to re-apply their preference.

For users who preferred not to highlight the current line in the previous version, let's say  v8.4, when they upgrade Notepad++ to v8.4.1 (which contains this new feature), there should be no highlighting nor frame on the current line, and the choice should set to "None" in the concerning setting.

So the purpose of my latest request is to read the old settings, convert the old setting's value to the new one, and write the new settings on exit. In this way the retro-compatibility is kept and there won't be the bug report.

@shriprem 
The commi [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) doesn't process the old settings value to the new one, so users' old settings cannot be conserved.
Sorry for not being clear. I'll provide the codes for this part.

Yaron10:
@donho,

> The latest request is for solving the problem that you pointed out

But this was solved in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9, wasn't it?
Am I missing anything?

PowerEditor/src/Parameters.cpp
```diff
@@ -5666,14 +5666,38 @@ void NppParameters::feedScintillaParam(TiXmlNode *node)
...
-			_svp._currentLineHilitingShow = true;
-		else if (!lstrcmp(nm, TEXT("hide")))
-			_svp._currentLineHilitingShow = false;
+		if (!lstrcmp(nm, TEXT("hide")))
+			_svp._currentLineHiliteMode = LINEHILITE_NONE;
+		else if (!lstrcmp(nm, TEXT("show")))
+			_svp._currentLineHiliteMode = LINEHILITE_HILITE;
+		else if (!lstrcmp(nm, TEXT("frame")))
+			_svp._currentLineHiliteMode = LINEHILITE_FRAME;
+	}
```
donho, line 5685: The old data should be  processed as the following way:
```cpp
	// Current Line Highlighting State
	nm = element->Attribute(TEXT("currentLineHilitingShow"));
	if (nm)
	{
		if (!lstrcmp(nm, TEXT("show")))
			svp._currentLineHiliteMode = LINEHILITE_HILITE;
		else
			_svp._currentLineHiliteMode = LINEHILITE_NONE;
	}
	else
	{		
		const TCHAR * currentLineModeStr = element->Attribute(TEXT("currentLineIndicator"));
		if (currentLineModeStr && currentLineModeStr[0])
		{
			if (lstrcmp(currentLineModeStr, TEXT("1")) == 0)
				svp._currentLineHiliteMode = LINEHILITE_HILITE;
			else if (lstrcmp(currentLineModeStr, TEXT("2")) == 0)
				_svp._currentLineHiliteMode = LINEHILITE_FRAME;
			else
				_svp._currentLineHiliteMode = LINEHILITE_NONE;
		}
	}
```

PowerEditor/src/Parameters.cpp
```diff
@@ -6062,7 +6086,11 @@ bool NppParameters::writeScintillaParams()
 								(_svp._lineWrapMethod == LINEWRAP_INDENT)?TEXT("indent"):TEXT("default");
 	(scintNode->ToElement())->SetAttribute(TEXT("lineWrapMethod"), pWrapMethodStr);
 
-	(scintNode->ToElement())->SetAttribute(TEXT("currentLineHilitingShow"), _svp._currentLineHilitingShow?TEXT("show"):TEXT("hide"));
+	const TCHAR *plineHiliteStr = (_svp._currentLineHiliteMode == LINEHILITE_NONE)?TEXT("hide"):
+								(_svp._currentLineHiliteMode == LINEHILITE_FRAME)?TEXT("frame"):TEXT("show");
+	(scintNode->ToElement())->SetAttribute(TEXT("markCurrentLineWith"), plineHiliteStr);
```
donho, line 6091: These 3 lines can be reduced in 1 line:
```cpp
(scintNode->ToElement())->SetAttribute(TEXT("currentLineIndicator"), _svp._currentLineHiliteMode);
```

PowerEditor/src/Parameters.cpp
```diff
@@ -6656,7 +6684,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("customColorLinkText"), _nppGUI._darkmode._customColors.linkText);
 	}
 
-	// <GUIConfig name="ScintillaPrimaryView" lineNumberMargin="show" bookMarkMargin="show" indentGuideLine="show" folderMarkStyle="box" lineWrapMethod="aligned" currentLineHilitingShow="show" scrollBeyondLastLine="no" rightClickKeepsSelection="no" disableAdvancedScrolling="no" wrapSymbolShow="hide" Wrap="no" borderEdge="yes" edge="no" edgeNbColumn="80" zoom="0" zoom2="0" whiteSpaceShow="hide" eolShow="hide" borderWidth="2" smoothFont="no" />
+	// <GUIConfig name="ScintillaPrimaryView" lineNumberMargin="show" bookMarkMargin="show" indentGuideLine="show" folderMarkStyle="box" lineWrapMethod="aligned" markCurrentLineWith="show" scrollBeyondLastLine="no" rightClickKeepsSelection="no" disableAdvancedScrolling="no" wrapSymbolShow="hide" Wrap="no" borderEdge="yes" edge="no" edgeNbColumn="80" zoom="0" zoom2="0" whiteSpaceShow="hide" eolShow="hide" borderWidth="2" smoothFont="no" />
```
donho, line 6687: Please restore this line.

shriprem:
@donho, I am a bit puzzled here.

While testing with the [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) builds, I would repeatedly switch between that newer NPP build and the NPP 8.3.3 executables. If the current line indicator option was either of "None" (_still being saved as_ "hide") or "Highlight" (_still being saved as_ "show") from either of those executables,  the other executable would seamlessly pick it up.

The choices made in [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9):
1. to continue to use the same attribute name of "currentLineHilitingShow" AND
2. to continue to save "None" as "hide", and "Highlight" as "show"

yielded that two-way compatibility. Only when the "Frame" option was selected with the newer NPP build, the NPP 8.3.3 executable would go back to the default "Highlight" (i.e., the checked state of "Enable Current Line Highlight") setting.

@Yaron10, please correct me if I am wrong about the claims on testing observations with the [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) builds.

@donho, 
>The commi [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) doesn't process the old settings value to the new one, so users' old settings cannot be conserved.
Sorry for not being clear. I'll provide the codes for this part.

I only implemented the attribute name change that was specified in your [comments](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534#issuecomment-1100271521). And I did so despite knowing that @Yaron10's keen-eyed observation had been fixed with the [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) commit; and also while silently acknowledging that there is actually no need for a two-way compatibility of the config settings that was being accomplished with that commit.

I have just now gone back and carefully re-read your change suggestions. I now note that you said:
>If it exists we use the settings, otherwise read "markCrurrentLineWith" which cotains 0 (None), 1 (HighLight) or 2 (Frame).

By this. if you mean that I should implement the attribute values of "0", "1" & "2" instead of the currently implemented "hide", "show" & "frame", I will still be happy do so. OTOH, if there is something else that I have missed, I look forward to seeing the code that you are planning to provide. Thank you.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -68,31 +68,44 @@ BEGIN
...
+    CONTROL         "",IDC_CARETBLINKRATE_SLIDER,"msctls_trackbar32",TBS_AUTOTICKS | TBS_BOTH | TBS_NOTICKS | TBS_TRANSPARENTBKGND | WS_TABSTOP,105,52,57,13
+    LTEXT           "S",IDC_CARETBLINKRATE_S_STATIC,166,52,12,8
+    LTEXT           "F",IDC_CARETBLINKRATE_F_STATIC,93,52,12,8,0,WS_EX_RIGHT
+
+    GROUPBOX        "Line Wrap",IDC_LW_GB_STATIC,193,6,78,67,BS_CENTER
+    CONTROL         "Default",IDC_RADIO_LWDEF,"Button",BS_AUTORADIOBUTTON | WS_GROUP,203,21,59,10
+    CONTROL         "Aligned",IDC_RADIO_LWALIGN,"Button",BS_AUTORADIOBUTTON,203,36,60,10
+    CONTROL         "Indent",IDC_RADIO_LWINDENT,"Button",BS_AUTORADIOBUTTON,203,51,62,10
+
+    GROUPBOX        "Mark Current Line with",IDC_CURRENTLINEMARK_STATIC,284,6,129,83,BS_CENTER
```
donho, line 88: I cannot find better name so far, so OK for `Current Line Indicator`.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -68,31 +68,44 @@ BEGIN
...
+    CONTROL         "None",IDC_RADIO_CLM_NONE,"Button",BS_AUTORADIOBUTTON | WS_GROUP,307,22,59,10
+    CONTROL         "Highlight",IDC_RADIO_CLM_HILITE,"Button",BS_AUTORADIOBUTTON,307,37,60,10
```
donho, line 90: Change the text to `Highlight Background` please

donho (changes requested):
Please consider the requests.

donho:
@Yaron10 
> But this was solved in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/62de7945d37b4c40817c18e5340cf4d7eb471de9, wasn't it?

Yes, it is. I missed this part.

@shriprem 
Sorry for not seeing the fixed commit.
> By this, if you mean that I should implement the attribute values of "0", "1" & "2" instead of the currently implemented "hide", "show" & "frame", I will still be happy do so. 

Yes, please - it's a cleaner way to me.

shriprem:
@donho, please note that in the [a972ea0](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/a972ea09a8abd4cfb5bc60a0bfd340c74495c129) commit, I made some slight changes in `preference.rc` as compared to the resource text you had provided.

The `Highlight Background` text was too long for the original width of "60". I increased it to "100", which will still fall well within the bounds of the enclosing groupbox. While doing so, just for consistency, I changed the width of the other two radio controls ("None" and "Frame") to "100" as well. I hope the width of "100" will prove sufficient for the translation texts for these new radio controls.

If you differ on these changes, please let me know. I will revert them. Except for these, all your suggested changes are in the [a972ea0](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/a972ea09a8abd4cfb5bc60a0bfd340c74495c129) commit as of now.

Yaron10:
@shriprem,

Thanks again.

> please correct me if I am wrong about the claims on testing observations with the [62de794](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534/commits/62de7945d37b4c40817c18e5340cf4d7eb471de9) builds

:+1: 

@donho,

Are both "currentLineHilitingShow" and "markCurrentLineWith" still needed?
May I ask why?

Thank you.

***

There seems to be some confusion in the current state.
If I'm wrong, there's a compatibility issue with the latest build.

Run the latest build.
Run an older build.
Run the latest build.

The current line is highlighted.

shriprem:
@Yaron10, I will jump in with my assumed (verbose) responses on behalf of @donho.

>Are both "currentLineHilitingShow" and "markCurrentLineWith" still needed?
May I ask why?

Moving forward, the 3-state `currentLineIndicator` attribute will be used to save the config state instead of the previous 2-state `currentLineHilitingShow` attribute. For backward compatibility, the confi.xml will be looked up for the `currentLineHilitingShow` attribute first, and only then the newer `currentLineIndicator` attribute will be looked up if the former is absent in the config.xml file. 

>Run the latest build.
Run an older build.
Run the latest build.
>
>The current line is highlighted.

My assumed response to this is: There is no need for two-way compatibility; just the one-time backward compatibility.

donho:
@Yaron10 
markCurrentLineWith is not needed anymore.
currentLineHilitingShow will be kept for the upgrade use.

> If I'm wrong, there's a compatibility issue with the latest build.
Run the latest build.
Run an older build.
Run the latest build.
The current line is highlighted.

What's your initial current line state?

shriprem:
I think this may be what @Yaron10 is referring to:
1. In the newer build,  set to "Frame" (with any width).
2. Next, in the older build, this will revert back to the default value of "show" since the old build cannot find `currentLineHilitingShow` in the config.xml that has been saved from the newer build in step 1.
3. Next, in the newer build, the setting will now be processed as for the "Highlight" setting.

So, between steps 1 & 3, the setting has changed from "Frame" to "Highlight". Inevitable, because of the loss of information in step 2.

Yaron10:
@shriprem,

Here is my assumption on Don's approach. :)

He wants the GroupBox title and the attribute in config.xml to be the same.
He doesn't like "Current Line Highlighting" as the GroupBox title.

And I'm wondering if the GroupBox title is worth the extra code and confusion.
What's wrong with "Current Line Highlighting" and the old attributes?

@donho,

> markCurrentLineWith is not needed anymore.

I should have asked about "currentLineIndicator" and "currentLineHilitingShow".

> What's your initial current line state?

Hide.

***

@shriprem,

> I think this may be what @Yaron10 is referring to:

No. :)
Just set to "Hide".

shriprem:
>No. :)
Just set to "Hide".

OK. That is still explainable by the differences in the attribute names and values to between the versions. Now the question is: should ensuring a two-way compatibility be a priority?

Yaron10:
> OK. That is still explainable by the differences in the attribute names and values to between the versions.

Sure.

> Now the question is: should ensuring a two-way compatibility be a priority?

It shouldn't be a high priority.
But if the implementation can be simpler + "a two-way compatibility", why not do that?

donho (approved):

Yaron10:
@shriprem,

On my machine, the frame max-width is 5px.
Could you please re-test it?

Thank you.

shriprem:
@Yaron10, I can confirm that there is a distinct change from frame-width of 5 to 6. I will upload a zipped archive of a .wmv file of the screen capture to validate that.

[NPP_Frame_Width.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/8541731/NPP_Frame_Width.zip)

I suspect that maybe you have some magnification setting in the Windows Display Settings that might be causing a setting of 5 to scale to 6px, and thereby maxing out at 5.

Yaron10:
@shriprem,

Thank you for testing and replying.

I'm getting max 5px both on Windows 7 and Windows 10.
Default scaling (100%) and DPI (96) on both machines.

תמונה

תמונה


Your background color is black, correct?
Could you please test it with a "normal text" file?

Yaron10:
@shriprem,

It actually depends on the font size: you can get an 8px frame too. :)

תמונה

So `6` is good. :+1: 

Thanks again.

shriprem:
@Yaron10 , that's a good find! The documentation on Scintilla.org is lacking for SCI_SETCARETLINEFRAME.

The allowed frame width seems to depend on both the font size _and_ the zoom level. At a font size of 11 and 100% zoom level, the max I could get was 6 px for the frame width. But with the same font size and a larger zoom level, I could get larger frame widths. Scintilla seems to be setting a threshold of around 50% of the size in pixels with the font size & zoom levels factored in.

Assuming that most users have font sizes set between 10 and 12, a max frame width of 6 seems to be a conservative choice while still keeping its implementation simple and user-comprehensible in NPP.

Yaron10:
@shriprem,

> The allowed frame width seems to depend on both the font size and the zoom level.

👍 
That makes sense.

> Assuming that most users have font sizes set between 10 and 12, a max frame width of 6 seems to be a conservative choice while still keeping its implementation simple and user-comprehensible in NPP.

I think so too.

Thanks again for your work.

***

In Visual Studio the frame is invisible if there's a selection.
I think it's very useful.

Select a line (including its EOL).
Result:
תמונה

I find it slightly confusing.

Implementing it in NPP doesn't seem too complex. I just wonder how it might affect performance.

I would also like to disable the frame if the line has another marker (e.g. Compare Plugin markers).

I might submit a Feature Request in Scintilla.

alankilborn:
@Yaron10 You find it "very useful" and "slightly confusing" at the same time? :-)

Yaron10:
@alankilborn

The VS behavior is "very useful". The current NPP behavior is "slightly confusing".
Is it clear now? Or still confusing? :)

Yaron10:
https://sourceforge.net/p/scintilla/feature-requests/1436/.

Yaron10:
@shriprem,

I'd like to make it clear that I did not mean your implementation was "slightly confusing".
On the contrary! Your work is excellent, and I truly appreciate it.

***

The purpose of "Current Line Highlighting" is help the user focus on the currently "important line".
Normally it's the line the caret is on.

But in the following case the currently "important line" is NOT the line the caret is on.

Select a line (including its EOL).
Result:

תמונה

Line 2 is selected, _and this is my currently "important line"_.
Line 3 being highlighted, is not helpful in this case but rather confusing/distracting.

Your new Frame **reduces** the confusion, but I'd still rather not have line 3 highlighted (as is the behavior in Visual Studio).

shriprem:
@Yaron10, thank you for your very kind clarification. 👍 

In fact, I have been using Scintilla's caret frame feature for over a year now as the default option with my [Fixed Width Data Visualizer](https://github.com/shriprem/FWDataViz) NPP plugin -- from its very first version. I had to do so since the current line full highlight would have undermined the core feature of the plugin on the current line.

FWDataViz

Since that time, I had always been thinking that NPP was missing a trick by not having the caret frame option as a built-in feature. My day-job obligations, hesitation and sheer apathy prevented me from taking that thought and translating it into action. I am glad I was able to do it now, finally.

The implementation in this PR offers users choice to vary the frame width, whereas my plugin set the frame width at a fixed width of 2. So, with the next release of my plugin, I will be removing the Frame Width toggle control.

On a side note, my FWDataViz plugin was unaffected by the recent Lexilla upgrade. This is since FWDataViz acts directly by utilizing Scintilla's style slots.

I mentioned this all here since it is relevant to your earlier statement [here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11534#issuecomment-1106821965):
>I would also like to disable the frame if the line has another marker (e.g. Compare Plugin markers).

The users of my plugin though would still like to keep the current line frame visibility. Different contexts bring different expectations.

Yaron10:
@shriprem,

Thank you.

Your plugin seems very interesting. 👍 
I've tried to install v2.4.0.2 via PluginsAdmin and also manually, but I get the following message:

תמונה

NPP 8.3.3. I've tried both x32 and x64.
I've already starred it, but I'd love to test it. :)

***

> The users of my plugin though would still like to keep the current line frame visibility. Different contexts bring different expectations.

Sure. It should be configurable.

***

I'll be back in a few hours.

shriprem:
@Yaron10, that's interesting as well as flummoxing for me.

More so, since the 2.4.0.2 release for the plugin has had 3.6K downloads ( GitHub release (latest by SemVer) ) till now but no one else has reported the issue that you are having.

I have opened up an [issue](https://github.com/shriprem/FWDataViz/issues/47) in my plugin repo. If you agree, we can take this issue over [there](https://github.com/shriprem/FWDataViz/issues/47) to keep the conversation on it separate from this PR.

PR #11536, opened 2022-04-14
Adjust padding to fix ShortcutMapper visual glitch

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10358


Yaron10:
@ozone10,

👍 
Thank you.

PR #11537, opened 2022-04-14
update with scintilla 5.2.2 and lexilla 5.1.6

chcg:


donho:
@chcg 
Thank you for update.
Is there any critical bug fix in Scintilla 5.2.2 that you judge it's better to include this PR for v8.4 ?
Or it can wait for being in v8.4.1 ?

rdipardo:
Lexilla 5.1.6 made improvements to the Ruby lexer, and introduced Ruby 3.x syntax: https://www.scintilla.org/LexillaHistory.html

@chcg, don't forget that you can now upgrade Lexilla all by itself; that's one benefit of its separation from the Scintilla source tree. It will be much in the future to fix code highlighting bugs, which the majority of user-generated issues are about.

chcg:
@donho I think there is no major fix in which can't wait for 8.4.1.

donho:
@chcg 
> I think there is no major fix in which can't wait for 8.4.1.

OK. It'll be in the v8.4.1.
But let's not update Scintilla & Lexilla on every release, for the sake of stability of Notepad++.

PR #11540, opened 2022-04-16
CustomFileDialog: Restore the selection after SetFileName

### LINKED ISSUE #11517: In Save dialog, File name is highlighted but it is not where focus is
### Description of the Issue

On Windows, when you go to save a file as  a new type, the file save dialog opens with the filename highlighted in white text with a blue background indicating is it the active field.  If you use the dropdown to select a new file type, the file name is still highlighted but if you start to type a new filename it changes the file type (based on the letter you type) not the name.

When you change the file type, either the focus should be removed from the file name or the file name should retain the focus after the type dropdown is closed.

1. Select File/Save as... with a file that already has a name - the filename textbox has the focus
2. Change the file type - notice the file name text entry is still highlighted 
3. Start typing as though you were entering a new file name.  The type changes

### Expected Behavior

I expect test entry to change the highlighted field (i.e., the file name).  (Most other program I tried remove the focus from the filename text entry box when the extension/type dropdown opens.)

### Actual Behavior

The file type (extension) changes.

### Debug Information

Notepad++ v8.3.3   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 2009
OS Build : 19044.1586
Current ANSI codepage : 1252

mere-human:
The selection is modified indirectly by SetFileName().
So it has to be restored to avoid confusion.

Fix #11517

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -303,11 +303,10 @@ public:
 		generic_string name = getDialogFileName(_dialog);
 		if (changeExt(name, dialogIndex - 1))
 		{
-			// Set file name and restore the previous selection in the edit box.
-			// The selection is modified by SetFileName().
+			// Set the file name and clear the selection in the edit box.
+			// The selection is implicitly modified by SetFileName().
 			DWORD selStart = 0;
 			DWORD selEnd = 0;
-			SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));
 			bool ok = SUCCEEDED(_dialog->SetFileName(name.c_str()));
 			if (ok)
 				SendMessage(_hwndNameEdit, EM_SETSEL, selStart, selEnd);
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -302,7 +302,17 @@ public:
 		_currentType = dialogIndex;
 		generic_string name = getDialogFileName(_dialog);
 		if (changeExt(name, dialogIndex - 1))
-			return SUCCEEDED(_dialog->SetFileName(name.c_str()));
+		{
+			// Set file name and restore the previous selection in the edit box.
+			// The selection is modified by SetFileName().
+			DWORD selStart = 0;
+			DWORD selEnd = 0;
+			SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));
+			bool ok = SUCCEEDED(_dialog->SetFileName(name.c_str()));
+			if (ok)
+				SendMessage(_hwndNameEdit, EM_SETSEL, selStart, selEnd);
+			return ok;
+		}
 		return false;
 	}
 


mere-human:
How to test:

1. Open an existing file
2. Select File > Save As... 
3. Expected: the filename edit box has the focus
4. Change the file type
5. Expected: the file name text entry is not highlighted and doesn't have the focus
6. Expected: the file type dropdown has the focus, can be navigated with arrow keys
7. Uncheck the "Append extension" checkbox
8. Change the file type
9. Expected:  the "Append extension" checkbox is checked
10. Uncheck the "Append extension" checkbox
11. Remove the file extension from the file name
12. Check the "Append extension" checkbox
13. Select a folder and press save
14. Expected: file is saved with the extension

mere-human:
@Yaron10  I'd appreciate if you can test this.

Yaron10:
@mere-human,

👍 
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11517 is fixed.
A similar case where you check "Append extension" *using the mouse* is also fixed.

Thank you for finding the time to work on it during these turbulent times in Ukraine. 

***

> Uncheck the "Append extension" checkbox
Change the file type
Expected: the "Append extension" checkbox is checked

I'm not sure why.
The user might want the "Append extension" preference OFF, but _this time only_ save as "A.cpp".

***

Open the Save dialog with "Append extension" _unchecked_.
Check "Append extension".

**Result:**
An extension is appended.

Open the Save dialog with "Append extension" _checked_.
Uncheck "Append extension".

**Result:**
The file-type is changed but the extension is not removed.

Isn't that somewhat inconsistent?
IOW: if you interpret checking "Append extension" as "in general and for this action too", shouldn't unchecking the box be interpreted the same way?

mere-human:
> #11517 is fixed.
> A similar case where you check "Append extension" _using the mouse_ is also fixed

Great, thanks for testing!



> **Result:** The file-type is changed but the extension is not removed.
> 
> Isn't that somewhat inconsistent? IOW: if you interpret checking "Append extension" as "in general and for this action too", shouldn't unchecking the box be interpreted the same way?

I consider this checkox to work as "Append extension if not already present". It could be really hard to figure out whether what user added after a dot (.) is an extension or just a part of the name. So we decided not to remove it at all. If a user really wants to remove the extension, it can be done manually from edit box. There was a discussion on this I just can't find it.

Yaron10:
@mere-human,

> I consider this checkox to work as "Append extension if not already present"...

👍 
Thanks again for the fix. I appreciate it.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -302,7 +302,17 @@ class FileDialogEventHandler : public IFileDialogEvents, public IFileDialogContr
 		_currentType = dialogIndex;
 		generic_string name = getDialogFileName(_dialog);
 		if (changeExt(name, dialogIndex - 1))
-			return SUCCEEDED(_dialog->SetFileName(name.c_str()));
+		{
+			// Set file name and restore the previous selection in the edit box.
+			// The selection is modified by SetFileName().
+			DWORD selStart = 0;
+			DWORD selEnd = 0;
+			SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));
```
donho, line 310: `selStart` & `selEnd` are always 0 after calling EM_GETSEL.
I also tried the following codes, the same result:
```cpp
			long res = SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));
			selStart = LOWORD(res);
			selEnd = HIWORD(res);
```
_hwndNameEdit is not null and I have no idea why it doesn't work on my Windows11.

donho:
@mere-human 
I hope you're well.

As the comment I left in the code review zone, this PR doesn't work at all on my Windows11 (version 21H2).
Both selStart & selEnd values are 0, even from the returned value of EM_GETSEL call.
`SendMessage(_hwndNameEdit, EM_SETSEL, selStart, selEnd);` call makes no selection, but the focus is still on "Save as type" combobox - when I type the on the keyboard the value changes on the type combobox, and the file extesion changes in "File name" field as well.

@Yaron10 @mere-human 
What's your OS? Windows10? Windows7?

donho:
One bug that I just noticed:
In save as dialog, even I uncheck "Append extension" checkbox, once I change in the combobox of file extension, the old file extension is still replaced by the new one in File name field, and  "Append extension" checkbox is checked.

Can you reproduce it on your OS?

Yaron10:
> but the focus is still on "Save as type" combobox - when I type the on the keyboard the value changes on the type combobox, and the file extesion changes in "File name" field as well.

This is the expected behavior.
Without this PR, the focus is correctly on "Save as type" combobox but the _"File name" is selected_ which is misleading (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11517).

Note: I don't understand @mere-human's code and comments.
Apparently `SendMessage(_hwndNameEdit, EM_SETSEL, 0, 0);` (or some `RemoveSelection`) should fix the issue. 

***

> In save as dialog, even I uncheck "Append extension" checkbox, once I change in the combobox of file extension, the old file extension is still replaced by the new one in File name field, and "Append extension" checkbox is checked.

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11540#issuecomment-1100760028

>  Uncheck the "Append extension" checkbox
    Change the file type
    Expected: the "Append extension" checkbox is checked
>> I'm not sure why.
The user might want the "Append extension" preference OFF, but this time only save as "A.cpp".

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3daf8d4ae979e39ae06bf45ac0cb3a1c849ab7aa/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp#L288

Should be:
```
		if (SUCCEEDED(_dialog->GetFileTypeIndex(&dialogIndex)))
		{
			OnTypeChange(dialogIndex);
		}
```

donho:
>> but the focus is still on "Save as type" combobox - when I type the on the keyboard the value changes on the type combobox, and the file extesion changes in "File name" field as well.

> This is the expected behavior.
Without this PR, the focus is correctly on "Save as type" combobox but the "File name" is selected which is misleading (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11517).

In this case, the PR can reduce to:
```
		{
			// Set file name and restore the previous selection in the edit box.
			// The selection is modified by SetFileName().
			DWORD selStart = 0;
			DWORD selEnd = 0;
			
			bool ok = SUCCEEDED(_dialog->SetFileName(name.c_str()));
			if (ok)
				SendMessage(_hwndNameEdit, EM_SETSEL, selStart, selEnd);
			return ok;
		}
```
The call of  `SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));` is not necessary at all.

Yaron10:
> In this case, the PR can reduce to:...

Unless @mere-human explains his intention and contradicts us, that is correct.

And the PR's title and code-comments should also be modified accordingly.

mere-human:
> @mere-human I hope you're well.
> 
> As the comment I left in the code review zone, this PR doesn't work at all on my Windows11 (version 21H2). 

It works for me on Windows 10. I'll try to find a VM with Windows 11 to check.

----
> 
> The call of `SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));` is not necessary at all.
> 


Yes, it's 0, 0 for most of the times. I don't know when selStart and selEnd could be != 0. But in case it would, we'll restore the correct selection. That was the intention.

mere-human:
@donho 

> `SendMessage(_hwndNameEdit, EM_SETSEL, selStart, selEnd);` call makes no selection, but the focus is still on "Save as type" combobox - when I type the on the keyboard the value changes on the type combobox, and the file extesion changes in "File name" field as well.
> 

That's the point. The point is to reset the selection to what it was before the call to `_dialog->SetFileName(name.c_str())`. Even if the selection was empty. Because the call to `SetFileName()` implicitly modifies selection for some reason.

---

> One bug that I just noticed: In save as dialog, even I uncheck "Append extension" checkbox, once I change in the combobox of file extension, the old file extension is still replaced by the new one in File name field, and "Append extension" checkbox is checked.
> 
> Can you reproduce it on your OS?
> 
> *Edit: Just saw your comment here:*[#11540 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11540#issuecomment-1100695041) _@mere-human From 7-9: how come it's expected behaviour?_

Because if the user selects a file type, it means they want to append the extension. Therefore, the checkbox is checked automatically. If the checkbox is unchecked, the file type is "All Types" which means "don't append the extension". So they are connected.

mere-human:
I've just checked this PR on Windows 11 and it works exactly the same as on Windows 10: the selection from the file name is removed when either the combo box or the checkbox is checked.

donho:
@mere-human 
So it's only about deselecting the file name field.
In this case, The call of `SendMessage(_hwndNameEdit, EM_GETSEL, reinterpret_cast<WPARAM>(&selStart), reinterpret_cast<LPARAM>(&selEnd));` is not necessary at all.

Are we on the same page?

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
mere-human: Done.

Yaron10:
The issue still exists in the following case:

Have a file named "new 1.txt" in the default folder.
Open and edit a new document.
File -> Save.
Press "Save" in the dialog *using the mouse*.

You get the "File already exists" message.
Press "Cancel".

**Result:**
The "File Name" is selected, but the "Save" button has the focus.

Note: this problem exists in some Microsoft programs (e.g. MS Paint) as well.

mere-human:
> 

Should we reopen the current issue or create a new one? Or do you consider it a non-issue?

Issue #11113, opened 2022-02-01
Replace just one more time


amtype:
### Description of the Issue

I'm requesting a new feature in the Replace dialog box: a button that replaces just one more instance of the searched text without continuing the search. The ancient (but wonderful) XyWrite word processor had such an option.

### Steps to Reproduce the Issue

1. In the Replace dialog box I insert text in "Find what" and "Replace with" fields
2. I search and replace until I land on the last instance I want to replace
3. There I want to replace once more and close the Replace dialog box, leaving the cursor at the last replacement

### Expected Behavior

If there already is a way to accomplish "replace once more and stop here", please tell me.

### Actual Behavior

I click on the Replace button and Notepad++ performs the replace and takes me to the next instance of the Find what text--as expected. Before I perform the last desired replacement, I have to write down text surrounding the replacement I'm about to perform so I can search back to find that exact replacement. Alternatively, instead of clicking Replace to make the last desired replacement, I can click Close to close the Replace dialog box and perform the final replacement manually, which is inconvenient.

alankilborn:
This issue has come up before and the somewhat lackluster solution that come out of it was:

![image](https://user-images.githubusercontent.com/77065706/151968862-7134b83d-a6cb-4b55-8a5e-856cf0f42063.png)

amtype:
Thanks! I agree, lackluster and more tedious, but workable without requiring me to write down what I'm replacing.

alankilborn:
Another way to do it is to _Undo_ (Ctrl+z) the last replacement and then _Redo_ (Ctrl+y) it.  This will leave your caret near your desired point.

amtype:
I assumed Ctrl+z would undo all the replacements, not just the last one. Thanks for the tip!

alankilborn:
Duplicate of #3214

donho:
Implemented.

Issue #3214, opened 2017-04-26
feature request: add "replace last" button in search and replace so don't jump past location of last desired replace


dvbrydon:
### Description of the Issue

When in the find and replace dialog, you click replace and it replaces that instance and jumps to the next.  But sometimes you are working on a local function and you come to the last replacement you want to make of that word, even though it appears later in the file.  You click replace, and it jumps you 300 lines down the file to the next instance.  

A replace last button in the dialog would replace that word and leave the cursor at the end of that word.

Ultraedit has this and it saves having to come back up the file to find the line you are working on.

### Steps to Reproduce the Issue

1.   ctrl+f to open find dialog.  click the replace tab.
2.  enter a word to find and a replacement.  click find.  click replace.  it jumps to next instance.

### Expected Behavior

3.  here we need the replace last button so the cursor will stay and not jump to the next instance.

### Actual Behavior

4.  it jumps down the file when I hit replace (correctly) but I have to come back up and find where I was.

donho:
Implemented.

dvbrydon:
This is awesome.  Thank you!!!

PR #11541, opened 2022-04-16
Moved 'Replace: Don't move to the following occurrence' button from Settings to Find and Replace

hadileonardd:
This resolves the issue [#11113](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11113), duplicate of [#3214](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3214).

The 'Replace: Don't move to the following occurrence' checkbox seems to be out of the way in Settings > Preferences > Searching, so I added the button directly in the Find and Replace window where it can be accessed immediately during any replace action.

This requires a design decision to _move_ the button instead of just _copying_ it and I thought it made sense. Is this okay? The side effect of this is that the `config.xml` file will not affect the checkbox anymore, since config.xml specifically affects options that are in settings (this change can be seen below in the `Parameters.cpp` file).

# UI Changes
### Before:
before preferences setting

### After:
after preferences

### Before:
before replace box

### After:
after replace box


alankilborn:
It is customary when coding something that will change the UI, to show a sample screenshot of the new proposed UI.

hadileonardd:
@alankilborn Thank you for the advice, I have added screenshots (before and after) for the UI changes.

alankilborn:
When the _Mark_ tab of the _Find_ dialog is active, _Don't move..._ appears but it shouldn't.
It covers _Purge for each search_.

hadileonardd:
@alankilborn Thanks for noticing the mistake, I believe I've pushed the fixed

alankilborn:
After some more thought on the matter, I'm wondering if this function isn't best as a new action button, i.e., like _Replace_ and _Replace All_, a new button called _Replace and Stop_ should be created and the checkbox removed.  What are the thoughts on this?

Yaron10:
@alankilborn,

> I'm wondering if this function isn't best as a new action button...

👍 
Finding the best wording... :)

@hadileonardd,

👍 
Thank you for your work.

hadileonardd:
@alankilborn @Yaron10 I was taking a look at this and here are my thoughts: how about we keep it this way for now and make a separate issue for changing it to be a button instead of a checkbox if needed?

One reason is that it doesn't seem to work too well with the current code organization. Changing it to a button would make it very similar in code to the original button (`IDREPLACE`), [shown here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L1331-L1350). My next thought would be: ideally, the best practice way to do this should be to just make a small change in that `IDREPLACE` section of code instead of having an entirely separate button with a different ID - because almost everything is the same except for a new boolean variable (e.g. `_replaceStopsWithoutFindingNext` as it is in our current PR) in `_options`, which can then be checked in [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2156) as follows:

     if (replaceOptions._replaceStopsWithoutFindingNext) // (change to check in replaceOptions instead of nppGui)

Unfortunately, though, it seems we are forced to have a separate button with a different ID, given how the [UI code](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/FindReplaceDlg.rc#L27-L84) works. I don't think it's possible with the way the current code is written to make a separate button that also enters and uses [the same `IDREPLACE` code](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L1331-L1350) while enabling a new `_replaceStopsWithoutFindingNext` option in `_options`.

The other reason, maybe more importantly, is that I also am not sure if this function is better as a button instead of a checkbox in the list of search modifying options. It feels to me that having too many "find"/"replace" buttons is a bigger potential issue than having too many options that modify those buttons. In other words, if we want to think long term where there may be a lot more find/replace functionality, I think it would better to have more options that modify the main buttons instead of adding a new button for each new functionality. It seems more easily scalable.

Yaron10:
@hadileonardd,

> how about we keep it this way for now and make a separate issue for changing it to be a button instead of a checkbox if needed?

👍 

You're right, there are pros and cons in each implementation.

Thanks again.

Yaron10:
@hadileonardd,

> how about we keep it this way for now and make a separate issue for changing it to be a button instead of a checkbox if needed?

👍 

You're right, there are pros and cons in each implementation.

Thanks again.

alankilborn:
The "danger" is that it could become an "annoying" feature if it is not a button.
As an example, what if there was no _Replace All_ button, but an "All" checkbox.  User could become annoyed because they didn't notice the state of the checkbox before performing the operation, and it wasn't what they wanted.  Or 50% of the time they want to replace a single match and 50% they want to do all -- having to do 2 steps instead of just 1 can get burdensome, when alternating functionalities.
These are not absolutes, just some thoughts...

Yaron10:
@alankilborn,

We've learned quite recently that our ideas and suggestion are a complete waste of time. :)
So, as Confucius said: "Let a ruler be a ruler". 
Or rather: "Let a supreme leader be a supreme leader".

alankilborn:
>We've learned quite recently that our ideas and suggestions are a complete waste of time. :)
So, as Confucius said: "Let a ruler be a ruler".
Or rather: "Let a supreme leader be a supreme leader".

Well...sure. :-)
But if we try "nudges" to try to make things "better", it has to count for something.  Even if suggestions are discarded, we can feel good about trying.

Ayaykos:
I'd like to chime in about this:

> As an example, what if there was no Replace All button, but an "All" checkbox. User could become annoyed because they didn't notice the state of the checkbox before performing the operation, and it wasn't what they wanted.

Assuming the premise that the existing `Replace All` button is directly comparable to the new button being proposed is true, I agree with the conclusion of this example. However, I think that premise itself is not necessarily true and that `Replace All` is not directly comparable.

I see the `Replace All` button as a more primary/fundamental button that also parallels existing software like the original Notepad itself:
image

For a new user of Notepad++ that comes from Notepad, it might be jarring to see `Replace All` transformed from a primary button in its own right to a checkbox that modifies the usage of the primary `Replace` button. On the other hand, it feels more intuitive that a Notepad feature that is less fundamental like `Match case` is a checkbox that modifies `Replace` instead of being a button like `Replace with matching case`. I think it has to do with the importance/significance of usage of that button.

I see the new proposed functionality of replacing without moving forward as less of a key functionality and it shouldn't be placed as a unique button alongside `Find Next`, `Replace`, and `Replace All`. As a user I think a checkbox that modifies the search would make more sense.

alankilborn:
The stop feature of replace is sort of unique as it modifies the behavior _after_ the prime functionality, not during the functionality, such as case matching.  Thus, while my _Replace All_ example was a bit contrived, there really is no parallel for this new functionality.  For me, as I said, I might use this capability 50% of the time (whereas where it is now, as a _Preference_ I simply ignore it and never use it).  Toggling a checkbox this often, would get annoying to me.

Mateos81:
Buttons are actions, and tick boxes are options/modifiers affecting these actions, right?

donho (changes requested):
I would say the following GUI (as the right checkbox of Find button) is more intuitive: 

image

PowerEditor/src/Parameters.cpp
```diff
@@ -6586,7 +6583,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("muteSounds"), _nppGUI._muteSounds ? TEXT("yes") : TEXT("no"));
 	}
 
-	// <GUIConfig name="Searching" "monospacedFontFindDlg"="no" stopFillingFindField="no" findDlgAlwaysVisible="no" confirmReplaceOpenDocs="yes" confirmMacroReplaceOpenDocs="yes" confirmReplaceInFiles="yes" confirmMacroReplaceInFiles="yes" replaceStopsWithoutFindingNext="no"/>
+	// <GUIConfig name="Searching" "monospacedFontFindDlg"="no" stopFillingFindField="no" findDlgAlwaysVisible="no" confirmReplaceOpenDocs="yes" confirmMacroReplaceOpenDocs="yes" confirmReplaceInFiles="yes" confirmMacroReplaceInFiles="yes" />
```
donho, line 6586: restore above line please

alankilborn:
Is the addition of this checkbox going to play well with macro recording and playback?

donho:
@alankilborn 
> Is the addition of this checkbox going to play well with macro recording and playback?

Yes, it would be great.

donho:
PR is closed due to no reply of code review request.

alankilborn:
>PR is closed due to no reply of code review request

That's disappointing.
What are the chances that @donho will finish coding this feature himself? :-)

donho:
@alankilborn 
Please do a feature request firstly, then we'll see.

alankilborn:
> Please do a feature request firstly, then we'll see.

Done...see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11696

PR #11557, opened 2022-04-20
exclude "Search result" from sorting of languages in Style Configurator 

CennoxX:
With https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2474cbeafd2db6483bfacd50768ec7ec07aa0be3 the listing of languages in Style Configurator was sorted in alphabetic order. In this process "Search Result" was sorted too, before it was placed at the bottom. This PR excludes "Search Results" from the sorting of languages and places it at the bottom as @donho wanted to do here: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2474cbeafd2db6483bfacd50768ec7ec07aa0be3#commitcomment-70894280

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -508,6 +508,10 @@ private :
 
 struct SortLexersInAlphabeticalOrder {
 	bool operator() (LexerStyler& l, LexerStyler& r) {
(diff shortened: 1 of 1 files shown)


Yaron10:
@CennoxX,

Great.
Thank you for your work. 👍

PR #11560, opened 2022-04-20
Fix behavior when using notepad image hijack with npp

### LINKED ISSUE #7902: When using Notepad++ as notepad replacement, it doesn't open a file without extension
### Description of the Issue
When using Notepad++ as replacement for `notepad.exe`, the behavior for opening files without extension is weird.

Using Notepad++ v7.8.4 x64
1. Replace default `notepad.exe` with Notepad++ (See [Notepad Replacement](https://npp-user-manual.org/docs/other-resources/))
2. Create a file **without** an extension, e.g. README
3. Try to open that file with `notepad.exe`
4. Error (it seems that Notepad++ tries to add `.txt` if ther is no file extension?!)

### Expected Behavior
Open files without extension too when using `notepad.exe` (e.g. README, LICENSE)

#### Proposed change
Add an additional check if the file without a file extension does exist or not before adding a `.txt` extension.
The [affected lines of code](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/8c6dc79df31fa4d97fab5fe7b07e366fb55f3dff/PowerEditor/src/winmain.cpp#L163-L166).
>		if ( *PathFindExtension(str.c_str()) == '\0' )
>		{
>			str.append(TEXT(".txt")); // If joined path has no extension, Notepad adds a .txt extension
>		}

### Actual Behavior
Notepad++ looks, if the same file with the `.txt` extension is available
1. if yes - it opens the wrong file (e.g. README.txt **instead** of README)
2. if no - it asks to create the file with a `.txt` extension

grafik

grafik

*Note: If I open any file without extension directly with Notepad++, it works as expected!*

### Debug Information
Notepad++ v7.8.4   (64-bit)
Build time : Jan 29 2020 - 01:31:03
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1909
OS Build : 18363.592

kapsiR:
This change aligns the behavior of the original notepad.exe with notepad++.exe when it's used as image hijack for notepad.exe.

Fixes #7902

#### New behavior:
*Important: This affects the notepad.exe image hijack with notepad++.exe, not notepad++.exe directly*  

1. File without extension exists:  
    The file will now be opened as expected

2. File without extension doesn't exist:  
    Notepad++ tries to open the file with a `.txt` extension  
    If this file also doesn't exist, it asks to create a file with the `.txt` extension

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -162,9 +162,9 @@ ParamVector convertParamsToNotepadStyle(PWSTR pCmdLine)
 	if ( *pCmdLine != '\0' )
 	{
 		generic_string str(pCmdLine);
-		if ( *PathFindExtension(str.c_str()) == '\0' )
+		if ( *PathFindExtension(str.c_str()) == '\0' && !PathFileExists(str.c_str()) )
 		{
-			str.append(TEXT(".txt")); // If joined path has no extension, Notepad adds a .txt extension
+			str.append(TEXT(".txt")); // If joined path has no extension and there is no file without an extension, Notepad adds a .txt extension
 		}
 		params.push_back(std::move(str));
 	}


kapsiR:
@donho Sorry for bothering you here, but may you include this in one of the next releases? (I know image hijacking does not work on Win11 without e.g. uninstalling the Notepad app, but this has nothing to do with this fix here?) 

Thanks!

donho:
The bug has been fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6c3659fe7300992ae7a3cd32410377bd88ac263a.

Issue #11118, opened 2022-02-02
Tiny Feature Request: Show number of results found


curya:
When searching for a word in Chrome, for example, you get something that looks like this:
![VIuRUC4](https://user-images.githubusercontent.com/71131343/152148385-07f2d0a8-2a1d-490b-b99a-b779413922a9.png)

And it changes numbers as you go through them:
![4zVjonJ](https://user-images.githubusercontent.com/71131343/152148636-3c52dd35-6506-4c1a-9aee-3dbd24034e05.png)

I know you already have features like "Count," "Mark," and "Find All in Current Document," but it would be very useful to see what number I'm on in the Find window as I search through the results.

alankilborn:
That works well for Chrome, where text is constant.  In an editor, after a find users typical make changes producing more/less matches with their changes.

So N++ would have to continually (with every text change by the user) scan an entire file (potentially a huge file) to make sure that the YY in XX/YY remains correct, while that box remains showing.

I wouldn't say it is a "tiny" feature request.  It's a feature request with possibly big implications and impact to the software.

curya:
Doesn't "Count" basically do that already...? I would just like the information displayed when I press "Find Next."

alankilborn:
>I would just like the information displayed when I press "Find Next."

I think it is a good feature request; it could update the Find window's status bar with every Find Next press.

silver5171:
What if the numerator becomes a '?' when a change comes through, and then when the arrow or "find next" is pressed N++ could scan the entire file, which would then change both the numerator and denominator?

Pros:
- deferred work so that N++ does not need to continually scan files
- Gives users extra information on what number they are on in the common case

Cons:
- In a naive implementation the denominator would not change, which may cause some confusion until "find next" is clicked and the file rescanned (unless we did something on a file change that would display something like '?/?')

vinsworldcom:
I'd love to see it on the incremental find - which already shows the "matches":

![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/9657169/a4dae774-4a1c-429d-8628-3ccb9963b79e)

I've had a quick look and not sure how easy it is to add this.  Maybe in this:

notepad-plus-plus/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp:
```
[...]

PR #11570, opened 2022-04-22
Add location indicator to find window

silver5171:
Issue #11118 
Added a relative location each time find replace is clicked. This gives users the ability to see more information about what match they are at and how many matches there are in an easy to view way.
The indicator is in the text in the bottom portion of the screen, and updates when a user clicks "Find Next".
ManualTest2
ManualTest1

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "cmake.configureOnOpen": false
+}
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -20,7 +20,6 @@
 #include "ScintillaEditView.h"
 #include "Notepad_plus_msgs.h"
 #include "localization.h"
-#include "Common.h"
 #include "Utf8.h"
 
 using namespace std;
@@ -883,6 +882,10 @@ std::mutex findOps_mutex;
 
 intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam)
 {
+	bool firstFindNext = true;
+	const TCHAR* prevString = _options._str2Search.c_str();
+	int total = 0;
+	int current = 0;
 	switch (message)
 	{
 		case WM_GETMINMAXINFO:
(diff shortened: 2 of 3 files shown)


chcg:
@silver5171 Your PR doesn't compile, see https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/43318570

donho (changes requested):
As @chcg comment : https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11570#issuecomment-1107891733
Your code cannot be compiled. 
Please fix it so the code review can be done.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4625,15 +4650,6 @@ intptr_t CALLBACK FindIncrementDlg::run_dlgProc(UINT message, WPARAM wParam, LPA
...
 
-		case WM_PRINTCLIENT:
-		{
-			if (NppDarkMode::isEnabled())
-			{
-				return TRUE;
-			}
-			break;
-		}
-
```
ozone10, line 4636: Why? This will make buttons have white corners.
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11506

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -5124,15 +5140,6 @@ LRESULT APIENTRY Progress::wndProc(HWND hwnd, UINT umsg, WPARAM wparam, LPARAM l
...
```
ozone10, line 5135: Again white corners here.

ozone10:
Maybe also add to incremental search.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "cmake.configureOnOpen": false
+}
```
ozone10, line 3: Remove

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: Deletion of these lines makes regression.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: Deletion of these lines makes regression.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -119,7 +119,7 @@ friend class FindReplaceDlg;
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
 	void addSearchHitCount(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection);
-	void add(FoundInfo fi, SearchResultMarking mi, const TCHAR* foundline, size_t totalLineNumber);
+	void add(FoundInfo fi, SearchResultMarking mi, const TCHAR* foundline);
```
donho, line 122: Deletion of this argument makes regression.

donho (commented):
This PR cannot be accepted due to the following reasons:
1. The deletions in the PR make 2 regressions (2 features are undone).
2. There are the performance issue in any quite large file (on each Find Next).

alankilborn:
@donho said:

>There are the performance issue in any quite large file (on each Find Next).

Someone mentioned _Incremental Search_.
I see that currently it displays the match count for the entire file, although it doesn't tell you the index of the current match.
But, my question is, doesn't that also have performance problems on a large file, and yet that exists currently?

donho:
@alankilborn 
> But, my question is, doesn't that also have performance problems on a large file, and yet that exists currently?

Just tested with the same file. Yes, there is the performance issue in Incremental search indeed.
I was not aware of the such issue, I will see what I can do about it.

Could you create an issue for it please?

alankilborn:
>Could you create an issue for it please?

Yes, I will.

>Just tested with the same file

But I lack this file that provides good example for this problem, so I will just create the issue as a theoretical one and if you want to attach the file...

PR #11572, opened 2022-04-23
Fix "Find progress bar" UI clipping

SinghRajenM:
image

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -5056,10 +5056,12 @@ int Progress::createProgressWindow()
 
 	int xStartPos = dpiManager.scaleX(5);
 	int yTextPos = dpiManager.scaleY(5);
+	auto ctrlWidth = width - widthPadding - xStartPos;
+
 	_hPText = ::CreateWindowEx(0, TEXT("STATIC"), TEXT(""),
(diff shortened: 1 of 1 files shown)


donho:
@SinghRajenM 
Sorry, I don't follow what you've fixed:
The image "Before" seems correct to me... Or I'm missing somethings.
Could you give me a hint?

alankilborn:
>I don't follow what you've fixed

This is probably because proper procedure wasn't followed.
There's supposed to be an _issue_ discussing and justifying a change, before a PR is created.
@SinghRajenM should know this, being a long-time contributor.

SinghRajenM:
> @SinghRajenM Sorry, I don't follow what you've fixed: The image "Before" seems correct to me... Or I'm missing somethings. Could you give me a hint?

Basically for both the controls (text and progress bar), width is going beyond the total width of the dialog due to starting position of controls (`xStartPos`). 

However, text control is able to adjust itself due to `SS_PATHELLIPSIS` property. But progress bar overflows (see, it touches dialog border in before image). With this change, same margin is provided end the right end too. Refer the back line in below screenshot - 

image

SinghRajenM:
> > I don't follow what you've fixed
> 
> This is probably because proper procedure wasn't followed. There's supposed to be an _issue_ discussing and justifying a change, before a PR is created. @SinghRajenM should know this, being a long-time contributor.

Yeah, I understand that. My bad.
But this was cosmetic change with less severity. So thought of submitting a PR directly.

Issue #11576, opened 2022-04-25
Re-invoking a mono-instance NPP using new command line parameters issue with plugin query & Debug Info display


shriprem:
### Issue
Some users prefer to leave an instance of Notepad++ running in the background to avoid the _brief_ wait for the NPP initial launch. Then they have their application/script file continually re-invoke Notepad++ to open some files in a specific way by using NPP's command line parameters. When they do so:
1. The fresh command line parameters are not being displayed in NPP's _Debug Info..._ dialog.
2. More importantly, these fresh command line parameters are not being made available to plugins like PythonScript for custom command line parameter processing (with the `-z` option).

### Steps to reproduce the issue
1. Have Notepad++ set at: _Default (mono-instance)_ under the _Multi-instance & Date_ section in Preferences.
3. Leave an instance of Notepad++ open or minimized.
4. Then launch Notepad++ with some new parameters either from a command line or from a batch file.
5. View the _Debug Info..._ dialog in NPP.  The new command line parameters are not being displayed in this dialog.
6. Use the **PythonScript** plugin console and execute: `notepad.getCommandLine()`. This will only report the original command line used to invoke the initial NPP process. This is since PythonScript is using [GetCommandLine](https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinea) Windows API.
![image](https://user-images.githubusercontent.com/45252729/136254902-711b3678-3e9f-4d86-8a17-93a9b9b5e565.png)

### What is needed
1. Add a new message **NPPM_GETCURRENTCMDLINE** to the NPP API. The plugin authors can utilize this query to obtain the current command line string, and use it to process custom command line parameters that are being passed with the `-z` option.
2. Optionally, a new message **NPPM_GETINITIALCMDLINE** to obtain the initial command line string can also be added to the NPP API for complementarity.
3. Add a new notification **NPPN_CMDLINECHANGED** to the NPP API. When Notepad++ has been re-invoked, it can notify the plugins that a fresh command line string is available.
4. When the most recent command line string used to re-invoke NPP is not an empty string, display both the "Initial Command Line" and the "Current Command Line" in the _Debug Info..._ dialog. This will be useful in providing technical support for NPP users.

### Additional Info
1. Please see my [comments](https://github.com/shriprem/Goto-Line-Col-NPP-Plugin/issues/11#issuecomment-936863058) in response to a user change request for my GotoLineCol plugin.
[...]

Issue #10602, opened 2021-09-28
[Feature] Command Line: Open file in Notepad++ by specifying the column (byte) position


pidgeon777:
![image](https://user-images.githubusercontent.com/13739892/135083053-013ca8c4-32d4-4f1a-8944-832a8c885876.png)

With `-c` it is possible to scroll to the wanted column position, but not in Bytes.

My idea would be to add another argument so scroll to the wanted column in Bytes.

By doing so, the equivalent width in spaces of the tab character wouldn't be significant for the calculation of the proper column position.

PR #11589, opened 2022-04-27
Initial & current command line parameters query API & Debug Info text refresh

shriprem:
Fixes: #11576 (Re-invoking a mono-instance NPP using new command line parameters issue with plugin query & Debug Info display) 
Fixes: #11633 (NPP crashes when re-invoked with -udl command line parameter with more than 7 characters as its value)

Additionally, this PR may also allow closing of issue #10602. With the additions to NPP messaging & notification API in this PR, GotoLineCol plugin will be able to meet the user's requirement. See [here](https://github.com/shriprem/Goto-Line-Col-NPP-Plugin/issues/11#issuecomment-935639520).

### Screenshots with this PR implementation:

1. When a fresh command line param string is not an empty string, _Debug Info..._ will display both the initial and current command lines.
image

2. When a fresh command line param string is either not available or empty, _Debug Info..._ will display just a command line (keeping current behavior for this use case).
image

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -281,8 +281,8 @@ struct CmdLineParamsDTO
 	intptr_t _pos2go = 0;
 
 	LangType _langType = L_EXTERNAL;
+	generic_string _udlName;
 
-	wchar_t _udlName[MAX_PATH];
 	wchar_t _pluginMessage[MAX_PATH];
 
 	static CmdLineParamsDTO FromCmdLineParams(const CmdLineParams& params)
@@ -300,7 +300,7 @@ struct CmdLineParamsDTO
 		dto._pos2go = params._pos2go;
 		
 		dto._langType = params._langType;
-		wcsncpy(dto._udlName, params._udlName.c_str(), MAX_PATH);
+		dto._udlName = params._udlName;
 		wcsncpy(dto._pluginMessage, params._pluginMessage.c_str(), MAX_PATH);
 		return dto;
 	}
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -248,6 +248,7 @@ struct CmdLineParams
 	LangType _langType = L_EXTERNAL;
 	generic_string _localizationPath;
 	generic_string _udlName;
+	generic_string _pluginMessage;
 
 	generic_string _easterEggName;
 	unsigned char _quoteType = 0;
@@ -280,7 +281,9 @@ struct CmdLineParamsDTO
 	intptr_t _pos2go = 0;
 
 	LangType _langType = L_EXTERNAL;
-	generic_string _udlName;
+
+	wchar_t _udlName[MAX_PATH];
+	wchar_t _pluginMessage[MAX_PATH];
 
 	static CmdLineParamsDTO FromCmdLineParams(const CmdLineParams& params)
 	{
@@ -297,7 +300,8 @@ struct CmdLineParamsDTO
 		dto._pos2go = params._pos2go;
 		
 		dto._langType = params._langType;
-		dto._udlName = params._udlName;
+		wcsncpy(dto._udlName, params._udlName.c_str(), MAX_PATH);
+		wcsncpy(dto._pluginMessage, params._pluginMessage.c_str(), MAX_PATH);
 		return dto;
 	}
 };
@@ -1481,13 +1485,11 @@ public:
 	{
 		_cmdLineParams = cmdLineParams;
 	}
+
 	const CmdLineParamsDTO & getCmdLineParams() const {return _cmdLineParams;};
 
 	const generic_string& getCmdLineString() const { return _cmdLineString; }
 	void setCmdLineString(const generic_string& str) { _cmdLineString = str; }
-
-	const generic_string& getCmdLineStringCurrent() const { return _cmdLineStringCurrent; }
-	void setCmdLineStringCurrent(const generic_string& str) { _cmdLineStringCurrent = str; }
 
 	void setFileSaveDlgFilterIndex(int ln) {_fileSaveDlgFilterIndex = ln;};
 	int getFileSaveDlgFilterIndex() const {return _fileSaveDlgFilterIndex;};
@@ -1734,7 +1736,6 @@ private:
 
 	CmdLineParamsDTO _cmdLineParams;
 	generic_string _cmdLineString;
-	generic_string _cmdLineStringCurrent;
 
 	int _fileSaveDlgFilterIndex = -1;
 
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -117,6 +117,7 @@ const int LANG_INDEX_TYPE7 = 8;
 const int COPYDATA_PARAMS = 0;
 const int COPYDATA_FILENAMESA = 1;
 const int COPYDATA_FILENAMESW = 2;
+const int COPYDATA_FULL_CMDLINE = 3;
 
 #define PURE_LC_NONE	0
 #define PURE_LC_BOL	 1
@@ -1485,6 +1486,9 @@ public:
 	const generic_string& getCmdLineString() const { return _cmdLineString; }
 	void setCmdLineString(const generic_string& str) { _cmdLineString = str; }
 
+	const generic_string& getCmdLineStringCurrent() const { return _cmdLineStringCurrent; }
+	void setCmdLineStringCurrent(const generic_string& str) { _cmdLineStringCurrent = str; }
+
 	void setFileSaveDlgFilterIndex(int ln) {_fileSaveDlgFilterIndex = ln;};
 	int getFileSaveDlgFilterIndex() const {return _fileSaveDlgFilterIndex;};
 
@@ -1730,6 +1734,7 @@ private:
 
 	CmdLineParamsDTO _cmdLineParams;
 	generic_string _cmdLineString;
+	generic_string _cmdLineStringCurrent;
 
 	int _fileSaveDlgFilterIndex = -1;
 
--- a/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
+++ b/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
@@ -145,170 +145,7 @@ intptr_t CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM 
 	{
 		case WM_INITDIALOG:
 		{
-			NppParameters& nppParam = NppParameters::getInstance();
-
 			NppDarkMode::autoSubclassAndThemeChildControls(_hSelf);
-
-			// Notepad++ version
-			_debugInfoStr = NOTEPAD_PLUS_VERSION;
-			_debugInfoStr += nppParam.archType() == IMAGE_FILE_MACHINE_I386 ? TEXT("   (32-bit)") : (nppParam.archType() == IMAGE_FILE_MACHINE_AMD64 ? TEXT("   (64-bit)") : TEXT("   (ARM 64-bit)"));
-			_debugInfoStr += TEXT("\r\n");
-
-			// Build time
-			_debugInfoStr += TEXT("Build time : ");
-			generic_string buildTime;
-			WcharMbcsConvertor& wmc = WcharMbcsConvertor::getInstance();
-			buildTime += wmc.char2wchar(__DATE__, CP_ACP);
-			buildTime += TEXT(" - ");
-			buildTime += wmc.char2wchar(__TIME__, CP_ACP);
-			_debugInfoStr += buildTime;
-			_debugInfoStr += TEXT("\r\n");
-
-#if defined(__GNUC__)
-			_debugInfoStr += TEXT("Built with : GCC ");
-			_debugInfoStr += wmc.char2wchar(__VERSION__, CP_ACP);
-			_debugInfoStr += TEXT("\r\n");
-#elif !defined(_MSC_VER)
-			_debugInfoStr += TEXT("Built with : (unknown)\r\n");
-#endif
-
-			// Binary path
-			_debugInfoStr += TEXT("Path : ");
-			TCHAR nppFullPath[MAX_PATH];
-			::GetModuleFileName(NULL, nppFullPath, MAX_PATH);
-			_debugInfoStr += nppFullPath;
-			_debugInfoStr += TEXT("\r\n");
-
-			// Command line as specified for program launch
-			_debugInfoStr += TEXT("Command Line : ");
-			_debugInfoStr += nppParam.getCmdLineString();
-			_debugInfoStr += TEXT("\r\n");
-
-			// Administrator mode
-			_debugInfoStr += TEXT("Admin mode : ");
-			_debugInfoStr += (_isAdmin ? TEXT("ON") : TEXT("OFF"));
-			_debugInfoStr += TEXT("\r\n");
-
-			// local conf
-			_debugInfoStr += TEXT("Local Conf mode : ");
-			bool doLocalConf = (NppParameters::getInstance()).isLocal();
-			_debugInfoStr += (doLocalConf ? TEXT("ON") : TEXT("OFF"));
-			_debugInfoStr += TEXT("\r\n");
-
-			// Cloud config directory
-			_debugInfoStr += TEXT("Cloud Config : ");
-			const generic_string& cloudPath = nppParam.getNppGUI()._cloudPath;
-			_debugInfoStr += cloudPath.empty() ? _T("OFF") : cloudPath;
-			_debugInfoStr += TEXT("\r\n");
-
-			// OS information
-			HKEY hKey;
-			DWORD dataSize = 0;
-			
-			TCHAR szProductName[96] = {'\0'};
-			TCHAR szCurrentBuildNumber[32] = {'\0'};
-			TCHAR szReleaseId[32] = {'\0'};
-			DWORD dwUBR = 0;
-			TCHAR szUBR[12] = TEXT("0");
-
-			// NOTE: RegQueryValueExW is not guaranteed to return null-terminated strings
-			if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, TEXT("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion"), 0, KEY_READ, &hKey) == ERROR_SUCCESS)
-			{
-				dataSize = sizeof(szProductName);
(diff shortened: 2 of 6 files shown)


PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -381,10 +218,191 @@ void DebugInfoDlg::doDialog()
...
+	}
+
+	// Plugins
+	_debugInfoStr += TEXT("Plugins : ");
+	_debugInfoStr += _loadedPlugins.length() == 0 ? TEXT("none") : _loadedPlugins;
+	_debugInfoStr += TEXT("\r\n");
+
+	// Set Debug Info text and leave the text in selected state
+	::SetDlgItemText(_hSelf, IDC_DEBUGINFO_EDIT, _debugInfoStr.c_str());
+	::SendDlgItemMessage(_hSelf, IDC_DEBUGINFO_EDIT, EM_SETSEL, 0, _debugInfoStr.length() - 1);
```
shriprem, line 403: Line 403 selects the entire Debug Info text string after refreshing it. This selection action preserves the exiting behavior in NPP.

donho:
According [your need](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11589), I would say what it should be done is : 
1. an additional Notepad++ cmd argument for plugin: `-pluginMessage="SPECIFIC_PLUGIN_MESSQGE"` where SPECIFIC_PLUGIN_MESSQGE is a string which can contain white space. In your case, the whole argument will be `-pluginMessage="-z"`.
2. NPPN_CMDLINEPLUGINMSG notification: which will be triggered by `-pluginMessage` and plugin authors can get their specific message via hwndFrom field:

```cpp
	//scnNotification->nmhdr.code = NPPN_CMDLINEPLUGINMSG;
	//scnNotification->nmhdr.hwndFrom = pluginMessage; //where pluginMessage is pointer of type TCHAR
	//scnNotification->nmhdr.idFrom = 0;
```
No need to distinguish between "initial cmdline" and "current cmdline", since plugins will get all message via `-pluginMessage`.

What do you think?

shriprem:
@donho, we will need access not just to the part of the command line that follows the `-z` argument; but also to the file names in the new command line. Since the user is expecting to position all those multiple files at the same column position using byte counts instead of character counts.

I initially thought that there could be multiple arguments following a single `-z` on the command line.  Something like:
`NPP_DIR\notepad++.exe file1 file2 file3 -n148 -c861 -z arg1 arg2 arg3`

But it turns out that NPP considers only the `arg1` as a custom argument and it prompts whether to create new files with names of `arg2` & `arg3`. To avoid that, each of those args have to be preceded by their own `-z`. So, the above line should actually be:
`NPP_DIR\notepad++.exe file1 file2 file3 -n148 -c861 -z arg1 -z arg2 -z arg3`

So, we need to consider that there could be multiple -z arguments in the command line. Where, each -z argument could be potentially targeted at a different plugin by the users.

Currently, the -z arguments are being allowed anywhere on the command line. That is, NPP is not minding if the file names, the NPP-specific arguments and the -z arguments are not grouped but mixed up, and in any order.

In view of all these, it will be much simpler if the new NPPM message passes the full command line and let the plugin worry about parsing it. It would be almost equivalent to what the plugin authors are getting with the Windows GetCommandLine API for the initial command line.

In the issue #11576 for this PR, I had indicated:
>Optionally, a new message NPPM_GETINITIALCMDLINE to obtain the initial command line string can also be added to the NPP API for complementarity.

So only the **_latest_** command line is actually what is needed. The initial command line is optional. I just threw it in their as an option for complementarity sake. The only minor difference would be that the GetCommandLine will include the NPP executable path but the new NPP message would skip the executable part of the command line and include the rest of the command line.

Yes, we could possibly use the NPPN_CMDLINEPLUGINMSG notification to pass the new command line in its entirety. But I was thinking if it is a good practice to be passing an LPWSTR instead of a window handle in a structure member named `hwndFrom`.

Also, what are your thoughts on the changes in this PR for the Debug Info dialog? How will it be affected by the changes you have in mind?

Thank you.

donho:
@shriprem 
> we will need access not just to the part of the command line that follows the -z argument; but also to the file names in the new command line. Since the user is expecting to position all those multiple files at the same column position using byte counts instead of character counts.

I understand your need. However, other plugin authors may not need the whole command line argument.
OTOH `-z` is not documented and I wonder if it's beneficial to the plugin authors.
My goal is provide a simple & complete API but not specific for 1 plugin's need.

My 2nd suggestion is:
`-pluginMessage="arg1 arg2 arg3"` generates a notification `NPPN_CMDLINEPLUGINMSG` which provides "arg1 arg2 arg3" (without quote) in nmhdr.hwndFrom.
If the plugin need more information, the message `NPPM_GETLASTCMDLINE` can be sent for getting the whole last command line.
In this way the API is simplified, clean/easy for its usage, and covers all need.
What do you think about it?

Regarding the Debug info, we can show the last CMD, but showing both initial and last cmd is not necessary IMO.

shriprem:
@donho

This is what I clearly understand from your suggestions: 
1. Retain only `NPPM_GETLASTCMDLINE`. Drop `NPPM_GETINITIALCMDLINE`.  [_Decision to keep `NPPM_GETLASTCMDLINE` is fantastic._]
2. Show only the last command line in Debug Info.

But I am not clear on the format of the text that needs to be embedded with nmhdr.hwndFrom. Is it the return from `nppParam.getCmdLineParams()`? Please clarify or provide code. Thanks!

shriprem:
Adding to my earlier question:
>But I am not clear on the format of the text that needs to be embedded with nmhdr.hwndFrom. Is it the return from nppParam.getCmdLineParams()? Please clarify or provide code. Thanks!

Or, should the code for generating `NPPN_CMDLINECHANGED` notification moved inside the `case COPYDATA_PARAMS:`?

donho:
@shriprem 
> But I am not clear on the format of the text that needs to be embedded with nmhdr.hwndFrom. Is it the return from nppParam.getCmdLineParams()? Please clarify or provide code. Thanks!

It's the notification about which I was talking:
```
	//scnNotification->nmhdr.code = NPPN_CMDLINEPLUGINMSG;
	//scnNotification->nmhdr.hwndFrom = pluginMessage; //where pluginMessage is pointer of type TCHAR
	//scnNotification->nmhdr.idFrom = 0;
```

> Also, should the code for generating NPPN_CMDLINECHANGED notification moved inside the case COPYDATA_PARAMS:?

On each "notepad++.exe -xxArg1 -ww=Arg2 -aaArg3" all the flags/options will be in "cmdLineParams" then pass the fullfilled structure via COPYDATA_PARAMS case. So `notepad++.exe -pluginMessage="arg1 arg2 arg3"` should follow the other options' schema to pass the value "arg1 arg2 arg3" into "cmdLineParams".
But the processing all the data of "cmdLineParams" should be in `loadCommandlineParams`, so in case of the value of pluginMessage is not empty or null, the notification NPPN_CMDLINEPLUGINMSG should be procasted in the same processing cmdLineParams function (`loadCommandlineParams`).

Feel free to ask if I'm not clear or  you have other questions.

shriprem:
@donho 
Are you suggesting a change from the [currently documented](https://npp-user-manual.org/docs/command-prompt/) `-z` option to `-pluginMessage="arg1 arg2 arg3"` format for placing the plugins options?

donho:
@shriprem 
`-z` was done for Notepad replacement, but not for plugin originally:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f0821d983593a0d3e1afcf5f61b1937c4f16be50

& see:
https://npp-user-manual.org/docs/other-resources/

> Are you suggesting a change from the [currently documented](https://npp-user-manual.org/docs/command-prompt/) -z option to -pluginMessage="arg1 arg2 arg3" format for placing the plugins options?

Yes, I think my suggestion above is better then `-z` because:
1. It's explicit by its naming '-pluginMessage' for the usage and the maintenance.
2. it's easy to get whichever form of argument(s) plugins need, and more flexible (NPPN_CMDLINECHANGED can be skipped).
3. it's more clear for its usage. 

Let's keep `-z` undocumented in the future and use `-pluginMessage="arg1 arg2 arg3"` mainly officially.

@pryrt FYI

shriprem:
Thanks @donho.

The implementation is much clearer now. Although, I will need a bit more time to implement it. May still have questions for you later as I move further with implementation.

shriprem:
@donho, I have completed the implementation of `-pluginMessage` parameter for plugins per your suggestion.

While doing so, I was trying to follow the recently added `-udl` parameter as a model pattern. I ran into crash issues with my initial attempt since I had declared a new member `generic_string _pluginMessage;` just like the preceding `generic_string _udlName;` member. After spending quite some time to see what was causing the crashes, I inferred that `CmdLineParamsDTO` is a POD class as noted clearly in the comment.

In fact, the crashes can occur even with the `-udl` parameter usage since NPP 8.1.2. I have opened issue #11633. (I will also edit the opening comment in this PR to reference that new issue from this PR.)

I have resolved the issue by converting both `generic_string _udlName` and `generic_string _pluginMessage` members to:
```C++
	wchar_t _udlName[MAX_PATH];
	wchar_t _pluginMessage[MAX_PATH];
```

With that change, I had to use `wscncpy` to copy the incoming string values in `CmdLineParamsDTO FromCmdLineParams(...)`.

Few more notes:
1. I am using `scnN.nmhdr.idFrom` instead of the suggested `scnN.nmhdr.hwndFrom` to pass the `-pluginMessage` param content to plugins. I had trouble finding a proper cast to the `<void *>` type.
2. The `-pluginMessage` param starts with a 'p' just like the `-p` parameter for positioning. Due to this overlap, I had to scan for `-pluginParam` before `-p`. There is this commment:
```C++
// getNumberFromParam should be run at the end, to not consuming the other params
```
But there are three other Params (`-settingsDir`, `-titleAdd` & `-udl`) that seem to have ignored this note. They have been lucky that they didn't have the initial letter overlap like the new param for this PR.

3. I am also setting the focus back on the debug info multiline control. This ensures that the multiline edit opens fully highlighted when the Debug Info dialog is reopened within the same session. This will be needed if the dialog has been closed earlier by clicking the _OK_ button -- which causes the focus to shift away from the mle control.
Finally, a sample clip of the Debug Info dialog with this change:
image

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -381,10 +218,181 @@ void DebugInfoDlg::doDialog()
...
+	::SetFocus(::GetDlgItem(_hSelf, IDC_DEBUGINFO_EDIT));
```
shriprem, line 393: * Line 392: Selects the entire Debug Info text string after refreshing it. This selection action preserves the existing behavior in NPP.

* Line 393: Set the focus back on the debug info multiline control. This ensures that the multiline edit opens fully highlighted when the Debug Info dialog is reopened within the same session. This will be needed if the dialog has been closed earlier by clicking the OK button, which causes the focus to shift away from the multilineedit control. **_This line is an enhancement of the existing behavior._**

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -145,171 +145,8 @@ intptr_t CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 	{
 		case WM_INITDIALOG:
 		{
-			NppParameters& nppParam = NppParameters::getInstance();
-
```
shriprem, line 149: Lines 148 & (152 thru' 312) used for constructing the Debug Info text has been moved out as is from `WM_INITDIALOG` into the new function `refreshDebugInfo()`. This new function will be called from `DebugInfoDlg::doDialog()`. This refactoring allows for the Debug Info text to be refreshed when the command line string changes.

PowerEditor/src/winmain.cpp
```diff
@@ -451,17 +455,26 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
...
+		if (pluginMessage.length() >= 2)
+		{
+			if (pluginMessage.front() == '"' && pluginMessage.back() == '"')
+			{
+				pluginMessage = pluginMessage.substr(1, pluginMessage.length() - 2);
+			}
+		}
+		cmdLineParams._pluginMessage = pluginMessage;
+	}
+
```
shriprem, line 470: The scan for `-pluginMessage=` will have to precede that for `-p`, which follows on line 474 below. Otherwise, the scan for `-p` will consume the `-plugin=` param because of the smaller string match.

PowerEditor/src/Parameters.h
```diff
@@ -279,7 +281,9 @@ struct CmdLineParamsDTO
 	intptr_t _pos2go = 0;
 
 	LangType _langType = L_EXTERNAL;
-	generic_string _udlName;
+
+	wchar_t _udlName[MAX_PATH];
+	wchar_t _pluginMessage[MAX_PATH];
```
shriprem, line 286: The previous version of `generic_string _udlName;` has to be converted into `wchar_t _udlName[MAX_PATH]` to prevent the crash reported in issue #11633.

PowerEditor/src/Parameters.h
```diff
@@ -296,7 +300,8 @@ struct CmdLineParamsDTO
 		dto._pos2go = params._pos2go;
 		
 		dto._langType = params._langType;
-		dto._udlName = params._udlName;
+		wcsncpy(dto._udlName, params._udlName.c_str(), MAX_PATH);
+		wcsncpy(dto._pluginMessage, params._pluginMessage.c_str(), MAX_PATH);
 		return dto;
```
shriprem, line 305: The `wscncpy(...)` will ensure the serialization of incoming string pointer data.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -145,171 +145,8 @@ intptr_t CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
-				_debugInfoStr += szWINEVersion;
-				_debugInfoStr += TEXT("\r\n");
-			}
-
-			// Plugins
-			_debugInfoStr += TEXT("Plugins : ");
-			_debugInfoStr += _loadedPlugins.length() == 0 ? TEXT("none") : _loadedPlugins;
-			_debugInfoStr += TEXT("\r\n");
-
-			::SetDlgItemText(_hSelf, IDC_DEBUGINFO_EDIT, _debugInfoStr.c_str());
```
donho, line 311: It's shame to get the static information over and over again.
I would say that we can keep `_debugInfoStr` here, and  call `::SetDlgItemText(_hSelf, IDC_DEBUGINFO_EDIT, _debugInfoStr.c_str());` later.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -145,171 +145,8 @@ intptr_t CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
-			// Binary path
-			_debugInfoStr += TEXT("Path : ");
-			TCHAR nppFullPath[MAX_PATH];
-			::GetModuleFileName(NULL, nppFullPath, MAX_PATH);
-			_debugInfoStr += nppFullPath;
-			_debugInfoStr += TEXT("\r\n");
-
-			// Command line as specified for program launch
-			_debugInfoStr += TEXT("Command Line : ");
-			_debugInfoStr += nppParam.getCmdLineString();
```
donho, line 184: Use `$COMMAND_LINE_PLACEHOLDER$` as place holder.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -381,10 +218,181 @@ void DebugInfoDlg::doDialog()
...
+	// Refresh the Debug Information.
+	// For example, the command line parameters may have changed since this dialog was last opened during this session.
+	refreshDebugInfo();
+
 	// Adjust the position of AboutBox
 	goToCenter();
 }
 
+void DebugInfoDlg::refreshDebugInfo()
+{
```
donho, line 230: Just copy _debugInfoStr into a string, then replace $COMMAND_LINE_PLACEHOLDER$ with the dynamic value in this function.
Set the string value into text zone by using `::SetDlgItemText(_hSelf, IDC_DEBUGINFO_EDIT, debugInfoNewStr.c_str());`.

donho (changes requested):
Please consider what I've suggested.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
shriprem: As suggested, I have moved the construction of the _debugInfoStr back inside WM_INITDIALOG. I am using the `$COMMAND_LINE_PLACEHOLDER$` for the command line parameter. This placeholder will later be replaced with the actual string value from `nppParam.getCmdLineString()` when `DebugInfoDlg::doDialog()` calls `refreshDebugInfo()`.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
shriprem: As suggested, I am using the `$COMMAND_LINE_PLACEHOLDER$` for the command line parameter. But I have abstracted 
 the `$COMMAND_LINE_PLACEHOLDER$` string into `const _cmdLinePlaceHolder`. This abstraction allows a single definition of this placeholder string value to be used in multiple places: placeholder insertion in WM_INITDIALOG, find placeholder position and replace placeholder in refreshDebugInfo().

donho (changes requested):
Please consider the following suggestion.

shriprem:
@donho
>https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11633 deserves a new PR.
>Your PR gets more chance to be accepted if the code review is simple.

>Let's not mix all the fixes in one PR. Please create a new PR for the issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11633

Fine with me. I have reverted the two changed lines relevant to #11633 from this PR. Will create a separate PR for it later.

But I am still retaining the serializing code for the `_pluginMessage` in these two lines:
Line 286: `wchar_t _pluginMessage[MAX_PATH];`
Line 304: `wcsncpy(dto._pluginMessage, params._pluginMessage.c_str(), MAX_PATH);`

Otherwise, I get a crash when re-invoking NPP with the following sample command:
`"notepad++.exe" license.txt -pluginMessage="arg1 arg2 arg3 arg4 arg5"`

Whereas with the serialized code for `_pluginMessage` I am able to get this:
image

donho (changes requested):

donho (approved):

PR #11597, opened 2022-04-27
Update portuguese.xml

hugok79:
Update portuguese translation

Minor corrections
Commit: c32da953cf9d1360ff4d30d8b26f31e1be993117


donho:
There's no way to make a huge diff like this.
Please redo your PR with only the necessary change.

hugok79:
Ok, my mistake. I just understand translation, still not very comfortable on github. Sorry @donho 
I created another PR: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11640

Issue #11596, opened 2022-04-27
Current Line Indicator: Editor vs. Search results


Yaron10:
In previous versions it was obvious that the "Highlight current line" preference only applied to the editing views and did not affect the Search Results window.
_We had a Full Line highlight in the Search Results window even if "Highlight current line" was unchecked._

With [Add "Framed current line" option beside of "Highlight current line background"](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d3e77c0e6f454374ad5e803da4c35fde9ce7ef45), it's not so clear: if you select "None" you get a Full Line in the Search Results, whereas if you select "Frame" you get a Frame.
(After switching from "None" to "Frame", the highlighting in the Search Results changes only after a new search).

Also - some users might prefer a Frame in the editing views, and a Full Line in the Search Results window.

**Suggestion:**
The simpler solution would be applying the "Current Line Indicator" pref to the editor only, and always have a Full Line in the Search Results.

PR #11598, opened 2022-04-27
Override with "Search result" style in "Search results" view

shriprem:
Fixes: #11596

With this PR, "Search results" view will only use the _Current line background color_ to set a full highlight on the caret line, while suppressing the caret frame that could get applied on the main and sub edit views.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2680,8 +2680,6 @@ void FindReplaceDlg::findAllIn(InWhat op)
 		_pFinder->_scintView.execute(SCI_SETCODEPAGE, SC_CP_UTF8);
 		_pFinder->_scintView.execute(SCI_USEPOPUP, FALSE);
 		_pFinder->_scintView.execute(SCI_SETUNDOCOLLECTION, false);	//dont store any undo information
-		_pFinder->_scintView.execute(SCI_SETCARETLINEVISIBLE, 1);
-		_pFinder->_scintView.execute(SCI_SETCARETLINEVISIBLEALWAYS, true);
 		_pFinder->_scintView.execute(SCI_SETCARETWIDTH, 0);
 		_pFinder->_scintView.showMargin(ScintillaEditView::_SC_MARGE_FOLDER, true);
 
@@ -2814,8 +2812,6 @@ Finder * FindReplaceDlg::createFinder()
 	pFinder->_scintView.execute(SCI_SETCODEPAGE, SC_CP_UTF8);
 	pFinder->_scintView.execute(SCI_USEPOPUP, FALSE);
 	pFinder->_scintView.execute(SCI_SETUNDOCOLLECTION, false);	//dont store any undo information
-	pFinder->_scintView.execute(SCI_SETCARETLINEVISIBLE, 1);
-	pFinder->_scintView.execute(SCI_SETCARETLINEVISIBLEALWAYS, true);
 	pFinder->_scintView.execute(SCI_SETCARETWIDTH, 0);
 	pFinder->_scintView.showMargin(ScintillaEditView::_SC_MARGE_FOLDER, true);
 
@@ -4331,6 +4327,7 @@ void Finder::setFinderStyle()
 		{
 			_scintView.execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, pStyle->_bgColor);
 			_scintView.execute(SCI_SETCARETLINEFRAME, 0);
+			_scintView.execute(SCI_SETCARETLINEVISIBLE, 1);
 			_scintView.execute(SCI_SETCARETLINEVISIBLEALWAYS, true);
 		}
 	}
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -4329,7 +4329,9 @@ void Finder::setFinderStyle()
 		const Style * pStyle = pStyler->findByID(SCE_SEARCHRESULT_CURRENT_LINE);
 		if (pStyle)
 		{
-			_scintView.execute(SCI_SETCARETLINEBACK, pStyle->_bgColor);
+			_scintView.execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, pStyle->_bgColor);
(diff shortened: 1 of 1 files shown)


Yaron10:
@shriprem,

👍 
Thank you for your work.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4329,7 +4329,9 @@ void Finder::setFinderStyle()
 		const Style * pStyle = pStyler->findByID(SCE_SEARCHRESULT_CURRENT_LINE);
 		if (pStyle)
 		{
-			_scintView.execute(SCI_SETCARETLINEBACK, pStyle->_bgColor);
+			_scintView.execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_CARET_LINE_BACK, pStyle->_bgColor);
+			_scintView.execute(SCI_SETCARETLINEFRAME, 0);
+			_scintView.execute(SCI_SETCARETLINEVISIBLEALWAYS, true);
```
donho, line 4334: I don't get it, in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d3e77c0e6f454374ad5e803da4c35fde9ce7ef45, only _mainEditView & _subEditView are applied by the setting of current line indicator. How come _scintView is affected?

shriprem:
It is happening here:

```C++
void Finder::setFinderStyle()
{
	// Set global styles for the finder
	_scintView.performGlobalStyles();
...
```

Yaron10:
If there's no better solution, the following lines can apparently be removed.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/292c057f655ec5e691d92045ab78c24c01b8b102/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2683

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/292c057f655ec5e691d92045ab78c24c01b8b102/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2684

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/292c057f655ec5e691d92045ab78c24c01b8b102/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2817

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/292c057f655ec5e691d92045ab78c24c01b8b102/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2818

donho:
@shriprem 
Thank you for the hint.

> If there's no better solution, the following lines can apparently be removed.

So removing indicated lines is the alternative solution of this PR? How come?

Yaron10:
@donho,

You wrote:
> I don't get it: in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d3e77c0e6f454374ad5e803da4c35fde9ce7ef45, only _mainEditView & _subEditView are applied by the setting of current line indicator. How come Finder::_scintView is affected?

So I meant: if you don't find the answer to that question, and this PR is the solution - "the following lines can apparently be removed".

donho:
> So I meant: if you don't find the answer to that question, and this PR is the solution - "the following lines can apparently be removed".

@shriprem 
Could you check if they are safe to be removed, and if they are, remove them in this PR please?

shriprem:
`Finder::setFinderStyle()` is being called from 3 other functions -- all within FindReplaceDlg.cpp & FindReplaceDlg.h.

1. `FindReplaceDlg::findAllIn`:  On line 2683-4, `SCI_SETCARETLINEVISIBLE` & `SCI_SETCARETLINEVISIBLEALWAYS`  are being called. Then a bit later on line 2716, `_pFinder->setFinderStyle();` is called which again calls just `SCI_SETCARETLINEVISIBLEALWAYS` (as of the 1st commit).

2.  `FindReplaceDlg::createFinder`:  On line 2817-8, `SCI_SETCARETLINEVISIBLE` & `SCI_SETCARETLINEVISIBLEALWAYS` are being called. Then a bit later on line 2845, `_pFinder->setFinderStyle();` is called which again calls just `SCI_SETCARETLINEVISIBLEALWAYS` (as of the 1st commit). 

3.  In FindReplaceDlg.h, `updateFinderScintilla()`:  On line 339, only `_pFinder->setFinderStyle();` is called which calls `SCI_SETCARETLINEVISIBLEALWAYS`.

Therefore:
* The `SCI_SETCARETLINEVISIBLEALWAYS` calls on 2684 & 2818 can be safely removed.
* But to remove the `SCI_SETCARETLINEVISIBLE` calls on 2683 & 2817, we will need to first add a new call on it within `Finder::setFinderStyle()`.

I have just pushed a commit with these changes.

Yaron10:
@shriprem,

Isn't `SC_ELEMENT_CARET_LINE_BACK` the replacement of the "discouraged" `SCI_SETCARETLINEVISIBLE`?
Just like in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e094e996979deb10ffbe8e8228e4b238ddbf202c.

shriprem:
@Yaron10, you are right. I had the blinders on. My analysis was only focused within the two files. I forgot that `SCI_SETCARETLINEVISIBLE` is now being discouraged by Scintilla.

@donho, should I push another commit with the `SCI_SETCARETLINEVISIBLE` call removed from inside `Finder::setFinderStyle()`?

Yaron10:
> I don't get it: in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d3e77c0e6f454374ad5e803da4c35fde9ce7ef45, only _mainEditView & _subEditView are applied by the setting of current line indicator. How come Finder::_scintView is affected?

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e094e996979deb10ffbe8e8228e4b238ddbf202c?

shriprem:
@Yaron10, I think I had answered Don's query on that already. And he has seen it too. To reprise:

>It is happening here:

```C++
void Finder::setFinderStyle()
{
	// Set global styles for the finder
	_scintView.performGlobalStyles();
...
```

Yaron10:
@shriprem,

You are right. I had the blinders on. :)

donho:
> should I push another commit with the `SCI_SETCARETLINEVISIBLE` call removed from inside `Finder::setFinderStyle()`?

Yes, please. If they are not needed anymore.

shriprem:
I have just now pushed a commit with the `SCI_SETCARETLINEVISIBLE` call removed from inside `Finder::setFinderStyle()`. That call is not needed with Scintilla5.

Only the `SCI_SETCARETLINEVISIBLEALWAYS` call is still valid and that stays.

PR #11607, opened 2022-04-29
Add patch to colorized theme dark mode buttons.

Leonard-The-Wise:
Hello @donho,

I've added a patch to fix the dialog buttons, so they won't draw black everytime on custom colored Dark Themes.

Original look:

image

Patched:

image

I also have a ready patch that can do the same with the recently added Tab UpDown.

Original unthemed version:

image

The current version:

image

My version:

image

Remark: the spin control (Tab UpDown) is NOT included on this pull yet, but if you also want it, I can add (it's already on code, but I did not commit).

What do you think?

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -25,6 +25,7 @@
 
 #include "Parameters.h"
 #include "resource.h"
+#include "ColourPicker.h"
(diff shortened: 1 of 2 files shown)


Leonard-The-Wise:
Wait, I've commited unwanted code here to the branch. Will redo pull later. =\

PR #11634, opened 2022-05-03
Add Dark Mode buttons fix for non-black dark themes;

Leonard-The-Wise:
Hello @donho,

I've added a patch to fix the dialog buttons, so they won't draw black everytime on custom colored Dark Themes.

Original look:

image

Patched:

image

I also modified the current (new) spin control to draw more seemingly to the original control - also now supports custom colors; also, now it will not re-subclass the control everytime it changes sizes. It will do it only once (at initialization).

Original unthemed version:

image

The current version in repository:

image

My version:

image

~~Lastly, I reverted the change to add new member to `struct Colors`, because that color is just derived from the existing `colors.edge`; no new customizable option exist. So It would be good to avoid breaking compatibility for `NPPM_GETDARKMODECOLORS` message unless the new color can be customized by the user in preferences dialog, instead of being a derived color. If you care to know how I achieved the same results without defining a custom color, I just added this function:~~ (not anymore. I saw the pull request to add the customized color).

Cheers.


sXNA:
Hi my first time writing here.

Great work and looks good, but can you change default background color for button to look more like button in Windows (before this commit), this commit makes it too dark. It looks like it's disabled.

I am using both Windows 10 and 11, and I don't want round corners on Windows 10 only on 11, could you fix this. 
image
image

Also this is why I prefer current tab overflow looks.

Current looks on Windows 10.
image

Can you add custom color for highlighted border, I want to change it,  

btw
I am learning c++, and have question.
Doesn't check for `if (hwndUpdown == nullptr);` avoid resubclassing.

Leonard-The-Wise:
Yeah, the button background was actually a bug, thanks for pointing.
I also made it square and modified the arrow - for windows 10 of course. I'm afraid Windows 11 have to stay the same for consistency though.

image

image


Also, for your questions:

- Can you add custom color for highlighted border, I want to change it,
Only if Don wants to add a customized color for that. You know, it would imply on an extra option here and I think colors are already harmonic the way they are... Changing the border edge just for hot buttons could break that harmony. Don't know, Don's call on that.

image


Also

- Doesn't check for `if (hwndUpdown == nullptr);` avoid resubclassing.

Don't work for tabs... they create and destroy their own spin controls at run time. Not sure, but I think `hwndUpdown` would keep changing and changing also. Rather subclass once and forget about it (and I don't think subclassing itself is a bad thing - if Microsoft did it right, it can be very cheap - processing time wise).

Leonard-The-Wise:
Undone the change to not add customized color to hot edge to conform to [this pull request](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11641)

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -1184,9 +1520,9 @@ namespace NppDarkMode
 		SetWindowSubclass(hwnd, GroupboxSubclass, g_groupboxSubclassID, pButtonData);
 	}
 
-	constexpr UINT_PTR g_tabSubclassID = 42;
+	constexpr UINT_PTR g_comboBoxSubclassID = 42;
```
donho, line 1523: There will be 2 same values (42) for both g_tabSubclassID and g_comboBoxSubclassID.
It seems fine for the compiled PR. However, will it have some risk to have any impact?

donho:
@Leonard-The-Wise 
The compiled binary looks good.
However, the diff of PR is hard to review due to some moved blocks and reformatting.
Could you close this one, keeping the current format and create a new PR for facilitating the code review?
Thank you!

sXNA:
> Yeah, the button background was actually a bug, thanks for pointing. I also made it square and modified the arrow - for windows 10 of course. I'm afraid Windows 11 arrows have to stay the same for consistency though.
> 

Sorry I think you misunderstand me I mean round corners in Windows 11 and square in Windows 10.

Also I am not sure how to trigger it, but sometimes tab overflow control disappear when I close tab, it will reappear if I move mouse in place where it was.

PowerEditor/src/NppDarkMode.cpp
Leonard-The-Wise: No, every ID there - for any control - is 42.

Leonard-The-Wise:
@donho believe me, I didn't move blocks around. Just added some functions, renamed a few and removed some to make more sense (eg: from `tabUpDownControl` to a more generic `spinControl`, because Spin controls are not only tab exclusives - and from renderButton to renderCheckBoxOrRadioButton, because it renders checkboxes, not buttons). And then the diff became like that.

The `ComboBoxSubclass` function is **as is** and bellow `GroupboxSubclass` as in original. Means, I didn't touch it... regardless, the diff is pointing as if it moved around. Don't know how to make it better, except if i moved the new code all way to the bottom of the file, which case would make the final version ugly and disorganized. To tell the truth, the `TabUpDownSubclass` function coming after `autoSubclassAndThemeChildControls` seemed somewhat out of place already so why I moved the spinControl code to be before it.

The diff here on git seems a mess. Try reviewing it on Visual Studio. For me is a lot better there. If still not good, then I'll see what can be done then.

Thanks.

Leonard-The-Wise:
Ah, I also just synchonized my fork with the current version. No conflicts either.

Leonard-The-Wise:
> @Leonard-The-Wise
> 
> > Yeah, the button background was actually a bug, thanks for pointing. I also made it square and modified the arrow - for windows 10 of course. I'm afraid Windows 11 arrows have to stay the same for consistency though.
> 
> Sorry I think you misunderstand me I mean round corners in Windows 11 and square in Windows 10.
> 
> Also I am not sure how to trigger it, but sometimes tab overflow control disappear when I close tab, it will reappear if I move mouse in place where it was.
> 
> btw won't it be better to split this PR between buttons and tab overflow control

I understood that. Windows 10 is square, 11 is rounded.
Also, I am did no work on tab overflows here. I don't understand what bug you are talking about. For me, closing and opening tabs is always fine here.

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -456,24 +506,30 @@ namespace NppDarkMode
...
+	HBRUSH getTextBrush()				 { return getTheme()._brushes.textColorBrush; }
+	HBRUSH getDarkerTextBrush()			 { return getTheme()._brushes.darkerTextColorBrush; }
+	HBRUSH getEdgeBrush()				 { return getTheme()._brushes.edgeBrush; }
+	HBRUSH getHotEdgeBrush()			 { return getTheme()._brushes.hotEdgeBrush; }
+	HBRUSH getDisabledEdgeBrush()		 { return getTheme()._brushes.disabledEdgeBrush; }
 
 	HPEN getDarkerTextPen()               { return getTheme()._pens.darkerTextPen; }
 	HPEN getEdgePen()                     { return getTheme()._pens.edgePen; }
-	HPEN getHotEdgePen()                  { return getTheme()._pens.hotEdgePen; }
+	HPEN getHotEdgePen()				  { return getTheme()._pens.hotEdgePen; }
```
donho, line 531: The above 2 lines (removed & added) are identical, please restore it.

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -456,24 +506,30 @@ namespace NppDarkMode
...
-	HBRUSH getDarkerBackgroundBrush()     { return getTheme()._brushes.pureBackground; }
-	HBRUSH getErrorBackgroundBrush()      { return getTheme()._brushes.errorBackground; }
-
-	HBRUSH getEdgeBrush()                 { return getTheme()._brushes.edgeBrush; }
-	HBRUSH getHotEdgeBrush()              { return getTheme()._brushes.hotEdgeBrush; }
+	HBRUSH getBackgroundBrush()          { return getTheme()._brushes.background; }
+	HBRUSH getSofterBackgroundBrush()    { return getTheme()._brushes.softerBackground; }
+	HBRUSH getHotBackgroundBrush()       { return getTheme()._brushes.hotBackground; }
+	HBRUSH getDarkerBackgroundBrush()    { return getTheme()._brushes.pureBackground; }
+	HBRUSH getErrorBackgroundBrush()     { return getTheme()._brushes.errorBackground; }
```
donho, line 520: The 5 above added lines are identical as the 5 removed lines, please restore them.

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -456,24 +506,30 @@ namespace NppDarkMode
...
+	HBRUSH getHardlightBackgroundBrush() { return getTheme()._brushes.hardlightBackground; }
+	HBRUSH getSoftlightBackgroundBrush() { return getTheme()._brushes.softlightBackground; }
+	HBRUSH getTextBrush()				 { return getTheme()._brushes.textColorBrush; }
+	HBRUSH getDarkerTextBrush()			 { return getTheme()._brushes.darkerTextColorBrush; }
+	HBRUSH getEdgeBrush()				 { return getTheme()._brushes.edgeBrush; }
+	HBRUSH getHotEdgeBrush()			 { return getTheme()._brushes.hotEdgeBrush; }
```
donho, line 526: The 2 above added lines are identical as the 2 removed lines, please restore them.

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -967,65 +1303,65 @@ namespace NppDarkMode
...
-				delete pButtonData;
+		case WM_UPDATEUISTATE:
+			if (HIWORD(wParam) & (UISF_HIDEACCEL | UISF_HIDEFOCUS))
+			{
+				InvalidateRect(hWnd, nullptr, FALSE);
+			}
+			break;
+		case WM_NCDESTROY:
+			RemoveWindowSubclass(hWnd, ButtonSubclass, g_buttonSubclassID);
+			delete pButtonData;
```
donho, line 1314: The 9 above added lines are identical as the 9 removed lines, please restore them.

donho (changes requested):
A lot of reformatting has been done  (no necessary change) so please redo a PR.
This PR will be closed.

donho:
@Leonard-The-Wise 

> Ah, I also just synchonized my fork with the current version. 

The merge is not necessary usually, but if you do it, please rebase all your commit locally then force push it. Otherwise I will have conflict to solve when I rebase the PR. By doing the suggestion above will save me the time to undo the merge and save you the time to redo a new PR.

> believe me, I didn't move blocks around. 

I do believe you. However, there are so many unnecessary modifications (reformatting by your IDE I guess) made the code review difficult. OTOH I don't think it's reasonable to include those reformatting change.
Sorry but please redo a new PR based on current master, without formatting anything.
Thank you in advance.

Issue #11025, opened 2022-01-10
Feature request for syntax highlighting for new features in some popular languages


FilipSzm:
### Description of the Issue

missing keyword highlighting:
- C# 9.0 - 'record'
- C++ 20 - 'optional'
- Java 17 - 'sealed'
- Python 3.10 - 'match', 'case'

### Expected Behavior

all keywords of selected languages are highlighted

### Actual Behavior

some newer keyewords are not highlighted (see missing keyword highlighting)

### Debug Information

Notepad++ v8.1.9.3   (64-bit)
Build time : Dec  6 2021 - 19:21:37
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19042.1415
Current ANSI codepage : 1250
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

WinterSilence:
+ PHP 8.*

ThePirate42:
For c++20, 'concept' and 'requires' also lack highlighting.

rdipardo:
> * <del>Java 17 - 'sealed'</del>

Added in 1d2f662

> * C# 9.0 - 'record'
> * C++ 20 - 'optional'
> . . . 
> * Python 3.10 - 'match', 'case'

Proposed in #11643, which was closed with no commit. The changes are [here](https://patch-diff.githubusercontent.com/raw/notepad-plus-plus/notepad-plus-plus/pull/11643.patch) if someone wants to base a new pull request on them.

~~~sh-session
$ curl -Ls https://patch-diff.githubusercontent.com/raw/notepad-plus-plus/notepad-plus-plus/pull/11643.patch | git apply -v

Checking patch PowerEditor/installer/APIs/cpp.xml...
Checking patch PowerEditor/installer/APIs/python.xml...
Checking patch PowerEditor/src/langs.model.xml...
Applied patch PowerEditor/installer/APIs/cpp.xml cleanly.
Applied patch PowerEditor/installer/APIs/python.xml cleanly.
Applied patch PowerEditor/src/langs.model.xml cleanly.
~~~

ArkadiuszMichalski:
Please create issue for each language separately. For those listed here such already exist.

PR #11643, opened 2022-05-06
new keywords for python and c sharp

maknight22:
Updated keywords for python with match and case.
Updated keywords for C# with record.
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11025


PowerEditor/installer/APIs/python.xml
```diff
@@ -30,6 +30,8 @@
 			<Overload retVal="" descr="Base class for warnings about bytes and buffer related problems, mostly&#x0a;related to conversion from str or comparing to str.">
 			</Overload>
 		</KeyWord>
+		<KeyWord name="case" />
+		<Keyword name="match" />
```
Mateos81, line 34: It has to be in alphabetical order in order to work, isn't it?

vinsworldcom: Yes.

vahida1157 (approved):

PR #11648, opened 2022-05-07
Get npp version add "add_zero_padding" option

### LINKED ISSUE #11535: `VERSION_VALUE` should always have at least 3 digits
### Description of the Issue

N++'s [`VERSION_VALUE` resource string][0] sometimes has 3 digits, e.g. "8.33". 
But with every "minor version" bump, it's reduced to only 2 digits, e.g. "8.4".
After N++ 8.3.3 comes 8.4, then, perhaps, 8.4.1, then 8.5, etc.

If a plugin is concerned about compatibility, it has to know that 4 comes *after* 33, and 5 comes *after* 41. 
This is a headache for developers targeting N++'s plugin API [^1] [^2] [^3].

#### Method 1

1. Install Notepad++ 8.4 (64-bit)
2. Install HTML Tag (1.2.2)
3. Start Notepad++
4. Notice that HTMLTag thinks Notepad++ is incompatible:

htmltag_v122_x64

5. Read [this explanation](https://community.notepad-plus-plus.org/topic/22503/new-version-of-html-tag/8) of what went wrong

#### Method 2

1. Clone the Notepad++ source tree
2. Edit `PowerEditor/src/resource.h` by adding a `0` to the end of the `VERSION_VALUE` string: `#define VERSION_VALUE "8.40"`
3. Repeat steps 2-3 of *Method 1*
4. No more issue!

#### Method 3

1. Write a plugin that calls into the [`NPPM_GETNPPVERSION` API][1]
2. Build 4 different Notepad++ binaries with 4 different strings assigned to `VERSION_VALUE`
3. Run the plugin in each version and note the results:

| HIWORD | LOWORD | N++ version |
| :---:  | :---:  |  ---------: |
| 8      | 33     | 8.3.3       |
| 8      | **4**  | 8.4         |
| 8      | 41     | 8.4.1       |
| 8      | **5**  | 8.5         |

### Expected Behavior

Every version string has at least 3 digits, e.g.,

N++ 8.4 -> `#define VERSION_VALUE "8.40"`
N++ 8.5 -> `#define VERSION_VALUE "8.50"`

What does the plugin see now?

| HIWORD | LOWORD | N++ version |
| :---:  | :---:  |  ---------: |
| 8      | 33     | 8.3.3       |
| 8      | **40** | 8.4         |
| 8      | 41     | 8.4.1       |
| 8      | **50** | 8.5         |

33 . . . 40 . . . 41 . . . 50 . . . 

Every new version has a low word greater than the last one! Easy!

### Actual Behavior

33 . . . 4 . . . 41 . . . 5 . . . 

### Debug Information

Notepad++ v8.4   (64-bit)
Build time : Apr 20 2022 - 03:31:06
Path : C:\bin\npp_8.4\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 2009
OS Build : 22000.651
Current ANSI codepage : 1252
Plugins : HTMLTag.dll mimeTools.dll NppConverter.dll NppExport.dll 

[^1]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11431
[^2]: https://github.com/nvaccess/nvda/pull/13382#issuecomment-1049294658
[^3]: https://github.com/oleg-shilo/cs-script.npp/issues/64#issuecomment-1108448994

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/resource.h#L24-L25
[1]: https://github.com/notepad-plus-plus/npp-usermanual/blob/master/content/docs/plugin-communication.md#nppm_getnppversion
[2]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f6ade2c9157e7c6a56c95dfd04c436df512ca028/PowerEditor/src/resource.h#L26-L27

donho:
Fix #11535

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -197,11 +197,31 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	//void NPPM_TRIGGERTABBARCONTEXTMENU(int view, int index2Activate)
 
 	#define NPPM_GETNPPVERSION (NPPMSG + 50)
-	// int NPPM_GETNPPVERSION(0, 0)
-	// return version
-	// ex : v4.6
-	// HIWORD(version) == 4
-	// LOWORD(version) == 6
+	// int NPPM_GETNPPVERSION(BOOL ADD_ZERO_PADDING, 0)
+	// Get Notepad++ version
+	// HIWORD(returned_value) is major part of version: the 1st number
+	// LOWORD(returned_value) is minor part of version: the 3 last numbers (digits)
+	// 
+	// ADD_ZERO_PADDING == TRUE
+	// 
+	// version  | HIWORD | LOWORD
+	//------------------------------
+	// 8.9.6.4  | 8      | 964
+	// 9        | 9      | 0
+	// 6.9      | 6      | 900
+	// 6.6.6    | 6      | 660
+	// 13.6.6.6 | 13     | 666
+	// 
+	// 
+	// ADD_ZERO_PADDING == FALSE
+	// 
+	// version  | HIWORD | LOWORD
+	//------------------------------
+	// 8.9.6.4  | 8      | 964
+	// 9        | 9      | 0
+	// 6.9      | 6      | 9
+	// 6.6.6    | 6      | 66
+	// 13.6.6.6 | 13     | 666
 
 	#define NPPM_HIDETABBAR (NPPMSG + 51)
 	// BOOL NPPM_HIDETABBAR(0, BOOL hideOrNot)
@@ -551,7 +571,6 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// return the caret current position column
 
 	#define NPPM_GETNPPFULLFILEPATH			(RUNCOMMAND_USER + NPP_FULL_FILE_PATH)
-
 
 
 // Notification code
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -200,7 +200,7 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// int NPPM_GETNPPVERSION(BOOL ADD_ZERO_PADDING, 0)
 	// Get Notepad++ version
 	// HIWORD(returned_value) is major part of version: the 1st number
-	// LOWORD(returned_value) is minor part of version: the 3 last numbers (digits)
+	// LOWORD(returned_value) is minor part of version: the 3 last numbers
 	// 
 	// ADD_ZERO_PADDING == TRUE
 	// 
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1265,15 +1265,35 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return TRUE;
 		}
 
+		// ADD_ZERO_PADDING == TRUE
+		// 
+		// version  | HIWORD | LOWORD
+		//------------------------------
+		// 8.9.6.4  | 8      | 964
+		// 9        | 9      | 0
+		// 6.9      | 6      | 900
+		// 6.6.6    | 6      | 660
+		// 13.6.6.6 | 13     | 666
+		// 
+		// 
+		// ADD_ZERO_PADDING == FALSE
+		// 
+		// version  | HIWORD | LOWORD
+		//------------------------------
+		// 8.9.6.4  | 8      | 964
+		// 9        | 9      | 0
+		// 6.9      | 6      | 9
+		// 6.6.6    | 6      | 66
+		// 13.6.6.6 | 13     | 666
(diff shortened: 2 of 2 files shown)


PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -197,11 +197,31 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	//void NPPM_TRIGGERTABBARCONTEXTMENU(int view, int index2Activate)
 
 	#define NPPM_GETNPPVERSION (NPPMSG + 50)
-	// int NPPM_GETNPPVERSION(0, 0)
-	// return version
-	// ex : v4.6
-	// HIWORD(version) == 4
-	// LOWORD(version) == 6
+	// int NPPM_GETNPPVERSION(BOOL ADD_ZERO_PADDING, 0)
```
rdipardo, line 200: OK, but how does a plugin use this feature?
You have to first discover the current version &#x2014; by calling `NPPM_GETNPPVERSION`!
If the running N++ has the new feature, you call `NPPM_GETNPPVERSION` a second time, with the extra parameter . . . ?

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -197,11 +197,31 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
...
 	#define NPPM_GETNPPVERSION (NPPMSG + 50)
-	// int NPPM_GETNPPVERSION(0, 0)
-	// return version
-	// ex : v4.6
-	// HIWORD(version) == 4
-	// LOWORD(version) == 6
+	// int NPPM_GETNPPVERSION(BOOL ADD_ZERO_PADDING, 0)
+	// Get Notepad++ version
+	// HIWORD(returned_value) is major part of version: the 1st number
+	// LOWORD(returned_value) is minor part of version: the 3 last numbers (digits)
```
MAPJe71: Why not use:
`HIBYTE(HIWORD(returned_value))` for first part of version (major),
`LOBYTE(HIWORD(returned_value))` for second part of version (minor digit 1),
`HIBYTE(LOWORD(returned_value))` for third part of version (minor digit 2),
`LOBYTE(LOWORD(returned_value))` for forth part of version (minor digit 3)?

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
donho: The best way IMO is to check number of digit in LOWORD - if it's not 3 digits, then it's for plugin to do the zero padding.
I have no better solution so far.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
donho: Yes, It's quite clean and clear. But it'll be a new API.

shriprem: Just a sidenote. In my plugins, I have been using this code for getting a properly sortable NPP version:

```C++
float Utils::getNPPVersion() {
   long versionNum{ static_cast<long>(nppMessage(NPPM_GETNPPVERSION, 0, 0)) };

   return std::stof(std::to_wstring(HIWORD(versionNum)) + L"." + std::to_wstring(LOWORD(versionNum)));
}
```

donho: @shriprem 
Good demo - that's the way to go!
Do you mind I use your function as document/comment of NPPM_GETNPPVERSION ?

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ef609c896f209ecffd8130c3e3327ca8a8157e72

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
shriprem: Please use it. My pleasure!

pnedev: @donho ,
@shriprem 's approach is very clever but the problem is that plugin developers don't know the Notepad++ versioning convention - at least I didn't know it (maybe I didn't search in the right places).
IMO it will be good to add to `Notepad_plus_msgs.h` as comment to NPPM_GETNPPVERSION the way Notepad++ versions increment (what are the possible values).
For example if Notepad++ supported versions like `8.0.3` @shriprem 's solution would not work (it would be the same as `8.3.0`) and I thought that such versions were completely valid.

donho: @pnedev 
> For example if Notepad++ supported versions like 8.0.3 @shriprem 's solution would not work (it would be the same as 8.3.0) and I thought that such versions were completely valid.

Indeed. In this case, it'll be better to use ADD_ZERO_PADDING come with v8.4.1 (ready to go) - so 8.0.3 (LOWORD 030 ie. 30) will be smaller than  v8.3  (LOWORD 300).

I will add a new API in v8.4.2 with the suggestion of @MAPJe71 : 
`HIBYTE(HIWORD(returned_value))` for first part of version (major),
`LOBYTE(HIWORD(returned_value))` for second part of version (minor digit 1),
`HIBYTE(LOWORD(returned_value))` for third part of version (minor digit 2),
`LOBYTE(LOWORD(returned_value))` for forth part of version (minor digit 3).

So there's no more ambiguity.

Issue #10824, opened 2021-11-23
Notepad++ does not save file extension when period in file name


KyleCrawford:
### Description of the Issue
When saving a file, if you include a '.' in the file name it will not automatically add the file extension at the end
eg.
Trying to save file named "9. Test" (without the quotes), selecting 'normal text file (*.txt)' as the type
![image](https://user-images.githubusercontent.com/14237962/143119365-275a2158-0620-4d32-bd74-43cff2126694.png)

The file extension will not get added to the file name unless specifically added.
![image](https://user-images.githubusercontent.com/14237962/143121318-044bbb80-4db7-4aaf-8b3f-4f54aa148c55.png)

### Steps to Reproduce the Issue
1.  Create new file, attempt to save it with a period in the name

### Expected Behavior
For the file extension to be added. Same as if there was no '.' in the name.

### Actual Behavior
No file extension is appended.

### Debug Information
Notepad++ v8.1.9.2   (32-bit)
Build time : Nov 21 2021 - 04:27:12
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2009
OS Build : 19043.1348
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

80ROkWOC4j:
There should be logic to check the file title before append extension.

donho:
I would say it's a tolerable limitation.
@Yaron10 & @mere-human what do you think about this issue?

mere-human:
> I would say it's a tolerable limitation. @Yaron10 & @mere-human what do you think about this issue?

@donho It is completely expected behavior.
Notepad++ thinks that anything that comes after `.` is a file extension.
There's no reliable way to differentiate a file extension from "just a suffix separated by a dot".
If you choose to use dots in the file name, then you should append the extension manually.
I'd suggest closing it.

donho:
Thank you @mere-human for your opinion.
Yes, as I said, the extension not being appended in case of dot being part of file name is a tolerable limitation.
Trying to manage such situation will make code complex, and I even wondering if there is a solution.

KyleCrawford:
"Thank you @mere-human <https://github.com/mere-human> for your opinion.
Yes, as I said, the extension not being appended in case of dot being part
[...]

PR #11649, opened 2022-05-08
Improve "Append extension" when save a file.

80ROkWOC4j:
## Impove Append extension

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10824

This PR enhances the "Append extension" feature.

Now :
If the file name contains a period, the extension will not be added even if the "Append extension" is checked.
If you change the file format from drop box, the file name after the period is replaced.

Changes :
Add extension if "Append extension" is checked and current extension is unkown extension.
If you change the file format from drop box, the file name after the period is not replace if it is unknown extension.
Custom extensions included in notepp's Style Configurator are excluded from this new rule. Those are kown extensions.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -76,7 +76,8 @@ namespace // anonymous
 	{
 		std::unordered_set<generic_string> result;
 
-		for (const auto& filter : filterSpec) {
+		for (const auto& filter : filterSpec) 
+		{
 			char delimiter = ';';
 			size_t pos = filter.ext.find('.'), posEnd = 0;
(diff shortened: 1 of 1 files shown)


Yaron10:
@80ROkWOC4j,

Thank you for your work.

The implementation in this PR can be useful in many cases.

What do you think of the following case?

You download a `Test.xul` file (and `xul` is an unknown extension).
You open it in NPP and want to save it as `Test.xml`.
In the Save dialog, you change the type to `xml` and save.

With your PR, the file would be saved as `Test.xul.xml`.

mere-human:
I'd discourage against this change. There could be many cases when the behavior would be more unexpected than without this change.
As mentioned above, there could be a lot of unregistered extensions.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10824#issuecomment-1121656620

PR #11660, opened 2022-05-12
new theme based on Ruby Blue

airduv:
new theme created from DarkModeDefault.xml (n++ v8.1.4) but based on Ruby Blue


airduv:
the theme is not perfect, i will test it some more and open another merge request when ready

PR #11662, opened 2022-05-12
Show installed plugin versions in Debug Info dialog

### LINKED ISSUE #11284: Debug info dialog lists the installed plug-ins but not the plug-in versions
### Debug info dialog lists plug-ins but not versions

In the menu `? -> Debug info -> Copy debug info to clipboard` on the last line it lists the plug-ins that are installed. However, it only lists the dll names and not the version for each plug-in.

This would be handy to determine if an old plugin is causing problems.

See menu `? -> Debug info -> Copy debug info to clipboard`

### Actual Behavior
The current debug info text is like this for example:

```
Notepad++ v8.2.1   (64-bit)
Command Line : "C:\Users\ReuverB\************DataDictionary.csv"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows Server 2019 Standard (64-bit) 
OS Version : 1809
OS Build : 17763.2452
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll CSVLint.dll CsvQuery.dll NppConverter.dll NppExport.dll
```

### Expected Behavior
It would be nice if it showed something like this:
```
..
Plugins : ComparePlugin.dll (v2.0.1) CSVLint.dll (v0.4.4) CsvQuery.dll (v1.2.8) NppConverter.dll (v4.3) NppExport.dll (v0.3)
```

rdipardo:
### Before

dbg_info_b4

### After

dbg_info_proposed

Fixes #11284

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -780,10 +780,15 @@ bool PluginsManager::allocateMarker(int numberRequired, int *start)
 generic_string PluginsManager::getLoadedPluginNames() const
 {
 	generic_string pluginPaths;
-	for (size_t i = 0; i < _loadedDlls.size(); ++i)
+	PluginUpdateInfo pl;
+	for (const auto &dll : _loadedDlls)
 	{
-		pluginPaths += _loadedDlls[i]._fileName;
-		pluginPaths += TEXT(" ");
+		pl = PluginUpdateInfo(dll._fullFilePath, dll._fileName);
+		pluginPaths += TEXT("\r\n    ");
+		pluginPaths += dll._displayName;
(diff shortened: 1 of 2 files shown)


rdipardo:
Someone will probably ask, so I'll answer now:

#### "Does Plugins Admin not *already* provide this information at 'Plugins > Plugins Admin > Installed'?"

Yes, but this can easily be copied and pasted into the issue template.
  
Note that this initial commit indents the plugin list by **4** spaces, but does **not** place a line break between the "Plugins:" label and the listed info, so everything is plain text in GH's markdown editor:

new_dbg_info_mrkd

The list could  be formatted to trigger GH's automatic code fence style; an extra line break at the top is all that's needed:

new_dbg_info_mrkd_code_fence

I personally don't like the latter format, but it's not for me to decide.

PR #11701, opened 2022-05-19
Make it possible to compile Plugins Admin with MinGW in debug mode

rdipardo:
Since [f428fba][0], the MinGW makefile *does not* pass the `DEBUG` compiler definition to GCC during a debug build. This is contrary to [Scintilla][1] and [Lexilla][2], which are built from their own makefiles and *do* pass the appropriate definition in debug mode.

The missing compiler def makes GCC skip the JSON loading feature in the Plugins Admin module:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/10f48e36ad34195adb02acbf9acbc77fdef8a842/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp#L706-L711

This means a manually edited plugin list can't be tested with GCC builds, as first reported in notepad-plus-plus/nppPluginList#454.

By sheer accident, this was preventing a compiler failure, as the debug code path currently breaks GCC builds:

<details>
<summary>* compiling ../src/WinControls/PluginsAdmin/pluginsAdmin.cpp</summary>
<pre><code>
../src/WinControls/PluginsAdmin/pluginsAdmin.cpp: In member function 'bool PluginsAdminDlg::initFromJson()':
../src/WinControls/PluginsAdmin/pluginsAdmin.cpp:709:55: error: no matching function for call to 'std::basic_ifstream<char>::basic_ifstream(generic_string&)'
  709 |         ifstream nppPluginListJson(_pluginListFullPath);
    |                                                       ^
In file included from ../src/WinControls/PluginsAdmin/pluginsAdmin.cpp:20:
c:\winlibs-mingw-msvcrt\current\include\c++\11.2.0\fstream:584:9: note: candidate: 'template<class _Path, class _Require> std::basic_ifstream<_CharT, _Traits>::basic_ifstream(const _Path&, std::ios_base::openmode) [with _Path = _Path; _Require = _Require; _CharT = char; _Traits = std::char_traits<char>]'
  584 |         basic_ifstream(const _Path& __s,
    |         ^~~~~~~~~~~~~~
c:\winlibs-mingw-msvcrt\current\include\c++\11.2.0\fstream:584:9: note:   template argument deduction/substitution failed:
c:\winlibs-mingw-msvcrt\current\include\c++\11.2.0\fstream:583:32: error: 'class std::__cxx11::basic_string<wchar_t>' has no member named 'make_preferred'
  583 |       template<typename _Path, typename _Require = _If_fs_path<_Path>>
[...]

Issue #11687, opened 2022-05-17
Are gcc debug builds and PluginAdmin incompatible?


Ekopalypse:
The use of a gcc debug build does not seem to be recognised by the plugin admin.
Can this be changed?
More detailed information about the problem has been posted as an [issue](https://github.com/notepad-plus-plus/nppPluginList/issues/454) in the nppPluginList repo.

If this issue makes no sense or is invalid, please feel free to close it.
Thank you.

rdipardo:
That you even got a GCC build to load Plugins Admin is an improvement on #10185.
 
> Neither toolbar nor Plugin Admin to be seen. [ . . . ]
> I even tried downloading the “Plugin Manager” plugin manually also with no success — it just doesn't show up anywhere.

Otherwise, I would close this as a duplicate of #10185 and (more tangentially) #9163.

As discussed in the [linked issue], the real problem is that no DEBUG compiler def is passed to GCC by [the makefile][1], so JSON loading is disabled in the [Plugins Admin module][3]. When the def *is* passed, GCC [fails to compile the module][2]. The git history does not suggest the DEBUG def was suppressed on purpose; it was just [added that way][0], and nobody noticed until now.

[0]: https://github.com/chcg/notepad-plus-plus/commit/f428fbab13c6bbea673b57457b35d9a4a9aa1f09#diff-81078873c7192ded6b5a93da2d62b0e6f86a5516905529f465fa5b7c33bb2df4R47
[1]: https://github.com/chcg/notepad-plus-plus/blob/master/PowerEditor/gcc/makefile#L63
[linked issue]: https://github.com/notepad-plus-plus/nppPluginList/issues/454
[2]: https://github.com/notepad-plus-plus/nppPluginList/issues/454#issuecomment-1130670309
[3]: https://github.com/chcg/notepad-plus-plus/blob/a577564699c993508d959304730e334fbc83d208/PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp#L706-L711

Ekopalypse:
Has been fixed, details in linked issue - thank you very much.

Ekopalypse:
Since the other issues mentioned are already closed, I am reopening this one so that a needed PR can link to an issue.

PR #11702, opened 2022-05-20
Test GCC Debug mode

donho:


rdipardo:
Don't forget to also change `appveyor.yml`:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b3a9ec5dfeb74161e8a39b6aab7f897854cc02de/appveyor.yml#L30-L34

rdipardo:
> Don't forget to also change `appveyor.yml`

That will fail the build (as expected);  #11701 can then be rebased with the updated `appveyor.yml` and the fix

donho:
@Ekopalypse 
Could you check GCC debug build to make sure if [the issue](https://github.com/notepad-plus-plus/nppPluginList/issues/454) is fixed ?

Ekopalypse:
Yes, I will try it tomorrow and report back to you.

rdipardo:
This compiles on the latest GCC [^1] 👍🏼
For some reason I can't get Plugins Admin to load, but that was true even before, and @Ekopalypse had no such problem. (I suspect Windows 11 is blocking the DLL with a strict security policy.)
 
@donho, you should link #11687 to this PR and close #11701 as redundant.

[^1]: gcc version 11.2.0 (MinGW-W64 x86_64-posix-seh, built by Brecht Sanders)

Ekopalypse:
Yes, I can confirm that this works like expected.

image

and normal, non-debug, build works also

image

@rdipardo, @donho - Thank you very much for the quick fix and the help in this matter.

Ekopalypse:
@rdipardo 
>  For some reason I can't get Plugins Admin to load

Are you talking about the debug version? If so, you must have an nppPluginList.json instead of the nppPluginList.dll in the plugins/config directory and the updater folder, in the npp root directory, must exist with all files as usual.

But if it's about the release version, then I can only confirm that it works under Windows 7, unfortunately I don't have access to a Windows 11 version.

image

rdipardo:
@Ekopalypse, 
> Are you talking about the debug version? If so, you must have an nppPluginList.json instead of the nppPluginList.dll in the plugins/config directory and the updater folder, in the npp root directory, must exist with all files as usual.
> 

The toolbar menu is not even populated with the "Plugin's Admin" item even when nppPluginList.dll is in the right place, so I'm guessing the DLL_ATTACH message is never received or it's silently intercepted by the OS.

donho:
@Ekopalypse 
> Yes, I can confirm that this works like expected.
> and normal, non-debug, build works also

Thank you for testing!

@rdipardo 
> you should link https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11687 to this PR and close https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11701 as redundant.

This PR is just POC for making sure the compiling working with the simpler modification.
You've found solution so let's use your PR. OTOH, could you please use:
```cpp
#ifdef DEBUG // if not debug, then it's release

	// load from nppPluginList.json instead of nppPluginList.dll
	ifstream nppPluginListJson(wstring2string(_pluginListFullPath, CP_UTF8));
	nppPluginListJson >> j;

#else //RELEASE
```
instead of your modification in [PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11701/files#diff-021d8f1c898c0698c6ed6dae6edfbf895a9235f1b833b464f1cb35c29893728d)
?

rdipardo:
@donho, you're minimal solution is better on all counts, yes.
I wonder, though, if it should still be guarded by `#ifdef __MINGW32__`, as that's the only compiler that actually needs it.
The MSVC debug builds are more useful and never had a problem opening streams to UTF-16 file paths.
There may be no risk of regression, but we can't be certain until we've actually seen what `wstring2string()` will do with a path in, say Korean Hangul.

rdipardo:
> There may be no risk of regression . . . 

Actually, a quick experiment suggests that `wstring2string()` does *not* convert wide file paths into usable ANSI equivalents.
Tested with MSVC 19.29.30145: 

wstring2string-msvc

and gcc 9.2.0 (tdm64-1):

wstring2string-gcc

I say we restrict the patch to MinGW debug builds to protect international users from nasty surprises.

pryrt:
@donho,
> What do you think?

There isn't any other information about compiling anywhere in the user manual, so I am hesitant.  

Would it go better in the [GitHub Wiki "Testing" page](https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/Testing) page, where it already shows how to download test builds?

Or an even more logical place would be to create a new page in that same [GitHub Wiki area](https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/) called "Compiling".  That page could then include the notes about MinGW, as well as be a place where other notes are kept that might help people correctly compile the app (for example, a new contributor might want advice on how to set up the Visual Studio build environment to get the compile to work properly to verify their fork/branch before submitting a PR) ... (I know that the Forum has had such discussion before, showing optimal sequence to compile Notepad++ executable under specific VS versions).  Thus, to me, a Wiki: "Compiling" page seems the best choice.

But if you really think it belongs in the User Manual instead, I will put a comment in the "other resources" section.

Let me know which you prefer: In the Wiki:"Testing", in a new Wiki:"Compiling", or in the User Manual.

donho:
@pryrt 
> There isn't any other information about compiling anywhere in the user manual, so I am hesitant.
>  Would it go better in the [GitHub Wiki "Testing" page](https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/Testing) page, where it already shows how to download test builds?

Yes, you're right. Putting it on GitHub Wiki is a better idea!

PR #11666, opened 2022-05-13
Finder window. "One result per line"\"Results in line" option.

vitali-kurylovich:
"Results in line" feature is useful for tables and logs analysis.

Indirectly partially softens the problem #9879

notepad++ Finder _One_result_per_line

Issue #9580, opened 2021-02-25
Please mark words on a SINGLE line in find results


a-einstein:
When I search for a word in file(s), let's say 'error' in a log-file,  I get every instance of the word in a line reported on a SEPARATE line, each one of them highlighted.

I would prefer to have all instances in a line reported on one SINGLE line, and of course all of them highlighted. 
That would create a more compact overview.
Maybe the number of instances in the line could be reported at the reporting line too.

So my request is to either implement it like that, or make it an option.

sasumner:
Similar to #1499 
Ongoing work will proceed via that issue.

Issue #2481, opened 2016-10-25
Display only one row in "Find All in Current Document"


VARGA-Peter:
### Description of the Issue

The "Find result" window displays each occurrences of the subject in one row.
![noptepad example](https://cloud.githubusercontent.com/assets/9625486/19682567/84ea7184-9aae-11e6-95b7-d825d495a15a.png)

There could be an option in the "Find" Window called something like "Group in one line" with which in the above example only 1 line would be displayed and the 3 occurrences would be marked three times in this line.

moymike:
I /agree.  [[my post]](https://notepad-plus-plus.org/community/topic/15631/find-results-multiple-matches-per-line-would-like-non-duplicate-line-result-highlighting-every-match/2)

sasumner:
Similar to #1499 so closing.

sasumner:
This feature-request is REJECTED by the author of Notepad++ ( https://github.com/notepad-plus-plus/notepad-plus-plus/pull/9088#issuecomment-720016310 )
It will not be reconsidered.
Please do not open any new issues on this topic as they will be immediately closed.

The suggestion is to use this workaround:
* Set your _Search mode_ to _Regular expression_.
* Add the following to the end of your _Find what_ expression:  `(?-s).*`

Example 1: To show only the first hit on a line for `foo`, make your _Find what_ expression `foo(?-s).*`

If the literal text you are intending to search for contains regular expression metacharacters, add `\Q` to its start and `\E(?-s).*` at its end.

Example 2: To show only the first hit on a line for `*foo*`, make your _Find what_ expression `\Q*foo*\E(?-s).*`

PR #11705, opened 2022-05-20
Finder window. "One result per line"\"Results in line" option.

vitali-kurylovich:
(Reopen PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11666)

Refers to issues https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2481, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1499, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9580

"Results in line" feature is useful for tables and logs analysis.

notepad++ Finder _One_result_per_line

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1495,6 +1495,7 @@ Find in all files but exclude all folders log or logs recursively:
             <finder-clear-all value="Clear all"/>
             <finder-open-all value="Open all"/>
             <finder-purge-for-every-search value="Purge for every search"/>
+            <finder-one-result-per-line value="Output each search result in a separate line"/>
             <finder-wrap-long-lines value="Word wrap long lines"/>
             <common-ok value="OK"/>
             <common-cancel value="Cancel"/>
--- a/PowerEditor/installer/nativeLang/russian.xml
+++ b/PowerEditor/installer/nativeLang/russian.xml
@@ -1483,6 +1483,7 @@ Updated to v8.4.1:
 			<finder-select-all           value="Выделить всё"/>
 			<finder-clear-all            value="Очистить поле поиска"/>
 			<finder-open-all             value="Открыть все файлы"/>
+			<finder-one-result-per-line  value="Выводить каждый результат поиска отдельной строкой"/>
 			<finder-wrap-long-lines      value="Перенос слов длинных строк"/>
 			<common-ok                   value="ОК"/>
 			<common-cancel               value="Отмена"/>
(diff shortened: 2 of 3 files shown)


alankilborn:
Bug:  Default is multiple results per line -- default should be current (8.4.1) behavior

Bug:  Out of the box, with the default bug above, when multiple results per single output line are shown, the context menu item must be ticked; it is not:

image

alankilborn:
OK...so I find "One result per line" a little bit confusing.  So much so that I starting posting a bug about how it was working here.  Can it be some better text?  "Multiple line hits generate single result line"?  It is longer, true...

alankilborn:
I think a user that is toggling this option from checked to unchecked (or from unchecked to checked) would expect to see all data currently in the _Search results_ window change its format.  If it is only to be for _future_ searches, maybe the UI text should indicate that?  Examples:

* One result per line (for future searches)
* Multiple line hits generate single result line (for future searches)

Typical experience:  User runs search and sees line 20 output 5 times in the _Search results_ window.  User goes in and ticks the "One result per line" option.  Then user first thinks it isn't working, then realizes and thinks "UGH!! I have to re-run the search to get it to work??!"

alankilborn:
General comment:  There is a lot of code change in this PR, for such a smallish feature; what is the regression potential?

alankilborn:
Bug: Default is the new behavior.  Default behavior should ALWAYS be existing behavior, and user can change existing/default behavior to new behavior if they want to.

alankilborn:
Question: How does one jump to the second match on a line?  Double-clicking on the result only allows jumping to the edit window at the point of the first match.

alankilborn:
How is the user to know how many matches are on a given line in this mode, when not all matches are currently visible in the _Search results_ area?

vitali-kurylovich:
> Bug: Default is the new behavior. Default behavior should ALWAYS be existing behavior, and user can change existing/default behavior to new behavior if they want to.

Fixed

vitali-kurylovich:
> Question: How does one jump to the second match on a line? Double-clicking on the result only allows jumping to the edit window at the point of the first match.

There is no such opportunity.
Behavior similarly to MS Visua Studio.

vitali-kurylovich:
Let's define the menu text. Variants:
- Output each search result in a separate line
- Output each result in a separate line
- Output each match in a separate line
- Separate each search result
- Isolate every match

Is the "(for future searches)" addition really necessary for understanding?

alankilborn:
>Is the "(for future searches)" addition really necessary for understanding?

More importantly:  Is it necessary so that users won't file bug reports because it doesn't change existing search output?

Issue #11710, opened 2022-05-22
[DarkMode] Dark mode on plugins with docking interface


ozone10:
### Description of the Issue

Some plugins are not maintained or don't have dark mode yet. For maintained we have to wait.
But for old ones there is no easy solution for user.
![image](https://user-images.githubusercontent.com/55940305/169709950-a0c40c45-807e-4cf4-9593-f568e3800d14.png)
![image](https://user-images.githubusercontent.com/55940305/169689171-743a0e63-ab96-478a-9a9c-5de5ac527d98.png)

### Steps to Reproduce the Issue

1. 
2. 
3. 

### Expected Behavior

Apply basic dark mode plugins docking window

### Actual Behavior

Plugins docking window remain light

### Debug Information

Notepad++ v8.4.1

PR #11711, opened 2022-05-22
[Done] Dark mode plugin dock panel

ozone10:
<s>Currently this PR is draft and for testing plugins dark mode.</s>

Some plugins are not maintained or don't have support for dark mode yet.
This PR attempt to apply basic dark mode on plugins with docking interface.

image

image

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11710

<s>TODO add option in preference (not) to apply dark mode on plugins, currently dark mode is applied on every plugins with docking interface without added flag `DWS_USEOWNDARKMODE` </s>

Remarks:
Option to apply dark mode on plugins currently requires Notepad++ to restart to use new setting
image

<s>Also missing translation for it, can be added later.

Colors for listview are not applied as some plugins use their own wndproc or some subclassing, which can cause issue like white background with white text.

For treeview it can be applied but there might be same issue as with listview.</s>

image
Using custom draw allowed to apply dark on treeview and listview, also fixed issue with item's highlight being invisible.

<s>Currently not applied to internal Notepad++ dialog/windows (ascii panel, document list, windows dialog, plugin admin ...) yet.</s>

There might be still issue when switching between dark and light mode but after restart it should work.

<s>Toolbar button highlight and checked state might need tweaks.</s>

image
image


PowerEditor/src/NppDarkMode.h
```diff
@@ -48,6 +48,7 @@ namespace NppDarkMode
 	{
 		bool enable = false;
 		bool enableMenubar = false;
+		bool enablePlugin = false;
```
donho, line 51: I would say, let's enable this option by default. 
And remove the option from  GUI. 
If users complain, we could add back the option on GUI, or a non-GUI option in config.xml.

PowerEditor/src/WinControls/Preference/preference_rc.h
```diff
@@ -421,4 +421,5 @@
 	#define IDD_CUSTOMIZED_COLOR11_STATIC  (IDD_PREFERENCE_SUB_DARKMODE + 26)
 	#define IDD_CUSTOMIZED_RESET_BUTTON    (IDD_PREFERENCE_SUB_DARKMODE + 30)
 	#define IDC_DARKMODE_TONES_GB_STATIC   (IDD_PREFERENCE_SUB_DARKMODE + 35)
+	#define IDC_CHECK_DARKMODE_PLUGIN      (IDD_PREFERENCE_SUB_DARKMODE + 36)
```
donho, line 424: Let's remove (comment) it. We will see users' reaction about this feature to add it back, or to add an non-GUI option  in config, or even an empty xml file as sentinel for enable this option.

donho (changes requested):
It's a great PR.
It's shame that it's not enabled by default.
I suggest to enable the option by default and remove the GUI option.

@ozone10 
If you think the suggestion makes sense but you can not do it, let me know. I will accept this PR  then change its behaviour.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -108,6 +108,7 @@ STYLE DS_SETFONT | DS_FIXEDSYS | DS_CONTROL | WS_CHILD
 FONT 8, "MS Shell Dlg", 0, 0, 0x1
 BEGIN
     CONTROL         "Enable &dark mode",IDC_CHECK_DARKMODE_ENABLE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,15,12,150,10
+    CONTROL         "Apply dark mode on plugins with docking window (requires Notepad++ restart)",IDC_CHECK_DARKMODE_PLUGIN,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,165,12,270,10
```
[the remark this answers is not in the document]
ozone10: I mainly added it because there might be plugin which can have issue with it. It also override plugins users colors. But on the other hand I also add flag to allow plugin developers to handle dark mode themself.

PowerEditor/src/WinControls/Preference/preference_rc.h
ozone10: I like your idea of using non-GUI option, especially empty xml.

PowerEditor/src/NppDarkMode.h
ozone10: Ok.

ozone10:
@donho 
Unfortunately I will leave in hour. I won't be able to make changes in PR. But I will be able answer question here.

PowerEditor/src/WinControls/Preference/preference_rc.h
donho: > Also current implementation require restart, user still must close Notepad++ to apply new config.

True. But it doesn't matter. This info will be noted in https://npp-user-manual.org/

shriprem:
@donho & @ozone10: Please see my impact analysis feedback for this PR at: https://community.notepad-plus-plus.org/topic/23080/impact-analysis-of-generic-darkmode-rendering-of-plugin-dock-panels-from-pr-11711. Thank you for your time!

ozone10:
@shriprem 
1. there might be visual glitches if plugin is written in C#, as it use WinForms for GUI, it also does not affect plugins which use their own custom control, but for other it should be safe

2. You misunderstand it, I mean disabling/enabling applying dark mode (subclassing) on plugins with panel GUI.
3. Because Notepad++ is not using it and plugins I use don't have this button, I did take it into account. Maybe you can do PR for it.
I also added richedit control because some plugins are using it, some plugins may use tabs, which I also had not take into account.
4. This is because there is no easy way to get handle for them.
5. 👍

shriprem:
@ozone10, thanks for your PR and the clarifications.

kczx3:
I have the Snippets plugin installed which uses a ListBox control but the scrollbar isn't painting correctly after this update.
image

shriprem:
@kczx3, I encountered this issue as well with my FWDataViz plugin.

After reviewing the code for `NPPDarkMode::CustomBorderSubclass`, I found that the `return` on [line 1501](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bbf3adb9bb10f61e0a932279ece42e471970b5da/PowerEditor/src/NppDarkMode.cpp#L1501) for `case WM_NCCALCSIZE` should be removed to allow the  `DefSubclassProc(...)` to execute subsequently on [line 1560](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bbf3adb9bb10f61e0a932279ece42e471970b5da/PowerEditor/src/NppDarkMode.cpp#L1560).

Note that `case WM_MOUSEMOVE`, `case WM_MOUSELEAVE` and `case WM_NCDESTROY` blocks also end with a `break;`. The `return` for `case WM_NCPAINT` is fine because it is executing `DefSubclassProc(...)` on [line 1431](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bbf3adb9bb10f61e0a932279ece42e471970b5da/PowerEditor/src/NppDarkMode.cpp#L1431).

Finally, I also looked at why this issue was not manifesting with the list boxes in Notepad++'s **Style Configurator** dialog. It did not manifest there because `NppDarkMode::autoSubclassAndThemeChildControls(_hSelf);` is being called (on [line 159](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bbf3adb9bb10f61e0a932279ece42e471970b5da/PowerEditor/src/WinControls/ColourPicker/WordStyleDlg.cpp#L159) of `WordStyleDlg::run_dlgProc` - `case WM_INITDIALOG`) only after the two list boxes have been populated with items.

So, the scrollbar rendering issue will manifest only when `NppDarkMode::autoSubclassAndThemeChildControls(...)` gets called in a dialog before its listboxes are populated.

This one is not a biggie. On [line 1450](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bbf3adb9bb10f61e0a932279ece42e471970b5da/PowerEditor/src/NppDarkMode.cpp#L1450), it should be `SM_CYHSCROLL` instead of `SM_CXHSCROLL`.

An adapted version of the changes that I am suggesting here in `NPPDarkMode::CustomBorderSubclass` can be viewed at my FWDataViz repo [here](https://github.com/shriprem/FWDataViz/blob/db4754afb5985c85f72bfdf80633733d60c3ba21/src/Darkmode/NppDarkMode.cpp#L929).

Issue #11229, opened 2022-02-17
Increase the number of open documents displayed in Window menu


Yaron10:
**STR:**
Menu -> Window.

**Result:**
The number of open documents displayed is limited to 10.

**Suggestion:**
Change the limit to 20.

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11190#issuecomment-1042859100

***

Notepad++ v8.3.1   (64-bit)
Build time : Feb 16 2022 - 22:55:56
Path : C:\---\notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

alankilborn:
I think the limit was originally designed because of the "underlining":

![image](https://user-images.githubusercontent.com/77065706/154496883-be316d27-65aa-4c53-addc-e449c4d2b097.png)

If there were more files they wouldn't have underlines in their number.

Just pointing this out for historical reasons; this limit could certainly be increased if the author agrees.

Yaron10:
@alankilborn,

👍 
Very nice! I loved that comment. :)

I've actually changed the number for my personal use.
Being a "mouse-oriented" user, I've never noticed it.

Thank you.

Yaron10:
@alankilborn,

Your point is relevant to the _Recent Files_ list as well. :)
And the number is not limited to 10.

alankilborn:
>Your point is relevant to the Recent Files list as well. :)
>And the number is not limited to 10.

But the default is 10.
Suggests same development history occurred there.
At least at first; then at some point it was expanded and also made configurable.

Issue #11724, opened 2022-05-25
[Request] [UI] Plus-button to the right of tabs => new tab


elofu17:
Feature request:

Apart from pressing Ctrl-N or clicking the [New] button to the left, please add a [+] button to the right of the tabs, to easily create a New tab.
(just like you do to start a new tab in web browsers)

Also see discussion in https://community.notepad-plus-plus.org/topic/23032/plus-to-the-right-of-tabs-new-tab where, among other things, an undocumented workaround was suggested to me:
Simply doubleclick in the whitespace where the Plus-button would be, and a New tab will appear.

alankilborn:
>please add a [+] button to the right of the tabs

Example from Edge browser, for hopefully maximum clarity of what is being requested:

![image](https://user-images.githubusercontent.com/77065706/170354190-645291af-64a6-4093-8e29-7f070db4d15f.png)

ozone10:
[+] button/tab right next to tabs can cause issue, when there are several tabs and overflow is created, as button than could be hidden.

Also what about multi-line tabs, should it be next to last tab on last line, or on first line?
What about other view, should [+] button be created for each views?

Using [+] button with fixed position seems better.

What do you think about this, we already have close button here.
![image](https://user-images.githubusercontent.com/55940305/170492515-186d9f13-3cee-4542-b538-f18150cacfeb.png)

elofu17:
To me, that location is unintuitive.
In any program I expect to find the New button up to the left (where it already is), and if the program has a tab-bar, then I expect to always see a + to the right of the last tab.

ozone10:
I know web browsers and some terminals have plus button next to last tab, but no text editor.
I have tried many text editor and IDE with GUI with tabs and IIRC only two have plus button right, not next to last tab, but in fixed position. So I don't expect plus button next to last tab in text editor.

Also how should we solve this
![image](https://user-images.githubusercontent.com/55940305/170774615-2e7e4e7b-233a-4769-8b17-d5e995dd5b30.png)
last tab is hidden.

Implementing this isn't easy, and for such minor ui feature I don't think it will done soon.

alankilborn:
Desired feature added in Notepad++ 8.4.3.
Please CLOSE this issue.

PR #11733, opened 2022-05-27
Add plus button menu bar

ozone10:
This PR is example solution for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11724 to add [+] button for new tab right

image


Yaron10:
@ozone10,

Thank you for your work.

I'm wondering what extras the `▼` popup has over the `Window` menu.
Is it "more than 10 tabs"? - Wouldn't it better to implement it the `Window` menu?

You can argue that `x` and `+` might be useful if one hides the Tool Bar.
But if you hide the Menu Bar, you wouldn't get the `Window` menu nor the `▼` popup.

ozone10:
@Yaron10 
I implemented this quickly, and did not want to touch localization code (translation). This is mainly proof of concept.

About `▼` I have done it mainly to avoid issue with accidentally pressing [x] close button instead of [+] button if they are next to each other.

And of course, if you hide menu bar, you can use toolbar buttons for new tab and close, document list for tabs.
There is always some alternative if you hide tabbar, toolbar or menu bar or document list. If you hide everything than you only have ctrl+n and ctrl+w.

Yaron10:
@ozone10,

Thanks for the explanation.

My point was that "too many alternatives" can actually be "too many duplicates".
One might request adding a "Close All" button to the Menu Bar. Where do you draw the line?

alankilborn:
>My point was that "too many alternatives" can actually be "too many duplicates".

@donho might say:  "Too many alternatives equals no alternative". :-)

ozone10:
@Yaron10, @alankilborn 

Point taken, I have simplify implementation, now it is only about adding [+] button.
Window menu is moved, also increased list limit to 50, should be sufficient.
image

Localization should work, it was easier than I've thought.

alankilborn:
@ozone10 I think it is fine; thanks for your efforts.

Yaron10:
@ozone10,

Increasing the list limit is good.
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11229.

Moving the `Window` menu is a major UI change. If you think it's worth it.

donho:
I have the problem to compile this PR (release x64):

```
...
3>ReadDirectoryChanges.cpp
3>ReadDirectoryChangesPrivate.cpp
3>ReadFileChanges.cpp
3>
3>..\src\Notepad_plus.rc(22): error RC2135: file not found: VERSIONINFO
3>
3>
3>..\src\Notepad_plus.rc(23): error RC2135: file not found: 4
3>
3>Done building project "notepadPlus.vcxproj" -- FAILED.
```

On Microsoft Visual Studio Community 2022 (64-bit)

donho:
@ozone10 
> Seems like Microsoft has not fixed issue with rc file saved with utf8 encoding.
I will try to save it with utf16 encoding.

If you're talking about `Notepad_plus.rc`, it should be UTF8, but without BOM.

ozone10:
@donho
ok I will push it with that encoding

ozone10:
I will create new PR

PR #2219, opened 2016-08-27
Fixed repository URLs

MaxXor:
Just updated the repository URLs to the current one.

Issue #2213, opened 2016-08-25
DLL side loading in notepad++ installer


adarshdinesh:
### Description of the Issue
[Security Bug] DLL side loading in Notepad++ installer
DLL in the name of "netutils.dll" gets loaded while installation. 

### Steps to Reproduce the Issue
1. Download latest Notepad++ (6.9.2) build in latest Windows 10 64 bit machine 
2. Create an on load execute DLL (Adding code in DLL main) and keep in the same directory where the Notepadd++ installer is situated. Change the DLL name to "netutils.dll"
3. Execute the Notepadd++ installer. 

### Expected Behavior
Application installer should take the dll from system32

### Actual Behavior
Remote code execution and privilege escalation. 

### Debug Information
![notepad poc 2](https://cloud.githubusercontent.com/assets/3956104/17982206/82318748-6b24-11e6-9840-ad5e1d22ca5f.png)

dail8859:
It will be fixed in the next release which uses a newer version of NSIS. Duplicate of #2163, #2086

MAPJe71:
Does anyone ever search the already posted issues before posting one? Would prevent a lot of duplicates!

PR #81, opened 2015-05-16
Merge with 'upstream/master'

MAPJe71:

Issue #80, opened 2015-05-15
Scintilla update


NN---:
Please update to the latest Scintilla.
They fixed many bugs and the annoying one with WM_GETTEXT ( http://sourceforge.net/p/scintilla/code/ci/d2579847448f4e59145b0892bed20247d4e8a781/ ) . .

Thanks,

ariccio:
I tried this, somewhere else (I can't find it - grr!). Good luck getting this merged.

NN---:
I compiled Scintilla from latest revision source without any change and everything is working.

ariccio:
I know, it did for me to.

Same thing [with SQLite](https://github.com/donho/notepad-plus-plus/pull/14).

cmeriaux:
:+1:

crayxt:
+1

NN---:
Another reasons for update.
The latest Scintilla doesn't have dependency on boost anymore ( #118 )
And it has a vcxproj file for easier compiling ( https://sourceforge.net/p/scintilla/code/ci/default/tree/win32/ ) making it possible to have a solution with project reference.

NN---:
And it is x64 build ready.
Actually Scintilla must be updated for 64-bit build.

MAPJe71:
Be aware that Notepad++ uses an adapted Scintilla (indicated by 'for Notepad++' in the file version).
Some of the changes are:
 - Added the 'Search Result'-, 'Objective-C'- and 'User Defined Language'-lexers.
 - Adapted HTML lexer to default to Visual Basic Script i.s.o. JavaScript for ASP embedded script.
 - Added possibility for PCRE search which depends on Boost::RegEx.

So simply compiling the latest Scintilla revision source without any change might work but it probably won't give you back all of Notepad++'s functionality.

NN---:
I see. So basically Notepad++ has an unmaintained fork of Scintiilla. Very sad:(
This issue blocks #64

don4d:
@NN--- Scintilla will be updated in 6.8 release.

donho:
@NN--- Scintilla will be updated in 6.8 release.

NN---:
@donho Glad to hear!
Maybe you should propose pull requests to the main repository ?
Also it will be good to have a normal fork in sorceforge.
It is going to be easier maintaining and updating.
And since it has a project file you should really make a simple solution with both projects and reference.
Add sln file to repo back.

[...]

PR #11734, opened 2022-05-27
Lexilla 5.1.7 + Scintilla 5.2.3

chcg:
Update with:
Released 22 May 2022.
    Add LexAccessor::StyleIndexAt to retrieve style values as unsigned to handle styles > 127 better. [Issue #61](https://github.com/ScintillaOrg/lexilla/issues/61).
    Associate more than one file extension with a setting for testing. [Issue #81](https://github.com/ScintillaOrg/lexilla/issues/81).
    CMake: Fix folding of "ElseIf". [Issue #77](https://github.com/ScintillaOrg/lexilla/issues/77), [Pull request #78](https://github.com/ScintillaOrg/lexilla/pull/78), [Bug #2213](https://sourceforge.net/p/scintilla/bugs/2213/).
    HTML: Fix folding of JavaScript doc comments. [Bug #2219](https://sourceforge.net/p/scintilla/bugs/2219/).
    Matlab: add "classdef" and "spmd" to folding keywords. [Pull request #70](https://github.com/ScintillaOrg/lexilla/pull/70).
    Matlab: handle "arguments" contextual keyword. [Pull request #70](https://github.com/ScintillaOrg/lexilla/pull/70).
    Matlab: improve support of class definition syntax. [Pull request #75](https://github.com/ScintillaOrg/lexilla/pull/75).
    Raku: fix escape detection. [Pull request #76](https://github.com/ScintillaOrg/lexilla/pull/76).
    Ruby: fix character sequence "?\\#" to not include '#' in SCE_RB_NUMBER as only second '\' is quoted. [Issue #69](https://github.com/ScintillaOrg/lexilla/issues/69).
    Ruby: improve styling of ternary expressions as commonly used. [Issue #69](https://github.com/ScintillaOrg/lexilla/issues/69).
    VHDL: support folding for VHDL 08 else-generate and case-generate. [Pull request #80](https://github.com/ScintillaOrg/lexilla/pull/80).


chcg:
@donho See https://www.scintilla.org/ScintillaHistory.html 5.2.3:

- Duplicate APIs to support 64-bit document positions on Win32: SCI_GETTEXTRANGEFULL, SCI_FINDTEXTFULL, and SCI_FORMATRANGEFULL. This adds new types to Scintilla.iface which may impact downstream projects. Applications should move to these APIs from their predecessors as they will be deprecated.

This collides with the change done within N++ by changing the type of Sci_PositionCR. How do you think we should handle that?

donho:
@chcg 
It's rather a good new: we can update SCI_GETTEXTRANGE, SCI_FINDTEXT, and SCI_FORMATRANGE to SCI_GETTEXTRANGEFULL, SCI_FINDTEXTFULL, and SCI_FORMATRANGEFULL respectively. OTOH, I would say to keep Notepad++'s modification `typedef intptr_t Sci_PositionCR;` though I don't see any impact if we restore it to `typedef long Sci_PositionCR;`. 

We will also encourage the plugin author update these API from their side, but it's not really urgent.

What's your option? Anyone in here has any suggestion?

rdipardo:
@chcg, @donho, the key word in the changelog is "duplicate". N++ *already* returns 64-bit character positions from the existing APIs &#x2014; nothing has changed except the name. The old messages can become type aliases for the new ones, like this:

~~~diff
@@ -473,10 +473,10 @@
 #define SCFIND_REGEXP 0x00200000
 #define SCFIND_POSIX 0x00400000
 #define SCFIND_CXX11REGEX 0x00800000
-#define SCI_FINDTEXT 2150
 #define SCI_FINDTEXTFULL 2196
-#define SCI_FORMATRANGE 2151
+#define SCI_FINDTEXT SCI_FINDTEXTFULL
 #define SCI_FORMATRANGEFULL 2777
+#define SCI_FORMATRANGE SCI_FORMATRANGEFULL
 #define SCI_GETFIRSTVISIBLELINE 2152
 #define SCI_GETLINE 2153
 #define SCI_GETLINECOUNT 2154
@@ -488,8 +488,8 @@
 #define SCI_GETMODIFY 2159
 #define SCI_SETSEL 2160
 #define SCI_GETSELTEXT 2161
-#define SCI_GETTEXTRANGE 2162
 #define SCI_GETTEXTRANGEFULL 2039
+#define SCI_GETTEXTRANGE SCI_GETTEXTRANGEFULL
 #define SCI_HIDESELECTION 2163
 #define SCI_POINTXFROMPOSITION 2164
 #define SCI_POINTYFROMPOSITION 2165
~~~

Completely removing the old messages would create a bigger mess than #11198, since even 32-bit plugins would stop working!

donho:
@rcdailey 
> nothing has changed except the name. 

No, it's not true. They are 2 different values in `scintilla.h` of v5.2.3:
```cpp
#define SCI_FINDTEXT 2150
#define SCI_FINDTEXTFULL 2196
```

> Completely removing the old messages would create a bigger mess than https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11198, since even 32-bit plugins would stop working!

The old messages are kept in Scintilla v5.2.3 so I see no reason to remove them from Scintilla of Notepad++.
All the plugins are compiled independently of Notepad++, they migrate these 3 API or not, they keep having always 64 bits result which is compatible with Notepad++. So the plugins won't stop working.

rdipardo:
*Apologies to @rcdailey for the accidental mention*

---

 > They are 2 different values in scintilla.h of v5.2.3

Plugins call the message *by name*, not *by integer value*. The *name* is still `SCI_FINDTEXT`. 
I was suggesting that the *name* be used to dispatch `SCI_FINDTEXTFULL` instead; a future call to `SCI_FINDTEXT` will get a 64-bit result like it has done since 8.3, and no plugin code needs refactoring.

> The old messages are kept in Scintilla v5.2.3 so I see no reason to remove them from Scintilla of Notepad++.

According to the [mailing list announcement][0]:

> SCI_GETTEXTRANGE will be deprecated at some point (likely 6.0) and __*removed even later (7.0?)*__.

Sounds like the old messages will be around for some time yet.

[0]: https://groups.google.com/g/scintilla-interest/c/mPLwYdC0-FE#:~:text=SCI_GETTEXTRANGE%20will%20be%20deprecated%20at%20some%20point%20(likely%206.0)%20and%20removed%20even%20later%20(7.0%3F)

rdipardo:
> I was suggesting that the name be used to dispatch `SCI_FINDTEXTFULL` instead

Of course that would be an *implementation* hack, probably touching `scintilla/src/Editor.cxx` &#x2014; not a *header* hack, as I wrongly said before. See my [edited comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11734#issuecomment-1140113901).

chcg:
Yes, with scintilla update 5.2.3 (keeping N++ change of Sci_PositionCR from long to intptr_t)
+ usage of the new interfaces

donho:
Auto-updating of v8.4.2 has just been triggered.
This PR will be integrated in 7-10 days if there's no critical issue in v8.4.2

donho:
@Yaron10
**Fix problem with horizontally inverted glyphs with buffered drawing and WS_EX_LAYOUTRTL set on Win32 GDI. [Feature #1435](https://sourceforge.net/p/scintilla/feature-requests/1435/).**
in https://www.scintilla.org/ScintillaHistory.html

Any idea about what it improves in Notepad++?

Yaron10:
@donho,

It seems they're referring to https://sourceforge.net/p/csscintilla/csscintilla/ci/fa80f5d42d95cb95030f587b00174471d946353b/.
But that commit was actually merged in an earlier Scintilla version.

If I'm correct, there shouldn't be any change (or potential change) in Notepad++.

scintilla/win32/Scintilla.vcxproj
```diff
@@ -95,6 +94,7 @@
     </ClCompile>
     <Link>
       <LinkTimeCodeGeneration>Default</LinkTimeCodeGeneration>
+      <CETCompat>true</CETCompat>
```
donho, line 97: Why is this line added?

donho (changes requested):
Please provide the argument for adding "CETCOMPAT" tag in vsproj file.

scintilla/win32/Scintilla.vcxproj
chcg: See Release Notes (https://www.scintilla.org/ScintillaHistory.html):
- Fix ARM64 builds with Visual C++ due to unsupported CETCOMPAT flag. [Bug #2324](https://sourceforge.net/p/scintilla/bugs/2324/).

The setting is already there at the common section, so it was just splitted out to the individual settings for x86, x64 because the feature is not available for arm64 (see also https://docs.microsoft.com/en-us/cpp/build/reference/cetcompat?view=msvc-170)

donho (approved):

PR #11738, opened 2022-05-28
Update german localisation

MarkusBodensee:
Add translation for item 6246 (current level folding/unfolding).
Add/update translations for dark mode items.
Synchronise translation of "commands" to "Befehle" to have same word in all places.


schnurlos:
Thanks for updating the german.xml - I'm very busy atm ... So any help as is done is nice!

Issue #11207, opened 2022-02-14
More granular theme control


UncleFred:
NotePad++ dark theme is great. Is it possible to control the indicated visual elements?

![2022-02-14_0-57-55](https://user-images.githubusercontent.com/5650331/153809691-1fd21c11-3cbd-4031-897f-239c5cda1d58.jpg)

ozone10:
IRC that yellow text is controlled by link color. And for other not possible.

Issue #11376, opened 2022-03-11
Settings for additional borders


depascalis:
### Description of the Issue
Borders are inconsistent and have too much contrast in dark mode for my taste.

### Suggestions
In short, add  color settings for various borders and fix inconsistent border widths. Maybe even add settings for custom border widths.

1. Dialog - Add settings for border color and width (some prefer high contrast borders, others prefer no borders at all)
2. List box and Text box
a. These borders are either `1.5 px`  wide or the control needs to shrink/grow `1 px` or the controls are misaligned `0.5 px`. 
b. Add setting for border color
3. Group box
a. Add setting for border color
b. Left align the title label (looks better when borders are invisible)
4. Buttons - Add setting for border color and width

Point 2 and 3 have the highest priority imo.

_Before_
![npp_borders_bef](https://user-images.githubusercontent.com/33980634/157950257-88e3f1f0-3722-48e2-b9a9-fc609d3f93e4.png)

_After_
![npp_borders_aft](https://user-images.githubusercontent.com/33980634/157950283-f0e4daef-0863-46fb-a282-bd7f83bbfa8e.png)

Suggestions above refer to controls in all dialogs, not only the Preferences dialog.

ozone10:
Notepad++ is using common control, which is "defined" by Windows OS.

So not possible, unless everything is owner/custom drawn, which would require a lot of code rewriting.

PR #11756, opened 2022-06-01
Use edge colors in dark mode for listbox and edit color border

ozone10:
tweaks to properly drawn listbox scrollbar, when using custom border,
add disabled edge color to dark mode preference dlg,
use disabled edge color for disabled combobox.

Before:
image
image

After
image
image
can 
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11207, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11376


ozone10:
gcc for x64 failed with error
```
C:\projects\notepad-plus-plus/PowerEditor/src/NppDarkMode.cpp:1754: undefined reference to `GetWindowSubclass'
C:\projects\notepad-plus-plus/PowerEditor/src/NppDarkMode.cpp:1798: undefined reference to `GetWindowSubclass'
```

anyone has idea how to fix it

rdipardo:
> anyone has idea how to fix it

I don't, but luckily StackOverflow does: <https://stackoverflow.com/a/16173383>

It seems GCC needs help selecting an appropriate [Common Control DLL Version](https://docs.microsoft.com/en-gb/windows/win32/controls/common-control-versions?redirectedfrom=MSDN#common-control-dll-versions-numbers). so:

~~~cpp
#define _WIN32_IE 0x0900
~~~

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -1547,6 +1740,43 @@ namespace NppDarkMode
...
+				bool isComboBox = (style & LBS_COMBOBOX) == LBS_COMBOBOX;
+				auto exStyle = ::GetWindowLongPtr(hwnd, GWL_EXSTYLE);
+				bool hasClientEdge = (exStyle & WS_EX_CLIENTEDGE) == WS_EX_CLIENTEDGE;
+
+				if (p.subclass && !isComboBox && hasClientEdge)
+				{
+					NppDarkMode::subclassCustomBorderForListBoxAndEditControls(hwnd);
+				}
+
+#ifndef __MINGW64__ // mingw build for 64 bit has issue with GetWindowSubclass, it is undefined
```
donho, line 1753: @ozone10 
Could you try `#define _WIN32_WINNT 0x502` instead?
Ref: https://stackoverflow.com/questions/60105064/setwindowsubclass-function-from-winapi

ozone10: Unfortunately it cause more issue. It needs higher number, but that also did not work.
I also tried `#define _WIN32_IE 0x0600`. 

But this is minor issue that code is mainly to avoid border flickering  when quickly moving to and out listbox.

ozone10: It's there in header but for some reason it's not used.

image

Issue #7735, opened 2019-12-14
Files with only CR are shown as separated lines, but not line-sorted correctly


bersbersbers:
### Description of the Issue
Sort the attached (Line Operations, Sort). The file consists of three lines (A, C, B), each terminated by only a carriage return (CR). In Hex, this is what this looks like:
```
41 0D 43 0D 42 0D
```

This is what it looks like in Notepad++:
![image](https://user-images.githubusercontent.com/12128514/70846640-a22df700-1e5b-11ea-8084-f58f7efd3f72.png)

So the file is displayed as three separate "lines", so line-sorting should be expected to sort them. However, line-sorting does not sort the "lines".

### Expected Behavior
Lines are sorted.

### Actual Behavior
They are not

### Debug Information
Notepad++ v7.8.1   (64-bit)
Build time : Oct 27 2019 - 22:57:19
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.535
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll 

Maybe related: #4736

xylographe:
The status bar reports "Unix (LF)" ?

bersbersbers:
@xylographe is that a question? Yes, the status bar reports "Unix (LF)".

The file is now attached by the way:
[bug.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/3963683/bug.txt)

sasumner:
@bersbersbers :

To @xylographe 's point, you are defining a "line" to be some text that ends with a LF (per your "choice" as shown in the status bar), and then you are creating some data which doesn't follow the line definition and are expecting it to be sorted correctly?

Suggest that you unify your example, so that line-endings used in the data match the line-ending type as shown on the status bar, and then see if sorting problems persist...

bersbersbers:
@sasumner I should add that I am not defining anything since "Unix (LF)" was not my own choice, it was Notepad++'s choice based on a slightly more complete example file that I am uploading now.
The file bug2.txt is
```
D(LF)
A(CR)
C(CR)
B(CR)
```

(A real-word example would be captured terminal output for some Unix process with progress:
```
Download started(LF)
0%(CR)
1%(CR)
2%(CR)
```
[...]

PR #11773, opened 2022-06-09
Make EOL (CRLF) display customizable

### LINKED ISSUE #11413: Feature request: Display e.g. CR/LF in a different colour
Recent Scintilla versions have support for displaying e.g. CR/LF in a different colour.

Search https://www.scintilla.org/ScintillaDoc.html for SC_REPRESENTATION_COLOUR and SCI_SETREPRESENTATIONCOLOUR

It would be great if this feature makes it into N++ as well.

Some messages on the Scintilla mailing list:
https://groups.google.com/g/scintilla-interest/c/-zLNwzJ18Q4/m/EZfyqGEaAQAJ
Not sure whether this is related. The search engine finds it if one searches for the SC_/SCI_ constants, but I don't see why:
https://groups.google.com/g/scintilla-interest/c/g1Ypo8qijJU/m/o9zZCGSSAAAJ

The now closed feature request for Scintilla: https://sourceforge.net/p/scintilla/feature-requests/1129/

donho:
Fix #11413


alankilborn:
How does one set a "Custom Color"?

Yaron10:
@alankilborn,

Is it working for you?

alankilborn:
>Is it working for you?

No.
All seems fine if one ignores the "Custom Color" checkbox (and leaves it unchecked).
But when "Custom Color" is ticked, any type of visible line-ending disappears, no matter what the color is specified as.  
And when "Default" is the active state, ticking "Custom Color" results in a no-show of any line-ending (doing this should have no effect in the "Default" state!).

Note:
I added this line to stylers.xml, after the "URL hovered" line to get the entry to show up in the Style Configurator:
`<WidgetStyle name="EOL custom color" styleID="0" fgColor="FF80C0" fontStyle="0" />`

Yaron10:
> But when Custom Color is ticked, any type of visible line-ending disappears

Interesting. I can not reproduce that.

My points:
1. `EOL custom color` availability in the Style Configurator _while by default `Custom Color` is unchecked_ - is highly confusing.
2. Why is `Custom Color` indented as if it depends on `Plain Text` being checked?

Thank you.

alankilborn:
>EOL custom color availability in the Style Configurator while by default Custom Color is unchecked - is highly confusing.

AFAIK _Style Configurator_ doesn't selectively show items based on other setup??

>Why is Custom Color indented as if it depends on Plain Text being checked?

I took this to mean that you can only set a custom color WHEN _Plain Text_ is the active selection -- perhaps I'm wrong about this.

Yaron10:
> AFAIK Style Configurator doesn't selectively show items based on other setup?

Do you have any other example to a style not changing anything **by default**?
I'm referring to default Settings. 

> I took this to mean that you can only set a custom color WHEN Plain Text is the active selection -- perhaps I'm wrong about this.

You can set a custom color even when `Default` (?) is selected

alankilborn:
>Do you have any other example to a style not changing anything by default?
I'm referring to default Settings.

I have absolutely no idea what you mean by that. :-)

>You can set a custom color even when Default (?) is selected

And I'm not sure what that means, either.

:-)

Sorry.

alankilborn:
One of the problems here is that _normally_ a person proposing a feature will describe it fully in the issue before creating a PR, including demo'ing the UI and discussing how it will work.  This is so Don can understand and "get interested" in the feature and how it will "feel" if adopted.  We don't have THAT need here (given the author of this PR), so...all we've got is guessing about how it is supposed to work. :-)

ozone10:
Default looks is inverted color for text and background
image
this is plain text
image

Unfortunately scintilla allows changing colors only in plain text mode.

alankilborn:
>scintilla allows changing colors only in plain text mode.

This is probably for the best because inverted color is too heavy (even if color was adjustable).
Plaintext with the correct color might just be something which would allow visible line-endings to remain on always!

ozone10:
I would prefer inverted looks but with adjusted transparency.

donho:
@ozone10 
> Unfortunately scintilla allows changing colors only in plain text mode.

The custom color can apply to both modes - you can try the latest commit (lite grey will apply to default & plain text).

donho:
@alankilborn 
> all we've got is guessing about how it is supposed to work. :-)

Which part? It means the GUI interface is not clear enough. Could you elaborate it?

donho:
@alankilborn 
> All seems fine if one ignores the Custom Color checkbox (and leaves it unchecked).
But when Custom Color is ticked, any type of visible line-ending disappears, no matter what the color is specified as.
And when Default is the active state, ticking Custom Color results in a no-show of any line-ending (doing this should have no effect in the Default state!).

Could you remove `<WidgetStyle name="EOL custom color" styleID="0" fgColor="FF80C0" fontStyle="0" />` from stylers.xml and try the latest commit please?

Yaron10:
>  Do you have any other example to a style not changing anything by default?

`Custom Color` is unchecked by default.
I think the average user would expect an immediate effect on changing `EOL custom color` _without changing a setting as well_.

> You can set a custom color even when Default (?) is selected.

When the RadioButton `Default` is selected you can still change `EOL custom color`.
So why is `Custom Color` indented?

PowerEditor/src/Parameters.cpp
```diff
@@ -3697,6 +3697,17 @@ bool NppParameters::feedStylerArray(TiXmlNode *node)
 			_widgetStyleArray.addStyler(styleID, childNode);
 		}
 	}
+	const Style* pStyle = _widgetStyleArray.findByName(TEXT("EOL custom color"));
+	if (!pStyle)
+	{
+		TiXmlNode* eolColorkNode = globalStyleRoot->InsertEndChild(TiXmlElement(TEXT("WidgetStyle")));
```
MarkusBodensee, line 3703: Should be:
`TiXmlNode* eolColorkNode = globalStyleRoot->InsertEndChild(TiXmlElement(TEXT("EOL custom color")));`

MarkusBodensee:
I just gave this PR a try and directly fall in love with it!! This new feature is somewhat great!

Two suggestions from my side:

- Is it better to set the default color of `White space symbol` as default color to `EOL custom color`, too? My expectation would be, that by default all Show Symbol options (ws/tab/eol) show up in the same color. In this way it would be visually even more clear that the CRLF at the end of the line are symbols and not letters.

- Remove the checkbox `Custom Color` completely from the settings? I think it's bloated and makes things more complicated. The color can be controlled via the Style Configurator. If a user really wants black symbols, the color can be changed there.


In a fresh installation, my expectation for this feature would be:
Choose: Menu View --> Show Symbols --> Show All Characters
Result: Spaces/Tabs and EOL show up in the same color

PowerEditor/src/WinControls/ColourPicker/WordStyleDlg.cpp
```diff
@@ -817,6 +817,34 @@ bool WordStyleDlg::selectThemeByName(const TCHAR* themeName)
...
+	::SendDlgItemMessage(_hSelf, IDC_LANGUAGES_LIST, LB_SETCURSEL, 0, i);
+	setStyleListFromLexer(static_cast<int>(i));
+
+	if (sections.size() == 1)
+		return true;
+
+	i = ::SendDlgItemMessage(_hSelf, IDC_STYLES_LIST, LB_FINDSTRING, (WPARAM)-1, (LPARAM)sections[1].c_str());
+	if (i == LB_ERR)
+		return false;
+	::SendDlgItemMessage(_hSelf, IDC_STYLES_LIST, LB_SETCURSEL, 0, i);
```
donho, line 842: @ozone10 
This line doesn't take effect. I would appreciate if you can take a look on it.

donho: Problem found:
It should be "(_hSelf, IDC_STYLES_LIST, LB_SETCURSEL, i, 0);" instead of "(_hSelf, IDC_STYLES_LIST, LB_SETCURSEL, 0, i);"
I just copied the existent code - it means there are some bugs....

ozone10: @donho 
I was away. IIRC WPARAM is used for handle and number value, and LPARAM for pointer. This should help next time.

donho:
@MarkusBodensee 
> Is it better to set the default color of White space symbol as default color to EOL custom color, too? 

>  Remove the checkbox Custom Color completely from the settings? 

Due to the complains on every change of behaviour, let's keep the current default behaviour - users can change it easily via settings anyway.

alankilborn:
>Due to the complains on every change of behaviour,

But this often happens in the natural course of a feature's development.
@MarkusBodensee had some GREAT suggestions, I thought.
Often you get one chance to make a feature the BEST it can be -- after that it is rarely revisited.

shriprem:
The checkboxes for Bold, Italics & Underline are currently enabled for the EOL style. But currently they don't have any desired effect.

Either those controls need to be disabled or their effects need to be implemented as well.

alankilborn:
After commit https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11773/commits/6ed242aad6372656c694b9638bc76bf6af34fb38 interestingly I get nothing showing when I select any custom color.

Working ok (no custom color):

image

Not working ok (with custom color, CRLF totally disappears!):

image

donho:
@alankilborn 
Could you check if you have "EOL custom color" entry in Style Configurator ?

image

donho:
@alankilborn 
OK, I can reproduce it now.
It's due to, once again, the DirectWrite!
I'll see what I can do about it.

MarkusBodensee:
@donho
> Due to the complains on every change of behaviour, let's keep the current default behaviour.

I think, here it is kind of different. Current default behaviour without this PR is counterintuitive. But until now it was not possible to fix or improve the situation. So the wrong behaviour only became default because of lack of correct alternative. The "default" black symbol are much too heavy in normal use, and in black "plain text" mode, it is nearly impossible to distinguish between normal text.
grafik

Less options for this feature is more user-friendliness. Nevertheless, I understand if you have doubts in changing the default behaviour, even if I would strongly vote for changing it and remove checkbox.

> users can change it easily via settings anyway.

Apart from the above. Instead of using lite grey for CRLF in all themes/styles, can you consider to apply the color of CRLF to be the same like White space symbol in the themes for consistency reasons? Some themes don't match well with lite grey as default, but of course do with white space color.

grafik
To get this well looking style (independent from "default" or "plain text" mode), you need to change 3 settings at 3 different places (much work):
- Change view -->Show symbol
- Change setting to use Custom Color
- Change style to use white space color

Thank you for reading :)

donho:
@alankilborn 
The issue has been fixed in the last commit.

MarkusBodensee:
> The checkboxes for Bold, Italics & Underline are currently enabled for the EOL style. But currently they don't have any desired effect.
> 
> Either those controls need to be disabled or their effects need to be implemented as well.

@donho @Yaron10 
The same is true for White space symbol style. The checkboxes are enabled without visual effect while changing them.

donho:
@MarkusBodensee 
About your requests, I remain my arguments and decision.
Thank you for your suggestion anyway.

alankilborn:
Nice work; one will actually be able to turn visible line-endings on and leave them on.

alankilborn:
@donho This is somewhat related, but tell me if I should create a new issue (or to forget about it):  What about a way to show ONLY the line-endings that don't match what is currently shown in the status bar for the file?  

So, for example if the status bar shows `Windows (CR LF)`, then only lines that have LF or CR as endings will show `LF` or `CR` respectively, visibly; lines that end with CR+LF will show nothing.  This draws user attention quickly to a "mixed line endings" problem (and it often is a big problem when it happens).

This would reduce what I consider a problem:  Scintilla has a notion of a file's line-ending type, but doesn't enforce that type.

This idea is not mine and has its roots here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/7735#issuecomment-1106879435

donho:
@alankilborn 
> So, for example if the status bar shows `Windows (CR LF)`, then only lines that have LF or CR as endings will show `LF` or `CR` respectively, visibly; lines that end with CR+LF will show nothing. This draws user attention quickly to a "mixed line endings" problem (and it often is a big problem when it happens).

AFAIK, there's no way to display part of kind of EOL in Scintilla. Event it existed, it would be better to display all.
User can use search features to find the different kind of EOL anyway.

PR #11790, opened 2022-06-13
Add NPPM_CREATELEXER message for plugins to instantiate lexers

### LINKED ISSUE #11789: Plugins can no longer set lexer for custom Scintilla object
### Description of the Issue
Notepad++ currently provides the `NPPM_CREATESCINTILLAHANDLE` call for plugins to create Scintilla instances to be used in plugins. However with the move to Scintilla v5+, a plugin can no longer use `SCI_SETLEXER` since this has been removed by Scintilla.

Notepad++ needs a new API call so that Notepad++ can call Lexilla's `CreateLexer()` function and return the pointer to the ILexer object it to the plugin.

1. A plugin creates a new Scintilla handle using `NPPM_CREATESCINTILLAHANDLE`
2. Call `SCI_SETLEXER` on the Scintilla handle
3. Nothing happens since Scintilla ignores this message

### Debug Information
```
Notepad++ v8.4.2   (64-bit)
Build time : May 29 2022 - 16:47:30
Command Line : $COMMAND_LINE_PLACEHOLDER$
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.1706
Current ANSI codepage : 1252
Plugins : 
    BetterMultiSelection (1.5)
    ElasticTabstops (1.4)
    ezLCDLua (1.1)
    LuaScript (0.12)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    NppMenuSearch (0.9.3)
```

dail8859:
Closes #11789

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -540,6 +540,9 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// Users should call it with commandLineStr as NULL to get the required number of TCHAR (not including the terminating nul character),
 	// allocate commandLineStr buffer with the return value + 1, then call it again to get the current command line string.
 
+	#define NPPM_CREATELEXER (NPPMSG + 110)
+	// void* NPPN_CREATELEXER(0, const TCHAR *lexer_name)
+	// Returns the ILexer pointer created by Lexilla
 
 #define VAR_NOT_RECOGNIZED 0
 #define FULL_CURRENT_PATH 1
@@ -734,4 +737,3 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	//scnNotification->nmhdr.code = NPPN_CMDLINEPLUGINMSG;
 	//scnNotification->nmhdr.hwndFrom = hwndNpp;
 	//scnNotification->nmhdr.idFrom = pluginMessage; //where pluginMessage is pointer of type wchar_t
-
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1386,6 +1386,13 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return cmdLineString.length();
 		}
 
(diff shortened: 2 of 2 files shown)


donho:
@dail8859 
Doesn't the usage `NPPM_MENUCOMMAND` for running one of Language menu items get job done? This is the most stable way to set language IMO.

But if setting lexer by plugin itself is really necessary, I prefer to use another straightforward name.
`NPPM_CREATELEXER` might not be explicit for users. Furthermore "ILexer pointer" is useful only for calling `SCI_SETLEXER`.
I would name it `NPPM_SETLANGBYID` or `NPPM_SETLEXERBYID`, so users (plugin authors) know what it is about.

```
#define NPPM_SETLANGBYID (NPPMSG + 110)
	// BOOL NPPM_SETLANGBYID(0, INT LangID)
	// Returns the TRUE if successful, otherwise FALSE
	// here are the lexer IDs:
	// https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h#L23
```

And you can just call [setLexerFromLangID](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L1943).

But once again, I don't think this API is necessary. What's your need for your plugin(s)?

dail8859:
> Doesn't the usage NPPM_MENUCOMMAND for running one of Language menu items get job done?

No. `NPPM_MENUCOMMAND` is for setting the language on a Scintilla/Buffer instance that is managed by Notepad++. If a plugin uses `NPPM_CREATESCINTILLAHANDLE`, all that Notepad++ is just return a handle to the instance for the plugin to have full control over.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/79e766755d9d466704c535f99eb3ce99ffbbb1bf/PowerEditor/src/NppBigSwitch.cpp#L1469-L1472

For example LuaScript has Notepad++ create two custom Scintilla instances by using `NPPM_CREATESCINTILLAHANDLE`. However there is no way for LuaScript to set the lexer to Lua on these two instances any more:

https://github.com/dail8859/LuaScript/blob/0e8f8cf9feb25beffe8076a81f3b6811cbd8fac3/src/LuaConsole.cpp#L209

This code needs a new API since it would require a call to lexilla's `CreateLexer()` function by the plugin.

donho:
@dail8859 
OK, I see.

Issue #8446, opened 2020-06-20
Move focus to editor on left-clicking a tab


Yaron10:
**STR:**
Give keyboard focus to a "not-editor" element (e.g. the Incremental Search Bar). 
Left-click the current document's tab.

**Result:**
The editor does not get the focus.

**Expected Result:** 
The expected behavior is that the editor should get the focus.
Also, practically it can be useful in many cases (e.g. if you want to move focus to the editor without changing the caret position or if you want to keep a selection). 

***
Notepad++ v7.8.3   (32-bit)
Build time : Jan 17 2020 - 18:20:02
Path : C:\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

Uhf7:
What you mean here is that the input focus does not switch to the edit window, if the tab of the current document is clicked, right? If the tab of another document than the current one is clicked (and the document switches), then the input focus goes to the edit window already.

Yaron10:
Correct. I've edited my Post.

notdodgeball:
That would be very useful, specially the "not changing the caret position" thing.

alankilborn:
Really cheesy way to do it:

When input focus isn't in editor, right-click the tab and then immediately press Esc (to cancel the tab context menu).  At this point, input focus is in the editor of the chosen tab.

alankilborn:
The original posting here is actually a pretty good idea.
I will prepare a PR implementing the behavior (since there has been no other action on this is over 2 years).

PR #11808, opened 2022-06-17
Add an option allows to show only 1 entry per found line in search result of find in files

### LINKED ISSUE #1499: Single line of output to Find-result window for lines with 1+ hits

### LINKED ISSUE #2481: Display only one row in "Find All in Current Document"

### LINKED ISSUE #2987: Feature Request: Find All to return only unique lines even if some lines have multiple matches of the text or expression

### LINKED ISSUE #3224: Find in active/files returns same line multiple times

### LINKED ISSUE #4525: Suggestion: don't duplicate lines with multiple hits in "find results"

### LINKED ISSUE #5547: Find All in... feature request (formerly #1499)

donho:
This option can be activated in "Searching" section of Preferences dialog.
Also mark several found occurrences in the same entry - user can click on the marked occurrence to jump among found result in the found line.

It's an alternative implementation of #11705

Fix #2481, fix #1499, fix #5547, fix #2987, fix #4525, fix #3224


PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -334,6 +334,7 @@ BEGIN
     CONTROL         "Find dialog remains open after search that outputs to results window",IDC_CHECK_FINDDLG_ALWAYS_VISIBLE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP,37,40,350,10
     CONTROL         "Confirm Replace All in All Opened Documents",IDC_CHECK_CONFIRMREPLOPENDOCS, "Button", BS_AUTOCHECKBOX | WS_TABSTOP,37,55,350,10
     CONTROL         "Replace: Don't move to the following occurrence", IDC_CHECK_REPLACEANDSTOP, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 70, 350, 10
+    CONTROL         "Find in files search result: show only one entry per found line", IDC_CHECK_SHOWONCEPERFOUNDLINE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 85, 350, 10
```
alankilborn, line 337: But it doesn't just affect "Find in files", it affects any search that outputs to "Search results" window, e.g. "Find all in current doc".  Suggest change wording to "Find All search result: ...".

alankilborn:
This feature (like "Replace: Don't move to the following occurrence") is much better as a "live action" choice in the "Find" window.

It's fine the way it is for people that ALWAYS want it only one way or the other, but power users will want it only sometimes -- and that means it is better served as a "Find" window choice.  Users that ALWAYS want it the same way can just tick it (or not) there ("Find" window) and then forever after forget about it.

PowerEditor/src/WinControls/Preference/preference.rc
pryrt: Or maybe even just "Search Result window: show only one entry per found line", to strongly emphasize it's for that particular window.

Yaron10:
@alankilborn,

Should it be very complex to show all entries per line and enable the user to jump to the corresponding occurrence?
This is how it's implemented in other editors.

תמונה

alankilborn:
@Yaron10 

I think in your comment you have 2 ideas that maybe weren't expressed in much detail:

* have an additional level of folding for multiple matches per line (if this were the case I would be happy with the enabling option in the Preferences rather than in the Find window!)
* each match is a link which when double clicked will jump to that specific match in the source file

Or, am I misinterpreting what you were trying to convey?

These are excellent suggestions but we don't know how far @donho wants to take this.

vinsworldcom:
Maybe I'm missing something, but I thought this was addressed in #11705 and I liked that implementation with a right-click option right in the Search Results window.

alankilborn:
> Maybe I'm missing something, but I thought this was addressed in #11705 and I liked that implementation with a right-click option right in the Search Results window.

Apparently @donho didn't like that implementation and so instead chose to roll his own.

Yaron10:
@alankilborn,

> have an additional level of folding for multiple matches per line...

I was not thinking of that.
Do you find it useful?

> each match is a link which when double clicked will jump to that specific match in the source file.

A link has crossed my mind, and indeed might be very helpful. But I'm not sure it's necessary.

My suggestion is quite simple:

If line `20` has `15` matches, display that line **once** (with all `15` matches) in the Search Results window.
If the user double-clicks the first match, go to the first occurrence in line `20` in the editor.
If the user double-clicks the second match, go to the second occurrence. And so on.

Using the keyboard: the caret would have to be visible in the Search Results window.
If the user presses Enter when the caret is on the third match, go to the third occurrence. And so on.

This is the implementation in EditPad. 
I think it's the best and most intuitive behavior. If you all agree, why add another setting for a behavior which isn't ideal either?

I don't really know how complex the implementation in NPP should be. :)

Thank you.

PowerEditor/src/WinControls/Preference/preference.rc
donho: @pryrt 
Done: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808/commits/a051f6c465ad050d7f9e255f8b13eaff46181791

donho:
@Yaron10 
> Should it be very complex to show all entries per line and enable the user to jump to the corresponding occurrence?
> This is how it's implemented in other editors.

I'll try, but I promise nothing.

donho:
@vinsworldcom 
> I liked that implementation with a right-click option right in the Search Results window.

@alankilborn 
> This feature (like "Replace: Don't move to the following occurrence") is much better as a "live action" choice in the "Find" window.

There are already too many actions on Search result window's context menu. I prefer to keep it on where it should be and keep the context menu clean.

donho:
@Yaron10 
> I don't really know how complex the implementation in NPP should be. :)

UI designer's dreams are usually the nightmares of devs.

I will see what I can do about your ideas...

donho:
(Sorry for tagging everyone here)
@Yaron10 @vinsworldcom @pryrt 
Could you check the issue #11814 and confirm me in the thread of #11814?

@alankilborn has already confirmed this bug, but it's strange to me that we have no such bug report whereas v8.4 and other later versions have come out since about 2 months ago.

Yaron10:
@donho,

> I will see what I can do about your ideas...

👍 
That would be a major improvement.

Thank you and best of luck.

alankilborn:
@Yaron10 

The "folding" idea could give everyone what they want, at all times.
When a search is run, we'd see in the output a single line under the pathname; this line would contain ALL matches for a line and would be a fold parent.  When unfolded it would be the parent of many one-hit-per-line children.
A preference setting would be needed for "initially collapsed or initially expanded".
This seems ideal but would a parent containing all matches per line being present annoy those that like the current one-hit-per-line format?
I don't know, maybe it is too "complex" to be implemented.

>If line 20 has 15 matches, display that line once (with all 15 matches) in the Search Results window.
If the user double-clicks the first match, go to the first occurrence in line 20 in the editor.
If the user double-clicks the second match, go to the second occurrence. And so on.

This is a fine idea; it is true that actual "links" aren't necessary.

Yaron10:
@alankilborn,

> When a search is run, we'd see in the output a single line under the pathname; this line would contain ALL matches for a line and would be a fold parent...

I understand it better now.
And if those "ALL matches" were "clickable", would you still want the one-hit-per-line children?
Or IOW: if my suggestion (again, not mine but EditPad's behavior) is implemented, do you think that the current one-hit-per-line results should be kept as an option?

alankilborn:
>And if those "ALL matches" were "clickable", would you still want the one-hit-per-line children?

I would.  Otherwise, for long lines, it is a guessing (or horizontal scrolling) game to see if I'm missing anything that I can't currently see off to the right.

>do you think that the current one-hit-per-line results should be kept as an option?

I do, and I'm sure @donho would agree with that as well -- devs don't take away things, only give new tools as options.

Yaron10:
> I would. Otherwise, for long lines, it is a guessing (or horizontal scrolling) game to see if I'm missing anything that I can't currently see off to the right.

Good point. 
Personally - if the caret was visible and the last double-clicked-match was "selected", it wouldn't be an issue.
But, to each his own. :)

AndiKiissel:
Now F4 "Next Search Result" (and Shift+F4 "Previous...") gives only first occurrences in lines. By me this is serious regression.

Yaron10:
@donho,

I've just briefly tested the new implementation.
It seems brilliant. One of the most important improvements sines I've started using NPP. 👍 

1. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808#issuecomment-1159539035
How about making the caret visible and uniquely highlight the last double-clicked-match (or last Enter-match) in the Search Results window?
(So that we have "current/last line" and "current/last match").

2. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808#issuecomment-1159562908.

I might not be by PC tomorrow. Could you please keep it open for a few more days?

Thank you for this great work.

donho:
@AndiKiissel 
> Now F4 "Next Search Result" (and Shift+F4 "Previous...") gives only first occurrences in lines. By me this is serious regression.

It's not a regression, but a limitation of this new option.

"Regression" means that the old feature has bug and you cannot use it anymore.
If you want to access to every occurrence with F4, just don't enable this new feature.

donho:
@Yaron10 
> I might not be by PC tomorrow. Could you please keep it open for a few more days?

No problem.

> How about making the caret visible

The original design of Search Result windows is make it as a control but not a editor, so I prefer to keep hiding caret.

> and uniquely highlight the last double-clicked-match (or last Enter-match) in the Search Results window?

This request is technically impossible, or (at least) very hard to implement. I keep it in mind but it won't be in this PR for sure.

> https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808#issuecomment-1159562908.

Here is my reply about your point 2:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808#issuecomment-1159900690

As I said, caret won't be added for the esthetic reason (and I don't really think it's a way for quick access).
However, I've implemented the shortcut 1-9 in the latest commit. Please let me know if you have other shortcut idea.

AndiKiissel:
> > Now F4 "Next Search Result" (and Shift+F4 "Previous...") gives only first occurrences in lines. By me this is serious regression.
> 
> It's not a regression, but a limitation of this new option.
> 
I understand, but then F4 not "Next Search Result" anymore and should be renamed, if this new option is selected. I'm sorry!

alankilborn:
Is it difficult to make F4 work in this new mode?  I tend to agree with @AndiKiissel that it sure seems like it should work the same in either mode.

alankilborn:
I notice with the addition of the keypresses for 1, 2, 3, ..., 9, that if you have 4 matches in a line, then pressing 5 or greater switches to the edit window and selects the text of the first match.  Would it be better to ignore such keypresses entirely (i.e., remain with focus in the _Search results_ window and do nothing -- no change -- in the editor window)?

Yaron10:
@donho,

> and uniquely highlight the last double-clicked-match (or last Enter-match)
>> This request is technically impossible, or (at least) very hard to implement.

I can't argue with that. :)
Should marking the match as "selected" be easier?

> As I said, caret won't be added for the esthetic reason (and I don't really think it's a way for quick access).

The reason to my suggestion was not quick access.
Please think of the following case:
תמונה
You place the (invisible) caret on match 15 (single-click; Search Results still has focus). 
You take a break and return after 30 minutes.
Looking at the screen now, you have no indication whatsoever which match NPP should jump to on pressing Enter.
I wouldn't consider it a bug, but I do think it's a major flaw.

> However, I've implemented the shortcut 1-9 in the latest commit.

It's a nice idea.
But is it really going to be useful? Would the user start counting the matches? Is that "quick access"?

Thanks again.

MarkusBodensee:
@Yaron10

> It's a nice idea. But is it really going to be useful? Would the user start counting the matches? Is that "quick access"?

It definitely is, consider following situation: Have many similar lines in a file, have same keyword multiple times in each line, you want to edit every line at the 4th occurrence of the keyword. So just go through the search results, press 4 and edit.

Yaron10:
@MarkusBodensee,

👍 
Very nice.

Thank you.

Yaron10:
@alankilborn,

> I notice with the addition of the keypresses for 1, 2, 3, ..., 9, that if you have 4 matches in a line, then pressing 5 or greater switches to the edit window and selects the text of the first match. Would it be better to ignore such keypresses entirely (i.e., remain with focus in the Search results window and do nothing -- no change -- in the editor window)?

Now that I understand the 1-9 shortcuts better...

If you double-click a word between matches or after the last match, you jump to the first match.
I thought it was a nice implementation.

If your suggestion is excepted, that behavior would have to be changed too.
I think that either behavior is legit, and it can be determined by the easier-implementation / shorter-code factor.
Personally, I tend to prefer the current behavior.

alankilborn:
@Yaron10 

>>if you have 4 matches in a line, then pressing 5 or greater switches to the edit window and selects the text of the first match

>I tend to prefer the current behavior.

Thus forcing my brain to work overtime, always striving to answer the question "am I on the 5th match, or are there really less than 5 and I'm back at the first match?"  I mean, I suppose I could keep a watch on the column indicator in the status bar...

Perhaps with very "regular" data (as @MarkusBodensee considers above), this is a minimal concern.

I don't know that I would use this entire feature (and that's fine, I don't have to).

But a larger concern is:  Is the implementation going to satisfy the many people that have requested this over the years...

Yaron10:
@alankilborn,

> Thus forcing my brain to work overtime...

That was my initial thought about the 1-9 shortcuts.

> But a larger concern is: Is the implementation going to satisfy the many people that have requested this over the years...

Why not?
If Don can manage to highlight the current match, it would be the best implementation IMO.
What else would you change?

Yaron10:
@AndiKiissel,

I think it's Work in Progress.

donho:
@AndiKiissel 
> Now I'm very afraid of all this implementation...

Don't be rude.
Read the comment of commit before whining...

donho:
@alankilborn @Yaron10 
> But a larger concern is: Is the implementation going to satisfy the many people that have requested this over the years...

I do believe people prefer this new behaviour. BTW, since Next Search result & Previous Search result will be implemented, the new behaviour will set as default.

donho:
@Yaron10 
> Should marking the match as "selected" be easier?

I'll try it. And once again, I promise nothing.

AndiKiissel:
@donho 

> Don't be rude. Read the comment of commit before whining...

I'm sorry.
I didn't mean to be rude.
Please excuse me.
Notepad++ is the best text editor I've tried and you have done great job.

Yaron10:
@donho,

> I'll try it. And once again, I promise nothing.

Thank you for trying. If it can't be done, it can't be done.

Allow me another point:
Uniquely highlighting the last double-clicked-match would be much better than "selecting" it.
A single-click or a new selection would remove last match "selection".

alankilborn:
I will try it, but I'll wait until the build errors are fixed.

donho:
@Yaron10 
Fixed. Thank you.

Yaron10:
@donho,

Very nice! 👍 
I do appreciate it.

I've tested https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808/commits/4c9a2efdc9ee8faed6709aaf189aa3652b17c97b.

I'll try that and the latest commit more thoroughly tomorrow.

In the meantime:

1. Shouldn't the last match be selected on Enter as well?

2. > The current match is selected in the Search Results window on pressing F4.
On the other hand - it removes an existing selection in the Search Results window.

What do you think?

donho:
@Yaron10 
> The current match is selected in the Search Results window on pressing F4.
On the other hand - it removes an existing selection in the Search Results window.

OK, I see now what you meant.
It's inevitable that the old selection done by user is unselected after F4.
In the v8.4.2 it has the same behaviour - if you hit F4 all selection is gone.

AndiKiissel:
Very good!
One bug found. Steps to reproduce:
*	open new file, write "x x x", find all "x", press Shift+F4 ... 
*	now add second line "x x x", find all "x", press Shift+F4 ...

alankilborn:
@AndiKiissel 

So when reporting something, it is common practice to not just list steps to reproduce, but also the behavior observed and the expected behavior.  All you gave were some steps...

When I do the actions of _only_ your first bullet point, what I get on the first Shift+F4 press is a fold-collapsing of the search results.   Another Shift+F4 press results in fold-expansion of the results.  Continuing to press just keeps toggling the fold state of the output.  Strange!

When I do the actions of both bullet points, the first press of Shift+F4 moves me to the last match on Line 1; I suppose I would have expected to be placed on the final match in Line 2.

Other quick notes:

With multiple horizontal "entities" now on a single line of output, it feels almost imperative that a caret should be visible in the _Search results_ window, aesthetic reasons for not having it are understood, but...  I think I'm agreeing with @Yaron10 on this.

alankilborn:
I noticed that when running a new search, old results aren't fold-collapsed.  I think this was mentioned before with an earlier commit, but it is still happening with the most recent commit.

Yaron10:
@donho,

> OK, I see now what you meant.
It's inevitable that the old selection done by user is unselected after F4.
In the v8.4.2 it has the same behaviour - if you hit F4 all selection is gone.

You're absolutely right. 👍 
So selecting the match in the Search Results window on F4 is a nice bounce.

Trying to properly test the various cases, I find it necessary to have the caret visible in the Search Results window.
Even if you don't want to reconsider and have it visible in the final implementation, could you temporarily enable it for testing?

And apropos reconsidering...
Do you think the 1-9 shortcuts are that important and useful?
Is it worth the more complex code and possible glitches? 

Thanks again for your on-going work.

***

@alankilborn,

> With multiple horizontal "entities" now on a single line of output, it feels almost imperative that a caret should be visible in the Search results window, aesthetic reasons for not having it are understood, but... I think I'm agreeing with @Yaron10 on this.

👍

alankilborn:
Another point about caret:  There are dedicated keyboardists out there -- those people that won't touch a mouse.  If those people are now expected to move horizontally in the _Search results_ window, how will they know where they are?

This was not a concern with one-match-per-line (old way) as they were merely moving up and down and the current-line-highlight was showing them which entire line they were on.  But now (all-matches-per-line) current horizontal location can be important.

@Yaron10, maybe we've beat the horse to death on this caret thing? :-)

AndiKiissel:
@alankilborn

I've already been many times sorry here and have to be again because my english is not good enough to write detailed descriptions and explanations. But I'll try...

@donho 

Bug description: if first step after "Find all in current document" or "Find all in all opened documents" is Shift+F4 then last occurrents (or many last occurrentses if they are in one line) will not be highlighted.

alankilborn:
@AndiKiissel 

>I've already been many times sorry here and have to be again because my english is not good enough to write detailed descriptions and explanations. But I'll try...

Ok, I'm sorry if it felt like I was being harsh to you.

donho:
@Yaron10 
> Do you think the 1-9 shortcuts are that important and useful?
Is it worth the more complex code and possible glitches?

It's not complex - just 2 lines of code.
What are possible glitches?

> Trying to properly test the various cases, I find it necessary to have the caret visible in the Search Results window.
Even if you don't want to reconsider and have it visible in the final implementation, could you temporarily enable it for testing?

OK, I will enable it temporarily for testing only.

alankilborn:
> OK, I will enable it temporarily for testing only.

Yaron10 will like it so much, removing it will be like trying to take a bone from a dog. :-)

( @Yaron10 , I have a script for enabling that caret, it is just a bit longish; but if your caret doesn't make it into the final PR, I'll share )

donho:
@alankilborn 
> Another point about caret: There are dedicated keyboardists out there -- those people that won't touch a mouse. If those people are now expected to move horizontally in the _Search results_ window, how will they know where they are?
> 
> This was not a concern with one-match-per-line (old way) as they were merely moving up and down and the current-line-highlight was showing them which entire line they were on. But now (all-matches-per-line) current horizontal location can be important.

That's a valid point. The caret will be there.

> @Yaron10, maybe we've beat the horse to death on this caret thing? :-)

If you provide valid argument, you can beat a dead horse to resurrection. :)

Yaron10:
@vitali-kurylovich,

Thank you for https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11705 which was the trigger of the current PR.
It would be appreciated if you could join us in testing and advising.

donho:
I just found a bug:

If you have 4 different files with the same content opened in Notepad++, and you search one occurrence which appears only once in the content, you've got only 1 line result instead of 4 lines.

Could you guys confirm the bug please?

Yaron10:
> I just found a bug:

👍 

תמונה

donho:
Please check the latest commit.
The change in this commit:
1. Fix only 1 search result instead of several ones.
2. Fix Shift-F4 issue.
3. Add caret.

donho:
@Yaron10 
If you think it's all OK for you, I would like to merge this PR.
Just give me a sign.

Yaron10:
@donho,

Great work. Thank you.

Some quick notes:

1. https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e5f366567dc23f33914ecd37d8abf75f2c8b8278/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L3001
Caret in Find in Finders.

2. I think it's _highly important_ that the caret should be visible on moving the focus to the Search Results window by clicking on its title bar (somehow related: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8446). 

3. @AndiKiissel's case:
Please copy & paste and search for `Test`.
```
Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test 

Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test Test 
```
Press `Shift+F4`.

Result:
You jump to the first `Test` on line 3.
(Same result on manually placing the caret on the headers).

***

This is a major new feature/implementation.
Please let us test it for a few more days.

AndiKiissel:
Shift-F4 still doesn't work right, if pressed right after "Find all..."

image

Yaron10:
1. When the Find dialog is first launched, the Search Results window does not get focus.
2. I've rarely used F4 before. I've just noticed that when pressing it in Find in Finders, you jump to the next result in the original finder (just an observation; same behavior in the old implementation).

AndiKiissel:
@Yaron10

Usage varies.
I use **only** F4 / Shift+F4 and old implementation was perfect for me.

Yaron10:
@AndiKiissel,

I usually use the mouse when possible.
The new implementation will be perfect too. And much better IMO.

Yaron10:
@AndiKiissel,

Could you please keep testing the F4 / Shift+F4 and the 1-9 shortcuts?
It's not "natural" to me.

Thank you.

AndiKiissel:
@Yaron10 

1-9 shortcuts are not "natural" to me too. With 10000 characters in line and 33 matches (sometimes) they will be not very helpful.
(I work with XHTML-files from unpacked EPUBs)

alankilborn:
> I usually use the mouse when possible.
> The new implementation will be perfect too. And much better IMO.

I too am a mouse user for this (at least for the "old" way).
In fact, when the talk started about F4 and Shift+F4, I actually had to go look up what those were!

But I wouldn't use "mouse" and "perfect" in subsequent sentences, about this. :-)
I may actually have to become an F4 user if I'm to use this feature.
Why?
Because using the mouse now ("new" way) is tied to visually seeing the match desired, and hitting it with a double-click to go to its source.
This is perfectly fine when all of your matches are shown on one screen-width.
But, as stated before, I often deal with longish lines.
Horizontal scrolling will be needed before I can see some matches -- have I accidentally h-scrolled too quickly and missed one?  Why am I now horizontal scrolling at all; with "old" way I didn't have to.
F4 instead of mouse would guarantee I haven't missed one...a workflow change.

BTW, double clicking between two matches brings one to the first match on a line -- why can't it bring one to the _next_ match?  Meaning that if you double click somewhere between match 14 and 15, isn't it logical that it takes you to match 15?  This would make the mouse much more viable for me with the "new" way.  Moving to match 1 like this seems very much a carryover from the "old" way, and IMO isn't natural or intuitive.

I guess the only thing I've brought to this discussion is achieving the caret for Yaron. :-)

Yaron10:
@alankilborn,

This might be a heretical question:
What did we actually gain by upgrading to Scintilla 5?

I'm joking!

donho:
The latest commit has fix the Shift-F4 jumping problem. There's still a small glitch on a specific condition (no space or other characters after the last search occurrence), but it should work better globally.

donho:
> > BTW, double clicking between two matches brings one to the first match on a line -- why can't it bring one to the next match? Meaning that if you double click somewhere between match 14 and 15, isn't it logical that it takes you to match 15?
> 
> I'll try to reciprocate for the caret. :) Don, what do you think? Would it be very complex to implement it that way?

Let's do it in another PR if it's possible.
What I want is make this PR right and merge it ASAP.

donho:
> This is a major new feature/implementation.
Please let us test it for a few more days.

OK. There won't be any major update if there's no bug found.

AndiKiissel:
@donho 

> The latest commit has fix the Shift-F4 jumping problem. There's still a small glitch on a specific condition (no space or other characters after the last search occurrence), but it should work better globally.

Searching must be reliable. If I search something I must be sure that I found all and see them all. Zero tolerance. Any "small glitch" is a fault.
Real life example: find and correct incorrect paragraph ends `\w</p>`. This happens at the end of line. Shift+F4 does not highlight the very last match:

image

Why I don't do this in "normal" order with F4? Because every added symbol will move next matches and next F4 highlight is shifted. If this happens 30 times in one file...
But this doesn't mater, text editor must be reliable in every situation.
I still hope that you can fix even "small glitches".

Yaron10:
@donho,

Thank you this new excellent feature. Much appreciated. 👍 

It's natural to have some issues in such a major implementation.
If indeed there are any, they'd be found in "real life" work.
As far as I'm concerned, this PR can be merged.

I'll leave the F4/Shift+F4 issue to you and @AndiKiissel.
(I do agree that "small glitches" can be a serious problem in one's workflow).

***

@alankilborn,

> BTW, double clicking between two matches brings one to the first match on a line -- why can't it bring one to the next match? Meaning that if you double click somewhere between match 14 and 15, isn't it logical that it takes you to match 15?
>> Let's do it in another PR if it's possible.

I still owe you. :)
(Thank you Don).

Yaron10:
@donho,

I've just encountered a major issue in RTL.

**STR:**
Change direction to RTL.
Copy & paste `את את את` and search for `את`.

**Result:**
Old implementation.
תמונה

New implementation.
תמונה

Thank you.

Yaron10:
@rdipardo,

Don wants to merge this PR ASAP.
Could you please help us testing it?

Thank you.

donho:
@Yaron10 
The Search result works badly under RTL in v8.4.1:

image

on which I can do nothing unfortunately :(

donho:
@AndiKiissel 
> I still hope that you can fix even "small glitches".

It's not that I don't want to fix this bug.
I have no solution with the current algorithm, OTOH, if you've got the solution, you're welcome to do your contribution.

Yaron10:
@donho,

Please don't test in English. 
That's an old issue, and not only in the Search Result but also in the editor. 

Copy & paste `את את את` and search for `את`.
Works correctly with the old implementation.
Could you try it?

Thank you.

Yaron10:
@AndiKiissel,

Does the problem exist if you use the old implementation (i.e. latest commit here and unchecking "Search Result window: show only one entry per found line" in Preferences)?

Yaron10:
@donho,

> Copy & paste את את את and search for את.
It works correctly with the old implementation.

And with "Direct Write" _disabled_.

donho:
@Yaron10 

>>    Copy & paste את את את and search for את.
>>    It works correctly with the old implementation.

>    And with "Direct Write" disabled.

Yes, I see the bug. I can reproduce it. As I said, I've no idea to make it work since it's the Microsoft function to mirror text.

Yaron10:
@donho,

Thank you for looking into it. I appreciate your time and work.

Most of my searches are in LTR. I'll use the old implementation for RTL.

Should I open an issue for that?

donho:
@Yaron10 
> Should I open an issue for that?

Sure, you can do it at least.

Yaron10:
I'll open an issue once this PR is merged.

Thanks again.

alankilborn:
>There is no "old implementation" anymore, as I understand.

Unchecking "Search Result window: show only one entry per found line" in Preferences is the "old implementation" _functionality_, but if you mean _code_, then it is true that the implementation is somewhat changed.

But the old _functionality_ should still be there, even though not the same exact code.
As far as I know that is the intent.

AndiKiissel:
I think, I give up and stay on last master-branch compilation before this PR, whitch I can trust.
Thank you all!

Yaron10:
I often argue with devs about the best way to implement various features.
But in this case we all agree that both the approach and the implementation are very good.
So, we've found some issues/inconveniences. - But this is natural/expected in every major work, isn't it?

I can only thank Don again for his work, and hope that the issues be fixed.

donho:
I just found a unexpected behaviour regarding the new feature:

1. Open a C++ file of Notepad++ project.
2. Took a line end of `;` and next line with `};` as following:
```
	void changeLineHiliteMode(bool enableSlider);
};
```
    Select from `;` to `};` then hit "Ctrl-F".
3. Hit the button "Find All in Current Document" - the results are trucked which is already the old behaviour so it's OK 
4. Double click on on of search result - The selection will cover to next line on the search result, which may lead confusion.

@Yaron10 
What's your opinion of such behaviour?

donho:
> @Yaron10
What's your opinion of such behaviour?

I think the behaviour of latest commit is fair enough.

Yaron10:
@donho,

The fix seems perfect, but I liked the catch even more. 👍 

Two enhancements. If you prefer, I can mention them later when this PR is merged.

**For keyboard users:**
When you execute a search, the caret is placed at the beginning of the header.

תמונה

On pressing Enter, the search is collapsed. 
It's not very likely the user would want to do that.

How about placing the caret at the beginning of the first match?

תמונה

(You can argue that not having the yellow background is a minus). 

**For mouse users:**
I've mentioned it before. I'll try to clarify it.
You place the caret on match 35 (in Search Results) with the intention of double-clicking on it later.
You move to the editor (i.e. the editor has focus now).
You want to get back to match 35 in the Search Results window.
- You have no way of doing that _with the mouse_ unless you have a phenomenal visual memory.

How about moving the focus to `_scintView` on clicking the Search Results' title bar?  

Thank you.

alankilborn:
@donho said:

> * Select from `;` to `};` then hit "Ctrl-F".
> * Hit the button "Find All in Current Document"

When I try this test (in "Normal" search mode), I notice that after pressing the "FAiCD" button, that my search mode has been changed to "Extended" ??

What really caused me to notice this is that my "Find What" box changed from looking like this (expected):

image

to looking like this (unexpected):

image

@Yaron10 or anyone...can you duplicate?

Yaron10:
@alankilborn,

> When I try this test (in "Normal" search mode), I notice that after pressing the "FAiCD" button, that my search mode has been changed to "Extended" ??

This has been the behavior for the past few years.
I remember dealing with the relevant code when I tweaked the Incremental Search Bar for my personal use.

BTW, for this and for Don's catch - just select 2 lines.

***

@donho,

Regarding the RTL issue:
How about the following temporary workaround?

We're checking before displaying the search-results if the active document is RTL.
Could you switch the mode from "new-implementation" to "old-implementation" if it's RTL?

Again - if you prefer, we can discuss it later.

alankilborn:
>>When I try this test (in "Normal" search mode), I notice that after pressing the "FAiCD" button, that my search mode has been changed to "Extended" ??

>This has been the behavior for the past few years.

IMO this is just insane. :-)
Is this _desired_ behavior by @donho ?  If not, should I open an issue?  I suspect this IS going to be "desired behavior". :-(

I suppose I have not noticed it because I rarely do searches that include multiline data.
If it going to switch modes on me, I would rather it switched to regex (because there's a great likelihood that my next search is going to be a regex one, saving me a step).

Yaron10:
@alankilborn,

I think it's a good point.
Are you only suggesting to restore the search-mode, or also not displaying `\r\n` in the `Find what` ComboBox?

***

I'm taking a break for a few hours. :)

alankilborn:
> Are you only suggesting to restore the search-mode, or also not displaying `\r\n` in the `Find what` ComboBox?

I'm suggesting not mucking with search mode (thus there is nothing to restore) or find-what.
And if there is no mucking, there would be no displaying of `\r\n` in find-what.
(Only coders know what `\r\n` is anyway)

Yaron10:
@alankilborn,

If your file contains the following **2** lines.

```
Test
Test
```

If you select the first line including its EOL, open the Find dialog and search - the result would _correctly_ be 1.
If you then manually select the `Normal` RadioButton, the result would be 0. 
NPP automatically selects the `Extended` search-mode so that the result should be 1.
NPP adds `\r\n` so that on executing the search again you'd get the same result.

How would you implement that?

donho:
@AndiKiissel 
The latest commit should fix last entry skipped issue by Shift-F4.
Could you try it out?

donho:
> When I try this test (in "Normal" search mode), I notice that after pressing the "FAiCD" button, that my search mode has been changed to "Extended" ??

@alankilborn
Please create another issue for this and tag me and @Yaron10 .  
Let's talk abut it outside of this issue - there are too much info here.

donho:
> We're checking before displaying the search-results if the active document is RTL. Could you switch the mode from "new-implementation" to "old-implementation" if it's RTL?
> 
> Again - if you prefer, we can discuss it later.

@Yaron10 
The old function has been removed due to the new one.
Regarding RTL, it's rather a bug and there's no reason to restore the old function.
Please open an issue for that - in the meanwhile, I'll take a look again to see what can I do for it.

Yaron10:
@donho,

> The latest commit should fix last entry skipped issue by Shift-F4.

👍

> Regarding RTL, it's rather a bug
...
I'll take a look again to see what can I do for it.

Great. Thank you and best of luck with that too.

> The old function has been removed due to the new one.

I meant that if it's RTL, navigate and display the results as if the new setting `Search Result window: show only one entry per found line` was unchecked.
Obviously, a proper fix would be much better.

donho:
@Yaron10 
> I meant that if it's RTL, navigate and display the results as if the new setting `Search Result window: show only one entry per found line` was unchecked.
> Obviously, a proper fix would be much better.

I have taken few hours to investigate this bug. The problem comes from lexer of LexSearchResult (LexSearchResult.cxx) in which an assert was triggered by Lexilla in debug mode. So in release mode it just stops colourising.

I tried to adapt to RTL mode in LexSearchResult.cxx but in vain. Even I could make it, there would be a problem to pass "isRTL" argument to LexSearchResult lexer.

I think your suggestion will be the solution finally, but instead of unchecking "Search Result window: show only one entry per found line", I would modify its label to "Search Result window: show only one entry per found line (not applied to RTL mode)".

What do you think?

Yaron10:
@donho,

Thank you for looking into it and for the explanation.

I did not mean to actually uncheck the new setting.
Let's make sure we're on the same page.

If `Search Result window: show only one entry per found line (not applied to RTL mode)` is checked.
Search 1, NOT isRTL: one entry per found line.
Search 2, isRTL: _old behavior_ multiple entries per found line.
Search 3, NOT isRTL: one entry per found line.

_Please remember "Find in Finders" too_. :)

***

The Shift+F4 issue seems to be fixed. 👍

AndiKiissel:
@donho 

> The latest commit should fix last entry skipped issue by Shift-F4. Could you try it out?

Yes, it's fixed. Well done!

Perhaps a stupid question - Replace All (in All Opened Documents) is not connected with this PR, right?

donho:
@AndiKiissel 
> Perhaps a stupid question - Replace All (in All Opened Documents) is not connected with this PR, right?

I confirm that Replace All (in All Opened Documents) is not concerning.

donho:
@Yaron10 
> If `Search Result window: show only one entry per found line (not applied to RTL mode)` is checked.
> Search 1, NOT isRTL: one entry per found line.
> Search 2, isRTL: _old behavior_ multiple entries per found line.
> Search 3, NOT isRTL: one entry per found line.

Done in the latest commit!

Yaron10:
@donho,

Great.
Thanks again for this major improvement to NPP. 👍

yasmise:
@donho 
This new feature is not working well in Japanese, Simplified Chinese and Korean. In English and French, it works. I didn't test in other languages.
Screen shot is shown below. Tested with English and Japanese.
Please see the orange markups. Found texts are wrongly highlighted.
Could you take a look at?

searchresult

donho:
@yasmise 
> This new feature is not working well in Japanese, Simplified Chinese and Korean. In English and French, it works. I didn't test in other languages.

Are you saying that the new feature is not working well in Japanese, Simplified Chinese and Korean **environment**?
Because the test you showed is English text. 

Have you ever tried in Simplified Chinese and Korean Windows? You are under Japanese Windows 11 I suppose?

donho:
Thanks to @alankilborn 's explanation, I see better the problem.

@yasmise Could you try to change the language to Japanese, restart Notepad++ and try it again?

yasmise:
@alankilborn is right. I tested by just switching language setting of Notepad++. OS is Windows 11 in Japanese. Thanks for your explanation.

@donho I just tried on another PC with portable version, but result is same. Here is steps to reproduce.

1. Download Portable (7z) version of Notepad++ 8.4.2 x64 from official site.
2. Extract to `C:\Users\******\Downloads\npp`.
3. Download AppVeyor artifact. This time, I used ed4bb1a93e763001aac842698fcde0856ba8b0bc ([download page is here](https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/44012124/job/s9tnk5s7bxpca69l/artifacts))
4. Move `Notepad++.x64.Release.exe` to the location of step 2, and launch it.
5. Change language to Japanese (日本語), and restart Notepad++.
6. Type `test test test` to new document.
7. `Search` > `Find...` > set `Find what` to `test` > `Find All in Current Document`. (Actually, these command name is shown in Japanese)
8. Screenshot is shown below.
screenshot

Debug info is here.
Notepad++ v8.4.2   (64-bit)
Build time : Jun 28 2022 - 17:35:29
Path : C:\Users\******\Downloads\npp\Notepad++.x64.Release.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.1766
Current ANSI codepage : 932
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

donho:
@yasmise 
Thank you for testing and report.
Could you open an issue for it?
I'll fix it.
Thank you

yasmise:
@donho 
I opened issue #11855. Step 6 and 7 of STR is slightly changed, since new STR describes issue better, in my opinion.
Many thanks.

AndiKiissel:
@donho 

First Shift+F4 doesn't catch very last match if searched string contains EOL.
For example:
open N++ last change.log
press Ctrl+F
Find what: `.\r\n1`, Search Mode: Extended, press "Find All in Current Document"
press Shift+F4
Last match goes to line "17. Update..." but it's not selected.

image

Used last AppVeyor release 7743.

Yaron10:
> @AndiKiissel's last case can be fixed by implementing https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808#issuecomment-1166244765.

That is correct if the first match is also the first word of the line.

If the first match is NOT the first word of the line, there's another issue.

**STR:**
Place the caret at the beginning of a match.
Press F4.
Place the caret at the beginning of a match.
Press Shift+F4.

**Result:**
In both cases NPP jumps to the match next to the caret.

**Expected:**
On pressing Shift+F4 NPP should jump to the _previous_ match.

donho:
@AndiKiissel 

> First Shift+F4 doesn't catch very last match if searched string contains EOL.

It won't be fixed in the release which is coming up.
Please open an issue by referencing @Yaron10 's reply.
I may work on that after v8.4.3 release.

Yaron10:
@donho,

> This bug has been fixed in [one of latest commits](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/059bca3315dbd71fbf235c8bc87eaacbd87a3cec). Since the bug is fixed, the RTL restriction has been removed as well.

👍 
It's working like a charm now.
I'm glad you've found the real cause. Thanks again for your time and work.

vlakoff:
Although this new mode can be useful, I think the setting shouldn't be enabled by default.

It makes very easy to overlook that an entry occurs several times on a same line. There is no hint that a line contains several matches, apart from the highlights that the user has to look for. Actually, for long lines that don't fit on screen width, there is no hint at all, unless the user horizontally scrolls the results pane while looking for the highlights...

alankilborn:
@vlakoff

> Although this new mode can be useful, I think the setting shouldn't be enabled by default.

> It makes very easy to overlook that an entry occurs several times on a same line. There is no hint that a line contains several matches, apart from the highlights that the user has to look for. Actually, for long lines that don't fit on screen width, there is no hint at all, unless the user horizontally scrolls the results pane while looking for the highlights...

You are absolutely right about this.  I tried to point it out several times as a deficiency of the feature, but my comments were not heeded.  Perhaps you could open a brand new issue on only this topic?

PR #11824, opened 2022-06-22
fix empty .NFO file crash

### LINKED ISSUE #11820: opening a empty .nfo file produce a instant crash with v8.4.2
### Description of the Issue

Since my update to v8.4.2 Notepad++ crashes when I open a empty .nfo File

1. Rightclick in Windows Explorer and select New->Notepad++ Document
2. Rename the new 0 Byte file from .txt to .nfo
3. Doubleclick the .nfo file (If you have selected Notepad++ as default tool für .nfo) or start Notepad++ and open the .nfo file

### Expected Behavior

An empty file to edit

### Actual Behavior

Notepad++ stops working and crash.

### Debug Information

I tested the behaviour on different Computers and with the 32&64 Bit version of Notepad++
That makes no difference.

xomx:
Fixes #11820 .

The infinite recursion tour:
- an empty nfo-file has been loaded, at the end the `activateBuffer` has been called
- the code flow reached the `ScintillaEditView::defineDocType` (`case L_ASCII`), where the file encoding has been changed to `NPP_CP_DOS_437` and the `IDM_FILE_RELOAD` has been sent after
- during the reloading the `FileManager::loadFileData` was called and there is this important part:
```
	//for empty files, if the default for new files is UTF8, and "Apply to opened ANSI files" is set, apply it
	if (fileSize == 0)
	{
		if (ndds._unicodeMode == uniCookie && ndds._openAnsiAsUtf8)
			fileFormat._encoding = SC_CP_UTF8;
	}
```
- it means that in this constellation the `fileFormat._encoding` has been reset back to `SC_CP_UTF8`
- so in the following `activateBuffer` call we ended up on the exactly same place and the file encoding has been set back to `NPP_CP_DOS_437` again and another `IDM_FILE_RELOAD` has been sent ...

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1601,7 +1601,7 @@ bool FileManager::loadFileData(Document doc, int64_t fileSize, const TCHAR * fil
 		fileFormat._eolFormat = ndds._format;
 
 		//for empty files, if the default for new files is UTF8, and "Apply to opened ANSI files" is set, apply it
-		if (fileSize == 0)
+		if ((fileSize == 0) && (fileFormat._encoding < 1))
 		{
 			if (ndds._unicodeMode == uniCookie && ndds._openAnsiAsUtf8)
 				fileFormat._encoding = SC_CP_UTF8;


Yaron10:
@xomx,

Well done. 👍 

I suppose you've tried it without https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d1347d501451e10c5d988282ca681760d2c2abff.

Thank you for the quick fix.

xomx:
@Yaron10 

> I suppose you've tried it without [d1347d5](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d1347d501451e10c5d988282ca681760d2c2abff).

Nope, with. Yesterday I downloaded the latest build for the translation purposes and when I used it, I tried also open an empty NFO just for the curiosity. My older N++ (8.3.3) does not crash on this.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1601,7 +1601,7 @@ bool FileManager::loadFileData(Document doc, int64_t fileSize, const TCHAR * fil
 		fileFormat._eolFormat = ndds._format;
 
 		//for empty files, if the default for new files is UTF8, and "Apply to opened ANSI files" is set, apply it
-		if (fileSize == 0)
+		if ((fileSize == 0) && (fileFormat._encoding < 1))
```
donho, line 1604: @xomx 
Thank you for your PR and very detail [explanation](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11824#issue-1280131509). 
As @Yaron10 said, I cannot reproduce the crash after https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d1347d501451e10c5d988282ca681760d2c2abff.

My question is why this specific condition: `(fileFormat._encoding < 1)` ?

donho: @xomx 
OK, I see. By putting a messagebox for displaying encoding value, I saw the infinite loop - the condition to filter 1 and greater than 1 works to stop the infinite loop.

@Yaron10 
 https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d1347d501451e10c5d988282ca681760d2c2abff (the optimization of speed) prevent program from crash (infinite loop), but the loop does happen, it just stop it at the end of ~10 times.

Yaron10:
@donho,

PR #11833, opened 2022-06-23
better specify the cmdline interpreter

### LINKED ISSUE #11818: FAW - "CMD here" does not work if there is a subdirectory named cmd
1. Make sure Settings->Preferences->Default Directory->Open all files of folder instead of ... is NOT checked
2. Create a directory called faw_test wherever you like
3. Create a subdirectory called cmd
4. Drag the faw_test directory onto Npp

This should result in something like

image

Select faw_test and use the context menu to run "Cmd here"

image

### Expected Behavior
a cmd shell opens in faw_test directory

### Actual Behavior
an EXPLORER opens in the CMD directory

### Debug Information
Notepad++ v8.4.2   (64-bit)
Build time : May 29 2022 - 16:47:30
Path : D:\Tests\npp\_latest\x64\notepad++.exe
Command Line : $COMMAND_LINE_PLACEHOLDER$
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Build : 7601.24546
Current ANSI codepage : 1252
Plugins : 
    NppLspClient (0.0.12)

xomx:
Fixes #11818 .

This solves a possible confusion between the cmdline interpreter and the possible folder named "cmd"' for the ShellExecute WINAPI.

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -841,7 +841,7 @@ struct NppGUI final
 	generic_string _definedSessionExt;
 	generic_string _definedWorkspaceExt;
 
-	generic_string _commandLineInterpreter = TEXT("cmd");
+	generic_string _commandLineInterpreter = TEXT("cmd.exe");
 
 	struct AutoUpdateOptions
 	{
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -46,6 +46,7 @@
 
 #endif
 
+#define CMD_INTERPRETER TEXT("%WINDIR%\\System32\\cmd.exe")
 
 class NativeLangSpeaker;
 
@@ -841,7 +842,7 @@ struct NppGUI final
 	generic_string _definedSessionExt;
 	generic_string _definedWorkspaceExt;
 
-	generic_string _commandLineInterpreter = TEXT("cmd.exe");
+	generic_string _commandLineInterpreter = CMD_INTERPRETER;
 
 	struct AutoUpdateOptions
 	{
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -46,7 +46,7 @@
 
 #endif
 
-#define CMD_INTERPRETER TEXT("%WINDIR%\\System32\\cmd.exe")
+#define CMD_INTERPRETER TEXT("%COMSPEC%")
 
 class NativeLangSpeaker;
 
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -6690,7 +6690,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 	}
 
 	// <GUIConfig name="commandLineInterpreter">powershell</GUIConfig>
-	if (_nppGUI._commandLineInterpreter.compare(TEXT("cmd")))
+	if (_nppGUI._commandLineInterpreter.compare(CMD_INTERPRETER))
 	{
 		TiXmlElement *GUIConfigElement = (newGUIRoot->InsertEndChild(TiXmlElement(TEXT("GUIConfig"))))->ToElement();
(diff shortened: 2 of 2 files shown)


dail8859:
Kind of a dumb question, but what about if there is a `cmd.exe` file in the local directory? Is this situation even avoidable?

https://docs.microsoft.com/en-us/windows/win32/shell/app-registration?redirectedfrom=MSDN#finding-an-application-executable

```
The file is sought in the following locations:

    The current working directory.
    The Windows directory only (no subdirectories are searched).
    The Windows\System32 directory.
    Directories listed in the PATH environment variable.
    Recommended: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths
```

xomx:
@dail8859 

> Kind of a dumb question, but what about if there is a `cmd.exe` file in the local directory?

It is a legitimate question, ShellExecute then really tries to open any file named "cmd.exe" from the dir selected (e.g. from the "faw_test" dir example mentioned in the issue). So if somebody copies there e.g. the Windows "notepad.exe" file, renames it to the "cmd.exe", then the Windows Notepad will be opened instead of the command prompt requested.


> Is this situation even avoidable?

Yes, when we use "%WINDIR%\\System32\\cmd.exe" instead of my current "cmd.exe". I tried that and it works ok.

But there is still something I need to understand - during my testing I have found that there is a N++ GUIConfig for this cmdline interpreter, which I cannot find in the N++ GUI or in the correspondent "config.xml": [commandLineInterpreter](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3b0479309729cbcdd29fa2a458b9b3e842543312/PowerEditor/src/Parameters.cpp#L6692)

And at the next line is a hard-coded comparison to `TEXT("cmd")`, so it needs the same adjustment. I will probably create a common define for it to be used at both places.

dail8859:
@xomx 

> But there is still something I need to understand

Luckily the documentation is kept up really well.

https://npp-user-manual.org/docs/preferences/#preferences-for-advanced-users

It is for advanced users that want to manually change the interpreter in the config file themselves.

vinsworldcom:
> It is for advanced users that want to manually change the interpreter in the config file themselves.

I do that:

````
C:\usr\bin\npp64\config.xml
115:         <GUIConfig name="commandLineInterpreter">powershell</GUIConfig>
````

Cheersl

**Comment by MarkusBodensee** [comment]
> Kind of a dumb question, but what about if there is a `cmd.exe` file in the local directory? Is this situation even avoidable?

And the next one... What happens now with a folder named `cmd.exe` in the local directory. I think same happens like mentioned in the issue, folder opens in explorer.

**Comment by xomx** [comment]
@MarkusBodensee 

Yes, the "cmd.exe" folder will be opened in the explorer.
It is interesting that when one creates both "cmd" and "cmd.exe" subfolders, the explorer opens the "cmd.exe" one.

My latest version ("%WINDIR%\System32\cmd.exe") solves even that (I checked).

File: PowerEditor/src/Parameters.h [hunk 1]
```diff
@@ -46,6 +46,7 @@
 
 #endif
 
+#define CMD_INTERPRETER TEXT("%WINDIR%\\System32\\cmd.exe")
```
MarkusBodensee, line 49: Better use constants in C++, instead of defines ?

`const TCHAR cmdInterpreter[] = TEXT("%WINDIR%\\System32\\cmd.exe");`

xomx: It is equivalent (the above `#define...` will become the same `const TCHAR*`).

tarkett2018:
@xomx 
> Yes, when we use "%WINDIR%\System32\cmd.exe" instead of my current "cmd.exe". I tried that and it works ok.

Only a suggestion: You can also use the environment variable %COMSPEC%. Then you get directly the command line interpreter which can be different in mixed x86/x64 systems.

xomx:
@tarkett2018 

Nice, that is even better, I will use it.

x86/x64 was not a problem - the WinOS redirector handles that for 32/64-bit apps.

PR #11842, opened 2022-06-25
Themes: Synchronise color of EOL and Whitespaces

### LINKED ISSUE #11840: Themes: Default setting of EOL is hard to distinguish from background or normal text
### Description of the Issue
Use new feature to modify EOL appearance and apply custom color. Try it with different themes in default settings. The EOL marks are either hard to distinguish from background or hard to distinguish from normal text.

1. Go to Settings --> Preferences --> Editing --> EOL (CRLF)
2. Choice of Default or Plain Text is not important, look at both
3. Check the Custom Color checkbox
4. Type some text with spaces, tabs and line ends
```
UPPER CASE TEXT
  		  
	CRLFCRLF

```

### Expected Behavior
A good look and feel with default settings.

### Actual Behavior
The EOL marks are either hard to distinguish from background or hard to distinguish from normal text with default settings. See Screenshot.

### Debug Information
I used the portable v8.4.2, copied artifact from master branch of Jun 18 2022 - 16:26:48, and copied the themes from master branch

Notepad++ v8.4.2   (64-bit)
Build time : Jun 18 2022 - 16:26:48
Path : C:\npp.8.4.2.portable.x64\Notepad++.x64.Release.exe
Command Line :
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Build : 7601.24546
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

grafik

MarkusBodensee:
Fix #11840

Before:
grafik

After:
grafik

This synchronises the color of EOL and Whitespace as default, so all symbols have same color which is most likely user expectation. This is IMO a good choice, because it is individually adapted to each theme instead of grey everywhere.

Checked it with normal text and also some source code files, for example, xml, ini, c/c++/c#, md. It looks quite well.

What do others think on this approach? Suggestions?

### DIFF
--- a/PowerEditor/installer/themes/Bespin.xml
+++ b/PowerEditor/installer/themes/Bespin.xml
@@ -803,6 +803,6 @@ Credits:
         <WidgetStyle name="Active tab text" styleID="0" fgColor="000000" bgColor="8000FF" />
         <WidgetStyle name="Inactive tabs" styleID="0" fgColor="808080" bgColor="C0C0C0" />
         <WidgetStyle name="URL hovered" styleID="0" fgColor="FFFFFF" />
-        <WidgetStyle name="EOL custom color" styleID="0" fgColor="808080" />
+        <WidgetStyle name="EOL custom color" styleID="0" fgColor="FCAF3E" />
     </GlobalStyles>
 </NotepadPlus>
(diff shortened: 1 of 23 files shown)


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11840#issuecomment-1166395682
I strongly dislike the new colour - even we need to adapt it for themes, it's preferable using the colour less bright.

PR #11858, opened 2022-06-29
Update chineseSimplified.xml

popcorner:


donho (changes requested):
Please remove the translation of "(not applied to RTL mode)".

popcorner:
the translation of "(not applied to RTL mode)" is removed.

donho:
@popcorner 
It's timing! Thank you!

PR #11871, opened 2022-07-04
Update czech.xml for v8.4.3

xomx:
Added only the id 6906 - "Search Result window: show only one entry per found line".


xomx:
Thank you Don for that feature!

PR #11885, opened 2022-07-08
Translate some installer stuff

Blueberryy:


donho:
There is no way to make installer translatable.
NSIS has its own multi-language system.

PR #11888, opened 2022-07-08
Update portuguese.xml

hugok79:
For commit: 059bca3315dbd71fbf235c8bc87eaacbd87a3cec


donho:
Thank you for the update.
Could you add one more entry which has been added in the commit: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/67ab4d55276ae99e632f19d1035de70cc022e3b9#diff-a6ac7ceba70d88bf1547fd1defd760bd55052dcdb78c44f9d46d99ef1f450472R551

Otherwise, no problem. I will take this PR as it is. The new entry can be added in the future release.

hugok79:
@donho Done.

Issue #11774, opened 2022-06-09
"Find in Files" is initially unresponsive when searching a large folder structure


UweKeim:
### Description of the Issue

When using the “Find in Files” dialog to search through a large folder structure, the dialog becomes unresponsive right after starting the searching by clicking the “Find All” button.

<img width="432" alt="image" src="https://user-images.githubusercontent.com/415439/172814057-4fec49dc-be1e-4c6c-9024-2752ec654f18.png">

In my opinion this seems to be the case because when starting the searching first a directory tree seems to be read fully into memory before actually starting to search at all.

For me this is a bug.

I would love to have a “Find in Files” dialog that is responsive in every stage of the search. I.e. even when initially reading in the the folder structure and not only when later searching inside the actual files.

### Steps to Reproduce the Issue

1. Open the "Find in Files" dialog
2. Set the "Directory" field to a folder with lots of files and/or sub-folders
3. Click the "Find all" button

### Expected Behavior

The "Find in Files" dialog should start searching/scanning the folders and files and stay responsive (i.e. redraw itself, be able to handle user inputs like canceling the find).

### Actual Behavior

The "Find in Files" dialog remains unresponsive for several seconds up to a lot of seconds or even minutes, depending on the machine power and the size of the folder structure to search in.

### Debug Information

Notepad++ v8.4.2   (64-bit)
Build time : May 29 2022 - 16:47:30
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : $COMMAND_LINE_PLACEHOLDER$
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 21H2
OS Build : 22000.708
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

alankilborn:
Related Community link: https://community.notepad-plus-plus.org/topic/23055

donho:
@UweKeim 
Find in files not responsive problem is rather on the 2nd (and after) launch, but not the very first launch.
Could you confirm me the description above?

donho:
@UweKeim 
When I said "the 1st time", I meant with Notepad++ fresh launch, the 1st time you launch Find in files.
Could you close your Notepad++, launch it again then run the operation Find in Files?

UweKeim:
[...]

PR #11889, opened 2022-07-08
Fix crash during Search result

### LINKED ISSUE #11883: [Regression] Wrong lexing in Search Results
**STR:**
Find in Files -> Search All.
Again, Find in Files -> Search All.

**Result:**
The second search-results are not lexed properly, and NPP may even crash.

תמונה

**Caused by:** https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88.

donho:
Fix #11883 partially.

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -963,6 +963,9 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				{
 					BufferID idd = _mainDocTab.getBufferByIndex(id);
 					Buffer * buf = MainFileManager.getBufferByID(idd);
+					if (buf == nullptr)
+						return FALSE;
+
 					tipTmp = buf->getFullPathName();
 
 					if (tipTmp.length() >= tipMaxLen)
@@ -975,6 +978,9 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				{
 					BufferID idd = _subDocTab.getBufferByIndex(id);
 					Buffer * buf = MainFileManager.getBufferByID(idd);
+					if (buf == nullptr)
+						return FALSE;
+
 					tipTmp = buf->getFullPathName();
 
 					if (tipTmp.length() >= tipMaxLen)
--- a/lexilla/lexers/LexSearchResult.cxx
+++ b/lexilla/lexers/LexSearchResult.cxx
@@ -79,7 +79,7 @@ static void ColouriseSearchResultLine(SearchResultMarkings* pMarkings, char *lin
 		
 		int currentState = SCE_SEARCHRESULT_DEFAULT;
 
-		if (pMarkings->_length >= linenum)
+		if (linenum >= pMarkings->_length)
 			return;
 		SearchResultMarkingLine miLine = pMarkings->_markings[linenum];
 
--- a/lexilla/lexers/LexSearchResult.cxx
+++ b/lexilla/lexers/LexSearchResult.cxx
@@ -77,8 +77,10 @@ static void ColouriseSearchResultLine(SearchResultMarkings* pMarkings, char *lin
 
 		styler.ColourTo(startLine + currentPos - 1, SCE_SEARCHRESULT_LINE_NUMBER);
 		
-		int currentStat = SCE_SEARCHRESULT_DEFAULT;
+		int currentState = SCE_SEARCHRESULT_DEFAULT;
 
+		if (pMarkings->_length >= linenum)
+			return;
 		SearchResultMarkingLine miLine = pMarkings->_markings[linenum];
 
 		for (std::pair<intptr_t, intptr_t> mi : miLine._segmentPostions)
@@ -92,11 +94,11 @@ static void ColouriseSearchResultLine(SearchResultMarkings* pMarkings, char *lin
 				if (match_end <= endPos)
 					styler.ColourTo(match_end, SCE_SEARCHRESULT_WORD2SEARCH);
 				else
-					currentStat = SCE_SEARCHRESULT_WORD2SEARCH;
+					currentState = SCE_SEARCHRESULT_WORD2SEARCH;
 			}
 		}
 
-		styler.ColourTo(endPos, currentStat);
+		styler.ColourTo(endPos, currentState);
 	}
 	else // every character - search header
 	{


Yaron10:
@donho,

The crash was not consistent (about 1 of 10 NPP startups).
Unfortunately I've just encountered a crash with the new binary. 

Thank you for looking into it.

Yaron10:
> Unfortunately I've just encountered a crash with the new binary.

Positive.

I might be wrong, but the crash seems to happen more frequently in NPP x32 (in x64 system).

Yaron10:
@ArkadiuszMichalski,

Could you please try reproducing the crash on your machine?

ArkadiuszMichalski:
Crash still exist:
```
Notepad++ v8.4.3   (32-bit)
Build time : Jul  8 2022 - 15:52:51
Path : E:\Download\_chrome\npp.8.4.3.portable\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 21H2
OS Build : 22000.778
Current ANSI codepage : 1250
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

```

donho:
@Yaron10 @Yaron10 
I cannot reproduce the crash at all with v8.4.3 x32 official portable version (neither with x32 Debug mode)

Could you provide a way to reproduce crash steadily please?

ArkadiuszMichalski:
@donho 
1. Open Notepad++
2. Open some file, like `license.txt`
3. Go to Find in Files (Ctrl+Shift+F) and mark `Follow current doc.`
4. Find some text.

If crash not happend after two searches close and run NPP again and repeat above steps. Currently I get crash after first run.

xomx:
@ArkadiuszMichalski 
I can reproduce the crash with your steps, but on the 32-bit N++ only, x64 does not crash.

ArkadiuszMichalski:
@xomx Try with version from this PR (looks like it breaks more often than the official one). If after catching a crash and after a few attempts, there is still no crash, try run system again (in the past I had a case  that after a few crashes in NPP they stopped appearing until the system restarted).

Yaron10:
@donho,

> Could you provide a way to reproduce crash steadily please?

That's the problem: it's inconsistent and can not be reproduced steadily.

You're trying to separate the crash from the wrong lexing. - Your knowledge and understanding are better than mine. 
Still, allow me to ask: did you find out why in "Find in Files" the lexing breaks after the first search? Perhaps it IS related to the random illusive crash?

@ArkadiuszMichalski has mentioned an important point:
> If crash not happend after two searches close and run NPP again and repeat above steps.

And also, try it with the x32 build.

Thank you.

donho:
> Try with version from this PR (looks like it breaks more often than the official one). 

It should be more stable though. In anyway, I believe it doesn't contribute to the crash.

> * Open Notepad++
> 
> * Open some file, like `license.txt`
> 
> * Go to Find in Files (Ctrl+Shift+F) and mark `Follow current doc.`
> 
> * Find some text.

Yes, now I can reproduce the crash with x86 official release. However, in debug mode (the same source code) I can see the syntax highlighting is lost, but no crash at all. 

It seems it's rather the race condition on Scintilla - the execution on debug binary is much slower; since Scintilla is not thread safe, when the execution speeds up, there might be several accesses to Scintilla, which lead crash.

Since I cannot find the crash reason from the debug, I'll revert  https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88 to see if the crash still happens.

Yaron10:
> I'll revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88 to see if the crash still happens.

We can all try it now with the previous commit.

Would you mind waiting a couple of days?
You or @xomx might still come up with a solution. :)

donho:
Found it!
I'll do a commit in 30 minutes!

donho:
The crash should be fixed in the latest commit.
There is still the colorizing problem though.
Once the crash-fix is confirmed, I will revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88 to fix the rest of problem.

Yaron10:
@donho,

The crash seems to be fixed. 👍 
Thank you.

I suppose you've noticed that with the recent commit the results in "Find in Files" _are lexed_ but wrongly.

> I will revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88 to fix the rest of problem.

Do you mean back to "no folding"?

Yaron10:
And the lexing in "Find All in Current Document" is broken.

donho:
@Yaron10 
Sorry my bad. I was too excited to commit the correct fix.
I just commit the correct fix. Please check the latest commit. Sorry again for the error.

ArkadiuszMichalski:
Crash was fixed.

Yaron10:
@donho,

Perfect! 👍 
Thank you for you time and work. 

I'm sharing your excitement. :)

So there's no need to revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88, correct?

donho:
@ArkadiuszMichalski  @Yaron10 
Thank you for the confirmation!

> So there's no need to revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88, correct?

Yes it's not necessary to revert this PR.
OTOH, I remark there's still the folding performance issue (on the 2nd search) when "color" is searched in all file from root of Notepad++ project. 
I may try to improve the method `ScintillaEditView::collapse()` in another PR.

Yaron10:
> Indeed, it's not necessary to revert this PR.

That's great.

> OTOH, I remark there's still the folding performance issue (on the 2nd search) when "color" is searched in all file from root of Notepad++ project.

There's a performance issue in the _first_ search too. :)
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11774.

> I may try to improve the method ScintillaEditView::collapse() in another PR.

👍 
Thank you and best of luck.

PR #11896, opened 2022-07-10
make code simplify.

dev-aniketj:
### Code more shortest to remove the else part in return.


dev-aniketj:
> Personally I like the previous one better.

Yes, I agree but, I belive to make code shorter, I just removed unnecessary {} backets from the else **return** part

PR #11904, opened 2022-07-12
Add some shortcuts in Styler Configurator to preferences dialog

### LINKED ISSUE #11890: "EOL custom color" does not appear to be working in v8.4.3
### Description of the Issue
`EOL custom color` is a new feature in v8.4.3.  But it doesn't seem to work. :(

1. Install v8.4.3
2. Ensure your style includes a custom colour for `EOL custom color`
3. Turn on the display of EOL markers
4. Notice no custom colour for EOL markers

### Expected Behavior
EOL markers to match the custom colour defined in the current style

### Actual Behavior
No custom colour for EOL markers

### Relevant Debug Information
Notepad++ v8.4.3   (32-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Plugins : 
    CustomizeToolbar (5.3)
    DSpellCheck (1.4.24)
    HexEditor (0.9.12)
    JSONViewer (1.40)
    NppQrCode32 (0.0.0.1)
    nppRegEx ()
    Remove Duplicate Lines (1.3)
    TopMost (1.4)

donho:
Shortcuts to Preferences dialog are added only for the following styles (Global Styles):

"Current line background colour"
"Caret colour"
"Edge colour"
"Line number margin"
"Bookmark margin"
"Fold"
"Fold active"
"Fold margin"
"Smart Highlighting"
"Tags match highlighting"
"Tags attribute"
"Mark Style 1"
"Mark Style 2"
"Mark Style 3"
"Mark Style 4"
"Mark Style 5"
"URL hovered"
"EOL custom color"

Fix #11890


ArkadiuszMichalski:
The label "Go to Settings" could be translated.

alankilborn:
I see that this adds a "Go to settings" link at various places in the "Style Configurator".  "Settings" is the top-level name (see the main menu) for both "Preferences" AND "Style Configurator", so the name is a bit self-referential.  Shouldn't it be "Go to Preferences" instead?

alankilborn:
Also, it is _one direction_, FROM "Style Configurator" TO "Preferences".

Maybe what inspired this change was the "EOL (CR LF)" ellipsis button which allows jumping FROM "Preferences" TO "Style Configurator"?  Note that this is the _other direction_.  Well, this implies that there should be two-direction jumping for ALL. :-)

Of course, adding a bunch of ellipsis buttons in the "Preferences" area would maybe not be great (would possibly add too much clutter), but what about a right-click popup menu with a "Go to Style Configurator" choice, on relevant controls/groupboxes?  The downside is that this would be less obvious for the user...

donho:
@alankilborn 
> I see that this adds a "Go to settings" link at various places in the "Style Configurator". "Settings" is the top-level name (see the main menu) for both "Preferences" AND "Style Configurator", so the name is a bit self-referential. Shouldn't it be "Go to Preferences" instead?

"Go to settings" means "Go to the settings of Preferences dialog", but for the moment all go to the Preferences dialog. However, it's possible in the future, one of the link go to another dialog. So ambiguity id the better solution IMO. In anyway users will understand once they click on link. 

> Maybe what inspired this change was the "EOL (CR LF)" ellipsis button which allows jumping FROM "Preferences" TO "Style Configurator"? Note that this is the other direction. Well, this implies that there should be two-direction jumping for ALL.

That's what I thought. The thing is, there are already too many button/text fields/combobox/checkbox so I don't looking for the exact same logic to make UI even heavier. It's enough (to me) to put where user needs (ie. where is confusing). 

>  what about a right-click popup menu with a "Go to Style Configurator" choice, on relevant controls/groupboxes? The downside is that this would be less obvious for the user...

If it's necessary - so far I don't see the need.

alankilborn:
>So far I don't see the need.

Well, it isn't about a _need_.  It is more about the _niceties_ ... 

And yes, I have noticed an uptick in the amount of niceties added in Notepad++ recently.  Very _nice_; good work.

PR #11911, opened 2022-07-13
Fix selection bg showed on print

donho:
Applying https://sourceforge.net/p/scintilla/code/ci/fc46d7ed1f9393b8f0a577d72016c3c4da9c465c/

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Printer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Printer.cpp
@@ -194,6 +194,7 @@ size_t Printer::doPrint(bool justDoIt)
 	// We print the range of selection
 	if ((!(_pdlg.Flags & PD_RETURNDEFAULT)) && (_pdlg.Flags & PD_SELECTION))
 	{
+		_pSEView->execute(SCI_SETSELECTIONLAYER, SC_LAYER_UNDER_TEXT);
 		if (_startPos > _endPos) 
 		{
 			lengthPrinted = _endPos;
--- a/scintilla/src/EditView.cxx
+++ b/scintilla/src/EditView.cxx
@@ -2694,8 +2694,12 @@ Sci::Position EditView::FormatRange(bool draw, CharacterRangeFull chrg, Rectangl
 	// Don't show the selection when printing
 	vsPrint.elementColours.clear();
 	vsPrint.elementBaseColours.clear();
-	// Transparent:
-	vsPrint.elementBaseColours[Element::SelectionBack] = ColourRGBA(0xc0, 0xc0, 0xc0, 0x0);
+	// Set all selection background colours to be transparent.
+	constexpr ColourRGBA transparent(0xc0, 0xc0, 0xc0, 0x0);
+	vsPrint.elementBaseColours[Element::SelectionBack] = transparent;
+	vsPrint.elementBaseColours[Element::SelectionAdditionalBack] = transparent;
+	vsPrint.elementBaseColours[Element::SelectionSecondaryBack] = transparent;
+	vsPrint.elementBaseColours[Element::SelectionInactiveBack] = transparent;
 	vsPrint.caretLine.alwaysShow = false;
 	// Don't highlight matching braces using indicators
 	vsPrint.braceHighlightIndicatorSet = false;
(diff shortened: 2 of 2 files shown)


rdipardo:
To verify, Scintilla's patch alone still doesn't make the selection transparent:

stp_1
res_1

donho:
As I posted here, it's not fixed by this PR (the fix in Scintilla).
For the unknown reason, I cannot login into SF anymore.
Could anyone post on link below to tell Neil that it doesn't fix the print selection issue?
https://sourceforge.net/p/scintilla/code/ci/fc46d7ed1f9393b8f0a577d72016c3c4da9c465c/

rdipardo:
> Could anyone post on link below to tell Neil that it doesn't fix the print selection issue?

https://sourceforge.net/p/scintilla/bugs/2335/#4c81

donho:
@rdipardo 
> Your patch is missing a line:

By adding `_pSEView->execute(SCI_SETSELECTIONLAYER, SC_LAYER_UNDER_TEXT);`, it works with "Microsoft Print to PDF" (No selection background), but it doesn't work with my real printer (with black font on the black background, so it's not even readable).

rdipardo:
Try this one: https://sourceforge.net/p/scintilla/bugs/_discuss/thread/e4019d46de/4c81/d151/attachment/TransparentBack.patch

donho:
It's the same result: Working with PDF, but not on printing:

image

donho:
> It's the same result: Working with PDF, but not on printing:

Furthermore, there's the regression on the 2nd patch:
While application loses the focus, all selection background colour is removed, like there is no selection.

rdipardo:
> Working with PDF, but not on printing

Can't reproduce. But if I revert the line added to `PowerEditor/src/ScintillaComponent/Printer.cpp`, then *both* copies look like your paper printout.

     git revert --no-edit 79b8684c
   
> While application loses the focus, all selection background colour is removed, like there is no selection

Probably because the default selection layer (`0`) is not restored after the print job is done. My hack for `Printer.cpp` was just a quick PoC.

rdipardo:
About this:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/49f6ee389c951df9c2964c8ed0bd31127b3c1a01/PowerEditor/src/ScintillaComponent/Printer.cpp#L214

Why are we using a *fait maison* type when Scintilla has `Sci_RangeToFormatFull`?

donho:
> Why are we using a _fait maison_ type when Scintilla has `Sci_RangeToFormatFull`?

Just check the definition of NPP_RangeToFormat and that's why.

I think it's better to wait the next version of Scintilla to evaluate if this bug has been fixed.

rdipardo:
Upstream suggested something like this:

~~~diff
diff --git a/PowerEditor/src/ScintillaComponent/Printer.cpp b/PowerEditor/src/ScintillaComponent/Printer.cpp
index c82c4b3b..24026635 100644
--- a/PowerEditor/src/ScintillaComponent/Printer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Printer.cpp
@@ -194,6 +194,7 @@ size_t Printer::doPrint(bool justDoIt)
 	// We print the range of selection
 	if ((!(_pdlg.Flags & PD_RETURNDEFAULT)) && (_pdlg.Flags & PD_SELECTION))
 	{
+		_pSEView->execute(SCI_HIDESELECTION, TRUE);
 		if (_startPos > _endPos) 
 		{
 			lengthPrinted = _endPos;
@@ -482,6 +483,7 @@ size_t Printer::doPrint(bool justDoIt)
 		_pSEView->showMargin(ScintillaEditView::_SC_MARGE_LINENUMBER, isShown);
 
 	_pSEView->execute(SCI_FORMATRANGEFULL, FALSE, 0);
+	_pSEView->execute(SCI_HIDESELECTION, FALSE);
 	::EndDoc(_pdlg.hDC);
 	::DeleteDC(_pdlg.hDC);
 

~~~

Works even when applied directly to dcf8b790, on paper or as PDF.

There won't be another Scintilla release in time for N++ 8.4.4(?), and it's not a Scintilla bug anyway: printouts already look picture perfect in SciTE ([5.2.4](https://www.scintilla.org/SciTEDownload.html)) and Notepad2 ([latest build](https://github.com/zufuliu/notepad2/actions/runs/2674093684#artifacts)).

Issue #11582, opened 2022-04-26
Problem to append extension when saving with filename in Korean


daejungkim:
### Description of the Issue

If the filename is Korean (maybe multi-byte character set) when saving file, 
the added file extension is incorrect.
The last character of filename is added to the following of the 'txt' extension.

### Steps to Reproduce the Issue

1. Write something starting with Ctrl + N
2. Save file with Ctrl + S
3. Input filename with '한글' in the blue highlighted namefield while checked on "Append extension". 
4. Type '한글' by keyboard!  
5. The box cursor will be blinking in the last character like '글'.
>> Doing copy and paste '한글', it does not repoduce.
>> If you move the box cursor to the left or to the right, it does not reproduce.
6. Press 'Enter' key! 
>> By clicking 'Save' button, it does not reproduce.
7. Then filename is '한글.txt글'

### Expected Behavior

In a saving dialog, the input filename is written by '한글' then the output filename is to be '한글.txt'. 

### Debug Information

Notepad++ v8.3.3   (64-bit)
Build time : Mar 13 2022 - 17:20:02
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.1500
Current ANSI codepage : 949
Plugins : DoxyIt.dll MarkdownViewerPlusPlus.dll mimeTools.dll NppConverter.dll NppExport.dll NppMarkdownPanel.dll Python Indent.dll

80ROkWOC4j:
Same in 8.3.2

**Debug Information**

Notepad++ v8.3.2   (64-bit)
Build time : Feb 27 2022 - 03:35:46
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 2009
OS Build : 22000.613
Current ANSI codepage : 949
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

80ROkWOC4j:
```C++
// Modifies the file name if necesary after user confirmed input.
// Called after the user input but before OnFileOk() and before any name validation. 
void onPreFileOk()
```
onPreFileOk is called before Korean IME ends its input. `KbdProcHook` might need to be changed.

[...]

PR #11912, opened 2022-07-13
Fix append extension issue when use Korean IME(letter composition bug).

80ROkWOC4j:
Fix Korean IME issue #11582 

I test this on Korean Windows and it works fine, but Japanese/Chinese Windows IME might need to be tested too.

I've been trying to find when the IME ends input, but i can't find the way. Even "WM_IME_ENDCOMPOSITION" is detected before the composition is complete. So I choose dirty way. Please consider change hooking keyboard to something else if this PR makes another problems.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -251,6 +251,10 @@ public:
 	IFACEMETHODIMP OnFileOk(IFileDialog* dlg) override
 	{
 		_lastUsedFolder = getDialogFolder(dlg);
+
+		if (isCJKMode)
+			return S_FALSE;
+
 		return S_OK;
 	}
 	IFACEMETHODIMP OnFolderChange(IFileDialog*) override
@@ -486,6 +490,7 @@ private:
 		generic_string fileName = getDialogFileName(_dialog);
 		expandEnv(fileName);
 		bool nameChanged = transformPath(fileName);
+
 		// Update the controls.
 		if (!::PathIsDirectory(getAbsPath(fileName).c_str()))
 		{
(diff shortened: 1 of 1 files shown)


yasmise:
@80ROkWOC4j 
I tried AppVeyor's build with this PR, and this has a bad side effect for Japanese IME. (With Japanese IME, issue #11582 does not happen).

Original release of Ver. 8.4.3 works like this:
1. Save new file with Ctrl + S.
2. Input `aiueo` with Japanese IME, then IME displays `あいうえお`.
3. Hit Enter key, then field of filename is changed to `あいうえお.txt`.
4. Hit Enter key to save file.

With this PR:
1. Same as above.
2. Same as above.
3. Hit Enter key, then save-dialog is closed and the file is saved. Filename is correctly set to `あいうえお.txt`.

Debug info:
Notepad++ v8.4.3   (64-bit)
Build time : Jul  7 2022 - 10:55:58
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 21H2
OS Build : 22000.795
Current ANSI codepage : 932
Plugins : 
    mimeTools (2.5)
    NppConverter (4.2.1)
    NppExport (0.2.9)

80ROkWOC4j:
Thanks for @yasmise 's test. Contrary to my expectation, this seems to occur only in Korean IME, not CJK, due to the difference in candidate (composition) behavior.
I've also tried rolling back hooking keyboard[#](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b5a5baf13bb3f0b041f49698afb74af2b7059f6a), but this doesn't seem to work because I ran into another problem.
In order to fix the [#issue](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11582) while maintaining the hooking keyobard and Add Extension feature, it seems to be necessary to create a logic that works only for Korean IME, not the CJK.

PR #11937, opened 2022-07-20
Add reduce mode in Find/Replace dialog

### LINKED ISSUE #11780: [Feature Request] Make Search/Replace Dialog Collapsible
### Description of the Enhancement Request

Enhancement request for usability: Make the search & replace dialog collapsible, and show only the minimum content needed when collapsed. The dialog can already be made semi-transparent, which helps a bit. However, the collapsed dialog will use very little space, and offers the functionality 90% of the time.

### Expected Behavior

1. The default is expanded, showing a collapse icon on the lower right:

notepadpp-search-1

Same for Replace and other tabs:

notepadpp-replace-1

2. Once you click on the collapse icon you get the collapsed view that shows the expand icon in the lower right:

notepadpp-search-2

notepadpp-replace-2

3. Remember the state between application runs.

Thank you for Notepad++, it's a great editor! Thank you for the consideration for this suggestion.

donho:
Add a button on the bottom-right corner of Find/Replace dialog to switch between collapsed & complete mode.

Fix #11780


dfs-:
Hi @donho,
the position of the button in the "closed state" looks a bit off:
image vs. image

Yaron10:
@donho,

The "Show More/Less" feature is very good and useful.
Thank you for your work.

I think it would be much better if you implement the following two points:

1. Reduce the _width_ of the dialog. There are some unused "dead" areas and it's _way too large_ IMO.
2. In the "Reduced" mode, hide only "Search Mode" and "Transparency". 
_If we take "Find in Files" as an example, the "Folder" field is essential._ 

_Reduced:_
תמונה

_Full:_
תמונה
 
Double-clicking the title-bar toggles "More/Less". It might not be obvious to users who do not read the manual. :)

***

If you keep the current implementation.

Hovering over the status bar in "Mark",  you see:
תמונה

In "Replace", you see the "In Selection" GroupBox:
תמונה

alankilborn:
When @Yaron10 said:

>In the "Reduced" mode, hide... "Transparency".

It made me wonder why _Transparency_ is even taking up space on this window, when _Searching_ is available in the _Preferences_ and would be a better home for _Transparency_.  I tend to think than almost every user regards this as a set-it-once-and-forget-it feature?

alankilborn:
Keyboardist concerns:

* is there to be a way to press the reduce/expand button directly, i.e., without using the mouse and without using the Tab key repeatedly to get it so the reduce/expand button has input focus?

* in "reduced" mode, input focus should not go to controls that are not shown when the Tab key is cycled.  BUT... it would be nice if the Alt key accelerators would still function for offscreen controls (e.g. Alt+a would do a _Replace All_ even if that button is not visible)

alankilborn:
The original requester for this feature asked that it "Remember the state between application runs."  With the commit thus far, this does NOT happen (after a restart of N++ the _Find_ window is in the expanded state, even if it was last-used in the reduced state).

rddim:
No arrow on Windows 7 :(

Image-2022-07-21-19-45-38

donho:
@dfs- 
> the position of the button in the "closed state" looks a bit off:

Fixed in the latest commit.

> the windows height get's collapsed a little too much, the input fields are not centered vertically.

That's the best balance I found between the needed feature and the current already complex implementation.
Its position won't be changed otherwise the source code won't be maintainable anymore.

> Also: Collapsed find in files is funny, because you can't select which files to search and where.

User can always uncollapse the dialog.

@Yaron10 
> Reduce the width of the dialog. There are some unused "dead" areas and it's way too large IMO

The Find/Replace dialog contains all the controls and hide/show them according the current tab (Find/Replace/FinF/FiP/Mark).
So the current minimum width has been optimized and cannot reduce it more.

> In the "Reduced" mode, hide only "Search Mode" and "Transparency".
If we take "Find in Files" as an example, the "Folder" field is essential.

I see your need, but it could not be the need of every one. Hence this feature reduces to the its maximum and shows only the essential operation controls. If user need to set more, just click on the button.

> Hovering over the status bar in "Mark", you see:
In "Replace", you see the "In Selection" GroupBox:

Fixed.

@alankilborn 
> The original requester for this feature asked that it "Remember the state between application runs." With the commit thus far, this does NOT happen (after a restart of N++ the Find window is in the expanded state, even if it was last-used in the reduced state).

Just implemented and committed.

> is there to be a way to press the reduce/expand button directly, i.e., without using the mouse and without using the Tab key repeatedly to get it so the reduce/expand button has input focus?

Using TAB is the only way so far - I have no other solution so far.

> in "reduced" mode, input focus should not go to controls that are not shown when the Tab key is cycled. BUT... it would be nice if the Alt key accelerators would still function for offscreen controls (e.g. Alt+a would do a Replace All even if that button is not visible)

In a perfect world, yes. Unfortunately we are in the real world that such request features cannot be done.  

@rddim 
> No arrow on Windows 7 :(

Damn. I will see what I can do about it.

Yaron10:
@alankilborn,

> Reduce the width of the dialog
>> But the dialog is resizable already in this dimension

You can't _reduce_ the width. 

>  and its smallest setting isn't really too wide

That's debatable. :) 

> Double-clicking the title-bar toggles "More/Less"
>> I don't see this behavior. Double-clicking the title bar does nothing for me. Or maybe this is a suggestion?

I certainly was not clear. It was a suggestion.

@donho,

> Hovering over the status bar in "Mark", you see:
> In "Replace", you see the "In Selection" GroupBox:
>> Fixed.

👍 
Thank you.

donho:
@rddim
> No arrow on Windows 7 :(

Could you try the latest commit please?

rddim:
@donho  
> Could you try the latest commit please?

Perfect, thank you very much

Image-2022-07-22-23-57-55

PR #11946, opened 2022-07-25
Fix erasing part of content make hanging issue

### LINKED ISSUE #10193: Crash while deleting text

1. Open a html file (more than 10MBytes).
2. View -> checked "Word wrap"
3. After selecting some text, press DELETE
4. Program crashed.

### Debug Information
Notepad++ v8.1.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro for Workstations (64-bit) 
OS Version : 2009
OS Build : 19042.804
Current ANSI codepage : 936

donho:
Deletion of one line where there's folding didn't show the hiden lines in the old version of Scitilla.
It's not anymore the case in the new (current used) version of Scintilla (which manages such situation correctly).
Therefore the redundant notification for the current used version of Scintilla can be removed, and it solves hanging problem.

Fix #10193

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -64,15 +64,7 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				_pEditView->getCurrentBuffer()->setModifiedStatus(true);
 			}
 
-			if (notification->modificationType & SC_MOD_CHANGEFOLD)
-			{
-				if (prevWasEdit)
-				{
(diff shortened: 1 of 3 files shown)


Yaron10:
@donho,

The crash is fixed. 👍 
Thank you for your work.

I rarely use the folding feature. I'll let others test that.

ArkadiuszMichalski:
@donho 
Works fine, I can't reproduce the crash.

Issue #11941, opened 2022-07-24
NPP with Wine, menu selection crash


jim-rosenvelt:
### Description of the Isue
I run NPP version 8.4.4 with Wine version 7.0 on Linux Debian 10.5. 
When I select Settings -> Style Configurator, NPP crashes right away. 
This happens with NPP 8.4.3 as well, but NPP 8.4.2 works fine.

### Steps to Reproduce the Issue

1. Install NPP on Linux with Wine
2. select Settings -> Style Configurator

### Expected Behavior
selecting a style configuration

### Actual Behavior
crash, see attached NppDump.dmp file

### Debug Information
Notepad++ v8.4.4   (64-bit)
Build time : Jul 15 2022 - 17:54:42
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Microsoft Windows 7 (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1252
WINE : 7.0
Plugins : none
[NppDump.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9175524/NppDump.zip)

pribeirojtm:
Same here. Problem started with last upgrade to 8.4.4 . Also a crash is happening when trying to access Plugin Admin...
![Screenshot from 2022-07-25 09-40-23](https://user-images.githubusercontent.com/18724236/180735404-10f8bfb2-a1f9-4a6a-b049-4cbaade911b3.png)

xomx:
It seems from the above crashdump that it is crashing in the ComCtl32.dll (MS Common Control library) call.

donho:
@xomx  
> It seems from the above crashdump that it is crashing in the ComCtl32.dll (MS Common Control library) call.

From the provided dump by @jim-rosenvelt, I saw the same problem. Regarding the change of Styler Configurator, I've checked the following commits:
79e766755d9d466704c535f99eb3ce99ffbbb1bf , 5c6fbd33103f3cde666c48a6a6ce0bcc8a90ff0e and 67ab4d55276ae99e632f19d1035de70cc022e3b9
But non of them seems causing the crash.

According the description of @pribeirojtm , now there's a crash on another dialog, I'm wondering it's due to d1347d501451e10c5d988282ca681760d2c2abff
 
@pribeirojtm @jim-rosenvelt 
Could you try other dialogs please?

[...]

PR #11961, opened 2022-07-31
Fix crash on Wine,

ozone10:
create functions for every classes in `EnumChildWindows` lambda,
and reorder class detection to improve stability,
add checks to improve stability on older OS and wine

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11941

I am currently away so I cannot test it properly, but it should fix crash with wine.
Issue seems to be `subclassCustomBorderForListBoxAndEditControls` but I am not sure why.

@xomx 
if you have time could you find out why?

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -350,6 +350,7 @@ namespace NppDarkMode
 	}
 
 	static bool g_isAtLeastWindows10 = false;
+	static bool g_isWine = false;
 
 	void initDarkMode()
 	{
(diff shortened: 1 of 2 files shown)


xomx:
@ozone10 

Just tested this PR and it is ok, no more crashing under Wine, good work.

I will try to find out the real reason, but this could also be some minor flaw in the Wine interpretation of MS COMCTL32.

zzhiyi:
Hi, Wine developer here. This bug should be fixed by https://source.winehq.org/git/wine.git/commitdiff/9b9fef2e112ff2ce0e1ce2da3aaacd2ffefa6d87

xomx:
@ozone10 

I reported the problem to WineHQ Bugzilla and they seem to have fixed their code pretty fast. The next Wine version (7.16 I think) will contain the fix,, so you will be able to adjust your `if (!g_isWine)` check for the newer Wine versions if you want to.

Issue #11963, opened 2022-08-01
[Feature Request] Use MsgWaitForMultipleObjectsEx in message loop in order to support overlapped IO.


Kered13:
Perhaps this is a stretch, but I am writing a plugin (LSP implementation) where I would like to be able to use overlapped IO (specifically, ReadFileEx and WriteFileEx). In order to use this Windows requires an event loop that puts the thread into an alertable state (see [Alertable I/O](https://docs.microsoft.com/en-us/windows/win32/fileio/alertable-i-o)). [MsgWaitForMultipleObjectsEx](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-msgwaitformultipleobjectsex) would work well for this, because it will also wakes upon receiving messages, allowing the main loop to handle both messages and overlapped IO.

I have prepared a pull request for this functionality if you're interested.

As an addendum, I believe this would also mean that CReadDirectoryChanges would no longer require a separate thread ([see here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/af56713079c9e31b8c81b5603a6c841afb3e2005/PowerEditor/src/WinControls/ReadDirectoryChanges/ReadDirectoryChanges.cpp#L52)), it also uses alertable IO, and therefore uses a thread calling SleepEx to handle the events.

PR #11964, opened 2022-08-01
Use MsgWaitForMultipleObjectsEx in message loop in order to support overlapped IO.

Kered13:
Implementation for feature request #11963.

I have tested basic functionality with this and everything seems to be working correctly. I have also tested that this allows plugins to use overlapped IO.

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -730,16 +730,23 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, PWSTR pCmdLine, int)
 		bool going = true;
 		while (going)
 		{
-			going = ::GetMessageW(&msg, NULL, 0, 0) != 0;
-			if (going)
+			DWORD wait_object = ::MsgWaitForMultipleObjectsEx(0, nullptr, INFINITE, QS_ALLINPUT, MWMO_ALERTABLE);
+			going = wait_object != WAIT_FAILED;
+			if(wait_object == WAIT_OBJECT_0)
 			{
-				// if the message doesn't belong to the notepad_plus_plus's dialog
-				if (!notepad_plus_plus.isDlgsMsg(&msg))
-				{
-					if (::TranslateAccelerator(notepad_plus_plus.getHSelf(), notepad_plus_plus.getAccTable(), &msg) == 0)
+				while (going && ::PeekMessageW(&msg, NULL, 0, 0, PM_REMOVE)) {
+					going = msg.message != WM_QUIT;
+					if (going)
 					{
-						::TranslateMessage(&msg);
-						::DispatchMessageW(&msg);
+						// if the message doesn't belong to the notepad_plus_plus's dialog
+						if (!notepad_plus_plus.isDlgsMsg(&msg))
+						{
+							if (::TranslateAccelerator(notepad_plus_plus.getHSelf(), notepad_plus_plus.getAccTable(), &msg) == 0)
+							{
+								::TranslateMessage(&msg);
+								::DispatchMessageW(&msg);
+							}
+						}
 					}
 				}
 			}


donho:
@Kered13 
I've tested the PR and find no anomaly so far. 
However, one more loop is added and the value of `pHandles` that you passed in `::MsgWaitForMultipleObjectsEx(0, nullptr, INFINITE, QS_ALLINPUT, MWMO_ALERTABLE);` is **nullptr**.
Could you provide more explanation for your modification and its advantage to your plugin code.  

> As an addendum, I believe this would also mean that CReadDirectoryChanges would no longer require a separate thread ([see here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/af56713079c9e31b8c81b5603a6c841afb3e2005/PowerEditor/src/WinControls/ReadDirectoryChanges/ReadDirectoryChanges.cpp#L52)), it also uses alertable IO, and therefore uses a thread calling SleepEx to handle the events.

Also could you provide a concrete example (or event a commit) for the modification you mentioned above.

Kered13:
> one more loop is added

The outer loop is the main event loop. It blocks in `MsgWaitForMultipleObjectsEx` until either a message has arrived, an [overlapped (asynchronous) IO](https://docs.microsoft.com/en-us/windows/win32/sync/synchronization-and-overlapped-input-and-output) operation has completed, or an [asynchronous procedure call](https://docs.microsoft.com/en-us/windows/win32/sync/asynchronous-procedure-calls) is queued. This function does not handle messages itself, it just waits for messages to enter the queue. It does handle overlapped IO operations though, by calling the corresponding callback (see `ReadFileEx` for an example).

The inner loop handles messages by looping over `PeekMessage`. A loop is used because multiple messages could be in the queue and we want to empty the queue before waiting again. `PeekMessage` is used because we don't want to block when the queue is empty.

> and the value of pHandles that you passed in is **nullptr**

In addition to messages and overlapped IO operations, `MsgWaitForMultipleObjectsEx` can also wait on certain objects such as [semaphores](https://docs.microsoft.com/en-us/windows/win32/sync/using-semaphore-objects), [events](https://docs.microsoft.com/en-us/windows/win32/sync/event-objects), and [threads](https://docs.microsoft.com/en-us/windows/win32/procthread/creating-threads). However this functionality is not useful to plugins unless there is a way for the plugin to communicate objects to wait on back to Notepad++ (something that could potentially be added later, but beyond the scope of this PR). In any case, my plugin does not need such functionality.

If `pHandles` is `nullptr` and `nCount` is `0` then `MsgWaitForMultipleObjectsEx` simply does not wait on any objects. But it will still wait on messages and asynchronous IO.

> Could you provide more explanation for your modification and its advantage to your plugin code.

My plugin uses pipes to communicate with an LSP server. This involves blocking calls to `ReadFileEx` and `WriteFileEx`. I do this in separate IO threads right now, but my code would be significantly simplified (in particular synchronization would be much easier) if I could do everything from a single thread using overlapped (asynchronous) IO.

> Also could you provide a concrete example (or event a commit) for the modification you mentioned above.

I can probably provide an example of what this would look like later today or tomorrow, for demonstrative purposes. It will be a larger change and beyond the scope of this PR though.

Kered13:
So it turns out that I was partially wrong about `ReadDirectoryChanges`. It is never directly called from the main thread, so this change to the main event loop has no effect. Instead, the main loop starts up separate threads in two locations: `fileBrowser.cpp` and `NppIO.cpp`, these threads then start `ReadDirectoryChanges` which creates *another* thread to handle the monitoring. I was able to eliminate the inner thread by rewriting `ReadDirectoryChanges` without too much difficulty. This still serves as a demonstration of how overlapped IO works. You can see what that commit looks like [here](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4439084b38f6d2b6a86ac7367ff62b2d582b1dd1). Note that it is not based on this PR, since that ended up being unnecessary.

I still believe it is *possible* to completely eliminate the extra threads in `fileBrowser.cpp` and `NppIO.cpp`, but that would be a much more complex change (they each run a custom message loop that would have to be integrated into the main message loop) and I'm not volunteering to do it.

donho:
@Kered13 
OK, I see your point. Thank you for your demo and coding effort.

Is your LSP plugin on GitHub or an open source? If it is, I can at least to see the diff (or at least to see the effect with your PR and without). Even the amount of change is not large, and it seems no regression with (very) quick test, I cannot just change this core part without any solid argument behind.

Note that if this PR is accepted with a solid argument from your plugin, it could be reverted in the future - with any regression found which is due to this PR.

Kered13:
My plugin is on SourceForge, the code that handles communication with the server is [here](https://sourceforge.net/p/kered13-notepad-plugins/code/ci/default/tree/src/LspClient/). The rest of the plugin code is [here](https://sourceforge.net/p/kered13-notepad-plugins/code/ci/default/tree/src/NppLsp/) and [here](https://sourceforge.net/p/kered13-notepad-plugins/code/ci/default/tree/src/NppLspPlugin/). It is not written to use overlapped IO right now though, as I didn't want to do that rewrite without knowing if this PR would be accepted.

My full plan, if this PR is accepted, is actually to rewrite the communication logic using C++20 coroutines using a little coroutines library that I have cooked up and tested which is based on overlapped IO. This would allow me to write code like `response = co_await client->request(json)` and the code would suspend execution (returning to the NPP main event loop) and then resume when the LSP server sent a response. Since everything would run on the main thread it would allow me to remove complex logic to deal with multithreading, such as in [this file](https://sourceforge.net/p/kered13-notepad-plugins/code/ci/default/tree/src/NppLsp/Handlers/Completion.cpp) (which also shows how I have used callbacks so far to handle asynchronous requests).

donho:
@Kered13 
Since this PR doesn't bring any bug-fix or feature, and it modified the application core part which could bring the regression, I cannot merge it into master for the future release.
However, you can always rewrite the communication logic of your plugin and test with the binary. If the new communication logic works with this PR, then we can talk about merging it (the policy of Notepad++ is favorite the plugins' dev). 
OTOH, I cannot guarantee that the merged code won't be reverted if any regression find.
So it's your call to rewrite a part of your plugin, without this PR merged.
I hope it sounds fair to you.

donho:
Outdated.

PR #11984, opened 2022-08-05
Improve English phrase "Go to Preferences" for item id 2234

Edditoria:
Capital "P" to keep consistent to item id 1766.


donho:
In the current cases it's going to Preference, however in the future it's not necessary to go to Preference dialog - it could be in Find dialog or even in Styler Configurator itself. Hence the word "Settings" was chosen.

Issue #11988, opened 2022-08-05
Only autofill search text in find and replace dialog if text is **_highlighted_** when opened


vertigo220:
### Description of the Issue

As the title says. I personally find it very useful to have the "Find what" box auto-filled with _selected_ text, but it's also very counter-productive when it auto-fills with the word the cursor is on, when that word isn't selected. I have to place the cursor on a blank line every time I open the search box if I don't want this to happen.

This has been brought up at least a couple times before, but this specific issue has been seemingly ignored/overlooked. In #2373, it's stated that it works this way, but that is not my experience, up to and including the latest version, so maybe something has changed since then. In #3243, the option "Searching > Don't fill find field in Find dialog with selected word" was added, but as mentioned by a couple users after the change, it only provides an all-or-nothing option. Those users, like myself, want the feature but only want it to work when a word is actually highlighted, which is how it's supposed to work, _not_ when nothing is highlighted.

### Steps to Reproduce the Issue

1. Open Find dialog and enter some text in the "Find what" box
2. Close the dialog
3. Place the cursor on a word but _don't_ highlight it
4. Open Find dialog and see that the word where the cursor is placed was auto-filled, replacing the text you previously had there

### Expected Behavior

With no selection, no autofill should occur

### Actual Behavior

With no selection, autofill does occur

### Debug Information

Notepad++ v8.4.4   (64-bit)
Build time : Jul 15 2022 - 17:54:42
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Program Files\Notepad++\change.log" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro N (64-bit) 
OS Version : 21H2
OS Build : 19044.1826
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NavigateTo (1.12.4.1)
    NppConverter (4.4)
    NppExport (0.4)
    NppMenuSearch (0.9.3)
    NppSaveAsAdmin (1.0.211)
    NppSnippets (1.6)
    RegexTrainer (1)
    RunMe (1.4)
    SurroundSelection (1.4.1)
    _CustomizeToolbar (4.2)

tools400:
Great to have that new option.

PR #11989, opened 2022-08-06
Turn off selecting text for find dialog

doug1234:
Added the ability to turn off selecting text when displaying the Field dialog for issue #11988.  Now the find dialog text will only be updated if there is a selection.


alankilborn:
Don't you also have to provide the base localization in english.xml?

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -323,11 +323,12 @@ STYLE DS_SETFONT | DS_FIXEDSYS | DS_CONTROL | WS_CHILD
...
 BEGIN
     CONTROL         "Don't fill find field in Find dialog with selected word", IDC_CHECK_STOPFILLINGFINDFIELD, "Button", BS_AUTOCHECKBOX | WS_TABSTOP,37,10,350,10
-    CONTROL         "Use Monospaced font in Find dialog (Need to restart Notepad++)",IDC_CHECK_MONOSPACEDFONT_FINDDLG, "Button", BS_AUTOCHECKBOX | WS_TABSTOP,37,25,350,10
-    CONTROL         "Find dialog remains open after search that outputs to results window",IDC_CHECK_FINDDLG_ALWAYS_VISIBLE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP,37,40,350,10
-    CONTROL         "Confirm Replace All in All Opened Documents",IDC_CHECK_CONFIRMREPLOPENDOCS, "Button", BS_AUTOCHECKBOX | WS_TABSTOP,37,55,350,10
-    CONTROL         "Replace: Don't move to the following occurrence", IDC_CHECK_REPLACEANDSTOP, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 70, 350, 10
-    CONTROL         "Search Result window: show only one entry per found line", IDC_CHECK_SHOWONCEPERFOUNDLINE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 85, 350, 10
+    CONTROL         "Don't select closest word when Find dialog is displayed", IDC_CHECK_STOPSELECTINGNEARESTWORDONFIND, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 25, 350, 10
```
alankilborn, line 326: Suggest reversing logic (all these "don't"s make my head hurt when trying to think about what it might do), and change wording to "Auto-fill Find-What with word at caret when no selection".

Of course, such a suggestion practically begs for the existing _Don't fill find field in Find dialog with selected word_ to be changed as well.

I'd suggest:

(checkbox) **Auto-fill Find-What when searching is invoked**
and then directly underneath with a slight right indentation:
(checkbox) **Use word at caret when no selection**

Both checkboxes would be ticked by default, preserving how N++ has operated for years.

Call me a dreamer.  :-)

Yaron10:
@doug1234,

That's a useful addition. Thank you for your work.

@alankilborn,

> Suggest reversing logic (all these "don't"s make my head hurt when trying to think about what it might do)...

Good point. Thank you.

PowerEditor/src/WinControls/Preference/preference.rc
doug1234: Thank you for the good suggestions @alankilborn .  I wasn't 100% happy with the wording either.  The only problem with reversing the logic of the current checkbox is how to deal with saved preferences from older versions of notepad++?  I could look for the old preference name and use the reversed version of the value if it is there but that is kind of messy.  What do you think?

alankilborn: Well, if my suggestion is REALLY liked (and even liked by @donho, a dubious prospect :-) ), then I'd say don't try to reuse what is already there, and let the existing setting "die off" and create two new ones.  This would have the advantage of clarity going forward, but the disadvantage of forcing users that have already made a non-default choice for the existing setting to come back and make a new choice -- but this could actually be an advantage because then they would be aware of the improved possible choices!

donho: @alankilborn 
> Call me a dreamer. :-)

What do you think the following GUI:
image

Do you guys think it needs to be reworded in order to make it more clear?

(RC code)
```
IDD_PREFERENCE_SUB_SEARCHING DIALOGEX 0, 0, 455, 185
BEGIN
    GROUPBOX        "When launch Find dialog",IDD_FILLFINDFIEL_GRPSTATIC,31,4,323,43,BS_CENTER
    CONTROL         "Select the word on which caret is",IDC_CHECK_FILLFINDFIEL_WITHSELECTED,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,52,31,275,10
    CONTROL         "Fill find field with selected word",IDC_CHECK_FILLFINDFIEL_SELECT_CARET,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,16,275,10
    CONTROL         "Use Monospaced font in Find dialog (Need to restart Notepad++)",IDC_CHECK_MONOSPACEDFONT_FINDDLG,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,52,350,10
    CONTROL         "Find dialog remains open after search that outputs to results window",IDC_CHECK_FINDDLG_ALWAYS_VISIBLE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,67,350,10
    CONTROL         "Confirm Replace All in All Opened Documents",IDC_CHECK_CONFIRMREPLOPENDOCS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,82,350,10
    CONTROL         "Replace: Don't move to the following occurrence",IDC_CHECK_REPLACEANDSTOP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,97,350,10
    CONTROL         "Search Result window: show only one entry per found line",IDC_CHECK_SHOWONCEPERFOUNDLINE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,112,350,10
END
```

alankilborn: * "When launch Find dialog" may not be ideal of meaning when the Find dialog is already open but the editor window has input focus (there's no "launch" because the window is already there).  More descriptive would be "When Find is initiated via menu or shortcut" but that is more words.
* "Fill find field with selected word" might be better as: "Fill find field with selected text" (because it could be more than a 'word' if user has more than that selected)
* "Select the word on which caret is" : I really think this needs to convey that this action only occurs where there is no pre-selected text, maybe "Select word under caret when nothing selected"

Sorry, Don, I'm not purposefully trying to change EVERY new line you've suggested for the UI.  :-)

I'm just trying to make the best UI possible. :-)

Perhaps solicit opinions on this from @pryrt @Yaron10 @vinsworldcom ... as a group we can make it the best!

vinsworldcom: @donho - I like it!  I do also like @alankilborn suggestions of:

- "Fill find field with selected text"
- "Select word under caret when nothing selected"

Instead of "When launch Find dialog" or the more verbose but explicit, can we just use "Find Dialog Options"?

Cheers.

donho: Thank you for your suggestions.

Indeed, "When Find is initiated via menu or shortcut" is verbose.
How about "When Find dialog is invoked"? So it's implied both actions.

image

Note that if "Fill find field with selected text" is unchecked, then "Select word under caret when nothing selected" will also be unchecked (if checked) and disabled.

Do you guys see any scenario could be incoherent with the new GUI behaviour?

alankilborn: I think the latest text is good enough; also the behavior seems consistent.

doug1234: I like the text changes.  My only thought is would it better to use "caret" or "cursor"?  I can make the changes when everyone is satisfied with the wording unless someone else is going to update the pull request.

vinsworldcom: "caret" is the Scintilla term.

donho: @doug1234 
Let's use "cursor" instead of "caret".

Here are the codes you need copy-past:
```
#define IDD_PREFERENCE_SUB_SEARCHING 6900 //(IDD_PREFERENCE_BOX + 900)
	//#define IDC_CHECK_STOPFILLINGFINDFIELD (IDD_PREFERENCE_SUB_SEARCHING + 1)
	#define IDC_CHECK_MONOSPACEDFONT_FINDDLG (IDD_PREFERENCE_SUB_SEARCHING + 2)
	#define IDC_CHECK_FINDDLG_ALWAYS_VISIBLE (IDD_PREFERENCE_SUB_SEARCHING + 3)
	#define IDC_CHECK_CONFIRMREPLOPENDOCS (IDD_PREFERENCE_SUB_SEARCHING + 4)
	#define IDC_CHECK_REPLACEANDSTOP (IDD_PREFERENCE_SUB_SEARCHING + 5)
	#define IDC_CHECK_SHOWONCEPERFOUNDLINE (IDD_PREFERENCE_SUB_SEARCHING + 6)
	#define IDD_FILLFINDFIELD_GRPSTATIC (IDD_PREFERENCE_SUB_SEARCHING + 7)
	#define IDC_CHECK_FILLFINDFIEL_WITHSELECTED (IDD_PREFERENCE_SUB_SEARCHING + 8)
	#define IDC_CHECK_FILLFINDFIEL_SELECTCARET (IDD_PREFERENCE_SUB_SEARCHING + 9)
```
and
```
IDD_PREFERENCE_SUB_SEARCHING DIALOGEX 0, 0, 455, 185
BEGIN
    GROUPBOX        "When Find dialog is invoked",IDD_FILLFINDFIELD_GRPSTATIC,31,4,323,43,BS_CENTER
    CONTROL         "Fill find field with selected text",IDC_CHECK_FILLFINDFIEL_WITHSELECTED,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,16,275,10
    CONTROL         "Select word under cursor when nothing selected",IDC_CHECK_FILLFINDFIEL_SELECTCARET,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,52,31,275,10
    CONTROL         "Use Monospaced font in Find dialog (Need to restart Notepad++)",IDC_CHECK_MONOSPACEDFONT_FINDDLG,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,52,350,10
    CONTROL         "Find dialog remains open after search that outputs to results window",IDC_CHECK_FINDDLG_ALWAYS_VISIBLE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,67,350,10
    CONTROL         "Confirm Replace All in All Opened Documents",IDC_CHECK_CONFIRMREPLOPENDOCS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,82,350,10
    CONTROL         "Replace: Don't move to the following occurrence",IDC_CHECK_REPLACEANDSTOP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,97,350,10
    CONTROL         "Search Result window: show only one entry per found line",IDC_CHECK_SHOWONCEPERFOUNDLINE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,37,112,350,10
END
```
1. The default settings of this feature should be both "Fill find field with selected text" & "Select word under cursor when nothing selected" are checked.
2. The old settings should be remembered by this new feature: if "Don't fill find field in Find dialog with selected word" is checked in the old config, the new feature should keep both "Fill find field with selected text" & "Select word under cursor when nothing selected" unchecked. 
3. GUI behaviour: if "Fill find field with selected text" is unchecked, then "Select word under caret when nothing selected" will also be unchecked (if checked) and disabled.

Feel free to ask question if you have any.

Thank you for your coding effort.

alankilborn: > Let's use "cursor" instead of "caret".

Oh, no, I'm absolutely HORRIFIED by this.
"Caret" is 100% the correct term (English is my only frame of reference, however).
Let's not introduce an inconsistency now.
Current usages in english.xml:
* `caret` : 3 hits
* `cursor` : 0 hits

alankilborn: User manual:
* `caret` : 39 hits
* `cursor` : 22 hits
If there is agreement on "caret" for this PR, I'll put an issue in on the user manual to evaluate/change all instances of "cursor" to "caret".

alankilborn: image

alankilborn: image

doug1234: I am OK with caret if it is what the majority wants.  I will try to get this wrapped up tomorrow night.  Thanks for the help!

pryrt: Since caret is used by scintilla, and everywhere in english.xml, that would be my vote, as well.

donho: It's true that "caret" is Scintilla term. However, the most users don't really use Scintilla directly.
And here is the definition of caret, which has nothing to do with the symbol that we are talking about: 
https://en.wikipedia.org/wiki/Caret_(computing)

Between "caret" and "cursor", I have no preference. All we should judge is from the aspect of the most of users.

donho: Notepad++ is my only editor of choice ;) 
But you guys may know which term the other editors use for this stuff?

alankilborn: >which term the other editors use for this stuff?

VS2019 editor hasn't settled on one term (or...maybe it has and really is using the terms correctly, per Don's finding of "The cursor is the graphical image that represents the mouse..."):

image

image

doug1234:
Looks like caret is the winner so I went with that.  I am going to spend a bit more time testing and then I will update the pull request.

doug1234:
OK... I think its ready to go with the suggested changes.

PowerEditor/src/WinControls/Preference/preference.rc
rdipardo: > VS2019 editor hasn't settled on one term (or...maybe it has and really is using the terms correctly

Properly speaking, a *caret* is an [insertion point][0], so more appropriate for what a text editor does, by metaphor with  the proof-reading symbol, an sub-scripted `^`.

A *cursor* points to or indexes a position (on the screen or in a file, like the "position indicator" in binary I/0 functions). The typical use case is (read-only) navigation, not editing.

Given their semantic and technical differences, there are perfectly valid reasons for using both terms in the same application.


[0]: https://www.merriam-webster.com/dictionary/caret

alankilborn: >there are perfectly valid reasons for using both terms in the same application

OK, but there aren't currently any uses of "cursor" in the UI, and the usages of "caret" are all reasonable as that term.

The user manual is currently under revision to change any improper usages to the proper one, evaluating each individual "caret" and "cursor".

PowerEditor/src/Parameters.cpp
```diff
@@ -5475,9 +5475,20 @@ void NppParameters::feedGUIParameters(TiXmlNode *node)
 			if (optNameMonoFont)
 				_nppGUI._monospacedFontFindDlg = (lstrcmp(optNameMonoFont, TEXT("yes")) == 0);
 
+			//This is an option from previous versions of notepad++.  It is handled for compatibility with older settings.
 			const TCHAR* optStopFillingFindField = element->Attribute(TEXT("stopFillingFindField"));
-			if (optStopFillingFindField)
-				_nppGUI._stopFillingFindField = (lstrcmp(optStopFillingFindField, TEXT("yes")) == 0);
+			if (optStopFillingFindField) {
```
donho, line 5480: Please respect coding standard:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#general

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -4982,9 +4984,14 @@ intptr_t CALLBACK SearchingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 		{
 			switch (wParam)
 			{
-				case IDC_CHECK_STOPFILLINGFINDFIELD:
+				case IDC_CHECK_FILL_FIND_FIELD_WITH_SELECTED:
 				{
-					nppGUI._stopFillingFindField = isCheckedOrNot(IDC_CHECK_STOPFILLINGFINDFIELD);
+					nppGUI._fillFindFieldWithSelected = isCheckedOrNot(IDC_CHECK_FILL_FIND_FIELD_WITH_SELECTED);
+					::EnableWindow(::GetDlgItem(_hSelf, IDC_CHECK_FILL_FIND_FIELD_SELECT_CARET), nppGUI._fillFindFieldWithSelected ? TRUE :FALSE);
+					if (!nppGUI._fillFindFieldWithSelected) {
```
donho, line 4991: Please respect coding standard:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#general

donho (changes requested):
1. Please consider the suggested changes.
2. It's OK for this time, but please make sure to create a new branch for each PR: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests

PowerEditor/src/WinControls/Preference/preference.rc
donho: OK. Let's keep using "caret".

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1046,12 +1046,15 @@ You can define several column markers by using white space to separate the diffe
                 </Print>
 
                 <Searching title="Searching">
-                    <Item id="6901" name="Don't fill find field in Find dialog with selected word"/>
+                    <!-- <Item id="6901" name="Don't fill find field in Find dialog with selected word"/>-->
```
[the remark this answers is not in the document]
doug1234: Removed.

PowerEditor/src/Parameters.cpp
doug1234: Sorry about that...

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
doug1234: Sorry about that...

doug1234:
Should be good now.

donho (approved):

PR #11992, opened 2022-08-06
Don't override detected file type with default language

### LINKED ISSUE #11504: shebang/xml heuristic not applied if default language is set
### Background

### Problem

### Reproduce

1. Create the file
```
<blah></endblah>
```
2. Save as `test`
3. Close the file

### Other

1. extension of the file
2. default language

### Debug Info
```
Notepad++ v8.3.3   (64-bit)
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1586
Current ANSI codepage : 1252
```

rdipardo:
### Problem

 1. Go to "Settings > Preferences > New Document" and change the "Default language" option, e.g. "C++"

 2. Create a new file with only a document type declaration at the top:

    `<?xml version="1.0" encoding="UTF-8" ?`

 3. Save the file _with no extension_

 4. Close and reopen the file

 5. The detected file type is "C++" (or whatever default language was chosen in step 1.)

 ### Analysis

 The logic of this condition is flawed:

 https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2e66fe000765d7923f731ef8f1e524a27ddc0912/PowerEditor/src/ScintillaComponent/Buffer.cpp#L744-L746

 When a default language has been set, and the file has no extension, a new `Buffer` instance will be constructed with *the default language*:

 https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2e66fe000765d7923f731ef8f1e524a27ddc0912/PowerEditor/src/ScintillaComponent/Buffer.cpp#L178-L181

So the condition `buf->getLangType() == L_TEXT` is *always* false, and the *detected* file type is *never* applied. The default language is intended for __*new, empty*__ files, not __*saved*__ files with existing content, especially if the content matches a pattern that can be inferred by `FileManager::detectLanguageFromTextBegining()`.

 ### Patch

 This PR prevents the default language from overriding the detected one by checking that a file is brand new (i.e. untitled) before deciding to skip the detected file type.

 Fixes #11504

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -742,7 +742,7 @@ BufferID FileManager::loadFile(const TCHAR* filename, Document doc, int encoding
 		buf->setEncoding(-1);
 
 		// if no file extension, and the language has been detected,  we use the detected value
-		if (!newBuf->_isLargeFile && ((buf->getLangType() == L_TEXT) && (loadedFileFormat._language != L_TEXT)))
+		if (!newBuf->_isLargeFile && loadedFileFormat._language != L_TEXT)
 			buf->setLangType(loadedFileFormat._language);
 
 		setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);


rdipardo:
_Cher_ Don,
> Maybe I'm missing somethings, but how can buffer could be untitled in the context of loading file (this modification/PR is on the `loadFile`)?

In fact, the PR'd condition is basically this:  

    (!newBuf->_isLargeFile && (TRUE && (loadedFileFormat._language != L_TEXT)))

which is identical to this:

    (!newBuf->_isLargeFile && loadedFileFormat._language != L_TEXT)
        
I was searching for some explicit method to say: "This file is __*not*__ brand new, nor is it plain text". As you just pointed out, no such thing is really necessary; brand new files will never take this code path.

We can erase the redundant check if you prefer; the effect will be exactly the same.

donho:
> I was searching for some explicit method to say: "This file is **_not_** brand new, nor is it plain text". As you just pointed out, no such thing is really necessary; brand new files will never take this code path.

Hmm... Just check it again, `newBuf->isUntitled()` could be **true** when `loadFile` is called during Notepad++ loading any session (unsaved files or a brand new text can be in session), but in this case, the value of `_lang` will be overridden anyway by the concerning value in the session. Except this case, there won't be any brand new text loaded by `loadFile`.

For making code simpler, let's remove `!newBuf->isUntitled()` condition.

donho (changes requested):
Remove `!newBuf->isUntitled()` condition, which is not necessary.

rdipardo:
Rebased + updated

donho (approved):

PR #11991, opened 2022-08-06
Update greek.xml

Blueddo:

PR #11994, opened 2022-08-06
Update greek.xml

Blueddo:


chcg:
@Blueddo Seems this is a correction to #11991. Correct? Than #11991 could be closed.

Issue #12010, opened 2022-08-13
Suggestion: Search in All Documents -> show how times there is corresponding results in same line


Tequiller:
To start, it has been a great step forward by putting in the Search result box only one time a line 👍 

But a little step more forward may be done by telling how many results there are by line.

- For example (sorry for french words, don't exactly know how it's in english version), today we have:

`Rechercher "b" (21 résultats dans 1 documents depuis 1 documents recherchés)`
`  nouveau 1 (21 résultats)`
`	Ligne 1: ababababababababababa`
`	Ligne 2: ababababa`
`	Ligne 3: aba`
`	Ligne 4: ababababababa`

- The update could be:

`Rechercher "b" (21 résultats dans 1 documents depuis 1 documents recherchés)`
`  nouveau 1 (21 résultats)`
`	Ligne 1 (10 résultats): ababababababababababa`
`	Ligne 2 ( 4 résultats): ababababa`
`	Ligne 3 ( 1 résultat ): aba`
`	Ligne 4 ( 6 résultats): ababababababa`

Thanks :)

alankilborn:
I think the repeated use of the word `results` on each line is a terrible waste of screen space.
Maybe with this waste one of my matches is pushed off the screen on the righthand side.

Truly repeating `Line` constantly is awful as well, when something very short like `L` would suffice, but `Line` is not a new feature.

I really think that something like `Line 2 (4)` would suffice.

Patriccollu:
@alankilborn 
> Truly, repeating Line constantly is awful as well, when something very short like L would suffice, but using Line is a long-established way that N++ does it.

In December 2020, the string `Line` was made translatable by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e3455a0f7e2e8aa2c97d1f7a6d2ddfde2d83f95e:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f2c4ecd7e1bfd5d7d5c14c854438c02f2ba0909a/PowerEditor/installer/nativeLang/english.xml#L1552

So if you really want, you can reduce its size to `L`. But you will have to redo your change after every new release installed.

alankilborn:
>So if you really want, you can reduce its size to L. 

Oh, I already did it, long ago...

>But you will have to redo your change after every new release installed.

Not true; why do you think so?

alankilborn:
[...]

PR #12014, opened 2022-08-14
Don't enable "show only one entry per found line" by default

vlakoff:
Per https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11808#issuecomment-1212701012:

> Although this new mode can be useful, I think the setting shouldn't be enabled by default.
> 
> It makes very easy to overlook that an entry occurs several times on a same line. There is no hint that a line contains several matches, apart from the highlights that the user has to look for. Actually, for long lines that don't fit on screen width, there is no hint at all, unless the user horizontally scrolls the results pane while looking for the highlights...

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -6335,7 +6335,7 @@ void NppParameters::createXmlTreeFromGUIParams()
 		GUIConfigElement->SetAttribute(TEXT("isLessModeOn"), _nppGUI._findWindowLessMode ? TEXT("yes") : TEXT("no"));
 	}
 
-	// <GUIConfig name="FinderConfig" wrappedLines="no" purgeBeforeEverySearch="no" showOnlyOneEntryPerFoundLine="yes"/>
+	// <GUIConfig name="FinderConfig" wrappedLines="no" purgeBeforeEverySearch="no" showOnlyOneEntryPerFoundLine="no"/>
 	{
 		TiXmlElement* GUIConfigElement = (newGUIRoot->InsertEndChild(TiXmlElement(TEXT("GUIConfig"))))->ToElement();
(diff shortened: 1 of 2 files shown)


alankilborn:
The author already decided this should be enabled by default; I don't think that is going to get revisited.  Those that don't like the mode can turn it off.

>There is no hint that a line contains several matches, apart from the highlights that the user has to look for. Actually, for long lines that don't fit on screen width, there is no hint at all

That is a separate thing.  Perhaps you should open an issue about that.

vlakoff:
It's not about liking the mode or not. The issue is that this mode can lead the user to trouble (miss results) if he is not aware of it.

alankilborn:
> this mode can lead the user to trouble (miss results) if he is not aware of it

This is very true.
But again, the way to address is is not by changing the default setting, necessarily, but by addressing the fact that "no hint that a line contains several matches".

>Perhaps you should open an issue about that.

This issue does it:  #12010

vlakoff:
Related: #12023

donho:
@vlakoff 
The PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14520 solves this problem.

vlakoff:
#12014 fixes entries _not displayed at all_, because they are located after column SC_SEARCHRESULT_LINEBUFFERMAXLENGTH, which is approximately 1024 characters.

Although it improves the situation (what could be worse than not displaying the entry at all…), it is unrelated to this issue.

> It makes very easy to overlook that an entry occurs several times on a same line. There is no hint that a line contains several matches, apart from the highlights that the user has to look for. Actually, for long lines that don't fit on screen width, there is no hint at all, unless the user horizontally scrolls the results pane while looking for the highlights...

donho:
The default value will be kept.
However, if there are many other voices are for not using "show only one entry" as default, it be might be considered.

PR #12016, opened 2022-08-15
Fix append extension issue when use Korean IME

80ROkWOC4j:
Fix Korean IME issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11582
This PR is an improved version of PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11912
Now only works on Korean keyboard layout.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -251,6 +251,10 @@ public:
 	IFACEMETHODIMP OnFileOk(IFileDialog* dlg) override
 	{
 		_lastUsedFolder = getDialogFolder(dlg);
+
+		if (isCJKMode)
+			return S_FALSE;
+
 		return S_OK;
 	}
 	IFACEMETHODIMP OnFolderChange(IFileDialog*) override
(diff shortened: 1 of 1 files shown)


donho:
@mere-human
Could you validate this PR to make sure there won't be any regression?

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -617,15 +630,88 @@ class FileDialogEventHandler : public IFileDialogEvents, public IFileDialogContr
...
+		static bool isCandidateMode = false;
+		static bool wasCandidateMode = false;
+		auto IsHangul = [](WCHAR locale[])
+		{
+			auto hwnd = GetFocus();
+			auto himc = ImmGetContext(hwnd);
+			auto isHangul = ImmGetOpenStatus(himc);
+			ImmReleaseContext(hwnd, himc);
+
+			return !wcscmp(locale, L"ko-KR") && isHangul;
```
mere-human, line 642: 1. Add a null check for `locale`.
2. Is this condition needed only for Korean? Do we need to check any other language?

mere-human:
> @mere-human Could you validate this PR to make sure there won't be any regression?

@donho I can't download the build from AppVeyor because it deletes artifacts older than 1 month.
I can build from sources and test locally probably on weekend if that's ok.
Visually, the logic is not very clear to me.

donho:
@mere-human 
> Visually, the logic is not very clear to me.

In this case, I prefer to not include it. Thank you for your review.

PR #12017, opened 2022-08-15
Function Completion menu item to cycle through choices

### LINKED ISSUE #11950: Function Completion: keystroke to cycle through choices
### Description of the Issue

Function Completion allows multiple “overload” variants in the definitions. I know when it pops down, you can click on the up and down arrows to cycle through those variants:
image

However, there doesn't appear to be a keyboard shortcut for that action.

In other programs, like Visual Studio, when doing Function parameter Completion, you can use the up and down arrows to navigate between the overloaded function definition popups: 
image

In Notepad++, when doing normal "word" autocomplete, you can use the up and down arrows to select between the various options: 
image

But in the Notepad++ Function Completion, it doesn't have that.

This was brought up in a post about IDE features in Notepad++ and other editors [in a Perl forum](https://perlmonks.org/?node_id=11145809).  This was a confusing discrepancy for the person trying Notepad++'s implementation for the first time.  Being able to navigate through the Function Completion choices by keyboard would make another selling point for Notepad++ in such discussions when talking with people in forums not related to Notepad++. ;-)

1. Language > Perl (or any other language that has Function Completion definitions
2. type `split(`, or use Ctrl+Shift+Spacebar after navigating to the right of the paren in `split(...)`
3. You can navigate the overload options by clicking on the arrow buttons, but you cannot navigate through those options using the keyboard

### Expected Behavior

Allow a keyboard input there: it doesn't really matter whether it's Up and Down arrows, or some shift+ or alt+ the arrows, or something similar.  (For comparison: VS uses the standard up/down arrows.)

### Actual Behavior

You cannot navigate those choices with the keyboard.

### Debug Information

```
Notepad++ v8.4.4   (64-bit)
Path : c:\usr\local\apps\npp\npp.8.4.4.portable.x64\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 20H2
OS Build : 19042.1586
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
```

vinsworldcom:
Fix #11950

Add menu item in the Edit => Auto-Completion sub menu to cycle through overloaded function call tips.

- This is the same menu spot where function call tips can be activated (Edit => Auto-Completion => Function Parameters Hint)
- I do not re-order, simply add my option to the end, though a re-ordering would be nice to group the "function" related items together
- I do not provide a language translation (english.xml) but it can be easily added with:
`    <Item id="50010" name="Function Parameters Next"/>`

- I only provide a "forward" / "next" (translates to a `position=2` in Scintilla.  This allows for overloads to go 1=>2=>3 .. =>n => 1, but not 3=>2=>1.  Adding another menu item "Function Parameters Previous" would be trivial to call with `position=1` but would add yet another menu item to just scroll through the choices in the opposite direction - I didn't think it was worth it.
- Added to shortcut mapper, but shortcut.cpp needs a specific check for this command since its number (50010) is out of the current range due to other commands already using those IDs:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f2c4ecd7e1bfd5d7d5c14c854438c02f2ba0909a/PowerEditor/src/resource.h#L425-L435


PowerEditor/src/menuCmdID.h
```diff
@@ -175,6 +175,7 @@
     #define    IDM_EDIT_AUTOCOMPLETE_CURRENTFILE                (50000 + 1)
     #define    IDM_EDIT_FUNCCALLTIP                             (50000 + 2)
     #define    IDM_EDIT_AUTOCOMPLETE_PATH                       (50000 + 6)
+    #define    IDM_EDIT_FUNCCALLTIP_NEXT                        (50000 + 10)
```
dfs-, line 178: Wouldn't be 50000 + 3 more logical? The gap before IDM_EDIT_AUTOCOMPLETE_PATH was left intentional I guess.

vinsworldcom: 50003 is already in use as is up to 50005.  Up to 50009 are in, but commented.  I wanted to avoid any issues:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f2c4ecd7e1bfd5d7d5c14c854438c02f2ba0909a/PowerEditor/src/resource.h#L425-L435

Cheers.

donho:
@vinsworldcom 
Could you provide the instructions to play and see the feature you've added (after pulling your PR)?

vinsworldcom:
@donho happily!  After this PR is incorporated, you can go to the Edit => Auto-Completion menu and see the new item "Function Parameters Next".  I find it best to set a shortcut for this, so menu Settings => Shortcut Mapper... and filter on "function".  I set the `Alt+Down Arrow` shortcut myself:

image

Next, open a new document and switch the language to Perl.  Perl because I know the AutoCompletion file for Perl has some nice overloads to test with.

Type `print(` and you should see the calltip come up with indication that it is 1 out of 4 options.

image

Now, use the shortcut you created (in my case, `Alt+Down Arrow` to cycle through 2 of 4, 3 of 4, 4 of 4 and back to 1 of 4.

image


Hope this helps!  It seems a pretty easy addition, the only "issue" I had was the menu item number "50010" since the others were taken and then not being able to just expand the `else if()` range in the 'shortcut.cpp' file, having to add the `or` statement.  I've been running with this change since I made the PR and it's pretty nifty!

Cheers.

alankilborn:
@vinsworldcom I have probably dumb questions:

* How come there is only a provision for down and not up?

* Why does this need a remappable shortcut key?  When the completion box is visible, shouldn't down (or up) arrow presses change the current entry?

vinsworldcom:
@alankilborn 

> How come there is only a provision for down and not up?

Typically, the Overloads start at 1 of # so you want to look at the next (rather than previous) and by looking at the next up to the last one and then one more, it rolls back over to 1.  If it stopped, then certainly we'd need a "previous".  I can add the previous, it just seemed like overkill to be able to scroll both directions for such a simple feature, but alas, users will probably ask for it.  Should I add that?

> Why does this need a remappable shortcut key? When the completion box is visible, shouldn't down (or up) arrow presses change the current entry?

That's exactly the issue described by the community post.  You can use the mouse to click the arrows, but there is no way to use the keyboard.  This PR - by assigning a shortcut to some keyboard combination - fixes that.

Cheers.

alankilborn:
>I can add the previous, it just seemed like overkill to be able to scroll both directions for such a simple feature, but alas, users will probably ask for it. Should I add that?

I would certainly leave that to your discretion. :-)

>That's exactly the issue described by the community post. You can use the mouse to click the arrows, but there is no way to use the keyboard. 

Right.  But I think my point was, can't it be fixed by just handling "down arrow" (or "up arrow") keypresses when the completion box is visible?  To my way of thinking, that would be what the user expects to happen.  If the user has to remember a "special" keycombo (e.g. Alt+down) then it IMO kills the utility of it for all but the most die-hard of function completion keyboardists.

vinsworldcom:
> Right. But I think my point was, can't it be fixed by just handling "down arrow" (or "up arrow") keypresses when the completion box is visible? To my way of thinking, that would be what the user expects to happen. If the user has to remember a "special" keycombo (e.g. Alt+down) then it IMO kills the utility of it for all but the most die-hard of function completion keyboardists.

That's a good question.  I'm sure it *can*, but it would involve a lot more coding that I know how to do.  I think you'd need to hook the keyboard to intercept up/down arrow presses and check if a calltip is active, and if so, then send the `SCN_CALLTIPCLICK` notification.  I also don't know how you'd handle a call tip without the overloads - in other words, nothing to scroll through.  Scintilla let's you know a call tip is active [`SCN_CALLTIPACTIVE`](https://www.scintilla.org/ScintillaDoc.html#SCI_CALLTIPACTIVE), but not whether it has overloads.  So if it doesn't do we just intercept and "drop" the up/down arrow key presses, or send them on to the editor window, which of course scrolls the cursor and by default closes the calltip?

Cheers.

PowerEditor/src/Notepad_plus.rc
```diff
@@ -509,6 +509,7 @@ BEGIN
             MENUITEM "Word Completion",             IDM_EDIT_AUTOCOMPLETE_CURRENTFILE
             MENUITEM "Function Parameters Hint",    IDM_EDIT_FUNCCALLTIP
             MENUITEM "Path Completion",             IDM_EDIT_AUTOCOMPLETE_PATH
+            MENUITEM "Function Parameters Next",    IDM_EDIT_FUNCCALLTIP_NEXT
```
donho, line 512: 1. I would rename it as "Function Parameters Next Hint"
2. Put it just blow `MENUITEM "Function Parameters Hint", IDM_EDIT_FUNCCALLTIP`
3. Add a new "Function Parameters Previous Hint" between "Function Parameters Hint" & "Function Parameters Next Hint"

PowerEditor/src/Notepad_plus.h
```diff
@@ -563,6 +563,7 @@ friend class FileManager;
 	void autoCompFromCurrentFile(bool autoInsert = true);
 	void showFunctionComp();
 	void showPathCompletion();
+	void showFunctionNext();
```
donho, line 566: Since we have added new item "Function Parameters Previous Hint", the function prototype should be:
`void showFunctionNextHint(bool isNext); // set isNext to false to reach the previous one`

PowerEditor/src/Parameters.cpp
```diff
@@ -149,6 +149,7 @@ static const WinMenuKeyDefinition winKeyDefs[] =
 	{ VK_SPACE,   IDM_EDIT_AUTOCOMPLETE_PATH,                   true,  true,  false, nullptr },
 	{ VK_RETURN,  IDM_EDIT_AUTOCOMPLETE_CURRENTFILE,            true,  false, false, nullptr },
 	{ VK_SPACE,   IDM_EDIT_FUNCCALLTIP,                         true,  false, true,  nullptr },
+	{ VK_NULL,    IDM_EDIT_FUNCCALLTIP_NEXT,                    false, false, false, nullptr },
```
donho, line 152: I'm wondering if we should add Alt-Down & Alt-Up by default.
What do you people think?

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -1235,7 +1235,7 @@ CommandShortcut::CommandShortcut(const Shortcut& sc, long id) :	Shortcut(sc), _i
 		_category = TEXT("File");
 	else if ( _id < IDM_SEARCH)
 		_category = TEXT("Edit");
-	else if (_id >= IDM_EDIT_AUTOCOMPLETE and _id <= IDM_EDIT_AUTOCOMPLETE_PATH)
+	else if (_id >= IDM_EDIT_AUTOCOMPLETE and _id <= IDM_EDIT_AUTOCOMPLETE_PATH or _id == IDM_EDIT_FUNCCALLTIP_NEXT)
```
donho, line 1238: I know you just follow the style, but it's the old style which should be changed. Let's do:
`else if ((_id >= IDM_EDIT_AUTOCOMPLETE) && (_id <= IDM_EDIT_AUTOCOMPLETE_PATH) || (_id == IDM_EDIT_FUNCCALLTIP_NEXT))`

donho (changes requested):
Please add new entries into "english.xml" and consider the given suggestions.

PowerEditor/src/Parameters.cpp
vinsworldcom: > I'm wondering if we should add Alt-Down & Alt-Up by default.
> What do you people think?

I prefer no defaults and let people create their own.  The other thought is @alankilborn 's comment about somehow making this "just work" with `UP`/`DOWN` arrows themselves when a calltip window is active, like `UP`/`DOWN` arrows just work on AutoComplete combo boxes when they appear.  I don't know how / if this is possible, but maybe worth looking at?

Cheers.

vinsworldcom:
> Please add new entries into "english.xml" and consider the given suggestions.

@donho - just pushed a new commit with your requested changes.  I did compile and test myself and the previous / next seemed to work as expected once I manually mapped shortcuts in the "Shortcut Mapper...".

Please pay attention to the `else if()` in shortcut.cpp; since we added both previous and next, it required a more "complex" conditional and I want to make sure you like the format I used.

Cheers.

donho:
@vinsworldcom 
> Please pay attention to the else if() in shortcut.cpp; since we added both previous and next, it required a more "complex" conditional and I want to make sure you like the format I used.

It's OK for this PR at least.

donho (changes requested):
Add ALT-UP & ALT-DOWN as their default shortcuts.

vinsworldcom:
@donho 

> Add ALT-UP & ALT-DOWN as their default shortcuts.

Done!  Tested on local portable 64-bit install - worked OK.

> Yes, I saw the coversation and I do agree that UP/DOWN is more intuitive and ideal.  But it needs to hack the code in Scintilla to do that.

That was what I thought - I see the Scintilla code for AutoComplete essentially instantiates a combobox which gets the `Up` / `Down` arrows handled.  I don't see where Scintilla CallTip source instantiates a Windows control - assuming it draws some kind of window, but again, don't see the callback handler so I thought some hacking would be needed for Scintilla to capture `Up` / `Down` arrows.

Alas, happy to see this feature make it into N++!

Cheers.

donho (approved):

PR #12038, opened 2022-08-25
new theme based on Ruby Blue

airduv:
New theme based on Ruby Blue but started from DarkModeDefault.xml
It's a bit darker, I hope more fancier and more up-to-date than Ruby Blue (as the original Ruby Blue missed several languages and latest theme improvements)

### DIFF
--- a/PowerEditor/installer/themes/DarkSide-of-RubyBlue.xml
+++ b/PowerEditor/installer/themes/DarkSide-of-RubyBlue.xml
@@ -1,0 +1,1404 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+File name:           Zenburn.xml
+Style Name:          Zenburn
+Description:         Zenburn-like style for Notepad++.
+                     Inspired by the original Zenburn colorscheme for Vim by Jani Nurminen.
+                     Official Vim Zenburn home page: http://kippura.org/zenburnpage/
+Supported languages: All the languages supported by release 7.7
+Created by:          Jani Kesänen (jani dot kesanen gmail com)
+Released:            25.06.2019
+License:             GPL2
+-->
+<NotepadPlus>
+    <LexerStyles>
+        <LexerType name="actionscript" desc="ActionScript" ext="">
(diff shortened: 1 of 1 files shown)


alankilborn:
@airduv User-supplied themes aren't likely to get accepted.  Just so you know.

alankilborn:
@donho @pryrt Any plans for a "user supplied themes" repo similar to the UDL repo?  If so then I think this would be the proper location for @airduv 's effort in this PR.

donho:
> @donho @pryrt Any plans for a "user supplied themes" repo similar to the UDL repo? If so then I think this would be the proper location for @airduv 's effort in this PR.

@pryrt  what do you think about the suggestion of @alankilborn ?

pryrt:
I think having a theme repo would be easier to maintain than having users submit new themes to the main repo.

If you created the repo, I could rework the CONTRIBUTING and README  based on the UDL version

vinsworldcom:
> I think having a theme repo would be easier to maintain than having users submit new themes to the main repo.

To go a step further, maybe removing all but the Dark Theme from the themes directory in Notepad++ distributions and letting users install themes from a repo.  Not that they would be anymore up-to-date if they are neglected there, but it maybe makes managing Notepad++ repo itself and pull requests to it easier? 

Just a thought ...

Cheers.

alankilborn:
@vinsworldcom It is a good idea, except it requires that "extra step" of an additional download for those that might want to quickly try out a theme, or even quickly understand what is meant by the term "theme".  It might be a good thing to keep some themes bundled with Notepad++ and then let users know there is a place where they can download more from. like UDL does on the _Language_ menu:

image

I envision a "Notepad++ Themes Collection" link, perhaps close to here:

image

or here:

image

pryrt:
Like @alankilborn , I think keeping the existing themes in the main app is best idea.  And for the additional themes, adding a link to the repo like the existing UDL link is a great idea.

donho:
I'm rather on the same page with @alankilborn & @pryrt 
Let's do it this way.

donho:
@airduv 
Could you close this PR, and do a new PR on:
https://github.com/notepad-plus-plus/nppThemes
please?

Issue #2341, opened 2016-09-26
Notepad ++ crashes everytime 


pshashank8:
### Description of the Issue

Whenever i open any file in notepad ++ it keeps crashing after 1,2 mins

I have other text editors also installed in the same machine, does this cause some conflict ? But i've selected notepad ++ to be my default editor
![npp](https://cloud.githubusercontent.com/assets/16872447/18828480/bfb6fa8c-83f5-11e6-8913-88fac189b5b4.PNG)

Please suggest a fix for this issue

cmeriaux:
Please, report debug info available in "?" menu
You should confirm your porblem with the last version (V7)
You should try to remove plugins

donho:
@pshashank8 You cannot just go to see the doctor by only telling him/her "Cure me please, I'm sick". 
So we can do nothing about your report. Please use issue template : 
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/ISSUE_TEMPLATE.md
to report your crash (and reopen it of course).

Issue #2338, opened 2016-09-26
its make a generally used software


raghvendrasingh92:
### Steps to Reproduce the Issue

1. 
2. 
3.

Issue #1441, opened 2016-02-04
Several tab's problem


jporter-sdl:
Hi guys!

Case:
Open 25 files (more tab's than it can display).
Close several tab's by pressing the X icon on the tab's itself until you have one left.

Now you have several, say 15 files open but it will only display 1 tab. It should just always display the maximum it can.

donho:
I believe this issue has been fixed since long time ago.
Update your Notepad++ to the latest version (v7.6.6).
If the issue persists, please create a new issue for it.
Thank you.

Issue #98, opened 2015-05-19
default FunctionList sorting


freggeldoe:
I always sort functions by name and have to change the sort order every time I reopen notepad++ for each of my opened files. Please keep the last sort-order as a default for newly opened files as well as open files when reopening notepad++ or provide a setting to define the default sort order.

Chrissstrahl:
I also like this idea, because I have the same problem :)

amiga-500:
And its still a problem after 2 years.

jmihalko:
I agree, this is a real pain.  I think just a setting to define the default sort order would be sufficient and hopefully quicker to implement so maybe we can still get the feature.....before I retire :)

elmonty:
If NP++ would simply remember the last sort order, that would be great!

donho:
Function list sorting has been implemented since long time ago.

MetaChuh:
@donho
verified label because:

the default setting is unsorted.
author wanted the alphabetical sort setting to be saved and remembered.
alternative he suggested to make alphabetical sort as default.

freggeldoe:
@donho: Please reconsider your close-decision. I was asking for a way to keep alphabetical functionList-sorting either by changed default or by a setting.

PR #12045, opened 2022-08-27
added scintilla 5.3.0 and lexilla 5.1.9

chcg:
https://www.scintilla.org/ScintillaHistory.html
[Release 5.3.0](https://www.scintilla.org/scintilla530.zip)

    Released 27 August 2022.
    Added change history which can display document changes (modified, saved, ...) in the margin or in the text.
    Add SC_MARK_BAR marker and INDIC_POINT_TOP indicator. SC_MARK_BAR is an outlined and filled rectangle that takes the full height of the line and 1/3 of the margin width. To give a connected appearance, it displays even on wrapped lines and draws end caps on the first and last line. INDIC_POINT_TOP is the same as INDIC_POINT but draws at the top of the line instead of the bottom. INDIC_POINT and INDIC_POINTCHARACTER were tweaked to be 1 pixel taller and 2 pixels wider in two-phase draw mode to be clearer. The translucency of INDIC_COMPOSITIONTHICK can be changed with SCI_INDICSETOUTLINEALPHA.
    Improve drawing of rounded rectangles on Direct2D.
    Line state optimized to avoid excess allocations by always allocating for every line. This makes SCI_GETMAXLINESTATE less useful although it can still distinguish cases where line state was never set for any lines. [Feature #1441](https://sourceforge.net/p/scintilla/feature-requests/1441/).
    Add SC_FOLDACTION_CONTRACT_EVERY_LEVEL option to contract every level for SCI_FOLDALL. [Bug #2340](https://sourceforge.net/p/scintilla/bugs/2340/).
    Enable multiline regex for gcc and clang when REGEX_MULTILINE defined. This requires gcc 11.3 or clang 14. [Bug #2338](https://sourceforge.net/p/scintilla/bugs/2338/).
    Stop including STYLE_CALLTIP when calculating line height. Allows a large font to be used for calltips without affecting text display. [Bug #2341](https://sourceforge.net/p/scintilla/bugs/2341/).
    Fix incorrect display of selection when printing in some modes. [Bug #2335](https://sourceforge.net/p/scintilla/bugs/2335/).
    Fix crash on Qt when showing autocompletion and the caret isn't on a screen. Move autocompletion onto screen when above it.

https://www.scintilla.org/LexillaHistory.html
[Release 5.1.9](https://www.scintilla.org/lexilla519.zip)

    Released 27 August 2022.
    Julia: Parse unicode forall and exists as identifiers #42314. [Pull request #98](https://github.com/ScintillaOrg/lexilla/pull/98).
    Julia: Parse apostrophe as char and not adjoint after exclamation. [Issue #97](https://github.com/ScintillaOrg/lexilla/issues/97), [Pull request #98](https://github.com/ScintillaOrg/lexilla/pull/98).
    Properties: Don't set header flag for empty section. [Issue #96](https://github.com/ScintillaOrg/lexilla/issues/96).


vinsworldcom:
@chcg @donho 

> Released 27 August 2022.
> Added change history which can display document changes (modified, saved, ...) in the margin or in the text.

[This feature](https://www.scintilla.org/ScintillaDoc.html#ChangeHistory) is AMAZING!  It basically removes the need for my [Changed Lines plugin](https://github.com/vinsworldcom/nppChangedLines).  I tried and enabled with NppExec:

```
SCI_SENDMSG 2780 3
```

Then this happens when typing / saving:

image

Notice the markers!  

# HOWEVER

The markers aren't what Scintilla says they should be and this is because Scintilla is using markers 21 - 24 for the change markers and Notepad++ uses those for something else.

So while this merges nicely and ran for me in a quick test, we probably need to renumber the Notepad++ markers 21-24 for this feature to work.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/08acfe99b8dcfef40c7e7c244fe5c75cab0338b5/PowerEditor/src/ScintillaComponent/ScintillaEditView.h#L116-L119

Cheers.

chcg:
As change history is not on without code modifications to N++ I would see the code for the activation + the necessary modifications @vinsworldcom mentioned as a new PR.

alankilborn:
> [This feature](https://www.scintilla.org/ScintillaDoc.html#ChangeHistory) is AMAZING! It basically removes the need for my [Changed Lines plugin](https://github.com/vinsworldcom/nppChangedLines)

Your comment might have been better placed in my new feature request issue, see #10246 !!
But obviously, yes, I agree that this is a great addition to Scintilla and will be to Notepad++ as well.

vinsworldcom:
@chcg @donho 

> As change history is not on without code modifications to N++ I would see the code for the activation + the necessary modifications @vinsworldcom mentioned as a new PR.

Agree.  And @alankilborn has an issue #12046 to track the requirements for a PR.  I'll move all further commenting to that issue.

Cheers.

donho:
@vinsworldcom 
> So while this merges nicely and ran for me in a quick test, we probably need to renumber the Notepad++ markers 21-24 for this feature to work.

Indeed. According [Scintilla documentation](https://www.scintilla.org/ScintillaDoc.html#Markers):
```
Marker numbers 21 to 24 are used for [Change history](https://www.scintilla.org/ScintillaDoc.html#ChangeHistory) if that is enabled but are otherwise free for application use.

Marker numbers 0 to 20 have no pre-defined function; you can use them to mark syntax errors or the current point of execution, break points, or whatever you need marking. If you do not need folding, you can use all 32 for any purpose you wish.
```
But in scintilla.h
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/include/Scintilla.h#L122
All the number are used. 

What's your suggestion?

AndiKiissel:
@donho 
`SC_MARK_...` are marker **symbols**, as I understand.
Currently N++ MARK_BOOKMARK (=24) uses SC_MARK_RGBAIMAGE (=30).

vinsworldcom:
@donho both @AndiKiissel and @alankilborn correctly explain this.  The `SC_MARK_` in Scintilla are the available "icons" to use as markers.  The `SC_MARKNUM` in Scintilla are the actual 0-31 markers defined:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b627d6a4554b095487d41f11765ab106d1884c9d/scintilla/include/Scintilla.h#L157-L167

I think in keeping with the current approach, we renumber the Notepad++ markers from 21-24:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b627d6a4554b095487d41f11765ab106d1884c9d/PowerEditor/src/ScintillaComponent/ScintillaEditView.h#L116-L119
 
to 17-20 to keep them at the top of the range the way Scintilla does, leaving the 0-16 lower range open for plugins to use.

What does everyone else think?

Cheers.

alankilborn:
It surely seems the most reasonable thing to do.

ozone10:
@chcg @donho 
Should flags /O2 and /Ot be enabled for scintilla and lexilla to maybe improve speed?
image
image

donho:
> Should flags /O2 and /Ot be enabled for scintilla and lexilla to maybe improve speed?

@ozone10 Have you noticed any performance change with the arguments you suggested?

ozone10:
@donho
sorry for late answer I was away. There might be some improvements in specific cases for scintilla. I tried those flags but because my workflow does not deal with intensive stuffs (big files), I didn't noticed anything. I've mainly asked because Notepad++ is already using them.

donho:
> I've mainly asked because Notepad++ is already using them.

That makes sense...
@chcg Could both Scintilla and Lexilla adapt the mentioning changes for aligning with Notepad++ vcxproj ?

chcg:
@donho Adapted the release builds of scintilla and lexilla to maxspeed.

donho:
@chcg 
Thank you for the modification and for the notification of size change!

Back to the 1st release of Notepad++ I still used 3"1/2 floppy disk (Verbatim) for transporting Notepad++ binary from  PC to PC to test it. Nowadays a text editor like MS VS CODE needs a GB of disk for enough room to download. So I guess 200 KB more is reasonable :)

donho:
@vinsworldcom 
Thank you for the suggestion.
The commit is here: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/9455684b429449f668c1aedb77ec9dcb1642c244

PR #12048, opened 2022-08-28
"Find All in current document" feature performance improvement.

### LINKED ISSUE #11878: Find All in current document freezes NPP++ for more than 2 minutes
Find All is deadly slow

### Description of the Issue
_Find All in current document_ for large files takes too much (minutes, not seconds!)

1. open for example this huge file [workshop.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9053080/workshop.zip)
2. 'Find All in current document' for ":" (443218 matches)
3. repeat the same search at point 2 multiple times

### Expected Behavior
Find and mark all occurrences in a few secs

### Actual Behavior
Can take more than 2:33 minutes on i7-8750h, 32GB RAM. Can also freeze forever sometimes randomly on repeat.
Same thing on other programs like 'CudaText' is almost instant, on 'NotepadNext' takes 2 secs, on 'Geany' takes less than 8 secs...

Ashfaaq18:
fix #11878 

This fix should improve the performance significantly. In my test results (using the workshop.zip file attached in the issue #11878). The "Find All in current document" feature to find the symbol ":" in this document produced the following results,
Before this fix. time taken in my pc ~36s
After this fix. time taken in my pc ~6s

This fix also improves the performance of the feature "Find in search results" in the search result window.


Yaron10:
@Ashfaaq18,

That's a significant major time improvement! 👍 
Thank you for your great work.

PR #12054, opened 2022-08-29
Fix supported OSes

LandarXT:
This is probably the smallest PR you've ever seen but I just noticed there was a small inconsistency in the supported os' document


alankilborn:
Note in the contributing guidelines typos are not to be fixed by submission of PR; simply create an issue that points out typos and if validated they will be fixed by the author himself.

PR #12065, opened 2022-08-31
Fix portuguese.xml

hugok79:


donho:
@hugok79 
I just reverted the bad commit.
Please create a new branch and a new PR to have a correct commit.

PR #12083, opened 2022-09-02
Fix button text in dark mode regression from Wine fix

ozone10:
Fix regression from https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f2c4ecd7e1bfd5d7d5c14c854438c02f2ba0909a

fix https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f2c4ecd7e1bfd5d7d5c14c854438c02f2ba0909a#commitcomment-82841877


donho:
Thank you @ozone10 for your (very) quick fix.

Issue #2271, opened 2016-09-12
[Feature request] Colored tabs


sas33:
### Description of the Issue

i got lost sometimes when i work with large numbers of php pages or html. if i could group by colors
the tabs it will be more easier to reach what i want without searching for the filename in the tabs.
### Steps to Reproduce the Issue

1. 
2. 
3. 
### Expected Behavior

colored tabs by the choice of the user (red,blue, green, etc)
### Actual Behavior

they are all grey and indistinguishable

Pyre909:
Why not set the color according to file type? Just food for thought.....

sas33:
@Pyre909 ok but when i have like 15 php file types and i want to group them by colors this will come in handy

spaxio:
It's old ticket, but I really needed this so I forked the repo and made it possible.
https://github.com/spaxio/notepad-plus-plus-colored-tabs/commit/db92d1fce38d110248cd6437e81ef8ba8c0ab422

https://i.imgur.com/Rlgt94T.png
We can select between 5 different colours. It works in dark/light mode.

I could open a pull request if @donho is interested, I'm pretty open to adjusting it more. 
But if not that's cool too. As I said I really needed this in my npp workflow :D

donho:
@spaxio 
Please do a PR - I will try it out.
It looks very promising - from the aspect of GUI and simplicity of modification.

alankilborn:
Would this color be saved in session?
Does this open the door to tab color by file type?

spaxio:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098

>Would this color be saved in session?
Yes it's saved in session.xml file 
>Does this open the door to tab color by file type?
We have access to the file type so it's possible but we would need to add some simple algorithm calculating colour for given file extension (or parent folder)

sas33:
Thank you!

JergenRoss:
> We can select between 5 different colours. It works in dark/light mode.

um does User can use custom color will work? also more select 5 to up?

JergenRoss:
> Why not set the color according to file type? Just food for thought.....

I think my suggest is name file type colored `.txt` `.bat` with tag instead tab

example:
tab yellow: `new document [.txt]` 

`[]` is colored by according to file type
`tab yellow` colored is user want to group them by colors.

User can prefer change color different file types 

Im not sure if who want color by file types or not.

Issue #12072, opened 2022-08-31
Close All - Confirm Dialog


Dvorak-Stepan:
### **Please add a function:**
**Preferences-->MISC.**
- Enable Close All - Confirm Dialog

ArkadiuszMichalski:
It worth recalling that this was rejected:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4795

Though calling `Close All` by mistake may be a problem for some users.
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10776

@donho Any thoughts, because I see that PR for the same problem has already appeared:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12116

alankilborn:
IMO there's really no excuse for not adding this.

PR #12098, opened 2022-09-04
Added feature: we can set colour for individual tabs

spaxio:
Fixes: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/2271

This pull request implements a feature where we can select a colour for an individual tab.

We can choose up to 5 hues (red/green/blue/orange/purple) in View/Tab/Color menu.
Colours are saved in session.xml as integer, where -1 (or no value at all) means no colouring.

I've selected 5 because in my opinion, anything more would get more confusing with lots of tabs.

This is how it looks like:
1. https://i.imgur.com/lutHHgB.jpg (lightmode with darkened inactive tabs)
2. https://i.imgur.com/ahtDt7x.jpg (lightmode)
3. https://i.imgur.com/Rlgt94T.png (darkmode)

It works in vertical mode too.
The colour is not retained when we try to pull a tab into a separate window.

Things worth discussing
- Should we allow user to select their own hues in the settings (right now they are fixed).
- Do we want to add possibility to select colour from tab context menu?
- When the tab is active then colouring is skipped, but maybe we could also colourize that top thin bar (which is orange by default)?


ArkadiuszMichalski:
Answer for asked question:
1) Yes, a few fixed values may not meet everyone's expectations.
2) Yes, it will be more convenient to set this in the tab context menu (at least for me).
3) I have no opinion, but if necessary in the theme settings in `Global Style> Active tab focused` we can add a switch `[ ] Overwrite with individual tab color`.

alankilborn:
>Should we allow user to select their own hues in the settings

Yes, like styles 1 - 5 are available there; it would be a disappointment to users to not be able to customize

>Do we want to add possibility to select colour from tab context menu?

Yes, with color samples, like styling:

image

Also, in a submenu (again like styling) would be appreciated.

Yaron10:
@spaxio,

Thank you for your work. 👍 

Would it be possible to apply the selected color to the *active tab*?
Isn't it confusing that the user does not see any effect to the color selection until after switching tabs?

alankilborn:
>Would it be possible to apply the selected color to the active tab?
Isn't it confusing that the user does not see any effect to the color selection until after switching tabs?

Wouldn't it be better served to only change the color of the narrow strip of colour in the top of the active tab to the colour user just selected?
I really don't want the active tab's entire colour to possibly match other tabs -- I really want it to stand out..it's the _active_ tab, after all.

Yaron10:
@alankilborn,

We agree that when the user is selecting a color, they should see some effect before switching tabs.

And I certainly see your point. 
But wouldn't it still be confusing if that effect is initially on the upper strip, and then (on switching tabs) on the tab's background color?

Wouldn't the upper strip suffice to make the active tab stand out?

As you and the others understand. :)

alankilborn:
>But wouldn't it still be confusing if that effect is initially on the upper strip, and then (on switching tabs) on the tab's background color?

Well, I haven't actually _seen_ it, but to me that doesn't sound confusing.

>Wouldn't the upper strip suffice to make the active tab stand out?

Not sure at all what this means, in the context of the recent conversation.

Yaron10:
> Wouldn't the upper strip suffice to make the active tab stand out?
>> Not sure at all what this means, in the context of the recent conversation.

In response to your previous comment:
> I really don't want the active tab's entire colour to possibly match other tabs -- I really want it to stand out..it's the active tab, after all.

alankilborn:
I suppose if the upper colour strip is really garish it would help it stand out.  But I wouldn't necessarily want the tab coloring that garish on other tabs of the same color.

spaxio:
Ok,

So I've added colourization of active bar highlight and in dark mode (default grey) it look awesome. But I'm not so sure about light ones (default theme, and with darkened inactive tabs). 

I'm also not sure about adding "Set colour" item to the document context menu since those should be related only to document itself. I would much more prefer to add it to the Tab context menu but the way it's implement right now I'm not able to add sub-menus there. For now I think it's best to leave it in View/Tab/Colour.

alankilborn:
>I'm also not sure about adding "Set colour" item to the document context menu

Correct.  Don't do it.

>I would much more prefer to add it to the Tab context menu

That could be a later change.
Although not having it will be a sad omission.
And it will be the first thing that users that want to colour tabs will request.

PowerEditor/src/Notepad_plus.rc
```diff
@@ -705,6 +705,17 @@ BEGIN
...
+            POPUP "Colour"
+            BEGIN
+                MENUITEM "Red",      IDM_VIEW_TAB_COLOUR_1
+                MENUITEM "Green",    IDM_VIEW_TAB_COLOUR_2
+                MENUITEM "Blue",     IDM_VIEW_TAB_COLOUR_3
+                MENUITEM "Orange",   IDM_VIEW_TAB_COLOUR_4
+                MENUITEM "Purple",   IDM_VIEW_TAB_COLOUR_5
+                MENUITEM SEPARATOR
+                MENUITEM "None",     IDM_VIEW_TAB_COLOUR_NONE
+            END
```
donho, line 718: 1. For the sake of usability, the 2nd level of sub-menu should be avoided.
2. The visual colour would be better than word, and it could be customizable by users.

So I would do : 
```
       POPUP "Tab"
        BEGIN
            MENUITEM "1st Tab",                 IDM_VIEW_TAB1
            MENUITEM "2nd Tab",                 IDM_VIEW_TAB2
            MENUITEM "3rd Tab",                 IDM_VIEW_TAB3
            MENUITEM "4th Tab",                 IDM_VIEW_TAB4
            MENUITEM "5th Tab",                 IDM_VIEW_TAB5
            MENUITEM "6th Tab",                 IDM_VIEW_TAB6
            MENUITEM "7th Tab",                 IDM_VIEW_TAB7
            MENUITEM "8th Tab",                 IDM_VIEW_TAB8
            MENUITEM "9th Tab",                 IDM_VIEW_TAB9
            MENUITEM SEPARATOR
            MENUITEM "Next Tab",                IDM_VIEW_TAB_NEXT
            MENUITEM "Previous Tab",            IDM_VIEW_TAB_PREV
            MENUITEM SEPARATOR
            MENUITEM "Move Tab Forward",        IDM_VIEW_TAB_MOVEFORWARD
            MENUITEM "Move Tab Backward",       IDM_VIEW_TAB_MOVEBACKWARD
            MENUITEM SEPARATOR
            MENUITEM "Apply Color 1",      IDM_VIEW_TAB_COLOUR_1
            MENUITEM "Apply Color 2",    IDM_VIEW_TAB_COLOUR_2
            MENUITEM "Apply Color 3",     IDM_VIEW_TAB_COLOUR_3
            MENUITEM "Apply Color 4",   IDM_VIEW_TAB_COLOUR_4
            MENUITEM "Apply Color 5",   IDM_VIEW_TAB_COLOUR_5
            MENUITEM "Remove Color",     IDM_VIEW_TAB_COLOUR_NONE
        END
```

donho (changes requested):
1. Move the 3rd level menu items to the 2nd level, and rename the label for the future customization.
2. Add these menu items also on the context menu of tab.
3. Add Color square to the menu items (as Style token).

The customization of colour is out of scope and will be in an another PR

donho:
BTW, I notice this PR is not from a new branch:

It's OK for the first time because there will no impact to the workflow of merging this PR. Please follow the suggestion for your next PR.

spaxio:
All points have been addressed. 🎉
I also adjusted colors for default themes (red, cyan, olive etc.). It looks better than before.

PowerEditor/src/localization.cpp
```diff
@@ -388,25 +388,31 @@ static const int tabContextMenuItemPos[] =
...
+   27,   // 20: Move to New Instance
+   28,   // 21: Open to New Instance
+    9,   // 22: Reload
+    2,   // 23: Close ALL to the Left
+    3,   // 24: Close ALL to the Right
+   12,   // 25: Open Containing Folder in Explorer
+   13,   // 26: Open Containing Folder in cmd
+   16,   // 27: Open in Default Viewer
+    4,   // 28: Close ALL Unchanged
+   14,   // 29: Open Containing Folder as Workspace
```
donho, line 415: The node `<TabBar>` in english.xml should be replaced by the following xml code:

```xml
            <TabBar>
                    <Item CMID="0" name="Close"/>
                    <Item CMID="1" name="Close All BUT This"/>
                    <Item CMID="2" name="Save"/>
                    <Item CMID="3" name="Save As..."/>
                    <Item CMID="4" name="Print..."/>
                    <Item CMID="5" name="Move to Other View"/>
                    <Item CMID="6" name="Clone to Other View"/>
                    <Item CMID="7" name="Full File Path to Clipboard"/>
                    <Item CMID="8" name="Filename to Clipboard"/>
                    <Item CMID="9" name="Current Dir. Path to Clipboard"/>
                    <Item CMID="10" name="Rename..."/>
                    <Item CMID="11" name="Move to Recycle Bin"/>
                    <Item CMID="12" name="Read-Only"/>
                    <Item CMID="13" name="Clear Read-Only Flag"/>
                    <Item CMID="14" name="Move to New Instance"/>
                    <Item CMID="15" name="Open in New Instance"/>
                    <Item CMID="16" name="Reload"/>
                    <Item CMID="17" name="Close All to the Left"/>
                    <Item CMID="18" name="Close All to the Right"/>
                    <Item CMID="19" name="Open Containing Folder in Explorer"/>
                    <Item CMID="20" name="Open Containing Folder in cmd"/>
                    <Item CMID="21" name="Open in Default Viewer"/>
                    <Item CMID="22" name="Close All Unchanged"/>
                    <Item CMID="23" name="Open Containing Folder as Workspace"/>
                    <Item CMID="24" name="Apply Color 1"/>
                    <Item CMID="25" name="Apply Color 2"/>
                    <Item CMID="26" name="Apply Color 3"/>
                    <Item CMID="27" name="Apply Color 4"/>
                    <Item CMID="28" name="Apply Color 5"/>
                    <Item CMID="29" name="Remove Color"/>
            </TabBar>
```

Then you replace whole `static const int tabContextMenuItemPos[] ={...]` by:

```cpp
static const int tabContextMenuItemPos[] =
{
	//  +-------------- The item positon in tab context menu
	//  |
	//  |        +------ Number in english.xml : <TabBar> CMDID
	//  |        |
        0,   //  0: Close
        1,   //  1: Close ALL BUT This
        5,   //  2: Save
        6,   //  3: Save As
       10,   //  4: Print
       32,   //  5: Move to Other View
       33,   //  6: Clone to Other View
       28,   //  7: Full File Path to Clipboard
       29,   //  8: Filename to Clipboard
       30,   //  9: Current Dir. Path to Clipboard
        7,   // 10: Rename
        8,   // 11: Move to Recycle Bin
       25,   // 12: Read-Only
       24,   // 13: Clear Read-Only Flag
       34,   // 14: Move to New Instance
       35,   // 15: Open to New Instance
        9,   // 16: Reload
        2,   // 17: Close ALL to the Left
        3,   // 18: Close ALL to the Right
       19,   // 19: Open Containing Folder in Explorer
       20,   // 20: Open Containing Folder in cmd
       23,   // 21: Open in Default Viewer
        4,   // 22: Close ALL Unchanged
       21,   // 23: Open Containing Folder as Workspace
       12,   // 24: Apply Color 
       13,   // 25: Apply Color 
       14,   // 26: Apply Color
       15,   // 27: Apply Color
       16,   // 28: Apply Color
       17,   // 29: Remove Color
       -1    //-------End
};
```

The comment of `tabContextMenuItemPos` wasn't clear. I have modified comment (above).
Each time if we add a new command, we have to "push back" its localization entry into `<TabBar>` in english.xml, so it's the index order of new command for localization. The index order of old commands keep the same.

Then we have to also "push back" new command's position into `tabContextMenuItemPos` and change the position of other commands, according where you insert the new command.

 Both array keep synchronized so Notepad++ know replace the label of menu item by localization string at the indicated position.

donho (changes requested):
Please fix the localization bug with the suggested solution.

ArkadiuszMichalski:
Maybe for another PR, but it would be nice to have these colors also for items in the Document List panel.

alankilborn:
With the addition of six items to the tab context menu, the menu seems to be getting to an almost obnoxious vertical size:

image

Shouldn't the new items be at the bottom? Users may complain that to "move to another instance" they now have to move the mouse a huge amount to reach it.  Of course, users always complain... :-)

I'm not sure if there is a "typical" size monitor that Notepad++ (maximized) is targeting.  In days of old, it might have been 1024 x 768...  Not sure about now.  But the point is, perhaps the tab context menu is getting dangerously close to appearing from top-screen-edge to bottom-screen-edge, and possibly even with the dreaded ▲ at the top and ▼ at the bottom.

Is it time to bite the bullet and do some submenus here?  There are obvious groupings (these "colours", the "close all..."s, the "opens", the "clipboards").

@donho Would this be considered?  Should I make an issue?

alankilborn:
> Maybe for another PR, but it would be nice to have these colors also for items in the Document List panel.

Sounds like a new issue.  :-)

spaxio:
>With the addition of six items to the tab context menu, the menu seems to be getting to an almost obnoxious vertical size

I agree, I would remove those. Features discovery will suffer but we can add those when submenu gets implemented. Or make it dependent on a flag in settings (in the future)

donho:
> I agree, I would remove those. Features discovery will suffer but we can add those when submenu gets implemented. Or make it dependent on a flag in settings (in the future)

It's shame to remove this feature from the tab context menu, that will make tab coloring meaningless.
OTOH sub-menu would be a good solution.

How about the following new tab context menu?

```
----------------------
Close                  
Close More           🢒
Save
Save As...
Open                 🢒
Rename
Reload
Move to Recycle Bin
Print
----------------------
Read Only
Clear Read Only Flag
Clipboard            🢒
Document to          🢒
----------------------
Apply Color          🢒
----------------------
```

**Close More** contains :
 * Close ALL BUT This
 * Close ALL to the Left
 * Close ALL to the Right
 * Close ALL Unchanged
  
**Open** contains:
* Open Containing Folder in
* Open Containing Folder in
* Open Containing Folder as
* ------------------------
* Open in Default Viewer

**Clipboard** contains:
* Full File Path to Clipboard
* Filename to Clipboard
* Current Dir. to Clipboard

**Document to** contains:
* Move to Other View
* Clone to Other View
* Move to New Instance
* Open in New Instance

**Apply Color** contains:
* Apply Color 1
* Apply Color 2
* Apply Color 3
* Apply Color 4
* Apply Color 5
* Remove Color

ozone10:
@donho 
I would use 
Close More  -> Close Multiple Tabs
Clipboard -> Copy to Clipboard
Document to -> Move/Clone Document
Apply Color  -> Apply Tab Color / Apply Color to Tab

It's better to be specific.

> **Clipboard** contains:
> 
>     * Full File Path to Clipboard
> 
>     * Filename to Clipboard
> 
>     * Current Dir. to Clipboard

Not really fan of "... to Clipboard", for submenus I would use

* Copy Full File Path
* Copy Filename
* Copy Current Dir.

Because other software are usually using "Copy".

spaxio:
What about instead of "Clipboard" as a submenu title we would would use "File Path"? Since items inside are a little more related with path than clipboard.

(oh and crazy idea, it would be great to do single menu items for all colors in a single row, similar to how win11 arrange copy/paste/cut in its contextmenu but I know WinAPI limitation 🙃)

ArkadiuszMichalski:
Proposal from https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098#issuecomment-1242770966 looks rational. The context menu with everything on one level is just too big. But I like more the names proposed in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098#issuecomment-1242776337.

alankilborn:
In general, I like @donho's layout https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098#issuecomment-1242770966 :-)

A lot of the proposed submenus appear elsewhere in the Notepad++ UI.

Example:  `Close More`
This is good because it would allow maximum reuse for localization; why define the same thing twice in exactly the same way?  And have 2 items for localization when 1 would suffice.

But `Clipboard` is new; why wouldn't this stay as `Copy to Clipboard` and get reuse from the _Edit_ menu's submenu of the same name?

In general, I DON'T like @ozone10 's layout https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098#issuecomment-1242776337 (sorry)

donho:
> Not really fan of "... to Clipboard", for submenus I would use
> *    Copy Full File Path
> *    Copy Filename
> *    Copy Current Dir.

It's necessary to rename these items in the main menu. I will see what I can do about it.

>  I would use
Close More -> Close Multiple Tabs

@alankilborn is right, **Close More** reuses the main menu term, with the same sub-menu items. So it should be remained.

OK for the other suggestions.

Here's the final version:

```
----------------------
Close                  
Close More           🢒
Save
Save As...
Open into            🢒
Rename
Reload
Move to Recycle Bin
Print
----------------------
Read Only
Clear Read Only Flag
Copy to Clipboard    🢒
Move Document        🢒
----------------------
Apply Color to Tab   🢒
----------------------
```

**Close More** contains :
 * Close ALL BUT This
 * Close ALL to the Left
 * Close ALL to the Right
 * Close ALL Unchanged
  
**Open** contains:
* Open Containing Folder in Explorer
* Open Containing Folder in Cmd
* Open Containing Folder as Workspace
* ------------------------
* Open in Default Viewer

**Copy to Clipboard** contains:
* Full File Path to Clipboard
* Filename to Clipboard
* Current Dir. to Clipboard

**Move Document** contains:
* Move to Other View
* Clone to Other View
* Move to New Instance
* Open in New Instance

**Apply Color to Tab** contains:
* Apply Color 1
* Apply Color 2
* Apply Color 3
* Apply Color 4
* Apply Color 5
* Remove Color

donho (approved):

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -2852,4 +2854,69 @@ namespace NppDarkMode
...
+	static constexpr IndividualTabColours individualTabHuesFor_DarkGreen { { HLSColour{0, 50, 60}, HLSColour{70, 50, 60}, HLSColour{144, 50, 60}, HLSColour{13, 50, 60}, HLSColour{195, 50, 60} } };
+	static constexpr IndividualTabColours individualTabHuesFor_DarkBlue  { { HLSColour{0, 50, 60}, HLSColour{70, 50, 60}, HLSColour{144, 50, 60}, HLSColour{13, 50, 60}, HLSColour{195, 50, 60} } };
+	static constexpr IndividualTabColours individualTabHuesFor_DarkPurple{ { HLSColour{0, 50, 60}, HLSColour{70, 50, 60}, HLSColour{144, 50, 60}, HLSColour{13, 50, 60}, HLSColour{195, 60, 60} } };
+	static constexpr IndividualTabColours individualTabHuesFor_DarkCyna  { { HLSColour{0, 60, 60}, HLSColour{70, 60, 60}, HLSColour{144, 70, 60}, HLSColour{13, 60, 60}, HLSColour{195, 60, 60} } };
+	static constexpr IndividualTabColours individualTabHuesFor_DarkOlive { { HLSColour{0, 60, 60}, HLSColour{70, 60, 60}, HLSColour{144, 60, 60}, HLSColour{13, 60, 60}, HLSColour{195, 60, 60} } };
+
+	static const IndividualTabColours individualTabHues              { { HLSColour{0, 210, 150}, HLSColour{70, 210, 150}, HLSColour{144, 210, 150}, HLSColour{13, 210, 150}, HLSColour{195, 210, 150} } };
```
donho, line 2877: I see that `individualTabHues` (light theme) and all other 7 `IndividualTabColours individualTabHuesFor_DarkXXX` arrays contains different 5 colour values.

It will be a major problem in the future if we want to make these 5 colours customizable.
Is it possible to have a function in which we pass a RGB value, and choice (light, dark, darkRed, darkBlue...) then the function in question return the RGB we need?

donho (changes requested):
Please fix the typo and the variable naming.

ozone10:
> > Not really fan of "... to Clipboard", for submenus I would use
> > 
> > * Copy Full File Path
> > * Copy Filename
> > * Copy Current Dir.
> 
> It's necessary to rename these items in the main menu. I will see what I can do about it.

Ok.
The other day I need to copy full path name and could not find it immediately, as I am used to term "Copy". 

> 
> > I would use
> > Close More -> Close Multiple Tabs
> 
> @alankilborn is right, **Close More** reuses the main menu term, with the same sub-menu items. So it should be remained.
> 

I see, I did not know about main menu, I rarely use it. I've used term "Close Multiple Tabs" to make it more familiar to users, as internet browsers are using it.

PowerEditor/src/NppDarkMode.cpp
spaxio: Notice that this is just implementation detail and what we are exposing is RGB `COLORREF getIndividualTabColour(...)`
So if in the future we want to add colour customization we only have to add a case for `if(g_colorToneChoice == customizedTone)`.

spaxio: >Is it possible to have a function in which we pass a RGB value, and choice (light, dark, darkRed, darkBlue...) then the function in question return the RGB we need?

And yes we can transform RGB into just a Hue and modify it according to theming rules back to RGB.

alankilborn:
What value are the SEPARATORS in @donho's https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098#issuecomment-1242844029 ?

image

They just make a user have to move the mouse further to get where they are going.

OK, maybe the one after `Print` is reasonable, but why have the one between "Move Document 🢒" and "Apply Color 🢒"?

Really there SHOULD be a separator between "Clear Readonly Flag" and "Copy to Clipboard 🢒" but even better than that would be putting "Read Only" and "Clear RO Flag" into a "Readonly 🢒" submenu.

"Open into 🢒" seems a bit obscure for opening Explorer, is there anything wrong with "Open Containing Folder 🢒" which is really obvious?

Summary of all my suggestions in this post:

```
Close
Close More             ▶
Save
Save As...
Open Containing Folder ▶
Rename
Reload
Move to Recycle Bin
Print
------------------------
Read Only              ▶  -----+---- Set Read-Only
                               |---- Clear Read-Only Flag
Copy to Clipboard      ▶
Move Document          ▶
Apply Color to Tab     ▶
```

PowerEditor/src/NppDarkMode.cpp
donho: @spaxio
> and modify it according to theming rules back to RGB.
 
When you say "modify it" you mean add it manually to other 7 vector right?
But if the customized color is completely dynamic added by user as we does in Style Configurator?
image

donho:
@alankilborn 
> What value are the SEPARATORS

The 1st part is about file operation which corresponds to the main menu "File" (Close, Save, Open, Delete, Print...).
The 2nd part is the OTHER operations and the 3rd part is the new added feature. But well, indeed, we can merge the 2nd & the 3rd parts in one section.
Note that there are only 2 Read-Only related commands, and Read-Only could be frequently used by users, so I prefer to keep both on the 1st level.

alankilborn:
>Read-Only could be frequently used by users

I'm very dubious of that, but as you like.
Thank you for revising an overlong tab context menu as part of the addition of colored tabs.

donho:
@ArkadiuszMichalski 
> Maybe for another PR, but it would be nice to have these colors also for items in the Document List panel.

It'll be harder to do in my opinion (I have no idea how to implement it anyway).
You can always create a PR BTW.

donho:
Hmm... Finally **Close Multiple Tabs** makes more sense than **Close More** .
I will also modify in the main menu.

alankilborn:
@donho said:

>Finally Close Multiple Tabs makes more sense than Close More .  I will also modify in the main menu

If you are making these sorts of changes, consider removing the word "All" from each of the "Close Multiple Tabs" items.

It is awkward because "all" suggests, well, everything.  It is a strong word in a native English speaker's mind.

Better would be:

* **Close Other Tabs** (instead of _Close All BUT This_  from right-click context menu, or _Close All but Active Document_ from _File_ menu -- why different text for the same thing in 2 places?)
* **Close Tabs to the Left**
* **Close Tabs to the Right**
* **Close Saved Tabs** (instead of _Close All Unchanged_)

The use of "unchanged" introduces another term to Notepad++ users; why not just use "saved" as "save" is extremely solid in a user's mind?

This would bring commonality with current naming trends that browser's use and users would be more familiar with; example from Chrome:

image

Also, maybe this is another opportunity to promote #12072 as a really good idea. :-)

donho:
> It is awkward because "all" suggests, well, everything. It is a strong word in a native English speaker's mind.

English is not my native language, but I just checked MS Studio, it uses also "Close All But this".
For the other suggestion, let's see if there are other users want to change them.

Yaron10:
The Close commands are used frequently, and I'd highly recommend not to place them in a sub-menu.
When Firefox did that - many users found it inconvenient.

alankilborn:
>The Close commands are used frequently, and I'd highly recommend not to place them in a sub-menu.

In direct opposition to that assertion, I find that `Close` itself is the only one of the group that is used frequently.  Putting the others in a submenu is best with the benefit that it provides a level of safety (harder to accidentally invoke a "Close All..." with a mouse stutter).

I think with the possibility of some name tweaking, that the "close" options are perfectly situated.

Yaron10:
@alankilborn,

You can disagree, and I respect your opinion.
If you want to comment and expect a response, please don't be rude!

alankilborn:
>please don't be rude!

Confused. Please tell me where in my response I was rude.

Yaron10:
@alankilborn,

I respect and appreciate you. Let's move on. :)

donho:
@Yaron10 
> The Close commands are used frequently, and I'd highly recommend not to place them in a sub-menu.
> When Firefox did that - many users found it inconvenient.

Just check the firefox's tab context - the sub-menus are still there.
But I do understand your point - I'll try to make it customizable completely.
Could you create an issue (FR) for that?

ArkadiuszMichalski:
@donho https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12170

Yaron10:
@donho,

Yes, Firefox did not revert that change.

> But I do understand your point - I'll try to make it customizable completely.

👍 
Thank you.

If you think it's worth the extra work and code. :)

@ArkadiuszMichalski,

👍 
Thank you.

alankilborn:
>I'll try to make it customizable completely.

Personally, I would love this; having `tabContextMenu.xml`

PR #12105, opened 2022-09-06
Add incompatible (unloaded) plugin list in Plugin Admin

### LINKED ISSUE #12069: Inform users which plugins have been blocked
After updating Notepad++ some plugins may become incompatible with the new version and automatically blocked. It comes from: 
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a06b4047
https://github.com/notepad-plus-plus/nppPluginList/commit/2137a2c0cd908ca0c615712dd1d6165645cbde29

The problem is that the user does not receive any information about such situation. Blocked plugins are still in the plugins folder and after updating Notepad++ it may not be clear why they are not working (not loading at all). There have been several report related to blocked NppTextFX.

Possible solutions:
- display such information in Plugins Admin, for example by adding a tab "Blocked" beside of tab "Installed". Or show such plugin also in "Installed" tab but blocked (greayed) with some explanation why it was blocked: this can be added in the description box or add third column with additional information (beside of Version). 
This solution will be convenient for standard users because they will instinctively analyze their plugins in the Plugin Admin window.

- display such information also in Debug Info, for example: 
```
Plugins : 
    mimeTools (2.8) (Blocked: incompatibilie)
    NppConverter (4.4)
    NppExport (0.4)
``` 
This solution will be convenient for people that analyzing reported issues.  Directly from the issue we see why some plugin doesn't work.

Other reports:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11900
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12057
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11932

donho:
The plugins which are not compatible but their newer versions are available are not only on the incompatible list, but also on the Updates list.

Fix #12069

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -317,7 +317,7 @@ int PluginsManager::loadPluginFromPath(const TCHAR *pluginFilePath)
 	}
 }
 
-bool PluginsManager::loadPlugins(const TCHAR* dir, const PluginViewList* pluginUpdateInfoList)
+bool PluginsManager::loadPlugins(const TCHAR* dir, const PluginViewList* pluginUpdateInfoList, PluginViewList* pluginImcompatibleList)
 {
 	if (_isDisabled)
 		return false;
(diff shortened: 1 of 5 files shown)


donho:
Could anyone test the feature of this PR?
No need to have the outdated plugins for testing it.
Use the Debug build binary then use modified `nppPluginList.json` (instead of `nppPluginList.dll`) for simulate the plugins incompatibility.

alankilborn:
I was going to give it a try.
Then I realized that I don't know where I'd obtain `nppPluginList.json`.
Some investigation led me here:
https://github.com/notepad-plus-plus/nppPluginList/tree/master/src
I suppose it is just a renamed `pl.x64.json` file?

rdipardo:
> I suppose it is just a renamed `pl.x64.json` file?

That's correct, assuming your Notepad++ targets x86_64.
The release DLL simply reads from an embedded copy: https://github.com/notepad-plus-plus/nppPluginList/blob/9364747e562a150a11bcf42ab78cf0beca19db80/src/resource.h#L33-L41

ArkadiuszMichalski:
@donho Seems works fine:

image

image

image

donho:
Thank you @ArkadiuszMichalski for the confirmation!

PR #12116, opened 2022-09-07
Added close all confirmation dialog

ChrisS02:
Added the ability to enable a confirmation dialog for closing multiple files. 

#12072


alankilborn:
I didn't test it -- yet -- but in theory a good PR.
I like it for those "mouse stutters" that happen more than I like, that end up closing way more files than I intended.

The user text code in this PR will need modification to incorporate localization.
You can model that after how `DoDeleteOrNot` works in english.xml.

BTW, this isn't really accurate:

>confirmation dialog for closing all open files

"...closing multiple files" is a better description.

But you didn't create the "Close All But..." nomenclature, that was pre-existing, so we'll let it stand. :-)

alankilborn:
>The default is disabled to avoid annoying users who don't care about this feature.

I would argue for On by default.
Safety measures, if they exist, should be enabled by default, to protect users from "loss".
(The loss here isn't actual text data, but rather "session" data -- their set of open tabs.)
If the safety measures are not liked, they can be turned Off.

ChrisS02:
>The user text code in this PR will need modification to incorporate localization.
>You can model that after how DoDeleteOrNot works in english.xml.

I'll look into that. Thanks

alankilborn:
>I didn't test it -- yet -- but in theory a good PR.

I've since tested it and it delivers what it promises. :-)

ChrisS02:
I've gone ahead and enabled it by default and made the necessary updates for localization. Please let me know if you see an issue with it. 


Thanks for the feedback @alankilborn

alankilborn:
>enabled it by default

Not sure the Notepad++ authors will like that or not, but I think you should give it a try to get it in that way -- and now you have.  But I'm just saying this because you may have to back off on it. We'll see.

ArkadiuszMichalski:
@ChrisS02 You can make the same solution as we have for `Save all`. Default is on but from dialog we can also change decision:
image

In tabbed programs, this is exactly what it looks like when we try to close all (e.g. in browsers).

alankilborn:
>I don't want to put more time into it if there is basically no chance of it getting approved.

Truly, there's no way of knowing.

alankilborn:
@ArkadiuszMichalski 's idea from https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12116#issuecomment-1240636087 is a good one, but I'm wondering about the wording on the UI for this case.  If it is merely "Are you sure you want to close all?" then that is totally misleading.

Again, it is "close multiple" not "close all" where "multiple" could mean 0 up to the number of tabs open.

If user has chosen to close 2 files to the right of the current tab, and then he's confronted with a "...close all?" confirmation, he's sure to be afraid of that and back out, puzzled.

alankilborn:
@ArkadiuszMichalski Why did you edit your post to remove this?:

>If the description in Preference is unclear we can be more precise `Enable Close more than one tab confirm dialog`.

Better would be `Confirm closing of Close All ... actions` ?

ArkadiuszMichalski:
> @ArkadiuszMichalski Why did you edit your post to remove this?:
> 
> > If the description in Preference is unclear we can be more precise `Enable Close more than one tab confirm dialog`.
> 
> Better would be `Confirm closing of Close All ... actions` ?

Because `Close more than one tab` does not match to `Close all` operation.  I'm not native english so I'm not the best person to discuss about translation. 

But if we have `Enable Save all confirm dialog` then `Enable Close all confirm dialog` looks rationally. In other site we have `Confirm Replace All...` in `Searching` tab so something similar to your proposition:
> Better would be `Confirm closing of Close All ... actions` ? 

As long as it's consistent and understandable, it doesn't really matter to me what the final characters will be (anyway, everyone will change it individually in their native languages).

image

donho:
The forced binary choice will drive Notepad++ crazy. 

@alankilborn @ArkadiuszMichalski 
If there are still such requests coming, please ping me - I'll see what I can do about it.

alankilborn:
Users can simply use the script provided here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12072#issuecomment-1234431489

Issue #12132, opened 2022-09-08
[Code] Fix -Wreorder warnings


ozone10:
Checking warnings in log from gcc debug builds, e.g. https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/44711787/job/dtswa3n5rp6mxniy?fullLog=true is hard.

Some warnings should be fixed to improve readability, for example [-Wreorder].
With less warnings it's easier to find out potential problematic ones.

PR #12133, opened 2022-09-08
Fix reorder warnings

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12132

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -185,7 +185,7 @@ public:
 struct sessionFileInfo : public Position
 {
 	sessionFileInfo(const TCHAR *fn, const TCHAR *ln, int encoding, bool userReadOnly, const Position& pos, const TCHAR *backupFilePath, FILETIME originalFileLastModifTimestamp, const MapPosition & mapPos) :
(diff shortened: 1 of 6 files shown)


PowerEditor/src/Parameters.h
```diff
@@ -185,7 +185,7 @@ struct MapPosition
 struct sessionFileInfo : public Position
 {
 	sessionFileInfo(const TCHAR *fn, const TCHAR *ln, int encoding, bool userReadOnly, const Position& pos, const TCHAR *backupFilePath, FILETIME originalFileLastModifTimestamp, const MapPosition & mapPos) :
-		_isUserReadOnly(userReadOnly), _encoding(encoding), Position(pos), _originalFileLastModifTimestamp(originalFileLastModifTimestamp), _mapPos(mapPos)
+		Position(pos), _encoding(encoding), _isUserReadOnly(userReadOnly), _originalFileLastModifTimestamp(originalFileLastModifTimestamp), _mapPos(mapPos)
```
Yaron10, line 188: Shouldn't it be `_encoding(encoding), _isUserReadOnly(userReadOnly), Position(pos)`?

ozone10: I tried to maximally respect order in constructor, but member initializers should be in same order as members appear/are declared. And since `Position` is declared earlier it must be "first".

Yaron10: 👍 
Thank you.

PR #12150, opened 2022-09-12
Revamp tab context menu

### LINKED ISSUE #12147: Revamp tab context menu
The context menu of Tab is currently too long & too cluttered:

image

It would be better to reorganize it in another way:

```
----------------------
Close                  
Close More           🢒
Save
Save As...
Open into            🢒
Rename
Reload
Move to Recycle Bin
Print
----------------------
Read Only
Clear Read Only Flag
Copy to Clipboard    🢒
Move Document        🢒
Apply Color to Tab   🢒
----------------------
```

**Close More** contains :
 * Close ALL BUT This
 * Close ALL to the Left
 * Close ALL to the Right
 * Close ALL Unchanged
  
**Open** contains:
* Open Containing Folder in Explorer
* Open Containing Folder in Cmd
* Open Containing Folder as Workspace
* Open in Default Viewer

**Copy to Clipboard** contains:
* Copy Full File Path
* Copy Filename
* Copy Current Dir.

**Move Document** contains:
* Move to Other View
* Clone to Other View
* Move to New Instance
* Open in New Instance

**Apply Color to Tab** contains:
* Apply Color 1
* Apply Color 2
* Apply Color 3
* Apply Color 4
* Apply Color 5
* Remove Color

donho:
Add sub-menu in tab context menu so it will be less cluttered.

Fix #12147


ArkadiuszMichalski:
It looks better now. It's possible that not everyone will be happy to move some items to an additional sub-menu. I wish that some day this context menu will be configurable as the main context menu (via an `.xml` file, even using the same `contextMenu.xml` file).
It is worth adding that now most (if not all) commands from the tab context menu are generally available, so they can be placed in the main context menu (and this menu we can modify as we want).

image

alankilborn:
>It's possible that not everyone will be happy to move some items to an additional sub-menu

Oh, I can almost hear the legions of unhappy users already.  "You've ruined my workflow by making me go through a submenu to get to the command I need".

>It is worth adding that now most (if not all) commands from the tab context menu are generally available, so they can be placed in the main context menu (and this menu we can modify as we want).

And that is what any submenu complainers should be told. :-)

vinsworldcom:
> > It is worth adding that now most (if not all) commands from the tab context menu are generally available, so they can be placed in the main context menu (and this menu we can modify as we want).
> 
> And that is what any submenu complainers should be told. :-)

I'll send them here from the Community forum when asking :-).  Seriously though, I like this improvement.  If I have to use the mouse anyway to activate the menu, I'm not opposed to pointing through a logical submenu that helps organization.

Cheers.

donho:
> It is worth adding that now most (if not all) commands from the tab context menu are generally available, so they can be placed in the main context menu (and this menu we can modify as we want).

You can always create a FR - I promise nothing but I can see at least what can I do about it.

alankilborn:
It appears that not many of my suggestions for the revamped tab context menu were implemented, but my feelings aren't hurt...

Here's a look at the new menu, since no one else provided it:

image

I expanded the last submenu for a reason: As a last ditch effort to get a suggestion of mine in, I'll put forth: 

* The submenu is called _Apply Color to Tab_ but the last item in the menu is in opposition to this: _Remove Color_.  Why wouldn't the submenu text be "Tab Color"? Then "Remove Color" fits better.

* The "Style...Token" menu items use the phraseology "Using Style 1"; why would this not be somewhat similar and be "Use Color 1"?

* Why is it "color" when most of the other references to color in the UI use "colour"?

CennoxX:
When I first saw this PR, I thought it would be difficult to get used to the new menu. But the opposite is true, I now need less time than before to navigate the context menu – before I had to search a lot, now the new menu is very intuitive. I don't want to miss it anymore! Thanks for this improvement @donho.

jmvdkolk98:
> It looks better now. It's possible that not everyone will be happy to move some items to an additional sub-menu. I wish that some day this context menu will be configurable as the main context menu (via an `.xml` file, even using the same `contextMenu.xml` file). It is worth adding that now most (if not all) commands from the tab context menu are generally available, so they can be placed in the main context menu (and this menu we can modify as we want).
> 
> image

I'm afraid I need help implementing this solution. I am really looking forward to have my three most used context menu commands back at the toplevel. I have added this to the contextMenu.xml file, I believe according to instructions:
		<Item id="10001" name="Move to Other View"/>
		<Item CMID="7" name="Copy Full File Path"/>
		<Item CMID="8" name="Copy Filename"/>
I don't see any commands however after restarting Notepad++. What am I doing wrong and where should I go for assistance (assuming that it is not here)?

PR #12159, opened 2022-09-12
Send NPPN_DARKMODECHANGED notification with hwndNpp as hwndFrom

### LINKED ISSUE #12144: hwndFrom in message NPPN_DARKMODECHANGED is not hwndNpp
From the document https://npp-user-manual.org/docs/plugin-communication/#nppn-darkmodechanged it is stated that hwndFrom in message NPPN_DARKMODECHANGED is hwndNpp. However, it is 0 instead. The problem is that in NppBigSwitch.cpp, when this notification is sent it uses lParam as hwndFrom, like this:

```
		case NPPM_INTERNAL_REFRESHDARKMODE:
		{
			refreshDarkMode(static_cast<bool>(wParam));
			// Notify plugins that Dark Mode changed
			SCNotification scnN;
			scnN.nmhdr.code = NPPN_DARKMODECHANGED;
			scnN.nmhdr.hwndFrom = reinterpret_cast<void*>(lParam);
			scnN.nmhdr.idFrom = 0;
			return TRUE;
		}
```

And in the code, most places send NPPM_INTERNAL_REFRESHDARKMODE with 0 as lParam, e.g. Notepad_plus::refreshDarkMode().

The code in NppBigSwitch should be changed to use hwnd as hwndFrom.

blu3mania:
So it matches the documented API info on https://npp-user-manual.org/docs/plugin-communication/#nppn-darkmodechanged. Previously it sends 0 which is of no use to plugin authors.

This fixes #12144

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -205,7 +205,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			// Notify plugins that Dark Mode changed
 			SCNotification scnN;
 			scnN.nmhdr.code = NPPN_DARKMODECHANGED;
-			scnN.nmhdr.hwndFrom = reinterpret_cast<void*>(lParam);
+			scnN.nmhdr.hwndFrom = hwnd;
 			scnN.nmhdr.idFrom = 0;
 			_pluginsManager.notify(&scnN);
 			return TRUE;


donho:
"blu3mania:master" is used repeatedly for several PR.

>    Create a new branch for each PR. Make sure your branch name wasn't used before - you can add date (for example patch3_20200528) to ensure its uniqueness.

Please create a new PR with a new branch. And make sure your branch name is unique. Thank you.

PR #12165, opened 2022-09-13
Add Change History markers for saved/unsaved/undone modification

### LINKED ISSUE #12164: Add new feature: Change History Margin
It'll be great that there's a margin for change history which marks saved/unsaved modification and undo states, so user can follow the changes of editing document.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12045#issuecomment-1229195785

donho:
Implement Change History by using Scintilla v5.3.0 new abilities.

Fix  #12164


Yaron10:
@donho,

That's a useful addition.
Thank you for your work. 👍 

1. Bookmarks and the Changed markers use the same margin. The alternative requires some thinking.
2. An option to disable this feature would be nice.
3. The empty rectangle for unsaved lines vs. the more prominent full rectangle for saved ones is interesting.
I suppose this is Scintilla.

vinsworldcom:
@Yaron10 

> 1. Bookmarks and the Changed markers use the same margin. The alternative requires some thinking.

Agree. I've been putting in a new margin for my plugin, but seeing it in the same is not that bad, see discussion and image here:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12164#issuecomment-1245597825

> 2. An option to disable this feature would be nice.

Agree.  Thoughts are again:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12164#issuecomment-1245597825

> 3. The empty rectangle for unsaved lines vs. the more prominent full rectangle for saved ones is interesting.
I suppose this is Scintilla.

Yes, this is Scintilla and not sure why.  I just added the "fill" in a PythonScript:

`editor.markerSetBack(23, (255, 128, 0))`

Cheers.

Yaron10:
@vinsworldcom,

I didn't see your comment when I posted mine. :)

The question is what should be better: the current implementation or adding a new narrow (6px ?) margin.
I tend to prefer the second option.

> An option to disable this feature would be nice.
>> Agree. Thoughts are again:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12164#issuecomment-1245597825

👍 

> The empty rectangle for unsaved lines vs. the more prominent full rectangle for saved ones is interesting.
        I suppose this is Scintilla.
>> Yes, this is Scintilla and not sure why. I just added the "fill" in a PythonScript:
editor.markerSetBack(23, (255, 128, 0))

Great! As long as the shape (and colors?) are not configurable, I'd suggest to add your line to NPP's code.

alankilborn:
>The question is what should be better: the current implementation or adding a new narrow (6px ?) margin.

Could be made configurable?  Radio buttons?:
* Share bookmark margin
* Dedicated margin

Yaron10:
> Could be made configurable?

Wouldn't that be too much?
Besides, the shared margin does not look professional.
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12164#issuecomment-1245597825

> Remember that the hidden lines markers already share the bookmark margin.
And, THAT effect isn't the greatest; example (the margin is shared with 2 symbols on line 14):

Good point.

@vinsworldcom,

I've just tried your ChangedLines plugin.
Great work. I should have tried it earlier.
You got a new star. :)

vinsworldcom:
Notepad++ has (by default) 5 margins - verify with `SCI_SENDMSG SGI_GETMARGINS` or `editor.getMargins()` numbered 0 - 4.

0 = line numbers
1 = symbols (bookmarks, folding and new change history)
2 = folding tree symbols
3 = unused
4 = unused

I know the ComparePlus plugin uses margin 4 for its symbols.  The Change History feature could use margin 3 as I don't know of any other plugins that use this.

Cheers.

vinsworldcom:
> I should have tried it earlier.

Or just waited a little longer for the native Change History!  I already uninstalled my Changed Lines as I'm using the native Change History activated with a PythonScript.  The only thing I miss is the "next / previous position" I baked into Changed Lines.

Cheers.

Yaron10:
@vinsworldcom,

> The only thing I miss is the "next / previous position" I baked into Changed Lines.

I suppose you'll get it in NPP at some point.

donho:
Thank you for your suggestion. The PR is still in progress.

>1.    Bookmarks and the Changed markers use the same margin. The alternative requires some thinking.
    
It's now separated. One margin has been added only for Change History Markers (as @vinsworldcom has suggested).

> 2. An option to disable this feature would be nice.

I would like launch Notepad++ v8.4.6 with this feature as only choice, to see the reaction of audience. We can always add a new option to disable it in the future version...

> 3.   The empty rectangle for unsaved lines vs. the more prominent full rectangle for saved ones is interesting.
    I suppose this is Scintilla.

It's done voluntarily in Scintilla, for the unknown reason: 
https://www.scintilla.org/ScintillaDoc.html#ChangeHistory

> Yes, this is Scintilla and not sure why. I just added the "fill" in a PythonScript:

If you have the C/C++ solution, please let me know.

alankilborn:
>I would like launch Notepad++ v8.4.6 with this feature as only choice

Nice -- I like this idea, personally.
However, Scintilla documentation says this feature can use up a lot of memory.
Until we know the implications from that, it might make sense to actually include the enable/disable from the start?
It can certainly be enabled by default, though, to gauge user reaction.

vinsworldcom:
@donho 

> If you have the C/C++ solution, please let me know.

I think it would just be:

```
SendMessage( hCurrentScintillaHandle, SCI_MARKERSETBACK, SC_MARKNUM_HISTORY_MODIFIED, 33023);
```

Where `hCurrentScintillaHandle` is the handle to the current Scintilla instance.

Where 33023 is the integer equivalent of the 0x00BBGGRR ([COLORREF](https://docs.microsoft.com/en-us/windows/win32/gdi/colorref)) for Red = 255, Blue = 128, Green = 0 (255, 128, 0) = (orange).  I got that value from a screengrab of Notepad++ using a color picker on the Scintilla change marker outline. 

Cheers.

vinsworldcom:
Looks good!

image

alankilborn:
> Looks good!

It does!
I might be tempted to use PS to make the change margin even narrower, like I currently have things set with _ChangedLines_.  :-)

Yaron10:
@donho,

Great! 👍 
Thanks again for your work.

I'd like to second @alankilborn's comments.

1. How about a simple enable/disable margin (just like the prefs for the other margins)?
2. The margins are "expensive real estate". Setting the history-margin's width to `10` - the image itself is `3px`, and you get `3px` on one side + `4px` on the other. 
This margin is *not clickable*, and `2px` on each side would suffice IMO.
If you use `width = dpiManager.scaleX(7);` - is the image itself still `3px`? Is it centered?

***

How about enabling the user to customize the background color?
ComparePlus uses the default color, and the current result is not ideal:

תמונה

I'm sure @pnedev would be kind enough to add customization for the ComparePlus' margin background color.
But NPP already provides this option for the other margins.

***

@All,

Do you prefer the current history-margin's position on the right next to the editor?
Isn't the fold-margin proximity to the text "more important"?
Would you like to see a demo with the history-margin's positioned after the lines-margin?

alankilborn:
Yaron10 said:

>I'd like to second @alankilborn's comments.
>How about a simple enable/disable margin

I think I meant my suggestion to go beyond the margin and have an enable/disable for the _entire feature_ right from the start.  Maybe I'm over-careful but I have performance concerns.  Maybe they will prove to be unfounded.

>Do you prefer the current history-margin's position on the right next to the editor?
Isn't the fold-margin proximity to the text "more important"?

I currently use _ChangedLines_ this way:

image

so maybe I prefer it that way? :-)
Or I've never been given a choice. :-)
With some thought, I think it is important to be as close to the text as possible.

Yaron10:
@alankilborn,

> I think I meant my suggestion to go beyond the margin and have an enable/disable for the entire feature right from the start.
Maybe I'm over-careful but I have performance concerns.

I see. This is beyond my level. I'll leave that to you and Don. :)

> With some thought, I think it is important to be as close to the text as possible.

I can think of other pros and cons for either position.
As you and the others prefer.

vinsworldcom:
@Yaron10  - good points, I'll try to address some with my thoughts:

> How about a simple enable/disable margin (just like the prefs for the other margins)?

I think we could get this for "free" with the enable / disable feature.  If the feature is disabled, it could automatically hide the margin (set width = 0) since the new use of margin 3 isn't used for anything else.  So enable / disable would also show / hide the new margin 3.

> The margins are "expensive real estate". Setting the history-margin's width to 10 - the image itself is 3px, and you get 3px on one side + 4px on the other.

I did a little playing with PythonScript and it seems the colored bar scales with the margin so if you make the margin much thinner, the bar ends up only being a pixel wide.

Width = 4 on left, Width = 50 on right:

image

I think 10 is a nice compromise.  This makes it big enough to click (`SCN_MARGINCLICK`) - which I use in ChangedLines to move to the next change.  I think that would be a nice feature add once this is incorporated into Notepad++ - a click / CTRL+Click or Shift+Click in the change margin to navigate to the "next" or "previous" change up or down the document.

> How about enabling the user to customize the background color?

We already can on the line number, bookmark and fold margins via Style Configurator.  I would hope this could be added there as well.

Cheers.

donho:
@vinsworldcom 
> I think it would just be:
SendMessage( hCurrentScintillaHandle, SCI_MARKERSETBACK, SC_MARKNUM_HISTORY_MODIFIED, 33023);

Works like charm! Thank you! (I guess it's needed to be configurable for users)

@alankilborn 
> However, Scintilla documentation says this feature can use up a lot of memory.

Yes. I was thinking of removing it for the big files, as not applying syntax highlighting for big files.

@Yaron10 
> Setting the history-margin's width to 10 - the image itself is 3px, and you get 3px on one side + 4px on the other.
If you use width = dpiManager.scaleX(7); - is the image itself still 3px? Is it centered?

As @vinsworldcom has pointed out, the width of markers will be centered in the margin, and it's shrunk with margin size.
I guess we have to also provide the option for resize Change History Margin.

The reason that I want to impose Change History Margin (ie. cannot disable it) is because I found a problem during playing around this new feature:
When a document is cloned beside, and it's been modified, enabling Change History will unsynchronize cloned document (but no problem if the document is cloned AFTER enabling Change History). 
However, it's necessary to have an option to disable/hide Change History Margin anyway, I think we are all in the same page.

I'll keep working on this feature. Thank you for all your suggestions.

alankilborn:
@vinsworldcom said:

> the colored bar scales with the margin so if you make the margin much thinner, the bar ends up only being a pixel wide

Not that you have any great insights into the minds of the Scintilla devs, but do you think there is a valid reason for this behavior?  I mean, if you're going to dedicate a certain width to a margin for putting something in it, why not use the maximum width possible...

vinsworldcom:
@alankilborn 

> Not that you have any great insights into the minds of the Scintilla devs, but do you think there is a valid reason for this behavior? I mean, if you're going to dedicate a certain width to a margin for putting something in it, why not use the maximum width possible...

They already have marker for that "maximum width" type = `SC_MARK_FULLRECT` (https://www.scintilla.org/ScintillaDoc.html#Markers).  I use this as default in ChangedLines plugin.

I think maybe the `SC_MARK_BAR` was added specifically for this Change History feature?

Cheers.

Yaron10:
@vinsworldcom,

Thank you for providing the C++ code for the "full rectangle". 👍 

> I think we could get this for "free" with the enable / disable feature. If the feature is disabled, it could automatically hide the margin (set width = 0) since the new use of margin 3 isn't used for anything else. So enable / disable would also show / hide the new margin 3.

👍 

> I did a little playing with PythonScript and it seems the colored bar scales with the margin so if you make the margin much thinner, the bar ends up only being a pixel wide.

Thank you for testing that.
Could you please post the Python code for changing the width?
I'd like the image to be centered within the margin. So perhaps 11 would be better than 10.

>  I think that would be a nice feature add once this is incorporated into Notepad++ - a click / CTRL+Click or Shift+Click in the change margin to navigate to the "next" or "previous" change up or down the document.

That would be a nice bonus.
And if the user clicks on the current changed line - no jumping, correct?

***

@donho,

Thank you for adapting the C++ code for the "full rectangle". 👍 

> (I guess it's needed to be configurable for users)

Do you mean "empty rectangle" vs. "full rectangle"?
Wouldn't that be too much? - I did not mean to suggest that in a previous comment.

> Yes. I was thinking of removing it for the big files, as not applying syntax highlighting for big files.

👍 

> As @vinsworldcom has pointed out, the width of markers will be centered in the margin

I'm afraid we have some misunderstanding here.
It would be aesthetically nicer if the image was *properly centered*: i.e. have an even number of pixels _on both sides_.
That's why I want to try 9 or 11.

> The reason that I want to impose Change History Margin...

Nice catch.  

> However, it's necessary to have an option to disable/hide Change History Margin anyway, I think we are all in the same page.

👍
Thanks again for you work.

***

@alankilborn,

Allow me to bring up the margin position again.

We actually have a demo of the history-margin positioned after the lines-margin. - Don's first version here.
IMO it's better. Would you give it a try?

1. The history-margin is "line-based" whereas the fold-margin is "block-based".
We do have a lines-margin. Let's place the history-margin next to it.
The text is the "block-margin" ( :) ). Let's place the fold-margin next to it.

תמונה

I think this is much clearer.

2. Users are used to having the fold-margin next to the text.
Why change that and cause some confusion?

@donho & @vinsworldcom,

What do you think?

***

@alankilborn,

> Remember that the hidden lines markers already share the bookmark margin.
And, THAT effect isn't the greatest; example (the margin is shared with 2 symbols on line 14):

Would you please open an issue for that?
What do you think of a dedicated hidden-lines margin displayed only if the document has some hidden lines?

Thank you.

vinsworldcom:
@Yaron10 

> > I did a little playing with PythonScript and it seems the colored bar scales with the margin so if you make the margin much thinner, the bar ends up only being a pixel wide.
> 
> Thank you for testing that. Could you please post the Python code for changing the width? I'd like the image to be centered within the margin. So perhaps 11 would be better than 10.

If that's the case, perhaps 9 instead of 11, let's not make it wider :-)

```
margin = 3
width = 9
editor.setMarginWidthN( margin, width )
```
 
> > I think that would be a nice feature add once this is incorporated into Notepad++ - a click / CTRL+Click or Shift+Click in the change margin to navigate to the "next" or "previous" change up or down the document.
> 
> That would be a nice bonus. And if the user clicks on the current changed line - no jumping, correct?

Yes - that's how ChangedLines works.  If you're on the line with the only change, no jumping.  If you're on any line and there are no changes at all, no jumping.  Otherwise find the "next" or "previous" change with each successive click, wrapping back to the top or bottom respectively and continuing.

I think you said you installed ChangedLines to see the marker effect, give the margin clicking a shot to see it in action!  It's not something I use - rarely if ever - just thought it was a nice addition since it was possible.  It certainly can be added to Notepad++ for the native Change History as a pull request for a later version once Change History is incorporated into the base code.

With respect to location of margins, since I've been using ChangedLines for this similar feature for years now and I've always had it in the margin closest to the text, so I'm happy with the new margin 3 exclusive usage for this.  It looks the most similar to what I'm accustomed to.  I'm sure we'll hear a lot of differing opinions once this is released, and then have to find some consensus for the majority of users.

To be honest, I'm not that worried since I've learned PythonScript.  So long as the Scintilla version is merged, I _could_ use PythonScript to disable the Notepad++ implementation and customize it how I like. Granted, the majority of users will not do this, hence my previous comment about "consensus for the majority".

I'm just so happy Scintilla made this native and now Notepad++ has merged the support!  Can't wait for 8.4.6 to see what we get and how users respond.

Cheers.

Yaron10:
@vinsworldcom,

Thank you very much.

With width = 9, the bar itself is still 3px and you have 3px on both sides.

תמונה

I must take a break now. I'll respond to the other points later.

alankilborn:
>CTRL+Click or Shift+Click in the change margin to navigate to the "next" or "previous" change up or down the document

A nice feature that I have thus far overlooked in the ChangedLines plugin.  :-(
Also an argument for a decently wide margin for the native implementation (my margin in ChangedLines is so narrow that it is difficult to get the click to fall on it every time).

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -233,6 +233,7 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 	execute(SCI_SETMARGINMASKN, _SC_MARGE_SYMBOL, (1 << MARK_BOOKMARK) | (1 << MARK_HIDELINESBEGIN) | (1 << MARK_HIDELINESEND) | (1 << MARK_HIDELINESUNDERLINE));
 
 	execute(SCI_SETMARGINMASKN, _SC_MARGE_CHANGEHISTORY, (1 << SC_MARKNUM_HISTORY_REVERTED_TO_ORIGIN) | (1 << SC_MARKNUM_HISTORY_SAVED) | (1 << SC_MARKNUM_HISTORY_MODIFIED) | (1 << SC_MARKNUM_HISTORY_REVERTED_TO_MODIFIED));
+	execute(SCI_MARKERSETBACK, SC_MARKNUM_HISTORY_MODIFIED, RGB(255, 0, 0));
```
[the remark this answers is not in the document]
Yaron10: @vinsworldcom,

With Don's `RGB(255, 0, 0))`, you get an *orange* "border" filled with 1px red.
I think it's nice. :)

vinsworldcom:
@Yaron10 

width = 9 seems like the way to go!

==

A picture is worth a thousand words:

image

I just had the background color of the fold margin set to some dark color, but when I matched it to my editor background the fold margin blends in with the editor and gives it this cool effect where the margin is almost invisible and the folding is part of the document.  **I can now see why you want to switch the fold and change margins** from:

fold=2
change =3
 
to:

change = 2
fold = 3

I tried compiling the pull request with the simple change of:

notepad-plus-plus\PowerEditor\src\ScintillaComponent\ScintillaEditView.cpp
```
35: const int ScintillaEditView::_SC_MARGE_LINENUMBER = 0;
36: const int ScintillaEditView::_SC_MARGE_SYMBOL = 1;
37: const int ScintillaEditView::_SC_MARGE_CHANGEHISTORY = 2;
38: const int ScintillaEditView::_SC_MARGE_FOLDER = 3;
```

and I get the margins re-ordered, but no change indicators appear in margin 2 (or anywhere for that matter) when I edit.  I must have missed some other stuff involved in switching margins?

Cheers.

ozone10:
@vinsworldcom 
looks great 👍

vinsworldcom:
@Yaron10 @donho 

> 1. The history-margin is "line-based" whereas the fold-margin is "block-based".
>    We do have a lines-margin. Let's place the history-margin next to it.
>    The text is the "block-margin" ( :) ). Let's place the fold-margin next to it.

After using my [compiled version](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12165#issuecomment-1248078815) with change history margin = 2 and fold margin = 3 all day, it's taken me only the day to conclude **_this is the way to go_**!  It's just so much more cleaner with my colors matched now.  And on the default stylers where the background is just white, the effect will be the same.

I think we should make the simple switch:

```diff
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -34,8 +34,8 @@ UserDefineDialog ScintillaEditView::_userDefineDlg;

 const int ScintillaEditView::_SC_MARGE_LINENUMBER = 0;
 const int ScintillaEditView::_SC_MARGE_SYMBOL = 1;
-const int ScintillaEditView::_SC_MARGE_FOLDER = 2;
-const int ScintillaEditView::_SC_MARGE_CHANGEHISTORY = 3;
+const int ScintillaEditView::_SC_MARGE_FOLDER = 3;
+const int ScintillaEditView::_SC_MARGE_CHANGEHISTORY = 2;

 WNDPROC ScintillaEditView::_scintillaDefaultProc = NULL;
 string ScintillaEditView::_defaultCharList = "";
```

Cheers.

Yaron10:
@vinsworldcom,

Thanks again for your work. 👍 

> width = 9 seems like the way to go!

Great.
And I think we all agree that if and when your "Next/Prev" feature is implemented, the width should be increased (as @alankilborn has pointed out).

I have tried the "Next/Prev" functionality in your plugin.
It's indeed very useful.
I have some questions/suggestions, and I'll bring them up in the relevant PR. :)

I'll allow myself to quote you.
"Not to nitpick, but" do you prefer the "lines-margin, bookmarks-margin, history-margin" order?

If I understand it correctly, your approach is based on the visual flow (which is certainly beautiful).
But other users would surely use completely different themes/background colors.

Both bookmarks-margin and history-margin are "line-based".
But the new history-margin would probably be more dominant for most users.
And if we don't place it next to the text, shouldn't it be next to the lines-margin?
(So that the order is: "lines-margin, history-margin, bookmarks-margin, fold-margin").

I don't have Visual Studio on the machine I'm using now.
Where is the bookmarks-margin positioned there?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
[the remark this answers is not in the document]
Yaron10: Seeing @vinsworldcom's images, the "border" is much more dominant on a darker background.
`(255,128,0)` would probably be better.

vinsworldcom:
> Where is the bookmarks-margin positioned there?

Before the line number margin:

image

PowerEditor/src/WinControls/ColourPicker/WordStyleDlg.cpp
```diff
@@ -524,7 +524,7 @@ intptr_t CALLBACK WordStyleDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM
 							case CPN_COLOURPICKED:
 							{
 								if (reinterpret_cast<HWND>(lParam) == _pFgColour->getHSelf())
-								{
+	sdds							{
 									updateColour(C_FOREGROUND);
 									notifyDataModified();
```
ArkadiuszMichalski, line 529: @donho Typo `sdds`?

donho: Just removed

donho:
PR current status: Add an option to disable/enable Change History:

image

**Please test this PR - It seems the feature is not mature - I've got several crash during the test.**

>  I vote for renumbering the margins as I did above - simple change - massive effect! I think it looks stunning and it will "feel" the same for most users with an extra margin inserted rather than "added" on the end.
Additionally, the re-order is consistent with Visual Studio for example:

I just want to show another possibility with the latest commit, then you guys tell me if you like or not:

image


Note that if you change the theme, recheck the option to apply the bg change.

donho:
Crash reproducible with the following step:
1. Open a document with Change History enabled.
2. Selection 1 character and type another character to replace it, then save the document.
3. Disable Change History and enable it again.
4. Ctrl-Z
5. Crash!

This feature is not ready obviously - it won't be in v8.5.6 for the sake of stability.

AndiKiissel:
https://www.scintilla.org/ScintillaDoc.html#ChangeHistory
...
Change history depends on the undo history and can only be enabled when undo history is enabled and empty. It should be enabled once when a file is loaded after calling SCI_SETUNDOCOLLECTION(true) and SCI_SETSAVEPOINT.
...

Yaron10:
@vinsworldcom,

> Where is the bookmarks-margin positioned there?
>> Before the line number margin:

Their concept might be: 
Both bookmarks and history should be next to the lines number.
Let's place them there - one on the left and the other on the right.

@donho,

Thanks again for your work.

> Crash reproducible with the following step:

👍 

**NOTE:** If you use the menu/toolbar Undo, the crash does NOT happen.
But the Undo/Redo buttons might be incorrectly enabled.

Another issue:
Start NPP with Change History enabled.
Edit a document.
Disable Change History.
Switch to another document and edit.

Result:
The current line background color is red.

***

> I just want to show another possibility with the latest commit, then you guys tell me if you like or not:

I still prefer the "lines-margin, history-margin, bookmarks-margin, fold-margin" order for the various reasons I've listed above.

alankilborn:
My preference is that I want the change history closest to the text as possible, but I will admit that the fold margin right next to the text has great appeal when setting the margin background color the same as the editing text background color, so for me:

linenumber-bookmark-change-fold

It might be too radical of a change to N++ for an existing feature, but perhaps bookmark placement like Visual Studio makes sense, so if this is considered I would change my preference to:

bookmark-linenumber-change-fold

vinsworldcom:
@AndiKiissel 

> https://www.scintilla.org/ScintillaDoc.html#ChangeHistory
> ...
> Change history depends on the undo history and can only be enabled when undo history is enabled and empty. It should be enabled once when a file is loaded after calling SCI_SETUNDOCOLLECTION(true) and SCI_SETSAVEPOINT.
> ...

Interesting.  I tried that with PythonScript (note I have written subs to enable / disable change history since PythonScript does not yet support this feature), so @donho  step 3 above becomes:

```Python
ch.setChangeHistory(0)
editor.setUndoCollection(True)
editor.setSavePoint()
ch.setChangeHistory(3)
```

and I still get the crash. 

I won't be enabling / disabling this feature probably at all - just keep it on ; however, as @donho mentioned, we may want to disable for large documents like we disable lexing for performance reasons and then re-enabling for a smaller doc would potentially cause this crash?

I'm going to keep using the PR artifact (slightly modified to put the margins:  NUM | BOOK | CHANGE | FOLD and with change margin 9 pixels) and see how it goes.  I'll report any ill effects - though pretty heavy usage yesterday and no issues at all!

Cheers.

Yaron10:
Another option for the order:
Let's follow Visual Studio's concept and tweak it.

Place the history-margin **before** the lines-margin.

The lines-margin's width is min 3 digits. If you use constant width, the min width might be much larger.
On opening a document, the users are accustomed to seeing some empty space before the number.

תמונה


So we just slightly increase the space before the line numbers.

What do you think?

alankilborn:
Perhaps so much debate about "best" margin ordering among a limited sample of 3/4 people means that the "best" option would be to allow end-users to configure any ordering they want. :-)

Of course even with that, there would be a "default" -- and what is best for that. Sigh. :-)

Yaron10:
> Perhaps so much debate about "best" margin ordering among a limited sample of 3/4 people means that the "best" option would be to allow end-users to configure any ordering they want. :-)

I was just thinking that. :)

vinsworldcom:
> ```python
> ch.setChangeHistory(0)
> editor.setUndoCollection(True)
> editor.setSavePoint()
> ch.setChangeHistory(3)
> ```

@AndiKiissel @donho  - change the above to:

> ```python
> ch.setChangeHistory(0)
> editor.emptyUndoBuffer
> ch.setChangeHistory(3)
> ```

and no crash!! :-)  Not ideal, forgetting the undo history, but if you're going to enable Change History, you need to start with a clean undo slate as per Scintilla docs.  I also tried with:

> ```python
> ch.setChangeHistory(0)
> editor.emptyUndoBuffer
> editor.setUndoCollection(True)
> editor.setSavePoint()
> ch.setChangeHistory(3)
> ```

and no crash.  Could be a "simple" fix - just adding the [`SCI_EMPTYUNDOBUFFER`](https://www.scintilla.org/ScintillaDoc.html#SCI_EMPTYUNDOBUFFER), [`SCI_SETUNDOCOLLECTION`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETUNDOCOLLECTION), and [`SCI_SETSAVEPOINT`](https://www.scintilla.org/ScintillaDoc.html#SCI_SETSAVEPOINT) calls just before enabling Change History [(`SCI_SETCHANGEHISTORY`)](https://www.scintilla.org/ScintillaDoc.html#SCI_SETCHANGEHISTORY).

Cheers.

alankilborn:
>Could be a "simple" fix - just adding the [SCI_EMPTYUNDOBUFFER](https://www.scintilla.org/ScintillaDoc.html#SCI_EMPTYUNDOBUFFER), [SCI_SETUNDOCOLLECTION](https://www.scintilla.org/ScintillaDoc.html#SCI_SETUNDOCOLLECTION), and [SCI_SETSAVEPOINT](https://www.scintilla.org/ScintillaDoc.html#SCI_SETSAVEPOINT) calls just before enabling Change History [(SCI_SETCHANGEHISTORY)](https://www.scintilla.org/ScintillaDoc.html#SCI_SETCHANGEHISTORY).

I'd imagine this warrants a change on the UI, possible example:
* _ Enable Change History (Warning: Clears undo buffer, if enabling)
or even:
* _ Enable Change History (Must restart to take effect, if enabling)

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -229,7 +226,17 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 	execute(SCI_SETMARGINMASKN, _SC_MARGE_FOLDER, SC_MASK_FOLDERS);
 	showMargin(_SC_MARGE_FOLDER, true);
 
-	execute(SCI_SETMARGINMASKN, _SC_MARGE_SYBOLE, (1<<MARK_BOOKMARK) | (1<<MARK_HIDELINESBEGIN) | (1<<MARK_HIDELINESEND) | (1<<MARK_HIDELINESUNDERLINE));
+	execute(SCI_SETMARGINMASKN, _SC_MARGE_SYMBOL, (1 << MARK_BOOKMARK) | (1 << MARK_HIDELINESBEGIN) | (1 << MARK_HIDELINESEND) | (1 << MARK_HIDELINESUNDERLINE));
+
+	execute(SCI_SETMARGINMASKN, _SC_MARGE_CHANGEHISTORY, (1 << SC_MARKNUM_HISTORY_REVERTED_TO_ORIGIN) | (1 << SC_MARKNUM_HISTORY_SAVED) | (1 << SC_MARKNUM_HISTORY_MODIFIED) | (1 << SC_MARKNUM_HISTORY_REVERTED_TO_MODIFIED));
```
vinsworldcom, line 231: @donho you need to clear the mask from the margin when hiding the margin otherwise the marks once in the margin are show as background in the text.  The same goes for bookmarks.

1) Set a bookmark
2) Settings=>Preferences=>Margins/Border/Edge=> uncheck "Display bookmark"

watch the line background turn the color of the bookmark.  With Change History, it's not immediately noticeable, but when you open a new doc and start typing, you see the background. 

These lines are probably OK here, but you'll need to add:

```C++
execute(SCI_SETMARGINMASKN, _SC_MARGE_SYMBOL, 0)
execute(SCI_SETMARGINMASKN, _SC_MARGE_CHANGEHISTORY, 0)
```

when disabling / hiding the margins.

vinsworldcom, line 231: @donho you need to clear the mask from the margin when hiding the margin otherwise the marks once in the margin are show as background in the text.  The same goes for bookmarks.

1) Set a bookmark
2) Settings=>Preferences=>Margins/Border/Edge=> uncheck "Display bookmark"

watch the line background turn the color of the bookmark.  With Change History, it's not immediately noticeable, but when you open a new doc and start typing, you see the background. 

These lines are probably OK here, but you'll need to add:

```C++
execute(SCI_SETMARGINMASKN, _SC_MARGE_SYMBOL, 0);
execute(SCI_SETMARGINMASKN, _SC_MARGE_CHANGEHISTORY, 0);
```

when disabling / hiding the margins.

Yaron10:
>  Set a bookmark
    Settings=>Preferences=>Margins/Border/Edge=> uncheck "Display bookmark"
>
> watch the line background turn the color of the bookmark.

I can't reproduce that.

vinsworldcom:
> I can't reproduce that.

Sorry, use a dark theme - I use one like Obsidian (customized).  You won't see it on the Default white background, but it's there.

image

image

image

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2000,7 +2008,7 @@ void ScintillaEditView::activateBuffer(BufferID buffer, bool force)
 	runMarkers(true, 0, true, false);
 
 	setCRLF();
-
+	execute(SCI_SETCHANGEHISTORY, SC_CHANGE_HISTORY_ENABLED | SC_CHANGE_HISTORY_MARKERS);
```
vinsworldcom, line 2011: @donho I don't think you want to enable it always here, I think you want to set it to the value of on/off depending on the UI:

```C++
	const ScintillaViewParams& svp = nppParam.getSVP();
	int enabledCH = svp._isChangeHistoryEnabled ? (SC_CHANGE_HISTORY_ENABLED | SC_CHANGE_HISTORY_MARKERS) : SC_CHANGE_HISTORY_DISABLED;
	execute(SCI_SETCHANGEHISTORY, enabledCH);

```

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
AndiKiissel: > We could probably use:
> 
> [`SCI_MARKERDELETEALL`](https://www.scintilla.org/ScintillaDoc.html#SCI_MARKERDELETEALL)
> 
> to remove the markers from the margin before hiding it. Maybe ignore it for the Bookmarks, that's a separate (but similar) issue.

Please do not delete Bookmarks when `Display bookmark` is switched off!
BTW, I know and use and like this "bug" for some time...

alankilborn: > do not delete Bookmarks when `Display bookmark` is switched off!

Is that current behavior for you @AndiKiissel ?
It is NOT my experience.

Yaron10:
@vinsworldcom,

I still can't reproduce the bookmarks issue.

@alankilborn,

Can you?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
AndiKiissel: Yes, 2. is bookmarked, 4. is active line:
image
Visible and not so intensive as default blue marker.
I move between them with `F2`/`Shift+F2`.

vinsworldcom:
> @vinsworldcom,
> 
> I still can't reproduce the bookmarks issue.
> 
> @alankilborn,
> 
> Can you?

Fresh install of Noteapad++ 64-bit 8.4.5.

1) open Notepad++
2) Settings => Style Configurator => Select Theme => Obsidian
3) Press Enter a few times to get some lines
4) Click in the bookmark margin to set some bookmarks on random lines
5) Settings => Preferences => Margins/Border/Edge => uncheck "Display bookmark"
6) Observe background color of lines with bookmarks has now changed.

Cheers.

Yaron10:
@vinsworldcom,

I can reproduce it now. :)
Nice catch. 👍 

Thank you for that and for the rest of your work here.

alankilborn:
> 6\. Observe background color of lines with bookmarks has now changed.

@vinsworldcom 

Making an independent new issue about that?

vinsworldcom:
> > 6. Observe background color of lines with bookmarks has now changed.
> 
> @vinsworldcom
> 
> Making an independent new issue about that?

We could?  @AndiKiissel calls it a "bug" but I too see how it's useful in that you can still get / set bookmarks without the margin showing.  They just show on the lines.  The only "fix" is to delete all bookmarks before hiding the margin so they won't show line highlights.  However, then you lose all bookmarks.  I tend to think this on is OK.

I do think it will be a problem with Change History though.  That's why I proposed no switch to hide the change history margin.  Rather turn on or of change.  Turning off would disable, clear all marks and then hide the margin.  I don't think we want to have a "hide" change history margin without actually disabling the feature.

Cheers.

donho:
I don't really want to deal with empty undo/redo buffer, cleaning the markers etc...
So the latest commit comes with my lazy (but cleaner) solution: this option can be disabled immediately, but if enable it, user has to restart Notepad++.
For the disabling for the big files, I have even no idea how to manage it. Let's see if the current implementation is stable, then we might see the next step.

For the margin order, it's set as the most people want: NUM | BOOK | CHANGE | FOLD. I heard the need to reorder them as people want, I will see what I can do about it.

Please let me know if anyone has crash.
Thank you everybody for your testing, useful tips and pertinent suggestions!

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -1732,9 +1732,21 @@ intptr_t CALLBACK MarginsBorderEdgeSubDlg::run_dlgProc(UINT message, WPARAM wPar
 					return TRUE;
 
 				case IDC_CHECK_CHANGHISTORYMARGE:
-					svp._isChangeHistoryEnabled = (BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_CHECK_CHANGHISTORYMARGE, BM_GETCHECK, 0, 0));
-					::SendMessage(::GetParent(_hParent), NPPM_INTERNAL_ENABLECHANGEHISTORY, 0, 0);
+				{
+					bool isChangeHistoryEnabled = (BST_CHECKED == ::SendDlgItemMessage(_hSelf, IDC_CHECK_CHANGHISTORYMARGE, BM_GETCHECK, 0, 0));
+					if (isChangeHistoryEnabled)
+					{
+						printStr(L"You have to restart Notepad++ to enable Change History.");
```
ArkadiuszMichalski, line 1739: @donho Is's worth adding a translation for this message.

Yaron10:
@donho,

> I don't really want to deal with empty undo/redo buffer, cleaning the markers etc...
So the latest commit comes with my lazy (but cleaner) solution: this option can be disabled immediately, but if enable it, user has to restart Notepad++.

👍 

> For the margin order, it's set as the most people want: NUM | BOOK | CHANGE | FOLD.

👍 
It looks great.

> I heard the need to reorder them as people want, I will see what I can do about it.

I appreciate that.
But IMO it would be too much, and I think @alankilborn was just joking.

Thank you (and all) for your work.

***

Remaining tasks:
1. Disabling in big files.
2. Customizing the margin background color.
3. `execute(SCI_MARKERSETBACK, SC_MARKNUM_HISTORY_MODIFIED, RGB(255, 0, 0));` to `(255,128,0)` as @vinsworldcom suggested.
4. Localizing the "You have to restart Notepad++..." message as @ArkadiuszMichalski suggested.

Anything else?

alankilborn:
>>I heard the need to reorder them as people want, I will see what I can do about it.

>... I think @alankilborn was just joking.

I wasn't joking -- hopefully I make it very clear when I am joking -- don't intend to make people here try to figure that out.

If custom reordering is too large of an effort, then it certainly makes sense to not do it.

>Thank you (and all) for your work.

Completely second that -- not joking. :-)

Yaron10:
@alankilborn,

> I wasn't joking

Are you serious?! :)

That would mean customizing the order of ALL margins.
Well - if Don is willing to do that, it might be useful.

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
donho: Sure. It's just a draft. I will make the whole PR commitable once everyone is on the same page.

Yaron10:
@donho,

Very nice! 👍 
Thank you.

I suppose you know the current english.xml does not contain some recently added entries.

donho:
@Yaron10 
> I suppose you know the current english.xml does not contain some recently added entries.

Are there some entries still missing?

Yaron10:
@donho,

תמונה

vinsworldcom:
That looks like the compare Plus plugin??

donho:
@Yaron10 
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4cb63ff011064dfee1a1b5a25e47311fa726f33b

Yaron10:
@donho,

👍 

@vinsworldcom,

Yes, it's ComparePlus with "Show Only Diffs" ON.
After https://github.com/pnedev/comparePlus/commit/54c37ebcd52c4ecb2c3a0f5e6a1221b21f885909, you might see some hidden lines vertical separators.

donho:
> Perhaps so much debate about "best" margin ordering among a limited sample of 3/4 people means that the "best" option would be to allow end-users to configure any ordering they want. :-)

It could be a bug in my code that I couldn't find, or it's just limitation in Scintilla that line number margin should be the first one.
Anyway, I just swapped the values of these 2 margins: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L36
The compiled result is a mess:

image

If you people find the bug I didn't spot in Notepad++, please let  me know.
Otherwise this feature is just not possible.

vinsworldcom:
> If you people find the bug I didn't spot in Notepad++, please let me know.

@donho no bug found, but I did that the other day - swapped the numbers compiled and saw the same result you had.  So I just dropped it.  Not sure if it's a Scintilla limitation as I could put a line number margin in the "second" margin with PythonScript, but I don't know why the simple swap didn't work as it did for change and fold margins.

Cheers.

donho:
@vinsworldcom 
It could be in the Notepad++ code to set line number line with 0 value - but I could not find anything similar.
Thank you for your confirmation.

Yaron10:
@vinsworldcom,

Could you please post the Python code for changing the history-margin background color?

Thank you.

vinsworldcom:
> Could you please post the Python code for changing the history-margin background color?

```Python
from Npp import editor, editor1, editor2, MARGINTYPE, STYLESCOMMON

margin = 2
color = editor.styleGetBack(STYLESCOMMON.DEFAULT)

for ed in editor1, editor 2:
            ed.setMarginTypeN(margin, MARGINTYPE.COLOUR)
            ed.setMarginBackN(margin, color)
```

Cheers.

Yaron10:
@vinsworldcom,

I've pasted your code in `startup.py` and it won't work.

Thank you.

Yaron10:
`editor 2` to `editor2`. :)

Great! 👍 

Would you consider submitting a PR for that?

vinsworldcom:
> I've pasted your code in `startup.py` and it won't work.

Fixed!  thanks ..

> Would you consider submitting a PR for that?

That may be a bit beyond my capabilities.  I can easily add those 2 lines to Notepad++ code, but actually getting the color should come from the Style Configurator and there should be a new style for change history margin background.

Cheers.

Yaron10:
I see.

Hopefully @donho might find some time for that.

Yaron10:
@donho,

Thank you for https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6a801cc0a46754ff1339e2e551b43844f417fc22. I appreciate it. 👍 
(I was not by my PC and couldn't open the issue).

AndiKiissel:
@donho 
> It could be a bug in my code that I couldn't find, or it's just limitation in Scintilla that line number margin should be the first one.

https://www.scintilla.org/ScintillaDoc.html#SCI_SETMARGINTYPEN
... By convention, margin 0 is used for line numbers ...

Yaron10:
@vinsworldcom,

Setting the margin width to 2px seems more convenient for my personal use (you get 2px bar with no "padding").
Even with the default 9px width, clicking the margin would not be easy. :)

Hence, in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12248 I only requested two menu commands "Go to Next/Prev block".

Issue #12122, opened 2022-09-08
Allow filling in selected word in search and replace dialog via option


CrHasher:
Sometimes when you search and replace all you want is turn some variable name from lowercase to camelCase, and it would be very nice to have an option here:
![image](https://user-images.githubusercontent.com/834518/189073609-dee760f6-4756-4789-8a69-125777b2b22a.png)
"Fill in replace with field in Replace dialog with selected word"

Hope it gets implemented it would help me a lot.

alankilborn:
>Fill in replace with field in Replace dialog with selected word

That won't get you to camelCase much quicker than:

* place caret on desired word
* invoke _Replace_ via Ctrl+h
* press Ctrl+c
* press Tab
* press Ctrl+v
* make changes to lowercase word to turn it into camelCase word

It works because when Replace is invoked, the word at the caret is put into the _Find what_ box and that box gets input focus and the entire content there is selected.  Ctrl+c copies the selection and Tab moves focus to the _Replace with_ box and selects everything there.  A Ctrl+v paste gets your data duplicated there.

It sounds like a lot when described, but it is really not.

Just a suggestion that is available to you NOW.

------------------------------------------------------------------

Otherwise, IMO your suggestion is not a bad one, although perhaps a better way of doing it might be via an additional button on the _Replace_ window itself, e.g. this button could be moved down:

![image](https://user-images.githubusercontent.com/77065706/189115331-76cae4c9-51d8-4a05-bd2e-2ca7f60a070e.png)

and a second button inserted above it that looks like this (sorry, ASCII art, but it gets the point across?):

```
+-----------------------+
|                       |
|                       |
|           |           |
|           |           |
|           |           |
|           |           |
|           |           |
|        \  |  /        |
|         \ | /         |
|          \|/          |
|           '           |
|                       |
|                       |
+-----------------------+
```

With that button available, when you have data in the _Find what_ box that you want duplicated into the _Replace with_ box, you simply press this new button.

CrHasher:
Button still requires an extra step, instead why not a check box? If selected text gets duplicated when the dialog opens and field gets cleared or left empty when unselected, this way I can leave that checkbox checked and would be applied every time I open the replace dialog or leave the checkbox unchecked for current behavior. Obviously, this would only work if options in the dialog are persisted between sessions which I bet they are.

alankilborn:
@CrHasher 

[...]

PR #12195, opened 2022-09-18
fill Replace field the same as Find field

xomx:
Implements #12122 . 

This option can increase productivity when one needs only to modify an already existing text.


xomx:
setting:

fill_ReplaceField_same_as_FindField

alankilborn:
@xomx 
Looks totally awesome; thank you for your efforts.

Yaron10:
@xomx,

👍 
Thank you for your work.

What do you think of the following case?

For whatever reasons you uncheck "Fill Find Field with Selected Text".
You're working on a document containing multiple occurrences of the misspelled word "apreciate".
The correctly spelled "appreciate" is selected.
You might want to "Fill Replace Field with Selected Text".

How about the following two independent options?
1. "Fill Find Field with Selected Text".
2. "Fill Replace Field with Selected Text".

Disable "Select Word Under Caret..." only _if both options are unchecked_.

xomx:
@Yaron10 

> For whatever reasons you uncheck "Fill Find Field with Selected Text".
> You're working on a document containing multiple occurrences of the misspelled word "apreciate".
> The correctly spelled "appreciate" is selected.
> You might want to "Fill Replace Field with Selected Text".

In your scenario you still need to fill the Find field of the Replace dlg with the misspelled "apreciate" somehow.

When I find such problem in my text written, I usually click on the relevant word or select some invalid text, press Ctrl+H, then Ctrl+C, move to the Replace field and Ctrl+V there, then edit... I have never done the opposite (filling the Replace field with the correct text, Ctrl+C there, then move to the Find field...), but maybe that is just my way how to accomplish this task.

IMO if we have separate independent "Fill Replace Field with Selected Text" checked and the "Fill Find Field with Selected Text" unchecked, we are in the same situation as before this productivity improvement PR but now from the Find field point of view.

alankilborn:
@xomx 

IMO The logic of the PR is perfect the way it is.

Yaron10:
@xomx,

> In your scenario you still need to fill the Find field of the Replace dlg with the misspelled "apreciate" somehow.

That is correct. But theoretically one might still prefer this configuration "For whatever reasons".

Let's analyze it from a broader design point of view.

1. "Fill Find Field with Selected Text" and "Fill Replace Field with Selected Text" are not "father and child".
2. You programmatically disable a setting only if it's absolutely contrary to the common sense to keep it enabled.
This is not the case here.
3. Why not implement it in a more flexible way? - You can still have your productivity improvement.

Thanks again.

xomx:
@Yaron10 

Understood.

I do not promise anything, but I will try to look at it.

Is this conforming to what you are talking about?
FillReplaceField-Yaron10var
I just do not know if it will be clear that the "Select Word Under Caret ..." is for both Fill-checkboxes in case of such layout.

Yaron10:
@xomx,

> I do not promise anything, but I will try to look at it.

👍 
Thank you very much.

> Is this conforming to what you are talking about?

Yes.

> I just do not know if it will be clear that the "Select Word Under Caret ..." is for both Fill-checkboxes in case of such layout.

Good point.

How about moving "Select Word Under Caret ..." to the right in the middle of the two other settings?

alankilborn:
I said:

>IMO The logic of the PR is perfect the way it is.

I spoke too soon. :-(

With the new checkbox checked, invoking with Ctrl+h results in BOTH the _Find what_ and _Replace with_ text selected in their boxes:

image

This is likely confusing to a user because he doesn't really know where input focus is.  It's with _Find what_, but I don't think that is best.

In that situation, this is what I'd expect:

image

The reason for that is the supposition that the text from the document is what you want to replace most often, so the _Find what_ data is less likely to need changing than the _Replace with_ data.

Of course, that's my opinion only, so do what you like with it.

Since to this point, this issue/PR doesn't have @donho blessing at all, 
I think Don should take a look at it, and if the base idea is judged to be a smart one, then Don should direct any changes needed.

Yaron10:
> With the new checkbox checked, invoking with Ctrl+h results in BOTH the Find what and Replace with text selected in their boxes:

👍

donho:
@xomx 
Thank you for the PR.
However, the added option makes UI more confusing and I'm not sure this feature is really needed.
Let's see if the same request coming then I will consider this PR (and redo the UI eventually if this feature is useful for a lot of people).

xomx:
@donho 
Ok

@alankilborn 
It was easy to adapt this PR for the replacing (Ctrl+H):
focus_on_Replace_field_(AK)
but it has to stay as it is for the FindInFiles, as there one cannot know in advance for what purposes a user triggered it:
FindInFiles_autofill

@Yaron10 
The above change in code logic together with a possible GUI ambiguity arising from a necessary changes (in case I implemented your version we discussed earlier) is for me a reason not to implement such version in case Don decides to accept this PR in future. Otherwise I have to cross-check that focus position (more code = chance to introduce a bug/regression...).

alankilborn:
>but it has to stay as it is for the FindInFiles, as there one cannot know in advance for what purposes a user triggered it

I'd say that in this case it isn't a problem to have only  _Find what_ text selected.
But, as you like.

donho:
Instead of adding the settings into Preferences dialog, I prefer to use the existent swap button in Find replace dialog - it makes more sense and better user experience. Please see #13332

xomx:
@donho 

> I prefer to use the existent swap button in Find replace dialog

Thanks for a implementation of this, I got used to this feature (I used my own build of N++ with this PR before). Your variant is IMO better than my previous attempt.

What about a tooltip for the swap-button in the Find-dlg? (to help announce the rightclick functionality to the users...)

donho:
@xomx 
Hope you and your family are well.

> What about a tooltip for the swap-button in the Find-dlg? (to help announce the rightclick functionality to the users...)

I have thought about tooltip, but it might be "too polluted" on a small button. In any way, it will be documented on Notepad++ User Manual, so I let user to discover it (or from Google, chatgpt...)

PR #12236, opened 2022-09-24
Fix #12226, #12253 and properly fix #11504

### LINKED ISSUE #11504: shebang/xml heuristic not applied if default language is set

### LINKED ISSUE #12226: 8.4.5 color highlighting: XML language definition overrides specializations of XML
### Description of the Issue

In Notepad v8.4.5, file formats that are based on XML get detected as plain XML instead of their more specialized version. For example, I have a sample file in XHTML format [test.htm.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9638112/test.htm.zip):

```
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<title>Test</title>
</head>

<body>
<p>This is a test.</p>
</body>
</html>
```

This should identify as (X)HTML and should get highlighted as such - and it does, in versions up to 8.4.4. In 8.4.5, though, the file is identified as plain XML and highlighted using the standard XML highlighter. The same happens in other XML-based files, e.g. MSBuild project files, provided that they use the XML header:
```
<?xml version="1.0"?>
```
I suspect it was caused by https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11504 - the description seems to explain what I am seeing.

In my opinion, the new behavior is not correct - a special highlighter should always take precedence over the generic one.

1. Download Notepad++ v8.4.5, portable version and unpack it to any directory.
2. Unpack my sample file above to the same directory.
3. Run `notepad++.exe test.htm`.

### Expected Behavior
The file should get highlighter using the HTML highlighter. The easiest way to check it is, go to the Language menu - `H`->`HTML` should be selected.

### Actual Behavior
The file gets highlighted using the XML highlighter. In the Language menu, `XML` is selected.

### LINKED ISSUE #12253: XML auto-detection overrides file extension language selection
### Description of the Issue
I have a consolidated logs file that contain a portion of XML at the beginning, and a large amount of ASCI logs (~80MB) following the XML structure. I use a user defined language to highlight particular data within the log, and I've associated certain extensions so notepad++ will correctly select my user defined language instead of attempting to parse the entire log as XML.  
After the recent update to v8.4.5, my logs no longer are detected using the *.logs extension and defaults to XML as the language.

1.  Create A User Defined Language, using 'logs' as the File extension type for the language, or ADD 'logs' as a user extension for the "dos style" language in the Style Configurator.
2.  Create a file, called 'test.logs', 
3. Add `<?xml version="1.0" encoding="UTF-8"?>` in the first line of the file.
4.  Save 'test.logs' and close the file.
5.  Open the file 'test.logs'. Open the language tab and observe the selected language.

### Expected Behavior
On step 4, it is expected 'test.logs' to be auto-detected as the User defined language.

### Actual Behavior
The file will be opened as the detected XML language.

### Debug Information
Notepad++ v8.4.5   (64-bit)
Build time : Sep  3 2022 - 04:05:32
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.1889
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExec (0.6)
    NppExport (0.4)

rdipardo:
Before 6263ce5, the user-preferred default language was being applied to files with _no extension_, but got _overwritten_ by `L_TEXT` when the extension was present _but unknown_. This was why the `buf->_lang == L_TEXT` condition failed _only for buffers with no extension_.

This fixes #11504 by handling the no-extension case, without causing regressions.
Also fixes #12226 and fixes #12253

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -742,7 +742,7 @@ BufferID FileManager::loadFile(const TCHAR* filename, Document doc, int encoding
 		buf->setEncoding(-1);
 
 		// if no file extension, and the language has been detected,  we use the detected value
-		if (!newBuf->_isLargeFile && loadedFileFormat._language != L_TEXT)
+		if (!newBuf->_isLargeFile && ((buf->getLangType() == L_TEXT) && (loadedFileFormat._language != L_TEXT)))
 			buf->setLangType(loadedFileFormat._language);
 
 		setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -211,6 +211,8 @@ void Buffer::setFileName(const TCHAR *fn, LangType defaultLang)
 			newLang = nppParamInst.getLangFromExt(ext);
 		}
 	}
+	else if (!isUntitled()) // existing file with no extension
+		newLang = L_TEXT;
 
 	if (newLang == defaultLang || newLang == L_TEXT)	//language can probably be refined
 	{
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -741,19 +741,10 @@ BufferID FileManager::loadFile(const TCHAR* filename, Document doc, int encoding
 		// restore the encoding (ANSI based) while opening the existing file
 		buf->setEncoding(-1);
 
-		NppParameters& nppParamInst = NppParameters::getInstance();
-		const NewDocDefaultSettings& ndds = (nppParamInst.getNppGUI()).getNewDocDefaultSettings();
-
-		// if no file extension, or the file type is unknown, and the language has been detected, we use the detected value
-		if (!newBuf->_isLargeFile &&
-			// case 1: no file extension
-			(!_tcsrchr(newBuf->_fileName, '.') ||
-				// case 2: file extension found, but the file type is unknown (i.e. unchanged from the default)
-				(buf->_lang == L_TEXT || buf->_lang == ndds._lang)
-			&& loadedFileFormat._language != L_TEXT))
-		{
+		// if no file extension, and the language has been detected,  we use the detected value
+		if (!newBuf->_isLargeFile && loadedFileFormat._language != L_TEXT)
 			buf->setLangType(loadedFileFormat._language);
-		}
+
 		setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
 
 		//determine buffer properties
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -741,10 +741,19 @@ BufferID FileManager::loadFile(const TCHAR* filename, Document doc, int encoding
 		// restore the encoding (ANSI based) while opening the existing file
 		buf->setEncoding(-1);
 
-		// if no file extension, and the language has been detected,  we use the detected value
-		if (!newBuf->_isLargeFile && loadedFileFormat._language != L_TEXT)
-			buf->setLangType(loadedFileFormat._language);
+		NppParameters& nppParamInst = NppParameters::getInstance();
+		const NewDocDefaultSettings& ndds = (nppParamInst.getNppGUI()).getNewDocDefaultSettings();
 
+		// if no file extension, or the file type is unknown, and the language has been detected, we use the detected value
+		if (!newBuf->_isLargeFile &&
+			// case 1: no file extension
+			(!_tcsrchr(newBuf->_fileName, '.') ||
+				// case 2: file extension found, but the file type is unknown (i.e. unchanged from the default)
+				(buf->_lang == L_TEXT || buf->_lang == ndds._lang)
+			&& loadedFileFormat._language != L_TEXT))
+		{
+			buf->setLangType(loadedFileFormat._language);
+		}
 		setLoadedBufferEncodingAndEol(buf, UnicodeConvertor, loadedFileFormat._encoding, loadedFileFormat._eolFormat);
 
 		//determine buffer properties
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1086,7 +1086,6 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 	std::lock_guard<std::mutex> lock(save_mutex);
 
 	Buffer* buffer = getBufferByID(id);
-	LangType language = buffer->_lang;
 	bool isHiddenOrSys = false;
 	DWORD attrib = 0;
 
@@ -1157,11 +1156,8 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 			}
 		}
 
-		if (!_tcsrchr(buffer->_fileName, '.'))
-		{
-			// no extension -- check the language du fichier
-			language = detectLanguageFromTextBegining((unsigned char *)buf, lengthDoc);
-		}
+		// check the language du fichier
+		LangType language = detectLanguageFromTextBegining((unsigned char *)buf, lengthDoc);
 
 		UnicodeConvertor.closeFile();
 
@@ -1511,7 +1507,7 @@ bool FileManager::loadFileData(Document doc, int64_t fileSize, const TCHAR * fil
 						fileFormat._encoding = detectCodepage(data, lenFile);
                 }
 
-				if (!_tcsrchr(filename, '.'))
+				if (fileFormat._language == L_TEXT)
 				{
 					// check the language du fichier
 					fileFormat._language = detectLanguageFromTextBegining((unsigned char *)data, lenFile);
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1086,6 +1086,7 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 	std::lock_guard<std::mutex> lock(save_mutex);
 
 	Buffer* buffer = getBufferByID(id);
+	LangType language = buffer->_lang;
 	bool isHiddenOrSys = false;
 	DWORD attrib = 0;
 
@@ -1156,8 +1157,11 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 			}
 		}
 
-		// check the language du fichier
-		LangType language = detectLanguageFromTextBegining((unsigned char *)buf, lengthDoc);
+		if (!_tcsrchr(buffer->_fileName, '.'))
+		{
+			// no extension -- check the language du fichier
+			language = detectLanguageFromTextBegining((unsigned char *)buf, lengthDoc);
+		}
 
 		UnicodeConvertor.closeFile();
 
@@ -1507,7 +1511,7 @@ bool FileManager::loadFileData(Document doc, int64_t fileSize, const TCHAR * fil
 						fileFormat._encoding = detectCodepage(data, lenFile);
                 }
 
-				if (fileFormat._language == L_TEXT)
+				if (!_tcsrchr(filename, '.'))
 				{
 					// check the language du fichier
 					fileFormat._language = detectLanguageFromTextBegining((unsigned char *)data, lenFile);


rdipardo:
Updated to check for more conditions:

| |Case | Preferred file type | Reason |
|:-- |:-- |:-- |:-- |
| 1. | no file extension  | auto-detected | scripting languages often omit the extension because they have shebang lines |
| 2. | file extension, but type is `L_TEXT` | auto-detected | `L_TEXT` means "no language", or, at least, "unrecognized language" |
| 3. | file extension, but type is user-preferred default | auto-detected | in this case the default *also* means "no language", because the choices of default language all have standard file extensions. If we reach this condition, it means the extension did not match with a builtin lexer (so probably an oddball extension) |

If __*any*__ of cases 1, 2, or 3 is false, the buffer's file type will not be changed.

As revealed by https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12226#issuecomment-1257084226, there are many undocumented use cases for the zealous auto-detection logic of the current implementation. Any changes to it have to be small and incremental to avoid a torrent of regression reports.

ArkadiuszMichalski:
It makes sense now. I played around a bit and it seems to be working as expected.

ArkadiuszMichalski:
@rdipardo Another thing, probably you put the parentheses in wrong place, it should be:
```c++
		// if no file extension, or the file type is unknown, and the language has been detected, we use the detected value
		if (!newBuf->_isLargeFile &&
			// case 1: no file extension
			(!_tcsrchr(newBuf->_fileName, '.') ||
				// case 2: file extension found, but the file type is unknown (i.e. unchanged from the default)
				(buf->_lang == L_TEXT || buf->_lang == ndds._lang))
			&& loadedFileFormat._language != L_TEXT)
		{
			buf->setLangType(loadedFileFormat._language);
		}
```

Without this change you brake name recognition  (without extension) and brake apply default language for unrecognized files (without extension). These two things work in older Notepad++ versions.

Btw, I don't know why the default language (from Preferences) is applied to unrecognized files without the extension, but not to unrecognized files with the extension.

rdipardo:
Thanks, @ArkadiuszMichalski, for the review.

After lot's of head-scratching, a much simpler fix for all the related issues has been found and pushed.

You can be certain that it fixes #12226. It's more important to check that it doesn't cause a recurrence of #11504 (which was pretty obscure, anyway).

### TL;DR

The user-preferred default language was being applied to files with *no extension*, but got *overwritten* by `L_TEXT` when the extension was present *but unknown*.

This was the true underlying cause of #11504. It should have been fixed at the source by simply handling the no-extension case in `Buffer::setFileName`.

Best way to explain is with examples:

| Case | Before 6263ce5 | After 6263ce5 | After this PR | Why it works |
| :-- | :-- | :-- | :-- | :-- |
| Open [a Code::Blocks project file][cbp] (extension `*.cbp`) with *no default language set* | XML highlighting | XML highlighting | XML highlighting | The `*.cbp` extension is unknown, so `buf->_lang == L_TEXT`. __*Auto-detection is performed because [this condition is always true][0]*__. Auto-detection finds the XML doctype declaration and sets the language to `L_XML` around line 1513. Now `buf->_lang == L_TEXT`, and `loadedFileFormat._language == L_XML`. Auto-detection wins. |
| Set a default language (e.g. C++); open the same Code::Blocks project file | XML highlighting | XML highlighting | XML highlighting | The `*.cbp` extension is unknown, so `buf->_lang == L_TEXT`, because `Buffer::setFileName` [calls `NppParameters::getLangFromExt` __*when there's a file extension*__][1], and `NppParameters::getLangFromExt` returns `L_TEXT` for unknown extensions (wiping out the user-preferred default language). __*Auto-detection is performed because [this condition is always true][0]*__. Auto-detection finds the XML doctype declaration. Now `buf->_lang == L_TEXT`, and `loadedFileFormat._language == L_XML`. Auto-detection wins.|
| Open [a simple bash script][sh], which has *no extension*, with *no default language set* | Bash highlighting | Bash highlighting | Bash highlighting | No extension, so `buf->_lang == L_TEXT`. __*Auto-detection is performed because [this condition is always true][0]*__. Auto-detection finds the shebang line and sets the language to `L_BASH` around line 1513. Now `buf->_lang == L_TEXT`, and `loadedFileFormat._language == L_BASH`. Auto-detection wins. |
| Set a default language (e.g. C++); open the same bash script | __*default language*__ highlighting | Bash highlighting | Bash highlighting | No extension, so  `Buffer::setFileName` [__*will not*__ call `NppParameters::getLangFromExt`][3], and the user-preferred language __*is not*__ reverted to `L_TEXT`. __*Auto-detection is performed because [this condition is always true][0]*__. Auto-detection finds the shebang line and sets the language to `L_BASH`. But `buf->_lang == L_CPP` (the user-preferred default). Before 6263ce5, the __*default language*__ wins, because the auto-detected value is never checked. After this PR, auto-detection wins, because we now handle the no-extension case (for named existing files only), so `buf->_lang == L_TEXT`, and `loadedFileFormat._language == L_BASH`. |
| Open any `*.html` document conforming to the [XHTML 1.0 specification] | HTML highlighting | __*XML*__ highlighting | HTML highlighting | The extension matches the HTML lexer. __*Auto-detection is performed because [this condition is always true][0]*__. Auto-detection finds the XML doctype declaration and sets the language to `L_XML`. Before 6263ce5, the HTML lexer wins, because `buf->_lang != L_TEXT`, so the auto-detected value is never checked. After 6263ce5, the auto-detected language is checked *in every case*, and wins, which is wrong. This PR restores the former behaviour |


[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0ce3bc08df1a7de8eff1ec1d70ee3b713224407c/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1510-L1514
[1]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e3a5f76af0ba15c64f95955a6b24d2e5431e024a/PowerEditor/src/ScintillaComponent/Buffer.cpp#L197
[2]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e3a5f76af0ba15c64f95955a6b24d2e5431e024a/PowerEditor/src/ScintillaComponent/Buffer.cpp#L195
[3]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e3a5f76af0ba15c64f95955a6b24d2e5431e024a/PowerEditor/src/ScintillaComponent/Buffer.cpp#L180

[XHTML 1.0 specification]: https://www.w3.org/1999/xhtml/
[cbp]:https://github.com/notepad-plus-plus/notepad-plus-plus/files/9650377/sample.cbp.zip
[sh]: https://github.com/notepad-plus-plus/notepad-plus-plus/files/9650380/sample.zip

ArkadiuszMichalski:
But you still introduce a regression for recognizing some knowing names:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/caa6a416349de6ec64d5c81a14d0ad9dd0a9b8f7/PowerEditor/src/ScintillaComponent/Buffer.cpp#L217

Your last fix can't recognize this files. In previous versions of Notepad++ this worked. This will work after move one `)` to other place: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12236#issuecomment-1257229421

I haven't analyzed the new changes yet because it's a bit twisted :), so I don't know what the problem is.

rdipardo:
> But you still introduce a regression for recognizing some knowing names:
> 
> https://github.com/notepad-plus-plus/notepad-plus-plus/blob/caa6a416349de6ec64d5c81a14d0ad9dd0a9b8f7/PowerEditor/src/ScintillaComponent/Buffer.cpp#L217
> 

In context:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/dfd0760f6cf6b71ab7ab20da8b7fd3e5a919262a/PowerEditor/src/ScintillaComponent/Buffer.cpp#L214-L217

So, `makefile`:

    if (FALSE)
    {
       // ....
    }
    else if (TRUE)
      newLang = L_TEXT;
    
    if (L_TEXT == defaultLang || L_TEXT == L_TEXT) // TRUE
    {
      // This runs
    }

ArkadiuszMichalski:
Sorry, looks like I use wrong binary. Download once again and it works.

rdipardo:
> I still thinking about this Default language option:
> [ . . .]
> apply default language from prefs also for opened document without extension (but not recognized)

That's an exact description of bug #11504. If that's the desired behaviour (assuming users are even aware of it), then #11504 is a "won't fix". We simply revert 6263ce5 and we're done.

>  I don't judge whether it was introduced intentionally or not.

I don't think the default language was meant for unidentified files. The user has created a file and knows what type it is. Automatically replacing it with a setting that could be switched on by accident seems like a bug. The decision branches are really, truly f****d up, so it's not surprising they make the wrong choice. There is no documentation explaining what the *correct* choice is supposed to be.

I say the editor should respect the file's contents, and limit the default setting to brand new files. I [said so in #11992][0], and no one objected.

Of course, it's @donho's decision to make.

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11992#issue-1330847258

alankilborn:
>There is no documentation explaining what the correct choice is supposed to be.

@pryrt Perhaps for the manual you may want to keep an eye on how this all shakes out.

ArkadiuszMichalski:
>That's an exact description of bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11504. If that's the desired behaviour (assuming users are even aware of it), then https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11504 is a "won't fix". We simply revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6263ce5dece595543bbfaa4207391075ae3a212a and we're done.

On this bug the problem was that shebangs was overwritten by default language from Prefs. So fixing this is proper decision. But still, after all checks, if nothing was matched, the default lang from prefs can be used (if it's other than L_TEXT). 

>There is no documentation explaining what the correct choice is supposed to be.

Sure, having it the rest is just a formality. This is why I am asking directly about each case so that later there will be no ambiguity again.

pryrt:
I will definitely be documenting the language-identification priority once this fix goes into effect, and will be borrowing heavily from @ArkadiuszMichalski's points#1-7 above.

I personally don't care whether the "new document: default language" gets applied to unidentified files, as long as it never overrides one of the other 6 rules.  But I do remember that some users have asked that "default language" gets applied to existing files with no extension or unknown extension -- so maybe giving an option to the user whether #7 is enabled or not:
image

rdipardo:
@donho,
> I will accept this PR (its final state), as it keeps the legacy behaviour. 

Just to be clear, if by "legacy behaviour" you mean "apply the default language to unknown files __*that have no file extension*__", then the final state of this PR is __*not*__ what you want. We need to back out of dfd0760 to restore that exact behaviour. 

If instead you mean "the behaviour that users consciously expect: Lexilla/UDL lexer 1st, auto-detected language 2nd, default language for brand new files", then everything is already set to go. 

Keep in mind, the default language has __*always been*__ ignored __*anytime the file has an extension AND the file type can be guessed from the content*__ (see the 2nd example of the [Code::Blocks project file][0] above). 

Changing that behaviour is outside the scope of this PR. The *only* thing this PR will do is make the *existing* behaviour more *consistent* by *also* applying it to files __*that have no extension*__.

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12236#:~:text=Set%20a%20default%20language%20(e.g.%20C%2B%2B)%3B%20open%20the%20same%20Code%3A%3ABlocks%20project%20file

donho:
> Just to be clear, if by "legacy behaviour" you mean "apply the default language to unknown files **_that have no file extension_**", then the final state of this PR is **_not_** what you want. We need to back out of [dfd0760](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/dfd0760f6cf6b71ab7ab20da8b7fd3e5a919262a) to restore that exact behaviour.

@rdipardo 
I'm confused.
Could you compare between the "lagacy behaviour" and the behaviour of this PR please?

ArkadiuszMichalski:
@donho This PR does exactly such steps: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12236#issuecomment-1257168802.

But one thing has been canceled (and it worked before Notepad++ 8.4.5): when we change default lang from Prefs then this lang is apply for unidentified files (with an unknown file extension, even when the file's content matches a detectable language). STR:
- create file `a` (with `<?xml version="1.0" encoding="utf-8"?>` as content)
- open NPP and change default lang in Prefs to somethin else (like ActionScript)
- open this file in NPP

And results:
- before NPP 8.4.5 we have: detected as Flash ActionScript file
- NPP 8.4.5 and 8.4.6 we have: detected as XML
- this PR use Normal text file (always)

This PR make that default lang from Prefs is applied only to new files (new file), not those loaded and unidentified (with or without extension).

Here is discussion if default lang from Prefs should have effect for such file (if yes then for which one and what to do with content detection in such case) https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12256.

donho:
@ArkadiuszMichalski 
> But one thing has been canceled (and it worked before Notepad++ 8.4.5): when we change default lang from Prefs then this lang is apply for unidentified files (with no extension, even when the file's content matches a detectable language). STR:
>
>   * create file a (with <?xml version="1.0" encoding="utf-8"?> as content)
>   * open NPP and change default lang in Prefs to somethin else (like ActionScript)
>   * open this file in NPP

Yes, I am aware this issue, and it's a bug to me.
So what I said "legacy behaviour", I meant the current behaviour with the bug fixed. 
The order is `File extension > Content detected > Default Language`.
Are we on the same page?

ArkadiuszMichalski:
@donho Yes if we assume that `Default Language` means always Normal Text (if detection did not find anything).

donho:
Conflict while rebasing:
```
$ git rebase -i
[detached HEAD ae2e424d] Fix content detection language overrides the file extension
 Author: Robert Di Pardo <59004801+rdipardo@users.noreply.github.com>
 Date: Sat Sep 24 18:58:11 2022 -0400
 1 file changed, 7 insertions(+), 3 deletions(-)
interactive rebase in progress; onto 23fc940a
Last commands done (2 commands done):
   reword 870677b6 Check file header only when there's no extension
   fixup 4cb1acee Revert "Check file header only when there's no extension"
Next commands to do (4 remaining commands):
   fixup caa6a416 Don't override builtin lexer language with auto-detected file type
   fixup d9f6064a Revert "Don't override builtin lexer language with auto-detected file type"
  (use "git rebase --edit-todo" to view and edit)
You are currently rebasing branch 'master' on '23fc940a'.
  (all conflicts fixed: run "git rebase --continue")

No changes
You asked to amend the most recent commit, but doing so would make
it empty. You can repeat your command with --allow-empty, or you can
remove the commit entirely with "git reset HEAD^".
Could not apply 4cb1acee... Revert "Check file header only when there's no extension"
```

I will do it myself.

donho:
@rdipardo 

I just revert the old commit and solve the conflict.
Could you create a new PR to apply the fix of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11504 please:
```
else if (!isUntitled()) // existing file with no extension
		newLang = L_TEXT;
```

rdipardo:
@ArkadiuszMichalski,
> Yes if we assume that `Default Language` means always Normal Text (if detection did not find anything).


Nothing has changed from the user's perspective because "Normal Text" means "no language at all". 
The language menu actually says "__*None*__ (Normal Text)"; nobody actively chooses it.
That is consistent with the internal logic, which depends on finding `L_TEXT` to decide when the file's content should prevail.

Like I said in #12256, a future option to prefer "Default Language" over "Content detected" is going to be more complicated 
than just a single check box; there are at least two cases with different outcomes:

     1. unkown file extension => choose file content (current and future behaviour, not affected by this PR)
 
     2. no extension at all =>  * choose default language (< 8.4.5)
                                * choose file content (>= 8.4.5; made more explicit by this PR)

rdipardo:
> So what I said "legacy behaviour", I mean the current behaviour with the bug fixed.
> That is: the order `File extension > Content detected > Default Language`.

@pryrt, that decision order looks like something we can offically document. 
If at some future time it becomes configurable, simply update it with a `Since version x.y.z` annotation.

pryrt:
>  that decision order looks like something we can offically document

I'll add that to my todo list

Issue #12117, opened 2022-09-07
Esc key should abort "Move to Recycle Bin" confirmation prompt


alankilborn:
### Description of the Issue
As a nicety, pressing the Esc key should dismiss the confirmation popup that appears in response to the "Move to Recycle Bin" command, effectively answering "No" to the prompt.

### Steps to Reproduce the Issue

1. Activate a hardnamed file tab
2. Right-click the tab and choose _Move to Recycle Bin_
3. Confirmation appears:
![image](https://user-images.githubusercontent.com/77065706/188895237-d33fa4ab-8c6c-4ead-88e4-8cc784dcf347.png)
4. Press the Esc key.

### Expected Behavior
The dialog box to close and my file NOT being moved to Recycle Bin and remain an open tab in Notepad++.

### Actual Behavior
Nothing happens.

### Debug Information
Notepad++ v8.4.5   (64-bit)
Build time : Sep  3 2022 - 04:05:32
Path : C:\\...\npp.8.4.5.portable.x64\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.1889
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    PythonScript (2)

smartyc30:
Hi, I am a first time contributor and I would like to try my hand at solving this issue. I use Notepad++ often, so it'd be pretty cool is I could make a contribution. Can you guys assign me this issue?

alankilborn:
@smartyc30 You don't need to be assigned to an issue to work on it.  However, this issue was just an idea I had for the software, and hasn't (yet) been agreed to as a good idea by Notepad++'s author.  The author may never evaluate this issue (there are a lot of issues open), but he will review all submitted PRs, judging acceptability at that time.  What I'm saying is there's no guarantee that if you work on it, that it will make its way into the software.

smartyc30:
@alankilborn Ok I understand, thanks for the heads. I'll still check it out anyway. Even if what I do doesn't make it into the software, it'd still be good experience for me to try.

PR #12239, opened 2022-09-25
ESC Key now aborts "Move to Recycle Bin" confirmation prompt

smartyc30:
#12117


smartyc30:
@donho Will this enhancement be included in the next version of np++?

donho:
@smartyc30 
It won't be in v8.4.6 since it's on its way:
https://community.notepad-plus-plus.org/topic/23524/notepad-v8-4-6-release-candidate-3

But the PR will be  merged after the release 8.4.6, so this enhancement will be in v8.4.7.

Issue #8149, opened 2020-04-21
hide-markers disappear when two subsequent lines are hidden separately


sskras:
### Description of the Issue
When two subsequent lines are hidden one by one, hide-markers seemingly combine into one block, that's OK:

![image](https://user-images.githubusercontent.com/7887758/79848282-90d89680-83c9-11ea-8d8b-4a77d03ae8e0.png)

But when I try to expand it, the first line remains hidden:

![image](https://user-images.githubusercontent.com/7887758/79848469-c8dfd980-83c9-11ea-8d81-77d97f5036c9.png)

And its hide-markers disappears. Line 2 got lost behind the scene. 
You don't see it in the tab and can damage it accidentally.

### Steps to Reproduce the Issue
1. Enter five lines into a new tab.
2. Go to line no. 2 and hide it (Alt-H).
3. See green hide-markers appearing on lines 1 and 3.
4. Go to line no. 3 and hide it too.
5. See closing hide-markers now jumping onto line 4.
6. Click the a hide-marker (any of the two).

### Expected Behavior
I expected both lines to be revealed.

### Actual Behavior
Only line no. 3 got revealed. Line no. 2 got visually lost.

### Debug Information
```
Notepad++ v7.8.5   (64-bit)
Build time : Mar  4 2020 - 11:08:34
Path : C:\Users\saukrs\_\bin\npp.7.8.5.bin.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1803
OS Build : 17134.1365
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll
```
### Real time video:

![2020 04 21-11_55_19-notepad-_v7 8](https://user-images.githubusercontent.com/7887758/79849692-7a333f00-83cb-11ea-8ae0-3044c3059b52.gif)

### Workaround

If this is an unsaved file, you can't just reload the content without losing part of it.
I worked around this by doing:
* Select All
* Cut
* Undo

Line no. 2 got back.

sskras:
This might be a continuation (an unfixed / corner case of) of bug #1112.

nCubed:
In addition to this issue with Hide Lines as well as bug #1112, there is no way to unhide lines if "bookmarks" are not enabled. Was going to open a new issue, but figured I'd piggy-back on this open issue and let the devs decide if they want to create a new issue.

[...]

PR #12250, opened 2022-09-26
Fix line hiding issues

### LINKED ISSUE #12184: NPP crashes on hiding lines
**STR:**
Paste the following lines in a new document.
```
Test
Test
Test
Test
Test
```
Place the caret at line 3.
View -> Hide Lines.
Place the caret at line 2.
View -> Hide Lines.
Click on the show-lines markers.
You should get this:
תמונה
Click on the single show-lines marker.

**Result:**
NPP crashes.

***

Notepad++ v8.4.5   (64-bit)
Build time : Sep 16 2022 - 07:53:18
Path : C:\***Notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Version : 20H2
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : 
    ComparePlus (1)
    HTMLTag (1.3.2)
    MenuIcons (1.2.5)
    PythonScript (2)
    _CustomizeToolbar (4.2)

Adrian104:
This pull request fixes line hiding problems caused by `hideLines()` not working properly.
Fixes #12184 and #8149

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -3525,7 +3525,7 @@ void ScintillaEditView::hideLines()
 	_currentBuffer->setHideLineChanged(true, startMarker);
 }
 
-bool ScintillaEditView::markerMarginClick(size_t lineNumber)
+bool ScintillaEditView::markerMarginClick(intptr_t lineNumber)
 {
 	auto state = execute(SCI_MARKERGET, lineNumber);
 	bool openPresent = ((state & (1 << MARK_HIDELINESBEGIN | 1 << MARK_HIDELINESUNDERLINE)) != 0);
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -3522,7 +3551,7 @@ bool ScintillaEditView::markerMarginClick(size_t lineNumber)
 
 		if (openPresent)
 		{
-			_currentBuffer->setHideLineChanged(false, lineNumber);
+			_currentBuffer->setHideLineChanged(false, lineNumber + 1);
```
Adrian104, line 3554: Variable `lineNumber` is always decremented before each iteration in this `for` loop, even before `lineNumber >= 0 && !openPresent` evaluates to false. As a result, after this loop `lineNumber` will always be decremented too much by one. To solve this, one is added back to `lineNumber`

Yaron10:
@Adrian104,

Both https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12184 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8149 are fixed. 👍 

Great job! Thank you.

donho (changes requested):
Please also change the signature of **markerMarginClick** from  `bool markerMarginClick(size_t lineNumber);` to `bool markerMarginClick(intptr_t lineNumber);`

donho (approved):

PR #12286, opened 2022-10-04
Properly fix #11504

### LINKED ISSUE #11504: shebang/xml heuristic not applied if default language is set

rdipardo:
Extracted from #12236 via:

    curl -sL https://github.com/rdipardo/notepad-plus-plus/commit/dfd0760f6cf6b71ab7ab20da8b7fd3e5a919262a.patch |\
        git am --ignore-space-change 

Fixes #11504 (without regressions)

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -211,6 +211,8 @@ void Buffer::setFileName(const TCHAR *fn, LangType defaultLang)
 			newLang = nppParamInst.getLangFromExt(ext);
 		}
 	}
+	else if (!isUntitled()) // existing file with no extension
+		newLang = L_TEXT;
 
 	if (newLang == defaultLang || newLang == L_TEXT)	//language can probably be refined
 	{


donho:
Thank you @rdipardo !

rdipardo:
@donho in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12236#issuecomment-1267623734,
> Conflict while rebasing:

*Comme ça*:

~~~
curl -sL https://github.com/notepad-plus-plus/notepad-plus-plus/commit/aee115eb5f66aa1f4571ed6a2eb52c638a4c8296.patch | git am --ignore-space-change
~~~

*__Toujours__ avec* [`--ignore-space-change`](https://git-scm.com/docs/git-am#Documentation/git-am.txt---ignore-space-change)

PR #12290, opened 2022-10-05
Support per-theme (and dark mode) config for external lexers

### LINKED ISSUE #9516: Lexer plugin is sorted unconventionally in language menu.
Currently, a lexer from a plugin is added to the language menu in an unconventional way.
When the name of the lexer comes before the “normal text” lexer, from the alphabetical point of view, it appears before “normal text”, but when it comes after that, it appears below it, but inside the Builtin lexers. Example see nim and vlang lexer from the image below.

image

### Expected Behavior
Either external lexers should appear within the sorting of the null lexer or within the builtin lexers

### Actual Behavior
Some here some there

### Debug Information
Notepad++ v7.9.3   (64-bit)
Path : D:\Tests\npp\_latest\x64_clean\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Build : 7601.24546
Current ANSI codepage : 1252
Plugins : BetterMultiSelection.dll mimeTools.dll NppConverter.dll NppExec.dll NppExport.dll nppnim.dll PythonScript.dll VLang.dll

### LINKED ISSUE #12190: Enhancement: Support themes and dark mode for external lexers
### Description of the enhancement
In current code, themes are supported for built-in lexers by having all of their styles list in individual theme configuration files under {data folder}\themes. However, for external lexers there isn't such a thing, The styles are always loaded and stored (when user configured) to {plugin}.xml under {data folder}\plugins\config, regardless what theme is currently used/configured for. This should be enhanced for better user experience, especially for users who switch between default and dark mode depending on the time of the day.

@donho If you don't mind, I can try to work on this. My idea is to have a new "themes" folder under {data folder}\plugins\config, and for each theme, a sub-folder under it to store each individual plugin's styling configuration. For example, for my Papyrus plugin, the config file for dark mode will be stored in {data folder}\plugins\config\themes\DarkModeDefault\Papyrus.xml.

To be compatible with existing lexer plugins, the logic for loading styles in current theme would be:
- Check if {data folder}\plugins\config\themes\{theme} exists
  - If yes, check if {data folder}\plugins\config\themes\{theme}\{plugin}.xml exists
    - If yes, load from it
- In all other cases, load from {data folder}\plugins\config\{plugin}.xml and save a copy as {data folder}\plugins\config\themes\{theme}\{plugin}.xml 

The logic for style configurator is simple. It will always save to {data folder}\plugins\config\themes\{configured theme}\{plugin}.xml for external lexers.

blu3mania:
When a theme (or dark mode) gets applied, external lexer's config will be loaded from {data folder}\plugins\config\themes{theme}. If this folder doesn't exist or the config file for the lexer doesn't exist, it is copied from the lexer's base config file under {data folder}\plugins\config.

Since enabling/disabling external language is affected, this PR also fixed the logic to insert disabled language back to Language menu after it is re-enabled (the old logic assumed a hardcoded number of items at the end of the menu and the number was incorrect anyway), which applies to adding external lexers to Language menu when Notepadd++ starts as well. Note, currently the logic to insert into Language menu is essentially duplicated in preferenceDlg.cpp and Notepad_plus.cpp, as there doesn't seem to be a good place to host a common utility function. Asking Notepad_plus to call this method on preferenceDlg would be weird. Alternatively it could be hosted by NppParameters, but it also seems weird...

This closes #12190 and also closes #9516.

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -199,35 +199,12 @@ int PluginsManager::loadPluginFromPath(const TCHAR *pluginFilePath)
 			//if (!GetNameSpace)
 				//throw generic_string(TEXT("Loading GetNameSpace function failed."));
 
-			// Assign a buffer for the lexer name.
-			char lexName[MAX_EXTERNAL_LEXER_NAME_LEN];
-			lexName[0] = '\0';
-
-			int numLexers = GetLexerCount();
-
-			ExternalLangContainer* containers[30];
-
-			for (int x = 0; x < numLexers; ++x)
-			{
-				GetLexerName(x, lexName, MAX_EXTERNAL_LEXER_NAME_LEN);
-				if (!nppParams.isExistingExternalLangName(lexName) && nppParams.ExternalLangHasRoom())
-				{
-					containers[x] = new ExternalLangContainer;
-					containers[x]->_name = lexName;
-					containers[x]->fnCL = CreateLexer;
-					//containers[x]->fnGLPN = GetLibraryPropertyNames;
-					//containers[x]->fnSLP = SetLibraryProperty;
-				}
-				else
-				{
-					containers[x] = NULL;
-				}
-			}
-
 			TCHAR xmlPath[MAX_PATH];
 			wcscpy_s(xmlPath, nppParams.getNppPath().c_str());
 			PathAppend(xmlPath, TEXT("plugins\\Config"));
-            PathAppend(xmlPath, pi->_moduleName.c_str());
+			size_t dirEndIndex = wcslen(xmlPath);
+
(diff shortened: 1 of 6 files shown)


blu3mania:
@donho The build failed with GCC on std::basic_string::ends_with() which was introduced in C++ 20. Does MinGW-w64 v8.1.0 even support C++ 20 standard? According to https://www.cppstories.com/2020/08/string-prefix-cpp20.html/ you may need at least GCC 9 for tis to work.

donho:
@blu3mania 
It's fixed now.
I see your PR passed for MSVC so it's OK.
Thank you for your hint.

donho:
The added code makes the source code more complex and more difficult to maintain. If there are more feature requests about this feature, I will consider it.

donho:
No request so far, so I think we can close this PR. 
If there are the similar request coming, I will reconsider it.

PR #12295, opened 2022-10-06
Refactoring For setting default language for the new document

### LINKED ISSUE #12256: New Document > Default Language: add "☐ Apply to unidentified files" option
### Description of the Request

In the **Settings > Preferences > New Document** interface, add "☐ Apply to unidentified files" option to the Default Language

see [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12236#issuecomment-1259988624) and following for context

donho:
~~Fix the wrong behaviour for the set default language which is applied to opened existing file.~~
~~The behaviour of current set default language in the new document section is wrong: it applies not only to the new document, but also to the opened existing file in which the language can't not be defined by Notepad++.
This PR makes this setting apply only to the new document so the behaviour is not inconsistent anymore.~~
~~Fix #12256~~

Code refactoring.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -7322,6 +7322,7 @@ static const QuoteParams quotes[] =
 	{TEXT("Anonymous #195"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("Why do programmers always mix up Halloween and Christmas?\nBecause Oct 31 == Dec 25\n") },
 	{TEXT("Anonymous #196"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("What happened to the function that ran away?\nIt never returned.\n") },
 	{TEXT("Anonymous #197"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("When I am tasked with sorting through a stack of résumés, I throw about half of them in the garbage.\nI do not want unlucky people working in our company.\n") },
+	{TEXT("Anonymous #198"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("The reason why we write SQL commands all in CAPITAL letters is because it stands for Screaming Query Language.\n") },
 	{TEXT("xkcd"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("Never have I felt so close to another soul\nAnd yet so helplessly alone\nAs when I Google an error\nAnd there's one result\nA thread by someone with the same problem\nAnd no answer\nLast posted to in 2003\n\n\"Who were you, DenverCoder9?\"\n\"What did you see?!\"\n\n(ref: https://xkcd.com/979/)") },
 	{TEXT("A developer"), QuoteParams::slow, false, SC_CP_UTF8, L_TEXT, TEXT("No hugs & kisses.\nOnly bugs & fixes.") },
 	{TEXT("Elon Musk"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("Don't set your password as your child's name.\nName your child after your password.") },
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -561,9 +561,13 @@ private :
 struct NewDocDefaultSettings final
 {
 	EolType _format = EolType::osdefault;
+
 	UniMode _unicodeMode = uniCookie;
 	bool _openAnsiAsUtf8 = true;
+
 	LangType _lang = L_TEXT;
+	//bool _openFileAsDefaultLang = true; // open file with the language non-identified as default language
+
 	int _codepage = -1; // -1 when not using
 };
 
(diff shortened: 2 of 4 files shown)


ArkadiuszMichalski:
@donho How this fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12256? It's a request to be able to do this. Anyway last changes cause that default language applied only on new document, so you make the same in a slightly different way? Can you show some STR to distinguish behavior from this PR from the previous one?

donho:
@ArkadiuszMichalski 
> Can you show some STR to distinguish behavior from this PR from the previous one?

**Before the PR**
1. Set Default Language to Ada.
2. Create a new file by using "New File..." menu entry.
3. New document is set to Ada. **(Expected ehaviour)**
4. Open a empty file, its file name has no extension.
5. The file is set to Ada. **(Non-expected ehaviour)**
 

**After the PR**
1. Set Default Language to Ada.
2. Create a new file by using "New File..." menu entry.
3. New document is set to Ada. **(Expected ehaviour)**
4. Open a empty file, its file name has no extension.
5. The file is set to Normal Text. **(Expected ehaviour)**


#12256 was created due to the inconsistency of **New Document -> Default Language** application. This PR restore its consistency.  Of course if there are requests from users for applying **Default Language** to opened existing files in which the language is Normal Text, #12256 can be reopened, and the feature shown on the image can be implemented.

But before that, the consistency of the behaviour is maintained. The UI is kept clean as it is (also the code), and I don't really think it's a demanded feature.

ArkadiuszMichalski:
@donho But this behavior `After the PR` I get also with actual code (without this PR), I mean step 4/5. This was fixed I guess by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a610387fb40411f992f1fea38c9e7eaa7cf5c8cc. I'am asking for the difference in visible behavior from the user's perspective, not how it was achieved in the code (because it can be achieved in different ways). I still don't see the difference, compared to this last build:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/44996878

donho:
@ArkadiuszMichalski 
You're absolutely right!
Indeed the behaviour of this PR is the same as before - This PR is rather a refactoring.  
I was mistaken. Sorry for the confusion.

ArkadiuszMichalski:
@donho Why do this?
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/cbb13f51feadde89207d01dbad950793a4173a86/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1180-L1188

Try with `test.php` with `<?xml version="1.0" encoding="utf-8"?>` as content. When we open we get PHP lang, but if we change file and save it then our PHP lang will change to XML. Combining on these detections can introduce this unexpected behavior (if we don't check everything carefully).

donho:
@ArkadiuszMichalski 
Nice catched. Fixed.

ArkadiuszMichalski:
@donho Another thing, when we save buffer we should not do similar logic as in:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3fae88f635087e1f63d7e71e68df6fcbe9285515/PowerEditor/src/ScintillaComponent/Buffer.cpp#L744 

STR:
1. Open some big file (>200MB) let say HTML. It will be treat as Plain Text.
2. Put on first line `<?xml version="1.0" encoding="utf-8"?>`
3. Save file. Our Plain Text will be change to XML.

So `_isLargeFile` should not be also check here?

donho:
> So _isLargeFile should not be also check here?

@ArkadiuszMichalski 
Nice caught, again! And thanks to your remark, save file routine has even been optimized for saving big files.

> But we have to take into account that before when we inovke save the buffer can go from large to small or vice versa

Indeed, such scenario is possible.

> Maybe we have to once again determine the current buffer size and set _isLargeFile flag. This can be considered in a separate issue.

The above problem not so critical so I guess we can ignore it for the moment - of course we can treat it separately if such scenario happens to a lot of ppl.

Please check the new commit and feel free to tell me if you find other problems.

ArkadiuszMichalski:
@donho
If `detectLanguageFromTextBegining()` when saving is dependent on `_isLargeFile` why not do the same when reading file?
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/afd876657501826029640fee6ed843fbc02a69e8/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1527-L1530

If the file is too big we will not use the result of this work anyway:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/afd876657501826029640fee6ed843fbc02a69e8/PowerEditor/src/ScintillaComponent/Buffer.cpp#L744

donho:
@ArkadiuszMichalski 
> If detectLanguageFromTextBegining() when saving is dependent on _isLargeFile why not do the same when reading file?

I did check the loadFile routine before committing the [Not detecte lang from content for big file](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12295/commits/afd876657501826029640fee6ed843fbc02a69e8), and didn't find the way to optimize this part.

Thanks to your comment I checked it again - the latest commit optimizes also the loadFile for the large files.
Thank you for checking the PR and providing all the pertinent suggestions!

ArkadiuszMichalski:
@donho Everything seems to be working fine, at least with my test files.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a26599794cfd5bc7c6754d240e7563f24b90e4cb

Issue #12284, opened 2022-10-04
"Go To" dialog does not update their values when current document is changed


ErnestLB:
### Description of the Issue
The "Go To" dialog does not update the values of "You are here:" and "You can't go further than:" when current document is changed by clicking in a tab of the other opened document.

### Steps to Reproduce the Issue
1. Open 2 documents in Notepad++ (in the current opened document it should appear 2 tabs)
2. Press Ctrl+G to open the "Go To" dialog, the values that appear in that dialog correspond to the first document (that is OK)
3. Press the second tab, to activate the second document; in the background the content is updated, but the values of the dialog still display the values of the first document.

### Expected Behavior
When you change the current document the values of "You are here:" and "You can't go further than:" must be updated with the values of the current document.

### Actual Behavior
When you change the current document the values of "You are here:" and "You can't go further than:" are displaying the values of the document that opened the dialog

NOTE: the values are properly updated when the dialog is clicked

Anklebitter:
Ok, I've managed to reproduce your error, however, I found that if I click the "Go To" dialog after changing tabs that the content then updates. Can you confirm the same on your end?

ErnestLB:
Yes, same behavior.
When the dialog is "printed" again (by cliking anywhere in the dialog or moving the dialog), the values are recalculated.
But losing or recovering the focus, does not recalculates the values of the dialog with the active document.

Anklebitter:
Just to clarify, is the "Go To" dialogue supposed to update regardless of whether it is in focus or not each time you change tabs?

alankilborn:
>is the "Go To" dialogue supposed to update regardless of whether it is in focus or not each time you change tabs?

That falls into the area of undefined behavior.
Since it is a modeless dialog, it certainly would be nice if it did that.

Anklebitter:
Wouldn't that be part of scintilla then? Since the window selection is responsible for making the "Go To" dialog inactive, and I believe that is handled by scintilla. Also if the "Go To" dialog is inactive it's unable to update/display the new information, essentially it's frozen.

alankilborn:
No, nothing to do with Scintilla.
And, an inactive dialog window can certainly be updated at any time.

alankilborn:
@Anklebitter 

[...]

PR #12302, opened 2022-10-07
Allow "Go To" dialog to update when switching file tabs

Anklebitter:
This allows the "Go To" dialog to update "You are here" and "You can't go further than" each time a different file tab is clicked or a different line of the current file tab is clicked. #12284 
FYI: This is my first pull request, feel free to give me some feedback!

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3873,6 +3873,9 @@ void Notepad_plus::updateStatusBar()
 	_statusBar.setText(strLnColSel, STATUSBAR_CUR_POS);
 
 	_statusBar.setText(_pEditView->execute(SCI_GETOVERTYPE) ? TEXT("OVR") : TEXT("INS"), STATUSBAR_TYPING_MODE);
+	
+	//updating "Go To" dialog
+	_goToLineDlg.updateLinesNumbers();
 }
 
 void Notepad_plus::dropFiles(HDROP hdrop)
@@ -6178,6 +6181,9 @@ void Notepad_plus::notifyBufferActivated(BufferID bufid, int view)
 	}
 
 	_linkTriggered = true;
+	
+	//updating "Go To" dialog
+	_goToLineDlg.updateLinesNumbers();
 }
 
 std::vector<generic_string> Notepad_plus::loadCommandlineParams(const TCHAR * commandLine, const CmdLineParamsDTO * pCmdParams)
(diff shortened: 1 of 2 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6178,6 +6181,9 @@ void Notepad_plus::notifyBufferActivated(BufferID bufid, int view)
 	}
 
 	_linkTriggered = true;
+	
+	//updating "Go To" dialog
+	_goToLineDlg.updateLinesNumbers();
```
alankilborn, line 6186: You already added this code in `updateStatusBar` which is called just above in this same function.  So this call is unneeded.

Anklebitter: You're absolutely right, "updateStatusBar()" is called each time a different file tab is clicked. There I've updated it now.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3873,6 +3873,9 @@ void Notepad_plus::updateStatusBar()
 	_statusBar.setText(strLnColSel, STATUSBAR_CUR_POS);
 
 	_statusBar.setText(_pEditView->execute(SCI_GETOVERTYPE) ? TEXT("OVR") : TEXT("INS"), STATUSBAR_TYPING_MODE);
+	
+	//updating "Go To" dialog
```
alankilborn, line 3877: So the comment is OK, but really I think the line of code itself explains it pretty well.  In such cases maybe it is best to not use the comment.

alankilborn: Maybe it isn't the best idea to call this function BEFORE the GoTo window has been created; it seems to work but: 
image

Anklebitter: How about using 
` if (_goToLineDlg->isCreated() && _goToLineDlg->isVisible()){_goToLineDlg.updateLinesNumbers();}`
instead of
`_goToLineDlg.updateLinesNumbers();`

alankilborn: It's your PR.  :)

Anklebitter: There I just updated the PR.

Anklebitter: So, do I just wait for review at this point?

alankilborn: Yes.

alankilborn: Note that review could entail rejection of the entire idea.  :-)

Anklebitter: Ok, understood. Thanks!

PR #12310, opened 2022-10-10
Make large file limit (for styling) configurable

### LINKED ISSUE #11389: Large XML File near 7mil lines --- TAG-Highlight not working (works in v8.1.9.2 64-bit)
### Description of the Issue

Large XML File near 7mil lines --- TAG-Highlight not working (works in v8.1.9.2 64-bit)

1. open attached big XML (171neu2utm-21.zip)
2. turn on Language/XML
3. 

### Expected Behavior

Highlighting as in  v8.1.9.2 64-bit

### Actual Behavior

TAG-Highlight not working
It takes minutes after changing to XML
As a sidecar effect, if "remember current session on next startup" is active
than n++ seams to hang minutes because of session.xml tells to open file as xml

### Debug Information

Notepad++ v8.3.3   (64-bit)
Path : C:\npp\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 2009
OS Build : 19042.1586
Current ANSI codepage : 1252

It takes minutes after changing to XML to become at least the folding but no highlighting appears

bad in v8.3.3 64-bit
image

iO in v8.1.9.2 64-bit
opening, folding and highlighting within seconds!
image

File:
[171neu2utm-21.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/8251839/171neu2utm-21.zip)

### LINKED ISSUE #11670: [Feature request]: Introduce an option to allow styling of large documents.
I understand the reason why the styling of large documents is disabled by default, but I think it would be beneficial for users if there was an option for this to be overridden, because it could be that the document is only slightly larger or one don't cares if it takes longer to load.

### LINKED ISSUE #12260: Big file issue: red highlights curly braces { } doesn't show up after update v8.4.6
### Description of the Issue
Last month August 2022, I could still see the curly braces/brackets/parentheses highlighted in red color. Which greatly helped me in my Crusader Kings save files modification. However, those braces were no longer highlighted in red color whenever I clicked around the braces. I tried looking in other unofficial forums for help and fixes but was unsuccessful. I also tried clean uninstall and reinstall Notepad++. It only worked the first time I open, and unfortunately stopped highlighting again during the next usage.

1. Opening any worded or coded files containing curly braces commands, such as _always = { yes }_, and click at the curly braces.
2. It doesn't highlight in red like it used to, like how it always highlighted in the past years.

### Expected Behavior
Just like the default settings in **Style Configurator --> Global Styles --> Brace highlight style**. The curly braces were supposed to highlight in red color, sometimes highlighting the dotted lines that's connecting the opening brace.

### Actual Behavior
Curly braces won't highlight in red or any color that I customized anymore. I already attempted every option in **Brace highlight style**. No curly braces were highlighted at all.

### Debug Information
Notepad++ v8.4.6   (64-bit)
Command Line : "C:\Users\James\Documents\Paradox Interactive\Crusader Kings III\save games\Blemmyia2.ck3"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2006
Current ANSI codepage : 932
Plugins : 
    ComparePlugin (2.0.2)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

donho:
Background: because styling large file can cause the performance issue, 200 MB (and bigger) file is not applied to its language (if any) while loading.
This PR makes large file restriction for styling and for other features (smart highlighting, brace match, etc...) configurable via the section `Performance` of Preferences dialog.

Fix #11389, fix #12260, fix #11670


ArkadiuszMichalski:
@donho When session is loaded it not switching big files to Normal text (we still have remembered language). It makes no sense because such a document will not receive a style anyway (and only extends the loading time). Or it's intentional to not modify the session file?

STR
1. Open Notepad++
2. Open this file [html_whatwg.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9750000/html_whatwg.zip) (11,6MB HTML after unpack) 
3. Close Notepad++
4. Edit `config.xml` and change `stylingMaxFileSizeMB` to `10`.
5. Open Notepad++

Our file will be open as HTML but should be Normal text. It will take longer to load and we won't have any styles in the document anyway.

donho:
@ArkadiuszMichalski 
Thank you for the testing and your suggestions.

> When session is loaded it not switching big files to Normal text (we still have remembered language). It makes no sense because such a document will not receive a style anyway (and only extends the loading time).

It's fixed in the latest commit.

> Btw. I played around with this stylingMaxFileSizeMB and it's nice to have such default limit but I can see one downside: this requires a restart of Notepad++.

I'll add UI for controlling all the options. However, the maximum file length for styling can not be totally dynamic but only for the future opened file due to this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.cpp#L688

> Apart from this limit, I would prefer to have some additional command that will enable or disable this mechanism permanently (eg in the View>Use optimization next to Word wrap).

I'll see what I can do about that.

ArkadiuszMichalski:
@donho 
> I'll add UI for controlling all the options. 

I had a feeling this was just a first step but you didn't say it outright at the beginning.

>However, the maximum file length for styling can not be totally dynamic but only for the future opened file due to this line:

I know about such limit but if we have some dynamic control (via command or Prefs) we can reload file or open it once again. Thanks to this we do not have to run Notepad ++ again and we do not interfere with other files. We can just apply or omit this limit depending on the situation.

donho:
@ArkadiuszMichalski 
> I had a feeling this was just a first step but you didn't say it outright at the beginning.

At beginning I did want to make only size option with no UI.
More I dig more I see the other potential ways.

> I will wait for the UI because it will solve this issue (although less convenient than quick access to the dedicated command).

On its way (tomorrow evening I guess).

donho:
@ArkadiuszMichalski 
The UI has been added in the latest commit (I'll work on the localization tomorrow).
Could you check it if it's what you expected please?

ArkadiuszMichalski:
@donho
>Could you check it if it's what you expected please?

Yes, it's a good base for future improvements. A few more things that can be added here:
- Highlight Matching Tags (now it's permanently disabled)
- URL detection (now it's permanently disabled)
- Backup (now it's permanently disabled - but it makes sense here anyway)
- Change History (only the option to suppress change tracking for large files, without modifying the margin if they already exist, let this be controlled by the general setting. I don't know if it's possible for a particular file, but if it would improve performance, it is also worth trying).

Btw., I know that these are just permissions, not full ON / OFF switches (which are in other Preferences places), but the following behavior may be a bit confusing:
1. Open Notepad++ and open some big files.
2. Turn on `Allow Smart Highlighting` on Performances tab.
3. Select some text to get at least 2 marking.
4. Turn of `Allow Smart Highlighting` on Performances tab.

Note that the styles are still not removed. Switching such options once again (and move focus to other place) or reload file remove them. Same happens for `Allow Brace Match`. So after switching to disallowed, these options in the currently open big file do not work (but the result of their previous work remains).
Standard switches ON/OFF (like `Smart Highlighting`) remove this previous styles completely. 

Screen of the new Performances tab:
image

Ekopalypse:
Am I correct in assuming that there is no file size limit for syntax highlighting if this is NOT enabled?
If so, this also fixes #11670.

ArkadiuszMichalski:
>Am I correct in assuming that there is no file size limit for syntax highlighting if this is NOT enabled?
If so, this also fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11670.

Yes, separate switch for lexer not exist, I assume that `Enable` cover such case.

donho:
@ArkadiuszMichalski 
> Note that the styles are still not removed

Fixed in the latest commit.

> * Highlight Matching Tags (now it's permanently disabled)

Logically if there's no xml/html styling, matching tags is kinda meaningless.
But OK, why not? 
However, it's limited by the following code, so it's unnecessary finally:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/xmlMatchedTagsHighlighter.cpp#L594

> * URL detection (now it's permanently disabled)

Not while the large file is in the session and opened on startup.
I'm debugging on it to figure out how to fix it. 

> * Backup (now it's permanently disabled - but it makes sense here anyway)

image


> * Change History

It's really a big challenge because it's a global setting for all document.
I prefer to not deal with it, at least so far.

ArkadiuszMichalski:
@donho 

>Change History
It's really a big challenge because it's a global setting for all document.
I prefer to not deal with it, at least for the moment.

It's rational, and before any work, it's worth measuring the impact of turning on change history in large files, because it's possible that the profit will be minimal and only time will be lost.

This commit https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310/commits/78ed5874aa4ac1a1e9048221041c3fd635ef9d4a fix only Smart Highlighting, but the same problem exist for brace when we have match and we get color for them.

donho:
@ArkadiuszMichalski 
> This commit [78ed587](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/78ed5874aa4ac1a1e9048221041c3fd635ef9d4a) fix only Smart Highlighting, but the same problem exist for brace when we have match and we get color for them.

Oups! Fixed. Sorry for not  reading your entire comment.

Yaron10:
@donho,

Default performance settings.
Word wrap: ON.
Brace match: ON.
Smart highlighting: ON

1. If you open a large file, "Word wrap" is disabled for *all files*.
"Brace match" and "Smart highlighting" are disabled for the large file only (per tab).

2. You can still enable "Word wrap" via "View -> Word wrap" for the large file as well.

Is that the expected behavior?
Even if "Word wrap" per tab is complex, shouldn't the original setting be restored after closing the large file?

I'm not sure it's worth the extra code. It's an observation.

Thank you for your ongoing work.

Yaron10:
I've just noticed that toggling "Enable Large File Restriction" takes effect only after closing and re-opening the large file.

ArkadiuszMichalski:
Hmm right `Word wrap` is global (affect all tab). This can be considered as separate problem because it occurs in the current stable version anyway.
Right now this can be always enabled in Performance (and then manually toggled in View if necessary). Not perfect but don't break actual behavior.

>I've just noticed that toggling "Enable Large File Restriction" takes effect only after closing and re-opening the large file.

This happen because of (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310#issuecomment-1275487526):
>I'll add UI for controlling all the options. However, the maximum file length for styling can not be totally dynamic but only for the future opened file due to this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.cpp#L688

This is acceptable, the most important thing is that you don't need to restart Notepad++.

donho:
> 1. If you open a large file, "Word wrap" is disabled for _all files_.
>        "Brace match" and "Smart highlighting" are disabled for the large file only (per tab).
> 
>     2. You can still enable "Word wrap" via "View -> Word wrap" for the large file as well.
> 
> 
> Is that the expected behavior?

@Yaron10 
Yes, it's expected behaviour.
In fact, it's the same behaviour since large file restriction has been implemented.

> I've just noticed that toggling "Enable Large File Restriction" takes effect only after closing and re-opening the large file.

Yes, this part also is expected - I'll do a tip for notifying users about this behaviour.

Yaron10:
@donho & @ArkadiuszMichalski,

My concern was as follows: if you *programmatically* disable "Word wrap" for a certain reason, you should re-enable it when that reason is no longer valid.
Well, it might be a good point in concept and theory but not worth the extra code. 

Thank you.

donho:
@Yaron10 I see your point.
I will see what I can do about it.

Yaron10:
@donho,

Great. Thank you.
And as Don would write: "If it doesn't require a lot of code".  :)

Yaron10:
Default performance settings.
Smart highlighting: ON (_in Preferences -> Highlighting tab_).

Open a large file.
Select a word.
Uncheck "Enable Large File Restriction".

**Result:**
Smart highlighting is still disabled.

**Expected:**
If it's dynamically enabled on checking "Allow Smart Highlighting" (in Performance), it certainly should be enabled on unchecking "Enable Large File Restriction". 

Relevant to "Brace Match" (and probably Auto-Completion) as well.

donho:
@ArkadiuszMichalski 
> Expected:
If it's dynamically enabled on checking "Allow Smart Highlighting" (in Performance), it certainly should be enabled on unchecking "Enable Large File Restriction".

> Relevant to "Brace Match" (and probably Auto-Completion) as well.

No. The logic is as you observed and @ArkadiuszMichalski  and I've explained. OTOH, to implement what you expected is complicate and make the code more fragile.
Les tooltips are added for explain the behaviour: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c02c23b7d431fe27ebad41619df4c2b4f1ea3a9a

Yaron10:
@donho,

My last comment was not a repetition of:
> I've just noticed that toggling "Enable Large File Restriction" takes effect only after closing and re-opening the large file.

I understood that toggling _syntax-highlighting_ without closing/re-opening was complex.
But you do toggle _smart-highlighting_ on checking/unchecking "Allow Smart Highlighting" without closing/re-opening.
So, from a user point of view, it IS expected that toggling "Enable Large File Restriction" would also immediately affect _smart-highlighting_. 
"Enable Large File Restriction" is "stronger" than "Allow Smart Highlighting".

The tooltips are a nice addition.
Thank you.

donho:
@Yaron10 @ArkadiuszMichalski 
> My concern was as follows: if you programmatically disable "Word wrap" for a certain reason, you should re-enable it when that reason is no longer valid.

After checking the code, it would be necessary to change the word wrap behaviour and its whole coding logic. Therefore it won't be implemented. OTOH, I realize also that the `Allow Word Wrap` option label is not accurate: since even it's not checked (not allowed), users can always turn Word Wrap ON after the large files' loading.

I would suggest new option label **Disable Word Wrap while loading Large File** and its default value is ON. Users can turn it OFF (Word Wrap ON) and there's no ambiguity.   

What do you think?

donho:
@Yaron10 
> But you do toggle smart-highlighting on checking/unchecking "Allow Smart Highlighting" without closing/re-opening.
So, from a user point of view, it IS expected that toggling "Enable Large File Restriction" would also immediately affect smart-highlighting.
"Enable Large File Restriction" is "stronger" than "Allow Smart Highlighting".

I do understand that from users' viewpoint such behaviour doesn't make sense. However, the root of implementation limit is, as my discussion with @ArkadiuszMichalski : 
> I'll add UI for controlling all the options. However, the maximum file length for styling can not be totally dynamic but only for the future opened file due to this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.cpp#L688

Since the tooltips has been added, I consider such confusion can be reduced. Otherwise I think the best solution is remove UI - in this case, everything is clear.

ArkadiuszMichalski:
@donho I think @Yaron10 want just delete previous style when we turn off `Enable Large File Restriction` what can be achieved by modifying this condition:
https://github.com/donho/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L5022
`if (!isAllowed)`
change to:
 `if (!nppGUI._largeFileLimit._isEnabled || !isAllowed)`
Same here:
https://github.com/donho/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L5038

And send msg `IDC_CHECK_PERFORMANCE_ALLOWBRACEMATCH` and `IDC_CHECK_PERFORMANCE_ALLOWSMARTHILITE` if we turn off `Enable Large File Restriction`:
https://github.com/donho/notepad-plus-plus/blob/master/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L5003
```
if (!largeFileRestrictionEnabled){
 send this two message so they can clean up
}
```
Or explicitly check the state of these two things inside above `case IDC_CHECK_PERFORMANCE_ENABLE:`:
```
 if (isCheckedOrNot(IDC_CHECK_PERFORMANCE_ALLOWBRACEMATCH)){
    ::SendMessage(::GetParent(_hParent), NPPM_INTERNAL_CLEANBRACEMATCH, 0, 0);
}
 if (isCheckedOrNot(IDC_CHECK_PERFORMANCE_ALLOWSMARTHILITE)){
    ::SendMessage(::GetParent(_hParent), NPPM_INTERNAL_CLEANSMARTHILITING, 0, 0);
}
```

Btw, I just started learning C++ so treat me like a typical noob :)

Coming back to the naming of these options:
1. `Enable Large File Restriction` << this is always confusing to me because it turns something on, but in practice it turns somethings off. In addition, it's responsible for switch the lexer, which is not informed directly. What about write:
`Enable Large File Restriction (always use Normal text file)` , or something similar, but make it more informative.
2. If there is space let the tooltip be also pointed by `?`, because then we know that such a hint exists. Check `Delimiter` setting tab (we have such approach).
3.  `Disable Word Wrap while loading Large File` << yes, this name is more appropriate, I would even give `Disable Word Wrap globally while loading Large File`, which will further indicate the scope of this switch. If any ambiguity appear also for this option then we can add a tooltip here too. Although the `ON/OFF` for `Word wrap` in Notepad++ works only globally anyway, so maybe it's unnecessary.

@Donho The current approach is still a better solution than not having any control over these limits. If in the future there are any reports related to unclear operation of these settings, we will react in a selectively manner.

Yaron10:
@donho,

> After checking the code, it would be necessary to change the word wrap behaviour and its whole coding logic. Therefore it won't be implemented.

Thank you for looking into it.

> I would suggest new option label Disable Word Wrap while loading Large File...

Currently the four Allow settings are "exceptions" to "Enable Large File Restriction".
If you change "Word Wrap", the other three would need to be changed as well.

So it could be:

GroupBox: `Large Files Performance`.
Tip: `Syntax Highlighting, Brace Match, Auto Completion, Smart Highlighting and Word Wrap may slow performance in large files. These features can be auto-disabled on opening a large file`.

`Disable "Slow Performance" features`.
- `Disable Brace Match`.
- `Disable Auto Completion`.
- `Disable Smart Highlighting`.
- `Disable Word Wrap`.

If you add `Disable Syntax Highlighting` it would be even clearer.
But then you might have to auto-uncheck `Disable "Slow Performance" features` if the user unchecks the five features.

I'm not sure it's worth the extra work.
`Allow Word Wrap` can be interpreted as "Do not auto-disable Word Wrap".

***

> However, the maximum file length for styling can not be totally dynamic but only for the future opened file due to this line:

I might be missing something here.
`SC_DOCUMENTOPTION_STYLES_NONE` and "styling can not be totally dynamic" are about `Syntax-Highlighting`, is that correct?
I was referring to `Smart-Highlighting` which you do toggle dynamically.

ArkadiuszMichalski:
@Yaron10 Smart-Highlighting is dependent on  `isLargeFile()`, and `_isLargeFile` does not change depending on `Enable Large File Restriction` state:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fc32fbdcce371bb669c9361d62c959b1b61e33f0/PowerEditor/src/NppNotification.cpp#L906

This problem does not exist with braces and tags because they are not controlled by the UI globally. This above condition could be slightly modified to make it work properly.

Yaron10:
@ArkadiuszMichalski,

A mid-air collusion. :)

> I think @Yaron10 want just delete previous style when we turn off Enable Large File Restriction

I mean enable/disable features which are dynamically toggled on checking/unchecking the other prefs.
(And regardless of the check-state of Allow if you uncheck `Enable Large File Restriction`).

> Edit: Hmm sorry, now I see @Yaron10 raises another problem, but here it is necessary to reopen the document, the tooltip makes it clear about it.

In https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310#issuecomment-1279826809 I was indeed referring to lexing.
But I understood the explanation.

Thank you.

ArkadiuszMichalski:
@Yaron10 You can play a bit with above condition, unfortunately I have to migrate to a newer version of MinGW first.

Yaron10:
@ArkadiuszMichalski,

I don't have VS on the PC I'm using at the moment.
And using AppVeyor for that would not be convenient.

> The current approach is still a better solution than not having any control over these limits. If in the future there are any reports related to unclear operation of these settings, we will react in a selectively manner.

Personally, I've never used really large files.
But the UI would certainly help users who do.

Yaron10:
So "Clickable Links" are disabled as well.

GroupBox: `Large Files Performance`.

Tip:
```
Syntax Highlighting, Brace Match, Auto Completion, Smart Highlighting, Word Wrap and URL-detection may slow performance in large files. These features can be auto-disabled on opening a large file.

NOTE:
1. Modifying the settings here will not affect the currently opened large files. - You can close and re-open them.
2. If "Disable Word Wrap" is checked and you open a large file, "Word Wrap" will be disabled for all files. - You can re-enable it via "View -> Word Wrap".
```

`Disable "Slow Performance" features`.
- `Disable Brace Match`.
- `Disable Auto Completion`.
- `Disable Smart Highlighting`.
- `Disable Word Wrap`.

To simplify the implementation and reduce confusion, I'd suggest not to enable/disable `Brace Match` and `Smart Highlighting` dynamically.
Let it take effect on newly opened files only.

If you use that long Tip, the tooltips can be removed.

***

@ArkadiuszMichalski,

The wrong URL-detection is a nice catch.

You can reproduce it without a session as well: 
Open a large file.
Open a new document.
Move the new document to the other view.

ArkadiuszMichalski:
@Yaron10 Dynamically changing other things than lexer (and Word Wrap) need only tweak some conditions. Can you check this binary if work as you excepted (GCC 64-bit build)?
[notepad++.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9824556/notepad%2B%2B.zip)

Yaron10:
@ArkadiuszMichalski,

If you and Don think there's a real-life advantage in it, so be it.

I've tested `Smart Highlighting` with your build.

On unchecking `Enable Large File Restriction`, it works as expected.
But on checking `Enable Large File Restriction`, the marks are not auto-cleared nor can they be removed "manually".

ArkadiuszMichalski:
@Yaron10 
>But on checking Enable Large File Restriction, the marks are not auto-cleared nor can they be removed "manually".

Yes, when we switch this `Enable large...` I don't add here:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/546745fe344be6050aa207ccb36821ca3cf52bf3/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L5003

this two lines:
```
::SendMessage(::GetParent(_hParent), NPPM_INTERNAL_CLEANBRACEMATCH, 0, 0);
::SendMessage(::GetParent(_hParent), NPPM_INTERNAL_CLEANSMARTHILITING, 0, 0);
```

For brace it's works fine, but for SmartHilite it gives no results, for some reason this removal only works for previous style when we switch `Allow Smart Highlighting` and I can't figure out why.

donho:
@ArkadiuszMichalski @Yaron10 

Please check latest commit on master: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6755daf223299ffcd49a8b6ca233c79627578f05

 

>> However, the maximum file length for styling can not be totally dynamic but only for the future opened file due to this line:

> I might be missing something here.
SC_DOCUMENTOPTION_STYLES_NONE and "styling can not be totally dynamic" are about Syntax-Highlighting, is that correct?
I was referring to Smart-Highlighting which you do toggle dynamically.

My bad - I was focusing on def of _isLargeFile and didn't realize it's possible. Thank you @ArkadiuszMichalski to point it out!

I will check the URL part tomorrow - it's bed time for me now :)

donho:
With new commit (https://github.com/notepad-plus-plus/notepad-plus-plus/commit/7e080a46bb333634fed435d7c88447b6bcaaa998) it works better. Sorry for the regression.

Yaron10:
@donho,

`Smart Highlighting` and `Brace Match`are dynamically toggled on checking/unchecking `Enable Large File Restriction`.
But only if the large file is opened when `Enable Large File Restriction` is ON.
If it's OFF (_isLargeFile == false), _even those two features are not dynamically toggled_.

***

The tooltips in RTL are not positioned properly. The issue was recently fixed in other tooltips.

Thank you.

Yaron10:
@ArkadiuszMichalski,

> After a few tries, I think there is no point in making such a dynamic change. isLargeFile is determine when files is loaded and also depend on Enable Large File Restriction. After that, toggling this option no longer corrects isLargeFile. So that would need to be changed a bit here, and then the rest of https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310#issuecomment-1284020006 (making them longer). And at the end we may find that something minor is wrong anyway and only reopening the file corrects it.
>
> So I conclude that the current solution is sufficient

I've just seen your EDIT.
So you still think the extra code for dynamic toggling (before https://github.com/notepad-plus-plus/notepad-plus-plus/commit/7e080a46bb333634fed435d7c88447b6bcaaa998) is worth it?

Yaron10:
@ArkadiuszMichalski,

I did not mean to request making the code more dynamic.
On the contrary - I think that consistency (changes take effect after closing/re-opening) would reduce confusion.

But your approach is certainly legit.

ArkadiuszMichalski:
But notice that the present solution is just practical. Suppose you have limits turned on, you open some big file, work with it, and suddenly you need e.g. highlighting, parentheses or other things (which are disabled by default). You can simply switched it dynamically. You don't need to reopen this big file, you don't lose change history and other things. Nice improvement, and it really didn't require a lot of code :)

donho:
@Yaron10 
> GroupBox: `Large Files Performance`.
> 
> Tip:
> 
> ```
> Syntax Highlighting, Brace Match, Auto Completion, Smart Highlighting, Word Wrap and URL-detection may slow performance in large files. These features can be auto-disabled on opening a large file.
> 
> NOTE:
> 1. Modifying the settings here will not affect the currently opened large files. - You can close and re-open them.
> 2. If "Disable Word Wrap" is checked and you open a large file, "Word Wrap" will be disabled for all files. - You can re-e
> ```

Implemented in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12369

For all the suggestion/evolution of large file restriction, let's talk in #12369

PR #12315, opened 2022-10-10
Increase hash dialog font height

### LINKED ISSUE #12314: Too small to read Hash dialog text
### Description of the Issue

Input text and generate hash font is too small to read. This is feature is useful while generating sha256 for plugin, but hard to read the text.

1. Open MD5 or SHA256 dialog to generate hash
2. Open any option "Generate..." or "Generate from file"
3. Observe both input and output text

### Expected Behavior

Text width and height should big enough to read it properly. 

### Actual Behavior

Text is too small to read.

### Debug Information

image

SinghRajenM:
#fixes #12314 and #closes #12314

### DIFF
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -32,10 +32,7 @@ intptr_t CALLBACK HashFromFilesDlg::run_dlgProc(UINT message, WPARAM wParam, LPA
 		{
 			NppDarkMode::autoSubclassAndThemeChildControls(_hSelf);
 
-			int fontDpiDynamicalHeight = NppParameters::getInstance()._dpiManager.scaleY(13);
-			HFONT hFont = ::CreateFontA(fontDpiDynamicalHeight, 0, 0, 0, 0, FALSE, FALSE, FALSE, ANSI_CHARSET,
-				OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS, DEFAULT_QUALITY,
-				DEFAULT_PITCH | FF_DONTCARE, "Courier New");
+			HFONT hFont = createFont(TEXT("Courier New"), 9, false, _hSelf);
 
 			const HWND hHashPathEdit = ::GetDlgItem(_hSelf, IDC_HASH_PATH_EDIT);
 			const HWND hHashResult = ::GetDlgItem(_hSelf, IDC_HASH_RESULT_EDIT);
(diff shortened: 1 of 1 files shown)


PowerEditor/src/MISC/md5/md5Dlgs.cpp
```diff
@@ -32,7 +32,7 @@ intptr_t CALLBACK HashFromFilesDlg::run_dlgProc(UINT message, WPARAM wParam, LPA
 		{
 			NppDarkMode::autoSubclassAndThemeChildControls(_hSelf);
 
-			int fontDpiDynamicalHeight = NppParameters::getInstance()._dpiManager.scaleY(13);
+			int fontDpiDynamicalHeight = NppParameters::getInstance()._dpiManager.scaleY(15);
```
SinghRajenM, line 35: I feel complete font can be removed. Let it take from system default.

donho: @SinghRajenM 
Do you mean that don't call CreateFontA ?

SinghRajenM: @donho Yes.

donho: @SinghRajenM 
I do remember now why CreateFontA is called:
The default font in listbox is not proportional, hence "Courier New" was used by calling CreateFontA.
If you have other idea than CreateFontA, feel free to let me know.

SinghRajenM: @donho 
Ok. If that is the case, then I think we should should be good with little bit increased font height. May be 15 (as in this PR) or 16.

donho:
It's the way a little bit too much on my hi-resolution screen (3840 x 2160, Scale 225%):

image

PowerEditor/src/MISC/md5/md5Dlgs.cpp
donho: @SinghRajenM 

Could you HFONT `hFont = createFont(TEXT("Courier New"), 9, false, _hSelf);` instead of 
```
int fontDpiDynamicalHeight = NppParameters::getInstance()._dpiManager.scaleY(13);
HFONT hFont = ::CreateFontA(fontDpiDynamicalHeight, 0, 0, 0, 0, FALSE, FALSE, FALSE, ANSI_CHARSET,
				OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS, DEFAULT_QUALITY,
				DEFAULT_PITCH | FF_DONTCARE, "Courier New");
```
to see if the font render better on your system?
If yes, please use `createFont` instead in this PR.

SinghRajenM: Yes, it looks better with `HFONT hFont = createFont(TEXT("Courier New"), 9, false, _hSelf);`


image

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/546745fe344be6050aa207ccb36821ca3cf52bf3

PR #12322, opened 2022-10-11
Update friulian.xml cleanup from old entries

### LINKED ISSUE #11561: Redundant Entries in Localization Files (Cleanup)
### Description of the Issue
Localization files often have redundant entries, which are not found in the English or the English Customizable Localization Files. 
I have encountered quite a few in just a short while of trying to give a go at updating the Tamil Localization File.

image

This turns out to be quite an annoyance while updating Localization Files, especially of those languages which haven't been updated for a very long time, as weeding out redundant entries becomes a huge part of the work.

I am not entirely sure why such entries exist, but it appears to me that it is because of entries of earlier versions being removed because of being outdated or turning out to be unused, and other languages not being updated thereafter; since, for instance, I _was_ able to find Items 24214, 24107 and 24103 (which are absent in the latest English and English Customizable Files, but present in the current Tamil File) in the 2012 version and before.

image

Presumably, this is an issue with several languages which do not get updated regularly, and possibly somewhat discourages those trying to update these files.
I think it would be very helpful if, on event of removal of entries from the English versions, the corresponding entries in every Localization File were removed too.  This is doable (I think. I do not know how to do it myself, but it strikes me that it _would_ be quite simple to do), and would greatly help those involved in updating the Localization Files.

Current Tamil Localization File (Lines 396, 397, 398):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/tamil.xml
English Localization File from October 2012 (Lines 423, 424, 425):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2352cb8d52552da3716fbd6a1295797851ef980a/PowerEditor/installer/nativeLang/english.xml

conky77:
ref #11561

### DIFF
--- a/PowerEditor/installer/nativeLang/friulian.xml
+++ b/PowerEditor/installer/nativeLang/friulian.xml
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
+<!--
+The comments are here for explanation, it's not necessary to translate them.
+
+*** This file contains untranslated Strings: please translate them and remove this comment (it hasn't translated since 2009 or earlier) ***
+-->
 <NotepadPlus>
-    <Native-Langue name="Furlan" filename="friulian.xml" >
+    <Native-Langue name="Furlan" filename="friulian.xml" version="8.4.7">
         <Menu>
             <Main>
                 <!-- Main Menu Entries -->
@@ -12,7 +17,7 @@
                     <Item menuId="encoding" name="For&amp;mât"/>
                     <Item menuId="language" name="&amp;Lengaç"/>
                     <Item menuId="settings" name="&amp;Impostazions"/>
-
+                    <Item menuId="tools" name="To&amp;ols"/>
                     <Item menuId="macro" name="&amp;Macro"/>
                     <Item menuId="run" name="Invie"/>
                     <Item menuId="Plugins" name="&amp;Plugins"/>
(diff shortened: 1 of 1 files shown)


conky77:
@milipili I cleaned up friulian.xlm from old entries. Are you still interested in maintaining the Friulian translation of Notepad++

conky77:
I cleaned up the old entries, created the main sections, and I inserted some blank lines to compare the file better with english.xml
Let me know if it is better to remove the blank lines or if it is better to insert the English entries when a translation is missing. What is better to facilitate a new translation?

Mateos81:
I believe the PR comment shouldn't contain the "Fix" keyword if other languages are concerned by the matter?

PowerEditor/installer/nativeLang/friulian.xml
```diff
@@ -260,336 +403,1178 @@
...
+
+
+
+
+
+
+
+
+
+
```
donho, line 1574: Why all these blank lines?

donho (changes requested):
Please fill the blank lines with English original entries - if it's not translatable by you.

conky77:
> Please fill the blank lines with English original entries - if it's not translatable by you.

done.

donho:

PR #12326, opened 2022-10-12
Installer - remember "Don't use_%APPDATA%" option

### LINKED ISSUE #12202: Installer won't persist on "dont use %appdata%" selection
### Description of the Issue
The installer, has a setting called "don't use %Appdata%" which is intended, among other uses, for portable instalation.s

An update installation will not honor this previous selection: users who select this option, will need to select it each time. Other wise sill result in a lose of settings.

1.  Install n++, selecting "don't use %appdata%" settings
2.  install an update to n++
3.  "don't use %appdata%" will not be eselected

### Expected Behavior
Expected to honor  "don't use %appdata%" preference

### Actual Behavior
"don't use %appdata%" option is always uncxhecked, on any instalation

### Debug Information
This bug is 10yr old.... God, my procastination took a while to report it!

xomx:
Fixes #12202 .

Solved by adding detection of the "doLocalConf.xml" trigger-file from the previous installation.

Links to patched installers for testing in the issue.

### DIFF
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -73,6 +73,17 @@ Var Dialog
 Var NoUserDataCheckboxHandle
 Var ShortcutCheckboxHandle
 Var WinVer
+Var noUserDataChecked
+Var createShortcutChecked
+
+; The definition of "OnChange" event for checkbox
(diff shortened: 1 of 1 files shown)


xomx:
In the PR is some old code moved before the new one. This is intentional as these declarations should be at the same place as the other ones and the new code also needs one of the variables defined beforehand.

Issue #2358, opened 2016-09-28
New files are not being added in "Folder as Workspace"


miguelofbc:
### Description of the Issue

I'm working with Notepad++ using a Network Map Drive and when I open a "Folder as Workspace" the files created in the remote directory are not being showed in the "Folder as Workspace" window.

Is there any way to refresh the files inside those directories?

I was trying close Notepad++ and open the editor again and the files are being updated. But that should work without closing window.

Is this a known issue?

SinghRajenM:
@miguelofbc  please provide debug information. (Menu -> ? -> Debug Info).
Have you tried with latest version?

miguelofbc:
Do you mean 7?

miguelofbc:
Yes, I'm seeing the same bug for Notepad++ 7 version...

innocentliny:
I have same problem.
"Folder as Workspace" should automatically add new files into workspace.
This is needed.

[DebugInfo]
`Notepad++ v7.3.3   (32-bit)
Build time : Mar  8 2017 - 03:37:37
Path : D:\Downloads\Motepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 7 (64-bit)
Plugins : Explorer.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 
`

Nuadin:
I'm seeing the same issue as well, and I really need to be able to use this function for the new folder/files I added in that section. The issue only seems to have started recently, as I was able to add something about a week or two ago. Here's my debugging info as well, in case the more info the better:

[Debug Info]
`Notepad++ v7.3.3   (32-bit)
Build time : Mar  8 2017 - 03:37:37
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 
`

MetaChuh:
please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html) and migrate or reinstall your plugins if you update from version 7.5.9 or below.
if your issue still persists, and it is still of relevance, please feel free to file a new issue as easy to understand and reproducible as possible.

thank you for your help and best regards.

Issue #2357, opened 2016-09-28
Support more encodings through ICU


garydgregory:
### Description of the Issue

I need supported for the IBM-1047 encoding. This can be done with http://site.icu-project.org/ which contains a gazillion encodings.
### Steps to Reproduce the Issue

There is such option
### Expected Behavior

Under Encodings/Character sets/I'd like to see an alphabetical list of encodings, not organized by country.
### Actual Behavior

Can't do it.
### Debug Information

N/A

donho:
> Under Encodings/Character sets/I'd like to see an alphabetical list of encodings, not organized by country.

It's rather a feature request for re-organization of encoding menu, nothing to do with ICU.

Though ICU is a good project, I don't see the advantage to use it in Notepad++.
Currently there's no the same request for re-organization of encoding menu, so it won't be implemented.

PR #2349, opened 2016-09-27
Shorten some properties of the executable

vlakoff:
A long lasting nitpick!
"Notepad++ : a free (GNU) source code editor" appears in various locations (mainly "open with" menus and dialogs), whereas it should display the software name only.

windows open with
firefox open with

Issue #1451, opened 2016-02-05
Textfiles 'lost' on an external hard drive when USB knocked out


soi521:
This is half a bug report, half a feature request.

I use Notepad++ extensively and have several textfiles open all the time that I edit over long periods of time (see attached), for example fragments of code that I'm still 'working on' but sometimes I haven't touched for months. It's useful for me just to keep them all open in Notepad++ so that I can use the search-through-all-open-files feature in case I need a bit of code from here or there.

I've recently changed job and now instead of having my files all on D Drive I have to use an external hard drive (I'm in a research institute which requires this for security reasons) so essentially a lot of those textfiles have now been moved to EHD. This has brought up a problem: the EHD is attached through a standard USB port but every once in a while the thing gets knocked out sometimes without my even noticing.

Whenever the USB attachment gets knocked out, next time I open Notepad++ it tells me that half my textfiles are no longer there and would I like to keep this nonexistent file in the editor? Notepad++ diligently asks me to confirm every single file and either closes them or turns them red and I can't get out of it without confirming every one even if I reattach the EHD. If I keep all of them in the editor then the texfiles on the EHD all turn red (indicating changes to be saved). If I then find and reattach the EHD then Notepad says these have been modified by another program and offers me to reload them and this can turn them all back to blue so fine, but if I had actually modified any of those files before knocking out the USB cable then those changes are all lost and one time I lost a lot of code changes that way.

Apologies for all the detail, but I get the impression this is something that you guys haven't imagined could be happening. Could I please suggest that something be done about this? It's clever that Notepad can tell when a file has become nonexistent, but if the source was on a removable drive then perhaps it wasn't actually deleted but the connection to that resource has been lost. If it's reconnected and the date stamp of the file is earlier than the datestamp of the last modification then Notepad shouldn't ask to reload the file, I think, but offer more options?

Thanks very much for looking at this!

Toby Marthews

![lots of open files](https://cloud.githubusercontent.com/assets/8268559/12853383/b22e1c56-cc2b-11e5-809e-6826a844b13a.png)
![prompt when usb kicked out](https://cloud.githubusercontent.com/assets/8268559/12853829/2477c9cc-cc2e-11e5-87dc-ed4f4237f5e5.png)

fade2gray:
Just curious; do you mean the connecting USB cable gets physically "knocked out" or the system stops recognising the EHD for whatever reason?

soi521:
[...]

PR #12327, opened 2022-10-12
Update to scintilla 5.3.1 and lexilla 5.2.0

chcg:
- See https://www.scintilla.org/ScintillaHistory.html

[Release 5.3.1](https://www.scintilla.org/scintilla531.zip)

    Released 12 October 2022.
    Added SCI_STYLESETINVISIBLEREPRESENTATION to make it easier to edit around invisible text. This also allows representing long lexemes with a single character to provide a summarized view. [Feature #1453](https://sourceforge.net/p/scintilla/feature-requests/1453/).
    Removed NotifyLexerChanged notification from DocWatcher. This is a private interface but could be used by independent platform layers and was exposed by ScintillaDocument in the Qt implementation of ScintillaEdit.
    Draw lines more consistently in SC_PHASES_TWO and SC_PHASES_ONE modes by clipping drawing to just the line rectangle. This stops drawing some extreme ascenders, descenders and portions of indicators which may appear and then disappear depending on which lines were drawn.
    Draw SC_MARK_BAR markers underneath other markers as they often cover multiple lines for change history and other markers mark individual lines.
    Enlarge point and point top indicators and scale to be larger with larger text.
    Suppress change history background line shading when printing. [Bug #2358](https://sourceforge.net/p/scintilla/bugs/2358/).
    Make SCI_LINESCROLL more accurate when width of space not integer. [Bug #2357](https://sourceforge.net/p/scintilla/bugs/2357/).
    On Win32 implement horizontal scrolling mouse wheel. [Feature #1450](https://sourceforge.net/p/scintilla/feature-requests/1450/).
    On Win32 implement horizontal scrolling with Shift + mouse wheel. [Feature #749](https://sourceforge.net/p/scintilla/feature-requests/749/), [Feature #1451](https://sourceforge.net/p/scintilla/feature-requests/1451/).
    On Win32 ensure page and step clicks on horizontal scroll bar do not overshoot document width.
    On GTK, fix bug where there were too many or too few lines when wrapping. [Bug #2349](https://sourceforge.net/p/scintilla/bugs/2349/).

- see https://www.scintilla.org/LexillaHistory.html

[Release 5.2.0](https://www.scintilla.org/lexilla520.zip)

    Released 12 October 2022.
    PowerShell: End comment before \r carriage return so \r and \n in same SCE_POWERSHELL_DEFAULT style. [Pull request #99](https://github.com/ScintillaOrg/lexilla/pull/99).
    PowerShell: Fix character truncation bug that lead to 'ġ' styled as an operator since its low 8 bits are equal to '!'.
    R: Support hexadecimal, float exponent and number suffix. [Issue #101](https://github.com/ScintillaOrg/lexilla/issues/101).
    R: Highlight backticks. [Pull request #102](https://github.com/ScintillaOrg/lexilla/pull/102).
    R: Highlight raw strings. [Issue #100](https://github.com/ScintillaOrg/lexilla/issues/100).
    R: Optionally highlight escape sequences in strings. [Issue #100](https://github.com/ScintillaOrg/lexilla/issues/100).
    Fix early truncation from LexAccessor::GetRange and LexAccessor::GetRangeLowered. [Issue #17](https://github.com/ScintillaOrg/lexilla/issues/17).


chcg:
Likely R style needs to be updated to support:
```

val SCE_R_INFIXEOL=11
val SCE_R_BACKTICKS=12
val SCE_R_RAWSTRING=13
val SCE_R_RAWSTRING2=14
val SCE_R_ESCAPESEQUENCE=15

```

xomx:
> On Win32 implement horizontal scrolling mouse wheel. [Feature #1450](https://sourceforge.net/p/scintilla/feature-requests/1450/).
> On Win32 implement horizontal scrolling with Shift + mouse wheel. [Feature #749](https://sourceforge.net/p/scintilla/feature-requests/749/), [Feature #1451](https://sourceforge.net/p/scintilla/feature-requests/1451/).

These work for me even in the old N++ v7.9.5 (Logitech 4-way scrolling wheel mouse). Or am I missing something?

donho:
> Likely R style needs to be updated to support:

@chcg 
Are they the new added styles in R ?

ArkadiuszMichalski:
This one probably added earlier but not yet exposed in `stylers.model.xml`:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/513ee3e3feae9e43af4fcf938b2b7acd064c693b/lexilla/include/SciLexer.h#L1524

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/513ee3e3feae9e43af4fcf938b2b7acd064c693b/PowerEditor/src/stylers.model.xml#L978

donho:
@chcg 
With https://github.com/notepad-plus-plus/notepad-plus-plus/commit/513ee3e3feae9e43af4fcf938b2b7acd064c693b, we can restore Scintilla's original `Sci_Position.h` and `ScintillaStructures.h`

donho:
@ArkadiuszMichalski @chcg 
> Likely R style needs to be updated to support:

OK I see. Will add them after merging this PR into master.

donho (changes requested):
Restore Scintilla's original Sci_Position.h and ScintillaStructures.h
(https://github.com/notepad-plus-plus/notepad-plus-plus/commit/513ee3e3feae9e43af4fcf938b2b7acd064c693b)

xomx:
@donho 

> Restore Scintilla's original Sci_Position.h

Just asking - Will it not cause a problem to the apps like [NVDA](https://github.com/nvaccess/nvda) that have already adapted to this somehow? Or some such N++ plugins?

If I remember correctly, at the end the NVDA has been patched like:
```
if ((NppVer > 8.3) && (NppBinary==x64))
    "use a longlong Sci_PositionCR;"
```

rdipardo:
@xomx,
> Just asking - Will it not cause a problem to the apps like [NVDA](https://github.com/nvaccess/nvda) that have already adapted to this somehow? Or some such N++ plugins?

If they integrated db33083 as they should have, they'll be OK as long as `sizeof(Sci_PositionCR) == sizeof(Sci_Position)` in their own code bases. As far as N++ is concerned, Scintilla just provided a cleaner way of doing the same thing as before.

donho:
> Just asking - Will it not cause a problem to the apps like [NVDA](https://github.com/nvaccess/nvda) that have already adapted to this somehow? Or some such N++ plugins?

@xomx 
No idea what NVDA has been done to fix the crash from their side. For Notepad++ it's the same thing to restore Sci_PositionCR as long instead of int64 because Sci_PositionCR is not used in Notepad++ anymore.

>  If I remember correctly, at the end the NVDA has been patched like:
```
if ((NppVer > 8.3) && (NppBinary==x64))
    "use a longlong Sci_PositionCR;"
```

As @rdipardo explained, I believe they use their code base but not the header of Notepad++ to compile their code. So I don't see any impact.

chcg:
@donho From my understanding we needed the change of Sci_PositionCR to support file handling beyond 2 GB (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10921) and the followup issue https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11047#issuecomment-1021540298. I would expect that this issue will reappear and also the failing plugins which needed to be recompiled after this API change might get again problems as the size of e.g. Sci_CharacterRange changes.

Unfortunately scintilla just added the feature to support 2GB+ files with Release 5.2.3:

>  Duplicate APIs to support 64-bit document positions on Win32: SCI_GETTEXTRANGEFULL, SCI_FINDTEXTFULL, and SCI_FORMATRANGEFULL. This adds new types to Scintilla.iface which may impact downstream projects. Applications should move to these APIs from their predecessors as they will be deprecated.

after it was introduced in N++ with the scintilla interface change :-(

chcg:
@xomx Yes, I took a look at the change for 2GB+ support at: https://github.com/chcg/NPP_HexEdit/commit/9f001874ce9f7921a7cca59e2f928c25a992a9e6#diff-3f57e03793ea35906312c840396d50ce188f298f17b02094053a30565a1c9336.

With the change back of `Sci_PositionCR` to `long` that code will break after a API update with current N++ files. So maybe it is a good idea as N++ no longer uses `Sci_PositionCR` to comment it out in the header file with a appropriate comment so upcoming adaptions from plugin author are forced to take a closer look at that structures and maybe scintilla calls which need to be updated.

chcg:
@donho @xomx see https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12435 for a proposal how to restrict access to deprecated interfaces by plugin authors

donho:
> So maybe it is a good idea as N++ no longer uses Sci_PositionCR to comment it out in the header file with an appropriate comment so upcoming adaptions from plugin author are forced to take a closer look at that structures and maybe scintilla calls which need to be updated to the full versions.

Applied in the plugin template:
https://github.com/npp-plugins/plugintemplate/commit/258e5a9a672964538a891141ee4fb90c7ffe8343#diff-e509170c025c0fc18c8484437be102e88539ea1d82edf73bdcd7d04cabd340daR21-R22

I will update header files for my other plugins. Let's keep this way and see if there's any problem/regression in v8.4.7.
If you find any problem/incoherence of this modification, please let me know.

Yaron10:
Wouldn't it be better to notify plugins authors of the upcoming change and give them some time to update their code?
You can announce _now_ that the headers should be modified in _v8.4.8_.
I don't understand why so many plugins should (_again_) be incompatible with the new release.

chcg:
@Yaron10 Current plugins should still work fine if the new scintilla interface files are not used. Just in case the plugin code is updated to use them it needs to be updated to the full versions of the structs and defines. So we shouldn't see incompatible versions immediately.

Yaron10:
@chcg,

I'm not sure I understand it.
Your PythonScript, @pnedev's ComparePlus and @rdipardo's "HTML Tag" crash with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3e9413d6152db40e1cf31ecc8f2dd057451c998f.

donho:
@Yaron10 
Do you mean 3 plugins PythonScript, ComparePlus and HTML Tag crash on the current build of master but without https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3e9413d6152db40e1cf31ecc8f2dd057451c998f ?

Yaron10:
@donho,

Yes. They crash only with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3e9413d6152db40e1cf31ecc8f2dd057451c998f.

donho:
@Yaron10 
How to reproduce the crash?
I installed PythonScript and HTML Tag - no crash on startup

xomx:
@donho 
Yaron is right, STR:
- use latest master for notepad++.exe
- install ComparePlus
- in N++ create two tabs for a comparison (e.g. with a content like "this is tab1" and "this is tab2")
- use Plugins -> ComparePlus -> Compare
npp847devel_crash_at Scintilla_GetTextRange

xomx:
Problem lies in the [ScintillaStructures.h](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/include/ScintillaStructures.h), there is:
```
using PositionCR = long;

struct CharacterRange {
	PositionCR cpMin;
	PositionCR cpMax;
};

...

struct TextRange {
	CharacterRange chrg;
	char *lpstrText;
}; 
```

So in the Scintilla Editor.cxx `case Message::GetTextRange:` is used the `TextRange` with 8B `CharacterRange`.

But all the x64 N++ plugins, compiled with the patched header in between the N++ 8.3-8.4.6, will not be ABI compatible because of these use `TextRange` with 16B `CharacterRange`.

chcg:
@Yaron10 Thanks for testing it. Your right. Also Hexeditor is crashing with the API change just on activation. The N++ change of PositionCR made the plugins to automagically behave like the FULL versions without that the plugins themself have to adapt the calls and structs.

@donho I think it makes sense to give the plugin authors to adapt to the FULL versions before changing back to the original behaviour of scintilla. Additionally the expansion of ´SCI_GETSTYLEDTEXT´ to the FULL versions should be available before forcing the plugins to be adapted to FULL.

Yaron10:
@xomx & @chcg,

Thank you for confirming.

And just to complete it: `HTML Tag` crashes on `Find matching tag`.

donho:
> I think it makes sense to give the plugin authors some time and a schedule to adapt to the FULL versions before changing back to the original behaviour of scintilla. 

Indeed. I will try to push plugins authors via Plugin Template and Plugin Demo to disable Sci_PositionCR in the header files.

The incompatible ABI problem has been fixed here:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/1d98d983006463848903d9c3e25c8c69912d662d

Thank you for all of you for testing and suggestions!

Yaron10:
Thank you Don.

rdipardo:
@Yaron10,
> `HTML Tag` crashes on `Find matching tag`.

It won't be a problem in the next release: https://bitbucket.org/rdipardo/htmltag/commits/5951d5b136bd8ad1013200365be42be2ed51fcfd

Yaron10:
@rdipardo,

Thank you for your work. 

Are you planing to release v1.3.6 with the latest commit?
Do you think the backward comparability is worth the extra code?

pnedev:
Hello @Yaron10 ,

ComparePlus should be working fine in [this release v1.1.0 pre-build](https://ci.appveyor.com/project/pnedev/compareplus/builds/45256334).

BR

rdipardo:
@Yaron10
> Are you planing to release v1.3.6 with the latest commit?

Plugin List v1.5.5 [just came out][0], so there's no point in rushing things.

> Do you think the backward compatibility is worth the extra code?

Forcing people to upgrade is anti-social behaviour, IMO. The most popular version of HTML Tag is still v1.0.0, if the [weekly download counter] means anything. Should I obsolesce every version from this year, before most users have had the chance to install one?

As for "the extra code," I've been sweeping out the cruft pretty aggressively since the Free Pascal migration. The binaries are generally smaller, partly because the DLL import table stays clean without Delphi sneaking in APIs like [`NetWkstaGetInfo`] and [`RegCloseKey`]:

netapi32

advapi32

I guess these are used to remotely attest a license key or do some other telemetry . . . ?

[0]: https://github.com/notepad-plus-plus/nppPluginList/commit/41425d16839eab38cdb42e020f9459f844df8605
[weekly download counter]: https://sourceforge.net/projects/npp-plugins/files/HTMLTag/
[`RegCloseKey`]: https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regclosekey
[`NetWkstaGetInfo`]: https://learn.microsoft.com/en-us/windows/win32/api/lmwksta/nf-lmwksta-netwkstagetinfo

rdipardo:
@Yaron10,
> But once the NPP headers have been modified and users update to that version, they would _have to_ update "HTML Tag" as well.

Upstream support for the non-`FULL` messages is not ending any sooner than "(likely [Scintilla] 6.0) and removed even later (7.0?)": https://groups.google.com/g/scintilla-interest/c/mPLwYdC0-FE. Npp took ~3 years to advance [from Scintilla v4 to v5](https://github.com/notepad-plus-plus/notepad-plus-plus/commits/master/scintilla/include).

As I've [said already][0], eliminating all the non-`FULL` messages from the code base would cripple __*every single*__ 32- and 64-bit plugin that calls into Scintilla. Much better to keep them as *aliases* for the new APIs, which they effectively are since v8.3. Plugins already expect pointer-int sized returned values from the "old" messages. All that Npp has to do is copy and paste the `*FULL` callbacks where the non-`*FULL` ones used to be. 

> That's why I thought the extra checks (`HasV5Apis`) in https://bitbucket.org/rdipardo/htmltag/commits/5951d5b136bd8ad1013200365be42be2ed51fcfd might not be necessary.

The job of that exact function is to watch out for `SCI_GETTEXT`, `SCI_GETSELTEXT` or `SCI_GETCURLINE`. There's currently no plugin command that actually calls them, so your  point is basically right. But then the span of Npp versions between v8.4 and v8.4.2 all have v5 APIs, except for the full-range ones. The plugin adapts to the host application, the way things ought to go, IMO. The alternative is asking users to adapt *their* editor to *my* plugin &#x2014; the tail wagging the dog.

> Why not release the next "HTML Tag" version for `HasV5Apis` only?

Both the 32- and 64-bit build of that version would require Npp v8.4.3 as a minimum. To this day, the 32-bit version works even on unsupported systems like Windows XP. That's a lot of stabaility to throw in the garbage just because Npp couldn't be bothered copy-and-paste a few callback handlers.

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11734#issuecomment-1140113901

Yaron10:
@rdipardo,

Thank you for the detailed explanation. 👍

Issue #4560, opened 2018-06-06
Incorrect collapsing of functions (multiple languages)


alex085:
STR:
1. Create new doc
2. Select "C" from "Language" menu
3. Type the code
```
void func1()
{
	//a {
	//}
	return;
}
```
"return" is not necessary. It is placed here to make the bug more visible.
4. Click menu "View->Fold All"

I expect folding like on picture 1.
![pic1](https://user-images.githubusercontent.com/18471286/41026422-32889b24-697d-11e8-9f68-c3151a572439.png)

Actually I get the case from picture 2.
![pic2](https://user-images.githubusercontent.com/18471286/41026440-3e431fd4-697d-11e8-9f0e-d8ea2b951dd6.png)

abek42:
Similar issue with Javascript function folding.  Code example to replicate (place in an empty JS file) :
``````````
function brokenFold(arrHnd,callerFSId,fileFor){
	console.log("folding fails for this function");
	console.log("more something");
	//{thisCommentedLine: "breaks folding",removeTheLastCharacter:"on this line to fix folding" }	
	console.log("Do more something");
}

//this should be visible post fold

function dontHideMe(){
	console.log("I will be folded");
}

Notepad Version: 7.5.4 32-bit on Win10

cgeroux:
This is bugging me also... if I put a space between the "//" and the "}" it works just fine. This seems like supper buggy behavior and inconsistent with other editors I have used which support code folding (e.g. MSVS). Is it possible to make that not the default setting on Notepad++ or to change it in the settings somewhere I haven't been able to find?

bpersonick:
There are problems with all of the folding of all languages in the newer
versions of notepad++, not just in C, but in every language I have observed
very inconsistent behaviors.  For XML and VB I was able to reliably fix
folds by making sure there was always a full tab stop (or 4/5 spaces)
indent for all lines within a fold level, blank lines with no tabs/spaces
were the most common cause of the issue.

Those are built in languages like C, so it may apply to C as well.

In custom languages, the folding was not consistent with the vb/xml
findings.  I forget exactly what I had to do, but I did hit upon something
that fixed my folding issues 95% of the time by removing one of the other
formatting features I can check which later, but that is for custom langs,
and C is a built in lang.l, so try the tabbing

zufuliu:
Isn't  XML/HTML folding based on start and end tag?

[...]

PR #12329, opened 2022-10-13
add option to select explicit comment folding for c++ like languages

niekma:
As discussed in issue #4560 there have been problems with `//{` and `//}`.

The problem was fixed in pull request #10948.

In my opinion it would be nice to have an option to turn this feature on if needed, because someone might use `//{` and `//}}` as markers for explicit region folding.

(please excuse my english, i'm not a native speaker)


donho:
@niekma 
Please create an issue, describe the problem and do a new PR to reference the issue:

PR #12331, opened 2022-10-14
Installer - ARM64 CPU check for arm64-installer

### LINKED ISSUE #12320: Installers do not check for compatibility when run
### Description of the Issue

When i ran npp.8.4.6.Installer.arm64.exe it just installed it, despite the fact that my system uses amd64 cpu.

1. Check that your system uses amd/intel cpu.
2. Install https://github.com/notepad-plus-plus/notepad-plus-plus/releases/download/v8.4.6/npp.8.4.6.Installer.arm64.exe
3. notepad++ is installed, but doesn't work

### Expected Behavior

Instaler refuses to install notepad++ on unsupported architecture.
### Actual Behavior

Installer installs arm64 notepad++ on amd64 hardware.

xomx:
Fixes #12320 .

To detect ARM64 CPU running, as we cannot run ARM64 binaries on a x86/x64 CPU  (the other way around is possible - x86 on ARM64 CPU, like we use the N++ x86 NSIS installer there...).

A patched arm64-installer for testing and how it looks like in action in the issue.

Also improved a little bit the visual look of the other abort-messages (MB_ICONSTOP and newline separator).

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -172,7 +172,7 @@ updaterDone:
 
 	Call SetRoughEstimation		; This is rough estimation of files present in function copyCommonFiles
 	InitPluginsDir			; Initializes the plug-ins dir ($PLUGINSDIR) if not already initialized.
-	Call preventInstallInWin9x
+	Call preventInstall		; check unsupported OSes and CPUs
 		
 	; look for previously selected language
 	ClearErrors
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -172,7 +172,7 @@ updaterDone:
 
 	Call SetRoughEstimation		; This is rough estimation of files present in function copyCommonFiles
 	InitPluginsDir			; Initializes the plug-ins dir ($PLUGINSDIR) if not already initialized.
-	Call preventInstall		; check unsupported OSes and CPUs
+	Call checkCompatibility		; check unsupported OSes and CPUs
 		
 	; look for previously selected language
 	ClearErrors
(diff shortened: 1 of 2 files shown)


PowerEditor/installer/nppSetup.nsi
```diff
@@ -172,7 +172,7 @@ updaterDone:
 
 	Call SetRoughEstimation		; This is rough estimation of files present in function copyCommonFiles
 	InitPluginsDir			; Initializes the plug-ins dir ($PLUGINSDIR) if not already initialized.
-	Call preventInstallInWin9x
+	Call preventInstall		; check unsupported OSes and CPUs
```
donho, line 175: Please rename `preventInstall` to `checkCompability`.

donho (changes requested):
Please rename `preventInstall` to `checkCompability`.

PowerEditor/installer/nppSetup.nsi
xomx: Done.

donho (approved):

PowerEditor/installer/nsisInclude/tools.nsh
```diff
@@ -107,41 +107,55 @@ Function ExtraOptions
...
-		MessageBox MB_YESNO "This version of Notepad++ doesn't support Windows Server 2003. The installation will be aborted.$\nDo you want to go to Notepad++ download page for downloading the last version which supports this OS?" IDYES ws2003_openDlPage IDNO ws2003_goQuit
+		MessageBox MB_YESNO|MB_ICONSTOP "This version of Notepad++ doesn't support Windows Server 2003. The installation will be aborted.$\n$\nDo you want to go to Notepad++ download page for downloading the last version which supports this OS?" IDYES ws2003_openDlPage IDNO ws2003_goQuit
 ws2003_openDlPage:
 		ExecShell "open" "https://notepad-plus-plus.org/downloads/v7.9.2/"
 ws2003_goQuit:
 		Abort
 ws2003_endTest:
+
+!ifdef ARCHARM64
+	${If} ${IsNativeARM64}
```
donho, line 147: The only case that I don't test is this case, as I have no ARM device.
Let's see if ARM users complain about it - hopefully not.

xomx: Yes, I also do not have any.

Issue #10882, opened 2021-12-06
[feature request] Distinguish between "indent width" and "tab width"


jwt27:
### Description of the Issue
Currently in Notepad++, there is only a single option for both tab and indent
width.  They should be considered separate concepts.

Each language should have separate options to set the indent size (the number of
spaces inserted when you press TAB), and the width of an ASCII `TAB` character.
The "replace by spaces" option is then more accurately named "use only spaces".

GNU indentation style is a good example, and this is currently very frustrating
to use in NP++.  Pressing TAB should indent by 2 spaces, and each sequence of 8
spaces should be replaced by a single `TAB` character.  So here, "indent width"
is 2, and "tab width" is 8.

When tab and indent size are both set to the same value, and "use only spaces"
is then unchecked, pressing TAB will just insert a single `TAB` character.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12333#issuecomment-1430684990

Issue #3251, opened 2017-05-07
Can you add "tab indentation space alignment" style?


vitaliylag:
Please make option to indent with tabs and align with spaces.

How it works:
- In beginning of a string tabs are used.
- After any not tab characters spaces are used.

I know there is a plugin for that but I want native support for such important feature (and also I cannot save plugin settings anymore https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3238, please fix it).

vrubleg:
Oh, yeah, it will be a very useful feature. Now I'm making it manually, pressing spaces as many times as needed, but automatic solution would be nice.

BTW, which plugin implements such behavior?

vitaliylag:
@vrubleg, TabIndentSpaceAlign plugin

vrubleg:
Unfortunately, TabIndentSpaceAlign is buggy. Sometimes it breaks default behavior of tab. But we need just "use spaces instead of tabs if there are other symbols than tabs on the left of the current line" behavior.

It would be nice if these "smart tabs" will be implemented natively in the Notepad++.

vitaliylag:
@donho, can you see this issue? This is most important feature in text editor for program code.

vitaliylag:
up

vitaliylag:
@donho up

donho:
If I do understand your need, what you want is using only TAB key, but in some specific situation it's white space which comes out ?

vrubleg:
@donho 

When there are nothing or just tabs before cursor, the tab key should insert tabs. Examples:
→→|test
|test

When there are non-tabs before the cursor, it should insert spaces:
→→test•|=•1;
→→•|test

As an additional convenience, it could convert adjacent tabs to spaces in such situations:
→→test→|=•1;

Legend: → is a tab, • is a space, | is a cursor.

This behavior is very convenient for the case when tabs are used for indentation, and spaces are used for alignment. You can find examples here: https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces

sasumner:
IMO this behavior would have to come from Scintilla to be an "as  you type" feature.  After-the-fact conversion functions could be in Notepad++, but without core Scintilla support for the dynamic case, conversions may be of lesser importance.

I did a little bit of searching but didn't see where Scintilla has any support for tab-indent-space-align behavior.  I could certainly be wrong about that, in which case I would appreciate someone pointing that out.

[...]

Issue #3784, opened 2017-09-29
How to enable tab indent space align?


vitaliylag:
I read all settings but didn't found how to enable tab indent space align.

What I found is "Tab size" and "Replace by space". Why should I replace tab by space. It makes no sense.

If I uncheck "Replace by space" then all spaces are replaced by tabs! It doesn't make any sense too!

I don't want my tabs to be replaced by spaces and spaces to be replaced by tabs. How can I do it?

sasumner:
This is not an issue (yet).  This is a question.  Perhaps it is better to ask it here:  https://notepad-plus-plus.org/community/

vitaliylag:
Well, if I can't, can this feature be added?

Now this is issue :) (I guess)

sasumner:
Well, I was hoping to get some discussion going about it where a lot more people will see it (on that website I linked), and maybe someone will understand what you are meaning (I don't).  If it is decided to be an issue, it can always come back here with a refined explanation.

vitaliylag:
Sorry, I don't understand what you mean.

I opened the forum you said but there's no good category I can post this issue.

> what you are meaning (I don't).

I didn't understand what you didn't understand (>_<)
Didn't you understand what is "TAB INDENT SPACE ALIGN"? If you didn't you can see it in google.com.
Or didn't you understand why I didn't create a topic on that forum? (sorry for my English)

> If it is decided to be an issue, it can always come back here with a refined explanation.

I guess there is full explanation in the first post. Is it unclear? How can I make it more clear?

sasumner:
Okay, I think I understand.  You want a capability to convert your text so that only tab characters exist before the first non-whitespace character on each line, with any tab characters existing after the first non-whitespace character to be converted to spaces.  That could be a menu item--to run that conversion on existing text.  Or do you want this capability dynamically, like this is a "setting" for the current file, and as you type the above rules are enforced?  Or maybe you'd like to see  both...

vitaliylag:
>  You want a capability to convert your text

No. I want this:

When I press tab key, Notepad++ should determinate if current line contains at least one non-tab character (even space!), but only on the left side from current cursor position.
- If there is at least one non-tab character, 4 spaces should be inserted (or less, if CUR_COLON % 4 > 0).
- If there is only tab characters or no any characters, 1 tab character should be inserted.

<br><b>Examples:</b>
- AABBCC❚DDEE — 2 spaces should be inserted after pressing TAB
[...]

PR #12333, opened 2022-10-14
Distinguish between "tab size" and "indent size"

### LINKED ISSUE #3382: Allow separate tab size and indent size
### Description of the Issue
The project I work on has the coding style: 4 column indent for each indented code block, tab size = 8. While Notepad++ is my preferred editor, trying to write code in with this style in Notepad++ is more trouble than it is worth. 

### Requested Enhancement
Provide separate indent size and tab size in language settings.

jwt27:
This patch splits the current "tab size" setting into two separate settings:

* Tab size   : determines the width of the ASCII `TAB` character.
* Indent size: determines the tab stop width.

The first option defaults to 8, the latter to 4.  Additionally, "Replace by spaces" is renamed to "Use only spaces", to better reflect what it actually does.

Pressing the TAB key simply inserts enough space characters to move the cursor to the next tab stop.  But if "Use only spaces" is unchecked, and the cursor column is divisible by the tab character size, those spaces are replaced by tabs.  This doesn't even require any extra code, it is a standard Scintilla feature.

For example, with tab size 8 and indent size 4, pressing the TAB key on an empty line inserts four spaces.  Pressing it again replaces those four spaces with a single `TAB` character.

When tab size and indent size are equal, the behavior is the same as before.

Fixes: #10882, fixes #3382


jwt27:
Also fixes: #3251 #3382 #3784

jwt27:
Rebased to fix merge conflict - no changes.

jwt27:
Rebased again, and added a commit to upgrade the user's config to the new format, so the change is less disruptive.

donho:
I don't think this feature is demanded by users.
Let's see if there are more similar request coming, then I will consider it.

jwt27:
Okay, that's a bit disappointing.  I hope then that others will see this and chime in.

I wouldn't call this a "feature", though.  It's a bug fix, since the current tab/indent behavior is just wrong.

mmoult:
How many users are required for this to qualify as "demanded"? I think that @jwt27 did a good job in linking several related issues that multiple other people commented on. Furthermore, a similar feature has been implemented in Visual Studio and Visual Studio Code, so that proves that there is interest in it.

I will vouch for my interest. I agree that many current tab/indent behaviors are incorrect. I am not sure how many of them this will fix, but it is a step in the right direction.

donho:
@mmoult 
I do believe @jwt27 has done a good job for this PR, but I think the implemented feature's behaviour in this PR will confuse a lot of users.
When a user hit a TAB, he/she does need a TAB. If a user makes TAB large as 7 char, he/she gets a 7 char width of TAB. If a user want to hit TAB to be ws, he/she enabled "Replace by space" options, then he/she gets 7 ws. It's clear and non-confusion.

ArkadiuszMichalski:
@jwt27 
What is the benefit of inserting at the beginning spaces when expecting tabs? I checked this PR and I don't quite understand its usefulness. When I use tabs as indentation at the beginning, why we should put spaces after pressing tab? What would these spaces do for me?

jwt27:
Strange and rather disappointing to see that this PR got closed so suddenly.

Many text and code files - probably the vast majority - are written with tab size 8 in mind.  But you generally don't want to indent each code block by 8 characters.  Or, the code style simply mandates the use of mixed tabs and spaces (GNU style).  Such files will be all jumbled up when viewed in Notepad++, and any further editing is a *very* painful experience.

I'd like to point out that no functionality is lost by applying this patch.  When tab and indent size are set to the same value, the behavior is unchanged.

Also note that this patch does not add any special code to handle this "feature".  It only uses functionality already present: `SCI_SETTABWIDTH` and `SCI_SETINDENT`.  Clearly this is the way tabs and indents are meant to be handled in Scintilla.

jwt27:
> @mmoult I do believe @jwt27 has done a good job for this PR, but I think the implemented feature's behaviour in this PR will confuse a lot of users. When a user hit a TAB, he/she does need a TAB. If a user makes TAB large as 7 char, he/she gets a 7 char width of TAB. If a user want to hit TAB to be ws, he/she enabled "Replace by space" options, then he/she gets 7 ws. It's clear and non-confusion.

Maybe my explanation was a bit confusing.  I would invite you to try this patch for yourself - I think you'll find that it makes a lot of sense.
It works the same way as in Visual Studio.  Has anyone ever complained that that was too confusing?

jwt27:
> @jwt27 What is the benefit of inserting at the beginning spaces when expecting tabs? I checked this PR and I don't quite understand its usefulness. When I use tabs as indentation at the beginning, why we should put spaces after pressing tab? What would these spaces do for me?

The example I gave is a specific use case for a specific coding style.  You'll know it when you need it.

The more general use case is, if you only indent with 2 or 4 spaces, this patch allows `TAB` characters to still be displayed correctly (8 chars).  Currently you are forced to use the same width for both, and it ruins the layout of most files.

jwt27:
> Just some comments:
> 
> > Many text and code files - probably the vast majority - are written with tab size 8 in mind.
> 
> > ...TAB characters to still be displayed correctly (8 chars)
> 
> Where are these "facts" about `8` coming from? PLEASE tell me the root origins are not because when you fire up Windows' notepad.exe and press TAB, you obtain what appears to be 8 spaces worth of horizontal caret movement.

The use of 8-width `TAB` chars *long* pre-dates notepad.exe and Windows itself.  I wouldn't be surprised if this tradition dates back to UNIX serial terminals, or maybe even the paper teletypes of the 1800's.

I don't have any hard statistics, but I am quite confident in saying that the vast majority of text files, especially those written before the advent of GUI text editors, were meant to be displayed with 8-character-width `TAB`s.  And even then, tab width may be different from the indent size.

> Isn't the modern/accepted way of handling such things, to use "editorconfig"? Notepad++ has a plugin for that and Visual Studio builds in the handling of it, and I'm sure most other power text editors support it as well, either natively or via plugin. I haven't investigated if it can handle all of the desires above (because I'm not personally interested in those scenarios, at all), but at first blush that's what comes to mind for this type of stuff.

I've never heard of "editorconfig" and never seen anyone use it.  I had to look it up, and, as expected, it has separate `tab_width` and `indent_size` options.

As I understand this would involve writing an `.editorconfig` file in each directory where one might want to open a text file with Notepad++.  I don't see why NP++ can't just do the right thing, out of the box.

alankilborn:
>The use of 8-width TAB chars long pre-dates notepad.exe and Windows itself. I wouldn't be surprised if this tradition dates back to UNIX serial terminals, or maybe even the paper teletypes of the 1800's.  I don't have any hard statistics, but I am quite confident in saying that the vast majority of text files, especially those written before the advent of GUI text editors, were meant to be displayed with 8-character-width TABs.

It's always good to remember the past.  Mainly so you don't repeat it.  :-)  I started working in the industry in 1986, and got more heavily into Unix in 1991, thankfully leaving that behind about 5 years later.  While I don't recall a tight tie of tab characters to "8", perhaps I just didn't notice.  My point is that I've been working continuously in the industry for (gasp!) 37 years now, and I don't feel this tight association with tabs and 8.  But ok, you do...  But I think you should be careful with terms like "vast majority" -- it may be your experience, but not everyone's or even the a majority of people's...

> "editorconfig" ..., as expected, it has separate tab_width and indent_size options.

So then its use would "solve" the current issue, I presume.

>As I understand this would involve writing an .editorconfig file in each directory where one might want to open a text file with Notepad++.

Well, you _could_ do that, but the way it is really meant to work is that it looks for `.editorconfig` in the same directory as the file, and if it doesn't find one, it climbs the directory tree, looking for the file in each directory encountered until it reaches the top (i.e. root of the drive).  When it finds one, it stops and uses its settings.

So, in practice, you'd create a master config file at the top level.  If this one meets your need for everything you'll ever do, great, you're done; forget about it, move on.  But if you need refinement in certain projects, you can create an additional `.editorconfig` and place it at the top level for a project, much like git creates a `.git` folder holding settings by project...or VSCode creates a `.vscode` folder in a project...or other similar examples of such things by other tools.

>I don't see why NP++ can't just do the right thing

Notepad++ does do the right thing...but that's always the author's view of what is right.  :-)

>...out of the box

Notepad++ rarely adds any functionality natively that already exists in a plugin.  The philosophy is that that's reinvention, and it is better to focus on new functionality rather than reinvent.  There are exceptions of course, e.g. "change history" went native after existing only in plugins for a long time -- but that was different because that feature came to Scintilla (N++'s underlying editing component and a separate project), and not a lot of work was required in actual N++ code.

jwt27:
> It's always good to remember the past. Mainly so you don't repeat it. :-)

Yeah, I wouldn't use something like GNU style in my own code.  But code in such styles does exist, and sometimes you need to work with it.  Currently that is just impossible in NP++.

> I started working in the industry in 1986, and got more heavily into Unix in 1991, thankfully leaving that behind about 5 years later. While I don't recall a tight tie of tab characters to "8", perhaps I just didn't notice. My point is that I've been working continuously in the industry for (gasp!) 37 years now, and I don't feel this tight association with tabs and 8. But ok, you do... But I think you should be careful with terms like "vast majority" -- it may be your experience, but not everyone's or even the a majority of people's...

Regardless of the specific values, the point of the patch is to make a distinction between the width of an ascii `TAB` character, and how far code blocks or paragraphs are indented (eg. what happens when you press the Tab key, or insert a new line).  There are plenty of files out there where those two settings are not equal, and it should be up to the user to determine them.

> > I don't see why NP++ can't just do the right thing
> 
> Notepad++ does do the right thing...but that's always the author's view of what is right. :-)
> 
> > ...out of the box
> 
> Notepad++ rarely adds any functionality natively that already exists in a plugin. The philosophy is that that's reinvention, and it is better to focus on new functionality rather than reinvent. There are exceptions of course, e.g. "change history" went native after existing only in plugins for a long time -- but that was different because that feature came to Scintilla (N++'s underlying editing component and a separate project), and not a lot of work was required in actual N++ code.

But that is exactly the case here - this patch only exposes functionality that is already present in Scintilla.

alankilborn:
>But that is exactly the case here - this patch only exposes functionality that is already present in Scintilla.

Then I suppose the only hurdle is convincing the N++ author.  Good luck and cheers.

jwt27:
> > But that is exactly the case here - this patch only exposes functionality that is already present in Scintilla.
> 
> Then I suppose the only hurdle is convincing the N++ author. Good luck and cheers.

Thanks.  I do hope he reads this and reconsiders this PR.

PR #12335, opened 2022-10-15
Update in Brazilian Portuguese translation

Marcello-mco:
Update in Brazilian Portuguese translation for this new feature.


Marcello-mco (commented):
Updated to address "Add "Allow clickable link" option in large file restriction" change

Issue #12337, opened 2022-10-15
add option to enable explicit comment folding for c++ like languages


niekma:
### Description of the Issue
You can not use `//{` and `//}` as explicit fold markers in c++ like languages (c, c++, java, ...) anymore.

As discussed in issue #4560 there have been problems with `//{` and `//}`. The problem was fixed in pull request #10948.

In my opinion it would be nice to have an option to turn this feature on, if needed.

### Steps to Reproduce the Issue
Use `//{` and `//}` in c++ like language.

### Expected Behavior
`//{` and `//}` should be useable as explicit fold markers.

### Actual Behavior
In recent Notepad++ versions `//{` and `//}` can not be used as explicit fold markers.

victorel-petrovich:
This would be good! 
And then extended to other languages, where instead of `//` - their particular symbols for rest-of-the-line commenting

Issue #12213, opened 2022-09-22
Add possibility to change lexer settings directly in Notepad++


ArkadiuszMichalski:
Some lexers have their own options (mostly for fold/unfold, but not only). It would be nice if we could change these options directly in Notepad++ (without additional plugins like NppExec or PythonScript). Not everyone have or can install these plugins (also know how to use them).

This can probably be done in many ways, these are just two suggestions:
- via GUI in the Preferences window. Show a list of lexers, each has its own text field where we enter the appropriate settings (each option on a separate line), e.g.
```
lexer.css.less.language=1
lexer.css.scss.language=1
```
This change can be saved by using the `xml` file below.
- via `langs.model.xml / langs.xml` file. For each `<Language>` we can add additional tag for configuration, like `<settings> or <config>` or other name. It contains another tag representing a specific option, like `<option>`. So for example:
```
<Language name="css" ext="css" commentStart="/*" commentEnd="*/">
    ....
    <settings>
        <option>lexer.css.less.language=1</option>
        <option>lexer.css.scss.language=1</option>
    </settings>
    ....
```

If user has made such changes (its lexer settings are different from the default ones) it's worth adding some info about it in Debug info. Enough is listing such changed lexers, for example:
```
Modified lexers: L_JS, L_CSS
```

I will just add that there are some open issues where we can solve them by using a specific option of a given lexer. Unfortunately, in Notepad++ only one such option is provided directly via the GUI for  escape `\` SQL. Selectively adding options is pointless, it should be possible to change any of them (as above).

Any suggestions are welcome.

niekma:
May I ask why you would like to have separate areas for each lexer?
I think all options have different names, so it would be sufficient to have one place (e.g. Advanded ScintillaEditView Settings).
This could be a simple TextArea in the preferences window of the GUI, for example.

ArkadiuszMichalski:
@niekma 
>I think all options have different names, so it would be sufficient to have one place

No, names can be very different (there is no specific pattern), in addition they are also the same for different lexers. You can check this two files by searching for `execute(SCI_SETPROPERTY`:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/83ee8f09bf84980809d7b58e008314b4672566e7/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
[...]

PR #12338, opened 2022-10-15
option to enable explicit comment folding for c++ like languages

niekma:
This pull request adds the option to enable explicit comment folding for c++ like languages mentioned in issue #12337.


dfs-:
I‘ve never seen a C++ IDE do this. Is this something common or a special feature?

niekma:
It is an option the ScintillaComponent has (fold.cpp.comment.explicit). I tend to think of it as a special feature.

In previous versions of Notepad++ this option was enabled by default. Because of issue #4560 it was turned off (see pull request #10948).

I used it a lot, because it was a convenient way to fold larger blocks of code. So it would be nice to be able to turn it back on again. Because of the issue it is off by default.

donho (commented):
I'm not convinced this feature is "most wanted" feature, and  it's even a confusing feature for users. This PR will be suspended to see if there are similar requests coming.

niekma:
Sorry, did not intend to give the impression that this is a "most wanted" feature.

My I ask, if there is another way to do region folding?

ArkadiuszMichalski:
@niekma You can use plugin like NppExec or PythonScript to get the same effect.

niekma:
@ArkadiuszMichalski
I tried NppExec. The script is working, but I have to run it manually, because the setting is overwritten. I also tries NppEventExec. It is possible to set it up, but as you described in #12213 it would be nice, if it would work out of the box.
Thank you for your suggestions.

victorel-petrovich:
> I‘ve never seen a C++ IDE do this. Is this something common or a special feature?

I was using it in Geany editor (also Scintilla based).
Reliably usable whenever you need hide lines that you cannot otherwise (the "Hide Lines" , Alt+H, is non-toggable and non-persistent). 
Just wish it were generalized to any other language, using `x{`, `x}`, with `x` being their own symbol for commenting rest of lines. 
But, one step at time...

victorel-petrovich:
> I'm not convinced this feature is "most wanted" feature, and it's even a confusing feature for users. This PR will be suspended to see if there are similar requests coming.

FWIW: I was a rather beginner user of advanced editors when using Geany, and that was not confusing at all. It just needs a good name , like `Enable folding by //{ and //}`, and it will be clear. Lovers of folding will attempt anything that has "folding" in mentioned ;).

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12338#pullrequestreview-1143339187

PR #11957, opened 2022-07-31
Update swedish.xml

Regnander:
Added new lines since my previous PR in May.

PR #12345, opened 2022-10-16
Update swedish.xml

Regnander:
Added new strings and made changes to existing ones based on recent commits -  including the changes from #11957, but now on a new branch.

While testing the translation, I found that the following string isn't included in English.xml. Should I create an issue about this?

Length


donho:
> While testing the translation, I found that the following string isn't included in English.xml. Should I create an issue about this?

@Regnander Yes please

Regnander:
> Yes please

Done! #12421

PR #12352, opened 2022-10-18
Add a new message NPPM_GETBUFFERIDFROMLEXER

### LINKED ISSUE #12351: Enhancement: allow external lexer to query for buffer ID that is being lexed by one of its instances
Sometimes a lexer may need to handle notifications from Scintilla and it may also need to know which instance should be handling a particular notification. For example, a lexer that manages function names in a buffer wants to handle SCN_MODIFIED notification so it can properly update its managed function name list. Surprisingly, in current Notepad++ it is impossible for a lexer to know which lexer instance triggered the notification. Scintilla's notification doesn't contain lexer instance info, and it also doesn't work for lexer to try to cache NPP's buffer ID in Lex() method, since when Lex is called, it may be happening on a different view instead of NPP's active view.

The suggestion is to introduce a new message (e.g. NPPM_GETBUFFERIDFROMLEXER) so a lexer instance can use its own "this" pointer to query the buffer ID that is currently being lexed by it,

blu3mania:
Add a new message NPPM_GETBUFFERIDFROMLEXER for external lexer to query for buffer ID by providing lexer instance pointer. Enhanced Buffer to cache associated external lexer pointer, and FileManager to provide a lookup method to find the buffer in its list that is associated with the given lexer instance.

Without such a mechanism, it is impossible for lexer to know which instance should be used to handle messages received from Scintilla.

This closes #12351.

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -548,6 +548,10 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// void* NPPM_GETBOOKMARKID(0, 0)
 	// Returns the bookmark ID
 
+	#define NPPM_GETBUFFERIDFROMLEXER (NPPMSG + 112)
+	// LRESULT NPPM_GETBUFFERIDFROMLEXER(void* pExternalLexer)
+	// wParam: Pointer to external lexer instance
+	// Returns Buffer ID if a buffer is lexed by given lexer instance, or 0 if invalid
 
 
 	// For RUNCOMMAND_USER
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -495,6 +495,11 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return reinterpret_cast<LRESULT>(BUFFER_INVALID);
 		}
 
+		case NPPM_GETBUFFERIDFROMLEXER:
+		{
+			return reinterpret_cast<LRESULT>(MainFileManager.getBufferByExternalLexer(reinterpret_cast<void*>(wParam)));
+		}
+
 		case NPPM_GETCURRENTBUFFERID:
 		{
 			return reinterpret_cast<LRESULT>(_pEditView->getCurrentBufferID());
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -615,6 +615,18 @@ Buffer* FileManager::getBufferByIndex(size_t index)
 	return _buffers.at(index);
 }
 
+Buffer* FileManager::getBufferByExternalLexer(void* pExternalLexer)
+{
+	for (const auto& buffer : _buffers)
+	{
+		if (buffer->getExternalLexer() == pExternalLexer)
+		{
+			return buffer;
+		}
+	}
+
+	return nullptr;
+}
 
 void FileManager::beNotifiedOfBufferChange(Buffer* theBuf, int mask)
 {
@@ -715,7 +727,7 @@ BufferID FileManager::loadFile(const TCHAR* filename, Document doc, int encoding
 	loadedFileFormat._eolFormat = EolType::unknown;
 	loadedFileFormat._language = L_TEXT;
 
-	bool res = loadFileData(doc, fileSize, backupFileName ? backupFileName : fullpath, data, &UnicodeConvertor, loadedFileFormat);
+	auto [res, pExternalLexer] = loadFileData(doc, fileSize, backupFileName ? backupFileName : fullpath, data, &UnicodeConvertor, loadedFileFormat);
 
 	delete[] data;
 
@@ -736,6 +748,8 @@ BufferID FileManager::loadFile(const TCHAR* filename, Document doc, int encoding
 		LONG res = CompareFileTime(&fileNameTimestamp, &zeroTimeStamp);
 		if (res != 0) // res == 1 or res == -1
 			newBuf->_timeStamp = fileNameTimestamp;
+
+		newBuf->setExternalLexer(pExternalLexer);
 
 		_buffers.push_back(newBuf);
 		++_nbBufs;
@@ -789,7 +803,7 @@ bool FileManager::reloadBuffer(BufferID id)
 	fclose(fp);
 
 	buf->_canNotify = false;	//disable notify during file load, we don't want dirty status to be triggered
-	bool res = loadFileData(doc, fileSize, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
+	auto [res, pExternalLexer] = loadFileData(doc, fileSize, buf->getFullPathName(), data, &UnicodeConvertor, loadedFileFormat);
 	buf->_canNotify = true;
 
 	delete[] data;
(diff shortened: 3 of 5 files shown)


donho:
Rather than complicating the logic in Buffer class, you can use NPPM_GETCURRENTBUFFERID & NPPM_GETCURRENTLANGTYPE to keep the buffer pointer in your external lexer.

blu3mania:
@donho The problem is that the plugin cannot get current buffer ID, but it doesn't know which lexer instance is associated with that buffer. When there are multiple documents open, for each document there is a separate lexer instance created, and there isn't anyway for the lexer instance to reliably get the buffer ID for the document it is lexing. Calling NPPM_GETCURRENTBUFFERID inside Lex method often doesn't return the correct buffer ID during startup, and also if  second view is opened, "current" buffer may not even be the one is being lexed.

donho:
@blu3mania 
You can use `NPPN_LANGCHANGED` to monitor the change of programming language:
	`#define NPPN_LANGCHANGED (NPPN_FIRST + 11) // To notify plugins that the language in the current doc is just changed.`

blu3mania:
@donho It is not that the plugin doesn't know which NPP buffer is managed by that plugin, but that it doesn't know which lexer instance is associated to that buffer. Let's consider a very simple scenario:

- User has 2 views open. In both view, the one displayed is lexed by current plugin
- NPP would initiate 2 instances of the lexer (not the plugin itself) and passes them to Scintilla. Note that NPP itself doesn't even know which buffer is managed by which Lexer instance, that is what this PR triest o address
- User hovers over one word in the view that is not current. Scintilla sends a SCN_DWELLSTART event
- NPP forwards this event to plugin 
- Now the plugin tries to find out which lexer instance should handle this event, but it cannot. NPPM_GETCURRENTBUFFERID returns the current buffer (which is not even the buffer where the event happens) and that cannot be mapped to a lexer instance

donho:
> NPP would initiate 2 instances of the lexer (not the plugin itself)

I thought this message (NPPM_GETBUFFERIDFROMLEXER) is needed for external lexer - which also is a plugin.
What are these 2 instances of lexers then?

blu3mania:
@donho Yes, external lexer belongs to a plugin, but lexer instance != plugin instance. Plugin instance is created by the plugin's WinMain and there is always only one instance, while lexer instances are created by this line:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1526

And it's one instance per document/buffer.

donho:
@blu3mania 
>while lexer instances are created by this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1526

I use an old version of Papyrus Plugin for debugging, I set a random file to Papyrus Script, close Notepad++, and reopen it - I cannot reach the above line. I reach this line instead:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppIO.cpp#L2147
then in [setLangType](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppIO.cpp#L2163) `NPPN_LANGCHANGED` will be sent to all plugins.
Normally, if you monitor `NPPN_LANGCHANGED` the notification of idFrom is a buffer pointer. Check the member _lang - its value (87) should correspond to your lexer ID (87) provided by Notepad++.

blu3mania:
@donho Sorry it has been quite a while since I modified the code, so I forgot where the Lexer instance was created. After checking changed code in this PR and adding some debugging code, I confirmed that the place where Lexer instance gets created is https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L900

Again, the problem is not that the plugin instance cannot get the buffer ID of current document, but that it doesn't know which Lexer instance is associated to the buffer.

donho:
> Again, the problem is not that the plugin instance cannot get the buffer ID of current document, but that it doesn't know which Lexer instance is associated to the buffer.

Each lexer instance has unique (lang) ID right? The Buffer ID IS indeed the Buffer pointer in which you can find the lang ID. If the both langID matched, we know your lexer instance is associated to the buffer.
Or there's somethings I am still missing?

blu3mania:
@donho No, the lang ID is always the same for all lexer instances. It is per plugin, not per lexer instance. I think you are still thinking lexer = plugin, but it's not. There could be many lexer instances in memory from the same plugin. You should be able to easily create a simple lexer to see the problem, or if you want, you can use the Lexer code in https://github.com/blu3mania/npp-papyrus/blob/main/src/Plugin/Lexer/Lexer.cpp, and add debug code in ctor/dtor to see when a new instance is created/destroyed.

blu3mania:
@donho No, you are still mistaking a lexer instance with lexer types. In a lexer plugin, GetLexerCount() usually returns 1, indicating there is just one type of lexer. However, when there are multiple documents of the same lexer type opened in NPP, multiple instances of the same lexer class is created via CreateLexer() method. They are different instances of the exact same class type. The class is inherited from ILexer5 interface. All of these instances have the exact same lang ID, as lang ID is assigned based on GetLexerCount() from the plugin, not when a lexer instance is created in ScintillaEditView. With current NPP code, there is no way a plugin can associate buffer ID to a lexer instance.

blu3mania:
@donho And to make it super clear, when I talk about instance, I mean something like:

instance = new LexerClass() <-- this is actually done by CreateLexer() call which is called by https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L900

For every document that NPP opened that has the same lang type selected in Language menu (e.g. Papyrus Script), they are all creating the exact same LexerClass, not LexerClass1, LexerClass2, and they all have the exact same lang ID.

Hopefully this makes it clear.

donho:
> However, when there are multiple documents of the same lexer type opened in NPP, multiple instances of the same lexer class is created via CreateLexer() method. 

So if I do understand well, if there are 3 documents in Notepad++ applied to Papyrus Script, there will be 3 instances of Papyrus lexer created in NppPapyrus plugin?

blu3mania:
> So if I do understand well, if there are 3 documents in Notepad++ applied to Papyrus Script, there will be 3 instances of Papyrus lexer created in NppPapyrus plugin to manager 3 documents in Notepad++ separately?

Yes. Technically they are created by NPP directly so Papyrus plugin doesn't even know their existence. Of course they can register their instances in a static list or so, but that doesn't really help, since there still isn't a way to associate buffer IDs to them. If you check current Papyrus plugin code, you will find that I am using a somewhat reliable way to detect buffer ID. Basically when Lexer's Lex() method is called, it gets current document from both views (if applicable) using NPPM_GETCURRENTDOCINDEX, then gets the full path using NPPM_GETBUFFERIDFROMPOS and NPPM_GETFULLPATHFROMBUFFERID, then compares the file name part against the script name defined in the current document the lexer instance is lexing (because Papyrus script always has a script name defined in the code). This mechanism works most of the time, but if there are multiple documents with same file name in different directories the detection of buffer ID is still not 100%, and that is why a new message that can retrieve buffer ID from lexer's instance (i.e. "this" pointer) is preferred. If you can think of a better way I'd be happy to use it as well. Thanks.

blu3mania:
@donho Just to make sure, there won't ever be multiple threads creating lexer instances in parallel, right? With your proposed solution, a plugin will first see Lexer's ctor being called, and then receive NPPN_EXTERNALLEXERBUFFER message. If it's guaranteed there won't be a separate Lexer instantiation (on another thread for example) before NPPN_EXTERNALLEXERBUFFER message then I don't see a problem. Otherwise I'd suggest passing both the buffer ID and ILexer5 pointer as parameters in the message. Thanks.

donho:
@blu3mania 
> Just to make sure, there won't ever be multiple threads creating lexer instances in parallel, right?

I confirm that it's done in the main thread and there's no other thread to do the task like setting lexer.
I'll implement it soon and I'll ping you for the testing.
Thank you!

PR #12357, opened 2022-10-18
Using an unique term for color

Patriccollu:
Hello,

I noticed 18 occurrences of '_color_' and 16 of '_colour_' in the current `english.xml` file. In order to use a unique term for the same thing, this PR replaces all the occurrences of '_colour_' with '_color_'.

It also reinstates the `contextMenu-PluginCommands` message which was erased during latest commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/8ff003412acc9b55413040b27110bf7ef887300b.

Cheers,
Patriccollu.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -549,12 +549,12 @@ The comments are here for explanation, it's not necessary to translate them.
                 <SubDialog>
                     <Item id="2204" name="Bold"/>
                     <Item id="2205" name="Italic"/>
-                    <Item id="2206" name="Foreground colour"/>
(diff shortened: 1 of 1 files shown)


chcg:
Color is preferred in American English. Colour is preferred in British English.

donho:
There are not only in the localization file, but in the source code.
I will do it myself.

donho:
@ArkadiuszMichalski 
> I also prefer `color` because it's shorter.

Indeed.

I just checked and found out that both "colour" and "color" are using in the source code for searching the element of stylers.xml.
In order to keep retro-compatibility, both of them in the code will be kept as they are.
I will modify only the english.xml

PR #12358, opened 2022-10-18
cleanup esperanto.xml from old entries

conky77:


conky77:
Ref #11561

donho:

PR #12367, opened 2022-10-20
Fixed the lua syntax highlighter once and for all.

ccuser44:


ccuser44:
This might be handled by scintilla/scintilla, if so someone should backport this change to there too.

donho (changes requested):
Reformatting makes code review harder.
Restore all the reformatting and make your modification compact and visible.

donho:
@ccuser44 
Reformatting is not allow. Please check:

PR #12369, opened 2022-10-20
Enhance large file restriction UI

donho:
ref:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310#issuecomment-1284561430


donho:
@Yaron10
> Smart Highlighting and Brace Matchare dynamically toggled on checking/unchecking Enable Large File Restriction.
But only if the large file is opened when Enable Large File Restriction is ON.
If it's OFF (_isLargeFile == false), even those two features are not dynamically toggled.

It works to me - could you provide the instructions to reproduce it please?

Yaron10:
@donho,

The design is cleaner an clearer now. 👍 

**STR:**
Uncheck all boxes.
Open a large file.
Check `Enable Large File Restriction`.

**Result:**
`Smart Highlighting` is enabled. - So it's still not completely dynamic.

***

The tooltip in RTL is not positioned properly.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3ab519e2e8391a97be71f0e97b4089578cfdfc3a/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L4483

is OK.

Thank you.

ArkadiuszMichalski:
>Uncheck all boxes.
Open a large file.
Check Enable Large File Restriction.

But your large file is not treat by Notepad++ as large file (it's just a normal file). Smart Highlighting is working because it's ON in other Preferences setting (global switch for this option). If file is large or not is determine by two factor:
- Enable Large File Restriction is ON
- and file size is >= your file size limit

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/546745fe344be6050aa207ccb36821ca3cf52bf3/PowerEditor/src/ScintillaComponent/Buffer.cpp#L674-L676

And this is only determined once when the file is opened. There is nothing fancy here, it's just how it determines whether a file is large or not. Switching `Enable Large File Restriction` may require a large file reopen, and you should be aware of this. Do not try finding error when you switch `Enable Large File Restriction` without reopening the file because they will appear (depending on the situation when you switch it.

Yaron10:
@ArkadiuszMichalski,

> But your large file is not treat by Notepad++ as large file (it's just a normal file).

I know that. But the average user would not.

I've just pointed out that toggling dynamically some of the features (not all) under certain circumstances (not always) might confuse users.

ArkadiuszMichalski:
>I've just pointed out that toggling dynamically some of the features (not all) under certain circumstances (not always) might confuse users.

But you still only focus on this once `Enable Large File Restriction`, where the tooltip clearly indicates that switching this option requires a reopen some already opened big file to take effect. 

Of course file size can be determine more dynamically (or at least remember it's size when loading file) and checking this size if necessary, but still that won't solve the problem with disabling styles (because it's done once):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/546745fe344be6050aa207ccb36821ca3cf52bf3/PowerEditor/src/ScintillaComponent/Buffer.cpp#L692

So reopen would be needed anyway.

Yaron10:
@ArkadiuszMichalski,

> But you still only focus on this once Enable Large File Restriction...

Correct. But in regards to `Smart Highlighting` which is now dynamically toggled on checking/unchecking `Enable Large File Restriction` (in some cases).

donho:
@Yaron10 
> Correct. But in regards to `Smart Highlighting` which is now dynamically toggled on checking/unchecking `Enable Large File Restriction` (in some cases).

So what's your suggestion for such eventual confusion ?

Yaron10:
@donho,

I've suggested to remove dynamic toggling altogether. - All changes will take effect after closing/re-opening.
But @ArkadiuszMichalski doesn't like it. If you also consider the current implementation "not-confusing", let's leave it as it is.

ArkadiuszMichalski:
We are discussing some potential ambiguity here when this control has not even been implemented yet and people cannot play with it. I am not so pessimistic and believe that people can read simple tips in a tooltip.
When I having many files open, where some of them are large (and with a general restriction is ON), and wanting to have matching parentheses or highlighting (only for a moment), I don't want to be forced to find them in the maze of files and reopen them again. Since it works right now, this is an improvement. Nobody forbids anyone to reopen files after toggling each options. You can even restart Notepad++ if it's more convenient for you.

Yaron10:
The current tooltip `Modifying "Enable Large File Restriction" and "Define Large File Size" will not affect the currently opened large files.` is not in-sync with the current implementation.

Modifying "Enable Large File Restriction" can affect the currently opened large files.

**We** are getting confused. :)

ArkadiuszMichalski:
So change description to:
`Modifying "Enable Large File Restriction" and "Define Large File Size" requires re-open currently opened large files to get proper behavior.`
Because we unnecessarily focus on the description of this option, when toggling this option `[] Enable Large File Restriction` (or change size) simply requires a re-open big files. So it should be clearly pointed out and that's it.

ArkadiuszMichalski:
Backing to this @yaron, you want that permissions for large files should be determine only once (when file is loading), and after that if anything would be switch in Performance, this will not to have any effect on those opened large files, but only on future opened (or reopened)?

Then this tip will actually be correct in any case: `Modifying "Enable Large File Restriction" and "Define Large File Size" will not affect the currently opened large files.`

@donho, Something like this can also be achieved quite easily, when file is opened and is large (determine here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12369#issuecomment-1286313051), then remember for it (for buffer) actual state from Performances (which options are ON or OFF).

Then, for large files, it is enough to check the status of these options directly from buffer (not actual state form Performance):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6755daf223299ffcd49a8b6ca233c79627578f05/PowerEditor/src/ScintillaComponent/Buffer.cpp#L547-L551

Pros:
We can make different settings for different large files (which is also interesting). Just change the settings and reload the individual files.

Cons:
We won't be able to dynamically switch these options without reopen large files.
Each buffer will have to hold the states of these options when loaded

I don't know which is better or more expected. I was just wondering how to do it in a completely non-dynamic way.

donho:
@ArkadiuszMichalski 
> Something like this can also be achieved quite easily, when file is opened and is large (determine here [#12369 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12369#issuecomment-1286313051)), then remember for it (for buffer) actual state from Performances (which options are ON or OFF).

I don't think it's confusing, and in the latest commit the tooltip has been modified to be even more explicit.
So let's keep the current behaviour, and see what users say.

Thank you @Yaron1O and @ArkadiuszMichalski  both for your pertinent suggestions.

ArkadiuszMichalski:
@donho Have you check this https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310#issuecomment-1284356351? Especially this automatically opening a second view when loading a session? This is the intended behavior?

donho:
@ArkadiuszMichalski 
Thank you very much for your analyse https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12310#issuecomment-1284356351, now the bug is fixed.

Yaron10:
@donho & @ArkadiuszMichalski,

Very nice. Thank you for your work.

ArkadiuszMichalski:
@donho We can switch this URL in dynamic manner (as other options): In this part:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/cd6a6ac3bb4d46a9ddd53e3bde2002db00ac00f5/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L5055-L5057
Remove `if (isAllowed)` line. And slightly modify the order in `addHotSpot()` to include deletion parts:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/cd6a6ac3bb4d46a9ddd53e3bde2002db00ac00f5/PowerEditor/src/Notepad_plus.cpp#L3012-L3019

I test this below code, but you might as well do a separate delete in above `if (!currentBuf->allowClickableLink())` to do this earlier and return. I moved this code a bit further because I didn't want to duplicate the deleting part. I was just checking if it could work at all.

```cpp
{
	ScintillaEditView* pView = view ? view : _pEditView;
	Buffer* currentBuf = pView->getCurrentBuffer();

	int urlAction = (NppParameters::getInstance()).getNppGUI()._styleURL;
	LPARAM indicStyle = (urlAction == urlNoUnderLineFg) || (urlAction == urlNoUnderLineBg) ? INDIC_HIDDEN : INDIC_PLAIN;
	LPARAM indicHoverStyle = (urlAction == urlNoUnderLineBg) || (urlAction == urlUnderLineBg) ? INDIC_FULLBOX : INDIC_EXPLORERLINK;
	LPARAM indicStyleCur = pView->execute(SCI_INDICGETSTYLE, URL_INDIC);
	LPARAM indicHoverStyleCur = pView->execute(SCI_INDICGETHOVERSTYLE, URL_INDIC);

	if ((indicStyleCur != indicStyle) || (indicHoverStyleCur != indicHoverStyle))
	{
		pView->execute(SCI_INDICSETSTYLE, URL_INDIC, indicStyle);
		pView->execute(SCI_INDICSETHOVERSTYLE, URL_INDIC, indicHoverStyle);
		pView->execute(SCI_INDICSETALPHA, URL_INDIC, 70);
		pView->execute(SCI_INDICSETFLAGS, URL_INDIC, SC_INDICFLAG_VALUEFORE);
	}

	intptr_t startPos = 0;
	intptr_t endPos = -1;
	pView->getVisibleStartAndEndPosition(&startPos, &endPos);
	if (startPos >= endPos) return;
	pView->execute(SCI_SETINDICATORCURRENT, URL_INDIC);
	if (urlAction == urlDisable || !currentBuf->allowClickableLink())
	{
		pView->execute(SCI_INDICATORCLEARRANGE, startPos, endPos - startPos);
		return;
	}

	if (!currentBuf->allowClickableLink())
		return;

	LRESULT indicFore = pView->execute(SCI_STYLEGETFORE, STYLE_DEFAULT);
	pView->execute(SCI_SETINDICATORVALUE, indicFore);
	...
```

Now when `Enable large file` is `ON`, we open some big file, we can switch `Allow URL` and it takes effects (same as for other options). Binary to test (GCC 64-bit) [notepad++.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9844648/notepad%2B%2B.zip).

donho:
@ArkadiuszMichalski 
Nice one. I should catch it.
Do you mind to do a PR for it?
I can always fix it but since it's you who found the enhancement...

PR #12379, opened 2022-10-22
Update belarusian.xml

a-polivanchuk:
Please update the Belarusian translation from @BORNAME

### DIFF
--- a/PowerEditor/installer/nativeLang/belarusian.xml
+++ b/PowerEditor/installer/nativeLang/belarusian.xml
@@ -3,7 +3,7 @@
 The comments are here for explanation, it's not necessary to translate them.
 -->
 <NotepadPlus>
-    <Native-Langue name="Беларуская" filename="belarusian.xml" version="8.2.1">
+    <Native-Langue name="Беларуская" filename="belarusian.xml" version="8.4.6">
         <Menu>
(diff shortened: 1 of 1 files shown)


PowerEditor/installer/nativeLang/belarusian.xml
```diff
@@ -1507,7 +1603,9 @@ The comments are here for explanation, it's not necessary to translate them.
             <contextMenu-styleAlloccurrencesOfToken value="Стыль усіх супадзенняў токена" />
             <contextMenu-styleOneToken value="Стыль аднаго токена" />
             <contextMenu-clearStyle value="Ачысціць стыль" />
-            <contextMenu-PluginCommands value="Каманды ўбудовы" />
```
donho, line 1510: to be restore

donho (changes requested):
Could you please confirm that updated localization file is matched with latest updated english.xml? 
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12404

a-polivanchuk:
@donho I've pulled the latest source strings after your request and pushed the related changes in a new commit. Would it be ok for merging? Thanks!

donho (approved):

PR #12407, opened 2022-10-27
Cleanup Sardinian.xml from old entries

conky77:
Ref  #11561


donho:

PR #12414, opened 2022-10-28
Use font "Segoe UI"

kokutoukiritsugu:
make UI font display in 4k 200% dpi scale batter.
Snipaste_2022-10-29_00-44-05


MAPJe71:
And how does it display on a non 4K and/or non 200% dpi scale?

donho:
AFAIK, the font in question is not provided as default font of OS.

Issue #12421, opened 2022-10-29
The "Length" label in the Customize Maximum Length popup is not translatable


Regnander:
### Description of the Issue

There is no language file entry for the "Length" label inside the popup that appears when you click the link next to the "Customize Maximum Length" radio button.

![](https://user-images.githubusercontent.com/17113053/196057496-907f45ee-072f-492e-94c5-88e7dc02b551.png)

### Steps to Reproduce the Issue

1. Open Preferences → Recent Files History
2. Click the link next to the "Customize Maximum Length" radio button
3. Take note of the label inside the popup
4. Go to [english.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/english.xml)
5. Search for "Length:"

### Expected Behavior

I expected to find a translatable entry for said string.

### Actual Behavior

There was no such entry to be found.

ArkadiuszMichalski:
It exist:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/english.xml#L1562

xomx:
But even if the above translation is contained, the item in N++ GUI is then not translated. At least for me it behaves that way.

Regnander:
> It exist

The string you are referring to is the popup for "Max. number of entries".

![](https://user-images.githubusercontent.com/17113053/198824882-16c3f3f9-86fd-4372-9e56-e58228f1a4b3.png)

ArkadiuszMichalski:
>The string you are referring to is the popup for "Max. number of entries".
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c3ffe35492ce01419919389f69c7dfb26a3977e4/PowerEditor/installer/nativeLang/english.xml#L1089

And after changing:
`<recent-file-history-maxfile value="IT WORKS: "/>`
we get: 
![image](https://user-images.githubusercontent.com/2730894/198825429-5a3c3e6a-06d4-44b6-bc24-b74a6077d88a.png)

Make sure you used actual `english.xml` file and after changing anything switch it to other localization and back to english or restart Noptepad++.

Yaron10:
This is translatable:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c3ffe35492ce01419919389f69c7dfb26a3977e4/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L2542

[...]

PR #12429, opened 2022-10-31
Add localization for Length label in the Customize Maximum Length popup

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12421.


donho:
@ArkadiuszMichalski 
Thank you for the PR.
In order to not re-integrate all the updated translation, this PR will be merged after releasing of v8.4.7.

PR #12435, opened 2022-10-31
Scintilla interface FULL for plugin proposal

chcg:
See discussion at #12327:

- disabled versions with usage of Sci_PositionCR restricted to 2GB files on windows
- fixed wrong cast for Editor::FindTextFull
- removed dependency to Sci_PositionCR from Editor::FindText
- didn't adapt ScintillaStructures.h yet to avoid disabling/patching of functions for SCI_GETTEXTRANGE, SCI_FINDTEXT, and SCI_FORMATRANGE already now
- seems a full equivalent for https://www.scintilla.org/ScintillaDoc.html#SCI_GETSTYLEDTEXT is still missing

Assumption:
- ScintillaStructures.h is yet not/not widely used by plugins

### DIFF
--- a/scintilla/include/Sci_Position.h
+++ b/scintilla/include/Sci_Position.h
@@ -18,7 +18,9 @@ typedef ptrdiff_t Sci_Position;
 typedef size_t Sci_PositionU;
 
 // For Sci_CharacterRange  which is defined as long to be compatible with Win32 CHARRANGE
-typedef long Sci_PositionCR;
+//deprecated by N++ 2GB+ support via new scintilla interfaces from 5.2.3 (see https://www.scintilla.org/ScintillaHistory.html), 
+//please use SCI_GETTEXTRANGEFULL, SCI_FINDTEXTFULL, and SCI_FORMATRANGEFULL and corresponding defines/structs
+//typedef long Sci_PositionCR;
 
 #ifdef _WIN32
 	#define SCI_METHOD __stdcall
--- a/scintilla/include/Scintilla.h
+++ b/scintilla/include/Scintilla.h
@@ -490,9 +490,13 @@ typedef sptr_t (*SciFnDirectStatus)(sptr_t ptr, unsigned int iMessage, uptr_t wP
 #define SCFIND_REGEXP 0x00200000
 #define SCFIND_POSIX 0x00400000
 #define SCFIND_CXX11REGEX 0x00800000
-#define SCI_FINDTEXT 2150
+//deprecated by N++ 2GB+ support via new scintilla interfaces from 5.2.3 (see https://www.scintilla.org/ScintillaHistory.html), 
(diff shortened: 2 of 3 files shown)


donho:
Hmm...
[The goal of restoring `Sci_Position.h` and `ScintillaStructures.h`](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12327#issuecomment-1296020276) is for reducing the effort of maintenance - there will be less diff so it makes updating easier.

This PR makes more diff.

I don't really think it's necessary because Plugin authors don't (or rarely) update their header files from Notepad++ repository.
OTOH, Plugin template and plugin demo's header files should be updated:
https://github.com/npp-plugins/plugintemplate/commit/258e5a9a672964538a891141ee4fb90c7ffe8343#diff-e509170c025c0fc18c8484437be102e88539ea1d82edf73bdcd7d04cabd340daR21-R22

xomx:
I did not expect so many necessary modifications.
Generally I prefer not touching an included external library source code, only when absolute necessary.
So I agree with Don, let's modify only the plugin template stuff.

rdipardo:
> seems a full equivalent for https://www.scintilla.org/ScintillaDoc.html#SCI_GETSTYLEDTEXT is still missing

The next release will have one: https://sourceforge.net/p/scintilla/feature-requests/1455

I have to agree with what's been said already. Plugin code is the responsibility of the plugin writers. It's [far from ideal][0], but that's what happens when your plugins are compiled modules instead of scripts in a dynamic language like Python.


[0]: https://community.notepad-plus-plus.org/topic/23098/v8-4-2-bug/18#:~:text=Breaking%20the%20ABI%20for%20the%20sake%20of%20one%20rarely%20used%20feature%20was%20pretty%20reckless

chcg:
Ok for me.

rdipardo:
> Ok for me. So the remaining point is then to check the problematic cast within Editor::FindTextFull. I think using Sci_PositionCR there is a failure.

Given that `SCI_FINDTEXT` and co. [will be deprecated someday][0], why not just de-duplicate things by pretending the `Sci_TextToFind` structure doesn't exist anymore?

Suppose every call to `SCI_FINDTEXT` was silently redirected to `Message::FINDTEXTFULL` instead. Plugins that already expect 64-bit positions will never know the difference; they get the same sized result by sending either message. If that works, you could do the same for `SCI_GETTEXTRANGE`, `SCI_FORMATRANGE`, etc.

That seems more maintainable than lugging around a hacked version of `Sci_PositionCR` from one release to the next.

[0]: https://groups.google.com/g/scintilla-interest/c/mPLwYdC0-FE#:~:text=SCI_GETTEXTRANGE%20will%20be%20deprecated%20at%20some%20point

donho:
https://community.notepad-plus-plus.org/topic/23733/move-from-deprecated-type-symbols-structures-to-valid-ones

Issue #8715, opened 2020-08-15
JSON Default theme too bright text


vanjmali:
### Description of the Issue

The yellow color for errors in the Default theme for JSON language is too bright and makes it really hard to read when the line is selected (see image below).
I haven't touch any of the themes or languages, it's all default settings.

### Steps to Reproduce the Issue

1. Use the Default theme
2. Set language to JSON
3. Type some JSON errors, such as `[Object object]` or `blah`

### Expected Behavior

Darker font

### Actual Behavior

Bright yellow font

### Debug Information

Notepad++ v7.8.7   (64-bit)
Build time : Jun  5 2020 - 02:38:56
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19042.450
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 

![image](https://user-images.githubusercontent.com/31445914/91891680-bfc51a00-ec91-11ea-81d7-fcf07e6e2636.png)

dail8859:
The default style should look like this:

vanjmali:
Oh, my bad, I did not realize that the image contained `[Object object]` instead of the example I mentioned, the bright yellow font color appears only if text, without any quotation marks, appears in an array `[]`.
And also now realized, it's yellow because it has a red background which is not visible when the line is selected, but still an issue.

dail8859:
Ok that clarifies things now. Since the "ERROR" style is yellow text on red background, when the current line is highlighted only the yellow text shows up.

Slava7572:
не только ошибки,но и коммиты начало выделять(раньше такого точно не было).Хотя я глянул,по стилю так и должно быть только для ошибок.
<a target="_blank" href="https://radikal.ru"><img src="https://d.radikal.ru/d23/2107/21/a52259c9cd2f.png" /></a>

PR #12441, opened 2022-11-01
Change foreground color for ERROR in JSON to be more readable

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8715. The red background for the ERROR is sufficient. I check also other themes and looks like they are readable enough. The only exception is `Solarized-light.xml` but here selected text is quite dark so setting black color for ERROR wouldn't look good.


donho:
@ArkadiuszMichalski 
Do you mind to provide screenshot "Before" and "After" please?

donho:
@ArkadiuszMichalski 
Thank you for providing the screenshots.
The middle line of "BEFORE" is improved indeed in "NOW", but the 1st line in NOW is less readable.
I have tried several colors, but cannot find a suitable foreground color.
Do you have a better suggestion other than black?

ArkadiuszMichalski:
@donho
Maybe instead of the typical red background, use something brighter?
image

donho:
@ArkadiuszMichalski 
OK for Orange variant.
Thank you

ArkadiuszMichalski:
@donho 
Do you have any specific color value? Above is #FFA448 (255, 164, 72).

donho:
@ArkadiuszMichalski 
No, I don't have any specific value.
The one you show me is perfect - you can just use it:
image

PR #12454, opened 2022-11-03
Reduce AppVeyor build time

ArkadiuszMichalski:
Detail https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12457.

### DIFF
--- a/PowerEditor/installer/functionList/ada.xml
+++ b/PowerEditor/installer/functionList/ada.xml
@@ -4,7 +4,7 @@
 |   To learn how to make your own language parser, please check the following
 |   link:
 |       https://npp-user-manual.org/docs/function-list/
-|
+|   
 \=========================================================================== -->
 <NotepadPlus>
 	<functionList>
--- a/PowerEditor/installer/functionList/ada.xml
+++ b/PowerEditor/installer/functionList/ada.xml
@@ -4,7 +4,7 @@
 |   To learn how to make your own language parser, please check the following
 |   link:
 |       https://npp-user-manual.org/docs/function-list/
-|  
+|   
 \=========================================================================== -->
 <NotepadPlus>
 	<functionList>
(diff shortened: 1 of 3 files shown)


donho:
@ArkadiuszMichalski 
Have you any conclusion via this PR?
If no, could we close it? Or you need more try?

ArkadiuszMichalski:
@donho 
I've made some potential improvements but I don't know if you want to include any of them. I will summarize them in
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12457.

donho:
Please close this one then.

ArkadiuszMichalski:
@donho But https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12457 is issue for discussing (not PR). I need some PR for test or show code example. I can close this PR, but I will probably reopen it again to do some testing until we determine whether we will adopt or not any changes from this PR.

donho:
@ArkadiuszMichalski 
Please ping me when the PR is ready.

ArkadiuszMichalski:
@donho 

It's basically ready. I can run a full build (10 jobs) to be sure, but it's rather unnecessary (last commit pass `Win32 Debug`). When you merge you can open some PR for test and play with it. f necessary we can change later the syntax of annotations or message descriptions.

donho:
@ArkadiuszMichalski 
Thank you. It's merged.
Let's just see what happens without any `[FLAG]`.
Regarding `[xml]`, I am thinking about `[Translation]` - which might more explicit for the translators. What do you think?

ArkadiuszMichalski:
@donho
This validation actually could be applies to all `xml` files (localization, themes, autoCompletion, functionList), generally those covered by [validator_xml.py](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/Test/xmlValidator/validator_xml.py). So `[Translation]` is too narrow and it's longer (with one upper case). On the other hand, if you prefer this form, we can change it. We can even use both `[xml]` or `[Translation]`, which will mean the same.

In my opinion `[xml]` fits best because it's short (easy to remember and little typing) and indicates that PR/commit only modified XML files and is expected to pass validation.

donho:
OK, let's stay with only [xml].

PR #12468, opened 2022-11-06
Update galician.xml

P0rsche-911:
[galician.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9946646/galician.zip)
Updated to version 8.4.7
Apologies if I am doing something wrong. I have no experience in these issues.


donho:
Wrong path

Issue #11936, opened 2022-07-20
C++ language formatting is missing int_least*_t and uint_least*_t types.


tderensis:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e90284ea4b7917f366223bdd3d176042d0fc7249/PowerEditor/src/langs.model.xml#L113

The following types are missing:
```
int_least8_t int_least16_t int_least32_t int_least64_t uint_least8_t uint_least16_t uint_least32_t uint_least64_t
```

Adding these types would match the C language types.

PR #12477, opened 2022-11-08
Update CPP keywords

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11936.

I add some missing keywords listed in https://en.cppreference.com/w/cpp/keyword. Also update autocompletion `cpp.xml` with missing words from `instre1` and `type1` keywords list.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/langs.model.xml#L111

Still missing some keywords from above ref:
```
atomic_cancel (TM TS)
atomic_commit (TM TS)
atomic_noexcept (TM TS)
char8_t (C++20)
						
co_await (C++20)
co_return (C++20)
co_yield (C++20)
						
reflexpr (reflection TS)
synchronized (TM TS)
```

I can add them (all or some), but if yes then to which group?

Btw, are there any guidelines on how to sort autocompletion files? Does this have any negative effect in addition to the order in which it appears in the popup? Because I can see that even in this `cpp.xml` the order is not strictly "lexicographically ascending".


donho:
@ArkadiuszMichalski 

atomic_* are **instruction word** to me.

char8_t is **type word** for sure.

co_* are rather **instruction word** to me (since they are "function"):
https://en.cppreference.com/w/cpp/language/coroutines#co_await

reflexpr is a function so it's **instruction word**, whereas  synchronized define a block so it's also **instruction word** to  me.

donho:
@ArkadiuszMichalski 
>  Btw, are there any guidelines on how to sort autocompletion files? Does this have any negative effect in addition to the order in which it appears in the popup? Because I can see that even in this cpp.xml the order is not strictly "lexicographically ascending".

You're welcome to define the guideline and reorder and complete any auto-completion files.

ArkadiuszMichalski:
> You're welcome to define the guideline and reorder and complete any auto-completion files.

I will open new issue for this.

PR #12481, opened 2022-11-09
Fix GCC

ArkadiuszMichalski:


ArkadiuszMichalski:
@donho 
Looks like Make was update from 4.3 to 4.4 and maybe there are some problems with the paths.
Before:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45256475/job/wseiwt73rmr7y4r1
Now:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45321946/job/e05n8wogo69hwg8x

That's why I did a PR showing the version of the tools.

donho:
@ArkadiuszMichalski 
Thank you for checking this issue.
Curiously [Fix GCC 4](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12481/commits/9c3bf3633c04a2afc882c99085f5593c0b6004d2) has passed. Do you know why and how?

ArkadiuszMichalski:
@donho Because there was no compilation Notepad++ here, I only returned the make 4.4 file in artifacts.

appveyor.yml
```diff
@@ -109,7 +109,7 @@ for:
         Write-Output (sh --version) | select-object -first 1
         Write-Output ""
   build_script:
-    - mingw32-make NUMBER_OF_PROCESSORS=1 -f PowerEditor\gcc\makefile
+    - mingw32-make SHELL=C:/Windows/System32/cmd.exe -f PowerEditor\gcc\makefile
```
donho, line 112: Should shell be bash?

ozone10:
@ArkadiuszMichalski
try `make` instead of `mingw32-make` for 32 bit build

ArkadiuszMichalski:
@donho @ozone10 
OK, looks like using older version of Make works fine. In `Fix GCC 9` was used `4.2.1` and in `Fix GCC10` was used `4.3`.
We can use `make` command (`4.3`) for bot `32/64` bit build or only for `32` bit and in future try find why `4.4` not works for `32` bit. There were many changes: http://git.savannah.gnu.org/cgit/make.git/log/.

donho:
@ArkadiuszMichalski @ozone10 
Thank you for fixing it.

@ArkadiuszMichalski 
Could you clean this PR (ie. remove unnecessary part) ?

ArkadiuszMichalski:
So use `make` for both builds?

donho:
@ArkadiuszMichalski 
> So use make for both builds?

Yes. I think it shall pass for both.

ozone10:
@ArkadiuszMichalski @donho 
using `make` or `mingw32-make` from different path than `C:\msys64\mingw32\bin` or `C:\msys64\mingw64\bin`
will require properly set variable `MSYSTEM` like here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12450

donho:
@ozone10 
So in your opinion, #12450 should be merged firstly?

ArkadiuszMichalski:
@ozone10 This var `MSYSTEM` is not use only by MSYS launchers? In appveyor we set the paths for specific tools ourselves (don't use any launcher), so this variable seems redundant anyway.

To use make from a different location we need just run it with the full path (inside makefile we use `$(MAKE)` so it's refers to itself, not to command `make/mingw32-make` somewhere in `Path`). I used the `make` command anyway, so it will come from the paths that we set `C:\msys64\mingw32\bin` or `C:\msys64\mingw64\bin`, actually 4.3 version (it's possible that in the future they will update also this tool and the problem will rise again).
Btw, Old MinGW-64 and WinLibs don't have `make` command (only `mingw32-make` or similar variants), only MSYS2 has it, so I wasn't even aware of its existence (I don't use MSYS2).

I will update also https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12450 later, to reflect actual change. 

@donho The last commit contains a minimal set of changes to pass build.

ozone10:
@ArkadiuszMichalski 
if we use binaries from `C:\msys64\mingw32\bin` or `C:\msys64\mingw64\bin` necessary variables will be automatically set.
But when using binaries outside these location, msys2 will not know which environment variables to use. so it will use default variables which might not be ones from mingw.

@donho 
yes

donho:
@ArkadiuszMichalski 
> I will update also #12450 later, to reflect actual change.

Oups, it's already merged.
I will merge this one now. Then feel free to do other PR to make it more accurate.

BuslikDrev:
So, when you run cmd files through the Start Menu, the path will be the cmd file itself, and not the notepad++ installation path?

PR #12484, opened 2022-11-09
Disable CI 32 build in GCC temporarily

donho:
AppVeyor has updated Msys2 and that causes GCC 32 bits build faillure. The investigation of this issue is in progress, in the meantime GCC 32 bits build is disable temporarily.

ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12481

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,8 +17,8 @@ environment:
     - compiler: MSC
       platform: arm64
 
-    - compiler: GCC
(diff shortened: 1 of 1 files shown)


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f4419f2be910e8fc4dcd8148406ea40fb3550aeb

Issue #12462, opened 2022-11-03
Remove Empty Lines is almost unusable so slow


Nuklon:
### Description of the Issue
If you have a large document with several thousand (most empty, some not) lines: Use Edit > Line Operations > Remove Empty Lines, the program freezes for a long time as it takes forever to complete. A workaround is to first remove duplicate lines (which is very fast) and then remove the remaining empty line(s) manually.

### Steps to Reproduce the Issue

1. Have a large document 2000+ lines, with say 80% empty lines (mostly \r\n, some with a space).
2. Edit > Line Operations > Remove Empty Lines.
3. This takes a long, long time.

### Expected Behavior
Should be quick.

### Actual Behavior
Is slow.

### Debug Information
Notepad++ v8.3   (64-bit)
Build time : Feb  3 2022 - 03:39:05
Path : D:\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19044.2130
Current ANSI codepage : 1252
Plugins : JSMinNPP.dll mimeTools.dll NppConverter.dll NppExport.dll

ArkadiuszMichalski:
Please add some file for test where such problem could be reproduce.

Nuklon:
Here:
[bug.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9995281/bug.txt)

ArkadiuszMichalski:
@Nuklon Could you check build from this [PR](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12512) if it's faster?

Nuklon:
I downloaded the exe from Environment: compiler=MSC, platform=x64; Configuration: Release and wow, what a difference! It's now almost instantaneously. Great work 🎉

PR #12512, opened 2022-11-14
Improve Remove Empty Lines command

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12462.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1461,11 +1461,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	FindOption env;
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("^[\\t ]*$(\\r\\n|\\r|\\n)");
+		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)+|^[\\t ]*$(\\r?\\n|\\r)");
 	}
 	else
 	{
-		env._str2Search = TEXT("^$(\\r\\n|\\r|\\n)");
+		env._str2Search = TEXT("^$(\\r?\\n|\\r)(\\1)+|^$(\\r?\\n|\\r)");
 	}
 	env._str4Replace = TEXT("");
 	env._searchType = FindRegex;
@@ -1479,11 +1479,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	// remove the last line if it's an empty line.
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("(\\r\\n|\\r|\\n)^[\\t ]*$");
+		env._str2Search = TEXT("(\\r?\\n|\\r)^[\\t ]*$");
 	}
 	else
 	{
-		env._str2Search = TEXT("(\\r\\n|\\r|\\n)^$");
+		env._str2Search = TEXT("(\\r?\\n|\\r)^$");
 	}
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 }
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1461,11 +1461,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	FindOption env;
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)+|^[\\t ]*$(\\r?\\n|\\r)");
+		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)*");
 	}
 	else
 	{
-		env._str2Search = TEXT("^$(\\r?\\n|\\r)(\\1)+|^$(\\r?\\n|\\r)");
+		env._str2Search = TEXT("^$(\\r?\\n|\\r)(\\1)*");
 	}
 	env._str4Replace = TEXT("");
 	env._searchType = FindRegex;


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1461,11 +1461,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	FindOption env;
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("^[\\t ]*$(\\r\\n|\\r|\\n)");
+		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)*");
```
alankilborn, line 1464: Isn't this a better regex here?:  `^([\t ]*\R)+`

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1461,11 +1461,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	FindOption env;
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("^[\\t ]*$(\\r\\n|\\r|\\n)");
+		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)*");
 	}
 	else
 	{
-		env._str2Search = TEXT("^$(\\r\\n|\\r|\\n)");
+		env._str2Search = TEXT("^$(\\r?\\n|\\r)(\\1)*");
```
alankilborn, line 1468: Isn't this a better regex here?: `^\R+`

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1479,11 +1479,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	// remove the last line if it's an empty line.
```
alankilborn, line 1479: Why not just check the last line's content to decide to remove it, rather than running a regex replace over the entire doc?

PowerEditor/src/Notepad_plus.cpp
ArkadiuszMichalski: https://www.boost.org/doc/libs/1_80_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html
For `\R`:
>The escape sequence \R matches any line ending character sequence, specifically it is identical to the expression (?>\x0D\x0A?|[\x0A-\x0C\x85\x{2028}\x{2029}]).

It looks like it will work too. I haven't checked exactly which characters represent these, but if necessary this will also be simpler `"^([\\t ]*(\\r?\\n|\\r))+`.

PowerEditor/src/Notepad_plus.cpp
ArkadiuszMichalski: As above if necessary this will also be simpler `^(\\r?\\n|\\r)+`.

PowerEditor/src/Notepad_plus.cpp
alankilborn: Perhaps this, taking a hint from the boost docs you cited:  `^([\x09\x20]*(?>\x0D\x0A?|\x0A))+`

donho:
@ArkadiuszMichalski 
Impressive! What's the sorcery?

ArkadiuszMichalski:
@donho 
Nothing special, we just grab also the repeating empty lines too in one shot (instead of replace them one by one). This still can be simplify (to not use `(\\1)` because it's not necessary here), check https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12512#pullrequestreview-1179159987. But it also depends which notation you prefer, (`\r` or hex, or use just `\R`).

alankilborn:
@ArkadiuszMichalski said:

>or use just \R

Perhaps `\R` includes "too much" (see boost docs).  Notepad++ should limit possible line ending characters to only `\x0D` and `\0x0A`.

The whitespace in the regex could be changed to `\h` instead of the lengthier `[\t ]`, but again, perhaps that includes "too much", i.e., more than ONLY `\x20` space and `\x09` tab characters.

PowerEditor/src/Notepad_plus.cpp
alankilborn: >Can you provide some code example

I don't currently have any such example. :-(

ArkadiuszMichalski:
In detail `\R` means:

0x0A | Line Feed | LF
0x0B | Vertical Tab | VT
0x0C | Form Feed | FF
0x0D | Carriage Return | CR
0x85 | NEXT LINE | NEL
0x2028 | LINE SEPARATOR 
0x2029 | PARAGRAPH SEPARATOR

alankilborn:
Not sure why this was adopted while the discussion was still ongoing.  :-(

Consider how the accepted regex replacement `^([\t ]*$(\r?\n|\r))(\1)*` walks through the attached file by doing it interactively in the Notepad++ UI.  It makes 20 separate replacements.

This one `^([\x09\x20]*(?>\x0D\x0A?|\x0A))+` will do the complete file in 4 replacements.

[new1.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10005983/new1.txt)

alankilborn:
>In detail \R means...

Which was why I suggested to NOT use it, even though I made the original suggestion TO use it. :-)

Yaron10:
@alankilborn,

Could you please post your suggested code of `removeEmptyLine()`?

Thank you.

alankilborn:
>Could you please post your suggested code of removeEmptyLine()?

Not sure what you mean.  I was only making suggestions (which were ignored, so why did I bother) for making the regular expressions "better", not rewriting the entire function.

Yaron10:
@alankilborn,

I meant the 4 regex in that function.

> Consider how the accepted regex replacement ^([\t ]*$(\r?\n|\r))(\1)* walks through the attached file by doing it interactively in the Notepad++ UI. It makes 20 separate replacements.
>
> This one ^([\x09\x20]*(?>\x0D\x0A?|\x0A))+ will do the complete file in 4 replacements.

I don't know anything about regex, but I trust your tests and judgement. :)
 
`^([\t ]*$(\r?\n|\r))(\1)*` is not in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d53ed27bb3a568f1c8095f32c3738d69519ae98e.

Thanks again.

alankilborn:
>`^([\t ]*$(\r?\n|\r))(\1)*` is not in...the committed code.

Hmm, well it _IS_, but with the backslashes all doubled.
They have to be doubled (escaped) in the code, or the code won't compile to the correct expression.

But they should really be discussed, if they're going to be discussed, in their native (not code) form.
(If one is going to try them out interactively, it is nice to NOT have to reduce double backslashes to single, before one can try it)

ArkadiuszMichalski:
@alankilborn How do you test how many changes will be made? I use https://regex101.com/ (PCRE) and file for test  [regex.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10025169/regex.txt). But these web implementations don't necessarily mirror what's in Notepad++. Regardless of this, after a few tries, the fastest regex what I get are:
```
^[\r\n]+

^(?>[\t ]*[\r\n]+)+
```

If you have something more efficient, let me know. I will make a correction when we determine the best variant.

Btw., I also play with this `// remove the last line if it's an empty line.` around some big files, but it's fast enough. Maybe using directly  Scintilla's commands would be faster but now it's waste of time to change this (unless someone reports it in the future, we'll do it).

alankilborn:
>How do you test how many changes will be made?

Using Notepad++ itself.  Put the regular expression in the _Find what_ box and nothing in the _Replace with_ box.  Untick _Wrap around_.  Press _Replace_ continuously, walking through the file and count the number of times you pressed _Replace_ until you get to no more changes.  That's what I did, but equivalent would be _Replace All_ and then see the number of replacements it made.

alankilborn:
>Btw., I also play with this // remove the last line if it's an empty line. around some big files, but it's fast enough. Maybe using directly Scintilla's commands would be faster but now it's waste of time to change this (unless someone reports it in the future, we'll do it).

Sure, it is fine to leave it the way it is, i.e., not change to Scintilla's commands.

alankilborn:
> `^[\r\n]+`

> `^(?>[\t ]*[\r\n]+)+`

Those seem excellent.

ArkadiuszMichalski:
@alankilborn 
What should be done when we have only 1 line with only spaces/tabs (without CRLF) and we invoke `Remove Empty Lines (Containing Blank characters)`. This line should be clear or not? Now it's not clear, but that looks like a mistake to me.

alankilborn:
>What should be done when we have only 1 line with only spaces/tabs (without CRLF) and we invoke Remove Empty Lines (Containing Blank characters). This line should be clear or not? Now it's not clear, but that looks like a mistake to me.

To be technically correct, I suppose the "line" should be cleared.
How likely is it to be encountered as a use-case, though?  Not very.

PR #12519, opened 2022-11-14
Add function list parser for NppExec plugin's scripting language

### LINKED ISSUE #12517: [Feature Request] Add function list parser for NppExec plugin's scripting language
### Description of the Issue
Since an [UDL lexer for _NppExec_ plugin's scripting language is available](https://github.com/notepad-plus-plus/userDefinedLanguages/blob/master/UDLs/NppExec_themeDefault_byAndreasHeim.xml), it would be nice to have a function list parser for this language as well.

1. [Install the UDL lexer](https://npp-user-manual.org/docs/user-defined-language-system/#import-a-udl) for _NppExec_ plugin's scripting language from the link above.
2. Load the [sample file for this language](https://github.com/notepad-plus-plus/userDefinedLanguages/blob/master/UDL-samples/NppExec_byAndreasHeim.exec) with _Notepad++_. 
3. Watch function list panel.

### Expected Behavior
Function list panel should display a list of functions contained in the code of the loaded _NppExec_ script.

### Actual Behavior
Function list panel stays empty.

### Applies to
All versions of _Notepad++_.

dinkumoil:
Fixes #12517


donho:
@dinkumoil 
Could you please provide the link(s) of NppExec Script (official or 3rd party) samples so I can test your parser?

dinkumoil:
@donho 
In case you also need a _NppExec_ script lexer for doing your parser tests, have a look at [the official repo for user defined languages](https://github.com/notepad-plus-plus/userDefinedLanguages/blob/15ce8cd34c79cd2221ecbfdd6f400dd779b48cbf/UDLs/NppExec_themeDefault_byAndreasHeim.xml).

donho:
> My PR contains a sample file in order to pass automated test workflow

Yes, I saw it.
I need rather the official example or the 3rd party example to validate the parser, but not the one provided with PR.

dinkumoil:
> I need rather the official example or the 3rd party example to validate the parser, but not the one provided with PR.

Sorry, I'm not aware of any official example scripts for _NppExec_ script code. There is only the [documentation of the language](https://github.com/d0vgan/nppexec/tree/master/NppExec/doc/NppExec).

Since _NppExec_ script language is restricted to the _Notepad++_ ecosystem, I also don't think that there are any 3rd party example scripts available. At least I'm not aware of any site that hosts such scripts.

dinkumoil:
As I already mentioned, most of the _NppExec_ scripts posted at the community forum are short snippets of code that are not suitable for testing a function list parser. The simple reason is that they do not contain any jump labels.

Thus, I point you to a file I wrote by myself: https://github.com/dinkumoil/ScriptCollection/blob/master/NppExec%20Scripts/npes_saved.txt

Additionally I think I should explain something important regarding my function list parser. Its intended use is to parse the _npes_saved.txt_ file of _NppExec_ plugin. This file is not a script at all, it is a **container file** for actual scripts and can contain many scripts a user has written for his special needs.

A script starts with a line consisting of **two** colons immediately followed by the name of the script (e.g. `::MyFancyScript`). A script, in turn, can contain jump labels that are build up from only **one** colon immediately followed by the name of the jump label (e.g. `:ForLoopStart` and `:ForLoopEnd`). By using _Advanced Options_ of _NppExec_ plugin a user can create menu items in `Macro` menu for every script, thus they can be assigned to keyboard shortcuts.

My function list parser generates one node in function list tree for every single script in the _npes_saved.txt_ file. The nodes display only the names of the scripts **without** the preceeding two colons. The leaves of these nodes consists of

  * another entry with the script's name **including** the preceeding two colons in order to be able to jump to the very beginning of the script by double-clicking this entry
  * and entries for every jump label that is part of the script.

To achieve this, I decided to create a class parser for the _NppExec_ script language though there are actually no classes in this language. It was the only way to get separate nodes in function list tree for every script contained in _npes_saved.txt_ file.

donho:
@dinkumoil 
Thank you for the explanation. I've check nppExec.xml, it does work. But LABEL & goto it belongs to BATCH.
Use the indent is a smart move, but it's rather a way used only by you.

I would say it's a good way for you to structure your data, but I don't see how other users can get advantage from this PR.
You can always include this parser in your distribution for encouraging such practice (which I think interesting), but I see no reason to include it.

Issue #12399, opened 2022-10-25
Join lines should not require lines to be selected


nneil:
Pressing Ctrl-J should join the next line to the end of the current line, with a single space between them. Like `j` in vi.

Requiring the user to select the lines in this simple case is pointless.

alankilborn:
Don't wait around for it to be "fixed" as some may not think it is broken; record a macro with the following actions, save it, and assign it to Ctrl+j (removing the original Ctrl+j binding afterwards):

1) Shift+End
2) Shift+RightArrow twice
3) Ctrl+j (original _Join Lines_)

nneil:
Again this requires more work from the user. I can't ask normal people to do this. 

What's the point of having a control sequence that does nothing some of the time, when there is something useful it could be doing?

alankilborn:
>I can't ask normal people to do this.

So then I guess your option is to sit back and wait to see if your issue is decided to be important enough that it gets acted upon.
There are already a good quantity of issues that are open.
Good luck.

alankilborn:
For other interested readers that might want to emulate the desired functionality in a PythonScript, here's a short segment of script code that does it:

```
if editor.getSelections() == 1 and editor.getSelectionEmpty():
    editor.rereplace('\R', ' ', 0, editor.getCurrentPos(), editor.getLength(), 1)
else:
    notepad.menuCommand(MENUCOMMAND.EDIT_JOIN_LINES)
```

ArkadiuszMichalski:
How does this relate to the rest of the multiline commands where unchecked means all the data? Shouldn't it be the same? Alternatively could be add separate command `Join current line and below line` and maybe `Join current line and above line`.

nneil:
I don't think it's helpful to join all the lines in the file, unless I hit Ctrl-A to select them all first. If that's what you mean. I just need something like Vi's `j` command or Emac's `M-^`

I think Vi has is right though: join the current line with the next one. With a single space between the joined lines.

alankilborn:
@ArkadiuszMichalski IMO this needs to be an exception to the normal rule (because how many users ever want to join all lines in a file into one long line? [not many]).  Thus, if it is going to be an exception, it might as well function like @nneil wants.

dgebel:
I could have sworn that in previous versions NPP would join the current line with the next like this request. ????

PR #12535, opened 2022-11-18
Change for Remove Empty Lines command how operate on selection and last empty line

ArkadiuszMichalski:
For https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12545.

I looked at all operations from `Edit>Line Operations` how they work with and without selection (`Sort`, `Reverse`, `Randomize` and `Remove empty`). And all except (`Remove empty`):
1. Without selection they include the last empty line.
2. With selection they omit the last empty line.
3. For partially selected lines they automatically extend the selection to the entire lines. For sorting it's obvious. But for other cases it also makes sense because we operate on the lines and we do not have to worry about the full (correct) line selecting.

\* `Join lines` command fit only to points 2 and 3 (not work without selection, maybe should).

Taking the above into account, I also made `Remove empty` consistent with this behavior (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535/commits/4b5369e8483521c7363c614ed4aa6611505c785c). 

It needs to be clearly stated that these listed sort commands are basically a variant of one command, so other commands should not necessarily behave the same.

This PR is for testing purposes to produce different behavior variants for this `Remove Empty Lines` command.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1473,29 +1473,33 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	auto mainSelEnd = _pEditView->execute(SCI_GETSELECTIONEND);
 	auto mainSelLength = mainSelEnd - mainSelStart;
 	bool isEntireDoc = mainSelLength == 0;
-	env._isInSelection = !isEntireDoc;
+	if (!isEntireDoc)
+	{
+		env._isInSelection = !isEntireDoc;
+		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
+		auto newSelStart = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.first);
+		auto newSelEnd = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.second) + _pEditView->execute(SCI_LINELENGTH, lineRange.second);
+		_pEditView->execute(SCI_SETSEL, newSelStart, newSelEnd);
+	}
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	auto lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	size_t lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
 	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
 	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLineDoc - 1);
 	auto endPos = _pEditView->execute(SCI_GETLENGTH);
 	if (!isEntireDoc)
 	{
+		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
 		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
 		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
-		auto endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
-		if (endPos != (_pEditView->execute(SCI_POSITIONFROMLINE, endLine) + _pEditView->execute(SCI_LINELENGTH, endLine)))
-		{
+		if (lineRange.second != lastLineDoc)
 			return;
-		}
 	}
 	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
 	auto posFound = _pEditView->searchInTarget(str2Search, lstrlen(str2Search), startPos, endPos);
-	if (posFound >= 0){
+	if (posFound >= 0)
 		_pEditView->replaceTarget(TEXT(""), posFound, endPos);
-	}
 }
 
 void Notepad_plus::removeDuplicateLines()
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1461,11 +1461,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	FindOption env;
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)*");
+		env._str2Search = TEXT("^(?>[\\t ]*[\\r\\n]+)+");
 	}
 	else
 	{
-		env._str2Search = TEXT("^$(\\r?\\n|\\r)(\\1)*");
+		env._str2Search = TEXT("^[\\r\\n]+");
 	}
 	env._str4Replace = TEXT("");
 	env._searchType = FindRegex;
@@ -1477,15 +1477,20 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	if (isBlankContained)
+	auto lastLine = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
+	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLine - 1);
+	auto endPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLine) + _pEditView->execute(SCI_LINELENGTH, lastLine);
+	if (!isEntireDoc)
 	{
-		env._str2Search = TEXT("(\\r?\\n|\\r)^[\\t ]*$");
+		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
+		endPos  = _pEditView->execute(SCI_GETSELECTIONEND);
 	}
-	else
-	{
-		env._str2Search = TEXT("(\\r?\\n|\\r)^$");
+	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
+	auto posFound = _pEditView->searchInTarget(str2Search, lstrlen(str2Search), startPos, endPos);
+	if (posFound >= 0){
+		_pEditView->replaceTarget(TEXT(""), posFound, endPos);
 	}
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 }
 
 void Notepad_plus::removeDuplicateLines()
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1487,7 +1487,8 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	nbTotal = _findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	auto lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto lineCount = _pEditView->execute(SCI_GETLINECOUNT);
+	auto lastLineDoc = lineCount - 1;
 	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
 	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLineDoc - 1);
 	auto endPos = _pEditView->execute(SCI_GETLENGTH);
@@ -1496,13 +1497,18 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
 		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
 		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
-		if (lineRange.second != static_cast<size_t>(lastLineDoc)){
+		if (lineRange.second != static_cast<size_t>(lastLineDoc))
+		{
 			if (nbTotal == 0)
 			{
 				_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
 				_pEditView->execute(SCI_SETCURRENTPOS, mainSelCaretPos);
 			}
 			return;
+		}
+		else if (lineCount > 1 && (endPos - startPos > 0))
+		{
+			startPos = _pEditView->execute(SCI_GETLINEENDPOSITION, lineRange.first - 1);
 		}
 	}
 	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1477,14 +1477,19 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	auto lastLine = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
 	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
-	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLine - 1);
-	auto endPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLine) + _pEditView->execute(SCI_LINELENGTH, lastLine);
+	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLineDoc - 1);
+	auto endPos = _pEditView->execute(SCI_GETLENGTH);
 	if (!isEntireDoc)
 	{
 		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
-		endPos  = _pEditView->execute(SCI_GETSELECTIONEND);
+		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
+		auto endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
+		if (endPos != (_pEditView->execute(SCI_POSITIONFROMLINE, endLine) + _pEditView->execute(SCI_LINELENGTH, endLine)))
+		{
+			return;
+		}
 	}
 	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
 	auto posFound = _pEditView->searchInTarget(str2Search, lstrlen(str2Search), startPos, endPos);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1471,8 +1471,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	env._searchType = FindRegex;
 	auto mainSelStart = _pEditView->execute(SCI_GETSELECTIONSTART);
 	auto mainSelEnd = _pEditView->execute(SCI_GETSELECTIONEND);
+	auto mainSelAnchor = _pEditView->execute(SCI_GETANCHOR);
+	auto mainSelCaretPos = _pEditView->execute(SCI_GETCURRENTPOS);
 	auto mainSelLength = mainSelEnd - mainSelStart;
 	bool isEntireDoc = mainSelLength == 0;
+	int nbTotal = 0;
 	if (!isEntireDoc)
 	{
 		env._isInSelection = !isEntireDoc;
@@ -1481,10 +1484,10 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		auto newSelEnd = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.second) + _pEditView->execute(SCI_LINELENGTH, lineRange.second);
 		_pEditView->execute(SCI_SETSEL, newSelStart, newSelEnd);
 	}
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
+	nbTotal = _findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	size_t lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
 	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
 	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLineDoc - 1);
 	auto endPos = _pEditView->execute(SCI_GETLENGTH);
@@ -1493,13 +1496,26 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
 		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
 		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
-		if (lineRange.second != lastLineDoc)
+		if (lineRange.second != static_cast<size_t>(lastLineDoc)){
+			if (nbTotal == 0)
+			{
+				_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
+				_pEditView->execute(SCI_SETCURRENTPOS, mainSelCaretPos);
+			}
 			return;
+		}
 	}
 	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
 	auto posFound = _pEditView->searchInTarget(str2Search, lstrlen(str2Search), startPos, endPos);
 	if (posFound >= 0)
+	{
 		_pEditView->replaceTarget(TEXT(""), posFound, endPos);
+	}
+	else if (!isEntireDoc && nbTotal == 0)
+	{
+		_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
+		_pEditView->execute(SCI_SETCURRENTPOS, mainSelCaretPos);
+	}
 }
 
 void Notepad_plus::removeDuplicateLines()


Yaron10:
Select the first line including its EOL.
תמונה

Remove Empty Lines.

**Result:**
תמונה

**Expected:**
No change.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1477,15 +1477,20 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	if (isBlankContained)
+	auto lastLine = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
+	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLine - 1);
+	auto endPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLine) + _pEditView->execute(SCI_LINELENGTH, lastLine);
```
Yaron10, line 1483: Can't you use docLength?

Yaron10, line 1483: Can't you use docLength?

ArkadiuszMichalski:
@Yaron10 This behavior is in the current version. I didn't change it, just focused on improving performance.

I know that such situations can occur with the selections. But this is because a simplified replace is done here that covers both cases (with or without selection). If you imagine that your selection is the only lines in the document and last line with a CRLF at the end, then this action is correct. Although this is probably not what you expect.

Initially, I tried to do it separately, i.e. get rid of spaces/tabs (for this second command) first, then remove the last CRLF. And additionally cover various rare cases (including selections). But there was always something new to come out and I had to add new conditions. Then the whole thing became unclear (without comments, basically difficult to understand).

So I finally returned to the original (simpler) form as it was until now. The only fix in the new approach is that if you have one line with only tabs / spaces (selected or not) then the second command will delete it https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12512#issuecomment-1317940411. The rest works the same as before.

PowerEditor/src/Notepad_plus.cpp
ArkadiuszMichalski: If you mean `_pEditView->execute(SCI_GETLENGTH)` then yes, would be simpler.

Yaron10: Yes, I meant `auto endPos = _pEditView->execute(SCI_GETLENGTH)`.

Yaron10:
@ArkadiuszMichalski,

> If you imagine that your selection is the only lines in the document and last line with a CRLF at the end, then this action is correct.

If you only select the first line including its EOL, the second line should NOT be considered "selected".

Thank you and @alankilborn for improving the performance.

alankilborn:
@Yaron10 said:

>If you only select the first line including its EOL, the second line should NOT be considered "selected".

Very true.
Notepad++ functions didn't agree until a couple of years ago when @sasumner made changes to correct most places that were affected by that.

Yaron10:
@alankilborn,

> Notepad++ functions didn't agree until a couple of years ago when @sasumner made changes to correct most places that were affected by that.

Do you know of any other function still "affected by that"?
I hardly use those Remove/Sort commands.

alankilborn:
>>>If you only select the first line including its EOL, the second line should NOT be considered "selected".

>>Notepad++ functions didn't agree until a couple of years ago when @sasumner made changes to correct most places that were affected by that.

@Yaron10 said:

> Do you know of any other function still "affected by that"?

Hmm, I was thinking about the following, but it is different (but I think possibly related?).

-----------------------------------

Start with nothing selected (I know, not what we were talking about...):

image

Do an ascending sort (which you never use :-) ) on it to obtain:

image

Yuck!

I think this actually got declined at some point, but maybe in a slightly different form?  Can't find it now...
Maybe it is time for a new issue describing exactly this?

---------------------------------------

You have to actually do a selection (even when you simply want to do the entire file) for it to come out correct:

Manual select from top to bottom:

image

OR Ctrl+a (select-all):

image

Either of the above result in the data I expect:

image

---------------------------------------

You can obtain the correct sort from all of the above starting points (nothing selected, manual selection, Ctrl+a selection), if this is your data:

image

Sort result:

image

But note, that is NOT my data, as line 4 has no line-ending (and I don't consider a final "line" without an ending to be a real line).

Yaron10:
@alankilborn,

תמונה

I suppose that in this case you'd agree line `5` should become line `1` after sorting.
So I'm not sure why in your case it should be different.
The average user would consider both line `5` and line `6` as empty lines. 

> and I don't consider a final "line" without an ending to be a real line

You have **5** lines in the status bar.

As for the discrepancy between none-selected and all-selected: this manifests itself in the status bar as well.
(And, all in all, I think it's OK).

> I think this actually got declined at some point, but maybe in a slightly different form? Can't find it now...

It would be interesting to see the declined PR. :)

Yaron10:
@alankilborn,

> I can't tell if you are in total agreement with me, absolute disagreement, or something in the middle.

Let's define it as "disagreement".

> And note that I'm not complaining about the sort itself, lest anyone think that. I'm complaining about what is used as input to the sort.

Yes, I _totaly_ understand that.

> Only if the average user has below-average intelligence.

I'd like to believe the average NPP user has above-average intelligence. :)

> A line is defined as some run of non LF/CR characters, followed by LF or CR or CRLF.

This seems to be the root of the disagreement.
I might be lacking some professional knowledge here. And yet - there are different types of lines, but our last line is still a "line". 
And although I do see your point, I still tend to regard your expected result as "unexpected".

Yaron10:
If I understand you correctly, you're arguing that our last line is a mere "placeholder" and not a proper line.
Again, I see that. But I think it's too sophisticated. :)

alankilborn:
> I still tend to regard your expected result as "unexpected".

As apparently do the N++ developers.
So I have to go through my life, remembering that I just can't sort my well-formed files, or I will corrupt them.
I have to remember to Ctrl+a their contents first, then sort them.
It seems ridiculous to me.
I may have "muscle" memory for this workaround by now, or maybe not.

Lightbulb: Script! :-)

alankilborn:
@ArkadiuszMichalski Sorry for issue/thread hijack.  I'll get back to commenting on the original topic soon!

alankilborn:
>A setting for that 

And interpreting the meaning of the setting I'm sure would confuse even the above-average intelligence users. :-)

ArkadiuszMichalski:
Please do not discuss about other problems here because it's difficult to analyze the actual problem (open new issue and move this discussion here).

Coming back to the problem, I will add some conditions to selections to avoid some unexpected situations.

Yaron10:
@ArkadiuszMichalski,

> Please do not discuss about other problems here because it's difficult to analyze the actual problem (open new issue and move this discussion here).

We have both apologized for that. Please don't exaggerate.

alankilborn:
Back to appropriate discussion...

What should the behavior be for _Remove Empty Lines (Containing Blank characters)_ with this data?:

image

ArkadiuszMichalski:
Please test last commit with some selections. This should fix @Yaron10 problem, also for above it only remove 3 line (which is correct), doesn't touch 2 and 4 line (because these lines are partially selected).

alankilborn:
>also for above it only remove 3 line (which is correct), doesn't touch 2 and 4 line (because these lines are partially selected).

OK, that's what I was getting at -- my understanding is that lines have to be "fully selected" to be removed (and meet the criterion of only space-and-tab content as well, of course).

ArkadiuszMichalski:
>OK, that's what I was getting at -- my understanding is that lines have to be "fully selected" to be removed (and meet the criterion of only space-and-tab content as well, of course).

Yes, and "fully selected" means also with the `EOL` (we must remember to select this last EOL).

In the previous approach, when selection is incomplete, whitespace was removed unnecessarily, which looked incorrect to me. Well, by adding a simple condition, that was fixed too.

alankilborn:
>"fully selected" means also with the EOL (we must remember to select this last EOL)

I could not agree with you more on that. :-)

Yaron10:
@ArkadiuszMichalski,

The *exact* case I mentioned above is fixed.
But if line 2 is empty, it's still wrongly deleted.

תמונה

***
> What should the behavior be for Remove Empty Lines (Containing Blank characters) with this data?
>> also for above it only remove 3 line

👍 

***
> Yes, and "fully selected" means also with the EOL (we must remember to select this last EOL).
>> I could not agree with you more on that.

I don't understand why.
If all characters are selected *excluding* EOL, the line is "fully selected".
And this is how it works in other commands (Sort, Indent etc.).

Yaron10:
And a partially selected line IS sorted. :)

alankilborn:
Here's the big question:

Should commands (including _Remove empty..._ and _Sort..._) that can act upon a selection (if a single stream selection is active, (`a`) act on text only between the start and end positions of the selection OR (`b`) should they act on text between the start-of-line-pos of the line containing the selection-start-pos and the end-of-line-pos (whatever end-of-line-pos really means -- line-ending char(s) in, or out?) of the line containing the selection-end-pos -- whew!

I like to refer to (`b`) as acting upon "whole lines touched" by the active stream selection.

Does it make sense to take a stance that all commands should act the same way (whichever one is decided upon), or are there certain commands that need to act differently?  Example:  _Sort..._ really can't act on partial line data...

Bottom line:  Complicated!

Yaron10:
> Acting the same in all situations has the obvious virtue of consistency (less confusion about it for the user).

👍 

Indent, if I remember correctly, is purely Scintilla.
And partially selected lines ARE indented.
So it seems that all commands should include partially selected lines.

"Execute the command on lines containing selections".

alankilborn:
>If all characters are selected excluding EOL, the line is "fully selected".

So if all characters on a "line" are selected _including_ the EOL character(s), what do we call that?  _super_-selected?  Or probably an answer to that will be we don't have to call it anything.  :-)

ArkadiuszMichalski:
@Yaron10 
> 
> The _exact_ case I mentioned above is fixed. But if line 2 is empty, it's still wrongly deleted.
> 
> תמונה

What excatlly is wrong here? What was before and after run cmd (and which cmd), you select something or not?. And please turn on `Show All Characters`.

Yaron10:
@ArkadiuszMichalski,

Line 1 is selected including the EOL.
Line 2 does not contain any characters.

Remove Empty Lines.

Expected:
Line 2 should not be deleted since it is NOT selected. 

If the purpose of this PR is improving performance, the other issues can be fixed separately. As you understand.
Thank you for your work.

ArkadiuszMichalski:
But this behawior is correct, when you select only Line1 + EOL it's the same as you select all (Ctrl+A) or not select anything (which means the entire document). So Line2 is empty and will be removed.

Yaron10:
So the off-topic comments are related. :)
Please read them.

I'll summarize it here.
תמונה

Line 1 is selected including the EOL.
Line 2 is empty.

Sort does NOT change the order as the second line is not selected. 
If nothing is selected, Sort changes the order.

ArkadiuszMichalski:
But what do you expect, that from belowe text and sorting:
image
you should get:
image
or what?

Take into account that in text like `aCRLF` (what you see in Line1) this `CRLF` basically means Line2 (but empty). You select this `CRLF` (this empty line) so it will be removed.

Let's not refer to sorting here because it's a completely different function (done differently, maybe it requires its own quirks).

Yaron10:
> Let's not refer to sorting here because it's a completely different function (done differently, maybe it requires its own quirks).

OK.

But I think we should assume the user knows what they have selected and act accordingly.
And the behavior should be consistent for all commands and in-sync with the status bar (selected lines Nb.).

If you accept the current status bar implementation on selecting a line including the EOL (selected lines: 1), then we can NOT treat the second line as selected.
And indeed there is some discrepancy between none-selected and all-selected.
Select All means select all that can be selected. And an empty line can not be selected.

Yaron10:
Taking a break until tomorrow.

Yaron10:
> Take into account that in text like aCRLF (what you see in Line1) this CRLF basically means Line2 (but empty). You select this CRLF (this empty line) so it will be removed.

I would define EOL as follows: CRLF are the last characters _in this line_ . Any char inserted *after* them should be placed in a new line.
So selecting EOL means selecting the end of the current line. You do not select the next line.

alankilborn:
>So selecting EOL means selecting the end of the current line. You do not select the next line.

This is all old news.  Reference earlier where I said that @sasumner fixed this behavior (for most commands) a few years ago.  Let's not go backwards.  Just because selecting a line and its EOL puts the caret at the start of the _following_ line, that following line should NOT be included in the upcoming operation (as Yaron10 said).  The unfortunate part about this is that it takes extra code to avoid getting the following line included.

alankilborn:
I think I found the code I was talking about:  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6803b69c0b92057d2ac5700b7a312a6cf2f82b39/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L2918-L2923

The `getSelectionLinesRange` function should be called when it is desired to get a starting line and ending line for either partially selected or wholly stream-selected lines.  Note that when we are talking about such lines, it can only be the first or last line of the selection that has some partial selection; obviously lines in the selection that aren't the first or last line are themselves wholly selected.

ArkadiuszMichalski:
This still doesn't mean that this selection is somehow magically skipping the last blank line. For:
image
and:
image
we have:
image

Where is my last empty line, I made selection without last line...

The current approach of these commands is:
1. Selecting something until the end (with last EOL)
2. Selecting by Ctrl+A (this also including last EOL)
3. Not selecting (which means the entire document)

Each of the above cases removes this last blank line (which is consistent). We can see that the person who wrote this function clearly wanted such behavior, because he added it at the end. Although it initially detects selection, it doesn't skip removing the last line for it (whether intentionally or not, I don't know).
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6803b69c0b92057d2ac5700b7a312a6cf2f82b39/PowerEditor/src/Notepad_plus.cpp#L1472

If the 1st and/or 2nd case are changed, then someone may object that now he always have to manually go to the end of the document and delete the last blank line. So this command doesn't work well because it was supposed to remove all blank lines, but it doesn't (everything is selected till the end of the document, I can't select any more). On the other hand, you can always select less :)

So far I haven't seen any reports related to these two commands and the last EOL. If there is no explicit request to change this behavior, I see no reason to do it.

Yaron10:
> This still doesn't mean that this selection is somehow magically skipping the last blank line.

This is a bug in the `In Selection` implementation which I've discussed with @sasumner.
Unfortunately he quit contributing to Notepad++ before fixing it. 

> So far I haven't seen any reports related to these two commands and the last EOL. If there is no explicit request to change this behavior, I see no reason to do it.

It's been requested here.
It's your PR. As you understand.

alankilborn:
>This is a bug in the In Selection implementation which I've discussed with @sasumner.

Off topic? (sorry)
@Yaron10 Do you have pointer to such discussion? I'm curious about it

alankilborn:
Back on topic:

>If there is no explicit request to change this behavior, I see no reason to do it.

@ArkadiuszMichalski You're in there making changes; it might as well be the BEST it can be, even if no complaints/requests have been made about it.  IMO.

Yaron10:
@alankilborn,

> Do you have pointer to such discussion? I'm curious about it

Unfortunately I don't. But the discussion was held here on GitHub.
If I remember correctly, the issue I raised was that `In Selection` includes the next line in the search (i.e. not our issue here in particular).
Please update if you find it.

Taking a break until Monday.

Yaron10:
@alankilborn,

Testing `In Selection` now, I'm not sure what the discussion was about. But I'm positive there was some issue.

Yaron10:
@alankilborn,

>

>This still doesn't mean that this selection is somehow magically skipping the last blank line.
>
>> This is a bug in the In Selection implementation which I've discussed with @sasumner.

I'm afraid I misunderstood @ArkadiuszMichalski's comment.
If I understand it correctly now, it's the expected behavior and not contrary to our approach here whatsoever.
The new line indicator _in the selected line_ is removed and therefore the next line is deleted.
Here we are not looking for new line indicators but for empty lines.

I must leave now. Sorry. :)  
Unfortunately he quit contributing to Notepad++ before fixing it.

ArkadiuszMichalski:
I looked at all operations from `Edit>Line Operations` how they work with and without selection (`Sort`, `Reverse`, `Randomize` and `Remove empty`). And all except (`Remove empty`):
1. Without selection they include the last empty line.
2. With selection they omit the last empty line.
3. For partially selected lines they automatically extend the selection to the entire lines. For sorting it's obvious. But for other cases it also makes sense because we operate on the lines and we do not have to worry about the full (correct) line selecting.

\* `Join lines` command fit only to points 2 and 3 (not work without selection, maybe should).

Taking the above into account, I also made `Remove empty` consistent with this behavior.

@alankilborn This case for sorting https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1320165449, although it looks strange, it's not such unusual. I checked out a few other editors, and where there was sorting command without selecting, the behavior was identical like in Notepad++ (it was Sublime Text). If you have other editors that suport sorting without selection, it's also worth checking how they behave.

alankilborn:
> Join lines command ... not work without selection, maybe should

See https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12399 for exactly that!

alankilborn:
>I looked at all operations from Edit>Line Operations how they work with and without selection (Sort, Reverse, Randomize and Remove empty).

This is unclear because reverse and randomize are actually a form of sort, so it appears you examined "sort" and "remove empty", not all line operations.

>Without selection they include the last empty line.
With selection they omit the last empty line.

I suppose more accurately you mean: Without selection of all text, either manually or with Ctrl+a

To me I don't understand the difference between acting upon the entire document text (when there is no selection) versus a selection that covers the text of the entire document.  Sure, they are different things, but the data indicated to use as input should be the same, so why not the result the same?

>This case for sorting, although it looks strange, it's not such unusual.

I still maintain that my data is being corrupted in the example I provided.
I think I will open a new issue about this, although it may go nowhere.
Note1 to self: If there is text but no EOL on last line of a file, this line should get an EOL appended before sorting.
Note2 to self: Line endings should be unified before sort, thus solving #7735 as well.

Yaron10:
@ArkadiuszMichalski,

תמונה

Remove Empty Lines (including blank).

Result:
Line 2 is not removed.

Could you please list the STR of the case https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535/commits/4b5369e8483521c7363c614ed4aa6611505c785c should fix?

Thank you for your work.

ArkadiuszMichalski:
@Yaron10 With build you test? From last commit this works for me:
Before:
image
After:
image


>Could you please list the STR of the case https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4b5369e8483521c7363c614ed4aa6611505c785c should fix?

When you have partially selected line this command (from last commit) automatically expand selection to full selected line. But when nothing was change we bring back original selection.
Before:
image
After (the same):
image
Not:
image

It also restores the correct position of the caret (if you selecting from right to left).

Yaron10:
@ArkadiuszMichalski,

> When you have partially selected line this command (from last commit)...

Very nice! 👍 

If you push another commit here, could you also replace the regex with the better ones from https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12544?

Thanks again for your work.

ArkadiuszMichalski:
@Yaron10 This PR use better regex.

Btw, what should be done for this?
image

It's obvious that only 1 line cannot be completely deleted, but in this situation it should be cleared or not?

Yaron10:
@ArkadiuszMichalski,

> This PR use better regex.

Great.
I didn't check the code and thought you found even better regex.

> Btw, what should be done for this?

The letter versus the spirit of the command. :)
I think it should be cleared.

ArkadiuszMichalski:
@Yaron10 Next fix for test is ready.

Yaron10:
@ArkadiuszMichalski,

Great work! 👍 

Testing it again, I've encountered another interesting case. 
If you still have patience and it doesn't require a lot of extra work.

תמונה

Remove Empty Lines (including blank).

Result:
תמונה

Expected:
תמונה

Thanks again.

ArkadiuszMichalski:
@Yaron10 I don't understand what the problem is. Line 2 has been deleted.

Yaron10:
@ArkadiuszMichalski,

It's a _minor_ issue regarding the selection after deleting the line.
Why should `Test` be selected instead of `st`?

ArkadiuszMichalski:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1322903389
>When you have partially selected line this command (from last commit) automatically expand selection to full selected line. But when nothing was change we bring back original selection.

Here we have changed (taking into account the entire selection) so in this situation it does not correct anything. I don't know if it can be traced in any simple way. Because we may have a situation where in selection the starting line is deleted, the last one, or anything in between, or even everything. I don't know how recognize that the original start /end line in selection has not actually changed.

Yaron10:
@ArkadiuszMichalski,

I see.
I'm exhausted now. Let's continue tomorrow.

Anyway, it's a minor issue/observation.

Yaron10:
@ArkadiuszMichalski,

Thinking about it with a clearer mind, your implementation may be better in concept as well.

For whatever reason the user had selected only part of the lines, their selection - _as it was_ - is no longer valid.
So the purpose of the post-deletion-selection is highlighting the remaining relevant lines. And it makes sense to select the entire lines.

We get the same result on Sort.

On block-selection (Alt) - there's no selection after Sort, whereas the entire lines are selected in your PR here.
Perhaps the Sort behavior should be changed.

Yaron10:
> On block-selection (Alt) - there's no selection after Sort, whereas the entire lines are selected in your PR here.
Perhaps the Sort behavior should be changed.

Another difference between Sort and `Remove Empty Lines`: on multiple-selections (Ctrl+) - Sort is not executed at all, whereas `Remove Empty Lines` is executed on the main selection only.

ArkadiuszMichalski:
@Yaron10 
When referring to selections, it's worth do consistent behavior for remove empty/dup lines because they delete something. Sorting doesn't delete anything so selecting here is less problem (can be determined separately).

So what would be appropriate?
1. Do not correct anything and let the selection be a result from the performed operations.
2. As in last PR, when nothing was change then restore original selection (before automatic expansion the entire line).
3. Try to keep the original position selection on the first/last line if something has changed between them (but it doesn't seem that simple).
4. Remove selection after all operations (if there was a change, or if there was no change, or always?).

Or something else?

alankilborn:
I believe that after command operations that are initiated with some text selected, the resulting text should be selected.  So, for removing-empty-lines, the remaining lines from the original selection should be selected after the removal.  Or, if no lines were removed, the selection could either be left as original, or if first/last line were originally partially selected, the whole-line version could be left selected.  If all selected lines were empty and thus removed, obviously no selection should be active, and caret would be placed at the removal point.

Gets complicated fast (again).

For sake of current issue/PR, maybe this could be a new issue: "What remains selected after a command that operates on the selection executes".

If upscoping of the current issue continues, it might never be completed. :-)

alankilborn:
>...difference between Sort and Remove Empty Lines: on multiple-selections (Ctrl+) - Sort is not executed at all, whereas Remove Empty Lines is executed on the main selection only.

AFAIK, no Notepad++ commands execute on multiple selections as they should; they all act on the main selection alone.

This should also be addressed, but again, in a separate issue, e.g. "Commands that execute on the main selection should act on multi-selections as well, where applicable".

Because sort uses multiple selections in the form of column selection for the sort key, this is probably why it doesn't also act on main stream selection only.  I'm not saying this is correct, just making an observation.

ASIDE: If the fact that in 2022 Notepad++ can't do a replacement operation in a selected column-block gets mentioned in this thread (besides right here), I may just have to permanently retire. :-)

Yaron10:
> When referring to selections, it's worth do consistent behavior for remove empty/dup lines because they delete something. Sorting doesn't delete anything so selecting here is less problem (can be determined separately).

👍 

> So what would be appropriate?

I think that your current implementation is very good. 👍 
If I understand @alankilborn's comment correctly, he seems to agree.

> AFAIK, no Notepad++ commands execute on multiple selections as they should; they all act on the main selection alone.
>
> This should also be addressed, but again, in a separate issue, e.g. "Commands that execute on the main selection should act on multi-selections as well, where applicable".

Referring only to `Remove Empty/Duplicate`, it seems rather complex. If it's worth it, it should indeed be addressed in a separate issue.
For the time being and for concluding this PR, what do you gentlemen suggest?
Act on the main selection only (current implementation), or do nothing?

> ASIDE: If the fact that in 2022 Notepad++ can't do a replacement operation in a selected column-block...

Ping @sasumner. :)

alankilborn:
WAY OFF-TOPIC:

>>ASIDE: If the fact that in 2022 Notepad++ can't do a replacement operation in a selected column-block...

>Ping @sasumner. :)

More like "ping Don".  I believe I remember that @sasumner wanted to tackle that one, and Don quashed it because it would be "too risky".  :-)

ArkadiuszMichalski:
>@Yaron10 I think that your current implementation is very good. 👍

Then suppose the selection behavior from this PR is expected and can be done similarly in the rest of the commands.

Regarding block selection or multiple selections: Find/Replace dialog doesn't offer this, so commands using this window (via `_findReplaceDlg`) will not handle it either. The only simple thing we can do this is:
1. Do nothing for both case.
2. For block selection expand it to full firt/last line as one main selection or do nothing.
3. For multiple selection use only first/last as main selection or do nothing.

Or don't use `_findReplaceDlg` but then we have to do everything ourselves, which will complicate these commands much more. It's not part of this PR anyway and it will certainly not change quickly.

The only thing that can be done at the moment is to somehow unify the behavior of these commands based on what we have now.

Yaron10:
> Then suppose the selection behavior from this PR is expected and can be done similarly in the rest of the commands.

👍 

> For block selection expand it to full firt/last line as one main selection

This is the current behavior and I think it's good.

> For multiple selection use only first/last as main selection or do nothing.

The current behavior (act on the main selection only) is slightly confusing IMO.
"do nothing" is not ideal either, but less confusing. 
As you understand.
@alankilborn, what's your take on that?

> Or don't use _findReplaceDlg...

:)

> The only thing that can be done at the moment is to somehow unify the behavior of these commands based on what we have now.

👍 

Thank you.

alankilborn:
>>For block selection expand it to full first/last line as one main selection

>This is the current behavior and I think it's good.

Currently after execution it leaves text selected strangely (to a user).  If that is being improved in this or another issue/PR then I think it is good.  If that part of it is not being improved, then, it seems confusing and I'd rather it do nothing if executed with a column block selected.

----------------------------------------------------------------

>>For multiple selection use only first/last as main selection or do nothing.

>The current behavior (act on the main selection only) is slightly confusing IMO.
"do nothing" is not ideal either, but less confusing.
@alankilborn, what's your take on that?

Agree that it is confusing as user may not remember which is the main selection, even though it is the most recently-selected.  And then user is very disappointed that he went to the trouble of making multiple selections to act upon--and it didn't.

"do nothing" may be just as confusing as acting on only main selection.

I think the best solution would be to loop over the selections in this case, performing the logic on (the whole-line version) of each.

Thank you for requesting and considering my thoughts on the matter. :-)

--------------------------------------------------------

>ASIDE: If the fact that in 2022 Notepad++ can't do a replacement operation in a selected column-block gets mentioned in this thread (besides right here) -- because everything but the kitchen sink seems to get brought up here -- I may just have to permanently retire. :-)

When I said the above, it was totally unrelated to the current topic.
But, we talked so long that it became relevant to it. :-)
Here's the connection:

>Or don't use _findReplaceDlg...

It is probably unreasonable to not use _findReplaceDlg for the current command.

But it is okay, since a column block should be expanded to its whole line version as a single stream selection before running the command anyway.

ArkadiuszMichalski:
@alankilborn 
>>>For block selection expand it to full first/last line as one main selection

>>This is the current behavior and I think it's good.

>Currently after execution it leaves text selected strangely (to a user). If that is being improved in this or another issue/PR then I think it is good. If that part of it is not being improved, then, it seems confusing and I'd rather it do nothing if executed with a column block selected.

I wouldn't say strangely, after executing the command the selection left on is what that command operated on. One improvement may be that in the case of block selection, if nothing has been changed, let's restore block selection as well.  
From:
image
And after execute command bring this selection back (as above), don't change to:
image
If something was removed then let's be the default behavior of selection.

>Agree that it is confusing as user may not remember which is the main selection, even though it is the most recently-selected. And then user is very disappointed that he went to the trouble of making multiple selections to act upon--and it didn't.
"do nothing" may be just as confusing as acting on only main selection.
I think the best solution would be to loop over the selections in this case, performing the logic on (the whole-line version) of each.

Yes, the loop would probably be the best, but it's more work and rather a matter of the future. And at this point, what's better, not doing anything (like sorting), or performing an operation only on the last selection like it's done now (I mean the last selection made, not selection positioned as the last one in the content)?

alankilborn:
>And at this point, what's better, not doing anything (like sorting), or performing an operation only on the last selection like it's done now

Keeping current behavior is better than doing nothing.

----------------------------------------------------------------------

>>Currently after execution it (remove-empty with column-block selection) leaves text selected strangely (to a user).

>I wouldn't say strangely

I _would_ say strangely, to a non-advanced user...his selection has changed from a column one to a stream one (he wonders WHY?, but his empty lines have been removed, so he shrugs and moves on), just as shown in your 2 screenshots in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1326407330

Yaron10:
**Column-block selection (Alt).**

> Currently after execution it leaves text selected strangely (to a user).

👍 

I understood you were referring to a case where nothing-was-changed only after seeing @ArkadiuszMichalski's screenshots.
I missed that case.

> One improvement may be that in the case of block selection, if nothing has been changed, let's restore block selection as well.

👍 
That would certainly be better.

Or...
_In real life - is any user going to hold Alt and select column-block for removing empty lines?_
I think it's only us testing the various possible cases. :)

I suppose the implementation/code should be shorter and simpler if we "do nothing" (return without changing anything).
What do you both think?

***

**Multiple-selections.**

> Thank you for requesting and considering my thoughts on the matter.

Well, thank *you* for replying. I do appreciate your opinion. :)

> Yes, the loop would probably be the best, but it's more work and rather a matter of the future...
>> Keeping current behavior is better than doing nothing.

👍 

Thank you both.

Yaron10:
Reading your last comments again, I think there has been some confusion regarding the terms "currently / current behavior".
@ArkadiuszMichalski and myself have been referring to the behavior of the last commit in this PR, whereas @alankilborn meant the current (latest) Notepad++ behavior.

donho:
Conflict - need to be solved. I recommend to redo a new PR based from the current master instead of  solving it.

PR #12544, opened 2022-11-21
More performance optimization for the Remove Empty Lines command

ArkadiuszMichalski:
I found regular expressions even more efficient than the [last ones ](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12512)(and simpler).
```
^[\r\n]+

^(?>[\t ]*[\r\n]+)+
```
This can be checked with https://regex101.com/ (PCRE) and file for test  [regex.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10025169/regex.txt).

I also play with this `// remove the last line if it's an empty line.` with some big file. There is no point to checking the entire file when we want to remove the last blank line. We need only check the last two (or one) lines, which can be done directly with Scintilla commands. From now on, this last empty line removal will not slow as the multi-line document grows.

This PR also cover two more cases:
- also works with one empty line [#12512 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12512#issuecomment-1317940411)
- add some condition for selection to avoid some unexpected situations ([like this](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1319497990))

This is just a performance improvement. In particular, it doesn't change the behavior of how to treat the last EOL with or without selection. It works as it has been until now. I suggest discussing changing this behavior elsewhere (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12545, https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535).

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1461,11 +1461,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	FindOption env;
 	if (isBlankContained)
 	{
-		env._str2Search = TEXT("^([\\t ]*$(\\r?\\n|\\r))(\\1)*");
+		env._str2Search = TEXT("^(?>[\\t ]*[\\r\\n]+)+");
 	}
 	else
 	{
-		env._str2Search = TEXT("^$(\\r?\\n|\\r)(\\1)*");
+		env._str2Search = TEXT("^[\\r\\n]+");
 	}
 	env._str4Replace = TEXT("");
 	env._searchType = FindRegex;
@@ -1477,15 +1477,22 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	if (isBlankContained)
+	auto lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
+	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLineDoc - 1);
+	auto endPos = _pEditView->execute(SCI_GETLENGTH);
+	if (!isEntireDoc)
 	{
-		env._str2Search = TEXT("(\\r?\\n|\\r)^[\\t ]*$");
+		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
+		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
+		auto endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
+		if (endPos != (_pEditView->execute(SCI_POSITIONFROMLINE, endLine) + _pEditView->execute(SCI_LINELENGTH, endLine)))
+			return;
 	}
-	else
-	{
-		env._str2Search = TEXT("(\\r?\\n|\\r)^$");
-	}
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
+	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
+	auto posFound = _pEditView->searchInTarget(str2Search, lstrlen(str2Search), startPos, endPos);
+	if (posFound >= 0)
+		_pEditView->replaceTarget(TEXT(""), posFound, endPos);
 }
 
 void Notepad_plus::removeDuplicateLines()


donho:
@ArkadiuszMichalski 
There is another PR #12548 which addresses the same issues - on which one should I focus?

ArkadiuszMichalski:
@donho 
This one basically improves the performance. So it could be approved without any additional problems. 

On https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12548 I'm trying to change the behavior of this command (ie how to treat the last empty line when doing a selection, how to set selections after doing operation, etc.). So it could be accepted separately (if this new approach is better than the current one), so then in case of any problems in the future it would be easier rejected.

But the performance improvement from this PR will be preserved.

donho:
Could you provide any test (file + instructions) to prove this PR has optimized the performance?

donho:
And please close #12548

ArkadiuszMichalski:
>@donho And please close https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12548

But https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12548 is related to other command `Remove Consecutive Duplicate Lines`. I plan solve it after close this PR and also after next PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535 (because I need to know earlier what behawior is more excepted, what will be accepted or not, this command is very similar). 

In the case of tests:

For the first improve you can play with https://regex101.com/ (PCRE or PCRE2) and file for test [regex.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10025169/regex.txt). You can see that this new regex are shorter and faster (takes less time and do less steps).
Just compare two scenerio:

For empty line:
```
^[\r\n]+

vs 

^$(\r?\n|\r)(\1)*

```

image
image


For empty line with white space:
```
^(?>[\t ]*[\r\n]+)+

vs

^([\t ]*$(\r?\n|\r))(\1)*
```
image
image
Here we also grab more data in one shot that why we see 33 vs 37 matches (it's not a bug).



For second improving `// remove the last line if it's an empty line` you need more detail? As from first post:
>There is no point to checking the entire file when we want to remove the last blank line. We need only check the last two (or one) lines, which can be done directly with Scintilla commands. From now on, this last empty line removal will not slow as the multi-line document grows.

This can be observed on larger files. We don't once again scan all data (all lines) to remove potentially last. We check directly only last, so basically it's instant.

donho:
> But #12548 is related to other command `Remove Consecutive Duplicate Lines`. I plan solve it after close this PR and also after next PR #12535 (because I need to know earlier what behawior is more excepted, what will be accepted or not, this command is very similar).

Sorry, I meant "close #12535"



> For the first improve you can play with https://regex101.com/ (PCRE or PCRE2) and file for test [regex.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10025169/regex.txt). You can see that this new regex are shorter and faster (takes less time and do less steps).

We cannot rely on https://regex101.com/  because we don't know about their PCRE implementation. Even they use the same version of boost PCRE in their engine, it's not the same binary code as in Notepad++.

So as far as I don't have the tangible proof that this PR improves the performance, the PR cannot be accepted.
All we need is a file which can show us the difference of time (BEFORE and AFTER applying the PR) to execute the command in question.

ArkadiuszMichalski:
@donho 
>We cannot rely on https://regex101.com/ because we don't know about their PCRE implementation. Even they use the same version of boost PCRE in their engine, it's not the same binary code as in Notepad++.

I know this is not the same implementation as in Notepad++. But it makes the presentation easier why these new Regex are more efficient. We don't use capturing group (which is always more efficient), we grab more data in one matching.

Hypothetical simple test case:
```
space_space_space CRLF
space_space CRLF
CR
LF
```
This new regex `^(?>[\t ]*[\r\n]+)+` make only one match (grab all this lines) and make one replace. This previous version `^([\t ]*$(\r?\n|\r))(\1)*` makes 4 matches and make 4 replace. So this newer version will definitely be faster (in any implementation), if so why not use it?

>So as far as I don't have the tangible proof that this PR improves the performance, the PR cannot be accepted.
All we need is a file which can show us the difference of time (BEFORE and AFTER applying the PR) to execute the command in question.

Well, but it's necessary to make a file with 1 million lines to prove it (and add time logging)? Please consider this practice case (use only Notepad++):
```
                 
             
		

```
image

vs

image

I can also make clearer example for improving `// remove the last line if it's an empty line`, but if the above does not convince you ...

Yaron10:
@donho,

> Sorry, I meant "close https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535"

May I ask why?
@ArkadiuszMichalski has been doing a great job there.
Obviously, we would appreciate your opinion too. :)

donho:
@Yaron10 
> @ArkadiuszMichalski has been doing a great job there.

I didn't say he does not do a good job.
There are 2 performance optimization PR for the Remove Empty Lines command: #12544 (this one) and #12535
I just don't see why we need 2 PR for the same optimization.

Yaron10:
@donho,

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535 started as an optimization PR but turned into fixing some issues in "Line Operations" commands.
The discussion and work there are about to conclude.
I suppose @ArkadiuszMichalski will ping you when it's done.

Thank you.

donho:
Thank you for your explanation @Yaron10 .
@ArkadiuszMichalski In this case, could you update the title **Change for Remove Empty Lines command how operate on selection and last empty line** to somethings else in order to make the description of PR #12535 more accurate?

ArkadiuszMichalski:
@donho 
But this title just hints at what that PR is about. In short:

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12544 (this PR) - is only for performance optimization for these two commands (better than the recently approved PR).

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535 - that PR change behavior for selection and last empty line for this two commands. I don't know if this change will be accepted (is expected), so I did it separately. It has optimizations from #12544, because I assume it will be accepted after all. But if not then I will revert to the previous approach, keeping the changes for selection and last empty line made following the discussion. What's wrong with the title `Change for Remove Empty Lines command how operate on selection and last empty line`?

It all depends on what will be done with this PR. I added a file to test https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12544#issuecomment-1326667130, so if you need anything else (additional files) let me know.

PR #12566, opened 2022-11-24
fix possible incorrect plugin architecture reports

### LINKED ISSUE #12565: Better info needed when a plugin cannot be loaded
### Description of the Issue

A user experienced the following error when a 64-bit plugin could not be loaded, on a 64-bit non-ARM system:
image

The problem turned out to NOT be that the plugin was 32-bit, so the error message was misleading, and led troubleshooting the problem down the wrong path.

The user's problem ended up being incorrect permissions:  _The issue was caused by my malformed robocopy command which stripped execute permissions from the DLL, so even though it was receiving a permissions failure, it was taking this as a malformed DLL and spitting out the erroneous error._

I could replicate the poor error message when I created a zero-length (intentionally “corrupt” dll) file and placed it in a proper location so that Notepad++ would attempt to load it.

More info in Community issue [HERE](https://community.notepad-plus-plus.org/topic/23789).

1. Set no-execute permissions from plugin DLL, or make a zero-length DLL file
2. Start Notepad++
3. Observe the above misleading error message about "32-bit" plugin.

### Expected Behavior
An error message that helps troubleshoot what is wrong.

### Actual Behavior
Misleading error message.

### Debug Information
Notepad++ v8.4.7   (64-bit)
Build time : Nov  1 2022 - 23:31:11
Path : C:\\...\npp.8.4.7.portable.x64\notepad++.exe
Command Line : -titleAdd="8.4.7 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2130
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

xomx:
Fixes #12565 .

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -116,11 +116,19 @@ int PluginsManager::loadPluginFromPath(const TCHAR *pluginFilePath)
 		int archType = nppParams.archType();
 		if (getBinaryArchitectureType(pluginFilePath) != archType)
 		{
-			const TCHAR *archErrMsg = TEXT("Cannot load 64-bit plugin."); // IMAGE_FILE_MACHINE_I386 by default
-			if (archType == IMAGE_FILE_MACHINE_ARM64)
-				archErrMsg = TEXT("Cannot load 32-bit or non-ARM64 plugin.");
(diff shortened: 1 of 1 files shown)


alankilborn:
Fixes a bug, but doesn't fulfill the feature request, so the issue should remain open!

PR #12575, opened 2022-11-27
fix v847 inconsistencies at OS-forced N++ exit

### LINKED ISSUE #12541: N++ v8.4.7 inconsistencies at exit during OS restart/shutdown
Two problems in the same area reported:
1. Single dirty file is directly saved during the OS restart without asking the user 1st (Community [report1](https://community.notepad-plus-plus.org/post/81726))
2. N++, with its backup engine ON, does not behave identically during the usual N++ exit initiated by the user and during the one initiated by the OS restart (Community [report2](https://community.notepad-plus-plus.org/post/81616)} 

STRs:

1)
- open only 1 existing file in N++
- add some text in the opened file without saving it, do not close N++
- restart Windows, “This program preventing restart...” dialog appears, click on the "Cancel" option there - the opened unsaved file is saved now (error - the N++ SaveAll dialog should be opened instead as in the case of 2 or more dirty files)
- this problem has to be fixed for sure, N++ should behave the same as for the 2 or more dirty files...

2)
- “Remember current session for next launch” and “Enable session snapshot and periodic backup” are enabled in the N++ 'Preferences'
- open some files, edit at least one of them
- now if the running N++ was closed by user manually ... no save dialog for the changed file(s), N++ just silently closes and on its next start all the files are opened again from the N++ backup
- but if the Windows OS restart tried to close that running N++ ...  we get the “This program preventing restart...” dialog and when the user clicks on its "Cancel" option,  N++ offers its  SaveAll dialog for all the unsaved files ("Restart anyway" option is ok from this point of view)
- this is intentional new safer behavior at the OS restart time, but it can be changed back to the previous state, when only the N++ backup engine silently handles the unsaved user's data

Possible solutions:

1)
- either we create a new separate N++ message (e.g. IDM_FILE_SAVEALL_AT_SHUTDOWN) and use it [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6803b69c0b92057d2ac5700b7a312a6cf2f82b39/PowerEditor/src/NppBigSwitch.cpp#L2131), then show in its handler new YESNO dialog “Windows restart/shutdown is closing the Notepad++ running. Do you want to save unsaved file(s)?” etc.
- or we adjust the v8.4.3+ [speed exception](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10995) for the only one dirty file in the [Notepad_plus::fileSaveAll()](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6803b69c0b92057d2ac5700b7a312a6cf2f82b39/PowerEditor/src/NppIO.cpp#L1669) at the OS restart time to:  `if ((nbDirty == 1) && curBuf->isDirty() && !(NppParameters::getInstance()).isEndSessionStarted())`

2)
If the consistency-view wins, we can use an additional check for the N++ active backup [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6803b69c0b92057d2ac5700b7a312a6cf2f82b39/PowerEditor/src/NppBigSwitch.cpp#L2114) like:

```
// we have unsaved filebuffer(s), give the user a chance to respond (non-critical shutdown and N++ backup-OFF only)
if (!isForcedShuttingDown && isFirstQueryEndSession && !nppParam.getNppGUI().isSnapshotMode())
```

xomx:
Fixes #12541 .


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2106,6 +2106,18 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return TRUE;
 		}
 
+		case NPPM_INTERNAL_WINSESSIONEXIT:
```
donho, line 2109: Add 4 more characters DOWS won't make it much longer but make it much more clear :)

donho (changes requested):
Change NPPM_INTERNAL_WINSESSIONEXIT to NPPM_INTERNAL_WINDOWSSESSIONEXIT

donho (approved):

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/bd4c323d75bbe9a565e9bc31bb522133d78421a6

PR #12576, opened 2022-11-28
Make tab context menu customizable

### LINKED ISSUE #12170: Add customization for tab context menu
Would be nice if we can modify tab context menu in the same way that the main context menu (via an `.xml` file, even using the same `contextMenu.xml` file) or other way.

After merge https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12150 it's possible that not everyone will be happy to move some items to an additional sub-menu. First unhappy report https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12098#issuecomment-1245605230.

Temporary solution:
Now most (if not all) commands from the tab context menu are generally available, so they can be placed in the main context menu (and this menu we can modify as we want via `contextMenu.xml` file).

donho:
Use new added tabContextMenu_example.xml for the customization: 

1. rename to tabContextMenu.xml
2. modify it
3. copy it beside notepad++.exe or to %APPDATA%\Notepad++\  according to your Notepad++ conf
4. relaunch Notepad++

Fix #12170

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -386,41 +386,41 @@ The comments are here for explanation, it's not necessary to translate them.
             <Splitter>
             </Splitter>
             <TabBar>
-                    <Item CMID="0" name="Close"/>
-                    <Item CMID="1" name="Close All BUT This"/>
-                    <Item CMID="2" name="Save"/>
-                    <Item CMID="3" name="Save As..."/>
-                    <Item CMID="4" name="Print..."/>
-                    <Item CMID="5" name="Move to Other View"/>
-                    <Item CMID="6" name="Clone to Other View"/>
-                    <Item CMID="7" name="Copy Full File Path"/>
-                    <Item CMID="8" name="Copy Filename"/>
-                    <Item CMID="9" name="Copy Current Dir. Path"/>
-                    <Item CMID="10" name="Rename..."/>
-                    <Item CMID="11" name="Move to Recycle Bin"/>
-                    <Item CMID="12" name="Read-Only"/>
-                    <Item CMID="13" name="Clear Read-Only Flag"/>
-                    <Item CMID="14" name="Move to New Instance"/>
-                    <Item CMID="15" name="Open in New Instance"/>
-                    <Item CMID="16" name="Reload"/>
-                    <Item CMID="17" name="Close All to the Left"/>
-                    <Item CMID="18" name="Close All to the Right"/>
-                    <Item CMID="19" name="Open Containing Folder in Explorer"/>
-                    <Item CMID="20" name="Open Containing Folder in cmd"/>
-                    <Item CMID="21" name="Open in Default Viewer"/>
-                    <Item CMID="22" name="Close All Unchanged"/>
-                    <Item CMID="23" name="Open Containing Folder as Workspace"/>
-                    <Item CMID="24" name="Apply Color 1"/>
-                    <Item CMID="25" name="Apply Color 2"/>
-                    <Item CMID="26" name="Apply Color 3"/>
-                    <Item CMID="27" name="Apply Color 4"/>
-                    <Item CMID="28" name="Apply Color 5"/>
-                    <Item CMID="29" name="Remove Color"/>
-                    <Item CMID="30" name="Close Multiple Tabs"/>
-                    <Item CMID="31" name="Open into"/>
-                    <Item CMID="32" name="Copy to Clipboard"/>
-                    <Item CMID="33" name="Move Document"/>
-                    <Item CMID="34" name="Apply Color to Tab"/>
+                <Item CMDID="41003" name="Close"/>
+                <Item CMDID="0" name="Close Multiple Tabs"/>
+                <Item CMDID="41005" name="Close All BUT This"/>
+                <Item CMDID="41009" name="Close All to the Left"/>
+                <Item CMDID="41018" name="Close All to the Right"/>
+                <Item CMDID="41024" name="Close All Unchanged"/>
+                <Item CMDID="41006" name="Save"/>
+                <Item CMDID="41008" name="Save As..."/>
+                <Item CMDID="1" name="Open into"/>
+                <Item CMDID="41019" name="Open Containing Folder in Explorer"/>
+                <Item CMDID="41020" name="Open Containing Folder in cmd"/>
+                <Item CMDID="41025" name="Open Containing Folder as Workspace"/>
+                <Item CMDID="41023" name="Open in Default Viewer"/>
+                <Item CMDID="41017" name="Rename..."/>
+                <Item CMDID="41016" name="Move to Recycle Bin"/>
+                <Item CMDID="41014" name="Reload"/>
+                <Item CMDID="41010" name="Print..."/>
+                <Item CMDID="42028" name="Read-Only"/>
+                <Item CMDID="42033" name="Clear Read-Only Flag"/>
+                <Item CMDID="2" name="Copy to Clipboard"/>
+                <Item CMDID="42029" name="Copy Full File Path"/>
+                <Item CMDID="42030" name="Copy Filename"/>
+                <Item CMDID="42031" name="Copy Current Dir. Path"/>
+                <Item CMDID="3" name="Move Document"/>
+                <Item CMDID="10001" name="Move to Other View"/>
+                <Item CMDID="10002" name="Clone to Other View"/>
+                <Item CMDID="10003" name="Move to New Instance"/>
+                <Item CMDID="10004" name="Open in New Instance"/>
+                <Item CMDID="4" name="Apply Color to Tab"/>
+                <Item CMDID="44111" name="Apply Color 1"/>
+                <Item CMDID="44112" name="Apply Color 2"/>
+                <Item CMDID="44113" name="Apply Color 3"/>
+                <Item CMDID="44114" name="Apply Color 4"/>
+                <Item CMDID="44115" name="Apply Color 5"/>
+                <Item CMDID="44110" name="Remove Color"/>
             </TabBar>
         </Menu>
 
--- a/PowerEditor/installer/nativeLang/french.xml
+++ b/PowerEditor/installer/nativeLang/french.xml
@@ -386,41 +386,41 @@ The comments are here for explanation, it's not necessary to translate them.
             <Splitter>
             </Splitter>
             <TabBar>
-                    <Item CMID="0" name="Fermer"/>
-                    <Item CMID="1" name="Fermer tout sauf ce fichier"/>
-                    <Item CMID="2" name="Enregistrer"/>
-                    <Item CMID="3" name="Enregistrer sous..."/>
-                    <Item CMID="4" name="Imprimer"/>
-                    <Item CMID="5" name="Déplacer vers l'autre vue"/>
-                    <Item CMID="6" name="Cloner dans l'autre vue"/>
-                    <Item CMID="7" name="Copier le chemin complet du document actuel"/>
-                    <Item CMID="8" name="Copier le nom du fichier actuel"/>
-                    <Item CMID="9" name="Copier le dossier du fichier actuel"/>
-                    <Item CMID="10" name="Renommer le fichier actuel"/>
-                    <Item CMID="11" name="Supprimer le fichier actuel"/>
-                    <Item CMID="12" name="Lecture seule pour le document actuel"/>
-                    <Item CMID="13" name="Enlever l’attribut de lecture seule du fichier"/>
-                    <Item CMID="14" name="Déplacer vers une nouvelle fenêtre"/>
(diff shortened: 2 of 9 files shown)


Yaron10:
@donho,

**Tab menu, `tabContextMenu.xml` exists.**
`Reload from Disk` is wrongly disabled if file is saved.
`Reload from Disk` is correctly disabled if doc is new. 

**Tab menu, `tabContextMenu.xml` does not exist.** 
`Reload from Disk` is _wrongly enabled_ if doc is new. 
(This started in an older tab-changes commit).

Thank you.

donho:
@Yaron10 
Thank you for testing it.

> This is an old issue

Could you indicate the issue number please?

Yaron10:
@donho,

> This is an old issue.

I meant "not caused by this PR".
I've opened a new issue: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12579.

Yaron10:
@donho,

Another minor issue:
On changing localization, the tab context-menu is displayed in the previous language (until the next startup).

@CennoxX,

If the tab context-menu was restored to the old layout (except for the new `Add Color to Tab` which should be in a sub-menu), would you still be interested in the new customization?

CennoxX:
> @CennoxX,
> 
> If the tab context-menu was restored to the old layout (except for the new `Add Color to Tab` which should be in a sub-menu), would you still be interested in the new customization?
I like the redesigned tab context menu layout, but I saw some users that don't like it. That's why I like this PR.

alankilborn:
>If the tab context-menu was restored to the old layout (except for the new Add Color to Tab which should be in a sub-menu)...

Are there plans for that reversion to the old layout as the standard?

Yaron10:
@CennoxX,

Thank you for replying. 
Understanding that you and @alankilborn prefer the new layout, I like this PR. :)

@alankilborn,

Personally, I prefer the old layout.
But I got the hint. :)

alankilborn:
>...and @alankilborn prefer the new layout

Actually, I have no preference as I have been overriding the tab context menu in a script for a long time now, figuring we'd never get customization capability.

But I think the fact that customization capability has come is very important (to others, not to me).

Yaron10:
@alankilborn,

Well, we should thank Don for his work here.

donho:
Sorry, I didn't follow you: What's the "New Layout" and "Old Layout"?

Using tabContextMenu.xml should have the same menu structure as the default one. The difference is only for some command item labels.

donho:
> Another minor issue:
> On changing localization, the tab context-menu is displayed in the previous language (until the next startup).

@Yaron10 It's fixed in the latest commit now.

Note:
 
1. You have use the new english.xml in the commit.
2. The customized sub-menu entry cannot be translated. User can use his/her native language as value of attribute "FolderName" in tabContextMenu.xml file.
3. User can always add any command beyond the default commands in tabContextMenu.xml file. But such command won't be translated.

Yaron10:
@donho,

👍 
Great.
Thank you for your work.

alankilborn:
The `tabContextMenu_example.xml` that is part of this PR provides the menu as it appears in 8.4.6 and later by default.  

If one wishes to recreate the tab context menu as it was in 8.4.5 and earlier, I have attached a sample file to do that here, as a TXT file (because XML files can't be added here).

[tabContextMenu_8.4.5_and_earlier.TXT](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10128562/tabContextMenu_8.4.5_and_earlier.TXT)

alankilborn:
@donho 

We have this currently:

image

Shouldn't there also be a menu entry here for `Edit Tab contextMenu` ?

alankilborn:
I see, but it is inconsistent.
And, being different, it (obviously) confused me.
I would think it would confuse other users (at least the ones that know you can "edit popup contextmenu").

alankilborn:
If you're going to provide `tabContextMenu_example.xml` to users, wouldn't it be better to have the example do something different than the default?

I mean, if a user tries activating it by renaming `tabContextMenu_example.xml` to `tabContextMenu.xml` (as a first step, without changing the contents of the file) and restarting, then invokes right-click-on-a-tab, I don't know if he gets a good feeling that anything happened, i.e., that it even affected the menu.

(Actually, one difference is that _Recycle Bin_ is missing, but it is very subtle)

Why not have the example file provide the context menu as it was in 8.4.5 and earlier, as there have been users wanting the non-submenu'd menu back?  It would be an easy solution on the support site for those that provide support:  "Rename the example file and restart--done"

donho:
@alankilborn 
> And, being different, it (obviously) confused me.

From the view of users, they see only the possibility to modify Edit Zone CM, that's my goal, and here's the reasons:
1. Default Edit Zone CM comes with a xml file, whereas Default Tab CM is hard coded.
2. By Edit Zone's nature (selected text), there are more commands (which can operate on selected text) can be added in Default Edit Zone CM - so more users need to modify Edit Zone CM; whereas Tab CM is limited for file operation, and I believe the essential commands are already in the Default Tab CM - I think the most of users can get used to the new layout, but for those who can't, they are be able to follow https://npp-user-manual.org/ to get what they want.
3. As an xml file is provided to override Tab CM, user can add whichever command, but the translation is limited only for the default Tab CM commands.



> If you're going to provide `tabContextMenu_example.xml` to users, wouldn't it be better to have the example do something different than the default?

The goal of providing the example xml is for showing the whole set of default commands and the all the possibilities (adding command, adding sub-menu entries and adding the separators).

> (Actually, one difference is that Recycle Bin is missing, but it is very subtle)

It's been forgotten, add it in: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c1cb0022c47eae0ef8f1e3872eb7d2969a3ac0ff



> Why not have the example file provide the context menu as it was in 8.4.5 and earlier, as there have been users wanting the non-submenu'd version back? It would be an easy solution on the support site for those that provide support: "Rename the example file and restart--done"

That's a good idea. Could you provide such file here, and it can be included in https://npp-user-manual.org/ 
 @pryrt what do you think about it?

alankilborn:
>>Why not have the example file provide the context menu as it was in 8.4.5 and earlier

>That's a good idea. Could you provide such file here

I tried to here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12576#issuecomment-1333032566
But for some reason, that now fails to download for me, so here's the text of it:

```
<!--
By modifying this file, you can customize your context menu popuped as right clicking on the tab zone.
It may be more convinient to access to your frequent used commands via context menu than via the top menu.

Please check "How to Customize the Context Menu" on:
https://npp-user-manual.org/docs/config-files/#the-context-menu-contextmenu-xml
-->
<NotepadPlus>
    <TabContextMenu>
        <!--
        Use MenuEntryName and MenuItemName to localize your commands to add.
        The values should be in English but not in translated language.
        (You can set Notepad++ language back to English from Preferences dialog via menu "Settings->Preferences...")
        -->
        <Item MenuEntryName="File" MenuItemName="Close"/>

        <!--
        Use FolderName (optional) to create sub-menu. FolderName value can be in any language (French, Japanese...).
        -->
        <Item MenuEntryName="File" MenuItemName="Close All BUT Active Document" ItemNameAs="Close All BUT This"/>
        <Item MenuEntryName="File" MenuItemName="Close All to the Left"/>
        <Item MenuEntryName="File" MenuItemName="Close All to the Right"/>
        <Item MenuEntryName="File" MenuItemName="Close All Unchanged"/>
        <Item MenuEntryName="File" MenuItemName="Save"/>
        <Item MenuEntryName="File" MenuItemName="Save As..."/>
        <Item MenuEntryName="File" MenuItemName="Rename..."/>
        <Item MenuEntryName="File" MenuItemName="Move to Recycle Bin"/>
        <Item MenuEntryName="File" MenuItemName="Reload from Disk" ItemNameAs="Reload"/>
        <Item MenuEntryName="File" MenuItemName="Print..."/>

        <Item id="0"/>      <!-- id="0" is the separator -->

        <Item MenuEntryName="File" MenuItemName="Explorer" ItemNameAs="Open Containing Folder in Explorer"/>
        <Item MenuEntryName="File" MenuItemName="cmd" ItemNameAs="Open Containing Folder in cmd"/>
        <Item MenuEntryName="File" MenuItemName="Folder as Workspace" ItemNameAs="Open Containing Folder as Workspace"/>

        <Item id="0"/>      <!-- id="0" is the separator -->

        <Item MenuEntryName="File" MenuItemName="Open in Default Viewer"/>

        <Item id="0"/>      <!-- id="0" is the separator -->

        <Item MenuEntryName="Edit" MenuItemName="Set Read-Only" ItemNameAs="Read-Only"/>
        <Item MenuEntryName="Edit" MenuItemName="Clear Read-Only Flag"/>

        <Item id="0"/>      <!-- id="0" is the separator -->

        <Item MenuEntryName="Edit" MenuItemName="Copy Current Full File Path" ItemNameAs="Full File Path to Clipboard"/>
        <Item MenuEntryName="Edit" MenuItemName="Copy Current Filename" ItemNameAs="Filename to Clipboard"/>
        <Item MenuEntryName="Edit" MenuItemName="Copy Current Dir. Path" ItemNameAs="Current Dir. Path to Clipboard"/>

        <Item id="0"/>      <!-- id="0" is the separator -->

        <Item MenuEntryName="View" MenuItemName="Move to Other View"/>
        <Item MenuEntryName="View" MenuItemName="Clone to Other View"/>
        <Item MenuEntryName="View" MenuItemName="Move to New Instance"/>
        <Item MenuEntryName="View" MenuItemName="Open in New Instance"/>

        <Item id="0"/>      <!-- id="0" is the separator -->

        <Item FolderName="Apply Color to Tab" MenuEntryName="View" MenuItemName="Apply Color 1"/>
        <Item FolderName="Apply Color to Tab" MenuEntryName="View" MenuItemName="Apply Color 2"/>
        <Item FolderName="Apply Color to Tab" MenuEntryName="View" MenuItemName="Apply Color 3"/>
        <Item FolderName="Apply Color to Tab" MenuEntryName="View" MenuItemName="Apply Color 4"/>
        <Item FolderName="Apply Color to Tab" MenuEntryName="View" MenuItemName="Apply Color 5"/>
        <Item FolderName="Apply Color to Tab" MenuEntryName="View" MenuItemName="Remove Color"/>

    </TabContextMenu>
</NotepadPlus>
```

donho:
@alankilborn @pryrt 
I think it's win win solution: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f400b682b509deae1c60f38d6522c967c6b6b2c4

Yaron10:
@alankilborn,

> Why not have the example file provide the context menu as it was in 8.4.5 and earlier, as there have been users wanting the non-submenu'd version back? It would be an easy solution on the support site for those that provide support: "Rename the example file and restart--done"

👍 
Great!

> @Yaron10 Would you be so kind as to test this file?

Thank you for providing this file.
The `ItemNameAs` entries override the localization. Shouldn't they be removed?

alankilborn:
>The ItemNameAs entries override the localization. Shouldn't they be removed?

If they are removed, then maybe some entries don't make as much sense, examples:

* Explorer
* cmd

..but, not showing _Open Containing Folder in_ in front of these is not a critical problem.

My goal with the file was to recreate the EXACT experience of the "old" (pre-8.4.6) tab context menu, for the purists that demand that back (with the noted exception of the tab-coloring submenu at the very bottom).  Well, at least for the English-using purists. :-)

I'd agree to removing `ItemNameAs=` sections as reasonable.

Yaron10:
> If they are removed, then maybe some entries don't make as much sense, examples:

That's a good point.

It reminds of a PR (accepted) which added `<MainCommandNames>` to the localization files and *a lot of code*.
I didn't understand why the simpler solution was not preferred: use *full commands* in sub-menus as well.
For example: `Open Containing Folder in Explorer` instead of  `Explorer`.

This would have solved our problem here as well.

alankilborn:
>It reminds of a PR (accepted) which added MainCommandNames to the localization files

As a user, the result of that seemed to make good sense.
As to the implementation adding "a lot of code", that I don't know.

>use full commands in sub-menus as well.
For example: Open Containing Folder in Explorer instead of Explorer
This would have solved our problem here as well.

Seems very reasonable.

Yaron10:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ddd2e5fd23412d5e8410c2a821461f1f578bca0f

> As a user, the result of that seemed to make good sense.

Indeed.
But if I remember the issue correctly, using full commands in sub-menus would have solved that too.

alankilborn:
>using full commands in sub-menus would have solved that too

I'm sure but often the full command there feels like double-talk, because a good portion of it is in the submenu.

Yaron10:
> I'm sure but often the full command there feels like double-talk

Yes, I understand that.
But is `Open Containing Folder in Explorer` so bad?
Anyway, https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ddd2e5fd23412d5e8410c2a821461f1f578bca0f is probably not going to be reverted. :)

Issue #12545, opened 2022-11-21
How operate on selection and last empty line Change for Remove Empty Lines command


ArkadiuszMichalski:
Context/discussion and some build for test was here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535.

For `Remove Empty Lines` command actual implementation works like this:
1. Selecting something until the end (with last EOL)
2. Selecting by Ctrl+A (this also including last EOL)
3. Not selecting (which means the entire document)

Each of the above cases removes the last blank line (which is consistent). Graphically it looks like this:

Before invoking `Remove Empty Lines`:
![image](https://user-images.githubusercontent.com/2730894/202935326-92d42b60-0be6-4ea3-97b5-daf614f02d8f.png)
or
![image](https://user-images.githubusercontent.com/2730894/202935339-0f8e8d5f-0048-4116-8d8e-79a153726506.png)
or
![image](https://user-images.githubusercontent.com/2730894/202935369-6f3e03fc-ac68-4f0b-83e6-d8914a1da13c.png)
After invoking `Remove Empty Lines`:
![image](https://user-images.githubusercontent.com/2730894/202935386-7d465a36-5dfb-40b3-a48e-67b3a1048fc8.png)

For some people it may be incorrect that such selection deletes the last line because the status bar clearly indicates that the line is not selected (like belowe `Sel 14|2`).
![image](https://user-images.githubusercontent.com/2730894/202935339-0f8e8d5f-0048-4116-8d8e-79a153726506.png)
![image](https://user-images.githubusercontent.com/2730894/202935633-54ba15e4-275d-465e-9aba-61443ef83faf.png)
Of course, this is a bit more complicated matter, because the last `CRLF` in line 2 has been selected, and it is responsible for the appearance of line 3.

The sort commands treating the last EOL differently with or without selection, see https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1319986638 or https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1319986638. , Such behavior, although it looks strange, it's not such unusual (I see that Sublime Text works identically).

For the test, I made the `Change for Remove Empty Lines` command work similarly, i.e .:
1. Without selection it include the last empty line.
2. With selection it omit the last empty line.
[...]

PR #12580, opened 2022-11-28
Change for Remove Empty Lines command how operate on selection and last empty line

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12545.

I looked at all operations from `Edit>Line Operations` how they work with and without selection (`Sort`, `Reverse`, `Randomize` and `Remove empty`). And all except (`Remove empty`):
1. Without selection they include the last empty line.
2. With selection they omit the last empty line.
3. For partially selected lines they automatically extend the selection to the entire lines. For sorting it's obvious. But for other cases it also makes sense because we operate on the lines and we do not have to worry about the full (correct) line selecting.

\* `Join lines` command fit only to points 2 and 3 (not work without selection, maybe should).

Taking the above into account, I also made `Remove empty` consistent with this behavior.

It needs to be clearly stated that these listed sort commands are basically a variant of one command, so other commands should not necessarily behave the same.

This PR is for testing purposes to produce different behavior variants for this `Remove Empty Lines` command.

Continuation of https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1505,8 +1505,9 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	auto mainSelCaretPos = _pEditView->execute(SCI_GETCURRENTPOS);
 	auto recSelAnchorVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONANCHORVIRTUALSPACE);
 	auto recSelCaretVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONCARETVIRTUALSPACE);
-	bool isSelRec = _pEditView->execute(SCI_GETSELECTIONS) > 1 && _pEditView->execute(SCI_GETSELECTIONMODE) == 1;
+	bool isRecSel = _pEditView->execute(SCI_GETSELECTIONS) > 1 && _pEditView->execute(SCI_GETSELECTIONMODE) == 1;
 	bool isEntireDoc = mainSelAnchor == mainSelCaretPos;
+	bool delRecSel = false;
 	int nbTotal = 0;
 	if (!isEntireDoc)
 	{
@@ -1514,6 +1515,12 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
 		auto newSelStart = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.first);
 		auto newSelEnd = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.second) + _pEditView->execute(SCI_LINELENGTH, lineRange.second);
+		if (isRecSel){
+			auto newLineEnd = _pEditView->execute(SCI_LINEFROMPOSITION, _pEditView->execute(SCI_GETSELECTIONEND));
+			newSelEnd = _pEditView->execute(SCI_POSITIONFROMLINE, newLineEnd) + _pEditView->execute(SCI_LINELENGTH, newLineEnd);
+			if (newLineEnd == _pEditView->execute(SCI_GETLINECOUNT) - 1)
+				delRecSel = true;
+		}
 		_pEditView->execute(SCI_SETSEL, newSelStart, newSelEnd);
 	}
 	nbTotal = _findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
@@ -1529,11 +1536,11 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
 		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
 		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
-		if (lineRange.second != static_cast<size_t>(lastLineDoc))
+		if ((!isRecSel && lineRange.second != static_cast<size_t>(lastLineDoc)) || (isRecSel && !delRecSel))
 		{
 			if (nbTotal == 0)
 			{
-				if (isSelRec)
+				if (isRecSel)
 				{
 					_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHOR, mainSelAnchor);
 					_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHORVIRTUALSPACE, recSelAnchorVirt);
@@ -1561,7 +1568,7 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	}
 	else if (!isEntireDoc && nbTotal == 0)
 	{
-		if (isSelRec)
+		if (isRecSel)
 		{
 			_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHOR, mainSelAnchor);
 			_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHORVIRTUALSPACE, recSelAnchorVirt);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1501,30 +1501,73 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	}
 	env._str4Replace = TEXT("");
 	env._searchType = FindRegex;
-	auto mainSelStart = _pEditView->execute(SCI_GETSELECTIONSTART);
-	auto mainSelEnd = _pEditView->execute(SCI_GETSELECTIONEND);
-	auto mainSelLength = mainSelEnd - mainSelStart;
-	bool isEntireDoc = mainSelLength == 0;
-	env._isInSelection = !isEntireDoc;
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
+	auto mainSelAnchor = _pEditView->execute(SCI_GETANCHOR);
+	auto mainSelCaretPos = _pEditView->execute(SCI_GETCURRENTPOS);
+	bool isSelRec = _pEditView->execute(SCI_GETSELECTIONS) > 1 && _pEditView->execute(SCI_GETSELECTIONMODE) == 1;
+	bool isEntireDoc = mainSelAnchor == mainSelCaretPos;
+	int nbTotal = 0;
+	if (!isEntireDoc)
+	{
+		env._isInSelection = !isEntireDoc;
+		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
+		auto newSelStart = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.first);
+		auto newSelEnd = _pEditView->execute(SCI_POSITIONFROMLINE, lineRange.second) + _pEditView->execute(SCI_LINELENGTH, lineRange.second);
+		_pEditView->execute(SCI_SETSEL, newSelStart, newSelEnd);
+	}
+	nbTotal = _findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 
 	// remove the last line if it's an empty line.
-	auto lastLineDoc = _pEditView->execute(SCI_GETLINECOUNT) - 1;
+	auto lineCount = _pEditView->execute(SCI_GETLINECOUNT);
+	auto lastLineDoc = lineCount - 1;
 	auto str2Search = isBlankContained ? TEXT("[\\r\\n]+^[\\t ]*$|^[\\t ]+$") : TEXT("[\\r\\n]+^$");
 	auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, lastLineDoc - 1);
 	auto endPos = _pEditView->execute(SCI_GETLENGTH);
 	if (!isEntireDoc)
 	{
+		pair<size_t, size_t> lineRange = _pEditView->getSelectionLinesRange();
 		startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
 		endPos = _pEditView->execute(SCI_GETSELECTIONEND);
-		auto endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
-		if (endPos != (_pEditView->execute(SCI_POSITIONFROMLINE, endLine) + _pEditView->execute(SCI_LINELENGTH, endLine)))
+		if (lineRange.second != static_cast<size_t>(lastLineDoc))
+		{
+			if (nbTotal == 0)
+			{
+				if (isSelRec)
+				{
+					_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHOR, mainSelAnchor);
+					_pEditView->execute(SCI_SETRECTANGULARSELECTIONCARET, mainSelCaretPos);
+				}
+				else
+				{
+				_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
+				_pEditView->execute(SCI_SETCURRENTPOS, mainSelCaretPos);
+				}
+			}
 			return;
+		}
+		else if (lineCount > 1 && (endPos - startPos > 0))
+		{
+			startPos = _pEditView->execute(SCI_GETLINEENDPOSITION, lineRange.first - 1);
+		}
 	}
 	_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP|SCFIND_POSIX);
 	auto posFound = _pEditView->searchInTarget(str2Search, lstrlen(str2Search), startPos, endPos);
 	if (posFound >= 0)
+	{
 		_pEditView->replaceTarget(TEXT(""), posFound, endPos);
+	}
+	else if (!isEntireDoc && nbTotal == 0)
+	{
+		if (isSelRec)
+		{
+			_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHOR, mainSelAnchor);
+			_pEditView->execute(SCI_SETRECTANGULARSELECTIONCARET, mainSelCaretPos);
+		}
+		else
+		{
+		_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
+		_pEditView->execute(SCI_SETCURRENTPOS, mainSelCaretPos);
+		}
+	}
 }
 
 void Notepad_plus::removeDuplicateLines()
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1506,7 +1506,7 @@ void Notepad_plus::removeEmptyLine(bool isBlankContained)
 	auto recSelAnchorVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONANCHORVIRTUALSPACE);
 	auto recSelCaretVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONCARETVIRTUALSPACE);
 	bool isRecSel = _pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE || _pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_THIN;
-	bool isEntireDoc = mainSelAnchor == mainSelCaretPos && !isRecSel;
+	bool isEntireDoc = mainSelAnchor == mainSelCaretPos && recSelAnchorVirt == recSelCaretVirt;
 	auto docLength = _pEditView->execute(SCI_GETLENGTH);
 	bool checkLastLine = true;
 	bool isRecSelLastLine = false;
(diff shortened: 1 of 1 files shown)


Yaron10:
@ArkadiuszMichalski,

👍 
Great work.

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12535#issuecomment-1326817824
> Or...
In real life - is any user going to hold Alt and select column-block for removing empty lines?
I think it's only us testing the various possible cases.

Do you think it's worth the extra code?
It gets more complex. :)

תמונה

Line 2 is empty.
Alt-select `st`+ part of line 2.
Remove Empty Lines.

Result:
תמונה

There might be some more cases. Why complicate the code?
I think that "do nothing" on column-selection would be better.

Thank you.

ArkadiuszMichalski:
@Yaron10 
I'll try to support this column mode in some expected way. If necessary we can easy remove this additional code and do nothing for column mode.

Yaron10:
@ArkadiuszMichalski,

As you understand.

Line 2 is still not deleted.

Thanks again.

ArkadiuszMichalski:
@Yaron10 
>Line 2 is still not deleted.

And it shouldn't be like that? This case can be treated as if the 2nd line has not been selected, because the last empty one cannot be selected? On status bar we have `2x2 -> 2` so in fact nothing has been selected from the second line. 

For such standard selection you also can't delete this last line:
image
image

Yaron10:
@ArkadiuszMichalski,

> On status bar we have 2x2 -> 2 so in fact nothing has been selected from the second line.

Good point. But you still have `2x2` and you see the selection background.
I know we're entering the extreme-corner-cases/arguments zone. But that's my point: in real life, column-selection would probably be irrelevant to `Remove Empty`.

In the following case, line 1 IS deleted although you have `2x2 -> 2`.
תמונה

And just to make the comparison more accurate, in the following case line 2 is NOT deleted.
תמונה

***
I'll continue tomorrow.

ArkadiuszMichalski:
@Yaron10 
But we in these commands expand the selections first. For column mode it would be equivalent in the form of line selection.
So for your first example it basically means:
image
and now we're doing line deletion.

Some more sophisticated column mode support is rather unnecessary for these commands. because there would not be any profit here. These commands delete entire lines, so extending partial selection to entire lines makes sense.

In the second example something is wrong because it should be:
image

I'll check later why this is happening because even this not work:
image

ArkadiuszMichalski:
I'm not insisting, it just worked so far (I don't go into how) so it may continue to work. But it would be easiest to just do nothing for this case.

There is a big difference between normal selection and block selection here. In a normal selection we cannot select the last empty line, but in a block looks like we can.

image
image

vs

image
image

So this complicates the conversion of a block selection into a single one. I'll see if it can be easily fixed with what it is now. I certainly have no intention of adding a new separate ligic to handle block selection here.

alankilborn:
If you have this:

image

or you have this:

image

You get the same result in either case to both SCI_GETSELECTIONSTART and SCI_GETSELECTIONEND.

"Line" 3 in the bottom screenshot is entirely in "virtual space" and thus the "selected" part shown for it really doesn't come into play.

>...what the status bar shows

What is shown in the status bar is "contrived".
I wouldn't take the status bar too seriously. :-)

Yaron10:
> it just worked so far (I don't go into how) so it may continue to work.

I didn't delve into your code, but it seems you've added quite a few lines for handling column-selection.
If I'm wrong and you can not shorten the code by "do nothing", I'm retracting my comments. :) 

> At least that's what the status bar shows.

It's also what we should assume the user wants to do.
_If we think that any user is going to hold Alt and column-select for removing empty lines._

> I'll see if it can be easily fixed with what it is now. I certainly have no intention of adding a new separate ligic to handle block selection here.

👍

alankilborn:
>You get the same result in either case to both SCI_GETSELECTIONSTART and SCI_GETSELECTIONEND.

Since this is true, can't the selection type when the user runs the command be mostly ignored?
It should work if using those two Scintilla commands (which are then expanded to "whole line" positions).

So the only case it should not handle is multiple stream selections?  In that case it would only operate on the "main" selection's lines?

It should handle fully, for no special coding:

* single stream selection
* thin column selection (zero width)
* rectangular column selection (non-zero width)

Yaron10:
@alankilborn,

Welcome back. :)

Do you agree that processing column-selection or "do nothing" should be determined by the "shorter-code" factor?

alankilborn:
>Welcome back.

Sorry for popping in/out.  I just don't always have time to fully contribute...

>agree that processing column-selection or "do nothing" should be determined by the "shorter-code" factor.

That seems reasonable, although from my earlier comments I don't see why column/thin selection would require any specialized coding.

ArkadiuszMichalski:
@alankilborn 
>You get the same result in either case to both SCI_GETSELECTIONSTART and SCI_GETSELECTIONEND.

But here is important how we obtain the first and last lines, directly from this points or use `getSelectionLinesRange()`, which makes some correction.

@Yaron10 
>I didn't delve into your code, but it seems you've added quite a few lines for handling column-selection.
If I'm wrong and you can not shorten the code by "do nothing", I'm retracting my comments. :)

These are still just additional conditions. It's easy to remove if needed.

>And as I wrote earlier:
>>I didn't delve into your code, but it seems you've added quite a few lines for handling column-selection.
If I'm wrong and you can not shorten the code by "do nothing", I'm retracting my comments.

Column selection has one additional advantage over standard selection, if we want to remove the last empty line too then we can do that, same as for the entire document. For standard selection this is impossible because it cannot be selected (of course I'm skipping the solution of removing the EOL from the previous line).

So:
image
will not remove line 2.
but:
image
will remove line 2.

You can check the last PR to see if its behavior is more appropriate.

Yaron10:
@ArkadiuszMichalski,

I appreciate your perseverance.
**If you don't want me to further comment about column-selection, please let me know. :)**

In the last commit there are ~15 lines for handling column-selection.
If it can't be implemented as @alankilborn suggested, why not return at the beginning? 
```
if (isRecSel)
   return;
```

> Column selection has one additional advantage over standard selection...

That's a nice point. But is it worth it?

And I've encountered some more issues.
***
תמונה

Remove Empty Lines.

Result:
תמונה

Expected:
3 lines only.
***
תמונה

Remove Empty Lines.

Result:
תמונה

alankilborn:
>If it can't be implemented as @alankilborn suggested

My suggestion doesn't "fit" as @ArkadiuszMichalski wants to use:

>getSelectionLinesRange(), which makes some correction

And I suppose that makes sense.

-------------------------------------------

Perhaps doing nothing on any type of column selection (thin or not) is best.
Be careful if trapping out column-selection, it is not as simple as testing for SCI_SELECTIONISRECTANGLE.

Yaron10:
@alankilborn,

Thank you for the clarification and for the additional info.
Is the term "Encyclopedia" still used these days? :)

ArkadiuszMichalski:
>@alankilborn
>You get the same result in either case to both SCI_GETSELECTIONSTART and SCI_GETSELECTIONEND.
>Since this is true, can't the selection type when the user runs the command be mostly ignored?
>It should work if using those two Scintilla commands (which are then expanded to "whole line" positions).
>It should handle fully, for no special coding:
>single stream selection
>thin column selection (zero width)
>rectangular column selection (non-zero width)

And it might even be so if it wasn't for the fact that it's not entirely consistent with what the status bar shows. And let me remind you that it didn't bother me at the beginning :)

@Yaron10
>If it can't be implemented as @alankilborn suggested, why not return at the beginning?
```
if (isRecSel)
   return;
```

I wrote that for this case (column mode) is simplest to do nothing, but this always can be achieved by adding the above at the beginning (and clean up rest). The same basically applies to multiselections, because it doesn't support this mode anyway.

I will experiment with this column mode to make it work as expected (even in line with bar status). But that doesn't mean it will be accepted in the end.

@Yaron10 
Try the last PR. I wonder if you'll find anything else :)

alankilborn:
>But that doesn't mean it will be accepted in the end.

Yep, I've often thought that over-thinking (and over-discussing) a feature makes Don quick to pick up his REJECT stamp.  :-)

Yaron10:
@ArkadiuszMichalski,

> I wonder if you'll find anything else.

The last commit seems very good.
But "I'll be back" (with the correct accent :) ).

I'll be back late this evening and have another look.
Anyway, I still think that column-selection would not be useful in real life.

Thank you for your work.

Yaron10:
@ArkadiuszMichalski,

The final implementation seems very good and solid.  👍 
I must mention again that I'm impressed by your patience and persistence. :)

Thank you.

@alankilborn,

Thank you too for your patience and very helpful comments. 👍

donho:
@ArkadiuszMichalski
> This PR is for testing purposes to produce different behavior variants for this `Remove Empty Lines` command.

So it's not for merging, is it?

ArkadiuszMichalski:
@donho 

Last commit can be considered to merging. All reported issues have been fixed. In short new behavior looks like this:
1. For partially selected lines they automatically extend the selection to the entire lines.
2. If nothing has been removed then the original selection is restored (before expand).
3. For selection the last empty line is treat in a consistent way (with what we see visually and what the status bar shows).
4. It works fine for rectangular selection.

So now it works consistently. But if any of the above changes are not accepted then can be removed. Based on the behavior established here the other line methods can be corrected to work similarly. But we must have at least one reference method of how it should look like.

donho:
I have no opinion about column mode, though I agree that we can just not do the remove empty lines operation when column mode is detected (as a lazy dev). But since it (column mode) is treated correctly, I've got nothing to say about it.

OTOH, the last blank line is (almost) impossible to be removed, except without selection (entire doc mode).
Is it a normal behaviour (Ctrl-A + empty lines operation) for users?

donho (changes requested):
Please consider the following remark.
~~Maybe redefine the last blank line with selection behaviour.~~
*Edit: last blank line with selection behaviour in this PR is a regression to me.*
*Please restore the current behaviour of last blank line with selection.*

donho:
@Yaron10 
Sorry I don't quite understand your understanding of what I say, let me make it more clear:

In this PR, the last blank line can be removed only in entire doc mode (No selection).
It's a regression to me because in the current version v8.4.7
* Ctrl-A + empty lines operation : remove the last blank line
* Manually selecting from 0 to the last 2 bytes CRLF  + empty lines operation : remove the last blank line
* Select last 2 bytes CRLF  + empty lines operation : remove the last blank line

Yaron10:
@donho,

תמונה

Line 2 is blank.
Ctrl+A.

See the status bar:
תמונה

You have 2 lines.
You've **selected all**.
Selected lines: **1**.

Is that correct?
If you think the current status bar implementation (**and the concept behind it**) is correct, then we can NOT treat the second line as selected.
"Select All" means select all that can be selected. And an empty line can not be selected.

The `Remove Empty Lines` code was written before the selection-concept changed, and was never updated.
Other operations (e.g. Sort) have been updated, and this PR is updating the `Remove Empty` operation and fixing the inconsistent behavior in v8.4.7.

donho:
@Yaron10 
I see your arguments. And I do understand them from the viewpoint of dev.
However in the shoes of users, when all the document is selected, and empty lines operation is applied, I would surprise about the last blank line is still there. The current behaviour for me is correct, if the PR was merged as it is, then there would be a regression, and people would be complain.

Regarding the info shown on status bar - I would say it's correct also, because it reflects graphically the selection, from the viewpoint of user. The 2 functions are not associated necessarily.

donho:
> That is: on Ctrl+A the line is excluded, whereas if nothing is selected - that line is included

Ctrl+A & nothing is selected should have the same behaviour - last line get removed.

Yaron10:
I respectfully disagree. But it's your code.

Yaron10:
Don,

Before making the final decision, please let @alankilborn and @ArkadiuszMichalski chime in.
They might convince you. :)

I must retire now.

donho:
@Yaron10 
Thank you for your suggestion!

alankilborn:
> Ctrl+A & nothing is selected should have the same behaviour

I agree with this.  (Sorry Yaron10.)
@donho, is this a global statement?  Meaning for ALL commands?  I think so, and hope you agree.

ArkadiuszMichalski:
@donho 

At some point someone changed the behavior for selection and last blank line.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/5867025842a45b4793e068684accb5a22fb5d3d0/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L2898
```
		// if the end of the selection includes the line-ending, 
		// then don't include the following line in the range
```

Wherever the `getSelectionLinesRange()` method is used, the last blank line is ignored. This includes, for example, what the status bar shows and working for all sorting methods. Invoking these methods with no selection or all selection (Ctrl+A) has a different effect. Let me just remind once again that this is not some weirdness at all, I saw the same thing in Sublime Text for sorting.
This was probably some prelude to unifying this behavior everywhere, but it was only done for sorting methods.

From the perspective of a standard user this is what we see visually (visually this last blank line is not selectable in a standard way). 
image
image
Of course, I omit the fact that the EOL in the 4th line is responsible for the appearance of the 5th line, but the standard user does not need to know what EOL is.
image
image

No matter how I look at it, I selected 4 lines here, 3 are empty (2,3,4) and only those line are deleted, which is true. I'm not arguing that because that's what I see on screen.

>OTOH, the last blank line is (almost) impossible to be removed, except without selection (entire doc mode).
Is it a normal behavior (Ctrl-A + empty lines operation) for users?

It can be selected by column selection, which is exactly in line with the visual presentation:
image
image
or even:
image
image

In the above I really see that this last blank line is part of selection.

**Due of how to treat this last blank line this PR and discussion is so long.** 


If Ctrl+A must be the same as no selection, what about the situation when we selecting something from the end?
image
I've selected line 3 and 4 and because of autoexpand this line will be removed (on this PR).  What should be done after reverting behavior, select 3-4 or 3-5 (assuming we keep the autoexpand)?

Same question for this:
image
image
or
image
image

Visually, I only select 2 lines (3-4), but if I have to restore the previous behavior, what should happen here?

In general, if we assume that a standard selection (full or partial from the end) can cover the last blank line, it probably doesn't make sense to make any changes and complicate the whole thing. If someone wants to remove the last blank line it should make selection including last EOL in document himself. The current implementation works according to this approach, even with block selection (at least for these two methods).

I will release one more commit with the fixes suggested by Don. If in the future the behavior for the last blank line will changed, there will already be some base code for further work.

donho:
@alankilborn 
> @donho, is this a global statement? Meaning for ALL commands? I think so, and hope you agree.

If you mean all Line Operations should have the same behaviour to process the last blank line, my reply is: yes, they should have, and it's better that they treat the last blank line in order to keep the consistency - for the reason that all lines are treated, but not for aligning the behaviour with other Line Operations.

OTOH the last blank line have the different visibility in the different Line Operations, in the eyes of users.
In "Remove Empty Lines", users focus on removing the empty lines - so while he/she selects the whole document (mouse selection or Ctrl-A), applies "Remove Empty Lines", and find the last blank line is still there - it's a bug.

For "Lines sort" operations or other ones ("Remove Duplicated Lines", "Join Lines", "Reverse Line Order"...), however, it's another story. A lot of people don't notice or even don't care of the fact that the last blank line is excluded in the operations. As the last non-empty line is processed, it doesn't matter for them.

So here is my opinion about make all Line Operations consistent: If one of you want to "align" the behaviour of all Line Operations, the last blank line should be included, but not excluded. From my side, I will make sure the PR is simple and no risk of regression, to decide if the PR is accepted - after all, it's rather BETTER TO HAVE than MUST HAVE feature.

Yaron10:
@alankilborn,

> I agree with this. (Sorry Yaron10.)

No need to be sorry. :)
There's an old saying (loosely translated): "Just as people's faces are not the same, so their opinions are not the same".

Unlike Don, you think that the last empty line should be excluded even if nothing is selected.

***
תמונה

If we could differentiate between Ctrl+A  and selecting the first line only, I'd agree Ctrl+A should be treated as "entire doc".

An example for Select All:
תמונה

But we can not make that distinction, and therefore I think the commands should be in-sync with the visual selection and the status bar.

Thank you all.

donho:
@ArkadiuszMichalski 


> Wherever the getSelectionLinesRange() method is used, the last blank line is ignored. This includes, for example, what the status bar shows and working for all sorting methods. Invoking these methods with no selection or all selection (Ctrl+A) has a different effect.

`getSelectionLinesRange() `might be modified for other Line Operations, to meet BETTER TO HAVE behaviour (the last blank line included). But here we are talking about "Remove Empty Lines".

> Let me just remind once again that this is not some weirdness at all, I saw the same thing in Sublime Text for sorting.
This was probably some prelude to unifying this behavior everywhere, but it was only done for sorting methods.

I believe that Sublime Text is a good editor, however what I'm looking for is not other editor's behaviour, but the feeling of an editor user: One want to remove all the empty lines of his/her document, after select all and execute the command in question, the last blank line is not removed. It's no doubt a bug. Since the current behaviour is correct (I don't see any repport for "Remove Empty Lines" operation's bad behaviour), acceptance of this PR as such will insert a regression.

> It can be selected by column selection, which is exactly in line with the visual presentation:
...
or even:
...
In the above I really see that this last blank line is part of selection.

True. I appreciate this PR for its selection part (persistent selection after operation and column selection).
But if the last 2 bytes of document are CRLF and they're selected with normal selection mode, the last line should be treated by "Remove Empty Lines" operation. It makes sense to me.  

> I've selected line 3 and 4 and because of autoexpand this line will be removed (on this PR). What should be done after reverting behavior, select 3-4 or 3-5 (assuming we keep the autoexpand)?
Same question for this
...
or
...
Visually, I only select 2 lines (3-4), but if I have to restore the previous behavior, what should happen here?

The "autoexpand" you are talking about is the selection after the "Remove Empty Lines" operation, right? If it is, I would say, if the last line which contains CRLF, the select the last 2 bytes CRLF:

image


Otherwise

image


> If someone wants to remove the last blank line it should make selection including last EOL in document himself. The current implementation works according to this approach, even with block selection (at least for these two methods).

I agree. However, in current PR, the last blank line can be removed only no selection, and column selection (which I omitted, sorry).
What it should be is besides above 2 cases, plus Select All & select the 2 last bytes CRLF:
image

Yaron10:
@donho,

> But if the last 2 bytes of document are CRLF and they're selected with normal selection mode, the last line should be treated by "Remove Empty Lines" operation. It makes sense to me.

תמונה

Should line 3 be removed in this case?

donho:
> And in v8.4.7 the user has no way to _exclude_ the last empty line.

It's not true. Select in this way, the last blank line is kept, under v8.4.7

image

Yaron10:
תמונה

Let's try to put aside Ctrl+A.
The user has manually selected the first 2 lines. Why should line 3 be removed in this case? What's the difference?
Why should we care about the content of the line after the selection?

> It's not true. Select in this way, the last blank line is kept, under v8.4.7

You're right.
I had in mind another part of this PR which expands partially selected lines.

donho:
> Why should we care about the content of the line after the selection?

Sorry, don't follow you.


> Let's try to put aside Ctrl+A.
> The user has manually selected the first 2 lines. Why should line 3 be removed in this case? What's the difference?

The cursor cross the 2nd line to the 3rd line:

image

Yaron10:
> Why should we care about the content of the line after the selection?
>> Sorry, don't follow you.

Let's look at the following 2 cases:

תמונה

תמונה

Why should the CRLF in line 3 (in the first case) make us _exclude_ that line, and the lack of CRLF in line 3 (in the second case) make us _include_ that line?

> The cursor crosses the 2nd line to the 3rd line:

But the caret is on the line after the selection _in both case_.
Also - when the status bar implementation changed, the idea/concept was that we don't care about the caret being on the following line.

Yaron10:
@donho,

And as I've mentioned: different opinions are a fundamental part of the human nature.

@ArkadiuszMichalski,

Sorry for the extra work.

ArkadiuszMichalski:
@donho 
>The "autoexpand" you are talking about is the selection after the "Remove Empty Lines" operation, right?

Yes, right after invoking the method, but before deleting anything. We operate on the (whole) lines so with autoexpand selection we don't have to worry about the full (correct) line selecting. 

The current implementation is buggy anyway:
image
We select 5 and 6 line and last CRLF (so also line 7 if we assuming that's the way it should be), but after invoking the second command we get:
image
So two line was removed (not three). We would also need to select CRLF on line 4 to get such behavior:
image
Effect:
image
Still possible, but completely unintuitive.


To try to keep the current behavior (Ctrl+A and weird last CRLF handling) it might look like this:
1.
image
Stretching to something other than the last blank line must also contain `CRLF` because that's how first regex works. When we do it manually, it's also looks like this. So this CRLF is considered as part of the line it appears in. So after all line 2 and 3 will be removed.
2. Let's play with last line:
image
If someone didn't explicitly select the last CRLF, we stretch without it. Lines 5 and 6 will be deleted here.

But if the last CRLF was selected:
image
we also do the stretching of the five line. But in this case lines 5,6 and 7 are removed.

Either way, the questions will continue to arise, what in this situation?
image
I don't know if someone here wants to remove only line 7 or 6-7? Probably 6-7 because for 7 only CRLF should be selected.

What about this?
image
Line 2-3 should be removed or only 3? For me should be 2-3 (this is not the last CRLF and should not be treated differently, it sit on line 2 so that line will be automatically selected).

donho:
@ArkadiuszMichalski 
What's the initial intention of this PR then?

ArkadiuszMichalski:
After all the current code is short and does the job. Sometimes we need to experiment with correct selecting CRLF in lines to achieve the desired result. It makes no sense to complicate this process.

donho:
@ArkadiuszMichalski 
Thank you for your effort !

Issue #12457, opened 2022-11-03
Reduce AppVeyor build time


ArkadiuszMichalski:
In normal mode for one commit we have 10 jobs, each takes 8 minutes on average, which gives over 1h for each commit. It needs some optimization, especially when someone does PR for files that do not require compilation, e.g. XML (from translators). The worst case is when someone doing a PR for one XML file with xxx commits one by one and blocks building for the whole day.

I have some questions and suggestions:
1. Does AppVeyor have any button to cancel the build? What I mean is that a PR author should be able to easily cancel his build if he finds that he will be making more changes anyway.
I see that cancel button might exist (https://pythonpackaging.info/_images/appveyor-fullrun-ongoing.png), but I don't know how to turn it on and what the build cancellation permissions look like.

2. There is something like [Rolling builds](https://www.appveyor.com/docs/build-configuration/#rolling-builds). If it can be set for specific branches, e.g. just for PR, it might be worth doing it. I don't know what it looks like in details because I don't see UI settings for AppVeyor.
Looks like that such an option exists:
![image](https://user-images.githubusercontent.com/2730894/200115328-51bb5b1f-c173-4e34-819d-885fb484bd0d.png)

4. Add some mode for XML type changes where only validation will be performed, not the entire compilation. I'm testing something like this here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12454. 

    If someone does a PR whose title will start as `(xml) Your title` (just put `(xml)` at the beginning) we start only two jobs, and only second (Debug) check syntax for XML file. For people making minor changes to XML files that don't require any compilation, this will suffice (and take a very short time).

    Btw, I don't know how to start build with just one job XML Debug, I'm still getting XML Release and XML Debug. But it's still better than full build.

5. Add to the FAQ description how make simple PR that will not block others. Details will depend on what would be approved from the above suggestions.

rdipardo:
I think all you need is a filter, like this:

~~~yaml
skip_commits:
  files:
    - '**/*.xml'
~~~

Reference: https://www.appveyor.com/docs/how-to/filtering-commits/#commit-files-github-and-bitbucket-only

> If someone does a PR whose title will start as `(xml) Your title`

You can filter by [commit message], too; but then contributors will have to follow the new rule . . .

[commit message]:
https://www.appveyor.com/docs/how-to/filtering-commits/#commit-message

ArkadiuszMichalski:
[...]

PR #12587, opened 2022-11-30
Fix typos in CONTRIBUTING.md

### LINKED ISSUE #12571: Typos in CONTRIBUTING.md
Found a few typos in CONTRIBUTING.md:

image

"Informations" should be changed to "information."

image

I believe it should say this instead: "Use C++ comment line style **rather** than C comment style."

image

For clarity. it might make more sense to say this: "Avoid using pointers. References are preferred instead."

donho:
Fix #12571

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ Bug reports are appreciated. Following a few guidelines listed below will help s
 
 1. Search the issue tracker to see if it has already been reported.
 2. Disable your plugins to see if one of them is the problem. You can do this by renaming your `plugins` folder to something else.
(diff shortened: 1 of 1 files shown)


donho:
@ArkadiuszMichalski 
I thought *.MD are excluded. It seems it's one of cases we have ignored.

ArkadiuszMichalski:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12457#issuecomment-1327316442
>For changes that don't need any build at all (eg .md, .txt,), we can add anywhere in the title [skip ci] or [ci skip] or [skip appveyor]. This is available out of the box and we don't have to do anything (we must remember that such annotation exists and sometimes it's worth using it).

Currently we only filter XML because tests are made for them. I know there exist other files that basically don't require anything. They can be covered with the above method.

I can make an additional filter that will simply return true for such files (if only they are changed). So we would have:
- without any flags 10  jobs will be started and aborted as passed
- with `[xml]` in the title 1 jobs will be started and aborted as passed

Which files are worth doing this way? From what I can see maybe `.md, .txt, .log`.

PR #12593, opened 2022-12-01
Issue 12453: Fix crash opening existing file...

nickdavi:


nickdavi:
No longer crashes on input of \\\\?\ at head of filename. 
IMG_3204

xomx:
@nickdavi 

I quickly checked your PR and found some problems. I put them here instead of making a review in the code.

1) Why you, after detection of a problematic unsupported filename with the \\\\?\ prefix, do not immediately show your warning messagebox and then `return BUFFER_INVALID;`? Even if you know that the current N++ code is not prepared for such filename and fails afterwards... And I do not know if abusing of the existing messagebox later in the `doOpen()` code is a good idea at all.

3) From your picture above is visible another possible problem - you are probably showing an incorrect filename (if you used the same example as in the corresponding issue) - there is no 'dot'-char at its end visible, it should be reported as the "\\\\?\\c:\\file." (this is probably because of you did not immediately show the messagebox and let the following intermediate N++ code cripple this unusual filename...)

3) Because of such raw Win32-filename can have otherwise impossible 'dot' or 'space' char(s) at its end, it would be better to show the filename surrounded by quotes. Otherwise user could overlook e.g. an ending space here "\\\\?\\c:\\file   ".

4) N++ now uses the latest C++ standard, so you can use simpler & safer `if (targetFileName.starts_with(TEXT("\\\\?\\"))` instead of your multiple `targetFileName.at()` (safer, because of if the `targetFileName` contains an empty string, the `targetFileName.at(0)` throws an exception)

nickdavi:
Thank you for the feedback, I will get to work on those changes!

nickdavi:
@xomx It appears that the getFileName function that is called within Notepad_plus.cpp returns the filename ( return _fileNames[index].str(); ) in such a way that the trailing white space is removed before it gets to the doOpen function. I could put in a check for the trailing white space in Notepad_plus.cpp if that would be appropriate. Otherwise, I believe I have all the changes you mentioned ready to go!

PowerEditor/src/NppIO.cpp
```diff
@@ -169,6 +170,13 @@ BufferID Notepad_plus::doOpen(const generic_string& fileName, bool isRecursive,
 	generic_string targetFileName = fileName;
 	resolveLinkFile(targetFileName);
 
+	bool WindowsFileNamingIssue = false;
+
+	if (targetFileName.at(0) == '\\' && targetFileName.at(1) == '\\' && targetFileName.at(2) == '?' && targetFileName.at(3) == '\\')
```
donho, line 175: You detect the specific case but it's not a generic solution.

nickdavi: I have made the alterations suggested by @xomx but have not submitted a new pull request yet as I am not certain how to handle the case of the " " at the end of the document.

Issue #12299, opened 2022-10-07
Strange behavior of the sort feature, with a zero-length column mode selection


guy038:
### Description of the Issue

In some cases, a **rectangular** selection, **followed** with an **ascending sort**, does **not** sort lines of that **selection** correctly !

### Steps to Reproduce the Issue

- - In a **new** tab, insert the *INPUT* text, below :

~~~diff
    one
    two
    three
    four
    five
    six
    seven
    eight
    nine
    ten
~~~

- Move the caret between the four **leading space** chars and the word **`three`**

- Hit **simustaneously** the **`Alt`** and **`Shift`** keys and, at the **same** time, hit **five** times on the **`Down arrow`** key

=> The indication **`Sel: 6x0`** should appear in the **Status** bar

- Now, hit, for instance, the **`=`** sign

=> A **`=`** sign is inserted in front of the lines **`three`** to **`eight`**

- Select the **`Edit > Line Operations > Sort lines Lexicographically Ascending`**

### Expected Behavior

The **alphabetic** sort should work !

### Actual Behavior

Nothing happens :-((

---

- Now, hit the  **`Left arrow`** key, **once**, to cancel the **rectangular** selection

- Hit **simustaneously** the **`Alt`** and **`Shift`** keys and, at the **same** time, hit **five** times on the **`Down arrow`** key

=> Again, a **rectangular** selection ( **`Sel: 6x0`** ), after the **`=`** sign, is created

- Select, again, the **`Edit > Line Operations > Sort lines Lexicographically Ascending`**

=> This time, the **expected** behavior occurs and we get this *OUTPUT* text :

~~~diff
    one
    two
    =eight
    =five
    =four
    =seven
    =six
    =three
    nine
    ten
~~~

As you can see, the lines of the **selection**, with an **`=`** sign, have been **correctly** sorted, in an **ascending** order !

### Debug Information

~~~diff
Notepad++ v8.4.6   (64-bit)
Build time : Sep 25 2022 - 19:51:39
Path : E:\846_x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 21H2
OS Build : 19044.2075
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    PythonScript (2)
    ComparePlus (1)
[...]

PR #12595, opened 2022-12-02
Support thin rectangular by sort commands

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12299.


Yaron10:
Thank you @ArkadiuszMichalski and @alankilborn for the fix.

PR #12600, opened 2022-12-04
Fix inconsistencies at OS-forced Notepad++ (v8.4.7) exit - 2nd part

xomx:
Fixes 2nd part of the  #12541 (when N++ backup mode is ON).

This fixes users reported inconsistency, in between the usual N++ exit initiated manually by a user and the Windows OS forced one, in case that the N++ backup mode is ON.


donho:
Finally the solution is very simple and efficient.
Bravo!

PR #12614, opened 2022-12-08
Update sql.xml (2)

malix0:
### Description of the Issue
SQL functionList does not catch all procedure and function signatures

Open an sql file and check that functionList contains all the signatures

### Solution
I change the sql.xml attached here and this one seems to work better than the one provided with Notepad++. Evaluate yourself if you want to use my version.

I provide the test_re.sql ([test_re.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10181753/test_re.txt) to be renamed) that I used for verification and the screenshot showing different result for default sql.xml and my customized version

FunctionList with my customized version, more function are catched by new RE
sql_custom

FunctionList with default sql.xml provided with the installer
sql_default


donho:
Create a new branch for each PR. Make sure your branch name wasn't used before - you can add date (for example patch3_20200528) to ensure its uniqueness.

Issue #12384, opened 2022-10-23
Delimiter selection crashes program


przemo20:
### Description of the Issue

Setting open and close characters to quotation mark ("), and then double clicking with the left mouse button while holding the Ctrl key on the last empty line in a file crashes the program.

### Steps to Reproduce the Issue

1. Go to: Settings → Preferences... → Delimiter → Switch open and close characters to quotation mark (") and close the Preferences window.
2. Create new document (Ctrl + N), type something and press Enter to create a new line
3. While holding the Left Control key, double click your left mouse button on the empty line
4. Program should now close itself

### Expected Behavior

Program should not select anything (as nothing matches) and not close itself.

### Actual Behavior

Program crashes.

### Debug Information

Notepad++ v8.4.6   (64-bit)
Build time : Sep 25 2022 - 19:51:39
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 22H2
OS Build : 19045.2130
Current ANSI codepage : 1252
Plugins : 
    ComparePlus (1)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

Video: https://www.youtube.com/watch?v=RSd6CAQn8gk

Yaron10:
It seems that the crash occurs if Open and Close contain the same character. _Any character._

xomx:
It crashes at this [line](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/cd6a6ac3bb4d46a9ddd53e3bde2002db00ac00f5/PowerEditor/src/NppNotification.cpp#L746):

`if (bufstring.at(i) == nppGUI._leftmostDelimiter)`

because of the `std::string bufstring` is an empty string ("") there, so calling the `bufstring.at(0)` is nonsense (there is no such `char` at that position to return from an empty string, so it throws an out of range ... "invalid string position" )

Maybe an additional check for empty `bufstring` there should do it (check also two other occurrences of the `bufstring.at(i)` further in the code).

If someone else could make a PR, I will be glad.

prez2307:
[...]

PR #12620, opened 2022-12-09
Fixed delimiter crash issue in NppNotification.cpp

prez2307:
This pull request is meant to fix issue #12384

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -743,6 +743,11 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 					// which are closest to the clicked position.
 					for (int32_t i = static_cast<int32_t>(position_of_click); i >= 0; --i)
 					{
+						if (bufstring.empty()) 
+						{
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppNotification.cpp
```diff
@@ -743,6 +743,11 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 					// which are closest to the clicked position.
 					for (int32_t i = static_cast<int32_t>(position_of_click); i >= 0; --i)
 					{
+						if (bufstring.empty())
```
xomx, line 746: I would rather do such check earlier, right after I have got the `bufstring` content.
E.g. on the line 739: 
```
if (bufstring.empty())
	break;
```

xomx:
And your current solution is not universal - when I check the "Allow on several lines" checkbox in Preferences -> Delimiter, N++ starts crashing again on such empty line.

donho:
The solution of @xomx makes more sense.
I will take care of it if @xomx has no time for creating the PR.

Issue #12602, opened 2022-12-04
"Blank Operations" should work on a selection of lines of text


alankilborn:
### Description of the Issue
Some users think that if they make a selection of lines, then choose _Edit_ > _Blank operations_ > _Trim Leading Space_, for example, that only the lines that are selected should be affected by the operation, not all lines in the currently active document.  Reference [THIS](https://community.notepad-plus-plus.org/topic/23847) Community thread.

### Steps to Reproduce the Issue

1. Create a new tab and paste the following data:
```
4
     Inf 795
     Sup 793
       Izq 1923
        Der 1923
5
         Inf 799
         Sup 800
          Izq 1920
           Der 1920

  6.    
           Inf 800
            Sup 800
              Inz 2010
               Der 2007

7.        
               Inf 797
                Sup  800
                 Izq 2090
                  Der 2093
```
2. Selected lines 1-10:
![image](https://user-images.githubusercontent.com/77065706/205515525-d1805586-c0d9-4d3b-900d-f0bd058cd72d.png)

3. Execute _Edit_ > _Blank operations_ > _Trim Leading Space_

### Expected Behavior
![image](https://user-images.githubusercontent.com/77065706/205515469-c185d247-4ca9-4de9-96ec-4aa4777e7715.png)

### Actual Behavior
![image](https://user-images.githubusercontent.com/77065706/205515499-35988142-a0fa-4f37-a21f-64c4871759bb.png)

### Debug Information
Notepad++ v8.4.7   (64-bit)
Build time : Nov  1 2022 - 23:31:11
Path : C:\\...\npp.8.4.7.portable.x64\notepad++.exe
Command Line : -titleAdd="8.4.7 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.2130
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    PythonScript (3.0.14)

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0c41cc5a5b42c1255d47cc7d29a7da0b02bc44d

PR #12626, opened 2022-12-11
Support selection for Blank Operations

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12602.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1486,8 +1486,10 @@ void Notepad_plus::doTrim(trimOp whichPart)
 	else
 		return;
 	env._str4Replace = TEXT("");
-    env._searchType = FindRegex;
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, true);
+	env._searchType = FindRegex;
+	bool isEntireDoc = _pEditView->execute(SCI_GETSELECTIONSTART) == _pEditView->execute(SCI_GETSELECTIONEND);
+	env._isInSelection = !isEntireDoc;
+	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
 }
 
 void Notepad_plus::removeEmptyLine(bool isBlankContained)


Yaron10:
What about the "built-in" macro `Trim Trailing Spaces and Save`?
Those who are used to saving via that command might want to *trim all* even if there's a selection.

alankilborn:
> What about the "built-in" macro `Trim Trailing Spaces and Save`?
> Those who are used to saving via that command might want to _trim all_ even if there's a selection.

Indeed.
Maybe a new sample macro that "cancels" any active selection(s) before trim-trailing?  Is there a simple way to cancel such selections?

Side note:  Users might be better off to turn control of this (meaning the trim-trailing + save action) over to the editorconfig plugin.

ArkadiuszMichalski:
>Is there a simple way to cancel such selections?

Press right arrow key? 

If the selection bothers someone, they can always create a new macro without it.

alankilborn:
>>Is there a simple way to cancel such selections?

>Press left or right arrow key?

Well, yes, but I was considering a more-automated way, e.g. a Scintilla command.
If one wanted to create a macro (like we are discussing), putting a "press right arrow" command in it, requires also a "press left arrow" to restore original position, and perhaps some side effects from this, I don't know...

--------------------------------

>If the selection bothers someone, they can always create a new macro without it.

Right, but one would have to make sure there is no selection somehow (and maybe that is the "press right" + "press left" before the trim operation).

ArkadiuszMichalski:
> Well, yes, but I was considering a more-automated way, e.g. a Scintilla command.
SCI_CLEARSELECTIONS
https://www.scintilla.org/ScintillaDoc.html#SCI_CLEARSELECTIONS

We can also read the current cursor position in the selection and set it without selection.

alankilborn:
>SCI_CLEARSELECTIONS

>We can also read the current cursor position in the selection and set it without selection.

Please consider what a user can do in a macro in your responses, as that is the point Yaron10 made and I've been responding to.
User can't (easily) do SCI_CLEARSELECTIONS or your other point.

ArkadiuszMichalski:
You asked about Scintilla's commands (although not everyone can have shortcuts).

The easiest way is record macro with press up+down (or reverse) keys before other command.  This causes that whether there was a selection or not, the cursor will remain in the same place.

alankilborn:
>You asked about Scintilla's commands (although not everyone can have shortcuts).

I was asking in the context of what user can do, i.e., what appears in ShortcutMapper.  But I should have been more specific.

"Press left and right" or "press down and up" is okay, but it just feels unclean and hacky.  :-)

But OK, so the suggestion then is:
Add "press left and right" to the _Trim Trailing and Save_ macro, to preserve its existing functionality of trimming from the entire file.

Or maybe that upsets purists?

ArkadiuszMichalski:
You can open a new issue for clearing selections command. Currently there is no such direct command. Even `Esc` is mapped to:
https://www.scintilla.org/ScintillaDoc.html#SCI_CANCEL
but that also leaves the main selection.

alankilborn:
>You can open a new issue for clearing selections command

I don't know that it is worth it.  What felt hacky to me is probably ok with others. :-)

Yaron10:
Another point:
some of the "Blank Operations" still do not work on selection only. This might be confusing.
Looking briefly at the relevant code, the implementation seems to be more complex.

Adding a menu-setting `Apply Commands to Selection Only` (and arrange the menu so that it's clear what those commands are) to "Blank Operations" and "Line Operations" should solve both issues.

But we have all learned quite recently not to engage in futile long discussions. :)

ArkadiuszMichalski:
@Yaron10 

Anyway something needs to be done here because `doTrim()` is also used in `IDM_EDIT_TRIMALL`, and the second part of this command operates on the entire document.

I think that the first 5 commands can operate on selection (it's easy to achieve), but another 3 (after separator) use more complex `wsTabConvert()`.

Yaron10:
@ArkadiuszMichalski,

> Anyway something needs to be done here because doTrim() is also used in IDM_EDIT_TRIMALL...

`EOL to Space` (`IDM_EDIT_EOL2WS`) also operates on the entire document.

> I think that the first 5 commands can operate on selection (it's easy to achieve), but another 3 (after separator)...

That would be better, but still somewhat confusing IMO.

ArkadiuszMichalski:
>EOL to Space (IDM_EDIT_EOL2WS) also operates on the entire document. Yes, but it doesn't use `doTrim()`, just its own method that can be handled separately.

I have a dilemma how to process selections because the default behavior is inconsistent https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12648. We can autoexpand to full lines, but I don't know if that's desirable. The best selection is the one made by the user, but it doesn't work properly.

Yaron10:
@ArkadiuszMichalski,

> Yes, but it doesn't use doTrim(), just its own method that can be handled separately.

If I understand you correctly, the problem is not with `doTrim()` - which apparently can easily be modified to work on selection only - but with `_findReplaceDlg.processAll()`.

> I have a dilemma how to process selections because the default behavior is inconsistent https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12648. We can autoexpand to full lines, but I don't know if that's desirable. The best selection is the one made by the user, but it doesn't work properly.

Unfortunately I don't know anything about RegEx.
Could you give me an example with the first 3 "Blank Operations" that are supposed to work properly at this stage?

If you mean the following case: 
תמונה

`Trim Trailing` removes the selection. - This is correct because the selected chars are indeed "trailing".
`Trim Leading` does not remove the selection. - This is correct because the selected chars are not "leading".

ArkadiuszMichalski:
I open issue for this:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12648

PR #12627, opened 2022-12-12
[xml] Add SQL unit tests

donho:


donho:
@ArkadiuszMichalski 

Adding unitTest with Prefix `[xml]`, I've got the following message at the end of Appveyor message: 
```
...
Build parameters:
Compiler : MSC
Platform : Win32
Configuration : Debug
Changed files on this commit require full build.
```

Can it be optimized?

ArkadiuszMichalski:
@donho 

I initially assumed this was only for XML files. Now I see that there are also files with other extensions (mostly related to testing).
I will add the entire `PowerEditor\Test` folder to the conditions, and in this case will be run only one job `Win32/Debug` with standard compilation and perform all test.

If something is not already covered, you can force the above behavior with the `[xml][force compile]` flags.
https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/FAQ#how-to-create-a-non-blocking-pull-request-pr

PR #12629, opened 2022-12-12
langs: Set ini syntax for some git related files

Biswa96:
This .gitconfig file syntax is as similar as .ini file
as documented here[1]. gitmodules[2] and gitattributes[3]
also follow same syntax. The VSCode ini extension[4]
maintain the similarities.

* https://git-scm.com/docs/git-config#_configuration_file
* https://www.git-scm.com/docs/gitmodules
* https://www.git-scm.com/docs/gitattributes
* https://github.com/microsoft/vscode/blob/main/extensions/ini/package.json


donho:
> See https://github.com/microsoft/vscode/blob/main/extensions/ini/package.json

@Biswa96 
It's not really an argument.
Could you give more arguments for your PR please?

Biswa96:
I have changed the commit message. Thank you for the review.

PR #12636, opened 2022-12-14
[xml][force xml] Switch to Python311-x64

ArkadiuszMichalski:
Looks like Python was update and we have a problem with `lxml` package:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45673917
Or maybe `lxml` was update? We are using a specific version `Python310`. 

I use `--use-pep517` but still build faild:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45674624

On this PR I use prebuild `lxml` provided on the website project (anyway it's marked as unofficial):
https://lxml.de/installation.html
https://www.lfd.uci.edu/~gohlke/pythonlibs/#lxml

Something like this works, but...

@alankilborn @pryrt 
Any tips how to do a proper build via pip? What's missing here?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ image: Visual Studio 2022
 #    - PowerEditor/installer/*/*.xml
 
 environment:
+  python_ver: Python311-x64
   matrix:
     - compiler: MSC
       platform: Win32
@@ -50,7 +51,7 @@ before_build:
                     Write-Output "XML validation mode`n"
                     if ("$env:platform/$env:configuration" -eq "Win32/Debug") {
                         if (@($files_modified | Where-Object {$_ -match $folders_onejob}).length -eq 0) {
-                            $env:Path = "C:\Python311-x64;C:\Python311-x64\Scripts;" + $env:Path
+                            $env:Path = "C:\$env:python_ver;C:\$env:python_ver\Scripts;" + $env:Path
                             python -m pip install requests rfc3987 pywin32 lxml
                             python PowerEditor\Test\xmlValidator\validator_xml.py
                             if ($LastExitCode -eq 0) {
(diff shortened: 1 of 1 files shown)


ArkadiuszMichalski:
Last working log:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45671050
```
Collecting lxml
  Downloading lxml-4.9.1-cp310-cp310-win32.whl (3.3 MB)
     ---------------------------------------- 3.3/3.3 MB 19.0 MB/s eta 0:00:00
```

I don't know where exactly this `.whl` file was located. There are no files for MS on the project page (even for earlier versions):
https://github.com/lxml/lxml/releases

Ohh I see it:
https://pypi.org/project/lxml/4.9.1/#files

ArkadiuszMichalski:
Last commit passed. Any objection for migration to Python311-x64?

ArkadiuszMichalski:
@donho 

It needs to be merged because since yesterday the xml tests will not pass.

donho:
> It needs to be merged because since yesterday the xml tests will not pass.

@ArkadiuszMichalski 
Why and how come "since yesterday"?

ArkadiuszMichalski:
@donho 
Yesterday `lxml` was update from 4.9.1 to 4.9.2, but pypi.org published `.whl` file only for Python-x64, that's why in another PR I had a problem with the XML tests. Today `.whl` was published also for Python-x32, so the problem is gone.

But still might be worth going to Python311-x64 anyway, it seems to have a higher priority than 32-bit. And we use it also in other place like [Plugin List](https://github.com/notepad-plus-plus/nppPluginList/commit/7879bb8810b0c49fdd972720e263a20c2ab1f5fe).

>For Python, do we install on the fly from the official site of Python?

Yes, we use official https://pypi.org/. Finally in this PR I only changed the python version, the rest is unchanged.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/bbe70f55a6fc7814819e35b674df15d999345ed5

PR #12641, opened 2022-12-14
Fixes in PHP autocompletion

vlakoff:
* [`file_get_contents()`](https://www.php.net/manual/en/function.file-get-contents.php):
    * `use_include_path` argument expects a boolean, and this is enforced when strict typing is enabled. See [note on the parameter](https://www.php.net/manual/en/function.file-get-contents.php#refsect1-function.file-get-contents-parameters).
    * `context` argument can be skipped using `null`.
    * `offset` argument default value is `0`. Using `-1` would count from the end of the stream (since PHP 7.1).
    * `length` argument produces a [warning](https://www.php.net/manual/en/function.file-get-contents.php#refsect1-function.file-get-contents-errors) if it is negative. It is nullable but only starting from PHP 8.
* [`file_put_contents()`](https://www.php.net/manual/en/function.file-put-contents.php):
    * `context` argument can be discarded using `null`.
* [`stream_get_contents()`](https://www.php.net/manual/en/function.stream-get-contents.php):
    * returns `false` on failure.
    * [documentation example](https://www.php.net/manual/en/function.stream-get-contents.php#example-3856) confirms the `length` argument can be skipped using `-1`. It is nullable but only starting from PHP 8.
    * `offset` argument default value is `-1`. Documentation states that no seeking occurs with a negative value.

### DIFF
--- a/PowerEditor/installer/APIs/php.xml
+++ b/PowerEditor/installer/APIs/php.xml
@@ -9698,7 +9698,7 @@
 		<KeyWord name="pcntl_signal" func="yes">
 			<Overload retVal="bool">
 				<Param name="int signo"/>
-				<Param name="callback handle"/>
+				<Param name="callback handler"/>
 				<Param name="[bool restart_syscalls]"/>
 			</Overload>
 		</KeyWord>
--- a/PowerEditor/installer/APIs/php.xml
+++ b/PowerEditor/installer/APIs/php.xml
@@ -2804,18 +2804,18 @@
 		<KeyWord name="file_get_contents" func="yes">
 			<Overload retVal="string|false">
 				<Param name="string filename"/>
-				<Param name="[int flags=0"/>
-				<Param name="[resource context"/>
-				<Param name="[int offset=-1"/>
-				<Param name="[int maxlen=-1]]]]"/>
+				<Param name="[bool use_include_path=false"/>
+				<Param name="[resource context=NULL"/>
+				<Param name="[int offset=0"/>
+				<Param name="[int length]]]]"/>
 			</Overload>
 		</KeyWord>
 		<KeyWord name="file_put_contents" func="yes">
 			<Overload retVal="int|false">
-				<Param name="string file"/>
+				<Param name="string filename"/>
 				<Param name="mixed data"/>
 				<Param name="[int flags=0"/>
-				<Param name="[resource context]]"/>
+				<Param name="[resource context=NULL]]"/>
 			</Overload>
 		</KeyWord>
 		<KeyWord name="file_register_wrapper" func="yes">
@@ -12221,10 +12221,10 @@
 			</Overload>
 		</KeyWord>
 		<KeyWord name="stream_get_contents" func="yes">
-			<Overload retVal="string">
-				<Param name="resource handle"/>
-				<Param name="[int maxlength=-1"/>
-				<Param name="[int offset=0]]"/>
+			<Overload retVal="string|false">
+				<Param name="resource stream"/>
+				<Param name="[int length=-1"/>
+				<Param name="[int offset=-1]]"/>
 			</Overload>
 		</KeyWord>
 		<KeyWord name="stream_get_filters" func="yes">


vlakoff:
About [`file_get_contents()`](https://www.php.net/manual/en/function.file-get-contents.php), its `length` parameter produces a warning with negative values, but I just discovered it accepts `-1` as a special case, for backwards compatibility. See https://github.com/php/php-src/commit/62dce97973436f1830b18304e7939a03b18d44ba. Maybe this special case will be removed sometime in the future.

What should we do?
* Keep `-1` as the default value (current situation).
* Put no default value (my initial choice in this PR). So it means the argument could be avoided only by omitting it.
* Put `null` as default value, but it is supported only starting from PHP 8.

I'm reverting it to `-1`, for user convenience. Then in some years, we could change it to `null`.

vlakoff:
My mistake, https://github.com/php/php-src/commit/62dce97973436f1830b18304e7939a03b18d44ba is about [`stream_get_contents()`](https://www.php.net/manual/en/function.stream-get-contents.php). For this function, they forbade negative values, while still allowing `-1` as a special case.

It seems that [`file_get_contents()`](https://www.php.net/manual/en/function.file-get-contents.php) is forbidding all negative values (including `-1`), and always has been. To get all contents, the default is a `PHP_STREAM_COPY_ALL` internal value. I'm looking into this.

vlakoff:
So, the internal [`PHP_STREAM_COPY_ALL`](https://github.com/php/php-src/blob/77ee92a50cee4ea6f9ae1994ea7950b110e4496e/main/php_streams.h#L465) is defined to `((size_t)-1)`, which is equal to `2147483647` (the max 32-bit signed integer), even on 64-bit platform.

Now, back to [`file_get_contents()`](https://www.php.net/manual/en/function.file-get-contents.php):
* Using a value larger than `2147483647` generates a warning: « Warning: file_get_contents(): maxlen truncated from [...] to 2147483647 ».
* **But**, specifying `2147483647` (or even less) generates an **error**: « Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 2147483672 bytes) ».
* On PHP < 8, passing `null` doesn't generate a warning… but it reads zero bytes from the file.

On a related note, an available constant is [`PHP_INT_MAX`](https://www.php.net/manual/en/reserved.constants.php#constant.php-int-max), which varies between 32-bit and 64-bit platforms, but there is no PHP constant for `2147483647`.

So, I guess our only option is to go back to this one:
> * Put no default value (my initial choice in this PR). So it means the argument could be avoided only by omitting it.

Refs also this message: [Hi again! . . .and file_get_contents() maxlen default value](https://externals.io/message/45493).

vlakoff:
A few other refs, somehow related to that `length` argument:
* https://github.com/php/php-src/commit/55fd7ac6013b39601b3242d62fe0e0244da2d063
* https://github.com/php/php-src/commit/34e686c5562bfd289d330f5dae2dafec4e58a20c
* https://github.com/php/php-src/commit/7e9479083d6dc6d84c393a96eb718bccc0185b9b ([PR #6648](https://github.com/php/php-src/pull/6648))

Anyway, as I stated above, I updated the PR to go back to the "no default value" option. The PR should now be ready.

PowerEditor/installer/APIs/php.xml
```diff
@@ -12221,10 +12221,10 @@
 			</Overload>
 		</KeyWord>
 		<KeyWord name="stream_get_contents" func="yes">
-			<Overload retVal="string">
-				<Param name="resource handle"/>
-				<Param name="[int maxlength=-1"/>
-				<Param name="[int offset=0]]"/>
+			<Overload retVal="string|false">
+				<Param name="resource stream"/>
+				<Param name="[int length=-1"/>
```
donho, line 12226: rather `int $length = null` ?
ref: https://www.php.net/manual/en/function.stream-get-contents.php

donho, line 12226: rather `int $length = null` ?
ref: https://www.php.net/manual/en/function.stream-get-contents.php

vlakoff: this review seems to be a duplicate of the other one

donho: @vlakoff 
Thank you for your reply.

donho: @vlakoff 
Thank you for your reply.

PR #12650, opened 2022-12-15
Add Go to settings for Change History margin style

ArkadiuszMichalski:
This one is missing:

image


PowerEditor/src/WinControls/ColourPicker/WordStyleDlg.cpp
```diff
@@ -943,6 +943,11 @@ std::pair<intptr_t, intptr_t> WordStyleDlg::goToPreferencesSettings()
 		result.first = 3;
 		result.second = IDC_CHECK_BOOKMARKMARGE;
 	}
+	else if (style._styleDesc == TEXT("Change History margin"))
+	{
+		result.first = 3;
+		result.second = IDC_CHECK_CHANGHISTORYMARGE;
+	}
```
donho, line 950: Ha! It's really good such system! :)

Issue #12648, opened 2022-12-15
Regex for selection is not matched at the start of the selection


ArkadiuszMichalski:
Why for selection regex have the following behavior?
1. Regex like `^patter` is always matched to the beginning of the line, not to the beginning of the selection. So below example for `Replace All` will not replace anything:
![image](https://user-images.githubusercontent.com/2730894/207901256-66acecb2-e429-4cc2-bdf1-d4cdc3bd8f26.png)

2. Regex like `patter$` is matched to the ending of the selection. So below example for `Replace All` will replace `ABCD`:
![image](https://user-images.githubusercontent.com/2730894/207901677-a049bc8b-d079-4c00-b2c2-e23f8f28f6e4.png)

This causes some methods that use `_findReplaceDlg.processAll()` also have the above behavior, which is a bit confusing (because sometimes they only change the last line in a partial selection and the first line might be unchanged).

alankilborn:
I can "sort of" understand the behavior of the above.  :-P

I presume this is the problem area?:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bf34ef0dd568b792627813a4445adf649a5c5ba6/PowerEditor/src/Notepad_plus.cpp#L1474-L1493

alankilborn:
So you could make your case 2 behave more like case 1 if you searched on `ABCD(?=\R)` or, because you and some others like last-lines without ends on them, `ABCD(?=\R|\z)`...but that second one behaves like your case 2. :-(

But I'm guessing (without thinking too hard about the exact thing you are trying to solve) that you really want case 1 to behave like case 2.

ArkadiuszMichalski:
>I presume this is the problem area?:

Not only here, but here is the easiest way to show it. 

For such example with selection:
![image](https://user-images.githubusercontent.com/2730894/207974161-b0967638-f444-4b85-a336-c436256a05f3.png)
Trim Trailing works
![image](https://user-images.githubusercontent.com/2730894/207974372-7a1c082c-3130-416d-84b6-b065ad5de7e6.png)
but Trim Leading not 
![image](https://user-images.githubusercontent.com/2730894/207974423-6c625b39-ff96-4818-8cb7-bf486129f795.png)

>But I'm guessing (without thinking too hard about the exact thing you are trying to solve) that you really want case 1 to behave like case 2.

Well, if we do not extend the selection to full lines, it would probably be the most appropriate.

Yaron10:
[...]

PR #12655, opened 2022-12-16
[xml][force compile] Fix trim operations for selection

ArkadiuszMichalski:
… line

Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12658

@alankilborn Can you check if it's consistent now?

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1477,17 +1477,52 @@ void Notepad_plus::doTrim(trimOp whichPart)
 	FindOption env;
 	if (whichPart == lineHeader)
 	{
-		env._str2Search = TEXT("^[	 ]+");
+		env._str2Search = TEXT("^[\\t ]+");
 	}
 	else if (whichPart == lineTail)
 	{
-		env._str2Search = TEXT("[	 ]+$");
+		env._str2Search = TEXT("[\\t ]+$");
+	}
+	else if (whichPart == lineFull)
+	{
+		env._str2Search = TEXT("^[\\t ]+|[\\t ]+$");
 	}
 	else
 		return;
 	env._str4Replace = TEXT("");
-    env._searchType = FindRegex;
-	_findReplaceDlg.processAll(ProcessReplaceAll, &env, true);
+	env._searchType = FindRegex;
+	bool isEntireDoc = _pEditView->execute(SCI_GETSELECTIONSTART) == _pEditView->execute(SCI_GETSELECTIONEND);
+	env._isInSelection = !isEntireDoc;
+	_findReplaceDlg.processAll(ProcessReplaceAll, &env, isEntireDoc);
+
+	// fix first partially selected line
+	if (!isEntireDoc && (whichPart == lineHeader || whichPart == lineFull))
+	{
+		auto startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
+		auto startLine = _pEditView->execute(SCI_LINEFROMPOSITION, startPos);
+		auto endPos = _pEditView->execute(SCI_GETSELECTIONEND);
+		auto endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
+
+		if (startPos == endPos || startPos == _pEditView->execute(SCI_POSITIONFROMLINE, startLine))
+			return;
+
+		if (startLine != endLine)
+			endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, startLine);
+
+		char *text = new char[endPos - startPos + 1];
+		int counter = 0;
+		_pEditView->execute(SCI_SETTARGETRANGE, startPos, endPos);
+		_pEditView->execute(SCI_GETTARGETTEXT, 0, reinterpret_cast<LPARAM>(text));
+
+		while (text[counter] == ' ' || text[counter] == '\t')
+			counter += 1;
+
+		if (counter)
+			_pEditView->replaceTarget(TEXT(""), startPos, startPos + counter);
+
+		delete [] text;
+	}
+
 }
 
 void Notepad_plus::removeEmptyLine(bool isBlankContained)
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1473,7 +1473,7 @@ void Notepad_plus::wsTabConvert(spaceTab whichWay)
 
 void Notepad_plus::doTrim(trimOp whichPart)
 {
-	// whichPart : line head or line tail
+	// whichPart : line head or line tail or line both
 	FindOption env;
 	if (whichPart == lineHeader)
 	{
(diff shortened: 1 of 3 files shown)


alankilborn:
I don't think 12648 is the correct issue for this PR?

ArkadiuszMichalski:
There is a discussion with descriptions of this problem, but I will move it to a new issue later.

alankilborn:
>>Can you check if it's consistent now?

>There is a discussion with descriptions of this problem for trim operations , but I will move it to a new issue later.

Hmm.  Without an issue, I don't know what it is that I'm checking...

ArkadiuszMichalski:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12648#issuecomment-1353755867

alankilborn:
As a user, I would not use "trim" operations on partially-selected lines -- I would select full lines with the mouse using line number margin click+drag.  Or, if I did partially select some group of lines, I would expect "trim" to operate on the full lines touched by selection.  I don't really want to get into the possibly endless debate about how such a thing should work (on partially selected lines), so perhaps @Yaron10 is a better and more thorough tester of this?

ArkadiuszMichalski:
>As a user, I would not use "trim" operations on partially-selected lines -- I would select full lines with the mouse using line number margin click+drag

So you still have a choice, and if that's your workflow, you can always select all. If we do automatic stretching on whole lines, there may be voices like this:
image
Why for above example line 1 and 4 are always include, I didn't mark them at all. Such automatic things done behind your back are better to do only as a last resort (e.g. when sorting, because taking only part of the first/last line is pointless here).

alankilborn:
Right, I was just trying to make the point that I am not the best person to give you the testing on this.

Yaron10:
> As a user, I would not use "trim" operations on partially-selected lines -- I would select full lines with the mouse using line number margin click+drag. Or, if I did partially select some group of lines, I would expect "trim" to operate on the full lines touched by selection.

👍 

> so perhaps @Yaron10 is a better and more thorough tester of this?

I think that removing the selection in the following case is wrong.
תמונה

Leading/Trailing tabs and non-Leading/Trailing tabs are two different "entities".
Let's imagine that Leading/Trailing tabs were represented by a different shape.
For example: ».

Selecting a non-Leading/Trailing tab
תמונה

does not transform it to:
תמונה

***
> Why for above example line 1 and 4 are always include,

Line 4 should not be included.

donho:
@ArkadiuszMichalski 

It's about "Trim **Trailling**" or "Trim **Leading**" operation so it's rather ones of Line Operations.
the selection on left should be considered extended as right:

image

Sorry, for https://github.com/notepad-plus-plus/notepad-plus-plus/commit/bf34ef0dd568b792627813a4445adf649a5c5ba6 I have checked only selection full lines.

ArkadiuszMichalski:
@donho 

It's not that simple, because then appear other borderline cases.

1
image
Cursor is at the beginning on line 4, so the extension should also include line 4, like this or not?

image

2 What in this situation, line 2 should be selected? We start selection before `CRLF` on line 2.
image

3 We start selection before `CRLF` on line 2, and finish with cursor at the beginning on line 3.
image

4 Here I assume that the whole line 4 should be selected (because the first character `A` was selected)?
image

What should be done with the last `CRLF`? Should be a part of this auto-expand selection or not? From the perspective of `doTrim()` it doesn't matter, but this function is also used with `IDM_EDIT_TRIMALL` (as first part), and for this command (named `Remove Unnecessary Blank and EOL`), every `CRLF` is important. In this case the selection will be processed with `doTrim()` first, so it matters more what exactly are we doing with selection here.

I've already seen what problems can arise if we try implement auto-expand selection in: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12580.

That's why I came to the conclusion that it's best if the commands operate on the selection exactly made by us. Then we always have the possibility to select exactly what we need. With auto-expand selection we are forced into default behavior, which may not always meet our expectations.
Maybe a solution would be to make, for example, a separate command `Expand selection to full lines` available from the macro level, and if someone needed it, they could combine the appropriate commands (through macro).

@donho 
But if you want auto-expand selection here then we can try, just write what should happen in the above cases.

donho:
@ArkadiuszMichalski 
No auto-expand selection please. We don't need visual update.
We need only the determination of which lines should apply to trim operation, from the selection. 
For the determination, you can see "Move up/down current line" commands' behaviour with the random selection.

ArkadiuszMichalski:
@donho 

Can you check the last commit is it better?  Now it follows what commands "Move up/down current line" do.

1
image
Selects lines 2 and 3.

2
image
Selects lines 2 and 3.

3
image
Selects lines 2.

4
image
Selects lines 2 and 3 and 4.


We never expand selection to contain the last `CRLF` because another command will not work properly (`IDM_EDIT_TRIMALL`). So:
image
would be:
image
as you suggested earlier https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12655#issuecomment-1356365695.

donho:
@ArkadiuszMichalski 
When multi-selections are present, it takes only the last selection - but I think it's tolerable.

`[xml][force compile]` - it might be a good practice to not consume too much resource for the earlier stage of PR and get quickly the result, but please don't forget to remove it in the latest commit in order to validate the whole set of compiling.

ArkadiuszMichalski:
@donho 
>When multi-selections are present, it takes only the last selection - but I think it's tolerable.

Yes, but that is due to the limitation of `_findReplaceDlg.processAll`. Wherever we use this function only the last selection will be processed. At the moment it may look like this, or we may not support multiselect at all (just return).
Multi-selection seems to be difficult to support because changes in one selection affect the next ones, I don't even know if Scintilla somehow properly saves the selected data in such a situation.

>[xml][force compile] - it might be a good practice to not consume too much resource for the earlier stage of PR and get quickly the result, but please don't forget to remove it in the latest commit in order to validate the whole set of compiling.

I know, even mention something like that here:
https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/FAQ#other-available-flags
I was just waiting for you to come back to this PR and check out this new approach, because I don't know if you like it in this form.

image
Now after committing, I will try to adjust these other methods so that they also work on the selection.

alankilborn:
>Multi-selection seems to be difficult to support because changes in one selection affect the next ones, I don't even know if Scintilla somehow properly saves the selected data in such a situation.

What you do for this situation is, for the first change, note how the document length changes, and adjust the location of the intended second change by this amount.  And so on for all the subsequent changes needed.

donho:
As I mentioned, it's tolerable for me. So let's take the simplest solution (the current one).

ArkadiuszMichalski:
>What you do for this situation is, for the first change, note how the document length changes, and adjust the location of the intended second change by this amount. And so on for all the subsequent changes needed.

Right, we can combine in this direction. But I would still expect that `_findReplaceDlg.processAll` support this firstly. It would be less work than doing everything ourselves every time we need this.

PR #12656, opened 2022-12-16
Fixed the lua syntax highlighter once and for all. (Redone)

ccuser44:
This is a redone version of #12367

### DIFF
--- a/PowerEditor/installer/APIs/lua.xml
+++ b/PowerEditor/installer/APIs/lua.xml
@@ -93,8 +93,7 @@ message.
         <KeyWord name="coroutine.running" func="yes">
             <Overload retVal="void" descr="
 Returns the running coroutine, or nil when called by the main thread. 
-        ">
-            </Overload>
+        "></Overload>
         </KeyWord>
         <KeyWord name="coroutine.status" func="yes">
             <Overload retVal="void" descr="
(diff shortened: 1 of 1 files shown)


ccuser44:
Ok I have re-done #12367 without the accidental formatting changes

ccuser44:
I'm not sure if I missed some newer Lua globals though because I added all of these manually.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12367#issuecomment-1314492403

PowerEditor/installer/APIs/lua.xml
```diff
@@ -1220,24 +1237,24 @@ captures from the pattern; otherwise it returns nil. If pattern specifies no cap
 the whole match is returned. A third, optional numerical argument init specifies where to 
 start the search; its default value is 1 and can be negative. 
         ">
-                <Param name="s" />
-                <Param name="pattern" />
-                <Param name="[, init]" />
+                <Param name="s"/>
+                <Param name="pattern"/>
+                <Param name="[, init]"/>
```
donho, line 1242: It's not necessary to remove all the white space.
Reformatting is not allowed, as I mentioned:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12367#issuecomment-1314492403

ccuser44: So do I need to fix the small whitespace additions as well?

PR #12657, opened 2022-12-16
Fixed the lua syntax highlighter once and for all. (Redone) (Redone)

ccuser44:
This is a redone version of #12656 which is a redone version of #12367
The XML formatter automatically reformatted it so I had to manually add it back. But now it is done. Phew, took a long time.

### DIFF
--- a/PowerEditor/installer/APIs/lua.xml
+++ b/PowerEditor/installer/APIs/lua.xml
@@ -2,7 +2,40 @@
 <NotepadPlus>
     <AutoComplete language="LUA">
         <Environment ignoreCase="yes" startFunc="(" stopFunc=")" paramSeparator="," terminal=";" additionalWordChar=".:" />
+
+        <!-- Lua syntax-->
+        <KeyWord name="break" func="no" />
         <KeyWord name="and" func="no" />
+        <KeyWord name="do" func="no" />
+        <KeyWord name="else" func="no" />
+        <KeyWord name="elseif" func="no" />
+        <KeyWord name="end" func="no" />
+        <KeyWord name="false" func="no" />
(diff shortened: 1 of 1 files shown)


donho:
@ccuser44 
The file you have modified has nothing to do with syntax highlighting.
Have you tested your modification to see the difference?
Could you provide screenshot BEFORE and AFTER your PR to see the improvement?

ccuser44:
> The file you have modified has nothing to do with syntax highlighting.

Well I don't know exactly what to call it. It's the thing that suggest functions&varables. And actually the syntax highligter does use this file to check if a function/variable is a Lua global.

But yes it removes unnecessary junk from the variables&function list from the thing that suggests them.

ccuser44:
> @ccuser44 The file you have modified has nothing to do with syntax highlighting. Have you tested your modification to see the difference? Could you provide screenshot BEFORE and AFTER your PR to see the improvement?

This removes:

- [BUG] Dublicate keywords (These dublicate keywords sometimes cause odd glitches. They also make some features of the Lua (Lua syntax highlighter&variable suggestor) not work)

  image

  image

- [BUG] Removes unnecessary keywords (These make developing Lua with NPP a pain. When you are typing code NPP suggest nonsensical keywords to you. And when you wan't to find valid keywords from a list of 500 junk keywords it makes developing harder. Many users often complain about this (who develop Lua with NPP))

  image

This adds:

- Keywords from newer Lua versions (Makes developing newer versions of Lua easier with NPP)

Mateos81:
> Well I don't know exactly what to call it. It's the thing that suggest functions&varables
It's the auto-completion :)

Issue #12583, opened 2022-11-29
`Settings on cloud` does not redirect `plugins` and `themes` folders


o-l-a-v:
### Description of the Issue

Setting the `Settings on cloud` -> `Set your cloud location here` to some other place than the default path of `%APPDATA%\Notepad++` works well for the `userDefineLangs` and `backup` folders, but not `plugins` and `themes` folders, which still gets read/ written to `%APPDATA%\Notepad++\<foldername>`.

### Steps to Reproduce the Issue

See previous point.

### Expected Behavior

All folders should be read and written to the set location, `%APPDATA%\Notepad++%` should not be used anymore for `themes` and `plugins` after setting `Settings on cloud`.

### Actual Behavior

`backup` and `userDefineLangs` works as expected, `plugins` and `themes` still reads and/ or writes from `%APPDATA%\Notepad++%`.

### Debug Information

```
Notepad++ v8.4.6   (64-bit)
Build time : Sep 25 2022 - 19:51:39
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : C:\Users\<removed>\OneDrive\_Share\AppData\Notepad++
OS Name : Windows 11 (64-bit) 
OS Version : 22H2
OS Build : 22621.819
Current ANSI codepage : 1252
Plugins : 
    JSMinNPP (1.2205)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    NppMarkdownPanel (0.6.2)
    XMLTools (3.1.1.13)
```

Issue #12296, opened 2022-10-06
Theme bug in portable version


qx87:
I already posted the issue #1344 but it was closed for no good reason.

There was said:
> if your issue still persists, and it is still of relevance to you, please feel free to file a new issue as detailed and reproducible as possible.

Yes, the issue still persists, and yes, it is still of relevance to me, nothing changed.

And people did write about it: #2579, #1589,  #9409.

clonsberry:
For others who are searching: I wrote [workaround instructions](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9409#issuecomment-1298496463) in 9409.

Issue #10801, opened 2021-11-19
Partial inoperability of dark mode on the first launch of the application and cloud synchronization of settings


Korb:
### Steps to Reproduce the Issue
1. Enable in-app dark theme and cloud sync via Google Drive
2. Turn on the PC or restart Windows
2. Launch the application with a shortcut from the Windows Quick Launch bar

### Expected Behavior
After installing a dark theme in the application settings - the application always starts with a dark theme.

### Actual Behavior
The dark theme does not apply to all elements of the application. To fix it, I have to open "Preferences" / "General" / "Dark Mode", uncheck "Enable dark mode" (dark mode will be disabled) and set it back (dark mode will be enabled including the main text field of the application).

### Debug Information
Notepad++ v8.1.9.1   (64-bit)
Build time: Nov 13 2021 - 15:28:20
Path: C:\SW\NotepadPP\notepad++.exe
Command Line: 
Admin mode: OFF
Local Conf mode: ON
Cloud Config: J:\Мой диск\Software\Microsoft Windows\Notepad++\Settings
OS Name: Windows 10 Pro (64-bit) 
OS Version: 2009
OS Build: 19044.1379
Current ANSI codepage: 65001
Plugins: DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll ShtirlitzNppPlugin.dll urlPlugin.dll

This is how the application looks after the first launch:
![2021-11-19_11-07-44](https://user-images.githubusercontent.com/3509797/142588991-13ec8954-e8ad-43d9-96cf-8c917e61c08a.png)

This is how the app looks after turning the dark theme off and on again:
![2021-11-19_11-10-24](https://user-images.githubusercontent.com/3509797/142589020-4d9be20d-676b-49cd-8391-0c1bacc090ef.png)

"Settings on cloud" / "Set your cloud location path here:"
![2021-11-19_11-08-58](https://user-images.githubusercontent.com/3509797/142589084-a3430013-6b56-4d20-82dc-070a082de80b.png)

"Global Styles: Global override" screen:
![2021-11-19_11-09-56](https://user-images.githubusercontent.com/3509797/142589086-da987079-cac5-4d97-8644-41aa120def32.png)

"Global Styles: Default Style" screen:
![2021-11-19_11-10-03](https://user-images.githubusercontent.com/3509797/142589087-c7972cab-8802-4382-8d13-1d47246d3616.png)

[...]

Issue #9409, opened 2021-01-18
Some settings still not portable


Tragen:
### Description of the Issue
There are at least two settings which are not "portable".
One is stylerTheme, already reported in #1589 and #2579 and #1344 but closed because
nobody reported the error for newer NP++ versions.

Another "setting" is in session.xml
If you automatically reopen unsaved text files, they get saved in NP++ folder\backup.
But the path isn't saved relative and so the files aren't reopened when the path changes, e.g. you work on another computer and the drive letter changes or when I update NP++ and have the version number in the path.

### Expected Behavior
Paths are saved relative to the NP++ folder.

### Actual Behavior
Paths are saved as absolute paths.

This happens in 7.9.2 also and every version before.

chcg:
Regarding the session.xml complain this is probably about entries like:

  `          <File firstVisibleLine="0" xOffset="0" scrollWidth="128" startPos="26" endPos="0" selMode="0" offset="0" wrapCount="1" lang="None (Normal Text)" encoding="-1" userReadOnly="no" filename="new 1" backupFilePath="C:\Users\XY\npp.7.9.2.portable.x64\backup\new 1@2021-01-06_213008" originalFileLastModifTimestamp="0" originalFileLastModifTimestampHigh="0" mapFirstVisibleDisplayLine="-1" mapFirstVisibleDocLine="-1" mapLastVisibleDocLine="-1" mapNbLine="-1" mapHigherPos="-1" mapWidth="-1" mapHeight="-1" mapKByteInDoc="0" mapWrapIndentMode="-1" mapIsWrap="no" />
`

clonsberry:
# How to use relative path for style config in the portable version
For others struggling to figure out the [workaround](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1344#issuecomment-171168759) mentioned in 1344, **make sure n++ is closed when you update the theme path** in your config.xml file.

## Works
1. Ensure n++ is closed. 
2. Open config.xml using your next favorite editor after n++ _(I was a bit slow to figure out that I needed to use another editor to change my config.xml file. 😊)_
3. Edit the `"stylerTheme"` key to use a relative path. Example: `path=".\themes\Vibrant Ink.xml"`
4. Save
5. When next opening n++, the relative path persists

### Does not work
**n++ seems to cache the config it had upon opening and rewrite the file on closing**
1. Open n++
2. Open config.xml
[...]

PR #12662, opened 2022-12-17
Fix issue with custom theme when changing dark/light mode

ozone10:
fix issue from https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12466

@donho 
When changing dark/light mode, only default themes (in "themes" folder in folder with notepad++.exe) will be used.

This PR will fix it.

Allows cloud setting to redirect themes folder, currently it uses `%AppData%\Notepad++\Themes\` or `Themes\` with `doLocalConf.xml` file.

Also use relative path for themes with `doLocalConf.xml` file.

should fix or improve issues below and maybe more issues with portable config and themes
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6092
partially fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9409
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10801
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12296
partially fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12583

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -7932,8 +7932,18 @@ void Notepad_plus::refreshDarkMode(bool resetStyle)
 		generic_string xmlFileName = NppDarkMode::getThemeName();
 		if (!xmlFileName.empty())
 		{
-			themePath = themeSwitcher.getThemeDirPath();
-			pathAppend(themePath, xmlFileName);
+			if (!nppParams.isLocal() || nppParams.isCloud())
+			{
+				themePath = nppParams.getUserPath();
+				pathAppend(themePath, TEXT("themes\\"));
+				pathAppend(themePath, xmlFileName);
+			}
+
+			if (::PathFileExists(themePath.c_str()) == FALSE || themePath.empty())
+			{
+				themePath = themeSwitcher.getThemeDirPath();
+				pathAppend(themePath, xmlFileName);
+			}
 
 			themeName = themeSwitcher.getThemeFromXmlFileName(themePath.c_str());
 		}
@@ -7946,7 +7956,7 @@ void Notepad_plus::refreshDarkMode(bool resetStyle)
 			themeName = themeSwitcher.getDefaultThemeLabel();
 		}
 
-		if (::PathFileExists(themePath.c_str()))
+		if (::PathFileExists(themePath.c_str()) == TRUE)
 		{
 			nppParams.getNppGUI()._themeName = themePath;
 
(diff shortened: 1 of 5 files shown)


ozone10:
STR
1. Use installed Notepad++
2. Remove default themes in `%AppData%\Notepad++\Themes`
3. Add theme with different name than ones in Themes\ subdirectory of your Notepad++ installation directory to `%AppData%\Notepad++\Themes` than ones in theme
4. Set that theme in either dark/light mode, only in one mode, and in other mode default theme (e.g. Obsidian)
5. change between dark/light modes

xomx:
@ozone10 
Did not test the cloud stuff, but all the local stuff works as announced.

donho:
@ozone10 
> STR
> 
>     1. Use installed Notepad++
> 
>     2. Remove default themes in `%AppData%\Notepad++\Themes`
> 
>     3. Add theme with different name than ones in Themes\ subdirectory of your Notepad++ installation directory to `%AppData%\Notepad++\Themes` than ones in theme
> 
>     4. Set that theme in either dark/light mode, only in one mode, and in other mode default theme (e.g. Obsidian)
> 
>     5. change between dark/light modes

I can reproduce the bug only when `%appdata%\themes\` is empty and config.xml is removed (Then launch Notepad++ like the first launch after Notepad++ installation).

donho:
> I can reproduce the bug only when `%appdata%\themes\` is empty and config.xml is removed (Then launch Notepad++ like the first launch after Notepad++ installation).

@ozone10 And the PR doesn't fix the scenario with which I reproduced the bug :(

donho:
> Did not test the cloud stuff, but all the local stuff works as announced.

@xomx Could you share your way to reproduce the bug?

ozone10:
@donho 
`%AppData%\Notepad++\themes\` contains only 1 theme `VS2019-Dark.xml`
`config.xml` does not exist
Notepad++ just opened with preference and styler configurator windows open, no settings changed at start of gif

str

donho:
@ozone10 

Step 3 is not clear to me:

> Add theme with different name than ones in Themes\ subdirectory of your Notepad++ installation directory to %AppData%\Notepad++\Themes than ones in theme

I guess with the scenario you provided, it's with the existence of config.xml.

ozone10:
@donho 
I mean themes not provided by Notepad++, not with same name as [here](https://github.com/notepad-plus-plus/notepad-plus-plus/tree/master/PowerEditor/installer/themes)

donho:
@ozone10 
I can reproduce now, and both bugs are fixed.
Thank you!

PR #12550, opened 2022-11-22
sharpen autocomplete icons,

vinsworldcom:
fix #12549 

image
 
and still looks fine on white backgrounds:

image

PR #12667, opened 2022-12-18
Use dark or light autocomplete icons based on theme bg

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12550#issuecomment-1323238165

image


xomx:
Perfect, works ok!
My tired eyes say - thanks :-)

PR #12671, opened 2022-12-19
Make theme portable

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6518f3e4b809bc9f56004795d38b88016be5088e#commitcomment-93601484


ozone10:
@vinsworldcom
I was quite naive thinking users would open in portable mode Notepad++.exe directly.

This PR should fix it.

donho:
@vinsworldcom 
Thank you for testing and reporting the bug.

@ozone10 
Thank you for the quick fixing.

AndiKiissel:
RC3 portable is still not portable:

image

image

donho:
@AndiKiissel 
> RC3 portable is still not portable:

It seems this settings is no more used (or can be replaced).
An issue has been created (#12685) and I will see what can I do about it.

Issue #12588, opened 2022-11-30
In dark mode search results window is with white background.


zilvinasklp:
### Description of the Issue
In Notepad++ dark mode search results are with white background

### Steps to Reproduce the Issue

1. Install Notepad++. I have 8.4.7.
2. Switch to dark mode (Settings > Preferences > Dark mode > Enable dark mode (checkbox) > Close).
3. Call search box (CTRL+F/CMD+F), select "Find in Files" and look for any phrase  in selected folder.

### Expected Behavior
Expected to see results frame/window also in dark mode / with dark background.

### Actual Behavior
After search is finished results are shown in separate search results frame/window at the bottom in white background with black font size.
Screenshot (blurred text):  https://i.imgur.com/1T6k8n3.png

ArkadiuszMichalski:
@ozone10
Looks like that in some case switching dark mode works correct for `Search results` when we firstly open `Search results`, and after this switch dark mode.

STR:
1. Open NPP when Light mode is default.
2. Call search box (CTRL+F) and find something by `Find All in Current Document`.
3. Search results appear, close it.
4. Switch to dark mode.
5. Repeat step 2.

After this our Search results is still on Light mode.

STR to fix this:
1. Call search box (CTRL+F) and find something by `Find All in Current Document`.
2. Search results appear (don't close it).
3. Now switch to dark/light mode depending what mode you want.

ozone10:
@ArkadiuszMichalski 
it does not depend on dark mode, it depends on theme. You can replace step 4 with changing theme.
It seems it's started from version 8.4.3, unfortunately I don't have time to investigate it.

ArkadiuszMichalski:
Confirmed, it's started from 8.4.3.

zilvinasklp:
Also if dark mode is disabled typing select suggestions still remains in dark background. https://prnt.sc/rOtltDhA8Oye

ozone10:
This is caused by commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0b5785cbecf82e5889f6e8680aeed85305b5fe88

Now that `_pFinder->setFinderStyle();` is moved to `if (!_pFinder)` block [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/58416114da14536dcc7a8ad36010da3382366ba6/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L3003), style won't be updated if finder is closed.

[...]

PR #12672, opened 2022-12-19
Fix search result not applying new theme

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12588


donho:
@ozone10 
I cannot reproduce the bug https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12588 at all with v8.4.8 RC2.
Could you provide the instruction to reproduce it?

ozone10:
@donho 
finder

STR:
1. Open some file, e.g. `stylers.model.xml`
2. Search something ( `Color` ) with Find All in Current Document
3. Change theme, or change color for `Hit Word` style in `Search result` language

PR #12682, opened 2022-12-21
Add two missing xml files to makefile

ArkadiuszMichalski:
Two xml files `tabContextMenu_example.xml` and `nppLogNulContentCorruptionIssue.xml` are added to the official package but are missing in the `makefile`.

@donho 
[`toolbarIcons.xml`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/toolbarIcons.xml) cannot be added to the official pack? Its default content will not affect the program, but it shows that Notepad++ offers the ability to change icons.

Edit: hmm `tabContextMenu_example.xml` is added only to `minimalist` ([packageAll.bat](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/packageAll.bat)), why not also to full?

### DIFF
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -26,8 +26,8 @@ BIN_DIRECTORY := ../bin
 INSTALLER_DIRECTORY := ../installer
 
 TARGET_BINARY := notepad++.exe
-SRC_DATA := contextMenu.xml langs.model.xml shortcuts.xml stylers.model.xml
-BIN_DATA := change.log doLocalConf.xml readme.txt userDefineLangs/
+SRC_DATA := contextMenu.xml langs.model.xml shortcuts.xml stylers.model.xml tabContextMenu_example.xml
+BIN_DATA := change.log doLocalConf.xml nppLogNulContentCorruptionIssue.xml readme.txt userDefineLangs/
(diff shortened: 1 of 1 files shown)


donho:
> Edit: hmm `tabContextMenu_example.xml` is added only to `minimalist` ([packageAll.bat](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/packageAll.bat)), why not also to full?

@ArkadiuszMichalski 
Is it not the case? If it's not, please add it to the full ones - it's already in installer.

ArkadiuszMichalski:
@donho Here or open next PR?

donho:
@ArkadiuszMichalski 
OK, I just added - thank you for your heads up!

donho:
@ArkadiuszMichalski 

> [`toolbarIcons.xml`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/toolbarIcons.xml) cannot be added to the official pack? Its default content will not affect the program, but it shows that Notepad++ offers the ability to change icons.

Nice caught. Could you take care (create a new PR) about it?

ArkadiuszMichalski:
>Nice caught. Could you take care (create a new PR) about it?

OK I'll create it a bit later.

PR #12489, opened 2022-11-10
Add Pascal/Delphi FunctionList parser

dinkumoil:
Fixes #3664

PR #12686, opened 2022-12-21
Update pascal.xml to fix issue with Pascal units

Zaaphod:
Pascal Units require a definition of each function and procedure as a header of the unit.  This causes the function list to have duplicates of each function and procedure in the definition.  To solve this duplication and make the list only take you to the actual functions, I've added another comment definition that uses the keywords Interface which must always be before the function definitions and implementation which always must be after the definitions and before the actual functions.  so everything between Interface and Implementation is ignored the way it should be.  If someone wants the duplicates for some strange reason, they can always comment that line out.


vinsworldcom:
@Zaaphod you need to link an issue.  I've create one for you.

Add the text "fix #12687, updates #12489" to your commit message please.

Cheers.

Zaaphod:
@vinsworldcom Thank your help.  I think I updated the commit message.

Issue #12685, opened 2022-12-21
FR: Remove "stylerTheme" attribute in config.xml


donho:
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12671#issuecomment-1360979204

PR #12691, opened 2022-12-22
Remove stylerTheme from config.xml

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12685


PowerEditor/src/WinControls/ColourPicker/WordStyleDlg.cpp
```diff
@@ -405,7 +405,7 @@ intptr_t CALLBACK WordStyleDlg::run_dlgProc(UINT Message, WPARAM wParam, LPARAM
 						::SendMessage(_hParent, WM_UPDATEMAINMENUBITMAPS, 0, 0);
 
 						const TCHAR* fn = ::PathFindFileName(_themeName.c_str());
-						NppDarkMode::setThemeName(lstrcmp(fn, L"stylers.xml") == 0 ? L"" : fn);
+						NppDarkMode::setThemeName((!NppDarkMode::isEnabled() && lstrcmp(fn, L"stylers.xml") == 0) ? L"" : fn);
```
ozone10, line 408: allow styler.xml to be used in dark mode

Issue #9710, opened 2021-03-28
Document List panel is not displayed when new instance is run


voronovmaksim:
### Description of the Issue
Document List panel a.k.a doc switcher is not displayed when new instance is run

### Steps to Reproduce the Issue

1. Enable document list panel (Doc panel is displayed in main instance)
2. Run second instance (alt + f6 or just clcik by icon of npp if multi-instance = always in multi-instance mode)

### Expected Behavior
Doc switcher is dispalyed

### Actual Behavior
Doc switcher is not dispalyed. 

I am forced  to go to Preferences->general->Document list panel->show is enable in otder to display it in new instance of npp.

win10 64bit, 7.9.5 version is used.

dinkumoil:
This issue has not only impact on `Document List` panel, **ALL** docked side panels are affected. Thus, multi instance mode is not usable anymore.

This is how Notepad++ v7.9.5 looks when only **one** instance is running:
![grafik](https://user-images.githubusercontent.com/24699812/116822227-eff4ed80-ab7d-11eb-9723-7955a166461e.png)

This is how it looks when it is started as **additional** instance (it makes no difference if the already running instance is also v7.9.5 or a much older one, e.g. v7.5.6):
![grafik](https://user-images.githubusercontent.com/24699812/116822238-08fd9e80-ab7e-11eb-9d73-2bff43caaf9c.png)

Panzerbjrn:
Is this something that is being worked on? It is a really annoying bug in an otherwise great product....

ArkadiuszMichalski:
Try run another instance with `-multiInst` argument. These Notepad++ commands which you mention probably use `-nosession`, and it closes all panels (see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9487).

Panzerbjrn:
> Try run another instance with `-multiInst` argument. These Notepad++ commands which you mention probably use `-nosession`, and it closes all panels (see #9487).

Sadly, that gives me the error:  "C:\Program Files (x86)\Notepad++\-multiinst" doesn't exist. Create it?
![image](https://user-images.githubusercontent.com/10132417/125609922-14c0f1a9-0c9a-4a31-9d34-148e02b2ef0c.png)

That's from a command line. I don't see an option for this when rightclicking a file in the Document List, and choosing "Open in New Instance":
[...]

PR #12699, opened 2022-12-24
Add setting for panels to ignore '-nosession'

ozone10:
<s>currently missing translation</s>

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9710


ozone10:
rebase onto master to check something

donho:
@ozone10 
What do you think `Remember Panel State in Multi-instance (-nosession) *` instead of   `Panel state ignore -nosession *` ?

ozone10:
@donho 
`-nosession` can be used even in mono-instance. And unfortunately there is no space for such long title.
There might be problem with translations, text running out of space.

Maybe adding [?] button or static text with more description.

donho:
@ozone10 
> `-nosession` can be used even in mono-instance. And unfortunately there is no space for such long title.

Yes, indeed.
What I think about is people who know nothing about Notepad++ command line arguments. 

> Maybe adding [?] button or static text with more description.

That's a good idea, so that will give the full explanation to the both sides of people.

Do you mind that I join this PR to finish that, or do you prefer to take care of it by yourself?

ozone10:
@donho 
I've added description (static text), if you want [?] button instead I will leave that to you.
image
image

ozone10:
Maybe keep only `Panel State *` for title

donho:
@ozone10 
It's much clear. Thank you!

PR #12703, opened 2022-12-26
Fix autocomplete to sort case insensitive

### LINKED ISSUE #12495: Autocompletion for ignoreCase="yes" and mixed words case works incorrect
When we have `ignoreCase="yes"` (default) and mixed words in `xml` file then the popup does not work properly.

Str:
1. Open Notepad++ and select `ActionScript` language.
2. Press `a`. We will get:
image
This is correct. We get a list of words starting with `a` and `A` (because we ignore case), and `a` is selected because it fits better in the text. 
3. Press `b`.  Popup will be closed. It's wrong because the word `abs` exist on list.

Same happend when in step 3. we press `c` or `d`. The interesting thing is that there is no problem with e.g. `l`, `n`, `p`, `s`. 

Another casae:
1. Open Notepad++ and select `ActionScript` language.
2. Press `a`. We will get:
image
This is correct. We get a list of words starting with `a` and `A` (because we ignore case), and `a` is selected because it fits better in the text. 
3. Press `r`. We will get:
image
Still acceptable because it fits better than `Array`.
4. Press `r`.  Popup will be closed. It's wrong because the word `Array` exist on the list.

This works fine when we have only one variant of `name` value in `.xml` (only upper or only lower) or when we set `ignoreCase="no"`.
But still I want to be able to put on the list mixed case even when language is case insensitive (names are more readable and often appear in this form in the specification).

Another question: would not be better if pressing next case will refresh popup list to show only the matching items? Now when we press one char like `a` we get list of all words started with `a`, now when press `b` this list is the same (still show all words started with `a`) but select `ab` word.
The problem is when we have `AZ1, AZ2, AZ3...` and `az` word and there are a lot of other words started wit `a` or `A` between them. Now when on the list `az` word was selected we don't see (don't know) that words like `AZ1, AZ2, AZ3....` exist (we don't see it). If pressing next case will refresh popup list to show only the matching items then this problem would be reduced (even if a scroll appears, it won't be so long).

```
Notepad++ v8.4.6   (32-bit)
Build time : Nov  7 2022 - 21:47:42
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 21H2
OS Build : 22000.1219
Current ANSI codepage : 1250
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
```

mpheath:
Fix #12495

Currently, case sensitive sort is done on the list items for autocomplete. This fix sorts the list items as case sensitive or case insensitive based on the value of `_ignoreCase`. This will improve the autocomplete list to operate better for case insensitive selections.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -376,7 +376,13 @@ bool AutoCompletion::showApiAndWordComplete()
 
 	// Sort word array and convert it to a single string with space-separated words
 
-	sort(wordArray.begin(), wordArray.end());
+	if (_ignoreCase)
+		sort(wordArray.begin(), wordArray.end(), [](const generic_string &a, const generic_string &b)
+		{
+			return (generic_stricmp(a.c_str(), b.c_str()) < 0);
+		});
+	else
+		sort(wordArray.begin(), wordArray.end());
 
 	generic_string words;
 
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -667,7 +667,13 @@ bool AutoCompletion::showWordComplete(bool autoInsert)
 
 	// Sort word array and convert it to a single string with space-separated words
 
-	sort(wordArray.begin(), wordArray.end());
+	if (_ignoreCase)
+		sort(wordArray.begin(), wordArray.end(), [](const generic_string &a, const generic_string &b)
+		{
+			return (generic_stricmp(a.c_str(), b.c_str()) < 0);
+		});
+	else
+		sort(wordArray.begin(), wordArray.end());
 
 	generic_string words(TEXT(""));
 
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -268,6 +268,25 @@ static bool isAllDigits(const generic_string &str)
 	return true;
 }
 
+void sortInsensitive(vector<generic_string> &wordArray)
+{
+	sort(
+		wordArray.begin(),
+		wordArray.end(),
+		[](const generic_string &a, const generic_string &b)
+		{
+			return lexicographical_compare(
+				a.begin(), a.end(),
+				b.begin(), b.end(),
+				[](const char &ch1, const char &ch2)
+				{
+					return toupper(ch1) < toupper(ch2);
+				}
+			);
+		}
+	);
+}
+
 
 bool AutoCompletion::showApiComplete()
 {
@@ -378,10 +397,7 @@ bool AutoCompletion::showApiAndWordComplete()
 	// Sort word array and convert it to a single string with space-separated words
 
 	if (_ignoreCase)
-		sort(wordArray.begin(), wordArray.end(), [](const generic_string &a, const generic_string &b)
-		{
-			return (generic_stricmp(a.c_str(), b.c_str()) < 0);
-		});
+		sortInsensitive(wordArray);
 	else
 		sort(wordArray.begin(), wordArray.end());
 
@@ -668,10 +684,7 @@ bool AutoCompletion::showWordComplete(bool autoInsert)
 	// Sort word array and convert it to a single string with space-separated words
 
 	if (_ignoreCase)
-		sort(wordArray.begin(), wordArray.end(), [](const generic_string &a, const generic_string &b)
-		{
-			return (generic_stricmp(a.c_str(), b.c_str()) < 0);
-		});
+		sortInsensitive(wordArray);
 	else
 		sort(wordArray.begin(), wordArray.end());
 
@@ -1210,10 +1223,7 @@ bool AutoCompletion::setLanguage(LangType language)
 		}
 
 		if (_ignoreCase)
-			sort(_keyWordArray.begin(), _keyWordArray.end(), [](const generic_string &a, const generic_string &b)
-			{
-				return (generic_stricmp(a.c_str(), b.c_str()) < 0);
-			});
+			sortInsensitive(_keyWordArray);
 		else
 			sort(_keyWordArray.begin(), _keyWordArray.end());
 
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -312,8 +312,14 @@ bool AutoCompletion::showApiComplete()
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
 	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
+
+	if (_ignoreCase)
+		_pEditView->execute(SCI_AUTOCSETORDER, SC_ORDER_PERFORMSORT);
+
 	_pEditView->showAutoComletion(curPos - startPos, _keyWords.c_str());
 
+	if (_ignoreCase)
+		_pEditView->execute(SCI_AUTOCSETORDER, SC_ORDER_PRESORTED);
 	return true;
 }
 
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -312,14 +312,8 @@ bool AutoCompletion::showApiComplete()
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
 	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
-
-	if (_ignoreCase)
-		_pEditView->execute(SCI_AUTOCSETORDER, SC_ORDER_PERFORMSORT);
-
 	_pEditView->showAutoComletion(curPos - startPos, _keyWords.c_str());
 
-	if (_ignoreCase)
-		_pEditView->execute(SCI_AUTOCSETORDER, SC_ORDER_PRESORTED);
 	return true;
 }
 
@@ -1209,7 +1203,13 @@ bool AutoCompletion::setLanguage(LangType language)
 			}
 		}
 
-		sort(_keyWordArray.begin(), _keyWordArray.end());
+		if (_ignoreCase)
+			sort(_keyWordArray.begin(), _keyWordArray.end(), [](const generic_string &a, const generic_string &b)
+			{
+				return (generic_stricmp(a.c_str(), b.c_str()) < 0);
+			});
+		else
+			sort(_keyWordArray.begin(), _keyWordArray.end());
 
 		for (size_t i = 0, len = _keyWordArray.size(); i < len; ++i)
 		{
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -311,6 +311,7 @@ bool AutoCompletion::showApiComplete()
 	_pEditView->execute(SCI_AUTOCSETTYPESEPARATOR, WPARAM('\x1E'));
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
+	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
 	_pEditView->showAutoComletion(curPos - startPos, _keyWords.c_str());
 
 	return true;
@@ -421,6 +422,7 @@ bool AutoCompletion::showApiAndWordComplete()
 	_pEditView->execute(SCI_AUTOCSETTYPESEPARATOR, WPARAM('\x1E'));
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
+	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
 	_pEditView->showAutoComletion(curPos - startPos, words.c_str());
 	return true;
 }
@@ -616,6 +618,7 @@ void AutoCompletion::showPathCompletion()
 	// Show autocompletion box.
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM('\n'));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, true);
+	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, true);
 	_pEditView->showAutoComletion(rawPath.length(), autoCompleteEntries.c_str());
 	return;
 }
@@ -679,6 +682,7 @@ bool AutoCompletion::showWordComplete(bool autoInsert)
 
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
+	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
 	_pEditView->showAutoComletion(curPos - startPos, words.c_str());
 	return true;
 }
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -278,7 +278,7 @@ void sortInsensitive(vector<generic_string> &wordArray)
 			return lexicographical_compare(
 				a.begin(), a.end(),
 				b.begin(), b.end(),
-				[](const char &ch1, const char &ch2)
+				[](const wchar_t &ch1, const wchar_t &ch2)
 				{
 					return toupper(ch1) < toupper(ch2);
 				}


mpheath:
This option

``` cpp
_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
```

https://www.scintilla.org/ScintillaDoc.html#SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR

tests as being better case insensitive behaviour with the matching in the autocomplete list. View this issue [comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12495#issuecomment-1365560810) regarding this option.

It just needs to inserted before this line that shows the autocomplete 

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/d87e996ff2da36b883906c3246b370b81b8653fc/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L418

Perhaps could be included with this PR if suitable? I would rather ask first than commit and revert the commit if the option is rejected.

mpheath:
Initial commit sorts case insensitive functions and words.
Second commit sets case insensitve behaviour.
Third commit to sort case insensitive for functions.

Word completion is still case sensitive sort. Please advise if word completion should be sorted case insensitive if needed.

ArkadiuszMichalski:
@mpheath 
In third commit sorting will be done with every show of autocomplete? If yes, maybe it's better to do sorting just after reading the xml file?
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/43b236aec65928d18c1362630b70994904025c09/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L1212

mpheath:
@ArkadiuszMichalski 
Nice observation. Certainly agree as it is being case sensitive sorted there so may as well do case insensitive sort there as needed which will eliminate the extra sorting being done in `showApiComplete()`. Expect another commit to implement this change.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
```diff
@@ -673,6 +682,7 @@ bool AutoCompletion::showWordComplete(bool autoInsert)
 
 	_pEditView->execute(SCI_AUTOCSETSEPARATOR, WPARAM(' '));
 	_pEditView->execute(SCI_AUTOCSETIGNORECASE, _ignoreCase);
+	_pEditView->execute(SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR, _ignoreCase);
```
ArkadiuszMichalski, line 685: @mpheath Any example how this line change actual behavior?

ArkadiuszMichalski:
>@mpheath
>Word completion is still case sensitive sort. Please advise if word completion should be sorted case insensitive if needed.

Good question. Looks like because of this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ff57f06b3d534fbb5ce5dbe8d3dfa91f760d7eb0/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L684
also this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/43b236aec65928d18c1362630b70994904025c09/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L676
should be correct (same as in other places).

Now I have wrong behavior for list like this (set ActionScript):
```
Acc
Acta
ActTa
AcTT
AcTa
```
When press `Act` or `AcT` it select `ActTa`, but should `AcTT`. Basically it can't select `AcTT`. 
image
When switch to `Function and world completion` then it works better:
image
so sorting for `Word completion` needs correct.

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: It may have no real impact there though the setting `SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR`  needs to be reset on the point of showing the calltip, else it inherits the last setting like is in path completion which is always sets it to true.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6a8a446eb61f96b00f74ae5baa0894ff16e7b18f/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L612

Word completion seems to be always be case sensitive, so the `SCI_AUTOCSETIGNORECASE` set to ignore case seems a strange as variable though it is there. And since it is there and `SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR` may need to be set to what that autocomplete function requires, it has been inserted there.

mpheath:
> Now I have wrong behavior for list like this (set ActionScript):

I will look into this. I may have to set `_ignoreCase` to `true` in this function for both `SCI_AUTOCSETIGNORECASE` and `SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR` and test the behaviour. Thanks for the STR.

ArkadiuszMichalski:
Change sorting is not better? We will have same order for `Word` and `Function and Word`.

mpheath:
> Change sorting is not better? We will have same order for `Word` and `Function and Word`.

Functions are read from the xml and have a setting of `ignoreCase="yes"` or `ignoreCase="no"`. Word completion has no such setting. So should word completion be case sensitive or case insensitive? Well, `getWordArray` uses `SCFIND_MATCHCASE` to get words and as Don mentioned to me in a [previous review](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11191#discussion_r809374017) that it is unrelated to function autocomplete. So I would rather not change it without some assurance, The ignore case setting should match the sort of the items in theory. If word completion is changed to ignore case to match the xml setting, then some confirmation might be needed for the change.

ArkadiuszMichalski:
But we don't change how `getWordArray` will work. Still will be case sensitive (use `SCFIND_MATCHCASE`). We just sort result of this work according to `ignoreCase`. This will keep the order in the autocomplete list the same for `Word` and `Function and Word`.

So these lines could stay as they are:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ff57f06b3d534fbb5ce5dbe8d3dfa91f760d7eb0/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L684-L685

Probably you can change above to `true` and don't change sort, but order of words in `Word` and `Function and Word` will be different. 

Can you check what will happens when you change:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ff57f06b3d534fbb5ce5dbe8d3dfa91f760d7eb0/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L670
to:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ff57f06b3d534fbb5ce5dbe8d3dfa91f760d7eb0/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L380-L386

Any other side effects besides the changed order of the words in the list?

ArkadiuszMichalski:
To always preserve case for `Word completion` this should be set:
```
pEditView->execute(SCI_AUTOCSETIGNORECASE, 0);
```
and my STR will works correct https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367004919.

It's a bit weird that `Word completion` and `Function and Word completion` works differently for `ignoreCase="no"`. Well, but if the `Word completion` option have to be always case-sensitive, then OK.

mpheath:
> It's a bit weird that `Word completion` and `Function and Word completion` works differently for `ignoreCase="yes"`. Well, but if the `Word completion` option have to be always case-sensitive, then OK.

I'll try to clear this confusion. Programming language case is important for case sensitive languages. Words like in code comments with programming languages,... or plain text files can be considered different. As such, case sensitive searching is done to build the list of words. Probable reason for case sensitive searching is to keep the list concise with what the user typed respecting case.

Now, that does not apply to sorting or selection of that list of words to my understanding.

SciTE as a comparison for word autocomplete, does follow the set language case to search for words. So Notepad++ is different with word search being only case sensitive. SciTE does not do `Function and Word completion` as one operation so Notepad++ has different behaviour with autocomplete. What Notepad++ does with word searching is a choice that was made and I am just trying to improve the autocomplete with sorting and selection.

I need more time for testing so I can find a better solution. So forgive my absence. I still try to monitor your comments while I am testing.

> `pEditView->execute(SCI_AUTOCSETIGNORECASE, false);`

Mixed case like `AcTAaB` can give unreliable results with this setting. Title case, all lowercase and all uppercase is probably good. Once mixing the autocomplete list with other mixed forms of case, then less than disirable events can happen. If case insensitive sorting and matching then accurate case sensitive matching is sacrificed for the autocomplete to remain open so as to find the desired word while typing. In my view, it may come down to a compromise.

mpheath:
@ArkadiuszMichalski 
I committed your suggestion of sort, `SCI_AUTOCSETIGNORECASE` and `SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR` based on `_ignoreCase`. My idea of just one way for all, I consider as overthinking the problem. Testing seems good for the later though the behaviour is different to `Function and Word completion` as you mentioned. So I did a full circle with testing and came back to your idea. If it works good enough is what matters I guess.

ArkadiuszMichalski:
@mpheath 

Sorry for confusion. From my last post https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367039433 I suggest change only `SCI_AUTOCSETIGNORECASE` and `SCI_AUTOCSETCASEINSENSITIVEBEHAVIOUR` to `false` and leave sorting as was (case sensitive) for `showWordComplete`. 
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ff57f06b3d534fbb5ce5dbe8d3dfa91f760d7eb0/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L684-L685
So for alone `Word completion` option we have always case sensitive behavior:
- when we obtain words by `getWordArray`
- when we select on autocomplete list (sort and selecting is cases sensitive)

Can you give me an example of when this changes would be wrong?

With your last commit we have:
- obtain words by `getWordArray` is always case sensitive
- select on autocomplete list (sort and selecting) depends on `ignoreCase` (so might be case insensitive). This would be more like how `Function and Word` completion works.

I don't know which behavior is more desirable for alone `Word completion` option. Anyway it needs to be corrected in one direction because it's not working as excepted right now (like I show earlier https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367004919). 

@donho can you look closer which behavior for `showWordComplete` is more expected?

mpheath:
The case insensitive sort algorithm needs to be compatible with Scintilla`s internal algorithm to make the autocomplete selection work properly.

Many of the functions like `generic_stricmp` do a lowercase compare. Scintilla does an uppercase compare. So with certain lists, the autocomplete selection failed to match due to the lowercase compare sort done before the autocomplete is displayed.

Quote [SCI_AUTOCSHOW](https://www.scintilla.org/ScintillaDoc.html#SCI_AUTOCSHOW) :

> One result of this is that the list should be sorted with the punctuation characters '[', '\', ']', '^', '_', and '`' sorted after letters.

Lowercase sort handles `a-z` as `\x61-\x7A` while uppercase sort handles `a-z` as `\x41-\x5A`. This affects the sorting order of characters `\x5B-\x60` which are between the uppercase alphabet and lowercase alphabet.

As for

> Can you give me an example of when this changes would be wrong?

There is no right or wrong in the word autocomplete sort. My choice was based on consistency of the sort rather than a contradiction of one autocomplete doing case sensitive sort and another doing case insensitive sort. What `getWordArray` does in searching is that it respects the case typed into the document to get a smaller word list. If the preference for everybody is to be inconsistent with the sorting of the autocompletion lists, then I could change the word complete sort to always be case sensitive if requested.

donho:
@ArkadiuszMichalski 
> @donho can you look closer which behavior for showWordComplete is more expected?

Sorry, I didn't see your request.

@mpheath @ArkadiuszMichalski 
Could you show me (with your scenario) the different behaviour between the original and this PR?

mpheath:
Based on actionscript.xml which should be case insensitive by uppercase comparison so the keywords with leading underscores should be below `A-Z` though above `a-z` if following a ASCII table. Scintilla does sorting and selection by the same algorithm for case insensitive lists.

Some original issues: 


<details>

<summary>showApiComplete()</summary>

Always case sensitive sort

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L1196

though the behaviour of the autocomplete is variable

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L313

and for ActionScript will be case insensitive behaviour.

This may fail with the order and selection in the autocomplete.

<hr>

</details>


<details>

<summary>showApiAndWordComplete()</summary>

Always case sensitive sort

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L379

though the behaviour of the autocomplete is variable

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L417

and for ActionScript will be case insensitive behaviour.

This may fail with the order and selection in the autocomplete.

<hr>

</details>


<details>

<summary>showWordComplete()</summary>

Always case sensitive sort

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L661

though the behaviour of the autocomplete is variable

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L675

and for ActionScript will be case insensitive behaviour.

This may fail with the order and selection in the autocomplete.

<hr>

</details>


What this PR does is to synchronize the sort of the list and behaviour of the autocomplete to make it compatible with Scintilla.

If not corrected, then selections can be mismatched and the autocomplete may close early as it fails to match words that may exist in the list.

If the order of the list items is not what Scintilla expects, then for a custom order to work correct may need an index to be created which could would be more work and would need to call

``` c++
_pEditView->execute(SCI_AUTOCSETORDER, SC_ORDER_CUSTOM);
```

or let Scintilla handle the sort order by

``` c++
_pEditView->execute(SCI_AUTOCSETORDER, SC_ORDER_PERFORMSORT);
```

These options are not efficient as the vector to string to vector to sort and possibly back to string... If the list was already a string then might be worth it to pass to Scintilla to internally sort. Indexing would probably create a cache for selection so would be intensive. So, getting the sort correct in Autocompletion.cpp would be the better option.

The sort needs to compatible with the selection code of

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f8593916a90bd1f37ce9820d2960607f28935fb7/scintilla/src/AutoComplete.cxx#L239

and what [SCI_AUTOCSHOW](https://www.scintilla.org/ScintillaDoc.html#SCI_AUTOCSHOW) advises about the sort algorithm needed.

Case sensitive languages should be the same as before so it is the case insensitive languages that will have different sorting and autocomplete behaviour.

Outstanding is perhaps `void sortInsensitive()` should be static and perhaps `toupper()` changed to `towupper()` for wide characters with use of type `wchar_t`. This could be updated with another commit.

donho:
@mpheath 

@ArkadiuszMichalski 
> @donho can you look closer which behavior for showWordComplete is more expected?

Honestly, I'm kind of like the behaviour of this PR, and I agree with @mpheath - when the words mixed with functions, it'll be better that the whole list has the same and unique behaviour. So when "Function & word completion" is selected in Auto-completion settings, it does make sense that words' order is case-insensitive in the auto-completion dropdown list, when the language is not case-sensitive.

OTOH, I'm wondering if it's good to make words case-insensitive if "Word completion" is selected in Auto-completion settings, even the language is not case-sensitive. Because in words list it's always case-sensitive, changing such behaviour might make users unhappy.

What do you think?

ArkadiuszMichalski:
@donho 
>Honestly, I'm kind of like the behavior of this PR (at least part of it), and I agree with @mpheath - when the words mixed with functions, it'll be better that the whole list has the same and unique order behavior . So when "Function & word completion" is selected in Auto-completion settings, it does make sense that words' order is case-insensitive in the auto-completion dropdown list to be aligned with functions, when the language is not case-sensitive.

When `Function & word completion` is selected we finally use only one list so the behavior will be consistent and rather dependent on the requirements for` Function`, because it's strictly related to the language and autocomplete `xml` file. This PR doesn't change the behavior for this setting here, it just fixes it (keep the sorting list consistent with Scintilla's behavior settings) because without it selecting in the list doesn't work properly (like on this example https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12495#issue-1445451873).

>OTOH, I'm wondering if it's good to make words case-insensitive if "Word completion" is selected in Auto-completion settings, even the language of document is not case-sensitive. Because in words list it's always case-sensitive, changing such behavior might make users unhappy.
What do you think?

For alone `Word completion` option we should talk about the two behaviors separately:
Code for test:
```
acc
Acc
Acta
ActTa
AcTT
AcTa
```
1. How we obtain words from document (case sensitive or case insensitive or follow with autocomplete `xml` file). From what I can see it's always been a case sensitive (regardless of other factors). There have been questions about this before, and it's pretty much established that this is the intended behavior and it's supposed to stay that way.
2. How we sort/select words in the list (case sensitive or case insensitive or follow with autocomplete `xml` file). From what I can see for selection the current master and this PR follow with `xml` file (and additionally this PR use correct sort). I propose here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367240055 to make this always case sensitive (same as we have in above point when obtain words from document).

Seems no one report any objection how selection on the list should work for `Word completion` (case sensitive/insensitive) so the current behavior (and from this PR) might stay. It's more important to do the correct sorting because without it nothing works properly.

Regardless of what we choose, it still be a bit confusing because `Word completion` and `Function completion`/`Function & word completion` work differently and we have to experiment to determine their proper behavior. Every time I'm back to this mechanism I forget that `Word completion` works differently.

>OTOH, I'm wondering if it's good to make words case-insensitive if "Word completion" is selected in Auto-completion settings, even the language of document is not case-sensitive. Because in words list it's always case-sensitive, changing such behavior might make users unhappy.

But where someone wanted to do case-insensitive for "Word completion" when document/language is case-sensitive? On the other hand, I'm not sure which exact behavior you're asking about when you writing "case sensitive/insensitive" (how obtain words from document, how sort/select words in list or both of them?).

donho:
@ArkadiuszMichalski 
IMO, how we obtain words from document is not so important since we sort the words/functions then put them into the list.
So "how we sort/select words in the list" is primordial to me. 

If my statement above is not unthoughtful and naïve, my following question, which is rather focused on the change of behaviour, is not related to "how we obtain words from document":
> I'm wondering if it's good to make words not case-sensitive if "Word completion" is selected in Auto-completion settings, even the language of document is not case-sensitive. Because in words list it's always case-sensitive, changing such behaviour might make users unhappy.

And I would like to have your opinion about that.

ArkadiuszMichalski:
@donho
>IMO, how we obtain words from document is not so important since we sort the words/functions then put them into the list.
So "how we sort/select words in the list" is primordial to me.

But before sorting we need to obtain them and depending on how we do it, we'll have it on the list. In the case of case insensitive languages, we still have case sensitive behavior in this aspect (here I mean only the behavior for `World completion`).

```
function Test(){
}
```

Writing `t` in any insensitive languages we never obtain `Test` word. Of course, if we assume that `Word completion` is supposed to work a bit differently, then OK. But in `Function and word completion` option this is a bit confusing because `Function` works in one way and `Word` works in another way (in context how we obtain words from document).

>I'm wondering if it's good to make words not case-sensitive if "Word completion" is selected in Auto-completion settings, when the language of document is not case-sensitive. Because in words list it's always case-sensitive, changing such behaviour might make users unhappy. 

Ahh previously I read it wrongly. I had the same questions here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367240055. So we might have two wariants:
- for `getWordArray` select is always case sensitive (same as we obtain words, what is always case sensitive)
- for `getWordArray` select on autocomplete list (sort and selecting) depends on `ignoreCase` (so might be case insensitive). This would be more like how `Function and Word completion` works.

> Because in words list it's always case-sensitive, changing such behaviour might make users unhappy.

But in this PR, we're not changing that.
Current master:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/535bd8fa0ea49f3a071c98f25fe743459baa37c7/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L675

This PR:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0158a31bd209ee18e34ba58eeae6a236486db3cd/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp#L703

This behavior still depends on the language and setting in the XML file.

If we set the above option to `false` for `Word completion` then the selection will always be case-sensitive, and then we change the current behavior.

I don't have any specific preference here. As I wrote, if so far there have been no reports regarding this behavior, there is no need to change it. It's more important to do the correct sorting (according to the Scintilla settings) because without it nothing works properly.

mpheath:
This PR is about fixing the case insensitive autocomplete. It might be beneficial to do what is needed now rather than creating another Issue and PR to resolve this entire subject.

Autocomplete setting is **function autocomplete**. Select the **Lua** language which is **case insensitive**. Enter the following into the document.

```
aab = 1

function AAC()
end

a
```

If the caret is positioned after the final `a` and press <kbd>Ctrl</kbd>+<kbd>Enter</kbd>, what autocomplete is expected?

```
aab
```

or

```
AAC
```

or

```
AAC
aab
```

It auto-inserts `aab`. This can be considered a bug to a case insensitive language as I would expect the 3rd option to show the autocomplete (if I did not know the current behaviour). To try to resolve this, undo back to `a`, backspace to remove `a` and then type `A` to try again. Press <kbd>Ctrl</kbd>+<kbd>Enter</kbd> and what autocomplete is expected? It auto-inserts `AAC`. The autocomplete that I would expect is never shown which is the 3rd option.

6 languages are case insensitive which are

 * AutoIt
 * Baan
 * Cobol
 * Lua
 * Powershell
 * SQL

so not sure of any resistance to correcting the behaviour. Most users will not notice. Those who use insensitive languages will appreciate the listing of words that should be there in the list.

Consider what the mind behind Scintilla implemented:

scite_word2

The users will be happy if the autocomplete works as expected and that has been not the case for a long time. When someone informs you that typing <kbd>backspace</kbd> naturally closes the autocomplete, then I can only say that it is buggy behaviour. Protecting buggy code is a bug. So, is it agreed that bugs should be fixed? as this would be a small fix.

ArkadiuszMichalski:
>6 languages are explicitly case insensitive which are
AutoIt
Baan
Cobol
Lua
Powershell
SQL

Default behavior is case insensitive so this list will be longer (without explicitly use `ignoreCase="no"` any language will be case insensitive).

donho:
@mpheath @ArkadiuszMichalski 
Sorry. I'm lost in your quite long discussion.
I need only an (or some) concrete example(s) to show the difference BEFORE & AFTER this PR, it's not necessary to mention the code (at least, not in the first step - that will lead more confusion).

Thank you

ArkadiuszMichalski:
@donho 
1 
Try steps from first post https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12495#issue-1445451873.

2 
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367004919
Set ActionScript and use `World completion`:
```
Acc
Acta
ActTa
AcTT
AcTa
```
When press `Act` or `AcT` it select `ActTa`, but should `AcTT`. Basically it can't select `AcTT`.

mpheath:
@donho I created an AutoIt3 script which reads actionscript.xml for the keywords, runs Notepad++ and types the keywords. If the autocomplete windows closes early then it is logged to file. 161 keywords failed to match which caused the autocomplete window to close early. The patched Notepad++ logged 0 keywords so passed the test.

[test_autocomplete.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10528950/test_autocomplete.zip)

This was tested with `Function and word completion`.

A readme.md in the zipfile describes more details. I do not expect the AutoIt3 script to be run as I have already done it and have supplied the logs.

donho:
@ArkadiuszMichalski 

Thank you for your simple demo
> 1
Try steps from first post https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12495#issue-1445451873.

I can reproduce this issue in v8.4.9.
And this PR has fixed this issue.

>2
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1367004919
Set ActionScript and use World completion:
Acc
Acta
ActTa
AcTT
AcTa
When press Act or AcT it select ActTa, but should AcTT. Basically it can't select AcTT.

I can also reproduce this issue in v8.4.9.
This PR has fixed the issue as well.

So do you validate this PR @ArkadiuszMichalski ?

ArkadiuszMichalski:
@donho 
>So do you validate this PR @ArkadiuszMichalski?

Yes, this PR fixes bugs I noticed.

There are other things mentioned here, but these should be considered separately, because it's hard to analyze such a long thread.
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1379757099
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1399154767
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1403361164

donho:
@ArkadiuszMichalski 
Thank you for the confirmation!

> There are other things mentioned here, but these should be considered separately, because it's hard to analyze such a long thread.
> [#12703 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1379757099)
> [#12703 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12703#issuecomment-1403361164)

Please create 2 issues for above comment-discusions.
I've merged this PR into master branch. So these issues (and eventual fixes) will based on the more correct behaviour.

donho:
> I created an AutoIt3 script which reads actionscript.xml for the keywords, runs Notepad++ and types the keywords. If the autocomplete windows closes early then it is logged to file. 161 keywords failed to match which caused the autocomplete window to close early. The patched Notepad++ logged 0 keywords so passed the test.

Thank you @mpheath
I'm checking AutoIt to see how to work.
Could you give me a hint for the advantage  for the test of auto-complletion of Actionscript ?

mpheath:
@donho The AutoIt script was made to do a **before** and **after** test of the autocomplete with keywords from actionscript.xml. At about 161 in 627 keywords is about 1 in 4 failure rate with the before patched code. The after patched code recorded no failures. Not something that would like to test manually is the advantage. If you are considering to use as a unitTest for building Notepad++, I very much would **not** recommend. Internal testing of functions... is fairly safe though external use of automation can be risky when using [`Send()`](https://www.autoitscript.com/autoit3/docs/functions/Send.htm) to similate keystrokes.

PR #12710, opened 2022-12-27
[xml] Update german.xml to v8.4.8

Karlo-F:
Fixed "tab context menu is not translated anymore"


schnurlos:
Thx, nice catch.

Karlo-F:
It is included in #12728

Issue #12702, opened 2022-12-26
Add support selection to EOL commands


ArkadiuszMichalski:
Would be nice to have selection support for these two commands: `EOL to Space` and `Remove Unnecessary Blank and EOL`. Same as we have for trim operations https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12655.

![image](https://user-images.githubusercontent.com/2730894/209535882-a5305430-da51-42a4-9f0e-d4a0e1b72ca7.png)

@donho I have a few questions before starting this. In `EOL` case it's a bit simpler, because here it's only important to select `EOL`, so what we select will be taken into account. So for `EOL to Space`:

1. 
![image](https://user-images.githubusercontent.com/2730894/209536658-6231a7f8-be96-4da2-9261-7a06e16224c7.png)
Will replace EOL in 1 line.

2. 
![image](https://user-images.githubusercontent.com/2730894/209536806-59f75121-58cc-49b0-aab4-7747001b161a.png)
Will replace EOL in 1 and 2 line.

3.
![image](https://user-images.githubusercontent.com/2730894/209536952-fd7a5039-5d1a-471f-a026-981f8040dcfe.png)
Will replace EOL in 1 line.

In this case it doesn't make sense to extend the selection to the whole data? For example:
![image](https://user-images.githubusercontent.com/2730894/209537144-5c9d031b-189a-48fe-92cb-e9365c3d355d.png)
Expanding will not affect the process because the selected EOL will not change. Well, but I'm not sure to not expand anyway, because the Trim operations make expanding and this one `EOL to Space` should not do it? Also `Remove Unnecessary Blank and EOL` firstly run `doTrim()` so expanding will be done here anyway.
These two commands should be considered in the context of whole lines or just according to the exactly selection made? A similar question will appear later for the last 3 commands `Tab to Space`, `Space to Tab (all/leading)`.

I have one more question about this command `Remove Unnecessary Blank and EOL`. Actually for such example:
![image](https://user-images.githubusercontent.com/2730894/209538580-c2d73df0-1192-459e-9615-889094df1709.png)
we get this:
![image](https://user-images.githubusercontent.com/2730894/209538128-1652e8e6-dad2-48a3-b924-14f7cb87ffac.png)

We get last 3 space because last 3 EOL was change to it. But this spaces are also not unnecessary, as command name suggest? Should not be removed?

One more example:
[...]

PR #12711, opened 2022-12-27
Support selection for EOL commands

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12702.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1534,6 +1534,13 @@ void Notepad_plus::doTrim(trimOp whichPart)
 	}
 }
 
+void Notepad_plus::eol2ws()
+{
+	bool isEntireDoc = _pEditView->execute(SCI_GETANCHOR) == _pEditView->execute(SCI_GETCURRENTPOS);
+	isEntireDoc ? _pEditView->execute(SCI_TARGETWHOLEDOCUMENT) : _pEditView->execute(SCI_TARGETFROMSELECTION);
+	_pEditView->execute(SCI_LINESJOIN);
+}
+
 void Notepad_plus::removeEmptyLine(bool isBlankContained)
 {
 	// whichPart : line head or line tail
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1536,9 +1536,48 @@ void Notepad_plus::doTrim(trimOp whichPart)
 
 void Notepad_plus::eol2ws()
 {
-	bool isEntireDoc = _pEditView->execute(SCI_GETANCHOR) == _pEditView->execute(SCI_GETCURRENTPOS);
-	isEntireDoc ? _pEditView->execute(SCI_TARGETWHOLEDOCUMENT) : _pEditView->execute(SCI_TARGETFROMSELECTION);
+	auto mainSelAnchor = _pEditView->execute(SCI_GETANCHOR);
+	auto mainSelCaretPos = _pEditView->execute(SCI_GETCURRENTPOS);
+	auto rectSelAnchorVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONANCHORVIRTUALSPACE);
+	auto rectSelCaretVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONCARETVIRTUALSPACE);
+	bool isRectSel = (_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE) || (_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_THIN);
+	bool isEntireDoc = (mainSelAnchor == mainSelCaretPos) && (rectSelAnchorVirt == rectSelCaretVirt);
+	auto docLength = _pEditView->execute(SCI_GETLENGTH);
+	auto lines = _pEditView->execute(SCI_GETLINECOUNT);
+	// auto-expand of partially selected lines
+	if (!isEntireDoc)
+	{
+		auto startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
+		auto startLine = _pEditView->execute(SCI_LINEFROMPOSITION, startPos);
+		auto endPos = _pEditView->execute(SCI_GETSELECTIONEND);
+		auto endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
+
+		if (startPos != _pEditView->execute(SCI_POSITIONFROMLINE, startLine))
+			startPos = _pEditView->execute(SCI_POSITIONFROMLINE, startLine);
+
+		if (endPos != _pEditView->execute(SCI_POSITIONFROMLINE, endLine) && endPos < _pEditView->execute(SCI_GETLINEENDPOSITION, endLine))
+			endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, endLine);
+
+		_pEditView->execute(SCI_SETSEL, startPos, endPos);
+	}
+	_pEditView->execute(isEntireDoc ? SCI_TARGETWHOLEDOCUMENT: SCI_TARGETFROMSELECTION);
 	_pEditView->execute(SCI_LINESJOIN);
+	// restore original selection if nothing has changed
+	if (!isEntireDoc && (docLength == _pEditView->execute(SCI_GETLENGTH)) && (lines == _pEditView->execute(SCI_GETLINECOUNT)))
+	{
+		if (isRectSel)
+		{
+			_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHOR, mainSelAnchor);
+			_pEditView->execute(SCI_SETRECTANGULARSELECTIONANCHORVIRTUALSPACE, rectSelAnchorVirt);
+			_pEditView->execute(SCI_SETRECTANGULARSELECTIONCARET, mainSelCaretPos);
+			_pEditView->execute(SCI_SETRECTANGULARSELECTIONCARETVIRTUALSPACE, rectSelCaretVirt);
+		}
+		else
+		{
+		_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
+		_pEditView->execute(SCI_SETCURRENTPOS, mainSelCaretPos);
+		}
+	}
 }
 
 void Notepad_plus::removeEmptyLine(bool isBlankContained)
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1493,11 +1493,13 @@ void Notepad_plus::doTrim(trimOp whichPart)
 	env._searchType = FindRegex;
 	auto mainSelAnchor = _pEditView->execute(SCI_GETANCHOR);
 	auto mainSelCaretPos = _pEditView->execute(SCI_GETCURRENTPOS);
-	auto rectSelAnchorVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONANCHORVIRTUALSPACE);
-	auto rectSelCaretVirt = _pEditView->execute(SCI_GETRECTANGULARSELECTIONCARETVIRTUALSPACE);
-	bool isRectSel = (_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE) || (_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_THIN);
-	bool isEntireDoc = (mainSelAnchor == mainSelCaretPos) && (rectSelAnchorVirt == rectSelCaretVirt);
+	bool isEntireDoc = (mainSelAnchor == mainSelCaretPos);
 	auto docLength = _pEditView->execute(SCI_GETLENGTH);
+
+	// block selection is not supported
+	if ((_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE) || (_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_THIN))
+		return;
+
 	// auto-expand of partially selected lines
 	if (!isEntireDoc)
 	{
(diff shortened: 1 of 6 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1534,6 +1534,13 @@ void Notepad_plus::doTrim(trimOp whichPart)
 	}
 }
 
+void Notepad_plus::eol2ws()
+{
+	bool isEntireDoc = _pEditView->execute(SCI_GETANCHOR) == _pEditView->execute(SCI_GETCURRENTPOS);
```
Yaron10, line 1539: Just curious: is there a reason not to use the previous common code?
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/31a392ae59e1dc3494c4193e7a3df3e4e34f3a5e/PowerEditor/src/Notepad_plus.cpp#L1561

A new function `getMainSelectionLenght()` could be used in most cases.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -1534,6 +1534,13 @@ void Notepad_plus::doTrim(trimOp whichPart)
 	}
 }
 
+void Notepad_plus::eol2ws()
+{
+	bool isEntireDoc = _pEditView->execute(SCI_GETANCHOR) == _pEditView->execute(SCI_GETCURRENTPOS);
+	isEntireDoc ? _pEditView->execute(SCI_TARGETWHOLEDOCUMENT) : _pEditView->execute(SCI_TARGETFROMSELECTION);
```
Yaron10, line 1540: Could also be:
`_pEditView->execute(isEntireDoc ? SCI_TARGETWHOLEDOCUMENT : SCI_TARGETFROMSELECTION);`

Yaron10:
תמונה

`EOL to Space`:
No change.

`Remove Unnecessary Blank and EOL`:
תמונה

Isn't that confusing?

ArkadiuszMichalski:
@Yaron10 

Name `Remove Unnecessary Blank and EOL` is wrong https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12702#issuecomment-1365705873, it's more like `Trim both and EOL to Space` or `Trim all and EOL to Space`, but `EOL to Space` is also wrong, it's more like join lines.

`SCI_LINESJOIN` not work on column selection, If we don't expand such selection to lines it will not work, but @donho don't want expand for `EOL to Space` https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12702#issuecomment-1365144908. Here, this expanding does not affect the modified data anyway, so it may be worth doing it so that both of these commands work consistently (along with Trim).

donho:
> Name `Remove Unnecessary Blank and EOL` is wrong [#12702 (comment)](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12702#issuecomment-1365705873), it's more like `Trim both and EOL to Space` or `Trim all and EOL to Space`, but `EOL to Space` is also wrong, it's more like join lines.

How about "Trim All and Join Lines" ?

Yaron10:
> Possible solution:
>
  >  expand block selection to standard selection for EOL to Space
    return for Remove Unnecessary Blank and EOL when detect block selection
    don't support block selection at all (for Trim and above two commands), just return.

> I thought it's not necessary. But if it could help the consistency, OK for expanding.

👍 

I would expand stream selections, and "don't support block selection at all" if it should shorten the code.

***

> How about:
>
>    Remove EOL to Space command. (It's redundant with Join Lines).
    Rename Remove Unnecessary Blank and EOL to Trim Both and Join Lines. ?

👍

ArkadiuszMichalski:
@donho 
I add autoexpand selection for partially selected lines for `EOL to Space` so this will fix Yaron example.

I checked more closely how `EOL to Space` and `Join Lines` commands works until now and they have some differences:
- `EOL to Space` always replace (or remove) the selected EOLs:
image
So here we visually select 2 lines (1,2), but in fact 3 lines was joined (1,2,3).

    `Join Lines` make some correction if nothing is selected on last line, so selected EOLs might change:
image
So here we visually select 2 lines (1,2) and 2 lines was joined (1,2).

- due to the above behavior also selecting everything works differently
`EOL to Space`
image
So here all lines was joined (1,2,3,4). If we want omit 4 line we should skip last EOL in selection:
image
`Join Lines` always skip last EOL so if it was such EOL then the empty line will always be there:
image
So here we visually select 3 lines (1,2,3) and 3 line was joined (1,2,3). It is not possible to join all lines into one here (last empty line will always stay).

So:
`EOL to Space` - does not strictly replace every EOL to space. If it's not needed it's just removed it. 
image
We still have one space (not two). If someone really wants to replace every EOL to space they can use Find/Replace dialog.
Actually `EOL to Space` command also not work like `Join Lines` command. In addition to the above differences `Join Lines` also not work without any selection (we need press `Ctrl+A`).
`EOL to Space` seems to be more flexible than `Join Lines` because we have more control what we select. I don't know if there's a better name for this behavior than `EOL to Space`, but if there isn't, it's best not to change it.

Maybe preserve `EOL to Space` command? It's been around for a long time and no one has reported it. Changing name `Remove Unnecessary Blank and EOL` to `Trim both and EOL to Space` would be more clear that it's a combination of these two commands.  

If someone need `Trim both and Join Lines` then always can record such macro with `Trim Leading/Trailing` + `Join Lines` commands.

The last commit doesn't change the existing behavior without selection (all works the same). I only added selection support.

But if `EOL to Space` should be removed and rename `Remove Unnecessary Blank and EOL` to `Trim Both and Join Lines` then how `Join Lines` should behave? Exactly like now what I show in above picture, or something should be tweak also for this command?

Yaron10:
@ArkadiuszMichalski,

> I checked more closely how EOL to Space and Join Lines commands works until now and they have some differences...

Very nice! 👍 

> In addition to the above differences Join Lines also not work without any selection (we need press Ctrl+A).

I thought that was next on your list. :)
It seems necessary now. 

And I think that without selection, the last empty line should be removed as well.

***

> I add autoexpand selection for partially selected lines for EOL to Space so this will fix Yaron example.

תמונה

`EOL to Space`:
תמונה

IMO it does not make sense to  remove the EOL of line 1 only. What's the difference between line 1 and line 2?
I thought Don agreed with that.

If that's the case, `EOL to Space` should include partially _stream-selected_ lines as well.
This may be useful and expected in real-life: the user might select lines from the editor (i.e. not from the Lines Margin), and not notice that the last EOL (in the selected lines) was actually not selected.

תמונה

ArkadiuszMichalski:
From such picture It's hard to guess what you're doing here:
image

>IMO it does not make sense to remove the EOL of line 1 only. What's the difference between line 1 and line 2?

If you want to remove both then you have to explicitly indicate it:
image

How expanding lines we established in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12655#issuecomment-1358749687. Currently this expanding was done only for block selection to work (because Trim supports it so let works also here). But we don't select more EOL's from what has been selected by user. Without expanding this will also works for standard selection, but still only on selected EOLs, so nothing was changes here.

Yaron10:
@ArkadiuszMichalski,

> From such picture It's hard to guess what you're doing here:

You're right.

Column selection.
תמונה

Why is line 4 "more selected" than line 5? And if the user starts the selection from the bottom up?

@donho,

What do you think?

donho:
@ArkadiuszMichalski 
> Maybe preserve EOL to Space command? It's been around for a long time and no one has reported it. Changing name Remove Unnecessary Blank and EOL to Trim both and EOL to Space would be more clear that it's a combination of these two commands.

The points are accepted.
Let's preserve EOL to Space command and changing name "Remove Unnecessary Blank and EOL" to "Trim both and EOL to Space". 

@Yaron10 
> Why is line 4 "more selected" than line 5?

Sorry, I don't follow you?

> And if the user starts the selection from the bottom up?

From bottom up or not, it's strange to me that EOL are replaced while the column selection being not on CRLF.

@ArkadiuszMichalski 
I'm finally for `don't support block selection at all (for Trim and above two commands), just return.`

Yaron10:
@donho,

> I'm finally for don't support block selection at all (for Trim and above two commands), just return.

👍 

> Sorry, I don't follow you?

It's irrelevant if "don't support block selection at all".

***

How about the following case?

תמונה

The same selection when `Show All Characters` is OFF (_normal use_):
תמונה

Wouldn't _excluding_ line 9 be confusing?
I think that *including* partially stream-selected lines (even if the EOL is not selected) would be clearer and more consistent.
And as you've mentioned recently: "Those commands are Line Operations".

Yaron10:
And what should happen on `Trim both and EOL to Space`?
Include line 9 for Trim and exclude it for `EOL to Space`?

donho:
@Yaron10 

image

For me in screen 1 the EOL of Line 4 (Test4) is selected so the Test4CRLF should be treated.
Whereas the  EOL of Line 4 in screen 2 is NOT selected, so the Test4CRLF should NOT be treated.

ArkadiuszMichalski:
@donho 

For me it's also visible and clear. When we have more control over selected EOL in the command `EOL to Space` then this command:
- can be treated more as `Join Lines` when we don't select last EOL in selection
- can be treated directly as `EOL to Space` when we select last EOL in selection

So still we have a choice, depending on what we need.

>I'm finally for don't support block selection at all (for Trim and above two commands), just return.

Hmm `Join Lines` support block selection. but if you don't want it here, we can remove it. But what should be in such a situation?

image

We use `Alt+Left mouse` outside the text (virtual selection). But the selection is still empty. We should return here or execute the command on the whole data (because nothing was selected)?

ArkadiuszMichalski:
OK, I removed block selection support for this commands. The code is a bit simpler. If in the future there are asking to support it, then we can think about it.

donho:
@ArkadiuszMichalski 
> Hmm `Join Lines` support block selection. but if you don't want it here, we can remove it. But what should be in such a situation?

Join Lines is a Line operation and column selection selects/includes 1 or several lines so it's not wired to users  (or to me).
Whereas "EOL to space" is NOT a Line operation, and as you said, it's different from "Join Lines", so we don't need to align the behaviour between "Join Lines" & "EOL to space".

Trim operations can be considered as a Line operation. So I think column selection could be treated on them. However, we have a command "Trim both and EOL to Space" which is combined with Trim operations (can use column selection) and "EOL to Space" (cannot use column selection).
I see 2 possible behaviours for the set of operations:
1. Trim operations treat column selection and "EOL to Space" does not treat column selection, and "Trim both and EOL to Space" does not treat column selection.
2. Trim operations, "EOL to Space" command, and "Trim both and EOL to Space" all do not treat column selection.

@ArkadiuszMichalski & @Yaron10 which solution you think more reasonable?

> We use Alt+Left mouse outside the text (virtual selection). But the selection is still empty. We should return here or execute the command on the whole data (because nothing was selected)?

It's for sure that the column selection outside of all lines shouldn't be treated. And this remind me another use case:
How about a column selection contains ONLY one long line and outside of all the other lines?
But if the behaviour 2 is chosen, then the problem above doesn't exist anymore, and the implementation will be much simpler.

ArkadiuszMichalski:
@donho 

I follow with 2 behavior, so always return when we detect column mode. This will be simpler and will not generate new borderline cases. 

I will just add that for me there was no problem with the interpretation of column mode for `EOL to Space`.
image

But now I've removed it. In practice, it's probably easier to make a normal selection than a block selection, so handling it seems unnecessary.

donho (changes requested):
The following line should be modified:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Notepad_plus.rc#L527

ArkadiuszMichalski:
@donho 

Right, can you cancel build?

donho (approved):

Yaron10:
> However, we have a command "Trim both and EOL to Space" which is combined with Trim operations (can use column selection) and "EOL to Space" (cannot use column selection).
I see 2 possible behaviours for the set of operations:
>
  >  Trim operations treat column selection and "EOL to Space" does not treat column selection, and "Trim both and EOL to Space" does not treat column selection.
    Trim operations, "EOL to Space" command, and "Trim both and EOL to Space" all do not treat column selection.

>> I follow with 2 behavior, so always return when we detect column mode. This will be simpler and will not generate new borderline cases.

👍 

But what about `Trim both and EOL to Space` in the following _stream selection_ case?
תמונה

I do respect and understand your approach to exclude partially selected lines in `EOL to Space`.
But I think that consistency and having less complicated rules (for @pryrt and the users reading his great documentation) are more important. :)

IMO, `EOL to Space` can certainly be interpreted as: execute the command on the selected lines (any line which have a selection).

Just some final thoughts. 
Thank you both. 👍

donho:
@Yaron10 

Here the test (your example) with this PR by using "Trim both and EOL to Space":
image

The result is expected (for me). What is the result you expected?

Yaron10:
@donho,

> However, we have a command "Trim both and EOL to Space" which is combined with Trim operations (can use column selection) and "EOL to Space" (cannot use column selection).

That was a very good point. It would have been very confusing.
Let's slightly modify your comment:

However, we have a command "Trim both and EOL to Space" which is combined with Trim operations (_which **includes** partially selected lines_) and "EOL to Space" (_which **excludes** partially selected lines_).

Isn't that as confusing and unexpected as your column-selection case?

donho:
@Yaron10 
> However, we have a command "Trim both and EOL to Space" which is combined with Trim operations (_which **includes** partially selected lines_) and "EOL to Space" (_which **excludes** partially selected lines_).

> Isn't that as confusing and unexpected as your column-selection case?

Which is your suggestion then?

Yaron10:
@donho,

I think that `EOL to Space` can be interpreted as: execute the command on the selected lines.
That is: include partially _stream-selected_ lines even if the EOL is not selected.

ArkadiuszMichalski:
Actually selection works like this:
- `EOL to Space` is not expand at all, you decide what you select.
- `Trim` is expand to full lines, but without last EOL in last line (so will not affect next command).
- `Trim both and EOL to Space` is just sequence `Trim` + `EOL to Space`. Expanding here from `Trim` doesn't affect `EOL to Space` so everything is fine. It's the same as if you manually invoked `Trim` and immediately `EOL to Space`.

If you want to include specific EOLs, just select them explicitly. I see no problem here.

PR #12712, opened 2022-12-27
Add toolbarIcons.xml to the build process

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12682#issuecomment-1365151173.

@donho I don't know exactly how this should be added to `copyCommonFiles` (if it should):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/58416114da14536dcc7a8ad36010da3382366ba6/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh#L73

### DIFF
--- a/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh
+++ b/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh
@@ -75,13 +75,13 @@ Function copyCommonFiles
 	SetOutPath "$UPDATE_PATH\"
 	File "..\bin\contextMenu.xml"
 	File "..\src\tabContextMenu_example.xml"
-	
+	File "..\src\toolbarIcons.xml"
+
 	SetOverwrite on
 	SetOutPath "$INSTDIR\"
 	File "..\bin\langs.model.xml"
 	File "..\bin\stylers.model.xml"
 	File "..\bin\contextMenu.xml"
-	File "..\bin\toolbarIcons.xml"
 
 	SetOverwrite off
 	File "..\bin\shortcuts.xml"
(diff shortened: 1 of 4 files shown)


PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh
```diff
@@ -81,6 +81,7 @@ Function copyCommonFiles
 	File "..\bin\langs.model.xml"
 	File "..\bin\stylers.model.xml"
 	File "..\bin\contextMenu.xml"
+	File "..\bin\toolbarIcons.xml"
```
donho, line 84: Should be `..\src\toolbarIcons.xml`

donho: And move up this line just behind `File "..\src\tabContextMenu_example.xml"` - due to `SetOverwrite on` (overwrite is turned ON, but installer should not overwrite user's data - icon could be already set).

donho (changes requested):

PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh
ArkadiuszMichalski: @donho 

But `toolbarIcons.xml` should be add only here (after this line):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f4517ef17e8dda08472166eda10d5cc3c52bcde4/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh#L77
where we have `SetOutPath "$UPDATE_PATH\"`

or also here (after this line)?
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f4517ef17e8dda08472166eda10d5cc3c52bcde4/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh#L87
where we have `SetOutPath "$INSTDIR\"`

donho (approved):

PR #12718, opened 2022-12-28
[xml] Update Turkish localization

Tmp341:
* Checked diff between 9809e2f and aa8ae48 commits via [Compare](https://github.com/notepad-plus-plus/notepad-plus-plus/compare/9809e2f..aa8ae48#diff-a6ac7ceba70d88bf1547fd1defd760bd55052dcdb78c44f9d46d99ef1f450472).
* I couldn't made changes for some time, hope these won't break something.


donho:
@Tmp341 
Please add `[xml]` as the prefix of title in your next commit.

Tmp341:
Okay. Will do.

Issue #12721, opened 2022-12-28
[Code] Undocumented functions for Windows XP


ozone10:
### Description of the Issue
Notepad++ does not support Windows XP anymore.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6a8a446eb61f96b00f74ae5baa0894ff16e7b18f/PowerEditor/src/Parameters.cpp#L1229

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6a8a446eb61f96b00f74ae5baa0894ff16e7b18f/PowerEditor/src/Parameters.cpp#L1236

Notepad++ should use [SetLayeredWindowAttributes](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setlayeredwindowattributes) and [EnableThemeDialogTexture ](https://learn.microsoft.com/en-us/windows/win32/api/uxtheme/nf-uxtheme-enablethemedialogtexture) directly instead of using undocumented way.

### Debug Information
Notepad++ v8.4.8

PR #12722, opened 2022-12-28
Replace undocumented functions

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12721


PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -167,8 +167,8 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	#define NPPM_MAKECURRENTBUFFERDIRTY (NPPMSG + 44)
 	//BOOL NPPM_MAKECURRENTBUFFERDIRTY(0, 0)
 
-	#define NPPM_GETENABLETHEMETEXTUREFUNC (NPPMSG + 45)
-	//BOOL NPPM_GETENABLETHEMETEXTUREFUNC(0, 0)
+	#define NPPM_GETENABLETHEMETEXTUREFUNC_DEPRECATED (NPPMSG + 45)
```
ozone10, line 170: @donho 
Plugin authors might need to update this header.

PowerEditor/src/Parameters.h
```diff
@@ -1545,7 +1545,7 @@ class NppParameters final
 	TiXmlDocument * getCustomizedToolIcons() const {return _pXmlToolIconsDoc;};
 
 	bool isTransparentAvailable() const {
-		return (_transparentFuncAddr != NULL);
+		return (_winVersion >= WV_VISTA);
```
ozone10, line 1548: [SetLayeredWindowAttributes](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setlayeredwindowattributes) is supported from Windows 2000 Professional but Notepad++ doesn't check if Windows 2000 is Professional or not.

ozone10: @donho 
btw
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6a8a446eb61f96b00f74ae5baa0894ff16e7b18f/PowerEditor/src/Parameters.cpp#L7623
does not return "Windows 11"

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
Yaron10: @ozone10,

As long as that message is kept, why change its name and cause potential incompatibility?

Thank you for your work. 👍

ozone10: @Yaron10 
the message can be removed, but for the sake of backward compatibility it is only renamed. The number value is not changed. `NPPM_GETENABLETHEMETEXTUREFUNC` used in plugin will be interpreted in Notepad++ as `NPPM_GETENABLETHEMETEXTUREFUNC_DEPRECATED`.
Renaming to `NPPM_GETENABLETHEMETEXTUREFUNC_DEPRECATED` will give plugin authors hint this message should not be used.
Plugin authors don't need to update, but it is recommended.
Plugin template should be updated.

rdipardo: > Renaming to `NPPM_GETENABLETHEMETEXTUREFUNC_DEPRECATED` will give plugin authors hint this message should not be used.

Why not just add [a C++-17-style annotation] to the *existing* message?

~~~cpp
enum { NPPM_GETENABLETHEMETEXTUREFUNC [[deprecated]] = (NPPMSG + 45) };
~~~

This makes every API call emit a compiler warning (potentially an error, if the "warnings as errors" option is set):

~~~
~\notepad-plus-plus\PowerEditor\src\NppBigSwitch.cpp(2662,8): warning C4996: 'NPPM_GETENABLETHEMETEXTUREFUNC': was declared deprecated
~~~

One problem, though: you'll have to add inline suppressions, or the AppVeyor build will fail:

~~~diff
diff --git a/PowerEditor/src/NppBigSwitch.cpp b/PowerEditor/src/NppBigSwitch.cpp
index fffe1cd8..4b52693e 100644
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2658,7 +2658,9 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return TRUE;
 		}

+#pragma warning(disable:4996)
 		case NPPM_GETENABLETHEMETEXTUREFUNC:
+#pragma warning(default:4996)
 		{
 			return (LRESULT)nppParam.getEnableThemeDlgTexture();
 		}

~~~

[a C++-17-style annotation]: https://en.cppreference.com/w/cpp/language/attributes/deprecated#:~:text=.%20(-,since%20C%2B%2B17,-)

PowerEditor/src/Parameters.h
donho: @ozone10 
OK, nice caught. I'll add case WV_WIN11.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
Yaron10: @ozone10,

How can the following line work properly if `NPPM_GETENABLETHEMETEXTUREFUNC` no longer exists?

https://github.com/pnedev/comparePlus/blob/a035877b65d77bebafb315b9743bd995f6c7278e/src/SettingsDlg/SettingsDialog.cpp#L53

Yaron10: @ozone10,

> The number value is not changed.

I missed the important part. 

Thanks again.

ozone10: @rdipardo 
I wanted to respect coding style, there are other macros with `_DEPRECATED` suffix. And using `[[deprecated]]` would force plugin authors to use C++17 or later.

@Yaron10
The numerical value is still same, so nothing changed, I just tested it ComparePlus plugin and it works.
It is now recommended to use [EnableThemeDialogTexture](https://learn.microsoft.com/en-us/windows/win32/api/uxtheme/nf-uxtheme-enablethemedialogtexture) directly instead of relying on Notepad++ to provide function.

Yaron10: @ozone10,

> The numerical value is still same, so nothing changed

Yes, I missed that. An amateur's mistake. :)

Thank you!

Issue #12737, opened 2022-12-31
[Dark Mode] Multiple-line tabs border


ozone10:
### Description of the Issue

Multiple-line tabs in dark mode with darken inactive tabs are hard to distinguish.
It's hard where one tab ends and another start between rows. 

![image](https://user-images.githubusercontent.com/55940305/210132987-d9524338-7bb9-4ce5-9317-27aef850a682.png)

### Steps to Reproduce the Issue

1. Enable dark mode
2. Enable Multi-line and Darken inactive tabs settings for tab bar
3. Have enough tabs to have multiple rows

### Expected Behavior

Some horizontal edge between rows.
### Actual Behavior

No horizontal edge between rows.
### Debug Information

Notepad++ v8.4.8

PR #12738, opened 2022-12-31
Align edge with tab (dark mode)

ozone10:
draw frame around tabs in multi-line state (dark mode), 
adjust padding for hiDPI,
add initializers

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12737
Before - After - Settings 
tabs


Yaron10:
@ozone10,

I don't use Dark Mode, but this is beautiful. 👍 

Happy New Year!

Issue #12720, opened 2022-12-28
Add support selection to TAB and Space conversion commands


ArkadiuszMichalski:
Would be nice to have selection support for these three commands: `TAB to Space`, `Space to TAB (All)` and `Space to TAB (Leading)`. Same as we have for trim operations https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12655 and maybe two others https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12711.

![image](https://user-images.githubusercontent.com/2730894/209812019-f3b588a3-50af-4f5b-bddb-394123f0b8e6.png)

@donho Here also is question how treat selection:
- exactly as it is
- or expand to full lines

If expand then for all three or only for some, like `Space to TAB (Leading)`?

Maybe in this submenu `Blank Operations` all commands should be treat in entire lines context, so partially selected line should be expand to full lines. It would be more consistent. Is there any command here for which a partially selected line would be better (more expected)?

ArkadiuszMichalski:
@donho 

Any advice on this? Only here the selection support is still missing. I would correct these commands too.

Yaron10:
@ArkadiuszMichalski,

In `Tab to Space` and `Space to Tab (All)` even column-selection can be useful in real-life. :)

In the following case the user might want to replace the selected spaces only.
![תמונה](https://user-images.githubusercontent.com/10229320/210111894-76623b20-9ddf-4650-b3fc-20efd7ff4b4f.png)

donho:
@ArkadiuszMichalski 
> Any advice on this? Only here the selection support is still missing. I would correct these commands too.

I would say that let's keep it as it is if there's no feature request.

alankilborn:
>In the following case the user might want to replace the selected spaces only.

No user I've ever met would want to do that.  :-)

donho:
I set suspended so far so if there's need we will unsuspended, ans see if we will do it.

ArkadiuszMichalski:
@donho
[...]

PR #12739, opened 2022-12-31
Add selection support to TAB and Space conversion commands

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12720, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12745.

Changes made related to the current state:
- add stream selection support. Partially selected lines are expanded to full lines without last EOL (same as `Trim`).
- block selection is not supported (just return).
- if nothing was change we restore original selection.
-  if nothing was change we don't replace original data (actually it was only for `Tab to Space`, now this applies to all commands). Replacing data with the same data unnecessarily changes the history.
- making some performance improvements by moving some things elsewhere (like obtaining bookmarks).

I didn't change code responsible for how Tab to Space conversion (and vice versa) works (add only one variable `changeDataCount` here). Changes were made only before this action and after this action.

Actually (before this PR), when we make a selection and execute commands, changes are made to the entire document, not just to the selection. In the case of a large file and a small selection, I don't know if the changes covered only the selection or the entire document. At first glance (if we don't see all the changed data), it seems that only for selection (but it's not true).

image

If selection has to be not supported at all then return should be done here (to not mislead us).

---
I have one additional question for the more experienced who can perform detailed debugging.

When we have some bookmark, make selection, copy data (`Ctrl+C`) and paste this data (`Ctrl+V`) we get this:

image
 
It seems Notepad (or Scintilla) incorrectly changing bookmark here. They should be deleted or keept, but we have some strange dislocation.

Now when we make undo we get this:

image

but this is not the initial state. Bookomark was on line 2 (not line 3).

This not  happens when we select all + copy and paste + undo:

image

This behavior also bothers me in this PR but I got around this for selection by deleting the bookmarks after remembering them. It's still not perfect because when we do undo such bookmark is restored. Well, but at the moment it must be like that because I do not know on whose side the problem is: Notepad++ or Scintilla?

Select => Tab to Space => Undo
image

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1350,7 +1350,8 @@ void Notepad_plus::wsTabConvert(spaceTab whichWay)
 		if (source == NULL)
 			continue;
 
-		_pEditView->execute(SCI_SETSEL, startPos, endPos);
+		_pEditView->execute(SCI_SETSELECTIONSTART, startPos);
+		_pEditView->execute(SCI_SETSELECTIONEND, endPos);
 		_pEditView->execute(SCI_GETSELTEXT, 0, reinterpret_cast<LPARAM>(source));
 
 		intptr_t count = 0;
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1526,7 +1526,7 @@ void Notepad_plus::wsTabConvert(spaceTab whichWay)
 
 		_pEditView->execute(SCI_BEGINUNDOACTION);
 
-		for (int i = static_cast<int>startLine; i <= endLine; ++i)
+		for (int i = static_cast<int>(startLine); i <= endLine; ++i)
 		{
 			if (bookmarkPresent(i))
 			{
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1387,7 +1387,7 @@ void Notepad_plus::wsTabConvert(spaceTab whichWay)
 					{
 						*dest++ = ' ';
 						changeDataCount += 1;
-						if (i <= currentPos)
+						if (i < currentPos)
 							++newCurrentPos;
 					}
 					column += insertTabs;
@@ -1395,7 +1395,7 @@ void Notepad_plus::wsTabConvert(spaceTab whichWay)
 				else
 				{
 					*dest++ = source[i];
-					if (i <= currentPos)
+					if (i < currentPos)
 						++newCurrentPos;
 					if ((source[i] == '\n') || (source[i] == '\r'))
 						column = 0;
@@ -1511,7 +1511,7 @@ void Notepad_plus::wsTabConvert(spaceTab whichWay)
 					}
 				}
 
-				if (i <= currentPos)
+				if (i < currentPos)
 					++newCurrentPos;
 			}
 			*dest = '\0';
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -1280,220 +1280,294 @@ bool Notepad_plus::replaceInOpenedFiles()
 	return true;
 }
 
-
 void Notepad_plus::wsTabConvert(spaceTab whichWay)
 {
+	// block selection is not supported
+	if ((_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_RECTANGLE) || (_pEditView->execute(SCI_GETSELECTIONMODE) == SC_SEL_THIN))
+		return;
+
 	intptr_t tabWidth = _pEditView->execute(SCI_GETTABWIDTH);
 	intptr_t currentPos = _pEditView->execute(SCI_GETCURRENTPOS);
-    intptr_t lastLine = _pEditView->lastZeroBasedLineNumber();
-	intptr_t docLength = _pEditView->execute(SCI_GETLENGTH) + 1;
-    if (docLength < 2)
-        return;
+	intptr_t startLine = 0;
+	intptr_t endLine = _pEditView->lastZeroBasedLineNumber();
+	intptr_t dataLength = _pEditView->execute(SCI_GETLENGTH) + 1;
+	intptr_t mainSelAnchor = _pEditView->execute(SCI_GETANCHOR);
+	bool isEntireDoc = (mainSelAnchor == currentPos);
 
-    intptr_t count = 0;
-    intptr_t column = 0;
-    intptr_t newCurrentPos = 0;
+	auto restoreSelection = [this, mainSelAnchor, currentPos, isEntireDoc]()
+	{
+		// restore original selection if nothing has changed
+		if (!isEntireDoc)
+		{
+			_pEditView->execute(SCI_SETANCHOR, mainSelAnchor);
+			_pEditView->execute(SCI_SETCURRENTPOS, currentPos);
+		}
+	};
+
+	// auto-expand of partially selected lines
+	if (!isEntireDoc)
+	{
+		intptr_t startPos = _pEditView->execute(SCI_GETSELECTIONSTART);
+		startLine = _pEditView->execute(SCI_LINEFROMPOSITION, startPos);
+		intptr_t endPos = _pEditView->execute(SCI_GETSELECTIONEND);
+		endLine = _pEditView->execute(SCI_LINEFROMPOSITION, endPos);
+
+		if (startPos != _pEditView->execute(SCI_POSITIONFROMLINE, startLine))
+			startPos = _pEditView->execute(SCI_POSITIONFROMLINE, startLine);
+
+		if (endPos != _pEditView->execute(SCI_POSITIONFROMLINE, endLine) && endPos < _pEditView->execute(SCI_GETLINEENDPOSITION, endLine))
+			endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, endLine);
+		
+		dataLength = endPos - startPos + 1;
+		_pEditView->execute(SCI_SETSEL, startPos, endPos);
+	}
+
+	if (dataLength < 2)
+	{
+		restoreSelection();
+		return;
+	}
+
+	char * source = new char[dataLength];
+	if (source == NULL)
+	{
+		restoreSelection();
+		return;
+	}
+
+	if (!isEntireDoc)
+		_pEditView->execute(SCI_GETSELTEXT, 0, reinterpret_cast<LPARAM>(source));
+	else
+		_pEditView->execute(SCI_GETTEXT, dataLength, reinterpret_cast<LPARAM>(source));
+
+	intptr_t count = 0;
+	intptr_t column = 0;
+	intptr_t newCurrentPos = 0;
 	intptr_t tabStop = tabWidth - 1;   // remember, counting from zero !
-    bool onlyLeading = false;
-    vector<int> bookmarks;
-    vector<int> folding;
+	bool onlyLeading = false;
 
-    for (int i=0; i<lastLine; ++i)
-    {
-        if (bookmarkPresent(i))
-            bookmarks.push_back(i);
+	if (whichWay == tab2Space)
+	{
+		// count how many tabs are there
+		for (const char * ch = source; *ch; ++ch)
+		{
+			if (*ch == '\t')
+				++count;
+		}
+		if (count == 0)
+		{
+			restoreSelection();
+			delete [] source;
+			return;
+		}
+	}
+	// allocate tabwidth-1 chars extra per tab, just to be safe
+	size_t newLen = dataLength + count * (tabWidth - 1) + 1;
+	char * destination = new char[newLen];
+	if (destination == NULL)
+	{
+		restoreSelection();
+		delete [] source;
+		return;
+	}
+	char * dest = destination;
+	intptr_t changeDataCount = 0;
 
-        if ((_pEditView->execute(SCI_GETFOLDLEVEL, i) & SC_FOLDLEVELHEADERFLAG))
-            if (_pEditView->execute(SCI_GETFOLDEXPANDED, i) == 0)
-                folding.push_back(i);
-    }
-
-    char * source = new char[docLength];
-    if (source == NULL)
-        return;
-    _pEditView->execute(SCI_GETTEXT, docLength, reinterpret_cast<LPARAM>(source));
-
-    if (whichWay == tab2Space)
-    {
-        // count how many tabs are there
-        for (const char * ch=source; *ch; ++ch)
-        {
-            if (*ch == '\t')
-                ++count;
-        }
-        if (count == 0)
-        {
-            delete [] source;
-            return;
-        }
-    }
-    // allocate tabwidth-1 chars extra per tab, just to be safe
-    size_t newlen = docLength + count * (tabWidth - 1) + 1;
-    char * destination = new char[newlen];
-    if (destination == NULL)
-    {
-        delete [] source;
-        return;
-    }
-    char * dest = destination;
-
-    switch (whichWay)
-    {
-        case tab2Space:
-        {
-            // rip through each line of the file
-            for (int i = 0; source[i] != '\0'; ++i)
-            {
-                if (source[i] == '\t')
-                {
-                    intptr_t insertTabs = tabWidth - (column % tabWidth);
-                    for (int j = 0; j < insertTabs; ++j)
-                    {
-                        *dest++ = ' ';
-                        if (i <= currentPos)
-                            ++newCurrentPos;
-                    }
-                    column += insertTabs;
-                }
-                else
-                {
-                    *dest++ = source[i];
-                    if (i <= currentPos)
-                        ++newCurrentPos;
-                    if ((source[i] == '\n') || (source[i] == '\r'))
-                        column = 0;
-                    else if ((source[i] & 0xC0) != 0x80)  // UTF_8 support: count only bytes that don't start with 10......
-                        ++column;
-                }
-            }
-            *dest = '\0';
-            break;
-        }
-        case space2TabLeading:
-        {
-            onlyLeading = true;
-        }
-        case space2TabAll:
-        {
-            bool nextChar = false;
+	switch (whichWay)
+	{
+		case tab2Space:
+		{
+			// rip through each line of the file
+			for (int i = 0; source[i] != '\0'; ++i)
+			{
+				if (source[i] == '\t')
+				{
+					intptr_t insertTabs = tabWidth - (column % tabWidth);
+					for (int j = 0; j < insertTabs; ++j)
+					{
+						*dest++ = ' ';
+						changeDataCount += 1;
+						if (i <= currentPos)
+							++newCurrentPos;
+					}
+					column += insertTabs;
+				}
+				else
+				{
+					*dest++ = source[i];
+					if (i <= currentPos)
+						++newCurrentPos;
+					if ((source[i] == '\n') || (source[i] == '\r'))
+						column = 0;
+					else if ((source[i] & 0xC0) != 0x80)  // UTF_8 support: count only bytes that don't start with 10......
+						++column;
+				}
+			}
+			*dest = '\0';
+			break;
+		}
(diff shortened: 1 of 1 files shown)


alankilborn:
>I have one additional question for the more experienced who can perform detailed debugging.
When we have some bookmark...

Bookmark handling has always been "strange" when lines are changed like this.

Here's an even simpler case than the one above; start with:

image

Press the Delete key, see:

image

IMO bookmark should have vanished.

Press Ctrl+z, bookmark reappears but in wrong place:

image

-------------------------

>I do not know on whose side the problem is: Notepad++ or Scintilla?

I don't believe Scintilla has anything to do with bookmarks, so the behavior is Notepad++'s.

--------------------------

This bookmark tangent probably deserves its own issue.  It should be ignored if possible while working on the main thrust of this current issue.

ArkadiuszMichalski:
@ozone10 

Can we somehow get the same behavior for GCC 64bit what we have form MSC 64bit?https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45808112

I mean stop compiling due such errors:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45808112/job/qpooi1lys830tvu1/messages

Yaron10:
@ArkadiuszMichalski,

I've just started to test this PR and encountered an interesting issue. *Reproducible with the current master as well.*

Paste the following text into a new document.
```
        Test         		Test      
Test         		Test      
        Test         		Test      
        Test         		Test      
        Test         		Test      
		
```
Place the caret at the end of line 1.
תמונה

`Space to Tab (All)`.

Result:
The caret disappears.
Not all spaces are converted.
תמונה

If you can not reproduce it, please restart NPP and try again.

Happy New Year!

Mateos81:
The unconverted spaces are the ones not making a full TAB length (2 spaces instead of 4 with the default setting)

donho:
@ArkadiuszMichalski 
> This behavior also bothers me in this PR but I got around this for selection by deleting the bookmarks after remembering them. It's still not perfect because when we do undo such bookmark is restored. Well, but at the moment it must be like that because I do not know on whose side the problem is: Notepad++ or Scintilla?

I don't think it's the right thing to do.
Though I cannot reproduce steadily, I've got several times that my markers disappear after the TAB<->Space operations by trying this PR.

ArkadiuszMichalski:
@donho
>I don't think it's the right thing to do.
Though I cannot reproduce steadily, I've got several times that my markers disappear after the TAB<->Space operations by trying this PR.

By markers do you mean bookmarks?

I would experiment more, but an exact STR would be useful. In the other site these bookmarks will be deleted anyway when the data is replaced, so I don't why they are not restored (we remembering them).

ArkadiuszMichalski:
@Yaron10 
>The caret disappears.

I think it was placed between `CR` and `LF`. Looks like after replacing the position of caret is incorrectly determined (for entire document).

image

Yaron10:
@Mateos81,

> The unconverted spaces are the ones not making a full TAB length (2 spaces instead of 4 with the default setting)

Thank you. 👍 

Using a mono-space font, the definition of `Space to Tab (All)` is much clearer: replace as many spaces as possible without changing the blanks width.
Please let me know if you can think of a better definition.

***

@ArkadiuszMichalski,

The "In Selection" part of this PR seems to work very well.

Regarding the bookmarks:
This issue deserves a separate discussion. I there a reason to mix it in this PR?

> The caret disappears.

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12745.

ArkadiuszMichalski:
@Yaron10 

I changed caret setting so you can try if it works better.

>Regarding the bookmarks:
>This issue deserves a separate discussion. Is there a reason to mix it in this PR?

@donho have some other problem with them, but I don't know exactly what it is without some STR.

Here we are trying to keep bookmarks after changes, so for selection it's worth doing the same. But if it doesn't work well we need to figure out what to do about it (not restore them at all?).

Yaron10:
@ArkadiuszMichalski,

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12745.

The caret is indeed between CR to LF.
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12739/commits/7f1ec9f9dd30387061f2dcdd3e437a5abbebdbf9 fixes the issue. 👍 
***
_Bookmarks:_

Seeing
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3da880b4f896254cc34daf42a37cb4da1b0c4df1/PowerEditor/src/Notepad_plus.cpp#L1298
I understand the relation to this PR.

I haven't tested it without the bookmarks code.
Could you list the exact cases the _original code_ should fix? And your code?

Yaron10:
I've played a bit with the bookmarks.

Paste the following text:
```
        Test         		Test
Test         		Test
        Test         		Test
        Test         		Test
        Test         		Test
		
        Test         		Test
		
```
Bookmark line 8.
Space to Tab (All).

**Result:**

Current master:
Bookmark is removed.

This PR:
Bookmark is kept.

ArkadiuszMichalski:
@Yaron10 
>Current master:
Bookmark is removed.
This PR:
Bookmark is kept.

Yes, and it should be kept. Current master is wrong here and I fixed this. 

@donho
>Though I cannot reproduce steadily, I've got several times that my markers disappear after the TAB<->Space operations by trying this PR.

I think at some point you make undo for all data and this action removes all bookmarks. But that has nothing to do with PR, that's just how Scintilla works.

STR:
1. Past this data:
```
	Test1
	Test2
	Test3
```
2. Invoke `Tab to Space` for entire document.
3. Add bookmark to line 1 and 2.
4. Make Undo.

We lost all bookmarks.

Same will happens with such scenerio:
 1. Past this data:
```
	Test1
	Test2
	Test3
```
2. Add bookmark to line 1 and 2.
3. Invoke `Tab to Space` for entire document (bookmarks are deleted here, but before this action we remember them and then restore them).
4. Make Undo.

We lost all bookmarks.

donho:
@ArkadiuszMichalski 

> I don't think it's the right thing to do.
> Though I cannot reproduce steadily, I've got several times that my markers disappear after the TAB<->Space operations by trying this PR.

I just checked the current code in master, the code memorizes the markers and uses SCI_SETTEXT then restores the markers - as I said, it's not the right thing to do. OTOH, this code lasted about 10 years, it might need to be rewritten.

Do you have other approach to replace SCI_SETTEXT (which makes disappear the markers) ?

ArkadiuszMichalski:
@donho 
>Do you have other approach to replace SCI_SETTEXT (which makes disappear the markers) ?

In the context of bookmarks seem they disappear when we replace the entire line (including last EOL). This does not apply only to `SCI_SETTEXT`, but to all actions.  Maybe for these methods here it would be possible to go through each line separately (without clearing EOL), just changing data before EOL. Well, but it will probably be much slower than operating on the whole data at once.

>I just checked the current code in master, the code memorizes the markers and uses SCI_SETTEXT then restores the markers - as I said, it's not the right thing to do. OTOH, this code lasted about 10 years, it might need to be rewritten.

What's wrong with this approach? If we want to keep the current bookmarks (while replacing whole lines) this is the only way out. We can combine as above suggestion, but then we may have a performance problem.

Yaron10:
The Sort commands do not restore bookmarks.

***

How about adding functions for the following operations? - The code is basically the same in all places. 
1. "auto-expand of partially selected lines".
2. "restore original selection if nothing has changed" (instead of `auto restoreSelection = [this`).
3. "block selection is not supported".

***
"block selection is not supported" can always be at the beginning of the function (`doTrim()` and `eol2ws()`).

ArkadiuszMichalski:
@donho
>For replacing TAB<->Space, do we need to reset EOL? For the Trim operations, bookmarks are untouched after the operation, aren't they?

Trim is simpler and we use just `_findReplaceDlg.processAll` and some regex, here all things are more complex. It is possible to do the replacement without touching the EOL, for example by doing:
- replace all data in line without EOL (maybe the current algorithm will cover it). So iterate over all lines, read data from single line, use current algorithm (or minimally modify it), replace data in single line without EOL. But here each line will be processed separately and it will be slow for sure. 
- same as above but operate on wole data at once, but without touching EOL  (here probably new code is needed).

This replacement works in some specific way (it's not strictly `fixed number of spaces to tab` or `tab to fixed number of spaces`)
image
image

I don't know if it's intentional or not, but it looks weird. Using simple regex and `_findReplaceDlg.processAll` probably won't be enough. All this conversion is a bit complicated and if a completely new approach is needed then it has to be done by someone else (right now it's not my skills).

Yaron10:
> This replacement works in some specific way (it's not strictly fixed number of spaces to tab or tab to fixed number of spaces)
...
I don't know if it's intentional or not, but it looks weird.

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12739#issuecomment-1368556471.
> Using a mono-space font, the definition of Space to Tab (All) is much clearer: replace as many spaces as possible without changing the blanks width.
Please let me know if you can think of a better definition.

In your first case:
You have 95px from the beginning of the line to the caret _before and after the conversion_.
You have 41px from the beginning of the line to the the first `k` _before and after the conversion_.

donho:
@ArkadiuszMichalski 
> I don't know if it's intentional or not, but it looks weird. Using simple regex and `_findReplaceDlg.processAll` probably won't be enough. All this conversion is a bit complicated and if a completely new approach is needed then it has to be done by someone else (right now it's not my skills).

I think it's the author (of this algorithm) was trying to do:

image

> replace all data in line without EOL (maybe the current algorithm will cover it). So iterate over all lines, read data from single line, use current algorithm (or minimally modify it), replace data in single line without EOL. But here each line will be processed separately and it will be slow for sure.

Sounds good - never mind if it gets slower, that's the price to pay.

ArkadiuszMichalski:
@donho 

The same problem is for all sorting command (they use common function) or for `Remove Duplicate Lines`:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c0e207328057b9ce321587a053ff1476cbb19562/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L3751

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/13dfaa96a8b224b000d470184ecae7a05057be45/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L4055

Maybe in other places too where we replace data including EOLs.

> Sounds good - never mind if it gets slower, that's the price to pay.

But you remember how much performance drops if all the lines have to be changed, and it's done separately (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12512)? I guess it will be similar here. Anyway I'll see if I can do it. At least will be able to compare the speed of both approaches.

ArkadiuszMichalski:
Another version for testing appeared. This time we modify the lines separately (without touching their EOL). I didn't check the performance case, I will do it later.

ArkadiuszMichalski:
Well, there is a difference in performance, we can see it immediately for this test file [test_space_20k.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10344531/test_space_20k.txt) . Such example is a bit extreme (all 20k lines must be changed), but in practice we can deal with different cases (including these).

donho:
@ArkadiuszMichalski 
So you are rather for the version in which the data is processed only once I guess?

ArkadiuszMichalski:
Yes, read data (all or from selection) to `source` as once and iterate over it and makes change (but still don't touch EOL's). Now I create `source` for each lines. Maybe it will help. I'll see if I can do that.

ArkadiuszMichalski:
@donho
Looks like this line is not optimal because `The caret is scrolled into view after this operation.` Doing this for every line is not best idea.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/4f9e488c7e76d69430df194d4de2af24c4b1e157/PowerEditor/src/Notepad_plus.cpp#L1353

If I change it to:
```
_pEditView->execute(SCI_SETSELECTIONSTART, startPos);
_pEditView->execute(SCI_SETSELECTIONEND, endPos);
```
it's better. Still I create `source` for each lines but it may not be necessary to change it.

I would prefer to operate directly on the target here. So instead of:
```
		_pEditView->execute(SCI_SETSELECTIONSTART, startPos);
		_pEditView->execute(SCI_SETSELECTIONEND, endPos);
		_pEditView->execute(SCI_GETSELTEXT, 0, reinterpret_cast<LPARAM>(source));
...
		if (changeDataLineCount)
		{
			_pEditView->execute(SCI_TARGETFROMSELECTION);
			_pEditView->execute(SCI_REPLACETARGET, static_cast<WPARAM>(-1), reinterpret_cast<LPARAM>(destination));
		}

```
use:

```
		_pEditView->execute(SCI_SETTARGETRANGE, startPos, endPos);
		_pEditView->execute(SCI_GETTARGETTEXT, 0, reinterpret_cast<LPARAM>(source));
...
		if (changeDataLineCount)
		{
			_pEditView->execute(SCI_REPLACETARGET, static_cast<WPARAM>(-1), reinterpret_cast<LPARAM>(destination));
		}

```

but on this second case sometimes read/replace data is wrong, and I don't know why. This is visible if right after starting Notepad++ we invoke `Tab to Space` on the whole document. If there is no crash and we undo + `Tab to Space` a few times, the problem disappears.

Any ideas what's going on here? The problem occurs when reading data or replacing it?

donho:
> Any ideas what's going on here? The problem occurs when reading data or replacing it? What's differences on these `SCI_SETTARGETRANGE/SCI_GETTARGETTEXT` vs `SCI_SETSELECTIONSTART/SCI_SETSELECTIONEND/SCI_GETSELTEXT`?

They should do the same effect. My guess is the absolute position startPos & endPos could be no more consistent AFTER several previous changes in the 2nd code case. 
Does the code which causes the problem has been committed in this PR?

ArkadiuszMichalski:
>They should do the same effect.

I thought the same but I get wrong result for `SCI_SETTARGETRANGE/SCI_GETTARGETTEXT`.

>My guess is the absolute position startPos & endPos could be no more consistent AFTER several previous changes in the 2nd code case.

But the number of lines does not change. And `startPos & endPos` for specific line we obtain before we change any data in this line (we should have the correct values after passing the line number). I see no reason why `startPos & endPos` could be wrong for target operation but not for selection operation, in both cases we retrieve these values in the same way.
```
		intptr_t startPos = _pEditView->execute(SCI_POSITIONFROMLINE, idx);
		intptr_t endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, idx);
```
It would be useful to know what the problem is here (even for future use). 

>Does the code which causes the problem has been committed in this PR?

No, last commit use `SCI_SETSELECTIONSTART/SCI_SETSELECTIONEND/SCI_GETSELTEXT` variant, and it works fine. I have never had wrong data. Well, it would be good if someone else tested it more precisely to be sure.

You can test last commit if this approach is better (related how bookmarks are handling). Here is some big file for performance testing https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12739#issuecomment-1370942447.

donho:
[notepad++.byProdMachine.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10356129/notepad%2B%2B.byProdMachine.zip)

donho:
[notepad++.byProdMachine.x86.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10356164/notepad%2B%2B.byProdMachine.x86.zip)

donho:
@ArkadiuszMichalski 

> So firstly we check if our data (entire document or selection) has any bookmarks. In no we can use faster approach, if yes we use slower approach.

This could be a solution, but it's surly not the best one for me (due to the complicity of code).

I have check this PR in 3 different PC: **A**. a super Gamer PC, **B**. a good decent 2 years laptop  & **C**. a 10 years old laptop, comparing with official release v8.4.8, by using [test_space_20k.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10344531/test_space_20k.txt). Here are the result:

|   |   | v8.4.8  |   |  PR |   |
|---|---|---|---|---|---|
|  A |   |  less 1 sec |   | less 1 sec  |   |
|  B |   | less 1 sec  |   | 1 sec  |   |
|  C |   |  1 sec |   |  4 sec |   |


For A & B category of PC, the results are acceptable, however, the result of C is not.
I would say it's 10 years old laptop, but still.

Obviously the old algorithm (no line by line loop) is the most optimal. And the workaround for memorizing and restoring bookmarkers is a better compromise finally. Thank you for your efforts for testing and providing the different solution, and sorry for make you roll back this PR to your initial modification (could you?).

ArkadiuszMichalski:
@donho 

But the same is for TRIM commands (now we use one by one line replace approach), so for more and more lines it will get slower. We could also do one replace + restore bookmarks here (of course in a different PR).

You can observe it, take this file [test_space_20k.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10344531/test_space_20k.txt) and few times copy/paste all data to get more data (like 100k lines) and experiment with different commands. On slower machine it will simply be worse.

The worst part here is that in most time we probably will operate on small files or data (small selection) so the latter approach would be more useful here. In my case such an optimized versions of some commands would be useful very rarely. Well, everyone works differently.

> Obviously the old algorithm (no line by line loop) is the most optimal. 

The iteration itself and reading the data is not a problem here. Separately overwriting data for each line causes such a performance dropping. I even try with turned off Undo memorization, but it didn't help. One-time overwrite of the entire data beats everything.

Regardless of the chosen solution in my free time I will make a plugin with the opposite behavior. So:
- if we have faster version in Notepad++ then I make slower version (with better bookmarks support) in plugin
- if we have slower version (with better bookmarks support) in Notepad++ then I make  faster version in plugin

>Thank you for your efforts for testing and providing the different solution, and sorry for make you roll back this PR to your initial modification (could you?).

OK, I'll restore the first version and see if there's anything else that needs to be corrected.

ArkadiuszMichalski:
@donho 

In most time we use `intptr_t` or `size_t`. But there are still places where `int` is used (derived from the original code):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/838bd58af1122947b87b7ec89a16aac808c4e646/PowerEditor/src/Notepad_plus.cpp#L1381
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/838bd58af1122947b87b7ec89a16aac808c4e646/PowerEditor/src/Notepad_plus.cpp#L1386
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/838bd58af1122947b87b7ec89a16aac808c4e646/PowerEditor/src/Notepad_plus.cpp#L1418

If for 64-bit and large files we need to change it to something bigger, let me know to what.

donho:
@ArkadiuszMichalski 
> If for 64-bit and large files we need to change it to something bigger, let me know to what.

It's no harm to make them 64-bit indeed...
After a very quick check for the type, it seems no cast needed to be done, so OK for me.
But in case contrary, just keep them as it.   

> This version is really fast, even for 1 million lines.

👍

ArkadiuszMichalski:
@donho
Replacing everything even if some line don't need it (because it doesn't change) has two more disadvantages:
- remove every markers in such line
- in the history such a line is shown as changed, so it's hard to find the lines that really have changed

`Tab to Space`:
image

By doing this line by line more stuff is preserved. This is how `Trim` currently works:
image

donho:
> Replacing everything even if some line don't need it (because it doesn't change) has more disadvantages:

Hmm... that's a good argument since now we've got change history.

What's your opinion @Yaron10? @alankilborn? anyone?

Yaron10:
@donho,

We all agree that the arguments for either approach are good and solid.
For my personal use, like @ArkadiuszMichalski's, the performance factor *in our case* is negligible.

From a broader Notepad++ point of view: I do not know.
However, I think we can also agree that the chosen approach should be consistent to all commands: "Line Operations", "Blank Operations" and possibly more.
And this (properly unifying all commands) might be another factor in your decision.

donho:
@ArkadiuszMichalski 
> Replacing everything even if some line don't need it (because it doesn't change) has more disadvantages:

I think we can try at least **line by line** algorithm. If there are some users complain about the performance, we will switch to the faster version. 

FYI, I have included this PR except the last commit. In case of complain of users,  we can always include the last commit [Switch to faster version](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12739/commits/838bd58af1122947b87b7ec89a16aac808c4e646).

Issue #12740, opened 2022-12-31
Let GCC behaves similar to MSC for implicit conversions that may alter a value


ArkadiuszMichalski:
Now MSC for implicit conversions that may alter a value throws an error (aborts compilation) for both: Release and Debug.
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/45808112

Unfortunately, GCC does not respect this (doesn't even warn about it). It looks like at least [`-Wconversion`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wconversion) should be add to:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/307fd2fcd2d4dd7a626092e6e1dc1a0467b596f5/PowerEditor/gcc/makefile#L40
so would be easier to catch such problem before we push on GitHub. It's a waste of time when MSC won't compile with these warnings anyway.

By using [`-Werror=`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Werror_003d) we can even turn such warning into an error so GCC will break compilation too.

donho:
@ArkadiuszMichalski 
If I understand well, you'ree using only GCC as your compiler?

ArkadiuszMichalski:
@donho 
Yes, and sometimes it happens to me that GCC doesn't give me any warning when MSC throws an error. I send a PR, block the queue and waste time unnecessarily. So from now on, I'll be adding these warnings flags to GCC, where in MSC we get errors.

I won't treat these warnings as errors in GCC because we can't get identical behavior in both compilers, and I don't want that GCC will block MSC compilation. Warnings in this case are enough and if they are sometimes more restrictive, they will not affect the compilation of MSC.

PR #12747, opened 2023-01-02
Add -Wconversion flag to makefile

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12740.

Few times I push PR (compiling in GCC without any problems) but it breaks on MSC so I waste of time. This is because MSC show warnings related to conversion problem and additionally treats them as an error. At the moment GCC ignores such thing so I added a proper flag that will support it.

*Anything that causes an error in MSC should be treated by GCC at least as a warning.

I din't change that warnings should be treated as errors. because I don't want GCC to be more restrictive than MSC and block the build (show warnings should be enough).

This change cause that for GCC and actual code we will see new warnings in log, but they can be corrected later.

### DIFF
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -53,13 +53,13 @@ SUBMAKEFLAGS := -O --no-print-directory
 ifeq "$(filter-out 0,$(DEBUG))" ""
 BUILD_TYPE := release
 BUILD_SUFFIX :=
-CXXFLAGS += -Os
+CXXFLAGS += -Os -Wconversion
 CPP_DEFINE += NDEBUG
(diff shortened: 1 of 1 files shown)


ozone10:
@ArkadiuszMichalski 
since GCC builds are mainly for code analysis,
how about adding `-Wextra` to debug builds, also both Scintilla and Lexilla are using it.

ArkadiuszMichalski:
@ozone10 

I'll see if `-Wextra` doesn't generate a bunch of new warnings for the entire build. If not then probably can be add. But if a lot of new warnings appear then analysis could be more difficult (at least until someone reduce all/some of them).

PR #12753, opened 2023-01-03
Update.xml

spkprs:
Sort and add few translates


donho:

Issue #12007, opened 2022-08-11
Plugin rendered wrong since v8.4.3


ffes:
### Description of the Issue

Since the update in v8.4.3 the feature `Add dark mode support for plugins` was added, But that doesn't work for my NppSnippets plugin in light mode.

### Steps to Reproduce the Issue

1. Install the NppSnippets plugin
2. Make sure you are in light mode
3. Select the "ANSI characters" library from the list (because that has a long list of snippets)
4. Scroll through the list and notice that the list is not rendered as expected.
5. Also notice that the combobox (with the list of libraries) isn't rendered correctly as well.

### Expected Behavior
The rendering should work, just like before v8.4.3.

### Remarks

See https://github.com/ffes/nppsnippets/issues/45 for more details and screenshots.

Note that I could not find any docs that plugin devs need to do anything to support this feature.

rdipardo:
It's been [independently reported][0] that:

> Since Notepad++ 8.4.4 enhanced Dark Mode [*sic*], the plugins’ docked dialogs seem to stop receiving the WM_CTLCOLOREDIT message.

Also, the size of the `NppDarkMode::Colors` structure was enlarged in d65967de (release version 8.4.2). Unannounced ABI changes like that can [affect plugins that don't account for them][1].


[0]: https://community.notepad-plus-plus.org/topic/23018/question-about-the-dark-mode-s-dark-scrollbars-and-toolbars/6
[1]: https://community.notepad-plus-plus.org/topic/23080/impact-analysis-of-generic-darkmode-rendering-of-plugin-dock-panels-from-pr-11711/7

ozone10:
I am currently away, so I cannot investigate it properly.
I can guess that plugin may use `WM_CTLCOLORLISTBOX` message which interfere with Notepad++ generic dark mode implementation.

@ffes 
currently for workaround you can update Notepad++ header files and use flag
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2e66fe000765d7923f731ef8f1e524a27ddc0912/PowerEditor/src/WinControls/DockingWnd/Docking.h#L39
to disable generic dark mode for plugin.

Or temporarily don't use `WM_CTLCOLORLISTBOX` message until there is workaround/fix for `WM_CTLCOLOR*` messages for Notepad++. There might be other code which might interfere with generic dark mode.

So IMHO I think using `DWS_USEOWNDARKMODE` flag might be easiest solution.

@rdipardo 
[...]

PR #12754, opened 2023-01-03
Fix docking GUI for plugin in light mode

ozone10:
improve combobox "v" position in hiDPI

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12007

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -1742,12 +1742,25 @@ namespace NppDarkMode
 
 				auto holdBrush = ::SelectObject(hdc, NppDarkMode::getDarkerBackgroundBrush());
 
-				auto dpiManager = NppParameters::getInstance()._dpiManager;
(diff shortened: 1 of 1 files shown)


donho:
@ozone10 
The enhancement is only for plugin panels' combobox or also for Notepad++ internal combox control?

ozone10:
@donho 
plugin with panel sometimes has visual glitch in hiDPI (I can't consistently reproduci it) but this improvement is also applied for every Notepad++ comboboxes in dark mode. 

It uses system provided rect instead of manually calculated positions to create rect.

ffes:
In the original issue #12007 is a description. You can also see https://github.com/ffes/nppsnippets/issues/45 for more details.

It is easy reproducible: Install the NppSnippets plugin v1.7.0 from https://github.com/ffes/nppsnippets/releases/tag/v1.7.0, be sure to be in light mode and scroll through the list of snippets. Don't forget to open the combo on top of the list.

Be sure not to use v1.7.1 because that has `DWS_USEOWNDARKMODE` to work around the bug.

Issue #12555, opened 2022-11-22
Deleting search results results in app lock-up


scott-smith-immersedvr:
### Description of the Issue
Deleting search results results in app lock-up ("Not Responding").

I've noticed this happening when I do searches that result in lots of hits over many files -- hereafter referred to as a "_big search_". So far, I've only noticed the lock-up when I try to clean up the results view after one or more big searches by removing other (usually non-big) search results. It's not necessary to attempt remove the big search itself.

I've first  noticed this problem not long ago, so I suspect that it was introduced relatively recently. I'm using Notepad++ in the same way I have been for years, and this problem just started occurring in the last few months or so.

### Steps to Reproduce the Issue
1. Do two or more **Find in Files** searches, one of which results in lots of hits over lots of files (e.g. "22539 hits in 1983 files of 13138 searched"). The other search(es) can have few or zero hits.
2. Try to delete one of the other (non-big) search results by clicking on its parent item in the search results view.
3. Even though the search being deleted has a small number of hits, the app enters a "Not Responding" state and never recovers.

### Expected Behavior
I might expect that deleting the results of a really big search could take some time. But deleting search results with a small number of hits shouldn't take long, just because the view also has one or more big search results in it. In fact, it's not clear why the mere presence of a large search result would impact the deletion of a small or zero-result search that's either earlier or later in the Search results view.

### Actual Behavior
The app enters a "Not Responding" state and never recovers (I've let it sit for hours on at least one occasion).

### Debug Information
Notepad++ v8.4.6   (32-bit)
Build time : Sep 25 2022 - 19:55:26
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.2251
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

![image](https://user-images.githubusercontent.com/110685010/203392414-8dfbc742-22c6-4d6f-8c39-e69d8bfdaaac.png)

Rinakles:
Same issue here. Started a few versions ago, been waiting for it to be fixed but no changes so far.
It doesn't actually lock up, the time to close the results has simply increased by something like 10000x.

[...]

PR #12765, opened 2023-01-06
When deleting search results use setLexer only when lexer is SCLEX_NULL

ozone10:
improve/partially fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12555

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -610,7 +610,10 @@ void Finder::deleteResult()
 	auto end = _scintView.execute(SCI_GETLINEENDPOSITION, lno);
 	if (start + 2 >= end) return; // avoid empty lines
 
-	_scintView.setLexer(L_SEARCHRESULT, LIST_NONE); // Restore searchResult lexer in case the lexer was changed to SCLEX_NULL in GotoFoundLine()
+	if (_scintView.execute(SCI_GETLEXER) == SCLEX_NULL)
+	{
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -610,7 +610,10 @@ void Finder::deleteResult()
 	auto end = _scintView.execute(SCI_GETLINEENDPOSITION, lno);
 	if (start + 2 >= end) return; // avoid empty lines
 
-	_scintView.setLexer(L_SEARCHRESULT, LIST_NONE); // Restore searchResult lexer in case the lexer was changed to SCLEX_NULL in GotoFoundLine()
+	if (_scintView.execute(SCI_GETLEXER) == SCLEX_NULL)
+	{
+		_scintView.setLexer(L_SEARCHRESULT, LIST_NONE); // Restore searchResult lexer in case the lexer was changed to SCLEX_NULL in GotoFoundLine()
```
ozone10, line 615: maybe `_scintView.makeStyle(L_SEARCHRESULT, nullptr);` is enough

ozone10:
still the main issue is with folding performance

donho:
@ozone10 
Good job!
But how come it fixes "partially" #12555 ?

ozone10:
@donho 
you can close it as it deal with the main issue.

But since upgrading to scintilla 5, usually "first" folding is really slow, and it can have impact on deleting lines.
I've also tried to commenting out some code related to it in `ScintillaEditView.cpp` (IIRC `runMarkers`) and it helps.
But I am not sure how to handle it, so I only fixed issue with `setLexer`.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b3934afd9210f7f580355c7eb54f263d425ebd43

Issue #12772, opened 2023-01-07
[DPI] Cursor for link text


ozone10:
### Description of the Issue

In high DPI display cursor for link is too small.

### Steps to Reproduce the Issue

1.  Have display with high DPI
2. Open window with link text e.g. Debug info
3. Hover above link text e.g. "Copy debug info into clipboard"

### Expected Behavior

Cursor is not small.

### Actual Behavior

Cursor is small.

### Debug Information

Notepad++ v8.4.8

PR #12773, opened 2023-01-07
Use predefined cursor (hand) for link text

ozone10:
before - after
cursor

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12772


PowerEditor/src/WinControls/AboutDlg/URLCtrl.h
```diff
@@ -26,17 +26,19 @@ class URLCtrl : public Window {
 	void create(HWND itemHandle, int cmd, HWND msgDest = NULL);
     void destroy();
 private:
+	HCURSOR& getCursor();
 	void action();
 protected :
     generic_string _URL;
     HFONT _hfUnderlined = nullptr;
     HCURSOR _hCursor = nullptr;
+	HCURSOR _hCursorPredefined = nullptr;
```
donho, line 35: Why don't we use system's one (ie. `::LoadCursor(nullptr, IDC_HAND);`) and remove XORMask & ANDMask ?

ozone10: Ok

PowerEditor/visual.net/notepadPlus.vcxproj
```diff
@@ -130,7 +130,7 @@
...
-      <PreprocessorDefinitions>WIN32;_WIN32_WINNT=0x0600;_WINDOWS;_USE_64BIT_TIME_T;TIXML_USE_STL;TIXMLA_USE_STL;_CRT_NONSTDC_NO_DEPRECATE;_CRT_SECURE_NO_WARNINGS;_CRT_NON_CONFORMING_SWPRINTFS=1;_CRT_NON_CONFORMING_WCSTOK;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>OEMRESOURCE;WIN32;_WIN32_WINNT=0x0600;_WINDOWS;_USE_64BIT_TIME_T;TIXML_USE_STL;TIXMLA_USE_STL;_CRT_NONSTDC_NO_DEPRECATE;_CRT_SECURE_NO_WARNINGS;_CRT_NON_CONFORMING_SWPRINTFS=1;_CRT_NON_CONFORMING_WCSTOK;_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
ozone10, line 133: @donho 
can we move Preprocessor Definitions outside to property sheet, so we don't have to add define to each builds

donho: @ozone10 
Go ahead please. If you think it's better, it's OK to me.

ozone10: @donho 
ok I will prepare it in another PR

PR #12779, opened 2023-01-09
Fix for Windows Notepad replacement regression (Notepad++ v8.4.8)

### LINKED ISSUE #12775: Notepad replacement regression
### Description of the Issue
Having replaced stock notepad with notepad++ in registry according to the  [docs](https://npp-user-manual.org/docs/other-resources/), selected formats open in notepad++ v8.4.8 but now additionally display a `Cannot open file " ".` (whitespace within doublequotes) error.

This can be observed by doubleclicking `.ps1` powershell files, or opening registry `.reg` files with the Edit action in the file context menu. Other file formats defaulting to Edit on open may be affected as well. Standard `.txt` files, as well as files potentially associated with (windows) notepad by the user (for example `.cfg`, `.json` , `.log`, `.xml` ...) are unaffected.

1. Replace notepad with notepad++ in registry according to the documentation
2. Create a dummy test.ps1 or test.reg with any content
3. Doubleclick test.ps1, or rightclick->Edit test.reg

### Expected Behavior
File opens (notepad++ <=v8.4.7)

### Actual Behavior
File opens, but a `Cannot open file " ".` message box is displayed first (notepad++ v8.4.8)

msgbox

### Debug Information
Notepad++ v8.4.8   (64-bit)
Path : C:\npp\notepad++.exe
Command Line : -notepadStyleCmdline -z "C:\Windows\System32\notepad.exe" "C:\test\test.ps1"
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.2364
Current ANSI codepage : 1250
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

xomx:
Fixes #12775 .

This fixes opening of unintentional non-existing file named " " in Notepad++ v8.4.8.

The bug was there all the time, but it was invisible for users until v8.4.8 checked for such nonsense filenames.
More details in the issue.

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -165,10 +165,10 @@ ParamVector convertParamsToNotepadStyle(PWSTR pCmdLine)
 		size_t len = wcslen(pCmdLine);
 		if (len > 0)
 		{
-			if (pCmdLine[0] == L'\"')
-				++pCmdLine;
 			if (pCmdLine[len - 1] == L'\"')
 				pCmdLine[len - 1] = L'\0';
+			if (pCmdLine[0] == L'\"')
+				++pCmdLine;
 		}
 
 		generic_string str(pCmdLine);
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -159,10 +159,20 @@ ParamVector convertParamsToNotepadStyle(PWSTR pCmdLine)
 	}
 
 	// Now form a file name from the remaining commandline (if any is left)
-	if ( *pCmdLine != '\0' )
+	if ( *pCmdLine != L'\0' )
 	{
+		// strip possible surrounding quotes
+		size_t len = wcslen(pCmdLine);
+		if (len > 0)
+		{
+			if (pCmdLine[0] == L'\"')
+				++pCmdLine;
+			if (pCmdLine[len - 1] == L'\"')
+				pCmdLine[len - 1] = L'\0';
+		}
+
 		generic_string str(pCmdLine);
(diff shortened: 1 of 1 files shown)


donho:
@xomx 
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12775#issuecomment-1375983014

donho (changes requested):
When the last argument is around the quotes as the following example:
-notepadStyleCmdline -z "C:\Windows\system32\NOTEPAD.EXE" `"`C:\Users\xxxxxx\xxxxxxxx\Nouveau document texte.txt`"`
It's parsed badly in `advanceCmdLine` function and make the inconsistent output.
Therefore the bug should be fixed in `advanceCmdLine` function, but not somewhere-else.

donho:
@xomx 
FYI, I'm refactoring (for making it simpler) and after the refactoring, this bug should be fixed.

xomx:
@donho 
OK, thanks a lot, so can I close this PR?

I have to deal with something serious IRL, it looks like I won't be able to get involved here for some time.

donho:
@xomx 
> OK, thanks a lot, so can I close this PR?

Yes, go ahead.
If you have time, please check  #12800 - it should fixes the problem.

> I have to deal with something serious IRL, it looks like I won't be able to get involved here for some time.

Hope everything will go into order to you.

Thank you very much for your analyze - that helped me a lot.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6c3659fe7300992ae7a3cd32410377bd88ac263a

xomx:
@donho 

> If you have time, please check #12800 - it should fixes the problem.

I checked now - it is ok, good job!

Issue #11147, opened 2022-02-07
Path Completion not working regression


donho:
### Description of the Issue
Path Completion has stopped working seems for quite moment.
It's a regression but it's not due to the removal 2GB restriction. 

### Steps to Reproduce the Issue

1. Type `C:\`
2. Choose menu command `Edit > Auto-Completion > Path Completion` or hit `Ctrl-Alt-Space`  

### Expected Behavior
Auto-completion list appears. All the sub-folder of C drive should be in the list.
  
### Actual Behavior
Nothing happens.

### Debug Information
Notepad++ v8.3   (64-bit)
Build time : Feb  5 2022 - 19:07:43
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 2009
OS Build : 22000.434
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

alankilborn:
I went on a version search to where this feature last worked.
I had to go back to v 6.9 !

mpheath:
@donho 

Testing autocompletion and path completion is one of tests. I repeatly get these 3 msgboxes using the menu entry

```
---------------------------
General Exception
---------------------------
std::bad_alloc
---------------------------
OK   
---------------------------


---------------------------
Recovery initiating
---------------------------
Notepad++ will attempt to save any unsaved data. However, dataloss is very likely.
---------------------------
OK   
---------------------------


---------------------------
Recovery success
---------------------------
Notepad++ was able to successfully recover some unsaved documents, or nothing to be saved could be found.

You can find the results at :

C:\Users\Michael\AppData\Local\Temp\\N++RECOV
---------------------------
OK   
---------------------------
```

and then Notepad++ exits.

With Ctrl+Alt+Space I get

```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.

---------------------------
OK   
---------------------------
```

Click OK and crashs Notepad++.

Strange the different behaviours of menu item to keys. Probably irrelevent if fixed. 

<hr>

Added some `printf` calls to get some information.

Output:

```
bufSize: '260'
currentPos: '41'
startPos: '-219'
```

[...]

PR #12784, opened 2023-01-09
Replace macro min/max with std::min/std::max

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11147#issuecomment-1365724574

this also fix some -Wconversion warnings for gcc builds


ozone10:
macro min/max should be disabled with preprocessor macro `NOMINMAX` to enforce using std::min/std::max instead

Issue #12790, opened 2023-01-11
[Code] Add default settings for vs project builds


ozone10:
### Description of the Issue

When changing msvc builds settings for compiler/linker, e.g. preprocessor macros, it has to be set for 6 builds.
This can lead into issue with accidentally forgetting to modify one or more build(s).

### Debug Information

Notepad++ v8.4.8

PR #12791, opened 2023-01-11
Add property sheet, clean project file

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12790

additional changes:
  * use constant instead of magic number for `_WIN32_WINNT=_WIN32_WINNT_VISTA`
  * replace `<AdditionalOptions>/fixed:no %(AdditionalOptions)</AdditionalOptions>` with `<FixedBaseAddress>false</FixedBaseAddress>`
  * use `_CRT_NON_CONFORMING_SWPRINTFS` instead of `_CRT_NON_CONFORMING_SWPRINTFS=1`
  * replace "MinimumRequiredVersion" 5.01/5.02 with 6.00 (Vista), for ARM64 6.02 (Windows 8)
  * add missing settings to ARM64 builds
  * remove `copy ..\src\functionList.xml ..\binarm64\functionList.xml`
  
some consideration:
  * maybe enable [/CETCOMPAT](https://learn.microsoft.com/en-us/cpp/build/reference/cetcompat?view=msvc-170) for x64 and win32 builds, since it's enabled for scintilla and lexilla, I can't test it since I don't have 11th or newer Intel processor or equivalent AMD processor
  * removing `_CRT_NON_CONFORMING_SWPRINTFS` and fix [swprintf](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-sprintf-l-swprintf-swprintf-l-swprintf-l?view=msvc-170), because it may be removed in future. Fix is simple.


ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12773#discussion_r1064351459

PR #12798, opened 2023-01-13
[xml] Update chineseSimplified.xml

skadomsky:
Retranslated "Function Completion" and "Word Completion".


skadomsky:
Make the translation shorter and more appropriate.

Issue #12792, opened 2023-01-11
[FunctionList] Not found functions with comment in JavaScript files


fursformula:
In notepad v8.4.8 functionList (javascript.js.xml) not found functions with comment in JavaScript files.

For example:
createCanvas = function(value) // value = "image"
{
 ...
}

This function don't add in functionList. If I delete '// value = "image"' in the end of string, it works correctly.
Need modify file 'javascript.js.xml' to fix this error. Thanks!

ArkadiuszMichalski:
Duplicate of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3685.

PR #12694, opened 2022-12-23
[xml] Prevent duplicate entries in function list panel of Pascal/Delphi files

dinkumoil:
Fixes #12693, #12687
Also closes #12686

Issue #3685, opened 2017-08-31
Function list does not list functions containing comment after declaration


bladel:
The function outline view does not show a function, that contains a comment in the same line as the function declaration.

Example C:

```
void somefunction(void) //some comment
{
}
```

Is not listed in the function list.

MAPJe71:
Duplicate/related #1232, #1657, #1861, #2910, #3095 and Community Topics [13286](https://notepad-plus-plus.org/community/topic/13286/functionlist-doesn-t-recognise-a-c-function-with-comment) and [13468](https://notepad-plus-plus.org/community/topic/13468/function-list-issue).

indrajitmajumdar:
@bladel @MAPJe71 
### Same in javascript also.
`function somename () { //
        //
}`
This function will not be shown in the Function List.

MAPJe71:
@indrajitmajumdar not the same problem i.e. comment embedded in function header.
You're probably using an older parser as your sample function appears in my FunctionList (using Notepad++ 7.6.3).

pryrt:
**bump** Two years later, this issue is still not resolved.

Aside from the list of related topics that @MAPJe71 posted, it's also come up again recently in Community topic [20452](https://community.notepad-plus-plus.org/topic/20452/).  The [Community FAQ](https://community.notepad-plus-plus.org/topic/19480/faq-desk-function-list-basics/2) even has to have this problem listed as a known issue.

And at least in v7.9.1, it requires _two_ whitespace characters (combination of at least two spaces, tabs, CR, LF) before the comment, which contradicts @MAPJe71's experience back with 7.6.3 that seemed to work with a single space before the comment.
![](https://community.notepad-plus-plus.org/assets/uploads/files/1607613557924-2af33f3d-a5c9-401f-bb58-c3555943085a-image.png)

As I commented in the Topic I just linked, telling people "you have to change your coding style if you want the function list to work right" is not a very user-friendly thing to say.  Especially if their coding style is mandated by their employer or by the lead of whatever open-source project they are contributing to.  

I can see justifications for why the parser for Function List would be easier to write if there's at least one space separating the function list regex from the comment regex.  But just because it's easier doesn't mean it's not solvable.  And these issues have been going back at least 5 years to #1232.

Is this something that @sasumner or @donho could spend some time investigating?  Can it be fixed so that it doesn't require two spaces before the comment, or require a newline or space at the end of the function before EOF?

pryrt:
[...]

PR #12799, opened 2023-01-13
Fix exclude functions in comment zones

### LINKED ISSUE #12759: Under certain circumstances function list parsers produce duplicate or unwanted entries in function list panel
### Description of the Issue
The C++ code that processes function list parser files and fills the tree in function list panel produces under certain circumstances unwanted  or duplicate entries in that tree.

It seems that this C++ code creates a data model of a source code file by inspecting it with the help of the content of a function list parser's file. The process of creating that data model and/or filling the tree in function list panel (based on that data model) seems to be faulty and leads under certain circumstances to wrong results.

In the following I give 3 examples of wrong results I found so far.

### Prerequisites
1. Download a fresh portable copy of _Notepad++_ from official web site.
2. Save the following _Pascal_ source code as  `FnListTest1.pas`:

```Pascal
unit FnListTest1;


interface

uses
  System.SysUtils, System.Types, System.Classes;


// function FreeFunc1(const Param: integer): integer;



implementation

{
function FreeFunc1(const Param: integer): integer;
begin
  Result := 0;
end;
}

end.
```

3. Save the following _Pascal_ source code as `FnListTest2.pas`:

```Pascal
unit FnListTest2;


interface

uses
  System.SysUtils, System.Types, System.Classes;


function  FreeFunc1(const Param: integer): integer;
procedure FreeProc1(AParam: integer);



implementation

function FreeFunc1(const Param: integer): integer;
begin
  Result := 0;
end;


procedure FreeProc1(AParam: integer);
begin
end;


end.
```

---

### Steps to Reproduce (Issue 1)
1. Navigate to directory `<Npp-folder>\functionList` and load file `pascal.xml` into downloaded _Notepad++_.
2. Download the following file [>> pascal.txt <<](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10355466/pascal.txt), copy&paste its content into `pascal.xml` and save it. **Please note:** This is the _Pascal/Delphi_ function list parser of the next version of _Notepad++_. You can test the regex contained in `parser` XML node's attribute `commentExpr` with normal _Search & Replace_ dialog of _Notepad++_ on the `FnListTest1.pas` file from the _**Prerequisites**_ section to verify that it matches _Pascal_ block comments.
3. Restart _Notepad++_ and load the `FnListTest1.pas` file from the _**Prerequisites**_ section.
4. Watch the tree in function list panel, it will contain **one** node.

### Expected Behavior (Issue 1)
In step 4, the tree in function list panel should contain **no** nodes as the only existing function is part of a block comment.

### Actual Behavior (Issue 1)
In step 4, the tree in function list panel contains one node `FreeFunc1` though it is commented out. This is unwanted behaviour. It happens though function `FreeFunc1` is part of a block comment.

---

### Steps to Reproduce (Issue 2)
1. Navigate to directory `<Npp-folder>\functionList` and load file `pascal.xml` into _Notepad++_.
2. Download the following file [>> pascal.txt <<](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10352197/pascal.txt), copy&paste its content into `pascal.xml` and save it. **Please note:** This is a function list parser that contains a mixed parser. However, the class parser part is stripped down to nearly empty regular expressions that do absolutely nothing. The function parser part is reduced to the absolute necessary minimum. In order to prevent matching procedure/function declarations in the interface section of _Pascal_ files, in XML node `parser` the `commentExpr` attribute contains the regex `(?is:^\h*INTERFACE\h*$.*?^\h*IMPLEMENTATION\h*$)`. This declares everything between the keywords `interface` and `implementation` as a comment, thus this section should not be parsed.
3. Restart _Notepad++_ and load the `FnListTest2.pas` file from the _**Prerequisites**_ section.
4. Watch the tree in function list panel, it will contain **four** nodes.
5. Navigate to directory `<Npp-folder>\functionList`, load file `pascal.xml` into _Notepad++_ and delete the whole `classRange` XML node.
6. Save file `pascal.xml`, restart _Notepad++_ and load the `FnListTest2.pas` file from the _**Prerequisites**_ section.
7. Watch the tree in function list panel, now it will contain only **two** nodes.

### Expected Behavior (Issue 2)
In step 4, the tree in function list panel should contain two nodes, `FreeFunc1` and `FreeProc1`, the same like in step 7.

### Actual Behavior (Issue 2)
In step 4, the tree in function list panel contains four nodes, two times `FreeFunc1` and two times `FreeProc1`. Clicking on these tree items shows that two of them refer to the declarations of the procedures/functions in the `interface` section of the _Pascal_ file. This is unwanted behaviour. Solely because the parser file contains a dummy class parser the tree in function list panel is filled with wrong data.

---

### Steps to Reproduce (Issue 3)
1. Navigate to directory `<Npp-folder>\functionList` and load file `pascal.xml` into _Notepad++_.
2. Download the following file [>> pascal.txt <<](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10352426/pascal.txt), copy&paste its content into `pascal.xml` and save it. **Please note:** This is a function list parser that contains only a function parser that is reduced to the absolute necessary minimum. In order to prevent matching procedure/function declarations in the interface section of _Pascal_ files, in XML node `function` the `mainExpr` attribute contains the regex `(?!(?s:.*?^\h*IMPLEMENTATION\h*$))`. This negative lookahead filters out patterns of procedures/functions that are followed by the `implementation` keyword at a line's beginning somewhere after them.
3. Restart _Notepad++_ and load the `FnListTest2.pas`  file from the _**Prerequisites**_ section.
4. Watch the tree in function list panel, it will contain **two** nodes.
5. Set the caret to one of lines 12, 13 or 14 in the loaded _Pascal_ file (the empty lines between the declarations of the procedures/functions and the `implementation` keyword) and add some type of comment. It can be a line comment (`// xxx`) or a block comment (`{ xxx }` or `(* xxx *)`).
6. In function list panel, click the `Update` toolbar button in order to make the function list to apply your edits from step 5.
7. Watch the tree in function list panel, now it will contain **four** nodes.

### Expected Behavior (Issue 3)
In step 7, the tree in function list panel should contain two nodes, `FreeFunc1` and `FreeProc1`, the same like in step 4.

### Actual Behavior (Issue 3)
In step 7, the tree in function list panel contains four nodes, two times `FreeFunc1` and two times `FreeProc1`. Clicking on these tree items shows that two of them refer to the declarations of the procedures/functions in the `interface` section of the _Pascal_ file. This is unwanted behaviour. Solely because of adding a comment in step 5. the tree in function list panel is filled with wrong data.

---

### Applies To
_Notepad++_ v8.4.8, 32 bit and 64 bit version

mpheath:
Fix #12759

Functions can be found in comment zones when using `FunctionMixParser::parse()`. Passing `commentZones` to `funcParse()` can invalidate functions found in comment zones.

### DIFF
--- a/PowerEditor/src/WinControls/FunctionList/functionParser.cpp
+++ b/PowerEditor/src/WinControls/FunctionList/functionParser.cpp
@@ -735,7 +735,7 @@ void FunctionMixParser::parse(std::vector<foundInfo> & foundInfos, size_t begin,
 	{
 		for (size_t i = 0, len = nonScannedZones.size(); i < len; ++i)
 		{
-			_funcUnitPaser->funcParse(foundInfos, nonScannedZones[i].first, nonScannedZones[i].second, ppEditView, classStructName);
+			_funcUnitPaser->funcParse(foundInfos, nonScannedZones[i].first, nonScannedZones[i].second, ppEditView, classStructName, &commentZones);
 		}
 	}
 }


dinkumoil:
Many thanks for that fix! However, there is still an issue left.

I've tested the _AppVeyor_ build with the test cases in #12759. Issues 1 and 2 are resolved, issue 3 is still valid. But it can be resolved by changing my Pascal/Delphi function list parser in a way that increases its performance as well, so I will do that anyway.

<details>
<summary>Read how I will do that.</summary>
I will extend the <code>commentExpr</code> attribute to define everything between the <i>Pascal/Delphi</i> keywords <code>interface</code> and <code>implementation</code> as a comment and remove the negative lookahead for the the keyword <code>implementation</code> that is currently used to prevent procedure/function declarations in the <code>interface</code> section of a unit from showing up in the function list tree. I already tested this approach and it works.
</details>

Nevertheless there is still a flaw in function list's C++ code. Would you mind to have a look at it once more?

mpheath:
@dinkumoil  The released [pascal.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/87a53d1ed2dbf5cf072498660002721effc06ca6/PowerEditor/installer/functionList/pascal.xml) from v8.4.8 works better than the [pascal.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10352426/pascal.txt) linked in the issue ticket.

<details>

<summary>Tested results</summary>

The released pascal.xml shows...

 * FreeFunc1
 * FreeProc1

initially and insert `// xxx` at line 13 causes no change.

The linked pascal.xml shows...

 * FreeFunc1
 * FreeProc1

initially and insert `// xxx` at line 13 causes...

 * FreeFunc1
 * FreeProc1
 * FreeFunc1
 * FreeProc1

which appears to be a possible regex pattern issue.

Perhaps the unfixed parser caused you to make a bugged regex pattern with the later? Can you confirm my results.

<hr>

</details>

Using `commentExpr` seems a good idea as the comment zones should be safer now from function detection.

>Nevertheless there is still a flaw in function list's C++ code. Would you mind to have a look at it once more?

You may need to give me details of the issue as I am lost to what the flaw is.

dinkumoil:
@mpheath 

> The released pascal.xml v8.4.8 works better than the pascal.xml linked in the issue ticket.

The _pascal.xml_ released with Npp v8.4.8 has other issues and is already outdated. The next version of Npp will be released with an updated version. 

> appears to be a possible regex pattern issue. Perhaps the unfixed parser caused you to make a bugged regex pattern with the later? Can you confirm my results.

Yes, I can confirm your results. But I don't think that issue 3 in #12759 is caused by a regex pattern issue. The function list parser and the _FnListTest2.pas_ source file I provided in issue 3 for testing are reduced to the absolute minimum to simplify testing and debugging. The regex pattern `(?!(?s:.*?^\h*IMPLEMENTATION\h*$))` can not be the cause of this unwanted behavior as it uses a negative lookahead (`?!`) to exclude procedure/function declarations whose characteristic is that they are followed by arbitrary characters (`.*?`,  non-greedy matching) until there is a line with the `implementation` keyword optionally surrounded with space or tab characters (`^\h*IMPLEMENTATION\h*$`). The `?s:` activates the regex engine's single line mode so that `.` matches line breaks too. Case insensitive mode has been activated with `(?im-s)` at the very beginning of the whole regex. So `.*?` should also match line comments and block comments.

> You may need to give me details of the issue as I am lost to what the flaw is.

Please read issue 3 in #12759 thoroughly and follow the provided steps to reproduce the issue. The flaw is that the content of function list tree changes only because a comment is added to the source code. This presumably happens because the created data model of the source file or the processing of this data model is buggy. It seems like the presence of a comment interrupts the matching area of the regex `(?!(?s:.*?^\h*IMPLEMENTATION\h*$))` so that it fails which in turn causes the declaration of _Pascal/Delphi_ functions to be part of the function list tree.

mpheath:
@dinkumoil  Thanks for the clarification.

`FunctionUnitParser::parse()` is being called with *FnListTest2.pas* as no **class** tags exist in the later xml file.

This is some output with some comments that I added:

``` python
commentZone: 190, 196        # the comment line
nonCommentZones: 0, 189      # up to the comment line
funcZone: 96, 147            # line 10 func header
funcZone: 149, 186           # line 11 func header
nonCommentZones: 197, 365    # after comment line until the eof
funcZone: 218, 268           # line 17 func defined
funcZone: 303, 340           # line 23 func defined

# implementation is on line 15 pos 200, 214
```
The searches are being done in `nonCommentZones`. One zone is from the beginning up to the comment line and the next zone is after the comment line to eof (end of file). Your look ahead pattern is trying to start in the first zone and search for `implementation` in the next zone which it will not find as it does not have access to cross over zones.

This function call does the searching
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/df658a601013bb74b0690196f24ae82ebec559b5/PowerEditor/src/WinControls/FunctionList/functionParser.cpp#L391
and you may notice the `begin` and `end` parameters that restricts the search to a zone. The regex pattern cannot go past `end` of which the first `nonCommentZones` ends at pos `189`. `implementation` is pos `200`.

Passing `commentZones` to `funcParser()` here will achieve nothing as it searches in `nonCommentZones`. `FunctionMixParser::parse` is different in comparison to where it searches and then validates if not in `CommentZones`. 

The behaviour of the code appears to be by design and is about 10 years old. Hmmm, this needs consideration as what to do about this code.

dinkumoil:
@mpheath 

Thank you for your investigations and detailed explanation.

> The behaviour of the code appears to be by design and is about 10 years old.

I agree, this is an important fact. Although this behavior you have discovered is indeed a flaw, changing this code should be given careful consideration.

So, @donho what do you think about it? In my last PR [you wrote](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12694#issuecomment-1368259863)

> I will see what I can do about it.

It seems like now it's up to you to make a decision what should be done. **Please note:** This PR already provides an important fix for function list parsers in general and resolves all issues known so far of my _Pascal/Delphi_ function list parser, it definitely should be accepted. Additionally I will create a new issue and a related PR to improve the current _Pascal/Delphi_ function list parser in order to increase its performance.

mpheath:
Recent Closed Issue #12792
Older Open Issue #3685

and probably others if searched more are caused by inserting comments. This splits the source into smaller zones and so the regex pattern can fail to match if too close to a inserted comment. `FunctionMixParser::parse` might be a better behaviour to mimick so as to allow freer range of searching and then invalidate if matched within a `commentZone`.

dinkumoil:
Wow, #3685 provides a really long list of similar issues found with a bunch of other programming languages. So, a workaround would be to apply the PR of @mpheath and convert all function list parsers, which are currently only function parsers, to mixed parsers with a dummy class parser like [the one in issue 2](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10352197/pascal.txt) of #12759.

@donho It seems its time to take action on this topic as it's a many years old "corpse in the cellar", like we say here in Germany.

Issue #9454, opened 2021-01-27
Document List Panel a.k.a. Doc Switcher disappears on minimize


wilkowy:
### Description of the Issue
Document List Panel a.k.a. Doc Switcher box disappears on minimizing and restoring pinned N++

### Steps to Reproduce the Issue
1. Pin N++ to taskbar
2. Settings->Preferences->General->Document List Panel - mark Show
3. Click N++ icon on taskbar to minimize window or minimize from minimize button
4. Restore N++ by clicking on taskbar

### Actual Behavior
![npp-docswitcher](https://user-images.githubusercontent.com/5204438/106015018-5c170f80-60be-11eb-8437-ce24c0de1402.jpg)

Important:
If restoring N++ from tray bar container (no matter if pinned or not) Doc Swicher box reappears.

### Debug Information
Notepad++ v7.9.2   (64-bit)
Build time : Dec 31 2020 - 04:01:34
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 7 Ultimate (64-bit) 
OS Build : 7601.23964
Current ANSI codepage : 1250
Plugins : AnalysePlugin.dll ComparePlugin.dll DSpellCheck.dll JSMinNPP.dll mimeTools.dll NppConverter.dll nppcrypt.dll NppExport.dll NPPJSONViewer.dll NppQrCode64.dll SurroundSelection.dll XMLTools.dll _CustomizeToolbar.dll

xaminmo:
Just to clarify, some version of this bug still exists in 8.4.4.

To clarify the trigger:
* Open NP++, docs, etc.
* Enable Document list if not already.
* Minimize NP++ window to tray.
* Open a new text file from windows context menu
* When NP++ window is restored, the document list is missing.
* The view menu still shows it's enabled.

Resolution is to disable then re-enable the document list from the View menu.

Notepad++ v8.4.4   (32-bit)
Build time : Jul 15 2022 - 17:56:38
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : "E:\GoogleDrive\Xaminmo\201x Money\ESS_Conflicts\2022-07_Meltdown_regarding_being_Excluded.txt"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.1826
[...]

PR #12817, opened 2023-01-19
Fixed dock-able panels not restoring for mono instances in tray

robotboy655:
Happens when NP++ is in the system tray (minimize to tray setting), and is restored via mono instance restore mechanism, i.e. opening a second NP++ instance.

Fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9454


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2490,6 +2490,33 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return TRUE;
 		}
 
+		case NPPM_INTERNAL_RESTOREMONOINSTANCE:
+		{
+			// When mono instance, bring this one to front
+			if (nullptr != _pTrayIco && _pTrayIco->isInTray())
```
donho, line 2496: Please use `_pTrayIco != nullptr !` instead of `nullptr != _pTrayIco`.

donho (changes requested):

PowerEditor/src/NppBigSwitch.cpp
robotboy655: Changed as per request.

donho (approved):

Issue #12818, opened 2023-01-19
[Code] Non-standard `swprintf` and `wcstok`


ozone10:
### Description of the Issue

Non-standard [swprintf](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-sprintf-l-swprintf-swprintf-l-swprintf-l?view=msvc-170) may be removed in future

The two argument version of [wcstok](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strtok-strtok-l-wcstok-wcstok-l-mbstok-mbstok-l?view=msvc-170) isn't standard

### Expected Behavior

Notepad++ use standard version of functions
### Actual Behavior

Notepad++ use non-standard version of functions

### Debug Information

Notepad++ v8.4.8

PR #12819, opened 2023-01-19
Use conformant code

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12818


PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
```diff
@@ -664,8 +664,9 @@ void PluginsManager::runPluginCommand(size_t i)
 			}
 			catch (...)
 			{
-				TCHAR funcInfo[128];
-				generic_sprintf(funcInfo, TEXT("runPluginCommand(size_t i : %zd)"), i);
```
donho, line 668: Hmm... The signature of swprintf is indeed:
```cpp
int swprintf(
   wchar_t *buffer,
   size_t count,
   const wchar_t *format [,
   argument]...
);
```
according the [MS document](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-sprintf-l-swprintf-swprintf-l-swprintf-l?view=msvc-170)

I don't understand how the above code can be compiled.
Do you have any explanation?

donho: OK, I see. There are another swprintf function with no `size_t count,` as argument.

PR #13003, opened 2023-01-30
Fix: use the correct attribute name of TabBar item

0xabcdefff:
Found somewhere use CMID, somewhere use CMDID attribute name of TabBar item.


donho:
The corrections are all wrong (please refer english.xml).
Even there is half part get fixed in this PR, it's better (IMO) to leave current state.
So if there are some translators want to work on the incorrect entries, they will do them right by testing the translations.

Issue #9897, opened 2021-05-21
PHP8 Missing Functions


valzargaming:
### PHP 8 functions

PHP 8 has added new functions such as with str_starts_with and str_contains that are not currently listed in Notepad++'s current PHP language extension.

### Steps to Reproduce the Issue

1. Start typing str_starts_with in a .php file with the language set to PHP
2. Expect it to show up in the auto-completion list with expected parameters, but it is not listed.

### Expected Behavior

PHP 8 functions should be listed.

### Actual Behavior

They aren't.

### Resources:
https://www.php.net/releases/8.0/en.php

PR #13011, opened 2023-01-31
[xml] Added new php 8 functions

Compaszer:
Added new php 8 functions to syntax highlighting and functions description as requested in issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9897


Mateos81:
When a commit/PR is only related to XML files, please add [XML] in front of them, so AppVeyor doesn't launch a full compile

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d2ba9cae0e7c5cd1650c5d224408f9e7c7cac43b

Issue #12949, opened 2023-01-27
middle button click on document list no longer closes a document


jkumeboshi:
### Description of the Issue
8.4.6 update has removed a useful function: when clicking with mouse middle button on a document in the document list the document closes.

### Steps to Reproduce the Issue
1. open a couple of files
2. show document list: menu->view->document list
3. middle click on any document entry in the list

### Expected Behavior
document closes

### Actual Behavior
nothing happens

### Debug Information
Notepad++ v8.4.8   (64-bit)
Build time : Dec 24 2022 - 19:40:27
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Users\gam\Downloads\logs_Fedora33 (1).txt"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 22H2
OS Build : 22621.1105
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExec (0.8.2)
    NppExport (0.4)

alankilborn:
Middle click close doc works fine for me.

Notepad++ v8.4.8   (64-bit)
Build time : Dec 24 2022 - 19:40:27
Path : W:\npp.misc\_NoBackup\Releases\npp.8.4.8\npp.8.4.8.portable.x64\notepad++.exe
Command Line : -titleAdd="8.4.8 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.2364
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppEditorConfig (0.4)
    NppExport (0.4)
    PythonScript (2)

jkumeboshi:
Maybe is a misbehavior in Windows 11?
Actually, I noticed it when recently I've upgrade to Windows 11.

alankilborn:
Oh wait, I misunderstood.  I didn't read closely enough.  I thought you meant for a file tab on the tab bar.

In "Document List", when did middle-click ever close a document?  I tried several recent versions and none of them did anything when middle-clicking on a filename in the "Document List".

jkumeboshi:
[...]

Issue #8445, opened 2020-06-20
No element gets focus after middle-clicking a tab


Yaron10:
**STR:**
Middle-click a tab.

**Result:**
No element has the keyboard focus.
(It happens after pressing the middle-click button and before releasing it).

**Expected Result:** 
Some element should get the focus.
***
Should it always be the editor or the previously focused element?

Moving focus to the editor on *left-clicking* a tab is the expected behavior and can be very useful.
If this is implemented, I'd suggest to always move focus to the editor on middle-clicking as well. 
***
Notepad++ v7.8.3   (32-bit)
Build time : Jan 17 2020 - 18:20:02
Path : C:\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

SkandaSrinivasan:
I would look to take a shot at fixing this and would appreciate some help in understanding if I am going about this right way. The code in TabBar.cpp is as below ,
```
case WM_MBUTTONUP:
		{
			int xPos = LOWORD(lParam);
			int yPos = HIWORD(lParam);
			int currentTabOn = getTabIndexAt(xPos, yPos);
			notify(TCN_TABDELETE, currentTabOn);
			return TRUE;
		}
```
So after triggering this event , the tab falls out of focus and the cursor disappears. Not exactly sure why, since it is similar to the close button action.

Would doing the below
```
case WM_MBUTTONUP:
		{
			int xPos = LOWORD(lParam);
			int yPos = HIWORD(lParam);
			int currentTabOn = getTabIndexAt(xPos, yPos);
			notify(TCN_TABDELETE, currentTabOn);
			----int currentTab = TabCtrl_GetCurFocus(hwnd);
			----setActiveTab(currentTab);
			return TRUE;
		}
```
be a valid solution? It fixes the bug, but I would like some feedback on if this is the right way to go about fixing this.

PR #13015, opened 2023-02-01
Close Document List documents with middle mouse click

robotboy655:
Adds parity with default tabs to Document List window, allowing to use middle mouse click to close documents.

Fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12949


alankilborn:
Shouldn't this be made configurable via a _Preference_ setting?
Bigger picture:  This AND the tab-bar middle click?
Perhaps a separate issue, but since new code relating to middle-click is being proposed here, why not make middle-click fully rounded out?

robotboy655:
What's the benefit in adding a setting for this? I have never accidentally clicked my mouse wheel, or accidentally closed a document with middle mouse click. 

Either way this is probably a better idea to discuss this in issue of its own.

alankilborn:
Some people are used to middle-click doing other actions in other applications, or perhaps something they've programmed via custom mouse config software.  When they come to Notepad++ they have "muscle memory" for those actions.  If their muscle memory fires, it is better to do nothing than the "drastic step" of closing a document.  Thus the desire for a preference setting.

I believe there is another issue requesting a preference setting for middle-click for the existing tab-bar file-close.  There has been no action on that issue.  There is already a setting for double-click to close, so why not one for middle click as well...

You are attempting to get what I'd call somewhat related code to be accepted.  My thinking was to take advantage of this situation, to make the best and most well-rounded solution.

Mateos81:
In all tab-based applications I can think of in Windows, middle-mouse click is to close a tab (web browsers, ...)
Only the task bar opens a new instance

Yaron10:
@robotboy655,

Thank you for your work. 👍 

Related: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8445.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1317,6 +1317,21 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return TRUE;
 		}
 
+		case NPPM_CLOSEDOC:
+		{
+			// Close a document without switching to it
+			int whichView = ((wParam != MAIN_VIEW) && (wParam != SUB_VIEW)) ? currentView() : static_cast<int32_t>(wParam);
```
Yaron10, line 1323: Just curious: why should `(wParam != MAIN_VIEW) && (wParam != SUB_VIEW)` be an option?

Thank you.

robotboy655: It's copy pasted directly from `NPPM_ACTIVATEDOC` message above.

Yaron10: Thank you. 👍

donho:
Please see:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12949#issuecomment-1420124089

PowerEditor/src/NppBigSwitch.cpp
donho: @Yaron10 
> Just curious: why should (wParam != MAIN_VIEW) && (wParam != SUB_VIEW) be an option?

I guess (I coded long time ago) it's for preventing wParam from other values than MAIN_VIEW or SUB_VIEW.

PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcher.cpp
```diff
@@ -89,6 +89,41 @@ intptr_t CALLBACK VerticalFileSwitcher::run_dlgProc(UINT message, WPARAM wParam,
 			return TRUE;
 		}
 
+		case WM_PARENTNOTIFY:
+		{
+			switch ( wParam )
+			{
+				case WM_MBUTTONDOWN:
```
donho, line 96: According several browsers and several windows OS middle click to close behaviour (that I checked) - they are all closed on mous up. So here should be `WM_MBUTTONUP` instead.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -549,7 +549,8 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// void* NPPM_GETBOOKMARKID(0, 0)
 	// Returns the bookmark ID
 
-
+	#define NPPM_CLOSEDOC (NPPMSG + 112)
```
donho, line 552: The message is wrong named (should be `NPPM_INTERNAL_CLOSEDOC` ) and wrong placed (should be in "resource.h").
Here is opened API for plugins - once a new API is added here, it should be maintained & available. Since close doc API has never been asked, let's not add it.

donho (changes requested):

PowerEditor/src/NppBigSwitch.cpp
Yaron10: @donho,

Being extra safe is good. :)
Thank you. 👍

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
Yaron10: @donho,

I'm wondering if you also meant that the new `closeDoc()` function wasn't necessary and an existing Close function could be used instead.

robotboy655:
@donho I have changed the internal message as requested, but with the issue of `WM_MBUTTONDOWN` is more challenging.

It appears that `ListView` element handles mouse up events in a special way that makes `WM_MBUTTONUP` event never happen for parent window, which makes it not possible to capture it the same was as I do `WM_MBUTTONDOWN` right now. I don't know if there's a way around that.

The closing being on `WM_MBUTTONDOWN` is also consistent with Visual Studio.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
donho: @Yaron10 
> I'm wondering if you also meant that the new closeDoc() function wasn't necessary and an existing Close function could be used instead.

I mean only for the API `NPPM_CLOSEDOC` which should be `NPPM_INTERNAL_CLOSEDOC`. But if you see any existent function which can replace closeDoc() function, please let me know.

Yaron10: @donho,

I did not have anything concrete in mind. Just a vague thought.

Thank you and @robotboy655 for your work. 👍

ozone10:
@robotboy655 
IIRC `WM_PARENTNOTIFY` doesn't handle "up" mouse messages.

You can subclass listview itself and handle `WM_MBUTTONUP` directly.

donho (approved):

donho:
@robotboy655 
> I have changed the internal message as requested, but with the issue of `WM_MBUTTONDOWN` is more challenging.

OK, I will keep working on `WM_MBUTTONUP` with thee suggestion provided by @ozone10 
Thank you for the PR.

Yaron10:
@donho,

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4d107e269106dcf136b635d6ff3ab42d3b0a8fce.

Great. 👍 
Thank you.

Yaron10:
@donho & @robotboy655,

**STR:**
Open a new document and edit it (no other docs or other docs are all saved).
Middle-click in the Documents List to close it.
When prompted - do not save.

**Result:**
`Save All` is wrongly enabled.

`checkDocState()` should be added after `fileClose()`.

Issue #8284, opened 2020-05-19
[Feature Request] Make ZERO WIDTH SPACE have some visual component


GhbSmwc:
Try sorting this:
```
https://pbs.twimg.com/media/
https://twitter.com/
​https://pbs.twimg.com/media/
```
Somehow, the last URL is positioned wrong.

Wait.., there is a zero-width character on the third (before the h in https) which is "​" (select to see that character). I think NP++'s show symbol should also show such characters.

xylographe:
The character is U+200B (Unicode name: ZERO WIDTH SPACE).

sasumner:
> I think NP++'s show symbol should also show such characters.

What's your opinion on what it should look like?

GhbSmwc:
A yellow line, much like when you make it show spaces and tabs.

sasumner:
@GhbSmwc I took the liberty of renaming your issue; hopefully that is okay with you. :-)

sasumner:
This is going to depend upon Scintilla.

sasumner:
This was "addressed" by the Scintilla project back in 2013:  https://sourceforge.net/p/scintilla/feature-requests/980/

Their response was to add an API so that you can represent any character you want, "specially":
https://www.scintilla.org/ScintillaDoc.html#SCI_SETREPRESENTATION

The representation mechanism is the "white text on black box" variety, much like line-endings look when visually enabled in Notepad++:

![image](https://user-images.githubusercontent.com/30118311/82374292-d7330c80-99ec-11ea-9020-48cc6c021ce4.png)

Here's an example of what it would look like to set the representation to `""` (an empty string), the narrowest possible thing, which is harmonious with a "zero-width" character:

![image](https://user-images.githubusercontent.com/30118311/82374618-4c064680-99ed-11ea-9346-33c2667ececb.png)

I suppose it is "workable" but I'd say it isn't ideal.

Daksol:
This request has come up more than once if you look back through the Issues history. I encountered this problem when "screen scraping" data out of certain HTML pages.

**If this is not taken up, then maybe add a Caution to NPP interface to say that the option "View-> Show Symbol -> Show All Characters" does NOT show characters such as ZeroWidthSpace and friends.**

Only way to detect a Zero Width Space is to move a cursor along a line of text while watching the character position number in the status bar,

Daksol:
Previous Issues raised on the same topic, going back to 2015

[...]

Issue #4731, opened 2018-08-01
Feature Request - Visualize Zero-width spaces in "show all chars" mode.


Daksol:
### Description of the Issue
Character Decimal 8203, Hex 200B can be put into a text file when screen scraping from certain webpages. These chars are invisible een when in View->Show All Chars mode.
### Steps to Reproduce the Issue

1.  Create a string in MS Excel including UNICODE(8203) 
2. Copy string and paste into NPP.
3. Can also be found when screen scraping from SharePoint O365 pages which have html tables.is first char in a cell.

### Expected Behavior
ZWSP should show up in View all chars

### Actual Behavior
Invisible, except by watching cursor movement when traversing s string wit arrow key.

### Debug Information
N/a

Some discussions on NPP community site, search "Zero-width space"

Daksol:
Earlier feature request issue #827 is essentially the same as this. #827 mentions other characters which can be problematic in addition to the ZWSP.

Issue #827, opened 2015-08-31
Show White Space and TAB does not show all whitespace


EffectivelyEfficient:
"View | Show Symbol | Show White Space and TAB" does not show all white space. This also affects "Show All Characters"

To reproduce, create an html page that contains &#8203;&_#_8203;&#8203; (without the underscores) and then copy the character to notepad++. [If this site allows it, the offending char might also be in this paragraph]

8203 is a Unicode character also know as ZERO WIDTH SPACE (ZWSP). This character is intended for invisible word separation and for line break control; it has no width, but its presence between two characters does not prevent increased letter spacing in justification.

dg9ngf:
There are more whitespace characters in Unicode. Perhaps the most interesting are:

* U+0020 "Space"
* U+00A0 "No-Break Space"
* U+202F "Narrow No-Break Space"

The normal space is the only of them that is visualised. But it's hardly visible so tiny is the symbol for it. It should definitely be a thicker graphic.

The other two mentioned should have different symbols. Microsoft Word for example shows something like ° for the non-breaking space, but also nothing for the narrow non-breaking space.

There's a more complete overview in [Wikipedia](https://de.wikipedia.org/wiki/Leerzeichen#Desktop-Publishing_und_Webtypographie) (not available in English).

aaaxx:
It would also be a good idea to have those other spaces as a separate item in Style Configurator, so that you can have normal spaces dimmed and special ones popping out.

Daksol:
See also feature request #4731, also asking for enhanced visuslisation. This feature request #827 is better as it covers a wider range of problematic characters.

MetaChuh:
we are closing this issue due to its age and multiple newer notepad++ versions since it was filed.

please update your notepad++ to the latest version [<kbd>Notepad++ 7.6.6</kbd>](https://notepad-plus-plus.org/download/v7.6.6.html).
if your issue persists, and it is still of relevance to you, please feel free to file a new issue.

thank you for your help and best regards.

-----
```
general notes:

please close your issues as soon as you are satisfied, to help
keeping the notepad++ issue tracker clean.

when creating a new issue, please provide all information as shown at the issue template.
your notepad++ debug info is part of the minimum requirements.
(? > Debug Info... > Copy debug info into clipboard).

please also conduct a search at the issue tracker beforehand, to check if a similar issue
already exists.

for general questions, or if it is not sure, whether your issue is directly related to
[...]

Issue #8111, opened 2020-04-13
Add feature to hide all non-printable characters


Gitoffthelawn:
From what I can tell, there is no way to hide all non-printable characters.

I've searched the web for ways to accomplish this in Notepad++, and it seems to be commonly desired functionality.

Can it be implemented?

sasumner:
What's a "non-printable" character?

>seems to be commonly desired functionality

Can you cite some references for that?

Gitoffthelawn:
> What's a "non-printable" character?

All ASCII characters below hexadecimal 20 (decimal 32), with the exception of hex 09 (dec 9), 0A (dec 10), and hex 0D (dec 13).  Also hex 81 (dec 129), hex 8D (dec 141), hex 8F (dec 143), hex 90 (dec 144), and hex 9D (dec 157).

Notepad++ represents them all as 2-3 character inverted text control strings (many of which are references to ancient serial communication protocols from 40+ years ago).

The best way to hide them is usually to hide all hex 00 (nulls), and show all sequences of 1 or more non-printable characters with a single space (hex 20).

> Can you cite some references for that?

Yes, if that will result in getting it implemented. :)

I performed a bunch of web searches looking for this Notepad++ functionality, and I'll I found was other people disappointed in the lack of this feature.  IIRC, I also found an old Notepad++ issue (on Sourceforge) from about a decade ago that said this would be resolved soon as the project was no longer going to use Scintilla!  ;)

sasumner:
>>What's a "non-printable" character?

>All ASCII characters below hexadecimal 20 (decimal 32), with the exception of hex 09 (dec 9), 0A (dec 10), and hex 0D (dec 13). Also hex 81 (dec 129), hex 8D (dec 141), hex 8F (dec 143), hex 90 (dec 144), and hex 9D (dec 157).

By that definition, why aren't the following also in the list, (decimal) 127, 128, 130-140, 142, 145-156, 158-255 ?

>The best way to hide them is usually to hide all hex 00 (nulls), and show all sequences of 1 or more non-printable characters with a single space (hex 20).

As one person's data point, I'm not convinced that hiding data from the user is the best thing to do.

>I performed a bunch of web searches

That really isn't citing references unless you share the results as explicit web links. :-)

Gitoffthelawn:
> By that definition, why aren't the following also in the list, (decimal) 127, 128, 130-140, 142, 145-156, 158-255 ?

[...]

PR #13020, opened 2023-02-01
Show non-printable characters

ozone10:
fix issue with tooltips in preference dialog, when changing between dark/light mode

image

defined for:
```
Codepoint : character : name : abbreviation
U+00A0 :   : no-break space : NBSP
U+1680 :   : ogham space mark : OSPM
U+180E : ᠎ : mongolian vowel separator : MVS
U+2000 :   : en quad : NQSP
U+2001 :   : em quad : MQSP
U+2002 :   : en space : ENSP
U+2003 :   : em space : EMSP
U+2004 :   : three-per-em space : 3/MSP
U+2005 :   : four-per-em space : 4/MSP
U+2006 :   : six-per-em space : 6/MSP
U+2007 :   : figure space : FSP
U+2008 :   : punctation space : PSP
U+2009 :   : thin space : THSP
U+200A :   : hair space : HSP
U+200B : ​ : zero-width space : ZWSP
U+200C : ‌ : zero-width non-joiner : ZWNJ
U+200D : ‍ : zero-width joiner : ZWJ
U+200E : ‎ : left-to-right mark : LRM
U+200F : ‏ : right-to-left mark : RLM
U+2028 :   : line separator : LS
U+2029 :   : paragraph separator : PS
U+202A : ‪ : left-to-right embedding : LRE
U+202B : ‫ : right-to-left embedding : RLE
U+202C : ‬ : pop directional formatting : PDF
U+202D : ‭ : left-to-right override : LRO
U+202E : ‮ : right-to-left override : RLO
U+202F :   : narrow no-break space : NNBSP
U+205F :   : medium mathematical space : MMSP
U+2060 : ⁠ : word joiner : WJ
U+2066 : ⁦ : left-to-right isolate : LRI
U+2067 : ⁧ : right-to-left isolate : RLI
U+2068 : ⁨ : first strong isolate : FSI
U+2069 : ⁩ : pop directional isolate : PDI
U+206A : ⁪ : inhibit symmetric swapping : ISS
U+206B : ⁫ : activate symmetric swapping : ASS
U+206C : ⁬ : inhibit arabic form shaping : IAFS
U+206D : ⁭ : activate arabic form shaping : AAFS
U+206E : ⁮ : national digit shapes : NADS
U+206F : ⁯ : nominal digit shapes : NODS
U+3000 : 　 : ideographic space : IDSP
U+FEFF : ﻿ : zero-width no-break space : ZWNBSP
```

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/827
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4731
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8284


pryrt:
From a UI perspective -- I originally thought those `1`, `2`, and `?` were supposed to be _buttons_ of some sort, and tried clicking them rather than just hovering.  And this was even having seen the "fix issue with tooltips" in the PR.  So I'm not sure that your average user is going to know to _hover_ rather than _click_ on those.  Is there a way to make clicking the box also pop up the hover text? 

Also, how critical is it, really, to have that full list in the UI?  Would it be better to just put those lists in the User Manual, and just have the Preferences dialog show the Abbreviation and Codepoint radio buttons, without the 1,2,? non-buttons.  Maybe have a short hover for each, so **Abbreviation** has a hover of `e.g. NBSP for the no-break space character; see the User Manual for the full list`, and **Codepoint** has a hover of `e.g. U+00A0 for the no-break-space character; see the User Manual for the full list`.

alankilborn:
Here's someone else's definition of "non-printable":  #8111

alankilborn:
>I originally thought those 1, 2, and ? were supposed to be buttons of some sort

I endorse @pryrt 's feelings about these buttons, including "is it really needed".

If user clicks one, nothing happens.  Worse, after the click attempt, user's mouse points to the button while user waits for something to happen.  This causes the tip to NEVER appear (without user moving mouse away and then back into button space).

Yaron10:
@ozone10,

Thank you for your work. 👍 

Toggling the command or changing the preference does not take effect until clicking in the editing area.
Can this be changed?

Yaron10:
> Would it be better to just put those lists in the User Manual, and just have the Preferences dialog show the Abbreviation and Codepoint radio buttons, without the 1,2,? non-buttons.

👍 

A link to the User Manual is another option.

ozone10:
@pryrt, @alankilborn, 
thx for feedback.
I did some quick tweaks for GUI could you check it.

@Yaron10 
> Toggling the command or changing the preference does not take effect until clicking in the editing area. Can this be changed?

I've noticed it too, but I am not sure how to fix it now.

Yaron10:
@ozone10,

> I've noticed it too, but I am not sure how to fix it now.

A known-issue is legit. :)

ozone10:
@alankilborn 
> Here's someone else's definition of "non-printable": #8111

well this PR is more about "non-ASCII" non-printable characters.

alankilborn:
>A link to the User Manual is another option.

There's already a link to the user manual in the `?` menu.
But I think you mean a (future) direct link to the topic at hand.
The problem with that is that then there becomes a maintenance issue as that link has to live forever.
And someone has to remember that it can't ever be changed.
I'm not saying I'm opposed, but it would be a great disappointment if someday such a link stops working.

alankilborn:
>>Here's someone else's definition of "non-printable": https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8111

>well this PR is more about "non-ASCII" non-printable characters.

Hmm, why not make a "complete" solution?  Not that I know what that looks like... :-)

alankilborn:
I really don't think `Show All Characters` is the correct name for showing whitespace+line-endings any longer, with the advent of the new command put forth by this PR.  Could this PR adjust that, or would that be a subject for a future issue/PR?

pryrt:
I don't know how I feel about `Emphasize`.  The fact that it's not a reverse-video symbol doesn't change the fact that it _is_ still showing a symbol for the space and tab characters: it might not be the same kind of symbol, but it's still a symbol, and the option does toggle whether or not that symbol is shown.

But more than that, `Whitespace and TAB Characters` was never an accurate description: "TAB" in all-caps is a holdover from people trying to make the tab character name look like the other ASCII control character abbreviations, though it's official abbreviation is actually HT.  (Though I understand that the `TAB` notation is used throughout the UI, so some might argue it should stay `TAB` here for consistency, since this PR isn't going to change `TAB` to `Tab` throughout the UI.  But I'd vote to use `Tab` here [first-letter capitalized, to match the capitalization of Space here and the capitalization of most other nouns and adjectives in the menu system, and then maybe put in a separate issue/PR to make `Tab` the consistent usage throughout the UI.)

And the tab character _is_ a [Whitespace character](https://en.wikipedia.org/wiki/Whitespace_character), as are newlines and vertical tabs and some of the "non-printable" Unicode characters, so separating "Tab" as somehow distinct from the "Whitespace" or "White Space" is not an accurate separation.  Since the only two characters affected by that toggle are the literal space (ASCII 32, U+0020) and the tab (ASCII 9, U+0009), I think that `Show Space and Tab` is more accurate and understandable.

As for the new name for the never-accurate `Show All Characters`, how about `Show Space, Tab, and End of Line` -- that is, just merge the names of the two previous menu entries into the third, multi-use entry.

ozone10:
@alankilborn 
> Hmm, why not make a "complete" solution? Not that I know what that looks like... :-)

Because ASCII characters are already visible. This PR is dealing with "invisible" characters.

IMHO I think `Show White Space and TAB` should be `Show Space and Tab` and `Show All Characters` can stay as it currently also enable/disable `Show Non-Printable Characters`

pryrt:
> `Show All Characters` can stay as it currently also enable/disable `Show Non-Printable Characters`

Oh, I hadn't noticed that toggling the Show All also forced the new menu entry to toggle.

I'm not sure I like that, because then to move from a current state of `Show Space and Tab` to the state where it shows space, tab, and newlines, would require first clicking on `Show All Characters` and then un-clicking `Show Non-Printable Characters`.  For people (like me) who are used to quickly toggling between their two favorite "show symbol" states, requiring us to go back to the menu a second time to change to the state we actually intended would get annoying really quickly.

alankilborn:
>>Hmm, why not make a "complete" solution? Not that I know what that looks like... :-)

>Because ASCII characters are already visible. This PR is dealing with "invisible" characters.

Sure, OK, in the spirit of keeping changes small.  Hopefully if someone does a solution for not showing the ASCII characters, it fits in well with this PR's changes.

alankilborn:
>Oh, I hadn't noticed that toggling the Show All also forced the new menu entry to toggle.

I hadn't noticed that either.
I guess we had to read the code in order to understand the nuances of how this PR works.

alankilborn:
> @ozone10,
> Thank you for your work. 👍

This is true.  Praise where praise is due.
NICE WORK, @ozone10 !
I hope this gets adopted, perhaps with some slight tweaks.
Again, great job.

Yaron10:
> There's already a link to the user manual in the ? menu.
But I think you mean a (future) direct link to the topic at hand.
The problem with that is that then there becomes a maintenance issue as that link has to live forever.

👍 

> Oh, I hadn't noticed that toggling the Show All also forced the new menu entry to toggle.
I'm not sure I like that...

👍 

https://github.com/pnedev/comparePlus/issues/345#issuecomment-1398751001.

> See how NPP implements "View -> Show Symbol -> Show White Space and Tab" and "View -> Show Symbol -> Show End of Line".
Checking one command, automatically unchecks the other.
IMO, that behavior is wrong in that case: the two commands are not contradicting whatsoever.
(The third command "Show All Characters" probably made the coder implement it that way).

***

@ozone10,

`Show White Space and TAB` and `Show End of Line` are currently (in Master) `Show White Space and TAB ONLY` and `Show End of Line ONLY`.
If you keep the linkage between `Show Non-Printable` to `Show All Characters`, checking either `Show White Space and TAB` or `Show End of Line` should uncheck `Show Non-Printable`.

(IMO, `Show All Characters` should be removed and each command should be independent).

If you use a MsgBox, how about replacing the `?` with `...`?

I also think that `Non-Printable` is not the best term.

Thanks again.

alankilborn:
> `Show All Characters` should be removed and each command should be independent

I agree, this would be better than existing implementations, both as currently in released N++, and in this PR.
 
> If you use a MsgBox, how about replacing the `?` with `...`?

Makes sense.

> I also think that `Non-Printable` is not the best term.

Agree, but your lack of suggesting a substitute shows how difficult coming up with a good substitute is.  :-)
Someone else suggested to me "Non-glyphed", which is growing on me...

ozone10:
Sorry wrong button.

@pryrt 
> I'm not sure I like that, because then to move from a current state of `Show Space and Tab` to the state where it shows space, tab, and newlines, would require first clicking on `Show All Characters` and then un-clicking `Show Non-Printable Characters`. For people (like me) who are used to quickly toggling between their two favorite "show symbol" states, requiring us to go back to the menu a second time to change to the state we actually intended would get annoying really quickly.

image

@alankilborn, @Yaron10 

Should it be "non-printing or control and whitespace characters"?

https://en.wikipedia.org/wiki/Control_character
https://en.wikipedia.org/wiki/Whitespace_character

Yaron10:
@alankilborn,

> but your lack of suggesting a substitute shows how difficult coming up with a good substitute is.

I haven't learned the subject and I don't fully understand it.
If it's that difficult, how about `Show Other Hidden Characters`? - Assuming that `Show All Characters` is removed.

@ozone10,

If @donho agrees to remove `Show All Characters`, the implementation should be much simpler.
If not - the extra `Sync...` setting seems to work properly, but personally I think it's a bit too much.
*And `Show Non-Printing` should be **before** `Show All Characters` and without the extra separator.*

What do you think about `Show Non-Printing` opening a sub-menu containing `Abbreviation` and `Codepoint`?
No need to add anything to the Preferences dialog. - Let the user look it up in the User Manual.

Thank you.

Yaron10:
> If not - the extra Sync... setting seems to work properly

It actually does not work as expected. :)
If `Show All Characters` is checked, `Show Non-Printing` should be unchecked.

I'd suggest to wait for @donho's opinion about removing `Show All Characters`.

ozone10:
@Yaron10 
> It actually does not work as expected. :) If `Show All Characters` is checked, `Show Non-Printing` should be unchecked.

`Show Non-Printing Character` without sync checked will work as `Show Indent Guide` (`Show Wrap Symbol`)

With sync checked it will same as if there is only `Show All Characters` menu

Currently this PR is treating `Show All Characters` as `Show Space, Tab and EOL`
`Show White Space and TAB` and `Show End of Line` with sync checked as `Show Only White Space and TAB` and `Show Only End of Line` .

Yaron10:
@ozone10,

> With sync checked it will same as if there is only Show All Characters menu

That's the part I find slightly confusing.
The first two commands are *only* (i.e. `Show Space and Tab` ONLY), whereas the new command is not.

> Show Non-Printing Character without sync doesn't affect those 3 menus, therefore the usage of separator.

But there's an option to sync.

> IMHO I don't think removing Show All Characters is good idea, people will complain.

Let's think of the following cases:

You have two shortcuts for `Space and Tab` and `EOL`.

**1.**
You display `Space and Tab`.
You now want to display `EOL` *as well*.
You intuitively use the `EOL` shortcut. 

The result is unexpected: `EOL` is displayed, and `Space and Tab` is not.

**2.**
You've "discovered" `Show All Characters`, pressed it and you now have both `Space and Tab` and `EOL` displayed.
You now want to display `EOL` *only*.
You intuitively use the `Space and Tab` shortcut to toggle and turn it off.

The result is unexpected: `Space and Tab` is displayed, and `EOL` is not.

So `Show All Characters` is only confusing and slowing one's workflow.
Let the two commands (now three) be independent. 

> People are lazy and don't read manual, then complain :).

As you understand. :)

donho:
I'm wondering... should a code/text editor show ALWAYS these non-printable characters? In this way, no ambiguity and always clear for the coders or IT people. What do you guys think about it?

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -1058,6 +1117,84 @@ intptr_t CALLBACK EditingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
+						L"U+206B : activate symmetric swapping : ASS\n"\
+						L"U+206C : inhibit arabic form shaping : IAFS\n"\
+						L"U+206D : activate arabic form shaping : AAFS\n"\
+						L"U+206E : national digit shapes : NADS\n"\
+						L"U+206F : nominal digit shapes : NODS\n"\
+						L"U+3000 : ideographic space : ISP\n"\
+						L"U+FEFF : zero-width no-break space : ZWNBSP",
+						L"Representation of selected whitespace and control characters",
+						MB_OK);
+				}
```
donho, line 1190: This dialog box is huge, and with not-well-presented information.
IMO, the display of this information is not necessarily in Notepad++, it could launch the browser to show the information in https://npp-user-manual.org/ in which we can do a pretty table easily in HTML to display such info (I did the following table in less 3 minutes in Markdown):
image

If users don't have Internet connection, never mind - it's a Nice-to-have but not Must-to-have.
Furthermore we can save the effort for the translation and its management - less is more.

donho (changes requested):
Please consider the suggestions.
[As I mentioned](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13020#issuecomment-1414582893), the `Sync with "Show All Character"` option may not be needed.
I would like have you guys' opinions to make a decision.

Yaron10:
@donho,

> A code/text editor should ALWAYS show these non-printable characters, shouldn't it?

My knowledge about it is not sufficient. Let's wait for the others' opinions.

alankilborn:
@donho

>A code/text editor should ALWAYS show these non-printable characters, shouldn't it?

For myself, up til now I have had a script that turns on visual representation of these characters (actually not ALL of these, but yet some OTHERs too -- should the list be configurable of what to show?).

And I always leave that script enabled, because I want to know if there is something there that normally I can't see.  So I suppose there is a strong argument for ALWAYS showing.

Also, consider the "ASCII control" characters, partial list to remove doubt about what I mean:

image

They are ALWAYS shown:

image

Always showing the control characters is a strong argument (from N++'s history) to always show all characters with normally no visual appearance.

HOWEVER, there has been a user request to NOT show the control chars ( #8111 ).

So the bottom line is users will want different things.

But Notepad++ should assist user so they don't make "mistakes" with their data because they can't see things that really are present -- maybe THAT is the bottom line.

ozone10:
close to force appveyor to cancel builds

ozone10:
@Yaron10 
I have modified how `Show *` menu works.
Now they work independently.

I still think `Show All Characters` is useful for quickly show/hide all symbols. 

@donho 
> A code/text editor should ALWAYS show these non-printable characters, shouldn't it? 

People want to read great fiction, but half of text is filled with `U+200C` 😁.

Some languages use format character to display alternative forms
https://en.wikipedia.org/wiki/Zero-width_non-joiner.

Or text with combination of RTL and LTR text.
`U+202E : ‮ : right-to-left override : RLO`
`U+202E : [RLO] : right-to-left override : RLO`
Displaying representation will disable character effect on text.

ozone10:
I added custom color, currently only default themes have styles defined.
image

Yaron10:
@ozone10,

Everything seems to be working like a charm. Great job! 👍 

> I have modified how Show * menu works.
Now they work independently.
>
>I still think Show All Characters is useful for quickly show/hide all symbols.

👍 
I hope Don and other users like the new implementation as much as I do. :)

> Toggling the command or changing the preference does not take effect until clicking in the editing area.

@donho's solution (adding `redraw();`) seems to fix the issue in both menu and preferences. 👍 

***

Two *minor* GUI comments:

1. Wouldn't it better to remove the separator before `Show All Characters`? - Currently it's not clear the command is only referring to the three commands above and NOT to the two commands below.

2. This is mini-minor (and can be ignored): wouldn't it be nicer to place the `EOL (CRLF)` and `Non-Printing Characters` settings in the Preferences dialog next to each other?

***

What's the "more appropriate" term - Symbol or Character?
We currently have `Show Symbol`, `Non-Printing Characters`, `Show All Characters` and then `Wrap Symbol`.

Thank you.

alankilborn:
>People want to read great fiction, but half of text is filled with U+200C

Being able to turn off these characters makes sense in a "reading" situation.

During active editing, however, I'm uncomfortable with changing text without being able to "visualize" all of the data.
Of course, I realize that this is the current situation in released N++.  :-)

pryrt:
> Being able to turn off these characters makes sense in a "reading" situation.

I agree.  Defaulting to showing is fine with me, but some people do large swathes of reading, and for those use-cases, being able to turn them off easily is important.

> What's the "more appropriate" term - Symbol or Character?
> We currently have `Show Symbol`, `Non-Printing Characters`, `Show All Characters` and then `Wrap Symbol`.

In my interpretation:

The "Symbol" is what is visible -- whether it's the middot for space or arrow for tab or the curled arrow for the line-wrap function or the inverted CR and LF and NBSP boxes.

The "Character" is what's in the underlying data -- whether it's the space or tab character or the carriage return or line feed or non-breaking-space character.

Since the menu is **Show Symbols**, I've always thought of it as having an implied "symbol for": **Show Symbols > Show _Symbol For_ Space and Tab Characters** and the like, which gets it to pass my pedantry filter.  But it wouldn't bother me if all (actually, I might prefer it) if these were:
 - Show Space and Tab Symbols
 - Show End of Line Symbols
 - Show Non-printing Symbols
 - Show All Symbols

ozone10:
I have adjusted abbreviation based on https://en.wikipedia.org/wiki/List_of_Unicode_characters#General_Punctuation
image

There are also other non-printing characters specific to language (Arabic, Korean, ...), but unless requested by users, they don't need to be added for now.

@Yaron10
image

Yaron10:
@pryrt,

> In my interpretation:...

Thank you for the good-common-sense analyses. 👍 
Needless to say, I agree with your conclusion.

@ozone10,

Thank you for the last improvements as well. I do appreciate it.
Again - great work. 👍

donho:
@ozone10 
> I have adjusted abbreviation based on https://en.wikipedia.org/wiki/List_of_Unicode_characters#General_Punctuation image
> 
> There are also other non-printing characters specific to language (Arabic, Korean, ...), but unless requested by users, they don't need to be added for now.

So that's the table we should put in https://npp-user-manual.org/
Do you confirm?

@pryrt 
In which section we should put this information? Could you provide the URL please?

Yaron10:
@ozone10,

May I ask you when `npcSync` should be `no`?

Thank you.

pryrt:
> @pryrt
> In which section we should put this information? Could you provide the URL please?

@donho , it will be in the https://npp-user-manual.org/docs/views/#show-symbol section

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -1058,6 +1119,44 @@ intptr_t CALLBACK EditingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
...
+					}
+
+					HWND grandParent = ::GetParent(_hParent);
+					::SendMessage(grandParent, NPPM_INTERNAL_SETNPC, 0, 0);
+					return TRUE;
+				}
+
+				case IDC_BUTTON_NPC_NOTE:
+				{
+					::ShellExecute(NULL, TEXT("open"), TEXT("https://npp-user-manual.org/"), NULL, NULL, SW_SHOWNORMAL);
```
donho, line 1141: Please use `https://npp-user-manual.org/docs/views/#show-symbol` instead.

donho:
@Yaron10 
> May I ask you when `npcSync` should be `no`?

Well caught.
npcSync is initialized as yes if this argument is absent. Itwill never be modified except user changes it manually in config.xml.
I suppose it was for "sync with show all" option initially but it's forgotten to be removed.

PowerEditor/src/Parameters.h
```diff
@@ -943,6 +943,11 @@ struct ScintillaViewParams
 	bool _eolShow = false;
 	enum crlfMode {plainText = 0, roundedRectangleText = 1, plainTextCustomColor = 2, roundedRectangleTextCustomColor = 3};
 	crlfMode _eolMode = roundedRectangleText;
+	bool _npcShow = false;
+	enum npcMode { identity = 0, abbreviation = 1, codepoint = 2 };
+	npcMode _npcMode = abbreviation;
+	bool _npcSync = false;
```
donho, line 949: Remove _npcSync since its GUI option is no more there.

donho (changes requested):
1. Please follow the suggestions.
2. Add **NPC custom color** node in `bool NppParameters::feedStylerArray(TiXmlNode *node)` similar as  **EOL custom color** node (so "NPC custom color" node will be always available, even the styler.xml is not updated):
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Parameters.cpp#L3886

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -70,36 +70,47 @@ IDD_PREFERENCE_SUB_EDITING DIALOGEX 115, 10, 460, 205
...
+    CONTROL         "Custom Color",IDC_CHECK_WITHCUSTOMCOLOR_CRLF,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,177,47,100,10
+    PUSHBUTTON      "...",IDC_BUTTON_LAUNCHSTYLECONF_CRLF,281,44,16,14
+
+    GROUPBOX        "Line Wrap",IDC_LW_GB_STATIC,311,3,80,59,BS_CENTER
+    CONTROL         "Default",IDC_RADIO_LWDEF,"Button",BS_AUTORADIOBUTTON | WS_GROUP,317,16,68,10
+    CONTROL         "Aligned",IDC_RADIO_LWALIGN,"Button",BS_AUTORADIOBUTTON,317,31,68,10
+    CONTROL         "Indent",IDC_RADIO_LWINDENT,"Button",BS_AUTORADIOBUTTON,317,46,68,10
+
+    CONTROL         "Enable smooth font",IDC_CHECK_SMOOTHFONT,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,173,81,270,10
+    CONTROL         "Enable virtual space",IDC_CHECK_VIRTUALSPACE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,173,96,266,10
```
donho, line 108: BTW, there's a forgotten one :)

ozone10:
@Yaron10
> May I ask you when `npcSync` should be `no`?

It was supposed to be hidden option for users who don't want change how `Show All Characters` currently works (show/hide space, tab and EOL).

But unless there is demand for it I have removed it for now.

@donho 
> So that's the table we should put in https://npp-user-manual.org/ ~Do you confirm?~ If it's, could you provide the code in Markdown or HTML please?

This was only for reference. This PR also include other non-printing and whitespace characters. 
Table should be based on https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13020#issue-1565611383

ozone10:
@donho 
| Name                         | Codepoint | Abbreviation |
|------------------------------|-----------|--------------|
| No-Break Space               | U+00A0    | NBSP         |
| Ogham Space Mark             | U+1680    | OSPM         |
| Mongolian Vowel Separator    | U+180E    | MVS          |
| En Quad                      | U+2000    | NQSP         |
| Em Quad                      | U+2001    | MQSP         |
| En Space                     | U+2002    | ENSP         |
| Em Space                     | U+2003    | EMSP         |
| Three-Per-Em Space           | U+2004    | 3/MSP        |
| Four-Per-Em Space            | U+2005    | 4/MSP        |
| Six-Per-Em Space             | U+2006    | 6/MSP        |
| Figure Space                 | U+2007    | FSP          |
| Punctation Space             | U+2008    | PSP          |
| Thin Space                   | U+2009    | THSP         |
| Hair Space                   | U+200A    | HSP          |
| Zero-Width Space             | U+200B    | ZWSP         |
| Zero-Width Non-Joiner        | U+200C    | ZWNJ         |
| Zero-Width Joiner            | U+200D    | ZWJ          |
| Left-To-Right Mark           | U+200E    | LRM          |
| Right-To-Left Mark           | U+200F    | RLM          |
| Line Separator               | U+2028    | LS           |
| Paragraph Separator          | U+2029    | PS           |
| Left-To-Right Embedding      | U+202A    | LRE          |
| Right-To-Left Embedding      | U+202B    | RLE          |
| Pop Directional Formatting   | U+202C    | PDF          |
| Left-To-Right Override       | U+202D    | LRO          |
| Right-To-Left Override       | U+202E    | RLO          |
| Narrow No-Break Space        | U+202F    | NNBSP        |
| Medium Mathematical Space    | U+205F    | MMSP         |
| Word Joiner                  | U+2060    | WJ           |
| Left-To-Right Isolate        | U+2066    | LRI          |
| Right-To-Left Isolate        | U+2067    | RLI          |
| First Strong Isolate         | U+2068    | FSI          |
| Pop Directional Isolate      | U+2069    | PDI          |
| Inhibit Symmetric Swapping   | U+206A    | ISS          |
| Activate Symmetric Swapping  | U+206B    | ASS          |
| Inhibit Arabic Form Shaping  | U+206C    | IAFS         |
| Activate Arabic Form Shaping | U+206D    | AAFS         |
| National Digit Shapes        | U+206E    | NADS         |
| Nominal Digit Shapes         | U+206F    | NODS         |
| Ideographic Space            | U+3000    | IDSP         |
| Zero-Width No-Break Space    | U+FEFF    | ZWNBSP       |

donho:
@ozone10 
[[xml] Update default themes](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13020/commits/042fbc0f9648c7f47cedc448c422ccb573083c24) is failed due to `[xml]`:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/46140254

Could you do a new force push please ?

Yaron10:
@donho & @ozone10,

> npcSync is initialized as yes...

> It was supposed to be hidden option...

Thank you both. 👍

ozone10:
> failed due to `[xml]`
> 
> Could you do a new force push please ?

I thought commit with changes only to xml should use `[xml]` in commit name.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -924,6 +925,10 @@ The comments are here for explanation, it's not necessary to translate them.
                     <Item id="6248" name="Default"/>
                     <Item id="6249" name="Plain Text"/>
                     <Item id="6250" name="Custom Color"/>
+                    <Item id="6252" name="Non-Printing Characters"/>
+                    <Item id="6254" name="Abbreviation"/>
+                    <Item id="6255" name="Codepoint"/>
+                    <Item id="6256" name="Sync with &quot;Show All Character&quot;"/>
```
Yaron10, line 931: `Sync with` has been removed.

Thank you.

donho (approved):

alankilborn:
Can this be made the *default* setting?:

image

This is such a valuable new feature, but if it is not on-by-default, 90+% of Notepad++ users won't ever know about it and won't be able to benefit from it (the typical user upgrades all the time, but doesn't read release notes).

Make it on by default, let those that complain figure out or ask how to turn it off (if they want, although turning it off is usually not a smart move).

alankilborn:
@ozone10 @donho

When _Show Non-Printing characters_ is enabled, how should they appear in _Search results_?

image

donho:
Hmm... Maybe it's better that Search result' Scintilla sync with settings.
@alankilborn Could you create an issue for that please?

alankilborn:
>Could you create an issue for that please?

@ozone10 's PR made it in before the issue could be created.

PR #13025, opened 2023-02-01
[xml] Update swedish.xml

Regnander:
Added new strings since my last PR in October.


donho:
Could you please check the following announcement for completing your localization file if it's not done?
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13209

Thank you @Regnander

Regnander:
Done! See #13220.

PR #13034, opened 2023-02-02
Add BeginEnd Select in Column Mode command

### LINKED ISSUE #12832: Assign default shortcut keycombo of Ctrl+Shift+b to Begin/End Select command
### Description of the Issue
The very useful command _Edit_ (menu) > _Begin/End Select_ has no default key binding.
Suggest defaulting the currently unused Ctrl+Shift+b to this command.  (NOTE: Ctrl+b might be better for this function, but it is already used in the default keymapping supplied with N++).

Usage of this command comes up frequently on the support site, for users seeking to select a lot of text without long scrolling with a keyboard or mouse selection.  It would be nice to tell users to use a keycombo to start/end their selection, instead of saying "go into the _Edit_ menu and choose _Begin/End Select_, then move your caret, then go back into the _Edit_ menu and choose _Begin/End Select_ again".

### Debug Information
Feature request based on default key assignments in Notepad++ 8.4.8

### LINKED ISSUE #12833: Add new command and default keycombo for Begin/End Select in column mode
### Description of the Issue

-----------------------------------------

PREAMBLE NOTE: If #12832 is adopted, this would be a nice addition as well.  If 12832 is NOT adopted, this probably doesn't make sense to implement on its own.

-----------------------------------------

Currently Notepad++ ships with one sample macro, _Trim Trailing Space and Save_.  I'd like to suggest a second one be created and shipped with Notepad++ releases.

I'd call this macro "Begin/End Select in Column Mode" and I would suggest a default key assignment of Alt+Shift+b.  The Alt+Shift part reminds the user of the column-mode nature of running the macro -- because the user is used to using Alt+Shift+arrows to make a keyboard column selection.

Here is the suggested macro content:

```
<Macro name="Begin/End Select in Column Mode" Ctrl="no" Alt="yes" Shift="yes" Key="66">
    <Action type="0" message="2429" wParam="0" lParam="0" sParam="" />
    <Action type="0" message="2428" wParam="0" lParam="0" sParam="" />
    <Action type="2" message="0" wParam="42020" lParam="0" sParam="" />
</Macro>
```

----------------------------------------

Here's how the suggested entry would appear on the _Macro_ menu:

image

----------------------------------------

Here's a usage example, for clarity:

1. Open Notepad++'s `license.txt` file.
2. Move the caret to line 1, column 5.
3. Invoke the (suggested) Shift+Alt+b keycombo (to begin-select).
4. Move the caret to the last line of the file, column 20.
5. Invoke the (suggested) Shift+Alt+b keycombo a second time (to end-select).
6. Observe the (longish) column block selection is made:

image

------------------------------------------

NOTE: This issue has its basis in discussions in this Community thread: https://community.notepad-plus-plus.org/topic/20616

-----------------------------------------

### Debug Information
Feature request based on current content of Notepad++ 8.4.8.

alankilborn:
Fix #12833 
Fix #12832


Yaron10:
@alankilborn,

Thank you for your work. 👍 :)

Yaron10:
@alankilborn,

Shouldn't you modify some values in `localization.cpp`?

If you use the old localization files, the menu is messed up.

תמונה

תמונה

Thank you.

donho:
@Yaron10 
> Shouldn't you modify some values in `localization.cpp`?

In the old system (long time ago) adding a new menu item need to modify the localization files - but it was not reasonable because once we changed the menu structures we have had to modify near 100 localization file.

Now adding a new menu item should just adjust in the source code so the translation keeps working without touching any localization file.

@alankilborn 
You should look at here:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/localization.cpp#L28

Yaron10:
@donho,

I've suggested to modify **localization.cpp**. Not the XML localization files.

alankilborn:
@Yaron10  @donho 

Thanks for the hint about `localization.cpp` !  

I had no idea that this type of change was necessary.

I made the change and tested the submenus below my new command on the _Edit_ menu.

Yaron10:
@alankilborn,

Great! 👍 

Thanks again.

donho:
> @donho,
> 
> I've suggested to modify **localization.cpp**. Not the XML localization files.

Oups! Sorry my bad!

Issue #13053, opened 2023-02-05
[Regression] Document Switcher last row is cut off


ozone10:
### Description of the Issue

Reference https://community.notepad-plus-plus.org/post/83819
### Steps to Reproduce the Issue

1. Have at least 2 open tab
2. Activate document switcher (Ctrl+TAB)

### Expected Behavior

Document switcher show fully 2 rows.
### Actual Behavior

Document switcher last row is cut off

### Debug Information

Notepad++ v8.4.9

PR #13054, opened 2023-02-05
Fix Doc Switcher cut off

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13053

### DIFF
--- a/PowerEditor/src/WinControls/TaskList/TaskList.cpp
+++ b/PowerEditor/src/WinControls/TaskList/TaskList.cpp
@@ -95,7 +95,8 @@ RECT TaskList::adjustSize()
 	ListView_GetItemRect(_hSelf, 0, &rc, LVIR_ICON);
 	const int imgWidth = rc.right - rc.left;
 	const int aSpaceWidth = ListView_GetStringWidth(_hSelf, TEXT(" "));
-	const int leftMarge = (::GetSystemMetrics(SM_CXFRAME) + ::GetSystemMetrics(SM_CXPADDEDBORDER)) * 2 + aSpaceWidth * 4;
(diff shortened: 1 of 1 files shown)


ozone10:
`::GetSystemMetrics(SM_CXFRAME)` return different value depending on minimum subsystem version. Windows Vista and newer needs to add `::GetSystemMetrics(SM_CXPADDEDBORDER)` to keep same value as subsystem version for Windows XP.

TaleTN:
FWIW, I would store ::GetSystemMetrics(SM_CXPADDEDBORDER) in a local variable, so it can be reused rather than call it twice i.e.:

`const int paddedBorder = ::GetSystemMetrics(SM_CXPADDEDBORDER);`
`const int leftMarge = (::GetSystemMetrics(SM_CXFRAME) + paddedBorder) * 2 + aSpaceWidth * 4;`
...
`_rc.bottom += (::GetSystemMetrics(SM_CYFRAME) + paddedBorder - 1) * 2;`

donho:
It's indeed a regression - v8.4.8 works fine but not for v8.4.9.

@ozone10 
Do you know where comes from this regression?

ozone10:
image

Previously it was set 5.01 for 32 bit and 5.02 for 64 bit both values for Windows XP. 
Value 6.00 is for Windows Vista.

This setting affect `::GetSystemMetrics(SM_CXFRAME)` as I said in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13054#issuecomment-1418259879

It comes from commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/662b54e28a5ad2787ba198082fc4c23a69849367

donho:
@ozone10 

Issue #12325, opened 2022-10-12
Auto-Completion for Normal Text doesn't work immediately


pryrt:
### Description of the Issue

When I have `autoCompletion\normal.xml` defined with `function` and `word`, Notepad++ doesn't auto-complete those in Normal Text files (*.txt or *.log) right away.  However, if I change the language to something else (like `XML` and then back to `None (Normal Text)`), then the auto-completions for Normal Text start working, and any new files opened or created after that

### Steps to Reproduce the Issue

0. Fresh unzip of portable v8.4.6 and run
1. `Settings > Preferences > Auto-Completion`: confirm
   - `> From _N_th character` = `1`
   - `Enable auto-completion on each input` = `Function and word completion`
2. Create new file:
    ```
    <?xml version="1.0" encoding="UTF-8" ?>
    <NotepadPlus>
        <AutoComplete language="Normal">
            <Environment ignoreCase="no" />
            <KeyWord name="function" func="yes" />
            <KeyWord name="word" />
        </AutoComplete>
    </NotepadPlus>
    ```
3. SaveAs `<portableDirectory>\autoCompletion\normal.xml`
4. **File > New**
5. Type `f` or `w` and it doesn't suggest anything
6. **Language > XML** then **Language > None (Normal Text)**
7. type `f` and it suggests `function`, or type `w` and it suggests `word`
   - at first I thought the difference between step 5 and step 7 was because normal.xml was new, and the toggle caused it to reload the normal.xml
8. from now on, any **File > New** followed by typing _will_ suggest the words/functions from `normal.xml`
9. **File > Close All**
10. Exit Notepad++ and restart
11. type `f` or `w` and it doesn't suggest anything
    - here, it proves that even if normal.xml already exists, Notepad++ doesn't immediately start autocompleting on normal text files
12. **Language > XML** then **Language > None (Normal Text)**
13. type `f` and it suggests `function`, or type `w` and it suggests `word`
    - but triggering a different language will allow Notepad++ to autocomplete on plain text files
14. from now on, any **File > New** followed by typing _will_ suggest the words/functions from `normal.xml`
15. **File > Close All**
16. Exit Notepad++ and restart
17. **Language > Perl** (so `new 1` now shows up as Perl): don't need to type anything
18. **File > New** (so now you are editing `new 2` which is still a normal text file)
19. type `f` and it suggests `function`, or type `w` and it suggests `word`
[...]

PR #13062, opened 2023-02-06
Fix missing items in word autocomplete

### LINKED ISSUE #13060: Missing items in word autocomplete
### Description of the Issue

Autocomplete setting is **function autocomplete** so that **word complete** is manually tested. Select the **Lua** language which is case insensitive. Enter the following into the document.

```lua
aab = 1

function AAC()
end

a
```

Possible word autocomplete list after pressing <kbd>Ctrl</kbd>+<kbd>Enter</kbd> following the last `a` :

```
aab
```

```
AAC
```

```
AAC
aab
```

 1. Position the caret after the last `a`, press <kbd>Ctrl</kbd>+<kbd>Enter</kbd>, `aab` is auto-inserted
 2. Undo and remove last `a`
 3. Type `A`, press <kbd>Ctrl</kbd>+<kbd>Enter</kbd>, `AAC` is auto-inserted

### Expected Behavior

An autocomplete list displaying

```
AAC
aab
```

as case insensitive `A` is considered the same as `a`.

This issue causes words expected in the list to be missing in the list. 

### Actual Behavior

Auto-inserts `aab` if `a` is 1st character or auto-inserts `AAC` if `A` is 1st character.

### Debug Information

```
Notepad++ v8.4.9   (64-bit)
Path : C:\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

mpheath:
Fix #13060

Word autocomplete change from case sensitive search to variable search based on `_ignoreCase` unless language type is text. This allows text documents to retain case sensitive search to respect typed case as the issue only affects case insensitive Data/Programming languages. If *normal.xml* exist in the autocomplete folder then xml defaults apply for text so would be case insensitive by default unless overridden by the xml setting `<Environment ignoreCase="no" />`.

If **full** variable search based on `_ignoreCase` is considered more suitable by including text, then a change is possible.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
+++ b/PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
@@ -458,7 +458,10 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | SCFIND_REGEXP | SCFIND_POSIX;
+
+	if (!_ignoreCase || (!_funcCompletionActive && _curLang == L_TEXT))
+		flags |= SCFIND_MATCHCASE;
 
 	_pEditView->execute(SCI_SETSEARCHFLAGS, flags);
 	intptr_t posFind = _pEditView->searchInTarget(expr.c_str(), expr.length(), 0, docLength);


PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
```diff
@@ -458,7 +458,10 @@ void AutoCompletion::getWordArray(vector<generic_string> & wordArray, TCHAR *beg
 
 	size_t docLength = _pEditView->execute(SCI_GETLENGTH);
 
-	int flags = SCFIND_WORDSTART | SCFIND_MATCHCASE | SCFIND_REGEXP | SCFIND_POSIX;
+	int flags = SCFIND_WORDSTART | SCFIND_REGEXP | SCFIND_POSIX;
+
+	if (!_ignoreCase || (!_funcCompletionActive && _curLang == L_TEXT))
```
donho, line 463: Hmm... if current lang is L_TEXT, `_funcCompletionActive` should be false, since there's no function xml to be load.
Is `!_funcCompletionActive` condition necessary?

donho (changes requested):
Is !_funcCompletionActive condition necessary?

PowerEditor/src/ScintillaComponent/AutoCompletion.cpp
mpheath: If *normal.xml* exist, then `_funcCompletionActive` will be true so then the xml settings will apply. If `_funcCompletionActive ` is not used there, then the xml settings will be overruled there and text would be always searched as case sensitive. Having `_funcCompletionActive` there prevents *normal.xml* Environment options being ignored.

*normal.xml* is mentioned [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12325#issue-1406561749) which made me aware of *normal.xml* .

mpheath: If the condition seems too confusing, then could change to only

```c++
int flags = SCFIND_WORDSTART | (_ignoreCase ? SCFIND_NONE : SCFIND_MATCHCASE) | SCFIND_REGEXP | SCFIND_POSIX;
```

though text would be case insensitive by default. I was treating text as special so that both `The` and `the` kind of word variations does not polute the list.

donho: @mpheath 
You're right about it. I forgot "normal.xml" case could happen.
So your original version is good. Nothing to be changed.

donho (approved):

PR #13072, opened 2023-02-07
Explicitly handle buffers with no extension

### LINKED ISSUE #13066: "New Document" language setting does not apply anymore
### Description of the Issue
In earlier versions of Notepad++, we could provide a default language for "new documents" which are files without an extension. This is still possible, but the setting seems to be ignored. We usually use this to quickly open c++ files w/o extension. We are then readily able to use block comments or uncomment blocks. Somewhere in recent versions this functionality was lost. We now have to set the language first each time we open a "new document".

1. Create a file w/o extension, save.
2. Open the file.
3. No syntax highlighting or functionality like block comments are available-

### Expected Behavior
Block comments readily available.

### Actual Behavior
Language is not recognized, no specific functionality available.

### Debug Information
Notepad++ v8.4.9   (32-bit)
Build time : Jan 27 2023 - 03:08:31
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.2486
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

rdipardo:
### Before

1. Set a preferred default language (e.g. C++) and open an empty file with no extension
2. `Buffer::setFileName` [looks for an extension][0], finds none, and exits without affecting the `_lang` member 
3. The file opens as plain text

### After

1. Set a preferred default language (e.g. C++) and open an empty file with no extension
2. `Buffer::setFileName` looks for an extension, finds none
3. `Buffer::setFileName` *notices that no extension is present*; now it will:

   - choose the auto-detected language if it finds a shebang/doctype
 
     *Note* this requires public visibility of `FileManager::detectLanguageFromTextBegining`

   - if that fails, choose the preferred default language

3. The file opens as the preferred default language

Fixes #13066

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f38195a0da5657f447ccc949bc77f38dba06aa61/PowerEditor/src/ScintillaComponent/Buffer.cpp#L197-L214


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13066#issuecomment-1462552710

Issue #12689, opened 2022-12-22
[Feature Request] Add a separator between two views in the Document List


Phroneris:
### Description of the Issue

I have 43 documents opening in Notepad++ now.
33 on the main (left) view and 10 on the sub (right).

Within the **Document List** a.k.a. Doc Switcher, I would like a **separator** (or something like that) between the main and sub file groups for visual organization.

<img width="250" height="450" title="click for full size" src="https://user-images.githubusercontent.com/9891867/209113145-d51716c2-e847-47fc-951f-a3b322e23c06.png">

### Steps to Reproduce the Issue

Because this isn't an issue, here is just an example situation.

1. Turn on [View] → [Document List]
2. Open many files on the main and sub view
3. Go to each file around the middle of the order on each view
4. In the Document List, try to go to the last file of the main view

... But sometimes the list order goes sub → main.
If that happens, try to find the last of sub instead.

### Expected Behavior

The separator indicates the boundary of two views so that you can easily find the desired file.

### Actual Behavior

You can't easily find the desired file unless you remember its name or your random clicks hit on it.

### Debug Information

- Notepad++ v8.4.7 (64-bit)
- Windows 10 (64-bit)

TBugReporter:
On a semi-related note (which I'm sure some here will see as me being too picky - which is why I'm tacking this on to this FR instead of creating yet another new one), it would also be nice if the Window menu would list the windows in **both** panes instead of just the one that's currently active.

alankilborn:
> On a semi-related note (which I'm sure some here will see as me being too picky - which is why I'm tacking this on to this FR instead of creating yet another new one), it would also be nice if the Window menu would list the windows in **both** panes instead of just the one that's currently active.

It's certainly NOT "too picky".
It is something I suggested when the "total documents: xx" was added to the title bar of the popup window when one chooses _Window_ > _Windows..._.  Or at least I think I suggested it -- I know I commented that if you have N documents open in the primary view and M docs open in the secondary view, then you have M+N total documents...
But it is off-topic to this current issue and deserves its own; I encourage you to create one (after searching to see if there is already one existing).

TBugReporter:
Found:  [Show open documents from both views in Window menu and ▼ #11764](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11764).

[...]

Issue #12155, opened 2022-09-12
Apply color from individual tab also in Document List panel


ArkadiuszMichalski:
Apply color from individual tab also in Document List panel. If necessary this can be controlled in the Preferences window or in the context menu of the Document List panel.

![image](https://user-images.githubusercontent.com/2730894/189631405-7022a57e-0a4d-4a28-b5f5-6bf0891a40d3.png)

PR #13080, opened 2023-02-09
Apply tab colors to document list items

ozone10:
add groups to document list

Groups were added because of sorting issue when applying colors.
Unfortunately group header color cannot be changed ([crHeader Not Implemented](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/ns-commctrl-lvgroupmetrics)).

image

Needs some testing and feedback.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12155
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12689

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -414,10 +414,6 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// VOID NPPM_DOCLISTDISABLEPATHCOLUMN(0, BOOL disableOrNot)
 	// Disable or enable path column of Document List
 
-	#define NPPM_DOCLISTDISABLEGROUPS    (NPPMSG + 112)
-	// VOID NPPM_DOCLISTDISABLEGROUPS(0, BOOL disableOrNot)
-	// Disable or enable groups of Document List
-
 	#define NPPM_GETEDITORDEFAULTFOREGROUNDCOLOR    (NPPMSG + 90)
 	// INT NPPM_GETEDITORDEFAULTFOREGROUNDCOLOR(0, 0)
 	// Return: current editor default foreground color. You should convert the returned value in COLORREF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -414,6 +414,10 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// VOID NPPM_DOCLISTDISABLEPATHCOLUMN(0, BOOL disableOrNot)
 	// Disable or enable path column of Document List
 
+	#define NPPM_DOCLISTDISABLEGROUPS    (NPPMSG + 112)
+	// VOID NPPM_DOCLISTDISABLEGROUPS(0, BOOL disableOrNot)
+	// Disable or enable groups of Document List
+
 	#define NPPM_GETEDITORDEFAULTFOREGROUNDCOLOR    (NPPMSG + 90)
 	// INT NPPM_GETEDITORDEFAULTFOREGROUNDCOLOR(0, 0)
 	// Return: current editor default foreground color. You should convert the returned value in COLORREF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -999,6 +999,7 @@ void Notepad_plus::command(int id)
 			const auto current_index = _pDocTab->getCurrentTabIndex();
 			BufferID buffer_id = _pDocTab->getBufferByIndex(current_index);
 			_pDocTab->setIndividualTabColour(buffer_id, color_id);
+			_pDocTab->redraw();
 			::SendMessage(_pPublicInterface->getHSelf(), WM_SIZE, 0, 0);
 
 			if (_pDocumentListPanel != nullptr)
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1000,6 +1000,11 @@ void Notepad_plus::command(int id)
 			BufferID buffer_id = _pDocTab->getBufferByIndex(current_index);
 			_pDocTab->setIndividualTabColour(buffer_id, color_id);
 			::SendMessage(_pPublicInterface->getHSelf(), WM_SIZE, 0, 0);
+
+			if (_pDocumentListPanel != nullptr)
+			{
+				_pDocumentListPanel->setItemColor(buffer_id);
+			}
 		}
 		break;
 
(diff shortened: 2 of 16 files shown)


ArkadiuszMichalski:
Works nice.

@ozone10 

It's possible to add `Compact view` or `Compact mode` switch where we reduce some space between elements? Original `Window Manager` plugin has less space between each docs so we see more docs (in my monitor I see 5 files more in `Window Manager` plugin).  

image

This `Compact mode` could also include these spacing for separators between views (because they are quite large).

image

BTW, `Window Manager` plugin also use separate list for both views but slightly different.

image
It also has a horizontal scrollbar, which is sometimes useful (if we don't see all the headers).

ozone10:
@Yaron10
I can separate PR but then I would need to solve merge conflict as both features are modifying same "parts" of code (Custom draw parts). 

About disabling, it's not worth it, I would need to deal with several situations: group setting disabled/enabled,  other view exist, ...

@ArkadiuszMichalski 
Can you create issue. I don't want to make this PR more complicated.

I will try to figure out how to make `Compact mode`.

Current Document list also have horizontal scrollbar.

ozone10:
Currently missing translation for menu item "Group by View", I will add it later.

Yaron10:
@ozone10,

> I can separate PR but then I would need to solve merge conflict as both features are modifying same "parts" of code (Custom draw parts).
>
>About disabling, it's not worth it...

It's great as it is.

How about adding a separator before `Group by View`?

PowerEditor/src/Parameters.h
```diff
@@ -897,6 +897,7 @@ struct NppGUI final
 	int _fileSwitcherExtWidth = 50;
 	bool _fileSwitcherWithoutPathColumn = true;
 	int _fileSwitcherPathWidth = 50;
+	bool _fileSwitcherDisableGroups = true;
```
Yaron10, line 900: Why should it be defaulted to `true`?
If you delete your config.xml, Groups would be disabled.

`_fileSwitcherWithoutExtColumn` was *mistakenly* changed to `true` when `_fileSwitcherWithoutPathColumn` was added.

PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcher.cpp
```diff
@@ -65,6 +68,129 @@ void VerticalFileSwitcher::startColumnSort()
...
+
+		case WM_NOTIFY:
+		{
+			auto nmhdr = reinterpret_cast<LPNMHDR>(lParam);
+			switch (nmhdr->code)
+			{
+				case NM_CUSTOMDRAW:
+				{
+					constexpr size_t classNameLen = 16;
+					TCHAR className[classNameLen]{};
```
Yaron10, line 173: Should it be `{}` or `{ '\0' }`?

ozone10: I usually use ` = { '\0' }` for "char" array, But it can be initialized with `{}`,`= {}`, ...

ozone10:
Added translation, added separator to menu.

ArkadiuszMichalski:
@ozone10 
>Current Document list also have horizontal scrollbar.

But it seems to only appear when the vertical appears. If we turn on `Ext.` and `Path` and our panel is not wide the horizontal bar does not appear (so we can't e.g. check multiple paths at once without resizing the panel).

image

Yaron10:
@ozone10,

Great. 👍 
Thanks again.

ozone10:
@ArkadiuszMichalski 
try to resize Path column.

ozone10:
Fix merge conflict.

donho:
I have got an unsync bug:

image

You can see the english.xml is not the same colour.
I don't know yet how to reproduce it. I will provide more info when I reproduce it.

donho (changes requested):
Please follow the suggestions

PowerEditor/src/Parameters.h
```diff
@@ -893,10 +893,11 @@ struct NppGUI final
 	bool _docMapKeepState = false;
 	bool _funcListKeepState = false;
 	bool _pluginPanelKeepState = false;
-	bool _fileSwitcherWithoutExtColumn = true;
+	bool _fileSwitcherWithoutExtColumn = false;
```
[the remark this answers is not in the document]
Yaron10: @donho,

Please see:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13080#discussion_r1103133486
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10539/files#r706698296
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/36d0d29cc756017bff911231320c4557b03f9d32#diff-d2f976c2d1c54b02bc8170fb05a99a96e30c9554abbdb4de60fe39d57b437406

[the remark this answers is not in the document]
Yaron10: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/156784722a46fd250128ef415d6813a8171de5c7/PowerEditor/src/Parameters.h

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -414,6 +414,10 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// VOID NPPM_DOCLISTDISABLEPATHCOLUMN(0, BOOL disableOrNot)
 	// Disable or enable path column of Document List
 
+	#define NPPM_DOCLISTDISABLEGROUPS    (NPPMSG + 112)
```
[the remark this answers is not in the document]
ozone10: ok I have removed it

PowerEditor/src/NppCommands.cpp
```diff
@@ -999,6 +999,7 @@ void Notepad_plus::command(int id)
 			const auto current_index = _pDocTab->getCurrentTabIndex();
 			BufferID buffer_id = _pDocTab->getBufferByIndex(current_index);
 			_pDocTab->setIndividualTabColour(buffer_id, color_id);
+			_pDocTab->redraw();
```
ozone10, line 1002: this should help with sync issue

donho: @ozone10 
Not really. Here's the way to reproduce (with your latest commit):
1. open 2 document 1.txt & 2.txt in the main view
2.  color 2.txt tab to green - we can see green color for 2.txt on both list & tab 
3. drag&drop 2.txt to sub view -  green color for 2.txt disappear on both list & tab (it's arguable if it's a bug, but...) 
4. drag&drop 2.txt back to main view -  green color is back on tab, but not on list (that's a bug for sure)

ozone10: Ok this should fix it, there was bug with tab color implementation for subview.

donho: Tested, fixed. Thank you!

donho (approved):

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/37963ea21ad7876903c669e2190e6f2b2dde7e39

PR #2375, opened 2016-10-01
Hindi Translation for V7.0

SinghRajenM:
- Corrected/updated whole translation (used user friendly words)
- Synced with english.xml (easy to compare with diff tool such as Beyond compare)
- Update for v7.0 (added missing strings)
- Added/updated translation in such a way so that it should not run out of space for the control which has limited width

Issue #2374, opened 2016-10-01
Battery issue on background


varosi:
### Steps to Reproduce the Issue
1. Leave Notepad++ for longer time
2. Settings | Battery usage
3. It's on the first place of battery usage

### Expected Behavior
When on background to not use the battery so much.

### Actual Behavior
Use the battery in background.

Windows 10 
Notepad++ 6.9.2

![image](https://cloud.githubusercontent.com/assets/6051718/19013522/c9c02cda-87db-11e6-9e37-1f69152d5f52.png)

dail8859:
I'd try disabling your plugins first to make sure its not one of them.

jwallis43:
I have it in first place at 22%

Issue #2373, opened 2016-09-30
Feature Request: Option for *not* Auto-Populating Search Box


KAH777:
When doing a Find (and/or Replace), I understand that it can be handy to have the search box auto-fill the text you are searching for with what is already highlighted.  However, there are many times that I want to bring up the Find window and have it search for the last item I searched for (or, sometimes I will switch from doing a Find to doing a "Mark All" or what have you).

Would it be possible and practical to modify Notepad++ with one of the following two behaviors?
  1.  Add an option in Settings to allow users like myself to turn off auto-populate (this is my #1 choice).
  2.  If no text is highlighted, then do not auto-populate search box.

Thank you for your consideration.

Kevin

kashalot:
> there are many times that I want to bring up the Find window and have it search for the last item I searched for

Isn't that already the current behavior?

- When I call up the search or replace box with no text highlighted, it's populated with the last thing I searched for.
- The text in the Find box is pre-selected when the search window comes up, so if you want to search for something different you can just start typing and overwrite it.  Having a special setting to remove autofill wouldn't save you a single keystroke in this scenario.
- If you specifically need to repeat the previous search while you have another chunk of text highlighted (which seems odd, given that the search is going to discard your current highlighting, but let's roll with it), pressing the Down arrow scrolls through your search history, so you can get to the last search you performed with one keystroke.

KAH777:
My apologies.  Using the version I have at home, I see that you are correct.  I'm not sure how I was missing that, unless my version at work has a plugin installed that is causing this behavior (unlikely, though).

Either way, it is clearly behaving as you describe.  Sorry to waste your time.  I will close the "issue".

Kevin

imthenachoman:
It would still be a nice feature to not auto-populate search period. I know I can press the down arrow but from a productivity perspective, it slows me down having to move my right hand from the mouse to the down arrow. I'd rather it keep my previous search query.

Issue #2372, opened 2016-09-30
Feature Request: Sort Plugins Menu Items by Name


KAH777:
Hi.  Currently, it looks like the plugins listed under the Plugins menu item are sorted by DLL name ("tip of the hat" to Dail for that information).  Please see attached screen shot for an example.

Would you consider changing it to sort by menu item name?  This would make it more efficient to find the plugins that we need to use.

Thank you,

Kevin
![pluginslist](https://cloud.githubusercontent.com/assets/22547861/18999742/be1cb970-8704-11e6-9013-01f2cc98499a.png)

dail8859:
Pending pull request #2347

dail8859:
Closed by d64a525f3e7921b6bed735e4298e8e85f09c71c1

Gitoffthelawn:
Is this patch working for others?  In v7.9.5 32-bit portable, *Plugins* menuitems appear in seemingly random order (not alphabetic).

Gitoffthelawn:
I figured out the issue.  Some plugins provide a unique "friendly name" while others do not.  For example, NppConverter.dll provides the English "friendly" name "Converter", but Notepad++ sorts it under "N" for "NppConverter.dll".  This is a plugin distributed with Notepad++... LOL.

Npp needs to sort by the actual menu name being displayed if the intention is to sort by the actual menu name.

Please re-open, this is broken.

Gitoffthelawn:
Note the `JSON Viewer` and `Customize Toolbar` plugins display the same behaviour as the included `Converter` plugin, so they can be used for testing as well.

Gitoffthelawn:
@donho Let me know if you need more details.

PR #13082, opened 2023-02-09
Update to Scintilla 5.3.3 and Lexilla 5.2.2

chcg:
update to https://www.scintilla.org/scintilla533.zip with:

    Released 8 February 2023.
    Fix SCI_LINESJOIN bug where carriage returns were incorrectly retained. Bug #2372.
    Fix SCI_VERTICALCENTRECARET to update the vertical scroll position.
    When an autocompletion list is shown in response to SCN_CHARADDED, do not process character as fill-up or stop. This avoids closing immediately when a character may both trigger and finish autocompletion.
    On Cocoa fix character input bug where dotless 'i' and some other extended Latin characters could not be entered. The change also stops SCI_ASSIGNCMDKEY from working with these characters on Cocoa. Bug #2374.
    On GTK, support IME context. Feature #1476.
    On GTK on Win32, fix scrolling speed to not be too fast. Bug #2375.
    On Qt, fix indicator drawing past left of text pane over margin. Bug #2373, Bug #1956.
    On Qt, allow scrolling with mouse wheel when scroll bar hidden.

and https://www.scintilla.org/lexilla522.zip with

    Released 8 February 2023.
    C++: Fix keywords that start with non-ASCII. Also affects other lexers. Issue #130.
    Matlab: Include more prefix and suffix characters in numeric literals. Issue #120.
    Matlab: More accurate treatment of line ends inside strings. Matlab and Octave are different here. Issue #18.
    Modula-3: Don't treat identifier suffix that matches keyword as keyword. Issue #129.
    Modula-3: Fix endless loop in folder. Issue #128.
    Modula-3: Fix access to lines beyond document end in folder. Issue #131.
    Python: Don't highlight match and case as keywords in contexts where they probably aren't used as keywords. Pull request #122.
    X12: Support empty envelopes. Bug #2369.

- update CMakeLists.txt to latest changes within vcxproj file


Yaron10:
@chcg,

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12219.

Thank you.

chcg:
@Yaron10 Thanks for finding this merge issue. Fixed.

Yaron10:
@chcg,

Great. 👍 
Thank you.

PowerEditor/src/CMakeLists.txt
```diff
@@ -385,13 +385,13 @@ IF (WIN32)
 if ( MSVC )
 	#do not use for mingw builds
 	SET(CMAKE_CXX_FLAGS "/EHa /MP /W4")
-	SET(defs -DUNICODE -D_UNICODE -D_WIN32_WINNT=0x0600 -D_USE_64BIT_TIME_T -DTIXML_USE_STL -DTIXMLA_USE_STL -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_CRT_NON_CONFORMING_SWPRINTFS )
+	SET(defs -DUNICODE -D_UNICODE -D_WIN32_WINNT=_WIN32_WINNT_VISTA -D_USE_64BIT_TIME_T -DTIXML_USE_STL -DTIXMLA_USE_STL -DNOMINMAX -DOEMRESOURCE -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_CRT_NON_CONFORMING_SWPRINTFS -D_CRT_NON_CONFORMING_WCSTOK -D_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING )
```
ozone10, line 388: @chcg 
Just a question. Are both `-D_CRT_NON_CONFORMING_SWPRINTFS` and `-D_CRT_NON_CONFORMING_WCSTOK` needed? Notepad++ msvc builds don't use these macros anymore.

chcg: @ozone10 Thanks. No, should be inline with the settings of the visual studio build. Corrected.

PR #13091, opened 2023-02-10
Add new notification for lexer plugin

### LINKED ISSUE #12351: Enhancement: allow external lexer to query for buffer ID that is being lexed by one of its instances

donho:
The NPPN_EXTERNALLEXERBUFFER is sent to plugins with buffer ID (idFrom) when a new lexer is applied to the buffer in question.

Fix #12351


donho:
@ blu3mania
Could you test it please?

rdipardo:
> @ blu3mania Could you test it please?

"@ blu3mania" definitely won't; but maybe @blu3mania will.

PowerEditor/src/resource.h
```diff
@@ -644,6 +644,7 @@
 	#define NPPM_INTERNAL_CLEANSMARTHILITING            (NOTEPADPLUS_USER_INTERNAL + 68)
 	#define NPPM_INTERNAL_CLEANBRACEMATCH               (NOTEPADPLUS_USER_INTERNAL + 69)
 	#define NPPM_INTERNAL_WINDOWSSESSIONEXIT            (NOTEPADPLUS_USER_INTERNAL + 70)
+	#define NPPM_INTERNAL_EXTERNALLEXERBUFFER           (NOTEPADPLUS_USER_INTERNAL + 71)
```
ozone10, line 647: @donho 
appveyor failed
Could you rebase on latest upstream/master and change value?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -902,6 +902,8 @@ void ScintillaEditView::setExternalLexer(LangType typeDoc)
 		return;
 	execute(SCI_SETILEXER, 0, reinterpret_cast<LPARAM>(iLex5));
 
+	::SendMessage(_hParent, NPPM_INTERNAL_EXTERNALLEXERBUFFER, 0, 0);
```
blu3mania, line 905: Since NppBigSwitch is expecting lParam for this message, shouldn't it be sent?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2715,7 +2717,8 @@ void ScintillaEditView::performGlobalStyles()
 		edgeColor = pStyle->_fgColor;
 	}
 	execute(SCI_SETEDGECOLOUR, edgeColor);
-	::SendMessage(_hParent, NPPM_INTERNAL_EDGEMULTISETSIZE, 0, 0);
+
+	::SendMessage(_hParent, NPPM_INTERNAL_EDGEMULTISETSIZE, 0, (LPARAM)getCurrentBufferID());
```
blu3mania, line 2721: Not sure if you meant to send the buffer ID in NPPM_INTERNAL_EXTERNALLEXERBUFFER but accidentally changed this message instead? I don't think NppBigSwitch uses lParam for this message.

blu3mania (changes requested):

donho:
@blu3mania 
The issues you mentioned above should solved now.

blu3mania:
@donho Tested and it was working correctly. Thanks!

Issue #13099, opened 2023-02-12
[GCC] Add -Wextra flag


ozone10:
### Description of the Issue

Adding -Wextra flag to GCC debug builds could allow us to spot potential issues in code.  

### Debug Information

Notepad++ v8.4.9

PR #13100, opened 2023-02-12
Add -Wextra flag for GCC builds

ozone10:
fix ignored-qualifiers warnings
fix inconsistent-missing-override warnings
use -O3 flag for GCC builds to keep inline with scintilla and lexilla 
add clang default settings in props
add clang detection in debug info

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13099

### DIFF
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -53,13 +53,13 @@ SUBMAKEFLAGS := -O --no-print-directory
 ifeq "$(filter-out 0,$(DEBUG))" ""
 BUILD_TYPE := release
 BUILD_SUFFIX :=
-CXXFLAGS += -Os -Wconversion
+CXXFLAGS += -O3 -Wconversion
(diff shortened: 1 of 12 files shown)


chcg:
@ozone10 https://github.com/chcg/notepad-plus-plus/actions/runs/4155988574/jobs/7189459667 based on https://github.com/chcg/notepad-plus-plus/tree/gh_actions_test might be interesting for your adaptions for clang. In this case combined with the analysis step.

ozone10:
@chcg 
I have looked. Currently I've added only default settings for clang to make it work. 

There is no configuration for it. Users will need to switch toolset manually if they want to built Notepad++ with clang.
IMHO currently it is not needed to add clang configurations, only this basic settings should be sufficient.

btw
Actions builds are failing because clang is treating warnings as error, you can temporarily turn `<TreatWarningAsError>false</TreatWarningAsError>` in props to make it pass.

PR #13103, opened 2023-02-12
[xml] Update spanish.xml to v. 8.5.0

P0rsche-911:
Nueva traducción.
Sería interesante que para futuras traducciones se permitiera igualar los archivos y facilitar el trabajo.
La comparación de archivos sería menos confusa y más eficaz.

New translation.
It would be interesting if for future translations it would be possible to match the files and make the work easier.
Comparing files would be less confusing and more efficient.

### DIFF
--- a/PowerEditor/installer/nativeLang/spanish.xml
+++ b/PowerEditor/installer/nativeLang/spanish.xml
@@ -1,1490 +1,1522 @@
 <?xml version="1.0" encoding="utf-8"?>
-	<!--
+<!--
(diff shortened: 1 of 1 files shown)


donho:
@P0rsche-911 
The translation with diff everywhere will never be merged into master:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13103/files

I suggest you use GitHub's interface to modify your file to prevent translation from format changing.

Issue #9209, opened 2020-11-29
"Folder as Workspace" pane does not update network share folder changes


xaviermdq:
### Description of the Issue
When I use Folder as Workspace (FAW) in a network share (we edit in a server where we develop web pages) any folder change doesn't update FAW pane.

### Steps to Reproduce the Issue
1. Activate Folder as Workspace (FAW) in a network share folder (//Server/Folder1/Folder2 structure)
2. With Explorer or Total Commander, change a folder name, add file, delete file or rename (all file inside that folder: //Server/Folder1/Folder2)
3. NPP doesn't update FAW panel

### Debug Information
Notepad++ v7.9   (64-bit)
Build time : Sep 22 2020 - 03:19:04
Path : D:\PORTABLES\_Programacion\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Pro (64-bit) 
OS Version : 1909
OS Build : 18363.1049
Current ANSI codepage : 1252
Plugins : CodeAlignmentNpp.dll JSMinNPP.dll LanguageHelp.dll Linter.dll mimeTools.dll NppConverter.dll NppExport.dll NPPJSONViewer.dll Python Indent.dll XMLTools.dll

paolo-sz:
I got exactly the same issue.
Digging in a little bit more I found the issue is due to the fact that, at least in my case, only items in the root folder are notified by the ADD or REMOVE event, modification on the lower folder levels only generate MODIFY event on the containing folder item.
The MODIFY event is currently disabled by the specific filter parameter in the call to <code>::ReadDirectoryChangesW</code> function in ReadDirectoryChangePrivate.cpp file.
Only enabling the management of that event is not enough by itself and makes the things a lot more complicated. Indeed the MODIFY event is also sent after an ADD or REMOVE event. This makes difficult to detect when the MODIFY event need to be managed or it is redundant, unless removing completely the management of the ADD and REMOVE events and implement an internal search of what was changed following a MODIFY event.
Surely a manual update is a much more easy workaround.
[...]

PR #13118, opened 2023-02-13
Added manual refresh to FAW (#9209)

paolo-sz:
As by reported by issue #9209, FAW has some issues when the watched folder is on a network share.

I added a couple of right button menu to refresh single watched folder or all at once.

Just a note, some useful function in class FileBrowser was already present but never used anywhere, I had the need to modify them a little bit.

Also, watcher thread only add/remove/rename from "GUI" tree but do not update accordingly the FolderInfo/FileInfo structure.
Now AddInTree, RemoveFromTree and RenameInTree call the related structure functions in order to keep the two representation updated.


donho:
I will see what I can do for #9209

Issue #8020, opened 2020-03-08
Bug and Translation: untranslated labels in User-defined form (Undock and sometimes Dock label) and scrollbar missing


conky77:
### Description of the Issue

Language-->User defined language-->define your language
1-the label for button "Undock" is not possible to translate (it is not present in the master English.xml)
2- the label of button "dock" is translated only when you get into the form the first time; if you press "dock" and "undock", the label "Dock" appears untranslated
3-in the form user-defined, in undock-mode, there is not any scrollbars

### Steps to Reproduce the Issue

1. Language-->User defined language-->define your language (the first time the label "Dock"  is translated)
![image](https://user-images.githubusercontent.com/59116246/76161216-6db39980-6131-11ea-996a-a938002e902f.png)

2.  press "Dock": the button "Undock" appears untranslated  
![image](https://user-images.githubusercontent.com/59116246/76161168-e49c6280-6130-11ea-8be0-b4337328bf9e.png)

3.  Press "undock": the button "Dock" appears untranslated

![image](https://user-images.githubusercontent.com/59116246/76161333-ba4ba480-6132-11ea-8053-083fb459cb88.png)

4. when the form is Undocked, there is not any scrollbars

![image](https://user-images.githubusercontent.com/59116246/76161222-8328c380-6131-11ea-9c21-6d0404c916a7.png)

### Expected Behavior

1. Include the label "undock" in the master English.xml
2. refresh the buttons Dock/Undock with the right translation
3. insert a scrollbar in the form in "undock-MODE"

### Actual Behavior

Labels not translated for "undock", the translation for "Dock" appears only the first time, scrollbar missing in "undock-mode"

### Debug Information

Notepad++ v7.8.5   (32-bit)
Build time : Mar  4 2020 - 11:04:20
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1903
OS Build : 18362.657
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll Translate.dll XMLTools.dll

ArkadiuszMichalski:
>3-in the form user-defined, in undock-mode, there is not any scrollbars

This dialog is not resizable so scrollbars are unnecessary.

PR #13127, opened 2023-02-14
Fix translation for Dock/Undock label in User-Defined dialog

ArkadiuszMichalski:
Fix [8020](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8020).

@donho 
I create additional function `getAttrNameByIdStr`. It's universal, we pass the `targetNode` for searching `<Item>` elements  (so before passing we can choose any one, regardless of the level of nesting). It retrieves the value of the `"name"` attribute based on the name and the value of the `"id"` attribute. Attribute names `"id"` and `"name"` can be changed (these are default).
I noticed that sometimes we have already translated value(s) in some node(s) and it doesn't make sense to add more nodes with the same value (especially when we are in the same parent). Unfortunately, there was no universal method defined to achieve this, so I added it.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/UserDefineDialog.cpp
+++ b/PowerEditor/src/ScintillaComponent/UserDefineDialog.cpp
@@ -933,9 +933,22 @@ void UserDefineDialog::changeStyle()
         if (targetNode)
             targetNode = targetNode->FirstChildElement("UserDefine");
     }
-    generic_string dockStr = pNativeSpeaker->getAttrNameByIdStr(TEXT("Dock"), targetNode, std::to_string(IDC_DOCK_BUTTON).c_str());
-    generic_string undockStr = pNativeSpeaker->getAttrNameByIdStr(TEXT("Undock"), targetNode, std::to_string(IDC_UNDOCK_BUTTON).c_str());
-    ::SetDlgItemText(_hSelf, IDC_DOCK_BUTTON, (_status == DOCK) ? undockStr.c_str() : dockStr.c_str());
+    generic_string dockButtonLabel;
+    generic_string defauleLabel;
+    string idStr;
+    if (_status == DOCK)
+    {
+        defauleLabel = TEXT("Undock");
+        idStr = std::to_string(IDC_UNDOCK_BUTTON);
+    }
+    else
+    {
+        defauleLabel = TEXT("Dock");
+        idStr = std::to_string(IDC_DOCK_BUTTON);
+    }
+
+    dockButtonLabel= pNativeSpeaker->getAttrNameByIdStr(defauleLabel.c_str(), targetNode, idStr.c_str());
+    ::SetDlgItemText(_hSelf, IDC_DOCK_BUTTON, dockButtonLabel.c_str());
 
     auto style = ::GetWindowLongPtr(_hSelf, GWL_STYLE);
     if (!style)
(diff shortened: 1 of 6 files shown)


PowerEditor/src/ScintillaComponent/UserDefineDialog.cpp
```diff
@@ -925,9 +925,14 @@ void UserDefineDialog::reloadLangCombo()
 void UserDefineDialog::changeStyle()
 {
     _status = !_status;
-    ::SetDlgItemText(_hSelf, IDC_DOCK_BUTTON, (_status == DOCK)?TEXT("Undock"):TEXT("Dock"));
-
-	auto style = ::GetWindowLongPtr(_hSelf, GWL_STYLE);
+    NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
+    TiXmlNodeA *targetNode = nullptr;
+    pNativeSpeaker->getNativeLangA() && (targetNode = (pNativeSpeaker->getNativeLangA())->FirstChildElement("Dialog")) && (targetNode = targetNode->FirstChildElement("UserDefine"));
```
chcg, line 930: C:\projects\notepad-plus-plus\PowerEditor\src\ScintillaComponent\UserDefineDialog.cpp(930): error C2220: the following warning is treated as an error [C:\projects\notepad-plus-plus\PowerEditor\visual.net\notepadPlus.vcxproj]
     5>C:\projects\notepad-plus-plus\PowerEditor\src\ScintillaComponent\UserDefineDialog.cpp(930): warning C4706: assignment within conditional expression [C:\projects\notepad-plus-plus\PowerEditor\visual.net\notepadPlus.vcxproj]

ArkadiuszMichalski: @chcg 
Is there any equivalent in GCC for this warning ([manual](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options))? For `Debug` we turn on a lot of things but I don't see any such message: https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/46232207/job/8aqqrg6cde100t7m.

PowerEditor/src/ScintillaComponent/UserDefineDialog.cpp
```diff
@@ -925,9 +925,19 @@ void UserDefineDialog::reloadLangCombo()
...
+    TiXmlNodeA *targetNode = nullptr;
+    if (pNativeSpeaker->getNativeLangA())
+    {
+        targetNode = (pNativeSpeaker->getNativeLangA())->FirstChildElement("Dialog");
+        if (targetNode)
+            targetNode = targetNode->FirstChildElement("UserDefine");
+    }
+    generic_string dockStr = pNativeSpeaker->getAttrNameByIdStr(TEXT("Dock"), targetNode, std::to_string(IDC_DOCK_BUTTON).c_str());
+    generic_string undockStr = pNativeSpeaker->getAttrNameByIdStr(TEXT("Undock"), targetNode, std::to_string(IDC_UNDOCK_BUTTON).c_str());
+    ::SetDlgItemText(_hSelf, IDC_DOCK_BUTTON, (_status == DOCK) ? undockStr.c_str() : dockStr.c_str());
```
donho, line 938: I would do somethings similar as following code:
```cpp
generic_string dockButtonLabel;
generic_string defauleLabel;
generic_string idStr;
if (_status == DOCK)
{
    defauleLabel = TEXT("Undock");
    idStr = std::to_string(IDC_UNDOCK_BUTTON);
}
else
{
    defauleLabel = TEXT("Dock");
    idStr = std::to_string(IDC_DOCK_BUTTON);
}

dockButtonLabel= pNativeSpeaker->getAttrNameByIdStr(defauleLabel.c_str(), targetNode, idStr.c_str());
::SetDlgItemText(_hSelf, IDC_DOCK_BUTTON, dockButtonLabel.c_str());
```

So we don't call `getAttrNameByIdStr` twice.

donho (changes requested):
Please follow the suggestion.

donho (approved):

PR #13128, opened 2023-02-14
Add support for non-printing char in main Search Result dock window

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13020#issuecomment-1430034599


ozone10:
@alankilborn 
could you test.

alankilborn:
It is now 6 hours later and this is still building.  Why?

Yaron10:
@ozone10,

👍 
Thank you for your work.

I'm wondering if toggling `Show NPC` should *immediately* affect the Search Results or only after the next search.
What do you all think?

alankilborn:
>I'm wondering if toggling Show NPC should immediately affect the Search Results or only after the next search.
>What do you all think?

IMO immediately is better, but my mind could be changed about that.
There are some "settings" changes that only affect future searches, but those likely work that way just because it was more convenient for the code that way.

alankilborn:
@ozone10 

STR:
Show Non-Printable Chars ON
Do a FACD search that results in NPC in _Search results_
All good so far...
Right click in _Search results_
Do _Find in these search results_
Use same search expression as before
Do _Find All_
Note that NPC are **_NOT_** shown in the new search results

ozone10:
@alankilborn 
this is why title contains word "main" 😎

Joking aside, I cannot reproduce it.
image

Secondary search result windows use same style as main at the time they were created. After that the style remains same.
Even changing theme or dark mode don't affect them, this seems like bug.

Yaron10:
> I'm wondering if toggling Show NPC should immediately affect the Search Results or only after the next search.
>> IMO immediately is better, but my mind could be changed about that.

This approach can certainly be useful and desirable in many cases.
But the user might search in a tab they want to see with NPC displayed, switch to another tab they want to see with NPC hidden - and still have the Search Results displaying NPC.

If we take LTR/RTL as an example, changing the direction in the editor does NOT affect the Search Results until a new search is executed. And this is how it should be.

Adding a separate NPC setting for the Search Results might also be an option.

***
> STR:
Show Non-Printable Chars ON
>> Joking aside, I cannot reproduce it.

I can't reproduce that either.

***
> Secondary search result windows use same style as main at the time they were created. After that the style remains same.
Even changing theme or dark mode don't affect them, this seems like bug.

👍 
Thank you for the fix. Great work.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -353,13 +353,29 @@ public :
 		if (_pFinder && _pFinder->isCreated())
 		{
 			_pFinder->setFinderStyle();
```
Yaron10, line 355: Is this line still needed?

ozone10:
> Future work to consider, just mentioning it here as a passing comment:
> 
> Now that user can SEE these characters in the editing window and _Search results_, when they select some text containing them and press Ctrl+f, how disappointed are they when they CAN'T see them in the _Find_ window's _Find what_ box?

This would be hard to implement, edit control is not scintilla.

donho:
@ozone10 
2 regressions has been detected with this PR:

1. white space & TAB don't display anymore.
2. NPC don't apply the colour anymore.


image

donho:
@ozone10 
Sorry for the false alert (deleted). I should get myself a pair of glasses.

The following 3 symbols display already without NPC enabled:  
```
U+0085     next line
U+2028       line separator
U+2029       paragraph separator
```

I'm wondering if they should be on the list.

ozone10:
@donho 
those are unicode EOL. I added them for the sake of completeness.

Scintilla support them, but lexers need to support them to work correctly.

Issue #5502, opened 2019-04-09
Confusing "Remove unmarked lines"


wilkowy:
```
<Item id="43021" name="Remove Bookmarked Lines"/>
<Item id="43051" name="Remove Unmarked Lines"/>
```
The "Remove Unmarked Lines" is confusing and it should state "remove not bookmarked lines" as it has nothing to do with marking from find/replace menu, unless bookmarking is checked. (yes, I know it is in "Bookmark" submenu, but still…)

alankilborn:
Tested in N++ 8.4.4 (current version as of this writing) and the problem still exists.
Could someone change the English text to be different (i.e., better) than `Remove unmarked lines` (see suggestions above)?

ArkadiuszMichalski:
Fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13149.

PR #13149, opened 2023-02-16
[XML] Rename "Remove Unmarked Lines" to "Remove Non-Bookmarked Lines"

ArkadiuszMichalski:
Fix [5502](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5502).

@alankilborn 
I use `Non-Bookmarked` term. We use `non-` pattern in other place (`"Show Non-Printing Characters"` or `"Non-ASCII Characters (128-255)"`). If this term (or any another) is not accepted and we stay with the current one, it's best to close issue 5502.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -202,7 +202,7 @@ The comments are here for explanation, it's not necessary to translate them.
                     <Item id="43019" name="Copy Bookmarked Lines"/>
                     <Item id="43020" name="Paste to (Replace) Bookmarked Lines"/>
                     <Item id="43021" name="Remove Bookmarked Lines"/>
-                    <Item id="43051" name="Remove Unmarked Lines"/>
+                    <Item id="43051" name="Remove Non-Bookmarked Lines"/>
                     <Item id="43050" name="Inverse Bookmark"/>
                     <Item id="43052" name="Find characters in rang&amp;e..."/>
                     <Item id="43053" name="Select All Between &amp;Matching Braces"/>
--- a/PowerEditor/installer/nativeLang/english_customizable.xml
+++ b/PowerEditor/installer/nativeLang/english_customizable.xml
@@ -191,7 +191,7 @@
                     <Item id="43019" name="Copy Bookmarked Lines"/>
                     <Item id="43020" name="Paste to (Replace) Bookmarked Lines"/>
                     <Item id="43021" name="Remove Bookmarked Lines"/>
-                    <Item id="43051" name="Remove Unmarked Lines"/>
+                    <Item id="43051" name="Remove Non-Bookmarked Lines"/>
                     <Item id="43050" name="Inverse Bookmark"/>
                     <Item id="43052" name="Find characters in rang&amp;e..."/>
                     <Item id="43053" name="Select All Between &amp;Matching Braces"/>
--- a/PowerEditor/src/Notepad_plus.rc
+++ b/PowerEditor/src/Notepad_plus.rc
@@ -643,7 +643,7 @@ BEGIN
             MENUITEM "Copy Bookmarked Lines",                  IDM_SEARCH_COPYMARKEDLINES
             MENUITEM "Paste to (Replace) Bookmarked Lines",    IDM_SEARCH_PASTEMARKEDLINES
             MENUITEM "Remove Bookmarked Lines",                IDM_SEARCH_DELETEMARKEDLINES
-            MENUITEM "Remove Unmarked Lines",                  IDM_SEARCH_DELETEUNMARKEDLINES
+            MENUITEM "Remove Non-Bookmarked Lines",            IDM_SEARCH_DELETEUNMARKEDLINES
             MENUITEM "Inverse Bookmark",                       IDM_SEARCH_INVERSEMARKS
         END
         MENUITEM SEPARATOR


alankilborn:
@ArkadiuszMichalski Looks reasonable; thank you.

ozone10:
@ArkadiuszMichalski 
it should be `Unbookmarked` or `Not Bookmarked`

alankilborn:
IMO `Non-Bookmarked` is best, but who's the authority besides @donho ?  He'll decide.

ozone10:
Grammatically correct should be `Unbookmarked`

ArkadiuszMichalski:
So we have:

Remove Non-Bookmarked Lines
Remove Not Bookmarked Lines
Remove Unbookmarked  Lines

Whatever @donho chooses, I will change. It's mainly about that to not using the word `marked`, because it can be confused with another mechanism provided in Notepad++.

alankilborn:
Don, as a non-native speaker, may not be the best authority, but yes it is his choice.
Don probably likes the original, as I'd think that was his choosing at the time of original implementation.

Perhaps some citing of a grammar rule should back up a choice.
Note that I have nothing to cite as I rely on what "sounds good" to me. :-)
I think strict grammar rules are relaxed in technical jargon situations, but I could be wrong.

Perhaps all uses of `Non-` should be evaluated for correctness?
`Show Unprinting Characters`? :-)

If _feels_ like to me that to "unbookmark", something had to be bookmarked in the past.

How about `Keep Only Bookmarked Lines` ?  :-)

ozone10:
[Non-printing characters](https://en.wikipedia.org/wiki/Control_character) is correct term 

I would use `Remove Not Bookmarked Lines`

ArkadiuszMichalski:
For me the following two opposing commands are clear:
`Remove Bookmarked Lines`
`Remove Not Bookmarked Lines`

pryrt:
My initial reaction, after reading the possibilities, was to vote "Non-Bookmarked", based on what "feels right" to my grammar-nitpicking mind.  But that's not authoritative, so I looked up some references to confirm.

For the same reason that "non-printing" is the correct way to modify "characters", "non-bookmarked" is the correct way to modify "lines".  Compound modifiers (where the multiple terms are _as a single_ unit modifying the noun) are hyphenated [[1]](https://www.grammarly.com/blog/hyphen/#:~:text=Hyphen%20with%20compound%20modifiers%3A%20multiple%2Dword%20adjectives%20before%20nouns) [[2]](https://prowritingaid.com/grammar/1000138/How-to-use-hyphen-after-adverbs#:~:text=When%20a%20hyphen%20connects%20an,compound%20modifier) [[3]](https://www.grammarbook.com/punctuation/hyphens.asp): hence, the "purple-people eater" of lyric fame only eats purple people, because the purple and people are compounded together, acting as a single unit. 

Here, the "non-bookmarked" modifies (as a unit) the line, so it needs the hyphen.

For `not` vs `non`, this [English.StackExchange comment](https://english.stackexchange.com/questions/277210/no-not-and-non#:~:text=So%20we%20can%20say%20non,not%20goes%20before%20a%20verb.) sums it up with "So we can say `non` goes before an adjective, `no` goes [before] a name, `not` goes before a verb."  Since `bookmarked` is an adjective, it needs `non-` as its compound-modifier prefix.

Hence, my vote is still `Remove Non-Bookmarked Lines`.

ozone10:
well `bookmarked` is adjective from verb - past participle

`un-` is commonly used to prefix participle to create opposite meaning.

pryrt:
`un-` vs `non-` is not as cut-and-dry:

- [the grammarist](https://grammarist.com/usage/negative-prefixes/#:~:text=Unlike%20in%2D%2C%20and%20un%2D,who%20absolutely%20does%20not%20conform.) says 
    >The prefix non- is the most useful negative prefix, as it can be attached to virtually any [noun](https://grammarist.com/grammar/nouns/), [verb](https://grammarist.com/grammar/verbs/), adjective, or [adverb](https://grammarist.com/grammar/adverbs/) and is not confusable with other common morphemes. Unlike in-, and un-, which often create nonabsolute negatives, non- is generally used to create a word that describes the complete opposite of its nonnegative form. In other words, for example, a nonconformist is someone who absolutely does not conform. If we were to coin the nouns uncomformist and inconformist, they wouldn’t have the same sense.

- but [this English.SE](https://english.stackexchange.com/questions/277906/when-is-the-prefix-non-used-vs-un) says the opposite:
    > "Un-" is [defined](http://dictionary.reference.com/browse/un-) as "a prefix meaning 'not,' freely used as an English formative, giving negative or opposite force in adjectives and their derivative adverbs and nouns... and less freely used in certain other nouns."

    > "Non-" is [defined](http://dictionary.reference.com/browse/non-?s=t) as "a prefix meaning 'not,' freely used as an English formative, usually with a simple negative force as implying mere negation or absence of something (rather than the opposite or reverse of it, as often expressed by un-)."

When reasonable sources disagree, I go back to my gut.  And my gut, like @alankilborn's assertion earlier, says that "unbookmarked" is "[adj.] something that has had the bookmark removed".  Which does not have the same connotation to me as "non-bookmarked" which is "[adj.] something that is not currently bookmarked, whether or not it has been bookmarked in the past".

_/pryrt is already regretting having nosed into a grammar war..._

alankilborn:
>a grammar war...

I don't think that's what it is, at its core.
It is people that care about making N++ the best that it can be, even in the smallest detail.

pryrt:
> It is people that care about making N++ the best that it can be, even in the smallest detail.

There is that, yes.

At this point, with opinions and research both adequately expressed, there needs to be input from the one with authority to make the executive decision as to the final wording, because I doubt this "committee" is ever going to come to complete agreement.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -202,7 +202,7 @@ The comments are here for explanation, it's not necessary to translate them.
                     <Item id="43019" name="Copy Bookmarked Lines"/>
                     <Item id="43020" name="Paste to (Replace) Bookmarked Lines"/>
                     <Item id="43021" name="Remove Bookmarked Lines"/>
-                    <Item id="43051" name="Remove Unmarked Lines"/>
+                    <Item id="43051" name="Remove Non-Bookmarked Lines"/>
```
donho, line 205: It's a better name indeed.
But then why don't we change also in the binary? (`MENUITEM "Remove Unmarked Lines",                  IDM_SEARCH_DELETEUNMARKEDLINES` in Notepad_plus.rc)

donho (changes requested):
Change also in Notepad_plus.rc (`MENUITEM "Remove Unmarked Lines",                  IDM_SEARCH_DELETEUNMARKEDLINES`)

donho:
> Don probably likes the original, as I'd think that was his choosing at the time of original implementation.
> ...
> It is people that care about making N++ the best that it can be, even in the smallest detail.

I've got my ego, as anyone here. But I'm part of people who care about making Notepad++ the best it can be.
So let the best choice be the final decision.
Thanks you all guys for your opinions and suggestions to make the best choice.

donho (approved):

PR #13159, opened 2023-02-17
Volatile Find uses least-strict option settings

### LINKED ISSUE #13145: "Find (volatile)" is always case sensitive
case sensitive -- casE sENsitive -- case sensitive   <== paste the present line in to a tab and try it

### Expected Behavior
Search should follow dialog settings, same as non-volatile analogs, esp. non-volatile Select and Find

### Debug Information
I discovered it in 8.1.9 (32-bit portable); confirmed in 8.4.5 (32-bit portable) & 8.4.6 (64-bit, normal install). They all have different plugins.

alankilborn:
Fix #13145 

-----------------------------------

Code unified with `IDM_SEARCH_SETANDFINDNEXT` block as much as reasonable (that block is currently fully working).  

-----------------------------------

To get a true picture of what this PR proposes, do this compare, with, e.g. ComparePlus plugin:

Left side of compare:
```
case IDM_SEARCH_SETANDFINDNEXT :
case IDM_SEARCH_SETANDFINDPREV :
{
...
}
```

Right side of compare (using the proposed new section from this PR):
```
case IDM_SEARCH_VOLATILE_FINDNEXT :
case IDM_SEARCH_VOLATILE_FINDPREV :
{
...
}
```

The difference should be that the "volatile" code (right-hand side) does NOT set the _Find what_ box data.

-----------------------------------

Perhaps it makes sense to do this with the code instead?:

```
case IDM_SEARCH_SETANDFINDNEXT :
case IDM_SEARCH_SETANDFINDPREV :
case IDM_SEARCH_VOLATILE_FINDNEXT :
case IDM_SEARCH_VOLATILE_FINDPREV :
{
...
}
```

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1372,15 +1372,26 @@ void Notepad_plus::command(int id)
 		case IDM_SEARCH_VOLATILE_FINDNEXT :
 		case IDM_SEARCH_VOLATILE_FINDPREV :
 		{
-			TCHAR text2Find[MAX_PATH] = { '\0' };
-			_pEditView->getGenericSelectedText(text2Find, MAX_PATH);
+			bool isFirstTime = !_findReplaceDlg.isCreated();
+			if (isFirstTime)
+			{
+				_findReplaceDlg.doDialog(FIND_DLG, _nativeLangSpeaker.isRTL(), false);
+			}
 
-			FindOption op;
-			op._isWholeWord = false;
-			op._whichDirection = (id == IDM_SEARCH_VOLATILE_FINDNEXT?DIR_DOWN:DIR_UP);
+			const int strSize = FINDREPLACE_MAXLENGTH;
+			TCHAR str[strSize] = { '\0' };
+			_pEditView->getGenericSelectedText(str, strSize);
+			if (isFirstTime)
+			{
+				_nativeLangSpeaker.changeFindReplaceDlgLang(_findReplaceDlg);
+			}
+
+			FindOption op = _findReplaceDlg.getCurrentOptions();
+			op._searchType = FindNormal;
+			op._whichDirection = (id == IDM_SEARCH_VOLATILE_FINDNEXT ? DIR_DOWN : DIR_UP);
 
 			FindStatus status = FSNoMessage;
-			_findReplaceDlg.processFindNext(text2Find, &op, &status);
+			_findReplaceDlg.processFindNext(str, &op, &status);
 			if (status == FSEndReached)
 			{
 				generic_string msg = _nativeLangSpeaker.getLocalizedStrFromID("find-status-end-reached", TEXT("Find: Found the 1st occurrence from the top. The end of the document has been reached."));
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -1372,21 +1372,14 @@ void Notepad_plus::command(int id)
 		case IDM_SEARCH_VOLATILE_FINDNEXT :
 		case IDM_SEARCH_VOLATILE_FINDPREV :
 		{
-			bool isFirstTime = !_findReplaceDlg.isCreated();
-			if (isFirstTime)
-			{
-				_findReplaceDlg.doDialog(FIND_DLG, _nativeLangSpeaker.isRTL(), false);
-			}
-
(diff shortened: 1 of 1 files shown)


Yaron10:
Isn't the *main* purpose of "volatile" is providing a search with fixed parameters NOT depending on the current settings in the Find dialog?

***
`MAX_PATH` to `FINDREPLACE_MAXLENGTH`.
👍

alankilborn:
>Isn't the main purpose of "volatile" providing a search with fixed parameters NOT depending on the current settings in the Find dialog?

The "volatile" part only refers to the _Find what_ data.
Otherwise, we have a problem -- and that is, how should the other things be treated?
Choice are:  
* "wide open" : match-case=no, whole-word=no
* "narrow" : match-case=yes, whole-word=yes
* just follow what user has set up for other searches they commonly do:  this PR
* chosen by yet-more _Preference_ settings

Yaron10:
> The "volatile" part only refers to the Find what data.

Literally that's correct. And "volatile" is indeed not the best term.

> "wide open" : match-case=no, whole-word=no

Current master behavior? - I think that's fine.

Yaron10:
> Current master behavior?

Sorry.

"wide open" : match-case=no, whole-word=no is a good option.

alankilborn:
>Current master behavior? - I think that's fine.

* Confuses users, see #13145 user complaint reasoning 
* Goes against documented behavior in the user manual for non-dialog searches

I had some offline discussion with @pryrt -- our concurrence was that this PR should be submitted.

alankilborn:
>And "volatile" is indeed not the best term

I'm sure Don is open to better verbiage if you've got a good suggestion. :-)

Yaron10:
> Being different from other non-dialog search

That's the point! We can use `IDM_SEARCH_SETANDFINDNEXT` with certain parameters, and  `IDM_SEARCH_VOLATILE_FINDNEXT` with other parameters.
I'm pretty sure that was the purpose and intention of adding `VOLATILE`.

> And "volatile" is indeed not the best term.

I meant to emphasize that "volatile" was negligible compared to the main feature: providing a different search.

alankilborn:
User confusion --> BAD
Even *I*, as a user, couldn't have told you how this worked before I got into it.

The point of this "volatile" search is that it is different from the other one in exactly one way, and that is that the _Find what_ is not-remembered.

Don will decide.

alankilborn:
@chcg

I disagree with the "enhancement" label as existing software does not conform to behavior described in the user manual.  I would put the "bug" label on it.

donho:
@alankilborn 
> I disagree with the "enhancement" label as existing software does not conform to behavior described in the user manual. I would put the "bug" label on it.

It could be an enhancement for some, and fix for others. 

On the behaviour of volatile searching, I agree totally with @Yaron10 
here's my argument: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13145#issuecomment-1440175719

alankilborn:
>It could be an enhancement for some, and fix for others.

I suppose so! :-)

>On the behaviour of volatile searching, I agree totally with @Yaron10

OK, then I should adjust this PR to conform to what I've hopefully outlined as the right behavior, here, if I've got it totally correct? :
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13145#issuecomment-1440292963

alankilborn:
Based on concurrence https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13145#issuecomment-1440331537 I'll adjust this PR...

donho (changes requested):
News specs:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13145#issuecomment-1440292963

Yaron10:
@alankilborn,

> Even I, as a user, couldn't have told you how this worked before I got into it.

I've refereed to you on various occasions as a "Notepad++ Encyclopedia". 
But NPP has so many features that even an expert like yourself may not have thoroughly used them all. :)

I happened to use "volatile" and realized it made "NPP Search ToolBox" much more flexible:
```
SelectedTab
Selected Tab
```
 you could search for `Selected` (whole word yes/no) *without changing any setting*.

And now that you "got into it" and suggested to make it case-insensitive it's even better.

Thank you for your work. 👍 

***

@donho,

Thank you for making Notepad++ as great as it is. 👍

alankilborn:
@donho Ready for review.

donho (approved):

PR #13165, opened 2023-02-18
[XML] update italian.xml v.8.5.0

conky77:


donho:
Could you please check the following announcement for completing your localization file if it's not done?
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13209

Thank you @conky77

PR #13143, opened 2023-02-15
Fix GCC implicit-fallthrough= warnings

ozone10:
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13142

PR #13168, opened 2023-02-18
Enable code analysis for debug builds

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13143#issuecomment-1434689064


ozone10:
@donho 
as I said enabling code analysis will significantly increase compile time.
Thus this PR only enable code analysis on x64 debug build. The increase is about 12 min.

Some warnings are helpful but hidden, to make it easier to spot more important warnings.
These warnings can be enabled later after dealing with enabled warnings.

Some warnings are set to none, these won't be perhaps fixed - don't use `reinterpret_cast`, ...

donho:
@ozone10 
I thought it will stop the compiling as error due to the warning, apparently there are only warning after the compiling...
I just comment randomly a "break;" statement in a switch instruction - I don't see any warning. Is it normal?

ozone10:
@donho 
after updating to latest MS Visual Studio 17.5 I cannot trigger warning c26819 too.

But there is new warning for 17.4, [c5262](https://learn.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default?view=msvc-170).

It does not detect all implicit fall-through but better than nothing.

ozone10:
There is some issue with external includes, I will fix it later

ozone10:
@donho 
After some research, warning C5262 causes issue with STL libraries for MS Visual Studio 2022 version below 17.5
Appveyor currently use version 17.4.

IMHO this PR should wait for appveyor update and then enable warning C5262.

ozone10:
Appveyor has been [updated](https://www.appveyor.com/updates/2023/02/25/).
Hopefully it will work.

Issue #13176, opened 2023-02-19
Preference Language Tab size replace "mini" dlg with edit control


ozone10:
### Description of the Issue

Setting tab size should be directly in preference dialog instead of using link with "mini" dialog
### Steps to Reproduce the Issue

1. Open Preference->Language
2. Click on link text - number next to `Tab size:`

### Expected Behavior

Change tab size directly in preference dialog.

### Actual Behavior

Use "mini" dlg to change tab size.

### Debug Information

Notepad++ v8.4.9

![image](https://user-images.githubusercontent.com/55940305/219969361-00f143fa-53e4-4ef1-a171-414454cb2d26.png)

Issue #11695, opened 2022-05-18
[DarkMode] Blurry static text in preference "tab size"


ozone10:
Issue for Notepad++ v8.4.1

![image](https://user-images.githubusercontent.com/55940305/169117467-8ce1333b-2374-4e71-84b1-04c3f61a187b.png)

Feel free to do PR, I am currently busy.
If no one will fix it, I will do it later when I have free time.

ts-50:
That's odd, I'm not getting that. I'm not even seeing a Default Value checkbox.
![image](https://user-images.githubusercontent.com/19561856/169307798-60e7c4ac-4fff-41da-b6d6-53d73ab7d08f.png)

Notepad++ v8.4.1   (64-bit)
Build time : May  8 2022 - 19:51:18
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 2009
OS Build : 22000.675
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

alankilborn:
>I'm not even seeing a Default Value checkbox.

Try selecting something other than `[Default]` in the listbox and it will appear.

Issue #12558, opened 2022-11-22
"Tab size" UI anomaly in the preferences


alankilborn:
### Description of the Issue
Weirdness in the UI for _Tab size_ on the _Preferences_ > _Language_ > _Tab Settings_ > _Tab size:_ control.

### Steps to Reproduce the Issue

0. Start with a fresh portable extract of N++ so that all language settings are at default values.
1. Open _Preferences_ > _Language_ and select `python` in the _Tab Settings_ area, note the setup:
![image](https://user-images.githubusercontent.com/77065706/203408150-78f912f3-2af1-4fac-9608-4cd50a883c1a.png)
2. Click on the `4` and change tab size to `3`:
![image](https://user-images.githubusercontent.com/77065706/203408316-52a1535e-ee21-4b65-9401-9ad7e92cbe59.png)
3. Check the box for _Use default value_:
![image](https://user-images.githubusercontent.com/77065706/203408592-23db7135-aff4-4af1-ac4d-e35a0832d651.png)
4. Move to a language choice different from `python`, e.g.:
![image](https://user-images.githubusercontent.com/77065706/203408781-52fc2dbd-687e-4ad7-96b1-a9e89d65e66e.png)
5. Return to `python`:
![image](https://user-images.githubusercontent.com/77065706/203408889-793d6beb-e009-4afa-bfc9-9f6e28ce599d.png)

### Expected Behavior
After step 3, I'd expect that _Tab size:_ to show 4

### Actual Behavior
After step 3, _Tab size:_ shows 3

### Notes
Note that after step 5, all is as it should be for `python`.  
But, it should indicate the correct tab-size for Python after step 3.

### Debug Information
Notepad++ v8.4.7   (64-bit)
Build time : Nov  1 2022 - 23:31:11
Path : C:\\...\npp.8.4.7\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.2130
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

alankilborn:
Appears to be fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/8b3f072a38657b521b14403d64d42d7d7db2a5cf

Can be CLOSED.

[...]

PR #13177, opened 2023-02-19
Replace tab size link mini dlg with edit control

ozone10:
image

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11695
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13176


Yaron10:
@ozone10,

👍 
Thank you for this too. Two more. :)

1. If the user enters `0` (zero) in the EditBox and moves focus to another element, that value is not auto-changed.
2. `Replace with Spaces` might still be blurry in DarkMode .

Yaron10:
@ozone10,

Thank you for the quick fix. 👍 

How about implementing it in `if (tabSize < 1)` under `case EN_CHANGE:`?
That is: emulate Windows behavior on trying to enter non-numeric chars, and if the users enters `0` - beep and restore the previous value.
(The user won't be able to type `01`, but that shouldn't be a problem IMO).

ozone10:
@Yaron10 

> 2. `Replace with Spaces` might still be blurry in DarkMode .

That is checkbox button text, and use different text drawing for disabled state than static text, if there was issue with it, it should also affect other places.

> How about implementing it in `if (tabSize < 1)` under `case EN_CHANGE:`? That is: emulate Windows behavior on trying to enter non-numeric chars, and if the users enters `0` - beep and restore the previous value. (The user won't be able to type `01`, but that shouldn't be a problem IMO).

Currently there is no difference between `0` and empty value. Differentiating would make code more complex.
Also consider value e.g. `10` user will put caret before `1` press `Delete` -> value will change to `0` -> value will be restored to `10`.  This is not good.

Yaron10:
@ozone10,

> Currently there is no difference between 0 and empty value.
...
Also consider value e.g. 10 user will put caret before 1...

Both arguments are very good. Thank you for the explanation. 👍 

***

תמונה
תמונה

If you can't reproduce it, it may be caused by my Windows 10 settings and custom theme.
You know I don't use DarkMode. :)

PowerEditor/src/WinControls/Window.h
```diff
@@ -64,8 +64,20 @@ class Window
 			::UpdateWindow(_hSelf);
 	}
 
+	virtual void redrawDlgItem(const int nIDDlgItem, bool forceUpdate = false) const
```
donho, line 67: What does it do exactly this function?
And why do we need it?

donho (changes requested):
`redrawDlgItem` uses MapWindowPoints function which converts (maps) a set of points from a coordinate space relative to one window to a coordinate space relative to another window, then redraw it with the new coordonate points.
I don't see the reason for it.
Please provide the arguments for this new added method.

PowerEditor/src/WinControls/Window.h
ozone10: @donho 
currently using `redraw()` will redraw all items in window which is unnecessary and can cause flickering.
Try switching fast options in Preference->Editing->Current Line Indicator.

`MapWindowPoints` is used to get correct coordinates of child rect for redrawing.
https://stackoverflow.com/questions/18034975/how-do-i-find-position-of-a-win32-control-window-relative-to-its-parent-window

donho: @ozone10 
In this case, this method should be in `class StaticDialog` instead of `class Window`.

donho (changes requested):
Please put the method redrawDlgItem in `class StaticDialog` instead of `class Window`.

Yaron10:
Fixes also https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12558.
👍

Issue #13187, opened 2023-02-21
[GUI] Edit control highlight on focus for incremental search and function panel


ozone10:
### Description of the Issue

It might be hard to notice that edit controls for incremental search and function panel have focus
### Steps to Reproduce the Issue

1. Open Function panel or incremental search
2. Click on edit control

### Expected Behavior

Edit control has border highlighted.

### Actual Behavior

No visual change to edit control border.

### Debug Information

Notepad++ v8.4.9

![image](https://user-images.githubusercontent.com/55940305/220367955-4650a030-5578-4677-8a5e-c1f178f2b451.png)

PR #13188, opened 2023-02-21
Enable edit control highlight on focus

ozone10:
for incremental search and function panel
add initializers

image

Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13187

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -5055,7 +5055,7 @@ void FindIncrementDlg::display(bool toShow) const
 	_pRebar->setIDVisible(_rbBand.wID, toShow);
 }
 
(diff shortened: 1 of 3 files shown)


donho:
The result looks nice in dark mode, but not in light mode:

image

Is it possible to be like comboboxes in Find in files while the edit control being on focus?

Yaron10:
The result is fine with 96 DPI.
תמונה

ozone10:
@donho 
unfortunately this is how Windows 11 highlight edit control.
To use "Windows 10" style, same as in dark mode, I would need to rewrite custom border subclass to allow light mode to use custom border. Also the same Windows 11 style  is already used in e.g. preference dlg, plugin admin dlg, ...

@Yaron10 
you have Windows 10

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2bf50bc191a580c4f57e90bddc8277a33ad03618

Issue #13194, opened 2023-02-22
NPP jumps from "new 1" to "new 3" when document map is visible


RodneyRabbit:
### Description of the Issue

If the document map is enabled, NPP will not create a "new 2" document but will jump to "new 3".
I believe this is similar to issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8677 that was fixed.

### Steps to Reproduce the Issue

1. Open NPP.
2. Click View > Document Map.
3. Create a new document.

### Expected Behavior

See "new 1" then "new 2" in the tab / document list.

### Actual Behavior

See "new 1" then "new 3" in the tab / document list.

### Debug Information

Notepad++ v8.4.9   (64-bit)
Build time : Jan 27 2023 - 03:11:16
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 22H2
OS Build : 19045.2673
Current ANSI codepage : 1252
Plugins : 
    DSpellCheck (1.4.24)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NppXmlTreeviewPlugin (2)
    XMLTools (3.1.1.13)
    zoomdisabler_x64 (1.2)

The issue no longer occurs when the document map is disabled.
Probably not high priority, just a mild inconvenience for me. I copy paste into 10+ documents at a time from other sources "output1.txt" to "output10.txt" etc, and like to use the document map. It would be easier to align source and destination documents if they were created as "new 1-10" instead of "new 1" then "new 3-11".

alankilborn:
It isn't related specifically to the document map.  

Try this alternative StR:

Close the document map if it is open.
Restart N++ so that you have a single document open, `new 1`.
Press Ctrl+f and do a _Find All in Current Document_ search so that the _Search results_ window/panel appears.
Create a new tab/doc via Ctrl+n; observe that `new 3` is created.

The numbering has to do with how many "Scintilla" buffers are open.
The _Search results_ window is a Scintilla buffer.
_Document Map_ is apparently also a Scintilla buffer.

[...]

Issue #11704, opened 2022-05-20
Multiple new docs have the same title


Yaron10:
**STR:**
Open two new docs.
Move "New 2" to the other view.
Move "New 2" back to the other view.
Open a new doc.
Open another new doc.

**Result:**
Two docs titled "New 2".

![תמונה](https://user-images.githubusercontent.com/10229320/169531156-c70e61db-4c2c-4ab1-b4ca-3110aa6b0888.png)

Repeating those steps with more tabs, you can get:

![תמונה](https://user-images.githubusercontent.com/10229320/169531325-474f9ca2-0a4c-4ad0-acbc-837a59f486a0.png)

***

Notepad++ v8.4.1   (64-bit)
Build time : May 20 2022 - 00:41:35
Path : C---Notepad++ Original.exe
Command Line : $COMMAND_LINE_PLACEHOLDER$
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : 
    ComparePlus (1)
    HTMLTag (1.2.3)
    MenuIcons (1.2.5)
    PythonScript (2)
    _CustomizeToolbar (4.2)

ArkadiuszMichalski:
@Yaron10 

I can't reproduce this with actual NPP 8.4.9. If we have only 2 untitled documents, move second to other view and move back to main view then first doc is replaced by this second, and finally we have only 1 doc. Such behavior always been here or has it changed at some point?

Yaron10:
@ArkadiuszMichalski,

> If we have only 2 untitled documents, move second to other view and move back to main view then first doc is replaced by this second, and finally we have only 1 doc.

And then:
Open a new doc.
Open another new doc.

Thank you for testing it.

ArkadiuszMichalski:
Ah right.

PR #13204, opened 2023-02-23
Fix title selection for untiled files

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13194, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8677 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11704.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1263,10 +1263,19 @@ size_t FileManager::nextUntitledNewNumber(bool toMainOrSubEditView) const
 	for (size_t i = 0; i < _buffers.size(); i++)
 	{
 		Buffer *buf = _buffers.at(i);
-		// if untitled document is invisible, then don't put its number into array (so its number is available to be used)
-		if (buf->isUntitled() && (buf->_referees[0])->isVisible())
+		if (buf->isUntitled())
 		{
-			if (!toMainOrSubEditView || (buf->indexOfReference(_pNotepadPlus->_pEditView) > -1 || buf->indexOfReference(_pNotepadPlus->_pNonEditView) > -1))
+			bool isVisible = false;
+			for (size_t k = 0; k < buf->_referees.size(); k++)
+			{
+				if (buf->_referees[k]->isVisible())
+				{
+					isVisible = true;
+					break;
+				}
+			}
+			// if untitled document is invisible, then don't put its number into array (so its number is available to be used)
+			if (isVisible && (!toMainOrSubEditView || (buf->indexOfReference(_pNotepadPlus->_pEditView) > -1 || buf->indexOfReference(_pNotepadPlus->_pNonEditView) > -1)))
 			{
 				generic_string newTitle = ((NppParameters::getInstance()).getNativeLangSpeaker())->getLocalizedStrFromID("tab-untitled-string", UNTITLED_STR);
 				TCHAR *numberStr = buf->_fileName + newTitle.length();
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1257,7 +1257,7 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 	}
 }
 
-size_t FileManager::nextUntitledNewNumber(bool toMainOrSubEditView) const
+size_t FileManager::nextUntitledNewNumber() const
 {
 	std::vector<size_t> usedNumbers;
 	for (size_t i = 0; i < _buffers.size(); i++)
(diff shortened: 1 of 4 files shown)


Yaron10:
@ArkadiuszMichalski,

Thank you for your work. 👍 

After a *quick test* - your PR fixes the issues and doesn't seem to cause any regression.
But this kind of change requires more than a "quick test" and a better understanding of the code (i.e. not myself).

Still, allow me to ask: if `toMainOrSubEditView == false` - should `nextUntitledNewNumber()` be called at all?

***

Related:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11704.

ArkadiuszMichalski:
@Yaron10 
>Still, allow me to ask: if toMainOrSubEditView == false - should nextUntitledNewNumber() be called at all?

From what I can see every buffer has `_fileName` set at creation time. Some are created and invisibly but for them we also need to give a name. To make it easier we use `nextUntitledNewNumber()` so the name is incremented (`New 1`, `New 2`) instead of something fixed (like `New` for all of them). Thanks to this we do not have to take care of it ourselves. Buffers not loaded into `MainEditView` or `SubEditView` (i.e. loaded into other View or not at all) also have some value in `_fileName`. Automatic generation of this value makes it easier.

https://github.com/notepad-plus-plus/notepad-plus-plus/search?q=nextUntitledNewNumber&type=code

Yaron10:
@ArkadiuszMichalski,

Thank you for the explanation and also for fixing https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11704. 👍 

I'd like to emphasize again that I don't fully understand the relevant code.
But I still think that my question deserves some further investigation. :)

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/99e7c5a84cc994b795b7b5e7acfa4943b9b03ed4/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1335

Should `nextUntitledNewNumber()` be called for the invisible view or the Search Results?

Apart from that - just replacing
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/99e7c5a84cc994b795b7b5e7acfa4943b9b03ed4/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1267

with

```
if (buf->isUntitled() && (buf->_referees[0])->isVisible())
		{
			if (buf->indexOfReference(_pNotepadPlus->_pEditView) > -1 || buf->indexOfReference(_pNotepadPlus->_pNonEditView) > -1)
			{
				generic_string newTitle = ((NppParameters::getInstance()).getNativeLangSpeaker())->getLocalizedStrFromID("tab-untitled-string", UNTITLED_STR);
				TCHAR *numberStr = buf->_fileName + newTitle.length();
				int usedNumber = _wtoi(numberStr);
				usedNumbers.push_back(usedNumber);
			}
		}
	}
```

seems to solve https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8677 (i.e. without any further changes).

Let's wait for Don's review.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1258,16 +1258,25 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
...
+			for (size_t k = 0; k < buf->_referees.size(); k++)
+			{
+				if (buf->_referees[k]->isVisible())
+				{
+					isVisible = true;
+					break;
+				}
+			}
 			// if untitled document is invisible, then don't put its number into array (so its number is available to be used)
-			if ((buf->_referees[0])->isVisible())
```
donho, line 1270: Please use the following code:
```cpp
		if (buf->isUntitled())
		{
			bool isVisible = false;
			for (size_t k = 0; k < buf->_referees.size(); k++)
			{
				if (buf->_referees[k]->isVisible())
				{
					isVisible = true;
					break;
				}
			}

			if (isVisible)
			{
				if (buf->indexOfReference(_pNotepadPlus->_pEditView) > -1 || buf->indexOfReference(_pNotepadPlus->_pNonEditView) > -1)
				{
					generic_string newTitle = ((NppParameters::getInstance()).getNativeLangSpeaker())->getLocalizedStrFromID("tab-untitled-string", UNTITLED_STR);
					TCHAR* numberStr = buf->_fileName + newTitle.length();
					int usedNumber = _wtoi(numberStr);
					usedNumbers.push_back(usedNumber);
				}
			}
		}
```
to replace the old code:
```cpp
		if (buf->isUntitled())
		{
			// if untitled document is invisible, then don't put its number into array (so its number is available to be used)
			if ((buf->_referees[0])->isVisible())
			{
				generic_string newTitle = ((NppParameters::getInstance()).getNativeLangSpeaker())->getLocalizedStrFromID("tab-untitled-string", UNTITLED_STR);
				TCHAR *numberStr = buf->_fileName + newTitle.length();
				int usedNumber = _wtoi(numberStr);
				usedNumbers.push_back(usedNumber);
			}
		}
```

The other modification is not necessary.

donho (changes requested):
@Yaron10 's simpler solution fixes #13194 and #8677. However it doesn't fix #11704.

I have to combine @ArkadiuszMichalski 's code to make #8677 fixed:
```
			bool isVisible = false;
			for (size_t k = 0; k < buf->_referees.size(); k++)
			{
				if (buf->_referees[k]->isVisible())
				{
					isVisible = true;
					break;
				}
			}
```

Thank you for the contribution from both of you :)

Yaron10:
@ArkadiuszMichalski,

So it seems your approach of modifying `nextUntitledNewNumber()` was correct. :)
Thanks again for your work. 👍 

@donho,

Thank you for reviewing and optimizing. 👍

donho (approved):

Issue #9343, opened 2021-01-03
Translation error: "Recent Files" main menu entry is not translated under certain circumstances


dinkumoil:
### Description of the Issue
If in main menu `File` the entry `Recent Files` is removed from the menu and added again later, it is not translated anymore.

### Steps to Reproduce the Issue
1. Open some files with Notepad++.
2. Go to `(menu) Settings -> Preferences -> General -> Localization` and set the UI language to something different than English.
3. Close Notepad++.
4. Restart Notepad++.
5. Go to `(menu) File -> Recent Files -> Open All Recent Files` (i.e. its translated pendant, **please note that the menu entry `Recent Files` is translated now**).
6. After all the files from your recent files list have been opened, have a look at the `File` main menu. The entry `Recent Files` (i.e. its translated pendant) has disappeared.
7. Close one of the opened files.
8. Have a look at the `File` main menu again. This time the `Recent Files` entry has reappeared but it is not translated.

### Screenshots
![grafik](https://user-images.githubusercontent.com/24699812/103478080-a3aec200-4dc4-11eb-8846-df26d8ce795b.png)

![grafik](https://user-images.githubusercontent.com/24699812/103478090-b0cbb100-4dc4-11eb-8fba-2551940350e7.png)

![grafik](https://user-images.githubusercontent.com/24699812/103478095-ba551900-4dc4-11eb-9eec-5f0b723ccd18.png)

### Expected Behavior
The `File -> Recent Files` menu entry should **always** be translated into the configured UI language.

### Actual Behavior
Following the test sequence above, the `File -> Recent Files` menu entry will be **not** translated in step 8.

### Debug Information
Notepad++ v7.9.2   (32-bit)
Build time : Dec 31 2020 - 03:58:36
Path : E:\Desktop\Notepad++ Portable\npp.7.9.2.bin.x86\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.1256
Current ANSI codepage : 1252
[...]

PR #13208, opened 2023-02-24
Fix Recent Files lost translation

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9343.

### DIFF
--- a/PowerEditor/src/localization.cpp
+++ b/PowerEditor/src/localization.cpp
@@ -144,14 +144,14 @@ void NativeLangSpeaker::init(TiXmlDocumentA *nativeLangDocRootA, bool loadIfEngl
     }
 }
 
-generic_string NativeLangSpeaker::getSpecialMenuEntryName(const char *entryName) const
+generic_string NativeLangSpeaker::getSpecialMenuEntryName(const char *entryName, const char *attrName, const char *nodeName) const
 {
 	if (!_nativeLangA) return TEXT("");
 	TiXmlNodeA *mainMenu = _nativeLangA->FirstChild("Menu");
 	if (!mainMenu) return TEXT("");
 	mainMenu = mainMenu->FirstChild("Main");
 	if (!mainMenu) return TEXT("");
-	TiXmlNodeA *entriesRoot = mainMenu->FirstChild("Entries");
+	TiXmlNodeA *entriesRoot = mainMenu->FirstChild(nodeName);
 	if (!entriesRoot) return TEXT("");
 
 	WcharMbcsConvertor& wmc = WcharMbcsConvertor::getInstance();
@@ -162,7 +162,7 @@ generic_string NativeLangSpeaker::getSpecialMenuEntryName(const char *entryName)
 	{
 		TiXmlElementA *element = childNode->ToElement();
 
-		const char *idName = element->Attribute("idName");
+		const char *idName = element->Attribute(attrName);
 		if (idName)
 		{
 			const char *name = element->Attribute("name");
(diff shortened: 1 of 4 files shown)


Yaron10:
@ArkadiuszMichalski,

Great. 👍 
Thank you for your work.

PowerEditor/src/localization.h
```diff
@@ -45,7 +45,7 @@ class NativeLangSpeaker {
 	TiXmlNodeA * searchDlgNode(TiXmlNodeA *node, const char *dlgTagName);
 	bool changeDlgLang(HWND hDlg, const char *dlgTagName, char *title = NULL, size_t titleMaxSize = 0);
 	void changeLangTabDropContextMenu(HMENU hCM);
-	generic_string getSpecialMenuEntryName(const char *entryName) const;
+	generic_string getSpecialMenuEntryName(const char *entryName, const char *attrName = "menuId", const char *nodeName = "Entries") const;
```
donho, line 48: This function was coded with the old XML structure of english.xml. So we can
1. Remove the added parameters
2. Rename `entryName` to `nodeName`
3. Rename `getSpecialMenuEntryName` to `getSubMenuEntryName`
4. ...

PowerEditor/src/localization.cpp
```diff
@@ -144,14 +144,14 @@ void NativeLangSpeaker::init(TiXmlDocumentA *nativeLangDocRootA, bool loadIfEngl
...
-generic_string NativeLangSpeaker::getSpecialMenuEntryName(const char *entryName) const
+generic_string NativeLangSpeaker::getSpecialMenuEntryName(const char *entryName, const char *attrName, const char *nodeName) const
 {
 	if (!_nativeLangA) return TEXT("");
 	TiXmlNodeA *mainMenu = _nativeLangA->FirstChild("Menu");
 	if (!mainMenu) return TEXT("");
 	mainMenu = mainMenu->FirstChild("Main");
 	if (!mainMenu) return TEXT("");
-	TiXmlNodeA *entriesRoot = mainMenu->FirstChild("Entries");
+	TiXmlNodeA *entriesRoot = mainMenu->FirstChild(nodeName);
```
donho, line 154: 4. ... use `"SubEntries"` instead of `"Entries"`
5. ...

PowerEditor/src/localization.cpp
```diff
@@ -162,7 +162,7 @@ generic_string NativeLangSpeaker::getSpecialMenuEntryName(const char *entryName)
 	{
 		TiXmlElementA *element = childNode->ToElement();
 
-		const char *idName = element->Attribute("idName");
+		const char *idName = element->Attribute(attrName);
```
donho, line 165: 5. ... use `"subMenuId"` instead of `"idName"`

donho (changes requested):
Please follow the suggestions

PowerEditor/src/localization.h
ArkadiuszMichalski: @donho 

I extended this function to work for any node in `<main>` (because such type doesn't exist), but if you want only for `<SubEntries>` then could be simplified.

donho: @ArkadiuszMichalski 
Yes, I did understand your intention. But this function is called only for recent files entry and I don't see the need for the other menu entry, so let's make it straight and simple.

donho (approved):

Yaron10:
@donho,

Thank you for the optimization. 👍 

@ArkadiuszMichalski,

Thanks again for the fix. 👍

PR #13211, opened 2023-02-24
[xml] Update czech.xml to v8.5

xomx:


donho:
@xomx 
Every branch should be unique.
Please create a new PR with a new branch.

Issue #13201, opened 2023-02-23
WANTED: "Edit with Notepad++" translation


donho:
[New explorer context menu entry **"Edit with Notepad++"**](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0031159cd15e2798996f53c05d98b10ec69ead0) has been added for Windows 11. This feature will be in the installer of next release. 

RP request: please find your language section to translate  **"Edit with Notepad++"** into your language in the following file:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/explorerContextMenuEntryLocal.ini

rddim:
@donho 

Please, sort languages alphabetically

ozone10:
Using `.gitattributes` file in root with
```
explorerContextMenuEntryLocal.ini   text diff working-tree-encoding=UTF-16LE-BOM eol=crlf
```

could help git process `explorerContextMenuEntryLocal.ini`  as text

maisondasilva:
[brazilian_portuguese.xml]
Edit_with_Notepad++=Editar com Notepad++

Regmos:
[danish.xml]
Edit_with_Notepad++=Redigér med Notepad++

Edditoria:
[hongKongCantonese.xml]
Edit_with_Notepad++=用 Notepad++ 編輯

donho:
> [hongKongCantonese.xml] Edit_with_Notepad++=用 Notepad++ 編輯

@Edditoria Unfortunately, it's limited by the language supported by NSIS installer.
- this entry is used when users chooses the language from the beginning of installation.
Thank you for your translation anyway.

IoeCmcomc:
[vietnamese.xml]
Edit_with_Notepad++=Sửa bằng Notepad++

GurliGebis:
@donho If the installer can write it some where in the registry, I'll update the handler to read it from there 🙂

GurliGebis:
@donho that would be an easy way to do it 🙂

SaiyajinK:
I've been waiting for this for a long time!
I use [Shell](https://github.com/moudey/Shell) to rename the entry and it works fine but natively it will be even better!

[**french.xml**]

`Edit_with_Notepad++=Éditer avec Notepad++`

or (with html special characters)

`Edit_with_Notepad++=&Eacute;diter avec Notepad++`

[...]

PR #13212, opened 2023-02-24
Czech translation of context menu entry

xomx:
translation for #13201


ArkadiuszMichalski:
@donho 
Treat all `ini` files as not requiring full compilation is safe? Are there any `ini` files whose contents affect compilation? If there are such files, alternatively we can add `nativeLang` folder to the exceptions (and only its `ini` file will be treated differently).

donho:
@ArkadiuszMichalski 
> Treat all ini files as not requiring full compilation is safe?
> Are there any ini files whose contents affect compilation? If there are such files, alternatively we can add nativeLang folder to the exceptions (and only its ini file will be treated differently).

Not sure why the term "safe" is used. There's no security issue AFAICT.
Also, this file can be exclude from our CI system, since it's needed only by NSIS build for the installer.

ArkadiuszMichalski:
@donho 
I ask if all `ini` files may be excluded or only this one?

PR #13213, opened 2023-02-24
Add Indonesian Context Menu Entry

sahidalizakaria:
Translation for #13201


donho:
You should create a new branch with a unique name for each PR.

PR #13217, opened 2023-02-24
Add ini files to exclusions for AppVeyor

ArkadiuszMichalski:
Changing this file [explorerContextMenuEntryLocal.ini](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/explorerContextMenuEntryLocal.ini) unnecessary run full build. INI files can be treated the same as `md`, `txt` or `log`.

But if for some reason all of them cannot be omitted, only this one can be added to the exceptions.


donho:
Thank you @ArkadiuszMichalski  !

PR #13223, opened 2023-02-24
Let Scintilla manage hidden lines separator; free marker 17

### LINKED ISSUE #12109: Refactor to simpler code and let Scintilla handle the hidden-lines separator line
This issue is being opened to explore the discussion begun [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12046#issuecomment-1230269665) and continued [THERE](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12046#issuecomment-1231100389).

Comments are based upon behavior current to N++ 8.4.4 and some possible new behavior.

It boils down to this:  Scintilla has the capability to display the green separator line that appears across the text viewable area when lines are hidden.  In current N++, this line is created by the N++ code itself by the dedication of a "marker" to it, specifically marker number 21 (in the code: `MARK_HIDELINESUNDERLINE = 21;`).

The N++ code can be simplified, and a "marker" freed up, by turning over the handling of this separator line to Scintilla.

The effect can be demo'd by starting with the 8.4.4 source code and:

* removing all code that currently references `MARK_HIDELINESUNDERLINE` (after the removal this identifier will not exist in the codebase at all)
* adding the following code at a convenient place (e.g. [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6d36d9d398497a6914c64a5804f0400b8bcd6fbd/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L2656) ):
  ```
  #define RGBA(r,g,b,a) ((a<<24) | RGB(r,g,b))
  long rgba = RGBA(0x77, 0xCC, 0x77, 0xFF);
  execute(SCI_SETELEMENTCOLOUR, SC_ELEMENT_HIDDEN_LINE, rgba);
  ```

Here's the effect of that demo (Scintilla-handled separator line):

 image

Compare to the N++  marker-drawn version from 8.4.4:

 image

I think the separator line handled by Scintilla is a bit less intrusive than the one from N++, because it is a bit "lighter weight".

alankilborn:
Fix #12109

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
@@ -117,7 +117,7 @@ const int MARK_BOOKMARK = 20;
 const int MARK_HIDELINESBEGIN = 19;
 const int MARK_HIDELINESEND = 18;
 // 20 - 18 reserved for Notepad++ internal used
-// 16 - 0  are free to use for plugins
+// 17 - 0  are free to use for plugins
 
 const std::vector<std::vector<const char*>> g_nonPrintingChars =
 {
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
@@ -116,8 +116,7 @@ const UCHAR BASE_02 = 0x03; // Bin
 const int MARK_BOOKMARK = 20;
 const int MARK_HIDELINESBEGIN = 19;
 const int MARK_HIDELINESEND = 18;
-const int MARK_HIDELINESUNDERLINE = 17;
-// 20 - 17 reserved for Notepad++ internal used
+// 20 - 18 reserved for Notepad++ internal used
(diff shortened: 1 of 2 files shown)


PowerEditor/src/ScintillaComponent/ScintillaEditView.h
```diff
@@ -116,8 +116,7 @@ const UCHAR BASE_02 = 0x03; // Bin
 const int MARK_BOOKMARK = 20;
 const int MARK_HIDELINESBEGIN = 19;
 const int MARK_HIDELINESEND = 18;
-const int MARK_HIDELINESUNDERLINE = 17;
-// 20 - 17 reserved for Notepad++ internal used
+// 20 - 18 reserved for Notepad++ internal used
 // 16 - 0  are free to use for plugins
```
CennoxX, line 120: shouldn't it be "17 - 0 are free to use for plugins"?

alankilborn:
@donho 
Thank you for agreeing with my code change and accepting my PR!

PR #13224, opened 2023-02-24
Change workspace "save as" behavior

### LINKED ISSUE #13086: Workspace file SaveAs should default to same folder as existing workspace file location
### Description of the Issue
See StR.

Note for what follows:  **_The use of the "dot workspace" extension on a workspace filename is my own convention._**

1. On a Project Panel, create and save a .workspace file (e.g. `foo.workspace`) to a desired location on disk.
2. Create a new editor window tab and save it to disk, in a folder that is NOT the same as the folder used for the .workspace file in step 1.
3. Right-click the workspace and choose _Save As..._ (or _Save a Copy As..._) from the ensuing menu:
image

### Expected Behavior

I expect the standard Windows SaveAs dialog to open up with the directory of the existing workspace as the default, and the existing .workspace file to be seen in the files area of that dialog.

### Actual Behavior

The standard Windows SaveAs dialog opens up to the directory of the active file tab of the editor window.

### Debug Information
Notepad++ v8.4.9 (64-bit)
Path : W:\npp.misc_NoBackup\Releases\npp.8.4.9\npp.8.4.9.portable.x64\notepad++.exe
Command Line : -titleAdd="8.4.9 64-bit"
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2486
Current ANSI codepage : 1252
Plugins :
mimeTools (2.9)
NppConverter (4.5)
NppExport (0.4)
PythonScript (2)

ripmoneygone2020:
Fix #13086 
Change workspace "save as" behavior to include setting the custom file dialog current directory, file name and extension using currently loaded workspace.

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -699,7 +699,7 @@ public:
 		if (SUCCEEDED(hr) && _defExt && _defExt[0] != '\0')
 			hr = _dialog->SetDefaultExtension(_defExt);
 
-		if (SUCCEEDED(hr) && _initialFileName)
+		if (SUCCEEDED(hr) && !_initialFileName.empty())
 		{
 			generic_string newFileName = _initialFileName;
 			if (_fileTypeIndex >= 0 && _fileTypeIndex < static_cast<int>(_filterSpec.size()))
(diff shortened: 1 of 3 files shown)


donho:
Not sure that #13936 is a "bug" (see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13086#issuecomment-1652759437).
If it will be "fixed", then there's a much simpler version here: #13936

PR #13226, opened 2023-02-24
[xml] Update explorerContextMenuEntryLocal.ini for Polish

ArkadiuszMichalski:
@donho 

This file can't be stored in UTF-8 (as other) and convert to UTF-16LE with BOM when generate installer? Now people can't make changes directly from Github because it does the conversion. And after sending the PR is treated as binary (we do not see modifications)

image


donho:
@ArkadiuszMichalski 
Thank you for your PR.
Due to the repeated conflict, I'll integrate myself.
Could you provide the translation here please?

ArkadiuszMichalski:
>Could you provide the translation here please?
```
[polish.xml]
Edit_with_Notepad++=Edytuj w Notepad++
```

donho:
Thank you!

PR #13232, opened 2023-02-25
Add german translation for Explorer context menu entry

dinkumoil:


donho:
Thank you @dinkumoil 
Due to the repeated conflict, I will integrate it myself.

PR #13234, opened 2023-02-25
Release 'data'

mjunix:


alankilborn:
See guidelines for contributing, in `CONTRIBUTING.md` in the root level of the project code.

chcg:
Moving the allocation down from line 804 -> 822 seems to be a better option to avoid this mem leak issue.

donho:
Use solution of @chcg here: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d4d4163cb3d1f827efd9ea2df9dd9838d12f5d72

PR #13236, opened 2023-02-25
Check index before using it

mjunix:


alankilborn:
@mjunix See guidelines for contributing, in `CONTRIBUTING.md` in the root level of the project code.

chcg:
@mjunix As this is scintilla code it should be reported here: https://sourceforge.net/p/scintilla/bugs/

donho:
Tourists are not welcome here.

PR #13260, opened 2023-02-27
Make Find-in-Projects a command

### LINKED ISSUE #13256: Find in Projects should be a command (so that it can be assigned a shortcut)
1. Look in _Search_ menu for a _Find in Projects..._ command (that when invoked would open the main _Find_ window  and activate the _Find in Projects_ tab), so that it will be available in the _Shortcut Mapper_ and I can assign my own custom keycombo to it.

### Expected Behavior
That I would locate _Find in Projects_ on the _Search_ menu, perhaps in either the indicated red (preferred) or blue (secondary) positions?:

image

### Actual Behavior
_Find in Projects_ command not found.

### Debug Information
Comments based on behavior current to Notepad++ 8.4.9.

alankilborn:
Fix #13256


alankilborn:
Rejection reason in issue:

`The reason that Find in project is not in the command menu is because without project context this action is meaningless.
This command is available on the context menu of project panel which makes sense, but not alone in the menu.
Hence the rejection.`

@donho

With due respect, IMO this rejection makes little sense, with the following 2 items offered as supporting evidence:

* Since _Find in Projects_ was implemented, as a user I have been forced to look at a _Find in Projects_ tab in the _Find_ window, even though (until recently) I haven't had a project panel open.  So how is this consistent with the rejection reason of this issue?  **SUMMARY:** If there is no "project context" (by virtue of no project panels open), why am I forced to see this _Find in Projects_ tab -- it should be invisible if I have no project panel open, right?  A consequence of this point: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13266

* As a dedicated keyboardist, having to grab the mouse and slog it over to the workspace/project panel, right-click on the workspace name and choose _Find in Projects_ (counting on my remembering that this is how it is done) is just, well, ugh, too much extra effort.  For keyboardists, there should be single keystroke access to _Find in Projects_ -- can this possibly be disputed?  A keyboarding user can't even do a workaround of Shift+Ctrl+f to invoke _Find in Files_ and then change to _Find in Projects_, because there is no way to switch tabs in _Find_ family of windows without grabbing for the mouse.  If the keyboardist needs to grab the mouse, they might as well directly click the _Find in Projects_ tab -- but this 100% destroys a keyboarding workflow.  **SUMMARY:** project searching remains klunky for keyboardists.

This PR took maybe 10 minutes of effort, so I'm not bothered by the "wasted time" result of rejection.  What bothers me is that the software is not as good as it could be without it.  Sure, I can and will script a workaround, and post it later in this thread, but this doesn't aid other users that don't know the script exists, or don't want to go to the trouble of setting it up, or are just generally annoyed by workarounds for missing things in the core software.

Have you ever really tried to use the "projects" concept in Notepad++?  As a user would?

I have recently had a need for a project, and so I've set one up and started real-world use.  My conclusion is that project handling in Notepad++ is not very good.  Thus, recently I've put in a flurry of issues regarding projects to get some attention to the deficiencies.  As this current issue was a smaller one, my skills were up to its 10-minute task.  Some of the other ones, if liked, would have to be dealt with by others with more skills.  And with ~1900 other open issues, perhaps I have low optimism for "projects" improving.  I suppose I'll keep trying, though... :-)

donho:
> And with ~1900 other open issues, perhaps I have low optimism for "projects" improving. I suppose I'll keep trying, though... :-)

I'm sorry for rejecting this PR and for wasting your time. I would make every contributor happy by accepting every single suggestion and PR, but I can't. If I did, then this project would be a big mess.
To be fair, OTOH, although there are so many rejected PR, there are many more accepted PR, so don't take it personal.

alankilborn:
>I'm sorry for rejecting this PR and for wasting your time.

It didn't waste my time -- I went out of my way to state that earlier.

Probably a more constructive use of everyone's time would have been to actually read and consider the points I made, and then come back with a really sensical rebuttals, but that didn't happen.

So, OK, I'll take the overall response as "projects" are always going to be poorly-handled in Notepad++, so I've changed my mind and won't bother attempting to make them better for other users, and won't be using them myself.

I will, however, share the script I mentioned before leaving "projects" alone for good.

ArkadiuszMichalski:
@alankilborn 
>A keyboarding user can't even do a workaround of Shift+Ctrl+f to invoke Find in Files and then change to Find in Projects, because there is no way to switch tabs in Find family of windows without grabbing for the mouse. 

`Shift+Ctrl+F > Shift+Tab > Right arrow`. I know it's not a dedicated shortcut, but there is a workaround.

alankilborn:
>Shift+Ctrl+F > Shift+Tab > Right arrow. I know it's not a dedicated shortcut, but there is a workaround.

OK, @ArkadiuszMichalski , I stand corrected.  Thank you for that.
It can be done with the keyboard in that workaround fashion.
But it is a burdensome workaround.

I really don't understand why we can't have one-keycombo access to project-wide searching.
It makes no sense and pretty much kills the usefulness of the project concept.

It really would be great if the author had a small group of "trusted advisors" that might say "hey, this something like the declined feature 'X' really makes sense, maybe you should reconsider".
No one person has the best idea all the time.  Heck, I know my ideas aren't best all the time, either (except in this case, I know I'm right, as it is a feature-limiting problem).
"Trusted advisors" might be appointed team members and/or people from the Community site with, I don't know, around 10000 reputation points. :-)
Oh well, letting it, and the possible use of projects for me, go now...

alankilborn:
As promised, a scripting solution:

Taking a cue from the @ArkadiuszMichalski multi-keycombo sequence, here's a simple PythonScript that can activate the _Find in Projects_ tab from the editor via a single keycombo (that invokes the script):

```
import SendKeys as sk
sk.SendKeys('^m+{TAB}{LEFT}')
```

Explanation:  Uses Ctrl+m to invoke _Mark_ via its default keymapping, then does Shift+Tab to get the _Mark_ tab control the input focus, then executes the left-arrow to switch to _Find in Projects_.

Certainly other approaches with this type of keystroke simulation are valid, e.g. AutoHotKey, I'd think (haven't tried).

PR #13265, opened 2023-02-28
Add translation of 'Edit with Notepad++'

donho:
Explorer context menu item under Windows 11 can be translated thanks to https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0031159cd15e2798996f53c05d98b10ec69ead0

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13201

People can also add translation in the comment below - I will add them into this PR.

### DIFF
--- a/PowerEditor/installer/nativeLang/explorerContextMenuEntryLocal.ini
+++ b/PowerEditor/installer/nativeLang/explorerContextMenuEntryLocal.ini
@@ -32,7 +32,7 @@ Edit_with_Notepad++=Mit Notepad++ öffnen
 Edit_with_Notepad++=Editar con Notepad++
 
 [italian.xml]
-Edit_with_Notepad++=Apri con Notepad++
+Edit_with_Notepad++=Modifica con Notepad++
 
 [portuguese.xml]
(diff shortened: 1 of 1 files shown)


dfs-:
Literal German translation would be „Mit Notepad++ bearbeiten“. The current translation is also okay meaning „Open in Notepad++“.

conky77:
I've already translated that. However testing the Release Candidate and comparing other windows programs, a better translation is:
[italian.xml]
Edit_with_Notepad++=Modifica con Notepad++

donho:
@dfs- 
Then it'll be "**Mit Notepad++ öffnen**" vs "**Mit Notepad++ bearbeiten**".
Is there anyone who seconds that "**Mit Notepad++ bearbeiten**" is better?

> I wonder how the accelerator key is decided?

I tried to figure out how to trigger the accelerator key of context menu commands for the Windows 11, without success.
If you can point me out how to use the accelerator key of context menu commands for the Windows 11, I will make it available.

ArkadiuszMichalski:
@donho If you add `[xml]` to commit title then only 1 job will be run without any action.

Issue #13166, opened 2023-02-18
window Command line Arguments is untranslatable


conky77:
### Description of the Issue

Window "Command line Arguments" is untranslatable
 
### Steps to Reproduce the Issue

1. click on "?", "Command line Arguments"
2. All strings are untranslated

### Expected Behavior

It should be possible to translate:
- title
- parameter description: every parameter should have a different entry in english.xml (string with only explanation, without parameter name)
- buttons

### Actual Behavior

All strings are untranslated

### Debug Information

Notepad++ v8.4.9   (64-bit)
Build time : Jan 27 2023 - 03:11:16
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 22H2
OS Build : 22621.1265
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    CSVLint (0.4.6.2)
    DSpellCheck (1.4.24)
    JsonTools (4.8.1)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NPPJSONViewer (2.0.4)
    NppXmlTreeviewPlugin (2)
    XMLTools (3.1.1.13)

![immagine](https://user-images.githubusercontent.com/59116246/219875934-0fbf03d0-c522-48d1-a75d-a7bef748b008.png)

conky77:
#8972

ArkadiuszMichalski:
@conky77 

You can test if [PR 13269](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13269) works as expected.

conky77:
@ArkadiuszMichalski Since I can't compile the project and test, I can just check the english.xml. From the point of view of a translator should be ok: you can translate all strings and it is not possible to miss or confuse strings.

ArkadiuszMichalski:
@conky77 
You don't need compile yourself, just use build from AppVeyor:
https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/46369591

Instruction:
https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/Testing#using-a-pull-request-pr-build-version-of-notepad

conky77:
[...]

PR #13269, opened 2023-03-01
Translate Command Argument Help dialog

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13166.

This can be done in many ways. In this approach:
- descriptions can be changed
- example values can be changed
- works with a `-L` argument
- the general format cannot be changed (there is a fixed template)
- it's easier to add something new because we don't have to play with reading/combining data in function (just add the appropriate entries in template/data/xml).


donho:
@ArkadiuszMichalski 
Unfortunately this PR cannot be accepted for the following reasons:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13166#issuecomment-1454836182

Sorry for your wasting time. Hope you understand it.

PR #13283, opened 2023-03-04
Update Hindi translation for (Edit with Notepad++)

SinghRajenM:


donho:
@SinghRajenM 
As I mentioned " I'll integrate it myself due to the repeated conflict." in #13201, it is added in PR #13265.
Thank you for the translation.

Issue #13286, opened 2023-03-04
Tooltips are not updated after changing translation


ArkadiuszMichalski:
### Description of the Issue
When we change translation all tooltips from `Preferences` and `Find/Replace` dialogs are not update. It's a bit annoying, especially when we doing a translation because we have to restart Notepad++ (and we don't always remember to do that and wonder why our translation doesn't work).

### Steps to Reproduce the Issue

1. Open Notepad++.
2. Open `Preferences` dialog and check tooltips (move mouse over `[?]` in `Editing/Delimiter/Performance`).
3. Check tooltips also in `Find/Replace` dialog.
4. Change translation to other and once again check above tooltips.

### Expected Behavior
Tooltips text should be update when we switch translation.

### Actual Behavior
Tooltips text is  not update when we switch translation.

Tooltips from other places are done differently and their text is updated when we switch translation.

### Debug Information
```
Notepad++ v8.4.9   (64-bit)
Build time : Mar  3 2023 - 18:39:25
Built with : GCC 12.2.0
```

alankilborn:
@donho said "not a bug to me" -- see [HERE](https://community.notepad-plus-plus.org/post/84525).

donho:
I personally don't fix that - but if you do and the solution is quite simple, it will be accepted.

donho:
Please see:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13288#issuecomment-1454937340

PR #13288, opened 2023-03-04
Update tooltips when change translation

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13286.


donho:
The PR implicates too much modification, for changing tooltip - which is not useful for the most of users. So as I said here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13286#issuecomment-1454824387, it won't be accepted.

Sorry for your wasted time, and thank you for your contribution anyway.

Issue #13280, opened 2023-03-03
Possible wrong value in "Custom Large File Size"


Yaron10:
**STR:**
Preferences -> Performance.
Set "Define Large File Size" to `3000`.
Exit NPP.
Restart NPP.
Preferences -> Performance.

**Result:**
"Custom Large File Size" is `4096`.

Exit NPP.
Restart NPP.
Preferences -> Performance.

**Result:**
"Custom Large File Size" is `200`.

**Suggestion:** 
Replace
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/081e8ad11b019a4cac082416ba855a98549a4fe1/PowerEditor/src/Parameters.cpp#L5734
with
```
if (fileSizeLimit4StylingMB > 0 && fileSizeLimit4StylingMB <= 4096)
	_nppGUI._largeFileRestriction._largeFileSizeDefInByte = (static_cast<int64_t>(fileSizeLimit4StylingMB) * 1024) * 1024;
```

***
Notepad++ v8.4.9   (64-bit)
Build time : Feb 25 2023 - 18:00:24
Path : C---\Notepad++.x64.Release.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 20H2
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : 
    ComparePlus (1.1)
    HTMLTag (1.3.6)
    MenuIcons (2.0.2)
    PythonScript (2)
    _CustomizeToolbar (4.2)

PR #13290, opened 2023-03-04
Use correct type for file size variables

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13280.

@donho 
Max value `4096` is store in XML file (and input accept it without any correct) but when read from XML file it's convert to default 200MB. I fixed it too (just use `<=` instead of `<` in one place).


Yaron10:
@ArkadiuszMichalski,

👍 
Thank you for the quick fix.

Issue #13178, opened 2023-02-19
[Feature Request] - Option to hide prompt when closing an unsaved file


vbergaaa:
### Description of the Issue
Needing to close multiple tabs that you don't want to save changes to is tedious due to the multiple prompts that appear when trying to close these tabs

### Use Case
I often use notepad++ as a scrapbook to write down notes I need to store for a limited time. The duration could range from 10s to a week.
When needing to create a note, I open Notepad++ and create a new tab to store the information. The note is named 'New n' for the duration of it's life, and because I know it's a temp file, I don't want to save it, I just rely on Notepad++ to store it for it's duration.
I might create 10-20 tabs like this during a day. This quickly builds up, and at so at the end of the day (or week), I often go through each file and decide if it's ready to be deleted or not.
If its something I want to keep for another day, I go to the next tab, if not, I hit the x to delete it, then have to navigate to the prompt to hit 'no' because it's unsaved.
This is the tedious part and the part I wish I didn't have to do. I'm more than aware I'm deleting an unsaved file and that I'll be losing those changes, having a prompt appear in the middle of the window every time is unnecessary to me. 

### Proposed Feature
Add some type of optional way that a user can delete the file without being prompted about losing the unsaved changes
A few possible ideas (with my prefered one at the top):
- Hold Shift while clicking close to close without prompt (and have a tooltip when the prompt appears saying 'hold shift while clicking to force delete without this prompt' (Discord does this and it's very effective).
- A checkbox (or two) on the prompt that says ("do not show this prompt again" and/or "do not show this prompt again for new files")
- A user preference you can navigate to that allows the prompt to not appear for new files.

Thank you for your consideration

alankilborn:
>Hold Shift while clicking close to close without prompt

Seems like a good idea.

doug1234:
I was able to add the shift+click feature @alankilborn but do you think it is a useful addition without some keyboard way of doing it?  Also, how would you document the feature's existence?

alankilborn:
@doug1234 

>but do you think it is a useful addition without some keyboard way of doing it

I'm not a good judge of this.  :-)
The procedure is to submit a PR for the feature and let the Notepad++ author decide.

>how would you document the feature's existence?

In the Notepad++ user manual, but no point in talking about that unless a change is going to be accepted.

Good luck.

doug1234:
Thank you!  I will get a PR together and see.

alankilborn:
[...]

PR #13291, opened 2023-03-04
Improvement for issue #13178.  Shift click to quickly close a tab.

doug1234:
Implementing the proposed feature for issue #13178:  Hold Shift while clicking close to close a tab without a dialog to ask if you want to save a modified file.

I can add documentation if this PR is approved.

### DIFF
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -170,7 +170,7 @@ public:
 	void fileOpen();
 	void fileNew();
     bool fileReload();
-	bool fileClose(BufferID id = BUFFER_INVALID, int curView = -1);	//use curView to override view to close from
+	bool fileClose(BufferID id = BUFFER_INVALID, int curView = -1, bool promptIfDirty = true);	//use curView to override view to close from
 	bool fileCloseAll(bool doDeleteBackup, bool isSnapshotMode = false);
 	bool fileCloseAllButCurrent();
 	bool fileCloseAllGiven(const std::vector<int>& krvecBufferIndexes);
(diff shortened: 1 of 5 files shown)


dfs-:
Does this work with CapsLock active? If the user accidentally activated it, it would be an user guidance issue to not ask if the file needs to be saved and is then gone forever.

alankilborn:
>Does this work with CapsLock active?

Caps Lock is absolutely not involved.

doug1234:
> > Does this work with CapsLock active?
> 
> Caps Lock is absolutely not involved.

Correct.  The shift button needs to be pressed.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13178#issuecomment-1464743469

Explorer09:
I personally don't like the idea either, but for the use case of the original post, may I propose an alternative way of doing the thing (closing multiple tabs without saving)?

Make the workflow something like this:
1. Open a dialog that list all tabs you opened in Notepad++ (I think this exists in the "Windows" menu of the menu bar already)
2. In the dialog, hold Shift and click on the tab/items you want to discard changes of. Note that this step only selects the tabs and nothing would be closed right away.
3. Click the "Close Without Saving" button on the list-of-tabs dialog.
4. There would be a confirmation prompt, click "Yes" or "Close Without Saving" to really close them, all at once.

alankilborn:
@Explorer09 said:

>Click the "Close Without Saving" button on the list-of-tabs dialog.

Where is this button?
Or is this a _proposed_ idea...to have such a button?

donho:
@Explorer09 
Yes, it's a better GUI solution. Could you create a feature request?

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13178#issuecomment-1585842456

Issue #13186, opened 2023-02-21
Change "Replace by" wording in Tab Settings


alankilborn:
### Description of the Issue
It is possible for a user to misinterpret the meaning of the _Replace by space_ control on _Settings_ (menu) > _Preferences..._ > _Language_ > _Tab Settings_.

The use of "replace by" could be interpreted to mean that, when the box is transitioned from unchecked to checked by the user, a replace (in the sense of _Search_ (menu) > _Replace..._ > _Replace All_) will be executed for the current document.

Ref.: #13182

Suggestions for change:

* _Replace by space_ --> "Tab press inserts space(s)"
* _Replace by space_ --> "Tab key inserts space(s)"

### Debug Information
Above comment made on behavior current to N++ v. 8.4.9.

### Screenshot
![image](https://user-images.githubusercontent.com/77065706/220342715-fc5b41e3-668d-455c-bdaa-a5b9b58bd217.png)

ArkadiuszMichalski:
But `Tab size` can also be changed from `4` to another value, and it doesn't mean conversion in the whole document. For me the current form is clear. Well, but if there is space and the new description will be more precise, it may be worth changing.

alankilborn:
@ArkadiuszMichalski

Why did you tag this with a "translation" label?
It isn't a translation issue.
If it is accepted and implemented then the code and `english.xml` will be changed.
Only AFTER that point does it becomes a translation issue.
Or do I misunderstand the usage of the "translation" label?

>For me the current form is clear.
Apparently it is not clear for all users.  Sure it is impossible to satisfy everyone, but sometimes it is best to reflect and think: "Could this be improved and is it worth it to do so?"  In this case, my opinion is yes.

pryrt:
I would vote for "Tab key inserts space(s)".  

(And I am not offended by the "translation" label, because changing this will _affect_ translation: looking at the other issues that have that tag, it seems to be both submissions for translation, and changes that will create a need for translations to be updated; since this is changing the English phrasing, it creates a need for the translators to assess whether there is the same translation need in other languages, or whether their existing phrasing already leaves no ambiguity.)

alankilborn:
>And I am not offended by the "translation" label

Well, "offended" would not describe my reaction either. :-)
But my conception of the "translation" label is that it cues translators that something needs translating (maybe I'm wrong?).  As nothing has been changed (yet) on this, why cue them?
[...]

PR #13297, opened 2023-03-06
[xml] Change "Replace by" wording in Tab Settings

ArkadiuszMichalski:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13186.

Actual text:
image

New variant from this PR:
image

Other proposed text:
image

There is some room for a more informative description.


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13186#issuecomment-1467187081

PR #13299, opened 2023-03-06
[xml] Update Bulgarian localization

rddim:
* Replace recent file ValueDlg with edit fields & fix DocSwitcher RTL problem · notepad-plus-plus/notepad-plus-plus@269e78b
* Make tray icon context menu translatable · notepad-plus-plus/notepad-plus-plus@52d3c36
* Add ability to copy "Find what" to "Replace with" and vice versa · notepad-plus-plus/notepad-plus-plus@12f649b


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d060b0f01a60882726850ba7298e4da3e0dcc7de

PR #13303, opened 2023-03-07
Fix Project Workspace "Save a Copy As" confusion issue

### LINKED ISSUE #13135: Project Workspace "Save a Copy As" confuses the GUI's shown workspace name
Per discussion in #13131, there is a definite bug that was brought up, and should be addressed, regardless of the other feature concerns in that previous issue.  

### Description of the Issue

If you use the Project Panel Workspace feature of **Save a Copy As**, it does not behave as expected: it shows that the workspace has changed name (it shouldn't) but is actually still using the old copy of the workspace (which it should do).

### Expected Behavior

When you **Save a Copy As**, there should be a new workspace file created, and Notepad++ should still continue to use the old workspace file and name, and the GUI should display that it's still using the old workspace name.

### Actual Behavior

When you **Save a Copy As**, there is be a new workspace file created (correct), and Notepad++ should still continue to use the old workspace file (correct) and name, but the GUI displays that it's using the new name when it should still show the old name (bug).

If you make a change to the active project (for example, add or remove a file), then save that project's workspace, the GUI suddenly shows the old name again (which it should have been showing all along)

### Steps to Reproduce (STR)

1. View > Project Panels > Project Panel 1
2. Workspace > New Workspace to clear any existing workspace
3. Workspace > Add New Project: `SubProject`
4. Workspace > Save: `one.wkspc` => contents of `one.wkspc` are
    ```
    <NotepadPlus>
          <Project name="SubProject" />
    </NotepadPlus>
    ```
5. Workspace > Save As: `two.wkspc`
   - GUI says `two.wkspc` is active, as it should
6. SubProject > Add File(s): `change.log` and `readme.txt`
7. Workspace > Save => contents of `two.wkspc` are now:
    ```
    <NotepadPlus>
        <Project name="SubProject">
            <File name="change.log" />
            <File name="readme.txt" />
        </Project>
    </NotepadPlus>
    ```
8. Workspace > Save a Copy As: `three.wkspc` =>
    - Expectation: GUI will still say `two.wkspc` is active
    - Actual behavior: GUI now says `three.wkspc` is active.  However, it's worse than that.  Continuing the experiment:
9. SubProject > Add File(s): `license.txt`
10. Workspace > Save =>
    - Unexpected: The GUI now says `two.wkspc` is active.   
    - Contents of `two.wkspc`:
        ```
        <NotepadPlus>
            <Project name="SubProject">
                <File name="change.log" />
                <File name="readme.txt" />
                <File name="license.txt" />
            </Project>
        </NotepadPlus>
        ```
    - Contents of `three.wkspc`:
        ```
        <NotepadPlus>
            <Project name="SubProject">
                <File name="change.log" />
                <File name="readme.txt" />
            </Project>
        </NotepadPlus>
        ```

Thus, for step 9-10, it was obviously actually correctly working with `two.wkspc`, but it claimed to be working with `three.wkspc`.

Conclusion: **Save a Copy As** is correctly saving a copy and keeping the old version active, but it incorrectly shows the name as the new version of the file, until the workspace is saved again.

### Debug Information

```
Notepad++ v8.4.9   (64-bit)
Path : C:\usr\local\apps\npp\npp.8.4.9.fresh\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 20H2
OS Build : 19042.2486
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

doug1234:
Project name is no longer updated in the tree when "save copy as" is used.

Fix #13135

### DIFF
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -384,7 +384,7 @@ bool ProjectPanel::saveWorkSpace()
 	} 
 }
 
-bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName)
+bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName, bool updateName)
 {
 	//write <NotepadPlus>: use the default file name if new file name is not given
 	const TCHAR * fn2write = projectFileName?projectFileName:_workSpaceFilePath.c_str();
@@ -429,7 +429,10 @@ bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName)
 		return false;
 	}
 	TCHAR * fileName = PathFindFileName(fn2write);
-	_treeView.renameItem(tvRoot, fileName);
+	if (updateName) 
+	{
+		_treeView.renameItem(tvRoot, fileName);
+	}
 	return true;
 }
 
@@ -1194,7 +1197,7 @@ bool ProjectPanel::saveWorkSpaceAs(bool saveCopyAs)
 	if (fn.empty())
 		return false;
 
-	if (!writeWorkSpace(fn.c_str()))
+	if (!writeWorkSpace(fn.c_str(), !saveCopyAs))
 		return false;
 
 	if (!saveCopyAs)
--- a/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
+++ b/PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
@@ -384,7 +384,7 @@ bool ProjectPanel::saveWorkSpace()
 	} 
 }
 
-bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName, bool updateName)
+bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName, bool doUpdateGUI)
 {
 	//write <NotepadPlus>: use the default file name if new file name is not given
 	const TCHAR * fn2write = projectFileName?projectFileName:_workSpaceFilePath.c_str();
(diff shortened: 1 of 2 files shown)


PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.h
```diff
@@ -125,7 +125,7 @@ class ProjectPanel : public DockingDlgInterface {
 	void recursiveAddFilesFrom(const TCHAR *folderPath, HTREEITEM hTreeItem);
 	HTREEITEM addFolder(HTREEITEM hTreeItem, const TCHAR *folderName);
 
-	bool writeWorkSpace(const TCHAR *projectFileName = NULL);
+	bool writeWorkSpace(const TCHAR *projectFileName = NULL, bool updateName = true);
```
donho, line 128: Please use `doUpdateGUI` instead of "updateName".

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
```diff
@@ -384,7 +384,7 @@ bool ProjectPanel::saveWorkSpace()
 	} 
 }
 
-bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName)
+bool ProjectPanel::writeWorkSpace(const TCHAR *projectFileName, bool updateName)
```
donho, line 387: Please use `doUpdateGUI` instead of "updateName".

donho (changes requested):
Please use `doUpdateGUI` instead of "updateName".

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.h
doug1234: Made the change as suggested.

PowerEditor/src/WinControls/ProjectPanel/ProjectPanel.cpp
doug1234: Made the change as suggested.

donho (approved):

PR #13311, opened 2023-03-08
[XML] Update Turkish translations

bitigchi:
Small improvements, that make it more in-line with the generally accepted Turkish software l10n.


donho:
There's no added entries, but only the modification of existent entries.
Could any Turkish speaker validate the modification?

bitigchi:
Just in case no one shows up, here's a screenshot of Windows Notepad with all the above terminology visible on screen.

Ekran görüntüsü 2023-03-22 200004

Issue #13320, opened 2023-03-09
Latest Notepad++ replaces "Pin to Quick access" in the Windows 11 menu


GurliGebis:
### Description of the Issue
The latest version of Notepad++ replaces the "Pin to Quick access" in the Windows 11 right click menu.
This is due to it being added in the wrong way.

Instead of just overwriting something else, it should be added like described here:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13170#issuecomment-1437229089

The correct way to do it is to use Sparse Manifests: https://blogs.windows.com/blog/2021/07/19/extending-the-context-menu-and-share-dialog-in-windows-11/

Also, please don't close this bug as "expected behavior", since you are not supposed to just overwrite the built-in menu options to add your own.

### Steps to Reproduce the Issue
1.  Install the latest Notepad++ version
2.  Watch as it takes over the "Pin to Quick access" menu option

### Expected Behavior
It shouldn't overwrite the default Windows behavior, but should instead add it's own menu option to the list.

### Actual Behavior
It takes the easy (and wrong) way out, and just replaces the existing menu item, instead of doing it the right way.

wallrik:
Please consider rolling back release 8.5 until this is fixed. 😣 Also, people have left [comments](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0031159cd15e2798996f53c05d98b10ec69ead0#commitcomment-103252580) on the commit and other places as well.

Rolling back the release would stop more people from getting messed up installs, and save a lot of headache.

riverar:
@donho Would consider rolling back a0031159cd15e2798996f53c05d98b10ec69ead0

ErikHumphrey:
Doesn't seem necessary to do it the way it's currently implemented. At minimum, it should be an option; I wouldn't expect this behaviour just by installing Notepad++, especially as uninstalling it doesn't revert the change.

donho:
Stop sharing the following link with me - it says it "should do" but not says "how to"
https://blogs.windows.com/blog/2021/07/19/extending-the-context-menu-and-share-dialog-in-windows-11/

If you have the concrete solution, please let me know.

For people who don't want the context menu, just uncheck **Context Menu Entry** during the installation:

![image](https://user-images.githubusercontent.com/90293/224195477-fecf2d3e-4660-48a0-9b99-690475b23aea.png)

donho:
[...]

PR #13330, opened 2023-03-10
Windows 11 context menu

GurliGebis:
**THIS IS STILL A WORK IN PROGRESS, DO NOT MERGE ANYTHING YET**

This pull request implements the Windows 11 context menu using the correct way of having a DLL file and a sparse package.
Creating draft PR now for people to follow along and comment with suggestions and feedback.

Once this is done, it will fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13320


donho:
@GurliGebis 
Thank you for working on it.
To make sure that you're not wasting your time, could you provide the ref link or screenshot (or even in short description) for showing how it will look like?

dongle-the-gadget:
It seems that the context menu item is not currently implemented yet.

GurliGebis:
> @GurliGebis Thank you for working on it. To make sure that you're not wasting your time, could you provide the ref link or screenshot (or even in short description) for showing how it will look like?

As it is still a work in progress, there is nothing to show yet.
But from how I understand it works, it will be a menu item with the correct icon and the "Edit with Notepad++" text.
The current hack will be reverted (people would have to cleanup manually, since the current hack is not being uninstalled)

dongle-the-gadget:
Also, what's your opinion on using C++/WinRT instead of WRL for the DLL? It generally provides a more modern syntax that more closely resembles newer versions of standard C++.

GurliGebis:
> Also, what's your opinion on using C++/WinRT instead of WRL for the DLL? It generally provides a more modern syntax that more closely resembles newer versions of standard C++.

My c++ is pretty rusty, so I'm following Microsofts documentation on how to call those APIs, and they use winrt for it.
So that's my reason for doing it like that 🙂

GurliGebis:
Just checked, WRL has been superseded by Winrt: https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl?view=msvc-170

contextmenu/dllmain.cpp
```diff
@@ -0,0 +1,178 @@
+// dllmain.cpp : Defines the entry point for the DLL application.
+#include "pch.h"
+
+using namespace Microsoft::WRL;
+using namespace winrt::Windows::Management::Deployment;
```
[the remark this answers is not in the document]
GurliGebis: That way all the installer needs to do is regsvr32 /i on it

sylveon:
Why not get wil via Nuget 8nstead of vendoring it in?

GurliGebis:
> Why not get wil via Nuget instead of vendoring it in?

I didn't want to add a nuget restore step to the build process, since it doesn't use it currently, as far as I can see (please correct me if I'm wrong)

My idea is to make as few changes as needed 🙂

sylveon:
Yeah but it also means that you're getting a +37k lines diff for this PR. One could argue adding a nuget restore step is less changes ;)

GurliGebis:
> Yeah but it also means that you're getting a +37k lines diff for this PR. One could argue adding a nuget restore step is less changes ;)

I agree that it is a bit huge, however, I'll leave it like this for the time being - we can always change it before this is merged, if it shows to be an easy change.

donho:
@GurliGebis 
There are 2 points I have to make it clear:
1. https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0031159cd15e2798996f53c05d98b10ec69ead0 should not to be revert because 
A. not only the "Edit with Notepad++" has been added, but some bugs of installer were fixed by this commit
B. if the new "Edit with Notepad++" is accepted and integrated into the next release, the "Edit with Notepad++" installed by installer of v8.5 will be removed by installer of v8.5.1 - so I will adapt the old code for this purpose.
2. This project should not be compiled with Notepad++ binary and it should have its own project file (if my assumption is correct - it's dependent of notepad++.exe completely) so it can be move into `notepad-plus-plus/PowerEditor/src/tools/` or be an independent project in [Notepad++ organization](https://github.com/notepad-plus-plus).

Anyway, we can see the next action when the new dll is ready to go. 
Feel free to ask me if you are stuck by some problem.

GurliGebis:
@donho Thanks, I'll move it and add it to it's own solution file and drop the revert commit.
I don't think I'll have time to work on it any more this weekend though :)

dongle-the-gadget:
I believe that the hijacking method is a pretty poor way to do it, and that it is better just to come up with a solution to the problems fixed by that hijacking instead of using it.

dongle-the-gadget:
> I didn't want to add a nuget restore step to the build process

IIRC it only involves adding `/restore /p:RestorePackagesConfig=true`

contextmenu/dllmain.cpp
```diff
@@ -0,0 +1,178 @@
...
+	case DLL_PROCESS_ATTACH:
+	case DLL_THREAD_ATTACH:
+	case DLL_THREAD_DETACH:
+	case DLL_PROCESS_DETACH:
+		break;
+	}
+	return TRUE;
+}
+
+class ExplorerCommandBase : public RuntimeClass<RuntimeClassFlags<ClassicCom>, IExplorerCommand, IObjectWithSite> {
```
dongle-the-gadget, line 23: Not sure if `IObjectWithSite` is actually needed here. In the sample project it is added to show message boxes as Explorer. I've tested removing `IObjectWithSite` on another project and it worked.

GurliGebis: I'll try and see about that, thanks

GurliGebis:
> > I didn't want to add a nuget restore step to the build process
> 
> IIRC it only involves adding `/restore /p:RestorePackagesConfig=true`

The project has to compile with both MSVC and GCC, so the problem then becomes - how do I restore nuget packages using GCC?

dongle-the-gadget:
Oh. Unfortunately, C++/WinRT is not supported on GCC.

donho:
@GurliGebis 
> The project has to compile with both MSVC and GCC , so the problem then becomes - how do I restore nuget packages using GCC?

Only Notepad++ binary is needed to be compiled by both MSVC and GCC (for people who don't have or don't want MSVC installed, and also for the cross-checking by 2 compilers for code sanity), that's why this project should be set out of the main project - it's enough to be compiled only by MSVC to get the binary we need.

GurliGebis:
@donho I will once I get it finished - it is still a work in progress.
I was wondering, can we limit it to x64 and arm64? (Since it is only going to be used on Windows 11, which is 64bit only)

donho:
@GurliGebis 
> I was wondering, can we limit it to x64 and arm64? (Since it is only going to be used on Windows 11, which is 64bit only)

Of course - it makes sense!

sylveon:
> Oh. Unfortunately, C++/WinRT is not supported on GCC.

C++/WinRT does support GCC now: https://github.com/microsoft/cppwinrt/pull/1245

contextmenu/dllmain.cpp
GurliGebis: I have removed it now

GurliGebis:
> > I didn't want to add a nuget restore step to the build process
> 
> IIRC it only involves adding `/restore /p:RestorePackagesConfig=true`

Looks like GCC is not required for tools, so I'll see what I can do later on

GurliGebis:
> @GurliGebis
> 
> > I was wondering, can we limit it to x64 and arm64? (Since it is only going to be used on Windows 11, which is 64bit only)
> 
> Of course - it makes sense!

Done 🙂

GurliGebis:
> @GurliGebis Thank you for your explanation. There are 2 points I have to make it clear:
> 
>     1. [a003115](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a0031159cd15e2798996f53c05d98b10ec69ead0) should not to be revert because
>        A. not only the "Edit with Notepad++" has been added, but some bugs of installer were fixed by this commit
>        B. if the new "Edit with Notepad++" is accepted and integrated into the next release, the "Edit with Notepad++" installed by installer of v8.5 will be removed by installer of v8.5.1 - so I will adapt the old code for this purpose.
> 
>     2. This project should not be compiled with Notepad++ binary and it should have its own project file (if my assumption is correct - it's independent of notepad++.exe completely) so it can be move into `notepad-plus-plus/PowerEditor/src/tools/` or be an independent project in [Notepad++ organization](https://github.com/notepad-plus-plus).
> 
> 
> Anyway, we can see the next action together when the new dll is ready to go. Feel free to ask me if you are stuck by some problems (technical or not).

I have moved it now

GurliGebis:
@donho how is the build of the existing NppShell integrated into the build setup?
What files should I change to get the new project included in the build output?

donho:
@GurliGebis 
> @donho how is the build of the existing NppShell integrated into the build setup? What files should I change to get the new project included in the build output?

NppShell.dll is placed (if needed) manually into the directory where there are the other binaries on my build machine - normally it doesn't need to be rebuilt so once it's put in the directory, it stays for several releases.

That's why I was talking about [Notepad++ organization](https://github.com/notepad-plus-plus). It should be an independent project as WinGUp.

GurliGebis:
> @GurliGebis
> 
> > @donho how is the build of the existing NppShell integrated into the build setup? What files should I change to get the new project included in the build output?
> 
> NppShell.dll is placed (if needed) manually into the directory where there are the other binaries on my build machine - normally it doesn't need to be rebuilt so once it's put in the directory, it stays for several releases.
> 
> That's why I was talking about [Notepad++ organization](https://github.com/notepad-plus-plus). It should be an independent project as WinGUp.

That makes sense, though it is project specific, so I think it makes sense to keep it inside this repository.

GurliGebis:
> > I didn't want to add a nuget restore step to the build process
> 
> IIRC it only involves adding `/restore /p:RestorePackagesConfig=true`

It is now included from a Nuget package.

GurliGebis:
Okay, so an end-of-day status:

Most of the library should be implemented, everything completely untested, but it compiles without warnings.
Next step is to create a msix package from the manifests, test-sign it and test it on a clean VM.

I expect nothing to work (after all, it's been 15+ years since I was playing around with C++, and going from C# to C++ is not trivial)

@donho I have a few things:
Would it make sense that I add a powershell script in the folder that builds the msix package? that way you just need to compile and sign the DLL, run the script to create and sign the msix package.
Then those two just needs to be included in the installer.
Once more thing - can you from the installer check if the version of Windows being run is Windows 11? (like build version 22000 or above), and only run it if that is the case?
You should be able to include the 64bit dll with both the 32bit and 64bit versions of NPP, since it exist seperately from the NPP executable and only interacts with it by executing it when the context menu item is selected (and executing a 32bit exe file from a 64bit dll isn't a problem).

donho:
@GurliGebis 
Great news!

> Would it make sense that I add a powershell script in the folder that builds the msix package? that way you just need to compile and sign the DLL, run the script to create and sign the msix package.

Is MSIX really need? I thought we need only dll + some reg hack.

> Once more thing - can you from the installer check if the version of Windows being run is Windows 11? (like build version 22000 or above), and only run it if that is the case?

Yes I confirm. The NSIS installer can detect all (or almost) the major windows version. So no problem at this point.

> You should be able to include the 64bit dll with both the 32bit and 64bit versions of NPP, since it exist seperately from the NPP executable and only interacts with it by executing it when the context menu item is selected (and executing a 32bit exe file from a 64bit dll isn't a problem).

Yes, that makes sense.
 
BTW, why do this project needs NuGet?

wallrik:
> > Once more thing - can you from the installer check if the version of Windows being run is Windows 11? (like build version 22000 or above), and only run it if that is the case?
> 
> Yes I confirm. The NSIS installer can detect all (or almost) the major windows version. So no problem at this point.

The current version of `WinVer.nsh` got updated by the now infamous commit to be able to detect Windows 11, but maybe consider updating to the latest version in the [NSIS svn source](https://sourceforge.net/p/nsis/code/HEAD/tree/NSIS/trunk/Include/WinVer.nsh)?

That way you can use something like `${If} ${AtLeastWin11}` in the [install script](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/1be0044663df209e7044d413532672ccbba7d1a3/PowerEditor/installer/nppSetup.nsi#L264-L284).

GurliGebis:
@donho the msix is not for installing in this case.
It is sparse (that is the sparse package) and only contains a small amount of metadata needed to register the package with the UI.
So think of it as any other file distributed by the installer - in this case it just shares the file extension with another installer 🙂

GurliGebis:
@donho regarding nuget - the idea is to get the Windows Implementation Library that way, to avoid bundling it with this PR.
I can include it if you prefer, but in general, downloading it as needed at build time is the usual way of doing it (which is why I'm doing it)

GurliGebis:
In other words, the msix sparse package only has the needed metadata to register the COM dll with the shell 🙂

dongle-the-gadget:
> the msix is not for installing in this case.

The sparse MSIX needs to be installed, but we just aren't bundling the whole app in MSIX.

GurliGebis:
> > the msix is not for installing in this case.
> 
> The sparse MSIX needs to be installed, but we just aren't bundling the whole app in MSIX.

Sure, but it is being done using the PackageManager API, so from the Nsis point of view, it just has to call regsvr32 on this dll

donho:
Thank you guys for your explanation & suggestion!

@GurliGebis 
Please mark this PR ready once it's all good - I'll make a new release for it ASAP.

GurliGebis:
Okay, so today's progress is that I'm able to generate the sparse package.
Currently the metadata is pointing to a self-signed certificate I have created, so those will be wrong until I have tested everything (Please note that several other places, I have used NotepadPlusPlus instead of Notepad++ in the manifests - this is due it being limited to alphanumeric characters.)

So, maybe I will get a little time tonight to test it in a clean VM (installing as I'm typing this).

The way I'm seeing it, it can either work in the first try, or it can explode horribly. (I suspect the last one, but you never know).

Anyway, we will have to see - also, the next few days will be quiet from my end, so please don't think I have gone missing, just because nothing is happening 🙂

GurliGebis:
End of day status:

I can register and unregister the package with a single rundll32 call, so that is great.
It looks like there is a bit of debugging needed, but attaching a debugger to the shell in a VM is always interesting...

Until later, if anyone has any ideas on how to have C++ code wait for a debugger to attach (no, I don't like a busy-loop), like you can do in .NET/C# - do let me know.

donho:
> Until later, if anyone has any ideas on how to have C++ code wait for a debugger to attach (no, I don't like a busy-loop), like you can do in .NET/C# - do let me know.

In DllMain `case DLL_PROCESS_ATTACH:`, add `::MessageBox(hModule, "", "Hold", MB_OK);` to block on the entry, once you see the "Hold" display, go to MSVC, and menu "Debug->Attach Process" - it should do the trick.

sylveon:
Waiting for a debugger to attach is also busy looping in C# anyways.

GurliGebis:
It's beginning to look promising:

screenshot

@donho Thanks 🙂
@sylveon Indeed it is, what I meant was that I didn't want to write a busy-loop by hand.

So far only a few things are missing, one of them being figuring out how to localize the "Edit with Notepad++" text string.
I'm looking into how other projects are doing it...

GurliGebis:
@donho when you ordered the certificate, how did you encode the ++, to be able to submit the request?

wallrik:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/60302a747a14097ac4d618a660145a7291cda261/PowerEditor/src/notepad%2B%2B.exe.manifest#L10-L13

Can I suggest that you make the publisher just `CN=NotepadPlusPlus` without an organization, location, province and country? This is how [Microsoft's example is done](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#msix).
```xml
<msix xmlns="urn:schemas-microsoft-com:msix.v1"
  publisher="CN=Contoso"
  packageName="ContosoPhotoStore"
  applicationId="ContosoPhotoStore"
/>
```
Unless the author wants it in there ofc 🇫🇷😊

GurliGebis:
> https://github.com/notepad-plus-plus/notepad-plus-plus/blob/60302a747a14097ac4d618a660145a7291cda261/PowerEditor/src/notepad%2B%2B.exe.manifest#L10-L13
> 
> Can I suggest that you make the publisher just `CN=NotepadPlusPlus` without an organization, location, province and country? This is how [Microsoft's example is done](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#msix).
> 
> ```
> <msix xmlns="urn:schemas-microsoft-com:msix.v1"
>   publisher="CN=Contoso"
>   packageName="ContosoPhotoStore"
>   applicationId="ContosoPhotoStore"
> />
> ```
> 
> Unless the author wants it in there ofc 🇫🇷😊

It has to match the subject of the certificate, and that has all that info.
I'll be pushing my changes later, where it is changed to Notepad++

GurliGebis:
@donho I'm pretty close to being done, hopefully I can push the final code tomorrow and let you know (I'll include a Readme.md file with instructions on how to build this).

I have a few questions/remarks:
1: It seems like you need to reload explorer.exe (or restart the machine) before the menu shows up - I haven't been able to figure out why, but it seems like other people are talking about the same issue - not sure if that should be included in the installer?
2: The context menu item is currently hardcoded to "Edit with Notepad++" - I don't know how you would like to localize it (if at all), so for now I'm just leaving it hardcoded.

donho:
@GurliGebis 
> hopefully I can push the final code tomorrow and let you know

That's a very good news!

> 1: It seems like you need to reload explorer.exe (or restart the machine) before the menu shows up - I haven't been able to figure out why, but it seems like other people are talking about the same issue

It sounds not good for the user experience - the old context menu has the same issue but only for the unregister time.

> - not sure if that should be included in the installer?

Don't follow you: what should not be included in the installer? The dll that you're implementing?

> 2: The context menu item is currently hardcoded to "Edit with Notepad++" - I don't know how you would like to localize it (if at all), so for now I'm just leaving it hardcoded.

By using dll way (but not registry hack) make it harder (or even impossible), but the translation is not MUST TO HAVE, but BETTER TO HAVE. So it's not a priority for the moment. We will see how to improve it when the new context menu dll is deployed widely.

sylveon:
GetTitle can just use whatever way is preferred to localize the string. For example, RC files and LoadString is an option.

GurliGebis:
> @GurliGebis
> 
> > hopefully I can push the final code tomorrow and let you know
> 
> That's a very good news!
> 
> > 1: It seems like you need to reload explorer.exe (or restart the machine) before the menu shows up - I haven't been able to figure out why, but it seems like other people are talking about the same issue
> 
> It sounds not good for the user experience - the old context menu has the same issue but only for the unregister time.

I know, but lets see what I come up with tomorrow.

> > * not sure if that should be included in the installer?
> 
> Don't follow you: what should not be included in the installer? The dll that you're implementing?

I was talking about asking the user if they want to restart explorer, so the dll is loaded.

> > 2: The context menu item is currently hardcoded to "Edit with Notepad++" - I don't know how you would like to localize it (if at all), so for now I'm just leaving it hardcoded.
> 
> By using dll way (but not registry hack) make it harder (or even impossible), but the translation is not MUST TO HAVE, but BETTER TO HAVE. So it's not a priority for the moment. We will see how to improve it when the new context menu dll is deployed widely.

I agree, that is a problem that can be solved later on.

GurliGebis:
Changing back to WRL didn't help with it not requiring restarting explorer 😥

dongle-the-gadget:
This is an OS limitation IIRC.

GurliGebis:
> This is an OS limitation IIRC.

That's my conclusion as well 🙂

GurliGebis:
@donho I'm done, it is ready for your review.
The Readme.md file should explain what is needed to include this - but to make a quick summary:

- Build the dll
- Create the msix sparse package
- Sign both of them
- Have both of them deployed into the program files folder
- Run the rundll32 command to run the RegisterSparsePackage function inside the DLL
- Restart explorer to test, due to the OS limitation.

donho:
@GurliGebis 
Amazing! I will try it this evening (in few hours) with my build machine!

PowerEditor/src/tools/NppModernShell/README.md
```diff
@@ -0,0 +1,61 @@
+# Windows 11 Modern UI Context Menu integration
+
+The purpose of this project is to allow Notepad++ to integrate into the new Windows 11 right-click context menu.
+Doing this requires two new things.
+
+* A dll library with some COM objects that the shell can communicate with.
+* A Sparse Package containing the metadata for the COM server.
+
+To build this, the following steps needs to be taken:
+
```
donho, line 10: I would add this line:
0. Install Windows 11 SDK from: https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/

:)

PowerEditor/src/tools/NppModernShell/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Build a Release dll file (NppModernShell.dll)
+2. Generate a Sparse Package (NppModernShell.msix)
+3. Sign both of these with signtool.exe
+4. Make sure they are included in the installer, so they are deployed next to the notepad++.exe program.
+5. The installer should, upon installation, install the package.
+6. The installer should, upon uninstallation, uninstall the package.
+
+## Build a Release dll file (NppModernShell.dll)
+Just open the NppModernShell.sln Visual Studio solution, select Release as the build type, and do a Rebuild of the solution.
```
donho, line 19: I would add this line:
(Require: [Windows 11 SDK](https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/)

ozone10:
Maybe ARM64 build should be added.

PowerEditor/src/tools/NppModernShell/README.md
```diff
@@ -0,0 +1,61 @@
...
+## Sign both of these with signtool.exe
+Now we have both the `NppModernShell.dll` and `NppModernShell.msix` files, we need to sign them with a valid certificate.
+To do this, once again run the `Developer Command Prompt for VS 2022` command prompt and change to the NppModernShell folder.
+The following command expects the following:
+* The pfx certificate is called MyCert.pfx
+* The password for the pfx certificate is: `Test1234`
+
+Make the needed changes to match the real certificate.
+```
+SignTool.exe sign /fd SHA256 /tr http://timestamp.digicert.com /td sha256 /a /f .\MyCert.pfx /p Test1234 /d "Notepad++" /du https://notepad-plus-plus.org/ Packaging\NppModernShell.msix
```
donho, line 38: Rather:
`SignTool.exe sign /fd SHA256 /tr http://timestamp.digicert.com /td sha256 /a /f .\MyCert.pfx /p Test1234 /d "Notepad++" /du https://notepad-plus-plus.org/ NppModernShell.msix`
(without *Packaging*)

donho:
@GurliGebis 
I've done all the steps indicated in the ReadMe.md (comiling, signing with official Notepad++ certificate, calling run32.exe and restart my PC), I can see "Edit with Notepad++" appears on the context menu:

image

However, clicking on it does just nothing (I supposed it's hard coded in "%ProgramFiles%\Notpad++\notepad++.exe").
Any idea for anything I've missed? How can I check the path?

PowerEditor/src/tools/NppModernShell/README.md
GurliGebis: Done, will be pushing shortly

GurliGebis:
> Maybe ARM64 build should be added.

Yep, but I don't have a way to test it, so that will have to wait - but it should be enough to just add a build target to the c++ project file.

PowerEditor/src/tools/NppModernShell/README.md
GurliGebis: Indeed, I have corrected and will push shortly

GurliGebis:
> @GurliGebis I've done all the steps indicated in the ReadMe.md (comiling, signing with official Notepad++ certificate, calling run32.exe and restart my PC), I can see "Edit with Notepad++" appears on the context menu:
> 
> However, clicking on it does just nothing (I supposed it's hard coded in "%ProgramFiles%\Notpad++\notepad++.exe"). Any idea for anything I've missed? How can I check the path?

It looks in the same folder as the DLL for the exe file - from the looks of it (since it reads the icon the same way), the DLL and msix files are not in the directory where notepad++ is installed.

```
const HMODULE GetThisModule()
{
    HMODULE hm = NULL;

    BOOL result = GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, (LPCWSTR)&GetInstallationPath, &hm);

    if (result == FALSE)
    {
        throw "Failed to locate current module, unable to proceed";
    }

    return hm;
}

const wstring NppModernShell::Helpers::GetInstallationPath()
{
    HMODULE thisModule = GetThisModule();

    wchar_t path[FILENAME_MAX] = { 0 };
    GetModuleFileName(thisModule, path, FILENAME_MAX);
    return std::filesystem::path(path).parent_path().wstring();
}
```

This is how it gets the directory for the icon and exe file to execute - it does it by reading the path to the dll file, and appending "\Notepad++.exe" to it.

GurliGebis:
To be exact, this is the relevant code from the context menu handler:

```
const wstring EditWithNppExplorerCommandHandler::GetNppExecutableFullPath()
{
    const wstring path = GetInstallationPath();
    const wstring fileName = L"\\notepad++.exe";

    return L"\"" + path + fileName + L"\"";
}

const wstring EditWithNppExplorerCommandHandler::Icon()
{
    const wstring fileName = GetNppExecutableFullPath();

    return fileName;
}

const wstring EditWithNppExplorerCommandHandler::GetCommandLine()
{
    const wstring fileName = GetNppExecutableFullPath();
    const wstring parameters = L"\"%1\"";

    return fileName + L" " + parameters;
}
```

Try unregistering the sparse package, move them into the folder where it is, and re-register it.

GurliGebis:
@donho I have sent you an email to the email address in your github profile, with a link to a youtube video showing how I did set it up in my VM, so you can try and replicate that.

GurliGebis:
> Maybe ARM64 build should be added.

I have added the ARM64 build target, so it can be compiled for ARM64 - It is untested though, since I don't have any ARM64 hardware to test on.

donho:
@GurliGebis 

Thank you for the video.
I was stupid to not put the signed msix and dll with notepad++.exe

However, after putting them in the same directory of notepad++.exe and recall rundll32.exe again, there's no more **"Edit with Notepad++"** entry on context menu.

After checking the both digital signatures on their propriety, I have got the following screenshots:

image

Not sure it's because of signature, but indeed, the ways to sign are not the same:
`SignTool.exe sign /fd SHA256 /tr http://timestamp.digicert.com /td sha256 /a /f %NPP_CERT% /p %NPP_CERT_PWD% /d "Notepad++" /du https://notepad-plus-plus.org/ x64\Release\NppModernShell.dll`

`signtool.exe sign /f %NPP_CERT% /p %NPP_CERT_PWD% /d "Notepad++" /du https://notepad-plus-plus.org/ /t http://timestamp.digicert.com/ notepad++.exe`

I was trying to sign both binaries with alternative way (switch both ways on both binaries) - of course with the same certificate,
the both results are error:`
```
C:\Program Files\Notepad++>SignTool.exe sign /fd SHA256 /tr http://timestamp.digicert.com /td sha256 /a /f %NPP_CERT% /p %NPP_CERT_PWD% /d "Notepad++" /du https://notepad-plus-plus.org/ notepad++.exe
Done Adding Additional Store
SignTool Error: Access is denied.
SignTool Error: An error occurred while attempting to sign: notepad++.exe

Number of errors: 1
```

```
C:\Program Files\Notepad++>signtool.exe sign /f %NPP_CERT% /p %NPP_CERT_PWD% /d "Notepad++" /du https://notepad-plus-plus.org/ /t http://timestamp.digicert.com/ NppModernShell.dll
SignTool Error: No file digest algorithm specified. Please specify the digest algorithm with the /fd flag. Using /fd SHA256 is recommended and more secure than SHA1. Calling signtool with /fd sha1 is equivalent to the previous behavior. In order to select the hash algorithm used in the signing certificate's signature, use the /fd certHash option.

C:\Program Files\Notepad++>signtool.exe sign /fd SHA256 %NPP_CERT% /p %NPP_CERT_PWD% /d "Notepad++" /du https://notepad-plus-plus.org/ /t http://timestamp.digicert.com/ NppModernShell.dll
SignTool Error: No certificates were found that met all the given criteria.
```
If it's due to signature, what's the solution?
If it's not due to the signature, what it will be?

GurliGebis:
@donho try unregistering it (use the get-appxpackage to verify it has been unregistered - it will return nothing once it is unregistered) and then registrering it again.

Also, the access denied can be due to the file being in use or the command prompt not being run as admin.

I think the easiest way is to just run the normal steps you do, which should producer a new signed exe file for you to test with 🙂

GurliGebis:
Also I don't think which signtool parameters are used is that important, just remember to timestamp them as well.

My internet access will be limited for the next few days, but I'll try and help when possible

donho:
@GurliGebis 
I've just done again the whole process of the video on my dev laptop - it does work and there's nothing to do with the different digest algorithm of signature.
Also for the build machine in which it didn't work last night - I recalled rundll32 and restarted the explorer - it works like a charm now. So I guess I was tired last night and mess up with some steps.
I will prepare for the next release.
Thank you very much for your assistance!

PowerEditor/src/notepad++.exe.manifest
```diff
@@ -7,6 +7,11 @@
     name="Notepad++"
     type="win32"
 />
+<msix xmlns="urn:schemas-microsoft-com:msix.v1"
+    publisher="CN=Notepad++, O=Notepad++, L=Saint Cloud, S=Ile-de-France, C=FR"
+    packageName="NotepadPlusPlus"
+    applicationId="NotepadPlusPlus"
+/>
```
donho, line 14: @GurliGebis 
I think the above lines are not necessary, since I can use the context menu with v8.5 (which has been compiled without these lines). If they are really needed, please remove organization, location, province and country, as @wallrik has suggested here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13330#issuecomment-1466748095

donho (changes requested):
Please consider to remove added block in notepad++.exe.manifest

PowerEditor/src/tools/NppModernShell/Packaging/AppxManifest.xml
```diff
@@ -0,0 +1,50 @@
...
+  xmlns:uap2="http://schemas.microsoft.com/appx/manifest/uap/windows10/2"
+  xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+  xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10"
+  xmlns:desktop4="http://schemas.microsoft.com/appx/manifest/desktop/windows10/4"
+  xmlns:desktop5="http://schemas.microsoft.com/appx/manifest/desktop/windows10/5"
+  xmlns:uap10="http://schemas.microsoft.com/appx/manifest/uap/windows10/10"
+  xmlns:com="http://schemas.microsoft.com/appx/manifest/com/windows10"
+  IgnorableNamespaces="uap uap2 uap3 rescap desktop desktop4 desktop5 uap10 com">
+  <Identity Name="NotepadPlusPlus" ProcessorArchitecture="neutral" Publisher="CN=&quot;Notepad++&quot;, O=&quot;Notepad++&quot;, L=Saint Cloud, S=Ile-de-France, C=FR" Version="1.0.0.0" />
```
donho, line 14: Are Location, Provence and Country needed here? 
With the renew of certificate every 3 years, these data could be changed, especially the locvation.

PowerEditor/src/notepad++.exe.manifest
GurliGebis: I will do that once I get close to my laptop in a few days

PowerEditor/src/tools/NppModernShell/Packaging/AppxManifest.xml
GurliGebis: I think it is, you can test it easily, just remove it, recreate the msix package and then try to sign it.
If signtool will sign it without errors, you should be good to go. (though you might need to reinstall the package to test that it works)

donho: @GurliGebis 
Sure, I will test it. I will keep you posted if it works.

GurliGebis: 👍

donho: @GurliGebis 
OK, then let's keep only publisher/CN then.

GurliGebis: @donho okay, just to verify, it allowed you to sign it with just the CN part of it?

donho: @GurliGebis 
I have not checked it yet - I am working on packaging NSIS script, and signing problem (Windows 11 SDK has brought the new SignTool.exe which has the different behaviour from the previous one).
I will check later after v8.5.1 being released, so let's keep it as it is.

donho (approved):

PR #13332, opened 2023-03-10
Copy "Find what" to "Replace with" and vice versa

### LINKED ISSUE #12122: Allow filling in selected word in search and replace dialog via option

donho:
This feature use the same swap button for new added actions: Use mouse right click on swap button to toggle among "Swap Replace and Find", "Copy from Find to Replace" and "Copy from Replace to Find".

Fix #12122, close #12195


alankilborn:
Problems with this:

* User will typically right-click right in the middle of the "swap" button, visually obscuring the symbol that gets put there when toggling, so user has to move mouse away to really see the new current setting (ok, maybe this isn't a huge problem, but...)

* User just wants to perform the desired action right NOW on current data in the boxes, but first has to find and set what the button will do in the future, THEN has to click on the button to make the action happen.  Wouldn't it be better if the right-click pops up a menu with choices ("Swap Find and Replace", "Copy Find to Replace", "Copy Replace to Find") and then when the user chooses from the menu, actually execute the chosen action (and then set the action symbol onto the button as the new default)?

pryrt:
I think a menu to do the action, rather than a menu to decide what the next action will be, would make more sense to the user.  Or it would to me, anyway.  And I can just see the complaints in the forum "I chose Copy From Find to Replace and nothing happened."

alankilborn:
@pryrt said:

>I think a menu to do the action

Isn't that what my suggestion was??

>...rather than a menu to decide what the next action will be

Actually, my suggestion was to do both: execute the action immediately, and set the chose action as the button's new action default.

pryrt:
> suggestion was to do both

That makes sense.

donho:
@alankilborn @pryrt 
Please check it's what you want.
I will add the translation after your confirmation.

alankilborn:
It's great, now!  Thank you for this most-excellent feature.

One small thing:  Since for some users it could be a "set once and leave it that way always" type of feature, can it be remembered over Notepad++ restarts?  Currently after a restart, it goes back to "swap" functionality.

donho:
@alankilborn 
Thank you for the review.

> One small thing: Since for some users it could be a "set once and leave it that way always" type of feature, can it be remembered over Notepad++ restarts? Currently after a restart, it goes back to "swap" functionality.

It's intentional. I prefer to keep Swap as default action. If user really stuck on the copy from Find to Replace action, he/she can set when it's needed, then it will last in the whole session.

If there are a lot of demand about it, then I will consider it.

alankilborn:
OK, thank you for the work.

For me, I only want "copy find to replace" functionality.
So, then, I will keep my workaround-script in place.

Yaron10:
@donho,

👍 
Thank you for you work.

It's certainly a nice addition, but it doesn't seem to fully solve the request in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12122.


How about adding a command (could be `IDC_` not in the menu) with a unique shortcut performing `Replace...` *and filling the `Replace with` field*?

alankilborn:
>It's certainly a nice addition, but it doesn't seem to fully solve the request in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12122.

👍

-------------------------------

>How about adding a command with a unique shortcut...

What does THIS mean; a shortcut that the user couldn't remap to a different keycombo?  👎

Yaron10:
@alankilborn,

> a shortcut that the user couldn't remap to a different keycombo?

Why shouldn't the user be able to remap to a different keycombo? :)

alankilborn:
NVM, for some reason I thought maybe you meant a hidden thing with a hardcoded shortcut. :-)

donho:
> How about adding a command (could be `IDC_` not in the menu) with a unique shortcut performing `Replace...` _and filling the `Replace with` field_?

Do we have already such command ?

Yaron10:
@donho,

I'm not sure what you mean.
There are some `IDC_` in NppCommands.cpp.

***

Implementing it this way for `Find` as well should probably be much more convenient for the user than the current settings in the Preferences dialog.
You might often want to fill the `Find with` field. But in many cases you'd rather NOT fill it.

But then it's not only `Find`, but `Find in Files` and the others.
So, at this stage, I only requested your opinion on `Replace`.

alankilborn:
IMO adding more shortcuts just means more I have to remember.  So I don't like that approach.
And, I have my scripted approach that always copies find text to replace when I invoke replace, so I have nothing more to add to this discussion, anyway.

Yaron10:
@alankilborn,

You could remap `Replace (Alt)` (_just an example_) to Ctrl+H.

The downside is that if it's not in the menu, you wouldn't be able execute it with the mouse.
And perhaps - if it's only `Replace`, adding it to the menu might be worth while.

donho:
@Yaron10 
> I'm not sure what you mean.

Do we have already some commands which are not in the menu, with unique shortcut performing some actions?

Yaron10:
@donho,

Switch to Prev/Next document.
Toggle macro recording.

donho:
I will think of such implementation if there are more such request.

Yaron10:
@donho,

Great. Thank you.
If @alankilborn doesn't like it, what's the point? :)

alankilborn:
>If @alankilborn doesn't like it, what's the point?

??????
Never heard of THIS argument before!
I've never seen any ONE user's opinion sway Don on a continual basis.

PR #13336, opened 2023-03-11
Add ability to do leading spaces with ColumnEditor

### LINKED ISSUE #11148: "Leading blanks" option to Column / Multi-Selection Editor
### Description of the Issue / enhancement
Please add "Leading blanks" option to Column / Multi-Selection Editor.

### Expected Behavior
Leading blanks would allow preserving column width. Leading zeros is not an option for languages that give leading zero special import, eg octal format.

### LINKED ISSUE #11309: Feature Request: Please give us a pin feature in search(ctrl+f) to pin 5-10 of our most common searches
Hi

This is a feature request. If possible, would you please give us a pin feature in search(ctrl+f) to pin 5-10 of our most common or favorite searches. I search for ip addresses or email patterns or many other things frequently, but new searches wipe out my older search strings and would be nice to pin a few frequent search strings.

I'm sorry if this feature has been requested before, I searched and couldn't find any similar request.

Thanks

Mac

### LINKED ISSUE #13309: Column Editor > Insert Number drops binary zero
### Description of the Issue
See StR.

1. Start with this setup:
image
2. Press Alt+c to invoke _Column Editor_ and set it up as follows:
image
3. Press OK

### Expected Behavior
A `0` on line 1.

### Actual Behavior
image

### Debug Information
Notepad++ v8.5   (64-bit)
Build time : Mar  6 2023 - 02:18:13
Path : W:\npp.misc\_NoBackup\Releases\npp.8.5\npp.8.5.portable.x64\notepad++.exe
Command Line : -titleAdd="8.5 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2486
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    PythonScript (3.0.14)

alankilborn:
Fix #11148 ; fix #13309 .


alankilborn:
Second commit changes `nbChiffre` to a more readable `nbDigits`; if this is not wanted, easy to separate out.

alankilborn:
Tested the PR code:

```txt
none               |  zeros              |  spaces
0   0   0   0      |  00  00  00  00000  |   0   0   0      0
1   1   1   1      |  01  01  01  00001  |   1   1   1      1
2   2   2   10     |  02  02  02  00010  |   2   2   2     10
3   3   3   11     |  03  03  03  00011  |   3   3   3     11
4   4   4   100    |  04  04  04  00100  |   4   4   4    100
5   5   5   101    |  05  05  05  00101  |   5   5   5    101
6   6   6   110    |  06  06  06  00110  |   6   6   6    110
7   7   7   111    |  07  07  07  00111  |   7   7   7    111
8   8   10  1000   |  08  08  10  01000  |   8   8  10   1000
9   9   11  1001   |  09  09  11  01001  |   9   9  11   1001
10  A   12  1010   |  10  0A  12  01010  |  10   A  12   1010
11  B   13  1011   |  11  0B  13  01011  |  11   B  13   1011
12  C   14  1100   |  12  0C  14  01100  |  12   C  14   1100
13  D   15  1101   |  13  0D  15  01101  |  13   D  15   1101
14  E   16  1110   |  14  0E  16  01110  |  14   E  16   1110
15  F   17  1111   |  15  0F  17  01111  |  15   F  17   1111
16  10  20  10000  |  16  10  20  10000  |  16  10  20  10000
```

Yaron10:
@alankilborn,

Very nice. 👍 
Thank you for your work.

***

Just another interesting case (NOT caused by this PR):
If you select a virtual column and insert text via the column-editor, the post-operation-selection is wrong.

alankilborn:
>the post-operation-selection is wrong.

That was noticed, but since it wasn't related to the change, I didn't worry about it.
The code change is already VERY large, with large risk of rejection on that basis.
(I finished this some time ago, but held back, and wasn't going to submit at all, given the amount of change for such small gain.  Then I decided that the risk was worth it, to potentially provide users with the improvement)

PowerEditor/src/ScintillaComponent/ScintillaEditView.h
```diff
@@ -170,7 +171,7 @@ const std::vector<std::vector<const char*>> g_nonPrintingChars =
 int getNbDigits(int aNum, int base);
 //HMODULE loadSciLexerDll();
 
-TCHAR * int2str(TCHAR *str, int strLen, int number, int base, int nbChiffre, bool isZeroLeading);
+TCHAR * int2str(TCHAR *str, int strLen, int number, int base, int nbDigits, bool isZeroLeading, bool isSpaceLeading);
```
donho, line 174: Use the enum instead of 2 boolean for keeping the consistence.

PowerEditor/src/ScintillaComponent/columnEditor_rc.h
```diff
@@ -33,8 +33,8 @@
 	#define	IDC_COL_FORMAT_GRP_STATIC	(IDD_COLUMNEDIT + 12)
 	#define	IDC_COL_NUM_RADIO			(IDD_COLUMNEDIT + 13)
 	#define	IDC_COL_TEXT_EDIT			(IDD_COLUMNEDIT + 14)
-	#define	IDC_COL_LEADZERO_CHECK		(IDD_COLUMNEDIT + 15)
+	#define	IDC_COL_LEADING_STATIC		(IDD_COLUMNEDIT + 15)
```
donho, line 36: Use a new number (id).

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -3235,41 +3229,70 @@ TCHAR * int2str(TCHAR *str, int strLen, int number, int base, int nbChiffre, boo
...
+		else if (!isZeroLeading)
 		{
-			// use sprintf or swprintf instead of wsprintf
-			// to make octal format work
-			swprintf(f, bufSize, TEXT("%%%s"), fStr);
-			swprintf(str, strLen, f, number);
+			// left-align within the field width, i.e. pad on right with space
+			
+			// first, remove leading zeros
+			for (TCHAR* j = str; *j != '\0'; ++j)
```
donho, line 3254: Please add the braces around the whole instruction of condition `if (*j == '1' || *(j + 1) == '\0')` for the readability & reducing the eventual error in the future maintenance.

donho (changes requested):

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
alankilborn: OK, the reason I didn't do it is because I didn't write this, I just modified it slightly, and wanted to change it as little as possible as to not increase my changed-lines count.  If I know that adding braces into old code is OK, I'll always do it (because I agree it should be done).

PowerEditor/src/ScintillaComponent/columnEditor_rc.h
alankilborn: OK.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1269,7 +1269,12 @@ You can define several column markers by using white space to separate the diffe
                 <Item id="2033" name="&amp;Number to Insert"/>
                 <Item id="2030" name="&amp;Initial number :"/>
                 <Item id="2031" name="Increase b&amp;y :"/>
-                <Item id="2035" name="Leading &amp;zeros"/>
+                <Item id="2035" name="&amp;Leading :"/>
```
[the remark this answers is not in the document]
alankilborn: OK.

Yaron10:
@alankilborn,

Should I open a new issue?

alankilborn:
@Yaron10 

>If the last selected section is "Text", we have "number" in config.xml. And vice-versa.
Should I open a new issue?

I'd say a resounding YES!

Yaron10:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13370.

PowerEditor/src/Parameters.h
```diff
@@ -1213,8 +1213,8 @@ struct ColumnEditorParam final
 	int _initialNum = -1;
 	int _increaseNum = -1;
 	int _repeatNum = -1;
-	bool _isLeadingZeros = false;
 	int _formatChoice = 0; // 0:Dec 1:Hex 2:Oct 3:Bin
+	int _leadingChoice = 0; // 0:None 1:Zeros 2:Spaces
```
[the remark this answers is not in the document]
alankilborn: OK, enum causes additional complication, the "worst" part of that was adding a `#include` in a file that previously didn't need it.  We'll see how you like my change; if not liked then I will ask for additional clarification on what would be liked.

PowerEditor/src/ScintillaComponent/ScintillaEditView.h
alankilborn: OK, enum causes additional complication, the "worst" part of that was adding a `#include` in a file that previously didn't need it.  We'll see how you like my change; if not liked then I will ask for additional clarification on what would be liked.

alankilborn:
@donho, if it builds (still building as I type this), then this PR is ready for re-review.

PowerEditor/src/ScintillaComponent/columnEditor.cpp
```diff
@@ -57,7 +57,15 @@ intptr_t CALLBACK ColumnEditorDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 			::SendDlgItemMessage(_hSelf, IDC_COL_LEADING_COMBO, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>(TEXT("None")));
 			::SendDlgItemMessage(_hSelf, IDC_COL_LEADING_COMBO, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>(TEXT("Zeros")));
 			::SendDlgItemMessage(_hSelf, IDC_COL_LEADING_COMBO, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>(TEXT("Spaces")));
-			::SendMessage(::GetDlgItem(_hSelf, IDC_COL_LEADING_COMBO), CB_SETCURSEL, colEditParam._leadingChoice, 0);
+			WPARAM curSel = 0;
```
Yaron10, line 60: May I ask why this block is necessary?
Can't `leadingChoice` be `int`?

alankilborn: Don wanted the "leading choice" to be an enum (see Don's prior comments).  My goal here was to fully uncouple things.  Again, I can't read Don's mind as to how "far" to take such things, so I proceeded on the presumption of this type of decoupling being wanted.  I really have no idea.  If the future gets "sticky" for me on this PR, I will abandon it and close it; of course an outright rejection will do that just as well, also.  :-)

Yaron10: > Don wanted the "leading choice" to be an enum

I meant to ask why the enum couldn't be `int`. :)

Well, let's for Don's review.

alankilborn: Yea, the way I read it was Don specifically wanted `enum`.
But maybe his real intent was to just not have two bool.  I do understand the desire to not have two bool; indeed I only went with 2 bool in the beginning in an effort to change as little as possible.
The high volume of changed lines in this makes me uncomfortable -- not at a C++ level, but at a "this is someone else's project" level.
I see others aren't afraid to change a lot of lines; I think a lot of changes are more acceptable if the feature is really wanted.
I'm truly sorry that I got involved in this one; my intent was to stay with smaller changes.

Yaron10: I'm glad you got involved in this one. - You did a great job. 👍
I'm sorry you didn't enjoy working on it.

donho: @alankilborn 
Yes, I did want an `enum` but not a `int`.
Regarding the quantity of modification, it depends on different context and a lot of factors, so there is no really a standard criteria.
But just keep one thing in mind - if the a function could have 3 lines but it has 20 lines - then the function is usually coded badly.
Also the unnecessary diff (reformatting the code or it could be generated badly by the tool) will be rejected for sure - the goal is not only to make code review easier, but also for the maintenance (it happens that we go back old commits for debugging reason).  
OK, I will do the enum myself after merging this PR, if you are not willing do it

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
donho: If the coding standard (in contribution.md) is respected, then the code review is easier - that increases the chance of PR acceptance (slightly) indeed. But it's not the real criteria for the acceptance of PR. 

To be more clear for the acceptance of PR (not really criteria, because is quite subject):
A PR on a core function (like open/save file) which is complex (such core functions are usually complex), could be rejected more easily than the other PR, except it brings the fix for the critical bug.

PowerEditor/src/ScintillaComponent/columnEditor.cpp
donho: @Yaron10 
> I meant to ask why the enum couldn't be int. :)

Because specific typed variable is better than a generic typed variable to make the function usage more explicit. Don't you think so?

Yaron10: @donho,

My main question was why the `switch (colEditParam._leadingChoice)` block was necessary.

In the first commit here, only `::SendMessage(::GetDlgItem(_hSelf, IDC_COL_LEADING_COMBO), CB_SETCURSEL...)` was called.

And this is relevant to `ColumnEditorDlg::getLeading()` as well.

alankilborn: >My main question was why the switch (colEditParam._leadingChoice) block was necessary.

"Necessary" might be a strong word, here. :-)
We can do things in a "hacky" way, and sure, that has been done in the past.

The code (that I think) you are questioning was changed to fully decouple the integer values of the combobox selection indices and the "values" of the `enum` type.

With this, the code for this line:
`enum leadingChoice : UCHAR { noneLeading, zeroLeading, spaceLeading };`
could be changed to this without any runtime difference:
`enum leadingChoice : UCHAR { noneLeading=27, zeroLeading=45, spaceLeading=106 };`

(Probably the `UCHAR` should be eliminated; this is a carryover of my habit because I work on smallish embedded systems where every byte can matter.)

Don seemed to want an `enum` ; when he requested this change, I provided, in a better and more complete fashion than an earlier commit (earlier commits aren't really relevant when changes are requested and made).

alankilborn: @donho 

>Edit: oups, I misread your comment. Good job, it's what it needs.

Thank you, Don, encouragement to your contributors that they are doing the right thing is very helpful indeed!

Yaron10: @alankilborn,

Thank you for the explanation. 👍 
You're a very good coder, and I was trying to understand and learn your approach.

Yaron10:
@alankilborn,

If the Format is Hex, Initial Number is `999` and Increase is blank - you always get _4 characters_.
If Initial Number is `99`, you get 2 characters.

What do you think?

alankilborn:
> If the Format is Hex, Initial Number is `999` and Increase is blank - you always get _4 characters_. If Initial Number is `99`, you get 2 characters.
> 
> What do you think?

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c59c07b4fae16c83f545e0da8772950d03ac4988/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L178-L179

If the hex number needs an odd number of digits for display, it is increased to the next even number of digits -- no idea why this would be this way; this PR isn't related.

Yaron10:
@alankilborn,

Yes, it's not related to this PR.
But it is related to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11148#issuecomment-1455239779, and I thought that mini-discussion was held here.

Thank you.

alankilborn:
>But it is related to 11148, and I thought that mini-discussion was held here.

I reread that now and I'm so confused.
Is THAT relevant to current issue and PR?
If not, it needs to be a separate issue, to avoid confusion.

Yaron10:
You asked for my opinion in that thread, and I raised that point as a side-note.
Sorry for the confusion.

alankilborn:
>I raised that point as a side-note.

OK...  I don't think the current issue and PR can/should be a catch-all to all wacky behavior with _Column Editor_.  I think current issue and PR should address the basic deficiency (of not currently -- 8.5.1 -- being able to insert numbers with leading spaces) and that other behavior should be address separately.  I suggest waiting until this PR is integrated/rejected and then making issues on further defects.

donho (approved):

Issue #13350, opened 2023-03-14
Notepad++ hard crash since V8.4.9 with a C# plugin


Redshark22:
### Description of the Issue
Notepad++ hard crash with my C# Plugin since V 8.4.9 (onward).

--> I'm aware that this page is only related to official plugins, but I feel that a bug might have been introduced in V8.4.9 of Notepad++.

### Debug Information
I've developped this c# plugin about 8 years ago and it was working fine with every Notepad++ versions all these years.
I'm using it almost daily, and also add few options to the pluggin monthly.

In the last 2 month I tried to understand what I did wrong, but can't figure out what changed in Notepad++ that lead to this hardcrash.
When debugging Notepad++ during the crash, it indicates that CLR (managed memory for C#) was violated. Since the plugin was working fine prior to that I wonder if modification within Notepad++ aren't now releasing memory that wasn't released before. 
And wonder if anyone could indicate which syntax are doing such action in C++, this could help understand which portion of code should be modify to send information to Notepad++.

To me Notepad++ should only be releasing the memory it allocated, and therefore memory blocked sent by external pluggins should be released by such plugin. (which the c# garbage collector does for me).

In hope that someone can help, thanks.

rdipardo:
> I've developped this c# plugin about 8 years ago

Where is the source code? 

If it's private, then sadly there's nothing anybody can do. GitHub is for *collaborating*, not for *asking*. There's [a forum](https://community.notepad-plus-plus.org) you can use to simply ask questions.

> ### Debug Information
> . . . 

Please follow [the template instructions](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/ISSUE_TEMPLATE.md). Information like the "Admin Mode" setting and a list of installed plugins is always useful in diagnosing a crash.

Redshark22:
Hi,
Yep, the source code isn't open. Sorry if I posted this question in the wrong section.

donho:
It's plugin issue.

PR #13362, opened 2023-03-16
Fix Issue 13350 by reverting minimumRequiredVersion to previous value…

Redshark22:
… for 32b environment.

This will fix issue #13350 so anyone that uses C# DLL can continue enjoying Notepad++.

The value set here was the one existing in previous version of Notepad++ prior to the 10/01/2023.


donho:
It's for the compatibility of version of Windows - 5.01 is even older than the old value in project.
You should change in your project for adapt new version of Notepad++ instead of downgrading this value.

PR #13373, opened 2023-03-19
Untested cleanup jobs for context menus

GurliGebis:
This PR makes sure the NppShell dlls are unregistered and removed when upgrading on Windows 11, since the new dll is used instead.
This also unregisters the modern shell context menu when uninstalling.

This is untested, since I cannot create the installer.

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -290,6 +290,20 @@ ${MementoSection} "Context Menu Entry" explorerContextMenu
 			File /oname=$INSTDIR\NppModernShell.dll "..\bin64\NppModernShell.dll"
 		!endif
 		Exec 'rundll32.exe"$INSTDIR\NppModernShell.dll,RegisterSparsePackage"'
+
+		; Make sure old NppShell dll's are unregistered and removed
+		Exec 'regsvr32 /u /s "$INSTDIR\NppShell_01.dll"'
+		Exec 'regsvr32 /u /s "$INSTDIR\NppShell_02.dll"'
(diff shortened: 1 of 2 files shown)


GurliGebis:
@donho This is untested though, since I cannot create the installer.
Can you test it to see if it works.

To test it:

1. Install 8.5 with context menu items.
2. Upgrade to 8.5.1 with these changes in the installer.
3. Make sure the old NppShell dll is unregistered and removed from the program files directory.
4. Uninstall Notepad++ and ensure the sparse package is unregistered, and the dll and msix is also removed from the program files directory.

GurliGebis:
@donho can you create a RC2 installer, so I can test and verify both cases work?

donho:
@GurliGebis 
Not yet - I'm debugging it.

GurliGebis:
@donho okay 👍
Let me know when it is ready for testning, and I'll test all the case.

donho:
Here is the fix (the part of fix that I've forgot to commit, but it was in the RC binary):
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/9c15bdc236d5f62f4f418f26bee189638a380f86
And in the commit 2 files are removed after unregister.

Here is the RC2:
https://community.notepad-plus-plus.org/topic/24262/notepad-v8-5-1-release-candidate/18?_=1679269990110

GurliGebis:
@donho I can confirm that RC2 fixes all the issues mentioned above.
Can you take a look at my latest comment here, regarding going for 1 dll? (I would like to take a look at it today if possible, but need your feedback first) 🙂 #13320

PR #13375, opened 2023-03-19
Fix configxml wrong val for col editor main choice

### LINKED ISSUE #13370: Column Editor: value in config.xml does not match user's preference

***

Notepad 8.5.1

alankilborn:
Fix #13370 

Note: non-user bug, unless user looks at config.xml content.

Note, this fixes the bug, but goes a bit beyond that to improve the code in a related way.  I hope this is liked/appreciated.


Yaron10:
@alankilborn,

Great. 👍 
Thank you for the quick fix.

PowerEditor/src/ScintillaComponent/columnEditor.cpp
```diff
@@ -448,7 +448,7 @@ intptr_t CALLBACK ColumnEditorDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 	return FALSE;
 }
 
-void ColumnEditorDlg::switchTo(bool toText)
+void ColumnEditorDlg::switchToText(bool toText)
```
Yaron10, line 451: How about `switchToSection`?

alankilborn: >How about `switchToSection`?

My goal is, as always, to change as little as possible, while still adding important value, because this lessens the chance of rejection.

It is challenging to read another developer's mind (Don's) and try to guess what is acceptable vs. not acceptable.

If Don likes something else, he'll tell me/us.

Notes:

* Changing the original `switchTo` has nothing whatsoever to do with fixing the bug; I only changed it because it suffers from the same problem as the source of the bug (poor naming).

* In my own code I would have made it `switchToTextNotNumberSection`, because that is the way I do things.

Yaron10: > In my own code I would have made it switchToTextNotNumberSection

I'm not sure I understand it.
`switchTo()` switches either to Text or to Number.

alankilborn: And **_I_** don't understand your followup...
But my main point was that everyone has a different way of doing things.

Yaron10: OK.
Thanks again for this PR.
It fixes the bug and the code is much clearer. 👍

ozone10:
@alankilborn 
when doing PR try to if possible use latest master, it might contains changes to files you're modifying.

chcg:
Current failures from CI build:

```
        C:\projects\notepad-plus-plus\PowerEditor\src\ScintillaComponent\columnEditor.cpp(128,83): error C2039: '_mainChoice': is not a member of 'ColumnEditorParam' [C:\projects\notepad-plus-plus\PowerEditor\visual.net\notepadPlus.vcxproj]
         C:\projects\notepad-plus-plus\PowerEditor\src\ScintillaComponent\columnEditor.cpp(128,7): error C2660: 'EnableWindow': function does not take 1 arguments [C:\projects\notepad-plus-plus\PowerEditor\visual.net\notepadPlus.vcxproj]

```

alankilborn:
@ozone10 / @chcg 

Updating to current master, it appears to be broken.  Apparently a project setting to warn for implicit fallthru on switch/case statements was enabled, without fixing the fallthru condition in many hundreds of places?  And with `/WX` (warnings are errors), master won't build.

For myself, for now, I'm turning off "warnings are errors", which allows the build to complete.

ozone10:
@alankilborn 
update msvc to 17.5

Yaron10:
@alankilborn,

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/73097461581987aa10aafa6deea7994d89e2f19c/PowerEditor/src/ScintillaComponent/columnEditor.cpp#L128

alankilborn:
Unfortunately, I don't have the github savvy to straighten this out.
I suppose I've learned my lesson about starting with an out-of-date master.
I'll close this and resubmit a new PR some time in the future.
Sorry for anyone's time and trouble...

Yaron10:
@alankilborn,

Thanks again for your work. 👍

Issue #13381, opened 2023-03-20
Select and Find Next: Create the Find Dialog Invisibly


Yaron10:
**STR:**
Open NPP.
Do not open the Find dialog.
Select and Find Next.

**Result:**
The Find dialog is displayed for a split second and then hidden.
This causes a "flickering effect".

**Expected:**
Avoid the flickering and create the Find dialog invisibly.

***

Notepad++ 8.5.1.

alankilborn:
#9536 ?

Yaron10:
@alankilborn,

👍 
Does the "Walking Encyclopedia" remember ALL issues? :)
Let's keep this one open as I've just referenced it in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13372#issuecomment-1476139576.

alankilborn:
>Does the "Walking Encyclopedia" remember ALL issues?

No, but I remembered this from our fairly recent discussion of the flicker when doing the recent "volatile" changes.

alankilborn:
I can't currently find it, but I know it came up along with "volatile"...
Regardless, this StR shows it:

Open NPP.
Do not open the Find dialog.
Find (Volatile) Next.

alankilborn:
>You probably mean that if your "volatile" proposal had been implemented, we would have had the flicker there as well.
Currently, Find (Volatile) Next does not create the Find dialog.

Ah yes, you are correct; ignore my earlier babble in my previous posting.

Yaron10:
I suppose you've implemented your proposal for your personal use. :)
This happened to me [here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13159#issuecomment-1435419788). I didn't even remember I changed that.

Yaron10:
Related:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13396.

Issue #9536, opened 2021-02-18
Find dialog flickers when Select and Find Next (Ctrl+F3) is pressed first time


d0vgan:
### Description of the Issue
When Notepad++ is started and Select and Find Next (Ctrl+F3) is pressed for the first time, the Find dialog flickers.

### Steps to Reproduce the Issue
1. Start Notepad++
2. Press Ctrl+F3 - the Find dialog flickers
3. To reproduce it better, open the source file "PowerEditor\src\ScintillaComponent\FindReplaceDlg.cpp", find the method "FindReplaceDlg::create", add `::Sleep(1000);` right after the line `::SetWindowPos(_hSelf, HWND_TOP, rc.left, rc.top, rc.right - rc.left, rc.bottom - rc.top, SWP_SHOWWINDOW);` and rebuild Notepad++.

### Expected Behavior
As Ctrl+F3 performs a "silent search" for a given word, no flickering is expected.

### Actual Behavior
The Find dialogs flickers.

One of possible ways to fix this issue is to extract the 
```
NppParameters& nppParam = NppParameters::getInstance();
NppGUI& nppGUI = const_cast<NppGUI&>(nppParam.getNppGUI());
if (nppGUI._findWindowPos.bottom - nppGUI._findWindowPos.top != 0)  // check height against 0 as a test of valid data from config
{
    RECT rc = getViewablePositionRect(nppGUI._findWindowPos);
    ::SetWindowPos(_hSelf, HWND_TOP, rc.left, rc.top, rc.right - rc.left, rc.bottom - rc.top, SWP_SHOWWINDOW);
    //Sleep(1000);
}
else
{
    goToCenter();
}
```
into a separate method that is called only when `toShow` is `true`.
Such approach, however, adds another problem: the keyboard focus is for some reason "stolen" from Scintilla and does not return back to it. According to my investigations, the focus is "stolen" as the result of calling `_hSelf = ::CreateDialogParam(_hInst, MAKEINTRESOURCE(dialogID), _hParent, dlgProc, reinterpret_cast<LPARAM>(this));` within "StaticDialog::create" (this one is called from the "FindReplaceDlg::create").

mere-human:
This is kind of expected behavior:
> Notepad++ uses a flashing of the Find dialog window and the main Notepad++ window itself (when the Find dialog is not open) to indicate that search text has not been found (or possibly that a Wrap around in the search has occurred).

[...]

PR #13383, opened 2023-03-20
Fix find dlg flicker

ozone10:
when using Select and Find cmds first time and find dlg is not already created

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13381
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9536


Yaron10:
@ozone10,

👍 
Perfect!

Thanks again. I do appreciate it.

PR #13389, opened 2023-03-21
Unified Shell Extension into one project

GurliGebis:
@donho 
This pull request does the following:

- Removes the old NppShell project.
- Moves the NppModernShell project in its place.
- Adds a call to `SHChangeNotify(SHCNE_ASSOCCHANGED, SHCNF_IDLIST, 0, 0);` to ensure the handlers are loaded without restarting explorer.
- Moves the cleanup logic from the NSIS install script into the code, so registering the DLL does all the clean, simplifying the installer script.
- Changes the DLL to support both the classic and modern context menu (when the DLL is registered, it detects which version of Windows it is running, and installs the correct type of handler (Sparse package on Win11, and a normal ExplorerCommandHandler for older Windows versions).

To use this, do the following:
- Build the dll three times (one for each arch) and sign them like always.
- Regenerate the msix and sign it (the ID's in there has changed to match the ID from the old DLL).
- Include those 4 in the installer, the DLL's should be named like they are outputted from Visual Studio).
- Recreate the installer, since the installer has changed as well.

The following DLL files should be included in these installers:
- 32bit installer: NppShell.x86.dll and NppShell.x64.dll
- 64bit installer: NppShell.x64.dll
- Arm64 installer: NppShell.arm64.dll


PowerEditor/src/tools/NppShell/Installer.cpp
```diff
@@ -0,0 +1,326 @@
...
+#endif
+
+struct DOREGSTRUCT {
+    HKEY hive;
+    wstring key;
+    wstring name;
+    wstring value;
+};
+
+bool IsWindows11Installation()
```
sylveon, line 32: The proper way to check for this is calling VerifyVersionInfo: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-verifyversioninfow

PowerEditor/src/tools/NppShell/Installer.cpp
```diff
@@ -0,0 +1,326 @@
...
+    result = RemoveRegistryKeyIfFound(HKEY_LOCAL_MACHINE, buffer);
+    if (result != ERROR_SUCCESS)
+    {
+        return result;
+    }
+
+    return ERROR_SUCCESS;
+}
+
+LRESULT CleanupHack()
```
sylveon, line 154: We *need* to only delete this when we know for sure it's Notepad++. I see a check, but then it isn't used to guard the call to RegDeleteKey.

sylveon:
You can try implementing this only with C++/WinRT instead of WRL, though you would need to implement DllGetClassObject yourself. Here's an example: https://github.com/TranslucentTB/TranslucentTB/blob/7d823bb71fe784a7b1df2f2dfba4f89a0480767b/ExplorerTAP/module.cpp


DllGetActivationFactory isn't needed since we aren't implementing any activatable WinRT object.

PowerEditor/src/tools/NppShell/Installer.cpp
GurliGebis: Indeed you are right - I have fixed it now.
Thanks 🙂

PowerEditor/src/tools/NppShell/Installer.cpp
GurliGebis: I did look at that first, but it seems like that function is deprecated from Windows 10 going forward.

> Note: This function has been deprecated for Windows 10. See [targeting your applications for Windows](https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1) for more information.

That is why I read the build number from the registry, since that is supported.

GurliGebis:
@sylveon I already tried with DllGetClassObject, and from debugging I can see it creating the object and returning it just fine, but it is like the UI just ignores it.

GurliGebis:
But I'll give it a try again, with your repo as a starting point

PowerEditor/src/tools/NppShell/Installer.cpp
sylveon: Reading from the registry isn't supported, no. There is no docs telling you that this registry key is a supported way for apps to get the OS build number.

Also the function is deprecated, sure, but they point you to version helper macros as the non-deprecated replacement, which ironically makes use of VerifyVersionInfo. So it isn't truly deprecated 😂

GurliGebis:
@sylveon looks like your code is doing something my code wasn't, since it is working 🙂
I'll update the PR to only use WIL instead.

PowerEditor/src/tools/NppShell/Installer.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+    return result;
+}
+
+wstring GetCLSIDString()
+{
+    auto uuid = __uuidof(NppShell::CommandHandlers::EditWithNppExplorerCommandHandler);
+
+    LPOLESTR guidString = 0;
+    HRESULT result = StringFromCLSID(uuid, &guidString);
```
sylveon, line 72: The caller is responsible for freeing the memory allocated for the string by calling the [CoTaskMemFree](https://learn.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cotaskmemfree) function.

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
+	!ifdef ARCH64
+		File /oname=$INSTDIR\NppShell.msix "..\binarm64\NppShell.msix"
+		File /oname=$INSTDIR\NppShell.x64.dll "..\binarm64\NppShell.dll"
+	!else ifdef ARCHARM64
+		File /oname=$INSTDIR\NppShell.msix "..\binarm64\NppShell.msix"
+		File /oname=$INSTDIR\NppShell.arm64.dll "..\binarm64\NppShell.dll"
+	!else
+		File /oname=$INSTDIR\NppShell.x86.dll "..\binarm64\NppShell.dll"
+	!endif
+	Exec 'regsvr32 /s "$INSTDIR\NppShell.dll"'
```
donho, line 285: Does it work also under Windows11?

donho:
@GurliGebis 
FYI, I stop the CI build because it takes time for nothing.

GurliGebis:
@donho is there a tag I can add to the PR for AppVeyor to skip the PR?

PowerEditor/installer/nppSetup.nsi
donho: I just tested by compiling the project into the x64 binary, and put both NppShell.x64.dll (renamed to NppShell.dll) & NppShell.msix into `C:\Program Files\Notepad++\`, then run `regsvr32 /s .\NppShell.dll` under the powershell - nothing happens. I then restarted explorer by typing `Get-process explorer | stop-process` - there's always no "Edit with Notepad++" on context menu of Windows 11, nor on the old context menu (Show more options).

Am I missing somethings?

GurliGebis: Can you try without the `/s` and see if it shows a dialog saying that it finished running.
If it doesn't, then the code is crashing somewhere.

GurliGebis: I just tested on a clean machine, installed npp 8.5, copied the dll and msix, installed my test cert, and tested with regsvr32 on it.
So it should be working, but there might be something that I have missed, that is causing it to fail (since I always test on a clean sandbox machine)

donho: @GurliGebis 
Nm, my bad - I've just forgotten to sign the binaries - after signing it works like a charm!
Sorry for the false alert.

donho: So both old (Show more options) and new context menu of Windows 11 have the "Edit with Notepad++" entry - is it intentional (or rather normal behaviour) ?

donho: > It starts by unregistering both the new and old kind, and then registering the sparse package for Win11, and the old context menu for anything below Win11 

Sorry, I didn't read it - I guess the answer is it :)

GurliGebis: Yep, the new one also shows up in the old menu on Windows 11 🙂

GurliGebis: No worries 🙂 I like how this cleans up the install script.

donho: So `NppShell.msix` is only for 64 bits binaries?

PowerEditor/installer/packageAll.bat
```diff
@@ -33,20 +33,15 @@ If ErrorLevel 1 goto End
...
-
-%signBinary% ..\bin64\NppModernShell.msix
-If ErrorLevel 1 goto End
-%signBinary% ..\bin64\NppModernShell.dll
-If ErrorLevel 1 goto End
-%signBinary% ..\binarm64\NppModernShell.msix
+%signBinary% ..\binarm64\NppShell.msix
 If ErrorLevel 1 goto End
-%signArmBinary% ..\binarm64\NppModernShell.dll
+%signArmBinary% ..\binarm64\NppShell.dll
```
donho, line 44: Same reason above, It should be `NppShell.arm64.dll`.

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
-		!endif
-		Exec 'regsvr32 /s "$INSTDIR\NppShell_06.dll"'
-
-	${EndIf}
+	SetOverwrite try
+	SetOutPath "$INSTDIR\"
+
+	!ifdef ARCH64
+		File /oname=$INSTDIR\NppShell.msix "..\binarm64\NppShell.msix"
+		File /oname=$INSTDIR\NppShell.x64.dll "..\binarm64\NppShell.dll"
```
donho, line 278: The syntax of `File` is `File destination source` (IIRC, `/oname` means Output name).
So here is rather `File /oname=$INSTDIR\NppShell.dll "..\binarm64\NppShell.x64.dll"`

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
+	!else ifdef ARCHARM64
+		File /oname=$INSTDIR\NppShell.msix "..\binarm64\NppShell.msix"
+		File /oname=$INSTDIR\NppShell.arm64.dll "..\binarm64\NppShell.dll"
```
donho, line 281: Same reason above, here is rather `File /oname=$INSTDIR\NppShell.dll "..\binarm64\NppShell.arm64.dll"`

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
+	!else
+		File /oname=$INSTDIR\NppShell.x86.dll "..\binarm64\NppShell.dll"
```
donho, line 283: Same reason above, here is rather `File /oname=$INSTDIR\NppShell.dll "..\binarm64\NppShell.x86.dll"`

donho (changes requested):
Please fix the errors in script.

PowerEditor/installer/nppSetup.nsi
GurliGebis: Yeah, I figured that out, it should be fine now

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
-		!endif
-		Exec 'regsvr32 /s "$INSTDIR\NppShell_06.dll"'
-
-	${EndIf}
+	SetOverwrite try
+	SetOutPath "$INSTDIR\"
+
+	!ifdef ARCH64
+		File /oname=$INSTDIR\NppShell.msix "..\bin\NppShell.msix"
+		File /oname=$INSTDIR\NppShell.dll "..\bin\NppShell.x64.dll"
```
donho, line 278: Should be "..\bin64\NppShell.x64.dll"

PowerEditor/installer/packageAll.bat
```diff
@@ -33,20 +33,15 @@ If ErrorLevel 1 goto End
 %signArmBinary% ..\binarm64\notepad++.exe
 If ErrorLevel 1 goto End
 
-%signBinary% ..\bin\NppShell_06.dll
+%signBinary% ..\bin\NppShell.dll
```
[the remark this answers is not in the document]
GurliGebis: Fixed 🙂

PowerEditor/installer/packageAll.bat
```diff
@@ -33,20 +33,15 @@ If ErrorLevel 1 goto End
...
 If ErrorLevel 1 goto End
 
-%signBinary% ..\bin\NppShell_06.dll
+%signBinary% ..\bin\NppShell.dll
 If ErrorLevel 1 goto End
-%signBinary% ..\bin\NppShell64_06.dll
+%signBinary% ..\bin64\NppShell.msix
 If ErrorLevel 1 goto End
-%signArmBinary% ..\binarm64\NppShell64.dll
+%signBinary% ..\bin64\NppShell.dll
```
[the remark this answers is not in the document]
GurliGebis: Fixed 🙂

GurliGebis:
@donho the nppSetup.nsi and packageAll.bat files should be fixed now 😀

donho (changes requested):
I thought it's triggered, sorry for the modif again.

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
-			${EndIf}
-		!endif
-		Exec 'regsvr32 /s "$INSTDIR\NppShell_06.dll"'
-
-	${EndIf}
+	SetOverwrite try
+	SetOutPath "$INSTDIR\"
+
+	!ifdef ARCH64
+		File /oname=$INSTDIR\NppShell.msix "..\bin\NppShell.msix"
```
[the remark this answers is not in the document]
GurliGebis: You are right - it is fixed now

PowerEditor/installer/nppSetup.nsi
GurliGebis: You are right - it is fixed now

donho (approved):

donho:
@GurliGebis 
Please check http://download.notepad-plus-plus.org/repository/8.x/8.5.1.RC3/
I'll keep the RC2 in case RC3 doesn't work well.
Thank you for your coding effort!

PowerEditor/installer/nppSetup.nsi
```diff
@@ -270,64 +270,33 @@ Var nppSubStr
...
-	${If} $WinVer == "11"
-	
-		; Clean up the hack of v8.5 installer
-		ReadRegStr $muiVerbStr HKLM "SOFTWARE\Classes\*\shell\pintohome" MUIVerb
-		${StrStr} $nppSubStr $muiVerbStr "Notepad++"
-		; Make sure there's an entry, and the entry belong to Notepad++ before deleting it
-		${If} $muiVerbStr != ""
-			${AndIf} $nppSubStr != ""  ; it contains "Notepad++"
-				DeleteRegKey HKLM "SOFTWARE\Classes\*\shell\pintohome"
-		${EndIf}
```
donho, line 282: Hold, the above code for clean up v8.5 hijacking is not in installer?
It should add back in the beginning of this block.

donho (changes requested):
Add back the clean up of v8.5's registry hijacking.

PowerEditor/src/tools/NppShell/Installer.cpp
```diff
@@ -0,0 +1,339 @@
...
+    {
+        return deployResult.ExtendedErrorCode();
+    }
+
+    return S_OK;
+}
+
+HRESULT NppShell::Installer::UnregisterSparsePackage()
+{
+    PackageManager packageManager;
```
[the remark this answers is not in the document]
GurliGebis: You got a point, I'll make sure it only tries to unregister it if it is on Win11

PowerEditor/installer/nppSetup.nsi
donho: Ha! Isn't it better to clean up via Installer?

GurliGebis: That is in the installer:

```
LRESULT CleanupHack()
{
    wstring keyName = L"SOFTWARE\\Classes\\*\\shell\\pintohome";
    wstring valueName = L"MUIVerb";

    LRESULT result = 0;
    bool found = false;

    HKEY hkey;
    HRESULT hResult = RegOpenKeyEx(HKEY_LOCAL_MACHINE, keyName.c_str(), 0, KEY_READ, &hkey);

    if (hResult == ERROR_SUCCESS)
    {
        DWORD cbData = 0;
        hResult = RegGetValue(hkey, nullptr, valueName.c_str(), RRF_RT_REG_SZ, nullptr, nullptr, &cbData);

        if (hResult == ERROR_SUCCESS)
        {
            vector<BYTE> buffer(cbData + 1);

            cbData = (DWORD)buffer.size();
            hResult = RegGetValue(hkey, nullptr, valueName.c_str(), RRF_RT_REG_SZ, nullptr, &buffer[0], &cbData);

            wstring valueString(reinterpret_cast<wchar_t*>(&buffer[0]));

            found = valueString.find(L"Notepad++") != wstring::npos;
        }
    }

    RegCloseKey(hkey);

    if (found)
    {
        result = RegDeleteTree(HKEY_LOCAL_MACHINE, keyName.c_str());
    }

    return result;
}
```

It is being called from `NppShell::Installer::UnregisterOldContextMenu()`

PowerEditor/src/tools/NppShell/Installer.cpp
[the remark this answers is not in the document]
sylveon: You might also need to delay load a few libs, as some of the statically linked C APIs for WinRT (`RoGetActivationFactory`) will also be missing. I think cppwinrt has some level of support for building a DLL that can also run on 7, so it might not actually be needed, but better to directly test on a Windows 7 machine/VM.

GurliGebis:
> @GurliGebis Please check http://download.notepad-plus-plus.org/repository/8.x/8.5.1.RC3/ I'll keep the RC2 in case RC3 doesn't work well. Thank you for your coding effort!

I just tested it, both with a clean install (works fine) and an upgrade from 8.5 (works as well, and cleans up the old hack as well)

donho (approved):

GurliGebis:
@donho after 8.5.1 has been released, should we look into moving this library into a separate github repository, and add it as a submodule to this repository?

donho:
@GurliGebis 
> after 8.5.1 has been released, should we look into moving this library into a separate github repository, and add it as a submodule to this repository?

Yes, I think it's better.
I'll create a empty repository so you can push all code into it.

GurliGebis:
> @GurliGebis
> 
> > after 8.5.1 has been released, should we look into moving this library into a separate github repository, and add it as a submodule to this repository?
> 
> Yes, I think it's better. I'll create a empty repository so you can push all code into it.

Great, thanks 😃

donho:
@GurliGebis 
I just rebuilt from your latest commit of this PR, and repackage binaries, on the same URL:
http://download.notepad-plus-plus.org/repository/8.x/8.5.1.RC3/

BTW, on which OS have you tested it?

GurliGebis:
> @GurliGebis I just rebuilt from your latest commit of this PR, and repackage binaries, on the same URL (content updated): http://download.notepad-plus-plus.org/repository/8.x/8.5.1.RC3/
> 
> BTW, on which OS have you tested it?

I have tested it on Windows 11, with a bit of hardcoding to make it think it is Windows 10. (I am using the Windows Sandbox feature on my dev machine to test it, since it is clean every time it starts)

GurliGebis:
@donho would there be time for another PR after this one is merged, before 8.5.1 is released?
There is an issue with uninstall that I want to look at (the file is locked, so I want to export a new dll function that we can call, that uses the MoveFileExW api call to move it away to the temp directory and tag it for deletion at next reboot.
I should be able to do that in a new branch pretty fast tomorrow 🙂

donho:
@GurliGebis 
In this case, let's maintain the RC2 as the official release.
So we can work on it with no rush, and it can be tested during enough of time by the community. 
(I didn't think you wanted it to be in v8.5.1 when you asked to work on this PR.)

GurliGebis:
@donho the same issue is in there.
It is in all of them, since the dll is loaded into the process.
Please use the latest with this PR included, and I'll create a PR with a fix for it tomorrow, since I know what needs to be done, it should me pretty fast to write.
Going for RC2 would be a worse choise, since we then have the dll with the wrong name in the wild, so please don't (it just complicates the cleanup going forward)

GurliGebis:
The solution is simple, the dll should be able to be moved even when it is loaded, so we can move it to the temp directory and then delete it from there on the next reboot.

GurliGebis:
@donho I couldn't sleep, so I coded up the solution and tested it, please see commit https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13389/commits/8f641720fef206d07c3c96c5e5ea4c88798e19d1
So now this should be good to go - it moves the dll out of the way upon uninstallation, so the uninstall process can complete without issues, and then registers it for deletion upon reboot, from the temp folder, where it doesn't block anything.

donho:
@GurliGebis 
OK. Just publish it here: https://community.notepad-plus-plus.org/topic/24262/notepad-v8-5-1-release-candidate/19?_=1679531912831
I will try to found out my windows7 VM. If you can test it on windows7, it will be great.

But you have to stop pushing new commit.
I cannot base on release on a floating code, as people cannot walking on non-freeze water (except one :) ).

GurliGebis:
@donho I am done now, so if the code I just pushed is included, we are good to go.
If so, I test it tomorrow, since I have shut down my machine for tonight 🙂

GurliGebis:
I do think this last commit should be included though, to ensure uninstall is smooth

donho:
@GurliGebis 
Last commit has been included: I just rebuilt all the package for the last time.
I guess now you can sleep peacefully :) Thank you for your coding effort and good night.

GurliGebis:
> @GurliGebis Last commit has been included: I just rebuilt all the package for the last time. I guess now you can sleep peacefully :) Thank you for your coding effort and good night.

Thank you, and you're welcome.
Should I just test the link you sent above, or is there a new one? 🙂

GurliGebis:
@donho I fixed the code, so it uses the std::filesystem api instead, so now it runs on Windows 7 as well.
Sorry for not noticing it before, but I just tried it on Windows 7 with my own built dll with that change, and it works perfectly.

So now finally, we should be good to go, all the way down to Windows 7 / Windows Server 2008 R2 😀🚀

donho:
@GurliGebis 
OK, I will do RC4.

However, RC3 has got a problem:
https://community.notepad-plus-plus.org/topic/24262/notepad-v8-5-1-release-candidate/20?_=1679541909944

Could you check with the user please?

GurliGebis:
@donho I'm not sure which one that is, since it is unable to scroll all the way down.
Anyway, if it is the one I'm thinking about, the issue most likely is that they have one of the previous RC versions installed, which we don't cleanup for.

My suggetion is that the user installs the RC they had before, and then uninstalls it. (Since upgrading doesn't uninstall the old version, the uninstaller parts for nsis is never run.

GurliGebis:
@donho I discovered a bit more cleanup that is needed, have you built RC4 yet?

donho:
@GurliGebis 
Please no more push

GurliGebis:
> @GurliGebis Please no more push

Deal, I'll test shortly and save the last little one for after 8.5.1 🙂

GurliGebis:
@donho 

Windows 7:
Install 8.4.9 and upgrade to 8.5.1 : OK
Install 8.5 and upgrade to 8.5.1: OK

We are good to go 🚀

Once we have 8.5.1 out the door, and you have created another repo for this project, I'll merge a few optimizations into there, for the 8.5.2 release - is that okay with you? 🙂

donho:
Unfortunately there's a bad news from my side.

I just tested RC4 32bits installer on Windows 10. The old entry has been removed after installation, but the new entry is not there. The same result after restarting explorer. 
Installing RC2 afterward restores the "Edit with Notepad++".

Here's the info of tested PC:
```
Notepad++ v8.5.1   (32-bit)
Build time : Mar 21 2023 - 19:13:46
Command Line :
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2486
Current ANSI codepage : 1252
Plugins :
    ComparePlugin (2)
    DSpellCheck (1.4.6)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    XMLTools (3.1.0.1)
```

sylveon:
Does the 32-bit installer only try to install the 32-bits NppShell DLL? If running on a 64-bits copy of Windows, the menu wouldn't show up in that case because Explorer is a 64 bits process and cannot load 32-bits DLLs.

Both 32 bits and 64 bits DLLs should be registered on a 64 bits OS (even if the main app is 32 bits). That way, the shell extension can show in:
- the file explorer, which is native 64 bits
- the browse file dialog of 64 bits apps
- the browse file dialog of 32 bits apps

Windows 11 doesn't have this issue because it hosts extensions which are sparse-packaged out of process (so bitness doesn't matter anymore), but this is not the case for Windows 10 and lower.

GurliGebis:
Installing Windows 10 32bit as we speak - will report back ASAP @donho

donho:
@GurliGebis 
> if you run regsvr32.exe NppShell.dll manually, what error does it give?

I cannot test it now - it's laptop of my gf and she's working...
I'll let you know in about 2 hours

GurliGebis:
@donho I'll test in a few minutes myself, just have to say good night to my daughter first 🙂

donho:
@sylveon 
> Does the 32-bit installer only try to install the 32-bits NppShell DLL? 

Yes, it does.

sylveon:
If you where trying on a 64 bits OS, then that is most likely why.

donho:
@GurliGebis 

FYI: I just created a repository in [Notepad++ organization](https://github.com/notepad-plus-plus/): 
https://github.com/notepad-plus-plus/nppShell

Feel free to do the PR in this repo :)

donho:
@sylveon 
> If you where trying on a 64 bits OS, then that is most likely why.

Well, the old 32bit NppShell.dll works though.
What's the difference between the 32 bits of NppShell.dll and NppModernShell.dll  then?

sylveon:
The old installer probably registered both correctly, I would guess (or only shipped and registered a 64 bits dll)

GurliGebis:
@donho I just tested on 32bit Windows 10, and it seems to be working.

I installed 8.5, and then upgraded to 8.5.1, which worked fine.
The first time I tried, it seems like it somehow forgot to create the CLSID registry key

donho:
@sylveon 
> The old installer probably registered both correctly, I would guess (or only shipped and registered a 64 bits dll)

After checking NSIS code (before & after this PR), the old and new installer (of this PR) follows the same logic - Notepad++ x64 binary comes with NppShell64.dll; Notepad++ x32 binary comes with NppShell32.dll

So for me the old NppShell 32 bits works fine under Windows 10 64 bits.

GurliGebis:
@donho I think I have fixed it (now it seems to work every time on my Windows 10 32bit machine).
I have added logic to flush the registry hive to disk after the cleanup, before it registers it, which seems to have made a difference.
See the latest push

GurliGebis:
Can you test locally with the code I just pushed?

donho:
@GurliGebis 
> I just tested on 32bit Windows 10, and it seems to be working.

The issue comes with the [32 bits installer on Windows 10 x64](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13389#issuecomment-1481600699).

GurliGebis:
@donho ahhh, that makes sense.
Can you confirm it is the 64bit dll that is placed inside the notepad++ folder?

donho:
@GurliGebis 
> I think I have fixed it (now it seems to work every time on my Windows 10 32bit machine).
> I have added logic to flush the registry hive to disk after the cleanup, before it registers it, which seems to have made a difference.
> See the latest push

Sorry but I can't now. I'm doing v8.5.1 release with RC2 (as we have agreed) 
RC2 may not be the best solution for devs (for you, me and other devs) but it provides the correct context menu entry and solves the problem of hijacking - and furthermore there won't be the regression.

Sincerely I prefer the solution provided by this PR, however we need more time for testing it. So let's move the code of this PR to new repo - I'll keep providing the package for testing and I will participate of testing as well.

GurliGebis:
@donho okay, I'll push it to there when 8.5.1 has been released, and then I'll figure out what updates are needed to cleanup for 8.5.1, since RC2 is a bit different than RC4 🙂

GurliGebis:
@donho I figured it out for real this time - the 32bit installer installs the 32bit DLL on 64bit windows.
It needs to install the 64bit dll on 64bit windows - is there a way to detect which arch we are running on inside the nsis script?

donho:
@GurliGebis 

As a dev, I do understand your desire for this feature to be included, but the PR need more time and more people to test clearly, so I won't take the risk.

What I can do at least is, to build the latest commit for you and people for testing, but it won't be in v8.5.1 definitely (in v8.5.2 for sure - if there's no regression found of course).

GurliGebis:
@donho thanks, I can agree with you on that.
I have already created a PR to the new repo with all this code here.
If you can merge it once 8.5.1 has been released, I'll update this PR to include a cleanup for NppModernShell.dll name, and a shift from the code being included to it being included as a submodule - is that okay with you?

GurliGebis:
@donho I have created a PR over there.
Once it is merged, I'll create a new PR that cleans up and adds it as a submodule instead

donho:
@GurliGebis 
Thank you.
I'll take care of it once v8.5.1 is released.

FYI, I have moved RC5 into: http://download.notepad-plus-plus.org/repository/MISC/8.5.1.RC5/

ZanzaZanzara:
OK so with 8.5.1 x64 the Edit with Notepad++ Option in Windows 11 Modern right click menu is gone again. I was happy to see it with 8.5.
I understand that it has been changed to "Pin to Quick access" function to something else in 8.5.1 but now its gone again like it was before 8.5.

GurliGebis:
@donho I can confirm that the RC5 above installs the 64bit dll with the 32bit installer on 64bit Windows, so the issue is fixed 🙂
If you can merge my PR https://github.com/notepad-plus-plus/nppShell/pull/1 , I'll make a new one here that updates the installer to cleanup 8.5.1 NppModernShell naming and uses a submodule instead.

GurliGebis:
@ZanzaZanzara can you try with the RC5 version that donho linked above, to see if that works better?
You might want to uninstall, restart the machine and delete the notepad++ folder from within Program Files before installing, since there is no cleanup added to that version yet.

donho:
@ZanzaZanzara 
Could you post here your debug info please ? (via menu `?->Debug Info...`)

donho:
@sylveon 
> After checking NSIS code (before & after this PR), the old and new installer (of this PR) follows the same logic - Notepad++ x64 binary comes with NppShell64.dll; Notepad++ x32 binary comes with NppShell32.dll
> 
> So for me the old NppShell 32 bits works fine under Windows 10 64 bits.

My bad. The old Installer contains both (32&64bits) dll and it will deploy the one matches with the OS.
You're right about it!

sylveon:
Psychic debugging!

GurliGebis:
@donho I have created a PR with the installer changes here: #13411

I think we should close this thread, since if people have any bugs related to it, it should be created as a new issue 🙂

PR #13394, opened 2023-03-23
Fix configxml wrong val for column editor main choice

### LINKED ISSUE #13370: Column Editor: value in config.xml does not match user's preference

alankilborn:
Fix #13370 

This is second try on a PR for fixing this; first try was the aborted 13375.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -2495,9 +2495,9 @@ void NppParameters::feedColumnEditorParameters(TiXmlNode *node)
 	if (strVal)
 	{
 		if (lstrcmp(strVal, TEXT("text")) == 0)
-			_columnEditParam._mainChoice = false;
+			_columnEditParam._mainChoiceIsText = true;
 		else
-			_columnEditParam._mainChoice = true;
+			_columnEditParam._mainChoiceIsText = false;
 	}
 	TiXmlNode *childNode = columnEditorRoot->FirstChildElement(TEXT("text"));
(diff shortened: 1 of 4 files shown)


donho:
@alankilborn 
This PR is good in principle.
However, I prefer the minimal change which matches the initiative intention of this PR.
Thank you for your coding effort anyway.

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d37c7c396a438e6866dfcce4e5eb99513b537ef4

alankilborn:
In https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d37c7c396a438e6866dfcce4e5eb99513b537ef4 you've created an artificial `enum` by going to extraordinary lengths to give `true` and `false` new names.  Why not just go all the way and make a real enum, if you're going that far?

But an enum is not necessary here, nor is redefining true/false.  All that's needed is a better name for `_mainChoice`.  It's a boolean, give it a name that reads like a boolean.

Example:  `isBanana` makes sense for a boolean; if it is true, it is a banana.  `_mainChoice` as the name of a boolean just makes me puzzled, as it conveys no additional meaning.

I'm certain that the poor naming is what caused the bug described by the issue to happen in the first place.  OK, the bug has been fixed, but why not fix the thing that caused the bug to occur at the same time?  Even better is to have a takeaway for a nice way to do things in the future.

No big deal, but leaving this poor code in place seems like an opportunity lost.

Thanks for your time.

PR #13407, opened 2023-03-25
CopyMarkedText button in Mark window does not record a macro action

### LINKED ISSUE #13405: CopyMarkedText button in Mark window does not record a macro action
### Description of Issue

A press of the _Copy Marked Text_ button during macro recording does not record the action into the macro.

Reference discussion here: https://community.notepad-plus-plus.org/topic/24271

1. Invoke the _Mark_ window with Ctrl+m
2. Start macro recording
3. Press the _Copy Marked Text_ button in the _Mark_ window:
image
4. Stop macro recording
5. Attempt to playback macro with Ctrl+Shift+P

### Expected Behavior
Expected any marked text in the active document to be copied to the clipboard

### Actual Behavior
Can't even playback anything (because nothing was recorded) and thus playback command is not enabled.

### Workaround
Record the _Search_ menu > _Copy Styled Text_ > _Find Mark Style_ action into your macro sequence where you would want to include the _Mark_ window > _Copy Marked Text_ press.

### Debug Information
Notepad++ v8.5.1   (64-bit)
Build time : Mar 21 2023 - 19:16:10
Path : W:\npp.misc\_NoBackup\Releases\npp.8.5.1\npp.8.5.1.portable.x64\notepad++.exe
Command Line : -titleAdd="8.5.1 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2604
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

doug1234:
Fix #13405

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2082,8 +2082,7 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 
 				case IDC_COPY_MARKED_TEXT:
 				{
-					if (isMacroRecording) saveInMacro(wParam, FR_OP_FIND);
-					(*_ppEditView)->markedTextToClipboard(SCE_UNIVERSAL_FOUND_STYLE);
+					::SendMessage(_hParent, WM_COMMAND, IDM_SEARCH_MARKEDTOCLIP, 0);
 				}
 				return TRUE;
 
@@ -3920,12 +3919,6 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 					case IDC_CLEAR_ALL:
 					{
 						clearMarks(*_env);
-						break;
-					}
-
-					case IDC_COPY_MARKED_TEXT:
-					{
-						(*_ppEditView)->markedTextToClipboard(SCE_UNIVERSAL_FOUND_STYLE);
 						break;
 					}
 
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2082,6 +2082,7 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 
 				case IDC_COPY_MARKED_TEXT:
 				{
+					if (isMacroRecording) saveInMacro(wParam, FR_OP_FIND);
 					(*_ppEditView)->markedTextToClipboard(SCE_UNIVERSAL_FOUND_STYLE);
 				}
 				return TRUE;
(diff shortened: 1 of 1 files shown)


alankilborn (commented):
As this is not really a "find" operation, it seems a bit awkward to encode it as one; here's how this PR records use of the command in a macro:

```
            <Action type="3" message="1700" wParam="0" lParam="0" sParam="" />
            <Action type="3" message="1601" wParam="0" lParam="0" sParam="free" />
            <Action type="3" message="1625" wParam="0" lParam="0" sParam="" />
            <Action type="3" message="1702" wParam="0" lParam="771" sParam="" />
            <Action type="3" message="1701" wParam="0" lParam="1725" sParam="" />
```

It seems more reasonable that it be encoded in a macro as a single line entry:

```
<Action type="2" message="0" wParam="43061" lParam="0" sParam="" />
```

alankilborn:
The "title" of your PR should _not_ contain "fix issue xxxx".

Your opening comment _should_ start with "Fix #xxxx"

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -2082,6 +2082,7 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 
 				case IDC_COPY_MARKED_TEXT:
 				{
+					if (isMacroRecording) saveInMacro(wParam, FR_OP_FIND);
 					(*_ppEditView)->markedTextToClipboard(SCE_UNIVERSAL_FOUND_STYLE);
```
alankilborn, line 2086: Maybe in between the braces for IDC_COPY_MARKED_TEXT, do _only_:  `::SendMessage(_hParent, WM_COMMAND, IDM_SEARCH_MARKEDTOCLIP, 0);`

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3922,6 +3923,12 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 						break;
 					}
 
+					case IDC_COPY_MARKED_TEXT:
```
alankilborn, line 3926: If https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13407/commits/8f4a1850a1f9d18c434715c662661eafa21d8384#r1148360378 is done, this change could probably be eliminated.

doug1234:
Thank you for the help on the title @alankilborn.  While I don't disagree about it not being a "find" operation it should be consistent with how IDCMARKALL is recorded which was my goal.  I think if one operation on the mark page was recorded differently then all of them should probably be changed to be consistent.

doug1234:
We are typing at the same time it seems... I will see if `::SendMessage(_hParent, WM_COMMAND, IDM_SEARCH_MARKEDTOCLIP, 0);`  Takes care of the whole problem.  That may be a better solution.

alankilborn:
>I think if one operation on the mark page was recorded differently then all of them should probably be changed to be consistent.

This command isn't like the others at all; it is more of a convenience function being here, as its functionality is duplicated on the _Search_ menu, as noted.

If it is encoded as a "find" operation, then it has misleading stuff bound to it, e.g. options like "match case", "in selection", etc., making someone looking at the data for the encoded macro be confused.

doug1234:
I changed things as you suggested and it seems to work well.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: Smart one!

Issue #13410, opened 2023-03-25
notepad++ 8.5.1 no icon when launching from the context menu (edit file)


AvenCores:
windows 11 last bild
npp.8.5.1.Installer.x64.exe

no icon when launching from the context menu (edit file)
![image](https://user-images.githubusercontent.com/64781822/227704077-75e7910b-0f23-4c84-a24d-39badcad2f3e.png)

but if you just run notepad++ as a normal application, the icon is displayed
![image](https://user-images.githubusercontent.com/64781822/227704114-73ca53c8-2f4e-4739-b2c5-4923f860cd5b.png)

donho:
Indeed, I can reproduce it.

@GurliGebis 
Not sure if it's link to NppModernShell - have you any idea about it?

donho:
BTW, @AvenCores Could you provide the debug info via menu `?->Debug Info...` please ?

AvenCores:
> BTW, @AvenCores Could you provide the debug info via menu `?->Debug Info...` please ?

Notepad++ v8.5.1   (64-bit)
Build time : Mar 21 2023 - 19:16:10
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 Pro (64-bit)
OS Version : 22H2
OS Build : 22621.1413
Current ANSI codepage : 1251
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

GurliGebis:
@donho the code ready the current path of the dll, combine that with notepad++.exe
The problem is that the RC2 it is based on is only half finished - so instead we should get a new test build with my latest PR, which people with issues can be pointed to.
No point in using a lot of energy debugging something that has already been changed alot in the next release.

donho:
@GurliGebis 
So this bug has not been fixed in the binaries we have tested - I have reproduced with both Notepad++ 32 & 64 installer under Windows 11.

GurliGebis:
@donho okay, I'll try and see if I can reproduce it.
It is a bit weird, since all that it does is call CreateProcess to launch it.

donho:
@GurliGebis 
> It is a bit weird, since all that it does is call CreateProcess to launch it.

It works - Notepad++ has been launched and the file is opened in launched Notepad++.
It's just that the icon of application is not shown on the taskbar.


[...]

PR #13411, opened 2023-03-25
NppShell cleanup

GurliGebis:
Now that the NppShell has been moved into it's own separate repository, the following steps are needed:

- Remove the old NppShell  and NppModernShell projects.
- Add the NppShell repo as a submodule instead.
- Changes to the installer to handle the installation correctly for 32bit NPP on 64bit Windows.

This pull request does all of these.


GurliGebis:
Reference: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13389
Reference: https://github.com/notepad-plus-plus/nppShell/pull/1

@donho once this has been merged, I would like if you can create an installer for me?
I need to make sure it works if I upgrade from <8.5.1, where the NppModernShell.dll doesn't exist.

If that is working (it should - but I have said that way too many times lately 😀), we should be good to go.

GurliGebis:
@donho since this moves the NppShell folder into a submodule, you might need to clear it for any untracked files, in case git complains

donho:
@GurliGebis 

Thank you. Here we go:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST01/

Please test as more OS as you can. 
I will do the same now.
If there's no bug/regression found, I will post the URL on the community to ask people test.

GurliGebis:
@donho thanks, I'll give it a test tonight and get back to you 🙂

GurliGebis:
@donho It seems like I have to add a check for if the NppModernShell.dll file exists (see the latest change https://github.com/notepad-plus-plus/notepad-plus-plus/compare/33e8d0ca4a12484b16afad70c133e9b4af66ec53..9ebd7c5ac388763608e6b38eefae578744a3a191)
I'll ignore that error and test the rest, and if you can build another installer from the new commit, I'll test this again.

I'll report back with my other tests in a while.

GurliGebis:
Windows 11
64bit debug:
```
Notepad++ v8.5.1   (64-bit)
Build time : Mar 24 2023 - 19:54:06
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 Enterprise (64-bit)
OS Version : 22H2
OS Build : 22621.1413
Current ANSI codepage : 1252
Plugins : 
	mimeTools (2.9)
	NppConverter (4.5)
	NppExport (0.4)
```
32bit debug:
```
Notepad++ v8.5.1   (32-bit)
Build time : Mar 21 2023 - 19:13:46
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit)
OS Version : 22H2
OS Build : 22621.1413
Current ANSI codepage : 1252
Plugins : 
	mimeTools (2.9)
	NppConverter (4.5)
	NppExport (0.4)
```

Findings:

- New test installer fails to unregister if the file is missing, I have added a check in the last commit.
- Everything else is working, both upgrading from 8.5 and 8.5.1 works, the menu is visible instantly, both for x86 and x64

Windows 10
64bit debug:
```
Notepad++ v8.5.1   (64-bit)
Build time : Mar 24 2023 - 19:54:06
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 1803
OS Build : 17134.165
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```
32bit debug:
```
Notepad++ v8.5.1   (32-bit)
Build time : Mar 21 2023 - 19:13:46
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 1803
OS Build : 17134.165
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

Findings:

- Fails if earlier version has been installed, due to HKEY_CLASSES_ROOT\*\shellex\ContextMenuHandlers\ANotepad++64 being present - PR in NppShell that fixes it is here: https://github.com/notepad-plus-plus/nppShell/pull/4
- Apart from above, everything works - if the new installer is run on a clean machine, that has never seen Notepad++ before (so the registry keys that needs to be cleaned up isn't there), it works fine. (So it is being registered correctly, just not cleanup'ed correctly from previous installation).

Windows 7
64bit debug:
```
Notepad++ v8.5.1   (64-bit)
Build time : Mar 24 2023 - 19:54:06
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 7 Enterprise (64-bit)
OS Build : 7601.24544
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```
32bit debug:
```
Notepad++ v8.5.1   (32-bit)
Build time : Mar 21 2023 - 19:13:46
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 7 Enterprise (64-bit)
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

Findings:

- Exactly the same as with Windows 10

-----------------------------------------------------------------------

@donho so in short, I have the following two things that needs to be done:

1. This PR needs to be merged into the NppShell repo, and it needs to be rebuilt: https://github.com/notepad-plus-plus/nppShell/pull/4 (Remember to update the submodule, so it is pointing to the latest commit after it is merged).
2: A new installer, based on this PR with the updated commit from 45 minutes ago needs to be built.

GurliGebis:
@donho if you want, you can start by merging the PR mentioned above, and I'll update this PR to point to the latest commit in the NppShell repo - I think that is easier, since the moving to a submodule is in this PR anyway 🙂

GurliGebis:
@donho I just force pushed the branch in NppShell, since I forgot to change a few lines 🙂 if you already have built it, please rebuild it (sorry)

donho:
@GurliGebis 
> Windows 11
64bit debug:

Thank you for testing and your report.
Sorry I don't get your **debug** meaning. All the packages I provided are release mode.

GurliGebis:
@donho output of the Debug dialog inside Notepad++ 🙂

donho:
@GurliGebis 
Please check the newest built from both repo:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST02/

GurliGebis:
@donho looks like the missing cleanup is working, but the installer still complains about the NppModernShell.dll not being there.
Did you pull the changes I pushed in this branch with the changes here before building the installer?
https://github.com/notepad-plus-plus/notepad-plus-plus/compare/33e8d0ca4a12484b16afad70c133e9b4af66ec53..9ebd7c5ac388763608e6b38eefae578744a3a191

GurliGebis:
I have updated the submodule pointer to the latest commits in the NppShell repo

donho:
Yes, https://github.com/notepad-plus-plus/notepad-plus-plus/compare/33e8d0ca4a12484b16afad70c133e9b4af66ec53..9ebd7c5ac388763608e6b38eefae578744a3a191 is included.

I will provide you soon another binary with dialog for making sure it's passed by there.

GurliGebis:
@donho would this work instead?

```
	; Since 8.5.1 was installed using the NppModernShell name, we have to clean that up.
	; But if the files does not exist, we cannot unregister it, so we must skip this line.
	IfFileExists "$INSTDIR\NppModernShell.dll" 0 noDeleteNppModernShell
		Exec 'rundll32.exe "$INSTDIR\NppModernShell.dll",UnregisterSparsePackage'
		Delete "$INSTDIR\NppModernShell.dll"
		Delete "$INSTDIR\NppModernShell.msix"
noDeleteNppModernShell:
```
I'm not really knowledgeable on NSIS 🙂

donho:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST02/

which contains:
```nsis
	; Since 8.5.1 was installed using the NppModernShell name, we have to clean that up.
	; But if the files does not exist, we cannot unregister it, so we must skip this line.
	IfFileExists "$INSTDIR\NppModernShell.dll" 0 +3
		MessageBox MB_OK "UnregisterSparsePackage $INSTDIR\NppModernShell.dll" /SD IDOK
		Exec 'rundll32.exe "$INSTDIR\NppModernShell.dll",UnregisterSparsePackage'

	Delete "$INSTDIR\NppModernShell.dll"
	Delete "$INSTDIR\NppModernShell.msix"
```

donho:
> would this work instead?

It's the same basically.
Please let me know if you've got the dialog from TEST3 packages.

GurliGebis:
@donho TEST3 doesn't show any dialog or errors.
I wonder why it fails with +2 like it is in this PR.

GurliGebis:
@donho okay, if I install the official 8.5.1 installation, and then TEST2 or TEST3, they both fail.
If I clean install TEST2 or TEST3, no errors are shown.

So it looks like the test is not working for some reason

GurliGebis:
@donho we got it backwards 🙂

`IfFileExists "$INSTDIR\NppModernShell.dll" 0 +2`

should be

`IfFileExists "$INSTDIR\NppModernShell.dll" +2 0`

https://nsis.sourceforge.io/Reference/IfFileExists

GurliGebis:
Or am I misunderstanding?

GurliGebis:
@donho okay, I was just thinking....

The installer runs the regsvr32 on the new dll, which will cleanup everything left behind, except the files.
So maybe we should skip the unregister part, and just delete those files.
I can update the regsvr32 part of the NppShell dll, so it moves those files into the temp folder and marks them for delete on next reboot - the result would be the same.

Should we go down that route instead? (the outcome is the same, it is it pretty simple)
I can have it coded up within 10 minutes.

GurliGebis:
@donho please see this PR: https://github.com/notepad-plus-plus/nppShell/pull/5
If we merge that, we can remove this part of the install script:

```
	IfFileExists "$INSTDIR\NppModernShell.dll" 0 +2
		Exec 'rundll32.exe "$INSTDIR\NppModernShell.dll",UnregisterSparsePackage'
```

GurliGebis:
Since the installer runs regsvr32 on the new dll, which will move the files away to the temp folder and schedule them for deletion.

GurliGebis:
@donho I see you have merged it, so I have pushed this branch again with those parts removed and the submodule pointer updated.

Can you do a TEST4? 😀

donho:
New packages:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST04/

GurliGebis:
@donho are you sure the latest NppShell.dll is in there?
The files aren't moved, and if I re-run the regsvr32 multiple times, nothing happens.
But if I build the code and run it, they are moved instantly.
You might need to pull from within the NppShell submodule folder for the changes I pushed in there to be included.

donho:
Sorry, I did pull and rebuild new NppShell, but I forgot to run the cp script:
Now here are the new packages:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST05/

GurliGebis:
@donho IT WORKS 😀
After a clean 8.5.1 installation:

before

After upgrading to TEST05:

after

I'll re-test on Windows 10, just to make sure the registry keys are removed as they should if upgrading from before 8.5.1 and report back, but so far it looks great.

GurliGebis:
@donho can confirm the registry issue is also fixed.
So we should be good to go 😀🚀
I'm off to bed for today

donho:
@GurliGebis 
Thank you very much for making work this enhancement!
I will test them from my side now!
Have a good sleep - you deserve it!

GurliGebis:
@donho thanks.

GurliGebis:
@donho the one with no icon, is it a total clean install?

GurliGebis:
@donho okay, so it was an upgrade from 32bit to 64bit, correct?
If so, I know exactly what the problem is and how to fix it

donho:
> okay, so it was an upgrade from 32bit to 64bit, correct?
If so, I know exactly what the problem is and how to fix it

It could be due to the upgrade from 32bit to 64bit, or not.
How can I do a "Clean install"?

GurliGebis:
@donho install 64bit followed by uninstall, 32bit install followed by uninstall should remove everything.
I'll create a PR with a fix for 32bit to 64bit migration later tonight

GurliGebis:
It is due to that, since the 64bit install doesn't clean up the registry for 32bit stuff - I'll fix that

GurliGebis:
@donho 8.5.1 does not know how to clean up after TEST5, so that would explain it as well.
Downgrading is not something we really can do (people will have to uninstall first, then install an earlier version).
This is due to the old version not knowing how to cleanup the newer versions (and we really cannot change the older versions that people have installed).

donho:
@GurliGebis 
> So all in call, I think we are good to go.

I don't think so: I have experienced this issue (No Icon) in official v8.5.1 and here is the issue created by a user who didn't have downgraded: #13410 

This bug is already in NppModernShell.dll BTW.

GurliGebis:
@donho is it also happening with only TEST5?

donho:
@GurliGebis 
Yes. Here are the TEST5 REPORT/
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13411#issuecomment-1485086576

GurliGebis:
@donho can you give me a step by step I can followed to reproduced the issue? Then I'll see if I can spot the issue

donho:
@GurliGebis 
I just reinstaller x64 of TEST5, uninstall it; reinstaller x32 of TEST5, uninstall it; then installer Notepad++ v8.5.1 x64, I get back the "Edit with Notepad++" entry, but no icon on taskbar issue persist.

GurliGebis:
@donho sounds like a explorer caching issue, since it caches icons.
However, we cannot change 8.5.1, so they are kind of stuck.
You can try pinning and unpinning from the taskbar (I have seen it on other program, where the icon is broken)
The taskbar icon is unrelated to this, but my guess is that it is because both 32bit and 64bit has been installed at some time, so the path the explorer has cached for the program is invalid (it caches based on exe name, and that is identical for both 32bit and 64bit)

To test that theory, rename the exe to something else and run it - if the icon shows in the taskbar, it is a caching issue.

GurliGebis:
But again, this is not related to this PR, since it is only for the context menu - we are not touching anything related to the exe icon (we only read it for the context menu item)

GurliGebis:
@donho okay, I think I understand the issue now.
Is it only present in 8.5.1 or is it also possible to trigger in TEST5?

GurliGebis:
@donho since the #13410 issue is in 8.5.1 as well, it is not something related to the changes in this PR.
So since that is the only issue left, can we merge this, so that part is done? Then we can work on figuring it out over there.

donho:
@GurliGebis 
Are they the latest commits you wanna push?

donho:
@GurliGebis 
Please remove submodule under `PowerEditor/src/tools/` - I 've got some conflict, and I have no idea where come from.
We don't need submodule - it makes things more complicated for nothing.

donho (changes requested):
Please remove submodule under PowerEditor/src/tools/

GurliGebis:
@donho I have removed it - still need to build the latest commit and include it though 😀

GurliGebis:
@donho yes, just take the latest code in NppShell and use that, and we should be good to go

GurliGebis:
@donho I have to go to bed within the next half hour - would you be able to create an installer for me to test before then? If not, I'll test it tomorrow instead.

donho:
@GurliGebis 
Sorry for taking time. New packages:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST06/

GurliGebis:
@donho no worries, I'll take a look right away 😀

GurliGebis:
@donho it seems like I doesn't like the way I try to install the files into a sub folder.

How should I change this, so the dll and msix is installed into a subfolder called contextmenu?

```
	SetOverwrite try
	SetOutPath "$INSTDIR\"

	!ifdef ARCH64
		File /oname=$INSTDIR\contextmenu\NppShell.msix "..\bin64\NppShell.msix"
		File /oname=$INSTDIR\contextmenu\NppShell.dll "..\bin64\NppShell.x64.dll"
	!else ifdef ARCHARM64
		File /oname=$INSTDIR\contextmenu\NppShell.msix "..\binarm64\NppShell.msix"
		File /oname=$INSTDIR\contextmenu\NppShell.dll "..\binarm64\NppShell.arm64.dll"
	!else
		; We need to test which arch we are running on, since 32bit exe can be run on both 32bit and 64bit Windows.
		${If} ${RunningX64}
			; We are running on 64bit Windows, so we need the msix as well, since it might be Windows 11.
			File /oname=$INSTDIR\contextmenu\NppShell.msix "..\bin64\NppShell.msix"
			File /oname=$INSTDIR\contextmenu\NppShell.dll "..\bin64\NppShell.x64.dll"
		${Else}
			; We are running on 32bit Windows, so no need for the msix file, since there is no way this could even be upgraded to Windows 11.
			File /oname=$INSTDIR\contextmenu\NppShell.dll "..\bin\NppShell.x86.dll"
		${EndIf}    

	!endif
	Exec 'regsvr32 /s "$INSTDIR\contextmenu\NppShell.dll"'
```

GurliGebis:
@donho I think I figured it out, I have made this change: https://github.com/notepad-plus-plus/notepad-plus-plus/compare/a4e8c591f555ee19b0a6c392f957c54bd578927f..bf2357f443df4311052b755b4ef04aefcba482ed

Can you try and build the installer and run it, and see if it places the dll and msix files inside a subfolder now?
If it doesn't, do let me know what I should change in the NSIS files (since they are pretty much a black box for me) 😀

donho:
@GurliGebis 

> it seems like I doesn't like the way I try to install the files into a sub folder.
> 
> How should I change this, so the dll and msix is installed into a subfolder called contextmenu?

Sorry, I don't follow you. What's the problem you have encountered?

donho:
Ah, OK I see

GurliGebis:
> @GurliGebis
> 
> > it seems like I doesn't like the way I try to install the files into a sub folder.
> > How should I change this, so the dll and msix is installed into a subfolder called contextmenu?
> 
> Sorry, I don't follow you. What's the problem you have encountered?

When I install it, it doesn't place the NppShell.dll and NppShell.msix into the subfolder called `contextmenu`.
So it looks like the changes I have made to the installer script didn't work, so I have tried changing something here: https://github.com/notepad-plus-plus/notepad-plus-plus/compare/a4e8c591f555ee19b0a6c392f957c54bd578927f..bf2357f443df4311052b755b4ef04aefcba482ed

If you can try and build the installer with that commit, and try and run it on a test machine, and see if it places the files in there, that would be great.
If it doesn't please guide me on what I should change for it to do that, since I'm not really familiar with NSIS.

donho:
@GurliGebis 
So it's better to remove contextmenu now.
Just find all the `contextmenu` in nsis files.

GurliGebis:
@donho If I do that, wouldn't it install the files in the same directory as notepad++.exe ?
They have to be placed in a subfolder for it to work.

donho:
> They have to be placed in a subfolder for it to work.

That's a new info for me. Why?

GurliGebis:
@donho Here is why 🙂 https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13410#issuecomment-1486936698
Long story short:```

When it registers the sparse package, Windows changes the permissions on the folder it is located inside.
Those permissions causes it to be unable to show the icon when notepad++.exe is launched from the context menu.
The solution is to move the msix and dll into a subfolder, so Windows only changes permissions on that, and not the main folder where notepad++.exe is, since that will cause it to work.
```

**Comment by GurliGebis** [comment]
So by moving those two into a subfolder instead, it "just works" 😀

**Comment by GurliGebis** [comment]
And resetting the ACL on the folder with the MSIX is, from what I can find online, a bad idea, so placing them in a subfolder is the best solution.

**Comment by donho** [comment]
@GurliGebis 
> When I install it, it doesn't place the NppShell.dll and NppShell.msix into the subfolder called `contextmenu`.
> So it looks like the changes I have made to the installer script didn't work, so I have tried changing something here: https://github.com/notepad-plus-plus/notepad-plus-plus/compare/a4e8c591f555ee19b0a6c392f957c54bd578927f..bf2357f443df4311052b755b4ef04aefcba482ed

Without answering the above question, and applying this diff, does TEST6 works for you?

**Comment by GurliGebis** [comment]
No, since it cleans up the files if they are stored in the root folder, they are gone.
So no NppShell.* files in TEST6, so we need to get NSIS tweaked, so it installs into the subfolder.

**Comment by donho** [comment]
OK, I will modify NSIS and try it locally

**Comment by GurliGebis** [comment]
> OK, I will modify NSIS and try it locally

Thanks, just let me know what changes are needed, and I'll apply them and push again.
I'll be going to bed now, but I should have some free time tomorrow before lunch to work on it 🙂

**Comment by donho** [comment]
https://github.com/notepad-plus-plus/notepad-plus-plus/compare/a4e8c591f555ee19b0a6c392f957c54bd578927f..bf2357f443df4311052b755b4ef04aefcba482ed is working.

Check this one:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST06/

**Comment by GurliGebis** [comment]
> https://github.com/notepad-plus-plus/notepad-plus-plus/compare/a4e8c591f555ee19b0a6c392f957c54bd578927f..bf2357f443df4311052b755b4ef04aefcba482ed is working.
> 
> Check this one: http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST06/

@donho yay - okay, now I'm off to bed 🙂

This fixes #13410, should fix #13434, should fix #13425 - can you merge this PR and then release an RC1 for those people to try?

**Comment by GurliGebis** [comment]
@donho one small change please, I forgot to remove the folder on uninstall - please see this one: https://github.com/notepad-plus-plus/notepad-plus-plus/compare/bf2357f443df4311052b755b4ef04aefcba482ed..65b546ee2af8939ae4f3f00bdbd26c80c51988d0
It is pushed now, so this PR should be good to go for merging 🚀

**Comment by donho** [comment]
@GurliGebis

Thank you for your coding effort. 
Here's packages which contain the lastest commit of this PR: 
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST08/

Could you test tomorrow roughly to make sure there's nothing to be left behind?
Once it's OK with your tests, I 'll do from my side.

**Comment by GurliGebis** [comment]
@donho thanks, I'll test later today and let you know 🙂

**Comment by GurliGebis** [comment]
@donho just tested - it seems like it leaves the contextmenu folder behind still, though it is empty.
I think I missed something - there are two places it removes directories.

I have updated the uninstall script to handle it - see the changes I just pushed: https://github.com/notepad-plus-plus/notepad-plus-plus/compare/65b546ee2af8939ae4f3f00bdbd26c80c51988d0..dab6f2873fb6c963e5471a787a9a392f7d4d842a

Hopefully this fixes the issue with the empty folder.

I'll test the rest, except the empty folder, on the other systems and report back.
If the rest work, and the next version you build can remove the folder, we should be good to go.

**Comment by GurliGebis** [comment]
@donho I have added a bit of cleanup, that I would like to have included in this PR: https://github.com/notepad-plus-plus/nppShell/pull/8
We should enable treat warnings as errors anyway, so it can be merged at the same time.
Can you build a TEST9 with a new dll with that PR and the installer changes above?

**Comment by donho** [comment]
@GurliGebis 
OK I'll integrate one last time - But please stop modifying the building package scripts.
If it's necessary, I will do it myself.

**Comment by donho** [comment]
Here we go:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST09/

**Comment by donho** [comment]
This PR is closed temporarily for freezing the code.
It'll be reopened once TEST9 passed.

**Comment by donho** [comment]
@GurliGebis 
Please let me know about your test results.
Once it's OK, I'll do from my side.

**Comment by GurliGebis** [comment]
@donho now the folder is removed like it should 🙂 so now I shouldn't have any more changes to the installer.
If more show up, I'll ask you to do them - no worries.

Can you test on Windows 10 and see if it works fine there as well, so we can get this one merged?
I'll try a clean Windows 10 machine later tonight, since my other test machine is broken.

**Comment by donho** [comment]
@GurliGebis 
Sounds good 👍 
I'll do the test once I get home.
Thank you!

**Comment by GurliGebis** [comment]
@donho You're welcome.
I'll test on Win10 as well tonight

**Comment by GurliGebis** [comment]
@donho I tested it on WIndows 10 and Windows 11 as well, and it works.
Both is with these three cases:
- Clean install of TEST9
- Upgrade from 8.5.0 to TEST9
- Upgrade from 8.5.1 to TEST9

**Comment by donho** [comment]
I have tested under Win11 & Win10. No major issue has been found.

**Comment by donho** [comment]
It works on Win 8.1 Pro. 
Though it still needs to restart windows to make context menu appear.

@GurliGebis It's kind of regression :(

**Comment by GurliGebis** [comment]
@donho okay, I'll take a look at it later tonight 🙂

**Comment by GurliGebis** [comment]
@donho Okay, I figured it out - It's my fault (I'm sorry) 🙂

The issue is in nppSetup.nsi: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/7b4e700d4967267f6434f0ed5e21a08097f540f4/PowerEditor/installer/nppSetup.nsi#L294

This line needs to be moved below this line: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/7b4e700d4967267f6434f0ed5e21a08097f540f4/PowerEditor/installer/nppSetup.nsi#L308

Here is what is happening now, if you upgrade from an older version:
- All the new files are installed.
- The new dll is registered, which makes sure all the correct registry keys are there.
- The old dll is unregistered, which removes all the registry keys it uses - here is the problem - the CLSID registry keys are shared between the old and new dlls, so they are removed.

So, if we move line 294 down below 308, so we get this, it should start working with upgrades (also explain why clean install and Windows 11 works):

```
............

!endif

; Make sure old NppShell dll's are unregistered and removed
Exec 'regsvr32 /u /s "$INSTDIR\NppShell_01.dll"'
Exec 'regsvr32 /u /s "$INSTDIR\NppShell_02.dll"'
Exec 'regsvr32 /u /s "$INSTDIR\NppShell_03.dll"'
Exec 'regsvr32 /u /s "$INSTDIR\NppShell_04.dll"'
Exec 'regsvr32 /u /s "$INSTDIR\NppShell_05.dll"'
Exec 'regsvr32 /u /s "$INSTDIR\NppShell_06.dll"'
Delete "$INSTDIR\NppShell_01.dll"
Delete "$INSTDIR\NppShell_02.dll"
Delete "$INSTDIR\NppShell_03.dll"
Delete "$INSTDIR\NppShell_04.dll"
Delete "$INSTDIR\NppShell_05.dll"
Delete "$INSTDIR\NppShell_06.dll"

Exec 'regsvr32 /s "$INSTDIR\contextmenu\NppShell.dll"'
```

Can I get you to move it and create a new installer with it? (If you want, I can create a PR with the changes)

**Comment by donho** [comment]
> * The old dll is unregistered, which removes all the registry keys it uses - here is the problem - the CLSID registry keys are shared between the old and new dlls, so they are removed.
> 
> 
> So, if we move line 294 down below 308, so we get this, it should start working with upgrades (also explain why clean install and Windows 11 works):

OK. I'll do a PR for fix it.
I'll do all the PR myself for all about packaging scripts. Thank you.

**Comment by GurliGebis** [comment]
@donho thanks, if you want, you can add me as a reviewer on it.

**Comment by GurliGebis** [comment]
@donho also, I still think, ever if there are a few issues, that this should be in 8.5.2, since it improves the situation from 8.5.1 where things are half implementere 🙂
But lets see how it works with the changes above

**Comment by GurliGebis** [comment]
@donho check https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13448#issuecomment-1491073098 - we are dealing with an Explorer issue it seems.
Unloader DLLs from a process seems to be the cause of the issue (Explorer really don't like it it seems).

**Comment by GurliGebis** [comment]
So no matter how much we try working around it, I think we won't be able to get it to work, since Explorer doesn't update the context menu after a DLL has been unloaded (is what it seems).
So they best we can do, is to notify people that they might need to restart explorer (it's a one time thing though, since once the new DLL is loaded, the old implementation is gone, and going forward, it won't be there any more, so nothing to try and unload).

**Comment by donho** [comment]
> So no matter how much we try working around it, I think we won't be able to get it to work, since Explorer doesn't update the context menu after a DLL has been unloaded (is what it seems).
> So they best we can do, is to notify people that they might need to restart explorer (it's a one time thing though, since once the new DLL is loaded, the old implementation is gone, and going forward, it won't be there any more, so nothing to try and unload).

I'm not talking about new installation under Win11, but all the upgraded Notepad++ under Win11, Win10 and previous version devices - the context menu should keep working just after upgrading Notepad++.

In the worst case, we keep using the old NppShell.dll.

**Comment by GurliGebis** [comment]
@donho we really should accept a single time, at people have to restart explorer or Windows once.
Going with the old 8.5.1 is not really a good solution, since that has a lot of other issues.

My suggestion would be having installer detect it is removing NppShell_06.dll (see if it exist before deleting it), and then asking the user the following messagebox:  “Due to changes in context menus, please restart Explorer - this is a one time occurance. Would you like ud to restart Explorer for you?"
If they click yes, after we delete the file, we restart the explorer process for them.

If we do it like that, we have automated it, and like I have mentioned, it will only be when upgrading from 8.5.1 and below to 8.5.2 and above, so a one time thing.

It would really be a shame sticking to the broken 8.5.1 installation, since it has other issues and is only half done.

A lot of other installeres just explorer without even telling or asking the user - I really think that we can do better than that - but having to ask the user one time, should be fine.
Eventually Microsoft will stop supporting the legacy API anyway, and then we are forced to move away from it.

**Comment by GurliGebis** [comment]
So if we do it like that, going from 8.5.1 to 8.5.2 will ask the user.
But then, later on when going from 8.5.2 to 8.5.3, nothing will happen, since it is only once.

**Comment by GurliGebis** [comment]
We can have the dll do it cleanly, so we just have to rundll32 a function if they click yes.
Here is how I would do it:
1. Add this to a new exported function in the dll: https://blog.adeltax.com/restart-explorer-exe-shell-in-c
2. Have you update the installer with the question above, and if they click yes, have it rundll32 that function.

**Comment by GurliGebis** [comment]
That would be the cleanest way to do it, and it is only once 🙂

**Comment by GurliGebis** [comment]
Okay, I have a working implementation of it - here is the function I will add to the dll:

```
STDAPI ReloadExplorer()
{
    // First we locate the correct Explorer process, based on the "Shell_TrayWnd" window class.
    HWND hwnd = FindWindowW(L"Shell_TrayWnd", NULL);
    
    // Then we find the process id of the process owning that window (which is the correct Explorer instance).
    DWORD pid(0);
    GetWindowThreadProcessId(hwnd, &pid);

    if (!pid)
    {
        return ERROR_FILE_NOT_FOUND;
    }

    // Now we have the pid of the process, we can open it.
    HANDLE hExplorer = OpenProcess(PROCESS_TERMINATE, false, pid);

    // Then we just restart the process.
    // Passing 2 as the 2nd parameter tells Explorer to restart automatically.
    TerminateProcess(hExplorer, 2);

    // Finally we cleanup by closing the handle.
    CloseHandle(hExplorer);

    return S_OK;
}
```

Here is the code that would be needed to add to the installer in between the Delete of NppShell_05 and NppShell_06:
```
	Delete "$INSTDIR\NppShell_05.dll"

	IfFileExists "$INSTDIR\NppShell_06.dll" 0 +3
		MessageBox MB_YESNO "Due to changes in context menus, please restart Explorer - this is a one time occurance. Would you like us to restart Explorer for you?" /SD IDYES IDYES doReloadExplorer IDNO noReloadExplorer ;IDYES restart
doReloadExplorer:
		Exec 'rundll32 "$INSTDIR\contextMenu\NppShell.dll,ReloadExplorer"'
noReloadExplorer:

	Delete "$INSTDIR\NppShell_06.dll"
```

That way we only ask if the NppShell_06 dll exists (so it is a one time thing, since it is removed here, so people won't ever see it again.
Only thing I'm unsure of, is if `+3` is the right number of lines to jump.

What do you think @donho?
This is the best we can do, if you ask me, since going back really really REALLY, isn't the way to go.

GurliGebis:
Also, all open programs (except file explorer windows) stats open, so no dataloss or anything 😃

GurliGebis:
@donho I have created a PR with the dll function here: https://github.com/notepad-plus-plus/nppShell/pull/12
The rest is just updating the install script, and you should be good to go.

I don't want to repeat myself, but this is really the way to go. (And the way other projects do, when they make changes to explorer configuration - if you think about it, it is quiet common for some installers to want to restart Explorer or even Windows, to make changes).

donho:
@GurliGebis 
I see your point.

Here is the minimum criteria that I find reasonable:
* The appearance of context menu entry after restarting explorer/os is acceptable, **only** on the 1st time installation: such situation contains both clean installation & update from nppShell_06.dll to new nppShell.dll.
* OTOH, for every update from new nppShell.dll (both x32 & x64) to new nppShell.dll, the context menu entry **should not** disappear after the updating installation, even it will reappear after restarting explorer/os.

I do believe TEST10 meets this criteria. But we need more tests.

> I have created a PR with the dll function here: https://github.com/notepad-plus-plus/nppShell/pull/12

This one rather a dirty hack and it won't be merged into master.

GurliGebis:
@donho that is the cleanest way to restart explorer, isn't it?
Upgrading (or just installing TEST10 on top of itself should test that quickly, since that is the same procedure as a future upgrade will be.

I'm not at my computer today, but I think using the dll function is the cleanest way, since it locates the rigt process (in case there are multiple), and restarts the right one.

Do you know of a better way to do it? (it's the best I could find, and it seems to be both simple and stable from the tests I did) 🙂

donho:
> I'm not at my computer today, but I think using the dll function is the cleanest way, since it locates the rigt process (in case there are multiple), and restarts the right one.
> 
> Do you know of a better way to do it? (it's the best I could find, and it seems to be both simple and stable from the tests I did)

No application should stop any process - it's even worse than hacking "Pin to Quick access" in registry.
Besides, the blink effect is a bad User XP. So no, we don't include such "feature".

OTOH, we can include a note in [Notepad++ User Manual](https://npp-user-manual.org/), to explain the installing and upgrade () result, and suggest user to restart explorer or windows.

GurliGebis:
@donho If just mentioning it in the guide is okay with you, I would prefer that as well.
The less hacky stuff, the better, totally agree 👍👍👍

donho:
I'm worry about the error dialog of unregistering NppModernShell.dll in the new version of NppShell.dll. Could it be silent?

GurliGebis:
It is only shown when uninstalling 8.5.1 from what I know?
Upgrading from 8.5.1 to TEST10 shouldn't show any errors.

GurliGebis:
and once you are on a later version than 8.5.1, the uninstaller shouldn't try to touch NppModernShell.dll anymore, since it is no longer there in that version.

PR #13416, opened 2023-03-25
Fix col editor always inserting even num of chars for hex num

### LINKED ISSUE #13413: ColumnEditor hex number insertion always uses even number of digits
### Description of the Issue
The _Column Editor_ number insertion feature always inserts its hex numbers with an _even_ number of field-width characters, when it should insert such numbers using the minimum number of characters needed to suit the value(s) of the numbers inserted.

This bug was mentioned here, by @Yaron10 : https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13336#issuecomment-1483234273

1. Set up the following _Column Editor_ operation and execute it on the data indicated in `new 1` tab:
image

### Actual Behavior
Resulting numbers inserted are done so with a field-width of 4, even though the hex numbers generated only need a field-width of 3:

image

### Expected Behavior
image

### Comments

Bug caused by the following code:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c59c07b4fae16c83f545e0da8772950d03ac4988/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L178-L179

The reason for the above code is probably lost to history...

### Debug Information

Evaluate bug after this commit:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/7f54a4b959c51323db635f76a4c037cb8f432c0e
but note that the bug is present _before_ the commit, in N++ 8.5.1 (when using _Leading Zeros_ checkbox either checked or unchecked).

alankilborn:
Fix #13413

Also modernizes code by changing French "chiffre" to English "digits"; hope that is OK to do.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -161,7 +161,7 @@ LanguageNameInfo ScintillaEditView::_langNameInfoArray[L_EXTERNAL + 1] = {
 
 int getNbDigits(int aNum, int base)
 {
-	int nbChiffre = 1;
+	int nbDigits = 1;
 	int diviseur = base;
 
 	for (;;)
@@ -172,13 +172,11 @@ int getNbDigits(int aNum, int base)
 		else
 		{
 			diviseur *= base;
-			++nbChiffre;
+			++nbDigits;
 		}
 	}
-	if ((base == 16) && (nbChiffre % 2 != 0))
-		nbChiffre += 1;
 
-	return nbChiffre;
+	return nbDigits;
 }
 
 void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)


Yaron10:
👍 
Thank you.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13413#issuecomment-1486109795

rdipardo:
> Also modernizes code by changing French "chiffre" to English "digits"

I though French was already [a "modern" language], as distinct from a "classical" one like Latin or Aramaic?

As for C++, identifiers can even contain [literal UTF-8 characters], French ones included, i.e. this works:

~~~cpp
// i18n.cxx
#include <iostream>
#include <locale>

struct virguleFrançaise : std::numpunct<char> {
    char_type do_decimal_point() const override { return ','; };
};

int main(void)
{
  int nombreŒufs = 6;
  float mlDuCafè = 473.18f;
  std::cout.imbue(std::locale(std::cout.getloc(), new virguleFrançaise()));
  std::cout << nombreŒufs << " oeufs et " << mlDuCafè << " ml du caf\xe8.\n";
  return 0;
}
~~~
~~~sh-session
> cl /EHsc /utf-8 i18n.cxx
Microsoft (R) C/C++ Optimizing Compiler Version 19.35.32216.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

i18n.cxx
Microsoft (R) Incremental Linker Version 14.35.32216.1
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:i18n.exe
i18n.obj

> i18n.exe
6 oeufs et 473,18 ml du cafè.

~~~
[a "modern" language]: https://en.wikipedia.org/wiki/Modern_language#:~:text=to%2Dday%20communication%20(-,such%20as%20French,-and%20German)%20and

[literal UTF-8 characters]: https://learn.microsoft.com/en-us/cpp/cpp/character-sets?view=msvc-170#universal-character-names:~:text=Character%20Literals.-,Microsoft%20Specific,-The%20Microsoft%20C

alankilborn:
>I thought French was already a "modern" language

I meant no disrespect to French, and I wasn't putting "modern" together with "French"; no, not at all.  I said "modernize code".

Once upon a time, coders around the world named variables in their own language; I refer to this as the "old days" of programming.  In my world (non-native English speaker), I've always been taught that source code should be in English, so I have always coded in English.  Notepad++ source code is mostly English, so "modernizing" some old choices for variable names makes sense to me.

PR #13427, opened 2023-03-26
FindInFinder MatchCase and WholeWord logic change

### LINKED ISSUE #13415: "Find in Search Results": Default "Whole Word" & Case to false
**STR:**
Open the Find dialog and execute a command displaying the Search Results window.
Open the "Find in Search Results" dialog.

**Result:**
The "Whole Word" & Case options are checked.

**Expected:**
Those options should be unchecked by default (just like the Find dialog)

***

Notepad++ 8.5.1

alankilborn:
Fix #13415

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -232,6 +232,7 @@ public:
 private:
 	Finder  *_pFinder2Search = nullptr;
 	FindOption _options;
+	bool firstTimeInit = true;
 	
 	virtual intptr_t CALLBACK run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam);
 	void initFromOptions();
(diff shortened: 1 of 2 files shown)


Yaron10:
@alankilborn,

👍 
Thank you.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -232,6 +232,7 @@ class FindInFinderDlg : public StaticDialog
 private:
 	Finder  *_pFinder2Search = nullptr;
 	FindOption _options;
+	bool firstTimeInit = true;
```
donho, line 235: Instead of adding new member variable, I would 	change the value of `_options._isWholeWordand` & `_options._isMatchCase` to `false` in the constructor of FindInFinderDlg.

donho (changes requested):
Make the modification as simple as possible.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
alankilborn: You're right; it is simpler and thus better.

alankilborn:
Ready for re-review.

donho (approved):

PR #13448, opened 2023-03-30
Fix "Edit with Notepad++" context menu needs restarting OS issue

### LINKED ISSUE #13446: v8.5.1 for Windows now needs AppXSVC always running
I'm running W10 x64 & Notepad++ v.851

This new version notepad++.exe needs AppXSVC running always.
I understand AppXSVC is needed to install *.msix modern shell package... but I don't install it, and I see no other reason to require AppXSVC so I don't understand why now this service is a must.

Can someone please explain the reasoning behind this? Can this be handled in a different way so this service isn't absolutely necessary?

### LINKED ISSUE #13457: Notepad++ is now dependent on the service "Application Information"
I have a hardened system with lots of services disabled. Since 8.5.1 (I'm using the portable version), NPP won't start unless the service "Application Information" ("AppInfo") is running. Otherwise, the error message "There are no more endpoints available from the endpoint mapper." is shown. 

I have never seen an app with this dependency before. Obviously, you can't support tweaked systems, but I'd like my favourite text editor not to be too much dependent on non-essential services. I like NPP because its small footprint and lack of dependencies, which means you know you can always rely on it, eg. when diagnosing a damaged / partly corrupt Windows machine.

donho:
Fix context menu disappears issue while update.

It works with https://github.com/notepad-plus-plus/nppShell/pull/14 to fix
https://github.com/notepad-plus-plus/nppShell/issues/13

It fixes also the added dependency of the service "Application Information".

Fix #13446, fix #13457

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -267,7 +267,7 @@ ${MementoSection} "Context Menu Entry" explorerContextMenu
 
 	SetOverwrite try
 	SetOutPath "$INSTDIR\contextMenu\"
-
+/*
 	; Make sure old NppShell dll's are unregistered and removed
 	Exec 'regsvr32 /u /s "$INSTDIR\NppShell_01.dll"'
 	Exec 'regsvr32 /u /s "$INSTDIR\NppShell_02.dll"'
@@ -281,7 +281,7 @@ ${MementoSection} "Context Menu Entry" explorerContextMenu
 	Delete "$INSTDIR\NppShell_04.dll"
 	Delete "$INSTDIR\NppShell_05.dll"
 	Delete "$INSTDIR\NppShell_06.dll"
-	
+*/
 	
 	!ifdef ARCH64
 		File /oname=$INSTDIR\contextMenu\NppShell.msix "..\bin64\NppShell.msix"
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -262,11 +262,6 @@ FunctionEnd
 
 !include "nsisInclude\themes.nsh"
 
-!include "StrFunc.nsh"
-${Using:StrFunc} StrStr
-
-Var muiVerbStr
-Var nppSubStr
 
 ${MementoSection} "Context Menu Entry" explorerContextMenu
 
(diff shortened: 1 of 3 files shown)


donho:
@GurliGebis 
Please review this PR and test the following packages:
http://download.notepad-plus-plus.org/repository/MISC/nppShell.TEST10/

GurliGebis:
@donho it looks correct.
I have tested it, and it is working fine on clean installs.

From all my testing and examining what is happening, I have come to the conclusion there is a bug in Explorer.
If you remove a context menu handler and add another one, Explorer needs to restart before it notices the new one.

If I install 8.5.1 and upgrade to this one, it is missing - if I compare the registry, everything is there now with TEST10.
If I clean install without 8.5.1 being there first, it works (the registry contains the same entries).

So the only conclusion I can come to, is that it is a bug in Explorer that causes it to not really like command handlers being removed. (Which makes sense, since it unloads a DLL from it, which is something Windows don't really support - see: https://stackoverflow.com/questions/8832381/can-i-unload-a-dll-from-another-process-win32)

So really, the only option we have, is to accept that on older versions of Windows, Explorer might have to restart before the "Edit with Notepad++" shows up.
It's going to be a one time thing though, since from 8.5.2 going forward, the DLL and registry entries are already there, so nothing to worry about.

Now the question is how do we tell users this in the easiest way?

GurliGebis:
Anyway, I'm off to bed for tonight.
But as I mentioned, there doesn't seems to be a safe way to unload a dll (see the linked article), so it is more a question of how do we tell users about it.

GurliGebis:
Just tested this on Win10:

1. Install old version
2. Unregister the old dll manually
3. Restart explorer
4. Upgrade to TEST10

This works fine, which points again to the unloading of the dll not really being supporter by Windows.
If you think about it, unloading potentially running code from an applications address space is an unsafe operation by design.

GurliGebis:
Come to think of it, uninstalling the old one only remove the registry entries, so most likely the old dll code is still loaded (Since you cannot force it to unload).
Not sure how to test that, but it doesn't make any difference anyway, it's just me thinking aloud while trying to all asleep 🙂

donho:
@GurliGebis 
Please review it.

GurliGebis:
@donho this should be fine, if we include notepad-plus-plus/nppShell#15, since it will schedule those files for removal on the next reboot, which will keep the menu working until then, and then on the next reboot, the new menu should take over at the same time.

GurliGebis:
@donho we have tested it now, so it is good to go 🙂 looking forward to RC1 of 8.5.2, so we can get a bit broder feedback on it

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e28db057544acdef37741b465d883c7ccae0862e

PR #13453, opened 2023-03-31
Add option to suppress warning when opening big files

DerivativeOfLog7:
Add an option to suppress the >=2GB file warning. Won't show on a x86 build. This is mostly useful when leaving npp searching inside files unattended as the warning could pop up and halt all progress until acted upon


lorenzo3421:
Would be useful 👍

donho:
Is there any opened issue about it?

DerivativeOfLog7:
> Is there any opened issue about it?

I don't know, but I added this out of necessity and figured I could make a pull request

donho:
Then please create one and attach this PR to it (ref it) - by doing it doesn't mean that PR will be accepted, but at least we have the full description & information to estimate if the PR is needed.
Ref:

DerivativeOfLog7:
Done; also had to reopen as #14053

Issue #13402, opened 2023-03-24
Panels GUI issues after running "Find in Search Results" 


Yaron10:
**STR:**
Open a panel (e.g. Functions List, Folder as Workspace).
Open the Find dialog and execute a command displaying the Search Results window.
Open the "Find in Search Results" dialog and run a search.

**Result:**
The panels GUI is messed up (background, no border, Search Edit Control).

![תמונה](https://user-images.githubusercontent.com/10229320/227509723-1eeefe64-81f0-4cb4-87f0-24ccc317b614.png)

_Dark Mode:_

![תמונה](https://user-images.githubusercontent.com/10229320/227512880-7cadcf76-2d49-4ddc-ab27-06a4513b456a.png)

**Expected:**

![תמונה](https://user-images.githubusercontent.com/10229320/227513463-a613abb3-9a25-47fb-bede-b9031bf299ba.png)

![תמונה](https://user-images.githubusercontent.com/10229320/227513371-93305190-7cd5-499c-ba02-8c9b11c09ae2.png)

***

Notepad++ v8.5 (64-bit)
Build time : Mar 19 2023 - 23:56:22
Path : C:---notepad++.exe
Command Line :
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 20H2
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins :
ComparePlus (1.1)
HTMLTag (1.3.6)
MenuIcons (2.0.2)
PythonScript (2)
_CustomizeToolbar (4.2)

alankilborn:
Can you clarify exactly _what_ is "messed up"?
I'm guessing the dark background shading, but maybe something else?

Yaron10:
@alankilborn,

I've edited the post.
Can't you reproduce that on your machine?

alankilborn:
I have scripting do an automatic "empty" _Find All in Current Document_ search when N++ starts up (to get _Search results_ visible for _other_ scripting to act upon).

I have noticed on rare occasion that my _Function List_ (also showing after N++ starts) will have the background shading in your first screenshot.

Yaron10:
And running "Find in Search Results"?
This is _consistent_ on my machine.

Yaron10:
@alankilborn,

Thank you for confirming. 👍

ozone10:
@Yaron10 
[...]

PR #13459, opened 2023-04-01
Fix dock panels visual glitch

ozone10:
caused by cmd 'Find in Search Results'

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13402


ozone10:
@Yaron10
could you test?

donho:
#13402 reproduced, PR tested, bug fixed.

Yaron10:
@ozone10,

👍 
Thank you for another great fix.

Issue #13342, opened 2023-03-13
A recorded macro with "Find Next" and then "Run until the end of file" crashes Notepad++


BdR76:
### Description of the Issue
When you record a macro that includes a step with "Find Next", and you select "Run a Macro Multiple times" with the option "Run until the end of file", then it will get stuck in a loop and Notepad++ freezes.

### Steps to Reproduce the Issue
Example file

          <mappingPart>
            <source column="v_0" />
            <destination observationCode="demin_leefstijl_0" />
          </mappingPart>
          <mappingPart>
            <source column="v_1" />
            <destination observationCode="demin_leefstijl_1" />
          </mappingPart>
          <mappingPart>
            <source column="v_2" />
            <destination observationCode="demin_leefstijl_2" />
          </mappingPart>

1. Type the text `pivotResultColumn="data" />`, select & copy/cut it so it's on the clipboard
2. Go to beginning of the example file
3. Go to menu `Macro > Start recording`
4. Ctrl+F to find next `<source column="v` to move the cursor to next instance
5. Press `End` key to go to end of that line
6. select last two characters `/>`
7. Paste `pivotResultColumn="data" />`
8. Go to menu `Macro > Stop recording`
9. Now go to `Macro > Run a Macro Multiple times`
10. Select `Run until the end of file` and press OK.

### Expected Behavior
The macro should run until the end of the file, so break/stop when "Find Next" wraps around to the beginning of the file, or alternatively when "Find Next" yields no more results.

### Actual Behavior
The macro runs indefinitely and Notepad++ freezes

### Debug Information
Notepad++ v8.4.8   (64-bit)
Build time : Dec 24 2022 - 19:40:27
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows Server 2019 Standard (64-bit) 
OS Version : 1809
OS Build : 17763.3887
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    CSVLint (0.4.6.2)
    CsvQuery (1.2.8)
    HexEditor (0.9.12)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NPPJSONViewer (1.40)
    NppMarkdownPanel (0.5)
    PoorMansTSqlFormatterNppPlugin (1.6.13.31508)
[...]

PR #13478, opened 2023-04-05
fix infinite loop in wrapped find/replace macros

molsonkiko:
This would address #13342.
Tested on Win32 and x64 builds.

In the past, the macro runner didn't check if the direction the cursor was moving changed. By adding a check to make sure that the cursor didn't switch from moving down to moving up from one macro run to the next, we avoid infinite loops.

The easiest way to test that this works is to do the following:
1. create a file containing the text
```
f
f
```
2. Start recording a macro.
3. Ctrl+F for `f`. Make sure that `Wrap around` is turned *on*!
4. Stop recording the macro.
5. Go to `Run a macro multiple times` and `run until end of file`. In previous versions of Notepad++, this will cause an infinite loop. In this version, it should stop at some point, though not necessarily the last line in the file.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1496,35 +1496,35 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 					{
 						if ( counter >= times )
 							break;
+						continue;
 					}
-					else // run until eof
+					// else run until eof
+					
+					// the line number must monotonically increase or decrease
+					// if direction ever changes, bail out to avoid infinite loop
+					if ((counter > 1) && ((deltaCurrLine < 0) != cursorMovedUp))
+						break;
+					cursorMovedUp = deltaCurrLine < 0;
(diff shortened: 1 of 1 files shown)


donho:
1. `molsonkiko:master` should not be used, you should create a new branch for each PR
2. `continue` should be avoided.
3. Make your PR as simple as possible.
4. as a result, only your 1st commit has been tested and it fixes indeed infinite loop, but it make this feature broken - "run until to end of line" run only one time. The PR has not been tested enough.

Contribution rules: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#pull-requests

molsonkiko:
Thank you for your feedback! I'll circle back and try to improve this. However, I'm a bit confused where you said this:

> but it make this feature broken - "run until to end of line" runs only once.

I'm only aware of a "run until end of file" feature, not a "run until end of line" feature. Did you mean to say "run until end of file"?

donho:
> I'm only aware of a "run until end of file" feature, not a "run until end of line" feature. Did you mean to say "run until end of file"?

Yes, sorry for the confusion. I was talking about  "run until end of file"

PR #13492, opened 2023-04-08
[xml] Update german.xml

schnurlos:
Following https://github.com/notepad-plus-plus/notepad-plus-plus/commit/508b3fabd657d32488eb9105d0cdc90055f55d2b


schnurlos:
Double entry, sorry.

PR #13508, opened 2023-04-11
[xml] Update french.xml

Mateos81:
New lines, restored comments, also spotted a wrong line while going through


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/76afd9284eed64b3bf4374c322db0607d996a771

PR #13537, opened 2023-04-15
[xml] Update spanish to v 8.5.3

P0rsche-911:
New revision and update


P0rsche-911:
I have made new corrections

Issue #13221, opened 2023-02-24
Notepad++ v8.4.9 regression: Folder display mess in non-western languges under Macro/Run menu 


rddim:
### Description of the Issue

The issue come from https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6e7901ecb311670e03fbe23adebd2417ea483d08. After creating a subfolder in Cyrillic everything seems normal. But when a macro/command change (rename, delete or record new one) is made via the GUI, the created submenu does not appear readable the next time Notepad++ is started. After opening `shortcuts.xml` it is seen that the subfolder name is not readable either.
### Steps to Reproduce the Issue

1. Make sure you have the latest v8.4.9
2. Open the `shortcuts.xml`
3. Follow the instructions in the file and add `макрос` for the submenu (i.e. `FolderName="макрос"`) to the existing `Trim Trailing Space and Save` macro
4. Save and close the file
5. Open `notepad++.exe` and rename the macro `Trim Trailing Space and Save` to something else.
6. Close and open Notepad++ again
7. Check the macro subfolder name

### Expected Behavior

I have expected to have a readable macros subfolder after changing something.
![npp_shortcuts_readable](https://user-images.githubusercontent.com/8998791/208510774-289d45cf-2401-468b-b58a-0c0fe66e43ac.png)
### Actual Behavior

After changes the subfolder name in `shortcuts.xml` is stored as `&#x043C;&#x0430;&#x043A;&#x0440;&#x043E;&#x0441;` or `&lt;0:@&gt;A` instead of `макрос`
![npp_unread_macro](https://user-images.githubusercontent.com/8998791/221255794-cdc6a741-b096-48b6-b1b0-d51355f672b3.png)

### Debug Information

```
Notepad++ v8.4.9   (64-bit)
Build time : Jan 27 2023 - 03:11:16
Path : D:\Install\Office Programs\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : D:\Install\Office Programs\Notepad++\_settings
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1251
Plugins : 
    AutoSave (1.6.1)
    ComparePlus (1.1)
    TakeNotes (1.2.4)
    VisualStudioLineCopy (1.0.0.2)
```

@donho pointed here

ArkadiuszMichalski:
@donho 
[...]

Issue #13222, opened 2023-02-24
Incorrect storing of Cyrillic in the `config.xml` and `shortcuts.xml`


rddim:
### Description of the Issue

When I use Cyrillic it is not stored in readable form in the `config.xml` and `shortcuts.xml` files.
### Steps to Reproduce the Issue

For `config.xml`
1. Use any version of Notepad++
2. Do a Search (Ctrl+F) in Cyrillic with `кирилица`
3. Close Notepad++ to save the changes in `config.xml`
4. Open `config.xml` and see the the first line starting with `<Find name=` - it should contain incorrect `"&#x043A;&#x0438;&#x0440;&#x0438;&#x043B;&#x0438;&#x0446;&#x0430;"` instead of proper`"кирилица"`

For `shortcuts.xml`
1. Record a macro with the actions above
2. Give it a name `макрос`
3. Close Notepad++ to save the changes in `shortcuts.xml`
4. Open the `shortcuts.xml`
5. See the line starting with `<Macro name=` - it should be stored incorrectly as `"&#x043C;&#x0430;&#x043A;&#x0440;&#x043E;&#x0441;"` instead of proper `"макрос"`
6. See the line starting with `<Action type="3" message="1601" wParam="0" lParam="0" sParam=` - it should contain incorrect `"&#x043A;&#x0438;&#x0440;&#x0438;&#x043B;&#x0438;&#x0446;&#x0430;` instead of proper`"кирилица"`
### Expected Behavior

I have expected to have readable cyrillic in this files, because sometimes I need to make changes and I don't know what is what
### Actual Behavior

The cyrillic is stored as `&#x04XX;` instead of human readable cyrillic
### Debug Information

```
Notepad++ v8.4.9   (64-bit)
Build time : Jan 27 2023 - 03:11:16
Path : D:\Install\Office Programs\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : D:\Install\Office Programs\Notepad++\_settings
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1251
Plugins : 
    AutoSave (1.6.1)
    ComparePlus (1.1)
    TakeNotes (1.2.4)
    VisualStudioLineCopy (1.0.0.2)
```

@donho pointed here

donho:
For shortcuts.xml it's been fixed in the previous version, and the current version works fine.
[...]

Issue #12670, opened 2022-12-19
Incorrect display of Cyrillic in the macros submenu


rddim:
### Description of the Issue

When I use Cyrillic in the macros submenu it appears unreadable in the `Macro` menu.
### Steps to Reproduce the Issue

1. Download the latest v8.4.8 RC2 `npp.8.4.8.portable.minimalist.x64.7z` from http://download.notepad-plus-plus.org/repository/8.x/8.4.8.RC2/
2. Open the `shortcuts.xml`
3. Follow the instructions in the file and add `макрос` for the submenu (i.e. `FolderName="макрос"`) to the existing `Trim Trailing Space and Save` macro
4. Save and close the file
5. Open `notepad++.exe` and click on `Macro` menu

### Expected Behavior

![npp_shortcuts_readable](https://user-images.githubusercontent.com/8998791/208510774-289d45cf-2401-468b-b58a-0c0fe66e43ac.png)

### Actual Behavior

![npp_shortcuts_unreadable](https://user-images.githubusercontent.com/8998791/208510844-8b034712-ea27-4d02-901e-09429063730e.png)

### Debug Information

```
Notepad++ v8.4.8   (64-bit)
Build time : Dec 19 2022 - 02:31:02
Path : X:\xxx\npp.8.4.8.portable.minimalist.x64\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 7 Professional (64-bit) 
OS Build : 7601.0
Current ANSI codepage : 1251
Plugins : none
```

rdipardo:
The bug is in the [TinyXML](https://sourceforge.net/p/tinyxml/bugs/) library. It parses files with [`fgetws()`] (a.k.a. `generic_fgets`) [^1], which does not properly decode multi-byte text.

<img width="732" alt="tinyxml-fgetws-wrong-decoding" src="https://user-images.githubusercontent.com/59004801/208567949-d59bd001-5f8d-4ea2-8bbb-c547ebf306da.png">

An explicit encoding declaration at the top of `shortcuts.xml` does not seem to make any difference [^2].

[^1]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/58416114da14536dcc7a8ad36010da3382366ba6/PowerEditor/src/TinyXml/tinyxml.cpp#L761-L764
[^2]: i.e., `<?xml version="1.0" encoding="UTF-8"?>`

[...]

PR #13538, opened 2023-04-15
Add the UTF8 character ability for Run & Macro menu

### LINKED ISSUE #13488: shortcuts.xml gets mixed line-endings on macro record and save
### Description of the Issue
See StR.

1. Extract a fresh portable version of 8.5.2.
2. Open its `shortcuts.xml` with it.
3. Do regex search for `(?<!\r)\n` and observe 0 hits.
4. Record and save any macro.
5. Restart.
6. Do regex search for `(?<!\r)\n` in `shortcuts.xml`.

### Expected Behavior
No hits.

### Actual Behavior
3 hits in the file.
Turn on visible line-endings.
Observe LF line-endings on lines 3 through 5 while the rest of the file has CRLF endings:
image
image

### Debug Information
Notepad++ v8.5.2   (64-bit)
Path : W:\npp.misc\_NoBackup\Releases\npp.8.5.2\test\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2728
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
Add the UTF8 character ability for Run & Macro menu by using GUI or by editing shortcut.xml.
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13221#issuecomment-1457325172

Fix #13488

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
+++ b/PowerEditor/src/MISC/PluginsManager/PluginsManager.cpp
@@ -597,23 +597,23 @@ void PluginsManager::addInMenuFromPMIndex(int i)
 
 		int cmdID = ID_PLUGINS_CMD + static_cast<int32_t>(_pluginsCommands.size() - 1);
 		_pluginInfos[i]->_funcItems[j]._cmdID = cmdID;
-		generic_string itemName = _pluginInfos[i]->_funcItems[j]._itemName;
+		string itemName = wstring2string(_pluginInfos[i]->_funcItems[j]._itemName, CP_UTF8);
 
 		if (_pluginInfos[i]->_funcItems[j]._pShKey)
 		{
 			ShortcutKey & sKey = *(_pluginInfos[i]->_funcItems[j]._pShKey);
-            PluginCmdShortcut pcs(Shortcut(itemName.c_str(), sKey._isCtrl, sKey._isAlt, sKey._isShift, sKey._key), cmdID, _pluginInfos[i]->_moduleName.c_str(), j);
+            PluginCmdShortcut pcs(Shortcut(itemName.c_str(), sKey._isCtrl, sKey._isAlt, sKey._isShift, sKey._key), cmdID, wstring2string(_pluginInfos[i]->_moduleName, CP_UTF8).c_str(), j);
 			pluginCmdSCList.push_back(pcs);
-			itemName += TEXT("\t");
+			itemName += "\t";
 			itemName += pcs.toString();
 		}
 		else
 		{	//no ShortcutKey is provided, add an disabled shortcut (so it can still be mapped, Paramaters class can still index any changes and the toolbar wont funk out
             Shortcut sc(itemName.c_str(), false, false, false, 0x00);
-            PluginCmdShortcut pcs(sc, cmdID, _pluginInfos[i]->_moduleName.c_str(), j);	//VK_NULL and everything disabled, the menu name is left alone
+            PluginCmdShortcut pcs(sc, cmdID, wstring2string(_pluginInfos[i]->_moduleName, CP_UTF8).c_str(), j);	//VK_NULL and everything disabled, the menu name is left alone
 			pluginCmdSCList.push_back(pcs);
 		}
-		::InsertMenu(_pluginInfos[i]->_pluginMenu, j, MF_BYPOSITION, cmdID, itemName.c_str());
+		::InsertMenu(_pluginInfos[i]->_pluginMenu, j, MF_BYPOSITION, cmdID, string2wstring(itemName, CP_UTF8).c_str());
 
 		if (_pluginInfos[i]->_funcItems[j]._init2Check)
 			::CheckMenuItem(_hPluginsMenu, cmdID, MF_BYCOMMAND | MF_CHECKED);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -5346,8 +5346,8 @@ bool Notepad_plus::addCurrentMacro()
 			::InsertMenu(hMacroMenu, posBase + nbTopLevelItem + 2, MF_BYCOMMAND, IDM_SETTING_SHORTCUT_MAPPER_MACRO, nativeLangShortcutMapperMacro.c_str());
         }
 		theMacros.push_back(ms);
-		macroMenu.push_back(MenuItemUnit(cmdID, ms.getName()));
-		::InsertMenu(hMacroMenu, static_cast<UINT>(posBase + nbTopLevelItem), MF_BYPOSITION, cmdID, ms.toMenuItemString().c_str());
+		macroMenu.push_back(MenuItemUnit(cmdID, string2wstring(ms.getName(), CP_UTF8)));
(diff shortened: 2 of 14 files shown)


alankilborn:
When just beginning to test, I noticed this:

image

ArkadiuszMichalski:
Have the same problem as above. And some other notes:
1. After save/change macro's name we lost some default comments in [`shortcuts.xml` ](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/shortcuts.xml). Only the first one is preserved.
2. >However, the modification of shortcut makes shortcut.xml being rewritten, and the Human readable UTF8 char become raw data - at least it's readable in the menu, and the raw data is the same for the same unicode characters.

   Yes, but it's not limited to just this file, but to all `.xml`. Try searching for `макрос`, close Notepad++ and check what character is in the `config.xml` file (`<Find name="&#x043C;&#x0430;&#x043A;&#x0440;&#x043E;&#x0441;" />`). We just have HTML entities, not plain text. I don't know why TinyXML writes it like that, but if we consider it a bug/defect, it should rather be considered separately.

alankilborn:
>After save/change macro's name we lost some default comments 

>Only the first one is preserved.

And, in that "first one", interestingly, this PR seems to solve #13488 !

Yaron10:
@donho,

Thank you for your work. 👍 
(I haven't tested the new Run structure).

I was about to ask you about the presentation of characters in the XML files. 
@ArkadiuszMichalski mentioned it first:
> close Notepad++ and check what character is in the config.xml file (<Find name="&#x043C;&#x0430;&#x043A;&#x0440;&#x043E;&#x0441;" />).

EditPad displays them in "readable characters" if the encoding is set to:

תמונה

rdipardo:
@ArkadiuszMichalski,
> I don't know why TinyXML writes it like that, but if we consider it a bug/defect, it should rather be considered separately.

As [was mentioned before](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12670#issuecomment-1359902802):

> TinyXML is not being developed anymore.
> 
> Maybe Notepad++ should try to move to [TinyXML2](https://github.com/leethomason/tinyxml2), but this will require major refactoring.

A good question (for a later time): the regex engine and Scintilla components get regular updates; has that XML engine *ever* been updated since the Subversion days?

Yaron10:
@rdipardo,

👍 
Thank you.

donho:
@alankilborn 
> When just beginning to test, I noticed this:

Fixed in the new commit.

> And, in that "first one", interestingly, this PR seems to solve https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13488 !

That's a good "side effect" of this PR. ;) 

@ArkadiuszMichalski 
> After save/change macro's name we lost some default comments in [shortcuts.xml ](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/shortcuts.xml). Only the first one is preserved.

Indeed. But it's not a regression. The way to remove whole node and write the new one makes the comments in the old nodes are removed together. It's a no-humful bug though.

> Yes, but it's not limited to just this file, but to all .xml. Try searching for макрос, close Notepad++ and check what character is in the config.xml file (<Find name="&#x043C;&#x0430;&#x043A;&#x0440;&#x043E;&#x0441;" />). We just have HTML entities, not plain text. I don't know why TinyXML writes it like that, but if we consider it a bug/defect, it should rather be considered separately.

So for only shortcut.xml has such bug. Please create a new issue about unreadability of shortcut.xml while there's non-English character, and assign to me. I will see what I can do about it.

@Yaron10 
> EditPad displays them in "readable characters" if the encoding is set to:

I think the better solution is make tinyxml write it correctly, instead of adding another system/mechanism over on it.


@rdipardo 
> As https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12670#issuecomment-1359902802:

    TinyXML is not being developed anymore.

We got tinyxml source code so I'm gonna try to make it work.

donho:
Now shortcut.xml can be saved as human readable Unicode attribute value.
Could you guy check it please?

alankilborn:
>Could you guy check it please?

With that commit, my complaint [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/96b56f61f90d8b3f27d8fab804676a8ddcbd4a22#commitcomment-108781009) is resolved, and everything is human readable.  Thanks, @donho !

Yaron10:
@donho,

> So for only shortcut.xml has such bug.

As @ArkadiuszMichalski has mentioned the problem was NOT unique to `shortcut.xml`.

> Yes, but it's not limited to just this file, but to all .xml. Try searching for макрос, close Notepad++ and check what character is in the config.xml file (). We just have HTML entities, not plain text.

Anyway, with https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13538/commits/9289bb2d4ae616e9042b40a883abaf81bcc32a9b the issue in `shortcut.xml` is solved. 👍 

"One small step for `tinyxmlA.cpp`, one giant leap for Notepad++." :)
Thank you for your work.

The problem in `config.xml` still exists.
Search-terms and file-names are still not "human readable".

Search for `שלום`.

תמונה

***

I may not be by my PC tomorrow.

donho:
> As @ArkadiuszMichalski has mentioned the problem was NOT unique to `shortcut.xml`.
> 
> > Yes, but it's not limited to just this file, but to all .xml. Try searching for макрос, close Notepad++ and check what character is in the config.xml file (). We just have HTML entities, not plain text.

Please check the latest commit which fixes this issue.

Yaron10:
@donho,

It's not consistent.

Closing/restarting NPP, I get different results both in `config.xml` and in `Find what`.

תמונה

First - correct.
Second - wrong.

תמונה

Good night.

donho:
Thank you @Yaron10 for your test.
I have checked in code source and there's no trivial way to make it right, so the latest commit is reverted.

ArkadiuszMichalski:
@donho 
>After save/change macro's name we lost some default comments in [shortcuts.xml ](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/shortcuts.xml). Only the first one is preserved.

>>Indeed. But it's not a regression. The way to remove whole node and write the new one makes the comments in the old nodes are removed together. It's a no-humful bug though.

Maybe add some function that will copy all of node's comments before deleting them and then insert them at the beginning when we restoring  a node (the order of occurrence can be memorized or omitted)? Thanks to this, the default ones would be remembered, but also all ours that we will add later (which is also rather expected). From what I can see this could cover `InternalCommands`, `Macros`, `UserDefinedCommands`, `PluginCommands`, `ScintillaKeys`.

Yaron10:
@donho,

> I have checked in code source and there's no trivial way to make it right, so the latest commit is reverted.

👍 
Thank you for looking into it and trying. 
It's surely not a bug anyway.

alankilborn:
Is this the end? :-(

rddim:
There is already opened issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13222

PR #13562, opened 2023-04-21
[xml] Update galician.xml to v 8.5.3

P0rsche-911:
New revision and update.
There seems to be a lot of changes but they are to match galician.xml to english.xml.

Changed ( = ) to (=) in the whole file.

Please, accept. It will make future revisions much easier for me.

Nueva revisión y actualización.
Parece que hay muchos cambios pero son por igualar galician.xml a english.xml.

Cambiado ( = ) por (=) en todo el archivo.

Pido por favor que se acepte. Me facilitará mucho revisiones futuras.

### DIFF
--- a/PowerEditor/installer/nativeLang/galician.xml
+++ b/PowerEditor/installer/nativeLang/galician.xml
@@ -3,424 +3,425 @@
 The comments are here for explanation, it's not necessary to translate them.
 -->
 <NotepadPlus>
-    <Native-Langue name = "Galego" filename="galician.xml" version="8.5.1">
+    <Native-Langue name="Galego" filename="galician.xml" version="8.5.3">
         <Menu>
             <Main>
                 <!-- Main Menu Entries -->
                 <Entries>
-                    <Item menuId = "file" name = "&amp;Arquivo"/>
-                    <Item menuId = "edit" name = "&amp;Editar"/>
-                    <Item menuId = "search" name = "&amp;Buscar"/>
-                    <Item menuId = "view" name = "&amp;Vista"/>
-                    <Item menuId = "encoding" name = "&amp;Codificación"/>
-                    <Item menuId = "language" name = "&amp;Linguaxe"/>
-                    <Item menuId = "settings" name = "C&amp;onfiguración"/>
-                    <Item menuId = "tools" name = "&amp;Ferramentas"/>
-                    <Item menuId = "macro" name = "&amp;Macro"/>
-                    <Item menuId = "run" name = "E&amp;xecutar"/>
-                    <Item menuId = "Plugins" name = "Complemen&amp;tos"/>
-                    <Item menuId = "Window" name = "Pesta&amp;nas"/>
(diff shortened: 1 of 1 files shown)


pryrt:
@donho, 

I know that in the past, you have [rejected a similar modification](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13103#issuecomment-1427029772) from @P0rsche-911.  

The reason the submission has the global change to the spaces is because `galician.xml` has a bunch of space-equal-space (" = ") throughout the XML.  The official English translation, and some of the more "common" languages like French and German and Spanish have no " = " attributes, or only a handful; many of the languages have under a couple hundred of them; but the  galician.xml and sardinian.xml each have more than 1000 instances of the " = " between the attribute name and its value.  This makes it difficult when @P0rsche-911 tries to compare the galician.xml with the spanish.xml or english.xml; and that, in turn, makes it much harder for @P0rsche-911 to do the translation updates to galician.xml.

I know you don't like "big" changes like that.... but somehow it got messed up compared to other languages, and it needs to get fixed to make it easier for him to continue maintaining the translation.  If doing it as a PR like he's tried multiple times will not be acceptable to you, how do you propose that it get fixed?  Is it possible that for one time, you could use the ComparePlus plugin, and do a compare between the current version of the file and @P0rsche-911's PR version with the "Ignore Spaces" option turned on -- that way, you can see which changes are real translation changes, and which changes are just fixing `name = "blah"` to `name="blah"`.  Or if that's not acceptable, could @P0rsche-911 split it into two separate PR -- one PR that fixes all the " = " fixes, and then later come back and add the v8.5.3 changes in a separate PR after the spaces are fixed?

alankilborn:
>split it into two separate PR -- one PR that fixes all the " = " fixes, and then later come back and add the v8.5.3 changes in a separate PR after the spaces are fixed?

This is a good idea.
Break changes into multiple types.
I see also that a case change was made for `subMenuId` -- maybe that could be another type of change to separate out.

P0rsche-911:
Thank you very much for your help.

From your comments, I don't know if I will be able to do what you suggest.

Keep in mind that I am inexperienced and can easily make mistakes.

I am waiting for a resolution.

More info: https://community.notepad-plus-plus.org/topic/24363/incorrect-translation-galician-xml

Os agradezco enormemente vuestra ayuda.

Por lo que comentáis, no sé si seré capaz de hacer lo que sugerís.

Tened en cuenta que soy inexperto y me puedo equivocar fácilmente.

Quedo a la espera de una resolución.

chcg:
From e.g. XML Tools plugin -> Pretty Print could be used to get a more consistent formatting of all the xml translation files currently used. Using tabs instead of spaces as indent we would also save some KB filesizes.
There are a couple of options to do this also via commandline: https://stackoverflow.com/questions/16090869/how-to-pretty-print-xml-from-the-command-line

donho:
Thank you guys for your suggestions.
I will standardize the format of translation and make rules in english.xml.

Issue #4526, opened 2018-05-25
Deleting a Macro doesn't remove it from the Run Macro multiple times dialog until restart.


Peter-AVID:
### Description of the Issue

If you have macros saved, and delete one, the "Run macro multiple times" dialog box still contains the macro name in the dropdown/combo box.  This causes the index of any macro "after" that to be off by 1.
If you restart notepad-plus-plus, this is corrected.

### Steps to Reproduce the Issue

1. Save 3 macros.
2. Delete the 1st macro.
3. Open the "Run a Macro Multiple Times Dialog. (The deleted macro will still be there)
4. Run the "Second" (should now be first but the bug leaves it there) Macro.  This will actually run the third macro because the combo box says run macro index 1.

### Expected Behavior

Deleted macros not to show up in the dialog box.

### Debug Information

Notepad++ v7.5.6   (32-bit)
Build time : Mar 19 2018 - 00:26:59
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

PR #13563, opened 2023-04-21
Fix regression Run Macro Multiple Times cmd

### LINKED ISSUE #13561: [Regression] Run Macro Multiple Times: "number of times" is ignored
**STR:**
Save a simple macro (e.g. press Enter once).
Run Macro Multiple Times: insert `4`.

**Result:**
The macro is run once.

***
Current master: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0457a112c76ddc883a99074cf459a970741cf5b6.

ozone10:
not changing times immediately from edit control

fix #13561

### DIFF
--- a/PowerEditor/src/WinControls/shortcut/RunMacroDlg.cpp
+++ b/PowerEditor/src/WinControls/shortcut/RunMacroDlg.cpp
@@ -130,10 +130,12 @@ intptr_t CALLBACK RunMacroDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM l
 					return TRUE;
 
 				case IDOK:
+				{
 					if (::SendDlgItemMessage(_hSelf, IDC_MACRO_COMBO, CB_GETCOUNT, 0, 0) > 0)
 						::SendMessage(_hParent, WM_MACRODLGRUNMACRO, 0, 0);
 
 					return TRUE;
+				}
 
 				default:
 				{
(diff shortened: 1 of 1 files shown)


ozone10:
@Yaron10 
because of fix for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4526

Yaron10:
@ozone10,

Thank you. 👍

PowerEditor/src/WinControls/shortcut/RunMacroDlg.cpp
```diff
@@ -130,17 +130,62 @@ intptr_t CALLBACK RunMacroDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM l
...
+										::SetDlgItemInt(_hSelf, IDC_M_RUN_TIMES, 1, FALSE);
+										return TRUE;
+									}
+
+									return FALSE;
+								}
+
+								case EN_CHANGE:
+								{
+									_times = std::max<int>(::GetDlgItemInt(_hSelf, IDC_M_RUN_TIMES, nullptr, FALSE), 1);
```
Yaron10, line 172: @ozone10,

Thank you for this improvement.

Can't `_times =` be executed in `case EN_KILLFOCUS:`?

ozone10: @Yaron10 
when you change value and press "enter", focus is not changed to different control

Yaron10: @ozone10,

Very good! 👍 

And you don't want to implement `case EN_CHANGE:` only.
We've discussed this option before. :)
It's actually implemented somewhere in the Preference dialog.

Yaron10: > It's actually implemented somewhere in the Preference dialog.

In Backup -> Backup in every --- seconds.

Yaron10: Enabling/disabling the Run button on `case EN_CHANG` might be a good option.

ozone10: @Yaron10 
> Enabling/disabling the Run button on `case EN_CHANGE:` might be a good option.

this will make it more complex, 
e.g. Run button disabled, switch to Run until the EOF or click on other control, 
now I need to handle this in `case EN_KILLFOCUS:`

> In Backup -> Backup in every --- seconds.

I don't think I've touched code for this.

Yaron10:
@ozone10,

> this will make it more complex...

OK. It's easy to suggest. :)

> I don't think I've touched code for this.

I was just generally referring to that option.

Thanks again.

Issue #13456, opened 2023-03-31
Alphabetical text sorting does not work very well, especially for not-English text (but not only)


grzegorj:
### Description of the Issue
The expected behaviour of the sort function is to operate correctly not only for English. Unfortunately, it is not the case.

### Steps to Reproduce the Issue
Open a new text file and write, one word in a line, ex.
laka
łąka
mąka
mika

(note EOL = the end of the line at the end of the text)
These are Polish words, and the lines have already been sorted correctly by the alphabetical order. Nevertheless, try to sort them in Notepad++.

### Expected Behavior
The function “Sort” should not change anything in this text

### Actual Behavior
The final EOL is transferred at the beginning of the file, and the words are not correctly sorted any longer (the correct alphabetical order is totally ignored). You receive:

laka
mika
mąka
łąka
(note the empty line at the beginning; there is no EOL after the word “łąka”)
 
The bug is checked in Notepad++ v. 8.5
BTW, the correct alphabetical order in Polish is aąbcćdeęfghijklłmnńoópqrsśtuvwxyzźż. Each letter is treated separately, not as a variant with a diacritic.

grzegorj:
Indeed, there are two issues (but within the same scope). For me, ignoring the correct alphabetical order is much more important than the issue with EOL (which I have noticed by the way). Either Notepad++ is “totally English” (and then it can sort only in accordance to ASCII codes), or it is international. But if the other alternative is correct, the lack of a possibility to sort text lines correctly is a very serious deficiency of the program. Of the programs I know, only MS Word can do it correctly, using the system locale (which is present under Windows) instead of low-level sorting based on ASCII codes. It would be REALLY nice if also Notepad++ had a similar option. And it would not be so English-biased any longer as it is now, despite existing superficial localizations (of the menu etc.). Really. The thing is worth fixing. Even if it may not be very simple.

rdipardo:
> Of the programs I know, only MS Word can do it correctly, using the system locale (which is present under Windows) instead of low-level sorting based on ASCII codes. It would be REALLY nice if also Notepad++ had a similar option.

Apples to oranges.

Notepad++ is a *source code* editor. Source code is almost always (for [historical reasons](https://pl.wikipedia.org/wiki/ASCII#Historia)) a plain text file of ASCII keywords and punctuation symbols.

MS Word is a *business document* editor. Businesses have offices around the world, so naturally their software can intelligently edit internationalized text.

[...]

PR #13571, opened 2023-04-23
Add culture-sensitive lexicographic ignorecase sort, partially address issue 13456

molsonkiko:
Partially address issue #13456 (the other part is asking for the terminal EOL a file not to be sorted to the top by an ascending sort, which is addressed in my other PR, #13498 )

The current ignorecase sort does not work well for non-ASCII text because it sorts characters like ß after all ASCII characters even if they are lexicographically between two ASCII characters. For example, "ß" is generally equivalent to "ss" in a case-insensitive regime.
__Add a button to the Edit main menu bar for locale-sensitive sorts: Sort Lines Lex. Asc. Culture-sensitively ignoring case and Sort Lines Lex. Desc. Culture-sensitively ignoring case.__

Consider the text
```
baßk
bazk
oyster
öyster
spä 2
spä 11
Spb
täco
tone
```
which is already in the most sensible order both lexicographically and numerically.

The existing case-insensitive ascending sort messes with this order by sorting it to
```
bazk
baßk
oyster
Spb
spä 11
spä 2
tone
täco
öyster
```
so all the German letters are in the wrong place and the numbers are in the wrong order.

My new *ascending* sort preserves the correct original order, and my new *descending* sort sorts it to
```
tone
täco
Spb
spä 11
spä 2
öyster
oyster
bazk
baßk
```

__Performance:__ for a very large file (770,000 lines) it is *at least 10 times slower* than the existing lexicographic ignorecase sort.


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -905,6 +905,37 @@ int OrdinalIgnoreCaseCompareStrings(LPCTSTR sz1, LPCTSTR sz2)
 	}
 }
 
+std::string CultureSensitiveIgnoreCaseMapString(LPCTSTR s) {
+	int len = wcslen(s);
```
rdipardo, line 909: The `*len` family of functions all return `size_t` (a dynamic type).
A 64-bit `size_t` is `typedef`'d as `unsigned long long`: https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/46860742/job/c0du01opkdnfbli1/messages

molsonkiko:
If you try sorting lines in Debug Mode, you may hit a failed assertion in `ScintillaEditView.cpp`. __This is not my fault!__ I tested and confirmed that this error is present in my master branch, not just in this branch.

My other PR #13498 should fix this problem.

molsonkiko:
Any thoughts from core devs on this new feature?
I just checked it again and it seems that it continues to work as advertised in my original comment.

donho:
@molsonkiko 
The result of **Sort Lines Lex. Asc. Culture-sensitively ignoring case** is not correct IMO.

Using the following example for testing:
```
<dfn id=map-get-t
<dfn id=map-get><
```

* The result after running **Sort Lines Lex. Asc. Culture-sensitively ignoring case** is:
```
<dfn id=map-get><
<dfn id=map-get-t
```

The above result is wrong because `<` > `-` (0x3E > 0x2D)

Any idea about such bug?

molsonkiko:
> The above result is wrong because `<` > `-` (0x3E > 0x2D)

Not sure I would call it *wrong* per se. I agree that the sort order of punctuation characters is strange, but the *whole point of this sort* is that it doesn't sort characters based on the numeric value of their Unicode codepoint.

I suppose I could special-case punctuation characters in the sort, but that might be a horrendously complicated and error-prone process because it would involve hacking the internals of `LCMapStringEx`, which is the string comparison method I used.

Also, for what it's worth, this is the same sort order that .NET uses. Using my [JsonTools plugin](https://github.com/molsonkiko/JsonToolsNppPlugin), I sorted this array:
```
[
    "<dfn id=map-get-t",
    "<dfn id=map-get><",
]
```
and the result of sorting was
```
[
    "<dfn id=map-get><",
    "<dfn id=map-get-t"
]
```

molsonkiko:
So I looked at the flags that are available for `LCMapStringEx`. It appears that the only flags that change the ordering of punctuation are `NORM_IGNORESYMBOLS`, which is *definitely not what we want*, and __`SORT_STRINGSORT`, which changes the order of punctuation but does not put them in proper ordinal order of the Unicode codepoints.__

So while I agree that it's a shame that the ordering of punctuation is unusual with this sort, __I think it's a price worth paying to make the sort order better in so many other ways.__

molsonkiko:
__Correct order__
```
a&b (a&b should come first)
a*b (all these a op b are in correct binary order)
a+b
a-b
a/b
a<b
a>b
a^b
a|b (a|b should come last)
```
__LCMapStringEx *(the PR as it currently is)*__
```
a&b (a&b should come first)
a*b (all these a op b are in correct binary order)
a/b
a^b
a|b (a|b should come last)
a+b
a<b
a>b
a-b
```
__LCMapStringEx *with the SORT_STRINGSORT flag*__
```
a-b
a&b (a&b should come first)
a*b (all these a op b are in correct binary order)
a/b
a^b
a|b (a|b should come last)
a+b
a<b
a>b
```

donho:
> Not sure I would call it wrong per se. I agree that the sort order of punctuation characters is strange, but the whole point of this sort is that it doesn't sort characters based on the numeric value of their Unicode codepoint.

The characters with diacritics can have also punctuation characters in the same line, therefore it's invalid argument to me.
**Sort Lines Lex. Asc. Culture-sensitively ignoring case** function, when it sorts an non unicode text, should just have the same sorting result comparing with the one of **Sort Lines Lex. Ascending ignoring case**. Since there's no Unicode character at all.

> So while I agree that it's a shame that the ordering of punctuation is unusual with this sort, I think it's a price worth paying to make the sort order better in so many other ways.

I wouldn't call it "price" but "bug". 
It's better to have a missing feature than to add a feature which is buggy. 
If you want to keep working on this bug, you're welcome to do it, but the PR is not ready to be merged with its current state.

molsonkiko:
@donho 
I respectfully disagree with your above reasoning.

The whole reason to add another sort method is that sorting things by the order of their Unicode codepoints is unsatisfactory in many situations.

The issue that prompted this PR pointed out that __NPP's case-insensitive sort is *inconsistent with the sort order of numerous other Windows applications*, including PowerShell's `sort` cmdlet and the Microsoft Word line-sorting tool.__ Thus, the goal of *this specific sort* (as opposed to the existing sorts, with which I have no issue) should arguably be to *be consistent with other Windows applications. *

When I use Microsoft word or the PowerShell `Sort-Object` cmdlet to sort 
```
a*b (all these a op b are in correct binary order)
a+b
a-b
a/b
a<b
a>b
a^b
a|b (a|b should come last)
```
I get 
```
a*b (all these a op b are in correct binary order)
a/b
a^b
a|b (a|b should come last)
a+b
a<b
a>b
a-b
```
__which is the same order you get from the PR in its current state.__

That said, the PR in its current state doesn't have *quite* the same sort order as Microsoft Word, because my PR sorts digits as numbers and Microsoft Word does not. Based on that reasoning, it might make sense to revert my PR and remove the ability to sort digits as numbers.

molsonkiko:
I guess my above argument wasn't persuasive enough. That's too bad, but I'm still interested in trying to salvage this PR, so long as I have a clear understanding of what I must do to be accepted.

@donho, what kind of punctuation must be sorted in order of its Unicode codepoints? Just ASCII punctuation? Or should I also order non-ASCII punctuation in order of the Unicode codepoints?

donho:
@molsonkiko 
> I guess my above argument wasn't persuasive enough.

I do understand totally your argument as dev, and I do understand also LCMapStringEx treats some symbols differently. 
My point is that a new feature's behaviour should not make users confused.

> but I'm still interested in trying to salvage this PR, so long as I have a clear understanding of what I must do to be accepted.
> what kind of punctuation must be sorted in order of its Unicode codepoints? Just ASCII punctuation? Or should I also order non-ASCII punctuation in order of the Unicode codepoints?

A user doesn't know and care it's Unicode or not. So the criteria to make PR accepted is "Sort Lines Lex. Asc. Culture-sensitively ignoring case" should have the same result of "Sort Lines Lex. Ascending ignoring case" while sorting a document which doesn't contain the character with [diacritic](https://en.wikipedia.org/wiki/Diacritic).

donho:
@molsonkiko 
Please let me know if you want to keep working on it. 
If yes, I'll reopen it.

molsonkiko:
The problem is that diacritics aren't the only consideration in a
case-insensitive sort. Also, some chars like German Eszett expand to
multiple characters when ignoring case. Any attempt to combine
LCMapStringEx with ordinal sorting is bound to be a garbage fire.

PR #13573, opened 2023-04-23
Installer should only write to the registry during the install phase

sredna:
.onInit should not write to the registry. If the user cancels the installer, no traces should be left behind.

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -177,16 +177,12 @@ updaterDone:
 		
 	; look for previously selected language
 	ClearErrors
-	Var /GLOBAL tempLng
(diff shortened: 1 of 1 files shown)


donho:
@sredna 
Logically, your code makes sense and looks better. However, it doesn't work after my testing of this PR:
It seems `WriteRegStr HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage' '$Language'` doesn't take effect - after changing the language and finishing the installation, the next launch of the same installer takes always the same language before changing.

I suggest you always test your PR locally before submitting.

sredna:
@donho I did test it. I tested it in a 32-bit VM which might be why it worked for me.

The installer is a bit schizophrenic about where it writes `HKLM "Software\${APPNAME}"` in the registry, it wants the path in the 64-bit view and the language in the 32-bit view.

I fixed the PR and added a note as to why it needs to be in the 32-bit view.

Issue #13437, opened 2023-03-28
Case Convert - Proper - Doesn't work properly with apostrophes in extended (unicode) text


au-JC:
### Description of the Issue
NotePad++ 8.5.x - Case Conversion - Proper (and Proper (Blend)) doesn't work properly with apostrophes in extended (unicode) text.

For example,
NotePad++ 8.5.0 – There’s An Issue (Case Convert - Proper)
Becomes:
Notepad++ 8.5.0 – There’S An Issue (Case Convert - Proper)

### Steps to Reproduce the Issue
1. Copy and Paste into NotePad++ some text, say from a website or Word document, containing an 'smart apostrophe'.
2. Select the text.
3. Select menu Edit | Convert Case to | Proper Case / Proper Case (Blend) [or use short-cut Alt+U]
4. Text will be converted similar to the above example.

### Expected Behavior
NotePad++ 8.5.0 – There’s An Issue (Case Convert - Proper)
Should become:
Notepad++ 8.5.0 – There’s An Issue (Case Convert - Proper)

### Actual Behavior
For example,
NotePad++ 8.5.0 – There’s An Issue (Case Convert - Proper)
Becomes:
Notepad++ 8.5.0 – There’S An Issue (Case Convert - Proper)

### Debug Information
NotePad++ 8.5.0

![SS_NotePad851_Issue-CaseConvert-Proper_20230329](https://user-images.githubusercontent.com/129229600/228392689-4af94070-03ec-430e-8354-a38ffd937299.jpg)

PR #13584, opened 2023-04-24
Fixed proper case conversion logic for single quotes and smart single quotes

cynthplusplus:
After applying proper case conversion, we expect uppercase for the first letter of selected words and lowercase for the remaining letters. Currently, two modes are supported, Force and Blend. Force changes all first letters to uppercase and changes all following letters to lowercase. Blend retains uppercase characters within the word.

However, apostrophes [typically do not split words](https://en.wikipedia.org/wiki/Apostrophe), necessitating special logic. 

The current logic only works for straight single quotes. It lowers the case of all alphabetic characters following a single quote, except when a single quote is the first character . This results in some undesirable behaviour as observed in issue #13437.

Specifically:
1) Smart single quotes (U+2018 and U+2019) are not supported.
2) Blend doesn't retain uppercase after an apostrophe. For example, we expect O'Connor -> O'Connor, but we get O'Connor -> O'connor.
3) Force doesn't lower the case after a number. For example, we expect 9'S -> 9's, but we get 9'S ->9'S.

This PR fixes these problems. Tested manually with the following cases:
'https://github.com/cynthplusplus/eecs481hw6/blob/main/case_convert_tests.txt


cynthplusplus:
Also, I changed TITLECASE_BLEND / TITLECASE_FORCE to PROPERCASE_BLEND / PROPERCASE_FORCE for consistency.

Yaron10:
@cynthplusplus,

👍 
Thank you for your work.

I don't use this feature frequently, and I've only tested "O'Connor", "9's" and "There’s An Issue".

donho:

It's OK for this time.

ref:

PR #13587, opened 2023-04-25
fix infinite loop in running macro to EOF

molsonkiko:
Fix issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13342
To test fix, try making file
```
f
f
```
and recording macro of find/replace form searching for `f` *with wrap-around enabled*, then running until EOF.
To confirm that this PR does not cause the loop to always end after one iteration, try adding more lines of f's, and `run until end of file` a few times. You should find that it will run until after the first wrap-around, just as one would expect.

This does not break any existing behavior, including:
- macros where the cursor moves towards EOF but line num doesn't change (those already stopped after one iteration)
- macros where line(s) are deleted with every iteration (even if line number doesn't change, they run until file empty)
- macros where the line number increases with each iteration
- macros where the cursor advances up or down with each iteration but would eventually stop anyway (those end at the correct time)

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1471,6 +1471,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 				int indexMacro = _runMacroDlg.getMacro2Exec();
 				intptr_t deltaLastLine = 0;
 				intptr_t deltaCurrLine = 0;
+				bool cursorMovedUp = false;
 
 				Macro m = _macro;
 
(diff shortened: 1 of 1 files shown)


molsonkiko:
Is anyone going to take action on this PR? I'd really like to see some action on this.

Seriously, I keep trying to make find/replace macros and using the "Run until end of file" option, only to have to shut down Notepad++ because of the infinite loop. This is an extremely irritating bug, and my PR is pretty simple.

My steps for checking if the PR addresses the issue are also simple. Just create a file with 8 lines of `f` and try running until end of file from various different points at the file until you've convinced yourself that (a) it doesn't cause an infinite loop and (b) it doesn't cause `run until end of file` to stop after one or two iterations.

donho:
Sorry @molsonkiko for the late merging.

molsonkiko:
@donho 
No worries. Thank you!

Issue #13574, opened 2023-04-23
Allow plugins to apply darkmode colors to dialogs using method call or API call?


BdR76:
### Description of the Issue
Is it possible to somehow call the [autoSubclassAndThemeChildControls](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppDarkMode.cpp#L1940) function from within a plug-in? afaik there isn't a API call for this. That way the plugins can use the default Notepad++ function to apply darkmode colors uniformly to all dialogs.

Currently most plugins display any dialogs in normal/lightmode colors even when Notepad++ is set to darkmode. I know there is `DWS_USEOWNDARKMODE` but for simple Settings/About forms when there are no special controls or anything, it would imho be better to just let Notepad++ change these to darkmode.

### Steps to Reproduce the Issue
1) Set Notepad++ to darkmode `Settings > Preferences > Dark Mode`
2) See "About" dialog in (almost?) all plugins, the CSV Query parse settings, CSV Lint setting dialogs etc.

![csvquery_example_dialog](https://user-images.githubusercontent.com/7793956/233845357-8ebfe4bf-8173-4cb9-9402-ffd8575e936d.png)

### Expected Behavior
Be able to call the function [autoSubclassAndThemeChildControls](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppDarkMode.cpp#L1940) to stylize any extra dialogs (settings, parameters, about etc) with the default darkmode colors, see mockup below.

![notepad_plugin_darkmode](https://user-images.githubusercontent.com/7793956/233845379-a6732b3a-51d2-4e0f-a0b2-281f07140761.png)

### Actual Behavior
When Notepad++ is set to darkmode, some plugin dialogs will display using the normal/lightmode colors.

### Debug Information
Tagging @ozone10 for his knowledge about darkmode

rdipardo:
To [quote the author of #11711](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11711#issuecomment-1146860165):

> there might be visual glitches if plugin is written in C#, as it use WinForms for GUI, it also does not affect plugins which use their own custom control [...]

Even assuming N++ could predict every possible GUI component a plugin may have, the implementation would be in C++. You still need some kind of bridge on the .NET side to make it usable.

[...]

Issue #13572, opened 2023-04-23
Generic darkmode for plug-in windows is not applied to all controls (groupbox, panels, labels etc)


BdR76:
### Description of the Issue
The generic darkmode for plug-in dock panels was added, see [PR 11711](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11711) but it doesn't apply darkmode colors to all wincontrols, for example groupbox, panel and labels. See screenshot below, the darkmode colors are not appied to the textbox within the panel.

![notepad_darkmode_window](https://user-images.githubusercontent.com/7793956/233844283-c60a325b-3714-4a8b-ac77-f98823750011.png)

### Steps to Reproduce the Issue
See some plug-ins with a docking panel, for example CSV Lint, NppMarkdownPanel, CSV Query (only the buttons), Json Tools (Tree viewer)

### Expected Behavior
The main docking window should use darkmode color for all controls/labels/textboxes etc.

### Actual Behavior
The panel shows some widgets/controls in light mode colors

### Debug Information

I think it has to do with function [autoSubclassAndThemeChildControls](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppDarkMode.cpp#L1940) which doesn't call `EnumChildWindows` recursively for container-type controls.

If I understand correctly, the container controls like Panels, GroupBoxes, SplitContainer etc are called [REBARCLASSNAME](https://learn.microsoft.com/en-us/windows/win32/controls/common-control-window-classes) controls

Tagging @ozone10 in this issue, because of the required knowledge of darkmode support, (was suggested in [thread](https://community.notepad-plus-plus.org/topic/24403/darkmode-doesn-t-stylize-all-plugin-controls-on-docked-window-groupbox-panel-labels-etc))

ozone10:
@BdR76 
Are those C# plugins?
IIRC C# are using some containers/controls with its classname which are not same as standard win32 classname.
They may behave same but subclassing may/may not work.

For edit, listbox, static text controls those are affected by `WM_CTLCOLOR*` messages processed by their parents.
So the edit control in container won't be affected since its parent - container does not handle these messages.

Notepad++ only process `WM_CTLCOLOR*` messages for main window.

I won't be able soon to investigate this, my pc recently broke.

ozone10:
@BdR76 
I did some quick testing, unfortunately there is no easy way to make generic dark mode work with C# plugins.

[...]

PR #13596, opened 2023-04-26
Add message NPPM_DARKMODESUBCLASSANDTHEME

ozone10:
to allow plugin authors to use generic dark mode

Needs some testing and feedback from plugin authors.
Unfortunately not for C# plugins.

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13574
related https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13572

### DIFF
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -45,7 +45,7 @@ CPP_DEFINE := UNICODE _UNICODE OEMRESOURCE NOMINMAX _WIN32_WINNT=_WIN32_WINNT_VI
 LD := $(CXX)
 LDFLAGS := -municode -mwindows
 LD_PATH :=
-LD_LINK := comctl32 crypt32 dbghelp ole32 sensapi shlwapi uuid uxtheme version wininet wintrust
+LD_LINK := comctl32 crypt32 dbghelp ole32 sensapi shlwapi uuid uxtheme version wininet wintrust dwmapi
 LD_LINK += $(patsubst lib%.a,%,$(SCINTILLA_TARGET)) $(patsubst lib%.a,%,$(LEXILLA_TARGET)) imm32 msimg32 ole32 oleaut32
 SUBMAKEFLAGS := -O --no-print-directory
 
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -558,65 +558,23 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	//
 	// Might not work properly in C# plugins.
 	//
-	// Returns succesful combinations from dmfRequiredMask.
+	// Returns succesful combinations of flags.
 	//
 
 	namespace NppDarkMode
 	{
-		// Used on parent of edit, listbox, static text, treeview, listview and toolbar controls.
-		// Should be used only one time on parent control after its creation
-		// even when starting in light mode.
-		// e.g. in WM_INITDIALOG, in WM_CREATE or after CreateWindow.
-		constexpr ULONG dmfSubclassParent =     0x00000001UL;
-		// Should be used only one time on main control/window after initializations of all its children controls
-		// even when starting in light mode.
-		// Will also use dmfSetThemeChildren flag.
-		// e.g. in WM_INITDIALOG, in WM_CREATE or after CreateWindow.
-		constexpr ULONG dmfSubclassChildren =   0x00000002UL;
-		// Will apply theme on buttons with style:
-		// BS_PUSHLIKE, BS_PUSHBUTTON, BS_DEFPUSHBUTTON, BS_SPLITBUTTON or BS_DEFSPLITBUTTON.
-		// Will apply theme for scrollbars on edit, listbox and rich edit controls.
-		// Will apply theme for tooltips on listview, treeview and toolbar buttons.
-		// Should be handled after controls initializations and in NPPN_DARKMODECHANGED.
-		// Requires at least Windows 10 to work properly.
-		constexpr ULONG dmfSetThemeChildren =   0x00000004UL;
-		// Set dark title bar.
-		// Should be handled after controls initializations and in NPPN_DARKMODECHANGED.
-		// Requires at least Windows 10 and WS_CAPTION style to work properly.
-		constexpr ULONG dmfSetTitleBar =        0x00000008UL;
-		// Will apply dark explorer theme.
-		// Used mainly for scrollbars and tooltips not handled with dmfSetThemeChildren.
-		// Might also change style for other elements.
-		// Should be handled after controls initializations and in NPPN_DARKMODECHANGED.
-		// Requires at least Windows 10 to work properly.
-		constexpr ULONG dmfSetThemeDirectly =   0x00000010UL;
-
 		// Standard flags for main parent after its children are initialized.
-		// Check individual flags for more information.
-		// 0x000000BUL
-		constexpr ULONG dmfAfterInitParent = dmfSubclassParent | dmfSubclassChildren | dmfSetTitleBar;
+		constexpr ULONG dmfInit =               0x0000000BUL;
+
 		// Standard flags for main parent usually used in NPPN_DARKMODECHANGED.
-		// Check individual flags for more information.
-		// 0x000000CUL
-		constexpr ULONG dmfHandleChangeParent = dmfSetThemeChildren | dmfSetTitleBar;
-
-		// At least one of these flags should be used.
-		// currently same as dmfAllMask
-		// used internally
-		// 0x0000001FUL
-		//constexpr ULONG dmfRequiredMask = dmfSubclassParent | dmfSubclassChildren | dmfSetThemeChildren | dmfSetTitleBar | dmfSetThemeDirectly;
-
-		// every flags
-		// currently not used
-		// 0x0000001FUL
-		//constexpr ULONG dmfAllMask = dmfSubclassParent | dmfSubclassChildren | dmfSetThemeChildren | dmfSetTitleBar | dmfSetThemeDirectly;
+		constexpr ULONG dmfHandleChange =       0x0000000CUL;
 	};
 
 	// Examples:
 	//
 	// - after controls initializations in WM_INITDIALOG, in WM_CREATE or after CreateWindow:
 	//
-	//auto success = static_cast<ULONG>(::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfAfterInitParent), reinterpret_cast<LPARAM>(mainHwnd)));
+	//auto success = static_cast<ULONG>(::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfInit), reinterpret_cast<LPARAM>(mainHwnd)));
 	//
 	// - handling dark mode change:
 	//
@@ -626,8 +584,8 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	//	{
 	//		case NPPN_DARKMODECHANGED:
 	//		{
-	//			::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(dmfHandleChangeParent), reinterpret_cast<LPARAM>(mainHwnd));
-	//			::SetWindowPos(mainHwnd, nullptr, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_FRAMECHANGED); // to redraw titlebar
+	//			::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(dmfHandleChange), reinterpret_cast<LPARAM>(mainHwnd));
+	//			::SetWindowPos(mainHwnd, nullptr, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_FRAMECHANGED); // to redraw titlebar and window
 	//			break;
 	//		}
 	//	}
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -574,7 +574,7 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 		// e.g. in WM_INITDIALOG, in WM_CREATE or after CreateWindow.
 		constexpr ULONG dmfSubclassChildren =   0x00000002UL;
 		// Will apply theme on buttons with style:
-		// BS_PUSHLIKE, BS_PUSHBUTTON, BS_DEFPUSHBUTTON, BS_SPLITBUTTON and BS_DEFSPLITBUTTON.
+		// BS_PUSHLIKE, BS_PUSHBUTTON, BS_DEFPUSHBUTTON, BS_SPLITBUTTON or BS_DEFSPLITBUTTON.
 		// Will apply theme for scrollbars on edit, listbox and rich edit controls.
 		// Will apply theme for tooltips on listview, treeview and toolbar buttons.
 		// Should be handled after controls initializations and in NPPN_DARKMODECHANGED.
@@ -592,11 +592,11 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 		constexpr ULONG dmfSetThemeDirectly =   0x00000010UL;
 
 		// Standard flags for main parent after its children are initialized.
-		// e.g. in WM_INITDIALOG, in WM_CREATE or after CreateWindow.
-		// Should be used only one time on main control/window after initializations of all its children controls.
+		// Check individual flags for more information.
 		// 0x000000BUL
 		constexpr ULONG dmfAfterInitParent = dmfSubclassParent | dmfSubclassChildren | dmfSetTitleBar;
 		// Standard flags for main parent usually used in NPPN_DARKMODECHANGED.
+		// Check individual flags for more information.
 		// 0x000000CUL
 		constexpr ULONG dmfHandleChangeParent = dmfSetThemeChildren | dmfSetTitleBar;
 
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -562,7 +562,8 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	//
 	// Might not work properly in C# plugins.
 	//
-	// Returns FALSE if parent subclass was not successful, hwnd == nullptr or dmFlags were invalid, otherwise TRUE.
+	// Returns FALSE if parent subclass was not successful, hwnd == nullptr or none of dmfSetParent and dmfSetChildren flags were set,
+	// otherwise TRUE.
 	//
 
 	namespace NppDarkMode
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -551,12 +551,8 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 
 	#define NPPM_DARKMODESUBCLASSANDTHEME (NPPMSG + 112)
 	// ULONG NPPM_DARKMODESUBCLASSANDTHEME(ULONG dmFlags, HWND hwnd)
-	// Subclass hwnd to add support for generic dark mode.
-	// When applying first time on main hwnd use "dmFlags = NppDarkMode::dmfAll".
-	// In case edit, listbox, static text, treeview, listview and toolbar controls don't have parent as main hwnd,
-	// call NPPM_DARKMODESUBCLASSANDTHEME on their parent and use "dmFlags = NppDarkMode::dmfSetParent".
-	// When handling dark mode change with NPPN_DARKMODECHANGED, use "dmFlags = NppDarkMode::dmfSetChildren".
-	// 
+	// Add support for generic dark mode.
+	//
 	// Docking panels don't need to call NPPM_DARKMODESUBCLASSANDTHEME for main hwnd.
 	// Subclassing is applied automatically unless DWS_USEOWNDARKMODE flag is used.
 	//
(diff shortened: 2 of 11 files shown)


ozone10:
@alankilborn 
could you test?

alankilborn:
I tested it in a very basic way and it seemed to work.  HOWEVER, I am not a plugin author and am just "fooling around" with the C++ plugin demo template (very simple code).  It would be much better if a real plugin author that currently sees dark mode deficiencies could test it.

Yaron10:
@shriprem,

Could you please test this PR?

Thank you.

shriprem:
@Yaron10,

Both my plugins have had their own independent Darkmode support, which I added pretty soon after NPP started offering it. For that, I created my own standalone classes (see: https://github.com/shriprem/NPP_Plugin_Darkmode ) which were basically code extracts from inside NPP's Darkmode implementation from that period.

So, my plugins are not the right candidates for evaluating this PR. Yet, I tested the build from this commit, and found no regressions due to overlapping concerns, if any, with my plugins' Darkmode rendering feature.

Yaron10:
@shriprem,

👍 
Thank you. I appreciate it.

I hope you're doing well.
Regards.

rdipardo:
@Yaron10, @shriprem

From what I read [here](https://github.com/npp-plugins/plugindemo/issues/4#issue-1684960322), the definitive test is a groupbox with at least one child control. The automatic subclassing currently misses the children.

rdipardo:
@ozone10

It looks like this PR actually addresses npp-plugins/plugindemo#4 (once the demo project integrates the new upstream API, of course).

#13572 and #13574 are .NET issues that can be resolved in the .NET plugin repo, e.g., kbilsted/NotepadPlusPlusPluginPack.Net#104

Yaron10:
@rdipardo,

:+1: 
Thank you.

ozone10:
> From what I read [here](https://github.com/npp-plugins/plugindemo/issues/4#issue-1684960322), the definitive test is a groupbox with at least one child control. The automatic subclassing currently misses the children.

Yes the generic dark mode does not take into account that edit, listbox and other controls which colors are handled by parent, don't have parent as docking container.

Implementing this is hard and will make the code very complex, there is also issue some with gcc builds (some win32 subclassing functions are not properly exported/defined?).

So generic dark mode add only basic support, as it is impossible to take into account every special situations.

If plugins use more complex controls, dark mode should be handled by its author.
This PR should make it easier for them.

ozone10:
Missing dark mode for title bar. I will add this later.

ozone10:
Result of just adding line
```c++
::SendMessage(_nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfMainParent), reinterpret_cast<LPARAM>(_hSelf));
```
image

Currently updown control is handled only for tab control with horizontal style.

Yaron10:
@ozone10,

👍 
Thank you.

Yaron10:
@ozone10,

> Result of just adding line

May I ask where this line should be added in the ComparePlus code?

Thank you.

ozone10:
@Yaron10 
> May I ask where this line should be added in the ComparePlus code?
> 
> Thank you.

I only did quick check for setting dlg 
https://github.com/pnedev/comparePlus/blob/64a9ee43bbda6755a0cdb784e6b1b3decdb0af6f/src/SettingsDlg/SettingsDialog.cpp#L50

In general this message should be used after `hwnd` creation, usually in `WM_INITDIALOG`, in `WM_CREATE` or after `CreateWindow` .


About dlg might require additional modification for url link text.
ColorPopup and NavBar's scrollbar can be dark too.

Also `NPPN_DARKMODECHANGED` need to be handled for dark mode change.

Yaron10:
@ozone10,

Thank you for the detailed reply. 👍 
I hope it should help @pnedev when he gets back to handling DarkMode.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -549,6 +549,71 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
...
+		constexpr ULONG dmfRequiredMask = dmfSetParent | dmfSetChildren | dmfSetTitleBar | dmfSetExplorerTheme; // 0x0000001BUL
+		// every flags
+		constexpr ULONG dmfAllMask = dmfSetParent | dmfSetChildren | dmfSubclassChildren | dmfSetTitleBar | dmfSetExplorerTheme; // 0x0000001FUL
+	};
+
+	// Examples:
+	//
+	// - first time:
+	//
+	//auto success = static_cast<ULONG>(::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfMainParent), reinterpret_cast<LPARAM>(mainHwnd)));
```
donho, line 600: Why do we need the 1st time?
Is it really necessary for this API?

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -549,6 +549,71 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
...
+	//
+	// - handling dark mode change:
+	//
+	//extern "C" __declspec(dllexport) void beNotified(SCNotification * notifyCode)
+	//{
+	//	switch (notifyCode->nmhdr.code)
+	//	{
+	//		case NPPN_DARKMODECHANGED:
+	//		{
+	//			constexpr auto flags = NppDarkMode::dmfSetChildren | NppDarkMode::dmfSetTitleBar;
```
donho, line 610: The settings of flags is quite confusing: in which case plugin authors should use which flag?

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
```diff
@@ -549,6 +549,71 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
...
+		// should be handled on first time and in NPPN_DARKMODECHANGED
+		// requires at least Windows 10
+		constexpr ULONG dmfSetExplorerTheme =   0x00000010UL;
+		// standard first time flags for main parent
+		constexpr ULONG dmfMainParent = dmfSetParent | dmfSetChildren | dmfSubclassChildren | dmfSetTitleBar; // 0x000000FUL
+		// at least one of these flag should be used
+		constexpr ULONG dmfRequiredMask = dmfSetParent | dmfSetChildren | dmfSetTitleBar | dmfSetExplorerTheme; // 0x0000001BUL
+		// every flags
+		constexpr ULONG dmfAllMask = dmfSetParent | dmfSetChildren | dmfSubclassChildren | dmfSetTitleBar | dmfSetExplorerTheme; // 0x0000001FUL
+	};
```
donho, line 594: All these flags are not explicit for people who are not familiar to dark mode.
Could we reduce some flags which won't be used, and make all the flag names simpler and more explicit, with more explanation and examples?

donho (changes requested):
Let's discuss the 3 points mentioned below. My goal is make API as simple/clear as possible. So every plugin dev can use it.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
ozone10: Mask flags are used mainly internally.
I am open to better name.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
ozone10: It's similar to way how Notepad++ apply dark mode, after controls are initialized they need to be subclassed.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
ozone10: Again same approach as Notepad++ dark mode change handling.

`dmfSetChildren` is required for mainly push buttons which use theme rather than subclassing.
 Title bars also don't use subclassing. If dialog has title bar it should use `dmfSetTitleBar`.

Maybe better explanation in comments or documentation should help.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
donho: @ozone10 
what are the flags which are used by plugins then?  
I suppose that they are dmfAllMask,  dmfRequiredMask  and dmfMainParent 
Could you confirm that?

donho: @ozone10 
What's the different between dmfSetParent & dmfSetChildren ?

ozone10: `dmfSetParent` is for edit, listbox, partially combobox and static text controls. They use `WM_CTLCOLOR*` messages, which are processed by their parents.
It might be better to rename it to `dmfSubclassParent`

`dmfSetChildren` currently sets theme for push buttons and some other controls, and allows subclass. I might change how it interacts with `dmfSubclassChildren` to make it easier for plugin's authors.

ozone10:
@donho 
I've changed flag names and added more explanation in comments.

PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
donho: > dmfSetParent is for edit, listbox, partially combobox and static text controls. They use WM_CTLCOLOR* messages, which are processed by their parents.
It might be better to rename it to dmfSubclassParent

So if a dlgHwnd contains a editHwnd, a comboHwnd and a buttonHwnd as children, how can plugin author call `NPPM_DARKMODESUBCLASSANDTHEME` to set dialog in dark mode?

ozone10: they should use 
```C++
::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfAfterInitParent), reinterpret_cast<LPARAM>(dlgHwnd));
```

donho: So why do we need other flags?
What's "Theme" exactly BTW?

ozone10: `dmfAfterInitParent` is combination of `dmfSubclassParent | dmfSubclassChildren | dmfSetTitleBar` flags, but plugin authors might want to handle some controls differently.

Maybe they want only dark title bar and handle other controls themself.

If editHwnd is not child of dlgHwnd but buttonHwnd, then
```c++
::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfSubclassParent), reinterpret_cast<LPARAM>(buttonHwnd));
```
should be used.

Also `dmfSubclassParent` should be used only once for required control (parent) and `dmfSubclassChildren` only once for main hwnd.


For theme, Notepad++ uses wrapper for https://learn.microsoft.com/en-us/windows/win32/api/uxtheme/nf-uxtheme-setwindowtheme. And controls need to apply new theme each time dark mode has changed, or they will remain in previous mode and have wrong colors.

Same for title bar.

donho: > Maybe they want only dark title bar and handle other controls themself.

That doesn't make sense. ppl handle whole things themself, or they use API completely to do all stuffs.
I have no problem to provide more control to users. But easier & simpler the API is, more ppl will use it. That's my point.

 > If editHwnd is not child of dlgHwnd but buttonHwnd, then...
 
 >  Also dmfSubclassParent should be used only once for required control (parent) and dmfSubclassChildren only once for main hwnd.
 
 Sorry but I don't follow you. Could you elaborate it please?
 
 
 > And controls need to apply new theme each time dark mode has changed, or they will remain in previous mode and have wrong colors.
Same for title bar.

Theme & title bar could be included and hidden in the API, or I'm too naive? For me, plugin authors don't care what it is theme - all they want is their dialog can follow Notepad++'s mode (dark or light mode).

ozone10: > That doesn't make sense. ppl handle whole things themself, or they use API completely to do all stuffs.
> I have no problem to provide more control to users. But easier & simpler the API is, more ppl will use it. That's my point.

This why I provide combined `dmfAfterInitParent` and `dmfHandleChangeParent` flags, if they want to have more granular control they could use individual flags.

They might also use some non-standard controls/parent-child relationship which are not handled by combined flags.

> Sorry but I don't follow you. Could you elaborate it please?
Control should be subclassed only once.

Check related issue https://github.com/npp-plugins/plugindemo/issues/4 for edit control and its parent interaction.

> Theme & title bar could be included and hidden in the API, or I'm too naive? For me, plugin authors don't care what it is theme - all they want is their dialog can follow Notepad++'s mode (dark or light mode).

The problem is that control has no idea that dark mode changed.

donho: @ozone10 
The following issue is not addressed: 
https://community.notepad-plus-plus.org/assets/uploads/files/1682252209282-notepad_plugin_darkmode.png

is it?

ozone10: Since this window does not have title bar using `dmfSubclassParent | dmfSubclassChildren` flags is enough to apply dark mode.

And because there is only one button we can use `dmfSetThemeDirectly` in `NPPN_DARKMODECHANGED` on it to handle dark mode change.

donho: image

```cpp
void about()
{
	HWND hSelf = ::CreateDialogParam(_hInst, MAKEINTRESOURCE(IDD_ABOUTBOX), nppData._nppHandle, (DLGPROC)dlgProc, (LPARAM)NULL);
		    
	// Go to center
	RECT rc;
	::GetClientRect(nppData._nppHandle, &rc);
	POINT center;
	int w = rc.right - rc.left;
	int h = rc.bottom - rc.top;
	center.x = rc.left + w/2;
	center.y = rc.top + h/2;
	::ClientToScreen(nppData._nppHandle, &center);

	RECT dlgRect;
	::GetClientRect(hSelf, &dlgRect);
	int x = center.x - (dlgRect.right - dlgRect.left)/2;
	int y = center.y - (dlgRect.bottom - dlgRect.top)/2;

	::SetWindowPos(hSelf, HWND_TOP, x, y, (dlgRect.right - dlgRect.left), (dlgRect.bottom - dlgRect.top), SWP_SHOWWINDOW);

	::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfSubclassParent | NppDarkMode::dmfSubclassChildren), reinterpret_cast<LPARAM>(hSelf));
}
```

ozone10: @donho 
`::SetWindowPos` should be last. `NPPM_DARKMODESUBCLASSANDTHEME` requires redraw if window is visible 

image

donho: Thank you @ozone10

donho:
@ozone10 
One bug I have noticed in this PR:
After aplying the PR, when I switch between dark mode & light mode, the title bar of Preferences dialog remains the precedant mode (doesn't change).

My debug info:
```
Notepad++ v8.5.3   (64-bit)
Build time : May 22 2023 - 09:40:55
Path : C:\sources\notepad-plus-plus\PowerEditor\bin64\Notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.1702
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.4)
    NppPluginDemo (4.3)

```

ozone10:
@donho 
that is strange, I did not change function with title bar.
I also tested on two PC without issue.

Maybe Windows 11 issue. I could try to replace undocumented methods with documented for build 22621.

donho:
@ozone10 
> Maybe Windows 11 issue. I could try to replace undocumented methods with documented for build 22621.

If it's Windows 11 issue, then it's critical - Windows 11 is the newest and mainstream version of Windows.

If the bug mentioned above can be fixed, I would suggest the following API:
```cpp
	#define NPPM_DARKMODESUBCLASSANDTHEME (NPPMSG + 112)
	// ULONG NPPM_DARKMODESUBCLASSANDTHEME(ULONG dmFlags, HWND hwnd)
	// Add support for generic dark mode.
	//
	// Docking panels don't need to call NPPM_DARKMODESUBCLASSANDTHEME for main hwnd.
	// Subclassing is applied automatically unless DWS_USEOWNDARKMODE flag is used.
	//
	// Might not work properly in C# plugins.
	//
	// Returns succesful combinations from dmfRequiredMask.
	//

	namespace NppDarkMode
	{
		// Standard flags for main parent after its children are initialized.
		constexpr ULONG dmfInit = 0x0000000BUL; 

		// Standard flags for main parent usually used in NPPN_DARKMODECHANGED.
		constexpr ULONG dmfHandleChange = 0x0000000CUL;
	};

	// Examples:
	//
	// - after controls initializations in WM_INITDIALOG, in WM_CREATE or after CreateWindow:
	//
	//auto success = static_cast<ULONG>(::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(NppDarkMode::dmfInit), reinterpret_cast<LPARAM>(mainHwnd)));
	//
	// - handling dark mode change:
	//
	//extern "C" __declspec(dllexport) void beNotified(SCNotification * notifyCode)
	//{
	//	switch (notifyCode->nmhdr.code)
	//	{
	//		case NPPN_DARKMODECHANGED:
	//		{
	//			::SendMessage(nppData._nppHandle, NPPM_DARKMODESUBCLASSANDTHEME, static_cast<WPARAM>(dmfHandleChange), reinterpret_cast<LPARAM>(mainHwnd));
	//			::SetWindowPos(mainHwnd, nullptr, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_FRAMECHANGED); // to redraw titlebar
	//			break;
	//		}
	//	}
	//}
```

I'm aware of that the way you suggested could make the finest call. However, my point (as I said before) is to increase the usability of API. IMO plugin authors don't need all the finest callings. But even if they do, all the other flags can be added in the future version, and it will be retro-compatible.

The new simplified interface has been tested, and it works just fine:
https://github.com/npp-plugins/mimetools/pull/20/files

I can do another PR to adapt new interface, if you don't want to be bothered by doing the modification.
What do you think?

ozone10:
@donho 
I've used documented function for dark title bar for Windows 11, on newer Windows 10 it's undocumented but function is same.
Could you test? I tested it on 3 PCs without issues.

For next 3 weeks I will be busy, so feel free to do modifications.

donho:
@ozone10 
Tested v8.5.2 on another PC and this issue is reproducible:
Quick switching between light mode and dark mode makes the title bar remain the previously mode randomly,
But if mouse left clicks on editor zone, then the title bar synchronizes with the mode.

Here's the Debug Info:
```
Notepad++ v8.5.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.1702
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    XMLTools (3.1.1.13)
```
So it's not a regression - sorry for the false alert.

donho:
> @ozone10 Tested v8.5.2 on another PC and this issue is reproducible: Quick switching between light mode and dark mode makes the title bar remain the previously mode randomly, But if mouse left clicks on editor zone, then the title bar synchronizes with the mode.
> 
> Here's the Debug Info:
> 
> ```
> Notepad++ v8.5.2   (64-bit)
> Build time : Apr  4 2023 - 19:55:32
> Path : C:\Program Files\Notepad++\notepad++.exe
> Command Line : 
> Admin mode : OFF
> Local Conf mode : OFF
> Cloud Config : OFF
> OS Name : Windows 11 Home (64-bit)
> OS Version : 22H2
> OS Build : 22621.1702
> Current ANSI codepage : 1252
> Plugins : 
>     mimeTools (2.9)
>     NppConverter (4.5)
>     NppExport (0.4)
>     XMLTools (3.1.1.13)
> ```
> 
> So it's not a regression - sorry for the false alert.

Updated v8.5.2 to v8.5.3, it seems to me that this issue is more visible (more reproducible) in the latest version (v8.5.3).

donho:
@ozone10 
> I've used documented function for dark title bar for Windows 11, on newer Windows 10 it's undocumented but function is same.
> Could you test? I tested it on 3 PCs without issues.

Unfortunately the result is still the same.

PowerEditor/gcc/makefile
```diff
@@ -45,7 +45,7 @@ CPP_DEFINE := UNICODE _UNICODE OEMRESOURCE NOMINMAX _WIN32_WINNT=_WIN32_WINNT_VI
 LD := $(CXX)
 LDFLAGS := -municode -mwindows
 LD_PATH :=
-LD_LINK := comctl32 crypt32 dbghelp ole32 sensapi shlwapi uuid uxtheme version wininet wintrust
+LD_LINK := comctl32 crypt32 dbghelp ole32 sensapi shlwapi uuid uxtheme version wininet wintrust dwmapi
```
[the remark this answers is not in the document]
ozone10: I would keep it since this is how MS recommends it
https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes#enable-a-dark-mode-title-bar-for-win32-applications

ozone10:
@donho 
I see, dark mode is applied correctly, but redraw sometimes does not work. This is known issue with Windows dark mode. Rapidly changing modes results in some elements are not redraw correctly. You can test it with changing dark mode from "Windows settings -> personalization -> colors -> choose your color" with several explorer windows open.

PowerEditor/gcc/makefile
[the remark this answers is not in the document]
donho: @ozone10 
Then why don't we use only this API DwmSetWindowAttribute for windows 11/10?
Why do we need to tell 19041U build for using it?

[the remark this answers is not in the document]
ozone10: @donho 
It's for builds 19041 and greater.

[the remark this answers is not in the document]
donho: @ozone10 
> It's for builds 19041 and greater.

Can't find this info in above link.
Do you have any ref regarding it?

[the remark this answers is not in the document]
ozone10: @donho 
for Windows 10 it's undocumented, but from Windows 10 20H1 build 19041 it's working same as in Windows 11.

PowerEditor/src/NppDarkMode.h
```diff
@@ -118,7 +118,9 @@ namespace NppDarkMode
 	void setAdvancedOptions();
 
 	bool isWindows10();
+	bool isWindows10(DWORD build);
 	bool isWindows11();
+	bool isWindows11(DWORD build);
```
donho, line 123: This function is never called. It shouldn't be here, should it?

ozone10: I've added it for convenience, currently it isn't used, but future Windows 11 builds might have new API for dark mode.

It can be removed.

PowerEditor/gcc/makefile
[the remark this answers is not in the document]
donho: @ozone10 
Are you telling me that **Windows 10 20H1 build 19041 (and later builds)** is **Windows 11** except the version number? 
Or you're saying that they have different look & version number, but sharing the same runtime library & Win32 API?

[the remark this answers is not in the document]
ozone10: @donho 
they use same Win32 API, it is just that this particular function did not have in Windows 10 SDK macro/enum defined. 

There are some win32 API that were undocumented in previous OS and then documented in next OS.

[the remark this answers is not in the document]
donho: @ozone10 
(part 1/2)
According your logic:
> but from Windows 10 20H1 build 19041 it's working same as in Windows 11.

I would remove:
```cpp
bool IsWindows10(DWORD build)
{
	return (IsWindows10() && g_buildNumber >= build);
}
```
and modify `bool IsWindows11(DWORD build)` to:
```cpp
bool IsWindows11(DWORD build)
{
	return (IsWindows10() && g_buildNumber >= build);
}
```
...

PowerEditor/src/NppDarkMode.cpp
```diff
@@ -2760,8 +2865,16 @@ namespace NppDarkMode
 
 	void setDarkTitleBar(HWND hwnd)
 	{
-		NppDarkMode::allowDarkModeForWindow(hwnd, NppDarkMode::isEnabled());
-		NppDarkMode::setTitleBarThemeColor(hwnd);
+		if (NppDarkMode::isWindows10(19041U))
```
donho, line 2868: ...
And use `if (NppDarkMode::isWindows11(19041U))` here instead?
(part 2/2)

donho (changes requested):
There might be a falure of logic:

PowerEditor/src/NppDarkMode.cpp
donho: And I would use `if (NppDarkMode::isWindows10() && NppDarkMode::g_buildNumber >= 19041U))` (somethings like that) instead.

ozone10: @donho 
done

alankilborn:
Now that this is accepted, is [THIS](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13596#issuecomment-1561601504) the final version of the advice for plugin authors on how to do darkmode support in their plugins?  Or did something else change since then?

Is this info supposed to be captured somewhere else?  E.g. plugins section of user manual?  @pryrt , thoughts?

PR #13599, opened 2023-04-26
Backup old version of shortcuts.xml

### LINKED ISSUE #13589: Macro regression in new version of Notepad++ with old macro recorded by v8.5.2 (and previous ones)
fff5f9b24c1e7a6f33633b665c06e8785f6a3455 makes Macro & Run menu completely localizable, which is good. However, it will bring a regression and critical bug for all macro recorded  by v8.5.2 and previous versions.

STR:
1. Run v8.5.2 or any previous version of Notepad++
2. Record a macro with 3 ENTER. Save it under the name "3 ENTER".
3. Run "3 ENTER" - observe that 3 new lines are added. Close Notepad++
4. Now launch Notepad++ of branch master.
5. Run "3 ENTER" - observe that `&#x000D;&#x000A;&#x000D;&#x000A;&#x000D;&#x000A;` are added instead of 3 new lines.

It'll be a critical bug because not only new version of Notepad++ run the macro recorded by the previous versions incorrectly, but also, it will saved the transformed macro - users won't even be able to go back to the old version because `shortcuts.xml` has been modified by the new version. As a result, users lose all/some macro they have recorded.

If anyone has a solution for it, please let me know. Otherwise, fff5f9b24c1e7a6f33633b665c06e8785f6a3455 (also 1a1ce04caa0229b58fae056c89b60ed45762f34e, 0457a112c76ddc883a99074cf459a970741cf5b6 and c76f178534bba518f263c8120b732801e13c4916) will be reverted.

donho:
An enhancement (fff5f9b - it will be in v8.5.3) makes Macro & Run menu completely localizable. However, it will bring a regression and critical bug for all macro recorded by v8.5.2 and previous versions. In order to remedy this issue, shortcuts.xml will be copied to shortcuts.xml.v8.5.2.backup before being written. So user can recover backup file if they want to back to v8.5.2 or previous version with the old data.

Fix #13589

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1444,6 +1444,12 @@ Continue?"/>
             <LanguageMenuCompactWarning title="Compact Language Menu" message="This option will be changed on the next launch."/> <!-- HowToReproduce: toggle "Make language menu compact" via Preferences dialog "Language/Language Menu. -->
             <SwitchUnsavedThemeWarning title="$STR_REPLACE$" message="Unsaved changes are about to be discarded!
 Do you want to save your changes before switching themes?"/> <!-- HowToReproduce: In the Style Configurator dialog change some theme and switch to other theme without saving. -->
+            <MacroAndRunCmdlWarning title="Macro and Run Commands Compatibility" message="Your Macro and Run commands saved in Notepad++ v.8.5.2 (or older) may not be compatible with the current version of Notepad++.
+Please test those commands and, if needed, re-edit them.
+
+Alternatively, you can downgrade to Notepad++ v8.5.2 and restore your previous data.
+Notepad++ will backup your old &quot;shortcuts.xml&quot; and save it as &quot;shortcuts.xml.v8.5.2.backup&quot;.
+Renaming &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, your commands should be restored and work properly.&quot;/> <!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
         </MessageBox>
         <ClipboardHistory>
             <PanelTitle name="Clipboard History"/>
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1449,7 +1449,7 @@ Please test those commands and, if needed, re-edit them.
 
 Alternatively, you can downgrade to Notepad++ v8.5.2 and restore your previous data.
 Notepad++ will backup your old &quot;shortcuts.xml&quot; and save it as &quot;shortcuts.xml.v8.5.2.backup&quot;.
-Renaming &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, your commands should be restored and work properly.&quot;/> <!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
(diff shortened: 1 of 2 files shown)


Yaron10:
@donho,

:+1: 

The prominent warning is still needed. 
The user might create a lot of new Macro/Run commands and only then realize the old ones are not working properly.

Thank you for your work.

donho:
> The prominent warning is still needed.

@Yaron10 
Do you mean the first time doing the backup?

Yaron10:
@donho,

A "prominent warning" both in the download-page and before the first change to _shortcuts.xml_ might help soothing "complaining users".

donho:
@Yaron10 
Thank you for your feedback.

The following message has been added on the very first time of shortcuts.xml backup :
```
Your macro recorded under v8.5.2 or previous versions could be corrupted by this version of Notepad++.
If it happens, please delete damaged macro then record it again under the current version of Notepad++; 
or you can always go back to Notepad++ v8.5.2 then restore "shortcuts.xml.v8.5.2.backup" to "shortcuts.xml".
```

Please let me know if you have a better text or any modified version of above text.

Yaron10:
@donho,

👍 
Thank you for your work.

Thinking about it a bit more: 

_1._ Users who will NOT edit their Macro/Run commands, but just run them and encounter a problem - will not see the message. The forum would still be bombarded by "complaining users". :)
_2._ Most users would prefer not to downgrade to a previous version.
_3._ Why should Notepad++ backup `shortcuts.xml` if it can easily be done by the user? 
_4._ Run commands may also not work properly if the path contains non-western Unicode characters.
  
How about displaying the warning on _startup_ if `shortcutsCompatibilityWarning.xml` has not been created?
(No message nor backup on saving `shortcuts.xml`).

_Title:_
`Macro and Run Commands Compatibility`.

_Message:_
```
Your Macro and Run commands saved in Notepad++ v.8.5.2 (or older) may not be compatible with the current version of Notepad++.

Please test those commands and, if needed, re-edit them.

* It is highly recommended to backup the `shortcuts.xml` file first.
```

donho:
@Yaron10 
>
>1. Users who will NOT edit their Macro/Run commands, but just run them and encounter a problem - will not see the message. The forum would still be bombarded by "complaining users". :)

Whatever is done, it will happen anyway. That's the reason I hesitate to do this change.

>2. Most users would prefer not to downgrade to a previous version.

Sure. But at least they are informed.

>3. Why should Notepad++ backup shortcuts.xml if it can easily be done by the user?

I've got a lot of feedback from users who use heavily Macro feature, and a lot of them have long list of recorded macro - the macro list is too long to display entirely on macro menu.  That's why macro folder feature has been implemented.
I have no idea about the content of their macro, all I know is that the shortcuts.xml can contain not only the folder name in non-western characters, but also any binary values in macro! I don't think, in this case, macro can be editable easily.

Therefore the old shorcuts.xml must be backup, before Notepad++ rewriting all and screw users' data.

>4. Run commands may also not work properly if the path contains non-western Unicode characters.

True.

> How about displaying the warning on startup if shortcutsCompatibilityWarning.xml has not been created?
(No message nor backup on saving shortcuts.xml).

I would rename `v852ShortcutsHasBeenBackup.xml` to `v852ShortcutsCompatibilityWarning.xml` for marking the incompatible version, for users and for devs.
However, as I mentioned above, the backup is important for some users who have huge amount of macro - They may take time to convert their macro for using the newest version of Notepad++, but at least they could use the old version of Notepad++ with the backup file.

>Your Macro and Run commands saved in Notepad++ v.8.5.2 (or older) may not be compatible with the current version of Notepad++.
>
>Please test those commands and, if needed, re-edit them.
>
>* It is highly recommended to backup the `shortcuts.xml` file first.

By taking account of my arguments above, here's new warning message:

```
Your Macro and Run commands saved in Notepad++ v.8.5.2 (or older) may not be compatible with the current version of Notepad++.
Please test those commands and, if needed, re-edit them.
You can also restore "shortcuts.xml.v8.5.2.backup" to "shortcuts.xml" and use restored "shortcuts.xml" Notepad++ v8.5.2.
```

Please let me know your thought.

Yaron10:
@donho,

Sure, you have more experience with complaining users.

```
Your Macro and Run commands saved in Notepad++ v.8.5.2 (or older) may not be compatible with the current version of Notepad++.
Please test those commands and, if needed, re-edit them.

Alternatively, you can downgrade to Notepad++ v8.5.2 and restore your previous data.
Notepad++ will backup your old "shortcuts.xml" and save it as "shortcuts.xml.v8.5.2.backup".
Renaming "shortcuts.xml.v8.5.2.backup" -> "shortcuts.xml", your commands should be restored and work properly.
```

Thanks again for your work.

Issue #13601, opened 2023-04-29
GUI glitch in "Preferences -> Print"


Yaron10:
**STR:**
_Preferences -> Print_.
Add some variables to the _Left part._

**Result:**
![תמונה](https://user-images.githubusercontent.com/10229320/235325176-b6329673-a421-4bc9-a65a-66c01a8c57dc.png)

***
Notepad++ v8.5.2.

ozone10:
I've mentioned it here https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10169
I will see what I can do.

Yaron10:
@ozone10,

👍 
Thank you for looking into it, and best of luck.

Issue #12812, opened 2023-01-17
Preferences Printing Option Dropdowns Too Narrow, Clip Font Names


kkm000:
### Description of the Issue
In the Preferences dialog, Printing page, the drop-down boxes are too narrow to accommodate longer font names, making them indistinguishable. See the screenshot below.

<img alt="Unlikely to affect screen readers" src="https://user-images.githubusercontent.com/8228229/212837307-b28ef9a7-de3a-4faa-952e-0c85ba94d317.png" width="360px" />

The sesquipedalian “Segoe UI Variable” font with multiple display axes presets, resulting in full display names such as “Segoe UI Variable Display Semibold” first came with Windows 11.

### Steps to Reproduce the Issue
1. From the top menu bar, select _Setting/Preferences…_, then _Print_ in the right pane.
2. Scroll down any either of the two the font selection drop-down boxes, and, if you are unlucky enough, you'll find groups of fonts with indistinguishable clipped names.

   In my case, the “Segoe UI Variable Sma” fonts could have been named “…ll Semibold,” “…rt Ass On Steroids,” “…dge Instead Of Characters” and “…ck Them Bold So They Don't Do That Again.” I'm only guessing the names; all I in fact know is that they look like Segoe UI Variable and print with a different weight each.

### Expected Behavior
I expected a triumphal procession through the streets of Ancient Rome, riding a black Bentley _quadriga,_ wearing a laurel wreath, followed by elephants, lions, koalas, platypuses and other exotic animals, and the Roman citizens cheering me from both sides of the road, throwing roses and cactuses at me.

### Actual Behavior
I was not granted a triumph. _Of course_ it was M. Anthonius' jiggery-pokery! What else would you expect from this guy?!

### Debug Information
```
Notepad++ v8.4.8   (64-bit)
Build time : Dec 24 2022 - 19:40:27
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Program Files\Notepad++\change.log" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 22H2
OS Build : 22621.1105
Current ANSI codepage : 1252
Plugins : 
    Comment Wrap (1.0.0.5)
    DSpellCheck (1.4.24)
    ExtSettings (1.2.1)
    HexEditor (0.9.12)
    JSMinNPP (1.2205)
    mimeTools (2.9)
[...]

PR #13602, opened 2023-05-01
GUIEnhancement: Preference Print subdlg

ozone10:
add initializers
reorder controls for tab focus
add accelerators
make comboboxes larger
replace variable static text with edit control
fix override warnings

image

fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13601
fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12812


Yaron10:
@ozone10,

👍 
Very nice!
Thank you for the quick fix.

donho (changes requested):

donho (approved):

Issue #12822, opened 2023-01-19
v8.4.8 hanging while opening some Pascal files


donho:
Notepad++ v8.4.8 hanging while opening some Pascal files.

### Description of the Issue
Notepad++ v8.4.8 hanging (with Function list panel opened) while opening [these 2 Pascal files](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10460178/2_files_hanging_npp.zip).

### Steps to Reproduce the Issue
1. Launch Notepad++ v8.4.8
2. Open the Function list panel
3. Open the 2 provided files in zip

### Expected Behavior
2 files can be opened in Notepad++.

### Actual Behavior
Notepad++ is hanging.

### Debug Information
```
Notepad++ v8.4.8   (64-bit)
Build time : Dec 24 2022 - 19:40:27
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 21H2
OS Build : 22000.1455
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

```

~~The hanging of opening 2 Pascal files is due to the PR #12489, obviously the hanging issue is fixed in #12811, so the issue will be fixed in the next release (v8.4.9) of Notepad++~~

The hanging of opening 2 Pascal files is due to function list parser (C++ code) problem. See [the analysis](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12822#issuecomment-1397862706)

dinkumoil:
The file `main.pas` which is contained in the provided ZIP file contains code that does not represent a valid _Pascal_ unit file. Testing function list parsers with invalid test files that are not compilable should be avoided.

donho:
@dinkumoil 
> The file `main.pas` which is contained in the provided ZIP file contains code that does not represent a valid _Pascal_ unit file.

Could you provide more detail of this argument please?

rdipardo:
> The file `main.pas` which is contained in the provided ZIP file contains code that does not represent a valid _Pascal_ unit file. 

Indeed, the other one (`Bass.pas`) has problems of its own.
There's an unfinished nested block comment starting at line 791:

    Free Pascal Compiler version 3.2.2 [2022/09/24] for i386
[...]

Issue #11728, opened 2022-05-26
[GCC] Debug builds missing


ozone10:
As of commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d6595ad836c6a8c9618631bab38b3f900e01994f we have debug builds in appveyor but no artifacts are pushed

ArkadiuszMichalski:
@donho
It was probably discussed earlier, but debug for MinGW in appveyor is necessary? Takes a bit of space (over 120MB) and extends the build time. Probably for these reasons debug was omitted in appveyor.

I brought it back for testing https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11730 but I don't know if it's needed now. Can be skipped again with lines :
```
matrix:
  exclude:
    - compiler: GCC
      configuration: Debug
```

rdipardo:
@ArkadiuszMichalski,
> It was probably discussed earlier

I did mention it [here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11701/files) but no one replied. (For some reason GitHub will only show reviews when you're logged in.)

Like I said before:

> It will add ~15 mins. of build time to every Appveyor run.
If someone has to pay for it at month's end, I don't believe supporting MinGW is worth the money.
There's really no good reason to use GCC unless you're compiling GTK widgets.

donho:
122 MB + 139 MB = 261 MB on Appveyor!
Do you guy think it's worth to push the GCC debug binaries?
Among:
1. excluding GCC Debug compiling (skipped lines suggested above) 
2. compiling GCC Debug but not pushing the binaries (current situation) 
3. including GCC Debug binaries

What do you guys suggest?

rdipardo:
It's really a choice between 1. or 3. 
Building *without* pushing the binary is a waste of time, so #11730 is appropriate &#x2014; __*if*__ we commit to debug builds in the first place: ["in for a penny, in for a pound"][0]

[0]: https://www.merriam-webster.com/dictionary/in%20for%20a%20penny,%20in%20for%20a%20pound

MarkusBodensee:
My 2 cents would be: Build time should never be an argument if there is any kind of benefit, like revealing errors with unportable code etc. Same is true for storage space.

Of course, only if there are no additional financial expenses which are not acceptable.

According storage space, is there a mechanism to drop build artifacts after a certain period of time?

donho:
[...]

PR #13604, opened 2023-05-01
Fix GCC type limits warnings

### LINKED ISSUE #13603: [Code] GCC type-limits warnings
Log from gcc builds, e.g. https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/46920728/job/39r4a6eg1eo3rci3 contains [-Wtype-limits] warnings.

These checks are always true or false, and thus unnecessary.

ozone10:
add initializers

fix #13603

### DIFF
--- a/PowerEditor/src/ScintillaComponent/DocTabView.cpp
+++ b/PowerEditor/src/ScintillaComponent/DocTabView.cpp
@@ -195,7 +195,7 @@ void DocTabView::bufferUpdated(Buffer * buffer, int mask)
 
 void DocTabView::setBuffer(size_t index, BufferID id)
 {
-	if (index < 0 || index >= _nbItem)
+	if (index >= _nbItem)
 		return;
 
 	TCITEM tie{};
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -1313,7 +1313,7 @@ namespace NppDarkMode
 
 			DWORD textFlags = isCenter ? DT_CENTER : DT_LEFT;
 
-			if(::SendMessage(hwnd, WM_QUERYUISTATE, 0, 0) != NULL)
+			if(::SendMessage(hwnd, WM_QUERYUISTATE, 0, 0) != static_cast<LRESULT>(NULL))
 			{
 				textFlags |= DT_HIDEPREFIX;
 			}
--- a/PowerEditor/src/ScintillaComponent/Printer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Printer.cpp
@@ -70,12 +70,12 @@ size_t Printer::doPrint(bool justDoIt)
 {
 	const NppGUI & nppGUI = (NppParameters::getInstance()).getNppGUI();
 
-	POINT ptPage;
-	POINT ptDpi;
+	POINT ptPage{};
+	POINT ptDpi{};
 
-	RECT rectMargins;
-	RECT rectPhysMargins;
-	RECT userMargins;
+	RECT rectMargins{};
+	RECT rectPhysMargins{};
+	RECT userMargins{};
 
 	// Get printer resolution
 	ptDpi.x = GetDeviceCaps(_pdlg.hDC, LOGPIXELSX);    // dpi in X direction
@@ -126,7 +126,7 @@ size_t Printer::doPrint(bool justDoIt)
 	// Convert page size to logical units and we're done!
 	DPtoLP(_pdlg.hDC, &ptPage, 1);
 
-	TEXTMETRIC tm;
+	TEXTMETRIC tm{};
 
 	int fontSize = nppGUI._printSettings._headerFontSize?nppGUI._printSettings._headerFontSize:9;
 	int fontWeight = (nppGUI._printSettings._headerFontStyle & FONTSTYLE_BOLD) ? FW_BOLD : FW_NORMAL;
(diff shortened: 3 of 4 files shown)


PowerEditor/src/ScintillaComponent/DocTabView.cpp
```diff
@@ -195,7 +195,7 @@ void DocTabView::bufferUpdated(Buffer * buffer, int mask)
 
 void DocTabView::setBuffer(size_t index, BufferID id)
 {
-	if (index < 0 || index >= _nbItem)
+	if (index >= _nbItem)
```
Yaron10, line 198: @ozone10,

I've been wondering why index/pos is `size_t` across the project.
1. `int` should be sufficient for the number of opened tabs.
2.  `if (index < 0)` is a good "safety check".

What do you think?

Thank you.

ozone10: @Yaron10 
1. `int` is large enough, but replacing all `size_t` does not give significant performance improvement since I don't expect user to have 10000+ tabs.
2. it's good check but only for signed type.

Yaron10: @ozone10,

I may not have been clear enough.

> it's good check but only for signed type.

That was my point.
This "good safety check" is removed here because _nbItem/index are `size_t`.

As you and @donho understand.

Thank you again.

donho: @Yaron10 
> I've been wondering why index/pos is size_t across the project.

It was `void DocTabView::setBuffer(int index, BufferID id)` hence `if (index < 0)`.
The signature of function has been changed, but `if (index < 0)` not being removed.
The changing is just for the semantic reason - the index should be size_t - an unsigned type which should never negative.

rdipardo: > The changing is just for the semantic reason - the index should be size_t - an unsigned type which should never negative.

You mean to say it <del>should</del> *can* never be negative:

~~~cpp
#include <cstdio>
#include <cinttypes>
int main(void)
{
  printf("%" PRIu64, (size_t)-1);
}
~~~

Prints: `18446744073709551615`

To better explain @Yaron10's concern, a negative index won't be negative anymore; it will overflow to `ULLONG_MAX`, testing `true` when it should be `false`.

Remember how the Pascal function parser [got into infinite loops] when `targetStart` was negative? [The fix] involved changing `size_t` to a (signed) `intptr_t`.

[got into infinite loops]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12822#issuecomment-1397862706
[The fix]: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/63d1bbae78d01c9e2142e8617c78b4e93d91938d

Yaron10: @donho,

Thank you for replying.

I'll add some words to @rdipardo's eloquent explanation (thank you Robert!). 

I suppose that when `if (index < 0)` was added, the main concern was not explicitly assigning `-1` to `index`.
It was the thought of (e.g.) decrementing `index` in a loop, and accidentally ending up with a negative value.

It's true that `size_t index` would not have a negative value in that case.
But getting a wrong value is still possible.

If we get that wrong value by accidentally decrementing `index` to `-1`, we can assume that `if (index >= _nbItem)` would be `true`. - But the range of possible wrong values is wider.
So, isn't the use of an unsigned variable a potential cause of trouble? Wouldn't it be better to continue using those "safety checks" with a signed variable?

donho: @Yaron10  @rdipardo 
I see your point.
I may pay attention to this issue in the future.
For the current PR, let's make as less change as possible.

rdipardo: > For the current PR, let's make as less change as possible.

*Et voilà!*

~~~diff
diff --git a/PowerEditor/gcc/makefile b/PowerEditor/gcc/makefile
index f060e860..68a2ee2e 100644
--- a/PowerEditor/gcc/makefile
+++ b/PowerEditor/gcc/makefile
@@ -37,7 +37,7 @@ LEXILLA_DIRECTORY := ../../lexilla
 LEXILLA_TARGET := liblexilla.a
 
 CXX := $(CROSS_COMPILE)g++
-CXXFLAGS := -include $(GCC_DIRECTORY)/gcc-fixes.h -std=c++20
+CXXFLAGS := -include $(GCC_DIRECTORY)/gcc-fixes.h -std=c++20 -Wno-type-limits
 RC := $(CROSS_COMPILE)windres
 RCFLAGS :=
 CPP_PATH := $(SCINTILLA_DIRECTORY)/include $(LEXILLA_DIRECTORY)/include

~~~

Seriously, though &#x2014; is a harmless GCC diagnostic really worth a permanent, wide-ranging change to the code base?

I've [given this rant before](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11728#issuecomment-114008078), but the facts remain that GCC is *not* a static analyzer, like [the one that comes free with Visual Studio](https://docs.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis); it's not even a legitimate compile target, since N++ has no dependency on GTK or any other Linux-only libraries.

ozone10: @Yaron10 
I can also say using signed `int` can cause issue, because we could forget `if (index < 0)`.
Count of wrong values is same, but for unsigned variables the range is continuous and so it needs to be checked only once.

btw 
MSVC has [C4296](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4296?view=msvc-170)
but it's not enabled by default

PR #13609, opened 2023-05-03
Enhance shortcuts backup & warning

donho:
Commit 5b5c30b0d33f1aade36f7a270e60d52a1ff38613 backup current shortcuts file before rewritting it, and gives a warning to users. However, when a PC installs v8.5.3 (or later version) from scratch (ie. no v8.5.2 or previous version installed), there should be no warning and no backup. This commit remove unnecessary cases for warning user (and backup shortcuts file).

Enhance #13589 & #13599


Yaron10:
@donho,

👍 
Another case of "the fix is good, but the catch is even better". :)

Thank you for your work.

PR #13621, opened 2023-05-07
[xml] Standarize english.xml format by using XML pretty print

donho:
Use XML Tools plugin for formatting english.xml. 
Command: via menu "Plugins -> XML Tools-> Pretty Print" command.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13562#issuecomment-1518543453


donho:
I'll take in charge of the modification for all localization file.

PR #13627, opened 2023-05-07
Update portuguese.xml

Blackspirits:
Small change, I reduced 4 lines, because the text didn't fit all inside the toolbar box

### DIFF
--- a/PowerEditor/installer/nativeLang/portuguese.xml
+++ b/PowerEditor/installer/nativeLang/portuguese.xml
@@ -900,10 +900,10 @@ The comments are here for explanation, it's not necessary to translate them.
                 <Global title="Geral">
                     <Item id="6101" name="Barra de ferramentas"/>
                     <Item id="6102" name="Ocultar"/>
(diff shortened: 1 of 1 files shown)


donho:
@hugok79

Do you validate this change?

hugok79:
@donho Looks good to me. Thanks

ozone10:
@donho 
why not increase width, there is still enough space
image

hugok79:
In version 8.5.2 it was not possible to resize the preferences window.... Meanwhile, with this fix: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13632 
it will not be necessary to apply this submission.

hugok79:
> @donho why not increase width, there is still enough space image

This fix the issue

hugok79:
@donho @ozone10 This PR can be closed.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a1d7db8049ce89bd537d4ca769af8d7f10feee45

PR #13629, opened 2023-05-08
Make all the localization files pretty printed

donho:
This commit makes all the localization files pretty printed, according to the following commit: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/dccab1f25ab61835bfe4042edd7780e5b7b3149d

Please feel free to let me know your thoughts about this PR.


donho:
@ozone10 
Thank you for your suggestion.
Could you point out which file has the problem?

ArkadiuszMichalski:
@donho 

I sometimes in `polish.xml` leave additional comments in line with the node, like:
```
<item></item> <!-- My additional comments -->
```

Thanks to this I don't increase the number of lines and it is easier to compare the whole thing with the original `english.xml`. So please don't move such comments to separate lines (as suggested by @ozone10).

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13631

PR #13631, opened 2023-05-08
Make all the localization files pretty printed (2nd try)

donho:
This commit makes all the localization files pretty printed (indent only).


donho:
@ozone10 @ArkadiuszMichalski 
Please check if this PR meet your criteria.

ArkadiuszMichalski:
@donho 
It seems to be OK, the number of lines in the file has not changed. Is this a one-time formatting or will you repeat it from time to time?

From what I can see, it's not indent only, but also does minor formatting adjustments. I caught the belowe corrections only in the `polish.xml` file. 
- use closing `/>` without any space, so `<item/>` instead of  `<item />`
- in-line comments also don't have any spaces in front of them, so `<item></item><!-- My additional comments -->` instead of `<item></item> <!-- My additional comments -->`
- remove unnecessary white space, so `<item id="1">` instead of `<item   id="1">`

Probably some others as well, but I haven't analyzed it thoroughly (I don't use this plugin). Personally, this new correction does not bother me at all.

donho:
@ArkadiuszMichalski 
Thank you for your analysis.

> Is this a one-time formatting or will you repeat it from time to time?

It's one time formatting.
It's just like a suggestion, but I will let translators choose the format they prefer.

ozone10:
👍

donho:
@conky77 
It's OK if you want to remove the obsolete items from english.xml or the languages that you maintain.
For the other languages, I prefer to let the maintainers to remove them.

PR #13645, opened 2023-05-12
Fix menu strings cut off regression for some localication

### LINKED ISSUE #13556: Regression: menu strings are cut off
**STR:**
Replace `hebrew.xml` with [this file](https://github.com/notepad-plus-plus/notepad-plus-plus/files/11290996/hebrew.zip) (I've modified the currently provided file with some English entries).
Use the binary of the current master (https://github.com/notepad-plus-plus/notepad-plus-plus/commit/15decdb3d93a165ce362743ba32164391373687a).
Change NPP interface to Hebrew.
Restart NPP.
Edit -> Line Operations.

**Result:**

תמונה

**Expected (with binary prior to https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fff5f9b24c1e7a6f33633b665c06e8785f6a3455):**

תמונה

***

Notepad++ v8.5.2   (64-bit)
Build time : Apr 20 2023 - 09:12:20
Path : ---Notepad++.x64.Release.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Version : 20H2
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : 
    ComparePlus (1.1)
    HTMLTag (1.3.6)
    PythonScript (2)
    _CustomizeToolbar (4.2)

donho:
ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13556#issuecomment-1518197329

Fix #13556

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -504,14 +504,13 @@ LRESULT Notepad_plus::init(HWND hwnd)
 		ExternalLangContainer & externalLangContainer = nppParam.getELCFromIndex(i);
 
 		int numLangs = ::GetMenuItemCount(hLangMenu);
-		const int bufferSize = 100;
-		TCHAR buffer[bufferSize] = { '\0' };
+		TCHAR buffer[menuItemStrLenMax]{};
 		const TCHAR* lexerNameW = wmc.char2wchar(externalLangContainer._name.c_str(), CP_ACP);
 
 		int x = 0;
 		for (; (x == 0 || lstrcmp(lexerNameW, buffer) > 0) && x < numLangs; ++x)
 		{
-			::GetMenuString(hLangMenu, x, buffer, bufferSize, MF_BYPOSITION);
+			::GetMenuString(hLangMenu, x, buffer, menuItemStrLenMax, MF_BYPOSITION);
 		}
 
 		::InsertMenu(hLangMenu, x - 1, MF_BYPOSITION, IDM_LANG_EXTERNAL + i, lexerNameW);
@@ -587,12 +586,12 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	// Update Scintilla context menu strings (translated)
 	vector<MenuItemUnit> & tmp = nppParam.getContextMenuItems();
 	size_t len = tmp.size();
-	TCHAR menuName[64];
+	TCHAR menuName[menuItemStrLenMax];
 	for (size_t i = 0; i < len; ++i)
 	{
 		if (tmp[i]._itemName.empty())
 		{
-			::GetMenuString(_mainMenuHandle, tmp[i]._cmdID, menuName, 64, MF_BYCOMMAND);
+			::GetMenuString(_mainMenuHandle, tmp[i]._cmdID, menuName, menuItemStrLenMax, MF_BYCOMMAND);
 			tmp[i]._itemName = purgeMenuItemString(menuName);
 		}
 	}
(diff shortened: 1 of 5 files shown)


donho:
@Yaron10 @rddim 

Yaron10:
@donho,

It fixes my particular case. 👍 

What about
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2577594e2c5d4b3a1f14284e8cb7754f2f659a17/PowerEditor/src/Notepad_plus.cpp#L8560
?

And other `GetMenuString` calls? - The values seem quite random and inconsistent.

Thank you.

rddim:
@donho 

It also works in my case. The max size is 96 bytes (in v8.5.2 is more). Is it possible to increase it to 128 or more without affecting the performance?

donho:
@Yaron10 
Well caught!
Latest commit contains the fixes. Please let me know if you see any inconstancy.

donho:
@rddim 
> The max size is 96 bytes (in v8.5.2 is more). 

I don't get it. How come v8.5.2 is more?

Yaron10:
@donho,

Great work. 👍 
Unified, consistent and clear.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/694dd1c530f821bd543fb8453ae4b46cc037450b/PowerEditor/src/Notepad_plus.cpp#L526

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/694dd1c530f821bd543fb8453ae4b46cc037450b/PowerEditor/src/Notepad_plus.cpp#L733

***

> The max size is 96 bytes (in v8.5.2 is more).

In v8.5.2 we could use `64` non-western Unicode characters.
With this PR we can only use `48` (96/2).

Thank you.

donho:
@Yaron10 
> In v8.5.2 we could use 64 non-western Unicode characters.
With this PR we can only use 48 (96/2).

Maybe I'm missing somethings, but it's length of TCHAR or the length of GetMenuStringW. So it's 64 length PLUS 32 = 96 which larger than before (64).

Yaron10:
@donho,

> The max size is 96 bytes (in v8.5.2 is more).

This sentence causes the confusion.
In this PR we have 96 bytes which is obviously more than the previous 64 bytes.
But currently each non-western Unicode character takes **2 bytes** (UTF-8), and therefore we can only use **48 chars**.
In v8.5.2 we could use **64 chars** with the 64 bytes.

donho:
@Yaron10 

```
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2577594e2c5d4b3a1f14284e8cb7754f2f659a17/PowerEditor/Test/FunctionList/cpp/unitTest#L1770

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/2577594e2c5d4b3a1f14284e8cb7754f2f659a17/PowerEditor/Test/FunctionList/cpp/unitTest#L1795
```

Both of them are in unitTest so it's OK to leave them. The other 2 are fixed. Thank you.

> But currently each non-western Unicode character takes 2 bytes (UTF-8), and therefore we can only use 48 chars.

Thanks your explanation, I realized the more accurate fix: the length 64 is restored, but the length of 2 char arrays is doubled. (Please see the latest commit)

@Yaron10 @rddim 
Could you retest the newest commit please?

Yaron10:
@donho,

> Both of them are in unitTest so it's OK to leave them. The other 2 are fixed.

👍 

> the length 64 is restored, but the length of 2 char arrays is doubled. 

The menus are cut off with the last commit (my case and I'm pretty sure @rddim's case as well).

rddim:
@donho 
The https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13645/commits/3344cea26e1e06496786f9f1a660899b3b4acbcf work as expected. The very long menus are cut off, but this is normal, they are very long. Thanks for fixing again.

Yaron10:
@donho,

👍 
Great.

Thank you.

PR #13653, opened 2023-05-13
Installer: Only call NppShell CleanupDll if the DLL exists

sredna:
If the user chose to not install the context menu entry, the uninstaller will display a RunDLL error message box because NppShell.dll does not exist.


donho:
What's the issue you're addressing to?

PR #13654, opened 2023-05-13
Installer: Correctly delete tabContextMenu_example.xml in the users profile

### LINKED ISSUE #13670: Uninstaller fails to delete tabContextMenu_example.xml in the users profile
### Description of the Issue

The uninstaller fails to delete the users copy of tabContextMenu_example.xml because it is deleted while the uninstaller is in the All Users context.

1. Install to the default location (and user settings in %Appdata%)
2. Uninstall and say yes to delete user settings.
3. %Appdata%\Notepad++ is not deleted because tabContextMenu_example.xml is left behind.

Fixed by this [PR](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13654). (Ping @donho as requested)

sredna:
tabContextMenu_example.xml must be deleted in the correct shell context so that $APPDATA resolves to the users profile.

Fix #13670


donho:
@sredna 
`All Pull Requests, except for translations and user documentation, need to be attached to a issue on GitHub. `

So please create an issue and ping me. I'll reopen this PR then.

PR #13661, opened 2023-05-16
Check translation sent in installer Windows 11 and Update explorerContextMenuEntryLocal.ini

maisondasilva:
Although I use Portuguese from Brazil, I received the Portuguese from Portugal translation, so I added the "O" so that the two translations are the same, but it would be good to check why I didn't receive the Brazilian one!

image

### DIFF
--- a/PowerEditor/installer/nativeLang/explorerContextMenuEntryLocal.ini
+++ b/PowerEditor/installer/nativeLang/explorerContextMenuEntryLocal.ini
@@ -38,7 +38,7 @@ Edit_with_Notepad++=Modifica con Notepad++
 Edit_with_Notepad++=Editar com o Notepad++
 
 [brazilian_portuguese.xml]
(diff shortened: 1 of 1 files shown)


GurliGebis:
This file is no longer being used, please just send the text in #13201, and I'll make sure it gets added to the NppShell project.

GurliGebis:
@maisondasilva so the one in the nppshell project is correct?
If so, it will be included in the next release of Notepad++.

maisondasilva:
@GurliGebis is the translation I received in version 8.5.3 correct is that what you are asking me?

image

GurliGebis:
Yes - I can see it is the correct value 🙂

maisondasilva:
@GurliGebis Congratulations for the great work!

GurliGebis:
@maisondasilva thank you 🙂

PR #13662, opened 2023-05-17
Remove old context menu ini file and attributes file.

GurliGebis:
The old explorerContextMenuEntryLocal.ini file is no longer used, and keeping it here causes confusion with people adding pull requests to update it.
This removes it, along with the .gitattributes file, since it is no longer needed with the ini file gone.


GurliGebis:
@donho since we have moved away from the old NppShell dll, this file is no longer needed, and can be removed.

Issue #13663, opened 2023-05-18
C language parser for function list bug


TomPomme:
### Description of the Issue
Functions that appear in a while or if condition will be added to the function list.

### Steps to Reproduce the Issue
in a source file:
while (isValue(value)) { ... }
then
isValue(value) will appear in the function list

### How to fix the issue
In file PowerEditor/installer/functionList/c.xml change line 39 from
39: )*
to
39: )

This will make sure that there is at least one return type/identifier present in front of the function name identifier.

This will also resolve issue:
> "C" Language parser for FunctionList hangs forever on long list of strings #12497 

### Debug Information
Notepad++ v8.5.3   (64-bit)
Build time : May 15 2023 - 06:09:36
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Home (64-bit)
OS Version : 22H2
OS Build : 19045.2965
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
It's not a valid bug report.

Issue #12497, opened 2022-11-11
"C" Language parser for FunctionList hangs forever on long list of strings


swgmike:
### Description of the Issue
Using notepad v8.4.6 and selecting the "C" Language to parse for Function Lists, notepad will hang forever i.e. 15 minutes plus on long lists of strings that are between two to 30 characters long each.
To stop it, notepad must be killed in the task manager. 

### Steps to Reproduce the Issue
1. Paste a long list of strings between two to 30 characters each into the notepad editor. In my case I'm testing my udl to colorize the keyword lists that were pasted into the editor. The UDL editor doesn't cause this problem.
2. My string lists are up to 1500 lines.  If only 150 lines are in the notepad editor, it will take about 10 seconds to show only the filename in the function list panel, and no functions. This is correct, since the list of strings don't contain functions.
3. If the list is around 1500 lines or more and the function list "C" parser is applied, notepad will hang forever. However, if any other
language parser such as fortran, CPP, C#, java etc is applied, it will not cause notepad to hang forever.

### Expected Behavior
Expected no function list data to appear in the function list panel except the filename being parsed since the 1500 lines are just strings.
In this case the list is all the content of all the keyword lists being used for my udl to test the colorization of them.

### Actual Behavior
When the "C" language function list parser is applied, notepad will hang for ever on lists longer than 150 strings whereas other language parsers will run quickly.  There may be similar issues when trying to process a line that exceeds 8000 chars (i.e. the max Unix stream in/out buffer characters and may be the same for windows).

### Debug Information
Notepad++ v8.4.6   (64-bit)
Build time : Sep 25 2022 - 19:51:39
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Users\mike\Documents\DOORS_RPE_Library\DOORS_DXL_Examples\NotePPlus_DXL_tool\DXL.xml"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 8.1 (64-bit) 
OS Build : 9600.20671
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

benmurrell:
I was able to reproduce the hanging behavior with [invalid-c.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10015740/invalid-c.txt).

Steps:
1. Open the "invalid-c.txt" file
[...]

PR #13664, opened 2023-05-18
Fix C language parser for function list hanging issue

TomPomme:
#13663
The fix will make sure there is at least one return type / identifier in front of the function name identifier.
while (isValue(value)) { ... }
isValue(value) would be added to the function list

#12497
It will also stop hanging on long lists of identifiers.

### DIFF
--- a/PowerEditor/installer/functionList/c.xml
+++ b/PowerEditor/installer/functionList/c.xml
@@ -36,7 +36,7 @@
 							)
 							\b
 							\s*
-						)*
+						)
 						(?'DECLARATOR'
 							(?'POINTER'
 								\*


donho (changes requested):
You should also provide Unit Test file which make hanging in current master but it works with your PR.
See:
https://npp-user-manual.org/docs/function-list/#unit-tests

donho:
The issue is fixed according the following comment/report:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12497#issuecomment-1563262639

This issue will be closed. However if the unit test is provided and it doesn't work with the master branch but solve in this PR, the PR in question will be reconsidered.

Issue #12844, opened 2023-01-24
Hidden Lines command not working correctly; simple case


alankilborn:
### Description of the Issue
The _View_ menu's _Hide Lines_ feature is not correctly hiding lines, in all cases.

### Steps to Reproduce the Issue

1. Open Notepad++'s `license.txt` file into a file tab
2. Click+drag down the line number margin, starting on line 2 and stopping when the caret is in the first column of line 5:
![image](https://user-images.githubusercontent.com/77065706/214373988-c9b1b1bf-fdc6-4846-94b1-856d709bf440.png)
3. Execute _View_menu's _Hide Lines_ command (or press Alt+h) to obtain the correct hiding of lines 2-5, inclusive:
![image](https://user-images.githubusercontent.com/77065706/214374298-af41c3bf-13a3-4492-9b0d-08c4978f0999.png)
4. Click+drag down the line number margin, starting on line 7 and stopping when the caret is in the first column of line 11:
![image](https://user-images.githubusercontent.com/77065706/214374515-1784015c-f343-4003-bba6-0dc28c4707dd.png)
5. Execute _View_menu's _Hide Lines_ command (or press Alt+h).

### Actual Behavior
Line 6 did not remain visible:
![image](https://user-images.githubusercontent.com/77065706/214374821-19df67ba-90a5-4671-8ec6-6b35c5a6aba7.png)

### Expected Behavior
I expected lines 1, 6, and 12+ to remain visible.

### Debug Information
Notepad++ v8.4.8   (64-bit)
Build time : Dec 24 2022 - 19:40:27
Path : W:\npp.misc\_NoBackup\Releases\npp.8.4.8\npp.8.4.8.portable.x64\notepad++.exe
Command Line : -titleAdd="8.4.8 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.2364
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppEditorConfig (0.4)
    NppExport (0.4)
    PythonScript (3.0.14)

cynthplusplus:
It seems like putting the carat on the first line, hitting alt+h, then doing the same on the second line will give you this merged result, regardless of whether there is a line in between.

[...]

PR #13681, opened 2023-05-21
Fixes hidden lines merging to avoid merging across a visible line, issue #12844

cynthplusplus:
Fixes issue #12844. The fix is slightly hacky (shows overlapping open / close markers) but the behaviour is closer to what the user expects now. Clicking the margin on an overlapping marker will always open the lower hidden section.

Tested opening single line hidden sections with the following case, by clicking through markers in different orders (eg from top to bottom, from bottom to top, from a middle marker).
open hide test 1

Tested merging two adjacent hidden sections as follows:
hide merge test
hide merge test result

Also tested with folding, which doesn't work, but this appears to not be a regression. It seems code folding and hidden lines is already broken, and should be fixed in a separate issue. (Edit: Opened issue #13682)

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -3675,7 +3675,7 @@ void ScintillaEditView::hideLines()
 	size_t endMarker = endLine + 1;
 
 	// Remove all previous markers in between new ones
-	for (size_t i = startMarker; i <= endMarker; ++i)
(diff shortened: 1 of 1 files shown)


victorel-petrovich:
@cynthplusplus , if it is of help, I could test it if you build again the artifact, as it's expired (I use win 32bit)

victorel-petrovich:
To avoid overlapping markers, how about having just a single marker? 
Just like for folding.

victorel-petrovich:
I just tested the Win32 release artifact. 
Yay, one little bug less ! :))

Issue #10349, opened 2021-08-07
NPP RTL: tabs Close button is cut off


Yaron10:
**STR:**
Settings -> Preferences -> General -> Localization: select a RTL language (e.g. Hebrew).
Restart NPP.

**Result:**
The tabs Close button is cut off by 1px.

![תמונה](https://user-images.githubusercontent.com/10229320/128597694-4c99ce81-6f25-4492-9ed8-194ca45f9445.png)

***
Notepad++ v8.1.2   (32-bit)
Build time : Aug  3 2021 - 20:04:28
Path : C:---\notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

PR #13702, opened 2023-05-24
GUIEnhancement: Tabbar

### LINKED ISSUE #13701: [GUI] Tabbar size inconsistencies
### Description of the Issue

Tabbar has size inconsistencies for icons and height.

1. Switch between reduced and unreduced states
2. observe how tabs look after launch and between states

---

1. Check/uncheck close button with reduced unchecked

---

1. Observe font size in hiDPI with reduced unchecked

### Expected Behavior

No inconsistencies 
### Actual Behavior

Inconsistencies 
### Debug Information

Notepad++ v8.5.3

ozone10:
-  prefer `SystemParametersInfo` fonts over `DEFAULT_GUI_FONT`, should improve DPI 
-  fix inconsistent tab bar size values
-  add initializer
-  fix override warnings

fix #13701

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -227,7 +227,7 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	int tabBarStatus = nppGUI._tabStatus;
 
 	_toReduceTabBar = ((tabBarStatus & TAB_REDUCE) != 0);
-	int iconDpiDynamicalSize = nppParam._dpiManager.scaleY(_toReduceTabBar ? 13 : 20);
+	int iconDpiDynamicalSize = nppParam._dpiManager.scaleY(_toReduceTabBar ? g_TabIconSize : g_TabIconSizeLarge);
 	_docTabIconList.create(iconDpiDynamicalSize, _pPublicInterface->getHinst(), docTabIconIDs, sizeof(docTabIconIDs) / sizeof(int));
 	_docTabIconListAlt.create(iconDpiDynamicalSize, _pPublicInterface->getHinst(), docTabIconIDs_alt, sizeof(docTabIconIDs_alt) / sizeof(int));
 	_docTabIconListDarkMode.create(iconDpiDynamicalSize, _pPublicInterface->getHinst(), docTabIconIDs_darkMode, sizeof(docTabIconIDs_darkMode) / sizeof(int));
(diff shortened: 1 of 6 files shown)


Yaron10:
@ozone10,

👍 
Thank you for your work.
Special thanks for fixing the _Reduce_ inconsistency.

Some comments/questions:

**1.**  Why should the tab icon be scaled to _13_ and _20_ instead of the **original size 16**?
If you use HiDPI, many images are "stretched", and not using the original size doesn't make much of a difference.
But many people _strongly prefer_ "100% Scaling" and seeing the images properly.

Here is the difference on my machine:
16x16.
תמונה

20x20.
תמונה

On toggling _Reduce_ you could change the font only and keep the 16x16 image.
(This would also solve the disappearing icons in the _Documents List_ on toggling _Reduce_).

**2.** We don't have an option to select the tab's font size, and it's way too large now.
תמונה

Before this commit:
תמונה

**3.** Using 16x16 _Disk & Close_ images and centering the text is much nicer IMO.
תמונה

**4.** May I ask you to have a look at https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10349?
I've been using some tricky solutions, and I hope you (as a GUI sorcerer) could find a proper fix. :)
Whenever you have some free time - I'd highly appreciate it.

ozone10:
@Yaron10 
1. I used larger of original values. I will see what I can do with document list and document switcher icons.
2. Some people were complaining that previous font size was too small, if more people complain that it's too large I will consider change.
3. Maybe, but I did not touch code with close button.
4. I will see what I can do, unfortunately I will be busy for next 3 weeks.

Yaron10:
@ozone10,

> I will see what I can do with document list and document switcher icons.

Before you start working on that, let's ask @donho for his opinion on always scaling the disk image to _16_.
It should be a much better and easier solution IMO.

> if more people complain that it's too large I will consider change.

I'm quite sure many users would consider the current font-size too large.

> I will see what I can do, unfortunately I will be busy for next 3 weeks.

👍 
I do appreciate it.

Thanks again.

donho:
> I'm referring to the font-size when Reduce is unchecked.

 I'm referring to the font-size when Reduce is checked:

image

The font size is a little bit large, but it doesn't matter. OTOH, the text is not centered vertically, which is bothering me.

And indeed, it's excessive:
image

I'll see what I can do about it if @ozone10 is not available.

alankilborn:
> the text is not centered vertically, which is bothering me.

Just noting:
In non-Reduce'd mode, the text is not centered vertically either, in 8.5.3.
I don't know that I have ever noticed this (thus it has never bothered me).

Yaron10:
@donho,

> OTOH, the text is not centered vertically, which is bothering me.

The question is what the reference point for "centered vertically" should be: is it the _Disk_ image? What about the _Close_ image?
Wouldn't it be much nicer to unify the size of both images?

And as @alankilborn has pointed out, it should be tested in _Reduce/non-Reduce_ states.

**Another very important point**: the entire layout should be tested with Windows scaling set to 100%.
This is the first and basic test.

That requires a lot of work and patience. :)

If you're planing to change the font-size only, I'd suggest to test it with `pointsToPixels(10);` for the _non-Reduce_ state.
(I hope @ozone10 wouldn't mind).

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b88456764b6d136dd2ace689a076fb43846afd4b/PowerEditor/src/WinControls/TabBar/TabBar.cpp#LL346C47-L346C47

Thank you.

ozone10:
@Yaron10 @alankilborn @donho 
could you check https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13709 ?

PR #13711, opened 2023-05-29
Set dpiAwareness to PerMonitorV2 through manifest file

### LINKED ISSUE #6284: Notepad++ still doesn't handle scaling correctly
### Description of the Issue
When opening Notepad++ on 1440p monitor and then dragging the window to a 4k (2160p) monitor with 150% scaling enabled, text is only zoomed in and becomes blurry.

Apps that are fully dpi/scaling aware don't show this and change font to a more suitable one. For example MS Outlook.

Issue does not occur when main monitor is the 4k monitor, only when main monitor is of lower resolution or scaling.

1. Open Notepad++ with main screen 1440p
2. Drag Notepad++ window to secondary 4k monitor
3. Observe blurry text.

### Expected Behavior
Expected font to change to higher DPI one and get same result as opening Notepad++ when the 4k monitor is configured as main display. We can see Notepad++ can actually do this, because it works correctly.

### Actual Behavior
Blurry text on 4k monitor after dragging window from 1440p to 2160p monitor with 150% scaling enabled.

### Debug Information
Screenshots:
Normal 1440p layout:
image

After dragging screen to 4k monitor (blurry text):
image

Opening notepad++ on 4k monitor with this 4k monitor as main display in Windows (shart and crispy text):
image

il-mix:
Proposal to fix #6284 (and several other DPI-related issues).
DPI Awareness level set on manifest file as PerMonitorV2.
For Windows < 10 (more generally, Windows where dpiAwareness is not available), dpiAware directive will be used, as per Microsoft documentation https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process
Tested on Windows 11 with dual monitor: main with scale 125%, second with scale 100%

### DIFF
--- a/PowerEditor/src/dpiAware.manifest
+++ b/PowerEditor/src/dpiAware.manifest
@@ -1,9 +1,9 @@
-<?xml version='1.0' encoding='UTF-8' standalone='yes'?>
-<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
-  <asmv3:application xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
-    <asmv3:windowsSettings
-         xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
-      <dpiAware>true</dpiAware>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+  <asmv3:application>
+    <asmv3:windowsSettings>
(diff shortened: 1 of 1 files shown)


donho:
**1. PerMonitorV2: it would be the ideal solution if there were no custom control in Notepad++ : some part of the custom controls like panels and Find dialog are not rendered properly in PerMonitorV2 mode.**
ref:
47ca5281fee5fa0df4bed5e5c5ae2391d5034084

donho:
@ozone10 
Can we use PerMonitorV2 with what you're doing (removing DEFAULT_GUI_FONT)?

il-mix:
Thought it won't be so easy...
Thanks for your work @ozone10 !

iampradip:
Just setting manifest is not enough. UI has to be repositioned and resized to adjust with new DPI when moved to other monitor.

PR #1923, opened 2016-06-02
Added a -quitOnEmpty command line flag that causes Notepad++ to quit when the last tab is closed.

bavis-m:
Useful for people who use Notead++ for things like editing Git commit messages (using -multiInst -notabbar -nosession), and want to signal they are done editing by closing the tab with Ctrl-W instead of Alt-F4.

Fixes #1006

Issue #154, opened 2015-06-01
Can't compile with VS2013 anymore


cmeriaux:
Hello,
with the new **Unprecompile headers** commit, I can't compile with visual studio 2013 anymore. Can someone help me please ?
I use to import VS 2005 project file, VS convert to its own format then I was able to compile. Now it's not possible anymore. 
What am I doing wrong ? Does someone have an idea ?
Thanks 
Christophe

ariccio:
> I use to import VS 2005 project file, VS convert to its own format then I was able to compile. Now it's not possible anymore. 

What?? Why not use `notepadPlus.vcxproj`??

cmeriaux:
Helle Ariccio, thanks for your respond.
When I open the file notepadPlus.vcxproj in visual studio, then compile (release mode), I get the message :

```

1>------ Build started: Project: Notepad++, Configuration: Unicode Release Win32 ------
1>Project file contains ToolsVersion="12.0". This toolset is unknown or missing. You may be able to resolve this by installing the appropriate .NET Framework for this toolset. Treating the project as if it had ToolsVersion="4.0".
1>C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cpp.Platform.targets(44,5): error MSB8020: The builds tools for v120_xp (Platform Toolset = 'v120_xp') cannot be found. To build using the v120_xp build tools, either click the Project menu or right-click the solution, and then select "Update VC++ Projects...". Install v120_xp to build using the v120_xp build tools.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

```


What am I doing wrong ?

I noticed that opening notepadPlus.vs2005.vcproj invoke a project import which makes the compilation possible. That's why I used this file.

NN---:
Oh, it seems that vcproj is not updated accordingly with vcxproj.

NN---:
vcproj file is not maintained.
You have VS2012, this why you don't have vs120_xp platform.
Either change in project configuration to use vs110_xp or upgrade to a newer VS.

cmeriaux:
Thanks, It's better now.

I still have this error :
```
Project file contains ToolsVersion="11.0". This toolset is unknown or missing. You may be able to resolve this by installing the appropriate .NET Framework for this toolset. Treating the project as if it had ToolsVersion="4.0".
```

Compilation is OK but linking is KO
```
1>  
1>  Searching libraries
[...]

Issue #137, opened 2015-05-29
The update dialog should be blocking


milipili:
When asking if n++ is up-to-date, a dialog is created "No update is available". However, the form is not blocking and it is possible to try it several times, or to do something else, and to launch several wingup.
![npp-multiple-update-dialog](https://cloud.githubusercontent.com/assets/5516919/7894077/2119936c-0665-11e5-8c4e-ab975ab9cfa7.png)

MetaChuh:
assigned, but closed due to:
* inactivity (4y)
* inactivity since assignment (2y)
* minor relevance for the majority of users
* workaround: don't press on `update notepad++` more than once 😉

please reopen if wanted

greetings

PR #13729, opened 2023-05-31
Scintilla 5.3.5 Lexilla 5.2.5

chcg:
update to Scintilla Release 5.3.5 (https://www.scintilla.org/scintilla535.zip)

    Released 31 May 2023.
    On Win32, implement IME context sensitivity with IMR_DOCUMENTFEED. Feature #1310.
    On Win32 remove dependence on MSIMG32.DLL by replacing AlphaBlend by GdiAlphaBlend. Bug #1923.
    On Qt, stop movement of IME candidate box.
    On Qt, report correct caret position within paragraph for IME retrieve surrounding text.
    On Qt for Cocoa, fix crash in entry of multi-character strings with IME.

and Lexilla Release 5.2.5 (https://www.scintilla.org/lexilla525.zip)

    Released 31 May 2023.
    Add CharacterSetArray constructor without setBase initial argument for common case where this is setNone and the initialSet argument completely defines the characters. This shortens and clarifies use of CharacterSetArray.
    Bash: implement highlighting inside quoted elements and here-docs. Controlled with properties lexer.bash.styling.inside.string, lexer.bash.styling.inside.backticks, lexer.bash.styling.inside.parameter, and lexer.bash.styling.inside.heredoc. Issue #154, Issue #153, Feature #1033.
    Bash: add property lexer.bash.command.substitution to choose how to style command substitutions. 0 → SCE_SH_BACKTICKS; 1 → surrounding "$(" and ")" as operators and contents styled as bash code; 2 → use distinct styles (base style + 64) for contents. Choice (2) is a provisional feature and details may change before it is finalized. Issue #153.
    Bash: fix nesting of parameters (SCE_SH_PARAM) like ${var/$sub/"${rep}}"}. Issue #154.
    Bash: fix single character special parameters like $? by limiting style. Issue #154.
    Bash: treat "$$" as special parameter and end scalars before "$". Issue #154.
    Bash: treat "<<" in arithmetic contexts as left bitwise shift operator instead of here-doc. Issue #137.
    Batch: style SCE_BAT_AFTER_LABEL used for rest of line after label which is not executed. Issue #148.
    F#: Lex interpolated verbatim strings as verbatim. Issue #156.
    VB: allow multiline strings when lexer.vb.strings.multiline set. Issue #151.


chcg:
|                  | Fixed in Lexilla version...  | Ref.                          |
| :-----:          |  :-----:                     | :----:                        |
| #13151           |  5.2.5                       | ScintillaOrg/lexilla@682f573  |

PR #13744, opened 2023-06-05
Fix language type not remembered through sessions due to language being hidden from menu

### LINKED ISSUE #13743: Hiding Languages from sub-menu Languages makes language type for the hidden languages not remembered though the sessions
### Description of the Issue
Hiding Languages from sub-menu **Languages** makes language type (syntax highlighting & folding) for the hidden languages not remembered though the sessions.

1. Open Preferences dialog, go to "Languages" section.
2. In "Language Menu" part, choose C++ & XML from "Available items" list, put them into "Disabled items" by clicking `->` button.
3. Open some CPP and XML files, and make sure the syntax highlighting & syntax folding working for this files.
4. Close Notepad++ and relaunch it.

### Expected Behavior
The C++ & XML files reopened with both syntax highlighting & syntax folding of its language (C++ or XML) applied. 

### Actual Behavior
The C++ & XML files reopened with no any language applied.

### Debug Information
Notepad++ v8.5.3   (64-bit)
Build time : May 15 2023 - 06:09:36
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.1778
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
Fix #13743

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -6072,8 +6072,24 @@ void Notepad_plus::getCurrentOpenedFiles(Session & session, bool includUntitledD
 					continue;
 
 
-			generic_string	languageName = getLangFromMenu(buf);
-			const TCHAR *langName = languageName.c_str();
+			wstring	languageName = getLangFromMenu(buf);
+
+			if (languageName.empty()) // The lang menu item in question is hidden
+			{
+				NppParameters& nppParam = NppParameters::getInstance();
+				const NppGUI& nppGUI = nppParam.getNppGUI();
+
+				for (size_t k = 0; k < nppGUI._excludedLangList.size(); ++k) // try to find it in exclude lang list
+				{
+					if (buf->getLangType() == nppGUI._excludedLangList[k]._langType)
+					{
+						languageName = nppGUI._excludedLangList[k]._langName;
+						break;
+					}
+				}
+			}
+
+			const wchar_t* langName = languageName.c_str();
 			sessionFileInfo sfi(buf->getFullPathName(), langName, buf->getEncoding(), buf->getUserReadOnly(), buf->getPosition(editView), buf->getBackupFileName().c_str(), buf->getLastModifiedTimestamp(), buf->getMapPosition());
 
 			sfi._isMonitoring = buf->isMonitoringOn();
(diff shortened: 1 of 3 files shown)


donho:
@Yaron10 
Thank you for your feedback, very appreciated.

> If you run this binary with sessions saved with the current master, the issue still exists.
> 
> I suppose that's expected.

Writing & reading language type into & from session.xml both operations need language menu info and excluded language list.
The v8.5.3 release doesn't use excluded language list, so if language type is excluded from menu, it will write EMPTY lang type into session.xml. The binary of this PR reads EMPTY lang type and it can do nothing.

So the behaviour described above is expected.

Yaron10:
@donho,

Could you please test the following case?
Uncheck _Remember current session_.
Do not exclude any language.
Close NPP.
Delete _session.xml_.
Restart NPP.
Exclude _C++_.
Check _Remember current session_.
Open a _C++_ file.
Close & restart NPP.

Result:
No _C++_ syntax.

donho:
@Yaron10 

> Result:
>  No C++ syntax.

Tested with master with the PR merged.

Result:
With C++ syntax.

Yaron10:
@donho,

Sorry for the late response. I just got back to my PC.

I've tested it on two machines - _Windows 7 32-bit_ & _Windows 10 64-bit_. No C++ syntax.
Could you please try again following the exact STR?

@alankilborn,

Could you please try it too?

Thank you both.

Yaron10:
@donho,

I understand why we got different results.

`C++` is displayed incorrectly in the _Language_ tab in RTL.

תמונה

I therefore localized `C++` as `CPP`. And this is the causes of the problem.
Could you please try it?

Thanks again.

donho:
@Yaron10 


> `C++` is displayed incorrectly in the _Language_ tab in RTL.
> 
> תמונה

Interesting...
If it's due to the words ++C & C++, then we don't need the complex instructions you have provided, just putting C++ into excluded list should reproduce the bug...

However, after excluding C++ and relaunching Notepad++, I still can't reproduce the bug:
 
image

donho:
> Localize _C++_ to _CPP_.

@Yaron10 
How? In current `english.xml`, we don't provide any way to user to change language name.

Yaron10:
@donho,

I've been using `<Item id="46003" name="CPP"/>` for a long time.

Anyway - once I realized the problem on my machine was caused by that localized item, I thought it was a "mini-minor" issue. 
Thank you again for the nice catch and fix.

PR #13761, opened 2023-06-10
Add ability to select multiple files to close without saving in Document list

### LINKED ISSUE #13178: [Feature Request] - Option to hide prompt when closing an unsaved file

donho:
The 2 commands "Close Selected Files Without Saving" & "Close Other Files Without Saving" (via context menu) get user's confirmation only once then close all selected (or unselected) files without saving files.

Fix #13178

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -246,6 +246,24 @@ void Notepad_plus::command(int id)
 			}
 			break;
 
+		case IDM_DOCLIST_FILESCLOSENOSAVE:
+		case IDM_DOCLIST_FILESCLOSEOTHERSNOSAVE:
+			if (_pDocumentListPanel)
+			{
+				vector<SwitcherFileInfo> files = _pDocumentListPanel->getSelectedFiles(id == IDM_DOCLIST_FILESCLOSEOTHERSNOSAVE);
+				for (size_t i = 0, len = files.size(); i < len; ++i)
+				{
+					bool isSnapshotMode = NppParameters::getInstance().getNppGUI().isSnapshotMode();
+					doClose(files[i]._bufID, files[i]._iView, isSnapshotMode);
+				}
+				if (id == IDM_DOCLIST_FILESCLOSEOTHERS)
+				{
+					// Get current buffer and its view
+					_pDocumentListPanel->activateItem(_pEditView->getCurrentBufferID(), currentView());
+				}
+			}
+			break;
+
 		case IDM_DOCLIST_COPYNAMES:
 		case IDM_DOCLIST_COPYPATHS:
 			if (_pDocumentListPanel)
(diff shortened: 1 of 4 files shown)


Yaron10:
@donho,

👍 
Thank you for your work.

> get user's confirmation only once

I don't get the confirmation dialog at all.

And more importantly: 
What's the purpose of these two new commands?
Why not use the multiple-files Save dialog with the existing commands? - This would be a good and consistent improvement.

Am I missing anything?

donho:
@Yaron10 
> I don't get the confirmation dialog at all.
> 

Not yet, I will push other commits for that and for the translation.

> And more importantly: What's the purpose of these two new commands? Why not use the multiple-files Save dialog with the existing commands? - This would be a good and consistent improvement.

Sorry I don't follow you. Which dialog are you talking about?

Yaron10:
@donho,

> Not yet, I will push other commits for that and for the translation.

OK.

> Sorry I don't follow you. Which dialog are you talking about?

When you _Close All_, _Close All But_ (etc.) from the main menu - you get either the single-file Save dialog or the multiple-files Save dialog.

תמונה

Wouldn't it much better and more consistent to use those Save dialogs in the _Documents List_ as well?
Why do we need the new commands?

Yaron10:
The current implementation in the _Documents List_ is indeed inconvenient & inconsistent.
If you execute one of the _existing_ Close commands, you might get the single-file Save dialog multiple times.

But isn't the "requested" solution displaying either the single-file or the multiple-files Save dialogs (as it's done in the main menu)?

donho:
@Yaron10 
> Wouldn't it be much better and more consistent to use those Save dialogs in the Documents List as well?

You're right about it. Please create a PR and ping me, I'll see what I can do about it.

Yaron10:
@donho,

👍 
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13768.

Thank you.

Issue #13514, opened 2023-04-12
I lost my opened files


andreasdc:
Today when I opened Notepad++ it didn't have opened files, including the ones that are not saved, I had like 30 files opened there.

donho:
@andreasdc 
Could you provide more details?
How did it happen?
Could you provide your Debug Info via menu `?->Debug Info...`?

andreasdc:
I just started PC and when I opened it all my opened files were gone. I closed my PC with notepad being open, but I do it the same all the time.

Notepad++ v8.4.8   (32-bit)
Build time : Dec 24 2022 - 19:30:12
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : "C:\Users\user\Desktop\test.html"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 22H2
OS Build : 19045.2728
Current ANSI codepage : 1250
Plugins : 
    mimeTools (2.6)
    NppConverter (4.3)
    NppExport (0.3)

andreasdc:
Where's the backup folder? I updated notepad++ long time ago.

alankilborn:
> Where's the backup folder? I updated notepad++ long time ago.

Have you read the user manual and the FAQ on the [Community](https://community.notepad-plus-plus.org/) site?

donho:
@andreasdc 
Just open them in Notepad++

xomx:
@andreasdc 
In your case, the `nppLogNulContentCorruptionIssue.log` file should have been created. Could you please find that log and paste its content here?

Location in your case should be `%APPDATA%\Notepad++\`  (this expands into `C:\Users\[loginname]\AppData\Roaming\Notepad++\`)
![APPDATA-NPP](https://user-images.githubusercontent.com/13075183/232251108-534ce274-ec9b-4872-9a65-606f98338f90.png)

andreasdc:
This is from this week
`2023-04-11 05:59:04  WM_QUERYENDSESSION (lParam: 0 - ordinary system shutdown/restart) =====================================
2023-04-11 05:59:05  WM_ENDSESSION (wParam: TRUE, lParam: 0 - ordinary system shutdown/restart)
2023-04-11 05:59:05  C:\Users\x\AppData\Roaming\Notepad++\config.xml is opened.
[...]

PR #13774, opened 2023-06-12
Fix ignored writeStr return value in TiXmlDocument::SaveFile

xomx:
@donho 

When I analyzed the file saving way for the #13514 I have found a common problem in the N++ code but forgot to report it.

The [TiXmlDocument::SaveFile](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/28594daef3440b1e791cb03c9f7c507a74c54813/PowerEditor/src/TinyXml/tinyxml.cpp#L788) could return incorrectly `true` even when the underlying [writeStr func](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/28594daef3440b1e791cb03c9f7c507a74c54813/PowerEditor/src/TinyXml/tinyxml.cpp#LL787C4-L787C4) failed. This can have unexpected results, e.g. in your commit https://github.com/donho/notepad-plus-plus/commit/45148fef5bd518f3ff071f35862ceeb12eb356a3 where you are testing the `sessionSaveOK`.

I am sorry, but currentIy I cannot test this simple patch myself, I am using the Github web interface only.

### DIFF
--- a/PowerEditor/src/TinyXml/tinyxml.cpp
+++ b/PowerEditor/src/TinyXml/tinyxml.cpp
@@ -784,8 +784,7 @@ bool TiXmlDocument::SaveFile( const TCHAR * filename ) const
 		std::unique_ptr<std::string> outputStr = std::make_unique<std::string>();
 		Print(*outputStr, 0);
 		if (!outputStr->empty())
-			file.writeStr(*outputStr);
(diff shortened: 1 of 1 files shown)


donho:
@xomx 
Thank you for fixing it!

Issue #13734, opened 2023-06-02
consider showing non-printing characters even when "show all characters" is not on


molsonkiko:
### Description of the Issue

While I *really like* the new setting where I can see the Unicode codepoint of non-printing characters instead of their abbreviation, I'm not crazy about needing to turn on the `Show all characters` setting to see them. I preferred being able to clearly see where all the weird characters were at a glance without also needing to see the CR and LF.

It would be nice to be able to see those characters all the time (possibly with a configurable setting) the way it used to be.
### Steps to Reproduce the Issue

1. Create a new document
2. Hit `Ctrl-E` (assuming you have default keybindings, this will create the `ENQ` character, `\x05`)
3. It will be invisible until you hit the `Show All Characters` button in 8.5.3, but in earlier NPP versions those characters were always visible.

### Expected Behavior

As stated above, always make non-printing characters visible.
### Actual Behavior

See above
### Debug Information

```
Notepad++ v8.5.3   (64-bit)
Build time : May 15 2023 - 06:09:36
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Home (64-bit)
OS Version : 22H2
OS Build : 19045.2965
Current ANSI codepage : 1252
Plugins : 
    ComparePlus (1.1)
    CSVLint (0.4.6.4)
    CsvQuery (0.1)
    Demo (0.7)
    DSpellCheck (1.5)
    EnhanceAnyLexer (1.1.3)
    HexEditor (0.9.12)
    HugeFiles (0.4.1)
    JsonTools (5.0.0.5)
    mimeTools (2.9)
    NavigateTo (2.4.3)
    NppConverter (4.5)
    NppExport (0.4)
    PythonScript (3.0.16)
    QuickText (0.2.5.1)
    XMLTools (3.1.1.13)
```

The below screenshot shows what happened when I hit Ctrl+E 7 times in Notepad++ 8.5.3 with `Show all characters` off.
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/46202915/530c0853-4467-44ad-b6ae-9e2492af7995)

pryrt:
Option already exists:
**All**
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/17455758/8d129f24-76fb-4833-9f9c-40b6d01b8619)

**None**
[...]

PR #13780, opened 2023-06-13
Remember each individual show characters setting after toggling "Show All Characters" OFF

molsonkiko:
Would address #13734

When the user selects "Show all characters", their previous character showing settings (e.g., show nonprinting and control chars but not EOL or whitespace) will be saved so that when they toggle "Show all characters" OFF it reverts to their previous settings.

TO TEST:
1. Add this text to a document (UTF-8 bytes are `[0xc2, 0xa0, 0x20, 0x9, 0x5, 0xd, 0xa]`):
image
2. Turn on "Show all characters"
3. Turn off "Show non-printing characters" so that the `NBSP` is no longer visible.
4. Turn on "Show all characters"
5. Turn off "Show all characters" again.
RESULT: You should now be able to see all characters except the `NBSP` at the beginning.

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2398,8 +2398,6 @@ void Notepad_plus::command(int id)
 		case IDM_VIEW_ALL_CHARACTERS:
 		{
 			const bool isChecked = !(::GetMenuState(_mainMenuHandle, id, MF_BYCOMMAND) == MF_CHECKED);
-			checkMenuItem(id, isChecked);
-			_toolBar.setCheck(id, isChecked);
 			
 			auto& svp1 = const_cast<ScintillaViewParams&>(NppParameters::getInstance().getSVP());
 			// when checking box, show all char types
@@ -2408,11 +2406,13 @@ void Notepad_plus::command(int id)
 			bool eolShow = isChecked || svp1._previousEolShow;
 			bool npcShow = isChecked || svp1._previousNpcShow;
 			bool ccUniEolShow = isChecked || svp1._previousCcUniEolShow;
+			bool allChecked = whiteSpaceShow && eolShow && npcShow && ccUniEolShow;
 			checkMenuItem(IDM_VIEW_TAB_SPACE, whiteSpaceShow);
 			checkMenuItem(IDM_VIEW_EOL, eolShow);
 			checkMenuItem(IDM_VIEW_NPC, npcShow);
 			checkMenuItem(IDM_VIEW_NPC_CCUNIEOL, ccUniEolShow);
-
+			checkMenuItem(id, allChecked);
+			_toolBar.setCheck(id, allChecked);
 
 			svp1._whiteSpaceShow = whiteSpaceShow;
 			svp1._eolShow = eolShow;
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -2321,6 +2321,7 @@ void Notepad_plus::command(int id)
 
 			auto& svp1 = const_cast<ScintillaViewParams&>(NppParameters::getInstance().getSVP());
 			svp1._whiteSpaceShow = isChecked;
+			svp1.updatePreviousSettings();
 
 			const bool allChecked = svp1._whiteSpaceShow && svp1._eolShow && svp1._npcShow && svp1._ccUniEolShow;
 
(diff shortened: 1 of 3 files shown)


donho:
@molsonkiko 
You should always create an issue (bug) then create PR to link with the issue.
See:

molsonkiko:
Just realized my PR doesn't cover an edge case where the previous settings had all checked (by checking each box individually) and the user tries to uncheck `Show All Characters`. I need to make another commit to cover this, but don't have time to recompile and test now. Will be able to later today.

donho:
So this PR is making each individual show characters setting remembered after switching "Show All Characters" OFF.
It change the current behaviour and could bring the confusion.
Any thought about this PR? @ozone10 ? Anyone?

molsonkiko:
If someone wanted to turn off all special char showing, maybe add another
button to turn everything off?

Yaron10:
@donho & @molsonkiko,

If "Show all characters" is checked, all symbols are displayed.
Doesn't that mean the opposite on unchecking it - _Hide all characters_?

But even if we don't interpret it so literally...

Manually check the the first 4 _Show_ commands (-> "Show all characters" is auto-checked).
Uncheck "Show all characters".

Result:
No change at all.
Isn't that highly confusing?

molsonkiko:
> Manually check the first 4 _Show_ commands (-> "Show all characters" is auto-checked). Uncheck "Show all characters".
> 
> Result: No change at all.

Yeah, that was a deliberate choice on my part. The alternative was to make a special case to turn everything off when all settings were checked manually. I agree that it's somewhat confusing, but it's completely self-consistent (remember exactly what the settings were before hitting `Show all characters`), unlike the alternative.

I really do understand your objections, @Yaron10 and @donho. I can totally see a case for doing things your way. But making it easy to see non-printing characters is one of the things that I like most about Notepad++ compared to, say, VSCode. From my perspective, making it so that one of my favorite buttons causes NPP to permanently forget all of the settings I had specifically chosen feels like a regression.

molsonkiko:
> If "Show all characters" is checked, all symbols are displayed.
> Doesn't that mean the opposite on unchecking it - _Hide all characters_?

I'd also just observe that no, that's not the case. `not (A and B) != (not A) and (not B)`

molsonkiko:
Is there any way to make the button tri-modal? Like make it so that it has three settings, `show no special characters`, `show user-chosen characters`, `show all characters`? Maybe the button would begin each session in the `show user-chosen` setting, then clicking once would go to the `show all` setting, and then clicking again would go to the `show none` setting?

donho:
@Yaron10 
> Manually check the first 4 _Show_ commands (-> "Show all characters" is auto-checked). Uncheck "Show all characters".
> 
> Result: No change at all. Isn't that highly confusing?

Yes, "Show all characters" cannot be unchecked. It makes this option pointless - users cannot hide all characters by one click.

@molsonkiko 
> Is there any way to make the button tri-modal (and have the text change to reflect that)? Like make it so that it has three settings, show no special characters, show user-chosen characters, show all characters? 

I don't think such control does exist.

> Maybe the button would begin each session in the show user-chosen setting, then clicking once would go to the show all setting, and then clicking again would go to the show none setting?
 
Even it does exist, we should keep our GUI as straightforward as possible. Change the current behaviour is not only pointless, but also confusing for the users.

The current behaviour will be kept.

Yaron10:
@molsonkiko,

I've never used Non-Printing/Control characters in real life.

Reading https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13734#issuecomment-1575171074 - would one toggleable "Show Non-Printing and Control Characters" command solve your issue?

@donho wants to improve NPP, and I'm sure he'd be willing to consider a good alternative solution.

molsonkiko:
@Yaron10
Thanks for asking! I'm not sure what would be the most satisfactory solution aside from the one I put forward. Maybe somebody else who talked in the #13734 discussion could come up with a good idea that would satisfy everyone.

> would one toggleable "Show Non-Printing and Control Characters" command solve your issue?

That would definitely help.

Yaron10:
@molsonkiko,

Thank _you_ for your work.

> would one toggleable "Show Non-Printing and Control Characters" command solve your issue?
>> That would definitely help.

I wasn't following the implementation of "Show Control Characters & Unicode EOL", and I don't know why it had to be separated from "Show NPC".

> Maybe somebody else who talked in the https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13734 discussion could come up with a good idea that would satisfy everyone.

That would be great. Best of luck.

Issue #12366, opened 2022-10-20
When saving in Korean file name in Save As, the file name is saved differently than intended.


lunamoth:
### Description of the Issue
When saving in Korean file name in Save As, the file name is saved differently than intended.

### Steps to Reproduce the Issue
1. Save As
2. Write "한글" and enter

### Expected Behavior
Save as "한글.txt" file

### Actual Behavior
Save as "한글글" file

There is a problem that the last character is repeated and the .txt extension is not attached.

### Debug Information
Notepad++ v8.4.6   (64-bit)
Build time : Sep 25 2022 - 19:51:39
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line :
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 21H2
OS Build : 19044.2006
Current ANSI codepage : 949
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    Remove Duplicate Lines (1.3)

Avtem:
Can not reproduce on **Windows 7, Notepad++ v8.4.6   (64-bit)**

lunamoth:
Occurs when you write "한글" and hit Enter when a block is set in "글"

It doesn't happen if you write "글" and hit enter with the arrow keys → block setting is off.

In Windows default Notepad, there is no problem even when blocks are set.

Avtem:
Hm, i think video here would be very useful, because i don't fully understand what you mean by "block". Also as i don't speak Korean i don't how to type those characters and  don't have the correct keyboard layout enabled on my machine
To run the test i simply copied the text - i didn't type it. Maybe that's relevant for this issue

tomasz1986:
Korean uses letters which form syllables. Each syllable is a "block". If you type Korean, the letters you type form syllables. This is done automatically by the OS input system. In this specific case, once you type (note: not copy and paste) "한글", and then you press enter, you will end up with a file named something like.

```
한글.txt글
```

instead of the desired

```
한글.txt
```

If you want to reproduce the issue yourself, you will likely need to actually install Korean language input first, then type "한글" (which translates to "gksrmf" on the English keyboard layout) in the "save as" window. In fact, you don't need to type "한글" specifically, a just using the Korean input to type any kind of gibberish will be enough, and you will see that the last block/character ends up being repeated at the end.

[...]

Issue #12225, opened 2022-09-24
filename appending extension's last letter is duplicated in Hangul


SageSage7:
### Description of the Issue
the end of the non English filename is duplicated at the end of the extension

### Steps to Reproduce the Issue
1. Ctrl+N make a new file and write something
2. Ctrl+S 한글파일이름확장자버그{enter} // with checked append extension {enter} is actual key

### Expected Behavior
we expect 
한글파일이름확장자버그.txt

### Actual Behavior
it makes 
한글파일이름확장자버그.txt그

the end '그' of the non English filename is duplicated at the end of the extension
this only occures in non english blocked state
block means combination state like in Korean Hangul which needs combination keystokes to make an one letter.
if you press arrow keys rectangle block state disappear and bug doesn't happen, but this is not natural way.

### Debug Information
Notepad++ v8.4.5   (64-bit)
Build time : Sep  3 2022 - 04:05:32
Path : C:\util\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 7 Ultimate (64-bit) 
OS Build : 7601.24533
Current ANSI codepage : 949
Plugins : 
    ComparePlugin (2.0.2)
    DSpellCheck (1.4.24)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

SageSage7:
this bug has happened since several months ago. i mean 8.4.5 is not the beginning version of this bug.

PR #13788, opened 2023-06-16
Korean IME append extension issue 

80ROkWOC4j:
Related issues
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11582
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12225
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12366

Currently, there is a bug in notepad++'s add extension feature only for Korean input after it was changed to hooking-based in the commit below.
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b5a5baf13bb3f0b041f49698afb74af2b7059f6a#diff-eeb5624a35a43795da4eb970149a9ce7d22858b678a242affd2357520ea3e9f2R607

### Bug
1. Attempting to save via Enter appends the last character to the extension.
2. Candidate mode is similar, with more varied issues depending on IME.

### Cause
1. the hooking function is executed before the Hangul composition is completed and the last character is added after the extension.
2. Same for Candidate mode.

### Workaround
It is almost impossible to fix issue 2 while maintaining the current Enter hooking
Exiting Candidate Mode can be done by pressing Enter, ESC, number key, or clicking on a candidate character, but there is too much code to cover all of these cases.
In addition, the Windows input framework is fragmented into IMM and TSF, and various IMEs have different implementations, so it is almost impossible to determine the state of Candidate or Hangul composition through IME hooking.
I have seen differences in the events fired by different Windows versions and different IME programs for the same IME behavior.
This PR solves problem 1 and partially solves problem 2 by not saving with Enter when in Hangul mode.

### Changes
When in Candidate/Hangul mode, the user completes the combination with the first Enter, encounters the text with the extension, and must press Enter again to save.
In this case, the same user experience can be achieved by sending one more Enter keystroke in the code section, but I think it's not intuitively understandable for non-Korean developers and I'm not sure this won't cause other problems, so I excluded it.

Now
ezgif com-gif-maker(3)

Now candidate
notepp_now_candidate

Now candidate, Windows' lastest version of Korean IME
notepp_now_candidate2

This PR
ezgif com-gif-maker(2)
If the user enters Enter in Korean IME mode, the user must enter Enter once more to complete the save.

This PR candidate
ezgif com-gif-maker(1)

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -709,8 +709,8 @@ private:
 	UINT _currentType = 0;  // File type currenly selected in dialog.
 	UINT _lastSelectedType = 0;  // Last selected non-wildcard file type.
 	UINT _wildcardType = 0;  // Wildcard *.* file type index (usually 1).
-	LANGID _keyboardLayoutLanguage;
-	bool _isHangul; // Korean IME specific flag
+	LANGID _keyboardLayoutLanguage = LANG_NEUTRAL;
+	bool _isHangul = false; // Korean IME specific flag
 };
 std::unordered_map<HWND, FileDialogEventHandler*> FileDialogEventHandler::s_handleMap;
 
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -529,10 +529,10 @@ private:
 	{
 		auto hwnd = GetFocus();
 		auto himc = ImmGetContext(hwnd);
-		auto isImeActivated = ImmGetOpenStatus(himc); // return true when CJK IME is in local language input mode
+		auto isLocalInputMode = ImmGetOpenStatus(himc); // return true when CJK IME is in local language input mode
 		ImmReleaseContext(hwnd, himc);
 
-		return (lang == LANG_KOREAN) && isImeActivated;
+		return (lang == LANG_KOREAN) && isLocalInputMode;
 	};
 
 	// Transforms a forward-slash path to a canonical Windows path.
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -374,6 +374,8 @@ public:
 		_lastSelectedType(fileIndex + 1), _wildcardType(wildcardIndex >= 0 ? wildcardIndex + 1 : 0)
 	{
 		installHooks();
+		_keyboardLayoutLanguage = getKeyboardLayout();
+		_isHangul = isHangul(_keyboardLayoutLanguage);
 	}
 
 	~FileDialogEventHandler()
@@ -518,6 +520,21 @@ private:
 		}
 	}
 
+	LANGID getKeyboardLayout() 
+	{
+		return PRIMARYLANGID(LOWORD(HandleToLong(GetKeyboardLayout(0))));
+	}
+
+	static bool isHangul(LANGID lang)
+	{
+		auto hwnd = GetFocus();
+		auto himc = ImmGetContext(hwnd);
+		auto isImeActivated = ImmGetOpenStatus(himc); // return true when CJK IME is in local language input mode
+		ImmReleaseContext(hwnd, himc);
+
+		return (lang == LANG_KOREAN) && isImeActivated;
+	};
+
 	// Transforms a forward-slash path to a canonical Windows path.
 	static bool transformPath(generic_string& fileName)
 	{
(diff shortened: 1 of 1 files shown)


rdipardo:
This is almost identical to #12016, except the [`IsHangul` lambda function](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13788/files#diff-eeb5624a35a43795da4eb970149a9ce7d22858b678a242affd2357520ea3e9f2R634) does not simulate an extra press of the <kbd>Enter</kbd> key the way [the first iteration did](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/12016/files#diff-eeb5624a35a43795da4eb970149a9ce7d22858b678a242affd2357520ea3e9f2R685).

80ROkWOC4j:
> 
The more I dug into the issue with candidate mode, the more I realized it was more complicated than I initially thought. I also tried IME hooking to cleanly solve more common input situations, but I couldn't find a consistent solution across IMEs or Windows versions, so I decided to simplify the old code and defer saving due to Enter when in Korean mode.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -625,7 +649,37 @@ class FileDialogEventHandler : public IFileDialogEvents, public IFileDialogContr
...
+	static LRESULT CALLBACK LanguageDetectHook(int nCode, WPARAM wParam, LPARAM lParam)
+	{
+		if (nCode == HSHELL_LANGUAGE)
+		{
+			HWND hwnd = GetFocus();
+			auto it = s_handleMap.find(hwnd);
+			if (it != s_handleMap.end() && it->second && hwnd == it->second->_hwndNameEdit)
+			{
+				HKL hkl = (HKL)lParam;
+				it->second->_keyboardLayoutLanguage = PRIMARYLANGID(LOWORD(hkl));
```
donho, line 682: After testing, this line has never been run, obviously the condition `if (it != s_handleMap.end() && it->second && hwnd == it->second->_hwndNameEdit)` is always false.
Any idea about this issue?

donho (changes requested):
`_keyboardLayoutLanguage` is never initialized, since `it->second->_keyboardLayoutLanguage = PRIMARYLANGID(LOWORD(hkl));` is not run in any moment.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
80ROkWOC4j: That section of code won't be called unless the user changes the input language.   
Currently, _keyboardLayoutLanguage is only checking to see if it's Korean, so I didn't initialize it.  
Do you think it would be better for _keyboardLayoutLanguage to be initialized to English or the system locale? Or any other ideas?

donho: > That section of code won't be called unless the user changes the input language.

How come the condition `if (it != s_handleMap.end() && it->second && hwnd == it->second->_hwndNameEdit)` will turn into true when user changes the input language? 
And BTW, I add the 2 Chinese keyboard layouts into my system, by switching into them it didn't change anything. 

> Do you think it would be better for _keyboardLayoutLanguage to be initialized to English or the system locale? Or any other ideas?

It will be the next step. But for the moment, I just can't see how and why the condition mentioned above can be true.
Please provide the instruction to make line `it->second->_keyboardLayoutLanguage = PRIMARYLANGID(LOWORD(hkl));` to be run.

80ROkWOC4j: Let me correct: when the user changes input language, ```if (nCode == HSHELL_LANGUAGE)``` becomes true. ```if (it != s_handleMap.end() && it->second && hwnd == it->second->_hwndNameEdit)``` is looking for a ```FileDialogEventHandler``` like the other hooking functions above to access ```_keyboardLayoutLanguage```. I don't understand what's wrong with that code, are you experiencing the same issue with the other hooking function above?

donho: Strangely now I can trigger this condition to true by switching ON my Chinese input...
I guess it's due to startup of Windows.

donho: But still, if my Chinese input is already enabled, and I trigger save dialog, _keyboardLayoutLanguage is 0.
So let's detect input language while dialog is launched and initialize  _keyboardLayoutLanguage correctely.

donho (changes requested):
1. Initialize 2 member variables.
2. Detect input language while dialog is launched and initialize _keyboardLayoutLanguage correctely.

donho (changes requested):
Initialize the member variables in the definition of class.

donho (approved):

donho:
@80ROkWOC4j 
FYI: The merged PR has been reverted (https://github.com/notepad-plus-plus/notepad-plus-plus/commit/906f6e4386cf5acebfc1ace3aac9b83b26e7c70a) due to the regression.

PR #13814, opened 2023-06-21
[xml] Update japanese.xml

yasmise:
Add translation texts for these commits:
* Fix menu Tools contains 2 SHA-256 item while using localization (5b52386)
* Add SHA-512 hash features (4ffd897)
* [xml] Add missing entries in english.xml for SHA-1/SHA-512 (d7aee68)
* Add change history navgation commands (d9b9868)
* Fix inaccurate find/replace in files result while using invalid regexp (e7f7c31)
* Remove "Temporary doc site:" from localization files (c143a4a)

### DIFF
--- a/PowerEditor/installer/nativeLang/japanese.xml
+++ b/PowerEditor/installer/nativeLang/japanese.xml
@@ -5,7 +5,7 @@ Translation note:
 2. All the comments are for explanation, they are not for translation.
 -->
 <NotepadPlus>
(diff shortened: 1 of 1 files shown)


donho:
Please merge https://github.com/notepad-plus-plus/notepad-plus-plus/commit/50a15f2e59fac6350d6011b2201eab5f0b3e892a for fixing build error.

PR #13826, opened 2023-06-23
[xml] Update Slovak translation

kubalav:


kubalav:
I don't know what's wrong. The XML file seems invalid, but it is valid 🤔

donho:
Please merge https://github.com/notepad-plus-plus/notepad-plus-plus/commit/50a15f2e59fac6350d6011b2201eab5f0b3e892a for fixing build error.

PR #13828, opened 2023-06-24
[xml] Update spanish.xml

P0rsche-911:
New revision and update

### DIFF
--- a/PowerEditor/installer/nativeLang/spanish.xml
+++ b/PowerEditor/installer/nativeLang/spanish.xml
@@ -5,7 +5,7 @@ Translation note:
 2. All the comments are for explanation, they are not for translation.
 -->
 <NotepadPlus>
-	<Native-Langue name="Español" filename="spanish.xml" version="8.5.4">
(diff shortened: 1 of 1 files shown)


Patriccollu:
@P0rsche-911 
Review messages can be found here:
	https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/47385625/messages

A quote is misplaced on line 257:
`					<Item id="43048" name=Seleccionar y buscar si"&amp;guiente"/>`

Line should probably look like this:
`					<Item id="43048" name="Seleccionar y buscar si&amp;guiente"/>`

P0rsche-911:
It is strange to me that inverted commas appear there.

Those ampersand position changes are to avoid duplication of letters and, when used with the Alt key, directly execute their function. And they use different positions in the Galician version.

I had asked in the Notepad++ forum if they could be removed as they already have a function with the Ctrl key. They are necessary.

Notepad++spa

Julio860JVL (approved):

Julio860JVL:
Is this one correct?
<Item menuId="Window" name="&amp;Pestañas"/>
Line 24.

PR #13833, opened 2023-06-27
Fix Clipboard History panel shows corrupted data issue

### LINKED ISSUE #13817: Clipboard History panel shows unexpected hieroglyph-like symbols
### Description of the Issue
When Notepad++ is active and its Clipboard History panel is shown, after copying and pasting in both Notepad++ and other applications (such as other text editors, file managers, browsers) during some noticeable time (let's say, 2 hours) one may notice unexpected hieroglyph-like symbols (that were not part of any copy-pasted text).
This behavior is reproduced with and without plugins.
Please look at the screenshot (I did not alter the screenshot - it shows exactly how the Clipboard History panel looks like).
Npp_ClipboardHistory

1. Start Notepad++
2. Show the Clipboard History panel and keep it visible
3. Copy and paste different text inside and outside Notepad++ during a few hours (if you are a developer, it actually means just to keep Notepad++ open with its Clipboard History panel visible during a few working hours).
4. Observe how the lines in the  Clipboard History panel look like
5. I noticed that when I copy-paste from and to [FAR Manager](https://farmanager.com/download.php?l=en), especially when copying multi-line text (in [FAR Manager](https://farmanager.com/download.php?l=en) it is e.g. Ctrl+Insert and Ctrl+Alt+Insert when there are selected files in the file panels), the chances to get corrupted text in the Clipboard History are higher.

### Expected Behavior
All the copy-pasted strings shoud look as is, without any unexpected symbols.

### Actual Behavior
Unexpected hieroglyph-like symbols appear (see the screenshot). Also, some other strings look truncated (some trailing characters are missed).
I believe it may be caused by out-of-buffer reading, when e.g. some data in the Clipboard is expected to be a null-terminated string while it is actually not (I mean, it does not contain the trailing \0); or, maybe, some code in Notepad++ uses structures where the trailing \0 is expected but not provided.

### Debug Information
Notepad++ v8.5.4   (32-bit)
Build time : Jun 17 2023 - 20:40:14
Path : C:\Progs\Notepadpp\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.3086
Current ANSI codepage : 1251
Plugins :

### LINKED ISSUE #13844: Clipboard History panel provids cut data issue with Copy Binary Content command
### Description of the Issue
While using Copy Binary Content command with Clipboard History panel displayed, the entry of copy is displayed in the panel, but double clicking on it paste only the partial data (cut content).

1. Open Clipboard History panel.
2. Copy and paste `void ClipboardHistoryPanel::addToClipboadHistory(ClipboardData cbd)` into Notepad++.
3. Run Menu `Edit`->`Paste Special`->`Copy Binary Content`.
4. Observe **"void ClipboardHistoryPanel::a..."** entry appears in Clipboard History panel.
5. Double click on **"void ClipboardHistoryPanel::a..."** entry.

### Expected Behavior
Seeing `void ClipboardHistoryPanel::addToClipboadHistory(ClipboardData cbd)` is pasted into Notepad++.

### Actual Behavior
Seeing `void ClipboardHistoryPanel::addToCli` is pasted into Notepad++.

### Debug Information
```
Notepad++ v8.5.4   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.1928
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

donho:
Fix #13817, fix #13844

### DIFF
--- a/PowerEditor/src/WinControls/ClipboardHistory/clipboardHistoryPanel.cpp
+++ b/PowerEditor/src/WinControls/ClipboardHistory/clipboardHistoryPanel.cpp
@@ -20,25 +20,22 @@
 #include "clipboardFormats.h"
 
 
-#define CLIPBOARD_TEXTFORMAT CF_UNICODETEXT
 #define MAX_DISPLAY_LENGTH 64
 
 ClipboardData ClipboardHistoryPanel::getClipboadData()
 {
 	ClipboardData clipboardData;
-	if (!IsClipboardFormatAvailable(CLIPBOARD_TEXTFORMAT))
+	if (!IsClipboardFormatAvailable(CF_UNICODETEXT))
 		return clipboardData;
 
 	if (!OpenClipboard(NULL))
 		return clipboardData;
 	 
-	HGLOBAL hglb = GetClipboardData(CLIPBOARD_TEXTFORMAT); 
+	HGLOBAL hglb = GetClipboardData(CF_UNICODETEXT);
 	if (hglb != NULL) 
 	{ 
-		char *lpchar = (char *)GlobalLock(hglb);
-		wchar_t *lpWchar = (wchar_t *)GlobalLock(hglb);
-		
-		if (lpchar != NULL) 
+		unsigned char* pData = static_cast<unsigned char*>(GlobalLock(hglb));
+		if (pData != NULL)
 		{
 			UINT cf_nppTextLen = RegisterClipboardFormat(CF_NPPTEXTLEN);
 			if (IsClipboardFormatAvailable(cf_nppTextLen))
@@ -49,20 +46,22 @@ ClipboardData ClipboardHistoryPanel::getClipboadData()
 					unsigned long *lpLen = (unsigned long *)GlobalLock(hglbLen); 
 					if (lpLen != NULL) 
 					{
-						for (size_t i = 0 ; i < (*lpLen) ; ++i)
+						size_t nbBytes = (*lpLen + 1) * sizeof(wchar_t);
+						for (size_t i = 0 ; i < nbBytes; ++i)
 						{
-							clipboardData.push_back(static_cast<unsigned char>(lpchar[i]));
+							clipboardData.push_back(static_cast<unsigned char>(pData[i]));
 						}
 						GlobalUnlock(hglbLen); 
 					}
 				}
 			}
-			else if (lpWchar != nullptr)
+			else
 			{
-				int nbBytes = (lstrlenW(lpWchar) + 1) * sizeof(wchar_t);
-				for (int i = 0 ; i < nbBytes ; ++i)
+				wchar_t* lpwchar = (wchar_t*)pData;
+				size_t nbBytes = (lstrlenW(lpwchar) + 1) * sizeof(wchar_t);
+				for (size_t i = 0 ; i < nbBytes ; ++i)
 				{
-					clipboardData.push_back(static_cast<unsigned char>(lpchar[i]));
+					clipboardData.push_back(static_cast<unsigned char>(pData[i]));
 				}
 			}
 			GlobalUnlock(hglb);
--- a/PowerEditor/src/WinControls/ClipboardHistory/clipboardHistoryPanel.cpp
+++ b/PowerEditor/src/WinControls/ClipboardHistory/clipboardHistoryPanel.cpp
@@ -22,9 +22,9 @@
 
 #define MAX_DISPLAY_LENGTH 64
 
-ClipboardData ClipboardHistoryPanel::getClipboadData()
+ClipboardDataInfo ClipboardHistoryPanel::getClipboadData()
 {
-	ClipboardData clipboardData;
+	ClipboardDataInfo clipboardData;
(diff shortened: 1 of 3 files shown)


donho:
@d0vgan 

d0vgan:
Thank you, seems to work correctly!
Could you also confirm from your side: does the Clipboard History panel behave as intended when you open a binary file in Notepad++ and copy some bytes (unreadable symbols) via Edit -> Paste Special -> Copy Binary Content ? I mean, in such cases the Clipboard History panel shows different content than the one shown in the editing window. If you confirm it is expected, I'm OK with it.

donho:
@d0vgan 
Thank you for testing.

> Notepad++ and copy some bytes (unreadable symbols) via Edit -> Paste Special -> Copy Binary Content ? I mean, in such cases the Clipboard History panel shows different content than the one shown in the editing window.

The last commit should fix issue from binary copy and Far.exe, could you test it please?

Note that for the problem of Far.exe, the fix is here: 
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13833/files#diff-78c77172abf458182e3d0b1fbf99b45bc9d0124ffccb55b589f7e73621b8a714R23

By changing the internal used format ID for clipboard, the clipboard data copied via far.exe is not treated anymore as binary data.
Without knowing the implementation of Win32 API `RegisterClipboardFormat`, I have 2 hypotheses (explanations):
1. The same string **"Notepad++ Binary Text Length"** is used in Far.exe. But it's unlikely and it doesn't make sense.
2. The  API `RegisterClipboardFormat` uses the text for a MD5 (or other hash) as the key of a map, and the internal used **"Notepad++ Binary Text Length"**'s MD5 hash value has the collision with the hash of string ID used in Far.exe

If the 2nd point is true, then there would be another problem with another application... Let's wait & see.

donho:
@d0vgan 
It turns out my 1st hypotheses is true. After googling **"Notepad++ binary text length"** I stumbled on the [change log of Far](https://raw.githubusercontent.com/FarGroup/FarManager/master/far/changelog):
```
drkns 2023-06-27 19:58:03+01:00 - build 6169

1. Using a foreign format exclusively for the clipboard text size was probably not a good idea.
   E.g. now copying text between 6167 and 6168 is broken. Oops.

   To avoid repeating the same mistake in the future, we now have our own format ("FAR_Text_Length")
   and resort to "Notepad++ binary text length" only when our own format is not available.

2. With some magic copying text from pre-6168 should now work.
   Unfortunately, time travel is not yet discovered, so copying text into pre-6168 remains broken.
```
and here:
```
w17 2015-12-29 13:10:24+03:00 - build 4496

1. Copy/paste text containing '\0' in editor. When copied text contains zero characters,
   an additional 4 byte block containing the length of the text is created.
   The block type is 'Notepad++ binary text length', it must be compatible with notepad++ Paste Special.

```

d0vgan:
Thank you, works good!
So, if FAR Team wants to use Notepad++'s clipboard format (as an external one) in the future, then need to confirm the latest implementation of it in Notepad++ and to use "Notepad++ Binary Length" instead of "Notepad++ Binary Text Length".

Yaron10:
@donho,

> It turns out my 1st hypotheses is true. After googling "Notepad++ binary text length" I stumbled on the [change log of Far](https://raw.githubusercontent.com/FarGroup/FarManager/master/far/changelog):

Interesting and impressive.
Thank you for your work. 👍 

Copying binary
תמונה

I get in the Clipboard History panel:
תמונה

I suppose this is also related to a Non-English-Windows environment.

donho:
@d0vgan 
> So, if FAR Team wants to use Notepad++'s clipboard format (as an external one) in the future, then need to confirm the latest implementation of it in Notepad++ and to use "Notepad++ Binary Length" instead of "Notepad++ Binary Text Length".

If FAR Team changes it to "Notepad++ Binary Length" for keep using Notepad++'s clipboard format in the future, there will be the bug due to string without terminal `0` at the end (My experience is whole string is displayed but with the the random symbols as tail).
Or they have to follow the internal rule of Notepad++ for treating binary: the string length without terminal `0`.

By fixing the problem of "Copy Binary Content" issue on Clipboard History panel, I realized the fix in the 1st commit is not correct:
Since logically no software will/should use "Notepad++ Binary Text Length", the workaround in the first commit is not coherent and break Notepad++ feature. There's no way to prevent other software from using Notepad++ internal mechanism, but if they want to use it, they should do it in the way of Notepad++.

donho:
> Copying binary תמונה
> 
> I get in the Clipboard History panel: תמונה
> 
> I suppose this is also related to a Non-English-Windows environment.

@Yaron10 
I do believe it's the same numeric value, but it's interpreted/displayed differently by Scintilla & Windows List control.

Yaron10:
@donho,

And thanks again for your time and work.

PR #13854, opened 2023-07-02
Show parent directory in tabs with duplicate filenames, address issue 13715

### LINKED ISSUE #13715: [Feature Request] Adding the last directory name to each tab
I've been an avid Notepad++ user for over 10 years now and what's always been confusing and bugging me since day one is that the tabs only contain the filename.

This results to confusion if you have multiple files open that have the same filename (see photo):
tabs

Would it be possible to have a setting added that, once enabled, would show the last directory path along with the filename as this would solve the majority of the issues as you can not have multiple files with the same name in the same directory.
So for example if the path were to be _**/home/user/public_html/index.php**_ the tab name would then be _**public_html/index.php**_

This would be immensely helpful and would probably save me in total a few hours per month as I won't have to jump between the filenames to find the file that I'm looking for.

In exchange for my time saved and for the effort of adding this, I would be more than happy to donate $200 to have this added (perhaps it can be seen as some sort of a bounty?) and all I need is the PayPal address of the developer (or developers if this would be split between multiple developers).

molsonkiko:
Would fix #13715.

See examples below:

Default (old behavior)
disambiguation filename only example
Parent directory + filename
disambiguation dirname + filename example
Full path
disambiguation full path example


Yaron10:
@molsonkiko,

This is a useful improvement. Great work! 👍 

It obviously requires a thorough testing.
A quick point for the time being:
Open 2 identical-named files.
Move one to the other view.
Close that view.

Result:
The remaining file still has the folder prefix.

***

Users might expect the the list in the Window menu (and possibly other lists) to reflect the tabs' titles.
I'd suggest to wait for Don's principle approval.

Thank you.

molsonkiko:
Fixed the multi-view bugs you mentioned, @Yaron10 . You should be able to move files between views or rename/close/delete a file in one view and see the changes be reflected in both views at the same time.

I agree that the Document List and perhaps the Window menu as well should show the same name as the tabs. I haven't gotten around to that yet because I've found the Document List code particularly difficult to understand.

In Win32 Debug mode, I've also noticed a message box pop up when reopening a file that had been deleted/renamed outside of Notepad++ when in multi-view mode. The message box doesn't seem to relate immediately to any of the code I've written, and it didn't cause a crash, so I didn't think too much of it.

Yaron10:
@molsonkiko,

Very nice! 👍 
Thank you for your work.

This is a major refactoring, and some issues are expected.
Don't let it discourage you. :)

I've encountered some crashes in double-view mode.
Here is a consistent case:

Set the new preference to `Directory + Name` and restart NPP ("New 1" - no previous session).
Open 3 identical-named files.
Move one to the other view.
Move another file from the left view to the right view.
"Close All to the Left".
"File -> New".

Result:
NPP crashes.

***

The Documents List requires some more thinking.
Its width is normally relatively small (personally I never use the Path column).
A full-path in the Name column (_not fully displayed_) might be unhelpful and even disturbing.
The folder-prefix might be useful, but I'm not sure it's worth the extra work and refactoring.

Anyway, I think it would be much better not to add it at this stage.
Don might think of some optimization/improvements to the current PR, and that can affect the Documents List code as well.

What do you think?

molsonkiko:
@Yaron10 

I haven't been able to cause a crash by following your instructions.

Yaron10:
@molsonkiko,

It turns out it's not as consistent as I thought. Sorry.

Could you please try the following simpler case? - Reproducible about 7 out of 10 times.

Open 2 identical-named files.
Move one to the other view.
Close that file in the other view (I clicked on the tab's Close button).

Result:
NPP crashes.

_If you can't reproduce it on the first time, please exit NPP and try again._

Thank you.

molsonkiko:
Thanks for your persistence! Based on you description, I'm 99% sure it's a
race condition due to one of my changes in the last commit. I'm going to
need to rethink how my code handles the transfer of files between views.

donho:
When mouse is on the tab, users can see the full file path. So it's not an issue.
If users want to distinguish both files, they can always change the tab color.
This feature could be helpful for few people, but it makes code more complex and the higher price of maintaining.
This feature won't be done in Notepad++

molsonkiko:
I don't expect you'll ever change your mind, but I FWIW I believe I fixed the bug that Yaron10 described.

Yaron10:
@molsonkiko,

When I first saw this thread I thought it was the boldest PR ever submitted here. :)
I think it could have been a useful improvement, but I can understand @donho's decision to suspend it. - It IS a major refactoring.

Thank you again for your work. 👍 

> but I FWIW I believe I fixed the bug that Yaron10 described.

Would you mind posting the fix?

***

You are clearly very talented.
And I can refer you to some more consensual issues. :)

molsonkiko:
@Yaron10 
The fix was actually very simple, just removing a kludge that was no longer necessary. [Here's the commit](https://github.com/molsonkiko/notepad-plus-plus/commit/bc0f6b3de89ca8a36ca38a362421ce03e282e809).

Saul510:
Came across this just now and I think it's a big shame that this ended like this as it's a truly vital feature that should have been added years ago.

I know for a fact as I've been using N++ daily for over 10 years and although I have a huge amount of respect for @donho who has provided us with the fantastic and free to use tool (although I, like many others, have donated multiple times throughout the years) that having to hover your mouse over the tab each time or temporarily adding color to tabs is a not a efficient solution as it disturbs the work flow.

Many web developers work with files that are called _index.js_, _index.html_ or _index.php_ and it disturbs the work flow when you have to take extra steps to navigate correctly.  You could also compare identical filenames to unnamed tabs as it's the exact same situation and having to remember the tab positions is what I've been doing for years and it gets tedious.

I had to express myself on this as it's been frustrating me for years and seeing the feature be added would be so immensely useful for me and probably thousands of other web developers that are bugged by this issue, perhaps even being unaware of how much of their time and energy would be saved by a feature like this.

Stimmenhotel:
I would be very interested in this feature too.
I am working on different systems with files sharing the same name.
Sometimes it happens I don't check the titelbar for the path, as it is "out of the view".
Hope this can come to NP++.

alankilborn:
The _NavigateTo_ plugin has been shown to help when working with many tabs of the same name.

donho:
@alankilborn 
If you're talking about this:

image

then let's not forget there is the same feature in built-in Document List:

image

alankilborn:
>then let's not forget there is the same feature in built-in Document List

True, comparable functionality.

molsonkiko:
The *same feature* is in the Document List? Not so fast. The Document List
is great, and I use it all the time, but saying that there exists a list of
the full paths to all your documents, and therefore there is no need to
make it easier to distinguish two documents, is a bit like saying that text
editors allow you to put two documents side-by-side and scan them for
differences, and therefore there is no need for diff algorithms.

We wouldn't be computer programmers if we believed that, just because it is
*possible* to do X, meant that there was no reason to make it *easier* to
do X.

donho:
> The *same feature* is in the Document List? Not so fast.

I was talking about Document List comparing with NavigateTo. Nothing more.

> he Document List is great, and I use it all the time, but saying that there exists a list of the full paths to all your documents, and therefore there is no need to make it easier to distinguish two documents...

This (feature request) could be good to have, but there are a lot of existing concerning features to be considered (for example, renaming Tab). Therefore it's really hard to implement it without any regression/critical bug (and some are detected during the PR).

A (only) "good-to-have" feature, which will not only insert regressions and critical bugs, but also make the code even more complex, is not worth to be implemented IMO.

Hope my point can be understood.

Yaron10:
Since we all agree it's a "nice feature"...

@molsonkiko,

Could you please post here the fix to the issues detected above?
@donho might reevaluate it.

@donho,

I do understand your concerns here.

> but there are a lot of existing concerning features to be considered (for example, renaming Tab).

Could you please list any other potential issues?

***

Let's take it slowly and see how it goes.

Thank you both.

molsonkiko:
I have already fixed the issues you mentioned, but they could not be
evaluated because the PR was prematurely closed.
I'm not going to bother resubmitting unless I can be given some confidence
that I will be given a fair shake. I am happy to work on fixing problems
with my code if anything else needs improvement.

Yaron10:
@molsonkiko,

> I have already fixed the issues you mentioned

Thank you again for that.

***

> Let's take it slowly and see how it goes.

I think that would make sense for the benefit of Notepad++.

donho:
@Yaron10 
> Could you please list any other potential issues?

Another potential issue is clone document - it will be complicated to deal with it.
And then users will ask the alignment for document list.

> @donho might reevaluate it.

I have already reevoluated this feature, and provided the reason of rejection. I won't say no more about it.

alankilborn:
>I have already reevoluated this feature, and provided the reason of rejection. I won't say no more about it.

FWIW, I tend to agree with Don's assessment of this.  I think the "workarounds" presented (_Document List_, _NavigateTo_ plugin, and the PythonScript on the Community) are acceptable for the situation.

Yaron10:
Thank you all.

Issue #5813, opened 2019-06-21
Re-add undo after file reload enhancement as an optional preference


weirdalsuperfan:
undo after file reload has been removed in 7.7, as it caused more problems than advantages.

thanks for your input.

_Originally posted by @MetaChuh in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5626#issuecomment-499711875_

Could you add this feature back as optional (perhaps off by default)??? This actually proved EXTREMELY useful for me for debugging things and pinpointing regression bugs, and didn't impact my workflow or memory. However, the most recent patch fixed a problem that I really needed fixed (the tab dragging issue), so I feel like I can't really do without either of these things now. That said, I would rather revert to an earlier version than do without this feature now.

vertigo220:
This would be very helpful for me as my computer likes to just randomly reboot overnight sometimes, and occasionally, though rarely, has issues with things freezing or crashing during use. If either were to happen after making a bunch of changes to a file and needing to undo those changes, I would potentially be screwed, with the only option to go back and find a previously saved version from before the changes were made. Of course, this wouldn't be helpful for an unsaved file or one in which I needed to see what changes were made, as I may need to undo some but not all of them. I can, and have, survived without it for years, but there have been times, especially recently, where I've found myself needing it.

weirdalsuperfan:
More references: #5141 and #5595 - as you can see if you read those issues, there are other people who want this back, too.

Has this been re-added as an option yet? I've been digging through the issue and version history and I can't tell - I saw something about auto-saving sessions, but I don't think that preserves undo history after individual file reload (e.g., on change detect), does it?

vertigo220:
Another case where this is useful, which I've experienced at least a few times in just the past couple weeks, is if you're working on files, with N++ open, and doing testing, and your script modifies one of the files. If you then access that file, N++ will say it's been changed and ask if you want to reload it. If you say yes, because you want to see the results, and they're not what you expected, perhaps wiping out the contents of the file entirely, then they're just gone, with no way to undo, even though N++ was never even closed. In this case, the change just occurred, and one would expect to simply be able to undo them to go back to the previous state of the file, before reloading it, but no such luck. Of course, I realize the smart thing to do in such a case is to save a copy of the file, but let's face it, we don't always think to do such things or bother to take the time.

[...]

PR #13858, opened 2023-07-03
Fix Change History wrong display after "Reload from Disk" command

### LINKED ISSUE #12261: Reloading a file will mark all lines as changed in change history bar
Title pretty much explain all there is to it.

Stock notepad++ 8.4.6 32bits (if that matter...). No plugins, besides standard ones.

### LINKED ISSUE #12319: Change History Margin - reloading document
First time posting in GitHub, so apologies if I am doing something wrong.

I have just updated to the latest version of Notepad++ (v8.4.6 - 64bit) on my Windows 11 Pro.
I really like the change history margin. I did however find a slight annoyance with it:

If I reload the document (Reload from disk, or CTRL + R) then the history margin of the entire document is marked with a green bar. This then makes it more difficult to distinguish new changes, as they get 'lost' in the green bar, defeating a bit its purpose. The only way I could find to 'clear' the history margin again was to close the document and then open it again.

I believe a better option would be to simply 'clear' the history margin on reload, as if the document had just been opened for the first time.

Thank you for your time.

### LINKED ISSUE #13735: [Feature requests] Option to clear 'Change history margin' when reloading file from disk
Please add option to clear/reset 'Change history margin' after file is reloaded from disk.

donho:
The code for fixing is from Change Lines plugin:
https://github.com/vinsworldcom/nppChangedLines/blob/main/PluginDefinition.cpp

Fix #12319, fix #12261, fix #13735

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -8656,3 +8656,16 @@ HBITMAP Notepad_plus::generateSolidColourMenuItemIcon(COLORREF colour)
 
 	return hNewBitmap;
 }
+
+
+void Notepad_plus::clearChangesHistory()
+{
+	Sci_Position pos = (Sci_Position)::SendMessage(_pEditView->getHSelf(), SCI_GETCURRENTPOS, 0, 0);
+	int chFlags = (int)::SendMessage(_pEditView->getHSelf(), SCI_GETCHANGEHISTORY, 0, 0);
+
+	SendMessage(_pEditView->getHSelf(), SCI_EMPTYUNDOBUFFER, 0, 0);
+	SendMessage(_pEditView->getHSelf(), SCI_SETCHANGEHISTORY, SC_CHANGE_HISTORY_DISABLED, 0);
+
+	SendMessage(_pEditView->getHSelf(), SCI_SETCHANGEHISTORY, chFlags, 0);
+	SendMessage(_pEditView->getHSelf(), SCI_GOTOPOS, pos, 0);
+}
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -644,4 +644,6 @@ private:
 	void updateCommandShortcuts();
 
 	HBITMAP generateSolidColourMenuItemIcon(COLORREF colour);
+
+	void clearChangesHistory();
 };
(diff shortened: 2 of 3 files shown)


rdipardo:
This would [wipe out the undo buffer] along with the change indicators! Seems like [*un marteau pour écraser une mouche*] to me. 

You could at least create a warning dialog, so users will know they can't get their revisions back. Some are still upset about the *last* time you began clearing the undo buffer (to save memory): #5813

[wipe out the undo buffer]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fcebb59fa4ccf3da42534a32a8980683e1da9764/PowerEditor/src/Notepad_plus.cpp#L8666
[*un marteau pour écraser une mouche*]: https://en.wiktionary.org/wiki/écraser_une_mouche_avec_un_marteau

Yaron10:
@donho,

Great. 👍 
Thank you for your work.

@vinsworldcom,

Thank you for your contribution to the Change History feature. 👍 

@rdipardo,

Could you please explain your concern? This PR is only addressing Reload.

vinsworldcom:
> This would [wipe out the undo buffer](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fcebb59fa4ccf3da42534a32a8980683e1da9764/PowerEditor/src/Notepad_plus.cpp#L8666) along with the change indicators!

Correct.  So use with caution I what @rdipardo  is saying - or perhaps inform users with a confirm dialog.  I know it does this so I don't care and just use it if the full green change margin for a new file really bothers me.

[Note from](https://www.scintilla.org/ScintillaDoc.html#ChangeHistory) :

> Change history depends on the undo history and can only be enabled when undo history is enabled and empty. It should be enabled once when a file is loaded after calling [SCI_SETUNDOCOLLECTION(true)](https://www.scintilla.org/ScintillaDoc.html#SCI_SETUNDOCOLLECTION) and [SCI_SETSAVEPOINT](https://www.scintilla.org/ScintillaDoc.html#SCI_SETSAVEPOINT).

So that's what I do.

Cheers.

donho:
@Yaron10 
Thank you for checking.

@vinsworldcom 

> or perhaps inform users with a confirm dialog. I know it does this so I don't care and just use it if the full green change margin for a new file really bothers me.

It's applied only to "Reload from Disk" command in which undo buffer is already wiped out.
It's a correct behaviour IMO after the command in question.

vinsworldcom:
> Reload from Disk" command in which undo buffer is already wiped out.
It's a correct behaviour IMO

In that case, I would agree!

Cheers.

PR #13822, opened 2023-06-22
Fix compiling on g++ 13 (#13821)

buzzcoding:
Fixes the missing imports of _cstdint_ for _uint8_t_ in _MISC/sha1/calc_sha1.cpp_ and _string_ for _std::wstring_ in  _WinControls/ReadDirectoryChanges/ReadDirectoryChanges{,Private}.h_. See issue #13821 for more details.

PR #13861, opened 2023-07-05
Add change history navgation commands

### LINKED ISSUE #12248: [Feature Request] Changed History: Go to Next/Prev block
"Changed History" margin was introduced in NPP 8.4.6.
Two menu commands "Go to Next/Prev block" would be a useful addition to that new feature.

Thank you.

donho:
Add Go to next/prev change & clear change history commands.

image

Not that the localization will be added later.
Also, I'm not sure that the beep warnings will be kept.

Fix #12248


rdipardo:
Look like GCC builds [are broken] again.

Merging #13822 might help.

[are broken]: https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/47465957/job/a3otlnnyvapgx30v

Yaron10:
@donho,

Very nice! 👍 
Thank you for your work.

1. After executing "Clear Change History" the Undo button is still enabled. Is it `checkUndoState()`?
2. Shouldn't "Clear Change History" (i.e. the menu item) be enabled/disabled as well?
3. Personally I like the beeps, but I think that many users would not.
4. Sorry I didn't explicitly mention it: I implemented `IDM_SEARCH_CHANGED_FIRST` in `changedHistoryGoTo()`.
If you don't want it, the code can be slightly shortened.
5. @rdipardo's [Warn about clearing Undo/Redo](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13858#issuecomment-1619275728): what do you think of `Clear Change History (and Undo/Redo)`?
It's not the best wording. Should a reference in the User Manual be enough?

donho:
@Yaron10 

>  After executing "Clear Change History" the Undo button is still enabled. Is it checkUndoState()?
Shouldn't "Clear Change History" (i.e. the menu item) be enabled/disabled as well?

Smart thinking. Fixed in the latest commit.

> but I think that many users would not.

Agree, it's removed for being coherent with other actions.

> Sorry I didn't explicitly mention it: I implemented IDM_SEARCH_CHANGED_FIRST in changedHistoryGoTo().
If you don't want it, the code can be slightly shortened.

Sorry I don't follow you - I don't see IDM_SEARCH_CHANGED_FIRST. Could you elaborate it please?

> @rdipardo's https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13858#issuecomment-1619275728: what do you think of Clear Change History (and Undo/Redo)?
It's not the best wording. Should a reference in the User Manual be enough?

A reference in the User Manual will be enough. Users will complain about warning everytime they clear change history.

Yaron10:
@donho,

> Fixed in the latest commit.

👍 
Thank you again for your work.

Allow me to ask: the change in `checkUndoState()` is indeed good for "Change History", but - are the `isChangeHistoryEnabled` checks needed for each and every Undo/SCN_UPDATEUI?
Wouldn't your initial implementation (+ a `checkUndoState()` call) be better?

***

_First:_

```cpp
if (idGoTo != IDM_SEARCH_CHANGED_PREV)		// Next or First.
```

```cpp
if (idGoTo == IDM_SEARCH_CHANGED_NEXT)
```
(That block is not executed for FIRST).

```cpp
intptr_t endRange = (idGoTo == IDM_SEARCH_CHANGED_NEXT) ? currentLine + 1 : lastLine;
```
(`endRange == currentLine + 1` on NEXT & `endRange == lastLine` on FIRST).

***

> Shouldn't "Clear Change History" (i.e. the menu item) be enabled/disabled as well?

I may not have been clear about it.

Isn't the convention to enable/disable the popup itself in such cases?
For example: if you have one tab opened in Firefox, the tab context-menu popups which are only relevant to multiple-tabs are disabled.
Also - it would be enabling/disabling one item instead of 3 sub-items.

(In "Edit -> CaSe conversion" it would be enabling/disabling one item instead of **8** sub-items).

***

Compare (Plugin/Plus) has been using `Ctrl+Alt+C` & `Ctrl+Alt+Shift+C` for a long time.
As you understand.

Yaron10:
@alankilborn,

> However, if Notepad++ feels it must do this, it should pick what it feels is the best keycombos available (to it). Plugin users should adapt.

Unfortunately you're right. :)

Yaron10:
> But adapting is not a big deal: Reassign the shortcuts you love back to the plugin.

Correct. But Compare users would still ask/complain.

donho:
@Yaron10 
> After executing "Clear Change History" the Undo button is still enabled. Is it checkUndoState()?

I totally misread the above sentence. I'll fix it as soon as I'm back to my dev pc.
But thanks that, the enabling/disabling of 3 commands are improved and they are more accurate now.

> Shouldn't "Clear Change History" (i.e. the menu item) be enabled/disabled as well?

3 commands are enabled if the current buffer "canUndo", and disabled otherwise. Of course, if Change History marge is disabled, 3 commands are disabled as well.

> Allow me to ask: the change in checkUndoState() is indeed good for "Change History", but - are the isChangeHistoryEnabled checks needed for each and every Undo/SCN_UPDATEUI?
Wouldn't your initial implementation (+ a checkUndoState() call) be better?

I have thought about your approach, it's much more accurate indeed, but it'll be difficult to maintain the code. In the way of the last commit there will be redundant checks on each call, but each call are on a simple undo/redo/init/disable (or enable) change history action. So these redundant checks are invisible.

> (That block is not executed for FIRST).

I still don't follow you - as my understanding, they are wrap feature: in the case of searching next, if the last line is reached but nothing found, we go to the first line for keeping searching. This ability should be kept IMO.

> Compare (Plugin/Plus) has been using Ctrl+Alt+C & Ctrl+Alt+Shift+C for a long time.

> IMO, at this "late" stage in its life, Notepad++ should not assign shortcut keys for such new features (users know how to assign shortcuts if they want them).

OK, I will remove the shortcuts from the commands, and let users assign them.

Yaron10:
@donho,

> In the way of the last commit there will be redundant checks on each call, but each call are on a simple undo/redo/init/disable (or enable) change history action. So these redundant checks are invisible.

The advantage in this approach is nice and clever. 👍
I'm still wondering if the Change History commands should be disabled if there's no History.
The Bookmarks go-to commands are not disabled if there are no bookmarks.
And the Style Jump Up/Down commands are not disabled if there are no highlights.

> I'll fix it as soon as I'm back to my dev pc.

Now it's my turn not to understand: what's there to fix if you prefer the last commit?

***

When wrapping around for NEXT, we start searching from the beginning of the file. 

```cpp
		if (line == -1)		// Wrap around.
		{
			intptr_t endRange = (idGoTo == IDM_SEARCH_CHANGED_NEXT) ? currentLine + 1 : lastLine;		//	"+ 1": currentLine might be *0*.
```

is used for wrapping around for NEXT **or GoTo First**:

```cpp
		if (line == -1)		// Wrap around **or GoTo First**.
		{
			intptr_t endRange = (idGoTo == IDM_SEARCH_CHANGED_NEXT) ? currentLine + 1 : lastLine;		//	"+ 1": currentLine might be *0*.
```

`IDM_SEARCH_CHANGED_FIRST` is (basically)  achieved with less than one line of code-change:

`intptr_t endRange = (idGoTo == IDM_SEARCH_CHANGED_NEXT) ? currentLine + 1 : lastLine;`

instead of

`intptr_t endRange = currentLine + 1;`.

**`changedHistoryGoTo()` with `IDM_SEARCH_CHANGED_FIRST`:**
```cpp
void Notepad_plus::changedHistoryGoTo(int idGoTo)
{
	int mask =	(1 << SC_MARKNUM_HISTORY_REVERTED_TO_ORIGIN) | 
					(1 << SC_MARKNUM_HISTORY_SAVED) |
					(1 << SC_MARKNUM_HISTORY_MODIFIED) |
					(1 << SC_MARKNUM_HISTORY_REVERTED_TO_MODIFIED);
	intptr_t line = -1;
	intptr_t blockIndicator = _pEditView->getCurrentLineNumber();
	intptr_t lastLine = _pEditView->execute(SCI_GETLINECOUNT);

	if (idGoTo != IDM_SEARCH_CHANGED_PREV)		// Next or First.
	{
		intptr_t currentLine = blockIndicator;

		if (idGoTo == IDM_SEARCH_CHANGED_NEXT)
		{
			// Start from currentLine (not currentLine + 1) in case currentLine is not-changed and the next line IS changed. lastLine is at least *1*.
			for (intptr_t i = currentLine; i < lastLine; i++)
				if (_pEditView->execute(SCI_MARKERGET, i) & mask)
				{
					if (i != blockIndicator)		// Changed-line found in a different block.
					{
						line = i;
						break;
					}
					else
						blockIndicator++;
				}
		}

		if (line == -1)		// Wrap around or GoTo First.
		{
			intptr_t endRange = (idGoTo == IDM_SEARCH_CHANGED_NEXT) ? currentLine + 1 : lastLine;		//	"+ 1": currentLine might be *0*.
			for (intptr_t i = 0; i < endRange; i++)
				if (_pEditView->execute(SCI_MARKERGET, i) & mask)
				{
					line = i;
					break;
				}
		}
	}
	else	// Prev.
	{
		while (true)
		{
			line = _pEditView->execute(SCI_MARKERPREVIOUS, blockIndicator, mask);
			// "line == -1": no changed-line found. "line != blockIndicator": changed-line found in a different block.
			if (line == -1 || line != blockIndicator)
				break;
			else
				blockIndicator--;
		}

		if (line == -1)	// Wrap around.
			line = _pEditView->execute(SCI_MARKERPREVIOUS, lastLine - 1, mask);
	}

	if (line != -1)
	{
		_pEditView->execute(SCI_ENSUREVISIBLEENFORCEPOLICY, line);
		_pEditView->execute(SCI_GOTOLINE, line);
	}
}
```

**`changedHistoryGoTo()` without `IDM_SEARCH_CHANGED_FIRST`:**
```cpp
void Notepad_plus::changedHistoryGoTo(int idGoTo)
{
	int mask =	(1 << SC_MARKNUM_HISTORY_REVERTED_TO_ORIGIN) | 
					(1 << SC_MARKNUM_HISTORY_SAVED) |
					(1 << SC_MARKNUM_HISTORY_MODIFIED) |
					(1 << SC_MARKNUM_HISTORY_REVERTED_TO_MODIFIED);
	intptr_t line = -1;
	intptr_t blockIndicator = _pEditView->getCurrentLineNumber();
	intptr_t lastLine = _pEditView->execute(SCI_GETLINECOUNT);

	if (idGoTo == IDM_SEARCH_CHANGED_NEXT)		// Next.
	{
		intptr_t currentLine = blockIndicator;

		// Start from currentLine (not currentLine + 1) in case currentLine is not-changed and the next line IS changed. lastLine is at least *1*.
		for (intptr_t i = currentLine; i < lastLine; i++)
			if (_pEditView->execute(SCI_MARKERGET, i) & mask)
			{
				if (i != blockIndicator)		// Changed-line found in a different block.
				{
					line = i;
					break;
				}
				else
					blockIndicator++;
			}

		if (line == -1)		// Wrap around.
		{
			intptr_t endRange = currentLine + 1;		//	"+ 1": currentLine might be *0*.
			for (intptr_t i = 0; i < endRange; i++)
				if (_pEditView->execute(SCI_MARKERGET, i) & mask)
				{
					line = i;
					break;
				}
		}
	}
	else	// Prev.
	{
		while (true)
		{
			line = _pEditView->execute(SCI_MARKERPREVIOUS, blockIndicator, mask);
			// "line == -1": no changed-line found. "line != blockIndicator": changed-line found in a different block.
			if (line == -1 || line != blockIndicator)
				break;
			else
				blockIndicator--;
		}

		if (line == -1)	// Wrap around.
			line = _pEditView->execute(SCI_MARKERPREVIOUS, lastLine - 1, mask);
	}

	if (line != -1)
	{
		_pEditView->execute(SCI_ENSUREVISIBLEENFORCEPOLICY, line);
		_pEditView->execute(SCI_GOTOLINE, line);
	}
}
```

***

> Shouldn't "Clear Change History" (i.e. the menu item) be enabled/disabled as well?

I'm reading this line and, my goodness!, no wonder you didn't have the slightest idea what I wanted. Sorry.
I meant **"Change History"**.
תמונה

And
תמונה

Disable one "father" instead of many "children".

***

> OK, I will remove the shortcuts from the commands, and let users assign them.

Thank you.

@alankilborn,

Thank you too for that.

donho:
@Yaron10 
> I'm still wondering if the Change History commands should be disabled if there's no History.

So you are for or against Change History commands being disabled when there's no more change history?

donho:
@Yaron10 
> I meant "Change History"

OK I was asking myself if you mean sub-folder "Change History". But I read it again and I saw it was "Clear Change History menu item".  My guess was right anyway :).

Well, I don't think it's necessary. Even I think on contrary, I don't know how to do (and have never done such thing).
Let's keep as it like other sub-folder.

Yaron10:
@donho,

> So you are for or against Change History commands being disabled when there's no more change history?

The advantage is clear.
Is it worth the redundant calls on every SCN_UPDATEUI? I thought it was not.
But you are the pro and I respect your opinion.

> My guess was right anyway :).

👍 :)

> Well, I don't think it's necessary...

👍 
Thank you again.

donho:
@Yaron10 
> Wouldn't your initial implementation (+ a checkUndoState() call) be better?

As I explained, when there's no impact of performance, I prefer the easier maintained code.  
Here all the enableCommand calls are in checkUndoState() and it's easier to maintained IMO.
If one day there's a performance issue because of checkUndoState(), I'll consider to do the refactoring.

> After executing "Clear Change History" the Undo button is still enabled. Is it checkUndoState()?

I cannot reproduce with the binary of the latest commit.

Yaron10:
@donho,

> As I explained, when there's no impact of performance, I prefer the easier maintained code.

OK. I understand and respect that.

> I cannot reproduce with the binary of the latest commit.

You fixed that. Did I bring it up again?

Yaron10:
@donho,

There's actually a problem with the last commit's approach.

Open a saved file.
Edit.
Save.
Undo.

Result:
The Change History commands are wrongly disabled.

donho:
@Yaron10 
> Is it worth the redundant calls on every SCN_UPDATEUI? I thought it was not.

You're right!
But the 3 commands are need the state of "canUndo" non? Otherwise what's your solution to update the state of 3 commands when the document cannot undo anymore?

Yaron10:
@donho,

Do you see it as an issue?

> The Bookmarks go-to commands are not disabled if there are no bookmarks.
And the Style Jump Up/Down commands are not disabled if there are no highlights.

Sure, it would be better to disable them. But is it worth the extra code?

Would you consider one beep if there's no history? :)

```cpp
	if (line != -1)
	{
		_pEditView->execute(SCI_ENSUREVISIBLEENFORCEPOLICY, line);
		_pEditView->execute(SCI_GOTOLINE, line);
	}
	else
		::MessageBeep(MB_ICONEXCLAMATION);
```

donho:
@Yaron10 
Could you check the latest commit is what you expect?

Yaron10:
@donho,

Great. 👍 

One last note:

> Would you consider one beep if there's no history?

There's no beep on "Clear Change History", and it's somewhat inconsistent. Sorry about that.
Should the beep be removed?

donho:
@Yaron10 
The beep for me is a warning to tell users that "There is no change to search". So it's good to apply it to the "go to next/previous change" commands. Whereas "Clear change history" command is executed conscientiously by user to empty change history, so IMO no beep should be applied here.

Yaron10:
@donho,

👍 
Thanks again.

donho:
@Yaron10 
Thank you for your suggestions & test!

Yaron10:
@donho,

There's a minor issue with "Clear Change History".

Clone a dirty document to the other view.
"Clear Change History".

Result:
The History markers in the other view are visible until you activate it.

Thank you.

donho:
@Yaron10 
Could you create an issue for it? I'm in holidays but I'll try to fix it when I've got time.

Yaron10:
@donho,

Enjoy your holiday.

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13877.

Thank you.

PR #13869, opened 2023-07-06
Enable optimization flags & CET

### LINKED ISSUE #13855: Feature Request: Add Exploit mitigation techniques and Sandbox App for Greater Security
If possible, you could enable exploit mitigations such as CFG, ACG and CET.
This way it will get rid of classes of exploit vulnerabilities.

Microsoft is creating a new sandbox environment that is easy to adapt and implement which will strengthen the security even further.

For more information:
https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/customize-exploit-protection?view=o365-worldwide)
https://github.com/microsoft/win32-app-isolation

ozone10:
partially fix #13855


donho:
@ozone10 
Could you point me out the advantage of this PR please?

ozone10:
@donho 
optimization flags are already enabled implicitly, to avoid some confusion in Visual Studio settings this PR change them to be explicit.

For [CET](https://learn.microsoft.com/en-us/cpp/build/reference/cetcompat?view=msvc-170), I am using it for about 2 month without issue. It is hardware based protection and thus there should be no performance impact.

PR #13886, opened 2023-07-12
Fixed typo in changelog

Tex2002ans:
appearence -> appearance

### DIFF
--- a/PowerEditor/bin/change.log
+++ b/PowerEditor/bin/change.log
@@ -5,7 +5,7 @@ Notepad++ v8.5.4 bug-fixes and new features:
  3. Fix opening multi-files on multi-instance mode regression.
  4. Update: Scintilla 5.3.5 Lexilla 5.2.5
  5. Fix Notepad++ hanging issue running macro to EOF.
- 6. Fix EOL symbol color/appearence being reset issue while changing encoding.
+ 6. Fix EOL symbol color/appearance being reset issue while changing encoding.
  7. Enable code folding in Assembly source files.
  8. Fix document language not remembered through sessions issue.
  9. Add SHA-1 hash features.


chcg:
As 8.5.4 is already released it doesn't seem to make sense to fix a typo in the changelog as it is just used for this specific release.

Tex2002ans:
> This doesn't follow the contributing guidelines.
> Typos are merely to be pointed out, not fixed.

Ahh, I wasn't aware. Thanks!

Last night, I just:

- Updated to v8.5.4
- Saw the typo in initial popup
- Did a search on Github for the misspelled word
- then decided to actually do my first PR.

I didn't even think to check out the main page/rules.

- - -

So, according to [Contribution Guidelines](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md):

> If [...] **fixing typos, please create the issue instead of the pull requests.** **All Pull Requests,** except for translations and user documentation, **need to be attached to a issue on GitHub.**
>
> 7. Typo fixing and code refactoring won't be accepted - please create issues with title started with `TYPO` to request the changing.

So, the correct thing to do is to submit the errors as an Issue instead?

And then just list:

- appearence -> appearance

and their exact sentences?

- - -

> As 8.5.4 is already released it doesn't seem to make sense to fix a typo in the "old" changelog as it is just used for this specific release. For the upcoming new release the changelog will be overwritten.

Over the years, I noticed quite a few typos throughout many of the changelogs, (and then yesterday, I finally decided to actually do something about it!).

Usually, I might be a few versions behind, like:

- v8.5

so when I update, I visit each version's "Downloads" page and go through multiple "older" changelogs, like:

- [v8.5.1](https://notepad-plus-plus.org/downloads/v8.5.1/)
- [v8.5.2](https://notepad-plus-plus.org/downloads/v8.5.2/)
- [v8.5.3](https://notepad-plus-plus.org/downloads/v8.5.3/)
- [v8.5.4](https://notepad-plus-plus.org/downloads/v8.5.4/)

to see exactly what changed or got updated.

I'm suspecting just doing one spellchecking/grammarchecking pass across all changelog text would be needed, and you'd squash many of those "old" typos once and for all.

If I just get pointed in the right direction, I wouldn't mind correcting them all in one shot. Thanks. :)

alankilborn:
>So, the correct thing to do is to submit the errors as an Issue instead?

Yes.

But:  @chcg had some wisdom here: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13886#issuecomment-1632953189

PR #13891, opened 2023-07-12
Fix 'Generate MD5 digest from files' 1st dlg show (v855 regression)

xomx:
@donho 
During testing of my new translation I have found a regression in the new v8.5.5 code. As it is not published yet, I will not create a new issue, ok? It was caused by the previous 'Add SHA-512 hash features' commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4ffd897ccf8d30d8dd945c42907e61381c590d08
The [dlg centering code](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ebdc910dcd7bf0577b8d54f6c6cb0fd2e1f6f703/PowerEditor/src/MISC/md5/md5Dlgs.cpp#L277) needed had been skipped for the 1st time only ...

### DIFF
--- a/PowerEditor/src/MISC/md5/md5Dlgs.cpp
+++ b/PowerEditor/src/MISC/md5/md5Dlgs.cpp
@@ -245,6 +245,13 @@ void HashFromFilesDlg::doDialog(bool isRTL)
 
 		switch (_ht)
 		{
+			case hash_md5:
+			{
+				title = TEXT("Generate MD5 digest from files");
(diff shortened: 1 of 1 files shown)


donho (commented):
Please add also here:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4ffd897ccf8d30d8dd945c42907e61381c590d08?diff=split#diff-c886868c635392447dbfa2bcd8a13fb6afb86e93578b3db1f7500d862359ec90R569

Issue #13164, opened 2023-02-18
Notepad++ v.8.4.9 Find & Replace with simple regular expressions doesn't work in file of 49 MB


KiaraValloti:
### Description of the Issue

Find & Replace with simple regular expressions doesn't work in file of 49 MB
### Steps to Reproduce the Issue

1. Click Search/Replace... 
2. Click Tab Find in Files
3. Find what (\d*);(\d*):(\d*):(\d*)
4. Replace with \1;\2:\3
5. Directory choose the directory where you unzipped the attached .csv file
6. Search Mode Regular expression
7. Click Replace in Files
8. You obtain 0 occurrences were replaced, you should obtain 948 occurrences
9. The same holds for tab Replace clicking Replace All
10. Open the file, cut off the first half of the text file, now its size is about 23 MB, repeat the experience on the last part, you obtain 948 occurrences
### Expected Behavior
You should obtain 948 occurrences of  replacement

### Actual Behavior
You should obtain 0 occurrences of  replacement

### Debug Information

Notepad++ v8.4.9   (64-bit)
Build time : Jan 27 2023 - 03:11:16
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "D:\all.ibdd\data\CME.GBP\CME.GBP_1.csv"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Home (64-bit) 
OS Version : 1803
OS Build : 17134.1130
Current ANSI codepage : 1252
Plugins : none

[CME.GBP_1.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10773812/CME.GBP_1.zip)

wallrik:
I tried replicating this, and there is indeed some type of error or timeout when I try your pattern.

However, if I change the regex to anchor at the start of each line, like so `^(\d*);(\d*):(\d*):(\d*)`, the replacement succeeds with 948 replacements. 👍

alankilborn:
>You obtain 0 occurrences were replaced

Do you happen to get this thing showing up when it occurs?:

![image](https://user-images.githubusercontent.com/77065706/219977552-6c54ebce-bb2f-4ccc-a794-0cac26440447.png)

KiaraValloti:
> > You obtain 0 occurrences were replaced
> 
> Do you happen to get this thing showing up when it occurs?:
> 
[...]

PR #13903, opened 2023-07-15
Reuse tooltip for Replace when Find fails with an exception

jofon:
In Find/Replace, when you press Replace and the Find step fails, the statusbar message is overriden with Replace's message. This unfortunately also replaces the tooltip if Find failed with an exception (see #13164). This commit reuses the tooltip, letting users get some more information on the problem, as they might not think about testing Find first.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2644,7 +2644,8 @@ bool FindReplaceDlg::processReplace(const TCHAR *txt2find, const TCHAR *txt2repl
 	{
 		NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
 		generic_string msg = pNativeSpeaker->getLocalizedStrFromID("find-status-replace-not-found", TEXT("Replace: no occurrence was found."));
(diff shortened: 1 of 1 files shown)


donho:
@jofon 
Though I see the intention from the code of PR, I cannot validate this PR (make this PR work on my PC with the privided file in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13164, running both release and debug binaries:

image

Do you have any other way to validate this PR?

jofon:
@donho 
This only adds the tooltip to the Replace tab. 
image

Previously, you'd only find the tooltip on the Find tab.

The Find In Files/Projects and Mark don't have tooltips of any kind, as they use different functions:
The Find and Replace tabs use FindReplaceDlg::processFindNext(), which sets tooltips, but the other tabs use FindReplaceDlg::processRange(), which doesn't.
So I didn't add it there as that would require several changes and a lot more work.

donho:
@jofon 

> So I didn't add it there as that would require several changes and a lot more work.


The current PR is clever but not evident: using the member variable as argument of function, while the same member variable takes the value of argument of function, is confusing.

I will be inspired by your solution and make find all functions return the error status for having accurate error message in find/replace/count/mark all functions.

Thank you for your contribution.

donho:
This PR is replaced by #13945, which provides more complex modification.

PR #13917, opened 2023-07-22
Ignore = at the end of a file name

### LINKED ISSUE #13344: Error while opening files whose name starts/ends with `=` (v8.5, 32b)
### Description of the Issue
Error while opening files whose name starts/ends with `=`

1. Create a file whose name starts and ends with `=`, e.g. `=NOTE=`
2. Open the file with this command: `"C:\Program Files\npp\notepad++.exe" "P:\ath\of\file\=NOTE="` (or similar, with the correct paths of course)
3. The program opens the file, but it also displays a "cannot open file" error box (see picture).

### Expected Behavior
(just open the file)

### Actual Behavior
(also an error messagebox)

### Debug Information
It's Notepad++ v8.5, 32 bit.
Doing the same as described above was not a problem with older versions, e.g. 6.7.9.2
immagine

jofon:
Opening a file that ends with '=', with quotes around the filepath, caused an extra quote to remain at the end of the string, as the code would enter the if on line 114, instead of line 133. Further on, the extra quote would be separated into an empty filepath by FileNameStringSplitter, and then opening would fail and present an error dialog (before the error dialog was introduced in #12613, the error would be silently ignored).

This commit prevents the code from entering the arg if (line 114) when it knows it's evaluating a file, and instead it'll go to the file if (line 133) and remove the extra quote.

Fix #13344

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -111,7 +111,7 @@ void parseCommandLine(const TCHAR* commandLine, ParamVector& paramVector)
 		{
 			case '\"': //quoted filename, ignore any following whitespace
 			{
-				if (!isStringInArg && i > 0 && cmdLinePtr[i-1] == '=')
+				if (!isStringInArg && !isInFile && i > 0 && cmdLinePtr[i-1] == '=')
 				{
 					isStringInArg = true;
 				}


PowerEditor/src/winmain.cpp
```diff
@@ -111,7 +111,7 @@ void parseCommandLine(const TCHAR* commandLine, ParamVector& paramVector)
 		{
 			case '\"': //quoted filename, ignore any following whitespace
 			{
-				if (!isStringInArg && i > 0 && cmdLinePtr[i-1] == '=')
+				if (!isStringInArg && !isInFile && i > 0 && cmdLinePtr[i-1] == '=')
```
donho, line 114: Change `isInFile` to `isBetweenFileNameQuotes`.

donho (changes requested):
The variable was named badly.
Please rename it as suggestion to make the intention more clear.

PR #13931, opened 2023-07-26
fix search results buffer collision with new buffs

### LINKED ISSUE #13636: Notepad++ fails to jump to search hits initiated by "Find All in" (Current Document/All Opened Documents) searches for the first, new, unsaved document under certain conditions.
### Description of the Issue
Notepad++ fails to jump to search hits initiated by "Find All in" (Current Document/All Opened Documents) searches for the first, new, unsaved document under certain conditions.

The two conditions:
* A saved document must be opened by NP++ at some point.
* A "Find All in" (Current Document/All Opened Documents) search for any search term must be initiated at some point.

After these two condtions are met:
* The first new, unsaved document that is created anytime after a saved document is opened in NP++ will be ignored by any double-clicking (jump to behavior) of search results for any subsequent "Find All in" (Current Document/All Opened Documents) searches.

1. Open Notepad++ ensuring that there are no open documents other than the default "new 1".
2. File > Open... "[needles in the haystack.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/11432007/needles.in.the.haystack.txt)" and select "Open".
3. Search > Find... "needle" (without quotes and Search Mode = Normal) and select "Find All in Current Document".

The Search results window will show a listing of 5 hits (1 file).

4. Double-click any of the hits from the Search results window and NP++ jumps to the location of the result within the document. **_This is expected behavior._**
5. Edit > Select All (Ctrl+A) from the content of the "needles in the haystack.txt" document.
6. Edit > Copy (Ctrl+C or Ctrl+INS).
7. File > New (Ctrl+N). A new document tab, "new 1", appears.
8. Edit > Paste (Ctrl+V or Shift+INS) into "new 1". Now the content from "needles in the haystack.txt" is duplicated into this new, unsaved document "new 1".
9. With the Search results window still up, right-click and select "Clear all".
10. Search > Find... "needle" (without quotes and Search Mode = Normal) and select "Find All in All Opened Documents".

The Search results window will show a listing of 10 hits (2 files).

11. Double-click any of the hits from the Search results window and NP++ jumps to the location of the result within "needles in the haystack.txt" but double-clicking any of the hits from "new 1" will fail to even activate the view of the document and/or jump to the location. **_This is unexpected behavior._**

Addendum:
If you repeat steps 7 and 8 several times for further new documents ("new 2", "new 3", "new 4") then proceed with steps 9 through 11. The behavior of jumping to the correct location works for all of the documents except for the first, new, unsaved document created after our opened document ("new 1" in our example).

### Expected Behavior
Double-clicking any of the hits from the Search results window will result in NP++ jumping to the correct doucment and line of the search hit.

### Actual Behavior
Double-clicking any of the hits from the Search results window will result in NP++ jumping to the correct doucment and line of the search hit with the exception of the first, new, unsaved document created after opening a saved document.

### Debug Information
Notepad++ v8.5.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Build : 7601.0
Current ANSI codepage : 1252

**_UPDATE:_** This bug seems to have been introduced in Notepad++ v8.5 as the last verison to exhibit expected behavior was in Notepad++ v8.4.9.

molsonkiko:
fix #13636.

To test:
1. Open a file containing at least two instances of some string (say, `foo`)
2. Open a file containing the same string (or just copy/paste into a new buffer)
3. Use `Find All in All Opened Documents` and `Find All in Current Document` and verify that they still work as expected (navigation is possible between matches within the same document, and also between different documents)
4. Use Ctrl+N to create a new buffer, and copy/paste the text of the first file into it.
5. Use `Find All in All Opened Documents` and `Find All in Current Document` and verify that they still work as expected (navigation is possible between matches within the same document, and also between different documents)
6. Also verify that `Use `Find All in All Opened Documents` and `Find All in Current Document` can search within the new file (the first file opened after the search results dialog was first opened)
7. Repeat steps 4 and 5 at least once.

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -346,8 +346,9 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// returned value : TRUE if this function call is successful and shortcut is enable, otherwise FALSE
 
 	#define NPPM_DOOPEN (NPPMSG + 77)
-	// BOOL NPPM_DOOPEN(0, const TCHAR *fullPathName2Open)
+	// BOOL NPPM_DOOPEN(BOOL onlyInEditViews, const TCHAR *fullPathName2Open)
 	// fullPathName2Open indicates the full file path name to be opened.
+	// if onlyInEditViews is true, only files in the two main edit views (not buffers associated with some forms) should be considered.
 	// The return value is TRUE (1) if the operation is successful, otherwise FALSE (0).
 
 	#define NPPM_SAVECURRENTFILEAS (NPPMSG + 78)
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -161,7 +161,7 @@ public:
 	//! \name File Operations
 	//@{
 	//The doXXX functions apply to a single buffer and dont need to worry about views, with the excpetion of doClose, since closing one view doesnt have to mean the document is gone
-	BufferID doOpen(const generic_string& fileName, bool isRecursive = false, bool isReadOnly = false, int encoding = -1, const TCHAR *backupFileName = NULL, FILETIME fileNameTimestamp = {});
+	BufferID doOpen(const generic_string& fileName, bool isRecursive = false, bool isReadOnly = false, int encoding = -1, const TCHAR *backupFileName = NULL, FILETIME fileNameTimestamp = {}, bool onlyInEditViews = false); // onlyInEditViews means if a file with that name is open, open it only if it's in an edit view.
 	bool doReload(BufferID id, bool alert = true);
 	bool doSave(BufferID, const TCHAR * filename, bool isSaveCopy = false);
 	void doClose(BufferID, int whichOne, bool doDeleteBackup = false);
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -437,7 +437,8 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		case NPPM_DOOPEN:
 		case WM_DOOPEN:
 		{
-			BufferID id = doOpen(reinterpret_cast<const TCHAR *>(lParam));
+			bool onlyInEditViews = wParam == 1;
+			BufferID id = doOpen(reinterpret_cast<const TCHAR*>(lParam), false, false, -1, NULL, {}, onlyInEditViews);
 			if (id != BUFFER_INVALID)
 				return switchToFile(id);
 			break;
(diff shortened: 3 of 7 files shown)


molsonkiko:
It turns out that the issue is more complex than I thought, because [`Document Map` can cause the same problem](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13636#issuecomment-1651052643). As such, __my PR is not currently acceptable.__

It looks like the problem may be that [FileManager::getBufferFromName still uses the old test for mapping buffer names to buffer ids](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/3bdefd88d0e227d8b5444318855e48b55b13fbe0/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1712).

A possible solution would be to change `FileManager::getBufferFromName` and add an optional boolean parameter, false by default, that restricts the search to only files in the two main edit views if true. I'll try working on that tomorrow.

Yaron10:
@molsonkiko,

Thank you for your work. 👍 

Allow me two questions:
1. Why is `onlyInEditViews` needed?
2. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13204#issuecomment-1443690407.
Wouldn't it better to NOT to set "New X" to the Search-Results & Document Map? Would that require a more serious refactoring?

molsonkiko:
> 1. Why is `onlyInEditViews` needed?

The issue that this PR endeavors to solve is that #13204 makes it so that the Document Map and Search Results forms will be linked to buffers that have the same name as buffers in the main edit views (i.e., the buffers that the user thinks of as "real buffers"). The `onlyInEditViews` parameter makes it possible to restrict the search of `WM_DOOPEN` or `NPPM_DOOPEN` to only files in the main edit views, which circumvents the problem created by #13204.

> Wouldn't it better NOT to set "New X" to the Search-Results & Document Map? Would that require a more serious refactoring?

Well... it *might* be better. I considered this myself. This would require a change to `nextUntitledNewNumber`. Essentially, a good solution from this angle might look something like this:
```
if the file is in one of the main edit views:
     name it `new N`, where N is the lowest number greater than 1 not already in use for `new i` documents in the main edit views
else:
     name it `new M`, where M is the highest number less than or equal to INT_MAX not already in use for `new i` documents NOT in the main edit views.
```
Essentially, the documents in the main edit views would be named `new 1`, `new 2`, `new 3`, ...
and the documents NOT in the main edit views (the search results and Document Map buffers among others) would be named `new 2147483647`, `new 2147483646`, `new 2147483645`, ...

Since nobody is very likely to manually change a document's name to `new 2147483647`, this would also fix the PR while not requiring too much code. Importantly, it would be more robust to changes in the codebase than simply changing the initialization code for the search results form and Document Map.

I'd rather not change my PR again, but if Don Ho thinks that alternative or something like it is a better solution, I'm happy to rework it.

Yaron10:
@molsonkiko,

> Why is onlyInEditViews needed?

I meant to ask if and when `onlyInEditViews` should be `false`.
If I understand it correctly, the fix is adding `if (isVisibleInEditViews(buf))` to `getBufferFromName()` (which is very nice).
Shouldn't this be the only change? 

Well, let's wait for Don's review.

donho:
#13636 is fixed by 891e8f936e03586b0bbf2850523a4af72701f622 which reverts 892ab0850fdfefe4f335e0977745e06b7a6ac4c8 due to it causes #13636.

PR #13945, opened 2023-07-29
Fix inaccurate find/replace in files result while using invalid regexp

### LINKED ISSUE #13164: Notepad++ v.8.4.9 Find & Replace with simple regular expressions doesn't work in file of 49 MB

donho:
Fix #13164


Yaron10:
@donho,

I hardly use RegEx. I've only tested the case in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13164.
It seems to work like a charm.  👍
Thank you. 

We currently have `<find-status-invalid-re value="Find: Invalid regular expression"/>`.
Can't the new `find-dlg-status-invalid-regexpr` replace it?
And what about the `---re-malformed` occurrences?

***

@jofon,

Thank you for the inspiration. 👍

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3649,6 +3665,20 @@ void FindReplaceDlg::setStatusbarMessage(const generic_string & msg, FindStatus
 	}
 }
 
+void FindReplaceDlg::setStatusbarMessageWithRegExprErr(ScintillaEditView* pEditView)
+{
+	if (!pEditView)
+		return;
+
+	char msg[511] = "";
```
Yaron10, line 3673: Just curious:
Is there a difference between `char msg[511] = "";` and `char msg[511]{};`?

donho: @Yaron10 
I do believe that, in `char msg[511] = "";` we declare a char array of 511 bytes and make the msg[0] = '\0';
whereas is `char msg[511]{};`, it fullfills 511 bytes with `\0`. 

Please let me know (anyone) if my understanding is wrong.
Therefore it's better to use `char msg[511]{};`.

I will do  another commit soon to fix it.

donho:
@Yaron10 
> 

> We currently have `<find-status-invalid-re value="Find: Invalid regular expression"/>`. Can't the new `find-dlg-status-invalid-regexpr` replace it? And what about the `---re-malformed` occurrences?

Finally we will use `find-status-invalid-re` instead of  `find-dlg-status-invalid-regexpr`.

> **EDIT:** The `---re-malformed` occurrences are apparently replaced by `find-dlg-status-invalid-regexpr`.

They will be removed.

Thank you for your review!

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
Yaron10: @donho,

Just some thoughts.

If your theory is correct - is there a reason to initialize all 511 bytes?
I understand it might be safer, but there must be some overhead to that robust initialization.
(Relevant to many other cases across the project).

In `POINT pt{}` (or `RECT rc{}`), there's no difference between `pt.x` and `pt.y`. - If it's better to initialize, both variables should be initialized.
But here we can check `msg[0]`. And I _suppose_ there's no warning if `char msg[511] = "";` is used.

> Please let me know (anyone) if my understanding is wrong.

@ozone10, @rdipardo.

Yaron10:
@donho,

Great. 👍 
Thanks again for your work.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: @Yaron10 
I could be wrong, but IMO, there's no overhead with modern C++ compiler initialization of `type var[512] {}` - I assume that compiler generate an array with all zero by default to you.

Yaron10: @donho,

> but IMO, there's no overhead with modern C++ compiler initialization of type var[512] {}

This is above my level.
I just thought there must be _some overhead_ at runtime.
But you're the pro, and I trust your judgment. :)

> I assume that compiler generate an array with all zero by default to you.

I use MS VS.

Yaron10:
@donho,

`if (!hasInvalidRegExpr && (_mainWindowStatus & WindowSubActive))`

Is there a reason not to use that in `Notepad_plus::findInOpenedFiles()` as well?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
rdipardo: @Yaron10 
> And I _suppose_ there's no warning if `char msg[511] = "";` is used.
> 
> > Please let me know (anyone) if my understanding is wrong.
> 
Since you asked, `char msg[511] = ""` is not only a warning but a hard failure.

`""` is a __*pointer*__ to an array of `char`, but `msg[511]` is an actual `char`:

    error C2440: '=': cannot convert from 'const char [1]' to 'char'

Yaron10: @rdipardo,

Thank you for replying.

> error C2440: '=': cannot convert from 'const char [1]' to 'char'

So how did the first commit not fail? :)

rdipardo:
> `if (!hasInvalidRegExpr && (_mainWindowStatus & WindowSubActive))`
>
> Is there a reason not to use that in `Notepad_plus::findInOpenedFiles()` as well?

In fact, the whole block is guarded by `if (_mainWindowStatus & WindowMainActive)` in [line 1255].
If I'm reading this correctly, the `WindowSubActive` and `WindowMainActive` flags are mutually exclusive:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/674cf95f7d8ffc6f95440232b3831b0b3503680a/PowerEditor/src/Notepad_plus.cpp#L4403

So, if you know `WindowMainActive` has been set, don't you basically have a known condition?, i.e.

~~~cpp
if (!hasInvalidRegExpr && true)
{
// . . .
~~~

[line 1255]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/674cf95f7d8ffc6f95440232b3831b0b3503680a/PowerEditor/src/Notepad_plus.cpp#L1255

donho:
> Is there a reason not to use that in Notepad_plus::findInOpenedFiles() as well?

@Yaron10 
You're right about it. Fixed in the latest commit.
Thank you!

Yaron10:
@rdipardo,

> In fact, the whole block is guarded by if (_mainWindowStatus & WindowMainActive) in [line 1255](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/674cf95f7d8ffc6f95440232b3831b0b3503680a/PowerEditor/src/Notepad_plus.cpp#L1255).

I don't quite understand it. - There are two separate blocks there.

@donho,

👍 
Thank you!

***

I must retire now.
Good night.

donho:
> I don't quite understand it. - There are two separate blocks there.

@Yaron10 
It tests firstly (the 1st block) if main edit zone is active (ie. it's being used). If it is, then searches into files opened in the main edit zone. Then (the 2nd block) it does the same thing on the second edit zone.

Yaron10:
@donho,

I DID understand that. :)

I did NOT understand @rdipardo's "In fact, the whole block is guarded by if (_mainWindowStatus & WindowMainActive)".

Thanks again for your work. It's very nice!

***

@rdipardo,

Am I missing anything?

`_mainWindowStatus |= (whichOne==MAIN_VIEW)?WindowMainActive:WindowSubActive;`

It's **`|=`**.

rdipardo:
@Yaron10 
> Am I missing anything?
>
> `_mainWindowStatus |= (whichOne==MAIN_VIEW)?WindowMainActive:WindowSubActive;`
>
> It's **`|=`**.

Yes, the terms of logical OR do not depend on each other (the notation is `A + B`, where `B` does not change anything when it's 0). But when you have one condition nested inside another, it's logical to assume (as I wrongly did), that the inner condition depends on the outer one. 

In this case, you have `A AND (1 OR B)`, because we don't get inside unless `A` is true (`WindowMainActive`). But `A AND (1 OR B)` is the same as `A AND 1`, or just `A`, which has nothing to do with the `B` nested inside. This is why bit flags are usually handled by a switch-case.

Cf. [Why You Shouldn't Nest Your Code](https://www.youtube.com/watch?v=CFRhGnuXG-4)

Yaron10:
@rdipardo,

PR #13960, opened 2023-08-04
Fix new document number jumps from "New 1" to "New 3" issue

### LINKED ISSUE #8677: NPP jumps from "New 1" to "New 3" if "Find Results" is displayed

### LINKED ISSUE #13194: NPP jumps from "new 1" to "new 3" when document map is visible

donho:
Solution from:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13931#issuecomment-1653523758

Fix #13194, fix #8677

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -7788,6 +7788,7 @@ static const QuoteParams quotes[] =
 	{TEXT("Anonymous #203"), QuoteParams::rapid , false, SC_CP_UTF8, L_TEXT, TEXT("My gynecologist follows me on Instagram, I really do not know what else he want to see.\n\n") },
 	{TEXT("Anonymous #204"), QuoteParams::slow , false, SC_CP_UTF8, L_TEXT, TEXT("The greatest security vulnerability in any computer system is located between the keyboard and the chair.\n\n") },
 	{TEXT("Anonymous #205"), QuoteParams::slow , false, SC_CP_UTF8, L_TEXT, TEXT("Courage is knowing it might hurt, and doing it anyway.\nStupidity is the same.\nAnd that's why life is hard.\n\n") },
+	{TEXT("Anonymous #206"), QuoteParams::rapid , false, SC_CP_UTF8, L_TEXT, TEXT("RegEx can save 30 minutes of your dev time with only 10 characters.\nBut you have to spend more than 40 minutes to figure out what these 10 characters need to be.\n\n") },
 	{TEXT("xkcd"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("Never have I felt so close to another soul\nAnd yet so helplessly alone\nAs when I Google an error\nAnd there's one result\nA thread by someone with the same problem\nAnd no answer\nLast posted to in 2003\n\n\"Who were you, DenverCoder9?\"\n\"What did you see?!\"\n\n(ref: https://xkcd.com/979/)") },
 	{TEXT("A developer"), QuoteParams::slow, false, SC_CP_UTF8, L_TEXT, TEXT("No hugs & kisses.\nOnly bugs & fixes.") },
 	{TEXT("Elon Musk"), QuoteParams::rapid, false, SC_CP_UTF8, L_TEXT, TEXT("Don't set your password as your child's name.\nName your child after your password.") },
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -50,6 +50,8 @@
 #define MENU 0x01
 #define TOOLBAR 0x02
 
+#define MAIN_EDIT_ZONE true
+
 enum FileTransferMode {
 	TransferClone		= 0x01,
 	TransferMove		= 0x02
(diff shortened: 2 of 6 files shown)


donho:
@Yaron10
Thank you for your solution.
I will check #11704 later to see what I can do about it.

donho:
@Yaron10 
Please check the latest commit which responds your remark here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13344#issuecomment-1665731435

Yaron10:
@donho,

We might have some misunderstanding here.

1. With the last commit - double-clicking an entry in the Search Results won't jump to a tab named "new".
2. If https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13344#issuecomment-1665619119 is fixed, there's no need to change the first commit.
3. If you agree with @alankilborn's "Maybe it should be given a better title, even if there is no purpose (user wouldn't see it) other than to avoid potential "N++ programmer" confusion at some future point.", how about `std::wstring newTitle = L"newNonMainEditZone ";` _with a space at the end_?

Thank you.

donho:
@Yaron10 
>With the last commit - double-clicking an entry in the Search Results won't jump to a tab named "new".

True! Nice caught!

> If https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13344#issuecomment-1665619119 is fixed, there's no need to change the first commit.

Do you mean the "bug" that allow users to add spaces in front & at the end of the tab name when they try to rename a  tab?

>  how about std::wstring newTitle = L"newNonMainEditZone "; with a space at the end?

Good idea. `newNonMainEditZoneInvisibleTitle ` for ensuring there won't be any collision.

Yaron10:
@donho,

> Do you mean the "bug" that allow users to add spaces in front & at the end of the tab name when they try to rename a tab?

Yes.

`newNonMainEditZoneInvisibleTitle `.

That's better. 👍
How about a short comment explaining the space? 

Thank you again. I appreciate it.

Yaron10:
@donho,

Very nice! 👍 

Thank you.

Yaron10:
@donho,

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f6e1b2cab6b42530869f600121443350f0292ad6/PowerEditor/src/WinControls/Window.h#L27

In MS Visual Studio **2017** I need to comment that line or change it to

```cpp
Window(const Window&) = default;
```

Otherwise, I get the `function was explicitly deleted` & `attempting to reference a deleted function` errors.

What do you think?

Thank you.

donho:
@Yaron10 
It doesn't make sense.
By searching ` = delete;` in the project, a lot results are found...
Or the constructor `Window(const Window&)` is really in use? If you can prove it, then this line should be removed.

Yaron10:
@donho,

Thank you for replying.

I know it works for you.
But I have a problem with the following lines:

```cpp
	ScintillaEditView _subEditView = ScintillaEditView(MAIN_EDIT_ZONE);	// only _mainEditView and _subEditView are MAIN_EDIT_ZONE comparing with other Scintilla controls
	ScintillaEditView _mainEditView = ScintillaEditView(MAIN_EDIT_ZONE);	// only _mainEditView and _subEditView are MAIN_EDIT_ZONE comparing with other Scintilla controls
```

תמונה

I'll have to figure it out.

donho:
@Yaron10 
In my opinion, it's one of limitation/bug in MS Visual Studio 2017. 
Couldn't you upgrade your Visual Studio to a more recent version?
By upgrading it, you keep always your MS Visual Studio 2017 for the other projects and you benefit especially all the support of modern C++ syntax of which Notepad++ will use in the future without doubt.

Yaron10:
@donho,

> In my opinion, it's one of limitation/bug in MS Visual Studio 2017.

That was my assumption too.
Thank you for confirming. :+1: 

> Couldn't you upgrade your Visual Studio to a more recent version?

Yes, I should do that.
Currently my Visual Studio is installed on an old Win 7 PC.
I might install the latest version on my Win 10 machine.

PR #13968, opened 2023-08-05
Fix leading & tailling spaces being allowed after renaming tab issue

### LINKED ISSUE #13967: Leading & tailling spaces should not be allowed after renaming tab
### Description of the Issue
Leading & tailling spaces should not be allowed after renaming tab

1. Renaming a tab by right clicking on a "new 1" tab for having the context menu
2. Choose "Rename"
3. In "New Name" field, type `   whatever   `, then  click OK button

### Expected Behavior
`whatever` as tab new name

### Actual Behavior
`   whatever   ` as tab new name

### Debug Information
```
Notepad++ v8.5.5   (64-bit)
Build time : Aug  2 2023 - 23:40:03
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.2070
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

donho:
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13344#issuecomment-1665687932

Fix #13967

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1383,6 +1383,7 @@ Continue?"/><!-- HowToReproduce: when you openned file is modified and saved, th
 			<FileLockedWarning title="Save failed" message="Please check whether if this file is opened in another program"/>
 			<FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/><!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
 			<RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use on another tab."/><!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
+			<RenameTabTemporaryNameIsEmpty title="Rename failed" message="The specified name cannot be empty, or it cannot contain only space(s) or TAB(s)."/><!-- HowToReproduce: Rename the tab of an untitled document and provide an empty string or only some white speces. -->
 			<DeleteFileFailed title="Delete File" message="Delete File failed"/><!-- HowToReproduce: this message prevents from system failure. It's hard to reproduce. -->
 
 			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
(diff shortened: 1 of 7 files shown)


Yaron10:
@donho,

Great. 👍 
Thank you for the quick fix.

Change a new tab name to "" (empty).
Open the Find dialog and "Find All in Current Document".
Double-click an entry in the Search Results window.

Result:
An error message: _Cannot open file ""_.

So an empty title (after trimming) should not be allowed either. :)

Yaron10:
@donho,

👍 
Thank you again for your time and work.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/391f4281efac81649a02aa8cdd4e765fa76d1237

alankilborn:
@donho @Yaron10 

Try this STR:

* Edit a localization file, e.g. `English_customizable.xml`
* Set this tag/value pair: `<tab-untitled-string value="         new "/>`
* Save the file
* Activate this Localization
* Press Ctrl+n a few times to create new tabs
* See that the new tabs have leading spaces

Yaron10:
@alankilborn,

Do you consider those spaces an issue? Consistency?
If we didn't have the _Cannot open file_ error, why shouldn't spaces be allowed?
(Now that an empty tab-title is not allowed, `std::wstring newTitle` could _theoretically_ be empty for NonMainEdit Scintillas and we shouldn't have any collisions).  

And even if spaces should not be allowed, is it worth the extra trimming? - Translators shoudn't add spaces to that string. :)
IIRC, NPP can crash if you manually edit config.xml with some wrong/unexpected values.

alankilborn:
>Do you consider those spaces an issue? Consistency?

Using the string I supplied, user could have a tab with name `         new 2`.  If they then rename the tab with intention on having `         new 2a`, result will be `new 2a`.  Confusing for user as to why that happened?  I'd say yes.  I think it is better to enforce consistency.

>Translators shoudn't add spaces to that string.

I suppose not, but I've long used english_customizable for my own purposes.

---------------------

My thoughts were:  After forming a potential new tab name via `tab-untitled-string`'s value, the result should be subjected to the same trimming as if a user renamed a tab.

Perhaps it isn't worthy of consideration.

Yaron10:
>  I think it is better to enforce consistency.
...
After forming a potential new tab name via tab-untitled-string's value, the result should be subjected to the same trimming as if a user renamed a tab.

I "love" consistency, and it's certainly a nice case.
@donho might agree with you that it IS worth the extra trimming.

PR #13985, opened 2023-08-09
Add indicator allocation for plugins

### LINKED ISSUE #5744: [Feature request] Add new plugins API messages for getting/releasing Scintilla indicator number
### Description of the Issue

There are currently several plugins that use Scintilla [indicators](https://www.scintilla.org/ScintillaDoc.html#Indicators) but there is no way for a plugin to know which indicator number is already used by another plugin so an unwanted plugins "cross-talk" happens as they track indicators placed by another plugin.

To name several plugins using indicators - DSpellCheck, DoxyIt, Compare

It appears that Compare v2.0 and DoxyIt plugins currently use the same indicator number.

1. Install both plugins.
2. Compare two files that have several lines differing by just one word (so you have changed lines, not only added/removed)
3. Press `TAB` key (DoxyIt shortcut for visiting function description fields)

DoxyIt will visit the Compare changed highlight indicators as well because the Compare's indicator number is the same as DoxyIt's.

This is just an example of the problem. It might appear in the future in different forms depending on the plugins.

### Suggestion
Add new NPPM_GET_SCINT_INDICATOR and NPPM_FREE_SCINT_INDICATOR plugin API messages so each plugin gets a different indicator number.
Of course Notepad++ should keep a list of the free indicator numbers.

alankilborn:
Fix #5744 

Note that the issue calls for ability to "release" indicator allocations, but since releasing isn't possible in N++ for command id allocation or marker allocation, it was not done for new indicator allocation.

### DIFF
--- a/PowerEditor/src/resource.h
+++ b/PowerEditor/src/resource.h
@@ -418,6 +418,9 @@
 #define MARKER_PLUGINS          1
 #define MARKER_PLUGINS_LIMIT   15
 
+#define INDICATOR_PLUGINS          9  // indicators 8 and below are reserved by Notepad++ (URL_INDIC=8)
+#define INDICATOR_PLUGINS_LIMIT   20  // indicators 21 and up are reserved by Notepad++ (SCE_UNIVERSAL_FOUND_STYLE_EXT5=21)
+
 
 
 //#define IDM 40000
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -368,7 +368,7 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 
 	#define NPPM_ALLOCATEMARKER  (NPPMSG + 82)
     // BOOL NPPM_ALLOCATEMARKER(int numberRequested, int* startNumber)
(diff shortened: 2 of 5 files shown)


Yaron10:
@alankilborn,

👍 
Thank you for your work.
(I haven't tested it).

PowerEditor/src/resource.h
```diff
@@ -418,6 +418,9 @@
 #define MARKER_PLUGINS          1
 #define MARKER_PLUGINS_LIMIT   15
 
+#define INDICATOR_PLUGINS          9  // indicators 8 and below are reserved by Notepad++ (URL_INDIC=8)
+#define INDICATOR_PLUGINS_LIMIT   20  // indicators 21 and up are reserved by Notepad++ (SCE_UNIVERSAL_FOUND_STYLE_EXT5=21)
```
donho, line 422: So the available indicators are from 9 to 20. Do you confirm?

donho (approved):

PowerEditor/src/resource.h
alankilborn: >So the available indicators are from 9 to 20. Do you confirm?

Correct; that is what I was trying to convey with my code plus comments.

alankilborn:
PythonScript test code for exercising the allocation of indicators may be found here: https://github.com/bruderstein/PythonScript/issues/303#issuecomment-1676452472

Issue #10266, opened 2021-07-27
Clone view creates two incompatible copies of the same file


lmattcris:
### Description of the Issue
Creating the clone of a file in the second view creates a possibly different file in the two views, if changes are added to two views separately. You can't then save any of the two views, without loosing the changes in the other view.
This looks to happen when after cloning you close/reopen the application and make different changes in the left and right views: changes in the right view aren't replicated in the left one and vice versa. Adding many changes you end up with two different versions of the same file, very hard to merge.

### Steps to Reproduce the Issue
1. Open a (long) text file in the main view 
2. Right click on file name tab and "Clone to other view"
3. Make a few changes in **left** pane, you'll see them appear in right pane
4. Make a few changes in **right** pane, still they will be updated in left one
5. **Close** Notepad++ with the main window close button, **without** saving anything
6. Reopen Notepad++, changes are _apparently_ preserved
7. Any change that you'll make from now on on left or right pane won't be replicated in the other one: file name is still the same, but you're working on two different copies (not cloned, not linked anymore)

### Expected Behavior
I just expected that reopening the application the link between the two windows is preserved, while it's not, even though they claim to be the same file on-disk

### Actual Behavior
I ended up with left pane having only the changes made in it, after the restart of the application. The same happens to right pane, and I can't track anymore all the changes as they're logged separately.

### Debug Information
Notepad++ v8.1.1   (32-bit)
Build time : Jul  1 2021 - 14:04:28
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 2009
OS Build : 19043.1110
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll

molsonkiko:
I encountered this just now on the current master (post-8.5.4). It's a bummer that this hasn't gotten more attention.

While the (*unfortunately numerous*) other bugs with document cloning could reasonably be put on the back burner, this is a __very serious problem__ with cloning that makes it borderline unusable.

I'm inclined to try fixing this myself.

molsonkiko:
A possible fix that occurs to me: add a `hasClone="yes/no"` attribute to the `<File>` elements in `session.xml`.

[...]

PR #14003, opened 2023-08-15
prevent reloading two incompatible copies of same file, fix issue #10266

molsonkiko:
Fix issue #10266, concerning unsaved cloned documents being reloaded.

When a buffer's status is saved in config.xml, the file manager checks if it is open in both edit views. If so, a hasClone attribute for that file's session.xml entry is set to `yes`.

When the session is loaded, all buffers in the main edit view are loaded as normal, and then when buffers in the subeditview are loaded, any that have the hasClone attribute check for a same-name file in the main edit view and clone it into the subeditview.

TO TEST:
1. Create a `new 1` document in one edit view, then make an edit.
2. Clone `new 1` into another view.
3. Open a real file (not a `new X` file) and clone it into another view, then edit one of the clones. DO NOT SAVE THE FILE.
4. Close the session.
5. Reopen the session.
6. Try editing the `new 1` document that was loaded from session. These edits should be mirrored in the copy in the other edit view.
7. Try editing the real file, and verify that it too is cloned, not split into two incompatible versions.

__CAVEAT:__
Arguably it would be preferable for Notepad++ to assume that *any time a document with the same name is open in both edit views, those two copies are clones.* The solution I have chosen *does not do that*, which means that if you clone a document in an older version of Notepad++, dirty the clones, and reopen those files in a version of Notepad++ with this fix, those will still be treated as two incompatible copies of the same document.

Presumably an alternate way to solve the problem would be to do something about the way that `Notepad_plus::loadSession` decides whether to clone a document in the subEditView when loading from session.xml. As it stands, it seems that this process involves checking to see if the buffer is dirty, and maybe that check should be eliminated.


Yaron10:
@molsonkiko,

👍 
Thank you for your work.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -6195,6 +6195,7 @@ void Notepad_plus::getCurrentOpenedFiles(Session & session, bool includUntitledD
 			const wchar_t* langName = languageName.c_str();
 			sessionFileInfo sfi(buf->getFullPathName(), langName, buf->getEncoding(), buf->getUserReadOnly(), buf->getPosition(editView), buf->getBackupFileName().c_str(), buf->getLastModifiedTimestamp(), buf->getMapPosition());
 
+			sfi._hasClone = MainFileManager.bufferHasClone(buf);
```
Yaron10, line 6198: @molsonkiko,

How about `sfi._hasClone = (k == 1 && _mainDocTab.getIndexByBuffer(buf) != -1)`?

(I haven't tested this line).

molsonkiko: Good idea.
While we're here, one thing that troubles me about this code block is the fact that the variable `size_t k` is shadowed a few lines earlier in the block `if (languageName.empty())`. This is a case of code smell at the very least, and at worst a potential bug.

I think this problem should be solved (by renaming the looping variable in the `if (languageName.empty())` block) whether or not my PR is accepted, but especially if I'm going to make use of `size_t k` here.

Yaron10: @molsonkiko,

Isn't `k` within the `if (languageName.empty())` block a completely different variable in a different scope?
But your point is good from a readability point of view, and I suppose Don would agree to replace that `k`.

PowerEditor/src/Parameters.h
```diff
@@ -207,6 +207,7 @@ struct sessionFileInfo : public Position
 	int	_encoding = -1;
 	bool _isUserReadOnly = false;
 	bool _isMonitoring = false;
+	bool _hasClone = false;
```
donho, line 210: New member `_hasClone` is not necessary.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1756,3 +1756,19 @@ size_t FileManager::docLength(Buffer* buffer) const
...
+	bool isInEditView = false;
+	bool isInNonEditView = false;
+	for (ScintillaEditView* referee : buf->_referees) {
+		if (referee == _pNotepadPlus->_pNonEditView)
+			isInNonEditView = true;
+		else if (referee == _pNotepadPlus->_pEditView)
+			isInEditView = true;
+	}
+	return isInEditView && isInNonEditView;
+}
```
donho, line 1774: New method `bufferHasClone` is not necessary.

PowerEditor/src/Parameters.cpp
```diff
@@ -3545,6 +3550,7 @@ void NppParameters::writeSession(const Session & session, const TCHAR *fileName)
 				(fileNameNode->ToElement())->SetAttribute(TEXT("originalFileLastModifTimestamp"), static_cast<int32_t>(viewSessionFiles[i]._originalFileLastModifTimestamp.dwLowDateTime));
 				(fileNameNode->ToElement())->SetAttribute(TEXT("originalFileLastModifTimestampHigh"), static_cast<int32_t>(viewSessionFiles[i]._originalFileLastModifTimestamp.dwHighDateTime));
 				(fileNameNode->ToElement())->SetAttribute(TEXT("tabColourId"), static_cast<int32_t>(viewSessionFiles[i]._individualTabColour));
+				(fileNameNode->ToElement())->SetAttribute(TEXT("hasClone"), viewSessionFiles[i]._hasClone ? TEXT("yes") : TEXT("no"));
```
donho, line 3553: No need to memorize whether if file is cloned.

PowerEditor/src/Notepad_plus.cpp
donho, line 6198: The detection of cloned buffer is not necessary here either.

PowerEditor/src/NppIO.cpp
```diff
@@ -2246,6 +2246,19 @@ bool Notepad_plus::loadSession(Session & session, bool isSnapshotMode, bool shou
...
+			// it appears that the user had this document cloned when they last saved the session
+			// if it's already open in the main edit view, we'll just clone it over from there.
+			BufferID copyInMainView = _mainDocTab.findBufferByName(pFn);
+			if (copyInMainView != BUFFER_INVALID)
+			{
+				loadBufferIntoView(copyInMainView, SUB_VIEW);
+				++k;
+				continue;
+			}
+		}
```
donho, line 2260: Detect here if the same file path can be found in main view - since user cannot open the same file in 2 views at the same time, except by mean of cloning.
So 
```cpp
			BufferID copyInMainView = _mainDocTab.findBufferByName(pFn);
			if (copyInMainView != BUFFER_INVALID)
			{
				loadBufferIntoView(copyInMainView, SUB_VIEW);
				++k;
				continue;
			}
```
is all we need to fix the problem.

donho (changes requested):
Please follow the suggestions and remove unnecessary code.

molsonkiko:
@donho 

I really like your simplification of my PR! 😄 I'd just like to make one observation, which isn't really a criticism per se. 

Part of the reason why I made a seemingly needlessly complex PR is to deal with one edge case that I have encountered when building plugins: the `NPPM_INTERNAL_SETFILENAME` message, which can potentially create two buffers with the same name. In principle, there *might* be some kind of reason why someone would want to use this message to create two different buffers with the same name.

I'm not saying that using `NPPM_INTERNAL_SETFILENAME` to create two buffers with the same name is a *legitimate* activity that Notepad++ should try to support, but this is definitely something you should be aware of.

donho:
All the message come with `NPPM_INTERNAL` suffix are used internally and they not part of plugin API. All plugin API will keep the compatibility with all the versions of Notepad++, but not for internal used messages. No need to say about `NPPM_INTERNAL_SETFILENAME` message.

Yaron10:
@donho & @molsonkiko,

👍 
Thank you for the nice fix.

donho:
@molsonkiko 
Please do 1 commit instead of several ones next time - the build of CI is quite long, and your PR consume 2 times of resource then the others which penalize other PRs' build.

Issue #12305, opened 2022-10-08
Show NUL in textdocument


KL7000F:
In the current version, Notepad++ always shows me NUL. It was not before. I assume this is a bug.
Version 8.4.5

![image](https://user-images.githubusercontent.com/16374462/194716917-d9e2c345-2b5d-4f9f-8460-3470853a0079.png)

alankilborn:
Check for a `nppLogNulContentCorruptionIssue.xml` file on your system; if you have one and it is not zero length, please attach it to this issue.

KL7000F:
This file is emty.

ArkadiuszMichalski:
Duplicate of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133.

alankilborn:
@ArkadiuszMichalski I don't know...closing this probably doesn't get any new attention on a critical old problem, that likely seems still happening on a very recent release, despite "recent" fix attempts.  Of course, maybe it isn't the exact same problem, but it appears to be a serious data corruption issue.

KL7000F:
Can i help or testing anything to solve the problem?

ArkadiuszMichalski:
@KL7000F If you can find steps on how to reproduce this bug, it would certainly help to fix it.

KL7000F:
Basically, I didn't actively do anything. The files have existed for many years. In Editor / Wordpad the files show some "Space" (see screenshot).
I can't say for sure, but I think it was after resetting a Windows restore point. Win11 Upgrade 22H2 back to 21H2. A connection?

![image](https://user-images.githubusercontent.com/16374462/194765829-0bee4c2e-9b50-435a-b9d6-7d730b3997c5.png)

Debug Info:

Notepad++ v8.4.5   (64-bit)
Build time : Sep  3 2022 - 04:05:32
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Users\kl7000f\Desktop\testfile.txt"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 21H2
OS Build : 22000.1042
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

[testfile.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9741925/testfile.txt)

xomx:
[...]

Issue #11196, opened 2022-02-12
Notepad++ x86 on 64Bit Windows can't save files in C:\Windows\Sysnative\...


dodmi:
### Description of the Issue
- Notepad++ 8.1.6 x86 and later fail to save files in C:\Windows\Sysnative\... which points to C:\Windows\System32\ on 64 bit Windows systems. I noticed this trying to change my hosts file. This works fine in Notepad++ 8.1.5.

### Steps to Reproduce the Issue

1.  Open an elevated command prompt
2.  Run `PathToNPP\notepad++.exe -noPlugin C:\Windows\Sysnative\drivers\etc\hosts`
3. Change something
4. Hit Save
5. An error will appear, that the file can't be saved

### Expected Behavior

The file should be saved.

### Actual Behavior

The file can't be saved.

### Debug Information

Notepad++ v8.2   (32-bit)
Build time : Dec 30 2021 - 03:36:49
Path : C:\Program Files (x86)\Office\Notepad++\notepad++.exe
Command Line : -noPlugin C:\Windows\Sysnative\drivers\etc\hosts
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 (64-bit) 
OS Version : 2009
OS Build : 19044.1526
Current ANSI codepage : 1252
Plugins : none

![image](https://user-images.githubusercontent.com/4572946/153731846-269277a1-233a-469d-9cd0-600ccf88fccf.png)

mere-human:
This file is probably protected and requires administrator rights in order to edit it.
I tried a similar command line:
`-nosession -multiInst -noPlugin  C:\Windows\System32\drivers\etc\hosts`
And I got this message:
![image](https://user-images.githubusercontent.com/9664141/154272010-85b0a065-7298-46e4-8c33-e412b7cc276b.png)
After pressing Yes I was able to modify and save the file.
Did you see a similar message?

xomx:
@mere-human 

From the @dodmi report is clear (e.g. look at the N++ screenshot taken - there is the [Administrator] visible), that the N++ has been already elevated before by launching from the elevated cmd. So there should not be the "Save failed - Do you want ... Administrator mode?" msgbox visible.

@dodmi 

I have tried exactly your steps but with the current 8.3.1 x86 portable. And I was able to save that file. So there should be something else in your way.

[...]

PR #14029, opened 2023-08-19
Fix: Saving files removes alternate data streams

### LINKED ISSUE #1498: Saving with Notepad++ removes alternate data streams
[I posted this on Notepad++ Community](https://notepad-plus-plus.org/community/topic/11280/saving-with-notepad-removes-alternate-data-streams).  Claudia has confirmed it, and has advised me to post here.  I'm using Notepad++ 6.8.8 on Windows 7 Pro SP1. 

---

Suppose that a file Test.txt has two alternate data stream files associated with it. When I edit the file with Notepad++, and then save it (Ctrl + S), the ADS files have been removed.

This doesn’t happen with Notepad, and with UEStudio, there is a setting to prevent it.

QUESTION: How do I save a file in Notepad++ without removing the alternate data streams?

Klotzi111:
This PR fixes #1498.

How to test this is described in that issue here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1498#issuecomment-483194646

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -19,12 +19,39 @@
 #include "FileInterface.h"
 #include "Parameters.h"
 
+// This function must be called "directly" after the call to CreateFile(A/W).
+// Because this function will get the last error code.
+// Returns whether the file was truncated. false can be returned if there is no need to truncate or if truncate failed.
+// But if truncate failed then the file handle will be closed (and set to invalid).
+bool Win32_IO_File::truncateFileAfterOpen()
+{
+	if (_hFile != INVALID_HANDLE_VALUE && ::GetLastError() == ERROR_ALREADY_EXISTS)
+	{
+		// the file pointer will be at the beginning after open
+		// so this call will truncate the file
+		BOOL result = SetEndOfFile(_hFile);
+		// last error might be changed here if the above function fails
+		// but that is more meaningful than the ERROR_ALREADY_EXISTS anyways since that value is not actually an error
+		if (!result)
+		{
+			// truncate failed
+			// we treat that as if opening the file failed
+			::CloseHandle(_hFile);
+			_hFile = INVALID_HANDLE_VALUE;
+			return false;
+		}
+		return true;
+	}
+	return false;
+}
+
 Win32_IO_File::Win32_IO_File(const char *fname)
 {
 	if (fname)
 	{
 		_path = fname;
 		_hFile = ::CreateFileA(fname, _accessParam, _shareParam, NULL, _dispParam, _attribParam, NULL);
+		truncateFileAfterOpen();
 	}
 }
 
@@ -37,6 +64,7 @@ Win32_IO_File::Win32_IO_File(const wchar_t *fname)
 		std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
 		_path = converter.to_bytes(fn);
 		_hFile = ::CreateFileW(fname, _accessParam, _shareParam, NULL, _dispParam, _attribParam, NULL);
+		truncateFileAfterOpen();
 
 		NppParameters& nppParam = NppParameters::getInstance();
 		if (nppParam.isEndSessionStarted() && nppParam.doNppLogNulContentCorruptionIssue())
--- a/PowerEditor/src/MISC/Common/FileInterface.h
+++ b/PowerEditor/src/MISC/Common/FileInterface.h
@@ -58,6 +58,8 @@ private:
 
 	const DWORD _accessParam  { GENERIC_READ | GENERIC_WRITE };
 	const DWORD _shareParam   { FILE_SHARE_READ | FILE_SHARE_WRITE };
-	const DWORD _dispParam    { CREATE_ALWAYS };
+	const DWORD _dispParam    { OPEN_ALWAYS };
 	const DWORD _attribParam  { FILE_ATTRIBUTE_NORMAL };
+	
+	bool truncateFileAfterOpen();
 };


donho:
@pnedev 
Could you review this PR please?
The PR is about modifying FileInterface header and source code.

PowerEditor/src/MISC/Common/FileInterface.cpp
```diff
@@ -19,12 +19,39 @@
...
+// Because this function will get the last error code.
+// Returns whether the file was truncated. false can be returned if there is no need to truncate or if truncate failed.
+// But if truncate failed then the file handle will be closed (and set to invalid).
+bool Win32_IO_File::truncateFileAfterOpen()
+{
+	if (_hFile != INVALID_HANDLE_VALUE && ::GetLastError() == ERROR_ALREADY_EXISTS)
+	{
+		// the file pointer will be at the beginning after open
+		// so this call will truncate the file
+		BOOL result = SetEndOfFile(_hFile);
```
pnedev, line 32: @Klotzi111 ,

Thank you for fixing this issue.
Could you please change the PR so it doesn't use a new WinAPI (SetEndOfFile()) but instead try using CreateFile() initially with TRUNCATE_EXISTING flag and if that call fails with ERROR_FILE_NOT_FOUND error call again CreateFile() with OPEN_ALWAYS ? I kind-of became paranoid with WinAPIs in time as they seem to work in most cases as described but have strange behaviors that are not pointed well in the documentation.
I remember that there were some problems with files truncation after open back then when Notepad Win32_IO_File was first introduced related to network mapped files.
This is just a suggestion of course. The fix looks perfectly fine and if @donho is willing to merge it this way that's OK. But at least first do some verification with network mapped files, please.

Thanks once again!

Klotzi111: I will try to do some of those tests. But can you please explain to me what exactly you mean by network mapped files. Do you just mean a volume mounted via SMB and accessible with an UNC path?

Also as the [documentation at msdn](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setendoffile) says the `SetEndOfFile` api is supported on all filesystems:

<html><body>


Technology | Supported
-- | --
Server Message Block (SMB) 3.0 protocol | Yes
SMB 3.0 Transparent Failover (TFO) | Yes
SMB 3.0 with Scale-out File Shares (SO) | Yes
Cluster Shared Volume File System (CsvFS) | Yes
Resilient File System (ReFS) | Yes


</body>
</html>

donho: @pnedev @Klotzi111 
I'm wondering if `SetEndOfFile` call (new or old API) is necessary.
I tested by modifying only `CREATE_ALWAYS` to `OPEN_ALWAYS`, and the AFS is kept after saving.

Klotzi111: It is necessary. Try this:
- Open a file with some text: `0123456789ABCDEF`
- Then open the file with you modified np++
- Remove some characters (example remove the letters at the end)
- Save the file
- Reload the file (or view the file with an other editor)
- -> The letters at the end are still there

Klotzi111: > I will try to do some of those tests

I tried this now with an network share from an windows server. Everything works as expected.
Because the server also uses an NTFS filesystem the files on it can also have ADS and they are preserved after save.

I also tried it with files on flash drives. One with NTFS filesystem and one with FAT32. And both also worked as expected.

donho: @pnedev 
Thank you for the review and suggestions.

donho: @Klotzi111 
>  The letters at the end are still there.

Indeed.
I can see the different behaviour between CREATE_ALWAYS and OPEN_ALWAYS, however, I don't understand why and how calling SetEndOfFile on only ***file opening*** makes file saving correctly, but on ***file writing*** it is not needed. 
Could you give me an hint please?

pnedev: > 

@Klotzi111 ,

I thing a Network share will suffice (computers in home network for instance).
Thanks!

pnedev: > @Klotzi111
> 
> > The letters at the end are still there.
> 
> Indeed. I can see the different behaviour between CREATE_ALWAYS and OPEN_ALWAYS, however, I don't understand why and how calling SetEndOfFile on only _**file opening**_ makes file saving correctly, but on _**file writing**_ it is not needed. Could you give me an hint please?

@donho ,

I think it is because when we call SetEndOfFile() when opening the file we explicitly truncate the file (dropping the existing contents). When writing the file it is a completely new content (from OS point of view even if you have changed in the editor a few characters) - you write again the total data you have in the editor buffers.
If you don't call SetEndOfFile() the file content is not erased - the EOF mark is still at the end of the existing file contents and writing new data somewhere only overwrites that portion but without moving the EOF mark. The EOF mark on file write in this case is moved only forward (if the new written data is beyond the current EOF) - it is as if the file can only grow larger if you append data but it is not shrunk if you change some data in the existing content. If we want it to be shrunk we must explicitly do so. AFAIU it is the consequence of using OPEN_ALWAYS on existing files.
My initial idea was exactly that - to write the new data and truncate the file to the new write position (depending on the editor buffer size but without rewriting the whole contents) and it worked perfectly on local drives but this approach appeared to have problems with network shared files AFAIR.

BR

xomx: @Klotzi111 
> can you please explain to me what exactly you mean by network mapped files

This is what I remember:
- there were problems reported by users using some kind of a network storage (I think especially in a heterogeneous systems Windows OS / *nix storage)
- if I remember correctly, we were not able to reproduce these errors, so solving of this was difficult and suboptimal (I also think that some of the affected users reported failures of SetEndOfFile WINAPI, when we tried to solve this(?))
- at the end it has been resolved by replacing the more appropriate `OPEN_ALWAYS` with `CREATE_ALWAYS`, look e.g. [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10751#issuecomment-984665070)
- this was the reason for the network logging [added](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10847)

pnedev: Thank you @xomx for reminding us the details, highly appreciated!

donho: @pnedev 
In this case, the API should be called `EnableSettingEndOfFile()` to make the intention of behaviour more clear.
Thank you for the very detail and understandable explanation!

Klotzi111: I did not read all of the linked issues by @xomx yet.
But it sounds difficult. What if we check where the file is located and use different approaches depending on the location (local, network, whatever).
This seems to be the only way how we can fix this issue without re-opening the old ones.

What are your thoughts?

donho: @xomx 
Thank you for reminding us this issue- very appreciated. It seems network file saving issue has been fixed by using `CREATE_ALWAYS` instead of `OPEN_ALWAYS`. I wonder if this fix of ADS will make the network file saving issue happens again.

What do you guys think about it?

xomx: @Klotzi111 
> What if we check where the file is located and use different approaches depending on the location (local, network, whatever).

Yes, that was one option we considered before (cannot find a link) but we did not find a simple & reliable way how to distinguish between the local and network saving (some network drives/folders could be mapped transparently for the users...).

@donho 
We could try the new `TRUNCATE_EXISTING` way suggested by @pnedev first (I think we did not try such option before(?)). Definitely don't try to use the SetEndOfFile somehow, I am sure we did that before with no good results for the users affected. But for this solution we would need help from the users who reported such networking problems last time - you or @Klotzi111 can try to reach out again to the users, who were active in the previous issues, with a testing N++ version...



One more option - I've always felt that the forced solution with the `CREATE_ALWAYS` disposition is not right (for the majority of the N++ users), so maybe the right way is to accept this fix by @Klotzi111 and create a new N++ setting somewhere in the Preferences, which enables the use of the `CREATE_ALWAYS` (the current N++) way somehow as a bypass for such networking problems.

Klotzi111: > distinguish between the local and network saving (some network drives/folders could be mapped transparently for the users...)

I am not sure what exactly you mean by `some network drives/folders could be mapped transparently for the users`.
But there are Win32 API functions with which you can determine the location of remote mapped paths: [WNetGetUniversalNameW ](https://learn.microsoft.com/en-us/windows/win32/api/winnetwk/nf-winnetwk-wnetgetuniversalnamew) and friends

donho: @xomx @pnedev 
Using only TRUNCATE_EXISTING instead of CREATE_ALWAYS does save file correctly (locally), but doesn't fix the losing ADS bug. :(

pnedev: @donho ,
That's not good. Thanks for trying anyway.
Well we are left then with the current @Klotzi111's approach but as @xomx suggested it would be very helpful if we could ping the users that reported the network related save problems before and ask them to retest this PR.
And the next thing would be either to use some APIs to determine the location of the file (to check whether it is network resource or not) as @Klotzi111 pointed or to re-create the saved file's alternative streams (I suspect it might be doable, CreateFile can also open the alternative stream/s but it will be rather difficult).
I think the first approach is better.

Klotzi111: We could also do both of these approaches combined. Maybe configurable via a setting when to (prioritise) use which approach.
I would suggest using `OPEN_ALWAYS` by default and the other options uses `CREATE_ALWAYS`.
This way users with such special cases could switch the behaviour and bypass their problem. For most users `OPEN_ALWAYS` should work perfectly fine.

donho: @pnedev @xomx 
> Using only TRUNCATE_EXISTING instead of CREATE_ALWAYS does save file correctly (locally), but doesn't fix the losing ADS bug. :(

I just relaunch and tested with several file by using TRUNCATE_EXISTING, it does work fine for saving (file is shrink, and ADS is preserved).

I guess the reason of the bad result of last test is that an old version of Notepad++ instance was opened, so after compiling and running Notepad++ via VS, it was the old Notepad++ instance which took over, with the old behaviour, as a result the ADS was removed after saving.

Sorry for the false alert.

donho: @Klotzi111 
Thank you for your contribution and suggestion.
The saving feature is crucial so I prefer to be careful.
I will open another PR which contains only `TRUNCATE_EXISTING`, and invite people who have had network problem to test both version (yours and mine). We will decide what we should do according the result of test.
Thank you for your understanding.

Klotzi111: (I commented your PR)

I am currently working on a version that also fixes #5664 , #6133 and some other problems mentioned in those issues.
I will create a new PR once I think that version is ready to be reviewed.

donho: @Klotzi111 
> (I commented your PR)

Thank you! I have replied you in my PR.

> I am currently working on a version that also fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664 , https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133 and some other problems mentioned in those issues.

All these problems link to the same routine (saving feature), and they could be fixed by this PR or #14039 when one of them are validated by testers. Don't you think it'll be better to wait the result of testing?

pnedev: @donho ,
The current PRs will not fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664 - they are targeted at the alternative data streams problem. https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664 is caused by file truncation on open and it will have to be addressed differently.

Klotzi111: > I suspect @Klotzi111 is working on another file saving approach that will address all the current existing issues altogether

Yes that is true.

Just to make things clear:

> Don't you think it'll be better to wait the result of testing?

Of course we should do testing and wait for some time (or issues) before making a public release. Because that saving routine is the heart of N++.

Klotzi111: @donho Does that mean you will deny any PRs regarding the saving routine?

donho: @Klotzi111 
> Does that mean you will deny any PRs regarding the saving routine?

If it was true, I wouldn't consider this PR.

For issue #6133, it's another story. 
Regardless all the heavy changes we have done in v8.1.5, the bug still persists. I'm not even sure now if it's a bug of Notepad++ or rather a failure of system. 
In anyway, since we cannot reproduce it steadily, there's no way to validate the fix. Should we take risk of regression to integrate a saving routine with heavy refactoring for a uncertain fix? I don't think so.

As a result, the heavy refactoring PR should be avoided, if you want to increase the chance of your PR's acceptation.
Here's the guideline of contribution, in case that you haven't read it yet:

xomx: @donho 
> For issue #6133, it's another story.
> Regardless all the heavy changes we have done in v8.1.5, the bug still persists.

IMO - The number of NUL-reports and alike here and in the Community forum has decreased rapidly after 8.1.5. So I see the change in 8.1.5+ as a 'mission accomplished'. Even if it brought some other problems like #11196 or #11339.

> I'm not even sure now if it's a bug of Notepad++ or rather a failure of system.

True for the remaining few NUL-reports, these either looks like #12305 (i.e. not a N++ problem at all) or the connection with N++ is discutable.

>In anyway, since we cannot reproduce it steadily, there's no way to validate the fix. Should we take risk of regression to integrate a saving routine with heavy refactoring for a uncertain fix? I don't think so.

Agree, it is hard to fix something we cannot reproduce.

@Klotzi111 
> @donho Does that mean you will deny any PRs regarding the saving routine?

I think that Don is not against fixing the ADS issue. He is against a total refactoring (sounded like you're going to re-fix even the old NUL-bug, which is a Pandora's box that shouldn't be opened again...).

It would be ideal if you could fix the ADS bug while also using OPEN_ALWAYS instead of CREATE_ALWAYS somehow and not causing the network problems mentioned.

pnedev: I totally agree with all @xomx's comments.

We could do a fix for the ADS problem using the TRUNCATE_EXISTING - CREATE_NEW - TRUNCATE_EXISTING and the only remaining file save issue I can think of is https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664.

If Don decides to live with it then OK. Otherwise a file save code refactoring approach is probably needed and @Klotzi111's ongoing work is justified.

donho: @xomx @pnedev 
Thank you for your opinion and suggestion.
The issue #5664 is good to fix, but not at all cost. The reason is that the disk room of nowadays is huge, so out of space while saving can happen, but very rarely. Any fix which address to #5664 issue will be considered except the fix with refactoring of code.

But one thing at a time - let's deal with the ADS issue firstly.

Klotzi111: > I now have developed a file saving routine

It contains the following changes/additions:

1. The file is opened with the following disposition values. `allowOverwrite ? OPEN_ALWAYS : CREATE_NEW`
    1. This opens the file without truncating it
        1. This is to ensure that if we fail to write the entire file (or were unable to write anything) we lost as little data as possible. (data is still lost in those cases. Backups are required anyways and are independent of these changes)
    2. We "reuse" the existing file. So we are not losing any metadata or ADS
2. Opens the file with the additional flags `FILE_FLAG_WRITE_THROUGH | FILE_FLAG_NO_BUFFERING`
    1. This causes the Windows file cache to be bypassed for writes to that file and to do atomic writes to the sectors of the storage device
        1. This improves performance because the device does not need to read, update and write a sector
        2. This should prevent cases in which data is lost after the write happened when a fatal system failure occurs (like power loss) because there is not cached data to be lost (at least from the OS side. The hardware cache might still be active and can be lost)
    2. This has some drawbacks, In order to do this the data buffer must be aligned to the physical sector size. And because of that we need to create a buffer that meets these requirements and copy the entrire data over in memory. This can require a lot of memory if the file is big.
        1. A possible solution would be to have the internal buffers of N++ already aligned correctly (Note: physical sector size can change between devices. So a big alignment needs to be used to be safe)
3. Before the actual write operation the required file space is preallocated
    1. This is to ensure we can write the entire content and a OutOfDiskSpace error can not occur (or write nothing if the preallocation fails)
4. The actual file size (more correct: the EOF mark) is set after a successful write
    1. This is required because we write in entire sector blocks and therefore the file can be larger than the wanted content after the write and before setting EOF
    2. If the process is killed before setting EOF there might be some NUL bytes at the end up to the next physical sectore aligned offset

pnedev: @Klotzi111 ,

It is great you made such fix but it appears there are some side cases (where things aren't working as expected or described). For example, please read this thread (thanks @xomx  for providing it here): https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11339 . You can more specifically look at this comment and below: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11339#issuecomment-1067318037 . Also there might be problems with setting the EOF (again due to SetEndOfFile() not working as expected on some network mounts).
Thanks again.

Klotzi111: @pnedev 
Thanks for pointing me at that.
The difference between the problematic code in #11339 and my version is that I also use the `FILE_FLAG_NO_BUFFERING` flag.
I am not 100% certain that this will work correctly. Maybe the disk drivers are also not honoring that. But at least it is not the same. So it could work in those cases.
Testing is required of course.

I repeat the question:
> Are you interested in having the test program to test certain cases?

donho: @Klotzi111 

Thank you for working for it and very detail explanation for your implementation.
However, as I mentioned above, it's very unlikely that Notepad++ integrate a heavy refactoring code of IO, for the sake of stability.

> I repeat the question:
>>    Are you interested in having the test program to test certain cases?

We can try your test program but I don't see the advantage, since all the reported saving issues are not reproducible by us (devs). And it's impossible to ask Notepad++ users to run whichever test program. So the best way to test it is a PR to integrate it into Notepad++, then it can be tested by people who have such issues.

As a result, thank you for your effort, but the test program cannot be used by people who have issues.
However, you can always create a new PR and push your implementation into it if you want to. Although there's very little chance to be integrated into master, I believe @pnedev, @xomx and others are interested in it.

Klotzi111: I understand that it is weird/bad for normal users to run my test program.

But
> I don't see the advantage, since all the reported saving issues are not reproducible by us (devs)

is not completely true. For example the disk full problem (#5664) can/could be reproduced (with a ram disk for example).

> However, you can always create a new PR and push your implementation into it if you want to

I will create a PR soon

donho:
Test request has been launched: #14085

PowerEditor/src/MISC/Common/FileInterface.cpp
Klotzi111: > I will create a PR soon

Here it is:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14090

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14085#issuecomment-1717677564

donho:
Thank you @Klotzi111 for your contribution anyway!

Issue #14030, opened 2023-08-19
Symbolic links and hard links can result in the same physical file being open in multiple tabs


Coises:
### Description of the Issue

This issue was raised in the forums as [Files open twice](https://community.notepad-plus-plus.org/topic/24833/files-open-twice).

Under some conditions, Notepad++ opens the same file in more than one tab. This can happen when a file is referenced by a symbolic link or a hard link. (Shortcuts are always handled properly.)

### Steps to Reproduce the Issue

Create a symbolic link to a file and a hard link to a file.
Try opening the original file and the links in various combinations by various methods.

### Expected Behavior

Symbolic links should always show the target file name when opened.

When a file is opened in a tab, the same physical file should never be opened in another tab, regardless of how it is referenced; instead Notepad++ should switch to the tab in which it is already open.

### Actual Behavior

Symbolic links opened from the File|Open command or by double-clicking in Windows File Explorer show the target name; symbolic links opened from the Explorer context menu or as an argument to Notepad++ on the command line show the name of the symbolic link.

For hard links and for symbolic links which show the name of the link instead of the target name, Notepad++ does not recognize that files it knows by different names refer to the same physical file and opens the file separately under each name.

### Debug Information

Notepad++ v8.5.6   (64-bit)
Build time : Aug 15 2023 - 15:29:28
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 22H2
OS Build : 19045.3208
Current ANSI codepage : 1252
Plugins : 
    ColumnsPlusPlus (0.6.1)
    ComparePlus (1.1)
    DSpellCheck (1.5)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NppMarkdownPanel (0.7.3)

Notepad++ v8.4.8   (32-bit)
Build time : Dec 24 2022 - 19:30:12
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 22H2
OS Build : 19045.3208
Current ANSI codepage : 1252
Plugins : 
    ColumnsPlusPlus (0.6.1)
    DSpellCheck (1.4.24)
[...]

Issue #13995, opened 2023-08-11
feature request: Add a way to open file shortcuts (.lnk files) themselves instead of just opening the file they point to


zbelios:
Currently, there doesn't seem to be a way to open .lnk files (shortcuts). Instead, Notepad++ just opens the file they point to.
Please add a way to open the shortcut itself instead.

OS is Windows 10 if that's relevant.

sredna:
.lnk files cannot be edited in a text editor, you can barely edit them in a hex editor.

Rename .lnk to .txt in cmd.exe if you really need to view one.

I prefer NPP to always follow the shortcut and open the target...

xomx:
@Sorontik 
> I think of Software as a tool like any other.
> If i want to loosen up a screw by tapping it with a hammer - should the hammer turn to jelly because it's creator only thought about nails?
> Think about what a mechanic would say to a screwdriver that forbids them to scrape of a little bit of rust, solely because a screwdriver is meant to screw and not to scrape.

IMO - I think it makes more sense to handle this in a file-manager. E.g. this Total Commander [plugin](https://totalcmd.net/plugring/linkinfo.html) has even a standalone editor app included for these lnk-files manipulation.

N++ would definitely do a bit of everything, albeit imperfectly. But I would prefer if it could do fewer things, but better. But that is a matter of opinion, of course.

> I think users should be allowed to use a tool or program for anything they think it might help them with - even when the creators originally didn't think of that or didn't see it as a relevant use-case.

It is true. But the creators must keep in mind that this is primarily a developer's text editor and think about whether to complicate the program's settings with another item (ignore lnk-targetting) because of such a marginal thing.

SoNickRND:
I tried to rename it to name_lnk.txt but N++ still tries to go to target.

xomx:
@SoNickRND 
Do you have the latest N++?
Patch for using the renaming trick has been included in v8.6.9+.

SoNickRND:
> @SoNickRND Do you have the latest N++? Patch for using the renaming trick has been included in v8.6.9+.

My bad, 8.4.8 still :_D

Issue #11089, opened 2022-01-26
File content, not extension is used to determine if a file is a shortcut


Taemporus:
### Description of the Issue
Somewhat related topics have already been opened and closed (e.g. #9643 or #10139) which are about the behaviour with regards to `.lnk` files and if they or their target should be opened. This is NOT about that; it has been made clear repeatedly that the change is here to stay. It is at least consistent with most other Windows-based tools, so it makes sense.

However, this also happens when a Windows shortcut file has been renamed to no longer have the `.lnk` extension, which can be a problem in a number of ways — and this is in fact not even consistent with other tools either, like the basic built-in Notepad.

### Steps to Reproduce the Issue
1. Take/copy/create a shortcut file `test.lnk`.
2. Rename it to `test.txt` in e.g. the standard command prompt.
3. Open it with Notepad++ (either provide as a command line argument, drag & drop, use the open dialog, or the Open with Notepad++ context menu entry).

(Alternatively, download the attached `desktop.txt` file. WARNING: Notepad++ will attempt to open all files on the desktop, including iterating over the contents of any directories placed there.)

### Expected Behavior
The user sees a `.txt` file, Notepad++ sees one too. One would expect a _text editor_ to open the file _as text_, whatever the text content may be. If a file does not have the `.lnk` extension, Notepad++ should just open it as a regular file, same as most other Windows software do.

### Actual Behavior
Notepad++ seems to read the file, identify it as a WIndows shortcut, then proceed to open the target (or try to do so).

### Why this is a problem
The established behaviour of Windows is for the file extension to determine what the default action is. While `.lnk` files are shown as shortcuts, `.txt` files are not, and the OS doesn’t treat them as such either, regardless of their content. Whether or not a user has, say, extensions turned on in the default File Explorer, it will show the file type as a text file. Something shown as a text file is expected to be opened by a text editor regardless of if the content happens to be gibberish.

[...]

PR #14031, opened 2023-08-20
Improve handling of symbolic links and hard links. Addresses issue #14030.

Coises:
Symbolic links don’t always resolve to the target file before Notepad++ has collected the file path and name it uses to label the tab and to identify when a new file to be opened is actually an already open file. Since the process used for shortcuts works perfectly, this patch adds logic at the beginning of `ResolveLinkFile` in `NppIO.cpp` to check for and resolve symlinks as well.

Hard links cannot be resolved in this way as they reference the file itself, not the original name. Still, the same file shouldn’t be opened twice. This patch adds logic to `FileManager::getBufferFromName` so that as it loops through open files, after checking to see if the requested path+name matches an open file, it uses `std::filesystem::equivalent` to check whether two different path+name specifications refer to the same physical file.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -28,6 +28,7 @@
 #include "fileBrowser.h"
 #include <tchar.h>
 #include <unordered_set>
+#include <filesystem>
 #include "Common.h"
 
 using namespace std;
@@ -123,6 +124,11 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 
 bool resolveLinkFile(generic_string& linkFilePath)
 {
+	if (std::error_code symlink_error; std::filesystem::is_symlink(linkFilePath, symlink_error)) {
+		generic_string target = std::filesystem::read_symlink(linkFilePath, symlink_error);
+		if (!symlink_error) linkFilePath = target;
+	}
+
 	bool isResolved = false;
 
 	IShellLink* psl = nullptr;
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -20,6 +20,7 @@
 #include <locale>
 #include <codecvt>
 #include <sys/stat.h>
+#include <filesystem>
 #include "Buffer.h"
 #include "Scintilla.h"
 #include "ILexer.h"
@@ -1711,6 +1712,7 @@ BufferID FileManager::getBufferFromName(const TCHAR* name)
 				return buf->getID();
 			}
 		}
+		if (std::error_code ec; std::filesystem::equivalent(name, buf->getFullPathName(), ec)) return buf->getID();
 	}
 	return BUFFER_INVALID;
 }


rdipardo:
> the process used for shortcuts works perfectly

The users who reported #13995 and #11089 may beg to differ about that.

mpheath:
```
---------------------------
Cannot open file
---------------------------
"D:\User\Downloads\Editors\Notepad++\A\main.cmd" cannot be opened:
Folder "D:\User\Downloads\Editors\Notepad++\A" doesn't exist.
---------------------------
OK   
---------------------------
```

Drag'n'drop of symbolic link `C:\B\main.cmd` with relative target `..\A\main.cmd` which is located at `C:\A\main.cmd`. It has built the invalid path from the Notepad++ directory instead of from the directory of the symbolic link. Unpatched v8.5.6.0 does not show this error message.

I disagree with the premise of this fix, as symbolic links should not be treated like shortcut links to change to another path. Made a new symbolic link with an absolute path and tested. Opening `C:\B\main.cmd` opens at the path `C:\A\main.cmd` instead. I consider this as a bug.

The contents:

```cmd
cd /d "%~dp0"
> test.log echo hello world
```

will create `test.log` in the files directory. If I go to the Run menu item and run the command `cmd /c "$(FULL_CURRENT_PATH)"`, will I get `C:\B\test.log` created? The answer is No as the intended `C:\B\main.cmd` has become `C:\A\main.cmd` so `C:\A\test.log` is created. The tab shows `main.cmd` though I would usually expect it to represent the path of `C:\B\main.cmd`. Similar happens with `File -> Open in default viewer` and perhaps other features where the use of the symbolic link path would be lost if this is committed. I could imagine that symlinked `index.html` files could make it unworkable due to the change of the symbolic link path for web developers to display their results in a browser with the `View -> View current file in` the choice of browser.

I expect applications to respect a symbolic link like as if it is the actual file that exists there where it is opened. Breaking this behavior makes symbolic links undesirable and that IMO is not acceptable as a fix.

Coises:
@mpheath Thank you for catching this! My proposed change does have a serious fault in handling relative symlinks. In my testing, if the symlink is to a file that opens by default in Notepad++ (such as .txt), unpatched Notepad++ opens the file under the target name when double-clicking; when right-clicking and selecting Open in Notepad++, it opens under the symlink name. With my patch, double-clicking behaves the same way, but right-clicking and selecting Open in Notepad++ fails as you described.

I still question the behavior of unpatched Notepad++; it doesn’t make sense to me that double-click and Open in Notepad++ should produce different results. But it did not occur to me that relative symlinks might behave differently, and I’m glad you caught that.

I’m not quite sure how to withdraw this pull request without losing the important comment you just made. Advice (from anyone) is welcome.

I do still think the other part of the change — the one that stops Notepad++ from opening the same physical file in two separate tabs — is appropriate. I think I made a mistake in combining them into a single commit.

Coises:
There is a significant flaw in this pull request. I withdraw it. There is also discussion at issue #14030 questioning the entire approach, so I will submit a new pull request when and if I have found a solution to the flaw and there is some consensus as to whether this should even be done.

PR #14039, opened 2023-08-23
Fix saving files loosing Alternate Data Streams issue

### LINKED ISSUE #1498: Saving with Notepad++ removes alternate data streams

### LINKED ISSUE #8013: NTFS alternate data streams are lost
If a file with NTFS alternate data streams is opened and saved in notepad++, the alternative data streams are lost.  This is unfortunate.  I use alternate data streams to store data sets and helper scripts to batch files and other scripts.  I noticed reports of this from 4 years ago.  Will it ever be fixed?  Thanks!

### LINKED ISSUE #8451: Saving a file with more than one file stream deletes them
# Description of the Issue
If you have a file with an alternate data stream, make some changes to the main one, then save. All the other data streams will disappear.

1. Create an alternate data stream on a file. ( In cmd: 'echo something > myfile.txt:secondstream' )
2. Open up myfile normally by just double-clicking it.
3. Do some changes and hit save.
4. The 'secondstream' you created is now gone.

### Expected Behavior
The expected behaviour is that all of the file's streams won't disappear.

### Actual Behavior
What happens is that all of the file's streams disappear.

### Debug Information

Notepad++ v7.8.5   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1903
OS Build : 18362.900

To show what happens, in the command prompt you can use the command 'dir /r' to see files including other streams.

Setup:
setup

Before saving with Notepad++:
before

After saving with Notepad++:
after

donho:
This PR related to #14029

Fix #1498, fix #8451, fix #8013

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -16,17 +16,9 @@
 
 #include <locale>
 #include <codecvt>
+#include <shlwapi.h>
 #include "FileInterface.h"
 #include "Parameters.h"
-
-Win32_IO_File::Win32_IO_File(const char *fname)
-{
-	if (fname)
-	{
-		_path = fname;
-		_hFile = ::CreateFileA(fname, _accessParam, _shareParam, NULL, _dispParam, _attribParam, NULL);
-	}
-}
 
 
 Win32_IO_File::Win32_IO_File(const wchar_t *fname)
@@ -36,7 +28,9 @@ Win32_IO_File::Win32_IO_File(const wchar_t *fname)
 		std::wstring fn = fname;
 		std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
 		_path = converter.to_bytes(fn);
-		_hFile = ::CreateFileW(fname, _accessParam, _shareParam, NULL, _dispParam, _attribParam, NULL);
+
+		DWORD dispParam = ::PathFileExistsW(fname) ? TRUNCATE_EXISTING : CREATE_ALWAYS;
+		_hFile = ::CreateFileW(fname, _accessParam, _shareParam, NULL, dispParam, _attribParam, NULL);
 
 		NppParameters& nppParam = NppParameters::getInstance();
 		if (nppParam.isEndSessionStarted() && nppParam.doNppLogNulContentCorruptionIssue())
@@ -104,31 +98,6 @@ void Win32_IO_File::close()
 	}
 }
 
-/*
-int_fast64_t Win32_IO_File::getSize()
-{
-	LARGE_INTEGER r;
-	r.QuadPart = -1;
-
-	if (isOpened())
-		::GetFileSizeEx(_hFile, &r);
-
-	return static_cast<int_fast64_t>(r.QuadPart);
-}
-
-unsigned long Win32_IO_File::read(void *rbuf, unsigned long buf_size)
-{
-	if (!isOpened() || (rbuf == nullptr) || (buf_size == 0))
-		return 0;
-
-	DWORD bytes_read = 0;
-
-	if (::ReadFile(_hFile, rbuf, buf_size, &bytes_read, NULL) == FALSE)
-		return 0;
-
-	return bytes_read;
-}
-*/
 
 bool Win32_IO_File::write(const void *wbuf, size_t buf_size)
 {
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -32,6 +32,14 @@ Win32_IO_File::Win32_IO_File(const wchar_t *fname)
 		DWORD dispParam = ::PathFileExistsW(fname) ? TRUNCATE_EXISTING : CREATE_ALWAYS;
 		_hFile = ::CreateFileW(fname, _accessParam, _shareParam, NULL, dispParam, _attribParam, NULL);
 
+		// Race condition management:
+		//  If file didn't exist while calling PathFileExistsW, but before calling CreateFileW, file is created:  use CREATE_ALWAYS is OK
+		//  If file did exist while calling PathFileExistsW, but before calling CreateFileW, file is deleted:  use TRUNCATE_EXISTING will cause the error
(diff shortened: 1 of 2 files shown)


donho:
@pnedev @xomx 
Your reviews are welcome.

PowerEditor/src/MISC/Common/FileInterface.cpp
```diff
@@ -36,7 +28,9 @@ Win32_IO_File::Win32_IO_File(const wchar_t *fname)
 		std::wstring fn = fname;
 		std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
 		_path = converter.to_bytes(fn);
-		_hFile = ::CreateFileW(fname, _accessParam, _shareParam, NULL, _dispParam, _attribParam, NULL);
+
+		DWORD dispParam = ::PathFileExistsW(fname) ? TRUNCATE_EXISTING : CREATE_ALWAYS;
+		_hFile = ::CreateFileW(fname, _accessParam, _shareParam, NULL, dispParam, _attribParam, NULL);
```
Klotzi111, line 33: I do not like that this is a race condition now. What if the file does not exist when `PathFileExistsW` is called but exists when `CreateFileW` is called? You then are opening the file with the `CREATE_ALWAYS` option and possibly kill the ADS.
You could change your code to first try to open the file with `TRUNCATE_EXISTING` and if that fails with last error = `ERROR_FILE_NOT_FOUND ` you then go on and try open with `CREATE_NEW`.
If now that fails with last error = `ERROR_FILE_EXISTS` you could try  `TRUNCATE_EXISTING` again.
Do those two methods in a loop with like 5 tries and if all fail then report save failure to user.
This guarantees we never loose ADS and we always have a truncated file after open.

PowerEditor/src/MISC/Common/FileInterface.cpp
```diff
@@ -104,31 +98,6 @@ void Win32_IO_File::close()
...
-		return 0;
-
-	DWORD bytes_read = 0;
-
-	if (::ReadFile(_hFile, rbuf, buf_size, &bytes_read, NULL) == FALSE)
-		return 0;
-
-	return bytes_read;
-}
-*/
```
Klotzi111, line 131: I like that this is now removed. Commented out code should not be in source control

Klotzi111 (changes requested):

PowerEditor/src/MISC/Common/FileInterface.cpp
pnedev: Good point. But I think one set of TRUNCATE_EXISTING - CREATE_NEW - TRUNCATE_EXISTING would suffice.
Thanks.

donho: @pnedev 
I see. Thank you for the explanation.
With the same logic, such race condition could happen anytime: a program can delete file just after CreateFileW is called. 
The race condition I do care about is in the same process, other program which can alter the same files opened in Notepad++ is out of our control. 
The problem what @Klotzi111 has pointed out could happen, but it happens very rarely. If we want to prevent it, we should enhance the code in Notepad++ everywhere and it would be not even efficient for all we had done.

pnedev: @donho ,

> With the same logic, such race condition could happen anytime. For exemple: a program can delete file just after CreateFileW is called.

That is not completely the case.
After CreateFileW() we already have the valid file handle and we also do not allow file deletion access (the sharing options used with CreateFileW do not include FILE_SHARE_DELETE).

> The race condition I do care about is in the same process, other program which can alter the same files opened in Notepad++ is out of our control.

This is also not quite true. Considering such scenarios allows us to take appropriate measures and we are responsible for the proper handling even if other processes/programs might be involved.
For example the OS provides the LockFile interface which idea is to properly synchronize OS resources shared by different programs.

> If we want to prevent it, we should enhance the code in Notepad++ everywhere and it would not be even efficient for all we had done.

Well, it is better to make the new code portions behave properly (to best of our knowledge and considerations) even-though there might be portions in the existing code that are prone to similar problems.
If we have a hole in our boat's bottom let's not puncture another one.

There is no functional difference between your current changes and using CreateFileW with TRUNCATE_EXISTING directly.

Thanks.

donho: @pnedev 

> After CreateFileW() we already have the valid file handle and we also do not allow file deletion access (the sharing options used with CreateFileW do not include FILE_SHARE_DELETE).

Makes sense. I wasn't aware of it.

> If we have a hole in our boat's bottom let's not puncture another one.

Love this quote and totally agree with it.

>  Good point. But I think one set of TRUNCATE_EXISTING - CREATE_NEW - TRUNCATE_EXISTING would suffice.
Thanks.

I don't really get your idea. Could you create a new PR from the master please? I then will close this one.

Klotzi111: > @Klotzi111 ,
Please confirm you are currently working on a fix, thank you.

Confirmed. But please give me some days. (Much work at work 😄 )

donho: @pnedev 
According to our discussion here:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14029#discussion_r1305584905

Please reconsider my request **Could you create a new PR from the master please? I then will close this one.**, or provide me the code that you feel suitable for this PR, in the comment.

Thank you.

pnedev: @donho , All
I have made a quick PR as discussed - haven't had time to verify it, sorry:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14051

Please feel free to modify it as you like.

donho:
Test request has been launched: #14085

PR #14053, opened 2023-08-25
Add option to suppress warning when opening big files

DerivativeOfLog7:
Add an option to suppress the >=2GB file warning. Won't show on a x86 build. This is mostly useful when leaving npp searching inside files unattended as the warning could pop up and halt all progress until acted upon
image

(the original PR had 4 👍 reacts)


alankilborn:
Just a thought: Maybe instead of on the `MISC.` subpage, this could go on the `Performance` subpage.  It isn't an exact fit there, but if I were looking for it, my first thought might be that subpage...

DerivativeOfLog7:
> Just a thought: Maybe instead of on the `MISC.` subpage, this could go on the `Performance` subpage. It isn't an exact fit there, but if I were looking for it, my first thought might be that subpage...

Made it on a new branch: https://github.com/DerivativeOfLog7/notepad-plus-plus/tree/alt_placement

image

Positioning can of course be adjusted

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1483,21 +1483,23 @@ bool FileManager::loadFileData(Document doc, int64_t fileSize, const TCHAR * fil
 		}
 		else // x64
 		{
+			NppGUI& nppGui = NppParameters::getInstance().getNppGUI();
+			if (!nppGui._suppress2GBWarning) {
```
donho, line 1487: Please respect coding convention in Notepad++ :
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#general

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -23,18 +23,18 @@
 #define IDC_STATIC -1
 #endif
 
-IDD_PREFERENCE_BOX DIALOGEX 0, 0, 580, 235
+IDD_PREFERENCE_BOX DIALOGEX 0, 0, 580, 250
```
donho, line 26: RESTORE ALL THE VALUE TO 235.

donho:
@alankilborn 
> Just a thought: Maybe instead of on the MISC. subpage, this could go on the Performance subpage. 

For one of rare times, I agree with you totally! :)

This option, if it's not hidden, it should be in Performance subpage.

@DerivativeOfLog7 
And it should be like:

image

donho (changes requested):
1. Please consider suggestions below.
2. Please consider https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14053#issuecomment-1694495210

PowerEditor/src/WinControls/Preference/preference.rc
DerivativeOfLog7: Already done on the fork

alankilborn:
>For one of rare times, I agree with you totally! :)

Well...really...why can't we make this happen more often??  :-)

DerivativeOfLog7:
Moved the checkbox and fixed the code style

image

DerivativeOfLog7:
I assume other variables added for the setting should be rearranged to be near the performance settings instead of the misc?

DerivativeOfLog7:
Would it be appropriate to add a note stating that the setting affects search in files too?

alankilborn:
> Would it be appropriate to add a note stating that the setting affects search in files too?

A note to that effect can be added to the user manual description for the checkbox.  BTW, my personal expectation (which is meaningless) is that developers that get changes to the code accepted also write the user manual text.  (When I do a code change, I also provide the user manual change)

donho:
image

donho:
> 11. Finally, please test your pull requests, at least once.

DerivativeOfLog7:
> > 11. Finally, please test your pull requests, at least once.
> 
> Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests

Sorry... it was working when I first added it; I just forgot to reword the settings cpp after moving it to the performance tab

PR #14065, opened 2023-08-28
[XML] Update romanian.xml part2

TigerxWood:
Part 2 of 3
Update localization and make it pretty printed.

Because the actual Romanian file is not pretty printed the GitHub comparison will not match correct some stings. I compared it against the english version and is a perfect match. The file is tested in the Notepad++, too.

Also, because the differences can't be viewed with full file changed, I split the file in 3 parts. 
P.S. If it is rejected again, I want a clear reason for why, not some generic link with too many instructions that does not apply to translations.

### DIFF
--- a/PowerEditor/installer/nativeLang/romanian.xml
+++ b/PowerEditor/installer/nativeLang/romanian.xml
@@ -628,153 +628,153 @@
             </StyleConfig>
 
             <ShortcutMapper title="Listă combinații taste">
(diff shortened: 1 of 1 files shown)


donho:
PR of reformatting (changing of ws/TAB, line endings or coding style) of source code won't be accepted. Use issue trackers for your request instead.
Ref:

PR #14067, opened 2023-08-29
Suppress 2GB warning option

### LINKED ISSUE #14055: Add an option to suppress large file warning
### Description of the Issue
Notepad++ will completely hang and await user confirmation when trying to open a file bigger than 2GB.

This includes while it's automatically searching for a string in a group of files  
If an user were to start a search in a folder without realizing there are big files, a single one of those files could waste time leaving Notepad++ doing nothing AND still leaving the rest of the files to be searched (especially if the big file were in the beginning of the file list); especially when leaving the computer unattended (but even while using the PC with Notepad++ in the background, as the warning can go unnoticed as it only plays a Windows message sound that could easily be dismissed)

There also may be users that simply don't feel the need to be warned when opening big files (especially on faster machines that don't take "several minutes" to open them, my RAM-starved laptop takes less than 10s for 2GB)

#14053 proposes to add an option to suppress the warning

1. Open a big file
#### Alternatively
1. Open the Search in files window
2. Start a search in a folder containing a big file
3. When Notepad++ tries to open a big file to search it's content, the warning will pop up and halt the search until acted upon

### Expected Behavior
Provide an option to suppress the warning when opening a big file.
Users that don't deem the warning necessary can disable it, and users who want to search in a folder that may contain big files won't have to worry about it halting 

### Actual Behavior
There's no option to suppress the warning
In case of the Search in files function, here's an example of the issue described above:
image

Completely halted, not searching in the rest of the files; if I didn't know about having ≥2GB files in a folder with a lot of files I could leave the PC unattended for nothing because of the warning

### Debug Information
Notepad++ v8.5.6   (64-bit)
Build time : Aug 15 2023 - 15:29:28
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.2215
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

DerivativeOfLog7:
Fix #14055  
Doesn't show up on an x86 build
image

(Fixed & squashed since the last pull request)


donho:

DerivativeOfLog7:
Sorry, I thought it didn't apply to rejects

PR #14069, opened 2023-08-29
Supperss large file warning

### LINKED ISSUE #14055: Add an option to suppress large file warning

DerivativeOfLog7:
Fix #14055  
Doesn't show up on an x86 build
image

(Fixed & squashed since the last pull request) (and made another branch) (hope I got all good this time)


PowerEditor/src/Parameters.h
```diff
@@ -784,6 +784,7 @@ struct NppGUI final
 	bool _isCmdlineNosessionActivated = false; // used for if -nosession is indicated on the launch time
 	bool _detectEncoding = true;
 	bool _saveAllConfirm = true;
+	bool _suppress2GBWarning = false;
```
donho, line 787: Make it into `LargeFileRestriction` structure.

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -539,6 +539,8 @@ BEGIN
     CONTROL         "Allow Smart Highlighting",IDC_CHECK_PERFORMANCE_ALLOWSMARTHILITE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,127,110,200,10
     CONTROL         "Allow Brace Match", IDC_CHECK_PERFORMANCE_ALLOWBRACEMATCH, "Button", BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP, 127, 125, 200, 10
     CONTROL         "Allow URL Clickable Link", IDC_CHECK_PERFORMANCE_ALLOWCLICKABLELINK, "Button", BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP, 127, 140, 200, 10
+    CONTROL         "Suppress warning when opening ≥2GB files", IDC_CHECK_SUPPRESS2GBWARNING, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 78, 178, 217, 10
```
donho, line 542: Use width 280 instead of 217.

PowerEditor/src/WinControls/Preference/preference_rc.h
```diff
@@ -480,4 +480,5 @@
 	#define IDC_CHECK_PERFORMANCE_ALLOWSMARTHILITE      (IDD_PREFERENCE_SUB_PERFORMANCE + 9)
 	#define IDC_CHECK_PERFORMANCE_DEACTIVATEWORDWRAP    (IDD_PREFERENCE_SUB_PERFORMANCE + 10)
 	#define IDC_CHECK_PERFORMANCE_ALLOWCLICKABLELINK    (IDD_PREFERENCE_SUB_PERFORMANCE + 11)
+	#define	IDC_CHECK_SUPPRESS2GBWARNING				(IDD_PREFERENCE_SUB_PERFORMANCE + 12)
```
donho, line 483: Rename `IDC_CHECK_SUPPRESS2GBWARNING` to `IDC_CHECK_PERFORMANCE_SUPPRESS2GBWARNING`

donho (changes requested):

PowerEditor/src/WinControls/Preference/preference_rc.h
Mateos81: Not LargeFileRestriction here? (curious)

DerivativeOfLog7:
@donho should I move the XML node for this in the config file inside the largeFileRestriction node?
image


(also, I'll mention this since I'm here - if I try to debug with VS2022, it complains about a missing [[fallthrough]] [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f5c60d26cde6b577d08787023b96c7836be8e1e6/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L3371) - I haven't put that in my commits since that's not the scope of this PR

donho:
> should I move the XML node for this in the config file inside the largeFileRestriction node?

Yes

> (also, I'll mention this since I'm [here](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f5c60d26cde6b577d08787023b96c7836be8e1e6/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L3371) - if I try to debug with VS2022, it complains about a missing [[fallthrough]] here - I haven't put that in my commits since that's not the scope of this PR)

Yes, I confirm it's not scope - I will merger #14005 to solve the problem.

donho (changes requested):
change for XML

DerivativeOfLog7:
> change for XML

Done, I checked numerous times playing with the setting and other in the performance tab, closing and reopening npp, and all seems to works fine

donho (approved):

donho (changes requested):
Thank you for the previous modification.
One last request: 
Please add the localization entry in [english.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/english.xml) & [english_customizable.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/english_customizable.xml).

DerivativeOfLog7:
> Thank you for your effort. One more (last) request: Please add the localization entry in [english.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/english.xml) & [english_customizable.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/english_customizable.xml).

Done!

donho (approved):

Issue #9722, opened 2021-04-01
Suggestion: Make Notepad++ a restartable Windows 10 application


jcrmatos:
Consider making Notepad++ a restartable Windows 10 application.

The opposite of this article
https://www.howtogeek.com/396884/how-to-stop-windows-10-from-reopening-your-previous-applications-after-restarting-your-pc/

rimrul:
> > https://docs.microsoft.com/en-us/windows/win32/recovery/registering-for-application-restart

That seems to be only for restarting an application after a crash. (Like Windows explorer does)

HarelM:
>Note for application installer authors: Registering for application restart will also cause Windows to automatically restart the application after the computer is restarted in cases where the computer is restarted due to a software update. For this to work, the installer of the application must call the [ExitWindowsEx](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-exitwindowsex) function with the EWX_RESTARTAPPS flag set or the [InitiateShutdown](https://docs.microsoft.com/en-us/windows/desktop/api/winreg/nf-winreg-initiateshutdowna) function with the SHUTDOWN_RESTARTAPPS flag set.

HarelM:
Great!! Thanks!!

Issue #11934, opened 2022-07-20
Open Notepad++ after windows restart


HarelM:
First and foremost, this is a great piece of software! I love it! Thanks you for all the hard work on this!

### Description of the Issue
If Notepad++ was open before a windows restart (for windows updates for example) it will not be reopened after the restart had completed.

### Steps to Reproduce the Issue
1. Open notepad++
2. Restart windows
3. --> Notepad++ is not lunched after the system finishes to boot up.

### Expected Behavior
This is more of a feature request, but it would be nice if notepad++ could be opened after restart

### Actual Behavior
I need to manually open it.

### Debug Information
Not a bug...

### More info
I think the following article might help, but I haven't been writing in C++ for a long time so I'm not sure:
https://docs.microsoft.com/en-us/windows/win32/recovery/registering-for-application-restart

Also worth noting:
I use windows and other applications such as outlook, edge, chrome, teams etc... have been adopting this behavior lately, and I think it's a great user experience.
I'm also using a mac (which sadly can't run notepad++ >>sob<<) and has this experience for longer time now.

CennoxX:
Duplicate of #9722.

HarelM:
My bad, thanks for the info!

Issue #11721, opened 2022-05-25
Notepad++ is not run when Windows 10 restores its previous session (in moment of user login)


piomiq:
### Description of the Issue
I'm not sure is Notepad++ is not run in moment when Windows 10 restore its previous session, because I use portable version or this is not some how supported in Notepad++

### Steps to Reproduce the Issue
1. Turn on restore session in Windows 10
2. Run Notepad++
3. Log off or restart Windows
4. Login again

### Expected Behavior
Notepad++ should be run when Windows 10 restores its session (in moment of user login)

### Actual Behavior
Notepad++ isn't run when Windows 10 restores its session (in moment of user login)

xomx:
@piomiq 
I am just finished a fix, which is needed before anyone could try a PR for this. If everything will be ok (let's say after that fix will be incorporated and released in a future N++ release, so N++ users will be able to test it and confirm its correctness), I will try to add this functionality (like some browsers, e.g. the Google Chrome, have) in a subsequent PR.

PR #14074, opened 2023-08-30
implement Win10+ app-restart feature in Notepad++

xomx:
Fix:  #9722 , #11721 , #11934 and Community forum requests like [this](https://community.notepad-plus-plus.org/topic/20972/are-there-any-plans-to-make-notepad-a-restartable-windows-10-application).

Implemented as per previous discussion: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12541#issuecomment-1332662024

It will allow Notepad++ to be a "restartable app", like some other SW today (eg Google Chrome, Mozilla Firefox or all the MS Windows UWP apps, Office apps etc).

This is to create a seamless experience wherein, if you have to reboot your PC, you can now pick back up from where you left off and resume being productive.

The OS app-restart feature needs at least Windows 10 (20H1, builds 18963+, but it was quietly introduced earlier in the Fall Creators Update (b1709+)) and the user has to switch on the "Restart apps" in the system Settings, Win10 screenshot:
Win10-Settings-Accounts-SignInOptions-RestartApps

The corresponding Registry-key ON/OFF for testing:
[RestartApps-Registry.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/12477112/RestartApps-Registry.zip)

N++ Preferences (default is OFF):
NppMiscAppRestart

Currently the translation xml is not updated with new strings from this PR - I will add them later if this PR and the names used are accepted.

PR has been successfully tested under various Win10 and Win11. I also run this under Win7, all ok (the feature is not accessible from there).


xomx:
Another way to achieve this is to use Explorer compatibility hack:
app_properties_Compatibility_tab

donho:
@xomx 
I'm kind of confused. According https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12541#issuecomment-1332662024, it's about a feature to restore application after Windows being restarted automatically due to a Windows update. So for me it's not necessary to have an option to activate it.

Am I missing some points about this feature?

xomx:
@donho 

Currently the N++ is not automatically restarted (reopened) by OS after the Windows Update restart at all.

If you activate this new PR option in N++, your actual N++ session will be automatically reopened not only in the case of the mentioned WU-restart, but also after usual user-initiated one (and even after user logoff/logon sequence).

I am currently using this feature e.g. in Firefox, where I usually have a lot of tabs open but for some reason I need a quick Windows restart and I do not want to bother with setting my browser to the exactly same state as it was before the restart.

donho:
@xomx 

> I am currently using this feature e.g. in Firefox, where I usually have a lot of tabs open but for some reason I need a quick Windows restart and I do not want to bother with setting my browser to the exactly same state as it was before the restart.

My question is: does exist such option **Restart Firefox for system app-restart** to be enabled to make it work?

xomx:
@donho 
> My question is: does exist such option **Restart Firefox for system app-restart** to be enabled to make it work?

Yes, in the Firefox settings for power-users, there is the `toolkit.winRegisterApplicationRestart`, which has to be `true` for that feature to work (I think this is currently the default option in Firefox builds and IMO - it's not a good choice):
Firefox-toolkit_winRegisterApplicationRestart

If you think in a way - let's enable this feature in N++ but save a corresponding on/off option in Preferences, I think it is not a good idea. This feature can have unpleasant consequences for the users privacy. If someone explicitly enables this option, that's fine (it is then up to such user to eventually close the N++ before the shutdown/restart/logoff, when one does not want N++ reopening at the next logon). It has a bit to do with the new N++ option "Always open a new document in addition at startup", which is there to prevent similar accidents for some users.

donho:
> If you think in a way - let's enable this feature in N++ but save a corresponding on/off option in Preferences, I think it is not a good idea. This feature can have unpleasant consequences for the users privacy. If someone explicitly enables this option, that's fine (it is then up to such user to eventually close the N++ before the shutdown/restart/logoff, when one does not want N++ reopening at the next logon). It has a bit to do with the new N++ option "Always open a new document in addition at startup", which is there to prevent similar accidents for some users.

I see your point. 
However, users won't be aware of this new option, so Notepad++ will be kept being ignored after OS restarting of Windows Update.
Furthermore, we've got already "Always open a new document in addition at startup", one more option for the same goal is kind of too much. 

Another argument for removing this option is, obviously there aren't so many complains from users of Firefox, Chrome & Edge (the amount of their users is huge) - otherwise, they would have considered making it as option since long time ago.

What do you think?

xomx:
@donho 

I searched for [such reports](https://support.mozilla.org/en-US/questions/1270889) and it looks like you're right - most users probably like this feature ON as the default. But e.g. Tor Browser, which is based on Firefox and focuses on the privacy more, has this setting [reversed](https://gitlab.torproject.org/tpo/applications/tor-browser/-/blob/tor-browser-115.2.0esr-13.0-1/browser/app/profile/001-base-profile.js?ref_type=heads#L560).

Anyway, if you rather like the GUI option in the Preferences removed, I propose these changes:

1) I will completely remove all my changes in the .\PowerEditor\src\WinControls\Preference\ subfolder, so the new GUI option simply disappear.

2) But I will not touch much the other code, so we can easily put the GUI back if we need so in the future.

3) For the power-users I am considering to use the hidden "GUIConfig commandLineInterpreter" [way](https://npp-user-manual.org/docs/preferences/#preferences-for-advanced-users). Using so we can have even more flexible app-restart setting - we can use there an `int` value representing the `dwFlags` param for the [RegisterApplicationRestart](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerapplicationrestart) (new GUIConfig item will have its default `int` value `3`, which is in fact the `RESTART_NO_CRASH | RESTART_NO_HANG` numerically) or `-1` if one wants to completely dissable this feature in N++. I have already successfully tested this approach and the code changes are not big.

donho:
@xomx
IMO, the best way to let the most of users benefit it, but not to penalize small amount of users who don't want this feature, is make it optional, enabled by default. 
It's a hidden option not on GUI, either in the config.xml, but by adding an empty xml (`noRestartAutomatically.xml` or you can find a better name) to disable this feature.

What do you think?

xomx:
@donho 
> IMO, the best way to let the most of users benefit it, but not to penalize small amount of users who don't want this feature, is make it optional, enabled by default.

Yes, that is exactly my intention now. 

> t's a hidden option not on GUI, either in the config.xml, but to disable it by adding an empty xml (`noRestartAutomatically.xml` or you can find a better name) to disable this feature.

Ok, I will show you first my already finished solution with the config.xml file only. If you don't like it, I will also add another file trigger possibility (the mentioned `noRestartAutomatically.xml`)

xomx:
@donho 

The OS app-restart possibility is now automatically enabled in N++ at its startup unless a power-user chooses otherwise.

To disable this feature in N++, one has to manually edit the config.xml file in a similar way how it is described for the changing of the cmd-line interpreter in the [Preferences for Advanced Users](https://npp-user-manual.org/docs/preferences/#preferences-for-advanced-users) (if this way is accepted, I will also issue a request for the npp-user-manual update...)

Disabling the N++ app-restart feature (it has to be `-1` and not `0`, as the zero is also a suitable flag for the [RegisterApplicationRestart](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerapplicationrestart) WINAPI:
GUIConfig-item

When there is not such line in the config.xml, N++ uses its default value for the `registerForOSAppRestart` (which is `3` aka the `RESTART_NO_CRASH | RESTART_NO_HANG`)

In this way, the user can also adjust the registering parameters for his needs. E.g. when one uses `7` (instead of `-1` for disabling), N++ uses `HRESTART_NO_CRASH | RESTART_NO_HANG | RESTART_NO_PATCH` combination (1+2+4) for the registration.
 
All possible OS flags values:
RESTART_NO_CRASH: 0x00000001
RESTART_NO_HANG: 0x00000002
RESTART_NO_PATCH: 0x00000004
RESTART_NO_REBOOT: 0x00000008

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -115,6 +116,8 @@ LRESULT Notepad_plus_Window::runProc(HWND hwnd, UINT message, WPARAM wParam, LPA
 						SWP_FRAMECHANGED);
 				}
 
+				::SendMessage(hwnd, NPPM_INTERNAL_HANDLE_OS_APP_RESTART, (WPARAM)false, 0);
```
donho, line 119: Why do we need `NPPM_INTERNAL_HANDLE_OS_APP_RESTART`?
Why don't we call `SetOSAppRestart(bool bLog)` directely?

PowerEditor/src/Parameters.h
```diff
@@ -849,7 +855,9 @@ struct NppGUI final
 	std::wstring _definedSessionExt;
 	std::wstring _definedWorkspaceExt;
 
+	// items with no Notepad++ GUI to set
 	std::wstring _commandLineInterpreter = CMD_INTERPRETER;
+	int _registerForOSAppRestart = DEFAULT_OS_APP_RESTART_FLAGS; // -1 ... disable (unregister)
```
donho, line 860: Why do we need `_registerForOSAppRestart`? After all, it's a argument which contains flags...
We can pass this argument directly while calling `RegisterApplicationRestart`.

donho (changes requested):
Please remove the setting from config.xml and use `noRestartAutomatically.xml` to disable this feature.
Remove unnecessary arguments/parameters.

donho:
@xomx 
Thank you for the modification!

However,
> To disable this feature in N++, one has to manually edit the config.xml file in a similar way how it is described for the changing of the cmd-line interpreter in the [Preferences for Advanced Users](https://npp-user-manual.org/docs/preferences/#preferences-for-advanced-users) (if this way is accepted, I will also issue a request for the npp-user-manual update...)

I remain the choice of `noRestartAutomatically.xml` - users don't need to modify config.xml at all. To re-enable this feature, remove this file - it's just simple and efficient.

> When there is not such line in the config.xml, N++ uses its default value for the registerForOSAppRestart (which is 3 aka the RESTART_NO_CRASH | RESTART_NO_HANG)
>
> In this way, the user can also adjust the registering parameters for his needs. E.g. when one uses 7 (instead of -1 for disabling), N++ uses HRESTART_NO_CRASH | RESTART_NO_HANG | RESTART_NO_PATCH combination (1+2+4) for the registration.

I'm wondering why user do not need RESTART_NO_PATCH. IMO, we use HRESTART_NO_CRASH | RESTART_NO_HANG | RESTART_NO_PATCH for the classic case and it's enough.

PowerEditor/src/Parameters.cpp
```diff
@@ -7329,6 +7343,15 @@ void NppParameters::createXmlTreeFromGUIParams()
...
 
+	// <GUIConfig name="registerForOSAppRestart">3</GUIConfig>
+	if (_nppGUI._registerForOSAppRestart != DEFAULT_OS_APP_RESTART_FLAGS)
+	{
+		std::wstring str = std::to_wstring(_nppGUI._registerForOSAppRestart);
+		TiXmlElement* GUIConfigElement = (newGUIRoot->InsertEndChild(TiXmlElement(TEXT("GUIConfig"))))->ToElement();
+		GUIConfigElement->SetAttribute(TEXT("name"), TEXT("registerForOSAppRestart"));
+		GUIConfigElement->InsertEndChild(TiXmlText(str.c_str()));
+	}
+
```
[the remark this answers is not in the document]
xomx: > No need to waste one more place in xml file.

I will switch to the requested noRestartAutomatically.xml way.
But I still would prefer one (optional) extra line in the already existing config.xml file over the another new xml trigger-file in the N++ dir.

PowerEditor/src/Notepad_plus.h
```diff
@@ -52,6 +52,8 @@
 
 #define MAIN_EDIT_ZONE true
 
+#define NPP_APP_RESTARTED_BY_OS_CMDLINE_PARAM (L"-restartedByOS")
```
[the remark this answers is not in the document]
xomx: > Why do we need this flag?

This is for the N++ (and its plugins) to know that an automatic OS-restart happened (it was intended as an adaptation possibility for a possible future needs...).

> I did check to see what _isRestartedByOS is for, but didn't see anything.
Or am I missing somethings?

I was using that in a previous PR version.

PowerEditor/src/NppBigSwitch.cpp
xomx: Again, this is a remnant from the previous PR version, now it is not needed. I will simplify that.

PowerEditor/src/Parameters.h
xomx: Because of user can fine adjust the OS Restart Manager setting (in my current PR variant with config.xml, where another set of these restart flags can be stored...).

xomx:
@donho 
Sorry for the delay, now I resume my work on this.

> I remain the choice of noRestartAutomatically.xml

ok

> I'm wondering why user do not need RESTART_NO_PATCH. IMO, we use HRESTART_NO_CRASH | RESTART_NO_HANG | RESTART_NO_PATCH for the classic case and it's enough.

Good point.
`RESTART_NO_CRASH | RESTART_NO_HANG | RESTART_NO_PATCH` is used e.g. in the Chrome source code, whereas other apps (e.g. Firefox) use `RESTART_NO_CRASH | RESTART_NO_HANG` only.

I will change it to the requested `RESTART_NO_CRASH | RESTART_NO_HANG | RESTART_NO_PATCH`.

xomx:
@donho 
Done.
Now N++ only automatically registers itself on its launch (at the end of successful WM_CREATE) for the OS app-restart.
To possibly suppress the registration, user can put the empty "noRestartAutomatically.xml" to the N++ app-folder or to the "%APPDATA%\Notepad++\".

donho:
@xomx 
Thank you for your work.
It works, but adding "noRestartAutomatically.xml" into "C:\temp\test\" doesn't prevent Notepad++ from starting, after OS being restarted.
Here is my Debug Info:
```
Notepad++ v8.5.7   (64-bit)
Build time : Sep 21 2023 - 14:30:28
Path : C:\XXXXXXXX\notepad-plus-plus\PowerEditor\bin64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : C:\temp\test
OS Version : 22H2
OS Build : 22621.2283
Current ANSI codepage : 1252
Plugins : 
    ChangedLines (0.3.2.1)
    mimeTools (2.9)
    NppConverter (4.4)
    NppPluginDemo (4.3)

```

xomx:
@donho
Maybe because of this (is that intended?):
> Cloud Config : C:\temp\test

donho:
@xomx 
Yes, that's what I said:
It works, but adding "noRestartAutomatically.xml" into **"C:\temp\test"** doesn't prevent Notepad++ from starting, after OS being restarted.

donho:
BTW, remove cloud setting and copy "noRestartAutomatically.xml" into "%APPDATA%\Notepad++" works.

donho:
@xomx 
Sorry for the false alert: I have checked in your code, and found it it should work, so I did try several time. It works with cloud settings finally. I guess I've got the cloud settings file corrupted, due to some modification and test for fixing https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14131
Sorry again for the false alert.

HarelM:
I see this feature in the release notes, do I need to activate it, or is this on by default?
I just restarted my PC and I didn't notice that N++ started again when the computer started.

xomx:
@HarelM 
In N++ v8.5.8+ it is ON by default.
But your Windows OS global setting for that feature has to be ON as well (in Win11 it is ON by default, but in Win10 it has to be activated manually (Settings > Accounts > Sign-in options > Restart Apps).

HarelM:
Thanks for the quick response!
Interesting, I'm using win 11 and N++ 8.5.8, I'll do another restart and see if I can reproduce the issue. I might have closed it before the restart so it might have been my fault.

HarelM:
Seems like this was off in the windows configuration, might be some company policy that turned it off.
Once turned in win 11, N++ is up and running after a restart.
THANKS!!!!! This is very very much appreciated.
Thanks for all the hours put into this amazing tool!

Issue #13479, opened 2023-04-05
Dragging tab takes too long if document list presents


maxox:
if the number of tabs is above 150, dragging tab from the last position would trigger the the document list refresh for couple seconds.
it would take like forever to move the last one to the first one.

https://user-images.githubusercontent.com/7052069/230031302-e6ce9f72-c7cb-479a-9748-3425c854e722.mp4

molsonkiko:
I tried opening like 150 tabs on both an older version of NPP and also the most recent version.
In both cases, moving tabs wasn't too slow.
Admittedly, the tabs were all empty documents (I just mashed Ctrl+n a bunch of times), but that probably shouldn't change how long it takes the document list to update.

Orden4:
This has been annoying me for months now. The thing is that I have a ton of tabs. I do try to periodically clear the bulk of them, but it builds up again before long. However, the ones at the top are always the same, in the same order. Sometimes I close a tab by accident, and then I have to go through this absolutely painful process of moving that tab to the start again, since restore recent closed file doesn't retain its position either. NavigateMe isn't really a solution to this, since it's just a lot more convenient to have them all at the top than to have them spread around and always navigate my way around via searches.

So anyway, to assist with getting this issue fixed: I've confirmed that this issue occurs even with all plugins removed
The reason that I suspect that you did not experience this issue is because you probably tried to move documents around near the top of the list. I don't really understand what exactly is happening behind the scenes, but it sort of boils down to that it will scroll through the document list until it has returned to your currently selected tab. If you try to get the last tab to the front in this way, you'll find that it speeds up over time as you get closer to the start of the document list.

Issue #12632, opened 2022-12-12
Moving tab very slow if Document List Panel is open with many tabs


SlugFiller:
### Description of the Issue
When the Document List Panel is open, and many tabs are open, dragging a tab, or moving it using shift+mouse wheel, causes the Document List to completely reload, by removing all the items, then re-adding them. If there are many tabs, the process takes a couple of seconds (!!), and is repeated on every single movement of the tab, even if still holding the button down in a mouse drag.

### Steps to Reproduce the Issue
1. Open NotePad++
2. View->Document List
3. Hold Ctrl+N for a couple of seconds
4. Hover the mouse over the tab bar
5. Hold shift, and roll the mouse wheel down.

### Expected Behavior
The tab just moves, and I can continue moving it or working with the document

### Actual Behavior
The Document List's scroll bar goes crazy for a while as all the opened tabs first disappear, then reappear in the new order. About 5-6 seconds of work time are lost, and that's for a single movement. If wanting to move, say, the first tab to the end, a few minutes would be required.

### Debug Information

Notepad++ v8.4.6   (32-bit)
Build time : Sep 25 2022 - 19:55:26
Path : C:\Apps\NotepadPP\notepad++.exe
Command Line : "C:\Docs\Programs\mollyim-android\libsignal\service\src\main\java\org\whispersystems\signalservice\internal\websocket\WebSocketConnection.java"
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 22H2
OS Build : 19045.2251
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    DSpellCheck (1.4.24)
    HTMLTag_unicode (1.2.3)
    LocationNavigate (0.4.8.1)
    mimeTools (2.8)
    MultiClipboard (2.1)
    NppConverter (4.4)
    NppExport (0.4)
    NPPJSONViewer (1.40)
    NppSaveAsAdmin (1.0.211)
    NppToolBucket (1.10.6622.41336)
    Remove Duplicate Lines (1.3)
    ScrollPastEOFUni (1)
    SessionMgr (1.4.4)

korpx:
Seen the same issue here for awhile

Notepad++ v8.4.6   (64-bit)
Build time : Sep 25 2022 - 19:51:39
Path : C:\Program Files\Notepad++\notepad++.exe
[...]

PR #14080, opened 2023-09-01
Greatly speed up reloading document list.

Orden4:
Addresses https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13479 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12632
Simply supplies an additional argument to `VerticalFileSwitcherListView::remove` so that it doesn't delete the item, allowing you to do it in bulk later, which greatly reduces the number of redraws required.

This reduces the time it takes to drag a tab from first to last/last to first with 138 files opened from ~1m55s to ~20s. Still a considerable time, but at least it's no longer increasing exponentially with number of files opened (see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13479#issuecomment-1682160614 for a more detailed explanation).

A more ideal solution would likely be to actively swap items instead of reloading everything, but I'm too inexperienced in C++ to really get anywhere on that, so I just did the best I can to address a thing that's been driving me mad for months to at least a bearable level.

### DIFF
--- a/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.h
+++ b/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.h
@@ -88,7 +88,7 @@ protected:
 
 	int find(BufferID bufferID, int iView) const;
 	int add(BufferID bufferID, int iView);
-	void remove(int index);
+	void remove(int index, bool del = true);
 	void removeAll();
 	void selectCurrentItem() const {
 		ListView_SetItemState(_hSelf, _currentIndex, LVIS_SELECTED | LVIS_FOCUSED, LVIS_SELECTED | LVIS_FOCUSED);
--- a/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
+++ b/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
@@ -352,7 +352,7 @@ int VerticalFileSwitcherListView::add(BufferID bufferID, int iView)
 }
 
 
-void VerticalFileSwitcherListView::remove(int index)
+void VerticalFileSwitcherListView::remove(int index, bool del)
 {
 	LVITEM item{};
 	item.mask = LVIF_PARAM;
@@ -360,7 +360,9 @@ void VerticalFileSwitcherListView::remove(int index)
 	ListView_GetItem(_hSelf, &item);
 	TaskLstFnStatus *tlfs = (TaskLstFnStatus *)item.lParam;
 	delete tlfs;
-	ListView_DeleteItem(_hSelf, index);
+	
+	if (del)
+		ListView_DeleteItem(_hSelf, index);
 }
 
 void VerticalFileSwitcherListView::removeAll()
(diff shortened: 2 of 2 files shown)


PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.h
```diff
@@ -88,7 +88,7 @@ class VerticalFileSwitcherListView : public Window
 
 	int find(BufferID bufferID, int iView) const;
 	int add(BufferID bufferID, int iView);
-	void remove(int index);
+	void remove(int index, bool del = true);
```
donho, line 91: Please change **del** to **removeFromListview**.

donho (changes requested):
Please change del to removeFromListview.

donho:
BTW, I notice that you're using master to create your PR. It's OK for this time, but please create a new branch for every new PR.
Ref:  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests

Orden4:
Sorry, am horrible at checking personal e-mail during workdays. Thanks for including the PR anyway!

SlugFiller:
This is only half the solution. Adding the items would still be slow. The correct way to handle bulk updates is with `WM_SETREDRAW`. You set it to `FALSE` at the start of the bulk update, set it to `TRUE` at the end, and then call `RedrawWindow` on the control to update everything at once. This should reduce it from 20s to under a second.

Orden4:
> This is only half the solution. Adding the items would still be slow. The correct way to handle bulk updates is with `WM_SETREDRAW`. You set it to `FALSE` at the start of the bulk update, set it to `TRUE` at the end, and then call `RedrawWindow` on the control to update everything at once. This should reduce it from 20s to under a second.

Ah, so that's is how you're meant to do that. I did try to find such a thing but google just wasn't any help.

Anyway, just for clarity let me first say that 20s is the time for moving it across 138 items 1 by 1 (by holding the file at the far left or right of the tab list), you don't really *want* it to move 138 items across in less than a second because that'd just be uncontrollable. Anyway, by adding `WM_SETREDRAW` around the first two lines of the reload method like this:
```cpp
	::SendMessage(_hParent, WM_SETREDRAW, false, 0);
	removeAll();
	initList();
	::SendMessage(_hParent, WM_SETREDRAW, true, 0);
```
It moves at something like 15 to 20 items per second regardless of list size, which is significantly faster but still just slow enough to follow, so that seems like a good balance to me.
However, the one oddity is that the selection will always be off by one during the dragging process, and I can't really seem to get that to draw properly.

SlugFiller:
I'm fairly sure that should be `_hSelf` not `_hParent`. ListView has special handling for `WM_SETREDRAW` which disables all sort of actions it performs during item add, like sorting, updating user data, etc. That's why it's important to send it to the ListView itself.

There's also a macro for it it's `SetWindowRedraw`. So that could be `SetWindowRedraw(_hSelf, FALSE);`

For the draw delay, are you calling `RedrawWindow` on the list after re-enabling drawing? If you want to be extra fast, you can pass the flag `RDW_UPDATENOW`, although even without it, redraw should happen within a frame. Another possibility is to call `ListView_RedrawItems` to inform it that it needs to update. That should be called after enabling drawing, and before calling `RedrawWindow`. And, again, the redraw should be on `_hSelf`, not `_hParent`

Orden4:
It's probably best to comment on the new PR that I made rather than this old PR, but anyway.

Suppressing redraw on `_hSelf` instead of `_hParent` will cause the list to rapidly flicker while moving items down. I guess it may be useful to suppress both or something just to make it slightly more performant. But at least on my own PC there's no meaningful performance gain anymore anyway, as I'd say having it move faster than 15~20 items per second will just make it difficult to control.

Furthermore, regarding the redrawing stuff, the biggest reason from what I could tell was just that the order isn't actually updated prior to calling `reload`. Later on there is a call to `activateItem` at which it *is* actually updated properly, so I just made it made it do an explicit repaint only at the end of that which gave an excellent end result that works fast and doesn't flicker.

Issue #5806, opened 2019-06-19
Digitally Sign Uninstaller


wrolisonaflac:
### Description of the Issue
Notepad+= Uninstaller is not digitally signed

### Steps to Reproduce the Issue

1. Install Notepad++
2. Uninstall Notepad++ via official uninstaller
3. Notice uninstaller is not digitally signed

### Expected Behavior
I expected the Notepad++ uninstaller to be digitally signed just like the installer and dlls

### Actual Behavior
Notepad+= Uninstaller is not digitally signed so some software may detect as malicious

https://notepad-plus-plus.org/community/topic/17838/notepad-uninstaller-is-not-digitally-signed

MetaChuh:
welcome to the notepad++ issue tracker, @wrolisonaflac 

the nsis installer's `uninstall.exe` is not detected as malicious by any of the top ~70 antivirus/antimalware engines.

please see the current virustotal.com results:

uninstall.exe (7.7)
https://www.virustotal.com/gui/file/a9e7777005905f608973696b4d83b8013ebfa82fc5dcad327390acba6ff3ce52/detection

uninstall.exe (7.7.1)
https://www.virustotal.com/gui/file/5ff7a866c0cce74680b01142bb9613388fef23b5cd68215c02dfe9baccc88487/detection

npp.7.7.1.Installer.exe
https://www.virustotal.com/gui/file/6787c524b0ac30a698237ffb035f932d7132343671b8fe8f0388ed380d19a51c/detection

please also make sure you only download notepad++ from our official resources at https://notepad-plus-plus.org/download/

many thanks and best regards.

wrolisonaflac:
Was being picked up by CrowdStrike. It's been fixed for that.

But still why close? And why not sign the uninstaller just like everything else?

MetaChuh:
@wrolisonaflac 
> But still why close? And why not sign the uninstaller just like everything else?

because `uninstall.exe` is not a shipped file, but a helper file created on the fly, by the nsis installer at the time of installation.
if you might ask: the version information you see, is also created on the fly by nsis, using the info provided at the installer script.

in short: there is no possible way to sign this file on a target system.
(except theoretically implementing a post installation script, and shipping everything needed to sign it, which will obviously never happen 😉)

many thanks and best regards.

wrolisonaflac:
Got ya. FYI: https://nsis.sourceforge.io/Signing_an_Uninstaller

[...]

Issue #4443, opened 2018-04-23
Virus found on file: uninstaller.exe (version 7.5.6, 32-bit)


gars:
### Steps to Reproduce the Issue

1. Download the 32bit installer from https://notepad-plus-plus.org/download/v7.5.6.html.
2. Install it (right click, Run As Administrator, and grant elevation privilege for installer)
3. Run notepad++.
4. Anti-virus catches the "win32-malware-gen" inside the file uninstall.exe

### Expected Behavior

Notepad++ running without viruts

### Actual Behavior

Virus detected

### Debug Information

![image](https://user-images.githubusercontent.com/422404/39101935-c1b74e88-4675-11e8-8e25-843c1f4fe571.png)

SinghRajenM:
This is false alarm. Uninstall.exe is auto generated nsis script (which is popular script to create installers including commercial setups). I feel it can be reported to Avast too.

Issue #4120, opened 2018-01-26
AntiVirus Alert


Gryffen1971:
[ComodoLogEvent(26_Jan_18_04_32_05).zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/1667292/ComodoLogEvent.26_Jan_18_04_32_05.zip)

### Description of the Issue
I received an Alert from Commodo Internet Security about a Virus. 
Alert Time; 2018-01-26 04:32:05
Malware/Virus name;  TrojWare.Win32.TrojanDownloader.Tovkater.ID@452171600
Location; C:\My Program Files\Notepad++\uninstall.exe
Program was not running at the time this occured.
Attached is the Antivirus Event as a ".htm" document with the same info provided above

### Steps to Reproduce the Issue

1. Don't know want too!
2. 
3. 

### Expected Behavior
Unknown, program was not running at time.

### Actual Behavior
Anti-Virus caught, Quarantined and Disinfected file with program NotePad++ not running.

### Debug Information
No Debug info need. This happened while program was not running.

dail8859:
What version of Notepad++ are you using and is it 32 or 64 bit? It is very likely it is a false positive.

www.virustotal.com shows the uninstallers for v7.5.4 are clean. Results:

- [32bit uninstaller](https://www.virustotal.com/#/file/84810d00344ec4a3c9e68498e98e974a262232cf07b4f21b6157efdb5d36a845/detection)
- [64bit uninstaller](https://www.virustotal.com/#/file/c6a12aeae1f7121c91fa4053b3f7fc408753edaa4d5fe6b74b2e3e8ba2c1241c/detection)

SinghRajenM:
> It is very likely it is a false positive.

Looks like false alarm only, if the installer (32bit or 64 bit) is downloaded from official site. File ```C:\My Program Files\Notepad++\uninstall.exe``` is auto generated by installer which seams very less likely virus infected.

Gryffen1971:
This was on my system for some time. It was downloaded from your secure site. I believe a virus tried and failed to infect your file. I wanted to report it to you. So, you that you would be aware of this. I know that you and your team have a secure site and that all of your files are safe. This was an 'FYI'. I was just looking out for ya.

PR #14087, opened 2023-09-03
Sign uninstall.exe

SinghRajenM:
There were numerous threads/issues which reported that npp's `uninstall.exe` is detected by various antivirus.

However, those were false positives. Mainly, these were detected by ML module which works upon static characteristics of binary. Sometimes, these ML modules detect just `Hello World` program too. Unfortunately, there are some limitation to handle these cases. But signing characteristics of binary play a very vital role there (due to signature based scanning/rules). Hence, valid signed binaries can help to mitigate these false positives.

So, it is good if we can sign the uninstaller exe. Prior to `nsis 3.08`, it was a complex process to sign the uninstaller binary. But `nsis version 3.08` makes it very easy with the help of `uninstfinalize`.  Refer [nsis documentation](https://nsis.sourceforge.io/Docs/Chapter5.html#uninstfinalize).

Additionally, it can sign installer too using `finalize`. Just an improvement to automating things.

Related issues #4120, #5806, #4443. May be few more.

@donho  further things to take care - 
1. Replace `npp_sign_cert.pfx` with actual/test npp certificate. I know original cert file can not be in repo. Hence, 
    1. You can generate test cert and put it in repo.
    2. Or you can define a macro like `SINGING_NEEDED` and both the `finalize` and `uninstfinalize` can be guarded using it.
    3. Either of these two will help others as nsis will not complain at compilation.
5. You can grab latest SignTool.exe from Widows SDK kit (e.g. `C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x86\signtool.exe`). This may require to update sign script too as there could be some new parameter.


PowerEditor/installer/sign-binaries.bat
```diff
@@ -0,0 +1,13 @@
...
+
+".\signCert\SignTool.exe" sign /f ".\signCert\npp_sign_cert.pfx" "%1"
+if errorlevel 1 goto SigningFailed
+goto SigningOK
+
+:SigningFailed
+echo Failed to sign file %1
+exit 1
+
+:SigningOK
```
donho, line 13: Please rename the file `sign-binaries.bat` to `sign-installers.bat`,
and use the following content instead:
 
```batch
@ECHO OFF

if [%SIGN%] == [] goto NoSignInstaller
if not %SIGN% == 1 goto NoSignInstaller

ECHO Start signing file: %1
%signBinary% "%1"

if errorlevel 1 goto SigningFailed
goto SigningOK

:SigningFailed
echo Failed to sign file %1
exit 1

:NoSignInstaller

:SigningOK
```

`%signBinary%` is already defined in `packageAll.bat`

PowerEditor/installer/signCert/npp_sign_cert.pfx
```diff
@@ -0,0 +1 @@
+Replace with your actual cert or test cert
```
donho, line 1: Remove both `npp_sign_cert.pfx` and `SignTool.exe` files.
They are not necessary, and should not be in the repository.

PowerEditor/installer/nppSetup.nsi
```diff
@@ -43,6 +43,10 @@ OutFile ".\build\npp.${APPVERSION}.Installer.arm64.exe"
 OutFile ".\build\npp.${APPVERSION}.Installer.exe"
 !endif
 
+; Sign both installer and uninstaller
+!finalize        'sign-binaries.bat "%1"' = 0     ; %1 is replaced by the installer exe to be signed.
+!uninstfinalize  'sign-binaries.bat "%1"' = 0     ; %1 is replaced by the uninstaller exe to be signed.
```
donho, line 48: 1. Rename the file sign-binaries.bat to sign-installers.bat.
2. Remove the following lines from [packageAll.bat](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/packageAll.bat#L545) :
``` batch
if %SIGN% == 0 goto NoSignInstaller
%signBinary% !nppInstallerVar!
If ErrorLevel 1 goto End
%signBinary% !nppInstallerVar64!
If ErrorLevel 1 goto End
%signArmBinary% !nppInstallerVarArm64!
If ErrorLevel 1 goto End
:NoSignInstaller
```

donho (changes requested):

PowerEditor/installer/sign-binaries.bat
SinghRajenM: It is not that easy to get the `%signBinary%` from `packageAll.bat` -> `nsis compiler` -> then `sign-installers.bat`.

However, I have made some changes which should work (I could not test as I don't have cert and password handy). Also, I assume that you must have two env variable for `%NPP_CERT%` and `%NPP_CERT_PWD%`.

PowerEditor/installer/nppSetup.nsi
```diff
@@ -43,6 +43,12 @@ OutFile ".\build\npp.${APPVERSION}.Installer.arm64.exe"
 OutFile ".\build\npp.${APPVERSION}.Installer.exe"
 !endif
 
+; Sign both installer and uninstaller
+!ifdef SIGN_INSTALLER
```
SinghRajenM, line 47: This compiler flag is passed from `PackageAll.bat`. That means if installer is created by just compiling `nppSetup.nsi` (e.g. right click on file and compile with nsis), then both installer and uninstaller will not be signed.

PowerEditor/installer/packageAll.bat
```diff
@@ -477,10 +477,23 @@ If ErrorLevel 1 goto End
...
+  set nsisPath="%PROGRAMFILES%\NSIS\makensis.exe"
+)
+
+set sign_flag=-DSIGN_INSTALLER
+if [%SIGN%]==[] (
+  set "sign_flag="
+)
+
+rem build x86, x64 and arm64 installer
+%nsisPath% /V4 %sign_flag% nppSetup.nsi > nsis_compile_log_x86.txt
```
SinghRajenM, line 494: Dump nsis compilation log to file. Easy for debugging. 😊

SinghRajenM:
> Please use the suggestions after my first review for the following reasons:
> 
> 1. They are tested on my prod machine.
> 2. They are simpler & less code.

I wonder how you tested and how it was working. When I was testing , neither %SIGN% nor %signBinary% were coming to `sing_installer.bat`. I echoed them, both were empty. Moreover, I was not getting what is happening when nsis compiler was running. Hence, /V4 flag & log redirection to file.


``` batch
@ECHO OFF

if [%SIGN%] == [] goto NoSignInstaller
if not %SIGN% == 1 goto NoSignInstaller

ECHO Start signing file: %1
%signBinary% "%1"

if errorlevel 1 goto SigningFailed
goto SigningOK

:SigningFailed
echo Failed to sign file %1
exit 1

:NoSignInstaller

:SigningOK
```

donho:
> I wonder how you tested and how it was working. When I was testing , neither %SIGN% nor %signBinary% were coming to sing_installer.bat. I echoed them, both were empty. Moreover, I was not getting what is happening when nsis compiler was running. Hence, /V4 flag & log redirection to file.

I fill %SIGN% and other variables (password & certif file) under cmd.exe before launching packageAll.exe.
There's no way to pass these varibles to nsis GUI compiler, so when users use nsis GUI compiler, the installers are not signed.
If any user want to sign Notepad++ installer, they have to use packageAll.bat way.

SinghRajenM:
> > I wonder how you tested and how it was working. When I was testing , neither %SIGN% nor %signBinary% were coming to sing_installer.bat. I echoed them, both were empty. Moreover, I was not getting what is happening when nsis compiler was running. Hence, /V4 flag & log redirection to file.
> 
> I fill %SIGN% and other variables (password & certif file) under cmd.exe before launching packageAll.bat (`set SIGN=1`). There's no way to pass these varibles to nsis GUI compiler, so when users use nsis GUI compiler, the installers are not signed. If any user want to sign Notepad++ installer, they have to use packageAll.bat way.

Done!

donho:
Thank you @SinghRajenM !

Issue #14085, opened 2023-09-02
Test request to Notepad++ users: ADS issue fixed, but eventual network file saving regression 


donho:
In order to fix loosing Alternate Data Stream while saving issue (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1498), the saving routine in Notepad++ has been modified.
The modification could bring the regression for the network file saving (#10751, #10787).

To prevent the merging of fix of loosing ADS in Notepad++ from the regression, we have 3 binary candidates for testing:

* [Binary 1](https://ci.appveyor.com/api/buildjobs/173f9uoc5kif9qcf/artifacts/Notepad%2B%2B.x64.Release.exe) (from  @Klotzi111 #14029)
* [Binary 2](https://ci.appveyor.com/api/buildjobs/956lmca62a1i07ah/artifacts/Notepad%2B%2B.x64.Release.exe) (from  @donho #14039)
* [Binary 3](https://ci.appveyor.com/api/buildjobs/cgbwabo42idisbyr/artifacts/Notepad%2B%2B.x64.Release.exe) (from  @pnedev #14051)

A request to people who have had networking file saving problem: Pleas test above binaries so you won't encounter any problem in the next release 
Report format (sample):
```
Binary 1   OK
Binary 2   KO: the description of saving network file problem
Binary 3   OK
```

Thank you in advance for your help!

donho:
To:
@SamweisWeb @blackraz @CNCHoward @btanner2 @CarlosMOGoncalves @pjcs00 @GTR8000 @jacobwatkinsgit

I am sorry to request your help in batch, but since you guys have had the network file saving problem, and I really don't want such problem happens to you again when the new release is deployed.  

Could you please test the 3 canditate binaries and report the result? 
Thank you in advance for your help!

pjcs00:
Hi Don

Binary 1 - error is present, ie I consistently get 'Text file busy' if I
access a file within a few seconds or two of NPP saving it on a
network drive
Binary 2 - the error does not occur
Binary 3 - the error does not occur

To be clear, I am editing a web page source using NPP on Windows 11 and
saving it on a mapped drive which is a Raspberry Pi running Ubuntu, and
then accessing the web page on a browser on the Windows machine. The
browser reports an error, and the Apache (webserver) error log file on the
Pi shows the error as 'text file busy'. Refreshing the page as quickly as I
can press F5 seems always to succees.

[...]

Issue #13074, opened 2023-02-07
Documentation - old supported OS table


sXNA:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/SUPPORTED_SYSTEM.md
is old.

Windows 8.1 support ended
https://learn.microsoft.com/en-us/lifecycle/products/windows-81

Windows Vista doesn't work with latest notepad++ 8.4.9.
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ce85323a9e899e02ebc4c682298a7398cad99f60#commitcomment-89820581
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12513

And Windows 8 has "No" in 3rd column other OS have "NO"

alankilborn:
>Windows 8.1 support ended

It doesn't matter that M$ has ended support; the question here for the table is "does N++ run on the OS"

sXNA:
It can run but it should be still supported?

"The column supported contains the status of Notepad++ team supporting Notepad++ under the specific OS. It means if a bug is present under an OS no more supported but not under supported OS, then this issue won't be treated by the team."

ozone10:
@donho 
Should be Vista removed from "can be run" or try to figure out why it does not work.

rimrul:
> Windows Vista doesn't work with latest notepad++ 8.4.9.
> https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ce85323a9e899e02ebc4c682298a7398cad99f60#commitcomment-89820581

That doesn't seem entirely accurate. There's at least one issue that clearly says it does work on Vista](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13078)

rdipardo:
> @donho Should be Vista removed from "can be run" or try to figure out why it does not work.

v8.4.7 implicitly calls `::CoGetApartmentType`, as revealed by the import table:

<img width="756" alt="npp-v8 4 7-pe-header" src="https://user-images.githubusercontent.com/59004801/221376757-b8671899-7d2a-41f8-bd68-f0dd8f0564a7.png">

The [API docs] for `::CoGetApartmentType` say that Windows __*7*__ is the minimum supported desktop OS (with Windows Server 2008 R2 being the minimum supported server OS).

v8.4.6 and earlier doesn't import `::CoGetApartmentType`, so no dependency on any Win 7 APIs:

[...]

PR #14090, opened 2023-09-03
Refactoring file saving routine

Klotzi111:
This PR replaces the file saving routine to be able to fix certain problems.
This PR fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1498, fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5664, fixes https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133

It contains the following changes/additions:

1. The file is opened with the following disposition values. `allowOverwrite ? OPEN_ALWAYS : CREATE_NEW`
    1. This opens the file without truncating it
        1. This is to ensure that if we fail to write the entire file (or were unable to write anything) we lost as little data as possible. (data is still lost in those cases. Backups are required anyways and are independent of these changes)
    2. We "reuse" the existing file. So we are not losing any metadata or ADS
2. Opens the file with the additional flags `FILE_FLAG_WRITE_THROUGH | FILE_FLAG_NO_BUFFERING`
    1. This causes the Windows file cache to be bypassed for writes to that file and to do atomic writes to the sectors of the storage device
        1. This improves performance because the device does not need to read, update and write a sector
        2. This should prevent cases in which data is lost after the write happened when a fatal system failure occurs (like power loss) because there is not cached data to be lost (at least from the OS side. The hardware cache might still be active and can be lost)
    2. This has some drawbacks, In order to do this the data buffer must be aligned to the physical sector size. And because of that we need to create a buffer that meets these requirements and copy the entrire data over in memory. This can require a lot of memory if the file is big.
        1. A possible solution would be to have the internal buffers of N++ already aligned correctly (Note: physical sector size can change between devices. So a big alignment needs to be used to be safe)
3. Before the actual write operation the required file space is preallocated
    1. This is to ensure we can write the entire content and a OutOfDiskSpace error can not occur (or write nothing if the preallocation fails)
    2. This also reduces data fragmentation because all the required space is allocated in one go
4. The actual file size (more correct: the EOF mark) is set after a successful write
    1. This is required because we write in entire sector blocks and therefore the file can be larger than the wanted content after the write and before setting EOF
    3. If the process is killed before setting EOF there might be some NUL bytes at the end up to the next physical sectore aligned offset

**Testing is still required.**

My IDE automatically changed some whitespaces in some files. I hope this is ok.
Your comments are welcome.

@donho Can you also add this PR to https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14085 ?


Klotzi111:
I changed `-D_WIN32_WINNT=_WIN32_WINNT_VISTA` to `-D_WIN32_WINNT=_WIN32_WINNT_WIN8` because otherwise some structs in the `Windows.h` are not available.
This should be ok since https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/SUPPORTED_SYSTEM.md says only Windows 8.1 and above is supported

rdipardo:
> This should be ok since https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/SUPPORTED_SYSTEM.md says only Windows 8.1 and above is supported

The table is not clear, but the "supported" colummn actually means "supported __*by Microsoft*__". Even in that sense it's out-of-date; 8.1 went [out of support in January][0].

Since N++ version v8.4.7, the minimum required Win32 API level *at runtime* is Windows 7 (i.e. `_WIN32_WINNT_WIN7`): 

- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13074#issuecomment-1445191786
- https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12546#issuecomment-1631733555

I say "runtime" because you actually need a Windows __*11*__ SDK to build the project. The minimum Windows version supporting `DWMWA_USE_IMMERSIVE_DARK_MODE` is build number 22000, [according to Microsoft], not "19041", as the code mistakenly assumes here:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b62b77b3ea55629ff8fdbc856fe2cd3511f8df04/PowerEditor/src/NppDarkMode.cpp#L2889-L2894


[0]: https://learn.microsoft.com/en-us/lifecycle/products/windows-81
[according to Microsoft]: https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute#:~:text=use%20light%20mode.-,This%20value%20is%20supported%20starting%20with%20Windows%2011%20Build%2022000.,-DWMWA_WINDOW_CORNER_PREFERENCE%0AUse%20with

Klotzi111:
What does that mean for this PR? Do I need to change something?
Also Dark Mode is totally unrelated to this PR.

rdipardo:
@Klotzi111,

> What does that mean for this PR?

Assuming you mean [my earlier comment][0], the point is that targeting Windows 8 should be justified by something better than the outdated table in SUPPORTED_SYSTEM.md, which does not reflect the actual minimum requirements of the source code, nor even Microsoft's current support policy. If the PR doesn't build without it, it's up to @donho to decide whether that's a fatal blocker or not. What's in SUPPORTED_SYSTEM.md has no authority whatsoever.

If it were my choice to make, Windows 7 would already be the minimum target, since Vista is definitely unsupported in all scenarios; there's a small but devoted group of users on Windows 7 (see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8108#issuecomment-612923727); and major vendors like Mozilla still support it ([for now][1]).

By the time there's a good reason to make Windows 8 the *mandatory* minimum, Windows 10 will be end-of-life, and you might as well make *that* the minimum target, because practically nobody will be running older versions by then.


[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14090#issuecomment-1704401404
[1]: https://support.mozilla.org/en-US/kb/firefox-users-windows-7-8-and-81-moving-extended-support

Klotzi111:
> targeting Windows 8 should be justified by something better than the outdated table in SUPPORTED_SYSTEM.md

The justification is that without increasing the minimum version certain features of my new file saving routine might not be supported. Then a version check at runtime would be required to only use those features if they are supported.
In both cases the change to `-D_WIN32_WINNT=_WIN32_WINNT_WIN8` will be required because otherwise the structs for the features will not be available. Which they need to be in order to compile the code for using the affected features (even if they are behind a runtime version check)

> What's in SUPPORTED_SYSTEM.md has no authority whatsoever.

Why does it even exist then? It makes things only confusing.

So we are waiting on a decision from @donho then.

donho:
@Klotzi111 
> I changed -D_WIN32_WINNT=_WIN32_WINNT_VISTA to -D_WIN32_WINNT=_WIN32_WINNT_WIN8 because otherwise some structs in the Windows.h are not available.
This should be ok since https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/SUPPORTED_SYSTEM.md says only Windows 8.1 and above is supported


The main issue for this PR is not the problem above.
As I explained before, [it's very unlikely that Notepad++ integrate a heavy refactoring code of IO, for the sake of stability.](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14029#discussion_r1306643455)
I have to admit that you have done a huge coding effort to bring this PR to us.  I appreciate that very much. However, it's out of the question for me to go through several versions again for stabilizing the core feature of Notepad++. As a result I have to decline your PR with a lot of regret. Still, thank you very much for your coding effort.

Klotzi111:
> As I explained before, it's very unlikely that Notepad++ integrate a heavy refactoring code of IO, for the sake of stability.

I do not understand that. Because the saving routine is already not stable. See all the issues about saving.

I will need to use my fork then.
Also I will continue testing this saving routine with the use cases described in #14085

xomx:
@Klotzi111 
I tried this PR also in the testing of #14103 (saving to a WebDAV mapped folder), but your code failed in your `FileSystemHelper::getStorageInfo`, where the `GetFileInformationByHandleEx` failed with the `ERROR_INVALID_FUNCTION` (0x1), which resulted in errdlg:
PR#14090-error
The usual saving of files to local disk worked fine though.

Anyway, thanks for the interesting solution.

Klotzi111:
> but your code failed

Thats odd. Because the error `ERROR_INVALID_FUNCTION` is not documented for `GetFileInformationByHandleEx` on msdn.
That probably means that the driver for your network share does not implement this method (or the requested information class)

xomx:
> That probably means that the driver for your network share does not implement this method

Exactly my suspicion, there is a [similar issue](https://github.com/gradle/native-platform/issues/43).

PR #14104, opened 2023-09-06
Fix an eventual memory leak in Utf8_16.cpp

### LINKED ISSUE #14099: Memory leak in Utf8_16.cpp
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/b62b77b3ea55629ff8fdbc856fe2cd3511f8df04/PowerEditor/src/Utf8_16.cpp#L361

If the condition !m_pFile->write(buf, bufIndex * sizeof(utf16)) evaluates to true, the delete[] operation will never be executed, potentially leading to a memory leak. Additionally, throwing exceptions might also trigger memory leak scenarios.

donho:
Fix #14099

### DIFF
--- a/PowerEditor/src/Utf8_16.cpp
+++ b/PowerEditor/src/Utf8_16.cpp
@@ -70,10 +70,12 @@ u78 Utf8_16_Read::utf8_7bits_8bits()
 		else if ((*sx & (0x80+0x40+0x20)) == (0x80+0x40))
 		{					  // 110xxxvv 10nnnnnn, 11 bit character
 			ASCII7only=0;
-			if (std::distance(sx, endx) < 2) {
+			if (std::distance(sx, endx) < 2)
+			{
 				rv=0; break;
 			}
-			if ( (sx[1]&(0x80+0x40)) != 0x80) {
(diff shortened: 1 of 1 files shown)


cezariuszmarek:
I think it is better not to mix formatting changes with code changes.

PowerEditor/src/Utf8_16.cpp
```diff
@@ -335,38 +353,50 @@ bool Utf8_16_Write::writeFile(const void* p, size_t _size)
...
 				while ((bufIndex < bufSize) && iter8.canGet())
 					iter8.get(&buf [bufIndex++]);
 
-				if (bufIndex == bufSize || !iter8) {
-					if (!m_pFile->write(buf, bufIndex*sizeof(utf16))) return 0;
+				if (bufIndex == bufSize || !iter8)
+				{
+					if (!m_pFile->write(buf, bufIndex * sizeof(utf16)))
+					{
+						delete[] buf;
```
chcg, line 387: Seems this is the relevant change. Likely usage of a smartpointer could help to avoid such missing deletes.

PR #14134, opened 2023-09-10
Fix data loss issue due to no room on disk for saving

### LINKED ISSUE #5664: File is erased on save when there is no free space left on disk

### LINKED ISSUE #14089: Massive data loss and total incompetency when drive becomes full!
When the drive on which the file/current tab being edited becomes full:

- All of the contents of the current tab are DELETED regardless of the fact that the file has already been saved on the disk
- The file is ZEROED removing all data from the file.
- No effort is made to retain a ~4KB text file from RAM in spite of there being ~20GB+ available.
- AppData\Roaming\Notepad++\backup does not provide any means of recovery.
- Using the undo function did not help recover the data.
- I believe I was asked to reload the file and pressed a corresponding key before I realized the prompt had opened.

This is a massive total failure that just wiped out several hours of my work this morning! Granted it's easy code to rewrite though that is beside the point. The file is literally necessary for a paid project that will allow me to replace some low capacity drives.

I think it may have been a download that filled the disk though I'm not sure exactly as I'd imagine most programs check for available drive space _first_.

### Debug Information

Notepad++ v8.5.2   (64-bit)
Path : C:\MEDIA\OFFICE\Notepad++\notepad++.exe
Command Line : "private.php"
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.2846
Current ANSI codepage : 1252
Plugins : 
    DSpellCheck (1.5)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
This PR is based the solution provided in #5664 by @pnedev (the 2nd one):
> Examine the free space left on disk BEFORE opening the file for writing with fopen...

VeraCrypt is used to crate a 10 MB small drive for testing to validate this PR.

Fix #5664, fix #14089

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1114,7 +1114,7 @@ bool FileManager::deleteBufferBackup(BufferID id)
 
 std::mutex save_mutex;
 
-SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool isCopy)
+SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR* filename, bool isCopy)
 {
 	std::lock_guard<std::mutex> lock(save_mutex);
 
@@ -1135,6 +1135,29 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 		{
 			::GetLongPathName(fullpath, fullpath, MAX_PATH);
 		}
+	}
+
+	const wchar_t* filePath = buffer->getFullPathName();
+	wchar_t dir[MAX_PATH];
+	wcscpy_s(dir, MAX_PATH, filePath);
+	::PathRemoveFileSpecW(dir);
+	ULARGE_INTEGER freeBytesForUser;
+	ULARGE_INTEGER totalBytesForUser;
+	 
+	BOOL getFreeSpaceRes = ::GetDiskFreeSpaceExW(dir, &freeBytesForUser, &totalBytesForUser, nullptr);
+	if (getFreeSpaceRes != FALSE)
+	{
+		int64_t fileSize = buffer->getFileLength();
+		if (fileSize >= 0)
+		{
+			freeBytesForUser.QuadPart += fileSize;
+		}
+		// else file doesn't exist
+		
+
+		// determinate if free space is enough
+		if (freeBytesForUser.QuadPart <= buffer->docLength())
+			return SavingStatus::NotEnoughRoom;
 	}
 
 	if (PathFileExists(fullpath))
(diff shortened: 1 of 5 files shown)


donho:
@pnedev @xomx and anyone's review are welcome.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1137,6 +1137,29 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
 		}
 	}
 
+	const wchar_t* filePath = buffer->getFullPathName();
```
pnedev, line 1140: Is there a chance `buffer->getFullPathName()` to be empty (in case the file is completely new and has never been written to disk before)?
Shouldn't we use `fullpath` instead the new `const wchar_t* filePath` ?

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1137,6 +1137,29 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
...
+	wchar_t dir[MAX_PATH];
+	wcscpy_s(dir, MAX_PATH, filePath);
+	::PathRemoveFileSpecW(dir);
+	ULARGE_INTEGER freeBytesForUser;
+	 
+	BOOL getFreeSpaceRes = ::GetDiskFreeSpaceExW(dir, &freeBytesForUser, nullptr, nullptr);
+	if (getFreeSpaceRes != FALSE)
+	{
+		int64_t fileSize = buffer->getFileLength();
```
pnedev, line 1149: If we are saving a new copy of the same file under different name then the original file will not be overwritten and its size should not be added to the free space calculation, right? That is if a new file copy is made this way...

PowerEditor/src/ScintillaComponent/Buffer.cpp
donho: You are right about it.

PowerEditor/src/ScintillaComponent/Buffer.cpp
donho: Do you mean "Save As..." or "Save a Copy As..."? If you do, then what you worry about won't happen, because the "filename" argument will be a new name so the size won't be added to the free space calculation.

OTOH, your remark makes me test with the both commands mentioned above to "Save as an existing file". It is rejected because this PR doesn't consider other existing file as the target. So in case of source & destination are not the same file,  and if source file length is smaller than destination one, it could be saved but operation is not permitted.

I do not want to manage the above case, for the reason that: 1. it'll make the code complex 2. it'll be more safe while no room left.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1136,6 +1136,28 @@ SavingStatus FileManager::saveBuffer(BufferID id, const TCHAR * filename, bool i
...
+		if (fileSize >= 0 && lstrcmp(fullpath, currentBufFilePath) == 0) // if file to save does exist, and it's an operation "Save" but not "Save As"
+		{
+			// if file exists and the operation "Save" but not "Save As", its current length should be considered as part of free room space since the file itself will be overrrided 
+			freeBytesForUser.QuadPart += fileSize;
+		}
+
+		// determinate if free space is enough
+		if (freeBytesForUser.QuadPart <= buffer->docLength())
```
xomx, line 1158: should be `freeBytesForUser.QuadPart < buffer->docLength()` only, otherwise when the disk is full and you open a file on it and e.g. change one char to another one (file retains the same size), you will not be able to save that file...

donho:
Thank you @pnedev & @xomx For your review!

PR #14142, opened 2023-09-13
Reduce number of document list redraws on reload.

Orden4:
Made as a follow up to https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14080#issuecomment-1716522593
This is basically what I wanted to do from the start, but I couldn't figure out how to control redraws myself.

Example on a list of about 80 documents: https://streamable.com/r5j5gc (that one slight hiccup is caused by recording software).
Like this it seems to go at a rate of about 15 items per second with no notable slowdown as the list gets longer/farther down, it's another step up from the previous version though not massively (~1.5x at most for my PC). However, because of these heavily reduced redraws, the document list no longer flickers when you are moving items like this. Previously you'd usually see flickers at the bottom or the scroll bar, as well as sometimes just the list in general.

I ended up completely disabling the redraws during the `reload` method, because if you trigger redraw from that method then the selected item won't have correctly updated yet, which then causes the darkened selection to always be off-by-one during the scrolling. From what I could tell every single thing that triggers a reload will immediately trigger a redraw afterwards for one reason or another anyway, so it didn't seem harmful to me, though by all means suggest a better approach (adding a `::SendMessage(_hSelf, WM_PAINT, 0, 0);` at the end seems fine?).

### DIFF
--- a/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
+++ b/PowerEditor/src/WinControls/VerticalFileSwitcher/VerticalFileSwitcherListView.cpp
@@ -169,6 +169,7 @@ void VerticalFileSwitcherListView::initList()
 
 void VerticalFileSwitcherListView::reload()
 {
+	// Suppress redraws for performance. We target _hParent to prevent scroll bar flickering.
 	::SendMessage(_hParent, WM_SETREDRAW, false, 0);
 	removeAll();
 	initList();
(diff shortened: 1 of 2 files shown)


Orden4:
The failed build does not appear to be related to any change made? It seems like it just randomly stalled.

Yaron10:
@Orden4,

Thank you for https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14080 and for this PR. 👍 

I think this PR requires some fine-tuning.

**STR:** 
Open the Documents List.
Press (and hold) Ctrl+N.

**Result**:

תמונה

Yaron10:
@SlugFiller,

Thank you as well. 👍

Orden4:
Fixed the issue and added some comments to explain the various redraws.
Added a `redrawItems` method which is now called at the end of both `reload` and `activateItem`. I'm still not really convinced whether it's needed for `reload`, but it didn't seem to matter performance/visual wise, so I figured to better play it safe. I tried a few other ways to open files/add to the list, but was not able to find any further problems.

Yaron10:
@Orden4,

Thank you for the fix.
Let's wait for Don's code-review.

PR #14146, opened 2023-09-14
Update langs.model.xml, Update PowerShell and Assembly Sintax

GabrielFrigo4:
# PowerShell:
 - Add some "INSTRUCTION WORD"

# Assembly:
 - Add AVX Registers;
 - Add missing instructions
 - Add missing macros
 - Add missing directives

# Things I couldn't add
 - Use (`) as a delimiter for strings in the Assembly that neither (') nor ("). But that I leave for a future update, when I figured out how to add this functionality :/


GabrielFrigo4:
Any feedback? If I did something wrong I would like to know the problem...

donho:
@GabrielFrigo4 
Sorry for the late review.
Could you provide the refs/links of all the keywords you added ?

donho:
Any Assembly expert can review this PR?

GabrielFrigo4:
> @GabrielFrigo4 Sorry for the late review. Could you provide the refs/links of all the keywords you added ?

# PowerShell (7.X)
I don't remember the PowerShell references very well, but it was something about comparison operators from microsoft's own website, I think it's this link:
[PowerShell Comparison Operators](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.3)

# Assembly X86-16, X86-32 and X86-64 (Intel Architecture)
[Instruction](https://www.felixcloutier.com/x86/)
[Registers](https://wiki.osdev.org/CPU_Registers_x86-64): I don't remember if the syntax already had everything that this link shows, but if not I probably used this link
[AVX Registers](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions)

# Assemblers MASM, NASM, FASM and GAS
To use the Intel assembly, you need to use the Assemblers, and the four main ones are MASM, NASM, FASM and GAS. And I put some keywords that these assemblers use (mainly NASM and FASM, since in this syntax there was already MASM, and GAS is mainly for Linux...) The links are:
[MASM](https://learn.microsoft.com/en-us/cpp/assembler/masm/microsoft-macro-assembler-reference?view=msvc-170): [Directives reference](https://learn.microsoft.com/en-us/cpp/assembler/masm/directives-reference?view=msvc-170), [Symbols Reference](https://learn.microsoft.com/en-us/cpp/assembler/masm/symbols-reference?view=msvc-170) and [Operators Reference](https://learn.microsoft.com/en-us/cpp/assembler/masm/operators-reference?view=msvc-170)
[NASM](https://www.nasm.us/xdoc/2.16.01/html/nasmdoc0.html)
[FASM](https://flatassembler.net/docs.php): [Manual](https://flatassembler.net/docs.php?article=manual) and [Windows](https://flatassembler.net/docs.php?article=win32)
[GAS](https://mentebinaria.gitbook.io/assembly/programando-junto-com-c/sintaxe-do-gas) It's in Portuguese... I didn't find any similar content in English... :/

rdipardo:
@GabrielFrigo4 

> [GAS](https://mentebinaria.gitbook.io/assembly/programando-junto-com-c/sintaxe-do-gas) It's in Portuguese... I didn't find any similar content in English... :/

My [UDL for GAS] has an ample keyword list, copied from a SciTE contributor's `gas.properties` file, which you can download [**here**](https://sourceforge.net/p/scintilla/feature-requests/1072/#5657).

> To use the Intel assembly, [. . .] the four main ones are MASM, NASM, FASM and GAS.

GAS is a dialect of [AT&T assembly](https://developer.ibm.com/articles/l-gas-nasm/#basic-structure), not Intel.

[UDL for GAS]: https://gist.githubusercontent.com/rdipardo/f659a3ebe7b4772e581e7b2d32d82f9f/raw/4f2709a9c237549812e86e4fd77c794c620c6824/GAS.udl.xml

donho:
@GabrielFrigo4 
So the keywords you're adding are the different varieties of Assembly, right?
I'm wondering if it's a good thing to put all different Assembly instructions all together - there will be the performance issue, without mentioning users confusion when they work on Intel architecture Assembly but miss type an instruction to one of GAS.

What's your opinion @rdipardo ?

GabrielFrigo4:
> @GabrielFrigo4 So the keywords you're adding are the different varieties of Assembly, right? I'm wondering if it's a good thing to put all different Assembly instructions all together - there will be the performance issue, without mentioning users confusion when they work on Intel architecture Assembly but miss type an instruction to one of GAS.
> 
> What's your opinion @rdipardo ?

No, I only added the keywords Intel Assembly (known as X86), I didn't add anything from ARM and RISCV... But I've also added some keywords that assemblers use (assemblers are equivalent to compilers, they take the code and convert it to binary) that are not standardized. Although Intel has its own syntax, there is for example the AT&T syntax that GAS uses (although it is also possible to use Intel syntax in GAS). In addition, there are also macros and directives specific to assemblers, such as "extern" in NASM, ".globl" in GAS, "executable" in FASM and ".ALLOCSTACK" in MASM...

But what I added was mainly the new AVX instructions, the new AVX registers and the other X86 instructions that were missing (and in general the names of these instructions are standardized, and sometimes they may be different in AT&T syntax, but in this case I didn't add these AT&T differences).

detail, "assembly" is different from "assembler"

GabrielFrigo4:
One of the notable differences in assembler syntax is that in NASM it is also possible to use the backquote (`) to declare strings, in addition to double quote (") and single quote (') that are already used normally...

But since I don't know how to add this in the notepad++ syntax, I left it out :(

donho:
@GabrielFrigo4 
I notice that you use your master to do this PR.
Here are some rules for the contribution, please follow the guidelines:

Furthermore, [Merge branch 'notepad-plus-plus:master' into master](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14146/commits/bd73f55a474be9ada7cdeb82a223949214751455) is not necessary and should not be done. It will cause the conflict while merging.

Therefore this PR will be closed - please create a new PR from the current master with a new branch.
Please list also the keywords you add, follow by the references (links) so it's clear to me which words are added in this PR.

PR #14152, opened 2023-09-16
keep missing files on load session

### LINKED ISSUE #12079: Loading Current Session with missing files does not meet user expectation
### Description of the Issue

3. Move `b.txt` to `be.txt`

### Desired Behavior

1. Notepad++ launches
3. For each file, in session:
   - if file exists, open it
   - if file missing,
4. (normal editor usage)
7. Notepad++ closes
8. Next run

### Debug Information

```
Notepad++ v8.4.4   (64-bit)
Command Line : a.txt b.txt
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 20H2
OS Build : 19042.1889
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
```

molsonkiko:
Many users find it undesirable that Notepad++ simply forgets about files that are missing with no backup at the time a session is loaded.

This PR fixes that issue by creating a new empty file for any file that was not found when loading the session and then deleting that empty file immediately so the user gets a "would you like to keep this deleted file" prompt.

THINGS THAT COULD PROBABLY BE IMPROVED WITH THIS PR:
1. I don't know how this could happen, but there's some concern that __deleting the new empty file that was just created could lead to data loss in some corner case__.
2. This approach is slower and clunkier than a theoretical alternate entirely in-memory approach that does not manipulate the filesystem. However, my attempts that involved creating a `new 1`  - type buffer and renaming it to the name of the missing file did not achieve the desired result.

TO TEST THIS PR:
1. Open a file (let's call it `file A.txt`) in one edit view. Add some text and save it.
2. Open a file (call it `file B.txt`) in the second edit view. Add some text and save it. *The purpose of this is to verify that the PR works in both edit views, not just the main one.*
3. Close Notepad++.
4. Delete or rename both `file A.txt` and `file B.txt`
5. Reopen Notepad++. You should get prompts asking you if you want to keep `file A.txt` and `file B.txt`, since those files no longer exist on disk.
6. Restore `file A.txt` and `file B.txt` __while Notepad++ is still open__.
7. For each of those files, ensure that you get the usual prompt saying that the file has been modified outside Notepad++ and asking if you want to discard the changes made inside Notepad++.

Fix #12079

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -2140,7 +2140,17 @@ bool Notepad_plus::loadSession(Session & session, bool isSnapshotMode, bool shou
 		}
 		else
 		{
-			lastOpened = BUFFER_INVALID;
+			// file doesn't exist and has no backup; probably it's a file that was deleted while NPP was closed
+			// as a hack to ensure that the file can be restored without issues, we will create an empty file
+			// with that name, open it normally, and then delete it. This way the user gets a missing file prompt for the file.
+			if (!PathFileExists(pFn) && MainFileManager.createEmptyFile(pFn))
+			{
+				lastOpened = doOpen(pFn, false, false, session._mainViewFiles[i]._encoding, session._mainViewFiles[i]._backupFilePath.c_str(), session._mainViewFiles[i]._originalFileLastModifTimestamp);
+				if (lastOpened->docLength() == 0 && PathFileExists(pFn))
+				{
+					MainFileManager.deleteFile(lastOpened);
+				}
+			}
 		}
 		if (isWow64Off)
 		{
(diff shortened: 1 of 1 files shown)


Yaron10:
@molsonkiko,

Thank you for your work.

> This PR fixes that issue by creating a new empty file for any file that was not found when loading the session and then deleting that empty file immediately so the user gets a "would you like to keep this deleted file" prompt.

1. Why not use the approach & message on trying to open a non-existing file from the "Recently Closed Files" list?
Adding the following line might be helpful.
```
Not creating the file will modify your Session file.
```

2. The issue is relevant also to Session files created by "Save Session...".

PowerEditor/src/NppIO.cpp
```diff
@@ -2140,7 +2140,17 @@ bool Notepad_plus::loadSession(Session & session, bool isSnapshotMode, bool shou
 		}
 		else
 		{
-			lastOpened = BUFFER_INVALID;
+			// file doesn't exist and has no backup; probably it's a file that was deleted while NPP was closed
+			// as a hack to ensure that the file can be restored without issues, we will create an empty file
+			// with that name, open it normally, and then delete it. This way the user gets a missing file prompt for the file.
+			if (!PathFileExists(pFn) && MainFileManager.createEmptyFile(pFn))
```
donho, line 2146: If we arrive here, PathFileExists(pFn) is surely FALSE, `!PathFileExists(pFn)` is not needed here and it shouldn't be used again.

donho: But `we will create an empty file with that name, open it normally, and then delete it. This way the user gets a missing file prompt for the file.` is a very bad workaround which shouldn't even be considered.

donho:
The solution of this PR for https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12079 is far from being perfect.

molsonkiko:
I understand your reasoning. I'm still interested in solving this issue, so I think I'll pursue a different approach based on creating a `new 1` - type document and making that track the missing file. Not sure if I can get it to work, but it seems like the most promising approach at this point.

PR #14156, opened 2023-09-18
Fix  Wrong Categories in Shortcuts Mapper

### LINKED ISSUE #13285: Wrong Categories in Shortcuts Mapper
**STR:**
Settings -> Shortcuts Mapper.

**Result:**
Some commands are listed under the wrong Category.
A few examples:

תמונה
תמונה
תמונה

***

Notepad++ v8.4.9   (64-bit)
Build time : Feb 24 2023 - 10:04:40
Path : C:---notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Version : 20H2
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : 
    ComparePlus (1.1)
    HTMLTag (1.3.6)
    MenuIcons (2.0.2)
    PythonScript (2)
    _CustomizeToolbar (4.2)

donho:
Fix #13285

### DIFF
--- a/PowerEditor/src/WinControls/shortcut/shortcut.cpp
+++ b/PowerEditor/src/WinControls/shortcut/shortcut.cpp
@@ -1226,44 +1226,46 @@ intptr_t CALLBACK ScintillaKeyMap::run_dlgProc(UINT Message, WPARAM wParam, LPAR
 
 void CommandShortcut::setCategoryFromMenu(HMENU hMenu)
 {
+	NativeLangSpeaker* pNativeSpeaker = NppParameters::getInstance().getNativeLangSpeaker();
+
 	if ( _id >= IDM_WINDOW_SORT_FN_ASC and _id <= IDM_WINDOW_SORT_FS_DSC)
-		getMainMenuEntryName(_category, hMenu, "Window", L"Window");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "Window", L"Window");
 	else if ( _id >= IDM_VIEW_GOTO_ANOTHER_VIEW and _id <= IDM_VIEW_LOAD_IN_NEW_INSTANCE)
-		getMainMenuEntryName(_category, hMenu, "view", L"View");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "view", L"View");
 	else if (_id == IDM_EDIT_LTR || _id == IDM_EDIT_RTL)
-		getMainMenuEntryName(_category, hMenu, "view", L"View");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "view", L"View");
 	else if (_id == IDC_PREV_DOC || _id == IDC_NEXT_DOC)
-		getMainMenuEntryName(_category, hMenu, "view", L"View");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "view", L"View");
 	else if (_id == IDM_FORMAT_TODOS || _id == IDM_FORMAT_TOUNIX || _id == IDM_FORMAT_TOMAC)
-		getMainMenuEntryName(_category, hMenu, "edit", L"Edit");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "edit", L"Edit");
 	else if (_id == IDM_EDIT_AUTOCOMPLETE || _id == IDM_EDIT_AUTOCOMPLETE_CURRENTFILE || _id == IDM_EDIT_FUNCCALLTIP ||
 		_id == IDM_EDIT_AUTOCOMPLETE_PATH || _id == IDM_EDIT_FUNCCALLTIP_PREVIOUS || _id == IDM_EDIT_FUNCCALLTIP_NEXT)
-		getMainMenuEntryName(_category, hMenu, "edit", L"Edit");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "edit", L"Edit");
 	else if (_id == IDM_LANGSTYLE_CONFIG_DLG)
-		getMainMenuEntryName(_category, hMenu, "settings", L"Settings");
+		pNativeSpeaker->getMainMenuEntryName(_category, hMenu, "settings", L"Settings");
 	else if (_id == IDM_MACRO_STARTRECORDINGMACRO  ||_id == IDM_MACRO_STOPRECORDINGMACRO  || _id == IDM_MACRO_RUNMULTIMACRODLG ||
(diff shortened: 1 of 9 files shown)


donho:
@Yaron10 
Could you check if it's fixed completely?

Yaron10:
@donho,

Great. 👍 
Thank you for your work.

***
תמונה

Wouldn't it be better to associate those two commands with View?
***

תמונה

Still Setting instead of Settings.
***

Shouldn't the XML localization files be updated?

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -1227,31 +1227,43 @@ intptr_t CALLBACK ScintillaKeyMap::run_dlgProc(UINT Message, WPARAM wParam, LPAR
...
+		_category = TEXT("View");
+	else if (_id == IDM_EDIT_LTR || _id == IDM_EDIT_RTL)
+		_category = TEXT("View");
+	else if (_id == IDM_FORMAT_TODOS || _id == IDM_FORMAT_TOUNIX || _id == IDM_FORMAT_TOMAC)
+		_category = TEXT("Edit");
+	else if (_id == IDM_EDIT_AUTOCOMPLETE || _id == IDM_EDIT_AUTOCOMPLETE_CURRENTFILE || _id == IDM_EDIT_FUNCCALLTIP ||
+		_id == IDM_EDIT_AUTOCOMPLETE_PATH || _id == IDM_EDIT_FUNCCALLTIP_PREVIOUS || _id == IDM_EDIT_FUNCCALLTIP_NEXT)
+		_category = TEXT("Edit");
+	else if (_id == IDM_LANGSTYLE_CONFIG_DLG)
+		_category = TEXT("Setting");
```
Yaron10, line 1243: Settings.

PowerEditor/src/WinControls/shortcut/shortcut.cpp
```diff
@@ -1227,31 +1227,43 @@ intptr_t CALLBACK ScintillaKeyMap::run_dlgProc(UINT Message, WPARAM wParam, LPAR
...
 	else if ( _id < IDM_ABOUT)
-		_category = TEXT("Lang");
+		_category = TEXT("Language");
 	else if ( _id < IDM_SETTING)
 		_category = TEXT("About");
 	else if ( _id < IDM_TOOL)
-		_category = TEXT("Setting");
+		_category = TEXT("Settings");
 	else if ( _id < IDM_EXECUTE)
 		_category = TEXT("Tool");
```
Yaron10, line 1266: Tools.

Yaron10:
@donho,

All commands are under the correct category now. 👍 
Thank you again.

Could you please explain to me the purpose of the last two commits?
The categories _were localized_ before them.

The only change needed was replacing:
```xml
				<WindowCategory name="Window"/>
				<FileCategory name="File"/>
				<EditCategory name="Edit"/>
				<SearchCategory name="Search"/>
				<ViewCategory name="View"/>
				<FormatCategory name="Format"/>
				<LangCategory name="Lang"/>
				<AboutCategory name="About"/>
				<SettingCategory name="Setting"/>
				<ToolCategory name="Tool"/>
				<ExecuteCategory name="Execute"/>
```
with
```xml
				<WindowCategory name="Window"/>
				<FileCategory name="File"/>
				<EditCategory name="Edit"/>
				<SearchCategory name="Search"/>
				<ViewCategory name="View"/>
				<EncodingCategory name="Encoding"/>
				<LanguageCategory name="Language"/>
				<AboutCategory name="About"/>
				<SettingsCategory name="Settings"/>
				<ToolsCategory name="Tools"/>
				<RunCategory name="Run"/>
```

Getting the categories from the main menu might be a good idea.
But then - the code getting them from `XCategory name` could and should be removed, shouldn't it?

Yaron10:
@donho,

> All the "*Category" strings are not using. I've checked in the codes - so they were not localized for sure.

What do you mean?

Here is v8.5.7.
תמונה

And this is after removing `<FileCategory name="קובץ"/>` from the localization file.
תמונה

Yaron10:
[Make categories in the Shortcut Mapper dialog translatable](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f7fcab4c219b7187ce9d6ca300c96f06f8f2e877).

donho:
> [Make categories in the Shortcut Mapper dialog translatable](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f7fcab4c219b7187ce9d6ca300c96f06f8f2e877).

Thank you for reminding. However, it doesn't work for me in v8.5.7

> And this is after removing <FileCategory name="קובץ"/> from the localization file.

Even with the whole PR ? I have tested it though. It does work on my PC.
Surely we don't deliver the PR with my PC...

Yaron10:
@donho,

> Even with the whole PR ?

I was referring to v8.5.7 (as a proof that those strings do work).
With this PR the categories are localized after removing the "*Category" strings (Except About).

> However, it doesn't work for me in v8.5.7.

Very interesting. It does work for me on Windows 10 & Windows 7.
Any idea what it can be? Did you try it with a fresh portable NPP? :)

@alankilborn, @rdipardo, @rddim,
Are the categories localized on your machine with v8.5.7?

xomx:
> Are the categories localized on your machine with v8.5.7?

Yes, they are (Win10, Win11).

Yaron10:
@xomx,

Thank you!

donho:
OK, I see why it didn't work for me.
On debug mode `doLocalConf.xml` is present, and there were already 3 obsolete localization files (English, French & Mandarin) in the localization folder beside the binary, whereas I have copied these 3 updated files into %APPDATA%\Notepad++\ - the debug binary load the obsolete localization (in which "*Category" entries are absent.

donho:
But while I searched "WindowCategory", "FileCategory"... etc in the code, I found nothing - which was kind of obscure IMO.
And since we can get these translated string from the menu directly, why should we store them in localization XML file?
Hence the following commits.

Yaron10:
@donho,

> And since we can get these translated string from the menu directly, why should we store them in localization XML file?

This is a question of an amateur to a master. :)
Getting the strings from the menu should apparently be a good choice for two reasons:
1. No double entries in the XML localization files.
2. The code _apparently_ should be shorter.

The code with this PR is longer and seems more complex.
Do you still think it's worth it?

Thank you.

Yaron10:
@donho,

Good points. I trust your judgement. 👍 
Thank you again for your time and work.

donho:
@Yaron10 
Thank you for your review!

PR #14167, opened 2023-09-21
Fix: unselected lines may become hidden

### LINKED ISSUE #14166: Unselected lines may become hidden
### Description of the Issue
Suppose that the content of the document alternates between visible individual lines and hidden blocks of text. In this situation, when the user tries to hide a single line, only that should become hidden. Instead, program hides all the lines that alternate between already hidden blocks of text.

**Step 1.** Copy and paste the following text in a new document:
```
Line 1
 Line 2    (hide)
  Line 3
   Line 4  (hide)
    Line 5
   Line 6  (hide)
  Line 7
 Line 8    (hide)
Line 9
```

**Step 2.** Hide labeled lines. It should look like this:  
A
**Step 3.** Select **only** the line number 5 and hide it.

### Expected Behavior
Line 1, 3, 7 and 9 should remain visible:  
B

### Actual Behavior
Only line 1 and 9 are visible:  
C

### Debug Information
```
Notepad++ v8.5.7   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.3448
Current ANSI codepage : 1252
Plugins : none
```

Adrian104:
Fix #14166

To solve this issue, lambda `removeMarker` can be set to only remove just begin or end marker. By doing so, the process of merging adjacent hidden sections is limited and as a result, lines that shouldn't be hidden remain visible.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -3663,9 +3663,9 @@ void ScintillaEditView::hideLines()
 	int scope = 0;
 	bool recentMarkerWasOpen = false;
 
-	auto removeMarker = [this, &scope, &recentMarkerWasOpen](size_t line)
+	auto removeMarker = [this, &scope, &recentMarkerWasOpen](size_t line, int markerMask)
 	{
-		auto state = execute(SCI_MARKERGET, line);
+		auto state = execute(SCI_MARKERGET, line) & markerMask;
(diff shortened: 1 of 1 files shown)


Yaron10:
@Adrian104,

👍 
Thank you for your work.

I hardly use "Hide Lines" in real-life. I've only tested the particular case in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14166.

PR #13245, opened 2023-02-26
Replace recent file ValueDlg with edit controls

ozone10:
remove ValueDlg class

image

fix #13244 
fix #13248

PR #14175, opened 2023-09-22
Add Find InSelection configurable size and possible disable

### LINKED ISSUE #12639: Check "In selection" checkbox by default if replace windows was opened while text was selected
### Description of the Issue
When the replace window is opened while text is selected (with control-H for example) the In selection checkbox is not checked.
Imho it would be better if it would be checked by default in that situation. If I have text selected and do a replace I want to replace in that selected text. Other editors I use select it by default which works better than the current notepad++ way.

1.  Select some text
2.  control-H

### Expected Behavior
"In Selection" checkbox checked by default

### Actual Behavior
"In Selection" checkbox not checked by default

### Debug Information
Notepad++ v8.4.6   (64-bit)
Drive\accounts\www.github.com.txt"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.2251
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)

### LINKED ISSUE #13677: Text strings longer than 1024 characters are not pre-filled in search dialog
### Description of the Issue
Selecting a string of text longer than 1024 characters doesn't pre-fill anymore the CTRL+F search dialog after upgrading from 8.4.9 to 8.5.3.

1. Copy this text in a new file:
`this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_is_a_long_repeated_string_that_exceeds_1204_characters___this_last_part_now_is_at_1032_characters`
2. Select all the text and press CTRL+F to open the search dialog.

### Expected Behavior
The search field in the search dialog should be **pre-filled with the selected text**.

### Actual Behavior
The search field in the search dialog that opens is **empty**.
I tested also the previous 8.5.x versions, and it seems the issue started with 8.5.0.

### Debug Information
Notepad++ v8.5.3   (64-bit)
Build time : May 15 2023 - 06:09:36
Command Line : "C:\Users\XXX\Desktop\SEARCH BUG EXAMPLE.txt"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.2846
Current ANSI codepage : 1252
Plugins : 
    HexEditor (0.9.12)
    JsonTools (4.8.1)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

### LINKED ISSUE #14108: Disable search option "in selection" constantly
Hi,
is there a way to disable the search 'in selection' option always?
we always search in whole document and this option disturbs our search,
since it is enabling by itself.
thanks for assistance.
cheers mike

alankilborn:
Submitters of the above issues should set the new setting to `0`.

Fix #14108, fix #13677, fix #12639


Yaron10:
@alankilborn,

👍 
Thank you for your work.
Another 👍 for the RTL tooltip. :)

**STR:**
Disable auto-checking.
Select some characters & activate the Find dialog.
Manually check "In selection".
Remove the selection. 
Activate the Find dialog.

**Result:**
"In selection" is unchecked.

**Expected:**
"In selection" should be disabled but **checked**.

If auto-checking is disabled, so should be auto-**unchecking**.
Let the user handle that.

***

The following issues are not directly related to this PR.

**STR:**
Check the two "Fill Find Field" options.
Remove all selections.
Place the caret on a word.
Invoke the Find dialog.

**Result:**
There's a selection, but "In selection" is disabled.
(See https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14108#issuecomment-1718358902).

___
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14108#issuecomment-1719653099.

alankilborn:
@Yaron10 said:

>If auto-checking is disabled, so should be auto-unchecking.
Let the user handle that.

I tried that and it had an awkward feeling when you want to uncheckmark _In selection_ and you can't because it is disabled.  You then have to go back to your document, make a (possibly) meaningless selection, just to be able to remove the checkmark.

I'm not opposed to it, but let's see what Don says.

Yaron10:
> I tried that and it had an awkward feeling when you want to uncheckmark In selection and you can't because it is disabled.

I'm afraid we have another misunderstanding. :)
It's **disabled** with this PR. I just don't want NPP to uncheck it.

alankilborn:
When I said "I tried that" I meant as I was developing, not when I try the code in this PR.
I don't know if that clarifies ANYTHING but I thought I should say it. :-)
Our misunderstandings continue, apparently. :-)

alankilborn:
> Could you please try the exact steps?

I saw no differences in the STR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1731634569 versus the (first) STR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1731530833

Should there be a difference in the two stated sets of STR?
If so, I don't see it.
If not, then I'm confused as to why you'd repeat the same exact STR.

Yaron10:
> I tried that and it had an awkward feeling when you want to uncheckmark In selection and you can't because it is disabled.

That made me wonder if we tried the same steps.
And the **Expected:** part is clearer now. I hope so. :)

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1493,38 +1493,42 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 			if (LOWORD(wParam) == WA_ACTIVE || LOWORD(wParam) == WA_CLICKACTIVE)
 			{
 				Sci_CharacterRangeFull cr = (*_ppEditView)->getSelection();
-				intptr_t nbSelected = cr.cpMax - cr.cpMin;
+				LRESULT nbSelected = (*_ppEditView)->execute(SCI_COUNTCHARACTERS, cr.cpMin, cr.cpMax);
```
donho, line 1496: I have some weird behaviour (the count of character not exact) while testing this PR. I'm not sure it's due to SCI_COUNTCHARACTERS. 
However, since the goal of the PR is adding "disable In selection auto-checked" feature, not change current behaviour, please restore `intptr_t nbSelected = cr.cpMax - cr.cpMin;`

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1493,38 +1493,42 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 			if (LOWORD(wParam) == WA_ACTIVE || LOWORD(wParam) == WA_CLICKACTIVE)
 			{
 				Sci_CharacterRangeFull cr = (*_ppEditView)->getSelection();
-				intptr_t nbSelected = cr.cpMax - cr.cpMin;
+				LRESULT nbSelected = (*_ppEditView)->execute(SCI_COUNTCHARACTERS, cr.cpMin, cr.cpMax);
 
-				_options._isInSelection = isCheckedOrNot(IDC_IN_SELECTION_CHECK)?1:0;
-				int checkVal = _options._isInSelection?BST_CHECKED:BST_UNCHECKED;
```
donho, line 1499: Restore the above lines please.

donho (changes requested):
There are probably other places to modify.
To make code review easier, please consider the following suggestions firstly;

donho:
@Yaron10 
> It's **disabled** with this PR. I just don't want NPP to uncheck it.

I see your point. The current PR is buggy to me

image

Let's wait the modification for the change request from @alankilborn.
Once the PR is bug free, then we will talk about your request.

Yaron10:
@donho,

> I see your point.
Once the PR is bug free, then we will talk about your request.

👍 

> The current PR is buggy to me

Current master is also buggy: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14108#issuecomment-1719653099.
I think it's hard to properly test this PR if that case is not addressed as well.

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -5596,65 +5611,100 @@ intptr_t CALLBACK SearchingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 
 		case WM_COMMAND:
 		{
-			switch (wParam)
+			if (LOWORD(wParam) == IDC_INSELECTION_THRESHOLD_EDIT)
 			{
-				case IDC_CHECK_FILL_FIND_FIELD_WITH_SELECTED:
+				if (HIWORD(wParam) == EN_KILLFOCUS)
```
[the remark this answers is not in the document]
Yaron10: @donho & @alankilborn,

I haven't tried it with `EN_CHANGE`.
But I'm wondering if it's `EN_KILLFOCUS` vs. `EN_CHANGE` or another variant of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14108#issuecomment-1719653099.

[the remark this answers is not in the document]
alankilborn: In order to create a numeric edit box control in the preferences, I went looking for an existing one to model upon.  I found the ones in _Recent Files History_:

image

These use `EN_KILLFOCUS`, hence my usage in currrent PR.  I will change current PR, but I'd think that `EN_KILLFOCUS` usage for  _Recent Files History_ should also change -- should I create an issue for that?

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -5596,65 +5611,100 @@ intptr_t CALLBACK SearchingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 
 		case WM_COMMAND:
 		{
-			switch (wParam)
```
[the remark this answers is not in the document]
alankilborn: OK

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -354,14 +354,16 @@ IDD_PREFERENCE_SUB_SEARCHING DIALOGEX 115, 10, 460, 205
...
-    CONTROL         "Select Word Under Caret when Nothing Selected", IDC_CHECK_FILL_FIND_FIELD_SELECT_CARET, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 52, 31, 275, 10
-    CONTROL         "Use Monospaced font in Find dialog (Need to restart Notepad++)", IDC_CHECK_MONOSPACEDFONT_FINDDLG, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 52, 350, 10
-    CONTROL         "Find dialog remains open after search that outputs to results window", IDC_CHECK_FINDDLG_ALWAYS_VISIBLE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 67, 350, 10
-    CONTROL         "Confirm Replace All in All Opened Documents", IDC_CHECK_CONFIRMREPLOPENDOCS, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 82, 350, 10
-    CONTROL         "Replace: Don't move to the following occurrence", IDC_CHECK_REPLACEANDSTOP, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 97, 350, 10
-    CONTROL         "Search Result window: show only one entry per found line", IDC_CHECK_SHOWONCEPERFOUNDLINE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 37, 112, 350, 10
+    GROUPBOX        "When Find Dialog is Invoked && Activated", IDD_FILL_FIND_FIELD_GRP_STATIC, 31, 4, 323, 65, BS_CENTER
```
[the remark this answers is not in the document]
alankilborn: OK

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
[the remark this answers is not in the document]
Yaron10: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13245#issuecomment-1450230191.

PowerEditor/src/Parameters.h
```diff
@@ -137,6 +137,9 @@ const int COPYDATA_FULL_CMDLINE = 3;
 
 #define NPP_STYLING_FILESIZE_LIMIT_DEFAULT (200 * 1024 * 1024) // 200MB+ file won't be styled
 
+const int FINDREPLACE_INSELECTION_THRESHOLD_DEFAULT = 1024;
+const int FINDREPLACE_INSELECTION_THRESHOLD_MAX = 1024;
```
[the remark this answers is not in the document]
alankilborn: OK.  (I originally had a different "max" numeric value, that's how this originated).

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
alankilborn: What is not liked here?

Surely the addition of whitespace around `?` and `:` in (existing) line 1499 is not a problem...

That leaves the removal of `?1:0` from (existing) line 1498 as what is objected to.
But that is nonsensical as `isCheckedOrNot()` returns `bool` and `options._isInSelection` is a bool.
So there is no pure numeric type here and thus no need to involve numeric constants `1` and `0`.
I'm truly lost...

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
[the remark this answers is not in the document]
alankilborn: @Yaron10, there wasn't much explanation there, but I think I understand.  In that case, as soon as the `3` is processed (by EN_CHANGE) the list of recent files is truncated to be of length 3, so that when the `0` is added and 30 is processed (by the following EN_CHANGE), there isn't any data remaining with which to re-lengthen the list.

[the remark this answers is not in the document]
Yaron10: @alankilborn,

Very well explained.
It was easier to understand when that binary (using `EN_CHANGE`) was available.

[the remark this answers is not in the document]
donho: > but I'd think that EN_KILLFOCUS usage for Recent Files History should also change -- should I create an issue for that?

@alankilborn 
If you can reproduce an issue, then please do it.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: The goal is to make minimum modification so the code review will focus on the essential stuff. Except the critical stake overflow or leak memory issue, it's suggested to keep the original code intact. 
[In short: The easier the code review is, the better the chance your pull request will get accepted.](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests)

donho:
@Yaron10 
> Current master is also buggy: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14108#issuecomment-1719653099.
I think it's hard to properly test this PR if that case is not addressed as well.

Just check it and I've got impression that this "feature" was i;ple;ented intentionally:

```cpp
				_options._isInSelection = isCheckedOrNot(IDC_IN_SELECTION_CHECK)?1:0;
				int checkVal = _options._isInSelection?BST_CHECKED:BST_UNCHECKED;

				if (!_options._isInSelection)
				{
					if (nbSelected <= FINDREPLACE_INSEL_TEXTSIZE_THRESHOLD)
					{
						checkVal = BST_UNCHECKED;
						_options._isInSelection = false;
					}
					else
					{
						checkVal = BST_CHECKED;
						_options._isInSelection = true;
					}
				}
```

(ref: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/33de57fe1dbb6164373258c297a98cc438b78c5f#diff-b5a0ff4b5a02de99b597d5c71ea278e3c7332dd8aec008c8b78604d06792a064)

My question is: What's the correct behaviour in your opinion?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
alankilborn: > it's suggested to keep the original code intact.

So, bad but working old code is allowed to stand, when there is an opportunity to make it better?
(Surely no one can argue that the modifications I suggested to these two lines don't make it better?)

No one would make a PR to only make such a change, but while someone is in there making changes to the general area, and theoretically testing the entire change well, IMO such changes make a great deal of sense.

donho: I'll do a refactoring after. In the meanwhile, let's focus on the essential part.

alankilborn: If you will refactor after, can I propose a list of such changes, with rationale for each?  Then, much like typos, you can pick-and-choose from that list.

donho: @alankilborn 
Please do your list in an issue, then ping me. I will assign to myself for sure. Also there might be some points that we don't agree, so we can talk directly in the issue.

Yaron10:
@donho,

I hope I understand you correctly.

If NPP should auto-check/uncheck "In selection", I expect it to be consistent.
When the Find dialog is activated: it should be checked if the selection >= 1024 (or custom size) , and unchecked otherwise. 

***

I also think that fixing the following issue (brought up in my first post here), would make it easier to test the implementation as a whole.

STR:
Check the two "Fill Find Field" options.
Remove all selections.
Place the caret on a word.
Invoke the Find dialog.

Result:
There's a selection, but "In selection" is disabled.

Yaron10:
There's actually a simpler and more fundamental case I don't understand in the auto-check/uncheck approach.

Select 1024.
Open the Find dialog (-> "In selection" is checked).
Manually **uncheck** it. 
Invoke the Find dialog.

Result:
"In selection" is re-checked.

Is that expected? Is it even an accessible behavior?

alankilborn:
> Select 1024. Open the Find dialog (-> "In selection" is checked). Manually **uncheck** it. Invoke the Find dialog.
> 
> Result: "In selection" is re-checked.

I'd say this is expected.

Yaron10:
I don't know what to expect any more. :)

donho:
> If NPP should auto-check/uncheck "In selection", I expect it to be consistent.
When the Find dialog is activated: it should be checked if the selection >= 1024 (or custom size) , and unchecked otherwise.

I will fix this part. 

> STR:
> Check the two "Fill Find Field" options.
> Remove all selections.
> Place the caret on a word.
> Invoke the Find dialog.
> 
> Result:
> There's a selection, but "In selection" is disabled.

Indeed, "In selection" should not be disabled.

donho:
> I will fix this part.

@Yaron10 
Just fixed it & committed into master: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/972329fb0cb411fcb43a23184a9aa368b2a0581b

Could you test it please?

PowerEditor/src/Parameters.h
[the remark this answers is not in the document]
donho: BTW we can use the existing one `FINDREPLACE_INSEL_TEXTSIZE_THRESHOLD`.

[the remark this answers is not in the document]
alankilborn: Wow.  That's bad. :-(

donho:
@alankilborn 
Could you please merge the current master then rebase your PR and push it with --force please?
Sorry for the inconvenience.

Yaron10:
@donho,

👍 
That issue is fixed. Thank you for your work.

***

@alankilborn, @pryrt (and obviously @donho),

I've been trying to find some explanation to the behavior in the following **common cases**.
Unfortunately I've failed. I apologize for bringing them up (again), but I'm sure we're all seeking the best implementation & solution.

**1.**
Select 1024.
Open the Find dialog (-> "In selection" is checked).
Manually **uncheck** it.
Switch to another tab (without activating the Find dialog).
Switch back to the previous tab, and activate the Find dialog.

Result:
"In selection" is re-checked.

**2.**
Select 900.
Open the Find dialog (-> "In selection" is unchecked).
Manually **check** it.
Switch to another tab (without activating the Find dialog).
Switch back to the previous tab, and activate the Find dialog.

Result:
"In selection" is re-unchecked.

**How can we ignore the user's explicit choice?** 

***

1024: bytes vs. characters.
1. The Selection part in the status-bar indicates characters. Isn't 1024 _bytes_ confusing?
2. If you have one tab containing English text, and another containing French text - you might get a different result (checked/unchecked) on selecting the same number of _characters_. 
Isn't that confusing as well?

alankilborn:
@Yaron10 said:

> 1024: bytes vs. characters.

I was going to use this PR to discuss this aspect, but I was holding back on such discussion until the PR is more mature.

Why wait for maturity?:  It is much easier to illustrate the "bytes vs. characters" thing when you can set the in-selection threshold size to e.g. 8, and then show some small data, than it is to deal with something as large as the current threshold (1024).

Another point regarding this:  It was one of the reasons the first version of the PR used `SCI_COUNTCHARACTERS` rather than subtracting the selection start _position_ from the selection end _position_.

alankilborn:
Yaron10 said:

>How can we ignore the user's explicit choice?

The user needs to understand that moving input focus between the editor and the _Find_ window retriggers the logic for _In selection_ auto-check / auto-uncheck.  I suppose the location of the new control (in the groupbox) and the proposed tooltip of this PR (or maybe we can make the tooltip text even better) helps the user's mind.

It is my guess that you are excited by this PR and a release of the code containing some version of it:  So that you can go in and set the setting to zero!  :-)

Yaron10:
> The user needs to understand

I apologize for failing to understand it.
And believe me - I've been trying really hard.
Believe me too that I'm not trying to win an argument here. 

> t is my guess that you are excited by this PR and a release of the code containing some version of it: So that you can go in and set the setting to zero! :-)

I haven't failed to understand that. :)
(I changed the behavior for my personal use a long time ago).

donho:
@Yaron10 
> How can we ignore the user's explicit choice?

Yes, I can see what you're trying to show us.
That's the reason of this PR - to add the option of auto-checking if it's bothering for user.

Yaron10:
@donho,

OK. :)
I can only wish good luck to @alankilborn on the rest of the work here.

alankilborn:
I did the rebase, and made minimal changes just to resolve the conflicts.  I still need to address all of the earlier changes @donho wanted.  My spare time is limited at the moment, but I will keep chipping away at it as I can.

Yaron10:
Changing the value and activating the Find dialog - the check-state reflects the old value.
It seems that @donho was right about `EN_CHANGE`. 👍 
(`WM_ACTIVATE` is processed before `EN_KILLFOCUS`?).

alankilborn:
@Yaron10, I suspect you will like these.  :-)

---

Here's a oddball scenario number 1, in 8.5.7:

- have editor window with a single stream selection > 1024
- invoke/activate _Find_ window (_In selection_ becomes enabled and checkmarked)
- uncheck _In selection_
- activate editor window, eliminate the selection
- activate _Find_ window **by using the mouse to checkmark** _In selection_ (it gets checkmarked and then disabled)

It shouldn't be checkmarked because there is no selection in editor window.

---

Here's a oddball scenario number 2, in 8.5.7:

- have editor window with a single stream selection > 1024
- invoke/activate _Find_ window (_In selection_ becomes enabled and checkmarked)
- activate editor window, eliminate the selection
- activate _Find_ window **by using the mouse to uncheck** _In selection_ (it gets disabled but stays checkmarked)

It shouldn't be checkmarked because there is no selection in editor window.

PowerEditor/src/NppCommands.cpp
```diff
@@ -1247,14 +1247,24 @@ void Notepad_plus::command(int id)
...
+					{
+						filling = false;
+					}
 				}
 			}
+			if (filling)
+			{
+				_pEditView->getGenericSelectedText(str, strSize, nppGui._fillFindFieldSelectCaret);
+				_findReplaceDlg.setSearchText(str);
+			}
```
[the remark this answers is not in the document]
alankilborn: I changed that code for a reason, but I can't seem to replicate the problem I saw now.  But OK, let's have this PR _only_ introduce a disable and settable threshold while we evaluate it.  If oddities are discovered during testing, then we'll make further logic changes.

PowerEditor/src/Parameters.h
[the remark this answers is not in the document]
alankilborn: At the risk of getting into trouble, I left the DEFAULT part on. :-)  Let's focus on behavior and agree on naming later?

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
alankilborn: I agree to defer related discussion to a later point on this; restoring position subtraction code for now.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1493,38 +1493,42 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
...
 					checkVal = BST_UNCHECKED;
 					_options._isInSelection = false;
 					nbSelected = 0;
 				}
+				
 				enableFindDlgItem(IDC_IN_SELECTION_CHECK, nbSelected != 0);
+				
 				// uncheck if the control is disable
-				if (!nbSelected)
+				if (nbSelected == 0)
```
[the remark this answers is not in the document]
alankilborn: Further discussion later; retored...

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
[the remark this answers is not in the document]
alankilborn: @donho : Discussion with @Yaron10 revealed no changes needed to Recent Files History section.

Changed the code in this PR to use EN_CHANGE.

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -5596,13 +5611,44 @@ intptr_t CALLBACK SearchingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
...
+
 			switch (wParam)
 			{
 				case IDC_CHECK_FILL_FIND_FIELD_WITH_SELECTED:
 				{
 					nppGUI._fillFindFieldWithSelected = isCheckedOrNot(IDC_CHECK_FILL_FIND_FIELD_WITH_SELECTED);
-					::EnableWindow(::GetDlgItem(_hSelf, IDC_CHECK_FILL_FIND_FIELD_SELECT_CARET), nppGUI._fillFindFieldWithSelected ? TRUE :FALSE);
-					if (!nppGUI._fillFindFieldWithSelected) 
+					::EnableWindow(::GetDlgItem(_hSelf, IDC_CHECK_FILL_FIND_FIELD_SELECT_CARET), nppGUI._fillFindFieldWithSelected ? TRUE : FALSE);
+					if (!nppGUI._fillFindFieldWithSelected)
```
alankilborn, line 5651: Hmm, I didn't mean to affect whitespace here...I hope it is OK and doesn't contribute too much to meaningless changes to review...

Yaron10:
Changing the value and activating the Find dialog - the check-state reflects the new value. 👍 
EditBox: Selecting all and deleting results in auto-filling the old value. isn't it slightly unconventional?

> I suspect you will like these.

Nice catches. I did like them. :)

alankilborn:
>EditBox: Selecting all and deleting results in auto-filling the old value. isn't it slightly unconventional?

Sigh.  I suppose I will go in search of a different "model" for integer entry of a parameter into an edit box...

If that's our biggest problem with the current state of the PR (...I don't think it is the biggest...), I'll gladly fix it and move on. :-)

BTW, I don't really like the way this PR works in its current state.  The current state of it tries to add in ONLY the disable and the adjustable size, no other logic changes (and I think to make it usable, more changes are going to be needed).  Hint: Set the size to something small (e.g. 8) and use it for a bit of time (with the other 2 checkboxes in _When Find Dialog Is Invoked_ checkmarked, of course).

Yaron10:
> EditBox: Selecting all and deleting results in auto-filling the old value. isn't it slightly unconventional?

Personally I don't mind that. If Don does, you'd probably need to use KILLFOCUS _as well_.

> BTW, I don't really like the way this PR works in its current state...

If I understood the discussion correctly, more changes _are expected_ later.

donho:
> BTW, I don't really like the way this PR works in its current state. The current state of it tries to add in ONLY the disable and the adjustable size, no other logic changes (and I think to make it usable, more changes are going to be needed). Hint: Set the size to something small (e.g. 8) and use it for a bit of time (with the other 2 checkboxes in When Find Dialog Is Invoked checkmarked, of course).

This feature's existed since long time ago, so it's hard to change its behaviour. OTOH, the current state of this PR can make every user happy: they can enable it with the size they want, or disable it completely.
What kind of change are you expecting exactly?

alankilborn:
>What kind of change are you expecting exactly?

I don't really like this:

Setup:
image

In editing tab, have text of `123456789` and don't select anything but put caret in that word and press Ctrl+f to obtain:

image

and then back in editor you see:

image

Click title bar of N++ main window, then click title bar of _Find_ window and see:

image

I guess I'm uncomfortable with _In selection_ not being enabled initially, but then with no other data change (other than moving focus among windows), _In selection_ becomes enabled.

It seems like the new setting (when non-zero) and _Select word under caret..._ compete with each other.

Yaron10:
@alankilborn,

> // Move it here so that "In Selection" can be correctly enabled/disabled on "WM_ACTIVATE" (if there was no selection and "getGenericSelectedText()" has set one).

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14108#issuecomment-1718358902.

***

@donho,

STR:
Disable auto-checking.
Select some characters & activate the Find dialog.
Manually check "In selection".
Remove the selection.
Activate the Find dialog.

Result:
"In selection" is unchecked.

Expected:
"In selection" should be disabled but checked.

If auto-checking is disabled, so should be auto-unchecking.
Let the user handle that.

> Once the PR is bug free, then we will talk about your request.

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1732585490.

Yaron10:
@alankilborn,

```cpp
		case IDM_SEARCH_FIND:
		case IDM_SEARCH_REPLACE:
		case IDM_SEARCH_MARK:
		{
			const int strSize = FINDREPLACE_MAXLENGTH;
			TCHAR str[strSize]{};

			const NppGUI & nppGui = NppParameters::getInstance().getNppGUI();
			if (nppGui._fillFindFieldWithSelected)
				_pEditView->getGenericSelectedText(str, strSize, nppGui._fillFindFieldSelectCaret);		// Move it here so that "In Selection" can be correctly enabled/disabled on "WM_ACTIVATE" (if there was no selection and "getGenericSelectedText()" has set one).

			bool isFirstTime = !_findReplaceDlg.isCreated();

			DIALOG_TYPE dlgID = FIND_DLG;
			if (id == IDM_SEARCH_REPLACE)
				dlgID = REPLACE_DLG;
			else if (id == IDM_SEARCH_MARK)
				dlgID = MARK_DLG;
			_findReplaceDlg.doDialog(dlgID, _nativeLangSpeaker.isRTL());

			if (nppGui._fillFindFieldWithSelected)
				_findReplaceDlg.setSearchText(str);
```

Can't it be relevant even if 1024 is set?

alankilborn:
>Can't it be relevant even if 1024 is set?

I use an example size of 8 because you can more easily get your head around the effects from the feature when the size is something you can visually count.
Anything discussed here is (or should be) relevant to any size from 1 to 1024 (so I will probably continue to use 8 in any examples I present).

donho:
@alankilborn 
> I don't really like this:

The above case you provided is for testing. IRL no user will set such value (8). They might reduce the value from 1024 to 256, or they disable this feature with value 0.

Yaron10:
@donho,

> IRL no user will set such value (8).

**STR:**
Check the two "Fill Find Field" options.
Disable auto-checking (set the value to 0).
Remove all selections.
Place the caret on a word.
Invoke the Find dialog.

**Result:**
There's a selection, but "In selection" is disabled.

***

> My scenario is different, but it shows the 0 value (disabling this feature) doesn't work

:+1: 
So both cases should be fixed.

alankilborn:
>I believe it doesn't meet the specification we have defined.

OK, well, let me keep evaluating it and working on it.

alankilborn:
>no user will set such value (8)

Now that you've brought up that pertinent fact, should the valid range for the setting be changed from `1 - 1024` to something different, to prevent non-real-life values from being entered?

alankilborn:
>8

BTW, 8 will be very useful and illustrative when we start talking about bytes versus characters in the selection.

donho:
@alankilborn 
> Now that you've brought up that pertinent fact, should the valid range for the setting be changed from `1 - 1024` to something different, to prevent non-real-life values from being entered?

My opinion is don't bother with that. If user want to set 6, 7, 8 or 9, it's their freedom to have wired behaviour.
OTOH, I suggest the following text ***"The maximum value is 1024. Set the value to 0 to disable auto-checking"*** instead of ***"Set the value to 0 to disable auto-checking; enabled range is 1 - 1024."*** to not guide people to set such value.

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -5575,6 +5575,21 @@ intptr_t CALLBACK SearchingSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPAR
 			::SendDlgItemMessage(_hSelf, IDC_CHECK_CONFIRMREPLOPENDOCS, BM_SETCHECK, nppGUI._confirmReplaceInAllOpenDocs, 0);
 			::SendDlgItemMessage(_hSelf, IDC_CHECK_REPLACEANDSTOP, BM_SETCHECK, nppGUI._replaceStopsWithoutFindingNext, 0);
 			::SendDlgItemMessage(_hSelf, IDC_CHECK_SHOWONCEPERFOUNDLINE, BM_SETCHECK, nppGUI._finderShowOnlyOneEntryPerFoundLine, 0);
+			::SetDlgItemInt(_hSelf, IDC_INSELECTION_THRESHOLD_EDIT, nppGUI._inSelectionAutocheckThreshold, 0);
+
+			NativeLangSpeaker* pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
+			generic_string tipText = pNativeSpeaker->getLocalizedStrFromID("searchingInSelThresh-tip", L"Number of selected characters in edit zone to automatically check the 'In selection' checkbox when the Find dialog is activated. Set the value to 0 to disable auto-checking; enabled range is 1 - $INT_REPLACE$.");
```
donho, line 5581: Please no `$INT_REPLACE$` please. Such place holder is for localization.
If we change this value in the future, we modify also in the localization files.

donho (changes requested):
Fix the disabling not working bug and not use $INT_REPLACE$.

alankilborn:
@donho 

@Yaron10 wants these to be true statements:

* "In selection" can be disabled and checked at the same time 
* If auto-checking is disabled, so should be auto-unchecking

I can't tell definitively from the (lengthy) above discussion if you (Don) agree with this or not, so please let me know.

donho:
> I can't tell definitively from the (lengthy) above discussion if you (Don) agree with this or not, so please let me know.

@alankilborn 
I think it's reasonable demand because if auto-checking In selection is disable by 0, all the automatic checking mechanism should be disabled. 
To be more precised:
If the value is 1-1024, auto-checking `In selection` is activated, the behaviour should be the same as the one in current master, but it reacts with the value set by users, instead of current fixed value 1024.
Otherwise, the value is 0. The auto-checking of `In selection` should be disable, and user keeps his/her checked or unchecked state, whenever the `In selection` checkbox is enabled or disabled.

alankilborn:
If _Count_ is pressed while _In selection_ is checkmarked and disabled:

image

then what should happen?

donho:
@alankilborn 
If Count is pressed while In selection is checkmarked and disabled:
then what should happen?

It should be consider **not** In selection - since there's no selection.

alankilborn:
>It should be consider not In selection - since there's no selection

OK, just wanted to be sure, since this is a bit of an odd condition.

To the user, since the box is checked, the thinking is that it is going to do some sort of selection search.
I suppose user must take note of the disabled state, and thus know that search will not be a selection search.

Further:
When I see a disabled but checked checkbox, I think that the program is forcing me into the checked behavior.  That's not the case here.

alankilborn:
If the new setting is 0 and _In selection_ is checkmarked when Notepad++ is exited, when Notepad++ is restarted and _Find_ is opened, _In selection_ should be checkmarked.

alankilborn:
If _In Selection_ is checkmarked and disabled, user might have a rectangular selection active, and might think that his _Count_ operation will be on that rectangular selection.

Yaron10:
@alankilborn,

> To the user, since the box is checked, the thinking is that it is going to do some sort of selection search.
...
If In Selection is checkmarked and disabled, user might have a rectangular selection active, and might think that his Count operation will be on that rectangular selection.

Why? - Disabled is basically like "element-not-displayed".

***

> If the new setting is 0 and In selection is checkmarked when Notepad++ is exited, when Notepad++ is restarted and Find is opened, In selection should be checkmarked.

My first thought was: "In selection" is a light-flexible per-search preference that shouldn't be saved across sessions.
But you can argue: why do you consider it lighter than "Match case"?
So, it's a nice point. :)

***

Your "oddball scenarios" in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1736388016 can actually result in a search being executed in the wrong range.

Replacing
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0d8a8dff7dc150db1ce4a8536994fa642e9969ef/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2179
with
```cpp
_options._isInSelection = (::IsWindowEnabled(---) && isCheckedOrNot(IDC_IN_SELECTION_CHECK));
```
might solve the issue.
_I haven't tested it._



Thank you for your work and patience.

alankilborn:
>Disabled is basically like "element-not-displayed".

I respectfully disagree.  :-)

>Thank you for your ... patience.

This one is tricky.  A struggle is to try to force changes in without changing existing code too much.  Sometimes this approach results in unreadable code, when considering the existing code wasn't very readable in the first place.  My decision was to rewrite entirely the WM_ACTIVATE section.  The result is much better,I think, but I don't know how well it will be received.  This is another one that I'm perhaps sorry I got involved in.  :-)

Yaron10:
> To me, disabled and checked In selection means "I'm going to do it in the selection, and you have no choice about that".

I respectfully disagree. :-)

> My decision was to rewrite entirely the WM_ACTIVATE section. The result is much better,I think, but I don't know how well it will be received.

Why not try it if you think it's better? Don would let you know if he didn't like it.

Thank you again. 👍 

I can hardly keep my eyes open now.
Good night.

alankilborn:
@Yaron10 

>_options._isInSelection = (::IsWindowEnabled(---) && isCheckedOrNot(IDC_IN_SELECTION_CHECK));
might solve the issue.
I haven't tested it.

I tried it with no improvement to the "oddball" scenarios.  But thanks for the suggestion.

donho:
@alankilborn 
> OK, just wanted to be sure, since this is a bit of an odd condition.
> 
> To the user, since the box is checked, the thinking is that it is going to do some sort of selection search. I suppose user must take note of the disabled state, and thus know that search will not be a selection search.
> 
> Further: When I see a disabled but checked checkbox, I think that the program is forcing me into the checked behavior. That's not the case here.
> 
> Also: An alternative would be to do nothing if the checkbox is checked and disabled when the _Count_ button is pressed, but as you didn't discuss that I won't consider it.

If user set this value to 0, I believe they know what they are doing.

Yaron10:
@alankilborn,

If auto-checking is disabled, the check-state in the "oddball scenarios" are expected and correct.
But I've found that in your first case the search is executed in-selection instead of entire-doc.

I've just tested my proposed code change and it does solve _that issue_.
`case IDC_IN_SELECTION_CHECK` is processed after `WM_ACTIVATE`.

***

Disabled-checked.
תמונה

There are more examples.

alankilborn:
>Disabled-checked....(Smart Highlighting)

Maybe that one isn't fair as it has a controlling _Enable_ and the slave controls are merely holding their current values waiting for the next enable...
Or maybe it is a fair comparison.
Regardless, I'm over my original objection, so no need to go further with it.

alankilborn:
Pushed more changes for further discussion and hopefully tryout by at least @Yaron10

alankilborn:
Consider the code in this PR and the following text in a Notepad++ tab, encoded as UTF-8 and having Windows line-endings:

```
one 💙
two 💙💙
thr 💙💙💙
fou 💙💙💙💙
fiv 💙💙💙💙💙
six 💙💙💙💙💙💙
sev 💙💙💙💙💙💙💙
eig 💙💙💙💙💙💙💙💙
nin 💙💙💙💙💙💙💙💙💙
```

and this setup:

image

Make this selection:

image

and press Ctrl+f to then see:

image

If _In selection_ is checkmarked (from a prior operation), uncheck it.

Back in editor window, make this selection:

image

Press Ctrl+f and see:

image

Note that _In selection_ has been autocheckmarked when the selection was made up of 2 characters (and the settings said 8 is the minimum size).

---

This is not a bug with this PR per se.  This behavior can be replicated with 8.5.7, it is just more difficult to illustrate because there the fixed size is 1024.  but let's do it:

In 8.5.7, have only this text in a tab, , encoded as UTF-8 and having Windows line-endings:

```
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙💙
```

If you do a Ctrl+a to select-all, you will see in the status bar: `Sel : 1,024 | 16` ; so... 1024 characters.

Do a Ctrl+f and in the _Find_ window, _In selection_ will be (automatically) checkmarked.  Uncheck it.

Now, select ONLY the first five lines.  The status bar will say `Sel : 320 | 5` ; the number of characters selected is now 320.

Press Ctrl+f and notice that _In selection_ will be (automatically) checkmarked.

Conclusion in 8.5.7 code:  320 characters caused autochecking of _In selection_, when the fixed threshold is 1024.

Of course, this is because one 💙character takes 4 bytes of "position", and because a selection-end-position minus a selection-start-position is used to determine "size".

---

So... to the user, what is this "minimum size" supposed to be:

image

Eight "characters" in the document, or eight "bytes" in the document?

Yaron10:
@alankilborn,

👍 
Thanks again for your work. And patience! :)

Disabled auto-checking seems to be working like a charm.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/39068b2b55b7293d10b4fd65144dcf2f6966b78c/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2757

Shouldn't be called if auto-checking is disabled.

***

**STR:**
Set the value (in the new EditBox) to 3.
Select 3 characters (bytes? :) ).
Open the Find dialog (-> enabled & checked).
Change the value to 6.
Activate the Find dialog.

**Result:** 
"In selection" is still enabled & checked.

***

This is not directly related to this issue & PR.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/39068b2b55b7293d10b4fd65144dcf2f6966b78c/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L4097

Any point in that line?

alankilborn:
>This is not directly related to this issue & PR.
> Line 4097
>Any point in that line?

I haven't worked on the saving/restoring of _In selection_ check / disabled state over N++ restart cycle yet.
It's coming...

Yaron10:
> It's coming...

To stay or not to stay? This is the question. :)

alankilborn:
> **STR:** Set the value (in the new EditBox) to 3. Select 3 characters (bytes? :) ). Open the Find dialog (-> enabled & checked). Change the value to 6. Activate the Find dialog.
> 
> **Result:** "In selection" is still enabled & checked.

In 8.5.7 this type of action (but with something >= 1024 first and something smaller than 1024 second) won't uncheck _In selection_, but I suppose we are OK with that.

The next push will address specifically this problem.

Yaron10:
👍 
Great.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/39068b2b55b7293d10b4fd65144dcf2f6966b78c/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L2757

Should be on the next commit?

Yaron10:
@alankilborn,

> If the new setting is 0 and In selection is checkmarked when Notepad++ is exited, when Notepad++ is restarted and Find is opened, In selection should be checkmarked.

>> My first thought was: "In selection" is a light-flexible per-search preference that shouldn't be saved across sessions.
But you can argue: why do you consider it lighter than "Match case"?
So, it's a nice point. :)

> I haven't worked on the saving/restoring of In selection check / disabled state over N++ restart cycle yet.

This is obviously up to you and Don.
I just want to clarify my reply: it IS a nice point, but I still think it's not needed.

Thank you again.

alankilborn:
>I just want to clarify my reply: it IS a nice point, but I still think it's not needed.

Hmm, I think if user has disabled the autocheckmarking and ALWAYS searches "in selection", then wouldn't user be annoyed that he has to go and recheck the checkbox every time he restarts Notepad++?  He should check it one time, and then it stays checked forevermore.

Yaron10:
> I think if user has disabled the autocheckmarking and ALWAYS searches "in selection"...

That would be quite rare, wouldn't it?
I was thinking of a session containing a file with a selection.

I usually search in entire-doc. I'm afraid that even in the session case that **I** checked "In selection", It might still cause some confusion on the next startup.

But it's not a solid argument. :)

alankilborn:
>That would be quite rare, wouldn't it?
> but I still think it's not needed.

OK, you've convinced me to leave it the way it is.
Another plus:  Less code changes.

Yaron10:
Your original point of view is certainly good, and some users might indeed want it to be saved across sessions.
As you (and Don) understand.

PowerEditor/src/NppCommands.cpp
```diff
@@ -1246,11 +1252,10 @@ void Notepad_plus::command(int id)
 				dlgID = MARK_DLG;
 			_findReplaceDlg.doDialog(dlgID, _nativeLangSpeaker.isRTL());
 
-			const NppGUI & nppGui = (NppParameters::getInstance()).getNppGUI();
 			if (nppGui._fillFindFieldWithSelected)
 			{
-				_pEditView->getGenericSelectedText(str, strSize, nppGui._fillFindFieldSelectCaret);
-				if (lstrlen(str) <= FINDREPLACE_INSEL_TEXTSIZE_THRESHOLD)
+				if ((nppGui._inSelectionAutocheckThreshold == 0) ||
+					(lstrlen(str) < nppGui._inSelectionAutocheckThreshold))
```
donho, line 1258: Selection Text to Find what should work independently. 
The condition of `_findReplaceDlg.setSearchText(str);` should be dissociated with **In selection**.
As a result, these lines should be restored.

donho (changes requested):
Please restore the old behaviour of **Selection into Find What**.

PowerEditor/src/NppCommands.cpp
donho: Sorry for not being clear.
What I mean is use the following code:
```cpp

				if (lstrlen(str) <= FINDREPLACE_INSELECTION_THRESHOLD_DEFAULT)
				{
					_findReplaceDlg.setSearchText(str);
				}
```
instead of 
```
				if ((nppGui._inSelectionAutocheckThreshold == 0) ||
					(lstrlen(str) < nppGui._inSelectionAutocheckThreshold))
				{
					_findReplaceDlg.setSearchText(str);
				}
```
to preserve the old behaviour of "selection to Find What" feature:

image

Yaron10: @donho,

Could you please list the exact STR?
(Master vs. PR).

Thank you.

donho: @Yaron10 
Master : Ctrl-F copy the selected text (less 1024) into Find What field.
PR: when set the value to 8,  Ctrl-F does not copy the selected text (more than 8) into Find What field.

The ***Ctrl-F copy the selected text (less 1024) into Find What field*** should not be interfered by this PR. That's my point.

alankilborn: @donho OK, will make this change in the next commit.

Yaron10: @donho & @alankilborn,

Thank you for the clarification and work.

Selected "Close" not being set in "Find what" is indeed weird.
But the latest commit creates two major issues:

1. If auto-checking is disabled (0), 1024 is irrelevant and even 1025-selected should be set in "Find what".
\- Why should `if ((nppGui._inSelectionAutocheckThreshold == 0)` be removed?
2. If auto-checking is enabled, and the value is set to 5 - we have an inherent logic contradiction: you wouldn't want to search "Close" in the _selected "Close"_.
(And that logic/implementation is still valid for 1024).

Limiting the valid range to 600-1024 might be a solution.

alankilborn: @Yaron10 

>But the latest commit creates two major issues

At this point, I just do what I'm told. :-)

Yaron10: > Limiting the valid range to 600-1024 might be a solution.

Or:

> My opinion is not bother with that. If user want to set wired value like 6, 7, 8 or 9, it's their freedom to get wired behaviour (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1739822133).

And the "freedom to get wired behaviour" should apply to _Selected "Close" not being set in "Find what"_ as well.

donho: @alankilborn @Yaron10 
The goal of this PR is to provide user the way to customize the length of auto-checking In selection and to disable the auto-checking In selection. 
Once user can do the above 2 things, they are on their own to set unreasonable value to have the inconsistent behaviour.
It's not critical for me such inconsistent behaviour because: 1. User will realize what is the reasonable value. 2. Even they don't, In selection can always be checked or unchecked. 3. In the worst case, they don't realize the reasonable value, and they don't check or uncheck conveniently, the search result is correct. The program is not buggy to me.
OTOH, change the behaviour of copying selected text into Find What due to this small enhancement is a big mistake to me - from the viewpoint of Notepad++ users.

donho (approved):

donho:
@alankilborn 
Thank you for you PR and your patience!
@Yaron10 
Thank you for your suggestion and your test!

alankilborn:
@donho We never got your comment on what "size" means.  See https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1743128860

donho:
@alankilborn 
> We never got your comment on what "size" means. See https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14175#issuecomment-1743128860

I see what you're trying to show.
By "character" it means western character, it's less accurate depending on which language, but for simplifying the conception of feature we use this term. 

Let's see users' reaction about that, then we'll see what we will do.

alankilborn:
>Let's see users' reaction about that, then we'll see what we will do.

Users probably won't react, they'll just _sigh_ and adjust.
IMO such sizes should always be in "characters", western or not.
When new features are added, at the very least, they should be done in characters.
That's why I originally used a call to SCI_COUNTCHARACTERS in the earliest commit in this PR (which you didn't like).
I do realize there are some corner cases in UTF-8 where a single "character" is hard to define.

It's becoming more of a "hot" topic, in Notepad++; here's a related user concern #14210

donho:
> That's why I originally used a call to SCI_COUNTCHARACTERS in the earliest commit in this PR (which you didn't like).
> I do realize there are some corner cases in UTF-8 where a single "character" is hard to define.

OK. Please do an PR to make character count for In selection.

Yaron10:
@donho,

> OK. Please do an PR to make character count for In selection.

👍 

***

Regarding the last commit here.

> due to this small enhancement

Small enhancement to some, and a major bug fix to others.

> **And the "freedom to get wired behaviour" should apply to Selected "Close" not being set in "Find what" as well.**

You 👍 that. :)

> If auto-checking is disabled (0), 1024 is irrelevant and even 1025-selected should be set in "Find what".
\- Why should if `((nppGui._inSelectionAutocheckThreshold == 0)` be removed?

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13677 (closed with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/591b00e538052543376e937a220aa6ef024ed10b) might be the first manifestation of that.

***

@alankilborn & @donho,

Thanks again for the hard work here. 👍

PR #14177, opened 2023-09-23
Store Unexecuted Search Terms In Search History

### LINKED ISSUE #14174: Recover search string after accessing search history
### Description of the Issue

I think it would be store the contents of the search line edit, in case someone accidentally presses the down key while writing a search term.

Tested in 8.5.2

### Current Behavior

1. Enter search text
2. Press down key (search text is replaced by history entry)
3. Press up (nothing changes, original search text is lost)

This can be pretty annoying if you try to write a complex regex term only to accidentally hit arrow up instead of arrow left so you have to start over.

### Expected Behavior

1. Enter search text (draft is stored somewhere)
2. Press arrow down
4. Press arrow up (draft is restored, only if going up from the first history element)

alankilborn:
Fix #14174.

@Cerno-b Can you [test](https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/Testing#using-a-pull-request-pr-build-version-of-notepad) this and see if it is what you expect?  Down arrow seems to work well, up arrow is a bit wonky (not sure if there is anything easy to do to improve it).

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -30,7 +30,7 @@ FindOption FindReplaceDlg::_options;
 
 #define SHIFTED 0x8000
 
-void addText2Combo(const TCHAR * txt2add, HWND hCombo)
+void addText2Combo(const TCHAR * txt2add, HWND hCombo, bool moveToTopIfFound = true)
 {
 	if (!hCombo) return;
 	if (!lstrcmp(txt2add, TEXT(""))) return;
@@ -38,6 +38,7 @@ void addText2Combo(const TCHAR * txt2add, HWND hCombo)
 	auto i = ::SendMessage(hCombo, CB_FINDSTRINGEXACT, static_cast<WPARAM>(-1), reinterpret_cast<LPARAM>(txt2add));
 	if (i != CB_ERR) // found
 	{
+		if (!moveToTopIfFound) return;
 		::SendMessage(hCombo, CB_DELETESTRING, i, 0);
 	}
 
@@ -4208,6 +4209,10 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwnd, UINT message, WPARAM
 	{
 		delLeftWordInEdit(hwnd);
 		return 0;
+	}
+	else if ((message == WM_KEYDOWN) && ((wParam == VK_DOWN) || (wParam == VK_UP)))
+	{
+		addText2Combo(getTextFromCombo(hwndCombo).c_str(), hwndCombo, false);
 	}
 	return CallWindowProc(originalComboEditProc, hwnd, message, wParam, lParam);
 }


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14174#issuecomment-1732589538

Cerno-b:
> Fix #14174.
> 
> @Cerno-b Can you [test](https://github.com/notepad-plus-plus/notepad-plus-plus/wiki/Testing#using-a-pull-request-pr-build-version-of-notepad) this and see if it is what you expect? Down arrow seems to work well, up arrow is a bit wonky (not sure if there is anything easy to do to improve it).

Wow that was fast, thanks a lot. Didn't have time over the weekend but I'll look into it as soon as I can.

alankilborn:
>Wow that was fast, thanks a lot. Didn't have time over the weekend but I'll look into it as soon as I can.

Don't get too excited; it appears fast-tracked for rejection.  :-)

xomx:
@alankilborn 
I tested this PR and it works ok.
So far I tend to think that it is a useful feature.
And one can always delete that search string from the combo-box when needed so.

alankilborn:
Thanks for the support @xomx

Cerno-b:
@alankilborn Ok, I checked it, but it's not quite what I had in mind.
The problem with this implementation is that every time I press up or down, the current edit is added to the top of my search, which will fill up the search and feels a bit clunky.

I think a better solution would be to add the draft to the top of the search history, but instead of adding a new item every time
the user presses down, re-use the top slot.

So basically this:

As we open the search dialog we have an existing history:

* one
* two

Now we start typing a new search phrase *new* and press down, the history now looks like this:

* new
* one
* two

Since we pressed down, the selected item is now *one*. Now we press up, and *new* is selected again

Now we change *new* to *new2* and press down again

Your implementation has a history that looks like this:

* new2
* new
* one
* two

I think it would be better to have a history that looks like this instead:

* new2
* one
* two

Cerno-b:
I am fairly new to Win32 programming so I had to read the code for a bit to see what options would be available with the way comboboxes work in this context.

Not sure this is desired behavior but wouldn't it be an option to always make the first history entry the current string of the line edit? So when the history is filled, the first entry would be filled with the empty string. When the user starts typing, the first history entry would be consistently be updated.

Now when the user presses down, we would have to skip the first entry so get the first actual entry of the history.
Also the first history entry would always be temporary until the search is actually executed.

Not sure if I am getting all the side-effects, I'm not very familiar with the codebase.

alankilborn:
My goal was a limited "fix" that would meet the goal of not losing data.  A bigger change doesn't interest me personally because, for one thing, lack of author interest has already been expressed, so it could be a time-waster.  If someone else wants to implement an alternate idea for a solution (that perhaps has a higher chance of adoption), please feel free.  For now, however, I'll leave this open.

Cerno-b:
@alankilborn Thank you very much for your effort on this.
I've been playing around with the code, but since I have almost no Win32 experience, this was a very interesting experience ;)

Might try my hand at a solution, if I can figure out how to make this work, but maybe you're right. If chances are low that this will be merged it might just end up as a practice exercise :)

Thanks again for your support!

donho:
The correct behaviour to me should be: 
1. When user types arrow up, nothing changes.
2. When user types arrow down, the sting is memorized temporarily. It will appear again when user types arrow up until list top - 1

The string in question is not in the list while it's not been searched anyway.

If the above criteria can be achieved easily, then it can be considered.

alankilborn:
>When user types arrow up, nothing changes.

Wouldn't an up-arrow press move the user _upward_ in the drop-down list (i.e., toward a smaller numeric index of the combobox)?  That's hardly "nothing changes".

donho:
@alankilborn 
> Wouldn't an up-arrow press move the user upward in the drop-down list (i.e., toward a smaller numeric index of the combobox)? That's hardly "nothing changes".

By "nothing changes", I mean the content should not be memorized temporarily.

There are 2 situations while up-arrow is pressed:
1. user is editing, then up-arrow is pressed by accident - nothing should be changed.
2. user is browsing the list, down-arrow has been pressed and content is already memorized temporarily - nothing should be changed.

alankilborn:
Closing because better PR for the issue was created by someone else:  PR 14187

PR #14183, opened 2023-09-25
Allow Window Windows command to be assigned a shortcut

### LINKED ISSUE #14179: add shortcut window/windows
please add shortcut window/windows

alankilborn:
Fix #14179 

Note:  Modernized boolean logic symbol (to `&&`) because I read in some other issue here that `and` is not to be used (but it was in existing code!)


alankilborn:
@P0rsche-911 

My goal for line 1233 was just to change `and` to `&&` because I noticed it and I know that the coding standard for Notepad++ specifically doesn't like `and` usage.

The whitespace is less important; I didn't notice it when I submitted.  If the author likes this PR generally and wants me to change that whitespace before acceptance, I'll certainly do that.

PR #14187, opened 2023-09-27
Store and restore search text on arrow key press

Cerno-b:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14174.

@alankilborn Since you were kind enough to attempt a fix, would you be interested in testing this version? I think it does what @donho commented in your PR.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -4170,6 +4170,8 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwnd, UINT message, WPARAM
 
 	bool isDropped = ::SendMessage(hwndCombo, CB_GETDROPPEDSTATE, 0, 0) != 0;
 
+	static TCHAR draftString[FINDREPLACE_MAXLENGTH] = { '\0' };
+
 	if (isDropped && (message == WM_KEYDOWN) && (wParam == VK_DELETE))
 	{
 		auto curSel = ::SendMessage(hwndCombo, CB_GETCURSEL, 0, 0);
@@ -4194,6 +4196,26 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwnd, UINT message, WPARAM
 	{
 		delLeftWordInEdit(hwnd);
 		return 0;
+	}
+	else if ((message == WM_KEYDOWN) && (wParam == VK_DOWN) && (::SendMessage(hwndCombo, CB_GETCURSEL, 0, 0) == CB_ERR))
+	{
+		// down key on unselected combobox item -> store current edit text as draft
+		::SendMessage(hwndCombo, WM_GETTEXT, FINDREPLACE_MAXLENGTH - 1, reinterpret_cast<LPARAM>(draftString));
+	}
+	else if ((message == WM_KEYDOWN) && (wParam == VK_UP) && (::SendMessage(hwndCombo, CB_GETCURSEL, 0, 0) == CB_ERR))
+	{
+		// up key on unselected combobox item -> no change but select current edit text
+		::SendMessage(hwndCombo, CB_SETEDITSEL, 0, MAKELPARAM(0, -1));
+		return 0;
+	}
+	else if ((message == WM_KEYDOWN) && (wParam == VK_UP) && (::SendMessage(hwndCombo, CB_GETCURSEL, 0, 0) == 0))
+	{
+		// up key on top selected combobox item -> restore draft to edit text
+		::SendMessage(hwndCombo, CB_SETCURSEL, WPARAM(-1), 0);
+		::SendMessage(hwndCombo, WM_SETTEXT, FINDREPLACE_MAXLENGTH - 1, reinterpret_cast<LPARAM>(draftString));
+		::SendMessage(hwndCombo, CB_SETEDITSEL, 0, MAKELPARAM(0, -1));
+		return 0;
+
 	}
 	return CallWindowProc(originalComboEditProc, hwnd, message, wParam, lParam);
 }


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4170,6 +4170,8 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwnd, UINT message, WPARAM
 
 	bool isDropped = ::SendMessage(hwndCombo, CB_GETDROPPEDSTATE, 0, 0) != 0;
 
+	static TCHAR draftString[FINDREPLACE_MAXLENGTH] = { '\0' };
+
```
alankilborn, line 4174: Since more than just find and replace comboboxes are subclassed with this proc, I'm not sure using FINDREPLACE_MAXLENGTH is the best choice here.  Do I know what would be better?  Nope. :-)
I suppose comparing hwndCombo to see if it is either the find box or the replace box would make it reasonable to then use FINDREPLACE_MAXLENGTH.

Cerno-b: Yeah, I took that from other parts of this code file like L50, L479, L5401, etc.

Cerno-b: But looking at the global sizes available, I think this is pretty much the only choice: 
image

alankilborn:
I went looking for weirdness because I thought that (from reading the code) a draft string from the _Find what_ combobox could be restored to the _Replace with_ combobox (or vice versa).  I could make this happen:

- Type some text into Fw
- Down-arrow to save as draft
- Type some text into Rw
- Down-arrow to save as draft
- Return to Fw
- Up-arrow to restore from draft
- The Rw draft is restored to Fw.

alankilborn:
Suggest more modern syntax on this:

`static TCHAR draftString[FINDREPLACE_MAXLENGTH] = { '\0' };`

as being:

`static wchar_t draftString[FINDREPLACE_MAXLENGTH]{};`

Cerno-b:
> Suggest more modern syntax on this:
> 
> `static TCHAR draftString[FINDREPLACE_MAXLENGTH] = { '\0' };`
> 
> as being:
> 
> `static wchar_t draftString[FINDREPLACE_MAXLENGTH]{};`

Will do, I've seen this in other places as well but according to coding rules I should only change my on bit of code, right?

Cerno-b:
> I went looking for weirdness because I thought that (from reading the code) a draft string from the _Find what_ combobox could be restored to the _Replace with_ combobox (or vice versa). I could make this happen:
> 
> * Type some text into Fw
> * Down-arrow to save as draft
> * Type some text into Rw
> * Down-arrow to save as draft
> * Return to Fw
> * Up-arrow to restore from draft; the Rw draft is restored to Fw
> 
> Note also that a similar thing could happen with the other comboboxes, e.g. _Directory_ in _Find in Files_ tab

Well spotted, thank you! I thought we would be safe because down arrow overwrites the draft but I forgot that the state of the other boxes is conserved and this can break the functionality. I'll check on it next week since I'm afk over the weekend.

alankilborn:
> I've seen this in other places as well but according to coding rules I should only change my on bit of code, right?

The `draftString` declaration line IS your line of code; as a new addition, it should be as "modern" as possible, I'd think.

alankilborn:
@Cerno-b It appears that your change was accepted and merged, along with the outstanding bugs that I pointed out, so you don't need to change anything.  While I like to think that someday I'll understand why something gets accepted versus rejected, I don't think I ever will.

Cerno-b:
> as

I agree. I was talking about all the other places that I copied from. It may make sense to modernize all of them but I think it may go against the coding rules to do it in the same PR.

Cerno-b:
> @Cerno-b It appears that your change was accepted and merged, along with the outstanding bugs that I pointed out, so you don't need to change anything. While I like to think that someday I'll understand why something gets accepted versus rejected, I don't think I ever will.

I was surprised as well. I think I'll just open another PR for the bugfix. Don't want to leave it in knowing that it exists now.

alankilborn:
>I was talking about all the other places that I copied from. It may make sense to modernize all of them but I think it may go against the coding rules to do it in the same PR.

That's called "refactoring" and the rules for that are to merely point it out in a issue, and let the author decide if HE wants to make changes like that (which, if he does, he will go ahead and do himself).

donho:
@alankilborn 


>    I went looking for weirdness because I thought that (from reading the code) a draft string from the Find what combobox could be restored to the Replace with combobox (or vice versa). I could make this happen:

> Type some text into Fw
> Down-arrow to save as draft
> Type some text into Rw
> Down-arrow to save as draft
> Return to Fw
> Up-arrow to restore from draft; the Rw draft is restored to Fw
>
>    Note also that a similar thing could happen with the other comboboxes, e.g. Directory in Find in Files tab

Well spotted! I didn't see your conversation while I was testing and merging the PR.
It's reverted in: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0d8a8dff7dc150db1ce4a8536994fa642e9969ef

Thank you for your heads up.

alankilborn:
BTW, I don't think this is really needed for _Directory_ and _Filters_ boxes, so logic should be added to exclude them.

Cerno-b:
> BTW, I don't think this is really needed for _Directory_ and _Filters_ boxes, so logic should be added to exclude them.

Was thinking about that as well but it thought it might make the code more complex and it's not really hurting anyone to have it in the other boxes. 

Still a bit weird to have it in all the boxes for the search dialogs but not globally. Personally I think this should be standard combo box behaviour across the board but since this would have to be manually added to all combo boxes of np++ it's probably overkill.

alankilborn:
>Was thinking about that as well but it thought it might make the code more complex and it's not really hurting anyone to have it in the other boxes.

True enough.

>Still a bit weird to have it in all the boxes for the search dialogs but not globally.

Which other edit/comboboxes are you thinking of?  I'm hard-pressed to think of any others quite like these in the Notepad++ UI.

Cerno-b:
> Which other edit/comboboxes are you thinking of? I'm hard-pressed to think of any others quite like these in the Notepad++ UI.

I was thinking in general, all comboboxes should have this feature imo since it's practical but does not hurt. 

However, if I understand this correctly, win32 widgets are not modeled as classes so we can't just subclass them and add this as a global feature.

As you said, it's not really needed anywhere else so it's not worth the effort to put it elsewhere.

donho:
> BTW, I don't think this is really needed for _Directory_ and _Filters_ boxes, so logic should be added to exclude them.

My thought about it is that if such behaviour is implemeted, it should be for every combobox of Notepad++.

alankilborn:
@Cerno-b said:

>I was thinking in general, all comboboxes should have this feature

@donho said:

My thought about it is that if such behaviour is implemeted, it should be for every combobox of Notepad++.

--------------

My point was that most comboboxes in Notepad++ are simple dropdowns; they don't have an in-line edit-an-entry feature.
Unless you can think of some others outside the _Find_ area that have this feature...

Cerno-b:
> My point was that most comboboxes in Notepad++ are simple dropdowns; they don't have an in-line edit-an-entry feature. Unless you can think of some others outside the _Find_ area that have this feature...

The only other place I could find was the *Run* window:

image

Cerno-b:
@alankilborn I created a new PR with the bugfix since this one was closed. https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14203

Is this the correct way to go about it? 

I also added your suggestion to modernize the string initialization.

PR #14193, opened 2023-09-28
Perl Stylers

### LINKED ISSUE #14192: Style Configurator / Themes: Perl styles need updating
### Description of the Issue

Scintilla/Lexilla's LexPerl has changed significantly since the `stylers.model.xml` list of styles for Perl was created/last-updated, so it is now out-of-date, and Notepad++ isn't settings colors for all the styles that are available in LexPerl.

- `stylers.model.xml` defines three styles (PREPROCESSOR, PUNCTUATION, LONGQUOTE) that aren't used by LexPerl.
- `stylers.model.xml` is missing 24 styles that are used by LexPerl.
- Some of the names for the styles should be updated for clarity.

I will be submitting a PR that includes the updates to the style list for both `stylers.model.xml` and `themes\DarkModeDefault.xml`, with colors that are reasonable for those default themes.

For the other themes that ship with Notepad++, I was thinking of setting `colorStyle="0"` for the new styles, so that they will inherit the foreground and background from their theme, to avoid defining colors that clash with the theme (it will mean that users who use those themes will not notice any changes in syntax highlighting, but if they go to Style Configurator, they will see more Styles available to Perl).  But if someone disagrees and thinks I should leave the other themes alone, I won't bother.

pryrt:
- update `stylers.model.xml` to have the 41 styles supported by LexPerl
- update `themes\DarkModeDefault.xml` to have the 41 styles supported by LexPerl

Why "Draft pull request"? We still need to determine if I should update the other themes or not.  I'm willing, but don't want to spend the effort if that portion isn't considered beneficial.

closes #14192


pryrt:
I decided to ping this PR, since @donho mentioned v8.5.8 release was coming soon, and I had been hoping that the theme updates (especially default and DefaultDarkMode) would be available in the next release, so that Style Configurator will make available all the styles that LexPerl uses.

PR #14202, opened 2023-10-02
Make session missing files remembered

### LINKED ISSUE #12079: Loading Current Session with missing files does not meet user expectation

### LINKED ISSUE #12744: Notepad++ doesnt show an error message when attempting to load sessions with missing file paths
When loading a session with missing file paths, notepad++ provides no error message so the user has no idea why the missing files are not being loaded with the rest of the session. Would it be possible to add an enhancement so that notepad++ displays an error message with the list of files that it cannot load with the session?

### Debug Information
Notepad++ v8.4.8 (64-bit)
Command Line : "C:\Users\quest\Documents\test session"
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.2364
Current ANSI codepage : 932
Plugins :
mimeTools (2.9)
NppConverter (4.5)
NppExport (0.4)
NPPJSONViewer (1.40)
NppToolBucket (1.10.6622.41516)

### LINKED ISSUE #13696: Restore opened file tabs for temporarly unavailable file paths
### Description of the Issue
When I forgot map network drive or mount crypted volume, or any else with logical hard drive letter is unavailable (E:\ F:\ to Z:\.) and run notepad++, then editor silently drops all opened files for that drive(s) without additional info, because paths are unavailable.

1. in command line emulate network drive or crypted volume
1a. **run cmd.exe**
1b. **mkdir C:\W**  (subdir for emulation **W:\** drive, of course if **W:\** is free on your system)
1c. **subst W: C:\W**  (at now you have drive W  . please do not close cmd.exe, it will be helpful later
2. open notepad++ , create new tab for text file, and write any text inside this tab
3. write this file for example as **W:\testfile.txt**
4. close notepad++ and run it again. you will see opened file **testfile.txt** , this is OK
5. close notepad++
6. click again cmd.exe window, and delete emulated drive by run **subst W: /D**
7. open notepad++ again.
8. this is all. you will not be noticed about lost path to **W:\testfile.txt** , and tab will not be open again.

### Expected Behavior
I need some configuration for this one, by selecting between:
1. **silent** do not open unavailable paths
2. **warning window**, with info about unavailable paths, short info similar to 'please map network or crypto drives W:, X: etc.... and or select below...', with two buttons 'Do not open listed files' , and second 'Try again'.
3. continuation to ad.2 , if still any paths are unavailable, do corrected list again described at point 2

### Actual Behavior
currently default is silently abort opening tabs for unavailable paths

### Debug Information
There is no debug. This behavior exist on all available versions notepad++

donho:
From #14168
~~Note that an option to activated this feature (remember all session files) has not been added yet.~~

Fix #12079, fix  #12744, fix  #13696


donho:
@Yaron10
I've just done the last effort to synchronize the execution of YesAll/NoAll dialog (message DOC_DELETED sent by PostMessage), the effort is in vain :(
Just like you, I prefer the solution YesAll/NoAll dialog. However it introduces a bug in some circumstance, I think I have no choice but opt for the solution of opening all with an alert to user.

You can always create an issue of feature request for YesAll/NoAll dialog, so someone more competent than I could implement this feature. 

Thank you for your understanding and for your participation of this issue.

Yaron10:
@donho,

Thank you for your hard work.

One last clarification:

> If the files were truly permanently gone (deleted intentionally), that behavior makes sense.

@pryrt in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12079#issue-1358915906.

**STR:**
Intentionally delete 50 files from your last session.
Open NPP.

**Result:**
NPP opens 50 place-holders _without asking the user_.

This a major bug IMO. Possibly even worse than the current (master) behavior.

\***

And one last question:
Would a "Yes/No" message be OK?
Something like:
```
Notepad++ can not open some files from your last Session.
Would you like to create place-holders for them? 
```

Ideally: open NPP, display the tab-bar (_and tabs_), prompt and run the place-holders mechanism from there.

***

A short code-snippet which can be added to the end of `Notepad_plus::loadSession()`.
(`sessionFileName` should be obtained there).
```cpp
	if (!allSessionFilesLoaded)
	{
		if (sessionFileName == nullptr)		// session.xml.
		{
			_nativeLangSpeaker.messageBox("AutoSessionFilesError",
				_pPublicInterface->getHSelf(),
				L"Notepad++ can not open some file(s) from your last Session.\n\nThe Session file \"session.xml\" will be modified on exit!\nYou can backup that file NOW.",
				L"Opening Session Files Error",
				MB_OK | MB_ICONWARNING);
		}
		else		// Manually-created session file (File -> Save Session).
		{
			_nativeLangSpeaker.messageBox("UserSessionFilesError",
				_pPublicInterface->getHSelf(),
				L"Notepad++ can not open some file(s) included in the following Session file:\n\n\"$STR_REPLACE$\"",
				L"Opening Session Files Error",
				MB_OK | MB_ICONWARNING,
				0,
				sessionFileName);
		}
	}
```

***

Anyway, the place-holders can be as useful in user-sessions as they are in session.xml.
The only difference is whether the session file should be auto-modified by NPP.

donho:
@Yaron10 
> NPP opens 50 place-holders without asking the user.
This a major bug IMO. Possibly even worse than the current (master) behavior.

It's not a bug for me. Users have the choice to not enable this feature, but if they enable it, they know what they are doing.

> And one last question:
Would a "Yes/No" message be OK?

Though I don't consider it (opening all place-holders if this option is enabled) as a bug as a bug, I prefer still that users can choose to open them or not when the option is enabled. So I did the last try last night, and it works quite good. The final solution as you suggested has been committed into this PR. For unknown reason, the CI doesn't work, you can check the PR and compile it locally if you want.

> Anyway, the place-holders can be as useful in user-sessions as they are in session.xml.
The only difference is whether the session file should be auto-modified by NPP.

All the work done in v8.5.8 for user created session is just for distinguishing user-session from session.xml. As a result they won't have the same behaviour. In the latest commit of the PR, the place-holder feature has been removed.

Yaron10:
@donho,

> I prefer still that users can choose to open them or not when the option is enabled. So I did the last try last night, and it works quite good.

Great! 👍 
I'd like to test it more thoroughly, but - so far - it seems to be working like a charm.
Thank you again for your work. I appreciate it.

> In the latest commit of the PR, the place-holder feature has been removed.

Could you please explain to me why the place-holders shouldn't work for user-sessions as well?
It might require some extra thinking and work, but isn't it worth while? Why not make it perfect?

***

https://ci.appveyor.com/project/donho/notepad-plus-plus/builds/48319984

donho:
@Yaron10
> I'd like to test it more thoroughly, but - so far - it seems to be working like a charm.

Nice to learn about that :)

> Could you please explain to me why the place-holders shouldn't work for user-sessions as well?
It might require some extra thinking and work, but isn't it worth while? Why not make it perfect?

The logic is simple and it came from you: user-session should not be modified by Notepad++ for users.
I realized that indeed it shouldn't because there's **Save session...** command for modify user-session.
Based on this rule, place-holder has no sense for user-session because it will create the confusion. If a user closes 1 or some place-holder(s), there could be 2 behaviours:
1. Modify user-session then in this case we violate the base rule and **Save session...** command become useless.
2. Not modify user-session then user won't understand why these place-holders are there.

Both brhaviours are not consist for users, so that's the reason to not have place-holders in user-session.

Yaron10:
@donho,

> place-holder has no sense for user-session

> The advantage with "place-holder" is, when user opens Notepad++, and realizes he/she forgets to mount a volume / insert a USB (because of the prompt), user can always insert the USB or mount the network driver, then switch to any place holder, the prompt of reloading message launch so he/she can get the files back without relaunching Notepad++. (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12079#issuecomment-1745599740).

This is valid for user-sessions as well, isn't it?

> because it will create the confusion

Why? I think it's quite consistent.
The user should know that NPP does not modify user-sessions. Period.
Keep the place-holders, close them. - NPP shouldn't care.

Yes, you'd need to use two different messages (one for session.xml and another for user-sessions), but isn't it worth it?

***

I haven't "learned" the code. Still, allow me to ask:
I see the advantage in handling the message(s) and result in `FileManager::newPlaceholderDocument()`.
But `FileManager::newPlaceholderDocument()` could be used (some day in the far future) for missing files in "Recently Closed Files".
Also - if you agree that place-holders are useful for user-sessions, it might be easier to handle the message(s) and result in `Notepad_plus::loadSession()`.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1481,8 +1481,12 @@ Please test those commands and, if needed, re-edit them.
 Alternatively, you can downgrade to Notepad++ v8.5.2 and restore your previous data.
 Notepad++ will backup your old &quot;shortcuts.xml&quot; and save it as &quot;shortcuts.xml.v8.5.2.backup&quot;.
 Renaming &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, your commands should be restored and work properly."/><!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
-			<NotEnoughRoom4Saving title="Save failed" message="Failed to save file.
-It seems there's not enough space on disk to save file. Your file is not saved."/>
+			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
+Are you sure you want to open them?"/>
+			<FileInaccessible title="File inaccessinble" message="Some files from your past session are inaccessible. They can be opened as empty and read-only files as placeholders.
```
xomx, line 1486: title="File inaccessible"

PowerEditor/installer/nativeLang/french.xml
```diff
@@ -1481,6 +1481,12 @@ Veuillez tester ces commandes et, si besoin, ré-éditer celles-ci.
 Autrement, vous pouvez revenir vers Notepad++ v8.5.2 et restaurer les données précédentes.
 Notepad++ va sauvegarder votre ancien &quot;shortcuts.xml&quot; et l'enregistrer sous &quot;shortcuts.xml.v8.5.2.backup&quot;.
 En renommant &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, vos commandes devraient être proprement restaurées."/><!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
+			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
+Are you sure you want to open them?"/>
+			<FileInaccessible title="Fichier inaccessinble" message="Certains fichiers de votre session précédente sont inaccessibles. Ils peuvent être ouverts comme des fichiers vides et en lecture seule en tant que espaces réservés.
```
xomx, line 1486: title="Fichier inaccessible"

PowerEditor/installer/nativeLang/english_customizable.xml
```diff
@@ -1493,7 +1494,12 @@ Notepad++ will backup your old &quot;shortcuts.xml&quot; and save it as &quot;sh
 Renaming &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, your commands should be restored and work properly."/><!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
 			<NotEnoughRoom4Saving title="Save failed" message="Failed to save file.
 It seems there's not enough space on disk to save file. Your file is not saved."/>
+			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
+Are you sure you want to open them?"/>
+			<FileInaccessible title="File inaccessinble" message="Some files from your past session are inaccessible. They can be opened as empty and read-only files as placeholders.
```
xomx, line 1499: title="File inaccessible"

xomx:
@donho @Yaron10 
So far everything works for me ok in my tests of this PR.
Thank you for your time & effort.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1347,6 +1357,31 @@ BufferID FileManager::newEmptyDocument()
...
 }
 
+BufferID FileManager::newPlaceholderDocument(const TCHAR* missingFilename, int whichOne)
+{
+	NppParameters& nppParamInst = NppParameters::getInstance();
+
+	static bool theWarningHasBeenGiven = false;
+	if (!theWarningHasBeenGiven)
+	{
+		int res = (nppParamInst.getNativeLangSpeaker())->messageBox(
```
Yaron10, line 1367: _I haven't actually tested it._
If res == IDNO - is there a reason to continue here and continue calling calling this function in `loadSession()`?

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1480,8 +1481,12 @@ Please test those commands and, if needed, re-edit them.
 Alternatively, you can downgrade to Notepad++ v8.5.2 and restore your previous data.
 Notepad++ will backup your old &quot;shortcuts.xml&quot; and save it as &quot;shortcuts.xml.v8.5.2.backup&quot;.
 Renaming &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, your commands should be restored and work properly."/><!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
-			<NotEnoughRoom4Saving title="Save failed" message="Failed to save file.
-It seems there's not enough space on disk to save file. Your file is not saved."/>
+			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
```
Yaron10, line 1484: Is the removal of `NotEnoughRoom4Saving` intentional

Yaron10:
@xomx,

Thank you, @donho & @molsonkiko.

Do you think that implementing place-holders for user-sessions might be confusing?
IMO, _not_ doing it would be confusing & unexpected.

Yaron10:
> IMO, not doing it would be confusing & unexpected.

It would actually make it much worse compared to current master.
The user sees the message for missing files in session.xml (-> "Great feature! NPP notifies me if files are missing.").
The user opens a user-session containing some missing files (-> "I haven't been notified. NO MISSING FILES.").

donho:
@Yaron10 
> I see the advantage in handling the message(s) and result in FileManager::newPlaceholderDocument().
But FileManager::newPlaceholderDocument() could be used (some day in the far future) for missing files in "Recently Closed Files".
Also - if you agree that place-holders are useful for user-sessions, it might be easier to handle the message(s) and result in Notepad_plus::loadSession().

Done in the latest commit.
Could you test it to be sure there's no new bug and either regression?

PowerEditor/installer/nativeLang/english.xml
donho: I'll redo another PR, so for the moment we can ignore it.

PowerEditor/src/ScintillaComponent/Buffer.cpp
donho: Well caught!
Fix it: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14202/commits/ccf6f75c63e002730f6a76aefc85cdd0daa4d198

Yaron10:
@donho,

Thank you for the recent changes and fixes.

I'm not sure which commit should be tested.
Also - it's currently "59 files changed", and I'd like to get some idea of this PR code-changes.

> I'll redo another PR, so for the moment we can ignore it.

Do you mean closing this PR and opening a new one?

PR #14203, opened 2023-10-02
Recover search string after accessing history (with bugfix)

Cerno-b:


Cerno-b:
This is a redo of PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14187 which was merged and then reverted due to a bug

Cerno-b:
Closed as duplicate. Use this one instead: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14204

PR #14204, opened 2023-10-02
Recover search string after accessing history with bugfix

Cerno-b:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14174.

This is a redo of PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14187 which was merged and then reverted due to a bug

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -4170,6 +4170,8 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwnd, UINT message, WPARAM
 
 	bool isDropped = ::SendMessage(hwndCombo, CB_GETDROPPEDSTATE, 0, 0) != 0;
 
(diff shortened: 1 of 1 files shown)


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4195,6 +4197,30 @@ LRESULT FAR PASCAL FindReplaceDlg::comboEditProc(HWND hwnd, UINT message, WPARAM
 		delLeftWordInEdit(hwnd);
 		return 0;
 	}
+	else if (message == WM_SETFOCUS)
+	{
+		draftString[0] = '\0';
+	}
```
donho, line 4203: It's tricky. I didn't find this solution 👍

Cerno-b: @donho Could you elaborate a bit what you didn't like about it?

I clear the draft string whenever this widget gets focus to ensure that the draft from another widget is not accidentally entered into the current one.

I still think that the feature has its merits, so if you could give me a hint what you don't like about this solution, I would like to find a better one.

Issue #12186, opened 2022-09-16
Backward search won't find match already at caret


alankilborn:
### Description of the Issue
Best to just see STR...

### Steps to Reproduce (STR) the Issue
1. In _Preferences_ > _Searching_, make sure _Fill Find Field with Selected Text_ is unchecked
2. Create a new Notepad++ tab
3. Copy this data WITHOUT a line-ending:

```
as;kjdf;anbperfmoai;jf;aiomnbo;egn;oweifjnp;gnsgas;dfmidkfljefk;ajfa;gbnad;ghjnapef;jafjnmrgag
```

4. Move to the Notepad++ tab and press Ctrl+v TWICE.
5. Your caret should be in column 189 and you should see only the number 1 in the line number margin (if you see the number 2, you haven't done steps 3 and 4 correctly).
6. Press Ctrl+f to bring up the _Find_ window.
7. Immediately press Ctrl+v to paste into _Find what_ to overwrite its selected contents.
8. Check the checkbox for _Backward direction_ and uncheck all other options (including _2 find buttons mode_).
9. Press the _Find Next_ button.

### Expected Behavior
Match to occur from column 95 to 189.

### Actual Behavior
Match occurred from column 1 to column 94.

### Debug Information
Notepad++ v8.4.5   (64-bit)
Build time : Sep  3 2022 - 04:05:32
Path : C:\\...\npp.8.4.5.portable.x64\notepad++.exe
Command Line : -titleAdd="8.4.5 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.1889
Current ANSI codepage : 1252
Plugins : 
    ComparePlus (1)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    PythonScript (2)

Yaron10:
@alankilborn,

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8661?

Issue #8661, opened 2020-08-03
Find Next vs. Find Prev


Yaron10:
**STR:**
Paste `AAAA` into a new document.
Select the first two letters.
Open the Find dialog, check `Wrap-Around` and press `Next`.
--The third and fourth letters are selected (correct).
Press `Prev`.
--The *second and third* letters are selected (?).

**Expected:**
The *first and second* letters should be selected.

`FindReplaceDlg::processFindNext` in `FindReplaceDlg.cpp`.
```
	if (pOptions->_whichDirection == DIR_UP)
	{
		//When searching upwards, start is the lower part, end the upper, for backwards search
		startPosition = cr.cpMax - 1;
		endPosition = 0;
	}
```
Shouldn't it be `startPosition = cr.cpMin;`?
***
Notepad++ v7.8.9   (32-bit)
Build time : Aug  1 2020 - 16:33:01
Path : C:\\---\notepad++.exe
Admin mode : ON
Local Conf mode : ON
OS Name : Windows 7 Enterprise (32-bit) 
OS Build : 7601.23964
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag_unicode.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

Uhf7:
I checked the behavior of Windows Notepad: Two steps forward and one step back. If you search "AA" within "AAAAAAAAAAAAAA", then it goes two characters forward with each forward search, and one character back with each backward search.

The behavior of MS Visual Studio: One step forward and one step back.

So at least the backward search seems to work same way as in Windows Notepad and in Visual Studio.

Yaron10:
@Uhf7,

Thank you for testing it.

It's indeed interesting.
In MS Visual Studio: if you have `AAAA`, it goes *two characters* back on Find Previous.
if you have `AAAAAAAAAAAAAA`, it goes *two characters* forward on the first Find Next and *one character* on the following Find Next.
Apparently, this is a bug. 

MS Notepad is at least consistent.

EditPad and Firefox (type it in the SearchBox of this page and open the Find Bar) go *two characters* on Next and Prev.
But more importantly: logically, there shouldn't be a difference between Next and Prev.

Uhf7:
MS VS seems to behave even more sophisticated. It jumps one character, if the search is repeated in the same direction. And it jumps two characters immediately after changing the search direction (when searching _AA_). I don't know, what this is good for. 

[...]

PR #14213, opened 2023-10-05
Fix Count when selection(s) active but not counting in-selection

### LINKED ISSUE #14200: Search Count inaccurate when a selection is active (and not counting in-selection)

Common setup; have only this text in a Notepad++ tab:

```
bbbbb bbbb bbbbbb
ccccc cccc cccccc
```

Steps To Reproduce (case 1):

* Put caret in the middle word (the one with 4 `b`)
* Ctrl+f to invoke _Find_; note `bbbb` gets put in the _Find what_ edit/combobox
* Type a single `b` so that it appears in _Find what_
* Uncheckmark all boxes in _Find_, select a _Search mode_ of _Normal_ (especially note that _In selection_ and _Wrap around_ are NOT to be checkmarked)
* Click on the N++ main window title bar (to activate the editor window so you can see the blinking caret); note that the caret is at the end of the second word
* Click on the _Count_ button in the _Find_ window and then observe the status bar of the _Find_ window
* **Expected**: `Count: 6 matches from caret to end-of-file`
* **Actual**: `Count: 10 matches from caret to end-of-file`

Steps To Reproduce (case 2):

* Select the middle word such that the blinking caret is before the first `b` in the word `bbbb`
* Ctrl+f to invoke _Find_; note `bbbb` gets put in the _Find what_ edit/combobox
* Type a single `b` so that it appears in _Find what_
* Uncheckmark all boxes in _Find_, select a _Search mode_ of _Normal_ (especially note that _In selection_ and _Wrap around_ are NOT to be checkmarked)
* Checkmark the _Backward direction_ checkbox
* Click on the N++ main window title bar (to activate the editor window so you can see the blinking caret); confirm that the caret is at the start of the second word
* Click on the _Count_ button in the _Find_ window and then observe the status bar of the _Find_ window
* **Expected**: `Count: 5 matches from start-of-file to caret`
* **Actual**: `Count: 9 matches from start-of-file to caret`

Steps To Reproduce (case 3):

* Make a column-block selection covering `bbbb` and `cccc` with the visible caret to the right/left of these words
* Try various counting operations of single `b` or single `c` characters in forward and/or backward directions
* Note various inaccurate counts...
* **Expected**: Accurate counts based off of the main-selection caret position
* **Actual**: Inaccurate counts

### Debug Information
Notepad++ v8.5.7   (64-bit)
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.5.7\virgin\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.2283
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

alankilborn:
Fix #14200 

Not sure if this will be wanted, but it was an easy enough fix so I went ahead with it.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2713,6 +2713,20 @@ int FindReplaceDlg::processAll(ProcessOperation op, const FindOption *opt, bool 
 		startPosition = cr.cpMin;
 		endPosition = cr.cpMax;
 	}
+	else if (op == ProcessCountAll && !pOptions->_isInSelection && !pOptions->_isWrapAround && !isEntire)
+	{
+		size_t caretPos = (*_ppEditView)->execute(SCI_GETCURRENTPOS);
+		if (direction == DIR_UP)
+		{
+			startPosition = 0;
+			endPosition = caretPos;
+		}
+		else
+		{
+			startPosition = caretPos;
+			endPosition = docLength;
+		}
+	}
 	else if (pOptions->_isWrapAround || isEntire)	//entire document needs to be scanned
 	{
 		startPosition = 0;


Yaron10:
@alankilborn,

👍 
Very nice!

Yaron10:
Let's please hold it until Sunday.
Thanks.

donho:
@Yaron10 
> Let's please hold it until Sunday.

Do you mean for discussion before the validation of this PR?

donho:
@alankilborn 
I don't approve this behaviour, for the following case:

image


The 8th "word" has been selected and Ctrl-F into Find What field.
With the same scenario you have provided, in current master: 
All unchecked then count: 5 matched.
Check Backward direction then count: 8 matched.

I consider it's correct behaviour. But this PR provided unexpected result, according how user selects the "word".
I would reject the PR, but let's wait @Yaron10  on Sunday for talking this behaviour.

alankilborn:
>All unchecked then count: 5 matched.
Check Backward direction then count: 8 matched.
I consider it's correct behaviour.

Any user could be confused by the results of your example "word" search.  Here's why:

* User checkmarks _Wrap around_ (only); runs _Count_; result: 12
* User unchecks _Wrap around_; checkmarks _Backward direction_; runs _Count_; result: 8
* User unchecks _Backward direction_; runs _Count_; result: 5

5 + 8 = 12 ??
User is uncertain -- are there 12 or 13 of the items searched for in this document?
This is even worse for the user when he cannot easily do the mental math, i.e., really large numbers of matches.

Results are supposed to be relative to the caret:

image

As the caret can only be placed between two characters, matches before the caret plus matches after the caret sum to total number of matches.

But...these are just my arguments... Do with it what you will.  Thank you for your consideration.  :-)

alankilborn:
I think user is going to be confused in this situation.
Maybe the best thing to do is to not even run the count if there is any selection (and _In selection_ is not checkmarked).
_Find_ window status bar could say: "Remove selection(s) before running Count" or some such...

Yaron10:
@donho & @alankilborn,

I apologize for the short "hold it" message. I hardly had a free moment.

@alankilborn was trying to lead us in the right direction. He wrote: **perhaps some of the the described behavior is "by design"** and **If Count is going to produce results in the form of "...from caret" or "...to caret"**.
Unfortunately I ignored that and got it all wrong. It was **my** fault.

```
word word word
```

1. On selecting the middle "word" and using Find Next/Prev, the caret position makes no difference.
I think we'd all agree this is correct and expected. 
\- Why should Count be different? 
The wording "from **caret**" is indeed confusing and inaccurate in our case. How about changing it to "from caret/selection to..."?
2. The behavior in Count, "Replace All" & "Mark All" should be consistent.
That is: if we take "from **caret**" literally, we should do so in the other two commands as well.

\***

@alankilborn's third case (column-selection) is surely a bug.
(Count, "Replace All" & "Mark All").

***
```
word word word
```

Uncheck "Wrap around".
Place the caret at the beginning of the third "word".
Find what: word.
Replace with: test.

**Replace All** (-> expected result: one replaced "from caret").
**Replace All in All Opened Documents**: (-> 3 replaced).

Though I can think of some lame argument to justify the different results, I think it should be "from caret/selection to..." in both cases.
And even if multiple documents are actually opened. 

***

We started discussing the following case in https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14200#issuecomment-1742190907.

```
word word word
```

Uncheck "Wrap around".
Place the caret at the beginning of the third "word".
Find what: word.
**Count** (-> 1 Match).
**FACD**  (-> 3 Matches).

IMO it's weird & unexpected.
@donho, what do you think?

If the search is executed in selection, we get "in 1 selection" in the Search Results window.
How about always adding the range & conditions to the Search Results?

For example:
"in selection; Case: ON; Whole word: OFF".
"from caret/selection **down**;  Case: OFF; Whole word: ON".

donho:
@Yaron10 
> I apologize for the short "hold it" message. I hardly had a free moment.

No need to apologize - we are all people who take our vacant time for contributing to this project :)

> Replace All (-> expected result: one replaced "from caret").
Replace All in All Opened Documents: (-> 3 replaced).

**Replace All in All Opened Documents** is not the same operation as **Replace All** at all.
The first considers the whole document as target, whereas the second takes into account the backward/forward argument.
So the different results is expected.

> Uncheck "Wrap around".
Place the caret at the beginning of the third "word".
Find what: word.
Count (-> 1 Match).
FACD (-> 3 Matches).
>
> IMO it's weird & unexpected.

It takes me few second to realize FACD means "Find All in Current Document" :)
The same argument as I used for the above reply:
**Find All in Current Document** is not the same operation as **Count** at all.
The first considers the whole document as target, whereas the second takes into account the backward/forward argument.
So the different results is expected.

> If the search is executed in selection, we get "in 1 selection" in the Search Results window.
How about always adding the range & conditions to the Search Results?
>
>For example:
"in selection; Case: ON; Whole word: OFF".
"from caret/selection down; Case: OFF; Whole word: ON".

Not mentioning there's no more room in Find dialog for adding more options, my position on User Interface is always "Keep It Simple, Stupid". I am, me too, an user quite simple & stupid and I do believe a complex UI will push users away (me firstly). So sorry, this option won't be added.

> @alankilborn was trying to lead us in the right direction. He wrote: perhaps some of the the described behavior is "by design" and If Count is going to produce results in the form of "...from caret" or "...to caret".

"right" direction is a big word, and it's subjective. I would say @alankilborn was trying to lead us in **his logic** direction. Note that I've never say his logic is wrong, but to me the example he has provided is not convincing: Very few people will search a letter in the document but the most of people search the whole word and in the case of whole word, they don't care where is the position of caret.

The goal is to build a user friendly tool to help people to get their job done, but it could be good additionally to make this tool as consistent as possible. Let's keep this PR opened. If there are similar requests coming, it will be considered.

alankilborn:
> @alankilborn's third case (column-selection) is surely a bug. (Count, "Replace All" & "Mark All").

I will break this out into a new issue dealing with only that aspect.

alankilborn:
> **Replace All** (-> expected result: one replaced "from caret"). **Replace All in All Opened Documents**: (-> 3 replaced).
> 
> Though I can think of some lame argument to justify the different results, I think it should be "from caret/selection to..." in both cases.

Hopefully the lame argument isn't going to be the one I'm about to say.  :-)

When a search affects multiple tabs/files (RAiAOD / FiF / RiF), that search is always the scope of the entire document.  Thus, _Wrap around_, _Backward direction_ and _In selection_ do not apply.  FiF and RiF are together on one tab where these checkboxes can be made invisible, but as RAiAOD shares space with single-file (and possibly scope-limited) replace actions, those checkboxes need to stay visible.

alankilborn:
>but to me the example he has provided is not convincing

>Very few people will search a letter in the document 

The example I used was just to illustrate; it wasn't supposed to be a "real world" or a "good use case" example.
We're supposed to provide a "small" example to show the problem, and I pretty much always try to reduce a real world problem into the smallest STR that I can.  I suppose I have regret with using the example I did if it was misconstrued.

alankilborn:
>Uncheck "Wrap around".
Count (-> 1 Match).
FACD (-> 3 Matches).

No inconsistency here to me.  FACD is immune to _Wrap around_ and _Backward direction_ state; it is always "entire document" unless overridden by _In selection_.

But a question is, can the user interface reasonably express this immunity??

Yaron10:
> No need to apologize...

👍 :)

> It takes me few second to realize FACD means "Find All in Current Document" :)

It took me a few seconds too when I first encountered it.
It might be easier for native English speakers. But for us - you're right, no more abbreviations please. :)

***

> @alankilborn's third case (column-selection) is surely a bug. (Count, "Replace All" & "Mark All").
>> I will break this out into a new issue dealing with only that aspect.

👍 
If you're thinking of a new PR as well, the current one can be closed, correct?
We can still use this thread for discussing the other points not directly related to that issue.

***
> If the search is executed in selection, we get "in 1 selection" in the Search Results window.
How about always adding the range & conditions to the Search Results?
>> Not mentioning there's no more room in Find dialog for adding more options...

Sure! But I was referring to the **Search Results window**.

תמונה

That could be a nice enhancement regardless of anything else discussed here. 

***
> The first considers the whole document as target, whereas the second takes into account the backward/forward argument.

> When a search affects multiple tabs/files (RAiAOD / FiF / RiF), that search is always the scope of the entire document.

_I know this is the current behavior._
Let's please put the GUI and the wording aside. (And also FiF (Find in Files) which is indeed a different case).
Couldn't "Find in Current Document" _taking  "Wrap around" into account_ be useful?
Wouldn't a new thinking-user be surprised when Count & "Find in Current Document" display different results _without changing any settings_?  

> The word "All" in FACD would seem to do it

I thought about this point too. :)

donho:
@Yaron10 
> Couldn't "Find in Current Document" taking "Wrap around" into account be useful?
Wouldn't a new thinking-user be surprised when Count & "Find in Current Document" display different results without changing any settings?

The name is rather "Find All in Current Document", so no. The actual behaviour will be kept, except a lot of requests for the changement.

donho:
The cause of race condition (2 dialogs launched even Yes/No to all) has been found finally.
Use `::SendMessage(hwnd, NPPM_INTERNAL_CHECKDOCSTATUS, 0, 0);` instead of [::PostMessage(hwnd, NPPM_INTERNAL_CHECKDOCSTATUS, 0, 0);](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppBigSwitch.cpp#L2023) fix 2 dialogs displayed problem (there are 2 dialogs launched simultaneously one after another if number of inaccessible file >= 2).
However, for the sake of performance, in `WM_ACTIVATEAPP` message we should use asynchronous function PostMessage and avoid SendMessage.
I don't have the solution so far. If anyone has any idea about this issue, please let me know.

alankilborn:
> The cause of race condition

@donho This seems unrelated to the current issue/PR?  Did you mean this comment in response to a different issue/PR?

alankilborn:
>Wouldn't a new thinking-user be surprised when Count & "Find in Current Document" display different results without changing any settings?

Again, the word "All" in FACD helps the user differentiate here.
If _Count_ was "Count All" instead, then there might be more of a problem.

---

But I think we have the definitive answer here anyway:

>The name is rather "Find All in Current Document", so no. The actual behaviour will be kept,

---

(Side note:  _Mark All_ is actually a problem; I'd lobby for the button being renamed to simply "Mark")

alankilborn:
@donho @Yaron10 

BUT...with all the discussions here, my original focused complaint stands, so I'll ask the question again and hope to get a definitive answer.

_Count_ provides results of the form:

* 20 matches from caret to end-of-file
* 16 matches from start-of-file to caret

If these statements of results are not correct, and indeed I've shown them not to be when there is a selection active, then the choices are:

* fix the logic so the statements are correct
* fix the statement of results so that it is correct with what the logic does
* do nothing

If the choice is "do nothing", let's close this; if the choice is something else, kindly advise what should be done.

I've been told that "user don't care about where is caret" so perhaps changing the statement of results to remove a reference to caret is the desired thing.

donho:
> @donho This seems unrelated to the current issue/PR? Did you mean this comment in response to a different issue/PR?

Oups! Sorry, deleted!
Another side effect of working late. Fortunately it's not sexting content :)

donho:
@alankilborn 

> 1.  fix the logic so the statements are correct
> 2.  fix the statement of results so that it is correct with what the logic does
> 3.  do nothing

> If the choice is "do nothing", let's close this; if the choice is something else, kindly advise what should be done.

The point 1 won't be possible to me, for the reason I've mentioned above.
I personally don't think the point 2 is necessary but why not, if we can find a term better and short (long description confuses always dumb people like me). 
If point 2 cannot be done, then definitely point 3 - do nothing.

Yaron10:
> BUT...with all the discussions here, my original focused complaint stands, so I'll ask the question again and hope to get a definitive answer.

> The wording "from caret" is indeed confusing and inaccurate in our case. How about changing it to "from caret/selection to..."?

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14213#issuecomment-1752036434.

***

@alankilborn,

> Again, the word "All" in FACD helps the user differentiate here.

1. The wording is secondary IMO. It's the concept & usefulness.
2. As you wrote: "but In selection makes that use of "All" less meaningful".

***

@donho,

> Fortunately it's not sexting content :)

We actually got some of that content in the past. :)

Yaron10:
@donho,

Before you get a heart attack: I was joking!

alankilborn:
After some further consideration, I've decided that changing the output to make more sense with what happens isn't all that workable due to the number of possible Count scenarios where all types of selections are concerned.  Thus, the best solution that will be accepted is to do nothing, so I'll close this PR and the associated issue.

Yaron10:
@alankilborn,

> due to the number of possible Count scenarios where all types of selections are concerned.

Could you elaborate please?

alankilborn:
>Could you elaborate please?

I was specifically thinking of multiple selections active.

Yaron10:
@alankilborn,

👍 
How about "from caret (or main selection)"?

alankilborn:
It doesn't particularly excite me, but I suppose I'll concede it does let the user know that something odd might be going on with the _Count_.  :-)

If Don likes it, we can go with it.

Yaron10:
> it does let the user know that something odd might be going on with the Count. :-)

And it might help _us_ too next year. :)
I'm pretty sure the issue has been discussed before.

alankilborn:
>I'm pretty sure the issue has been discussed before.

That's the problem when you get an unsatisfying non-solution to a problem.  The problem just resurfaces later.  :-)

Yaron10:
Indeed.
It might be time to fix https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12186 & https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8661. :)

alankilborn:
>It might be time to fix

Ah...you found some old discussions.  Nice.
But... these probably aren't perceived as problems by the powers-that-be, so...

alankilborn:
>Have the "powers-that-be" opposed the fix?

I see no record of that, but...
It gets a bit tiresome to put effort into debate and/or putting forth a fix, only to have nothing come from it.
So when you say "it might be time to fix...", the extent of my effort on it will be to say "I agree".

donho:
@alankilborn 
> It gets a bit tiresome to put effort into debate and/or putting forth a fix, only to have nothing come from it.
> So when you say "it might be time to fix...", the extent of my effort on it will be to say "I agree".

Sorry for what you feel...
I wish I could accept all the contribution and make everyone happy.
Unfortunately my role makes me apply what I believe in.

alankilborn:
@donho said:

>Sorry for what you feel...
I wish I could accept all the contribution and make everyone happy.
Unfortunately my role makes me apply what I believe in.

It's absolutely no problem -- everyone has to stand up for what they believe in.

A good number of my suggestions and my code have made it into Notepad++, and for that I am grateful and happy about it.

With my comments earlier, I was trying to explain to @Yaron10 why I wasn't going to pursue certain things any further.

alankilborn:
I created my own Counting implementation via script, which fixes the bugs I've noted here, and additionally provides even more capability that Notepad++'s native _Count_ feature, specifically multiple selection and column-block counting capability, and the ability to count _outside_ of selected text region(s).  This can be seen in its UI:

image

For further detail, see https://community.notepad-plus-plus.org/topic/25022

Issue #7574, opened 2019-11-14
Installer language is Afrikaans instead of English on Windows Server 2016.


rbarte:
### Description of the Issue

The Notepad++ installer instead of English is shown in Afrikaans. Afterwards also the notepad++ is shown in this language.
Language and regional settings on Windows 2016 OS are set to US (English).

Found similar closed bug: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3844
and 
https://community.notepad-plus-plus.org/topic/14564/dutch-as-install-language-instead-of-english

### Steps to Reproduce the Issue

1. Double click  on the "npp.7.8.1.Installer.x64.exe".
2. Confirm UAC.
3. Installer is shown in Afrikaans.
4. After installation the Notepad++ application is also in Afrikaans.

### Expected Behavior

Installer is shown in English (like other installers on US/English OS).
### Actual Behavior

Installer is shown in Afrikaans, also the installed application defaults to it.

### Debug Information

Windows Device Guard is active on this system.

![2019-11-14 10_21_25-NotepadPP](https://user-images.githubusercontent.com/31691084/68844140-9d91e980-06c9-11ea-92d8-bf7cbfdb73af.png)
![2019-11-14 10_30_37-Windows-Regional1](https://user-images.githubusercontent.com/31691084/68844567-4d675700-06ca-11ea-8f8f-6067fa57b4c4.png)
![2019-11-14 10_31_00-Windows-Regional2](https://user-images.githubusercontent.com/31691084/68844568-4d675700-06ca-11ea-88d2-b3f88619d530.png)
![2019-11-14 10_31_00-Windows-Regional3](https://user-images.githubusercontent.com/31691084/68844569-4dffed80-06ca-11ea-9869-582cce505ed9.png)
![2019-11-14 10_31_00-Windows-Regional4](https://user-images.githubusercontent.com/31691084/68844570-4dffed80-06ca-11ea-8384-d8043ab78840.png)
![2019-11-14 10_37_09-notepadpp-gui](https://user-images.githubusercontent.com/31691084/68844909-d41c3400-06ca-11ea-94df-65af7fb25009.png)

aaimaai:
That's not Dutch but Afrikaans. Because of its name (or its iso-code "af") it's probably the first language in the language list.

rbarte:
> 
> 
> That's not Dutch but Afrikaans. Because of its name (or its iso-code "af") it's probably the first language in the language list.

[...]

Issue #3844, opened 2017-10-24
7.5.1 Only installs as "Afrikaans"


rhisgen:
### Description of the Issue
Installing ver 7.5.1, 32 or 64 on MS Server 2012 R2 Standard. I get (Dutch) as the language. My servers are set to English. Previous versions all installed as English. 
Attached screenshot is first screen after launching installer.  It does not matter what login is used, or if you choose Install as Administrator.
Last version I have loaded was 7.4.2

### Steps to Reproduce the Issue

1. download 7.5.1 32 or 64bit
2. try to install on MS Server 2012 R2 Standard
3. 

### Expected Behavior
Install in english or prompt for language

### Actual Behavior
starts install in Dutch, 

### Debug Information

![notepad_dutch](https://user-images.githubusercontent.com/33067443/31957341-127abf80-b8b4-11e7-8514-b038ea6053a7.png)

nlsdg:
This screenshot isn't Dutch, but Afrikaans, correct?
Is there any language pack or setting related to Dutch or Afrikaans on the server or in the user settings?
I mean is this a random language (it is also alphabetically first!) or there anything on the server/user that could it Dutch/Afrtikaans?

rhisgen:
No other languages have been used on the server. It is set to english. Many
previous versions have been installed with no language issue.  Np other
program has the issue.   I only say Dutch .. because thats what google
translate thought it was.

Rich


On Nov 9, 2017 8:17 PM, "Niels de Groot" <notifications@github.com> wrote:

This screenshot isn't Dutch, but Afrikaans, correct?
Is there any language pack or setting related to Dutch or Afrikaans on the
server or in the user settings?
I mean is this a random language (it is also alphabetically first!) or
there anything on the server/user that could it Dutch/Afrtikaans?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3844#issuecomment-343354940>,
or mute the thread

rhisgen:
As of 7.5.6  still having install language issues.  Corrected I believe it is Afrikaans,  the language selection screen does not come up.   I cannot use the -L command line switch, because English is not in the list of language codes because it is supposed to be default... it is not.  This is Windows 2012 R2 Datacenter.  All language options on the servers are set to English.  I have not seen this issue on any other installed program.  This seems to have started with the 7.5 versions.   Any Ideas ?

[...]

PR #14125, opened 2023-09-09
Fix unsupported language set to english

mpheath:
Fix #3844
Fix #7574

PR #14215, opened 2023-10-06
Fix unintended Afrikaans installer language

xomx:
Fix (partially) #3844 #7574 

Some users (probably with some Windows installations like a server edition etc. or with some specific Windows locales like Icelandic) have reported a problem with unwanted Afrikaans Notepad++ installer language.

Partial-fix note:
Some users with such 'Afrikaans' problem also reported a complete omission of the installer language-selection dialog (the Welcome Screen was then shown directly in the Afrikaans language and the Notepad++ Registry lang-key had been set to the "Installer Language" nonsense value). For these users this PR fixes the Afrikaans language preselection (to English), but still the installer language selection dialog will not be shown at all. This currently looks like a NSIS-failure and it needs further investigation (and possibly a future NSIS upstream bug-report ...).

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -178,16 +178,28 @@ updaterDone:
 	Call SetRoughEstimation		; This is rough estimation of files present in function copyCommonFiles
 	InitPluginsDir			; Initializes the plug-ins dir ($PLUGINSDIR) if not already initialized.
 	Call checkCompatibility		; check unsupported OSes and CPUs
-		
+	
 	; look for previously selected language
 	ClearErrors
 	Var /GLOBAL tempLng
 	ReadRegStr $tempLng HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage'
 	${IfNot} ${Errors}
-		StrCpy $LANGUAGE "$tempLng" ; set default language
+		${If} "$tempLng" == "Installer Language" ; check for the possible previous inappropriate value (from some of the 'Afrikaans' issues...)
+			WriteRegStr HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage' '${LANG_ENGLISH}' ; bad, so repair that with our factory default
+			StrCpy $LANGUAGE "${LANG_ENGLISH}" ; and preset accordingly to the above
+		${Else}
+			StrCpy $LANGUAGE "$tempLng" ; ok, so offer that previously used language as the installer default
+		${EndIf}
 	${EndIf}
 	
 	!insertmacro MUI_LANGDLL_DISPLAY
+	
+	; check for the possible NSIS lang-engine fault
+	; - some users (maybe with some Windows server editions?) reported a complete omission of the installer language-selection dlg
+	;   (the Welcome Screen is then shown directly instead and the $LANGUAGE var is filled up with the "Installer Language" nonsense)
+	${If} "$LANGUAGE" == "Installer Language"
+		StrCpy $LANGUAGE "${LANG_ENGLISH}" ; a NSIS inherent failure(?!), at least repair that to our factory default
+	${EndIf}
 
 	; save selected language to registry
 	WriteRegStr HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage' '$Language'
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -184,23 +184,18 @@ updaterDone:
 	Var /GLOBAL tempLng
 	ReadRegStr $tempLng HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage'
 	${IfNot} ${Errors}
-		${If} "$tempLng" == "Installer Language" ; check for the possible previous inappropriate value (from some of the 'Afrikaans' issues...)
-			WriteRegStr HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage' '${LANG_ENGLISH}' ; bad, so repair that with our factory default
-			StrCpy $LANGUAGE "${LANG_ENGLISH}" ; and preset accordingly to the above
-		${Else}
-			StrCpy $LANGUAGE "$tempLng" ; ok, so offer that previously used language as the installer default
-		${EndIf}
+		StrCpy $LANGUAGE "$tempLng" ; set default language
 	${EndIf}
+	
+	Call unsupportedLanguageToEnglish
 	
 	!insertmacro MUI_LANGDLL_DISPLAY
 	
 	; check for the possible NSIS lang-engine fault
 	; - some users (maybe with some Windows server editions?) reported a complete omission of the installer language-selection dlg
 	;   (the Welcome Screen is then shown directly instead and the $LANGUAGE var is filled up with the "Installer Language" nonsense)
-	${If} "$LANGUAGE" == "Installer Language"
-		StrCpy $LANGUAGE "${LANG_ENGLISH}" ; a NSIS inherent failure(?!), at least repair that to our factory default
-	${EndIf}
-
+	Call unsupportedLanguageToEnglish
+	
 	; save selected language to registry
 	WriteRegStr HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage' '$Language'
 
--- a/PowerEditor/installer/nsisInclude/langs4Installer.nsh
+++ b/PowerEditor/installer/nsisInclude/langs4Installer.nsh
@@ -15,11 +15,12 @@
 ; along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 
-; Set languages (first is default language)
-;!insertmacro MUI_LANGUAGE "English"
+!define MUI_LANGDLL_ALWAYSSHOW
 !define MUI_LANGDLL_ALLLANGUAGES
 ;Languages
 
+  ; Set the Notepad++ static languages list
+  !insertmacro MUI_LANGUAGE "English" ; first is the default language (important for a possible language preselection failure...)
   !insertmacro MUI_LANGUAGE "Afrikaans"
   !insertmacro MUI_LANGUAGE "Albanian"
   !insertmacro MUI_LANGUAGE "Arabic"
@@ -35,7 +36,6 @@
   !insertmacro MUI_LANGUAGE "Czech"
   !insertmacro MUI_LANGUAGE "Danish"
   !insertmacro MUI_LANGUAGE "Dutch"
-  !insertmacro MUI_LANGUAGE "English"
   !insertmacro MUI_LANGUAGE "Farsi"
   !insertmacro MUI_LANGUAGE "Finnish"
   !insertmacro MUI_LANGUAGE "Estonian"
--- a/PowerEditor/installer/nsisInclude/langs4Installer.nsh
+++ b/PowerEditor/installer/nsisInclude/langs4Installer.nsh
@@ -143,3 +143,73 @@ LangString langFileName ${LANG_VIETNAMESE} "vietnamese.xml"
 LangString langFileName ${LANG_WELSH} "welsh.xml"
 LangString langFileName ${LANG_KURDISH} "kurdish.xml"
 LangString langFileName ${LANG_IRISH} "irish.xml"
+
+
+Function unsupportedLanguageToEnglish
+  ${Switch} $LANGUAGE
+    ${Case} ${LANG_AFRIKAANS}
+    ${Case} ${LANG_ALBANIAN}
+    ${Case} ${LANG_ARABIC}
+    ${Case} ${LANG_BASQUE}
+    ${Case} ${LANG_BELARUSIAN}
+    ${Case} ${LANG_BOSNIAN}
+    ${Case} ${LANG_BRETON}
+    ${Case} ${LANG_BULGARIAN}
+    ${Case} ${LANG_CATALAN}
+    ${Case} ${LANG_CORSICAN}
+    ${Case} ${LANG_CROATIAN}
+    ${Case} ${LANG_CZECH}
+    ${Case} ${LANG_DANISH}
+    ${Case} ${LANG_DUTCH}
+    ${Case} ${LANG_ENGLISH}
+    ${Case} ${LANG_ESTONIAN}
+    ${Case} ${LANG_FARSI}
+    ${Case} ${LANG_FINNISH}
+    ${Case} ${LANG_FRENCH}
+    ${Case} ${LANG_GALICIAN}
+    ${Case} ${LANG_GEORGIAN}
+    ${Case} ${LANG_GERMAN}
+    ${Case} ${LANG_GREEK}
+    ${Case} ${LANG_HEBREW}
+    ${Case} ${LANG_HINDI}
+    ${Case} ${LANG_HUNGARIAN}
+    ${Case} ${LANG_INDONESIAN}
+    ${Case} ${LANG_IRISH}
+    ${Case} ${LANG_ITALIAN}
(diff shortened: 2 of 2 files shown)


PowerEditor/installer/nsisInclude/langs4Installer.nsh
```diff
@@ -15,11 +15,12 @@
...
 
 
-; Set languages (first is default language)
-;!insertmacro MUI_LANGUAGE "English"
+!define MUI_LANGDLL_ALWAYSSHOW
 !define MUI_LANGDLL_ALLLANGUAGES
 ;Languages
 
+  ; Set the Notepad++ static languages list
+  !insertmacro MUI_LANGUAGE "English" ; first is the default language (important for a possible language preselection failure...)
```
donho, line 23: Have this code has been tested?
My binary which works for you is the following code:
```nsis
; Set languages (first is default language)
!insertmacro MUI_LANGUAGE "English"
!define MUI_LANGDLL_ALLLANGUAGES
;Languages

  !insertmacro MUI_LANGUAGE "Afrikaans"
  !insertmacro MUI_LANGUAGE "Albanian"
  !insertmacro MUI_LANGUAGE "Arabic"
  ;!insertmacro MUI_LANGUAGE "Armenian"
  !insertmacro MUI_LANGUAGE "Basque"
  !insertmacro MUI_LANGUAGE "Belarusian"
  !insertmacro MUI_LANGUAGE "Bosnian"
  !insertmacro MUI_LANGUAGE "Breton"
  !insertmacro MUI_LANGUAGE "Bulgarian"
  !insertmacro MUI_LANGUAGE "Catalan"
  !insertmacro MUI_LANGUAGE "Corsican"
  !insertmacro MUI_LANGUAGE "Croatian"
  !insertmacro MUI_LANGUAGE "Czech"
  !insertmacro MUI_LANGUAGE "Danish"
  !insertmacro MUI_LANGUAGE "Dutch"
  ;!insertmacro MUI_LANGUAGE "English"
  ...
```

PowerEditor/installer/nppSetup.nsi
```diff
@@ -178,16 +178,28 @@ updaterDone:
...
 	${EndIf}
 	
 	!insertmacro MUI_LANGDLL_DISPLAY
+	
+	; check for the possible NSIS lang-engine fault
+	; - some users (maybe with some Windows server editions?) reported a complete omission of the installer language-selection dlg
+	;   (the Welcome Screen is then shown directly instead and the $LANGUAGE var is filled up with the "Installer Language" nonsense)
+	${If} "$LANGUAGE" == "Installer Language"
+		StrCpy $LANGUAGE "${LANG_ENGLISH}" ; a NSIS inherent failure(?!), at least repair that to our factory default
+	${EndIf}
```
donho, line 202: I know the above 2 parts are the conclusion of test from you and @MalakymR.
However, I'm wondering if it's necessary.

As you said, current installer has 2 problems with iceland windows:
1. Language Selector appear, but detection failure so it's always Afrikaans which is the first choice.
2. Language Selector doesn't appear, the installer language is in Afrikaans during the installation.

Both problem should solve (partially, with acceptable solution) with my binary as I mention here:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14215/files#r1349304416

So with it, we could have 2 situation:
1. Language Selector appear, but detection failure so it's always English which is the first choice.
2. Language Selector doesn't appear, the installer language is in English during the installation.

Could you confirm the above description?

donho (changes requested):
Please confirm my hypotheses.
If confirmed, please remove 2 parts of change which is not necessary to me.

PowerEditor/installer/nsisInclude/langs4Installer.nsh
xomx: Yes, of course, many times by me (I use a Win11 VM with the Icelandic locale for that) and by @MalakymR (he can test on a computer with that more serious NSIS problem - skipping the lang-selection dlg + "Install Language" regkey...).

PowerEditor/installer/nppSetup.nsi
xomx: > Could you confirm the above description?

No, that is not correct.

> As you said, current installer has 2 problems with iceland windows:
> 
>     1. Language Selector appear, but detection failure so it's always Afrikaans which is the first choice.
> 
>     2. Language Selector doesn't appear, the installer language is in Afrikaans during the installation.

No. As I tried to explain [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3844#issuecomment-1748610142), we have 2 independent types of the 'Afrikaans' problem:

A) Lang-selection dlg is shown but currently with invalid Afrikaans
- it is caused either by a specific Windows locale like Icelandic or by a previous invalid N++ lang-regkey value
- this one is easy to fix by that "English-1st-in-our-static-lang-list"

B) Lang-selection dlg is not shown at all (directly the  "Welcome Screen" with invalid Afrikaans)
- this one is NOT caused by a specific Windows locale like Icelandic and I cannot simulate this type of the NSIS error myself
- this is what @MalakymR has on his computer with [English (United Kingdom) locale](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3844#issuecomment-1711349415)
- it is the worse (not fully fixable) one, here even a previous correct lang-regkey is ignored totally and $LANGUAGE installer variable is overwritten at runtime to the nonsense value "Installer Language" by NSIS itself (then this was consequently also written by us to the lang-regkey later because of we could not previously know that something like that is even possible...)

donho: What do you mean by "invalid Afrikaans" ?
Do you mean language detection failure so it's Afrikaans set as default?

donho: > A) Lang-selection dlg is shown but currently with invalid Afrikaans

So A has been fixed by setting English at the 1st position right?

donho: > B) Lang-selection dlg is not shown at all (directly the "Welcome Screen" with invalid Afrikaans)

B is not fixed completely, because  Lang-selection dlg is not shown at all, but "Welcome Screen" with English, thanks to the fix of setting English at the 1st position, right ?

xomx: > B is not fixed completely, because Lang-selection dlg is not shown at all, but "Welcome Screen" with English, thanks to the fix of setting English at the 1st position, right ?

Yes and no.
Even when that "Welcome Screen" is then in English due to that English 1st patch, NSIS still updates our $LANGUAGE var to "Installer Language" (this is what the [!insertmacro MUI_LANGDLL_DISPLAY](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/591b00e538052543376e937a220aa6ef024ed10b/PowerEditor/installer/nppSetup.nsi#L190) should do after a user selected some specific language, but this time it fails completely somehow).
So any further dependence on the $LANGUAGE in our NSIS script is affected and we then also write that "Installer Language" to N++ lang-regkey for the future use (and not the 1033 value for English).

donho: > Yes and no.

Not really clear to me - Yes for what and No for what?

donho: OK thank you for your confirmation @xomx.
So for me the fix should only contain the part of putting English on the first position.
The other registry part should not be included in this PR.

xomx: And what about the other possible `$LANGUAGE` dependencies like [this](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/591b00e538052543376e937a220aa6ef024ed10b/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh#L129)?

PowerEditor/installer/nppSetup.nsi
```diff
@@ -178,16 +178,28 @@ updaterDone:
...
 	Call checkCompatibility		; check unsupported OSes and CPUs
-		
+	
 	; look for previously selected language
 	ClearErrors
 	Var /GLOBAL tempLng
 	ReadRegStr $tempLng HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage'
 	${IfNot} ${Errors}
-		StrCpy $LANGUAGE "$tempLng" ; set default language
+		${If} "$tempLng" == "Installer Language" ; check for the possible previous inappropriate value (from some of the 'Afrikaans' issues...)
```
donho, line 187: PART2

So here could it be a function which detects whether if `$tempLang` is a legal value - all values other than legal values will fall down to  ${LANG_ENGLISH}?

donho (changes requested):
@xomx 
The PR could be accepted as it is. However, a better management of exception could be added.
Please let me know if you want to work on it. If not, I will merge this PR than add the exception management mentioning blow.

PowerEditor/installer/nppSetup.nsi
xomx: Yes.

I originally considered making my PR more universal here by taking advantage of the new `unsupportedLanguageToEnglish` function proposed by @mpheath (previous PR https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14125). But then I thought that you would not like it because it increases the size of the PR (IMO - in the wild, there are unlikely any other invalid values than the NSIS generated "Installer Language" one.)

> The PR could be accepted as it is. However, a better management of exception could be added.
Please let me know if you want to work on it.

In fact, I welcome a more universal solution. I will extend my solution with the above mentioned `unsupportedLanguageToEnglish` function.

xomx:
@donho 
Done. 
I tested what I can on both the x86 & x64 installers, all ok.

I would not want to annoy @MalakymR with demand for another test, but maybe it would be appropriate before the real commitment. What do you think?

PowerEditor/installer/nppSetup.nsi
```diff
@@ -187,8 +187,15 @@ updaterDone:
 		StrCpy $LANGUAGE "$tempLng" ; set default language
 	${EndIf}
 	
+	Call unsupportedLanguageToEnglish
```
donho, line 190: That's PR #14125's which didn't work - why should we apply it again?

PowerEditor/installer/nppSetup.nsi
```diff
@@ -187,8 +187,15 @@ updaterDone:
...
 	
+	Call unsupportedLanguageToEnglish
+	
 	!insertmacro MUI_LANGDLL_DISPLAY
-
+	
+	; check for the possible NSIS lang-engine fault
+	; - some users (maybe with some Windows server editions?) reported a complete omission of the installer language-selection dlg
+	;   (the Welcome Screen is then shown directly instead and the $LANGUAGE var is filled up with the "Installer Language" nonsense)
+	Call unsupportedLanguageToEnglish
```
donho, line 197: And why `unsupportedLanguageToEnglish` is called again here?

donho (changes requested):
Please provide the explanation of this commit.

PowerEditor/installer/nppSetup.nsi
xomx: Because otherwise the NSIS can left us here with $LANGUAGE=="Installer Language" and then after .onInit any [existing](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/591b00e538052543376e937a220aa6ef024ed10b/PowerEditor/installer/nsisInclude/mainSectionFuncs.nsh#L129) (or future) code dependent on the $LANGUAGE will have a problem...

donho: I see your point.
Why the first call then?

PowerEditor/installer/nppSetup.nsi
xomx: I have already [explained](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/3844#issuecomment-1736463273) this, I am copying here my previous note:

"@mpheath has already correctly described most of the problems and possible solutions. But his PR is not working because while he fixed the possibly wrong regkey/$LANGUAGE value before invoking that NSIS MUI_LANGDLL_DISPLAY at .onInit, if NSIS own lang-detection method of that MUI_LANGDLL_DISPLAY fails before displaying the lang-selection dlg, NSIS engine reaches rather for the statically predefined first member of the N++ lang-list than for the already correctly preset $LANGUAGE var. And then it updates the $LANGUAGE accordingly (currently either with that Afrikaans or with the complete nonsense value "Installer Language""

Now I am using his `unsupportedLanguageToEnglish` func because of [you have asked](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14215#discussion_r1349532697) for such more universal detection method...

PowerEditor/installer/nppSetup.nsi
xomx: Hmm, it's getting really complicated. Short answer is :

Now I think the 1st call is no longer necessary because of that other patch (English-1st-in-our-lang-list).
But there is also no harm in left it there.

Explanation:
First such call had to be there for any user with the NSIS lang-engine working without inherent failures, e.g.:
- user has Icelandic Windows locale
- on the installer start, before our .onInit had been called, the $LANGUAGE variable was preset by NSIS with its Icelandic equivalent value
- then this was offered as default for the language selection dlg combobox
- because we do not have NSIS-lang for that Icelandic, Afrikaans had been correctly (it was our 1st choice before) preselected instead in the lang-combobox by NSIS
- but now we switched to the English-1st, so that 1st call should not be needed anymore there (sorry, I did not really tested, no more time...)

donho: @xomx 
Here the current code in master:

```nsis
	; look for previously selected language
	ClearErrors
	Var /GLOBAL tempLng
	ReadRegStr $tempLng HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage'
	${IfNot} ${Errors}
		StrCpy $LANGUAGE "$tempLng" ; set default language
	${EndIf}
	
	!insertmacro MUI_LANGDLL_DISPLAY

	; save selected language to registry
	WriteRegStr HKLM "SOFTWARE\${APPNAME}" 'InstallerLanguage' '$Language'
```

The code does the following stuffs:
1. read a value from registry, fills $Language with the value.
2. then NSIS macro MUI_LANGDLL_DISPLAY uses $Language to pre-select a language for user. And $Language can be changed, If user changes the selection. 
3. write the value of $Language into registry for the next installation.


With only your fix **English-1st-in-our-lang-list**, in case of problem of read/write value from/into registry, $Language will be always the value we expected (1033). So it's the fix sufficient and there won't be the problem of copying nativeLang.xml file.

> But there is also no harm in left it there.

The PR #14125 of @mpheath should work (without the fix **English-1st-in-our-lang-list**) but for unknown reason it doesn't. So it's reasonable IMO to not add this code on which we know it doesn't work.

Please keep only **English-1st-in-our-lang-list** and remove other unnecessary code.

mpheath: @xomx Recognition of the issue as being valid has helped to propose a fix (or two if needed). I agree with a minimal fix as suggested by @donho and then evaluate again after a release to see if the issue has been fixed or needs more attention.

xomx: @donho 

> Please keep only English-1st-in-our-lang-list and remove other unnecessary code.

Done in #14224 .
Sorry for the delay (no time and I am on the GH web-interface only).

> With only your fix English-1st-in-our-lang-list, in case of problem of read/write value from/into registry, $Language will be always the value we expected (1033).

No. I obviously failed here as a explainer :-)

donho: @xomx 
> Done in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14224 .

Thank you!

> No. I obviously failed here as an explainer :-)

I may miss somethings...
But when I am checking the code, I don't see how $Language is not 1033 if we apply English-1st-in-our-lang-list fix.

PR #14223, opened 2023-10-09
Fix JSON5 to use JSON keywords

### LINKED ISSUE #14205: JSON5 highlights true and false values as errors
JSON5 highlights true and false values as errors

Select JSON5 in Language menu

### Expected Behavior
Correct highlighting of JSON5 content

### Actual Behavior
true and false values highlighted in red as errors

### Debug Information
Notepad++ v8.5.7   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.2283
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

image

mpheath:
Fix #14205

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -703,8 +703,7 @@ void ScintillaEditView::setEmbeddedJSLexer()
 
 void ScintillaEditView::setJsonLexer(bool isJson5)
 {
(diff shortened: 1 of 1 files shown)


mpheath:
No keywords are set for JSON5 so keywords are not recognized. This causes these keywords to be highlighted with the error style. Resolve is JSON5 to use JSON keywords to avoid duplication. Removed `j` as an uneeded variable to simplify the code.

PR #14232, opened 2023-10-14
Fix possible missing 'SessionFileInvalidError' message in MultiInst configs

### LINKED ISSUE #14228: "Session file is corrupted" message may not be displayed
**STR:**
Manually save a session via _File -> Save Session_.
Open that session file in another editor and delete the first `<NotepadPlus>` line (-> file is corrupted).
_Preferences -> Multi-Instance_: select one of the non-default options.

תמונה

Restart NPP.
Open some file.
Open the corrupted session file.

**Result:**
The "Session file is corrupted" message is not displayed.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/64e95cc61c69eaceb889584ced414f88f19b25f7/PowerEditor/src/NppIO.cpp#L2478

***

Notepad++ v8.5.7.

xomx:
Fix #14228 

Moving the error-dlg to the `NppParameters::loadSession` is more universal.

Other minor improvements (in `Notepad_plus::fileLoadSession`):
- checking also for a possible ShellExecute error
- removing the `isAllSuccessful` boolean, as it is no longer needed (due to previous patch by @donho )

Enhance also NPPM_GETNBSESSIONFILES API.

[New specs of NPPM_GETNBSESSIONFILES](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14228#issuecomment-1767258536)  (which is retro-compatible):
***********
**NPPM_GETNBSESSIONFILES**

Retrieves the number of files to load in the session sessionFileName. sessionFileName should be a full path name of an xml file.

**Parameters:**

wParam [out]
    BOOL* isValidXML, if this pointer is null, then this parameter will be ignored. TRUE if XML is valid, otherwise FALSE.

lParam [in]
    const TCHAR * sessionFileName

**Return value:**
Returns 0 if sessionFileName is an empty string/NULL, or XML session file is corrupted/invalid, else the number of files.
******************

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1297,16 +1297,16 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		{
 			size_t nbSessionFiles = 0;
 			const TCHAR* sessionFileName = reinterpret_cast<const TCHAR*>(lParam);
-			BOOL* pbOut = reinterpret_cast<BOOL*>(wParam);
-			if (pbOut)
-				*pbOut = false;
+			BOOL* pbIsValidXML = reinterpret_cast<BOOL*>(wParam);
+			if (pbIsValidXML)
+				*pbIsValidXML = false;
 			if (sessionFileName && (sessionFileName[0] != '\0'))
 			{
 				Session session2Load;
 				if (nppParam.loadSession(session2Load, sessionFileName, true))
 				{
-					if (pbOut)
-						*pbOut = true;
+					if (pbIsValidXML)
+						*pbIsValidXML = true;
 					nbSessionFiles = session2Load.nbMainFiles() + session2Load.nbSubFiles();
 				}
 			}
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1295,13 +1295,22 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_GETNBSESSIONFILES:
 		{
-			const TCHAR *sessionFileName = reinterpret_cast<const TCHAR *>(lParam);
-			if ((!sessionFileName) || (sessionFileName[0] == '\0'))
-				return 0;
-			Session session2Load;
-			if (nppParam.loadSession(session2Load, sessionFileName))
-				return session2Load.nbMainFiles() + session2Load.nbSubFiles();
-			return 0;
+			size_t nbSessionFiles = 0;
+			const TCHAR* sessionFileName = reinterpret_cast<const TCHAR*>(lParam);
+			BOOL* pbOut = reinterpret_cast<BOOL*>(wParam);
+			if (pbOut)
+				*pbOut = false;
+			if (sessionFileName && (sessionFileName[0] != '\0'))
+			{
+				Session session2Load;
+				if (nppParam.loadSession(session2Load, sessionFileName, true))
+				{
+					if (pbOut)
+						*pbOut = true;
+					nbSessionFiles = session2Load.nbMainFiles() + session2Load.nbSubFiles();
+				}
+			}
+			return nbSessionFiles;
 		}
 
 		case NPPM_GETSESSIONFILES:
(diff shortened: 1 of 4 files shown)


Yaron10:
@xomx,

👍 
Thank you for the nice fix and the fine-turning.

donho:
@xomx 
Thank you for the fix.
It was assigned to me for not forgetting to check it. However, I'm not sure that #14228 is a valid bug to me.
I will check it ASAP.

Yaron10:
> Thank you for the nice fix

The "Thank you" is valid and reinforced. :)
As for the fix: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14228#issuecomment-1763141359.
Removing the message altogether might be an option. - You don't prompt if session.xml or other XML files are corrupted.

donho:
@Yaron10 
> Removing the message altogether might be an option. - You don't prompt if session.xml or other XML files are corrupted.

Sorry if I misunderstand you (again), but I'm confused: Does the issue you created #14228 **"Session file is corrupted" message may not be displayed** ask to display a message if XML is corrupted?

donho:
@Yaron10 


OK, I see. you mean that in the case of plugin message NPPM_GETNBSESSIONFILES & NPPM_GETSESSIONFILES, the message box should not be displayed - that's a good point, and I agree with you.

Yaron10:
@donho,

The issue was evolving. :)

1. I did not (and still don't) suggest to display a message if any XML file used by NPP was corrupted.
2. As we all agree the message should not be displayed in `NPPM_GETNBSESSIONFILES` & `NPPM_GETSESSIONFILES`,
we have two options: modifying `nppParam.loadSession()`, or removing the message altogether.
As long as NPP do not prompt on corrupted session.xml (and others), I don't understand why corrupted user-session should be different (so IMO Removing the message altogether would be better).

donho:
@Yaron10 
> I did not (and still don't) suggest to display a message if any XML file used by NPP was corrupted.

So what the issue #14228 about is to remove the message box but not to display it?

> As long as NPP does not prompt on corrupted session.xml (and others), I don't understand why corrupted user-session should be different (so IMO removing the message altogether would be better).

The session.xml file is managed by the application - Notepad++ loads and saves it automatically, so users have no control over it. In contrast, user-created sessions are entirely in the hands of the users - they load and save them manually, and Notepad++ does not modify them. From this viewpoint, invalid session.xml files can be ignored by Notepad++ for the sake of user experience.

Speaking of user experience, session.xml files are loaded on startup of Notepad++. Therefore, the message box that appears on startup can break the user experience. On the other hand, user-created sessions are loaded manually by users who know what they are doing, so an error message box does not bother them.

Of course, my sense follows both user experience and consistency of tool. But when there’s a conflict between them, then for me the user experience should be the priority.

Yaron10:
@donho,

> So what the issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14228 about is to remove the message box but not to display it?

As I wrote earlier: the issue & solution were evolving.

> The session.xml file is managed by the application - Notepad++ loads and saves it automatically, so users have no control over it. In contrast...

When I started reading that paragraph, I thought your argument would be: NPP should over-write and "re-valid" session.xml on exit.

Anyway, I understand you prefer to keep the message.
So be it. :)

donho (changes requested):
Please 
1. solve the conflict
2. Remove the message from the 2 plugin messages
3. implement as following specs: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14228#issuecomment-1767258536

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1295,13 +1295,22 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
-			const TCHAR *sessionFileName = reinterpret_cast<const TCHAR *>(lParam);
-			if ((!sessionFileName) || (sessionFileName[0] == '\0'))
-				return 0;
-			Session session2Load;
-			if (nppParam.loadSession(session2Load, sessionFileName))
-				return session2Load.nbMainFiles() + session2Load.nbSubFiles();
-			return 0;
+			size_t nbSessionFiles = 0;
+			const TCHAR* sessionFileName = reinterpret_cast<const TCHAR*>(lParam);
+			BOOL* pbOut = reinterpret_cast<BOOL*>(wParam);
```
donho, line 1300: Please use isValidXML instead of pOut for being more explicit.
```cpp
BOOL* isValidXML = reinterpret_cast<BOOL*>(wParam);
```

donho (changes requested):
Please change the variable name for following the convention in this block (ie. "sessionFileName", but not "pIn") and improving the readability of code.

xomx:
@donho 
Sorry (now I re-read your comment) - did you really mean using `isValidXML` (instead of the current `pbIsValidXML`) for a pointer?

donho:
> @donho Sorry (now I re-read your comment) - did you really mean using `isValidXML` (instead of the current `pbIsValidXML`) for a pointer?

@xomx 
It's OK `pbIsValidXML` though it's not my coding style.
The most important is to make it meaningful so 10 years later when the maintainer checks the code he/she can understand what is the code doing without checking Notepad++ user-manual.

donho (approved):

Issue #14236, opened 2023-10-16
Application is getting upgraded,  while Notepad++.exe will not upgrade if in use


metaad:
While trying to upgrade the app from version 8.5.3 to 8.5.7 silently, below issue is noted.
Shortcut was launched of version 8.5.3 (Notepad++.exe was running).
Once the upgrade was triggered for 8.5.7, application's other files got upgraded but the Notepad++.exe retained the same and shows the version 8.5.3.
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/106653425/9a71e423-2c7a-406d-83b0-23233cb2c594)

The installer did not error out stating app is in use or should have been terminated without upgrade.

Could someone look into the issue.

Issue #10277, opened 2021-07-27
Trouble upgrading notepad++ using winget


scadden365:
### Description of the Issue

I want to use winget to upgrade notepad++. There is a problem, if notepad++ is open when the winget command is run, it doesn't upgrade, but winget reports that it has upgraded successfully. 
### Steps to Reproduce the Issue

1. open notepad++
2. run "winget upgrade notepad++" from windows terminal
Output
Found Notepad++ [Notepad++.Notepad++]
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licences to, third-party packages.
Downloading https://github.com/notepad-plus-plus/notepad-plus-plus/releases/download/v8.1.2/npp.8.1.2.Installer.x64.exe
Successfully verified installer hash
Starting package install...
Successfully installed

### Expected Behavior
For notepad++ installer to have alerted that the application was open, and to ask to close it before the installation continues. 
This is the functionality with WinSCP

### Actual Behavior
It appears to be installed, and running : "winget list notepad++"
Name Id Version
Notepad++ Notepad++.Notepad++ 8.1.2

Winget thinks the latest version is installed.
But when you check "?" "About Notepad++" , it still shows the old version v8.1.1

### Debug Information

See winget ticket: https://github.com/microsoft/winget-pkgs/issues/22514

OfficialEsco:
Again, this is false blaming. This has nothing to do with WinGet.
The issue is with the Silent Install, if you run the Notepad++ installer in interactive mode it will show this message
![image](https://user-images.githubusercontent.com/15158490/127215621-ec082ae0-a6ae-4d77-8c38-4a0d7adca04c.png)
If you run the Notepad++ installer in silent mode it will not display any message and exit with code 0, since it exists with code 0 it will "successfully" install Notepad++, however the program itself is not updated, but the registry strings are so it displays as it updated.

You can replicate this by having Notepad++ 8.1.1 installed and then having Notepad++ open while executing the 8.1.2 installer in silent mode.

The difference between the Notepad++ and WinSCP installer is that WinSCP is Inno while Notepad++ is NSIS.

mere-human:
Related: #8514

scadden365:
Related: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/10189

donho:
Notepad++ is not in Windows Store. The one on Windows store IS NOT Notepad++ official release.

[...]

Issue #8514, opened 2020-07-03
Feature request: silent update


PGNetHun:
### Feature request: silent update
Whenever there is a new update for Notepad++, it shows a popup window, and then the setup application starts with UI, in which we usually hit Next-Next-Finish... that's total waste of time.
It would be better to update Notepad++ silently, like Visual Studio Code does: it shows just a small bubble, that there is an update, you hit Update button, and it automatically updates application in the background, all you have to do is to restart VS Code, and that's it.

Notepad++ should update similarly:
- either update automatically in the background (if it is enabled in settings), and restart if done
- or when you close Notepad++, and there is an update (actual flow), then provide a button to update silently: no Next-Next-Finish UI, just update silently with default options.

gnodarse:
Just offering a perspective:

I like the way notepad++ updates. It makes updates more of an "event". The concrete benefit of that is that I am cognizant of every update that occurs I read the changes in every update. Automatic updates are good for applications that want to sneak in features that users don't want.

Blightbuster:
The problem for me is, that I dont want to install a update (which takes about a minute) before being able to quickly view or edit a text / config file. The way it currently updates makes me not update it for atleast a week until it gets to anoying to always have the pop up when im just trying to use np++ for 30 seconds.

JonathanLydall:
> Just offering a perspective:
> 
> I like the way notepad++ updates. It makes updates more of an "event". The concrete benefit of that is that I am cognizant of every update that occurs I read the changes in every update. Automatic updates are good for applications that want to sneak in features that users don't want.

Clearly it needs to be a user option because while you may not want software updating itself, I on the other hand find the current way that Notepad++ does updates very disruptive and distracting, it actively annoys me every single time that I have to do 10 clicks to update the app.

Filezilla is similarly annoying, but overall less so as updating it is a two click process, however, it is worse than Notepad++ in that the installer is left lying around in my downloads folder.

I would love to have the option for updates to happen automatically and silently in the background and when I re-open the application it shows the changelog.

aarner42:
[...]

PR #14251, opened 2023-10-20
fix silent installer mode when a previously installed Notepad++ is running

### LINKED ISSUE #10189: Silent Update does not update application if open
### Description of the Issue
When deploying a new version on top an existing one where notepad++ is opened, the notepad++.exe is not updated but the Add/Remove App receives the new version stamp.

1. Install Notepad++ 8.1 using /S
2. Open Notepad++ 8.1
3. Install Notepad++ 8.1.1 /S

### Expected Behavior
Its expected that upon installation open files are checked and the installation does not procceed in case the main application is open

or

The installer force closes the application and re-opens upon installing the new version.
### Actual Behavior
All files but notepad++.exe and nppshell_06.dll got overwritten with the latest version and the Add/Remove Programs is updated showing the version installed as 8.1.1

xomx:
Fix #10189 #10277 https://github.com/microsoft/winget-pkgs/issues/22514 #14236 partially #8514

In the future, this fix should be expanded to include also a new installer option (like "/closeRunningNpp"), which then ensures automatic termination of a previously installed & running Notepad++.

When Notepad++ installer with the '/S' optional param is used while a previously installed Notepad++ instance is running, it does not update the locked main 'notepad++.exe' binary, but the installer still reports success (exit code == 0, regkeys actualized according to the installer version used...).

Moreover - because the installation/upgrade then looks like as being done, some automatic update services like winget cannot be used until a next N++ version is released (winget then erroneously thinks that the current version is already installed...).

There is not such problem when we run that N++ installer in its interactive mode, as it shows the msg "Cannot continue the installation: Notepad++ is running. Please close Notepad++, then click 'Retry'".

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -150,6 +150,16 @@ Function .onInit
 	;
 	; --- PATCH END ---
 
+	; handle the possible Silent Mode (/S) & already running Notepad++ (without this an incorrect partial installation is possible)
+	IfSilent 0 notInSilentMode
+	System::Call 'kernel32::OpenMutex(i 0x100000, b 0, t "nppInstance") i .R0'
+	IntCmp $R0 0 nppNotRunning
+	System::Call 'kernel32::CloseHandle(i $R0)' ; a Notepad++ instance is running, tidy-up the opened mutex handle only
+	SetErrorLevel 5 ; set an exit code > 0 otherwise the installer returns 0 aka SUCCESS ('5' means here the future ERROR_ACCESS_DENIED when trying to overwrite the notepad++.exe file...)
+	Quit ; silent installation is silent, currently we cannot continue without a user interaction (TODO: a new "/closeRunningNppAutomatically" installer optional param...)
+nppNotRunning:
+notInSilentMode:
+
 	${GetParameters} $R0 
 	${GetOptions} $R0 "/noUpdater" $R1 ;case insensitive 
 	IfErrors withUpdater withoutUpdater


Yaron10:
@xomx,

Thank you for your work. 👍 
I've never used the installer. :)

xomx:
@Yaron10 
I also mostly use the portable N++. But sometimes I also move in an environment restricted by domain policies, limited user rights, etc. So I understand the request of sysadmins for a silent and unattended installation/upgrade.

Yaron10:
@xomx,

I just wanted to thank you for this PR and your other recent PRs.
I've learned it's better not to interfere in PRs related to features I don't normally use.
(That's why I added "I've never used the installer").

donho:
@xomx 
So the current fix is set error code when there's an instance of Notepad++ open, and installer in silent mode. Do you confirm?

PowerEditor/installer/nppSetup.nsi
```diff
@@ -150,6 +150,16 @@ Function .onInit
 	;
 	; --- PATCH END ---
 
+	; handle the possible Silent Mode (/S) & already running Notepad++ (without this an incorrect partial installation is possible)
+	IfSilent 0 notInSilentMode
+	System::Call 'kernel32::OpenMutex(i 0x100000, b 0, t "nppInstance") i .R0'
+	IntCmp $R0 0 nppNotRunning
+	System::Call 'kernel32::CloseHandle(i $R0)' ; a Notepad++ instance is running, tidy-up the opened mutex handle only
```
donho, line 157: In the case of `nppNotRunning`, is there no mutex handle ?

xomx: Yes. 
In such case the OpenMutex returns NULL (and possible GetLastError() then shows ERROR_FILE_NOT_FOUND).

xomx:
@donho 
The 3 open issues mentioned above can be also closed.

Issue #14161, opened 2023-09-20
Opening multiple files at once will cause previous tabs to disappear


OsakaWebbie:
I love the fact that NPP will remember unsaved files when I close and reopen it. And if I open one file (using double-click or right-click->Open or Open With in File Explorer, or any other method), NPP will launch and open the file in a fresh tab with my other in-progress stuff still in the other tabs. But just now I selected **two files** and chose "Open" from the context menu, and NPP launched with **only** those files - the other in-process stuff I was trusting NPP to save for me is gone!

### To replicate:
1. With one or more non-empty files still in tabs, close NPP. (It doesn't matter if the files are saved or unsaved.)
2. In File Explorer (or your favorite substitute - I use Directory Opus), select more than one file.
3. Right-click and select Open With (or just Open if the file type is associated with NPP). The previous files will no longer be in tabs, with no warning.

### Debug Info:
```
Notepad++ v8.5.6   (64-bit)
Build time : Aug 15 2023 - 15:29:28
Path : C:\PortableStuff\PortableApps\Notepad++Portable\App\Notepad++64\notepad++.exe
Command Line : "C:\Users\karen\Documents\Kizuna\Clients\DMI\Student_list.php"
Admin mode : OFF
Local Conf mode : ON
Cloud Config : C:\PortableStuff\PortableApps\Notepad++Portable\Data\Config
OS Name : Windows 11 Home (64-bit)
OS Version : 22H2
OS Build : 22621.2134
Current ANSI codepage : 932
Plugins : 
    ComparePlugin (2.0.2)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
```

xomx:
Ok, I found some problem, but I was able to add additional file-tabs to my existing N++ session.

This is my N++ session before opening two additional files from the Windows Explorer ('new 2' is an empty N++ tab):
![npp-opened_session](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/13075183/17a4ca23-e909-43ad-a636-804c45c8fc9b)

When I selected one from my two additional files in the Explorer, I can see & use the mentioned 'Open with' (and that file is then added to the N++ tabs ok):
![npp_open_one_additional_file_from_Explorer](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/13075183/75f1584b-fb24-43be-adc2-9ff523445839)

[...]

PR #14252, opened 2023-10-20
Make session inaccessible files remembered (part 2/2)

### LINKED ISSUE #12079: Loading Current Session with missing files does not meet user expectation

### LINKED ISSUE #12744: Notepad++ doesnt show an error message when attempting to load sessions with missing file paths

### LINKED ISSUE #13696: Restore opened file tabs for temporarly unavailable file paths

donho:
Based from #14168
Continue with #14202

Fix #12079, fix #12744, fix #13696


donho:
@Yaron10

Done, please check it.

Yaron10:
@donho,

👍 
Thanks again for your work and patience.

I'll continue testing it tomorrow.
So far:

***

**STR:**
Open file `A`.
File -> Save Session: save it as SessionA.ses.
Close NPP.
Delete file `A`.
Open SessionA.ses from File-Explorer.
Click "Yes" in the two messages.

**Result:**
Two `A` tabs in the same view. 

***

**STR:**
Create two user-sessions.
Delete some file(s) from _one session file_.
Select the two user-sessions in File_explorer and open them in one go.

**Result:**
You can not know which session file has missing files.

How about replacing `bool loadSession(Session & session, bool isSnapshotMode = false, bool isUserCreatedSession = false);` with `bool loadSession(Session & session, bool isSnapshotMode = false, const TCHAR *sessionFileName = nullptr);`?

\- You could then display the user-session file-name in the message.

***

"Open Session in New Instance (and save on exit)" complicates the implementation/messages.

***

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c8fafcd12b1f88c4817d7fee0a3f214d3772659c/PowerEditor/src/Notepad_plus.cpp#L6487

Is that still needed?

donho:
@Yaron10 
> STR:
Open file A.
File -> Save Session: save it as SessionA.ses.
Close NPP.
Delete file A.
Open SessionA.ses from File-Explorer.
Click "Yes" in the two messages.

> Result:
Two A tabs in the same view.

Fixed.

> "Open Session in New Instance (and save on exit)" complicates the implementation/messages.

Could you elaborate it?

> Is that still needed?

Simplified.

Please check the latest commit.

PowerEditor/installer/nativeLang/french.xml
```diff
@@ -1167,6 +1167,7 @@ Vous pouvez définir plusieurs marqueurs de colonne en utilisant un espace pour
 					<Item id="6821" name="secondes"/>
 					<Item id="6822" name="dans le dossier"/>
 					<Item id="6309" name="Mémoriser la session actuelle"/>
+					<Item id="6825" name="Mémoriser les fichiers inaccessibles de la session précédante"/>
```
Mateos81, line 1170: Précédente*

PowerEditor/installer/nativeLang/french.xml
```diff
@@ -1481,6 +1482,13 @@ Veuillez tester ces commandes et, si besoin, ré-éditer celles-ci.
 Autrement, vous pouvez revenir vers Notepad++ v8.5.2 et restaurer les données précédentes.
 Notepad++ va sauvegarder votre ancien &quot;shortcuts.xml&quot; et l'enregistrer sous &quot;shortcuts.xml.v8.5.2.backup&quot;.
 En renommant &quot;shortcuts.xml.v8.5.2.backup&quot; -&gt; &quot;shortcuts.xml&quot;, vos commandes devraient être proprement restaurées."/><!-- HowToReproduce: Close Notepad++, remove shortcuts.xml.v8.5.2.backup & v852ShortcutsCompatibilityWarning.xml if present, relaunch Notepad++, delete or modify a shortcuts via Shortcut Mapper, close Notepad++, then the message will show up -->
+			<FileInaccessibleUserSession title="Fichier inaccessible" message="Certains fichiers de votre session sauvegardée sont inaccessibles. Ils peuvent être ouverts comme des fichiers vides et en lecture seule en tant que espaces réservés.
```
Mateos81, line 1485: Que e -> qu'e (twice here, in 2 diffs)

Yaron10:
@donho,

> Fixed.

👍 

But this creates another issue.
**STR:**
Open file A.
Save session as SesA.ses.
Open file B.
Move B to the other view.
Save session as SesAB.ses.
Move B to the other view (-> A and B in one view).
Close NPP.
Delete A and B.
Select the two user-sessions in File-Explorer and open them in one go.

**Result:**
No B in MAIN_VIEW.

***

> How about replacing bool loadSession...

I understand you don't like it.

***

> "Open Session in New Instance (and save on exit)" complicates the implementation/messages.

I'm referring to the potential confusion you mentioned in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14202#issuecomment-1771799125.

The messages should indeed make it clear.

**User-sessions.**
```
Some files from your manually-saved session are inaccessible. They can be opened as empty and read-only place-holders.
Would you like to create those place-holders?

NOTE: Choosing not to create the place-holders or closing them later, your manually-saved session will NOT be modified on exit.
```

**session.xml.**
```
Some files from your past session are inaccessible. They can be opened as empty and read-only place-holders.
Would you like to create those place-holders?

NOTE: Choosing not to create the place-holders or closing them later, your session WILL BE MODIFIED ON EXIT!
You can backup your \"session.xml\" now.
```

"Open Session in New Instance (and save on exit)" **will indeed cause confusion**.
Regardless, I don't understand why you should make an exception in this case and modify the user-session.

Yaron10:
> But this creates another issue.
STR:

You get the same result without deleting the files. - Probably not related to this PR.
It should be further investigated with current master **and older NPP versions**.

donho:
> But this creates another issue. 
**STR:** Open file A. Save session as SesA.ses. Open file B. Move B to the other view. Save session as SesAB.ses. Move B to the other view (-> A and B in one view). Close NPP. Delete A and B. Select the two user-sessions in File-Explorer and open them in one go.
> 
> **Result:** No B in MAIN_VIEW.

Cannot reproduce it:

image

>> How about replacing bool loadSession...
> I understand you don't like it.

Implementing...

donho:
@Yaron10 
> The messages should indeed make it clear.

Done. However, according google, "place-holder" should be "placeholder", so I maintain "placeholder".

> "Open Session in New Instance (and save on exit)" will indeed cause confusion.
Regardless, I don't understand why you should make an exception in this case and modify the user-session.

"Open Session in New Instance (and save on exit)" was a frequent-request feature which was implemented quite long time ago, so I believe a lot of users depend on this feature.

You can check the latest commit in which the messages have been refined (with the user session file name).

Yaron10:
@donho,

It seems to work like  charm. 👍 
I hope we haven't missed some corner cases.

Thanks again.

> Done. However, according google, "place-holder" should be "placeholder", so I maintain "placeholder".

👍 
We need a native English speaker.

```
They can be opened as empty and read-only placeholder files.
```
Instead of
```
They can be opened as empty and read-only files as placeholders.
```
?

And `\n` before `You can backup`?

***

> "Open Session in New Instance (and save on exit)" was a frequent-request feature which was implemented quite long time ago,

Do you mean the **(and save on exit)** part was a frequent-request feature?

***

> You get the same result without deleting the files. - Probably not related to this PR.
It should be further investigated with current master **and older NPP versions**.

Could you please try the _exact steps_?

**STR:**
Open file A.
Save session as SesA.ses.
Open file B.
Move B to the other view.
Save session as SesAB.ses.
Move B to the other view (**-> A and B in one view on exit**).
Close NPP.
Select the two user-sessions in File-Explorer and open them in one go.

**Result:**
No B in MAIN_VIEW.

Yaron10:
@alankilborn,

Could you please test this PR?
Thank you.

pryrt:
Regarding "place-holder" vs "placeholder", I definitely agree that "placeholder" is the right choice.

@alankilborn suggested I chime in, since I was the one who submitted #12079.  Based on my experiments, this meets my expectations for reasonable behavior regarding the session when a file/directory/mounted-drive goes temporarily missing, and I would be happy if this is the version that goes "live".

Thank you for all the effort that the whole team put into going through various edge cases, and all the back-and-forth that occurred.  The end result shows the benefit of getting a lot of different perspectives on this issue.

Yaron10:
@pryrt,

Thank you for submitting the issue and testing this PR.

@alankilborn,

That was a nice dodge. :)

donho:
@Yaron10 
I do believe "They can be opened as empty and read-only files **as** placeholders" since it's chatgpt has suggested me :)
However, I think to use "documents" instead of "files" is even better: "They can be opened as empty and read-only **documents** as placeholders".

> Do you mean the (and save on exit) part was a frequent-request feature?

Yes.

> Could you please try the exact steps?

Cannot always reproduce it.

image

Note that I copied new compilerf bin to replace `Path : C:\Program Files\Notepad++\notepad++.exe` and I used `Edit with Notepad++` to open both session A.sess & AB.sess.

Yaron10:
@donho,

> I do believe "They can be opened as...

👍 

***

> and I used Edit with Notepad++ to open both session A.sess & AB.sess.

I'm using NPP portable, and I don't have that entry in the context menu.

Both files are selected, but does it make a difference if you right-click on A.sess or AB.sess?
_And if you use Open?_

I'm using [QTTabBar](http://qttabbar.wikidot.com/) and [OldNewExplorer](https://www.softpedia.com/get/Tweak/System-Tweak/OldNewExplorer.shtml) on Windows 10.

On using Open in the bar
תמונה

I always get a wrong result: No B in MAIN_VIEW.

On using Open in the context menu, it depends on the file I right-click.

_https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14161 might be related._

donho:
@Yaron10 
Thank you for the more detail info

> Both files are selected, but does it make a difference if you right-click on A.sess or AB.sess?

I suppose by "right-click" you means **Edit with Notepad++** of context menu under Explorer which comes Windows 10/11. 
When you right-click on A.sess or AB.sess and run **Edit with Notepad++** from the context menu, NppShell.dll runs command `notepad++.exe "A.sess" "AB.sess"` - both session files are opened in one call, and session files are opened one after another.

In this scenario, it works fin, as expected.

> And if you use Open?

Always under Explorer which comes Windows 10/11, when you select 2 files and use Open command from the context menu, or type ENTER key to open 2 files:  the system run `notepad++.exe "A.sess"` then `notepad++.exe "AB.sess"` (or in inverse order, it depends in which order you select the 2 files).

In this scenario, I can reproduce the problem, not exactly the same result though:

image

I will see what I can do about it.

donho:
> I will see what I can do about it.

The test results (with **open**) are different on mono-instance - multi-instance works perfectly - and it's surely due to the race condition: System launches several Notepad++ commands simultaneously and all commands lead to one instance for the execution in a random order.
Unfortunately I've no solution so far. And at the same time, this issue is related to this feature but it's not a regression due to this PR, so I will merge this PR.

Yaron10:
@donho,


> In this scenario, I can reproduce the problem, not exactly the same result though (but even worse):

Yes, I encountered a similar case after posting my STR.

> I will see what I can do about it.

Thank you for looking into it and good luck.
Could you pleas open an issue? - Your specs are (obviously) much more common.

> And at the same time, this issue is related to this feature but it's not a regression due to this PR, so I will merge this PR.

👍

donho:
@Yaron10 
> Could you please open an issue? - Your specs are (obviously) much more common.

It's indeed the same issue of #14161: 
I tested with 1 modified untitled text file in the session. With no Notepad++ open, I select 4 text files and launch **Open** via Explorer.
After several tests I cannot reproduce the lost of the modified untitled document in the current session as @OsakaWebbie has experienced. However, I have reproduced once (among 6-8 times) that 1 of these 4 files was missing - it's what it has happened to you with session files.
As I have explain [here ](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14252#issuecomment-1774152943) & [there](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14161#issuecomment-1774128291), system launches several commands to call Notepad++, that will raise the race condition (probably it's [better explained by portableapps](https://portableapps.com/comment/254940#comment-254940)).

As a result, I won't open a similar issue as #14161, and this issue won't be fixed and probably can't be fixed.
The best practice is using the "Edit with Notepad++" of context menu or a one command with all files as arguments in command line.

Yaron10:
@donho,

First, I'd like to emphasize again that I was initially wrong: the issue is not related to this PR, and can be reproduced in v8.5.7 as well.

> In this scenario, I can reproduce the problem, not exactly the same result though (but even worse)

I understand your explanation and recommendation.
But ending up with two identical-path tabs in the same view is a major problem, isn't it?
(And again, it can happen with **accessible files** as well).

The question is if NPP should (and could) make sure this is not the end result.
If no other solution is viable - how about checking if an identical-path already exists in the view, and prevent opening another one?

Yaron10:
Alternatively, and as it's a rare case, it can be labeled as a "known issue".
The best applications have them. :)

donho:
> If no other solution is viable - how about checking if an identical-path already exists in the view, and prevent opening another one?

It's already done in the code.
The problem of race condition, that is the same resource is accessed at the same time, and identical-path checking for 2 threads or 2 processes have the same result. That's the reason that the identical-path checking not working in my test.

> Alternatively, and as it's a rare case, it can be labeled as a "known issue".

Why not? I can always labeled #14161 as "known issue". But what's the advantage?

Yaron10:

> I can always labeled https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14161 as "known issue".

That issue is only reproducible with PortableApps.
The cases we've encountered her are valid in common regular setups. 

> But what's the advantage?

So that we at least have an open issue and not just some comments in a closed PR.
As you understand. :)

PR #14268, opened 2023-10-25
Hotkey for "Copy Selected Line(s)" in find space.

GOLEM777:
Ctrl+Shift+X hotkey for Copy Selected Line(s) in find field.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -5115,6 +5115,7 @@ intptr_t CALLBACK Finder::run_dlgProc(UINT message, WPARAM wParam, LPARAM lParam
 				generic_string collapseAll = pNativeSpeaker->getLocalizedStrFromID("finder-collapse-all", TEXT("Fold all"));
 				generic_string uncollapseAll = pNativeSpeaker->getLocalizedStrFromID("finder-uncollapse-all", TEXT("Unfold all"));
 				generic_string copyLines = pNativeSpeaker->getLocalizedStrFromID("finder-copy", TEXT("Copy Selected Line(s)"));
+				copyLines += TEXT("\tCtrl+Shift+X");
 				generic_string copyVerbatim = pNativeSpeaker->getLocalizedStrFromID("finder-copy-verbatim", TEXT("Copy"));
 				copyVerbatim += TEXT("\tCtrl+C");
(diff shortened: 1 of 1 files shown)


GOLEM777:
Macro doesn't record operation///
```
		case NPPM_INTERNAL_SCINTILLAFINDERCOPY :
				{
					copy();
					return TRUE;
				}
```
replacement  on SCI_LINECOPY  won't help for hot key?
you can help?

alankilborn:
>you can help?

I can try, if @donho approves and has an idea for how it should work.

donho:
@alankilborn 

SCI_LINECOPY is already in Shortcut Mapper.

But I wonder [why Finder::copy()](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L4838) shouldn't use SCI_LINECOPY instead of current more complex code:
```cpp
void Finder::copy()
{
	_scintView.execute(SCI_LINECOPY );
}
```
 Could you check it please?

GOLEM777:
SCI_GETSELECTIONS  - tied selection by column
Therefore the field ignores the column selection.
**The problem is copying associated garbage!**
I think this is a bad idea. I was stupid!
------
I don't know how to compile and can't test it.  )))
But  judging by the code, it will be so.

alankilborn:
>But I wonder why Finder::copy() shouldn't use SCI_LINECOPY instead of current more complex code

This is the _Copy Selected Line(s)_ choice from the _Search results_ right click menu.  It does a very special type of copy (thus needing the complex code) as detailed in the [user manual](https://npp-user-manual.org/docs/searching/#rightclick-commands-in-the-client-area-of-a-search-results-window-s-tab:~:text=What%20gets%20copied%20when%20RightClick%20%3E%20Copy%20Selected%20Line(s)%20is%20run).

Code is correct as is.

donho:
@alankilborn 
> This is the Copy Selected Line(s) choice from the Search results right click menu. It does a very special type of copy (thus needing the complex code) as detailed in the [user manual](https://npp-user-manual.org/docs/searching/#rightclick-commands-in-the-client-area-of-a-search-results-window-s-tab:~:text=What%20gets%20copied%20when%20RightClick%20%3E%20Copy%20Selected%20Line(s)%20is%20run).

Hmm... indeed.

In this case, I see no trivial way to attribute Ctrl+Shift+X to this command.

kor44:
Is it possible add option to redefine shortcut `Ctrl+C` not for `Copy` but for `Copy Selected Line` in `Search result Window`?

PR #14278, opened 2023-10-27
Fix bugs for remembering inaccessible files of past session (placeholder)

### LINKED ISSUE #14271: Placeholders are not marked as read-only if "Session Snapshot" is disabled
**STR:**
Use the binary from https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e5ee4c3638a4b7d96504de6beba088016f1922f3.
Enable "Remember current session" & "Remember inaccessible files".
_Disable "Session Snapshot"._
Open file A.
Close NPP.
Delete file A.
Open NPP.
Choose "Yes" to create the placeholder.

**Result:**
File A is not marked as read-only.

NOTE: activating another program and returning to NPP - it IS marked as read-only.

### LINKED ISSUE #14272: Placeholders: missing "File does not exist" message
**STR:**
Use the binary from https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e5ee4c3638a4b7d96504de6beba088016f1922f3.
Enable "Remember current session" & "Remember inaccessible files".
Open file A.
Close NPP.
Delete file A.
Open NPP.
Choose "Yes" to create the placeholder.
Restore file A.
Reload on prompt (-> _file A should be a normal file now_).
Delete file A.
Activate NPP.

**Result:**
No "File does not exist" message.

### LINKED ISSUE #14273: Placeholders: "Move to New Instance" should be disabled
**STR:**
Use the binary from https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e5ee4c3638a4b7d96504de6beba088016f1922f3.
Enable "Remember current session" & "Remember inaccessible files".
Open file A.
Close NPP.
Delete file A.
Open NPP.
Choose "Yes" to create the placeholder.

**Result:**
The commands "Move to New Instance" & "Open in New Instance" _are enabled_.

donho:
Fix #14271, fix #14272, fix #14273

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2606,7 +2606,6 @@ void Notepad_plus::checkDocState()
 	::CheckMenuItem(_mainMenuHandle, IDM_EDIT_SETREADONLY, MF_BYCOMMAND | (isUserReadOnly ? MF_CHECKED : MF_UNCHECKED));
 
 	enableCommand(IDM_FILE_DELETE, isFileExisting, MENU);
-	//enableCommand(IDM_FILE_RENAME, isFileExisting, MENU);
 	enableCommand(IDM_FILE_OPEN_CMD, isFileExisting, MENU);
 	enableCommand(IDM_FILE_OPEN_FOLDER, isFileExisting, MENU);
 	enableCommand(IDM_FILE_RELOAD, isFileExisting, MENU);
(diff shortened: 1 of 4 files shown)


donho:
@Yaron10
Please check this PR.

xomx:
@donho 
Commit tested for all above issues, thanks!

donho:
@Yaron10 
> enableCommand(IDM_FILE_SAVEAS, !curBuf->isInaccessible(), MENU);
enableCommand(IDM_FILE_RENAME, !curBuf->isInaccessible(), MENU);
enableCommand(IDM_VIEW_GOTO_ANOTHER_VIEW, !curBuf->isInaccessible(), MENU);
enableCommand(IDM_VIEW_CLONE_TO_ANOTHER_VIEW, !curBuf->isInaccessible(), MENU); 
>
> Why should those commands be disabled (especially the last two)?

* There is no sense to me to save as an empty file, though it's no harm to do it.
* Not allow to rename is the request from @xomx - but he is right: Not only renaming a placeholder is meaningless, but also it could have the undefined behaviour (tested).
* for go & clone to other view, IMO  the placeholder has only one job: holding the place, so why should we let them fly around between the views?

The code is already complex. With the addition of placeholder, the regressions & bugs could be introduced even more easier.
Making placeholders as restrict as possible (but let them do their essential job) to prevent from bugs & regressions is a good strategy to me.

> enableCommand(IDM_FILE_CONTAININGFOLDERASWORKSPACE, !curBuf->isInaccessible(), MENU);
>
>Why not isFileExisting?

For me the placeholder feature (isInaccessible state) for dealing with files existing on unmounted drives. whereas isFileExisting is rather deleted files (their parent folder is accessible) during Notepad++ session. But you're right - we can remove a network drive or a USB during the session. I will modify this part.

> How about adding
>
>bool oneTabOnly = (_pDocTab->nbItem() == 1 && !bothActive());
enableCommand(IDM_VIEW_GOTO_ANOTHER_VIEW, !oneTabOnly, MENU);
>
>?
>
>And what do you think of IDM_VIEW_GOTO_NEW_INSTANCE if oneTabOnly ?

Why not? But it's beyond of the scope of placeholders.
And the current behaviour is not incoherent, furthermore there's no complains about it until now.
Therefore it won't be done (at least not in this PR).

Yaron10:
@donho,

Thanks again for fixing the bugs.

donho:
Thank you @Yaron10 & @xomx for checking & testing & the pertinent suggestion!

PR #14291, opened 2023-10-29
first version of GH CI build

### LINKED ISSUE #12177: Move CI from Appveyor to GitHub Actions
The current CI configuration with Appveyor takes 1 hour approximately. It does not build various configuration in parallel. So, every jobs are waiting for another to complete. Even CI builds in PRs wait for previous commit or PR.

May I suggest to move the CI from Appveyor to GitHub Actions? I have tested the CI with msvc builds only in GitHub Actions which took between 7-8 minutes. The jobs are done in parallel. There is total 20 jobs allocated for free tier. So, CI builds in PRs will not wait for another commit or PR. What do you think about this proposal?

chcg:
fix #12177 

- fix merge issue for lexilla.mak
- fix clang compiler issue: ..\lexers\LexObjC.cxx(50,20): warning : unused function 'IsADigit' [-Wunused-function] [D:\a\notepad-plus-plus\notepad-plus-plus\lexilla\src\Lexilla.vcxproj]
- adapted unit test powershell scripts to run also on powershell 7
- bash unittest provides the correct result on GH CI

Todos:
- check if xml optimization could also be applied


chcg:
@donho I think you could have a look at the changes now. See https://github.com/chcg/notepad-plus-plus/actions/runs/6723197855 on the output of a CI run. I added one cmake job to ensure that the cmake config doesn't get broken. Apart from that the build jobs should be the same as for appveyor. Artefacts are slightly different as GH action always creates a zip container. This contains the Notepad++.exe. Do you think the N++.exe within the zip container should also have a modified name with the build infos like "Notepad++.Win32.Debug.exe" as we have it on appveyor?

I have to modify the bash functionlist unittest because on GH runners we get the correct expected result. This is unfortunately breaking the appveyor run. So it is a problem to have both CIs running in parallel for a while to check if differences appear.

I didn't adapt the pre check to avoid unnecessary builds on xml translation checkins, etc.

donho:
@chcg 
Thank you for your work.

> Do you think the N++.exe within the zip container should also have a modified name with the build infos like "Notepad++.Win32.Debug.exe" as we have it on appveyor?

No, it's not necessary. The zip file name indicate clearly the architecture and compiler platform. I think it's enough.

> I had to modify the bash functionlist unittest because on GH runners we get the correct expected result. This is unfortunately breaking the appveyor run. So it is a problem to have both CIs running in parallel for a while to check if differences appear.

IIRC, it's due to the different EOL (Linux vs Windows) - the parser of batch file is buggy while running with EOL of Linux. It's good that Github CI doesn't transform file from Windows' EOL to Linux's one.


> We could add another workflow which just runs occasionally e.g. with each tag containing e.g. clang, MSVC code analysis, ubuntu linux.

Why not? While these tasks don't penalize the whole CI performance, it's OK to me.

> I didn't adapt the pre check to avoid unnecessary builds on xml translation checkins, etc.
I think this could also be done in an additional PR later.

OK, no problem.
I'll merge this PR now. Feel free to the additional PR whenever you want.

PR #14295, opened 2023-10-31
Add multi-select commands

### LINKED ISSUE #8203: [Feature Request] Add SCI_MULTIPLESELECTADDEACH and SCI_MULTIPLESELECTADDNEXT
[Both](https://www.scintilla.org/ScintillaDoc.html#SCI_MULTIPLESELECTADDNEXT) are parameter-less functions like SCI_CUT, SCI_PASTE etc.
but provide powerful functionality to npp, so I suggest to add them to the list of scintilla commands (shortcut mapper) and respective places.

donho:
Add "Multi-select All" & "Multi-select Next" for 4 modes to multi-select the occurrences respectively:
1. Match Case Only
2. Match Whole Word Only
3. Match Case & Whole Word
4. Ignore Case & Whole Word

Add also 2 commands "Undo Last Multi-select" & "Skip Current Multi-select" for undoing last slection & skipping current selection and do the next one.

Users can asign the shortcuts to these 10 commands so multi-occurrence can be easily selected via shortcuts and be modified silmutanously.

This feature is inspired by: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8203#issuecomment-623712205

Fix #8203


donho:
Translation entries will be added after the rewording of command names being done.

Yaron10:
@donho,

👍 

```
cpp
Acpp
Cpp
cpp
```

**1.**
Select the first `cpp`.
"Multi-Select Next".
-> The Case/Word commands do not work properly.

**2.**
"Undo Last Multi-Select" is actually "Undo **Current** Multi-Select" (i.e. main caret), isn't it?

**3.**
"Undo Last Multi-Select" does NOT work on the last remaining selection.

**4.** 
I understand there's no way to remove a selection at a random position.

**5.**
"Skip Current Multi-Select" is actually "**Move** Current Multi-Select Forward/Backward", isn't it?

**6.**
If you select the first `cpp` and "Multi-Select All", the main caret is at the end of the third line.
If you just place the caret on the first `cpp` and "Multi-Select All", the main caret is at the end of the last line.

Yaron10:
**7.**
The position of the new commands in the Shortcut Mapper could be changed.

donho:
@Yaron10 
> Select the first cpp.
"Multi-Select Next".
-> The Case/Word commands do not work properly.

Sorry for not testing thoroughly before committing. It's fixed in the latest commit.
And thank you for spotting the bug.

> "Undo Last Multi-Select" is actually "Undo Current Multi-Select" (i.e. main caret), isn't it?

Hmm... if user runs "Multi-select All" command, it's **"Undo Last Multi-Select"**; however, if user runs "Multi-select All"  it's **"Undo Current Multi-Select"**.
Can we reword this command name to make it suites both situation?

> "Undo Last Multi-Select" does NOT work on the last remaining selection.

It's the behaviour of Scintilla: https://www.scintilla.org/ScintillaDoc.html#SCI_DROPSELECTIONN

> I understand there's no way to remove a selection at a random position.

From the current implementation, via keyboard shortcut, no. But user can always use Ctrl+Mouse for unselecting on any random position.

> "Skip Current Multi-Select" is actually "Move Current Multi-Select Forward/Backward", isn't it?

You've got the idea. The explicit naming should be: "Skip Current Selection & Move to Forward Multi-select", but it's too long. Sometimes, shorter is better in our life :)
How about ***Skip Current & Go to Next Multi-select*** ?

> If you select the first cpp and "Multi-Select All", the main caret is at the end of the third line.
If you just place the caret on the first cpp and "Multi-Select All", the main caret is at the end of the last line.

Well spotted! It's fixed in the latest commit.

> The position of the new commands in the Shortcut Mapper could be changed.

Done in the latest commit.

Yaron10:
@donho,

👍 
Thank you for fixing the bugs.

The Undo/Skip commands can be tricky.

**STR:**
Select `Cpp`.
"Multi-Select Next" -> "Match Case & Word".
Remove the selection.
Manually select the first `cpp`.
"Skip Current".

**Result:**
The _last_ `cpp` is selected.
If you ignore the previous "Match Case & Word", the _second_ `cpp` (in `Acpp`) should be selected.

Are those commands worth the extra complication and confusion?
Are they really useful?

Yaron10:
> If you ignore the previous "Match Case & Word"...

I'd like to rephrase that.
On manually selecting the first `cpp`, it's "a whole new ballgame".
Using the old parameters would be confusing.

donho:
@Yaron10 

> **Result:**
> The _last_ `cpp` is selected.
> If you ignore the previous "Match Case & Word", the _second_ `cpp` (in `Acpp`) should be selected.

Case and whole word settings are set on each "Multi-select All" & "Multi-select Next" command, so "Skip" command will follow the last Case & Word settings for moving onto the right selection.
These commands are supposed to use sequentially with shortcuts, so the result of your scenario seems a little bit wired, but not illogical.

> Are those commands worth the extra complication and confusion?
Are they really useful?

Yes, I do believe they are useful for people who use only keyboard for setting multi-selection.

Yaron10:
@donho,

As you understand.
Thank you again. The multi-select is a nice addition. 👍

donho:
@Yaron10 
Thank you for your review and pertinent remark.

BTW:
>> "Undo Last Multi-Select" is actually "Undo Current Multi-Select" (i.e. main caret), isn't it?
>
>Hmm... if user runs "Multi-select All" command, it's "Undo Last Multi-Select"; however, if user runs "Multi-select All" it's "Undo Current Multi-Select".
Can we reword this command name to make it suites both situation?

***"Undo Latest Added Multi-Select"*** could be a good compromise.

Yaron10:
@donho,

I don't really understand the use of "Undo" here. 
And "Skip" is wrong IMO. - You do not skip; you MOVE the selection.

I'd use "Remove main-selection" or "Unselect main-selection".
And "Move main-selection to the next occurrence".

It's not the best either. :)
We need a native English speaker.

donho:
I don't think it's the problem of being native speaker or not.
We have the different view and sense for this feature.
FWIW:
https://stackoverflow.com/questions/16495727/multiple-select-in-visual-studio

I will merge this PR, but I am opened for renaming these commands in the future (before Notepad++ next release).

Thank you for your suggestions.

alankilborn:
@Yaron10 said:

---

>Select the first . (try from right to left & from left to right).  Multi-select All.

All possible choices lead to `cpp` being selected in every line.

---

>Macro?

Are you asking for all of these commands to be macro-able?

---

Sorry if I'm being dense but sometimes, without explanation, things aren't very clear to me.

Yaron10:
@alankilborn,

> All possible choices lead to `cpp` being selected in every line.

If I select from from right to left, the `NPP` occurrences are selected.

> Are you asking for all of these commands to be macro-able?

Whatever you and Don think can be useful.

Yaron10:
@alankilborn,

> I understand there's no way to remove a selection at a random position.
>> But user can always use Ctrl+Mouse for unselecting on any random position.

Does that (Ctrl+Mouse) work for you?

alankilborn:
>Does that (Ctrl+Mouse) work for you?

Nope; I don't see any way to remove one of the existing selections.

Yaron10:
> Nope; I don't see any way to remove one of the existing selections.

👍 

***

Line `10` can be the (visible-)main-caret-line before "Multi-Select All".
After that command is executed, line `5000` might be the visible-main-caret-line.
Is it possible to make the original main-caret-line visible?

***

Not directly related.
A unique style for the main-selection might be a nice improvement.

alankilborn:
> A unique style for the main-selection might be a nice improvement.

Agreed.

What I do for now is have my main-caret in red:
image

Yaron10:
> What I do for now is have my main-caret in red

👍 
That's nice. I suppose it's red only when you have multiple selections.

It can be either the caret or the selection background color.
Playing with "Multi-Select", I've realized the current distinction may not be sufficient.

alankilborn:
>the current distinction

...which is absolutely no distinction? :-)

Yaron10:
תמונה

Black vs. dark gray. Practically, you're right, no clear distinction. :)

donho:
@Yaron10 
> Select the first . (try from right to left & from left to right).
Multi-select All.

What is it an expected result for you?

> Macro?

I'll consider it.

donho:
> What I do for now is have my main-caret in red:

> Black vs. dark gray. Practically, you're right, no clear distinction. :)

Sorry guys, I don't think it's possible.
If you see the possibility, please do a PR, or let me know how.

Yaron10:
@donho,

> Select the first . (try from right to left & from left to right).
 > Multi-select All.
>> What is it an expected result for you?

I'd expect all the dots (`.`) to be selected.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/7f2c303ac6583f370e06ad9ad121fa70db23c114/PowerEditor/src/NppCommands.cpp#L2015

Why not call that function only if there's no selection?

***

> If you see any possibility, please do a PR, or let me know how.

Again, not directly relate.
If anyone wants to further explore it.
https://www.scintilla.org/ScintillaDoc.html#SC_ELEMENT_CARET_ADDITIONAL.
https://www.scintilla.org/ScintillaDoc.html#SCI_SETMAINSELECTION.

donho:
@Yaron10 
> I'd expect all the dots (`.`) to be selected.
> Why not call that function only if there's no selection?

Yes, what you say makes sens.
Fixed in #14301.

> Macro?

Make the multi-select commands set macro record-able in #14301.

donho:
@Yaron10 
> Again, not directly related.
If anyone wants to further explore it.
https://www.scintilla.org/ScintillaDoc.html#SC_ELEMENT_CARET_ADDITIONAL.
https://www.scintilla.org/ScintillaDoc.html#SCI_SETMAINSELECTION.

Thank you, I will check these links.
In the meanwhile, @alankilborn or @Yaron10 could you create a issue as FR for this feature so there is a simple "specification" that I can follow?

Yaron10:
@donho,

Great. 👍 
Thank you again.

Now only the localization is missing. :)

> could you create a issue as FR for this feature so there is a simple "specification" that I can follow?

If @alankilborn doesn't do it, I'll open the issue tomorrow.
I must retire now.

Thank you for that too.

alankilborn:
Considering @Yaron10 's comment https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14295#issuecomment-1791643979 ...

In a fresh portable, I see even LESS distinction among carets that you showed; as an example, I am having difficulty quickly determining the main selection/caret here:

image

That's why I said "absolutely no distinction"...because I didn't originally believe there was any (I was wrong...it is very slight but it is there if you look at it hard and long enough).

Anyway...as stated...it could be better.

Yaron10:
@donho,

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14302.
Thank you!

@alankilborn,

I hope you don't mind I've used your image in that issue.

Yaron10:
https://sourceforge.net/p/scintilla/bugs/2410/

PR #14309, opened 2023-11-03
Inactive appveyor partially

donho:
Keep only x64 release build in Appveyor


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/02e4202d681a8cb79ff5db5ceabeba262969ce99

PR #14330, opened 2023-11-06
Fix hasSelection wrongly detected

### LINKED ISSUE #14322: Multiple Selections: Copy, Cut and "Convert Case" commends may be wrongly disabled
**STR:**
Enable "Multiple Selections".
```
Test1
Test2
```
Select `Test1`.
Press Ctrl and **single-click** on `Test2` (thin selection).

**Result:**
The Copy, Cut and "Convert Case" commends are wrongly disabled.

 https://github.com/notepad-plus-plus/notepad-plus-plus/blob/950236bb8f8e85d678912939c00183e85c6417f6/PowerEditor/src/Notepad_plus.cpp#L2558

Replacing that line with
```cpp
bool hasSelection = !_pEditView->execute(SCI_GETSELECTIONEMPTY);
```
solves the issue.

_There might be other issues in selection-related operations caused by not using `SCI_GETSELECTIONEMPTY`._

***

Notepad++ v8.5.8.

donho:
Fix #14322

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2555,7 +2555,7 @@ void Notepad_plus::enableCommand(int cmdID, bool doEnable, int which) const
 
 void Notepad_plus::checkClipboard()
 {
-	bool hasSelection = (_pEditView->execute(SCI_GETSELECTIONSTART) != _pEditView->execute(SCI_GETSELECTIONEND));
+	bool hasSelection = _pEditView->hasSelection();
 	bool canPaste = (_pEditView->execute(SCI_CANPASTE) != 0);
 	enableCommand(IDM_EDIT_CUT, hasSelection, MENU | TOOLBAR);
 	enableCommand(IDM_EDIT_COPY, hasSelection, MENU | TOOLBAR);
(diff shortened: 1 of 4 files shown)


Yaron10:
@donho,

Thank you for your work.

```
Test1
Test2
```
Select `Test1`.
Press Ctrl and **single-click** on `Test2` (thin selection).
Multi-select All. 

**Result:**
NPP crashes.

In this case we shouldn't care about other-selections. The commands should operate on the main-caret.
\- If you invoke the Find dialog, `Test2` would be correctly set in "Find what".

I haven't tested the shortcuts as I don't understand the concept.

donho:
@Yaron10 
> Result:
NPP crashes.

>In this case we shouldn't care about other-selections. The commands should operate on the main-caret.
>- If you invoke the Find dialog, Test2 would be correctly set in "Find what".

Thank you for finding it!
Hmm... there's an infinite loop on SCI_MULTIPLESELECTADDEACH while multi-selection is on and main selection has no selection, wired.
I will try to prevent from such situation.

> I haven't tested the shortcuts as I don't understand the concept.

No problem.

Yaron10:
Regardless of the crash, the commands should operate on the main-caret.
1. The other-selections are irrelevant in this case.
2. What should happen if you have **two** other-selections?

donho:
> What should happen if you have two other-selections?

Fixed it with the old code. It should work better now.

Yaron10:
Great. 👍

Thank you again.

Issue #14266, opened 2023-10-25
[Feature Request] Add a feature for transforming the column mode to multi-caret on HOME/END/Arrow keys


nrv-96:
### Description of the Feature

Please incorporate a feature similar to VS Code, as demonstrated in the attached video, enabling the editing of the selected line at the end of the line.

### Expected Behavior
At present, this particular feature is absent in Notepad++. Could you kindly consider integrating this functionality into the application?

https://github.com/notepad-plus-plus/notepad-plus-plus/assets/62826272/bbfab116-db6e-4e54-83f0-9cfc299c12c1

Ekopalypse:
There is a plugin called [BetterMultiSelection](https://github.com/dail8859/BetterMultiSelection/) which then allows Npp to do this kind of editing.

nrv-96:
@Ekopalypse , can we add it as core functionality? This would be helpful to everyone.

Ekopalypse:
@nrv-96 - it's up to @donho to decide, but in the past the policy has tended to be - "if there's a plugin that can do this, it's more likely not to be included in core".
Personally, I would like to see this as a core feature as well.

nrv-96:
I agree with you @Ekopalypse . While I understand the policy about plugins, having this as a core feature could greatly enhance the overall functionality. It's up to @donho  to decide, but considering its potential benefits, I also support the idea of incorporating it into the core. Everyone can enjoy a seamless and integrated experience without the need for external plugins.

@donho, I kindly request you to consider incorporating  **BetterMultiSelection** as a core feature instead of an external plugin.

alankilborn:
>this plugin (BetterMultiselection) was last updated in 2022

There's not much value in that statement, as 2022 is hardly ancient history.

alankilborn:
I'd think that a "problem" with adding this as a "core" feature is that these types of functions are really Scintilla owned, and not Notepad++ owned.  Thus, shouldn't something like this be in Scintilla's careting support?

Ekopalypse:
In general, I agree that Scintilla would be the best place to ask for such a feature, but if I'm not mistaken, there was already such a request and it was rejected (I hope I don't remember wrong - I'm going to research scintillas FRs).
Npp would be the next best option.

donho:
Frankly, I kind of like this feature. If it's not hard/complex for implementing, it will be great to include it in Notepad++.
I will see what I can do about it.

@nrv-96 
[...]

Issue #9260, opened 2020-12-13
Edit menu Paste is NOT the same as a Ctrl+v paste


sasumner:
### Description of the Issue

The _Edit_ menu has an entry called _Paste_ and by default keymapping is shown as mapping to Ctrl+v and Shift+Ins:

![image](https://user-images.githubusercontent.com/30118311/102012536-0efefa00-3d19-11eb-970e-b27ccba821ce.png)

However, the actions performed when the menu item is invoked from the menu are different from the actions performed when Ctrl+v is pressed.

This seems wrong because from what appears on the _Edit_ menu, the commands are one-and-the-same.

### Steps to Reproduce the Issue

1. Turn visible line endings on by selecting _View_ menu > _Show Symbol_ > _Show End of Line_
2. Create some data in a tab; make the tab be of "Windows (CR LF)" line-ending type:
![image](https://user-images.githubusercontent.com/30118311/102012730-212d6800-3d1a-11eb-8cb9-f4c169415fe5.png)
3. Change some or all line-endings in this tab to "LF" by a regular expression replacement of `\R` with `\n` (okay, not something normally done, but this is for point of illustration that the menu item paste is doing something different than Ctrl+v); observe:
![image](https://user-images.githubusercontent.com/30118311/102012879-b4ff3400-3d1a-11eb-89d4-32f287318e3a.png)
4. Select lines 3 and 4 and press Ctrl+c ; move to end of document and press Ctrl+v ; observe:
![image](https://user-images.githubusercontent.com/30118311/102012921-f8f23900-3d1a-11eb-86c3-1df91386220b.png)
5. Press Ctrl+z to undo the actions of step 4 and return to the state shown in step 3.
6. Select lines 3 and 4 and press Ctrl+c ; move to end of document and invoke _Paste_ from the _Edit_ menu ; observe:
![image](https://user-images.githubusercontent.com/30118311/102012972-335bd600-3d1b-11eb-8b93-80f130a34dea.png)

### Expected Behavior

A consistent paste action whether the _Paste_ menu entry is invoked, or Ctrl+v is pressed.
In other words, the line-endings after step 4 and after step 6 should be the same.

### Actual Behavior

Different line-endings result after step 4 and step 6.
Specifically, the _Edit_ menu _Paste_ has changed all lines of the file in addition to pasting the desired data, whereas Ctrl+v has only pasted the desired data.

### Other

[...]

PR #14338, opened 2023-11-08
Make multi-select Copy/Paste behaviour as same as column edit's one

donho:
Copy some multi-select texts and paste them make all text glued all together. This commit makes pasted texts separated by EOL, as column selection's Copy/Paste behaviour.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14266#issuecomment-1799203336


donho:
@Ekopalypse @vinsworldcom

vinsworldcom:
> @Ekopalypse @vinsworldcom
> Could you check this PR please?

I adds the newlines, but not quite as expected:

```
two = obj[""]
three = obj[""]
four = obj[""]
```

With the cursor at position 0 (to the left of the `t` in `two`):

1) use `ALT+SHIFT+Down` to extend caret to the `f` in `four`.
2) Next, use `CTRL+SHIFT+Right` to create a multi-selection of the words "two", "three" and "four".
3) `CTRL+C` to copy the selected words
4) Use the `Right` arrow key to move the multi-carets to between the double-quotes (on each line)
5) `CTRL+V`

Outcome:

```
two = obj["two
three
four
"]
three = obj["two
three
four
"]
four = obj["two
three
four
"]
```

Expected (and what happens in [BetterMultiSelection](https://github.com/dail8859/BetterMultiSelection)):

```
two = obj["two"]
three = obj["three"]
four = obj["four"]
```

Cheers.

donho:
@vinsworldcom 
Thank you for testing.
After committing, I just realized there was a problem of copied text format, so I did a fix.
I just committed the fix.
Could you try the latest commit and confirm me if it's OK the current state of PR?

vinsworldcom:
> Could you try the latest commit and confirm me if it's OK the current state of PR?

Better, but still not quite.  Doing the same test above, the output it:

```
two = obj[""]
three = obj[""]
four = obj["two"]
            three
            four
```

Cheers.

Yaron10:
```
Test1
Test2
```

Select `Test1`.
Press Ctrl and select `Test2`.
Copy.
Add a new line.
Paste.

**Result:**
```
Test1
Test2
```

**Expected:**
```
Test1Test2
```

vinsworldcom:
I would not expect that. You're copying Text2  which is under Text1
visually, thus when pasted I would expect it to appear the same..

I'm pretty sure if you use better multi-selection you will get the results
I describe.

If we can't agree on how to implement this feature pasting, I will just
continue to use better multi-select to get the results I expect.

Happy either way..

Cheers.

Yaron10:
@vinsworldcom,

Try it with
```
Test1 Test2
```

donho:
@Yaron10 
Frankly, I've never liked the original copy/paste behaviour of multi-select - it makes no sense to me to put all words together without delimiter. 
Using EOL might not be the preferred choice of all users, but at least users can distinguish the different selections/words they have copied. And since the multi-select paste feature needs such format, I think it's (EOL as delimiter) the best compromise.

vinsworldcom:
@donho for the few quick tests I did, selecting from top-down, bottum-up, partial, whole-word, navigation with arrow-key and `CTRL` + arrow keys, `HOME`, `END` ... It seemed to me the same as BetterMultiSelection plugin.  I had two instances open, one with this new commit and one with 8.5.8+BetterMultiSelection plugin - they both behaved the same for my limited testing.

I'm sure an immense amount of work dealing with corner cases - so thank you for your dedication to getting this done!

I'm going to remove BetterMultiSelection for the next few days while going through my normal editing and use of N++.  I'll see if anything pops up.

Cheers.

Yaron10:
@donho,

> Frankly, I've never liked the original copy/paste behaviour of multi-select - it makes no sense to me to put all words together without delimiter.

The current master behavior can be useful in many cases.
But, I agree, in many other cases splitting the words would be preferred.

***
If a doc is read-only, the flag is removed after Ctrl+V.
***

Following @vinsworldcom STR in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14338#issuecomment-1802645330.

**Ctrl+V:**
```
two "]
three "]
four "]
```
(Undo is disabled).

**Edit -> Paste:**
```
"]
"]
two "]
three 
four 
```

Yaron10:
> Use the Right arrow key to move the multi-carets to between the double-quotes (on each line)

I thought Shift should be pressed.

Edit -> Paste:
```
two = obj[""]
three = obj[""]
four = obj["two "]
            three 
            four 
```

donho:
@Yaron10 
Thank you for testing and finding bugs.

> If a doc is read-only, the flag is removed after Ctrl+V.

Fixed in the latest commit.

> Edit -> Paste:

In the perfect world it should work also with Ctrl-V, but in real life ppl just don't use the the Paste on the menu item.
But we all wish be in the perfect world, so it's also fixed in the latest commit. :)

vinsworldcom:
@Yaron10 , if I do your latest test case in the latest commit artifact and in N++ with BetterMultiSelection, I get the same results - so at least behavior is consistent - if not ideal.

Cheers.

donho:
>>Use ALT+SHIFT+Down to extend caret to the f in four.
Next, use CTRL+SHIFT+Right to create a multi-selection of the words "two", "three" and "four".
CTRL+C to copy the selected words
--
Place the caret between "" in the last line.
Press Ctrl and place the caret between "" in the first line.
Press Ctrl and place the caret between "" in the second line.
Press Ctrl and place the caret between "" in the third line.
Paste.

>> Why should the result in the first three lines be different? - The main caret and additional carets are in the same positions as in 

> if I do your latest test case in the latest commit artifact and in N++ with BetterMultiSelection, I get the same results - so at least behavior is consistent - if not ideal.

In BetterMultiSelection & this PR current state, the behaviour is the same - it ensures both number of source (occurrences to paste in clipboard) and number of targets (multi-selection) are the same to do this operation.

@Yaron10 & @vinsworldcom  Such behaviour can be changed - what's your preferred behaviour then?

vinsworldcom:
The current behaviour is consistent with what I would expect - though I've been user BetterMultiSelection for years now so my view of behaviour is probably skewed to the way that plugin operates vs. what new possibilities could be.

I generally don't do the thing @Yaron10  posits above - manually clicking the paste-to locations in a different order expecting they would appear in that new order and perhaps repeat if more paste-to locations are selected than objects copied.  I don't really have a guess as to how this *should* behave since I've never really had a use case for it.

With respect, I will caution that looking at these corner cases and creating an interpretation for them will surely be "incorrect" in at least someone else's opinion so not sure how far down the added implementation road you want to go?

Cheers.

Yaron10:
As opposed to @vinsworldcom, I rarely use multiple-selections in real life.
If you think the current behavior is OK, so be it.

Yaron10:
@donho,

Thank you for fixing the Undo issue.

alankilborn:
@donho said:

> but in real life ppl just don't use the the Paste on the menu item.

Au contraire ... the only people that use the menu item for Paste are the ones that want to convert all line-endings in the current file.  #9260  :-)

alankilborn:
@donho / @Yaron10 

Let me say one more thing about this:

>but in real life ppl just don't use the the Paste on the menu item.

@Yaron noted:

>I usually use Paste in the context menu.

---

Let's try to remember going forward that users have different ways of doing things.  With configurable editor-context menu, configurable tab-right-click menu, mapped-keycombos, or just plain using something on the "main menu", ALL are equally valid ways to do something, and user picks which is best for himself.  These things should all in the end execute the same code, no matter how invoked.  

I'm not sure how "paste" ever became different depending upon how invoked...

I recently took care of the strangeness with _Line Duplicate_...

All of these are good steps.

vinsworldcom:
> Finally, assume the following code in python
> 
> ```python
> def test1():
>     pass
> 
> def test2():
>     pass
> ```
> 
> Select both instances of `pass` and press `enter` This will work if you have disabled the `settings->preferences->auto-completion->auto-indent` but NOT if this setting is active, which is the default.

Yikes!  Can't believe I missed that one.  Perhaps since I was using the default settings in a fresh portable install to test.  I always have that setting enabled and you're right, it doesn't work as expected, but does with BetterMultiSelection plugin enabled.

Cheers.

Ekopalypse:
Just to make sure, all three tests behave differently and in my opinion correctly with the BetterMultiSelection plugin. The first and second are rather ignorable for me, the last one is more of a problem for me as I use npp 99% of the time with the autoindent function.

vinsworldcom:
> the last one is more of a problem for me as I use npp 99% of the time with the autoindent feature.

Agree!

donho:
PR #14355 has enhanced multi-edit **paste** and **Enter** key-in.
It also disables auto-indent while multi-editing.

Please check it to see if it's what you need.

vinsworldcom:
> It also disables auto-indent while multi-editing.

This is a non-starter for me.  I always have auto-indent enabled and expect it to function during multi-caret editing - as it does today.  Making it disabled during multi-caret editing breaks current capability and even with BetterMultiSelect, this will no longer work how it does today.  I disabling auto-indent during multi-select as regressive.

Cheers.

CennoxX:
If I use the builds of https://github.com/notepad-plus-plus/notepad-plus-plus/commit/176475866985b43e767222c786870e2024ee104f then the deletions at the beginning/end of lines are without effect:

**delete key on line end**
If I press the delete key on the line end, while having multiple cursors, nothing happens.
I can simulate the expected behaviour by pressing `[shift]`+`[right]` and `[delete]`
output

**backspace on line start**
Same problem on line start, while having multiple cursors, and pressing backspace, though this works for example, if I pressed enter before.
output2

vinsworldcom:
@CennoxX , indeed, I can confirm.  With BetterMultiSelection, it does behave as expected - Delete at the end of the lines removes the "newline" and all the lines become one with multi-carets dispersed throughout continuing to delete as `Del` key is pressed.

Same with `Backspace` - although with BetterMultiSelection, no `Enter` is required first.

Cheers.

donho:
@CennoxX @vinsworldcom
> If I press the delete key on the line end, while having multiple cursors, nothing happens.

I have no solution yet. I know it's not perfect, but in the meanwhile, use Ctrl-DEL for deleting the EOL.

vinsworldcom:
It seems to fix the `Del` and `Backspace` issue for me!

You are a rock star!

Cheers.

donho:
Please check new commit in master: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c517985beb08c3f3478bcdbdffb58e6f0d912f3e.
The code is different from #14359 after some refactoring. Though I have checked roughly, there could be some regressions for Multi-edition features alltogether.

vinsworldcom:
There seems to be some inconsistency in caret movement with multi carets.  Use my previous example:

```

sub  {
    my ($) = @_;

    test
}


sub  {
    my ($) = @_;

    test
}

```

Highlight both "sub" and then press the down arrow key three times, to put the multi-carets on the lines with "test".  They should both be at the same position in the line; however, they are not:

```
sub  {
    my ($) = @_;

|   test
}


sub  {
    my ($) = @_;

  |  test
}
```

Where the pipe (`|`) above shows where the carets end up for me.  I would expect they should both be at the beginning of the "test" line, but the second one is 3 spaces in.

Cheers.

donho:
@vinsworldcom 
Just checked with BetterMultiSelect, and found out that it cannot do such thing. Could you confirm that?
Otherwise, it seems such wrong behaviour is a bug of Scintilla. Could you create an issue in Scintilla project please?

vinsworldcom:
My bad, I should have said both carets should be at the 3-spaces in position, since [sticky caret](https://www.scintilla.org/ScintillaDoc.html#SCI_SETCARETSTICKY) is enabled by default and the carets start at 3 positions in from the left on the "sub" line as they are placed after the `b` in "sub".  The output expected should be:

```
sub  {
    my ($) = @_;

  |  test
}


sub  {
    my ($) = @_;

  |  test
}
```

Where the pipe (|) above shows where the carets end up for me when using BetterMultiSelection.  This would be the case if using only a single caret - it would end up 3-spaces in due to sticky behavior.  So with multi carets, they both should "obey" this rule.

Cheers.

donho:
@vinsworldcom 
> I think the behavior of BetterMultiSelection is correct and both carets should end up 3 spaces in on the "test" line like I show in this post. My previous post was incorrect and for that I apologize.

Strangely, I cannot make BetterMultiSelection work with my v8.5.8 (official release) - to select both `sub` with Ctrl + Mouse Duble Clicked, can you? Or you're using BetterMultiSelection with the binary of the current master?

vinsworldcom:
I tried with both the latest artifact and the official 8.5.8 release - both with BetterMultiSelection - behave as described in my second post - with the carets being 3 spaces in on the "test" line after moving them down from the `b` on the "sub" line.

Note, you need to have multi-editing enabled:

image

It is not on by default in the 8.5.8 64-bit portable.

Cheers.

donho:
@vinsworldcom 
Thank you for you explanation. I can reproduce your result now.

> I tried with both the latest artifact and the official 8.5.8 release - both with BetterMultiSelection - behave as described in my second post - with the carets being 3 spaces in on the "test" line after moving them down from the b on the "sub" line.

Just tested in SciTE v5.3.9 - it's the same behaviour as under current master of Notepad++.
The bug has been reported here:
https://sourceforge.net/p/scintilla/bugs/2415/

Let's wait for the answer from Neil for this issue.

PR #14344, opened 2023-11-10
Fix network shared files saving regression

### LINKED ISSUE #14300: Corrupt file after editing/saving on macOS SMB share
### Description of the Issue
When saving a file on a macOS SMB share that was previously longer, the size doesn't actually change, resulting in duplicate data near the end of the file, thus kinda corrupting the file in the process.

1. Create a text file with different 3 lines (of the same length, because that makes it easier to explain/test the issue).
2. Save it on a macOS SMB share.
3. Delete the 2nd (middle) line.
4. Save the file again.
5. Reload from Disk (or load the file in Windows Notepad or macOS TextEdit).

### Expected Behavior
I would expect a file with only 2 lines, and the file size should now be smaller.

### Actual Behavior
There are still 3 lines in the file; the last 2 lines are now the same. Also, the file size hasn't changed.

This is with Noteapad++ v8.5.8, but it doesn't seem to happen with v8.5.7 (and probably earlier, because I haven't encountered this issue before). When saving the file directly on my Windows machine this doesn't happen, nor when saving it on a Windows network share.

### Debug Information
Notepad++ v8.5.8   (64-bit)
Build time : Oct 15 2023 - 21:43:56
Command Line : "\\smbshare\username\test.txt" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.3636
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
Cause:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14300#issuecomment-1791313760

Solution:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14300#issuecomment-1798075334

Enhancement:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14300#issuecomment-1804910680

Fix  #14300

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -37,7 +37,7 @@ Win32_IO_File::Win32_IO_File(const wchar_t *fname)
 			// Store the file creation date & attributes for a possible use later...
 			::GetFileAttributesExW(fname, GetFileExInfoStandard, &attributes_original);
 
-			// Test
+			// Check the existence of Alternate Data Streams
 			WIN32_FIND_STREAM_DATA findData;
 			HANDLE hFind = FindFirstStreamW(fname, FindStreamInfoStandard, &findData, 0);
 			if (hFind != INVALID_HANDLE_VALUE) // Alternate Data Streams found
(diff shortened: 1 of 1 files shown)


TaleTN:
Shouldn't it be like this?

```
if (hFind != INVALID_HANDLE_VALUE) // Alternate Data Streams found
{
	dispParam = TRUNCATE_EXISTING;
	FindClose(hFind);
}
```

xomx:
@donho 
I only found the same problem as @TaleTN .
Otherwise it seems to me ok.
I would like to test it more thoroughly, but I cannot until Monday/Tuesday, sorry.

donho:
@TaleTN 
Well spotted! Thank you.

@xomx 
No problem. I will merge this PR right now but feel free to report here when you find any issue.
Thank you again for your investigation and your help for this PR.

PR #14350, opened 2023-11-13
Make multi-select background & caret colour customizable

### LINKED ISSUE #14302: [Enhancement] Better distinction between main-selection and other-selections
**STR:**
Create multiple selections.

**Result:**
The distinction between the main-selection and the other-selections is not clear.

תמונה

**Suggestion:**
1. Add a unique style for the other-selections caret color.
2. Add a unique style for the other-selections background color.

Personally I prefer the first option.

**Ref:**
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14295#issuecomment-1791745374.

***

Notepad++ v.8.5.8.

donho:
Fix #14302

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -3081,7 +3081,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			_subEditView.setCRLF();
 			return TRUE;
 		}
-		
+
 		case NPPM_INTERNAL_NPCFORMCHANGED:
 		{
 			_mainEditView.setNpcAndCcUniEOL();
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2874,7 +2874,7 @@ void ScintillaEditView::performGlobalStyles()
 
 
 	COLORREF selectMultiSelectColorBack = liteGrey;
-	pStyle = stylers.findByName(TEXT("Multi-selected text colour"));
+	pStyle = stylers.findByName(TEXT("Multi-selected text color"));
 	if (pStyle)
 	{
 		selectMultiSelectColorBack = pStyle->_bgColor;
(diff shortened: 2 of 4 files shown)


alankilborn:
Comments from testing:

* Is "Multi-selected..." and "Multi-edit..." the best wording here?  Just asking...and I know the next thing that will happen is I'll be asked for what I'd suggest, so:  "Additional...", "Secondary...", "Non-main..." [not sure any of these are better :-( ].

* Unlike other modifications to settings, the effect from changing the text color won't appear until user saves and the Style Configurator window closes.  OR user can leave Style Conf. window open and switch focus back to editor window to see change--but user may not know this and be slightly confused on why they don't get instant feedback on their color change.  (This could be a Scintilla limitation -- because it colors selected text in an unfocused window _differently_) 

All in all, though...GREAT WORK for those of us that will use this setting (maybe not many).

donho:
@alankilborn 
I'm happy about that you're happy about this implementation :)

> Is "Multi-selected..." and "Multi-edit..." the best wording here? Just asking...and I know the next thing that will happen is I'll be asked for what I'd suggest, so: "Additional...", "Secondary...", "Non-main..." [not sure any of these are better :-( ].

"Additional...", "Secondary...", "Non-main..." are probably the terms more accurate, if you are already an expert of Notepad++ & Scintilla in the context of multi-selection. However, IMO, they don't ring the bell for the most of users. Whereas "Multi-selected..." and "Multi-edit...", users know immediately what they are about without guessing, though I agree the current naming might be  a little bit confusing:
* **Multi-selected text color** vs **Selected text colour**
* **Multi-edit carets color** vs **Caret colour**

It could be less confusing to add **Main** in front of **Selected text colour** and **Caret colour**. Unfortunately they cannot be changed because the both names are used in both stylers.xml and source codes for getting style info.
I'm not worrying about it anyway - users will play around, then they will realize the difference.
Of course any other suggestions are welcome.

> Unlike other modifications to settings, the effect from changing the text color won't appear until user saves and the Style Configurator window closes. 

By "text color", I suppose that you mean "**selected** text color **background**". In Scintilla API, there's only one selection element for inactive window: **SC_ELEMENT_SELECTION_INACTIVE_BACK**
ref: https://www.scintilla.org/ScintillaDoc.html#ElementColours
If you find other ways to enhance this point, please let me know.

Yaron10:
@donho,

👍 
Great!

We should hope that Scintilla implements `SC_ELEMENT_SELECTION_ADDITIONAL_INACTIVE_BACK`.

vinsworldcom:
I currently do this with a PythonScript:

image

```
editor.setAdditionalCaretFore((255, 0, 0))
```

Cheers.

alankilborn:
@vinsworldcom 

> I currently do this with a PythonScript

Yes, but soon you won't have to. :-)

PR #14355, opened 2023-11-14
Enhance multi-edit paste and Enter key type

donho:
Also disable auto-indent during multi-editing.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14338#issuecomment-1809045648

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -356,6 +356,10 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_mainEditView.execute(SCI_SETMULTIPASTE, SC_MULTIPASTE_EACH);
 	_subEditView.execute(SCI_SETMULTIPASTE, SC_MULTIPASTE_EACH);
 
+	// Turn auto-completion into each multi-select on
+	_mainEditView.execute(SCI_AUTOCSETMULTI, SC_MULTIAUTOC_EACH);
+	_subEditView.execute(SCI_AUTOCSETMULTI, SC_MULTIAUTOC_EACH);
+
 	// allow user to start selecting as a stream block, then switch to a column block by adding Alt keypress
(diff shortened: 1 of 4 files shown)


donho:
@Ekopalypse @vinsworldcom

This PR should fix [the bugs](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14338#issuecomment-1809045648) reported by @Ekopalypse.
Could you test it and let me know if there are somethings need to be enhanced?

vinsworldcom:
I don't like having to disable auto-indent during multi caret edit.  This is not needed for BetterMultiSelect and it really impacts how I use it.  Using @Ekopalypse example:

```Perl
sub test1 {
    my ($) = @_;

    |
}


sub test2 {
    my ($) = @_;

    |
}

```

I should be able to add commands with multi-caret at the pipe (`|`) symbols and press `Enter` and get the auto-indent.  Before this commit with BetterMultiSelection, this works.  After this commit - **_even with_** BetterMultiSelect - it does not work.

Cheers.

donho:
@vinsworldcom 
> I don't like having to disable auto-indent during multi caret edit. This is not needed for BetterMultiSelect and it really impacts how I use it. Using @Ekopalypse example:

OK, I will see what I can do for it.

donho:
@vinsworldcom 
Could you test the latest commit please? It should work now.
OTOH, it could break BetterMultiSelet... I didn't test with it though.

Ekopalypse:
Works as expected, thank you very much. I'll be testing this version as my "workhorse" for the next couple days to see if anything else pops up, but so far it looks good.

vinsworldcom:
> Works as expected, thank you very much. I'll be testing this version as my "workhorse" for the next couple days to see if anything else pops up, but so far it looks good.

My sentiments exactly.  I'll continue to test, but it works the same both with or without BetterMultiSeletion.

Cheers.

Issue #13084, opened 2023-02-09
[Feature request] Solution to the "lost panel" problem


alankilborn:
### Description of the Issue
Sometimes panels seemingly get "lost" in Notepad++.
Typically this happens with the _Search results_ window/panel.
The size of the window is reduced (somehow) to zero or near zero, and the user can no longer locate the panel.
Users seek technical support when this happens, resulting in a cumbersome and raw "edit the config file" approach to solution, meaning restoration to a viewable and workable size.

It would be nice if some means for locating such a "lost" window were available in the Notepad++ UI itself; perhaps some sort of minimal size limit, or perhaps a command to provide more size to a suspected lost panel.  We leave it to Notepad++ devs to come up with the best solution.

More discussion on this problem is located in the Community site, including ideas for solutions, [HERE](https://community.notepad-plus-plus.org/topic/23344/).

### Debug Information
Comments based on behavior observed in Notepad++ 8.4.8.

Slayer95:
This is a bug happening since at least 8.4.2. <i>Sometimes</i>, a ``Search All`` execution will suddenly fail and subsequent fails will too.

pryrt:
To the Developers: [here's another report](https://community.notepad-plus-plus.org/topic/24588/find-all-in-all-open-documents-simply-closes-find-window-without-results) in the forum

A possible practical and simple solution: Add a **View** menu command called **Reset Docking Panel Sizes** which would take all four docking panel edges immediately back to their default width -- that would then immediately show any Panels that had been hidden because they were 0 pixels.  Or call it **Reset Any Tiny Docking Panels** which would only change Panels sizes that were 10pixels or fewer.  Or something.

Slayer95:
Since these reports didn't exist before, it's reasonable to hypothesize that something has gone wrong in the software side, rather than suddenly developers becoming dumb and triggering these "accidents".

Last time this happened, I couldn't reproduce your fixing steps as shown in the images posted, but editing the config manually did work. Note, however, that I _can_ use the mouse resize a normally-working search results panel to 0, and back to normal.

pryrt:
> Since these reports didn't exist before

[...]

PR #14356, opened 2023-11-15
Fix possible PluginsManager FindClose WINAPI errors

xomx:
Otherwise the ERROR_INVALID_HANDLE (0x6) can be triggered.

@donho 
I got these (potentially harmless) ERROR_INVALID_HANDLE during my plugin-loading debugging (test for the DockingManager "lost panel" problem: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13084#issuecomment-1789040331 ).

I do not think that this fix deserves its own separate issue report, but if you think otherwise, I will add also the the corresponding issue.(?)


donho:
@xomx 
Thank you for fixing it.

> I do not think that this fix deserves its own separate issue report

Agree, it's not necessary to open an issue for that.

PR #14359, opened 2023-11-16
Fix DELETE key not working in multi-editing

donho:
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14338#issuecomment-1813006657

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -532,6 +532,50 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
 						                                              // https://sourceforge.net/p/scintilla/bugs/2412/
 						break;
 
+					case VK_DELETE:
+					{
+						SHORT ctrl = GetKeyState(VK_CONTROL);
+						SHORT alt = GetKeyState(VK_MENU);
+						SHORT shift = GetKeyState(VK_SHIFT);
+						
+						if (!(shift & 0x8000) && !(ctrl & 0x8000) && !(alt & 0x8000)) // DEL + Column-edit
+						{
+							size_t nbSelections = execute(SCI_GETSELECTIONS);
+							if (nbSelections > 1)
+							{
+								execute(SCI_BEGINUNDOACTION);
+								for (size_t i = 0; i < nbSelections; ++i)
+								{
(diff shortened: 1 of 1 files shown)


donho:
@CennoxX @vinsworldcom
It should fix the issue found here https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14338#issuecomment-1813006657 by @CennoxX .

vinsworldcom:
It seems to fix the Del and Backspace issue for me!

You are a rock star!

Cheers.

CennoxX:
Hi @donho, this does work at the line end with <kbd>delete</kbd>, but it doesn't work for <kbd>backspace</kbd> at the line start, if I use <kbd>alt</kbd> and the curser to drag a multi cursor at the beginning of a line.

This only works, if I pressed enter before, or if the cursors were set with <kbd>ctrl</kbd> and individual clicks.
output2

donho:
@CennoxX 
Fixed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b24734df8ff1c0b1d4f033f3fb78089301a2e9bc (in master)
Could you check it?

CennoxX:
It's working.
Thank you for these excellent improvements to an already great tool!

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -532,6 +532,50 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
...
+									LRESULT posStart = execute(SCI_GETSELECTIONNSTART, i);
+									LRESULT posEnd = execute(SCI_GETSELECTIONNEND, i);
+									if (posStart != posEnd)
+									{
+										replaceTarget(L"", posStart, posEnd);
+									}
+									else // posStart == posEnd)
+									{
+										char eolStr[3];
+										Sci_TextRange tr;
```
chcg, line 558: @donho Shouldn't be Sci_TextRangeFull used instead of deprecated Sci_TextRange? Together with the corresponding SCI_GETTEXTRANGEFULL.

donho: @chcg 
You are absolutely right!
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/8ae049afbcfa4d37ec5abd81e2df8bf92af1b7d0

PR #14360, opened 2023-11-17
[xml] Update czech.xml to v8.5.9

xomx:


xomx:
@donho 
Now there is one duplicate message in the English master:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/24af65c6020757c8cf21e137906f6a6c5b40b6e3/PowerEditor/installer/nativeLang/english.xml#L1393
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/24af65c6020757c8cf21e137906f6a6c5b40b6e3/PowerEditor/installer/nativeLang/english.xml#L1496

donho:
@xomx 
Thank you for finding the redundant entry!
Fixed in https://github.com/notepad-plus-plus/notepad-plus-plus/commit/faba181e8fe7605dbff9ffc3d617f5306d9a1376.

PR #14373, opened 2023-11-19
Add search options output to Search-results

### LINKED ISSUE #14306: Search Results: display range & conditions
Quite often users are bewildered after executing a search and seeing the result(s).
[A recent example](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14304).

**Suggestion:**
 Display the range & conditions in the Search Results window.

***

Notepad++ v8.5.8.

alankilborn:
Fix #14306 purely for experimental evaluation.
Attention @Yaron10

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -82,7 +82,7 @@ struct FindOption
 	bool _isProjectPanel_2 = false;
 	bool _isProjectPanel_3 = false;
 	bool _dotMatchesNewline = false;
-	bool _isMatchLineNumber = true; // only for Find in Folder
+	bool _isMatchLineNumber = false; // always false for main search
 };
 
 //This class contains generic search functions as static functions for easy access
@@ -229,6 +229,7 @@ public:
 	FindInFinderDlg() {
 		_options._isMatchCase = false;
 		_options._isWholeWord = false;
+		_options._isMatchLineNumber = true;
 	};
 
 private:
@@ -251,7 +252,6 @@ public :
 	FindReplaceDlg() {
 		_uniFileName = new char[(_fileNameLenMax + 3) * 2];
 		_winVer = (NppParameters::getInstance()).getWinVersion();
-		_options._isMatchLineNumber = false;  // always false for main search
 		_env = &_options;
 	};
 
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -124,7 +124,7 @@ public:
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
-	void addSearchHitCountAndOptionsInfo(int count, int countSearched, bool searchedEntireNotSelection, const FindOption *pFindOpt);
+	void addSearchHitCountAndOptionsInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
 	const char* foundLine(FoundInfo fi, SearchResultMarkingLine mi, const TCHAR* foundline, size_t totalLineNumber);
 	void setFinderStyle();
 	void setFinderStyleForNpc(bool onlyColor = false);
@@ -135,7 +135,7 @@ public:
 	void copy();
 	void copyPathnames();
 	void beginNewFilesSearch();
-	void finishFilesSearch(int count, int searchedCount, bool searchedEntireNotSelection, const FindOption *pFindOpt);
+	void finishFilesSearch(int count, int searchedCount, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
 	
 	void gotoNextFoundResult(int direction);
 	std::pair<intptr_t, intptr_t> gotoFoundLine(size_t nOccurrence = 0); // value 0 means this argument is not used
@@ -336,7 +336,8 @@ public :
 
 	void finishFilesSearch(int count, int searchedCount, bool searchedEntireNotSelection)
 	{
-		_pFinder->finishFilesSearch(count, searchedCount, searchedEntireNotSelection, _env);
+		const bool isMatchLines = false;
+		_pFinder->finishFilesSearch(count, searchedCount, isMatchLines, searchedEntireNotSelection, _env);
 	}
 
 	void focusOnFinder() {
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -124,7 +124,7 @@ public:
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
-	void addSearchHitCountAndOptionsInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pfo);
+	void addSearchHitCountAndOptionsInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
 	const char* foundLine(FoundInfo fi, SearchResultMarkingLine mi, const TCHAR* foundline, size_t totalLineNumber);
 	void setFinderStyle();
 	void setFinderStyleForNpc(bool onlyColor = false);
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -82,7 +82,7 @@ struct FindOption
 	bool _isProjectPanel_2 = false;
 	bool _isProjectPanel_3 = false;
 	bool _dotMatchesNewline = false;
-	bool _isMatchLineNumber = false; // sometimes true for "Find in these search results"; never true for main search
+	bool _isMatchLineNumber = true; // only for Find in Folder
 };
 
 //This class contains generic search functions as static functions for easy access
@@ -124,7 +124,7 @@ public:
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
-	void addSearchHitCountAndOptionsInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
+	void addSearchResultInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
 	const char* foundLine(FoundInfo fi, SearchResultMarkingLine mi, const TCHAR* foundline, size_t totalLineNumber);
 	void setFinderStyle();
 	void setFinderStyleForNpc(bool onlyColor = false);
@@ -229,7 +229,6 @@ public:
 	FindInFinderDlg() {
 		_options._isMatchCase = false;
 		_options._isWholeWord = false;
-		_options._isMatchLineNumber = true;
 	};
 
 private:
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -124,7 +124,7 @@ public:
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
-	void addSearchResultInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
+	void addSearchResultInfo(int count, int countSearched, bool searchedEntireNotSelection, const FindOption *pFindOpt);
 	const char* foundLine(FoundInfo fi, SearchResultMarkingLine mi, const TCHAR* foundline, size_t totalLineNumber);
 	void setFinderStyle();
 	void setFinderStyleForNpc(bool onlyColor = false);
@@ -135,7 +135,7 @@ public:
 	void copy();
 	void copyPathnames();
 	void beginNewFilesSearch();
-	void finishFilesSearch(int count, int searchedCount, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
+	void finishFilesSearch(int count, int searchedCount, bool searchedEntireNotSelection, const FindOption *pFindOpt);
 	
 	void gotoNextFoundResult(int direction);
 	std::pair<intptr_t, intptr_t> gotoFoundLine(size_t nOccurrence = 0); // value 0 means this argument is not used
@@ -335,8 +335,7 @@ public :
 
 	void finishFilesSearch(int count, int searchedCount, bool searchedEntireNotSelection)
 	{
-		const bool isMatchLines = false;
-		_pFinder->finishFilesSearch(count, searchedCount, isMatchLines, searchedEntireNotSelection, _env);
+		_pFinder->finishFilesSearch(count, searchedCount, searchedEntireNotSelection, _env);
 	}
 
 	void focusOnFinder() {
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
@@ -124,7 +124,7 @@ public:
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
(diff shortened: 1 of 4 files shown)


alankilborn:
Example:

image

Note `[cwn]` and `[ce]` in the search output, meaning that the first search was done in Extended mode with case sensitivity, and the second search was done in Normal mode with whole-word being added to case sensitivity for the search.

Yaron10:
@alankilborn,

Very nice and useful. 👍 
Thank you for that.

1. The single letters are too cryptic IMO.
I'd prefer `[Normal/Extended/Reg, Word, Case]`.
2. How about adding commas?
3. `Word` appear before `Case` in the Find dialog (I should have noted that in "Multi-select" as well).
4. You've decided that `Reg.` is not necessary, correct?
5. I understand the `Search "foo" [cwn] (2 hits in 1 file of 1 searched)` structure.
\- Still, `Search "foo" (2 hits in 1 file of 1 searched) [cwn]` might be better for two reasons:
A. The old flow users are used to is kept. There's some addition at the end.
B. The number of hits should be more important in most cases.

alankilborn:
Regarding 5:  I actually had it that way first, then I realized my original proposal was different, so I changed it.  But "at the end" really does make more sense; I'll change it.

Regarding 1:  If the new text is at the end, it can be longer.  So I'll move away from single letters.

Regarding 4:  In my quick mockup, I forgot about `. matches newline`.  I'll include it next try.

Regarding 3:  Ordering isn't that important, but we can do `W` before `C`.

Regarding 2:  No commas. :-)

alankilborn:
image

Afterthought: `Norm` -> `Normal`

pryrt:
I like the added context/information provided by those -- coming back some time later to view previous search results, one might not remember which options were active at the time, so seeing them enumerated in the results could be quite helpful (especially, as in the screenshot above, the search text is the same, but the different options then find different numbers of matches).

Yaron10:
@alankilborn,

> No commas. :-)

Anything against them? :)

It's great now. 👍 

Allow me another minor comment:
Shouldn't the search-mode be first? - You could also shorten the code and use `TEXT("/Word")` & `TEXT("/Case")` (or `TEXT(" ,Word")`).

Thank you again.

vinsworldcom:
This is magnificent!  I'll echo @pryrt comment above - the context of the search when going back to look at it really helps - especially if I change the text and then search again - knowing the context allows me to redo the search with the same options (case, regex, word, etc) to truly see what I changed.

Cheers.

alankilborn:
>Shouldn't the search-mode be first?

Word and case appear _first_ in the _Find_ window.  :-)

It's all a judgement call, but I think for purposes of a demo, it suffices as is.  If there's any interest from the author, he'll decide where to take it.

Yaron10:
> I understand the Search "foo" [cwn] (2 hits in 1 file of 1 searched) structure.

The search mode & parameters should _in principle_ be displayed before the hits.
This is the order: you set the mode & parameters and then get the hits/results.
I still thought it might be better to change the structure. 

> Word and case appear first in the Find window.

That's correct.
And yet, the search mode is more "basic" when displaying the _Search Info_.

(`addSearchHitCount()` to `addSearchInfo()`? :) ).

As you understand.

alankilborn:
>The search mode & parameters should in principle be displayed before the hits.

I like it as the rightmost data.  YOU talked me into liking it.  :-)

>addSearchHitCount() to addSearchInfo()?

Again, just a demo.  No more effort to be put in unless there is a real reason to.
Another example or more effort:  It would need to be "localized".

Yaron10:
@alankilborn,

> I like it as the rightmost data. YOU talked me into liking it. :-)

And I'm glad you agreed.

I meant that "in principle" and consistency are important, but there are other factors as well.
So it's true that Word & Case are before the search mode in the dialog, but we can flexible here as well.

Anyway, it was a minor comment.
You did a great job! Thanks again.

donho:
@alankilborn 
I have to admit, it looks promissing :)
I will finish the release v8.6, then I will be back for the code review in few days.

Yaron10:
@alankilborn,

Some ideas for later.

**1.** "Search only in found lines" is also an option. - You can display it in the same section.

תמונה

**2.** Find in Files: what do you think of adding "Sub-Folders" & "Hidden Folders"?

Yaron10:
"Sub Folders" & "Hidden Folders" are search-expanding as opposed to the search-restricting Word & Case.
`!Sub` & `!Hidden` might be better. It requires some thinking.

donho:
@alankilborn 
> I think so. See now how it was different from that other issue? :-)

I still think it's a Nice-To-Have but not Must-To-Have feature comparing the tip on the status bar for the countless PEBCEK. 
(the tip on the status bar will prevent from PEBCEK in the future, whereas the tip for Find in Files, there is no request for that)
But the result looks clean and so does the code, so why not.

> The search mode & parameters should in principle be displayed before the hits.

I agree with @Yaron10. I would suggest the following formats:

* [Normal: Case/Word]
* [Normal]
* [Extended: Case]
* [RegEx.: Case/Word]
* [RegEx]


@Yaron10 
> 1. "Search only in found lines" is also an option. - You can display it in the same section.
> 2. Find in Files: what do you think of adding "Sub-Folders" & "Hidden Folders"?

> "Sub Folders" & "Hidden Folders" are search-expanding as opposed to the search-restricting Word & Case.
!Sub & !Hidden might be better. It requires some thinking.

Thank you for the cautious thinking. But for keeping the result clean, the above info won't be considered.

alankilborn:
>But the result looks clean and so does the code, so why not.

So, OK then!  :-)

I'll make the requested changes and add localization.

alankilborn:
Here's a sample of the evolving output:

image

IMO, the output format at https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373#issuecomment-1819150606 is better.  I like how the number of hits (and, everything, really) align better vertically.  Plus, these new parts of the output aren't something I need to check _every_ time (just when I'm unsure what I did), and for that, if they are on the far right, they are mostly out of sight and out of mind (but they are there, for security).

What do other think?

alankilborn:
I'm so confused / misunderstood.
I await @donho telling me what is wrong with my PR.
:-(

Yaron10:
@alankilborn,

Your PR is great.
We've had a chain of misunderstandings.

alankilborn:
Perhaps there is merit to @Yaron10's idea https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373#issuecomment-1826296758

STR:

* do a _Find All_ such that _Search results_ gets data
* right-click in the white area of _Search results_
* choose _Find in these search results..._
* checkmark _Search only in found lines_ and _Match case_
* click _Find All_

This resulting text is longish and hard to understand what it means:

image

(Tangent: Note that title bar says `1 hit` but there were really 2 hits; already open issue for this somewhere)

Maybe it could be changed into:

image

I've abbreviated it as much as possible, given its current location.  If the whole `[ ... ]` field is moved to be the right-most on the line, probably don't need to be so concerned about bfrevity...

donho:
@alankilborn 
> YES, exactly that statement. I like the previous one too:
> 
> Search "string" (hits) [configuration]

+1.

I don't think there's a misunderstanding between us.

alankilborn:
OK, I will move the `[ configuration ]` to the end in the next commit.

---

>...misunderstanding...

Sometimes I think things are discussed in these comments _too_ much; I am guilty of this as much as anyone.
The problem is, things can get misinterpreted.
If we were all sitting together around a table at the Notepad++ Headquarters building, discussing things, there'd be less confusion potential.  :-)

alankilborn:
The commit https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373/commits/90754b58dd6cf675146dcd6f7b68f1d3450cc8d5 puts the new data on the far right, as requested once the confusion died down. :-)

alankilborn:
The commit https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373/commits/6675a65d6dc70d13011659eb5741d3b63b57a670 eliminates the `isMatchLines` argument from multiple functions; this argument is no longer needed since the `pFindOpt` argument was added (for the needs of this issue); the same data as the old argument can now be found in a member variable of the new argument.

alankilborn:
The commit https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373/commits/69f5ede967e84b6ff6d06d1f05916a0ff48c8cee implements "In Found Lines" as described https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373#issuecomment-1833874842 .

Not sure if this will be wanted, but easy to revert if not.

Yaron10:
@alankilborn,

👍 

Just curious: is `_env->_isMatchLineNumber = false; // always false for main search` needed?

Thank you.

alankilborn:
> is `_env->_isMatchLineNumber = false; // always false for main search` needed?

Maybe not now (but maybe it is good emphasis? :-) )...
Originally `_isMatchLineNumber` was intialized to `true` in the structure initializer.
One of my changes for that was the line you are questioning.
Later I decided to set `_isMatchLineNumber` to `false` in the structure initializer.

Yaron10:

alankilborn:
It probably also makes it very explicit that when a search macro runs, it doesn't affect the search -- there are no search commands defined to date that that member would be true for.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -82,7 +82,7 @@ struct FindOption
 	bool _isProjectPanel_2 = false;
 	bool _isProjectPanel_3 = false;
 	bool _dotMatchesNewline = false;
-	bool _isMatchLineNumber = true; // only for Find in Folder
+	bool _isMatchLineNumber = false; // sometimes true for "Find in these search results"; never true for main search
```
donho, line 85: Restore it please.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -229,6 +229,7 @@ class FindInFinderDlg : public StaticDialog
 	FindInFinderDlg() {
 		_options._isMatchCase = false;
 		_options._isWholeWord = false;
+		_options._isMatchLineNumber = true;
```
donho, line 232: Remove it please

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -124,7 +124,7 @@ friend class FindReplaceDlg;
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
-	void addSearchHitCount(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection);
+	void addSearchHitCountAndOptionsInfo(int count, int countSearched, bool searchedEntireNotSelection, const FindOption *pFindOpt);
```
donho, line 127: Please rename `addSearchHitCountAndOptionsInfo` to `addSearchResultInfo`.

donho (changes requested):
The scope of the FR of this PR to me is provide user "What we search", so the filter won't be in our target [Normal/Expanded/RegEx: Word/Case/Wrap] (as "In Selection" is not in the target as well).
So in the case of "Search Only In Found Lines" is checked, let's do the following:
```
Search "AppData" (64 hits in 2 files of 2 searched)  [Normal: Word/Case]  - Line Filter Mode: only display the filtered results
```

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
alankilborn: Done.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
alankilborn: Done.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
alankilborn: Done.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
```diff
@@ -124,7 +124,7 @@ friend class FindReplaceDlg;
 	void addSearchLine(const TCHAR *searchName);
 	void addFileNameTitle(const TCHAR * fileName);
 	void addFileHitCount(int count);
-	void addSearchHitCount(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection);
+	void addSearchResultInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);
```
donho, line 127: I see you have added back `bool isMatchLines`. I didn't do any request to add it back. Contrarily, I think it's good to remove it, since the value of "isMatchLines" can be found in **pFindOpt**.

donho (changes requested):
1. Please change the signature of method from:
`void addSearchResultInfo(int count, int countSearched, bool isMatchLines, bool searchedEntireNotSelection, const FindOption *pFindOpt);`
to:
`void addSearchResultInfo(int count, int countSearched, bool searchedEntireNotSelection, const FindOption *pFindOpt);`
2. Please apply the same thing for `finishFilesSearch` method.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.h
alankilborn: >I see you have added back bool isMatchLines. I didn't do any request to add it back.

You requested it indirectly with the other changes you wanted me to put back.

---

>good to remove it, since the value of "isMatchLines" can be found in pFindOpt.

Yes, I agree.  Which was why I did it in the earlier state of the PR.

But...someone at some earlier point decided to do make this variable `true` instead of `false`:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/bf37d4714f572efa13088863ef6e8ef0132c4021/PowerEditor/src/ScintillaComponent/FindReplaceDlg.h#L85

which interferes with a main search being able to use the `pFindOpt` member (instead of the `isMatchLines` variable) without further changes.

So, several of the other changes I did in my earlier PR became necessary to compensate for this.
And you had previously wanted me to restore those.

So I'm at a quandary... now I can put _some_ of those changes back in and you won't disapprove, in order to get rid of passing `isMatchLines` as a separate parameter?

(Note that I'm NOT talking about the "line filter" output stuff -- that's been put back)

donho: > But...someone at some earlier point decided to do make this variable true instead of false:

You can use "YOU" instead of "someone", can't you? Otherwise, I suggest you to use "Your Majesty" :)

The reason of the member `_isMatchLineNumber` being set to true, because it's used only in "Find in Finder" operation. So it won't influence the other operations. I don't think it's necessary to change this logic. Please let me know if I'm wrong.

alankilborn:
>You can use "YOU" instead of "someone", can't you? 

Well, I haven't been around long enough to know when Find-in-Finder was originally adopted.
So it could have been a contributor's PR and thus not YOU.
But you would have to have accepted it into master, so at that point all the code becomes YOURS.  :-)
So, OK, I could have said YOU.  :-)

---

>I suggest you to use "Your Majesty" :)

I will try to remember to do this.  :-P
:-)

---

>I don't think it's necessary to change this logic. Please let me know if I'm wrong.

OK, I will take another look at it.  But when I was doing it the first time, I only made what I felt were necessary changes...

[ Except for the "line filter" part, which I liked, but kept very separate for easy reversion, because I knew Your Majesty (LOL) wouldn't like it.  :-) ]

alankilborn:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373/commits/4ff53930cab7ecb5bc27851126a1d351af0b5195 removes the `isMatchLines` argument to mutliple functions.  This results in "line filter mode" showing for a main search, something that shouldn't happen:

image

I'll next work on an alternate method (from the one proposed earlier) to fix that...

alankilborn:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373/commits/c5848672c550098587631805f998b93fd1ae5e29 is really just another way of doing what was tried (and disliked) before.

I don't think it is better; what I think is better is setting `._isMatchLineNumber` to `false` in the `FindOption` structure instantiation.

Note that setting `_env->isMatchLineNumber` to `false` is needed so that macros that output to _Search results_ don't output the "line filter mode" text.  Again, this would not strictly be needed (see note [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14373#issuecomment-1836792061)) if the structure instantiation used `false`.

donho:
OK.
It seems make more sense.
Please restore:
`bool _isMatchLineNumber = false; // always false for main search`
if you want.

alankilborn:
>bool _isMatchLineNumber = false;

Done.  Ready for further review.

donho (approved):

alankilborn:
Thanks, Don, er, Y.M.
I'm glad we got this done.
It's a nice improvement to search output.

donho:
@alankilborn 
> It's a nice improvement to search output.

Yes, indeed :)
Thank you for your PR.

Issue #11520, opened 2022-04-11
RTL accidentally enables font ligatures


rdipardo:
### Description of the Issue

Switching to RTL mode enables ligatures (e.g. != -> &#x2260;) if the active font supports them (e.g. Cascadia Code).

Ligatures normally require the deliberate activation of DirectWrite at Settings > Preferences > MISC. > "Use DirectWrite . . . ".  As noted [here][0], RTL is not available in DirectWrite mode. This was true before a16f084 and remains true now. Yet RTL appears to temporarily enable ligatures even when DirectWrite is off. Switching back to LTR restores non-ligatured (GDI) rendering.

### Steps to Reproduce the Issue

1. Make sure DirectWrite is not selected at Settings > Preferences > MISC. > "Use DirectWrite . . . "
2. At Settings > "Style Configurator", set "Font name" to Cascadia Code (you may need to [install the font] first)
3. Paste the following into a new buffer:

~~~
2 >= 1
false != true
~~~

4. Select View > Text Direction RTL, or press <kbd>Ctrl + Alt + R</kbd>
5. The operators become ligatures

### Expected Behavior

RTL mode should not change how the device context renders fonts.

### Actual Behavior

RTL mode triggers DirectWrite rendering when it should be disabled.

#### 8.3.3 release build

![npp_833_rel](https://user-images.githubusercontent.com/59004801/162649265-16d95352-e830-440a-bb02-5423d7ccbdea.gif)

#### Git (721f994d)

![npp_833_git](https://user-images.githubusercontent.com/59004801/162649270-cd6b6467-73fd-4fe2-9ecd-f19eed3e9487.gif)

*NOTE*  
The horizontal inversion of `1` and `2` is a separate [known issue][1] that also predates a16f084

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11476#issuecomment-1092681783
[1]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11476#issuecomment-1094104174
[install the font]: https://github.com/microsoft/cascadia-code/releases

rdipardo:
Don't look now, but this bug may be the price of fixing #11476 &#x2014; the commit before a16f084 doesn't reproduce it!

I guess the creaky old GDI implementation will be good enough until [Notepad2] figures out a stable bidirectional API and ports it upstream.

[Notepad2]: https://github.com/zufuliu/notepad2/issues/392

Yaron10:
[...]

PR #14374, opened 2023-11-19
Prevent Direct Write being set if user uses RTL

donho:
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/975d29b30bfd624b6392541b133c2bd32f3212b0#commitcomment-132946815

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -5,7 +5,7 @@ Translation note:
 2. All the comments are for explanation, they are not for translation.
 -->
 <NotepadPlus>
-	<Native-Langue name="English" filename="english.xml" version="8.5.9">
+	<Native-Langue name="English" filename="english.xml" version="8.6">
 		<Menu>
 			<Main>
 				<!-- Main Menu Entries -->
@@ -1505,6 +1505,7 @@ NOTE: Choosing not to create the placeholders or closing them later, your manual
 Would you like to create those placeholders?
 
 NOTE: Choosing not to create the placeholders or closing them later, your session WILL BE MODIFIED ON EXIT! We suggest you backup your &quot;session.xml&quot; now."/>
+			<RTLvsDirectWrite title="Cannot run RTL" message="RTL is not compatible with DirectWrite mode. Please disable DirectWrite mode in MISC. section of Preferences dialog, restart Notepad++, and try this command again."/>
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Clipboard History"/>
--- a/PowerEditor/installer/nativeLang/english_customizable.xml
+++ b/PowerEditor/installer/nativeLang/english_customizable.xml
@@ -5,7 +5,7 @@ Translation note:
 2. All the comments are for explanation, they are not for translation.
 -->
 <NotepadPlus>
-	<Native-Langue name="English" filename="english_customizable.xml" version="8.5.9">
+	<Native-Langue name="English" filename="english_customizable.xml" version="8.6">
 		<Menu>
 			<Main>
 				<!-- Main Menu Entries -->
@@ -1505,6 +1505,7 @@ NOTE: Choosing not to create the placeholders or closing them later, your manual
 Would you like to create those placeholders?
 
 NOTE: Choosing not to create the placeholders or closing them later, your session WILL BE MODIFIED ON EXIT! We suggest you backup your &quot;session.xml&quot; now."/>
+			<RTLvsDirectWrite title="Cannot run RTL" message="RTL is not compatible with DirectWrite mode. Please disable DirectWrite mode in MISC. section of Preferences dialog, restart Notepad++, and try this command again."/>
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Clipboard History"/>
--- a/PowerEditor/installer/nativeLang/french.xml
+++ b/PowerEditor/installer/nativeLang/french.xml
@@ -5,7 +5,7 @@ Translation note:
 2. All the comments are for explanation, they are not for translation.
 -->
 <NotepadPlus>
-	<Native-Langue name="français" filename="french.xml" version="8.5.9">
+	<Native-Langue name="français" filename="french.xml" version="8.6">
 		<Menu>
 			<Main>
 				<!-- Main Menu Entries -->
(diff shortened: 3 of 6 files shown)


donho:
@Yaron10
Please check this PR

Yaron10:
@donho,

1. If you use NPP RTL, the text-direction is RTL on startup (off-topic, this should be configurable IMO).
If you now use "Text Direction RTL", you should get the beep.
Your new block should be under `if (toRTL && !isRTL)`.
2. Even if you fix the previous point - It would be weird that on startup "Text Direction RTL" is OK, but after "Direction **LTR**" & "Direction **RTL**", you get the message.
3. How would you use DirectWrite in NPP RTL? - You disable it on startup.

rdipardo:
@Yaron10 
> 3. How would you use DirectWrite in NPP RTL?

Answer: select `Cascadia Code` or another ligatured font, then enter RTL mode: #11520

Yaron10:
@rdipardo,

Thank you for reminding me of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11520. :)

donho:
@Yaron10 
Just commit a new modification. It should be more coherent.

Yaron10:
@donho,

I suppose that `if (!isWINE && !isTextDirectionRTL())` should be `if (isWINE || isTextDirectionRTL())`.

But we still have a problem: you can't use DirectWrite in NPP RTL.
_And NPP RTL users might want to test DirectWrite_.

How about setting the default to DirectWrite only if !NPP RTL?
if NPP RTL - you can use the (slightly modified) `RTLvsDirectWrite` message on checking `Use DirectWrite` in Preferences.
```
"Text Direction RTL" is not compatible with Direct Write mode.
Continue?
```

donho:
@Yaron10 
> I suppose that if (!isWINE && !isTextDirectionRTL()) should be if (isWINE || isTextDirectionRTL()).

Thank you for correct me the inattentional error.

> How about setting the default to DirectWrite only if !NPP RTL?

What do you mean by "!NPP RTL"?
And "NPP RTL users" ?

Yaron10:
@donho,

I understand the confusion now.

> If RTL localization is not used, but user tries to set Scintilla RTL via menu command "Text Direction RTL", then it's still "NPP RTL" case.

**NPP RTL:** `NppParameters::getInstance().getNativeLangSpeaker()->isRTL()` is `true`.
**Text Direction RTL:** `isTextDirectionRTL()` is `true`.

"Text Direction LTR/RTL" is available whether the localization is LTR or RTL.

If RTL localization is used (**NPP RTL**), how would you use DirectWrite?
I'd like to test it with "**Text Direction LTR**".

donho:
>How about setting the default to DirectWrite only if !NPP RTL?
if NPP RTL - you can use the (slightly modified) RTLvsDirectWrite message on checking Use DirectWrite in Preferences.
```
Text Direction RTL" is not compatible with Direct Write mode.
Continue?
```

@Yaron10 
OK, I think I understand your intention, but I don't see the advantage.
A new user installs Notepad++, the Direct Write Technology is set, and he/she wants to make Scintilla RTL via menu command.
With your suggestion, the following scenario will happen:
1. A Message Box ask the above question, and user clicks Yes.
2. RTL doesn't work.
3. User complains

donho:
@Yaron10 
> If RTL localization is used (NPP RTL), how would you use DirectWrite?

If RTL localization is used, according current PR state, isTextDirectionRTL() will be true. So _writeTechnologyEngine will be affected to the default one - Direct Write won't be set.

Therefore RTL localization will work as expected.

Yaron10:
@donho,

> Therefore RTL localization will work as expected.

That's correct.
But I'd still like to test DirectWrite with "Text Direction **LTR**". And I can't because on startup the direction is **RTL** and you disable it.

```
Text Direction RTL" is not compatible with Direct Write mode.
Continue?
```

I didn't mean this should replace the message you've added in `if (toRTL &&`. That's good whether the localization is LTR or RTL. 
I suggested to _add_ a message on checking `Use DirectWrite` in Preferences if NPP RTL.

> How about setting the default to DirectWrite only if !NPP RTL?

With a new portable NPP, the default interface language is English.
It's getting complex. :)

Yaron10:
@donho,

> OK I see what's your situation (and it can happen to other RTL users) now.

I'd just like to emphasize that `Use DirectWrite` is useless/has-no-effect (if NPP RTL) in the current state of this PR.

***

> How about setting the default to DirectWrite only if !NPP RTL?

In v8.6 the default value of `_writeTechnologyEngine` is `defaultTechnology`. The user can handle that as they wish.
My initial thought was as follows:
If you think it's important to change the default value to `directWriteTechnology`, do it only if NPP layout is LTR.
If NPP RTL, the default value should remain `defaultTechnology`.

But I then realized that in a new setup, the default language is English (-> DirectWrite).
What should happen when the user is switching to Hebrew?  

***

> So the GUI will be RTL, but Scintilla will be LTR.
If editZoneRTL is absent, its default value is "no". What do you think?

1. Do you mean changing the RTL XML localization files?
2. And if the user wants the default text-direction to be RTL (as it is in v8.6)?

***

I don't like it, but you can hide `Use DirectWrite` if NPP layout is RTL.

Yaron10:
Another option.

Add a new setting: `Default Text Direction:` LTR/RTL.
(This can be very useful regardless of DirectWrite: I use NPP RTL, but I want the default text-direction to be LTR.).

Hide `Use DirectWrite` if the new preference is RTL.
Obviously this would require some more thinking & work.

donho:
@Yaron10 
> I'd just like to emphasize that Use DirectWrite is useless/has-no-effect (if NPP RTL) in the current state of this PR.

True. That's why I suggest to make Notepad++ read `editZoneRTL` in the future version. Once it is implemented, the current state of PR can work without changing.


> Do you mean changing the RTL XML localization files?
> And if the user wants the default text-direction to be RTL (as it is in v8.6)?

I might not be clear enough.

In official distribution, all the xml will be kept as they are now.
But user can add `editZoneRTL="no"` in hebrew.xml or any xml for having GUI RTL but Scintilla LTR.
From
```xml
<Native-Langue name="Hebrew" RTL="yes" filename="hebrew.xml">
```
to
```xml
<Native-Langue name="Hebrew" RTL="yes" editZoneRTL="no" filename="hebrew.xml">
```

So in Notepad++ v8.6.1 binary, it finds `RTL="yes"`, then it will try to find `editZoneRTL="???"`. If `editZoneRTL` attribute is absent, then Scintilla follow the `RTL` attribute (normally YES). Otherwise, if `editZoneRTL` value is no, then Scintilla is LTR, even the GUI is RTL.

> Add a new setting: Default Text Direction: LTR/RTL.

I don't like adding options. I guess this policy is clear for the most of community people. 

> Obviously this would require some more thinking & work.

Yes, indeed. Let's think & talk about it for v8.6.1.
But in v8.6, I think Direct Write deserves to be set as default value, at least to try. The old value can always be set to back in the next version.

Yaron10:
@donho,

> But user can add editZoneRTL="no" in hebrew.xml or any xml for having GUI RTL but Scintilla LTR.

It would be easier to switch to English IMO.

> I don't like adding options. I guess this policy is clear for the most of community people.

Nobody is forcing you to do anything!

PR #14379, opened 2023-11-20
Updated indexes according to comments and improved cross-referencing

### LINKED ISSUE #14378: Incorrect Localization Offsets
### Description of the Issue
While working on a fork of Notepad++, I ran across the comments in `BOOL Notepad_plus::notify(SCNotification *notification)` near line 532. In my fork, I plan to update `itemUnitArray` - so I went over to localization.cpp to review the mappings. Much to my surprise, they are no longer lined up.

The offsets in tabCmSubMenuEntryPos from localization.cpp should point to valid rows in itemUnitArray. Currently, they do not. I don't use localization, so I'm not entirely sure how this manifests in the UI...

The old indexes map 5 commands - of which only close multiple tabs appears to work properly.
0. "Close Multiple Tabs" - offset 1 - seems OK
1. "Open into" - offset 4
2. "Copy to Clipboard" - offset 13
3. "Move Document" - offset 14
4. "Apply Color to Tab" - offset 15

### Expected Behavior
I expected these offsets to actually match the observed indexes.

### Actual Behavior
The given indexes would map as follows:
1. Open into -> Close All Unchanged
2. Copy to Clipboard -> Move to Recycle Bin
3. Move Document -> Reload
4. Apply Color to Tab -> Print

### Debug Information
Notepad++ v8.5.8   (64-bit)
Build time : Nov 19 2023 - 18:23:39
Path : C:\Source\notepad-plus-plus\PowerEditor\visual.net\x64\Debug\Notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.2715
Current ANSI codepage : 1252
Plugins : none

Notepad_plus.cpp tells you to keep location.cpp matched up...
image

But localization.cpp doesn't currently match!
image

wbic16:
These offsets don't seem to match `Notepad_plus` in the latest code. I re-mapped them to match the current code.

I haven't worked with Notepad++ much, so I'm not sure if this is a real problem or not. I'm not entirely certain I mapped the old entries properly (it seems like someone expanded them without updating localization perhaps?).

This fixes #14378.

### DIFF
--- a/PowerEditor/src/localization.cpp
+++ b/PowerEditor/src/localization.cpp
@@ -418,7 +418,7 @@ void NativeLangSpeaker::changeMenuLang(HMENU menuHandle)
 	}
 }
 
-
+// Important: these indexes must match itemUnitArray in `Notepad_plus::notify(SCNotification* notification)`.
 static const int tabCmSubMenuEntryPos[] =
 {
 //   +-------------- The submenu entry item position on the top level of tab context menu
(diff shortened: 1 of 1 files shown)


donho:
There's no wrong on the current code:

image

wbic16:
Thanks for the screenshot. Makes sense now.

PR #14384, opened 2023-11-21
Fix DEL regression

donho:
Fix https://community.notepad-plus-plus.org/topic/25150/notepad-v8-6-release-candidate/5?_=1700578401629


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c62e2229850c3b111b555b0448be5c9f4e38be03

PR #14404, opened 2023-11-24
[xml] Update danish.xml to v8.6

Regmos:


donho:
Create a new branch for each PR. Make sure your branch name wasn't used before - you can add date (for example patch3_20200528) to ensure its uniqueness.

Issue #14323, opened 2023-11-06
open link when double click include ' of end tag in html


duong273:
open link when double click include ' of end tag in html
### Description of the Issue

I have same code in html:
<li class='menu-item-has-children menu-item-type-label menu-item' data-class='' data-enable_mega='' data-icon=''>
        <a data-href-default='#' data-text-default='' href='/#Label2'>Quà tặng Doanh Nghiệp</a>
        <ul class='sub-menu'>
          <li class='menu-item-type-post menu-item-type-label menu-item' data-class='menu-item-type-post' data-enable_mega='' data-icon=''>
            <a data-href-default='/' data-text-default='' href='https://gifthouse.jwvietnam.vn/p/gioi-thieu.html'>Giới thiệu JW Gifthouse</a></li>
          <li class='menu-item-type-post menu-item-type-label menu-item' data-class='menu-item-type-post' data-enable_mega='' data-icon=''>
            <a data-href-default='/' data-text-default='' href='https://luxgift.jwvietnam.vn/p/gioi-thieu.html'>Giới thiệu JW Luxgift</a></li>
          <li class='menu-item-type-post menu-item-type-label menu-item' data-class='menu-item-type-post' data-enable_mega='' data-icon=''>
            <a data-href-default='/' data-text-default='' href='https://gift.jwvietnam.vn/p/gioi-thieu.html'>Giới thiệu JW Gift</a></li>
        </ul>
      </li>
But when double click the link it will open with the ' at end of the link
### Steps to Reproduce the Issue

1. Copy the code as same or write your html code with link in the apostrophe. The link will have hyperlink with underscore but it include the apostrophe at the end.
2. Click on that link will open in default browser the link have apostrophe at the end of link.

### Expected Behavior

Open the webpage correct
### Actual Behavior

It open wrong webpages or have 404 error page
### Debug Information

Notepad++ v8.5.4   (64-bit)
Build time : Jun 17 2023 - 20:42:45
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Temp\Xiaomi_Global_Debloater_v1.9\Debloat-16-Duo.bat" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 Home Single Language (64-bit)
OS Version : 22H2
OS Build : 22621.2428
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
[...]

PR #14408, opened 2023-11-25
add apostrophe to URL check

nampng:
Fix for #14323 for the indicator not terminating when an apostrophe is at the end of a URL.


Yaron10:
@nampng,

An apostrophe _in the middle_ should be part of the URL.
https://somedomain's.com

Thank you for your work.

Issue #11437, opened 2022-03-25
Copy/Cut state is not updated on switching between views


Yaron10:
**STR:**
Open two files: _A_ and _B_.
Make sure _A_ has no selection.
Move _B_ to the other view.
Select some text in _B_ 
Click on _A_'s tab.

**Result:**
Copy/Cut/Delete and "Convert Case to" are enabled.

***
Notepad++ v8.3.3   (64-bit)
Build time : Mar 15 2022 - 00:15:25
Path : C:\---\Notepad++ Original.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit) 
OS Version : 2009
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : ComparePlus.dll HTMLTag.dll MenuIcons.dll PythonScript.dll _CustomizeToolbar.dll

Yaron10:
Undo/Redo state is not updated either.
Doc state is properly updated.

PR #14415, opened 2023-11-27
Enable Copy & Cut commands on menu all the time

### LINKED ISSUE #14401: line operation (delete, copy & cut) problem
v8.6, 2 Problem

1.When text is not selected, the Cut/Copy menu and toolbar icons are disabled and cannot be used.
The Cut/Copy menu and toolbar icons should always be active for line operation.
Ultraedit 'Enable copy/cut of current line with no selection active' option work like this.
line

2.Shift + Delete: without selected text, it shoud 'Cut'(not 'Delete') the whole line.
This is the standard and UltraEdit works like this too.

donho:
Before the implementation of https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e9c50ed967fe44c758cd25c3907ef8aee37e6f05, Copy & Cut command were disabled when there were no selection, which makes sense because there was nothing to be copied & cut. Now users can copy/cut current line without selection.

Fix #14401

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2561,11 +2561,8 @@ void Notepad_plus::checkClipboard()
 {
 	bool hasSelection = _pEditView->hasSelection();
 	bool canPaste = (_pEditView->execute(SCI_CANPASTE) != 0);
-	enableCommand(IDM_EDIT_CUT, hasSelection, MENU | TOOLBAR);
-	enableCommand(IDM_EDIT_COPY, hasSelection, MENU | TOOLBAR);
 
 	enableCommand(IDM_EDIT_PASTE, canPaste, MENU | TOOLBAR);
-	enableCommand(IDM_EDIT_DELETE, hasSelection, MENU | TOOLBAR);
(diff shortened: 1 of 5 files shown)


donho:
@Yaron10 & @alankilborn 
I hope this PR will make you guys happier :)
I'm happy with it anyway.

alankilborn:
Unintended consequence (but in a good way):  
Seems to partially solve: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11437

Yaron10:
The menu commands & their shortcuts are not "synced".
Ctrl+C & Ctrl+X do not copy/cut the current line if there's no selection.

donho:
@Yaron10 
> The menu commands & their shortcuts are not "synced".
> Ctrl+C & Ctrl+X do not copy/cut the current line if there's no selection.


I've forgot Ctrl-C and Ctrl-X are managed directly by Scintilla. I should have tested it.
Now it's fixed. Thank you!

Yaron10:
If you replace Ctrl+X with (e.g.) Ctrl+Shift+C, the command and shortcut are not synced.

PR #14431, opened 2023-11-30
Make RTL per document & remembered across the sessions

### LINKED ISSUE #9665: [Suggestion] Add remember Text Direction support
Hello. I want to suggest adding support to remember the set option of "Text Direction RTL" or "Text Direction LTR" so it can be used on next launch just like Word Wrap is remembered. Also, if it is possible, it would be nice if Notepad++ could detect document languages and auto switch between RTL and LTR.

Thanks for your time.

### LINKED ISSUE #9950: Text direction per tab
When you switch the text direction via **View > Text Direction LTR / RTL**, or by **Ctrl Alt L / R**, the text direction changes for all the tabs instead of changing only for the active tab.

I think it would be better if it worked tab-wise, because if I have tabs with multiple languages open, I would want each of them to be displayed in a certain direction.

### LINKED ISSUE #14385: Add edit zone LTR ability for RTL localization files
For all localization files for the languages Right to Left like Arabic & Hebrew has an attribute "RTL = yes". That makes Notepad++ GUI RTL, but also edit zone (Scintilla) RTL.
It will be good that Notepad++can load RTL language with its GUI RTL, but keep its edit zone as LTR.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14374#issuecomment-1819972587

donho:
This PR add the ability of RTL per document (tab), and memorizes RTL setting of each file in the session.xml.

It also adds an attribute "editZoneRTL" in localization XML file. if "RTL" attribute is present and its value is "yes", then user can add  "editZoneRTL" attribute beside, and set the value of the attribute in question to "no", so Notepad++ GUI will be RTL, but Scintilla zone will be LTR by opening files.  Of course, user can set any direction they want afterward, and what they have set on document will be remembered across the sessions.

Fix #9665, fix #9950, fix #14385

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1505,7 +1505,8 @@ NOTE: Choosing not to create the placeholders or closing them later, your manual
 Would you like to create those placeholders?
 
 NOTE: Choosing not to create the placeholders or closing them later, your session WILL BE MODIFIED ON EXIT! We suggest you backup your &quot;session.xml&quot; now."/>
-			<RTLvsDirectWrite title="Cannot run RTL" message="RTL is not compatible with DirectWrite mode. Please disable DirectWrite mode in MISC. section of Preferences dialog, restart Notepad++, and try this command again."/>
+			<RTLvsDirectWrite title="Cannot run RTL" message="RTL is not compatible with DirectWrite mode. Please disable DirectWrite mode in MISC. section of Preferences dialog, and restart Notepad++."/>
+			<FileMemoryAllocationFailed title="Exception: File memory allocation failed" message="There is probably not enough contiguous free memory for the file being loaded by Notepad++."/><!-- HowToReproduce: Try to open multiple files with total size > ~700MB in the x86 Notepad++ (it will depend on the PC memory configuration and the current system memory usage...). -->
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Clipboard History"/>
--- a/PowerEditor/installer/nativeLang/english_customizable.xml
+++ b/PowerEditor/installer/nativeLang/english_customizable.xml
@@ -1505,7 +1505,8 @@ NOTE: Choosing not to create the placeholders or closing them later, your manual
 Would you like to create those placeholders?
 
 NOTE: Choosing not to create the placeholders or closing them later, your session WILL BE MODIFIED ON EXIT! We suggest you backup your &quot;session.xml&quot; now."/>
-			<RTLvsDirectWrite title="Cannot run RTL" message="RTL is not compatible with DirectWrite mode. Please disable DirectWrite mode in MISC. section of Preferences dialog, restart Notepad++, and try this command again."/>
+			<RTLvsDirectWrite title="Cannot run RTL" message="RTL is not compatible with DirectWrite mode. Please disable DirectWrite mode in MISC. section of Preferences dialog, and restart Notepad++."/>
+			<FileMemoryAllocationFailed title="Exception: File memory allocation failed" message="There is probably not enough contiguous free memory for the file being loaded by Notepad++."/><!-- HowToReproduce: Try to open multiple files with total size > ~700MB in the x86 Notepad++ (it will depend on the PC memory configuration and the current system memory usage...). -->
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Clipboard History"/>
--- a/PowerEditor/installer/nativeLang/french.xml
+++ b/PowerEditor/installer/nativeLang/french.xml
@@ -1505,7 +1505,8 @@ Notez que si vous choisissez de ne pas créer d'espaces réservés ou de les fer
 Voulez-vous créer des espaces réservés pour eux ?
 
 NOTE : Si vous choisissez de ne pas créer d'espaces réservés ou de les fermer plus tard, votre fichier de session sera modifié à la sortie. Nous vous suggérons de faire une sauvegarde du fichier de session &quot;session.xml&quot; maintenant."/>
-			<RTLvsDirectWrite title="Exécution de la commande &quot;Texte de droite à gauche&quot; impossible" message="&quot;Texte de droite à gauche&quot; n'est pas compatible avec le mode &quot;DirectWrite&quot;. Veuillez désactiver ce mode dans la section &quot;Divers&quot; du dialogue Préférences, redémarer Notepad++, pour exécuter la commande de nouveau."/>
+			<RTLvsDirectWrite title="Exécution de la commande &quot;Texte de droite à gauche&quot; impossible" message="&quot;Texte de droite à gauche&quot; n'est pas compatible avec le mode &quot;DirectWrite&quot;. Veuillez désactiver ce mode dans la section &quot;Divers&quot; du dialogue Préférences, et redémarer Notepad++."/>
+			<FileMemoryAllocationFailed title="Exception : échec de l'allocation de mémoire du fichier" message="Il n'y a probablement pas assez de mémoire libre contiguë pour le fichier chargé par Notepad++."/>
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Historique du presse-papier"/>
--- a/PowerEditor/installer/nativeLang/taiwaneseMandarin.xml
+++ b/PowerEditor/installer/nativeLang/taiwaneseMandarin.xml
@@ -1390,7 +1390,8 @@
 你想為它們創建佔位欄嗎？
 
 請注意，如果你選擇不創建佔位欄或稍後關閉它們，工作階段檔案將在 Notepad++ 關閉時被修改。我們建議你現在備份工作階段檔案 &quot;session.xml&quot; 。"/>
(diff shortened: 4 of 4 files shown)


alankilborn:
>Make RTL per document & remembered across the sessions

Oooh, can this be far behind?:

>Make 'word wrap' per document & remembered across the sessions

:-)

donho:
@alankilborn 
> Make 'word wrap' per document & remembered across the sessions

I would like to make EVERY options per document & remembered across the sessions...
But some options have performance issue, Word wrap option especially.

Yaron10:
@donho,

> That's why editZoneRTL="no" is introduced to localization XML file.

> and since the RTL="yes" is in the XML file, it makes sense to me that editZoneRTL="no" is beside. Users modify just once and they get their desire behaviour forever.

When you explain yourself, I understand you better. :)
Please read the following lines carefully, and try to understand the bi-LTR/RTL user experience.

Setting the interface language (and NPP LTR/RTL layout defined in it) is indeed "Users modify just once and they get their desire behaviour forever".

This is NOT the case with Text Direction. One might change that 10 times in 10 minutes!
If I open a file containing English text, the text-direction should be LTR.
If I open a file containing Hebrew text, the text-direction should be RTL.

Think of the following common case:
I open NPP and want to work on 10 English-text files.
I then close them, and open 10 Hebrew-text files.
Back to English and so on.

With current master I execute "Text Direction LTR/RTL" **once for each set of files**.
With the current PR, I'd need to execute "Text Direction LTR/RTL" **10 times for each set of files**.

_So there must be a way to override the default direction with one command._

Think also of another common case:
One week I work on a project containing 200 English-text files. - For that week, I'd like the default direction to be LTR.
Another week I work on a project containing 200 Hebrew-text files. - For that week, I'd like the default direction to be RTL.

_And therefore `editZoneRTL` in the XML localization file is extremely impractical and inconvenient._

What do you think of the following approach?
**1.** `editZoneRTL` should be a parameter saved to config.xml. 
**2.** The default/initial value should be `true` if NPP layout is RTL, and `false` otherwise.
**3.** On executing "Text Direction LTR/RTL", change `editZoneRTL` according to the new direction.
(And save to config.xml the value on exit).

***

_No `editZoneRTL`._

Open a file (-> Text Direction RTL as expected).

**Result**:
The keys (Arrow Left etc.) are in LTR mode.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0978b2e29f21e7797f488e9e6d2e85763b25a406/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L4223

Please change direction to RTL and try the left/right arrow keys with current master and then with this PR.

***

No session or no files in session. editZoneRTL="no".

Open NPP (-> Text Direction RTL).

Expected:
Text Direction LTR.

> Item.

?

***

> Fixed. Please see the latest commit.
Since the RTL can be done in non-RTL localization, the new behaviour is that the user will get warning (only once) if he/she has RTL doc, and RTL won't be applied.

You don't get the warning in a new setup or if you delete config.xml.

How about setting the default value to DirectWrite only if editZoneRTL is `false`?
The warning can be in the preference itself:

תמונה

donho:
@Yaron10 

>Setting the interface language (and NPP LTR/RTL layout defined in it) is indeed "Users modify just once and they get their desire behaviour forever".
>
>This is NOT the case with Text Direction. One might change that 10 times in 10 minutes!
If I open a file containing English text, the text-direction should be LTR.
If I open a file containing Hebrew text, the text-direction should be RTL.
>
>Think of the following common case:
I open NPP and want to work on 10 English-text files.
I then close them, and open 10 Hebrew-text files.
Back to English and so on.
>
>With current master I execute "Text Direction LTR/RTL" once for each set of files.
With the current PR, I'd need to execute "Text Direction LTR/RTL" 10 times for each set of files.
>
>So there must be a way to override the default direction with one command.
>
>Think also of another common case:
One week I work on a project containing 200 English-text files. - For that week, I'd like the default direction to be LTR.
Another week I work on a project containing 200 Hebrew-text files. - For that week, I'd like the default direction to be RTL.
>
>And therefore editZoneRTL in the XML localization file is extremely impractical and inconvenient.
>
>What do you think of the following approach?
>1. editZoneRTL should be a parameter saved to config.xml.
>2. The default/initial value should be true if NPP layout is RTL, and false otherwise.
>3. On executing "Text Direction LTR/RTL", change editZoneRTL according to the new direction.
>(And save to config.xml the value on exit).




You're talking about ONLY YOUR use case.
I can imagine another use case that only RTL because he/she is not a dev/IT person.
Or here's another use case: A dev who is comfortable with RTL GUI, but he/she uses Notepad++ for coding, so edit zone should always LTR.

All these 3 (and even more) cases can happen, so it's not fair to consider only the first case.


> With current master I execute "Text Direction LTR/RTL" once for each set of files.
With the current PR, I'd need to execute "Text Direction LTR/RTL" 10 times for each set of files.

In current PR, you can NOT set editZoneRTL, then it'll be the same behaviour as the current master.
Or you can set editZoneRTL to "no", so you have to set Hebrew file set to RTL. 
In both way, I don't see how this PR will make your use case even more fastidious.

I still believe editZoneRTL is the best solution for above 3 use cases.
For your use case I would say that you need 2 portable Notepad++, the 1st one contains only RTL="yes" settings, the 2nd one contains both RTL="yes" & editZoneRTL="no", then you open English files in the secod one and all Hebrew files in the 1st one.

> Think also of another common case:
One week I work on a project containing 200 English-text files. - For that week, I'd like the default direction to be LTR.
Another week I work on a project containing 200 Hebrew-text files. - For that week, I'd like the default direction to be RTL.

I think it's one of use casees, certainly your use case, but it's NOT a common case. 

> >What do you think of the following approach?
>1. editZoneRTL should be a parameter saved to config.xml.
>2. The default/initial value should be true if NPP layout is RTL, and false otherwise.
>3. On executing "Text Direction LTR/RTL", change editZoneRTL according to the new direction.
>(And save to config.xml the value on exit).

I don't like this approach at all.
Here are the reasons I don't like it:
1. As I've said, it's more reasonable to me to put editZoneRTL setting into localization XML file since RTL setting is there also.
2. Not only the logic is complex, that makes the implementation complex as well.
3. I don't believe it's a common case. And it could be confusing for the other RTL users.

Note that it's not a definitive reject: I don't think it's the common case. But if there are more similar requests coming, I will consider it.

Yaron10:
@donho,

With all due respect, your reply clearly indicates that you do not adequately understand RTL.

Thank you for your work, and best of luck in implementing this feature as you prefer.

alankilborn:
@Yaron10 Forgive me if this has been mentioned before, but is there a text editor that in your opinion handles RTL well?  If so, maybe a Notepad++ implementation could model that.

Yaron10:
@alankilborn,

If one claims that `2 + 2 = 5`, you can prove him wrong.
If someone insists - assertively and pompously - that `2 + 2 = potatoes`, you should nod in agreement and quietly walk away.

Thank you.

donho:
> If someone insists - assertively and pompously - that 2 + 2 = potatoes, you should nod in agreement and quietly walk away.

It could be, from my side, I've stated 2 + 2 = potatoes. But it's what  you did from your side as well.

I don't like strongly the behaviour of the feature you want, so I don't implement it for the moment. And you have been told that your suggestion will be considered if there are similar requests coming. I think it's fair what I've suggested, and let's end this story.

AndiKiissel:
@Yaron10 
Have you seen this  script https://community.notepad-plus-plus.org/topic/23402/assign-shortcut-to-a-saved-function/24 ?
This thread is not nice reading, but @alankilborn's script may help you.

AndiKiissel:
@Yaron10 

Unfortunately, not only RTL is broken or messed up lately :-(
But some older versions are still pretty good, plus helpful scripts :-)

Yaron10:
@AndiKiissel,

> Unfortunately, not only RTL is broken or messed up lately :-(

That's true.
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d7630fa42df0d3949e15353272a883c2696c817c is another example.

I've given up.

alankilborn:
@Yaron10 said:

---

>https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d7630fa42df0d3949e15353272a883c2696c817c is another example.

Another example of "broken or messed up"?
That commit is supposed to fix 14225 and indeed if I try the STR in that issue (with 8.6.1RC1), I don't see any failure...

---

>I've given up.

I am sorry to hear.
Would you change your mind if you simply ignore RTL in Notepad++ (as a forever-broken-and-not-possible-to-fix thing)?

Yaron10:
@alankilborn,

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/d7630fa42df0d3949e15353272a883c2696c817c.
> Another example of "broken or messed up"?

**STR:**
Open files A and B.
Move B to the other view (pos: 1).
Activate A.
Externally modify and save B.
Activate B _by left-clicking line 10 in the editing area_.

**Result:**
Wrong "selection" and flickering cursor.
See:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/d039eaa9009e3561589cb3b5e9849ab9aefa3e02/PowerEditor/src/Notepad_plus.cpp#L6393

***

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a51e62c16dc45160afe6c7deb19096e98ad18372

Try it with new untitled cloned document, and with **2** sets of cloned files.

***

> I am sorry to hear.

Thank you.

1. I don't have the time and energy for the futile discussions we both know so well. :)
2. Basic RTL support is fundamental for my personal use.

Issue #14245, opened 2023-10-19
Feature request: Alphabetic order for languages


rozsahegyip:
### Description of the Issue
It's a feature request. It would be great if the list of languages would be in alphabetic order on Settings -> Program settings -> Languages like in Languages menu point of the menu bar.

### Steps to Reproduce the Issue

1. Click Settings -> Program settings -> Languages
2. The languages are listed in random order

### Expected Behavior
Show languages in alphabetic order like in the Language menu or add a button or switch to turn on/off ABC sorting.

### Actual Behavior
Randomized order of languages w/o any obvious logic (at least for me).
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/148427828/d17cd064-e57a-4f84-839a-199f378518d2)

### Debug Information
Notepad++ v8.5.6   (64-bit)
Build time : Aug 15 2023 - 15:29:28
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 Pro N for Workstations (64-bit)
OS Version : 22H2
OS Build : 22621.2428
Current ANSI codepage : 1250
Plugins : 
    JsonTools (4.0.0.1)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NppFTP (0.29.12)
    NPPJSONViewer (2.0.3)

saivaka:
Could this be assigned to me

PR #14442, opened 2023-12-04
Saivaka task1 - Feature Addition - Sort Languages in Settings -> Preferences -> Languages 

saivaka:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14245

A new feature request implementation to add a functionality to sort languages in settings -> preferences -> languages:

In order to implement the feature, we discovered where in the code files dealt with the language settings. Once found, we added our command boxes (buttons) into the .rc files and used its “onclick” functionality in the .cpp file. Once we discovered which data structures dealt with the coding language names, we used the button onclick event to sort the languages in the data structure and display it in the settings. Additionally, we had to create a comparator, since the data structure used a vector of structs, in order to use the built in c++ sort function. 

Please provide any feedback to merge this functionality into your software. I'm open to making any necessary code changes such as moving the comparator to a different, more appropriate file, or even moving the buttons down below the listboxes. 

Screenshot 2023-12-03 234022

PR #14443, opened 2023-12-04
Fix 14245: Alphabetic order for languages

briangood35:
#14245 

Feature:
Sort languages (and excluded languages) alphabetically.

Changes:
-Override less than operator for LangMenuItem struct in order to call std::sort on the vector of languages. This operator will sort alphabetically regardless of upper vs. lower case for better navigation.
-Call std::sort on language vector before displaying the language lists.

Results:
-Lists are now sorted each time Notepad++ is opened.
image

### DIFF
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -586,6 +586,14 @@ struct LangMenuItem final
 
 	LangMenuItem(LangType lt, int cmdID = 0, const std::wstring& langName = TEXT("")):
 	_langType(lt), _cmdID(cmdID), _langName(langName){};
+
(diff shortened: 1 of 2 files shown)


chcg:
@briangood35 This seems to be a second PR to fix the issue. PR #14442 from @Saivaka already exists. Were you aware of each other?

donho:
@chcg 
It doesn't matter.
I will choose the better implementation... Or if I'm not satisfied on the both ones, I'll provide my implementation.

briangood35:
@chcg I saw the other PR, however I had already almost finished my implementation. I figured our methods were different enough (i.e. no UI change) that I might as well offer my changes for review.

PR #14473, opened 2023-12-11
"-loadingTime" cmdline param enhancement

### LINKED ISSUE #14472: [FR] "-loadingTime" cmdline param enhancement
Currently (v8.6) the "-loadingTime" only roughly reports Notepad++ loading time in seconds.

I would like to:
- improve the reported time accuracy (from seconds to milliseconds, so e.g. different N++ settings benchmarking will be possible)
- add the ability to analyze various types of problems (e.g. by distinguishing in between the time taken by the app/plugin code-init from the possible startup file loading time)
- maintain the existing simplicity of the "-loadingTime" option

xomx:
Fix #14472

Improves:
- the reported time accuracy (from seconds to milliseconds, now e.g. different N++ settings benchmarking is possible)
- the ability to analyze various problems (now it is possible to distinguish the time taken by the app/plugin code-init from the possible file loading time)

### DIFF
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -48,6 +48,10 @@
 #include <chrono>
 
 extern std::chrono::steady_clock::time_point g_nppStartTimePoint;
+extern std::chrono::steady_clock::duration g_nppInitTime;
+extern std::chrono::steady_clock::duration g_pluginsLoadingTime;
+extern std::chrono::steady_clock::duration g_sessionLoadingTime;
+extern std::chrono::steady_clock::duration g_cmdlineParamsLoadingTime;
 
 
 #define MENU 0x01
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -48,10 +48,7 @@
 #include <chrono>
 
 extern std::chrono::steady_clock::time_point g_nppStartTimePoint;
-extern std::chrono::steady_clock::duration g_nppInitTime;
 extern std::chrono::steady_clock::duration g_pluginsLoadingTime;
-extern std::chrono::steady_clock::duration g_sessionLoadingTime;
-extern std::chrono::steady_clock::duration g_cmdlineParamsLoadingTime;
 
 
 #define MENU 0x01
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -45,6 +45,9 @@
 #include "md5Dlgs.h"
 #include <vector>
 #include <iso646.h>
+#include <chrono>
+
+extern std::chrono::steady_clock::time_point g_nppStartTimePoint;
 
 
 #define MENU 0x01
--- a/PowerEditor/src/Notepad_plus_Window.cpp
+++ b/PowerEditor/src/Notepad_plus_Window.cpp
@@ -21,10 +21,6 @@
 const TCHAR Notepad_plus_Window::_className[32] = TEXT("Notepad++");
 HWND Notepad_plus_Window::gNppHWND = NULL;
 
-std::chrono::steady_clock::duration g_nppInitTime{};
-std::chrono::steady_clock::duration g_sessionLoadingTime{};
-std::chrono::steady_clock::duration g_cmdlineParamsLoadingTime{};
-
 
 namespace // anonymous
 {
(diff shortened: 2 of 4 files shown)


PowerEditor/src/Notepad_plus_Window.cpp
```diff
@@ -385,12 +390,13 @@ void Notepad_plus_Window::init(HINSTANCE hInst, HWND parent, const TCHAR *cmdLin
...
-		char dest[256];
-		sprintf(dest, "Loading time : %.0lf seconds", loadTime);
-		::MessageBoxA(NULL, dest, "", MB_OK);
+		std::chrono::steady_clock::duration nppInitTime = (std::chrono::steady_clock::now() - g_nppStartTimePoint) - sessionLoadingTime - cmdlineParamsLoadingTime;
+		std::wstringstream wss;
+		wss << L"app & plugin(s) initialization: " << std::chrono::hh_mm_ss{ std::chrono::duration_cast<std::chrono::milliseconds>(nppInitTime) } << std::endl;
+		wss << L"last session loading: " << std::chrono::hh_mm_ss{ std::chrono::duration_cast<std::chrono::milliseconds>(sessionLoadingTime) } << std::endl;
+		wss << L"cmdline params handling: " << std::chrono::hh_mm_ss{ std::chrono::duration_cast<std::chrono::milliseconds>(cmdlineParamsLoadingTime) } << std::endl << std::endl;
+		wss << L"overall: " << std::chrono::hh_mm_ss{ std::chrono::duration_cast<std::chrono::milliseconds>(nppInitTime + sessionLoadingTime + cmdlineParamsLoadingTime) };
+		::MessageBoxW(NULL, wss.str().c_str(), L"Notepad++ loading time (hh:mm:ss.ms)", MB_OK);
```
donho, line 399: Very good PR. But if we can get plugins loading time, it'll be even better.
How about the following message:
```
Notepad++ initialization: 0:0:0.007
Plugins loading: 0:0:0.666
Last session loading: 0:0:0.042
Command line params handling: 0:0:0.069
Total loading time: 0:0:0.784
```
Since you have already the number of *"app & plugin(s) initialization:"*, you can get **"Notepad++ initialization:"** easily if you have also "Plugins loading:":
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Notepad_plus.cpp#L460

What do you think?

donho (changes requested):

PowerEditor/src/Notepad_plus_Window.cpp
xomx: I like both the message and the addition of separate time measurement for plugins.

I have already considered some others (e.g. to have exact pure file loading time instead of the sessionTime and the cmdlineTime measurements) but I think that the current state of PR is enough (+ the pluginsTime).
We can always add or refine some criteria later.

xomx:
@donho 
And what do you think about the possible addition of the translation?
For my purposes it is fine as is, but maybe we should be consistent in the N++ GUI and do not leave this untranslated...

donho:
@xomx 
> And what do you think about the possible addition of the translation?

It's used for devs (debugging), not for the users. So the translation is not necessary IMO.

xomx:
@donho 

Done:

npp-loading-time-enh

PowerEditor/src/Notepad_plus.h
```diff
@@ -45,6 +45,13 @@
 #include "md5Dlgs.h"
 #include <vector>
 #include <iso646.h>
+#include <chrono>
+
+extern std::chrono::steady_clock::time_point g_nppStartTimePoint;
+extern std::chrono::steady_clock::duration g_nppInitTime;
+extern std::chrono::steady_clock::duration g_pluginsLoadingTime;
+extern std::chrono::steady_clock::duration g_sessionLoadingTime;
+extern std::chrono::steady_clock::duration g_cmdlineParamsLoadingTime;
```
donho, line 54: Only g_nppStartTimePoint & g_pluginsLoadingTime need to be external.
The others please make them local, as in the first commit.

donho (changes requested):
Please reduce the variable scopes if they are not necessary to be external.

donho (approved):

xomx:
@donho 
> By checking the method `loadCommandlineParams()`, I'm wondering if it needs to be measured.
> Let's see in the future and tune it if it's necessary.

Files can be loaded via cmdline too...

donho:
@xomx 
> Files can be loaded via cmdline too...

Indeed! That's a good and valid argument!

Issue #14296, opened 2023-10-31
[Feature Request] Shortcut Key Enhancement


nrv-96:
### Description of the Feature

Please consider adding the following shortcut, as it would be more impactful and useful:

**Shift + Delete**:  without selected text, it will delete the whole line.
**Ctrl + C**:  without selected text, it will copy the whole line
**Ctrl + X**: without selected text, it will cut the whole line

p.s. : This shortcut works well in VSCode.

alankilborn:
For <kbd>Shift</kbd>+<kbd>Delete</kbd>, just do some remapping in _Shortcut Mapper_ such that that keycombo is tied to the _SCI_LINEDELETE_ command.

For <kbd>Ctrl</kbd>+<kbd>c</kbd> and <kbd>Ctrl</kbd>+<kbd>x</kbd>, the _Visual Studio Line Copy_ plugin should work.

nrv-96:
Thanks @alankilborn, It works. 

@donho Can we make the **Visual Studio Line Copy plugin** a core functionality? It will make a developer's life easier :)

Yaron10:
@nrv-96,

You can remap in the Shortcut Mapper `SCI_LINECOPY` & `SCI_LINECUT` as well.

alankilborn:
>You can remap in the Shortcut Mapper SCI_LINECOPY & SCI_LINECUT as well.

That's not (all that) useful.  User wants one keycombo to have dual functionality depending upon selected text or not.  Reading again, my advice about Shift+Delete violates that (so it is bad advice and should be ignored).

donho:
> Shift + Delete: without selected text, it will delete the whole line.

@alankilborn @Yaron10 
Without selection, Shift-DEL, Ctrl-C and Ctrl-X do nothing currently in Notepad++.
If the following behaviours
```
Shift + Delete: without selected text, it will delete the whole line.
Ctrl + C: without selected text, it will copy the whole line
Ctrl + X: without selected text, it will cut the whole line
```
are in the most popular code/text editors, why does not Notepad++ align with?

alankilborn:
I have no argument against Notepad++ aligning with the described behavior.

Yaron10:
> ...are in the most popular code/text editors, why does not Notepad++ align with?

👍 
IMO _Copy/Delete/Cut Current Line_ should be also added to the menu.

MAPJe71:
[...]

PR #14476, opened 2023-12-12
Make column to multi-select be abled to be disabled

### LINKED ISSUE #14464: Is there a key combination to cancel column mode (or multi-edit) since v8.6?
I use column mode a lot with the simple shift-Alt key combination.
But since v8.6 I'm getting stuck in column mode  (or multi-edit or whatever this new feature is called) and can't get out. 
I can't find any preference to re-enable the pre-8.6 behavior and can't find a key combination to cancel column mode.
All I can do now is change my selection into 1 line and only then does column mode seem to cancel itself. It seems very unintuitive.

donho:
In order to disable the "Column to multi-select" ability (new feature introduced in v8.6), users must add an empty file named "noColumnToMultiSelect.xml", in "%APPDATA%\Notepad++\" directory (or in the Notepad++ installed directory in portable mode) to prevent this behaviour.

ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14296#issuecomment-1848076816

Fix #14464

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -1635,6 +1635,21 @@ bool NppParameters::load()
 		_isRegForOSAppRestartDisabled = (::PathFileExists(filePath.c_str()) == TRUE);
 	}
 
+
+	//-------------------------------------------------------------//
+	// noColumnToMultiSelect.xml                                   //
+	// This empty xml file is optional - user adds this empty file //
+	// manually in order to prevent Notepad++ transform column     //
+	// selection into multi-select.                                //
+	//-------------------------------------------------------------//
+	std::wstring enableNoColumn2MultiSelectPath = _userPath;
+	pathAppend(enableNoColumn2MultiSelectPath, TEXT("noColumnToMultiSelect.xml"));
+
+	if (PathFileExists(enableNoColumn2MultiSelectPath.c_str()))
+	{
+		_column2MultiSelect = false;
+	}
+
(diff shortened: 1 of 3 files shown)


cmanley:
Wasn't it possible to use a checkbox preference to achieve the goal instead of using an empty file?

Shararamosh:
I closed Notepad++, added this file to %appdata%/Notepad++, opened Notepad++ again, but it has not affected it - it still has multi-line editing.

Shararamosh:
It's an unmodified official version of Notepad++. I was always using the latest version and updated every time app prompted me.

alankilborn:
@Shararamosh 

The change that uses the `noColumnToMultiSelect.xml` isn't in a official version of Notepad++ yet, so that explains why it did nothing for you when you tried it.

Shararamosh:
Oh, boy.
Alright, guess I'll just wait for the update.

alankilborn:
@Shararamosh If you want to try it NOW, you can get the .exe file containing the change by clicking "Checks" and then "Artifacts" and then picking the correct thing to download, e.g. MSVC x64 release, if you are using the 64-bit version of Notepad++ (what most people run):

image

Rename what you download to be `notepad++.exe` and copy it over the original of that file wherever you run Notepad++ from (perhaps rename the original to e.g. `notepad++_orig.exe` first).

DougALug:
Creating the noColumnToMultiSelect.xml worked perfectly. Thank you so much for that.

Issue #14353, opened 2023-11-14
Python f-strings are not recognised as text strings


IanSav:
### Description of the Issue
Python 3.10 and higher now offer f-strings.  These are quoted text strings that allow inline specification of python variables and code.  This code like:
`print("Variable fred has a value of '%s'." % fred)`
can now be written as:
`print(f"Variable fred has a value of '{fred'}'.")`
In the first version Notepad++ shows the quoted string in the text highlight while the second is missing the text highlight.

### Steps to Reproduce the Issue
Any use of f-strings within the editor in Python mode will show the issue.

### Expected Behavior
If Notepad++ is correctly detecting the new syntax then the string checking and highlighting should work as expected.

It would be helpful if the text within the braces "{}" of f-strings is shown as code and not text.

### Actual Behavior
At the moment text highlighting is not happening.

### Debug Information

Notepad++ v8.5.8   (64-bit)
Build time : Oct 15 2023 - 21:43:56
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "Filename.py" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 22H2
OS Build : 19045.3636
Current ANSI codepage : 1252
Plugins : 
    DSpellCheck (1.4.24)
    HexEditor (0.9.12)
    HTMLTag (1.3.5)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    Remove Duplicate Lines (1.3)
    SpeechPlugin (0.4)
    XMLTools (3.1.1.13)
    zoomdisabler_x64 (1.2)

![Notepad++Highlighting](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/6904736/06634d8f-d01c-49f2-a35c-ed43e5dcfd39)

alankilborn:
In Notepad++ 8.5.8:

![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/77065706/ce872b66-6f80-47d1-8dc1-dc61d7c03dca)

You just need to update your langs.xml file.
Get new one from downloading 8.5.8 portable and extracting that file.
Either copy over your outdated langs.xml file, or, if you've ever modified settings for languages and want to keep those mods, diff the 8.5.8 file with yours and merge relevant updates into your version of the file.

IanSav:
[...]

PR #14507, opened 2023-12-19
Fix python wrong decorator attribute colors

### LINKED ISSUE #5894: Incorrect highlight python decorator
### Description of the Issue
The decorator color is not yellow if it contains a dot.

Paste this code into Notepad ++ 7.7.1 32 bit:

```python
import test_module


@test_module.test_decorator
def test_function():
    pass
```

### Expected Behavior
The entire decorator should be yellow.

### Actual Behavior
decorator_color_test

PavelBlend:
Fix #5894

Decorator color are now displayed correctly:

The solution to the problem was suggested here:
https://github.com/ScintillaOrg/lexilla/pull/219#issuecomment-1849782847

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
@@ -887,6 +887,7 @@ protected:
 	void setPythonLexer() {
 		setLexer(L_PYTHON, LIST_0 | LIST_1);
 		execute(SCI_SETPROPERTY, reinterpret_cast<WPARAM>("fold.quotes.python"), reinterpret_cast<LPARAM>("1"));
+		execute(SCI_SETPROPERTY, reinterpret_cast<WPARAM>("lexer.python.decorator.attributes"), reinterpret_cast<LPARAM>("1"));
 	};
 	
 	void setGDScriptLexer() {


vinsworldcom:
Please update the text of your pull request from:

> Solves this problem: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5894

to:

Fix #5894 

so that the pull request will be "linked" to that issue and have a better time getting accepted and incorporated.

Cheers.

rdipardo:
@PavelBlend

> Decorator color are now displayed correctly

Where does this color come from? So far this PR only changes the lexer setup function. To make decorator attributes display properly, you still need to associate a style with the ID number of the new `SCE_P_ATTRIBUTE` lexical class (which is `20`). At the very least, a new style should to be added to [the default theme's configuration file][0], e.g.,

~~~xml
<LexerType name="python" desc="Python" ext="">
    <!-- . . . -->
    <WordsStyle name="DECORATOR" styleID="15" fgColor="FF8000" bgColor="FFFFFF" fontName="" fontStyle="2" fontSize="" />
    <!-- . . . -->
    <WordsStyle name="ATTRIBUTE" styleID="20" fgColor="FF8000" bgColor="FFFFFF" fontName="" fontStyle="2" fontSize="" />
</LexerType>
~~~

People using other N++ themes will not see any difference until their theme's own XML file gets updated. In fact, no difference will be apparent even in the default theme unless they *also* delete their existing copy of `%AppData%\Notepad++\stylers.xml`, since new installations [do not overwrite existing configuration files][2].


[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/7419770d2eacc4ad9cf94556887b37ad0de3a1de/PowerEditor/src/stylers.model.xml#L1036
[1]: https://github.com/ScintillaOrg/lexilla/pull/219#issuecomment-1849782847
[2]: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14353#issuecomment-1810174152

PavelBlend:
@rdipardo It works for me without specifying the color. I think the color is specified by the decorator color parameter:
```xml
<WordsStyle name="DECORATOR" styleID="15" fgColor="009900" bgColor="FFFFFF" fontName="" fontStyle="2" fontSize="" />
```
Color set to `009900`. Here is the result:


Here's another theme:

rdipardo:
Looking [here][0], notice where `{20}` (for `SCE_P_ATTRIBUTE`) shows up, e.g.,

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/7419770d2eacc4ad9cf94556887b37ad0de3a1de/lexilla/test/examples/python/attributes/attributes.py.styled#L5

In fact this PR *does not* actually introduce the `SCE_P_ATTRIBUTE` class at all (as I mistakenly thought before). It simply extends the existing `SCE_P_DECORATOR` style past the `.` attribute accessor (and only within decorators), like elsewhere in the same test file, e.g.,

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/7419770d2eacc4ad9cf94556887b37ad0de3a1de/lexilla/test/examples/python/attributes/attributes.py.styled#L11

This explains why the decorator attribute is already `#FF8000` in the screen capture above.

Actually making `SCE_P_ATTRIBUTE` a real lexical class would involve turning on the `lexer.python.identifier.attributes` option. But that's a separate issue, and a messier one to integrate, because then every theme really would need updating.

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/lexilla/test/examples/python/attributes/attributes.py.styled

PR #14520, opened 2023-12-26
Fix "Find all ..." commands results hidden issue  for long lines

### LINKED ISSUE #12023: "Find All..." does not work right with long lines
### Description of the Issue

"Find All..." does not show all findings in long lines

1. Unpack 64bit (or 32bit) Portable Minimalist Notepad++ v8.4.4 in new folder
2. Open this testfile [_long_line_test.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/9373796/_long_line_test.txt)
3. Find...
4. Find what: a
5. Find All in Current Document

image

6. Try to catch all "a"-s

### Expected Behavior

All "a"-s found, **unindependent of line length**

### Actual Behavior

"a"-s not found after Col:~2048

image

**Before v8.4.3 all "a"-s could be found with F4 / Shift-F4, but not now.**

### Debug Information

Notepad++ v8.4.4   (64-bit)
Path : F:\NppTest64\notepad++.exe
Command Line : F:\NppTest64\_long_line_test.txt
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.1889
Current ANSI codepage : 1257
Plugins : none

donho:
With this PR, all the results of "Find all in..." commands will be displayed, even with long lines.
All the results displayed in the cut long lines will be kept as the current behaviour. Additionally, each result beyond the cut long line will be displayed as default mode (ie. each entry will be displayed, so user can double click on the entry line for reaching the result).

Related: #12014

Fix #12023


alankilborn:
The new behavior could be confusing from a user perspective, so I'll add some notes. 
@pryrt can decide if it should be described in the user manual, and if so, how best to do that.

When "show only one entry per found line if possible" is active...
And not all of the matches can be shown on one output line because the maximum output line length is encountered (and the ellipsis `...` is shown)...
The remaining matches are "shown" (but not highlighted in red with yellow background) by using one additional output line per match.  All of these additional output lines appear identical in the _Search results_ window (and, in fact, are identical in content to the first line of output -- but that first line has red on yellow match highlights).

Thus, for the original STR example in #12023, the first output line will show 64 highlighted matches and then 136 more copies of the line (sans highlighting) will be shown, representing 136 more matches.  Thus, the total number of matches for the file is 64+136 = 200.

PR #14534, opened 2023-12-30
[xml] Update japanese.xml

yasmise:
Add translation texts for these commits:
* Add new localization entries (d039eaa)
* Add navigation to the 1st & last tab abilities (4e6bbbc)

This also includes fix-ups to existing texts, to make it clearer.

### DIFF
--- a/PowerEditor/installer/nativeLang/japanese.xml
+++ b/PowerEditor/installer/nativeLang/japanese.xml
@@ -304,13 +304,13 @@ Translation note:
 					<Item id="44092" name="タブ7"/>
 					<Item id="44093" name="タブ8"/>
 					<Item id="44094" name="タブ9"/>
-					<Item id="44116" name="最初のタブ"/>
-					<Item id="44117" name="最後のタブ"/>
-					<Item id="44095" name="次のタブ"/>
-					<Item id="44096" name="前のタブ"/>
+					<Item id="44116" name="先頭のタブを表示"/>
+					<Item id="44117" name="末尾のタブを表示"/>
+					<Item id="44095" name="次のタブを表示"/>
+					<Item id="44096" name="前のタブを表示"/>
 					<Item id="44097" name="ファイル監視(tail -f)"/>
-					<Item id="44098" name="タブを右へ移動"/>
-					<Item id="44099" name="タブを左へ移動"/>
+					<Item id="44098" name="タブを後ろへ移動"/>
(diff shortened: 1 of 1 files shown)


yasmise:
GitHub Action's `xml validation test` failed, but syntax checking by XML Tools plugin reports no error. Maybe this fail comes from temporary error of Action.

yasmise:
@donho Thanks for looking into this.
Reading the Action's log, building lxml caused the fail, and I found that [the last good run used `lxml-4.9.4`](https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/7359505432/job/20034381548), but now, `lxml-5.0.0` is used. Is it possible that version change of lxml affects Actions?

rdipardo:
> I found that [the last good run used `lxml-4.9.4`](https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/7359505432/job/20034381548), but now, `lxml-5.0.0` is used. Is it possible that version change of lxml affects Actions?

@yasmise, @chcg

This does not appear to affect Python 3.12. Could be the runner's [built-in software][0] needs updating. Instead of waiting for GitHub to do it, why not install a newer Python by ourselves?, e.g.

```diff
diff --git a/.github/workflows/CI_build.yml b/.github/workflows/CI_build.yml
index 84ff0eb5e..f9528f895 100644
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -58,6 +58,11 @@ jobs:
           name: Notepad++.MSVC.${{ matrix.build_platform}}.${{ matrix.build_configuration}}
           path: PowerEditor\visual.net\${{ matrix.build_configuration}}\Notepad++.exe
 
+    - name: Set up python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+
     - name: Run xml validation test for Win32 / Debug only
       if: matrix.build_platform == 'Win32' && matrix.build_configuration == 'Debug'
       working-directory: .\
```

[0]: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#language-and-runtime

donho:
> This does not appear to affect Python 3.12. Could be the runner's [built-in software](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#language-and-runtime) needs updating. Instead of waiting for GitHub to do it, why not install a newer Python by ourselves?, e.g.

It seems Github has fixed the problem now.

PR #14536, opened 2023-12-31
Fix deleting in column mode also delete an unexpected EOL

### LINKED ISSUE #14426: v 8.6 column delete also deleted an unexpected end-of-lines
### Description of the Issue
This is a v8.6 regression.

I had some text with varying line lengths and wanted to add a second column to the right at column 37.  On lines 6 and 7 below I manually added the tabs. I saw that lines 1 to 5 were shorter and so rather than manually adding tabs needing to pay attention I selected the area and then did a search/replace to replace `$` with 9 tabs `\t\t\t\t\t\t\t\t\t`.

My normal practice at this point is I moved the typing cursor to the end of line 7 and while holding Shift+Alt I did up-arrow to line 1 and then held down the right arrow key until things stopped jumping around meaning I was past the end of the longest line. I then pressed Del but was surprised that line 7 was appended on line 6 and line 8 was appended to line 7.

With Notepad v8.5.8 this worked as I expected and it deleted the excess tabs without also deleting any end of line marks.  To make the end of line visible in this post I appended a tilde `~` to each line. They were not part of the test and so should be removed if you want to copy the test data below into Notepad to reproduce the issue.

The 9_9_9_... text was already there as the second column of text for line 8. I was about to add the second column of text for lines 1-7 which lead me to add the tabs and then try to trim the excess so that all lines ended at column 37 meaning I could type stuff and then up-arrow as I worked my way up the table.

### Original data
```txt
1_1_1_									~
2_2_2									~
3_3_3_3_3									~
4_4_4_4_4_4_4_4_4_4_4_									~
5_5_5_5_5_									~
6_6_6_6_6_6_6_6_6_6_6_6_6_6_6_		~
7_7_7_7_7_7_7_7_7_7_7_7_7_7_7		~
8_8_8_8_8_8							9_9_9_9_9_9_9_9_9_~
```

### Results after the `Del` using Notepad++ v8.6
```txt
1_1_1_								~
2_2_2								~
3_3_3_3_3							~
4_4_4_4_4_4_4_4_4_4_4_				~
5_5_5_5_5_							~
6_6_6_6_6_6_6_6_6_6_6_6_6_6_6_		7_7_7_7_7_7_7_7_7_7_7_7_7_7_7		8_8_8_8_8_8							9_9_9_9_9_9_9_9_9_~
```

### Results after the `Del` using Notepad++ v8.5.8 - this is the result I expected.
```txt
1_1_1_								~
2_2_2								~
3_3_3_3_3							~
4_4_4_4_4_4_4_4_4_4_4_				~
5_5_5_5_5_							~
6_6_6_6_6_6_6_6_6_6_6_6_6_6_6_		~
7_7_7_7_7_7_7_7_7_7_7_7_7_7_7		~
8_8_8_8_8_8							9_9_9_9_9_9_9_9_9_~
```

donho:
Fix #14426

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -592,17 +592,20 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
 						if (nbCaseForScint)
 							_callWindowProc(_scintillaDefaultProc, hwnd, Message, wParam, lParam);
 
-						// then do our job, if any
-						for (const auto& i : edgeOfEol)
+						// then do our job, if it's not column mode
+						if (!isColumnSelection)
 						{
-							// because the current caret modification will change the other caret positions,
-							// so we get them dynamically in the loop.
-							LRESULT posStart = execute(SCI_GETSELECTIONNSTART, i._selIndex);
-							LRESULT posEnd = execute(SCI_GETSELECTIONNEND, i._selIndex);
+							for (const auto& i : edgeOfEol)
+							{
+								// because the current caret modification will change the other caret positions,
+								// so we get them dynamically in the loop.
+								LRESULT posStart = execute(SCI_GETSELECTIONNSTART, i._selIndex);
+								LRESULT posEnd = execute(SCI_GETSELECTIONNEND, i._selIndex);
 
-							replaceTarget(L"", posStart, posEnd + i._len2remove);
-							execute(SCI_SETSELECTIONNSTART, i._selIndex, posStart);
-							execute(SCI_SETSELECTIONNEND, i._selIndex, posStart);
+								replaceTarget(L"", posStart, posEnd + i._len2remove);
+								execute(SCI_SETSELECTIONNSTART, i._selIndex, posStart);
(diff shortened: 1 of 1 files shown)


mkupper:
@donho, I can't figure out where the most recently built but not yet released installer.exe and portable.zip/7z are.

alankilborn:
@mkupper 
All you get is an `.exe`.
For this case, grab it here: https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/7372306928

mkupper:
Thank you @alankilborn - I downloaded and am using
```
Notepad++ v8.6   (64-bit)
Build time : Dec 31 2023 - 19:18:45
```
Column delete and the end of line issue has been fixed.  Thank you!

@donho - I noticed the version number in debug info is v8.6. I would have expected v8.6.1

The v8.6.1 release candidates show  v8.6.1 in their debug info.

alankilborn:
>I noticed the version number in debug info is v8.6. I would have expected v8.6.1

As 8.6.1 hasn't been released yet (just in the release-candidate stage), I'd say 8.6 is what is expected.
But, "build time" should be accurate and can be used to tell where one `.exe` differs from another (if you lose track).

When I grab such an `.exe` for testing, I usually rename it to `notepad++_PRxxxxx.exe` where `xxxxx` is the pull-request number.  It helps me keep track of what I'm looking at, if I "go away and come back later" from something.

PR #14586, opened 2024-01-13
Fix Copy/Cut/Paste issue in Veritical Edge text field of preferences

### LINKED ISSUE #13874: Copy/paste using the keyboard does not work in Preferences / Vertical Edge Settings field
### Description of the Issue
I was experimenting with Preferences / Margins/Border/Edge / Vertical Edge Settings to better understand how it works and discovered that I can't use Ctrl-C, Ctrl-V, nor Ctrl-X to copy/paste/cut values to or from the Vertical Edge Settings field.

Ctrl-A for Select-all and Ctrl-Z for undo, and Ctrl-Y for redo also do not work in this field.

Ctrl-H for backspace works meaning the Ctrl key seems to be working but not Ctrl with A, C, V, X, Y, or Z.

Copy, paste, cut, and select-all all work if I use the mouse to right-click in the field and select Copy, Paste, etc. from the right-click menu.

Ctrl-C using the keyboard works to get the values from other Preferences fields meaning it's an issue that seems specific to the Vertical Edge Settings field.

This is not a v8.5.4 regression as older versions such as v8.3.3 also behave the same way with the keyboard's Ctrl-C, Ctrl-V, and Ctrl-X not working in the Vertical Edge Settings. A [comment in issue 9296](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9296#issuecomment-753485478) from January 2021 also mentions the issue with using the keyboard in this field. I'm creating this one as a separate issue as 9296 is asking for another feature with a suggested workaround being to copy/paste the values.

### Expected Behavior
That the Ctrl keys work the way they usually do in text/numeric fields.
Ctrl-A to select all text in the field.
Ctrl-C to copy the selected test into the Windows clipboard.
Ctrl-V to paste from Windows clipboard into the edit field.
Ctrl-X to cut the selected test into the Windows clipboard.
Ctrl-Y for redo from the redo-history.
Ctrl-Z for undo the undo history.

### Actual Behavior
The Ctrl keys other than Ctrl-H seem to do nothing in this field. I don't have speakers on my computer and so don't know if there are any beeps.

### Debug Information
(while this debug information dump shows some plugins in use the issue also happens with all plugins disabled)

Notepad++ v8.5.4   (32-bit)
Build time : Jun 17 2023 - 20:40:14
Path : c:\npp\npp854-RTM\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.1848
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
Fix #13874

### DIFF
--- a/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
+++ b/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
@@ -844,11 +844,21 @@ static LRESULT CALLBACK editNumSpaceProc(HWND hwnd, UINT message, WPARAM wParam,
 		case WM_CHAR:
 		{
 			// All non decimal numbers and non white space and non backspace are ignored
-			if ((wParam != 8 && wParam != 32 && wParam < 48) || wParam > 57)
+			SHORT ctrl = GetKeyState(VK_CONTROL);
+			SHORT alt = GetKeyState(VK_MENU);
+			SHORT shift = GetKeyState(VK_SHIFT);
+			if (!ctrl && !alt && !shift)
 			{
-				return TRUE;
+				if ((wParam != 8 && wParam != ' ' && wParam < '0') || wParam > '9')
+				{
+					return TRUE;
+				}
 			}
 		}
+		break;
+
+		default:
+			break;
 	}
 	return oldFunclstToolbarProc(hwnd, message, wParam, lParam);
 }


Mateos81:
Little typo in the commit/PR message: Veritical -> Vertical

PR #14588, opened 2024-01-13
Fix Copy/Cut/Paste shortcut modification conflict problem

### LINKED ISSUE #14585: Shortcut Mapper in v8.6.2 RC misbehaves when changing Edit Cut, Copy or Paste
### Description of the Issue
In v8.6.2 RC, the shortcut mappers shows two each of Edit Cut, Copy and Paste commands.
After making any change to the default for any of these commands, once Notepad++ is closed and then opened again, both commands of any modified pairs are set to the same shortcut and highlighted in red. If one of a pair is cleared, both are cleared on next launch. There appears to be no way to restore defaults except to modify or delete shortcuts.xml.

This also affects the initial launch after upgrading if any of these commands were set in shortcuts.xml.

1. Change any one of the shortcut mappings for Edit Cut, Copy or Paste.
2. Close and reopen Notepad++.
3. Open Shortcut Mapper and look at the mappings.

### Expected Behavior
The mappings set in one session would persist to the next.

### Actual Behavior
The changed value is duplicated as the value of the other command with the same name.

### Debug Information
Notepad++ v8.6.2   (64-bit)
Build time : Jan 11 2024 - 17:11:54
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.3803
Current ANSI codepage : 1252
Plugins : 
    ColumnsPlusPlus (1.0.2)
    CSVLint (0.4.6.6)
    DSpellCheck (1.5)
    JsonTools (5.6)
    mimeTools (3)
    MultiReplace (2.1.0.8)
    nppConverter (4.5)
    NppExport (0.4)
    NppMarkdownPanel (0.7.3)

donho:
Fix #14585

### DIFF
--- a/PowerEditor/src/NppCommands.cpp
+++ b/PowerEditor/src/NppCommands.cpp
@@ -364,7 +364,7 @@ void Notepad_plus::command(int id)
 				{
 					_pEditView->execute(WM_CUT);
 				}
-				else // Cul the entire line
+				else // Cut the entire line
 				{
 					_pEditView->execute(SCI_COPYALLOWLINE);
 					_pEditView->execute(SCI_LINEDELETE);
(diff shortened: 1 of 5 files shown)


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/75ae73c17333e6a83ce84c11ef052b6e2e5693ca

Issue #13722, opened 2023-05-31
"Go to matching brace" not working as expected with macros.


Buggfix42:
### Description of the Issue
The "Go to matching brace" command seems to be actually split in two: Go from opening brace to closing brace and vice versa.
Depending on the brace type one of both is executed. Normally this isn't noticeable, except when you define a macro for it.

### Steps to Reproduce the Issue

1. Define a macro that uses the functionality "Go to matching brace"
2. Execution of that macro only works in one direction (opening brace to closing brace or vice versa)

### Debug Information
Reproduceable on v8.5.3

alankilborn:
Probably need to elaborate your "Steps to Reproduce" with a real example.  From the limited description, it sounds to me as if it is working as intended.

CennoxX:
1. This is a text with cursor: `|(example)`
2. Record a macro
3. Menu > Search > Go to Matching Brace
4. cursor changes position: `(example|)`
5. Stop macro
6. Set cursor at starting position: `|(example)`
7. Run macro
8. Expected result: `(example|)`, but the cursor doesn't change position
9. Set cursor at this position: `(example|)`
10. Run macro
11. Position of cursor changed as expected: `|(example)`

As said, the execution of that macro only works in one direction (opening brace to closing brace or vice versa)

Buggfix42:
No. It depends on how you recorded the macro. If you recorded the macro by placing the cursor on a closing brace and execute "Go to matching brace" then the macro can only jump from a opening brace to an closing brace.
If you recorded the macro by placing the cursor on a opening brace and execute "Go to matching brace" then the macro can only jump from a closing brace to an opening brace.

anatoly77g:
If you save the recorded macro, it appears in this file (Windows 11): %AppData%\Notepad++\shortcuts.xml
Looking in this file sheds some light on what is going on:
"Go to matching brace" is recorded in a macro in the same way as "Search -> Go to... -> Offset". That is, when playing back the macro, the cursor will jump to a fixed offset in the editing buffer, regardless of what text there is now.

So the problem is not that it jumps only forward or only backward.
The problem is that it jumps to a fixed position, regardless of the text currently being edited.

victorel-petrovich:
@anatoly77g  
Perhaps this has changed a bit, but in N++ version 8.8.5, it records 2 actions, of which the first one is the one you mention and seems redundant, as detailed below: 
text file contents: 
```
 a(example) 

```
[...]

PR #14613, opened 2024-01-17
Fix macro recording problem for some commands

### LINKED ISSUE #5217: Move current line up/down in a macro shifts up/down 2 lines
### Description of the Issue

When running a macro, which you just made by recording, and it contains a MoveUpCurrentLine or a MoveDownCurrentLine, the result of this macro will be as if these 2 commands have been run twice, while the rest of the macro works as expected. While using the commands manually, shifts the line(s) only by 1, as expected.

1. have a tab open with multiple lines in it
2. have the cursor in any line (or select some line, effect does not depend on this)
3. record a macro which contains MoveUpCurrentLine or MoveDownCurrentLine
4. run the macro
5. see that the current/selected line(s) has been moved up/down 2 lines

### Expected Behavior

I expected the current/selected lines to only move 1 line, like it does when I run the command manually(either by selecting from the edit menu, or by pressing [Ctrl]+[Shift]+[Up/Down])

### Actual Behavior

every MoveDownCurrentLine which is contained in a recorded macro results in the selection having shifted 2 lines down
every MoveUpCurrentLine which is contained in a recorded macro results in the selection having shifted 2 lines up

### Debug Information

Notepad++ v7.5.9   (32-bit)
Build time : Oct 14 2018 - 15:02:52
Admin mode : ON
Local Conf mode : OFF
OS : Windows 7 (64-bit)
Plugins : NppConverter.dll NppExport.dll CodeAlignmentNpp.dll ComparePlugin.dll HexEditor.dll NotepadStarterPlugin.dll NppTextFX.dll NppToolBucket.dll OpenSelection.dll PluginManager.dll Python Indent.dll regrexplace.dll ReloadButton.dll XMLTools.dll

### LINKED ISSUE #14634: Cut and paste in macro doesn't work properly
### Description of the Issue
Since version 8.6.1, macros using cut and paste fail.

I was using versions 8.6.1 and 8.6.2, but I had to downgrade to 8.6, since macros were needed to a specific job and they were almost unusable.

1. Start recording a Macro;
2. Cut (Ctrl+X) an entire line of text or just some part of a word;
3. Paste (Ctrl+V) it at a blank line at the end of the document;
4. Stop recording;
5. Put the caret at the beginning of another line of text and select Playback (Ctrl+Shift+P), in order to repeat the work.
6. You will see the content will NOT be pasted at the end of the document.

### Expected Behavior
The macro should work as expected. I've been using Notepad++ for many years, and have never seen this cut/paste macro issue.

### Actual Behavior
Cut and paste in macro doesn't work properly, as described above.

### Debug Information

I saw this issue extensively in Notepad++ v8.6.2 (64-bit). Is this really a bug? Is there any workaround?
(Meanwhile, I'll stick with v8.6)
Thanks!

anatoly77g:
Fixed #5217, fixed #14634
When recording a keyboard macro, some commands are added to the macro twice. When later playing back the macro, they happen twice, which leads to incorrect edits.

Fixed commands:
IDM_EDIT_CUT (Cut)
IDM_EDIT_INS_TAB (Increase Line Indent)
IDM_EDIT_RMV_TAB (Decrease Line Indent)
IDM_EDIT_LINE_UP (Move Up Current Line)
IDM_EDIT_LINE_DOWN (Move Down Current Line)

Not fixed: issue #13722:
IDM_SEARCH_GOTOMATCHINGBRACE (Go to Matching Brace)


MarkusBodensee:
See also issue #14634

donho:
@anatoly77g 
The following commands are removed from recording in your PR: 
IDM_EDIT_INS_TAB (Increase Line Indent)
IDM_EDIT_RMV_TAB (Decrease Line Indent)

Could you provide the way for reproduce the duplication please?

PowerEditor/src/NppCommands.cpp
```diff
@@ -4193,7 +4193,6 @@ void Notepad_plus::command(int id)
 			case IDM_FILE_RELOAD:
 			case IDM_EDIT_UNDO:
 			case IDM_EDIT_REDO:
-			case IDM_EDIT_CUT:
 			case IDM_EDIT_COPY:
```
donho, line 4196: IDM_EDIT_COPY can also be removed

donho (changes requested):
Remove IDM_EDIT_COPY please.

anatoly77g:
> @anatoly77g The following commands are removed from recording in your PR: IDM_EDIT_INS_TAB (Increase Line Indent) IDM_EDIT_RMV_TAB (Decrease Line Indent)
> 
> Could you provide the way for reproduce the duplication please?

Start with an empty buffer.
1. Press 'a' and Enter.
2. Press Shift + Up to select the line with 'a' in it.
3. Press Ctrl + Shift + R to start recording a macro.
4. Choose from the menu: Edit - Indent - Increase Line Indent.
5. Press Ctrl + Shift + R to stop the recording.
6. Press Ctrl + Shift + P to play the recorded macro.
7. The line is indented twice.

I now see that my change doesn't fix this "Increase Line Indent" properly — it spoils the macro recording in other scenarios involving indentation. Probably worth making a separate issue for this; please confirm!
Meanwhile, I remove all changes related to indentation.

anatoly77g (commented):
Finished my review

PowerEditor/src/NppCommands.cpp
```diff
@@ -4208,6 +4207,8 @@ void Notepad_plus::command(int id)
 			case IDM_SEARCH_CLEAR_BOOKMARKS:
 			case IDM_SEARCH_INVERSEMARKS:
 			case IDM_EDIT_SELECTALL:
+			case IDM_EDIT_INS_TAB:
+			case IDM_EDIT_RMV_TAB:
```
donho, line 4211: OK I can reproduce the duplication.
But why IDM_EDIT_INS_TAB & IDM_EDIT_RMV_TAB can be removed but not always?

anatoly77g:
I believe commenting-out is a poor programming practice, so I prefer doing it more explicitly.
I suggest creating a dedicated do-nothing case in the switch statement, with comments which explain what's going on, and move all related commands to it.
Not sure this is acceptable as I am not familiar with pull-request rules in this project.
Please express your opinion.

donho:
> I suggest creating a dedicated do-nothing case in the switch statement, with comments which explain what's going on, and move all related commands to it.

It's a better idea indeed. OK let's do it this way.

donho (approved):

anatoly77g:
Behavior of "increase/decrease indentation" commands depends on what is selected. At certain conditions "depart from Scintilla behavior and do it our way", as some comment mentions. So correct recording of indentation commands in a macro requires some logic — sometimes, record the menu-command, other times record the Scintilla-command.
I guess several other commands have this problem; it may require some systematic approach.

donho:
@anatoly77g 

victorel-petrovich:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14613#issuecomment-1925676045
> I now see that my change doesn't fix this "Increase Line Indent" properly — it spoils the macro recording in other scenarios involving indentation. Probably worth making a separate issue for this; please confirm!

Here is the issue, https://github.com/notepad-plus-plus/notepad-plus-plus/issues/16960 ,  together with a case where removing the menu actions from recording would not work.

Issue #13908, opened 2023-07-18
Function outline does not recognize async Python functions


JC3:
### Description of the Issue

The function outline view does not recognize async Python functions. Note `example2` is not in the outline:

![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/5651184/72bab639-939b-44c4-acee-b37c47835b67)

### Steps to Reproduce the Issue

1. Create a Python source file with a top level async function
2. View -> Function List

### Expected Behavior

All top-level functions appear in function outline.

### Actual Behavior

Async functions do not appear in function outline.

### Debug Information

```Notepad++ v8.5.4   (64-bit)
Build time : Jun 17 2023 - 20:42:45
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Program Files\Notepad++\change.log" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 22H2
OS Build : 19045.3208
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    HexEditor (0.9.12)
    JSMinNPP (1.2205)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExec (0.8.2)
    NppExport (0.4)
    NppSaveAsAdmin (1.0.211)
    PreviewHTML (1.3.2)
    XMLTools (3.1.1.13)
```

sherdim:
fast patch in config file  `\functionList\python.xml`,  from line 32:
As a bonus it also corrects ambiguity with colon in the typed arguments (check a dollar sign - symbol.

                         <function
				mainExpr="^(async )?def\x20\K.+?(?=:$)"
                         >
				<functionName>
					<nameExpr expr=".*" />
				</functionName>
			</function>

PR #14628, opened 2024-01-23
Enhance Function List for Python

sherdim:
support for async def
resolves issue #13908


donho (changes requested):
The PR for the enhancement of Function List parser should also include unit test. Please refer [here](https://npp-user-manual.org/docs/function-list/#contribute-your-new-or-enhanced-parser-rule-to-the-notepad-codebase) for more information. 

donho:
Close due to oo answer from the PR author.
I'll complete the task.

donho:
Tested, the PR doesn't cause any regression so far.
I'll test it more and add the Unit test for it.

Issue #10746, opened 2021-11-04
Enhancement: Associate the .mjs file extension with JavaScript


startrekdude:
### Description of the Issue

Associate the `.mjs` file extension, used by ECMAScript 6 module files, with JavaScript. This would enable syntax highlighting by default when opening a `.mjs` file.

The `.mjs` extension is recommended for modules by the [V8 project](https://v8.dev/features/modules#mjs).

### Steps to Reproduce the Issue

1. Open a file with the `.mjs` file extension.

### Expected Behavior

The file is displayed with JavaScript syntax highlighting.

### Actual Behavior

The file is displayed without syntax highlighting.

### Debug Information

```
Notepad++ v8.1.9   (64-bit)
Build time : Oct 21 2021 - 23:37:53
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Education (64-bit) 
OS Version : 2009
OS Build : 19043.1288
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 
```

termdew:
I'll have a look at it.

htho:
Until it is fixed, this is how to manually set the extension:

In the menu go to: "Settings" -> "Style Configurator"
Select "JavaScript" (not "JavaScript (embedded)")) and add the extension in the "User extension" edit box.
To add multiple extensions separate them using Space.

You then need to re-open your file to see the syntax highlighting applied.

Thanks to: https://superuser.com/questions/40876/assigning-custom-extensions-to-a-languages-syntax-highlighting-in-notepad

ArkadiuszMichalski:
More info:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js

PR #14644, opened 2024-01-28
Update extensions for the "web script" file associations

vlakoff:
Add `.mjs` extension to the "web script" file associations.

Related: #10746 (fixed by #12704).

Refs: [JavaScript modules#Aside — .mjs versus .js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js) on MDN

Side notes:
* `.jsm` seems to be a legacy extension for the same thing, dating back from when the module specs weren't fixed yet. I haven't looked into it further.
* `.jsp` is unrelated to JavaScript, it's [JavaServer Pages](https://en.wikipedia.org/wiki/Jakarta_Server_Pages) (basically Java embedded in HTML content, just like PHP embedded in HTML content in `.php` files).

### DIFF
--- a/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
+++ b/PowerEditor/src/MISC/RegExt/regExtDlg.cpp
@@ -25,7 +25,7 @@ const TCHAR* nppBackup = TEXT("Notepad++_backup");
 const TCHAR* nppDoc    = TEXT("Notepad++ Document");
 
 const int nbSupportedLang = 10;
-const int nbExtMax = 27;
+const int nbExtMax = 28;
 const int extNameMax = 18;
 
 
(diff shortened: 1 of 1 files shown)


vlakoff:
Force-pushed to increase `nbExtMax` in order to fix this error:

> error C2078: too many initializers

vlakoff:
Added a commit to remove the `.phtml` file extension.

As noted in this [answer on Stack Overflow](https://stackoverflow.com/questions/11859015/what-is-phtml-and-when-should-i-use-a-phtml-extension-rather-than-php/11859122#11859122):

> `.phtml` was the standard file extension for PHP 2 programs. `.php3` took over for PHP 3. When PHP 4 came out they switched to a straight `.php`.

As a reminder, `.phpN` file extensions are still used by some web hosting providers, to let the user indicate which PHP major version should execute the script.

PowerEditor/src/MISC/RegExt/regExtDlg.cpp
```diff
@@ -48,8 +48,8 @@ const TCHAR defExtArray[nbSupportedLang][nbExtMax][extNameMax] =
 	{TEXT("web script"),
 		TEXT(".html"), TEXT(".htm"), TEXT(".shtml"), TEXT(".shtm"), TEXT(".hta"),
 		TEXT(".asp"), TEXT(".aspx"),
-		TEXT(".css"), TEXT(".js"), TEXT(".json"), TEXT(".jsm"), TEXT(".jsp"),
-		TEXT(".php"), TEXT(".php3"), TEXT(".php4"), TEXT(".php5"), TEXT(".phps"), TEXT(".phpt"), TEXT(".phtml"),
```
donho, line 52: .phtml "was" the standard file extension for PHP 2 programs, doesn't mean it doesn't exist anymore.
I could imagine there are still some servers use php2 so some phtml files still exist. 
OTOH, there's no reason to remove it if we keep php3 & php4 which are taken over by php5.

donho (changes requested):
Restore .phtml please.

PowerEditor/src/MISC/RegExt/regExtDlg.cpp
vlakoff: Come on, PHP 2 has been released on 1997! It's so old it's not even listed on the [Unsupported Historical Releases
](https://www.php.net/releases/index.php) page lol.

Here I'm just applying the [Less is more](https://en.wikipedia.org/wiki/Less_is_more) principle, removing a tiny bit of low-hanging bloat, to improves usability of the relevant parts.

donho: My point is phtml could still be used.

donho:
Create a new PR if you want to add only .jsm

vlakoff:
New PR: #14689

PR #14654, opened 2024-01-31
Fix 'Open File' command when tab characters are used on the same line

### LINKED ISSUE #14543: Open File: wrong determination of the caret position
### Description and Steps to Reproduce the Issue

I have a line of text:
(tab)(tab)(tab)tst.txt
three tab characters and the file name (the file is in the same directory as the file I'm editing).

I want to open the file with the context menu and the option 'Open File'.
I can't open the file if I place the caret on the file name, but I can do that if I place the caret between the first and the second tab.

### Analysis

I think the problem is in NppBigSwitch.cpp
...
case NPPM_GETFILENAMEATCURSOR:
...
col = _pEditView->getCurrentColumnNumber();
This method uses SCI_GETCOLUMN to determine the caret position in the string.
We have tabs as single characters in the string, but SCI_GETCOLUMN takes into account how many 'spaces' every tab (on the left side of the caret) takes.

SCI_GETCOLUMN(position pos) → position
This message returns the column number of a position pos within the document taking the width of tabs into account. This returns the column number of the last tab on the line before pos, plus the number of characters between the last tab and pos. 

Guessing where to click could be fun for short lines, but for long ones with many tab characters... ;)

### Debug Information
Notepad++ v8.6   (32-bit)
Build time : Nov 23 2023 - 16:56:22
Path : C:\b\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF

alankilborn:
Fix #14543

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1065,11 +1065,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		{
 			constexpr int strSize = CURRENTWORD_MAXLENGTH;
 			TCHAR str[strSize]{};
-			TCHAR strLine[strSize]{};
-			size_t lineNumber = 0;
-			intptr_t col = 0;
 			int hasSlash = 0;
-			TCHAR *pTchar = reinterpret_cast<TCHAR *>(lParam);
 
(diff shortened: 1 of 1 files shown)


alankilborn:
Note: I fixed the bug AND moved some declarations around to be better scoped.  If the scoping changes aren't wanted, I'll certainly back them out.

Issue #14568, opened 2024-01-09
Can’t use Ctrl+V to paste text into the text fields of forms for certain plugins


donho:
Ref:
1. https://community.notepad-plus-plus.org/topic/25315/notepad-v8-6-1-release/2?_=1704770077793
2. https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14557#issuecomment-1879914173

### Description of the Issue
Can’t use Ctrl+V to paste text into the text fields of forms for certain plugins in v8.6.1

### Steps to Reproduce the Issue

1. Install [JsonTools 6.1.1 ](https://github.com/molsonkiko/JsonToolsNppPlugin) in Notepad++ v8.6.1
2. Start Notepad++ and  launch JsonTools GUI
4. Try to paste a text by using Ctrl-V

### Expected Behavior
Can paste text by using Ctrl-V

### Actual Behavior
Cannot paste text by using Ctrl-V

### Debug Information
```
Notepad++ v8.6.1   (64-bit)
Build time : Jan  4 2024 - 05:01:52
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 Home (64-bit)
OS Version : 23H2
OS Build : 22631.2861
Current ANSI codepage : 1252
Plugins : 
    BetterMultiSelection (1.5)
    mimeTools (3)
    NppConverter (4.5)
    NppExport (0.4)
    PythonScript (2)
    VisualStudioLineCopy (1.0.0.2)
```

Coises:
I mentioned this in the #14557 discussion, but one possible way to fix this would be to put code like this:

	for (HWND msgAncestor = msg->hwnd; msgAncestor && msgAncestor != _hSelf; msgAncestor = ::GetParent(msgAncestor))
	{
		if (WC_DIALOG == MAKEINTATOM(GetClassLong(msgAncestor, GCW_ATOM))) return ::IsDialogMessage(msgAncestor, msg);
	}

into [Notepad_plus_Window::isDlgsMsg](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/eb0922c2233bff484f91e7f1cf26446e6a9bc2a5/PowerEditor/src/Notepad_plus_Window.cpp#L422).

[...]

PR #14660, opened 2024-02-01
Make copy/cut line while no selection optional

### LINKED ISSUE #14638: In v8.6.2 (32-bit), when no text selection is made, Ctrl-C copies and Ctrl-X deletes the whole current line
Is there a workaround for this?

donho:
Fix #14638


donho:
@molsonkiko 
With this PR, there's always the same issue that you describe [here ](https://community.notepad-plus-plus.org/topic/25315/notepad-v8-6-1-release/2?_=1706795482129) in JsonTools.
To reproduce: 
1. Uncheck "Enable Copy/Cut Line without selection" checkbox in **Editing** section of Preferences.
2. Make sure there's no selection in the current editor
3. Launch "Open JSON tree viewer", and try to Cut "@" symbol on the top left text field with Ctrl-X

You have mentioned ***[NavigateTo 2.7 ](https://github.com/young-developer/nppNavigateTo), also a C# plugin, has this bug.*** in above link. Could you provide the way to reproduce the same issue in NavigateTo  plugin please?

molsonkiko:
@donho 

Thanks for asking for my input! Here are my findings:

__Enable Copy/Cut Line without selection *disabled*__ 
Copy/cut __does not work__ in the text boxes for JsonTools and NavigateTo
__Enable Copy/Cut Line without selection *enabled*__
Copy/cut __does work__ in the text boxes for JsonTools and NavigateTo.

I installed the [GCC_x86_64.Release GH Actions artifact](https://github.com/donho/notepad-plus-plus/actions/runs/7741238289), and added it to the ProgramFiles folder for my existing Notepad++ 8.6.2 64bit installation.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -2567,9 +2567,12 @@ void Notepad_plus::checkClipboard()
 {
 	bool hasSelection = _pEditView->hasSelection();
 	bool canPaste = (_pEditView->execute(SCI_CANPASTE) != 0);
-	//enableCommand(IDM_EDIT_CUT, hasSelection, MENU | TOOLBAR);
-	//enableCommand(IDM_EDIT_COPY, hasSelection, MENU | TOOLBAR);
 
+	if (!NppParameters::getInstance().getSVP()._lineCopyCutWithoutSelection)
```
molsonkiko, line 2571: I think you need to get rid of this check.

Based on my observation that copy/cut in the NavigateTo and JsonTools windows __only works when text is selected__ after this commit, the problem would seem to be that when you disable the menu item here, there's never an opportunity for the WM_CUT message to get sent to plugin textboxes.

donho:
@molsonkiko 
OK, I think I know what's going on finally.
For C# controls in dockable dialog, their common shortcuts (Ctrl-C, Ctrl-X, Ctrl-V, Ctrl-A, etc...) will be hijacked by Notepad++.
You can test on the (only) first commit - when you remap Ctrl-X to other shortcut for Cut command, the Ctrl-X works on  C# controls of dockable dialog, even Cut command is disabled for system.

To remedy the problem (my 2nd commit), Cut & Copy commands are disabled only on toolbar but not in the menu so the shortcuts can always passed by system - you can try the latest commit, it should be OK for both JSONTools and NavigatorTo plugins, though the status are not synchronized between menu and toolbar for these 2 commands :(

donho:
@Coises You might be interested in above explanation.

Coises:
@donho:
> You might be interested in above explanation.

I am reasonably certain this has nothing to do with C# or dockable dialogs *per se*; the relevant factor is the use of a non-modal dialog that is not registered using NPPM_MODELESSDIALOG. Why this is apparently common for C# plugins and/or dockable dialogs is not something I can guess, since I haven’t worked with either. Without NPPM_MODELESSDIALOG, [TranslateAccelerator](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/winmain.cpp#L720) picks up every key combination assigned to a Notepad++ function (but *not* ones assigned to Scintilla functions).

You can even swap keys around in 8.6.2 — for example, make Ctrl+X copy and Ctrl+C cut in Notepad++ — and those changes will carry over to unregistered modeless dialogs in plugins; modal dialogs and dialogs registered with NPPM_MODELESSDIALOG are unaffected.

Falsifying the enabled state of menu commands is not going to go over well.

At root, there is a logical inconsistency here. A modeless dialog that does not use NPPM_MODELESSDIALOG is, in effect, saying: I want Notepad++ accelerator keys (but not Scintilla accelerator keys) to work as they are defined to Notepad++ without first returning keyboard focus to the main Notepad++ window... except for when I don’t want that. Until Notepad++ took over Ctrl+X/C/V from Scintilla, this was apparently not a problem for many plugins.

If you really want to make exceptions, I think the place to do it is in [isDlgsMsg](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/Notepad_plus_Window.cpp#L422). Hard-code a list of key combinations (e.g., Ctrl+X/C/V) that are never translated when sent to a non-Notepad++ window regardless of whether it is registered as a modeless dialog or not.

rdipardo:
> You can even swap keys around in 8.6.2 — for example, make Ctrl+X copy and Ctrl+C cut in Notepad++ — and those changes will carry over to unregistered modeless dialogs in plugins; modal dialogs and dialogs registered with NPPM_MODELESSDIALOG are unaffected.

Just to be clear, by "unregistered modeless dialogs" we are talking specifically about *docked* plugin dialogs? As far as I know, sending NPPM_DMMREGASDCKDLG at any time will cancel the effect of NPPM_MODELESSDIALOG, no matter what order you send both messages in.

Debug any plugin that implements its own window procedure and you will see `::DefWindowProc` near the bottom of the call stack. Under normal conditions, a plugin receives messages directly from the operating system; the particular FFI used to wrap `::DefWindowProc` makes no difference. <kbd>Ctrl+C/X/V</kbd> will "just work" is the plugin's native GUI library has already mapped them to the OS's universal accelerators. The host application's custom mappings are never consulted.

Sending NPPM_DMMREGASDCKDLG effectively makes the plugin dialog a "child" of the Notepad++ application window. A single docking manager looks after the "internal plugin" dialogs like the search results window, as well as any external ones created by third-party modules. So the "a logical inconsistency" is perhaps the overbearing control assumed by the docking manager, which does everything from drawing the dialog window (at times [confounding the OS](https://community.notepad-plus-plus.org/post/92248)) to playing traffic cop with the message loop.

Coises:
@rdipardo:

> Just to be clear, by "unregistered modeless dialogs" we are talking specifically about docked plugin dialogs? As far as I know, sending NPPM_DMMREGASDCKDLG at any time will cancel the effect of NPPM_MODELESSDIALOG, no matter what order you send both messages in.

No, I mean any *non-modal* dialog that wasn’t the subject of an NPPM_MODELESSDIALOG call. NPPM_MODELESSDIALOG has one effect: it [adds the HWND of the dialog to the list of modeless dialogs](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/NppBigSwitch.cpp#L1922) which is consulted in [isDlgsMsg](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/Notepad_plus_Window.cpp#L422) to determine whether to [bypass TranslateAccelerator](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/winmain.cpp#L718) in the Notepad++ message loop. I don’t see any evidence that NPPM_DMMREGASDCKDLG does something to remove the HWND from that list, but I wouldn’t say I’m certain.

> any plugin that implements its own window procedure

I might misunderstand, but I think you are confusing the window procedure with the message loop. Modal windows have a separate message loop — that’s what makes them modal — otherwise, [all messages pass through the message loop in Notepad++’s winmain](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/winmain.cpp#L712). Keyboard messages are dispatched to the window procedure of the focused window *unless TranslateAccelerator picks them off*.

rdipardo:
Some code to consider:

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f431a379671936ae611b8a231ee0a43c8a7fc0f9/PowerEditor/src/NppCommands.cpp#L380-L395

The finder dialog's edit control implicitly knows what to do with `WM_COPY`, but not with `IDM_EDIT_COPY`; the `else` branch [became necessary][0] because the finder no longer responds to <kbd>Ctrl+C</kbd> without it.

The finder is a typical child window, as it follows the dark mode theme, achieved by automatically sub-classing all the windows to which the application has a handle. The same applies to docked plugin dialogs (provided the DWS_USEOWNDARKMODE flag was not used to construct them).

If the finder gets a special exception, maybe third-party dialogs can, too, as was [tentatively suggested](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14568#issuecomment-1882371206) before, or a whitelist of some universal key combos, as suggested above.


[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/0978b2e29f21e7797f488e9e6d2e85763b25a40

donho:
@Coises 
> I am reasonably certain this has nothing to do with C# or dockable dialogs per se; the relevant factor is the use of a non-modal dialog that is not registered using NPPM_MODELESSDIALOG.

I think you're right about it:
I just removed `::SendMessage(_hParent, NPPM_MODELESSDIALOG, MODELESSDIALOGADD, reinterpret_cast<WPARAM>(_hSelf));` from the plugin NppConverter for testing - and the text field of NppConverter has the exact same behaviour as text control of JsonTools & NavigateTo.

It's definitely a plugin issue. (Plugin Modeless dialog should call NPPM_MODELESSDIALOG with MODELESSDIALOGADD)

donho:
The latest commit freezes (disables) the Cut/Copy menu item while copy/cut line without selection is disabled and there's no selection.
As a result, the cut/copy problem in text field of JsonTools & NavigateTo will be reproduced in the latest commit.

@molsonkiko 
I've checked JsonTools v6.1.1 code, [NPPM_MODELESSDIALOG](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h#L93) is not called with your plugins - you should retrieve Window handle of TreeViewer form and call NPPM_MODELESSDIALOG with MODELESSDIALOGADD and retrieved HWND of TreeViewer form in the constructor.

PowerEditor/src/Notepad_plus.cpp
donho: Please check my comment:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14660#issuecomment-1924391237

molsonkiko:
> I've checked JsonTools v6.1.1 code, [NPPM_MODELESSDIALOG](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h#L93) is not called with your plugins - you should retrieve Window handle of TreeViewer form, then call NPPM_MODELESSDIALOG with MODELESSDIALOGADD and retrieved HWND of TreeViewer form in the constructor.

I tried that before, during the original discussion about the copy/cut bugs after the 8.6.1 release, and it completely messed up all the keypress and keyup handlers in my plugin. If I register my forms with MODELESSDIALOGADD, all my internal logic about the order in which tab is supposed to navigate between fields is not respected, and I won't compromise that aspect of my user experience.

rdipardo:
@donho 
caa68f8 works for me 👍🏼 

BTW, I pasted into a docked plugin window while a dev build of caa68f8 was running. GDB (9.2) captured the following when breaking inside [this window procedure](https://github.com/rdipardo/nppFSIPlugin/blob/b7d9cd93b10670b7648288fee27a03a04d7cb221/Source/Plugin/Src/FSIWrapper.pas#L390):

<details>
<summary>Thread 1 hit Breakpoint 2, editorWndProc (this=0x2a1c0dba690, Message=...) at Src/FSIWrapper.pas:394</summary>
<pre>
NPPFSIPlugin.dll!editorWndProc(TFSIViewer this, TMessage Message) (c:\Users\Rob\dev\nppFSIPlugin\Source\Plugin\Src\FSIWrapper.pas:394)
NPPFSIPlugin.dll!DELIVERMESSAGE(TOBJECT TARGET, formal AMESSAGE) (c:\Users\Rob\scoop\apps\lazarus\3.0\lcl\lclmessageglue.pas:114)
NPPFSIPlugin.dll!DOWINDOWPROC(TWINDOWPROCHELPER this) (c:\Users\Rob\scoop\apps\lazarus\3.0\lcl\interfaces\win32\win32callback.inc:2642)
NPPFSIPlugin.dll!WINDOWPROC(QWORD WINDOW, LONGWORD MSG, INT64 WPARAM, INT64 LPARAM) (c:\Users\Rob\scoop\apps\lazarus\3.0\lcl\interfaces\win32\win32callback.inc:2807)
NPPFSIPlugin.dll!RICHEDITPROC(QWORD WINDOW, LONGWORD MSG, INT64 WPARAM, INT64 LPARAM) (c:\Users\Rob\dev\nppFSIPlugin\Source\RichMemo\win32\win32richmemo.pas:501)
user32.dll!USER32!CallWindowProcW (Unknown Source:0)
user32.dll!USER32!DispatchMessageW (Unknown Source:0)
user32.dll!USER32!SendMessageTimeoutW (Unknown Source:0)
ntdll.dll!ntdll!KiUserCallbackDispatcher (Unknown Source:0)
win32u.dll!win32u!NtUserMessageCall (Unknown Source:0)
user32.dll!USER32!SendMessageW (Unknown Source:0)
user32.dll!USER32!SendMessageA (Unknown Source:0)
NPPFSIPlugin.dll!GETTEXTLENGTH(^__vtbl_ptr_type self, QWORD RICHEDITWND) (c:\Users\Rob\dev\nppFSIPlugin\Source\RichMemo\win32\win32richmemoproc.pas:407)
NPPFSIPlugin.dll!GETTEXTLEN(^__vtbl_ptr_type self, TWINCONTROL AWINCONTROL, LONGINT ALENGTH) (c:\Users\Rob\dev\nppFSIPlugin\Source\RichMemo\win32\win32richmemo.pas:612)
NPPFSIPlugin.dll!GETTEXTLEN(TWINCONTROL this) (c:\Users\Rob\scoop\apps\lazarus\3.0\lcl\include\wincontrol.inc:8315)
NPPFSIPlugin.dll!IsConsoleRunning(TFSIViewer this, Boolean echo) (c:\Users\Rob\dev\nppFSIPlugin\Source\Plugin\Src\FSIWrapper.pas:409)
NPPFSIPlugin.dll!SendText(TFSIViewer this, WideString selText, Boolean addDelimiter, Boolean appendEditor, Boolean echoErr) (c:\Users\Rob\dev\nppFSIPlugin\Source\Plugin\Src\FSIWrapper.pas:254)
NPPFSIPlugin.dll!SendSelectedTextInNPPToFSI(TFrmFSIHost this) (c:\Users\Rob\dev\nppFSIPlugin\Source\Plugin\Forms\FSIHostForm.pas:163)
NPPFSIPlugin.dll!SendText(TFsiPlugin this) (c:\Users\Rob\dev\nppFSIPlugin\Source\Plugin\Src\FSIPlugin.pas:457)
NPPFSIPlugin.dll!_sendText() (c:\Users\Rob\dev\nppFSIPlugin\Source\Plugin\Src\FSIPlugin.pas:105)
[Unknown/Just-In-Time compiled code] (Unknown Source:0)
</pre>
</details>

Also tied a release build of 8.6.1 just to compare. The call stack was identical 🤷🏼 

I expect a C# plugin's call stack would differ only in terms of more indirection, i.e., `mscoree.dll` would take the place of the individually named system modules which it wraps. Object Pascal is "unmanaged" and binary compatible with plain C, so it calls the APIs directly.

donho:
@molsonkiko 
> If I register my forms with MODELESSDIALOGADD, all my internal logic about the order in which tab is supposed to navigate between fields is not respected, and I won't compromise that aspect of my user experience.

It sounds not good but I still suggest using NPPM_MODELESSDIALOG and changing your plugin's internal logic. Or you would use [the hack you have published](https://community.notepad-plus-plus.org/topic/25315/notepad-v8-6-1-release/3?_=1706919117240). However in the feature Notepad++ might manage directly more shortcuts like Ctrl-Z, Ctrl-A and other - then you might have to add more hacks for your plugin. It's your call to choose the best way for JsonTools anyway. Good luck.

molsonkiko:
> It sounds not good but I still suggest using NPPM_MODELESSDIALOG and changing your plugin's internal logic
@donho 
I respect your decisions, and I appreciate your continued work to maintain Notepad++. I suppose I will try to adopt to the new situation, but I want to give you a warning: __many plugin maintainers will likely simply give up rather than deal with the escalating demands you are placing on them.__ If you make Notepad++ a less hospitable place to build plugins, do not be surprised if the plugin ecosystem suffers as a result.

donho:
@molsonkiko 
>  If you make Notepad++ a less hospitable place to build plugins, do not be surprised if the plugin ecosystem suffers as a result.

A less hospitable place is (for me) is a project where the application doesn't care about the contract with plugins.
The contract here is API for the plugins, which is clear and always respected by Notepad++, without regression.
BTW, NPPM_MODELESSDIALOG has existed since long time ago, and its documentation is always explicit.
It's not reasonable to ask application not doing the right thing, instead of following official way (which is documented) to achieve your task.

SN-CH:
@donho I must admit to be not tech-savvy enough to follow the discussion in this thread. So, I would appreciate if you could answer the following question in less technical terms: is it possible or is it not to disable cutting/copying the whole paragraph when no text ist selected (what could have been achieved with this deprecated disableLineCopyCutDelete.xml file in 8.6.1)? If yes, how to activate this function? If not, is there any chance it will be implemented in the future (in any form: a file, a GUI toggle, a parameter in some xml-config, whatever)?

donho:
@SN-CH 
In this PR, a GUI option has been added to disable cutting/copying the whole paragraph feature.
If you are using cutting/copying the whole paragraph feature in v8.6.2, then there will be no change to you when you upgrade your Notepad++ to v8.6.3.

alankilborn:
> cutting/copying the whole paragraph

Wondering about the use of "paragraph" here...does it mean something different than "line"?

donho:
@alankilborn 
I think by "paragraph" @SN-CH meant **"entire line"**. I just reused this term to make the discussion more comprehensive.

SN-CH:
@donho Thank you very much for the explanation (and even more so for the work you have been doing). By the way, I can confirm the toggle works perfectly.

@alankilborn Yes, «paragraph» is the another name for «entire line» in this context.

molsonkiko:
I finally figured out an [acceptable solution that doesn't break tab ordering for C# plugins](https://community.notepad-plus-plus.org/topic/25459/generalized-fix-for-ctrl-c-x-woes-in-c-plugins-beginning-in-notepad-8-6-1). Many thanks to Coises for helping with the tab ordering issue, and apologies to Don Ho for my somewhat overwrought and misguided complaints.

PR #14671, opened 2024-02-04
 tag completion html.xml

Enjoyop:
First of all, I would like to thank everyone who contributed to a simple and beautiful editor that I have been using since my childhood.

I also know that this is not a full IDE, but it is still used in web development in a few places. For web programming maybe you can define a few variables. except just "KeyWord name".

I don't know how accurate my comment is since I don't have a full command of C++. Unfortunately, I won't be able to write what I want to say right now into src as code, since the code completion part is a bit dirty and there is no base class structure.

Firstly:
We can't call the "!DOCTYPE html" keyword anyway.
! Dictionaries starting with characters like < > do not appear in code completion. On the other hand, why are there CSS codes in the HTML syntax? We either use them as embedded styles or just use them in head <style></style>, we can read them from the css.xml file. On the other hand, we can write the labels ourselves and close them automatically. for example <meta> or <h1></h1> Of course, you need to define the keyword type here <meta></meta>, it would be meaningless. On the other hand, the ":" character comes after the CSS properties, then ";" for the value and ending. This should also be defined with keyword types. When we write an HTML code, we need a few basic codes. can call this instead of DOCKTYPE.
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1.0"> <title></title>
</head>
<body>

</body>
</html>

In other words, I have never seen 2 different bodies or heads defined in one HTML file. It should be able to give a warning when such a thing happens.

Finally, you can do this in all languages.
Sometimes, instead of searching for what a tag, method or function does, it can be defined into a description variable that can be translated into any language. Thank you for reading them all.

### DIFF
--- a/PowerEditor/installer/APIs/html.xml
+++ b/PowerEditor/installer/APIs/html.xml
@@ -204,14 +204,14 @@
 		<KeyWord name="grad" />
 		<KeyWord name="gray" />
 		<KeyWord name="green" />
-		<KeyWord name="h1" />
-		<KeyWord name="h2" />
-		<KeyWord name="h3" />
-		<KeyWord name="h4" />
-		<KeyWord name="h5" />
-		<KeyWord name="h6" />
-		<KeyWord name="head" />
-		<KeyWord name="header" />
+		<KeyWord name="h1></h1>" />
+		<KeyWord name="h2></h2>" />
+		<KeyWord name="h3></h3>" />
+		<KeyWord name="h4></h4>" />
+		<KeyWord name="h5></h5>" />
+		<KeyWord name="h6></h6>" />
+		<KeyWord name="head></head>" />
+		<KeyWord name="header></header>" />
 		<KeyWord name="headers" />
 		<KeyWord name="hebrew" />
 		<KeyWord name="height" />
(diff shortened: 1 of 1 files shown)


chcg:
See https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/7772518691/job/21195034827?pr=14671. XML is not valid as escaping is missing.

Enjoyop:
image
There seems to be no problem.

alankilborn:
Also, this PR is not likely to be accepted because the protocol is to first create an issue describing the problem, THEN fix it with a PR.  See CONTRIBUTING.md.

donho:
1. [All Pull Requests, except for translations and user documentation, need to be attached to a issue on GitHub](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#pull-requests).
2. Your XML file is not valid: See https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/7772518691/job/21195034827?pr=14671. XML is not valid as escaping is missing.

PR #14685, opened 2024-02-06
Fix Replace All crash & performance issue

### LINKED ISSUE #14630: app-crash after find and replace in a large file
### Description of the Issue
I have a csv file with over 500,000 lines (file size 50MB) and 20 comma's on each line. After a find and replace commas to semicolons NPP crashes after 5 minutes
(Windows notepad did the trick in 5 seconds)

### Debug Information
Notepad++ v8.6.2   (32-bit)
Build time : Jan 14 2024 - 02:18:41
Command Line : "D:\user\logins.txt"
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Build : 7601.0
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2)
    DSpellCheck (1.4.6)
    JSMinNPP (1.1903)
    mimeTools (3)
    NppConverter (4.5)
    NppExport (0.4)
    NppTextFX (0.2.6)
    PoorMansTSqlFormatterNppPlugin (1.6.10.1279)
    XMLTools (2.4.11)

donho:
Fix  #14630

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2737,11 +2737,18 @@ int FindReplaceDlg::processAll(ProcessOperation op, const FindOption *opt, bool 
 {
 	if (op == ProcessReplaceAll && (*_ppEditView)->getCurrentBuffer()->isReadOnly())
 	{
-		NativeLangSpeaker *pNativeSpeaker = (NppParameters::getInstance()).getNativeLangSpeaker();
+		NppParameters& nppParam = NppParameters::getInstance();
+		NativeLangSpeaker *pNativeSpeaker = nppParam.getNativeLangSpeaker();
 		generic_string msg = pNativeSpeaker->getLocalizedStrFromID("find-status-replaceall-readonly", TEXT("Replace All: Cannot replace text. The current document is read only."));
 		setStatusbarMessage(msg, FSNotFound);
 		return 0;
 	}
+
+	// Turn OFF all the notification of modification (SCN_MODIFIED) for the sake of performance
+	LRESULT notifFlag = (*_ppEditView)->execute(SCI_GETMODEVENTMASK);
+	(*_ppEditView)->execute(SCI_SETMODEVENTMASK, 0);
+
+
 
 	const FindOption *pOptions = opt?opt:_env;
 	const TCHAR *txt2find = pOptions->_str2Search.c_str();
@@ -2804,6 +2811,13 @@ int FindReplaceDlg::processAll(ProcessOperation op, const FindOption *opt, bool 
 	findReplaceInfo._endRange = endPosition;
 
 	int nbProcessed = processRange(op, findReplaceInfo, pFindersInfo, pOptions, colourStyleID);
+
+
+	// Turn ON the notifications after operations
+	(*_ppEditView)->execute(SCI_SETMODEVENTMASK, notifFlag);
+	if (op == ProcessReplaceAll && nbProcessed > 0) // All the notification of modification (SCN_MODIFIED) were removed during the operations, so we set modified status true here
+		(*_ppEditView)->getCurrentBuffer()->setModifiedStatus(true);
+
 
 	if (nbProcessed == FIND_INVALID_REGULAR_EXPRESSION)
 		return FIND_INVALID_REGULAR_EXPRESSION;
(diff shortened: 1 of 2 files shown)


xomx:
@donho 
Just tested, superb performance enhancement.
Are there any side-effects of that temporary notifications disabling?

alankilborn:
> Are there any side-effects of that temporary notifications disabling?

Well, plugins that act on "modified" notifies to do their "whatever" thing obviously aren't going to be doing it. :-)

But I'd say that's negligible in trade for substantial performance improvement.

donho:
@xomx 

>>    Are there any side-effects of that temporary notifications disabling?

> Well, plugins that act on "modified" notifies to do their "whatever" thing obviously aren't going to be doing it. :-)

What @alankilborn has said is pertinent... let's see if there any complaint from plugin authors, then we will see what we do about it.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -2805,6 +2812,13 @@ int FindReplaceDlg::processAll(ProcessOperation op, const FindOption *opt, bool
 
 	int nbProcessed = processRange(op, findReplaceInfo, pFindersInfo, pOptions, colourStyleID);
 
+
+	// Turn ON the notifications after operations
+	(*_ppEditView)->execute(SCI_SETMODEVENTMASK, notifFlag);
+	if (op == ProcessReplaceAll && nbProcessed > 0) // All the notification of modification (SCN_MODIFIED) were removed during the operations, so we set modified status true here
```
donho, line 2818: `NPPN_MODIFIEDREPLACEALL` could be added here for notifying the plugins "the current document has been changed after Replace All action".

donho:
@Coises 
> I hadn’t happened to do a Replace All in a file where my plugin’s Elastic tabstops function was affected since installing a version of Notepad++ with this change.

Could you create an issue with a step by step instructions for reproduce the bug?

>   advertised well before it’s implemented.

You're right about it. But I neglected the part of plugins being able to get Scintilla's notification. Sorry for neglecting it.

>  complemented with an NPPN notification indicating that multiple changes were (or at least may have been) made while SCN_MODIFIED notifications were suppressed, or some similar mechanism, so plugins know they need to rescan the entire file.

`NPPN_MODIFIEDREPLACEALL` could be added [here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14685/files#r1496765277) for notifying the plugins "the current document has been changed after Replace All action".

What do you think?

Coises:
@donho 
> Could you create an issue with a step by step instructions for reproduce the bug?

Issue #14767.
> 
> `NPPN_MODIFIEDREPLACEALL` could be added [here](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14685/files#r1496765277) for notifying the plugins "the current document has been changed after Replace All action".
> 
> What do you think?

I can think of two options, depending on whether you prioritize not breaking existing plugins, or being as “honest” as possible with plugins.

The notification you described would tell plugins exactly what happened and allow them to choose how to respond. However, existing plugins would be broken until and unless their authors update them.

Alternatively, you could “forge” the SCN_MODIFIED notifications that would occur if the entire document were deleted and replaced. (Save the old document length before you start; then after the replace is complete, send SC_MOD_BEFOREDELETE, SC_MOD_DELETETEXT, SC_MOD_BEFOREINSERT and SC_MOD_INSERTTEXT. A potential complication is that these should occur before Scintilla sends an SCN_UPDATEUI notification, and I don’t know if that is how it would happen.) That would avoid breaking existing plugins, at the expense of “lying” to plugins about what actually happened.

Much as I hate! hate! hate! the idea of breaking existing plugins (especially when — unlike the NPPM_MODELESSDIALOG fiasco –— they are already following the rules), if I had to choose, I guess I’d go for the NPPN message rather than forging Scintilla messages... there are too many unforeseeable ways the latter could come back to bite us.

rdipardo:
@Coises 

> I hate! hate! hate! the idea of breaking existing plugins 
 
Even if a kludgy workaround could avoid it now, the next breaking change is only around the corner. Scintilla is an *application* interface, not a plugin interface. SciTE's mainatiner had the wisdom to natively support [Lua scripting](https://www.scintilla.org/SciTELua.html) like many other extensible editors, [NotepadNext](https://github.com/dail8859/NotepadNext/tree/master/src/LuaBridge) included.

donho:
@Coises 
> I guess I’d go for the NPPN message rather than forging Scintilla messages... there are too many unforeseeable ways the latter could come back to bite us.

Amen!

PR #14689, opened 2024-02-07
Add .mjs extension to the "web script" file associations

vlakoff:
Add `.mjs` extension to the "web script" file associations.

Related: #10746 (fixed by #12704).

Refs: [JavaScript modules#Aside — .mjs versus .js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js) on MDN

Side notes:
* `.jsm` seems to be a legacy extension for the same thing, dating back from when the module specs weren't fixed yet. I haven't looked into it further.
* `.jsp` is unrelated to JavaScript, it's [JavaServer Pages](https://en.wikipedia.org/wiki/Jakarta_Server_Pages) (basically Java embedded in HTML content, just like PHP embedded in HTML content in `.php` files).


vlakoff:
While working on some project, I thought back of this ticket: we should add the `.cjs` file extension as well.

Refs https://github.com/nodejs/node/pull/50096.

PR #14692, opened 2024-02-07
fix cmdline & Preferences multi-inst inconsistency

### LINKED ISSUE #14691: possible overwriting of backups of unsaved tabs with non-unique filenames from "-multiInst" instances
Description:
https://community.notepad-plus-plus.org/topic/25434/bug-npp-loses-all-open-files-after-restarting-as-admin/17

The following is due to the "new x" tab filenames are not multi-instance unique.
The problem is only with the cmdline parameter "-multiInst", not with such setting in Preferences (which effectively disables possible N++ session backup feature in any non-first instance).

STR (using x64 v8.6.2 portable):

1)
- check Preferences > Backup > Remember current session for next launch & Enable session snapshot and periodic backup
- note how many seconds you have there in the Backup in every XXX seconds (default here is 7, I will use this number in my STR)
- set Preferences > Multi-Instance & Date > Muti-instance settings == Always in multi-instance mode
- close N++
- launch 1st instance of N++, there will be the "new 1" tab, edit its content e.g. by adding number "1"
- wait 7 secs, then check the backup created (a  .\\backup\\new 1@YYYY-MM-DD_HHMMSS) and its content (should be that "1")
- launch 2nd N++ instance, there will be the same "new 1" tab,  edit its content e.g. by adding number "2"
- wait another 7 secs, then check the original backup file (the same  .\\backup\\new 1@YYYY-MM-DD_HHMMSS)), its content remains that "1"

2)
- change the Preferences > Multi-Instance & Date > Muti-instance settings == Default (mono-instance)
- close N++
- repeat the remaining test as in the above first case, but this time launch all the instances of N++ with "-multiInst" cmdline param
- result: 2nd N++ instance will overwrite the backup of the 1st N++ instance with "2"

xomx:
Fix #14691 

Now both the cmdline "-multiInst" and the same setting in the Notepad++ Preferences will have the same effect on the backup feature.

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -527,11 +527,15 @@ int WINAPI wWinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE /*hPrevInstance
 	// override the settings if multiInst is choosen by user in the preference dialog
 	const NppGUI & nppGUI = nppParameters.getNppGUI();
 	if (nppGUI._multiInstSetting == multiInst)
-	{
 		isMultiInst = true;
-		// Only the first launch remembers the session
-		if (!TheFirstOne)
-			cmdLineParams._isNoSession = true;
+
+	// Only the first launch remembers the session in multiInst
+	if (isMultiInst && !TheFirstOne)
+	{
+		// this will also correctly disable the possible unsaved tabs filebuffer backup in any other Notepad++ multi-inst instance later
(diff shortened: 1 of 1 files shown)


donho:
@xomx 
The PR makes the regression of saving file with Admin, with default mono-instance ON.
The solution of issue #14691 should be https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14691#issuecomment-1935267457.
(I do belive this fix is after investigation of issue #14694, which is fixed by PR #14701)

PR #14721, opened 2024-02-13
Change colors for Mossylawn: selected text, bad brace, INI files, JSON, Python

### LINKED ISSUE #14720: Mossylawn theme- bad brace is nearly invisible, JSON STRINGEOL and ESCAPE SEQUENCE styles hard to see
### Description of the Issue

1. Bad braces are very hard to see, because they are too dark
image
2. Selected text has poor contrast with the default background for the current line.
image
3. For the JSON lexer, the `STRINGEOL` color has rather poor contrast, and the `ESCAPE SEQUENCE` color is so bright it kind of hurts the eyes on unselected lines.
image

1. Switch to MossyLawn theme.
2. Create a document containing only `[foo` and observe that the `[` is too dark to see well when the caret is next to it.
3. Select the second `foo` of the document `foofoo` and observe that contrast is mediocre.
4. Create a document containing the text
    ```
    "foo

    "bar\r \t"
    ```
5. Set the lexer language to JSON
4. Select the line containing `"foo` and observe that the contrast is poor.
6. Also observe that the `\r` and `\t` are so bright that they are rather hard and jarring to look at.

### Proposed fix

I will make a PR with proposed changes that would remedy these issues, which I will link here.

### Debug Information
```
Notepad++ v8.6.2   (64-bit)
Path : C:\Program Files\Notepad++ 8p6p2\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.3930
Current ANSI codepage : 1252
Plugins : 
    CSharpPluginPack (0.0.2)
    JsonTools (7)
    mimeTools (3)
    NavigateTo (2.7)
    NppConverter (4.5)
    NppExport (0.4)

```

molsonkiko:
This would fix #14720 .

- __Improved bad brace visibility (it was nearly invisible before).__ I consider the other changes negotiable, but this one is really important to me.
- Increased diversity of coloring for markers (they had good contrast before, but now each marker is IMO more visually distinct from the other markers)
- Improved contrast of selection background with selected line background (previously they were a bit too similar).
mossylawn marks and normal text and bad brace example
- improved contrast and color diversity for python files, and __added styling for the `DECORATOR` style.__
mossylawn python example
- improved contrast and color diversity for JSON files. __Major improvement to contrast for `STRINGEOL` and `ESCAPE SEQUENCE` styles.__
    - also add `jsonl` (JSON Lines) and `ipynb` (Jupyter notebook) to list of extensions that use JSON lexer.

  mossylawn json example
- Improved contrast and color diversity for INI files. This is the smallest improvement IMO.
mossylawn ini example

And before you ask: I simulated these new colors with a colorblindness simulator (https://pilestone.com/pages/color-blindness-simulator-1) and verified that the new color schemes have good contrast for colorblind people.

### DIFF
--- a/PowerEditor/installer/themes/MossyLawn.xml
+++ b/PowerEditor/installer/themes/MossyLawn.xml
@@ -13,6 +13,12 @@ Last Modified:       1/14/2015
                      Added support for CoffeeScript.\
 					 Improved contrast for readbility.
 					 2023-09-30: update Perl support.
+                     2024-02-12:
+                     improved contrast and color diversity for Python, JSON and INI files.
+                     Increased diversity of coloring for markers.
+                     Improved bad brace visibility.
+                     Improved visibility of hovered URLS.
+                     Improved contrast of selection background with normal text color.
 Released:            4/17/2012
 License:             Feel free to modify this theme. 
                      This theme is available under the terms of the Creative Commons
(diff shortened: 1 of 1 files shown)


molsonkiko:
By the way, I've looked at how my new (light blue) selected text color looks with the lexers for RC, C++, C, C#, Perl, R, SQL, XML, HTML, YAML, and probably some others. __The selected text color has reasonably good contrast with all the styles for every lexer I've seen.__

molsonkiko:
One shortcoming of this PR as it currently stands - the Mark 2 background style (pretty much black) has almost no contrast with black text, which is pretty commonly used for comments in a lot of different lexers (like XML and C#). I'd be open to changing that if people want, but I don't think it's a huge issue because it's only a Mark style, and it has good contrast with almost everything else.

donho:
I prefer the current selection bg colour:
image

but I respect that you've got your personal preferences - that's why all the themes can be modified by the users.
You can keep your preferences in your local usage. But besides adding the missing entries, let's keep the themes as they are currently.

molsonkiko:
I understand and respect your decision. But I very much hope you will
consider changing the bad brace style to make it easier to see. Of all my
proposed changes, that feels like the most useful.

PR #14725, opened 2024-02-13
fix hanging Notepad++ when closed minimized

### LINKED ISSUE #14718: Freezes after attempting to close from task bar.
### Description of the Issue

Unable to close, save, or click anything in app, and must kill from Task Manager. If notepad++ is minimized, with 2 or more docs edited, and you try to close it using right click close on task bar icon.

1. Uncheck "Remember current session for next launch" in Backup
2. Have 2 or more unsaved docs.
3. Minimize app
4. Right click on task bar icon and click close.
5. Nothing shows and app is unusable if you maximize

### Expected Behavior

Show the "Save" dialog.

### Actual Behavior

The application acts like a dialog has been displayed and locks up everything. At this point you're unable to save or even just close it and must kill task.

### Debug Information

Windows 10 device:
Notepad++ v8.5.7   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Pro for Workstations (64-bit)
OS Version : 22H2
OS Build : 19045.3930
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    JSMinNPP (1.2205)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NPPJSONViewer (1.40)
    XMLTools (3.1.1.13)

Windows 11 device:
Notepad++ v8.5.4   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.3007
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    NPPJSONViewer (2.0.5)

xomx:
Fix #14718 .

STR & debugging in the issue.

### DIFF
--- a/PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
+++ b/PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
@@ -79,10 +79,15 @@ void StaticDialog::goToCenter(UINT swpFlags)
 {
 	RECT rc{};
 	::GetClientRect(_hParent, &rc);
+	if ((rc.left == rc.right) || (rc.top == rc.bottom))
+		swpFlags |= SWP_NOSIZE; // sizing has no sense here
+
 	POINT center{};
 	center.x = rc.left + (rc.right - rc.left)/2;
 	center.y = rc.top + (rc.bottom - rc.top)/2;
 	::ClientToScreen(_hParent, &center);
+	if ((center.x == -32000) && (center.y == -32000))
+		swpFlags |= SWP_NOMOVE; // moving has no sense here (owner wnd is minimized)
 
 	int x = center.x - (_rc.right - _rc.left)/2;
 	int y = center.y - (_rc.bottom - _rc.top)/2;
--- a/PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
+++ b/PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
@@ -86,7 +86,7 @@ void StaticDialog::goToCenter(UINT swpFlags)
 	center.x = rc.left + (rc.right - rc.left)/2;
 	center.y = rc.top + (rc.bottom - rc.top)/2;
 	::ClientToScreen(_hParent, &center);
-	if ((center.x == -32000) && (center.y == -32000))
+	if ((center.x == -32000) && (center.y == -32000)) // https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453
 		swpFlags |= SWP_NOMOVE; // moving has no sense here (owner wnd is minimized)
 
 	int x = center.x - (_rc.right - _rc.left)/2;
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -2450,13 +2450,20 @@ int Notepad_plus::doSaveOrNot(const TCHAR* fn, bool isMulti)
 	if ((NppParameters::getInstance()).isEndSessionCritical())
 		return IDCANCEL; // simulate Esc-key or Cancel-button as there should not be any big delay / code-flow block
 
-	// In case Notepad++ is iconized into notification zone
-	if (!::IsWindowVisible(_pPublicInterface->getHSelf()))
+	// In case Notepad++ is minimized into taskbar or iconized into notification zone
+	if (::IsIconic(_pPublicInterface->getHSelf()))
 	{
-		::ShowWindow(_pPublicInterface->getHSelf(), SW_SHOW);
+		::ShowWindow(_pPublicInterface->getHSelf(), SW_RESTORE);
+	}
+	else
+	{
+		if (!::IsWindowVisible(_pPublicInterface->getHSelf()))
+		{
+			::ShowWindow(_pPublicInterface->getHSelf(), SW_SHOW);
 
-		// Send sizing info to make window fit (specially to show tool bar.)
-		::SendMessage(_pPublicInterface->getHSelf(), WM_SIZE, 0, 0);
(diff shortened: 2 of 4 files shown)


xomx:
@donho 
I found some invalid logic in my PR, so please ignore this until I finish my more thorough debugging.
Fortunately, this bug does not appear in the default N++ config, so IMO it is not a blocker of a new version release.

xomx:
@donho 
Now I read your comment in the issue to be fixed here about splitting the fixing "one by one". So will be acceptable to:
- leave fixes (from the above last commit) for the Notepad_plus.cpp & StaticDialog.cpp in this PR
- and create a separate issue and PR for the possible tray-icon NppBigSwitch.cpp WM_CLOSE no-GUI problem ?

Fix in the StaticDialog.cpp generally fixes the root of the #14718 problem (invisible wnd positioning when owner-wnd is minimized), also for any future possible occurrence of this situation (a N++ dlg created & centered while the N++ main-wnd is minimized).

Fix for the Notepad_plus.cpp only ensures restoring of the minimized N++ main-wnd (to be coherent here with the behavior when the N++ is minimized to systray and not to the taskbar) when the doSaveOrNotBox dlg has been invoked.

donho:
@xomx 
I've said:
> It might be related, but let's fix the issue one by one, **if it's possible.**

So feel free to take action that you judge "reasonable" :)

xomx:
@donho 
Done.

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
```diff
@@ -79,10 +79,15 @@ void StaticDialog::goToCenter(UINT swpFlags)
...
 	::GetClientRect(_hParent, &rc);
+	if ((rc.left == rc.right) || (rc.top == rc.bottom))
+		swpFlags |= SWP_NOSIZE; // sizing has no sense here
+
 	POINT center{};
 	center.x = rc.left + (rc.right - rc.left)/2;
 	center.y = rc.top + (rc.bottom - rc.top)/2;
 	::ClientToScreen(_hParent, &center);
+	if ((center.x == -32000) && (center.y == -32000))
+		swpFlags |= SWP_NOMOVE; // moving has no sense here (owner wnd is minimized)
```
donho, line 90: I restored `void StaticDialog::goToCenter(UINT swpFlags)` as it is in master and keep your modification in `int Notepad_plus::doSaveOrNot(const TCHAR* fn, bool isMulti)` - it fixes issue #14718.

So my questions here are: 
1. if the modification of `void StaticDialog::goToCenter(UINT swpFlags)` is necessary, what's the reason? 
2. Why -32000?

donho (changes requested):
@xomx 
I do believe you have the right reason. Please provide the reason.

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
xomx: > 1\. if the modification of void StaticDialog::goToCenter(UINT swpFlags) is necessary, what's the reason?

The `StaticDialog::goToCenter` fix is necessary to correct the very essence of the problem and not just one specific occurrence of it. You can try the opposite (restore the `Notepad_plus::doSaveOrNot` code and keep the fix for the `StaticDialog::goToCenter`) and the current problem will be solved too, but in this case also for any future possible occurrence of this situation (a N++ modal child/pop-up wnd centering while its owner-wnd is minimized).

The additional `Notepad_plus::doSaveOrNot` fix is only there to mimic the behavior when the N++ is minimized to systray instead to taskbar and the `doSaveOrNotBox` dlg is being invoked.

> 2\. Why -32000?

Windows NT family just stuck with that `-32000,-32000` coordinates for the minimized wnds, probably for some prehistoric compatibility reasons. I am copying here my explanation from the issue:
"it is also centered relative to the owner (main N++ wnd), but when that owner is minimized (systray seems to be ok), the `ClientToScreen` WINAPI returns [-32000, -32000](https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453) as the `center` `x,y`"

donho: > The StaticDialog::goToCenter fix is necessary to correct the very essence of the problem and not just one specific occurrence of it. You can try the opposite (restore the Notepad_plus::doSaveOrNot code and keep the fix for the StaticDialog::goToCenter) and the current problem will be solved too, but in this case also for any future possible occurrence of this situation (a N++ modal child/pop-up wnd centering while its owner-wnd is minimized).

***"A good programmer is someone who always looks both ways before crossing a one-way street."***
:)

> "it is also centered relative to the owner (main N++ wnd), but when that owner is minimized (systray seems to be ok), the ClientToScreen WINAPI returns [-32000, -32000](https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453) as the center x,y"

Then it would be great to add https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453 in a comment to justify the magic number, wouldn't be it?

donho (changes requested):
Please add the comment with url https://devblogs.microsoft.com/oldnewthing/20041028-00/?p=37453 to justify the value "-32000".

PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
xomx: @donho 
Done.

donho (approved):

PR #14729, opened 2024-02-14
Update css.xml

Enjoyop:
Latest update on css keywords.


donho:
@Enjoyop 
You have added already 2 times the css keywords in the same files.
I wonder why don't you commit in one shot?

Enjoyop:
Since I'm not doing it with full focus, is there any priority or disadvantage?

I won't add anything else to the CSS section, I guess I've already added what I can.

PR #14730, opened 2024-02-14
Context menu popup location at current text position when invoked via keyboard

### LINKED ISSUE #14727: Pressing the context menu key shows the context menu on the mouse position.
1.  Open a document in NotePad++
2.  Move the caret into the text
3.  Move the mouse cursor somewhere outside the text editor
4. Presse the context menu key (Windows Key, or Shift+F10)
5. The context menu pops up at the mouse position not near the caret position

I would expect that the context menu shows up at the caret position´(input position in the text)
Using x64 Notepad++ v8.6.2

Hint for the programmer:
The WM_CONTEXTMENU message is sent, with a position. If the context menu key is pressed than the x/y position is (-1,-1).

If the keyboard is used the caret position in the text should be used...

alankilborn:
Fix #14727 

Notes:

* created new HWND variable to avoid casting `wParam` over and over
* moved getting/calculating of `p` closer to where something is done with it

If it is "too much" change, the above noted items are not strictly necessary and can be backed-out if desired.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -1959,20 +1959,37 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			}
 			else
 			{
-				if ((HWND(wParam) == _mainEditView.getHSelf()) || (HWND(wParam) == _subEditView.getHSelf()))
+				HWND activeViewHwnd = reinterpret_cast<HWND>(wParam);
+
+				if ((activeViewHwnd == _mainEditView.getHSelf()) || (activeViewHwnd == _subEditView.getHSelf()))
 				{
-					if ((HWND(wParam) == _mainEditView.getHSelf()))
+					if (activeViewHwnd == _mainEditView.getHSelf())
 						switchEditViewTo(MAIN_VIEW);
 					else
 						switchEditViewTo(SUB_VIEW);
 
-					POINT p;
-					::GetCursorPos(&p);
 					ContextMenu scintillaContextmenu;
+					
 					std::vector<MenuItemUnit>& tmp = nppParam.getContextMenuItems();
 					bool copyLink = (_pEditView->getSelectedTextCount() == 0) && _pEditView->getIndicatorRange(URL_INDIC);
 					scintillaContextmenu.create(hwnd, tmp, _mainMenuHandle, copyLink);
+					
+					POINT p;
+					if (lParam == -1)
+					{
+						// context menu activated via keyboard; pop up at text caret position
+						auto caretPos = _pEditView->execute(SCI_GETCURRENTPOS);
+						p.x = static_cast<LONG>(_pEditView->execute(SCI_POINTXFROMPOSITION, 0, caretPos));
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1959,20 +1959,37 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
-					POINT p;
-					::GetCursorPos(&p);
 					ContextMenu scintillaContextmenu;
+					
 					std::vector<MenuItemUnit>& tmp = nppParam.getContextMenuItems();
 					bool copyLink = (_pEditView->getSelectedTextCount() == 0) && _pEditView->getIndicatorRange(URL_INDIC);
 					scintillaContextmenu.create(hwnd, tmp, _mainMenuHandle, copyLink);
+					
+					POINT p;
+					if (lParam == -1)
```
donho, line 1978: I cannot find the Microsoft's document for indicating the meaning of lParam value "-1" for  WM_CONTEXTMENU.
OTOH, ***If the context menu is generated from the keyboard for example, if the user types SHIFT+F10 then the x- and y-coordinates are -1 and the application should display the context menu at the location of the current selection rather than at (xPos, yPos).***

https://learn.microsoft.com/en-us/windows/win32/menurc/wm-contextmenu

donho (changes requested):
Please provide the online MS doc for `(lParam == -1)`.

PowerEditor/src/NppBigSwitch.cpp
alankilborn: It is in the _Remarks_ section:  https://learn.microsoft.com/en-us/windows/win32/menurc/wm-contextmenu#:~:text=the%20VK_APPS%20key.-,If%20the%20context%20menu,-is%20generated%20from

alankilborn: Ready for review again.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -1975,7 +1976,9 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 					scintillaContextmenu.create(hwnd, tmp, _mainMenuHandle, copyLink);
 					
 					POINT p;
-					if (lParam == -1)
+					int xPos = GET_X_LPARAM(lParam);
+					int yPos = GET_Y_LPARAM(lParam);
+					if ((xPos == -1) && (yPos == -1))
```
donho, line 1981: I would do the following:
```cpp
					p.x = GET_X_LPARAM(lParam);
					p.y = GET_Y_LPARAM(lParam);
					if ((p.x== -1) && (p.y == -1))
					{
						// context menu activated via keyboard; pop up at text caret position
						...
					}
```
And remove
```cpp
					else
					{
						// context menu activated via mouse right-click; pop up at click location
						::GetCursorPos(&p);
					}
```

Does it make more sense?

donho (changes requested):
Please modify as suggested.

PowerEditor/src/NppBigSwitch.cpp
alankilborn: Sure, I wasn't sure how much change would be tolerated here; I had thought you liked `GetCursorPos` for some reason...

PR #14735, opened 2024-02-15
Create smali.xml

Enjoyop:
smali code completion.


donho:
https://github.com/notepad-plus-plus/userDefinedLanguages

PR #14742, opened 2024-02-16
Implement OpenTheCopy checkbox on SaveACopyAs dialog

### LINKED ISSUE #11860: "Save a Copy As..." window title should be "Save a Copy As"
`File` -> `Save As...` -> window title is "Save As".
`File` -> `Save a Copy As...` -> window title is the same "Save As", which is misleading - what did I _actualy_ select from `File`...?

### LINKED ISSUE #11861: [Feature Request] Save a Copy As and then OPEN THE COPY

### Debug Information

### Workarounds

### LINKED ISSUE #14737: "Save As" dialog title bar leaves user unsure which command got them there

1. Execute _File_ > _Save As_ and a standard file-saving dialog window opens
2. Note the title bar is "Save As" (English localization) and also note the appearance of the rest of the dialog
3. Cancel the dialog
4. Execute _File_ > _Save a Copy As_ and a standard file-saving dialog window opens
5. Note the title bar is "Save As" (English localization) and also note the appearance of the rest of the dialog
6. Cancel the dialog

### Actual Behavior

After step 2 or step 5 I am unsure which command brought me there -- the resulting dialogs are exactly the same in each case, so I get no visual confirmation that I'm in the "right place".

Of course I know for the purposes of the above STR, but in general, since _Save As_ and _Save a Copy As_ are adjacent items in the _File_ menu, a "mouse stutter" may not select my true intention, and, once the saving dialog opens, I have no way of knowing which command brought me to that point.

### Expected Behavior
* I expect the title bar of the dialog to be "Save As" when I've selected _File_ > _Save As_ (it is so this is not a problem)
* I expect the title bar of the dialog to be "Save Copy As" when I've selected _File_ > _Save a Copy As_ (this is the PROBLEM area)

### Debug Information
Notepad++ v8.6.2   (64-bit)
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.2\npp.8.6.2.portable.x64\notepad++.exe
Command Line : -titleAdd="8.6.2 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.3155
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3)
    NppConverter (4.5)
    NppExport (0.4)
    PythonScript (2)

alankilborn:
Fix #11861 

Fix #14737 
Fix #11860

---

UI details of change:

image

---

Notes:

Fixes for 14737 / 11860 included adding localization for the `Save a copy as` window title bar text.  This joins the existing localization for the `Rename` window title bar text.

For completeness, the `Open` window title bar text and the `Save As` title bar text were also made localization-capable with this change (instead of relying on the OS to choose the title bar text).  If that isn't desired, I can certainly back that out.

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1636,6 +1636,9 @@ NOTE: Choosing not to create the placeholders or closing them later, your sessio
 			<shift-change-direction-tip value="Use Shift+Enter to search in the opposite direction"/>
 			<two-find-buttons-tip value="2 find buttons mode"/>
 			<file-rename-title value="Rename"/>
+			<file-open-title value="Open"/>
+			<file-save-as-title value="Save As"/>
+			<file-save-as-copy-title value="Save Copy As"/>
 			<find-in-files-filter-tip value="Find in cpp, cxx, h, hxx &amp;&amp; hpp:
 *.cpp *.cxx *.h *.hxx *.hpp
 
@@ -1737,6 +1740,7 @@ Find in all files but exclude all folders log or logs recursively:
 			<session-save-folder-as-workspace value="Save Folder as Workspace"/>
 			<tab-untitled-string value="new "/>
 			<file-save-assign-type value="&amp;Append extension"/>
+			<file-save-open-copy value="&amp;Open copy after saving"/>
 			<close-panel-tip value="Close"/>
 			<IncrementalFind-FSFound value="$INT_REPLACE$ matches"/>
 			<IncrementalFind-FSNotFound value="Phrase not found"/>
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1785,29 +1785,55 @@ bool Notepad_plus::fileSaveAs(BufferID id, bool isSaveCopy)
 
 	LangType langType = buf->getLangType();
 
-	const bool defaultAllTypes = NppParameters::getInstance().getNppGUI()._setSaveDlgExtFiltToAllTypes;
+	NppParameters& nppParam = NppParameters::getInstance();
+	NppGUI& nppGUI = nppParam.getNppGUI();
+
+	const bool defaultAllTypes = nppGUI._setSaveDlgExtFiltToAllTypes;
 	const int langTypeIndex = setFileOpenSaveDlgFilters(fDlg, false, langType);
 	
 	fDlg.setDefFileName(buf->getFileName());
 
 	fDlg.setExtIndex(langTypeIndex + 1); // +1 for "All types"
 
-	const generic_string checkboxLabel = _nativeLangSpeaker.getLocalizedStrFromID("file-save-assign-type",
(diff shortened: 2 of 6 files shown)


donho:
I strongly dislike the idea. But I will check if the feature is needed.

alankilborn:
>I strongly dislike the idea. But I will check if the feature is needed.

The only reason I played around with an implementation for this is because two other users expressed interest in it (in the issue).

One of those users actually said: _I also struggle with this_ (see https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11861#issuecomment-1937846641).  Users shouldn't struggle.  :-)

Perhaps the idea of the command the way it exists is purely for archival purposes, i.e., take a snapshot of the file as it exists right now and save it, out of sight and out of mind.  That's not how I and presumably other users use it:  We use it because we want to have a derivative file, but continue editing both the original file and the copy.

Is the complaint about the "idea" or about the proposed implementation?

donho:
Not only I don't like "Save a Copy As and then OPEN THE COPY" feature, but also I don't think "open copy after saving" checkbox should be beside of "Append extension" checkbox.
From user's aspect "Append extension" will apply only in the dialog, whereas "open copy after saving" applies AFTER the closing of dialog. Both options sit together make users confused.

Opening a saved file is not a big deal, if the option "Remember last directory" is ON. 
So this feature won't be merged in master.

donho:
@alankilborn 
After reconsidering the feature, I'm wondering if it's possible to add a button instead of a checkbox as following:

image

?

If it's possible and there's no a lot of coding, it's worth IMO.

alankilborn:
>I'm wondering if it's possible to add a button instead of a checkbox in the Save As a Copy dialog

>If it's possible ... it's worth IMO.

I'll look into doing a button.
Thank you for considering this to be valuable enough to be added in some manner.

alankilborn:
@donho 

I've looked into this somewhat, and a new custom "Save & Open" button may have some problems, among them:

* The necessary size of the button may be an issue -- sure, for English it is rather short text keeping the button small, but is this going to be the case for all translations?
* I don't see in the code where the XY location of the _Append extension_ checkbox is set; could this be something that the OS decides?  If so, then a "Save & Open" button may not appear in as good of a place as the screenshot above shows
* The existing `CustomFileDialog.cpp` code makes it fairly clear that "finding" a button in the dialog is not a trivial task; adding another probably makes this even more complicated
* The code required to implement may not be short, violating "...and there's not a lot of coding"

Alternative idea:  What if the save-and-open-copy functionality is run if the Shift key is held down during the _Save_ button press?  Perhaps the _Save_ button could be given a tooltip so the user could discover this functionality?

donho:
@alankilborn 

> Alternative idea: What if the save-and-open-copy functionality is run if the Shift key is held down during the Save button press? Perhaps the Save button could be given a tooltip so the user could discover this functionality?

Shift-Click on Save button to save and open is a good way to me. Of course, the tooltip on the button can only show on SaveACopyAs dialog.

donho:
@alankilborn 
There are some conflicts. This PR will be closed. Please do a new PR based on current master.

PR #14766, opened 2024-02-20
Allow prevention of text entry of typed control characters

### LINKED ISSUE #13279: Notepad++ prints ENQ symbol when i press ctrl+e
### Description of the Issue

Notepad++ prints ENQ symbol when i press ctrl+e

1. Open empty file
2. press ctrl+e
3. 

### Expected Behavior

Nothing
### Actual Behavior

Notepad++ inputs 
[новый 1.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/10882197/1.txt)
hex 05  
### Debug Information

Notepad++ v8.4.9   (64-bit)
Command Line : "F:\cygwin64\home\frolo\git\cygcheck.out"
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 22H2
OS Build : 22621.1105
Current ANSI codepage : 1251
Plugins : 
    ComparePlugin (2.0.2)
    HexEditor (0.9.12)
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

Снимок экрана (29)

alankilborn:
Fix #13279 

Floating this as a "first step" to a solution as a RFC (request for comments).
Still has a TODO if it is liked:  a preference setting to enable/disable ?

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -501,6 +501,17 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
 			break;
 		}
 
+		case WM_CHAR:
+		{
+			// prevent "control characters" from being entered in text
+			// (don't need to be concerned about Tab or CR or LF etc here)
+			if ((wParam >= 0 && wParam <= 31) || wParam == 127)
+			{
+				return FALSE;
+			}
+			break;
+		}
+
 		case WM_KEYUP:
 		{
 			if (wParam == VK_PRIOR || wParam == VK_NEXT)


alankilborn:
> But there may be times when I want to temporarily switch it to allow me to enter the control characters (without having to restart the application).

Double-clicking here will still insert the control character:

image

(But, I agree with a new preference setting for the feature of this PR)

alankilborn:
@donho 

Looking for a yes/no on the basic idea here.
If a "yes" then I will add a preference setting or zero-length xml file or whatever is wanted to control the behavior.

donho:
@alankilborn 
> Looking for a yes/no on the basic idea here.

Maybe
:D 
(BTW, I always want to build a dialog with 3 buttons: "Yes", "No" & "Maybe")

> If a "yes" then I will add a preference setting or zero-length xml file or whatever is wanted to control the behavior.

The feature deserves a GUI option - where you think to put this option more suitable?

alankilborn:
> The feature deserves a GUI option - where you think to put this option more suitable?

As Peter suggested earlier, it could go here where the yellow is:

image

---

It could also simply be put on the _Editing 2_ page.

---

Either way, I envision it as a checkbox:  "Prevent control character typing into document"

---

Still Yes / No / Maybe ?
:-)

donho:
> As Peter suggested earlier, it could go here where the yellow is:

Yes, I think it'll be the better place.

> Either way, I envision it as a checkbox: "Prevent control character typing into document"

If it'll be in "Non-Printing Characters" section, is it rather "Prevent non-printable character typing into document" ?

> Still Yes / No / Maybe ?

If the answer is no, there won't be `The feature deserves a GUI option...`
However, there'll be no enough place to add the option. 
I'll move some options from "Editing 1" to "Editing 2". 
Once it's done (I'm doing it), you can continue this PR (rather close this one and do a new PR by being based on the master).

What do you think?

donho:
@alankilborn 
> I'll move some options from "Editing 1" to "Editing 2".
Once it's done (I'm doing it), you can continue this PR (rather close this one and do a new PR by being based on the master).

It's done in [12548b6](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/12548b6abb607de94163e9171908de4a1b7a718c).
Just uncomment this line:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/12548b6abb607de94163e9171908de4a1b7a718c#diff-e5c81a5d2fb35d00259941d97465c669073c5f49060dee607b1a1c670736a385R130
then you can continue this the coding of this PR (Please create a new PR as I suggested).

donho:
@alankilborn 
> Either way, I envision it as a checkbox: "Prevent control character typing into document"

Will you still want to link the GUI option to the feature you have done in this PR?
Or I merge this PR as it is, and I can continue the implementation.

alankilborn:
> Or I merge this PR as it is, and I can continue the implementation.

YOU can go ahead with it -- thanks

Coises:
Maybe I’m too late with this...

I have to think anyone who is randomly typing Ctrl+ combinations with any frequency could hardly be able to use a computer keyboard at all. I can see, though, that if you type one of these combinations by accident (and we all make occasional mistakes) and **View** | **Show Symbols** | **Show control characters...** is off, you can’t see that you typed it.

Just floating an idea... perhaps a setting could be avoided by making this contingent on the **View** | **Show Symbols** setting. If you can’t see the character, you can’t type it. If you can see it, it will be easy enough to recognize it and delete it, so the suppression won’t be needed.

donho:
@Coises 
> Maybe I’m too late with this...

Not at all. I just began for the implementation.

> Just floating an idea... perhaps a setting could be avoided by making this contingent on the View | Show Symbols setting. If you can’t see the character, you can’t type it. If you can see it, it will be easy enough to recognize it and delete it, so the suppression won’t be needed.

If I understand you, what you have suggested is checking the visibility of control characters to determinate the ON or OFF of this feature, instead of GUI option.

It could confuse people IMO, I can imagine the following scenario:
The visibility of C0 is OFF, user type Ctrl-E on purpose, nothing happen, but user think `ENO` is keyed in.

With the GUI option, user turn it ON or OFF. No ambiguity. My question is, what's the value by default? ON or OFF?

pryrt:
@Coises ,
> randomly typing Ctrl+ combinations with any frequency could hardly be able to use a computer keyboard at all

It annoyed me when you said this before, but I couldn't articulate why at that point.  But thinking about it again, I believe I've figured out something that might make sense to you:

If we are trying to encourage someone who is coming from VSCode (or any other editor) to use Notepad++ instead, they may have a <kbd>Ctrl+___</kbd> keyboard shortcut that they are used to, which isn't bound to any action by default in Notepad++.  In this case, it's not "random typing"; it might be their muscle-memory that needs training for Notepad++.  That's nothing to insult them over, but rather applaud their new choice in text editor.

And one of the benefits of this feature is to give those people the choice of whether it's better for training their muscle-memory to insert weird characters, or better to train their muscle-memory to not insert any characters with those keystrokes -- depending on the person, they might prefer to be trained by it doing nothing, or prefer to be trained by seeing a weird character; either way, who are we to judge (or to accuse them of "random typing" or being incapable of using a computer keyboard).

@donho,
> My question is, what's the value by default? ON or OFF?

If the default is OFF (so it doesn't prevent the characters), there is no change in default behavior -- which means fewer complaints as issues here or posts in the forum.  But as Alan is fond of pointing out, for good features, keeping them off by default means that more people don't find out about the feature.  

Despite my mention of complaints, I would think this feature wouldn't get a lot of complaints even if it were defaulted ON.  Since I'm willing to say that, I guess that means I'm willing to deal with the complaints in the Forum as they arise. ;-)

So my vote is ON.

alankilborn:
>someone who is coming from VSCode

Probably no one migrates from VSCode to Notepad++.  :-)

alankilborn:
>If the default is OFF (so it doesn't prevent the characters)...

Default should be ON.  Prevent the characters.
In the unlikely event that there's someone out there that inserts some control characters using the keyboard, that user can post a question to the forum and I'll gladly tell them how to restore their favorite behavior.

alankilborn:
No offense, but I think @Coises is misguided.

Scenario:
Novice user fat-fingers Ctrl+e and gets an `ENQ` inserted into his document.  User has absolutely no idea what this crazy white on black `ENQ` is, and how to get rid of it.  User figures out they can backspace delete it, but is left wondering how they got that in there anyway.
Better to just prevent the fat-fingering incident in its entirety (user never sees anything).

Coises:
I apologize for what, in retrospect, was a flip-cum-insulting tone in my remarks.

For what it’s worth:

Since pretty much everyone else’s sense of this is opposite of mine, I can accept that my intuition in this case is misguided.

I agree that the default should be to suppress the control-character generation. The only combination which leads to silently broken files is the one in which the user did not intend to enter a control character, viewing control characters is off, and a control character is entered. We can’t change the first and wouldn’t want to mess with the second, so making the third default to the “safe” value is the sensible choice.

PR #14789, opened 2024-02-26
Reorganize "Editing" sections in Preferences dialog

donho:


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/12548b6abb607de94163e9171908de4a1b7a718c

PR #14744, opened 2024-02-17
Shortcut mapper changes: split filter box into words and require each to match

molsonkiko:
~~Fix #12557, fix #10486~~ (already fixed by PR #14732), fix #14743

### What this changes
~~1. Fix: The shortcut mapper will filter Scintilla Commands by keycombo(s). Previously it only filtered Scintilla Commands by name, even though it filtered every other type of command by keycombo as well.~~ (fixed by #14732)
2. Rather than requiring that at least one field for a command match the entire contents of the Filter text box, the Filter text box will be split into a list of words, each of which must be matched by at least one field of the command.

### To test this PR

1. Open the shortcut mapper.
2. If you don't have any plugins installed, install the `mimeTools` plugin.
3. In the `Macros` tab, verify that the macro `Trim Trailing Space and Save` (and no others) is matched by the filter "trim +S" (note that this assumes that the default keycombo is bound here)
4. In the `Main menu` tab, verify that the menu command `New` (and no others) is matched by the filter `w +n`
5. In the `Main menu` tab, verify that the menu command `Reload from Disk` (and no others) is matched by the filter `load disk ctrl`
6. In the `Run commands` tab, verify that the `Get PHP help` command (and no others) is matched by `f1 php help`
7. Verify that filters match all the Plugin commands fields (name, plugin, shortcut)
8. Verify that filters match all the Scintilla commands fields (name, shortcuts)

### Examples of how the new behavior should work
__1. Consider a plugin command named "do thing" with plugin "BlahLint" and keycombo "Alt+G"__
    Filters that match:
    - "lint do +" (the "+" is matched by "alt+g")
    - "+G thing blah"
    Filters that *do not* match:
    - "lint foo +g" ("foo" is not in any field)
    - "thing alt quz" ("quz" is not in any field)
__2. Consider a Scintilla command named "foo baz" with keycombos "Alt+H or Ctrl+Shift+Q"__
    Filters that match:
    - "baz or" (the "or" is matched by "Alt+H or Ctrl+Shift+Q")
    - "+q foo"
    Filters that *do not* match:
    - "+q +z" ("+z" is not in any field)
    - "quz"
    - "baz Tab" ("tab" is not in any field)
__3. Consider any other type of command named "cute kitty" with keycombo "Alt+K"__
    Filters that match:
    - "alt k" matches ("k" is matched by "kitty" *and* "alt+k")
    - "kit alt"
    Filters that *do not* match:
    - "cute q" ("q" is not in any field)
    - "z kit"

PR #14816, opened 2024-03-01
filter shortcut mapper by words (split filter by whitespace)

### LINKED ISSUE #14743: Feature proposal: filter shortcut mapper by each word in the filter text box separately
Before I begin, I should note that this is not a request for someone else to implement a feature. I have already implemented this feature in my own fork, and can submit a PR if core devs are interested.

### Description of the Issue

Currently the shortcut mapper filters using the entire contents of the `Filter` search box (I will hereafter call this `filterStr`). If any field for a shortcut contains `filterStr` in its entirety, the shortcut is displayed.

This behavior is fine, but there are many times when a more flexible system would be useful. A simple, user-friendly way to do this is to split `filterStr` into a list of whitespace-separated words (I will hereafter call this `filterWords`), and then require *every word in `filterWords` to be matched by at least one field.*

For example, let's look at a screenshot from 8.5.8.

image

I have used the `jsont` search term to find all the shortcuts associated with the `JsonTools` plugin. But JsonTools has a lot of shortcuts, and maybe I want to see all the JsonTools shortcuts that relate to `regex`. With the current implementation, this can only be done manually.

Here's a screenshot showing how to do exactly that with my proposed implementation.

image

Hopefully the advantages of the new behavior are obvious.

### Disadvantages of the proposed behavior, and my counterarguments

1. __It's different from the existing behavior, and would cause initial confusion.__
    - __Anything matched by the existing behavior will still be matched by the new behavior.__ For example, suppose the user wants to find a shortcut called `csv to json`, so they input `csv to json`. The new behavior will match `csv to json`, but it will also match any shortcut containing the word `csv` associated with the `JsonTools` plugin.
2. __Literal space characters in the search term are no longer matched.__
    - I think this is a pretty niche case.
3. __It might make performance worse.__
    - I tested it. There's no obvious difference.
4. __It makes the code more complex.__
    - my current implementation causes a roughly 20-line diff, including about 20 changed lines in `ShortcutMapper.cs` and 2 changed lines in `ShortcutMapper.h`. About half of the new lines are explanatory comments.
    - in addition, the changed functions are only called in `ShortcutMapper.cs`, __which is a modal dialog__, meaning that ripple effects to the broader codebase are essentially impossible.
    - __However, the code footprint gets a lot larger if a new setting is desired to allow toggling between the old behavior and the new behavior__ (say, maybe a checkbox next to the `Filter` textbox). For this reason, I would prefer to avoid adding a new configurable setting, especially one that is saved to a config file.

molsonkiko:
Fix #14743 
Close #14744 (this is a re-submission with no changes)

@alankilborn you said you were interested in this feature. Want to test it before Don Ho merges it?

### What this changes
Rather than requiring that at least one field for a command match the entire contents of the Filter text box, the Filter text box will be split into a list of words, each of which must be matched by at least one field of the command.

### To test this PR

1. Open the shortcut mapper.
2. If you don't have any plugins installed, install the `mimeTools` plugin.
3. In the `Macros` tab, verify that the macro `Trim Trailing Space and Save` (and no others) is matched by the filter "trim +S" (note that this assumes that the default keycombo is bound here)
4. In the `Main menu` tab, verify that the menu command `New` (and no others) is matched by the filter `w +n`
5. In the `Main menu` tab, verify that the menu command `Reload from Disk` (and no others) is matched by the filter `load disk ctrl`
6. In the `Run commands` tab, verify that the `Get PHP help` command (and no others) is matched by `f1 php help`
7. Verify that filters match all the Plugin commands fields (name, plugin, shortcut)
8. Verify that filters match all the Scintilla commands fields (name, shortcuts)

### Examples of how the new behavior should work
__1. Consider a plugin command named "do thing" with plugin "BlahLint" and keycombo "Alt+G"__
    Filters that match:
    - "lint do +" (the "+" is matched by "alt+g")
    - "+G thing blah"
    Filters that *do not* match:
    - "lint foo +g" ("foo" is not in any field)
    - "thing alt quz" ("quz" is not in any field)
__2. Consider a Scintilla command named "foo baz" with keycombos "Alt+H or Ctrl+Shift+Q"__
    Filters that match:
    - "baz or" (the "or" is matched by "Alt+H or Ctrl+Shift+Q")
    - "+q foo"
    Filters that *do not* match:
    - "+q +z" ("+z" is not in any field)
    - "quz"
    - "baz Tab" ("tab" is not in any field)
__3. Consider any other type of command named "cute kitty" with keycombo "Alt+K"__
    Filters that match:
    - "alt k" matches ("k" is matched by "kitty" *and* "alt+k")
    - "kit alt"
    Filters that *do not* match:
    - "cute q" ("q" is not in any field)
    - "z kit"

### ideas for future changes

- maybe add a checkbox next to the filter text box, to change the way text is filtered?
- maybe create a method with a variable number of arguments that tests each word of the filter on each arg, to reduce repetitiveness of code in implementation

### DIFF
--- a/PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
+++ b/PowerEditor/src/WinControls/Grid/ShortcutMapper.cpp
@@ -194,32 +194,76 @@ generic_string ShortcutMapper::getTextFromCombo(HWND hCombo)
 
 bool ShortcutMapper::isFilterValid(Shortcut sc)
 {
-	if (_shortcutFilter.empty())
+	// all words in _shortcutFilter must be in the name or keycombo
+	// For example, the shortcut with name "foo bar baz" and keycombo "Ctrl+A"
+	// would be matched by the filter "foo ctrl" and the filter "bar +a"
+	// but *not* by the filter "foo shift" or the filter "quz +a"
+	size_t filterSize = _shortcutFilter.size();
+	if (filterSize == 0)
 		return true;
 
 	wstring shortcut_name = stringToLower(string2wstring(sc.getName(), CP_UTF8));
 	wstring shortcut_value = stringToLower(string2wstring(sc.toString(), CP_UTF8));
 
-	// test the filter on the shortcut name and value
-	return (shortcut_name.find(_shortcutFilter) != std::string::npos) || 
-		(shortcut_value.find(_shortcutFilter) != std::string::npos);
+	for (size_t i = 0; i < filterSize; ++i)
+	{
+		generic_string filterWord = _shortcutFilter.at(i);
+		// every word must be matched by keycombo or name
+		if (shortcut_name.find(filterWord) == std::string::npos &&
+			shortcut_value.find(filterWord) == std::string::npos)
+			return false;
+	}
+	return true;
 }
 
 bool ShortcutMapper::isFilterValid(PluginCmdShortcut sc)
 {
-	// Do like a classic search on shortcut name, then search on the plugin name.
-	Shortcut shortcut = sc;
-	bool match = false;
-	wstring module_name = stringToLower(string2wstring(sc.getModuleName(), CP_UTF8));
-	if (isFilterValid(shortcut)){
+	// all words in _shortcutFilter must be in the name or the keycombo or the plugin name
+	// For example, the shortcut with name "foo bar baz" and keycombo "Ctrl+A" and plugin "BlahLint"
+	// would be matched by the filter "foo ctrl" and the filter "bar +a blah"
+	// but *not* by the filter "baz shift" or the filter "lint quz" 
+	size_t filterSize = _shortcutFilter.size();
+	if (filterSize == 0)
 		return true;
-	}
-	size_t match_pos = module_name.find(_shortcutFilter);
-	if (match_pos != std::string::npos){
-		match = true;
-	}
 
-	return match;
+	wstring shortcut_name = stringToLower(string2wstring(sc.getName(), CP_UTF8));
+	wstring shortcut_value = stringToLower(string2wstring(sc.toString(), CP_UTF8));
+	wstring module_name = stringToLower(string2wstring(sc.getModuleName(), CP_UTF8));
+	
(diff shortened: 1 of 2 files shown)


donho:
@molsonkiko 
There is error on CI
You should build and test it  locally before committing:
> 12. Finally, please test your pull requests, at least once.

molsonkiko:
> You should build and test it locally before committing:

I actually did test it locally before submitting my PR, on `Debug|x64` and `Debug|Win32`, and it worked fine. I saw no compiler errors or warnings that indicated I should make further changes.

In any case, I've been staring at the errors from the CI build, and it looks like all the errors are saying something along the lines that `bool ShortcutMapper::isFilterValid(ScintillaKeyMap sc)` has been implemented multiple times in `ShortcutMapper.cpp`. Does anyone have any idea why that would be?

molsonkiko:
In any case, I figured out what was wrong, and it should be fixed now.

I can resubmit if you're still interested.

molsonkiko:
[CI succeeded in this branch](https://github.com/molsonkiko/notepad-plus-plus/tree/filter_shortcut_mapper_by_words_revision1).

Sorry again for the confusion.

donho:
Did you synchronize your master with the Notepad++ one, create a branch and import your modification of this PR then do a new PR? If you did, it should not happen.

molsonkiko:
When I submitted this PR, I thought I had synchronized my master with the Notepad++ master, but I hadn't. The branch I just linked to is synchronized with the current NPP master - I just checked.

In the future, I intend to double-check (and probably triple-check) the commit history of my PRs relative to the current master before I submit a PR.

alankilborn:
> you said you were interested in this feature. Want to test it

Tested in your fork/branch as you shared https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14816#issuecomment-1974189089

It seems to work as advertised.  :-)

molsonkiko:
re-submitted the previously linked branch as #14824. CI should work this time around.

Issue #14539, opened 2024-01-01
Incorrect syntax highlighting in JavaScript with backtrick (`) in code


starsolaris:
### Description of the Issue
Incorrect syntax highlighting in JavaScript with backtrick (`) in code

### Steps to Reproduce the Issue
```javascript
const a = () => /`/;
```

### Expected Behavior
Text should not be with grey background after backtrick (`) symbol.

### Actual Behavior
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/715636/ba6c8b17-695f-4b38-a317-1761a8363dfe)

### Debug Information
Notepad++ v8.6   (64-bit)
Build time : Nov 23 2023 - 16:58:44
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 22H2
OS Build : 19045.3803
Current ANSI codepage : 1251
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

rdipardo:
Lexilla bug: ScintillaOrg/lexilla#94

Duplicate of, at least:
- #3252
- #3331
- #4007
- #4106
- #6193
- #7879
- #8421
- #9812
- #11056
- #11412
- #11562
- #12287

donho:
Duplicate also of:  #3822

donho:
![image](https://github.com/user-attachments/assets/af20c77c-e80a-441a-b766-2e94cef5e303)

PR #2432, opened 2016-10-16
Add possibility to customize ColourPopup's color palette

A-R-C-A:
This adds the possibility to customize the color palette used by the ColourPopup dialog. Customization can be done by right clicking in the list box. The color palette will be saved and loaded per theme, so every theme can have its own palette.

Issue #2420, opened 2016-10-13
Identifying Win XP and and Win2K ????


SinghRajenM:
Below piece of code does not provide guarantee about detecting WinXp and Win2K

From "ScintillaEditView.h"

``` C++
void convertSelectedTextToLowerCase() {
         // if system is w2k or xp
         if ((NppParameters::getInstance())->isTransparentAvailable())
                  convertSelectedTextTo(LOWERCASE);
         else
                  execute(SCI_LOWERCASE);
    };
```

 whereas transparentFuncAddr points to "SetLayeredWindowAttributes".

``` C++//-------------------------------------//
    // Transparent function for w2k and xp //
    //-------------------------------------//
    HMODULE hUser32 = ::GetModuleHandle(TEXT("User32"));
    if (hUser32)
        _transparentFuncAddr = (WNDPROC)::GetProcAddress(hUser32, "SetLayeredWindowAttributes");
```

I was debugging issue #2417 on Win10. As code is running on win10, I expected to statement execute(SCI_LOWERCASE); (from code snippet 1) to be executed, but it goes to convertSelectedTextTo(LOWERCASE);

In convertSelectedTextTo(LOWERCASE), API CharLowerW does not convert uppercase dotted (İ) to lowercase.
As per MSDN https://msdn.microsoft.com/en-us/library/windows/desktop/ms647467(v=vs.85).aspx unfortunately, there is no indication of success or failure. However, failure is rare. :(

So concerns:
1: Why npp does not hand over case conversion task Scintilla?
2: If Scintilla conversion does not work properly on winXp and Win2K, condition to check winXp and Win2K should be modified, So issue #2417 (or similar) will not occur atleast on Win10 as execute(SCI_LOWERCASE); working perfectly.

Notepad++ v7   (32-bit)
Build time : Oct 13 2016 - 23:29:49
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10
Plugins : DSpellCheck.dll mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll

SinghRajenM:
Not much attention paid..... closing.

Issue #2403, opened 2016-10-07
when switching from mono-instance to multi-instance, user has to close each tab individually


bennetthaselton:
### Steps to Reproduce the Issue

1. With Notepad++ still in mono-instance mode (the default), open two or more files, say a.txt and b.txt. In mono-instance mode, a single instance of Notepad++ shows each file as a separate tab, as expected.
2. Go to Settings->Preferences->Multi-instance and switch from "Default (mono-instance)" to "Always in multi-instance mode".  Note the text at the bottom of the dialog: "*The modification of this setting needs to restart Notepad++".
3. Close Notepad++
4. In Windows Explorer double-click to open a text file.

(Note that Actual Behavior and Expected Behavior are switched below, because Expected Behavior involves several possibilities, each of which are variations on current Actual Behavior.)
### Actual Behavior

Notepad++ opens up still in mono-instance mode, with one tab for each file.  The files that you had open in step 1 above will still be open in tabs.  Even if you open another file, it will open as an additinoal tab.  As far as I can tell, in order to truly switch from mono-instance mode to multi-instance mode, you have to close each tab individually that is open in mono-instance mode, and then you can start opening files in multi-instance mode.
### Expected Behavior

I think there are several possibilities that would be more user-friendly:
(A) At minimum, if you want to keep current behavior, then I think the text at the bottom of the dialog box in step 2 above should say "Changing this setting requires Notepad++ to be restarted.  If you change from mono-instance mode to multi-instance mode, you must close each individual tab open in mono-instance mode, and then Notepad++ can be reopened in multi-instance mode."  This requires only a text change. (And fixes the bag grammar that is there now :) )
(B) After switching from mono-instance mode to multi-instance mode and closing Notepad++, next time you run it, open only the file that was opened (or a new file), without any tabs.  The drawback here is that the user may had several files open that they were working on, and now those files don't come back after re-starting Notepad.
(C) After switching from mono-instance mode to multi-instance mode and closing Notepad++, each file that was previously open in a tab should open in a new window.  This prevents the user from losing open documents, but some users don't find it a good experience if a bunch of windows open suddenly (even if it's the same documents they had open previously in tabs).
[...]

PR #14834, opened 2024-03-06
Updated to Scintilla 5.4.2 & Lexilla 5.3.1

chcg:
Added [Release 5.4.3](https://www.scintilla.org/scintilla543.zip)

    Released 9 March 2024.
    Fix redo failure introduced with 5.4.2. [Bug #2432](https://sourceforge.net/p/scintilla/bugs/2432/).
    Add SC_AUTOCOMPLETE_SELECT_FIRST_ITEM option to always selects the first item in the autocompletion list. [Bug #2403](https://sourceforge.net/p/scintilla/bugs/2403/).

https://www.scintilla.org/scintilla542.zip
Release 5.4.2

    Released 5 March 2024.
    Significantly reduce memory used for undo actions, often to a half or quarter of previous versions. Feature #1458.
    Add APIs for saving and restoring undo history.
    For GTK, when laying out text, detect runs with both left-to-right and right-to-left ranges and divide into an ASCII prefix and more complex suffix. Lay out the ASCII prefix in the standard manner but, for the suffix, measure the whole width and spread that over the suffix bytes. This produces more usable results where the caret moves over the ASCII prefix correctly and over the suffix reasonably but not accurately.
    For ScintillaEdit on Qt, fix reference from ScintillaDocument to Document to match change in 5.4.1 using IDocumentEditable for SCI_GETDOCPOINTER and SCI_SETDOCPOINTER.
    For Direct2D on Win32, use the multi-threaded option to avoid crashes when Scintilla instances created on different threads. There may be more problems with this scenario so it should be avoided. Bug #2420.
    For Win32, ensure keyboard-initiated context menu appears in multi-screen situations.

https://www.scintilla.org/lexilla531.zip
Release 5.3.1

    Released 5 March 2024.
    Assembler: After comments, treat \r\n line ends the same as \n. This makes testing easier.
    Bash: Fix folding when line changed to/from comment and previous line is comment. Issue #224.
    Batch: Fix handling ':' next to keywords. Issue #222.
    JavaScript: in cpp lexer, add lexer.cpp.backquoted.strings=2 mode to treat ` back-quoted strings as template literals which allow embedded ${expressions}. Issue #94.
    Python: fix lexing of rb'' and rf'' strings. Issue #223, Pull request #227.
    Ruby: fix lexing of methods on numeric literals like '3.times' so the '.' and method name do not appear in numeric style. Issue #225.


mpheath:
Looks like Notepad++ and Scintilla committed code are competing with the same Scintilla context menu (originally handled by Scintilla).

Notepad++: [Fix context menu popup location issue](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/fbd4a402143082178a006891e3a055dd332b334f)

Scintilla: [For Win32, ensure keyboard-initiated context menu appears in multi-screen situations](https://sourceforge.net/p/scintilla/code/ci/e40542d4b507adfcb300b6c21e8bc224026e76b7/)

alankilborn:
>Looks like Notepad++ and Scintilla committed code are competing...

Notepad++ has "always" handled this popup menu, so I don't think there is any "competition"...

rdipardo:
> JavaScript: in cpp lexer, add lexer.cpp.backquoted.strings=2 mode to treat ` back-quoted strings as template literals which allow embedded ${expressions}.

To benefit from this, N++ has to configure `lexer.cpp.backquoted.strings` and provide a new `<WordsStyle name="STRING RAW" styleID="20" .... />` for C++.

All that just to close #14539 and [a long list](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14539#issuecomment-1874791869) of duplicates going back long before VS Code became the JavaScript IDE *du jour*.

Is it even worth the bother?

PR #14842, opened 2024-03-08
adding FlushFileBuffers-fail user notification

xomx:
Helps in fixing the #6133 .

With this PR the affected users will be at least alerted that something bad is happening.

Simulated case message:
npp-ffb-fail-notification-smaller

### DIFF
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -92,6 +92,8 @@ void Win32_IO_File::close()
 {
 	if (isOpened())
 	{
+		NppParameters& nppParam = NppParameters::getInstance();
+
 		DWORD flushError = NOERROR;
 		if (_written)
 		{
@@ -99,16 +101,17 @@ void Win32_IO_File::close()
 			{
 				flushError = ::GetLastError();
 
-				if (!NppParameters::getInstance().isEndSessionCritical())
+				if (!nppParam.isEndSessionCritical())
 				{
-					// at least alert the user that the file data could not actually be saved
+					// because of there is not an externally forced shutdown/restart of Windows in progress,
+					// we can at least alert the user that the file data could not actually be saved
 
 					std::wstring curFilePath;
 					const DWORD cchPathBuf = MAX_PATH + 128;
 					WCHAR pathbuf[cchPathBuf]{};
 					// the dwFlags used below are the most error-proof and informative
 					DWORD dwRet = ::GetFinalPathNameByHandle(_hFile, pathbuf, cchPathBuf, FILE_NAME_OPENED | VOLUME_NAME_NT);
-					if ( (dwRet == 0) || (dwRet >= cchPathBuf))
+					if ((dwRet == 0) || (dwRet >= cchPathBuf))
 					{
 						// probably insufficient path-buffer length, the classic style must suffice
 						std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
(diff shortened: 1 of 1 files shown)


xomx:
@donho 
During testing, I discovered the possibility that under certain conditions the N++ can crash (and thus possibly cause some disk corruptions - but this I have not proven yet) when using snapshot backups. It is all about the ability to recursively (ab)use the N++ backup-mutex. I will investigate and let you know.

PowerEditor/src/MISC/Common/FileInterface.cpp
```diff
@@ -96,7 +96,44 @@ void Win32_IO_File::close()
 		if (_written)
 		{
 			if (!::FlushFileBuffers(_hFile))
+			{
 				flushError = ::GetLastError();
+
+				if (!NppParameters::getInstance().isEndSessionCritical())
```
donho, line 102: If FlushFileBuffers returns FALSE in any time, it's critical, isn't it?
So instead of isEndSessionCritical(), we can use nppParam.isEndSessionStarted() as condition.

If what I said makes sens then we can merge the following block into this one:
```cpp
if (nppParam.isEndSessionStarted() && nppParam.doNppLogNulContentCorruptionIssue())
{
...
}
```
so once we have flush error in the end of session, we can send a user alert and also write the error into the log.

donho (changes requested):
Please consider to merge the 2 treatments into one block.

PowerEditor/src/MISC/Common/FileInterface.cpp
xomx: > If FlushFileBuffers returns FALSE in any time, it's critical, isn't it?

 Yes.

> So instead of isEndSessionCritical(), we can use nppParam.isEndSessionStarted() as condition.

No.

Precisely because of what you said before (that it is critical when that func returns an error), we must always show this error to users but with only one exception - unless an externally forced shutdown of Windows is already in progress. In that case, the 2nd block starting with the `if (nppParam.isEndSessionStarted() && nppParam.doNppLogNulContentCorruptionIssue())` will do the FlushFileBuffers-fail report instead..

donho: >> If FlushFileBuffers returns FALSE in any time, it's critical, isn't it?

> Yes.

If the answer is YES, then the reply of following sentence:

> So instead of isEndSessionCritical(), we can use nppParam.isEndSessionStarted() as condition.

should not be NO.

Please consider the following lines:
```cpp
			nppParam.endSessionStart();
			if (isForcedShuttingDown)
				nppParam.makeEndSessionCritical();
```
in https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppBigSwitch.cpp#L2350

You can see `isEndSessionCritical` is a condition much stronger (or much more restrict) than `isEndSessionStarted` condition. 

According the code of this PR, the error of flush will be displayed to users on ONLY **force shutdown of WM_QUERYENDSESSION**, whereas the error will be written into log on **WM_QUERYENDSESSION**.

~~If I'm not right, then I must be left, because I can't be wrong :P~~
If I'm not right, please tell me of which I am not aware.

xomx: > You can see `isEndSessionCritical` is a condition much stronger (or much more restrict) than `isEndSessionStarted` condition.

Yes, but we are still allowed to show the users a msgbox while it is only in the `isEndSessionStarted` phase but not in the `isEndSessionCritical` one (a msgbox in that one will forever block the N++ from exiting at the shutdown time and the OS will forcibly terminate N++ without any N++ tidy-up code or saving). The not so critical `isEndSessionStarted` phase can even lead to canceling the Windows shutdown. 

> According the code of this PR, the error of flush will be displayed to users on ONLY **force shutdown of WM_QUERYENDSESSION**,

No, there is the `!` before the `nppParam.isEndSessionCritical()`, so "NOT in case of isEndSessionCritical".

> whereas the error will be written into log on WM_QUERYENDSESSION.

Yes.

donho: > No, there is the `!` before the `nppParam.isEndSessionCritical()`, so "NOT in case of isEndSessionCritical".

Oups! My bad! I didn't see `!` in front of `nppParam.isEndSessionCritical()`.

donho (approved):

PowerEditor/src/MISC/Common/FileInterface.cpp
xomx: I forgot to add that our critical-flag is also set here, when the shutdown is already inevitable:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d92b0a86939e391a8fc709c97de0929532b986f/PowerEditor/src/NppBigSwitch.cpp#L2528

PR #14844, opened 2024-03-08
Code optimization

### LINKED ISSUE #14843: Small code optimization
### Description of the Issue
- Scope of var wmc is bigger than needed in FileManager::saveBuffer()
- grabSize is not initialized
- remove unused var in TreeView::searchLeafRecusivelyAndBuildTree()
- missing nullpointer check in ScintillaEditView::init()

N/A

### Expected Behavior
N/A

### Actual Behavior
N/A

### Debug Information
N/A

chcg:
fixes #14843 

- init grabSize
- optimize scope of WcharMbcsConvertor instance
- remove unused var in TreeView::searchLeafRecusivelyAndBuildTree()
- missing nullpointer check in ScintillaEditView::init()

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1054,7 +1054,7 @@ bool FileManager::backupCurrentBuffer()
 				else
 				{
 					WcharMbcsConvertor& wmc = WcharMbcsConvertor::getInstance();
-					size_t grabSize;
+					size_t grabSize = 0;
 					for (size_t i = 0; i < lengthDoc; i += grabSize)
 					{
 						grabSize = lengthDoc - i;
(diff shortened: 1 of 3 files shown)


chcg:
@donho I added some further fixes from code analysis runs. I hope that's ok.

donho (approved):

chcg:
grabSize is set inside of the loop to `lengthDoc - i`. So I think it is no actual issue.

vlakoff:
The initialization isn't used actually, as the `i += grabSize` increment of the loop is executed only starting from the 2nd iteration, and inside the loop there is the `grabSize = lengthDoc - i` assignment that is executed starting from the 1st iteration.

This made me suspect that the control structure might be improved. Thus, I may suggest the following:

```c++
size_t i = 0;

while (i < lengthDoc) {
	size_t grabSize = lengthDoc - i;

	if (grabSize > blocSize) {
		grabSize = blocSize;
	}

	int newDataLen = 0;
	int incompleteMultibyteChar = 0;
	const char* newData = wmc.encode(SC_CP_UTF8, encoding, buf + i, static_cast<int>(grabSize), &newDataLen, &incompleteMultibyteChar);
	grabSize -= incompleteMultibyteChar;
	isWrittenSuccessful = UnicodeConvertor.writeFile(newData, newDataLen);

	i += grabSize;
}
```

It does the same thing, but I think it is clearer, and it allows for a locally scoped `grabSize`.

donho:
@vlakoff 
> The initialization isn't used actually, as the i += grabSize increment of the loop is executed only starting from the 2nd iteration

You're right about it. However, by the principle that all variables should be initialized on its declaration, I think the current "fix" makes sense.

> It does the same thing, but I think it is clearer, and it allows for a locally scoped grabSize.

Thank you for your suggestion. But let's keep using the current code which already works.

PR #14855, opened 2024-03-13
Use undocumented API to support win7

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14818#issuecomment-1993472569


ozone10:
@astewart77
could you test?

astewart77:
It runs without any missing entry point. Just a quick launch of Notepad++.MSVC.Win32.Release. I have multiple monitors, but none are hiDPI. Looks OK for me. Thanks.

donho:
By "undocumented" API you mean `SystemParametersInfoForDpi`, `GetSystemMetricsForDpi`, `GetDpiForWindow` & `GetDpiForSystem`?

They are documented here:
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfofordpi
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemmetricsfordpi
https://learn.microsoft.com/fr-fr/windows/win32/api/winuser/nf-winuser-getdpiforwindow
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdpiforsystem

or am I missing something?

MarkusBodensee:
Thank you for including c8a2bcb6b0c32c93fd0501378ebaeaa25fd6e984 and for continuing Windows 7 support :-)

[#commitcomment-139807687](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c8a2bcb6b0c32c93fd0501378ebaeaa25fd6e984#commitcomment-139807687)

ozone10:
@donho 
yes. But even if unused in older OS, they will try to be loaded, which will cause error about missing entry point
`The procedure entry point GetDpiForWindow could not be located in the dynamic link library USER32.dll.`

I will also add other dpi functions this way, if needed, to not block Windows 7 - 8.1 and very old Windows 10 builds.

astewart77:
Well, everything looked good until I tried to open the Shortcut Mapper. The dialog did not open.
 `N++ has stopped working`
`Debug the program`
`An unhandled win32 exception occurred in [9168] notepad++.exe.`
I don't have debugging tools installed, so I can't carry it further. I haven't had any other problem so far.

ozone10:
@astewart77 
I cannot reproduce it.

MarkusBodensee:
@ozone10 @astewart77 
I cannot reproduce it, too. Tested it with the x64 version and it opened as before.

astewart77:
I tried an fresh x64 minimalist install with this .exe and did not get the error. A fresh minimalist x86 does fail for me.

xomx:
@ozone10 
I can reproduce the x86 crash on my old Win7 PC.

It is probably some uninitialized stuff as the MSVC.Win32.Debug binary works fine. The binary in question quits with code 3, which probably means that the abort() has been called in it.

I do not have the needed MSVS2022 there for debugging (it is unsupported under the Win7 but up to the v17.6 LTS it should be at least installable). If you install it, then you can try to switch temporarily off the Win32 Release build compiler & linker optimization and re-test to confirm my theory...

ozone10:
@xomx 
it also crash on Windows 10 and 11
It seems to be caused by [/Og](https://learn.microsoft.com/en-us/cpp/build/reference/og-global-optimizations?view=msvc-170)

Further discussion should be in https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14871

Issue #14839, opened 2024-03-08
The function "PathFileExists" causes confusion in the judgment of directories and files.


softmgr:
The program extensively uses the function "PathFileExists" to determine directories and files.
In the source code, the judgments about directories and files are confusing.
For example, if the directory "plugins" or "backup" is a file, it will also be judged as successful, but there may be problems with the subsequent execution of the program.

alankilborn:
This is not a properly formed issue.  Please follow the issue template or you risk closure without further comment.

xomx:
@alankilborn 
I think I know what the OP is trying to say. I also ran into this when I recently experimented with replacing this function in the N++ code with my own non-blocking cancellable variant (to solve the issues here with suddenly inaccessible network resources...). 

For example, N++ code here uses this function to determine if a file exists:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/83f0232d910444c6683a5dc104732b7c73c53b87/PowerEditor/src/MISC/Common/Common.cpp#L53-L61

But if the `file2read` here corresponds to an existing directory (and not to a file), the PathFileExists WINAPI also returns TRUE, which IMO is not what the author of this code-part intended to achieve.

softmgr:
> This is not a properly formed issue. Please follow the issue template or you risk closure without further comment.

Sorry, I'm using the GITHUB mobile app to submit issue, but there are no issue templates in the app interface.

softmgr:
> @alankilborn I think I know what the OP is trying to say. I also ran into this when I recently experimented with replacing this function in the N++ code with my own non-blocking cancellable variant (to solve the issues here with suddenly inaccessible network resources...).
> 
> For example, N++ code here uses this function to determine if a file exists:
> 
> https://github.com/notepad-plus-plus/notepad-plus-plus/blob/83f0232d910444c6683a5dc104732b7c73c53b87/PowerEditor/src/MISC/Common/Common.cpp#L53-L61
> 
> But if the `file2read` here corresponds to an existing directory (and not to a file), the PathFileExists WINAPI also returns TRUE, which IMO is not what the author of this code-part intended to achieve.

Add this function to Common.h and Common.cpp, I have tested.
`bool fileExists(const TCHAR* filename);`

`bool fileExists(const TCHAR* filename) {`
`	WIN32_FILE_ATTRIBUTE_DATA fileData;`
`	// Append special suffix \\?\ to allow path lengths up to 32767`
[...]

PR #14858, opened 2024-03-14
Remedy losing session problem after the power outrages

### LINKED ISSUE #14781: Tabs/session lost after forced shutdown
### Description of the Issue

This is now the second time that I have lost the session (that is, the open tabs) after a forced shutdown. 

1. Open Notepad++ and open a few files.
2. Force a shutdown of the computer (long press on the power button on most laptops).
3. Boot the computer and open Notepad++.

### Expected Behavior

The prior session opens; that is, the same tabs are open as before.

### Actual Behavior

The session is empty; no files are open.

### Debug Information

```
Notepad++ v8.6.2   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 22621.3155
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3)
    NppConverter (4.5)
    NppExport (0.4)
    NppPluginDemo (4.3)
```

donho:
This PR doesn't really "fix" the issue #14781. Instead, it provides a workaround: 
In case of "session.xml" being corrupted after the power outrages, user can get **"session.xml.inCaseOfCorruption.bak"** - the before last session.xml backup file.

Fix #14781

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -3591,10 +3591,14 @@ void NppParameters::writeSession(const Session & session, const TCHAR *fileName)
 {
 	const TCHAR *sessionPathName = fileName ? fileName : _sessionPath.c_str();
 
+	//
 	// Make sure session file is not read-only
+	//
 	removeReadOnlyFlagFromFileAttributes(sessionPathName);
 
+	// 
 	// Backup session file before overriting it
+	//
 	TCHAR backupPathName[MAX_PATH]{};
 	BOOL doesBackupCopyExist = FALSE;
 	if (PathFileExists(sessionPathName))
@@ -3613,11 +3617,12 @@ void NppParameters::writeSession(const Session & session, const TCHAR *fileName)
 		}
 	}
 
+	//
+	// Prepare for writing
+	//
 	TiXmlDocument* pXmlSessionDoc = new TiXmlDocument(sessionPathName);
-
 	TiXmlDeclaration* decl = new TiXmlDeclaration(TEXT("1.0"), TEXT("UTF-8"), TEXT(""));
 	pXmlSessionDoc->LinkEndChild(decl);
-
 	TiXmlNode *root = pXmlSessionDoc->InsertEndChild(TiXmlElement(TEXT("NotepadPlus")));
 
 	if (root)
@@ -3709,20 +3714,28 @@ void NppParameters::writeSession(const Session & session, const TCHAR *fileName)
 		}
 	}
 
+	//
+	// Write the session file
+	//
 	bool sessionSaveOK = pXmlSessionDoc->SaveFile();
 
-	if (!sessionSaveOK)
+	//
+	// Double checking: prevent written session file corrupted while writting
+	//
+	if (sessionSaveOK)
 	{
-		::MessageBox(nullptr, sessionPathName, L"Error of saving session XML file", MB_OK | MB_APPLMODAL | MB_ICONWARNING);
-	}
-	else
-	{
-		// Double checking: prevent written session file corrupted while writting
 		TiXmlDocument* pXmlSessionCheck = new TiXmlDocument(sessionPathName);
 		sessionSaveOK = pXmlSessionCheck->LoadFile();
 		delete pXmlSessionCheck;
 	}
+	else
+	{
+		::MessageBox(nullptr, sessionPathName, L"Error of saving session XML file", MB_OK | MB_APPLMODAL | MB_ICONWARNING);
+	}
 
+	//
+	// If error after double checking, restore session backup file
+	//
 	if (!sessionSaveOK)
 	{
 		if (doesBackupCopyExist) // session backup file exists, restore it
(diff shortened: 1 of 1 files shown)


donho:
@xomx @pnedev 
Could you review this PR please?
As @pnedev has said, there's nothing more we can do about it. So I remove the part of deletion session backup file - user can get back their before last session which is under the name "session.xml.inCaseOfCorruption.bak".

pnedev:
@donho ,
Looks good to me. :+1:

xomx:
@donho 
I have debugged this PR and found these problems (I put all my notes here at one place, if you want me to do also a review in the code, please let me know):


1. Instead of the used `L"Saving session error - restore backup"`, I would rather choose the wording `L"Saving session error - restoring from the backup:"` in the `NppParameters::writeSession` msgbox.

2. Wrong `ReplaceFile` use (or I do not get your intent).
    - I simulated a session.xml write failure - while running under the debugger, immediately after the session.xml file was successfully saved in the `NppParameters::WriteSession`, I paused the N++ and externally corrupted the contents of the session.xml file (e.g. to simulate the NUL-bytes problem...) and then let the code continue to the ReplaceFile part
    - the above resulted in: no session.xml file at all, session.xml.inCaseOfCorruption.bak file containing the incorrect session.xml data and session.xml.fail2Load file created containing the right (previous backup) data - is this what you intended?

3. Consider please also the following addition of the content-corruption check of the written session.xml in the `NppParameters::WriteSession` (this immediately reveals e.g. the possible NULs...)
```
  //
  // Double checking: prevent written session file corrupted while writting
  //
  if (sessionSaveOK)
  {
  	TiXmlDocument* pXmlSessionCheck = new TiXmlDocument(sessionPathName);
  	sessionSaveOK = pXmlSessionCheck->LoadFile();
  	if (sessionSaveOK)
  	{
  		Session sessionCheck;
  		sessionSaveOK = getSessionFromXmlTree(pXmlSessionCheck, sessionCheck);
  	}
  	delete pXmlSessionCheck;
  }
  else
```

4. Adjust please also the session.xml [loading part](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d09ee818a2effcb54e4e7452dbd2c377bed835b/PowerEditor/src/Parameters.cpp#L1573-L1587) in the `NppParameters::load()` (and maybe also in the `NppParameters::loadSession`)
    - If, for any reason, it is not possible to write a new session.xml file (or it is then damaged somehow) but there will be a valid backup copy session.xml.inCaseOfCorruption.bak, after restarting the N++, the backup will not be used automatically, but an empty session will be created, which will even overwrite the existing backup file when the N++ gets closed!
    - By the above I mean this scenario - the N++ user usually discovers this problem by opening the N++ and finding that their session is empty. If he/she closes the N++ at this point before starts looking for the backup, that backup file could be destroyed by overwriting with the current empty-session when the N++ quits...
    - so in case of a session.xml loading fail, but when the session.xml.inCaseOfCorruption.bak exists, show also a modal msgbox informing the user that the session.xml.inCaseOfCorruption.bak has been found and will be used instead of the session.xml file (+ add such loading code from the session.xml.inCaseOfCorruption.bak)

5. Complement the existing session.xml loading code part above (4.) with the same [content checking](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d09ee818a2effcb54e4e7452dbd2c377bed835b/PowerEditor/src/Parameters.cpp#L2305) as used for the `NppParameters::loadSession` func.
   - Is there a reason why the [NppParameters::loadSession](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d09ee818a2effcb54e4e7452dbd2c377bed835b/PowerEditor/src/Parameters.cpp#L2300) is not re-used at the startup session.xml [loading place](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d09ee818a2effcb54e4e7452dbd2c377bed835b/PowerEditor/src/Parameters.cpp#L1573-L1587) ?

6. Consider adding of the `!nppParam.isEndSessionCritical()` before any possible msgbox in the `NppParameters::WriteSession`
    - this critical-flag is also [set](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d92b0a86939e391a8fc709c97de0929532b986f/PowerEditor/src/NppBigSwitch.cpp#L2528) when the shutdown is already inevitable
    - because of the N++ session.xml is written at the app-exit, we should take this into account and not use the blocking modal msgbox when Windows shutdown is already in progress
 
One more general note - improper use of the the PathFileExists WINAPI in the `NppParameters::WriteSession` (but this is a common problem in the N++ code base, more in the #14839 ).

donho:
@xomx 
Thank you. Very good code review.
> Instead of the used L"Saving session error - restore backup", I would rather choose the wording L"Saving session error - restoring from the backup:" in the NppParameters::writeSession msgbox.

Done in the latest commit.

> Wrong ReplaceFile use (or I do not get your intent).

Fixed in the latest commit - tested also with your debug break point way then modify session.xml.

> Consider please also the following addition of the content-corruption check of the written session.xml in the NppParameters::WriteSession (this immediately reveals e.g. the possible NULs...)

Done in the latest commit - it detects more the session file error indeed.

> Adjust please also the session.xml [loading part](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d09ee818a2effcb54e4e7452dbd2c377bed835b/PowerEditor/src/Parameters.cpp#L1573-L1587) in the NppParameters::load() (and maybe also in the NppParameters::loadSession)

Good thinking!
Done in the latest commit. Tested also with your debug break point way then modify session.xml.

> Complement the existing session.xml loading code part above (4.) with the same [content checking](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/9d09ee818a2effcb54e4e7452dbd2c377bed835b/PowerEditor/src/Parameters.cpp#L2305) as used for the NppParameters::loadSession func.

NppParameters::loadSession()` method is for loading session manually via File command "Load Session..." & it contains messagebox warning, whereas in `NppParameters::load()` there should be no message warning, and there are more specific stuff as added in the current PR. However, I will check in the future for the possibility of refactoring.

> Consider adding of the !nppParam.isEndSessionCritical() before any possible msgbox in the NppParameters::WriteSession

Done in the latest commit.

xomx:
@donho 
I have found some problems while debugging the new code and I will do an adjustment PR for that.

The N++ startup `NppParameters::load()` session-loading code does not handle right the situation where there is a session.xml.inCaseOfCorruption.bak file but no session.xml file. My patch will be likewise the:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0bc28df616fd8ecd2f13f111501e5d7c53bcb6ee/PowerEditor/src/ScintillaComponent/Buffer.cpp#L1080-L1083

Also it should check for a possible ReplaceFile error return.

PR #14860, opened 2024-03-14
Fix NUL characters file corruption after power outrages (1st step solution)

### LINKED ISSUE #6133: The NUL file-corruption bug appears to not be completely gone (v7.7.1 reported)
Unfortunately, Maddock Emerson has [reported in the Community forum](https://notepad-plus-plus.org/community/topic/18206/notepad-just-deleted-all-the-text-from-my-files/9) that he has seen the file-replaced-by-NULs in Notepad++ v7.7.1, when he allowed the OS to reboot due to OS update.

This means that abc6bc1 (and @pnedev's changes that led up to that commit) appears to not have been sufficient to eliminate all the possible conditions that can cause the NUL file.  If @pnedev is able to join into the forum discussion, as the one who made the fixes for v7.5.9, he might be able to elicit more info from Maddock Emerson about the problem, and see what's different about this circumstance than the ones that were fixed.

donho:
### Sernario:
When a user modifies a file in Notepad++, and the time of periodic backup (defaulted to 7 seconds) is reached, the backup of the modified file is being written. However, if a power outage occurs during this precise moment while the file is being written, file corruption may occur.

### Remedy:
The goal is to maintain a non-corrupted file sample even during power outages. Here are the steps:

0. Begin
1. Write the file A as A.temp
2. Replace A by A.temp
3. End

During these steps, the cutoff can happen at any moment, but the user will always have a non-corrupted file sample (either A or A.temp).

### Note:
The solution is applied only to "new #" files, since these files are generally small in length and do not have a second "physical" file existing on the hard drive.

ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133#issuecomment-1987037043

Fix #6133

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -1028,7 +1028,7 @@ bool FileManager::backupCurrentBuffer()
 				hasModifForSession = true;
 			}
 
-			TCHAR fullpath[MAX_PATH];
+			TCHAR fullpath[MAX_PATH]{};
 			::GetFullPathName(backupFilePath.c_str(), MAX_PATH, fullpath, NULL);
 			if (wcschr(fullpath, '~'))
 			{
@@ -1038,8 +1038,9 @@ bool FileManager::backupCurrentBuffer()
 			// Make sure the backup file is not read only
 			removeReadOnlyFlagFromFileAttributes(fullpath);
 
-
-			if (UnicodeConvertor.openFile(fullpath))
+			std::wstring fullpathTemp = fullpath;
+			fullpathTemp += L".tmp";
+			if (UnicodeConvertor.openFile(buffer->isUntitled() ? fullpathTemp.c_str() : fullpath)) // Use temp only for "new #" due to they don't have the original physical existance on the hard drive
 			{
 				size_t lengthDoc = _pNotepadPlus->_pEditView->getCurrentDocLen();
 				char* buf = (char*)_pNotepadPlus->_pEditView->execute(SCI_GETCHARACTERPOINTER);	//to get characters directly from Scintilla buffer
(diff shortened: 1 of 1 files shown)


donho:
@pnedev @xomx 
Please review this PR.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -1074,6 +1075,14 @@ bool FileManager::backupCurrentBuffer()
 
 				if (isWrittenSuccessful) // backup file has been saved
 				{
+					if (buffer->isUntitled()) // "new #" file is saved successfully, then we replace its only physical existence by its temp
+					{
+						if (::PathFileExists(fullpath))
+							::ReplaceFile(fullpath, fullpathTemp.c_str(), nullptr, REPLACEFILE_IGNORE_MERGE_ERRORS | REPLACEFILE_IGNORE_ACL_ERRORS, 0, 0);
+						else
+							::MoveFileEx(fullpathTemp.c_str(), fullpath, MOVEFILE_REPLACE_EXISTING);
+					}
```
donho, line 1084: @xomx 
The above block can use only `::MoveFileEx(fullpathTemp.c_str(), fullpath, MOVEFILE_REPLACE_EXISTING);`.
Is there any advantage to combined with `::ReplaceFile(fullpath, fullpathTemp.c_str(), nullptr, REPLACEFILE_IGNORE_MERGE_ERRORS | REPLACEFILE_IGNORE_ACL_ERRORS, 0, 0);` as you suggested?

pnedev:
@donho ,
Looks good to me :+1:

xomx:
@donho 
I will try to check this tomorrow.

donho:
@xomx 
No rush, take your time for reviewing the PR.
I'll merge it into master for v8.6.5 Release Candidate. Even it's merged into master, we are still in RC stage.
So RC2 could be done for the optimisation of the PR if it's necessary.

xomx:
@donho 
Thanks for the patience, I know that you would like to release the 8.6.5, but unfortunately not enough time at my side just now.
I know about at least one serious problem in the now closed #14858 (will report there), I did not even started checking this one :-(

donho:
Thank you @xomx  !

PR #14866, opened 2024-03-15
[xml] Update romanian.xml

AndreiMiloiu:
I came back with a better translation.


donho:
@AndreiMiloiu 
**You still didn't test your translation.**
You should download the current master CI binary, for example, here:
https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/8309682752
and use it for checking if your translation have covered new added entries:
https://npp-user-manual.org/docs/binary-translation/

AndreiMiloiu:
I understand. I'll try again.

PR #14871, opened 2024-03-17
Temporarily disable new DPI API for MSVC 32bit release build

ozone10:
to avoid crash with Shortcut mapper

ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14855#issuecomment-2002031940

### DIFF
--- a/PowerEditor/src/winmain.cpp
+++ b/PowerEditor/src/winmain.cpp
@@ -503,7 +503,13 @@ int WINAPI wWinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE /*hPrevInstance
 	NppGUI & nppGui = nppParameters.getNppGUI();
 
 	NppDarkMode::initDarkMode();
+
+	// There is issue with MSVC x86 release build,
+	// some optimalization will cause crash when opening shortcut mapper.
+	// Needs further investigation.
+#if defined(_WIN32) && (defined(_WIN64) || defined(DEBUG))
 	DPIManagerV2::initDpiAPI();
+#endif
 
 	bool doUpdateNpp = nppGui._autoUpdateOpt._doAutoUpdate;
 	bool doUpdatePluginList = nppGui._autoUpdateOpt._doAutoUpdate;
--- a/PowerEditor/src/dpiManagerV2.cpp
+++ b/PowerEditor/src/dpiManagerV2.cpp
@@ -29,10 +29,10 @@ bool ptrFn(HMODULE handle, P& pointer, const char* name)
 	return false;
 }
 
-using fnGetDpiForSystem = UINT (WINAPI*)();
-using fnGetDpiForWindow = UINT (WINAPI*)(HWND);
-using fnGetSystemMetricsForDpi = int (WINAPI*)(int, UINT);
-using fnSystemParametersInfoForDpi = BOOL(WINAPI*)(UINT, UINT, PVOID, UINT, UINT);
+using fnGetDpiForSystem = UINT (WINAPI*)(VOID);
+using fnGetDpiForWindow = UINT (WINAPI*)(HWND hwnd);
+using fnGetSystemMetricsForDpi = int (WINAPI*)(int nIndex, UINT dpi);
+using fnSystemParametersInfoForDpi = BOOL (WINAPI*)(UINT uiAction, UINT uiParam, PVOID pvParam, UINT fWinIni, UINT dpi);
 
 fnGetDpiForSystem _fnGetDpiForSystem = nullptr;
 fnGetDpiForWindow _fnGetDpiForWindow = nullptr;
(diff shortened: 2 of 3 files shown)


PowerEditor/src/winmain.cpp
```diff
@@ -503,7 +503,13 @@ int WINAPI wWinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE /*hPrevInstance
 	NppGUI & nppGui = nppParameters.getNppGUI();
 
 	NppDarkMode::initDarkMode();
+
+	// There is issue with MSVC x86 release build,
+	// some optimalization will cause crash when opening shortcut mapper.
+	// Needs further investigation.
+#if defined(_WIN32) && (defined(_WIN64) || defined(DEBUG))
 	DPIManagerV2::initDpiAPI();
```
ozone10, line 511: Disabling `NppDarkMode::initDarkMode();` instead of `DPIManagerV2::initDpiAPI();` will also prevent crash.
Making `DPIManagerV2::initDpiAPI()` do nothing will also cause crash.
GCC builds works fine.

ozone10:
@donho 
there is issue with MSVC 32bit release build. My debug skills are not enough good to figure what is it causing.
I suspect that it is caused by buffer overrun somewhere.

@astewart77 @MarkusBodensee
could you check GCC 32 bit builds, if they crash when opening Shortcut mapper.
I could only reproduce it with MSVC 32 bit release build.

ozone10:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14855#issuecomment-2002434176
> it also crashes on Windows 10 and 11 
> It seems to be caused by [/Og](https://learn.microsoft.com/en-us/cpp/build/reference/og-global-optimizations?view=msvc-170)

xomx:
@ozone10 

I do not quite understand it yet, but it is crashing here:
- last N++ code-line before the crash is calling the CRT `_wtoi` (the N++ wide `buffer` content is ok there - `L"00000"`): 
  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/45cdbbfe15e4bb6622cc86c206a7ea47afce5f72/PowerEditor/src/WinControls/Grid/BabyGrid.cpp#L1699
- `_wtoi` call crashes when reaches the CRT `__acrt_update_thread_locale_data` code in the "C:\Program Files (x86)\Windows Kits\10\Source\10.0.22621.0\ucrt\locale\locale_refcounting.cpp", line 291, it is calling the `abort()` as I expected:
```
  if (!ptloci)
  {
      abort();
  }
```

So it seems to be connected with some locale-stuff but I do not understand why, as the calling of the `_wtoi(oneparamonly)` results in the CRT:
```
extern "C" int __cdecl _wtoi(wchar_t const* const string)
{
    // Note:  We parse as a long to avoid an extra specialization of parse_integer_from_string
    return __crt_strtox::parse_integer_from_string<long>(string, nullptr, 10, nullptr);
}
````
so the locale param is and should be that `nullptr` there...

**Comment by astewart77** [comment]
> could you check GCC 32 bit builds, if they crash when opening Shortcut mapper, you can try builds from this PR.

No Shortcut Mapper crash in either the latest master GCC 32, or MSVC 32 in this PR for me.

File: PowerEditor/src/winmain.cpp [hunk 2]
```diff
@@ -503,7 +503,13 @@ int WINAPI wWinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE /*hPrevInstance
 	NppGUI & nppGui = nppParameters.getNppGUI();
 
 	NppDarkMode::initDarkMode();
+
+	// There is issue with MSVC x86 release build,
+	// some optimalization will cause crash when opening shortcut mapper.
+	// Needs further investigation.
+#if defined(_WIN32) && (defined(_WIN64) || defined(DEBUG))
```
donho, line 510: I'm not convinced by this line... it ***Needs further investigation*** indeed.

donho:
I do believe the culprit is not DpiManager2's code but it has triggered the darkside in *BabyGrid.cpp*.
The code in *BabyGrid.cpp* needs to be modernized and enhanced clearly. 
However, since releasing v8.6.5 RC (in progress) should be secured, I will reverse 2 commits regarding DpiManager2 (#14875) for this release.

donho:
I close this PR now, because of the conflict. 
@ozone10 please feel free to open a new PR.

ozone10:
@xomx 
about locale I figured it too, I also tried disabling some code to always ended with locale issue.

@donho 
ok #14876

xomx:
@ozone10 

> I do believe the culprit is not DpiManager2's code but it has triggered the darkside in _BabyGrid.cpp_.

@donho is right, there is a nasty BabyGrid bug causing extraneous memory overwriting.
That DpiManager2's code is ok, it only helped revealing that long-hidden bug (probably by shifting of some important objects memory into area of overwriting...).

The fix is easy, I will do a quick PR.

PR #14873, opened 2024-03-17
[xml] Update galician v.8.6.5.xml

P0rsche-911:
New revision and update


donho:
@P0rsche-911 
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/aa0be99 part is missing.

PR #14875, opened 2024-03-17
Revert 2 commits of DpiManager2  

donho:
Revert "Use undocumented dpi API to support win7"
This reverts commit c8a2bcb6b0c32c93fd0501378ebaeaa25fd6e984.

Fix the crash reported in: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14855#issuecomment-2002031940


donho:
@ozone10
I do believe the culprit is not DpiManager2's code but it has triggered the darkside in BabyGrid.cpp.
[As pointed](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14871#issuecomment-2002485089) by @xomx, the code in BabyGrid.cpp needs to be modernized and enhanced clearly.
However, since releasing v8.6.5 RC (in progress) should be secured, I will reverse 2 commits regarding DpiManager2 (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14875) for this release.
That leave us more time to find the cause.

PR #14880, opened 2024-03-19
Fix memory overwriting bug by BabyGrid

xomx:
Fix https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14855#issuecomment-2001066992 , https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14871#issuecomment-2002485089

BabyGrid code was overwriting foreign memory on its initialization and deinitialization. At that time (WM_NCCREATE, WM_NCCALCSIZE, WM_CREATE and WM_NCDESTROY) the relevant `FindGrid` func returns -1, which was used as an index pointing to a memory area before the whole `BGHS` object (`BGHS[-1]`...)!

This was a long-standing hidden bug that only started to manifest itself probably when the app memory layout shifted somehow and important objects/data started to be overwritten, resulting in the visible app crashes.

### DIFF
--- a/PowerEditor/src/WinControls/Grid/BabyGrid.cpp
+++ b/PowerEditor/src/WinControls/Grid/BabyGrid.cpp
@@ -1341,7 +1341,7 @@ LRESULT CALLBACK GridProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 
 	SelfIndex=FindGrid(GetMenu(hWnd));
 
-	// the following check will prevent memory overwriting bug by BabyGrip during the WM_NCCREATE, WM_NCCALCSIZE, WM_CREATE and WM_NCDESTROY
+	// the following check will prevent memory overwriting bug by BabyGrid during the WM_NCCREATE, WM_NCCALCSIZE, WM_CREATE and WM_NCDESTROY
 	// - TODO: if the above FindGrid can theoretically return -1, the whole BabyGrid code here is theoretically flawed and needs a review
 	//         (luckily it seems it never returns -1 apart from the aforementioned 4 WMs above)
 	if (SelfIndex != -1)
--- a/PowerEditor/src/WinControls/Grid/BabyGrid.cpp
+++ b/PowerEditor/src/WinControls/Grid/BabyGrid.cpp
@@ -1341,14 +1341,16 @@ LRESULT CALLBACK GridProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 
 	SelfIndex=FindGrid(GetMenu(hWnd));
 
-	//update the grid width and height variable
+	// the following check will prevent memory overwriting bug by BabyGrip during the WM_NCCREATE, WM_NCCALCSIZE, WM_CREATE and WM_NCDESTROY
+	// - TODO: if the above FindGrid can theoretically return -1, the whole BabyGrid code here is theoretically flawed and needs a review
+	//         (luckily it seems it never returns -1 apart from the aforementioned 4 WMs above)
+	if (SelfIndex != -1)
 	{
-	 RECT rect;
-
-	 GetClientRect(hWnd,&rect);
-	 BGHS[SelfIndex].gridwidth = rect.right - rect.left;
-	 BGHS[SelfIndex].gridheight = rect.bottom - rect.top;
-
+		// update the grid width and height variable
+		RECT rect{};
+		::GetClientRect(hWnd, &rect);
+		BGHS[SelfIndex].gridwidth = rect.right - rect.left;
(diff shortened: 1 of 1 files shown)


ozone10:
@xomx 
cool. Could you tell me how did you find out the issue, for future reference, thx.

rdipardo:
> Could you tell me how did you find out the issue, for future reference, thx.

You would need a dissembler to really find out. What @xomx means by "app memory layout" is the stack alignment of the function parameters and local variables. When you add a parameter or declare a new variable, more data gets pushed to the stack, and the existing data has to be aligned differently.

In a 32-bit process, each data item is aligned on a boundary of 4 bytes. If the data's byte count in not a perfect multiple of 4, padding is used to push it into the next boundary. A 32-bit pointer is also 4 bytes. Before the hDPI stuff, the pointer to the Baby Grid probably had padding around it, so a negative index would land in the padded zone where no real data was stored, having no a fatal consequence. A 64-bit process uses 8 byte alignment, which probably results in padding, before and after the hDPI stuff, so the crash was always avoided.

xomx:
> What @xomx means by "app memory layout" is the stack alignment of the function parameters and local variables

> the Baby Grid probably had padding around it, so a negative index would land in the padded zone where no real data was stored, without any fatal consequence.

Not this time, the sizeof(BGHS) > 60kB for sure!
So the BGHS[-1] can influence that much of the app memory...
IIRC from my night-debugging - the @ozone10 hidpi commit needs some extra static-memory and the CRT stuff uses the same mem-segment, so that commit revealed the bug by shifting that mem-layout and caused overwriting of the already preinitialized static CRT internals like the locale-stuff...

> Could you tell me how did you find out the issue

Ok, I will describe what I did to find that later (in short - I used HW memory breakpoints to watch the relevant NULLed CRT mem-address).

ozone10:
@xomx
what about
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/33ac6e7ecfa67715be1d87fc5fb50d66ad5a1ecc/PowerEditor/src/WinControls/Grid/BabyGrid.cpp#L150

it also uses `FindGrid`

PowerEditor/src/WinControls/Grid/BabyGrid.cpp
```diff
@@ -1341,14 +1341,16 @@ LRESULT CALLBACK GridProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 
 	SelfIndex=FindGrid(GetMenu(hWnd));
 
-	//update the grid width and height variable
+	// the following check will prevent memory overwriting bug by BabyGrid during the WM_NCCREATE, WM_NCCALCSIZE, WM_CREATE and WM_NCDESTROY
+	// - TODO: if the above FindGrid can theoretically return -1, the whole BabyGrid code here is theoretically flawed and needs a review
+	//         (luckily it seems it never returns -1 apart from the aforementioned 4 WMs above)
+	if (SelfIndex != -1)
```
donho, line 1347: I just don't understand, with Win32 Debug, why the following 2 lines won't crash or table index exception while SelfIndex == -1:
```cpp

BGHS[SelfIndex].gridwidth = rect.right - rect.left;
BGHS[SelfIndex].gridheight = rect.bottom - rect.top;

```
whereas BGHS has declared for containing only 20 element.
@xomx Any explanation?

donho (changes requested):
`RefreshGrid` is called usually after initialization, but since `FindGrid` could return -1, it might worth to do the following:
```cpp
void RefreshGrid(HWND hWnd)
{
	 RECT rect;
     int SI;
	 GetClientRect(hWnd,&rect);
	 InvalidateRect(hWnd,&rect,FALSE);
	 SI=FindGrid(GetMenu(hWnd));
     if(SI != -1 && BGHS[SI].EDITING)
         {
          DisplayEditString(hWnd, SI, TEXT(""));
         }

}
```

xomx:
sorry for the delay, here is the final proof:

npp-BabyGrid-CRT-locale-init-vs-BGHS

PowerEditor/src/WinControls/Grid/BabyGrid.cpp
xomx: @donho 
Win32.Debug build will have a completely different static memory-segment layout, so it is (silently) overwriting something else.

We are extremely lucky that @ozone10 hidpi commit makes it possible to unhide that.
Otherwise, this nasty bug would continue to merrily corrupt some data in the N++ memory forever, each time with different effects (depending on the actual memory layout only)...

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/73bc79ee502a271bce216b0d8b46beed7ac4f47d

PR #14883, opened 2024-03-19
Change History GUI enhancement and tweaks

### LINKED ISSUE #14882: [regression] Wrong setting for 'Go to setting' for Change History margin
### Description of the Issue

Go to setting' for Change History margin will open Dark mode settings.
This is due to adding Editing 2 settings.

1. Open Style Configurator
2. Select Change History margin style
3. Click on Go to setting

### Expected Behavior

Open Preference on Margin/Border/Edge settings
### Actual Behavior

Open Preference on Dark Mode settings
### Debug Information

Notepad++ v8.6.5 Release Candidate

ozone10:
- add 'Go to setting' for other Change History styles
- make spacing consistent in preference
- minor warning fix


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -3206,25 +3206,52 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 		}
 
 		case NPPM_INTERNAL_CRLFLAUNCHSTYLECONF:
+		case NPPM_INTERNAL_NPCLAUNCHSTYLECONF:
```
ozone10, line 3209: Can be simplified if define directives can be removed in resource.h

PowerEditor/src/Parameters.cpp
```diff
@@ -4140,40 +4140,73 @@ bool NppParameters::feedStylerArray(TiXmlNode *node)
...
+				return result;
+			}
+			return -1;
 		}
+		return result;
+	};
 
-		_widgetStyleArray.addStyler(0, npcColorkNode);
-	}
+	addStyle(L"Change History modified", L"FF8000", L"FF8000");
```
ozone10, line 4203: missing styles will be added at end, they won't be sorted same as in fresh new installs,
but still better than searching for non-existing entry in style configurator

PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
```diff
@@ -1982,6 +1982,18 @@ intptr_t CALLBACK MarginsBorderEdgeSubDlg::run_dlgProc(UINT message, WPARAM wPar
 				// Make tip stay 30 seconds
 				::SendMessage(_verticalEdgeTip, TTM_SETDELAYTIME, TTDT_AUTOPOP, MAKELPARAM((30000), (0)));
 			}
+
+			generic_string tipCHM2Show = pNativeSpeaker->getLocalizedStrFromID("changeHistoryColor-tip",
```
ozone10, line 1986: forgot language files, will add later if approved

donho:
@ozone10 
Despite the fact that there is no additional room for translations after adding the `...` button, I believe it is unnecessary to overly complicate the GUI for customizing these features. The feature in question is less technical than others, such as “Non-printing characters,” which might require color customization for clarity.

OTOH, if we were to consistently link all style colors to the settings, we would need to introduce numerous similar buttons in Preferences dialog. This is precisely what I am trying to avoid.

For these reasons, the PR is rejected. I hope you understand the rationale behind this decision.

ozone10:
@donho 
Ok, what about other enhancement:
-  spacing and positions, it does not look good in hiDPI situations as it's more noticeable
-  missing new change history styles for older users or users with custom theme
-  'Go to setting' for new change history styles, I think at least this should be added, as it might cause confusion to new users, they change color and nothing happen, if indicators are not enabled

donho:
> spacing and positions, it does not look good in hiDPI situations as it's more noticeable

Which part needs to be tuned? Could you provide me 2 screenshots of before and after please?

> missing new change history styles for older users or users with custom theme

I am aware of it. Currently some global styles have the same situation (if a user installs a new version of Notepad++ on an old version, without remove the old stylers.xml). If user really need to customize some styles, he/she can just removed stylers.xml then start Notepad++. Otherwise, rather the styles which are linked with "customized button" are necessary to be inserted.

> 'Go to setting' for new change history styles, I think at least this should be added, as it might cause confusion to new users, they change color and nothing happen, if indicators are not enabled

OK for adding 'Go to setting' for new change history styles.

ozone10:
> Which part needs to be tuned? Could you provide 2 screenshots of before and after please?

edit 1
-  line wrap group box has more distance from caret box, than caret box from current line box
-  buttons in line wrap are more left than in current line box

before
edit1p
after
edit1

edit 2
-  make spacing between boxes and buttons more consistent
-  reduce space between Apply Appearance setting... and Prevent control character... to avoid confusion fro some users that there is missing checkbox between them

before
edit2p
after
edit2

margins
-  make spacing between boxes and buttons more consistent

before
marginp
after
margin



> I am aware of it. Currently some global styles have the same situation (if a user installs a new version of Notepad++ on an old version, without remove the old stylers.xml). If user really need to customize some styles, he/she can just removed stylers.xml then start Notepad++. Otherwise, rather the styles which are linked with "customized button" are necessary to be inserted.

IMHO most users are using different themes than `stylers.xml` or they already customized with different colors, custom ext (removing `stylers.xml` will reset their customization).
Notepad++ is already applying default colors for some styles in case of missing them in themes xml files
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/12c6777bc11b47074b7d9848eac48a91c49b2594/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L2902
so why not to add them in Style configurator to make it more convenient for user.
And if IIRC there were some questions about missing styles in community forums, usually after user updates Notepad++ to newer version.



> OK for adding 'Go to setting' for new change history styles.
OK

donho:
> edit 1
>    line wrap group box has more distance from caret box, than caret box from current line box
>    buttons in line wrap are more left than in current line box

OK

> edit 2
>    make spacing between boxes and buttons more consistent

OK

>    reduce space between Apply Appearance setting... and Prevent control character... to avoid confusion fro some users that there is missing checkbox between them

It's on purpose. 
`"Apply Appearance settings to C0, C1 & Unicode EOL"` proposes to user to apply the set of settings above to C0, C1 & Unicode EOL.
Whereas `"Prevent control character (C0 code) typing into document"` has nothing to do with these settings, but it still belongs to "Non-Printing Characters" category.

It might be much clear for user if a new groupbox could be added:
image

If you have no better suggestion to make it more clear, I will consider the above solution after the release of v8.6.5.

> margins
>    make spacing between boxes and buttons more consistent

OK
But please don't modify the length of controls or labels which contain the string. They are always made longer then English strings need because the localization strings are usually longer. 

> so why not to add them in Style configurator to make it more convenient for user.
And if IIRC there were some questions about missing styles in community forums, usually after user updates Notepad++ to newer version.

OK I see your point. In this case, please do it in another PR.

donho:
OK, let's continue the GUI enhancement with this PR

PR #14894, opened 2024-03-22
faster URL scanning when one very long line is visible

### LINKED ISSUE #13916: Loading/Pasting included string causes program to hang
### Description of the Issue

Notepad++ hangs (Not Responding) when trying to load or pasting the included one line string.
[bug.txt](https://github.com/notepad-plus-plus/notepad-plus-plus/files/12135827/bug.txt)

Open the attached txt file or paste its contents

### Expected Behavior

txt file load/paste normally.

### Actual Behavior

txt file attempts to load, then program hangs. If pasting, the string displays and immediately the program hangs.

### Debug Information

Notepad++ v8.5.4   (64-bit)
Command Line : -noPlugin
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.3208
Current ANSI codepage : 1252
Plugins : none

It looks like Notepad++ is trying to parse the string as a url.
When the program has stopped responding, the thread stack cycles through the following:

0, ntdll.dll!RtlUnicodeToUTF8N+0x166
1, KernelBase.dll!WideCharToMultiByte+0x69c
2, wininet.dll!InternetCrackUrlW+0x177
3, notepad++.exe!SetLibraryProperty+0xab32c
n, ...

0, ntdll.dll!RtlUnicodeToUTF8N+0x180
1, KernelBase.dll!WideCharToMultiByte+0x69c
2, wininet.dll!InternetCrackUrlW+0x177
3, notepad++.exe!SetLibraryProperty+0xab32c
n, ...

0, wininet.dll!InternetCrackUrlW+0xbe
1, notepad++.exe!SetLibraryProperty+0xab32c
n, ...

0, notepad++.exe!SetLibraryProperty+0xab044
1, notepad++.exe!SetLibraryProperty+0xab2e1
n, ...

0, notepad++.exe!SetLibraryProperty+0xab13e
1, notepad++.exe!SetLibraryProperty+0xab2e1
n, ...

molsonkiko:
Fix #13916

## Introduction

Currently [`Notepad_plus::addHotSpot`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/cccd99b791c98ef4a9a729e31da45acb7f9bcdfd/PowerEditor/src/Notepad_plus.cpp#L3401) uses [`ScintillaEditView::getVisibleStartAndEndPosition`](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/cccd99b791c98ef4a9a729e31da45acb7f9bcdfd/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp#L2565) to get the start and end positions to scan for URLs, which sets startPos to the start of the first line in view and endPos to the end of the last line in view.

This means that when a very long line is in view, URL scanning can take an extremely long time, and in some unusual cases (see #13916, original message) can lead to loading or even scrolling a document to take minutes.

This commit fixes this issue by ensuring that, *if only one line is in view*, startPos and endPos are clamped as follows:
- startPos is set to the first character visible MINUS 32767 (see below for why this number is chosen) or the first character of the line, whichever is greater.
- endPos is set to the last character visible PLUS 32767 or the last character of the line, whichever is less.

*If multiple lines are in view*, startPos and endPos are set using the same rules as before.

This means that __a multi-line document similar to the one that originally caused this issue would have the same awful performance as before.__ See this comment by Coises (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13916#issuecomment-2010952251) for a proposal to change the URL-parsing algorithm in a way that would improve worst-case performance at the cost of introducing a complicated regular expression and noticeably reducing average-case performance. I do not think the proposed algorithmic change is worth the performance hit, but I can understand why others would disagree.

__In addition to eliminating insane latency in the most extreme cases, this commit also noticeably improves the vertical scroll rate (and sometimes improves load speed) on *all* word-wrapped documents when only one very long (>=2MB) line is in view.__

Because of the above-described change, horizontal scrolling now triggers URL scanning *if there is one line in view* (since it is still unnecessary when multiple lines are in view). Horizontal scrolling is still about as fast as it was before whether there was one line in view or multiple.

### An unrelated quality-of-life improvement

This commit turns off URL scanning while Notepad++ is shutting down, to help reduce shutdown time.

## A very slight disadvantage of this change 

Because of the changes in this commit, when only a single line is in view, *if an extremely long URL (longer than approximately 33 thousand characters) is only partially in view, it will not be highlighted.*

However, double-clicking a URL longer than 32767 characters cannot cause the URL to be opened in a web browser due to the behavior of ShellExecuteW (see https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry for explanation). *This is the way Notepad++ worked before this commit.* Note that the `Copy link` right-click context menu option *does currently work* for such links. That is the only feature that this commit would break.

Since double-clicking URLs that long already failed to open the page in a browser before this commit,
and since URLs that long are exceedingly rare,
and this commit only breaks highlighting of such URLs when one line is in view,
it seems obvious to me that this is an acceptable sacrifice for a potentially large performance improvement in much more common situations.

## How to test this commit
1. download the test file from the original message of #13916
2. open the test file in Notepad++. Verify that it takes only a short time to open (not minutes).
3. Turn ON word wrap. Try scrolling up and down. Verify that scrolling does not cause long lag.
4. Turn OFF word wrap. Try scrolling side to side. Verify that the scrolling does not cause long lag.
5. try replacing all instances of ":" with " https://".
6. Verify that URLs are highlighted as you scroll through the document. Once again, lag should not be too bad.
7. Turn ON word wrap. Verify that URLs are highlighted as you scroll through the document.
8. Save the test document with the URLs.
9. Close and reopen Notepad++. Verify that load speed is acceptable.
10. Repeat steps 3, 4, and 7.
11. Test URL scanning for a document with multiple lines. Make sure it works as normal.
12. Try creating a document with multiple lines that each have many characters (say, 100 lines with 200 thousand characters each).
13. Verify that horizontal scrolling is not too slow when word wrap is turned *off* and only one line is in view.
14. Verify that vertical scrolling is not too slow when word wrap is turned *on* and only one line is in view.
15. Download [`super long url.txt`](https://github.com/notepad-plus-plus/notepad-plus-plus/files/14716663/super.long.url.txt), which contains a very long URL.
16. Copy the text of `super long url.txt`, and paste it at the end of the file. Do not add a trailing EOL.
17. Turn word wrap off.
18. Place the caret between the two very long URLs.
19. Verify that both URLs can be double-clicked such that the URL is opened on your web browser.
20. Repeat step 19 with word wrap turned on.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3400,6 +3400,8 @@ bool isUrl(TCHAR * text, int textLen, int start, int* segmentLen)
 
 void Notepad_plus::addHotSpot(ScintillaEditView* view)
 {
+	if (_isAttemptingCloseOnQuit)
+		return; // don't recalculate URLs when shutting down
 	ScintillaEditView* pView = view ? view : _pEditView;
 	Buffer* currentBuf = pView->getCurrentBuffer();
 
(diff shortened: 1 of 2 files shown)


molsonkiko:
I currently believe that this PR could benefit from some changes [as I describe in this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13916#issuecomment-2015966588).

However, I'm not going to make any further changes without some feedback.

molsonkiko:
I have tested [Coises' PR](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14900) and I believe that it should be favored over this one. Although that PR does not deliver the performance improvements for scrolling and loading for a word-wrapped single-line document, it does not have any of the associated drawbacks, and it provides a much more robust solution to #13916.

This PR is of course not mutually exclusive with #14900, but don't accept this PR if all you care about is closing #13916.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3400,6 +3400,8 @@ bool isUrl(TCHAR * text, int textLen, int start, int* segmentLen)
 
 void Notepad_plus::addHotSpot(ScintillaEditView* view)
 {
+	if (_isAttemptingCloseOnQuit)
+		return; // don't recalculate URLs when shutting down
```
donho, line 3404: It's good to keep! 👍

donho (changes requested):
After trying #13916 with or without the PR, I see no difference.
Please remove the other code but keep 
```cpp
if (_isAttemptingCloseOnQuit)
		return; 
```

PowerEditor/src/Notepad_plus.cpp
molsonkiko: I recommend you use [this PR](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14931) instead and close this one, because that PR better reflects the fact that it doesn't relate to the originally linked issue #13916, and it's just a quality-of-life improvement.

Issue #14885, opened 2024-03-20
Search-results output has unconventional subfolder ordering


alankilborn:
### Description of the Issue
When doing a _Find in Files_ on a folder tree, it is conventional to show results closest to the top/root of the tree _first_ in the output; Notepad++ currently displays the tree inverted, i.e., closest-to-top _last_.

### Steps to Reproduce (STR) the Issue

1. Extract the attached `dir0.zip` to a temporary folder on your disk (I chose to do it to `C:\JUNK\` such that after the extraction I had `C:\JUNK\dir0\...`) [dir0.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/14666453/dir0.zip)
2. Invoke _Find in Files_ via Shift+Ctrl+f and set up the following search (changing the _Directory_ as to your specific temporary folder extraction of the zip file)
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/77065706/1a3100a1-18ac-4ac1-a347-31cb2e9a1e6b)
3. Press _Find All_
4. Right-click in the _Search results_ window and choose _Fold All_
5. Click the `+` in the box to the left of `Search "t"`...

### Actual Behavior
Bottom-up output:

![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/77065706/5831f1d5-da8c-4250-9e36-b8570645e3f7)

### Expected Behavior
Top-down output:
```txt
C:\JUNK\dir0\t.txt
C:\JUNK\dir0\dir1\t.txt
C:\JUNK\dir0\dir1\dir2\t.txt
C:\JUNK\dir0\dir1\dir2\dir3\t.txt
C:\JUNK\dir0\dir1\dir2\dir3\dir4\t.txt
C:\JUNK\dir0\dir1\dir2\dir3\dir4\dir5\t.txt
C:\JUNK\dir0\dir1\dir2\dir3\dir4\dir5\dir6\t.txt
```

---

### **EDIT** to add:

It appears that the output might be sorted (ascending) by the full file pathname.

Repeat the STR with after extracting this alternate zip: 
[dir0_a.zip](https://github.com/notepad-plus-plus/notepad-plus-plus/files/14694236/dir0_a.zip)

to obtain:

![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/77065706/070c1cc7-4cfd-4bec-bd3f-ed4637ba8a36)

[...]

PR #14899, opened 2024-03-23
BFS traversal for Find in Files

alexprabhat99:
This PR attempts to order subfolders on the basis of their path depth in Find in Files search results output as is expected in issue: #14885


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14885#issuecomment-2016558241

PR #14900, opened 2024-03-23
Prevent freeze due to backtracking when identifying clickable links in certain files

### LINKED ISSUE #13916: Loading/Pasting included string causes program to hang

Coises:
addHotSpot can become very slow when the screen contains certain sequences of characters that look like URLs but are not valid, due to a form of backtracking. This change eliminates the possibility of backtracking.

Resolves #13916.

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3367,8 +3367,13 @@ bool isUrl(TCHAR * text, int textLen, int start, int* segmentLen)
 				*segmentLen = len;
 				return true;
 			}
+			else // to avoid potentially catastrophic backtracking, skip the entire text that looked like a URL
+			{
+				*segmentLen = len;
+				return false;
+			}
 		}
-		else len = 1;
+		len = 1;
 		int lMax = textLen - start;
 		while (isUrlSchemeStartChar(text[start+len]) && (len < lMax)) len++;
 		*segmentLen = len;
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3120,37 +3120,17 @@ bool isUrlQueryDelimiter(TCHAR const c)
 	return false;
 }
 
-bool isUrlSchemeSupported(INTERNET_SCHEME s, TCHAR *url)
+bool isUrlSchemeSupported(TCHAR *url, int remainingLength)
 {
-	switch (s)
-	{
-		case INTERNET_SCHEME_FTP:
-		case INTERNET_SCHEME_HTTP:
-		case INTERNET_SCHEME_HTTPS:
-		case INTERNET_SCHEME_MAILTO:
-		case INTERNET_SCHEME_FILE:
-			return true;
-
-		case INTERNET_SCHEME_PARTIAL:
-		case INTERNET_SCHEME_UNKNOWN:
-		case INTERNET_SCHEME_DEFAULT:
-		case INTERNET_SCHEME_GOPHER:
-		case INTERNET_SCHEME_NEWS:
-		case INTERNET_SCHEME_SOCKS:
-		case INTERNET_SCHEME_JAVASCRIPT:
-		case INTERNET_SCHEME_VBSCRIPT:
-		case INTERNET_SCHEME_RES:
-		default:
-			break;
-	}
-	generic_string const mySchemes = (NppParameters::getInstance()).getNppGUI()._uriSchemes + TEXT(" ");
+	generic_string const mySchemes = L"ftp:// http:// https:// mailto: file:// "
+		                           + (NppParameters::getInstance()).getNppGUI()._uriSchemes + L" ";
 	TCHAR *p = (TCHAR *)mySchemes.c_str();
 	while (*p)
 	{
 		int i = 0;
 		while (p [i] && (p [i] != ' ')) i++;
 		if (i == 0) return false;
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3380,15 +3360,15 @@ bool isUrl(TCHAR * text, int textLen, int start, int* segmentLen)
...
+			bool r  = InternetCrackUrl(& text [start], len, 0, & url);
 			if (r)
 			{
 				while (removeUnwantedTrailingCharFromUrl (& text [start], & len));
 				*segmentLen = len;
 				return true;
 			}
 		}
-		len = 1;
+		else len = 1;
```
donho, line 3371: Hmm... this line has changed the semantic of this function.
In your PR, if `r` is FALSE, then `len` will keep its value, and then finish the function with the evaluated value by returning false.
Is it your intention? 
If it is, could you provide brief explanation, and make `else len = 1;` as
```cpp
else
{
    len = 1;
}
```
so the intention will not be unnoticed.

donho (changes requested):
Please confirm the intention, provide explanation and make the modification.

PowerEditor/src/Notepad_plus.cpp
Coises: It is intended — it is, in fact, the entire point of the change.

Issue #13916 happened because URLs can contain other URLs, and the previous process for finding URLs could wind up scanning the same text within a very long line very many times if something that "looked like" a URL but wasn’t, contained many other strings that also “looked like” URLs but were not.

This PR does two things.

First, it tightens the requirements for “looks like a URL” by checking the scheme earlier in the process. That is necessary to keep the next step from skipping valid URLs in reasonable contexts.

Second, once the beginning of a potential URL passes the tighter initial scanning and the end of the URL is found, we “commit” to that portion of the line. If the potential URL fails InternetCrackUrl validation, we restart scanning from the end of of the string that looked like a URL but wasn’t, rather than from just after the scheme.

In rare cases, this can cause a valid URL that “looks like” part of an invalid URL to be missed; for example, in the text ```ftp://this:4k,ftp://stop``` the existing logic underlines ```ftp://stop```; but with this PR nothing is underlined. I believe that is the unavoidable cost of preventing the “backtracking” loop observed in #13916.

In attempting to explain this, I have found a subtle error. The fall-through to the `while` loop that checks `isUrlSchemeStartChar` is usually unnecessary; that loop avoids restarting the search for a scheme in the middle of a longer string of alphabetic characters. The only time the end of a (potential) URL falls just before an alphabetic character is when the URL was ended by a delimited query string. In that situation — e.g., `ftp://this:4k?that="12"ftp://stop` — the second ftp *should* be recognized as the start of a URL (since it was not part of the invalid URL already scanned), but my change causes it to be skipped.

I will re-examine and retest; I think the correction, which will also be clearer, will be to add an `else` clause to the `if (r)` which will return false and the correct segment length — thus avoiding the fall-through — and remove the `else` I added before `len = 1;`. I’ll add a commit to this branch when I’m sure.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3380,13 +3360,18 @@ bool isUrl(TCHAR * text, int textLen, int start, int* segmentLen)
...
 			{
 				while (removeUnwantedTrailingCharFromUrl (& text [start], & len));
 				*segmentLen = len;
 				return true;
 			}
+			else // to avoid potentially catastrophic backtracking, skip the entire text that looked like a URL
+			{
+				*segmentLen = len;
+				return false;
+			}
```
donho, line 3374: It indeed makes more sense. 👍

donho (approved):

PR #14901, opened 2024-03-23
URL: exclude closing single quote in href attribute value

alexprabhat99:
Fixes: #14323


MarkusBodensee:
Looks like this PR just trims ' and " from the end of a url which is not the expectation if a url ends with such a char. Would like to test it if possible.

The check should be more like: is url is enclosed in ' or ".

chcg:
See https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14408 about a rejected PR.

Coises:
I tested by making this change to a copy of the current master branch, and it does not appear to fix the test case from #14323. That makes sense, since the changed code is executed only when [isUrlTextChar](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6204f00e881f7fc9459aff7134c12badbe6794bd/PowerEditor/src/Notepad_plus.cpp#L3092) is false for the potential ending character, and the test case is about an apostrophe, for which isUrlTextChar is true.

Might this be better addressed in [removeUnwantedTrailingCharFromUrl](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6204f00e881f7fc9459aff7134c12badbe6794bd/PowerEditor/src/Notepad_plus.cpp#L3296) instead of [scanToUrlEnd](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6204f00e881f7fc9459aff7134c12badbe6794bd/PowerEditor/src/Notepad_plus.cpp#L3198)?

The problem does not exist for URLs enclosed in double quotes, only when apostrophes are used. I think the idea is that if the character immediately preceding the beginning of the URL is an apostrophe, it is highly likely that the URL is enclosed in apostrophes and thus an apostrophe at the end is almost certainly a delimiter, not a part of the URL. That’s tricky to add to scanToUrlEnd because that routine is quite complicated. It looks to me, though, to be consistent with the sort of logic found in removeUnwantedTrailingCharFromUrl.

PR #14947, opened 2024-04-06
Use dpi scaled fonts for some controls

ozone10:
-  prepare static dialog for hiDPI


ozone10:
@donho 
should I create some generic issues for DPI pull requests?
HiDPI support will require a lot of work and I don't want to create large PR.

donho:
@ozone10
> should I create some generic issues for DPI pull requests?

One (or several) main issue(s) is (are) enough I guess (one or several, you judge it, it depends on the steps of the progress I guess).

> HiDPI support will require a lot of work and I don't want to create large PR.

Yes, that's the way to go.

ozone10:
@donho 
Ok I will create one main DPI issue to track what needs to be implemented.

Issue #14354, opened 2023-11-14
Dropped file may be opened in the wrong view


Yaron10:
**STR:**
Open two files.
Move one to the other view (-> SUB_VIEW is active).
Execute a search opening the Search Results window.
Drag & drop a third file _into the Search Results window_.

**Result:**
The file is opened in MAIN_VIEW.

```
Notepad++ v8.5.8   (64-bit)
Build time : Nov 14 2023 - 03:38:15
Path : ---notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : ON
Cloud Config : OFF
OS Name : Windows 10 Pro (64-bit)
OS Version : 20H2
OS Build : 19042.985
Current ANSI codepage : 1255
Plugins : 
    ComparePlus (1.1)
    HTMLTag (1.4.1)
    MenuIcons (2.0.5)
    PythonScript (2)
    _CustomizeToolbar (4.2)
```

alankilborn:
When I try dragging and dropping into anywhere in Notepad++ that is not view 1 nor view 2, the result is that the file opens in view 1.  I don't know that the expectation that a dropped file should open in the active view (when not dropped specifically there) is bug-worthy.  Maybe more of a feature-request. :-)

An argument could be made that if the drop target isn't specifically view 1 or view 2 that nothing should happen (i.e., the mouse cursor should not change, and no file should open if dropped).  I personally don't like that because often I have most of the N++ window covered with other stuff and I can just see a tiny corner -- enough of a file drop target for me to get my file open and N++ put on top in the Z order.

Yaron10:
> When I try dragging and dropping into anywhere in Notepad++ that is not view 1 nor view 2, the result is that the file opens in view 1.

The Search Results window was just an example.

> I don't know that the expectation that a dropped file should open in the active view (when not dropped specifically there) is bug-worthy.

Allow me to disagree.

> An argument could be made that if the drop target isn't specifically view 1 or view 2 that nothing should happen
...
I personally don't like that

I wouldn't like that either.

Yaron10:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/05f339b0cfe11095dc5590241bc8ff8cc4aec835/PowerEditor/src/Notepad_plus.cpp#L4313

Replace with:
```cpp
if ((hWin == _mainEditView.getHSelf()) || (hWin == _mainDocTab.getHSelf()))
[...]

PR #14953, opened 2024-04-07
Fix wrong dropped file view

### LINKED ISSUE #14951: Enhancement: When in side-by-side mode, when dragging a file to NP++, open file on side file dropped
### Description of the Issue
With NP++ displaying in side-by-side mode, open dragged file on side where dropped

1. Activate side-by-side display
2. Have a file on both sides
3. Drag a file from file explorer to side not most recently active

### Expected Behavior
Dragged file opens on side it was dropped

### Actual Behavior
The file opens on the side that was most recently active

### Debug Information
Notepad++ v8.6.4   (64-bit)
Build time : Feb 20 2024 - 00:12:59
Command Line :
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Version : 22H2
OS Build : 19045.4170
Current ANSI codepage : 1252
Plugins :
    DSpellCheck (1.5)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)

xomx:
Fix #14951 .

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -4281,7 +4281,7 @@ void Notepad_plus::dropFiles(HDROP hdrop)
 		HWND hWin = ::ChildWindowFromPointEx(_pPublicInterface->getHSelf(), p, CWP_SKIPINVISIBLE);
 		if (!hWin) return;
 
-		if ((_subEditView.getHSelf() == hWin) || (_subDocTab.getHSelf() == hWin) || currentView() == SUB_VIEW)
+		if ((_subEditView.getHSelf() == hWin) || (_subDocTab.getHSelf() == hWin))
 			switchEditViewTo(SUB_VIEW);
 		else
 			switchEditViewTo(MAIN_VIEW);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -4276,15 +4276,17 @@ void Notepad_plus::dropFiles(HDROP hdrop)
 	if (hdrop)
 	{
 		// Determinate in which view the file(s) is (are) dropped
-		POINT p;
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -4281,7 +4281,7 @@ void Notepad_plus::dropFiles(HDROP hdrop)
 		HWND hWin = ::ChildWindowFromPointEx(_pPublicInterface->getHSelf(), p, CWP_SKIPINVISIBLE);
 		if (!hWin) return;
 
-		if ((_subEditView.getHSelf() == hWin) || (_subDocTab.getHSelf() == hWin) || currentView() == SUB_VIEW)
+		if ((_subEditView.getHSelf() == hWin) || (_subDocTab.getHSelf() == hWin))
```
donho, line 4284: It kind of revert https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5f13c847191b38313d456fd69f575c5d84137ea0, which is supposed to fix #14354

donho (changes requested):
Could you find another solution to fix both #14354 & #14951?

PowerEditor/src/Notepad_plus.cpp
xomx: @donho 
You are right, my previous fix was not universal.

xomx:
@donho 
> Could you find another solution to fix both https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14354 & https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14951?

Ok, I have fixed both.
But kudos here goes to @Yaron10, who suggested [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14354#issuecomment-1811637865) the correct solution before, but you chose a different but incorrect solution.

donho (approved):

PR #14955, opened 2024-04-07
Enhance source code

donho:


PowerEditor/WinControls - Shortcut.lnk
alankilborn, line 1: A shortcut file...in the source code??

ozone10:
@donho 
maybe include this small modification in this PR:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/8d3d8aa6429a7408ca47c2fc15e0b8729992ef69/PowerEditor/src/WinControls/Grid/BabyGrid.cpp#L677

could be deleted since it's not used anyway.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
```diff
@@ -1094,8 +1076,8 @@ void DockingCont::drawTabItem(DRAWITEMSTRUCT *pDrawItemStruct)
 		if ((hImageList != NULL) && (iPosImage >= 0))
 		{
 			// Get height of image so we
-			IMAGEINFO	info		= {};
-			RECT &		imageRect	= info.rcImage;
+			IMAGEINFO info = {};
```
ozone10, line 1079: for consistency I would use `IMAGEINFO info{};`, same for `TCITEM` variables below

casey-chesshir:
your build is failing because this says cons instead of const 
image

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e6558a36188532bbf12f5aee985a96b27c9a23ad

PR #14958, opened 2024-04-08
Improve HiDPI compatibility for tab control labels

### LINKED ISSUE #11012: with Dark mode, in Search Dialog, the tab labels are cut (windows 11)
### Description of the Issue
with Dark mode, in Search Dialog, the tab labels are cut (i.e. "replace" and "find in files" )
OS: windows 11
resolution: 1920 X 1080, zoom 150%
changing to 1280 X 768 zoom 100%, the result is the same
with a different computer, with win 10, the labels are more visible (just a little bit cut)

1.  set dark mode
2.  open search dialog
3.  see the labels "replace" and "find in files" cut

### Debug Information
Notepad++ v8.2   (64-bit)
Build time : Dec 30 2021 - 03:40:42
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Home (64-bit)  --> Actually windows 11 (N++ shows wrong information)
OS Version : 2009
OS Build : 22000.376
Current ANSI codepage : 1252
Plugins : ComparePlugin.dll mimeTools.dll NppConverter.dll NppExport.dll NppXmlTreeviewPlugin.dll 

image

ozone10:
-  fix typo

may fix #11012

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -593,10 +593,10 @@ namespace NppDarkMode
 
 	static TreeViewStyle g_treeViewStyle = TreeViewStyle::classic;
 	static COLORREF g_treeViewBg = NppParameters::getInstance().getCurrentDefaultBgColor();
-	static double g_lighnessTreeView = 50.0;
+	static double g_lightnessTreeView = 50.0;
 
 	// adapted from https://stackoverflow.com/a/56678483
-	double calculatePerceivedLighness(COLORREF c)
+	double calculatePerceivedLightness(COLORREF c)
 	{
 		auto linearValue = [](double colorChannel) -> double
 		{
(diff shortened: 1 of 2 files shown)


PowerEditor/src/NppDarkMode.cpp
```diff
@@ -1582,7 +1584,7 @@ namespace NppDarkMode
 
 		case WM_NCDESTROY:
 		{
-			RemoveWindowSubclass(hWnd, TabSubclass, g_tabSubclassID);
+			::RemoveWindowSubclass(hWnd, TabSubclass, uIdSubclass);
```
donho, line 1587: Are you sure about it?
The current code in master we have symetric calls `SetWindowSubclass(hwnd, TabSubclass, g_tabSubclassID, 0);` & `RemoveWindowSubclass(hWnd, TabSubclass, g_tabSubclassID);`.
In this PR  I wonder if `g_tabSubclassID` is still useful.

ozone10: Yes.
`SetWindowSubclass` will send `g_tabSubclassID` via `TabSubclass` (https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-setwindowsubclass)

IIRC there was error with some subclass which had wrong subclass id variable, probably due to copy&paste. Fortunately every subclass ids Notepad++ use have same value.

So this is mainly to avoid future issue when copy&paste or if several subclassing is required.

Issue #14959, opened 2024-04-08
[DPI] Elements required works to properly support hiDPI


ozone10:
This issue is for tracking issues with hiDPI environment and works on it.

## Main window
 - [ ] Tab control
 - [ ] Toolbar
 - [ ] Editor
 - [ ] Status bar

## Dialogs
- [ ] FindInFinder
- [ ] About
- [ ] DebugInfo
- [ ] Run 
- [ ] HashFromFiles
- [ ] GoToLine
- [ ] ColumnEditor 
- [ ] Style Configurator 
- [ ] Preference
- [ ] FindCharsInRange 
- [ ] Rename Current Tab
- [ ] Save all
- [ ] Shortcut
- [ ] Run Macro Multiple Times
- [ ] UDL styler

## More complex
- [ ] Find and Replace
- [ ] PluginsAdmin 
- [ ] Windows 
- [ ] Shortcut Mapper
- [ ] Progress
- [ ] Color Popup
- [ ] Color dialog
- [ ] UDL

## Panels
- [ ] Title
- [ ] Tab control
- [ ] Main field
- [ ] Plugin support

donho:
@ozone10 

Just found 2 dialogs (Preferences & Find Replace) partially working & 1 (Shortcut mapper) totally not working:
  
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/90293/425d4a06-f663-4cad-abe1-e287aa4a8059)

ozone10:
@donho 
I will be away for 2 weeks.

For preference add `DPIManagerV2::setDpiWithParent(_hSelf);` before line
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/018d4f9a1f01a4fc17ebc1d0d8e372db1cf5539d/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp#L1541

Shortcut mapper, Find&Replace and Windows dialogs are sizeable which require more works.

btw
While working on toolbar I noticed some issue with custom icons for toolbar
Could you check it:
1. Have custom icons for toolbar enabled.
2. Open preference and change between large and small icons several times.

donho:
> Just found 2 dialogs (Preferences & Find Replace) partially working & 1 (Shortcut mapper) totally not working:

@ozone10 
Sorry, I wasn't clear. What I meant is:
[...]

PR #14980, opened 2024-04-11
Add dpi messages

ozone10:
- process WM_DPICHANGED_AFTERPARENT message
- remove unused function
- tweak 2 switches, which use enums to avoid warnings

ref #14959


ozone10:
@donho 
this should be last "preparation" PR for now.
After this I will try to implement tasks from #14959

PR #14989, opened 2024-04-13
Add dpi support to dialogs:

ozone10:
- About
- DebugInfo
- Save All
- Close All

ref #14959

### DIFF
--- a/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
+++ b/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
@@ -85,31 +85,41 @@ intptr_t CALLBACK AboutDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM lPar
 		case NPPM_INTERNAL_REFRESHDARKMODE:
 		{
 			NppDarkMode::autoThemeChildControls(_hSelf);
+			if (_hIcon != nullptr)
+			{
+				::DestroyIcon(_hIcon);
+				_hIcon = nullptr;
+			}
 			return TRUE;
 		}
 
-		case WM_DRAWITEM :
+		case WM_DPICHANGED:
 		{
-			DPIManager& dpiManager = NppParameters::getInstance()._dpiManager;
-			int iconSideSize = 80;
-			int w = dpiManager.scaleX(iconSideSize);
-			int h = dpiManager.scaleY(iconSideSize);
+			DPIManagerV2::setDpiWP(wParam);
+			destroy();
+			//setPositionDpi(lParam);
+			getClientRect(_rc);
 
-			HICON hIcon = nullptr;
-			if (NppDarkMode::isEnabled())
-				hIcon = (HICON)::LoadImage(_hInst, MAKEINTRESOURCE(IDI_CHAMELEON_DM), IMAGE_ICON, w, h, LR_DEFAULTSIZE);
-			else
-				hIcon = (HICON)::LoadImage(_hInst, MAKEINTRESOURCE(IDI_CHAMELEON), IMAGE_ICON, w, h, LR_DEFAULTSIZE);
+			return TRUE;
+		}
+
+		case WM_DRAWITEM:
+		{
+			const int iconSize = DPIManagerV2::scale(80);
+			if (_hIcon == nullptr)
+			{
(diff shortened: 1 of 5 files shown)


PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -85,31 +85,41 @@ intptr_t CALLBACK AboutDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM lPar
...
 
-		case WM_DRAWITEM :
+		case WM_DPICHANGED:
 		{
-			DPIManager& dpiManager = NppParameters::getInstance()._dpiManager;
-			int iconSideSize = 80;
-			int w = dpiManager.scaleX(iconSideSize);
-			int h = dpiManager.scaleY(iconSideSize);
+			DPIManagerV2::setDpiWP(wParam);
+			destroy();
```
donho, line 99: I suppose `WM_DPICHANGED` message is sent before `WM_DRAWITEM`, so I see your logic to call the destroy method. However, in destroy() method, not only the `_hIcon` is destroyed, but also `_pageLink`.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -85,31 +85,41 @@ intptr_t CALLBACK AboutDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM lPar
...
+			//setPositionDpi(lParam);
```
donho, line 100: Is the comment here on purpose?

donho (changes requested):
Please reply some points which is unclear to me.
Also, please add the necessary information for testing the PR.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -409,7 +430,8 @@ void DebugInfoDlg::doDialog()
 	// For example, the command line parameters may have changed since this dialog was last opened during this session.
 	refreshDebugInfo();
 
-	// Adjust the position of AboutBox
+	// Adjust the position of DebugBox
+	goToCenter(SWP_HIDEWINDOW | SWP_NOSIZE | SWP_NOACTIVATE);
```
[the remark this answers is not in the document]
ozone10: To trigger `WM_DPICHANGED`, 
Let's consider a scenario: A user opens the debug dialog on screen 1 with a scale of 100%. 
Next, they move the main Notepad++ window to the second screen, which has a scale of 150%. 
The user then clicks on the menu bar, selects "? -> Debug Info," causing the debug dialog to move to the second screen.
This action triggers the `WM_DPICHANGED` event, resulting in the debug dialog resizing.
Unfortunately, the `goToCenter` function was called before `WM_DPICHANGED`, so it uses a rectangle with the old dimensions to calculate center position, causing the debug dialog not to be centered.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
```diff
@@ -364,6 +376,15 @@ intptr_t CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM
 			return TRUE;
 		}
 
+		case WM_DPICHANGED:
```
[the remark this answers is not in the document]
ozone10: Change this line
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/161bb41425853168dc4dc85ea02b88e4267685d2/PowerEditor/src/dpiAware.manifest#L8

to 
```
<dpiAwareness>PerMonitorV2, system, unaware</dpiAwareness>
```

Later I will create testing PR with this change to make it easier to test. PR will include commits from other dpi PRs.

PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
ozone10: `_pageLink` has old font size.

donho: My point is _hIcon is destroyed in WM_DPICHANGED, but it will be recreated in WM_DRAWITEM.
Whereas it's not the case for _pageLink. Or am I missing somethings?

donho: IMO, here we should use 
```cpp
if (_hIcon != nullptr)
{
	::DestroyIcon(_hIcon);
	_hIcon = nullptr;
}
```
instead of `destroy()`

PR #15007, opened 2024-04-16
Resize checkboxes/radio buttons according to text length

### LINKED ISSUE #15006: [BUG] Find window: Checkboxes working if clicked in blank space.

### Description of the Issue

In the find menu, 
The check boxes get checked/unchecked if we click in blank space which is inline with the option
notepad++-misclick

### Steps To Reproduce

1.  CTRL + F : to open Find window
2.  Click next to the option Backward direction(or any other option) in the blank space

### Current Behavior

Check boxes getting checked/unchecked if clicked in blank space.

### Expected Behavior

Check boxes should not be checked/unchecked when clicked in blank space next to it.
They should be functional only when clicked on the text or the checkbox.

### Debug Information

```shell
Notepad++ v8.6.5   (64-bit)
Command Line : D:\
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 11 Pro (64-bit
OS Version : 23H2
OS Build : 22631.3447
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2.0.2)
    Explorer (1.9.9)
    JsonTools (5)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    NPPJSONViewer (2.0.5)
    Python Indent (1.0.0.4)
    PythonScript (2)
    XMLTools (3.1.1.13)
```

### Anything else?

_No response_

ozone10:
fix #15006


xomx:
Tested, works ok.

ozone10:
still need works on multiline buttons

ozone10:
Ok should work now. I don't want to make it more complicated so multiline buttons are ignored. Only 3 in preference so IMHO it shouldn't be problem.

ozone10:
This will also remove width restriction. Few localizations will require changing some long lines, which are currently "cut" due to width restriction, to avoid the button label going under other controls.

donho (changes requested):
It's nice this PR.
However:

image

The current implementation of master is not dynamic indeed, but prevent the above case.

For the PR being accepted, the criteria is following:
1. The RC files cannot be modified, so the maximal length of control can be retrieved.
2. Use the retrieved maximal length as threshold to resize the control for every language: in the same session users could switch among 3 - n languages (I personally do 3 translations in the same time).

Good luck.

ozone10:
@donho 
don't merge this PR yet, I am rewriting it now to avoid using `GWLP_USERDATA` for better compatibility for 3rd party apps

ozone10:
@donho 
sorry for several pushes, I am forgetting to save files.

But it should be ok for review now.

Issue #14975, opened 2024-04-11
Find dialog doesn't allow to select a different tab easily via the keyboard


xMRi:
### Steps to Reproduce the Issue
1.  Open the find dialog
2.  The first tab is active 
3. Try to access the Tab "Mark" just with the keyboard

### Expected Behavior
I want to use F6, or Shift+F6, Ctrl+PageUp, Ctrl+PageDown, or press Alt+Hotkey to activate another tab page in the Find Dialog 
This are standard hotkey in Windows!

Or Ctrl+Number to access the first, second, third tab.

### Actual Behavior
The onyl way is to press Shift+Tab, move the focus to the tab, press right or left.
The focus turns back onto the first control in the tab.
I have to repeat Shift+Tab, Right.

### Comment
The keyboard interface of Notepad++ is horrible and!

sXNA:
@xMRi 
I just tried your standard hotkeys in some tabbed applications and not all of them did work (Windows 11 explorer, explorer properties window, notepad, terminal, ...). All hotkeys work only in browsers for me. 

So these "standard" hotkeys are not so standard even for MS.
And if keyboard interface of Notepad++ is horrible, you can always change shortcuts in Shortcut mapper or do some PR to improve it.

Except for Find in Project tab I can easily switch between them using keyboard.
Ctrl+F : Find
Ctrl+H : Replace
Ctrl+Shift+F : Find in Files
Ctrl+M : Mark

xMRi:
Standard Windows Property Sheets:
Open Explorer (Windows+E)
Open Properties of A file (Alt+Enter

Paging in Property Sheet: Ctrl+PgUp/dn

xomx:
@alankilborn 

I do remember that Ctrl+Tab / Ctrl+Shift+Tab request: #7932
There was even a good PR for that, but unfortunately it was rejected: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/10170

I am somewhat used to that shortcuts from another apps, so having this also in N++ would be a nice feature for me.

xMRi:
If I look into the Wine code I find Ctrl+Tab, Shift+Ctrl+Tab, Ctrl+PgUp, Ctrl+PgDn
https://github.com/wine-mirror/wine/blob/master/dlls/comctl32/propsheet.c

I just debugged into the proerty sheet message handler and it seams that there is the same code in it.

ozone10:
@xMRi 
[...]

PR #15015, opened 2024-04-19
Allow Ctrl-TAB to switch tabs in PluginAdmin and UDL dialogs

### LINKED ISSUE #7932: Ctrl+Tab doesn't switch tabs in dialogs
### Description of the Issue

In dialogs like Find/Replace, Shortcut mapper, Plugins admin, User Defined Language... when I press <kbd>Ctrl</kbd>+<kbd>Tab</kbd> then the focus just moves to the next tab stop instead of moving to the next tab in the dialog

1. Open any dialogs with tabs
2. Press <kbd>Ctrl</kbd>+<kbd>Tab</kbd> or <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Tab</kbd>

### Expected Behavior

Tabs should be switched instead of changing focus

ozone10:
ref #14975

partially fix #7932


ozone10:
I won't be doing shortcut mapper soon, I will be busy for at least 2 months.

PR #15016, opened 2024-04-19
Add dpi support to status bar

ozone10:
ref #14959


ozone10:
@donho 
build is failing because I was maybe creating many PRs too fast.

In my run it's passing https://github.com/ozone10/notepad-plus-plus/actions/runs/8755410563

PR #15023, opened 2024-04-21
Add dpi support to plugin admin dialog

ozone10:
ref #14959


PowerEditor/src/WinControls/PluginsAdmin/pluginsAdmin.cpp
```diff
@@ -183,6 +182,9 @@ void PluginsAdminDlg::create(int dialogID, bool isRTL, bool msgDestParent)
 		ListView_SetBkColor(hList, bgColor);
 		ListView_SetTextBkColor(hList, bgColor);
 		ListView_SetTextColor(hList, fgColor);
+		const auto style = ::GetWindowLongPtr(hList, GWL_STYLE);
+		::SetWindowLongPtr(hList, GWL_STYLE, style | WS_TABSTOP);
```
donho, line 186: 👍

PR #15024, opened 2024-04-21
Add dpi support to UDL dialogs

ozone10:
- StringDlg (Rename Current Tab, ...)
- UDL Styler
- UDL in undock state

- UDL allow tab switching between main dialog and subdialogs
- make controls spacing and size consistent

ref #14959


donho:
@ozone10 
UDL is fine, but there are some dialogs not working:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14959#issuecomment-2068108871

PR #15029, opened 2024-04-22
Add dpi support to Style Configurator dialog

ozone10:
- reorder controls for consistent tab switch
- make spacing and position more consistent
- use system aware or gdi scaling for Color dialog and Color popup

ref #14959


PowerEditor/src/WinControls/ColourPicker/ColourPopup.cpp
```diff
@@ -43,6 +45,17 @@ void ColourPopup::create(int dialogID)
 	display();
 }
 
+void ColourPopup::createForDpi(int dialogID)
+{
+	const DPI_AWARENESS_CONTEXT dpiContextSystem = NppDarkMode::isWindows10() ? DPI_AWARENESS_CONTEXT_UNAWARE_GDISCALED : DPI_AWARENESS_CONTEXT_SYSTEM_AWARE;
+	const auto dpiContext = DPIManagerV2::setThreadDpiAwarenessContext(dpiContextSystem);
+	create(dialogID);
```
donho, line 52: I don't like this wrapper function.
I'll see what I can do for it.

donho: @ozone10 
FYI:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/4ffa8fcdf93fdce5b63ac469e2834e545fb7ac2a

Here is my reason:
Using functions around the create() is OK - we know what we are doing. However the name "createForDpi" is confusing IMO. Hence the modification.

PR #15031, opened 2024-04-23
Fixes for the npp backup thread

### LINKED ISSUE #14906: possible crash when recursively entering the Notepad++ backup_mutex

xomx:
Fix #14906 .

This PR fixes:
- possible backup thread deadlock exception (re-entering the std::mutex from the same thread)
- possible thread message queue flooding by the multiple PostMessages when a backup IO is blocked somehow or takes a longer time than the backup period set
- unhandled backup worker thread closing at the very Notepad++ exit (e.g. in the nppLogNulContentCorruptionIssue.log you can sometimes see that the backup thread is still running & doing its job even in the very last stages of the Notepad++ app existence, which is IMO not completely safe...)


donho:
Currentely #14906 cannot be reproducible, so the PR is on standby.

donho:
@xomx 
Any news for the reproduction of crash?

It seems https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6133 has been fixed by https://github.com/notepad-plus-plus/notepad-plus-plus/commit/bbeaafac8bde911d3268f2d40b90329feb5c57f4

xomx:
@donho Vacation until Aug 5th. Then I will recheck this again.Dne 21. 7. 2024 13:01 napsal uživatel Don HO ***@***.***>:
@xomx
Any news for the reproduction of crash?
It seems #6133 has been fixed by bbeaafa

donho:
@xomx 
Enjoy your holidays!

donho:
@xomx 
No rush.
I just want to know if we can close the PR.

donho:
FYI: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15669#issuecomment-2397916755

xomx:
@donho 
I still think that this PR fixes the first two points mentioned in this PR 1st post better than the https://github.com/notepad-plus-plus/notepad-plus-plus/commit/bbeaafac8bde911d3268f2d40b90329feb5c57f4 (this PR will not block the main N++ thread in case that the backup worker thread is busy or does not respond at all). But this PR is not as simple as the mentioned PostMessage->SendMessage change, so it's clear to me that you probably won't be inclined to accept it.

But there is also the 3rd point

> * unhandled backup worker thread closing at the very Notepad++ exit (e.g. in the nppLogNulContentCorruptionIssue.log you can sometimes see that the backup thread is still running & doing its job even in the very last stages of the Notepad++ app existence, which is IMO not completely safe...)

for which we now have a solid proof https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15681#issuecomment-2403168094 why it is bad.

Changes needed to fix this backup-thread-crash-after-npp-winmain-exit are only few:
- adding new atomic exit flag
  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0478122f8af69fce07652d8ccbc189dc265980f1/PowerEditor/src/NppBigSwitch.cpp#L42
  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0478122f8af69fce07652d8ccbc189dc265980f1/PowerEditor/src/Notepad_plus.h#L51
- set this in an appropriate place
  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0478122f8af69fce07652d8ccbc189dc265980f1/PowerEditor/src/NppBigSwitch.cpp#L2610-L2613
- and use it in the backup worker thread loop
  https://github.com/notepad-plus-plus/notepad-plus-plus/blob/0478122f8af69fce07652d8ccbc189dc265980f1/PowerEditor/src/Notepad_plus.cpp#L8470

Do you agree with closing of this PR and creating new simpler one for the point No.3 only?

donho:
@xomx 
Does the fix of the point No.3 has been done in this PR entirely?

xomx:
@donho 
Yes, the snippets in my previous post above is from this PR branch.

donho:
@xomx 
> Do you agree with closing of this PR and creating new simpler one for the point No.3 only?

OK please.

xomx:
Closed, will be replaced by a simpler variant later.

xomx:
@donho 
For now, I will continue here with some questions before I create new backup-PR.

1)
While debugging I noticed one strange variable declaration. In fact I noticed that because I saw the `'\x1'`:

npp-boolean-insteadof_bool

I am guessing this `boolean` (boolean == unsigned char type) should be `bool` instead. Fortunately, it does not matter functionally in the code, but I would replace it (I found it in two places) with the standard `bool` type - are you in favor of that change?

2)
Until now, I mistakenly believed that the N++ only writes a backup when changes are detected in an open filebuffer/tab, when the "Trigger backup on modification in every xxx seconds" interval has elapsed. Now I see that backup with exactly the same data are written over and over again, even if nothing else has been changed from the last change detection.

STR:
- open one file (or use one unsaved new buffer/tab, it does not matter)
- modify that buffer/tab somehow
- because the buffer/tab is now marked as "dirty" and "modified", the same data will be written over and over again by the backup thread regardless of whether the user actually makes any further changes to it afterwards

Is this an intended way how the backup thread should operate?

donho:
@xomx 
> I am guessing this boolean (boolean == unsigned char type) should be bool instead. Fortunately, it does not matter functionally in the code, but I would replace it (I found it in two places) with the standard bool type - are you in favor of that change?

I must use it by mistake - please replace them.

> Is this an intended way how the backup thread should operate?

Yes. But if you've got a better idea & the implementation is not complicated, you can do a another PR for changing this logic.

Issue #2417, opened 2016-10-12
"İ" not the lowercase


gevv:
hi,

İ character is not lowercase
### Debug Information

Notepad++ v6.9.2
Build time : May 18 2016 - 00:34:05
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS : Windows XP
Plugins : mimeTools.dll NppConverter.dll NppExport.dll PluginManager.dll 

![untitled 1](https://cloud.githubusercontent.com/assets/5096676/19324351/d37c0ab8-90c9-11e6-8a40-4ce664c641c7.png)

![untitled 2](https://cloud.githubusercontent.com/assets/5096676/19324356/d9a80a54-90c9-11e6-93e6-43d67aaf402e.png)

SinghRajenM:
By any change are you using Turkish Keyboard? I could not reproduce this issue with english keyboard.
If yes, this issue is occuring with dotted (İ) and dotless (I) seems to work fine.

guy038:
Hello,  **gevv**,

You're really **not lucky** for that matter :-(((

Indeed, it happens that, among **all** the **Latin**, **Greek** or **Cyrillic** **`capital`** letters, of the default **Courrier New** font, which have an **lower-case equivalent** letter, the capital Azerbaijani Turkish letter **İ** ( **`\x{0130}`** ) is the **`unique`** letter **without** an **`lower-case`** equivalent, in Notepad++ !!

Indeed, in the **Unicode 9.0 Latin-Extended-A** code-chart, for the item **LATIN CAPITAL LETTER I WITH DOT ABOVE** of code point **`0130`** it's specified that the **lower-case** form of the **capital** letter **İ** is, simply, the normal lower-case **i** ( **`\x69`** )

Refer to the link, below :

http://www.unicode.org/charts/PDF/U0100.pdf

Below, I gather a **sorted** ( in the **A** .. to .. **Z** way ) list, of all **Latin**, **Greek** and **Cyrillic** letters, of the default **Courrier New** font, in their **both** forms ( **upper-case** and **lower-case** )

---

**Capital** Latin letters :

[...]

Issue #2321, opened 2016-09-22
uninstall should not be deleting configuration (by default)


bughit:
sometimes you need to uninstall and reinstall or upgrade, for example:

x64 version does not upgrade the x86 one, so you need to uninstall x86 then install x64

removing config could be an option in the uninstall

dail8859:
Closed by bb01b0f528482ff97b84685b82812d7684e1c32a

PR #15042, opened 2024-04-24
Update to scintilla 5.5.0 & Lexilla 5.3.2

chcg:
Sintilla Release 5.5.0 (https://www.scintilla.org/scintilla550.zip)

    Released 23 April 2024.
    Add elements for inactive additional selections SC_ELEMENT_SELECTION_INACTIVE_ADDITIONAL_TEXT and SC_ELEMENT_SELECTION_INACTIVE_ADDITIONAL_BACK. When not set these default to SC_ELEMENT_SELECTION_INACTIVE_TEXT and SC_ELEMENT_SELECTION_INACTIVE_BACK. Bug #2417.
    On Cocoa, avoid use of NSUserDefaults which will soon require justification when used in applications on the App Store.
    Fix Win32 IME crash in windowed mode. Bug #2433.
    Scale reverse arrow cursor for margins to match other cursors when user changes pointer size. Bug #2321.

Lexilla Release 5.3.2 (https://www.scintilla.org/lexilla532.zip)

    Released 23 April 2024.
    COBOL: Stop string literal continuing over line end. Issue #229.
    COBOL: Stop doc comment assigning different styles to \r and \n at line end. Issue #229.
    COBOL: Recognize keywords that start with 'V'. Issue #230.
    COBOL: Recognize comments after tag or that start with '/'. Issue #231.
    HTML: Implement substyles for tags, attributes, and identifiers SCE_H_TAG, SCE_H_ATTRIBUTE, SCE_HJ_WORD, SCE_HJA_WORD, SCE_HB_WORD, SCE_HP_WORD, SCE_HPHP_WORD.
    HTML: Implement context-sensitive attributes. "tag.attribute" matches "attribute" only inside "tag".
    HTML: Match standard handling of comments. Issue #232.
    Lua: Implement substyles for identifiers SCE_LUA_IDENTIFIER.
    Ruby: Allow non-ASCII here-doc delimiters. Issue #234.
    Ruby: Allow modifier if, unless, while and until after heredoc delimiter. Issue #236.
    Rust: Recognize raw identifiers. Issue #239, Pull request #240.


chcg:
@donho @ArkadiuszMichalski 

With this update all files are opened as language Go. Need to be investigated. Any hint appreciated.

chcg:
Just happens if file is opened via context menu and therefore WM_COPYDATA with COPYDATA_PARAMS is used at Notepad_plus::process()

ozone10:
Maybe related to similar issues with JSON5 https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13746 and https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13042

chcg:
Could be. I'm testing with a portable version 8.5.6 update with this compilation but there is also an installation of N++ 8.5.6 available.

donho:
> With this update all files are opened and autodetected as language Go but could be manually changed to the correct one. Need to be investigated. Any hint appreciated.

It can happen when the arguments are passed by `%ProgramFiles%\Notepad++\notepad++.exe` to (YOUR LAUNCHED) `whereever\notepad++.exe` and both notepad++.exe are not the same version.
Here is the detailed explanation:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13042#issuecomment-1416802964

Issue #15077, opened 2024-05-03
[BUG] Custom draw not working in TreeView controls inside plugins


oviradoi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

The subclass that is added by the Dark Mode code to will handle the NM_CUSTOMDRAW message sent by TreeView controls.

This subclass will return CDRF_DODEFAULT and this will block any plugins trying to handle this message themselves. 

The result will be that the customizations made by the plugins will not be applied.

This will only happen if the plugin is not using the DWS_USEOWNDARKMODE flag.

### Steps To Reproduce

1. Install the Explorer plugin in the plugin manager.
2. Open the Favorites panel and add a currently opened file to the Favorites list.

### Current Behavior

If the file is open, the entry in the Favorites tree should be underlined to indicate the file is open.

Because the subclass overrides the custom draw messages, the entry is not underlined.

### Expected Behavior

The entry in the Favorites panel should be underlined if the file is open in Notepad++

### Debug Information

```shell
Notepad++ v8.6.5   (64-bit)
Build time : Mar 29 2024 - 17:04:43
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 10 Pro (64-bit
OS Version : 22H2
OS Build : 19045.4291
Current ANSI codepage : 1252
Plugins : 
    Explorer (1.9.9)
    JSMinNPP (1.2205)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    NppXmlTreeviewPlugin (2)
    XMLTools (3.1.1.13)
```

donho:
@oviradoi 

I don't see any unusual stuff:
 
![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/90293/d28084e7-7851-4b7b-adad-d117a264cb38)

oviradoi:
@donho 
If a file is open in Notepad++, the item in the TreeView should be underlined.

The Explorer plugin has code that handles the NM_CUSTOMDRAW message and sets an underlined font if the file is open. But since Notepad++ does not call DefSubclassProc, this code is not called, and the underlined font is not set for the files that are open.

Here is how it should look:
[...]

PR #11767, opened 2022-06-06
Add support for plugin custom draw for listview subitems,

ozone10:
add support for 3state checkbox button

Enhancement to https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11711.
Based on information from https://community.notepad-plus-plus.org/topic/23080/impact-analysis-of-generic-darkmode-rendering-of-plugin-dock-panels-from-pr-11711.

I have not tested this locally as I am currently away and cannot compile it.

PR #15078, opened 2024-05-03
Allow plugins to handle Custom Draw in TreeView controls (bug #15077)

oviradoi:
This PR attempts to fix bug #15077 .

I have modified the code that handles the CustomDraw messages sent by TreeView controls to also call the base subclass and merge the result with the one from the dark mode code.

This will allow the plugins to also handle the custom draw messages and add customizations to the TreeView items.

I have tested this with the Explorer plugin and also several built-in panels (Folder as Workspace, Document List).

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -2707,7 +2707,7 @@ namespace NppDarkMode
 		return ::DefSubclassProc(hWnd, uMsg, wParam, lParam);
 	}
 
-	LRESULT darkTreeViewNotifyCustomDraw(LPARAM lParam)
+	LRESULT darkTreeViewNotifyCustomDraw(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, bool isPlugin)
 	{
 		auto lptvcd = reinterpret_cast<LPNMTVCUSTOMDRAW>(lParam);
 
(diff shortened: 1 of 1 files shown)


donho:
The eventual problem of this PR is, it could make the other plugins which don't want to deal with the dark mode broken.

@ozone10 
What do you think about this PR?

ozone10:
@donho 
We already do this for listview. IMHO it would be better to align the code with the listview’s implementation for consistency. However, I still find the current PR acceptable.

I would also include similar handling for toolbar in another PR, if requested.

oviradoi:
I don't think it would break any plugin. If a plugin does not subclass a TreeView, the call to DefSubclassProc will not have any effect and the previous behavior will be kept. If a plugin does subclass TreeView, these changes will allow the plugin to add its own customizations.

donho:
@oviradoi 
Could you show me what's wrong of my screenshot:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15077#issuecomment-2107956015

donho:
@ozone10 
> We already do this for listview.

Could you point it out please?

ozone10:
@donho 
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/11767

donho:
@ozone10 
Thank you!

PR #15089, opened 2024-05-05
Initial new dpi management support for Shortcut Mapper

ozone10:
ref #14959


donho:
@ozone10 
By "Initial new dpi management support for Shortcut Mapper" I guess it doesn't do the multi-display support, but it should scale
correctly in the mono screen.

However, I've got this when I scale my current display from 225% to 300% :
image

ozone10:
@donho 
changing dpi in single screen is same as dragging to second screen.
This PR only handle initial opening of shorcut mapper, mainly font handling.
To fully support DPI, `WM_DPICHANGED`, `WM_GETDPISCALEDSIZE` need to be handled and `WM_SIZE` rewritten.

PR #15090, opened 2024-05-05
Improve status bar resize handling

ozone10:
ref #14959


donho:
@ozone10 
By "Improve status bar resize handling", it improves the status improvement display or it's a code refactoring?

ozone10:
@donho 
technically it's more code refactoring. But it helps in some rare case to avoid visual glitch (status bar not repositioning immediately) when changing between full and less mode.

PR #15129, opened 2024-05-11
Fix regression of multi-edit cursors placed wrongly issue

### LINKED ISSUE #15126: [BUG] The caret position is unnatural when repeating column mode.

### Description of the Issue

[Multi Edit resists escape after typing · Issue #14649 · notepad-plus-plus/notepad-plus-plus · GitHub](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14649)
This issue has been resolved in v8.6.6, but
I would like to report that the behavior when starting column mode again after canceling column mode was different from v8.6.5 and earlier.

### Steps To Reproduce

1. Press Alt+Shift+Down two or more times to place the caret in three or more rows.
2. Press Esc to exit column mode.
3. Press Alt+Shift+Up, then the caret will be placed up to the first line selected in 1.

### Current Behavior

Column mode attempts to restore the last manipulated position.

### Expected Behavior

Column mode must operate anew, regardless of the previously set range.

### Debug Information

```shell
Notepad++ v8.6.6   (64-bit)
Build time : May  6 2024 - 21:59:00
Command Line : "C:\Program Files\Notepad++\change.log" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : OFF
OS Version : 22H2
OS Build : 19045.4291
Current ANSI codepage : 932
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

donho:
Fix #15126

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -650,6 +650,7 @@ LRESULT ScintillaEditView::scintillaNew_Proc(HWND hwnd, UINT Message, WPARAM wPa
 						int selection = static_cast<int>(execute(SCI_GETMAINSELECTION, 0, 0));
 						int caret = static_cast<int>(execute(SCI_GETSELECTIONNCARET, selection, 0));
 						execute(SCI_SETSELECTION, caret, caret);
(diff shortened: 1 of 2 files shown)


xomx:
Tried this PR - it fixes the problem reported.

Issue #15076, opened 2024-05-02
[Feature request] More than one bookmark color


Your-Pal-Al:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

Admittedly, I probably don't use bookmarks as they are intended. But they're pretty handy when I want to mark certain lines for various reasons. I think they would be even more useful if they came in different colors.

### Describe the solution you'd like.

Clicking an unbookmarked line will toggle the standard blue bookmark. Clicking the blue bookmark will toggle to red. Clicking red toggles to green. Clicking green toggles the bookmark off again. (Or any other method, but this one seems simple enough. Also, there's no reason the colors should be limited to red, blue, green)

### Debug Information

```shell
Notepad++ v8.6.5   (64-bit)
Build time : Mar 29 2024 - 17:04:43
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 11 Enterprise (64-bit
OS Version : 22H2
OS Build : 22621.3447
Current ANSI codepage : 1252
Plugins : 
    BracketsCheck (1.2.2)
    CodeAlignmentNpp (14.1.107)
    ComparePlus (1.1)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    NppFavorites (1.0.0.1)
    NppQrCode64 (0.0.0.1)
    PlanetCNCNpp64 (3001.21.1123.1)
    RandomValuesNppPlugin (0.2.1)
    RegexTrainer (1)
```

alankilborn:
It's not a bad idea, but it probably isn't the best for all users of bookmarks (burdening them with "n" extra clicks to turn off a blue bookmark).

Your-Pal-Al:
I agree, I figured there would need to be a setting within preferences to enable/disable this feature. Or even a menu for enable each color of bookmarks. Alternatively, right clicking the bookmark and selecting a color from a context menu would be cool

PR #15133, opened 2024-05-13
Red blue toggle of bookmarks 20240513

ockedee:
ref #15076 
Added logic to toggle the color of bookmarks between red and blue.


donho:
The bookmark behaviour of the PR becomes very awkward.
The idea from #15076 is too vague to be implemented.

PR #15142, opened 2024-05-15
Add plugin command NPPM_GETTABCOLORID

### LINKED ISSUE #15115: [Feature request] Allow plugins to get/set the existing color choice for a tab

### Description of the Issue

Currently no way for plugins to manipulate tab coloring, as is settable in N++'s _View_ menu > _Tab_ > _Apply Color ..._

Implementation of NPPM_GET_TABCOLORINDEX and NPPM_SET_TABCOLORINDEX as plugin commands.

### Debug Information

```shell
Notepad++ v8.6.6   (64-bit)
Build time : May  6 2024 - 21:59:00
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.6\npp.8.6.6.portable.x64\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 23H2
OS Build : 22631.3447
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

A scripting reason for wanting to be able to get tab coloring is discussed here: https://community.notepad-plus-plus.org/post/94744

alankilborn:
Fix #15115 

Note that the implementation I chose was to allow "setting" only for the active tab in the active view.  "Getting", however, can be done on any tab in either view.  Mainly this choice was to "keep it simple", but in reality it isn't that big of an "ask" to whatever calls the "set" to make the tab active first.

Other choices were arbitrary as well, e.g. using `-1` to refer to the active tab (or view).  I don't know if these choices will be liked.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2955,9 +2955,11 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_GETTABCOLORID:
 		{
-			const auto view = static_cast<INT>(wParam);
-			auto tabIndex = static_cast<INT>(lParam);
+			const auto view = static_cast<int>(wParam);
+			auto tabIndex = static_cast<int>(lParam);
+
 			auto colorId = -1;  // no color (or unknown)
+
 			auto pDt = _pDocTab;  // active view
 			if (view == MAIN_VIEW)
 			{
@@ -2967,14 +2969,17 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			{
 				pDt = &_subDocTab;
 			}
+
 			if (tabIndex == -1)
 			{
 				tabIndex = pDt->getCurrentTabIndex();
 			}
-			if ((tabIndex >= 0) && (tabIndex < static_cast<INT>(pDt->nbItem())))
+			
+			if ((tabIndex >= 0) && (tabIndex < static_cast<int>(pDt->nbItem())))
 			{
 				colorId = pDt->getIndividualTabColour(tabIndex);
 			}
+			
 			return colorId;
 		}
 
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2953,6 +2953,51 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return _pluginsManager.allocateIndicator(static_cast<int32_t>(wParam), reinterpret_cast<int *>(lParam));
 		}
 
+		case NPPM_NPPM_GET_TABCOLORINDEX:
+		{
+			auto tabIndex = static_cast<INT>(wParam);
+			const auto view = static_cast<INT>(lParam);
+			auto colorIndex = -1;  // no color (or unknown)
+			auto pDt = _pDocTab;  // active view
+			if (view == MAIN_VIEW)
+			{
+				pDt = &_mainDocTab;
+			}
+			else if (view == SUB_VIEW)
+			{
+				pDt = &_subDocTab;
+			}
+			if (tabIndex == -1)
+			{
+				tabIndex = pDt->getCurrentTabIndex();
+			}
+			if ((tabIndex >= 0) && (tabIndex < pDt->nbItem()))
+			{
+				colorIndex = pDt->getIndividualTabColour(tabIndex);
+			}
+			return colorIndex;
+		}
+
+		case NPPM_NPPM_SET_TABCOLORINDEX:
+		{
+			// operates only upon active tab in active view
+			const auto newColorIndex = static_cast<INT>(wParam);
+			const auto currentTabIndex = _pDocTab->getCurrentTabIndex();
+			const auto oldColorIndex = _pDocTab->getIndividualTabColour(currentTabIndex);
+			if ((newColorIndex != oldColorIndex) && (newColorIndex >= -1) &&
+				(newColorIndex <= IDM_VIEW_TAB_COLOUR_5 - IDM_VIEW_TAB_COLOUR_NONE - 1))
+			{
+				const auto bufferID = _pDocTab->getBufferByIndex(currentTabIndex);
+				_pDocTab->setIndividualTabColour(bufferID, newColorIndex);
+				_pDocTab->redraw();
+				if (_pDocumentListPanel != nullptr)
+				{
+					_pDocumentListPanel->setItemColor(bufferID);
+				}
+			}
+			return oldColorIndex;
+		}
+
 		case NPPM_GETBOOKMARKID:
 		{
 			return MARK_BOOKMARK;
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2953,11 +2953,11 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return _pluginsManager.allocateIndicator(static_cast<int32_t>(wParam), reinterpret_cast<int *>(lParam));
 		}
 
-		case NPPM_NPPM_GET_TABCOLORINDEX:
+		case NPPM_GETTABCOLORID:
 		{
-			auto tabIndex = static_cast<INT>(wParam);
-			const auto view = static_cast<INT>(lParam);
-			auto colorIndex = -1;  // no color (or unknown)
+			const auto view = static_cast<INT>(wParam);
+			auto tabIndex = static_cast<INT>(lParam);
+			auto colorId = -1;  // no color (or unknown)
 			auto pDt = _pDocTab;  // active view
 			if (view == MAIN_VIEW)
 			{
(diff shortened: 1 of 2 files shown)


PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2953,6 +2953,51 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
+			}
+			else if (view == SUB_VIEW)
+			{
+				pDt = &_subDocTab;
+			}
+			if (tabIndex == -1)
+			{
+				tabIndex = pDt->getCurrentTabIndex();
+			}
+			if ((tabIndex >= 0) && (tabIndex < pDt->nbItem()))
```
rdipardo, line 2974: `nbItem()` [returns][0] a `size_t`.

> [!TIP]
> You could actually dispense with that first `static_cast` in l. 2958. Comparing unsigned values with `-1` works exactly as you would expect; the literal `-1` will be [implicitly converted][1] to an unsigned `DWORD`.

[0]: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/8329a9fec9229bdda61493f88be867219d6b0385/PowerEditor/src/WinControls/TabBar/TabBar.h#L94
[1]: https://godbolt.org/z/96xoW1YKs

donho (changes requested):
The set of the API is not coherent.
Please consider the suggested API:
```cpp
	// int NPPM_GETTABCOLORNUMBER (0, int viewIndex)
	// Get the tab color number with given tab index and view.
	// wParam[in]: 0 (not used)
	// lParam[in]: VIEW & INDEX or -1 (active view)
		// VIEW takes 2 highest bits and INDEX (0 based) takes the rest (30 bits)
		// Here's the values for the view:
		//  MAIN_VIEW 0
		//  SUB_VIEW  1
	//
	// Return tab color number
		// tab color number contains the following values:
		// -1 (no color)
		//  0 (yellow)
		//  1 (green)
		//  2 (blue)
		//  3 (orange)
		//  4 (pink)
 
 .
 
	// int NPPM_SETTABCOLORNUMBER(int tabColorNum, int viewIndex)
	// Set a new tab color by tab color number for the active tab in the active view.
	// wParam[in]: tab color number
		// tab color number contains the following values:
		// -1 (no color)
		//  0 (yellow)
		//  1 (green)
		//  2 (blue)
		//  3 (orange)
		//  4 (pink)
	//
	// lParam: VIEW & INDEX or -1 (active view)
		// VIEW takes 2 highest bits and INDEX (0 based) takes the rest (30 bits)
		// Here's the values for the view:
		//  MAIN_VIEW 0
		//  SUB_VIEW  1
	//
	// Return previous color number of the tab (from -1 to 4)
```
Please check `NPPM_GETPOSFROMBUFFERID` for how to set and retrieve viewIndex

pryrt:
> The set of the API is not coherent. Please consider the suggested API:
> 
> ```c++
> 	// lParam[in]: VIEW & INDEX or -1 (active view)
> 		// VIEW takes 2 highest bits and INDEX (0 based) takes the rest (30 bits)
> 		// Here's the values for the view:
> 		//  MAIN_VIEW 0
> 		//  SUB_VIEW  1
> ```

That's rather complicated from an end-user perspective.  Trying to encode properly, and knowing whether that means
* 0xFFFFFFFF = active ("-1" refers to the whole 32bit integer)
* 0x80000000 = SUB_VIEW ("1" in the MSB)
* 0x00000000 = MAIN_VIEW
... or
* 0xFFFFFFFF (-1) = active
* 0x40000000 = SUB_VIEW  ("1" in a "2 highest bit" perspective is "01", which then encodes to this value)
* 0x00000000 = MAIN_VIEW
... or
* 0xC0000000 = active (because those "2 highest bits" are -1)
* 0x80000000 = SUB_VIEW
* 0x00000000 = MAIN_VIEW
... or
* 0xC0000000 = active (because those "2 highest bits" are -1)
* 0x40000000 = SUB_VIEW
* 0x00000000 = MAIN_VIEW

All four of those are reasonable interpretations of what you wrote.  I think that would cause a lot of headache for plugin/script developers.

Since the IDM_VIEW_TAB_COLOUR_* menu-command-ids already provide a way to SET the color from a menu command (and menu commands can be activated by their IDM through any plugin), we don't really _need_ a SET version of the API message. There are plenty of other messages that only have one of the two directions, so it wouldn't be inconsistent with the API to only have a getter for this.

To me, it would make more sense to drop the SET_ version, and instead just have `GET_TABCOLORNUMBER(tabIndex, inView)` -- that makes the interface simpler, and all the same functionality is available.  I would find that a lot easier to use than having to do a complicated encoding to be able to indicate "active" or "MAIN_VIEW | index" or "SUB_VIEW | index" into a single message argument.

donho:
@pryrt 
> Since the IDM_VIEW_TAB_COLOUR_* menu-command-ids already provide a way to SET the color from a menu command (and menu commands can be activated by their IDM through any plugin), we don't really need a SET version of the API message. 

Indeed, `NPPM_MENUCOMMAND` can be use to replace `NPPM_SETTABCOLORNUMBER`.

> To me, it would make more sense to drop the SET_ version, and instead just have GET_TABCOLORNUMBER(tabIndex, inView) -- that makes the interface simpler, and all the same functionality is available. I would find that a lot easier to use than having to do a complicated encoding to be able to indicate "active" or "MAIN_VIEW | index" or "SUB_VIEW | index" into a single message argument.

It makes sense to me.

@alankilborn 
What do you think?

alankilborn:
@donho said:

>alankilborn, What do you think?

I think Peter, as usual, makes a great deal of sense.
I will rework the PR in the spirit of his excellent guidelines.

alankilborn:
@donho @pryrt 

The only "trouble" remaining before I submit revisions is this:

> `NPPM_NPPM_GET_TABCOLORINDEX` should be `NPPM_GET_TABCOLORNUMBER` 

(BTW, I'm not sure why I originally had doubled up the `NPPM` part.  :-( )

---

When one looks at the menu for coloring a tab, reference:

image

then if we are considering a color _number_, then I'd say the numbering is 1, 2, 3, 4, 5.  However, if one thinks in term of a color _index_, then one might see it as 0, 1, 2, 3, 4.

So, with the current guideline, the nomenclature would be "number" but we'd be returning basically an "index".
I realize that it might be a "fine" point, but I just wanted to be sure that calling it a "number" and returning a zero-based value for the color is really what we want to do?

My preference would be to call it a "number", return 1, 2, 3, 4, 5 for the colors, and return 0 for a tab without color.  But, I often don't get what I want here.  :-)

donho:
@alankilborn 
Your viewpoint about "number" not wrong. However, I would try to avoid index because 
1. tab color index has no meaning to me and to other dev I believe.
2. it's confusing with "tab index". 
3. tab colors in the code is not array, so it's not index

I would use **tab color ID** instead.
For the API name, let's keep simple, so we can just use "NPPM_GETTABCOLOR".

Here's the suggested API specs:
```cpp
// int NPPM_GETTABCOLOR(int inView, int index2Activate)
// Get the tab color ID with given tab index and view.
// wParam[in]: inView - main view (0) or sub-view (1) or -1 (active view)
// lParam[in]: index2Activate - index (in the view indicated above). -1 for currently active tab
// Return tab color ID which contains the following values: 0 (yellow), 1 (green), 2 (blue), 3 (orange), 4 (pink) or -1 (no color)
```

What do you think?

alankilborn:
>What do you think?

Sounds good.
I will implement it that way and update this PR.

ozone10:
I would use `NPPM_GETTABCOLORID`. `NPPM_GETTABCOLOR` can be confusing. Someone could expect that it return color (yellow, ...) directly, and not its ID.

alankilborn:
I think @donho would approve of @ozone10 's suggestion, so I'll go ahead with that direction.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2953,6 +2953,31 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return _pluginsManager.allocateIndicator(static_cast<int32_t>(wParam), reinterpret_cast<int *>(lParam));
 		}
 
+		case NPPM_GETTABCOLORID:
+		{
+			const auto view = static_cast<INT>(wParam);
+			auto tabIndex = static_cast<INT>(lParam);
```
donho, line 2959: In general, it's better to use primitive type than macro defined type.
Here I would use `int tabIndex = static_cast<int>(lParam);` instead of `auto tabIndex = static_cast<INT>(lParam);`

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2953,6 +2953,31 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
+			auto colorId = -1;  // no color (or unknown)
+			auto pDt = _pDocTab;  // active view
+			if (view == MAIN_VIEW)
+			{
+				pDt = &_mainDocTab;
+			}
+			else if (view == SUB_VIEW)
+			{
+				pDt = &_subDocTab;
+			}
```
donho, line 2969: Please add an empty line for the sake of readability.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -2953,6 +2953,31 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
...
+			if (tabIndex == -1)
+			{
+				tabIndex = pDt->getCurrentTabIndex();
+			}
+			if ((tabIndex >= 0) && (tabIndex < static_cast<INT>(pDt->nbItem())))
```
donho, line 2974: Please add `else` for the sake of visibility.

donho (changes requested):
Please consider the following suggestions.

PowerEditor/src/NppBigSwitch.cpp
alankilborn: I added more empty lines throughout; hopefully it is more appealing.

PowerEditor/src/NppBigSwitch.cpp
alankilborn: Hmm, no "else" functionality intended here -- maybe improved empty-line whitespacing helps?

donho (approved):

MarkusBodensee:
This PR is such a great example how collaboration works and how it results in a solid implementation. Nice to read through it, thank you. Wonderful job guys :-)

PR #15146, opened 2024-05-16
Handle dpi messages in Find dialog

ozone10:
ref #14959


ozone10:
@donho 
the only thing missing is handling situation when `goToCenter` is invoked and https://github.com/notepad-plus-plus/notepad-plus-plus/blob/71914ed5453c550882929c1ef19ff6b77d80a5ff/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp#L334

donho:
@ozone10 

Almost...

After dragging from 200% screen to 100% screen:

image

donho:
@ozone10 

The bug of dragging from 200% screen to 100% screen still persist:
image

donho:
@ozone10 
> But could you provide some information. Which scale does your primary screen have?

My main display's scale rate is 225%, external screen 100%

Just tested your latest commit, the issue seems gone - bravo!
The only remain issue is: while the 2nd Ctrl-F to make Find dialog display on the center of Notepad++ (225%) from the external screen (100%), Find dialog increases its width considerably - but you have mentioned "the only thing missing is handling situation when goToCenter is invoked..."

So let's merge this PR and treat the dialog width increasing issue in another PR.

Thank you!

PR #15152, opened 2024-05-17
Fix division by zero

### LINKED ISSUE #15144: [BUG] Inserting numbers with Column Editor causes crash

### Description of the Issue

Attempting to insert certain non-decimal number sequences using the Column Editor crashes the program.

### Steps To Reproduce

1.  Select at least 28 lines in Column Mode or multi-select at least 28 positions
2.  Open Column Editor, select "Number to Insert", set "Increase by" to 9999999, "Repeat" to 1, and "Format" to "Hex"
3.  Press "OK"

### Current Behavior

Program crashes

### Expected Behavior

Numbers are inserted at selected positions

### Debug Information

```shell
Notepad++ v8.6.7   (64-bit)
Command Line :
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 1809
OS Build : 17763.1579
Current ANSI codepage : 1252
Plugins :
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

I have localized the problem and will create a PR as soon as possible.

cddiffz:
Fix #15144

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -165,24 +165,17 @@ LanguageNameInfo ScintillaEditView::_langNameInfoArray[L_EXTERNAL + 1] = {
 
 int getNbDigits(int aNum, int base)
 {
-	int nbChiffre = 1;
-	int diviseur = base;
+	int nbDigits = 0;
 
-	for (;;)
+	do
 	{
-		int result = aNum / diviseur;
-		if (!result)
-			break;
-		else
-		{
-			diviseur *= base;
-			++nbChiffre;
-		}
-	}
-	if ((base == 16) && (nbChiffre % 2 != 0))
-		nbChiffre += 1;
+		++nbDigits;
+		aNum /= base;
+	} while (aNum != 0);
+	if (base == 16 && nbDigits % 2 != 0)
+		++nbDigits;
 
-	return nbChiffre;
+	return nbDigits;
 }
 
 bool isCharSingleQuote(__inout wchar_t const c)


PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -165,20 +165,13 @@ LanguageNameInfo ScintillaEditView::_langNameInfoArray[L_EXTERNAL + 1] = {
 
 int getNbDigits(int aNum, int base)
 {
-	int nbChiffre = 1;
-	int diviseur = base;
+	int nbChiffre = 0;
```
alankilborn, line 168: Why not take the opportunity to convert the code to English at this time?
Suggest `nbChiffre` become `nbDigits`.

cddiffz: You are right. As this is my first contribution, I wanted to change as little as possible. I found more bugs, so I'll know better next time.

cddiffz: Changed identifier as suggested
Fixed author

alankilborn: @donho Hopefully it is ok for me to make the suggestion to convert French to English in the code.  I know that when I submitted some Column Editor code changes previously, I did such language updates and it was accepted...

donho: @alankilborn @cddiffz 
Yes, it's good. 
I did it (using inconsistently French word) because both are not my mother tongue.
However, the code should be done by a international language for the sake that everyone can understand so everyone can maintain it.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -165,24 +165,17 @@ LanguageNameInfo ScintillaEditView::_langNameInfoArray[L_EXTERNAL + 1] = {
...
 int getNbDigits(int aNum, int base)
 {
-	int nbChiffre = 1;
-	int diviseur = base;
+	int nbDigits = 0;
 
-	for (;;)
+	do
 	{
-		int result = aNum / diviseur;
```
chcg, line 173: @cddiffz Did the division by zero happen here?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -165,24 +165,17 @@ LanguageNameInfo ScintillaEditView::_langNameInfoArray[L_EXTERNAL + 1] = {
...
-			diviseur *= base;
-			++nbChiffre;
-		}
-	}
-	if ((base == 16) && (nbChiffre % 2 != 0))
-		nbChiffre += 1;
+		++nbDigits;
+		aNum /= base;
+	} while (aNum != 0);
+	if (base == 16 && nbDigits % 2 != 0)
```
chcg, line 175: @cddiffz For input aNum = 0 and base =16 this would result in nbDigits = 1 from do-while above already here and as 1%2 = 1 != 0 would result in another increment to nbDigits.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
cddiffz: Yes, the divisor can become zero due to an arithmetic overflow.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
cddiffz: But that was already the case before. Ididn't change the semantics at all, just fixed the bug.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
chcg: ok, I see, because also in the new code a div by zero is possible if the provided base parameter is already zero by the caller. So I think an additional check for that makes sense.

cddiffz: The actual calling code first sets the base to 10, then possibly changes it to either 2, 8 or 16. If zero were passed as the base, the calling code would be flawed anyway, wouldn't it? Trying to catch this makes no sense.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
chcg: Yes, that is correct also it seems to be strange.

cddiffz: This is another bug for me. With "Format" set to "Hex", e.g. instead of "100", "100 " is inserted because of the rounding to the next even number.

cddiffz: See #15168

PR #15158, opened 2024-05-19
Move window for dpi change for goToCenter()

ozone10:
ref #14959

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -349,7 +349,9 @@ void FindReplaceDlg::create(int dialogID, bool isRTL, bool msgDestParent, bool t
 	if (nppGUI._findWindowPos.bottom - nppGUI._findWindowPos.top != 0)  // check height against 0 as a test of valid data from config
 	{
 		RECT rc = getViewablePositionRect(nppGUI._findWindowPos);
-		::SetWindowPos(_hSelf, HWND_TOP, rc.left, rc.top, rc.right - rc.left, rc.bottom - rc.top, swpFlags);
+		::SetWindowPos(_hSelf, HWND_TOP, rc.left, rc.top, 0, 0, SWP_HIDEWINDOW | SWP_NOSIZE | SWP_NOACTIVATE);
+		::SetWindowPos(_hSelf, HWND_TOP, 0, 0, _rc.right - _rc.left, _rc.bottom - _rc.top, swpFlags | SWP_NOMOVE);
+
 		if ((swpFlags & SWP_SHOWWINDOW) == SWP_SHOWWINDOW)
 			::SendMessageW(_hSelf, DM_REPOSITION, 0, 0);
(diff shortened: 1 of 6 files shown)


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -349,7 +349,9 @@ void FindReplaceDlg::create(int dialogID, bool isRTL, bool msgDestParent, bool t
 	if (nppGUI._findWindowPos.bottom - nppGUI._findWindowPos.top != 0)  // check height against 0 as a test of valid data from config
 	{
 		RECT rc = getViewablePositionRect(nppGUI._findWindowPos);
-		::SetWindowPos(_hSelf, HWND_TOP, rc.left, rc.top, rc.right - rc.left, rc.bottom - rc.top, swpFlags);
+		::SetWindowPos(_hSelf, HWND_TOP, rc.left, rc.top, 0, 0, SWP_HIDEWINDOW | SWP_NOSIZE | SWP_NOACTIVATE);
+		::SetWindowPos(_hSelf, HWND_TOP, 0, 0, _rc.right - _rc.left, _rc.bottom - _rc.top, swpFlags | SWP_NOMOVE);
```
ozone10, line 353: @donho 
currently it resets saved dialog size to min size.
To used saved size, more works need to be done. Like saving unscaled size and than scale it or saving used dpi.

ozone10:
PR needs testing for hash, run, about, debug, and mainly find dialogs.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
donho: @ozone10 
I think this behaviour is OK - it's original behaviour.

MarkusBodensee: Upvote: It is a good way to reset position and size of the dialog. Please keep it this way :-)

vlakoff:
Side note: I noticed that if pressing "Ctrl + F" when the Find dialog is already open (i.e. press "Ctrl + F" two times), the dialog gets "reset": its size is reset, and it is centered in the window.

alankilborn:
> press "Ctrl + F" two times), the dialog gets "reset": its size is reset, and it is centered in the window.

The centering is intentional; I don't know about the size being reset (but it does get resized on 8.7).

vlakoff:
Agree this recentering may be useful. On the other hand, the dialog size should be preserved (I hope it's possible).

vlakoff:
I just noticed it was also pointed out in [this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15158/files#diff-62585f441298d642a3dbd871af2800eac3606f9fe148372ab119feed21e050b0R353):

> currently it resets saved dialog size to min size.
To used saved size, more works need to be done. Like saving unscaled size and than scale it or saving used dpi.

I would still love to have the dialog size preserved, but well, we'll have to make do.

PR #15159, opened 2024-05-19
Add auto-indent for Python

### LINKED ISSUE #15122: [Feature request] Auto-indentation in Python

- [ ] I have searched the existing issues

### Description of the Issue

In Python it's nessessary to indent every statement inside an if, loop, function definition or similar block. 

Therefore e.g. IDLE automatically indents the next line with an additional tab when you press enter after a colon. 
But apparently Notepad++ doesn't support this.

It would be very helpful if you could add this feature to Notepad++ too. 
(But maybe _only_ for files in Python language, as in other contexts you may not want this to happen. 
And you should be able to disable it in the settings.)

### Debug Information

```shell
Notepad++ v8.6.5   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 10 Pro (64-bit
OS Version : 22H2
OS Build : 19045.4291
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

alankilborn:
Fix #15122

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3683,6 +3683,35 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 			*/
 		}
 	}
+	else if (type == L_PYTHON)
+	{
+		if (((eolMode == SC_EOL_CRLF || eolMode == SC_EOL_LF) && ch == '\n') ||
+			(eolMode == SC_EOL_CR && ch == '\r'))
+		{
+			// Search the non-empty previous line
+			while (prevLine >= 0 && _pEditView->getLineLength(prevLine) == 0)
+				prevLine--;
+
+			// Get previous line's Indent
+			if (prevLine >= 0)
+			{
+				indentAmountPrevLine = _pEditView->getLineIndent(prevLine);
+			}
+
+			_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP | SCFIND_POSIX);
+			_pEditView->execute(SCI_SETTARGETRANGE, _pEditView->execute(SCI_POSITIONFROMLINE, prevLine), 
+				_pEditView->execute(SCI_GETLINEENDPOSITION, prevLine));
+			const char colonExpr[] = ":[ \t]*(#|$)";  // colon optionally followed by only whitespace and/or start-of-comment
+			if (_pEditView->execute(SCI_SEARCHINTARGET, strlen(colonExpr), reinterpret_cast<LPARAM>(colonExpr)) >= 0)
+			{
+				_pEditView->setLineIndent(curLine, indentAmountPrevLine + tabWidth);
+			}
+			else if (indentAmountPrevLine > 0)
+			{
+				_pEditView->setLineIndent(curLine, indentAmountPrevLine);
+			}
+		}
+	}
 	else // Basic indentation mode
 	{
 		if (((eolMode == SC_EOL_CRLF || eolMode == SC_EOL_LF) && ch == '\n') ||
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3699,9 +3699,13 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
 			}
 
 			_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP | SCFIND_POSIX);
-			_pEditView->execute(SCI_SETTARGETRANGE, _pEditView->execute(SCI_POSITIONFROMLINE, prevLine), 
-				_pEditView->execute(SCI_GETLINEENDPOSITION, prevLine));
+
+			auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, prevLine);
+			auto endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, prevLine);
+			_pEditView->execute(SCI_SETTARGETRANGE, startPos, endPos);
+
 			const char colonExpr[] = ":[ \t]*(#|$)";  // colon optionally followed by only whitespace and/or start-of-comment
+
 			if (_pEditView->execute(SCI_SEARCHINTARGET, strlen(colonExpr), reinterpret_cast<LPARAM>(colonExpr)) >= 0)
 			{
 				_pEditView->setLineIndent(curLine, indentAmountPrevLine + tabWidth);


alankilborn:
@BlueBatSamurai
If you want to test this implementation of your feature request, that would be great.

donho:
Can any one here knows Python test this PR please?

mpheath:
Test by inserting the caret to the EOL and pressing <kbd>Enter</kbd>

```py
:
": #"
""":
:#"""
else:
elsse:
```

The `else:` should indent the next line and the others should not indent the next line.

Currently, the *Python Indent* plugin installable from *Plugin Admin* is the better choice IMO as it determines conditionally to indent or not to indent. Indenting almost any line with a colon at EOL as the PR does can be disruptive if the indenting is incorrect. When both plugin and PR indenting are enabled, the plugin fixes the actions of the PR indenting and then needs more undo actions so it is worse for undo to have both enabled. Disabling Auto-indent checkbox prevents the extra undo behaviour though affects all tabs that may have other languages where Auto-indent is preferred.

Builtin auto indenting for Python seems like a nice idea and the code looks tidy, I suggest to improve the regex to lessen indenting that should not occur:

```
"^\\h*(?:async|case|class|def|elif|else|except|finally|for|if|match|try|while|with)\\b.*?:\\h*(?:#|$)"
```

or perhaps make `async` with the following keyword more robust:

```
"^\\h*(?:async\\h+def|async\\h+for|async\\h+with|case|class|def|elif|else|except|finally|for|if|match|try|while|with)\\b.*?:\\h*(?:#|$)"
```


The *Python Indent* plugin checks for these keywords. [plugin code](https://sourceforge.net/p/kered13-notepad-plugins/code/ci/default/tree/src/PythonIndent/PythonIndent.cpp)

alankilborn:
@mpheath 

Thanks for killing my PR.  :-)

I was unaware that there was a plugin that did this functionality; I have long used a script to do it, but when the 15122 issue came along I thought it made sense to code it natively.

The idea behind the PR code was to keep it _simple_:  If a line "ends" with a colon (`:`), indent the next line one more level.  Although...I did get a _little_ fancy with it by allowing whitespace and/or a comment to be after the colon (because I myself very often put a comment on an "if :" line).

C-like auto-indent behavior for other languages in N++ is far from perfect, and again I was trying to keep Python indenting simple and not "perfect".

---

>Indenting almost any line with a colon at EOL ... can be disruptive if the indenting is incorrect.

I don't know what this means.  IMO indenting after any line with a `:` as the last real character is fine, so can you provide a counter example(s)?

---

>When both plugin and PR indenting are enabled, the plugin fixes the actions of the PR indenting and then needs more undo actions so it is worse for undo to have both enabled. 

Why even try it with plugin and this PR together?  Makes no sense.

---

> I suggest to improve the regex 

I'd prefer not, mainly because I don't think it is necessary.
But I would, to get a code change accepted.
And also to prevent future nay-sayers saying that the plugin works better than the native code.

---

>Builtin auto indenting for Python seems like a nice idea and the code looks tidy

Well...something positive!  :-)
Thanks for your comments and opinions; if they kill the PR, which I think is extremely likely...so be it.

donho:
@mpheath 
Thank you for testing!

@alankilborn 
> Thanks for killing my PR. :-)

Not necessarily.
It's better if the feature is complete. But if not, the idea is to provide a rudimentary functionality:
The new user can benefit immediately from the builtin basic Python auto-indent, without knowing the existence of the plugin.
For the experienced users who are not satisfied with the builtin basic feature, they can always install the plugin to override the builtin functionality.

alankilborn:
@donho 

OK, I like your "basic functionality" statement.

...So what should be done next to move this PR forward?

donho:
> ...So what should be done next to move this PR forward?

I will test it as well, to make sure it's useful for the new users, then do the code review. :)

BlueBatSamurai:
> @BlueBatSamurai If you want to test this implementation of your feature request, that would be great.

Sorry but I’m new on GitHub…
Could anyone tell me how to use this in Notepad++?
I’ve seen this linked code, but what do I have to do with it?

alankilborn:
@BlueBatSamurai 
Get new `notepad++.exe` from [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/9147359139/artifacts/1517017509).
Rename to something like `notepad++_PR15159.exe`.
Copy into the folder where your existing `notepad++.exe` is.
Close Notepad++ if you happen to have it running.
Run `notepad++_PR15159.exe`.

BlueBatSamurai:
@alankilborn 
Thanks!

BlueBatSamurai:
I've dowloaded and renamed the new `Notepad++.exe`. 

And it works! Even if there is a comment behind the colon before you press Enter.
Well done!

(And yes, it doesn't check whether there is a valid keyword before the colon... Something like `qwert: + Enter` causes indentation too. But the user should know that for themselves.)

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3683,6 +3683,35 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
...
+
+			// Get previous line's Indent
+			if (prevLine >= 0)
+			{
+				indentAmountPrevLine = _pEditView->getLineIndent(prevLine);
+			}
+
+			_pEditView->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP | SCFIND_POSIX);
+			_pEditView->execute(SCI_SETTARGETRANGE, _pEditView->execute(SCI_POSITIONFROMLINE, prevLine), 
+				_pEditView->execute(SCI_GETLINEENDPOSITION, prevLine));
```
donho, line 3703: Improve readability:
```cpp
auto startPos = _pEditView->execute(SCI_POSITIONFROMLINE, prevLine);
auto endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, prevLine);
_pEditView->execute(SCI_SETTARGETRANGE, startPos, endPos);
```

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -3683,6 +3683,35 @@ void Notepad_plus::maintainIndentation(TCHAR ch)
...
+			const char colonExpr[] = ":[ \t]*(#|$)";  // colon optionally followed by only whitespace and/or start-of-comment
```
donho, line 3704: Please add a empty line after the above line.

donho (changes requested):
Please follow the suggestion:

PR #15160, opened 2024-05-19
Make top bar indicate applied tab color for unfocused view

### LINKED ISSUE #15140: [BUG] Tab color of active tab in inactive view is unknown

### Description of the Issue

When a tab in the inactive view _has_ color, that color is not known to the user because the tab visually appears like a non-colored tab.

### Steps To Reproduce

1.  In a fresh N++, create `new 1`, `new 2`, `new 3`, `new 4` and `new 5` tabs
2.  Color each tab (except `new 5`) with the same color number as its tab title, e.g. tab `new 1` gets "color 1", etc., via _View_ (menu) > _Tab_ > _Apply Color ..._
3.  Move documents `new 3` and `new 4` to the "other" (secondary or sub) view
4. Experiment with making the tabs active, in turn.

### Current Behavior

Note that when either `new 3` or `new 4` is the active document in the active secondary view:

* the appearance of `new 5` (an uncolored tab) in the non-active view, when it is the active tab in that view:
image

* the appearance of `new 2` (a colored tab) in the non-active view, when it is the active tab in that view:
image

These tabs appear visually the same, even though one is colored and one is not.

The bottom line: User has no way to know that `new 2` is a tab that has had color applied to it in this situation.

### Expected Behavior

Some way to know the color of the active tab in the non-active view.

### Debug Information

```shell
Notepad++ v8.6.7   (64-bit)
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.7\npp.8.6.7.portable.x64\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 23H2
OS Build : 22631.3593
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

ozone10:
- use dpi scaled metrics

fix #15140


alankilborn:
Tested and works great; thanks @ozone10

PR #15169, opened 2024-05-21
Refactor Shortcut Mapper resizing for dpi compatibility

ozone10:
ref #14959


donho:
@ArkadiuszMichalski 
Now Github CI doesn't work parallel anymore. Any idea about that? 
Hope it's temporarily (due to unavailable resource at this moment).

ozone10:
@donho
https://www.githubstatus.com/

>
>Update - We have identified network connectivity issues causing delays in Actions Workflow Runs, Workflow Job Runs, and Check >Steps. We are continuing to investigate.
>May 21, 2024 - 14:34 UTC
>Update - We are investigating delayed updates to Actions job statuses.
>May 21, 2024 - 13:58 UTC
>Investigating - We are investigating reports of degraded performance for Actions
>May 21, 2024 - 12:45 UTC

PR #15173, opened 2024-05-22
Handle dpi message in Shortcut Mapper

ozone10:
ref #14959


ozone10:
Needs testing.
Known issues: currently column widths are not automatically resized after dpi change and mainly cannot shrink. I haven't check babygrid code yet.

PR #15174, opened 2024-05-22
Update Swedish localization

krilbe:
The Mark dialog has "always" been missing a few keyboard hotkeys in the Swedish translation.

I've added them, and tried to adhere to letters used for similar command elsewhere.

I've tried to make sure there are no key collisions (before my changes, the keys I chose produced a Windows "ding" sound when I tried them).


myTempTemp:
傻B侯今吾，侯今吾天天吃屎

PR #15204, opened 2024-05-27
Replace bmp with ico for panels with toolbar

ozone10:
- use correct styles for toolbar
- change panel toolbar icons if dark/light mode has changed
- align panel toolbar icons
- add initialization

ref #14959


donho:
@ozone10 
Could you provide the argument(s) for justifying using ICO over BMP?

ozone10:
@donho 
- ico can contain multiple image with different sizes -> this is main purpose of this PR to handle dpi
- better handling of transparency
- bmp files are typically larger

PR #15211, opened 2024-05-28
Disallow tab character via pasting when renaming a filetab

### LINKED ISSUE #15202: [BUG] Using a tab character when renaming an unsaved tab is problematic

### Description of the Issue

See Steps To Reproduce.

### Steps To Reproduce

1. Get `b4tab\tafterTab` into your clipboard (where `\t` is not literal but is a true tab character) 
2. Right-click any `new #` tab and choose _Rename_ from the popup menu
3. Paste into the `New name` box in the `Rename Current Tab` dialog box, see:
image

5. Press OK to obtain:
image

6. Attempt to exit N++; receive this prompt:
image

### Current Behavior

A tab character gets into the tab/file name (and causes some problems).

### Expected Behavior

Tab characters should be disallowed here.

### Debug Information

```shell
Notepad++ v8.6.7   (64-bit)
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.7\npp.8.6.7.portable.x64\notepad++.exe
Command Line : -titleAdd="8.6.7 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 23H2
OS Build : 22631.3593
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    PythonScript (3.0.16)
```

### Anything else?

_No response_

alankilborn:
Fix #15202

Note that since tab characters are prevented from being used, the error message that mentions them has been updated to NOT mention them (as it is unnecessary to mention them).

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1402,7 +1402,8 @@ Continue?"/><!-- HowToReproduce: when you openned file is modified and saved, th
 			<FileLockedWarning title="Save failed" message="Please check whether if this file is opened in another program"/>
 			<FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/><!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
 			<RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use on another tab."/><!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
-			<RenameTabTemporaryNameIsEmpty title="Rename failed" message="The specified name cannot be empty, or it cannot contain only space(s)."/><!-- HowToReproduce: Rename the tab of an untitled document and provide an empty string or only some white speces. -->
+			<RenameTabTemporaryNameIsEmpty title="Rename failed" message="The specified name cannot be empty, or it cannot contain only space(s) or TAB(s)."/><!-- HowToReproduce: Rename the tab of an untitled document and provide an empty string or only some white speces. -->
+			<RenameTabTemporaryNameGeneralFailure title="Rename failed" message="The specified name is not valid."/><!-- HowToReproduce: Rename the tab of an untitled document and provide a reserved name such as COM1 for the new name. -->
 			<DeleteFileFailed title="Delete File" message="Delete File failed"/><!-- HowToReproduce: this message prevents from system failure. It's hard to reproduce. -->
 
 			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1403,7 +1403,6 @@ Continue?"/><!-- HowToReproduce: when you openned file is modified and saved, th
 			<FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/><!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
 			<RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use on another tab."/><!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
 			<RenameTabTemporaryNameIsEmpty title="Rename failed" message="The specified name cannot be empty, or it cannot contain only space(s) or TAB(s)."/><!-- HowToReproduce: Rename the tab of an untitled document and provide an empty string or only some white speces. -->
-			<RenameTabTemporaryNameGeneralFailure title="Rename failed" message="The specified name is not valid."/><!-- HowToReproduce: Rename the tab of an untitled document and provide a reserved name such as COM1 for the new name. -->
 			<DeleteFileFailed title="Delete File" message="Delete File failed"/><!-- HowToReproduce: this message prevents from system failure. It's hard to reproduce. -->
 
 			<NbFileToOpenImportantWarning title="Amount of files to open is too large" message="$INT_REPLACE$ files are about to be opened.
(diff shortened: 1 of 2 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -1934,7 +1935,7 @@ bool Notepad_plus::fileRename(BufferID id)
 			{
 				_nativeLangSpeaker.messageBox("RenameTabTemporaryNameIsEmpty",
 					_pPublicInterface->getHSelf(),
-					L"The specified name cannot be empty, or it cannot contain only space(s) or TAB(s).",
+					L"The specified name cannot be empty, or it cannot contain only space(s).",
```
donho, line 1938: There's no reason to change it, since it's the fact.
Please keep as it was.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1402,7 +1402,7 @@ Continue?"/><!-- HowToReproduce: when you openned file is modified and saved, th
...
 			<FileAlreadyOpenedInNpp title="" message="The file is already opened in Notepad++."/><!-- HowToReproduce: Open a new document and open a file "c:/tmp/foo", save this new document by choosing "c:/tmp/foo" as file to save, reply the override popup "yes", then this message appears. -->
 			<RenameTabTemporaryNameAlreadyInUse title="Rename failed" message="The specified name is already in use on another tab."/><!-- HowToReproduce: Rename the tab of an untitled document and provide a name that is the same as an already-existing tab of an untitled document. -->
-			<RenameTabTemporaryNameIsEmpty title="Rename failed" message="The specified name cannot be empty, or it cannot contain only space(s) or TAB(s)."/><!-- HowToReproduce: Rename the tab of an untitled document and provide an empty string or only some white speces. -->
+			<RenameTabTemporaryNameIsEmpty title="Rename failed" message="The specified name cannot be empty, or it cannot contain only space(s)."/><!-- HowToReproduce: Rename the tab of an untitled document and provide an empty string or only some white speces. -->
```
donho, line 1405: Please keep as it was.

donho (changes requested):
Please follow the suggestion.

PowerEditor/src/NppIO.cpp
MarkusBodensee: Or alternatively maybe more generic "... cannot contain only white space(s)." ?

ozone10:
Aren't all ASCII characters with number bellow 31 forbidden. What about renaming only to ".." or ".",should this also be forbidden?
And IIRC there are also some reserved names like "COM" and "NUL".

alankilborn:
>Aren't all ASCII characters with number bellow 31 forbidden. What about renaming only to ".." or ".",should this also be forbidden?
And IIRC there are also some reserved names like "COM" and "NUL".

Actually... I'm (now) not sure really what the point of this entire feature is:

image

My original thinking was that this is to allow a user to get rid of the meaningless "new #" name for tabs that they want to keep but not (ever) save into the file-system, i.e., let N++ manage this data via save-unsaved-tabs-on-exit.  Call this case "A".

But now, with this discussion and some further experimentation (and really thinking about the code and the reserved-characters detection), this feature might be a way to provide a new tab with its eventual-save-into-the-file-system name (filename only for now, no path).  Call this case "B".

My intention with this issue+PR was to handle a corner case of getting a tab character into the case "A" situation, because a tab character really causes problems (as demonstrated in the issue).

With case "A", it doesn't matter if user wants to rename a `new #` tab to `<testing>` or `con` or `nul` or `..`.  Hmm, ok, it DOES matter because N++ needs to use this name to create its 7-second backup...where "con" would work but "<" in a filename would not work.

But if reality is more of the case "B" situation, then there's more to it, as @ozone10 starts pointing out, above.

Maybe let the user rename using ANY character -- well, maybe except tab :-) -- where a backup file name could be created and used.  If the user eventually wants to save into the file system, and the name isn't good, the standard-dialog FileSaveAs logic will detect that and force the user to change.

I'm interested in your further thoughts, @ozone10, and really anyone that cares to chime in on this.

ozone10:
@alankilborn 
I am more for 'B'.

I would like at least that `..` and `.` to be included in this PR. Just try to rename tab to them. And try to rename it again. 
And as you said there is issue with backup with tabs with those names. 

I can imagine someone working 3 months on document and not saving it and have backup enabled. Then some corruption happen, "uff I am okay there is backup" -> "where is my backup!!!!!!!!!!". Sorry backup did not work due to naming.

donho:
@alankilborn 

> I would like at least that .. and . to be included in this PR. Just try to rename tab to them. And try to rename it again.
And as you said there is issue with backup with tabs with those names.

Agree with @ozone10 .

> You can use this existing func (or parts of its logic, as you like): 

Also, please check @xomx mentioned function so we can ensure that snapshot will work in every single tab naming.

alankilborn:
Yes, I will turn this into a better thing. :-)

alankilborn:
After some research, it seems that determining if a filename is going to be valid is not really a straightforward thing.  I don't get a good feeling that `isUnsupportedFileName` does the definitive job, one simple example is that it checks for `NUL` but not `nul`.  Sure, it could be changed, but even then does it really hit all of the invalid cases?  It's hard to know...

Since in the case for this issue we are only concerned about a valid/invalid file name (and not the complete path), would a reasonable approach be to do a `::CreateFile()` on a pathname constructed with the user-supplied file name and see if it succeeds/fails?  The directory part of the pathname could be e.g. the backup directory.

Comments from previous issue commentors welcomed (and needed), as well as from any other lurkers. :-)

xomx:
> I don't get a good feeling that `isUnsupportedFileName` does the definitive job, one simple example is that it checks for `NUL` but not `nul`.

Good point. The fix is simple - because of the names in blacklist for comparison use only basic ASCII chars, we can simply  upperize the `fileNameOnly` string there before the final case sensitive `std::find`.

> Since in the case for this issue we are only concerned about a valid/invalid file name (and not the complete path), would a reasonable approach be to do a ::CreateFile() on a pathname constructed with the user-supplied file name and see if it succeeds/fails?

Interesting universal idea. IDK exactly why, but I would probably be reluctant to use a real IO at this point, I need to think more about it. Maybe others will think otherwise(?).

> The directory part of the pathname could be e.g. the backup directory.

If so, rather use %TEMP% dir, user can have backups disabled or its path invalid etc.

alankilborn:
I've thought more about it myself, and I don't really think it is worth "fixing".
Using a name such as `con` (or `CON`, etal) isn't going to be a problem for Notepad++'s backup system, because the name is "mangled" before the backup-file is saved.
If the user names his temporary tab `con` and then at some point attempts to save the file, the _Save As_ dialog will reject `con` as a filename and user will have to choose something else, i.e., correct the problem at that point.

What **_IS_** a problem for Notepad++ is a tab character getting in there (original subject of this PR and issue).
So I'm going to make the changes that were requested regarding that, and be done with it.

donho (approved):

PR #15212, opened 2024-05-28
Fluent Toolbar Icons for better scaling

WildByDesign:
- 16x16, 20x20, 24x24, 32x32, 40x40, 48x48, 64x64 and 96x96

I have tested in dark mode quite thorough. Please do further testing in light mode also before merging.


ozone10:
@WildByDesign 
icons appear to be smaller. 
For macro I would use previous icons square for stop and triangle for playback, now everything is circle and it can be hard to distinguish them. 
Also provide icon for monitoring disabled state (gray color), it's currently not used, but should. 
Copy icons in disabled state in dark mode don't have same color as other icons in disabled state.

WildByDesign:
Excellent feedback, I appreciate it.

> icons appear to be smaller.

Yes. I actually started working on an updated set about an hour ago to deal with this problem. I used the Fluent SVGs identical to how they are supplied from their repo. In the Fluent sets from several years back, I actually addressed this problem by using the SVG object from the next size up. For example, I would stuff the 20px SVG into the 16px SVG, the 24px SVG into the 20px SVG. The only downside is that I will have to manually edit each SVG file individually and so it will take a few extra days.

I'm excited to fix this because my 125% scaling uses the 20px icons and they need to be sharper.

> For macro I would use previous icons square for stop and triangle for playback, now everything is circle and it can be hard to distinguish them.

I agree with you. After I made that change, I started to second guess it. I will change those for sure.

> Also provide icon for monitoring disabled state (gray color), it's currently not used, but should. Copy icons in disabled state in dark mode don't have same color as other icons in disabled state.
> 
> For disabled state use color #616060 instead of #4e4e4e in dark mode and in light mode use #c2c2c2 instead of #b1b1b1.

I will address these details as well. Thank you for your feedback. Let's hold off on this PR until I can recreate the icons to address the issues mentioned. I will close this PR now.

The good thing is that I have more scripting to make this a smoother process this time.

PR #15218, opened 2024-05-30
Add missing monitoring icon for disabled state

### LINKED ISSUE #15217: [BUG] Missing disabled fluent icon for monitoring

### Description of the Issue

When using fluent icon sets, and "new 1" tab "Monitoring (tail -f)" button on toolbar is disabled, but icon show enabled state.

### Steps To Reproduce

1.  Have "New 1" (some document not saved on disk) opened
2.  Notice Monitoring button on toolbar, it's not possible to click on it
3.  Switch to tab with document saved on disk, there is no change in button appearance, you should be able now click on it now

### Current Behavior

Same appearance for both disabled and enabled states.

### Expected Behavior

Different appearance for both states. 

### Debug Information

```shell
Notepad++ v8.6.7
```

### Anything else?

_No response_

ozone10:
fix #15217


ozone10:
Ico files should be treated as placeholder until @WildByDesign prepare better ones.

PR #15222, opened 2024-05-31
Fix integer overflow issue in Column Editor

### LINKED ISSUE #15167: [BUG] Integer overflow in Column Editor

### Description of the Issue

When you insert a sequence of numbers with the Column Editor, you expect the sequence to be monotonically increasing, but this is not the case if the number becomes too large at some point.

### Steps To Reproduce

1.  Create a new file with 216 empty lines
2.  Place the cursor in the first line
3.  Open Column Editor, select "Number to Insert", set "Increase by" to 9999999, "Repeat" to 1, and "Format" to "Dec"
4.  Press "OK"

### Current Behavior

A negative number is displayed on the last line.

### Expected Behavior

A consistent positive number is displayed on the last line.

### Debug Information

```shell
Notepad++ v8.6.7   (64-bit)
Command Line :
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 1809
OS Build : 17763.1579
Current ANSI codepage : 1252
Plugins :
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

donho:
Fix #15167

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.h
@@ -245,10 +245,106 @@ const std::vector<std::vector<const char*>> g_nonPrintingChars =
 	{"\xEF\xBF\xBB", "IAT", "U+FFFB"}        // U+FFFB : interlinear annotation terminator
 };
 
-int getNbDigits(int aNum, int base);
-//HMODULE loadSciLexerDll();
+size_t getNbDigits(size_t aNum, size_t base);
 
-TCHAR* int2str(TCHAR* str, int strLen, int number, int base, int nbDigits, ColumnEditorParam::leadingChoice lead);
+template<typename T>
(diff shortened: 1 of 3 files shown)


PowerEditor/src/ScintillaComponent/ScintillaEditView.h
```diff
@@ -245,10 +245,106 @@ const std::vector<std::vector<const char*>> g_nonPrintingChars =
...
+	size_t numberStrLen = numberStr.length();
+	size_t noneUsedZoneLen = nbDigits - numberStrLen;
+
+	size_t nbStart = 0;
+	size_t nbEnd = 0;
+
+	size_t noneUsedStart = 0;
+	size_t noneUsedEnd = 0;
+
+	char noneUsedSymbol = ' ';
```
rdipardo, line 309: Shouldn't this be *__T__ noneUsedSymbol = ...* ?

PR #15224, opened 2024-05-31
[XML] Update french.xml - Align with english.xml

Mateos81:
Update french.xml - Align with english.xml


donho:
There are only the format changes, so #15272 will be used instead.

Mateos81:
Aight, reported a mistake over there btw

PR #15227, opened 2024-05-31
Make clang warnings flags in makefile and msvc consistent

ozone10:
- fix some clang warnings


ozone10:
PR #15189 added option to use clang with gcc makefile. This PR makes clang warnings flags in makefile and msvc consistent.

PR #15229, opened 2024-06-01
Status bar indentation indicator

### LINKED ISSUE #11739: [Feature Request] Indentation type in status bar
Hi, sometimes I am working with than one files and they can have different indentation styles.

I would like to maintain same style, but changing between tab and space in notepad++ is not simple.
Currently I have to go to preference -> Language and check/uncheck Replace by space.

I can set indentation based on language, but it does not help much if different files from same language have different style.

So I would like to request to have indentation type in status bar, with (double) left clicking to change between tab and space. And right clicking to open context menu with option to select common tab size (2, 4, 8) and custom size.

Thanks.

ozone10:
fix #11739

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -377,9 +377,10 @@ Translation note:
 					<Item id="48020" name="Indent using tabs"/>
 					<Item id="48021" name="Indent using spaces"/>
 					<Item id="48022" name="Indent size 2"/>
-					<Item id="48023" name="Indent size 4"/>
-					<Item id="48024" name="Indent size 2"/>
-					<Item id="48025" name="Open settings"/>
+					<Item id="48023" name="Indent size 3"/>
+					<Item id="48024" name="Indent size 4"/>
+					<Item id="48025" name="Indent size 8"/>
+					<Item id="48026" name="Open settings"/>
 					<Item id="48009" name="Shortcut Mapper..."/>
 					<Item id="48011" name="Preferences..."/>
 					<Item id="48014" name="Open Plugins Folder..."/>
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -75,6 +75,7 @@ Translation note:
 					<Item subMenuId="encoding-westernEuropean" name="Western European"/>
 					<Item subMenuId="encoding-vietnamese" name="Vietnamese"/>
 					<Item subMenuId="language-userDefinedLanguage" name="User Defined Language"/>
+					<Item subMenuId="settings-indentation" name="Indentation"/>
 					<Item subMenuId="settings-import" name="Import"/>
 					<Item subMenuId="tools-md5" name="MD5"/>
 					<Item subMenuId="tools-sha1" name="SHA-1"/>
@@ -372,6 +373,13 @@ Translation note:
 					<Item id="48005" name="Import plugin(s)..."/>
 					<Item id="48006" name="Import style theme(s)..."/>
 					<Item id="48018" name="Edit Popup ContextMenu"/>
+					<Item id="48019" name="Use default values"/>
+					<Item id="48020" name="Indent using tabs"/>
+					<Item id="48021" name="Indent using spaces"/>
+					<Item id="48022" name="Indent size 2"/>
+					<Item id="48023" name="Indent size 4"/>
+					<Item id="48024" name="Indent size 2"/>
+					<Item id="48025" name="Open settings"/>
 					<Item id="48009" name="Shortcut Mapper..."/>
 					<Item id="48011" name="Preferences..."/>
 					<Item id="48014" name="Open Plugins Folder..."/>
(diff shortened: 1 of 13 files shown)


ozone10:
It's getting stuck on test with udl-NppExec and I don't know why.

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/11739#issuecomment-2143561425

sXNA:
Yes this is what i want. Good work.

alankilborn:
If one chooses the "Open settings" choice from the new popup menu, and the active file/tab is e.g. a Python one, when the _Preferences_ window opens, should Python be made the current choice in the list box?

alankilborn:
The creator of the issue said:

>sometimes I am working with (more) than one file and they can have different indentation styles

I tested this PR and it does not support this:

* Open two .cpp files; when done have one of them be the active file/tab
* In status bar is visible the N++ default settings for .cpp files: `Tabs: 4`
* Right-click the new status bar element and select "spaces" and "2" -- now status bar shows `Spaces: 2` (correct)
* Switch the active tab to the other .cpp file -- status bar still shows `Spaces: 2` when, I expect to see `Tabs: 4`

In my view, this new feature should allow setting non-standard values for individual tabs, and, it should not permanently change any settings in _Preferences_ > _Language_ > _Indent settings_.  If this new feature can't do that, get rid of the popup menu and when user double-clicks on the status bar element, simply open the _Preferences_ window and pre-select the language of the active tab in the listbox there.

---

Also, a question:  Where I work, tab size of 3 is the standard for C-family files -- why can't I achieve such a choice easily with this new feature?

ozone10:
@alankilborn 
this is "basic" PR which does only what was requested.
>So I would like to request to have indentation type in status bar, with (double) left clicking to change between tab and space. And right clicking to open context menu with option to select common tab size (2, 4, 8) and custom size.

I usually work with files with indent size 2 or 4, and know that some source files from unix use 8, and also based on request, I added only those 3 sizes. Adding size 3 is trivial.
For custom size I added option to open settings, since most common sizes are those 3. 

Pre-selecting language could be also done. But with no sure that this PR will be accepted, I don't know if working on it will result only my waste of time.  

I would like to implement it so indent can be per file, meaning to have 3 option: global default, per language, and per file.
This will also help with implementing indentation detection when file is open or check on demand, similarly how it is used in VScode and geany.

pryrt:
> * Switch the active tab to the other .cpp file -- status bar still shows `Spaces: 2` when, I expect to see `Tabs: 4` (and it isn't a display glitch, all my .cpp file tabs are now space-2 indented -- I don't want or expect this).

Even worse: assume I start with **default** = `Spaces: 4`, **normal** = `use default value`, and **makefile** = `Tabs: 4` (and thus to _not_ use default value).  Then while I am editing my makefile, if I change in the status-bar-menu to `Tabs: 8` for my makefile, then create a new plaintext file (_not_ makefile), the plaintext sees the 8: if I check the settings, I can confirm that both **default** and **makefile** are now set to 8, even though I only changed it for the **makefile**, and **makefile** was not set to match the **default**.

So even if you are willing to accept the limitation of it affecting all files of a specific type rather than just the current file (and like the OP and Alan, I really think that per-file would be useful), this implementation does not correctly maintain the separation between filetypes that are _not_ supposed to follow the default setting by changing both that filetype's setting _and_ the default setting when a status-bar change is made for the specific filetype.

ozone10:
@pryrt 
there was issue with implementation, it should work now.

I will put this PR on hold, until per file setting is implemented

alankilborn:
>I would like to implement it so indent can be per file, meaning to have 3 option: global default, per language, and per file.

IMO, changes to a set value from the status bar should affect ONLY the current file; please don't doodle around with setting "global" or "per file type" stuff via this mechanism -- if user needs that, then they need to go to the _Preferences_ to do it (it's certainly fine to have a popup menu item to take them right to the correct area of the prefs if invoked).

>I usually work with files with indent size 2 or 4, and know that some source files from unix use 8, and also based on request, I added only those 3 sizes. Adding size 3 is trivial.

Yes, but do you add 1, 3, 5, 7, 9, etc as discrete choices?  I'd say the current 2, 4, and 8 are fine, but maybe add a "Input custom size" menu item that opens a integer prompt.  It isn't good for user to go to Prefs to set e.g. 3, because maybe they want 3 ONLY for the file currently active (I have seen this in practice...where I want/need to do exactly this).

ozone10:
@alankilborn 
For setting lets wait until per file is implemented.
For custom size, prompt is good idea, it can be implemented separately, this "basic" PR is already quite complex.

Just note anyone is free to do implementation for per file setting.
I will be away for two weeks and also I need to finish rest of hiDPI stuffs.

sXNA:
If tab indicator in status bar is problem, can at least indentation submenu in setting be accepted

donho:
@alankilborn 
>The creator of the issue said:

>>    sometimes I am working with (more) than one file and they can have different indentation styles

> I tested this PR and it does not support that:

Regarding the detection: it's hard to make the detection persist: we need a thread for monitoring every key-in from the users. Even with a tread, it's not easy to be accurate. So I don't think it'll be in Notepad++.


@ozone10 

Though I don't like the added submenu "Identation" of the main menu "Settings", it's well implemented - at least it's coherent with the existent settings.

Regarding the behaviour, however, one point bothers me is what @alankilborn  has described below:

> Switch the active tab to the other .cpp file -- status bar still shows Spaces: 2 when, I expect to see Tabs: 4 (and it isn't a display glitch, all my .cpp file tabs are now space-2 indented -- I don't want or expect this).

Indeed, except the power users, it's confusing without choosing the programming language explicitly.   


> I would like to implement it so indent can be per file, meaning to have 3 option: global default, per language, and per file.
> This will also help with implementing indentation detection when file is open or check on demand, similarly how it is used in VScode and geany.

Concerning indentation detection, as I said above, the detection cannot  (or is hard to ) be accurate, and it needs to following the key-in of the users: The file could be detected as Tab, then user could type 4 spaces so the indent of document should turn into "Mixed". Under the Mixed mode, what should happen when user hits the Tab key?
Also, I've tested the feature in Visual Studio, the detection is far from accurate.

Even indentation detection works well, and we do settings per file, what's the purpose of per language and global default then?
It's a complex issue, and needs a good definition before the implementation IMO.

ozone10:
close for now until I get more time and finish DPI

PR #15230, opened 2024-06-01
Implement installer '/closeRunningNpp' cmdline option

xomx:
Partially fix the #8514 .
The promised followup of the #14251.

If specified on the NSIS installer cmdline:
- it detects a possible running Notepad++ (by FindWindow WINAPI on the Notepad++ registered wnd-class)
- then it first tries to use the usual app-closing by sending the WM_CLOSE message to the running Notepad++ wnd found (it waits for that closing 5 secs max using the WaitForSingleObject WINAPI)
- if that standard closing fails, it uses consequently the forceful TerminateProcess WINAPI way (one can easily simulate this e.g. by suspending a running N++ process via an advanced task manager...)
- it also handles the possible Notepad++ multi-instance mode (it will close all the running instances)

(if accepted, the N++ manual will also need a PR about this new switch)

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -114,6 +114,7 @@ InstType "Minimalist"
 
 Var diffArchDir2Remove
 Var noUpdater
+Var closeRunningNpp
 
 
 !ifdef ARCH64 || ARCHARM64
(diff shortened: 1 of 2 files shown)


donho:
@xomx 
From the description of PR title, if the PR is accepted, in the future the following command can be launched:
`npp.8.6.9.Installer.x64.exe /closeRunningNpp /S`
So the installer will close all instances of Notepad++, then install version 8.6.9 silently, and launches the new version when the installation is done.

Do you confirm above description?

xomx:
@donho 

Yes.

Of course, this can also be used without that "/S" switch, but the main purpose of this PR is to support automated installation services like winget that uses the N++ installer in its silent mode.

donho:
@xomx 
Tested in any scene I could imagine, it looks robust. Thank you!

> they (MS WinGet) are watching our efforts and accommodating to it.

Ah, I was not aware of it. It's rather a good news, isn't it?
I mean, users have more choice to update/install Notepad++ (though I've never used WinGet).

xomx:
@donho 
> I mean, users have more choice to update/install Notepad++ (though I've never used WinGet).

IMO - this has the potential to replace the strongly required MSI installer for sysadmins and partially also the publication via the MS Store. It could also satisfy all the users requirements for seamless silent auto-update.

PR #15236, opened 2024-06-03
Fix for the "lost" panels problem

### LINKED ISSUE #13084: [Feature request] Solution to the "lost panel" problem

xomx:
Fix #13084 .
https://community.notepad-plus-plus.org/topic/23344/

Implement some size limits for the Notepad++ docked and floating panels, as the users often "lost" its panels (e.g. when they minimized them to zero width or height).

This also fixes the config.xml possible previous (or future) "out-of-the-allowed-range" data.

Fix in action:
npp-lost-panels-fix

Some videos of the problem in the issue: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13084#issuecomment-1775662467

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -799,7 +799,7 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	dmd._minDockedPanelVisibility = DPIManagerV2::scale(nppGUI._dockingData._minDockedPanelVisibility, dpi);
 	dmd._minFloatingPanelSize.cy = nppGUI._dockingData._minDockedPanelVisibility;
 	dmd._minFloatingPanelSize.cx = std::max(static_cast<int>(nppGUI._dockingData._minFloatingPanelSize.cy * 6),
-		::GetSystemMetrics(SM_CXMINTRACK));
+		DPIManagerV2::getSystemMetricsForDpi(SM_CXMINTRACK, dpi));
 
 	_dockingManager.setDockedContSize(CONT_LEFT, nppGUI._dockingData._leftWidth);
 	_dockingManager.setDockedContSize(CONT_RIGHT, nppGUI._dockingData._rightWidth);
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -798,7 +798,7 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	_dockingManager.setDockedContSize(CONT_LEFT, nppGUI._dockingData._leftWidth);
 	_dockingManager.setDockedContSize(CONT_RIGHT, nppGUI._dockingData._rightWidth);
 	_dockingManager.setDockedContSize(CONT_TOP, nppGUI._dockingData._topHeight);
-	_dockingManager.setDockedContSize(CONT_BOTTOM, nppGUI._dockingData._bottomHight);
+	_dockingManager.setDockedContSize(CONT_BOTTOM, nppGUI._dockingData._bottomHeight);
 
 	{
 		for (size_t i = 0, len = dmd._pluginDockInfo.size(); i < len; ++i)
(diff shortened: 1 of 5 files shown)


donho:
@xomx 
The amount of the fix code is quite important... 
Could you provide the instruction to reproduce the bug in order to validate your PR?

PowerEditor/src/Parameters.cpp
```diff
@@ -1012,6 +1013,16 @@ NppParameters::NppParameters()
 	//Load initial accelerator key definitions
 	initMenuKeys();
 	initScintillaKeys();
+
+	// Rescale default panels dimensions according to the current DPI
+	_nppGUI._dockingData._leftWidth = _dpiManager.scaleX(_nppGUI._dockingData._leftWidth);
```
donho, line 1018: @ozone10 Please correct me if I say somethings wrong.

@xomx 
`DPIManager` is deprecated and it will be removed in the future. Please use `DPIManagerV2` instead.

And the value of docking data should not be modified here, but when it is used IMO.

For example, instead of adding the above line, please rather use the following line:
`_dockingManager.setDockedContSize(CONT_LEFT, DPIManagerV2::scale(nppGUI._dockingData._leftWidth, dpi));`
instead of https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Notepad_plus.cpp#L798

donho (changes requested):
Since the member `_dpiManager` will be removed from Parameter in short/middle term, please consider scaling _dockingData where they are used.

PowerEditor/src/Parameters.cpp
ozone10: @donho @xomx 
when using `DPIManagerV2` panels sizes need to be saved unscaled. And when loaded they need to be scaled according to dpi on screen they currently are. And because of this multi-screen situation, especially for floating panels, some code from `Parameters.*` needs to be moved to "docking" files.

But since this will require a lot of work, simple "in place" replacing old `DPIManager` with `DPIManagerV2` with correct dpi should be enough for current dpi manifest.

xomx: @donho @ozone10 
Thanks for your hints.

To not further complicate this PR, I will leave the mentioned right Docking Manager data handling to @ozone10 (storing only unscaled data, otherwise it will slowly grow with every N++ relaunch without a user-initiated dimension change, right?). Inherent data needed for this PR are not stored to the config.xml via Parameters.cpp handling anyway, the base for the minimum panel sizes calculations is the DockingCont.h constant HIGH_CAPTION.

So now this PR uses the requested DPIManagerV2 and the initial scale adjustment was moved from the NppParameters ctor to the suggested Notepad_plus::init.

At runtime, watching for the possible scaling changes has been already added by me to the DockingManager::runProc -> case DMM_MOVE_SPLITTER.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -795,10 +795,16 @@ LRESULT Notepad_plus::init(HWND hwnd)
 
 	DockingManagerData& dmd = nppGUI._dockingData;
 
+	// preset minimal panel dimensions according to the current DPI
+	dmd._minDockedPanelVisibility = DPIManagerV2::scale(nppGUI._dockingData._minDockedPanelVisibility, dpi);
+	dmd._minFloatingPanelSize.cy = nppGUI._dockingData._minDockedPanelVisibility;
+	dmd._minFloatingPanelSize.cx = std::max(static_cast<int>(nppGUI._dockingData._minFloatingPanelSize.cy * 6),
+		::GetSystemMetrics(SM_CXMINTRACK));
```
ozone10, line 802: @xomx 
could you replace `GetSystemMetrics` with https://github.com/notepad-plus-plus/notepad-plus-plus/blob/e7e88a380ac01f8def627f1672f0192461603d6b/PowerEditor/src/dpiManagerV2.h#L49

xomx: @ozone10 
Done.

PR #15243, opened 2024-06-05
Add message box with information about disabled backward regex searching

### LINKED ISSUE #14647: Shift F3 search backward does not work...

1. 
2. 
3.

### LINKED ISSUE #15239: [BUG] Find Previous (Shift + F3) doesn't work in the last update 8.6.7

### Description of the Issue

Find Previous (Shift + F3) doesn't work in the last update 8.6.7

### Steps To Reproduce

1. Search a word (with F3), it finds it properly. 
2. Try do go to the previous find, and it doesn't do anything.
3. Find Previous (Shift + F3) doesn't work in the last update 8.6.7. 
    - It doesn't work also from Search menu

* In find Menu, backward option is disabled

### Current Behavior

Notepad++ doesn't do anything when pressing Shift + F3 or Find Previous from menu.

### Expected Behavior

Shift + F3 or Find Previous should work (similar to F3)

### Debug Information

```shell
Notepad++ v8.6.7   (64-bit)
Command Line : "C:\Program Files\Notepad++\change.log" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4412
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

ozone10:
close #15239
close #14647

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -601,6 +601,12 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	//Search menu
 	//disable "Search Results Window" under Search Menu 
 	::EnableMenuItem(_mainMenuHandle, IDM_FOCUS_ON_FOUND_RESULTS, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_GOTONEXTFOUND, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_GOTOPREVFOUND, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
+
+	// Find dialog is not created yet
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_FINDNEXT, MF_BYCOMMAND | MF_DISABLED | MF_GRAYED);
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_FINDPREV, MF_BYCOMMAND | MF_DISABLED | MF_GRAYED);
 
 	//Main menu is loaded, now load editor context menu items
 	nppParam.getContextMenuFromXmlTree(_mainMenuHandle, _pluginsManager.getMenuHandle());
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1881,6 +1881,8 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 						// this can only happen when shift-key was pressed
 						// regex upward search is disabled
 						// turn user action into a no-action step
+
+						regexBackwardMsgBox();
 					}
 					else
 					{
@@ -3754,6 +3756,38 @@ Finder* FindReplaceDlg::getFinderFrom(HWND hwnd)
 	return nullptr;
 }
 
+int FindReplaceDlg::regexBackwardMsgBox()
+{
+	NppParameters& nppParam = NppParameters::getInstance();
+
+	const int msgboxID = nppParam.getNativeLangSpeaker()->messageBox("FindRegexBackwardDisabled",
+		(*_ppEditView)->getHParent(),
+		L"By default, backward regex searching is disabled due to potentially unexpected results. " \
+		L"To perform a backward search, open the Find dialog and select either normal or extended search mode instead of regular expression.\r\n" \
+		L"Press the OK button to open the Find dialog or set focus on it.\r\n" \
+		L"\r\n" \
+		L"If you require the backward regex searching feature, consult the user manual for instructions on enabling it.",
+		L"Regex backward search disabled",
+		MB_OKCANCEL | MB_APPLMODAL | MB_ICONINFORMATION);
+
+	switch (msgboxID)
+	{
+		case IDOK:
+		{
+			doDialog(FIND_DLG, nppParam.getNativeLangSpeaker()->isRTL());
+			goToCenter();
+			::SetFocus(::GetDlgItem(_hSelf, IDREGEXP));
+			break;
+		}
+
+		case IDCANCEL:
+		default:
+			break;
+	}
+
+	return msgboxID;
+}
+
 void FindReplaceDlg::setSearchText(TCHAR * txt2find)
 {
 	HWND hCombo = ::GetDlgItem(_hSelf, IDFINDWHAT);
@@ -4131,6 +4165,8 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 							// regex upward search is disabled
 							// this macro step could have been recorded in an earlier version before it was not allowed, or hand-edited
 							// make this a no-action macro step
+
+							regexBackwardMsgBox();
 						}
 						else
 						{
@@ -4157,6 +4193,8 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 							// regex upward search is disabled
 							// this macro step could have been recorded in an earlier version before it was not allowed, or hand-edited
 							// make this a no-action macro step
+
+							regexBackwardMsgBox();
 						}
 						else
 						{
@@ -4173,6 +4211,8 @@ void FindReplaceDlg::execSavedCommand(int cmd, uptr_t intValue, const generic_st
 							// regex upward search is disabled
 							// this macro step could have been recorded in an earlier version before it was disabled, or hand-edited
 							// make this a no-action macro step
+
+							regexBackwardMsgBox();
 						}
 						else
 						{
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -2472,6 +2472,9 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 					}
 
 					bool isRegex = (_options._searchType == FindRegex);
+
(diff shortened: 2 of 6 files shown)


ozone10:
Instead of silently doing nothing with regex backward search it's better to inform user why backward search does not nothing.

alankilborn:
IMO the problem isn't silently-doing-nothing, it's that _Find previous_ is not disabled in the main menu (_Search_).

ozone10:
@alankilborn 
> IMO the problem isn't silently-doing-nothing, it's that _Find previous_ is not disabled in the main menu (_Search_).

Okay, menu is disabled -> shortcut is also disabled -> user will fill issue that `Find previous` is disabled.
Similar to disabled "down triangle" in find dialog, where shortcut still works.

But if this is what you think is OK, here is commit.

xomx:
@ozone10 
I can reproduce your crash STR:
npp-Search-NextSearchResult-crash

@donho @alankilborn 
I tried the original and updated PR with disabled menu and have ambivalent feeling. I agree that having the relevant menu cmd disabled is consistent with the Find dlg behavior, but the original PR is better from the user's POV (they get the right grasp what is going on and what to do to correct it ...).

alankilborn:
> Okay, menu is disabled -> shortcut is also disabled -> user will fill issue that `Find previous` is disabled. Similar to disabled "down triangle" in find dialog, where shortcut still works.
> 
> But if this is what you think is OK

These are just MY feelings, as a long-time user and new-to-Npp developer.  :-)

I just tend to not like "chatty" software.  Sure, there are places where a message box is really important, but I don't feel like this is one of them.  Users should "accept" the disabled condition, RTFM, or ask.  Of course, it is preferred that the asking is on the Community and not in the form of an "issue", but...

alankilborn:
> btw
> does Notepad++ crash with these steps

I can replicate the crash as well.  :-(

Maybe it is time to prevent a _Find All in Current Document_ search when the _Find what_ text is empty?

Unfortunately (for me), I use this quirk in behavior so that I can easily get the _Search results_ created on (scripting) startup, so that I can get an hwnd to the window (for further scripting).
I suppose that I can change my automated search to look for `\A` in regex mode instead.  :-)

Regardless, shouldn't the steps [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15243#issuecomment-2151494889) be made into a separate issue?

donho:
@ozone10 
> btw does Notepad++ crash with these steps
> 
>     1. Open empty document (e.g. New 1)
> 
>     2. Open Find dialog (ctrl+f)
> 
>     3. Press "Find All in Current Document" button -> Search results with 0 hits
> 
>     4. Open menu Search->Next Search Result

I can reproduce it too.
The concerning issue #15247 is created and assigned to me. If there's no PR dealing with the issue, I'll work on it when I have time.

ozone10:
@alankilborn
> Regardless, shouldn't the steps [HERE](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15243#issuecomment-2151494889) be made into a separate issue?

I did not have time, and I wanted to confirm my findings.

> I just tend to not like "chatty" software. Sure, there are places where a message box is really important, but I don't feel like this is one of them. Users should "accept" the disabled condition, RTFM, or ask. Of course, it is preferred that the asking is on the Community and not in the form of an "issue", but...

I am mainly concerned for users who use only shortcuts, so they won't notice that menu item is disabled. They press shift+f3 and nothing happen -> ok there is no other hits.

@donho 
also if you get search results with some hits and then delete search results with delete key it will also cause crash
e.g.
1. Open `stylers.xml`
2. Select `WordsStyle` and press ctrl+f
3. Press "Find All in Current Document" button -> Search results with hits for `WordsStyle`
4. Now delete 2nd line -> search results is empty except 1st line with Search "WordsStyle" ...
5. Open menu Search->Next Search Result

donho:
@xomx I see your point.

@ozone10 
> I am mainly concerned for users who use only shortcuts, so they won't notice that menu item is disabled. They press shift+f3 and nothing happen -> ok there is no other hits.

Indeed. In this case Find Replace dialog is often closed. So let's not block commands & shortcuts so the users will get the information. But in Find Replace dialog the concerning operation keep being blocked.

PowerEditor/src/Notepad_plus.cpp
```diff
@@ -601,6 +601,12 @@ LRESULT Notepad_plus::init(HWND hwnd)
 	//Search menu
 	//disable "Search Results Window" under Search Menu 
 	::EnableMenuItem(_mainMenuHandle, IDM_FOCUS_ON_FOUND_RESULTS, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_GOTONEXTFOUND, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_GOTOPREVFOUND, MF_DISABLED | MF_GRAYED | MF_BYCOMMAND);
+
+	// Find dialog is not created yet
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_FINDNEXT, MF_BYCOMMAND | MF_DISABLED | MF_GRAYED);
+	::EnableMenuItem(_mainMenuHandle, IDM_SEARCH_FINDPREV, MF_BYCOMMAND | MF_DISABLED | MF_GRAYED);
```
donho, line 609: Can be removed now.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3590,6 +3595,8 @@ void FindReplaceDlg::findAllIn(InWhat op)
 
 		//enable "Search Results Window" under Search Menu
 		::EnableMenuItem(::GetMenu(_hParent), IDM_FOCUS_ON_FOUND_RESULTS, MF_ENABLED | MF_BYCOMMAND);
+		::EnableMenuItem(::GetMenu(_hParent), IDM_SEARCH_GOTONEXTFOUND, MF_ENABLED | MF_BYCOMMAND);
+		::EnableMenuItem(::GetMenu(_hParent), IDM_SEARCH_GOTOPREVFOUND, MF_ENABLED | MF_BYCOMMAND);
```
donho, line 3599: Can be removed now.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -4517,6 +4562,12 @@ void FindReplaceDlg::doDialog(DIALOG_TYPE whichType, bool isRTL, bool toShow)
 	{
 		_isRTL = isRTL;
 		create(IDD_FIND_REPLACE_DLG, isRTL, true, toShow);
+
+		::EnableMenuItem(::GetMenu(_hParent), IDM_SEARCH_FINDNEXT, MF_BYCOMMAND | MF_ENABLED);
+
+		NppParameters& nppParam = NppParameters::getInstance();
+		const bool enableBackwardSearch = (nppParam.getFindHistory()._searchMode != FindHistory::regExpr) || nppParam.regexBackward4PowerUser();
+		::EnableMenuItem(::GetMenu(_hParent), IDM_SEARCH_FINDPREV, MF_BYCOMMAND | (enableBackwardSearch ? MF_ENABLED : MF_DISABLED | MF_GRAYED));
```
donho: Can be removed now.

ozone10:
@donho 
> Indeed. In this case Find Replace dialog is often closed.
> So let's not block commands & shortcuts so the users will get the information. But in Find Replace dialog the concerning operation keep being blocked.

Done.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -3754,6 +3761,38 @@ Finder* FindReplaceDlg::getFinderFrom(HWND hwnd)
...
+{
+	NppParameters& nppParam = NppParameters::getInstance();
+
+	const int msgboxID = nppParam.getNativeLangSpeaker()->messageBox("FindRegexBackwardDisabled",
+		(*_ppEditView)->getHParent(),
+		L"By default, backward regex searching is disabled due to potentially unexpected results. " \
+		L"To perform a backward search, open the Find dialog and select either normal or extended search mode instead of regular expression.\r\n" \
+		L"Press the OK button to open the Find dialog or set focus on it.\r\n" \
+		L"\r\n" \
+		L"If you require the backward regex searching feature, consult the user manual for instructions on enabling it.",
```
donho, line 3774: I'm thinking of providing the following link:
https://npp-user-manual.org/docs/searching/#regular-expressions

I'll see what I can do about it, after merging the PR.

donho (changes requested):
As agreed in my post, please remove the blocking commands so users can get the information.

donho (approved):

PowerEditor/src/Notepad_plus.cpp
ozone10: It does nothing if find dialog is not created
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ea71d5ea4e2116fa30d1ab0dda6b1cb2c160b8de/PowerEditor/src/NppCommands.cpp#L1362

And It makes no sense that IDM_SEARCH_GOTONEXTFOUND and  DM_SEARCH_GOTOPREVFOUND are enabled if search results panel was not created.

jedomed:
> Sure, there are places where a message box is really important, but I don't feel like this is one of them. Users should "accept" the disabled condition, RTFM, or ask.

I got here after wondering why the button was disabled. IMO there should be *some* in-app hint to why this happens, even just a tooltip when you hover over the button would be great. I don't think a user should have to go to an online forum to understand why a functionality this commonly used suddenly stopped working for them.

> RTFM

https://github.com/notepad-plus-plus/npp-usermanual/blob/master/content/docs/searching.md

I didn't read the entire page (look at the lenght of it), but I couldn't find anything by Ctrl+Fing "find previous", "disabled", "search mode". Either it's not there at all, or it's well hidden :-(

There IS a mention of "Match whole" being disabled, "Find previous" could be mentioned here as well:

"Selecting Search Mode of Regular expression will cause the Match whole word only option to become unchecked and disabled (greyed out)."

PR #15245, opened 2024-06-06
Fix short checkbox 'Indent Settings' for some languages.

GOLEM777:


donho:
I will take care of it.

PR #15278, opened 2024-06-13
Fix selection mode after EndSelect

### LINKED ISSUE #15221: [BUG] Begin/End Select Problem

### Description of the Issue

After using Begin/End Select to start a selection and another Begin/End Select to end the selection, and deleting the selection, pressing cursor keys to move causes a continued selection.

### Steps To Reproduce

1.  Ctrl-Shift-B
2.  Select some text
3.  Ctrl-Shift-B
4. Press Delete to delete it
5. Press a cursor key and see a selection get expanded instead of just moving the cursor.

### Current Behavior

Trying to move the cursor with the cursor keys after using Begin/End Select is selecting text along the cursor's path.

### Expected Behavior

I expect that once the selection is ended that any cursor movements would simply be cursor movements.

### Debug Information

```shell
Notepad++ v8.6.7   (64-bit)
Command Line : "C:\Users\j.ritter\AppData\Roaming\Notepad++\shortcuts.xml" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4412
Current ANSI codepage : 1252
Plugins : 
    HexEditor (0.9.12)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    NppFTP (0.29.12)
```

### Anything else?

I have v7.3.3 on another machine and it acts as expected with a Begin/End Select sequence and normal cursor movements afterward.

alankilborn:
Fix #15221

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2786,14 +2786,18 @@ void ScintillaEditView::addText(size_t length, const char *buf)
 
 void ScintillaEditView::beginOrEndSelect(bool isColumnMode)
 {
+	auto currPos = execute(SCI_GETCURRENTPOS);
+
 	if (_beginSelectPosition == -1)
 	{
-		_beginSelectPosition = execute(SCI_GETCURRENTPOS);
+		_beginSelectPosition = currPos;
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


donho (changes requested):
In general, please use 1 line instruction.

donho (approved):

PR #15280, opened 2024-06-13
Make install/update easy & quiet

donho:
Add command line parameter "/runNppAfterSilentInstall" for Notepad++ installer. This parameter launches installed Notepad++ after silent installation. Of course, the parameter takes effect only when "/S" (silent installation) is used.

Fix partially #8514, follow up #15230

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -90,6 +90,7 @@ page Custom ExtraOptions
 !define MUI_FINISHPAGE_RUN_FUNCTION "LaunchNpp"
 !insertmacro MUI_PAGE_FINISH
 
+
 !insertmacro MUI_UNPAGE_CONFIRM
 !define MUI_PAGE_CUSTOMFUNCTION_SHOW "un.CheckIfRunning"
 !insertmacro MUI_UNPAGE_INSTFILES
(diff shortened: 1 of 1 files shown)


xstaindx:
If I was to deploy this version of your software with automation tools that use SYSTEM for the install with these new comamndl ines, would these launch the software in the SYSTEM context after install in Session 0 or would it be launched into the logged on user in Session 1?  I would think session 1 would be bad.

donho:
No idea about concept of session 0 or of session 1.
It's launched under the accound of user.

xstaindx:
I would strongly suggest that you do not allow a silent install to launch in the context of the user who installed it if the installation is SYSTEM. (like through tools such as SCCM, BIGFIX, Tanium or others)  You end up leaving an open session that a user cannot access that has SYSTEM rights, nor can user see or interact with.

xstaindx:
> No idea about concept of session 0 or of session 1. It's launched under the accound of user.

https://techcommunity.microsoft.com/t5/ask-the-performance-team/sessions-desktops-and-windows-stations/ba-p/372473

schenardie:
Is there a way that /runNppAfterSilentInstall will only run if /closeRunningNpp is also called? I would like to automatically re-open npp only if the installation closed it, but if it was already closed it should not open it.

xomx:
@schenardie 

> I would like to automatically re-open npp only if the installation closed it, but if it was already closed it should not open it.

In the upcoming N++ version, there will be new installer param `/relaunchNppAfterSilentInstall` for it (for details check the current PR [#16626](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/16626#discussion_r2155473082) which added option to auto-update N++ also at its exit)

PR #15286, opened 2024-06-13
Add dark mode search results panel tab bar icon

ozone10:
ref #14959


ozone10:
ico is placeholder, waiting for @WildByDesign

PR #15289, opened 2024-06-14
Make panel tabbar icons using BMP use transparency

ozone10:
- modify standard ico files to use transparency

ref #14959


ozone10:
@donho 
there is small regression with https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f2fbeadfea670b0001f3b44e3156fd5022daf71a and https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f37f2d0ffeb5e75aea6d54806b06b0f0e112aa87

`LoadIconWithScaleDown` scale better with for dpi, but has worse support for transparency. Icon needs to be already transparent. Unlike `LoadImage` with flags `LR_LOADMAP3DCOLORS | LR_LOADTRANSPARENT` which allows to fake transparency.

So I have modified standard icons to have "true" transparency.

Similar to`LoadIconWithScaleDown`, `DrawIconEx` also needs that loaded `HICON` already has transparency,
Because this mainly affect plugins, ImageList is used to get modified `HICON` with fake transparency.

donho:
@ozone10 
Really, except few ico (for example: filebrowser.ico), I cannot see the regression in the other icons.
But since the icons' size are reduced, it's good that you change them.

BTW, which software do you use for modifying the icons?

ozone10:
@donho 
>Really, except few ico (for example: filebrowser.ico), I cannot see the regression in the other icons.
>But since the icons' size are reduced, it's good that you change them.

For some of them it's more noticeable, when switching from light mode to dark mode. I also enlarged some of them to size 16x16.

 >BTW, which software do you use for modifying the icons?

Simple MS Paint for modification and saving them as PNG. Then usually python script to save them as ICO with PNG.
Or for quick modification https://redketchup.io/.

WildByDesign:
@ozone10 @donho For PNG to ICO, there is also a really nice CLI program (https://github.com/J-Siu/go-png2ico) that does the correct format. It can be used by itself from the command line quite simply, but also I have it scripted with PowerShell to create arrays of all of the PNG file names to make everything automated when doing many icons.

donho:
@ozone10 
> Simple MS Paint for modification

LOL! I use it all the time, and I bet a lot of devs use it instead of PhotoShop :)

> Or for quick modification https://redketchup.io/

I will try it in the future. Thank you!

@WildByDesign 
> For PNG to ICO, there is also a really nice CLI program (https://github.com/J-Siu/go-png2ico) that does the correct format.

Downloaded, thank you!

ozone10:
> LOL! I use it all the time, and I bet a lot of devs use it instead of PhotoShop :)

MS even added new features to it, e.g. layers, remove background, ...

PR #15291, opened 2024-06-15
Add plugin method to rename untitled tab

### LINKED ISSUE #8916: Add plugin interface for renaming an unsaved tab
Notepad++ version 7.6.4 (I think) introduced the capability to rename an unsaved tab.

For example, a tab with the title "new 2" can be right-clicked upon, "Rename" chosen from the resulting popup menu, and the following box presents:

image

What is needed is a way for a plugin to be able to effect this sort of rename, directly, without the popup input box shown above.

[ Comments above based upon features available current to Notepad++ 7.9 ]

alankilborn:
Fix #8916

### DIFF
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -190,7 +190,7 @@ public:
 	bool fileSaveAs(BufferID id = BUFFER_INVALID, bool isSaveCopy = false);
 	bool fileDelete(BufferID id = BUFFER_INVALID);
 	bool fileRename(BufferID id = BUFFER_INVALID);
-	bool fileRenameUntitled(BufferID id = BUFFER_INVALID, const wchar_t* tabNewName = nullptr);
+	bool fileRenameUntitled(BufferID id, const wchar_t* tabNewName);
 
 	bool switchToFile(BufferID buffer);			//find buffer in active view then in other view.
 	//@}
--- a/PowerEditor/src/Notepad_plus.h
+++ b/PowerEditor/src/Notepad_plus.h
@@ -190,6 +190,7 @@ public:
 	bool fileSaveAs(BufferID id = BUFFER_INVALID, bool isSaveCopy = false);
 	bool fileDelete(BufferID id = BUFFER_INVALID);
 	bool fileRename(BufferID id = BUFFER_INVALID);
+	bool fileRenameUntitled(BufferID id = BUFFER_INVALID, const wchar_t* tabNewName = nullptr);
 
 	bool switchToFile(BufferID buffer);			//find buffer in active view then in other view.
 	//@}
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -553,20 +553,6 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			break;
 		}
 
-		case NPPM_INTERNAL_SETFILENAME:
-		{
-			if (!lParam && !wParam)
-				return FALSE;
-			BufferID id = (BufferID)wParam;
-			Buffer * b = MainFileManager.getBufferByID(id);
-			if (b && b->getStatus() == DOC_UNNAMED)
-			{
-				b->setFileName(reinterpret_cast<const TCHAR*>(lParam));
-				return TRUE;
-			}
-			return FALSE;
-		}
-
 		case NPPM_GETBUFFERLANGTYPE:
 		{
 			if (!wParam)
@@ -3140,6 +3126,11 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			}
 			
 			return colorId;
+		}
+
+		case NPPM_SETUNTITLEDNAME:
+		{
+			return fileRenameUntitled(reinterpret_cast<BufferID>(wParam), reinterpret_cast<const wchar_t*>(lParam));
 		}
 
 		case NPPM_GETBOOKMARKID:
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -3130,7 +3130,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_SETUNTITLEDNAME:
 		{
-			if (!lParam) return false;
+			if (!wParam || !lParam) return false;
 			return fileRenameUntitled(reinterpret_cast<BufferID>(wParam), reinterpret_cast<const wchar_t*>(lParam));
 		}
 
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -3130,6 +3130,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_SETUNTITLEDNAME:
 		{
+			if (!lParam) return false;
 			return fileRenameUntitled(reinterpret_cast<BufferID>(wParam), reinterpret_cast<const wchar_t*>(lParam));
 		}
 
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -3130,7 +3130,7 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_SETUNTITLEDNAME:
 		{
-			if (!wParam || !lParam) return false;
+			if (!wParam || !lParam) return FALSE;
 			return fileRenameUntitled(reinterpret_cast<BufferID>(wParam), reinterpret_cast<const wchar_t*>(lParam));
 		}
 
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1969,6 +1969,86 @@ bool Notepad_plus::fileRename(BufferID id)
 	return success;
 }
 
+bool Notepad_plus::fileRenameUntitled(BufferID id /* =BUFFER_INVALID */, const wchar_t* tabNewName /* =nullptr */)
+{
+	BufferID bufferID = id;
+	if (id == BUFFER_INVALID)
+	{
+		bufferID = _pEditView->getCurrentBufferID();
+	}
+	Buffer* buf = MainFileManager.getBufferByID(bufferID);
+
+	SCNotification scnN{};
+	scnN.nmhdr.code = NPPN_FILEBEFORERENAME;
+	scnN.nmhdr.hwndFrom = _pPublicInterface->getHSelf();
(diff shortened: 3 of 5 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -1969,6 +1969,86 @@ bool Notepad_plus::fileRename(BufferID id)
...
+			trim(tabNewNameStr); // No leading and tailing space allowed
+
+			if (!tabNewNameStr.empty())
+			{
+				if (tabNewNameStr.length() < langNameLenMax)
+				{
+					// https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
+					// Reserved characters:  < > : " / \ | ? * tab  
+					//  ("tab" is not in the official list, but it is good to avoid it)
+					const std::wstring reservedChars = TEXT("<>:\"/\\|\?*\t");
```
donho, line 2008: Use the same RegExpr here:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppIO.cpp#L1906

Please define a symbol and use it in the both places.

PowerEditor/src/NppIO.cpp
```diff
@@ -1969,6 +1969,86 @@ bool Notepad_plus::fileRename(BufferID id)
...
+
+								// Create new backup
+								buf->setModifiedStatus(true);
+								bool bRes = MainFileManager.backupCurrentBuffer();
+
+								// Delete old backup
+								if (bRes)
+								{
+									::DeleteFile(oldBackUpFile.c_str());
+								}
```
donho, line 2038: 8 levels of depth is too much IMO...

PowerEditor/src/NppIO.cpp
```diff
@@ -1969,6 +1969,86 @@ bool Notepad_plus::fileRename(BufferID id)
...
+				}
+			}
+		}
+	}
+	
+	scnN.nmhdr.code = success ? NPPN_FILERENAMED : NPPN_FILERENAMECANCEL;
+	_pluginsManager.notify(&scnN);
+
+	return success;
+}
```
donho, line 2051: ... in order to improve the readability, I suggest to code in the following way:
```cpp


bool Notepad_plus::fileRenameUntitled(BufferID id, const wchar_t* tabNewName)
{
	BufferID bufferID = id;
	if (id == BUFFER_INVALID)
	{
		bufferID = _pEditView->getCurrentBufferID();
	}
	Buffer* buf = MainFileManager.getBufferByID(bufferID);

	bool isFileExisting = PathFileExists(buf->getFullPathName()) != FALSE;
	if (isFileExisting) return false;
	
	// We are just going to rename the tab nothing else
	// So just rename the tab and rename the backup file too if applicable

	if (!tabNewName) return false;
		
        std::wstring tabNewNameStr = tabNewName;

        trim(tabNewNameStr); // No leading and tailing space allowed

        if (tabNewNameStr.empty()) return false;

	if (tabNewNameStr.length() >= langNameLenMax) return false;
				
        // https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
        // Reserved characters:  < > : " / \ | ? * tab  
        //  ("tab" is not in the official list, but it is good to avoid it)
        const std::wstring reservedChars = TEXT("<>:\"/\\|\?*\t");

        if (tabNewNameStr.find_first_of(reservedChars) != std::wstring::npos) return false; // no reserved characters in tabNewNameStr

        BufferID sameNamedBufferId = _pDocTab->findBufferByName(tabNewNameStr.c_str());
        if (sameNamedBufferId == BUFFER_INVALID)
        {
                sameNamedBufferId = _pNonDocTab->findBufferByName(tabNewNameStr.c_str());
        }

        if (sameNamedBufferId == BUFFER_INVALID)
        {
                SCNotification scnN{};
                scnN.nmhdr.code = NPPN_FILEBEFORERENAME;
                scnN.nmhdr.hwndFrom = _pPublicInterface->getHSelf();
                scnN.nmhdr.idFrom = (uptr_t)bufferID;
                
                buf->setFileName(tabNewNameStr.c_str());
                
                scnN.nmhdr.code = NPPN_FILERENAMED;
                
                
                bool isSnapshotMode = NppParameters::getInstance().getNppGUI().isSnapshotMode();
                if (isSnapshotMode)
                {
                        std::wstring oldBackUpFile = buf->getBackupFileName();

                        // Change the backup file name and let MainFileManager decide the new filename
                        buf->setBackupFileName(TEXT(""));

                        // Create new backup
                        buf->setModifiedStatus(true);
                        bool bRes = MainFileManager.backupCurrentBuffer();

                        // Delete old backup
                        if (bRes)
                        {
                                ::DeleteFile(oldBackUpFile.c_str());
                        }
                }
        }

	return true;
}
```

donho (changes requested):
Please following the suggestions.

PowerEditor/src/NppIO.cpp
```diff
@@ -1969,6 +1969,86 @@ bool Notepad_plus::fileRename(BufferID id)
...
+	{
+		bufferID = _pEditView->getCurrentBufferID();
+	}
+	Buffer* buf = MainFileManager.getBufferByID(bufferID);
+
+	SCNotification scnN{};
+	scnN.nmhdr.code = NPPN_FILEBEFORERENAME;
+	scnN.nmhdr.hwndFrom = _pPublicInterface->getHSelf();
+	scnN.nmhdr.idFrom = (uptr_t)bufferID;
+	_pluginsManager.notify(&scnN);
```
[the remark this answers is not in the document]
alankilborn: The model for the new function is the existing `bool Notepad_plus::fileRename(BufferID id)` function.  In `fileRename()`, it triggers the `NPPN_FILEBEFORERENAME` _before_ doing any error checking.  Are we sure we want to do something different here?

PowerEditor/src/NppIO.cpp
alankilborn: It's fine (and I will make the desired changes); however, it goes against my normal coding style to `return` from many different places within a function, just to save some indent "depth".  I think mine in this regard is a more-modern coding style, but, as I said, I will certainly adapt to what is wanted.

PowerEditor/src/Notepad_plus.h
```diff
@@ -190,6 +190,7 @@ friend class FileManager;
 	bool fileSaveAs(BufferID id = BUFFER_INVALID, bool isSaveCopy = false);
 	bool fileDelete(BufferID id = BUFFER_INVALID);
 	bool fileRename(BufferID id = BUFFER_INVALID);
+	bool fileRenameUntitled(BufferID id = BUFFER_INVALID, const wchar_t* tabNewName = nullptr);
```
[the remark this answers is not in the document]
alankilborn: Sure, I was just following the "model" set by how `fileSaveAs`, `fileDelete` and `fileRename` were done in the preceding lines.

PowerEditor/src/NppIO.cpp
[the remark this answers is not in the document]
donho: > Are we sure we want to do something different here?

Yes. If the code could be easier and more readable, I don't see why we shouldn't make it better.
 
I will review the code of `Notepad_plus::fileRename(BufferID id)`, and do a refactoring if it's necessary.

PowerEditor/src/NppIO.cpp
donho: > I think mine in this regard is a more-modern coding style

It's subjective.

It's not about saving some indent characters. It's about efficiency & easy to be maintained:
* For the efficiency: in the case that one plugin calls this API and fails to change the tab, all plugins will get the notification 2 times for nothing.
* For the facility of maintenance: The code of this PR have too many levels of imbrication, that leads code not easy to read. Since I am the principle maintainer and the chance is big that people's code will be maintained by me in the future, I have to ask people respect the coding style. Hope you understand.

myTempTemp:
340082238-bd0869f0-a8c0-4d0a-9663-dee784d9c1bb

PowerEditor/src/NppIO.cpp
alankilborn: No worries, I will adapt.  It's just hard to know what is wanted before doing it.

PowerEditor/src/NppIO.cpp
alankilborn: I wasn't sure where to put it, so I put it close to top-of-file -- hope that is reasonable.

I also defined a shared constant for the maximum length of the name.  Existing code has it using the max length for a language name, but this seems unrelated, so I gave it its own constant.

PowerEditor/src/NppBigSwitch.cpp
```diff
@@ -3142,6 +3128,12 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 			return colorId;
 		}
 
+		case NPPM_SETUNTITLEDNAME:
+		{
+			if (!wParam || !lParam) return false;
```
donho, line 3133: Please use `FALSE` instead of `false` to meet the coding standard of this file.

PowerEditor/src/NppIO.cpp
```diff
@@ -1900,16 +1907,11 @@ bool Notepad_plus::fileRename(BufferID id)
...
-		// Reserved characters:  < > : " / \ | ? * tab  
-		//  ("tab" is not in the official list, but it is good to avoid it)
-		std::wstring reservedChars = TEXT("<>:\"/\\|\?*\t");
-
 		std::wstring staticName = _nativeLangSpeaker.getLocalizedStrFromID("tabrename-newname", L"New name");
 
 		StringDlg strDlg;
 		std::wstring title = _nativeLangSpeaker.getLocalizedStrFromID("tabrename-title", L"Rename Current Tab");
-		strDlg.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), title.c_str(), staticName.c_str(), buf->getFileName(), langNameLenMax - 1, reservedChars.c_str(), true);
+		strDlg.init(_pPublicInterface->getHinst(), _pPublicInterface->getHSelf(), title.c_str(), staticName.c_str(), buf->getFileName(), untitledMaxNameLength, filenameReservedChars.c_str(), true);
```
donho, line 1914: StringDlg class comes from `UserDefineDialog.h`, so please use original value `langNameLenMax - 1` instead of `untitledMaxNameLength`.

donho (changes requested):

alankilborn:
I will be away until June 21, but I will make the requested changes then.

alankilborn:
@donho, okay...ready for more review

Ekopalypse:
@alankilborn 

I don't understand how this is supposed to work if -1 is provided by a plugin.
The check for BUFFER_INVALID, which is 0, works, but because of the -1 Npp crashes as this is not a valid BufferID.
What have I missed?

alankilborn:
@Ekopalypse 

>What have I missed?

_You_ haven't missed anything.
_I_, on the other hand, missed doing things correctly.  :-(
I will fix it.

Ekopalypse:
Are you going to use 0 as the default?

alankilborn:
>Are you going to use 0 as the default?

I think that makes sense here.
Obviously no one but me is using this plugin command, and I'm (clearly) not sending -1 to it when I call it.

@donho What do you think?  Can we change wParam == 0 to mean "currently active buffer"?  We need a fix, so I will start by doing this...

alankilborn:
Created separate issue:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15970

PR #15296, opened 2024-06-15
Add double buffering for status bar and tab bar to avoid flickering

### LINKED ISSUE #15260: [BUG] Status bar flickers when updated

### Description of the Issue

Hi everyone.

Sometimes I use Notepad++ to write a lot of text in a kind of distraction-free mode (F11 full-screen, dark mode, etc.) and I have noticed during this that the status bar flickers quite often.

Because this ruins the nice experience a bit, I want to propose fixing this by implementing double buffering.

### Steps To Reproduce

1.  Open Notepad++
2.  Start writing some long text, or even just hold down a random character key

### Current Behavior

The status bar (`length : 51    lines : 1 | Ln : 1    Col : 52    Pos : 52 | ...` ) is redrawn after each keystroke.
This is done in two steps, as typical for Win32 controls - first erase the background and then write the new text over it.
Sometimes these two steps happen separately and become visible to the user as flickering of the text.

### Expected Behavior

Updates of the status bar are flicker-free. User experience is smooth, fluent and without distractions.

### Debug Information

Tested also on the latest commit of "master" from today.
Happens for both normal mode (standard control) and dark mode (custom drawn).

### Anything else?

~Proposed fix in a fork: https://github.com/yirkha/notepad-plus-plus/commit/8ecc6697214c1c2d03099b3f5763dddbf4dff5c1~ Edit: obsolete

Thank you for your consideration

yirkha:
Fixes #15260

### DIFF
--- a/PowerEditor/src/WinControls/DoubleBuffer.h
+++ b/PowerEditor/src/WinControls/DoubleBuffer.h
@@ -1,0 +1,101 @@
+// This file is part of Notepad++ project
+// Copyright (C) 2024 Jiri Hruska <jirka@fud.cz>
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// at your option any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#pragma once
+#include <windows.h>
+#include <cassert>
+
+class DoubleBuffer final
+{
+private:
+	SIZE _size{};
+	HDC _hMemoryDC = nullptr;
+	HBITMAP _hDefaultBitmap = nullptr;
+	HBITMAP _hBufferBitmap = nullptr;
+
+public:
+	DoubleBuffer() {}
+	DoubleBuffer(const DoubleBuffer&) = delete;
+	DoubleBuffer& operator=(const DoubleBuffer&) = delete;
+
+	~DoubleBuffer()
+	{
+		if (_hBufferBitmap)
+		{
+			::SelectObject(_hMemoryDC, _hDefaultBitmap);
+			::DeleteObject(_hBufferBitmap);
+		}
+
+		if (_hMemoryDC)
+		{
+			::DeleteDC(_hMemoryDC);
+		}
+	}
+
+	HDC beginPaint(HWND hWnd, PAINTSTRUCT* ps)
+	{
+		if (!::BeginPaint(hWnd, ps))
+		{
+			return nullptr;
+		}
+
+		if (!_hMemoryDC)
+		{
+			_hMemoryDC = ::CreateCompatibleDC(ps->hdc);
+		}
+
+		RECT clientRc{};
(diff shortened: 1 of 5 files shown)


donho:
@ozone10 
I suppose you have reviewed the code and it's OK for you.

ozone10:
@donho 
I only reviewed older version with only status bar. There some code style which I would change. 
And mainly double buffer won't work correctly for tab bar in light mode.

@yirkha 
unfortunately there is issue with `WM_PRINTCLIENT`, it only take into account visible part of window. In dark mode there is no issue, because we don't use it. For status bar also no issue, because it is always visible. But for tab bar, when it is resized, it will create black parts.

So the solution for now is to use double buffer only for dark mode for controls, which can be resized.

PowerEditor/src/WinControls/TabBar/TabBar.cpp
```diff
@@ -915,35 +915,29 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 
 		case WM_ERASEBKGND:
```
ozone10, line 916: For now use something like this
```c++
case WM_ERASEBKGND:
{
	if (!NppDarkMode::isEnabled())
	{
		break;
	}
	return FALSE;
}

case WM_PAINT:
{
	LONG_PTR dwStyle = GetWindowLongPtr(hwnd, GWL_STYLE);
	if (!NppDarkMode::isEnabled() || !(dwStyle & TCS_OWNERDRAWFIXED))
	{
		break;
	}

	PAINTSTRUCT ps{};
	HDC hdc = _dblBuf.beginPaint(hwnd, &ps);
```

replacing everything until 
`const bool hasMultipleLines = ((dwStyle & TCS_BUTTONS) == TCS_BUTTONS);`

PowerEditor/src/WinControls/DoubleBuffer.h
```diff
@@ -0,0 +1,101 @@
...
+	{
+		::BitBlt(
+			ps->hdc, ps->rcPaint.left, ps->rcPaint.top, ps->rcPaint.right - ps->rcPaint.left, ps->rcPaint.bottom - ps->rcPaint.top,
+			_hMemoryDC, ps->rcPaint.left, ps->rcPaint.top,
+			SRCCOPY);
+
+		::EndPaint(hWnd, ps);
+	}
+
+	int getWidth() const
```
ozone10, line 92: check https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#general for braces style

yirkha:
@ozone10 Thank you for the review and sorry it got closed before I could react to your feedback. I was fine with making sure everything was correct before merging it &ndash; all this is really just cosmetic improvement, nothing that needs to be rushed.

> check CONTRIBUTING.md#general for braces style
Gotcha, for some reason I thought that applied only to the classes split between .cpp/.h files.

I'll align it with the guidelines it in the next PR , when I'll figure out what to do with the docking container captions etc.
[Edit: or for the issue below 😞]

> unfortunately there is issue with WM_PRINTCLIENT, it only take into account visible part of window. (...)
> for tab bar, when it is resized, it will create black parts.

Ouch, sounds serious. I've tested all kinds of things in the default (non-dark) mode, add/remove tabs, resize the window from various edges, even with parts of the window off-screen, and I did not see any of that... Now I realized I had always "[x] Multi-line" tab bar enabled and it needs the default (facepalm).

Anyway, now that I could repro it, I debugged it, seems just like a bug with the DoubleBuffer class, that it does not set/reset clipping region between WM_PAINTs correctly (and the common Tab control unusually leaves its own clipping region set on the inner HDC between calls). Thanks for noticing this, I'll post a fix asap (and test all combinations better this time)

PR #15304, opened 2024-06-16
Update Notepad_plus.rc

myTempTemp:


conky77:
@donho spam in pull requests as well

myTempTemp:
> @donho spam in pull requests as well

This account is used to scold you

PR #15314, opened 2024-06-17
Fix crash on Windows Core Server

### LINKED ISSUE #15313: NOTEPAD++ v8.6.8 Crashed and throwing exception when launching on Windows Core Server 2022

### Description of the Issue

Getting error: 

### Steps To Reproduce

### Current Behavior

Throwing exception- 

### Expected Behavior

Notepad++ should be launched.

### Debug Information

```shell
```

### Anything else?

Notepad++ErrorOnCORE

donho:
LoadIconWithScaleDown function could not be in runtime of Windows Core Server. 
In order to prevent Notepad++ from crash, this commit use the old call (LoadIcon) in case of loading failed.

Fix #15313

### DIFF
--- a/PowerEditor/src/dpiManagerV2.cpp
+++ b/PowerEditor/src/dpiManagerV2.cpp
@@ -286,4 +286,10 @@ void DPIManagerV2::loadIcon(HINSTANCE hinst, const wchar_t* pszName, int cx, int
 	{
 		*phico = static_cast<HICON>(::LoadImage(hinst, pszName, IMAGE_ICON, cx, cy, fuLoad));
 	}
+
+	// in case of LoadIconWithScaleDown is not compatible on Windows version: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15313
+	if (!*phico)
+	{
+		*phico = ::LoadIcon(hinst, pszName);
+	}
 }


PowerEditor/src/dpiManagerV2.cpp
```diff
@@ -286,4 +286,10 @@ void DPIManagerV2::loadIcon(HINSTANCE hinst, const wchar_t* pszName, int cx, int
 	{
 		*phico = static_cast<HICON>(::LoadImage(hinst, pszName, IMAGE_ICON, cx, cy, fuLoad));
 	}
+
+	// in case of LoadIconWithScaleDown is not compatible on Windows version: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15313
+	if (!*phico)
```
ozone10, line 291: this can be added to previous check
`if ((::LoadIconWithScaleDown(hinst, pszName, cx, cy, phico) != S_OK) || (*phico == nullptr))`

[LoadImage](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadimagew) is supported from Windows 2000 and even MS recommend to use it instead of `LoadIcon`

ozone10:
If this PR does not fix it, I could try to use undocumented way to check if `LoadIconWithScaleDown` is available.

Pmalhotra2024:
image

still crashing with same error.

donho:
@ozone10 
> If this PR does not fix it, I could try to use undocumented way to check if LoadIconWithScaleDown is available.

Obviously the PR doesn't work.
Please create a new PR if you've got other solution.

PR #15315, opened 2024-06-17
Fix exception/crash on Windows Server Core 2022

### LINKED ISSUE #15313: NOTEPAD++ v8.6.8 Crashed and throwing exception when launching on Windows Core Server 2022

ozone10:
Make loading icons fail-safe.

Fix #15313

### DIFF
--- a/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
+++ b/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
@@ -43,16 +43,47 @@ void IconList::create(int iconSize, HINSTANCE hInst, int* iconIDArray, int iconI
 		addIcon(iconIDArray[i], iconSize, iconSize);
 }
 
-void IconList::addIcon(int iconID, int cx, int cy) const
+void IconList::addIcon(int iconID, int cx, int cy, int failIconID) const
 {
 	HICON hIcon = nullptr;
 	DPIManagerV2::loadIcon(_hInst, MAKEINTRESOURCE(iconID), cx, cy, &hIcon, LR_DEFAULTSIZE);
 
 	if (!hIcon)
-		throw std::runtime_error("IconList::addIcon : LoadIcon() function return null");
+	{
+		static bool ignoreWarning = false;
+		int userAnswer = 0;
+		if (!ignoreWarning)
+		{
+			userAnswer = ::MessageBoxA(NULL, "IconList::addIcon : LoadIcon() function return null.\nIgnore warning?", std::to_string(iconID).c_str(), MB_YESNOCANCEL | MB_ICONWARNING);
+			ignoreWarning = userAnswer == IDYES;
+		}
 
-	ImageList_AddIcon(_hImglst, hIcon);
-	::DestroyIcon(hIcon);
+		if (userAnswer == IDNO)
+		{
+			throw std::runtime_error("IconList::addIcon : LoadIcon() function returns null");
+		}
+		else
+		{
+			if (failIconID != -1)
+			{
+				HBITMAP hBmp = static_cast<HBITMAP>(::LoadImage(_hInst, MAKEINTRESOURCE(failIconID), IMAGE_BITMAP, cx, cy, LR_LOADMAP3DCOLORS | LR_LOADTRANSPARENT));
+				if (hBmp != nullptr)
+				{
+					::ImageList_AddMasked(_hImglst, hBmp, ::GetSysColor(COLOR_3DFACE));
+					::DeleteObject(hBmp);
+					return;
+				}
+			}
+			constexpr int IDI_ICONABSENT = 104; // check resource.h for correct id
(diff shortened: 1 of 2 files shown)


ozone10:
@donho 
I cannot test is since I am away.

xomx:
@ozone10  @donho 

I do not think this will fix the issue, we probably jumped on that LoadIconWithScaleDown too quickly.

If it is problem with missing/unsupported LoadIconWithScaleDown comctl32 func, the error message will be different (it will originate directly from the OS).

I rather think that there is this code flow:

```
void DPIManagerV2::loadIcon(HINSTANCE hinst, const wchar_t* pszName, int cx, int cy, HICON* phico, UINT fuLoad)
{
	if (::LoadIconWithScaleDown(hinst, pszName, cx, cy, phico) != S_OK) // -> fails but no exception, so it goes to the LoadImage part...
	{
		*phico = static_cast<HICON>(::LoadImage(hinst, pszName, IMAGE_ICON, cx, cy, fuLoad)); // -> this fails too, so hIcon is still NULL...
	}
}
	
if (!hIcon)
	throw std::runtime_error("IconList::addIcon : LoadIcon() function return null"); // -> so it goes here and the message is from the main N++ exception handler...
```

---

I did not have time to search for the comctl32 suport in a Win Core Server but it makes sense to me that it supports that, otherwise it could not run common Win apps...

ozone10:
@xomx 
you might be right, because change to `DPIManagerV2::loadIcon` was already in 8.6.6

donho:
@xomx 
So you think it's failure (problem) from the OS?
If your assumption is true, why did v8.6.7 work?

xomx:
@donho 
> why did v8.6.7 work?

IDK. Maybe the v8.6.8 commits connected to icons somehow?
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/cdf32d198b84927539fc0428e3fdbbf09fee85df
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c89033bd8a4bb4505f26377d6a4deae36a505ac0
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6c660a71b38ab662f9cf8346c84bc6a39e3d957c

donho:
@xomx 
Let's wait and see the test result from @Pmalhotra2024

xomx:
@donho 

To find out more, create please a new DEBUG PR, with this patch for the DPIManagerV2::loadIcon:

```
#include <format> // TEMP: DEBUG:

void DPIManagerV2::loadIcon(HINSTANCE hinst, const wchar_t* pszName, int cx, int cy, HICON* phico, UINT fuLoad)
{
	HRESULT hr = ::LoadIconWithScaleDown(hinst, pszName, cx, cy, phico);
	if (hr != S_OK)
	{
		std::wstring msg = L"ERROR: LoadIconWithScaleDown WINAPI failed! (HRESULT: ";
		msg += std::format(L"{:#010x}", static_cast<DWORD>(hr)) + L")";
		::MessageBoxW(NULL, msg.c_str(), L"DPIManagerV2::loadIcon", MB_OK | MB_SYSTEMMODAL | MB_ICONERROR); // TEMP: DEBUG:

		*phico = static_cast<HICON>(::LoadImage(hinst, pszName, IMAGE_ICON, cx, cy, fuLoad));
		if (*phico == NULL)
		{
			msg = L"ERROR: LoadImage WINAPI failed! (ErrorCode: ";
			msg += std::to_wstring(::GetLastError()) + L")";
			::MessageBoxW(NULL, msg.c_str(), L"DPIManagerV2::loadIcon", MB_OK | MB_SYSTEMMODAL | MB_ICONERROR); // TEMP: DEBUG:
		}
	}
}
```

donho:
@xomx 

Thank you: done: #15316

ozone10:
issue is due to unsupported ico format

ozone10:
Reuse PR

ozone10:
msys2 failure is due to failure of updating msys2 packages

Maybe using other than transparent icon would be better.

donho:
> Maybe using other than transparent icon would be better.

@ozone10 Why "using other than transparent icon" would make updating msys2 packages better?

ozone10:
@donho 
I have accidentally removed part of text.

"msys2 failure is due to failure of updating msys2 packages", because their site cannot be reached.

Current PR use transparent placeholder. "Maybe using other than transparent icon would be better" for indication.

donho:
The PR is better than https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15331 is, though the dialog needs more clarification. 
I will add more text for it.

ozone10:
@xomx 
that is intended, because all fluent icons now use new ico files with only png images, so it will fail on every toolbar icons.
This will result in blank icons (as seen on tab bar, indicator and close button). 
But because toolbar standard icons use bmp files and these are loaded fine, I use them as some kind of fail-safe solution. User should still switch to standard icons.

Issue #15298, opened 2024-06-15
[Feature request] Hide Close button on inactive tabs


WildByDesign:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

This is just a concept at the moment. The inactive tabs might look cleaner and less busy by hiding the close button.

### Describe the solution you'd like.

Hide the close button on inactive tabs. Only show the close button on inactive tabs when the mouse hovers over.

This will require the close button BMP images to be converted to ICO to benefit from transparency and better scaling.

### Debug Information

```shell
Notepad++ v8.6.8   (64-bit)
Build time : Jun 15 2024 - 08:53:28
Path : D:\Tools\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 11 Pro (64-bit)
OS Version : 24H2
OS Build : 26100.863
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    RainLexer (2.21)
```

### Anything else?

![animation](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/26308319/53966498-5683-42df-a635-a0df1580cb5e)

alankilborn:
The proposal is not a good one.
It would turn a one-step close action into a two-step action -- first have to activate the tab, then press the `x`.

CennoxX:
It would be better, if the x shows direct, when hovering above the tab, not the place, where I suspect the close button to come. And I think it has to be possible to deactivate this behaviour.

myTempTemp:
侯今吾是一条狗 我家里有一条狗，我给它起名叫侯今吾，来吃点屎吧！

WildByDesign:
> when hovering above the tab, not the place, where I suspect the close button to come

This is a very good point, I agree.

donho:
@WildByDesign 
I will see if it's easy to be implemented. But I promise nothing.

donho:
@WildByDesign 
Different from file state images (saved, dirty, RO & monitoring) on the document tab, close buttons are in BMP format.

After some googling, the bitmap support the transparency:
https://stackoverflow.com/questions/14195289/can-a-bmp-image-format-handle-transparency

So if you manage to create a transparent bmp and use it to replace **closeTabButton_inact.bmp**, then this FR is feasible.
[...]

PR #15317, opened 2024-06-17
Replace tab bar close button BMP files with ICO

ozone10:
ref #15298


donho:
@ozone10 
image

The left icons (the PR) looks not the same as the right ones.

WildByDesign:
@ozone10 @donho  Thank you for the work on this. I have something really special coming for these icons that I started on last night, but they’ve got to be 100% perfect first.

PR #15319, opened 2024-06-18
Replace WM_PRINTCLIENT with WM_PRINT to avoid black parts when resizing tabbar

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15296#issuecomment-2171221062

### DIFF
--- a/PowerEditor/src/WinControls/TabBar/TabBar.cpp
+++ b/PowerEditor/src/WinControls/TabBar/TabBar.cpp
@@ -966,8 +966,8 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
 			{
 				// Even if the tab bar common control is used directly, e.g. in non-dark mode,
 				// it suffers from flickering during updates, so let it paint into a back buffer
-				::CallWindowProc(_tabBarDefaultProc, hwnd, WM_ERASEBKGND, reinterpret_cast<WPARAM>(hdc), 0);
-				::CallWindowProc(_tabBarDefaultProc, hwnd, WM_PRINTCLIENT, reinterpret_cast<WPARAM>(hdc), PRF_NONCLIENT | PRF_CLIENT);
+				::DefWindowProc(hwnd, WM_ERASEBKGND, reinterpret_cast<WPARAM>(hdc), 0);
+				::DefWindowProc(hwnd, WM_PRINT, reinterpret_cast<WPARAM>(hdc), PRF_NONCLIENT | PRF_CLIENT);
 				_dblBuf.endPaint(hwnd, &ps);
 				return 0;
 			}


donho:
@ozone10 

How can we reproduce the black bar?

ozone10:
@donho 
1. open some panel on right side (function list, document map, ...)
2. make sure that multiline tabbar is disabled
3. resize panel via splitter

ozone10:
Also I would like to replace old subclassing with newer recommended one https://learn.microsoft.com/en-us/windows/win32/controls/subclassing-overview to improve stability and to avoid using `CallWindowProc` and `GWLP_USERDATA`

ozone10:
image

donho:
@ozone10 
> Also I would like to replace old subclassing with newer recommended one https://learn.microsoft.com/en-us/windows/win32/controls/subclassing-overview to improve stability and to avoid using CallWindowProc and GWLP_USERDATA

OK, let's do it little by little.
If you want to start from this PR, please let me know.

ozone10:
@donho
> OK, let's do it little by little.
> If you want to start from this PR, please let me know.

I am away so maybe next PR, when I get time to properly do test. 

I did this PR mainly to fix regression.

yirkha:
>>  let's do it little by little.
>
> I did this PR mainly to fix regression.

Yeah I agree, let's fix broken master first and then think of the next steps.

yirkha:
> From the looks of it, I'd think that DefWindowProc() would only call this (subclassed) window procedure again. Just maybe a bit differently, per whatever it does with WM_PRINT, which avoids problems?

Okay, I checked what `DefWindowProc()` does with `WM_PRINT` and it boils down to this:

```c++
HDC savedDc = SaveDC(hdc);
SendMessage(hWnd, WM_ERASEBKGND, (WPARAM) hdc, 0);
SendMessage(hWnd, WM_PRINTCLIENT, (WPARAM) hdc, lParam);
RestoreDC(hdc, savedDc);
```

So, now the comctl's tab control's changes of clipping rectangle are not kept on the raw HDC sent in anymore, and the clipping is restored at the `RestoreDC()` time.

So I still think that fixing DoubleBuffer with the clipping region reset would be worth it, as this PR fixed the issue with DC reuse only in this specific case..

Whether to actually call WM_PRINT or WM_PRINTCLIENT or what is another matter. Looking at the `DefWindowProc()` handling again, my original approach with two separate message dispatches (`WM_ERASEBKGND` and `WM_PRINTCLIENT`) might be unoptimal, we could have just one `WM_PRINT` with `PRF_ERASEBKGND | PRF_CLIENT`.

ozone10:
@yirkha 
this is just quick fix, feel free to do additional PR.

ozone10:
@yirkha @donho 
another issue with status bar in light mode
image

STR
1. Open Find dialog (ctrl+f)
2. Trigger text in status bar, e.g. press Count

PR #15320, opened 2024-06-19
Fix black text background in Find dialog status bar

ozone10:
ref https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15319#issuecomment-2177773329


yirkha:
Thanks for the fix, @ozone10. It was caused by a bug in the `DoubleBuffer` implementation (which I'm gonna fix now), but still the whole `getBackgroundColor()` here seemed superfluous and `TRANSPARENT` is so much better 👍

PR #15321, opened 2024-06-19
Fluent Icons - Close Buttons

WildByDesign:
- update `g_TabCloseBtnSize` size from 11 to 16
- having an uneven numbered (11) size was problematic
- updated close button Fluent icons (custom)
- added 16, 20, 24 and 32 pixel icons for better scaling
- updated Standard close button icons from 11 to 16 pixels
- added a small amount of transparent padding to avoid icon stretching


PowerEditor/src/WinControls/TabBar/TabBar.h
```diff
@@ -53,7 +53,7 @@ constexpr int g_TabHeight = 22;
 constexpr int g_TabHeightLarge = 25;
 constexpr int g_TabWidth = 45;
 constexpr int g_TabWidthCloseBtn = 60;
-constexpr int g_TabCloseBtnSize = 11;
+constexpr int g_TabCloseBtnSize = 16;
```
donho, line 56: The left icons are the current old one, and the right ones are resized of current PR:

image

Can't we keep the size 11?

donho (changes requested):
Please restore the icons under standard/tabbar/ and size 11
Let's change only fluent icons.

PowerEditor/src/WinControls/TabBar/TabBar.h
WildByDesign: > Can't we keep the size 11?

Is there any possible way to keep 11 for the Standard icons and have the code do 16 when dark mode is selected? 

That would be the best for both scenarios if that is possible.

donho: @WildByDesign 
Yes. I think.
Please check the above request - I have just changed it.

WildByDesign: Thank you. I will do this but I away from my main system right now. I will test this out later in the day and less you know.

WildByDesign:
@donho  I made the changes that you suggested, but I am getting build errors related to those changes:

`Error	C2065	'g_TabCloseBtnSize_DM': undeclared identifier	Notepad++	D:\Tools\_compile\notepad-plus-plus-fluent-icons-closebuttons\PowerEditor\src\WinControls\TabBar\TabBar.cpp	471`

`Error	C2737	'btnSize': const object must be initialized	Notepad++	D:\Tools\_compile\notepad-plus-plus-fluent-icons-closebuttons\PowerEditor\src\WinControls\TabBar\TabBar.cpp	471`

Do you know what I can do to fix this? Thanks

donho:
> Add constexpr int g_TabCloseBtnSize_DM = 16; under constexpr int g_TabCloseBtnSize = 11;

@WildByDesign 
Have you the following 2 lines:

```cpp
constexpr int g_TabCloseBtnSize = 11;
constexpr int g_TabCloseBtnSize_DM = 16; 
````

in TabBar.h?

**Comment by WildByDesign** [comment]
> in TabBar.h?

@donho  Yes, I have both lines in that file.

**Comment by donho** [comment]
@WildByDesign 
Please push your change directely, so I can check it for you.

**Comment by ozone10** [comment]
@donho @WildByDesign 
don't separate size between dark/light mode, with current implementation it won't work correctly.
If even size is needed, 12 should be enough. Also don't add transparent padding, with current hover zone implementation, it will create confusion for users. Hovering or pressing button when cursor is not on "button" will cause issue.

**Comment by donho** [comment]
> don't separate size between dark/light mode, with current implementation it won't work correctly.
If even size is needed, 12 should be enough.

@WildByDesign 
I think @ozone10 is right about it.
Please try to make it 12.

**Comment by WildByDesign** [comment]
The problem with 12 is that we still end up with odd numbers.

For example:
```
Scaling				
100 (base)	16	14	12	11	20
125		20	18	15	14	25
150		24	21	18	17	29
```

I did try with 12 pixel base icons yesterday. The problem is that with my own system (125% scaling) it ends up with a 15 pixel icon that looks terrible. All of those examples above end up with odd numbers, except the one with a base of 16.

I will look into this in a few hours when I get some time in the evening.

Even if we go with 12 as a base pixel size, it will end up stretching the 11 pixel standard icons slightly.

donho:
@WildByDesign 
I see your point.

@ozone10 
> don't separate size between dark/light mode, with current implementation it won't work correctly.
m
Though I'm not for complicating the code and agree with you, why 11 & 16 won't work? I didn't see any problem with my quick test.

WildByDesign:
There is also the possibility of adding light mode (inverted) versions of the Fluent icons for the close buttons. 

That could potentially be used instead of the standard icons since those don’t scale without stretching up in size.

ozone10:
@donho 
> Though I'm not for complicating the code and agree with you, why 11 & 16 won't work? I didn't see any problem with my quick test.

Implementation uses one imagelist for all icons, and thus one size will be used. Icons which don't have same size as imagelist will be adapted to imagelist's size and won't be sharp.

ozone10:
PR #15326

ozone10:
@WildByDesign 
could you make inactive close button icon brighter, not everyone uses darker colors for inactive tabs.
image

This subjective, but I would switch hover and pushed icons.

WildByDesign:
> could you make inactive close button icon brighter, not everyone uses darker colors for inactive tabs. image
> 
> This subjective, but I would switch hover and pushed icons.

Both are great points, thank you. I was actually thinking about both of those points last night. I will do both suggestions and resubmit in a few hours.

WildByDesign:
I believe that I have followed through with all suggestions. I brightened up the inactive close button. I swapped the hover and push icons. 

Please note that this absolutely must be merged with @ozone10 's PR (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15326) to work properly. I have tested these with that PR and everything looks great from my testing.

Thank you both for all of the help and suggestions.

donho:
@ozone10 
> PR #15326

So the PR #15326 allow both sizes: 11 (standard) & 16 (dark mode) on document tab?

donho:
> So the PR #15326 allow both sizes: 11 (standard) & 16 (dark mode) on document tab?

OK, just checked the PR, it is.

donho:
Though I don't use dark mode, it looks really good!
Thank you @WildByDesign & @ozone10 for making this happen!

WildByDesign:
You're welcome, my pleasure. I'm happy to help.

donho:
@WildByDesign 
`[Merge branch 'notepad-plus-plus:master' into fluent-icons-closebuttons](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15321/commits/fbdb90c748b26266601672bd4ea82d785ea39d23)`

Never merge the master into your branch.
I cannot integrate this PR because of it.
I have to close this one.
Could you please:
1. Update your local master
2. create a NEW branch then create a NEW PR

WildByDesign:
@donho  Sorry about that mistake. I will ensure that I don't do that again.

Here is updated PR: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15336

PR #15331, opened 2024-06-20
Fix exception/crash on Windows Server Core 2022 (Datacenter)

### LINKED ISSUE #15313: NOTEPAD++ v8.6.8 Crashed and throwing exception when launching on Windows Core Server 2022

donho:
The exception/crash occurs due to the unsupported icon format in Windows Server Core 2022 (due to the lack of a graphical library). 

The commit represents a balance between the functionality of loading icons and the minimal graphic support required for Notepad++:
In the event of loading problems in a typical Windows version with complete graphical library support (which is unlikely but possible), users will receive alerts, and the ICO ID in question will be displayed in the caption. 
However, under Windows Server Core 2022, multiple alert message boxes may appear during Notepad++ launch, but they can be stopped after the first one, allowing the user to use Notepad++ normally.

image

Fix #15313

### DIFF
--- a/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
+++ b/PowerEditor/src/WinControls/ImageListSet/ImageListSet.cpp
@@ -49,7 +49,15 @@ void IconList::addIcon(int iconID, int cx, int cy) const
 	DPIManagerV2::loadIcon(_hInst, MAKEINTRESOURCE(iconID), cx, cy, &hIcon, LR_DEFAULTSIZE);
 
 	if (!hIcon)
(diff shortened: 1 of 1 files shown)


donho:
@xomx @ozone10 
What do you think this solution?

ozone10:
@donho 
I have just noticed that we have similar idea for message, check reworked PR #15315. It also contains logic for loading bmp files for toolbar icons, since standard icons use them and it's better than having blank icons.

PR #15345, opened 2024-06-23
WIP: Splitting of sidebars

### LINKED ISSUE #14745: [FR] Allow multiple simultaneous windows in single sidebar
### Description of the feature
Would it be possible to allow multiple windows to be shown in single sidebar at the same time? There are windows that would be really helpful to be able to see, but are almost empty, so there isn't a reason why they should take up the whole sidebar.
I'd honestly be surprised if this isn't a duplicate, because this seems like a no-brainer, but I haven't found anything similar.

image

sdasda7777:
Aims to fix #14745 (at some point in future)


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14745#issuecomment-2186350782

PR #15351, opened 2024-06-24
Improve description for settings "Backup"

donho:
Fix #14560

PR #15353, opened 2024-06-24
[xml] Update german.xml

schnurlos:
Modified backspace description.

### DIFF
--- a/PowerEditor/installer/nativeLang/german.xml
+++ b/PowerEditor/installer/nativeLang/german.xml
@@ -1115,7 +1115,7 @@ Translation note:
 					<Item id="6310" name="Einrückung verwendet:"/>
 					<Item id="6311" name="Tab Zeichen"/>
 					<Item id="6510" name="Standard verwenden"/>
-					<Item id="6512" name="Die Rückschrittaste U+232B entfernt Einrückungen, anstatt einzelne Leerzeichen zu entfernen"/>
+					<Item id="6512" name="Die Rückschrittaste U+232B entfernt Einrückungen statt einzelne Leerzeichen"/>
 					<Item id="6335" name="Backslash ist Escapezeichen für SQL"/>
 				</Language>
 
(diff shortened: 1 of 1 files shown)


PowerEditor/installer/nativeLang/german.xml
```diff
@@ -1115,7 +1115,7 @@ Translation note:
 					<Item id="6310" name="Einrückung verwendet:"/>
 					<Item id="6311" name="Tab Zeichen"/>
 					<Item id="6510" name="Standard verwenden"/>
-					<Item id="6512" name="Rückschritttaste entfernt Einrückungen, anstatt einzelne Leerzeichen zu entfernen"/>
+					<Item id="6512" name="Die Rückschrittaste U+232B entfernt Einrückungen statt einzelne Leerzeichen"/>
```
MarkusBodensee, line 1118: I would just call it Rücktaste which seems the most common for me, U+232B is imo not needed, so:

"Rücktaste entfernt Einrückungen anstatt einzelner Leerzeichen"

donho: Are there more German speakers who can provide the opinion regarding this change?

MarkusBodensee: LEO.org validates it, too. I think Rücktaste is the correct and most common word for the backspace key.

https://dict.leo.org/englisch-deutsch/backspace

grafik

schnurlos:
Now waiting for https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15351 getting committed.

PowerEditor/installer/nativeLang/german.xml
CennoxX: As a German I say Backspace-Taste (which is the name in the according Wikipedia article too). With Rücktaste I would know what is meant. Rückschritttaste is rather uncommon.

PR #15364, opened 2024-06-26
Do not recognize files without lnk-extension as Windows shortcuts

### LINKED ISSUE #9643: Opening .lnk opens destination file for editing
### Description of the Issue
As the title says, opening `.lnk` file using Notepad++ opens shortcut's destination file for editing.

1. Navigate to `C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Administrative Tools` using Explorer
2. Right-click `Registry Editor.lnk` and select `Edit with Notepad++`.

### Expected Behavior
`Registry Editor.lnk` file is opened for editing.

### Actual Behavior
`%windir%\regedit.exe` file is opened for editing instead.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Admin mode : ON
Local Conf mode : OFF
OS Version : 2009
OS Build : 19042.867
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll nppcrypt.dll NppExport.dll NPPJSONViewer.dll NppMarkdownPanel.dll 

I see that there was an issue #8483 where someone requested this functionality.

If you wonder why would anyone want to edit `.lnk` file itself, I have at least two use cases:

1. To delete its contents (i.e. to make a 0-byte file), and set it to hidden/system/readonly so that rude program installers and updaters cannot litter my start menu with shortcuts I don't need or use.
2. To copy links of special shortcuts that are otherwise not possible to copy because they are grayed in shortcut properties dialog.

I also believe that a text editor should not be opening `.exe` files by default (unless it specifically supports modifications to `PE/PE+` executable structure), because people can accidentally change and save them and thus break their programs or worse yet their Windows install.

If there is no option to turn this off then please consider it as a feature request.

If adding an option is too much of a hassle, then please at least exclude shortcuts to `*.exe` from opening the target instead of shortcut itself.

### LINKED ISSUE #10139: CAN NOT edit the "*.lnk" file correctly
### Description of the Issue

can not edit the "*.lnk" file correctly 

(If the old way gone forever with no choice, then, it is ok, but I have to force myself to turn to other editors. Sorry for the trouble. I do not want to edit the "target file" when I open the "*.lnk" file, I do know where the target files are, I can and I will open them directly if I want to edit them.)

1. drag&drop the "*.lnk" file to the npp or open by File-->Open

### Expected Behavior

can edit the "*.lnk" file itself

### Actual Behavior

edit the linked/target file

### LINKED ISSUE #11089: File content, not extension is used to determine if a file is a shortcut

xomx:
Fix #11089, fix #10139, fix #9643

Also contains minor fix (using the same "Upperize" method) for the reported https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15211#issuecomment-2156495651 .

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1552,11 +1552,7 @@ bool isUnsupportedFileName(const generic_string& fileName)
 					fileNameOnly = fileNameOnly.substr(pos + 1);
 
 				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
-#ifdef UNICODE
 				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
-#else
-				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::toupper);
-#endif
 
 				const std::vector<generic_string>  reservedWin32NamespaceDeviceList{
 				TEXT("CON"), TEXT("PRN"), TEXT("AUX"), TEXT("NUL"),
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1551,6 +1551,13 @@ bool isUnsupportedFileName(const generic_string& fileName)
 				if (pos != std::string::npos)
 					fileNameOnly = fileNameOnly.substr(pos + 1);
 
+				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
+#ifdef UNICODE
+				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
+#else
+				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::toupper);
+#endif
+
 				const std::vector<generic_string>  reservedWin32NamespaceDeviceList{
 				TEXT("CON"), TEXT("PRN"), TEXT("AUX"), TEXT("NUL"),
 				TEXT("COM1"), TEXT("COM2"), TEXT("COM3"), TEXT("COM4"), TEXT("COM5"), TEXT("COM6"), TEXT("COM7"), TEXT("COM8"), TEXT("COM9"),
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -128,6 +128,16 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 
 bool resolveLinkFile(generic_string& linkFilePath)
 {
+	// upperize for the following comparison because the ends_with is case sensitive unlike the Windows OS filesystem
+	generic_string linkFilePathUp = linkFilePath;
+#ifdef UNICODE
+	std::transform(linkFilePathUp.begin(), linkFilePathUp.end(), linkFilePathUp.begin(), ::towupper);
+#else
+	std::transform(linkFilePathUp.begin(), linkFilePathUp.end(), linkFilePathUp.begin(), ::toupper);
+#endif
+	if (!linkFilePathUp.ends_with(TEXT(".LNK")))
+		return false; // we will not check the renamed shortcuts like "file.lnk.txt"
+
 	bool isResolved = false;
 
 	IShellLink* psl = nullptr;
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -126,16 +126,12 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 	return EXIT_SUCCESS;
 }
 
-bool resolveLinkFile(generic_string& linkFilePath)
+bool resolveLinkFile(std::wstring& linkFilePath)
 {
 	// upperize for the following comparison because the ends_with is case sensitive unlike the Windows OS filesystem
-	generic_string linkFilePathUp = linkFilePath;
-#ifdef UNICODE
+	std::wstring linkFilePathUp = linkFilePath;
 	std::transform(linkFilePathUp.begin(), linkFilePathUp.end(), linkFilePathUp.begin(), ::towupper);
-#else
-	std::transform(linkFilePathUp.begin(), linkFilePathUp.end(), linkFilePathUp.begin(), ::toupper);
-#endif
-	if (!linkFilePathUp.ends_with(TEXT(".LNK")))
+	if (!linkFilePathUp.ends_with(L".LNK"))
 		return false; // we will not check the renamed shortcuts like "file.lnk.txt"
 
 	bool isResolved = false;


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1551,6 +1551,13 @@ bool isUnsupportedFileName(const generic_string& fileName)
 				if (pos != std::string::npos)
 					fileNameOnly = fileNameOnly.substr(pos + 1);
 
+				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
+#ifdef UNICODE
+				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
+#else
+				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::toupper);
+#endif
+
```
donho, line 1560: The UNICODE symbol is always defined in Notepad++ project, and Notepad++ has been a Unicode application since long time ago.
So the 2nd instruction could be removed IMO.

donho (changes requested):
Please follow the suggestion.

PowerEditor/src/MISC/Common/Common.cpp
xomx: Until there are the `generic_string` signatures,(here the `bool isUnsupportedFileName(const generic_string& fileName)` and `bool resolveLinkFile(generic_string& linkFilePath)`), the correct way is to use these UNICODE constructs.

In the future, it will need a some breakpoint PR with refactoring, where the switching to such UNICODE-only chars/strings will be finished. I mean the `generic_string` to `std::wstring`, `TCHAR` to `WCHAR` or `char` to `wchar_t`). And also removing all of these #ifdefs.

But if you rather want to do it in small steps, I will left there only the UNICODE version lines, as requested. (?)

donho: > Until there are the generic_string signatures,(here the bool isUnsupportedFileName(const generic_string& fileName) and bool resolveLinkFile(generic_string& linkFilePath)), the correct way is to use these UNICODE constructs.

That's the reason that I use rather `std::wstring` instead of `generic_string`. And in long term, `generic_string` will be removed.

In this case, please change the signature of function resolveLinkFile to `bool resolveLinkFile(std::wstring& linkFilePath)`.

> But if you rather want to do it in small steps, I will left there only the UNICODE version lines, as requested. (?)

Yes, please remove UNICODE - that might lead the confusion that Notepad++ still supports ANSI version.

xomx: @donho 
Done.

donho: @xomx 
Sorry if I wasn't clear.

I saw that you have changed the signature of resolveLinkFile - it's good.
I said "So the 2nd instruction could be removed IMO.", I mean:
instead of 
```cpp
				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
#ifdef UNICODE
				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
#else
				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::toupper);
#endif
```

use 

```cpp
				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
```

UNICODE is defined in Notepad++ VCPROJ file, TCHAR is always wchar_t - also the generic symbols TCHAR & generic_string will be replaced progressively by wchar_t & wstring respectively.
Since a lot of effort was taken to remove `#ifdef UNICODE` in the past, I don't think it's wise to introduce again such symbol.

donho (changes requested):
Please use 

```cpp
				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
```

instead of 
```cpp
				// upperize because the std::find is case sensitive unlike the Windows OS filesystem
#ifdef UNICODE
				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::towupper);
#else
				std::transform(fileNameOnly.begin(), fileNameOnly.end(), fileNameOnly.begin(), ::toupper);
#endif
```

PowerEditor/src/MISC/Common/Common.cpp
xomx: @donho 
Done.
Sorry, I was not sure before and did not want to bother you with another question…

donho: @xomx 
No problem. Look at the bright side, you have pushed me to do the PR #15371 (not finished yet) :)

donho (approved):

PR #15366, opened 2024-06-26
Update english.xml

Need74:
Good morning,
This modification of the english.xml file aims to allow better translation for other languages ​​by preferring more commonly accepted terms. I also added a certain number of empty lines to harmonize the file in its structure and ease of reading. Some terms were incomplete or confusing and should therefore be removed. I'll stay tuned to talk about this simple proposition. Sincerely
Dennis

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -13,11 +13,11 @@ Translation note:
 					<Item menuId="file" name="&amp;File"/>
 					<Item menuId="edit" name="&amp;Edit"/>
 					<Item menuId="search" name="&amp;Search"/>
-					<Item menuId="view" name="&amp;View"/>
+					<Item menuId="view" name="&amp;Display"/><!-- Because the term View is also used in the Menu "Move to Other View" and "Clone to Other View" -->
 					<Item menuId="encoding" name="E&amp;ncoding"/>
 					<Item menuId="language" name="&amp;Language"/>
 					<Item menuId="settings" name="Se&amp;ttings"/>
-					<Item menuId="tools" name="To&amp;ols"/>
+					<Item menuId="tools" name="To&amp;ols"/> and
 					<Item menuId="macro" name="&amp;Macro"/>
 					<Item menuId="run" name="&amp;Run"/>
 					<Item menuId="Plugins" name="&amp;Plugins"/>
@@ -26,7 +26,7 @@ Translation note:
 				<!-- Sub Menu Entries -->
 				<SubEntries>
 					<Item subMenuId="file-openFolder" name="Open Containing &amp;Folder"/>
-					<Item subMenuId="file-closeMore" name="Close &amp;Multiple Documents"/>
+					<Item subMenuId="file-closeMore" name="Close &amp;Multiple Files"/>
 					<Item subMenuId="file-recentFiles" name="&amp;Recent Files"/>
 					<Item subMenuId="edit-insert" name="Insert"/>
 					<Item subMenuId="edit-copyToClipboard" name="Cop&amp;y to Clipboard"/>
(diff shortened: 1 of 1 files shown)


PowerEditor/installer/nativeLang/english.xml
```diff
@@ -13,11 +13,11 @@ Translation note:
 					<Item menuId="file" name="&amp;File"/>
 					<Item menuId="edit" name="&amp;Edit"/>
 					<Item menuId="search" name="&amp;Search"/>
-					<Item menuId="view" name="&amp;View"/>
+					<Item menuId="view" name="&amp;Display"/><!-- Because the term View is also used in the Menu "Move to Other View" and "Clone to Other View" -->
 					<Item menuId="encoding" name="E&amp;ncoding"/>
 					<Item menuId="language" name="&amp;Language"/>
 					<Item menuId="settings" name="Se&amp;ttings"/>
-					<Item menuId="tools" name="To&amp;ols"/>
+					<Item menuId="tools" name="To&amp;ols"/> and
```
donho, line 20: It shouldn't be there.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -13,11 +13,11 @@ Translation note:
 					<Item menuId="file" name="&amp;File"/>
 					<Item menuId="edit" name="&amp;Edit"/>
 					<Item menuId="search" name="&amp;Search"/>
-					<Item menuId="view" name="&amp;View"/>
+					<Item menuId="view" name="&amp;Display"/><!-- Because the term View is also used in the Menu "Move to Other View" and "Clone to Other View" -->
```
donho, line 16: The term "View" is a convention of Windows applications, and used by the major software under Windows.

donho (commented):
Good suggestions for some terms. However users have become accustomed to the use of these terms.
I'll consider some of your suggestions and do the modification in english.xml myself.

PowerEditor/installer/nativeLang/english.xml
Need74: For this point it was an error on my part and I subsequently deleted it.

PowerEditor/installer/nativeLang/english.xml
Need74: > The term "View" is a convention of Windows applications, and used by the major software under Windows.

I am well aware of this and if I made this proposal it is just to be able to retain the use of the term view for the right zone and left zone which are called views.

I think that it is in fact this name that a change should be made.
I suggest talking about display area where appropriate.

The lines concerned would be:
328 <Item id="44072" name="Focus on Another View"/>
351 <Item id="10001" name="Move to Other View"/>
352 <Item id="10002" name="Clone to Other View"/>
444 <Item CMDID="10001" name="Move to Other View"/>
445 <Item CMDID="10002" name="Clone to Other View"/>

They could become:
328 <Item id="44072" name="Focus on Another Display Area"/>
351 <Item id="10001" name="Move to Other Display Area"/>
352 <Item id="10002" name="Clone to Other Display Area"/>
444 <Item CMDID="10001" name="Move to Other Display Area"/>
445 <Item CMDID="10002" name="Clone to Other Display Area"/>

This is only a suggestion but it seems to me that it would provide better clarity when using the software.

PowerEditor/installer/nativeLang/english.xml
Need74: > It shouldn't be there.

For this point it was an error on my part and I subsequently deleted it.

Need74:
Thank you for taking the time to read and understand the meaning of my PR.
It seems obvious to me that modifying the "English language master" file on which all translations depend must be done with the greatest care.
I don't know if you understood it, but the multiple line breaks I made were intended to make it easier to read the file in text mode and to better identify the beginnings and ends of main sections.
I remain at your disposal for any further information and thank you for your involvement in the world of free software.
Sincerely
Dennis

PR #15371, opened 2024-06-28
Remove ambiguous symbols - TCHAR & generic_string

donho:
The "ambiguous symbols (TCHAR & generic_string) are used when Notepad++ suppoted both ANSI & Unicode formats. Since ANSI format of Notepad++ has been dropped (since long time ago), TCHAR & generic_string will be replaced by wchar_t & wstring respectively.

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -29,6 +29,8 @@
 #include <Parameters.h>
 #include "Buffer.h"
 
+using namespace std;
+
 void printInt(int int2print)
 {
 	wchar_t str[32];
@@ -42,7 +44,7 @@ void printStr(const wchar_t *str2print)
 	::MessageBox(NULL, str2print, TEXT(""), MB_OK);
 }
 
-generic_string commafyInt(size_t n)
+wstring commafyInt(size_t n)
 {
 	generic_stringstream ss;
 	ss.imbue(std::locale(""));
@@ -89,9 +91,9 @@ char getDriveLetter()
 }
 
 
-generic_string relativeFilePathToFullFilePath(const wchar_t *relativeFilePath)
+wstring relativeFilePathToFullFilePath(const wchar_t *relativeFilePath)
 {
-	generic_string fullFilePathName;
+	wstring fullFilePathName;
 	wchar_t fullFileName[MAX_PATH];
 	BOOL isRelative = ::PathIsRelative(relativeFilePath);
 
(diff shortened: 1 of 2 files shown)


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -31,13 +31,13 @@
 
 void printInt(int int2print)
 {
-	TCHAR str[32];
+	wchar_t str[32];
 	wsprintf(str, TEXT("%d"), int2print);
 	::MessageBox(NULL, str, TEXT(""), MB_OK);
```
xomx, line 38: now it should be:
```
	wsprintfW(str, L"%d", int2print);
	::MessageBoxW(NULL, str, L"", MB_OK);
```

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -31,13 +31,13 @@
...
+	wchar_t str[32];
 	wsprintf(str, TEXT("%d"), int2print);
 	::MessageBox(NULL, str, TEXT(""), MB_OK);
 }
 
 
-void printStr(const TCHAR *str2print)
+void printStr(const wchar_t *str2print)
 {
 	::MessageBox(NULL, str2print, TEXT(""), MB_OK);
```
xomx, line 44: ::MessageBoxW(NULL, str2print, L"", MB_OK);

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -50,7 +50,7 @@ generic_string commafyInt(size_t n)
 	return ss.str();
 }
 
-std::string getFileContent(const TCHAR *file2read)
+std::string getFileContent(const wchar_t *file2read)
 {
 	if (!::PathFileExists(file2read))
```
xomx, line 57: PathFileExistsW

@donho 
I will not continue in this review, there are so many problems of this type.
E.g. the nearby line 63 should be: `FILE *fp = _wfopen(file2read, L"rb");`

xomx (commented):
@donho 
This PR is not completed.
- the `TEXT("")` and `_T('')` should be replaced by the `L""`.
- A- and T-WINAPIs replaced by the W-variants (`MessageBoxA` and `MessageBox` -> `MessageBoxW`) where appropriate
- ditto A-func and T-funcs with W-funcs ( `fopen` and `_tfopen` -> `_wfopen` )

PowerEditor/src/MISC/Common/Common.cpp
donho: @xomx 
Agree for the TEXT("") and _T('') being replaced by the L"", to make the intention (semantic) of using char or wchar_t more clear. However, adding suffix "W" for all the win32 functions is not necessary to me.

PR #15379, opened 2024-06-30
Remove ambiguous symbols (part 5)

donho:
Relace TCHAR, generic_string & TEXT("") par wchar_t, wstring & L"" respectively.

Follow up: 94af271

### DIFF
--- a/PowerEditor/src/Notepad_plus_Window.cpp
+++ b/PowerEditor/src/Notepad_plus_Window.cpp
@@ -18,7 +18,7 @@
 #include <shlwapi.h>
 #include "Notepad_plus_Window.h"
 
-const TCHAR Notepad_plus_Window::_className[32] = TEXT("Notepad++");
+const wchar_t Notepad_plus_Window::_className[32] = L"Notepad++";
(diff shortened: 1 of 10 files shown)


PowerEditor/src/Notepad_plus_Window.cpp
```diff
@@ -18,7 +18,7 @@
 #include <shlwapi.h>
 #include "Notepad_plus_Window.h"
 
-const TCHAR Notepad_plus_Window::_className[32] = TEXT("Notepad++");
+const wchar_t Notepad_plus_Window::_className[32] = L"Notepad++";
```
xomx, line 21: @donho 
Now (C++11+) you can remove this line from the .cpp and initialize directly in the Notepad_plus_Window.h:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/4d8ef955e28e2164224946261e5dcdb00f313f87/PowerEditor/src/Notepad_plus_Window.h#L112
`static constexpr wchar_t _className[] = TEXT("Notepad++");`

PR #15387, opened 2024-07-01
Allow to reset custom tone colors to other default colors

### LINKED ISSUE #15055: [Feature Request] - Ability to pre-populate the predefined color sets to the list of Tones

### Description of the Issue

In **Preferences > Dark Mode**, when defining the Customized Tones instead of using one of the predefined Dark Mode color sets, the **Reset** will take it back to the default Dark Mode tones.  It would be really helpful if you could start from one of the predefined color sets, instead of just starting from the default.

For example, if I mostly like "Olive", but want to change the "Error" color, right now I would have to Customize, trying to figure out what the "Olive" colors are first (or dig into [source code](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c823ca815056410a255ba7e62e870298c4be0968/PowerEditor/src/NppDarkMode.cpp#L179-L305)) and set those to match what "Olive" uses, and _then_ customize the one color I actually wanted to change from "Olive"'s color scheme.  Being able to populate from "olive" to "custom" would allow easier tweaking of the Tones.

I can think of two ways to implement it:

1. If you start from (for example) "Olive", when you enable "Customized", it could pre-populate the tones from the "Olive", instead of starting from the tones that match "Black"
2. Normal-Click **Reset** would reset to "Black"-based tones; Right-Click **Reset** would allow picking which of the preset color-sets to start from (so Right-Click **Reset** then "Olive" would show all the Tones that match the "Olive" choice"

the second is probably better, because that will avoid losing your previous customized set every time you pick one of the predefined colors (so it would allow easily switching back and forth between Named and Custom without resetting the Custom colors)

(If #13255 is implemented -- which it should be -- then the Right-Click-**Reset** list should include Light as one of the drop-down color sets to choose from)

### Debug Information

```shell
Notepad++ v8.6.5   (64-bit)
Path : C:\usr\local\apps\npp\npp.8.6.5.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 11 Home (64-bit
OS Version : 23H2
OS Build : 22631.3447
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

ozone10:
fix #15055


ozone10:
image

ozone10:
msys failure is from master

ArkadiuszMichalski:
@ozone10 The `Reset` label can also be translated?

ozone10:
@ArkadiuszMichalski 
PR is reusing existing translations.
image

rddim:
Is it possible to add a check mark for the selected tone? The default should be on `Black`

npp_check_mark

ozone10:
@rddim 
For current implementation it does not makes sense. Reset button will always reset to black. Dropdown menu will reset to specific color. It could be possible to make it behave similarly to find dialog swap button, but how many times user needs to reset color.

PR #15388, opened 2024-07-01
Fix comiling error in msys2

donho:


PowerEditor/src/MISC/sha1/sha1.h
```diff
@@ -154,22 +153,20 @@
...
-#define _T(__x) (__x)
-#define _tmain main
-#define _tprintf printf
-#define _getts gets
-#define _tcslen strlen
-#define _tfopen fopen
-#define _tcscpy strcpy
-#define _tcscat strcat
-#define _sntprintf snprintf
+#define TCHAR wchar_t
```
ozone10, line 156: while these mappings fix compilation for msys2 gcc builds, won't it be better to remove this code block and use those string functions directly (`_snwprintf`, `wcscpy`, ...)


https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c2da907f349ec5decf6fb3e641b0bcb72a82fb78/PowerEditor/src/MISC/sha1/sha1.cpp#L148

to
`FILE* fpIn = _wfopen(tszFileName, L"rb");`

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/c2da907f349ec5decf6fb3e641b0bcb72a82fb78/PowerEditor/src/MISC/sha1/sha1.cpp#L206-L238

to
```c++
bool CSHA1::ReportHash(wchar_t* tszReport, REPORT_TYPE rtReportType) const
{
	if(tszReport == NULL) return false;

	wchar_t tszTemp[16]{};

	if((rtReportType == REPORT_HEX) || (rtReportType == REPORT_HEX_SHORT))
	{
		_snwprintf(tszTemp, 15, L"%02X", m_digest[0]);
		wcscpy(tszReport, tszTemp);

		const wchar_t* lpFmt = ((rtReportType == REPORT_HEX) ? L" %02X" : L"%02X");
		for(size_t i = 1; i < 20; ++i)
		{
			_snwprintf(tszTemp, 15, lpFmt, m_digest[i]);
			wcscat(tszReport, tszTemp);
		}
	}
	else if(rtReportType == REPORT_DIGIT)
	{
		_snwprintf(tszTemp, 15, L"%u", m_digest[0]);
		wcscpy(tszReport, tszTemp);

		for(size_t i = 1; i < 20; ++i)
		{
			_snwprintf(tszTemp, 15, L" %u", m_digest[i]);
			wcscat(tszReport, tszTemp);
		}
	}
	else return false;

	return true;
}
```

PR #15389, opened 2024-07-01
Refactor DoubleBuffer through subclassing, fix bugs, add comments

### LINKED ISSUE #15363: [BUG] DoubleBuffer beginPaint assert in Debug version

### Description of the Issue

MS VC++ Runtime Library shows a failed assert message when the N++ is resized as is shown on this pic:
npp-DoubleBuffer-Debug-assert

### Steps To Reproduce

1.  Launch N++.
2.  Resize to some minimal size.
3.  Assert.

### Current Behavior

Assert.

### Expected Behavior

Even the Debug N++ build should not assert.

### Debug Information

```shell
v8.6.9devel
```

### Anything else?

_No response_

yirkha:
This is a follow-up to my previous efforts with double buffering for status bar, tab bar etc.

- This PR mostly reverts previous changes in `TabBar` and `StatusBar` and makes them just implement WM_PAINT and WM_PRINTCLIENT as usual.
- Double buffering is provided through a transparent subclassing class with automatic lifetime, instead of the previous intrusive implementation that required adding a member field and changing WM_PAINT implementation.

⇒ The whole double buffering can be easily turned off by commenting out one line for a control like the status bar, or globally for the whole app by commenting/ifdeffing out `DoubleBuffer::subclass()` now. This should make it safer for any future regressions.

- WM_ERASEBKGND and clipping rectangle is handled properly now (the issue that was worked around by #15319)
- The code is protected against empty paint rectangle (fixes #15363)
- There are more explanatory comments to make it more maintainable

----

It gets a bit tedious to test all the scenarios (add tabs, remove tabs, resize the window, move/drag tabs around, move between tabs (clicking the arrows - single row), jump between tabs (ctrl-tab or similar), scroll the tabs with mouse wheel, find dialog status bar), and variants (light mode / dark mode, single row / multi row, horizontal / vertical, standard / reduced, ...), but I've played with everything possible for a while and I've not seen anything new broken. Fingers crossed.

Please check, @ozone10, @xomx 
Also in general, if this more detached/separatable kind of subclassing the windows is better from your pov.
@donho too, of course.
Thanks


xomx:
@yirkha 
I checked the N++ tab-bar & status-bar behavior and so far so good, the previously reported problems by @ozone10 are not there, good job. The Debug assert bug #15363 is also fixed.

Now when I know about it, I cannot stop focusing on the N++ panels (like the Clipboard History) flickering while resizing them :-)
So that could be another suitable target for some future patch.

PowerEditor/src/WinControls/DoubleBuffer/DoubleBuffer.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include <CommCtrl.h>
+#include <cassert>
+#include <memory>
+
+void DoubleBuffer::subclass(HWND hWnd)
+{
+	std::unique_ptr<DoubleBuffer> self(new DoubleBuffer);
+	if (::SetWindowSubclass(hWnd, s_subclassWndProc, 0, reinterpret_cast<DWORD_PTR>(self.get())))
+	{
+		self.release();
```
xomx, line 27: 👍  Clever use of the unique_ptr.

If the subclassion succeeds, the released unmanaged memory (new DoubleBuffer object) will be destroyed at the DoubleBuffer::subclassWndProc -> WM_NCDESTROY time later.

donho:
@yirkha 
So if I do understand, the PR is rather a refactoring (ie. no bug fix & new enhancement)? If yes, why? For the future enhancement fot light mode maybe?

donho (dismissed):
Remove SetWindowSubclass call for not modifying the retured value and keep the legacy windows subclass way.

xomx:
@donho
> So if I do understand, the PR is rather a refactoring (ie. no bug fix & new enhancement)?

This was not for me but I try to answer that.

Enhancement:
- https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15319#issuecomment-2176498554   (see the PR description - "Double buffering is provided through a transparent subclassing class with automatic lifetime...")

Bugs (the last two items were fixed by the patches from @ozone10 but if I understand it correctly, they were "interim" master branch patches to the less-than-ideal original DoubleBuffer version:
- #15363
- #15319
-  #15320

yirkha:
Yes, this has a few **bug fixes** (the assert in #15363, DC clip rectangle and other state saving for what was worked around in #15319, proper usage of WM_PRINT with a PRF_ERASEBKGND flag to avoid background issues like in #15320, or more error checking around Win32 APIs to handle exceptional cases better).

__________

It also brings other **improvements**. Most importantly, the _whole_ logic of double buffering is moved to the separate class now. Previously some parts of it were unnecessarily "leaking" out to the individual controls, resulting in repeated code. This was already annoying with just the Status bar and the Tab bar.

The other improvement is utilizing the [modern subclassing](https://learn.microsoft.com/en-us/windows/win32/controls/subclassing-overview) as recommended by @ozone10 (well, and Microsoft), as it handles potentially multiple subclassing and other edge cases internally.

Unfortunately, I did not have enough foresight when I came with the original double buffering patch, so there are some reverts needed in StatusBar.cpp and TabBar.cpp. But compared to the original state of those files >1 month ago, there would be effectively fewer changes now. And the same would be true for any other control that might need buffering in the future like this.

yirkha:
> Remove SetWindowSubclass call for not modifying the retured value and keep the legacy windows subclass way.

You mean `SetWindowLongPtr(GWL_WNDPROC, ...)`, right? Where should the state be saved then, though, `GWL_USERDATA` already belongs to someone else. `SetProp()` with an atom?

donho:
@xomx @yirkha 
Thank you for your reply. Very clear & useful. 

> You mean SetWindowLongPtr(GWL_WNDPROC, ...), right?

Yes, but please hold on - I've miss-understood what @ozone10 has explained about the SetWindowSubclass.
Once the PR #15394 is validated and merged, I'll be back to this PR

PowerEditor/src/WinControls/DoubleBuffer/DoubleBuffer.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+		_hBufferBitmap = ::CreateCompatibleBitmap(hOriginalDC, _size.cx, _size.cy);
+		if (_hBufferBitmap)
+		{
+			_hDefaultBitmap = static_cast<HBITMAP>(::SelectObject(_hMemoryDC, _hBufferBitmap));
+		}
+	}
+}
+
+LRESULT CALLBACK DoubleBuffer::s_subclassWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, UINT_PTR, DWORD_PTR dwRefData)
```
donho, line 70: By prefix "s_" in s_subclassWndProc, what does it mean?

PowerEditor/src/WinControls/StatusBar/StatusBar.cpp
```diff
@@ -109,26 +108,28 @@ static LRESULT CALLBACK StatusBarSubclass(HWND hWnd, UINT uMsg, WPARAM wParam, L
...
-			return FALSE;
+			if (!NppDarkMode::isEnabled())
+			{
+				break;  // Let the control paint background the default way
+			}
+
+			RECT rc{};
+			::GetClientRect(hWnd, &rc);
+			::FillRect(reinterpret_cast<HDC>(wParam), &rc, NppDarkMode::getBackgroundBrush());
+			return TRUE;
```
donho, line 119: @yirkha 
It should return 0 according the doc:
https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-erasebkgnd

In PR #15394, @ozone10 has explained as following:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15394#issuecomment-2235347525 

So it's about a Windows procedure, its return value should follow the online document mentioned above.

PowerEditor/src/WinControls/TabBar/TabBar.cpp
```diff
@@ -962,30 +965,30 @@ LRESULT TabBarPlus::runProc(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lPara
...
-			return FALSE;
+			if (!NppDarkMode::isEnabled())
+			{
+				break;	// Let the control paint background the default way
+			}
+
+			RECT rc{};
+			::GetClientRect(hwnd, &rc);
+			::FillRect(reinterpret_cast<HDC>(wParam), &rc, NppDarkMode::getDarkerBackgroundBrush());
+			return TRUE;
```
donho, line 976: It's about a Windows procedure, its return value should follow the online document mentioned above. (return 0)

donho (dismissed):
Please check the return value.

PowerEditor/src/WinControls/StatusBar/StatusBar.cpp
ozone10: @donho 
`WM_ERASEBKGND` has specific returns.
>An application should return nonzero if it erases the background; otherwise, it should return zero.
It's okay here, we erase so it should be nonzero (`TRUE`).

PowerEditor/src/WinControls/DoubleBuffer/DoubleBuffer.cpp
yirkha: Yes, correct

yirkha:
Thanks for the review & merge. Indeed, panels have come up before to be the next inline, so I'll see what's up with them now.

PR #15394, opened 2024-07-02
Add initial dpi support for panel tab and caption

ozone10:
- use recommended subclassing for docking panel tab and caption
- remove unused function

ref #14959


ozone10:
known issue: tab icons won't be sharp when moving to screen with different dpi

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
```diff
@@ -952,18 +974,18 @@ LRESULT DockingCont::runProcTab(HWND hwnd, UINT Message, WPARAM wParam, LPARAM l
...
-			return TRUE;
+			return 0;
 		}
 
 		case WM_MOUSELEAVE:
 		{
 			toolTip.destroy();
 			_bTabTTHover = FALSE;
-			return TRUE;
+			return 0;
```
donho, line 988: The above return values are changed:
1. from BOOL to int
2. from TRUE to 0 & FALSE to 1 - they are reversed

Is it about unintentional error?

donho (changes requested):
If it's due to unintentional error, please restore all returned BOOL value.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
ozone10: @donho 
we are subclassing to window procedure which use "reversed" values to dialog procedure

donho: @ozone10 
Could you elaborate please?

ozone10: @donho 
[SetWindowSubclass](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-setwindowsubclass) will install window procedure, which is different to dialog box procedure. 
Dialog box procedure returns for most messages TRUE if it processes a message or FALSE if it does not.

[Control-Color Messages](https://learn.microsoft.com/en-us/windows/win32/dlgbox/dlgbox-programming-considerations#control-color-messages) handling in dialog box procedure is main hint to differentiate procedures.

When returning FALSE from not subclassed dialog, default coloring is used (message is ignored), but if subclassed, controls will be colored in black since FALSE == NULL, meaning returned HBRUSH is nullptr which is invalid.

donho: > [SetWindowSubclass](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-setwindowsubclass) will install window procedure, which is different to dialog box procedure.
Dialog box procedure returns for most messages TRUE if it processes a message or FALSE if it does not.

In this case, I prefer to not use SetWindowSubclass - Let's not modify the return value and keep the legacy windows subclass way.

donho (changes requested):
Remove SetWindowSubclass call for not modifying the retured value and keep the legacy windows subclass way.

ozone10:
@donho 
the used legacy subclassing is using `GWLP_WNDPROC` which is also installing window procedure. Also on controls (button, tab) which are not dialog. It makes no sense to return dialog procedure values in window procedure. I am even surprised there is no issue with this.

donho:
> the used legacy subclassing is using `GWLP_WNDPROC` which is also installing window procedure. Also on controls (button, tab) which are not dialog. It makes no sense to return dialog procedure values in window procedure. I am even surprised there is no issue with this.

According the MS online document, for some messages they should return 0, if the messages are processed.
So what you have said:
> we are subclassing to window procedure which use "reversed" values to dialog procedure

is not accurate, or I did miss-understand you.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
```diff
@@ -918,7 +940,7 @@ LRESULT DockingCont::runProcTab(HWND hwnd, UINT Message, WPARAM wParam, LPARAM l
 						tcItem.mask		= TCIF_PARAM;
 						::SendMessage(hwnd, TCM_GETITEM, iItem, reinterpret_cast<LPARAM>(&tcItem));
 						if (!tcItem.lParam)
-							return FALSE;
+							return 1;
```
donho, line 943: In online document of WM_MOUSMOVE:
https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousemove
say:
```
Return value
If an application processes this message, it should return zero.
```
So IMO it should return 0 anyway.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
```diff
@@ -952,18 +974,18 @@ LRESULT DockingCont::runProcTab(HWND hwnd, UINT Message, WPARAM wParam, LPARAM l
 			tcItem.mask		= TCIF_PARAM;
 			::SendMessage(hwnd, TCM_GETITEM, iItem, reinterpret_cast<LPARAM>(&tcItem));
 			if (!tcItem.lParam)
-				return FALSE;
+				return 1;
```
donho, line 977: IMO it should return 0 anyway.

ozone10:
@ozone10 
"reversed" is in quotation, this is maybe oversimplification from me. I should have been more specific. If message is in dialog procedure processed it should return TRUE (which is 1), if not then FALSE (0). But in window procedure, as you have said, if message is processed it should usually return 0, the correct return value is usually in documentation but only for window procedure.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
```diff
@@ -1227,7 +1265,52 @@ intptr_t CALLBACK DockingCont::run_dlgProc(UINT Message, WPARAM wParam, LPARAM l
...
+			if (Message == WM_DPICHANGED)
+			{
+				_dpiManager.setPositionDpi(lParam, _hSelf);
+			}
+			else
+			{
+				onSize();
+			}
+
+			return TRUE;
```
donho, line 1310: For [WM_DPICHANGED_AFTERPARENT](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged-afterparent):
```
Return value
This value is unused and ignored by the system.
```
So it doesn't matter the processing this message returns TRUE.
However,  since [WM_DPICHANGED ](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged)use the same return value, there's a problem to me:
```
Return value
If an application processes this message, it should return zero.
```

So I guess it should be `return TRUE;`

donho (changes requested):

donho:
> If message is in dialog procedure processed it should return TRUE (which is 1), if not then FALSE (0). But in window procedure, as you have said, if message is processed it should usually return 0, the correct return value is usually in documentation but only for window procedure.

@ozone10 
Interesting. Just found this info:
```
A dialog box procedure is similar to a window procedure in that the system sends messages to the procedure when it has information to give or tasks to carry out. Unlike a window procedure, a dialog box procedure never calls the [DefWindowProc](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-defwindowproca) function. Instead, it returns TRUE if it processes a message or FALSE if it does not.
```
https://learn.microsoft.com/en-us/windows/win32/dlgbox/dlgbox-programming-considerations

donho:
@ozone10 
> If message is in dialog procedure processed it should return TRUE (which is 1), if not then FALSE (0). 

So the most of `return TRUE;` that you modified to `return 0;` should be remained, since `LRESULT DockingCont::runProcTab(HWND hwnd, UINT Message, WPARAM wParam, LPARAM lParam)` is a dialog procedure.

Or am I missing some points?

ozone10:
@donho 
`DockingCont::runProcTab` was always window procedure, legacy subclassing is using `GWLP_WNDPROC` and not `DWLP_DLGPROC`. Also tab control is not dialog anyway.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
ozone10: `DockingCont::run_dlgProc` is dialog procedure. `WM_DPICHANGED` is processed so it should return `TRUE`.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
ozone10: This should be `break;` instead, to pass it to default window processing.

PowerEditor/src/WinControls/DockingWnd/DockingCont.cpp
ozone10: This should be `break;` instead, to pass it to default window processing.

ozone10:
@donho 
here is summary.
Dialog procedure returns `TRUE` (for most messages) if it processes message, otherwise it's `FALSE`.
If message has some special return value unless specified for dialog procedure (e.g. `WM_CTLCOLOR*`) it should use `SetWindowLongPtr` with `DWLP_MSGRESULT`.

Dialog procedure should be used only for dialogs.

Window procedure returns `0` (for most messages) if it processes message, otherwise it should use `break` to leave it to default window procedure.
If message has some special return value it can return it directly.

Window procedure can be used for everything, even dialogs.

The main purpose of using new subclassing method is to allow multiple subclassing and stability (to avoid using `GWLP_USERDATA`).

donho (approved):

donho:
@ozone10 
Thank you for the summary.
For the change of return values in this PR (from TRUE to 0), I miss-understood as that SetWindowSubclass requires "reverse" return values. It turns out these return values in window procedure were wrongly coded.

Sorry for the bad change request. Feel free to change the bad return values if you see ones in your future PRs.

donho:
@ozone10 
A crash has been detected: while drag&drop one of the docked views (my case is functionList) out by grabbing the tab, Notepad++ crashes. 
The crash is not reproduced after switching to the master. However, I cannot reproduce the crash after switching back to this PR.

Could you try to reproduce the crash please?

ozone10:
@donho 
no I cannot reproduce it at all.

donho:
With sevevral tries I still cannot reproduce the crash.
The PR will be merged and we will see.

xomx:
@donho 
> A crash has been detected: while drag&drop one of the docked views (my case is functionList) **out** by grabbing the tab, Notepad++ crashes.

When I was working on the docked/floating panel issues, I also occasionally got such hardly reproducible crash. I found there one Docking Manager heap buffer overrun (has not been fixed yet) which could be connected to this. I will try to get back there and finish my previous unsuccessful work on the issue 14478.

donho:
@xomx 
Thank you for your note.
It's indeed the crash not necessarily due to this PR. Thought it's the first time I've got dockable window's crash, I change rarely the position of dockable windows - the crash could have been there since long time ago.

PR #15401, opened 2024-07-04
[xml] Update Bulgarian localization

rddim:
* Fix `Add message box with information about disabled backward regex searching` · notepad-plus-plus/notepad-plus-plus@07e9503


GOLEM777:
New lines in darkMode block:
<Item id="7136" name="Red"/>
<Item id="7137" name="Green"/>
<Item id="7138" name="Blue"/>
<Item id="7139" name="Purple"/>
<Item id="7140" name="Cyan"/>
<Item id="7141" name="Olive"/>

PR #15409, opened 2024-07-07
Fix closing unsaved clone document makes periodic backup lost issue

### LINKED ISSUE #8227: Closing modified clone produces make periodic backup lost
### Description of the Issue

When a file tab is cloned into another view, by definition there is an exact-copy file tab in the opposite view.  When one of the cloned tab is closed by the user, it can be discarded without any save prompting because the file will still be open in a tab in the opposite view, with all data intact.  

Unfortunately, it doesn't happen this way and save prompting occurs.

1. Create a new file or open an existing disk file in Notepad++
2. Modify the file in some way
3. Clone the file (e.g. right-click its tab, choose _Clone to Other View_)
4. Close either of the tabs for this file

### Expected Behavior
I expect the file tab to quietly disappear in the view I've told it to close in; because of its cloned nature nothing can be lost.

### Actual Behavior
I'm prompted to save:
image
I'm not ready to save at this point, but I want the clone closed, so I choose `No`.
The clone closes, which is good, but I should not have been prompted in the first place

Even stranger, if I leave the file unsaved and attempt to quit Notepad++, I get this prompt:
image

Backup preferences are set this way:
image

### Debug Information
Notepad++ v7.8.6   (32-bit)
Path : C:\..........\npp.7.8.6.bin\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS Version : 1809
OS Build : 17763.1039

donho:
Fix #8227

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1020,7 +1020,7 @@ bool Notepad_plus::fileClose(BufferID id, int curView)
 		// Do nothing
 	}
 	else if (buf->isDirty())
-	{	
+	{
 		const wchar_t* fileNamePath = buf->getFullPathName();
 		int res = doSaveOrNot(fileNamePath);
 
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -1036,7 +1036,19 @@ bool Notepad_plus::fileClose(BufferID id, int curView)
 		viewToClose = curView;
 
 	bool isSnapshotMode = NppParameters::getInstance().getNppGUI().isSnapshotMode();
-	doClose(bufferID, viewToClose, isSnapshotMode);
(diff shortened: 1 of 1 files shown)


donho:
@alankilborn 
Could you check it and confirm the issue is fixed by this PR?

alankilborn:
This PR fixes the worst part of the issue.  Thank you.

Do you really want to keep the save prompt when user initiates the close process on a dirty cloned doc?
Nothing is lost by NOT prompting, as the same doc (also dirty) is still in the other view.

I reiterate my point from the issue comments, that user may simply want to end cloned status on his doc, not think about whether or not to save the doc at this time.

donho:
@alankilborn 
Could you check the latest commit to see if it's what you have expected?

alankilborn:
With the latest commit it does meet (my) expectation.  :-)
I like it, good work.

PR #15413, opened 2024-07-09
Implement way for SaveACopyAs to open the copy post-save

### LINKED ISSUE #11861: [Feature Request] Save a Copy As and then OPEN THE COPY

alankilborn:
Fix #11861

Note, the design of this PR is based on some comments in a previously abandoned PR; see https://github.com/notepad-plus-plus/notepad-plus-plus/pull/14742#issuecomment-2087640927


dfs-:
I liked the initial idea after sometimes having the same workflow as described in the issue. That SHIFT+Click _hidden feature_ is a nice touch, as it does not interfere with other users workflow.

alankilborn:
I should have included a visual:

image

---

Note that it also works without the mouse being involved, as the _Save_ button is the default button, so holding Shift while pressing Enter also actuates the proposed new functionality.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -902,6 +927,8 @@ class CustomFileDialog::Impl
 	bool _enableFileTypeCheckbox = false;
 	bool _fileTypeCheckboxValue = false;	// initial value
 	wstring _fileTypeCheckboxLabel;
+	bool _isSavingAsCopy = false;
+	bool _isOpeningTheCopy = false;
```
donho, line 931: IMO, for this feature we need only 1 member variable than 2 bool _isSavingAsCopy & _isOpeningTheCopy, for the reason to make the class class `CustomFileDialog::Impl` simpler.
The 2 bool member variables can be reduced indeed to 1 unsigned char member variable, for 1 feature.
```cpp
unsigned char _savingAsCopyInfo = 0;
```

and use mask SAVE_AS_COPY_DLG (value 0x02) and mask SAVE_AS_COPY_OPEN (value 0x01) to retrieve these 2 booleans when it's necessary.

donho (changes requested):
Squeeze 2 member variables in 1 to make the concerning class interface more trivial & intuitive.

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
alankilborn: OK, I will do it when I return from my most recent trip (extends through 21 July).

donho: Bon voyage @alankilborn !

alankilborn:
@donho 's "Merge branch 'master' into i11861b" messed me up somehow; I'll have to close this and create a new PR.

PR #15476, opened 2024-07-25
Implement means for SaveACopyAs to open the copy post-save

### LINKED ISSUE #11861: [Feature Request] Save a Copy As and then OPEN THE COPY

alankilborn:
Fix #11861

### DIFF
--- a/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
+++ b/PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
@@ -25,6 +25,7 @@
 #include <unordered_map>
 #include "CustomFileDialog.h"
 #include "Parameters.h"
+#include "localization.h"
 
 using namespace std;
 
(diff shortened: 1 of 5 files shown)


MarkusBodensee:
And futher comments there in PR #15413

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
```diff
@@ -902,6 +927,8 @@ class CustomFileDialog::Impl
 	bool _enableFileTypeCheckbox = false;
 	bool _fileTypeCheckboxValue = false;	// initial value
 	wstring _fileTypeCheckboxLabel;
+	bool _isSavingAsCopy = false;
+	bool _isOpeningTheCopy = false;
```
donho, line 931: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15413#discussion_r1678674104

donho (changes requested):
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15413#discussion_r1678674104

PowerEditor/src/WinControls/OpenSaveFileDialog/CustomFileDialog.cpp
alankilborn: Oops, missed deletion.  :-(

alankilborn: Hmm, something bigger went wrong.  :-(

donho (approved):

PR #15499, opened 2024-07-30
disable project warnings

Hurrieam:
https://learn.microsoft.com/zh-cn/cpp/build/reference/vcxproj-files-and-wildcards?view=msvc-170


donho:
@Hurrieam 
What's the point of the first commit?

PR #15517, opened 2024-08-06
Warn user before opening binary files - help address #15470

### LINKED ISSUE #15470: [BUG] Notepad++ chokes on large binary files

### Description of the Issue

Couple of days ago I accidentally opened a 650 MB dat file (i.e. binary) and Notepad++ got stuck in an unresponsive state with one CPU core stuck at 100% forcing me to kill it using Task Manager.

I tried to reproduce that today by creating a 650 MB file filled with zeros (Powershell code):
```
$f = new-object System.IO.FileStream c:\test.dat, Create, ReadWrite
$f.SetLength(650MB)
$f.Close()
```
And it behaved the same &mdash; even with what is apparently not a text file, Notepad++ is stuck seemingly trying to identify the encoding or the file type or to count lines or whatever it is doing wasting 17.33 GB of RAM to load 650 MB file (note that CPU usage is for 24 threads, hence 4.16% because it's single-threaded):

image

### Steps To Reproduce

1.  Create a large empty file
2.  Open with Notepad++
3.  Wait until the file loads
4. Try to close Notepad++
5. Wait until it closes

### Current Behavior

**Unresponsive GUI during loading (1m 45s for 650 MB on my Xeon workstation):**
image

**Unresponsive GUI during closing**
After clicking Close the window becomes unresponsive and grayed out if you try to interact with it further. It finally closes in about 20 seconds, despite no changes to the file being made.

### Expected Behavior

Even if you insist on parsing and displaying the file which is apparently empty and has a wrong extension, it shouldn't use 17.33 GB to render 650 MB of zeros.

image

Also, it shouldn't claim it's a `Normal text file`.

Finally, unresponsive GUI can mean only one thing.

Parsing / processing is done on the UI thread which should never be done &mdash; it should be done on a background thread with a progress bar and it should be cancellable.

### Debug Information

```shell
Notepad++ v8.6.8   (64-bit)
Build time : Jun  4 2024 - 00:30:00
Command Line : "C:\test.dat"
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : OFF
OS Name : Windows 11 Enterprise (64-bit)
OS Version : 23H2
OS Build : 22631.3880
Current ANSI codepage : 1252
Plugins : 
    JsonTools (7.2)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

molsonkiko:
This might help close #15470. The OP in that issue mentioned some performance issues, but those seem a lot less tractable than just giving the user the opportunity to close a jumbo-sized binary file before it eats all their RAM and fills their life with sadness.

The basic idea of this PR is very simple: after a file has been loaded into a Scintilla buffer but before it is rendered, my new code does the following:
1. Look at the file's extension. If the file's extension is on a special whitelist (called `_extensionsToNotCheckIfBinary` in the code, stored in the `fileExtensionsToNotCheckBinary` attribute of the `MISC` tag in `config.xml`), skip step 2 and render the file.
2. If the file "is binary", warn the user that the file is binary. The user has an opportunity to click __No__ and close the file. If they click __Yes__, they have an opportunity to stop showing these warnings for all files with the same extension.
    1. Currently, my code says that a file "is binary" if there are *any* characters with ASCII values `\x00` to `\x08` in the first 10000 characters of the file.
    2. Since this is done after Notepad++ has done its best to decode the file's encoding, this avoids false positives due to things like the empty second bytes of an all-ASCII UTF-16 encoded file. [See this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15470#issuecomment-2261827094) for a list of all the tests I did to determine how well my code avoids false positives.

I am very open to changing the algorithm for determining whether to warn users whether a file is binary. I like a simple approach (in part because I'm not a great C++ programmer), but other approaches might be able to further reduce the false positive rate. That said, I think a fixed "blacklist" of file extensions that *are* binary (like `exe`, `dll`, `pdf`, `jpg`, etc) is a __bad idea__, because (a) we will never think of every binary file extension out there, and (b) this blacklist would add yet another config file or more bloat to the codebase.

As I see it, the main potential downside of this PR (aside from the usual hazards of null dereferences and the like, which I believe I've guarded against) is that people will be turned off by another new thing that throws up modal dialogs. I believe that in most cases, these modal dialogs are an __actively good__ thing, because:
- Even though it should be pretty obvious that editing binary files with Notepad++ is a bad idea, the community forum still regularly gets people who don't realize that text editors are only for editing text documents (yes, yes, hex editor plugins exist, but I haven't found one for NPP that I like)
- Suppose that someone opens a text document, and is warned that this is a binary file. Perhaps they start looking through their file, and realize that it is corrupted with `NUL` characters or the like. Perhaps they go to the community forum and complain, and some regular helps them find the corruption. Perhaps this isn't corruption, they just decided to put control characters in their text file. The worst case scenario (I hope) is that they simply add the file's extension to the whitelist, and aren't bothered by the notifications again.


PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -812,6 +812,107 @@ BufferID FileManager::loadFile(const wchar_t* filename, Document doc, int encodi
...
+}
+
+wstring FileManager::getExtensionsToNotCheckBinary()
+{
+	std::wstring joinedList;
+	stringJoin(_extensionsToNotCheckIfBinary, wstring(L"|"), joinedList);
+	return joinedList;
+}
+
+void FileManager::setExtensionsToNotCheckBinary(const TCHAR* joinedList)
```
donho, line 832: No more macros TCHAR & TEXT("") in the project. Use wchar_t & L"" respectively instead.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -812,6 +812,107 @@ BufferID FileManager::loadFile(const wchar_t* filename, Document doc, int encodi
...
+bool FileManager::isBinary(BufferID id)
+{
+	if (id == BUFFER_INVALID)
+		return false;
+	// figure out if the file contains any characters with ASCII char code less than 9 ('\t')
+	//     (we assume that such files are binary, and thus probably inappropriate to open with Notepad++)
+	// we can also do some tests on the filename (if the ext isn't recognized, maybe it's binary)
+	_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, id->getDocument());
+	_pscratchTilla->execute(SCI_SETSEARCHFLAGS, SCFIND_REGEXP);
+	const TCHAR * binaryRecognizer = TEXT("[\\x00-\\x08]");
```
donho, line 856: TEXT("") is forbiden in the project.

donho:
The PR add a lot of complex logic for not opening the files in some conditions - that's bad IMO since users should be able to open whatever type of file they want without any restriction, and the complex logic makes both source code and binary behaviour difficult to be maintained.

PR #15529, opened 2024-08-09
Python smart indent, do not indent when comment line ends in colon

### LINKED ISSUE #15528: [BUG] With Python auto-indent, lines following a comment line should not be smart-indented

### Description of the Issue

In Python, I often create a comment like the one shown here on line 1:

image

The comment describes what is happening on the next line, so I "form that connection" with a colon (`:`) at the end of the comment line.

Unfortunately, with the advent of Python smart-indent resulting from #15122, the colon at the end of the comment line will cause the following line to be indented when _Enter_ is pressed after the colon:

image

Ideally, since the comment line isn't executable code, the colon at its end should not trigger smart auto-indent on the following line.

### Steps To Reproduce

See Description above.

### Current Behavior

image

### Expected Behavior

image

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.9\npp.8.6.9.portable.x64\notepad++.exe
Command Line : -titleAdd="8.6.9 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 23H2
OS Build : 22631.3880
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    PythonScript (3.0.18)
```

### Anything else?

_No response_

alankilborn:
Fix #15528

Comments definitely welcome on the regular expression change this PR makes!

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3768,7 +3768,8 @@ void Notepad_plus::maintainIndentation(wchar_t ch)
 			auto endPos = _pEditView->execute(SCI_GETLINEENDPOSITION, prevLine);
 			_pEditView->execute(SCI_SETTARGETRANGE, startPos, endPos);
 
-			const char colonExpr[] = ":[ \t]*(#|$)";  // colon optionally followed by only whitespace and/or start-of-comment
+			// colon optionally followed by only whitespace and/or start-of-comment, but NOT on a line that is already a comment
+			const char colonExpr[] = "^[^#]*\\K:[ \t]*(#|$)";
 
 			if (_pEditView->execute(SCI_SEARCHINTARGET, strlen(colonExpr), reinterpret_cast<LPARAM>(colonExpr)) >= 0)
 			{


donho (dismissed):
image
The above case is not fixed.

alankilborn:
> The above case is not fixed.

I think that case is too burdensome to fix -- it would rely on substantial lookback.
Triple-quoted strings are NOT comments, and the comment case is all the issue and PR are attempting to fix.

alankilborn:
@donho Maybe you were too agreeable.  :-)

I was thinking that we could check the "context" of the colon... using SCI_GETSTYLEAT we could tell if the colon is part of a comment or part of a string, and then, if it is, don't do the smart-indent on the following line after Enter is pressed.  If you think this approach has merit, please say so and I'll create a new issue for it.

donho:
Sounds good. But how is it in term of the code? If the code is simple, then OK to me.

alankilborn:
> But how is it in term of the code?

I looked at it.  It should be simple.  When the style context of the colon is "operator", then we can smart-indent the next line.  If the context is something else (which could be "comment", "string", etc), we _don't_ smart-indent the next line.

I'll create a new issue and do the coding + PR.

BTW, we wouldn't need _this_ PR's changes with that new scheme.  You can either revert this PR, or I'll just undo it in the new issue + PR.

donho:
@alankilborn 
No need to revert the PR.
Just create a new PR based on the current master.

PR #15535, opened 2024-08-11
Improve Python smart indent to use style context

### LINKED ISSUE #15534: [Feature request] Improve Python smart-indent conditions by looking at context of colon

### Description of the Issue

Post - #15529 I thought of an improved way to do Python smart-indent:  use Scintilla to detect when a colon at the end of a line is styled as a comment, string, etc.  Basically, require that its style be "operator" as the colon is defined as an operator in the Python language.

Rely on style context for determining next-line smart-indent, not a somewhat complicated regular expression.

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.9\npp.8.6.9.portable.x64\notepad++.exe
Command Line : -titleAdd="8.6.9 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 23H2
OS Build : 22631.3880
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    PythonScript (3.0.18)
```

### Anything else?

_No response_

alankilborn:
Fix #15534

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -3770,7 +3770,10 @@ void Notepad_plus::maintainIndentation(wchar_t ch)
 
 			const char colonExpr[] = ":[ \t]*(#|$)";  // colon optionally followed by only whitespace and/or start-of-comment
 
-			if (_pEditView->execute(SCI_SEARCHINTARGET, strlen(colonExpr), reinterpret_cast<LPARAM>(colonExpr)) >= 0)
+			auto posColon = _pEditView->execute(SCI_SEARCHINTARGET, strlen(colonExpr), reinterpret_cast<LPARAM>(colonExpr));
+
+			// when colon found, additionally check that it is not in a comment, inside a string, etc.
+			if ((posColon >= 0) && (_pEditView->execute(SCI_GETSTYLEINDEXAT, posColon) == SCE_P_OPERATOR))
 			{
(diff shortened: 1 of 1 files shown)


MarkusBodensee:
@alankilborn 
I am a bit confused now after reading through #15529. Didn't you want create a PR based on latest master and revert the Regex because a double check for comments seems unnecessary? 

But looks like you created this PR based on "a previous to #15529 commit to master" version. Or is everything as intended to be?

alankilborn:
@MarkusBodensee is correct -- the `^[^#]*\\K` part of the regular expression doesn't need to be there (and in fact shouldn't be, because it is now confusing).

MarkusBodensee:
@alankilborn @donho
Any plans to fix and go back to the previous regex? Since it looks to me like it was the original suggestion of @donho in #15529.

alankilborn:
>Any plans to fix and go back to the previous regex? 

Yes, I just haven't gotten to it yet. :-)

---

>Since it looks to me like it was the original suggestion of don

I'm missing something here, because I don't think Don suggested anything on this, except he hinted that it should work inside triple-quoted strings, which is something that I did end up implementing (indirectly).

alankilborn:
I got around to it:  https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15574

PR #15537, opened 2024-08-11
Add new SubStyles and their available keyword lists

### LINKED ISSUE #15520: [Feature request] Please give StyleCofigurator access to Lexilla's SubStyles

### Description of the Issue

Over the years, I have seen multiple requests in the forum for adding new keyword lists with their own colors to Notepad++.

Here are a selection I found from the last 3 years:

- [PHP](https://community.notepad-plus-plus.org/topic/25980/highlighting-with-self-created-words-in-langs-xml-does-not-work)
- [PHP](https://community.notepad-plus-plus.org/topic/24054/can-t-define-php-color-for-functions)
- [HTML](https://community.notepad-plus-plus.org/topic/25047/how-to-add-keywords-for-the-html-syntax-highlighter)
- [HTML](https://community.notepad-plus-plus.org/topic/21646/modifying-existing-language-settings)
- [CPP](https://community.notepad-plus-plus.org/topic/24228/can-t-make-default-c-style-to-highlight-custom-macros)
- [CPP](https://community.notepad-plus-plus.org/topic/22324/modify-the-c-xml-language-file-to-include-math-functions)
- [JSON](https://community.notepad-plus-plus.org/topic/22250/how-to-add-custom-color-to-a-tag-in-json-language)
- [any](https://community.notepad-plus-plus.org/topic/25678/syntax-keywords)

There have also historically been some issues requesting the enabling of additional keyword lists from , some of which were rejected because Scintilla/Lexilla didn't have them allocated, and others that were eventually implemented.  And I found #3063 requesting it for Verilog and other languages (that was closed for "age" in the 2019 issue cleanup) -- so there has definitely been historical interest in having more keyword lists available on various languages.

Up until now, we've been suggesting that requestors use the EnhanceAnyLexer plugin to add a regex for their list of keywords -- but the regex post-processing slow, and isn't optimized for keywords (it's better for pattern matches), and doesn't have the benefit of being in the Style Configurator (which makes it more difficult for users to find and update).

I recently learned about the Scintilla/Lexilla "SubStyles" feature, which allows additional keywords lists (beyond the default keywords-keywords7) for any lexer that has substyles enabled, which includes C/C++ and HTML/PHP. The "substyles" are essentially on-request allocated styles (with their own color/bold/italic/underline style settings) that come with multiple keyword lists (all of the lexilla lexers with substyles enabled allow for up to 64 substyles for a given document).

In [this post](https://community.notepad-plus-plus.org/post/96064), I share a script for the PythonScript plugin that enables the substyles.  But from a user perspective, it's not nearly as nice as just being able to use the Style Configurator to add more keywords into some user-defined keyword lists that are available in the GUI next to all the other keyword lists for a given language.  (Actually, my script's interface for changing the colors or keyword lists is horrible right now.  But I didn't want to put effort into making it prettier or easier if we could add the feature to Notepad++'s core code, instead.)

I'd like Notepad++ to natively enable the Scintilla/Lexilla SubStyles feature on applicable lexers/languages, rather than having to do it from a script or other plugin.

Based on my experiments in PythonScript, and my recent experience with how to enable lexilla lexers, I have some ideas on how I would implement it in the codebase, that would essentially make it look to the user like it's just another keyword list in the Style Configurator.  My thought was to add just 8 substyles to each of the lexer languages that allows them, which should be plenty for most users (64 would be overkill, and would clutter the StyleConfigurator too much, IMO).  (For HTML, I would propose doing 4 custom tag substyles and 4 custom attribute substyles)  But I wouldn't want to spend any time working on a PR to implement this if you were completely against it.

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Path : C:\usr\local\apps\npp\npp.8.6.9.portable.x64\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4651
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

The following lexers (and their associated languages) have SubStyles available

- LexBASH
    - languages: BASH
    - styles: SCE_SH_IDENTIFIER=8, SCE_SH_SCALAR=9
- LexCPP
    - languages: C, CPP, JAVA, JS, RC, CS, FLASH, SWIFT, TYPESCRIPT
    - styles: SCE_C_IDENTIFIER=11, SCE_C_COMMENTDOCKEYWORD=17
- LexGDScript
    - languages: GDScript
    - styles: SCE_GD_IDENTIFIER=11
- LexHTML
    - languages: HTML, XML, PHP, ASP, JSP
    - styles: SCE_H_TAG=1, SCE_H_ATTRIBUTE=3, SCE_HJ_WORD=46, SCE_HJA_WORD=61, SCE_HB_WORD=74, SCE_HP_WORD=96, SCE_HPHP_WORD=121
- LexLua
    - languages: LUA
    - styles: SCE_LUA_IDENTIFIER=11
- LexPython
    - languages: PYTHON
    - styles: SCE_P_IDENTIFIER=11
- LexVerilog
    - languages: Verilog
    - styles: <none>
    - comment: no, I don't know why the LexVerilog enables the substyles feature, but doesn't have them for any styles

Of the 8 requests I listed above, 6 could be solved completely by having the SubStyle feature enabled in Notepad++. Unfortunately, most lexers don't have the feature, so JSON and "any" will not be feasible without Lexilla updates (which isn't going to happen quickly, if at all).  But it would still give a feature to a bunch of people who have been asking for it (and those are normally representative of many other users who don't both logging into the forum or putting in a github issue.

pryrt:
- Enable 8 Scintilla's SubStyles (each), which allow for new keyword lists and styles for the languages with SubStyles available: ActionScript, ASP, Bash, C, C++, C#, GDScript, Go, HTML, Java, JavaScript (standalone and embedded), JSP, Lua, PHP, Python, Resource (RC), Swift, TypeScript, and XML.
- The new SubStyles have been added to `langs.model.xml`, `stylers.model.xml`, and the themes, which means users can just use the StyleConfigurator for the SubStyles, just like they do for the original Styles.

Fixes #15520


pryrt:
@alankilborn messaged me -- he thought it was confusing that Styles that have an empty list in `langs.xml` would still show the empty **Default keywords** box:
image

His suggestion was to hide that box, because users would only be expecting to see the **User-defined keywords** box for these new lists.

I would have no clue how to hide that box "on the fly", but he said he'd try to look into it and figure it out.  In the meantime, he asked me to mark this as "work in progress" rather than "ready for review"

@donho ,

Which of these would you prefer?

1. The current implementation, where **Default keywords** shows up even when empty.  (If so, please let us know, so that Alan can avoid wasting his time digging into it, and I can convert it back to "ready for review".)
2. On the SubStyle-based entries, hide the **Default keywords** entries if `langs.xml` doesn't define default keywords for it.  (But for ones like PHP, where I used some of the substyle# entries for splitting up the 9000-keyword `instre1`, or if any user edits their own `langs.xml`, still show the **Default keywords** if they _do_ populate the "default" values.)
3. On _any_ styler entry, hide the **Default keywords** list if the `langs.xml` doesn't have a value.
4. Something else that I haven't thought of.

donho:
@pryrt 
In you opinion, should people have any way to add Default keywords manually in langs.xml?

pryrt:
@donho , my vote would be for "yes". 

I actually used it to help clean up the PHP keywords (which had > 9000 all in the `instre1`) into separate categories in the langs.xml that I submitted.  

I think we would eventually regret treating it differently to a "normal" keyword list, when the user won't see the technical difference, and the Lexilla team treats it as the official way to get additional keyword lists.  (It would enable, for example, allowing an IT department to distribute their own lands.xml with company-common keywords for a given language.  And give us room, in the future, to rename use one of the substyle entries to distribute additional keywords separately from the already-used keyword slots)

donho:
@pryrt 
> my vote would be for "yes".
I actually used it to help clean up the PHP keywords (which had > 9000 all in the instre1) into separate categories in the langs.xml that I submitted.

For which reason the PHP keywords are needed to separated?

BTW
I've tried PR, and the result of **USER KEYWORDS 1**'s Default keywords in PHP is empty:

image

The path of langs.xml is `%APPDATA%\notepad++\langs.xml`.
Is it normal?
> I think we would eventually regret treating it differently to a "normal" keyword list, when the user won't see the technical difference, and the Lexilla team treats it as the official way to get additional keyword lists....

Sorry, I don't follow you. Could you elaborate it please?

pryrt:
> @pryrt
> For which reason the PHP keywords are needed to separated?

The original [forum question](https://community.notepad-plus-plus.org/topic/25980/highlighting-with-self-created-words-in-langs-xml-does-not-work) which prompted my finding out that SubStyles exist and learning how to implement them was because users want to be able to use a different color for the language constructs (`if`, `for`, etc) compared to the default-library functions-and-methods (`imagecreatefrombmp`, `sin`, `bin2hex`, ... and thousands more)

> BTW I've tried PR, and the result of **USER KEYWORDS 1**'s Default keywords in PHP is empty:
> 
> image
> The path of langs.xml is `%APPDATA%\notepad++\langs.xml`. Is it normal?

That's because for `substyle1`, `substyle2`, and `substyle3`, I called them `PREDEFINED` and `FUNCS AND METHODS 1` and `FUNCS AND METHODS 2`, which are shown in your screenshot.  That's why there's only five ones called `USER KEYWORDS #` in your screenshot, because `substyle4` - `substyle8` weren't pre-populated, but the first three were.

> > I think we would eventually regret treating it differently to a "normal" keyword list, when the user won't see the technical difference, and the Lexilla team treats it as the official way to get additional keyword lists....
> 
> Sorry, I don't follow you. Could you elaborate it please?

From the user perspective, there is currently no difference between the `instre1` list and the `substyle1` list, in terms of how they interact with it from the Style Configurator, or in terms of the fact that once they set the colors and the keyword list, the `substyle#` behave just like the `instre1`.  

As a result, I think it would come as a surprise to them if everything else were the same, but that they could not edit langs.xml list for the `substyle#`-based ones even though they can for the `instre#`- and `type#`-based lists.  I don't see a good reason to add more code to _take away_ that power from the user (or from a user's IT department) when it will work for them as currently implemented, and when it would then make it different in their mind from the other keyword lists.  

The responses I would expect from users, based on this confusion: "Why is this keyword list different than the other keyword list(s) for this language?  Why won't it let me add Default keywords in langs.xml?  If it's a second-class citizen, does that mean it doesn't work as well, or is dangerous to use compared to the other keyword list?  Please make it the same as the other keyword lists.  GIVE ME THE ABILITY TO ADD DEFAULT KEYWORDS IN langs.xml FOR REASON XYZ".  

To sum up my arguments:
- it's more natural to the user to treat both keyword lists identically
- it gives more power to the user right now
- the code is cleaner and less confusing
- it will be easier to give it to them now with the full power than it would be to add in the logic to disable that aspect of SubStyles and then try to remember how to add it back into the code in the future when users ask for it

pryrt:
... To clarify, my arguments about the user-confusion about the differences is only important for your question, "should people have any way to add their default keywords manually in langs.xml, in this feature".  

With respect to Alan's idea, of just hiding the Default box when there aren't any default keywords for a SubStyle, I could see why that would help reduce confusion for the user (so there isn't the extra empty box).  He's shown me where to go in the code to implement the logic to show or hide those GUI elements, so if you wanted me to hide that on SubStyles that don't have defaults, I could do that.  (If the user ever added defaults to langs.xml, then the next time they ran, the Defaults box would show up, so they can see those values.)

alankilborn:
I'm getting named-dropped a lot here (fine), but in short my argument was that I thought when user selects _USER KEYWORDS 1_ he doesn't mind seeing a box titled _User Defined Keywords_, but he's confused when he sees another box called _Default keywords_.

pryrt:
If I had called them `EXTRA KEYWORDS #` instead of `USER KEYWORDS #`, would that have alleviated your original concern?

alankilborn:
> If I had called them `EXTRA KEYWORDS #` instead of `USER KEYWORDS #`, would that have alleviated your original concern?

If that had happened, I probably wouldn't have keyed in on it.
It's fine as is if it doesn't bother anyone else participating in this discussion or reading this thread.

donho:
@pryrt 
> That's because for substyle1, substyle2, and substyle3, I called them PREDEFINED and FUNCS AND METHODS 1 and FUNCS AND METHODS 2, which are shown in your screenshot. That's why there's only five ones called USER KEYWORDS # in your screenshot, because substyle4 - substyle8 weren't pre-populated, but the first three were.

OK, I see.
In this case, I would change the substyle1, substyle2, and substyle3 to substyle6, substyle7, and substyle8 respectively, so if any user want to add their user-defined keywords, they add in substyle1 and they can see their keywords are shown in USER KEYWORDS 1's Default keywords in PHP - that'll reduce a lot of confusion.

I also suggest to add comment for the explanation of user-defined usage in langs.model.xml here:
```xml
<NotepadPlus>
   <!-- The key words of the supported languages, don't touch them! -->
   <!-- For the language C, C++... you can add your own keywords into substyle1..6 ... 
   blablabla... -->
   <Languages>
   ...
```

> On the SubStyle-based entries, hide the Default keywords entries if langs.xml doesn't define default keywords for it. (But for ones like PHP, where I used some of the substyle# entries for splitting up the 9000-keyword instre1, or if any user edits their own langs.xml, still show the Default keywords if they do populate the "default" values.)

There's nothing to hide in this case - if the default keywords could not make user know about this ability, at least it's harmless.

donho (changes requested):
1. change the substyle1, substyle2, and substyle3 to substyle6, substyle7, and substyle8 respectively.
2. Add the comment in langs.model.xml for the explanation.

donho (approved):

PR #15560, opened 2024-08-20
Fix Styler Configurator performance issue

### LINKED ISSUE #12436: Performance issues when using FolderAsWorkspace(FAW) and the Style Configurator dialog.
### Description of the Issue
When using a root directory with > 40000 files, > 2000 directories (~9.5GB) in FAW, Npp freezes on my machine for ~10 seconds for each action executed in the Style Configuration dialog like changing the language or changing style attributes.

1.  Open FAW
2.  add a directory with many files and folders
3.  open style configurator
4.  change to another language

### Expected Behavior
Instant update of the style attributes of the selected language

### Actual Behavior
Npp freezes

### Debug Information
Notepad++ v8.4.6   (64-bit)
Path : D:\Tests\npp\846\x64_2\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 21H2
OS Build : 19044.1889
Current ANSI codepage : 1252
Plugins : none

donho:
Improve Styler Configurator modification application performance considerably.

Fix #12436

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -8563,7 +8563,7 @@ void Notepad_plus::refreshDarkMode(bool resetStyle)
 			else
 			{
 				nppParams.reloadStylers(themePath.c_str());
-				::SendMessage(_pPublicInterface->getHSelf(), WM_UPDATESCINTILLAS, 0, 0);
+				::SendMessage(_pPublicInterface->getHSelf(), WM_UPDATESCINTILLAS, TRUE, 0);
 			}
 		}
 
--- a/PowerEditor/src/Notepad_plus_Window.cpp
+++ b/PowerEditor/src/Notepad_plus_Window.cpp
@@ -287,7 +287,7 @@ void Notepad_plus_Window::init(HINSTANCE hInst, HWND parent, const wchar_t *cmdL
 		{
 			nppGUI._themeName.assign(themePath);
 			nppParams.reloadStylers(themePath.c_str());
-			::SendMessage(_hSelf, WM_UPDATESCINTILLAS, 0, 0);
+			::SendMessage(_hSelf, WM_UPDATESCINTILLAS, TRUE, 0);
 		}
 	}
 
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -2286,6 +2286,8 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case WM_UPDATESCINTILLAS:
 		{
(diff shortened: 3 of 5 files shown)


donho:
@Ekopalypse 
Could you check it please?

Ekopalypse:
@donho, first, thank you for the ongoing effort to make npp better and better.

I tested with the x64 release binary from [here](https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/10463300411/artifacts/1830188979) and a clean portable version of the latest official npp package.

I added two repos to the FAW dialog, notepad-plus-plus and the [windows-rs](https://github.com/microsoft/windows-rs.git) repo.

Starting the latest official version, 6.8.9, with the `-loadingTime` parameter shows a loading time of ~3 seconds and about ~5 seconds for the artifact version.

Changing the default style to the bespin theme took ~20 seconds for the official version and ~22 seconds for the artifact version. Canceling the theme changes took another ~20 seconds for the official version and was instant for the artifact version, BUT the FAW dialog was not reset, it still had the bespin theme.

Switching to a language like Python in the style configurator was instant for both versions, as was changing the attributes, like a different foreground/background color, for that language. It seems that this part has already been solved since I reported the problem.

So, all in all, it seems to be worse than before.

donho:
@Ekopalypse 
Thank you for your testing and the report of result.
The only change in this PR is for Styler Configurator.
So regarding the much longer loading and switching theme time, it should due to the different compiler or the way of compiling, or even some commits after v8.6.9 release (that I doubt a lot), but it has nothing to do with this PR.

The issue #12436 is about the interaction between Styler Configurator & FaW: While FaW is loaded heavily, any action (checkboxes & comboboxes) on Styler Configurator reacts very slowly.

This PR fixes such problem. Could you check it again?

donho:
@Ekopalypse 
Thank you for testing.
> With the exception of GlobalStyles->DefaultStyles, where changing foreground or background color takes time, but only for these two

It's because that the both changing foreground or background color of GlobalStyles->DefaultStyles need to apply also to FaW.

> Undoing changes, pressing the cancel button, works again with the latest artifact build.

It's for the reason that we don't know if FaW should be revert. I'll see what I can do about it.

donho:
> With the exception of GlobalStyles->DefaultStyles, where changing foreground or background color takes time, but only for these two

BTW, not only changing foreground or background color of GlobalStyles->DefaultStyles will to apply to FaW, but also changing the theme via *Select theme* combobox.

> Undoing changes, pressing the cancel button, works again with the latest artifact build.

It's been fixed in the latest commit: only after changing theme or foreground or background color of GlobalStyles->DefaultStyles, the FaW will be applied changing (it'll take time), otherwise there won't be any impact.

ozone10:
@donho 
`SetWindowTheme` has large impact on performance.
PR #15565 should improve performance for some cases.

Issue #1530, opened 2016-02-23
Feature Request: folder as workspace: ignore files using wildcards manually


gdh1995:
I've noticed that Notepad++ has ignored `.git` folder, but this is far less than actual needs.

For example, `*.pyc` should be hidden, and in my workspace `.idea` and `.venv` are also not needed.

I want to be able to:
- On a file/folder's context menu, click a menu item to hide it manually
- Edit a multi-line setting file to decide what files should be hidden
  (just like `.gitignore` but it should use another filename)

pquiring:
I want something like this.  Maybe not a file, but just somewhere in NotePad++'s config you specify extensions of files you don't want listed (*.o *.a *.dll etc).  I just want source files listed.

charleshb:
I do agree! I do **not** want Notepad++ to display DLLs in the listing for Folder As Workspace. If the list of extensions is configurable, that should be great! 
I tried to use the option under Settings -> Preferences -> MISC. -> "Workspace file ext.:". I put this "txt,config,log,xml,js" in that field, but it doesn't seem to have any effect.
Any ideas?

CircuitSerialKiller:
I'd like to see a file filter option too!

matthewparkes:
+1 I would like to hide *.pyc files.

iccor56:
i would really appreciate if you could filter out files too. a list of binary files you don't want to see/open would clean up the workspace alot

Issue #1524, opened 2016-02-21
Broken x64 build since v6.8.9-3-g07ce6c2


linquize:
Broken x64 build since v6.8.9-3-g07ce6c2

NN---:
See  #1525

cmeriaux:
@linquize ,
when you compile x64, you have to compile scintilla in x64. Am I right ? 
How do you compile scintilla in x64 ?

linquize:
You need to patch `scintilla.mak` first
(https://github.com/linquize/notepad-plus-plus/commit/461784b8ae2432bbdc56ad484f9a57f022e8d7e1)

Open VS2015 x64 Command Prompt
```
cd scintilla\win32
nmake -f scintilla.mak
```

cmeriaux:
@linquize  Thanks

Issue #259, opened 2015-06-19
Blinking cursor position is confusing


rook747:
The blinking cursor always stays in front of the "(" or on top of it. 
This is very confusing, any way to fix this ?

![select-cursor-issue-example2](https://cloud.githubusercontent.com/assets/12961985/8247458/a25f7936-1651-11e5-8b3d-ee18bce0d710.gif)
![select-cursor-issue-example1](https://cloud.githubusercontent.com/assets/12961985/8247457/a25d940e-1651-11e5-9665-c42e3a6d35ec.gif)

milipili:
With another font ?

Issue #258, opened 2015-06-18
When NP++ crashes, it can't upload the crash dump file


GregUzelac:
I like to send crash details to developers. So when NP++ crashes I always choose the option to send feedback & dump files. However, NP++ never appears to succeed at this. It displays something about error 5 (ERROR_ACCESS_DENIED??). Suggestions? I'm fluent with the likes of ProcMon, FileMon, and WinDbg.

milipili:
From my point of view, I would prefer a description to reproduce the bug.

PR #15564, opened 2024-08-22
Update to scintilla 5.5.2 & Lexilla 5.4.0

chcg:
Release 5.5.2 ( https://www.scintilla.org/scintilla552.zip )

    Released 21 August 2024.
    Add SCI_SETCOPYSEPARATOR for separator between parts of a multiple selection when copied to the clipboard. Feature #1530.
    Add SCI_GETUNDOSEQUENCE to determine whether an undo sequence is active and its nesting depth.
    Add SCI_STYLESETSTRETCH to support condensed and expanded text styles.
    Add SCI_LINEINDENT and SCI_LINEDEDENT. Feature #1524.
    Fix bug on Cocoa where double-click stopped working when system had been running for a long time.
    On Cocoa implement more values of font weight and stretch.

Release 5.4.0 ( https://www.scintilla.org/lexilla540.zip )

    Released 21 August 2024.
    Inside Lexilla, LexerModule instances are now const. This will require changes to applications that modify Lexilla.cxx, which may be done to add custom lexers.
    Lexer added for TOML "toml".
    Bash: Handle backslash in heredoc delimiter. Issue #257.
    Progress: Fix lexing of nested comments. Pull request #258.
    Force lower-casing of case-insensitive keyword lists so keywords match in some lexers. Issue #259.


scintilla/src/Editor.cxx
```diff
@@ -4361,14 +4363,16 @@ void Editor::CopySelectionRange(SelectionText *ss, bool allowLineCopy) {
...
+		const std::string_view separator = (sel.selType == Selection::SelTypes::rectangle) ? pdoc->EOLString() : copySeparator;
+		for (size_t part = 0; part < rangesInOrder.size(); part++) {
+			text.append(RangeText(rangesInOrder[part].Start().Position(), rangesInOrder[part].End().Position()));
+			if ((sel.selType == Selection::SelTypes::rectangle) || (part < rangesInOrder.size() - 1)) {
+				// Append unless simple selection or last part of multiple selection
+				text.append(separator);
 			}
 		}
 		ss->Copy(text, pdoc->dbcsCodePage,
-			vs.styles[StyleDefault].characterSet, rangesInOrder.size() > 1, sel.selType == Selection::SelTypes::lines);
```
chcg, line 4371: @donho Conflict between 4ff9d77d2bfbb323e2fcba10f02570710872db94 and the current modification by scintilla 5.5.2 update which still needs to be checked and adapted.

donho: @chcg 
IMO, for not having regression, `vs.styles[StyleDefault].characterSet, sel.IsRectangular(), sel.selType == Selection::SelTypes::lines);` should be `vs.styles[StyleDefault].characterSet, rangesInOrder.size() > 1, sel.selType == Selection::SelTypes::lines);`.
I'll test it and let you know the test result.

ozone10:
follow up PR to enable `TOML` would be great
ref #5631

scintilla/src/Editor.cxx
donho: @chcg 
After checking the old behaviour, the above modification is necessary, but not enough.
We should also change from:
`if ((sel.selType == Selection::SelTypes::rectangle) || (part < rangesInOrder.size() - 1)) {`
to
`if ((rangesInOrder.size() > 1) || (part < rangesInOrder.size() - 1)) {`

and as well from:
`const std::string_view separator = (sel.selType == Selection::SelTypes::rectangle) ? pdoc->EOLString() : copySeparator;`
to
`const std::string_view separator = (rangesInOrder.size() > 1) ? pdoc->EOLString() : copySeparator;`

donho (changes requested):
Please change from:
`if ((sel.selType == Selection::SelTypes::rectangle) || (part < rangesInOrder.size() - 1)) {`
to
`if ((rangesInOrder.size() > 1) || (part < rangesInOrder.size() - 1)) {`

and also from:
`const std::string_view separator = (sel.selType == Selection::SelTypes::rectangle) ? pdoc->EOLString() : copySeparator;`
to
`const std::string_view separator = (rangesInOrder.size() > 1) ? pdoc->EOLString() : copySeparator;`

Finally change from
`vs.styles[StyleDefault].characterSet, sel.IsRectangular(), sel.selType == Selection::SelTypes::lines);` 
should be 
`vs.styles[StyleDefault].characterSet, rangesInOrder.size() > 1, sel.selType == Selection::SelTypes::lines);`.

in order to keep the same behaviour of Copy/Paste Multi-selection as previous version.

donho (approved):

PR #15565, opened 2024-08-22
Improve Styler Configurator performance

ozone10:
ref #15560

### DIFF
--- a/PowerEditor/src/NppDarkMode.cpp
+++ b/PowerEditor/src/NppDarkMode.cpp
@@ -591,6 +591,7 @@ namespace NppDarkMode
 		return invert_c;
 	}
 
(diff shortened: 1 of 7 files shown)


PowerEditor/src/NppDarkMode.cpp
```diff
@@ -1094,7 +1095,7 @@ namespace NppDarkMode
 	static void renderButton(HWND hwnd, HDC hdc, HTHEME hTheme, int iPartID, int iStateID)
 	{
 		RECT rcClient{};
-		WCHAR szText[256] = { '\0' };
+		wchar_t szText[256] = { '\0' };
```
donho, line 1098: 👍

PR #15572, opened 2024-08-24
FEATURE: Notify plugins of the user's native language preference (would address #15513)

molsonkiko:
### Rationale

This PR implements `NPPN_RELOADNATIVELANG`, which makes it easy for plugins to discover the user's native language preference without having to parse the `nativeLang.xml` file or subclass the Notepad++ native window, [as illustrated here](https://github.com/molsonkiko/NppCSharpPluginPack/blob/main/NppCSharpPluginPack/Utils/Translator.cs).

Arguments in favor of implementing a notification like this are discussed at length in #15513. Given the relatively low code footprint of this implementation, and the fact that it mostly is just doing work that each plugin wishing to translate itself otherwise would have had to do itself, this seems like a net improvement, assuming we are serious about trying to make Notepad++ as user-friendly as possible for non-English speakers. 

### Description of implementation

The `SCNotification` structs of `NPPN_RELOADNATIVELANG` have the following fields filled:
```
scnNotification->nmhdr.code = NPPN_RELOADNATIVELANG;
scnNotification->nmhdr.hwndFrom = hPluginMenu; // the menu handle to the plugin menu, so that plugins can find their plugin's submenu and re-translate their menu items
scnNotification->nmhdr.idFrom = languageName; // the ANSI encoded English name of the current UI language for Notepad++ (this is the same as the name of the localization XML file used by Notepad++ to translate the UI).
```

__EDIT:__ As this notification is currently implemented, __the `hwndFrom` and `idFrom` *must not be `NULL`* for the notification to be sent.__

This notification is emitted once at startup (at the very end of `Notepad_plus::init`) and also once each time the user changes their language preference in the Preferences form.

In order to avoid having to send this message twice every time the user changes their language preference, this PR has also changed `NPPM_INTERNAL_RELOADNATIVELANG`, changing the previously unused `WPARAM` into a boolean that, if `TRUE`, indicates that plugins should be notified of the native language change.

Aside from new documentation, most of the code footprint of this PR is associated with finding the menu handle of the plugins menu, which most plugins will need so that they can find the menu handle of their plugin's submenu. This code could probably be simplified by assuming that the plugins menu is always at `MENUINDEX_PLUGINS`, but this will not always be true, because some plugins like [Columns++](https://github.com/Coises/ColumnsPlusPlus) add their plugin menu to the Notepad++ main menu, causing the plugins menu to be at a different position than it normally would be.

### How to test

[...]

PR #15582, opened 2024-08-27
Add new plugin command & notification while native lang being changed

### LINKED ISSUE #15513: [Feature request] Send message to plugins when native language preference is changed.

### Description of the Issue

[See this community forum post for a description of the issue](https://community.notepad-plus-plus.org/post/96012).

In brief, it is currently *possible* for plugins to learn whether the user has changed their localization preference, but it is cumbersome and unintuitive to do so. There have been many requests (in various places) over the years to make plugins translatable, and adding an `NPPN_*` notification would definitely help plugin maintainers add translation to their plugins.

It seems to me that relatively little code would be required to send an `NPPN_*` notification to all the plugins after Notepad++ has finished changing all of its own UI elements. I would probably recommend continuing to use the old `NPPM_INTERNAL_RELOADNATIVELANG` notification internally, and just create a new notification, maybe called `NPPN_RELOADNATIVELANG`, to notify plugins.

Based on my own experience [implementing a response to `NPPN_INTERNAL_RELOADNATIVELANG`](https://github.com/molsonkiko/NppCSharpPluginPack/commit/82edb829a39426d1a2cba8d7fb6284fc07267c5c) in my own plugin, I believe that this notification should probably also have the following payload:
```
// Assume that const char * nativeLangName is the filename associated with the nativeLang.xml file (minus the .xml extension)
//     for example, if the current language is English, nativeLangName = "english", if the current language is Italian, nativeLangName = "italian", and so on
//  Also assume that HMENU pluginMenuHandle is the HMENU to the Plugins submenu of the Notepad++ main menu.
//      We want to send this to plugin too, so that they can always find the submenu associated with their plugin, and then re-translate all their menu items.
//     You might be thinking "Wait, can't the plugins menu be found by using GetSubMenu(hNotepadPlusMainMenu, MENUINDEX_PLUGINS)?"
//     No! This is not possible, because some plugins like Columns++ can add their plugin's submenu to the Notepad++ main menu, and that means that the Plugins submenu is no longer at index MENUINDEX_PLUGINS of the main menu.
SCNotification scnN{};
scnN.nmhdr.code = NPPN_RELOADNATIVELANG
scnN.text = nativeLangName;
scnN.length = strlen(nativeLangName);
scnN.lParam = pluginMenuHandle;
```

By telling the plugin what the user's new native language preference is, and where the Plugins submenu of the Notepad++ main menu is located, you supply every plugin with enough information to re-translate their plugin into the current UI language of Notepad++, and you eliminate the need for every plugin to re-read the `nativeLang.xml` file every time the notification is fired.

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Command Line : "C:\Users\mjols\AppData\Roaming\Notepad++\nativeLang.xml"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4651
Current ANSI codepage : 1252
Plugins : 
    ColumnsPlusPlus (1.1.1)
    ComparePlus (1.1)
    CSharpPluginPack (0.0.3.11)
    CSVLint (0.4.6.6)
    EnhanceAnyLexer (1.1.3)
    HTMLTag (1.4.3.1)
    HugeFiles (0.4.1)
    JsonTools (8.0.0.17)
    mimeTools (3.1)
    NavigateTo (2.7)
    NppConverter (4.6)
    NppExport (0.4)
    NppLspClient (0.0.21)
    PythonScript (3.0.16)
    XMLTools (3.1.1.13)
```

### Anything else?

_No response_

donho:
Add new plugin notification `NPPN_NATIVELANGCHANGED` when the native language has been changed, and add new plugin message `NPPM_GETNATIVELANGFILENAME` for plugins calling after getting `NPPN_NATIVELANGCHANGED` notification.

Usage of the command:
```cpp
	#define NPPM_GETNATIVELANGFILENAME (NPPMSG + 116)
	// int NPPM_GETNATIVELANGFILENAME(size_t strLen, char* nativeLangFileName)
	// Get the Current native language file name string.
	// Users should call it with nativeLangFileName as NULL to get the required number of char (not including the terminating nul character),
	// allocate commandLineStr buffer with the return value + 1, then call it again to get the current native language file name string.
	// wParam[in]: strLen is "commandLineStr" buffer length
	// lParam[out]: commandLineStr recieves all copied native language file name string
	// Return the number of char copied/to copy
```
Usage of the notification:
```cpp
	#define NPPN_NATIVELANGCHANGED (NPPN_FIRST + 31)  // To notify plugins that the current native language is just changed to another one.
                                                      // Use NPPM_GETNATIVELANGFILENAME to get current native language file name.
                                                      // Use NPPM_GETMENUHANDLE(NPPPLUGINMENU, 0) to get submenu "Plugins" handle (HMENU)
	//scnNotification->nmhdr.code = NPPN_NATIVELANGCHANGED;
	//scnNotification->nmhdr.hwndFrom = hwndNpp
	//scnNotification->nmhdr.idFrom = 0; // preserved for the future use, must be zero
```

Fix #15513

### DIFF
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -980,6 +980,15 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 	// lParam[in]: newName - the desired new name of the tab
 	// Return TRUE upon success; FALSE upon failure
 
+	#define NPPM_GETNATIVELANGFILENAME (NPPMSG + 116)
+	// int NPPM_GETNATIVELANGFILENAME(size_t strLen, char* nativeLangFileName)
+	// Get the Current native language file name string.
+	// Users should call it with nativeLangFileName as NULL to get the required number of char (not including the terminating nul character),
+	// allocate commandLineStr buffer with the return value + 1, then call it again to get the current native language file name string.
+	// wParam[in]: strLen is "commandLineStr" buffer length
+	// lParam[out]: commandLineStr recieves all copied native language file name string
+	// Return the number of char copied/to copy
+
 	// For RUNCOMMAND_USER
 	#define VAR_NOT_RECOGNIZED 0
 	#define FULL_CURRENT_PATH 1
@@ -1195,4 +1204,11 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
                                                    // This notification is implemented in Notepad++ v8.6.5.
 	//scnNotification->nmhdr.code = NPPN_GLOBALMODIFIED;
 	//scnNotification->nmhdr.hwndFrom = BufferID;
+	//scnNotification->nmhdr.idFrom = 0; // preserved for the future use, must be zero
+
+	#define NPPN_NATIVELANGCHANGED (NPPN_FIRST + 31)  // To notify plugins that the current native language is just changed to another one.
+                                                      // Use NPPM_GETNATIVELANGFILENAME to get current native language file name.
+                                                      // Use NPPM_GETMENUHANDLE(NPPPLUGINMENU, 0) to get submenu "Plugins" handle (HMENU)
+	//scnNotification->nmhdr.code = NPPN_NATIVELANGCHANGED;
+	//scnNotification->nmhdr.hwndFrom = hwndNpp
 	//scnNotification->nmhdr.idFrom = 0; // preserved for the future use, must be zero
--- a/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
+++ b/PowerEditor/src/MISC/PluginsManager/Notepad_plus_msgs.h
@@ -982,7 +982,7 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 
 	#define NPPM_GETNATIVELANGFILENAME (NPPMSG + 116)
 	// int NPPM_GETNATIVELANGFILENAME(size_t strLen, char* nativeLangFileName)
-	// Get the Current native language file name string.
+	// Get the Current native language file name string. Use it after getting NPPN_READY notification to find out which native language is used.
 	// Users should call it with nativeLangFileName as NULL to get the required number of char (not including the terminating nul character),
 	// allocate commandLineStr buffer with the return value + 1, then call it again to get the current native language file name string.
 	// wParam[in]: strLen is "commandLineStr" buffer length
@@ -1041,7 +1041,7 @@ enum Platform { PF_UNKNOWN, PF_X86, PF_X64, PF_IA64, PF_ARM64 };
 
 // Notification code
 #define NPPN_FIRST 1000
-	#define NPPN_READY (NPPN_FIRST + 1) // To notify plugins that all the procedures of launchment of notepad++ are done.
+	#define NPPN_READY (NPPN_FIRST + 1) // To notify plugins that all the initialization for launching Notepad++ is complete.
 	//scnNotification->nmhdr.code = NPPN_READY;
 	//scnNotification->nmhdr.hwndFrom = hwndNpp;
 	//scnNotification->nmhdr.idFrom = 0;
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -7248,13 +7248,6 @@ bool Notepad_plus::reloadLang()
 	}
 
 	_lastRecentFileList.setLangEncoding(_nativeLangSpeaker.getLangEncoding());
-
-	SCNotification scnN{};
-	scnN.nmhdr.code = NPPN_NATIVELANGCHANGED;
-	scnN.nmhdr.hwndFrom = _pPublicInterface->getHSelf();
-	scnN.nmhdr.idFrom = 0;
-	_pluginsManager.notify(&scnN);
-
 	return true;
 }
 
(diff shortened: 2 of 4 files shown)


rdipardo:
Well, this is still better than having to parse `nativeLang.xml`. 

I don't really have a dog in this hunt, so I'll be watching on the sidelines from here.

As final thoughts, let's reflect on the fact that any new API is usually a compromise between:

1. general utility, across platforms, programming languages, build systems, etc.; and,
2. suitability for a specific purpose, i.e., what the individual requester wants.

It's been said by Scintilla's maintainer that knowing what to leave out is more important than maximizing what you put in, since the latter leads to feature creep. The more you promise, the more users expect &mdash; and before you know it you'll have 2.5K open issues on your hands.

molsonkiko:
@donho 

Thank you for your new idea for this notification. I'll let you know what I think of the current implementation once I've tested to see how it works in NppCSharpPluginPack.

molsonkiko:
@donho 

I've re-implemented NppCSharpPluginPack to work with `NPPN_NATIVELANGCHANGED` as it is currently implemented, and I find that the current implementation works fine, except for the following problems (__all of which were resolved by my PR #15572 that you rejected__):
1.  The most logical time for my plugin to translate its UI to conform with Notepad++ is in the [`setInfo` plugin method](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/8e26e08c1cf799c632c3cdc067804ed128f6e743/PowerEditor/src/MISC/PluginsManager/PluginInterface.h#L65), but __I can't use `SendMessage(hwndNpp, NPPM_GETMENUHANDLE, NPPPLUGINMENU, 0)` during `setInfo` because the PluginsManager menu handle is not initialized at that time.__ It is not at all obvious to me when I *should* ask for the Plugins menu handle, and I probably have more patience for this sort of thing than the average plugin maintainer.
    - An easy way to solve this problem would be to send an `NPPN_NATIVELANGCHANGED` notification at the end of `Notepad_plus::init()`, as I did in my PR. You still wouldn't need to send the menu handle, because by that point the plugins menu handle has been assigned.
3. My plugin has to either ignore every other `NPPN_NATIVELANGCHANGED` notification, or waste time responding to an extra notification, because the `NPPM_INTERNAL_RELOADNATIVELANG` notifications always come in pairs. This is why I changed the `WPARAM` of `NPPM_INTERNAL_RELOADNATIVELANG` to a `BOOL`. The code footprint is quite small (four lines, including documentation), and it improves performance by reducing the number of notifications that are sent.
4. I don't see why plugins should need to send *two instances* of `NPPM_GETNATIVELANGFILENAME` to NPP to save NPP the trouble of sending them a string that will have *less than 30 characters.* __Considering how infrequently these notifications are sent__, I believe that sending the name of the native language file with the notification (probably in the `idFrom` field) would not meaningfully impact performance. 

I completely understand your desire to use less code and improve performance. However, the current implementation does not make much sense to me.

If you are completely unwilling to allocate more lines of code to this notification than this PR currently has, I would recommend closing this PR and issue #15513, because the whole point of creating this notification is to make life easier for plugin maintainers, and the current implementation does not make my life easier.

donho:
@molsonkiko 
Thank you for testing.

> This is why I changed the WPARAM of NPPM_INTERNAL_RELOADNATIVELANG to a BOOL. The code footprint is quite small (four lines, including documentation), and it improves performance by reducing the number of notifications that are sent.

It's indeed a bug. It's fixed in the latest commit.

> I don't see why plugins should need to send two instances of NPPM_GETNATIVELANGFILENAME to NPP to save NPP the trouble of sending them a string that will have less than 30 characters....

Firstly, NPPM_GETNATIVELANGFILENAME command is necessary because currently there's no notification for the change of language from Notepad++ startup. For example, if user changed Notepad++ native language from English to Japanese, notification is sent. The next startup of Notepad++ Japanese will be loaded over English implicitly, but without notification.
In this case, NPPM_GETNATIVELANGFILENAME is useful for the plugins after NPPN_READY notification.

Secondly, for keeping consistency of all the notification, I prefer to not put the string into idFrom or code field.

> The most logical time for my plugin to translate its UI to conform with Notepad++ is in the [setInfo plugin method](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/8e26e08c1cf799c632c3cdc067804ed128f6e743/PowerEditor/src/MISC/PluginsManager/PluginInterface.h#L65), but I can't use SendMessage(hwndNpp, NPPM_GETMENUHANDLE, NPPPLUGINMENU, 0) during setInfo because the PluginsManager menu handle is not initialized at that time.

The most logical time to do it is waiting for NPPN_READY notification.

molsonkiko:
@donho 

Thanks for fixing the double-notification bug!

> Firstly, NPPM_GETNATIVELANGFILENAME command is necessary because currently there's no notification for the change of language from Notepad++ startup. [other justification...]

I think this is a reasonable perspective, and I'm satisfied enough that I see no reason to remove `NPPM_GETNATIVELANGFILENAME`.

> The most logical time to do it is waiting for NPPN_READY notification.

Thank you for clarifying that. Now that I use the same translation code in `NPPN_READY` as I do in `NPPN_NATIVELANGCHANGED`, my plugin is translated at startup with significantly less code than before. At this point, I believe that the code I believe that other plugin maintainers would have a much easier time using this new API if you stated this in the documentation for `NPPN_NATIVELANGCHANGED` or `NPPM_GETNATIVELANGFILENAME`.

I would probably add the following comment at the end of the documentation for `NPPN_NATIVELANGCHANGED`.

```
// If you want to translate your plugin's UI as Notepad++ is starting up,
//     you can run the same translation code in response to NPPN_READY that you run in response to NPPN_NATIVELANGCHANGED.
```

PR #15587, opened 2024-08-28
Fix issue with multiple selections in combobox fields in Find dialog

### LINKED ISSUE #15583: [BUG] Find/Replace/FiF/... dialog "selects" multiple fields

### Description of the Issue

_originally described by user in Community Forum, in https://community.notepad-plus-plus.org/topic/26053/find-in-files-all-fields-highlighted, quoted here:_

> In the latest version of Notepad++ (v8.6.9), when I open the "Find in Files" dialog box, the text in all fields is highlighted, instead of only the text of the active field:
> 

Myself and mkupper and Coises were all able to confirm.

Essentially, if you have a search history (previous FIND WHAT or REPLACE WITH or FILTERS values) in the Find / Replace / Find In Files dialog, and if you've stretched the size of the dialog bigger than the default width, then when you open the dialog, it will make it look like the text of all the input fields (FIND WHAT, REPLACE WITH, FILTERS, ...) are selected, instead of only the FIND WHAT having the text selected.

### Steps To Reproduce

In the Forum, we were able to come up with three different sequences of Steps to Reproduce:

**Pryrt (PeterJones)'s steps**
1. Fresh unzip of portable v8.6.9-64bit
2. Run once and exit so it creates `config.xml`
3. Use external editor on `config.xml` to set 
   ```
    <FindHistory nbMaxFindHistoryPath="10" nbMaxFindHistoryFilter="10" nbMaxFindHistoryFind="10" nbMaxFindHistoryReplace="10" matchWord="no" matchCase="no" wrap="yes" directionDown="yes" fifRecuisive="no" fifInHiddenFolder="no" fifProjectPanel1="no" fifProjectPanel2="no" fifProjectPanel3="no" fifFilterFollowsDoc="no" fifFolderFollowsDoc="yes" searchMode="0" transparencyMode="1" transparency="150" dotMatchesNewline="no" isSearch2ButtonsMode="no" regexBackward4PowerUser="no" bookmarkLine="no" purge="no">
        <Filter name="*.*" />
        <Find name="hello" />
        <Replace name="something" />
    </FindHistory>

   ```
   ... and save
4. Run Notepad++, `Ctrl+Shift+F` looks normal: 
 
5. Exit Notepad++.
6. Use external editor on `config.xml` to edit `<GUIConfig name="FindWindowPosition" ...`: edit the `right="852"` to add one to the value (so in my example, it would be `right="853"`), then save
6. Run Notepad++, `Ctrl+Shift+F` highlights all three fields: 
 

**mkupper's steps**

> I was able to reproduce by dragging the find-in-files box larger but also see that it happens dynamically.
> 
> * I did `Ctrl+Shift+F` to bring up the find-in-files dialog box. The fields looked ok with only `Find what` highlighted.
> * Using the anchor or whatever it's called in the lower-right corner of the dialog box I dragged it a little wider. The instant I started dragging all four of the main fields of find-in-files became highlighted.
> 
> I then discovered the same issue happens with Replace.  
> * Now that the dialog box was dragged a little wider I closed and reopened Notepad++.
> * When I do `Ctrl+H` both the `Find what` and `Replace with` fields are highlighted.
> 
> As an extra data point.
> * If I drag the box back to its smallest width, which is also the default width then the problem persists until I exit/restart > Notepad++. This is with both `Ctrl+H` or `Ctrl+Shift+F` to activate the dialog box.

**Coises' steps**
1. Open any tab of the Find/Replace/... dialog.
2. Drag it to minimum width.
3. Close the dialog.
4. Close Notepad++.
5. Open Notepad++.
6. Open the Replace, Find in Files or Find in Projects tab of the search dialog.
7. *Highlighting appears only in the first field.*
8. Drag the dialog width larger.
9. *Highlighting immediately appears in all the text input fields.*
10. Close and re-open Notepad++.
11. Open the Replace, Find in Files or Find in Projects tab of the search dialog.
12. *Highlighting appears in all the text input fields.*
13. Drag the dialog back to minimum width.
14. *Highlighting still appears in all the text input fields.*
15. Close Notepad++.
16. Open Notepad++.
17. Open the Replace, Find in Files or Find in Projects tab of the search dialog.
18. *Highlighting appears only in the first field.*

### Current Behavior

When you bring up the **Find/Replace/FiF/...** when there is a search history and the dialog has been enlarged beyond the default width, it shows all fields selected, though if you try to type, it only tries to overwrite the **Find what** field

 

### Expected Behavior

When you bring up the **Find/Replace/FiF/...** when there is a search history and the dialog has been enlarged beyond the default width, it should show just the **Find what** field selected.

 

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Path : C:\usr\local\apps\npp\npp869\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4780
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

mkupper further narrowed it down: v8.6.7 didn't have the problem, v8.6.8 and v8.6.9 do.  (My StR won't show it in v8.6.8, because v8.6.8 ignored the Find-dialog's size/position from `config.xml`.  But mkupper's StR are confirmed to show the problem in v8.6.8.)
```


### Anything else?

_No response_

ozone10:
fix #15583

### DIFF
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1181,7 +1181,8 @@ void FindReplaceDlg::resizeDialogElements()
 	const bool isLessModeOn = NppParameters::getInstance().getNppGUI()._findWindowLessMode;
 
 	//elements that need to be resized horizontally (all edit/combo boxes etc.)
-	const auto resizeWindowIDs = { IDFINDWHAT, IDREPLACEWITH, IDD_FINDINFILES_FILTERS_COMBO, IDD_FINDINFILES_DIR_COMBO };
+	const std::vector<int> resizeWindowIDs = { IDFINDWHAT, IDREPLACEWITH, IDD_FINDINFILES_FILTERS_COMBO, IDD_FINDINFILES_DIR_COMBO };
+	const size_t nComboboxes = resizeWindowIDs.size();
 
 	//elements that need to be moved
 	const auto moveCheckIds = {
@@ -1275,24 +1276,20 @@ void FindReplaceDlg::resizeDialogElements()
 
 	getMappedChildRect(hSwapBtn, rcSwapBtn);
 
-	nCtrls = resizeWindowIDs.size() + moveLaterIDs.size() + (isLessModeOn ? 0 : moveTransIDs.size()) + 1; // 1 is for tab control
+	nCtrls = nComboboxes + moveLaterIDs.size() + (isLessModeOn ? 0 : moveTransIDs.size()) + 1; // 1 is for tab control
 	hdwp = ::BeginDeferWindowPos(static_cast<int>(nCtrls));
 
-	for (int id : resizeWindowIDs)
+	std::vector<DWORD> endSelections(nComboboxes, 0);
+
+	for (size_t i = 0; i < nComboboxes; ++i)
 	{
-		HWND resizeHwnd = ::GetDlgItem(_hSelf, id);
+		HWND resizeHwnd = ::GetDlgItem(_hSelf, resizeWindowIDs[i]);
 		getMappedChildRect(resizeHwnd, rcTmp);
 
 		// Combo box for some reasons selects text on resize. So let's check before resize if selection is present and clear it manually after resize.
-		DWORD endSelection = 0;
-		::SendMessage(resizeHwnd, CB_GETEDITSEL, 0, reinterpret_cast<LPARAM>(&endSelection));
+		::SendMessage(resizeHwnd, CB_GETEDITSEL, 0, reinterpret_cast<LPARAM>(&endSelections[i]));
 
 		hdwp = setOrDeferWindowPos(hdwp, resizeHwnd, nullptr, 0, 0, rcSwapBtn.left - rcTmp.left - gap, rcTmp.bottom - rcTmp.top, SWP_NOMOVE | flags);
-
-		if (endSelection == 0)
-		{
-			::SendMessage(resizeHwnd, CB_SETEDITSEL, 0, 0);
-		}
 	}
 
 	RECT rcFPrevBtn{};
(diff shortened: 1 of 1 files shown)


ozone10:
@pryrt @StijnBousard 
could you test PR, if it is fixed?

pryrt:
@ozone10,
This PR fixes the problem for me.  Thanks!

@StijnBousard, in case you don't know how: there are instructions [on this page](/notepad-plus-plus/notepad-plus-plus/wiki/Testing#using-a-pull-request-pr-build-version-of-notepad), but to sum up: go to the **Checks** page from this PR, use the "Artifacts" pulldown and choose "MSCV.x64.Release", which will download a zipfile.
image
Inside a zipfile is the PR-build of the `notepad++.exe` -- if you run that build, it will be the current build, with the bugfix applied.

donho:
@ozone10 
Not for finding a culprit, but just for keeping a track:
Is it a regression due to e25e15b9b25afee7f56fe6e02ef4a314468f6d7b ?

donho:
> Not for finding a culprit, but just for keeping a track: Is it a regression due to [e25e15b](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/e25e15b9b25afee7f56fe6e02ef4a314468f6d7b) ?

OK, found the commit:
...
bdc11a47fffb5af3f3b9cc0012a3b9230f08e07e **Reproducable**
e25e15b9b25afee7f56fe6e02ef4a314468f6d7b **Reproducable**
71914ed5453c550882929c1ef19ff6b77d80a5ff **Not Reproducable**
...

So the regression was introduced by e25e15b9b25afee7f56fe6e02ef4a314468f6d7b

ozone10:
@donho 
yes, I have replaced `SetWindowPos` with `DeferWindowPos`. It is used to move controls together at once for better performance, since paint update happens only once instead individually for each controls. 

Because I have not found that issue at that time I have not updated code with `::SendMessage(resizeHwnd, CB_SETEDITSEL, 0, 0);`, which needs to be after moving control, which is after `::EndDeferWindowPos(hdwp);`, not after `DeferWindowPos`.

donho:
> Because I have not found that issue at that time I have not updated code with `::SendMessage(resizeHwnd, CB_SETEDITSEL, 0, 0);`, which needs to be after moving control, which is after `::EndDeferWindowPos(hdwp);`, not after `DeferWindowPos`.

@ozone10 
No problem at all - with the huge amount of code that you have contributed, there are very few regressions or bugs come from the code you made. 
I just need the the commit which introduced the regression to have the track for the future (just in case). 
That remind me a joke: ***With great refactoring comes great regressions.*** :)

StijnBousard:
> @pryrt @StijnBousard could you test PR, if it is fixed?

Looks good to me now. Thanx for the fix!

PR #15598, opened 2024-09-02
Fix monitoring large file with frequent writes freezes UI

### LINKED ISSUE #9661: When monitoring a large file with frequent writes, notepad++ locks up
### Description of the Issue
The NPPM_INTERNAL_RELOADSCROLLTOEND message is sent every time the monitored file is updated.  The handler for this message calls buf->reload() synchronously.
If writes are frequent and the file is large (or IO slow), reload() will take longer to finish than it takes for more NPPM_INTERNAL_RELOADSCROLLTOEND messages to come in.  Messages received while reload() is still running are queued up.
Each message will be handled in the order received until all have completed, even after writes have stopped.

Suppose reload() takes 250ms, and 100 writes are made over 10 seconds.
Notepad++ will be unresponsive for 25 seconds (the time it takes for 100 reloads), even though writing stopped after 10 seconds.

Ideally, the message handler should only set a flag showing that a reload is needed, and then return.
A separate timer should check the flag and if set, unset it and reload the file.

1. In a command prompt run the following:
`for /l %x in (1,1,1000) do echo line number %x >> testfile.txt`
2. While this runs, open the resulting testfile.txt in notepad++ and turn on file monitoring

### Expected Behavior
Notepad++ should remain responsive as it updates the file, and stop reloading it once the command is finished.

### Actual Behavior
Notepad++ is completely unresponsive until long after the command has finished.

### Debug Information
Notepad++ v7.9.1   (64-bit)
Admin mode : OFF
Local Conf mode : OFF
OS Version : 1803
OS Build : 17134.2026
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll XMLTools.dll

BubuC:
Fix for annoying issue I commonly encounter when viewing actively written log files with monitoring enabled; as reported by others in #9661

Fix #9661

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -98,7 +98,7 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 							{
 								::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
 								bDoneOnce = true;
-								Sleep(250);
+								Sleep(250);	// Limit refresh rate
 							}
 						}
 						else if ((dwAction == FILE_ACTION_REMOVED) || (dwAction == FILE_ACTION_RENAMED_OLD_NAME))
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -79,6 +79,7 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 			case WAIT_OBJECT_0 + 1:
 			// We've received a notification in the queue.
 			{
+				bool bDoneOnce = false;
 				DWORD dwAction = 0;
 				wstring fn;
 				// Process all available changes, ignore User actions
@@ -93,7 +94,12 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 					{
 						if (dwAction == FILE_ACTION_MODIFIED)
 						{
-							::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
+							if (!bDoneOnce)
+							{
+								::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
+								bDoneOnce = true;
+								Sleep(250);
+							}
 						}
 						else if ((dwAction == FILE_ACTION_REMOVED) || (dwAction == FILE_ACTION_RENAMED_OLD_NAME))
 						{


PowerEditor/src/NppIO.cpp
```diff
@@ -93,7 +94,12 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 					{
 						if (dwAction == FILE_ACTION_MODIFIED)
 						{
-							::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
+							if (!bDoneOnce)
+							{
+								::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
+								bDoneOnce = true;
+								Sleep(250);
```
donho, line 101: Where come from the magic number "250" ?

donho:
It's absolutely a solution for me, with Sleep() is not bad for me, and with the magic number "250" is even worse.

PowerEditor/src/NppIO.cpp
```diff
@@ -93,7 +94,12 @@ DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 					{
 						if (dwAction == FILE_ACTION_MODIFIED)
 						{
-							::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
+							if (!bDoneOnce)
```
donho, line 97: What's the point of this boolean? It'll be always true this condition.

PowerEditor/src/NppIO.cpp
BubuC: Consider the preceding line:
`69 	DWORD waitStatus = ::WaitForMultipleObjects(_countof(changeHandles), changeHandles, FALSE, 250);`

WaitForMultipleObjects waits for file modification events, or times out after 250ms, acting as a polling loop.

But, when monitoring log files actively being written, hundreds of events can be returned in one call, and each matching FILE_ACTION_MODIFIED event will PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND... leading to a freeze.
```
	if (!bDoneOnce)
	{
		::PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND, reinterpret_cast<WPARAM>(buf), 0);
		bDoneOnce = true;
		Sleep(250);
	}
```
Above, I am using a boolean flag to only PostMessage for first matching event, and then Sleeping for a similar 250ms as used in the polling loop, to break the flood of reloads.

PowerEditor/src/NppIO.cpp
BubuC: Please see my changes in context by looking at the whole case statement [here](https://github.com/BubuC/notepad-plus-plus/blob/cf99b9ec59a20129f6093a807896600f16ff958d/PowerEditor/src/NppIO.cpp#L79)

The boolean is set to false outside the loop, and set to true once inside, so that only first of a flood of matching events triggers PostMessage(h, NPPM_INTERNAL_RELOADSCROLLTOEND...

PowerEditor/src/NppIO.cpp
MarkusBodensee: I don't get the complete logic here... Can you explain? Why is both needed? The boolean flag to trigger PostMessage only once and the sleep afterwards?

If the PostMessage is triggered only once, the application does not freeze anymore. Only first event is processed and the all others will be ignored.

Sleep(...) blocks the application from running, which seems not needed (could even lead to non-responsive app, too). IMO Sleep(...) should be avoided.

BubuC: No, it does not block the UI, even if you use a delay like 10 seconds. It simply limits the refresh rate for a monitored document with rapid continuous updates. It is without it that the UI becomes unresponsive. 

Just because everyone has an opinion on Sleep(...) doesn't mean this approach is not correct here.

MarkusBodensee: Thank you for clarifiing and confirming, that it will not block the application. That's fine.

But still don't get completely why calling Sleep(...) is needed at all.

As from my understanding, there are multiple FILE_ACTION_MODIFIED which get popped from the available dirChanges (in while loop). And all of them would invoke a PostMessage(... NPPM_INTERNAL_RELOADSCROLLTOEND ...) which is the cause of Notepad++ becoming unresponsive for a while.

But since with your change, PostMessage(... NPPM_INTERNAL_RELOADSCROLLTOEND ...) is invoked only once (the first time a FILE_ACTION_MODIFIED is received), all further FILE_ACTION_MODIFIED will be ignored anyway, even without calling Sleep(...). So the root cause of unresponsiveness seems gone, even without calling Sleep(...).

Limiting the refresh rate seems already been done by WaitForMultipleObjects(...) because it also waits for 250ms.

Not sure if I have formulated my thoughts well, nor whether they are correct. Will try to test a bit.

BubuC: @MarkusBodensee WaitForMultipleObjects rarely waits the full 250ms, unless there are no events queued. Once the PostMessage is sent, editor will get busy with loading, parsing and layout and whatever else with the plugins, all this while more events accumulate. If there is no breathing space between each refresh, it become unresponsive because it is busy all the time. It gets worse as the file grows.

To test under load, run a command like `for /l %x in (1,1,100000) do echo line number %x >> testfile.txt` in a windows command prompt, and then load and monitor the file testfile.txt while it is being continuously updated.

MarkusBodensee: Thank you for your explanations.

PR #15602, opened 2024-09-03
Simplify customizable tab color feature

donho:
In commit a16261caaa173689da468d5be66e32866681e152, customizable tab color feature was added. However the entries were needed for both "Tab color #" & "Tab color dark mode #". This commit keeps only 5 entries of "Tab color #" and removes 5 entries of "Tab color dark mode #" from Style Configurator, and uses tab light mode color for generating tab dark mode color on the fly.


donho:
With this PR, while setting the tab color under dark mode in Style Configurator, the color chosen from the color picker doesn't match with the applied tab color - it will make users confused. 
Therefore this PR won't be merged.

Issue #13181, opened 2023-02-20
Disable monitoring mode deactivates read-only


rotemit:
### Description: 

I want to open a file as read only, and keep it as read only as long as it's open no matter what. It's a log file and I don't want the user to be able to delete important parts accidentally.
While the file is open, I want to toggle monitoring mode, because I need it to be on at times and off at other times (the file is being written to continuously).
The problem arises when I disable monitoring: this deactivates the read only feature.

### Steps to Reproduce the Issue

1. Open a file in Notepad++ via command line with "-ro" and "-monitor" flags
2. When the file is open, turn off monitoring using the eye icon
3. Edit and save the file

**Other ways to reproduce**;
Instead of step 1 above, do one of the following:
- Open the file from command line with "-ro" flag and when the file is open turn on monitoring.
- Open the file from command line with "-monitor" flag and when the file is open make it read-only, from either "edit menu" or "tab context menu".
- Open the file manually from Notepad++, turn on monitoring and read-only manually, then turn off monitoring.

### Expected Behavior
Editing the file shouldn't be allowed. Especially in the scenario where it was opened from command line using "-ro" flag.

### Actual Behavior
We can write, delete, and save the file. In essence, read-only was disabled once monitoring was disabled.

### Debug Information
Notepad++ v8.4.9   (64-bit)
Build time : Jan 27 2023 - 03:11:16
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 1909
OS Build : 18363.2274
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)

donho:
The issue cannot be reproduced in v8.7 - I do believe it's been solved in previous versions since long time ago.
BTW, using only `-monitor` is enough. `-RO` is not necessary.

BubuC:
@donho This bug in one form or other is still there... I submitted a suggested fix for it in #15605

To reproduce: Launch without switches, open a log file, set as readonly from Edit menu. If later we enable monitoring, then disable monitoring, file is no longer in readonly, which is unexpected behaviour to us.

PR #15605, opened 2024-09-05
#15604 open files from FolderAsWorkspace as read-only with switches -openFoldersAsWorkspace -ro

### LINKED ISSUE #15604: [Feature] LogViewer use-case: FoldersAsWorkspace open files as read-only with -ro option

### Description of the Issue

When I launch Notepad++ with command line switches `-openFoldersAsWorkspace -ro ` and I pick files from the Workspace, the -ro switch is ignored.

We all hate modifying log files by accident!

I would prefer if the editor operates in read-only mode when picking files from Workspace, unless I explicity unset the option for a file. 

### Debug Information

```shell
n/a
```

### Anything else?

Working on potential fix

BubuC:
(LogViewer use-case)
- When using -openFoldersAsWorkspace -ro  switches, I would prefer if the editor operates in read-only mode, unless I explicity unset the option for a file, for the session. 
- When I turn on monitoring, if file is stale, I expect file to be reloaded and scrolled to end.
- When I turn off monitoring, the read-only UI state should be restored to prior status, not cleared.

note: nppParam.getCmdLineParams not synced with nppParam.getCmdLineString from launch, so used latter

fix #15604  #15660 #13181

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -8863,13 +8863,19 @@ void Notepad_plus::monitoringStartOrStopAndUpdateUI(Buffer* pBuf, bool isStartin
 	if (pBuf)
 	{
 		if (isStarting)
+		{
 			pBuf->startMonitoring();
+			pBuf->saveReadOnlyState();
+			pBuf->setUserReadOnly(true);
+		}
 		else
+		{
 			pBuf->stopMonitoring();
+			pBuf->setUserReadOnly(pBuf->restoreReadOnlyState());
+		}
 
 		checkMenuItem(IDM_VIEW_MONITORING, isStarting);
 		_toolBar.setCheck(IDM_VIEW_MONITORING, isStarting);
-		pBuf->setUserReadOnly(isStarting);
 	}
 }
 
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -8864,9 +8864,10 @@ void Notepad_plus::monitoringStartOrStopAndUpdateUI(Buffer* pBuf, bool isStartin
 	{
 		if (isStarting)
 		{
+			pBuf->reload();
 			pBuf->startMonitoring();
 			pBuf->saveReadOnlyState();
-			pBuf->setUserReadOnly(true);
+			pBuf->setUserReadOnly(true);		
 		}
 		else
 		{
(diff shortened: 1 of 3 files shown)


donho:
There are 2 logic flaws in the behaviour of this PR.

Notepad++ can be launched from CMD to open files directly:
1. by listing their path one by one.
2. by listing a session file, for opening a group of files of the saved session (`-openSession`)

Currently `-RO` is applied only to the 1st case. Such behaviour is coherent to me.
It can be discussed for `-RO` being applied also to the 2nd case, since the result of `-openSession` is making the files opened directly.

So the 1st logic flaw in the behaviour is that `-openFoldersAsWorkspace` doesn't open any file but Folder as Workspace with the folder only. 

The 2nd logic flaw in the behaviour of the PR is, consider the following command:
`.\notepad++.exe -multiinst -nosession -openFoldersAsWorkspace -ro c:\temp`

After launching the above command, *c:\temp* is opened in Folder as Workspace, and open the files from *c:\temp* from Folder as Workspace, they are all read-only, which is good. However, when you add a new folder in FaW, and open any file under the new opened folder, the file is opened as read-only, but it shouldn't be - because `-RO` should be applied only to *c:\temp*.

As a result, `-RO` should not be applied to Folder as Workspace logically, IMO.

BubuC:
I don't see it as an unreasonable request, but yes, it requires an accomodation from the prior strict logic. Currently -ro with -openFaW does nothing, so why not make the combination do something some users actually need and would reasonably expect?

My usecase is viewing log files in a folder, which I'd like to do with the editor in readonly by default, for the session. My suggested enhancement is simple, limited in scope, and does not break anything. So, why not consider it?

donho:
@BubuC 
> My suggested enhancement is simple, limited in scope, and does not break anything. So, why not consider it?

I do consider it. And indeed the enhancement is simple & limited in scope. I'm finding a concise way to integrate this feature.

How about using `-readOnlyFoldersAsWorkspace` instead of `-openFoldersAsWorkspace -ro` ?
So with `-readOnlyFoldersAsWorkspace`, FaW will keep read-only constantly during the whole session whenever user add any folder in it - il'll be less confusing than `-openFoldersAsWorkspace -ro`.

Except making FaW read-only, `-readOnlyFoldersAsWorkspace` should have the same behaviour as `-openFoldersAsWorkspace`.

donho:
@BubuC 
What do you think my suggestion?

donho:
Due to the absence of reply, the issue is closed.

Issue #15612, opened 2024-09-06
[BUG] Lua language prioritizes capitalized buffer words over language keywords


Craftidore:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

When using the lua language, capitalized keywords in comments are prioritized by autocomplete over language keywords. Previously they were not.

(My boss got annoyed with the new behavior and started downgrading all instances of notepad++ on our machines.)

### Steps To Reproduce

1. Open notepad++
2. Open new tab, set language to lua
3. Paste the following code into the editor:

```lua
-- IF
x = 2
if (x == 2) then
	print("x is 2!")
end


```

4. At the end of the buffer, on a new line, type `i`.

### Current Behavior

![image](https://github.com/user-attachments/assets/494499f2-1932-417c-ad02-6ba3a42093da)

`IF`, a word from the buffer, is prioritized over the language keyword.

### Expected Behavior

The language keyword `if` is prioritized over the word `IF`.

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Build time : Jul 12 2024 - 05:09:25
Path : C:\Users\craftidore\scoop\apps\notepadplusplus\current\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 10 Pro (64-bit)
OS Version : 22H2
OS Build : 19045.4780
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

In `{NP++ Directory}/autoCompletion/lua.xml`, if you modify the third line to say `ignoreCase="no"` and restart notepad++, the original behavior returns.

Lua is _not_ case insensitive, not even for keywords.
![image](https://github.com/user-attachments/assets/221042d1-268d-472b-bdc4-b04d2f5aa82f)

PR #15613, opened 2024-09-06
#15612 [xml] Update autoCompletion/lua.xml to be case sensitive

Craftidore:
Fix for issue #15612


donho:
Can any Lua user confirm the PR is helpful?

MarkusBodensee:
@donho @Craftidore 
I can confirm at least, that Lua is case-sensitive. So it seems, this change is valid.

For reference, see official Lua documentation: https://www.lua.org/pil/1.3.html

MarkusBodensee:
Haha, just realised, my first sentence was a rhyme.

Issue #11627, opened 2022-05-02
[feature request] Minimize to tray instead of closing.


MARK2580:
### Description of the Issue

I would like a setting to appear so that Notepad++, when you click on the window button to close [X], is minimized to the system tray in the form of an icon (similar to various launchers and instant messengers).

pryrt:
Developers: this is related to the conversation [in the forum](https://community.notepad-plus-plus.org/topic/22937/minimize-notepad-to-tray-instead-of-closing)

@alankilborn actually shared a possible (but slightly overpowered) PythonScript workaround, which hijacks the `X`, and **File > Exit**, and `Alt+F4`, and makes you use `Shift+WindowCloseX` to actually close the application -- anything else will minimize (or minimize to tray, if that option is on).

Given the overpowered nature of the workaround, if it were implemented by the Development team as part of Notepad++, I would suggest that they need to make sure that a compiled version of this setting/option would only intercept the WindowCloseX, and that **File > Close** and `Alt+F4` would still really close the application (or at least the Windows-standard `Alt+F4`).  Because, as Alan said in that discussion, there needs to be an obvious answer to "how do I actually exit Notepad++ when this feature is enabled".  (I guess as an alternative, if **File > Close** and `Alt+F4` have to match WindowCloseX behavior, then the Tray Icon Right Click menu (which I didn't remember had been implemented, since I rarely use the minimize-to-tray feature) should add an option to "Fully Exit Notepad++" below the "Close Tray Icon" entry.

pryrt:
duplicate of #4075

PR #15617, opened 2024-09-08
Add "Close to system tray" MISC preference

### LINKED ISSUE #4075: Allow Notepad++ to stay permanently in the system tray
Notepad++ currently has the ability to minimize to the system tray.  However, I think it would be good to also add the option of "permanently stay in the system tray".  This would mean that the Notepad++ system tray icon would always remain, and when you click the close button, it closes all tabs, closes the main window, but not the system tray icon.  The way to really close Notepad++ would be to right-click on the system tray icon and click "close tray icon".  This would allow you to quickly close all documents in Notepad++, yet still have the convenient system tray icon to open it up again with a blank tab.

pawelzwronek:
Main motivation is to speed up opening files. Now when I have opened like 20 tabs, most of then not saved, opening new file takes about 1-2 secs. Closing all tabs shorten this sub 1sec. This is still quite long compared to instant opening(editing) an image with MSPaint.

obraz

Fix #4075 #11627


xomx:
@pawelzwronek 

If I got it right, you only changed the N++ closing action to minimizing to systray.
But at least in the mentioned issue #4075 to fix, the users requested a different behavior:

> , and when you click the close button, `it closes all tabs`, closes the main window, but not the system tray icon. 

I am not saying that your current PR is bad, I like it as it is, but it might be better to point out that this is not how your code exactly behaves.

pawelzwronek:
@xomx you are right. This issue is more complex, so I suggested the author to open new issue with new preference "Close all tabs on close/exit".

PR #15624, opened 2024-09-13
[xml] Update Croatian.xml for Notepad++ 8.7

Karlo-F:
[xml] Update Croatian translation for Notepad++ 8.7


donho:

PR #15631, opened 2024-09-15
add NULL check in NPPM_GETNATIVELANGFILENAME

### LINKED ISSUE #15627: [BUG] In Npp 8.7 RC, if there is no nativeLang.xml, receiving NPPM_GETNATIVELANGFILENAME from a plugin causes crash

### Description of the Issue

As described in the title. [First described here](https://community.notepad-plus-plus.org/topic/26105/notepad-v8-7-release-candidate/6).

I discovered this accidentally when I installed the latest version of [NppCSharpPluginPack](https://github.com/molsonkiko/NppCSharpPluginPack) in a fresh install of the [Notepad++ 8.7 RC](http://download.notepad-plus-plus.org/repository/8.x/8.7.RC/), and found that Notepad++ was crashing on startup when that plugin was enabled. Running the plugin in the Visual Studio debugger determined that __the crash happened while Notepad++ was processing the `NPPM_GETNATIVELANGFILENAME` message__ that my plugin sent in response to the `NPPN_READY` notification.

When I run the same version of the plugin on Notepad++ 8.6.9, or a slightly earlier version of the plugin that doesn't sent the message on Notepad++ 8.7, I have no problem.

Presumably the best fix to this issue is to have `NPPM_GETNATIVELANGFILENAME` return `NULL` if there is no `nativeLang.xml` file, or else return `english` since that's the default language.

### Steps To Reproduce

1. Download the NppCSharpPluginPack ([64-bit files](https://github.com/molsonkiko/NppCSharpPluginPack/blob/main/NppCSharpPluginPack/Release_x64.zip), and the [32-bit files](https://github.com/molsonkiko/NppCSharpPluginPack/blob/main/NppCSharpPluginPack/Release_x86.zip))
2.  Unzip the zipped folder into a folder named `CSharpPluginPack` and put it into the `plugins` folder of the latest master
3.  Try to run `notepad++.exe`
4. Watch it crash

### Current Behavior

See above

### Expected Behavior

See above

__NOTE:__ the Debug Information below obviously does not show CSharpPluginPack 0.0.3.14, because that plugin caused the crash.

### Debug Information

```shell
Notepad++ v8.7   (64-bit)
Build time : Sep 11 2024 - 19:03:37
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4894
Current ANSI codepage : 1252
Plugins : 
    JsonTools (8.1.0.6)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

molsonkiko:
Would fix #15627 

To test, follow the replication instructions [in this comment](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15627#issue-2526729405). You may also wish to verify that the message still works normally when `nativeLang.xml` *does* exist by setting the UI language to `Español` and verify that the `CSharpPluginPack` plugin has been translated into spanish.

### DIFF
--- a/PowerEditor/src/NppBigSwitch.cpp
+++ b/PowerEditor/src/NppBigSwitch.cpp
@@ -3761,7 +3761,10 @@ LRESULT Notepad_plus::process(HWND hwnd, UINT message, WPARAM wParam, LPARAM lPa
 
 		case NPPM_GETNATIVELANGFILENAME:
 		{
-			string fileName = _nativeLangSpeaker.getFileName();
+			const char * fileNameBuf = _nativeLangSpeaker.getFileName();
+			if (fileNameBuf == NULL)
+				return 0;
+			string fileName(fileNameBuf);
 			if (lParam != 0)
 			{
 				if (fileName.length() >= static_cast<size_t>(wParam))


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/ea08a8929d506fd40af07c5111ee2fad3d76bbac

PR #15632, opened 2024-09-15
Fix hanging issue while hiding lines

### LINKED ISSUE #15630: [BUG] If we overlap line masking, Notepad++ freezes/crashes.

### Description of the Issue

When we hide already hidden lines, Notepad++ freezes/crash. It's not systematic, and I mostly feel that the issue occurs when the **End of File (EoF) is included** in the hidden section.

### Steps To Reproduce

---

1/ Download a Notepad++ Portable Minimalist version to ensure there are no unwanted plugins.

---

2/ Extract it.

---

3/ Launch Notepad++

---

4/ Write the following lines:
```
{
	// sample
	// sample
}
{
	// sample
	// sample
}
```
<details>
<summary>Image</summary>

Description de l'image

</details>

---

5/ Highlight the last 4 lines (lines 8 to 5) and press `ALT + H` to hide them.
<details>
<summary>Image</summary>

Description de l'image

</details>

---

6/ Now, you should only see this:
```
{
	// sample
	// sample
}
}
```
With the fold markers on line 4 (>) and line 8 (<).

<details>
<summary>Image</summary>

Description de l'image

</details>

---

7/ Highlight all 5 lines (lines 8 to 1, including lines 5, 6, and 7, which are hidden) and press `ALT + H` to hide them.

<details>
<summary>Image</summary>

Description de l'image

</details>

---

8/ Notepad++ freezes.

<details>
<summary>Image</summary>

Description de l'image

</details>

---

### Current Behavior

Notepad++ Freeze.

### Expected Behavior

I expect the lines to hide and unhide normally while respecting previous fixes that ensure hidden lines remain hidden when folding or unfolding.

I'm a beginner programmer, and I acknowledge that understanding the logic behind what should be hidden or not, keeping track of what needs to remain hidden, and managing overlapping 'un/hide line' and 'un/fold line' operations can be quite a headache.

I would fully understand and accept that this bug is not fixable in order to avoid breaking the rest of Notepad++'s code.

### Debug Information

```shell
Notepad++ v8.6.9   (64-bit)
Path : C:\Downloads\npp.8.6.9.portable.minimalist.x64\notepad++.exe
Command Line : C:\Downloads\npp.8.6.9.portable.minimalist.x64\bug.txt
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 8.1 (64-bit)
OS Build : 9600.20045
Current ANSI codepage : 1252
Plugins : none
```

### Anything else?

Related issues ?

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13682
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12184

Adrian104:
Fix #15630

In some specific cases, old markers weren't removed properly and this could lead to a crash.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -3983,7 +3983,7 @@ void ScintillaEditView::hideLines()
 		// Special case: user wants to hide every line in between other hidden sections.
 		// Both "while" loops are executed (merge with above AND below hidden section):
 
-		while (scope == 0 && startMarker < nbLines)
+		while (scope == 0 && static_cast<intptr_t>(startMarker) >= 0)
 			removeMarker(--startMarker, 1 << MARK_HIDELINESBEGIN);
 
 		while (scope != 0 && endMarker < nbLines)
@@ -3994,7 +3994,7 @@ void ScintillaEditView::hideLines()
 		// User wants to hide some lines below/above other hidden section.
 		// If true, only one "while" loop is executed (merge with adjacent hidden section):
 
-		while (scope < 0 && startMarker < nbLines)
+		while (scope < 0 && static_cast<intptr_t>(startMarker) >= 0)
 			removeMarker(--startMarker, 1 << MARK_HIDELINESBEGIN);
 
 		while (scope > 0 && endMarker < nbLines)
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -3983,10 +3983,10 @@ void ScintillaEditView::hideLines()
 		// Special case: user wants to hide every line in between other hidden sections.
 		// Both "while" loops are executed (merge with above AND below hidden section):
 
-		while (scope == 0)
+		while (scope == 0 && startMarker < nbLines)
 			removeMarker(--startMarker, 1 << MARK_HIDELINESBEGIN);
 
-		while (scope != 0)
+		while (scope != 0 && endMarker < nbLines)
 			removeMarker(++endMarker, 1 << MARK_HIDELINESEND);
 	}
 	else
@@ -3994,10 +3994,10 @@ void ScintillaEditView::hideLines()
 		// User wants to hide some lines below/above other hidden section.
 		// If true, only one "while" loop is executed (merge with adjacent hidden section):
 
-		while (scope < 0)
+		while (scope < 0 && startMarker < nbLines)
 			removeMarker(--startMarker, 1 << MARK_HIDELINESBEGIN);
 
-		while (scope > 0)
+		while (scope > 0 && endMarker < nbLines)
 			removeMarker(++endMarker, 1 << MARK_HIDELINESEND);
 	}
 
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -3968,9 +3968,12 @@ void ScintillaEditView::hideLines()
 	size_t startMarker = startLine - 1;
 	size_t endMarker = endLine + 1;
 
-	// Remove all previous markers in between new ones
+	// Previous markers must be removed in the selected region:
+
+	removeMarker(startMarker, 1 << MARK_HIDELINESBEGIN);
 	for (size_t i = startLine; i <= endLine; ++i)
 		removeMarker(i, (1 << MARK_HIDELINESBEGIN) | (1 << MARK_HIDELINESEND));
+	removeMarker(endMarker, 1 << MARK_HIDELINESEND);
 
 	// When hiding lines just below/above other hidden lines,
 	// merge them into one hidden section:


Adrian104:
Despite the fix, program may still crash/freeze in some other rare cases when hiding lines. I'll try to investigate this issue and when everything is working properly, I'll convert this draft to pull request. Sorry for the inconvenience.

Adrian104:
It seems like there exists a problem when deleting hidden lines (or adjacent lines to them). A detailed explanation can be found here: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15630#issuecomment-2392024991

My PR unfortunately doesn't fix that. Instead, it improves reliability of the `hideLines()` method, so that it won't produce an invalid state of the markers in the first place. Still, invalid state can happen after deletion, but it originates from somewhere else.

I think that this PR is ready for a review. It will not fix everything, but the primary bug (in #15630) should be fixed.

donho (changes requested):
Please follow the suggestion.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -3968,9 +3968,12 @@ void ScintillaEditView::hideLines()
 	size_t startMarker = startLine - 1;
 	size_t endMarker = endLine + 1;
 
-	// Remove all previous markers in between new ones
+	// Previous markers must be removed in the selected region:
+
+	removeMarker(startMarker, 1 << MARK_HIDELINESBEGIN);
 	for (size_t i = startLine; i <= endLine; ++i)
 		removeMarker(i, (1 << MARK_HIDELINESBEGIN) | (1 << MARK_HIDELINESEND));
+	removeMarker(endMarker, 1 << MARK_HIDELINESEND);
```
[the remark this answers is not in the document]
Adrian104: @donho
Thank you for the feedback!
The reason I didn't combine 3 `removeMarker()` calls into a single `for` loop is because it would slightly change the way lines are hidden — not better or worse, just different. Let's say we have the following situation:



The question is: what should happen if we try to hide line number 7 (just that line)? Currently, and after my PR, the program hides just line number 7:



However, after following your suggestion, the program will behave like this:



Instead of overlapping begin and end markers, it will also merge the hidden section and hide line number 6. In other words, this suggestion would undo #13681.

Why this happens?



Hiding line number 7 will not only try to remove the **non-existent begin** marker on line number 6, but also the **existing end** marker on the same line and this behavior will propagate further upwards, causing the merge effect.

Please let me know what you think, and I'll align myself with your opinion.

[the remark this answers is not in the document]
donho: Yes, indeed.
After playing your crash example, I realize the difference.
OK, let's keep the current code.

donho (changes requested):
Please fix the infinit loop (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15630#issuecomment-2420033254) in this PR.

Adrian104:
I've added additional checks in the `while` loops. This fix applies to all of them, because I think that invalid values could theoretically happen in `startMarker` and `endMarker` in both `if` control flows.
I check `startMarker < nbLines` despite `startMarker` being decremented, but `startMarker` is unsigned, so it will just overflow and `startMarker < nbLines` should evaluate to false.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -3980,21 +3983,21 @@ void ScintillaEditView::hideLines()
 		// Special case: user wants to hide every line in between other hidden sections.
 		// Both "while" loops are executed (merge with above AND below hidden section):
 
-		while (scope == 0)
+		while (scope == 0 && startMarker < nbLines)
 			removeMarker(--startMarker, 1 << MARK_HIDELINESBEGIN);
```
donho, line 3987: How `--startMarker` can lead to condition `startMarker < nbLines` ??

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -3980,21 +3983,21 @@ void ScintillaEditView::hideLines()
 		// Special case: user wants to hide every line in between other hidden sections.
 		// Both "while" loops are executed (merge with above AND below hidden section):
 
-		while (scope == 0)
+		while (scope == 0 && startMarker < nbLines)
 			removeMarker(--startMarker, 1 << MARK_HIDELINESBEGIN);
 
-		while (scope != 0)
+		while (scope != 0 && endMarker < nbLines)
```
donho, line 3989: Even it can stop the infinit loop, it doesn't make sense to me.
Could you explain the situation in the comment:
```cpp
// When hiding lines just below/above other hidden lines,
// merge them into one hidden section:
```
?

donho (changes requested):
Please consider the review comment.

donho:
> I check startMarker < nbLines despite startMarker being decremented, but startMarker is unsigned, so it will just overflow and startMarker < nbLines should evaluate to false.

It's a very bad practice IMO: Why should we loop inefficiently if we can do less operations??

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
Adrian104: This could possibly happen if control flow enters `while` loop when there is an invalid state of the markers. In a nutshell, begin/end markers don't match -> this leads to `while` loops searching for the paired marker -> because paired marker is not found, `scope` variable is not changing and program freezes.
Variable `startMarker` can cause issues, because if there is end marker, but no begin marker, `startMarker` will keep decrementing (because it searches for the begin marker), starting from some line X, then 0, then overflow will occur. When we check `startMarker < nbLines`, then we basically can detect overflow.

Adrian104:
@donho 
Sorry, I don't understand what should I check then.
> Controling the scope variable is not enough to stop the loop. the endMarker should be checked as well.

How should I detect that infinite loop will occur, if the only "symptom" is that `scope` is constantly the same and `endMarker` is constantly incrementing?

After that fix, the program will check every line, and if it doesn't find matching marker, it will just exit `while` loop like nothing happened. Yes, it is bad, it doesn't fix invalid state, but if you say this wastes CPU cycles, then what should I fix exactly?

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
Adrian104: When line number N is hidden, and the user wants to hide the N-1 or the N+1 line number, then the algorithm should extend the previous hidden section.

Adrian104:
> > I check startMarker < nbLines despite startMarker being decremented, but startMarker is unsigned, so it will just overflow and startMarker < nbLines should evaluate to false.
> 
> It's a very bad practice IMO: Why should we loop inefficiently if we can do less operations??

I will also add that we don't check between 0 and the maximum value `size_t` can handle, but rather we check between 0 and N, where N is the number of lines in the document.

Without my fix, this situation where `while` loops iterate over a lot of lines could still happen, when the user have huge hidden sections and wants to hide some extra lines just below/above them.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
donho: Then please use the type converting:
`while (scope == 0 && static_cast<intptr_t>(startMarker) < 0)`

donho (changes requested):
Please use the type converting:
`while (scope == 0 && static_cast<intptr_t>(startMarker) < 0)`
when the variable is decreasing.

Adrian104:
Done, it should be better now - but I think you meant this:
`while (scope == 0 && static_cast<intptr_t>(startMarker) >= 0)`
not this:
`while (scope == 0 && static_cast<intptr_t>(startMarker) < 0)`

donho (approved):

PR #15635, opened 2024-09-16
[xml] Update french 8.6.9

Need74:
Hoping that this last proposal will allow to finalize this PR

### DIFF
--- a/PowerEditor/installer/nativeLang/french.xml
+++ b/PowerEditor/installer/nativeLang/french.xml
@@ -25,7 +25,7 @@ Translation note:
 				</Entries>
 				<!-- Sub Menu Entries -->
 				<SubEntries>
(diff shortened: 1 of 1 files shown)


donho:
@Need74 Thank you for your PR.
Reviewing a huge modification is time consuming, and the new commits for replying of modification request need to be review the verification on the whole PR - it is not an efficient way to do it.

As a result, this PR is closed, and I suggest the following method: 
1. Do a PR in which you can put the modification that I've never request the modification. Make sure it won't exceed 20 items.
2. Once the PR is validated and merged, repeat 1 & 2
3. do the remain part, with which the modification has been requested.  Make sure it won't exceed 20 items.
4. Once the PR is validated and merged, repeat 3 & 4

So that such review process is easier for both of us .

PR #15658, opened 2024-09-27
Fix opened network files hanging issue while the network disconnected

### LINKED ISSUE #4306: Feature Request: Configurable Timeout for File Status Auto-Detection
_Short version:_ I'd like to be able to configure the timeout for the file status auto-detection that happens when Notepad++ gets back focus.

_Background:_ Sometimes I will open a file on the network at work and put my laptop to sleep without closing the file. If I open my laptop off of the network later and go back to Notepad++, the whole program freezes while it tries to reach the network share that the file is on. It seems to take about 30 seconds for it to finally give up. And I can't even do anything else during that time, because if I leave Notepad++ and come back after 30 seconds, the whole process starts again.

So I'd like to be able to shorten that timeout. I'd probably set it to something really short, like 2 seconds.

The other option is to maybe do that auto-detection on a background thread so the UI isn't locked up.

### LINKED ISSUE #6178: Npp freezez when oppening workspace that is not available due to network
### Description of the Issue
I use View Foldr as Workspace feature, and the folder is on a network drive.

When I'm outside my home network and that given folder is not available, Npp does not start, it freezes while trying to open the folder.

1. Open a network folder as a workspace. See that it functions as expected
2.  Close Npp, close computer
3.  Connect the computer to another network, srat windows and open Npp
4. Npp freezes

### Expected Behavior
Npp would start without freezing.
perhaps informing that workspace folder is not available, or maybe silent,

Anyway, user should not be expected to remove folder from workspace every time before closing Npp, if there is expected to be some change in network.

### Actual Behavior
Described above

### LINKED ISSUE #8055: Notepad++ not responding state when network file is disconnected
### Description of the Issue

Notepad++ not responding state when network file is opened and the network location is no longer available

1. Open a network located file in notepad++
2. Disconnect the network drive or lose connectivity to the network location
3. Open notepad again or make the window active

### Expected Behavior

A notification that the network drive is no longer available. A faster response.

### Actual Behavior

Notepad++ went to not responding state.

### LINKED ISSUE #11388: Freezing when internet lost to network drive file
### Description of the Issue
When I have a network drive file open and the connection is lost, Notepad++ freezes before showing a dialogue box. This was the case in the past and was annoying. In the past, though, once the dialogue box was dealt with, there was no more freezing. At the moment, there's freezing each time there's a change in focus to Notepad++. 

1. Open a network drive file
2. Lose internet connection to the network drive
3. Change focus to the file

### Expected Behavior
I don't think Notepad++ should be freezing. Can it not check for connection to the drive independently of being able to edit the file? And I think it shouldn't be doing it each time there's a change in focus. 

### Actual Behavior
There's a freezing each time there's a change in focus to a file in a network drive where access has been severed. 

### Debug Information
Notepad++ v8.2.1   (64-bit)
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 (64-bit) 
OS Version : 2009
OS Build : 22000.556
Current ANSI codepage : 1252

### LINKED ISSUE #12553: Notepad++ hangs when network files are unreachable
### Description of the Issue
If you have a networked file open and you close notepad++, then disconnect the resource of the file notepad++ will crash in an infinity loop when you try to open the program. (White screen)

1. Open for instance a smb networked file
2. Close notepad++
3. disconnect the smb resource
4. open notepad++

### Expected Behavior
Should open normaly and maybe give a message saying the files cant be reached and if it should discard them 

### Actual Behavior
Program hangs with white screen, never reached a timeout.

### Debug Information
not possible.

### LINKED ISSUE #15540: [BUG] Process hangs if network resource unavailable

### Description of the Issue

Notepad++ hangs if the process is attempting to edit a file that resides on a network resource that is unavailable for some reason

### Steps To Reproduce

1.  Attempt to edit a previously opened file that resides on a network resource THAT IS DOWN.  For example: \\NAS1\volume1\home\stuff\workingonit.txt
2. NotePad++ hangs and becomes unresponsive.
3. Attempt to close NPP with the upper right corner "X".
4. "NotePad++ is not reponding" pop-up appears.
5.  "Close the Program" is selected in order to force close NPP.
6. Attempts to reopen NPP also hang with the same sequence.

### Current Behavior

see steps to repoduce

### Expected Behavior

some other sort of error message

### Debug Information

```shell
Unavailable - NPP is currently totally unresponsive until I either uninstall/reinstall or somehow get the network resource to return online
```

### Anything else?

_No response_

donho:
The issue is due to WinAPI's `GetFileAttributes` function can take a huge amount of time in various situations, like when the target server being offline (usalaly). The current solution is to wait at least 3 seconds for each `GetFileAttributes` function, then return the default value in case of the problem (or the returned value of the function, in case of normal situation - if it has been executed completely during the 3 seconds). So there will still be the hanging time (3 or 6 seconds) while the problem of network connection, but the hanging time (it could be 30 seconds more) is reduced considerably.

`Wow64EnableWow64FsRedirection` call is also removed from x64 build (in which this call is unnecessary) in this PR to reduce the IO calls.

Fix #4306, fix #6178, fix #8055, fix #11388, fix #12553, fix #15540


donho:
@xomx 
Please review the PR.

xomx:
***@***.*** Until Sunday evening, I am only on a mobile phone, so excuse me if I did not get something right.I see two problems:1) You fire up a worker thread for the file attributes check and rely on the fact that after a possible delay, caused by a missing network resource, the worker thread resumes ok and finishes itself. You resigned on any control if that happens so. So in a case,  when such worker thread freezes for some unpredictable reason for good, you will have a big memory leak (every thread has its own stack allocated...). 2) Bigger issue is that if the worker thread freezes, waiting for an unavailable network resource, the main thread continues after 3 secs, so theoretically after the worker resumes and the GetFileAttributes WINAPI tries to finish its work and stores the file attributes to the passed struct, that struct could not exist at that time anymore, as the parent  getFileAttrWaitFewSec already exited and destroyed its vars passed to the worker long time before...Dne 28. 9. 2024 4:39 napsal uživatel Don HO ***@***.***>:
@xomx
Please review the PR.

donho:
@xomx 
> 2) Bigger issue is that if the worker thread freezes, waiting for an unavailable network resource, the main thread continues after 3 secs, so theoretically after the worker resumes and the GetFileAttributes WINAPI tries to finish its work and stores the file attributes to the passed struct, that struct could not exist at that time anymore, as the parent  getFileAttrWaitFewSec already exited and destroyed its vars passed to the worker long time before...

Good point.
Making `GetAttrParamResult data;` as **static** (in the latest commit) solves the problem mentioned above. 

> 1) You fire up a worker thread for the file attributes check and rely on the fact that after a possible delay, caused by a missing network resource, the worker thread resumes ok and finishes itself. You resigned on any control if that happens so. So in a case,  when such worker thread freezes for some unpredictable reason for good, you will have a big memory leak (every thread has its own stack allocated...). 

The thread solution is for solving the hanging due to the inaccessible IO with network problem, and all the reports of hanging are about network connection problem. So ***when such worker thread freezes for some unpredictable reason for good*** situation cannot happen unlikely. I would say, let's see the situation above happen, then we find another solution.

xomx:
@donho 
> Making `GetAttrParamResult data;` as **static**

Are you sure that there cannot be two such concurrent worker threads (== another `getFileAttrWaitFewSec` call in action) at the same time?  (`static` var will be shared...)

donho:
@xomx 
> Are you sure that there cannot be two such concurrent worker threads (== another `getFileAttrWaitFewSec` call in action) at the same time? (`static` var will be shared...)

Yes, there will be the race condition.
In the latest commit, `static` is removed, and TerminateThread is added after WaitForSingleObject. So [2 issues you've mentioned](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15658#issuecomment-2380880836) are solved in the latest commit.

Note that I tried to replace **TerminateThread** by **CancelSynchronousIo**, but the later one has no effect - the cancel IO time is like the time waiting for **GetFileAttributes** finishes.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1764,60 +1764,105 @@ bool Version::isCompatibleTo(const Version& from, const Version& to) const
...
-	static GetAttrParamResult data;
+	GetAttrParamResult data;
 	data._fileAttr = INVALID_FILE_ATTRIBUTES;
 	data._filePath = filePath;
+	data._isNetworkFailure = true;
+
 	HANDLE hThread = ::CreateThread(NULL, 0, getFileAttributesWorker, &data, 0, NULL);
 	if (!hThread)
 	{
 		return false;
```
xomx, line 1791: return INVALID_FILE_ATTRIBUTES;

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1764,60 +1764,105 @@ bool Version::isCompatibleTo(const Version& from, const Version& to) const
...
 	}
 
-	// Wait for 3 sec
-	::WaitForSingleObject(hThread, 3000);
+	// Wait for few sec
+	::WaitForSingleObject(hThread, milleSec2wait == 0 ? 1000 : milleSec2wait);
+	TerminateThread(hThread, 2);
```
xomx, line 1796: For threads, the TerminateThread is a really nasty shutdown, so if possible it should not be a part of the usual code flow.

So I would use something like (you said the `CancelSynchronousIo` WINAPI does not work for you - in that case, delete my WAIT_TIMEOUT part entirely):

// wait for our worker thread to complete or cancel it when the required timeout has elapsed
DWORD dwWaitStatus = ::WaitForSingleObject(hThread, milleSec2wait == 0 ? 1000 : milleSec2wait);
switch (dwWaitStatus)
{
	case WAIT_OBJECT_0:
		// Ok, the state of our worker thread is signaled, so it finished itself in the timeout given
		// - nothing else to do here, except the thread handle closing later
		break;

	case WAIT_TIMEOUT:
		// the timeout interval elapsed, but the worker's state is still non-signaled
		// - try to cancel our worker thread's synchronous IO from here
		if (!::CancelSynchronousIo(hThread))
		{
			// if the CancelSynchronousIo cannot find a request to cancel,
			// its return value is 0 and the GetLastError() returns ERROR_NOT_FOUND
			dwRet = ::GetLastError();
			if (dwRet != ERROR_NOT_FOUND)
				::TerminateThread(hThread, WAIT_TIMEOUT);
		}
		break;

		// any other dwWaitStatus is a BAD one here
	default:
		// WAIT_FAILED or WAIT_ABANDONED
		::TerminateThread(hThread, dwWaitStatus);
		break;
}

xomx:
> I tried to replace **TerminateThread** by **CancelSynchronousIo**, but the later one has no effect - the cancel IO time is like the time waiting for **GetFileAttributes** finishes.

Weird - the `CancelSynchronousIo` function should not wait for the canceled operations to complete.

From the MS doc: "Most types of operations can be canceled immediately; other operations can continue toward completion before they are actually canceled and the caller is notified. The CancelSynchronousIo function `does not wait` for all canceled operations to complete."

So the  `CancelSynchronousIo` itself should return immediately, at least I understand it that way.

PowerEditor/src/MISC/Common/Common.cpp
xomx: @donho 
> To reproduce the delay, you just need to open 1 network file then cut off the connection, open some new tab, make it dirty, and close the dirty tab. When the focus come back to the disconnected network file, you'll see the huge delay.

Thank you, with this STR I am finally able to reproduce the issue steadily (my previous attempts mostly ended up with immediate "Keep non existing file" dlg).

I will recheck the `CancelSynchronousIo` sln against your latest commits and let you know.

donho:
> From the MS doc: "Most types of operations can be canceled immediately; other operations can continue toward completion before they are actually canceled and the caller is notified. The CancelSynchronousIo function does not wait for all canceled operations to complete."
> So the `CancelSynchronousIo` itself should return immediately, at least I understand it that way.

@xomx 
I did understand that way as well.
However, by adding `CancelSynchronousIo(hThread)` line, it made all the improvement in vain. 

In [the example provided by Microsoft](https://learn.microsoft.com/en-us/windows/win32/fileio/canceling-pending-i-o-operations), the worker calls `CreateFileA` but not `GetFileAttributesW`. That could make the difference.

OTOH, I would believe in the MS online doc, but I trust more my own experience - so I prefer not use it because adding it doesn't fix at all the issue.

xomx:
@donho 

> In [the example provided by Microsoft](https://learn.microsoft.com/en-us/windows/win32/fileio/canceling-pending-i-o-operations), the worker calls `CreateFileA` but not `GetFileAttributesW` (in our case) - which might make the difference.

I tried that but it is not any better (I rewrote your PR with its latest commits to use the `CreateFileW` & then consequently the [GetFileInformationByHandle](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileinformationbyhandle) on the opened `hFile` to completely replace the `GetFileAttributesW` functionality and to use the `CancelSynchronousIo` when the `case WAIT_TIMEOUT:` ). The annoying delay is still there, so let's forget the `CancelSynchronousIo` way.

The currently used ThreadTerminate-way works for me. But I detected another biiig-lag (`_wfopen`-one) when I opened first a network file in N++, closed N++, unplugged cable from my server and relaunched N++, callstack at the lag-time when I "BreakAll":

```
 	ntdll.dll!00007fff4004e004()	Unknown
 	KernelBase.dll!00007fff3d8a1a09()	Unknown
 	KernelBase.dll!00007fff3d8a1476()	Unknown
 	notepad++.exe!create_file(const wchar_t * const path, _SECURITY_ATTRIBUTES * const security_attributes, const `anonymous-namespace'::file_options options) Line 234	C++
 	notepad++.exe!_wsopen_nolock(int * punlock_flag, int * pfh, const wchar_t * path, int oflag, int shflag, int pmode, int secure) Line 702	C++
 	notepad++.exe!__crt_char_traits<wchar_t>::tsopen_nolock<int *,int * const &,wchar_t const * const &,int const &,int const &,int const &,int const &>(int * && <args_0>, int * const & <args_1>, const wchar_t * const & <args_2>, const int & <args_3>, const int & <args_4>, const int & <args_5>, const int & <args_6>) Line 122	C++
 	notepad++.exe!common_sopen_dispatch<wchar_t>(const wchar_t * const path, const int oflag, const int shflag, const int pmode, int * const pfh, const int secure) Line 172	C++
 	notepad++.exe!_wsopen_dispatch(const wchar_t * path, int oflag, int shflag, int pmode, int * pfh, int secure) Line 216	C++
 	notepad++.exe!_wsopen_s(int * pfh, const wchar_t * path, int oflag, int shflag, int pmode) Line 907	C++
 	notepad++.exe!__crt_char_traits<wchar_t>::tsopen_s<int *,wchar_t const * const &,int const &,int const &,int>(int * && <args_0>, const wchar_t * const & <args_1>, const int & <args_2>, const int & <args_3>, int && <args_4>) Line 122	C++
 	notepad++.exe!common_openfile<wchar_t>(const wchar_t * const file_name, const wchar_t * const mode, const int share_flag, const __crt_stdio_stream stream) Line 38	C++
 	notepad++.exe!_wopenfile(const wchar_t * file_name, const wchar_t * mode, int share_flag, _iobuf * public_stream) Line 77	C++
 	notepad++.exe!__crt_char_traits<wchar_t>::open_file<wchar_t const * const &,wchar_t const * const &,int const &,_iobuf *>(const wchar_t * const & <args_0>, const wchar_t * const & <args_1>, const int & <args_2>, _iobuf * && <args_3>) Line 122	C++
 	notepad++.exe!common_fsopen<wchar_t>(const wchar_t * const file_name, const wchar_t * const mode, const int share_flag) Line 54	C++
 	notepad++.exe!_wfopen(const wchar_t * file, const wchar_t * mode) Line 130	C++
>	notepad++.exe!FileManager::loadFile(const wchar_t * filename, __int64 doc, int encoding, const wchar_t * backupFileName, _FILETIME fileNameTimestamp) Line 712	C++
 	notepad++.exe!Notepad_plus::doOpen(const std::wstring & fileName, bool isRecursive, bool isReadOnly, int encoding, const wchar_t * backupFileName, _FILETIME fileNameTimestamp) Line 442	C++
 	notepad++.exe!Notepad_plus::loadSession(Session & session, bool isSnapshotMode, const wchar_t * userCreatedSessionName) Line 2347	C++
 	notepad++.exe!Notepad_plus::loadLastSession() Line 2295	C++
 	notepad++.exe!Notepad_plus_Window::init(HINSTANCE__ * hInst, HWND__ * parent, const wchar_t * cmdLine, CmdLineParams * cmdLineParams) Line 173	C++
 	notepad++.exe!wWinMain(HINSTANCE__ * hInstance, HINSTANCE__ * __formal, wchar_t * pCmdLine, int __formal) Line 712	C++

``` 

The last N++ inherent func is `_wfopen` from here:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/fc051a1231065731205ff5e315684ce3bfb19033/PowerEditor/src/ScintillaComponent/Buffer.cpp#L707

I do not have simple sln for this _wfopen-lag.

I also have some minor review-things, I will do them later.

donho:
@xomx 
> But I detected another biiig-lag (_wfopen-one) when I opened first a network file in N++, closed N++, unplugged cable from my server and relaunched N++, callstack at the lag-time when I "BreakAll":

> The last N++ inherent func is _wfopen from here:

Are you sure about it?
I cannot reproduce it and here's the code:

```cpp
BufferID FileManager::loadFile(const wchar_t* filename, Document doc, int encoding, const wchar_t* backupFileName, FILETIME fileNameTimestamp)
{
	if (!filename)
		return BUFFER_INVALID;

	//Get file size
	int64_t fileSize = -1;
	const wchar_t* pPath = filename;
	if (!doesFileExist(pPath))                 // block 1
	{
		pPath = backupFileName;
	}

	if (pPath)                                         // block 2
	{
		FILE* fp = _wfopen(pPath, L"rb");
		if (fp)
		{
			_fseeki64(fp, 0, SEEK_END);
			fileSize = _ftelli64(fp);
			fclose(fp);
		}
	}
...
```
Logically if the `filename` is a disconnected network file *\\myNetwork\tmp\aFile.txt*, we will enter into block1, and `pPath` will be:
1. empty: if aFile.txt was NOT dirty in the previous session
2. *\\myNetwork\tmp\aFile.txt@2024-09-30_122334*: if aFile.txt WAS dirty in the previous session

In the 1st case, we won't enter into the block 2, so _wfopen won't be called.
In the 2nd case, we will enter into the block 2, _wfopen call will have no issue since *\\myNetwork\tmp\aFile.txt@2024-09-30_122334* does exist.

Have you made a break point on `FILE* fp = _wfopen(pPath, L"rb");` and notice the latency between `FILE* fp = _wfopen(pPath, L"rb");` and  `if (fp)` ?

donho:
@xomx 

```cpp
BufferID FileManager::loadFile(const wchar_t* filename, Document doc, int encoding, const wchar_t* backupFileName, FILETIME fileNameTimestamp)
{
	if (!filename)
		return BUFFER_INVALID;

	//Get file size
	int64_t fileSize = -1;
	const wchar_t* pPath = filename;
	writeLog(L"c:\\tmp\\networkIssue.log", "1");
	if (!doesFileExist(pPath))
	{
		writeLog(L"c:\\tmp\\networkIssue.log", "2");
		pPath = backupFileName;
	}
	writeLog(L"c:\\tmp\\networkIssue.log", "3");
	if (pPath)
	{
		writeLog(L"c:\\tmp\\networkIssue.log", "4");
		FILE* fp = _wfopen(pPath, L"rb");
		writeLog(L"c:\\tmp\\networkIssue.log", "5");
		if (fp)
		{
			writeLog(L"c:\\tmp\\networkIssue.log", "6");
			_fseeki64(fp, 0, SEEK_END);
			fileSize = _ftelli64(fp);
			fclose(fp);
			writeLog(L"c:\\tmp\\networkIssue.log", "7");
		}
	}
	writeLog(L"c:\\tmp\\networkIssue.log", "8");
```
Here's the log:
```
2024-10-02 00:32:48  1
2024-10-02 00:32:49  2
2024-10-02 00:32:49  3
2024-10-02 00:32:49  4
2024-10-02 00:32:49  5
2024-10-02 00:32:49  6
2024-10-02 00:32:49  7
2024-10-02 00:32:49  8
2024-10-02 00:34:00  1
2024-10-02 00:34:01  3
2024-10-02 00:34:01  4
2024-10-02 00:34:01  5
2024-10-02 00:34:01  6
2024-10-02 00:34:01  7
2024-10-02 00:34:01  8
```

There's indeed more than 1 minute of latency between 2 loads of file. But obviously, the culprit is not `_wfopen`.
I keep tracing it...

donho:
@xomx 
```
2024-10-02 00:34:00  1
2024-10-02 00:34:01  3
2024-10-02 00:34:01  4
2024-10-02 00:34:01  5
2024-10-02 00:34:01  6
2024-10-02 00:34:01  7
2024-10-02 00:34:01  8
```
Stupid me! the above log are the traces of "networkIssue.log" opened in Notepad++.

donho:
@xomx 

With the confirmation of https://github.com/notepad-plus-plus/notepad-plus-plus/issues/4306#issuecomment-2386529017 & https://github.com/notepad-plus-plus/notepad-plus-plus/issues/6178#issuecomment-2386524585, I think the PR can be merged.

Here is the log on opening only one offline network file (dirty):
```
2024-10-02 01:48:00  BEGIN
2024-10-02 01:48:02  1 \\MYNETWORK\tmp\NUL_char.txt
2024-10-02 01:48:03  2 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:03  3 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:03  4 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:03  5 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:03  6 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:03  7 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:03  8 C:\Users\donh\AppData\Roaming\Notepad++\backup\NUL_char.txt@2024-10-01_234312
2024-10-02 01:48:08  End of loadFile
2024-10-02 01:48:15  END
```

As you can see in the log, there is still room to improve, I'll keep working on it.

donho:
@xomx 
> I detected another biiig-lag (`_wfopen`-one) when I opened first a network file in N++, closed N++, unplugged cable from my server and relaunched N++, callstack at the lag-time when I "BreakAll":

The issue has been fixed by PR #15681.

PR #15679, opened 2024-10-05
Fix network files hanging while the network disconnected (part 3)

donho:
Add thread for CreateFile & log file (commented) for debugging to fix saving disconnected network file hanging. 

STR: Open a network file, modify it. Disconnect the network, then save the file. A huge latency (more than 1 minute) can be observed.

Not that the crash is not reproducible anymore by this PR. If any crash happens for you, please let me know (with the STR).

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15669

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1977,7 +1977,7 @@ HANDLE createFileWaitSec(const wchar_t* filePath, DWORD accessParam, DWORD share
 	HANDLE hThread = ::CreateThread(NULL, 0, createFileWorker, &data, 0, NULL);
 	if (!hThread)
 	{
-		return FALSE;
+		return INVALID_HANDLE_VALUE;
 	}
 
 	// wait for our worker thread to complete or terminate it when the required timeout has elapsed
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1945,3 +1945,59 @@ DWORD getFileAttributesExWaitSec(const wchar_t* filePath, WIN32_FILE_ATTRIBUTE_D
 
 	return data._result;
 }
+
+
+//----------------------------------------------------
+
+struct CreateFileParamResult
+{
+	wstring _filePath;
+	HANDLE _hFile = INVALID_HANDLE_VALUE;
+	DWORD _accessParam = GENERIC_READ | GENERIC_WRITE;
+	DWORD _shareParam = FILE_SHARE_READ | FILE_SHARE_WRITE;
+	DWORD _dispParam = CREATE_ALWAYS;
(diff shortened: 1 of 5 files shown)


PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1945,3 +1945,59 @@ DWORD getFileAttributesExWaitSec(const wchar_t* filePath, WIN32_FILE_ATTRIBUTE_D
...
+};
+
+HANDLE createFileWaitSec(const wchar_t* filePath, DWORD accessParam, DWORD shareParam, DWORD dispParam, DWORD attribParam, DWORD milliSec2wait, bool* isNetWorkProblem)
+{
+	CreateFileParamResult data(filePath, accessParam, shareParam, dispParam, attribParam);
+
+	HANDLE hThread = ::CreateThread(NULL, 0, createFileWorker, &data, 0, NULL);
+	if (!hThread)
+	{
+		return FALSE;
```
xomx, line 1980: return INVALID_HANDLE_VALUE;

donho:
@xomx 
The crash come back :(

With random manipulation on release binary of the PR, I've got this:

image

But with the following STR, in debug mode I have the crash, I do believe it's the same scenario:
1. Open a network file, modify it. Disconnect the network, then save the file. 
2. The "Saved failed" dialog with the message *"Please check the network where the file is located is connected."* appears, click OK.
3. Click **X** button on the file in question to close it, then the dialog with "The file "xxxxxx" doesn't exist anymore. Keep this file in Editor?" appear, click "No".
4. Crash with "Access violation".

donho:
Also, during the test with this PR, I have seen once the zombie, but it didn't lasted for the eternity.

donho:
BTW, I just replace `::TerminateThread(hThread, dwWaitStatus);` by `CancelSynchronousIo(hThread);`. It seems work better the `CancelSynchronousIo` call. But it leads the same crash.

donho:
> But with the following STR, in debug mode I have the crash, I do believe it's the same scenario:
> 
>     1. Open a network file, modify it. Disconnect the network, then save the file.
> 
>     2. The "Saved failed" dialog with the message _"Please check the network where the file is located is connected."_ appears, click OK.
> 
>     3. Click **X** button on the file in question to close it, then the dialog with the message _"The file "xxxxxx" doesn't exist anymore. Keep this file in Editor?"_ appear, click "No".
> 
>     4. Crash with "Access violation".

Finally, the crash is not due to this PR. It's the regression of 2 disconnection hanging enhancement, and the crash can be reproduced by the current master as well.
A new issue is created here: #15684, and the solution PR is here:  #15685

donho:
Finally, adding `CancelIoEx(data._hFile, NULL);` seems remedy the process persists in the memory after quitting.
The results of test are variable but globally I can see the improvement (in average less 20 seconds remained in memory).

PR #15681, opened 2024-10-07
Improve code & refactoring

donho:
~~Fix network file hanging issue due to win32API GetFileAttributes cache (unsynchronized).~~

~~STR:~~
~~1. Open a network file.~~
~~2. Close Notepad++ to have it in the session.~~
~~3. Disconnect the network, and launch Notepad++ immediately.~~
~~4. Around more than 1 minute's delay, then the "Error" dialog displayed.~~

~~The reason of hanging is that the network file was detected by "doesFileExist" as true, so Notepad++ was trying to open non-existent file (by _wfopen). I believe that there's some kind of cache during the very short period for the IO function (here's our case GetFileAttributes), and such cache is not immediately synchronized (cleared) while network disconnected. As a result, when we launch Notepad++ after the disconnection of network, GetFileAttributes keeps its memory & responds "FileExists". However for _wfopen it can't find the resource of network anymore - that makes hanging.~~

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15658#issuecomment-2386662974

~~Improve #4306, #6178, #8055, #11388, #12553, #15540~~

***Edit: due to the zombie process persistent issue due to the PR, the timeout thread won't be applied to _wfopen***


donho:
Log file before the fix:

```
2024-10-07 02:20:45  \\XXX\tmp\scintilla\bin\empty.txt	 BEFORE doesFileExist
2024-10-07 02:20:45  \\XXX\tmp\scintilla\bin\empty.txt	 AFTER doesFileExist
2024-10-07 02:20:46  \\XXX\tmp\scintilla\bin\empty.txt	 BEFORE getFileAttributesExWaitSec
2024-10-07 02:20:46  \\XXX\tmp\scintilla\bin\empty.txt	 getFileAttributesExWaitSec OK - let's retrieve file length
2024-10-07 02:20:46  \\XXX\tmp\scintilla\bin\empty.txt	 AFTER getFileAttributesExWaitSec
2024-10-07 02:20:46  \\XXX\tmp\scintilla\bin\empty.txt	 BEFORE loadFileData
2024-10-07 02:20:46  \\XXX\tmp\scintilla\bin\empty.txt	 BEFORE _wfopen
2024-10-07 02:22:08  \\XXX\tmp\scintilla\bin\empty.txt	 AFTER _wfopen
2024-10-07 02:22:08  \\XXX\tmp\scintilla\bin\empty.txt	 AFTER loadFileData
```

Log file after the fix:

```
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 BEFORE doesFileExist
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 AFTER doesFileExist
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 BEFORE getFileAttributesExWaitSec
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 getFileAttributesExWaitSec OK - let's retrieve file length
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 AFTER getFileAttributesExWaitSec
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 BEFORE loadFileData
2024-10-07 02:58:12  \\xxx\tmp\scintilla\bin\empty.txt	 BEFORE _wfopen
2024-10-07 02:58:17  \\xxx\tmp\scintilla\bin\empty.txt	 AFTER _wfopen
2024-10-07 02:58:17  \\xxx\tmp\scintilla\bin\empty.txt	 AFTER loadFileData
```

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1837,6 +1837,58 @@ bool doesPathExist(const wchar_t* path, DWORD milliSec2wait, bool* isNetWorkProb
...
+};
+
+FILE* wfopenWaitSec(const wchar_t* filePath, DWORD milliSec2wait, bool* isNetWorkProblem)
+{
+	wfopenParamResult data(filePath);
+
+	HANDLE hThread = ::CreateThread(NULL, 0, wfopenWorker, &data, 0, NULL);
+	if (!hThread)
+	{
+		return FALSE;
```
xomx, line 1866: rather `return nullptr;`

even though the `FALSE` is also numerically ok here

xomx:
@donho 
My reported `_wfopen` big-lag issue is fixed but there seems to be now a serious problem at the N++ exit (N++ deadlocked at the CRT internal exit). Investigating...

PowerEditor/src/MISC/Common/Common.cpp
donho: Well spotted!

donho:
@xomx 
> I dl kernel PDB symbols for my Windows and here is the callstack of my N++ exit deadlock - the last ntdll!NtWaitForAlertByThreadId is apparently going to wait for the signaled state by some N++ thread forever...

Thank you for testing and reproducing the critical issue.
I will:
1. check the code merged in master to see whether if the same issue happens also.
2. try to remedy the issue of this PR, if the same issue doesn't happen in 1.

xomx:
@donho 
> * check the code merged in master to see whether if the same issue happens also.
> 
> * try to remedy the issue of this PR, if the same issue doesn't happen in 1.

I did exactly that, it is ok.
I should have added a new post instead of editing my old one above, sorry.

>Edit2:
I checked the same situation with the net-hanging commits 1 & 2: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f884a39dd422cda908b36e9439fb00c340c4d85e & https://github.com/notepad-plus-plus/notepad-plus-plus/commit/a3535f385f297a0825cad50cde7f8cfa59aca6e0 and there is not such problem visible. So the problem has to be in this part4 PR.

donho:
@xomx 
> I did exactly that, it is ok.

Thank you for checking them.
I just did the same check quickly, no issue found either.

It seems there's no similar issue for GetFileAttributes, GetFileAttributesEx & GetDiskFreeSpaceEx. I'm pretty sure that CreateFile has the same problem as the wfopen's one.

I'll check #15679 and let you know the result.

xomx:
@donho 
I probably will not have a time for anything N++ in the next two days, so I simply put my next two ideas here without really trying them out:

1) Possibly safer `ThreadTerminate` use:
```
	switch (dwWaitStatus)
	{
		case WAIT_OBJECT_0: // Ok, the state of our worker thread is signaled, so it finished itself in the timeout given		
			// - nothing else to do here, except the thread handle closing later
			break;
		case WAIT_TIMEOUT: // the timeout interval elapsed, but the worker's state is still non-signaled
		default: // any other dwWaitStatus is a BAD one here
			// WAIT_FAILED or WAIT_ABANDONED
			{
				DWORD dwDummyFlagsOut;
				if (::GetHandleInformation(hThread, &dwDummyFlagsOut) // ? still a valid thread object handle
				{
					::TerminateThread(hThread, dwWaitStatus);
				}
			}
			break;
	}
```

2) Just an idea how to improve further the user experience in this area:
We could improve the existing mechanism of waiting for the completion of the worker threads ops. We can even extend the default timeouts for ops to complete (if we want to, of course), but give the user the option to immediately abort this wait at any time by pressing the Esc-key. A code is better explaining than many words, here are the changes required:
```
#define DEFAULT_MILLISEC 200


bool userWantsToCancelWaitingByEsc()
{
	bool bRet = false;
	HWND hForegroundWnd = ::GetForegroundWindow();
	if (hForegroundWnd)
	{
		DWORD pid = 0;
		::GetWindowThreadProcessId(hForegroundWnd, &pid);
		if (pid == ::GetCurrentProcessId())
		{
			// Notepad++ app-wnd is active, now check for the possible user Esc
			bRet = (::GetAsyncKeyState(VK_ESCAPE) & 0x8000);
		}
	}
	return bRet;
}


// and use in all the relevant funcs:

	// user-cancelable wait for some worker thread op
	while (true)
	{
		DWORD dwWaitStatus = ::WaitForSingleObject(hThread, milliSec2wait == 0 ? DEFAULT_MILLISEC : milliSec2wait);
		if (dwWaitStatus == WAIT_OBJECT_0)
		{
			// Ok, the state of our worker thread is signaled, so it finished itself in the timeout given
			// - nothing else to do here, except the thread handle closing later
			break;
		}
		
		bool bTerminate = false;
		if (dwWaitStatus == WAIT_TIMEOUT)
		{
			// the timeout interval elapsed, but the worker's state is still non-signaled
			// - check if the user wants to cancel the op earlier
			if (userWantsToCancelWaitingByEsc())
				bTerminate = true;
		}
		else
		{
			// any other wait status (WAIT_FAILED, WAIT_ABANDONED) is a bad one here
			bTerminate = true;
		}
		if (bTerminate)
		{
			DWORD dwDummyFlagsOut;
			if (::GetHandleInformation(hThread, &dwDummyFlagsOut) // ? still a valid thread object handle
			{
				::TerminateThread(hThread, dwWaitStatus);
			}
			break;
		}
	}
```

donho:
@xomx 
Thank you for the idea and your thought.

Unfortunately, `::GetHandleInformation(hThread, &dwDummyFlagsOut)` doesn't make any difference.
I've tried several solutions, none of them can solve the problem of process which remains in the memory.
I will give a last try. If the zombie problem persists, this PR won't be merged.

Regarding the idea about ESC, it could be a good feature. However, since we have implemented the timeout (1 second) for each problematic function call to optimize the hanging, I don't see how it could improve more UX.
I would say, it could help without any timeout for those problematic function calls - but in this case I doubt how many users will hit ESC.
As a result, I prefer the solution that application do the optimal solution for users, instead of letting users guess what they should do.

xomx:
@donho 
I re-tested this PR before the last "Remedy" commit and it seems better than before.
There is still a chance to see the Notepad++ being in the memory after its exit but it is no longer a "permanent zombie", the temporarily frozen process disappears after a while with the closing of the last `getFileAttributesWorker` thread:

npp-exit-temp-frozen

donho:
@xomx 
Though you will not like the code, the latest status of PR fix the problem of zombie process.
As you can see, removing "TerminateThread" makes zombie process not persistent.

I do understand your concern, but we all have no idea how it work inside of this black box (It took me at least 36 hours to try everything), and  from my aspect, it does work like a charm without any consequence, as we can notice - so I consider it as solution.
Of course, let's wait and see in the next release if there's any impact with this solution - if any, we can always do somethings about it.

xomx:
@donho 
Now I tried it with the last Remedy-commit and what a coincidence - I repeatedly get an Access Violation exception due to the backup snapshot worker thread. This is probably what I also got in the [#14906](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14906#issuecomment-2071160480), but I have not been able to reproduce it (it seems to be a specific race condition where the backup snapshot worker thread is awaken to do its job right after the N++ `wWinMain` exit, but before this worker thread is autodestructed by the CRT internals.

npp-exit-backup_snapshot_access_violation_exception

> have no idea how it works inside of this black box

Hope I have a fairly accurate idea of ​​what is probably going on. At the very end of the app-exit (CRT internals, after the N++ `wWinMain` exit), the loader-lock is active which means that any attempt to terminate (or create) a thread will deadlock (since that thread would in turn try to acquire the same loader-lock...). 

So the solution could be to have a list of N++ existing worker threads and do not let the N++ `wWinMain` exits until these are finished or terminated.

donho:
@xomx 
> Edit:
Hmm, maybe luck before, now I try again and get "permanent zombie" again.

I'll try it from my side.

donho:
@xomx 
> Now I tried it with the last Remedy-commit and what a coincidence - I repeatedly get an Access Violation exception due to the backup snapshot worker thread. 

So let me put the situation more clearly: When you were testing the binary of this PR in debug mode (ie. you launched Notepad++ via VS Studio by hitting F5), Notepad++ crashed and the VS Studio's shown the crash is due to "Access violation"?
Is the above description correct?

If it is, could you provide me the following information:
1. How many files opened in Notepad++? More specifically, how many unsaved files? What are these files' length (environ)
2. Under which situation you've got crash? While you were modifying one of these files? Or while you were trying to close Notepad++?

> This is probably what I also got in the #14906, but I have not been able to reproduce it (it seems to be a specific race condition where the backup snapshot worker thread is awaken to do its job right after the N++ wWinMain exit, but before this worker thread is autodestructed by the CRT internals.

With the scenario you will provide (if you can), I will try to reproduce #14906, with & without the PR.

donho:
@xomx 
This summarizes the situation of the PR :(

image

xomx:
@donho 
Yes, I was wondering when something like this will be mentioned here :-)
To be fair, it is not just about the multithreading here, it is a deadly combination of multithreading + IO + networking.

Every time I see how the users here lightheartedly suggest multithreading (especially for something IO-related, e.g. # 8329), I wonder if they have any idea what it can entail.

OTOH - this helps me to reproduce & understand some of my previous backup worker thread crashes.

donho:
@xomx 
Thank you for your very detailed explanation for the scenario de crash. I will try to reproduce it when I get home.

I've done my last try to fix the hanging of disconnected network file due to GetAttributeEx not synchronizing on time with the disconnected network (it returns immediately TRUE even no more network connection, and the file-existing situation is true (wrong detected).

The issue is annoying but I think we will keep it as it finally. Except we find a better solution in the future.

xomx:
@donho 

> I've done my last try (similar solution in #15691) to fix

I have not even had time to really test this latest attempt yet, will try next week.

But I think you overlooked one of my comments about it:
- you have already fixed this one: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3b6f22b3572aaf4f5b81438d69d5ca32d5306d5d#r147875995
- but there was another similar note above: https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3b6f22b3572aaf4f5b81438d69d5ca32d5306d5d#r147875940

> we DID already a lot improvement regarding the disconnected network file issue.

Yes, definitely.
When I first simulated that issue, it really looked as a complete N++ crash, my comp was frozen for a very long time. It was probably mainly due to the repeated querying of the status of the net-file (probably in the `Buffer::checkFileState()`), I had a problem even closing the N++ normally. Now, the last time I tested the previous PRs, it was almost OK.

donho:
@xomx 
FYI: I've just done the latest modification in #15701 to remedy the issue of hanging of relaunching.
It works in debug mode, but not in release binary. I still keep it because the added layer of protection it cost nothing in the normal case, but in the case of problem it could save in some network situations.

Issue #12044, opened 2022-08-26
renamed / deleted file not kept in editor when in background and closing tab (red cross)


carlos-ddd:
### Description of the Issue
When a file is renamed or deleted while opened in npp, the file is usually kept and you're asked to if you want to keep the file as soon as the tab (file) is selected in the editor again. However if the deleted file's tab is not selected, the deleted file's tab can be closed (red cross) without warning.

### Steps to Reproduce the Issue
1. create file 'test1.txt'
2. create file 'test2.txt'
3. open 'test1.txt' and 'test2.txt' in npp
4. select 'test2.txt' in npp
5. delete 'test1.txt' in explorer
6. close 'test1.txt' in npp by clicking the red cross

### Expected Behavior
I'd expect the warning if I want to save the file / keep it.

### Actual Behavior
Closed without warning -> file lost

### Debug Information
Notepad++ v8.4.4   (32-bit)
Build time : Jul 15 2022 - 17:56:38
Path : C:\Program Files (x86)\Notepad++\notepad++.exe
Command Line : "D:\temp\compare\op_working.py"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 10 Enterprise (64-bit) 
OS Version : 21H2
OS Build : 19044.1889
Current ANSI codepage : 1252
Plugins : 
    ComparePlugin (2)
    mimeTools (2.8)
    NppConverter (4.4)
    NppExport (0.4)
    NPPJSONViewer (1.40)

alankilborn:
The "red cross":

![image](https://user-images.githubusercontent.com/77065706/186895469-d655bf61-9591-41d4-ae35-deb6d815f381.png)

alankilborn:
Reproducible as written.

alankilborn:
@carlos-ddd Pro-tip:  I find that due to the core nature of "tabs", i.e., clicking on them to activate, with the "red cross" I tend to accidentally hit that area and close my file.  This is not catastrophic as if the file is unsaved I get prompted to save it and can cancel.  But if the file is saved it is immediately closed and then I have to go to the trouble of reopening it.  Thus, I turn the "red cross" off in the _Preferences_ to avoid this:

![image](https://user-images.githubusercontent.com/77065706/186903383-a610a9c7-bc8a-447e-ae23-6f561ca2cdce.png)

[...]

PR #15685, opened 2024-10-08
Fix crash on closing a disconnected network file

### LINKED ISSUE #15684: [BUG] After saving failure of disconnected network file, close the file make crash

### Description of the Issue

After saving a disconnected network file with failure result due to the disconnection, closing the disconnected network file makes Notepad++ crash.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15679#issuecomment-2397556015

### Steps To Reproduce

1. Open a network file, modify it. Disconnect the network, then save the file.
2. Wait for about 1-2 minutes hanging.
3. The "Saved failed" dialog with the message "Please check the network where the file is located is connected." appears, click OK.
4. Click X button on the file in question to close it, then the dialog with the message "The file "xxxxxx" doesn't exist anymore. Keep this file in Editor?" appear, click "No".

### Current Behavior

Crash

### Expected Behavior

No Crash

### Debug Information

```shell
The crash can only be reproduced in the current master
```

### Anything else?

_No response_

donho:
Context: While the network file is disconnected, it's considered non-existent and might to be deleted.

The reason of crash: 
"doClose" on the same buffer happens twice while users clicks X button, then chooses "No" when they are asked whether if this file should be kept in editor.

Explanation:
When user clicks X button on the tab of the file in question to close it, "activateBuffer" is called, then "fileClose" is called.
* activateBuffer: Following "activeateBuffer" call in the depth, "checkFileState" is invoked - that leads the message "The file "xxxxxx" doesn't exist anymore. Keep this file in Editor?". If user clicks on No, "doClose" will close the tab and destroy the buffer.
* fileClose: in "fileClose" call, "doClose" will be invoked and try to close the already-destroyed buffer.

Solution:
Retrieve the buffer ID once again, after "activateBuffer" call, for comparing with old buffer ID. If the buffer was destroyed, the new retrieved buffer ID is not the same. 

Note: this commit fixes the crash, but it doesn't fix the misbehaviour: If user clicks on "Yes" to answer the message "The file "xxxxxx" doesn't exist anymore. Keep this file in Editor?", the tab will be still closed. I have so far no solution for it.

Fix #15684

### DIFF
--- a/PowerEditor/src/NppNotification.cpp
+++ b/PowerEditor/src/NppNotification.cpp
@@ -323,7 +323,11 @@ BOOL Notepad_plus::notify(SCNotification *notification)
 				activateBuffer(bufferToClose, iView);
 			}
 
-			if (fileClose(bufferToClose, iView))
+			BufferID bufferToClose2ndCheck = notifyDocTab->getBufferByIndex(index);
+
+			if ((bufferToClose == bufferToClose2ndCheck) // Here we make sure the buffer is the same to prevent from the situation that the buffer to be close was already closed,
+			                                             // because the precedent call "activateBuffer(bufferToClose, iView)" could finally lead "doClose" call as well (in case of file non-existent).
+				&& fileClose(bufferToClose, iView))
 				checkDocState();
 
 			break;


xomx:
@donho 
Nice find & fix!

The described problem was somehow familiar to me, so I searched a bit in the history and finally found: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/12044#issuecomment-1232721079
Your fix also solved my old problem described there (I tested before and after this PR), thanks.

PR #15691, opened 2024-10-11
Fix network files hanging while the network disconnected (part 3.5)

donho:
This PR prevent hanging when user: Open a network file, modify it. Disconnect the network, then save the file.
It also prevents the zombie process due to blocked CreateFile left behind.

Remove the timeout thread for CreateFile to prevent the zombie process. Use another way for the detection:
If the result of network file existent detection is false, and the network problem found (timeout reached), we just stop and don't call CreateFile routine.

Ref: 1445487bb76b92a3a30cf71d8175b8b416b4c0c9


donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/3b6f22b3572aaf4f5b81438d69d5ca32d5306d5d

PR #15696, opened 2024-10-12
[xml] Update serbian & serbianCyrillic to v8.7.1

rammba:
Hello @donho, I'm glad to contribute to such a great project, because I'm using Notepad++ for years.

I've noticed that serbian files are outdated. My intention is to sync them first with the english version, and then work on the 
translation.
I also want to use same terms in both latin and cyrillic version since they should be the same, but with different set of letters.

Regarding my translation work, I'm currently working on [the translation](https://github.com/reactjs/sr.react.dev) of official react documetation. Also, I've translated few Steam games as well, and I'm trying to contribute to some smaller translations.


donho:
@rammba 
> I've noticed that serbian files are outdated. My intention is to sync them first with the english version, and then work on the
translation.

I saw the 2 files of your PR - it seems that you've finished the synchronization for these 2 files.
Please keep translating directly in this PR.

rammba:
Thanks for the quick response @donho. I can do that, but not sure how long it will take and that's why I created PR only for sync. But if you think it's better to do all at once, it's fine by me.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1387,7 +1387,7 @@ Translation note:
 				<Item id="1720" name="&amp;. matches newline"/>
 			</FindInFinder>
 			<DoSaveOrNot title="Save">
-				<Item id="1761" name="Save file &quot;$STR_REPLACE$&quot; ?"/>
+				<Item id="1761" name="Save file &quot;$STR_REPLACE$&quot; ?"/><!-- Don't translate $STR_REPLACE$ -->
```
donho, line 1390: you shouldn't modify this file.

donho (changes requested):
you shouldn't modify english.xml file.

PowerEditor/installer/nativeLang/english.xml
rammba: Ok, I will revert. Thought this change is needed.
Btw, I need to rewrite those changes on latin and PR will be ready for review.

donho: The modified xml are not validated XML files.
Please fix them.

PowerEditor/installer/nativeLang/serbian.xml
```diff
@@ -1,438 +1,1828 @@
 <?xml version="1.0" encoding="utf-8" ?>
+<!--
+	Serbian localization for Notepad++
+	Modified by:
+		- Radoš Milićev (https://github.com/rammba)
+-->
 <NotepadPlus>
-	<Native-Langue name="Srpski(Serbian)" filename="serbian.xml">
+	<Native-Langue name="Srpski(Serbian)" filename="serbian.xml"  version="8.6.9">
```
rammba, line 8: Is the version correct here, or it should be 8.7.1? I followed structure of english file.

rammba (commented):
Hello @donho, I finished translations for Serbian. Thanks for the patience and chance to improve translations for my mother tongue.
Please let me know if I made mistakes in xml structure or something similar.
Are there any other Notepad++ related content which can be translated, either on this repo or some other. I found only [wingup](https://github.com/notepad-plus-plus/wingup/tree/master)‚ and I plan to add Serbian there as well if you agree.

P.S. I found "small issue" with language change. Namely, when I change language in preferences dialog, some tips will have language which was selected on last Notepad++ load. Those tips will have same language until I restart Notepad++.

PowerEditor/installer/nativeLang/serbian.xml
MarkusBodensee: You added the latest translations for current master, so you can set version to 8.7.1 which will be the upcoming release. The version is a good indicator to see when the last update to the translations was made.

MarkusBodensee:
> P.S. I found "small issue" with language change. Namely, when I change language in preferences dialog, some tips will have language which was selected on last Notepad++ load. Those tips will have same language until I restart Notepad++.

Looks like you found a bug in current code base. Would be good if you create a bug report in the issue tracker.

donho:
@MarkusBodensee 
> Looks like you found a bug in current code base. Would be good if you create a bug report in the issue tracker.

It's a limitation surely. I don't consider it as a bug.

donho:
@rammba 
> Don't merge upstream/master (using git or via github sync), if your PR is based on older upstream/master. If you need to base it on latest master (e.g. to check and fix merge conflict), use commands git fetch upstream to get latest master and then git rebase upstream/master to rebase it onto this latest upstream/master.

ref: https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests

As a result, the PR is not merged. 
Please create a new PR with a new branch.

PR #15700, opened 2024-10-13
[xml] Update polish.xml

pawelzwronek:
Update translation texts for these commits:
- Add "Close to system tray" in MISC preference (035ef19)
- Improve GUI for commands for the system tray in Preferences (cd45afc)
- Avoid user confusion between Global override & Default Styles (1890ea6) - (in polish it's not confusing)
- Fix Find dialog status bar wrong messaging (9010147)
- Fix wrong messages for replacement status (bfe835e)


donho:

PR #15701, opened 2024-10-13
Reduce network hanging issue due to win32API GetFileAttributes cache

donho:
STR:
1. Open a network file.
2. Close Notepad++ to have it in the session.
3. Disconnect the network, and launch Notepad++ immediately.
4. Around more than 2-3 minute's hanging, then the "Error" dialog displayed.

The reason for the hanging is that the network file was incorrectly detected by doesFileExist (GetFileAttributesEx) as present, leading Notepad++ to attempt opening a non-existent file with _wfopen. This issue seems to stem from a caching mechanism within the IO function (GetFileAttributesEx). When the network disconnects, the cache is not immediately cleared, causing GetFileAttributesEx to falsely report that the file exists. Consequently, when Notepad++ is launched after a network disconnection, GetFileAttributesEx retains its cache, indicating the file exists, while _wfopen fails to locate the network resource, resulting in a hang.

Unfortunately, there's no efficient solution for this problem. The commit's remedy is to check if the file is on the network and whether its directory still exists. If the directory doesn't exist, we avoid calling _wfopen. We verify the directory's existence instead of the file itself because the cache issue with GetFileAttributesEx occurs before _wfopen is executed. Checking the directory avoids the cache problem due to the identical argument being used.

I've tested this remedy in debug mode, and it works fine. However, the problem persists in release mode. Despite this, I believe it's worth keeping this solution, as it provides some protection in a variable network environment, potentially mitigating the issue when it arises.

Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15658#issuecomment-2386662974

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -629,7 +629,7 @@ wstring BuildMenuFileName(int filenameLen, unsigned int pos, const wstring &file
 }
 
 
-wstring PathRemoveFileSpec(wstring& path)
+wstring pathRemoveFileSpec(wstring& path)
 {
     wstring::size_type lastBackslash = path.find_last_of(L'\\');
     if (lastBackslash == wstring::npos)
--- a/PowerEditor/src/MISC/Common/Common.h
+++ b/PowerEditor/src/MISC/Common/Common.h
@@ -154,7 +154,7 @@ protected:
 
 #define REBARBAND_SIZE sizeof(REBARBANDINFO)
 
-std::wstring PathRemoveFileSpec(std::wstring & path);
+std::wstring pathRemoveFileSpec(std::wstring & path);
 std::wstring pathAppend(std::wstring &strDest, const std::wstring & str2append);
 COLORREF getCtrlBgColor(HWND hWnd);
 std::wstring stringToUpper(std::wstring strToConvert);
--- a/PowerEditor/src/MISC/Common/FileInterface.cpp
+++ b/PowerEditor/src/MISC/Common/FileInterface.cpp
@@ -37,7 +37,7 @@ Win32_IO_File::Win32_IO_File(const wchar_t *fname)
 		// Store the file creation date & attributes for a possible use later...
 		if (getFileAttributesExWithTimeout(fname, &attributes_original, 0, &isTimeoutReached))
 		{
-			fileExists = (attributes_original.dwFileAttributes != INVALID_FILE_ATTRIBUTES);
+			fileExists = (attributes_original.dwFileAttributes != INVALID_FILE_ATTRIBUTES && !(attributes_original.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY));
 		}
 
 		if (fileExists)
@@ -120,7 +120,7 @@ void Win32_IO_File::close()
 
 					std::wstring curFilePath;
 					const DWORD cchPathBuf = MAX_PATH + 128;
-					WCHAR pathbuf[cchPathBuf]{};
+					wchar_t pathbuf[cchPathBuf]{};
 					// the dwFlags used below are the most error-proof and informative
 					DWORD dwRet = ::GetFinalPathNameByHandle(_hFile, pathbuf, cchPathBuf, FILE_NAME_OPENED | VOLUME_NAME_NT);
 					if ((dwRet == 0) || (dwRet >= cchPathBuf))
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -6679,7 +6679,7 @@ void Notepad_plus::notifyBufferChanged(Buffer * buffer, int mask)
 		checkDocState();
 		setTitle();
 		wstring dir(buffer->getFullPathName());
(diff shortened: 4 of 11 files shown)


donho:
@xomx 
> Got a ~6 secs delay at the N++ exit with your STR, but no N++ zombie process - good!

Do you mean "Got a ~6 secs delay at the N++ exit **AFTER RUN** your STR"?

> I got the same results both in the Debug and in the Release build.

That's a good news!
Note that after disconnection of network, you might wait for few seconds to relaunch Notepad++, in order to have network disconnected totally.
My experience with 2 sharing files laptops via WiFi is 5-8 seconds: 
* Within 5 seconds Notepad++ can often open the network file.
* Beyond 8 seconds the disconnection is taken into account and the cache of GetFileAttributesEx is cleared, so "doesFileExist" will return the correct value (false), then [File Inaccessible dialog](https://private-user-images.githubusercontent.com/90293/373347974-7938cad0-1738-448d-a3a2-104e232d7d9d.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mjg5MTgxMDYsIm5iZiI6MTcyODkxNzgwNiwicGF0aCI6Ii85MDI5My8zNzMzNDc5NzQtNzkzOGNhZDAtMTczOC00NDhkLWEzYTItMTA0ZTIzMmQ3ZDlkLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDE0VDE0NTY0NlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTExNTVlMTQ0YzNlN2IxZDQwMzgxYTc2MWM3YWM2ODNmYzRhMzEyNzdjNzIxMDA0ZmQ2YTNmMGNhMjQ0NDdjZjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.h04A_gS4zDMINbyvHQGGdmNWoU8X5Wr8Ehfn8ox612k) will show.
* Only the timing during 5-8 seconds Notepad++ hangs 2-3 minutes (PR not working), or show immediately (PR working) the [Error dialog](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15669#issuecomment-2392242230)

xomx:
@donho 

Some more testing. I used the N++ defaults (so placeholders not enabled) and Debug build (to be able to see the N++ symbol names in my SystemInformer task manager).

1st test

1. opened a net-file
2. create/switch to another N++ tab (e.g. a newxxx)
3. unplug the net-file server cable
4. switching back and forth in between the two N++ tabs until a short lag and then the N++ "Keep non existing file" dlg
5. selecting "No"
   npp-net-hangs-tests-1
6. immediately exit the N++
7. switching to my task manager to inspect the N++ exiting process - main thread is already gone but there are many `getFileAttributesExWorker` ones waiting for an end (each one takes approx 15 secs before it disappears)
  npp-net-hangs-tests-2
8. after some longer delay (~2 mins), the whole N++ process is finally gone & tidied up ok

2nd test

1.-4. as before, then answering "Yes"
  npp-net-hangs-tests-3
5. a short delay then (this dlg seems not right to me here)
   npp-net-hangs-tests-4
6. A - when answering "Yes", the file is not saved (as expected - the net-resource is not reachable...)
   npp-net-hangs-tests-5
6. B - when answering "No", the N++ tab filebuffer is closed ok
7. exiting N++, this time it takes much much more time for the complete N++ exit, here is visible why...
   npp-net-hangs-tests-6
8. after a very long delay the N++ is tidied up ok (no zombie remain...)

donho:
@xomx 
Thank you for your test.

But it seems the scenario you've tested is not what the PR tries to fix.
The hanging scenario is:

STR of hanging:

1.    Open a network file.
2.    Close Notepad++ to have it in the session.
3.    Disconnect the network, and launch Notepad++ immediately.
4.    Around more than 2-3 minute's hanging, then the "Error" dialog displayed.

The PR in Debug mode fixes (the point 4) 2-3 minutes of hanging.

donho:
@xomx 
> a short delay then (this dlg seems not right to me here)

Yes, true. I will see what I can do about it.

Though you haven't tested the problematic scenario, it doesn't matter - this PR can fix the problem in question in the best case, and it can make no difference in the worst case.

I will merge the PR now. In case of any problem detected due to the PR (though I don't think so), I can always catch it up.

xomx:
@donho 

> But it seems the scenario you've tested is not what the PR tries to fix.

I know. I tested your STR before and it was ok. I just tried some other scenarios reported by the users in their net-hanging issues.

donho:
Thank you @xomx !

PR #15703, opened 2024-10-13
fix DWM taskbar thumbnail live app-preview for modeless WS_DLGFRAME h…

### LINKED ISSUE #15690: [BUG?] Taskbar thumbnail live previews show previously closed dialogs

### Description of the Issue

When hovering over the taskbar icon then hovering over a window's thumbnail, the "live preview" will show dialogs that were previously open but have since been closed on some machines/setups.

### Steps To Reproduce

1.  Open a dialog (ex. ctrl+F for "Find")
2.  Close the dialog
3.  Hover over the Notepad++ icon in the Windows taskbar. A thumbnail image of the window appears.
4.  Hover over the window's thumbnail image. A "live preview" of the window is displayed.

### Current Behavior

On some machines/setups*, the "live preview" will display not only the Notepad++ window that's currently open, but will also display the dialog opened in step 1 (ex. "Find") as if it were still open.

### Expected Behavior

The "live preview" should reflect the current state of the window and not display any dialogs that were previously open then later closed.

### Debug Information

```shell
Notepad++ v8.7   (64-bit)
Build time : Sep 17 2024 - 17:06:31
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4894
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

**Additional details**
- When opening this issue, I just realized that this appears to happen on some machines/setups but not others. While this happens daily on my main computer, I was unable to recreate it on a different machine with a very similar setup in terms of Windows and Notepad++ settings/preferences.
- This has been happening for a long time (ex. months) and isn't something new after a recent update.
- I'm not sure if this is caused by how Windows is rendering the previews or how Notepad++ is handling opening/closing the dialogs, but since this doesn't happen with any other apps, this leads me to believe it's something about how Notepad++ is handling things.
- While all the dialogs I normally use seem to cause this to happen, there are apparently others that do not. In troubleshooting I found the "About", "Windows", and "Shortcut mapper" dialogs do *not* display in the "live preview" after being closed.
- The steps above are direct and to the point, but the live preview continues to show the old dialogs even after doing other things in Notepad++ (ex. you can do a find, go to results, close the dialog, switch around in tabs, add/remove text, and continue working for hours, and anytime you look at the "live preview", it will show the current state of the main window but also display the "find" dialog that was closed hours ago).

**Screenshots**
I opened "Find" (ctrl+F) then closed it, opened "Debug Info" then closed it, opened "Preferences" then closed it, then hovered over the thumbnail to see the live preview:
image

I clicked the thumbnail to bring Notepad++ to the foreground and this was the actual current state:
image

Note that I never had more than one dialog open at any given time leading up to these screenshots.

This can be a bit of an annoyance but in the grand scheme of thing is definitely a very minor problem. I just wanted to report it since it's been happening for quite a while now and I wasn't sure if anyone was aware of it.

xomx:
Fix #15690

The WinOS Desktop Window Manager taskbar thumbnail live preview has obviously a problem with modeless only hidden dlgs which have the WS_DLGFRAME style - the preview initially shows such dialogs, even if they were already hidden.

Temporarily using the WS_EX_TOOLWINDOW style on such a problematic dlg, right before its hiding and then after the hiding immediately removing that style, does the trick (floating toolbar-wnds are not allowed on the taskbar (or in the ALT-TAB), so DWM probably will not consider them at all for such a preview...).

### DIFF
--- a/PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
+++ b/PowerEditor/src/WinControls/StaticDialog/StaticDialog.cpp
@@ -109,6 +109,9 @@ bool StaticDialog::moveForDpiChange()
 
 void StaticDialog::display(bool toShow, bool enhancedPositioningCheckWhenShowing) const
 {
+	bool bRemoveTemporaryWndStyle = false;
+	LONG exStyle = 0;
+
 	if (toShow)
 	{
 		if (enhancedPositioningCheckWhenShowing)
@@ -152,8 +155,35 @@ void StaticDialog::display(bool toShow, bool enhancedPositioningCheckWhenShowing
 				::SendMessageW(_hSelf, DM_REPOSITION, 0, 0);
 		}
 	}
+	else
+	{
+		// toHide
+
+		// As the WinOS Desktop Window Manager taskbar thumbnail live preview has obviously a problem
+		// with modeless only hidden dlgs which have the WS_DLGFRAME style (the preview initially shows such dialogs,
+		// even if they were already hidden), we will use a simple workaround:
+		// - we temporarily set the WS_EX_TOOLWINDOW style on such a problematic dlg, right before its hiding
+		// - then do the hiding as before and immediately remove that style
+		// This does the trick (probably because of the floating toolbar-wnds are not allowed on the taskbar
+		// (or in the ALT-TAB), so they will not be previewed at all...).
+		if (::IsWindowEnabled(_hParent)) // ? is it modeless dlg (modal ones are ok for the DWM)
+		{
+			if (::GetWindowLongW(_hSelf, GWL_STYLE) & WS_DLGFRAME) // ? has it the WS_DLGFRAME (problematic for the DWM)
+			{
+				exStyle = ::GetWindowLongW(_hSelf, GWL_EXSTYLE);
+				if (exStyle && !(exStyle & WS_EX_TOOLWINDOW)) // ? has not that style already
+				{
+					if (::SetWindowLongW(_hSelf, GWL_EXSTYLE, exStyle | WS_EX_TOOLWINDOW))
+						bRemoveTemporaryWndStyle = true; // WS_EX_TOOLWINDOW temporary assigned before the hiding
+				}
+			}
+		}
+	}
 
 	Window::display(toShow);
+
+	if (bRemoveTemporaryWndStyle)
+		::SetWindowLongW(_hSelf, GWL_EXSTYLE, exStyle & ~WS_EX_TOOLWINDOW);
 }
 
 RECT StaticDialog::getViewablePositionRect(RECT testPositionRc) const


donho:
@xomx 
It's very tricky...
I'm wondering if we use directly WS_EX_TOOLWINDOW  on all dialog, we might save this tricky code.

The WS_EX_TOOLWINDOW have been replaced by  WS_EX_DLGMODALFRAME | WS_EX_WINDOWEDGE in this commit:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/5e899f2a6ce2e1adf64a693f465d6bbb7601cf46

@ozone10 
We need your expertise: I think the reason to replace WS_EX_TOOLWINDOW  by WS_EX_DLGMODALFRAME | WS_EX_WINDOWEDGE is for the look of DarkMode. If we restore WS_EX_TOOLWINDOW, will the look change to the annoying way?

ozone10:
@donho 
main issue with `WS_EX_TOOLWINDOW` is that it has no border edge which could make it hard to distinguish between dialog and editor field if user uses same color for them.

Also close button is smaller, same as undocked panels.

IMHO this is Windows bug, IIRC there is similar issue with some QT apps, also not updating live preview.

btw I cannot reproduce it on 2 machines, one with Windows 10 other 11 24H2

xomx:
@ozone10 

> IMHO this is Windows bug

I suppose so - if I restore N++ back and minimize again, the live-preview problem is gone, hidden dlgs are not visible (until I show/hide them again).

I somewhere saw a Java app having exactly this live-preview problem under Windows 10&11 and using successfully such a temporary WS_EX_TOOLWINDOW DWM-workaround, so I tried my luck in N++ and it works.

> btw I cannot reproduce it on 2 machines, one with Windows 10 other 11 24H2

I can reproduce on Win10Ent LTSC 2021 and Win11Pro 23H2.

> on which dialogs is WS_DLGFRAME used? Using spy++ I cannot find them.

I noticed that even though About-dlg is also based on the `StaticDialog` class, it does not have that problem with the live-preview when hidden. So I compared all the styles (& ex-styles) from the About and e.g. DebugInfo dlg with and found the necessary condition - `WS_DLGFRAME`.

npp-DebugInfo-WS_DLGFRAME-DWM-preview-problem

npp-About-no-WS_DLGFRAME-DWM-preview-ok

Another DWM problematic one is Find-dlg.

ozone10:
@xomx 
I see, it's part of `WS_CAPTION`

donho:
@ozone10 

> IMHO this is Windows bug

We all agree that it's Windows' bug.
It's not reasonable to fix it in the applications.
I will wait and see if Microsoft fixes their bug. In the meanwhile, let's keep this PR opened. 
If the bug is not fixed in Windows in the future, I will consider to merge the PR.

xomx:
@donho 

> I will wait and see if Microsoft fixes their bug

:-)

This is an overly optimistic expectation. Unless a miracle happens, that evidently long standing bug will remain there for good.

donho:
@xomx 
> This is an overly optimistic expectation. Unless a miracle happens, that evidently long standing bug will remain there for good.

I see your point, but I still want to give Microsoft a chance to fix their bug. ;)
If the bug persists before Notepad++ release v8.7.2, then the PR will be merged into v8.7.2.

xomx:
@donho 

> but I still want to give Microsoft a chance to fix their bug. ;)

Ok, I will try not to behave like a spoilsport to your optimism. :-)

Let's pretend the MS does not know about this problem yet. Then a simple POC & video for the report is needed:

[PreviewBugPOC.zip](https://github.com/user-attachments/files/17412665/PreviewBugPOC.zip)

Win10 video (Win11 shows the same, but Win7SP1 in a VM is ok, so it seems to be a Win10+ bug, I do not have Win8 ready for such a test):

https://github.com/user-attachments/assets/f807f241-dc4e-463b-b4bc-b5e5b2a74b64

Now the hard part - reporting it to them:
- since Win10+ you can use the MS Feedback app but this is an one-way blackhole for reports, you have to report & pray:
  https://support.microsoft.com/en-us/windows/send-feedback-to-microsoft-with-the-feedback-hub-app-f59187f8-8739-22d6-ba93-f66612949332
- then there are some MS forums to try, but I have the impression that these are not operated by the MS staff itself (outsourcing, volunteers?), so you cannot rely on it to much:
  https://answers.microsoft.com
  https://techcommunity.microsoft.com  
- best possibility is still to know a suitable MS insider

If anyone wants to try any of this, feel free to do it (as I will not, good luck).

There is one more "dirty" possibility - somehow pretend/try to pass it off as a security issue and then write a blog post about it and try to make it go viral somehow. It is stupid but it works, nowadays people hear about it and the MS pays attention to its possible negative audience. (How to do it here - I would e.g. write something like: "It is a potential security issue because there could be a deliberately hidden dialog containing sensitive information... blablabla".

Endnote: Do not ever try the MS call support for such a thing - real purpose of this 100% outsourced stuff is to get rid of thousands people annoying the MS every day with often trivial or non-existent problems. It is a filter-wall. For a person with a real technical problem to report, it is a multi-level circular Hell where the only goal is to get the caller to hang up.

donho:
@xomx 
> Now the hard part - reporting it to them:

Very true :(

Feedback Hub has been downloaded, and bug was reported (with the POC zip you provided) via Feedback Hub to Microsoft.
And the same bug report was submitted on the both sites:

https://answers.microsoft.com/en-us/windows/forum/windows_11-start-win_taskbar/bug-taskbar-thumbnail-live-previews-show/c36fa139-bc60-48e2-9821-04ec433fc2da

https://techcommunity.microsoft.com/discussions/windows11/bug-windows-11-taskbar-thumbnail-live-previews-show-previously-closed-dialogs/4297001

Thank you very much for the very detail instructions and the POC file.

donho:
The answer from Microsoft:
https://techcommunity.microsoft.com/discussions/windows11/bug-windows-11-taskbar-thumbnail-live-previews-show-previously-closed-dialogs/4297001/replies/4443869

xomx:
> The answer from Microsoft:
> https://techcommunity.microsoft.com/discussions/windows11/bug-windows-11-taskbar-thumbnail-live-previews-show-previously-closed-dialogs/4297001/replies/4443869

Unfortunately this is not a sln, now I tested on another PC with Win11 25H2 (build 26200.8037), with GPU drivers from late 2025 and the problem is still there. It's a MS bug, not a GPU one.

ozone10:
That bug is also affecting Windows 10. And since Windows 10 is practically EOL and can run on many old PCs, which don't have new GPU driver I don't think that solution is correct.

donho:
@xomx @ozone10 
Thank you. I just replied with the arguments you provided, let's see what MS will say.

PR #15706, opened 2024-10-15
Avoid user confusion between Global override & Default Styles

### LINKED ISSUE #15640: [Feature request] Please de-emphasize Global Override in the style list

### Description of the Issue

There is a long history of people not understanding the right way to change the normal text font, which is inherited by all the languages.

Because the Style Configurator places **Global override** before **Default Style**, then people assume that's the "right" place to change the font.  But really, the **Default Style** should be the primary location for making that change (because people don't always understand what the true consequences of **Global override** really are).

image

Examples of Global Override being used when Default Style would have been more appropriate
- https://community.notepad-plus-plus.org/topic/26132/changing-the-font/
- https://community.notepad-plus-plus.org/topic/25701/font-size-on-popup-window-please
- https://community.notepad-plus-plus.org/topic/25224/failed-to-display-text-using-simultaneously-combined-chinese-fonts-characters-correctley/4
- https://community.notepad-plus-plus.org/topic/24571/change-tab-font-size-and-document-font-color/3
- https://community.notepad-plus-plus.org/topic/23768/default-font-for-utf-8
- https://community.notepad-plus-plus.org/topic/23590/good-old-highlighting-issue-topic/5
- https://community.notepad-plus-plus.org/topic/23496/background-colors-since-8-4-5-upgrade-of-notepad
- https://community.notepad-plus-plus.org/topic/23420/japanese-font-not-displaying-in-notepad
- https://community.notepad-plus-plus.org/topic/22443/persian-farsi-font-has-different-style-in-np-and-ms-np
- https://community.notepad-plus-plus.org/topic/21451/styles-keyword-highlighting-stopped-working/3
- https://community.notepad-plus-plus.org/topic/21009/underline-a-single-word/3
- https://community.notepad-plus-plus.org/topic/20128/help-selected-lines-are-blacked-out-how-to-fix
- https://github.com/tonsky/FiraCode/wiki/Notepad-Plus-Plus-instructions

These are just examples going back to 2020... I didn't dig any further back; and I didn't include ones in the list where I diagnosed the problem as someone had set the Global Override instead of Default Style, but they never came back and replied to let us know whether it was the cause of their problem; there were about as many of those as the list above.  The last was an external link I happened to run across while researching something else, where a font developer gave their "official" recommendation as to how to use their font in Notepad++, thus showing that the misunderstanding is pervasive across the internet, not just people having problems once they do set a Global Override and come asking for help.

There are at least ways that I can think of to improve this: implementing one or both of these would be quite helpful:
1. Deprioritize **Global override** in the **Global Styles**'s list of style choices -- put **Default Style** first, and either put **Global Override** second, or (even better) put it at the very end of the list, so that users have to go looking for it.
2. Make it obvious in the **Global override** panel that overriding is an "option of last resort" and that most of the time, what a user really wants is to set the **Default Style**.
    - Suggested statement: `Enabling global override here will override that parameter in all language styles.  What you probably really want is to use the "Default Style" settings instead`
        - maybe put that text as a label with a description -- though it takes up a lot of space
            image
        - maybe put one of your [?] buttons in that dialog, and have that pop up the text instead
    - include a link on the `"Default Style"` text to the **Default Style** page

### Debug Information

```shell
n/a
```

### Anything else?

_No response_

donho:
Fix #15640


donho:
@pryrt 
I've just refined the PR for the last time. If you see somethings could be improved, feel free to let me know.

donho:
I will merge the PR now. In case of any problem detected due to the PR (though I don't think so), I can always catch it up.

PR #15710, opened 2024-10-16
[XML] Update brazilian_portuguese.xml

Marcellomco:
Added translation for "Fix Find dialog status bar wrong messaging".


Marcellomco (commented):
Also added translation for "Avoid user confusion between Global override & Default Styles".

donho:

Marcellomco (commented):
Added correction for old updates not implemented. Downloaded and fully compared to English version.

PR #15719, opened 2024-10-20
Installer - always allow to use the "Show details" functionality

### LINKED ISSUE #15718: Installer - Ensuring that one can always use the "Show details" button to review installer actions.

### Description of the Issue

1) There can be a problem with successfully clicking the "Show details" button during the installing-phase (sometimes it takes several clicks to finally show the details).

   npp-installer-ShowDetailsBtn-issue

2) These details cannot be left displayed for some possible inspection after the installer has finished its ops, as the installer immediately switches to its last page “Completing Notepad++ vXXX Setup with Finish-btn” after completing its job.

   npp-installer-ShowDetailsBtn-FinishPage

### Steps To Reproduce

1.  Run the Notepad++ installer.
2.  Try to click the "Show details" button.
3.  If ok - repeat 1&2 until the described problem occurs.  

### Current Behavior

Only sometimes user can see the details during the installation phase, but never at the installer actions end.

### Expected Behavior

User can review the installation details either during the installation phase or at least at the installers steps end.

### Debug Information

```shell
Tested with x64 & x86 v8.7 installer.
```

### Anything else?

https://community.notepad-plus-plus.org/post/96738

xomx:
Fix #15718 

By using the NSIS MUI_FINISHPAGE_NOAUTOCLOSE, the installer will always wait for the user Next-click confirmation before it goes to the last Finish-page.

Otherwise, sometimes even during the installation phase, the details of the installation activities cannot be displayed (probably some problem with the focus of the NSIS installer prevents the "Show details" button from being successfully clicked).

npp-installer-ShowDetailsBtn-fix

### DIFF
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -283,6 +283,11 @@ FunctionEnd
 
 
 Section -"Notepad++" mainSection
+	${If} $showDetailsChecked == ${BST_CHECKED}
+		SetDetailsView show
+		SetAutoClose false
+	${endIf}
+
 	${If} $diffArchDir2Remove != ""
 		!insertmacro uninstallRegKey
 		!insertmacro uninstallDir $diffArchDir2Remove 
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -85,7 +85,7 @@ page Custom ExtraOptions
 !define MUI_PAGE_CUSTOMFUNCTION_SHOW "CheckIfRunning"
 !insertmacro MUI_PAGE_INSTFILES
 
-!define MUI_FINISHPAGE_NOAUTOCLOSE
+
 !define MUI_FINISHPAGE_RUN
 !define MUI_FINISHPAGE_RUN_FUNCTION "LaunchNpp"
 !insertmacro MUI_PAGE_FINISH
--- a/PowerEditor/installer/nppSetup.nsi
+++ b/PowerEditor/installer/nppSetup.nsi
@@ -85,7 +85,7 @@ page Custom ExtraOptions
 !define MUI_PAGE_CUSTOMFUNCTION_SHOW "CheckIfRunning"
 !insertmacro MUI_PAGE_INSTFILES
 
-
+!define MUI_FINISHPAGE_NOAUTOCLOSE
 !define MUI_FINISHPAGE_RUN
 !define MUI_FINISHPAGE_RUN_FUNCTION "LaunchNpp"
 !insertmacro MUI_PAGE_FINISH
--- a/PowerEditor/installer/nsisInclude/tools.nsh
+++ b/PowerEditor/installer/nsisInclude/tools.nsh
@@ -72,9 +72,11 @@ FunctionEnd
 Var Dialog
 Var NoUserDataCheckboxHandle
 Var ShortcutCheckboxHandle
+Var ShowDetailsCheckboxHandle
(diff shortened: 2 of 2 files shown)


donho:
@donho said in [Notepad\+\+ release 8\.7](/post/96784):
> @xomx said in [Notepad\+\+ release 8\.7](/post/96738):
> > This can be simply changed (MUI_FINISHPAGE_NOAUTOCLOSE ) in the installer source but it will add one more Next-btn click to the whole installation. What do you all think about it, is it worth doing so?
> 
> If there's not a lot of code change, and no regression risk of course, why not? Otherwise, I would say no.

(ref: https://community.notepad-plus-plus.org/topic/26122/notepad-release-8-7/9?_=1729691844236)

@xomx 
Just tried this PR [Installer - always allow to use the "Show details" functionality](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15719), and realized this will add one more step in the installation which is complain already by a lot of users for its long and boring procedure.

I'll see if it's possible to make it optional (if you have any idea for it, please let me know).

It was my bad: your description above was clear for the consequence. Sorry about it.

xomx:
@donho 

> I'll see if it's possible to make it optional (if you have any idea for it, please let me know).

Now I do not have time to check for such a possibility but IIRC there is no easy (NSIS built-in) way.
And I am not for creating a new cmdline-switch only for this minor thing.

> It was my bad: your description above was clear for the consequence. Sorry about it.

Ok, never mind. We could postpone this PR from the upcoming release, but I think that now users will rather just click on that new "Yes (silent)" option after the initial N++ installation - see e.g. [here](https://community.notepad-plus-plus.org/topic/26188/big-thanks-for-new-silent-update-feature) how much they appreciate it ;-)

So maybe one extra Next-click in the initial N++ installation might not be so much of a hassle here after all...

donho:
> Now I do not have time to check for such a possibility but IIRC there is no easy (NSIS built-in) way.

Hmm... I was playing around with the following code:
```nsis
!if $showDetailsChecked == ${BST_CHECKED}
!insertmacro MUI_FINISHPAGE_NOAUTOCLOSE
!endif
```
and 
```nsis
Function OnChange_ShowDetailsCheckbox
	${NSD_GetState} $ShowDetailsCheckboxHandle $showDetailsChecked
FunctionEnd

Function ExtraOptions
	nsDialogs::Create 1018
	Pop $Dialog

	${If} $Dialog == error
		Abort
	${EndIf}

	${NSD_CreateCheckbox} 0 0 100% 30u "Create Shortcut on Desktop"
	Pop $ShortcutCheckboxHandle
	StrCmp $WinVer "8" 0 +2
	${NSD_Check} $ShortcutCheckboxHandle
	${NSD_OnClick} $ShortcutCheckboxHandle OnChange_ShortcutCheckBox


	${NSD_CreateCheckbox} 0 60 100% 30u "Show installation details"
	Pop $ShowDetailsCheckboxHandle
	${NSD_Check} $ShowDetailsCheckboxHandle
	${NSD_OnClick} $ShowDetailsCheckboxHandle OnChange_ShowDetailsCheckbox

```

It doesn't work.
It seems  that NSIS page can only be done statically.

This PR will be stand by. If we find a way to insert it dynamically, then it's good. Otherwise it doesn't worth IMO.
Sorry again for not merging the PR.

whindsaks:
SetAutoClose can be called in a Section to control if it moves to the next page or not.

xomx:
@donho 
> If we find a way to insert it dynamically, then it's good.

Now it works this way, try new checkbox:

npp-installer-show-details

PR #15750, opened 2024-11-01
Add Pin Tab feature

### LINKED ISSUE #5786: Feature request: pin tab
I have a problem with tabs in NPP - I usually have too many open!  Some files I have open all the time.  It would be great if NPP had a pin tab feature.  This would work the same way as the major browsers - the pinned tab would shrink in size and always be visible, to the left of all other tabs.

donho:
Fix #5786


xomx:
npp-pin-PR-problem2

npp-pin-PR-problem3

donho:
@xomx 
Thank you for your review.

Though the part of non-GUI functionality has been done, I shouldn't make PR as "review ready". I wanted to modify/simplify the icon state and indeed there are still some visual glichy.
Regarding Dark mode, once I finish to modelized the lite mode icon state and get the glichy fixed, I will request @WildByDesign for the Dark mode icons, if he is willing to help.

I'll ping you when it's all done.

donho:
@xomx 
FYI, the issues are fixed.

@WildByDesign 
Currently I use the icons of lite mode in dark mode:

image

Surely the result is far from perfect:

image


Could you provide the icons for dark mode or please tell me where I can find dark mode icons.
Thank you!

schnurlos:
https://icons8.com/icons/set/pin--black as an example

WildByDesign:
I'm working on some Fluent icons for dark mode right now for the pin tab feature. I'll let you know when it's done.

The only concern that stands out to me so far is that the filename in the tabbar is very close to the pin tab icon. I think that there should be a 1 or 2 pixel space between filename and pin tab icon.

WildByDesign:
@donho Here are the dark mode Fluent icons for the new Pin to tab feature.

[pinTab-v1.zip](https://github.com/user-attachments/files/17679143/pinTab-v1.zip)

Most of my testing has been at 125% scaling and I am happy with the results. Please test at 100% , 150% and 200% and let me know if there are any issues. Thanks.

donho:
@WildByDesign 
> The only concern that stands out to me so far is that the filename in the tabbar is very close to the pin tab icon. 

It's due to the icon size value "16" instead of "11" - the latest commit fixes it.

donho:
@WildByDesign 

Icon size value "11":

image

donho:
Icon size value "16"

image

WildByDesign:
I just tested the latest commit and it seems to have shrunk the icons, assuming to base size of 11.

I am pretty sure that we had done all of the other tabbar-specific icons at a base size of 16 specifically for dark mode.

Also, I am pretty sure that the light mode tabbar-specific icons remained at a base size of 11.

So I think that you were right in the first place and that most recent commit (https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15750/commits/40c68ce0ed78bd5fbc5d956e3ff5529c2710728c) was not needed. I think that commit should be reverted.

But that does leave us with the issue of space needed between filename and pin tab icon.

If you want, I can create that 1 or 2 pixel of space with the icons by simply aligning the icon more to the right. That way you wont have to create a space within the code.

donho:
@WildByDesign 
> Here are the dark mode Fluent icons for the new Pin to tab feature.

Thank you for the icons!
The 2 above screenshots are done on 2 laptops - the 1st with Icon size value "11" & the 2nd with "16".
And we can see that the 2nd is cristal clear, whereas the 1st one is a little bit blurred comparing the 1st one.

My question is: could we keep icon size "11", and with the sharp icon look (another set provided by you) ?

If it's not possible, I will find another solution.

donho:
@WildByDesign 
> But that does leave us with the issue of space needed between filename and pin tab icon.

OK. I will try to improve it. Let's keep the current icon. I'll revert the latest commit.

donho:
@WildByDesign 
It looks quite good the latest commit to me.
What do you think?

WildByDesign:
@donho That looks absolutely perfect. I love it!

xomx:
@donho 
There is a different pinning behavior in the light vs dark mode.

LMC ... left mouse click

Light:
- move mouse on the tab pinning-ico area (hovering only) == pinning state visually changes (pin->pinned or pinned->pin)
- A) without LMC, after moving the mouse out of that ico area, the pinning-ico state is reverted back to its pre-hovering one
- B) with LMC it does not change from the hovering visual look at all, but after moving the mouse out of that ico area, the changed hovering pinning ico state remains (not changes back to its pre-hovering state)

Dark:
- move mouse on the tab pinning-ico area (hovering only) == pinning-ico only reflects the mouse hovering visually (pin/pinned state does not change)
- A) without LMC, after moving the mouse out of that ico area, the pinning-ico look is reverted back to its non-hovering representation
- B) with LMC it does not immediately change visually from the hovering state, but after moving the mouse out of that ico area, the pinning state changes (pin->pinned or pinned->pin)

xomx:
With the latest commits I have this incorrect pinning-ico position in darkmode (lightmode is ok, Win10 1920x1080 100%):

npp-pinning-darkmode-position

donho:
@xomx 
> There is a different pinning behavior in the light vs dark mode.

True. The latest commit fixes it.

@WildByDesign 
FYI: In order to align the dark mode hover behaviour with lite mode one, I've modified the 2 hover icons.

donho:
@xomx 
> With the latest commits I have this incorrect pinning-ico position in darkmode (lightmode is ok, Win10 1920x1080 100%):

Please check in the latest commit: It should be OK now.

WildByDesign:
I just checked the latest commit and everything looks good to me and functions very well.

xomx:
@donho 
Perfect!

WildByDesign:
> If tab is pinned, pin button will replace close button. Clicking on pin button will unpin tab and close button will appear again.
> This will also reduce size of tabs, as there will be no needs for unpinned tab indicator.

I agree, this would be best. This is how VS Code behaves which helps to prevent accidentally closing the pinned tabs.

donho:
@ozone10 
> I would like to suggest to create menu item to pin tab in right click context menu of tab.
If tab is pinned, pin button will replace close button. Clicking on pin button will unpin tab and close button will appear again.
This will also reduce size of tabs, as there will be no needs for unpinned tab indicator.

It's a smart way to achieve this feature for some users, but for the other users it's the fastidious way for pinning the tab (right clicked & "Pin" command).
I prefer the current (the PR) way - it is clear and intuitive. If users don't like them (2 buttons), they can hide one of them, or all of them.

OTOH, I like the behaviour of VS Studio - hide them when they are not needed. I might implement them in the future PR.


 
The tab context menu "Pin/Unpin" command is planed to be implemented in the next PR.

PR #15753, opened 2024-11-02
[xml] Update Ukrainian translation

artpoli:


donho (changes requested):
New added items for translations are not present. Please check the updated english.xml

Issue #6021, opened 2019-08-08
focus moves when switching back to n++ in find/replace


glebsts:
Hello, thank you for awesome tool.
Did not find similar issue here, so reporting.
v7.7.1
1. Open find/replace dialog, put smth in "find", move to "replace", put smth there
2. Switch to some other window with Alt-Tab (that's important)
3. Switch back to N++ with Alt-Tab (that's important)

### Expected Behavior
Focus should stay in "Replace" field, I can continue typing

### Actual Behavior
Focus moves back to "Find" field, and field text is selected (typing will replace entered text)

### Debug Information
Win10 1809 17763.557
N++ with administrator privileges.

Ekopalypse:
Strange, I don't see your reported behavior. 
If I switch from npp to another one and back to npp the caret is in the current document
but not within the replace dialog.

glebsts:
"Caret is in the current document" happens if I switch i.e. with mouse using taskbar window buttons. Reported behavior with focus moving happens on Alt+Tab switching.

sasumner:
Verified OP's reported behavior with the following:

Notepad++ v7.7.1   (64-bit)
Build time : Jun 16 2019 - 21:24:47
Path : C:\........\npp.7.7.1.bin.x64\notepad++.exe
Admin mode : OFF
Local Conf mode : ON
OS : Windows 10 (64-bit)
Plugins : mimeTools.dll NppConverter.dll

Ekopalypse:
I'm using alt+tab, tested on 7.6.6 and 7.7.1 both 64bit BUT on windows 7.
Opening the replace dialog, alt+tab to browser, don't do anything within browser but immediately do alt+tab to return to npp. Focus is on document.

sasumner:
OP did not specify OS; if they would have followed instructions about creating an issue (fully), the actual `Debug Info` output would be there and we would know their OS...

glebsts:
sorry, will add more info to the original post. Win10 1809 17763.557

Ekopalypse:
Fired up win10 in VirtualBox and yes, I can see it there which raises the question if a 
possible future change affects different windows version differently?

sasumner:
@Ekopalypse A solution should be derived that works on Win7 and Win10.  Or do I miss your point?

glebsts:
Thank you for confirmation.
[...]

PR #15756, opened 2024-11-03
Don't force focus to Find field when returning to Find/Replace dialog

Coises:
Addresses Issue #6021


PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1708,7 +1708,7 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 				enableFindDlgItem(IDREDOTMATCHNL, false);
 			}
 			enableProjectCheckmarks();
-			return TRUE;
+			return FALSE;
```
donho, line 1711: Though the value of FALSE is also 0, I think it'll be better use 0 instead, for the semantic reason.
As well, in the online document: `If an application processes this message, it should return zero.`
https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-activate

donho (changes requested):
Use 0 instead of FALSE.

donho (approved):

PR #15781, opened 2024-11-10
Hide inactive tab buttons (Close & Pin)

### LINKED ISSUE #15298: [Feature request] Hide Close button on inactive tabs

donho:
Note that the implementation for Pin button will be in the next commit. 
Fix  #15298


xomx:
@donho 
Very nice!
Currently only the close button is hidden on inactive tabs (so not like MSVS where both buttons are hidden).

WildByDesign:
@donho  I absolutely love this. This is exactly how I envisioned it and you've implemented it perfectly. Great work!

alankilborn:
@donho 

Start with both the "close button" and "pin tab" checkboxes unchecked:

image

Checkmark the "close button" checkbox:

image

Finally, checkmark the "pin tab" checkbox:

image

Note the overlap between the tab name and the "pin".

donho:
@WildByDesign 
> If you have a moment, please drop those in `src\icons\dark\tabbar` in your own build.
> 
> Please let me know what you think. If you like them, please feel free to include them in your commit.

The old icons are more fancy, and the new ones are simpler and sober. I like both new & old ones.
But if you suggest them, I guess you have your reason from the designer viewpoint. They are included and committed.

xomx:
@donho 
> > Note the overlap between the tab name and the "pin".
> 
> Yes, I noticed this bug, but I didn't find the reason of causing the bug - on the debug mode the change has been considered, but for the reason I ignore, the tab doesn't shrink. Could you please create an issue for it? Maybe some people who have more Win32 competence than I have can fix the bug.

Not so great fix but visually ok:

1) Add new public method to the .\PowerEditor\src\WinControls\TabBar\TabBar.h:
```
	void refresh() {
		int index = insertAtEnd(L"");
		deletItemAt(index);
	}
```
2) Add the following two calls to the .\PowerEditor\src\NppBigSwitch.cpp at very end of the `case NPPM_INTERNAL_DRAWTABBARCLOSEBUTTON:` and `case NPPM_INTERNAL_DRAWTABBARPINBUTTON:` message handlers:
```
			_mainDocTab.refresh();
			_subDocTab.refresh();
```

(during testing of your PR I just noticed that adding or deleting a tab fixes the situation, that's all...)

PR #15794, opened 2024-11-13
Highlight inactive tab on mouse hover

### LINKED ISSUE #15791: [Feature request] Highlight inactive tab (background color) on hover

### Description of the Issue

This is a follow-up feature request to the recent hiding of the Close tab button on inactive tab (https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15298).

When hovering over an inactive tab, the hidden Close button appears which is great. Another good usability experience to add to this would be adding a highlight to the tab background color as well.

Hovering over an inactive tab should change the background color of that tab as a visual indication.

I believe that the background color of that inactive (now hovering) tab should be taken from the same color as the active tab. That background color, however, should be slightly reduced from the active tab background color.

### Debug Information

```shell
Notepad++ v8.7.1   (64-bit)
Build time : Nov 11 2024 - 16:35:42
Path : D:\Tools\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 24H2
OS Build : 26100.2033
Current ANSI codepage : 1252
Plugins : 
    mimeTools (2.9)
    NppConverter (4.5)
    NppExport (0.4)
    RainLexer (2.21)
```

### Anything else?

_No response_

donho:
Fix #15791


donho:
@WildByDesign 
If you're not sastisfied with the result, you can play with the values lightnessMore (25) & saturationLess (0) - both can be negative:
`brushColour = hls.toRGB4DarkModWithTuning(25, 0);`
And let me know the values you prefer.

donho:
> Great work! I amazed that something like that can be done with such a small amount of code.

That's because there are already a lot of tools in this mature project, thanks to nombrous contributors :) 

> Light mode with Darken inactive tabs unchecked is a problem. This would display the inactive tabs as black when you hover the mouse over, no matter what. I could not fix this.
So I don't know how to proceed with Light mode.

This feature is excluded when "Darken inactive tab" option is OFF. It makes sense, and since the option is ON by default, the most of users will benefit from this new feature.


The new close icon is included in the latest commit. 
Thank you for the new icon and for your time for the tunning!

WildByDesign:
> This feature is excluded when "Darken inactive tab" option is OFF, for both lite & dark mode. It makes sense, and since the option is ON by default, the most of users will benefit from this new feature.
> 
> The new close icon is included in the latest commit. Thank you for the new icon and for your time for the tunning!

You're welcome. And thank you for getting this feature request done so fast.

This turned out even better than I had originally envisioned. The implementation is quite literally perfect.

The fact that it will be default for the majority of users makes me happy.

PR #15811, opened 2024-11-16
Fix empty.ico - removed random semi-transparent pixel

pawelzwronek:
empty.ico seems to has some lost semi-transparent pixel, causing to render black pixel when Close Button is hidden:
obraz

Fixed by clearing whole canvas in GIMP and exporting to .ico as 1 bpp, 1-bit alpha, 2-slot palette.

obraz


donho:
@pawelzwronek 
I don't have such issue:

image

chcg:
grafik
Directly looking at the icon shows that there is a block dot in it. It is also not visible in my test with N++ 8.7.1 .

@pawelzwronek Could you please add the N++ debug info of the environment you use.

pawelzwronek:
Sure:
```Notepad++ v8.7.1   (32-bit)
Build time : Nov 16 2024 - 19:31:40
Path : D:\git\notepad-plus-plus\PowerEditor\visual.net\Debug\Notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4780
Current ANSI codepage : 1250
Plugins : none
```

PR #15816, opened 2024-11-17
Introduced styles check utility

jdknight:
Adding a utility script to compare the stylers model to themes. This script can be used to detect missing lexer types as well as word styles for said lexer types.

----

The goal of introducing this script is to see if it could help support the maintenance of themes. My preferenced theme is `Twilight` and noticed that it was not styling `Rust` code blocks. With a quick inspection, it was noticed that some stock themes had more lexer types defined than others. While I could update the theme directly to match, I figured I might be able to push some theme corrections upstream to prevent having to update my local theme over time. Figured I would help a helper script to see what is missing for a given theme to help bring a theme into align with the stylers model.

If such a script is not desired, no worries.

----

Example output:

```
$ ./StylesCheck.py DarkModeDefault
Loading styles model: C:\gitwork\notepad-plus-plus\PowerEditor\src\stylers.model.xml
Styles model defines 89 lexers types.

DarkModeDefault
  MISSING: Nim (nim)
  UNKNOWN: Nimrod (nimrod)

Detected an issue (total: 2)!
```

### DIFF
--- a/PowerEditor/scripts/StylesCheck.py
+++ b/PowerEditor/scripts/StylesCheck.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Introduces a styles check utility. This can be used to help check for
+# possible updates needed to apply for a given theme.
+
+from enum import Enum
+from enum import auto
+from pathlib import Path
+import xml.etree.ElementTree as ET
+import argparse
+
+# version of this script
+__version__ = '0.0.0'
+
+# default filename for the stylers model
+DEFAULT_STYLERS_MODEL = 'stylers.model.xml'
+
+# expected tag at the root of the xml model/style document
+XML_ROOT_TAG = 'NotepadPlus'
+
+# xml query for finding lexer types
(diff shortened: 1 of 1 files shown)


alankilborn:
Perhaps has some overlap/commonality with [THIS](https://community.notepad-plus-plus.org/topic/26049/config-files-need-updating-too), although to be honest I haven't looked very deeply at the functionality of either one (I've blindy used `ConfigUpdater.py` though).

jdknight:
> Perhaps has some overlap/commonality with [THIS](https://community.notepad-plus-plus.org/topic/26049/config-files-need-updating-too)...

For sure. I was not aware this script existed, so thanks for mentioning it.

From my inspection of what [ConfigUpdater.py](https://github.com/pryrt/nppStuff/blob/771d873fc8d0da0a6ba757b0ba3222551c1a53f3/pythonScripts/useful/ConfigUpdater.py) provides, it looks to be a great tool for managing updates to style documents, pulling in new lexer types into themes without overriding existing modifications. The focus towards both style creators/maintainers, as well as some end users.

The script being provided in this pull request is a bit different, where the focus is more so on simple linting (i.e. mainly for maintainers wanting to do sanity checks).

Issue #15776, opened 2024-11-09
[BUG] Opening files from Smb share fails


Absolutely-Free:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

I have a file share from a Samba server on my home network that I use to store Windows related notes in txt files. In previous versions of notepad++, I was able to open those files directly from the shared directory and edit them there. This functionality has broken in a recent version. I am not experiencing any problems with the SMB server. I can still open these files directly from the SMB server with regular notepad just fine, and if I copy the files to a local directory I can edit the copy with notepad++, but I always get errors when I try to open the files directly from the shared path with notepad++. For example:

![image](https://github.com/user-attachments/assets/59052e41-fd5d-4de4-8916-f22fa8cb9290)

This is the path to the files. I double click on one of the files, and I get this error when opening it:

![image](https://github.com/user-attachments/assets/b1e53d13-3a25-4658-bd57-6783edc64649)

Followed by this message when I click OK:

![image](https://github.com/user-attachments/assets/9b37d7db-4e37-446f-9463-fd6db7243c93)

### Steps To Reproduce

1.  In file explorer, navigate to an SMB shared directory containing a txt file.
2.  Open the txt file in notepad++

### Current Behavior

Currently I get error messages as shown above

### Expected Behavior

The files should be able to be opened with notepad++ and edited as usual.

### Debug Information

```shell
Notepad++ v8.7.1   (64-bit)
Build time : Oct 31 2024 - 00:48:56
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 10 Pro N (64-bit)
OS Version : 22H2
OS Build : 19045.5011
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

I tried renaming the plugins folder and then opening the files from the SMB share, but it had no effect.

xomx:
@Absolutely-Free 

Could you download the fix draft x64 binary 

https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/11761446995/artifacts/2167545451

replace your "notepad++.exe" file and test your scenario again?

Absolutely-Free:
[...]

PR #15820, opened 2024-11-18
Code enhancement for file checking

xomx:
This enhancement reports the real reason for a possible file size check fail and does not allow the code proceed further (to the Scintilla `SC_STATUS_FAILURE`) as it was seen in the issue #15776 .

It also added all the missing `dwFileAttributes = INVALID_FILE_ATTRIBUTES` inits for surety.

And changes one misleading `DWORD` to `BOOL`.
- this was probably caused by mixing together the worker thread's DWORD exit code with the worker thread's WINAPI functionality purpose (`GetFileAttributesEx`, `GetDiskFreeSpaceExW`) which returns BOOLs (and this is what we need to work with here)


xomx:
@donho 

While completing the inits of `dwFileAttributes`, I ran into place where the success of the GetFileAttributesEx WINAPI is not checked at all:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/af2b90f0a586028afb3f26159cfaa4eaf250ed84/PowerEditor/src/WinControls/ReadDirectoryChanges/ReadFileChanges.cpp#L9

I did only the missing init there and some formatting but nothing else. Do not know if it really needs a error checking there or not, just FYI.

PR #15829, opened 2024-11-20
Updated Norwegian translation 8.6.9

Njardarheim-1337:
Updated the Norwegian translation to 8.6.9


donho (changes requested):
New added items for translations are not present. Please check the updated english.xml

PR #15833, opened 2024-11-21
Fix French translations

w-ap-admin:
Some outdated translations are no longer commonly used in everyday language and may not be easily understood by younger users.


nzhiti:
The actual translations of those buttons are more a rough language than fancy. Thank you

EtienneTeuf:
You are right w-ap-admin, thx for noticing it!

Issue #9513, opened 2021-02-15
REGRESSION: DPI scaling broken in v7.9.3


Whissi:
### Description of the Issue
I am using multiple monitors with different DPI scaling settings, like 100% and 120%.

It was never a problem to open Notepad++ on the monitor with the 100% setting and move it to the 120% screen -- once the program window was shown on the 120% screen, UI scaled up as expected. When I moved the UI back to the other other monitor, UI scaled back down to 100%.

This doesn't work anymore with v7.9.3. I'll get 120% scaling (which is coming from my main monitor) on all screens.
Downgraded back to v7.9.2 fixes the problem.

### Steps to Reproduce the Issue
1. Get multiple monitors and apply different DPI scaling.
2. Open Notepad++ on first monitor, check that everything looks like you are expecting and now move UI to a monitor with different DPI scaling.

### Expected Behavior
UI should adapt to different scaling like in <=7.9.2.

### Actual Behavior
UI always renders with highest DPI scaling (or with DPI settings from main monitor) which makes it looking too big on monitors where no DPI scaling should apply on.

### Debug Information
```
Notepad++ v7.9.2   (64-bit)
Build time : Dec 31 2020 - 04:01:34
Path : C:\Program Files\Notepad++\notepad++.exe
Admin mode : OFF
Local Conf mode : OFF
OS Name : Windows 10 Pro for Workstations (64-bit) 
OS Version : 2009
OS Build : 19042.804
Current ANSI codepage : 1252
Plugins : mimeTools.dll NppConverter.dll NppExport.dll 
```

sasumner:
You need to provide the _Debug Information_ that was required in the issue template.

Whissi:
I added debug information. I expect that for v7.9.3 only version number has changed since I already downgraded.

sasumner:
So for me in 7.9.3, display is crisp and great on all four of my various screens, in contrast to 7.9.2 where it was blurry on some.  This is always using the "Recommended" _Scale and layout_ settings for all monitors; which causes a mix of 125% and 100% settings, examples:

![image](https://user-images.githubusercontent.com/30118311/107955209-847e8500-6f6b-11eb-8964-a7a52e0dbb81.png)

![image](https://user-images.githubusercontent.com/30118311/107955237-8ea08380-6f6b-11eb-87ad-be22b04359c2.png)

[...]

Issue #8660, opened 2020-08-03
Maximised Notepad++ has 1px margin next to "X"-button


Fonze:
### Description of the Issue

At least with multiple monitors maximized Notepad++ has like 1px white margin on the right side of its "X"-button and if you click it, the click event goes through the Notepad++ window and will be registered on whatever is behind it (so you might accidentally close another app). This behavior is very annoying as I (and I guess many other users too) just push my cursor to the top right corner of the monitor and click without watching what I'm doing.

Happened on Windows 10 and Notepad++ v7.8.7 64-bit

### Steps to Reproduce the Issue

(if this doesn't work with one monitor, try having multiple monitors)
1.  Open any app with "X"-button on the top right corner and maximize it.
2.  Open Notepad++ and maximize it
3. Push your cursor to the top right corner of the screen and click.
4. Notice how you closed the window below Notepad++.

kspalaiologos:
![picture](https://user-images.githubusercontent.com/27734421/89405456-e0926200-d71b-11ea-9253-c7d28e2f062c.png)
v7.8.8, can't reproduce. Windows 10, 64-bit. The "X" button is getting focused, although the 1px margin seems to be a thing

Fonze:
> ![picture](https://user-images.githubusercontent.com/27734421/89405456-e0926200-d71b-11ea-9253-c7d28e2f062c.png)
> v7.8.8, can't reproduce. Windows 10, 64-bit. The "X" button is getting focused, although the 1px margin seems to be a thing

@kspalaiologos Just speculating, but is it possible that the your taskbars location at the top of the screen could affect this? I have mine at the bottom of the screen so I can push my cursor to the very top right pixel on the screen. And yes I forgot to mention that in my case the X gets highlighted too.

kspalaiologos:
hmm, tried moving it back to the bottom, didn't really change much. Maybe updating to the newest version will help?

KnIfER:
Not the close button, but anything on the right **(scollbar and dock panel)** is affected by this 1px padding

sasumner:
Related? : #5102 #7674 #8380

sasumner:
Also possibly related:  https://community.notepad-plus-plus.org/topic/20077/through-the-title-bar

Kabbah:
Just wanted to say I have a similar problem in my laptop.

[...]

PR #15837, opened 2024-11-21
Update dpiAware.manifest to support per-monitor DPI awareness

pawelzwronek:
After reading almost all of the issues related dpi awareness I found that for the current version of N++ setting `notepadd++.exe -> properties -> compatibility -> change high DPI settings -> Override...`  to `Application` works without any issue in my system configuration. No blurred window nor any moved buttons, see attached screenshots.
Also I found that setting `<dpiAware>` to `true/pw` was not tested often and this is a way to gain `Application`/`PerMonitor` setting, [see here](https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process).

I invite others to see if setting `notepad++.exe DPI settings` to `Application` and confirm that the app is ok in their multi-monior configurations.

```
Notepad++ v8.7.1   (32-bit)
Build time : Nov 21 2024 - 21:16:16
Path : D:\git\notepad-plus-plus\PowerEditor\visual.net\Debug\Notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 22H2
OS Build : 19045.4780
Current ANSI codepage : 1250
Plugins : none
```
obraz

Monitor 1:
obraz

Monitor 2:
obraz

obraz

### DIFF
--- a/PowerEditor/src/dpiAware.manifest
+++ b/PowerEditor/src/dpiAware.manifest
@@ -2,10 +2,10 @@
 <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
     <asmv3:application xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
         <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
-            <dpiAware>true</dpiAware>
+            <dpiAware>true/pm</dpiAware>
         </asmv3:windowsSettings>
         <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
-            <dpiAwareness>system, unaware</dpiAwareness>
+            <dpiAwareness>PerMonitor</dpiAwareness>
         </asmv3:windowsSettings>
         <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2017/WindowsSettings">
             <gdiScaling>false</gdiScaling>


donho:
@ozone10 
In your opinion, is it "safe" (no visual bug) for Notepad++ to transition to **PerMonitor** on the way to **PerMonitorV2**?

pawelzwronek:
I might to create this PR too soon, as I mentioned [here](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/8660#issuecomment-2494387469):
> @codeceptsDE I see. I didn't tested on such big zoom levels. It seems that in `Aplication` setting, N++ track the dpi level of the monitor where it lunches, but should react to moving between monitors better. I've tested `System(Enhanced)` and it seems to works quite good on larger zoom levels >= 150%. The blurriness on zoomed monitor is not as much visible the bigger the zoom is, and moving between monitors works good. On zoom level 125%, the blurriness is not acceptable and better option would be `Application` setting.

donho:
@pawelzwronek 
So this PR can be closed.
Could you close it please?

ozone10:
@donho 
PerMonitor requires processing `WM_DPICHANGED` so I would rather work with PerMonitorV2 directly, since it's supported by current MS supported OS. 

Also related issue https://github.com/notepad-plus-plus/notepad-plus-plus/issues/9513

Sorry I am currently very busy, so I don't have time to work on it. Mainly docked panels require work.

@pawelzwronek 
you can check https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14959 for what needs to be implemented for hidpi support.

PR #15840, opened 2024-11-22
Rename 'Darken inactive tabs'

### LINKED ISSUE #15800: [BUG] The tab bar's "Darken inactive tabs" should be renamed

### Description of the Issue

I'm calling this a bug because the name of the setting is "Darken inactive tabs" and the problem is with the word "darken" -- darkening is the default behavior, but it doesn't have to be used this way; it could be made to make the tabs "brighter" (depending upon the chosen color) as well.

### Steps To Reproduce

1.  Visit _Settings_ menu and choose _Preferences_
2.  Select _General_ along the left side of the dialog that appears after step 1
3.  Under _Tab Bar_ note that there is a checkbox with caption _Darken inactive tabs_

### Current Behavior

Less than accurate terminology in the UI.

### Expected Behavior

Better terminology.  :-)

A better caption for this checkbox would be:  "Color background of inactive tabs".

Ideally, there would also be a link next to this checkbox that would take the user to the _Style Configurator_ > _Global styles_ > _Inactive tabs_ area, much like this one in another area of the settings:
image
The hover bubble text could be:  "Go to Style Configurator to change the background color for inactive tabs"

And there could be a corresponding link back to settings at _Style Configurator_ > _Global styles_ > _Inactive tabs_, similar to this:
image
that would take the user back to _Settings_ > _Preferences_ > _General_ > _Tab Bar_ , when pressed.

### Debug Information

```shell
Notepad++ v8.7   (64-bit)
Build time : Sep 17 2024 - 17:06:31
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.7\npp.8.7.portable.x64\notepad++.exe
Command Line : -titleAdd="8.7 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Version : 23H2
OS Build : 22631.4317
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    PythonScript (2)
```

### Anything else?

_No response_

alankilborn:
Fix #15800 

I'd hope that even if the suggested name change isn't liked, the new back-and-forth links between _Preferences_ and _Style Configurator_ are at least a nice addition?


PowerEditor/installer/nativeLang/english.xml
```diff
@@ -954,7 +954,7 @@ Translation note:
 					<Item id="6106" name="Tab Bar"/>
 					<Item id="6107" name="Reduce"/>
 					<Item id="6108" name="Lock (no drag and drop)"/>
-					<Item id="6109" name="Darken inactive tabs"/>
+					<Item id="6109" name="Color background of inactive tabs"/>
```
xomx, line 957: @alankilborn 
IMO - That's not the gist of the meaning here, that "darken" is used here more as the opposite of "highlight". At least that is what I use in my translation. I am not a native English speaker, so maybe my English attempts will not be ok, but I would use here expressions like "Fade inactive tab" or "Weaken inactive tab color" or simply "Dim inactive tabs".

PowerEditor/installer/nativeLang/english_customizable.xml
```diff
@@ -954,7 +954,7 @@ Translation note:
 					<Item id="6106" name="Tab Bar"/>
 					<Item id="6107" name="Reduce"/>
 					<Item id="6108" name="Lock (no drag and drop)"/>
-					<Item id="6109" name="Darken inactive tabs"/>
+					<Item id="6109" name="Color background of inactive tabs"/>
```
xomx, line 957: ditto above

pawelzwronek:
I like the idea, but to avoid adding extra text for translators, I would change the button to an icon that speaks for itself, and remove the tooltip:
Also I would disable the button/icon when the checkbox is unchecked.

obraz

or just:

obraz

PowerEditor/src/WinControls/Preference/preference.rc
```diff
@@ -56,7 +56,7 @@ BEGIN
     CONTROL         "Reduce",IDC_CHECK_REDUCE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,229,47,164,10
     CONTROL         "Alternate icons",IDC_CHECK_TAB_ALTICONS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,229,60,164,10
     CONTROL         "Lock (no drag and drop)",IDC_CHECK_LOCK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,229,73,164,10
-    CONTROL         "Darken inactive tabs",IDC_CHECK_DRAWINACTIVE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,229,86,164,10
+    CONTROL         "Color background of inactive tabs",IDC_CHECK_DRAWINACTIVE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,229,86,164,10
```
donho, line 59: Between "Darken inactive tabs" & "Color background of inactive tabs", I'm rather for the 1st one.

donho:
For the naming, instead of "Color background of inactive tabs", how about "Contrast inactive tab" or "Shade inactive tab"?
What do you guys think?

alankilborn:
> instead of "Color background of inactive tabs", how about "Contrast inactive tab" or "Shade inactive tab"?

Well, "Contrast inactive tab" or "Shade inactive tab" doesn't really help the user to know what to do when he gets to Style Configurator (if he wants to change the "contrast"/"shading").  Using the word "background" here tells him exactly what he needs to do to change it, i.e., adjust the "Background" color setting in Style Configurator.

---

>Third option could be short "Color inactive tabs"

Decent, but still lacks the IMO critical "background".

---

>I would use here expressions like "Fade inactive tab" or "Weaken inactive tab color" or simply "Dim inactive tabs".

This misses the point of the issue.  
Is it "darkening", "fading" or "dimming" if I set the background color of inactive tabs to white? (no):

image

donho:
@alankilborn 
> I'd hope that even if the suggested name change isn't liked, the new back-and-forth links between Preferences and Style Configurator are at least a nice addition?

Indeed, "Go to Settings" link is a nice addition.
For the tooltip button, not mentioning it will reduce the room for the translation, a button among the group of checkbox looks wired. So for the sake of simplicity, I prefer to not have it.

pawelzwronek:
Fourth option `Background of inactive tabs`:

obraz

or

obraz

alankilborn:
So, as a user of Notepad++, I spent a lot of time last week trying to figure out why the background color of my inactive tabs was not working.  No matter how I changed the setting in Style Configurator, nothing happened.  I thought, this works in a fresh install, which one of my scripts is messing this up??.  I then spent a lot of time disabling/reenabling scripts, to no avail -- still no way I could change background color on inactive tabs.  This was _after_ I consulted the user manual (there were no hints there).

Long story short: I had absolutely no idea that _Darken inactive tabs_ being unchecked was preventing me from setting a background color.  How did I figure it out? It started with defaulting my `config.xml` and it wasn't long after that that it dawned on me.

My point: I'm a power user (I think), and a non-power user could also be confused by this.  Shouldn't we do all we can to help, e.g. a name change _and_ a button to link to the other setting that is affected by the option?

pawelzwronek:
@alankilborn I agree that that there should be a hint in Style Configurator about the check box in Preferences or check it in the code when a user change the color by hand.

pawelzwronek:
Whole Preferences window, where you see at first glance you can change the color of one of the option

obraz

pryrt:
As with @alankilborn , I think **Darken inactive tabs** is not the best choice, because dark-vs-light depends on another setting as to whether it is accurate or not.  I'm not sure that any of the alternatives to **Color background of inactive tabs** that have been suggested so far really fulfill the brief, IMO.  

I'm not always great at picking GUI labels, but I think **Change background of inactive tabs** or **Change inactive tab background** might be good choices (without trying to use "Color" as a verb, which probably does confuse non-native speakers, since except in "color between the lines" and other coloring-book nomenclature, "color" is almost always a noun, not a verb, in English).

And having some sort of a link from the Preference to the Style Configurator entry is exceedingly useful, for the exact same reasons that Custom Color for Non-Printing Characters and EOL link to the Style Configurator; and similarly, having a link from Style Configurator's **Inactive tabs** entry back to the Preference page is just as useful and important as the existing links from ** EOL custom color** and **Non-printing characters custom color** back to their associated preference.  Cross-linking makes the preference-and-color settings _much_ more user-friendly.

alankilborn:
@pawelzwronek 

I like your suggestion that there should also be a button for _Draw a colored bar on active tab_.
This makes it clearer that those buttons are related to the checkboxes that they are vertically aligned with (and not some "general" button that applies to all _Tab bar_options; although if that part was confusing all one has to do is read the button's hover text in order to understand what it applies to).

I don't know about the "eye dropper" icon on the button, first, because it is harder (more code) to do than the simple `...` button caption, and also because to me it implies that it might _immediately_ open a "standard" color picker.

pawelzwronek:
@alankilborn we can change the icons to more suitable but cant find any better than that:

obraz

pryrt:
> we can change the icons to more suitable but cant find any better than that:

The same `[...]` button that is currently used on the other preferences with links to the Style Configurator (which I mentioned above) should be used for UI consistency.

alankilborn:
>As for buttons, I would say they are too brutal/rough and user will think twice before hitting them. 

A cautious user will point to the `...` button, slowly, and the hover text will pop up, giving user confidence.  :-)

>make N++ more modern and eye pleasing we should go towards minimalism and put more icons instead of buttons. I like Fluent UI icons very much. If we could puts more such icons to the interface it would be for better.

They are nice icons but as @pryrt said, we need to stay consistent in the UI and the `...` button is used elsewhere.
Maybe a larger change later could change buttons to icons.

pawelzwronek:
Like this?

obraz

alankilborn:
>Like this?

Yes, I think that is ideal for now.
I didn't think to add the 2nd `...` button in the original PR (and it isn't related to exactly what I was changing).
But, if @donho approves of it, I'll certainly add it.

pawelzwronek:
Group a style related setting together and move to the end:

obraz

donho:
"Change inactive tab background" is not bad, but it's still too long. I would use **"Change inactive tab color"**. I think it's the best compromise: when users enable this option, they see the bg color of inactive tabs are changed, so it's obviously for the background color. Also the terms echo each other with *"Inactive tab"* of Styler Configurator and *"Change inactive tab color"* of this option - if there's "Go to Settings" link there - no more confusion.

Personally I like the [suggestion ](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15840#issuecomment-2494453887)of @pawelzwronek (with 2 extra icons), but as @alankilborn & @pryrt said, the consistence in the UI should be kept. OTOH, I rethinking about the simplicity, and I ask myself if the extra button & tip are really necessary.

From my viewpoint, it's not. For the following reason:
My ideal application is, I run it when I need it, I get the job done and I close it without even I notice its existence. That's the main goal when I try to build Notepad++. With the principle I mentioned above, I try to see if Notepad++ can do without the extra button with the the aspect of normal users. 
A. A new user come into the preferences, he/she sees the option **Change inactive tab color** is enabled (default behaviour), the user disables it and enables it then he/she sees the difference of tab background color - it's enough for this user.
B. The same user has experienced Notepad++ for several weeks or even several months, and he/she comes into Style Configurator and found the "Inactive tab" item in Global Styles. The user may not understand what it is, but then he/she click on "Go to Settings", it will bring him/her to **Change inactive tab color** option, and the user will know what it is about.

Both 2 above scenarios are enough (for me) to a simplest application. As a result, no button will be necessary to add behind any checkbox.

Hope my arguments are understandable.

alankilborn:
>Please consider my suggestions

I will make those modifications to the PR.

pawelzwronek:
@donho I see your point of view, those button are just shortcuts and might overwhelmed the user and are unnecessary unless the change is visible on checking the checkbox. 
But, such change is invisible when user go into Style Configurator and change the color of Inactive tab when "Change inactive tab color" is unchecked, so vital would be or disable/shadeout a color picker in Style Configurator and leave only "Go to Settings" link enabled, or check/set in the code "Change inactive tab color" checkbox when the color of Inactive tab is changed by the user. Third option would be warn a user with an info message "Enable ... in Settings to see this change" or sth like that, but as you might already noticed I'm an enemy of avoidable message dialogs 😈 4. option would be a hint below the ""Go to Settings" link with a copy of the name of the checkbox, so `(Change inactive tab color)` in this situation.

donho:
@pawelzwronek 
I think we have the different definition about "avoidable message dialog".

pawelzwronek:
@donho I must have not clarify enough my point of view in a discussion about deleted files. For me a message boxes interrupt a workflow, because all messageboxes looks the same and you are forced to read the text carefully to take right decision. In the situation of deleted files, a red icon is an instant indicator that sth has gone wrong and IMO in 99% of situation there is no need to bother the user with a messagebox. For this 1% of a corrupted file the user still have an option to save the tab.

And what do you think about the rest of given options about changing the color of inactive tabs?

donho (approved):

PR #15849, opened 2024-11-24
[xml] Update Portuguese.xml

hugok79:
@horus68 @Blackspirits

### DIFF
--- a/PowerEditor/installer/nativeLang/portuguese.xml
+++ b/PowerEditor/installer/nativeLang/portuguese.xml
@@ -5,7 +5,7 @@ Nota de tradução:
 2. Todos os comentários são de caráter explicativo, não são de tradução.
 -->
 <NotepadPlus>
-	<Native-Langue name="Portuguese Portugal" filename="portuguese.xml" version="8.7.1">
+	<Native-Langue name="Portuguese Portugal" filename="portuguese.xml" version="8.7.2">
 		<Menu>
 			<Main>
 				<!-- Main Menu Entries -->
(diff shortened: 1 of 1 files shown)


Blackspirits:
@hugok79 @horus68

In general, I would only make a few small changes

I would change “Minimizar / Fechar para” -> “Minimizar/Fechar para”
because you don't normally put a space before and after the slash

"Aparência" -> "Aspeto"
chrome, edge and many apps use the word “Aspeto”

would change all strings with the word “usando” to “utilizando”
“usando” sounds very Brazilian to me, I believe that we Portuguese use ‘utilizando’ more, but it often also depends on the context, but on a computer level I think ‘utilizando’ is better.

hugok79:
@Blackspirits @horus68
I see that I'm on the same line of thought as both of you!
In the case of the word `Aparência` I agree with @Blackspirits because I recently changed that particular word in a web browser(Vivaldi) that I translate and I think that to maintain better consistency the word `Aspeto` is better.
I also avoid using the "gerúndio"(like the word "você" for instance)as much as possible and always prefer to use the words "utilizar/utilizando" instead of "usar/usando"
I've also changed the string `Minimizar/Fechar para`

donho (changes requested):
Please add a new entry also in Tabbar context menu, as latest commit in english.xml:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/f18d99e24cb3609f912e7c7bf0f1d3de2b48268f

kalil-camera:
Isn't time to separate PT and PT-BR? I grew in Brazil for the last 40 years and never heard or read Aspeto in my life.

Blackspirits:
If you're Brazilian and live in Brazil it's normal that you don't understand, pt-PT and pt-BR have a lot of differences.
And you're in the wrong section, there's been pt-BR since 2009.
[https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/brazilian_portuguese.xml](https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/installer/nativeLang/brazilian_portuguese.xml)

hugok79:
@kalil-camera https://www.infopedia.pt/dicionarios/lingua-portuguesa/aspeto?express=Aspeto

hugok79 (commented):
@donho Done

donho (approved):

PR #15851, opened 2024-11-25
Fix possible DockingManager::FindEmptyContainer() buffer overflow

### LINKED ISSUE #15850: [BUG] DockingManager::FindEmptyContainer() possible buffer overflow

### Description of the Issue

Due to the N++ Docking Manager design imperfection, there is a possibility of a buffer overflow in its `DockingManager::FindEmptyContainer()` method.

Erroneous memory overwriting in the above mentioned func usually does not lead to an immediate visible N++ crash, but if the user continues to use the undocking/docking of the N++ panels during the app life, it does lead to constant corruption of unpredictable parts of the app heap memory and consequently to the possible app-crash (or worse - to a possible data corruption).

This bug can also be intentionally exploited if an attacker manually changes the saved value of `prev="..."` to, say, `prev="99999999"`. Then the N++ would crash immediately upon any attempt to undock a panel.

### Steps To Reproduce

1.  Build a fresh Debug build of the N++ and launch it.
2.  Ctrl+F and then click `Find All in Current Document`
3.  Undock the `Search results` panel to the floating state
4.  Redock it back
5.  Close N++

Now in your relevant config.xml should be these Docking Manager data:
```
        <GUIConfig name="DockingManager" leftWidth="200" rightWidth="200" topHeight="200" bottomHeight="200">
            <FloatingWindow cont="4" x="183" y="344" width="878" height="624" />
            <PluginDlg pluginName="Notepad++::InternalFunction" id="0" curr="3" prev="4" isVisible="yes" />
            <ActiveTabs cont="0" activeTab="-1" />
            <ActiveTabs cont="1" activeTab="-1" />
            <ActiveTabs cont="2" activeTab="-1" />
            <ActiveTabs cont="3" activeTab="0" />
            <ActiveTabs cont="4" activeTab="-1" />
        </GUIConfig>
```

6.  Launch the N++ again and close it immediately.

Now your relevant config.xml data should be like:
```
        <GUIConfig name="DockingManager" leftWidth="200" rightWidth="200" topHeight="200" bottomHeight="200">
            <PluginDlg pluginName="Notepad++::InternalFunction" id="0" curr="3" prev="4" isVisible="yes" />
            <ActiveTabs cont="0" activeTab="-1" />
            <ActiveTabs cont="1" activeTab="-1" />
            <ActiveTabs cont="2" activeTab="-1" />
            <ActiveTabs cont="3" activeTab="-1" />
        </GUIConfig>
```

7. Set a breakpoint in MSVS on the `DockingManager::FindEmptyContainer()` method.
8. Launch a N++ debugging session.
9. Ctrl+F in N++ and then click `Find All in Current Document`
10. Try to undock the `Search results` panel to the floating state.
11. Now watch in debugger how the `pPrevDockList` buffer is being corrupted (see the pics below).
12. Buffer overflow heap manager msgbox warning, when you reach the `delete [] pPrevDockList;` line.

### Current Behavior

npp-DockingManager-buffer-overflow-before

npp-DockingManager-buffer-overflow-after

npp-DockingManager-buffer-overflow-warning-msg

### Expected Behavior

Such buffer overflow is not possible under any circumstances.

### Debug Information

```shell
Probably any N++ version (with the docking panels possibility).
```

### Anything else?

_No response_

xomx:
Fix #15850

### DIFF
--- a/PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
+++ b/PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
@@ -951,25 +951,25 @@ int DockingManager::GetContainer(DockingCont* pCont)
 
 int DockingManager::FindEmptyContainer()
 {
-    int      iRetCont       = -1;
+	int iRetCont = -1;
 	const size_t dockingContVectorSize = _vContainer.size();
 	const size_t prevDockListBufSize = dockingContVectorSize + 1;
 	BOOL* pPrevDockList = new BOOL[prevDockListBufSize];
 	BOOL* pArrayPos = &pPrevDockList[1]; // make a room for the possible iPrevCont==-1 later
 
-    // reset all entries
-    for (size_t iCont = 0, len = prevDockListBufSize; iCont < len; ++iCont)
-    {
-        pPrevDockList[iCont] = FALSE;
-    }
+	// reset all entries
+	for (size_t iCont = 0, len = prevDockListBufSize; iCont < len; ++iCont)
+	{
+		pPrevDockList[iCont] = FALSE;
+	}
 
-    // search for used floated containers
-    for (size_t iCont = 0; iCont < DOCKCONT_MAX; ++iCont)
-    {
-        vector<tTbData*>    vTbData = _vContainer[iCont]->getDataOfAllTb();
+	// search for used floating containers
+	for (size_t iCont = 0; iCont < DOCKCONT_MAX; ++iCont)
+	{
+		vector<tTbData*> vTbData = _vContainer[iCont]->getDataOfAllTb();
 
-        for (size_t iTb = 0, len = vTbData.size(); iTb < len; ++iTb)
-        {
+		for (size_t iTb = 0, len = vTbData.size(); iTb < len; ++iTb)
+		{
 			if ((vTbData[iTb]->iPrevCont < static_cast<int>(dockingContVectorSize)) && (vTbData[iTb]->iPrevCont >= -1))
 			{
 				pArrayPos[vTbData[iTb]->iPrevCont] = TRUE;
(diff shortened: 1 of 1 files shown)


xomx:
@donho 

I encountered this bug while working on another docking stuff bug.
I did not reformat the func in question yet for better fix readability.
If you accept the fix, I can handle the mixture of tabs and spaces there(?).

PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
```diff
@@ -968,12 +970,21 @@ int DockingManager::FindEmptyContainer()
 
         for (size_t iTb = 0, len = vTbData.size(); iTb < len; ++iTb)
         {
-            pArrayPos[vTbData[iTb]->iPrevCont] = TRUE;
+			if ((vTbData[iTb]->iPrevCont < static_cast<int>(dockingContVectorSize)) && (vTbData[iTb]->iPrevCont >= -1))
+			{
+				pArrayPos[vTbData[iTb]->iPrevCont] = TRUE;
```
donho, line 975: When **vTbData[iTb]->iPrevCont == -1**, it satisfies the condition `vTbData[iTb]->iPrevCont >= -1`, then `pArrayPos[-1] = TRUE;` will make an explosion.

PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
```diff
@@ -968,12 +970,21 @@ int DockingManager::FindEmptyContainer()
...
         {
-            pArrayPos[vTbData[iTb]->iPrevCont] = TRUE;
+			if ((vTbData[iTb]->iPrevCont < static_cast<int>(dockingContVectorSize)) && (vTbData[iTb]->iPrevCont >= -1))
+			{
+				pArrayPos[vTbData[iTb]->iPrevCont] = TRUE;
+			}
+			else
+			{
+				// ? invalid config.xml input data
+				assert((vTbData[iTb]->iPrevCont < static_cast<int>(dockingContVectorSize)) && (vTbData[iTb]->iPrevCont >= -1));
```
donho, line 980: Is assert necessary?

donho (changes requested):
Please check the condition to make sure there won't be the exception.

PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
xomx: @donho 
No, `-1` is OK here.
It is a Docking Manager expected value meaning `no-previous-one` and the original code is prepared for that (I only added a comment there to explain better that code meaning):
https://github.com/xomx/notepad-plus-plus/blob/ceead7526bb120bee465531f26fd5cb0de347209/PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp#L958

PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
xomx: I would prefer to have this assert there so that one can at least in the Debug build be notified that the current config.xml data for the Docking Manager is not quite right.

PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
donho: pArrayPos is a pointer of BOOL which points to the 2nd position of a BOOL array.
So I don't understand why `pArrayPos[-1]` won't cause the runtime error.

xomx: @donho 

npp-pArrayPos-minus-one-index-area

donho: Holly S... I didn't see that coming! (Docking system was done about 18 years ago, by a contributor)
Surely it's not a good practice, but do we have any better way for it?

donho:
@xomx 
> If you accept the fix, I can handle the mixture of tabs and spaces there(?).

Please go ahead.

donho (approved):

PowerEditor/src/WinControls/DockingWnd/DockingManager.cpp
donho: BTW, I thought index `-1` in the `[` `]` will be translated to an unsigned int (size_t?) on the runtime?

donho: OK, it's about the **POINTER** but not **ARRAY**
https://stackoverflow.com/questions/47170740/c-negative-array-index

xomx:
> > If you accept the fix, I can handle the mixture of tabs and spaces there(?).
> 
> Please go ahead.

@donho 
All done.

PR #15852, opened 2024-11-25
Add Pin/Unpin Tab context menu item

donho:


xomx:
@donho 
Found a glitch - when I pin/unpin via this new popup menu a tab, which does not move when pining/unpining (e.g. pin/unpin only the 1st N++ tab), then its pin-ico changes accordingly only if I move mouse over (or do something with the other N++ tabs like switching their order etc)

PR #15870, opened 2024-11-28
remove pointless "File to load size-check failed" message

### LINKED ISSUE #15869: [Regression] Drag&drop a folder in Notepad++ launch "File to load size-check failed" dialog

### Description of the Issue

This is a v8.7.2 regression caused by the commit https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c021c861959765db9efa0a7d6b156538a0e6f4db

Reported at https://community.notepad-plus-plus.org/post/98250

### Steps To Reproduce

Try to drag&drop a folder with files on the N++ app icon or launch the N++ via a cmdline `notepad++.exe FolderWithFilesToOpen` or try to use globbing at the N++ launch.

### Current Behavior

Pointless message, then the N++ opens that files in dir ok, etc.

### Expected Behavior

No such message, only returning `BUFFER_INVALID` in `FileManager::loadFile` immediately (it also speeds up the loading stage a bit).

### Debug Information

```shell
N++ v8.7.2
```

### Anything else?

_No response_

xomx:
Fix #15869 .

Only returning BUFFER_INVALID in FileManager::loadFile immediately if the file size check == -1. This is a must here because of the following code does not expect such -1 in fileSize.

But an error msg is not appropriate here, as it could also be a legal attempt to open a dir (with files inside) or an attempt to use the N++ globbing feature.

This early return also speeds up the loading stage a bit.

### DIFF
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -729,7 +729,7 @@ BufferID FileManager::loadFile(const wchar_t* filename, Document doc, int encodi
 	if (fileSize == -1)
 	{
 		// we cannot continue (or Scintilla will throw SC_STATUS_FAILURE in the loadFileData later)
-		::MessageBox(_pNotepadPlus->_pEditView->getHSelf(), L"Cannot obtain the file size before loading!", L"File to load size-check failed", MB_OK | MB_APPLMODAL);
+		// - no error message here as this can be also a dir with files to open or globbing attempt
 		return BUFFER_INVALID;
 	}
 


PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -729,7 +729,7 @@ BufferID FileManager::loadFile(const wchar_t* filename, Document doc, int encodi
 	if (fileSize == -1)
 	{
 		// we cannot continue (or Scintilla will throw SC_STATUS_FAILURE in the loadFileData later)
-		::MessageBox(_pNotepadPlus->_pEditView->getHSelf(), L"Cannot obtain the file size before loading!", L"File to load size-check failed", MB_OK | MB_APPLMODAL);
+		// - no error message here as this can be also a dir with files to open or globbing attempt
```
donho, line 732: 👍

PR #15871, opened 2024-11-28
Update sql.xml

demetrio72:
My contribution.
It recognizes packages though not yet as classes, minor fixes to expressions, deleted comments until bug is fixed


rdipardo:
Changes to the Function List require [unit tests](https://github.com/notepad-plus-plus/notepad-plus-plus/tree/master/PowerEditor/Test/FunctionList/sql):

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/6229e0f5d47687ca7be19dad4b32f00edb3f3dbb/CONTRIBUTING.md?plain=1#L33

donho (changes requested):
1. Fix build error
2. Add Unit tests

PR #15874, opened 2024-11-29
Fix v8.6.9+ possible opened docked panels invisibility in multi-instance mode

### LINKED ISSUE #15873: [BUG] Docked panels in v8.6.9+ can be invisible in multi-instance mode

### Description of the Issue

When one uses any Notepad++ v8.6.9+ in multi-instance mode and left the 1st instance launched minimized, then any other N++ instance will have its docked panels invisible.

### Steps To Reproduce

https://community.notepad-plus-plus.org/post/98013

### Current Behavior

Cannot see the opened panels.

### Expected Behavior

Panels can be used at the described setup.

### Debug Information

```shell
Any Notepad++ v8.6.9+
```

### Anything else?

_No response_

xomx:
Fix #15873 and many Community reports like [this](https://community.notepad-plus-plus.org/topic/26164/v8-7-search-results-missing).

When Notepad++ v8.6.9+ was set to "Always in multi-instance mode" and the 1st instance has been minimized to the task-bar, then any subsequently launched instance had its opened docked panels invisible.

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -6774,9 +6774,8 @@ void NppParameters::feedDockingManager(TiXmlNode *node)
 	HWND hwndNpp = ::FindWindow(Notepad_plus_Window::getClassName(), NULL);
 	if (hwndNpp)
 	{
-		// TODO: 
-		// the problem here is that this code-branch cannot be currently reached
-		// (as it is called at the Notepad++ startup in the wWinMain nppParameters.load())
+		// this code-branch is currently reached only if the Notepad++ multi-instance mode is ON and it is not the 1st Notepad++ instance
+		// (the feedDockingManager() is called at the Notepad++ init via the wWinMain nppParameters.load()))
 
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Parameters.cpp
```diff
@@ -6774,9 +6774,8 @@ void NppParameters::feedDockingManager(TiXmlNode *node)
 	HWND hwndNpp = ::FindWindow(Notepad_plus_Window::getClassName(), NULL);
 	if (hwndNpp)
 	{
-		// TODO: 
-		// the problem here is that this code-branch cannot be currently reached
-		// (as it is called at the Notepad++ startup in the wWinMain nppParameters.load())
+		// this code-branch is currently reached only if the Notepad++ multi-instance mode is ON and it is not the 1st Notepad++ instance
+		// (the feedDockingManager() is called at the Notepad++ init via the wWinMain nppParameters.load()))
```
donho, line 6778: 👍

donho:
https://github.com/notepad-plus-plus/notepad-plus-plus/commit/b43eba3f4bc1a0fda65cf550a15e7157f4011e9c

PR #15879, opened 2024-11-29
Update sql.xml

demetrio72:
Sorry I'm not able to make tests using command line tool tool because I cannot use git locally on my PC. I tested the file unitTest.sql and as far I can see it work, I send you a screenshot to let yo see the result. 
Note that I'm still not able to indent functions and procedures in package properly so they appears unlinked but the output is still good enough. It's usefull to keep package definition in functions list in order to jump directly to declaration, this is much more usefull when you have package declaration an body in the same file.
sql xml test

PS: I'm not sure I sent you the screenshot because a firewall popup blocked me.

### DIFF
--- a/PowerEditor/installer/functionList/sql.xml
+++ b/PowerEditor/installer/functionList/sql.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.1" encoding="UTF-8" ?>
 <!-- ==========================================================================\
 |
 |   To learn how to make your own language parser, please check the following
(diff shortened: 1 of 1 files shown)


donho:
@demetrio72 
Could you provide your sql sample file that doesn't work well with the current parser, but works better with your PR please?

demetrio72:
Hi.
I opened another fork (demetrio72-patch-20250529) to push a completely new
version of sql.xml (today I made a little fix)
The output si different than the unitTest.expected.result file is
different, and it's correct. So I tried to align latter file with the new
output, but it looks like the build always take older reference file.

How can I solve the situation?

Thanks
Demetrio

Il lun 2 dic 2024, 18:29 Don HO ***@***.***> ha scritto:

PR #15888, opened 2024-11-30
Check user defined extensions for styles first when detecting lang from extension

### LINKED ISSUE #15826: [BUG] Cannot override language selection for .nfo files

### Description of the Issue

I would like to use the "ini file" language definition for .nfo files, but it seems to be hardcoded for "Dos Style" language.
I have set the nfo file exention
image

But it still open as Dos style even after restart of NPP

### Steps To Reproduce

NA

### Current Behavior

NA

### Expected Behavior

NA

### Debug Information

```shell
Notepad++ v8.7.1   (64-bit)
Build time : Oct 31 2024 - 00:48:56
Command Line : "C:\data\Development\LIBRARIES WIT\W-CommonLib\W-CLib_1.03.nfo"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Version : 21H1
OS Build : 19043.1466
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

_No response_

pawelzwronek:
Before when a language was detected from an extension, a list of languages was looped from the end. When matching extension was found for a default extension or used defined extension, loop ended.

Now when a user add an extension from build-in languages (e.g. .nfo) in Style Configurator, N++ will prioritizes this setting, by looking this list first.

Fix #15826

### DIFF
--- a/PowerEditor/src/Parameters.cpp
+++ b/PowerEditor/src/Parameters.cpp
@@ -3224,32 +3224,26 @@ bool NppParameters::exportUDLToFile(size_t langIndex2export, const std::wstring&
 
 LangType NppParameters::getLangFromExt(const wchar_t *ext)
 {
-	int i = getNbLang();
-	i--;
+	// first check a user defined extensions for styles
+	LexerStylerArray &lexStyleList = getLStylerArray();
+	for (size_t i = 0 ; i < lexStyleList.getNbLexer(); ++i)
+	{
+		LexerStyler &styler = lexStyleList.getLexerFromIndex(i);
+		const wchar_t *extList = styler.getLexerUserExt();
+
+		if (isInList(ext, extList)) {
+			return getLangIDFromStr(styler.getLexerName());
+		}
+	}
+
+	// then check languages extensions
+	int i = getNbLang() - 1;
 	while (i >= 0)
 	{
 		Lang *l = getLangFromIndex(i--);
-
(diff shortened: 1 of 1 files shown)


PowerEditor/src/Parameters.cpp
```diff
@@ -3224,32 +3224,26 @@ bool NppParameters::exportUDLToFile(size_t langIndex2export, const std::wstring&
...
-	int i = getNbLang();
-	i--;
+	// first check a user defined extensions for styles
+	LexerStylerArray &lexStyleList = getLStylerArray();
+	for (size_t i = 0 ; i < lexStyleList.getNbLexer(); ++i)
+	{
+		LexerStyler &styler = lexStyleList.getLexerFromIndex(i);
+		const wchar_t *extList = styler.getLexerUserExt();
+
+		if (isInList(ext, extList)) {
```
donho, line 3234: Please use 
```cpp
if (isInList(ext, extList))
{
...
```
instead of 
```cpp
if (isInList(ext, extList)) {
...
```

donho (changes requested):
Please follow the coding style.

PowerEditor/src/Parameters.cpp
pawelzwronek: My bad, fixed.

donho (approved):

Issue #15819, opened 2024-11-18
[BUG] 8.7.1 introduced a network file modification detection bug.


cmanley:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

This release as introduced a new bug that didn't exist in v8.7.0 and prior:

I open/edit files on network drives (rclone SFTP mount), and now I'm frequently getting popup messages that the underlying file has been modified with the question if I want to reload it. I choose no, and then it says the file has gone missing with the question if I want to keep it open in my editor. I choose yes, and then save it. This process keeps on repeating. It's so annoying. I suspect that it was introduced by one of the network bug fixes listed in the release notes for 8.7.1

### Steps To Reproduce

1.  Install 8.7.1
2.  Install and configure a rclone SFTP mount in Windows
3.  Open and save a file on that mount.
4. Wait for the annoying popup dialog.

### Current Behavior

Much annoyance.

### Expected Behavior

No annoyance as in 8.7.0 and prior.

### Debug Information

```shell
I downgraded in order to work so I don't have this anymore.
```

xomx:
@cmanley 

Please add to your 1st post the requested Notepad++ `menu > ? > Debug Info...`.

> I open/edit files on network drives (rclone SFTP mount), and now I'm frequently getting popup messages that the underlying file has been modified with the question if I want to reload it. I choose no, and then it says the file has gone missing with the question if I want to keep it open in my editor.

From your description it looks like the N++ file checking code gets info from the OS that the file is not present anymore. I am wondering if it cannot be some type of a SFTP connection timeout? (I know you said the v8.7.0 has not the same problem, so this is probably not so relevant here).

How is the exact sequence of your steps and at which stage you have got the 1st message?
- right after the file opening
- after some specific time while still working with the opened network file in the N++
- do you need to switch to another app (or another N++ tab) and back to N++ (or to the previously opened problematic network file tab in N++) before the 1st message


> 2\. Install and configure a rclone SFTP mount in Windows

Can you replicate your network saving problem without this specific rclone network storage usage (I mean mount/reach your network storage in another way to exclude the rclone influence to narrow down the possibilities...)?
[...]

PR #15890, opened 2024-11-30
[TEST] Network file change detection issue

xomx:
Test for #15819 .


xomx:
Closing, will be replaced by a final version without the excessive debug code.

Issue #15903, opened 2024-12-03
[Enhancement] Add some more info to the Notepad++ Debug Info


xomx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

Including additional information can speed up the diagnostic of user-reported issues.
e.g.: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15793#issuecomment-2474298926

### Describe the solution you'd like.

Additional info that may be related to common N++ problems.

### Debug Information

```shell
v8.7.3+
```

Issue #15793, opened 2024-11-13
[BUG] Some Unicode characters rendered incorrectly at beginning of line


agaace:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

Some Unicode characters (but not all!) that are from the Basic Multilingual Plane are incorrectly rendered some of the time. It usually occurs when they're at the beginning of the line. Compare to rendering in Windows Notepad, Visual Studio Code, Chrome or Edge browser. All of them render this content correctly.

1. Have Notepad++ configured to use default font (Courier New, size 10).
2. Open the attached file Repro.txt.
3. Observe how the U+25B4 character (small triangle) is rendered correctly in line 1 where it's not at the beginning of a new line.
4. Observe how the same character is rendered incorrectly (missing glyph) in line 2 where it's at the beginning of a new line. Adding another multiplane character in front of it, such as the U+25B2 (big triangle) fixes the rendering of U+25B4.
5. In line 5, delete the first character (➤) and see what happens to the rendering of the remaining characters.

[Repro.txt](https://github.com/user-attachments/files/17724894/Repro.txt)

![image](https://github.com/user-attachments/assets/5d90a6d6-c945-4c16-816e-f85fca3bbea8)

### Steps To Reproduce

1. Have Notepad++ configured to use default font (Courier New, size 10).
2. Open the attached file Repro.txt.
3. Observe how the U+25B4 character (small triangle) is rendered correctly in line 1 where it's not at the beginning of a new line.
4. Observe how the same character is rendered incorrectly (missing glyph) in line 2 where it's at the beginning of a new line. Adding another multiplane character in front of it, such as the U+25B2 (big triangle) fixes the rendering of U+25B4.
5. In line 5, delete the first character (➤) and see what happens to the rendering of the remaining characters.

[Repro.txt](https://github.com/user-attachments/files/17724894/Repro.txt)

![image](https://github.com/user-attachments/assets/5d90a6d6-c945-4c16-816e-f85fca3bbea8)

### Current Behavior

4. Observe how the same character is rendered incorrectly (missing glyph) in line 2 where it's at the beginning of a new line. Adding another multiplane character in front of it, such as the U+25B2 (big triangle) fixes the rendering of U+25B4.
5. In line 5, deleting the first character (➤) makes the rest of the characters rendered incorrectly.

![image](https://github.com/user-attachments/assets/3c2378da-a26d-4ffa-965c-218dba82ba3c)

[...]

PR #15904, opened 2024-12-03
Some more Debug Info

xomx:
Implement #15903 .

Adding:
+ Placeholders
+ DirectWrite
+ Multi-instance Mode
+ File Status Auto-Detection
+ Dark Mode

### DIFF
--- a/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
+++ b/PowerEditor/src/WinControls/AboutDlg/AboutDlg.cpp
@@ -230,6 +230,61 @@ intptr_t CALLBACK DebugInfoDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM 
 			_debugInfoStr += nppGui.isSnapshotMode() ? L"ON" : L"OFF";
 			_debugInfoStr += L"\r\n";
 
+			// Placeholders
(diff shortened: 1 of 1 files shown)


xomx:
@ozone10 
Is there another one you'd like to see?
I'm considering some sort of Display Info there (e.g. number of monitors present, virtual screen rectangle, monitors enum with resolution and scaling, cur N++ mon).

ozone10:
@xomx 
I am late :(
N++ is currently using only system scaling, so maybe only system dpi value should enough for now.

donho:
@ozone10 
> N++ is currently using only system scaling, so maybe only system dpi value should enough for now.

There is nothing about dpi thing in added info. Or is it not right PR you were commenting?

xomx:
@donho 
> There is nothing about dpi thing in added info. Or is it not right PR you were commenting?

This PR was a reaction to this @ozone10 suggestion: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15793#issuecomment-2474298926

I included also some other info, which I considered most relevant for a possible future users problems solving.

Then I asked him if he does not have any other suitable candidates, but you closed this PR before he could react.

I talked specifically about the possible Display Info because there are a lot of issues reported, which are related to this area (e.g. problems with the multi-monitors setups, HiDpi, etc.)

PR #15911, opened 2024-12-05
[xml] Update Serbian (cyrillic & latin) to v8.7.4

rammba:


PowerEditor/installer/nativeLang/serbian.xml
```diff
@@ -5,7 +5,7 @@
 		- Radoš Milićev (https://github.com/rammba)
 -->
 <NotepadPlus>
-	<Native-Langue name="Srpski(Serbian)" filename="serbian.xml"  version="8.7.2">
```
rammba, line 8: In language dropdown in the preferences dialog only label `Srpski` is present, so I omitted english name. Is this ok?

PR #15921, opened 2024-12-08
allow using the Scintilla DirectWrite mode in non-Core Server Windows editions

### LINKED ISSUE #15916: [BUG] DirectWrite / Font Ligature support not working on Server 2025

### Description of the Issue

Font Ligature support is functioning on my Server 2025 install in other applications such as Windows Terminal, but in Notepad++ the DirectWrite option gets unchecked at startup, which I believe is causing font ligature support to be disabled. 

I have ensured that the DirectX runtimes are installed and available, and have rebooted the machine. 

I have tried running as both Administrator and standard user.

### Steps To Reproduce

1.  Use a font that supports ligatures (Cascadia Code or FiraCode)
2.  Enabled DirectWrite support from the Settings menu
3.  Restart Notepad++ to find the option toggled off

### Current Behavior

DirectWrite option gets unchecked at application startup, which I believe is causing font ligature support to be disabled. 

### Expected Behavior

DirectWrite is supported on this OS and has all dependencies available for use

### Debug Information

```shell
Notepad++ v8.7.4   (64-bit)
Build time : Dec  4 2024 - 23:50:05
Command Line : "C:\Users\Administrator\AppData\Roaming\Notepad++\config.xml" 
Admin mode : ON
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
Placeholders : OFF
DirectWrite : OFF
Dark Mode : ON
OS Name : Windows Server 2025 Standard (64-bit)
OS Version : 24H2
OS Build : 26100.2314
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
```

### Anything else?

Attached debug information is from a fresh startup, when the DirectWrite option was forced off.

xomx:
Fix #15916 .

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -318,7 +318,15 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 
 	if (isWINE || // There is a performance issue under WINE when DirectWrite is ON, so we turn it off if user uses Notepad++ under WINE
 		isCoreWindows()) // In the case of Windows Server Core, DirectWrite cannot be on.
-		nppGui._writeTechnologyEngine = defaultTechnology;
+	{
+		nppGui._writeTechnologyEngine = directWriteTechnologyUnavailable;
+	}
+	else
+	{
+		// allow IDC_CHECK_DIRECTWRITE_ENABLE to be set in Preferences > MISC. again
+		if (nppGui._writeTechnologyEngine == directWriteTechnologyUnavailable)
+			nppGui._writeTechnologyEngine = defaultTechnology;
+	}
 
 	if (nppGui._writeTechnologyEngine == directWriteTechnology)
 	{
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -317,7 +317,7 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 		isWINE = ::GetProcAddress(hNtdllModule, "wine_get_version");
 
 	if (isWINE || // There is a performance issue under WINE when DirectWrite is ON, so we turn it off if user uses Notepad++ under WINE
-		::IsWindowsServer()) // In the case of Windows Server Core, DirectWrite cannot be on.
+		isCoreWindows()) // In the case of Windows Server Core, DirectWrite cannot be on.
 		nppGui._writeTechnologyEngine = defaultTechnology;
 
 	if (nppGui._writeTechnologyEngine == directWriteTechnology)
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1906,3 +1906,62 @@ bool doesPathExist(const wchar_t* path, DWORD milliSec2wait, bool* isTimeoutReac
 	getFileAttributesExWithTimeout(path, &attributes, milliSec2wait, isTimeoutReached);
 	return (attributes.dwFileAttributes != INVALID_FILE_ATTRIBUTES);
 }
+
+
+#pragma warning(disable:4996) // 'GetVersionExW': was declared deprecated
+bool isCoreWindows()
+{
+	bool isCoreWindows = false;
+
+	// older Windows (Windows Server 2008 R2-) check 1st
+	OSVERSIONINFOEXW osviex{};
+	osviex.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEXW);
+	if (::GetVersionEx(reinterpret_cast<LPOSVERSIONINFOW>(&osviex)))
+	{
+		DWORD dwReturnedProductType = 0;
+		if (::GetProductInfo(osviex.dwMajorVersion, osviex.dwMinorVersion, osviex.wServicePackMajor, osviex.wServicePackMinor, &dwReturnedProductType))
+		{
+			switch (dwReturnedProductType)
+			{
+				case PRODUCT_STANDARD_SERVER_CORE:
+				case PRODUCT_STANDARD_A_SERVER_CORE:
+				case PRODUCT_STANDARD_SERVER_CORE_V:
+				case PRODUCT_STANDARD_SERVER_SOLUTIONS_CORE:
+				case PRODUCT_SMALLBUSINESS_SERVER_PREMIUM_CORE:
+				case PRODUCT_ENTERPRISE_SERVER_CORE:
+				case PRODUCT_ENTERPRISE_SERVER_CORE_V:
+				case PRODUCT_DATACENTER_SERVER_CORE:
+				case PRODUCT_DATACENTER_A_SERVER_CORE:
+				case PRODUCT_DATACENTER_SERVER_CORE_V:
+				case PRODUCT_STORAGE_STANDARD_SERVER_CORE:
+				case PRODUCT_STORAGE_WORKGROUP_SERVER_CORE:
+				case PRODUCT_STORAGE_ENTERPRISE_SERVER_CORE:
+				case PRODUCT_STORAGE_EXPRESS_SERVER_CORE:
+				case PRODUCT_WEB_SERVER_CORE:
+					isCoreWindows = true;
+			}
+		}
+	}
+
+	if (!isCoreWindows)
+	{
+		// in Core Server 2012+, the recommended way to determine is via the Registry
+		HKEY hKey = nullptr;
+		if (::RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Server\\ServerLevels",
+			0, KEY_READ, &hKey) == ERROR_SUCCESS)
+		{
+			DWORD dwBuf = 0;
+			DWORD dataSize = sizeof(dwBuf);
+			if (::RegQueryValueExW(hKey, L"ServerCore", nullptr, nullptr, reinterpret_cast<LPBYTE>(&dwBuf), &dataSize) == ERROR_SUCCESS)
+			{
+				if (dwBuf == 1)
+					isCoreWindows = true;
+			}
+			::RegCloseKey(hKey);
+			hKey = nullptr;
+		}
+	}
+
+	return isCoreWindows;
+}
+#pragma warning(default:4996)
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1947,14 +1947,15 @@ bool isCoreWindows()
 	{
 		// in Core Server 2012+, the recommended way to determine is via the Registry
 		HKEY hKey = nullptr;
-		if (::RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Server\\ServerLevels",
+		if (::RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion",
 			0, KEY_READ, &hKey) == ERROR_SUCCESS)
 		{
-			DWORD dwBuf = 0;
-			DWORD dataSize = sizeof(dwBuf);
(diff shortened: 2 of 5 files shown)


donho (changes requested):
Works great! Could you add 3 lines in the PR?

```diff
diff --git a/PowerEditor/src/Parameters.h b/PowerEditor/src/Parameters.h
index 8742c0101..ec594a5e3 100644
--- a/PowerEditor/src/Parameters.h
+++ b/PowerEditor/src/Parameters.h
@@ -116,7 +116,7 @@ enum ChangeDetect { cdDisabled = 0x0, cdEnabledOld = 0x01, cdEnabledNew = 0x02,
 enum BackupFeature {bak_none = 0, bak_simple = 1, bak_verbose = 2};
 enum OpenSaveDirSetting {dir_followCurrent = 0, dir_last = 1, dir_userDef = 2};
 enum MultiInstSetting {monoInst = 0, multiInstOnSession = 1, multiInst = 2};
-enum writeTechnologyEngine {defaultTechnology = 0, directWriteTechnology = 1};
+enum writeTechnologyEngine {defaultTechnology = 0, directWriteTechnology = 1, directWriteTechnologyUnavailable = 2};^M
 enum urlMode {urlDisable = 0, urlNoUnderLineFg, urlUnderLineFg, urlNoUnderLineBg, urlUnderLineBg,
               urlMin = urlDisable,
               urlMax = urlUnderLineBg};
diff --git a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
index b74ecc11b..ea901c2a0 100644
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -318,7 +318,7 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)

        if (isWINE || // There is a performance issue under WINE when DirectWrite is ON, so we turn it off if user uses Notepad++ under WINE
                isCoreWindows()) // In the case of Windows Server Core, DirectWrite cannot be on.
-               nppGui._writeTechnologyEngine = defaultTechnology;
+               nppGui._writeTechnologyEngine = directWriteTechnologyUnavailable;^M

        if (nppGui._writeTechnologyEngine == directWriteTechnology)
        {
diff --git a/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp b/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
index 99d21d240..d6f337dcc 100644
--- a/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
+++ b/PowerEditor/src/WinControls/Preference/preferenceDlg.cpp
@@ -2433,6 +2433,7 @@ intptr_t CALLBACK MiscSubDlg::run_dlgProc(UINT message, WPARAM wParam, LPARAM)
                        ::SendDlgItemMessage(_hSelf, IDC_CHECK_DETECTENCODING, BM_SETCHECK, nppGUI._detectEncoding, 0);
                        ::SendDlgItemMessage(_hSelf, IDC_CHECK_SAVEALLCONFIRM, BM_SETCHECK, nppGUI._saveAllConfirm, 0);
                        ::SendDlgItemMessage(_hSelf, IDC_CHECK_AUTOUPDATE, BM_SETCHECK, nppGUI._autoUpdateOpt._doAutoUpdate, 0);
+                       ::EnableWindow(::GetDlgItem(_hSelf, IDC_CHECK_DIRECTWRITE_ENABLE), nppGUI._writeTechnologyEngine != directWriteTechnologyUnavailable);^M
                        ::SendDlgItemMessage(_hSelf, IDC_CHECK_DIRECTWRITE_ENABLE, BM_SETCHECK, nppGUI._writeTechnologyEngine == directWriteTechnology, 0);
                        ::SendDlgItemMessage(_hSelf, IDC_CHECK_ENABLEDOCPEEKER, BM_SETCHECK, nppGUI._isDocPeekOnTab ? BST_CHECKED : BST_UNCHECKED, 0);
                        ::SendDlgItemMessage(_hSelf, IDC_CHECK_ENABLEDOCPEEKONMAP, BM_SETCHECK, nppGUI._isDocPeekOnMap ? BST_CHECKED : BST_UNCHECKED, 0);
```

xomx:
@donho 
Done.
I added also the possible re-enabling of the `Use DirectWrite` checkbox.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -317,8 +317,16 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
...
-		nppGui._writeTechnologyEngine = defaultTechnology;
+		isCoreWindows()) // In the case of Windows Server Core, DirectWrite cannot be on.
+	{
+		nppGui._writeTechnologyEngine = directWriteTechnologyUnavailable;
+	}
+	else
+	{
+		// allow IDC_CHECK_DIRECTWRITE_ENABLE to be set in Preferences > MISC. again
+		if (nppGui._writeTechnologyEngine == directWriteTechnologyUnavailable)
+			nppGui._writeTechnologyEngine = defaultTechnology;
```
donho, line 328: 👍

donho (approved):

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
donho: Just re-read the code, here nppGui._writeTechnologyEngine will never be directWriteTechnologyUnavailable, because of this line: `_nppGUI._writeTechnologyEngine = (lstrcmp(optNameWriteTechnologyEngine, L"1") == 0) ? directWriteTechnology : defaultTechnology;` in https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/Parameters.cpp#L6189

Nevertheless, we can keep it - it shows more clearly our intention.

xomx: @donho 
Ok, you are right, I missed that.

I only saw that in the config.xml that `writeTechnologyEngine="2"` has been really written there and I imagined that someone might have a portable N++ setup on a usb-flash and once used it in Wine or Core env, that checkbox could have been disabled forever.

xomx:
reported by PR binary as a Core but in reality it is not:
>
> [on Dec 9, 2024](https://github.com/notepad-plus-plus/notepad-plus-plus/commit/6b441b738a951ed225632f4085a46c95cd4aa13f#r150088153)
> 
> It's non-Core:
> 
> ```
> Notepad++ v8.7.4   (64-bit)
> Build time : Dec  8 2024 - 16:02:19
> Path : C:\Program Files\Notepad++\notepad++.exe
> Command Line : 
> Admin mode : ON
> Local Conf mode : OFF
> Cloud Config : OFF
> Periodic Backup : ON
> Placeholders : OFF
> DirectWrite : OFF
> Multi-instance Mode : monoInst
> File Status Auto-Detection : cdEnabledNew (for current file/tab only)
> Dark Mode : OFF
> OS Name : Windows Server 2022 Datacenter (64-bit)
> OS Version : 21H2
> OS Build : 20348.2849
> Current ANSI codepage : 1252
> Plugins : 
>     mimeTools (3.1)
>     NppConverter (4.6)
>     NppExport (0.4)
> ```

@0xfeeddeadbeef

This PR uses the MS recommended way for checking if the Windows installation is a Core one ([Ref1](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ee391629(v=vs.85)), [Ref2](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/hh846315(v=vs.85))). But I see a possibility that I implemented it the wrong way, so I need your help, otherwise I have to install another Windows Server variant in VM for debugging (your reported Windows Server 2022 Datacenter).

Could you please check 1st your Registry for the `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Server\ServerLevels` if there is the key `ServerCore` (with the DWORD value 1)?
E.g. run the Windows Registry Editor:

npp-Registry-ServerLevels

0xfeeddeadbeef:
@xomx All values are set to 1
image
which, according to documentation, is `Full Server`:
image
It is Core only when `ServerCore = 1` **and** others are zero or missing.

0xfeeddeadbeef:
I don't know if this can be of any use, but this is how Microsoft's .NET Runtime unit tests detect Core and Nano:

[PlatformDetection.Windows.cs#L29-L30](https://github.com/dotnet/runtime/blob/v9.0.0/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.Windows.cs#L29-L30)

[PlatformDetection.Windows.cs#L117-L131](https://github.com/dotnet/runtime/blob/v9.0.0/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.Windows.cs#L117-L131)

xomx:
@0xfeeddeadbeef 

> which, according to documentation, is `Full Server`:

That's it, thanks!


> I don't know if this can be of any use, but this is how Microsoft's .NET Runtime unit tests detect Core and Nano:
> 
> [PlatformDetection.Windows.cs#L29-L30](https://github.com/dotnet/runtime/blob/v9.0.0/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.Windows.cs#L29-L30)
> 
> [PlatformDetection.Windows.cs#L117-L131](https://github.com/dotnet/runtime/blob/v9.0.0/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.Windows.cs#L117-L131)

Yes, it is useful, because of it is exactly what the [Ref1](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ee391629(v=vs.85)) says:
`Another way to test if Server Core is running is to check the registry under 
HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\CurrentVersion for the value of InstallationType.
If it is set to "Server Core", then the Server Core installation option is installed.`

I will switch this PR to this much simpler method. My bad that I chose the [Ref2](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/hh846315(v=vs.85)) method implementation before.

---

Do you have any experience with the other options mentioned in the doc?
I mean that the "Server Core" and "Full Server" are now clear to me, but what about the "Nano Server" and "Server Core with Management Tools" - should the N++ DirectWrite mode be disabled there as well, in the same way as it will be in the "Server Core" case?

xomx:
Ok, I've got this "Nano Server", it is more restricted than the Core ones, no GUI-apps == N++ does not have to worry about anything at all because it will not be able to run:

"Windows Server Core includes a command-line interface and limited support for local graphical management tools. Nano Server is designed to be managed remotely with no local graphical user interface."

So the only remaining question is the "Server Core with Management Tools" DirectWrite support.

0xfeeddeadbeef:
> Do you have any experience with the other options mentioned in the doc?

I've never worked with Core. The doc says "This configuration is not supported starting with Windows Server 2016" — IMHO, it is safe to consider it as "not having DirectWrite support".

xomx:
@0xfeeddeadbeef  @mbaker-e2open

Could you re-test with the PR latest notepad++.exe x64 binary?
https://github.com/notepad-plus-plus/notepad-plus-plus/actions/runs/12242808602/artifacts/2295825250

This one should allow you to use the DirectWrite on non-Core servers.

0xfeeddeadbeef:
@xomx Debug Info says `DirectWrite : ON`
Also, I can check/uncheck it in the settings UI.
Thanks.

xomx:
@donho 
I fixed the Core-detection method by switching to a simpler one mentioned in the MS doc.

One user above confirmed that the latest PR binary no longer marks a non-Core Windows Server installation as the Core-one.

And I checked it in my Core-Windows VM that it is detected as the Core-one:
npp-CoreServer-DirectWrite-disabled

From the pic above you can see that I have different problem there - tab pin-button is visible whereas the close-button not (?)

donho:
@xomx 
I'm traveling and the laptop with me has no core server VM. I'll test it when I get home (in about 2 months).
Also the close button display issue can be fixed at that time.
Thank you for the fix.

xomx:
@donho 
Enjoy your travel, even if it is just for work and not a vacation!

Now I noticed that your [testing pic](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15916#issuecomment-2525270722) in the issue does not exhibit that problem, so this is probably something local in my VM installation.

Issue #5089, opened 2018-11-30
Feature request: search text in a scope


ferenczy:
It would be great to have a **possibility to limit text search to a specific scope**. A scope would be a language element like comment, keyword, string, etc.

Notepad++ already knows about those scopes - they correspond to language elements which syntax highlighter is using to colorize the text. Any element the syntax highlighter knows might be used as a scope. Obviously, it depends on the currently active language.

The list of scopes in the _Find text_ dialog would be dynamic, it would simply display elements defined by the currently active language (you can see them in the _Style_ list box in the _Style Configurator_ dialog). Only some of those elements would make sense, for example comment, keyword, identifier, string, etc. Other elements like number, operator, etc. wouldn't be very useful but it's not a big deal I guess.

So it's just a new usage of something Notepad++ already has (language elements). Some editors have such feature.

What do you think? I would appreciate additional thoughts on this.

MetaChuh:
unfortunately we currently don't have time to do that, hence also the delayed reply.
thanks for understanding and best regards.

Issue #5098, opened 2018-12-03
Option to save session with relative pathnames for current drive


peripatetes:
### Description of the Issue

When notepad++ is used on a portable drive, it would be a great benefit to make all session references to files on the same drive relative rather than absolute paths. This is perhaps an extension of issue #1587.

### Steps to Reproduce the Issue

1. Use portable Notepad++ on a removable drive. Save some files on same removable drive.
2. Close Notepad++, saving session and backups.
3. Do something which causes the drive to have another drive letter - move to another computer, insert another removable drive, etc.
4. Launch Notepad++ again.

### Expected Behavior

Notepad++ launches with saved session, including all the previously open files on the same removable drive.

### Actual Behavior

Notepad++ finds that it can't locate the files from the saved session, and opens an empty session instead.

If you have sufficient user privileges and nothing open on any conflicting disk, you can work around this by going into Disk Management and changing drive letters around - but only if you remember to do that before launching Notepad++. If you launch Notepad++ first and then remember you needed to do that, it's too late - the saved session file has been overwritten with a blank one.

badrelmers:
this is a must,please add it.

donho:
This feature won't be implemented due to the huge impact with its all functionality.

Issue #1587, opened 2016-03-04
Bug] Unsaved documents lost on next launch if portable npp change its path 


Tragen:
Currently the default backup path for the session is the
N+ folder\backup
This is a fixed path in the settings. But when I keep N++ portable on y USB stick, this doesn't work.
It would be great to have a variable or an option to make the path relative to the current N++ executable

ClaudiaFrank:
Temporary [workaround](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/1344) should
work here as well

PR #15928, opened 2024-12-09
Add missing backup recovery functionality

### LINKED ISSUE #15886: [BUG] session.xml backup paths are stored as absolute paths but the backups themselves are stored relative to npp's folder which breaks session.xml if npp's folder were renamed or moved

- [x] I have searched the existing issues
Related: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/5098

### Description of the Issue

So, session.xml backup paths are stored as absolute paths but the backups themselves are stored relative to npp's folder which breaks session.xml if npp being launched after its folder was renamed or moved. This just happens to me between windows reinstallations as I use it in portable mode.
It can also break absolutely easy if npp is used from flash drive and windows just assign another letter to the drive

### Steps To Reproduce

1. unpack portable npp
2. start npp
3. make new file
4. edit it
5. close npp
6. check session.xml that it have absolute path to new file's backup
7. rename or move npp's folder
8. start npp
9. you lost your unsaved edited new file

### Current Behavior

as described in steps 7~9

### Expected Behavior

as backups stored relative to npp's executable folder, session.xml should also have paths that are relative to npp's executable.

Current: `backupFilePath="D:\Users\Rustam\Desktop\0Notepad++\backup\new 5@2024-11-30_165530"`
Expected: `backupFilePath="backup\new 5@2024-11-30_165530"`

### Debug Information

```shell
Notepad++ v8.7.1   (64-bit)
Build time : Oct 31 2024 - 00:48:56
Path : D:\Users\kww\Desktop\0Notepad++\notepad++.exe
Command Line : 
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 10 Enterprise LTSC 2021 (64-bit)
OS Version : 21H2
OS Build : 19044.1288
Current ANSI codepage : 1251
Plugins : 
    ComparePlugin (2.0.2)
    EmmetNPP (1.0.2)
    HexEditor (0.9.12)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    NPPJSONViewer (1.40)
    PythonScript (2)
```

### Anything else?

_No response_

Trubisky:
Fixes #15886 and also #5089 by providing an auto-recover attempt for missing backup files relative to the CWD. The patch causes NppIO.cp to attempt to find and load a backup file from currentWorkingDirectioy/backup/[file name] if the session it's trying to load has a backup file path as part of it, but no backup file exists at the expected file path.

From my testing, seems to work as expected without modifying or breaking the functionality of anything else.

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -37,20 +37,20 @@ using namespace std;
 //  ("tab" is not in the official list, but it is good to avoid it)
 const std::wstring filenameReservedChars = L"<>:\"/\\|\?*\t";
 
-DWORD WINAPI Notepad_plus::monitorFileOnChange(void* params)
+DWORD WINAPI Notepad_plus::monitorFileOnChange(void * params)
 {
-	MonitorInfo* monitorInfo = static_cast<MonitorInfo*>(params);
-	Buffer* buf = monitorInfo->_buffer;
+	MonitorInfo *monitorInfo = static_cast<MonitorInfo *>(params);
+	Buffer *buf = monitorInfo->_buffer;
 	HWND h = monitorInfo->_nppHandle;
 
-	const wchar_t* fullFileName = (const wchar_t*)buf->getFullPathName();
+	const wchar_t *fullFileName = (const wchar_t *)buf->getFullPathName();
 
 	//The folder to watch :
 	wchar_t folderToMonitor[MAX_PATH]{};
 	wcscpy_s(folderToMonitor, fullFileName);
 
 	::PathRemoveFileSpecW(folderToMonitor);
-
+	
 	const DWORD dwNotificationFlags = FILE_NOTIFY_CHANGE_LAST_WRITE | FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_SIZE;
 
 	// Create the monitor and add directory to watch.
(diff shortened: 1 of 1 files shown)


Trubisky:
My original test commit pushed an older version of the NppIO file, and the second commit resolves that to be newest and include added functionality.

alankilborn:
@Trubisky 

> and also 5089

??

PowerEditor/src/NppIO.cpp
```diff
@@ -2399,9 +2399,36 @@ bool Notepad_plus::loadSession(Session & session, bool isSnapshotMode, const wch
...
 		else
 		{
-			BufferID foundBufID = MainFileManager.getBufferFromName(pFn);
-			if (foundBufID == BUFFER_INVALID)
-				lastOpened = nppGUI._keepSessionAbsentFileEntries ? MainFileManager.newPlaceholderDocument(pFn, MAIN_VIEW, userCreatedSessionName) : BUFFER_INVALID;
+			bool recoveredBackup = false;
+			//If it's in snapshot mode, there's a backupFilePath present, but the existing path doesn't exist
+			if (isSnapshotMode && !session._mainViewFiles[i]._backupFilePath.empty())
+			{
+				//Jusko was here, December 2024 <3
```
alankilborn, line 2406: I don't think such a comment is going to be acceptable.

xomx: No offense, but this is just so similar to a doggy does when it marks its spots outside...  :-)

donho: @xomx Could you provide the link please?

xomx: @donho 
I meaned the same thing (`//Jusko was here, December 2024 <3`) as @alankilborn pointed out.
It just seems ridiculously similar to what dogs do outside when they want to let other dogs know they have been there... :-)

Anyway, due to the extensive reformatting (right one, though) in this PR, it took me a while to find a place with any relevant changes to the code.

rust2:
Does this work if 

> @Trubisky
> 
> > and also 5089
> 
> ??

see #5098

PowerEditor/src/NppIO.cpp
Trubisky: No offense taken. Is this PR okay as is, or would the two of you rather I revise it without the commentary and with only relevant code changes (and not the reformatting)?

Trubisky:
Yeah, 5098. Also, #1587

PowerEditor/src/NppIO.cpp
donho: > //Jusko was here, December 2024 <3

If I allow it, Notepad++ source code will be a graffiti wall.

donho:
> //Jusko was here, December 2024 <3

If I allowed it, Notepad++ source code would be a graffiti wall.
PR rejected.

alankilborn:
>//Jusko was here, December 2024 <3

I'm not even sure how somebody in the act of typing that would think it was a good idea.  Wowza.

PR #15931, opened 2024-12-10
[Xml] Update brazilian_portuguese.xml

Marcellomco:
Added translation for "Fix the localization files to match the new behaviour"


Marcellomco (commented):
Updated pre-existing PR with  "Remove the obsolete entry from localization files"

PR #15935, opened 2024-12-10
Prevent backup worker thread crash at npp exit

xomx:
Fix: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15681#issuecomment-2403168094

This patch addresses uncontrolled termination of the backup worker thread at the very Notepad++ app-exit. This could previously lead to:
- race condition crash, when the backup thread has been awaken right after the Notepad++ `wWinMain` exit but before it was terminated in the CRT internals
- possible unsafe file I/O (e.g. in the `nppLogNulContentCorruptionIssue.log` could be sometimes seen that the backup thread was still running & doing its job even in the very last stages of the Notepad++ app existence, which is not completely safe...)

There is also small code optimization of the `Notepad_plus::backupDocument` worker thread body.

This PR is a simpler variant of the previously closed #15031 .

### DIFF
--- a/PowerEditor/src/Notepad_plus.cpp
+++ b/PowerEditor/src/Notepad_plus.cpp
@@ -8653,23 +8653,22 @@ void Notepad_plus::launchDocumentBackupTask()
 
 DWORD WINAPI Notepad_plus::backupDocument(void * /*param*/)
 {
-	bool isSnapshotMode = true;
-	while (isSnapshotMode)
-	{
-		NppParameters& nppParam = NppParameters::getInstance();
+	NppGUI& nppGUI = (NppParameters::getInstance()).getNppGUI();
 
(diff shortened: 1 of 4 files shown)


PowerEditor/src/Notepad_plus.cpp
```diff
@@ -8653,23 +8653,22 @@ void Notepad_plus::launchDocumentBackupTask()
 
 DWORD WINAPI Notepad_plus::backupDocument(void * /*param*/)
 {
-	bool isSnapshotMode = true;
-	while (isSnapshotMode)
-	{
-		NppParameters& nppParam = NppParameters::getInstance();
+	NppGUI& nppGUI = (NppParameters::getInstance()).getNppGUI();
```
donho, line 8656: 👍

PR #15936, opened 2024-12-10
Fix v8.7.1+ network file modification detection

### LINKED ISSUE #15819: [BUG] 8.7.1 introduced a network file modification detection bug.

xomx:
Fix #15819 .

- increased `DEFAULT_MILLISEC` constant from 1000 to 3000 (some slower networks obviously cannot respond within the previous 1sec timeout)

- slightly rewritten the `Buffer::checkFileState()`, as the problematic situation was made worse by the fact that the `Buffer::checkFileState()` did not issue one GetFileAttributesEx call but also an unnecessary second one (this was probably caused by an inappropriate literal rewriting of the previous version of this func for using the new non-blocking thread method of calling the GetFileAttributesEx)

- other `Buffer::checkFileState()` problem was that it did not check at all whether the threaded call to the GetFileAttributesEx succeeded or had to be forcibly terminated from the main app thread and expected successful use only

- the `nppLogNetworkDriveIssue` logging has been enhanced in the `Buffer::checkFileState()`, `Buffer::updateTimeStamp()` and `NPPM_INTERNAL_CHECKDOCSTATUS` message handler

- added option to check possible WIN32API error code of the threaded GetFileAttributesEx call from the main N++ thread

### DIFF
--- a/PowerEditor/src/MISC/Common/Common.cpp
+++ b/PowerEditor/src/MISC/Common/Common.cpp
@@ -1751,7 +1751,7 @@ bool Version::isCompatibleTo(const Version& from, const Version& to) const
 }
 
 
-#define DEFAULT_MILLISEC 1000
+#define DEFAULT_MILLISEC 3000
 
 
 //----------------------------------------------------
@@ -1816,6 +1816,7 @@ struct GetAttrExParamResult
 	wstring _filePath;
 	WIN32_FILE_ATTRIBUTE_DATA _attributes{};
 	BOOL _result = FALSE;
+	DWORD _error = NO_ERROR;
 	bool _isTimeoutReached = true;
 
 	GetAttrExParamResult(wstring filePath): _filePath(filePath) {
@@ -1826,12 +1827,16 @@ struct GetAttrExParamResult
 DWORD WINAPI getFileAttributesExWorker(void* data)
 {
 	GetAttrExParamResult* inAndOut = static_cast<GetAttrExParamResult*>(data);
+	::SetLastError(NO_ERROR);
 	inAndOut->_result = ::GetFileAttributesExW(inAndOut->_filePath.c_str(), GetFileExInfoStandard, &(inAndOut->_attributes));
+	if (!(inAndOut->_result))
+		inAndOut->_error = ::GetLastError();
 	inAndOut->_isTimeoutReached = false;
 	return ERROR_SUCCESS;
 };
 
-BOOL getFileAttributesExWithTimeout(const wchar_t* filePath, WIN32_FILE_ATTRIBUTE_DATA* fileAttr, DWORD milliSec2wait, bool* isTimeoutReached)
+BOOL getFileAttributesExWithTimeout(const wchar_t* filePath, WIN32_FILE_ATTRIBUTE_DATA* fileAttr,
+	DWORD milliSec2wait, bool* isTimeoutReached, DWORD* pdwWin32ApiError)
 {
 	GetAttrExParamResult data(filePath);
 
@@ -1855,12 +1860,15 @@ BOOL getFileAttributesExWithTimeout(const wchar_t* filePath, WIN32_FILE_ATTRIBUT
 			::TerminateThread(hThread, dwWaitStatus);
 			break;
 	}
-	CloseHandle(hThread);
+	::CloseHandle(hThread);
 
 	*fileAttr = data._attributes;
 
 	if (isTimeoutReached != nullptr)
 		*isTimeoutReached = data._isTimeoutReached;
+
+	if (pdwWin32ApiError != nullptr)
+		*pdwWin32ApiError = data._error;
 
 	return data._result;
 }
--- a/PowerEditor/src/ScintillaComponent/Buffer.cpp
+++ b/PowerEditor/src/ScintillaComponent/Buffer.cpp
@@ -142,11 +142,48 @@ void Buffer::setLangType(LangType lang, const wchar_t* userLangName)
 void Buffer::updateTimeStamp()
 {
 	FILETIME timeStampLive {};
+
 	WIN32_FILE_ATTRIBUTE_DATA attributes{};
 	attributes.dwFileAttributes = INVALID_FILE_ATTRIBUTES;
-	if (getFileAttributesExWithTimeout(_fullPathName.c_str(), &attributes))
+	bool bWorkerThreadTerminated = true;
+	DWORD dwWin32ApiError = NO_ERROR;
+	BOOL bGetFileAttributesExSucceeded = getFileAttributesExWithTimeout(_fullPathName.c_str(), &attributes, 0, &bWorkerThreadTerminated, &dwWin32ApiError);
+	if (bGetFileAttributesExSucceeded && (attributes.dwFileAttributes != INVALID_FILE_ATTRIBUTES) && !(attributes.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY))
 	{
 		timeStampLive = attributes.ftLastWriteTime;
+	}
+	else
+	{
+		if (_currentStatus != DOC_UNNAMED)
+		{
+			NppParameters& nppParam = NppParameters::getInstance();
+			if (nppParam.doNppLogNetworkDriveIssue())
+			{
+				wstring issueFn = nppLogNetworkDriveIssue;
+				issueFn += L".log";
+				wstring nppIssueLog = nppParam.getUserPath();
+				pathAppend(nppIssueLog, issueFn);
+				std::string msg = wstring2string(_fullPathName, CP_UTF8);
+				msg += "  in Buffer::updateTimeStamp(), getFileAttributesExWithTimeout returned ";
+				if (bGetFileAttributesExSucceeded)
+					msg += "TRUE";
+				else
+					msg += "FALSE";
+				if (bWorkerThreadTerminated)
+				{
+					msg += ", its worker thread had to be forcefully terminated due to timeout reached!";
+				}
+				else
+				{
+					msg += ", its worker thread finished successfully within the timeout given, ";
+					if (attributes.dwFileAttributes == INVALID_FILE_ATTRIBUTES)
+						msg += "dwFileAttributes == INVALID_FILE_ATTRIBUTES ! (WIN32API Error Code: " + std::to_string(dwWin32ApiError) + ")";
+					else
+						msg += "dwFileAttributes has the FILE_ATTRIBUTE_DIRECTORY flag set!";
+				}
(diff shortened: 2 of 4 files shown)


PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -324,30 +399,39 @@ bool Buffer::checkFileState() // returns true if the status has been changed (it
...
+				if (res == 1)
+				{
+					sprintf_s(buf, _countof(buf) - 1, "  in checkFileState(): attributes.ftLastWriteTime (%lu/%lu) < _timeStamp (%lu/%lu)",
+						attributes.ftLastWriteTime.dwLowDateTime, attributes.ftLastWriteTime.dwHighDateTime, _timeStamp.dwLowDateTime, _timeStamp.dwHighDateTime);
+				}
+				else
+				{
+					// usual state after a file modification, uncomment if needed to log too
+					//sprintf_s(buf, _countof(buf) - 1, "  in checkFileState(): attributes.ftLastWriteTime (%lu/%lu) > _timeStamp (%lu/%lu)",
+					//	attributes.ftLastWriteTime.dwLowDateTime, attributes.ftLastWriteTime.dwHighDateTime, _timeStamp.dwLowDateTime, _timeStamp.dwHighDateTime);
```
donho, line 421: Is it necessary to keep the commented code?

donho:
@xomx 
It's mainly log in this PR. Could you indicate what's the fix?

PowerEditor/src/ScintillaComponent/Buffer.cpp
xomx: @xomx 
This probably will not be needed normally, but right now I needed it to solve the reported problems twice in a row, so I left the debugging code there for a next possible case.

xomx:
@donho 
> It's mainly log in this PR. Could you indicate what's the fix?

OK, the main stuff is here:

> * increased `DEFAULT_MILLISEC` constant from 1000 to 3000 (some slower networks obviously cannot respond within the previous 1sec timeout)

https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15936/commits/2211f701ff27362491d1524282724c2d1178ea0b#diff-b24b3194baf7cdfb71dcaba855cf7a9e228d1a332404f056959d33736b119052L1754


> * slightly rewritten the `Buffer::checkFileState()`, as the problematic situation was made worse by the fact that the `Buffer::checkFileState()` did not issue one GetFileAttributesEx call but also an unnecessary second one
> * other `Buffer::checkFileState()` problem was that it did not check at all whether the threaded call to the GetFileAttributesEx succeeded or had to be forcibly terminated from the main app thread

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2211f701ff27362491d1524282724c2d1178ea0b#diff-2d5c277be2799a7e43a90232d8ef9add2e788e252c8692bf9416abc69836957dL266-R316

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2211f701ff27362491d1524282724c2d1178ea0b#diff-2d5c277be2799a7e43a90232d8ef9add2e788e252c8692bf9416abc69836957dL301-R380

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/2211f701ff27362491d1524282724c2d1178ea0b#diff-2d5c277be2799a7e43a90232d8ef9add2e788e252c8692bf9416abc69836957dL316-R391

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -260,10 +306,43 @@ bool Buffer::checkFileState() // returns true if the status has been changed (it
...
+			{
+				msg += ", its worker thread finished successfully within the timeout given, ";
+				if (attributes.dwFileAttributes == INVALID_FILE_ATTRIBUTES)
+					msg += "dwFileAttributes == INVALID_FILE_ATTRIBUTES ! (WIN32API Error Code: " + std::to_string(dwWin32ApiError) + ")";
+				else
+					msg += "dwFileAttributes has the FILE_ATTRIBUTE_DIRECTORY flag set!";
+			}
+			writeLog(nppIssueLog.c_str(), msg.c_str());
+		}
+	}
```
donho, line 345: Use rather:
```cpp
if (!fileExists && nppParam.doNppLogNetworkDriveIssue())
{
    // log here
}
```

Also, this log block should be on line 278/357 (after `#ifndef	_WIN64`), instead being here.

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -291,29 +370,25 @@ bool Buffer::checkFileState() // returns true if the status has been changed (it
...
-		_isDirty = true;	//dirty sicne no match with filesystem
+		_isDirty = true;	//dirty since no match with filesystem
 		_timeStamp = {};
 		doNotify(BufferChangeStatus | BufferChangeReadonly | BufferChangeTimestamp);
 		isOK = true;
 	}
 	else if (_currentStatus == DOC_DELETED && fileExists) //document has returned from its grave
 	{
-		if (GetFileAttributesEx(_fullPathName.c_str(), GetFileExInfoStandard, &attributes) != 0) // fileExists so it's safe to call GetFileAttributesEx directly
+		// fileExists==true here means that we can use the previously obtained attributes safely
```
donho, line 380: 👍

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -291,29 +370,25 @@ bool Buffer::checkFileState() // returns true if the status has been changed (it
...
-				_reloadFromDiskRequestGuard.unlock();
-			}
-			isOK = true;
+			doNotify(BufferChangeStatus | BufferChangeReadonly | BufferChangeTimestamp);
+			_reloadFromDiskRequestGuard.unlock();
 		}
+		isOK = true;
 	}
-	else if (getFileAttributesExWithTimeout(_fullPathName.c_str(), &attributes))
+	else if (bGetFileAttributesExSucceeded)
```
donho, line 391: 👍

donho (changes requested):
Please consider the review remark.

PowerEditor/src/MISC/Common/Common.cpp
```diff
@@ -1826,12 +1827,16 @@ struct GetAttrExParamResult
 DWORD WINAPI getFileAttributesExWorker(void* data)
 {
 	GetAttrExParamResult* inAndOut = static_cast<GetAttrExParamResult*>(data);
+	::SetLastError(NO_ERROR);
```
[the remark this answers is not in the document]
xomx: @donho 

No, it is ok, the WINAPI error codes are per-thread ([Ref.](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-setlasterror)): `"The last-error code is kept in thread local storage so that multiple threads do not overwrite each other's values."`

And resetting before the GetFileAttributesEx calling is necessary because of `"Most (WINAPI) functions call SetLastError(Ex) ONLY when they fail. However, some system functions call SetLastError(Ex) under conditions of success;"`. So without the reset, it can contain a misleading previously set non-zero value even after a successful GetFileAttributesEx call...

PowerEditor/src/ScintillaComponent/Buffer.cpp
```diff
@@ -142,12 +142,49 @@ void Buffer::setLangType(LangType lang, const wchar_t* userLangName)
...
+					msg += ", its worker thread finished successfully within the timeout given, ";
+					if (attributes.dwFileAttributes == INVALID_FILE_ATTRIBUTES)
+						msg += "dwFileAttributes == INVALID_FILE_ATTRIBUTES ! (WIN32API Error Code: " + std::to_string(dwWin32ApiError) + ")";
+					else
+						msg += "dwFileAttributes has the FILE_ATTRIBUTE_DIRECTORY flag set!";
+				}
+				writeLog(nppIssueLog.c_str(), msg.c_str());
+			}
+		}
+	}
```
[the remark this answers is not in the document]
xomx: @donho 
You should tell me, IDK :-) 
This needs more debugging. As I did not change the code logic there, only better logging, we can now let it be as it is and if one of us will have a positive answer to your question, we can change it in another PR.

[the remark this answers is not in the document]
donho: @xomx 
OK, let's keep it as it is (your modification) now. 
Once the PR is merged into master, I will do another PR to add return part.

donho:
@xomx 
So please 
1. Remove log followeing the code : 
```cpp
			Buffer* currBuf = getCurrentBuffer();
			if (currBuf && currBuf->isMonitoringOn())
				doesFileExist(currBuf->getFullPathName());
```
2. use `if (!fileExists && nppParam.doNppLogNetworkDriveIssue())`

donho (changes requested):
Please consider:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15936#issuecomment-2547452331

xomx:
@donho 
Done.

Replaced also one remaining `doesFileExist` by `getFileAttributesExWithTimeout` in `Buffer::checkFileState()` `#ifndef _WIN64` code branch.

donho:
@xomx 
Just want to make sure: though the enhancement of this PR, #15819 is not (yet) fixed by this PR, right?

xomx:
@donho 

> #15819 is not (yet) fixed by this PR, right?

It should be fixed.
If you mean the last posts in that issue from igtbap, from the [desc](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15819#issuecomment-2533819359) his issue is probably not the same (but IDK, there seems to be a language barrier(?)).

I have also registered another user report which could be connected to this v8.7.1+ too short threaded GetFileAttributesEx call timeout: https://community.notepad-plus-plus.org/topic/26462/file-does-not-exist

donho (approved):

PR #15981, opened 2024-12-24
Enhance large files with syntax highlighting performance

### LINKED ISSUE #15952: [Feature request] Enhance large files with syntax highlighting performance

### Description of the Issue

**Steps to Reproduce Performance Issues in Notepad++:**
1. Open a large file (e.g., an HTML file larger than 100MB) in Notepad++ with syntax highlighting enabled.
2. Press Ctrl + End. At this point, Notepad++ becomes unresponsive and takes a significant amount of time to recover.
3. Reopen the large file, press Ctrl + G, input a very large line number, and press Enter. Again, Notepad++ becomes unresponsive and takes a long time to respond.
4. Move the cursor to the end of the large file. Then, create a new file or open another file. Finally, switch back to the tab containing the large file. Notepad++ takes a significant amount of time to respond.

**Fix Description:**
@donho I have optimized the behavior for the above scenarios. The attached source code includes my modifications, based on the latest master branch as of today.

[notepad-plus-plus-master[20241215].zip](https://github.com/user-attachments/files/18140068/notepad-plus-plus-master.20241215.zip)

I have optimized the behavior for the above scenarios. 

### Debug Information

```shell
Notepad++ v8.7.4   (64-bit)
Build time : Dec  4 2024 - 23:50:05
Command Line : 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
Placeholders : OFF
DirectWrite : OFF
Dark Mode : ON
OS Version : 22H2
OS Build : 19045.5131
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
```

### Anything else?

**Relevant links:**
https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15653
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15926

donho:
Fix #15952

### DIFF
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2260,29 +2260,7 @@ void ScintillaEditView::restyleBuffer()
 
 void ScintillaEditView::styleChange()
 {
-	const bool isSameLangType = _prevBuffer != nullptr && ((_prevBuffer == _currentBuffer) || (_prevBuffer->getLangType() == _currentBuffer->getLangType()));
-	const int currentLangInt = static_cast<int>(_currentBuffer->getLangType());
-	const bool isFirstActiveBuffer = (_currentBuffer->getLastLangType() != currentLangInt) || (_currentBuffer->isUntitled());
-
-	if (isFirstActiveBuffer)
-	{
-		defineDocType(_currentBuffer->getLangType());
-	}
-	else if (!isSameLangType)  // When entering the tab for the second or more times
-	{
-		Document prevDoc = execute(SCI_GETDOCPOINTER);
-		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
-		execute(SCI_SETDOCPOINTER, 0, getBlankDocument());
-		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
-
-		defineDocType(_currentBuffer->getLangType());
-
-		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
-		execute(SCI_SETDOCPOINTER, 0, prevDoc);
-		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
-	}
-
-	_currentBuffer->setLastLangType(currentLangInt);
+	defineDocType(_currentBuffer->getLangType());
 	restyleBuffer();
 }
 
@@ -2323,9 +2301,9 @@ void ScintillaEditView::activateBuffer(BufferID buffer, bool force)
 
 	const bool isSameLangType = _prevBuffer != nullptr && ((_prevBuffer == _currentBuffer) || (_prevBuffer->getLangType() == _currentBuffer->getLangType()));
 	const int currentLangInt = static_cast<int>(_currentBuffer->getLangType());
-	const bool isFirstActiveBuffer = (_currentBuffer->getLastLangType() != currentLangInt) || (_currentBuffer->isUntitled());
+	const bool isFirstActiveBuffer = (_currentBuffer->getLastLangType() != currentLangInt);
 
-	if (isFirstActiveBuffer)  // Entering the tab for the first time
+	if (isFirstActiveBuffer)  // Entering the tab for the 1st time
 	{
 		// change the doc, this operation will decrease
 		// the ref count of old current doc and increase the one of the new doc. FileManager should manage the rest
@@ -2338,8 +2316,17 @@ void ScintillaEditView::activateBuffer(BufferID buffer, bool force)
 		// defineDocType() function should be called here, but not be after the fold info loop
 		defineDocType(_currentBuffer->getLangType());
 	}
-	else if (!isSameLangType) // When entering the tab for the second or more times
+	else if (isSameLangType) // After the 2nd entering with the same language type
 	{
+		// No need to call defineDocType() since it's the same language type
+		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
+		execute(SCI_SETDOCPOINTER, 0, _currentBuffer->getDocument());
+		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
+	}
+	else // Entering the tab for the 2nd or more times, with the different language type
+	{
+		// In order to improve the performance of switch-in on the 2nd or more times for the large files,
+		// a blank document is used for accelerate defineDocType() call.
 		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
 		execute(SCI_SETDOCPOINTER, 0, getBlankDocument());
 		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -230,6 +230,8 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 		throw std::runtime_error("ScintillaEditView::init : SCI_GETDIRECTPOINTER message failed");
 	}
 
+	execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
+	execute(SCI_SETIDLESTYLING, SC_IDLESTYLING_ALL, 0);
 	execute(SCI_SETMARGINMASKN, _SC_MARGE_FOLDER, SC_MASK_FOLDERS);
 	showMargin(_SC_MARGE_FOLDER, true);
 
@@ -345,6 +347,7 @@ void ScintillaEditView::init(HINSTANCE hInst, HWND hPere)
 			delete[] defaultCharList;
 		}
 	}
+	execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
 	//Get the startup document and make a buffer for it so it can be accessed like a file
 	attachDefaultDoc();
 }
@@ -2109,6 +2112,9 @@ void ScintillaEditView::defineDocType(LangType typeDoc)
 		if (currentIndentMode != docIndentMode)
 			execute(SCI_SETINDENTATIONGUIDES, docIndentMode);
 	}
+
+	execute(SCI_SETLAYOUTCACHE, SC_CACHE_DOCUMENT, 0);
+	execute(SCI_STARTSTYLING, 0, 0);
 }
 
 BufferID ScintillaEditView::attachDefaultDoc()
@@ -2283,7 +2289,9 @@ void ScintillaEditView::activateBuffer(BufferID buffer, bool force)
 	// change the doc, this operation will decrease
 	// the ref count of old current doc and increase the one of the new doc. FileManager should manage the rest
 	// Note that the actual reference in the Buffer itself is NOT decreased, Notepad_plus does that if neccessary
+	execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
 	execute(SCI_SETDOCPOINTER, 0, _currentBuffer->getDocument());
+	execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
 
 	// Due to execute(SCI_CLEARDOCUMENTSTYLE); in defineDocType() function
 	// defineDocType() function should be called here, but not be after the fold info loop
--- a/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
+++ b/PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
@@ -2117,6 +2117,16 @@ void ScintillaEditView::defineDocType(LangType typeDoc)
 	execute(SCI_STARTSTYLING, 0, 0);
 }
 
+Document ScintillaEditView::getBlankDocument()
+{
+	if(_blankDocument==0)
+	{
+		_blankDocument=static_cast<Document>(execute(SCI_CREATEDOCUMENT,0,SC_DOCUMENTOPTION_TEXT_LARGE));
+		execute(SCI_ADDREFDOCUMENT,0,_blankDocument);
+	}
+	return _blankDocument;
+}
+
 BufferID ScintillaEditView::attachDefaultDoc()
 {
 	// get the doc pointer attached (by default) on the view Scintilla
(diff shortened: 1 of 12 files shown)


donho:
@softmgr 
Please review the 1 commit. I will break down #15952 into several commits.

softmgr:
> @softmgr 
> 
> Please review the 1 commit. For preventing the regressions, I will break down #15952 into several commits. 

All these modifications must be applied. If only some of the modifications are applied, crashes may occur when switching files or performing similar operations. In fact, I have been using the program generated with the modified code for several months now, and no significant issues have been observed. (I have not tested most third-party plugins, so there might be occasional issues with some plugins due to many notification messages being suppressed.)

donho:
@softmgr 
Thank you for the review. Sorry for the quite long progression for this PR - I'm in a trip.

Here's my question about your change: Why we need set document on a BlankDocument then set back to the current document?
What's  BlankDocument really useful?

softmgr:
@donho 

Background:
The original logic of Notepad++ involves invoking the `defineDocType()` function to initialize syntax highlighting after opening a document. However, when switching between documents, the `defineDocType()` function is called repeatedly. This can lead to unresponsiveness when working with large files.

Through testing, I discovered that if two documents have the same language type, it is possible to skip re-executing the defineDocType() function during document switching. For example:
 • When switching from a large file to a new document with the same language type, executing defineDocType() only on the new document, and then switching back to the large file without re-executing defineDocType() still works correctly. The document-switching process remains smooth.

This observation allows us to optimize the function calls:
 1. Initial load of a large file: Follow the original Notepad++ logic and execute defineDocType().
 2. Switching back to a large file:
 • First, execute the `SCI_SETDOCPOINTER` command to switch to a blank document by calling the `getBlankDocument()` function.
 • Then, perform operations like `defineDocType()` on this blank document.
 • After completing these operations, execute the `SCI_SETDOCPOINTER` command again to switch back to the original document using `_currentBuffer->getDocument()`.
 • Continue executing other functions to complete the document switch efficiently.
 • Use `execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);` to suppress event messages during the process, as message triggers during this time can cause Notepad++ to crash.
 3. Set `MODEVENTMASK_ON` with `SC_MOD_INSERTTEXT | SC_MOD_DELETETEXT` only: Adding additional flags generates excessive and unnecessary event messages, degrading performance. For instance, in the `NppNotification.cpp`, the `notify(SCNotification *notification)` function includes the following code:
```
if (notification->modificationType & SC_MOD_CHANGEINDICATOR)  
{  
    ::InvalidateRect(notifyView->getHSelf(), NULL, FALSE);  
}  
```
Based on my tests, removing this call does not impact the `Mark Style` functionality. However, it increases unnecessary UI refresh actions during operations.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2250,7 +2260,29 @@ void ScintillaEditView::restyleBuffer()
 
 void ScintillaEditView::styleChange()
 {
-	defineDocType(_currentBuffer->getLangType());
+	const bool isSameLangType = _prevBuffer != nullptr && ((_prevBuffer == _currentBuffer) || (_prevBuffer->getLangType() == _currentBuffer->getLangType()));
+	const int currentLangInt = static_cast<int>(_currentBuffer->getLangType());
+	const bool isFirstActiveBuffer = (_currentBuffer->getLastLangType() != currentLangInt) || (_currentBuffer->isUntitled());
+
```
donho, line 2266: @softmgr 
Why `|| (_currentBuffer->isUntitled()` ?
If we copy all data from a large XML file, paste it to an untitled new document and set it to XML - switch off then switch back in it - there will be the performance issue.

donho:
@softmgr 
Thank you for very detailed explanation.
I just did the 2nd commit with some refactoring to simplify the logic, please review it.

> removing this call does not impact the Mark Style functionality. However, it increases unnecessary UI refresh actions during operations.

I'll check the history, and delete the code (of your suggestion to remove) in another PR.

> Set MODEVENTMASK_ON with SC_MOD_INSERTTEXT | SC_MOD_DELETETEXT only

undo & redo notification should be monitoring, see:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppNotification.cpp#L59

softmgr:
> > Set MODEVENTMASK_ON with SC_MOD_INSERTTEXT | SC_MOD_DELETETEXT only使用 SC_MOD_INSERTTEXT 设置 MODEVENTMASK_ON |仅限 SC_MOD_DELETETEXT
> 
> undo & redo notification should be monitoring, see:撤消和重做通知应该受到监控，请参阅： https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppNotification.cpp#L59

The notification messages for undo and redo actions only execute `_pEditView->getCurrentBuffer()->setModifiedStatus(true);`. However, before performing undo and redo, the modification actions have actually been completed (after all, modifications must exist before undo and redo can be performed, right?). So, can we just focus on monitoring the modification actions instead?

donho:
@softmgr 
I guess the 2nd commit has been reviewed by you, and there's no issue.
However, there's one question has been left:
https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15981/commits/07bcfcbd5a6222ca14ff27baaf6d5c16deccf10d?diff=unified&w=0#r1898788988

`|| (_currentBuffer->isUntitled()` can be removed safely IMO.

We can enhance redo & undo afterward. But it is not in the scope of this PR.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
donho: OK, I see the problem when `|| (_currentBuffer->isUntitled())` is removed.
So it should be kept.

PowerEditor/src/ScintillaComponent/ScintillaEditView.cpp
```diff
@@ -2250,7 +2260,29 @@ void ScintillaEditView::restyleBuffer()
...
+		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
+
+		defineDocType(_currentBuffer->getLangType());
+
+		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF);
+		execute(SCI_SETDOCPOINTER, 0, prevDoc);
+		execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON);
+	}
+
+	_currentBuffer->setLastLangType(currentLangInt);
```
donho, line 2285: `void ScintillaEditView::styleChange()` is called when user stays on the same document and is adding/deleting/modifying UDL element on UDL dialog. The aim of calling the fuction is updating the visual effect of document while the modification. 
As a result, the above code is not necessary.

donho:
The 2nd commit brought a regression. The 3rd commit has fixed it.

softmgr:
> The 2nd commit brought a regression. The 3rd commit has fixed it.

After reviewing, I confirm that the `restyleBuffer()` function does not need to be modified.

donho:
@softmgr 
FYI
In the latest commit SCI_SETMODEVENTMASK are removed for the non-mainEditView & non-subEditView

Only main-EditView & sub-EditView's notification codes (SC_MOD_DELETETEXT | SC_MOD_INSERTTEXT | SC_PERFORMED_UNDO | SC_PERFORMED_REDO) in SCN_MODIFIED notification:
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/PowerEditor/src/NppNotification.cpp#L49

So it's not necessary to turn off/on the notification of modification for other Scintilla controls.
OTOH, it'll be better to turn off the notification of modification completely for them. I'll do it in the next PR.

donho:
```cpp
			//if (notification->modificationType & SC_MOD_CHANGEINDICATOR)
			//{
			//	::InvalidateRect(notifyView->getHSelf(), NULL, FALSE);
			//}
```
Remove these lines make the regression.
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8685

softmgr:
> ```cpp
> 
> 			//if (notification->modificationType & SC_MOD_CHANGEINDICATOR)
> 
> 			//{
> 
> 			//	::InvalidateRect(notifyView->getHSelf(), NULL, FALSE);
> 
> 			//}
> 
> ```
> 
> Removing these lines makes the regression.
> 
> Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/8685


Since I removed the "Clickable Link" feature in the modified code, I didn’t test this part of the issue (I use the context menu to implement copying and opening of clickable link). The reason, as mentioned earlier, was to improve performance.

donho:
> Since I removed the "Clickable Link" feature in the modified code, I didn’t test this part of the issue (I use the context menu to implement copying and opening of clickable link). The reason, as mentioned earlier, was to improve performance.

OK, I see.
Removing "Clickable Link" feature will make a lot of people complain. So it will be kept.

softmgr:
> > Since I removed the "Clickable Link" feature in the modified code, I didn’t test this part of the issue (I use the context menu to implement copying and opening of clickable link). The reason, as mentioned earlier, was to improve performance.
> 
> 
> 
> OK, I see.
> 
> Removing "Clickable Link" feature will make a lot of people complain. So it will be kept.

After removing the clickable link feature, I used the "Copy Link" and "Open File" options in the context menu to achieve the same functionality. The advantage of this approach is that it significantly reduces the refresh time during each rendering. Other editors, such as Notepad4, adopt a similar approach.

donho:
User can always  disable Clickable link feature.

Since SC_MOD_CHANGEINDICATOR is triggered when an indicator has been added or removed from a range of text, I don't think the following line will make any impact:
```cpp
			if (notification->modificationType & SC_MOD_CHANGEINDICATOR)
			{
				::InvalidateRect(notifyView->getHSelf(), NULL, FALSE);
			}
```

softmgr:
> User can always  disable Clickable link feature.
> 
> 
> 

Based on my testing, the `addHotSpot()` function is executed regardless of whether `Clickable Link Settings` in the `Preferences` is enabled or disabled. This function runs during interface refresh events, such as typing characters, scrolling, or switching tabs.
For smaller files, the execution is so quick that it’s unnoticeable, but for larger files, it can still have some impact.

donho:
> Based on my testing, the `addHotSpot()` function is executed regardless of whether `Clickable Link Settings` in the `Preferences` is enabled or disabled. This function runs during interface refresh events, such as typing characters, scrolling, or switching tabs.
> For smaller files, the execution is so quick that it’s unnoticeable, but for larger files, it can still have some impact.

@softmgr 
Please check PR #16021 - the issue you mentioned above is fixed in the PR.

rdipardo:
@softmgr

> Through testing, I discovered that if two documents have the same language type, it is possible to skip re-executing the defineDocType() function during document switching.

What about *reloading* the current buffer: #16027 ?

softmgr:
> @softmgr
> 
> 
> 
> > Through testing, I discovered that if two documents have the same language type, it is possible to skip re-executing the defineDocType() function during document switching.
> 
> 
> 
> What about *reloading* the current buffer: #16027 ?

@donho 
After saving the file as a new file, if the file type changes, we should handle syntax highlighting in the same way as when the document was first opened. I modified the following code, which should resolve the issue you mentioned:

NppIO.cpp
```
bool Notepad_plus::fileSaveAs(BufferID id, bool isSaveCopy)
{
	// ...
			if (res && isSaveCopy && fDlg.getOpenTheCopyAfterSaveAsCopy())
			{
				BufferID bid = doOpen(fn);
				if (bid != BUFFER_INVALID)
				{
					switchToFile(bid);
				}
			}
			else if (res && langType != buf->getLangType()) // After saving the file, if the file type changes, it should be handled in the same way as when the document was first opened.
			{
				buf->setLastLangType(-1);
				activateBuffer(id, currentView(), true);
			}

			return res;
	// ...
}

ScintillaEditView.cpp
```
void ScintillaEditView::activateBuffer(BufferID buffer, bool force)
{
	// ...
	const bool isSameLangType = _prevBuffer != nullptr && (_prevBuffer->getLangType() == _currentBuffer->getLangType());
	// ...
}
```

**Comment by donho** [comment]
@softmgr 
> After saving the file as a new file, if the file type changes, we should handle syntax highlighting in the same way as when the document was first opened. I modified the following code, which should resolve the issue you mentioned:

The problem is not about "save as" action, but "reload" action. The issue #16027 can be reproduced just by reloading any opened existing files.

**Comment by softmgr** [comment]
> @softmgr 
> 
> > After saving the file as a new file, if the file type changes, we should handle syntax highlighting in the same way as when the document was first opened. I modified the following code, which should resolve the issue you mentioned:
> 
> 
> 
> The problem is not about "save as" action, but "reload" action. The issue #16027 can be reproduced just by reloading any opened existing files.
> 
> 

After modifying the code above, reloading the file now applies syntax highlighting correctly. As for code folding, it was not retained even before this code optimization: reloading the file always resets to an unfolded state, just like when the file is first opened.

**Comment by donho** [comment]
@softmgr 
The [above modification you provided](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15981#issuecomment-2581609660) is for:
1. `bool Notepad_plus::fileSaveAs(BufferID id, bool isSaveCopy)` which is about only "file save as" action.
2. `void ScintillaEditView::activateBuffer(BufferID buffer, bool force)` in which `(_prevBuffer->getLangType() == _currentBuffer->getLangType())` condition will be always true for reloading a file.

As a result I cannot see how the modification can remedy the issue.

Maybe I'm missing somethings, could you enlight me?

**Comment by softmgr** [comment]
> @softmgr The [above modification you provided](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/15981#issuecomment-2581609660) is for:
> 
> 1. `bool Notepad_plus::fileSaveAs(BufferID id, bool isSaveCopy)` which is about only "file save as" action.
> 2. `void ScintillaEditView::activateBuffer(BufferID buffer, bool force)` in which `(_prevBuffer->getLangType() == _currentBuffer->getLangType())` condition will be always true for reloading a file.
> 
> As a result I don't see how the modification can remedy the issue.
> 
> Maybe I'm missing somethings, could you enlight me?


@donho @rdipardo 
This time, there should be no issues.

NppIO.cpp
```
bool Notepad_plus::doReload(BufferID id, bool alert)
{
	// ...
	bool res = MainFileManager.reloadBuffer(id);
	Buffer * pBuf = MainFileManager.getBufferByID(id);
	pBuf->setLastLangType(-1); // added
	// ...
}
```

ScintillaEditView.cpp
```
void ScintillaEditView::activateBuffer(BufferID buffer, bool force)
{
	// ...
	const bool isSameLangType = _prevBuffer != nullptr && (_prevBuffer->getLangType() == _currentBuffer->getLangType());
	// ...
}
```
Buffer.cpp
```
SavingStatus FileManager::saveBuffer(BufferID id, const wchar_t* filename, bool isCopy)
{
	// ...
		if (isCopy) // "Save a Copy As..." command
		{
			//_pscratchTilla->execute(SCI_SETMODEVENTMASK, MODEVENTMASK_OFF); // removed
			_pscratchTilla->execute(SCI_SETDOCPOINTER, 0, _scratchDocDefault);
			//_pscratchTilla->execute(SCI_SETMODEVENTMASK, MODEVENTMASK_ON); // removed
			return SavingStatus::SaveOK;	//all done - we don't change the current buffer's path to "fullpath", since it's "Save a Copy As..." action.
		}
	// ...
}

donho:
@softmgr 
Thank you for your code. However, I prefer this appraoch: #16042

Issue #15992, opened 2024-12-27
[BUG] Text replacement in selected text, second time, Replace All In Selection


Sediqwe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the Issue

If I replace a given text, so that I want it in the selected text, it will replace the whole document if it is repeated, even if the checkbox is apparently checked!

### Steps To Reproduce

1.  Select text section
2.  ctrl - h
3.  mark "in selection" (Extended)
4. Replace ALL -> OK
5. Another text section select 
6 Replace ALL  -> Replace in ALL text! (Not only in the selected)

### Current Behavior

The text has not only been replaced in the selected area!

### Expected Behavior

Only the field is replaced because I did not uncheck the checkbox

### Debug Information

```shell
Notepad++ v8.7.1   (64-bit)
Build time : Oct 31 2024 - 00:48:56
Path : C:\Program Files\Notepad++\notepad++.exe
Command Line : "C:\Program Files\Notepad++\change.log" 
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
Periodic Backup : ON
OS Name : Windows 11 Pro (64-bit)
OS Version : 23H2
OS Build : 22631.4602
Current ANSI codepage : 1250
Plugins : 
    ColumnTools (1.4.5.1)
    ComparePlugin (2.0.2)
    CSVLint (0.4.6.5)
    HTMLTag (1.4.1)
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    NppRegExTractorPlugin (2.1)
    RegexTrainer (1)
    XMLTools (3.1.1.13)
```

alankilborn:
Duplicate of #14897 
See workaround there until this problem is dealt with.

fir999:
@alankilborn: The workaround, setting the option _Searching set Minimum Size for Auto-Checking "In selection"_ to _0_, solveld my problem. Thanks!

Issue #14624, opened 2024-01-21
"In Selection" becomes unchecked on clicking "Replace All" 


Gilgamech:
### Description of the Issue
The behavior of the `In Selection` checkbox is highly inconsistent. Usually it is in the wrong mode, so I'm used to clicking the checkbox after opening the Replace window. I understand that there is other logic that attempts to determine the correct mode, and am willing to work around that.

And until recently, the checkbox would remain the same as long as the dialog box is open. Now, after highlighting and replacing text in a section with the checkbox checked, highlighting another section (and without closing or modifying the Replace dialog) causes the `In Selection` checkbox to become unchecked. But the UI doesn't update until it gains focus again, which occurs when I click the `Replace All` button. 

### Steps to Reproduce the Issue
1. Have a document with text in it. Highlight a multi-line section
2. Press `Ctrl+R` to open the Replace window and check the `In Selection` checkbox.
3. Highlight a section of text., and press `Replace All`. 
4. Highlight another section of text and press `Replace All`. 

### Expected Behavior
The replaceable text only within the selection would be replaced.

### Actual Behavior
The replaceable text throughout the document was replaced.

### Debug Information
Notepad++ v8.6.2   (64-bit)
Build time : Jan 14 2024 - 02:16:00
Path : C:\Program Files\n++\notepad++.exe
Command Line : "C:\repos\website\www\calc.md"
Admin mode : OFF
Local Conf mode : OFF
Cloud Config : OFF
OS Name : Windows 11 Home (64-bit)
OS Version : 22H2
OS Build : 22621.3007
Current ANSI codepage : 1252
Plugins : 
    AutoSave (1.6.1)
    ComparePlugin (2.0.2)
    mimeTools (3)
    NppConverter (4.5)
    NppExport (0.4)

I don't mind that the checkbox is usually in the wrong state. What I mind is that it changes state while the Replace window is open, and that its behavior continues to change. It seems as though Notepad++ is determined to always replace across the wrong set, necessitating a `Ctrl+Z`, re-check selection, re-check settings, and `Replace All` again. Which is an inefficient workflow.

alankilborn:
Note: Step 2 should say `Ctrl+h`, not `Ctrl+r`.

---

I think setting this avoids the above problem:

![image](https://github.com/notepad-plus-plus/notepad-plus-plus/assets/77065706/3b1bef77-70dd-45d6-a5e6-395a3a590843)

[...]

PR #15995, opened 2024-12-30
Prevent search from running with unintended In-selection parameter

### LINKED ISSUE #14897: ReplaceAll can execute opposite of user InSelection intention
### Issue Description

Logic executes to auto-set or auto-clear the _In selection_ checkbox when the _Replace_ window becomes the active window, depending upon the amount of selected text in the recently-active editor window.  Because pressing _Replace All_ on an inactive _Replace_ window will activate the window, and then run the _Replace All_ functionality, it is possible for the _In selection_ checkbox to appear in one state to the user (at the moment of button press), but yet be in the opposite state when the _Replace All_ is actually conducted.

1. In _Preferences_ > _Searching_ set _Minimum Size for Auto-Checking "In selection"_ to 6.
2. Create a new Notepad++ tab with this text:
```txt
abcdefg
abcdefg
abcdefg
```
3. Select `a` through `g` on the line 1 and press Ctrl+h to invoke the _Replace_ window.
4. Specify _Find what_ of `c`.
5. Specify _Replace with_ of `Z`.
6. Note that _In selection_ is checkmarked (automatically, because 7 characters were selected when replace was invoked).
7. Checkmark _Wrap around_ and uncheckmark everything else (except _In selection_).
8. Press the _Replace All_ button.
9. Observe (correct replacement):
```txt
abZdefg
abcdefg
abcdefg
```
10. Leave the _Replace_ window open and move input focus back to the editor window.
11. Select `b` through `d` on line 2.
12. Without otherwise activating the _Replace_ window, and noting that _In selection_ is still checkmarked, press the _Replace All_ button.

### Actual behavior:
```txt
abZdefg
abZdefg
abZdefg
```
Line 3 was changed.

### Expected behavior:
```txt
abZdefg
abZdefg
abcdefg
```
Line 3 should not have been changed and should still have a `c` in it.

### Notes

This can also "go the other way", where the user sees an uncheckmarked _In selection_ at the moment of _Replace All_ press, and then observes that _only_ the selected text affected by the replacement.

This is a restatement of #14624 which sort of "lost its way".

Workaround:  Activate _Replace_ without clicking on an activatable control (e.g., click the _Replace_ window's title bar, then alter _In selection_ if needed, then finally press _Replace All_.

### Debug Information
Notepad++ v8.6.4   (64-bit)
Build time : Feb 20 2024 - 00:12:59
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.6.4\npp.8.6.4.portable.x64\notepad++.exe
Command Line : -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
OS Version : 23H2
OS Build : 22631.3155
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)

alankilborn:
Fix #14897 #14624 #15992

### DIFF
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1514,7 +1514,7 @@ Do you want to continue?"/>
 please give another one."/>
 			<UDLRemoveCurrentLang title="Remove the current language" message="Are you sure?"/>
 			<SCMapperDoDeleteOrNot title="Are you sure?" message="Are you sure you want to delete this shortcut?"/>
-			<FindCharRangeValueError title="Range Value problem" message="You should type between 0 and 255."/>
+			<FindCharRangeValueError title="Range Value problem" message="You should type between 0 and 255."/> <!-- HowToReproduce: Search menu, then Find characters in range, select Custom range, enter 999 in either edit box, press Find. -->
 			<OpenInAdminMode title="Save failed" message="The file cannot be saved and it may be protected.
 Do you want to launch Notepad++ in Administrator mode?"/>
 			<OpenInAdminModeWithoutCloseCurrent title="Save failed" message="The file cannot be saved and it may be protected.
@@ -1555,6 +1555,7 @@ Press the OK button to open the Find dialog or set focus on it.
 			
 If you require the backward regex searching feature, consult the user manual for instructions on enabling it."/>
 			<PrintError title="0" message="Cannot start printer document."/><!-- Use title="0" to use Windows OS default translated "Error" title. -->
+			<FindAutoChangeOfInSelectionWarning title="Search warning" message="The &quot;In selection&quot; checkbox state has been automatically modified. Please verify the search condition before performing the action."/> <!-- HowToReproduce: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14897#issuecomment-2564316224 -->
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Clipboard History"/>
--- a/PowerEditor/installer/nativeLang/english.xml
+++ b/PowerEditor/installer/nativeLang/english.xml
@@ -1555,7 +1555,8 @@ Press the OK button to open the Find dialog or set focus on it.
 			
 If you require the backward regex searching feature, consult the user manual for instructions on enabling it."/>
 			<PrintError title="0" message="Cannot start printer document."/><!-- Use title="0" to use Windows OS default translated "Error" title. -->
-			<FindAutoChangeOfInSelectionWarning title="Search warning" message="The &quot;In selection&quot; checkbox state has been automatically modified. Please verify the search condition before performing the action."/> <!-- HowToReproduce: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14897#issuecomment-2564316224 -->
+			<FindAutoChangeOfInSelectionWarning title="Search warning" message="The &quot;In selection&quot; checkbox state has been automatically modified.
+Please verify the search condition before performing the action."/> <!-- HowToReproduce: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14897#issuecomment-2564316224 -->
 		</MessageBox>
 		<ClipboardHistory>
 			<PanelTitle name="Clipboard History"/>
--- a/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
+++ b/PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
@@ -1692,6 +1692,8 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 
 		case WM_ACTIVATE :
 		{
+			bool inSelAutoChangeSoAbortSearch = false;
+
 			if (LOWORD(wParam) == WA_ACTIVE || LOWORD(wParam) == WA_CLICKACTIVE)
 			{
 				Sci_CharacterRangeFull cr = (*_ppEditView)->getSelection();
@@ -1708,7 +1710,7 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 
 				enableFindDlgItem(IDC_IN_SELECTION_CHECK, inSelEnabled);
 
-				bool inSelChecked = isCheckedOrNot(IDC_IN_SELECTION_CHECK);
+				bool inSelChecked = isCheckedOrNot(IDC_IN_SELECTION_CHECK), origInSelChecked = inSelChecked;
 
 				const NppGUI& nppGui = (NppParameters::getInstance()).getNppGUI();
 				if (nppGui._inSelectionAutocheckThreshold != 0)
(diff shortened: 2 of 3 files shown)


alankilborn:
@xomx @donho Comments?

alankilborn:
image

xomx:
@alankilborn 
It will take me some time to get back to some coding for N++. So for now, I will only comment from a user's perspective.

- I tried the PR binary and it works as promised, now I am stopped when I try to directly click on some action button in the opened semi transparent Find-dlg if the "In selection" gets auto-changed somehow, good!
- maybe also add a tooltip for the Find-dlg "In selection" checkbox pointing to the Preferences setting/explanation(?)
- but I have to admit that I have here the "manual" control myself (Preferences > Searching > Minimum Size for Auto-Checking "In selection": 0) , I do not like this automation much as it has already bitten me several times

alankilborn:
@xomx Thank you for trying it and for your comments.

I suppose I was specifically looking for comments about any potential downsides to the use of PeekMessage in this application.

>I do not like this automation much as it has already bitten me several times

Maybe this PR, in its final form, will take the "bite" out of the feature.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1746,7 +1772,18 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 			{
 				enableFindDlgItem(IDREDOTMATCHNL, false);
 			}
+
 			enableProjectCheckmarks();
+
+			if (inSelAutoChangeSoAbortSearch)
+			{
+				// TODO: center on Find, respect dark mode, localization
+				::MessageBox(NULL, L"The state of 'In selection' has automatically "
```
donho, line 1781: The message box should be modal for blocking of Find/Replace dialog's access.

donho:
> image

I would use:
***The "In Selection" checkbox state has been automatically modified.***
***Please verify the search condition before performing the replace or mark action.***

donho (changes requested):
1. Please set the message box modal.
2. Please modify the content of message.
3. Make the message localizable.

alankilborn:
@donho said:

> "Please verify the search condition before performing the replace or mark action."

The bug can happen with _Find All in Current Document_ or _Count_ actions as well, that's why the text I used was more generic and simply said "search operation".

Wouldn't user be confused if he pressed _Count_ and the msgbox said "...before performing the replace or mark action" ?

alankilborn:
How about simply saying "Please verify the search condition before continuing" ??

donho:
> Wouldn't user be confused if he pressed Count and the msgbox said "...before performing the replace or mark action" ?

I see.
Let's use ***Please verify the search condition before performing the action.***

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1720,6 +1722,30 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
 					setChecked(IDC_IN_SELECTION_CHECK, inSelChecked);
 				}
```
donho, line 1724: Please add the following comment here:
/*
In the scenario where the user clicks the action button (Count, Find All in Current Document, Replace All, Mark All, or Clear All marks) without activating the Find/Replace dialog, the "In Selection" checkbox could be auto-changed after the button click. To prevent the search from running with this unintended state, the search message has been removed from the queue. Then, launch a message box to alert the user that the search didn't run and they need to verify the settings.
*/

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1746,7 +1772,23 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
...
 				enableFindDlgItem(IDREDOTMATCHNL, false);
 			}
+
 			enableProjectCheckmarks();
+
+			if (inSelAutoChangeSoAbortSearch)
+			{
+				// show this to the user only after all UI changes have been made,
+				// so user doesn't see something out-of-date if he looks at the
+				// search window while the msgbox is still open
```
donho, line 1782: Please remove the comment - the message auto-explains our intention.

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1746,7 +1772,23 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
...
+				NppParameters& nppParamInst = NppParameters::getInstance();
+				(nppParamInst.getNativeLangSpeaker())->messageBox(
+					"FindAutoChangeOfInSelectionWarning",
+					_hSelf,
+					L"The \"In selection\" checkbox state has been automatically modified. Please verify the search condition before performing the action.",
```
donho, line 1787: Please break down the one line sentence `The "In selection" checkbox state has been automatically modified. Please verify the search condition before performing the action.` into 2 lines.

PowerEditor/installer/nativeLang/english.xml
```diff
@@ -1555,6 +1555,7 @@ Press the OK button to open the Find dialog or set focus on it.
 			
 If you require the backward regex searching feature, consult the user manual for instructions on enabling it."/>
 			<PrintError title="0" message="Cannot start printer document."/><!-- Use title="0" to use Windows OS default translated "Error" title. -->
+			<FindAutoChangeOfInSelectionWarning title="Search warning" message="The &quot;In selection&quot; checkbox state has been automatically modified. Please verify the search condition before performing the action."/> <!-- HowToReproduce: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/14897#issuecomment-2564316224 -->
```
donho, line 1558: Please break down the one line sentence `The "In selection" checkbox state has been automatically modified. Please verify the search condition before performing the action.` into 2 lines.

donho (changes requested):
Please consider the requested changes

PowerEditor/src/ScintillaComponent/FindReplaceDlg.cpp
```diff
@@ -1720,6 +1723,32 @@ intptr_t CALLBACK FindReplaceDlg::run_dlgProc(UINT message, WPARAM wParam, LPARA
...
+				Then, launch a message box to alert the user that the search didn't run and 
+				they need to verify the settings.
+				*/
+				if (inSelChecked != origInSelChecked)
+				{
+					const auto inSelActionIds = { IDCCOUNTALL, IDC_FINDALL_CURRENTFILE, IDREPLACEALL, IDCMARKALL, IDC_CLEAR_ALL };
+					for (auto id : inSelActionIds)
+					{
+						MSG msg;
+						while (PeekMessage(&msg, ::GetDlgItem(_hSelf, id), 0, 0, PM_REMOVE))
```
donho, line 1741: @alankilborn 
Sorry for not pointing out the problem on the previous review:
Though there must be the message of button clicked action logically, the while loop could make a dead lock if there's no message in the queue. According the [online doc](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-peekmessagea):
`If no messages are available, the return value is zero.`

As a result, I would suggest the following code for preventing infinit loop & improving the readability:
```cpp
if (inSelChecked != origInSelChecked)
{
	const std::vector<int> inSelActionIds = { IDCCOUNTALL, IDC_FINDALL_CURRENTFILE, IDREPLACEALL, IDCMARKALL, IDC_CLEAR_ALL };
	for (const auto& id : inSelActionIds)
	{
		MSG msg;
		if (PeekMessage(&msg, ::GetDlgItem(_hSelf, id), 0, 0, PM_REMOVE))
		{
			isInSelectionAutoChange = true;
			break;
		}
	}
}
```

donho (changes requested):

PR #15998, opened 2024-12-31
Remove extraneous space in find status message.

asdfchlwnsgy1236:
FIND_STATUS_TOP_REACHED_TEXT has two spaces after "Find:" instead of one space like all other status messages do. This commit fixes that inconsistency.


chcg:
See https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/CONTRIBUTING.md

> PR of reformatting (changing of ws/TAB, line endings or coding style) of source code won't be accepted. Use issue trackers for your request instead.

Anyhow most likely this define is not used but the text from the translation files.

asdfchlwnsgy1236:
Right, of course it would be.
And I forgot to even check for the CONTRIBUTING.md file....
I really should not have made a PR while sick.
I will close this then.
Sorry about this....

PR #16010, opened 2025-01-03
Improve performance: Optimize activateBuffer function

donho:
Ref: https://github.com/notepad-plus-plus/notepad-plus-plus/issues/15952#issuecomment-2567521808


softmgr (approved):
The code works correctly and effectively improves the performance of file switching.

PR #16030, opened 2025-01-08
[XML] Update brazilian_portuguese.xml

Marcellomco:
Added translation for "Fix wrong replace all while 2nd time replace in selection"


Marcellomco (commented):
Added translation for "Add localization entry & clean up"

Marcellomco (commented):
Added translation for "GUI enhancement: hide right menu shorcuts on the fly"

PR #16047, opened 2025-01-11
Fix backup file renaming bug for untitled tabs

### LINKED ISSUE #16043: [BUG] Backup filename corruption when renaming untitled tab

### Description of the Issue

See STR.

### Steps To Reproduce

1.  Press Ctrl+n to create a new tab with name, e.g. `new 22`
2.  Type some text into the new tab (so that it is modified)
3.  Right-click the tab and rename it to `w`
4.  Look at the N++ "backup" folder in Windows Explorer.

### Current Behavior

The backup filename is `new 22@2025-01-10w0`.

### Expected Behavior

The backup filename should be, e.g. `w@2025-01-10_082007`.

### Debug Information

```shell
Notepad++ v8.7.5   (64-bit)
Build time : Dec 21 2024 - 05:13:03
Path : C:\NotepadPlusPlus\Misc\_NoBackup\Releases\npp.8.7.5\npp.8.7.5.portable.x64\notepad++.exe
Command Line : -titleAdd="8.7.5 64-bit" -multiInst
Admin mode : OFF
Local Conf mode : ON
Cloud Config : OFF
Periodic Backup : ON
Placeholders : OFF
DirectWrite : ON
Dark Mode : OFF
OS Version : 23H2
OS Build : 22631.4602
Current ANSI codepage : 1252
Plugins : 
    mimeTools (3.1)
    NppConverter (4.6)
    NppExport (0.4)
    PythonScript (3.0.20)
```

### Anything else?

_No response_

alankilborn:
Fix #16043

### DIFF
--- a/PowerEditor/src/NppIO.cpp
+++ b/PowerEditor/src/NppIO.cpp
@@ -2108,9 +2108,7 @@ bool Notepad_plus::fileRename(BufferID id)
 						return success;
 
 					wstring newBackUpFileName = oldBackUpFileName;
-
-					size_t index = newBackUpFileName.find_last_of(oldFileNamePath) - oldFileNamePath.length() + 1;
-					newBackUpFileName.replace(index, oldFileNamePath.length(), tabNewNameStr);
+					newBackUpFileName.replace(newBackUpFileName.rfind(oldFileNamePath), oldFileNamePath.length(), tabNewNameStr);
 
 					if (doesFileExist(newBackUpFileName.c_str()))
 						::ReplaceFile(newBackUpFileName.c_str(), oldBackUpFileName.c_str(), nullptr, REPLACEFILE_IGNORE_MERGE_ERRORS | REPLACEFILE_IGNORE_ACL_ERRORS, 0, 0);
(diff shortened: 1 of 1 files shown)


PowerEditor/src/NppIO.cpp
```diff
@@ -2179,15 +2178,14 @@ bool Notepad_plus::fileRenameUntitledPluginAPI(BufferID id, const wchar_t* tabNe
 	bool isSnapshotMode = NppParameters::getInstance().getNppGUI().isSnapshotMode();
 	if (isSnapshotMode)
 	{
-		wstring oldName = buf->getFullPathName();
```
donho, line 2182: The local variable should be declared in the scope where it is used.
Is there any reason to move out `oldName` from its scope?

donho (changes requested):

PowerEditor/src/NppIO.cpp
alankilborn: >Is there any reason to move out oldName from its scope?

Yes.

`wstring oldName = buf->getFullPathName();`

needs to be BEFORE:

`buf->setFileName(tabNewNameStr.c_str());`

otherwise, the "old name" is lost.

donho: @alankilborn 
Oups! You're right about it!

donho (approved):

PR #16065, opened 2025-01-14
GUI enhancement: hide right menu shorcuts on the fly

donho:
Hide right menu shorcuts on the fly in Preferences->General->Menu: `Hide right shortcuts...`


donho:
3921812175b7c537106af934ba4b9cbd4350c89f

